PR #7515, opened 2018-03-08
[WASM] Initial code drop for debugger

kumpera:
This supports setting, resolving and hitting breakpoints.

This support reporting stack frames and their params & locals.

### DIFF
--- a/mono/metadata/object-offsets.h
+++ b/mono/metadata/object-offsets.h
@@ -183,7 +183,6 @@ DECL_OFFSET(MonoContext, wasm_bp)
 DECL_OFFSET(MonoContext, wasm_sp)
 DECL_OFFSET(MonoContext, llvm_exc_reg)
 
-DECL_OFFSET(MonoLMF, method)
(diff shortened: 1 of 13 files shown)


kumpera:
The two failures are unrelated.

mono/metadata/object-offsets.h
```diff
@@ -183,7 +183,6 @@ DECL_OFFSET(MonoContext, wasm_bp)
 DECL_OFFSET(MonoContext, wasm_sp)
 DECL_OFFSET(MonoContext, llvm_exc_reg)
 
-DECL_OFFSET(MonoLMF, method)
```
lewurm: why was this removed? this breaks at least ARM (it likely others): https://github.com/mono/mono/blob/99198da52bee6d5a648a3e92923693bf56cb4d18/mono/mini/tramp-arm.c#L245

lewurm: nevermind, I didn't see the `ifdef` above.

PR #3165, opened 2016-06-14
[utils] Work around CLOCK_MONOTONIC being defined on macOS 10.12 SDK despite targeting earlier releases

akoeplinger:
Should fix https://bugzilla.xamarin.com/show_bug.cgi?id=41786

/cc @rolfbjarne

PR #7520, opened 2018-03-08
[mono-time] Replace mono_msec_boottime() with CoreCLR implementation.

### LINKED ISSUE #7056: Environment.TickCount not monotonic
This may or may not be a platform problem with our devices but our old devices running a different operating system had Environment.TickCount as monotonic. 

What I mean by this is that changing the date/time of the device/system is not supposed to effect the Environment.TickCount.

## Steps to Reproduce

1. Run c# application printing Environment.TickCount
2. Change the date/time of the device using date +%T -s "00:00:00"
3. See that TickCount changed when changing the time. If you added a hour to the time, basically a hour of milliseconds would be added to the tickcount.

## Current Behavior

Changing the date/time of the device Changes the Environment.TickCount.

## Expected Behavior

Environment.TickCount should be monotonic and not change if time is changed.

[x ] Linux

**Version Used**:

5.4.1.7

alexischr:
Fixes #7056
Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=58413

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1654,6 +1654,33 @@ if test x$platform_android = xyes; then
 fi
 
 if test x$host_win32 = xno; then
+	dnl *************************************
+	dnl *** Checks for time capabilities ***
+	dnl *************************************
+
+	AC_MSG_CHECKING(for CLOCK_MONOTONIC)
+	AC_TRY_COMPILE([#include <time.h>], [
+		const int foo = CLOCK_MONOTONIC;
+	],[
+		AC_MSG_RESULT(yes)
+		AC_DEFINE(HAVE_CLOCK_MONOTONIC, 1, [CLOCK_MONOTONIC])
+	], [
+		AC_MSG_RESULT(no)
+	])
+
+	AC_MSG_CHECKING(for CLOCK_MONOTONIC_COARSE)
+	AC_TRY_COMPILE([#include <time.h>], [
+		const int foo = CLOCK_MONOTONIC_COARSE;
+	],[
+		AC_MSG_RESULT(yes)
+		AC_DEFINE(HAVE_CLOCK_MONOTONIC_COARSE, 1, [CLOCK_MONOTONIC_COARSE])
+	], [
+		AC_MSG_RESULT(no)
+	])
+
+	AC_CHECK_FUNC(mach_absolute_time, [AC_DEFINE(HAVE_MACH_ABSOLUTE_TIME, 1, [mach_absolute_time])])
+	AC_CHECK_FUNC(gethrtime, [AC_DEFINE(HAVE_GETHRTIME, 1, [gethrtime])])
+	AC_CHECK_FUNC(read_real_time, [AC_DEFINE(HAVE_READ_REAL_TIME, 1, [read_real_time])])
 
 	dnl hires monotonic clock support
 	AC_SEARCH_LIBS(clock_gettime, rt)
--- a/mono/utils/mono-time.c
+++ b/mono/utils/mono-time.c
@@ -9,6 +9,7 @@
 #include <config.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <errno.h>
 
 #ifdef HAVE_SYS_TIME_H
 #include <sys/time.h>
@@ -18,6 +19,18 @@
 
 
 #define MTICKS_PER_SEC (10 * 1000 * 1000)
+
+typedef enum _TimeConversionConstants
+{
+	tccSecondsToMillieSeconds       = 1000,         // 10^3
+	tccSecondsToMicroSeconds        = 1000000,      // 10^6
+	tccSecondsToNanoSeconds         = 1000000000,   // 10^9
+	tccMillieSecondsToMicroSeconds  = 1000,         // 10^3
+	tccMillieSecondsToNanoSeconds   = 1000000,      // 10^6
+	tccMicroSecondsToNanoSeconds    = 1000,         // 10^3
+	tccSecondsTo100NanoSeconds      = 10000000,     // 10^7
+	tccMicroSecondsTo100NanoSeconds = 10            // 10^1
+} TimeConversionConstants;
 
 gint64
 mono_msec_ticks (void)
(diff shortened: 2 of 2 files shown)


mono/utils/mono-time.c
```diff
@@ -126,16 +139,66 @@ get_boot_time (void)
...
+    {
+        clockid_t clockType =
+#if HAVE_CLOCK_MONOTONIC_COARSE
+            CLOCK_MONOTONIC_COARSE; // good enough resolution, fastest speed
+#else
+            CLOCK_MONOTONIC;
+#endif
+        struct timespec ts;
+        if (clock_gettime(clockType, &ts) != 0)
+        {
```
luhenry: Adapt to Mono's code convention.

mono/utils/mono-time.c
```diff
@@ -126,16 +139,66 @@ get_boot_time (void)
...
-	static gint64 boot_time = 0;
-	gint64 now;
-	if (!boot_time)
-		boot_time = get_boot_time ();
-	now = mono_100ns_datetime ();
-	/*printf ("now: %llu (boot: %llu) ticks: %llu\n", (gint64)now, (gint64)boot_time, (gint64)(now - boot_time));*/
-	return (now - boot_time)/10000;
+    uint64_t retval = 0;
+
+#if HAVE_CLOCK_MONOTONIC_COARSE || HAVE_CLOCK_MONOTONIC
```
kumpera: Where are the configure.ac changes to detect all this new stuff?

alexischr: WIP folks :) we already have `HAVE_CLOCK_MONOTONIC` so the for the fix we want this should be fine.

alexischr: @kumpera lol nope, I thought we had it as I was grepping in Ludo's System.Native integration branch where he brought them over. I'll add them :)

mono/utils/mono-time.c
```diff
@@ -126,16 +139,66 @@ get_boot_time (void)
...
+        if (s_TimebaseInfo.denom == 0)
+        {
+            g_error ("s_TimebaseInfo is uninitialized.\n");
+            goto EXIT;
+        }
+        retval = (mach_absolute_time() * s_TimebaseInfo.numer / s_TimebaseInfo.denom) / tccMillieSecondsToNanoSeconds;
+    }
+#elif HAVE_GETHRTIME
+    {
+        retval = (ULONGLONG)(gethrtime() / tccMillieSecondsToNanoSeconds);
```
alexischr: this won't compile, will fix

mono/utils/mono-time.c
```diff
@@ -126,16 +139,57 @@ get_boot_time (void)
...
 mono_msec_boottime (void)
 {
-	static gint64 boot_time = 0;
-	gint64 now;
-	if (!boot_time)
-		boot_time = get_boot_time ();
-	now = mono_100ns_datetime ();
-	/*printf ("now: %llu (boot: %llu) ticks: %llu\n", (gint64)now, (gint64)boot_time, (gint64)(now - boot_time));*/
-	return (now - boot_time)/10000;
+    gint64 retval = 0;
```
alexrp: This code is using spaces instead of tabs for indentation.

jaykrell: Consistent rendering. No editor settings required. Typical spaces=4 uses screen better than typical tab=8. Alas. :(

configure.ac
```diff
@@ -1654,6 +1654,49 @@ if test x$platform_android = xyes; then
...
+		AC_MSG_RESULT(yes)
+		AC_DEFINE(HAVE_CLOCK_MONOTONIC, 1, [CLOCK_MONOTONIC])
+	], [
+		AC_MSG_RESULT(no)
+	])
+
+	AC_MSG_CHECKING(for CLOCK_MONOTONIC_COARSE)
+	AC_TRY_RUN([
+		#include <stdlib.h>
+		#include <time.h>
```
jaykrell: Should probably instead just compile (not run) extern const int foo; extern const int foo = clock_monotic; You don't need main and you don't need to link or run. This is more cross build friendly and more efficient. Autoconf is slow.

mono/utils/mono-time.c
```diff
@@ -19,6 +20,18 @@
...
 
+typedef enum _TimeConversionConstants
+{
+    tccSecondsToMillieSeconds       = 1000,         // 10^3
+    tccSecondsToMicroSeconds        = 1000000,      // 10^6
+    tccSecondsToNanoSeconds         = 1000000000,   // 10^9
+    tccMillieSecondsToMicroSeconds  = 1000,         // 10^3
+    tccMillieSecondsToNanoSeconds   = 1000000,      // 10^6
+    tccMicroSecondsToNanoSeconds    = 1000,         // 10^3
+    tccSecondsTo100NanoSeconds      = 10000000,     // 10^7
```
jaykrell: Personally I think "100NanoSeconds" should be called "NtTime".

mono/utils/mono-time.c
```diff
@@ -126,16 +139,57 @@ get_boot_time (void)
...
+    gint64 retval = 0;
+
+#if HAVE_CLOCK_MONOTONIC_COARSE || HAVE_CLOCK_MONOTONIC
+    clockid_t clockType =
+#if HAVE_CLOCK_MONOTONIC_COARSE
+        CLOCK_MONOTONIC_COARSE; /* good enough resolution, fastest speed */
+#else
+        CLOCK_MONOTONIC;
+#endif
+    struct timespec ts;
```
jaykrell: I like to zero-initialize, but this might actually be performance sensitive.

mono/utils/mono-time.c
```diff
@@ -126,16 +139,57 @@ get_boot_time (void)
...
+    timebasestruct_t tb;
+    read_real_time (&tb, TIMEBASE_SZ);
+    if (time_base_to_time (&tb, TIMEBASE_SZ) != 0) {
+        g_error ("time_base_to_time() failed; errno is %d (%s)", errno, strerror (errno));
+        goto exit;
+    }
+    retval = (tb.tb_high * tccSecondsToMillieSeconds)+(tb.tb_low / tccMillieSecondsToNanoSeconds);
+
+#else
+    struct timeval tv;
```
jaykrell: What do you get in error path? Maybe zero or -1-fll it at entry or in exit path?

mono/utils/mono-time.c
luhenry: `NtTime` implies implicit knowledge that it is 100ns, which is absolutely unclear to anybody who didn't do a lot of Windows dev.

configure.ac
alexrp: Indeed, we should not be adding new `AC_TRY_RUN` calls to configure unless the check absolutely can't function without being run. And even in that case, it should probably be a run-time decision in the C code (as we do for e.g. some WSL workarounds). `AC_TRY_RUN` is almost always the wrong thing to use.

kumpera: AC_TRY_RUN doesn't work with cross compiler and we'd have to special case all the new checks there as well.

alexischr:
@jaykrell @kumpera @alexrp switched to AC_TRY_COMPILE

alexischr:
@monojenkins build failed

akoeplinger:
@alexischr can you please verify this won't run into the same issue that was fixed by https://github.com/mono/mono/pull/3165 ?

alexischr:
@monojenkins build pkg

alexischr:
@monojenkins build pkg

alexischr:
@monojenkins build pkg

alexischr:
@akoeplinger I have ordered a MacSDK package and though it is failing (for an F# issue I am about to file) it does succeed in building/running mono with `MACOSX_DEPLOYMENT_TARGET=10.7` which if I remember the problem correctly was enough for the problem to show up. So I think we are fine.

Please correct me or thumbsup me :)

alexischr:
@monojenkins build pkg

alexischr:

alexischr:

akoeplinger:
@alexischr hmm, we should definitely test this on a pre-10.12 OSX now that we backported this to 2017-12.

akoeplinger:
we're disabling clock_getres in bockbuild since https://github.com/mono/bockbuild/commit/2c8891c41d097331a151fccb122acb73fa6d15c5, not sure how that interacts with the autoconf checks.

PR #7524, opened 2018-03-09
[Tailcall] Fix virtual generic methods that F# tests uses.

jaykrell:


dsyme:
I don't know enough (anything) about the Mono VM implementation to comment on that part - I mean I read through it, and the changes make sense, but I'm not a competent reviewer.

Re testing: 

1. Are we testing with SGEN?

2. Can we add tests that go across assembly boundaries?  That would involve splitting up some of the tests to got Assembly1 --> Assembly2 --> Assembly1 etc.  e.e. could be done via a pair of interfaces/classes, one class in each assembly.

3. Is the testing giving us any guarantees about whether this will work with Android and iOS backends?  Those are two we really care about.

jaykrell:
Just an FYI to you, but good questions.

I know ECMA says something about inter vs. intra assembly but I lost track of that.
This code should work the same inter or intra assembly. The same no-guarantee-but-we-are-getting better. I should reread the spec. And yes that extra testing wouldn't be a bad idea. This particular change doesn't yet have a good test story at all. We don't use F# in tests. I have to distill it down to IL an then if we want I can further split it into two assemblies for more test coverage.

Mono of course is a mix of portable and OS-specific and architecture-specific code.
We also have some funnyness where some architecture backends have more builtins than others.
So we have "classes" of processor architectures.
In particular, amd64, x86, arm32, arm64, do have better tailcall support than the others.

So, this change is very portable to all of those, yes.

Another recent change is so far AMD64-specific but that is just because it isn't finished.

There are underlying processor-specifities just below the surface that we might be able to reduce or are maybe stuck with. There are some processor-specific restrictions on tailcalls. I don't have a solid handle on them. But, like, for example, probably signatures do not have to match, if all the parameters are in registers, for example.

We do have the same priorities -- ios and android -- but amd64 is so far easiest for me so I do stuff there first. With a little more ramp up or maybe other people doing other processors, arm32, arm64, and x86 should follow.

The matrix is a little larger than that, because there  is JIT, AOT, FullAOT, LLVM, LLVMONLY, and interpreter. But there is much shared code among the scenarios, so it isn't necessarily that bad.

jaykrell:
We test with SGEN and Boehm but don't care much for Boehm.
You think it matters, with respect to tailcall?

jaykrell:
With this and earlier commits:
 F# tests pass AMD64 JIT, all but two tests. Those two are difficult and I believe CoreCLR fails them and more.
 AMD64 Full AOT same thing.
 AMD64 AOT hangs part way through.
 All other architectures are behind at least a little I believe (lack of tall.callvirt support).

dsyme:
> AMD64 AOT hangs part way through.

Can we do something about this hang - do we know the cause of it?

> With this and earlier commits..Those two are difficult and I believe CoreCLR fails them and more.
> AMD64 Full AOT same thing.

I wasn't aware of this failure in CoreCLR.  Does .NET Framework fail?  If not, let's add a bug on CoreCLR and link it here

> All other architectures are behind at least a little I believe (lack of tall.callvirt support).

OK thanks - are the failures hard (hang/crash) or soft (just doesn't take the tailcall).

If you could add a tracking bug for all known failures on Mono too that would be great.  One of the most important things is to get to a known, stable state.

luhenry:

mono/tests/tailcall/fsharp.fs
```diff
@@ -0,0 +1,233 @@
+(**************************************
```
luhenry: This test needs to be run as part of CI.

luhenry:

luhenry:

mono/tests/tailcall/fsharp.fs
jaykrell: Agreed I'll distill down the tests to IL. I assume we still can't use F#, not even under autoconf. Even though acceptance tests seem to use. I can add that to this PR. All in one or split it up? Some of its passing-ness is architecture specific at least temporarily.

Don: I'll have to go back and test CoreCLR; the two failures I believe are signature mismatch, I think CoreCLR might fail others too. I don't want the bugs to last long so opening bugs might be overkill but ok. The other architectures will use stack in their tailcalls so stack overflow on large input, and maybe nested stack overflow attempting to report that (whether or not stack overflow works well is a separate point) We are pretty close to known stable, just that the AOT hang and not all architectures in sync in one specific way.

mono/mini/method-to-ir.c
```diff
@@ -9107,21 +9107,8 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
-					 * OP_TAILCALL has no return value, so skip the CEE_RET if it is
-					 * only reachable from this call.
-					 */
-					GET_BBLOCK (cfg, tblock, ip + 5);
-					if (tblock == cfg->cbb || tblock->in_count == 0)
-						skip_ret = TRUE;
-					push_res = FALSE;
 
-					goto call_end;
+					goto tailcall_remove_ret;
```
kumpera: This is jumping to the taken branch of an `if` statement. How feasible is to move all the code for CEE_CALL / CEE_CALLVIRT to a separate file and properly factor into a saner set of functions?

mono/mini/method-to-ir.c
```diff
@@ -9070,10 +9079,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 
 				//printf ("HIT: %s -> %s\n", mono_method_full_name (cfg->method, TRUE), mono_method_full_name (cmethod, TRUE));
 
-				if (cfg->backend->have_op_tail_call) {
-					/* Handle tail calls similarly to normal calls */
-					tail_call = TRUE;
-				} else {
+				if (!cfg->backend->have_op_tail_call) {
```
kumpera: Another thing to look into is whether we can move all backends to OP_TAILCALL to simplify this code.

mono/tests/tailcall/fsharp.fs
luhenry: The issue with fsharp is we don't build the compiler as part of Mono, so for any platform that do not depend on a system mono to be installed, we won't be able to run f# tests. Compiling and committing the IL would indeed be the best option here, since we can't rely on the C# compiler to generate the `tail.` prefixes.

mono/mini/method-to-ir.c
jaykrell: Agreed but ppc, sparc, s390x are not high priority.

mono/tests/tailcall/fsharp.fs
jaykrell: The other way around, we can rely on C# to never output tail. It never ever does.
I'll work on getting more IL tests here. This might also fix existing IL tests.

mono/mini/method-to-ir.c
jaykrell: Yes when I have sparc, ppc, s390x setups and the priority reaches top.

jaykrell:
@monojenkins build failed

mono/tests/tailcall/fsharp.fs
luhenry: We want the IL test integrated to our testing infrastructure in place of this F# test before merging.

jaykrell:
There is fsharpc — standalone that will “just work” but is large. That, or shrink it down? Also I’d feel safer in the short term to only test amd64. Ok? Also keep the tests as one big one (“just works”), or split it up?

mono/mini/method-to-ir.c
jaykrell: Almost anything is possible. I’d rather take this change in the short term & refactor separately/later.  

Placing a label here and jmping to is not a problem & better than duplicating code. 

Goto is not actually evil despite what some people say.

mono/mini/method-to-ir.c
```diff
@@ -9099,21 +9100,8 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
-					 * only reachable from this call.
-					 */
-					GET_BBLOCK (cfg, tblock, ip + 5);
-					if (tblock == cfg->cbb || tblock->in_count == 0)
-						skip_ret = TRUE;
-					push_res = FALSE;
-
-					goto call_end;
+					goto tailcall_remove_ret;
 				}
```
vargaz: The way this is currently done is we have one call_end label, and a bunch of flags are set to control its behavior, you can just add 1 more flag.

jaykrell: An option or insist? I guess I call it tailcall_remove_ret, move the label up, use if on that gboolean.

jaykrell: @vargaz Commit 94b5bb4 is an attempt to do what you say.

jaykrell:
@luhenry ping?

jaykrell:
@vargaz better with 94b5bb4 ?
@kumpera still ok, with 94b5bb4 ?

jaykrell:
@vargaz @kumpera @luhenry ok with latest?

jaykrell:
Need to merge with master..still a PR in flight ahead of this..

mono/tests/tailcall/fsharp.fs
luhenry: My previous review isn't stale: this test is still in fsharp and thus unusable by CI. If https://github.com/mono/mono/pull/7524/files#diff-35d1a6f41e6c8c560c59586e7dbb9385 is the equivalent test in IL, then this `.fs` file should be removed

jaykrell:
The F# file is important because it tries to explain what is the IL, and it is the basis of future IL. And if we ever figure out how to use fsharpc then we have it available.

jaykrell:
This will unbreak some/many tests in
https://github.com/mono/mono/pull/7524

jaykrell:
I moved the F# code into comments in the .il, which is the role the separate .fs file was filling.

jaykrell:
Anyone wanting to test the F# code will have to extract the comment instead of having the convenient separate file.

mono/tests/Makefile.am
```diff
@@ -718,6 +718,9 @@ PLATFORM_DISABLED_TESTS += tailcall-virt.exe \
 	tailcall/fsharp.exe
 endif
 
+# FIXME Temporary pending https://github.com/mono/mono/pull/7658
+PLATFORM_DISABLED_TESTS += tailcall/fsharp.exe
```
luhenry: This should go in KNOWN_FAILING_TESTS

jaykrell: I'm going to remove it as soon as Jenkins catches up..or even let 7658 go in first.

jaykrell:

PR #7530, opened 2018-03-09
[arm64] Add a known TLS offset.

rolfbjarne:

### DIFF
--- a/mono/utils/mach-support-arm64.c
+++ b/mono/utils/mach-support-arm64.c
@@ -32,6 +32,7 @@ static const int known_tls_offsets[] = {
 	0x48, /*Found on iOS 6 */
 	0xA4,
 	0xA8,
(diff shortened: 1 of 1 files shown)


lewurm:
is this needed on `2017-12` and `2018-02` too?

rolfbjarne:
@lewurn: no, it's not necessary to backport this, it just silences a warning message.

BrzVlad:
I think this code is dead. Do we still get tls values from other threads ?

rolfbjarne:
@BrzVlad every iOS app prints this warning without this fix: https://github.com/mono/mono/blob/91b64b4560b78cf05ec99faec693737d7f63d093/mono/utils/mach-support-arm64.c#L193

BrzVlad:
@rolfbjarne Yes. I think it basically prints a warning about dead code not working.

luhenry:

PR #7512, opened 2018-03-08
[bcl] Fix test build with mcs

akoeplinger:
It'd fail with the following errors since e6de24c2d196b064e64aeea6e45350e985a1bb23:

```
MCS     [net_4_x-darwin] net_4_x_System.Data.OracleClient_test.dll
Test/System.Data.OracleClient.jvm/MonoTests.System.Data.Utils/ADONetTesterClass.cs(337,71): error CS0012: The type `System.Xml.Serialization.IXmlSerializable' is defined in an assembly that is not referenced. Consider adding a reference to assembly `System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'

...

MCS     [net_4_x-darwin] net_4_x_System.Json.Microsoft_test.dll
Test/System.Json/JsonValueTest.cs(23,12): error CS0012: The type `System.Dynamic.IDynamicMetaObjectProvider' is defined in an assembly that is not referenced. Consider adding a reference to assembly `System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'

...

MCS     [net_4_x-darwin] DummyAssembly.dll
error CS0518: The predefined type `System.Object' is not defined or imported
error CS0518: The predefined type `System.ValueType' is not defined or imported
```

@marek-safar this sounds like mcs bugs to me as it compiles fine with csc.

PR #7531, opened 2018-03-09
[System.Windows.Forms] Fixes tests build

marek-safar:


akoeplinger:
I've merged this into https://github.com/mono/mono/pull/7512.

Issue #7536, opened 2018-03-09
Missing Facades from net_4_x that are in 4.7.1.


akoeplinger:
We're missing those:

```
System.Net.Http.Rtc.dll
System.Net.Ping.dll
System.Resources.Reader.dll
System.Resources.Writer.dll
System.Runtime.Serialization.Formatters.dll
System.Security.Cryptography.Csp.dll
```

PR #7541, opened 2018-03-09
[Facades] Add missing Facades that exist in .NET 4.7.1

akoeplinger:
Some of them we had on mobile only, others were missing completely.

Fixes https://github.com/mono/mono/issues/7536

### DIFF
--- a/mcs/class/Facades/System.Net.Http.Rtc/AssemblyInfo.cs
+++ b/mcs/class/Facades/System.Net.Http.Rtc/AssemblyInfo.cs
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2013 Xamarin Inc. (http://www.xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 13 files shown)


akoeplinger:
@monojenkins commit apidiff

akoeplinger:
@monojenkins backport 2018-02

marek-safar:
Review:

- System.Net.Http.Rtc.dll -- remove completely I don't think it's worth the pain as it's windows only
- System.Resources.Reader/System.Resources.Writer we need to report CoreFX issue to update the package to include our mobile version TFMs

akoeplinger:
> System.Net.Http.Rtc.dll -- remove completely I don't think it's worth the pain as it's windows only

I'd like to keep System.Net.Http.Rtc.dll since it's a bigger pain to exclude it from all the compares I do with .NET Fx. It's not really a pain since we need to edit some files for  https://github.com/mono/mono/issues/7573 anyway and it's just one more line.

> System.Resources.Reader/System.Resources.Writer we need to report CoreFX issue to update the package to include our mobile version TFMs

https://github.com/dotnet/corefx/issues/27995

marek-safar:
`System.Net.Http.Rtc.dll` I am fine to keep it on desktop even if it's useless as it cannot be referenced but don't want this in any mobile/modern profile

marek-safar:
We should also update `api` Facades

akoeplinger:
@marek-safar what do you mean by that?

marek-safar:
Should we add them also to https://github.com/mono/reference-assemblies/tree/master/v4.7.1/Facades ?

akoeplinger:
@marek-safar they are there, that's why we noticed them missing from mono in the first place :)

marek-safar:
oh (used wrong search pattern). We are good then

Issue #7525, opened 2018-03-09
Profiler exception clause reporting is broken


alexrp:
```bash
bash ~/tests $ cat test.cs
using System;

class Program {
        static void Main () {
                try {
                        throw new Exception ();
                } catch (Exception) when (Filter ()) {
                        Console.WriteLine ("catch");
                } finally {
                        Console.WriteLine ("finally");
                }
        }
        static bool Filter () {
                Console.WriteLine ("filter");
                return true;
        }
}
bash ~/tests $ mono --profile=log:nodefaults,exception test.exe
filter
catch
finally
bash ~/tests $ mprof-report --reports=exception output.mlpd

Exception summary
        Throws: 1
        Executed filter clauses: 1
```

This is a bug on the runtime side. Some of the `exception_clause` events are not being raised for some reason.

Issue #6183, opened 2017-12-06
Graduate Mono.Profiler.Log to stable status


alexrp:
What needs to happen:

- [ ]  Refine and stabilize the API.
- [ ] Remove the library from `MONO_API_ASSEMBLIES_IGNORED`.
- [ ] Include the library in the `mobile` and `xammac_4_5` profiles (possibly others).
- [ ] Start shipping the library in XA/XI/XM.

PR #7543, opened 2018-03-09
[Mono.Profiler.Log] Add a new profiler test suite.

### LINKED ISSUE #6979: Create Mono.Profiler.Log test suite
We want to have a Mono.Profiler.Log test suite to ensure we do not regress the log profiler over time. It should test profiling from the `metadata/profiler.c` API all the way to the `Mono.Profiler.Log` API

alexrp:
This is the beginning of the new profiler test suite based on the `Mono.Profiler.Log` library. This new test suite will exist alongside the old profiler tests in `mono/profiler` until I'm confident that the new tests cover everything well enough. This PR also fixes some bugs caught by the new tests.

Note that two tests are currently disabled due to #7525 and #7560.

Fixes #6979.


alexrp:
Feel free to review, but this is probably not ready to merge yet. I expect some of these tests will fail on ARM and maybe OS X.

kumpera:
THIS IS AMAZING AND DESERVES TO BE PRAISED IN CAPS.

alexrp:

alexrp:
@monojenkins build Linux i386

alexrp:

alexrp:
@monojenkins build Linux i386

alexrp:
Also need to rewrite the monitor event test to be a bit more robust on ARM.

alexrp:
@monojenkins build Linux AAarch64

alexrp:

alexrp:
OK, this is ready for review, and assuming no related build failures, can probably be merged also.

alexrp:
@monojenkins build failed

mcs/class/Mono.Profiler.Log/Makefile
```diff
@@ -2,14 +2,26 @@ thisdir = class/Mono.Profiler.Log
...
-LIB_REFS = System System.Core
 KEYFILE = $(LIBRARY_SNK)
+
+LIB_REFS = System System.Core
 LIB_MCS_FLAGS = /unsafe /publicsign /nowarn:0618
 
-LIBRARY_WARN_AS_ERROR = yes
+XTEST_LIB_REFS = System System.Core
+
+xunit-test-local: log-profiler-test.exe
```
marek-safar: We already have support for this, called REMOTE_EXECUTOR see https://github.com/mono/mono/blob/master/mcs/class/System/Makefile#L31 how to enable it. 

You then derive from RemoteExecutorTestBase and write your asserts as normal code

```
    public class Foo : RemoteExecutorTestBase
    {
        [Fact]
        public static void Test()
        {
            RemoteInvoke(() => { });
        }
```

I think it'd make sense to use this instead of building another one

mcs/class/Mono.Profiler.Log/Mono.Profiler.Log/LogEvents.cs
```diff
@@ -276,7 +276,7 @@ public sealed class HeapRootsEvent : LogEvent {
 
 		public struct HeapRoot {
 
-			public long AddressPointer { get; internal set; }
+			public long SlotPointer { get; internal set; }
```
marek-safar: Isn't this breaking change?

mcs/class/Mono.Profiler.Log/Makefile
alexrp: I don't think that would apply to this use case since I need to pass arguments to the runtime itself?

mcs/class/Mono.Profiler.Log/Mono.Profiler.Log/LogEvents.cs
marek-safar: Then just extend it, it's still better than maintaining two very similar support code

mcs/class/Mono.Profiler.Log/Makefile
marek-safar: Then just extend it, it's still better than maintaining two very similar support code

mcs/class/Mono.Profiler.Log/Mono.Profiler.Log/LogEvents.cs
alexrp: The API of the library is explicitly not stable right now (see #6183) and the old name of this property was pretty bad.

marek-safar:
@alexrp could you resolve the failing test?

alexrp:
Note that this currently depends on #7940 being merged first.

alexrp:
build profiler stress

alexrp:
If tests pass (they should), this is good to merge.

alexrp:
build profiler stress

luhenry:
@monojenkins build OS X i386

PR #7549, opened 2018-03-09
[WIP - DONT MERGE][wasm/sdb] Introduce debugger-engine.c file that will host the common code between sdb and the wasm debugger.

kumpera:
This PR includes the refactoring to unify the breakpoint engine between sdb and the wasm debugger.

It moves/adjusts code from sdb into debugger-engine.c and will then replace tons of code in mini-wasm.c with it.

Making an early PR to draw feedback on the design before I land too much code into it.

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -451,6 +451,8 @@ common_sources = \
 	mini-gc.c		\
 	debugger-agent.h 	\
 	debugger-agent.c	\
+	debugger-engine.h	\
+	debugger-engine.c	\
 	xdebug.c			\
 	mini-llvm.h			\
 	mini-llvm-cpp.h	\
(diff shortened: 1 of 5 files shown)


vargaz:
Can't we extend the current file with the new behavior ?

joncham:
We have a somewhat modified debugger-agent.c to support IL2CPP. Would be good to see what the split is going to look like. We'll have a painful merge, but I think that's unavoidable.

But perhaps between this and work to use MonoClass accessors we can work with less changes.

kumpera:
@vargaz I think keeping the code in debugger-agent.c will make it a lot harder to keep the wasm debugger working.

It means a bunch of functions & globals have a different set of ifdefs to compile them in. The amount of ifdef spaguetti to make sure we don't compile the sdb-only bits would be ugly.

It would make initialization/state maintenance a lot more fragile cuz one changing sdb would not see obvious impacts into wasm.

The subsystems I plan to move for now (breakpoints/single stepping) are reasonably well factored, I won't need a lot of changes beyond moving the files.

@joncham I plan to move logging, breakpoints and single stepping into the shared file. Now having said that, I believe some of the future changes I plan on doing to support compiled code debugging will make il2cpp debugging significantly more straightforward.

kumpera:
This was merged by #8750

PR #7555, opened 2018-03-10
[profiler] Increment ref count of an image's method while we're keeping around pointers to it.

### LINKED ISSUE #6211: Rare shutdown crash in profiler stress tests
Example: https://jenkins.mono-project.com/job/test-mono-mainline-linux-profilerstress-x86-x64/label=high-mem-ubuntu-1404-i386/500/testReport/junit/MonoTests/profiler-stress/couchbase/

It looks like we try to call `mono_method_full_name` on a `MonoMethod` whose signature contains a type that's already been freed. This happens because we keep `MonoMethod` pointers around in the profiler for longer than we should - see #6171. The fix will probably be similar to that issue.

alexrp:
Fixes #6211.

### DIFF
--- a/mono/profiler/log.c
+++ b/mono/profiler/log.c
@@ -852,6 +852,18 @@ emit_method_inner (LogBuffer *logbuffer, void *method)
 	g_assert (logbuffer->cursor <= logbuffer->buf_end && "Why are we writing past the buffer end?");
 }
 
+static void
+inc_method_ref_count (MonoMethod *method)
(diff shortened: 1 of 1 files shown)


alexrp:
build profiler stress

alexrp:
Failures unrelated.

alexrp:
@monojenkins build failed

alexrp:
@monojenkins build failed

PR #7567, opened 2018-03-11
Update gitignore, keep paths in windows format for csproj

EgorBo:
1) Add temp stuff generated by tools/ to .gitignore
2) Jenkins bot updates all csproj on macOS so when I want to re-generate them locally the script overwrites, let' say:
`<OutputPath>./../../../class/lib/net_4_x-$(HostPlatform)/Facades</OutputPath>`
to
`<OutputPath>.\..\..\..\class\lib\net_4_x-$(HostPlatform)\Facades</OutputPath>`
since the second one is "more default" for msbuild and still works on macOS we can use it

### DIFF
--- a/msvc/scripts/genproj.cs
+++ b/msvc/scripts/genproj.cs
@@ -929,9 +929,9 @@ class MsbuildGenerator {
 		// ../class/lib/build/tmp/System.Xml.dll  [No longer possible, we should be removing this from order.xml]
 		//   /class/lib/basic/System.Core.dll
 		// <library_output>mcs.exe</library_output>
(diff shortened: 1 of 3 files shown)


alexrp:
I already have a commit in #7543 that will add the `.libs` directory as a whole to the root `.gitignore`.

alexrp:
By the way, you should push stuff like this to a branch in your fork, not the upstream repository.

EgorBo:
@alexrp oops, I didn't notice I have an access to do it actually at all :) I though I am in my fork

PR #7493, opened 2018-03-08
[System.Windows.Forms] Fix bugs in PrintPreviewControl.

filipnavara:

PR #7571, opened 2018-03-12
[System.Windows.Forms] Fix custom test targets in makefile.

filipnavara:

### DIFF
--- a/mcs/class/System.Windows.Forms/Makefile
+++ b/mcs/class/System.Windows.Forms/Makefile
@@ -131,14 +131,14 @@ $(PREBUILT): %.prebuilt: %
 
 dist-default: $(PREBUILT)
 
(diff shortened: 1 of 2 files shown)


filipnavara:
Broken since https://github.com/mono/mono/commit/e6de24c2d196b064e64aeea6e45350e985a1bb23.

akoeplinger:
Hmm, looks like due to e6de24c the WinForms tests got skipped (since simple-test.exe failed) and in the meantime we had some regression as they hang now...

filipnavara:
I am running the unit tests on OS X w/ X11 backend just fine, but I haven't recompiled the runtime in quite a while.

akoeplinger:
It seems to hang in `***** MonoTests.System.Windows.Forms.PrintPreviewControlTest.StartPage` so I assume it is caused by https://github.com/mono/mono/pull/7493, could you try disabling the test for now?

filipnavara:
It hangs because there are no printers on the machine. I will add check to the test and skip it.

PR #7574, opened 2018-03-12
[Facades] Move types from Facades into class libraries

akoeplinger:
We implemented them in the facades until now but in .NET 4.7.1 they were promoted into the regular class libraries.


akoeplinger:
We need to coordinate this with the other corefx bumps that @EgorBo is doing, just putting it up so I don't forget.

marek-safar:
Please add `PLATFORM_DEBUG_FLAGS =` to Facades which now have no IL code, we are using this to avoid generating pdb file.

I don't think we need to wait for @EgorBo as this case land independently

akoeplinger:
@monojenkins commit apidiff

Issue #7558, opened 2018-03-10
F# build fails on master with "Specified cast is not valid"


akoeplinger:
from https://jenkins.mono-project.com/job/build-package-osx-mono/job/master/686/:

```
15:25:30   mono @/external/bockbuild/builds/fsharp-fsharp-561af8b/packages/FSharp.Compiler.Tools.4.1.27/tools/fsi.exe --exec "@/external/bockbuild/builds/fsharp-fsharp-561af8b/src/scripts/fssrgen.fsx" "@/external/bockbuild/builds/fsharp-fsharp-561af8b/src/fsharp/FSharp.Build/FSBuild.txt" "obj/Proto/net40/FSBuild.fs"  "obj/Proto/net40/FSBuild.resx" 
15:25:30   
15:25:30   
15:25:30 error FS0193 : internal error : Specified cast is not valid. [@/external/bockbuild/builds/fsharp-fsharp-561af8b/src/fsharp/FSharp.Build-proto/FSharp.Build-proto.fsproj]
15:25:30   
15:25:30   Exception by fsi.exe:
15:25:30   System.Exception: Error creating evaluation session: System.InvalidCastException: Specified cast is not valid.
15:25:30     at (wrapper castclass) System.Object.__castclass_with_cache(object,intptr,intptr)
15:25:30     at Microsoft.FSharp.Core.LanguagePrimitives+IntrinsicFunctions.UnboxGeneric[T] (System.Object source) [0x00019] in <5893d081904cf4daa745038381d09358>:0 
15:25:30     at FSIstrings.SR.GetStringFunc[T] (System.String messageID, Microsoft.FSharp.Core.PrintfFormat`4[TPrinter,TState,TResidue,TResult] fmt) [0x00038] in <59d43c662274556da7450383663cd459>:0 
15:25:30     at Microsoft.FSharp.Compiler.Interactive.Shell+FsiCommandLineOptions.fsiUsagePrefix (Microsoft.FSharp.Compiler.CompileOps+TcConfigBuilder tcConfigB) [0x00000] in <59d43c662274556da7450383663cd459>:0 
[...]

PR #7579, opened 2018-03-12
[Tailcall] Dial back tailcalls slightly -- for unboxing

jaykrell:
and for virtual on unsupported targets.
 This should fix https://github.com/mono/mono/issues/7558 but still testing.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -2160,6 +2160,12 @@ check_method_sharing (MonoCompile *cfg, MonoMethod *cmethod, gboolean *out_pass_
 		*out_pass_mrgctx = pass_mrgctx;
 }
 
+inline static gboolean
+mono_need_unbox_trampoline (MonoMethod *method)
+{
+	return method->klass == mono_defaults.object_class || mono_class_is_interface (method->klass);
+}
+
 inline static MonoCallInst *
 mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig, 
 					 MonoInst **args, int calli, int virtual_, int tail, int rgctx, int unbox_trampoline, MonoMethod *target)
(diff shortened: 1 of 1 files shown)


marek-safar:
Please always add tests for regressions

mono/mini/method-to-ir.c
```diff
@@ -2160,6 +2160,12 @@ check_method_sharing (MonoCompile *cfg, MonoMethod *cmethod, gboolean *out_pass_
 		*out_pass_mrgctx = pass_mrgctx;
 }
 
+inline static gboolean
+mono_need_unbox_trampoline (MonoMethod *method)
+{
+	return method->klass == mono_defaults.object_class || mono_class_is_interface (method->klass);
```
jaykrell: This hopefully isn't all needed -- unable to tailcall an interface would be bad.
What is the unbox trampoline? I see it used only in x86.

jaykrell:
Need to test/debug first.

mono/mini/method-to-ir.c
kumpera: The unbox trampoline is present on all arches, it's used when you do an interface invoke into a boxed value type.
A boxed value type is represented in memory all of its fields after an object header.

ValueType methods are compiled with `this` as a byref to the first field.
interface invokes are compiled with `this` as a pointer to the object header.

This means we need an unbox trampoline that essentially does `this + sizeof (MonoObject)`.

jaykrell: This was just bad guessing on my part. It wasn't so much the questionable changing of tail to false in a few places, such as for the unbox trampoline and llvmonly, which is still questionable, but too much turning into tail in the change. I didn't debug it just a little trial/error based on the original change.

jaykrell: Only x86 seems to reference this though.

jaykrell:
"Please always add tests for regressions" ok this will take a while as I don't really understand the problem, just dialing back the change some.

Issue #6940, opened 2018-02-10
DNS lookup of IPv6-only hosts fails


TekkiWuff:
## Steps to Reproduce

1. Ensure you have propper IPv6 connectivity e. g. by opening `ipv6.google.com` in a browser
2. Write a simple oneliner C# program to do a DNS lookup
```C#
var addresse = Dns.GetHostAddresses("ipv6.google.com");
```
3. An exception is thrown

## Current Behavior

The DNS lookup does not return any IPv6 addresses and fails with an exception for IPv6-only hosts.

## Expected Behavior

The DNS lookup returns all IPv6 addresses associated with the hostname.

### On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

```
Distributor ID: Ubuntu
Description:    Ubuntu 17.10
Release:        17.10
Codename:       artful
```

**Version Used**:

```
Mono JIT compiler version 5.8.0.108 (tarball Fri Jan 19 18:15:21 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        LLVM:          supported, not enabled.
        GC:            sgen (concurrent by default)
```
Installed from repository as described here http://www.mono-project.com/download/#download-lin

### Stacktrace

```
Unhandled Exception:
System.Net.Sockets.SocketException (0x80004005): Could not resolve host 'ipv6.google.com'
  at System.Net.Dns.Error_11001 (System.String hostName) [0x00015] in <a63b52a3a7644f60a95b388cb72441c2>:0
  at System.Net.Dns.hostent_to_IPHostEntry (System.String originalHostName, System.String h_name, System.String[] h_aliases, System.String[] h_addrlist) [0x00068] in <a63b52a3a7644f60a95b388cb72441c2>:0
  at System.Net.Dns.GetHostByName (System.String hostName) [0x00027] in <a63b52a3a7644f60a95b388cb72441c2>:0
  at System.Net.Dns.GetHostEntry (System.String hostNameOrAddress) [0x00052] in <a63b52a3a7644f60a95b388cb72441c2>:0
  at System.Net.Dns.GetHostAddresses (System.String hostNameOrAddress) [0x00056] in <a63b52a3a7644f60a95b388cb72441c2>:0
[...]

PR #7584, opened 2018-03-12
[System.Net] Replace use of obsolete `SupportsIPv6` property.

alexischr:
The obsolete property was always set to false. Fixes https://github.com/mono/mono/issues/6940

### DIFF
--- a/mcs/class/System/System.Net/Dns.cs
+++ b/mcs/class/System/System.Net/Dns.cs
@@ -320,11 +320,9 @@ namespace System.Net {
 				try {
 					IPAddress newAddress = IPAddress.Parse(h_addrlist[i]);
 
-#pragma warning disable 618
-					if( (Socket.SupportsIPv6 && newAddress.AddressFamily == AddressFamily.InterNetworkV6) ||
-					    (Socket.SupportsIPv4 && newAddress.AddressFamily == AddressFamily.InterNetwork) )
+					if( (Socket.OSSupportsIPv6 && newAddress.AddressFamily == AddressFamily.InterNetworkV6) ||
+					    (Socket.OSSupportsIPv4 && newAddress.AddressFamily == AddressFamily.InterNetwork) )
 						addrlist.Add(newAddress);
-#pragma warning restore 618
 				} catch (ArgumentNullException) {
 					/* Ignore this, as the
 					 * internal call might have
--- a/mcs/class/System/Test/System.Net/DnsTest.cs
+++ b/mcs/class/System/Test/System.Net/DnsTest.cs
@@ -210,6 +210,13 @@ namespace MonoTests.System.Net
 		}
 
 		[Test]
+		[Category("NotOnWindows")]
+		public void GetHostAddresses_IPv6 ()
+		{
+			var address = Dns.GetHostAddresses("ipv6.google.com");
+		}
+
+		[Test]
 		public void GetHostName ()
 		{
 			string hostName = Dns.GetHostName ();
--- a/mcs/class/System/Test/System.Net/IPAddressTest.cs
+++ b/mcs/class/System/Test/System.Net/IPAddressTest.cs
@@ -241,9 +241,6 @@ public class IPAddressTest
 	[Test]
 	public void IsLoopbackV6 ()
 	{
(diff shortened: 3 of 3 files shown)


akoeplinger:
The codepath quoted on the issue which always sets the obsolete property to false isn't used on Mono so I don't understand why this happens.

alexischr:
@akoeplinger we define `FEATURE_PAL`: https://github.com/mono/mono/blob/master/mcs/class/System/Makefile#L62

marek-safar:
I think the fix should be to move the code we have to location which is compiled 
https://github.com/mono/mono/blob/4e3389667f6e29480cf978eb524d0c4308fe6cb5/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs#L6065

to

https://github.com/mono/mono/blob/4e3389667f6e29480cf978eb524d0c4308fe6cb5/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs#L6129

akoeplinger:
@alexischr you missed this little nugget at the top of the file 😄 :
https://github.com/mono/mono/blob/4e3389667f6e29480cf978eb524d0c4308fe6cb5/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs#L7-L9

alexischr:
@akoeplinger  Sorry, I saw that today 🤦‍♂️ 

That got me confused and had to verify that the change in `Dns.cs` fixes the behavior (it does, and the other change is not needed). I have to figure how `OSSupportsIPv6` can possibly be true when `SupportsIPv6` apparently wasn't.

alexischr:
oh: https://github.com/mono/mono/blob/4e3389667f6e29480cf978eb524d0c4308fe6cb5/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs#L6117

akoeplinger:
Yeah but that should just return true if the config section isn't there: https://github.com/mono/mono/blob/4e3389667f6e29480cf978eb524d0c4308fe6cb5/mcs/class/System/ReferenceSources/SettingsSectionInternal.cs#L31-L43

... and looking at our machine.config file we're not setting it.

alexischr:
@akoeplinger "system.net/settings" section is declared here: https://github.com/mono/mono/blob/29c2ba9387fe094c672adedd27c56e46bfa3669c/data/net_4_5/machine.config#L87

And the "ipv6/enabled" defaults to false: https://github.com/mono/mono/blob/29c2ba9387fe094c672adedd27c56e46bfa3669c/mcs/class/System/System.Net.Configuration/Ipv6Element.cs#L65

mcs/class/referencesource/System/net/System/Net/Internal.cs
```diff
@@ -318,7 +318,7 @@ private static IPHostEntry GetLocalHost()
             //               old IPv4 gethostbyname(null). Instead we need
             //               to do a more complete lookup.
             //
-            if (Socket.SupportsIPv6)
+            if (Socket.OSSupportsIPv6)
```
akoeplinger: please revert this change, we're not using this code.

mcs/class/System/System.Net/Dns.cs
```diff
@@ -320,11 +320,9 @@ private static IPHostEntry hostent_to_IPHostEntry(string originalHostName, strin
 				try {
 					IPAddress newAddress = IPAddress.Parse(h_addrlist[i]);
 
-#pragma warning disable 618
-					if( (Socket.SupportsIPv6 && newAddress.AddressFamily == AddressFamily.InterNetworkV6) ||
+					if( (Socket.OSSupportsIPv6 && newAddress.AddressFamily == AddressFamily.InterNetworkV6) ||
 					    (Socket.SupportsIPv4 && newAddress.AddressFamily == AddressFamily.InterNetwork) )
```
akoeplinger: please use `Socket.OSSupportsIPv4` here so we're consistent.

marek-safar:
@monojenkins backport 2017-12

marek-safar:
@monojenkins backport 2018-02

marek-safar:
@alexischr please backport manually

PR #7590, opened 2018-03-13
[Facades] Make System.Net.Http.Rtc.dll a net_4_x only facade

akoeplinger:
It doesn't make sense on mobile but we'd like to keep it in net_4_x to match Desktop .NET


akoeplinger:
@monojenkins commit apidiff

akoeplinger:
@monojenkins backport 2018-02

PR #7591, opened 2018-03-13
Debugger: send VM_DISPOSE on connection error

brianradunity:
If a bad header is received from the client or the connection is broken,
send an internal VM_DISPOSE command to shut down the debugger thread
properly.  This prevents runtime hangs that occur during profile
events after the debugger shuts down.


luhenry:
@vargaz @alexanderkyte  can you please review?

akoeplinger:
@monojenkins whitelist

vargaz:
This looks a bit weird by its ok.

PR #7599, opened 2018-03-13
[threads] Delay abort via handler block guard when in abort protected block.

### LINKED ISSUE #7588: Class should be viable even if cctor is aborted in nested finally clause
Consider the following code (https://github.com/lambdageek/mono/commit/144c3df2130e4d39d5b765ba013c33813a5f8da6):
```csharp
	public static bool got_to_the_end_of_outer_finally = false;
	public static bool got_to_the_end_of_inner_finally = false;
	class StaticConstructor6 {
		static StaticConstructor6 ()
		{
			try {
				Setup6 ();
			} finally {
				Driver.got_to_the_end_of_outer_finally = true;
			}
		}

		[MethodImplAttribute (MethodImplOptions.NoInlining)]
		public static void Setup6 () {
			try {
			} finally {
				Driver.sema1.Release ();
				Thread.Sleep (1000); /* hopefully we get woken up here */
				Driver.got_to_the_end_of_inner_finally = true;
			}
		}
	}
```

If a thread running the cctor `StaticConstructor6` is aborted by a call to `thread.Abort()` when it is in the `finally` block in `Setup6`, we should get to the end of both finally clauses and the class `StaticConstructor6` should be viable.

But that doesn't happen.  Instead the thread is self-aborted (which bypasses the abort protection from `mono_runtime_class_init_full` and `StaticConstruct6` is marked as not viable.

The reason is because the main thread ends up calling `mono_install_handler_block_guard` which sets a flag on the child thread at the end up the inner finally block, which causes the child thread to self-abort:
https://github.com/mono/mono/blob/8c3d7c81af7a3b37d21e8d17c9302015258570ce/mono/mini/method-to-ir.c#L11543-L11546

This fails in both with coop and async suspend.

lambdageek:
`mono_install_handler_block_guard` is called when we want to signal an async
abort to another thread.  In case the other thread is executing native code
while in a `finally` block, the handler block guard sets a byte in the finally
block that used to trigger a self abort (`mono_thread_self_abort`).

This PR renames `mono_thread_self_abort` to
`ves_icall_thread_finish_async_abort`.  It also changes the code from
immediately executing a synchronous abort, to checking whether we are in an
abort protected block (most commonly - whether we were called from a cctor) and
in that case merely toggling the "async interrupt"
MonoInternalThread:thread_state bit.  If we're not in an abort-protected block,
trigger a self interrupt (which will turn into an abort exception in the
icall wrapper of `ves_icall_thread_finish_async_abort`).

Fixes #7588

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -384,7 +384,8 @@ static gboolean
 mono_thread_set_self_interruption_respect_abort_prot (void)
 {
 	MonoInternalThread *thread = mono_thread_internal_current ();
-	/* N.B. Sets the ASYNC_REQUESTED_BIT even though for this thread. */
+	/* N.B. Sets the ASYNC_REQUESTED_BIT for current this thread,
+	 * which is unusual. */
 	return mono_thread_set_interruption_requested_flags (thread, FALSE);
 }
 
@@ -4117,13 +4118,38 @@ mono_threads_abort_appdomain_threads (MonoDomain *domain, int timeout)
 void
 ves_icall_thread_finish_async_abort (void)
 {
-	/* If we're in an abort protected block, just set the async requested bit and return.
-	 * Otherwise self abort now.
+	/* We were called from the handler block and are about to
+	 * leave it.  (If we end up postponing the abort because we're
+	 * in an abort protected block, the unwinder won't run and
+	 * won't clear the handler block itself which will confuse the
+	 * unwinder if we're in a try {} catch {} and we throw again.
+	 * ie, this:
+	 * static Constructor () {
+	 *   try {
+	 *     try {
+	 *     } finally {
+	 *       icall (); // Thread.Abort landed here,
+	 *                 // and caused the handler block to be installed
+	 *       if (exvar)
+	 *         ves_icall_thread_finish_async_abort (); // we're here
+	 *     }
+	 *     throw E ();
+	 *   } catch (E) {
+	 *     // unwinder will get confused here and synthesize a self abort
+	 *   }
+	 * }
+	 *
+	 * More interestingly, this doesn't only happen with icalls - a JIT
+	 * trampoline is native code that will cause a handler to be installed.
+	 * So the above situation can happen with any code in a "finally"
+	 * clause.
 	 */
-	if (!mono_thread_set_self_interruption_respect_abort_prot ())
-		return;
-	else
-		mono_thread_info_self_interrupt ();
+	mono_get_eh_callbacks ()->mono_uninstall_current_handler_block_guard ();
+	/* Just set the async interruption requested bit.  Rely on the icall
+	 * wrapper of this icall to process the thread interruption, respecting
+	 * any abort protection blocks in our call stack.
+	 */
+	mono_thread_set_self_interruption_respect_abort_prot ();
 }
 
 /*
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -193,6 +193,9 @@ static void self_abort_internal (MonoError *error);
 static void async_suspend_internal (MonoInternalThread *thread, gboolean interrupt);
 static void self_suspend_internal (void);
 
+static gboolean
+mono_thread_set_interruption_requested_flags (MonoInternalThread *thread, gboolean sync);
+
 static MonoException* mono_thread_execute_interruption (void);
 static void ref_stack_destroy (gpointer rs);
 
@@ -366,6 +369,29 @@ mono_thread_set_interruption_requested (MonoInternalThread *thread)
 {
 	//always force when the current thread is doing it to itself.
 	gboolean sync = thread == mono_thread_internal_current ();
+	/* Normally synchronous interruptions can bypass abort protection. */
+	return mono_thread_set_interruption_requested_flags (thread, sync);
+}
+
+/* Returns TRUE if there was a state change and the interruption can be
+ * processed.  This variant defers a self abort when inside an abort protected
+ * block.  Normally this should only be done when a thread has received an
+ * outside indication that it should abort.  (For example when the JIT sets a
+ * flag in an finally block.)
+ */
+
+static gboolean
+mono_thread_set_self_interruption_respect_abort_prot (void)
+{
+	MonoInternalThread *thread = mono_thread_internal_current ();
+	/* N.B. Sets the ASYNC_REQUESTED_BIT even though for this thread. */
+	return mono_thread_set_interruption_requested_flags (thread, FALSE);
+}
+
+/* Returns TRUE if there was a state change and the interruption can be processed. */
+static gboolean
+mono_thread_set_interruption_requested_flags (MonoInternalThread *thread, gboolean sync)
+{
 	gsize old_state, new_state;
 	do {
 		old_state = thread->thread_state;
@@ -4081,12 +4107,23 @@ mono_threads_abort_appdomain_threads (MonoDomain *domain, int timeout)
 	return TRUE;
 }
 
+/* This is a JIT icall.  This icall is called from a finally block when
+ * mono_install_handler_block_guard called by another thread has flipped the
+ * finally block's exvar (see mono_find_exvar_for_offset).  In that case, if
+ * the finally is in an abort protected block, we must defer the abort
+ * exception until we leave the abort protected block.  Otherwise we proceed
+ * with a synchronous self-abort.
+ */
 void
-mono_thread_self_abort (void)
+ves_icall_thread_finish_async_abort (void)
 {
-	ERROR_DECL (error);
-	self_abort_internal (error);
-	mono_error_set_pending_exception (error);
+	/* If we're in an abort protected block, just set the async requested bit and return.
+	 * Otherwise self abort now.
+	 */
+	if (!mono_thread_set_self_interruption_respect_abort_prot ())
+		return;
+	else
+		mono_thread_info_self_interrupt ();
 }
 
 /*
--- a/mono/mini/mini-exceptions.c
+++ b/mono/mini/mini-exceptions.c
@@ -117,6 +117,7 @@ static void mono_raise_exception_with_ctx (MonoException *exc, MonoContext *ctx)
 static void mono_runtime_walk_stack_with_ctx (MonoJitStackWalk func, MonoContext *start_ctx, MonoUnwindOptions unwind_options, void *user_data);
 static gboolean mono_current_thread_has_handle_block_guard (void);
 static gboolean mono_install_handler_block_guard (MonoThreadUnwindState *ctx);
+static void mono_uninstall_current_handler_block_guard (void);
 
 static gboolean
 first_managed (MonoStackFrameInfo *frame, MonoContext *ctx, gpointer addr)
@@ -232,6 +233,7 @@ mono_exceptions_init (void)
 	cbs.mono_raise_exception_with_ctx = mono_raise_exception_with_ctx;
 	cbs.mono_exception_walk_trace = mono_exception_walk_trace;
 	cbs.mono_install_handler_block_guard = mono_install_handler_block_guard;
+	cbs.mono_uninstall_current_handler_block_guard = mono_uninstall_current_handler_block_guard;
 	cbs.mono_current_thread_has_handle_block_guard = mono_current_thread_has_handle_block_guard;
 	cbs.mono_clear_abort_threshold = mini_clear_abort_threshold;
 	cbs.mono_above_abort_threshold = mini_above_abort_threshold;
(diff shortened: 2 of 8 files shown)


mono/metadata/threads.c
```diff
@@ -4059,12 +4085,23 @@ mono_threads_abort_appdomain_threads (MonoDomain *domain, int timeout)
...
+ves_icall_thread_finish_async_abort (void)
 {
-	ERROR_DECL (error);
-	self_abort_internal (error);
-	mono_error_set_pending_exception (error);
+	/* If we're in an abort protected block, just set the async requested bit and return.
+	 * Otherwise self abort now.
+	 */
+	if (!mono_thread_set_self_interruption_respect_abort_prot ())
+		return;
```
kumpera: Can't his be
```
if (mono_thread_set_self_interruption_respect_abort_prot ())
   mono_thread_info_self_interrupt ();
```
?

mono/metadata/threads.c
```diff
@@ -365,6 +368,29 @@ mono_thread_set_interruption_requested (MonoInternalThread *thread)
 {
 	//always force when the current thread is doing it to itself.
 	gboolean sync = thread == mono_thread_internal_current ();
+	/* Normally synchronous interruptions can bypass abort protection. */
+	return mono_thread_set_interruption_requested_flags (thread, sync);
```
kumpera: What about renaming `mono_thread_set_interruption_requested` to follow `mono_thread_set_self_interruption_respect_abort_prot`? 

This would allow us to get rid of the weird `_flags` variant.

lambdageek:
Locally, with coop and under high load, I get a crash in `abort-cctor.exe - Test5()` at `object.c:548`

```
#0  0xb7fdcd40 in __kernel_vsyscall ()
#1  0xb7dde367 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#2  0xb7ddfa23 in __GI_abort () at abort.c:89
#3  0x08318bfa in mono_log_write_logfile (log_domain=0x0, level=G_LOG_LEVEL_ERROR, hdr=0, 
    message=0xb54315f0 "* Assertion at object.c:548, condition `is_ok (error)' not met, function:mono_runtime_class_init_full, (null) assembly:/home/aleksey/mono/mono/mcs/class/lib/net_4_x-linux/mscorlib.dll type:ThreadAbort"...) at mono-log-common.c:135
#4  0x083130b7 in structured_log_adapter (log_domain=0x0, log_level=G_LOG_LEVEL_ERROR, 
    message=0xb54315f0 "* Assertion at object.c:548, condition `is_ok (error)' not met, function:mono_runtime_class_init_full, (null) assembly:/home/aleksey/mono/mono/mcs/class/lib/net_4_x-linux/mscorlib.dll type:ThreadAbort"..., user_data=0x0) at mono-logger.c:458
#5  0x0832e4bd in monoeg_g_logv (log_domain=0x0, log_level=G_LOG_LEVEL_ERROR, 
    format=0x8338504 "* Assertion at %s:%d, condition `%s' not met, function:%s, %s\n", 
    args=0xb5151f34 "A\272\067\b$\002") at goutput.c:115
#6  0x0832e531 in monoeg_assertion_message (
    format=0x8338504 "* Assertion at %s:%d, condition `%s' not met, function:%s, %s\n") at goutput.c:135
#7  0x08245845 in mono_runtime_class_init_full (vtable=0x85d5dd8, error=0xb5151fcc) at object.c:548
#8  0x080c7886 in mono_generic_class_init (vtable=0x85d5dd8) at jit-icalls.c:1466
#9  0xb7a0db41 in ?? ()
#10 0xb7a14588 in ?? ()
#11 0xb5c74b0f in System_Threading_ThreadHelper_ThreadStart_Context_object (state=0xb7c16040)
    at /home/aleksey/mono/mono/mcs/class/referencesource/mscorlib/system/threading/thread.cs:68
#12 0x00000000 in ?? ()
(gdb) p mono_pmip ((void*) 0xb7a0db41)
$5 = 0xb54316d8 " (wrapper managed-to-native) object:__icall_wrapper_mono_generic_class_init (intptr) {0x85be750} + 0x49 (0xb7a0daf8 0xb7a0db74) [0x8512bc0 - abort-cctor.exe]"
(gdb) p mono_pmip ((void*) 0xb7a14588)
$6 = 0xb5431290 " Driver/<>c__DisplayClass18_0:<Test5>b__0 () {0x85dc460} + 0x38 (0xb7a14550 0xb7a145f4) [0x8512bc0 - abort-cctor.exe]"
```

So it tried to call `mono_get_exception_type_initialization_checked` but when it invoked `TIE`'s constructor, we got a TAE.  So something is wrong.

lambdageek:
I'm confused about why Test5 is failing for me.

It looks like I'm exiting from `object.c:519` (invoking cctor of `StaticConstructor5`)
```c
		mono_threads_begin_abort_protected_block ();
		mono_runtime_try_invoke (method, NULL, NULL, (MonoObject**) &exc, error);
		mono_threads_end_abort_protected_block ();
```
with `exc` set to a TAE. So a synchronous abort happens somehow even with my changes.

mono/metadata/threads.c
BrzVlad: @lambdageek why do we need to call mono_thread_info_self_interrupt here ?

lambdageek: Good question. I'm not sure.

I'm copying from `self_abort_internal`:
1.  I don't need to do `mono_thread_request_interruption (TRUE)` anymore - because I already called mono_thread_set_self_interruption_respect_abort_prot` and it returned TRUE,
2. so... (looking at the body of `mono_thread_request_interruption`) the next thing I need to do is `mono_thread_info_self_interrupt()`.

BrzVlad: @lambdageek the comment in `mono_thread_request_interruption` says that the method can be called from signal handler. Maybe it needs to break some waits if it is called while we are in unmanaged or something like that. In this case we know we call it directly from managed code after returning from a finally clause, so I don't see any waits that would need to be broken.

lambdageek:
@kumpera @BrzVlad I added a new callback `MonoRuntimeExceptionHandlingCallbacks:mono_uninstall_current_handler_block_guard`

its job is to clear `MonoJitTlsData:handler_block`.

The issue is with code like this:

```csharp
icall_with_wrapper () {
  //try {
       icall ();  // Thread.Abort lands here, so we install a handler block in the "finally" below
  //} finally {
       // if (exvar)
       //     ves_icall_thread_finish_async_abort () 
   //}
}
static Constructor () {
  try {
    icall_with_wrapper ();
    throw E(); // this starts the unwinder
  } catch (E) {
     // at the end of this block if handler_block is still non-NULL we resume the abort!
  }
}
```

This is basically the situation in `abort-cctor.exe` in `Test5` - the `try { icall () } finally { }` is the icall wrapper for `ManualResetEvent.Set`

lambdageek:
There's yet more subtlety here.

I'm was a little bit worried that just blindingly setting `handler_block` to NULL is wrong if the victim thread is aborted multiple times.  What if the `handler_block` was set multiple times before `ves_icall_thread_finish_async_abort` got a chance to run?

But the answer is that can't happen.  If `MonoInternalThread:thread_state` is `ThreadStart_AbortRequested`, subsequent `Thread.Abort` calls will be ignored.

I think that means once a thread is in `ThreadStart_AbortRequested` there can be at most one `handler_block` install call.

lambdageek:
@monojenkins build failed

mono/metadata/threads.c
```diff
@@ -4095,13 +4096,33 @@ mono_threads_abort_appdomain_threads (MonoDomain *domain, int timeout)
...
+	 * leave it.  (If we end up postponing the abort because we're
+	 * in an abort protected block, the unwinder won't run and
+	 * won't clear the handler block itself which will confuse the
+	 * unwinder if we're in a try {} catch {} and we throw again.
+	 * ie, this:
+	 * static Constructor () {
+	 *   try {
+	 *     try {
+	 *       icall (); // Thread.Abort landed here,
+	 *                 // and caused the handler block to be installed
```
BrzVlad: @lambdageek  I think you meant for this icall () to be inside the finally clause ?

lambdageek: I meant to show is that `icall` has a wrapper that has a `finally` clause. When we get the abort, we're in native and the handler block is installed on the `finally` of the wrapper.

BrzVlad: A handler block is installed only if we are inside a finally clause, not if we are guarded by a finally clause. Or am I still missing something from your explanation ?

lambdageek: You're right.

This is what the victim thread looks like (abort-cctor.exe Test1) when the handler is installed:
```
#0  0xb7fdcd40 in __kernel_vsyscall ()
#1  0xb7f6a036 in sem_post@@GLIBC_2.1 ()
    at ../nptl/sysdeps/unix/sysv/linux/i386/i686/../i486/sem_post.S:62
#2  0x08322064 in mono_os_sem_post (sem=0x850b25c <suspend_semaphore>)
    at ../../mono/utils/mono-os-semaphore.h:282
#3  mono_threads_notify_initiator_of_suspend (info=0xb5400468) at mono-threads.c:106
#4  0x08326469 in mono_threads_state_poll_with_info (info=0xb5400468) at mono-threads-coop.c:147
#5  0x08175905 in mono_threads_safepoint () at ../../mono/utils/mono-threads-coop.h:43
#6  mini_method_compile (method=0x854ad40, opts=378629119, domain=0x8512bc0, flags=JIT_FLAG_RUN_CCTORS, 
    parts=0, aot_method_index=-1) at mini.c:3561
#7  0x08176175 in mono_jit_compile_method_inner (method=0x854ad40, target_domain=0x8512bc0, 
    opt=378629119, error=0xb56feb3c) at mini.c:4167
#8  0x08074751 in mono_jit_compile_method_with_opt (method=0x854ad40, opt=0, jit_only=0, error=0xb56feb3c)
    at mini-runtime.c:2200
#9  0x0810612b in common_call_trampoline (regs=0x1, regs@entry=0xb56feba0, 
    code=0x1 <error: Cannot access memory at address 0x1>, 
    code@entry=0xb5c807c0 <System_Threading_NativeEventCalls_SetEvent_Microsoft_Win32_SafeHandles_SafeWaitHandle+144> "\017\266E\354\205\300t\r\213E\b\211\004$9", m=0x854ad40, vt=0x0, vtable_slot=0x0, 
    error=0xb56feb3c) at mini-trampolines.c:739
#10 0x08106a7e in mono_magic_trampoline (regs=0xb56feba0, 
    code=0xb5c807c0 <System_Threading_NativeEventCalls_SetEvent_Microsoft_Win32_SafeHandles_SafeWaitHandle+144> "\017\266E\354\205\300t\r\213E\b\211\004$9", arg=0x854ad40, tramp=0x0) at mini-trampolines.c:872
#11 0xb7c15188 in ?? ()
#12 0x0854ad40 in ?? ()
#13 0xb5c6bc25 in System_Threading_EventWaitHandle_Set (this=...)
    at /home/aleksey/mono/mono/mcs/class/referencesource/mscorlib/system/threading/eventwaithandle.cs:319
#14 0xb5400f58 in ?? ()
#15 0xb7a10b20 in ?? ()
#16 0xb7a10930 in ?? ()
#17 0xb7a0c27d in ?? ()
#18 0x08075683 in mono_jit_runtime_invoke (method=0x85b82d8, obj=0x0, params=0x0, exc=0xb7401578, 
    error=0xb56fec78) at mini-runtime.c:2878
```

Frame 6 `method` arg is `"__icall_wrapper_mono_threads_state_poll"`, Frame 11 is `"<0xb7c15188 - generic_trampoline_jit trampoline>"` and Frame 12 stubbornly refuses to `mono_pmip`.  So I'm actually not sure where I am.

lambdageek:
@BrzVlad Aside from my messed up example (I will fix it), do you the handler block changes make sense - is it sensible to uninstall the handler block once we are acting on `exvar` being non-zero?

BrzVlad:
@lambdageek Overall LGTM

lambdageek:
@monojenkins merge

lambdageek:
@monojenkins build failed

luhenry:

mono/mini/mini-exceptions.c
```diff
@@ -2955,10 +2957,10 @@ install_handler_block_guard (MonoJitInfo *ji, MonoContext *ctx)
 			break;
 	}
 
-	/*no matching finally */
-	if (i == ji->num_clauses)
-		return;
+	/*no matching finally - can't happen, we parallel the logic in find_last_handler_block. */
+	g_assert (i < ji->num_clauses);
 
+	g_message (" installed handler block guard at %p\n", ip);
```
alexischr: escaped debug line?

PR #7600, opened 2018-03-13
[Tailcall] Cleanup, handle varargs, cleanup varargs handling, be sure not to use unsupported tailcall_membase, pass r11 through, align patches.

jaykrell:


jaykrell:
Passing through r11 was an attempt to fix the F# regression, but it doesn't fix it.
It still seems like a good idea.
There is larger "problem", significantly whittled away here, that call has a lot of handling that tailcall lacks. Possibly I should refactor more significantly though, and reuse emit_call_body?

mono/mini/method-to-ir.c
```diff
@@ -9041,7 +9041,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			// 2. a Non-generic static methods of reference types and b. non-generic methods
 			//    of value types need to be passed a pointer to the caller’s class’s VTable in the MONO_ARCH_RGCTX_REG register.
 			// 3. Generic methods need to be passed a pointer to the MRGCTX in the MONO_ARCH_RGCTX_REG register
-			if (inst_tailcall && cfg->verbose_level >= 2)
+			if (inst_tailcall && cfg->verbose_level)
 				g_print ("tail.%s %s -> %s supported_tail_call:%d gshared:%d vtable_arg:%d\n",
```
vargaz: verbose level==1 is for very high level stuff, >= 2 should be ok.

vargaz:
The is_supported_tail_call () changes look ok, could you put it into another PR?

jaykrell:
Yes I can definitely split it up. I like extra verbose— it is, you know, one line per rare instruction (or maybe more due to inline checks?).

jaykrell:
I’m not sure more refactoring is needed, in that tailcall is only currently for patchinfo method, not abs, but still to debug. Ok on the r11 & varargs & abi work? Certainly can split up.

vargaz:
Currently, verbose_level==1 only prints out about 2 lines per method, everything else is at a higher level.

mono/mini/mini-amd64.c
```diff
@@ -181,7 +181,8 @@ amd64_use_imm32 (gint64 val)
 }
 
 static void
-amd64_patch (unsigned char* code, gpointer target)
+amd64_patch_aligned (unsigned char* code, gpointer target, gboolean aligned)
+// "aligned" just generated and never run, vs. in the midst of running.
 {
```
vargaz: Why is this new argument needed ?

jaykrell: We frequently patch during initial codegen when alignment doesn’t matter, code not yet running. And also patch later when running and alignment matters.

mono/mini/method-to-ir.c
jaykrell: Ok. I “need” another flag perhaps.

mono/mini/mini-amd64.c
jaykrell: Granted: which instructions patched when? I didn’t check. I can try always asserting & no new param.

mono/mini/mini-amd64.c
```diff
@@ -4672,28 +4738,8 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
-				for (i = 0; i < AMD64_XMM_NREG; ++i)
-					if (call->used_fregs & (1 << i))
-						nregs ++;
-				if (!nregs)
-					amd64_alu_reg_reg (code, X86_XOR, AMD64_RAX, AMD64_RAX);
-				else
-					amd64_mov_reg_imm (code, AMD64_RAX, nregs);
-			}
 
+			code = amd64_handle_varargs_call (cfg, code, call, TRUE);
```
jaykrell: Last param should be false here but probably does not matter.

jaykrell:
Moved method-to-ir.c to https://github.com/mono/mono/pull/7608

jaykrell:
On the alignment thing..well..my asserts are a start but really miss stuff.
When I hardcode aligned to true, it fails, but not clearly in an interesting way.

I think as well, the running-ness of code is subject to race conditions. Not just in the obvious way that code is usually not running, but in the sense that the JIT can be running on multiple threads, and therefore JIT code in varying order, therefore might have resolved targets when it JITs, or not until it runs.

So, I think there is "something" to do here, but my fix is perhaps not it, incomplete, wrong, etc.
The alignment I added to tailcall, btw, I didn't find via assertion failure or disassembly/print of actual values, but the code sure looks wrong, and the fix is based on call -- part of what I was saying where tailcall kind of a pale impersonation of call. It must be different of course, and I suppose it will always be a small subset, but it also seemed/seems to be missing pieces.

I guess remove the assert for now?


   frame #2: 0x0000000101f19a80 mono`sigabrt_signal_handler(_dummy=6, _info=0x00007fff5df1df18, context=0x00007fff5df1df80) at mini-posix.c:209
    frame #3: 0x00007fffdf9b7b3a libsystem_platform.dylib`_sigtramp + 26
    frame #4: 0x00007fffdf8d6d43 libsystem_kernel.dylib`__pthread_kill + 11
    frame #5: 0x00007fffdf9c4457 libsystem_pthread.dylib`pthread_kill + 90
    frame #6: 0x00007fffdf83c420 libsystem_c.dylib`abort + 129
    frame #7: 0x00000001021b8044 mono`mono_log_write_logfile(log_domain=0x0000000000000000, level=G_LOG_LEVEL_ERROR, hdr=0, message="* Assertion at mini-amd64.c:214, condition `!aligned || ((gsize)(code + 1) % 4) == 0' not met\n") at mono-log-common.c:135
    frame #8: 0x00000001021adfac mono`structured_log_adapter(log_domain=0x0000000000000000, log_level=G_LOG_LEVEL_ERROR, message="* Assertion at mini-amd64.c:214, condition `!aligned || ((gsize)(code + 1) % 4) == 0' not met\n", user_data=0x0000000000000000) at mono-logger.c:458
    frame #9: 0x00000001021dd2cd mono`monoeg_g_logv(log_domain=0x0000000000000000, log_level=G_LOG_LEVEL_ERROR, format="* Assertion at %s:%d, condition `%s' not met\n", args=0x00007fff5df1e3b0) at goutput.c:115
    frame #10: 0x00000001021dd6c4 mono`monoeg_assertion_message(format="* Assertion at %s:%d, condition `%s' not met\n") at goutput.c:135
    frame #11: 0x0000000101e83d96 mono`amd64_patch_aligned(code="?, target=0x00000001025cecd6, aligned=1) at mini-amd64.c:214
    frame #12: 0x0000000101ee97bf mono`amd64_patch(code="?, target=0x00000001025cecd6) at mini-amd64.c:226
    frame #13: 0x0000000101eec41a mono`mono_arch_patch_code_new(cfg=0x00007f8534800000, domain=0x00007f8531c10170, code="UH\x8b?H??`L\x89e?L?m?L?u?L?}?L\x8b?L??H?U\xa8L\x8b?H?E\xb8", ji=0x00007f853480d210, target=0x00000001025cecd6) at mini-amd64.c:6670
    frame #14: 0x0000000101ce805c mono`mono_codegen(cfg=0x00007f8534800000) at mini.c:2351

jaykrell:
Ping? Maybe I should write a varargs tailcall..

luhenry:
@vargaz is that PR good to go? Please approve if so. Thank you.

jaykrell:
@lateralusX fixed some NT/amd64 ABI violation fixes -- epilogue must adjust rsp, restore nonvolatiles, and transfer control (jmp, ret, iret). It cannot mov once it starts the nonvolatile restore -- it must be recognized by the unwinder, as seen here: https://github.com/dotnet/coreclr/blob/master/src/unwinder/amd64/unwinder_amd64.cpp#L460

jaykrell:
See here also for the ABI violation fixed:
https://github.com/dotnet/coreclr/blob/master/src/unwinder/amd64/unwinder_amd64.cpp#L1212

PR #7607, opened 2018-03-13
[mini] Improve error reporting by using MonoError around more places.

kumpera:


mono/mini/mini.h
```diff
@@ -2488,12 +2488,18 @@ void mono_generic_sharing_cleanup (void);
 MonoClass* mini_class_get_container_class (MonoClass *klass);
 MonoGenericContext* mini_class_get_context (MonoClass *klass);
 
+typedef enum {
+	SharedModeNone = 0x0,
+	AllValueTypes = 0x1,
+	GSharedVT = 0x2,
+} GetSharedMethodFlags;
```
vargaz: These should be uppercase.

mono/mini/mini-generic-sharing.c
```diff
@@ -3651,26 +3655,29 @@ get_shared_inst (MonoGenericInst *inst, MonoGenericInst *shared_inst, MonoGeneri
...
- * METHOD can be a non-inflated generic method.
+ * \returns The method which is actually compiled/registered when doing generic sharing.
+
+ * If flags & SHARE_MODE_GSHAREDVT, produce a method using the gsharedvt instantiation.
+ * \p method can be a non-inflated generic method.
  */
 MonoMethod*
-mini_get_shared_method_full (MonoMethod *method, gboolean all_vt, gboolean is_gsharedvt)
+mini_get_shared_method_full (MonoMethod *method, GetSharedMethodFlags flags, MonoError *error)
 {
```
vargaz: Now that this only has one argument, shouldn't the previous bool is_gsharedvt be enough ?

vargaz: Why do we want to pass up the error ? Is there a case where this function can legitimely fail ?

PR #7612, opened 2018-03-14
[System.Windows.Forms] Unify layout engines for Controls and ToolStripItems

filipnavara:
Lot of the ToolStrip layout code was held by duct tape. This PR introduces the IArrangedElement and IArrangedContainer interfaces that are implmented by Control, ToolStripItem and other ToolStrip classes. This facilitates sharing the same layout engine code for both regular controls and the tool strips. The benefit is that it enables use of table layout in ToolStrips (as used by StatusStrip) and proper use of FlowLayout in ToolStripDropDown/ToolStripDropDownMenu/ToolStripOverflow. It makes the overflow code working for KeePass on X11 backend, properly flowing the buttons and combo box from the tool strip into the overflow window.


filipnavara:
The duct tape layout is still kept for ToolStripPanel and ToolStripPanelRow since I had no code to test it.

filipnavara:
As for the API diff, there are two options. I can either keep it as-is, the public API doesn't change, only protected overridden methods, or I can move the code from CreateLayoutSettings into OnLayout. The later option is less clear in the intent.

mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/DefaultLayout.cs
```diff
@@ -43,33 +43,33 @@ private DefaultLayout ()
 	
 		public override void InitLayout (object child, BoundsSpecified specified)
 		{
-			Control control = child as Control;
-			Control parent = control.Parent;
+			IArrangedElement control = child as IArrangedElement;
```
marek-safar: Use IArrangedElement explicit cast

mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/FlowLayout.cs
```diff
@@ -47,24 +47,26 @@ public override void InitLayout (object child, BoundsSpecified specified)
 			base.InitLayout (child, specified);
 		}
 
+		private FlowLayoutSettings GetLayoutSettings (IArrangedContainer container)
```
marek-safar: should be static

mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/FlowLayout.cs
```diff
@@ -47,24 +47,26 @@ public override void InitLayout (object child, BoundsSpecified specified)
...
+		{
+			if (container is FlowLayoutPanel)
```
marek-safar: Use `if (container is FlowLayoutPanel flp)` syntax

mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/FlowLayout.cs
```diff
@@ -47,24 +47,26 @@ public override void InitLayout (object child, BoundsSpecified specified)
...
-				return LayoutToolStrip ((ToolStrip)container);
-				
-			Control parent = container as Control;
-			
-			FlowLayoutSettings settings;
-			if (parent is FlowLayoutPanel)
-				settings = (parent as FlowLayoutPanel).LayoutSettings;
-			else
-				settings = default_settings;
+			IArrangedContainer parent = container as IArrangedContainer;
```
marek-safar: Change it to explicit cast

marek-safar:
API diff changes are fine, they are not breaking changes

filipnavara:
Thanks for review! I updated the code with appropriate fixes.

marek-safar:
@monojenkins commit apidiff

PR #7621, opened 2018-03-15
[Tailcall] Proper fix for F# regression.

jaykrell:
(regression test not yet available; existing repro is 200MB+)


jaykrell:
This helps, but testing has shown it is probably inadequate.

jaykrell:
This fix is correct. Tests coming soon.

mono/mini/mini-trampolines.c
```diff
@@ -732,10 +732,6 @@ common_call_trampoline (mgreg_t *regs, guint8 *code, MonoMethod *m, MonoVTable *
 		need_rgctx_tramp = FALSE;
 	}
 
-	/* Calls made through delegates on platforms without delegate trampolines */
-	if (!code && mono_method_needs_static_rgctx_invoke (m, FALSE))
-		need_rgctx_tramp = TRUE;
-
```
vargaz: There is a reason this code is there, why is it removed ?

jaykrell: It breaks things, and does not seem to fix anything -- still to run a full CI/test pass.

The comment is about delegates, but it actually triggers for jmp trampolines, and appears to trigger for nothing else, including not for delegates.

From what I see, the only time code is null, is for jump trampolines.
If we get here for a call, or anything else, this code does nothing.
If we get here for a jump, this code sometimes sets need_rgctx_tramp to true and that breaks things. I don't fully understand what this trying to do vs. what is already done, but I have a vague sense that this is redundant, and then less clearly, does the wrong thing.

In any case, there are tests now..and either this is the right fix, or something is wrong elsewhere or nearby.

mono/mini/method-to-ir.c
```diff
@@ -9024,8 +9033,13 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			if (ins)
 				goto call_end;
 
+#if 0 // debug code -- attempt to tailcall every call followed by a ret
```
luhenry: Use a `#define` which conveys what you are trying to debug. Also you can factor the `gboolean inst_tailcall` to reduce chances of code bitrot.

mono/mini/method-to-ir.c
```diff
@@ -2173,6 +2174,15 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 	if (cfg->llvm_only)
 		tail = FALSE;
 
+#if 0 // debug code; can change tail to false in debugger
```
luhenry: Use a `#define` which conveys what you are trying to debug

mono/mini/mini-runtime.c
```diff
@@ -3473,6 +3473,21 @@ mono_get_delegate_virtual_invoke_impl (MonoMethodSignature *sig, MonoMethod *met
...
+			return tailcall->break_compile = TRUE;
+		else if (strcmp (option, "tailcall_break_run") == 0)
+			return tailcall->break_run = TRUE;
+		else if (strcmp (option, "tailcall_log") == 0)
+			return tailcall->log = TRUE;
+		else if (strcmp (option, "tailcall_try_all") == 0)
+			return tailcall->try_all = TRUE;
+		else if (strcmp (option, "tailcall_all") == 0)
+			return tailcall->all = -1;
+		return FALSE;
```
jaykrell: Could also remove the wierd returns and say "else return FALSE; return TRUE;"

mono/mini/mini-trampolines.c
vargaz: Why does this break things ?

mono/mini/mini-runtime.c
```diff
@@ -3473,6 +3473,21 @@ mono_get_delegate_virtual_invoke_impl (MonoMethodSignature *sig, MonoMethod *met
 gboolean
 mini_parse_debug_option (const char *option)
 {
+	MonoDebugTailcallOptions * const tailcall = &mini_debug_options.tailcall;
+	switch (option [0]) {
+	case 't':
+		if (strcmp (option, "tailcall_break_compile") == 0)
```
luhenry: This is completely unnecessary and overengineered over what's actually needed. Even the previous approach of `#if 0` is much better than this.

mono/mini/mini-trampolines.c
jaykrell: It breaks the tests and F#, if you allow tailcall with vtable_arg.
I didn't determine what is the new parameter in r10 or r11. What was already coming in was correct and works, w/o this code changing things.
I noticed 1) it is where tailcall vs. non-tailcall diverge and 2) it changes if code works or not.

mono/mini/mini-runtime.c
jaykrell: MONO_DEBUG is a thing. I didn't make it up.

You should be able to debug w/o edit and recompile/relink.

I find myself constantly changing the logging to be enabled under -v instead of -v -v, so I am not flooded with other output -- that even grep takes a while to get through -- and which runs tools that aren't usually on PATH on windows (objdump or somesuch).

vargaz: MONO_DEBUG should contain options which are somewhat commonly used. Its pretty common to recompile while debugging.

jaykrell:
On ARM, new test fails which is maybe ok. Need to check if correctness vs. tailcalled. Mono_Debug=tailcall_log will be useful.

But an existing test also fails and that needs investigation.

jaykrell:
ARM32 shouldn't be ignored though right? Let me see..

jaykrell:
Arg, CI loses the link.

jaykrell:
Small regression test maybe to be added. Comment/uncomment the tail. to see bug.
Problem is that r8 is loaded with rgctx, and then nonvolatiles are restored, and then tailcall jmped to.
With non-tailcall , the nonvolatiles aren't restored till after the call and rgctx makes it through.

```
.assembly extern mscorlib { }

.assembly 'tailcall-generic-cast' { }

.class public B
{
        .method public instance void .ctor ( )
        {
                ldarg.0
                call instance void [mscorlib]System.Object::.ctor ( )
                ret
        }
}

.class public C
{
        .method public static void
        cast1<T> (object o) noinlining
        {
                ldarg.0
                unbox.any !!T
                pop
                ret
        }

        .method public static void
        cast2 (object o) noinlining
        {
                ldarg.0
//              tail.
                call void C::cast1<class B> (object)
                ret
        }

}

.class public E
{
        .method public static void
        Main (string[] args) noinlining
        {
                .entrypoint

                newobj instance void B::.ctor ( )
                call void C::cast2 (object)
                ret
        }
}
```

luhenry:
I missed the failing ARM builds.

jaykrell:
I don't think CI of the update had finished, but only ARM32 was of concern.

PR #7627, opened 2018-03-15
[jit] Incorrect code under mono --trace.

jaykrell:


jaykrell:
Also, a test case is possible here, but not present.
It'd need to be very very small, and use mono --trace.
I looked quickly at the other architectures and they might be ok.

The failure mode is kinda arbitrary incorrect code -- regoffset instruction with arbitrary offset.
But only under mono --trace.

akoeplinger:
Failures are unrelated, merging.

PR #7617, opened 2018-03-14
[sdks] IOS fixes.

vargaz:

PR #7628, opened 2018-03-15
[2017-12] [sdks] IOS fixes. (#7617)

vargaz:
* [sdks] Fix ios build.

* [sdks] Pass -DMONOTOUCH to the ios cross compiler builds. Compile monotouch/tvos/watchos profiles for the bcl build.


lewurm:
also needs to be backported to `2018-02`

PR #7634, opened 2018-03-15
Fix for stepping in VSTU taking multiple steps to get to next line when there is a breakpoint

brianradunity:
process_breakpoint() will process breakpoint, stepping, and method
exit/entry events all at the same time when checking a sequence point.
So, if there is is a breakpoint and a stepping event at the same time,
the debugger agent will suspend the current thread twice, which means
you have to click the step button in the debugger client twice on that
line.  Considering only one of these events at a time seems to fix
the issue without breaking any current unit tests or showing any other
noticeable issues in testing.


luhenry:
@vargaz @alexanderkyte can you pelase review?

marek-safar:
approve

luhenry:

luhenry:
@monojenkins build Linux x64

luhenry:

PR #7639, opened 2018-03-15
[sdks] Build the offsets tool after running autogen.sh, otherwise parallel builds will fail since the offsets tool build is done in the srcdir.

vargaz:


jaykrell:
So..do we need that build vs. host vs. target kind of thing?
i.e. this is "build" tool, whereas almost all software and build systems just deal with "host"?

But there is plenty of precedent, gcc uses autoconf and deals reasonably with the potentially three platforms in one build (though again, mostly host).

genmdesc(.c)is also "build", and if we make such a clear and correct split, writing build tools in C isn't so bad, vs. restricting to "portable executables".

As well, mcs should be handled this way, if it isn't already.

luhenry:
Then we should add the dependency on `build-ios-$(1)`

vargaz:
There is no such target anymore, they are generated by this rule:

$(foreach target,$(TARGETS),build-$(target)): build-%: .stamp-%-configure
	$(MAKE) build-custom-$*

luhenry:
There are such targets: by defining `TARGETS += $(1)` at https://github.com/vargaz/mono/blob/0af4951e99bf6ef6dbaf50a4995b4b5dd98e2a23/sdks/builds/runtime.mk#L85 and passing `ios-$(1)` at https://github.com/vargaz/mono/blob/0af4951e99bf6ef6dbaf50a4995b4b5dd98e2a23/sdks/builds/ios.mk#L386 you are generating the following targets:

```
build-ios-cross32 build-ios-cross64 build-ios-crosswatch: build-%: .stamp-%-configure
	$(MAKE) build-custom-$*
```

https://github.com/mono/mono/pull/7652/commits/993566e69a894dbad964b6ca195d0cdcf4da505f is the way to fix it.

PR #7640, opened 2018-03-15
Fix bug #60843 and add test

kg:
If an enum was nested inside a generic class, using it as an argument to a custom attribute would cause a runtime error in the custom attribute decoder. Update the decoder to forward this special case through the normal primitive loading path.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=60843


mono/metadata/custom-attrs.c
```diff
@@ -215,6 +220,43 @@ load_cattr_enum_type (MonoImage *image, const char *p, const char *boundp, const
...
 
+static MonoReflectionType*
+load_cattr_type_object (MonoImage *image, MonoType *t, const char *p, const char *boundp, const char **end, MonoError *error, guint32 *slen)
+{
+	MonoReflectionType *rt;
+	char *n;
+
+	if (!decode_blob_value_checked (p, boundp, slen, &p, error))
+		return NULL;
+	if (slen > 0 && !bcheck_blob (p, *slen - 1, boundp, error))
```
lambdageek: should be `*slen > 0` ?

(i'm also uncertain if `slen` should be modifier on error paths (as you have it now) or only on success.)

mono/tests/bug-60843.cs
```diff
@@ -0,0 +1,38 @@
...
+class A : Attribute
+{
+    public object X;
+
+    public static void Main()
+    {
+        var x = (C<int>.E)AttributeTest(typeof(C<>.E));
+        Assert(C<int>.E.V == x);
+        var y = (C<int>.E2[])AttributeTest(typeof(C<>.E2));
+        Assert(y.Length == 2);
```
lambdageek: Maybe pass a `String` to each assert so we could tell from CI which piece of the test failed? or use exit codes like other `mono/tests` tests or the TestDriver.cs - style tests.

mono/metadata/custom-attrs.c
lambdageek: should be `*slen > 0`

(i'm also uncertain if `slen` should be modified on error paths (as you have it now) or only on success.)

mono/tests/bug-60843.cs
```diff
@@ -0,0 +1,48 @@
...
+        var a = (A)cas[0];
+        return a.X;
+    }
+
+    private static int AssertCount = 0;
+
+    public static void Assert (
+        bool b, 
+        [CallerFilePath] string sourceFile = null, 
+        [CallerLineNumber] int lineNumber = 0
```
akoeplinger: FullAOT profiles don't have these attributes:

```
bug-60843.cs(29,10): error CS0246: The type or namespace name 'CallerFilePathAttribute' could not be found (are you missing a using directive or an assembly reference?)
bug-60843.cs(29,10): error CS0246: The type or namespace name 'CallerFilePath' could not be found (are you missing a using directive or an assembly reference?)
bug-60843.cs(30,10): error CS0246: The type or namespace name 'CallerLineNumberAttribute' could not be found (are you missing a using directive or an assembly reference?)
bug-60843.cs(30,10): error CS0246: The type or namespace name 'CallerLineNumber' could not be found (are you missing a using directive or an assembly reference?)
```

kg: Weird, thanks for pointing it out

marek-safar: Well, that's quite weird. We should add them

luhenry:
The newly added test is failing with the interpreter:
https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/3933/testReport/junit/MonoTests/runtime-interp/bug_60843_exe/
https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-armhf/3443/testReport/junit/MonoTests/runtime-interp/bug_60843_exe/

kg:
Thanks for the heads-up ludovic, I'll look into interp once I figure out the caller info attributes

luhenry:
@kg it might be that the interpreter just doesn't support the necessary feature, so just disabling the test with interpreter could be ok.

@BrzVlad does that change looks like it's regressing the interpreter, or is it just something that it doesn't support yet?

BrzVlad:
@luhenry Not sure but I think it might just expose something that we're doing wrong there. Doubt it's a regression.

kg:
The interpreter failure is definitely a bug, the attribute is getting a garbage value instead of the correct enum value. Looking into it

kg:
The interpreter failure is some sort of bug in the interpreter or a really obscure bug in the cattr loader I can't identify. Tracing through it everything seems to be working (loading the right value from the file, constructing a valid mono object, etc). So the boxed object might be getting collected or who knows what. On that basis I think maybe we should merge this (I'll disable the test on the interpreter) - thoughts?

lewurm:
@kg since you filed https://github.com/mono/mono/issues/7698, disabling it on the interpreter + merge sounds good to me 👍

kg:
7698 is actually a separate bug, but I'll file a bug for the test failure and disable the test. Thanks.

kg:
Filed https://github.com/mono/mono/issues/7752 for the interpreter test failure

PR #7644, opened 2018-03-15
Imports FtpWebRequest from CoreFX

EgorBo:
Contributes to #7682

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=12850
FtpWebRequest used to work incorrectly with sub directories.
Let's say we have a following FTP structure
```
myserver
├── file1
└── subdir
    └── file2
    └── file3
```
Old behavior:

### 1) For `ftp://myserver/subdir/` printed:
```
file2
file3
```
Expected:
```
file2
file3
```

### 2) For `ftp://myserver/subdir` printed:
```
file1
subdir
```
Expected:
```
subdir/file2
subdir/file3
```
### 3) For `ftp://myserver/subdir/file2` printed:
```
file2
file3
```
Expected:
```
file2
```

With the fix in this PR it works as expected (matches netfx and netcore behaviors).


EgorBo:
There is an issue with the hand-made ftp server we use in tests - I could not connect to it via .NET, it throws `The underlying connection was closed: The server committed a protocol violation`

mcs/class/referencesource/mscorlib/system/threading/LazyInitializer.cs
```diff
@@ -156,6 +156,40 @@ public static class LazyInitializer
             return target;
         }
 
+#if MONO
```
EgorBo: a few methods from coref'x LazyInitializer.
LazyInitializer can be switched to corefx implementation but will require a small change in corefx

mcs/class/System/System_test.dll.sources
```diff
@@ -217,7 +217,7 @@ System.Net/DnsTest.cs
 System.Net/EndPointTest.cs
 System.Net/FileWebRequestTest.cs
 System.Net/FileWebResponseTest.cs
-System.Net/FtpWebRequestTest.cs
+#System.Net/FtpWebRequestTest.cs
```
EgorBo: Old FTP tests use a simple hand-made FTP Server based on raw sockets but it's not compatible with .NET and .NET Core (both refuse to connect to it due to some protocol violations).
Instead of this test I enabled a corefx one below - tested on my local ftp server, all pass.

mcs/class/System/common_networking.sources
```diff
@@ -20,12 +20,17 @@ System.Net/AuthenticationManager.cs
 System.Net/ChunkedInputStream.cs
 System.Net/EndPointListener.cs
 System.Net/EndPointManager.cs
-System.Net/FtpAsyncResult.cs
-System.Net/FtpDataStream.cs
+
 System.Net/FtpRequestCreator.cs
```
akoeplinger: This can be replaced by FtpWebRequestCreator class in corefx/src/System.Net.Requests/src/System/Net/FtpWebRequest.cs

EgorBo: @akoeplinger thanks! However this change will require changes in machine.configs

marek-safar: That can be done with follow up commit

mcs/class/System/common_networking.sources
```diff
@@ -21,7 +21,6 @@ System.Net/ChunkedInputStream.cs
 System.Net/EndPointListener.cs
 System.Net/EndPointManager.cs
 
-System.Net/FtpRequestCreator.cs
```
akoeplinger: The file can be deleted too.

I also just saw we have FtpRequestCreator.platformnotsupported.cs for WatchOS, that will need adjustments too (probably by moving the code into the corefx FtpWebRequest.cs)

akoeplinger:
@monojenkins commit apidiff

PR #7658, opened 2018-03-16
[Tailcall] Proper fix for F# regression, at least some architectures.

jaykrell:
Try this again -- it got commited and reverted, while possibly waiting for CI to finish again, due to ARM32-related update.

Fixes https://github.com/mono/mono/issues/7558


jaykrell:
(non-required ARM32 to be clear -- perhaps they should be required?? at least ARMv7/ARMHF?)

luhenry:
For the failures on ARM: no platforms should regress as part of a change, but their is no mandate to get all the platforms on-par as part of one PR. What this means in this PR case: as long as ARM is not broken by this change, it's ok to have the tests disabled there for now.

But, please keep in mind that we want to have x86, x86_64, arm and arm64 on-par regarding this tailcall work, so you will have to come back to it at some point.

jaykrell:
Fixing this for ARM is going to be exceedingly difficult, possibly not possible.

mono/mini/method-to-ir.c
```diff
@@ -2173,6 +2179,23 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 	if (cfg->llvm_only)
 		tail = FALSE;
 
+#if defined (DEBUG_TAILCALL_BREAK_COMPILE) || defined (DEBUG_TAILCALL_BREAK_RUN) || defined (DEBUG_TAILCALL)
+	if (tail && mono_is_usermode_native_debugger_present ()) {
+
```
vargaz: Putting ifs inside #ifdefs confuses the indentation of some editors, so its discouraged.

vargaz: I'd have preferred the old #if 0 code, its just there for debugging, I don't think commented out code like that needs to be named with specific ifdefs.

mono/mini/method-to-ir.c
```diff
@@ -6996,8 +7019,25 @@ is_supported_tail_call (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmetho
...
+		// Imagine F1 calls F2, and F2 tailcalls F3.
+		// F2 and F3 are managed. F1 is native.
+		// Without a tailcall, F2 can save and restore everything needed for F1.
+		// However if the extra parameter were in a non-volatile, such as ARM32 V5/R8,
+		// F3 cannot easily restore it for F1, in the current scheme. The current
+		// scheme where the extra parameter is not merely an extra parameter, but
+		// passed "outside of the ABI".
+		|| (vtable_arg && !cfg->backend->have_volatile_non_param_register)
+
 		|| ((vtable_arg || cfg->gshared) && !cfg->backend->have_op_tail_call)
```
vargaz: Extra empty line here.

mono/mini/mini-amd64.h
```diff
@@ -476,6 +476,10 @@ typedef struct {
             MONO_EMIT_NEW_COND_EXC (cfg, LE_UN, "IndexOutOfRangeException"); \
        } while (0)
 
+// Does the ABI have a volatile non-parameter register, so tailcall
+// can pass context to generics or interfaces?
+#define MONO_ARCH_HAVE_VOLATILE_NON_PARAM_REGISTER 1
+
```
vargaz: This is ok, I'd have preferred simply MONO_ARCH_HAVE_GSHARED_TAILCALLS or something, 'have non volatile non param register' looks way to generic.

jaykrell: I believe this covers more than just GSHARED_TAILCALLS, which is why I named it this way. It will cover tailcalls to interface methods too.

mono/mini/method-to-ir.c
jaykrell: Either that is line wrap, or it is better bind the comment to the line.

mono/mini/method-to-ir.c
```diff
@@ -6996,8 +7019,25 @@ is_supported_tail_call (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmetho
 		|| cfg->method->save_lmf
 		|| (cmethod->wrapper_type && cmethod->wrapper_type != MONO_WRAPPER_DYNAMIC_METHOD)
 		|| call_opcode == CEE_CALLI
-		|| ((virtual_ || call_opcode == CEE_CALLVIRT) && !cfg->backend->have_op_tail_call_membase)
-		|| vtable_arg // FIXME
+		|| (virtual_ && !cfg->backend->have_op_tail_call_membase)
+
+		// Passing vtable_arg uses (requires?) a volatile non-parameter register,
```
vargaz: It would probably be better to put these comments into some arm specific file like mini-arm.h.

mono/mini/method-to-ir.c
jaykrell: Ludovic did not like if 0, wanted a name, and I don't think they deserve the same name, really.

mono/mini/mini-amd64.h
vargaz: Ok.

mono/mini/method-to-ir.c
luhenry: In the end, it's up to you @vargaz.

mono/mini/method-to-ir.c
jaykrell: There is a problem with the comment and the define that neither is specific about rgctx being in the register alluded to.

mono/mini/method-to-ir.c
jaykrell: if and ifdef seem like pretty basic features that we should be able to use, together. I could make a const static gboolean and just use if, not idef. The compiler will see through it and optimize and the code will not be allowed to rot.

vargaz: Doing 
#ifdef 
  if ()
     ..;
#endif 
is ok, but doing
#  ifdef
   if ()
#endif 
is not.
The static gboolean approach is ok too.

mono/mini/mini-arm.h
```diff
@@ -373,6 +373,10 @@ typedef struct MonoCompileArch {
 /* First argument reg */
 #define MONO_ARCH_VTABLE_REG ARMREG_R0
 
+// Does the ABI have a volatile non-parameter register, so tailcall
+// can pass context to generics or interfaces?
+#define MONO_ARCH_HAVE_VOLATILE_NON_PARAM_REGISTER 0
```
jaykrell: Well darn it, there is an easy mistake -- if vs. ifdef.
Obviously I mixed them up purely by accident, I know the language.
I assume we want to remove this line and stick with pound-ifdef, but pound-if and regular-if will also work.

luhenry:

luhenry:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

Issue #7624, opened 2018-03-15
Strict loader still loads incorrect assembly


marek-safar:
## Steps to Reproduce

1. Unzip the attachment [Archive.zip](https://github.com/mono/mono/files/1814739/Archive.zip)
2. `mono --assembly-loader=strict nunit-console.exe Library.Tests.dll`

## Current Behavior

```
Library.Tests.Test.TestCase : System.TypeLoadException : Could not resolve type with token 01000001 from typeref (expected class 'System.Object' in assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a')
```

## Expected Behavior

No error as System.Runtime assembly in the folder should not be used

### On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows

**Version Used**:
```
Mono JIT compiler version 5.13.0 (master/1c88a115453 Thu Mar 15 00:03:33 CET 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug
	Interpreter:   yes
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
```

lambdageek:
So it's doing this:
```
Mono: Assembly System.Runtime is a framework Facade asseembly
Mono: Assembly System.Runtime is a framework Facade asseembly
Mono: Domain domain-Library.Tests.dll search path is:
Mono: 	path[0] = '/Users/alekseyk/work/bugs/gh-7624/gh'
Mono: End of domain domain-Library.Tests.dll search path.
Mono: Assembly Loader probing location: '/Users/alekseyk/work/bugs/gh-7624/gh/System.Runtime.dll'.
Mono: Assembly Loader shadow-copied assembly to: '/var/folders/xj/lznyzvqx45j_qy49h8f09ld80000gn/T/nunit20/ShadowCopyCache/22905_636567142141853290/Tests/assembly/shadow/8415508f/b3c63656_37d366d9_00000001/System.Runtime.dll'.
[...]

PR #7663, opened 2018-03-16
[loader] Don't load framework Facade assemblies that are older than expected.

lambdageek:
If there is an older framework Facade assembly in the search order before the
one that we want, don't load it.

"older" here means just compare the major component of the version.

Fixes https://github.com/mono/mono/issues/7624


lambdageek:
Is this what we want?  The rules for these assemblies where we want to override the normal strong name logic are getting pretty complicated...

marek-safar:
I am not sure about minor (e.g. 4.0 and 4.1) but quite sure about major version mismatch

mono/metadata/assembly.c
```diff
@@ -666,6 +666,39 @@ assembly_names_equal_flags (MonoAssemblyName *l, MonoAssemblyName *r, AssemblyNa
...
+ */
+static int
+assembly_names_compare_versions (MonoAssemblyName *l, MonoAssemblyName *r, int maxcomps)
+{
+	int i = 0;
+	if (maxcomps < 0) maxcomps = 4;
+#define CMP(field) do {				\
+		if (l-> field < r-> field && i < maxcomps) return -1;	\
+		if (l-> field > r-> field && i < maxcomps) return 1;	\
+	} while (0);
```
luhenry: Stylistic: no need for a trailing `;`

lambdageek:

lambdageek:

Issue #7668, opened 2018-03-17
Race condition between `WebOperation` and `ServicePointScheduler`.


baulig:
The latest `WebCompletionSource<T>` changes (PR #7358) exposed and intensified the side-effects of a race condition between `WebOperation` and `ServicePointScheduler`.

This race condition already existed prior to those changes, but it was much harder to trigger and reproduce.  It can now reliably be triggered by the web-tests via [`HttpClientTestType.PutRedirectKeepAlive`](https://github.com/xamarin/web-tests/blob/9d3eabdf101597bf05fffa9e326a01bc545ae103/Xamarin.WebTests.Framework/Xamarin.WebTests.TestRunners/HttpClientTestType.cs#L51).

The problem is that the `ServicePointScheduler` [waits](https://github.com/mono/mono/blob/4b88e2a0e29b9bbb8e5ec4e0994c5587b27c302d/mcs/class/System/System.Net/ServicePointScheduler.cs#L163) on a task that's completed by [`WebOperation.FinishReading()`](https://github.com/mono/mono/blob/4b88e2a0e29b9bbb8e5ec4e0994c5587b27c302d/mcs/class/System/System.Net/WebOperation.cs#L333).

When redirecting on the current connection, `HttpWebRequest` completes this via [`WebResponseStream.ReadAllAsync()`](https://github.com/mono/mono/blob/4b88e2a0e29b9bbb8e5ec4e0994c5587b27c302d/mcs/class/System/System.Net/WebResponseStream.cs#L402) in [`MyGetResponseAsync()`](https://github.com/mono/mono/blob/4b88e2a0e29b9bbb8e5ec4e0994c5587b27c302d/mcs/class/System/System.Net/HttpWebRequest.cs#L1054) prior to queuing the new request.

On rare occasions, the `ServicePointScheduler`s main loop may get woken up by this new request prior to completing the previous one.  This would result in the connection being redirected on a new connection.

baulig:
I have a PR to fix this, which will be up shortly.

baulig:
To trigger this prior to the `WebCompletionSource<T>` changes, checkout an older commit (such as `dfe6e2d` for instance), go to [`WebOperation.FinishReading()`](https://github.com/mono/mono/blob/dfe6e2d4a2c0cc05f63b3421207daf4dccf6ff5e/mcs/class/System/System.Net/WebOperation.cs#L324) and insert `await Task.Delay (500)` prior to the `finishedTask.TrySetResult()` call.

PR #7358, opened 2018-03-01
[System]: Introduce `WebCompletionSource<T>` and use it in `WebOperation`.

baulig:
Fixes #7194
Fixes #7355

PR #7669, opened 2018-03-17
[System]: Fix a race condition that was exposed by #7358 (see #7668).

baulig:
The latest `WebCompletionSource<T>` changes (PR #7358) exposed and
intensified the side-effects of a race condition between `WebOperation`
and `ServicePointScheduler`.

See #7668 for an in-depth explanation of this race condition.

* `WebCompletionSource<T>` now contains new `CurrentResult` and `CurrentStatus`
  properties, which allow completion status to be checked in a non-blocking way.

* `WebOperation.CompleteResponseRead()` has been renamed into `Finish()` and
  contains the continuation that was previouly known as `FinishReading()`.
  Calling this new `Finish()` method will now complete the `finishedTask`
  completion, which has a new accessor called `Finished`.

* The ServicePointScheduler's main-loop is still woken up by completion of
  `WebOperation.Finished.Task`, but we use the new `CurrentResult` property
  to check state in a non-blocking way.  This ensures the the clean-up and
  bookkeeping work that's necessary after `WebCompletion.Finish()` always
  happens prior to starting new connections.

### DIFF
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -983,7 +983,7 @@ namespace System.Net
 				WebConnection.Debug ($"HWR GET RESPONSE: Req={ID} {oldCompletion != null}");
 				if (oldCompletion != null) {
 					oldCompletion.ThrowOnError ();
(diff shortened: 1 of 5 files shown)


marek-safar:
@monojenkins backport 2018-02

PR #7670, opened 2018-03-17
Update server.py to run on both python 2 and 3

davidlsharp1:
This is just a small python code change so users with both python 2 and 3 should be able to use the server.py


davidlsharp1:
deleting branch since pull request successfully merged and closed.

PR #7671, opened 2018-03-17
[runtime] Fix mono_msec_boottime () so it compiles on older ios versions, clock_gettime () is only available from ios 10.

vargaz:


akoeplinger:
We'll need to backport to 2018-02 and 2017-12. Though given that we already found a problem with this I wonder if we should just revert the boottime change in 2017-12?

vargaz:
It would be better to revert for now.

vargaz:
@monojenkins build OS X x64

akoeplinger:
@monojenkins backport 2018-02

akoeplinger:
Oh, seems @alexischr already reverted in 2018-02: https://github.com/mono/mono/commit/c2fe39f13efb5aea8df6651052d06963a1d2288b

PR #7676, opened 2018-03-19
[metadata] Don't leak methods when running with a profiler attached.

### LINKED ISSUE #6171: Runtime should free methods when profilers are attached
`mono_free_method` currently does nothing if a profiler is attached. This is because of a bunch of hacks in the log profiler that result in keeping `MonoMethod` pointers around potentially for longer than their actual lifetime. The thread-local method list is one such hack.

We should investigate whether this can be fixed.

alexrp:
Let's see what happens now that #7555 is merged.

Fixes #6171.

### DIFF
--- a/mono/metadata/loader.c
+++ b/mono/metadata/loader.c
@@ -1955,11 +1955,7 @@ void
 mono_free_method  (MonoMethod *method)
 {
 	MONO_PROFILER_RAISE (method_free, (method));
-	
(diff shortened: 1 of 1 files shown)


alexrp:
This seems to break IronJS. Will investigate.

alexrp:
Will revisit this later. Closing for now.

Issue #7449, opened 2018-03-06
PInvoke to function that takes a struct with an array of double gets incorrect results on Android


brianradunity:
## Steps to Reproduce

This issue was discovered by a Unity user running a Mono player on an Android device.  When calling the DLLImport function defined below on Android, the native function receives incorrect data.

Source code for native function is located here: https://github.com/Gobra/UAM_Params/tree/master/source/cpplib-1.0

This corresponds to Unity issue: https://issuetracker.unity3d.com/issues/android-mono-dllimport-function-doesnt-receive-passed-parameters

c# code calling native function:

```csharp
[Serializable]
[StructLayout(LayoutKind.Sequential)]
public struct Rectangle
{
	public int X;
	public int Y;
	public int Width;
	public int Height;
		
	public Rectangle(int x, int y, int width, int height)
	{
		X = x;
		Y = y;
		Width = width;
		Height = height;
	}
}

[Serializable]
public struct Scalar
{
	public double Val0;
	public double Val1;
	public double Val2;
	public double Val3;

	public Scalar(double v0, double v1, double v2, double v3)
	{
		Val0 = v0;
		Val1 = v1;
		Val2 = v2;
		Val3 = v3;
	}
}

/// <summary>
/// DllExports
/// </summary>
internal static class Native
{
	[DllImport("CppLib", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
	[return: MarshalAs(UnmanagedType.LPStr)]
	public static extern string binding(IntPtr ptr, Rectangle rect, Scalar color, int thickness, int lineType, int shift);
}
 public static void main()
{
    string formatted = Native.binding(IntPtr.Zero, new Rectangle(10, 10, 100, 20), new Scalar(32, 64, 128, 256), 1, 2, 3);
}
```

## Current Behavior

Return string is:
"ptr, {10, 10, 100, 20 }, color = { 0.0, 0.0, 0.0, 0.0 }, 2, 3, 0"

## Expected Behavior

The output string of the above native.binding() call should be:
"ptr, { 10, 10, 100, 20 }, color = { 32.0, 64.0, 128.0, 256.0 }, 1, 2, 3"

### On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows
[x] Android

**Version Used**:

recent master

### Stacktrace

```
Please paste the stacktrace here if available.
```

EgorBo:
[...]

PR #7691, opened 2018-03-19
[arm] fix stack alignment for structs

lewurm:
fixes https://github.com/mono/mono/issues/7449

/cc @joncham

### DIFF
--- a/mono/tests/pinvoke11.cs
+++ b/mono/tests/pinvoke11.cs
@@ -60,6 +60,39 @@ public struct NestedFloat {
 	public float f4;
 }
 
+[Serializable]
+[StructLayout(LayoutKind.Sequential)]
+public struct Rectangle
+{
+	public int X;
+	public int Y;
+	public int Width;
+	public int Height;
+
+	public Rectangle(int x, int y, int width, int height)
+	{
+		X = x;
+		Y = y;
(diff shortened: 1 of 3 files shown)


vargaz:
Looks good in theory, hopefully it won't break anything.

lewurm:
I've tested it on Linux armv5 and armv7. I'm a bit concerned about iOS 32bit.

mono/tests/pinvoke11.cs
```diff
@@ -159,6 +196,12 @@ static int Main()
 		if (f.fi.f1 != 1.0)
 			return 12;
 
+		Rectangle rect = new Rectangle (10, 10, 100, 20);
+		Scalar4 sc4 = new Scalar4 (32, 64, 128, 256);
+		var sc4_ret = mono_return_struct_4_double (IntPtr.Zero, rect, sc4, 0x1337, 0x1234, 0x9876);
+		if (sc4_ret != "sc4 = {32.0, 64.0, 128.0, 256.0 }, a=1337, b=1234, c=9876\n")
+			return 13;
```
luhenry: You should print `sc4_ret` in case it fails, so we have some information in CI.

luhenry:

PR #7695, opened 2018-03-19
Add tailcall support & fix floating point ABI

nealef:
1. Implement OP_TAILCALL for s390x
-- Remove OP_JUMP

2. Fix float ABI implementation. basic-float fails as s390x is supposed to save FP registers 8-15 if they are used within a method. 
-- Added FP register tracking with cfg->arch.used_fp_regs
-- Added unwind support as FP registers map to dwarf registers 16-31

3. Enable BTLS build for s390x
-- Support is in boringssl submodule but configure.ac requires update

4. Prepare for float32 support
-- Added opcodes but haven't enabled


mono/mini/mini-s390x.c
```diff
@@ -7364,3 +7627,43 @@ mono_arch_opcode_supported (int opcode)
...
+	CallInfo *c1, *c2;
+	gboolean res;
+	int i;
+
+	c1 = get_call_info (NULL, caller_sig);
+	c2 = get_call_info (NULL, callee_sig);
+	res = c1->stack_usage >= c2->stack_usage;
+	if (callee_sig->ret && MONO_TYPE_ISSTRUCT (callee_sig->ret))
+		/* An address on the callee's stack is passed as the first argument */
+		res = FALSE;
```
jaykrell: goto end?

mono/mini/mini-s390x.c
```diff
@@ -7364,3 +7627,43 @@ mono_arch_opcode_supported (int opcode)
...
+	for (i = 0; i < c2->nargs; ++i) {
```
jaykrell: Or add res to the for loop terminal condition?

mono/mini/unwind.c
```diff
@@ -91,8 +91,15 @@ static int map_hw_reg_to_dwarf_reg [ppc_lr + 1] = { 0, 1, 2, 3, 4, 5, 6, 7, 8,
 #endif
 #define NUM_DWARF_REGS (DWARF_PC_REG + 1)
 #elif defined (TARGET_S390X)
-static int map_hw_reg_to_dwarf_reg [] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };
-#define NUM_DWARF_REGS 16
+/*
+ * 0-15 = GR0-15
+ * 16-31 = FP0-15
+ */
+static int map_hw_reg_to_dwarf_reg [] = {  0,  1,  2,  3,  4,  5,  6,  7,
```
jaykrell: const maybe?

mono/mini/mini-codegen.c
```diff
@@ -1039,6 +1039,9 @@ alloc_general_reg (MonoCompile *cfg, MonoBasicBlock *bb, MonoInst **last, MonoIn
 	if (val < 0)
 		val = get_register_spilling (cfg, bb, last, ins, dest_mask, sym_reg, bank);
 
+#ifdef MONO_ARCH_HAVE_TRACK_FPREGS
+	cfg->arch.used_fp_regs |= 1 << val;
+#endif
```
vargaz: This should be somewhere else.

vargaz: If I understand correctly, s390x uses callee saved fp regs for local register allocation ? It is the only platform that currently does that then. Since the registers need to be saved/restored, using them might actually lead to worse code.

nealef:
Where is the better place to put this?



@vargaz commented on this pull request.

In mono/mini/mini-codegen.c<https://github.com/mono/mono/pull/7695#discussion_r175612471>:

> @@ -1039,6 +1039,9 @@ alloc_general_reg (MonoCompile *cfg, MonoBasicBlock *bb, MonoInst **last, MonoIn
        if (val < 0)
                val = get_register_spilling (cfg, bb, last, ins, dest_mask, sym_reg, bank);

+#ifdef MONO_ARCH_HAVE_TRACK_FPREGS
+       cfg->arch.used_fp_regs |= 1 << val;
+#endif


This should be somewhere else.

nealef:
Maybe but none of the other entries are defined that way. I can update the other arch as well if it makes sense.

@jaykrell commented on this pull request.

In mono/mini/unwind.c<https://github.com/mono/mono/pull/7695#discussion_r175612059>:

> @@ -91,8 +91,15 @@ static int map_hw_reg_to_dwarf_reg [ppc_lr + 1] = { 0, 1, 2, 3, 4, 5, 6, 7, 8,
 #endif
 #define NUM_DWARF_REGS (DWARF_PC_REG + 1)
 #elif defined (TARGET_S390X)
-static int map_hw_reg_to_dwarf_reg [] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };
-#define NUM_DWARF_REGS 16
+/*
+ * 0-15 = GR0-15
+ * 16-31 = FP0-15
+ */
+static int map_hw_reg_to_dwarf_reg [] = {  0,  1,  2,  3,  4,  5,  6,  7,


const maybe?

nealef:
Added a terminal condition.

mono/mini/mini-codegen.c
nealef: If it is going to use that many registers (FP0-7 aren't saved) so by the time it gets to FP8 onwards it won't be much different to situations where we have to spill anyway. Also, I'd eventually like to implement SIMD and that will require those registers as well.

PR #7699, opened 2018-03-20
add some OpenBSD patches

thfrwn:
Hi,

A few trivial additions of OpenBSD to ifdefs, plus the change of the abort signal to SIGUSR1 which fixes the SIGABRT problem (other compilation problems remain). SIGTTIN is broadcast to the process group which leads to the SIGABRT (at least on OpenBSD).

### DIFF
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -21,6 +21,9 @@
 #include <sys/socket.h>
 #endif
 #ifdef HAVE_NET_IF_H
+#ifdef (__OpenBSD__)
+#include <sys/socket.h>
+#endif
 #include <net/if.h>
 #endif
 #ifdef HAVE_UNISTD_H
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -17,7 +17,7 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
-#if defined(HAVE_SYS_SOCKET_H) || defined(__OpenBSD__)
+#if defined(HAVE_SYS_SOCKET_H)
 #include <sys/socket.h>
 #endif
 #ifdef HAVE_NET_IF_H
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -17,7 +17,7 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
-#if defined(HAVE_SYS_SOCKET_H)
+#ifdef HAVE_SYS_SOCKET_H
 #include <sys/socket.h>
 #endif
 #ifdef HAVE_NET_IF_H
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -17,13 +17,10 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
-#ifdef HAVE_SYS_SOCKET_H
+#if defined(HAVE_SYS_SOCKET_H) || defined(__OpenBSD__)
 #include <sys/socket.h>
 #endif
 #ifdef HAVE_NET_IF_H
-#ifdef (__OpenBSD__)
-#include <sys/socket.h>
-#endif
 #include <net/if.h>
 #endif
 #ifdef HAVE_UNISTD_H
--- a/libgc/dyn_load.c
+++ b/libgc/dyn_load.c
@@ -396,7 +396,8 @@ GC_bool GC_register_main_static_data()
 
 # if (defined(LINUX) || defined (__GLIBC__) || defined(NACL)) /* Are others OK here, too? */ \
      && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ > 2) \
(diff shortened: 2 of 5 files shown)


mono/utils/networking-posix.c
```diff
@@ -21,6 +21,9 @@
 #include <sys/socket.h>
 #endif
 #ifdef HAVE_NET_IF_H
+#ifdef (__OpenBSD__)
+#include <sys/socket.h>
```
vargaz: There is already a sys/socket.h include above, isn't HAVE_SYS_SOCKET_H defined on openbsd ?

vargaz:
Looks good otherwise.

thfrwn:
HAVE_SYS_SOCKET_H is not defined in openbsd clang's preprocessor defines.

vargaz:
There is a AC_CHECK_HEADERS () check for it in configure.ac, why does that fail ?

thfrwn:
Hmm... AC_CHECK_HEADERS should take care of this - maybe that patch isn't needed. I'm doing a test build without that patch. Give me ~30 minutes.

thfrwn:
Okay, that took less time to find out. It doesn't build without the patch

`In file included from networking-posix.c:21:`
`/usr/include/net/if.h:121:17: error: field has incomplete type 'struct timeval'`

vargaz:
I mean why does the configure check fail to find sys/socket.h ?

thfrwn:
I honestly have no idea why AC_CHECK_HEADERS fails. It's not patched out or anything. currently building based on 5.10.0.179.

output from configure:

`checking for sys/socket.h... (cached) yes`

Noticed there's an error where it doesn't find `./libtool`

Full output of `make clean configure` here:

https://gist.github.com/thfrwn/db3544a215fbc190f1e8c71daecd15c1

jaykrell:
I don't mind this sort of thing -- "static conf" instead of "auto conf", but you might also look at config.log and config.h. configure --verbose might also help, might.

mono/utils/networking-posix.c
```diff
@@ -17,7 +17,7 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
-#ifdef HAVE_SYS_SOCKET_H
+#if defined(HAVE_SYS_SOCKET_H) || defined(__OpenBSD__)
```
jaykrell: This might just be a bad merge.
The earlier code did not include socket.h, just in.h and had an ifdef around that. The earlier path added another include within the ifdef.

mono/utils/networking-posix.c
jaykrell: Yes I'm pretty convinced this is just a bad merge. The older Mono code did not include sys/socket.h here (or nearby).

mono/utils/networking-posix.c
jaykrell: I'm pretty certain this is rooted in an bad merge, and is not needed.

vargaz: The configure output indicates that it does find sys/socket.h, is HAVE_SYS_SOCKET_H defined in config.h ?

akoeplinger:
The message about libtool not found should be harmless, filed https://github.com/mono/mono/issues/7705

mono/utils/networking-posix.c
thfrwn: HAVE_SYS_SOCKET_H is set to 1 in config.h

I looked through config.h, config.h.in, config.log, and the output of `./configure --verbose`, but failed to find anything that would explain it to me.

Here are the files if you wanna take a look:

https://gist.github.com/thfrwn/ec78cfbf680d5847e26610e64df085fe

vargaz: So HAVE_SYS_SOCKET_H is defined in config.h, and networking-posix.c includes that file in the beginning, so not sure what is happening here.

vargaz: Those lines were added by e31ecb3b4198.

thfrwn: I'm doing this with the 5.10.0.179 from https://download.mono-project.com/sources/mono/ - would be surprised if a Jan 8 commit wasn't in that?

akoeplinger: 5.10 was branched in December (2017-12 branch) so that does not surprise me :)

thfrwn: did it not enter the stable build?

akoeplinger: We do branch for release every two months so depending on when it was committed it is included or not. It did make the 2018-02 branch which will be Mono 5.12.

thfrwn: gotcha. I'll just take this patch out then

mono/utils/networking-posix.c
```diff
@@ -17,7 +17,7 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
-#ifdef HAVE_SYS_SOCKET_H
+#if defined(HAVE_SYS_SOCKET_H)
 #include <sys/socket.h>
```
vargaz: This change is not needed.

thfrwn: Absolutely. Missed this one.

vargaz:
@monojenkins build

alexischr:
@monojenkins squash

vargaz:
@monojenkins build

luhenry:

PR #7703, opened 2018-03-20
[Tailcall] Add interface tailcall tests, and generics thereof.

jaykrell:
These all fail on Mono. All but one succeeds on desktop.

### DIFF
--- a/mono/tests/tailcall-interface.cs
+++ b/mono/tests/tailcall-interface.cs
@@ -1,0 +1,517 @@
+/*
+Author:
+    Jay Krell (jaykrell@microsoft.com)
+
+Copyright 2018 Microsoft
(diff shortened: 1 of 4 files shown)


mono/tests/tailcall-interface.cs
```diff
@@ -0,0 +1,531 @@
...
+		//Console.WriteLine("done");
+		//Console.WriteLine("success");
+
+		print (c.cast2 (b));
+		print (c.cast3<B> (b));
+		print (c.cast3<B[]> (ba));
+		print (c.cast4 (ba));
+		print (c.cast5<B> (ba));
+
+		print (da.cast2 (b));
```
jaykrell: Not right -- either duplication or forgot to vary.

jaykrell: Should fixed now.

PR #7714, opened 2018-03-20
Remove obsolete code for getting TLS values from other threads.

### LINKED ISSUE #7692: Remove mono_mach_get_tls_address_from_thread / mono_mach_arch_get_tls_value_from_thread?
As mentioned in https://github.com/mono/mono/pull/7530 the two functions no longer seem to be used. Can the code be removed altogether?

filipnavara:
Fixes #7692, in response to pull request #7530.

### DIFF
--- a/mono/utils/mach-support-amd64.c
+++ b/mono/utils/mach-support-amd64.c
@@ -22,21 +22,6 @@
 //For reg numbers
 #include <mono/arch/amd64/amd64-codegen.h>
 
-/* Known offsets used for TLS storage*/
-
-/* All OSX versions up to 10.8 */
-#define TLS_VECTOR_OFFSET_CATS 0x60
-#define TLS_VECTOR_OFFSET_10_9 0xe0
-#define TLS_VECTOR_OFFSET_10_11 0x100
-
-/* This is 2 slots less than the known low */
-#define TLS_PROBE_LOW_WATERMARK 0x50
-/* This is 28 slots above the know high, which is more than the known high-low*/
-#define TLS_PROBE_HIGH_WATERMARK 0x200
-
-
-static int tls_vector_offset;
-
 void *
 mono_mach_arch_get_ip (thread_state_t state)
 {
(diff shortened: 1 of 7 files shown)


mono/utils/mach-support-x86.c
```diff
@@ -151,64 +135,9 @@ mono_mach_arch_set_thread_states (thread_port_t thread, thread_state_t state, ma
...
-}
-
-void *
-mono_mach_arch_get_tls_value_from_thread (pthread_t thread, guint32 key)
-{
-	return *(void**)mono_mach_get_tls_address_from_thread (thread, key);
-}
-
 void
 mono_mach_init (pthread_key_t key)
```
luhenry: Why not removing fully the function?

mono/utils/mach-support-arm64.c
```diff
@@ -148,58 +129,9 @@ mono_mach_arch_set_thread_states (thread_port_t thread, thread_state_t state, ma
...
-}
-
-void *
-mono_mach_arch_get_tls_value_from_thread (pthread_t thread, guint32 key)
-{
-	return *(void**)mono_mach_get_tls_address_from_thread (thread, key);
-}
-
 void
 mono_mach_init (pthread_key_t key)
```
luhenry: Why not removing fully the function?

mono/utils/mach-support-arm.c
```diff
@@ -136,58 +118,9 @@ mono_mach_arch_set_thread_states (thread_port_t thread, thread_state_t state, ma
...
-}
-
-void *
-mono_mach_arch_get_tls_value_from_thread (pthread_t thread, guint32 key)
-{
-	return *(void**)mono_mach_get_tls_address_from_thread (thread, key);
-}
-
 void
 mono_mach_init (pthread_key_t key)
```
luhenry: Why not removing fully the function?

mono/utils/mach-support-amd64.c
```diff
@@ -160,64 +145,9 @@ mono_mach_arch_set_thread_states (thread_port_t thread, thread_state_t state, ma
...
-}
-
-void *
-mono_mach_arch_get_tls_value_from_thread (pthread_t thread, guint32 key)
-{
-	return *(void**)mono_mach_get_tls_address_from_thread (thread, key);
-}
-
 void
 mono_mach_init (pthread_key_t key)
```
luhenry: Why not removing fully the function?

mono/utils/mach-support-arm.c
```diff
@@ -25,24 +25,6 @@
...
-
-/* This is 2 slots less than the known low */
-#define TLS_PROBE_LOW_WATERMARK 0x40
-/* This is 24 slots above the know high, which is the same diff as the knowns high-low*/
-#define TLS_PROBE_HIGH_WATERMARK 0x108
-
-static int tls_vector_offset;
-
 void *
 mono_mach_arch_get_ip (thread_state_t state)
```
luhenry: `mono_mach_arch_get_ip` and `mono_mach_arch_get_sp` seems dead too.

mono/utils/mach-support-x86.c
filipnavara: Well, I can... it just seemed like this one could have some use in the future.

mono/utils/mach-support-arm.c
filipnavara: I can remove them too if desired. Originally I just wanted to remove the TLS code and keep the patch small because it was dead code that was actively causing trouble. These functions are harmless compared to that.

luhenry: If the code is dead, then it's ok to remove it. It will always be part of the history so it's not completely lost anyway.

PR #6333, opened 2017-12-22
[object] mono_string_new should not assert on UTF conversion failures

lambdageek:
Revert the embedding API behavior change introduced by dcdfb3c5005297b375c985ec5f0a4af99f8ad7cf

`mono_string_new` will:
 * return `NULL` if the given byte sequence is not a valid UTF-8 sequence
 * assert if there is not enough memory to allocate a new MonoString.

This only changes the behavior of the API function.  The runtime will continue
to use `mono_string_new_checked` which sets `MonoError*` for both sorts of
failures.

PR #7718, opened 2018-03-20
Forward mono_string_new_wrapper to mono_string_new for error handling.

joncham:
In commit 8ba48b5a41ac48bf6fe9351406165d22c753722d
mono_string_new was adjusted to only assert for
out of memory rather than all errors to preserve
previous behavior for invalid strings. This
change preserves behavior for mono_string_new_wrapper.

This is similar to PR https://github.com/mono/mono/pull/6333

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -6358,15 +6358,8 @@ mono_string_new_len_checked (MonoDomain *domain, const char *text, guint length,
 	return o;
 }
 
-/**
- * mono_string_new:
- * \param text a pointer to a UTF-8 string
- * \deprecated Use \c mono_string_new_checked in new code.
- * This function asserts if it cannot allocate a new string.
- * \returns A newly created string object which contains \p text.
- */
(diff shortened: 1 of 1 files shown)


mono/metadata/object.c
```diff
@@ -6486,16 +6486,7 @@ mono_string_new_wrapper (const char *text)
...
-
-	if (text) {
-		ERROR_DECL (error);
-		MonoString *result = mono_string_new_checked (domain, text, error);
-		mono_error_assert_ok (error);
-		return result;
-	}
-
-	return NULL;
+	return mono_string_new (mono_domain_get (), text);
```
lambdageek: This should result in a compilation error about calling a `MONO_RT_EXTERNAL_ONLY` function.

joncham: Argh. I work on Windows where that doesn't error.

joncham:
Updated @lambdageek

PR #7719, opened 2018-03-20
[RFC][win32] Deprecation error if using MONO_RT_EXTERNAL_ONLY function

lambdageek:
We do this under clang and gcc on Linux and Mac; so might as well try Windows with MSC, also.

One caveat is that it turns _all_ deprecations warnings into errors.  That may not be what we want.  (Or else we may need `#pragma warning (suppress : 4996)` if we use old API anywhere else.)

I'm not a Windows person - looking for feedback.

### DIFF
--- a/mono/utils/mono-publib.h
+++ b/mono/utils/mono-publib.h
@@ -144,6 +144,12 @@ mono_set_allocator_vtable (MonoAllocatorVTable* vtable);
 #pragma GCC diagnostic error "-Wdeprecated-declarations"
 #endif
 
+#elif defined (_MSC_VER)
+#define MONO_RT_EXTERNAL_ONLY __declspec (deprecated ("The mono runtime must not call this function.")) \
+	MONO_RT_CENTRINEL_SUPPRESS
+// Turn deprecation warnings into errors
+#pragma warning (error : 4996)
+
 #else
 #define MONO_RT_EXTERNAL_ONLY MONO_RT_CENTRINEL_SUPPRESS
 #endif // clang or gcc


mono/utils/mono-publib.h
```diff
@@ -144,6 +144,12 @@ mono_set_allocator_vtable (MonoAllocatorVTable* vtable);
 #pragma GCC diagnostic error "-Wdeprecated-declarations"
 #endif
 
+#elif defined (_MSC_VER)
+#define MONO_RT_EXTERNAL_ONLY __declspec (deprecated ("The mono runtime must not call this function.")) \
```
jaykrell: The version that takes a message is "new". Old enough?

jaykrell: (just like gcc)

mono/utils/mono-publib.h
```diff
@@ -144,6 +144,12 @@ mono_set_allocator_vtable (MonoAllocatorVTable* vtable);
...
+	MONO_RT_CENTRINEL_SUPPRESS
+// Turn deprecation warnings into errors
+#pragma warning (error : 4996)
```
jaykrell: Same question as you -- what about all the other errors?
Our windows build is so full of warnings so keeping it as a warning would be worthless (which you aren't doing here, understood).

mono/utils/mono-publib.h
jaykrell: To be safe, if _MSC_VER greater than something. To be deliberately harsh with old compilers, not a bad thing, asis.

lambdageek: What's our lower bound compiler on Windows? Is it VS2015 or something older?

The oldest MSDN docs that mention deprecated with a message that I can see are for VS2008.

lambdageek:
Hm.
I see some 4996 warnings in the Windows x64 job output but they didn't kill the build for some reason...

```
 environment.c
..\mono\metadata\environment.c(59): warning C4996: 'GetVersionExW': was declared deprecated [d:\j\workspace\x\msvc\libmono-dynamic.vcxproj]
  C:\Program Files (x86)\Windows Kits\8.1\Include\um\sysinfoapi.h(442): note: see declaration of 'GetVersionExW'
```
and 
```
..\mono\metadata\threadpool-io.c(512): warning C4996: 'inet_addr': Use inet_pton() or InetPton() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings [d:\j\workspace\x\msvc\libmono-dynamic.vcxproj]
  C:\Program Files (x86)\Windows Kits\8.1\Include\um\winsock2.h(1850): note: see declaration of 'inet_addr'
```

I expected the pragma to be in scope.

mono/utils/mono-publib.h
lateralusX: Should be fine, in master we have upgraded the solution and vcxproj's to VS2015. In other places (like mkbundle) we have dropped support for VS2013 toolchain as well, so if this was introduced in VS2008 then just using _MCS_VER should be enough.

mono/utils/mono-publib.h
lateralusX: Most of the warnings are in common code, but other platforms are building with lower warning levels or excluding several of the categories causing most of the warnings in windows builds. One option could be to reduce warning level on windows to similar levels used in gcc/clang builds and then keeping this as a warning.

lateralusX:
@lambdageek Several of the vcxproj's disables some specific warnings like 4996 so maybe that interfere with the pragma above. One option could be to remove 4996 from disabled list in vcxproj and see what needs to be adjusted. Just did a quick build of libmono-dynamic vcxproj (complete mono runtime build as a dynamic library) and only two API's gets flagged with warning 4996 so should be easy to suppress local or just fix to none deprecated API's. When I removed 4996 from disable list and added:

`#pragma warning (error : 4996)`

to top of environment.c I got expected results:

```
1>------ Build started: Project: libmonoruntime, Configuration: Debug x64 ------
1>  environment.c
1>..\mono\metadata\environment.c(61): error C4996: 'GetVersionExW': was declared deprecated
1>  C:\Program Files (x86)\Windows Kits\8.1\Include\um\sysinfoapi.h(442): note: see declaration of 'GetVersionExW'
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

luhenry:
@monojenkins build OS X i386

luhenry:

jaykrell:
Note that msvc has no warning for missing prototype but implemented. How about just skipping the prototypes?

lambdageek:
@jaykrell I don't follow.  Skip the prototypes for the `MONO_RT_EXTERNAL_ONLY` functions you mean? So that they wouldn't be callable from another translation unit?  I guess that kind of gets us what we want in a roundabout way - provided we don't call an external only function from the same translation unit where it is defined.  (Not a good assumption, we have large .c files that call their own functions very often).

Last time I looked through the Windows builds' logs, there were only a handful (maybe 2?) of deprecation warnings - it seems more straightforward to just fix those (or locally disable 4996, as needed) and henceforth treat deprecation warnings as errors.

jaykrell:
For emphasis, agreed, that pragma warning(suppress:4996) right before lines that call deprecated functions is a good idea. You can still also skip the prototypes. These aren't contradictory proposals.

Also, for internal builds, non-Windows -- can also skip the prototype in the .h file, and duplicate it in .c right before the function. That will again fix  "2 out of 3" ways that mono time can do it wrong -- from other files or from same file but earlier. However it does involve duplication of the prototypes.

steveisok:
Since this is older, I'm closing.  Please see the [old pull request policy](https://github.com/mono/mono/wiki/Pull-Request-Guide#old-pull-request-policy) for more info.

Issue #7418, opened 2018-03-05
Figure out what it will take to build the BCL with the interpreter on CI


luhenry:
We want to build the BCL and the tests - on top of running the latter - with the interpreter on its CI lanes. We must try to use Roslyn, but should fallback to MCS if there are too many failures.

This would allow us:
 - Augmented trust of our CI: if we are able to run Roslyn or MCS to compile the BCL, this is a good sign that the interpreter is solid; it already allows us to uncovers JIT bugs
 - Easier testing for mixed-mode: by using the already AOTted core libraries and executables (mscorlib, System, roslyn/mcs, etc.) to compile the BCL with the interpreter, we would test mixed-mode.

First, we need to assess the situation and figure out what's broken and how much will be needed to fix it.

luhenry:
Compiling the BCL with MCS doesn't crash all the time, so it should be possible to use the interpreter to compile the BCL and tests, and run the tests.

marek-safar:
This adds almost no value. We are already running mcs tests suite which covers many more cases than running mcs for BCL.

We need to run csc compiled BCL tests with interpreter. That's significantly more useful target.

luhenry:
@marek-safar csc is a much bigger target to support than mcs, possibly putting it many month away from doable. mcs is the only solution over csc that would allow us to guarantee we do not regress the interpreter when compiling the BCL. Discarding it as "does not add value" is just pushing aside the existing context.

marek-safar:
I didn't say we have to run csc using the interpreter. I just don't see little value in using the interpreter to build BCL. What does it add us on top of what we have today? If you can run mcs test suite using the interpreter we can almost certainly build BCL using mcs. There are not special IL patterns for the different BCL libraries used in mcs

marek-safar:
I disagree, the mcs code is pretty self-contained and uses almost no BCL APIs it might use different IL patterns but they should all be hit by mcs tests. We might want to run mscorlib compilation using the interpreter but apart from that I'd focus elsewhere, there are many more tests we could run which would give us significantly better coverage in less time.

luhenry:
[...]

PR #7722, opened 2018-03-20
[ci] Compile BCL with interpreter

lewurm:
Fixes https://github.com/mono/mono/issues/7418


mcs/build/config-default.make
```diff
@@ -9,7 +9,11 @@
 # Use UTF-8 as the default encoding for compilers
 CODEPAGE = 65001
 
+ifdef TEST_WITH_INTERPRETER
+RUNTIME_FLAGS = --interpreter
+else
 RUNTIME_FLAGS =
+endif
```
akoeplinger: I'd move it to the TEST_WITH_INTERPRETER if clause we have already right below.

luhenry: +1

luhenry:
@lewurm why `do-not-merge`? I would love to see it happen sooner rather than later, and it won't impact any of our customers, just CI

BrzVlad:
@lewurm if it builds fine, I think it would be nice try it a few more times before merging,  just in case, since this would be a required lane.

akoeplinger:
@luhenry the `do-not-merge` label is fine, it shows that this is still experimental so someone (like me 😄) doesn't accidentally merge it

luhenry:

luhenry:

luhenry:

luhenry:

lewurm:
@marek-safar there is still an mcs issue on `sizeof-empty-structs.cs`:
```
sizeof-empty-structs.cs(87,28): error CS1929: Type `System.Collections.Generic.List<(string, int)>' does not contain a member `SequenceEqual' and the best extension method overload 
Compilation failed: 1 error(s), 0 warnings
```

luhenry:
Some progress: we now can run testinterp, and the list of failing tests are `bug-60848.exe`, `DebuggerTests.LineNumbers`, `DebuggerTests.SetIP, `DebuggerTests.SetIPSingleStep, `DebuggerTests.SingleStepping`. They all fail on all platforms so I would assume they are related to the change to mcs.

@marek-safar @kumpera if we want to even just avoid regressions with MCS we should run the tests at least with the JIT (on top of the build-only CI check we have nowadays).

marek-safar:
The debugger test needs mcs specific changes.

We could build runtime tests with mcs but the failures are from new tests so no regressions found and it'll block us from using any new C# syntax mcs does not support there

marek-safar:
>For the debugger tests, that means that we don't support the debugger with code generated by MCS in general then.

We do but as generated IL is different the debugger verification checks are different because the tests are quite low level

>For the runtime tests, we already have a status check for Linux x64 mcs which compiles all BCL, BCL tests and runtimes tests, and it's a required status check, meaning any breakage shouldn't happen.

Incorrect, we don't build runtime tests with mcs in `Linux x64 mcs`

>So the question bilds down to: how much do we care about MCS? Is it still a first-class compiler, in this case it should work just as well with MCS or CSC; or is it a second-class compiler?

mcs is a fallback compiler right now, no reason to fight for first-class vs second-class.

luhenry:

luhenry:
@monojenkins build Linux AArch64Interpreter

luhenry:

luhenry:

luhenry:

luhenry:

luhenry:

luhenry:

lewurm:

marek-safar:
@monojenkins build failed

luhenry:
@monojenkins build failed

Issue #6966, opened 2018-02-13
CoreFX System.IO.FileSystem.Watcher


marek-safar:

alexischr:
@marek-safar What do you mean by "Mobile profile version"? We have discussed adding the CoreFX INotify and FSEvent watchers to the monodroid and monotouch profiles, but CoreFX doesn't have a "generic" FSW. I thought our DefaultWatcher could work but you mentioned it's not very functional. Are we talking about making one?

marek-safar:
I though CoreFX has universal one, which does something and throws PNSE elsewhere. If that's not the case I fine to change the task item to XI/XM + XA only

AdamDotNet:
This looks promising for aspnet/Hosting/issues/1287! Will this get rid of the NotImplementedExceptions inside the FileSystemWatcher class on Mono.Android, seeing the "Mobile profiles" checklist item?

jamiewest:
I tried using some bits from Microsoft.Extensions.Hosting, a netstandard2.0 package, and I get a non implemented exception on a Xamarin.Forms app. Looking forward to this issue being incorporated.

PR #7740, opened 2018-03-21
[System.IO] Use CoreFX FileSystemWatcher stub for mobile profile

alexischr:
https://github.com/mono/mono/issues/6966


alexischr:
@monojenkins build failed

alexischr:
@monojenkins commit apidiff

PR #7744, opened 2018-03-21
[System.Windows.Forms] Reset check state of item when removing it.

### LINKED ISSUE #7664: CheckListBox maintains check state
## Steps to Reproduce

1. Fill CheckListBox with items: CheckListBox.Items.Add(item,false)
2. Check items in CheckListBox
3. Clear CheckListBox: CheckListBox.Items.Clear()
4. Fill CheckListBox with items: CheckListBox.Items.Add(item,false)
5. Check state from line 2 is maintained even though it was cleared and Item.Count is 0 before step 4.

## Current Behavior
Check state from line 2 is maintained even though it was cleared, items were added expressly setting
check state to false and Item.Count is 0 before step 4.

## Expected Behavior
Items added in step 4 should be unchecked.

[ ] macOS
[ X] Linux (Fedora 27)
[ ] Windows

Same code under Windows 10, VS2017 produces the correct results.

**Version Used**:

Mono version: 5.10.0.160

### Stacktrace
 NA

filipnavara:
Fixes #7664.


mcs/class/System.Windows.Forms/System.Windows.Forms/CheckedListBox.cs
```diff
@@ -423,7 +423,18 @@ public int Add (object item, CheckState check)
...
+
+			public override void Clear ()
+			{
+				owner.check_states.Clear ();
+				base.Clear ();
+			}
+
+			internal override void UpdateSelection (int removed_index)
+			{
+				owner.check_states.Remove (this[removed_index]);
```
akoeplinger: don't you need to call `base.UpdateSelection (removed_index)` ?

filipnavara: Eh, of course. I was sure I pushed the version where the line was present.

filipnavara: Weird. One `git push --force` later and it's there.

akoeplinger:
@monojenkins commit apidiff

PR #7747, opened 2018-03-21
[interp] Allocate ThreadContext structures on the heap instead of on the stack, the latter doesn't work in mixed mode if the native frame gets unwound during exception handling, so the set_context (NULL) call is never executed, leaving a dangling ref in the tls variable.

vargaz:


mono/mini/interp/interp.c
```diff
@@ -1597,8 +1591,7 @@ interp_entry (InterpEntryData *data)
 
 	frame.ex = NULL;
 	if (context == NULL) {
-		context = &context_struct;
-		memset (context, 0, sizeof (ThreadContext));
+		context = g_new0 (ThreadContext, 1);
 		set_context (context);
 	} else {
 		old_frame = context->current_frame;
```
lewurm: oops, shouldn't that be outside of the else? otherwise `old_frame` might isn't assigned later in that function

PR #7748, opened 2018-03-21
[metadata] Move mono_class_compute_gc_descriptor back to object.c

lambdageek:
Partially revert 2713dfc,
but preserve 523ba15


lambdageek:
@monojenkins build failed

Issue #6858, opened 2018-02-07
Convert existing per profile assembly csproj files into single csproj


marek-safar:

akoeplinger:
This should be done with https://github.com/mono/mono/pull/7755.

PR #7755, opened 2018-03-22
Genproj updates for BCL build in Windows Visual Studio

kg:
This batch of commits makes further updates to our support for building the BCL using msbuild (Visual Studio, to be specific).

* Generates a single sln and set of csproj files with support for all of our build profiles (orbis, monodroid, net_4_x, etc)
* Project configuration selector (in VS or as parameter to msbuild) can be used to select which profile to build
* Includes partial dependency information so that culevel and jay are compiled before the rest of the sln (this keeps pre/post events from from failing on first build)

As of this rev things are working pretty good, but a lot of the measures I took here are temporary so I'd appreciate thoughts on how to go about them. The hard-coded System.Web -> Culevel dependency feels gross, for example - I kind of want to go through and build an automatic solution for it but I'm not sure it's necessary.

Implements #6858

### DIFF
--- a/mcs/tools/Makefile
+++ b/mcs/tools/Makefile
@@ -10,7 +10,6 @@ net_4_5_dirs := \
 	mono-service	\
 	mono-xsd	\
 	resgen		\
-	gacutil		\
 	wsdl		\
 	xbuild		\
 	csharp		\
--- a/msvc/scripts/csproj.tmpl
+++ b/msvc/scripts/csproj.tmpl
@@ -25,7 +25,7 @@
     <RootNamespace>
     </RootNamespace>
     <AssemblyName>@ASSEMBLYNAME@</AssemblyName>
-    <TargetFrameworkVersion>v@FX_VERSION</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v@FX_VERSION@</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
 @SIGNATURE@
   </PropertyGroup>
@@ -53,22 +53,14 @@
     <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
   </PropertyGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <ItemGroup>
-@SOURCES@  </ItemGroup>
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
-       Other similar extension points exist, see Microsoft.Common.targets.
-  <Target Name="BeforeBuild">
-  </Target>
-  <Target Name="AfterBuild">
-  </Target>
-  -->
+  <!-- @ALL_SOURCES@ -->
+  <!-- @ALL_REFERENCES@ -->
   <PropertyGroup>
 @PREBUILD@
 @POSTBUILD@
   </PropertyGroup>
   <ItemGroup>
-@REFERENCES@  </ItemGroup>
-  <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
-@RESOURCES@</Project>
+  <!-- @ALL_RESOURCES@ -->
+</Project>
--- a/msvc/scripts/csproj.tmpl
+++ b/msvc/scripts/csproj.tmpl
@@ -3,7 +3,6 @@
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <ProductVersion>9.0.30729</ProductVersion>
     <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>@PROJECTGUID@</ProjectGuid>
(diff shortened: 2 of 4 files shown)


mcs/tools/Makefile
```diff
@@ -10,7 +10,6 @@ net_4_5_dirs := \
 	mono-service	\
 	mono-xsd	\
 	resgen		\
-	gacutil		\
```
marek-safar: Why was it removed

msvc/scripts/genproj.cs
```diff
@@ -71,18 +72,23 @@ public void Add (MsbuildGenerator.VsCsproj vsproj)
...
 	{
+		// HACK
+		library = library.Replace("-net_4_x", "");
 		sln.WriteLine (project_start, prefixGuid, library, relativePath, projectGuid);
 
-		foreach (var guid in dependencyGuids) {
-    		sln.WriteLine ("    ProjectSection(ProjectDependencies) = postProject");
-    		sln.WriteLine ("        {0} = {0}", guid);
-    		sln.WriteLine ("    EndProjectSection");
+		if (
```
marek-safar: The extra parens are not necessary

msvc/scripts/genproj.cs
```diff
@@ -71,18 +72,23 @@ public void Add (MsbuildGenerator.VsCsproj vsproj)
...
+			(dependencyGuids != null) && (dependencyGuids.Length > 0)
+		) {
+			sln.WriteLine ("\tProjectSection(ProjectDependencies) = postProject");
+			foreach (var guid in dependencyGuids)
+	    		sln.WriteLine ("\t\t{0} = {0}", guid);
```
marek-safar: This looks suspicious, why is the same thing assign from and to?

mcs/tools/Makefile
kg: We build it twice, which causes order.xml to have bad information in it

msvc/scripts/genproj.cs
kg: I don't know, .sln is a really bad mysterious format. I discovered using spaces instead of tabs causes silent parsing errors :(

marek-safar:
For the hard coded dependencies. Did you try to add the projects as normal references that should ensure that they are built before and then adjust their execution path to point to temporary obj/ location?

kg:
We currently have no way to feed dependencies through, both of these are EXE files so I'm not sure how to make a reference work right... the compiled dll would depend on it to load wouldn't it? It would also copy the exe into the build output directory, I think.

culevel is managed so maybe the ref would work. jay is native so i'm not sure if it would work at all. I can test it if you think it's worth doing

marek-safar:
It seems @kzu found a solution for that http://www.cazzulino.com/project-dependencies-as-project-references.html

mcs/tools/Makefile
akoeplinger: yep we already have it in `net_4_x_SUBDIRS = gacutil` so this is fine.

msvc/scripts/csproj.tmpl
```diff
@@ -13,62 +12,60 @@
...
+  <PropertyGroup>
+    <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
+  </PropertyGroup>
+
+<!-- @ALL_PROFILE_PROPERTIES@ -->
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <NoWarn>@DISABLEDWARNINGS@</NoWarn>
```
akoeplinger: this can be removed here since we already set it in the global PropertyGroup

msvc/scripts/csproj.tmpl
```diff
@@ -13,62 +12,60 @@
...
-    <DefineConstants>TRACE;@DEFINECONSTANTS@</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <!-- TRACE is set only for Debug configuration, so inherit from platform-specific value -->
+    <DefineConstants>TRACE;$(DefineConstants)</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <DebugType>pdbonly</DebugType>
     <NoWarn>@DISABLEDWARNINGS@</NoWarn>
```
akoeplinger: same

msvc/scripts/csproj.tmpl
```diff
@@ -13,62 +12,60 @@
...
     <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>
-    </RootNamespace>
+    <RootNamespace></RootNamespace>
     <AssemblyName>@ASSEMBLYNAME@</AssemblyName>
-    <TargetFrameworkVersion>v@FX_VERSION</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v@FX_VERSION@</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
-@SIGNATURE@
+    @SIGNATURE@
```
akoeplinger: this shouldn't be indented, the variable that gets inserted here already has spaces in it...

kg:
Ludovic pointed out that because this changes some of the csproj filenames, the old ones will stick around. Does it make sense for genproj to delete the old ones or do we just want to do that once by hand?

akoeplinger:
We'll do deleting in a separate PR.

PR #7757, opened 2018-03-22
[interp] always set old_frame in interp_entry

lewurm:
fixup for https://github.com/mono/mono/pull/7747


lewurm:
@monojenkins rebase

PR #4654, opened 2017-04-06
[runtime] Make mono_set_pending_exception public.

rolfbjarne:
Xamarin.Mac uses mono_set_pending_exception to convert Objective-C exceptions
to managed exceptions.

PR #7759, opened 2018-03-22
[runtime] Add public API to set the pending exception.

rolfbjarne:
Xamarin.iOS and Xamarin.Mac need a public API to set the pending exception
when converting Objective-C exceptions to managed exceptions.

This adds a public API to support this scenario.

Partially fixes https://bugzilla.xamarin.com/show_bug.cgi?id=60065.

### DIFF
--- a/mono/metadata/object.h
+++ b/mono/metadata/object.h
@@ -248,6 +248,10 @@ MONO_API void
 mono_raise_exception	    (MonoException *ex);
 
 MONO_RT_EXTERNAL_ONLY
+MONO_API mono_bool
+mono_runtime_set_pending_exception (MonoException *exc, mono_bool overwrite);
+
+MONO_RT_EXTERNAL_ONLY
 MONO_API void
 mono_reraise_exception	    (MonoException *ex);
 
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -4814,7 +4814,9 @@ mono_set_pending_exception (MonoException *exc)
  *
  *   Set the pending exception of the current thread to \p exc.
  *   The exception will be thrown when execution returns to managed code.
- *   Can optionally \p overwrite any existing pending exceptions.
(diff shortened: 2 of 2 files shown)


rolfbjarne:
This is a continuation of PR #4654.

kumpera:
I'm not sure allowing embedders to overwrite a pending exception is a good idea.
I believe it would interfere with the abort machinery.

rolfbjarne:
I can change/remove it if you wish (afaik we don't need it).

luhenry:
Wouldn't it being marked `MONO_API` that you then load dynamically fix your issue?

migueldeicaza:
The code does not override a pending exception in the current patch.

That said, there is value in this capability, I think we could document any downsides as a warning for embedders about potential side effects.

rolfbjarne:
@kumpera when you mean that setting the pending exception would interfere with the "abort machinery", do you mean just Thread.Abort? Or is it something else?

@luhenry I first tried marking it `MONO_API` (PR #4654), but that was rejected.

rolfbjarne:
@kumpera I can document that overwriting a pending thread abort exception will result in undefined behavior, would that be good enough?

kumpera:
@rolfbjarne why are we settling for the worse option of all? Can we either handle the scenario or remove the override argument?

rolfbjarne:
@kumpera I've documented that overwriting pending exceptions when the runtime is processing thread aborts is a bad idea. Is there anything else this PR needs?

kumpera:
No, it's more than ready for merging. Now we just need to go over the failures and ensure they are unrelated.

kumpera:
OSX 32bits failure is the dns hang in System.Net.Http.
Linux aarch64 failures in corlib xunit are unrelated. This PR doesn't change managed code behavior.

rolfbjarne:
Can it be back ported to 2018-02 and later branches?

luhenry:

luhenry:

Issue #7696, opened 2018-03-19
LINQ bug (ConcatNIterator.ToArray()?)


EgorBo:
## Steps to Reproduce
```csharp
using System;
using System.Collections.Generic;
using System.Linq;

public static class Program
{
    public static void Main()
    {
        var slots = GetSlots().Concat(new[] { new Slot("q") });
        var flatten = slots.Flatten(container => container.AllParentSlots)/*.Select(i => i)*/.ToArray();

        var str = string.Join(", ", flatten.Select(x => x == null ? "NULL" : x.Id));
        Console.WriteLine(str);
        // Mono 5.x including master:
        //   "a, b, c, d, e, f, g, q, x, NULL"
        // .NET, .NET Core, Mono 4.2:
        //   "a, b, c, d, e, f, g, q, x, y"
    }
    
    public class Slot
    {
        public readonly string Id;
        public Slot(string id) => Id = id;
        public SlotsHolder ContainedSlotsHolder;
        public IEnumerable<Slot> AllParentSlots => ContainedSlotsHolder == null ? new Slot[0] : ContainedSlotsHolder.Slots;
    }

    public class SlotsHolder
    {
        public Slot[] Slots;
    }
    
    static IEnumerable<T> Flatten<T>(this IEnumerable<T> e, Func<T, IEnumerable<T>> f) 
        => e.Concat(e.SelectMany(c => f(c).Flatten(f)));

    static IEnumerable<Slot> GetSlots()
    {
        yield return new Slot("a");
        yield return new Slot("b");
        yield return new Slot("c");
        yield return new Slot("d");
        yield return new Slot("e");
        yield return new Slot("f");
        yield return new Slot("g") {
            ContainedSlotsHolder = new SlotsHolder { Slots = new[] {
                new Slot("x"),
                new Slot("y")
            }}};
    }
}
```

It feels like the issue is somewhere in `ConcatNIterator.ToLazyArray`

## Current Behavior

The code prints `a, b, c, d, e, f, g, q, x, NULL`

## Expected Behavior

The code should print `a, b, c, d, e, f, g, q, x, y` just like .NET/.NET Core do.

### On which platforms did you notice this

 [ ] macOS
 [ ] Linux
 [x] Windows

**Version Used**:
mono 5.0-5.13
Does not reproduce on old versions of mono when we had a "slow" `Enumerable.ToArray` implementation.

mderoy:
This is affecting Unity users as well:
[...]

PR #7760, opened 2018-03-22
Bump corefx to bring a fix for #7696

marek-safar:


marek-safar:
@monojenkins backport 2018-02

PR #7764, opened 2018-03-22
[interp] Improve mixed mode execution support, using full-aot+interp as the configuration.

vargaz:
The assemblies need to be AOTed using:
--aot=full,interp
and run using:
--full-aot --interp


lewurm:
what's the equivalent for the "old" `--aot=interp` now? We need some way to "AOT compile" `mscorlib.dll` without all the methods included for the interpreter-only mode on iOS.

luhenry:
@vargaz please submit a follow up PR to update the man page as requested.

vargaz:
There is no needed, the semantics of the option haven't changed in the final version.

PR #7766, opened 2018-03-22
DO NOT MERGE! PR for a build with lock tracker enabled

Therzok:
To investigate 2s of lock contention on domain lock on VSMac with hybrid AOT

### DIFF
--- a/mono/metadata/lock-tracer.c
+++ b/mono/metadata/lock-tracer.c
@@ -22,6 +22,7 @@
 #endif
 
 #include <mono/utils/mono-compiler.h>
+#include <mono/utils/mono-threads.h>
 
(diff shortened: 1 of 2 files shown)


Therzok:
@monojenkins buildpkg

Therzok:
@monojenkins build pkg

Therzok:
Is there any way to stop other builds from happening? Only need the osx pkg

Therzok:
@monojenkins build pkg

akoeplinger:
> Is there any way to stop other builds from happening? Only need the osx pkg

Yes, add `@monojenkins skip` to the PR description and the default builds won't be triggered.

PR #7767, opened 2018-03-22
[runtime] Fix lock tracer compilation error

lambdageek:


lambdageek:
Test failures are unrelated

PR #7773, opened 2018-03-22
[mini] Use MonoClass getters in aot-compiler.c

lambdageek:


lambdageek:
@monojenkins build failed

PR #7779, opened 2018-03-23
[icall] remove DISABLE_PROCESS_HANDLING

lewurm:
The only user was Xamarin.iOS in a weird way: The define is only set for
device builds where the icall table is excluded anyway. For the cross
compiler however, the define isn't set, and therefore those icalls can
be resolved by the AOT compiler.

With the interpreter now, this define actually had an effect: The
guarded icalls weren't inclueded in `libmono-icall-table.a`.


lewurm:
needs `2018-02` backport

lewurm:
@monojenkins backport 2018-02

PR #7784, opened 2018-03-23
Remove incorrect docs

migueldeicaza:
This documentation has not been updated, despite the fact that the capabilities were implemented in the TermInfo driver.


marek-safar:
@monojenkins commit apidiff

PR #7797, opened 2018-03-23
[System] Makes TLS providers linker friendly

marek-safar:


joncham:
Can you elaborate a bit on this change? Is the purpose of the `typeof` usage to preserve those types?

marek-safar:
@joncham yes and also make it compilation error if they are not present

marek-safar:
@monojenkins backport 2018-02

PR #7799, opened 2018-03-23
[mini] Use MonoClass getters on non-AMD64 platforms

lambdageek:
Best effort to replace direct MonoClass field access by getters.


lambdageek:
@monojenkins build failed

PR #7802, opened 2018-03-23
[ci] For checked builds, pass -k to make

lambdageek:
I'd like to see more compilations errors in checked build results - so run with `make -k`


lambdageek:
@monojenkins build Linux x64 Checked Build

lambdageek:
Looks decent enough.

PR #7803, opened 2018-03-23
[sre-save] Remove type_token assignment in mono_image_fill_export_table

lambdageek:
In 9725d50c8726b9f34e897737176c64a94865ba0a we added the `g_error`, that should
fire if the `MonoClass:type_token` and the `TypeBuilder` token ever disagree. As a
result, this assignment is strictly redundant.


lambdageek:
@monojenkins build Linux x64

PR #7835, opened 2018-03-26
[WCF] fix index of message body parts

hgy59:
Do not use MessageBodyMemberAttribute.Order as index.

When working with Order attribute of MessageBodyMember in MessageContracts, the deserialization of messages often fails with array index out of bounds exception.
WCF services with such message contracts run with dotnet for windows but fail with mono.

Debugging the System.ServiceModel.Dispatcher.DataContractMessagesFormatter method MessageToParts showed the source of the exception: part.Index is used as index to an array of all parts. Therefor the index is expected to be 0 to (number of parts -1).
Since the collection in the message body is already ordered by the name and order of the parts, the solution is to adjust the indexes after all parts are added to the collection.


marek-safar:
approve

PR #7836, opened 2018-03-26
[runtime] One more MonoClass getter fixup

lambdageek:
- Fixup recent ilgen change
- Fix linking error in *monograph* due to static linking an extra copy of metadata.
- Fix runtime crash in `monograph`, too, while I'm at it.


lambdageek:
@monojenkins build failed

lambdageek:

😦

Issue #7856, opened 2018-03-27
[RFC]  mono_class_init_ready


lambdageek:
## Summary of design

We would like to add an overall notion of "`MonoClass` readiness" to replace the various `MonoClass:*_inited` bitfields.  This PR does **not** do that.  But it adds the corresponding
change to the loading API:

```c
typedef enum _MonoClassReady {
	/* Barebones readiness: just allocate a MonoClass and set its image, name and name space. */
	MONO_CLASS_READY_BAREBONES = 0,
	/* Instantiate readiness: metadata vtable is initialized, ready to create runtime vtables. */
	MONO_CLASS_READY_INSTANTIATE = 0xff,
	MONO_CLASS_READY_MIN = MONO_CLASS_READY_BAREBONES,
	MONO_CLASS_READY_MAX = MONO_CLASS_READY_INSTANTIATE,
} MonoClassReady;

gboolean
mono_class_init_ready (MonoClass *klass, MonoClassReady readiness);
```

1. We would eventually add multiple readiness levels between `MIN` and `MAX` that represent a `MonoClass` that is partially initialized.  (I don't know precisely how many we will need, so the examples below should not be taken too seriously.) For example:

   - Some information about a class is collected from its parent.  For example `MonoClass:delegate` is set based on the value of the parent.  So we may have `MONO_CLASS_READY_PARENT` above `MONO_CLASS_READY_BAREBONES` to signal that the parent class is ready to provide that information.  (This is in `mono_class_setup_parent` today and its protected by the loader lock and the `gclass_recorded_list` in the generic instance case).

   - Instance size information is currently guarded by `MonoClass:size_inited`.  We would add a 
 `MONO_CLASS_READY_INSTANCE_SIZE` above `MONO_CLASS_READY_PARENT` and below `MONO_CLASS_READY_INSTANTIATE`.

2. The semantics of `mono_class_init_ready (klass, level)` is that it is a no-op if `klass` is already
ready at level `level` or higher, otherwise it will perform whatever initialization is needed to bring the class from its current level up to at least the requested level.

3. Eventually the various legacy initialization functions (see `mono/metadata/class-init.h` - e.g. `mono_class_init_sizes`, `mono_class_layout_fields`, etc) will be removed (or marked `MONO_RT_EXTERNAL_ONLY`) and there will be a single initialization entrypoint.

[...]

PR #7857, opened 2018-03-27
Define mono_class_init_ready and MIN and MAX readiness levels

lambdageek:
# FOR REFERENCE ONLY

Please comment on the [RFC and design discussion](https://github.com/mono/mono/issues/7856)

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -4097,6 +4097,12 @@ mono_get_unique_iid (MonoClass *klass)
 gboolean
 mono_class_init (MonoClass *klass)
 {
+	return mono_class_init_ready (klass, MONO_CLASS_READY_MAX);
+}
+
+gboolean
+mono_class_init_ready (MonoClass *klass, MonoClassReady readiness)
+{
 	int i, vtable_size = 0, array_method_count = 0;
 	MonoCachedClassInfo cached_info;
 	gboolean has_cached_info;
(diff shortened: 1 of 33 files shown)


mono/metadata/class-init.c
```diff
@@ -4107,6 +4113,17 @@ mono_class_init (MonoClass *klass)
...
 
+	g_assert (readiness >= MONO_CLASS_READY_MIN && readiness <= MONO_CLASS_READY_MAX);
+
+	switch (readiness) {
+	case MONO_CLASS_READY_BAREBONES:
+		return !mono_class_has_failure (klass);
+	case MONO_CLASS_READY_INSTANTIATE:
+		break; /* continue with initialization */
+	default:
+		g_assert_not_reached ();
```
luhenry: a `g_error` with the value passed would help debugging.

kumpera:
LGTM, should this cover the many lazy init steps in the _setup_ functions?

One thing, I believe the comment in MONO_CLASS_READY_INSTANTIATE is incorrect as mono_class_init AFAICT doesn't do vtable setup.

lambdageek:
@kumpera 1. yes this should eventually take over from the `_setup_` functions.  2. thanks - looks like `mono_class_init` only sets up the vtable for generic instances. I'll rename. 3. We have the overall design discussion over at #7856

steveisok:
Since this is older, I'm closing.  Please see the [old pull request policy](https://github.com/mono/mono/wiki/Pull-Request-Guide#old-pull-request-policy) for more info.

PR #7863, opened 2018-03-28
Add BindingFlags.DoNotWrapExceptions

AustinWise:
The original design review is here: dotnet/corefx#22866 . To summarize, this adds a new flag to `BindingFlags` called `DoNotWrapExceptions`. When this flag is used to invoke a method using reflection, any exceptions thrown are not wrapped with a `TargetInvocationException`.

It has already been implemented in some other version of .NET:
- dotnet/corert#4437
- dotnet/coreclr#13767

I would be delighted if this handy feature was also available in Mono.

I have in part based my changes on the CoreCLR implementation. There all already tests for this feature in corefx, so I have added those to the corlib_xtest.dll.sources file.

I have a couple of concerns about my implementation:
- I notice that `DynamicMethod.Invoke` was ignoring the BindingFlags and other arguments. I changed this method to pass along all the arguments. For what it's worth, it appears that [CoreCLR respects](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/DynamicMethod.cs#L488) these arguments.

### DIFF
--- a/mcs/class/referencesource/mscorlib/system/activator.cs
+++ b/mcs/class/referencesource/mscorlib/system/activator.cs
@@ -184,9 +184,14 @@ namespace System {
                                   null,
                                   ref stackMark);
         }
-            
+
+        public static object CreateInstance(Type type, bool nonPublic)
+        {
+            return CreateInstance(type, nonPublic, wrapExceptions: true);
+        }
+
         [MethodImplAttribute(MethodImplOptions.NoInlining)] // Methods containing StackCrawlMark local var has to be marked non-inlineable
-        static public Object CreateInstance(Type type, bool nonPublic)
+        internal static object CreateInstance(Type type, bool nonPublic, bool wrapExceptions)
         {
             if ((object)type == null)
                 throw new ArgumentNullException("type");
@@ -198,7 +203,7 @@ namespace System {
                 throw new ArgumentException(Environment.GetResourceString("Arg_MustBeType"), "type");
 
             StackCrawlMark stackMark = StackCrawlMark.LookForMyCaller;
-            return rt.CreateInstanceDefaultCtor(!nonPublic, false, true, ref stackMark);
+            return rt.CreateInstanceDefaultCtor(!nonPublic, false, true, wrapExceptions, ref stackMark);
         }
 
         [MethodImplAttribute(MethodImplOptions.NoInlining)] // Methods containing StackCrawlMark local var has to be marked non-inlineable
@@ -228,7 +233,7 @@ namespace System {
                 return (T)rt.CreateInstanceSlow(true /*publicOnly*/, true /*skipCheckThis*/, false /*fillCache*/, ref stackMark);
             else
 #endif // FEATURE_CORECLR
-            return (T)rt.CreateInstanceDefaultCtor(true /*publicOnly*/, true /*skipCheckThis*/, true /*fillCache*/, ref stackMark);
+            return (T)rt.CreateInstanceDefaultCtor(true /*publicOnly*/, true /*skipCheckThis*/, true /*fillCache*/, true /*wrapExceptions*/, ref stackMark);
         }
 
         [ResourceExposure(ResourceScope.Machine)]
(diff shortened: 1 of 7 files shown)


vargaz:
Support for filter clauses is not perfect, it would be better to simply duplicate the body:
if (do-not-wrap) {
     <call>
} else {
   try {
     <call>
   } catch (...) {
   ...
  }
}

AustinWise:
I have update my pull request to not use exception filter clauses.

marek-safar:
approve

mcs/class/referencesource/mscorlib/system/reflection/bindingflags.cs
```diff
@@ -64,5 +64,6 @@ public enum BindingFlags
 
         // These are a couple of misc attributes used
         IgnoreReturn        = 0x01000000,    // This is used in COM Interop
+        DoNotWrapExceptions = 0x02000000, // Disables wrapping exceptions in TargetInvocationException
```
marek-safar: Please instead of changing this file replace it in `corlib.dll.sources` with the version from `external/corefx/src/Common/src/CoreLib/System/Reflection/BindingFlags.cs`

mcs/class/referencesource/mscorlib/system/activator.cs
```diff
@@ -184,9 +184,14 @@ static public Object CreateInstance(Type type)
...
-            
+
+        public static object CreateInstance(Type type, bool nonPublic)
+        {
+            return CreateInstance(type, nonPublic, wrapExceptions: true);
+        }
+
         [MethodImplAttribute(MethodImplOptions.NoInlining)] // Methods containing StackCrawlMark local var has to be marked non-inlineable
-        static public Object CreateInstance(Type type, bool nonPublic)
+        internal static object CreateInstance(Type type, bool nonPublic, bool wrapExceptions)
```
marek-safar: Should be just private

mcs/class/referencesource/mscorlib/system/reflection/bindingflags.cs
AustinWise: Done

marek-safar:
approve

mcs/class/referencesource/mscorlib/system/activator.cs
AustinWise: Before this change RuntimeType called `Activator.CreateInstance(Type type, bool nonPublic)` when a constructor has no arguments. After this change I needed a way to propagate the DoNotWrapExceptions flag, so I added this internal overload.

AustinWise:
I noticed that API Diff check failed because of the addition of the new flag. Is that something I should fix in this pull request?

marek-safar:
We'll take care of apidiff handling

mcs/class/referencesource/mscorlib/system/activator.cs
marek-safar: As it's used by CreateInstance only, it does have to be `internal`

mcs/class/corlib/corlib-net_4_x.csproj
```diff
@@ -68,6 +68,7 @@
     <Compile Include="..\..\..\external\corefx\src\Common\src\CoreLib\System\Memory.cs" />
     <Compile Include="..\..\..\external\corefx\src\Common\src\CoreLib\System\MemoryDebugView.cs" />
     <Compile Include="..\..\..\external\corefx\src\Common\src\CoreLib\System\ReadOnlyMemory.cs" />
+    <Compile Include="..\..\..\external\corefx\src\Common\src\CoreLib\System\Reflection\BindingFlags.cs" />
```
luhenry: You should do these 2 modifications to https://github.com/mono/mono/blob/master/mcs/class/corlib/corlib.dll.sources instead. This `corlib-net_4_x.csproj` is autogenerated from the `mcs/class/corlib/*.sources` files

AustinWise: I have removed my changes from the generated csproj file.

AustinWise:
@luhenry  It appears that a bot periodically regenerates the csproj file so I have removed my changes to that file. Is that what you had in mind?

marek-safar:
@monojenkins commit apidiff

luhenry:
@AustinWise yes we do regenerate periodically the csproj files from the `*.dll.sources` files, so modifying the former will just be overridden by the automatic generation, while modifying the latter will make it happen in the former.

Issue #6991, opened 2018-02-13
Epic: Roslyn tests


marek-safar:
We need to resolve bugs and tests problems in all Roslyn unit tests assemblies as they are important for us to run.

## Milestone I
- [x] Roslyn.Compilers.CSharp.Syntax.UnitTests
- [ ] Roslyn.Compilers.CSharp.Semantic.UnitTests
- [ ] Roslyn.Compilers.CSharp.Symbol.UnitTests
- [ ] Roslyn.Compilers.CSharp.Emit.UnitTests
- [ ] Microsoft.CodeAnalysis.CSharp.Scripting.UnitTests

The list comes from 
https://github.com/mono/roslyn/blob/mono-testing/mono-testing.sh

## Milestone II

There are additional tests we could run as well as we need to figure out how keep the tests repo synchronized with bundled csc we test against

jaredpar:
> https://github.com/mono/roslyn/blob/mono-testing/mono-testing.sh

Why have a custom script here versus using `./build.sh --test --mono`. The latter will run the tests as we're expecting using mono from the command line. 

The script in the current form will get broken from time to time as our build layout and TFs change. It's already broken if you sync up with dotnet/master as our tests changed from net461 to net46 and all of our test assemblies got renamed. 

What problems are you all having getting MS.CS.Symbols.UnitTests running? Those should have worked on Mono for some time now.

jaredpar:
Here is the issue we are tracking on our side to do this work. 

https://github.com/dotnet/roslyn/issues/28011

marek-safar:
@jaredpar I am happy to switch to `./build.sh --test --mono` if it does the same thing but it was not working as expected when we played with it.

Last time we ran `Roslyn.Compilers.CSharp.Semantic.UnitTests.dll` we had about 59 failures but don't have that list at hand but we are going to re-run it next week.

jaredpar:
@marek-safar 

> I am happy to switch to ./build.sh --test --mono if it does the same thing but it was not working as expected when we played with it.

This is what we've been investing in. If you hit issues with it, or it's not meeting your needs, please let us know and we can work it out. 

The intent with that switch is to essentially "run all test using the `mono` available on `$PATH`". Our CI is slightly different in that it creates [a docker container](https://github.com/dotnet/roslyn/blob/master/build/scripts/docker/Dockerfile), installs the latest stable mono package and then runs tests. 

[...]

PR #7866, opened 2018-03-28
Implement Field SetValueDirect on non-struct targets. Add test case from roslyn issue 19038.

kg:
A test case from Roslyn issue 19308 uses SetValueDirect to modify a field on a class. Mono currently only implements SetValueDirect for fields of structs. Adding class support makes the test pass as expected.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -2163,10 +2163,17 @@ leave:
 		mono_gchandle_free (value_gchandle);
 }
 
+ICALL_EXPORT MonoObject*
+mono_TypedReference_ToObject (MonoTypedRef* tref);
+
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
+	ERROR_DECL (error);
+
 	MonoClassField *f;
+	MonoObjectHandle objHandle, valueHandle;
+	MonoReflectionFieldHandle fieldHandle;
 
 	g_assert (field);
 	g_assert (obj);
@@ -2174,7 +2181,11 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, 
 
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
+		fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		objHandle = MONO_HANDLE_NEW (MonoObject, mono_TypedReference_ToObject (obj));
+		valueHandle = MONO_HANDLE_NEW (MonoObject, value);
+		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
+		mono_error_set_pending_exception (error);
 		return;
 	}
 
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -2169,6 +2169,8 @@ mono_TypedReference_ToObject (MonoTypedRef* tref);
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
+	HANDLE_FUNCTION_ENTER ();
+
 	ERROR_DECL (error);
 
 	MonoClassField *f;
@@ -2186,13 +2188,14 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, 
 		valueHandle = MONO_HANDLE_NEW (MonoObject, value);
 		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
 		mono_error_set_pending_exception (error);
-		return;
+	} else {
+		if (MONO_TYPE_IS_REFERENCE (f->type))
+			mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), value, FALSE);
+		else
+			mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), mono_object_unbox (value), FALSE);
 	}
 
-	if (MONO_TYPE_IS_REFERENCE (f->type))
-		mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), value, FALSE);
-	else
-		mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), mono_object_unbox (value), FALSE);
+	HANDLE_FUNCTION_RETURN ();
 }
 
 ICALL_EXPORT MonoObject *
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -2174,7 +2174,7 @@ typed_reference_to_object (MonoTypedRef *tref, MonoError *error)
 	} else {
 		result = mono_value_box_handle (mono_domain_get (), tref->klass, tref->value, error);
 	}
-	HANDLE_FUNCTION_RETURN_OBJ (result);
+	HANDLE_FUNCTION_RETURN_REF (MonoObject, result);
 }
 
 ICALL_EXPORT void
@@ -2195,12 +2195,10 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, 
 			valueHandle = MONO_HANDLE_NEW (MonoObject, value);
 		goto_if_nok (error, leave);
 		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
-	} else {
-		if (MONO_TYPE_IS_REFERENCE (f->type))
-			mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), value, FALSE);
-		else
-			mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), mono_object_unbox (value), FALSE);
-	}
+	} else if (MONO_TYPE_IS_REFERENCE (f->type))
+		mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), value, FALSE);
+	else
+		mono_copy_value (f->type, (guint8*)obj->value + f->offset - sizeof (MonoObject), mono_object_unbox (value), FALSE);
 
 leave:
 	ICALL_RETURN ();
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -2163,19 +2163,26 @@ leave:
 		mono_gchandle_free (value_gchandle);
 }
 
-ICALL_EXPORT MonoObject*
-mono_TypedReference_ToObject (MonoTypedRef* tref);
+static MonoObjectHandle
+typed_reference_to_object (MonoTypedRef *tref, MonoError *error)
+{
+	HANDLE_FUNCTION_ENTER ();
+	MonoObjectHandle result;
+	if (MONO_TYPE_IS_REFERENCE (tref->type)) {
+		MonoObject** objp = (MonoObject **)tref->value;
+		result = MONO_HANDLE_NEW (MonoObject, *objp);
+	} else {
+		result = mono_value_box_handle (mono_domain_get (), tref->klass, tref->value, error);
+	}
+	HANDLE_FUNCTION_RETURN_OBJ (result);
+}
 
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
-	HANDLE_FUNCTION_ENTER ();
-
-	ERROR_DECL (error);
+	ICALL_ENTRY ();
 
 	MonoClassField *f;
-	MonoObjectHandle objHandle, valueHandle;
-	MonoReflectionFieldHandle fieldHandle;
 
 	g_assert (field);
 	g_assert (obj);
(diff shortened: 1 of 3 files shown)


kg:
Forgot to tag the epic https://github.com/mono/mono/issues/6991 on this.

mono/metadata/icall.c
```diff
@@ -2163,18 +2163,29 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
 		mono_gchandle_free (value_gchandle);
 }
 
+ICALL_EXPORT MonoObject*
+mono_TypedReference_ToObject (MonoTypedRef* tref);
+
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
```
jaykrell: Need the function_enter/exit stuff, as this icall is evidentally not marked HANDLES in icall-def.h.

mono/metadata/icall.c
```diff
@@ -2163,18 +2163,29 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
 
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
+		fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		objHandle = MONO_HANDLE_NEW (MonoObject, mono_TypedReference_ToObject (obj));
+		valueHandle = MONO_HANDLE_NEW (MonoObject, value);
+		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
+		mono_error_set_pending_exception (error);
 		return;
```
jaykrell: And can't return.

mono/metadata/icall.c
```diff
@@ -2163,18 +2163,29 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
+ICALL_EXPORT MonoObject*
+mono_TypedReference_ToObject (MonoTypedRef* tref);
+
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
+	ERROR_DECL (error);
+
 	MonoClassField *f;
+	MonoObjectHandle objHandle, valueHandle;
```
jaykrell: You can move these down to the init. Save lines. Lose no readability.
FooHandle foo = MONO_NEW_HANDLE (foo, …);

mono/metadata/icall.c
```diff
@@ -2186,13 +2188,14 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field,
 		valueHandle = MONO_HANDLE_NEW (MonoObject, value);
 		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
 		mono_error_set_pending_exception (error);
-		return;
+	} else {
+		if (MONO_TYPE_IS_REFERENCE (f->type))
```
jaykrell: No need to indent and brace, can just change first if to else if.

mono/metadata/icall.c
```diff
@@ -2163,25 +2163,39 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
+
 ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
 {
+	HANDLE_FUNCTION_ENTER ();
+
+	ERROR_DECL (error);
+
 	MonoClassField *f;
+	MonoObjectHandle objHandle, valueHandle;
```
jaykrell: Suggest move these down to the initialization instead of leaving them uninitialized and wasting vertical space.

mono/metadata/icall.c
```diff
@@ -2163,25 +2163,39 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
 	g_assert (obj);
 	g_assert (value);
 
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
-		return;
+		fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		objHandle = MONO_HANDLE_NEW (MonoObject, mono_TypedReference_ToObject (obj));
+		valueHandle = MONO_HANDLE_NEW (MonoObject, value);
```
jaykrell: For that matter, you can pass temporaries, don't need the local variables.

kg: Makes sense, I thought they needed to be in scope on the stack but knowing what I do now about setup/teardown for coop I can understand why this is safe.

jaykrell: There is also a shorthand available using token pasting, like if you name the pointers foo_raw then you can get handles foo. HANDLE_DCL or such. But I think best is just to pass temporaries and have no locals.

mono/metadata/icall.c
```diff
@@ -2163,25 +2163,39 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
+		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
+		mono_error_set_pending_exception (error);
```
lambdageek: better to pair this up with `HANDLE_FUNCTION_RETURN()`.

Even better would be to write the function like this:
```c
ICALL_EXPORT void
 ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field, MonoReflectionType *field_type, MonoTypedRef *obj, MonoObject *value, MonoReflectionType *context_type)
{
	ICALL_ENTRY();
	/* "error" already declared by ICALL_ENTRY, HANDLE_FUNCTION_ENTRY also called. */
	...
	/* mono_error_set_pending_exception already called by ICALL_RETURN, HANDLE_FUNCTION_RETURN also called. */
	ICALL_RETURN();
}
```

mono/tests/roslyn-bug-19038.cs
```diff
@@ -0,0 +1,35 @@
...
+            throw new Exception("TypedReference");
+
+        info.SetValueDirect(reference0, 34);
+
+        Console.WriteLine($"a.Value = {a.Value}");
+        if (a.Value != 34)
+            throw new Exception("SetValueDirect");
+
+        int z = 56;
+        if (CopyRefInt(ref z) != 56)
```
lambdageek: Is this testing the same bug? Does CopyRefInt use a typed reference under the hood somehow?

mono/metadata/icall.c
```diff
@@ -2163,25 +2163,39 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
 	g_assert (field);
 	g_assert (obj);
 	g_assert (value);
 
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
-		return;
+		fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		objHandle = MONO_HANDLE_NEW (MonoObject, mono_TypedReference_ToObject (obj));
```
lambdageek: This is hiding a call to `mono_error_set_pending_exception`.
Which means that we don't do any error checking before we call the next thing which may overwrite the pending exception.

It is better not to call icalls from elsewhere in the runtime, ie change `mono_TypedReference_ToObject` to call a helper, and also call that helper from here.

icalls and internal runtime functions don't always behave the same way (for example: icalls that haven't been converted to use coop handles usually call `mono_error_set_pending_exception`, which is something that we generally don't like anymore in internal runtime functions).

```c
static MonoObjectHandle
typed_reference_to_object (MonoTypedRef *tref, MonoError *error)
{
	HANDLE_FUNCTION_ENTER();
	MonoObjectHandle result;
	if (MONO_TYPE_IS_REFERENCE (tref->type)) {
		MonoObject** objp = (MonoObject **)tref->value;
		result = MONO_HANDLE_NEW (MonoObject, *objp);
	} else {
		result = mono_value_box_handle (mono_domain_get (), tref->klass, tref->value, error);
	}
    HANDLE_FUNCTION_RETURN_OBJ(result);
}

ICALL_EXPORT MonoObject*
mono_TypedReference_ToObject (MonoTypedRef* tref)
{
	ICALL_ENTRY();
	MonoObjectHandle result = typed_reference_to_object (tref, error);
	mono_error_set_pending_exception (error);
	ICALL_RETURN_OBJ (result);
	/* or convert ToObject to use handles and then its body is a one-liner */
}

...
/* in ves_icall_System_RuntimeFieldHandle_SetValueDirect */
{
	ICALL_ENTRY();
	...
	{
		objHandle = typed_reference_to_object (obj, error);
		goto_if_nok (error, leave);
		...
	}
	...

leave:
	ICALL_RETURN();
}
```

mono/metadata/icall.c
jaykrell: I never used ICALL_ENTRY/RETURN. Is it worth another pattern?

mono/tests/roslyn-bug-19038.cs
kg: I don't know anything about the test itself, it just happened to use SetValueIndirect. I think they're just testing the overall code generation, which is why it has multiple throws.

kg:
I tried @lambdageek's suggested changes but they silently break setvaluedirect. Still trying to figure out why...

kg:
I can't figure out why the new suggested handle approach doesn't work. I stepped through it in gdb and read over it a couple times and everything looks right. The code runs successfully but the field value isn't updated. Thoughts?

jaykrell:
With all the casting and type unsafety it is difficult to reason about the source without digging more -- anything will compile.
I would just do various printfs.
Set the value to 123 "normally" and 456 dynamically. Use unsafe/fixed as allowed to print addresses, etc.

jaykrell:
You could also call the native code directly from other native code as a test, to cut out some layers.

mono/metadata/icall.c
```diff
@@ -2163,9 +2163,25 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
...
+{
+	HANDLE_FUNCTION_ENTER ();
+	MonoObjectHandle result;
+	if (MONO_TYPE_IS_REFERENCE (tref->type)) {
+		MonoObject** objp = (MonoObject **)tref->value;
+		result = MONO_HANDLE_NEW (MonoObject, *objp);
+	} else {
+		result = mono_value_box_handle (mono_domain_get (), tref->klass, tref->value, error);
+	}
+	HANDLE_FUNCTION_RETURN_OBJ (result);
```
jaykrell: This looks wrong. Should _REF I believe.

mono/metadata/icall.c
```diff
@@ -2174,14 +2190,20 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field,
...
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
-		return;
+		MonoReflectionFieldHandle fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		MonoObjectHandle objHandle = typed_reference_to_object (obj, error), 
+			valueHandle = MONO_HANDLE_NEW (MonoObject, value);
+		goto_if_nok (error, leave);
+		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
+	} else {
```
jaykrell: I'd use "else if {" here, not "else { if {".

mono/metadata/icall.c
kg: I think you're right, the comment says _OBJ returns a bare pointer. I'm surprised it compiles...

jaykrell: Yes. Here are the comments:
// Return a raw pointer from coop handle.
#define HANDLE_FUNCTION_RETURN_OBJ(HANDLE)			\

// Return a coop handle from coop handle.
#define HANDLE_FUNCTION_RETURN_REF(TYPE, HANDLE)			\

So you want _REF. Anything will compile.

mono/tests/roslyn-bug-19038.cs
lambdageek: I think it's worth checking the IL if this compiles to `SetValueIndirect` and if it doesn't, leave it out of this test.

mono/metadata/icall.c
```diff
@@ -2174,14 +2190,18 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field,
...
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
-		return;
-	}
-
-	if (MONO_TYPE_IS_REFERENCE (f->type))
+		MonoReflectionFieldHandle fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		MonoObjectHandle objHandle = typed_reference_to_object (obj, error), 
+			valueHandle = MONO_HANDLE_NEW (MonoObject, value);
+		goto_if_nok (error, leave);
+		ves_icall_MonoField_SetValueInternal (fieldHandle, objHandle, valueHandle, error);
```
lambdageek: I'd change this one to calling helper function too, but it's not incorrect as is.

mono/metadata/icall.c
```diff
@@ -2174,14 +2190,18 @@ ves_icall_System_RuntimeFieldHandle_SetValueDirect (MonoReflectionField *field,
...
 	f = field->field;
 	if (!MONO_TYPE_ISSTRUCT (m_class_get_byval_arg (f->parent))) {
-		mono_set_pending_exception (mono_get_exception_not_implemented (NULL));
-		return;
-	}
-
-	if (MONO_TYPE_IS_REFERENCE (f->type))
+		MonoReflectionFieldHandle fieldHandle = MONO_HANDLE_NEW (MonoReflectionField, field);
+		MonoObjectHandle objHandle = typed_reference_to_object (obj, error), 
+			valueHandle = MONO_HANDLE_NEW (MonoObject, value);
```
jaykrell: I would state MonoObjectHandle twice but it is cosmetic.
```MonoObjectHandle valueHandle = MONO_HANDLE_NEW (MonoObject, value);
MonoObjectHandle objHandle = typed_reference_to_object (obj, error);```

PR #7870, opened 2018-03-28
[genproj] Move obj to common folder and remove unnecessary properties

akoeplinger:
We now default to mcs/class/obj instead of writing to the assembly folder.


akoeplinger:
Tested on Mac and Windows/VS and it seems to work :)

PR #1984, opened 2015-08-16
check for iconv only once

saper:
It is enough to check for iconv
dependencies once in the eglib/configure
- just use the generated libtool
  library file to automatically
  add additional dependencies (if any).

While here, make sue libeglib.a
has libraries returned by LTLIBTOOL
as a libtool dependency.

PR #7873, opened 2018-03-28
check for iconv only once

luhenry:
It is enough to check for iconv
dependencies once in the eglib/configure
- just use the generated libtool
library file to automatically
add additional dependencies (if any).

While here, make sue libeglib.a
has libraries returned by LTLIBTOOL
as a libtool dependency.

Resubmission of https://github.com/mono/mono/pull/1984


luhenry:

luhenry:

Issue #7864, opened 2018-03-28
[wasm] Don't fail when TaskCreationOptions.LongRunning is used


kumpera:
Repro code:
```
Task.Factory.StartNew (() => { }, TaskCreationOptions.LongRunning);
```

This will fail with

```
System.Threading.Tasks.TaskSchedulerException: An exception was thrown by a TaskScheduler. ---> System.ExecutionEngineException: MonoError reused after mono_error_cleanup
  at System.Threading.Thread.StartInternal (System.Security.Principal.IPrincipal principal, System.Threading.StackCrawlMark& stackMark) <0x1df96b0 + 0x00018> in <bd43c24f03094d6f9ea2cbc005ee124e>:0 
```

We're better ignoring `TaskCreationOptions.LongRunning` and treating it as a regular task that failing like that.

kumpera:
Shouldn't this be closed only once I get the mono PR to bump corert merged?

marek-safar:
It closes automatically when you use any of the right keywords like "Fixes"

kumpera:
Oh, didn't know about such automation.
I have probably complained with you about stuff github did on your behalf. Sorry for that. :(

marek-safar:
Here are the keywords https://help.github.com/articles/closing-issues-using-keywords/ (just in case)

Issue #7661, opened 2018-03-16
Suppress the degraded allocation warning for the WebAssembly target


kumpera:
GC is implemented in wasm by switching to degraded mode once the nursery is full and scheduling a GC during the next animation frame.

We still emit the warning, even though it's harmless, and this is causing confusion among our users.

See this report as proof of it: https://github.com/aspnet/Blazor/issues/276

PR #7874, opened 2018-03-28
[wasm] Fix the build plus #7661 and #7864.

kumpera:
Fixes https://github.com/mono/mono/issues/7661
Fixes https://github.com/mono/mono/issues/7864

### DIFF
--- a/mono/metadata/sgen-mono.c
+++ b/mono/metadata/sgen-mono.c
@@ -2703,14 +2703,18 @@ sgen_client_degraded_allocation (void)
 	static gint32 num_degraded = 0;
 
 	gint32 major_gc_count = mono_atomic_load_i32 (&mono_gc_stats.major_gc_count);
+	//The WASM target aways triggers degrated allocation before collecting. So no point in printing the warning as it will just confuse users
+#if !defined (TARGET_WASM)
 	if (mono_atomic_load_i32 (&last_major_gc_warned) < major_gc_count) {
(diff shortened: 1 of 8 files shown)


lewurm:
I'm not a fan of yet another boolean flag (`mono_interpreter_only`). What parts of the runtime need `mono_aot_only` to be true in order to work on wasm?

kumpera:
wasm interp only works under "FullAOT" so... all of them.

This is a regression from @vargaz recent change to support fixed mode that effectively disables the interpreter under FullAOT.

It's not clear to me why his work needs things this way and I didn't want to break him, so this new far was introduced to address the behavior.

As I mentioned, the right fix is to remove all of the mode flags and replace them with feature flags. Things will make more sense and we'll actually be able to reason on specific checks.

vargaz:
Those changes don't break the interpreter, they just prioritize full-aot-ed code when running with both --full-aot and --interp.

kumpera:
all failures are due to github going offline for a bit.

kumpera:
@monojenkins build

luhenry:
The git error is because it can't find the proper commit on external/corert.

luhenry:
@monojenkins build Linux i386

luhenry:
@monojenkins build Linux x64 Coop GC

kumpera:
Superseded by https://github.com/mono/mono/pull/7946

PR #7875, opened 2018-03-28
[corlib] Moves DynamicMethod to mscorlib to work better with wasm

marek-safar:


akoeplinger:
@monojenkins commit apidiff

PR #7877, opened 2018-03-28
[wasm] Add OS detection for WASM. Fixes #7344 [DEPEND ON #7874]

### LINKED ISSUE #7344: Change OSName to return WebAssembly
Change corlib to return WebAssembly so user code can detect when running on the web.

kumpera:
The change was done by ifdef'ing WASM because I don't see a way to change
Environment.Platform / Environment.OSVersion without touching the public
API.

Right now OSDescription is being used to return the execution environment.
Once we properly support node and other non-web embeddings, it will be
adjusted to do so.

Fixes https://github.com/mono/mono/issues/7344
Depends on #7874


kumpera:
When reviewing, please old look at the last commit, as the others are part of a separate PR.

kumpera:
Superseded by https://github.com/mono/mono/pull/7946

PR #7886, opened 2018-03-28
Hide new API which didn't make it to stable on time for dev15.7

marek-safar:


marek-safar:
@monojenkins commit apidiff

PR #7892, opened 2018-03-29
[sdks] Disable btls and support/ on the bcl build, they are not needed.

vargaz:


marek-safar:
I believe this is not the case for Android. It'd be better to do it on per sdk level

lewurm:
@marek-safar there is a separate target for android runtime build, btls should be covered by that I think

marek-safar:
@luhenry how do you build monodroid profile then and why do we need https://github.com/mono/mono/blob/master/sdks/builds/bcl.mk#L13 ?

I think this change is wrong because it misconfigures monodroid profile

vargaz:
This is wrong, it does change the bcl assemblies as well.

luhenry:
@marek-safar Android have a `host-{Linux,Darwin}` which builds the BCL and they pick it from there.

PR #7900, opened 2018-03-29
Bump msbuild to track xplat-master

radical:


akoeplinger:
@monojenkins build pkg

PR #7902, opened 2018-03-29
[Tailcall] Enable tailcall of virtual functions on x86 and ARM64.

jaykrell:


mono/mini/method-to-ir.c
```diff
@@ -7024,14 +7024,16 @@ is_supported_tail_call (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmetho
...
+	gboolean const llvm_only = cfg->llvm_only;
 
 	g_assertf (call_opcode == CEE_CALL || call_opcode == CEE_CALLVIRT || call_opcode == CEE_CALLI, "%s (%d)", mono_opcode_name (call_opcode), (int)call_opcode);
 
 	if (       (fsig->hasthis && m_class_is_valuetype (cmethod->klass))  // This might point to the current method's stack. Emit range check?
 		|| (cmethod->flags & METHOD_ATTRIBUTE_PINVOKE_IMPL)
 		|| cfg->method->save_lmf
 		|| (cmethod->wrapper_type && cmethod->wrapper_type != MONO_WRAPPER_DYNAMIC_METHOD)
-		|| call_opcode == CEE_CALLI
+		|| (!llvm_only && (
```
jaykrell: Is LLVM too willing? For example could we pass a stale local address as part of returning a struct by value? I'd think LLVM owns the ABI there and everywhere and so won't mess it up, so we probably really still have too many restrictions here.

jaykrell:
Woohoo tests passed in one try (there was just one small followup upon the initial submit, probably just undoing more of the arm32 change but I don't remember)

jaykrell:
A few relevant new tests broken, looking.

jaykrell:
@monojenkins build OS X i386

jaykrell:
Ping?

mono/mini/mini-x86.c
```diff
@@ -70,7 +70,42 @@ static mono_mutex_t mini_arch_mutex;
...
+	int bitset;
+	int push [callee_saved_regs_count];
+	int pop [callee_saved_regs_count];
+	int monolmf [callee_saved_regs_count + 1];          // +1 for EBP special case
+	int monolmf_offset [callee_saved_regs_count + 1];   // +1 for EBP special case
+} MonoX86CalleeSavedRegisters;
+
+const static MonoX86CalleeSavedRegisters
+callee_saved_regs = {
+	callee_saved_regs_count,                            // count
```
vargaz: What is the purpose of these changes ? Is the set of registers on x86 going to change in future ? Does it differ between OSes ? The new code looks harder to understand than the old one.

jaykrell: Factoring out the repetition makes it clear what is different vs. the same in each reptition. i.e. just the registers.

jaykrell:
I'll close this and propose similar but cleaner and with some of the commits in separate PR.

jaykrell:
In particular I want to save the arm changes in this branch for maybe another day.

PR #7904, opened 2018-03-29
Revert "[sdks] Disable btls and support/ on the bcl build, they are not needed."

vargaz:
Reverts mono/mono#7892

The change was affecting the bcl build, not just the runtime build.


marek-safar:
@vargaz You can keep `--disable-support-build` it has no BCL influence

PR #7906, opened 2018-03-29
[perfcounters] fix alignment issue on 32bit

lewurm:
clang fails to properly align 64bit members of the `MonoPerfCounters` struct on
32bit builds:

```
* thread #1, name = 'tid_403', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_ARM_DA_ALIGN, address=0x16933124)
    frame #0: 0x0066f4b2 mini`mono_atomic_cas_i64(dest=0x16933124, exch=<unavailable>, comp=<unavailable>) at atomic.c:519 [opt]
   516 	gint64
   517 	mono_atomic_cas_i64(volatile gint64 *dest, gint64 exch, gint64 comp)
   518 	{
-> 519 		return  __sync_val_compare_and_swap (dest, comp, exch);
   520 	}
   521
   522 	#elif defined (TARGET_ANDROID)
(lldb) up
frame #1: 0x006292ea mini`ves_icall_System_Threading_ThreadPool_NotifyWorkItemQueued [inlined] mono_atomic_inc_i64 at atomic.h:397 [opt]
   394 		do {
   395 			get = *val;
   396 			set = get + 1;
-> 397 		} while (mono_atomic_cas_i64 (val, set, get) != get);
   398 		return set;
   399 	}
   400
(lldb) up
frame #2: 0x006292d6 mini`ves_icall_System_Threading_ThreadPool_NotifyWorkItemQueued at threadpool.c:743 [opt]
   740 	ves_icall_System_Threading_ThreadPool_NotifyWorkItemQueued (void)
   741 	{
   742 	#ifndef DISABLE_PERFCOUNTERS
-> 743 		mono_atomic_inc_i64 (&mono_perfcounters->threadpool_workitems);
   744 	#endif
   745 	}
   746
```

Note that the member `threadpool_workitems` is on an unnatural alignment
(`0x16933124`) regarding to its type `gint64`.

related https://github.com/xamarin/xamarin-macios/issues/3826

### DIFF
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -729,9 +729,9 @@ typedef struct {
 	gint32 security_depth;
 	gint32 unused;
 	/* Threadpool */
+	gint32 threadpool_threads;
 	gint64 threadpool_workitems;
 	gint64 threadpool_ioworkitems;
-	gint32 threadpool_threads;
 	gint32 threadpool_iothreads;
 } MonoPerfCounters;
 


lewurm:
this needs to be backported to `2018-02`

jaykrell:
Oh darn. I like to rely on the compiler to align things but I understand how this works -- the operation on the type has a higher alignment requirement than the type -- 64bit integer operations are usually done 32bits at a time so 64bit integers only have 32bit alignment. Is there a portable form of __declspec(align(sizeof(long long)) we can put in front of it? Or a typedef to capture that? I don't like just moving struct fields around, if there is something better. There is no comment and it feels fragile. Would it help to union it with a double? Windows LARGE_INTEGER used to do that, for the very reason to achieve alignment. I can try that.

lewurm:
I think we shouldn't overcomplicate this, if there is an issue on the target it will be apparent pretty quickly.

jaykrell:
It is isn't complicated.
We should use gcc __attribute__(aligned(8)) or msvc __declspec(align(8)) or C11 alignas(16) here. And moving it is ok too to avoid padding. There are warnings that tell you if padding got added that can be nice.

luhenry:

jaykrell:
static_assert (offset(...) % 8) == 0) would also do, it'd fail to compile until moved around

lewurm:
@jaykrell I'm gonna go with this change for now as it's needed to unblock the XI `2018-02` integration. Please feel free to submit a PR with your suggestion for `master`.

lewurm:
@monojenkins backport 2018-02

PR #7907, opened 2018-03-29
[runtime] Fix class visibility check for protected nested classes.

### LINKED ISSUE #7657: Method is inaccessible from method in nested classes with inheritance.
## Steps to Reproduce

1. Compile Program.cs `mcs Program.cs`
2. Run Program.exe `mono Program.exe`

```c#
namespace ConsoleApplication1
{
    public abstract class SuperClass
    {
        protected abstract class SuperInnerAbstractClass
        {
            protected class SuperInnerInnerClass
            {
            }
        }
    }

    public class ChildClass : SuperClass
    {
        private class ChildInnerClass : SuperInnerAbstractClass
        {
            private readonly SuperInnerInnerClass s_class = new SuperInnerInnerClass();
        }

        public ChildClass()
        {
            var childInnerClass = new ChildInnerClass();
        }
    }
    
    internal class Program
    {
        {
            new ChildClass();
        }
    }
}
```

## Current Behavior
The code compiles successfully, but when run on mono `System.MethodAccessException` is thrown.

## Expected Behavior
Code should execute without exception.  I have tested this on Windows 10, Visual Studio 2017 and this code executes, and runs without error.

- [x] macOS
- [ ] Linux
- [ ] Windows

**Version Used**:
```
Mono JIT compiler version 5.10.0.176 (2017-12/a7adb18ce2f Wed Mar 14 06:05:24 EDT 2018)
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-master/8b1520c8aae)
```

### Stacktrace

```
Unhandled Exception:
System.MethodAccessException: Method `ConsoleApplication1.SuperClass+SuperInnerAbstractClass+SuperInnerInnerClass..ctor()' is inaccessible from method `ConsoleApplication1.ChildClass+ChildInnerClass..ctor()'
  at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_throw_method_access(intptr,intptr)
  at ConsoleApplication1.Program.Main (System.String[] args) [0x00001] in /Users/jeff/Projects/ConsoleApplication1/ConsoleApplication1/Program.cs:30 
[ERROR] FATAL UNHANDLED EXCEPTION: System.MethodAccessException: Method `ConsoleApplication1.SuperClass+SuperInnerAbstractClass+SuperInnerInnerClass..ctor()' is inaccessible from method `ConsoleApplication1.ChildClass+ChildInnerClass..ctor()'
  at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_throw_method_access(intptr,intptr)
  at ConsoleApplication1.Program.Main (System.String[] args) [0x00001] in /Users/jeff/Projects/ConsoleApplication1/ConsoleApplication1/Program.cs:30 

```

kumpera:
Fixes https://github.com/mono/mono/issues/7657

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -5236,14 +5236,11 @@ can_access_type (MonoClass *access_klass, MonoClass *member_klass)
 	if (is_nesting_type (access_klass, member_klass) || (m_class_get_nested_in (access_klass) && is_nesting_type (m_class_get_nested_in (access_klass), member_klass)))
 		return TRUE;
 
-	MonoClass *member_klass_nested_in = m_class_get_nested_in (member_klass);
-	if (member_klass_nested_in && !can_access_type (access_klass, member_klass_nested_in))
-		return FALSE;
-
 	/*Non nested type with nested visibility. We just fail it.*/
 	if (access_level >= TYPE_ATTRIBUTE_NESTED_PRIVATE && access_level <= TYPE_ATTRIBUTE_NESTED_FAM_OR_ASSEM && m_class_get_nested_in (member_klass) == NULL)
 		return FALSE;
 
+	MonoClass *member_klass_nested_in = m_class_get_nested_in (member_klass);
 	switch (access_level) {
 	case TYPE_ATTRIBUTE_NOT_PUBLIC:
 		return can_access_internals (access_klass_assembly, member_klass_assembly);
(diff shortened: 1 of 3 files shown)


mono/metadata/class.c
```diff
@@ -5252,16 +5249,16 @@ can_access_type (MonoClass *access_klass, MonoClass *member_klass)
 		return TRUE;
 
 	case TYPE_ATTRIBUTE_NESTED_PUBLIC:
-		return TRUE;
+		return member_klass_nested_in && can_access_type (access_klass, member_klass_nested_in);
```
lambdageek: I don't think this is right.

If you just bring down the original condition, we have:
```
   if (member_klass_nested_in && !can_access (acess_klass, member_klass_nested_in))
     return FALSE;
   else
     return TRUE;
```
which is equivalent to
```c
    return !member_klass_nested_in || can_access_type (access_klass, member_klass_nested_in);
```

lambdageek: On second thought, the old code makes no sense.  Your way is better.

lambdageek:
@monojenkins build

lambdageek:
Fixed merge conflict with master

mono/metadata/class.c
kumpera: It's ok to fail visibility checks for broken code. A class can have nested public visibility and not be nested. I think there's even a check for that early on.

PR #7923, opened 2018-03-30
Mitigate problem seen here where are no functions in the callstack.

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-coop/2298/parsed_console/log.html

***** MonoTests.System.Runtime.Remoting.RemotingServicesTest
***** MonoTests.System.Runtime.Remoting.RemotingServicesTest.Bug46473
Assertion failure in thread 0x7f21a504b740 due to: Cannot transition thread 0x7f21a504b740 from STATE_BLOCKING with DO_BLOCKING
Last 3 state transitions: (most recent first)
[DO_BLOCKING] RUNNING -> STATE_BLOCKING (0) 0 at:
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ef31a) [0x560336a6f31a]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2eb477) [0x560336a6b477]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ec7a2) [0x560336a6c7a2]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ec7e6) [0x560336a6c7e6]

This will provide one function instead of 0 or n, and could be expanded to include more.

This bloats shipping binary some (under condfigure only) which can be addressed with
an #if that condtionally turns the strings to null or empty. And slight code bloat.


mono/utils/mono-threads-coop.c
```diff
@@ -279,7 +279,7 @@ mono_threads_exit_gc_safe_region_unbalanced (gpointer cookie, gpointer *stackdat
 
 	info = (MonoThreadInfo *)cookie;
 
-	check_info (info, "exit", "safe");
+	check_info (info, "exit", "safe", *stackdata);
 
 	switch (mono_threads_transition_done_blocking (info)) {
```
lambdageek: We should augment this one, too. so we get a function name on the `BLOCKING->RUNNING` transition.

mono/utils/mono-threads-coop.c
```diff
@@ -331,20 +331,20 @@ mono_threads_enter_gc_unsafe_region_with_info (THREAD_INFO_TYPE *info, gpointer
 }
 
 gpointer
-mono_threads_enter_gc_unsafe_region_unbalanced (gpointer *stackdata)
+mono_threads_enter_gc_unsafe_region_unbalanced (char const * const *stackdata)
```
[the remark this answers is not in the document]
jaykrell: On the larger point -- this is redundant with backtrace_symbols and so far a subset, but it could match it. And this works but it doesn't.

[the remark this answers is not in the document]
jaykrell: From the JIT. I understand the need, but how? I'm looking.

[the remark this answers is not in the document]
lambdageek: https://github.com/mono/mono/blob/1931cb1dcbca5c24f0184b29db6c3b3dec26f5b0/mono/metadata/marshal-ilgen.c#L4088-L4090
Right here in `emit_thunk_invoke_wrapper_ilgen` - need to store something (empty string?) into the `coop_gc_stack_dummy` var.  Something like:
```
    mono_mb_emit_ldstr (mb, "<invoke wrapper>");
    mono_mb_emit_stloc (mb, coop_gc_stack_dummy);
```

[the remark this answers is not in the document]
jaykrell: How about foo_from_jit that JIT calls instead and initializes it? Which is what I'd do without your answer and would work.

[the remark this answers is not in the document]
jaykrell: mono_mb_emit_ldstr  is what I was missing though.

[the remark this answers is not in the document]
lambdageek: My preference is always for having separate entrypoints for icalls 👍 
(with distinctive `ves_icall_foo` names).

mono/mini/mini-exceptions.c
```diff
@@ -2130,7 +2130,7 @@ mono_handle_exception_internal (MonoContext *ctx, MonoObject *obj, gboolean resu
 
 		if (method->wrapper_type == MONO_WRAPPER_NATIVE_TO_MANAGED && ftnptr_eh_callback) {
 			guint32 handle = mono_gchandle_new (obj, FALSE);
-			gpointer stackptr;
+			const char *stackptr = __func__;
```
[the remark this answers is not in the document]
jaykrell: Yes I'll do that.

mono/utils/mono-threads-coop.c
jaykrell: Got it I think. And maybe trace_state_change?

mono/utils/mono-threads-coop.c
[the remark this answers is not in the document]
jaykrell: What is register_icall for anyway? Just so C# can use dllimport?
Or so managed stack walks get symbols?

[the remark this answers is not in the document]
jaykrell: And for patching, I see.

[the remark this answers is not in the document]
jaykrell: That is, I think we register 4 but only emit 2 here so only register 2.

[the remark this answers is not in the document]
jaykrell: My mistake. We emit calls to all 4.

mono/metadata/threads.c
```diff
@@ -5577,6 +5577,7 @@ mono_threads_detach_coop (gpointer cookie, gpointer *dummy)
 	if (mono_threads_is_blocking_transition_enabled ()) {
 		/* it won't do anything if cookie is NULL
 		 * thread state RUNNING -> (RUNNING|BLOCKING) */
+		dummy = __func__;
```
jaykrell: oops

mono/utils/mono-threads-coop.c
```diff
@@ -167,12 +167,11 @@ return_stack_ptr (gpointer *i)
 }
 
 static void
-copy_stack_data (MonoThreadInfo *info, gpointer *stackdata_begin)
+copy_stack_data (MonoThreadInfo *info, MonoStackDataParam stackdata_begin)
 {
 	MonoThreadUnwindState *state;
 	int stackdata_size;
-	gpointer dummy;
-	void* stackdata_end = return_stack_ptr (&dummy);
```
jaykrell: oops

jaykrell:
Ok added them back, but still internally use the cleaner form.

mono/metadata/marshal-ilgen.c
```diff
@@ -1790,7 +1790,7 @@ emit_native_wrapper_ilgen (MonoImage *image, MonoMethodBuilder *mb, MonoMethodSi
 #endif
 
 		mono_mb_emit_ldloc_addr (mb, coop_gc_stack_dummy);
-		mono_mb_emit_icall (mb, mono_threads_enter_gc_safe_region_unbalanced);
+		mono_mb_emit_icall (mb, ves_icall_mono_threads_enter_gc_safe_region_unbalanced);
 		mono_mb_emit_stloc (mb, coop_gc_var);
```
vargaz: We use ves_icall for real icalls called from managed code, there is no need to rename these imho.

jaykrell: Lambdageek like it, as I understand, both from JIT and C#, though I realize JIT usually does not. I wonder if it should be three naming conventions though. I was going to call it foo_jit or foo_from_jit before lambdageek reiterated ves_icall.

I do need a second name now, but foo_internal would suffice (and foo_helper or foo_impl for foo and foo_internal to bottleneck on)

mono/mini/method-to-ir.c
```diff
@@ -11675,7 +11675,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				CHECK_STACK (info->sig->param_count);
 				sp -= info->sig->param_count;
 
-				if (!strcmp (info->name, "mono_threads_attach_coop")) {
+				if (!strcmp (info->name, "ves_icall_mono_threads_attach_coop")) {
```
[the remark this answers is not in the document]
jaykrell: Maybe it is from mono/mini/aot-compiler.c

mono/metadata/marshal.c
```diff
@@ -264,12 +264,12 @@ mono_marshal_init (void)
 		register_icall (mono_gc_wbarrier_generic_nostore, "wb_generic", "void ptr", FALSE);
 		register_icall (mono_gchandle_get_target, "mono_gchandle_get_target", "object int32", TRUE);
 		register_icall (mono_marshal_isinst_with_cache, "mono_marshal_isinst_with_cache", "object object ptr ptr", FALSE);
-		register_icall (mono_threads_enter_gc_safe_region_unbalanced, "mono_threads_enter_gc_safe_region_unbalanced", "ptr ptr", TRUE);
-		register_icall (mono_threads_exit_gc_safe_region_unbalanced, "mono_threads_exit_gc_safe_region_unbalanced", "void ptr ptr", TRUE);
-		register_icall (mono_threads_enter_gc_unsafe_region_unbalanced, "mono_threads_enter_gc_unsafe_region_unbalanced", "ptr ptr", TRUE);
-		register_icall (mono_threads_exit_gc_unsafe_region_unbalanced, "mono_threads_exit_gc_unsafe_region_unbalanced", "void ptr ptr", TRUE);
-		register_icall (mono_threads_attach_coop, "mono_threads_attach_coop", "ptr ptr ptr", TRUE);
-		register_icall (mono_threads_detach_coop, "mono_threads_detach_coop", "void ptr ptr", TRUE);
+		register_icall (ves_icall_mono_threads_enter_gc_safe_region_unbalanced, "ves_icall_mono_threads_enter_gc_safe_region_unbalanced", "ptr ptr", TRUE);
```
luhenry: These renames create a loooooooot of line damage, let's not do it.

mono/utils/mono-threads-api.h
```diff
@@ -22,34 +22,54 @@ MONO_BEGIN_DECLS
...
+	const char *function_name;
+} MonoStackData;
+typedef const MonoStackData *MonoStackDataParam; // usual
+typedef       MonoStackData *MonoStackDataParamMutable; // rare
+// const here seems to trigger compiler bug, giving static instead of stack storage
+#define MONO_STACK_DATA(x) /*const*/ MonoStackData  x##_value = { __func__ }; \
+                           const MonoStackDataParam x = &x##_value
+
 MONO_API gpointer
-mono_threads_enter_gc_unsafe_region (gpointer* stackdata);
```
luhenry: All these functions are `MONO_API` and used by the products. Changing them effectively mean you will take care of fixing the products for it.

mono/utils/mono-threads-api.h
```diff
@@ -22,34 +22,54 @@ MONO_BEGIN_DECLS
...
+mono_threads_exit_gc_unsafe_region (gpointer cookie, MonoStackDataParam stackdata);
 
 MONO_API gpointer
-mono_threads_enter_gc_unsafe_region_unbalanced (gpointer* stackdata);
+mono_threads_enter_gc_unsafe_region_unbalanced (MonoStackDataParam stackdata);
 
-MONO_API void
-mono_threads_exit_gc_unsafe_region_unbalanced (gpointer cookie, gpointer* stackdata);
+gpointer
+ves_icall_mono_threads_enter_gc_unsafe_region_unbalanced (MonoStackDataParamMutable stackdata);
```
luhenry: The `ves_icall_` should not be in this file. Add them to mono-threads.h

mono/utils/mono-threads-state-machine.c
```diff
@@ -85,10 +85,11 @@ check_thread_state (MonoThreadInfo* info)
 }
 
 static inline void
-trace_state_change (const char *transition, MonoThreadInfo *info, int cur_raw_state, int next_state, int suspend_count_delta)
+trace_state_change_with_func (const char *transition, MonoThreadInfo *info, int cur_raw_state, int next_state, int suspend_count_delta, const char *func)
 {
 	check_thread_state (info);
-	THREADS_STATE_MACHINE_DEBUG ("[%s][%p] %s -> %s (%d -> %d)\n",
+	THREADS_STATE_MACHINE_DEBUG ("%s [%s][%p] %s -> %s (%d -> %d)\n",
```
luhenry: Add the funct to the end. Having the same prefix makes it easier to read a lot of logs.

mono/metadata/marshal.c
jaykrell: ok, I thought they were wanted

mono/utils/mono-threads-api.h
jaykrell: They should still work.

mono/utils/mono-threads-state-machine.c
jaykrell: ok

mono/utils/mono-threads-api.h
jaykrell: The name will change, but probably the move is desired the same, ok.

mono/metadata/marshal.c
jaykrell: Actually I need some new name here, since I'm not keen on updating the products.

jaykrell: I need some new name, so ves_icall_mono_foo, mono_foo_internal?
It is line damage but just a few lines, and I 'need' a new name -- the structuring was kinda messy and confusing with "dummy" and "cookie" and combining them into the same parameter.

mono/utils/mono-threads-api.h
jaykrell: I put them in this file so that mono_foo and similar ves_icall_mono_foo are adjacent but ok.

mono/metadata/marshal-ilgen.c
jaykrell: Maybe I confused the functions and don't require a new name. Let me see again.

jaykrell: Should be able to rename just two functions instead of six..

mono/metadata/marshal.c
luhenry: You can also use these names but declare the functions as static in the current file. This is the approach we have taken elsewhere.

luhenry:
The last review was meant as comments following up the previous one.

jaykrell:
Hm. So, there remains a difference in those four functions, not external vs. internal, but JIT vs. non-JIT. That difference can be eliminated by having the JIT pass a function name (or dummy string).

jaykrell:
gpointer vs. MonoDtackDataParam was abstracting. It changed. It could change again.

jaykrell:
I would rather not move function definitions from one file to another -- too much line damage -- would rather add declarations somewhere. I understand "static" optimizes away the PLT however, that does bug me, is a reason to move functions around, in order to be able to add static.

jaykrell:
Oh, what you mean is.. we only need external_only for non-static (and in a header).
By using static, we don't need to say external_only, as it is very clear and implied.
So that is a second reason for static. Let me look again.

mono/metadata/marshal.c
jaykrell: The functions are referenced in both marshal.c and marshal-ilgen.c, and currently implemented in a third file -- mono-threads-coop.c. "Fixing" that to make them static does not seem worthwhile.

mono/metadata/threads-types.h
```diff
@@ -274,12 +274,18 @@ void mono_thread_join (gpointer tid);
 
 void ves_icall_System_Threading_Thread_GetStackTraces (MonoArray **out_threads, MonoArray **out_stack_traces);
 
-MONO_API gpointer
+MONO_RT_EXTERNAL_ONLY MONO_API gpointer
```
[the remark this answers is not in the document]
jaykrell: They are in public API as lambdageek pointed out, just the two attach/detach functions as I understand.

mono/metadata/marshal-ilgen.c
jaykrell: No, there is reason to rename/extend all six. Though for four of them unclear it is worth it. There are calls from JIT and calls from the macros. The macros can be more easily maintained to fill in the function name. The JIT is less statically checked, so I felt it more maintainable to leave more of the work in C instead of generated IL. However the dynamic binding renames with some level of coupling either way. Originally I didn't know how to maintain the ilgen but that is flimsy and moot -- it is mono_emit_ldstr or such.

mono/utils/mono-threads-api.h
jaykrell: I don't think the enter/exit paths are used by the products. I'll check.
Only coop attach/detach still are.

jaykrell: oh, my mistake.
jays-MacBook-Pro:xamarin-macios jay$ git grep mono_threads_enter_gc_unsafe_region 
runtime/exports.t4:             new Export (true, "void*", "mono_threads_enter_gc_unsafe_region",
runtime/xamarin/runtime.h:              gpointer __gc_unsafe_cookie = mono_threads_enter_gc_unsafe_region (&__dummy)    \
jays-MacBook-Pro:xamarin-macios jay$ git grep mono_threads_enter_
runtime/exports.t4:             new Export (true, "void*", "mono_threads_enter_gc_unsafe_region",
runtime/exports.t4:             new Export (true, "void*", "mono_threads_enter_gc_safe_region",
runtime/xamarin/runtime.h:              gpointer __gc_unsafe_cookie = mono_threads_enter_gc_unsafe_region (&__dummy)    \
runtime/xamarin/runtime.h:              gpointer __gc_safe_cookie = mono_threads_enter_gc_safe_region (&__dummy)        \
jays-MacBook-Pro:xamarin-macios jay$

jaykrell: So I need to treat these like coop attach/detach then and have more variations, or give up on the whole thing..

mono/metadata/threads-types.h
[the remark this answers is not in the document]
luhenry: They are not in the public API since they are not declared in a header that is public. And they do not have any raw object pointers. So they do not qualify.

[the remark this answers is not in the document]
jaykrell: Oh you mean "public API" vs. "used by xamarin", so the second is very bounded and fixable/changable? I should update them? It turns out they duplicate all the GC state transition macros. Perhaps that duplicate can/should be eliminated?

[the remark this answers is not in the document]
jaykrell: I think I should have foo be unchanged be used external and JIT;
foo_internal is the new one; or foo_with_func but that is maybe too specific
foo_common is underlying implementation
So line damage in JIT removed and some other line damage maybe undone.
Just the runtime's use of the macros will improve.

Or maybe put time into fixing backtrace on linux, to use libelf??? And then scrap all this??

mono/metadata/marshal.c
luhenry: So all these functions can be implemented in `marshal.c` and declared in `marshal-internals.h`. There is no need to pollute `mono-threads-api.h` or any other non-marshal file with these declarations since they are only useful for marshalling.

jaykrell:
I meant libdwarf not libelf. But another idea is that checked builds can just make everything visible (except static). That would let backtrace just work more.

mono/utils/mono-threads-coop.h
```diff
@@ -51,27 +51,36 @@ mono_threads_safepoint (void)
...
 		gpointer __gc_unsafe_cookie = mono_threads_enter_gc_unsafe_region_with_info ((info), &__gc_unsafe_dummy)
 
 #define MONO_EXIT_GC_UNSAFE_WITH_INFO	MONO_EXIT_GC_UNSAFE
 
 gpointer
-mono_threads_enter_gc_unsafe_region_unbalanced_with_info (THREAD_INFO_TYPE *info, gpointer *stackdata);
+mono_threads_enter_gc_unsafe_region_unbalanced_with_info (THREAD_INFO_TYPE *info, MonoStackData *stackdata);
+
+gpointer
+mono_threads_enter_gc_unsafe_region_unbalanced_with_info (THREAD_INFO_TYPE *info, MonoStackData *stackdata);
```
jaykrell: accidental duplicate

jaykrell:
Broke profiler tests.

mono/utils/mono-threads-api.h
```diff
@@ -22,35 +22,63 @@ MONO_BEGIN_DECLS
 This API is experimental. It will eventually be required to properly use the rest of the raw-omp embedding API.
 */
 
+typedef struct _MonoStackData {
+	const char *function_name;
+} MonoStackData;
```
luhenry: The issue I have with this approach is that by casting `gpointer*` to `MonoStackData*` (in all the `mono_threads_{enter,exit}_gc_{safe,unsafe}_region*` functions you are encoding in the API - without any other check than hope - that you are not going to change this `MonoStackData` structure to anything else than a `char*` (or anything equivalent to a `gpointer`). And if you do happen to change that structure for some reasons, and you forget to do the same on XI/XM/XA side of things, then it's going to break with random stack corruption. So how will you protect against that?

vargaz:
This looks like a huge amount of changes just to add some debug logging.

mono/utils/mono-threads-api.h
alexanderkyte: Please makes this an opaque type with getters/setters, and add some verification logic in the cast.

jaykrell: @luhenry I thought about this a bit. I think if the struct evolves, the _internal function would need flags or a size, or the struct would have to start with flags (or bitflag) or a size. Or in fact the struct is not great and today we can/should pass a separate const char* parameter. Structs are preferred because intermediate functions can just pass them through by pointer without worrying about their precise contents -- evolving the struct takes work, but less work.

Basically a flag per field, to indicate it is present, or you can use size. Size is easier to initialize, a little awkward to use. flags or bitfield is awkward to initialize but easy to check. Flags or bitfield get used up faster but probably slow enough. It can be a flag per field, easy to understand but used up fast or a "version". 

@alexanderkyte no verification logic is particlarly possible on the cast due to the v1 API, but getter/setter is. Well, we could pass in a size.

Also, to be clear, there is an extra subtletly here. "Normally" the v1 would likely be adapted to v2 with more of a copy in/out. The reason for the cast is because the pointer doubles as a stack pointer. I didn't consider that it could kinda be replaced in the deeper stack. I think that was a slight error though.

However, upon further thought:

How about more like this:

struct MonoStackData {
  union {
    gint32 as_int; // this always present
    struct {
     gbool has_stack_ptr : 1;  // required, but still a flag
     gbool has_function_name : 1; // optional 
   };
  };
  gpointer stack_ptr;
  const char *function_name;
};

jaykrell: To be clear, I agree. What I have in the PR is a bit short sighted and I'll try something more evolvable and safe instead. Even those flags aren't the full story for API evolution -- you need in flags_required, in flags_optional, and out flags_honored, but likely people lose then lose their taste for compatible evolution. Just one set of flags is all people can stomach, and hand wave around the larger problem.

jaykrell:
Safe/evolvable -- missed some of it yet

jaykrell:
Consider removing underscore in function_name -- sort of for longer "words" for search, as I did for stack_*.

jaykrell:
@vargaz Well, agreed, but there is already a lot of code for debugging/logging. It is true all code is costly and bloating, logging, asserts, maintenance, build, etc. We need to find a balance in all things. In this case we could try to lean more on symbols. We have fatal_with_history and render_native_backtrace, very similar, except the first does not use symbols on Linux and the second does, for fairly obvious reasons (one is in-proc library based the other is debugger-based).

mono/utils/mono-threads-coop.c
```diff
@@ -331,22 +353,32 @@ mono_threads_enter_gc_unsafe_region_with_info (THREAD_INFO_TYPE *info, gpointer
...
 		return NULL;
 
 	++coop_reset_blocking_count;
 
-	check_info (info, "enter", "unsafe");
+	const char *function_name = mono_stackdata_get_function_name (stackdata, "");
+
+	check_info (info, "enter", "unsafe", function_name);
 
-	copy_stack_data (info, stackdata);
```
jaykrell: oops

mono/utils/mono-threads-api.h
```diff
@@ -22,35 +22,97 @@ MONO_BEGIN_DECLS
 This API is experimental. It will eventually be required to properly use the rest of the raw-omp embedding API.
 */
 
+typedef struct _MonoStackData {
+	union {
+		int as_int;
+		struct {
+			gboolean has_stackpointer : 1;
```
luhenry: This is quite the overengineering: if stackpointer is NULL then has_stackpointer is FALSE and if stackpointer is non-NULL then has_stackpointer is TRUE, same for function_name and has_function_name, and these 2 `has_*` fields are then used to return NULL or the value (which is itself already NULL or not). Just simplify it and remove this union which is useless.

Otherwise, not-modifying the existing `MONO_API` seems the way to go (as you did it), so looks good to me once you removed this useless union.

jaykrell: The boolean meaning false means the struct is not large enough to access the field it refers to. 

That allows safely growing the size of the struct on one side but not the other.

Currently it does not matter, but this addresses your valid concern of updating XA/XM, and then updating runtime again. With the flags in place, we don't have to keep introducing newly-named functions -- instead add flags to the struct (but then the function name isn't great -- "_internal" -- it'll have to be renamed).

Let's see if this somehow fixes the profiler tests too.

mono/metadata/threads.c
```diff
@@ -5546,38 +5546,55 @@ mono_threads_attach_coop (MonoDomain *domain, gpointer *dummy)
...
-void
-mono_threads_detach_coop (gpointer cookie, gpointer *dummy)
+gpointer
+mono_threads_attach_coop (MonoDomain *domain, gpointer *dummy)
 {
-	MonoDomain *domain, *orig;
-
-	orig = (MonoDomain*) cookie;
+	MonoStackData stackdata;
+	mono_stackdata_init (&stackdata, dummy, NULL);
```
luhenry: Why not passing `__func__` here?

mono/metadata/threads.c
```diff
@@ -5588,6 +5605,23 @@ mono_threads_detach_coop (gpointer cookie, gpointer *dummy)
...
+ *  - @dummy:
+ *    - blocking mode: contains gc unsafe transition cookie
+ *    - non-blocking mode: contains random data
+ *    - a pointer to stack, used for some checks
+ */
+void
+mono_threads_detach_coop (gpointer orig, gpointer *dummy)
+{
+	MonoStackData stackdata;
+	mono_stackdata_init (&stackdata, dummy, NULL);
```
luhenry: Why not passing `__func__` here?

mono/metadata/threads.c
jaykrell: Fair enough.

mono/metadata/threads.c
jaykrell: Fair enough.

jaykrell: At which point I suggest further reduction, in the next commit.

mono/utils/mono-threads-api.h
```diff
@@ -72,13 +72,13 @@ mono_threads_assert_gc_unsafe_region (void);
 MONO_API gpointer
 mono_threads_enter_gc_safe_region (gpointer *stackdata);
 
-gpointer
+MONO_API gpointer // used by profiler
```
luhenry: `MONO_PROFILER_API` is what you want.

mono/utils/mono-threads-api.h
```diff
@@ -22,35 +22,77 @@ MONO_BEGIN_DECLS
...
+
 MONO_API void
 mono_threads_assert_gc_unsafe_region (void);
 
-
-
 MONO_API gpointer
 mono_threads_enter_gc_safe_region (gpointer *stackdata);
 
+MONO_API gpointer // used by profiler
```
luhenry: `MONO_PROFILER_API` is what you want.

mono/utils/mono-threads-api.h
jaykrell: I tried that first and it did not compile. I can try again.

mono/utils/mono-threads-api.h
luhenry: It is defined in `mono-compiler.h`

jaykrell: Yes I saw that but it didn't compile and didn't seem worth fighting. I'll try again.

jaykrell: I tried MONO_PROFILER_API again. It failed again.
It is common to include mono-threads-api.h. without mono-compiler.h.
So I added include mono-compiler.h, that I was avoiding before.

lambdageek:
`thunks.exe` failure on coop looks related.

```
thunks.exe
=============== thunks.exe.stdout ===============
Assertion failure in thread 0x7f9a66204740 due to: object.c:6593: Expected GC Unsafe mode but was in STATE_BLOCKING_AND_SUSPENDED state
Last 3 state transitions: (most recent first)
[DO_BLOCKING] RUNNING -> STATE_BLOCKING_AND_SUSPENDED (0) 0 at:
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2f0daa) [0x558cbd05bdaa]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ecdcb) [0x558cbd057dcb]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ee135) [0x558cbd059135]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(mono_method_get_unmanaged_thunk+0x72) [0x558cbcf70652]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/tests/.libs/libtest.so.0(test_method_thunk+0xbe) [0x7f9a621f2e9e]
	at [0x40080811]
[ABORT_BLOCKING] STATE_BLOCKING_AND_SUSPENDED -> RUNNING (0) 0 at:
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2f0daa) [0x558cbd05bdaa]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ed1bc) [0x558cbd0581bc]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ee37e) [0x558cbd05937e]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ee436) [0x558cbd059436]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(mono_method_get_unmanaged_thunk+0x41) [0x558cbcf70621]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/tests/.libs/libtest.so.0(test_method_thunk+0xbe) [0x7f9a621f2e9e]
	at [0x40080811]
[DO_BLOCKING] RUNNING -> STATE_BLOCKING_AND_SUSPENDED (0) 0 at:
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2f0daa) [0x558cbd05bdaa]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ecdcb) [0x558cbd057dcb]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2ee135) [0x558cbd059135]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(mono_method_get_unmanaged_thunk+0x72) [0x558cbcf70652]
	at /mnt/jenkins/workspace/test-mono-pull-request-coop/mono/tests/.libs/libtest.so.0(test_method_thunk+0xa7) [0x7f9a621f2e87]
	at [0x40080811]


=============== thunks.exe.stderr ===============
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) Test.test_method_thunk (int,intptr,intptr) [0x0000c] in <56467047b41b401cb2f39814338c05a1>:0
  at Test.RunTests (int,System.Type) [0x0007a] in <56467047b41b401cb2f39814338c05a1>:0
  at Test.Main () [0x00001] in <56467047b41b401cb2f39814338c05a1>:0
  at (wrapper runtime-invoke) <Module>.runtime_invoke_int (object,intptr,intptr,intptr) [0x0001e] in <56467047b41b401cb2f39814338c05a1>:0
...
Native stacktrace:

	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0xcc669) [0x558cbce37669]
	/lib/x86_64-linux-gnu/libpthread.so.0(+0x110c0) [0x7f9a656d30c0]
	/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcf) [0x7f9a6513efff]
	/lib/x86_64-linux-gnu/libc.so.6(abort+0x16a) [0x7f9a6514042a]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2e02f4) [0x558cbd04b2f4]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(monoeg_g_logv+0x40) [0x558cbd060500]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(monoeg_g_log+0x82) [0x558cbd0605b2]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2f0f44) [0x558cbd05bf44]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(+0x2f11a9) [0x558cbd05c1a9]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/mini/mono(mono_string_new_wrapper+0x15) [0x558cbcf776a5]
	/mnt/jenkins/workspace/test-mono-pull-request-coop/mono/tests/.libs/libtest.so.0(test_method_thunk+0x1d6) [0x7f9a621f2fb6]
	[0x40080811]

Debug info from gdb:

[New LWP 17149]
[New LWP 17150]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x00007f9a656d2b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
  Id   Target Id         Frame 
* 1    Thread 0x7f9a66204740 (LWP 17125) "mono" 0x00007f9a656d2b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
  2    Thread 0x7f9a64bff700 (LWP 17149) "SGen worker" 0x00007f9a656cf15f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
  3    Thread 0x7f9a62622700 (LWP 17150) "Finalizer" 0x00007f9a656d1536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0

Thread 3 (Thread 0x7f9a62622700 (LWP 17150)):
#0  0x00007f9a656d1536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007f9a656d15e4 in __new_sem_wait_slow.constprop.0 () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x0000558cbcfdbaab in mono_os_sem_wait (flags=MONO_SEM_FLAGS_ALERTABLE, sem=0x558cbd3a1820 <finalizer_sem>) at ../../mono/utils/mono-os-semaphore.h:209
#3  mono_coop_sem_wait (flags=MONO_SEM_FLAGS_ALERTABLE, sem=0x558cbd3a1820 <finalizer_sem>) at ../../mono/utils/mono-coop-semaphore.h:43
#4  finalizer_thread (unused=unused@entry=0x0) at gc.c:903
#5  0x0000558cbcf93fd2 in start_wrapper_internal (stack_ptr=<optimized out>, start_info=0x0) at threads.c:1102
#6  start_wrapper (data=0x558cbf0c6900) at threads.c:1162
#7  0x00007f9a656c9494 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#8  0x00007f9a651f4acf in clone () from /lib/x86_64-linux-gnu/libc.so.6

Thread 2 (Thread 0x7f9a64bff700 (LWP 17149)):
#0  0x00007f9a656cf15f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x0000558cbd0410eb in mono_os_cond_wait (mutex=0x558cbd3b0680 <lock>, cond=0x558cbd3b0640 <work_cond>) at ../../mono/utils/mono-os-mutex.h:173
#2  get_work (job=<synthetic pointer>, do_idle=<synthetic pointer>, work_context=<synthetic pointer>, worker_index=0) at sgen-thread-pool.c:165
#3  thread_func (data=<optimized out>) at sgen-thread-pool.c:196
#4  0x00007f9a656c9494 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#5  0x00007f9a651f4acf in clone () from /lib/x86_64-linux-gnu/libc.so.6

Thread 1 (Thread 0x7f9a66204740 (LWP 17125)):
#0  0x00007f9a656d2b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x0000558cbce3773e in mono_handle_native_crash (signal=0x558cbd085c96 "SIGABRT", ctx=<optimized out>, info=<optimized out>) at mini-exceptions.c:2780
#2  <signal handler called>
#3  0x00007f9a6513efff in raise () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007f9a6514042a in abort () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x0000558cbd04b2f4 in mono_log_write_logfile (log_domain=<optimized out>, level=<optimized out>, hdr=<optimized out>, message=0x558cbf155600 "Assertion failure in thread 0x7f9a66204740 due to: object.c:6593: Expected GC Unsafe mode but was in STATE_BLOCKING_AND_SUSPENDED state\nLast 3 state transitions: (most recent first)\n[DO_BLOCKING] RUNN"...) at mono-log-common.c:135
#6  0x0000558cbd060500 in monoeg_g_logv (log_domain=0x0, log_level=G_LOG_LEVEL_ERROR, format=<optimized out>, args=args@entry=0x7ffdf35e1e38) at goutput.c:115
#7  0x0000558cbd0605b2 in monoeg_g_log (log_domain=log_domain@entry=0x0, log_level=log_level@entry=G_LOG_LEVEL_ERROR, format=<optimized out>) at goutput.c:125
#8  0x0000558cbd05bf44 in mono_fatal_with_history (msg=msg@entry=0x558cbd121d08 "%s:%d: Expected GC Unsafe mode but was in %s state") at checked-build.c:244
#9  0x0000558cbd05c1a9 in assert_gc_unsafe_mode (file=file@entry=0x558cbd0ae08f "object.c", lineno=lineno@entry=6593) at checked-build.c:290
#10 0x0000558cbcf776a5 in mono_string_new_wrapper (text=text@entry=0x7f9a6220d5ac "foo") at object.c:6593
#11 0x00007f9a621f2fb6 in test_method_thunk (test_id=2, test_method_handle=<optimized out>, create_object_method_handle=0x558cbf0eb710) at libtest.c:3751
#12 0x0000000040080811 in ?? ()
#13 0x0000000000000002 in ?? ()
#14 0x00007f9a64c09728 in ?? ()
#15 0x00007f9a64c01df8 in ?? ()
#16 0x0000000000000000 in ?? ()

=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================


```

mono/utils/mono-threads-coop.c
```diff
@@ -196,16 +198,16 @@ copy_stack_data (MonoThreadInfo *info, gpointer *stackdata_begin)
 }
 
 static gpointer
-mono_threads_enter_gc_safe_region_unbalanced_with_info (MonoThreadInfo *info, gpointer *stackdata);
+mono_threads_enter_gc_safe_region_unbalanced_with_info (MonoThreadInfo *info, MonoStackData *stackdata);
 
 gpointer
-mono_threads_enter_gc_safe_region (gpointer *stackdata)
+mono_threads_enter_gc_safe_region_internal (MonoStackData *stackdata)
```
lambdageek: We still need `mono_threads_enter_gc_safe_region` (and `mono_threads_exit_gc_safe_region`) are still in the header, but don't have an implementation anymore?

Also the `thunks.exe` failure on coop is directly related - it tries to use these functions to move from GC Safe mode to GC Unsafe, and back.

jaykrell: Agreed the failure is related and I will followup.
I will double check that I didn't remove anything, quite possible.

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X i386

PR #7931, opened 2018-03-31
Remove failing jpeg flags tests

hughbe:
Fixed in https://github.com/mono/libgdiplus/pull/261


akoeplinger:
For the record: gdiplus.dll returns different flags on Windows/.NET and we're aligning libgdiplus with that so the tests would start to fail.

PR #7940, opened 2018-04-02
[profiler] Fix exception clause reporting.

### LINKED ISSUE #7525: Profiler exception clause reporting is broken

alexrp:
This fixes three separate issues:

* Finally clauses in the non-exceptional case were not reported because we implement these by simply calling them from managed code, as opposed to the exceptional case where we call them from runtime C code.
* The catch portion of a try-filter-catch clause was being incorrectly reported as a filter.
* Actual exception filters were not being reported at all.

Because we have to instrument JIT'd code to report non-exceptional finally clauses, a new mono_profiler_enable_clauses () function has been introduced which profiler API users must call in order to get exception clause events, in a similar fashion to mono_profiler_enable_allocations (). Since this is a breaking change, the profiler API version has been bumped.

Fixes #7525.

---

I included some profiler API cleanups since I bumped the API version anyway.


alexrp:
build profiler stress

luhenry:
@monojenkins build API Diff

alexrp:
@vargaz do the JIT parts look ok?

mono/mini/method-to-ir.c
```diff
@@ -710,7 +710,11 @@ mono_find_leave_clauses (MonoCompile *cfg, unsigned char *ip, unsigned char *tar
 		clause = &header->clauses [i];
 		if (MONO_OFFSET_IN_CLAUSE (clause, (ip - header->code)) && 
 		    (!MONO_OFFSET_IN_CLAUSE (clause, (target - header->code)))) {
-			res = g_list_append_mempool (cfg->mempool, res, clause);
+			MonoLeaveClause *leave = mono_mempool_alloc (cfg->mempool, sizeof (MonoLeaveClause));
+			leave->index = i;
```
vargaz: Suggest using alloc0 for defensive coding.

jaykrell: Strongly agreed, zeroing is great, and allows for evolution of types without visiting every initializer. It is the next best thing to C++ constructors, or perhaps better due to efficiency and works with globals. memset, ZeroMemory, HEAP_ZERO_MEMORY, calloc, etc.

alexrp:
@monojenkins build failed

PR #7941, opened 2018-04-02
[System.MathF] Bring support for the single-precission Math operations to Mono

migueldeicaza:
This pull request brings support to Mono's class libraries for the new `System.MathF` type which includes operations for single-floating point.   

This is needed to run some new applications like:

https://github.com/aras-p/ToyPathTracer

With this change, it is possible to run this unmodified.  With this support, and the `-O=float32` option, it is possible to improve the performance from 6.3Mrays/sec to 7.8Mrays/sec on a Mac.

Oddly, this seems to be faster than .NET Core 2.1.4, which is at 4.2Mrays/sec.


mcs/class/referencesource/mscorlib/system/math.cs
```diff
@@ -62,7 +62,233 @@ public static class Math {
       [ResourceExposure(ResourceScope.None)]
       [MethodImplAttribute(MethodImplOptions.InternalCall)]
       public static extern double Ceiling(double a);      
-      
+
+#if MONO
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static byte Clamp(byte value, byte min, byte max)
+        {
```
vargaz: This should use K&R indentation.

mono/metadata/sysmath.c
```diff
@@ -316,3 +316,142 @@ ves_icall_System_Math_SplitFractionDouble (gdouble *v)
 {
 	return modf (*v, v);
 }
+
+gfloat
+ves_icall_System_MathF_Acos (gfloat x)
```
vargaz: This should be just float, we don't use gfloat.

mono/metadata/icall-def.h
```diff
@@ -439,6 +439,31 @@ ICALL(MATH_16, "Sqrt", ves_icall_System_Math_Sqrt)
 ICALL(MATH_17, "Tan", ves_icall_System_Math_Tan)
 ICALL(MATH_18, "Tanh", ves_icall_System_Math_Tanh)
 
+ICALL_TYPE(MATHF, "System.MathF", MATHF_1)
+ICALL(MATHF_1, "Acos", ves_icall_System_MathF_Acos)
```
vargaz: These can be HANDLES(..) I think.

mcs/class/referencesource/mscorlib/system/math.cs
migueldeicaza: I copy pasted this from CoreCLR, I rather not change the style of the copy-pasted code.

mono/metadata/sysmath.c
migueldeicaza: We do in that file, just a couple of lines above ;-)

I dont mind updating.

mono/metadata/icall-def.h
migueldeicaza: What do you mean?

jaykrell: Many but not all lines icall-def.h have HANDLES() around them.
This is how icalls get converted to be coop-friendly.
1. passes/returns coop handles instead of raw pointers
2. passes an extra, initialized MonoError*
3. Check the error at the end.
Eventually every line should either have HANDLES or some comment or other macro explaining why not -- for example, maybe the only one, if no pointers and no errors and performance sensitive, that combination does not want HANDLES.

kumpera: @vargaz wouldn't HANDLE'izing those functions produce a net loss of perf for something they don't need? IE, none raise exceptions or take ref pointers.

mono/metadata/icall-def.h
```diff
@@ -439,6 +439,31 @@ ICALL(MATH_16, "Sqrt", ves_icall_System_Math_Sqrt)
...
+ICALL(MATHF_8, "Cbrt", ves_icall_System_MathF_Cbrt)
+ICALL(MATHF_9, "Ceiling", ves_icall_System_MathF_Ceiling)
+ICALL(MATHF_10, "Cos", ves_icall_System_MathF_Cos)
+ICALL(MATHF_11, "Cosh", ves_icall_System_MathF_Cosh)
+ICALL(MATHF_12, "Exp", ves_icall_System_MathF_Exp)
+ICALL(MATHF_22, "FMod", ves_icall_System_MathF_FMod)
+ICALL(MATHF_13, "Floor", ves_icall_System_MathF_Floor)
+ICALL(MATHF_14, "Log", ves_icall_System_MathF_Log)
+ICALL(MATHF_15, "Log10", ves_icall_System_MathF_Log10)
+ICALL(MATHF_23, "ModF(float,float&)", ves_icall_System_MathF_ModF)
```
kumpera: This signature doesn't match the managed signature.

mono/metadata/sysmath.c
```diff
@@ -316,3 +316,142 @@ ves_icall_System_Math_SplitFractionDouble (gdouble *v)
...
+}
+
+float
+ves_icall_System_MathF_FMod (float x, float y)
+{
+	return fmodf (x, y);
+}
+
+float
+ves_icall_System_MathF_ModF (float x, float *d)
```
kumpera: This icall must null check `d`

migueldeicaza: I changed this to be ref, do we still need the check in that case?

CoreCLR does not have that.

mono/metadata/icall-def.h
migueldeicaza: Good catch, thanks!

mono/metadata/sysmath.c
kumpera: If managed was using a byref instead of a pointer, we won't need it as we emit caller side checks. That doesn't happen with pointers.

kumpera: Looks like there's it's only used by Truncate and Round with locals, so it's just lousy code on their end but we should be safe.

mono/metadata/icall-def.h
migueldeicaza: I changed my mind, I kept the .NET signature.

migueldeicaza: Short version: done

mono/metadata/sysmath.c
migueldeicaza: Added a check, just for sanity's sake.

migueldeicaza:
Mhm, the build breaks elsewhere with a series of errors like this:

```
/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx/external/corefx/src/System.Numerics.Vectors/src/System/Numerics/Matrix3x2.cs(281,26): error CS0436: The type 'MathF' in '/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx/mcs/class/System.Numerics/../../../external/corefx/src/Common/src/System/MathF.netstandard.cs' conflicts with the imported type 'MathF' in 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. Using the type defined in '/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx/mcs/class/System.Numerics/../../../external/corefx/src/Common/src/System/MathF.netstandard.cs'.
```

luhenry:
@monojenkins build Linux x64

luhenry:

luhenry:
@akoeplinger @marek-safar can you please check the API Diff

akoeplinger:
The API diff looks fine to me.

mcs/class/corlib/System/MathF.mono.cs
```diff
@@ -0,0 +1,88 @@
...
+
+//This class contains only static members and doesn't require serialization.
+
+using System.Runtime.CompilerServices;
+
+namespace System
+{
+    public static partial class MathF
+    {
+        [MethodImpl(MethodImplOptions.InternalCall)]
```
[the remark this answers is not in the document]
migueldeicaza: I rather not for code that is copy pasted, makes it simpler to track changes.

mcs/class/corlib/corlib.dll.sources
```diff
@@ -978,6 +978,8 @@ ReferenceSources/Type.cs
 ../referencesource/mscorlib/system/invalidtimezoneexception.cs
 ../referencesource/mscorlib/system/Lazy.cs
 ../referencesource/mscorlib/system/math.cs
+../../../external/corert//src/System.Private.CoreLib/shared/System/MathF.cs
```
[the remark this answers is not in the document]
migueldeicaza: Done

marek-safar:
The System.Runtime.Extensions change for a desktop could be problematic but let's see if anyone hits it

marek-safar:
@monojenkins commit apidiff

Issue #7247, opened 2018-02-23
CoreFX System.Text namespace


marek-safar:
Import all code and tests

PR #7949, opened 2018-04-02
[corlib] StringBuilder from corefx

EgorBo:
Part of #7247


kumpera:
Does this mean we can we remove the runtime hacks for mono's stringbuilder?

vargaz:
Mono depends on the exact layout/behavior of stringbuilder for marshalling stringbuilders to/from native code, so this change will probably break that.

EgorBo:
Well it's more like WIP, in order to see test results on CI.
@vargaz the layout is the same, is not it?

EgorBo:
PS: important node - Append won't use Spans until all primitive types implement `ISpanFormattable` 
https://github.com/mono/corefx/blob/master/src/Common/src/CoreLib/System/Text/StringBuilder.cs#L1034-L1037

marek-safar:
@kumpera @vargaz it didn't break the marshalling code or are we missing tests for it?

On the other hand, it'd be great to remove that code from runtime and just use the code from CoreRT https://github.com/dotnet/corert/blob/7a6d611ee5dc93a7cc94b48530c208c4b42821c9/src/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs#L131-L162 as it looks like we don't handle it in fully compatible way

kumpera:
@marek-safar I'm not aware of any issues WRT marshalling of string-builder. Those one-liners from corert won't help at all, we'd still need to deal a lot by hand.

What I meant was whether we still need the GC support for String trimming as our StringBuilder used to.

marek-safar:
@kumpera they are not one-liners they are just entry points, we could use that instead of writing our StringBuilder marshalling code in runtime

> What I meant was whether we still need the GC support for String trimming

I remember we experimented with something like that but I cannot remember if that code was ever merged into master. Do you remember where did it end up?

EgorBo:
Tests for marshalling: https://github.com/mono/mono/blob/master/mono/tests/pinvoke2.cs#L310-L329 (including array of SB). According to the results ^ works fine

kumpera:
@marek-safar we have support for trimming for ages now, it was a major PITA for sgen.

marek-safar:
@kumpera I was looking at the logs but could not find it. Could you point me to the code?

kumpera:
@marek-safar see 860083dcf14525ebc6979b5763cf7116837a6367

marek-safar:
Yeah, we for a long time don't use that code because our StringBuilder uses smaller chunks for temp buffers.

We might need to update the constant if it does not fit well to our GC though.

https://github.com/mono/mono/blob/eda870e6d62aefb723780b82be89e8b0f49e0b47/mcs/class/referencesource/mscorlib/system/text/stringbuilder.cs#L74-L78

marek-safar:
@EgorBo please update monodoc tests

marek-safar:

PR #7952, opened 2018-04-02
Fix stack overflow on ServiceHost initialisation

### LINKED ISSUE #7305: Stack overflow when hosting a WCF service when contract interface inherits other interfaces
## Steps to Reproduce

```
using System;
using System.ServiceModel;

namespace Test
{
    public class AmBuggy : IAmBuggy
    {
        public void BugMe()
        {
            throw new Exception("Something bad happened!");
        }

        public void Dispose()
        {
            throw new NotImplementedException();
        }
    }

    [ServiceContract]
    public interface IAmBuggy : IDisposable
    {
        [OperationContract]
        void BugMe();
    }

    internal static class Program
    {
        {
            var binding = new BasicHttpBinding();
            var address = new Uri("http://localhost:8080");
            var host = new ServiceHost(typeof(AmBuggy));
            host.AddServiceEndpoint(typeof(IAmBuggy), binding, address);
            host.Open();
            Console.ReadLine();
            host.Close();
        }
    }
}
```

## Current Behavior

Crashing

## Expected Behavior

Not crashing

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 5.8.0 ((HEAD/1eac626536b Thu Feb 15 15:05:54 CET 2018)
	TLS:           __thread
	SIGSEGV:       normal
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 

### Stacktrace

I don't have a stacktrace but I have manually debugged and observed that this is a stack overflow bug: the function ContractDescriptionGenerator.GetContractInternal is being recursively called to eternity, at the first call recursion the function variables are set like this:
```
givenContractType=typeof(Test.IAmBuggy)
givenServiceType=typeof(Test.AmBuggy)
serviceTypeForCallback=null
```
and I suppose this is good.

At the second call recursion the function variables are set like this:
```
givenContractType=typeof(System.IDisposable)
givenServiceType=typeof(Test.AmBuggy)
serviceTypeForCallback=null
```
still good, but then a third call recursion occurs and from this point on everything just repeats where the function variables are set like this:
```
givenContractType=typeof(System.IDisposable)
givenServiceType=typeof(Test.AmBuggy)
serviceTypeForCallback=null
```
Eventually the app segfaults from the stack overflow.

This problem can be worked around by giving the AmBuggy class inheritance of IDisposable interface, then no stack overflow would occur.

tdinucci:
Fixes https://github.com/mono/mono/issues/7305

### DIFF
--- a/mcs/class/System.ServiceModel/System.ServiceModel.Description/ContractDescriptionGenerator.cs
+++ b/mcs/class/System.ServiceModel/System.ServiceModel.Description/ContractDescriptionGenerator.cs
@@ -233,7 +233,9 @@ namespace System.ServiceModel.Description
 			 */
 
 			var inherited = new Collection<ContractDescription> ();
-		var interfaces = cd.ContractType.GetInterfaces ();
+			var interfaces = cd.ContractType.GetInterfaces ()
+				.Where (t => t != givenContractType);
+				
 			foreach (var it in interfaces ) {
 				var icd = GetContractInternal (it, givenServiceType, null);
 				if (icd != null)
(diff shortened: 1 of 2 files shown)


luhenry:
build

marek-safar:
Thank you for the contribution

PR #7956, opened 2018-04-03
Fix .assembly directive to make some tailcall tests AOT compatible.

jaykrell:


luhenry:
@monojenkins build Linux x64

PR #7961, opened 2018-04-03
[jit] Make x86 callee saved register codegen data driven to reduce repetition.

jaykrell:
repetition. Make it clearer what is same vs. different in repetitive code. Correct references to caller saved registers to be callee save.


jaykrell:

luhenry:
It didn’t get any traction and it does not impact anything.

jaykrell:
It makes the code easier to read.

PR #7962, opened 2018-04-03
[Tailcall] require exact matching return type

jaykrell:


jaykrell:
I know this looks faulty, because it disables working tests, but 2 of them are definitely dubious and it was my gut feeling before I noticed the failing tests, and @vargaz insisted.

jaykrell:
@monojenkins build failed

Issue #6914, opened 2018-02-09
Epic: Tail calls optimizations


marek-safar:
The goal is to make our tail call optimization to match netcore/.NET

## Milestone I

[ ] Bring coreclr tests for tail calls to Mono source tree
[ ] Update our JIT to pass all tail calls tests we have
[ ] Extend our AOT compiler to support tail calls and pass all our tests

marek-safar:
Collecting notes for somewhat future tailcall work.

csc can emit things like:
IL_0002: callvirt instance string Test.Base::AbstractOverrideNil()
IL_0007: stloc.0
IL_0008: br.s IL_000a
IL_000a: ldloc.0
IL_000b: ret

or if you say csc -optimize:
IL_0001: callvirt instance string Test.Base::VirtualOverrideNil()
IL_0006: ret

JIT should recognize the second and possibly the first pattern and attempt to tail call. Even without the tail prefix.

The first pattern is somewhat generalizable in that "effectively nop" is unbounded -- a general optimization problem. Some of this this work might involve more surveying of what csc/mcs tends to emit. And surveying of people tend to use csc -optimize. The second pattern is trivial.

If these are already handled, then just add tests to assert it.

Meta-point is that csc never outputs "tail." yet programmers might deliberately write code amenable to tail optimizations. Until/unless C#/csc are fixed, the burden is all on the JIT. C# should gain a syntax though, and then this bug can be skipped.

jaykrell:
Pretty much everything is believed to work. There are a few known problems. Traversing the test matrix takes a long time.
Known problems:

1. arm32 can't pass "extra arg" in a tailcall, so that breaks many generic/interface calls.

2. gsharedvt has wrappers which break tailcall on FullAOT only. This can probably be reduced and work in in-only scenarios. gsharedvt isn't just about recursive generic parameters, but also interface calls and/or virtual calls, because FullAOT doesn't see the functions as used so doesn't compile them. Suggestion is intersected tailcaled interface/virtual signatures with interface/virtual signatures and compile those. This is overkill, but would not penalize C# FullAOT at all (it has no tailcalls).

3. Of course, passing more parameters on stack than recieved. Big difficult problem and CoreCLR on Unix is believed no better, but CLR on desktop is better.

4. Some scenarios involving valuetypes, like passing them by value on some architectures (arm64, amd64), and returning them by value similarly. This can be improved by reusing space from incoming parameters.

[...]

PR #7963, opened 2018-04-03
[Tailcall] Give LLVM more leeway.

jaykrell:


mono/mini/method-to-ir.c
```diff
@@ -7046,7 +7046,8 @@ is_supported_tail_call (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmetho
 		|| (cmethod->flags & METHOD_ATTRIBUTE_PINVOKE_IMPL)
 		|| cfg->method->save_lmf
 		|| (cmethod->wrapper_type && cmethod->wrapper_type != MONO_WRAPPER_DYNAMIC_METHOD)
-		|| call_opcode == CEE_CALLI
+		|| (!cfg->llvm_only && (
+		   call_opcode == CEE_CALLI
 		|| (virtual_ && !cfg->backend->have_op_tail_call_membase)
```
vargaz: This if looks pretty complicated now.

jaykrell: There is a slight restructuring that might help you read it. Let me try.

jaykrell:
Better? This is a bit experimental as I haven't done the test but I think it is reasonable.
In either case, this makes sense to separate from the other PR in which it first appeared.

jaykrell:

jaykrell:
It should have slightly less, and it should/could decide for itself.
In particular, it should have fewer restrictions around generic context and maybe interface method dispatch. i.e. should enable ARM32 to tailcall more, with LLVM.

jaykrell:
As I understand, llvmonly has at least one fewer restrictions than mini and we should let it take advantage that.
Specifically that rgctx/imt is passed “within” the ABI and not “outside” of the ABI.
That should let more things work with llvmonly — at least with ARM32.
It might have advantages in gsharedvt also.

I feels like a lot of our checks are about mini, and LLVM should just be allowed to do whatever it can, at least given the tail. prefix — maybe even without it. Even things like lmf it should be able to figure out itself cannot be tailcalled, because there is stuff after the call. I think we are just unnecessarily tyings its hands.

Other examples are the stack_usage==0 in arm64, or stack_usage < 64 in arm32. I’d like to visit these, understand them, see about fixing them, but perhaps they can be lifted right away for llvmonly.

jaykrell:
Maybe I add some tests that this moves from failing to passing? I believe I can.

vargaz:
Even if llvm supports this, I don't think we should support different tail calls with different configurations, it would mean that an app would work in some config and not in others. Also, llvmonly is only used on the watch, and any potential problems are hard to debug there.

jaykrell:
Right, I wasn't sure about that -- @luhenry is the goal to fill as many spots in the matrix, or to have it be kinda "level".

Will any F# user be serious enough for about arm32 tailcall so as to use llvmonly because it can be better?

And what are we willing to do for arm32? For example, generate llvm-like generic thunks instead of using the extra register? Or just don't do arm32 well?

jaykrell:
Ps: We can certainly enable llvmonly CI on more platforms, independently of the shipping scenarios. But that doesn't address the kinda jaggy/sparse resulting matrix which I think is the larger concern.

jaykrell:
The matrix: https://github.com/mono/mono/issues/6914

luhenry:
@jaykrell we want to strive to have the same features working across the board, with the expected slight differences between platforms. Any difference needs to be justified, limited and understood; this is to minimize chances of a bug appearing in 1 configuration and not the other, thus for obvious reasons of limited manpower and the load of maintaining all these configurations. Unfortunately, giving LLVM more leeway is not limited nor justified, since it will give to the `bitcode`/`llvmonly` configuration different constraints than to the others.

PR #7985, opened 2018-04-03
[Tailcall] Dial back single -v output, and enable under MONO_LOG_LEVEL=debug MONO_LOG_MASK=tailcall.

jaykrell:


mono/mini/mini-runtime.c
```diff
@@ -3581,6 +3581,8 @@ mini_parse_debug_option (const char *option)
 	// It is asserted.
 	else if (!strcmp (option, "test-tailcall-require"))
 		mini_debug_options.test_tailcall_require = TRUE;
+	else if (!strcmp (option, "tailcall"))
+		mini_debug_options.tailcall = TRUE;
 	else
```
vargaz: This might be better named 'log-tailcall' or something.

jaykrell: I thought of that and am ok with that but I figured printing is the most basic of debugging feature. Perhaps that is my bias.

mono/mini/method-to-ir.c
```diff
@@ -2176,7 +2175,7 @@ test_tailcall (MonoCompile *cfg, MonoBoolean tailcall)
 	// Do not change "tailcalllog" here without changing other places, e.g. tests that search for it.
 	//
 	g_assertf (tailcall || !mini_get_debug_options ()->test_tailcall_require, "tailcalllog fail from %s", cfg->method->name);
-	if (cfg->verbose_level)
+	if (cfg->verbose_level > 1 || mini_get_debug_options ()->tailcall)
```
luhenry: We should use `mono_trace` instead, it's exactly done for these kind of use cases.

jaykrell: I see printf and g_print a lot and was unaware. Let me look more then.

jaykrell: MONO_LOG_LEVEL=debug MONO_LOG_MASK=tailcall and I fill in flag_mask_map? I'll try that.
ok to also support -v -v, or we should stop adding anything under verbose_level?

jaykrell: oh, no, can't really have -v -v also, that'd be messy. just mono_trace.

jaykrell: notice method-to-ir.c never uses mono_trace..thus my mistake

PR #7991, opened 2018-04-03
[genproj] Add Mono.CSharp.pre and fix genproj postbuild bug

akoeplinger:
We weren't emitting the postbuild script.


akoeplinger:
Nope, genproj.cs looks it up based on libraryName + ".pre"

PR #8003, opened 2018-04-04
[interp] workaround undefined behaviour in r8 to u1 cast

lewurm:
This fixes `test_0_simple_double_casts` on iOS/arm64.
According to the C standard (section [conv.fpint]), the behaviour of
such a cast is undefined if the truncated value doesn't fit into the
destination type.

```c
unsigned char __attribute__ ((noinline)) convr8u1 (double i)
{
        return (unsigned char) i;
}
```

for arm64, `clang -O0` produces (works as you would expect):

```
_convr8u1:
100006848:  ff 43 00 d1     sub sp, sp, #16
10000684c:  e0 07 00 fd     str d0, [sp, #8]
100006850:  e0 07 40 fd     ldr d0, [sp, #8]
100006854:  08 00 78 1e     fcvtzs  w8, d0
100006858:  00 1d 00 53     uxtb    w0, w8
10000685c:  ff 43 00 91     add sp, sp, #16
100006860:  c0 03 5f d6     ret
```

while `clang -O2` produces:

```
_convr8u1:
100006ae8:  00 00 78 1e     fcvtzs  w0, d0
100006aec:  c0 03 5f d6     ret
```

which is fine according the standard: Any value < 256 gets casted just
fine, but for example `convr8u1 ((double) 0xffff)` would end up with
`0xffff` in `w0`.

We workaround this behaviour by casting to a 32bit int first, the
resulting code looks like this then:

```
_convr8u1:
100006ae4:  08 00 78 1e     fcvtzs  w8, d0
100006ae8:  00 1d 00 12     and w0, w8, #0xff
100006aec:  c0 03 5f d6     ret
```

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3332,7 +3332,14 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, guint16 *st
 			++ip;
 			MINT_IN_BREAK;
 		MINT_IN_CASE(MINT_CONV_I1_R8)
-			sp [-1].data.i = (gint8)sp [-1].data.f;
+			/* without gint32 cast, C compiler is allowed to use undefined
+			 * behaviour if data.f is bigger than >255. See conv.fpint section
+			 * in C standard:
+			 * > The conversion truncates; that is, the fractional  part
+			 * > is discarded.  The behavior is undefined if the truncated
+			 * > value cannot be represented in the destination type.
+			 * */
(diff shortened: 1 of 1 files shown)


lewurm:
needs backport to `2018-02` and `2018-04`

lewurm:
@monojenkins backport 2018-02

lewurm:
@monojenkins backport 2018-04

lewurm:
green build without a single rebuild, what the heck

PR #8014, opened 2018-04-04
[coop] Remove AbortBlockingIgnoreAndPoll

lambdageek:
This is used to indicate that `mono_threads_transition_abort_blocking` was called
by a thread in the `STATE_ASYNC_SUSPEND_REQUESTED` state.  But we have no
transitions from `STATE_BLOCKING` to `STATE_ASYNC_SUSPEND_REQUESTED` (as
`mono_threads_transition_request_async_suspension` just increments the suspend
count when in `STATE_BLOCKING`), therefore abort is always a programmer error.

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -352,9 +352,6 @@ mono_threads_enter_gc_unsafe_region_unbalanced_with_info (MonoThreadInfo *info, 
 	case AbortBlockingIgnore:
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		return NULL;
-	case AbortBlockingIgnoreAndPoll:
-		mono_threads_state_poll_with_info (info);
-		return NULL;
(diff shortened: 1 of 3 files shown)


lambdageek:
`Cannot transition thread 0xf51ffb40 from ASYNC_SUSPEND_REQUESTED with DONE_BLOCKING`
  (╯°□°）╯︵ ┻━┻

Can't tell yet if removing the `abort_blocking` `ASYNC_SUSPEND_REQUESTED` case was wrong or if it just exposed some underlying issue.

lambdageek:
This is wrong.

We always execute `mono_threads_transition_abort_blocking` for `MONO_ENTER_GC_UNSAFE` in a victim thread - even when the victim is already in `RUNNING`.  Which means that we could race with another thread calling `mono_threads_request_async_suspend` on the victim thread - which means we could be in `ASYNC_SUSPEND_REQUESTED` when the victim tries to run `mono_threads_transition_abort_blocking`.

So we need all this code.

PR #8015, opened 2018-04-04
[SDKS] Fix Android SDK build on Linux

grendello:
Current build uses `tar` to extract ZIP archives. This works with BSD
tar but not with GNU tar, alas. Tar was used in order to strip the root
directory of Android archive while moving files into place. This commit
replaces `tar` with a call to a shell script (because doing what it does
using Make defines is an exercise in pain) which works both on macOS/BSD
and on Linux.


luhenry:

luhenry:

luhenry:

Issue #7356, opened 2018-03-01
Memory leak in HttpWebRequest / HttpWebResponse


smr888:
## Steps to Reproduce

1. Use Ubuntu 16.04.  Install mono-devel, mono-utils. mono-profiler, whatever is necessary to have the compiler and profiler.  I doubt this is specific to the Linux version, however.

2. The program below simply loops over a set of popular websites and gets their homepages using HttpWebRequest and HttpWebResponse.  Compile and run it under the profiler, e.g.

```
mono --gc=sgen --profile=log:heapshot=50gc TestMemGrow.exe
```

3.  Let it run for 15 - 30 minutes.  After that, to view the profiler output enter:

```
mprof-report output.mlpd > out.txt
grep System.Byte out.txt
```

You should see a list of allocations of System.Byte[] that grows increasingly larger.  You can leave the program running and re-execute those two commands, to see the continued growth of memory usage as the profiler adds more heap shots to its data file.

I put in the GC calls to see what would happen, and to see if the problem was only in the large object space.  It appears not to be limited to those objects.  If you take the GC calls out, memory usage fluctuates quite a bit up and down as the collections take place, and possibly grows more rapidly.  The line of code that actually reads the response stream and writes the web page data to a file is currently commented out. If you comment that back in, memory use grows more rapidly than without it.

**CODE:**

```C#
using System;
using System.Collections.Generic;
using System.IO;
using System.Net;
using System.Runtime;
using System.Threading;

namespace TestMemGrow
{
    class Program
    {
        static void Main()
        {
            List<string> websites = new List<string>() { "http://www.amazon.com", "http://www.google.com", "http://www.yahoo.com", "http://www.ebay.com", "http://www.overstock.com",
                "http://www.marketwatch.com", "http://www.bn.com", "http://www.newegg.com", "http://www.wsj.com", "http://www.arstechnica.com", "http://www.slashdot.org",
                "http://www.mediaite.com", "http://www.disqus.com", "http://www.twitter.com", "http://www.snap.com", "http://www.facebook.com", "http://www.usps.com",
                "http://www.ups.com", "http://www.techcrunch.com", "http://www.oracle.com", "http://www.java.com", "http://www.apple.com", "http://www.microsoft.com",
                "http://www.ibm.com", "http://www.dell.com", "http://www.asus.com", "http://www.gigabyte.com", "http://www.intel.com", "http://www.crucial.com",
[...]

PR #8016, opened 2018-04-04
[System]: Fix memory leak in the web stack (partial fix of #7356).

baulig:
Using `Task.Delay (timeout)` has the potential of possibly allocating and holding large
chunks of memory throughout the timeout period.  While this is not strictly speaking a
memory leak because those objects will eventually be disposed and garbage collected, it
can cause memory usage of apps to grow quite significantly in the short run.

The default timeout values are as follows:

* `HttpWebRequest.ReadWriteTimeout`: 5 minutes.

* `HttpWebRequest.Timeout`: 100 seconds.

* `ServicePointManager.MaxServicePointIdleTime`: 100 seconds.

And these values can be increased by the user.  This means that we're potentially holding
hundreds, maybe even thousands of timeout objects in memory for 5+ minutes.

This has now been fixed by using the `Task.Delay(int,CancellationToken)` constructor.

Note that disposing a `CancellationTokenSource` will not cancel it, so the following code
will still leak:

	using (var cts = new CancellationTokenSource ()) {
		Task.Delay (TimeSpan.FromDays (3), cts.Token);
		// Work
	}

To actually make the `Task.Delay` instance go away when we're done with it, we need to
use this pattern:

	var cts = new CancellationTokenSource ();
	try {
		Task.Delay (TimeSpan.FromDays (3), cts.Token);
		// Work
	} finally {
		cts.Cancel ();
		cts.Dispose ();
	}

We also cannot use `CancellationToken.CancelAfter()` for timeouts because unfortunately
some of the underlying framework code that we're using does not honor cancellation
requests - like for instance `NetworkStream`.

This is a partial fix for #7356.  While this commit significantly improves the problem,
there may be other unrelated issues left that needs to be investigated.

### DIFF
--- a/mcs/class/System/System.Net/ServicePointScheduler.cs
+++ b/mcs/class/System/System.Net/ServicePointScheduler.cs
@@ -442,6 +442,19 @@ namespace System.Net
 			Interlocked.Decrement (ref currentConnections);
 		}
 
+		public static async Task<bool> WaitAsync (Task workerTask, int millisecondTimeout)
+		{
+			var cts = new CancellationTokenSource ();
+			try {
+				var timeoutTask = Task.Delay (millisecondTimeout, cts.Token);
+				var ret = await Task.WhenAny (workerTask, timeoutTask).ConfigureAwait (false);
+				return ret != timeoutTask;
+			} finally {
+				cts.Cancel ();
+				cts.Dispose ();
+			}
+		}
+
 		class ConnectionGroup
 		{
 			public ServicePointScheduler Scheduler {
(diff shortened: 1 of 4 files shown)


marek-safar:
@monojenkins backport 2018-02

marek-safar:
@monojenkins backport 2018-04

mcs/class/System/System.Net/ServicePointScheduler.cs
```diff
@@ -442,6 +442,19 @@ void OnConnectionClosed (WebConnection connection)
 			Interlocked.Decrement (ref currentConnections);
 		}
 
+		public static async Task<bool> WaitAsync (Task workerTask, int millisecondTimeout)
+		{
+			var cts = new CancellationTokenSource ();
```
marek-safar: Would be nice to optimize for Infinite timeout

pr0vieh:
can we get this in master also?
this really improves performance without loss of quality
but this not reach the performance before btls implementation
but I think more improvements incoming

PR #8018, opened 2018-04-04
[coop] We never return SelfSuspendWait and AsyncSuspendWait

lambdageek:
These are the last two uses of `SelfSuspendWait` and `AsyncSuspendWait`


lambdageek:
@monojenkins build failed

lambdageek:
@monojenkins rebase

PR #8020, opened 2018-04-04
Silently ignore mono_gchandle_free(0) so callers do not have to check.

jaykrell:
Generally this makes for smaller easier to maintain code.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -854,10 +854,8 @@ mono_thread_detach_internal (MonoInternalThread *thread)
 	mono_w32mutex_abandon (thread);
 #endif
 
-	if (thread->abort_state_handle) {
-		mono_gchandle_free (thread->abort_state_handle);
-		thread->abort_state_handle = 0;
-	}
+	mono_gchandle_free (thread->abort_state_handle);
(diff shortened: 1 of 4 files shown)


mono/metadata/threads.c
```diff
@@ -2543,13 +2539,11 @@ mono_thread_internal_reset_abort (MonoInternalThread *thread)
...
 		mono_get_eh_callbacks ()->mono_clear_abort_threshold ();
 		thread->abort_exc = NULL;
-		if (thread->abort_state_handle) {
-			mono_gchandle_free (thread->abort_state_handle);
-			/* This is actually not necessary - the handle
-			   only counts if the exception is set */
-			thread->abort_state_handle = 0;
-		}
 	}
+	mono_gchandle_free (thread->abort_state_handle);
```
vargaz: This was inside an if () before, but not any more.

jaykrell: Deliberate but I can put it back.

jaykrell: oh..deliberate..and suspicious, got it.

jaykrell:
Ping?

jaykrell:
ping?

PR #8021, opened 2018-04-05
[interp] do not bake object reference into code stream

lewurm:
it won't be scanned by the GC, defer string allocation to execution-time instead.

this fixes a GC crash when running mini regression suite on iOS/arm64.


lewurm:
needs backporting to `2018-02` and `2018-04`.

lewurm:
@monojenkins backport 2018-02

lewurm:
@monojenkins backport 2018-04

lewurm:
@monojenkins rebase

PR #8025, opened 2018-04-05
Allow the LLVM optimizer options to be passed on the command line.

migueldeicaza:
Just to make it simpler to try different LLVM optimization flags.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -205,6 +205,7 @@ typedef struct MonoAotOptions {
 	char *temp_path;
 	char *instances_logfile_path;
 	char *logfile;
+	char *llvm_opts;
 	gboolean dump_json;
 	gboolean profile_only;
 } MonoAotOptions;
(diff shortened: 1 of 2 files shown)


mono/mini/aot-compiler.c
```diff
@@ -8990,6 +8993,9 @@ emit_llvm_file (MonoAotCompile *acfg)
 #else
 		opts = g_strdup ("-targetlibinfo -no-aa -basicaa -notti -instcombine -simplifycfg -inline-cost -inline -sroa -domtree -early-cse -lazy-value-info -correlated-propagation -simplifycfg -instcombine -simplifycfg -reassociate -domtree -loops -loop-simplify -lcssa -loop-rotate -licm -lcssa -loop-unswitch -instcombine -scalar-evolution -loop-simplify -lcssa -indvars -loop-idiom -loop-deletion -loop-unroll -memdep -gvn -memdep -memcpyopt -sccp -instcombine -lazy-value-info -correlated-propagation -domtree -memdep -adce -simplifycfg -instcombine -strip-dead-prototypes -domtree -verify");
 #endif
+	if (acfg->llvm_opts != NULL)
```
luhenry: This will leak memory, and will override the default one. If overriding is the expected behavior, we should put lines 8987 to 8995 in the else branch of this `acfg->llvm_opts != NULL`

migueldeicaza: Well, I could point out that this is a tiny leak on a one-off build, this is not a long running process.

Anyways, patch attached.

luhenry:
@monojenkins build failed

PR #8037, opened 2018-04-05
Remove System.Array.SetGenericValueImpl and InternalArray__set_Item<T>.

jaykrell:
They are not used. They are only reachable via reflection to non-public functions.

### DIFF
--- a/mcs/class/corlib/System/Array.cs
+++ b/mcs/class/corlib/System/Array.cs
@@ -180,26 +180,9 @@ namespace System
 			return value;
 		}
 
-		internal void InternalArray__set_Item<T> (int index, T item)
-		{
-			if (unchecked ((uint) index) >= unchecked ((uint) Length))
-				throw new ArgumentOutOfRangeException ("index");
-
-			object[] oarray = this as object [];
-			if (oarray != null) {
-				oarray [index] = (object)item;
-				return;
-			}
-			SetGenericValueImpl (index, ref item);
-		}
-
 		// CAUTION! No bounds checking!
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		internal extern void GetGenericValueImpl<T> (int pos, out T value);
-
-		// CAUTION! No bounds checking!
-		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		internal extern void SetGenericValueImpl<T> (int pos, ref T value);
 
 		internal struct InternalEnumerator<T> : IEnumerator<T>
 		{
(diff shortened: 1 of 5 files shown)


jaykrell:
There might be something to this set_Item thing. Maybe best to leave it. But then coop/handle conversion must be correct.

mcs/class/corlib/System/Array.cs
```diff
@@ -180,27 +180,10 @@ internal T InternalArray__get_Item<T> (int index)
 			return value;
 		}
 
-		internal void InternalArray__set_Item<T> (int index, T item)
-		{
-			if (unchecked ((uint) index) >= unchecked ((uint) Length))
-				throw new ArgumentOutOfRangeException ("index");
-
-			object[] oarray = this as object [];
```
lambdageek: Jay, I was wrong, all this stuff is live.

Change `InternalArray__set_Item<T>` to be:

```csharp
			Console.WriteLine ("I was called!!!");
			if (unchecked ((uint) index) >= unchecked ((uint) Length))
				throw new ArgumentOutOfRangeException ("index");

			object[] oarray = this as object [];
			if (oarray != null) {
				Console.WriteLine ("Successful cast!");
				oarray [index] = (object)item;
				return;
			}
			Console.WriteLine ("Sad cast :-(");
			SetGenericValueImpl (index, ref item);
```

And run this example:
```csharp
public struct P {
	public int x;
	public int y;
	public static P[] MakePs () {
		var ps = new P[10];
		for (int i = 0; i < ps.Length; i++) {
			ps[i].x = 1+i;
			ps[i].y = 1-i;
		}
		return ps;
	}
};

public class M {
	public static P[] M_naive () {
		P h;
		h.x = 42;
		h.y = 99;
		P[] ps = P.MakePs();
		(ps as System.Collections.Generic.IList<P>)[0] = h;
		return ps;
	}
	public static void Main () {
		P[] ps = M_naive ();
		Console.WriteLine ("ps[0] = [x = {0}, y = {1}]", ps[0].x, ps[0].y);
      }
}
```

The output is:
```
I was called!!!
Sad cast :-(
pval_naive[0] = [x = 42, y = 99]
```

if you if you comment out this bit in method-to-ir.c you can see the runtime hit the SetGenericValueImpl icall:
```c
		else if (strcmp (cmethod->name, "SetGenericValueImpl") == 0 && fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt)
			return emit_array_generic_access (cfg, fsig, args, TRUE);
```

PR #8038, opened 2018-04-05
[SDKS] fix Linux (and possibly other OSes) build for Android

grendello:
OS: Ubuntu 18.03
CMake: 3.10

For some reason cmake fails to detect any make program (be it ninja or make)
when cross-compiling for Android in XA. It happens despite the programs being in
$PATH and working correctly. The effect is that CMAKE_MAKE_PROGRAM is set to
"not found" in CMakeCache.txt and configuration fails as follows:
```
 Call Stack (most recent call first):
   external/mono/external/boringssl/util/android-cmake/android.toolchain.cmake:1140 (CMAKE_FORCE_CXX_COMPILER)
   external/mono/sdks/builds/android-armeabi-v7a-debug/mono/btls/build-shared/CMakeFiles/3.10.2/CMakeSystem.cmake:6 (include)
      CMakeLists.txt:3 (project)
   CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
      -- Configuring incomplete, errors occurred!
```
This commit does not fix the issue, per se, but it works around it by finding
the first ninja/make (as configured) executable in $PATH and setting the
CMAKE_MAKE_PROGRAM to full path of the executable. I haven't spent too much time
investigating *why* cmake fails to find the executable (I merely ran it with
--trace-expand --debug-output which revealed nothing of consequence) but this
workaround should be safe for any version of CMake and any OS we use makefiles
on.

### DIFF
--- a/mono/btls/Makefile.am
+++ b/mono/btls/Makefile.am
@@ -44,12 +44,14 @@ NINJA_VERBOSE=$(if ($V),-v,)
 if NINJA
 NINJA_ARGS = -G Ninja
 BUILDFILE = build.ninja
+CMAKE_MAKE_PROGRAM = $(shell which ninja-build || which ninja)
 else
 NINJA_ARGS =
 BUILDFILE = Makefile
+CMAKE_MAKE_PROGRAM = $(shell which gmake || which gnumake || which make)
 endif
 
-CMAKE_ARGS = -D CMAKE_INSTALL_PREFIX:PATH=$(prefix) -D BTLS_ROOT:PATH=$(BTLS_ROOT) \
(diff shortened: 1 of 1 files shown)


grendello:
> We have a --enable-ninja configure parameter. I know it didn't fix the issue but could you please provide the output you get when passing that?
> 
> Could you also share the log of cmake so we can investigate more.
> 

The output is identical with the difference that instead of `Ninja` CMake shows `Unix Makefile`

jaykrell:
We might be able to remove -enable-ninja with this change?

grendello:
@luhenry [full cmake log](https://git.io/vxyoK) (with `--trace-expand --debug-output`), full paths to the source dir truncated to remove some noise

luhenry:

luhenry:

luhenry:

PR #6672, opened 2018-01-25
[corlib] CoreRT System.Threading.Tasks

EgorBo:
Replaces `System.Threading.Tasks` implementation with corefx+corert.
WIP
Requires changes in the corert and corefx submodules to compile.

Implements #6967

PR #8040, opened 2018-04-05
[corlib] ConcurrentExclusiveSchedulerPair from corefx

EgorBo:
These types were not replaced with corefx in https://github.com/mono/mono/pull/6672


luhenry:
@marek-safar @EgorBo there is a change in attribute for `ConcurrentExclusiveSchedulerPair` in `net_4_x`, `monodroid` and `monotouch` profiles, does it matter? See https://jenkins.mono-project.com/job/test-mono-pull-request-amd64/13335//Public_API_Diff/

mcs/class/corlib/corlib.dll.sources
```diff
@@ -1586,11 +1586,10 @@ ReferenceSources/AppContextDefaultValues.cs
...
 
 ../referencesource/mscorlib/system/threading/Tasks/AsyncCausalityTracer.cs
-
-../referencesource/mscorlib/system/threading/Tasks/TaskExceptionHolder.cs
-../referencesource/mscorlib/system/threading/Tasks/ConcurrentExclusiveSchedulerPair.cs
-../referencesource/mscorlib/system/threading/Tasks/ProducerConsumerQueues.cs
-../referencesource/mscorlib/system/threading/Tasks/TaskToApm.cs
+../../../external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/TaskExceptionHolder.cs
+../../../external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/ConcurrentExclusiveSchedulerPair.cs
+../../../external/corefx/src/System.Threading.Tasks.Dataflow/src/Internal/ProducerConsumerQueues.cs
```
akoeplinger: can we use external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/ProducerConsumerQueues.cs instead of the copy in Dataflow?

EgorBo: @akoeplinger well, those are almost the same (@marek-safar said if a type is presented in both corefx and corert I should prefer corefx). The only difference - `MultiProducerMultiConsumerQueue` subclasses `LowLevelConcurrentQueue` in corert instead of `ConcurrentQueue` in corefx. Which is a new dependency.

akoeplinger: That's a fine rule of thumb but in this case the corefx ProducerConsumerQueues.cs is a private source copy in a completely unrelated (and essentially "maintenance mode") assembly.

I feel we should be using the matching implementation from corert since that's where the dependency to ProducerConsumerQueues comes from.

We could #ifdef the subclassing change and use `ConcurrentQueue` on Mono.

marek-safar: In this case `../../../external/corefx/src/System.Threading.Tasks.Dataflow/src/Internal/ProducerConsumerQueues.cs` really looks like some old version. For now we should use corert and fill corefx issue about moving ProducerConsumerQueues.cs to common location to share it acress the repos

marek-safar:
@monojenkins commit apidiff

mcs/class/corlib/corlib.dll.sources
akoeplinger: @EgorBo I saw that you bumped corert with a ProducerConsumerQueues.cs change but this still points to the corefx file, could you please adjust it? :)

akoeplinger:
@monojenkins commit csproj

PR #8041, opened 2018-04-05
Report the proper socket error

joshpeterson:
When a connection fails, the socket error was being incorrectly reported
as an address family not supported error, instead of connection refused.

Why did this happen? The order of addresses returned from
`Dns.GetHostAddresses` changed. The newer code returns the IPv6 address
first, and the IPv4 address second. While `GetCheckedIPs` mentioned in a
comment that it skips addresses that don't match the address family, it
did not. The code using the addresses returned by `GetCheckedIPs` will
use the first address in the array. If this address happens to match the
address family of the socket, things will work "properly" (the correct
socket error will occur).

If, however the first entry has the wrong address family, then the
socket error will be an address family not supported error, since the
socket was created with different address family.

This change modifies `GetCheckedIPs` to filter the addresses, removing
those which do not match the address family.

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -1168,6 +1168,16 @@ namespace System.Net.Sockets
 			DnsEndPoint dep = e.RemoteEndPoint as DnsEndPoint;
 			if (dep != null) {
 				addresses = Dns.GetHostAddresses (dep.Host);
+				int last_valid = 0;
+				for (int i = 0; i < addresses.Length; ++i) {
+					if (addresses [i].AddressFamily != dep.AddressFamily)
+						continue;
+
+					addresses [last_valid++] = addresses [i];
+				}
+
+				if (last_valid != addresses.Length)
+					Array.Resize (ref addresses, last_valid);
 				return true;
 			} else {
 				e.ConnectByNameError = null;


mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -1167,7 +1167,19 @@ bool GetCheckedIPs (SocketAsyncEventArgs e, out IPAddress [] addresses)
 			// while skipping entries that do not match the address family
 			DnsEndPoint dep = e.RemoteEndPoint as DnsEndPoint;
 			if (dep != null) {
-				addresses = Dns.GetHostAddresses (dep.Host);
+				var possibleAddresses = Dns.GetHostAddresses (dep.Host);
```
marek-safar: I don't think we need to allocate 2 additional arrays. It should be enough to shuffle the items in the original array and resize it at the end (if necessary)

joshpeterson: I agree that would improve performance, but I think it would make the code more difficult to understand. I'd lean toward easier to understand code here, as I don't see this as being performance critical.

joshpeterson:
@pr0vieh: Note that this change is not to the implementation of `Dns.GetHostAddresses`, but rather to a consumer of the information that method provides. It should not impact the performance of `Dns.GetHostAddresses`.

pr0vieh:
this is probably true but our environment can not handle IPv6 currently!
therefore it probably just generates bad urls because this bug exists
That's why I think it's relevant to our environment

but 👍 for fixing it

mcs/class/System/System.Net.Sockets/Socket.cs
marek-safar: I don't this so. Typing blindly but following code should do same

```csharp
		var possibleAddresses = Dns.GetHostAddresses (dep.Host);
		int last_valid = 0;
		for (int i = 0; i < possibleAddresses.Length; ++i) {
			if (possibleAddresses [i].AddressFamily != dep.AddressFamily)
				continue;

			possibleAddresses [last_valid++] = possibleAddresses [i];
		}

		if (last_valid != possibleAddresses.Length)
			Array.Resize (ref possibleAddresses, last_valid);
```

and it looks to me to be easier to understand

joshpeterson: Ok, I'll go with a change similar to this one. Thanks!

PR #8051, opened 2018-04-05
[corefx] Build System.Native

luhenry:


configure.ac
```diff
@@ -2827,6 +2829,560 @@ if test x$host_win32 = xno; then
...
+		],[
+			AC_MSG_RESULT(yes)
+			AC_DEFINE(HAVE_FLOCK64, 1, struct flock64)
+		], [
+			AC_MSG_RESULT(no)
+		])
+	fi
+
+	AC_CHECK_FUNC(lseek64,           [AC_DEFINE(HAVE_LSEEK64, 1, [lseek64])])
+	AC_CHECK_FUNC(mmap64,            [AC_DEFINE(HAVE_MMAP64, 1, [mmap64])])
```
vargaz: Using AC_CHECK_FUNCS(....) would be simpler.

configure.ac
```diff
@@ -2827,6 +2829,560 @@ if test x$host_win32 = xno; then
...
+	AC_CHECK_FUNC(sched_setaffinity, [AC_DEFINE(HAVE_SCHED_SETAFFINITY, 1, [sched_setaffinity])])
+	AC_CHECK_DECL(TIOCGWINSZ,        [AC_DEFINE(HAVE_TIOCGWINSZ, 1, [TIOCGWINSZ])], [], [[#include <sys/ioctl.h>]])
+	AC_CHECK_FUNC(tcgetattr,         [AC_DEFINE(HAVE_TCGETATTR, 1, [tcgetattr])])
+	AC_CHECK_FUNC(tcsetattr,         [AC_DEFINE(HAVE_TCSETATTR, 1, [tcsetattr])])
+	AC_CHECK_DECL(ECHO,              [AC_DEFINE(HAVE_ECHO, 1, [ECHO])], [], [[#include <termios.h>]])
+	AC_CHECK_DECL(ICANON,            [AC_DEFINE(HAVE_ICANON, 1, [ICANON])], [], [[#include <termios.h>]])
+	AC_CHECK_DECL(TCSANOW,           [AC_DEFINE(HAVE_TCSANOW, 1, [TCSANOW])], [], [[#include <termios.h>]])
+
+	AC_CHECK_MEMBER(struct stat.st_birthtimespec, [AC_DEFINE(HAVE_STAT_BIRTHTIME, 1, [struct stat.st_birthtime])], [], [[#include <sys/types.h>], [#include <sys/stat.h>]])
+	AC_CHECK_MEMBER(struct stat.st_atimespec,     [AC_DEFINE(HAVE_STAT_TIMESPEC, 1, [struct stat.st_atimespec])], [], [[#include <sys/types.h>], [#include <sys/stat.h>]])
```
vargaz: Same for AC_CHECK_MEMBERS.

configure.ac
```diff
@@ -738,9 +738,9 @@ else
 		;;
 	esac
 fi
-CFLAGS="$CFLAGS -g $WARN"
+CFLAGS="$WARN $CFLAGS -g"
 CFLAGS_FOR_LIBGC="$CFLAGS_FOR_LIBGC -g"
```
vargaz: Why is this needed ? The usual way of adding flags is to append to them.

mono/metadata/Makefile.am
```diff
@@ -89,7 +89,11 @@ if !ENABLE_ILGEN
 ilgen_libraries = libmono-ilgen.la
 endif
 
-noinst_LTLIBRARIES = libmonoruntime-config.la $(boehm_libraries) $(sgen_libraries)
+if !HOST_WIN32
+sysnat_libraries = libmonoruntime-sysnat.la
+endif
```
vargaz: It would be easier to just add them to libmonoruntime, having convenience libs just slows down compilation.

configure.ac
jaykrell: Personally I like a line per "thing" for easier to see diffs (whatever is a "thing"). Instead of edits to long lines where you have to find the intra-line diff. For example there are coding conventions with a parameter per line. But conserving vertical space is also a goal.

I suspect this got copied from elsewhere.

mono/metadata/Makefile.am
luhenry: We have different CFLAGS for `libmonoruntime-sysnat.la` than for `libmonoruntime.la`.

configure.ac
luhenry: Otherwise we have the following command line being executed for `pal_io.c` for example:

```
echo "  CC      " ../../external/corefx/src/Native/Unix/System.Native/libmonoruntime_sysnat_la-pal_io.lo;../../doltlibtool --silent --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../..  -I/Users/luhenry/git/mono/external/corefx/src/Native/Unix/Common -I/Users/luhenry/git/mono/external/corefx/src/Native/Unix/System.Native -Wno-typedef-redefinition  -Wno-missing-prototypes -D_THREAD_SAFE -DGC_MACOSX_THREADS -DUSE_MMAP -DUSE_MUNMAP -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -Qunused-arguments -Wno-unused-function -Wno-tautological-compare -Wno-parentheses-equality -Wno-self-assign -Wno-return-stack-address -Wno-constant-logical-operand -Wno-zero-length-array   -O0 -g -std=gnu99 -fno-strict-aliasing -fwrapv -DMONO_DLL_EXPORT -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -Qunused-arguments -Wno-unused-function -Wno-tautological-compare -Wno-parentheses-equality -Wno-self-assign -Wno-return-stack-address -Wno-constant-logical-operand -Wno-zero-length-array -Werror-implicit-function-declaration -MT ../../external/corefx/src/Native/Unix/System.Native/libmonoruntime_sysnat_la-pal_io.lo -MD -MP -MF ../../external/corefx/src/Native/Unix/System.Native/.deps/libmonoruntime_sysnat_la-pal_io.Tpo -c -o ../../external/corefx/src/Native/Unix/System.Native/libmonoruntime_sysnat_la-pal_io.lo `test -f '../../external/corefx/src/Native/Unix/System.Native/pal_io.c' || echo './'`../../external/corefx/src/Native/Unix/System.Native/pal_io.c
  CC       ../../external/corefx/src/Native/Unix/System.Native/libmonoruntime_sysnat_la-pal_io.lo
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:180:19: warning: 'stat64' is deprecated: first deprecated in macOS 10.6 [-Wdeprecated-declarations]
    while ((ret = stat_(path, &result)) < 0 && errno == EINTR);
                  ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:41:15: note: expanded from macro 'stat_'
#define stat_ stat64
              ^
/usr/include/sys/stat.h:402:5: note: 'stat64' has been explicitly marked deprecated here
int     stat64(const char *, struct stat64 *) __OSX_AVAILABLE_BUT_DEPRECATED(__MAC_10_5,__MAC_10_6,__IPHONE_NA,__IPHONE_NA);
        ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:176:9: warning: no previous prototype for function 'SystemNative_Stat2' [-Wmissing-prototypes]
int32_t SystemNative_Stat2(const char* path, struct FileStatus* output)
        ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:194:19: warning: 'fstat64' is deprecated: first deprecated in macOS 10.6 [-Wdeprecated-declarations]
    while ((ret = fstat_(ToFileDescriptor(fd), &result)) < 0 && errno == EINTR);
                  ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:42:16: note: expanded from macro 'fstat_'
#define fstat_ fstat64
               ^
/usr/include/sys/stat.h:400:5: note: 'fstat64' has been explicitly marked deprecated here
int     fstat64(int, struct stat64 *) __OSX_AVAILABLE_BUT_DEPRECATED(__MAC_10_5,__MAC_10_6,__IPHONE_NA,__IPHONE_NA);
        ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:190:9: warning: no previous prototype for function 'SystemNative_FStat2' [-Wmissing-prototypes]
int32_t SystemNative_FStat2(intptr_t fd, struct FileStatus* output)
        ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:207:15: warning: 'lstat64' is deprecated: first deprecated in macOS 10.6 [-Wdeprecated-declarations]
    int ret = lstat_(path, &result);
              ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:43:16: note: expanded from macro 'lstat_'
#define lstat_ lstat64
               ^
/usr/include/sys/stat.h:401:5: note: 'lstat64' has been explicitly marked deprecated here
int     lstat64(const char *, struct stat64 *) __OSX_AVAILABLE_BUT_DEPRECATED(__MAC_10_5,__MAC_10_6,__IPHONE_NA,__IPHONE_NA);
        ^
../../external/corefx/src/Native/Unix/System.Native/pal_io.c:204:9: warning: no previous prototype for function 'SystemNative_LStat2' [-Wmissing-prototypes]
int32_t SystemNative_LStat2(const char* path, struct FileStatus* output)
        ^
6 warnings generated.
```

What you can note is we define `-Wno-missing-prototypes` in `libmonoruntime_sysnat_la_CPPFLAGS` but it gets overwritten by `$WARN` at https://github.com/mono/mono/pull/8051#discussion-diff-179934409L741. By changing the order of `$CFLAGS` and `$WARN`, we don't get this warning anymore.

configure.ac
luhenry: We need to control exactly what's the name of the defined variable since it needs to match the corefx one, and `AC_CHECK_MEMBERS` wouldn't allow for it.

luhenry:
@monojenkins build Linux x64

luhenry:
@monojenkins build failed

luhenry:
Note: the `System.Net.NetworkInformation` integration from corefx is mostly to have a test case to make sure the `System.Native` integration actually work, I don't think it would be used by anything otherwise. So if there are too many issues with `System.Net.NetworkInformation`, we can just let the tests run with, and once it's green, we just remove it and merge.

mcs/class/System/Test/System.Net.NetworkInformation/IPInterfacePropertiesTest.cs
```diff
@@ -116,6 +116,8 @@ public void DnsEnabled ()
 		{
 			if (Environment.OSVersion.Platform == PlatformID.Win32NT)
 				Assert.Ignore ("IsDnsEnabled is not nessasarily enabled for all interfaces on windows.");
+			if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+				Assert.Ignore ("IsDnsEnabled is not supported on Unix.");
```
akoeplinger: It seems corefx has an implementation for this. It might throw a PlatformNotSupportedException in some cases but we can catch that and use Assert.Inconclusive instead of essentially disabling the test everywhere.

mcs/class/System/Test/System.Net.NetworkInformation/NetworkInterfaceTest.cs
```diff
@@ -70,6 +77,9 @@ public void FirstInterfaceOperationalStatus ()
 		[Test]
 		public void FirstInterfaceSpeed ()
 		{
+			if (RunningOnUnix)
+				Assert.Ignore ();
+
 			NetworkInterface[] adapters = NetworkInterface.GetAllNetworkInterfaces ();
 			Assert.IsTrue (adapters[0].Speed > 0);
```
akoeplinger: we seem to be just returning a hardcoded `1000000` in the Mono implementation while corefx always throws a NotImplementedException. I wonder how many apps that will break and if we need to do something about it?

mcs/class/System/darwin_System.dll.sources
```diff
@@ -0,0 +1 @@
+#include darwin_net_4_x_System.dll.sources
```
akoeplinger: This logic seems backwards to me. darwin_net_4_x_System.dll.sources should `#include darwin_System.dll.sources` (which contains all the source files) not the other way round if we want to have a common darwin file for other profiles.

marek-safar: Yeah, that's what we do elsewhere and should follow the same approach everywhere

mono/metadata/Makefile.am
```diff
@@ -147,6 +147,39 @@ libmono_ilgen_la_LIBADD = $(glib_libs) ../utils/libmonoutils.la ../sgen/libmonos
 endif
 endif
 
+if !HOST_WIN32
+if !HOST_ANDROID
+if !HOST_IOS
```
alexischr: As we are trying to enable the FSEvent FileSystemWatcher on `2018-04` either we need to build with `HOST_IOS`, or the two tiny ported functions needed should stay in `metadata/filewatcher.*` for now with `#ifdef  TARGET_IOS`

alexischr:

mcs/class/System/darwin_System.dll.sources
luhenry: Will fix that. Just FYI, these changes were done some months ago and this PR is just reapplying on top of master.

luhenry:

luhenry:

mcs/class/System/Test/System.Net.NetworkInformation/IPInterfacePropertiesTest.cs
```diff
@@ -117,12 +117,18 @@ public void DnsEnabled ()
...
+
 			NetworkInterface[] adapters = NetworkInterface.GetAllNetworkInterfaces ();
 			foreach (NetworkInterface adapter in adapters)
 			{
 				IPInterfaceProperties adapterProperties = adapter.GetIPProperties ();
 				Assert.IsTrue (adapterProperties.IsDnsEnabled);
 			}
+
+			} catch (PlatformNotSupportedException e) {
+				Assert.Inconclusive (e.ToString ());
```
marek-safar: On which platforms is it not supported?

luhenry:
@monojenkins commit apidiff

luhenry:

PR #8054, opened 2018-04-05
Throw MemberAccessException when abstract class is used with newobj instruction

joncham:
Previously an abstract instance would be created and potentially crash when vtable members were accessed.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -3641,6 +3641,14 @@ handle_alloc (MonoCompile *cfg, MonoClass *klass, gboolean for_box, int context_
 	MonoInst *iargs [2];
 	void *alloc_ftn;
 
+	if (mono_class_get_flags (klass) & TYPE_ATTRIBUTE_ABSTRACT) {
+		char* full_name = mono_type_get_full_name (klass);
+		mono_cfg_set_exception (cfg, MONO_EXCEPTION_MONO_ERROR);
+		mono_error_set_member_access (&cfg->error, "Cannot create an abstract class: %s", full_name);
+		g_free (full_name);
+		return NULL;
+	}
+
 	if (context_used) {
 		MonoInst *data;
 		MonoRgctxInfoType rgctx_info;
--- a/mono/utils/mono-error.c
+++ b/mono/utils/mono-error.c
@@ -428,6 +428,17 @@ mono_error_set_invalid_program (MonoError *oerror, const char *msg_format, ...)
 	set_error_message ();
 }
 
+void
+mono_error_set_member_access (MonoError *oerror, const char *msg_format, ...)
+{
+	MonoErrorInternal *error = (MonoErrorInternal*)oerror;
+
+	mono_error_prepare (error);
+	error->error_code = MONO_ERROR_MEMBER_ACCESS;
+
+	set_error_message ();
+}
+
 /**
  * mono_error_set_invalid_cast:
  *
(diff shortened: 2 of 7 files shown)


mono/mini/method-to-ir.c
```diff
@@ -3641,6 +3641,14 @@ handle_alloc (MonoCompile *cfg, MonoClass *klass, gboolean for_box, int context_
 	MonoInst *iargs [2];
 	void *alloc_ftn;
 
+	if (mono_class_get_flags (klass) & TYPE_ATTRIBUTE_ABSTRACT)	{
+		char* full_name = mono_type_get_full_name (klass);
+		cfg->exception_message = g_strdup_printf ("Cannot create an abstract class: %s", full_name);
+		g_free (full_name);
+		mono_cfg_set_exception (cfg, MONO_EXCEPTION_MEMBER_ACCESS);
```
lambdageek: Could do
```c
		mono_cfg_set_exception (cfg, MONO_EXCEPTION_MONO_ERROR);
		mono_error_set_generic_error (&cfg->error, "System", "MethodAccessException", "Cannot crerate an abstract class: %s", full_name);
		g_free (full_name);
```

For bonus points add `mono_error_set_member_access (MonoError *error, const char *msg_format, ...) MONO_ATTR_FORMAT_PRINTF(2,3)` to `utils/mono-error-internals.h` analogously to `mono_error_set_invalid_program` and others.

mono/utils/mono-error.c
```diff
@@ -588,6 +599,9 @@ mono_error_prepare_exception (MonoError *oerror, MonoError *error_out)
 	case MONO_ERROR_MISSING_FIELD:
 		exception = mono_corlib_exception_new_with_args ("System", "MissingFieldException", error->full_message, error->first_argument, error_out);
 		break;
+	case MONO_ERROR_MEMBER_ACCESS:
+		exception = mono_corlib_exception_new_with_args ("System", "MemberAccessException", error->full_message, error->first_argument, error_out);
```
lambdageek: I think this should be `mono_exception_from_name_msg (mono_defaults.corlib, "System", "MemberAccessException", error->full_message)`.  There's no `MemberAccessException(String,String)` constructor, afaict.

mono/mini/method-to-ir.c
```diff
@@ -3641,6 +3641,14 @@ handle_alloc (MonoCompile *cfg, MonoClass *klass, gboolean for_box, int context_
 	MonoInst *iargs [2];
 	void *alloc_ftn;
 
+	if (mono_class_get_flags (klass) & TYPE_ATTRIBUTE_ABSTRACT) {
+		char* full_name = mono_type_get_full_name (klass);
+		mono_cfg_set_exception (cfg, MONO_EXCEPTION_MONO_ERROR);
+		mono_error_set_member_access (&cfg->error, "Cannot crerate an abstract class: %s", full_name);
```
Therzok: typo: create

joncham:
@lambdageek I notice my test fails in AOT mode since the AOT compile fails and it's thus attempted to be JIT'd later on: https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot/9506/testReport/junit/MonoTests/runtime/newobj_abstract_exe/

lambdageek:
@joncham That makes sense. AOT turned it effectively into a compile-time error.

@vargaz @alexanderkyte Do we have a way of writing negative tests for AOT?

@joncham if we can't turn it into a negative test, just turn the test off for AOT - add it to somewhere around https://github.com/mono/mono/blob/master/mono/tests/Makefile.am#L1015

lambdageek:
@lewurm @BrzVlad Looks like the interpreter needs to handle newobj of an abstract class, too.

BrzVlad:
let's disable the test on interp

lewurm:
I pushed a fix for the interpreter

lambdageek:
I pushed to disable the test on FullAOT

lambdageek:
@monojenkins rebase

lambdageek:
@monojenkins build failed

PR #8057, opened 2018-04-05
[mono-api-html] Update tool to optionally output markdown

spouliot:
Output formatting is now delegate to a `Formatter` subclass. The default
one remains HTML with `HtmlFormatter` so any script should not see any
difference.

The output file is largely identical, except a few whitespace changes
that could not easily be avoided and how modified base types were
reported, since that was different from the rest.

The big change is the `--md` option which can output markdown, which
is the format to use from docs.microsoft.com. There are less options
available, e.g. colors, so additions are rendered using the `csharp`
blocks, while modification are rendered as a `diff` block.

Each assembly output is made to be easy to re-assemble into a single
document if needed, e.g. https://github.com/xamarin/xamarin-macios/pull/3814

### DIFF
--- a/mcs/tools/mono-api-html/MarkdownFormatter.cs
+++ b/mcs/tools/mono-api-html/MarkdownFormatter.cs
@@ -1,0 +1,199 @@
+﻿// 
+// Authors
+//    Sebastien Pouliot  <sebastien.pouliot@microsoft.com>
+//
+// Copyright 2018 Microsoft Inc.
+// 
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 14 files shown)


spouliot:
Diff between the current and new HTML output on xamarin-macios assemblies
https://gist.github.com/spouliot/2e36a0986cd8afc3d5363d471ab1b715

mcs/tools/mono-api-html/MarkdownFormatter.cs
```diff
@@ -0,0 +1,199 @@
...
+			} else {
+				output.WriteLine ("Added {0}:", list.Count () > 1 ? member.GroupName : member.ElementName);
+			}
+			output.WriteLine ();
+			output.WriteLine ("```csharp");
+		}
+
+		public override void AddMember (TextWriter output, MemberComparer member, bool isInterfaceBreakingChange, string obsolete, string description)
+		{
+			output.Write (obsolete);
```
marek-safar: You should check mosts values for special characters (e.g. backtick) and escape them

spouliot:
@rolfbjarne https://gist.github.com/spouliot/1c66b0f3431d3d2669e28aafdda56242
screen shot 2018-04-06 at 9 49 17 am

spouliot:
@marek-safar it should not be needed, this only happens inside blocks and the syntax to get out (of the block) cannot be in the formatted text, e.g. https://gist.github.com/spouliot/11ebc3fda871c3674bc6dc12dad4dc0e

alexischr:

PR #7994, opened 2018-04-03
[sdks] More fixes for running device tests

vargaz:

PR #7901, opened 2018-03-29
[sdks] add Make.config to ignore file

lewurm:

PR #8062, opened 2018-04-05
[2018-02] more iOS sdk backports

lewurm:
backport of
#7901
#7958
#7994
#8019


lewurm:
@vargaz please don't forget to backport your changes in the future, it slows down monthly integration

vargaz:
These don't need to be backported, they only affect the under development ios sdk lane which only runs on master.

PR #8066, opened 2018-04-05
Fix new warnings about missing prototypes.

jaykrell:


jaykrell:
@monojenkins build failed

PR #8068, opened 2018-04-06
[coop] Hybrid suspend

lambdageek:
## Summary

This is a new suspend mode where threads that are running managed or native runtime code (what we call "GC Unsafe") are suspended cooperatively using safepoints, but threads running either embedder code or P/Invoke code or blocking syscalls (what we call "GC Safe") are suspended preemptively using signals.

The motivation is to make an embedder-friendly coop suspend mechanism where the runtime will do GC Safe -> GC Unsafe transitions on calls to `MONO_API` functions, but the embedders themselves don't have to have any special knowledge of coop safepoints or state transitions.

To try this out, set **both** `MONO_ENABLE_COOP` and `MONO_ENABLE_HYBRID_COOP` environment variables.

## Implementation

Here's a picture of the new state machine.  There's two new states `Blocking_Suspend_Requested` and `Blocking_Async_Suspended` and one updated transition `req_s` (request suspension).

coop-state-machine

### State machine states and transitions

The hybrid cooperative suspend mechanism works like this:
 - when a thread is executing in GC Unsafe mode, we use cooperative suspend and expect
   the thread to periodically checkpoint its execution (as in the ordinary full
   coop mode).
 - when a thread is executing in GC Safe mode, we use a preemptive signal-based
   suspend.  This is new - previously in full coop we would allow `BLOCKING`
   threads to continue executing and only suspend them when they wanted to go
   from GC Safe to GC Unsafe (via the `BLOCKING_SELF_SUSPENDED` state).

There are two new states and one updated transition - the transition will
service both running and blocking threads: The idea is that suspend mechanism
is determined by the suspend initiator, and is not embodied in the the state
machine.  The state machine just has to return distinctive enough values from
the transitions for the initiator to select a policy.

Resume mechanism is determined by the state that a thread finds itself in when
it is resumed, and the original suspend policy.

The primary differences between suspend mechanisms are: the preemptive
suspend is two phase - the `finish_async_suspension` does not
apply for cooperative suspend; the cooperative policy uses the `poll` transition when the thread is running.

We renamed `mono_threads_transition_request_async_suspension` to `mono_threads_transition_request_suspension`.

 - It must be called by a suspend initiator on a victim thread that is not
   itself.

 - There can only be one suspend initiator for a victim at a time - if the
   suspend initiator initiates a suspension it must follow through with the
   whole protocol.

 - If a victim is `RUNNING` we transition it to `ASYNC_SUSPEND_REQUSTED` and return
   `ReqSuspendInitSuspendRunning` which signals that the caller must initiate
   suspension for a running thread.  (Same as the old `AsyncSuspendInitSuspend`).

 - If a victim is `BLOCKING` we transition it to `BLOCKING_SUSPEND_REQUESTED`.
   Note that this is different from the old `request_async_suspension` which just
   incremented the suspend count and returned `AsyncSuspendBlocking`.  Now we
   return `ReqSuspendInitSuspendBlocking`.  The return of
   `ReqSuspendInitSuspendBlocking` means the initiator may signal the victim and
   must wait to be notified of the suspension (in preemptive suspend it will be
   notified by the signal handler, in cooperative when the blocking thread
   attempts to exit from blocking mode).

 - In `BLOCKING_SUSPEND_REQUESTED` we
   may increment the suspend count.  This is slightly too lax if we're going to
   be using preemptive suspend on blocking threads: in that case we're in the
   middle of a two-phase suspend and since there is only one suspend initiator,
   we wouldn't expect another suspend to be initiated until we have a chance to
   finish suspending.  We leave it to the suspend initiator to rule out by returning
  `ReqSuspendAlreadySuspendedBlocking` from this state.

 - In all the not executing states (`SELF_SUSPENDED`, `ASYNC_SUSPENDED`,
  `BLOCKING_SELF_SUSPENDED`, `BLOCKING_ASYNC_SUSPENDED`) we just
   increment the suspend count.

 - All other transitions are illegal.

- Note that `BLOCKING` must always have a suspend count of 0 and
   `BLOCKING_SUSPEND_REQUESTED` must have suspend_count > 0.

We add two new states:

 - `STATE_BLOCKING_SUSPEND_REQUESTED` - this is a new state that indicates that a
   suspend initiator wants to suspend this victim thread.
   - The thread is still executing in this state.
   - We only transition into this state with
     `mono_threads_transition_request_suspension`.
   - The only legal transitions out from this state are:
     - `resume` - decrements the `suspend_count`. If the `suspend_count` becomes 0 we
       go back to `BLOCKING`, otherwise we stay in `BLOCKING_SUSPEND_REQUESTED`.
     - `finish_async_suspension` (executed by the suspend signal handler to
       finish the two phase preemptive suspend request and notify the suspend initiator),
     - `done_blocking` and `abort_blocking`.  It means that the a
       suspend initiator requested a suspend but we got to done (or abort)
       blocking before the signal was delivered. We return `NotifyAndWait` and the
       caller is supposed to send a notification to the suspend initiator and
       wait for a resume(with suspend count == 0) which will return to
       `RUNNING` (the victim thread goes to `STATE_BLOCKING_SELF_SUSPENDED`).
 - `STATE_BLOCKING_ASYNC_SUSPENDED`
   - The thread is not executing in this state (it is waiting for a `resume`).
   - a thread transitions into this state from `STATE_BLOCKING_SUSPEND_REQUESTED`
     on a `finish_async_suspension` transition (see above).
   - on a `resume` transition we decrement the suspend count. if it's still
     positive we stay in this state. if it's 0 the thread transitions out of
     this state.  Unlike other resume transitions, in this case the thread goes
     back to `BLOCKING` and resumes executing in GC Safe mode.  The resume caller
     must notify the resume initiator (ie this is a `ResumeInitAsyncResume`).
   - on a `request_suspension` we stay in this state and just increment the
     suspend count.
   - all other transitions from this state are illegal.

We add two new return values for `abort_blocking` and `done_blocking`:
  - `DoneBlockingNotifyAndWait` and `AbortBlockingNotifyAndWait`. Similar to
    `SelfSuspendNotifyAndWait` - there was a race between the second phase of a
    preemptive suspend and another operation (in that case a poll, in this case a
    done or abort blocking) and the other operation won, so its caller should
    notify the suspend initiator. Since the thread is now suspended, the caller
    should wait for the resume signal.

We add new results for `request_initiate_suspension` `MonoRequestSuspendResult`
(formerly `MonoRequestAsyncSuspendResult`):
  - `ReqSuspendAlreadySuspended` means the thread is not executing and we just
    incremented its suspend count and there's nothing else for the suspend
    initiator to do. (The old `AsyncSuspendAlreadySuspended`)
  - `ReqSuspendAlreadySuspendedBlocking` means the thread is in
    `BLOCKING_SUSPEND_REQUESTED` and we initiated another suspend request.  This
    should only happen with full cooperative suspend - with hybrid suspend we
    will use a two phase preemptive suspend on a blocking thread and since only
    a single suspend initiator is active, this state should not be visible to
    another suspend initiator.
  - `ReqSuspendInitSuspendRunning` - Thread is executing in GC Unsafe mode and the
    caller should suspend it.  This is the old `AsyncSuspendInitSuspend`.  In
    full preemptive suspend this is the only suspend initiation action and the caller
    should begin the preemptive suspend procedure.  In full coop we expect the
    victim thread to reach a safepoint and poll to finish suspending.
  - `ReqSuspendInitSuspendBlocking` - Thread is executing in GC Safe mode and
    the caller should initiate a suspend.  In full coop the caller has nothing
    to do - a thread executing in blocking mode is assumed to be suspended.  In
    hybrid suspend, the caller has to initiate a preemptive suspend.

### Suspend initiator, signal handler, self suspender

Basically we call `mono_threads_transition_request_suspension` instead of `mono_threads_transition_request_async_suspension`.
Then if it says to suspend a running thread we call `begin_suspend_for_running_thread`, if it says initiate a suspension of a blocking thread, we call `begin_suspend_for_blocking_thread`, and if it says do nothing we do nothing.

**Questions**
1. We have an assumption in `safe_interrupt_thread` that we can return from a self interrupt if we're using coop suspend.  I'm not sure if that makes sense for hybrid suspend, and not clear what to do about it.
2. In `mono_thread_info_safe_suspend_and_run` we expect that the callback won't return `KeepSuspended` if coop is enabled.  I'm not sure what this means for hybrid.
3. A bunch of tests like `monitor-abort` and `appdomain-threadpool-unload` are failing if I enable hybrid suspend.  Don't know why yet.

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -312,9 +312,14 @@ mono_threads_exit_gc_safe_region_unbalanced_internal (gpointer cookie, MonoStack
 	case DoneBlockingOk:
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		break;
-	case DoneBlockingWait:
 	case DoneBlockingNotifyAndWait:
 		// in full coop NotifyAndWait doesn't notify
+		if (mono_threads_is_hybrid_suspension_enabled ()) {
+			mono_threads_notify_initiator_of_suspend (info);
+		}
+		mono_thread_info_wait_for_resume (info);
+		break;
+	case DoneBlockingWait:
 		THREADS_SUSPEND_DEBUG ("state polling done, notifying of resume\n");
 		mono_thread_info_wait_for_resume (info);
 		break;
@@ -413,9 +418,14 @@ mono_threads_enter_gc_unsafe_region_unbalanced_with_info (MonoThreadInfo *info, 
 	case AbortBlockingOk:
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		break;
-	case AbortBlockingWait:
 	case AbortBlockingNotifyAndWait:
 		// in full coop, notify and wait doesn't need to notify
+		if (mono_threads_is_hybrid_suspension_enabled ()) {
+			mono_threads_notify_initiator_of_suspend (info);
+		}
+		mono_thread_info_wait_for_resume (info);
+		break;
+	case AbortBlockingWait:
 		mono_thread_info_wait_for_resume (info);
 		break;
 	default:
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -313,6 +313,8 @@ mono_threads_exit_gc_safe_region_unbalanced_internal (gpointer cookie, MonoStack
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		break;
 	case DoneBlockingWait:
+	case DoneBlockingNotifyAndWait:
+		// in full coop NotifyAndWait doesn't notify
 		THREADS_SUSPEND_DEBUG ("state polling done, notifying of resume\n");
 		mono_thread_info_wait_for_resume (info);
 		break;
@@ -412,6 +414,8 @@ mono_threads_enter_gc_unsafe_region_unbalanced_with_info (MonoThreadInfo *info, 
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		break;
 	case AbortBlockingWait:
+	case AbortBlockingNotifyAndWait:
+		// in full coop, notify and wait doesn't need to notify
 		mono_thread_info_wait_for_resume (info);
 		break;
 	default:
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -524,6 +524,15 @@ mono_threads_is_blocking_transition_enabled (void)
 #endif
 }
 
+gboolean
+mono_threads_is_hybrid_suspension_enabled (void)
+{
+	static int is_hybrid_suspension_enabled = -1;
+	if (G_UNLIKELY (is_hybrid_suspension_enabled == -1))
+		is_hybrid_suspension_enabled = g_hasenv ("MONO_ENABLE_COOP") && g_hasenv ("MONO_ENABLE_HYBRID_SUSPEND");
+	return is_hybrid_suspension_enabled == 1;
+}
+
 
 void
 mono_threads_coop_init (void)
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -537,10 +537,14 @@ mono_threads_is_blocking_transition_enabled (void)
 gboolean
 mono_threads_is_hybrid_suspension_enabled (void)
 {
+#if defined(USE_HYBRID_SUSPEND)
+	return TRUE;
+#else
 	static int is_hybrid_suspension_enabled = -1;
 	if (G_UNLIKELY (is_hybrid_suspension_enabled == -1))
 		is_hybrid_suspension_enabled = g_hasenv ("MONO_ENABLE_COOP") && g_hasenv ("MONO_ENABLE_HYBRID_SUSPEND");
 	return is_hybrid_suspension_enabled == 1;
+#endif
 }
 
 
--- a/mono/utils/mono-threads-state-machine.c
+++ b/mono/utils/mono-threads-state-machine.c
@@ -598,29 +598,6 @@ STATE_BLOCKING_ASYNC_SUSPENDED: This is an exit state of abort blocking, can't h
 	}
 }
 
-MonoThreadUnwindState*
-mono_thread_info_get_suspend_state (MonoThreadInfo *info)
-{
-	int raw_state, cur_state, suspend_count;
-	UNWRAP_THREAD_STATE (raw_state, cur_state, suspend_count, info);
-	switch (cur_state) {
-	case STATE_ASYNC_SUSPENDED:
-	case STATE_BLOCKING_ASYNC_SUSPENDED:
-		return &info->thread_saved_state [ASYNC_SUSPEND_STATE_INDEX];
-	case STATE_SELF_SUSPENDED:
-	case STATE_BLOCKING_SELF_SUSPENDED:
-		return &info->thread_saved_state [SELF_SUSPEND_STATE_INDEX];
-	default:
-/*
-STATE_RUNNING
-STATE_BLOCKING_SUSPEND_REQUESTED
-STATE_ASYNC_SUSPEND_REQUESTED
-STATE_BLOCKING: All those are invalid suspend states.
-*/
-		g_error ("Cannot read suspend state when target %p is in the %s state", mono_thread_info_get_tid (info), state_name (cur_state));
-	}
-}
-
 // State checking code
 /**
  * Return TRUE is the thread is in a runnable state.
--- a/mono/utils/mono-threads-state-machine.c
+++ b/mono/utils/mono-threads-state-machine.c
@@ -41,13 +41,17 @@ state_name (int state)
 {
 	static const char *state_names [] = {
 		"STARTING",
-		"RUNNING",
 		"DETACHED",
+
+		"RUNNING",
 		"ASYNC_SUSPENDED",
 		"SELF_SUSPENDED",
 		"ASYNC_SUSPEND_REQUESTED",
+
 		"STATE_BLOCKING",
+		"STATE_BLOCKING_ASYNC_SUSPENDED",
 		"STATE_BLOCKING_SELF_SUSPENDED",
+		"STATE_BLOCKING_SUSPEND_REQUESTED",
 	};
 	return state_names [get_thread_state (state)];
 }
@@ -73,9 +77,12 @@ check_thread_state (MonoThreadInfo* info)
 	case STATE_SELF_SUSPENDED:
 	case STATE_ASYNC_SUSPEND_REQUESTED:
 	case STATE_BLOCKING_SELF_SUSPENDED:
+	case STATE_BLOCKING_SUSPEND_REQUESTED:
+	case STATE_BLOCKING_ASYNC_SUSPENDED:
 		g_assert (suspend_count > 0);
 		break;
-	case STATE_BLOCKING: //this is a special state that can have zero or positive suspend count.
+	case STATE_BLOCKING:
+		g_assert (suspend_count == 0);
 		break;
 	default:
 		g_error ("Invalid state %d", cur_state);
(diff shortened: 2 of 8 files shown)


lambdageek:
Attn @kumpera @luhenry

mono/utils/mono-threads-state-machine.c
```diff
@@ -322,6 +396,13 @@ mono_threads_transition_request_resume (MonoThreadInfo* info)
 
 			return ResumeOk; //Resume worked and there's nothing for the caller to do.
 		} else {
+			if (cur_state == STATE_BLOCKING_ASYNC_SUSPENDED) {
```
luhenry: We should probably just have it in a different case statement to ease undestanding the code and keep it consistent with the rest

mono/utils/mono-threads-state-machine.c
```diff
@@ -539,13 +645,15 @@ mono_thread_info_get_suspend_state (MonoThreadInfo *info)
 	case STATE_SELF_SUSPENDED:
 	case STATE_BLOCKING_AND_SUSPENDED:
 		return &info->thread_saved_state [SELF_SUSPEND_STATE_INDEX];
+	case STATE_BLOCKING_ASYNC_SUSPENDED:
```
luhenry: You can just add this case at https://github.com/mono/mono/pull/8068/files#diff-6efe503cfde87f09553ab6c97c061604R644

mono/utils/mono-threads.c
```diff
@@ -860,16 +873,32 @@ mono_thread_info_resume (MonoNativeThreadId tid)
...
+	default:
+		g_assert_not_reached ();
+	}
+}
+
 gboolean
 mono_thread_info_begin_suspend (MonoThreadInfo *info)
 {
+	if (G_UNLIKELY (mono_threads_is_hybrid_suspension_enabled ())) {
+		return begin_suspend_request_hybrid_suspension (info);
```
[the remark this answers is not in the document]
lambdageek: Let me see if I'm getting this: have a single "begin suspending" transition that subsumes both `request_async_suspension` and `request_hybrid_suspension`;  give it enough return codes so that we can switch on the results and in all cases do the correct `begin_safepoint_suspend`/`begin_preemptive_suspend`?

That makes sense but there are a couple problems:
1. (minor) right now `ASYNC_SUSPENDED` has a `request_async_suspension` selfloop, but no legal `request_hybrid_suspension` transition (if we decided to suspend that thread via safepointing, it coudn't have preemptively suspended and shouldn't try to preemptively suspend for suspend_count >1.  This one is probably harmless to add.).  The symmetric situation on `BLOCKING_ASYNC_SUSPENDED` - it has a `request_hybrid_suspension` selfloop but no legal `request_async_suspension` transition - if we decided to suspend it preemptively we shouldn't try to suspend it again via checkpointing. I think it's better when more things are illegal.
2. (serious) in `BLOCKING` state, `request_async_suspension` loops back to `BLOCKING`, but `request_hybrid_suspension` goes to `BLOCKING_SUSPEND_REQUESTED` - this decision is driven purely by external considerations, not by the state machine state.  So we need two distinct transitions for this case.

In general, the suspend transitions represent *policy* - different suspend techniques starting from the same states want different decisions.  On the resume side it's different - the resume policy is encoded in the state that the thread entered when we suspended it.

Alternately, maybe we could split the blocking state into, let's say `BLOCKING_NONPREEMPTABLE` (kind of like the original meaning of BLOCKING - it's in a syscall or foreign code that doesn't touch managed at all which will suspend at safepoints on the done/abort transitions) and `BLOCKING_PREEMPTABLE` which is preempted.  In that case we push all the complexity into runtime code (we now need to do BEGIN_SYSCALL/END_SYSCALL and BEGIN_FOREGIN/END_FOREIGN).  On the bright side we won't have to restart syscalls.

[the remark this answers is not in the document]
lambdageek: In the morning things look better.  The solution is to do a two-phase suspend in blocking always, the same way that we do it in the running case (ie: `request_suspension` should move from `Blocking` to `Blocking_Suspend_Requested`, always, which communicates enough information to the suspend initiator to then decide on a suspend policy).  The previous `request_async_suspension` transition from `Blocking` to `Blocking` was just an artifact of not having an explicit `Blocking_Suspend_Requested` state.

And there's a benefit for the state machine, too: we can establish the invariant that in `Blocking` the suspend count is always equal to 0 and in `Blocking_Suspend_Requested` it's always strictly positive.

I'm going to update the PR to make it work this way.

Thanks @luhenry !

lambdageek:
@luhenry updated:
1. Use a single unified `mono_threads_transition_request_initiate_suspension`
2. Add `--with-hybrid-suspend` configure option. Defaults to 'no'. To enable requires '--with-cooperative-gc'.

lambdageek:
The `monitor-abort` and `appdomain-threadpool-unload` tests are still failing if hybrid coop is enabled.  They work with full preemptive suspend and full cooperative suspend.

mono/utils/mono-threads-coop.c
```diff
@@ -510,7 +523,7 @@ mono_threads_are_safepoints_enabled (void)
 gboolean
 mono_threads_is_blocking_transition_enabled (void)
 {
-#if defined(USE_COOP_GC)
+#if defined(USE_HYBRID_COOP)
```
luhenry: We want it for both `USE_COOP_GC` *and* `USE_HYBRID_COOP`.

mono/utils/mono-threads-coop.c
```diff
@@ -412,6 +420,11 @@ mono_threads_enter_gc_unsafe_region_unbalanced_with_info (MonoThreadInfo *info,
 		info->thread_saved_state [SELF_SUSPEND_STATE_INDEX].valid = FALSE;
 		break;
 	case AbortBlockingWait:
+	case AbortBlockingNotifyAndWait:
+		// in full coop, notify and wait doesn't need to notify
+		if (result == AbortBlockingNotifyAndWait && mono_threads_is_hybrid_suspension_enabled ()) {
```
luhenry: We should probably just duplicate the code for `AbortBlockingWait` and `AbortBlockingNotifyAndWait` to keep it consistent across the board.

mono/utils/mono-threads.c
```diff
@@ -935,21 +983,45 @@ suspend_sync (MonoNativeThreadId tid, gboolean interrupt_kernel)
...
 			mono_hazard_pointer_clear (hp, 1);
 			return NULL;
 		}
 		break;
-	case AsyncSuspendBlocking:
-		if (interrupt_kernel)
+	case ReqSuspendAlreadySuspendedBlocking:
+	case ReqSuspendInitSuspendBlocking: {
+		gboolean did_interrupt = FALSE;
+		if (result == ReqSuspendInitSuspendBlocking) {
```
luhenry: We should duplicate the code for `ReqSuspendAlreadySuspendedBlocking` and `ReqSuspendInitSuspendBlocking`

mono/utils/mono-threads-state-machine.c
```diff
@@ -170,12 +179,15 @@ This transition initiates the suspension of another thread.
...
+- ReqSuspendInitSuspendBlocking: Thread in blocking state, caller may initiate suspend.
+- ReqSuspendAlreadySuspended: Thread was already suspended and not executing, nothing to do.
+- ReqSuspendAlreadySuspendedBlocking: Thread was already in blocking and a suspend was requested
+                                      and the thread is still executing (perhaps in a syscall),
+                                      nothing to do.
 */
-MonoRequestAsyncSuspendResult
-mono_threads_transition_request_async_suspension (MonoThreadInfo *info)
+MonoRequestSuspendResult
+mono_threads_transition_request_initiate_suspension (MonoThreadInfo *info)
```
luhenry: `mono_threads_transition_request_suspension` should be enough and provides a good counter-part to `mono_threads_transition_request_resume`.

luhenry:
The individual comments are nitpick and are not blocking the PR in general, just the requested change at https://github.com/mono/mono/pull/8068#discussion_r180464543 is.

mono/utils/mono-threads-coop.c
luhenry: Also, Hybrid Coop doesn't make much sense, Hybrid Suspend would be more meaningful.

lambdageek:
`[threads] Return suspend sort from begin_suspend_for_running_thread` fixed `monitor-abort.exe` and `appdomain-threadpool-unload.exe` under hybrid suspend.

Now the only one that's broken is `thread-suspend-suspended.exe`.

----

The `mono_thread_info_get_suspend_state` change fixed the massive CI breakage of coop: `STATE_BLOCKING_SUSPEND_REQUESTED` is only a suspend state for full coop.

lambdageek:
@luhenry I pushed a commit that wraps all the `mono_os_event_*` calls in `threads.c` with `MONO_ENTER_GC_SAFE/MONO_EXIT_GC_SAFE` - it makes `appdomain-threadpool-unload.exe` work under hybrid suspend.  I don't really have a good sense whether this was the _right_ fix.
(`MonoOSEvent` on non-Windows is backed by a mutex, so it seems right to do transitions - but it's in the suspend machinery itself, so...)

luhenry:
@lambdageek the longer term fix would be to have `mono_coop_event_*` which would wrap the `mono_os_event_*` functions with `MONO_{ENTER,EXIT}_GC_SAFE` and use these instead, but that can be done in a follow-up PR.

luhenry:
And maybe the test is flaky just because we have a lower chance of hitting what you observe with hybrid suspend.

lambdageek:
@monojenkins merge

PR #8072, opened 2018-04-06
[mcs] Fixes codegen when probing reference type expression is known t…

marek-safar:
…o pass and pattern variable has to be created


marek-safar:
@monojenkins backport 2018-02

marek-safar:
@monojenkins backport 2018-04

Issue #8074, opened 2018-04-06
signing problems in preset fullaot build


jaykrell:
I see this 100%.
Why others do not?
I usually skip aot so usually I don't see it.
Linux Ubuntu 1604 and 1710 tested.
mono-devel package is installed, which is probably not relevant

git clean -xddf (or new git clone)
./configure --with-runtime_preset=fullaot 
likely also with others (hybridaot)
make

Many errors like
```
MONO_PATH="./../../class/lib/build:$MONO_PATH" /s/mono/runtime/mono-wrapper  ./../../class/lib/build/sn.exe -q -R ../../class/lib/testing_aot_full/mscorlib.dll ./../../class/mono.snk
Couldn't sign the assembly ../../class/lib/testing_aot_full/mscorlib.dll with this key pair. Public key of assembly did not match signing public key.
../../build/library.make:342: recipe for target '../../class/lib/testing_aot_full/mscorlib.dll' failed
```
If you run make V=1 you see the sn command.
 You can see the installed sn works ok.
Remove the -Q for quiet and you see a warning about missing machine.config.
Run with ptrace and you see it looking for 4.5/machine.config.
Look in the tree and you see 1.1 and 2.0.
Copy the 2.0 to 4.5 and problem goes away.
And can be edited down a bunch.
Actual fix is to move symlink of the file to outside a conditional.

PR shortly.

Maybe people aren't using fullaot or git clean??
Repro is 100% for me. Even back in September 2017.

PR #8075, opened 2018-04-06
Symlink the 4.5 machine.config unconditionally.

jaykrell:
This fixes https://github.com/mono/mono/issues/8074.

I see this 100%. Why others do not? I usually skip aot so usually I don't see it. Linux Ubuntu 1604 and
1710 tested. mono-devel package is installed, which is probably not relevant.

git clean -xddf (or new git clone)
./configure --with-runtime_preset=fullaot   likely also with others (hybridaot)
make

Many errors like:

MONO_PATH="./../../class/lib/build:$MONO_PATH" /s/mono/runtime/mono-wrapper ./../../class/lib/build/sn.exe
-q -R ../../class/lib/testing_aot_full/mscorlib.dll ./../../class/mono.snk Couldn't sign the assembly
../../class/lib/testing_aot_full/mscorlib.dll with this key pair. Public key of assembly did not match
signing public key. ../../build/library.make:342: recipe for target
'../../class/lib/testing_aot_full/mscorlib.dll' failed

If you run make V=1 you see the sn command.
 You can see the installed sn works ok. Remove the -Q for quiet and you see a warning about missing
machine.config. Run with ptrace and you see it looking for 4.5/machine.config. Look in the tree and you
see 1.1 and 2.0. Copy the 2.0 to 4.5 and problem goes away. And can be edited down a bunch. Actual fix is
to move symlink of the file to outside a conditional.

Maybe people aren't using fullaot or git clean? Repro is 100% for me.
I was going to bisect, but it was occuring as far back as September 2017 so I did not.

### DIFF
--- a/runtime/Makefile.am
+++ b/runtime/Makefile.am
@@ -6,8 +6,11 @@ etctmp = etc
 symlinks = etc/mono/1.0/machine.config etc/mono/2.0/machine.config etc/mono/2.0/web.config etc/mono/browscap.ini etc/mono/2.0/Browsers/Compat.browser
 
 if INSTALL_4_x
-symlinks += etc/mono/4.0/machine.config etc/mono/4.0/web.config etc/mono/4.5/web.config etc/mono/4.5/machine.config etc/mono/4.0/Browsers/Compat.browser etc/mono/4.5/Browsers/Compat.browser
+symlinks += etc/mono/4.0/machine.config etc/mono/4.0/web.config etc/mono/4.5/web.config etc/mono/4.0/Browsers/Compat.browser etc/mono/4.5/Browsers/Compat.browser
 endif
+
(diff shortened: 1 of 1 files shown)


jaykrell:
This runtime preset stuff causes no runtime to be installed..which is why the ifdef makes some sense theres. But we still run stuff in-tree that needs the machine.config (sn.exe).

It'd be great if all the variations could be installed sidebyside and mono chose the right one.

The current scheme is has so many variations, and testing them is then more difficult than it could be...

jaykrell:
@monojenkins build OSX i386

jaykrell:
@monojenkins build OS X i386

jaykrell:
Ping?

jaykrell:
Ping?

PR #8053, opened 2018-04-05
[ci] Add {,hybrid,full}aot_llvm presets to make sure we compile with LLVM

luhenry:

PR #8076, opened 2018-04-06
Fix build race conditions with random temporary file names.

jaykrell:


jaykrell:
Note quite right.

good:

king.exe.so.tmp "temp-hijacking-llvm.o" /tmp/mono_aot_cvprsc.o 
Executing the native linker: "ld"  -shared -o /s/mono/mono/tests/tailcall/coreclr/JIT/opt/ETW/TailCallCases.exe.so.tmp "temp-TailCallCases-llvm.o" /tmp/mono_aot_Up7k5F.o 

not good:

Compiled: 21/21 (100%), LLVM: 21 (100%), No GOT slots: 19 (90%), Direct calls: 0 (100%)
Executing the native assembler: "as" --64  -o /tmp/mono_aot_ZJsdsp.o /tmp/mono_aot_ZJsdsp
Executing the native linker: "ld"  -shared -o /s/mono/mono/tests/tailcall/coreclr/JIT/Methodical/tailcall_v4/tailcall_AV.exe.so.tmp "temp-o-llvm.o" /tmp/mono_aot_ZJsdsp.o 

But the assembly is actually named "o".

The approach is an improvement but still leaves room perhaps..

jaykrell:
This might might might address both kumpera and luhenry's problem..

vargaz:
There is a temp-path aot option for this purpose.

jaykrell:
Fixes obscure severe problem easy to fall into by accident. And existing switch not easy to use when compiling same files concurrently,

jaykrell:
This also resembles what the code does a few lines later in other cases.

https://github.com/mono/mono/pull/8076/files#diff-499ab2170cdda019ebc31342389233b3R12753

luhenry:
Pretty much cherry-picked it in https://github.com/mono/mono/pull/8053/commits/320a1f981631b42db39b0c074edbdf8470e50eba

jaykrell:
Ah, I thought closed as redundant with -temp-path or such and did not see that. Good. You do need to cast NULL to char* or const char* or perhaps void* for portability. A problem in the prior code. I didn't want a directory, just a file, but ok.

jaykrell:
Also that unused temp_path[32] throws off anyone looking for "temp" in the code. They have to go look for the uses and find none.

PR #8078, opened 2018-04-06
[interp] use same domain as transform phase to register debug data for a method

lewurm:
needs backport to `2018-02` and `2018-04`


lewurm:
@monojenkins rebase

PR #8091, opened 2018-04-07
Enable mono_runtime_set_execution_mode to handle MONO_AOT_MODE_NORMAL

alexanderkyte:
This recently added function doesn't have a case for the default AOT mode.

### DIFF
--- a/mono/mini/driver.c
+++ b/mono/mini/driver.c
@@ -2602,6 +2602,9 @@ mono_runtime_set_execution_mode (MonoEEMode mode)
 		mono_ee_features.force_use_interpreter = TRUE;
 		break;
 
+	case MONO_AOT_MODE_NORMAL:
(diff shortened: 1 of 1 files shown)


sourceskyboxer:
Same here I have tried but still error "Unknown execution-mode 1"
If I generate mkbundle with my c# application

sourceskyboxer:
Always error "Unknown execution-mode 1"

Windows 10 x64 - mkbundle can not generate and use with clang.exe and cl.exe than run bundled exe in cygwin terminal than it says "Unknown execution-mode 1"

"passing" is wrong -> "unstable" is correct

Please stop to lie us! Because you have to write "passing" but why does it not work for me? That is why I am mad now.

alexanderkyte:
Oops typo, let me fix that

PR #8093, opened 2018-04-07
[mini-llvm] Remove unused struct field.

jaykrell:


jaykrell:
@monojenkins build failed

PR #8096, opened 2018-04-07
[llvm] Add intrinsics support for some of the new MathF icalls when using llvm+float32.

vargaz:


marek-safar:
There seems to be an issue with LLVM conversion

```
Call parameter type does not match function signature!
  %27 = call cc16 float @p_881_plt_mscorlib_System_MathF_Round_single_llvm(float %t114)
```

PR #8102, opened 2018-04-08
Fix generic-stack-traces2.2.cs under FullAOT+LLVM.

jaykrell:
It relies on poorly defined behavior that mono cannot likely well define.

https://msdn.microsoft.com/en-us/library/system.diagnostics.stacktrace(v=vs.110).aspx

"StackTrace might not report as many method calls as expected, due to code transformations
that occur during optimization."


jaykrell:
@monojenkins build Linux x64

jaykrell:
Ping?

PR #8111, opened 2018-04-09
[bcl] Mark some some code with DISABLE_COM which depends on icalls which are disabled in the runtime when --enable-minimal=com is used.

vargaz:


marek-safar:
System.Runtime.InteropServices.WindowsRuntime fails to build. I'd recommend move the ifdef to have WindowsRuntimeMarshal class with no body kept for `DISABLE_COM`

PR #8120, opened 2018-04-09
[build] Exclude platform specific xunit tests

marek-safar:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5046,18 +5046,16 @@ fi
 
     if test x$host_darwin = xyes; then
       echo "BUILD_PLATFORM = darwin" >> $mcs_topdir/build/config.make
+      echo "HOST_PLATFORM ?= darwin" >> $mcs_topdir/build/config.make
+      echo "XTEST_PLATFORM ?= osx" >> $mcs_topdir/build/config.make
     elif test x$host_win32 = xyes; then
       echo "BUILD_PLATFORM = win32" >> $mcs_topdir/build/config.make
+      echo "HOST_PLATFORM ?= win32" >> $mcs_topdir/build/config.make
+      echo "XTEST_PLATFORM ?= windows" >> $mcs_topdir/build/config.make
     else
       echo "BUILD_PLATFORM = linux" >> $mcs_topdir/build/config.make
-    fi
-
-    if test x$host_darwin = xyes; then
-      echo "HOST_PLATFORM ?= darwin" >> $mcs_topdir/build/config.make
-    elif test x$host_win32 = xyes; then
-      echo "HOST_PLATFORM ?= win32" >> $mcs_topdir/build/config.make
-    else
       echo "HOST_PLATFORM ?= linux" >> $mcs_topdir/build/config.make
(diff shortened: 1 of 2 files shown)


mcs/build/tests.make
```diff
@@ -268,6 +268,11 @@ XTEST_HARNESS_PATH := $(topdir)/../external/xunit-binaries
 XTEST_HARNESS = $(XTEST_HARNESS_PATH)/xunit.console.exe
 XTEST_HARNESS_FLAGS := -noappdomain -noshadow -parallel none -nunit TestResult-$(PROFILE)-xunit.xml
 XTEST_TRAIT := -notrait category=failing -notrait category=nonmonotests -notrait Benchmark=true -notrait category=outerloop
+# The horrible CoreFX tests logic is double inverted so this actually excludes Windows tests
+# best to search for `property name="category"` in the xml output to see what's going on
+# https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/PlatformSpecificDiscoverer.cs
+XTEST_TRAIT_PLATFORM := -notrait category=nonlinuxtests
```
akoeplinger: This doesn't make sense. The way it works is that if you add `[PlatformSpecific(TestPlatforms.Windows)]` to a test then the properties it adds are:

```
<property name="category" value="nonlinuxtests" />
<property name="category" value="nonosxtests" />
<property name="category" value="nonfreebsdtests" />
<property name="category" value="nonnetbsdtests" />
```

... which is based on the platforms that corefx knows about.

The idea is that you pass `-notrait category=non<CURRENT_PLATFORM>tests` to xunit and it automatically excludes tests that don't apply to this platform.

We should be adding logic based on the platfrom, e.g. by adding `XTEST_TRAIT_PLATFORM` to mcs/build/platforms/darwin.make etc and set it to ` -notrait category=nonosxtests` and so on.

marek-safar: As I said in the comment, the purpose was to exclude always now windows tests which this trait does but I'll look into making this work properly

akoeplinger: Well it also excludes OSX specific tests this way :)

Not sure if you need it urgently but I'd prefer doing it properly.

configure.ac
```diff
@@ -5046,18 +5046,16 @@ fi
...
       echo "BUILD_PLATFORM = darwin" >> $mcs_topdir/build/config.make
-    elif test x$host_win32 = xyes; then
-      echo "BUILD_PLATFORM = win32" >> $mcs_topdir/build/config.make
-    else
-      echo "BUILD_PLATFORM = linux" >> $mcs_topdir/build/config.make
-    fi
-
-    if test x$host_darwin = xyes; then
       echo "HOST_PLATFORM ?= darwin" >> $mcs_topdir/build/config.make
+      echo "XTEST_PLATFORM ?= osx" >> $mcs_topdir/build/config.make
```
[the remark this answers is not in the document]
marek-safar: It has to match https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/XunitConstants.cs#L12-L16

PR #8133, opened 2018-04-09
[jit] Various float32 fixes.

vargaz:


mono/mini/iltests.il
```diff
@@ -2264,14 +2264,14 @@ HAS_VALUE:	ldc.i4.1
 		ret
 	}
 
+/*	   Disabled until they can be fixed to run on amd64
```
kumpera: Can we file issues on this? I didn't even knew we had disabled tests on iltests.il

PR #8138, opened 2018-04-09
[console-unix] retry `tcsetattr` if it was interrupted by signal

lewurm:
if configuring the terminal failed, we still try to continue (see
a3ce2f008bdb2fa5108034f1687fc7f87c92a0b8 ) and therefore succedings
`read`s can fail.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=26983

### DIFF
--- a/mono/metadata/console-unix.c
+++ b/mono/metadata/console-unix.c
@@ -465,7 +465,9 @@ ves_icall_System_ConsoleDriver_TtySetup (MonoStringHandle keypad, MonoStringHand
 #endif
 	gint ret;
 	do {
+		MONO_ENTER_GC_SAFE;
 		ret = tcsetattr (STDIN_FILENO, TCSANOW, &mono_attr);
+		MONO_EXIT_GC_SAFE;
 	} while (ret == -1 && errno == EINTR);
 
 	if (ret == -1)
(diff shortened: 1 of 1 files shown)


lewurm:
@coop guys: does the syscall need a `MONO_ENTER_GC_SAFE`/`MONO_EXIT_GC_SAFE` guard as other syscalls in this file?

vargaz:
Probly yes, if it can block.

mono/metadata/console-unix.c
```diff
@@ -463,7 +463,14 @@ ves_icall_System_ConsoleDriver_TtySetup (MonoStringHandle keypad, MonoStringHand
...
-	if (tcsetattr (STDIN_FILENO, TCSANOW, &mono_attr) == -1)
+	gint ret;
+	do {
+		MONO_ENTER_GC_SAFE;
+		ret = tcsetattr (STDIN_FILENO, TCSANOW, &mono_attr);
+		MONO_EXIT_GC_SAFE;
+	} while (ret == -1 && errno == EINTR);
+
+	if (ret == -1)
 		return FALSE;
```
migueldeicaza: Whoa!   very good catch!   This is indeed broken code.

lewurm:
@monojenkins squash

PR #8166, opened 2018-04-11
[runtime] Make GetFrames consider nested exceptions

### LINKED ISSUE #7260: StackTrace.GetFrames() does not return frames from captured traces
Sorry in advance if this is not correct, that's the first time I fill an issue on this repository.

## Current Behavior

Sometimes, frames printed from `StackTrace.ToString()` and frames returned by `StackTrace.GetFrames()` are different. This is because `.GetFrames()` don't care about `captured_traces` while `.ToString()` do care.

## Steps to Reproduce

1. Use `ExceptionDispatchInfo.Capture(...).Throw()`
2. Catch the thrown exception
3. Print the resulting stacktrace
4. Iterate over all the frames in the stacktrace
5. See that the two steps above do not send exactly the same information

Here's a small (and dumb) example: [test.zip](https://github.com/mono/mono/files/1754271/test.zip).

## Expected Behavior

There must be a way to access captured frames in `StackTrace`.  

Here is the output of the same bit of code (that I linked above) on both mono and Microsoft .Net.
Mono:
capture du 2018-02-24 11-46-09
Microsoft .Net:
sans titre

-[ ] macOS
-[x] Linux
-[ ] Windows

**Version Used**:
5.8.0.127

alexanderkyte:
Fixes #7260

### DIFF
--- a/mcs/class/corlib/System.Diagnostics/StackTrace.cs
+++ b/mcs/class/corlib/System.Diagnostics/StackTrace.cs
@@ -180,7 +180,15 @@ namespace System.Diagnostics {
 		[ComVisibleAttribute (false)]
 		public virtual StackFrame[] GetFrames ()
 		{
-			return frames;
+			if (captured_traces == null)
+				return frames;
+
+			var accum = new List<StackFrame> ();
+			foreach (var t in captured_traces)
+				accum.AddRange(t.GetFrames ());
+
+			accum.AddRange (frames);
+			return accum.ToArray ();
 		}
 
 		static bool isAotidSet;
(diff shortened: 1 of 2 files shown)


luhenry:
This breaks on Linux x64 FullAOT https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot/9673/testReport/junit/(root)/StackFrameTest3/GetFramesAndNestedExc/

alexanderkyte:
That was my mistake, I copy and pasted and left the string "Main" in there. It's working now on my machine.

PR #8196, opened 2018-04-12
[jit] Use r8 for r4 stack vars to avoid problems with stack merges when one side is an r4 and the other an r8. Add conversions for compares.

vargaz:


mono/mini/method-to-ir.c
```diff
@@ -1436,10 +1436,20 @@ mono_compile_get_interface_var (MonoCompile *cfg, int slot, MonoInst *ins)
...
 	MonoInst *res;
 	int pos, vnum;
+	MonoType *type;
+
+	/*
+	 * Use double for r4 stack vars to avoid narrowing with r4-r8 stack
+	 * merges.
+	 */
+	if (ins->type == STACK_R4)
+		type = &mono_defaults.double_class->byval_arg;
```
luhenry: This will require a `m_class_get_byval_arg`.

vargaz:
Thats kinda hard to do since by the time we realize we need an r8, the variable is already allocated and assigned. Note that this is the same behavior as the one without -O=float32.

kumpera:
We can backpatch the r4 site to convert & store to a r8.

vargaz:
Will add that as a fixme.

luhenry:
@monojenkins build Linux x64

luhenry:
@monojenkins build Linux i386

PR #8198, opened 2018-04-12
[runtime] Improve SIGQUIT thread dumps.

alexrp:
The goal of this PR is to make thread dumps triggered via `SIGQUIT` use the same GDB/LLDB code as we do on native crashes. That way, we can provide more useful backtraces when a hang is happening in native code.

`mono/tests/test-runner.cs` already had some separate code to use GDB/LLDB, but any tests in the tree that aren't using this test runner won't benefit from that code, nor would any external users who send `SIGQUIT` to Mono to diagnose hangs.

The main motivator for this PR is the profiler stress test suite, which currently does not get GDB/LLDB output on a test timeout. I could have duplicated the code in `mono/tests/test-runner.cs`, but this approach seemed like a better solution overall.


mono/tests/test-runner.cs
```diff
@@ -579,116 +581,12 @@ class TestInfo {
...
 	{
-		try {
-			TryGDB (pid, data);
-			return;
-		} catch {
-		}
-
-#if !FULL_AOT_DESKTOP && !MOBILE
-		/* LLDB cannot produce managed stacktraces for all the threads */
+#if !MOBILE
```
alexrp: Not 100% sure about this one. The original check seemed weird to me - why wouldn't we want to use `SIGQUIT` just because we're testing the FullAOT profile?

mono/metadata/threads.c
```diff
@@ -3788,7 +3792,7 @@ dump_thread (MonoInternalThread *thread, ThreadDumpUserData *ud)
 		}
 	}
 
-	fprintf (stdout, "%s", text->str);
+	fprintf (stderr, "%s", text->str);
```
lewurm: we should use `g_printerr ()` here

mono/metadata/threads.c
```diff
@@ -3808,7 +3812,7 @@ mono_threads_perform_thread_dump (void)
 	if (!thread_dump_requested)
 		return;
 
-	printf ("Full thread dump:\n");
+	fprintf (stderr, "Full thread dump:\n");
```
lewurm: we should use `g_printerr ()` here

mono/metadata/threads.c
```diff
@@ -3822,6 +3826,8 @@ mono_threads_perform_thread_dump (void)
 
 	g_free (ud.frames);
 
+	fprintf (stderr, "\n");
```
lewurm: we should use `g_printerr ()` here

mono/mini/mini-posix.c
```diff
@@ -886,7 +901,7 @@ mono_gdb_render_native_backtraces (pid_t crashed_pid)
 		goto exec;
 #endif
 
-	fprintf (stderr, "mono_gdb_render_native_backtraces not supported on this platform, unable to find gdb or lldb\n");
+	fprintf (stderr, "gdb_render_native_backtraces not supported on this platform, unable to find gdb or lldb\n");
```
luhenry: You could replase that by `%s` and `__func__`.

mono/mini/mini-posix.c
```diff
@@ -898,10 +913,81 @@ mono_gdb_render_native_backtraces (pid_t crashed_pid)
...
+		int status;
+		pid_t crashed_pid = getpid ();
+
+		/*
+		 * glibc fork acquires some locks, so if the crash happened inside malloc/free,
+		 * it will deadlock. Call the syscall directly instead.
+		 */
+#if defined(HOST_ANDROID)
+		/* SYS_fork is defined to be __NR_fork which is not defined in some ndk versions */
+		if (crash)
```
luhenry: You could replace that by 
```
g_assert (!crash);
return;
```

steveisok:
@lambdageek This PR is approved, but never got merged.  Should we close it or is it still good?

lambdageek:
@steveisok seems useful.  @alexanderkyte, does the crash reporter now gives more info for cases like this?

If it's not redundant with the crash reporter, I'll try to spend a bit of time to resolve the merge conflicts and try to get it merged.

lewing:
This needs work and lacks an owner

PR #8211, opened 2018-04-12
[jit] Use m_class_get_byval_arg () in more places.

vargaz:


luhenry:
@monojenkis build Linux x64

luhenry:
@monojenkins build Linux x64

PR #8219, opened 2018-04-12
Rename MonoClass field byval_arg to _byval_arg

lambdageek:
Replace and remaining uses by `m_class_get_byval_arg` when possible.

----

Add `mono_get_*_type` for `void`, `object`, `int32` and `int` (aka `IntPtr`).

Replace uses of `m_class_get_byval_arg (mono_defaults.*_class)` by the new getters.


lambdageek:
@luhenry in the cases where we're accessing the properties of another class, we should use the getters even in class-init.  In the cases where we're accessing the properties of the same class, I tried to only use the accessors for what I think would be in available at the minimal initialization level (e.g.  name, name_space, token, rank, byval_arg should all be available)

lambdageek:
@monojenkins build Linux x64

PR #8222, opened 2018-04-12
Verify all 4 bytes of PE signature.

jaykrell:


jaykrell:
@monojenkins build Linux x64

PR #8223, opened 2018-04-12
Default platform to net_4_x if none is specified, to fix tools that build without setting a platform

kg:
@alexischr pointed out that the previous updates to the csproj files broke apidiff, causing problems for builds that use it. This PR adds a default Platform value to the csprojs, so that tools that build a csproj without specifying a platform get the default they previously expected (net_4_x).

Alexis is going to test this later to make sure it fixes the issue but I figure it's worth running CI against it now.


luhenry:
@monojenkins build Linux x64

luhenry:
@monojenkins build Linux i386

luhenry:
@monojenkins commit csproj

PR #8224, opened 2018-04-13
 Teach metadata-verify.c about PE32+. Remove extra string copies.

jaykrell:


jaykrell:
The over-verification is still over-verification.
For example, Windows doesn't care about imagebase, and the filealign and sectionalign verifications are not right by far. Both values are quite flexible, but do have some limitations.

jaykrell:

mono/metadata/metadata-verify.c
```diff
@@ -464,40 +462,58 @@ verify_pe_optional_header (VerifyContext *ctx)
...
+	//
+	// Mono requires at least 216 (or pe64:232) size.
+
+	/* LAMESPEC MS plays around this value and ignore it during validation
+	if (read32 (pe_optional_header + 28) != 0x400000)
+	ADD_ERROR (ctx, g_strdup_printf ("Invalid Image base %x", read32 (pe_optional_header + 28)));*/
+	if (read32 (pe_optional_header + 32) != 0x2000)
+		ADD_ERROR (ctx, g_strdup_printf ("Invalid Section Aligmnent %x", read32 (pe_optional_header + 32)));
+	file_alignment = read32 (pe_optional_header + 36);
+	if (file_alignment != 0x200 && file_alignment != 0x1000)
```
jaykrell: This check is not correct for Windows. There is much more flexibility in the value.

mono/metadata/metadata-verify.c
```diff
@@ -464,40 +462,58 @@ verify_pe_optional_header (VerifyContext *ctx)
...
+	//
+	// Mono requires at least 15 data directories. More than that are ignored,
+	// except to require zeros in the 16th.
+	//
+	// Mono requires at least 216 (or pe64:232) size.
+
+	/* LAMESPEC MS plays around this value and ignore it during validation
+	if (read32 (pe_optional_header + 28) != 0x400000)
+	ADD_ERROR (ctx, g_strdup_printf ("Invalid Image base %x", read32 (pe_optional_header + 28)));*/
+	if (read32 (pe_optional_header + 32) != 0x2000)
```
jaykrell: This check is not correct for Windows. There is much more flexibility in the value.

jaykrell: ImageBase is pretty meaningless these days (i.e. since Vista's ASLR).

mono/metadata/metadata-verify.c
kumpera: Well, those are the ECMA required values. The spec is a lot more strict than windows, apparently.

kumpera: I think all compilers but the manage C++ one emits the ECMA values blindly.
I agree that there's little value in keeping or removing them.
I have no strong opinion on the matter, TBH.

PR #8226, opened 2018-04-13
Remove null checks at higher levels -- add or change or preexisting at lower levels.

jaykrell:


jaykrell:
Created https://github.com/mono/mono/pull/8228 with just those then.

mono/metadata/mono-hash.c
```diff
@@ -249,8 +249,8 @@ rehash (MonoGHashTable *hash)
 guint
 mono_g_hash_table_size (MonoGHashTable *hash)
 {
-	g_return_val_if_fail (hash != NULL, 0);
```
kumpera: I like pushing the null check to the callees, I think it brings consistency with what we expect of C APIs.
What I'm on the fence if wether we want to do it like this, expanding g_return_val_if_fail.

vargaz: I think these are internal apis, so they should just assert.

luhenry:
As also pointed out by @kumpera and @vargaz, I would prefer having the check in the callee, and even replace the `g_return_val_if_fail` by `g_assert`. This drastically reduce surprises and chances of introducing bugs.

lambdageek:
Agree with all of the above, except that the functions that deallocate should work with NULL. `g_thing_free (NULL)` or `g_thing_destroy (NULL)` should be allowed.

vargaz:
free () is a well know api which accepts null, but its harder to determine which of our api's are 'free' apis.

jaykrell:
I was surprised but g_return_val_if_fail does assert, so it is close.
Most of the changes aren't needed, to enable most of the null removals.
It was really mainly a foreach or two, and a destroy or two. I can try again.

jaykrell:
Pushing the checks to the callee is exactly what this does.
Perhaps caller was intended??

PR #8233, opened 2018-04-13
Add PowerPC configure flags for LLVM, and clean up failover linker flags

directhex:
Since the Mono runtime doesn't do cross-compiling, it seems sorta weird to link multiple LLVM codegen targets in the failover path in configure. So stop doing that.


luhenry:

Issue #8236, opened 2018-04-13
Add SIGTERM signal handler reporting


alexanderkyte:
On OSX, MonoDevelop would like to have the Force Quit action register a telemetry event. This is useful because excessive quitting might indicate livelock or performance regressions. 

This is lower priority for now. It should be implemented after the free-form mono telemetry is implemented. This hook is mostly going to be used to fire that one. 

The biggest goal here is to make sure that this doesn't have any machinery that can break / will be too annoying. A smooth SIGTERM is important for cleanly quitting the runtime in automated scenarios.

lewurm:
why does it need to be SIGTERM?

luhenry:
@monojenkins backport to 2018-02

luhenry:
@monojenkins backport to 2018-04

PR #8240, opened 2018-04-13
[runtime] Add a signal handler for sigterm

alexanderkyte:
Fixes https://github.com/mono/mono/issues/8236

### DIFF
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -377,6 +377,15 @@ remove_signal_handler (int signo)
 }
 
 void
+mini_register_sigterm_handler (void)
+{
+#ifdef TARGET_OSX
+	/* always catch SIGTERM, conditionals inside of handler */
+	add_signal_handler (SIGTERM, sigterm_signal_handler, 0);
+#endif
+}
+
+void
 mono_runtime_posix_install_handlers (void)
 {
 
@@ -407,9 +416,6 @@ mono_runtime_posix_install_handlers (void)
 	signal (SIGPIPE, SIG_IGN);
 
 	add_signal_handler (SIGABRT, sigabrt_signal_handler, 0);
-
-	/* always catch SIGTERM, conditionals inside of handler */
-	add_signal_handler (SIGTERM, sigterm_signal_handler, 0);
 
 	/* catch SIGSEGV */
 	add_signal_handler (SIGSEGV, mono_sigsegv_signal_handler, 0);
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -73,6 +73,7 @@
 #include <mach/mach.h>
 #include <mach/mach_time.h>
 #include <mach/clock.h>
+#include <mono/utils/mono-merp.h>
 #endif
 
 #if defined(HOST_WATCHOS)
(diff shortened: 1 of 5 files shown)


alexanderkyte:
Fixes https://github.com/mono/mono/issues/8236

alexanderkyte:
@monojenkins build pkg

kumpera:
What's the impact will that have on regular usage of ctrl+c from CLI usage?

vargaz:
Yes, users expect ctlr+c to just terminate the app without doing anything else.

alexanderkyte:
Ctrl+c sends a sigint, sigterm is sent by Force Quit and by POSIX `kill` .

alexanderkyte:
SIGTERM is like SIGKILL, but the dying process is allowed resources to shut down cleanly. This was requested by VS4M for detecting force quit of applications due to hangs.

alexanderkyte:
@monojenkins build pkg

mono/mini/mini-posix.c
```diff
@@ -393,6 +402,13 @@ mono_runtime_posix_install_handlers (void)
 
 	add_signal_handler (SIGABRT, sigabrt_signal_handler, 0);
 
+	gchar *sigterm_env = g_getenv ("MONO_HANDLE_SIGTERM");
```
alexrp: This should be documented in `man/mono.1`.

luhenry: Use `g_hasenv`

luhenry:
@lewurm if we want to keep this feature for MERP only, then we will just enable it in the same icall as we enable MERP support, no need to create another one.

mono/mini/mini-posix.c
```diff
@@ -210,6 +211,20 @@ MONO_SIG_HANDLER_FUNC (static, sigabrt_signal_handler)
 	}
 }
 
+MONO_SIG_HANDLER_FUNC (static, sigterm_signal_handler)
+{
+	MONO_SIG_HANDLER_INFO_TYPE *info = MONO_SIG_HANDLER_GET_INFO ();
+	MONO_SIG_HANDLER_GET_CONTEXT;
+
+#ifdef HOST_DARWIN
```
lewurm: this should be `TARGET_OSX`

mono/mini/mini-posix.c
```diff
@@ -210,6 +211,20 @@ MONO_SIG_HANDLER_FUNC (static, sigabrt_signal_handler)
...
+	if (mono_merp_enabled ())
+		mono_handle_native_crash ("SIGTERM", ctx, info);
+#endif
+
+	mono_chain_signal (MONO_SIG_HANDLER_PARAMS);
+	exit (1);
```
luhenry: Why keeping the `exit (1);`?

alexanderkyte: Without an exit(1), you won't exit? 
I think there's a minor API confusion here. mono_chain_signal only calls another signal handler if it's already been registered. There's no other sigterm handler, so if you put a print between the chain signal and the exit call, it fires. 

```Before chain signal``` And
``` After chain signal, before exit```

Both fire.

This corresponds to the expected state in add_signal_handler, when ```previous_sa.sa_flags & SA_SIGINFO``` is false, and we have no already-registered signal handler to save. 

For this to work the way you want it to, where removing `exit` doesn't make it not `exit`, you'd need to explicitly register the handler at the start of every run that makes mono `exit` when sent sigaction. POSIX just defaults to externally killing a process when there's no signal handler, that doesn't mean that there's automatically a saved signal handler.

luhenry:
It's broken on Android and Windows

luhenry:

luhenry:
@monojenkins build Linux x64

luhenry:

luhenry:

luhenry:

PR #8249, opened 2018-04-14
CoreFX: System.Runtime.Caching

EgorBo:


marek-safar:
The missing types need to come back

PR #8257, opened 2018-04-15
x86/amd64 require a control transfer to notice writes to code

jaykrell:
(citation needed).

Therefore, while mono_arch_flush_icache does not need to do anything, it cannot be inlined.


jaykrell:
Granted, you aren't likely to "fall from" the caller of this into the modified code -- there is a likely a call/ret or at least a ret somewhere already.

PR #8258, opened 2018-04-15
[ppc] make CI a bit more green

lewurm:
linux/ppc64le is one of those targets that can't run roslyn (yet), and thus suffer from runtime tests not being able to compile with `mcs`. Adding a flag to disable those tests under such conditions.

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -682,10 +682,13 @@ TESTS_CS_SRC=		\
 	bug-59400.cs \
 	tailcall-generic-cast-cs.cs \
 	tailcall-interface.cs \
-	sizeof-empty-structs.cs \
 	bug-60843.cs	\
-	roslyn-bug-19038.cs	\
 	nested_type_visibility.cs
+
+# some tests fail to compile on mcs
+if CSC_IS_ROSLYN
+TESTS_CS_SRC += roslyn-bug-19038.cs sizeof-empty-structs.cs
+endif
 
 if AMD64
 TESTS_CS_SRC += async-exc-compilation.cs finally_guard.cs finally_block_ending_in_dead_bb.cs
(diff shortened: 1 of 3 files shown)


lewurm:

mono/tests/Makefile.am
```diff
@@ -682,11 +682,14 @@ TESTS_CS_SRC=		\
 	bug-59400.cs \
 	tailcall-generic-cast-cs.cs \
 	tailcall-interface.cs \
-	sizeof-empty-structs.cs \
 	bug-60843.cs	\
-	roslyn-bug-19038.cs	\
 	nested_type_visibility.cs
 
+# some tests fail to compile on mcs
+if CSC_IS_ROSLYN
```
luhenry: You can use `ifeq ($(CSC),csc)`, that would avoid using another variable in `configure.ac`

lewurm: it's not that simple, `$(CSC)` can for example be `cd $srcdir && pwd`/external/roslyn-binaries/Microsoft.Net.Compilers/Microsoft.Net.Compilers.2.7.0/tools/csc.exe`. it's possible to fiddle around with some text functions, but it's kinda ugly

lewurm:

lewurm:
added more failing tests:
```
tailcall/coreclr/JIT/Methodical/tailcall/gcval.exe
tailcall/coreclr/JIT/Methodical/tailcall/gcval_sideeffect.exe
tailcall/coreclr/JIT/Methodical/tailcall/test_3b.exe
```

lewurm:

lewurm:

mono/tests/Makefile.am
akoeplinger: IMO it would be easier to wrap the test inside of `#if !__MonoCS__` which is a define that is only set by the mcs compiler.

lewurm: @akoeplinger I disagree. In my understanding, `Makefile` serves as some sort of index about what tests are run in a given configuration. Your suggested solution would hide that information, and would require someone to know that a test can be disabled from _within_ a test. That could end up in a confusing state, as for example on CI, the test would still be showed as successfully executed.

PR #8260, opened 2018-04-16
[sdks] Use cmake to build llvm.

vargaz:


sdks/builds/llvm.mk
```diff
@@ -32,9 +32,18 @@ _llvm_$(1)_CONFIGURE_FLAGS= \
 	--enable-targets="arm,aarch64,x86" \
 	$$(if $$(filter $$(UNAME),Darwin),--enable-libcpp)
 
+_llvm_$(1)_CMAKE_FLAGS = \
+	-DCMAKE_INSTALL_PREFIX=$$(TOP)/sdks/out/llvm-$(1) \
+	-DCMAKE_BUILD_TYPE=Release \
+	-DLLVM_TARGETS_TO_BUILD="X86;ARM;AArch64" \
+	-G Ninja \
+	$(LLVM_SRC)
```
luhenry: We need to be able to build 32 and 64 bits versions.

luhenry: It also support building with cmake.

sdks/builds/llvm.mk
```diff
@@ -32,9 +32,18 @@ _llvm_$(1)_CONFIGURE_FLAGS= \
 	--enable-targets="arm,aarch64,x86" \
 	$$(if $$(filter $$(UNAME),Darwin),--enable-libcpp)
 
+_llvm_$(1)_CMAKE_FLAGS = \
+	-DCMAKE_INSTALL_PREFIX=$$(TOP)/sdks/out/llvm-$(1) \
+	-DCMAKE_BUILD_TYPE=Release \
+	-DLLVM_TARGETS_TO_BUILD="X86;ARM;AArch64" \
+	-G Ninja \
```
luhenry: As for Ninja, we don't want to add this forced dependency, because we have no idea if it's supported everywhere we need it to be. Could you replace that by `$$(if $$(shell which ninja),-G ninja)` to make it conditional?

sdks/builds/llvm.mk
luhenry: I have just tried with https://github.com/mono/llvm/commit/bdb3a116dbf4a77ef5f6f70d7f845bdaca2b24ca and using cmake just works.

luhenry:
We should wait on https://github.com/mono/mono/pull/8298 to be merged first, to make sure we didn't break it for Android

vargaz:
This seems to work now both with the normal build, and the mxe build. Only tested with the release_60 branch of llvm.

sdks/builds/Makefile
```diff
@@ -40,10 +41,7 @@ $(TOP)/tools/offsets-tool/MonoAotOffsetsDumper.exe: $(wildcard $(TOP)/tools/offs
 TARGETS=
 
 ## MXE targets
-ifndef DISABLE_ANDROID
```
luhenry: Why removing that? It's only useful for Android.

sdks/builds/llvm.mk
```diff
@@ -23,26 +32,35 @@ _llvm_$(1)_CONFIGURE_ENVIRONMENT= \
...
 	mkdir -p $$(TOP)/sdks/builds/llvm-$(1)
-	cd $$(TOP)/sdks/builds/llvm-$(1) && $$< $$(_llvm_$(1)_CONFIGURE_ENVIRONMENT) $$(_llvm_$(1)_CONFIGURE_FLAGS)
+	cd $$(TOP)/sdks/builds/llvm-$(1) && cmake $$(_llvm_$(1)_CMAKE_FLAGS)
 	touch $$@
 
 .PHONY: package-llvm-$(1)
+
+ifneq ($(NINJA),)
+build-custom-llvm-$(1):
+	ninja -C $$(TOP)/sdks/builds/llvm-$(1)
```
luhenry: This could be `$(if $(NINJA),ninja,make)`, allowing to remove the duplicate https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R52 and https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R58. Same at https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R56

sdks/builds/llvm.mk
```diff
@@ -64,56 +88,38 @@ $(eval $(call LLVMTemplate,llvm64,x86_64))
...
-.stamp-llvm-$(1)-configure: $$(LLVM_SRC)/configure | $(if $(IGNORE_PROVISION_MXE),,provision-mxe)
+.stamp-llvm-$(1)-configure:
 	mkdir -p $$(TOP)/sdks/builds/llvm-$(1)
-	cd $$(TOP)/sdks/builds/llvm-$(1) && PATH="$$$$PATH:$$(_llvm_$(1)_PATH)" $$< $$(_llvm_$(1)_CONFIGURE_ENVIRONMENT) $$(_llvm_$(1)_CONFIGURE_FLAGS)
+	cd $$(TOP)/sdks/builds/llvm-$(1) && $$(llvm_$(1)_CMAKE) $$(_llvm_$(1)_CMAKE_FLAGS)
 	touch $$@
 
-.PHONY: package-llvm-$(1)
+ifneq ($(NINJA),)
+build-custom-llvm-$(1):
```
luhenry: Same as https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R53

sdks/builds/llvm.mk
```diff
@@ -123,6 +129,16 @@ TARGETS += llvm-$(1)
 
 endef
 
+# -DLLVM_ENABLE_THREADS=0 is needed because mxe doesn't define std::mutex etc.
+# -DLLVM_BUILD_EXECUTION_ENGINE=Off is needed because it depends on threads
+# -DCROSS_TOOLCHAIN_FLAGS_NATIVE is needed to compile the native tools (tlbgen) using the host compilers
+llvm_win_configure_flags = -DLLVM_ENABLE_THREADS=0 -DCROSS_TOOLCHAIN_FLAGS_NATIVE=-DCMAKE_TOOLCHAIN_FILE=$(LLVM_SRC)/cmake/modules/NATIVE.cmake -DLLVM_BUILD_EXECUTION_ENGINE=Off
```
luhenry: is the attached `-DDCROSS_TOOLCHAIN_FLAGS_NATIVE=-DCMAKE_TOOLCHAIN_FILE` on purpose? And put these values in `_llvm_$(1)_CMAKE_FLAGS` directly at https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R95

sdks/builds/llvm.mk
```diff
@@ -123,6 +129,16 @@ TARGETS += llvm-$(1)
...
+
+llvm_llvmwin64_CMAKE = $(MXE_PREFIX)/bin/x86_64-w64-mingw32.static-cmake
```
luhenry: Construct this variable at https://github.com/mono/mono/pull/8260/files#diff-e19948cc4c89121d1aa1068fd3d60b01R107 with `$(2)-w64-mingw32$$(if $$(filter $$(UNAME),Darwin),.static)`, since you need to treat both Linux and macOS case.

sdks/builds/Makefile
vargaz: So it can be tested without having an android sdk.

luhenry: If you don't define `DISABLE_ANDROID`, then you can just run `make provision-mxe`.

vargaz: Its not going to define MXE_PREFIX if its not included.

luhenry: I am sorry, I don't understand the issue you are running into. If you define `DISABLE_ANDROID`, then neither mxe.mk neither android.mk are included, so you don't need `MXE_PREFIX` anywhere. If you don't define `DISABLE_ANDROID`, then both mxe.mk and android.mk are included, and `MXE_PREFIX` is defined. Also, in android.mk, `$(MXE_PREFIX)` is going to be lazily evaluated (`=` vs `:=` for assignment). Finally, it's working as-is and I don't see where you are using `MXE_PREFIX` any differently now.

sdks/builds/llvm.mk
vargaz: It could, but these are different tools which happen to have a common subset of command line options.

vargaz: We could probably use cmake --build to invoke the appropriate build tool.

luhenry: It's mostly to avoid repetition, so I would prefer the `$(if $(NINJA),ninja,make)` but that's not-blocking to the PR.

vargaz:

vargaz:

vargaz:

vargaz:
Both android/ios looking good, ios builds+runs device runs with the new llvm.

vargaz:
There are some new failures on the fullaot+llvm lane.

luhenry:

luhenry:
@monojenkins build Linux x64

luhenry:

PR #8261, opened 2018-04-16
[threads] Remove configuration dependency of Hybrid Suspend on Coop Suspend

luhenry:


luhenry:

luhenry:

luhenry:

luhenry:
@monojenkins build Linux AArch64 Hybrid Suspend

luhenry:
@monojenkins build Linux AArch64 Hybrid Suspend

luhenry:

configure.ac
```diff
@@ -4513,26 +4513,30 @@ AC_ARG_WITH(lazy_gc_thread_creation, [  --with-lazy-gc-thread-creation=yes|no
 	fi
 ], [with_lazy_gc_thread_creation=no])
 
-AC_ARG_WITH(cooperative_gc, [  --with-cooperative-gc=yes|no      Enable cooperative stop-the-world garbage collection (sgen only) (defaults to no)], [], [with_cooperative_gc=default])
+AC_ARG_WITH(cooperative_gc,      [  --with-cooperative-gc=yes|no           Enable cooperative stop-the-world garbage collection (sgen only) (defaults to no)], [], [with_cooperative_gc=default])
+AC_ARG_WITH(cooperative_suspend, [  --with-cooperative-suspend=yes|no      Enable cooperative stop-the-world garbage collection (sgen only) (defaults to no)], [], [with_cooperative_suspend=default])
```
akoeplinger: should we mark the `--with-cooperative-gc` option as deprecated and recommend the new one?

configure.ac
```diff
@@ -4513,26 +4513,30 @@ AC_ARG_WITH(lazy_gc_thread_creation, [  --with-lazy-gc-thread-creation=yes|no
...
 
-if test x$with_cooperative_gc = xdefault; then
-	with_cooperative_gc=$with_cooperative_gc_default
+if test x$with_cooperative_suspend = xdefault -a test x$with_cooperative_gc != xdefault; then
+	with_cooperative_suspend=$with_cooperative_gc
```
akoeplinger: if I'm reading the logic correctly we never assign the actual value to `with_cooperative_gc` (only `default` in case the option is omitted)

configure.ac
```diff
@@ -4513,26 +4513,30 @@ AC_ARG_WITH(lazy_gc_thread_creation, [  --with-lazy-gc-thread-creation=yes|no
...
 fi
 
-if test x$with_cooperative_gc != xno; then
-	AC_DEFINE(USE_COOP_GC,1,[Enable cooperative stop-the-world garbage collection.])
+if test x$with_cooperative_suspend = xdefault; then
```
akoeplinger: same here

scripts/ci/run-jenkins.sh
```diff
@@ -40,7 +40,10 @@ if [[ ${CI_TAGS} == *'osx-amd64'* ]]; then CFLAGS="$CFLAGS -m64 -arch x86_64"; L
 if [[ ${CI_TAGS} == *'win-i386'* ]]; then PLATFORM=Win32; EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --host=i686-w64-mingw32"; export MONO_EXECUTABLE="${MONO_REPO_ROOT}/msvc/build/sgen/Win32/bin/Release/mono-sgen.exe"; fi
 if [[ ${CI_TAGS} == *'win-amd64'* ]]; then PLATFORM=x64; EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --host=x86_64-w64-mingw32 --disable-boehm"; export MONO_EXECUTABLE="${MONO_REPO_ROOT}/msvc/build/sgen/x64/bin/Release/mono-sgen.exe"; fi
 
-if [[ ${CI_TAGS} == *'coop-gc'* ]]; then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --with-cooperative-gc=yes"; fi
+if   [[ ${CI_TAGS} == *'coop-gc'* ]]; then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --with-cooperative-suspend=yes";
```
akoeplinger: why do we still need this?

configure.ac
luhenry: We use `with_cooperative_gc` only if it's none default and `with_cooperative_suspend` is default, ie we specified `--with-cooperative-gc=yes|no` but not `--with-cooperative-suspend=yes|no`. If we specified `--with-cooperative-suspend=yes|no`, then we discard `--with-cooperative-gc`.

configure.ac
luhenry: How do we do that? Just put in the message "(deprecated)"?

scripts/ci/run-jenkins.sh
akoeplinger: The only way we can merge this without breaking existing CI is by _adding_ the `coop-suspend` CI_TAGS since earlier branches will still need the `coop-gc` CI_TAGS (remember we only have one config for PR builds) until we no longer care about them.

This means that we don't need the old check since we'd have the new one already.

configure.ac
akoeplinger: Yes. Just so that someone reading this in 6 months doesn't wonder why we have two flags that do the same.

configure.ac
akoeplinger: Ah yeah I misremembered that `AC_ARG_WITH ()` defaults the `action-if-present` to setting the variable to the value. I thought we'd need to set it manually.

configure.ac
luhenry: I added a `AC_MSG_WARN` for `--with-cooperative-gc`

scripts/ci/run-jenkins.sh
luhenry: Fair enough, let me remove it then.

luhenry: I fixed it on CI and we are now passing both `coop-gc` and `coop-suspend`.

akoeplinger: the `${CI_TAGS} == *'coop-gc'*` check is still there :)

akoeplinger: Btw. I assume having both `coop-suspend` and `hybrid-suspend` is a valid case right? We need to do separate `if` instead of `if-else` then.

luhenry: @akoeplinger the push was failing due to a github issue, but should be there now. And having both `coop-suspend` and `hybrid-suspend` at the same time is not valid, because you either use cooperative suspended or hybrid suspend, not both. Let me add this contraint to `configure.ac`.

PR #8265, opened 2018-04-16
[Tailcall amd64] Use calli/tailcall_reg in method-to-ir for F# FullAOT.

jaykrell:
This fixes some F# tests under FullAOT.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -92,6 +92,23 @@ static const gboolean debug_tailcall_break_run = FALSE;     // insert breakpoint
 static const gboolean debug_tailcall_try_all = FALSE;       // consider any call followed by ret
 static const gboolean debug_tailcall = FALSE;               // logging
 
+static gboolean
+tailcall_print_enabled (void)
+{
+	return debug_tailcall || MONO_TRACE_IS_TRACED (G_LOG_LEVEL_DEBUG, MONO_TRACE_TAILCALL);
+}
+
+static void
+tailcall_print (const char *format, ...)
+{
+	if (!tailcall_print_enabled ())
+		return;
+	va_list args;
+	va_start (args, format);
+	g_printv (format, args);
+	va_end (args);
+}
+
 /* These have 'cfg' as an implicit argument */
 #define INLINE_FAILURE(msg) do {									\
 	if ((cfg->method != cfg->current_method) && (cfg->current_method->wrapper_type == MONO_WRAPPER_NONE)) { \
@@ -2230,23 +2247,21 @@ test_tailcall (MonoCompile *cfg, MonoBoolean tailcall)
 	// Do not change "tailcalllog" here without changing other places, e.g. tests that search for it.
 	//
 	g_assertf (tailcall || !mini_get_debug_options ()->test_tailcall_require, "tailcalllog fail from %s", cfg->method->name);
-	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_TAILCALL, "tailcalllog %s from %s", tailcall ? "success" : "fail", cfg->method->name);
+	tailcall_print ("tailcalllog %s from %s\n", tailcall ? "success" : "fail", cfg->method->name);
 }
 
-inline static MonoCallInst *
+static MonoCallInst *
 mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig, 
-					 MonoInst **args, int calli, int virtual_, int tailcall, int rgctx, int unbox_trampoline, MonoMethod *target)
+		     MonoInst **args, gboolean calli, gboolean virtual_, gboolean tailcall,
+		     gboolean rgctx, gboolean unbox_trampoline, MonoMethod *target)
 {
 	MonoType *sig_ret;
 	MonoCallInst *call;
-#ifdef MONO_ARCH_SOFT_FLOAT_FALLBACK
-	int i;
-#endif
 
-	if (cfg->llvm_only) {
-		if (tailcall)
-			test_tailcall (cfg, FALSE);
+	if (tailcall && cfg->llvm_only) { // FIXME?
 		tailcall = FALSE;
+		tailcall_print ("losing tailcall in %s due to llvm_only\n", cfg->method->name);
+		test_tailcall (cfg, FALSE);
 	}
 
 	if (tailcall && (debug_tailcall_break_compile || debug_tailcall_break_run)
@@ -2264,7 +2279,7 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 
 	if (tailcall) {
 		mini_profiler_emit_tail_call (cfg, target);
-		MONO_INST_NEW_CALL (cfg, call, virtual_ ? OP_TAILCALL_MEMBASE : OP_TAILCALL);
+		MONO_INST_NEW_CALL (cfg, call, calli ? OP_TAILCALL_REG : virtual_ ? OP_TAILCALL_MEMBASE : OP_TAILCALL);
 	} else
 		MONO_INST_NEW_CALL (cfg, call, ret_type_to_call_opcode (cfg, sig->ret, calli, virtual_));
 
@@ -2314,7 +2329,7 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 		 * an icall, but that cannot be done during the call sequence since it would clobber
 		 * the call registers + the stack. So we do it before emitting the call.
 		 */
-		for (i = 0; i < sig->param_count + sig->hasthis; ++i) {
+		for (int i = 0; i < sig->param_count + sig->hasthis; ++i) {
 			MonoType *t;
 			MonoInst *in = call->args [i];
 
@@ -2366,20 +2381,26 @@ set_rgctx_arg (MonoCompile *cfg, MonoCallInst *call, int rgctx_reg, MonoInst *rg
 #endif
 }	
 
-MonoInst*
-mini_emit_calli (MonoCompile *cfg, MonoMethodSignature *sig, MonoInst **args, MonoInst *addr, MonoInst *imt_arg, MonoInst *rgctx_arg)
+static MonoInst*
+mini_emit_calli_full (MonoCompile *cfg, MonoMethodSignature *sig, MonoInst **args, MonoInst *addr,
+	MonoInst *imt_arg, MonoInst *rgctx_arg, gboolean tailcall)
 {
 	MonoCallInst *call;
 	MonoInst *ins;
 	int rgctx_reg = -1;
 	gboolean check_sp = FALSE;
 
+	// This logic is duplicated in mini_emit_calli_full and is_supported_tailcall,
+	// in order to compute tailcall_supported earlier. Alternatively it could be passed
+	// out from here -- assuming it has not been copied around
+	// or decisions made based on it.
 	if (cfg->check_pinvoke_callconv && cfg->method->wrapper_type == MONO_WRAPPER_MANAGED_TO_NATIVE) {
 		WrapperInfo *info = mono_marshal_get_wrapper_info (cfg->method);
 
-		if (info && info->subtype == WRAPPER_SUBTYPE_PINVOKE)
-			check_sp = TRUE;
+		check_sp = info && info->subtype == WRAPPER_SUBTYPE_PINVOKE;
 	}
+
+	g_assert (!check_sp || !tailcall);
 
 	if (rgctx_arg) {
 		rgctx_reg = mono_alloc_preg (cfg);
@@ -2395,7 +2416,7 @@ mini_emit_calli (MonoCompile *cfg, MonoMethodSignature *sig, MonoInst **args, Mo
 		MONO_ADD_INS (cfg->cbb, ins);
 	}
 
-	call = mono_emit_call_args (cfg, sig, args, TRUE, FALSE, FALSE, rgctx_arg ? TRUE : FALSE, FALSE, NULL);
+	call = mono_emit_call_args (cfg, sig, args, TRUE, FALSE, tailcall, rgctx_arg ? TRUE : FALSE, FALSE, NULL);
 
 	call->inst.sreg1 = addr->dreg;
 
@@ -2428,12 +2449,19 @@ mini_emit_calli (MonoCompile *cfg, MonoMethodSignature *sig, MonoInst **args, Mo
 	return (MonoInst*)call;
 }
 
+MonoInst*
+mini_emit_calli (MonoCompile *cfg, MonoMethodSignature *sig, MonoInst **args, MonoInst *addr, MonoInst *imt_arg, MonoInst *rgctx_arg)
+// Historical version without gboolean tailcall parameter.
+{
+	return mini_emit_calli_full (cfg, sig, args, addr, imt_arg, rgctx_arg, FALSE);
+}
+
 static MonoInst*
 emit_get_rgctx_method (MonoCompile *cfg, int context_used, MonoMethod *cmethod, MonoRgctxInfoType rgctx_type);
 
 static MonoInst*
 mono_emit_method_call_full (MonoCompile *cfg, MonoMethod *method, MonoMethodSignature *sig, gboolean tailcall,
-							MonoInst **args, MonoInst *this_ins, MonoInst *imt_arg, MonoInst *rgctx_arg)
+			    MonoInst **args, MonoInst *this_ins, MonoInst *imt_arg, MonoInst *rgctx_arg)
 {
 #ifndef DISABLE_REMOTING
 	gboolean might_be_remote = FALSE;
@@ -2448,7 +2476,7 @@ mono_emit_method_call_full (MonoCompile *cfg, MonoMethod *method, MonoMethodSign
 	if (!sig)
 		sig = mono_method_signature (method);
 
-	if (cfg->llvm_only && (mono_class_is_interface (method->klass)))
+	if (cfg->llvm_only && mono_class_is_interface (method->klass))
 		g_assert_not_reached ();
 
 	if (rgctx_arg) {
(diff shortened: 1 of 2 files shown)


mono/mini/method-to-ir.c
```diff
@@ -2479,7 +2506,7 @@ mono_emit_method_call_full (MonoCompile *cfg, MonoMethod *method, MonoMethodSign
 
 		addr = emit_get_rgctx_method (cfg, context_used, method, MONO_RGCTX_INFO_REMOTING_INVOKE_WITH_CHECK);
 
-		return mini_emit_calli (cfg, sig, args, addr, NULL, NULL);
+		return mini_emit_calli (cfg, sig, args, addr, NULL, NULL, FALSE);
 	}
```
vargaz: It might be better to add a emit_calli_tailcall helper since most places pass FALSE.

jaykrell: mini_emit_calli original and mini_emit_calli_full?

mono/mini/method-to-ir.c
```diff
@@ -92,6 +92,17 @@ static const gboolean debug_tailcall_break_run = FALSE;     // insert breakpoint
 static const gboolean debug_tailcall_try_all = FALSE;       // consider any call followed by ret
 static const gboolean debug_tailcall = FALSE;               // logging
 
+static void
+tailcall_print (const char *format, ...)
```
jaykrell: This has an affect of adding a few more strings to the image than before, because the optimizer was previously presented with a static const false and now an environment variable. Ok?

mono/mini/method-to-ir.c
```diff
@@ -7138,8 +7165,8 @@ is_jit_optimizer_disabled (MonoMethod *m)
 static gboolean
 is_not_supported_tailcall_helper (gboolean value, const char *svalue, MonoMethod *method, MonoMethod *cmethod)
 {
-	if (value && debug_tailcall)
-		g_print ("%s %s -> %s %s:true\n", __func__, method->name, cmethod->name, svalue);
+	if (value)
```
jaykrell: This is where optimizer will now fail and add more strings to the code vs. before. ok?

jaykrell:
Not sure why the arm32 failures, kinda related.

mono/mini/method-to-ir.c
```diff
@@ -2263,7 +2265,7 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 		no_tailcall = "llvm_only";
 
 	if (tail && no_tailcall) {
-		tail = FALSE;
+		*ptailcall = FALSE;
```
jaykrell: Also need tail =FALSE here.

mono/mini/method-to-ir.c
```diff
@@ -2246,41 +2246,39 @@ test_tailcall (MonoCompile *cfg, MonoBoolean tailcall)
 
 static MonoCallInst *
 mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig, 
-		     MonoInst **args, gboolean calli, gboolean virtual_, gboolean *ptailcall,
```
vargaz: Wouldn't be better to do the cfg->check_pinvoke_callconv check in CEE_CALLI  so tailcall doesn't have to be an out argument anymore ? These checks should be done on an upper level and not emit_calli which is a low level function.

jaykrell: I was thinking the opposite -- put them lower -- but put them in one place.
Maybe. You are right if they are earlier/higher than it doesn't change as much and doesn't have to be passed back out. Blocking? I'm debugging other related problems.. but I'll look..

jaykrell: Maybe do it in is_supported_tail_call? Still a bit disjointed, because that means is_supported_tail_call understands the implication that check_sp is needed.

jaykrell: Not so simple, because the decision of call vs. calli is spread around.

jaykrell: One way, maybe gross, to bottleneck it, is is_supported_tail_call return bitfields -- is call supported, is calli supported..

jaykrell: mono_emit_call_args was already making a decision, based on llvm_only, and moving this up..does it apply across every single call?

jaykrell: The out parameter as I did it, keeps various decision making where it was, and avoids spreading out decisions, except for tailcall decisions. Kinda tough to have it all ways -- can't have everything in one place.

vargaz: is_supported_tail_call () seems like a good place to do it. just add a g_assert (!tailcall) to the 
check_sp code.

jaykrell: For check_sp I have that coded..it isn't too too terrible, but slightly bad, put comments in both, and the assert as you say...but what about the llvm_only check?

And is the check_sp logic specific to calli? I think not so ok.

jaykrell: To redundantly put it -- mini_emit_calli_full seems ok to fix, maybe, the logic can go in is_tail_call_supported I think. But that still leaves mono_emit_call_args.
It is worse there, because we don't know it is calli currently at that point.

vargaz: emit_call_args () used to disable tail calls because the caller didn't need the information, if it does now, the check should be done in the caller, or the caller's caller. There are only two places which can pass tail==TRUE to that function.

jaykrell: I understand what you mean, but it is messy. I believe it is more complicated than just the two places.

check_sp involves duplicating code but I understand what to do.
Because this is about the function being called.

But something like:
mono_emit_call_args:
	if (calli && !cfg->backend->have_op_tail_call_reg)
		no_tailcall = "architecture lacking tail_call_reg";

is messier because higher up we aren't even doing a calli.
But we need/want to omit the return code after the call, that got turned into calli.
When I left in the return code, I hit assert(reg > -1).

I think what we'd have to do, to kinda satisfy you is right before mini_emit_calli_full, we'd call something like is_tail_call_supported, but another variation of it. is_tailcall_supported_calli.

Or in fact, the same function, but add a calli parameter to it, false on the first call, and then later try again with true. Or pass in CEE_CALLI as a close enough approximation.

I can try this but I don't think the result will be particularly cleaner than what I have.

jaykrell:
@monojenkins build failed

jaykrell:
coop break might also be related.

jaykrell:
tailcall_is_supported(calli == true) needs to be called in more places in order to take the place of emit.. like ahead of each one, I think... messy, more than I expected.

mono/mini/method-to-ir.c
```diff
@@ -2230,41 +2241,35 @@ test_tailcall (MonoCompile *cfg, MonoBoolean tailcall)
...
 			MONO_ADD_INS (cfg->cbb, brk);
 		}
 	}
 
-	if (tail) {
+	if (tailcall) {
 		mini_profiler_emit_tail_call (cfg, target);
-		MONO_INST_NEW_CALL (cfg, call, virtual_ ? OP_TAILCALL_MEMBASE : OP_TAILCALL);
+		MONO_INST_NEW_CALL (cfg, call, calli ? OP_TAILCALL_REG : virtual_ ? OP_TAILCALL_MEMBASE : OP_TAILCALL);
 	} else
```
vargaz: This could use a parentheses.

jaykrell: It works intuitively but if you insist..

jaykrell: Is the pinvoke_callconv thing calli-specific? It is messier to keep it so but I can.

jaykrell:
arm problem is moving is_tailcall_supported earlier

We interleave computings its inputs and acting on them, somewhat.

Three possible approaches:
 1. compute all inputs
  then act
This is best, but the biggest change.

2. partially reevaluate as new inputs become available or change (see how virtual_ is assigned a few times)
This is efficient but a little messy.
I tried it and got it somewhat wrong.

3. Fully reevaluate a "few" times as inputs become available, intersected with places in the code where we can use an up to date value. Initially "few" is probably..statically as many as four places in the code but dynamically anywhere 1-3. This is a smaller and easier to understand and somewhat-easier-tio-maintain change, but less efficient. The inefficiency would be gated by tail. prefix so never seen compiling C#, for example. Though we should "auto tailcall" C# more in future.

4. Capture locals by address and evaluate whenever we pass tailcall parameter.
This potentially is easiest to maintain and most efficient (aside from option 1), and not a huge change.
Efficiency depends on passing tailcall few times (once) or memoizing it.

jaykrell:
To be clear, this is the problem:
```commit b75447a0c1a784ad04267e5bf90e928df4666015
Author: Jay Krell <jaykrell@microsoft.com>
Date:   Mon Apr 16 02:30:07 2018 -0700

    [Tailcall] Move checks earlier for accurate logging and maybe fixes.
```

jaykrell:
Option 3 tempting. Put the moved code into a macro and call it a small number of times more than 1.

vargaz:
Just compute it once before the first place where it is really used.

jaykrell:
Computing it once before first use is a big change, as the inputs keep changing, after use. That is why it is broken now.

The inputs that change are virtual_ and vtable_arg.
Also I'd like to replace is_interface with imt_arg, which is more accurate and is the point, but also does change.

If you consider the prints to only need the approximate inst_tailcall, at least for now, that should help.
If you consider that calli is not sensitive to some of the inputs (vtable_arg/interface/imt_arg), that might help too.
I am going to try those avenues.

However, many of the prints are potential later "real" uses which will increase the "pressure".

jaykrell:
I'm settling on, of course, a compromise.
The one function, with additional complexity, wrapped in a macro, that we evaluate in about three places statically, typically dynamically once. The debugprints are inaccurate. is_interface changed to imt_arg. We'll see how it fairs in CI and PR...

mono/mini/cpu-amd64.md
```diff
@@ -57,8 +57,9 @@
 #
 
 break: len:2
-tailcall: len:120 clob:c
-tailcall_membase: src1:b len:120 clob:c # FIXME len?
+tailcall: len:255 clob:c
+tailcall_reg: src1:b len:255 clob:c # FIXME len?
```
jaykrell: I thought these were in master already.

mono/mini/cpu-x86.md
```diff
@@ -64,8 +64,8 @@
 #
 break: len:1
 call: dest:a clob:c len:17
-tailcall: len:120 clob:c
-tailcall_membase: src1:b len:120 clob:c # FIXME len?
+tailcall: len:255 clob:c
+tailcall_membase: src1:b len:255 clob:c # FIXME len?
```
jaykrell: I thought this in master already.

jaykrell:
Geez, nice try I thought, but it is worse now. Still iterating..

jaykrell:
I think this is actually going to work.

jaykrell:

jaykrell:
@vargaz if this works, are you generally ok with it? Or horrible to call is_supported_tailcall more than once or from more than one place? (more than once is imho worse than from more than one place).

To further clean it up, I think, some of the ifs should be captured into variables or functions, for duplicate evaluation, to split up computation from side effects. Can attempt this but, would rather not until after this PR.

vargaz:
Couldn't you disable tailcalls when the variables it depends on changes, like when setting vtable_reg, just follow it with tailcall = FALSE; ?

jaykrell:
It is difficult to do this w/o exposing to the caller which variables matter.
There approximately two ideals here.

1. Only one call. Then that call has the obligation to know. Once there are multiple calls, I don't like that.
I think this might be achievable but I'd rather not try this week. i.e. by moving more of the code around.

2. Put the address of the relevant locals in a struct "MethodToIrTailcall" and whenever we reference tailcall local, instead method_to_ir_tailcall_get(), which conceptually reevaluates the value, but actually what it'd do is cache the values of the locals and the result and only reevaluate if any changed. Many people would say this is complicated but it is not and it cuts down the complexity and cost of most other solutions.

The present solution is a compromise. It is generally "simpler" and "faster", except it spreads knowledge around -- which variables are relevant, when to reevaluate -- and it does a full evaluation when it might  not.

jaykrell:
@monojenkins build failed

mono/mini/method-to-ir.c
```diff
@@ -7173,13 +7213,28 @@ is_supported_tailcall (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmethod
...
 		// Interface method dispatch has the same problem.
-		//
-		|| IS_NOT_SUPPORTED_TAILCALL ((vtable_arg || is_interface) && !cfg->backend->have_volatile_non_param_register)
-		)
-		return FALSE;
+		|| IS_NOT_SUPPORTED_TAILCALL ((vtable_arg || imt_arg) && !cfg->backend->have_volatile_non_param_register)
 
-	MonoMethodSignature *caller_signature = mono_method_signature (method);
-	MonoMethodSignature *callee_signature = mono_method_signature (cmethod);
+		// FIXME?
```
jaykrell: This should be stricken but I don't want to reset CI.

jaykrell:
CI good. @vargaz ok?

mono/mini/method-to-ir.c
```diff
@@ -2366,21 +2381,27 @@ set_rgctx_arg (MonoCompile *cfg, MonoCallInst *call, int rgctx_reg, MonoInst *rg
...
 {
 	MonoCallInst *call;
 	MonoInst *ins;
 	int rgctx_reg = -1;
 	gboolean check_sp = FALSE;
 
+	// This logic is duplicated in mini_emit_calli_full and is_supported_tailcall,
+	// in order to compute tailcall_supported earlier. Alternatively it could be passed
+	// out from here -- assuming it has not been copied around
+	// or decisions made based on it.
```
jaykrell: Should probably be a function. Shoud_check_stack().

mono/mini/method-to-ir.c
```diff
@@ -8902,58 +8974,82 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
+			const gboolean tailcall_calli = tailcall_calli_temp;
+			// Writes to imt_arg, vtable_arg, virtual_, cmethod, must not occur from here (inputs to is_supported_tailcall).
+
+			if (virtual_generic) {
+				if (virtual_generic_imt) {
 					if (tailcall) {
 						/* Prevent inlining of methods with tailcalls (the call stack would be altered) */
 						INLINE_FAILURE ("tailcall");
+						// FIXME? leaks imt_arg, vtable_arg?
 					}
```
vargaz: This won't leak anything.

mono/mini/method-to-ir.c
```diff
@@ -8902,58 +8974,82 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
+			/* FIXME: runtime generic context pointer for jumps? */
+			/* FIXME: handle this for generic sharing eventually */
+
+			// http://www.mono-project.com/docs/advanced/runtime/docs/generic-sharing/
+			// 1. Non-generic non-static methods of reference types have access to the
+			//    RGCTX via the “this” argument (this->vtable->rgctx).
+			// 2. a Non-generic static methods of reference types and b. non-generic methods
+			//    of value types need to be passed a pointer to the caller’s class’s VTable in the MONO_ARCH_RGCTX_REG register.
+			// 3. Generic methods need to be passed a pointer to the MRGCTX in the MONO_ARCH_RGCTX_REG register
+			//
```
vargaz: What do these comments refer to ?

mono/mini/method-to-ir.c
```diff
@@ -8902,58 +8974,82 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
+
 				/* Prevent inlining of methods that contain indirect calls */
 				INLINE_FAILURE ("virtual generic call");
 
 				if (cfg->gsharedvt && mini_is_gsharedvt_signature (fsig))
 					GSHAREDVT_FAILURE (*ip);
 
 				if (cfg->backend->have_generalized_imt_trampoline && cfg->backend->gshared_supported && cmethod->wrapper_type == MONO_WRAPPER_NONE) {
+
+					virtual_generic_imt = TRUE;
```
vargaz: Extra empty line.

mono/mini/method-to-ir.c
```diff
@@ -8902,58 +8974,82 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
 					g_assert (!imt_arg);
-					if (!context_used)
-						g_assert (cmethod->is_inflated);
+					g_assert (context_used || cmethod->is_inflated);
+
```
vargaz: This might have the same meaning, but its harder to understand.

jaykrell: It should make good sense, that the second part depends on the first, and it is a longer string in the assert failure to correlate with the source (but also larger bloat).

vargaz:
Looks ok otherwise.

mono/mini/method-to-ir.c
jaykrell: They are from the online docs but I agree maybe a bit large size/value ratio.
They aren't add here, they are a move. Git does not seem to ever indicate movement, just del and add. I can remove them.

mono/mini/method-to-ir.c
jaykrell: ok, comment should be removed

mono/mini/method-to-ir.c
jaykrell: The comment was also better placed before more logic moved into the function. Now it only kinda helps explain what parameters are useful. I agree it isn't great. I feel like the area could use heavy commenting for casual readers, or maybe casual reading is not possible Or maybe the arm thing in the function is the point. Not sure all around.

jaykrell: Maybe the only viable path is what I took: ask questions, find docs; code and comments can only go so far. Or maybe that is just my inability to read it.

mono/mini/method-to-ir.c
jaykrell: removed in upcoming commit either way

vargaz: Extra newlines are ok, but putting them after ifs doesn't really add anything imho.

jaykrell: on other parts, I'm not entirely happy with all the fixmes or their elaborate comments.
They are an attempt at something.

The print(...tailcall skipped...) might suffice. Or maybe even remove them, maybe.
There is already:
```
			if (common_call) // FIXME goto call_end && !common_call often skips tailcall processing.
```
maybe that suffices -- you then go and look for all the gotos and try to figure it out, and "try to figure it out" might go better than reading my comments; meanwhile things aren't exactly falling apart either.

Though the prints did help. You can "often" run a test, look for the print, it pinpoints the problem. Often as in, well, once or twice.

mono/mini/method-to-ir.c
jaykrell: We used to check gshared here, which connected the code to the comments a little better. Now the code is more concerned specifically with the extra vtable/imt parameters. I'll try to polish it a little.

jaykrell: Ok I moved the comment and tried to wordsmith it more. If you just want to delete it all that's ok too. It does not clearly add any value.

mono/mini/method-to-ir.c
```diff
@@ -7188,24 +7264,30 @@ is_supported_tailcall (MonoCompile *cfg, MonoMethod *method, MonoMethod *cmethod
...
-	if (!mono_debug_count ())
-		return FALSE;
+	if (!mono_debug_count ()) {
+		tailcall_calli = FALSE;
+		tailcall = FALSE;
+		goto exit;
+	}
 #endif
+	// See check_sp in mini_emit_calli_full.
+	tailcall_calli = tailcall_calli && !IS_NOT_SUPPORTED_TAILCALL (should_check_stack_pointer (cfg));
```
jaykrell: I try not to write double negatives, but there is one. !IS_NOT...

jaykrell:
@monojenkins Linux AArch64 Interpreter

jaykrell:
@monojenkins build failed

jaykrell:

PR #8268, opened 2018-04-16
Update CMake version for BTLS

luhenry:


akoeplinger:
what's the reason for this change?

luhenry:
@akoeplinger no reason in particular, I was just going over the cmake version BTLS was using and it turns out we can use a later one.

akoeplinger:
hm that doesn't sound like a compelling reason to take this patch :)

luhenry:
@akoeplinger fair enough 😄

PR #8270, opened 2018-04-16
[wasm] JS engine testing improvements.

kumpera:
This PR switch from manual compiling of V8 to using jsvu for engine provisioning.
Plus it adds jsc, ch and sm testing as it's all super easy now.

Finally, add ch and sm testing to CI as they work fine on OSX.

This PR might break CI as I'm not 100% we have node.js installed on the bots.

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -17,7 +17,8 @@ MINI_PATH=$(TOP)/mono/mini
 BCL_ASSEMBLIES= \
         mscorlib.dll    \
         System.dll      \
-        System.Core.dll
+        System.Core.dll	\
+        Mono.Security.dll
 
 DEPS_ASSEMBLIES= \
         nunitlite.dll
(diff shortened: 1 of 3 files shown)


sdks/wasm/Makefile
```diff
@@ -138,9 +129,29 @@ build-managed: managed/nunitlite.dll managed/mini_tests.dll managed/main.exe man
 
 build: build-native build-managed
 
-run-%: toolchain build
+CHACKA=~/.jsvu/ch
```
akoeplinger: typo: should be CHAKRA

kumpera:
@monojenkins build

sdks/wasm/test.js
```diff
@@ -1,3 +1,24 @@
+//glue code to deal with the differences between ch, d8, jsc and sm.
+if (print == undefined)
```
akoeplinger: should probably use `===`

PR #8271, opened 2018-04-16
[Tailcall] Generally remove underscores and spaces in "tail_call", "T…

jaykrell:
…AIL_CALL", "tail call", etc.

Also standalone "tail" to "tailcall".
For consistency and ease of search -- "longer words are better".
Leave profiler alone generally.

This is pure line damage, yes, and makes the code easier to search and maintain.


lewurm:
line damage is okay if it isn't mixed with actual changes imho

PR #8276, opened 2018-04-17
Fix Gdi+ WrongState exception type and remove incorrect test for GdiPlus status code

hughbe:
This now matches the Windows behaviour: https://github.com/dotnet/corefx/blob/0eb5e7451028e9374b8bb03972aa945c128193e1/src/System.Drawing.Common/src/System/Drawing/Gdiplus.cs#L289-290

The exception code for locking an already locked bitmap is incorrect, as demonstrated here: https://github.com/dotnet/corefx/blob/ec22edc155aa13289770c9f49630cc7ab588fca4/src/System.Drawing.Common/src/System/Drawing/Bitmap.cs#L244-L248

This has been fixed in https://github.com/mono/libgdiplus/pull/367

@akoeplinger FYI

### DIFF
--- a/mcs/class/System.Drawing/System.Drawing.Drawing2D/LinearGradientBrush.cs
+++ b/mcs/class/System.Drawing/System.Drawing.Drawing2D/LinearGradientBrush.cs
@@ -36,6 +36,7 @@ namespace System.Drawing.Drawing2D {
 	public sealed class LinearGradientBrush : Brush
 	{
 		RectangleF rectangle;
+		private bool _interpolationColorsWasSet;
 		
 		internal LinearGradientBrush (IntPtr native)
 		{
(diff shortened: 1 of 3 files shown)


luhenry:
The exception change is breaking some tests:
 - LinearGradientBrushTest.InterpolationColors_Colors_InvalidBlend
 - LinearGradientBrushTest.InterpolationColors_Positions_InvalidBlend

They both expect the `ArgumentException` instead of the `InvalidOperationException`, please verify the test behaviour on .NET Framework and update the test (or code) accordingly.

Thank you.

hughbe:
The issue here is that the lineargradientbrush product code doesn't have the same validation as .NET Core or .NET Framework:

https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Advanced/LinearGradientBrush.cs,9384d19bfda49bbc,references
https://github.com/dotnet/corefx/blob/eb851a2b0243859be21e0d9d6fc3c4a4c5f9c7c5/src/System.Drawing.Common/src/System/Drawing/Drawing2D/LinearGradientBrush.cs#L381-L382

PR #8281, opened 2018-04-17
[threads] Check that thread is in a valid suspend state before notifying

lambdageek:


lambdageek:

lambdageek:
@monojenkins build Linux AArch64 Hybrid Suspend

lambdageek:

lambdageek:
@monojenkins build Linux AArch64 Hybrid Suspend

lambdageek:
@luhenry I did a bad job reviewing 282ee4bfd7ea91dc0015891010c71d8bd1421bc8 - in `begin_suspend_for_running_thread` we want to use coop for both full cooperative suspend and hybrid suspend.

I took out the atomic loads. I don't think they do anything here.

lambdageek:

lambdageek:
@monojenkins build Linux AArch64 Hybrid Suspend

lambdageek:
Some of the hybrid suspend hangs look like they're deadlocks in `backtrace()`: two threads are doing thread transitions at the same time and both try to capture backtraces to record in the checked thread transition histories.

Googling "`_Unwind_Find_FDE` deadlock" finds a lot of discouraging info.

lambdageek:

lambdageek:
@luhenry @kumpera another review?

Two new things:
1. A hacky mutex `trylock` around `backtrace` in checked builds to work around `backtrace` internally taking a mutex (there's a comment with a timeline of an example in the code). We trylock and yield a couple of times to do our best to get a stack trace and if there's contention, just give up.
2. Change who's responsible for notifying the suspend initiator when in hybrid suspend we have a race between done/abort blocking and the suspend signal.

   Previously: done blocking would notify the initiator and then wait for a resume - which means if the signal arrives after we've resumed we do an illegal `finish_async_suspend` transition when we're already back in `RUNNING`.

   Now: done blocking just waits for the resume - signal handler must run and notify the initiator and then immediately return).

lambdageek:

Issue #8267, opened 2018-04-16
TimeZoneInfo.Local Id and DisplayName are always: Local


bruno-garcia:
## Steps to Reproduce

1. Call TimeZoneInfo.Local while running on Mono

## Current Behavior

Returns:

```
TimeZoneInfo.Local.DisplayName:                         Local
TimeZoneInfo.Local.Id:                                  Local
```

## Expected Behavior
Data that can be used to pin the device's timezone.

CoreCLR returns something more meaningful:

*On macOS 10.12.2*
Darwin Brunos-MacBook-Pro.local 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64

```
// .NET Core 2.0.6:
TimeZoneInfo.Local.DisplayName:                       GMT+01:00
TimeZoneInfo.Local.Id:                                Europe/Vienna

// Mono JIT compiler version 5.8.0.129 (2017-10/835a98d8559 Fri Feb 23 21:09:37 EST 2018)
TimeZoneInfo.Local.DisplayName:                        Local
TimeZoneInfo.Local.Id:                                 Local
```

*On CentOS 7*
Linux centos-7.shared 3.10.0-693.5.2.el7.x86_64 #1 SMP Fri Oct 20 20:32:50 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
// .NET Core 2.0.6:
TimeZoneInfo.Local.DisplayName:                       Central European Standard Time
TimeZoneInfo.Local.Id:                                Europe/Vienna

// Mono JIT compiler version 5.10.1.20 (tarball Thu Mar 29 14:44:43 UTC 2018)
TimeZoneInfo.Local.DisplayName:                         Local
TimeZoneInfo.Local.Id:                                  Local
```

*On Kali under WSL*
Linux FLORIPA 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014 x86_64 GNU/Linux

*This seems to be a separate issue. /etc/timezone is not a symlink on WSL.*

```
// .NET Core 2.0.6:
TimeZoneInfo.Local.DisplayName:                          GMT
TimeZoneInfo.Local.Id:                                   Msft/localtime

// Mono JIT compiler version 4.6.2 (Debian 4.6.2.7+dfsg-1)
TimeZoneInfo.Local.DisplayName:                         Local
[...]

PR #8283, opened 2018-04-17
[corlib] Fix TimeZoneInfo.Local when /usr/share/zoneinfo is a symlink

akoeplinger:
We'd be using the path verbatim in `TryGetNameFromPath()` to compare with the resolved value of
`/etc/localtime` but if it's a symlink (like on OSX 10.13+) the comparison failed.

The fix is to resolve any symlinks for /usr/share/zoneinfo too.

Fixes https://github.com/mono/mono/issues/8267 and https://github.com/xamarin/maccore/issues/628

### DIFF
--- a/mcs/class/corlib/System/TimeZoneInfo.cs
+++ b/mcs/class/corlib/System/TimeZoneInfo.cs
@@ -266,11 +266,12 @@ namespace System
 			}
 		}
 #if LIBC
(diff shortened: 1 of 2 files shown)


filipnavara:
Should be backported to 2017-12, 2018-02 and 2018-04 once approved.

akoeplinger:
@monojenkins backport 2018-04

akoeplinger:
@monojenkins backport 2018-02

PR #8291, opened 2018-04-17
[corlib] SynchronizationContext.Current does not work on non-root AppDomain on Android

luhenry:
AndroidEnvironment.GetDefaultSyncContext just crashes when called on an AppDomain which is not the root AppDomain. This started appearing with the mono:2018-02 integration during which https://github.com/mono/mono/commit/252d61d0558b0bfa3461f3339ce617e2e4c285bb was introduced.

Fixes https://github.com/xamarin/xamarin-android/issues/1561


luhenry:
This is blocking any further testing for `2018-02` for XA.

akoeplinger:
But we're still not sure why we have another AppDomain right?

kg:
This probably should work in other appdomains, it's not a scenario I tested extensively. I still think our implementation of sync contexts is wrong but I wasn't able to figure it out in enough depth to build out a perfect solution. Maybe worth going back to it. netcore and netframework do it very differently from us

PR #8231, opened 2018-04-13
Fix timestamp for file copy

lewurm:
4a8d840 introduced that the destination
file of a copy operation has the same modification and access timestamps
as the source file.

With e0e185d this behaviour regresses,
since getting timestamps provides more precise information. Fixing this
by being more accurate when applying timestamps to destination file.

also: convert nano seconds to microseconds properly, by using `TICKS_PER_MILLISECOND` (which has L suffix in its definition)

fixes https://github.com/xamarin/xamarin-macios/issues/3929

needs backporting to `2018-02` and `2018-04`

PR #8293, opened 2018-04-17
[w32file-unix] fix mistake in 8430155946c31f0f3f5f04017eebd79ac15e566d

lewurm:
[w32file-unix] fix mistake in 8430155946c31f0f3f5f04017eebd79ac15e566d

commit message said
> [w32file-unix] convert nano seconds to microseconds properly, by using
> TICKS_PER_MILLISECOND (which has L suffix in its definition)

first it talks about microseconds, but then uses constant for _milli_
seconds

fixes regression introduced and mentioned here: https://github.com/mono/mono/pull/8231#issuecomment-381900732


lewurm:
@monojenkins backport 2018-02

lewurm:
@monojenkins backport 2018-04

lewurm:
@akoeplinger I thought about checking the difference being e.g. <1s, but I'm afraid such a test is too fragile

jaykrell:
Can we run touch to set a file's time and then read it back?

jaykrell:
Or a little bit of C?

Issue #8290, opened 2018-04-17
Only support one of dynamic linking or static linking for LLVM into Mono


luhenry:
We support 2 ways of linking Mono with LLVM:
 - static linking
 - dynamic linking (== "loaded" llvm)

Dynamic linking has the usual advantages with static linking, plus an avoided 0.1s added to Mono startup time (most likely static constructors) even when not using LLVM in the case of static linking.

As part of the work on distributing LLVM on Linux as part of Mono, we should settle on one of the 2 mechanism for the sake of simplifying our build. If choosing the dynamic linking approach, we need to make sure the packaging story is taken care of.

This is following discussion at https://xamarinhq.slack.com/archives/C98FGEDJR/p1523987857000829

/cc @kumpera @vargaz

vargaz:
The dynamic linking approach also has the disadvantage of complicating runtime development since running mono --llvm will pick up a random libmonollvm.so from somewhere on the system.

kumpera:
What if we fix our in-tree dev story for probing for assemblies & DSOs?

Cuz we either completely give up on being able to use mono/mini/mono and depend exclusively on runtime/mono-wrapper or we add enough probing under desktop.

I rather have the later, of adding the smarts to the runtime to probe enough additional locations on disk if it finds itself in a build tree.

jaykrell:
@vargaz I suggest using @origin to strongly guide the search -- the .sos must live in the same directory. Adequate then? Mac has a similar mechanism, need to make sure it is so to so (dylib) and not just executable to so (dylib).

Otherwise explicit dlopen+dlsym might be ok, if we have only a few imports, or automate.

Or, is there a model of dlopen-only, and then let the dynamic linker do all the dlsym? As long as you dlopen before it goes looking?

I do not want to be beholden to mono-wrapper. I avoid it all the time.

jaykrell:
I believe for Mac, @loader_path is the answer -- dylib to dylib, vs. @executable_path which would be main executable to dylib. Or if "all three" (mono, libmono, libmonollvm) in same directory, then either, that is a good solution imho, though I know people like to put so/dylib in lib instead of bin..

PR #8294, opened 2018-04-17
[configure.ac] Remove --enable-loadedllvm

luhenry:
Fixes https://github.com/mono/mono/issues/8290

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3219,9 +3219,8 @@ dnl ***  LLVM  ***
 dnl **************
 
 AC_ARG_ENABLE(llvm,[  --enable-llvm	Enable the LLVM back-end], enable_llvm=$enableval, enable_llvm=default)
-AC_ARG_ENABLE(loadedllvm,[  --enable-loadedllvm	Load the LLVM back-end dynamically], enable_llvm=$enableval && enable_loadedllvm=$enableval, enable_loadedllvm=no)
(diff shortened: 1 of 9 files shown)


vargaz:
As I mentioned, this will make development more painful. Also, on ios, we have 2 cross compilers, both with a libmonollvm.dylib, which are going to have a name collision, or they have to be lipo-d together.

jaykrell:
I have only used the non-loaded form, but only a little.

configure.ac
```diff
@@ -3219,9 +3219,8 @@ dnl ***  LLVM  ***
 dnl **************
 
 AC_ARG_ENABLE(llvm,[  --enable-llvm	Enable the LLVM back-end], enable_llvm=$enableval, enable_llvm=default)
-AC_ARG_ENABLE(loadedllvm,[  --enable-loadedllvm	Load the LLVM back-end dynamically], enable_llvm=$enableval && enable_loadedllvm=$enableval, enable_loadedllvm=no)
```
akoeplinger: autoconf will just issue a warning if someone passes this now unknown option right?

akoeplinger: We still have --enable-loadedllvm in packaging/MacSDK/mono.py and README.md

luhenry:
Too many drawbacks for little removed code. We will revisit whenever we have a good story of use of in-tree builds of our libraries.

PR #8298, opened 2018-04-18
[sdks] Build LLVM configurations as part of Android CI

luhenry:


luhenry:
@monojenkins build Linux ARMv7 Bitcode

PR #8302, opened 2018-04-18
Fix mono_perfcounter_instance_exists to match prototype when #ifndef …

jaykrell:
…DISABLE_PERFCOUNTERS.


ntherning:
@jaykrell @luhenry Could we have this backported to 2018-02 and 2018-04?

jaykrell:
I not sure if it requires higher authority or if we are all allowed. Of course this is a trivial change..

lewurm:
@monojenkins backport 2018-02

PR #8306, opened 2018-04-18
[sdks] Wrap configure to provide better output.

rolfbjarne:
Wrap configure in a script that:

* Captures stdout/stderr and only shows it in case of errors.
* If configure fails, and a cache file was used, deletes the cache file and
  tries again.
* If configure really fails, then print config.log in addition to
  stdout/stderr, since a lot of valuable diagnostic output is in that file
  (and it's annoying to track down configure failures on bots when the file
  with the important output has already been deleted).

### DIFF
--- a/sdks/builds/runtime.mk
+++ b/sdks/builds/runtime.mk
@@ -50,7 +50,7 @@ __$(1)_CONFIGURE_ENVIRONMENT = \
 
 .stamp-$(1)-$$(CONFIGURATION)-configure: $$(TOP)/configure .stamp-$(1)-toolchain
 	mkdir -p $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)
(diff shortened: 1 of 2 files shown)


vargaz:
Would it be possible to add an option to turn off the dumping of config.log ? Its very annoying when running locally.

rolfbjarne:
@vargaz how would you prefer to set that option?

vargaz:
I think the best would be to make it only run on ci by only doing it if some ci specific env variable is set, like CI_TAGS on jenkins.

jaykrell:
I would like autogen to use this, or maybe build it into configure.
Well, what I really want, is for all make output to always be output to a file.
With accurate command (V=1, and remove a lot of '@') and clear indication of where errors are and what directory and command caused them....

rolfbjarne:
@vargaz fixed to check for JENKINS_HOME or BUILD_REVISION before printing config.log

luhenry:
@vargaz monojenkins backport to 2018-02

luhenry:

luhenry:

PR #8309, opened 2018-04-18
ifdef reduction

jaykrell:
Typo: is_adressable_valuetype_load => is_addressable_valuetype_load
Don't mark large multi-called function inline.
some formatting cleanup.

Viewed in github, there is a formatting trick being removed that doesn't always work, so I didn't even see it. Specifically lines are started with 10 tabs to effectively give a newline and start at the start of the line.

Seen doing other work..

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -453,7 +453,7 @@ gsharedvt_failure (MonoCompile *cfg, int opcode, const char *file, int line)
 		(tblock) = cfg->cil_offset_to_bb [(ip) - cfg->cil_start]; \
 		if (!(tblock)) {	\
 			if ((ip) >= end || (ip) < header->code) UNVERIFIED; \
-            NEW_BBLOCK (cfg, (tblock)); \
+			NEW_BBLOCK (cfg, (tblock)); \
 			(tblock)->cil_code = (ip);	\
 			ADD_BBLOCK (cfg, (tblock));	\
 		} \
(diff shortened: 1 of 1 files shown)


mono/mini/method-to-ir.c
```diff
@@ -3172,7 +3169,7 @@ emit_get_rgctx_sig (MonoCompile *cfg, int context_used,
 
 static MonoInst*
 emit_get_rgctx_gsharedvt_call (MonoCompile *cfg, int context_used,
-							   MonoMethodSignature *sig, MonoMethod *cmethod, MonoRgctxInfoType rgctx_type)
+			       MonoMethodSignature *sig, MonoMethod *cmethod, MonoRgctxInfoType rgctx_type)
 {
```
vargaz: This lines up fine with tab size==4 why does it need to be changed ?

jaykrell: I thought we use tab=8, so I have been. This is the case on a few lines then probably.

jaykrell: http://www.mono-project.com/community/contributing/coding-guidelines/
says 8

I would prefer 4 myself. If most everything looks good that way I can switch.
4 is the default in some editors, 8 in others -- which is why spaces are good.
Comments at the top aren't understood by most editors, so that isn't a good choice.

alexrp: Why are we even using tabs here? Spaces should be used for alignment.

vargaz: That line has tabs extended by spaces to match up the two lines.

alexrp: I mean we should use tabs up to the indentation level of the previous line, and then spaces for alignment. Since the previous line had no indentation, there shouldn't be any tabs. That way it doesn't matter how someone configures their editor; indentation and alignment will be correct regardless.

jaykrell: I was not aware of this convention, but anything involving spaces instead of tabs I am for.

jaykrell: If I remove the ifdef, and remove whitespace at ends of lines, and fix the spelling of the identifer, and remove inline on large but rarely called function, and leave the line wrapping or very large indentation alone, ok?

vargaz: Y.

jaykrell: ok will do later thank you

luhenry:
@monojenkins build

Issue #7333, opened 2018-02-28
Improve build of LLVM-based CI configurations


akoeplinger:
We already have some CI setup for testing AOT+LLVM, hybridAOT+LLVM and fullAOT+LLVM, but  all these configurations are broken because of missing pre-built llvm packages.

1. Automate the build of these llvm pre-built binaries
  
2. Use caching of these pre-built binaries so if a package for a specific version of llvm hasn’t been built yet, we do not fail the build and we just fallback to the slower option of building llvm

vargaz:
https://jenkins.mono-project.com/view/All/job/build-package-osx-llvm/

This lane builds llvm packages for ios, it could be extended to build the desktop ones as well.

marek-safar:
@luhenry has a plan to work on some caching infrastructure as part of SDK work we should coordinate the work

kumpera:
Can we have regular OSX executed with —llvm for all steps?

jaykrell:
That matrix is missing a column also -- the LLVM/JIT scenario.

luhenry:
`FullAOT+LLVM` and `JIT+LLVM` are available on PRs with `@monojenkins build Linux x64 FullAOT+LLVM` and `@monojenkins build Linux x64 JIT+LLVM` respectively. The former is enabled on all PR, and the latter is available on-demand. They both have some breakages, but the later doesn't even pass the build stage.

Once we have a solid `FullAOT+LLVM`, I will add `{,hybrid}AOT{,+LLVM}` to all PRs with the same steps: if working then enabled on all PR, otherwise on demand until fixed.

PR #8317, opened 2018-04-18
[ci] Add `jit_llvm` configuration

luhenry:
https://github.com/mono/mono/issues/7333


vargaz:
This will compile and run everything with llvm, which is probably going to take a lot of time and have a lot of failures.

jaykrell:
Is that a bad thing? Should we run some hybrid, like, where we run csc w/o LLVM, i.e. to focus on the unit tests and not the entire system? Maybe have two forms -- "everything" that runs far less often and "hybrid/test" that runs more often? I noticed what you say -- I accidentally used LLVM+JIT a few week ago and build/csc was slow.

luhenry:
@monojenkins build Linux x64 JIT+LLVM

Issue #8311, opened 2018-04-18
bug-82022.exe fails in LLVM JIT


akoeplinger:
## Steps to Reproduce

1. Run bug-82022.exe with LLVM JIT: `make -C mono/tests -k test-wrench MONO_ENV_OPTIONS=--llvm V=1 CI=1 M=1`
2. It fails

It seems to have started with https://github.com/mono/mono/pull/8141.

## Current Behavior

Fails reproducibly on Jenkins: https://jenkins.mono-project.com/job/test-mono-mainline-mac-sdk/label=osx-amd64/1274:

```
bug-82022.exe
=============== bug-82022.exe.stdout ===============

=============== EOF ===============
=============== bug-82022.exe.stderr ===============

=============== EOF ===============
```

## Expected Behavior

No failure

### On which platforms did you notice this

[ x ] macOS
[ ] Linux
[ ] Windows

**Version Used**: master

/cc @vargaz

PR #8318, opened 2018-04-18
[runtime] Handle failed classes in mono_type_get_checked ().

vargaz:
Fixes https://github.com/mono/mono/issues/8311.

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -2603,8 +2603,12 @@ mono_type_get_checked (MonoImage *image, guint32 type_token, MonoGenericContext 
 	if ((type_token & 0xff000000) != MONO_TOKEN_TYPE_SPEC) {
 		MonoClass *klass = mono_class_get_checked (image, type_token, error);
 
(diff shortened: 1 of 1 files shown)


mono/metadata/class.c
```diff
@@ -2603,11 +2603,8 @@ mono_type_get_checked (MonoImage *image, guint32 type_token, MonoGenericContext
 	if ((type_token & 0xff000000) != MONO_TOKEN_TYPE_SPEC) {
 		MonoClass *klass = mono_class_get_checked (image, type_token, error);
 
-		if (!klass) {
+		if (!klass || m_class_has_failure (klass))
```
kumpera: I assume this is needed because mono_class_get_checked won't return NULL for broken classes.

This change breaks the invariant that mono_type_get_checked will set an error if it returns NULL, we should avoid this as much as possible and explicit document those cases when unavoidable.

marek-safar:
@monojenkins backport 2018-04

Issue #8319, opened 2018-04-18
Failing tests on FullAOT+LLVM


luhenry:
We have the following tests failing consistently on the FullAOT+LLVM PR job:

 - `MonoTests.runtime.generic-stack-traces2.2.exe`
 - `MonoTests.runtime.finally_block_ending_in_dead_bb.exe`
 - `FileTest.Create_Path_ReadOnly`
 - `FileTest.GetAttributes_ReadOnly`

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot+llvm/46/testReport/

akoeplinger:
Let's use https://github.com/mono/mono/issues/8321 for tracking since the runtime tests were actually recently broken.

PR #8320, opened 2018-04-18
[ci] Disable tests consistently failing on FullAOT+LLVM

luhenry:
See https://github.com/mono/mono/issues/8319

### DIFF
--- a/mcs/class/corlib/Test/System.IO/FileTest.cs
+++ b/mcs/class/corlib/Test/System.IO/FileTest.cs
@@ -147,6 +147,10 @@ namespace MonoTests.System.IO
 		}
 
 		[Test]
+#if FULL_AOT_RUNTIME
+		// This test doesn't work with FullAOT+LLVM
+		[Ignore ("https://github.com/mono/mono/issues/8319")]
+#endif
 		public void Create_Path_ReadOnly ()
 		{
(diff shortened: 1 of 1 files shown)


mono/tests/Makefile.am
```diff
@@ -1223,14 +1223,22 @@ CI_PR_DISABLED_TESTS = \
...
+LLVM_DISABLED_TESTS = \
+	finally_block_ending_in_dead_bb.exe \
+	generic-stack-traces2.2.exe
+
 # https://github.com/mono/mono/issues/6490
 LLVM_DISABLED_TESTS += \
 	weak-fields.exe
 
-LLVM = $(filter --llvm, $(MONO_ENV_OPTIONS))
+if ENABLE_LLVM
```
akoeplinger: does ENABLE_LLVM here mean that llvm is actually used or just that it is available?

luhenry: That it is available. Unfortunately we have no good way of checking if we have are running with llvm enabled.

luhenry: Since it doesn't seem to work, and the tests have only started to fail recently (https://jenkins.mono-project.com/job/test-mono-mainline-linux-fullaot+llvm/507/) I don't think we should disable them, and just have somebody look at what the regression was.

Issue #8149, opened 2018-04-10
Assembly.LoadFile and Assembly.LoadFrom have same behavior


LePtitDev:
https://github.com/mono/mono/blob/8b5d58285518bf509dcd21fd299114009cb1e908/mcs/class/corlib/System.Reflection/Assembly.cs#L578

I try to load 2 different assemblies but when I call Assembly.LoadFile twice, the second returned assembly is the first (locations are same).

https://blogs.msdn.microsoft.com/suzcook/2003/09/19/loadfile-vs-loadfrom/

"Use the LoadFile method to load and examine assemblies that have the same identity, but are located in different paths.LoadFile does not load files into the load-from context, and does not resolve dependencies using the load path, as the LoadFrom method does. LoadFile is useful in this limited scenario because LoadFrom cannot be used to load assemblies that have the same identities but different paths; it will load only the first such assembly."

lambdageek:
@LePtitDev to clarify: the issue here is that you wish to call `LoadFile` multiple times with the same path.  And in between you update the on-disk assembly?  Something like:

```csharp
   var a = Assembly.LoadFile ("dir1/Lib.dll");
   File.Copy ("dir2/Lib.dll", "dir1/Lib.dll");
   var b = Assembly.LoadFile ("dir1/Lib.dll");
   // error is here - no new library loaded - b is the same as a
```

lambdageek:
This is happening for the same reason as https://bugzilla.xamarin.com/show_bug.cgi?id=11199 and has to do with how Mono caches loaded assemblies.

jairbubbles:
No the issue is more:
```csharp
var a = Assembly.LoadFile ("CompilationCache/Dir1/Lib.dll");
var b = Assembly.LoadFile ("CompilationCache/Dir2/Lib.dll");
// we expect a != b
```
We rely on this CLR implementation which is well documented by Microsoft:
> Use the LoadFile method to load and examine assemblies that **have the same identity, but are located in different paths**

Our scenario is I guess not that commo: we use Roslyn compiler to compile C# code on the fly. We do not want to use app domains so we load new assemblies each time we recompile.

LePtitDev:
The problem is that I use Roslyn to compile assemblies but if I write this code :
```csharp
var content = @"
use System;
class MyProgram
{
    public static void Main(string[] args)
    {
        Console.WriteLine(""test1"");
    }
}";
compiler.CompileAndSave(content, "dir1/assembly.dll");
[...]

Issue #8152, opened 2018-04-10
Assembly.LoadFrom behaves differently in Mono as with Microsoft's CLR WRT binding redirect


louis-z:
## Steps to Reproduce
[BindingRedirectTest.zip](https://github.com/mono/mono/files/1895031/BindingRedirectTest.zip)

1. Using the attached ZIP file, build the provided solution in Debug with VS2017.
2. Successively run the resulting executable with Microsoft's CLR and with Mono.
3. Notice differences in the console output.

_Note: The executable will explicitly invoke nuget.exe. Make sure it can be found in the PATH._

## Current Behavior

The executable has a .config file that should redirect to version 1.2.2.0 any request for a System.Collections.Immutable version smaller than 1.2.2.0:

```xml
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.Collections.Immutable" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-1.2.2.0" newVersion="1.2.2.0" />
      </dependentAssembly>
    </assemblyBinding>
```

Mono does not seem to take this into account when `Assembly.LoadFrom` is called.

## Expected Behavior

The _right_ behavior should be that of the Microsoft CLR's. In other words, Mono should not load version 1.2.1.0 of the System.Collections.Immutable assembly but simply redirect towards version 1.2.2.0.

### On which platforms did you notice this

[ ] macOS
[ ] Linux
[x] Windows

**Version Used**:

Mono JIT compiler version 5.10.0 (Visual Studio built mono)

lambdageek:
Related #8149: Mono's `LoadFrom` really has the semantics of `LoadFile` - just load the assembly from the given path without probing.  we should keep the existing mechanism as `LoadFile` and add a proper `LoadFrom`.

jairbubbles:
@lambdageek Well in #8149 the issue is that `LoadFile` do return an already loaded assembly from a different location while it shouldn't. So I would say neither of them do behave like CLR!

lambdageek:
@jairbubbles You're right. I didn't read the other one closely enough.

Our current `LoadFrom` is closer to what `LoadFile` is supposed to do, but still wrong.

lambdageek:
@jairbubbles @louis-z Mono's `LoadFrom` doesn't know about binding redirects at all.

----

`Load(String)` understands binding redirects.

[...]

PR #8326, opened 2018-04-19
[loader] Add loader contexts

lambdageek:
There are 4 different ways that an assembly can be loaded in .NET:
1. default context - `Assembly.Load(string)` does this.
2. reflection only context - `Assembly.ReflectionOnlyLoad` and `Assembly.ReflectionOnlyLoadFrom` do this.
2. load from context - `Assembly.LoadFrom(string)` does this. The interesting bit is that the assembly at the specified path is probed for its assembly name, and then binding redirects are applied to figure out if we need to load a different version of the assembly (from elsewhere in the search path, presumably).
3. "no context" (.NET Core's [AssemblyLoadContext](https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/assemblyloadcontext.md) design doc calls it "custom context", and internally they implement it as `IndividualAssemblyLoadContext`) - in Mono I'm going to call it the "individual context".  This is what `Assembly.LoadFile(string)` and `Assembly.Load(byte[])` do: you get exactly the thing you asked for even if something with the same assembly name is already loaded and you get to have problems with type resolution.

So this PR adds:
1. `MonoAssemblyLoadContextKind` enum - default, refonly, loadfrom and individual.
2. Store the `MonoAssemblyLoadContext` in every `MonoAssembly` instead of a boolean `refonly` flag.
3. When loading references from assemblies, consider the load context of the requesting assembly.
4. Do the wacky `LoadFrom` probe and rebind dance.
5. Don't check for duplicated for the individual assembly context.

Not in this PR:

The managed assembly load context stuff in .NET Core.  If we eventually want to add it, it will likely build on this work, but it's not here now.

Fixes: #8149, #8152, [bugzilla #11199](https://bugzilla.xamarin.com/show_bug.cgi?id=11199), [bugzilla #33728](https://bugzilla.xamarin.com/show_bug.cgi?id=33728)

TODO:
- [x] Add tests.  (I tried the repros from all the bugs above, but didn't turn them into tests yet).
- [x] Check that `ReflectionOnlyLoadFrom` is correct and binding redirects do not apply to it.
- [x] ~Find out what happens in this situation: directory `d` contains `Asm1.dll` version 1 and `Asm2.dll` version 1. App has a binding redirect for `Asm1` all versions to version 2.  normal app probing path has `Asm1` version 2 and `Asm2` version 2. `LoadFrom(d/Asm1.dll)` and then cause the `Asm2` reference to be loaded - do we get `Asm2` version 1 or version 2.  (ie did we retain `d` as a probing path even though the redirect made us look for `Asm1` somewhere else)~ it's fine... there's nothing that remembers that we started with a `LoadFrom` in `d/`. 
- [x] FIXME: resolving references from individual context assemblies should only fire the `AssemblyResolve` event and not look in the default context.
- [x] FIXME: SRE assemblies should be in the individual context.
- [x] FIXME: "assemblies that are loaded from byte arrays are loaded without context unless their identity (after policy is applied) establishes that they are in the global assembly cache." - I'm not applying policy or looking in the GAC...
- [x] Figure out the FIXMEs by trying .NET Framework
  - [x] ~If `LoadFrom` is subject to a binding redirect, where should `LoadFrom` look for the replacement assembly. Is the given filename's basename relevant at all? Is the new assembly loaded in the default context?~ No
  - [x] Should any sort of deduplication happen for `LoadFile` and `Load(byte[])` - should you really get brand new `MonoAssembly*` pointers every time (even if for `LoadFile` the `MonoImage` is shared).
  - [x] Am I firing all the right managed events? As best I can tell, yes.

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -1966,7 +1966,7 @@ load_aot_module (MonoAssembly *assembly, gpointer user_data)
 		 */
 		return;
 
-	if (image_is_dynamic (assembly->image) || assembly->ref_only || mono_domain_get () != mono_get_root_domain ())
+	if (image_is_dynamic (assembly->image) || mono_asmctx_get_kind (&assembly->context) == MONO_ASMCTX_REFONLY || mono_domain_get () != mono_get_root_domain ())
 		return;
 
 	mono_aot_lock ();
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -1976,10 +1976,10 @@ if (container_assm_name && !container_amodule) {
 	container_assm_name = NULL;
 	MonoImageOpenStatus status = MONO_IMAGE_OK;
 	gchar *dll = g_strdup_printf (		"%s.dll", local_ref);
-	MonoAssembly *assm = mono_assembly_open_a_lot (dll, &status, FALSE, FALSE);
+	MonoAssembly *assm = mono_assembly_open_a_lot (dll, &status, MONO_ASMCTX_DEFAULT);
 	if (!assm) {
 		gchar *exe = g_strdup_printf ("%s.exe", local_ref);
-		assm = mono_assembly_open_a_lot (exe, &status, FALSE, FALSE);
+		assm = mono_assembly_open_a_lot (exe, &status, MONO_ASMCTX_DEFAULT);
 	}
 	g_assert (assm);
 	load_aot_module (assm, NULL);
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -272,6 +272,8 @@ load_image (MonoAotModule *amodule, int index, MonoError *error)
 
 	error_init (error);
 
+	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_AOT, "AOT: module %s wants to load image %d: %s", amodule->aot_name, index, amodule->image_names[index].name);
+
 	if (amodule->image_table [index])
 		return amodule->image_table [index];
 	if (amodule->out_of_date) {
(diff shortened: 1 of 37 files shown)


luhenry:

akoeplinger:
> The managed assembly load context stuff in .NET Core. If we eventually want to add it, it will likely build on this work, but it's not here now.

FWIW I've seen e.g. Roslyn use reflection to probe for this type to check whether they're on .NET Core so we might need to coordinate this when we add it.

lambdageek:
@kumpera It's definitely the case here that the code isn't making a distinction between `ReflectionOnlyLoad`and `ReflectionOnlyLoadFrom` - I'll rework the PR to add that.

@akoeplinger I'll make sure to coordinate if I start making changes to the managed interfaces.

kumpera:
@lambdageek on refonly, we should validate my claim first. I'm not 100% sure on it.

lambdageek:
Interesting. `ReflectionOnlyLoadFrom` in .NET Framework does _not_ have binding redirects applied. It  loads straight from the given path.  So what I have now is correct.

lambdageek:
Changing `LoadFile` to use `MONO_ASMCTX_INDIVIDUAL` broke xUnit for some reason (changing it to `MONO_ASMCTX_DEFAULT` or even `LOADFROM` makes it work).  I can see that they use `LoadFile` on .NET Framework sometimes.  Investigating...

lambdageek:
The reason xUnit broke is because apparently even for a `LoadFile`/"no context"/indivudal assembly context, the .NET Framework will still cause GAC assemblies (or at least framework Facade assemblies like `System.Runtime`) to be loaded (and won't fire an `AssemblyLoad` event to announce it, for some reason)).

That's... going to require a slight hack.  Mono doesn't really have a way of saying "Load from the GAC (and from MONO_PATH) but not from the current domain's paths").  So I'll have to add that.

luhenry:
Looks like `Linux x64 FullAOT` is missing a AOTted assembly : https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot/10068/testReport/junit/MonoTests/runtime/assembly_loadfile_exe/

Do you think the `Linux AArch64 Interpreter` failure is related? https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-aarch64/4408/parsed_console/log_content.html#WARNING1. I'll retry it to check if it's a flake.

luhenry:

luhenry:

mono/mini/aot-runtime.c
```diff
@@ -2326,7 +2347,10 @@ if (container_assm_name && !container_amodule) {
 		for (i = 0; i < amodule->image_table_len; ++i) {
 			ERROR_DECL (error);
 			load_image (amodule, i, error);
-			mono_error_cleanup (error); /* FIXME don't swallow the error */
+			if (!is_ok (error)) {
+				mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_AOT, "AOT: image '%s' could not load image %d of %d due to: %s", found_aot_name, i, amodule->image_table_len, mono_error_get_message (error));
+				mono_error_cleanup (error); /* FIXME don't swallow the error */
```
vargaz: We already do a mono_trace from load_image ().

luhenry:

luhenry:

luhenry:

luhenry:
@monojenkins build Linux x64

lambdageek:
@monojenkins build Linux i386

marek-safar:
@mhutch no, we didn't test that with MD yet. We do that regularly during monthly integration work but if you have a specific concern we could do that earlier

lambdageek:
@mhutch interesting.

I'll try VSfM on top of mono master today.  What's something specific I should try that would demonstrate the problem?  Open VS Mac and try to install some extensions?

KirillOsenkov:
@lambdageek: @davkean and @marklio are my go to people for assembly loading, fusion, loader contexts, etc. in .NET Framework. 

Guys, who would be the best person to give this a sanity check? I know assembly loading is very tricky and this has potential to break stuff since I know Mono used to be much more lenient and I know people took dependency on the old behavior.

PR #8327, opened 2018-04-19
[Tailcall] Remove OP_JMP support. Require OP_TAILCALL.

jaykrell:
Per @vargaz and what I was considering myself.
This only hurts mips and sparc -- it breaks them.
Remove front end support for OP_JMP and conditionalality of OP_TAILCALL.

In place of `tailcall` (backend) and `supported_tailcall` (portable), now just `tailcall`.
Though perhaps that is too short and `supported_tailcall` would be preferred, perhaps.

We might be able to keep mips/sparc working by using OP_CALL instead.


jaykrell:
mips/sparc C# will probably still work, but F# will hit assertion failures due to lack of switch/case op_tailcall.
A tiny restoration could keep that hobbling along -- retain advertisement of the op_tailcall and when not present, ignore tail. prefix. If anyone complains we can do that.

vargaz:
Neither sparc nor mips is supported, and both are probably broken, this is just going to break them a bit more.

vargaz:
MONO_ARCH_HAVE_OP_TAILCALL should also be removed.

jaykrell:
And I thought I was thorough, oops, agreed.

alexrp:
Pretty sure Debian still builds Mono on MIPS? cc @directhex

jaykrell:
Do they include F#? C# will probably work.
The way to keep this working is pretty easy.
Provide mono_arch_tailcall_supported that always returns false.

vargaz:
We don't build on mips which means there is a  90% change its broken.

lewurm:

directhex:
Debian has removed SPARC.

MIPS packages are provided by Debian, but we don't CI them in Xamarin, which as @vargaz points out means it's likely perma-broken. Miguel indicated that there wasn't sufficient justification for MIPS work due to removal of MIPS from Android SDK, so I guess breaking MIPS is fine.

jaykrell:
"Shrink or eliminate small local arrays." went to wrong branch.

jaykrell:
OpenBSD and probably NetBSD and maybe FreeBSD support MIPS and SPARC. I would like to see them continue to work, but I don't have much "community time" to offer these days.

But I'd prefer a pure C++ + autoconf option, along the lines of interpreter/il2cpp/llvmonly/wasm, pushing the backend off to other projects..

barracuda156:
@jaykrell For `ppc` this change is okay, right? Just to make sure.

PR #8330, opened 2018-04-19
[aot] Avoid making the plt labels referenced by the llvm generated code temporary, since they need to be accessed from another object file.

vargaz:


mono/mini/aot-compiler.c
```diff
@@ -3609,6 +3609,12 @@ get_plt_entry (MonoAotCompile *acfg, MonoJumpInfo *patch_info)
 			res->llvm_symbol = g_strdup_printf ("%s_%s_llvm", res->symbol, res->debug_sym);
 		else
 			res->llvm_symbol = g_strdup_printf ("%s_llvm", res->symbol);
+		if (strstr (res->llvm_symbol, acfg->temp_prefix) == res->llvm_symbol) {
+			/* The llvm symbol shouldn't be temporary, since the llvm generated object file references it */
+			char *tmp = res->llvm_symbol;
```
luhenry: The `tmp` variable isn't used. Did you mean to have `g_strdup (tmp`?

Issue #8273, opened 2018-04-17
FullAOT takes 9 minutes vs. AOT or JIT under a second, on this F#-derived code


jaykrell:
This takes under a second under regular AOT.
JIT is also under a second.
But over 9 minutes with FullAOT.
It doesn't work in master today but should soon.

If you change string to int32 it runs much faster.
And you can reduce the iteration count to make it tolerable (look for 999).

Why does it go into trampolines?

Command lines are like:

`ilasm 3.il` and then, based on mono/tests:

```
MONO_PATH=/s/mono/mcs/class/lib/testing_aot_full /s/mono/mono/mini/mono -O=gsharedvt --aot=full,nimt-trampolines=2000,ntrampolines=10000,nrgctx-fetch-trampolines=256,ngsharedvt-trampolines=4000 tailcall/3.exe
```

```
MONO_PATH=/s/mono/mcs/class/lib/testing_aot_full /s/mono/mono/mini/mono  tailcall/3.exe
```
vs.
```
MONO_PATH=/s/mono/mcs/class/lib/testing_aot_full /s/mono/mono/mini/mono --full-aot tailcall/3.exe
```

Based on this F# and C#.
```fsharp
type TailCallLoopGenericClassAndMethod<'T1>(resultA: 'T1) =
    member this.Method1<'T2>(x:int, resultB: 'T2) =
        if x = 0 then (resultA, resultB) else this.Method2 (x - 1, resultB)
    member this.Method2<'T2>(x:int, resultB: 'T2) =
        if x = 0 then (resultA, resultB) else this.Method1 (x  - 1, resultB)
```

```csharp
using System;
using System.Runtime.CompilerServices;
using static System.Runtime.CompilerServices.MethodImplOptions;

unsafe class A<T1>
{
	T1 resultA;

	[MethodImpl (NoInlining)]
	static void check (long stack1, long stack2)
	{
// NOTE: This is wierd in order to feed into the edited IL.
		if (stack1 != 0)
			return;
		if (stack1 == stack2)
			return;
		Console.WriteLine ("tailcall failure {0} {1}", stack1, stack2);
		Environment.Exit (1);
	}

	[MethodImpl (NoInlining)]
	public A(T1 a)
	{
		resultA = a;
	}

	[MethodImpl (NoInlining)]
	public System.Tuple<T1, T2> Method1<T2> (T2 resultB, long depth = 100, long stack = 0)
	{
		int local;
		if (depth > 0)
[...]

PR #8333, opened 2018-04-19
[Tailcall] Add some tests derived from F# code to be used soon.

jaykrell:
Includes case for https://github.com/mono/mono/issues/8273 where
FullAOT is hundreds of time slower than AOT or JIT (under a second vs. 9 minutes).


jaykrell:
@vargaz Not sure you were intending I split these off from the other PR or not.

PR #8342, opened 2018-04-19
[Mono.Profiler.Log] Support cancellation in LogReader

rodrmoya:
We recently found a bug in the Xamarin.Profiler that, when profiling an
app for a very short time, we've come to the need to distinguish between
reading and processing cancellation, so this adds support to LogReader
for the reading part.

### DIFF
--- a/mcs/class/Mono.Profiler.Log/Mono.Profiler.Log/LogStream.cs
+++ b/mcs/class/Mono.Profiler.Log/Mono.Profiler.Log/LogStream.cs
@@ -4,14 +4,17 @@
 
 using System;
 using System.IO;
+using System.Threading;
 
 namespace Mono.Profiler.Log {
 
 	public class LogStream : Stream {
 
+		protected CancellationToken Token { get; }
+
 		public Stream BaseStream { get; }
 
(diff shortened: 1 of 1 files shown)


alexrp:
1. It's not immediately obvious to me why you need to distinguish between these two types of cancellations. What's the use case here?
2. Even if there is a compelling use case, why not just have a `LogStream`-derived class in the Xamarin Profiler where you can do the cancellation check in your `EndOfStream` override? It doesn't seem like this *needs* to live in the upstream library per se.

rodrmoya:
For #1 the use case is that we read the file in one thread (read it and provide an immediate visitor and a sorted visitor, which just add events to a queue) and process it in another, so just using a single cancellation token for `LogProcessor.Process` makes us cancel the processing, and the bug we found is that on very short runs, because we were cancelling the *processing*, our sorted visitor was never getting any event. So, with this separation, we can just cancel reading but still leave the processor thread alive until all read-till-now events can be processed.

As for #2, we already do that, for live profiling, so this just adds the same "API" to the base LogStream. We can indeed have a subclass for LogStream indeed, just thought it would make sense in the MPL API, as there is already the processing cancellation (in `LogProcessor.Process`).

Issue #8329, opened 2018-04-19
mono --interpreter got * Assertion at ..\mono\mini\interp\interp.c:3557, condition `csig->call_convention == MONO_CALL_DEFAULT' not met


yindongfei:
## Steps to Reproduce

[client.zip](https://github.com/mono/mono/files/1926610/client.zip)

## Current Behavior

after build the project attached, I use mono client.exe, the right result is produced. but when I use mono --interpreter client.exe, I got * Assertion at ..\mono\mini\interp\interp.c:3557, condition `csig->call_convention == MONO_CALL_DEFAULT' not met

## Expected Behavior

same as the result with out --interpreter.

### On which platforms did you notice this

[ ] Windows 10

**Version Used**:
c:\Program Files\Mono\bin>mono --version
Mono JIT compiler version 5.10.1 (Visual Studio built mono)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           normal
        SIGSEGV:       normal
        Notification:  Thread + polling
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          supported, not enabled.
        GC:            sgen (concurrent by default)

lewurm:
Thanks for your bug report. We actually never had the interpreter running on Windows. Luckily it was a trivial change to get it working there 🙂 

Did you compile mono yourself or did you obtain it somewhere?

yindongfei:
I got mono from official download link. I can build it from source code and try again.

luhenry:
@lewurm should we have a `Windows x64 Interpreter` CI lane?

lewurm:
@yindongfei you can actually get a master build here: https://xamjenkinsartifact.azureedge.net/build-package-win-mono/master/992/8127483e8a6b76f2febfdd72968db34f004990e2/mono-5.15.0.228-x64-0.msi

lewurm:
@luhenry I don't think we need one

PR #8343, opened 2018-04-19
[interp] remove useless assert in mint_newobj

lewurm:
Fixes https://github.com/mono/mono/issues/8329. That is, make the interpreter work on Windows.

No idea what the original indent of this assert was, it came with the import of the old interpreter code base.


lewurm:
with this PR we pass the mini regression test suite on windows

lewurm:

lewurm:
@monojenkins rebase

PR #8347, opened 2018-04-19
[llvm] Remove the 'llvm' directory, llvm can be compiled using the sdk makefiles now.

vargaz:


llvm/Makefile.am
```diff
@@ -1,30 +0,0 @@
-#
```
akoeplinger: configure.ac has another mention of this in AC_OUTPUT that needs to be removed

luhenry:
Following the discussion on making LLVM more easily compilable for anybody who wants to use Mono, we should probably have it the other way around: have the `sdks/` and `packaging/MacSDK/` depends on `llvm/`. The logic to compile LLVM would then be moved to `llvm/` and `configure.ac` would be modified to make it possible to compile LLVM if none are already available (via `--with-llvm` or installed on the system).

vargaz:
We already have the sdk code, it works, while the code in llvm/ never really worked, and it was never used.

PR #8361, opened 2018-04-20
Reduce #ifdef MONO_ARCH_SOFT_FLOAT_FALLBACK, and #if SIZEOF_VOID_P == 4.

jaykrell:
The optimizer should be able to do the same.

This reduces ARM32-specific breaks, as more code compiles
for all architectures (but is optimized away).

### DIFF
--- a/mono/mini/alias-analysis.c
+++ b/mono/mini/alias-analysis.c
@@ -195,6 +195,10 @@ handle_instruction:
 				}
 				break;
 
+			case OP_LOADR4_MEMBASE:
+				if (MONO_ARCH_SOFT_FLOAT_FALLBACK)
+					break;
+				// fallthrough
 			case OP_LOADV_MEMBASE:
 			case OP_LOAD_MEMBASE:
 			case OP_LOADU1_MEMBASE:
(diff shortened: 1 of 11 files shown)


vargaz:
It was kept ifdef-ed because its a large body of code, and we only want to include it if needed.

jaykrell:
Optimizer should handle it the same. All the runtime ifs will evaluate to constant false when the ifdef was false.

jaykrell:
I don't see how. Let me double check.
Taking the address of a function under if (0) should not keep the function, for example.

jaykrell:
Huh, apparently Unix linkers don't throw out unused functions, unless they are also static.
Most of these functions in jit-icalls.c are only referenced from mini-runtime.c so I moved them there and made them static and then they are gone. As well, though I didn't check, they should have been comprised of "only" unreached() (which could be bloaty with assertion failure, file, line).

vargaz:
Those functions are jit icalls, so logically they should be in jit-icalls.c, is it really worth doing all this work just to get rid of one ifdef ?

jaykrell:
The overall symbol or the specific instance(s)? I can put back one or two instances if desired, i.e. around the functions and their registration, and still achieve similar. The "work" is easy and quick, but the churn and desired structure of the code matter, granted.

static does have side benefits that weren't the original point.

jaykrell:
Hm. I like the static. Can the registration be moved to jit-icalls instead of mini-runtime.c? Let me see.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

luhenry:
This doesn’t have any positive impact, and it seems to be mostly about fitting personal preferences.

Issue #8007, opened 2018-04-04
Implement support for pinvoke for wasm target.


yowl:
## Steps to Reproduce

1. See https://github.com/mono/mono/blob/63795c0ce8e564344a791c6406a6862525b7309f/mono/utils/mono-dl-wasm.c#L46.  
2. Its not implemented so calls to another SO using Pinvoke fail.  This would be useful to add Skia support.
3. 

### On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows
[*] Wasm

**Version Used**:

master

### Stacktrace

```
Please paste the stack trace here if available.
```

kumpera:
Dynamic linking on WebAssembly is very tricky, it requires to compile the runtime with such feature enabled and it comes at a performance & footprint cost. 

I'm not entirely sure on how to proceed with this, but I agree that this is a feature we need to support.

kekekeks:
@kumpera You are [providing a way](https://github.com/mono/mono/tree/master/sdks/wasm) to build a custom `.wasm` file by linking libmonosgen-2.0.a and using a custom driver.c. 

That allows people to statically link additional modules which means that functions from them will be callable using `call_indirect` without additional performance costs. So at least `[DllImport("__Internal")]` should be supported, which would be sufficient for projects like Avalonia that will probably need to ship a custom runtime anyway.

kekekeks:
P/Invoke would also be very nice for calling external JS functions "directly" (yes, I understand that JS side will need manual marshaling) instead of using `eval`.

kumpera:
@kekekeks yeah, the problem with __Internal is that there's no way to locate those symbols in the wasm file as they don't have names.

On using pinvoke to directly call JS, that's even more complicated as the marshaling on the C# side is non trivial.

I explored this problem in the past and the linkage model of WebAssembly is quite unfriendly to dotnet's pinvoke.

I expect this to be solved by a mix of runtime code plus tooling to generate the stubs needed to locate those functions.

kumpera:
@kekekeks makes sense on adding the extra call to register functions, I'll look into it.

When I say i's non trivial to marshal from the C# side is that pinvoking into JS requires currying an extra argument as the only way to call into JS is through the imports table.

Now, on linking shared modules, it requires a lot more than just passing the function pointers around, we need to deal with linear memory linking or you won't be able to pass pointers around.

TheSpydog:
[...]

PR #8363, opened 2018-04-20
Added P/Invoke support to wasm backend

kekekeks:
Since it's not quite viable to use dynamic linking provided by WebAssembly itself, one has to statically link the native library to resulting `mono.wasm`. However emsdk has no idea which functions are used and linker removes them. It's also not possible to get a function pointer by its name.

This PR enables accessing such libraries via P/Invoke by manually registering needed imports from a custom driver.c file

#8007
https://github.com/AvaloniaUI/Avalonia/issues/1387

@kumpera

### DIFF
--- a/mono/utils/mono-dl-wasm.c
+++ b/mono/utils/mono-dl-wasm.c
@@ -12,6 +12,63 @@
 #include <string.h>
 #include <glib.h>
 
+typedef struct MonoDlWasmFunctionEntry
+{
+	char *name;
+	void *ptr;
+	struct MonoDlWasmFunctionEntry *next;
+
+} MonoDlWasmFunctionEntry;
+
+typedef struct MonoDlWasmLibraryEntry
+{
+	char *name;
+	MonoDlWasmFunctionEntry *functions;
+	struct MonoDlWasmLibraryEntry *next;
+
+} MonoDlWasmLibraryEntry;
+
+
+static MonoDlWasmLibraryEntry* mono_dl_wasm_entries = NULL;
+
+
+static MonoDlWasmLibraryEntry *
+mono_dl_wasm_get_library_entry (const char* name, int create)
+{
+	if (name == NULL)
+		name = "__Internal";
+
+	MonoDlWasmLibraryEntry *entry = mono_dl_wasm_entries;
+	while (entry)
+	{
+		if (0 == strcmp (name, entry->name))
+			return entry;
+		entry = entry->next;
+	}
+
+	if (!create)
+		return NULL;
+
+	entry = (MonoDlWasmLibraryEntry *)g_malloc0 (sizeof (MonoDlWasmFunctionEntry));
+	entry->next = mono_dl_wasm_entries;
+	entry->name = g_strdup (name);
+	entry->functions = NULL;
+	mono_dl_wasm_entries = entry;
+	return entry;
+}
+
+void mono_dl_wasm_add_internal_pinvoke (const char *library_name, const char *name, void *ptr)
+{
+	struct MonoDlWasmLibraryEntry *library = mono_dl_wasm_get_library_entry (library_name, TRUE);
+
+	MonoDlWasmFunctionEntry *entry = (MonoDlWasmFunctionEntry *)g_malloc0 (sizeof (MonoDlWasmFunctionEntry));
+	entry->next = library->functions;
+	entry->name = g_strdup (name);
+	entry->ptr = ptr;
+	library->functions = entry;
+}
+
+
 const char *
 mono_dl_get_so_prefix (void)
 {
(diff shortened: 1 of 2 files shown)


kekekeks:
@kumpera It also seems  to work fine with JavaScript functions. 

I've added an additional `--js-library`:
```js
mergeInto(LibraryManager.library, {
  js_foo: function(number) {
    alert("Number" + number);
  },
});
```
imported it with 
```c
extern void js_foo(int num)
```

and then added it with 
```c
mono_dl_wasm_add_internal_pinvoke(NULL, "js_foo", js_foo);
```

It's being called just fine. Certainly that means that said function should be WASM-aware, just like everything that's being called from a regular C code compiled by emscripten. So I'm not sure what you have meant by 'the marshaling on the C# side is non trivial'. Were you referring to accessing JS objects?

migueldeicaza:
This code looks like it duplicates existing functionality that already exists in `mono_dl_register_library` and `mono_dl_fallback_register`.

Is there a reason to get this that is not covered by the other two?

Tooling-wise, I would expect that we could make it simple to pass a set of native libraries, and we would register those to `mono_dl_fallback_register`

kumpera:
Miguel is right that `mono_dl_register_library` / `mono_dl_fallback_register` should do the trick. Perhaps I broke something in the wasm backend and it's not working.
I'll that it works today.

mono/utils/mono-dl-wasm.c
```diff
@@ -12,6 +12,63 @@
...
+typedef struct MonoDlWasmLibraryEntry
+{
+	char *name;
+	MonoDlWasmFunctionEntry *functions;
+	struct MonoDlWasmLibraryEntry *next;
+
+} MonoDlWasmLibraryEntry;
+
+
+static MonoDlWasmLibraryEntry* mono_dl_wasm_entries = NULL;
```
weitzhandler: Wouldn't it be better to rename 'entries' to 'libraries', so we can consistently distinguish between `MonoDlWasmLibraryEntry` and `MonoDlWasmFunctionEntry`?
Also in the local variables inside the functions.

mono/utils/mono-dl-wasm.c
```diff
@@ -12,6 +12,63 @@
...
+
+
+static MonoDlWasmLibraryEntry *
+mono_dl_wasm_get_library_entry (const char* name, int create)
+{
+	if (name == NULL)
+		name = "__Internal";
+
+	MonoDlWasmLibraryEntry *entry = mono_dl_wasm_entries;
```
weitzhandler: Rename `entry` to library.

mono/utils/mono-dl-wasm.c
```diff
@@ -12,6 +12,63 @@
...
+	entry->functions = NULL;
+	mono_dl_wasm_entries = entry;
+	return entry;
+}
+
+void mono_dl_wasm_add_internal_pinvoke (const char *library_name, const char *name, void *ptr)
+{
+	struct MonoDlWasmLibraryEntry *library = mono_dl_wasm_get_library_entry (library_name, TRUE);
+
+	MonoDlWasmFunctionEntry *entry = (MonoDlWasmFunctionEntry *)g_malloc0 (sizeof (MonoDlWasmFunctionEntry));
```
weitzhandler: Rename `entry` to `function`.

weitzhandler:
@kekekeks 
Thank you so much for this continuous effort! You rawwwkz!

kekekeks:
@migueldeicaza `mono_dl_register_library` can't handle `__Internal`, unfortunately.

`__Internal` is being replaced by `NULL` at https://github.com/mono/mono/blob/d16ca8d0373e5f525fd0bb3442ffc6e8b4877a24/mono/metadata/loader.c#L1257

So handling it requires to register `NULL` as a library name which isn't supported by g_str_hash which requires strings to be non-NULL

kumpera:
@kekekeks is that's all that's blocking mono_dl_register_library, I'll fix that. :)

kekekeks:
I've moved my code to `driver.c` and made it use `mono_dl_fallback_register` and it seems to be working. I don't see why `mono_dl_register_library` shouldn't.

kekekeks:
I'm now encountering `CANNOT HANDLE COOKIE VIIFFI` in `icall_trampoline_dispatch` which currently blocks further attempts of using Skia.

kekekeks:
Is it possible `icall_trampoline_dispatch` to also be pluggable from `driver.c`? P/Invoke kinda requires to generate implementations for imported function signatures.

luhenry:
@kumpera should we accept this change, or is everything already available with `mono_dl_fallback_register` / `mono_dl_fallback_register`?

roozbehid-ic:
@kekekeks and @kumpera so this PR is closed without merging? I also need this PR, is there a reason it is not merged?

kekekeks:
@roozbehid-ic `mono_dl_register_library` and `mono_dl_fallback_register` should work if you don't need to handle `__Internal`

roozbehid:
Well I want to use dllimport to call a function defined in driver.c. How can I do that. I thought I have to use "__Internal" for that purpose

kekekeks:
You can register another name e. g. "MyInterop" and use it in [DllImport]

roozbehid:
got you. thanks

Issue #8360, opened 2018-04-20
[System.Net.Sockets] flakey failures of SocketTest.TestSelect1 on Darwin systems


lewurm:
I couldn't reproduce it locally yet, but it seems to happen often enough.

I first saw it on the XI 2018-02 device tests run:
https://github.com/xamarin/xamarin-macios/pull/3402#issuecomment-382528658

later in the tvOS simulator, again for XI 2018-02:
https://github.com/xamarin/xamarin-macios/pull/3402#issuecomment-382815289

and now the mono OS X i386 lane (master):
https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/9798/testReport/junit/(root)/SocketTest/TestSelect1/

lewurm:
can someone with knowledge in this area verify that the test itself does make sense?
https://github.com/mono/mono/blob/42d425cb62db513146b1a463e22323f4d5b577f4/mcs/class/System/Test/System.Net.Sockets/SocketTest.cs#L253-L315

lewurm:
it looks like a bit fishy: Is `Thread.Sleep` used as a synchronisation primitive in this test?

filipnavara:
I am really concerned about the structure of the test... Can you try what happens if you:
1) Add Thread.Sleep at the beginning of ConnectSleepClose method?
2) Add Thread.Sleep between th.Start and srv.Accept calls?

I am almost certain it will affect the outcome one way or another and just show that the test has race condition.

lewurm:
If I move `clnt.Write ()` after the first `Select ()`, it fails in the same way, but I don't know how this could happen with the current test. If I try what you suggested @filipnavara, it ends up with exceptions being thrown due to accessing a disposed object (which is a problem with this test case, but not the problem that we are seeing).

I submitted #8364 which at least gets rid of the one factor of uncertainty (thread).

akoeplinger:
XI issue: https://github.com/xamarin/xamarin-macios/issues/3239

PR #8364, opened 2018-04-20
[System.Net] do not use Thread for SocketTest.TestSelect1

lewurm:
it makes the test less reliable. Potentially helps with
https://github.com/mono/mono/issues/8360


lewurm:
I agree @akoeplinger, but I don't know what that reason would be. The original commit doesn't mention anything regarding that: https://github.com/lewurm/mono/commit/fd02ac62fa1da66b5b7adaed205138824d6ea0f8

luhenry:

luhenry:

Issue #8100, opened 2018-04-07
NuGet Restore crashes runtime on Ubuntu 18.04 ARM


directhex:
https://jenkins.mono-project.com/blue/rest/organizations/jenkins/pipelines/ng-release-generate-debian-packages/runs/799/nodes/297/steps/459/log/?start=0

```Restoring NuGet package ICSharpCode.Decompiler.3.0.0.3403-beta4.
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) Mono.Btls.MonoBtlsSsl.mono_btls_ssl_connect (intptr) <0x00037>
  at Mono.Btls.MonoBtlsSsl.Connect () [0x00011] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at Mono.Btls.MonoBtlsContext.DoProcessHandshake () [0x00028] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at Mono.Btls.MonoBtlsContext.ProcessHandshake () [0x00009] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at Mono.Net.Security.MobileAuthenticatedStream.ProcessHandshake (Mono.Net.Security.AsyncOperationStatus) [0x0003e] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at (wrapper remoting-invoke-with-check) Mono.Net.Security.MobileAuthenticatedStream.ProcessHandshake (Mono.Net.Security.AsyncOperationStatus) [0x00032] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at Mono.Net.Security.AsyncHandshakeRequest.Run (Mono.Net.Security.AsyncOperationStatus) [0x00006] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at Mono.Net.Security.AsyncProtocolRequest/<ProcessOperation>d__24.MoveNext () [0x000ff] in <eddcf0f405ca4b228ee5ec6bf726ef42>:0
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore/MoveNextRunner.InvokeMoveNext (object) [0x00000] in <cf49dfc616a841db9a7a6637587a11b0>:0
  at System.Threading.ExecutionContext.RunInternal (System.Threading.ExecutionContext,System.Threading.ContextCallback,object,bool) [0x00071] in <cf49dfc616a841db9a7a6637587a11b0>:0
[...]

PR #8366, opened 2018-04-20
Turn off BTLS assembly optimizations on 32-bit ARM.

directhex:
See https://github.com/mono/mono/issues/8100


luhenry:
@monojenkins build Linux x64

Issue #7053, opened 2018-02-14
Make the interpreter pass all BCL tests


luhenry:
We want the interpreter to be on par with the JIT in regards to the BCL test suite. As we don't have support for running tests in parallel we should add them in the following order.

For all assemblies include both add xunit and nunit versions.

- [x] mscorlib
- [x] mscorlib / xunit
- [x] Mono.Debugger.Soft
- [x] Mono.Profiler.Log
- [x] System.Core
- [x] System.Core / xunit
- [x] System.Xml
- [x] System.Numerics
- [x] System.Numerics / xunit
- [x] Microsoft.CSharp
- [ ] Microsoft.CSharp / xunit
- [x] System.Security
- [ ] System.Security / xunit
- [x] System
- [x] System / xunit
- [x] System.Data
- [x] System.Data / xunit

luhenry:
We could have a lane, but there are still around 50 tests that are failing. We should disable these tests for now, and fix them as we go.

marek-safar:
We should focus on core (mobile) assemblies only

luhenry:
Only very few assemblies left after https://github.com/mono/mono/pull/9648

lewurm:
https://github.com/mono/mono/pull/10095 added `System.Core / xunit`.

@BrzVlad do you remember what's the issue with `Microsoft.CSharp / xunit`?

BrzVlad:
@lewurm I don't remember. I stressed  it for some time with a recent mono on amd64 and it it passed all the time. Maybe we should enable it.

PR #8370, opened 2018-04-20
[interp] add System tests to CI

lewurm:
https://github.com/mono/mono/issues/7053

LSWCIS (let's see what CI says 😛 )


marek-safar:
I am not very keen on this one. I think we should run full mscorlib tests instead

luhenry:
@marek-safar this is in complement to mscorlib, not a replacement, so what do you mean in https://github.com/mono/mono/pull/8370#issuecomment-383228514?

marek-safar:
Sure, it's complement of other dlls but there is not that much interesting stuff in System.dll for the interpreter. We should add the interesting assemblies first and only if the run is not too long add System.dll.

Right now we run only about 10k of mscorlib tests and ignore remaining 40k+ in there.

lewurm:
> Right now we run only about 10k of mscorlib tests and ignore remaining 40k+ in there.

oh, how would I run the remaining 40k+?

lewurm:

PR #8371, opened 2018-04-20
[interp] add System.Core tests to CI

lewurm:
https://github.com/mono/mono/issues/7053

LSWCIS (let's see what CI says 😛 )

should be good, as @kumpera is already running those tests as part of the webassembly tests


marek-safar:
Please try xunit version there are more interesting cases than we have

lewurm:
@marek-safar will do in a follow-up PR

lewurm:

PR #8385, opened 2018-04-20
Use g_print to dump threads when SIGQUIT is sent to Mono

grendello:
`printf`, `fprintf` don't work on Android, using g_print will redirect the output to
`mono-stdout` so that it can appear in Android `logcat` output.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -3785,7 +3785,7 @@ dump_thread (MonoInternalThread *thread, ThreadDumpUserData *ud)
 		}
 	}
 
-	fprintf (stdout, "%s", text->str);
+	g_print ("%s", text->str);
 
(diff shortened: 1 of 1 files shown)


mono/metadata/threads.c
```diff
@@ -3785,7 +3785,7 @@ dump_thread (MonoInternalThread *thread, ThreadDumpUserData *ud)
 		}
 	}
 
-	fprintf (stdout, "%s", text->str);
+	g_print (text->str);
```
jaykrell: What if text->str contains percent signs?
use g_print ("%s", text->str).

luhenry:

luhenry:

luhenry:

luhenry:

PR #8406, opened 2018-04-23
[Mono.Profiler.Log] Use AssertExtensions for comparison asserts

akoeplinger:
`Assert.True (actual >= expected)` doesn't log useful information when
the condition is false, making it hard to see what went wrong:

https://jenkins.mono-project.com/job/test-mono-mainline/label=osx-i386/8525/testReport/junit/Test%20collection%20for%20MonoTests.Mono.Profiler.Log/ProfilerTests/MonoTests_Mono_Profiler_Log_ProfilerTests_MonitorEventsAreRecorded/

Unfortunately xUnit doesn't provide a built-in comparison assert but
CoreFX already built one so we can just use that one.

### DIFF
--- a/mcs/class/Mono.Profiler.Log/Makefile
+++ b/mcs/class/Mono.Profiler.Log/Makefile
@@ -9,7 +9,7 @@ KEYFILE = $(LIBRARY_SNK)
 LIB_REFS = System System.Core
 LIB_MCS_FLAGS = /unsafe /publicsign /nowarn:0618
 
-XTEST_LIB_REFS = System System.Core
+XTEST_LIB_REFS = System System.Core Facades/System.Threading.Tasks
 
 xunit-test-local: log-profiler-test.exe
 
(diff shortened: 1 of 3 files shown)


mcs/class/Mono.Profiler.Log/Makefile
```diff
@@ -9,7 +9,7 @@ KEYFILE = $(LIBRARY_SNK)
 LIB_REFS = System System.Core
 LIB_MCS_FLAGS = /unsafe /publicsign /nowarn:0618
 
-XTEST_LIB_REFS = System System.Core
+XTEST_LIB_REFS = System System.Core Facades/System.Threading.Tasks
```
luhenry: What's the need for this added  facade?

luhenry:

luhenry:

luhenry:

luhenry:

luhenry:

mcs/class/Mono.Profiler.Log/Makefile
akoeplinger: it's needed by the xunit assemblies (we already add it on all the other assemblies that run xunit tests):

```
/Users/alexander/dev/mono/external/corefx/src/CoreFx.Private.TestUtilities/src/System/AssertExtensions.cs(20,26): error CS0012: The type 'Task' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Threading.Tasks, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```

luhenry:

PR #8419, opened 2018-04-23
[Tailcall] Let watchos make more tailcalls.

jaykrell:
This is a subset of https://github.com/mono/mono/pull/7963, perhaps most of what might be interesting to customers.

arm32 will have a bunch of holes in the matrix, but seemingly watchos is the main or half the interesting targets and this shores it up, while leaving more consistency elsewhere.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -7239,7 +7239,7 @@ is_supported_tailcall (MonoCompile *cfg, const guint8 *ip, MonoMethod *method, M
 		//
 		// Interface method dispatch has the same problem (imt_arg).
 
-		|| IS_NOT_SUPPORTED_TAILCALL ((vtable_arg || imt_arg) && !cfg->backend->have_volatile_non_param_register)
+		|| IS_NOT_SUPPORTED_TAILCALL ((vtable_arg || imt_arg) && !cfg->backend->have_volatile_non_param_register && !cfg->llvm_only)
 		) {
 		tailcall_calli = FALSE;
 		tailcall = FALSE;


luhenry:
@jaykrell is this going to make more tailcall available on watchos temporarily (until it gets fixed on other platforms) or is it going to favor `llvmonly`/`bitcode` only?

jaykrell:
It is in-between. There is an arm32-specific problem, that is likely permanent.
It lifts that problem, in a way that from a customer point of view, is just for watch.
It doesn't address arm32iphone which doesn't really matter, nor arm32android which is declining.

jaykrell:
I said it wrong perhaps. It lets watchos match x86/arm64/amd64, in a way that other arm32 platforms probably never will.

jaykrell:
Otherwise arm32 falls behind across the board, but for desktop and iphone that doesn't really matter.

luhenry:
bitcode is only available on watchos, so trying to fix `ARMv5`/`ARMv7` by letting the bitcode-specific code handle it is not the way to go. I would prefer to have a `ARMv5`/`ARMv7` wide lack of functionality rather than a watchOS specific fix.

luhenry:
A comparable case is float32: it's not available across all of x86 because it's too much work, so instead of fixing a specific platform, it's just disabled across the whole x86 architecture.

jaykrell:
"I would prefer to have a ARMv5/ARMv7 wide lack of functionality rather than a watchOS specific fix."
That is the critical decision, ok.
float32 is a little less comparable in that tailcall lets code not crash (i.e. for large input), not "just" run faster.
I'll try to find another solution much later.

luhenry:
Don't be so hyperbolic 😉 tailcall is an optimization, which, by definition, is not required for the program to run correctly. Also, don't twist my word: "having a watchos specific fix (which doesn't work on other ARMv5/ARMv7 platforms) as the long term fix" is what I don't want. I would encourage you to figure out something that works across all ARMv5/ARMv7 platforms, but as long as you tell me "tailcall on ARMv5/ARMv7 is not possible, except with `llvmonly`", then it's not going to happen. I hope you understand what we are trying to achieve here, and if you don't, please don't hesitate to ask questions, and don't simply disparage it.

jaykrell:
There are two kinda contradictory view of tailcall.
1. "It is just an optimization", if you miss it occasionally, ok.
2. It is required by some programs, in order to process large data, without crashing (stack overflow). This isn't really hyperbole. This is how some code is written, with some validity. Granted, within C# you have to use /optimize and mutual recursion to stand a chance. F# is much more able to produce such code.

Many people see #1 and miss #2.

I understand the desire for consistency, and that an arm32/llvmonly-only fix is far from ideal. But I am very pessimistic that there is another solution here. I know of a few maybe-solutions, but they don't seem likely. I'll try to put this off while dealing with other easier limitations, and should see what netcore does.

A thing is though, that Linuxdesktop/arm32 isn't a customer scenario, iphone/arm32 is seemingly gone per Apple policy as a customer scenario (maybe targeting old OS???), leaving only two customer arm32 platforms as I understand: watch and android. So the proposed partial fix fixes a full half the problem, a higher fraction than if more arm32 platforms were interesting. That is, I wouldn't want a 20% solution, I would definitely like a 100% solution, but 50% seems maybe ok.

1. Use intra-procedural scratch r20 for imt_arg/vtable_arg.
This would require quite some change, to confidentally fix its current uses and prevent other use.
gcc uses this for nested function context, so there is hope, but not much.
In particular, I'm not sure there is a way to force dynamic loader resolution.
You can't use this technique with any risk of traversing dynamic loading.

2. Use r9 I think it is. This might be easy. It doesn't work on very old iphone, probably no matter. It probably always works on watchos, good. It isn't clear about Linux/android.

3. Use the "regular ABI", like llvmonly does, but use it for mini/jit/etc. There is at least precedent that we do this sometimes, code is there already subject to flags. It does grow code size and it'd have be done pretty much everywhere I think, not just in the presence of tailcalls.

4. Store it in a thread local. Awfully inefficient.

5. Store it in redzone. I don't think arm32 has one.

6. If we control all native to managed transitions, then the current scheme might just work.
Every native to managed transition could not be a tailcall. That is likely already the case, another albeit smaller thorn in our side. Mutual recursion involving native/managed transitions cannot likely be fully tailcalled.

I think the most likely solutions by far are 0. do nothing and 3. use the regular/llvmonly ABI when not using llvmonly.

Anyway other fish in the fryer a while. Let me ask about #3 other folks.

jaykrell:
Hm that approach 3 might have trouble anyway -- requiring a thunk, that breaks tailcall.

To fix that, you'd have to consider every function as taking an extra parameter and pass it as 0 all over the place and the thunk could replace it, w/o requiring returning through the thunk. Not great. Large change.

Maybe r9. Maybe. It should work on watch at least, not just llvmonly. Not sure about Linux/android.

PR #8428, opened 2018-04-24
[jit] Decompose CEE_BOX later, to enable more optimization opportunities. Optimize Enum.HasFlags ().

vargaz:
Also mark the temporary bblocks used by decompose_vtypes () as out-of-line if
the bblock under processing is out-of-line. Make the domainvar as volatile to avoid
optimizing it away since references to it can now be generated later from the
decomposition of OP_BOX.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -9485,7 +9485,8 @@ calli_end:
 				break;
 			}
 
-			if (mono_class_is_enum (klass)) {
+			if (mono_class_is_enum (klass) && !(val->type == STACK_I8 && SIZEOF_VOID_P == 4)) {
+				/* Can't do this with 64 bit enums on 32 bit since the vtype decomp pass is ran after the long decomp pass */
 				if (val->opcode == OP_ICONST) {
 					MONO_INST_NEW (cfg, ins, OP_BOX_ICONST);
 					ins->type = STACK_OBJ;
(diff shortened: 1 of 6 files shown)


mono/mini/method-to-ir.c
```diff
@@ -10605,8 +10610,42 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				break;
 			}
 
-			*sp++ = handle_box (cfg, val, klass, context_used);
+			if (mono_class_is_nullable (klass) || mini_is_gsharedvt_klass (klass) || context_used) {
```
kumpera: I'm curious, why can't we emit box ops for those 3 cases? It would be nice to document the constraints.

mono/mini/method-to-ir.c
```diff
@@ -10605,8 +10610,42 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
+			} else {
+				MONO_INST_NEW (cfg, ins, OP_BOX);
+				ins->type = STACK_OBJ;
+				ins->klass = klass;
+				ins->sreg1 = val->dreg;
+				ins->dreg = alloc_dreg (cfg, val->type);
+
+				switch (val->type) {
+				case STACK_I4:
+				case STACK_I8:
```
kumpera: Hmmmmmmm, how does this handle longs on 32bits? Don't we need a OP_LBOX that takes as input the regpair?

vargaz: Possible.

vargaz:
So it looks like this is still not enough to support the HasFlags optimization in mini_emit_inst_for_method (), since it will receive two OP_BOX instructions, and has no access to the input operands of those instructions. So the optimization cannot really be done there.

lewurm:
LGTM, but it seems to make roslyn crash more often

vargaz:
@monojenkins build

vargaz:
There seems to be some random failures during the aot process.

PR #8433, opened 2018-04-24
Cleanup instruction sizing and code reallocation.

jaykrell:


jaykrell:
This should really all be std::vector<uchar>::reserve and push_back.

jaykrell:
Crashing though . it worked locally, will have to try this much more carefully.

jaykrell:
close while I work on it morea

PR #8434, opened 2018-04-24
check_call_signature return gboolean and run second time if break, to step through in debugger

jaykrell:


vargaz:
What is the purpose of this ? You can call check_call_signature () from the debugger after hitting the breakpoint.

jaykrell:
I'm not experienced with that -- can set a breakpoint on the function and then have debugger call it and step through it? I'm used to kernel debuggers where the debugger definitely cannot call user code so I rarely try to have debugger make calls, and they do seem to act funny in windbg. But it is a debugging convenience in any case.

vargaz:
At least in gdb, you can do
b <func>
and p <func> (args)
and it will break.

jaykrell:
Take the boolean? Or neither?

jaykrell:
Sorry to make repeat self -- but to ensure no offense/mistake, entire thing approved, or just the boolean?

PR #6639, opened 2018-01-23
Thread_Abort coop/handle conversion

jaykrell:

PR #8435, opened 2018-04-24
[Coop] Abort_internal and ResetAbortNative to coop/handle.

jaykrell:
This is a subset of https://github.com/mono/mono/pull/6639/files to see if it doesn't experience the hangs on Windows that the larger PR encountered.


mono/metadata/threads.c
```diff
@@ -2514,7 +2557,8 @@ ves_icall_System_Threading_Thread_ResetAbort (MonoThread *this_obj)
 
 	if (!was_aborting) {
 		const char *msg = "Unable to reset abort because no abort was requested";
-		mono_set_pending_exception (mono_get_exception_thread_state (msg));
+		mono_set_pending_exception_handle (mono_exception_new_thread_state (msg, error));
+		mono_error_assert_ok (error);
```
jaykrell: This assert is confusing but probably ok. It is asserting that creating the exception succeeded. i.e. not low memory. The actual assert created and set is in the thread.

jaykrell:
Looking ok on Windows? Want another Windows run?

jaykrell:

jaykrell:

PR #8436, opened 2018-04-24
Cleanup instruction sizing and code reallocation.

jaykrell:


mono/mini/mini-amd64.c
```diff
@@ -50,6 +50,8 @@
 #include "mini-runtime.h"
 #include "aot-runtime.h"
 
+#define EXTRA_CODE_SPACE (16)
```
kumpera: Is this still needed as the constant was moved to mini.h?

mono/mini/mini.c
```diff
@@ -4402,13 +4402,26 @@ mono_add_patch_info (MonoCompile *cfg, int ip, MonoJumpInfoType type, gconstpoin
 	g_assert_not_reached ();
 }
 
+#else // DISABLE_JIT
+
+guint8*
+mini_realloc_code_slow (MonoCompile *cfg, int size)
+{
+	const int EXTRA_CODE_SPACE = 16;
```
kumpera: Couldn't this const be shared with the one on mini.h?

mono/mini/mini-amd64.c
jaykrell: I left it a local variable in mini.h but yes we could make it a define there.

mono/mini/mini.c
jaykrell: As a #define (or const static) yes. I was limiting the scope but agree it isn't great.

mono/mini/mini-amd64.c
```diff
@@ -6880,9 +6870,12 @@ mono_arch_emit_prolog (MonoCompile *cfg)
 
 	cfg->arch.stack_alloc_size = alloc_size;
 
+	set_code_cursor (cfg, code);
```
jaykrell: Some of the set_code_cursor calls are redundant.

The idea is, we optimize out the realloc checks in the existing code, so we don't need code_len to be accurate/up-to-date -- only before a realloc check. It isn't pretty, but it is incremental.

But some of the calls are required, so that realloc can use code_len.

mono/mini/mini-amd64.c
jaykrell: EXTRA_CODE_SPACE in mini-amd64.c will go away with https://github.com/mono/mono/pull/8437.
We already pad out the base tailcall size to 255, and then with 8437 we can count up the parameters accurately, so the extra 64 seems like overkill upon overkill (the 255 is overkill, the parameter counting only slightly overkill)

jaykrell:
arm emit_float_args needs to increase max_len, the out parameter.
Arguably the .md should increase, but does the realloc so ok.

jaykrell:

jaykrell:

jaykrell:

jaykrell:
Accidentally pushed to wrong branch, reseting CI.

jaykrell:

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build Linux i386

jaykrell:
@monojenkins build Linux i386

PR #8437, opened 2018-04-24
[Tailcall] Better sizing of tailcalls, allow arm64 r4, copy arm64 stacked parameters

jaykrell:
Better sizing of tailcalls via new OP_TAILCALL_PARAMETER.
Allow arm64 r4.
Copy arm64 stacked parameters.
Better interface to instruction sizes.
Adding possibly missing reallocs.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -2252,6 +2252,25 @@ test_tailcall (MonoCompile *cfg, MonoBoolean tailcall)
 	tailcall_print ("tailcalllog %s from %s\n", tailcall ? "success" : "fail", cfg->method->name);
 }
 
+static void
+emit_tailcall_parameters (MonoCompile *cfg, MonoMethodSignature *sig)
+{
+	// OP_TAILCALL_PARAMETER helps compute the size of code, in order
+	// to size branches around OP_TAILCALL_[REG,MEMBASE].
+	//
+	// The actual bytes are output from OP_TAILCALL_[REG,MEMBASE].
+	// OP_TAILCALL_PARAMETER is an overestimate because typically
+	// many parameters are in registers.
+
+	int n = sig->param_count + sig->hasthis + 1; // always at least one
+	for (int i = 0; i < n; ++i) {
+		MonoInst *ins;
+		MONO_INST_NEW (cfg, ins, OP_TAILCALL_PARAMETER);
+		MONO_ADD_INS (cfg->cbb, ins);
+	}
+
+}
+
 static MonoCallInst *
 mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig, 
 		     MonoInst **args, gboolean calli, gboolean virtual_, gboolean tailcall,
@@ -2281,6 +2300,7 @@ mono_emit_call_args (MonoCompile *cfg, MonoMethodSignature *sig,
 
 	if (tailcall) {
 		mini_profiler_emit_tail_call (cfg, target);
+		emit_tailcall_parameters (cfg, sig);
 		MONO_INST_NEW_CALL (cfg, call, calli ? OP_TAILCALL_REG : virtual_ ? OP_TAILCALL_MEMBASE : OP_TAILCALL);
 	} else
 		MONO_INST_NEW_CALL (cfg, call, ret_type_to_call_opcode (cfg, sig->ret, calli, virtual_));
@@ -8410,10 +8430,13 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				/* Handle tailcalls similarly to calls */
 				DISABLE_AOT (cfg);
 
+				MonoMethodSignature *sig = mono_method_signature (cmethod);
+
+				emit_tailcall_parameters (cfg, sig);
 				MONO_INST_NEW_CALL (cfg, call, OP_TAILCALL);
 				call->method = cmethod;
 				call->tailcall = TRUE;
-				call->signature = mono_method_signature (cmethod);
+				call->signature = sig;
 				call->args = (MonoInst **)mono_mempool_alloc (cfg->mempool, sizeof (MonoInst*) * n);
 				call->inst.inst_p0 = cmethod;
 				for (i = 0; i < n; ++i)
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -8430,13 +8430,11 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				/* Handle tailcalls similarly to calls */
 				DISABLE_AOT (cfg);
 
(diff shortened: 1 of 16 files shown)


mono/mini/mini-arm.c
```diff
@@ -1758,7 +1758,10 @@ mono_arch_tailcall_supported (MonoCompile *cfg, MonoMethodSignature *caller_sig,
 	if (!res && !debug_tailcall)
 		goto exit;
 
-	res &= IS_SUPPORTED_TAILCALL (callee_info->stack_usage <= 16 * 4); // FIXME why?
+	// FIXME The limit here is that moving the parameters requires addressing the parameters
+	// with 12bit (4K) immediate offsets.
+	res &= IS_SUPPORTED_TAILCALL (callee_info->stack_usage < 4096);
```
jaykrell: Maybe feels safer to stay away from the limit? 1k? 2k, 3k?

jaykrell:
Thoughts?

mono/mini/mini.h
```diff
@@ -1698,6 +1698,12 @@ extern const char MONO_ARCH_CPU_SPEC [];
 extern const guint16 MONO_ARCH_CPU_SPEC_IDX(MONO_ARCH_CPU_SPEC) [];
 #define ins_get_spec(op) ((const char*)&MONO_ARCH_CPU_SPEC + MONO_ARCH_CPU_SPEC_IDX(MONO_ARCH_CPU_SPEC)[(op) - OP_LOAD])
 
+static inline unsigned
+ins_get_size (int opcode)
```
vargaz: This can return simply an int.

mono/mini/mini-arm64.c
```diff
@@ -4340,7 +4340,13 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
 			}
 
-			// FIXME: Copy stack arguments
+			// Copy stack arguments.
+			// FIXME a fixed size memcpy is desirable here,
+			// at least for larger values of stack_usage.
+			for (i = 0; i < call->stack_usage; i += sizeof (mgreg_t)) {
+				code = emit_ldrx (code, ARMREG_LR, ARMREG_SP, i);
+				code = emit_strx (code, ARMREG_LR, ARMREG_R28, i);
+			}
```
vargaz: Why does this use R28 ?

mono/mini/method-to-ir.c
```diff
@@ -8408,10 +8428,13 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				/* Handle tailcalls similarly to calls */
 				DISABLE_AOT (cfg);
 
+				MonoMethodSignature *sig = mono_method_signature (cmethod);
+
+				emit_tailcall_parameters (cfg, sig);
```
vargaz: There is already an fsig computed above.

mono/mini/method-to-ir.c
```diff
@@ -8408,10 +8428,13 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
 				MONO_INST_NEW_CALL (cfg, call, OP_TAILCALL);
 				call->method = cmethod;
 				call->tailcall = TRUE;
-				call->signature = mono_method_signature (cmethod);
+				call->signature = sig;
 				call->args = (MonoInst **)mono_mempool_alloc (cfg->mempool, sizeof (MonoInst*) * n);
```
vargaz: This should use fsig.

vargaz:
Looks ok otherwise.

mono/mini/mini.h
jaykrell: Ok. It has to cast to unsigned char first though, to get the full 0-255 range.

mono/mini/mini-arm64.c
jaykrell: What should it use? cfg->arch.args_reg?

jaykrell: Same thing, right?

jaykrell:
I wonder if this shouldn't be two separate PRs.
They are related -- the sizing is needed, if we are to loop over stacked parameters without bound.
r4 and stacked arm64 parameter handling could be handled together.
At least the subject lists it all.

mono/mini/method-to-ir.c
jaykrell: Agreed, fixed.

mono/mini/method-to-ir.c
jaykrell: Agreed, fixed.

mono/mini/mini-arm64.c
jaykrell: I still confused on this one -- what should it be?

mono/mini/mini-amd64.c
```diff
@@ -4650,12 +4661,9 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 
 			g_assert (!cfg->method->save_lmf);
 
-			/* the size of the tailcall op depends on signature, let's check for enough
-			 * space in the code buffer here again */
-			max_len += AMD64_NREG * 4 + call->stack_usage * 15 + EXTRA_CODE_SPACE;
-			max_len += 64; // FIXME make this two pass for an accurate size
-
-			if (G_UNLIKELY (offset + max_len > cfg->code_size)) {
+			max_len += AMD64_NREG * 4;
```
jaykrell: This AMD64_NREG  * 4 should be part of the .md,and perhaps is, but leaving it alone for now.

mono/mini/mini-amd64.c
```diff
@@ -4650,12 +4661,9 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
+			max_len += call->stack_usage / sizeof (mgreg_t) * ins_get_size (OP_TAILCALL_PARAMETER);
```
jaykrell: Removing this line is kinda the point, depending on how else the rest of the code works, which I'm not sure of.

mono/mini/mini-arm64.c
```diff
@@ -4331,14 +4351,23 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 				g_assert (ins->inst_basereg != -1);
 				g_assert (ins->inst_basereg != ARMREG_IP0);
 				g_assert (ins->inst_basereg != ARMREG_IP1);
+				g_assert (ins->inst_basereg != ARMREG_LR);
```
jaykrell: assert could be relaxed, i.e. if stack_usage == 0, which is common, but ok.

jaykrell:
@monojenkins build Linux i386

jaykrell:
I'm a little confused as to where accurate length is needed.
call updates max_len, e.g. arm emit_float_args.
For realloc purposes, length can be computed late.
Is that it? Or should arm call .md be increased?

PR #8442, opened 2018-04-24
[wasm] Add Sample binding engine.

kumpera:


kumpera:
Interp aarch64 failed with `Error: Garbage collector could not allocate 67108864 bytes of memory for nursery. `

aarch64 failed System with: ```
1) AuthenticateClientAndServer_ClientSendsNoData (MonoTests.System.Net.Security.SslStreamTest.AuthenticateClientAndServer_ClientSendsNoData)
   System.IO.IOException : The authentication or decryption has failed.
  ----> System.IO.IOException : Error while sending TLS Alert (Fatal:InternalError): System.IO.IOException: The authentication or decryption has failed. ---> System.ObjectDisposedException: Cannot access a disposed object.
```

**Comment by kumpera** [comment]

**Comment by kumpera** [comment]

File: sdks/wasm/bindings.cs [hunk 1]
```diff
@@ -0,0 +1,174 @@
...
+		Add JS helpers to fetch wrapped methods, like to Module.cwrap
+		Better Wrap C# exception when passing them as object (IE, on task failure)
+		Make JSObject disposable (same for js objects)
+	*/
+	public sealed class Runtime {
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern string InvokeJS (string str, out int exceptional_result);
+
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		internal static extern object InvokeJSWithArgs (int js_obj_handle, string method, object[] _params, out int exceptional_result);
```
alexanderkyte: Are these all safe for concurrent access? I can't quite picture where you'd need to lock, but I feel like I see some accesses and mutations that would be unsafe concurrently. Things like `(int)(IntPtr)obj.Handle`.

sdks/wasm/bindings.cs
```diff
@@ -0,0 +1,174 @@
...
+		static void SetupJSContinuation (Task task, JSObject cont_obj) {
+			if (task.GetType () == typeof (Task)) {
+				task.GetAwaiter().OnCompleted (() => {
+					//FIXME we should dispose cont_obj after completing the Promise
+					if (task.Exception != null)
+						cont_obj.Invoke ("reject", task.Exception.ToString ());
+					else
+						cont_obj.Invoke ("resolve", null);
+				});
+			} else {
```
alexanderkyte: Could you comment what you expect this to be when `task.GetType () != typeof(Task)` ?

sdks/wasm/main.cs
```diff
@@ -11,26 +11,6 @@
...
-	}
-
-	public class JSException : Exception {
-		public JSException (string msg) : base (msg) {}
-	}
-}
-
 public class Driver {
 	static void Main () {
 		Console.WriteLine ("hello");
```
alexanderkyte: Hello

sdks/wasm/test.js
```diff
@@ -147,6 +119,24 @@ var send_message = find_method (driver_class, "Send", -1)
...
+var binding_test_module = assembly_load ("binding_tests");
+
+//binding test suite support code
+var binding_test_class = find_class (binding_test_module, "", "TestClass");
+if (!binding_test_class)
+	throw 9;
+
+function call_test_method(method_name, signature, args)
+{
+	var target_method = find_method (binding_test_class, method_name, -1)
```
alexanderkyte: Why -1? To get rid of magic numbers, could we make this a `#define FIND_METHOD_UNLIMITED` or something?

sdks/wasm/binding_support.js
```diff
@@ -0,0 +1,363 @@
...
+
+		set_task_result: function (tcs, result) {
+			this.call_method (this.set_tcs_result, null, "oo", [ tcs, result ]);
+		},
+
+		set_task_failure: function (tcs, reason) {
+			this.call_method (this.set_tcs_failure, null, "os", [ tcs, reason.toString () ]);
+		},
+
+		js_to_mono_obj: function (js_obj) {
```
alexanderkyte: This looks correct assuming the caller uses it correctly, but at some point I feel this function is going to be the function that throws when we have nulls/bugs propagate from elsewhere in the code. Some more error handling might make it easier to file bugs that throw here.

sdks/wasm/bindings.cs
kumpera: So, JS is single threaded so this is not an issue today.
Even later, with the multithreaded wasm proposal, the JS heap will only be accessible from the main thread.

sdks/wasm/binding_support.js
kumpera: This support code is only visible to the binding engine, users can't call it.
The only thing it's visible is call_method, cuz I exported it early, in the `BINDING__postset` part

sdks/wasm/bindings.cs
kumpera: It's going to be `Task<T>`.

Notice that `task` is of type `Task` so the only possible outcome of `task.GetType () != typeof (Task)` is a subclass of it.

What will happen in the case of a different subclass is that we'll throw an exception at runtime.

sdks/wasm/test.js
kumpera: Good point, I'll change the code to use a constant.

alexanderkyte:
Looks good to me, after the constant change let's merge this.

luhenry:
@kumpera what’s the status of this PR?

PR #8445, opened 2018-04-25
[Tailcall] Fix patching of generic GOT under FullAOT.

jaykrell:
This fixes bug https://github.com/mono/mono/issues/8273 where FullAOT generic-shared (valuetype?) runs hundreds of times slower than regular AOT due to hitting the trampoline for every call.

[Coop] Fork test 8273.il from 3.il and increase iteration count
so it takes a long time previously.

Also possibly fix a JIT path where gshared but not gsharedvt handled.

### DIFF
--- a/mono/mini/mini-generic-sharing.c
+++ b/mono/mini/mini-generic-sharing.c
@@ -4114,25 +4114,28 @@ mini_is_gsharedvt_variable_signature (MonoMethodSignature *sig)
 }
 
 MonoMethod*
-mini_method_to_shared (MonoMethod *method)
+mono_method_to_shared (MonoMethod *method)
 {
 	if (!mono_method_is_generic_impl (method))
-		return NULL;
+		return method;
 
 	ERROR_DECL (error);
+
+	MonoMethod *shared_method;
 
 	// This pattern is based on add_extra_method_with_depth.
 
 	if (mono_method_is_generic_sharable_full (method, TRUE, TRUE, FALSE))
 		// gshared over reference type
-		method = mini_get_shared_method_full (method, SHARE_MODE_NONE, error);
+		shared_method = mini_get_shared_method_full (method, SHARE_MODE_NONE, error);
 	else if (mono_method_is_generic_sharable_full (method, FALSE, FALSE, TRUE))
 		// gshared over valuetype (or primitive?)
-		method = mini_get_shared_method_full (method, SHARE_MODE_GSHAREDVT, error);
+		shared_method = mini_get_shared_method_full (method, SHARE_MODE_GSHAREDVT, error);
 	else
-		return NULL;
+		return method;
+	g_assert (shared_method != method);
 	mono_error_assert_ok (error);
-	return method;
+	return shared_method;
 }
 
 #else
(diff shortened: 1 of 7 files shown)


mono/tests/tailcall/8273.il
```diff
@@ -0,0 +1,221 @@
...
+
+// Using string instead of int32 here makes this test really slow,'
+// so use a low iteration count -- anything small is ok to test tailcall.
+
+ldstr "a"
+//ldc.i4 1
+newobj instance void class A`1<string>::'.ctor'(!0)
+ldstr "b"
+//ldc.i4 2
+ldc.i8 99999999
```
jaykrell: This takes between half a full second on fast machine.
I'm not sure we want that.
I'm not sure how to properly test this.
It works before and after, but very slow before.

mono/mini/mini-generic-sharing.c
```diff
@@ -4059,6 +4059,29 @@ mini_is_gsharedvt_variable_signature (MonoMethodSignature *sig)
...
+	// This pattern is based on add_extra_method_with_depth.
+
+	if (mono_method_is_generic_sharable_full (method, TRUE, TRUE, FALSE))
+		// gshared over reference type
+		method = mini_get_shared_method_full (method, SHARE_MODE_NONE, error);
+	else if (mono_method_is_generic_sharable_full (method, FALSE, FALSE, TRUE))
+		// gshared over valuetype (or primitive?)
+		method = mini_get_shared_method_full (method, SHARE_MODE_GSHAREDVT, error);
+	else
+		return NULL;
```
jaykrell: Arguably this should return method. All the callers want that. But this preserves more clearly information as to if it was shared (which can also be gleaned perhaps by comparing output to input).

mono/mini/mini-generic-sharing.c
```diff
@@ -4097,4 +4120,10 @@ mini_is_gsharedvt_variable_signature (MonoMethodSignature *sig)
 	return FALSE;
 }
 
+MonoMethod*
+mini_method_to_shared (MonoMethod *method)
+{
+	return NULL;
```
jaykrell: NULL vs. method -- same as other ifdef fork.

mono/mini/mini.h
```diff
@@ -2604,4 +2604,7 @@ gboolean    mono_class_is_magic_float (MonoClass *klass);
 MonoInst*   mono_emit_native_types_intrinsics (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSignature *fsig, MonoInst **args);
 MonoType*   mini_native_type_replace_type (MonoType *type) MONO_LLVM_INTERNAL;
 
+MonoMethod*
+mini_method_to_shared (MonoMethod *method); // null if not shared
```
jaykrell: Should probably be called "mono_" not "mini_" since it worked on MonoMethod.

jaykrell:
CI all passes on original commit. Second is bikesheddy.

mono/mini/mini-trampolines.c
```diff
@@ -792,6 +792,7 @@ common_call_trampoline (mgreg_t *regs, guint8 *code, MonoMethod *m, MonoVTable *
 		 */
 		if (domain_jit_info (domain)->jump_target_got_slot_hash) {
 			GSList *list, *tmp;
+			m = mono_method_to_shared (m);
```
vargaz: Would be better to move this whole block of code to mini_patch_jump_sites ().

jaykrell: Also..mini_patch_jump_sites locks more briefly, but leaks the list?
This will also make one instead of two conversion to shared.
Merge them with another PR or in this PR?

jaykrell: Looks like my "bikeshed" broke things. I'll undo that.

jaykrell:

PR #8455, opened 2018-04-25
[interp] enable mcs tests on armv7

lewurm:
there was a technical limitiation to run it reliably (aka. "the r8 problem"), which was resolved by https://github.com/mono/mono/pull/8056


lewurm:
it's green, but I'm not sure if we should merge: it takes around 90 minutes.

akoeplinger:
Hm, I'd say we shouldn't enable it as this job runs on all PRs and we're already low on ARM capacity.

You could enable it for non-PR runs only (by checking whether the CI_TAGS contain `pull-request`) so it gets some exposure during the nightly runs.

luhenry:
@akoeplinger we are low, but we are not low to the point we are limiting ourselves on what we really care about, and the interpreter is one of these projects, so we really should run this on all PR. If we want to lower our load, we should instead reduce the frequency of some of our master builds, so we don't come to a full build queue every day.

akoeplinger:
@luhenry we already moved all ARM master builds to be nightly so there's not a lot left :)

If we really think there's huge value in testing on both ARM64 and ARM on each PR then so be it and we'll live with it.

PR #8458, opened 2018-04-26
Free or destroy, of any type, should do nothing, silently, given a null pointer.

jaykrell:
Remove excess null checks.

### DIFF
--- a/mono/eglib/ghashtable.c
+++ b/mono/eglib/ghashtable.c
@@ -545,8 +545,9 @@ void
 g_hash_table_destroy (GHashTable *hash)
 {
 	int i;
-	
-	g_return_if_fail (hash != NULL);
+
+	if (!hash)
+		return;
 
 	for (i = 0; i < hash->table_size; i++){
 		Slot *s, *next;
--- a/mono/metadata/metadata.c
+++ b/mono/metadata/metadata.c
@@ -4236,7 +4236,7 @@ mono_metadata_free_mh (MonoMethodHeader *mh)
 	 * or a SRE-generated method, so the lifetime in that case is
 	 * dictated by the method's own lifetime
 	 */
-	if (mh->is_transient) {
+	if (mh && mh->is_transient) {
 		for (i = 0; i < mh->num_locals; ++i)
 			mono_metadata_free_type (mh->locals [i]);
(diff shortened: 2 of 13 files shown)


akoeplinger:
doesn't this deviate from the original glib contract? or do we not care?

jaykrell:

luhenry:
This goes against what was discussed in https://github.com/mono/mono/pull/8226, and this is just resubmitting in a different form what was already refused.

jaykrell:
This is the approved subset. That one also skipped iteration, and in that one free & destory were both approved.

jaykrell:
It doesn’t break assumptions that matter. It changes the interface slightly in a way that is reasonable to all callers & cleans things up a little.

luhenry:
@jaykrell I disagree, it breaks a big assumption that matters: when passing an incorrect argument, it would previously break, and not it would just silently succeed. This is the same kind of problem where you print a warning instead of an assertion: it's most likely you are just going to ignore the error, and even more importantly, it might mask a bigger/worse bug. A not-so-old case that we hit in Mono: we would previously silently ignore any error returned from any `pthread_mutex_*` call, and when not silently ignoring these errors anymore, we discovered a whole set of bugs which we wouldn't previously hit (use-after-destroy, non-initialized, etc.).

Keeping the conditions of execution tight greatly help us in reasoning about the code, and what possible cases are possible. Your change break these assumptions and just make the code harder to follow in general.

jaykrell:
@lambdageek's assertion here:
  https://github.com/mono/mono/pull/8226#issuecomment-381161580
is what this PR specifically does.

kumpera:
NULL ignoring free/destroy is a common pattern in C. It does help a lot dealing with optional data. We gain nothing by forcing callees to check for those.

Worse than that, probably a lot of the callees are missing NULL checks because we never ran into the error condition that triggers it.

I don't think the comparison between pthread_* return values and ignoring NULLs is fair. In the former, we're in an unknown state, while the later we're just handling one valid form of input.

jaykrell:
Good point about the potential missing null checks that this fixes/covers-up.
Overly strict assertion failures lying dormant, about harmless conditions.

PR #8459, opened 2018-04-26
Fix leak in mini_patch_jump_sites.

jaykrell:


jaykrell:
Maybe not. mono_domain_alloc0 and then? Let me see.

PR #8470, opened 2018-04-26
Expose two public types

joshpeterson:
These types are public in the .NET 4.7.1 profile, so they should be public in Mono as well.


akoeplinger:
@monojenkins commit apidiff

PR #8476, opened 2018-04-26
[tests] Fix out-of-tree build.

madewokherd:


mono/tests/Makefile.am
```diff
@@ -1458,6 +1458,7 @@ tailcall/coreclr/JIT/opt/Tailcall/TailcallVerifyWithPrefix.exe: \
 		tailcall/coreclr/JIT/opt/Tailcall/TailcallVerifyVerifiableLibraryWithPrefix.il
 
 %.exe: %.il
+	$(MKDIR_P) $(dir $@)
```
jaykrell: Can say $(@D) for the directory.
mkdir should have - in front of it, maybe it does.

jaykrell: I had addressed in an earlier PR but it was rejected as unrelated.

PR #8479, opened 2018-04-27
[Coop] Convert thread interrupt, resume, join, cleanup raising ThreadState exceptions.

jaykrell:


jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OSx x86

jaykrell:
@monojenkins build os x i386

mono/metadata/threads.c
```diff
@@ -2699,16 +2704,23 @@ mono_thread_resume (MonoInternalThread *thread)
...
+
+	if (!internal_thread) {
+		exception = TRUE;
 	} else {
-		LOCK_THREAD (thread->internal_thread);
-		if (!mono_thread_resume (thread->internal_thread))
-			mono_set_pending_exception (mono_get_exception_thread_state ("Thread has not been started, or is dead."));
-		UNLOCK_THREAD (thread->internal_thread);
+		LOCK_THREAD (internal_thread);
+		if (!mono_thread_resume (internal_thread))
```
jaykrell: Move exception out of the lock. Similar cases are not locked, and the raise is not on the same thread (?) as the locking -- or can it be on self?

mono/metadata/threads.c
```diff
@@ -1514,6 +1512,18 @@ thread_handle_to_internal_ptr (MonoThreadObjectHandle thread_handle)
 	return MONO_HANDLE_GETVAL(thread_handle, internal_thread); // InternalThreads are always pinned.
 }
 
+static void
+mono_error_set_exception_thread_state (MonoError *error, const char *exception_message)
```
lambdageek: I usually try to make new `mono_error_set_FOO` functions varargs (calling `mono_error_set_generic_errorv`) so that the call site can have an easy way to put more context into the messages.

mono/metadata/threads.c
```diff
@@ -1947,7 +1957,7 @@ mono_join_uninterrupted (MonoThreadHandle* thread_to_join, gint32 ms, MonoError
 	MonoThreadInfoWaitRet ret;
 	gint32 wait = ms;
 
-	gint64 start = (ms == -1) ? 0 : mono_msec_ticks ();
+	const gint64 start = (ms == -1) ? 0 : mono_msec_ticks ();
```
lambdageek: 👍

mono/metadata/threads.c
lambdageek: Could add an assert to find out.

mono/metadata/threads.c
```diff
@@ -2545,14 +2553,11 @@ ves_icall_System_Threading_Thread_ResetAbort (MonoThreadObjectHandle this_obj, M
...
-		const char *msg = "Unable to reset abort because no abort was requested";
-		mono_set_pending_exception_handle (mono_exception_new_thread_state (msg, error));
-		mono_error_assert_ok (error);
-		return;
-	} else if (is_domain_abort) {
-		/* Silently ignore abort resets in unloading appdomains */
+		mono_error_set_exception_thread_state (error, "Unable to reset abort because no abort was requested");
 		return;
 	}
+	if (is_domain_abort) // Silently ignore abort resets in unloading appdomains
```
lambdageek: I think it was clearer as an `else if`.

mono/metadata/threads.c
jaykrell: None of them provide such information currently, though it could be added later.

mono/metadata/threads.c
jaykrell: return does not continue through the if ladder. else is also confusing.

mono/metadata/threads.c
jaykrell: Resetting CI but ok.

jaykrell: Or just put it back in the lock?

jaykrell: Raise under lock or not? It is always best to hold locks as short of a time/range as possible, and no shorter. The key is finding the minimum sufficient way.

lambdageek: That's true, I guess there's a strdup at least.

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build Linux aarch64

PR #8483, opened 2018-04-27
[Coop] Remove around 140 uncooperative pointers due to calls to mono_set_pending_exception ().

jaykrell:
Bottleneck them all in small helper functions, most of which avoid pointers already, one of which needs further work.

### DIFF
--- a/mono/metadata/exception-internals.h
+++ b/mono/metadata/exception-internals.h
@@ -71,4 +71,49 @@ mono_exception_new_thread_abort (MonoError *error);
 MonoExceptionHandle
 mono_exception_new_thread_state (const char *msg, MonoError *error);
 
+void
+mono_set_pending_exception_class_failure (MonoClass *klass);
+
+void
+mono_set_pending_exception_remoting (const char *message);
+
+void
+mono_set_pending_exception_appdomain_unloaded (void);
+
+void
+mono_set_pending_exception_argument (const char *name, const char *message);
+
+void
+mono_set_pending_exception_argument_null (const char *name);
+
+void
+mono_set_pending_exception_argument_out_of_range (const char *name);
+
+void
+mono_set_pending_exception_divide_by_zero (void);
+
+void
+mono_set_pending_exception_execution_engine (const char *message);
+
+void
+mono_set_pending_exception_index_out_of_range (void);
+
+void
+mono_set_pending_exception_invalid_operation (const char *message);
+
+void
+mono_set_pending_exception_not_supported (const char *message);
+
+void
+mono_set_pending_exception_overflow (void);
+
+void
+mono_set_pending_exception_synchronization_lock (const char *message);
+
+void
+mono_set_pending_exception_thread_interrupted (void);
+
+void
+mono_set_pending_exception_not_implemented (const char *message);
+
 #endif
--- a/mono/metadata/exception-internals.h
+++ b/mono/metadata/exception-internals.h
@@ -71,49 +71,4 @@ mono_exception_new_thread_abort (MonoError *error);
 MonoExceptionHandle
 mono_exception_new_thread_state (const char *msg, MonoError *error);
 
-void
-mono_set_pending_exception_class_failure (MonoClass *klass);
-
-void
-mono_set_pending_exception_remoting (const char *message);
-
-void
-mono_set_pending_exception_appdomain_unloaded (void);
-
-void
-mono_set_pending_exception_argument (const char *name, const char *message);
-
-void
-mono_set_pending_exception_argument_null (const char *name);
-
-void
-mono_set_pending_exception_argument_out_of_range (const char *name);
-
-void
-mono_set_pending_exception_divide_by_zero (void);
-
-void
-mono_set_pending_exception_execution_engine (const char *message);
-
-void
-mono_set_pending_exception_index_out_of_range (void);
-
-void
-mono_set_pending_exception_invalid_operation (const char *message);
-
-void
-mono_set_pending_exception_not_supported (const char *message);
-
-void
-mono_set_pending_exception_overflow (void);
-
-void
-mono_set_pending_exception_synchronization_lock (const char *message);
-
-void
-mono_set_pending_exception_thread_interrupted (void);
-
-void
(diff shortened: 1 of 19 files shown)


mono/metadata/exception.c
```diff
@@ -290,7 +290,7 @@ mono_exception_from_name_two_strings_checked (MonoImage *image, const char *name
  * \returns the initialized exception instance.
  */
 static MonoExceptionHandle
-mono_exception_new_by_name_msg (MonoImage *image, const char *name_space,
+mono_exception_new_by_name_message (MonoImage *image, const char *name_space,
```
lambdageek: Why rename this?

mono/metadata/exception-internals.h
```diff
@@ -71,4 +71,49 @@ mono_exception_new_thread_abort (MonoError *error);
...
+
+void
+mono_set_pending_exception_synchronization_lock (const char *message);
+
+void
+mono_set_pending_exception_thread_interrupted (void);
+
+void
+mono_set_pending_exception_not_implemented (const char *message);
+
```
lambdageek: We are trying to discourage the implicit use of the pending exception in favor of explicitly threading `MonoError*` through the code - rather than adding a dozen new pending exception functions, I think we should be trying to replace uses of `mono_set_pending_exception (mono_exception_get_FOO(...))` by `mono_error_set_FOO (error, ...)` followed by a `mono_error_set_pending_exception(error)` or by routing the mono-error out of a `HANDLES()` icall.

mono/metadata/exception.c
jaykrell: msg is too short.

mono/metadata/exception-internals.h
jaykrell: I think that can come later but it is also optional I believe. This is an easy effective mechanical change. For example if the functions don't have pointers, this suffices. And if they gain handles() and MonoError, this still works. Other than removing a small amount of code, what is the advantage of the other approach?

jaykrell:
This does not make complete sense.

For example:
```
mono/mini/jit-icalls.c:mono_llmult_ovf_un (guint64 a, guint64 b)
mono/mini/jit-icalls.h:guint64 mono_llmult_ovf_un (guint64 a, guint64 b);
mono/mini/mini-runtime.c:       register_opcode_emulation (OP_LMUL_OVF_UN, "__emul_lmul_ovf_un", "long long long", mono_llmult_ovf_un, "mono_llmult_ovf_un", FALSE);
```

how do you convert that? We still need the pending exception and no MonoError.

jaykrell:
The way this is now formed, there are groups of 2 or usually 3 lines.
As functions are further visited, in some cases, two lines will be removed, but the third (middle) line should remain unchanged.
In many cases the functions will not be further visited however and these will stand asis. For example the math stuff.

jaykrell:
Strange seeming compile errors. It worked for me. Will look tonight.

jaykrell:
I think we should consider a class of icalls that create no objects, possibly traffic in no objects, except in the rare error path. Accessing thread locals outside of the error path is very inefficient.

jaykrell:
Possibly something fragile here & we should take the original. Or maybe simply missing a handle frame, will check soon.

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X x64

jaykrell:
Something still wrong here, will retry & dig again.

jaykrell:
@monojenkins build OS X i386

jaykrell:
One incorrect change around ‘endinvoke_called’, to see if that is the problem,  before retrying first commit — removing 140+ raw pointers while leaving pending exceptions alone seems like good progress.

jaykrell:

mono/metadata/exception.c
```diff
@@ -436,6 +436,7 @@ mono_exception_new_thread_interrupted (MonoError *error)
  * mono_get_exception_thread_interrupted:
  * \returns a new instance of the \c System.Threading.ThreadInterruptedException
  */
+MONO_RT_EXTERNAL_ONLY
```
lambdageek: Curious why in a couple of places you put `MONO_RT_EXTERNAL_ONLY` in the C file - they should be in `exception.h`

mono/metadata/exception.c
```diff
@@ -570,6 +571,7 @@ mono_get_exception_type_load (MonoString *class_name, char *assembly_name)
  * \param msg the message to pass to the user
  * \returns a new instance of the \c System.NotImplementedException
  */
+MONO_RT_EXTERNAL_ONLY
```
lambdageek: Should be in exception.h

mono/metadata/exception.c
```diff
@@ -851,6 +853,7 @@ mono_get_exception_type_initialization_checked (const gchar *type_name, MonoExce
  * \param inner the inner exception.
  * \returns a new instance of the \c System.SynchronizationLockException
  */
+MONO_RT_EXTERNAL_ONLY
```
lambdageek: Should be in exception.h

mono/metadata/exception.c
```diff
@@ -1437,3 +1440,10 @@ mono_error_set_simple_file_not_found (MonoError *error, const char *file_name, g
 	else
 		mono_error_set_file_not_found (error, file_name, "Could not load file or assembly '%s' or one of its dependencies.", file_name);
 }
+
+void
+mono_error_set_argument_out_of_range (MonoError *error, const char *name)
+{
+	//FIXMEcoop
+	mono_error_set_exception_instance (error, mono_get_exception_argument_out_of_range (name));
```
lambdageek: I added a task #8516 to fix this up later.

jaykrell:
I find it helps read the code to mark it in both places — so you see it no matter where you are looking.

luhenry:
@jaykrell we don't mark functions with `MONO_RT_EXTERNAL_ONLY` in the `.c` file, only in the `.h` file. Please, respect the coding convention of the project (not the first time it was requested).

jaykrell:
@monojenkins build Linux x64

PR #8497, opened 2018-04-27
[profiler] Do GC state transitions in profiler helper threads.

### LINKED ISSUE #8355: Fix Mono.Profiler.Log-xunit tests under hybrid suspend
The `Mono.Profiler.Log-xunit` testsuite is failing on Jenkins all platforms using hybrid suspend.

[hybrid coop master builds](https://jenkins.mono-project.com/job/test-mono-mainline-linux-hybrid-suspend/) and every PR, too: [Linux AArch64](https://jenkins.mono-project.com/view/Pull%20requests/job/test-mono-pull-request-hybrid-suspend-arm64/) and [Linux x64](https://jenkins.mono-project.com/view/Pull%20requests/job/test-mono-pull-request-hybrid-suspend/)

lambdageek:
After the calls to `profiler_thread_check_detach`, the GC doesn't skip
the profiler helper threads anymore (because of the call to
`mono_thread_info_set_flags (MONO_THREAD_INFO_FLAGS_NONE)`.

Under hybrid or full cooperative GC, the helper threads in GC Unsafe
mode should not block without periodically polling.  Otherwise we get
into a deadlock on shutdown when finalizing the root domain:  The profiler
threads have been told to detach and we run a GC in preparation for
running the finalizers.  The GC STW ends up waiting for the dumper and
writer threads which never self-suspend.

---

In the mach and windows suspend code, don't add the suspending thread to the pending operations set. 

Under hybrid suspend if the suspend initiator raced with
done_blocking/abort_blocking, then the thread self-suspended and is already
waiting for the resume semaphor - there's nothing else for the suspend initiator
to do, so don't add victim to the pending operation set.

---

Fixes #8355

### DIFF
--- a/mono/profiler/log.c
+++ b/mono/profiler/log.c
@@ -3420,7 +3420,9 @@ writer_thread (void *arg)
 	MonoProfilerThread *thread = profiler_thread_begin ("Profiler Writer", FALSE);
 
 	while (mono_atomic_load_i32 (&log_profiler.run_writer_thread)) {
+		MONO_ENTER_GC_SAFE;
 		mono_os_sem_wait (&log_profiler.writer_queue_sem, MONO_SEM_FLAGS_NONE);
+		MONO_EXIT_GC_SAFE;
 		handle_writer_queue_entry ();
 
 		profiler_thread_check_detach (thread);
@@ -3533,11 +3535,15 @@ dumper_thread (void *arg)
 	MonoProfilerThread *thread = profiler_thread_begin ("Profiler Dumper", TRUE);
 
 	while (mono_atomic_load_i32 (&log_profiler.run_dumper_thread)) {
+		gboolean timedout = FALSE;
+		MONO_ENTER_GC_SAFE;
 		/*
 		 * Flush samples every second so it doesn't seem like the profiler is
 		 * not working if the program is mostly idle.
 		 */
-		if (mono_os_sem_timedwait (&log_profiler.dumper_queue_sem, 1000, MONO_SEM_FLAGS_NONE) == MONO_SEM_TIMEDWAIT_RET_TIMEDOUT)
+		timedout = mono_os_sem_timedwait (&log_profiler.dumper_queue_sem, 1000, MONO_SEM_FLAGS_NONE) == MONO_SEM_TIMEDWAIT_RET_TIMEDOUT;
+		MONO_EXIT_GC_SAFE;
+		if (timedout)
 			send_log_unsafe (FALSE);
 
 		handle_dumper_queue_entry ();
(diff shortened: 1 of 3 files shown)


lambdageek:
Linux x64 Hybrid Suspend is green 🎉

alexrp:
@lambdageek I don't fully understand the change here. Immediately after the `mono_thread_info_set_flags (MONO_THREAD_INFO_FLAGS_NONE)`, the thread calls `mono_thread_detach (mono_thread_current ())`. So it's unclear to me why waiting on a semaphore after that point could be problematic for coop as the GC should no longer care about the thread at all.

lambdageek:
@alexrp It's a race in this chunk of code:

https://github.com/mono/mono/blob/23cfac00c572ea43165e8fd0b12ef8909056d6e0/mono/mini/mini-runtime.c#L4763-L4776

raising `runtime_shutdown_begin` will toggle the dumper thread to `MONO_THREAD_INFO_FLAGS_NONE`, then the call to `mono_domain_finalize` on the main thread will start a major GC https://github.com/mono/mono/blob/23cfac00c572ea43165e8fd0b12ef8909056d6e0/mono/metadata/gc.c#L443

At that point the dumper thread is still attached but not skippable.  So under coop we will wait for it to checkpoint and self-suspend

alexrp:
OK, that makes sense.

Question: What happens when a thread does `MONO_ENTER_GC_SAFE` while having `MONO_THREAD_FLAGS_NO_GC` set? Will it block if a GC happens? This matters because the profiler threads have to be unaffected by STW under normal conditions (it doesn't really matter during shutdown).

lambdageek:
@alexrp when a thread performs a `MONO_ENTER_GC_SAFE` it will just toggle its thread state from running to blocking (and the reverse on exit).  If it's marked with `MONO_THREAD_FLAGS_NO_GC`, that's the end of the story - the GC will skip the thread during stop-the-world - the running/blocking thread state will just be a little bit of irrelevant bit twiddling.

If a thread is not marked as skippable, then it's a question of suspend policy:
 - under cooperative suspend, if a thread in the blocking state is requested to suspend, it will toggle to the "blocking suspend requested" state, the GC will then assume it's suspended and the thread will continue executing. If it reaches `MONO_EXIT_GC_SAFE` before it receives a resume notification it will toggle into a self-suspended state (and from there it will step back into running state once resumed) .
- under hybrid suspend and preemptive suspend, a thread in blocking state will be async-suspended (using signals on linux, or using the native suspend APIs on Darwin and Win32) and resumed back into blocking state.

There's a picture of the possible transitions here: http://www.mono-project.com/docs/advanced/runtime/docs/coop-suspend/#current-implementation

coop state machine

alexrp:
OK, makes sense, thanks for the explanation. Then this PR looks reasonable to me.

PR #8506, opened 2018-04-29
Do not set thread name repeatedly.

jaykrell:


jaykrell:
This isn't as bad as it looks, or even perhaps bad at all.
The repetition appears not to be per work item, but per scheduling quantum, due to:

```
                //
                // Loop until our quantum expires.
                //
                while ((Environment.TickCount - quantumStartTime) < ThreadPoolGlobals.tpQuantum)
```

and if a thread can be a threadpool worker, and then not, and back and worth, then setting the name repeatedly is appropriate. I suspect that threads never stop being a threadpool worker but I could be wrong.

jaykrell:

luhenry:
We need to reset the thread name because it might be changed in managed, and the API wants the thread name to be reset.

vargaz:
We reset the thread name because that is what MS.NET does.

PR #8507, opened 2018-04-29
More thread join coop conversion and cleanup and some timekeeping cleanup.

jaykrell:


mono/metadata/threads.c
```diff
@@ -1952,42 +1973,6 @@ mono_thread_internal_current_handle (void)
 	return MONO_HANDLE_NEW (MonoInternalThread, mono_thread_internal_current ());
 }
 
-static MonoThreadInfoWaitRet
```
luhenry: This function was extracted to more easily review and implement it. Don't remove it. (PS: look at the history).

mono/metadata/threads.c
```diff
@@ -1952,42 +1973,6 @@ mono_thread_internal_current_handle (void)
...
-	const gint64 start = (ms == -1) ? 0 : mono_msec_ticks ();
-	while (TRUE) {
-		MONO_ENTER_GC_SAFE;
-		ret = mono_thread_info_wait_one_handle (thread_to_join, wait, TRUE);
-		MONO_EXIT_GC_SAFE;
-
-		if (ret != MONO_THREAD_INFO_WAIT_RET_ALERTED)
-			return ret;
-
-		MonoException *exc = mono_thread_execute_interruption_ptr ();
```
luhenry: The only useful change in this whole PR is changing 
```
MonoException *exc = mono_thread_execute_interruption_ptr ();
if (exc) {
	mono_error_set_exception_instance (error, exc);
	return ret;
}
```
to
```
MonoExceptionHandle exc = MONO_HANDLE_NEW (MonoException, NULL); // And yes this should be out of the `while(TRUE)` loop so you don't repetitively allocate handles
if (mono_thread_execute_interruption (&exc)) {
	mono_error_set_exception_handle (error, exc);
	break;
}
```

jaykrell:
@lambdageek & I discussed this & some other parts.

The inlining is useful because it counteracts the lack of coop handle frame. Should either inline or add a frame.

The return type is incorrect.

The unlock is duplicated.

The timekeeping could use more cleanup. It converts unnecessarily between -1 and infinity — same thing.

jaykrell:
Also extra call to set pending exception.

mono/metadata/threads.c
jaykrell: It seems no easier to review or implement either way. The non-inline version needs a frame, to be resilient against other callers materializing, as discussed in another PR and this resolves that without growing the code. If there was another caller, or a unit test just for the function, or the function was large,  or cold vs. its one caller, those would be good reasons for keeping it separate.

jaykrell:
@monojenkins build failed

PR #8508, opened 2018-04-29
[Coop] Convert Microsoft.Win32.NativeMethod to coop.

jaykrell:
These are all pretty trivial. There are no pointers or handles.
The only small wrinkle is internal calls -- split the function into w/ and w/o MonoError*.


lambdageek:
@monojenkins squash

PR #8523, opened 2018-04-30
[msbuild] Zip test results and log files for surfacing in jenkins

radical:

### DIFF
--- a/packaging/MacSDK/msbuild.py
+++ b/packaging/MacSDK/msbuild.py
@@ -7,6 +7,7 @@ class MSBuild (GitHubPackage):
 
 	def build (self):
 		self.sh ('./build.sh -host mono -configuration Release -skipTests')
+		self.sh ('zip msbuild-bin-logs.zip artifacts/Release-MONO/log/*')
 
 	def install (self):
 		# use the bootstrap msbuild as the system might not have one available!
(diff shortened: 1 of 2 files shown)


scripts/ci/run-test-mac-sdk.sh
```diff
@@ -8,6 +8,7 @@ export PATH=${MONO_REPO_ROOT}/external/bockbuild/stage/bin:$PATH
 # Bundled MSBuild
 cd ${MONO_REPO_ROOT}/external/bockbuild/builds/msbuild-15/
 ${TESTCMD} --label="msbuild-tests" --timeout=180m ./build.sh -host mono -configuration Release
+zip msbuild-test-results.zip artifacts/2/Release-MONO/TestResults/*
```
akoeplinger: this should use `${MONO_REPO_ROOT}/msbuild-test-results.zip` so it ends up in the repo root

radical: Ok. How do I get the same path in `msbuild.py`, to copy the binlogs?

akoeplinger:

scripts/ci/run-test-mac-sdk.sh
akoeplinger: I actually don't know, we probably need to bug Alexis about it :)

packaging/MacSDK/msbuild.py
```diff
@@ -7,6 +7,7 @@ def __init__ (self):
 
 	def build (self):
 		self.sh ('./build.sh -host mono -configuration Release -skipTests')
+		self.sh ('zip msbuild-bin-logs.zip artifacts/Release-MONO/log/*')
```
radical: @akoeplinger @alexischr  How do I get the path for `mono_repo` here, short a hardcoded relative path?

PR #8500, opened 2018-04-27
[Tailcall PowerPC] OP_TAILCALL_PARAMETER support.

jaykrell:

PR #8524, opened 2018-04-30
[ppc] Fix build

directhex:
Build fix from https://github.com/mono/mono/pull/8500#issuecomment-385438385

Already tested on `xam-power8-1`


directhex:

directhex:
Runtime builds on PPC64; abusing admin powers to merge immediately

Issue #8356, opened 2018-04-19
Hybrid suspend: backtrace in GC Unsafe mode considered harmful (on Linux)


lambdageek:
It can race with the dynamic linker 😠 

This is from https://jenkins.mono-project.com/job/test-mono-mainline-linux-hybrid-suspend/label=debian-9-armel/5/parsed_console/log.html

Thread 1 is... I don't even know. But it's apparently GC Safe, so we async suspend it. it's holding a lock that Thread 4 needs.  Thread 4 just did a GC Safe -> GC Unsafe transition (so we want to cooperatively suspend it), and it wants to record the state transition and is collecting a backtrace.  So it ends up in `dl_iterate_phdr` and tries to lock a mutex, but Thread 1 got it, so T4 waits.

Meanwhile the suspend initiator is waiting for T1 to coop suspend itself, and the whole runtime is stuck.

We previously had a similar issue where two threads were racing for a backtrace because they were both doing thread transitions. Attempted fix here: https://github.com/mono/mono/blob/006d6ce09734e8b555a53324c21bbaab3f1ebcb4/mono/utils/checked-build.c#L200-L232

But this is worse - this time the suspended thread was not doing a `backtrace` - it just needed the linker for whatever reason.

```
Thread 4 (Thread 0xf52ff450 (LWP 20777)):
#0  0xf7b381b8 in __lll_lock_wait () from /lib/arm-linux-gnueabi/libpthread.so.0
#1  0xf7b30630 in pthread_mutex_lock () from /lib/arm-linux-gnueabi/libpthread.so.0
#2  0xf7abf0dc in dl_iterate_phdr () from /lib/arm-linux-gnueabi/libc.so.6
#3  0xf7ac00f4 in __gnu_Unwind_Find_exidx () from /lib/arm-linux-gnueabi/libc.so.6
#4  0xf7b12454 in ?? () from /lib/arm-linux-gnueabi/libgcc_s.so.1
#5  0xf7b12b58 in ?? () from /lib/arm-linux-gnueabi/libgcc_s.so.1
#6  0xf7b1358c in _Unwind_Backtrace () from /lib/arm-linux-gnueabi/libgcc_s.so.1
#7  0xf7a92c04 in backtrace () from /lib/arm-linux-gnueabi/libc.so.6
#8  0x00d1c4bc in collect_backtrace (out_data=0xf5000cf0) at checked-build.c:224
[...]

PR #8530, opened 2018-04-30
[threads] Don't call backtrace on BLOCKING->RUNNING transitions

lambdageek:
See https://github.com/mono/mono/issues/8356

Backtrace can deadlock with threads that have been suspended while holding a
dynamic linker lock on Linux.

In checked build mode we collect backtraces on state machine transitions.

When we transition from BLOCKING->RUNNING under hybrid suspend the thread must
not block without polling.  So we must not collect a backtrace on this
transition.


luhenry:

luhenry:
@monojenkins build Linux x64

luhenry:

PR #8532, opened 2018-05-01
nuget-hash-extractor: Update to track xplat-master

radical:


radical:
`OS X x64`: failing test is `TemplateControlCompilerTest.NoBindForMethodsWithBindInNameTest`, should not be related to this PR. This PR effectively doesn't make any change.

PR #8533, opened 2018-05-01
[Tailcall] Fix CoreCLR tests to return success:0 instead of success:100 and run some.

jaykrell:


jaykrell:
Makefile.am might still need tweaks, might need to disable stuff still, and maybe I missed some .cs/.il, and likely I modified .cs/.il that fail for other reasons, but I expect the .cs/.il changes to be approved and such.

jaykrell:
@monojenkins build OS X x64

jaykrell:

jaykrell:

luhenry:
The build sim broken.

jaykrell:
Yes there is a net increase in enabled tests, but not quite all of them should be enabled, still working through that.

jaykrell:
Oh and working through them I had a typo too.

mono/tests/Makefile.am
```diff
@@ -1287,6 +1209,7 @@ LLVM = $(filter --llvm, $(MONO_ENV_OPTIONS))
 # threads-init.exe: runs out of system threads
 # dim-constrainedcall.exe: fails on dontnet as well (https://github.com/dotnet/coreclr/issues/15353)
 # tailcall-rgctxb.exe seems like it would but it does not, needs debugging
+# tailcall-member-function-in-valuetype.exe never works, for understood reason
```
luhenry: What understood reasons?

jaykrell: Or i’ll it inline, later.

jaykrell:
Any time we have a byref/out parameter, be it
an integer, or valuetype-this, we pessimistically
will never tailcall.

In this case we have valuetype-this.

I suggest we insert range checks.
If all parameters are outside the frame,
do a tailcall. If any are not, regular call.

It'll be a little bigger and have a slowdown,
and a speedup (stack locality speeds things up),
and enable code that does not work otherwise.

Again, tailcall is not just an optimization.
It lets code process large data within small stack.
If it was just an optimization, desktop CLR would not
go so far out of their way to implement it.

mono/tests/Makefile.am
jaykrell: Any time we have a byref/out parameter, be it
an integer, or valuetype-this, we pessimistically
will never tailcall.

In this case we have valuetype-this.

I suggest we insert range checks.
If all parameters are outside the frame,
do a tailcall. If any are not, regular call.

It'll be a little bigger and have a slowdown,
and a speedup (stack locality speeds things up),
and enable code that does not work otherwise.

Again, tailcall is not just an optimization.
It lets code process large data within small stack.
If it was just an optimization, desktop CLR would not
go so far out of their way to implement it.

jaykrell:
@monojenkins build failed

jaykrell:

luhenry:
@jaykrell there are code conflicts, could you please rebase on master? Thank you.

jaykrell:
Of course.

jaykrell:
@monojenkins build Linux x64

jaykrell:

jaykrell:
@monojenkins build API Diff

PR #8536, opened 2018-05-01
Update monollvm to 0b3cb8ac12cd839f8110775d4085e822e8af4d7b for mono/tests/generic-stack-traces2.2

jaykrell:
commit 0b3cb8ac12cd839f8110775d4085e822e8af4d7b (HEAD -> master, origin/master, origin/HEAD)
	Author: Zoltan Varga <vargaz@gmail.com>
	Date:   Tue May 1 14:52:22 2018 -0400

	Emit an LSDA for methods without landing pads too since we need the information about the 'this' register.

to fix generic-stack-traces2.2 -- so it gets a generic type definition
instead of a generic type instance, consistent with other execution modes.

### DIFF
--- a/packaging/MacSDK/mono-llvm.py
+++ b/packaging/MacSDK/mono-llvm.py
@@ -5,7 +5,7 @@ class MonoLlvmPackage (GitHubPackage):
 
     def __init__(self):
         GitHubPackage.__init__(self, 'mono', 'llvm', '3.0',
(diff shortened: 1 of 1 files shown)


luhenry:
It's always good to bump, but this is not going to fix the test: you bumped the Mac package's llvm version, and this has nothing to do with our CI.

jaykrell:
This is what I told to do and it was approved. I really didn't know what I was doing or how to test it (in terms of CI, I did test it locally).

Issue #8488, opened 2018-04-27
Doesn't compile with Clang on ARM64


directhex:
```
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:32:2: error: instruction requires: crypto
08:54:17  pmull v0.1q,v20.1d,v20.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:34:2: error: instruction requires: crypto
08:54:17  pmull2 v2.1q,v20.2d,v20.2d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:35:2: error: instruction requires: crypto
08:54:17  pmull v1.1q,v16.1d,v16.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:41:2: error: instruction requires: crypto
08:54:17  pmull v18.1q,v0.1d,v19.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:48:2: error: instruction requires: crypto
08:54:17  pmull v0.1q,v0.1d,v19.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:73:2: error: instruction requires: crypto
08:54:17  pmull v0.1q,v20.1d,v3.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:75:2: error: instruction requires: crypto
08:54:17  pmull2 v2.1q,v20.2d,v3.2d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:76:2: error: instruction requires: crypto
08:54:17  pmull v1.1q,v21.1d,v17.1d
08:54:17  ^
08:54:17 /build/mono-5.15.0.292/mono/btls/build-shared/boringssl/crypto/modes/ghashv8-armx.S:82:2: error: instruction requires: crypto
[...]

PR #8542, opened 2018-05-02
[build] Add a -march forcing crypto extensions on ARM64

directhex:
We only care about CPUs with this extension, and it fixes https://github.com/mono/mono/issues/8488

GCC doesn't object to this flag, so don't bother gating it as Clang-only


luhenry:

luhenry:

Issue #8540, opened 2018-05-02
VS for Mac add-ins cannot be loaded


lambdageek:
## Steps to Reproduce

1. Install VS for Mac (I have Visual Studio Community 2017 for Mac Version 7.4.3 (build 10))
2. Install mono master (Mono JIT compiler version 5.15.0.321 (master/5eafab5dabf Tue May  1 21:53:23 EDT 2018))
3. Open VS Mac

### Current Behavior

![vsmac-loadcontext](https://user-images.githubusercontent.com/480437/39535348-ab82624a-4e01-11e8-802f-9a3f1a573470.png)

Multiple extensions throw a System.IO.FileNotFoundException

### Expected Behavior

All the addins load correctly.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.15.0.321 (master/5eafab5dabf Tue May  1 21:53:23 EDT 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-master/0b3cb8ac12c)
	GC:            sgen (concurrent by default)
```

## Stacktrace

```
ERROR [2018-05-02 12:10:04Z]: Add-in error (Xamarin.AzureSupport,7.4.3): Add-in could not be loaded
System.IO.FileNotFoundException: Invalid Image
File name: '/Applications/Visual Studio.app/Contents/Resources/lib/monodevelop/AddIns/Xamarin.Ide.Identity/System.Net.Http.Extensions.dll'
  at (wrapper managed-to-native) System.Reflection.Assembly.LoadFrom(string,bool)
  at System.Reflection.Assembly.LoadFrom (System.String assemblyFile) [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/master/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Reflection/Assembly.cs:539 
[...]

PR #8547, opened 2018-05-02
[loader] If binding redirection redirects to the same version, ignore it

lambdageek:
If you load Foo.dll and it happens to be version 2.0.0.0 and an app.config says
to redirect all versions of Foo.dll to version 2.0.0.0, we're there already -
don't close the image and probe again, that's silly.

Followup for #8326. This should address #8540, but I want to keep that issue open in case I note more breakage with more testing.


lambdageek:
@monojenkins build OS X x64

PR #8549, opened 2018-05-02
[System]: Wrap `ServicePointScheduler.StartScheduler()` call with `Task.Run()` to avoid a deadlock.

baulig:


mcs/class/System/System.Net/ServicePointScheduler.cs
```diff
@@ -126,7 +126,7 @@ public void Run ()
 		{
 			lock (ServicePoint) {
```
marek-safar: This lock is now not used and can be removed

marek-safar:
ok, closing it

PR #8550, opened 2018-05-03
[Tailcall] Allow tailcall when returning valuetypes.

jaykrell:


mono/mini/mini-ppc.c
```diff
@@ -1346,24 +1346,10 @@ mono_arch_tailcall_supported (MonoCompile *cfg, MonoMethodSignature *caller_sig,
...
 	gboolean res;
-	int i;
 
 	c1 = get_call_info (caller_sig);
 	c2 = get_call_info (callee_sig);
 	res = c1->stack_usage >= c2->stack_usage;
-	if (callee_sig->ret && MONO_TYPE_ISSTRUCT (callee_sig->ret))
-		/* An address on the callee's stack is passed as the first argument */
-		res = FALSE;
-	for (i = 0; i < c2->nargs; ++i) {
```
jaykrell: oops, fixed later;  will repair history

jaykrell:
tailcall/2.exe.timeout is due to

commit 3ec6733f74c46616b61a6c7cdf854d5aef6ab059
Author: Jay Krell <jaykrell@microsoft.com>
Date:   Tue May 1 16:59:45 2018 -0700
    Remove the CallInst.tailcall field, which is not currently accurate
    in the face of llvm_only.
        Provide MONO_IS_TAILCALL which is accurate.
        Handle all tailcall operations in arm32 parameter construction.
        Change all call->tailcall reads to MONO_IS_TAILCALL (&call->inst).

which is surprising.

jaykrell:
The reason 2.exe hung, is because the change from call->tailcall to IS_TAILCALL_OPCODE exposed that tailcall isn't actually written usually until after it is used, usually. So when you'd expect true, it was false. So parameter passing was incorrect. And probably still is for CEE_JMP. I'll write another test for that before fixing it completely.

jaykrell:
@monojenkins build @monojenkins OS X i386

jaykrell:
@monojenkins build OS X i386

mono/mini/method-to-ir.c
```diff
@@ -2559,6 +2562,9 @@ mono_emit_method_call_full (MonoCompile *cfg, MonoMethod *method, MonoMethodSign
 		call->method = method;
 	call->inst.flags |= MONO_INST_HAS_METHOD;
 	call->inst.inst_left = this_ins;
+
+	// FIXME This has already been read in amd64 parameter construction.
+	// Fixing it generates incorrect code. CEE_JMP needs attention.
```
vargaz: Shouldn't this be set earlier then ?

jaykrell: No. That makes it generate incorrect code.

vargaz: Why ? If the arch specific code depends on this flag, then shouldn't it be set before that code is called ?

jaykrell: No. If you set it, it sends the code down incorrect path, generating incorrect code. The checks for it should probably be removed. Or fixed. But I had to test jmp before making any change, which I did later. Jmp sets the flag & gets incorrect code as a result.

jaykrell: The use of the tailcall flag is incorrect. It “worked” because the setting is incorrect too. If you just fix one, it breaks. Either remove both or fix both, or leave it alone. Here I have left it alone, just adding comments. Another change goes further but not this one.

jaykrell: Related to https://github.com/mono/mono/pull/8571

jaykrell: So yes I agree it should be fixed, but 1. For now commenting 2. What is the fix? What is the point of the amd64-specific code? All I can see is it breaks things. And lots works reasonably without it. But perhaps there is a small repair for it, that can make even more work?

jaykrell: Either way I’d like to take this PR asis. It does not regress anything and enables more tailcalls.

jaykrell:
Approve/deny?

mono/mini/method-to-ir.c
```diff
@@ -8433,6 +8439,8 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				emit_tailcall_parameters (cfg, fsig);
 				MONO_INST_NEW_CALL (cfg, call, OP_TAILCALL);
 				call->method = cmethod;
+				// FIXME Other initialization of the tailcall field occurs after
+				// it is used. So this is the only "real" use and needs more attention.
 				call->tailcall = TRUE;
```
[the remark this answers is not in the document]
jaykrell: That's exactly where I was going.
It is exactly what I tried. And it failed.
But now I know more and can do better. But not in this PR.

PR #8551, opened 2018-05-03
[Tailcall] Add a test for passing large valuetypes by value. Should fail. Will fix.

jaykrell:
This is expected to fail, at least on arm64, and then a fix is available.
This is believed not to be a regression.
This isn't likely "properly" fixable. We just cannot allow it.

Passing valuetypes sometimes involves a local temporary passed by address.
Passing locals by address inhibits tailcall.

It is concievable that some cases can be handled by passing on what
we got from our caller -- which isn't guaranteed to exist -- and
can't clearly be used w/o breaking semantics, unless const-analysis is perhaps done.


jaykrell:
Need to understand how this manages to work in some architectures.

jaykrell:
Let's combine this with https://github.com/mono/mono/pull/8550 and only separate if requested.

PR #8552, opened 2018-05-03
[native types] fix nfloat:NaN on arm64 JIT and interp

lewurm:
interp: If one of the operands is NaN, we have a default result: It is false (0)
when doing a compare (`ceq_r8`), but must be true (1) on `cne_r8`.

JIT: match behaviour what C# compiler would generate for float/double.


lewurm:
@monojenkins backport 2018-02

lewurm:
@monojenkins backport 2018-04

Issue #8518, opened 2018-04-30
Support Response Files


jonpryor:
It would be useful if `mono` supported [response files](https://msdn.microsoft.com/en-us/library/windows/desktop/aa367156(v=vs.85).aspx).

A "response file" is a plain-text file that contains strings to be treated as arguments for `mono`.

For example, assume a file `foo.rsp` with the contents:

```
--debug=casts foo.exe
```

Then `mono @foo.rsp` would be treated *as if* the user had instead entered `mono --debug=casts foo.exe`.

Response files are useful for working around [Windows command-line length limits](https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation) (8191 characters on Windows XP and later), which *sounds* like a lot but can quickly disappear when e.g. [providing AOT commands](https://developercommunity.visualstudio.com/content/problem/237753/xamarin-android-aot-complition-does-not-work-visua.html). (Linked-to bug is a Xamarin.Android bug, but an underlying "cause" of the bug is that Xamarin.Android prefers "8.3" names on Windows when invoking `mono` so that paths are shorter, and so we don't need to deal with path quoting. If we could generate a response file *instead*, it might be easier to support paths-with-spaces.)

migueldeicaza:
Agreed, Mono's C# compiler also supports response files, I think that this request is for the runtime itself.

The issue that I have with @foo.rsp is that it will potentially break (albeit this sounds rare) any executable that was named '@foo.rsp'.

Perhaps this shoudl be mono --response foo

lambdageek:
@migueldeicaza Do we already have `mono -- args ...` meaning "stop interpreting arguments after `--`"?

If yes, we could support `mono @foo.rsp` for response files and if someone really names their assembly "@foo.rsp" then can do `mono -- @foo.rsp` to run it.

jaykrell:
$ mono -- foo.exe
Unknown command line option: '--'
Add that too?

migueldeicaza:
We could explore the "Stop interpreting" option "--", but for now, we can just add "--response=FILE"

PR #8560, opened 2018-05-03
[mono] Add support for response files, using --response=FILE

migueldeicaza:
Implements the feature request here:

https://github.com/mono/mono/issues/8518

### DIFF
--- a/man/mono.1
+++ b/man/mono.1
@@ -588,6 +588,11 @@ Controls whether the runtime should attempt to inline (the default),
 or not inline methods invocations
 .ne
 .RE
+\fB--response=FILE\fR
+Provides a response file, this instructs the Mono command to read
+other command line options from the specified file, as if the
+options had been specified on the command line.   Useful when you have
+very long command lines.
 .TP
 \fB--runtime=VERSION\fR
 Mono supports different runtime versions. The version used depends on the program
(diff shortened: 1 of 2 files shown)


mono/mini/driver.c
```diff
@@ -2729,7 +2747,7 @@ mono_parse_options_from (const char *options, int *ref_argc, char **ref_argv [])
 	
 	for (p = options; *p; p++){
 		switch (*p){
-		case ' ': case '\t':
+		case ' ': case '\t': case '\n':
```
jaykrell: Please also \r.
I wrote the file on Windows and copied it to Unix..

mono/mini/driver.c
```diff
@@ -2160,6 +2162,16 @@ mono_main (int argc, char* argv[])
 		} else if (strcmp (argv [i], "--help-handlers") == 0) {
 			mono_runtime_install_custom_handlers_usage ();
 			return 0;
+		} else if (strncmp (argv [i], "--response=", 11) == 0){
```
jaykrell: What if this is the entire string?

man/mono.1
```diff
@@ -588,6 +588,11 @@ Controls whether the runtime should attempt to inline (the default),
 or not inline methods invocations
 .ne
 .RE
+\fB--response=FILE\fR
```
akoeplinger: @migueldeicaza I'd prefer if we made this `@file.rsp` to match the convention set by roslyn/csc and other tools: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/response-file-compiler-option

Issue #8559, opened 2018-05-03
Deprecate and eventually remove the Legacy TLS Implementation


baulig:
We should at some point first deprecate, then remove the old legacy TLS implementation.  Nobody should really be using TLS 1.0 anymore these days and we're also not making any fixes or even just taking bug reports against that old implementation.

Ideally, I would like to deprecate it as soon as possible, then schedule it for removal.

Are there any objections against this?

migueldeicaza:
No objections from me.

baulig:
PR for the first stage of the removal is up: https://github.com/mono/mono/pull/8568.

baulig:
There are a few places where the Legacy TLS Code is still in use:

- There is some code in `Novell.Directory.Ldap`, which should be easy to replace with `SslStream`: https://github.com/mono/mono/blob/master/mcs/class/Novell.Directory.Ldap/Novell.Directory.Ldap/Connection.cs.
- `System.ServiceModel` has its own copy of the entire code, which only seems to be used here: https://github.com/mono/mono/blob/master/mcs/class/System.ServiceModel/System.ServiceModel/ClientCredentialsSecurityTokenManager.cs#L281

I can easily replace the Ldap code, but what should we do with `System.ServiceModel`?

baulig:
Ok, the `Novell.Directory.Ldap` issue was trivial, I've already replaced that with `SslStream`.

marek-safar:
@baulig I looked at ServiceModel some time ago and it was just a copy of old version because of some private/internal code. I am not sure how well that code works anyone so it might be fine to just leave it there for now

PR #8567, opened 2018-05-03
[Mono.Security] Fix session ID handling in legacy TLS provider.

davidben:
The legacy TLS provider echos back the session ID from the ClientHello
to the ServerHello. This is incorrect. Echoing back the session ID means
the server is resuming the session that the client offered. The legacy
provider does not implement session resumption as a server at all, thus
it should unconditionally send an empty value in the ServerHello field,
signaling that it neither resumed the offered session nor issued a new
session ID for future connections.

This bug causes problems for clients which support TLS 1.3 which, in
order to work around different bugs in the ecosystem, involves the
client filling the session ID with a random string. It would
additionally cause problems if a deployment were migrating from a
different, more featureful, stack to Mono. Clients may then attempt to
offer a cached session from the other stack.


davidben:
Oh, I see @baulig has a PR to just remove this code altogether in #8568. If that's the plan, even better! Happy to close this in that case. :-)

baulig:
Yes, the plan is to remove the Legacy TLS Provider as soon as possible.  See also https://github.com/mono/mono/issues/8559.

PR #8570, opened 2018-05-04
Use size_t for gsize if it matches, else C99 uintptr_t. type-based instead of size-based autoconf printf(size_t)

jaykrell:
If C99 zu is not portable enough, then autoconf can find the exact type.
Iu always works on Win64.
Very old Win32 would want plain u.

GINT_TO_POINTER and such need no autoconf, beyond chosing gsize, gssize.

Using size_t provides for an exact match with functions like malloc and pthread_attr_getstack, no warnings.

### DIFF
--- a/mono/eglib/eglib-config.hw
+++ b/mono/eglib/eglib-config.hw
@@ -9,6 +9,7 @@
 #ifdef _MSC_VER
 
 #include <io.h>
+#include <stddef.h>
 
 #define G_GNUC_PRETTY_FUNCTION   __FUNCTION__
 #define G_GNUC_UNUSED            
@@ -17,23 +18,16 @@
 #define G_BREAKPOINT()           __debugbreak()
 #define MAXPATHLEN 242
 
-typedef uintptr_t gsize;
-typedef intptr_t gssize;
+typedef size_t gsize;
+typedef ptrdiff_t gssize;
 typedef int pid_t;
 
 #define G_DIR_SEPARATOR          '\\'
 #define G_DIR_SEPARATOR_S        "\\"
 #define G_SEARCHPATH_SEPARATOR_S ";"
 #define G_SEARCHPATH_SEPARATOR   ';'
-#ifdef _WIN64
-#define G_GSIZE_FORMAT   "I64u" // assuming Microsoft C runtime back to circa year 2000, newer supports "ll"
-#else
-#define G_GSIZE_FORMAT   "u"
-#endif
-#define GPOINTER_TO_INT(ptr)   ((gint)(intptr_t) (ptr))
-#define GPOINTER_TO_UINT(ptr)  ((guint)(intptr_t) (ptr))
-#define GINT_TO_POINTER(v)     ((gpointer)(intptr_t) (v))
-#define GUINT_TO_POINTER(v)    ((gpointer)(intptr_t) (v))
+
+#define G_GSIZE_FORMAT   "Iu"
 
 // https://msdn.microsoft.com/en-us/library/40bbyw78.aspx
 #define STDOUT_FILENO (1)
--- a/mono/eglib/eglib-config.hw
+++ b/mono/eglib/eglib-config.hw
@@ -27,7 +27,13 @@ typedef int pid_t;
 #define G_SEARCHPATH_SEPARATOR_S ";"
 #define G_SEARCHPATH_SEPARATOR   ';'
 
+#ifdef _WIN64
+// Correct for Win32 and Win64 since circa 2000.
 #define G_GSIZE_FORMAT   "Iu"
+#else
+// Correct for Win32 since circa 1992.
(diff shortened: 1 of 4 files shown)


mono/eglib/eglib-config.hw
```diff
@@ -17,23 +18,16 @@
 #define G_BREAKPOINT()           __debugbreak()
 #define MAXPATHLEN 242
 
-typedef uintptr_t gsize;
-typedef intptr_t gssize;
+typedef size_t gsize;
+typedef ptrdiff_t gssize;
```
luhenry: Why not ssize_t?

jaykrell: That is much newer. Same thing, nicer name, granted.

jaykrell:
How committed to C99 compiler and runtime are we? As a minimum? I know, compiler, we are heavily dependent on. This adds a runtime dependency though. I can avoid it. I have the autoconf done & working to chose among u, lu, llu, or I64u, if zu is too new-fangled. Autoconf for zu could be done with sprintf, but not for crossbuild.

mono/eglib/eglib-config.hw
luhenry: Please switch to ssize_t granted it passes CI.

luhenry:
Since, as you pointed out, we didn’t depend on the C99 runtime, I don’t want to introduce a dependency on it. That’s for all the reasons you already gave, portability being the most important one.

mono/eglib/eglib-config.hw
jaykrell: type 1.c

#include <stddef.h>

cl -EP 1.c | findstr  size_t
Microsoft (R) C/C++ Optimizing Compiler Version 19.12.25835 for x86

1.c
    typedef unsigned int     size_t;
    typedef size_t rsize_t;

cl -EP 1.c | findstr  ptrdiff
Microsoft (R) C/C++ Optimizing Compiler Version 19.12.25835 for x86

1.c
    typedef int              ptrdiff_t;

jaykrell:
I will "rewrite" the zu part. I have it done already.  It is "more complicated", but reuses preexisting heavily used mechanism -- autoconf.  To decide among u lu llu I64u. Please withhold judgement til you see it -- you might like it,  you might not.

mono/eglib/eglib-config.hw
jaykrell: So I suggest ptrdiff_t is really preferred for portability.
I tried stdint.h too and it isn't there.

jaykrell:
@monojenkins build failed

jaykrell:
Review? This makes gsize actually match size_t and therefore fix warnings.

jaykrell:
@luhenry ?

jaykrell:
@monojenkins build  Linux x64

jaykrell:

mono/eglib/eglib-config.hw
kumpera: Makes sense to use ptrdiff_t then.

jaykrell:
@monojenkins build Linux x64

jaykrell:
Fixes warnings e.g.:

```

mono-publib.c:27:46: warning: incompatible function pointer types initializing 'gpointer (*)(gpointer, gsize)' (aka 'void *(*)(void *, unsigned int)') with an
      expression of type 'void *(*)(void *, size_t)' (aka 'void *(*)(void *, unsigned long)') [-Wincompatible-function-pointer-types]
        GMemVTable g_mem_vtable = { vtable->malloc, vtable->realloc, vtable->free, vtable->calloc};
                                                    ^~~~~~~~~~~~~~~

mono-publib.c:27:77: warning: incompsatible function pointer types initializing 'gpointer (*)(gsize, gsize)' (aka 'void *(*)(unsigned int, unsigned int)') with an
      expression of type 'void *(*)(size_t, size_t)' (aka 'void *(*)(unsigned long, unsigned long)') [-Wincompatible-function-pointer-types]
        GMemVTable g_mem_vtable = { vtable->malloc, vtable->realloc, vtable->free, vtable->calloc};
 
mono-threads-posix.c:92:43: warning: incompatible pointer types passing 'gsize *const' (aka 'unsigned int *const') to parameter of type 'size_t *'
      (aka 'unsigned long *') [-Wincompatible-pointer-types]
                res = pthread_attr_getstacksize (&attr, stack_size);
                                                        ^~~~~~~~~~

/usr/include/pthread.h:250:85: note: passing argument to parameter here
int pthread_attr_getstacksize(const pthread_attr_t * __restrict, size_t * __restrict);
                                                                                    ^                                                                                                             

marshal.c:4770:72: warning: format specifies type 'unsigned long' but the argument has type 'gsize' (aka 'unsigned int') [-Wformat]
                mono_error_set_out_of_memory (error, "Could not allocate %lu bytes", size);
                                                                         ~~~         ^~~~
                                                                         %u
```

PR #8571, opened 2018-05-04
Add a jmp test, fix AMD64 passing stacked valuetypes, allow AOT.

jaykrell:
I see a bug in the code and this proves it.
The return value is unpredictable due to the bug.
It works on Windows.
Once CI fails, I'll add the fix.

The only other apparent testing of jmp is in coreclr tests and maybe not run.


jaykrell:
The AOT block is deliberate in the code but w/o comments.
ARM32 is flake.

jaykrell:

jaykrell:
Fixes https://github.com/mono/mono/issues/8577.

jaykrell:
Fixes https://github.com/mono/mono/issues/8578.

jaykrell:
I missed that arm64 was still broken, because CI passing, because I disabled it.. will look into. Opened issue in case it goes after this PR.

jaykrell:
The problems here run a bit deeper than I . realized.
CEE_JMP isn't maybe tailcall, it is definitely tailcall.
The signatures have to match (we don't check), which helps a lot.
In particular, incoming valuetype temporaries should be used for outgoing valuetypes.
tailcall should do that too, when they are available (there, signatures do not have to match).

jaykrell:
It is very close, agreed.


Here is my understanding.
The signature is supposed to match. In that retrospect, it is easier. We don't check that.
It is not followed by a ret. Fallback is therefore slightly more difficult -- have to fake up that ret.
There is no clear provision in the spec for not tailcalling.
In CEE_JMP, the parameters are implied in ldarg 0..n. The evaluation stack is implied empty (or ignored?)
In CEE_TAIL + CEE_CALL, the parameters are on the evaluation stack.
There is clear provision in CEE_TAL + CEE_CALL that you can fail to tailcall, and just call.
Though clearly you are supposed to try hard --  the spec says to document when you do not.



Now, in our implementation, CEE_JMP is a bit different, but shares code.
It hardcodes tail as TRUE, and fails to consider virtual or generics, etc.
I have to check if desktop handles those.


I would suggest sharing more code between CEE_JMP and tailcall.
Exact proposals are quite wordy.

jaykrell:
I've been deliberately holding back here as there are actually a few problems with CEE_JMP, and some more to be confirmed. This fixes only  some.

Issue #8587, opened 2018-05-05
configure.ac G_HAVE_ISO_VARARGS problems


jaykrell:
In configure.ac we have:

```
AC_MSG_CHECKING(for varargs macros)
AC_TRY_COMPILE([],[
int foo (int i, int j);
#define bar(...) foo (1, __VA_ARGS__)
void main () {
	 bar (2);
}
],have_iso_varargs=yes,have_iso_varargs=no)
AC_MSG_RESULT($have_iso_varargs)
G_HAVE_ISO_VARARGS=0
if test "x$have_iso_varargs" = "xyes"; then
   G_HAVE_ISO_VARARGS=1
fi
AC_SUBST(G_HAVE_ISO_VARARGS)
```

This does not work in the way intended.
A main() wrapper is supplied and it tries to compile:

```c
int
main ()
{
		int foo (int i, int j);
		#define bar(...) foo (1, __VA_ARGS__)
		void main () {
			 bar (2);
		}
  ;
  return 0;
}

```

giving:

```
1.c:7:14: error: function definition is not allowed here
void main () {
```

marek-safar:
From @jaykrell comment


Sometimes gcc allows nested functions, sometimes not, i.e. Apple configuration.

Solution is use newer-fangled AC_COMPILE_IFELSE.
Or rename main and put it in the first parameter.
Or remove the entire thing, really, and see what @luhenry did here:

```
#if defined(_MSC_VER) && (_MSC_VER < 1910)
#define g_assertf(x, format, ...) (G_LIKELY((x)) ? 1 : (g_assertion_message ("* Assertion at %s:%d, condition `%s' not met, function:%s, " format "\n", __FILE__, __LINE__, #x, __func__, __VA_ARGS__), 0))
#else
#define g_assertf(x, format, ...) (G_LIKELY((x)) ? 1 : (g_assertion_message ("* Assertion at %s:%d, condition `%s' not met, function:%s, " format "\n", __FILE__, __LINE__, #x, __func__, ##__VA_ARGS__), 0))
#endif
```

Also I don't think the intermediate have_iso_varargs is needed.
Also in the  config.h you can just say `#if @foo@`,  you don't need `#if @foo@ == 1`.

jaykrell:
Or if the parameters end up adjacent, unlike g_assertf,  just consider "format" part of "..." and just say __VA_ARGS__ to take both, with whatever comma the user put in.
[...]

PR #8589, opened 2018-05-05
Cleanup varargs -- 3 variations to 1 and no malfunctioning autoconf.

jaykrell:
Change 3 ifdef versions to one portable one.
Remove supporting autoconf that did not work as intended.
See https://github.com/mono/mono/issues/8587.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4144,21 +4144,6 @@ if test $found_export_dynamic = no; then
 	LDFLAGS="${old_ldflags}"
 fi
 
-AC_MSG_CHECKING(for varargs macros)
-AC_TRY_COMPILE([],[
-int foo (int i, int j);
-#define bar(...) foo (1, __VA_ARGS__)
-void main () {
-	 bar (2);
-}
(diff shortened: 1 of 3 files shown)


luhenry:
How is that an improvement? You removed the check to figure out if there is a functioning __VA_ARGS__ and provide no explanation why it should work.

jaykrell:
The autoconf does not work. I explained it in the bug.

jaykrell:
The third non-macro is also portable, i.e. to C89. And a decent compiler will inline it anyway.
It seems not worthwhile having 3 forms and a broken autoconf.
The reason the autoconf didn't work is because it was accidentally using nested functions, which only some configurations of gcc and maybe clang support. They generate code on the stack which more secure platforms (Mac) don't allow, so Apple compilers don't support it.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
Approve?

jaykrell:
Approve?

PR #8591, opened 2018-05-05
Fill in some missing tailcall pieces: x86reg, armreg, armmembase, align x86patch, x86aot, armhardfloat

jaykrell:


jaykrell:
Interesting this breaks arm.. probably a problem with the extra parameter...

mono/mini/mini-x86.c
```diff
@@ -3223,16 +3244,15 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
-				// FIXME alignment
-				mono_add_patch_info (cfg, code - cfg->native_code, MONO_PATCH_INFO_METHOD_JUMP, call->method);
+				// FIXME Patch data instead of code.
+				code = x86_align_and_patch (cfg, code, MONO_PATCH_INFO_METHOD_JUMP, call->method, 1);
 				x86_jump32 (code, 0);
 			}
 
 			ins->flags |= MONO_INST_GC_CALLSITE;
-			cfg->disable_aot = TRUE;
 			break;
```
vargaz: Where is the aot support above ?

vargaz:
Looks ok otherwise.

mono/mini/mini-x86.c
```diff
@@ -1925,30 +1925,43 @@ if (ins->inst_true_bb->native_offset) { \
 	x86_fnstsw (code); \
 } while (0); 
 
-
 static guint8*
-emit_call (MonoCompile *cfg, guint8 *code, guint32 patch_type, gconstpointer data)
+x86_align_for_patch (MonoCompile *cfg, guint8 *code, guint32 patch_type, gconstpointer data, int align_offset)
 {
-	gboolean needs_paddings = TRUE;
```
vargaz: These changes don't seem to be neccessary.

jaykrell: x86_align_and_patch is reused.
x86_align_for_patch granted only has one use and could be put back inline.
Would that suffice?
I do also like "short circuit with return" instead of updating local booleans but that is style.

This line:
`	guint32 pad_size = (guint32)(code + align_offset - cfg->native_code) & 0x3;
`
confuses me some.
Why the subtract of native_code?
Is it not aligned, but later one there will be some aligned base?
That would explain it.

mono/mini/mini-x86.c
jaykrell: Oops good point.

jaykrell: Arguably it is in x86_align_and_patch but that isn't clear. Let me look again.

jaykrell: Ok, I'm convinced, it is x86_align_and_patch.
Perhaps it would be clearer, if emit_call was refactored slightly, to accept a boolean tailcall? Then call and tailcall would share code more clearly?

jaykrell:
@monojenkins build failed

jaykrell:
I'm inclined to dial some of method-to-ir.c back, or make it aot or llvm-dependent, see if the Linux x64 AOT+LLVM failures drop back to 8 or 9.

jaykrell:
(but let's see how armv5/7 fair)

vargaz:
Can we split the stable parts into another PR/PRs ?

jaykrell:
Almost always possible, just takes a little time.

1. Linux x64 AOT LLVM regression seems to be in master though, now that I look more.

2. The arm32 reg/membase is a little wierd, in that, there is kinda presently no useful use. Just enabling it, exposed more badness in method-to-ir. The hardfloat/x86 stuff is more innocent/useful though. It feels kinda "rightly completionist" to fill it in, make arm32 look complete relative to x86/amd64/arm64, but because of how rgctx/imt are passed, might never be useful.

A compromise is to leave in the cases but omit the defines so they don't trigger.

And/or get method-to-ir.c straightened out -- again, eager to see the latest CI there.


3.  Judge again the x86 changes? Ok asis? Inline single-use function?  AOT correct but too unclear? (really need x86 AOT lanes).

jaykrell:
Need to look arm32 more..probably on the right track with my method-to-ir.c change but missed something obvious.

jaykrell:
Problem on arm is need to consider "is_interface".

jaykrell:

jaykrell:
@monojenkins build ARMv5

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build ARMv5

jaykrell:
Wierd ARMv7 passes but not ARMv5..

jaykrell:
@monojenkins build ARMv5

jaykrell:

jaykrell:
CI all green now. Approved? Denied? Split up?

mono/mini/method-to-ir.c
```diff
@@ -9281,22 +9332,28 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 					if (fsig->generic_param_count) {
 						/* virtual generic call */
 						g_assert (!imt_arg);
+						g_assert (will_have_imt_arg);
 						/* Same as the virtual generic case above */
 						imt_arg = emit_get_rgctx_method (cfg, context_used,
 														 cmethod, MONO_RGCTX_INFO_METHOD);
+						g_assert (imt_arg);
```
vargaz: Can this call ever return null ?

jaykrell: Probably not. I added this implicit “pointer is/was boolean” notion so more sensitive to it.

mono/mini/mini-arm.c
```diff
@@ -5116,23 +5135,55 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
+			// For reg and membase, get destination in IP.
+
+			if (tailcall_reg) {
+				g_assert (ins->sreg1 > -1);
+				if (ins->sreg1 != ARMREG_IP)
+					ARM_MOV_REG_REG (code, ARMREG_IP, ins->sreg1);
+			} else if (tailcall_membase) {
+				g_assert (ins->sreg1 > -1);
+				if (!arm_is_imm12 (ins->inst_offset)) {
+					g_assert (ins->sreg1 != ARMREG_IP); //. temp in emit_big_add
```
jaykrell: Mac keyboard keeps inserting random spaces and then Mac inserts dots. I'll fix the dot setting at least..

PR #8592, opened 2018-05-05
[wasm] Add .wasm mime mapping for python3 in server.py too

akoeplinger:
It was only done in the python2 case. Noticed this while answering a contributor question on gitter.


kumpera:
you guys... 💖

PR #8599, opened 2018-05-06
[Coop] Convert Mono.Runtime

jaykrell:
Trivial.


jaykrell:
Linux x64 FullAOT+LLVM regressed from 8 or 9, to 44, but obviously not from this change.

PR #8605, opened 2018-05-07
[Coop] Convert System.ArgIterator

jaykrell:
Trivial.


vargaz:
Since the HANDLES () code has a certain amount of overhead, it would be better to add a NOHANDLES () macro or something, and don't convert these icalls to handles.

jaykrell:
This stuff is frequently used?
Isn't the overhead just:
  1. stack space for MonoError
  2. one instruction to initialize the MonoError
  3. check the monoerror upon exist -- instead of or in addition to thread's pending exception?

Too much?

jaykrell:
I suggest this overhead is affordable for the vast majority of icalls, that don't traffic in pointers.
But identifying the exact set is not scientific, granted.
Perhaps a scientific policy would be to exclude ones that have no pointers and no errors?

vargaz:
There are two calls made to c functions as well, i.e. mono_icall_begin () and mono_icall_end ().

jaykrell:
Fails miserably, come back to it later..

PR #8607, opened 2018-05-07
[Coop] Convert System.Decimal

jaykrell:
Fairly trivial.


vargaz:
Esp. some of these ones need to remain non-handle to avoid overhead.

jaykrell:
I'm told we don't like mono_error_set_pending_exception.
And removal of mono_error_set_pending_exception is conflated with icall-def.h HANDLES().
System.Decimal is probably little used, though when used perhaps performance sensitive.
Perhaps we should plan to rewrite this in C#, or corefx already has?
Or perhaps we don't dislike mono_error_set_pending_exception enough to really remove it.
It is much less bad than raw pointers?

marek-safar:
yes, this code will migrate to managed once I have someone available to do it

jaykrell:
Hold off on this.

PR #8609, opened 2018-05-07
[jit x86] Move microoptimization down a level: x86_mov_reg_reg do not…

jaykrell:
…hing if source == desta and size == 4 (size is always 4, except in amd64 reuse,
which was only used once, unnecessarily, remove it).

[jit x86] x86_mov_reg_reg size is always 4,
split into unused. x86_mov_reg_reg_size and specialized x86_mov_reg_reg


vargaz:
Not a fan of this, our convention is that the codegen macros simply generate code and optimizations are done by the caller.

jaykrell:
Why not optimize here?

Does caller need/want to know exact byte count or instruction count?
Or require "inst" parameter always be evaluated? (more than once)?

Byte count kinda varies. Kinda -- because of the size==2 path, which is unused.
Would optimization here really surprise or hurt any caller?

mono/mini/exceptions-x86.c
```diff
@@ -92,7 +92,7 @@ mono_win32_get_handle_stackoverflow (void)
 	x86_mov_reg_membase (code, X86_EBX, X86_ESP, 4, 4);
 
 	/* move current stack into edi for later restore */
-	x86_mov_reg_reg (code, X86_EDI, X86_ESP, 4);
+	x86_mov_reg_reg (code, X86_EDI, X86_ESP);
```
vargaz: This looks ok. The commit message says 'x86_mov_membase_reg' through.

jaykrell: True. I was looking at that too. I'd fix the squash/commit-to-master message.
Please clarify:
1. I have convinced you of the overall PR and it is approved?
or 2. Just removing the usually/always==4 last parameter?

jaykrell:
@monojenkins build OS X x64

PR #8619, opened 2018-05-08
[testing] Replace gensources.sh with gensources.cs

kg:
This PR introduces a new tool, gensources.cs, which replaces gensources.sh. It is able to parse the whole set of sources files for a library in one go so that genproj can use that information to encode all our platform and profile specific files into one csproj file.

For now this PR just introduces it and switches to using it instead of gensources.sh. Making the PR in order to see whether any builds or tests fail (it works locally).


mcs/build/library.make
```diff
@@ -295,24 +295,37 @@ endif
...
+	$(with_mono_path_monolite) $(RUNTIME) $(monolite_path)/mcs.exe -debug -out:$(gensources) $(topdir)/build/gensources.cs
 
 # Note, gensources.sh can create a $(sourcefile).makefrag if it sees any '#include's
 # We don't include it in the dependencies since it isn't always created
-sourcefile = $(depsdir)/$(PROFILE_PLATFORM)_$(PROFILE)_$(LIBRARY_SUBDIR)_$(LIBRARY).sources
-$(sourcefile): $(PROFILE_sources) $(PROFILE_excludes) $(topdir)/build/gensources.sh $(depsdir)/.stamp
-	$(SHELL) $(topdir)/build/gensources.sh $@ '$(PROFILE_sources)' '$(PROFILE_excludes)'
+sourcefile = $(depsdir)/$(SOURCES_PLATFORM)_$(PROFILE)_$(LIBRARY_SUBDIR)_$(LIBRARY).sources
+$(sourcefile): $(PROFILE_sources) $(PROFILE_excludes) $(topdir)/build/gensources.sh $(depsdir)/.stamp $(gensources)
+	$(SHELL) $(topdir)/build/gensources.sh "$@.shout" '$(PROFILE_sources)' '$(PROFILE_excludes)'
```
luhenry: Why are we still using both `gensources.sh` and `gensources.exe`? I understand it's useful during development to guarantee it generates the same thing, but once you verified it, you should not commit both. Otherwise, you should add a `diff "/tmp/$(LIBRARY).csgensources" "/tmp/$(LIBRARY).shgensources"` and guarantee it's the same.

mcs/build/library.make
```diff
@@ -295,24 +295,34 @@ endif
...
+PROFILE_excludes = $(firstword $(if $(SOURCES_PLATFORM),$(wildcard $(SOURCES_PLATFORM)_$(PROFILE)_$(LIBRARY).exclude.sources)) $(wildcard $(PROFILE)_$(LIBRARY).exclude.sources))
 
-# Note, gensources.sh can create a $(sourcefile).makefrag if it sees any '#include's
-# We don't include it in the dependencies since it isn't always created
-sourcefile = $(depsdir)/$(PROFILE_PLATFORM)_$(PROFILE)_$(LIBRARY_SUBDIR)_$(LIBRARY).sources
-$(sourcefile): $(PROFILE_sources) $(PROFILE_excludes) $(topdir)/build/gensources.sh $(depsdir)/.stamp
-	$(SHELL) $(topdir)/build/gensources.sh $@ '$(PROFILE_sources)' '$(PROFILE_excludes)'
+gensources = $(topdir)/build/gensources.exe
+$(gensources): $(topdir)/build/gensources.cs
+	echo $(BOOTSTRAP_MCS) -noconfig -debug -r:mscorlib.dll -r:System.dll -r:System.Core.dll -out:$(gensources) $(topdir)/build/gensources.cs
```
luhenry: There is some debugging leftover. FYI, you can use `make --trace` to see exactly which commands are getting executed, even without `V=1`.

mcs/build/profiles/basic.make
```diff
@@ -118,4 +118,4 @@ $(PROFILE_EXE): $(topdir)/build/common/basic-profile-check.cs
 	$(BOOTSTRAP_MCS) /warn:0 /noconfig /langversion:latest /r:System.dll /r:mscorlib.dll /out:$@ $<
 
 $(PROFILE_OUT): $(PROFILE_EXE)
-	$(PROFILE_RUNTIME) $< > $@ 2>&1
+	$(PROFILE_RUNTIME) $< > $@ 2>&1
```
luhenry: Line damage

mcs/build/library.make
akoeplinger: he's talking about the `echo`. that one is already printed when building with V=1

luhenry:

akoeplinger:
@luhenry please don't do the apidiff commit, it shouldn't show up (Katelyn is looking at it).

kg:
@monojenkins build failed

mcs/build/gensources.cs
```diff
@@ -0,0 +1,411 @@
...
+                    break;
+                case "--trace2":
+                    SourcesParser.TraceLevel = 2;
+                    break;
+                case "--trace3":
+                    SourcesParser.TraceLevel = 3;
+                    break;
+                case "--trace4":
+                    SourcesParser.TraceLevel = 4;
+                    break;
```
akoeplinger: not sure we really need four logging levels for this :) I'd consolidate it into a single `--verbose`

mcs/build/gensources.cs
```diff
@@ -0,0 +1,411 @@
...
+                    Console.Error.WriteLine ($"// {entry.Directory} / {entry.Pattern} -> {absoluteDirectory} / {absolutePattern}");
+            }            
+
+            if (!Directory.Exists (absoluteDirectory)) {
+                Console.Error.WriteLine ($"Directory does not exist: {Path.GetFullPath (absoluteDirectory)}");
+                ErrorCount += 1;
+                continue;
+            }
+
+            var matchingFiles = Directory.GetFiles (absoluteDirectory, absolutePattern);
```
akoeplinger: We should only call `.GetFiles()` when we actually have a `*` pattern, it is quite expensive to do it for each file otherwise. This would save 100ms when generating net_4_x corlib.

mcs/build/gensources.cs
```diff
@@ -0,0 +1,411 @@
+using System;
```
akoeplinger: please add the usual license header.

kg:

kg:

PR #8620, opened 2018-05-08
[sdks] Transition for storing the llvm version info in llvm/SUBMODULE…

vargaz:
…S.json and using the makefiles/scripts there to clone the repo etc.


vargaz:
There is a dependency problem since llvm/Makefile is created by configure, and configure needs an already built llvm. So we will probably need to make llvm/Makefile a normal makefile.

luhenry:
I'm fine with making `llvm/Makefile` a normal Makefile since there are no sources or anything in there other than the logic to fetch llvm and build it.

vargaz:
So making llvm/Makefile a normal makefile is a little bit messy, it might be a better approach to get rid of it completely, and move the logic there to the sdk makefiles which are already normal makefiles.

luhenry:
@vargaz you are saying "configure needs an already built llvm", why is that? And instead of having these dependencies at configure-time, could we have them at compile-time instead?

vargaz:
configure needs to run the llvm-config script to obtain the list of include dirs, libraries, gcc flags, etc. Running it at compile time would be kinda complicated with automake.

luhenry:
Ok, let's get it to depend on `sdks/` then. I don't have a strong opinion on which way we should go, so it's up to you.

sdks/builds/llvm.mk
```diff
@@ -1,9 +1,16 @@
 
 LLVM_SRC?=$(TOP)/sdks/builds/toolchains/llvm
 
-$(TOP)/sdks/builds/toolchains/llvm:
```
luhenry: Why don't we just keep the existing code since it does the same thing?

vargaz: The code in versions.mk reads the hashes from SUBMODULES.json and makes sure that the proper version is checked out if the hash changes. Its also used by acceptance-tests/ so there is code sharing.

vargaz:
Superseded by https://github.com/mono/mono/pull/8792

PR #8621, opened 2018-05-08
[Tailcall] Classify some CoreCLR test problems, enable more, partially implement tail.calli for them.

jaykrell:


jaykrell:

jaykrell:

jaykrell:
Finally passing CI. Review?

mono/mini/method-to-ir.c
```diff
@@ -2679,6 +2679,7 @@ mono_emit_method_call (MonoCompile *cfg, MonoMethod *method, MonoInst **args, Mo
 MonoInst*
 mono_emit_native_call (MonoCompile *cfg, gconstpointer func, MonoMethodSignature *sig,
 					   MonoInst **args)
+// FIXME tailcall?
 {
```
vargaz: These are never going to be tailcalls.

mono/mini/method-to-ir.c
```diff
@@ -2694,6 +2695,7 @@ mono_emit_native_call (MonoCompile *cfg, gconstpointer func, MonoMethodSignature
 
 MonoInst*
 mono_emit_jit_icall (MonoCompile *cfg, gconstpointer func, MonoInst **args)
+// FIXME tailcall?
 {
```
vargaz: Ditto.

mono/mini/method-to-ir.c
```diff
@@ -2707,9 +2709,10 @@ mono_emit_jit_icall (MonoCompile *cfg, gconstpointer func, MonoInst **args)
  *
  *   Emit a call to the runtime function described by PATCH_TYPE and DATA.
  */
-inline static MonoInst*
+static MonoInst*
 mono_emit_abs_call (MonoCompile *cfg, MonoJumpInfoType patch_type, gconstpointer data, 
 					MonoMethodSignature *sig, MonoInst **args)
+// FIXME tailcall?
 {
```
vargaz: Ditto.

mono/mini/method-to-ir.c
```diff
@@ -8560,24 +8583,43 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
-			ins = (MonoInst*)mini_emit_calli (cfg, fsig, sp, addr, NULL, NULL);
+			ins = (MonoInst*)mini_emit_calli_full (cfg, fsig, sp, addr, NULL, NULL, tailcall_calli);
 
 			calli_end:
 			if (ins_flag & MONO_INST_TAILCALL)
-				test_tailcall (cfg, FALSE); // Currently we never tailcall calli.
+				test_tailcall (cfg, tailcall_calli);
+
+			// Tailcall does not return, and the subsequent code confuses the JIT, and
+			// can usually be optimized away. However note that valid code can branch
```
vargaz: What is the problem with having code past the tailcall ? For the rest of the JIT, a tailcall looks just like a normal call.

mono/mini/method-to-ir.c
jaykrell: Because we have to save lmf?

mono/mini/method-to-ir.c
jaykrell: Because we have to save lmf?  I can remove/reduce the comments.

mono/mini/method-to-ir.c
jaykrell: Whenever I don't do this, I get an assertion failure. A register is -1.
Perhaps the problem is when only doing half the work though.

jaykrell: I  could just reduce the commentary and leave it looking like a nice little optimization. I can reproduce the assertion failure.

jaykrell: Actually..they don't entirely go together. There is a reason to have two variables.
I am missing the test case for why they are two variables -- because you can branch around the call to ret. Anyway, there was always an assertion failure when I didn't "finish" this, that I didn't try to fix/understand otherwise.

jaykrell: I meant -- I can try to reproduce the failure.
It was possibly
				g_assert (vreg != -1);
in
mono_handle_global_vregs but I'm guessing from memory.

vargaz: One solution is to just generate the same code, i.e. dont skip ret, etc. It will lead to some extra code generated, but keep the JIT simpler.

mono/mini/method-to-ir.c
vargaz: Because these types of calls are generated by the JIT, and we don't use tail calls ourselves.

mono/mini/method-to-ir.c
jaykrell: I'm pretty sure if I do nothing, leave the push and the ret, it failed an assert. I'd have to disable  the code and  rerun the test to be sure. It is a little wierd when you look at the generated code and see the obviously dead code, but maybe only 3 or so of us look and maybe only 1 thinks it wierd.

The optimization is based on regular call/tailcall of course, I didn't come up with it.

jaykrell: I'll try it. Not worth so much talk.

jaykrell: Wild guess while I wait for tests..
```
mono_emit_call_args
	if (tailcall) {
...
	} else if (!MONO_TYPE_IS_VOID (sig_ret))
		call->inst.dreg = alloc_dreg (cfg, (MonoStackType)call->inst.type);
```

and then you ret after that, probably not happy.. so, it is fixable, if you want tailcalls to look more like calls. They'd still be saving stack which is the point.

I also suspect the branch around call doesn't work.
Though maybe only allowed for void functions??
i.e. I don't know what is verifiable, or ECMA allows for verification to be easier.

jaykrell: I'm not seeing the failure now with a little local testing.
Want me to remove all  skip_ret/push_reslogic??
It does seem suspicious if removed.
From call/callvirt too?

Another option, really, is to share more code between calli and regular  call/callvirt.
How about that? I meant to in the first place but waffled.  Let me try..

vargaz:
CALLI used to be handled the same as the other calls, but was moved to its own code section to simplify the code.

mono/mini/method-to-ir.c
vargaz: I think we should try to remove that logic if possible, it will lead to slightly bigger generated code, but the JIT code will be simpler and less error prone.

jaykrell:
Well, I'm not suggesting merge the entire thing, but just the ending. I'll try and we can judge..

mono/mini/method-to-ir.c
jaykrell: I don't think this is all that complicated, but it is subjective.
Let me try to have one copy instead of two, and then we can consider zero or one.
Two is harder to justify than one or zero.

jaykrell:
ok and will cleanup shortly

PR #8631, opened 2018-05-08
Bump msbuild to track mono-2018-02

radical:


akoeplinger:

PR #8633, opened 2018-05-08
[jit] Fix support for directly calling icalls.

vargaz:


mono/metadata/marshal.h
```diff
@@ -259,7 +260,7 @@ typedef enum {
 } MonoStelemrefKind;
 
 
-#define MONO_MARSHAL_CALLBACKS_VERSION 2
+#define MONO_MARSHAL_CALLBACKS_VERSION 3
```
jaykrell: Maybe comment the version history? Maybe.

jaykrell: (almost nothing is worth reseting CI imho though)

vargaz:
Still causes some regressions.

PR #8641, opened 2018-05-09
Fix powerpc and s390x compilation.

jaykrell:


lewurm:

luhenry:
@lewurm there are quite a lot of tests failures, so unsure which one are meaningful and which one are not.

akoeplinger:
I'm pretty sure they are meaningful, e.g. this one for PPC:

```
Test run: image=/home/builder/jenkins/workspace/test-mono-pull-request-ppc64el/mono/mini/iltests.exe, opts=
Opcode 'tailcall_parameter' missing from machine description file.
Stacktrace:

  at <unknown> <0xffffffff>
  at Tests.test_3_tail_call_vret_by_addr () [0x00010] in <e556237dcffe49ad868d98c343b249fd>:0
```

... but given that PPC failed to even _compile_ before it is progress 😄

akoeplinger:
@jaykrell please file an issue for tracking the above failure if it's not something that can be fixed quickly.

lewurm:
@luhenry runtime build, bcl build and mini regression test suite should be green, which is the case with #8645 👍

PR #8645, opened 2018-05-09
Fix PowerPC64 not just PowerPC32.

jaykrell:


akoeplinger:
do we need the same fix for cpu-s390x.md?

jaykrell:
S390x same or similar yes

jaykrell:
I’ll get it soon today.

akoeplinger:
Ok, looks like PPC is back to where it was before the break so let's get this in ;)

PR #8667, opened 2018-05-10
Some method-to-ir cleanup.

jaykrell:

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -6301,7 +6301,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 	MonoBasicBlock *tblock = NULL;
 	MonoBasicBlock *init_localsbb = NULL, *init_localsbb2 = NULL;
 	MonoSimpleBasicBlock *bb = NULL, *original_bb = NULL;
-	MonoMethod *cmethod, *method_definition;
+	MonoMethod *method_definition;
 	MonoInst **arg_array;
 	MonoMethodHeader *header;
 	MonoImage *image;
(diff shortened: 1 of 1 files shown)


mono/mini/method-to-ir.c
```diff
@@ -8592,72 +8609,26 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				}
 			}
 			ins = (MonoInst*)mini_emit_calli_full (cfg, fsig, sp, addr, NULL, NULL, tailcall);
-
```
vargaz: I don't think we should goto between switch cases, its really confusing. Just keep the duplicate code.

jaykrell: I think it is too much duplicated code. It was less before, and it has incorrectly diverged even then. Now it is larger, though maybe could be dialed back.

Ultimately I think the common tail becomes a function and then can call it from two places.

mono/mini/method-to-ir.c
```diff
@@ -8583,87 +8598,38 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
 				} else if (info_type == MONO_PATCH_INFO_JIT_ICALL_ADDR) {
-					// FIXME tailcall?
-					tailcall_calli = FALSE;
+					tailcall = FALSE;
 					ins = (MonoInst*)mono_emit_abs_call (cfg, info_type, info_data, fsig, sp);
 					NULLIFY_INS (addr);
 					goto calli_end;
 				}
 			}
-			ins = (MonoInst*)mini_emit_calli_full (cfg, fsig, sp, addr, NULL, NULL, tailcall_calli);
```
jaykrell: A whole screenful of duplication (for me), and it was missing the sequence point logic.

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build OS X x64

mono/mini/method-to-ir.c
jaykrell: I don't want this duplication. I can try writing a function for it and call it from two places.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build Linux i386

jaykrell:

PR #8669, opened 2018-05-11
"On" field breaks compatibility with System.Net.Http because of "Internals Visible To" & member type mis-match

TonyValenti:
Hi Mono Team,
This commit fixes an interesting bug that relates to the  System.Net.Http nuget packages.

So the short story is that System.Net.Http can view the internals of this assembly because of InternalsVisibleTo.  Part of HttpClient references System.Net.Logging's "On" _property_.  The problem is that mono has implemented "On" as a _field_ . When the latest version of System.Net.Http.HttpClient tries to invoke a HttpClient.SyncAsync using Mono, you get a MissingMethodException because:
bool System.Net.Logging.get_On()
does not exist.

This commit is simple and only changes the signature of On to match that of the reference source (property instead of field).  This will cause the Missing Method Exception not to be thrown.

### DIFF
--- a/mcs/class/System/ReferenceSources/Logging.cs
+++ b/mcs/class/System/ReferenceSources/Logging.cs
@@ -4,14 +4,13 @@ namespace System.Net {
 	static class Logging
 	{
 
-        internal static bool On {
-            get {
-                return false;
-            }
-        }
+		internal static bool On {
+			get {
+				return false;
+			    }
+		}
 
-
-        internal static TraceSource Web {
+		internal static TraceSource Web {
 			get {
 				return null;
 			}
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

TonyValenti:
@akoeplinger @marek-safar 
Hi Guys,
Thank you so much for approving and accepting this change.  What is the process I need to go through in order to get a build that incorporates these changes and is used by Visual Studio for Mac?  Is it as simple as installing tonight's nightly build?

marek-safar:
Yes nightly build or you could also download the package directly from the commit on https://github.com/mono/mono/commits/master page, look for the green tick next to the commit message

PR #8670, opened 2018-05-11
Cleanup IL size checks.

jaykrell:
Make it much more data driven and remove about 200 lines of code.


jaykrell:
Still working on it.

PR #8673, opened 2018-05-11
Fix sizing of cee_switch, mono_dyn_call, mono_icall.

jaykrell:


mono/cil/cil-opcodes.xml
```diff
@@ -292,7 +292,7 @@
 <opcode name="unused70" input="Pop0" output="Push0" args="InlineNone" o1="0xFE" o2="0x22" flow="next" />
```
jaykrell: Might want to change this file. Or rename it away. Or make sure its timestamp is old. I edited it second, oops.

jaykrell: Looks like timestamps are not controllable in git. If this causes problems we can edit it away in Makefile.am and add comments that the file was generated but maybe no longer is.
Or we could always generate it..

jaykrell:
@monojenkins build  Linux x64

jaykrell:

PR #8687, opened 2018-05-11
Backport "Implement Freeform Unmanaged Error" to 2018-02

alexanderkyte:


mono/eglib/goutput.c
```diff
@@ -126,12 +126,23 @@ g_log (const gchar *log_domain, GLogLevelFlags log_level, const gchar *format, .
...
+}
+
 void
 g_assertion_message (const gchar *format, ...)
 {
 	va_list args;
 
 	va_start (args, format);
+
+	g_vasprintf (&failure_assertion, format, args);
```
jaykrell: This breaks assert. Put va_start after it. Or sprintf into a local and use it for both (g_log ("%s", foo)).

mono/eglib/glib.h
```diff
@@ -576,6 +576,8 @@ GLogLevelFlags g_log_set_fatal_mask   (const gchar *log_domain, GLogLevelFlags f
 void           g_logv                 (const gchar *log_domain, GLogLevelFlags log_level, const gchar *format, va_list args);
 void           g_log                  (const gchar *log_domain, GLogLevelFlags log_level, const gchar *format, ...);
 void           g_assertion_message    (const gchar *format, ...) G_GNUC_NORETURN;
+void           g_assertion_message    (const gchar *format, ...) G_GNUC_NORETURN;
```
jaykrell: No need to declare twice.

mono/eglib/goutput.c
jaykrell: https://github.com/mono/mono/pull/8727/commits/a1baf487d073f6d66d11ac0134f2fa6b0e06408e fixes this.

PR #8690, opened 2018-05-12
[jit] Implement devirtualization of some EqualityComparer<T>.Equals calls, the corefx code assumes this.

vargaz:


joncham:
What does `The corefx code excepts these calls to be devirtualized.` mean? That the behavior is wrong without devirtualization? Or just that code assumes `EqualityComparer<T>.Equals` is fast?

vargaz:
It assumes its fast, it calls it in inner loops etc.

vargaz:
https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Collections/Generic/Dictionary.cs#L291

joncham:
Thanks!

migueldeicaza:
I do not understand @kumpera 's comment.   It seems like this is a way of attaching a meaning that is only handled at a later stage.   What is it that you are suggesting?

vargaz:

vargaz:

PR #8717, opened 2018-05-15
Use guchar instead of unsigned char for brevity/style.

jaykrell:


jaykrell:

PR #8718, opened 2018-05-15
method_to_ir: factor out operand decoding and  stack checking

jaykrell:


mono/dis/push-pop.h
```diff
@@ -1,3 +1,5 @@
+#error This file is not used.
```
kumpera: just remove it?

mono/mini/method-to-ir.c
```diff
@@ -6897,7 +7017,25 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 		if (cfg->verbose_level > 3)
 			printf ("converting (in B%d: stack: %d) %s", cfg->cbb->block_num, (int)(sp - stack_start), mono_disasm_code_one (NULL, method, ip, NULL));
 
-		switch (*ip) {
+		MonoOpcodeParameter parameter;
+		const MonoOpcodeInfo* info = mono_opcode_decode (ip, op_size, il_op, &parameter);
```
kumpera: Could you benchmark if compilation times significantly change because of this?
Something as simple as compiling the whole of corlib (--compile-all) should do it.
This was a perf issue I hit in the past.

jaykrell: I doubt it but yes I will.
There is an unfortunate double  array lookup that could be brought back to one (was zero).

mono/dis/push-pop.h
jaykrell: It demonstrates some nice patterns..could  also not change  it..

jaykrell:

jaykrell:
@monojenkins build OS X x64

mono/mini/method-to-ir.c
jaykrell: Timgings looked identical/noisy/sometimes-one-faster-sometimes-the-other building mscorlib arm32 on builder.

jaykrell:

jaykrell:
@monojenkins build Linux Aarch64 Interpreter

PR #8736, opened 2018-05-16
[sre] ParameterBuilder.SetConstant for a Nullable<T> parameter

### LINKED ISSUE #8504: System.Reflection.Emit: ParameterBuilder.SetConstant(null) throws for parameters of type Nullable<T>
Given a `ParameterBuilder` for a parameter of type `Nullable<T>`, a call to [`ParameterBuilder.SetConstant(null)`](https://github.com/mono/mono/blob/89ecd02722d1aa484dca476a800435639b544fd8/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs#L95-L109 "see source code for `ParameterBuilder.SetConstant`") will throw an exception, claiming that `null` is not a supported constant type.

However, it should be perfectly possible to assign a null reference to such a parameter: The Roslyn C# compiler will turn this C# type definition:

```csharp
public interface IInterface
{
    void Method(int? arg = null);
}
```

into this IL:

```
.class interface public auto ansi abstract IInterface
{
    .method public hidebysig newslot abstract virtual instance void Method ([opt] valuetype [mscorlib]System.Nullable`1<int32> arg) cil managed 
    {
        .param [1] = nullref
    }
}
```

## Steps to reproduce

```csharp
using System;
using System.Reflection;
using System.Reflection.Emit;

class Program
{
    static void Main()
    {
        var assemblyBuilder = AppDomain.CurrentDomain.DefineDynamicAssembly(new AssemblyName("DynamicAssembly"), AssemblyBuilderAccess.RunAndSave);
        var moduleBuilder = assemblyBuilder.DefineDynamicModule("DynamicModule", "DynamicAssembly.dll");
        var typeBuilder = moduleBuilder.DefineType("IInterface", TypeAttributes.Abstract | TypeAttributes.Interface);
        var methodBuilder = typeBuilder.DefineMethod("Method", MethodAttributes.Public | MethodAttributes.Abstract | MethodAttributes.Virtual, typeof(void), new[] { typeof(int?) });
        var parameterBuilder = methodBuilder.DefineParameter(1, ParameterAttributes.Optional | ParameterAttributes.HasDefault, "arg");
        parameterBuilder.SetConstant(null);
    }
}
```

### Current behavior

`parameterBuilder.SetConstant(null)` causes the following run-time exception:

> System.ArgumentException: System.Nullable`1[System.Int32] is not a supported constant type.

### Expected behavior

The call to `parameterBuilder.SetConstant(null)` should succeed for parameters with type `Nullable<T>`, as it does on the CLR / on Windows.

## Platform on which I noticed this

[ ] macOS
[X] Bash on Ubuntu on Windows subsystem of Windows 10 (Ubuntu 16.04.4 LTS)
[ ] Windows

**Version used**:

Mono 5.10.1.47

## Stacktrace

```
Unhandled Exception:
System.ArgumentException: System.Nullable`1[System.Int32] is not a supported constant type.
  at System.Reflection.Emit.ParameterBuilder.SetConstant (System.Object defaultValue) [0x0008c] in <9bbab8f8a2a246e98480e70b0839fd67>:0
  at Program.Main () [0x0007a] in <aef196c188e64047a149a0675e76255d>:0
[ERROR] FATAL UNHANDLED EXCEPTION: System.ArgumentException: System.Nullable`1[System.Int32] is not a supported constant type.
  at System.Reflection.Emit.ParameterBuilder.SetConstant (System.Object defaultValue) [0x0008c] in <9bbab8f8a2a246e98480e70b0839fd67>:0
  at Program.Main () [0x0007a] in <aef196c188e64047a149a0675e76255d>:0
```

### LINKED ISSUE #8597: System.Reflection.Emit: ParameterBuilder.SetConstant(value) throws ArgumentException for nullable value typed parameters
For example, given a `ParameterBuilder parameterBuilder` for a parameter of type `int?`, `parameterBuilder.SetConstant(42)` will throw an `ArgumentException` ("Constant does not match the defined type").

## Steps to reproduce

```csharp
using System;
using System.Reflection;
using System.Reflection.Emit;

class Program
{
    static void Main()
    {
        var assembly = AppDomain.CurrentDomain.DefineDynamicAssembly(new AssemblyName("Assembly"), AssemblyBuilderAccess.Run);
        var module = assembly.DefineDynamicModule("Module");
        var type = module.DefineType("Type", TypeAttributes.Public | TypeAttributes.Abstract | TypeAttributes.Interface);
        var method = type.DefineMethod("Method", MethodAttributes.Public | MethodAttributes.Abstract | MethodAttributes.Virtual, typeof(void), new[] { typeof(int?) });
        var parameter = method.DefineParameter(1, ParameterAttributes.Optional | ParameterAttributes.HasDefault, "arg");
        parameter.SetConstant(42);
    }
}
```

### Current behavior

The call to `parameter.SetConstant` fails with a `System.ArgumentException` ("Constant does not match the defined type").

### Expected behavior

The call to `ParameterBuilder.SetConstant` should succeed (as it does on the CLR and CoreCLR) because 42 is a valid value for an `int?`-typed parameter.

## Platform on which I noticed this

[ ] macOS
[ ] Linux
[X] Bash on Ubuntu on Windows subsystem of Windows 10 (Ubuntu 16.04.4 LTS)

**Version used**:

Mono 5.10.1.47

## Stacktrace

```
System.ArgumentException: Constant does not match the defined type.
  at System.Reflection.Emit.ParameterBuilder.SetConstant (System.Object defaultValue) [0x0004c]
  at Program.Main()
```

lambdageek:
You should be able to do the moral equivalent of the code below using SRE:

```csharp
public enum E {E1, E2};
public interface Itf {
    void MyMethod (int? i = null, int? i2 = 42, E? e1 = null, E? e2 = E.E2);
}
```

Fixes https://github.com/mono/mono/issues/8504
Fixes https://github.com/mono/mono/issues/8597

### DIFF
--- a/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
@@ -96,12 +96,24 @@ namespace System.Reflection.Emit {
 		{
 			if (position > 0) {
 				Type t = methodb.GetParameterType (position - 1);
-				if (defaultValue != null && t != defaultValue.GetType ()) {
-					if(!t.IsEnum || t.UnderlyingSystemType != defaultValue.GetType ())
-						throw new ArgumentException ("Constant does not match the defined type.");
+				if (defaultValue == null) {
+					if (!(t.IsGenericType && t.GetGenericTypeDefinition () == typeof(Nullable<>)))
+						throw new ArgumentException("Null is not a valid constant value for this type");
+				} else {
+					if (t.IsEnum) {
+						Type underlyingType = Enum.GetUnderlyingType (t);
+						if (defaultValue.GetType () != underlyingType && defaultValue.GetType () != t) {
+							throw new ArgumentException ("Constant does not match the defined type.");
+						}
+					} else {
+						if (!t.IsAssignableFrom (defaultValue.GetType ()))
+							throw new ArgumentException ("Constant does not match the defined type.");
+					}
+
+					if (t.IsValueType && !t.IsPrimitive && !t.IsEnum && t != typeof (DateTime) &&
+					    !(t.IsGenericType && t.GetGenericTypeDefinition () == typeof (Nullable<>)))
+						throw new ArgumentException ("" + t + " is not a supported constant type.");
 				}
-				if (t.IsValueType && !t.IsPrimitive && !t.IsEnum && t != typeof (DateTime))
-					throw new ArgumentException ("" + t + " is not a supported constant type.");
 			}
 
 			def_value = defaultValue;
--- a/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
@@ -95,25 +95,8 @@ namespace System.Reflection.Emit {
 		public virtual void SetConstant (object defaultValue)
 		{
 			if (position > 0) {
-				Type t = methodb.GetParameterType (position - 1);
-				if (defaultValue == null) {
-					if (!(t.IsGenericType && t.GetGenericTypeDefinition () == typeof(Nullable<>)))
-						throw new ArgumentException("Null is not a valid constant value for this type");
-				} else {
-					if (t.IsEnum) {
-						Type underlyingType = Enum.GetUnderlyingType (t);
-						if (defaultValue.GetType () != underlyingType && defaultValue.GetType () != t) {
-							throw new ArgumentException ("Constant does not match the defined type.");
-						}
-					} else {
-						if (!t.IsAssignableFrom (defaultValue.GetType ()))
-							throw new ArgumentException ("Constant does not match the defined type.");
-					}
-
-					if (t.IsValueType && !t.IsPrimitive && !t.IsEnum && t != typeof (DateTime) &&
-					    !(t.IsGenericType && t.GetGenericTypeDefinition () == typeof (Nullable<>)))
-						throw new ArgumentException ("" + t + " is not a supported constant type.");
-				}
+				TypeBuilder.SetConstantValue (methodb.GetParameterType (position - 1),
+							      defaultValue, ref defaultValue);
 			}
 
 			def_value = defaultValue;
--- a/mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
@@ -1903,6 +1903,125 @@ namespace System.Reflection.Emit
 		{
 			return base.IsAssignableFrom (typeInfo);
 		}
+
+		internal static bool SetConstantValue (Type destType, Object value, ref Object destValue)
+		{
+			// Mono: This is based on the CoreCLR
+			// TypeBuilder.SetConstantValue except it writes to an
+			// out argument instead of doing an icall, and it uses
+			// TypeCode instead of CorElementType (like
+			// MonoTypeEnum) which we don't have in our corlib and
+			// our member fields are different.
+
+			// This is a helper function that is used by ParameterBuilder, PropertyBuilder,
+			// and FieldBuilder to validate a default value and save it in the meta-data.
+
+			if (value != null) {
+				Type type = value.GetType ();
+
+				// We should allow setting a constant value on a ByRef parameter
+				if (destType.IsByRef)
+					destType = destType.GetElementType ();
+
+				// Convert nullable types to their underlying type.
+				// This is necessary for nullable enum types to pass the IsEnum check that's coming next.
+				destType = Nullable.GetUnderlyingType (destType) ?? destType;
+
+				if (destType.IsEnum)
+				{
+					//                                   |  UnderlyingSystemType     |  Enum.GetUnderlyingType() |  IsEnum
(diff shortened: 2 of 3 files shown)


mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
```diff
@@ -96,12 +96,24 @@ public virtual void SetConstant (object defaultValue)
 		{
 			if (position > 0) {
 				Type t = methodb.GetParameterType (position - 1);
-				if (defaultValue != null && t != defaultValue.GetType ()) {
-					if(!t.IsEnum || t.UnderlyingSystemType != defaultValue.GetType ())
-						throw new ArgumentException ("Constant does not match the defined type.");
+				if (defaultValue == null) {
+					if (!(t.IsGenericType && t.GetGenericTypeDefinition () == typeof(Nullable<>)))
+						throw new ArgumentException("Null is not a valid constant value for this type");
```
stakx: The Roslyn compilers can be made to encode a metadata constant of `null` for *any* parameter type, e.g. in this scenario:

```csharp
void Method<T>(T arg = default(T)) { }
```

Because the type of `T` is unknown at compile-time, the Roslyn compilers always encode `default(T)` as `null`.

It should be possible to do the same with Reflection.Emit. The above `if`s might collide with this particular scenario.

stakx: Also, the Roslyn C# compiler generally encodes `default(SomeStruct)` as a `null` metadata constant:

```
struct SomeStruct { }

void Method(SomeStruct arg = default(SomeStruct)) { }
```

(`decimal` and `DateTime` are notable exceptions because they can be encoded in custom attributes, too, which the C# compiler does for `decimal` and the VB.NET compiler does for both types.)

Again, it should be possible to do the same with Reflection.Emit, and these `if`s might collide with that.

Basically, setting a default value of `null` should be allowed always. (I agree this is weird, but that's what the Roslyn compilers do today and what people might encounter via Reflection.)

mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
```diff
@@ -96,12 +96,24 @@ public virtual void SetConstant (object defaultValue)
...
+				} else {
+					if (t.IsEnum) {
+						Type underlyingType = Enum.GetUnderlyingType (t);
```
stakx: This check might fail if `t` is `Nullable<SomeEnum>`. You would then be unable to set the default value for a nullable enum parameter.

You might have to "unwrap" nullable types before the check, e.g.:

```csharp
var unpackedIfNullableT = Nullable.GetUnderlyingType(t) ?? t;
if (unpackedIfNullableT.IsEnum) ...
```

mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
stakx: (Oh and if `t` is a reference type, `null` should definitely be an allowed default value. Looks like the above `if`s collide with that, too.)

mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
```diff
@@ -96,12 +96,24 @@ public virtual void SetConstant (object defaultValue)
...
+							throw new ArgumentException ("Constant does not match the defined type.");
+						}
+					} else {
+						if (!t.IsAssignableFrom (defaultValue.GetType ()))
+							throw new ArgumentException ("Constant does not match the defined type.");
+					}
+
+					if (t.IsValueType && !t.IsPrimitive && !t.IsEnum && t != typeof (DateTime) &&
+					    !(t.IsGenericType && t.GetGenericTypeDefinition () == typeof (Nullable<>)))
+						throw new ArgumentException ("" + t + " is not a supported constant type.");
```
stakx: What if `t` is `Nullable<T>` where `T` is an unsupported type? For example, what if `t == typeof(decimal?)`? In that case, an exception should be thrown as well.

mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs
```diff
@@ -96,12 +96,24 @@ public virtual void SetConstant (object defaultValue)
...
+						Type underlyingType = Enum.GetUnderlyingType (t);
+						if (defaultValue.GetType () != underlyingType && defaultValue.GetType () != t) {
+							throw new ArgumentException ("Constant does not match the defined type.");
+						}
+					} else {
+						if (!t.IsAssignableFrom (defaultValue.GetType ()))
+							throw new ArgumentException ("Constant does not match the defined type.");
+					}
+
+					if (t.IsValueType && !t.IsPrimitive && !t.IsEnum && t != typeof (DateTime) &&
```
marek-safar: nit: you already have the t.IsEnum check above so this could be moved to `else` clause

lambdageek:
What's the best way to stay in sync with CoreCLR - use the same logic as CoreCLR.
Rewrote SetConstant in terms of a helper that has the logic from https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/Reflection/Emit/TypeBuilder.cs

marek-safar:
@lambdageek best way is to use their code directly but we are not there yet, though copying the code from there is fine as it'll make the future work easier

Issue #8726, opened 2018-05-16
Loading denied assemblies via `Assembly.LoadFrom` can cause issues, eg. for VSMac


radical:
Usually, if an assembly depends on a known "bad"(windows specific implementation) assembly then the mono runtime can deny loading
it. But if the load is initiated explicitly like via `Assembly.LoadFrom`, then the assembly gets loaded (https://github.com/mono/mono/blob/2018-02/mono/metadata/image.c#L1363-L1365)
with a `Loading problematic image ...`.

For example, take `System.Runtime.InteropServices.RuntimeInformation.dll`. It's `RuntimeInformation.IsOSPlatform` API's response
depends on what platform it was built was for. So, a windows specific version would return true even it is running on macOS. Normally,
this gets caught and denied loading. But if it is explicitly loaded via `Assembly.LoadFrom`, then that will break
any subsequent code that depends on that API!

This can show up in VSMac, if some addin loads the bad assembly before the good copy has been loaded into the app domain. Then the
subsequent use of msbuild APIs that depend on `IsOSPlatform` break:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.TypeInitializationException: The type initializer for 'Microsoft.Build.Collections.MSBuildNameIgnoreCaseComparer' threw an exception. ---> System.EntryPointNotFoundException: GetSystemInfo
  at (wrapper managed-to-native) Microsoft.Build.Shared.NativeMethodsShared.GetSystemInfo(Microsoft.Build.Shared.NativeMethodsShared/SYSTEM_INFO&)
```

Here the windows specific code path runs on macOS, because of the `IsOSPlatform`'s result!

Related: https://github.com/mono/monodevelop/issues/4788

radical:
@lambdageek @kumpera @marek-safar @mhutch @akoeplinger

Therzok:
Nice find!

marek-safar:
@lambdageek did your context load changes fix the problem?

marek-safar:
@lambdageek I certainly prefer consistent behaviour in this case and we should use same logic as we do when loading the assembly using other API. That leaves use with only reflection context APIs behaving differently but that sounds reasonable to me.

I'd like to see this in 2018-04 because we are still testing the release and we can take it out if necessary. I agree that backporting to earlier versions (15.8 and before) might be problematic but we'd have to find a different workaround in that case.

marek-safar:
[...]

PR #8737, opened 2018-05-17
[loader] LoadFrom of problematic images should reprobe

lambdageek:
The "problematic images" are Windows-specific assemblies that are often
included with applications that do not work on Mono, but for which Mono has its
own implementation.

Previously (9ec8ec5389ae9177853442bdadbff3eedcf93000) we changed mono to deny
loading problematic images using Assembly.Load or when one assembly references
another and we find a problematic image using the usual assembly search.

For Assembly.LoadFrom, we used to emit a warning and then load the assembly anyway.

In this commit, we change the behavior: If Assembly.LoadFrom tries to open a
problematic image, we will instead probe for an assembly of that name in the
default context.  That should eventually end up opening Mono's version of that
assembly.

Example:
  Suppose a problematic System.Runtime.InteropServices.RuntimeInformation.dll
  is in the application bin path: /home/user/myapp/bin

  1. User code does (for whatever reason)
  Assembly.LoadFrom ("/home/user/myapp/bin/System.Runtime.InteropServices.RuntimeInformation.dll")

  2. We find the image and try to open it; add it to the images absfpath to
  image hash; note that there are no binding redirects; but that it is a
  problematic image.  We extract the assembly
  name ("System.Runtime.InteropServices.RuntimeInformation", version x.y.z.w
  and public key token 123456789a) from the problematic image.

  3. We probe for "System.Runtime.InteropServices.RuntimeInformation version
  x.y.z.w public key token 123456789a" in the default context.

  4. We find
  /home/user/myapp/bin/System.Runtime.InteropServices.RuntimeInformation.dll in
  the application bin path; we try to open it, we see it's already in the
  images hash, but it's a problematic image,  and we're not in a loadfrom
  context (we re-probed with default!) so we keep probing

  5. We find
  <prefix>/lib/mono/4.5/Facaded/System.Runtime.InteropServices.RuntimeInformation.dll,
  it's not a problematic image, so we open that and return it to (3)

  6. We return to (2) and close the problematic image and return the good one
  instead.

  7. Assembly.LoadFrom returns with the good image.

Fixes https://github.com/mono/mono/issues/8726

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -2372,7 +2372,7 @@ mono_assembly_has_reference_assembly_attribute (MonoAssembly *assembly, MonoErro
 MonoAssembly*
 chain_redirections_loadfrom (MonoImage *image, MonoImageOpenStatus *out_status)
 {
-	MonoImageOpenStatus status;
+	MonoImageOpenStatus status = MONO_IMAGE_OK;
 	MonoAssembly *redirected = NULL;
 
 	redirected = mono_assembly_binding_applies_to_image (image, &status);
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -358,6 +358,10 @@ static MonoAssembly*
 mono_assembly_load_full_gac_base_default (MonoAssemblyName *aname, const char *basedir, MonoAssemblyContextKind asmctx, MonoImageOpenStatus *status);
 static MonoAssembly*
 mono_assembly_load_full_nodomain (MonoAssemblyName *aname, MonoAssemblyContextKind asmctx, MonoImageOpenStatus *status);
+static MonoAssembly*
+chain_redirections_loadfrom (MonoImage *image, MonoImageOpenStatus *status);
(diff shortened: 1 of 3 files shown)


luhenry:
@monojenkins build failed

lambdageek:
All the xunit tests failed on linux with `System.IO.fileNotFoundException: Invalid Image` (but not on OSX??) that's... pretty related

marek-safar:
@kumpera please review

lambdageek:
@monojenkins build failed

lambdageek:
I don't think a later `Load` in a default context would get in trouble.   This should make it more difficult for a problematic image to be loaded into a domain.

Previously if there was a `LoadFrom` followed by a `Load` the `Load` would get the problematic image, now neither `LoadFrom` nor `Load` would see it.

Additionally on master `LoadFile` should obey binding redirects now, so if `Load(asmName)` would be redirected, the initial `LoadFile` will be redirected too, once we probe the assembly name.

PR #8742, opened 2018-05-17
mkbundle interop changes

grendello:
mkbundle works by generating a small C source file which, in addition to
embedded assemblies/configs/etc, contains a certain amount of code to initialize
the bundle, uncompress the assemblies and set runtime options (e.g. AOT mode).
This source code is generated based on a number of templates shipped as
resources inside `mkbundle.exe` and is not part of any public, maintained API.

The generated code contains calls to Mono runtime which work just fine when
mkbundle is used to generate a desktop application bundle with the default
template code. However, it breaks when a third party (like Xamarin.Android)
embeds Mono runtime and loads it dynamically **after** the application is
initialized. Such third parties must ensure that the bundle's startup code is
able to access the Mono runtime functions it needs by providing pointers to
them, loaded using e.g. the `dlsym` function.

However, as the Mono calls are hard-coded into the generated source code, the
third party needs to resort to search-and-replace methods (e.g. [Xamarin.Android](https://github.com/xamarin/xamarin-android/blob/62e7792ca52949f2fa213330d292865ad78d2bfa/src/Xamarin.Android.Build.Tasks/Tasks/MakeBundleNativeCodeExternal.cs#L159-L166))
which are very fragile and prone to undetected breakage.

One example of such breakage which slipped under the radar is
https://github.com/xamarin/xamarin-android/issues/1651 which was caused by
addition of the `mono_jit_set_aot_mode` call to the generated code. The change
slipped under the radar and was discovered only after release.

This commit attempts to make the situation better by establishing a sort of a
code contract between "upstream" mkbundle and the third party. Since the
template code is not "public" (nor should it be), the third party cannot simply
include a header file with the API definitions. Instead, the idea here is to
ensure that all and any Mono runtime calls needed by the template code are
placed in a dispatch structure as pointers to functions and that the structure
needs to be initialized by the embedding party before calling the
`mono_mkbundle_init()` function. The third party is responsible for maintenance
of its own version of the dispatch structure as well as for its initialization
via the `initialize_mono_api` function in the template code. The generated code
must include the third party's version of the structure, in place of the default
one as shipped with `mkbundle`, and to make it possible this commit adds the
`--mono-api-struct-path` mkbundle parameter which points to a file that's
injected into the generated code, replacing the default structure definition.

Following that, the `mono_mkbundle_init()` code calls the `validate_api_struct`
function which checks each and every pointer in the structure for `null`,
exiting the application with error if any `null` pointer is encountered.
This (along with the `initialize_mono_api` function in the template) has the
additional effect that if the third party's structure is incompatible with the
default one a build error will happen, thus allowing early detection of obvious
incompatibilities/omissions.

As long as all the new Mono runtime calls are added to the structure, the above
should make it possible for issues like the one linked to above to not happen in
the future.

Sample of the generated `temp.c` code [is available here](https://gist.github.com/grendello/465ecec96b3bee801e6bbc5a28019833) and 
the Xamarin.Android counterpart to the chanes proposed in this commit [is found here](https://github.com/grendello/xamarin-android/commit/33380c5ec397b1c4085eccf191942451715b4f5f)

Interim fix for [Xamarin.Android issue1651](https://github.com/xamarin/xamarin-android/issues/1651) [is available here](https://github.com/xamarin/xamarin-android/pull/1683)


mcs/tools/mkbundle/mkbundle.cs
```diff
@@ -1012,6 +1032,12 @@ static void GenerateBundles (List<string> files)
 			}
 			tc.WriteLine ("\tNULL\n};\n");
 
+			// This must go before any attempt to access `mono_api`
+			template_stream = System.Reflection.Assembly.GetAssembly (typeof(MakeBundle)).GetManifestResourceStream ("template_common.inc");
```
jonpryor: Why not `using (template_stream = ...) {...}`?

mcs/tools/mkbundle/bundle-mono-api.inc
```diff
@@ -0,0 +1,25 @@
...
+static void
+mkbundle_log_error (const char *format, ...)
+{
+	va_list ap;
+
+	va_start (ap, format);
+	vfprintf (stderr, format, ap);
+	va_end (ap);
+}
+#endif
```
jonpryor: This should be:

```c
#endif /* ! USE_DEFAULT_MONO_API_STRUCT */
```

jonpryor:
This PR should also update [`mkbundle.1`](https://github.com/mono/mono/blob/master/man/mkbundle.1) to mention the new `mkbundle --mono-api-struct-path FILE` option and the *semantics* of the specified `FILE`, in particular that it needs to define a `BundleMonoAPI` containing various function pointer fields, and that it needs to define a `mkbundle_log_error()` function.

mcs/tools/mkbundle/template_common.inc
```diff
@@ -0,0 +1,61 @@
...
+init_default_mono_api_struct ()
+{
+#ifdef USE_DEFAULT_MONO_API_STRUCT
+	mono_api.mono_register_bundled_assemblies = mono_register_bundled_assemblies;
+	mono_api.mono_register_config_for_assembly = mono_register_config_for_assembly;
+	mono_api.mono_jit_set_aot_mode = mono_jit_set_aot_mode;
+	mono_api.mono_aot_register_module = mono_aot_register_module;
+	mono_api.mono_config_parse_memory = mono_config_parse_memory;
+	mono_api.mono_register_machine_config = mono_register_machine_config;
+#endif
```
jonpryor: `#endif` should contain a trailing comment for what symbols it's ending.

marek-safar:
@monojenkins backport 2018-04

sourceskyboxer:
But I have tried because generation failure for Windows, Linux and Mac and other operating systems.....
https://gist.github.com/sourceskyboxer/fce14d8cb694f2218c704d303ab9a268

If you love failure - I will ignore you because you make mkbundle failure of generation

Issue #7615, opened 2018-03-14
Custom MarshalAs attributes are not included in ParameterInfo objects retrieved via reflection


Nihlus:
## Steps to Reproduce

1. Create a method with an attribute decorated with the `MarshalAs` attribute.
2. Use reflection to retrieve a `MethodInfo` instance, and then a `ParameterInfo` instance corresponding to the attribute.
3. Inspect the `CustomAttributes` property.
4. Observe that the `MarshalAs` attribute is not included in the collection.

-- alternatively --

1. Compile the [provided minimal test case][1] using `csc Program.cs` (substituting `Program.cs` for the filename you saved the source as)
3. Execute the test case using `mono Program.exe` (substituting `Program.exe` for the output filename)
2. Observe the output of the application. You should see an output similar to the following:

```
Custom attribute count: 0
Running on Mono.
Type of value in internal field: MarshalAsAttribute
```

## Current Behavior
In the current development version of Mono, `MarshalAs` attributes are not included in the collection assigned to `CustomAttributes`  property. Instead, they are present as an instance of the attribute, assigned to a private field (`marshalAs`).

## Expected Behavior
The `MarshalAs` attribute should be included in the collection assigned to the `CustomAttributes` property, as is the behaviour on the .NET Framework and .NET Core.

### On which platforms did you notice this

-[ ] macOS
-[x] Linux ([System Information][2])
-[ ] Windows

**Version Used**:
```
Mono JIT compiler version 5.10.0.170 (tarball Sat Mar 10 09:49:08 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)

```

[1]: https://hastebin.com/subovodode.cs
[2]: https://gist.github.com/400659620841d3e1a96b792e52b00841

lambdageek:
The example call the getter of the `S.R.E.ParameterInfo.CustomAttributes` property which ultimately ends up in `MonoCustomAttrs.GetCustomAttributesData` which doesn't look like it picks up the pseudo custom attributes from the `ICustomAttributeProvider`.

[...]

PR #8744, opened 2018-05-17
Include MarshalAs attributes in the collection assigned to CustomAttributes property

MaximLipnin:
Fixes: #7615

### DIFF
--- a/mcs/class/corlib/System/MonoCustomAttrs.cs
+++ b/mcs/class/corlib/System/MonoCustomAttrs.cs
@@ -301,6 +301,35 @@ namespace System
 				throw new ArgumentNullException ("obj");
 
 			CustomAttributeData [] attrs = GetCustomAttributesDataInternal (obj);
+
+			object [] pseudoAttrs = System.Array.FindAll (GetPseudoCustomAttributes (obj, null), attr => attr is MarshalAsAttribute);
+			if (pseudoAttrs.Length > 0)
+			{
+				CustomAttributeData [] maAttrsData = new CustomAttributeData [pseudoAttrs.Length];
+				for (var i = 0; i < pseudoAttrs.Length; i++) {
+					MarshalAsAttribute maAttr = (MarshalAsAttribute)pseudoAttrs[i];
+					var attrValue = maAttr.Value;
+					var customAttributeData = (CustomAttributeData)Activator.CreateInstance (typeof (CustomAttributeData), true);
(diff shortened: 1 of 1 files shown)


MaximLipnin:
@lambdageek Could you also take a look at this draft? Is it correct way to proceed?

mcs/class/corlib/System/MonoCustomAttrs.cs
```diff
@@ -301,6 +301,35 @@ internal static IList<CustomAttributeData> GetCustomAttributesData (ICustomAttri
...
 			CustomAttributeData [] attrs = GetCustomAttributesDataInternal (obj);
+
+			object [] pseudoAttrs = System.Array.FindAll (GetPseudoCustomAttributes (obj, null), attr => attr is MarshalAsAttribute);
+			if (pseudoAttrs.Length > 0)
+			{
+				CustomAttributeData [] maAttrsData = new CustomAttributeData [pseudoAttrs.Length];
+				for (var i = 0; i < pseudoAttrs.Length; i++) {
+					MarshalAsAttribute maAttr = (MarshalAsAttribute)pseudoAttrs[i];
+					var attrValue = maAttr.Value;
+					var customAttributeData = (CustomAttributeData)Activator.CreateInstance (typeof (CustomAttributeData), true);
```
lambdageek: Maybe add a static method to `CustomAttributeData.CreateFromPseudoCustomAttr()` that initializes a `CustomAttributeData` instance directly with the ctorArgs and ctorInfo?  Then you wouldn't have to jump through reflection.

mcs/class/corlib/System/MonoCustomAttrs.cs
```diff
@@ -301,6 +301,35 @@ internal static IList<CustomAttributeData> GetCustomAttributesData (ICustomAttri
 				throw new ArgumentNullException ("obj");
 
 			CustomAttributeData [] attrs = GetCustomAttributesDataInternal (obj);
+
+			object [] pseudoAttrs = System.Array.FindAll (GetPseudoCustomAttributes (obj, null), attr => attr is MarshalAsAttribute);
```
lambdageek: Should we do this for all pseudo custom attributes, not just `MarshalAs`?

marek-safar:
I commented on the original issue we are missing handling for all pseudo custom attribute in this scenario and we need more light-weight way to get them (less allocation)

PR #8752, opened 2018-05-17
Do not mixup Unicode and ASCII.

jaykrell:
Not sure what is happening here. Is the type incorrect throughout, or being passed to wrong function?

### DIFF
--- a/support/minizip/iowin32.c
+++ b/support/minizip/iowin32.c
@@ -94,7 +94,7 @@ voidpf ZCALLBACK win32_open_file_func (opaque, filename, mode)
     }
 
     if ((filename!=NULL) && (dwDesiredAccess != 0))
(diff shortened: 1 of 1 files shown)


lateralusX:
Looks like this code can be reached through the monoposixhelper library and DLLImported by NativeZip.cs that in turn should marshal this as ANSI in the call, so if called this bug should have been hit. Looks like there is a test suite cover parts of the ZipArchive, MonoTests.System.IO.Compression.ZipArchiveTests and that completes with this fix. It also looks like the code path hitting the CreateFile will not be hit by any of the tests since GetStream is not used (and that will trickle down to CreateFile). Best could be to write a test that hits the codepath and validate the behavior before the fix (expecting a failure) and after the fix.

jaykrell:
Code looks dead. Let's prune.

PR #8754, opened 2018-05-17
More turnkey use of LLVM on Mac -- add CFLAGS=-lc++.

jaykrell:


vargaz:
Why is this needed ? llvm-config should already return the flags needed to compile/link.

jaykrell:
I have been doing this manually every time I run configure or autogen.

Otherwise I get link errors.

Today I ran run-jenkins.sh which formed CFLAGS itself and I got the usual  link errors so finally  suggest fix it (perhaps there are parameters to  run-jenkins.sh)

I  built monollvm in the first place with:
```
$ more /o/monollvm/config.log 
 $ /s/monollvm/configure -prefix=/i/monollvm --enable-optimized --enable-targets=x86 x86_64
```

llvm-config  says:

```
$ llvm-config --cflags
-I/i/monollvm/include  -D_DEBUG -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -O3  -fno-common

$ llvm-config --cxxflags

-I/i/monollvm/include  -D_DEBUG -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -O3  -std=c++11 -fvisibility-inlines-hidden -fno-exceptions -fno-rtti -fno-common -Woverloaded-virtual -Wcast-qual

$ llvm-config --ldflags
-L/i/monollvm/lib 

$ which  llvm-config
/i/monollvm/bin/llvm-config

```

Maybe I built LLVM incorrectly?

vargaz:
It looks like autoconf adds it:
```
   # The c++ standard library used by llvm doesn't show up in $LLVM_SYSTEM_LIBS so add it manually
   if echo $LLVM_CXXFLAGS | grep -q -- '-stdlib=libc++'; then
      LLVM_LIBS="$LLVM_LIBS -lc++"
   else
      LLVM_LIBS="$LLVM_LIBS -lstdc++"
   fi
```

jaykrell:
I guess I just been using the -with-runtime-preset lately and that should share this logic.
I probably  stopped using --with-llvm, I just set PATH and -with-runtime-preset.
I can poke some more. I agree that logic is better-looking than mine if it works.

jaykrell:
Actually, you can see, that logic doesn't match what  I do, and  maybe doesn't work..

vargaz:
llvm-config --cxxflags contains -std=c++11, so configure adds -lstdc++ to LLVM_LIBS. This should work imho.

jaykrell:
That’s not what it greps for there.

vargaz:
The grep fails, so it will add:
LLVM_LIBS="$LLVM_LIBS -lstdc++"

jaykrell:
libtool is resolving -lstdc++ to /usr/local/lib/libstdc++.dylib.
Probably I built/installed gcc or  clang at some point (they aren't  homebrew symlnks), before giving up on any kind of piecemeal system, all the bindings are  too lose). If I delete those it works. dll hell.  :(

jaykrell:
I guess not needed. Either -lc++ or -lstdc++ usually work and autoconf will add one. What I had done is broken one of them, so used the other, it seems. I didn't realize.

Issue #7075, opened 2018-02-15
Epic: Client Certificate Support


baulig:
I have a pretty decent understanding what needs to be done (and what needs a bit of research) in order for us to implement client certificates.

This summary will be updated once I started to actually work on the implementation.

Use this meta issue to track bugs and feature requests regarding client certificates.

##### X509Store versus Client Certificates

Please read the section called "X509Store versus Client Certificates" in #7019 for an in-depth explanation of the difference between the X509Store and Client Certificate Support.

##### TLS Renegotiation

Supporting Client Certificates requires TLS Renegotiation.  This does not only change the way the TLS Handshake works, it also changes the semantics of an already-established connection.  First, we need to establish rules when a renegotiation will be allowed and when it won't be.  Then we need to deal with violations of these rules and make sure we handle them in a reasonable manner.  Some of these corner cases cannot be tested with automated unit tests and will require manual interaction.

# Milestone I
- [x] TLS Renegotiation
	- [x] Handshake update
- [x] SslStream properties update
- [ ] HttpClient support (ideally after CoreFX import)
- [ ] HttpListener update (should wait for CoreFX import)

# Milestone II
- [ ] X509Store support for the client certificates #7019
	- [ ] Windows
	- [ ] Mac/iOS
	- [ ] Android

baulig:
There are several bugs on bugzilla related to client certificates.  I will add them here while working on the implementation to track which issues I'm already written tests for.

baulig:
We are getting ready to land - anything that's not addressed / fixed with #8756 and #8758 should be added here, so I can have a look at it.

PR #8756, opened 2018-05-17
[System]: Epic: Client Certificate Support - Part One.

baulig:
This is the first of two Pull Requests to implement Client Certificates :-)

Part One binds the new native APIs that will be used internally, finishes
the certificate selection callbacks, but without the more riskly changes
to the underlying handshake and I/O layer.

Part Two will bring support for TLS Renegotiation - and due to the required
changes in the underlying handshake, it is the more risky one.

* `Mono.Security.Interface.MonoTlsSettings`: Add `ClientCertificateIssuers`.

* `MobileTlsContext`:
  - fully implement `SelectClientCertificate()`; the `acceptableIssuers` parameter
    is now actually set and we also have a reasonable default selection.
  - add `CanRenegotiate` and `RenegotiateAsync()` - these are not hooked up yet.

* `AppleTlsContext`:
  - we will only ever call `RequirePeerTrust()` once per session, so we can
    also remove it alltogether and just use `EvaluatePeerTrust()` instead.
  - use proper exceptions for `SslStatus.PeerNoRenegotiation` and `PeerUnexpectedMsg`.
  - don't call `SetClientSideAuthenticate()` on the client side.
  - bind and hook up `SSLAddDistinguishedName()` and `SSLCopyDistinguishedNames()`.
  - bind `SSLReHandshake()`.

* `MobileAuthenticatedStream`: minor cleanups; there will be more uses of the new
  `GetInvalidNestedCallException()` helper class once Part Two lands.

* Enable some more constants in `SecureTransport.cs`.

* Add new `MonoBtlsError.GetErrorReason()` and `mono_btls_error_get_reason()`
  implementation, only supporting `SSL_R_NO_RENEGOTIATION` at the moment.

* Add new native `mono_btls_ssl_ctx_set_client_ca_list()` function and managed
  `MonoBtlsSslCtx.SetClientCertificateIssuers()`; hooked up via
  `MonoTlsSettings.ClientCertificateIssuers`.

* According to a comment in the header file, `SSL_get_client_CA_list()` may only
  be called during the selection callback or while the handshake is paused.
  To respect this restriction, we now call it during the client certificate
  selection callback and pass the list from native to managed.
  - changed signature of `MonoBtlsSelectFunc` from
    `int (* MonoBtlsSelectFunc) (void *instance)` to
    `int (* MonoBtlsSelectFunc) (void *instance, int countIssuers, const int *sizes, void **issuerData)`.
  - the managed counter-part is in `MonoBtlsSslCtx.NativeSelectFunc` / `NativeSelectCallback`.

* MonoBtlsContext:
  - use the new `MonoBtlsError.GetErrorReason()` to throw a `TlsException` with
    `AlertDescription.NoRenegotiation` that can be checked for by user code.
  - `SelectCallback()` now has a `string[] acceptableIssuers` argument; pass it
    to `SelectClientCertificate()`.
  - the native backend does not support TLS Renegotiation, so `CanRenegotiate`
    always returns false.

Implements #7075


baulig:
@monojenkins commit apidiff

mcs/class/System/Mono.Btls/MonoBtlsError.cs
```diff
@@ -56,6 +56,9 @@ static class MonoBtlsError
 		[DllImport (MonoBtlsObject.BTLS_DYLIB)]
 		extern static void mono_btls_error_get_error_string_n (int error, IntPtr buf, int len);
 
+		[DllImport (MonoBtlsObject.BTLS_DYLIB)]
+		extern static int mono_btls_error_get_reason (int error);
```
marek-safar: We should bump mscorlib (to ensure unmanaged bits don't get out of sync)

PR #8760, opened 2018-05-18
No _Static_assert in old compiler.

### LINKED ISSUE #8759: Build of Master fails with GCC 4.4.7
## Steps to Reproduce

1. Clone mono/master on a rhel6.x/centos 6.x machine with GCC 4.4.7 as the default compiler
2. configure and make
3. observe build failure

### Current Behavior

Build fails as GCC 4.4.7 does not understand `_Static_assert`

### Expected Behavior

`make` succeeds OR `configure` reports that the GCC version is not supported and you need at least GCC 4.7.x.

[ ] macOS
[x ] Linux
[ ] Windows

**Version Used**:

```
$ gcc --version
gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-17)
```

## Stacktrace

Offending lines in [method-to-ir.c](https://github.com/mono/mono/blob/master/mono/mini/method-to-ir.c#L6294):
```
#ifndef _MSC_VER
_Static_assert (sizeof (float) == 4, "");
_Static_assert (sizeof (double) == 8, "");
#endif
```

jaykrell:
Fixes #8759

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -6290,11 +6290,6 @@ typedef union _MonoOpcodeParameter {
 	guchar *branch_target;
 } MonoOpcodeParameter;
 
-#ifndef _MSC_VER
-_Static_assert (sizeof (float) == 4, "");
-_Static_assert (sizeof (double) == 8, "");
-#endif
-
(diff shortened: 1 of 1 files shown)


jaykrell:

PR #8764, opened 2018-05-18
Small cleanups for mini-ppc.h

NattyNarwhal:
* mini-ppc-os.h doesn't exist. If the OS lacks a define,
	  give an error instead of a broken include.

	* Remove redundant Mac OS case.

	* Move AIX define for sigaction above cross-compile case,
	  so it won't get lost in the noise there.


lewurm:

Issue #8692, opened 2018-05-12
TcpClient throws "An address incompatible with the requested protocol was used"


cosminstirbu:
## Steps to Reproduce

```
using System;
using System.Net;
using System.Net.Sockets;
using System.Threading.Tasks;

namespace TcpClientTest
{
    class MainClass
    {
        public static async Task Main(string[] args)
        {
            var tcpClient = new TcpClient(new IPEndPoint(IPAddress.Any, 0));
            await tcpClient.ConnectAsync("localhost", 51139);
        }
    }
}
```

Running the snippet above throws "An address incompatible with the requested protocol was used" on Mono versions newer than 5.8.1.0.

### Current Behavior

"An address incompatible with the requested protocol was used" is thrown.

### Expected Behavior

No exception should be thrown

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 5.8.1.0 (2017-10/6bf3922f3fd Thu Mar  8 17:00:31 EST 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          yes(3.6.0svn-mono-master/8b1520c8aae)
	GC:            sgen (concurrent by default)

marek-safar:
It appears to be a regression in 5.10

PR #8765, opened 2018-05-18
[System] BeginConnect fails on ipv6 if Socket is ipv4

EgorBo:
Fixes: https://github.com/mono/mono/issues/8692
In the following code:
```
var tcpClient = new TcpClient(new IPEndPoint(IPAddress.Any, 0));
tcpClient.BeginConnect("localhost", 51139, _ =>
{
    tcpClient.EndConnect(_);
    Console.WriteLine("Done");
}, null);
```
`Dns.GetHostAddresses` resolves "localhost"  as two addresess: ipv6 (::1) and ipv4 (127.0.0.1). Currently mono tries the first one and throws an exception (my Socket is ipv4) - `An address incompatible with the requested protocol was used.`
instead of that it should try ipv4 after that and only throw an exception if this second attempt also fails.

so it will match .NET Core behavior for `BeginConnect()`: https://github.com/dotnet/corefx/blob/7a1007f2ef04fd9ec576cdb198ff53af4f18c77b/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L4978-L4981
BTW, synchronous `Connect()` in Mono has correct behavior (https://github.com/mono/mono/blob/master/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs#L1134), so the bug reproduces only if `BeginConnect()` or `ConnectAsync()` are used.

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -1046,6 +1046,10 @@ namespace System.Net.Sockets
 				sockares.EndPoint = remoteEP = sockares.socket.RemapIPEndPoint (ep);
 			}
 
+			if (!sockares.socket.CanTryAddressFamily(sockares.EndPoint.AddressFamily)) {
+				throw new ArgumentException(SR.net_invalidAddressList);
+			}
+
 			int error = 0;
 
 			if (sockares.socket.connect_in_progress) {


EgorBo:
@monojenkins build failed

marek-safar:
@monojenkins backport 2018-04

marek-safar:
@monojenkins backport 2018-02

Issue #8701, opened 2018-05-14
AOT failure if ref readonly with generics is used


dalexsoto:
_From @Belorus on May 11, 2018 11:34_

### Steps to Reproduce

Compile attached project in **Debug**

### Expected Behavior

It compiles

### Actual Behavior

It doesn't compile. Fails on AOTing

### Environment

```
Xamarin.iOS
Version: 11.10.1.177 (Visual Studio Community)
Hash: 7e782c1e
Branch: d15-7
Build date: 2018-04-25 15:27:13-0400
```

### Build Logs

```
/Users/antonl/Projects/monosyne/Samples/Scene/Projects/IOS/MTOUCH: Error MT3001: Could not AOT the assembly '/Users/antonl/Projects/monosyne/Samples/Scene/Projects/IOS/obj/iPhone/Debug/mtouch-cache/64/Build/Scene.exe' (MT3001) (Scene.IOS)
```
### Example Project (If Possible)

[Scene.zip](https://github.com/xamarin/xamarin-macios/files/1995134/Scene.zip)

### Important Code 

```
public ref readonly Components<T> GetComponents()
{
    return ref _components;
}
```

_Copied from original issue: xamarin/xamarin-macios#4077_

marek-safar:
@vargaz could you look into it (will probably have to backport it)

PR #8729, opened 2018-05-16
[aot] Fix encoding/decoding of types with custom modifiers.

vargaz:
Fixes https://github.com/mono/mono/issues/8701.

PR #8772, opened 2018-05-18
[aot] Fix encoding/decoding of types with custom modifiers. (#8729)

vargaz:
* [aot] Fix encoding/decoding of types with custom modifiers.

Fixes https://github.com/mono/mono/issues/8701.

* [aot] Add a test, disabling for now because it depends on c# 7.2.

* [aot] Bump AOT file format version.

* [aot] Only disable the test with mcs.

* [aot] Fix compilation of test.


vargaz:
@monojenkins backport 2018-02

PR #8793, opened 2018-05-21
Interrupt syncronous file read/write IO on Win32 by registering inter…

joncham:
…rupt callback.

Fixes interrupt/abort on Windows while a thread is doing a Read/Write.

### DIFF
--- a/mono/metadata/w32file-win32.c
+++ b/mono/metadata/w32file-win32.c
@@ -81,23 +81,47 @@ mono_w32file_delete (const gunichar2 *name)
 	return res;
 }
 
-gboolean
-mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
+static void
+cancel_w32_io (HANDLE file_handle)
 {
-	gboolean res;
-	MONO_ENTER_GC_SAFE;
-	res = ReadFile (handle, buffer, numbytes, bytesread, NULL);
-	MONO_EXIT_GC_SAFE;
+	CancelIoEx (file_handle, NULL);
+}
+
+gboolean
+mono_w32file_read (gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
+{
+	gboolean res = FALSE;
+	gboolean interrupted;
+
+	mono_thread_info_install_interrupt (cancel_w32_io, handle, &interrupted);
+	if (!interrupted)
+	{
+		MONO_ENTER_GC_SAFE;
+		res = ReadFile (handle, buffer, numbytes, bytesread, NULL);
+		MONO_EXIT_GC_SAFE;
+
+		mono_thread_info_uninstall_interrupt (&interrupted);
+	}
+
 	return res;
 }
 
 gboolean
 mono_w32file_write (gpointer handle, gconstpointer buffer, guint32 numbytes, guint32 *byteswritten)
 {
-	gboolean res;
-	MONO_ENTER_GC_SAFE;
-	res = WriteFile (handle, buffer, numbytes, byteswritten, NULL);
-	MONO_EXIT_GC_SAFE;
+	gboolean res = FALSE;
+	gboolean interrupted;
+
+	mono_thread_info_install_interrupt (cancel_w32_io, handle, &interrupted);
+	if (!interrupted)
+	{
+		MONO_ENTER_GC_SAFE;
+		res = WriteFile (handle, buffer, numbytes, byteswritten, NULL);
+		MONO_EXIT_GC_SAFE;
+
+		mono_thread_info_uninstall_interrupt (&interrupted);
+	}
+
 	return res;
 }
 


mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,63 @@ mono_w32file_delete (const gunichar2 *name)
...
+
 gboolean
-mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
+mono_w32file_read (gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
 {
 	gboolean res;
+	gboolean interrupted;
+	HANDLE duplicate;
+
+	duplicate = OpenThread (THREAD_ALL_ACCESS, FALSE, GetCurrentThreadId ());
```
jaykrell: MSDN says  THREAD_TERMINATE, not THREAD_ALL_ACCESS.
There is also MAXIMUM_ALLOWED if  you want the maximum you can get and  be prepared for access denied on later use.

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,63 @@ mono_w32file_delete (const gunichar2 *name)
...
+	if (!duplicate)
+		return FALSE;
+
+	mono_thread_info_install_interrupt (abort_w32_synchronous_io, duplicate, &interrupted);
+	if (interrupted) {
+		CloseHandle (duplicate);
```
jaykrell: Ideally just one call to CloseHandle at the end of function.

mono/metadata/w32file-win32.c
jaykrell: Actually  since you are using  GetCurrentThreadId anyway, just use GetCurrentThread(). It is very much faster (just returns a constant -2) and grants all access. No need to duplicate or close, as long as you  are only using it on the current thread. If are you passing it across thread/process boundaries, then use DuplicateHandle(GetCurrentThread()), not OpenThread.

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,63 @@ mono_w32file_delete (const gunichar2 *name)
...
 }
 
 gboolean
 mono_w32file_write (gpointer handle, gconstpointer buffer, guint32 numbytes, guint32 *byteswritten)
 {
 	gboolean res;
+	gboolean interrupted;
+	HANDLE duplicate;
+
+	duplicate = OpenThread (THREAD_ALL_ACCESS, FALSE, GetCurrentThreadId ());
```
jaykrell: This should be GetCurrentThread() or DuplicateHandle(GetCurrentThread()).

mono/metadata/w32file-win32.c
joncham: Hmm, I initially had `GetCurrentThread` and the call to `CancelSynchronousIo` was failing. Let me revisit.

jaykrell: I can understand that -- the Cancel happens from another thread, right? In that case, DuplicateHandle(GetCurrentThread()) is slightly preferred. Same number of syscalls either way -- open(GetCurrentThreadId()) vs. duplicate(GetCurrentThread()). But Duplicate is kinda better. Granted it takes more parameters. GetCurrentThreadId and GetCurrentThread are both fast/small, not syscalls. GetCurrentThreadId just reads memory,  GetCurrentThread just returns -2.

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,61 @@ mono_w32file_delete (const gunichar2 *name)
...
+
+	mono_thread_info_install_interrupt (abort_w32_synchronous_io, duplicate, &interrupted);
+	if (interrupted)
+		goto cleanup;
+
 	MONO_ENTER_GC_SAFE;
 	res = WriteFile (handle, buffer, numbytes, byteswritten, NULL);
 	MONO_EXIT_GC_SAFE;
+
+	mono_thread_info_uninstall_interrupt (&interrupted);
```
luhenry: The problem with `mono_thread_info_uninstall_interrupt` is that it races with `abort_w32_synchronous_io`, since the interrupt might happen at any moment. See https://github.com/mono/mono/blob/dde9ee5e79e756212625dc995ea5d2c0d479e372/mono/metadata/w32handle.c#L766 for how to protect yourself against that.

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,61 @@ mono_w32file_delete (const gunichar2 *name)
...
 gboolean
-mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
+mono_w32file_read (gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
 {
-	gboolean res;
+	gboolean res = FALSE;
+	gboolean interrupted;
+	HANDLE duplicate;
+
+	if (!DuplicateHandle (GetCurrentProcess (), GetCurrentThread (), GetCurrentProcess (), &duplicate, THREAD_TERMINATE, FALSE, 0))
```
jaykrell: Initialize duplicate to NULL and check it for NULL before CloseHandle.

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,61 @@ mono_w32file_delete (const gunichar2 *name)
...
+}
+
 gboolean
-mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
+mono_w32file_read (gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread)
 {
-	gboolean res;
+	gboolean res = FALSE;
+	gboolean interrupted;
+	HANDLE duplicate;
```
jaykrell: HANDLE duplicate = NULL;

mono/metadata/w32file-win32.c
```diff
@@ -81,23 +81,61 @@ mono_w32file_delete (const gunichar2 *name)
...
+		goto cleanup;
+
 	MONO_ENTER_GC_SAFE;
 	res = ReadFile (handle, buffer, numbytes, bytesread, NULL);
 	MONO_EXIT_GC_SAFE;
+
+	mono_thread_info_uninstall_interrupt (&interrupted);
+
+cleanup:
+	CloseHandle (duplicate);
```
jaykrell: if  (duplicate)
 CloseHandle (duplicate);

mono/metadata/w32file-win32.c
jaykrell: Maybe the interrupt should take address of the local? And InterlockedExchange it? Might be easier to understand?

joncham: What is the race? I am not closing the handle in the callback

joncham:
Updated to use `CancelIoEx`

jaykrell:
My idea of the InterlockedExchange was bogus, because cancel does not close the handle.
However this whole thing seems racy.
What if the cancel comes in after the first check but before the ReadFile?
It will be lost.
It kinda seems like to make this work, you have  to be using I/O completion ports, and cancelation would be queueing a completion. That what you can do it before the read or during.

jaykrell:
I suspect there isn't a race-free solution here for synchronous I/O.

joncham:
So where does that leave this? I plan on applying this to our branch as it fixes (well enough) real issues we are encountering.

jaykrell:
I guess we should take, but bugs remain.

jaykrell:
I'm still leary of cancel for the file vs the thread and sync vs. async.
 Can the handle be in use by another thread, and is the desired cancellation for all threads? Or would it be a mistake to cancel arbitrary thread?
 Do we know if this is sync or async (i.e.  there is no overlapped)? Async would seem to have much more capability and more easily canceled/interrupted?
 Can/should we discourage sync I/O?? Seems difficult. Can we turn sync I/O into async + interruptable wait? (I know this tries but it isn't convincing, because of the race).

For example, can we allocate an event, pass it in overlapped, and signal for interruption? i.e. even in the synchronous case? Overlapped is not just for async -- for example you can use it to give a seek pointer with the read. This seems worth a quick try, or close read of the docs.

How does this work on Unix? You send a signal to the thread?

lateralusX:
I believe the best solution (as pointed out in a separate discussions) would be to use async IO and alterable wait to get overlapped result, but that mandates that the handle is opened to support async IO in the first place and that's potentially not something we can enforce and changing the implementation to use async IO in mono_w32file_create will trickle all the way up to the BCL and, as far as I can see, potentially reach users as a native handle through FileStream.Handle (deprecated), so changing this might cause regressions. It also looks like Mono's implementation of FileStream async support is done through async delegates, but that will end up in the same call to ReadFile with the same native handle (doing a synchronous read), so even for that scenario it will still be a synch handle (but potentially that could be change for at least that scenario), but regardless we might end up in a blocking ReadFile call in some scenarios anyways.

One alternative to reduce the risk of unwanted side effects is to use an OVERLAPPED struct in the call to ReadFile and use that in the call to CancelIoEx? This should be possible even for file handles opened without async IO support and should only abort that specific IO request.

This won't solve the potential race between registering the callback, and getting the abort call before the call to ReadFile getting to a place where cancel could have effect, but I believe that won't be solvable without switching over to async IO and synchronizing the abort thread with the thread registering the callback and doing the IO calls.

lateralusX:
One thing that actually might close the race condition mentioned above is doing logic in the callback based on CancelIOEx result. If there is no pending IO request on that file handle for that specific OVERLAPPED instance the API should behave like this (from the docs):

"If this function cannot find a request to cancel, the return value is 0 (zero), and GetLastError returns ERROR_NOT_FOUND."

One option could be to implement retry logic in the callback when this happens, it could also cordinate with the thread doing IO to break this retry logic in case there is a problem with register/unregister of the callback and the completion of executing the callback (using an atomic state or similar).

luhenry:
@lateralusX what needs to be done next to get this PR in?

jaykrell:
I believe Jon indicated that just registering the interrupt, with an empty handler, broke things.

lateralusX:
@joncham Did you have any success implementing using an OVERLAPPED structure and just cancel the specific IO request when interrupting the call?

jaykrell:
The problem was uninstall interrupt changes last error and there is a form of this can progress, i.e. whatever is in the Unity repo?
I think uninstall interrupt should never change lasterror.

lateralusX:
@joncham A slight variation of this fix has been merged in https://github.com/mono/mono/pull/12654, that also fixes the race in socket aborts. I leverage the same semantics to add support interrupting blocking read/write on files when requesting a syscall abort (via Thread.Abort) that should solve the same problem addressed in this PR, but without the need to register/unregister the interrupt handler by the blocking thread (instead extending the APC mechanism we already had in place on Windows).

PR #8799, opened 2018-05-21
[System]: Bring 'System.Diagnostics.DiagnosticsSource' from CoreFX.

baulig:


baulig:
It is used by some code in `System.Net.Http` - I'd need to check how much of it and whether we could possibly stub some of it out for mobile.

PR #8727, opened 2018-05-16
Fix assert to print and not crash, build the eglib tests and fix their warnings.

jaykrell:
g_assert isn't printing correctly, and can crash, depending on which target.

Also build the eglib tests and fix their warnings.

PR #8801, opened 2018-05-21
[runtime] Fixup freeform, incorporate feedback from backports

alexanderkyte:
In the backports, @jaykrell noticed a mistake that was merged with the initial freeform PR. This commit incorporates the fixes made during backport that apply to master.

### DIFF
--- a/mono/eglib/goutput.c
+++ b/mono/eglib/goutput.c
@@ -143,10 +143,9 @@ g_assertion_message (const gchar *format, ...)
 {
 	va_list args;
 
-	va_start (args, format);
-
 	g_vasprintf (&failure_assertion, format, args);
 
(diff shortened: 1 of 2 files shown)


mono/eglib/goutput.c
```diff
@@ -143,10 +143,9 @@ g_assertion_message (const gchar *format, ...)
 {
 	va_list args;
 
-	va_start (args, format);
-
 	g_vasprintf (&failure_assertion, format, args);
 
+	va_start (args, format);
```
jaykrell: This is not correct.

jaykrell: Fix is  this PR:
https://github.com/mono/mono/pull/8727

PR #8802, opened 2018-05-21
[System]: Bring 'HashCode' from CoreFX.

baulig:


mcs/class/System/ReferenceSources/Interop.cs
```diff
@@ -0,0 +1,21 @@
+using System;
```
marek-safar: Don't do it this way but import their PAL implementation instead

mcs/class/System/common.sources
```diff
@@ -794,6 +795,8 @@ corefx/SR.cs
 
 ../../../external/corefx/src/Common/src/CoreLib/System/Text/ValueStringBuilder.cs
 
+../../../external/corefx/src/Common/src/CoreLib/System/HashCode.cs
```
marek-safar: This is System.Runtime type better to have it in mscorlib

baulig:
I'll completely refactor this and do a new PR when ready.

PR #8807, opened 2018-05-22
Backport "Implement Freeform Unmanaged Error" to 2018-02

alexanderkyte:


alexanderkyte:
Failure blocking merging: ```RuntimeReflectionExtensionsTests.GetRuntimeMethod``` has a missing method exception

alexanderkyte:
@monojenkins retry OS X i386

mono/eglib/goutput.c
```diff
@@ -139,6 +139,7 @@ g_assertion_message (const gchar *format, ...)
 {
 	va_list args;
 
+	va_start (args, format);
```
jaykrell: We are not sure this is portable and  it is not efficient -- formats twice.
The master fix avoids both  problems.

alexanderkyte: @jaykrell merged onto 2018-02 is code that moves the va_start below the other function rather than making two. I had misunderstood your feedback before. I think that a fix/undo for that should be merged.

alexanderkyte:
@luhenry can I have a review? This commit was merged into 2018-04 already

PR #8810, opened 2018-05-22
Fix bunch of new gcc/linux warnings I recently introduced by accident.

jaykrell:


lewurm:
what warnings does it fix?

jaykrell:
Using the values from one enum with the type of another.

jaykrell:
Approve?

lewurm:
Good practice would be to copy/paste the warning into the commit message, so it's searchable if someone else hits this issue. Also mention the commit sha1 somewhere that introduced the regression.

jaykrell:
I can't find the warnings in CI logs (of other PRs).
We went from switch (int or unsigned char) case CEE to switch (enum) CEE and should be switch (enum) MONO_CEE.
It took a few commits and is really hard to see in the history.

jaykrell:
```
$ git show 1507b9b531798f4229de8b2891bbd531e947264d | grep switch
-		switch (*ip) {
+		switch (il_op) {
-				switch (op) {
+				switch (il_op) {
```

However, yet:

```
$ git show 1507b9b531798f4229de8b2891bbd531e947264d | grep il_op
+		int il_op = *ip;
+		switch (il_op) {
```

and then..

```
$ git show 298aa78a47c6f027b7083c374d4d2868717a7295 |  grep il_op\ \= | more
                MonoOpcodeEnum il_op = MonoOpcodeEnum_Invalid;
-               int il_op = *ip;
```

so we briefly had two locals il_op, one int, one enum.
Removing the int left the enum.
Originally it was switch (*ip).
It is confusing, but did allow smaller steps.

luhenry:
With this change we went from opcode starting with `MONO_` being mono specific to all opcodes being prefixed with `MONO_` even if they are not mono specific. This breaks assumptions and the regression should have been avoided in the first place.

jaykrell:
The MONO_ prefix was already available on all opcodes. That didn't change.

jaykrell:
I believe the mono-specific ones all have "_mono_" in them, not just at the start.
```
		case MONO_CEE_MONO_LDPTR_CARD_TABLE:
		case MONO_CEE_MONO_LDPTR_NURSERY_START:
		case MONO_CEE_MONO_LDPTR_NURSERY_BITS:
		case MONO_CEE_MONO_LDPTR_INT_REQ_FLAG:
		case MONO_CEE_MONO_LDPTR_PROFILER_ALLOCATION_COUNT:
```
etc. There are lots of those.

PR #8816, opened 2018-05-22
Add missing sre features to the mscorlib link.xml

joshpeterson:
To make sure that System.Reflection.Emit types in the class libraries
are properly removed by the linker, mark them with the "sre" feature in
the embedded linker descriptor file for mscorlib.


joshpeterson:
@marek-safar: Are you referring to the `feature` attributes on the `method` elements? If so, we found that they are necessary. The linker will not preserve a given method required by a feature because its type is required by that feature.

marek-safar:
@joshpeterson yeah it's kind of confusing where the feature is used to preserve fields.

Thanks for the fix

PR #8822, opened 2018-05-22
[corefx] Import System.Hashcode

alexischr:
/cc @baulig


mono/metadata/Makefile.am
```diff
@@ -178,7 +178,8 @@ libmono_system_native_la_SOURCES = \
 	../../external/corefx/src/Native/Unix/System.Native/pal_networkstatistics.c \
 	../../external/corefx/src/Native/Unix/System.Native/pal_networkstatistics.h \
 	../../external/corefx/src/Native/Unix/System.Native/pal_tcpstate.c \
-	../../external/corefx/src/Native/Unix/System.Native/pal_tcpstate.h
+	../../external/corefx/src/Native/Unix/System.Native/pal_tcpstate.h \
+	../../external/corefx/src/Native/Unix/System.Native/pal_random.c
```
luhenry: Add `../../external/corefx/src/Native/Unix/System.Native/pal_random.h`

baulig:
Upstream PR: https://github.com/dotnet/corefx/pull/29876

mcs/class/corlib/corlib.dll.sources
```diff
@@ -1676,6 +1677,9 @@ corert/RuntimeAugments.cs
 
 ../../../external/corefx/src/Common/src/CoreLib/System/Reflection/BindingFlags.cs
 
+../../../external/corefx/src/Common/src/CoreLib/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
```
akoeplinger: This will break on Windows as we don't build System.Native there. You'll need to include this only for Unix profiles and include `external/corefx/src/Common/src/CoreLib/Interop/Windows/BCrypt/Interop.BCryptGenRandom.cs` on Windows instead (like System.dll does it with similar Interop files).

marek-safar:
@monojenkins commit apidiff

akoeplinger:
FWIW this only adds Hashcode to net_4_x, we'll need to add it to the mobile profiles too if we want to use it in types that are available on all profiles.

akoeplinger:
I had to revert this as it broke the Windows build (https://github.com/mono/mono/pull/8879), please redo the PR :)

PR #8826, opened 2018-05-23
[jit] Add a 'no-verbose-gdb' option to turn off very verbose gdb crash dumps.

vargaz:


mono/mini/mini-posix.c
```diff
@@ -863,11 +863,13 @@ native_stack_with_gdb (pid_t crashed_pid, const char **argv, FILE *commands, cha
 	fprintf (commands, "attach %ld\n", (long) crashed_pid);
 	fprintf (commands, "info threads\n");
 	fprintf (commands, "thread apply all bt\n");
-	for (int i = 0; i < 32; ++i) {
-		fprintf (commands, "info registers\n");
-		fprintf (commands, "info frame\n");
-		fprintf (commands, "info locals\n");
-		fprintf (commands, "up\n");
+	if (!mini_get_debug_options ()->no_verbose_gdb) {
+		for (int i = 0; i < 32; ++i) {
```
lewurm: can we decrease the number of frames too? 10?

jaykrell: It is too verbose, agreed. [author of verbosity]

jaykrell: Might want to invert it.

vargaz:
Alternatively, we can make add an option to turn it on and enable it on CI.

PR #8831, opened 2018-05-23
[llvm] Disable the 'x86-call-frame-opt' pass on x86, it generates code which breaks unwinding.

vargaz:


jaykrell:
This is related to e.g.:
```
 MONO_PATH="mcs/class/lib/testing_aot_full/:mcs/class/lib/testing_aot_full//tests:.:$MONO_PATH" gdb --args mono/mini/mono  --full-aot mcs/class/lib/testing_aot_full/nunit-lite-console.exe  mcs/class/lib/testing_aot_full/tests/testing_aot_full_System.Xml_test.dll      -test=MonoTests.System.XmlSerialization.XmlSerializerTests.TestSerializeZeroFlagEnum_InvalidValue
```

The function `System_Xml_Serialization_XmlSerializationWriterInterpreter_WriteRoot_object`.

and should make a good dent in CI of Linux LLVM FullAOT -- specifically Linux.

luhenry:

jaykrell:
1. I think this is actually supposed to work, but the mono unwind codes are being produced incorrectly,  for reasons I don't know.
2. Remove the  check for >500? 3.9 has similar set of test failures.

PR #8841, opened 2018-05-23
[ppc] implement machinery for thunks that works on ppc64

lewurm:
* use gcc builtin for icache flushing
---------------------------------------------------
* narrow immediate value for relative branches

   since we don't know the exact distance yet, we have to guess. sometimes this guess was not defensive enough.

---------------------------------------------------
* use different thunk machinery that works on ppc64

   The old thunk machinery was sort of working on 64bit (before f4bd084 ), but it relies on having on
the memory layout of allocated chunks for code. Either the property got lost due to some refactoring, or it was never true on ppc64. Anyway, this patch switches the implementation to something we do on arm32 too: If we potentially need some thunks for the emitted code, we reserve some space for it at the end of the code buffer. That wastes a bit more memory, because we often don't need it, but at least it works reliable.

Untested on 32bit systems.

### DIFF
--- a/mono/mini/mini-ppc.c
+++ b/mono/mini/mini-ppc.c
@@ -2028,7 +2028,7 @@ if (0 && ins->inst_true_bb->native_offset) { \
 	ppc_bc (code, (b0), (b1), (code - cfg->native_code + ins->inst_true_bb->native_offset) & 0xffff); \
 } else { \
 	int br_disp = ins->inst_true_bb->max_offset - offset;	\
-	if (!ppc_is_imm16 (br_disp + 1024) || ! ppc_is_imm16 (ppc_is_imm16 (br_disp - 1024))) {	\
+	if (!ppc_is_imm16 (br_disp + 8 * 1024) || !ppc_is_imm16 (br_disp - 8 * 1024)) {	\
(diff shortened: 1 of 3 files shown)


lewurm:

lewurm:

lewurm:
@monojenkins rebase

mono/mini/mini-ppc.c
```diff
@@ -734,6 +734,9 @@ void
 mono_arch_flush_icache (guint8 *code, gint size)
 {
 #ifdef MONO_CROSS_COMPILE
```
jaykrell: Why is cross compile broken?
Can we not cross compile a JIT?

lewurm: we don't need to invalidate the instruction cache as we don't try to execute it right away

jaykrell: Cross compile might not mean what I though. I thought, for some reason, we check if build != host/target, and that is what cross_compile meant. Which wouldn't lead to this.
If it is host != target, then aot only, then understood.
i.e. are we cross compiling mono (first case), or building a cross compiler (second case).

lewurm: it's `host != target`:
https://github.com/mono/mono/blob/4260bd28d0c025ce7e4cbc3734a519d6fabe9279/configure.ac#L4114-L4115

PR #8842, opened 2018-05-23
[2018-02] Gracefully handle crashes before mono-state dependencies inited

alexanderkyte:


alexanderkyte:
I'm seeing slightly different tests fail on each arch. I'm going to retry them.

jaykrell:
The non-required LLVM break is a compile error.
Maybe https://github.com/mono/mono/pull/8850 ?

PR #8844, opened 2018-05-23
[jit] Add a 'verbose-gdb' option to turn on very verbose gdb crash du…

vargaz:
…mps.


jaykrell:
I think we need to go further :(
One thread?
Fewer/16 frames?
Redirect to a file?
Off by default but on  in CI?

jaykrell:
I read it backwards. Thought an option  to turn off, not on. So  yeah, good.

PR #8849, opened 2018-05-23
[btls] rewrite mono_btls_ssl_ctx_set_ciphers

davidben:
Rather than reaching into the library internals and potentially violating its invariants, assemble a cipher string and call the supported public API. (There's no need to touch the version-specific cipher lists because mono/btls never configures them.) The existing code also had a bug; it incorrectly assumed `OPENSSL_malloc` returned a zero-initialized buffer for `in_group_flags`. (Though Mono doesn't set `SSL_OP_CIPHER_SERVER_PREFERENCE`, so I suspect this was a no-op.)

Additionally, this should hopefully ease updating to a less outdated version of BoringSSL in the future.

Note I've changed the return value of this function from returning the number of ciphers to merely returning 0/1 for whether the operation succeeded. The only caller of this function, `MonoBtlsSslCtx.SetCiphers`, only cares about success vs. failure.

I ran `make run-test` in `mcs/class/System`, but it wasn't clear there were any tests which exercised `MonoTlsSettings.EnabledCiphers`, so I did some tests manually to confirm it worked.

### DIFF
--- a/mono/btls/btls-ssl-ctx.c
+++ b/mono/btls/btls-ssl-ctx.c
@@ -8,6 +8,8 @@
 
 #include "btls-ssl-ctx.h"
 #include "btls-x509-verify-param.h"
+#include <openssl/bytestring.h>
+#include <string.h>
 
 struct MonoBtlsSslCtx {
 	CRYPTO_refcount_t references;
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

baulig:
We are getting rid of this function, but I don't see a reason not to have this change.

Issue #8767, opened 2018-05-18
[soft debugger] make sure debugee process runs with `--interpreter` too on CI


lewurm:
`TEST_WITH_INTERPRETER=1` only makes the debugger client run with the interpreter, but not the process running the debugger agent

PR #8851, opened 2018-05-23
[Mono.Debugger.Soft.Test] Use interp on debugee

lewurm:
We only used the interpreter on the client side (by mistake). Luckily most tests are working when using the interpreter on the debugee side. 

--------------------------------------------

* [debugger tests] use runtime flags for debugee

------------------------------------------
* [debugger] fix debugger_invoke frame in interp case

fixes DebuggerTests.BreakpointDuringInvoke on the interpreter

------------------------------------------

* [debugger] handle managed-to-native frames

fixes DebuggerTests.InspectThreadSuspenedOnWaitOne on the interpreter

------------------------------------------

* [interp] set SP in monoctx that is passed to exception handling

fixes DebuggerTests.ExceptionFilter on the interpreter

------------------------------------------

Fixes https://github.com/mono/mono/issues/8767

### DIFF
--- a/mcs/build/tests.make
+++ b/mcs/build/tests.make
@@ -212,7 +212,7 @@ endif
 ## FIXME: i18n problem in the 'sed' command below
 run-test-lib: test-local test-local-aot-compile patch-nunitlite-appconfig
 	ok=:; \
-	PATH="$(TEST_RUNTIME_WRAPPERS_PATH):$(PATH)" MONO_REGISTRY_PATH="$(HOME)/.mono/registry" MONO_TESTS_IN_PROGRESS="yes" $(TEST_HARNESS_EXEC) $(test_assemblies) $(NOSHADOW_FLAG) $(TEST_HARNESS_FLAGS) $(LOCAL_TEST_HARNESS_FLAGS) $(TEST_HARNESS_EXCLUDES) $(LABELS_ARG) -format:nunit2 -result:TestResult-$(PROFILE).xml $(FIXTURE_ARG) $(TESTNAME_ARG)|| ok=false; \
+	PATH="$(TEST_RUNTIME_WRAPPERS_PATH):$(PATH)" MONO_REGISTRY_PATH="$(HOME)/.mono/registry" MONO_TESTS_IN_PROGRESS="yes" DBG_RUNTIME_ARGS="$(TEST_RUNTIME_FLAGS)" $(TEST_HARNESS_EXEC) $(test_assemblies) $(NOSHADOW_FLAG) $(TEST_HARNESS_FLAGS) $(LOCAL_TEST_HARNESS_FLAGS) $(TEST_HARNESS_EXCLUDES) $(LABELS_ARG) -format:nunit2 -result:TestResult-$(PROFILE).xml $(FIXTURE_ARG) $(TESTNAME_ARG)|| ok=false; \
 	if [ ! -f "TestResult-$(PROFILE).xml" ]; then echo "<?xml version='1.0' encoding='utf-8'?><test-results failures='1' total='1' not-run='0' name='bcl-tests' date='$$(date +%F)' time='$$(date +%T)'><test-suite name='$(strip $(test_assemblies))' success='False' time='0'><results><test-case name='$(notdir $(strip $(test_assemblies))).crash' executed='True' success='False' time='0'><failure><message>The test runner didn't produce a test result XML, probably due to a crash of the runtime. Check the log for more details.</message><stack-trace></stack-trace></failure></test-case></results></test-suite></test-results>" > TestResult-$(PROFILE).xml; fi; \
 	$$ok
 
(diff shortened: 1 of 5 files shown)


mcs/build/tests.make
```diff
@@ -212,7 +212,7 @@ endif
 ## FIXME: i18n problem in the 'sed' command below
 run-test-lib: test-local test-local-aot-compile patch-nunitlite-appconfig
 	ok=:; \
-	PATH="$(TEST_RUNTIME_WRAPPERS_PATH):$(PATH)" MONO_REGISTRY_PATH="$(HOME)/.mono/registry" MONO_TESTS_IN_PROGRESS="yes" $(TEST_HARNESS_EXEC) $(test_assemblies) $(NOSHADOW_FLAG) $(TEST_HARNESS_FLAGS) $(LOCAL_TEST_HARNESS_FLAGS) $(TEST_HARNESS_EXCLUDES) $(LABELS_ARG) -format:nunit2 -result:TestResult-$(PROFILE).xml $(FIXTURE_ARG) $(TESTNAME_ARG)|| ok=false; \
+	PATH="$(TEST_RUNTIME_WRAPPERS_PATH):$(PATH)" MONO_REGISTRY_PATH="$(HOME)/.mono/registry" MONO_TESTS_IN_PROGRESS="yes" TEST_RUNTIME_FLAGS="$(TEST_RUNTIME_FLAGS)" $(TEST_HARNESS_EXEC) $(test_assemblies) $(NOSHADOW_FLAG) $(TEST_HARNESS_FLAGS) $(LOCAL_TEST_HARNESS_FLAGS) $(TEST_HARNESS_EXCLUDES) $(LABELS_ARG) -format:nunit2 -result:TestResult-$(PROFILE).xml $(FIXTURE_ARG) $(TESTNAME_ARG)|| ok=false; \
```
lewurm: @akoeplinger what do you think about sneaking `TEST_RUNTIME_FLAGS` in here?

vargaz:
The debugger tests used to run with the interpreter, did something regress ?

mcs/build/tests.make
marek-safar: I think this ok even if not very transparent

lewurm:
@vargaz yes, mentioned tests in the PR description regressed. I guess due to the changes around exception handling since then.

mcs/class/Mono.Debugger.Soft/Test/dtest.cs
```diff
@@ -45,6 +45,7 @@ public class DebuggerTests
 	public static bool listening = Environment.GetEnvironmentVariable ("DBG_SUSPEND") != null;
 #endif
 	public static string runtime = Environment.GetEnvironmentVariable ("DBG_RUNTIME");
+	public static string runtime_args = Environment.GetEnvironmentVariable ("TEST_RUNTIME_FLAGS");
 	public static string agent_args = Environment.GetEnvironmentVariable ("DBG_AGENT_ARGS");
```
vargaz: This should be DBG_RUNTIME_ARGS imho.

lewurm: done 👍

mono/mini/interp/interp.c
```diff
@@ -702,6 +702,7 @@ interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, gco
 
 	MonoContext ctx;
 	memset (&ctx, 0, sizeof (MonoContext));
+	MONO_CONTEXT_SET_SP (&ctx, &ext);
```
vargaz: Why is this needed ? ctx should be 0 for interpreter frames, we use MonoStackFrameInfo.frame_addr to compare frames now.

lewurm: I added it to make this check happy:
https://github.com/mono/mono/blob/2c54579df918cf756f38962ac4a106d8a71a14c1/mono/mini/debugger-agent.c#L3213-L3214

`filter_state.ctx` is setup by this:
https://github.com/mono/mono/blob/2c54579df918cf756f38962ac4a106d8a71a14c1/mono/mini/mini-exceptions.c#L1983

is there a better solution?

mono/mini/mini-exceptions.c
```diff
@@ -712,7 +712,7 @@ unwinder_unwind_frame (Unwinder *unwinder,
 
 		/* Process debugger invokes */
 		/* The DEBUGGER_INVOKE should be returned before the first interpreter frame for the invoke */
-		if (unwinder->last_frame_addr > (gpointer)(*lmf)) {
+		if (unwinder->last_frame_addr < (gpointer)(*lmf)) {
```
luhenry: Doesn't that depend on the orientation of the stack? We do support platforms that have different stack orientation, so maybe having the same approach would be useful.

luhenry:
There is still a test failing on x64: `DebuggerTests.BreakpointDuringInvoke` https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/5952/testReport/junit/(root)/DebuggerTests/BreakpointDuringInvoke/

mono/mini/mini-exceptions.c
lewurm: what platforms are those?

luhenry: Turns out I was mistaken. I remembered some places in the code checked for it, but that wasn't enough to support it fully.

PR #8855, opened 2018-05-24
[corlib] Optimizes Stream:CopyTo

marek-safar:
/cc @benaadams


benaadams:
Worth doing Stream.CopyToAsync in same manner while here? https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/IO/Stream.cs#L138-L154

marek-safar:
As I stated in the email there are many more tweaks in the file (e.g. whole ReadWriteTask, etc) so unless you are hitting this I'd prefer to wait for whole implementation replace

marek-safar:
@monojenkins build failed

PR #8856, opened 2018-05-24
[jit] Disable the EqualityComparer.get_Default () optimization for no…

vargaz:
…w, it seems to cause failures in the corlib tests on the ios simulator.


luhenry:
Otherwise, could we just disable this optimization for iOS?

PR #8860, opened 2018-05-24
Remove two methods with incorrect arguments

joshpeterson:
These two methods on `IPersonalizable` don't match the .NET 4.7.1 API.
In both cases the methods should take one `PersonalizableDictionary`
argument, but _that_ type does not exist.

Since nothing in the current class library code is using these methods,
just remove them.


akoeplinger:
@monojenkins commit apidiff

PR #8864, opened 2018-05-24
[corlib]: Add more `System.Buffers` types.

baulig:
This brings `ReadOnlySequence`, `ReadOnlySequenceSegment` and a few others.

### DIFF
--- a/mcs/class/corlib/corlib.dll.sources
+++ b/mcs/class/corlib/corlib.dll.sources
@@ -1758,8 +1758,7 @@ corert/RuntimeAugments.cs
 ../../../external/corefx/src/System.Memory/src/System/Runtime/InteropServices/*.cs
 
 ../../../external/corefx/src/Common/src/CoreLib/System/Buffers/Text/FormattingHelpers.CountDigits.cs
-../../../external/corefx/src/System.Memory/src/System/Buffers/StandardFormat.cs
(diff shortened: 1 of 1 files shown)


mcs/class/corlib/corlib.dll.sources
```diff
@@ -1758,8 +1758,7 @@ corert/RuntimeAugments.cs
 ../../../external/corefx/src/System.Memory/src/System/Runtime/InteropServices/*.cs
 
 ../../../external/corefx/src/Common/src/CoreLib/System/Buffers/Text/FormattingHelpers.CountDigits.cs
-../../../external/corefx/src/System.Memory/src/System/Buffers/StandardFormat.cs
-../../../external/corefx/src/System.Memory/src/System/Buffers/OperationStatus.cs
+../../../external/corefx/src/System.Memory/src/System/Buffers/*.cs
```
marek-safar: We don't want to include all the types as they are old implementation. What is weird though is why it didn't fail and why there is no public API change

akoeplinger: The reason is that we have this line in master already at line 1756 😄

akoeplinger: Looks like you added it in https://github.com/mono/mono/commit/e1e94cb6b183096c06cfe80add57e6cd4bffe339

PR #8865, opened 2018-05-24
[runtime] Make debugger log to ring buffer and maintain thread state model

alexanderkyte:
By more explicitly modeling the transitions made by the debugger as changes in a state machine, we will more easily be able to refactor the control flow and resource lifetimes in the debugger agent. This change should have little to no observable changes to someone using the debugger, but on crash they will have a dump that looks like 

https://gist.github.com/alexanderkyte/05243ad173691092e7cfaf6b7448eba7
or
https://gist.github.com/alexanderkyte/c20a74b496fd7d55c91c8d9c23e302da
.

This information comes from the ring buffer and the modeled state.


vargaz:
I'm not really a fan of this. Why do we have to move the debugger internal state to some other public header ? Can somebody really debug debugger problems using just this log file ? Why is it better than simply enabling logging with --debugger-agent=loglevel=10 ?

luhenry:
@vargaz I assume this is useful to people who do not want to enable the very verbose logging, and still help figuring out what could have gone wrong.

vargaz:
If we really want to do this, then we should add some functions to the debugger code which is called by this state code to collect the state that it needs, instead of just exporting everything.

luhenry:
The property of a state machine is that the state transitions can be done separately from the actions resulting of transitioning state. For example, if you take a look at `mono/utils/mono-threads-state-machine.c`, only the state transitions and logging are done in this file, not all the actions that are the result of transitioning state.

Moving all the code that transitions state to a different file doesn't make it a state machine, it's just moving code. So try to model the various states, the various transitions and the various actions resulting of these transitions; then you can come up with code to encode these. The state machine doesn't need to be thread safe nor atomic, that still can be done by the caller.

alexanderkyte:
I intended to change most of this moved code shortly but it's easy enough to do that then. I'll rebase the old code back into debugger-agent.

mono/mini/debugger-agent.c
```diff
@@ -113,144 +115,6 @@ typedef struct {
 	gboolean setpgid;
 } AgentConfig;
 
-typedef struct
```
luhenry: Why do these struct and define need to move? The state-machine shouldn't have any knowledge of what the various actions are, or you have a leaky abstraction.

alexanderkyte: When the structured logger takes a more enum-based view of the commands and events, a lot of these types will have to be public.

vargaz: Can't the logging be added inside debugger-agent.c ? We are already doing text logging there.

alexanderkyte: These specific named types would have to be made visible in any refactoring of the debugger making any real changes, because they have to be exported in order to export DebuggerTlsData

alexanderkyte: Why do we want the debugger to be a single, impenetrable file with untracked global variables determining state rather than factoring it into easy-to-read subsystems with logged interop points?

kumpera:
I find it particularly weird that we're getting what's essentially 2 logging systems for the debugger.

It's fine to have a flight-recorder that only dumps on failure and that usually means it needs to be low overhead.

I think a similar thing could be accomplished by augmenting how sdb logging is done by:

1) making it an integral part of mono's logging/tracing infrastructure
2) improve mono's logging/tracing infrastructure to have finer controls and a flight recorder (which means we might need to support structured logging).

DavidKarlas:
@monojenkins build pkg

DavidKarlas:
@monojenkins build pkg

DavidKarlas:
@monojenkins build pkg

alexanderkyte:
@monojenkins build pkg

alexanderkyte:
@kumpera 

I think that it's no more strange than how SGEN has two logging systems (the logs, and the binary protocol) and how the threading system has the logs you get with a compile-time change as well as the state machine + logs. 

I agree with you that we should bring these many disparate logging components throughout the runtime together under a couple of abstractions (verbose disk logger, lightweight flight recorder, lightweight non-disk logger) with component-specific implementations. 

Whether you want this PR to be the one that creates the shared logging infrastructure to improve all of these many different logging systems throughout the runtime, or whether this should just add the thread-states-type flight recorder to the debugger, is a choice that I'll leave up to the reviewers. I, personally, think that this PR should stay simple. 

If I'm going to make this a full logging unification PR, we should have a meeting about which interfaces we want to use globally and what the characteristics about each should be.

alexanderkyte:
I moved the public types back into debugger-agent to make the diff look smaller and used some getters and setters to access the fields I needed. Most of the relevant types had already been broken out by Kumpera in the previous debugger-engine PR. 

Also moved the types into mini and made the code only take a separate lock. I do this because the non-debugger-agent-thread threads will never end up paused while holding the lock. Other threads often end up paused while holding the loader lock, making all of the normal metadata code that we use (as well as these helpers) unsafe to access from that debugger-agent thread. These dumpers will be called from that thread. 

A lot of stress testing on my machine shows the locking changes to be resilient to a race. I ran the debugger test suite 100x times without seeing any failures or crashes (aren't we supposed to have some races on master?). I got a deadlock/loader-lock-contention reproduction after 10 test runs previously. 

For anybody interested in reproducing locking problems related to taking the loader lock from debugger-agent.c, they happen incredibly often on Linux+x64 and not at all on OSX desktop.

alexanderkyte:
@vargaz I would appreciate a re-review of this code if you have the time

mono/mini/debugger-agent.c
```diff
@@ -858,6 +863,56 @@ debugger_agent_parse_options (char *options)
 	}
 }
 
+void
+mono_debugger_set_thread_state (gpointer ref, MonoDebuggerThreadState expected, MonoDebuggerThreadState set)
+{
+	DebuggerTlsData *tls = (DebuggerTlsData *) ref;
+
+	if (!tls)
```
luhenry: Use `g_assertf(tls, "Cannot get state of null thread")`.

mono/mini/debugger-agent.c
```diff
@@ -10075,6 +10143,10 @@ debugger_thread (void *arg)
 
 	DEBUG_PRINTF (1, "[dbg] Agent thread started, pid=%p\n", (gpointer) (gsize) mono_native_thread_id_get ());
 
+	gchar *log_each_step_var = g_getenv ("MONO_DEBUGGER_LOG_AFTER_COMMAND");
```
luhenry: Use `g_hasenv`.

mono/mini/debugger-agent.c
```diff
@@ -858,6 +863,56 @@ debugger_agent_parse_options (char *options)
...
+
+	return;
+}
+
+MonoDebuggerThreadState
+mono_debugger_get_thread_state (gpointer ref)
+{
+	DebuggerTlsData *tls = (DebuggerTlsData *) ref;
+
+	if (!tls)
```
luhenry: Use `g_assertf(tls, "Cannot get state of null thread")`.

mono/mini/debugger-agent.c
```diff
@@ -249,6 +251,9 @@ typedef struct {
 
 	/* The currently unloading appdomain */
 	MonoDomain *domain_unloading;
+
+	// The state that the debugger expects the thread to be in
+	MonoDebuggerThreadState thread_state;
 } DebuggerTlsData;
```
luhenry: To avoid having `mono_debugger_tls_thread_id (gpointer debuggerTlsData)` with `debuggerTlsData` being a `gpointer`, you can simply define `DebuggerTlsData` as an opaque struct in `debugger-agent.h` like `typedef struct _ DebuggerTlsData DebuggerTlsData;` and have `struct _DebuggerTlsData` at https://github.com/mono/mono/pull/8865/files#diff-cad67a3642d7f7e00161e3e81c29f25eL168. This will allow you to use `DebuggerTlsData*` anywhere you want, without having to expose its definition.

mono/mini/debugger-engine.h
```diff
@@ -154,6 +157,21 @@ typedef struct {
...
 } DbgEngineStackFrame;
 
+MonoGHashTable *
+mono_debugger_get_thread_states (void);
+
+gboolean
+mono_debugger_is_disconnected (void);
+
+intptr_t 
+mono_debugger_tls_thread_id (gpointer debuggerTlsData);
```
luhenry: Coding convention: variable name is `debugger_tls_data` and return value is `gsize`.

mono/mini/debugger-engine.h
```diff
@@ -5,6 +5,9 @@
 #ifndef __MONO_DEBUGGER_ENGINE_H__
 #define __MONO_DEBUGGER_ENGINE_H__
 
+#include <mono/metadata/seq-points-data.h>
```
luhenry: Why do we need to include `seq-points-data.h`? If that's necessary only for `debugger-agent.c` then we prefer to put it there to reduce the need to rebuild everything whenever we modify one header file.

mono/mini/debugger-agent.c
```diff
@@ -858,6 +863,56 @@ debugger_agent_parse_options (char *options)
...
+{
+	DebuggerTlsData *tls = (DebuggerTlsData *) ref;
+
+	if (!tls)
+		g_error ("Cannot get state of null thread");
+
+	return tls->thread_state;
+}
+
+intptr_t
```
luhenry: Use `gsize`.

mono/mini/debugger-engine.h
alexanderkyte: The debugger-agent.c failed to compile for me for a while, master was broken. I think a subsequent fix was pushed. This was my workaround. I'll remove it. Me pushing this was an accident.

alexanderkyte:
@monojenkins rebuild failed

alexanderkyte:
Looking at windows failure right now, a bit hard to see how it could be triggered by my changes. And it's not wrong, the code there is not typesafe. That's a type of assignment we often do though.

alexanderkyte:
I'm building and running this on my VS2017 install, having problems reproducing the CI failure

mono/mini/debugger-state-machine.h
```diff
@@ -0,0 +1,71 @@
...
+ *   Alexander Kyte (alkyte@microsoft.com)
+ *
+ * (C) 2018 Microsoft, Inc.
+ *
+ */
+#ifndef __MONO_DEBUGGER_STATE_MACHINE__
+#define __MONO_DEBUGGER_STATE_MACHINE__
+
+
+#include <config.h>
```
lambdageek: You shouldn't include config.h in a header.

msvc/libmini-common.targets
```diff
@@ -60,6 +60,8 @@
     <ClCompile Include="$(MonoSourceLocation)\mono\mini\debugger-agent-stubs.c" />
     <ClInclude Include="$(MonoSourceLocation)\mono\mini\debugger-engine.h" />
     <ClCompile Include="$(MonoSourceLocation)\mono\mini\debugger-engine.c" />
+    <ClCompile Include="$(MonoSourceLocation)\mono\mini\debugger-state-machine.c" />
+    <ClCompile Include="$(MonoSourceLocation)\mono\mini\debugger-state-machine.h" />
```
lambdageek: ClInclude, not ClCompile

mono/mini/debugger-state-machine.h
luhenry: I missed this one.

alexanderkyte: Fixed, thanks

msvc/libmini-common.targets
alexanderkyte: Oh my. Thank you so much. I think I looked at that file too many times and became blind to it.

mono/metadata/handle.h
```diff
@@ -487,7 +487,7 @@ extern const MonoObjectHandle mono_null_value_handle;
 static inline void
 mono_handle_assign (MonoObjectHandleOut dest, MonoObjectHandle src)
 {
-	MONO_HANDLE_SUPPRESS (dest->__raw = (gpointer)(src ? MONO_HANDLE_RAW (src) : NULL));
+	MONO_HANDLE_SUPPRESS (dest->__raw = (MonoObject *)(src ? MONO_HANDLE_RAW (src) : NULL));
```
jaykrell: Change is good but why was it failing?

alexanderkyte: Because on windows it's not valid to assign a gpointer rvalue to a MonoObject * lvalue, according to VS's output.

alexanderkyte:
@monojenkins build pkg

PR #8873, opened 2018-05-25
Add missing error check and factor out common code.

jaykrell:


jaykrell:
@monojenkins build Linux i386

PR #8879, opened 2018-05-25
Fix Windows build

akoeplinger:


akoeplinger:
Hmm this is more complicated. We need to move all win32-specific dependencies from net_4_x_corlib.dll.sources to win32_net_4_x_corlib.dll.sources.

Issue #8225, opened 2018-04-13
Mono mkbundle output executable throws: “Error mapping file: No error”


dns:
I wrote simple "hello world" console program in C#, and build it using VS2013 (I also try to build it using mono "csc" but the error is same).

```
> mkbundle --sdk "C:\Program Files\Mono" --deps -o test.exe ConsoleApplication2.exe
Using runtime: C:\Program Files\Mono\bin\mono
     Assembly: D:\SANDBOX\Silverlight\ConsoleApplication2\ConsoleApplication2\bin\Release\ConsoleApplication2.exe
       Config: D:\SANDBOX\Silverlight\ConsoleApplication2\ConsoleApplication2\bin\Release\ConsoleApplication2.exe.config
     Assembly: C:\Program Files\Mono\lib\mono\4.5\mscorlib.dll
     Assembly: C:\Program Files\Mono\lib\mono\4.5\I18N.West.dll
     Assembly: C:\Program Files\Mono\lib\mono\4.5\I18N.dll
Generated test.exe
```

The new "test.exe" need mono-2.0-sgen.dll, so I copy it to "C:\Program Files\Mono\bin" and then run test.exe, but the new executable throw a error message to stdout:

```
> test.exe
Error mapping file: No error
```

**Version Used**:
- Mono 5.12.0.226.x64
- Win10 (64-bit).

marek-safar:
@lateralusX does this look familiar?

lateralusX:
Nothing that I have seen before. This mkbundle execution use the --sdk feature that packages bundles differently into the final binary not using the installed C toolchain. Instead it appends bundles directly to a copy of the mono binary (that in this case is not a static linked version, but depending on mono-2.0-sgen.dll). The error seems to occurs when the runtime tries to load (memory mapping the bundles) the bundles. If this is still an issue I could add it to my list and see if I could repro the problem.

jaykrell:
Get us the .exe to debug?

jaykrell:
Or run it along with sysinternals sysmon or such, watch the file opens for errors, near the end. Is something like what I'd do.

dns:
@jaykrell, Here [out.zip](https://github.com/mono/mono/files/2039956/out.zip)

jaykrell:
It is like I thought from code inspection.
mkbundle aligns things to 4k but 64k alignment is needed.

KERNEL32!MapViewOfFileStub:
[...]

PR #8884, opened 2018-05-25
Fix mkbundle alignment to work with Windows. See https://github.com/mono/mono/issues/8225.

jaykrell:
Consider doing this only for Windows.

### DIFF
--- a/mcs/tools/mkbundle/mkbundle.cs
+++ b/mcs/tools/mkbundle/mkbundle.cs
@@ -470,7 +470,7 @@ class MakeBundle {
 		// Modern bundling starts here
 		if (!custom_mode){
 			if (runtime != null){
-				// Nothing to do here, the user has chosen to manually specify --runtime nad libraries
+				// Nothing to do here, the user has chosen to manually specify --runtime and libraries
 			} else if (sdk_path != null) {
 				VerifySdk (sdk_path);
 			} else if (cross_target == "default" || cross_target == null){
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Could we do this for Windows only for now?   On Linux the page alignment should work at the 4k boundary for all the current distributions - otherwise we would pay the 61k price (well, say 32k on average) for those

jaykrell:
Yeah, based on the “runtime” variable? Does arm64 require 64k? I should try. I also think, we are just mapping them all readonly, no execute, could map them all at once and only align the first to 64k, the rest to 4k. So 32k on average total instead of per image.

jaykrell:
Oh, it looks like we already map once? So how about just align the first to 64k and the rest to 4k.
32k new waste per bundle, not per embedded file? For all platforms?

jaykrell:
Strings are over-aligned?
And no guaranteed terminal nul?

jaykrell:
Is 30k per bundle (not per file) worth optimizing out?
Esp. when we align strings to 4k unnecessarily?

migueldeicaza:
You are right, we only map the file once, and then pass this single mapped file to the rest of the runtime.

Let us get your patch.

migueldeicaza:
The way to check if the target is Windows is to check that the embedded Mono runtime has the signature MZ at the beginning.

migueldeicaza:
The "style" usage here is wrong.    The style indicates what is used on the host, what you want to do is do the right thing for the target OS.

So what you want to do is determine that the alignment is 64k when the runtime that is being bundled i is a Windows runtime, which you can check if the executable that you are packaging up starts with "MZ"

jaykrell:
?

jaykrell:

migueldeicaza:
The new changes for runtime detection are fine

jaykrell:
@monojenkins build failed

jaykrell:
Approve?

PR #8895, opened 2018-05-25
 [threads] Move interrupt mechanism to threads.c 

luhenry:


vargaz:
This stuff was in mono-threads.c because its supposed to be lower level functionality not related to .net stuff.

jaykrell:
Is such distinction useful? What will it be used for other than .net?

kumpera:
I think the distinction of whether it's related or not to dotnet is a bit misguided. The distinction is around whether it requires to interop with managed or not to be implemented.

In this case, interrupt can, and probably should, be implemented without dealing with managed.

My original motivator for mono-threads.c was to start extracting functionality out of metadata/threads.c to slowly cleanup the mess that we have there. We made quite some progress but the effort stalled as the more egregious parts have being addressed.

I believe MonoThreadInternal (and most stuff in System.Thread) need to go away and replaced by simpler portable constructs.

If anything, this PR is a bad idea cuz it's touching more managed memory and not less.

jaykrell:
Is/was there nice aspect here, of working to consolidate Thread, ThreadInternal, ThreadInfo, by reducing ThreadInfo? Or is that hopeless or counterproductive?

PR #8898, opened 2018-05-26
Fix u16to8 for non-Windows.

jaykrell:
Fix u16to8 for non-Windows. I was about to use this for debugging and it didn't compile,
because it passes gunichar2* to strlen.

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -797,13 +797,17 @@ gunichar  *g_utf16_to_ucs4 (const gunichar2 *str, glong len, glong *items_read, 
 gchar     *g_ucs4_to_utf8  (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 gunichar2 *g_ucs4_to_utf16 (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 
-#define u8to16(str) g_utf8_to_utf16(str, (glong)strlen(str), NULL, NULL, NULL)
+#define g_u16len monoeg_g_u16len
+gsize
+g_u16len (const gunichar2 *str);
 
-#ifdef G_OS_WIN32
-#define u16to8(str) g_utf16_to_utf8((gunichar2 *) (str), (glong)wcslen((wchar_t *) (str)), NULL, NULL, NULL)
-#else
-#define u16to8(str) g_utf16_to_utf8(str, (glong)strlen(str), NULL, NULL, NULL)
-#endif
+#define u8to16 monoeg_g_u8to16
+gunichar2*
+u8to16 (const gchar *str);
+
+#define u16to8 monoeg_g_u16to8
+gchar*
+u16to8 (const gunichar2 *str);
 
 /*
  * Path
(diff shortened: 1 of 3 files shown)


mono/eglib/glib.h
```diff
@@ -797,13 +797,17 @@ gunichar  *g_utf16_to_ucs4 (const gunichar2 *str, glong len, glong *items_read,
 gchar     *g_ucs4_to_utf8  (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 gunichar2 *g_ucs4_to_utf16 (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 
-#define u8to16(str) g_utf8_to_utf16(str, (glong)strlen(str), NULL, NULL, NULL)
+#define g_u16len monoeg_g_u16len
```
luhenry: This needs to be moved to `eglib-remap.h`

It's not the first time I request it so at least try to follow the existing convention. You might not agree with them but, again, we are open to discussion, not to disruption 😉

mono/eglib/glib.h
```diff
@@ -797,13 +797,17 @@ gunichar  *g_utf16_to_ucs4 (const gunichar2 *str, glong len, glong *items_read,
...
+g_u16len (const gunichar2 *str);
 
-#ifdef G_OS_WIN32
-#define u16to8(str) g_utf16_to_utf8((gunichar2 *) (str), (glong)wcslen((wchar_t *) (str)), NULL, NULL, NULL)
-#else
-#define u16to8(str) g_utf16_to_utf8(str, (glong)strlen(str), NULL, NULL, NULL)
-#endif
+#define u8to16 monoeg_g_u8to16
+gunichar2*
+u8to16 (const gchar *str);
```
luhenry: Revert to the macro, switching to a function is useless.

mono/eglib/glib.h
```diff
@@ -797,13 +797,17 @@ gunichar  *g_utf16_to_ucs4 (const gunichar2 *str, glong len, glong *items_read,
...
+
+#define u16to8 monoeg_g_u16to8
+gchar*
+u16to8 (const gunichar2 *str);
```
luhenry: Revert to the macro, switching to a function is useless.

mono/eglib/glib.h
```diff
@@ -797,13 +797,17 @@ gunichar  *g_utf16_to_ucs4 (const gunichar2 *str, glong len, glong *items_read,
 gchar     *g_ucs4_to_utf8  (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 gunichar2 *g_ucs4_to_utf16 (const gunichar *str, glong len, glong *items_read, glong *items_written, GError **err);
 
-#define u8to16(str) g_utf8_to_utf16(str, (glong)strlen(str), NULL, NULL, NULL)
+#define g_u16len monoeg_g_u16len
+gsize
+g_u16len (const gunichar2 *str);
```
luhenry: This needs to be renamed to `g_utf16_strlen`.

jaykrell: We have both conventions and I thought about this a while and chose the shorter name. ok.

jaykrell: `str` implies char*, `wcs` implies wchar_t* in ANSI C library. How about omiting `str` here? One of g_utf16len, g_utf16_length, g_utf16length? The other argument, though, is that `wcs` is obscure out side of Windows, `str` is all that is known, so prepend g_utf16_ and  then use exactly existing `str` function: g_utf16_strchr, g_utf16_strstr, g_utf16_strcmp, etc.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

PR #8901, opened 2018-05-26
Allow monodis to disassemble csc.

jaykrell:
This fixes/hacks up monodis enough so it can disassemble csc w/o sigsegv or assertion failures.
Is it right? I don't know.


vargaz:
We don't need to fix problems in monodis anymore, use ikdasm.

jaykrell:
But ok to do so in half-hearted way?
ikdasm also fails many cases, though not here, so I had stopped using it in favor of monodis.
i.e. it fails with F# output.

marek-safar:
We could still fix monodis problems even if ikdasm is prefered tool for disassembly going forward.

However, this is not fix it just hides the underlying issue

PR #8902, opened 2018-05-27
Fix failure to set out parameter.

jaykrell:


mono/metadata/w32file.c
```diff
@@ -713,6 +713,8 @@ ves_icall_System_IO_MonoIO_get_ConsoleError ()
 MonoBoolean
 ves_icall_System_IO_MonoIO_CreatePipe (HANDLE *read_handle, HANDLE *write_handle, gint32 *error)
 {
+	*error = 0;
```
marek-safar: better is `*error = ERROR_SUCCESS;`

jaykrell:

Issue #8903, opened 2018-05-27
Two tests failing due to running as root vs. non-root.


jaykrell:
The tests `Create_Path_ReadOnly` and `GetAttributes_ReadOnly` fail for FullAOT Linux amd64 LLVM, but pass for FullAOT Linux amd64 non-LLVM (and all other configurations).

I believe this is because the LLVM CI is running as root, and the non-LLVM CI is running as builder.

I believe there are two problems here, not related to LLVM.

1. There are runtime behaviors that are probably incorrect when running as root. i.e. how it determines writability of a file -- the code is quite complicated on Unix.

2. CI needs to be run both ways to catch these discrepancies.

marek-safar:
@akoeplinger could you check the root builder claim?

@jaykrell I think we should fix the bug if we behave differently under root instead of fixing the tests

jaykrell:
It is very clear the builder is running as root. There is checkbox checked indicating so.
This is so the monollvm .deb can be installed, though there should be a solution that allows that, w/o continuing as root.

I think there is no good solution here and the mono File tests are hacked up already.

As I understand, the Mono code to return the readonly bit look at owner/group/other readability, and calls access(). The owner/group/other readability can indicate that nobody can write, but then access() will say you can write, if you are root.

The mono code to set the readonly bit clear one or more of owner/group/other readability.
This makes the code readonly to most, but does not affect root's ability.

Explained another way:
chmod 444 1.txt
 sudo sh  "-c" "echo >> 1.txt"

works, even though:
chmod 444 1.txt
sh  "-c" "echo >> 1.txt"

does not.

baulig:
If you're only running as root to install a package, then why aren't you just using `sudo` to do so?  This is what I'm doing for the web-tests to auto-provision the bots from within a Jenkins job.

akoeplinger:
Jenkins was fixed to no longer run these tests as root but I still merged https://github.com/mono/mono/pull/8904 since root can still write to a file even if it's set readonly (this matches the corefx behavior).

PR #8904, opened 2018-05-28
Ignore GetAttributes_ReadOnly and Create_Path_ReadOnly when running as root, as LLVM CI does.

jaykrell:


jaykrell:
Workaround for https://github.com/mono/mono/issues/8903

lewurm:
I think the proper fix is to avoid running those tests as root.

jaykrell:
That's kinda what this is doing. Is there another way?

jaykrell:
Oh, you mean, fix Jenkins? Well, seeing if things work as root -- the rest does -- seems kinda reasonable.

akoeplinger:
Jenkins was fixed to no longer run these tests as root but I still think we should merge this since it makes sense.

PR #8910, opened 2018-05-28
[native crash] limit output of process mapping

lewurm:
less noise

### DIFF
--- a/mono/mini/mini-exceptions.c
+++ b/mono/mini/mini-exceptions.c
@@ -2895,8 +2895,10 @@ static void print_process_map (void)
 	}
 
 	mono_runtime_printf_err ("/proc/self/maps:");
+	const int max_lines = 25;
+	int i = 0;
 
-	while (fgets (line, sizeof (line), fp)) {
+	while (fgets (line, sizeof (line), fp) && i++ < max_lines) {
 		// strip newline
 		size_t len = strlen (line) - 1;
 		if (len >= 0 && line [len] == '\n')


mono/mini/mini-exceptions.c
```diff
@@ -2895,8 +2895,10 @@ static void print_process_map (void)
 	}
 
 	mono_runtime_printf_err ("/proc/self/maps:");
+	const int max_lines = 25;
+	int i = 0;
 
-	while (fgets (line, sizeof (line), fp)) {
+	while (fgets (line, sizeof (line), fp) && i++ < max_lines) {
 		// strip newline
 		size_t len = strlen (line) - 1;
```
jaykrell: This doesn't handle zero length correctly.

lewurm: why?

jaykrell: `size_t len = strlen("") - 1;`

lewurm: ah, is the problem `size_t` being unsigned?

jaykrell: Yes. As it probably should be. Negative numbers are not useful for sizes. Compare to 0 (or 1, perhaps) before subtracting. I realize this was preexisting code and not what you changing.

lewurm: fixed it. good catch 🙂

lewurm:
@monojenkins squash

PR #8911, opened 2018-05-28
[native crash] dump memory around instruction pointer

lewurm:
It can help to identify issues in generated code. We don't want to ship a disassembler, so we just do a hexdump. It looks like this:

```
Memory around native instruction pointer (0x10ca1b50e):
0x10ca1b4fe  48 c7 04 24 00 00 00 00 48 8b 44 24 10 48 8b f8  H..$....H.D$.H..
0x10ca1b50e  83 38 00 48 8d 64 24 00 49 bb 3a b5 a1 0c 01 00  .8.H.d$.I.:.....
0x10ca1b51e  00 00 41 ff d3 48 89 04 24 e9 00 00 00 00 48 8b  ..A..H..$.....H.
0x10ca1b52e  04 24 e9 00 00 00 00 48 83 c4 18 c3 e8 c1 1c 53  .$.....H.......S
```

Offline tools can be used to disassemble the code.

### DIFF
--- a/mono/mini/mini-exceptions.c
+++ b/mono/mini/mini-exceptions.c
@@ -2882,6 +2882,55 @@ print_stack_frame_to_string (StackFrameInfo *frame, MonoContext *ctx, gpointer d
 }
 
 #ifndef MONO_CROSS_COMPILE
+static gchar
+conv_ascii_char (gchar s)
+{
+	if (s < 0x20)
+		return '.';
+	if (s > 0x7e)
+		return '.';
+	return s;
+}
+
+static void
+xxd_mem (gpointer d, int len)
+{
+	guint8 *data = (guint8 *) d;
+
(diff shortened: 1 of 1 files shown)


lewurm:
@monojenkins build

luhenry:

luhenry:

luhenry:

jaykrell:
It'd be nice to pipe the bytes to objdump or something, if it is present.

lewurm:
we actually have infrastructure for that, see:
https://github.com/mono/mono/blob/663754bf468811ae82a28976240c98d12190e3c0/mono/mini/helpers.c#L136-L280
but for now I'm okay with using tools like https://onlinedisassembler.com/odaweb/

lewurm:
@monojenkins squash

PR #8914, opened 2018-05-28
Use only stat, not access to determine readonly, when root.

jaykrell:
This is closer to how Windows works, closer to how CoreCLR appears,
and should fix one of GetAttributes_ReadOnly and Create_Path_ReadOnly running as root (LLVM CI).
It should fix GetAttributes_ReadOnly.

"Fixing" Create_Path_ReadOnly would would require File.Open/Create to synthesize the readonly bit
with stat (like here) and implements its own denial instead of just calling underlying
open/create to decide, thereby denying access when the underlying system would not.

See https://github.com/mono/mono/issues/8903.

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -938,26 +938,31 @@ is_file_writable (struct stat *st, const gchar *path)
 	// OS X Finder "locked" or `ls -lO` "uchg".
 	// This only covers one of several cases where an OS X file could be unwritable through special flags.
 	if (st->st_flags & (UF_IMMUTABLE|SF_IMMUTABLE))
-		return 0;
+		return FALSE;
 #endif
 
 	/* Is it globally writable? */
 	if (st->st_mode & S_IWOTH)
-		return 1;
+		return TRUE;
+
+	uid_t const euid = geteuid ();
 
 	/* Am I the owner? */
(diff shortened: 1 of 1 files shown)


jaykrell:
See also https://github.com/mono/mono/pull/8904

mono/metadata/w32file-unix.c
```diff
@@ -953,16 +946,9 @@ is_file_writable (struct stat *st, const gchar *path)
...
 		return 1;
 
-	/* Fallback to using access(2). It's not ideal as it might not take into consideration euid/egid
-	 * but it's the only sane option we have on unix.
-	 */
-	MONO_ENTER_GC_SAFE;
-	ret = access (path, W_OK) == 0;
-	MONO_EXIT_GC_SAFE;
-	return ret;
+	return 0;
```
lewurm: can we do the fallback when the runtime isn't executed as root?

jaykrell:
Sure. It is all a bit nebulous. I’ve concluded the portable subset of file system features is tiny, and mono tries to exceed it.

jaykrell:
Does access make a difference? Maybe handles extended functionality on Unix?

jaykrell:
Better? This should bring “llvm” failures from 3 to 2.

akoeplinger:
Went with https://github.com/mono/mono/pull/8904 instead.

PR #8930, opened 2018-05-29
Fix Make syntax error causing spurious error ("ONO_PATH in MONO_PATH doesn't exist")

mcclure:
I tried building mono this morning from scratch on OS X 10.13.2. In the make scroll, I happened to notice this line:

    '/Users/mcc/work/g/other/mono/mcs/class/System.Web/ONO_PATH' in MONO_PATH doesn't exist or has wrong permissions.

I looked and this was caused by the use of `$$MONO_PATH` in `mcs/class/System.Web/Makefile`. I think this is a Automake-ism that someone accidentally pasted into a Makefile. I think Make interprets `$$M` as one token.

This patch fixes that Makefile plus `mcs/build/rules.make`, which contained the same error and was being included in a way which caused this warning message to appear in 15 other makefiles.

I think usually this patch will have no effect other than removing the ONO_PATH warning. On certain strange configurations (if the contents of MONO_PATH are important at build time) this patch might prevent a build error. Mostly the error was just bugging me


akoeplinger:
Should be fixed by https://github.com/mono/mono/pull/8933. Thanks for your work though!

Issue #8897, opened 2018-05-26
"Microsoft.Managed.Core.targets" is missing from roslyn-binaries


atsushieno:
... as a result of this problem, monodevelop fails to build projects.

## Steps to Reproduce

1. build and install mono to certain destination.
2. With the built mono, run monodevelop
3. open some project and build it.

### Current Behavior

Build fails due to lack of `/path/to/mono/msbuild/15.0/bin/Roslyn/Microsoft.Managed.Core.targets`

### Expected Behavior

Project just builds.

## On which platforms did you notice this

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:

92d9a3c

I changed external/roslyn-binaries checkout to the latest master and run `cp external/roslyn-binaries/Microsoft.Net.Compilers/Microsoft.Net.Compilers.2.8.2/tools/* /path/to/mono/msbuild/15.0/bin/Roslyn/` and then it worked again. I'm not sure if bumping roslyn-binaries is enough though.

marek-safar:
@radical are we missing symlink for the file?

PR #8931, opened 2018-05-29
[roslyn][msbuild] Copy the new Microsoft.Managed.Core.targets file in roslyn.

radical:
Fixes https://github.com/mono/mono/issues/8897 .


radical:
`2018-04` is still using roslyn 2.7.0, which doesn't have this file, so no backport required.

PR #8934, opened 2018-05-29
[corlib]: Add `System.HashCode` - revisiting #8822 and #8879.

baulig:
Enable `System.HashCode` again, cleanup the `.sources` files and fix the Windows build.

### DIFF
--- a/mcs/class/corlib/corert/PreAllocatedOverlapped.platformnotsupported.cs
+++ b/mcs/class/corlib/corert/PreAllocatedOverlapped.platformnotsupported.cs
@@ -1,0 +1,19 @@
+namespace System.Threading
+{
+	public sealed class PreAllocatedOverlapped : IDisposable
+	{
+		[CLSCompliant (false)]
+		public unsafe PreAllocatedOverlapped (IOCompletionCallback callback, object state, object pinData)
+		{
+			throw new PlatformNotSupportedException ();
+		}
+
+		public void Dispose ()
+		{
+		}
+
+		~PreAllocatedOverlapped ()
+		{
+		}
+	}
+}
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,5 @@
 <!--
-If you are new to the project get yourself familiar with http://mono-project.com/community/bugs/make-a-good-bug-report before filling the issue
+If you are new to the project get yourself familiar with https://www.mono-project.com/community/bugs/make-a-good-bug-report/ before filling the issue
 -->
 
 ## Steps to Reproduce
--- a/mcs/build/gensources.cs
+++ b/mcs/build/gensources.cs
@@ -286,6 +286,15 @@ public class SourcesParser {
             return state.Result;
         }
 
+        testPath = Path.Combine (libraryDirectory, $"{hostPlatform}_{libraryName}");
+        ok = TryParseSingleFile (state, testPath + ".sources", false);
+        TryParseSingleFile (state, testPath + ".exclude.sources", true);
+
+        if (ok) {
+            PrintSummary (state);
+            return state.Result;
+        }
+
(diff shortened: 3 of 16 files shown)


mcs/class/corlib/corert/PreAllocatedOverlapped.platformnotsupported.cs
```diff
@@ -0,0 +1,31 @@
+namespace System.Threading
+{
+	public sealed class PreAllocatedOverlapped : IDisposable
+	{
+		static PreAllocatedOverlapped ()
```
marek-safar: static ctor is not needed

mcs/class/corlib/corert/PreAllocatedOverlapped.platformnotsupported.cs
```diff
@@ -0,0 +1,31 @@
...
+
+		[CLSCompliant (false)]
+		public unsafe PreAllocatedOverlapped (IOCompletionCallback callback, object state, object pinData)
+		{
+			throw new PlatformNotSupportedException ();
+		}
+
+		public void Dispose ()
+		{
+			GC.SuppressFinalize (this);
```
marek-safar: Should be just no-op

mcs/class/corlib/corert/PreAllocatedOverlapped.platformnotsupported.cs
```diff
@@ -0,0 +1,31 @@
...
+		}
+
+		~PreAllocatedOverlapped ()
+		{
+			//
```
marek-safar: Should be no-op

mcs/class/corlib/wasm_corlib.dll.sources
```diff
@@ -0,0 +1,5 @@
+#include corlib.dll.sources
+
+corert/ThreadPoolBoundHandle.platformnotsupported.cs
+corert/PreAllocatedOverlapped.platformnotsupported.cs
```
akoeplinger: hmm, I think it might be easier to add these to corlib.dll.sources instead and exclude them just in win32_net_4_x_corlib.dll.exclude.sources and winaot_corlib.dll.exclude.sources. That way we shouldn't have to add .sources for all the profiles.

baulig: Good idea.  I really didn't like the addition of these one million additional .sources files.

baulig:
Dismissed stale reviews.  Everything should be fixed now and this passed a `--with-runtime-preset=all` build locally.

Please review again.

baulig:
The api-diff removes the constructor `~ThreadPoolBoundHandle() { }`, but I checked with the [reference](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Overlapped/ref/System.Threading.Overlapped.cs) and [ThreadPoolBoundHandle](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Overlapped/ref/System.Threading.Overlapped.cs#L27) does not have a destructor in there.

The [PreAllocatedOverlapped](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Overlapped/ref/System.Threading.Overlapped.cs#L21) class also does not have a destructor there, but our `corert/PreAllocatedOverlapped.platformnotsupported.cs` does.

Do you have any advise for handling situations like that?

marek-safar:

marek-safar:
@baulig I merge it to unblock dependant work but do follow up PR with
- Add corefx tests for Hashcode
- Enable the type in `Facades/System.Runtime/TypeForwarders.netcore.cs`
- Why is the type in desktop profile only? Should we move it do corlib.sources

baulig:
Moving it into `corlib.dll.sources` to enable it on all profiles was actually what I originally intended, but it required additional sources work and the `gensources.cs` fix to land first.

This is now in a follow-up PR: https://github.com/mono/mono/pull/8984.

PR #8937, opened 2018-05-29
[sdb][wasm][debugger-engine] Cleanup & extract enough that it's possible to move the breakpoint code over.

kumpera:
Which this PR, we're almost 1/2 done. There are quite a few more hard bit ahead once we start to deal with single-stepping.

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -115,10 +115,10 @@ typedef struct {
 
 typedef struct
 {
+	//Must be the first field to ensure pointer equivalence
+	DbgEngineStackFrame de;
 	int id;
-	guint32 il_offset, native_offset;
-	MonoDomain *domain;
-	MonoMethod *method;
+	guint32 il_offset;
 	/*
 	 * If method is gshared, this is the actual instance, otherwise this is equal to
 	 * method.
@@ -131,7 +131,6 @@ typedef struct
 	MonoMethod *api_method;
 	MonoContext ctx;
 	MonoDebugMethodJitInfo *jit;
-	MonoJitInfo *ji;
 	MonoInterpFrameHandle interp_frame;
 	gpointer frame_addr;
 	int flags;
@@ -3068,13 +3067,15 @@ process_frame (StackFrameInfo *info, MonoContext *ctx, gpointer user_data)
 	}
 
 	frame = g_new0 (StackFrame, 1);
-	frame->method = method;
+	frame->de.ji = info->ji;
+	frame->de.domain = info->domain;
+	frame->de.method = method;
+	frame->de.native_offset = info->native_offset;
+
 	frame->actual_method = actual_method;
 	frame->api_method = api_method;
 	frame->il_offset = info->il_offset;
-	frame->native_offset = info->native_offset;
 	frame->flags = flags;
-	frame->ji = info->ji;
 	frame->interp_frame = info->interp_frame;
 	frame->frame_addr = info->frame_addr;
 	if (info->reg_locations)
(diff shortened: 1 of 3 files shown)


mono/mini/debugger-agent.c
```diff
@@ -663,25 +653,16 @@ static void jit_failed (MonoProfiler *prof, MonoMethod *method);
...
 static void clear_types_for_assembly (MonoAssembly *assembly);
 
-static void clear_breakpoints_for_domain (MonoDomain *domain);
-
 static void process_profiler_event (EventKind event, gpointer arg);
 
 /* Submodule init/cleanup */
-static void breakpoints_init (void);
-static void breakpoints_cleanup (void);
+static void event_requests_cleanup (void);
```
jaykrell: Function name at start of line please for grep ^foo.

mono/mini/debugger-agent.c
```diff
@@ -2714,6 +2692,21 @@ process_suspend (DebuggerTlsData *tls, MonoContext *ctx)
...
 }
 
+
+/* Conditionally call process_suspend depending oh the current state */
+static gboolean
+try_process_suspend (DebuggerTlsData *tls, MonoContext *ctx)
+{
+	if (suspend_count > 0) {
+		/* Fastpath during invokes, see in process_suspend () */
+		if (suspend_count - tls->resume_count == 0)
```
jaykrell: How about a == b instead of a - b == 0?
Do you need to check if one is greater than the other, or it works in either direction?

mono/mini/debugger-engine.h
```diff
@@ -147,11 +147,35 @@ typedef struct {
...
 void mono_de_init (void);
 void mono_de_cleanup (void);
+void mono_de_set_log_level (int level, FILE *file);
 
 //locking - we expose the lock object from the debugging engine to ensure we keep the same locking semantics of sdb.
 void mono_de_lock (void);
 void mono_de_unlock (void);
 
+// domain handling
+void mono_de_foreach_domain (GHFunc func, gpointer user_data);
```
jaykrell: Prefer function name at start of line for grep ^foo.

mono/mini/debugger-agent.c
kumpera: This is our coding convention, not sure I want to deviate from it.

mono/mini/debugger-agent.c
kumpera: This was just code motion, I wanted to make it as identical as possible to facilitate review.

kumpera:

kumpera:

mono/mini/debugger-agent.c
```diff
@@ -2734,7 +2727,7 @@ suspend_vm (void)
 
 	if (suspend_count == 1) {
 		// FIXME: Is it safe to call this inside the lock ?
-		start_single_stepping ();
+		mono_de_start_single_stepping ();
```
alexanderkyte: mono_germany_start_single_stepping? 
Can we use dbg or something, rather than "de"?

alexanderkyte: mono_germany_start_single_stepping?
Can we use dbg rather than "de"? We don't use "de" to mean the debugger anywhere else.

jaykrell: Or even debug or debugger or dbgeng. I also don't like "de", but I didn't speak up.
Some projects to have usually two character prefixes (Ex, Ps, Ke, Mm, Io) but it seemed obscure to me..but I realize what it is..probably we all do..

jaykrell: I find "engine" a dubious part of any name. What do you call the non-engine parts?
Thing and ThingEngine? Thing and ThingCore? ThingFoo and ThingBar, but Foo and Bar are hard to name, so Foo is empty and Bar is Engine or Core?
.NET Engine?
d and de?
Dbg and DbgEng?

PR #8940, opened 2018-05-30
[corlib]: Use 'RandomNumberGenerator' from CoreFX.

baulig:
* Add new mono_btls_get_random_bytes() function.

* corefx/RandomNumberGeneratorImplementation.OSX.cs: copy from CoreFX
  and P/Invoke directly instead of calling a native wrapper function.

* corefx/corefx/RandomNumberGeneratorImplementation.Unix.cs:
  CoreFX uses OpenSSL here, but the same function also exists in boringssl.
  Use a custom implementation P/Invoking into the new mono_btls_get_random_bytes().

* On Windows, we use the native CoreFX implementation.


baulig:
I am currently reworking this, but it depends on the `HashCode` and `Guid` changes.

baulig:
Closing as I am going to completely rework this and have a bunch of broken stuff in the work branch that I'd like to push on top of this.

PR #8942, opened 2018-05-30
[Coop] Convert System.Buffer.

jaykrell:


mono/metadata/icall.c
```diff
@@ -6438,8 +6438,9 @@ ves_icall_System_Delegate_GetVirtualMethod_internal (MonoDelegateHandle delegate
 /* System.Buffer */
 
 static inline gint32 
-mono_array_get_byte_length (MonoArray *array)
+mono_array_get_byte_length (MonoArrayHandle array_handle)
```
Therzok: No MonoError?

jaykrell: Not currently needed.

luhenry:
@marek-safar how perf-sensitive in System.Buffer? This seems like a case where you don't want coop handles because of the perf penalty.

marek-safar:
Any type in mscorlib System namespace is perf sensitive. Secondly, this type is not yet ported from corefx so be prepared for futher work when it changes.

kumpera:
In the case of System.Buffer it makes sense to have a managed fast path that does some memcpy in managed before jumping to native anyways - this is not a suggestion to make such a change.

PR #8945, opened 2018-05-30
[Coop] Partially convert System.Threading.Monitor.

jaykrell:


jaykrell:
Note that the entire "Pulse" idea is buggy and deprecated in Win32, so probably here/CLR/C# as well. That is some of this. Maybe move the rest to C#?

jaykrell:
Can the little readers be in unsafe C#?

jaykrell:
@kumpera indicated that fast paths make the perf of at least the small functions here not matter.
Otherwise I really wonder if/how to move them to C#. The obvious of casting object in unsafe to a struct* didn't compile. That would seem, with some syntax, a bridge toward writing more runtime in C#. I'm/I'd-be surprised if there is a way. Maybe taking ref of first data member and subtracting offset but that'd seem unnecessarily fragile (granted, all options are fragile).

PR #8947, opened 2018-05-30
[Coop] Partially convert COM interop.

jaykrell:


jaykrell:
@monojenkins build  failed

jaykrell:
Windows Cominterop test fails with access violation, so no.

jaykrell:
@monojenkins build failed

mono/metadata/cominterop.c
```diff
@@ -1752,31 +1737,17 @@ ves_icall_System_Runtime_InteropServices_Marshal_GetComSlotForMethodInfoInternal
...
-		return NULL;
-	obj = mono_object_new_alloc_specific_checked (vtable, error);
-	if (mono_error_set_pending_exception (error))
-		return NULL;
+	MonoDomain * const domain = MONO_HANDLE_DOMAIN (ref_type);
+	MonoType * const type = MONO_HANDLE_GETVAL (ref_type, type);
+	MonoClass * const klass = mono_class_from_mono_type (type);
 
-	return obj;
+	// mono_object_new_handle => mono_object_new_by_vtable handle
```
luhenry: You completely ignored the comment that explicitly warns against using `mono_object_new` directly. Is this change warranted and correct?

mono/metadata/cominterop.c
```diff
@@ -1850,22 +1821,17 @@ mono_cominterop_release_all_rcws (void)
...
+	if (!mono_class_init_checked (klass, error))
 		return NULL;
-	}
 
-	gpointer itf = cominterop_get_interface_checked (obj, klass, error);
-	if (throw_exception)
-		mono_error_set_pending_exception (error);
-	else
-		mono_error_cleanup (error);
+	ERROR_DECL (error_ignored);
```
luhenry: Instead of creating a new error, can't you more simply conditionally cleanup the existing `error`? Something like the following would be equivalent to what you did without creating a temporary error.

```
gpointer const itf = cominterop_get_interface_checked (MONO_HANDLE_RAW (obj), klass, error);
if (!throw_exception)
	mono_error_cleanup (error);
```

mono/metadata/cominterop.c
jaykrell: No I didn't but I agree it is confusing.
When we converted to coop creating functions we updated the names/meanings.
`mono_object_new` is deprecated and external-only/mostly.
There is no significant use for this, so "regular" `mono_object_new_handle`  has the new better meaning. If this is  confusing, which I agree it could be, perhaps the handle functions should be more mechanically named and resemble the non-handle ones more closely?

I kinda pushed for that but was more than once advised to pick better names, instead of just sticking "handle" somewhere. (exception_new vs. exception_get or something).

mono/metadata/cominterop.c
jaykrell: Maybe but not clearly/completely. I think you'd have to reinit. Which is a rare pattern, perhaps to be avoided?

It seems slightly better to perturb ignored extra state on the side, if intending to ignore, instead of mutating the "real" state temporarily, only to restore it.
Basically to reduce visible-or-even-temporary-maybe-not-visible side effects in general, if perf is ok.
It isn't clear cut though.

There is also the off chance of:
foo(error);
bar(ignore, error); // harmless to proceed a little?
if (ok(error)) // error from foo, unperturbed by bar.

Though given the scoped nature of icalls, this might be impossible to construct.

jaykrell: As to  what is simpler is also subjective. What I have isn't complicated (subjectively) and is simpler to reason about, due to the reduced side effects.
What you suggest does save stack and perhaps perf, indeed.

mono/metadata/cominterop.c
jaykrell: The renaming is really confusing I agree. Maybe we should revisit the handle allocating names first, line them up with the rawpointer names, as ugly as they are?

jaykrell: To be clear, I am not even asserting that I got right. But I did not ignore the comment and I tried to maintain unchanged semantics, except for pointer vs. handle. If I failed, and your question, is perhaps testament that cleaning up the names vs. "just append handle" is not warranted.

jaykrell:
Relevant test failures on Windows still.

jaykrell:
I think I read the comment backwards. It was actually avoiding the extra work, not needing it.

jaykrell:

jaykrell:

jaykrell:
Passes CI. Approve?

PR #8956, opened 2018-05-30
[corlib] Remove `implicit operator String (ReadOnlySpan<char> value)`

EgorBo:
This operator allows to cast ROS<char> to string but it's not used anywhere
```
ReadOnlySpan<char> span = ...
string str = span;
```
.NET Core also doesn't have it and it can in theory cause "ambiguous call" compilation errors.


marek-safar:
@monojenkins commit apidiff

PR #8960, opened 2018-05-30
 Refactor ves_icall_System_Threading_Monitor_Monitor_pulse and pulse_all.

jaykrell:
These  functions are the same except for if vs. while, so can be one function with a boolean parameter. (and unused func in debugging)

### DIFF
--- a/mono/metadata/monitor.c
+++ b/mono/metadata/monitor.c
@@ -611,7 +611,7 @@ mono_object_hash (MonoObject* obj)
 		mono_monitor_inflate (obj);
 		lw.sync = obj->synchronisation;
 	} else if (lock_word_is_flat (lw)) {
-		int id = mono_thread_info_get_small_id ();
+		int const id = mono_thread_info_get_small_id ();
 		if (lock_word_get_owner (lw) == id)
 			mono_monitor_inflate_owned (obj, id);
 		else
@@ -967,7 +967,7 @@ static inline gint32
 mono_monitor_try_enter_internal (MonoObject *obj, guint32 ms, gboolean allow_interruption)
 {
 	LockWord lw;
-	int id = mono_thread_info_get_small_id ();
+	int const id = mono_thread_info_get_small_id ();
 
 	LOCK_DEBUG (g_message("%s: (%d) Trying to lock object %p (%d ms)", __func__, id, obj, ms));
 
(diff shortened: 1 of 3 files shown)


mono/metadata/monitor.c
```diff
@@ -1292,46 +1292,28 @@ ves_icall_System_Threading_Monitor_Monitor_pulse (MonoObject *obj)
...
+}
+
 void
 ves_icall_System_Threading_Monitor_Monitor_pulse_all (MonoObject *obj)
 {
-	int id;
-	LockWord lw;
-	MonoThreadsSync *mon;
-	
-	LOCK_DEBUG (g_message("%s: (%d) Pulsing all %p", __func__, mono_thread_info_get_small_id (), obj));
```
jaykrell: "all" in the format string is another divergence but redundant with func.

jaykrell:
@monojenkins build failed

mono/metadata/monitor.c
```diff
@@ -1292,46 +1292,29 @@ ves_icall_System_Threading_Monitor_Monitor_pulse (MonoObject *obj)
...
+again:
 	if (mon->wait_list != NULL) {
-		LOCK_DEBUG (g_message ("%s: (%d) signalling and dequeuing handle %p", __func__, mono_thread_info_get_small_id (), mon->wait_list->data));
-	
+		LOCK_DEBUG (g_message ("%s: (%d) signalling and dequeuing handle %p", func, mono_thread_info_get_small_id (), mon->wait_list->data));
+
 		mono_w32event_set (mon->wait_list->data);
 		mon->wait_list = g_slist_remove (mon->wait_list, mon->wait_list->data);
+		if (all)
+			goto again;
```
lewurm: can we use a loop instead of `goto`?

jaykrell: Yes but it isn’t pretty. Something like do while all and not null.

jaykrell: Or while true and break if null or not all.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build  failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build Linux x64

PR #8981, opened 2018-05-31
Make the sendfile with 6 args check use AC_TRY_LINK

### LINKED ISSUE #8979: configure script incorrectly assumes functions are available with some conditions
Working on the [CoreFX PAL fixes for AIX](https://github.com/dotnet/corefx/pull/30012), when I noticed that it was incorrectly assuming that sendfile was available, confusing CoreFX. Example, with a comparison against a case where it wasn't successful and thus caught by the linker:

```
configure:26246: checking for sendfile with 6 arguments
configure:26266: gcc -c -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -g -O2 -std=gnu99 -fno-strict-aliasing -fwrapv -DMONO_DLL_EXPORT -Wno-unused-but-set-variable -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -D__PASE__ -I/opt/mono/include -maix64 -DGC_AIX_THREADS -D_ALL_SOURCE -D_THREAD_SAFE -D_LARGE_FILES -D_REENTRANT -g  conftest.c >&5
conftest.c:251:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 main ()
 ^~~~
conftest.c: In function 'main':
conftest.c:254:16: warning: implicit declaration of function 'sendfile' [-Wimplicit-function-declaration]
   int result = sendfile(0, 0, 0, NULL, NULL, 0);
                ^~~~~~~~
conftest.c:254:3: warning: nested extern declaration of 'sendfile' [-Wnested-externs]
   int result = sendfile(0, 0, 0, NULL, NULL, 0);
   ^~~
conftest.c:254:7: warning: unused variable 'result' [-Wunused-variable]
   int result = sendfile(0, 0, 0, NULL, NULL, 0);
       ^~~~~~
configure:26266: $? = 0
configure:26268: result: yes
configure:26282: checking for fcopyfile
configure:26282: gcc -o conftest -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -g -O2 -std=gnu99 -fno-strict-aliasing -fwrapv -DMONO_DLL_EXPORT -Wno-unused-but-set-variable -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -D__PASE__ -I/opt/mono/include -maix64 -DGC_AIX_THREADS -D_ALL_SOURCE -D_THREAD_SAFE -D_LARGE_FILES -D_REENTRANT -g  -maix64 conftest.c -lm   -lpthread >&5
conftest.c:268:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 char fcopyfile ();
 ^~~~
conftest.c:277:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 main ()
 ^~~~
ld: 0711-317 ERROR: Undefined symbol: .fcopyfile
ld: 0711-345 Use the -bloadmap or -bnoquiet option to obtain more information.
collect2: error: ld returned 8 exit status
configure:26282: $? = 1
configure: failed program was:
[...]
```

The difference between the two; one checks as a code fragment while the other lets autotools handle it. In source, on configure.ac+3072:

```
        AC_MSG_CHECKING(for sendfile with 6 arguments)
        AC_TRY_COMPILE([
                #include <stdlib.h>
                #include <sys/types.h>
                #include <sys/socket.h>
                #include <sys/uio.h>
        ], [
                int result = sendfile(0, 0, 0, NULL, NULL, 0);
        ],[
                AC_MSG_RESULT(yes)
                AC_DEFINE(HAVE_SENDFILE_6, 1, [sendfile with 6 arguments])
        ], [
                AC_MSG_RESULT(no)
        ])

        AC_CHECK_FUNC(fcopyfile,     [AC_DEFINE(HAVE_FCOPYFILE, 1, [fcopyfile])])
```

It seems that the linker catches it on the `AC_CHECK_FUNC` case, whereas the `AC_TRY_COMPILE` case merely makes an object without linking with the `-c` option passed to GCC. (FWIW, the "4 arguments" case for detecting sendfile for CoreFX fails because of a missing `sendfile.h` header - basically again, that case only handles compile errors.)

My guess for possible solutions are either:

1. Make `AC_TRY_COMPILE` try linking

2. Pass the compiler flag that makes GCC error out on implied definitions; though there *are* a few cases where fixincludes included with GCC has mangled system headers to not include something in headers that exists in libc, but I believe I caught them all already and explicitly make explicit declarations whenever GCC was being stupid

I'm not an autotools expert, so IDK how to implement either of these.

NattyNarwhal:
This lets the linker catch any missing symbols, and fixes a problem
I had building AIX with CoreFX PAL. Fixes #8979, though there may
be more cases like this.

dotnet/corefx#30012 is blocking on this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3070,7 +3070,7 @@ if test x$host_win32 = xno; then
 	])
 
 	AC_MSG_CHECKING(for sendfile with 6 arguments)
-	AC_TRY_COMPILE([
+	AC_TRY_LINK([
 		#include <stdlib.h>
 		#include <sys/types.h>
 		#include <sys/socket.h>


configure.ac
```diff
@@ -3070,7 +3070,7 @@ if test x$host_win32 = xno; then
 	])
 
 	AC_MSG_CHECKING(for sendfile with 6 arguments)
-	AC_TRY_COMPILE([
+	AC_TRY_LINK([
```
jaykrell: How can you tell 4 parameter vs. 6 parameter? They will each compile and each link. You should compile the test program as C++ to decide. See what I did with size_t.

NattyNarwhal: The four-argument test will fail on not finding the sendfile header. The six-argument test will *compile* due to implied declaration, (it uses headers found on mostly any system) but linking will make it fail, which is what we'd want.

jaykrell: Yeah I see that now..on Mac..	#include <sys/sendfile.h>
 not found. Wierd.

But Linux appears to have sys/sendfile and 4 parameters.

But then, the thing being compiled..the number of parameters is irrelevant. You can just take the address of sendfile in a global.

It is really testing if sys/sendfile exists, and if so, assume it has 4 parameters (if it doesn't, it'll just warn and succeed).
Or else, does sendfile link, in which case, assume 6 parameters.

There isn't really a check for number of parameters. It is just that sys/sendfile.h implies 4 parameters. 

It isn't really doing  what it looks like.

jaykrell: Suggest at least comments. That the second check will always compile, and if it links, the signature is merely assumed. There is no actual checking of the signature.

And that the first test will compile if sys/sendfile.h exists, no matter what is the signature or existance of sendfile.

NattyNarwhal: FWIW, if sendfile-with-4 is caught, the PAL defines IIRC wont fall back to sendfile-with-6; either way, this patch wouldn't change *that* behaviour.

jaykrell: Indeed, it was already wonky. You always get one of the symbols, sometimes both, must check them in order, this doesn't change it. It is a minor bug that it defines both, that we must workaround.

PR #8984, opened 2018-05-31
[corefx] Enable HashCode on all profiles.

baulig:
* HashCode.cs has been moved into `corlib.dll.sources`.

* Add `PLATFORMS = darwin linux win32` to the following profiles:
  basic, testing_aot_full, testing_aot_hybrid.

* Add fallback implementation using `RNGCryptoServiceProvider` (like our [`Guid.FastNewGuidArray`](https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Guid.cs#L66)) and use in on profiles which don't have a native implementation in CoreFX.

* Add per-platform sources which are now used by all profiles which
  do not have an explicit profile-platform sources file:
  - `darwin_corlib.dll.sources`
  - `linux_corlib.dll.sources`
  - `win32_corlib.dll.sources`

* Add new `monodroid_corlib.dll.sources`.

* Add new `xammac_net_4_5_corlib.dll.sources`.

* Remove `testing_aot_full_corlib_test.dll.sources`.

* Add the platform-specific non-cryptographic RNG to all profiles.


mcs/build/profiles/orbis.make
```diff
@@ -5,6 +5,8 @@ BOOTSTRAP_PROFILE = build
 BOOTSTRAP_MCS = MONO_PATH="$(topdir)/class/lib/$(BOOTSTRAP_PROFILE)$(PLATFORM_PATH_SEPARATOR)$$MONO_PATH" $(INTERNAL_CSC)
 MCS = $(BOOTSTRAP_MCS)
 
+PLATFORMS = darwin linux win32
```
luhenry: This doesn't make any sense: `orbis` cannot run on anything else than `orbis`.

mcs/build/profiles/basic.make
```diff
@@ -9,6 +9,8 @@ use_monolite := $(wildcard $(monolite_flag))
 
 MONOLITE_MSCORLIB = $(monolite_path)/mscorlib.dll
 
+PLATFORMS = darwin linux win32
```
luhenry: This is not how you want to fix it, look at https://github.com/mono/mono/compare/master...luhenry:corefx-System.Net.NetworkInformation#diff-8a17f52e0062026d1a9555a3bae175e8R298 for the fix.

mcs/build/profiles/unreal.make
```diff
@@ -5,6 +5,8 @@ BOOTSTRAP_PROFILE = build
 BOOTSTRAP_MCS = MONO_PATH="$(topdir)/class/lib/$(BOOTSTRAP_PROFILE)$(PLATFORM_PATH_SEPARATOR)$$MONO_PATH" $(INTERNAL_CSC)
 MCS = $(BOOTSTRAP_MCS)
 
+PLATFORMS = darwin linux win32
```
luhenry: Same as `orbis`, it doesn't make any sense.

mcs/build/profiles/wasm.make
```diff
@@ -5,6 +5,8 @@ BOOTSTRAP_PROFILE = build
 BOOTSTRAP_MCS = MONO_PATH="$(topdir)/class/lib/$(BOOTSTRAP_PROFILE)$(PLATFORM_PATH_SEPARATOR)$$MONO_PATH" $(INTERNAL_CSC)
 MCS = $(BOOTSTRAP_MCS)
 
+PLATFORMS = darwin linux win32
```
luhenry: Same as `orbis`, doesn't make any sense.

mcs/build/profiles/basic.make
baulig: That link doesn't work.

luhenry: [This link](https://github.com/mono/mono/compare/master...luhenry:corefx-System.Net.NetworkInformation#diff-8a17f52e0062026d1a9555a3bae175e8R298) should work

mcs/class/corlib/monotouch_corlib.dll.sources
```diff
@@ -13,3 +13,6 @@ System.Text/EncodingHelper.MonoTouch.cs
 System/Environment.iOS.cs
 System/Guid.MonoTouch.cs
 System/NotSupportedException.iOS.cs
+
+../../../external/corefx/src/Common/src/CoreLib/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
```
luhenry: System.Native is not available on iOS

mcs/class/corlib/monodroid_corlib.dll.sources
```diff
@@ -0,0 +1,4 @@
+#include corlib.dll.sources
+
+../../../external/corefx/src/Common/src/CoreLib/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
```
luhenry: System.Native is not available on Android

mcs/class/corlib/corefx/Interop.GetRandomBytes.cs
```diff
@@ -0,0 +1,29 @@
...
+				if (_rng == null)
+					_rng = new RNGCryptoServiceProvider ();
+				_rng.GetBytes (buffer);
+			}
+		}
+	}
+
+	internal static unsafe void GetRandomBytes (byte* buffer, int length)
+	{
+		var bytes = new byte[length];
```
luhenry: If we are taking this approach, we should not have to allocate here, but instead pass the byte* to the runtime so we populate it without copying.

mcs/class/corlib/monodroid_corlib.dll.sources
marek-safar: @luhenry what is blocking us to include it?

mcs/class/corlib/orbis_corlib.dll.sources
```diff
@@ -0,0 +1,3 @@
+#include corlib.dll.sources
+
+corefx/Interop.GetRandomBytes.cs
```
marek-safar: This should be just unix implementation

mcs/build/profiles/testing_aot_full.make
```diff
@@ -5,6 +5,8 @@ BOOTSTRAP_PROFILE = build
 BOOTSTRAP_MCS = MONO_PATH="$(topdir)/class/lib/$(BOOTSTRAP_PROFILE)$(PLATFORM_PATH_SEPARATOR)$$MONO_PATH" $(INTERNAL_CSC)
 MCS = $(BOOTSTRAP_MCS)
 
+PLATFORMS = darwin linux win32
```
luhenry: For that, you also want to use [this](https://github.com/mono/mono/compare/master...luhenry:corefx-System.Net.NetworkInformation#diff-8a17f52e0062026d1a9555a3bae175e8R298) since `PLATFORMS` is useful when you want to cross-compile to another platform, and there is no point of cross-compiling `testing_aot_full` since it's only for testing.

mcs/build/profiles/testing_aot_hybrid.make
```diff
@@ -5,6 +5,8 @@ BOOTSTRAP_PROFILE = build
 BOOTSTRAP_MCS = MONO_PATH="$(topdir)/class/lib/$(BOOTSTRAP_PROFILE)$(PLATFORM_PATH_SEPARATOR)$$MONO_PATH" $(INTERNAL_CSC)
 MCS = $(BOOTSTRAP_MCS)
 
+PLATFORMS = darwin linux win32
```
luhenry: For that, you also want to use [this](https://github.com/mono/mono/compare/master...luhenry:corefx-System.Net.NetworkInformation#diff-8a17f52e0062026d1a9555a3bae175e8R298) since `PLATFORMS` is useful when you want to cross-compile to another platform, and there is no point of cross-compiling `testing_aot_hybrid` since it's only for testing.

mcs/class/corlib/monodroid_corlib.dll.sources
luhenry: Porting it and making sure it does not break Xamarin.Android?

PR #8985, opened 2018-05-31
Rework genproj to use gensources to build sources list for each profile and host platform

### LINKED ISSUE #8046: Genproj includes only Linux platform C# specific files
We need to include platform-specific files for example `external/corefx/src/Common/src/Interop/OSX/Interop.Libraries.cs` from `darwin_net_4_x_System.Data.dll.sources` to build correctly on non-Linux platforms

kg:
Fixes #8046


mcs/build/executable.make
```diff
@@ -192,6 +192,7 @@ csproj-local:
 	(echo $(is_boot); \
 	echo $(USE_MCS_FLAGS) $(LIBRARY_FLAGS) $(LIB_MCS_FLAGS) $(patsubst %,-r:%,$(LIB_REFS)); \
 	echo $(PROGRAM); \
```
luhenry: Do we need to duplicate it?

msvc/scripts/Makefile
```diff
@@ -1,8 +1,8 @@
 all: genproj.exe prepare.exe
 	mono --debug genproj.exe
 
-genproj.exe: genproj.cs
-	mcs -debug:portable -g genproj.cs -r:System.Xml.Linq
+genproj.exe: genproj.cs ../../mcs/build/gensources.cs
+	mcs -debug:portable genproj.cs ../../mcs/build/gensources.cs -main:Driver -r:System.Xml.Linq -r:System.dll -r:System.Core.dll
```
luhenry: Should we use csc instead?

kg:

kg:
The one other big problem I forgot to mention is this:
screenshot 2018-06-28 10 48 40
I think it's a VS bug as I only started seeing it after an update. It goes away if you Reload jay, and I haven't touched the vcxproj, so it's an annoyance at most. I hope I can fix it in a follow-up PR somehow though.

marek-safar:
@monojenkins commit csproj

akoeplinger:
Awesome work, thank you @kg !

PR #8989, opened 2018-05-31
Fix race condition in ves_icall_Mono_Interop_ComInteropProxy_AddProxy.

jaykrell:
Multiple threads can see the null, then proceed to lock and create new one.
The check for null needs to be under the lock, either duplicated, or before and within.
Since we take the lock always unconditionally anyway soon afterward, just do it within.


luhenry:

luhenry:

luhenry:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins Windows i386

PR #8969, opened 2018-05-31
[bitcode] Fix building -with-runtime-preset=bitcode.

jaykrell:
Not clear if this is correct, but let build progress.

PR #8990, opened 2018-06-01
[bitcode] Pipe llvm-dis to clang -xir to slowly fix bitcode versioning problem.

jaykrell:


vargaz:
I'm not sure we need this right now with the current xcode (9.3), but we could keep it in an ifdef #0 just in case.

mono/profiler/Makefile.am
```diff
@@ -85,8 +85,17 @@ else
 if HOST_TIZEN
 libmono_dep = $(top_builddir)/mono/mini/$(LIBMONO_LA)
 else
+# BITCODE => -no-undefined
```
luhenry: This change is already part of https://github.com/mono/mono/pull/8969, please remove

mono/mini/aot-compiler.c
```diff
@@ -9042,12 +9042,18 @@ emit_llvm_file (MonoAotCompile *acfg)
 	if (acfg->aot_opts.llvm_only) {
 		/* Use the stock clang from xcode */
 		// FIXME: arch
-		command = g_strdup_printf ("clang++ -fexceptions -march=x86-64 -fpic -msse -msse2 -msse3 -msse4 -O2 -fno-optimize-sibling-calls -Wno-override-module -c -o \"%s\" \"%s.opt.bc\"", acfg->llvm_ofile, acfg->tmpbasename);
+		// Only output unsightly quotes for the unusual case of spaces in paths.
```
luhenry: Why does it hurt to always have the `"`? That seems like an unnecessary "optimization".

jaykrell: It is ugly and unnecessarily verbose when viewing logs & copy/pasting their commands.

mono/profiler/Makefile.am
jaykrell: Right. I needed something that works more, will be fixed by merging.

jaykrell: Llvm-dis “needed” on regular Sierra. Maybe not on HighSierra. Could also autoconf but ok if 0, once I confirm highsierra works for me.

jaykrell:
Confirmed newer OS and then udpating Xcode, it is working.
Maybe autoconf it or something? Or if it fails, fallback to the dis path? I'll PR that. Not pretty, I realize.

jaykrell:
@monojenkins build failed

mono/mini/aot-compiler.c
jaykrell: Note that the code was already using wrap_path for similar reason, though it was leaking everything it returned.
Another idea is to use wrap_path, but it have it store all its results for freeing.

vargaz:
If this only fails with older xcodes, then we can ignore it. ios development requires the newest xcode always.

jaykrell:
Not just ios development, but Mac development. Granted, Mac bitcode.

jaykrell:
I'd like to take the automatic workaround, where it  first tries what it always did, and if that fails, uses llvm-dis | clang -xir -.
And remove all the unnecessary quotes.

jaykrell:
Granted, the llvm-dis might still fail. The code doesn't know the exact error. It is difficult to.

vargaz:
I think we should keep this code commented out, so we can enable it in the future if it ever becomes necessary.

jaykrell:
I like it there as a more active fallback but let me sleep on it, and will probably comment it out tomorrow.
A drawback of my position is that it'd just silently start happening w/o notice and we'd have a slower compile. As well, it might cover up other fixable problems, like if the bitcode just went slightly wrong somehow.
As to the user population using bitcode + old xcode + llvm6.0, perhaps that was only me.
And understood, it could come up again and this might save us in the future.

alexanderkyte:
Can we document this somewhere so it's not lost?

PR #9005, opened 2018-06-02
Fix shutdown hangs e.g. thread6 and possibly Alpine/muslc problem.

jaykrell:
Prior to 5f5c5e97a08f7086d7c18af37352c4a03dc4c0d1, threadpool workers were in an interruptable wait, of random duration. There was coordination I believe with shutdown and that got lost.

The performance of shutting down the runtime became therefore random.
This is likely the cause of the AlpineLinux/muslc slowdown,  and can be seen more  easily by running mono/tests/thread6.exe. It used to run quickly, now randomly slowly.

This attempts to wake the workers, but still seems racy.
I am skeptical that there is an adequate replacement for an interruptable wait.
Or that we need in general to wait on multiple events, one of them being a shutdown event.

Unfortunately the wait on multiple events, I don't think composes with condition variables, mutexes, etc. i.e. this code needs either an undo or a different rewrite.

Really the world needs interruptable waits on condition variables  and effident mutexes. i.e. missing in Win32, unless you use the expensive kernel events/mutexes/semaphores.

### DIFF
--- a/mono/metadata/debug-helpers.c
+++ b/mono/metadata/debug-helpers.c
@@ -443,6 +443,8 @@ mono_method_desc_from_method (MonoMethod *method)
 void
 mono_method_desc_free (MonoMethodDesc *desc)
 {
+	if (!desc)
+		return;
 	if (desc->name_space)
 		g_free (desc->name_space);
 	else if (desc->klass)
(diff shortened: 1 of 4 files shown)


jaykrell:
@akoeplinger does this fix Alpine?

jaykrell:
@monojenkins build failed

jaykrell:
This wasn't going well.
Let's revert instead -- https://github.com/mono/mono/pull/9007

PR #9006, opened 2018-06-03
Ensure order of evaluation is guaranteed.

jaykrell:


jaykrell:
@monojenkins build failed

PR #6222, opened 2017-12-11
[threadpool] Replace parked_threads condition variable by a semaphore

luhenry:

PR #9007, opened 2018-06-03
do not merge, just trying CI

jaykrell:
…emaphore (#6222)"

This reverts commit 5f5c5e97a08f7086d7c18af37352c4a03dc4c0d1.

While that change simplifies the code, it caused regression.
In particular, it appears there is no longer coordination between shutdown
and the threadpool, leading to hangs at shutdown, of a bounded random
but significant duration. i.e. because threadpool workers do a random wait,
and then check for shutdown in progress. Previously the wait was interrupted.

The hang is often long enough to cause mono/tests/thread6.exe to fail CI.
The hang is usually very obvious running mono/tests/thread6.exe, even if it is within
the threshold allowed by CI.
The hang is long enough to make building mono many times slower on Alpine/muslc.
  I'm guessing due to different random number generation?
  I'm guessing this is the same issue.
  But the problems with mono/tests/thread6.exe are severe enough.

### DIFF
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -131,8 +131,9 @@ typedef struct {
 
 	ThreadPoolWorkerCounter counters;
 
-	MonoCoopSem parked_threads_sem;
+	MonoCoopMutex parked_threads_lock;
 	gint32 parked_threads_count;
+	MonoCoopCond parked_threads_cond;
 
 	volatile gint32 work_items_count;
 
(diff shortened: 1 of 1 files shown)


luhenry:
What is the root cause of this bug? Did you understood exactly how it is supposed to work, and how the assumptions are broken and trigger this regression? You are trying to revert the symptom without figuring out the root of the problem.

jaykrell:
I need to see if this really fixes the problem in CI.
A very similar PR is not.

PR #9014, opened 2018-06-04
[Coop] Convert System.Security.Cryptography.RNGCryptoServiceProvider.

jaykrell:


jaykrell:
Looks like relevant CI failure. Will address later.

jaykrell:
@monojenkins build failes

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/utils/mono-rand.c
```diff
@@ -93,19 +93,19 @@ mono_getrandom (guchar *buffer, gint buffer_size, gint flags, MonoError *error)
...
 static gint
-mono_getentropy (guchar *buffer, gint buffer_size, MonoError *error)
+mono_getentropy (guchar *buffer, intptr_t buffer_size, MonoError *error)
 {
 	static gint getentropy_works = 1;
 	gint count = 0;
 	gint len;
 	gint err;
 
+	error_init (error);
```
luhenry: Remove this `error_init`. `mono_getentropy` is only called by `mono_rand_try_get_bytes` which already calls `error_init`.

mono/utils/mono-rand.c
```diff
@@ -53,18 +53,18 @@ static gint file = -1;
...
    getrandom() is retried if it failed with EINTR: interrupted by a signal. */
 static gint
-mono_getrandom (guchar *buffer, gint buffer_size, gint flags, MonoError *error)
+mono_getrandom (guchar *buffer, intptr_t buffer_size, gint flags, MonoError *error)
 {
 	static gint getrandom_works = 1;
 	gint count = 0;
 	gint err;
 
+	error_init (error);
```
luhenry: Remove this `error_init`. `mono_getrandom` is only called by `mono_rand_try_get_bytes` which already calls `error_init`.

mcs/class/corlib/System.Security.Cryptography/RNGCryptoServiceProvider.cs
```diff
@@ -94,31 +98,33 @@ private void Check ()
 		private static extern bool RngOpen ();
 		
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
-		private static extern IntPtr RngInitialize (byte[] seed);
+		unsafe private static extern IntPtr RngInitialize (byte* seed, IntPtr seed_length);
```
luhenry: Pass the `seed_length` as a `gint32`.

mcs/class/corlib/System.Security.Cryptography/RNGCryptoServiceProvider.cs
```diff
@@ -94,31 +98,33 @@ private void Check ()
...
 
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
-		private static extern IntPtr RngGetBytes (IntPtr handle, byte[] data);
+		unsafe private static extern IntPtr RngGetBytes (IntPtr handle, byte* data, IntPtr data_length);
```
luhenry: Pass the `data_length` as a `gint32`.

jaykrell: `uintptr_t mono_array_handle_length (MonoArrayHandle arr);`
but I used intptr_r as compromise to preserve previous signedness in existing code.

jaykrell:
Later changes maybe crashing, will see.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mcs/class/corlib/System.Security.Cryptography/RNGCryptoServiceProvider.cs
jaykrell: We also have:
```
#ifdef MONO_BIG_ARRAYS
typedef uint64_t mono_array_size_t;
#else
typedef uint32_t mono_array_size_t;
#endif
```
however the ifdef does not translate into the BCL interface -- we only build it once without ifdef. So you have to pick one type that works either way, with either configuration of C and with C#, so the wider type is what works. The smaller type would be silently truncated and silent truncation of integers is dangerous and not worth much risk.

jaykrell:
@monojenkins build pkg with monolite

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

configure.ac
```diff
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=9
+MONO_CORLIB_COUNTER=0
```
jaykrell: Should be 10 but wasn't working, trying 1.0.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
Subtle break on Windows where an invalid parameter is getting a different exception than expected. I don't see where it would have originated in the original. It is obvious where it originates in the revision. Will probably have to build/run on windows and break on prepare_exception.

mcs/class/corlib/System.Security.Cryptography/RNGCryptoServiceProvider.cs
luhenry: https://msdn.microsoft.com/en-us/library/system.array.length%28v=vs.110%29.aspx. `Array.Length` returns `int` which is equivalent to `gint32`.

jaykrell: But what about Mono's large array feature? Worthless? Only for embedding uses that aren't interacting with C#/IL?

jaykrell:
Still an error to figure out in a deliberately degenerate test case.

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build Linux x64

jaykrell:
review? Also checking with @baulig, I think he is kinda but not quite obsoleting this code, and conflicting on approximately one line. Maybe he goes first.

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
CI passing. Approve?

jaykrell:
@monojenkins build failed

PR #9019, opened 2018-06-04
[runtime] Store the interpreter method invoked by the delegate in MonoDelegate.interp_method, make MonoDelegate.method_ptr always point to a JIT callable entry point.

vargaz:


mono/mini/interp/interp.c
```diff
@@ -5156,7 +5172,8 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, guint16 *st
 
 		   --sp;
 		   del = (MonoDelegate*)sp->data.p;
-		   sp->data.p = del->method_ptr;
+		   g_assert (del->interp_method);
+		   sp->data.p = del->interp_method;
```
lewurm: could we make the interp go through `del->method_ptr` as well too? it should be always set, right?

I don't like to introduce another field to `MonoDelegate`.

vargaz: It wouldn't work in mixed mode, the jit assumes method_ptr is a JIT entry point, while the intepreter needs it to be an InterpMethod.

lewurm: couldn't we go through the JIT entry point too with the interpreter?

After `CEE_MONO_LD_DELEGATE_METHOD_PTR` we emit `CEE_MONO_CALLI_EXTRA_ARG`, which is sort of special already. could we convert `CEE_MONO_CALLI_EXTRA_ARG` into a `MINT_JIT_CALL`?

vargaz: We could, but it would be very slow for the normal interpreter, since we would go interp->jit->interp.
MonoDelegate is not a very common object, it doesn't have to be very small.

lewurm: ok

lewurm:
so this fixes the situation if the JIT tries to call a delegate that is suppose to run on the interpreter, right?

what about the other way around, if the interpreter tries to call a jitted delegate. does it go through a `JIT_CALL`?

vargaz:
No, it will stay in the interpreter.

vargaz:
This causes some interpreter failures.

vargaz:

vargaz:
@monojenkins build failed

lewurm:
I think the wasm failure is related:
```
WASM: * Assertion at /mnt/jenkins/workspace/test-mono-pull-request-wasm/mono/mini/interp/interp.c:2386, condition `is_ok (error)' not met, 
function:interp_create_method_pointer, couldn't compile wrapper "(wrapper unknown) byte object:interp_in (object,object)" for "bool System.__Filters:FilterAttribute (System.Reflection.MemberInfo,object)",
due to Attempting to JIT compile method '(wrapper unknown) object:interp_in (object,object)' while
 running in aot-only mode. See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.
```

PR #9021, opened 2018-06-04
Mixed testing

vargaz:


luhenry:
@vargaz is that fixes for mixed-mode? How did you run it locally, and could we run what you are running on CI as well?

vargaz:
Its ongoing work.

PR #9022, opened 2018-06-04
[build] Rename darwin platform to macos + Add `unix` host platform

luhenry:
Darwin is common to both macOS and iOS/tvOS/watchOS, while we use the darwin platform only for macOS.

We also add `unix` platform for any platform other than `linux`, `macos` and `win32`. This is close to `AnyUnix` from corefx.


luhenry:

luhenry:
@akoeplinger I'll update the `linux_*.sources` and `unix_*.sources` to invert their inclusion.

On the other hand, `unix_*.sources` is not a subset of `linux_*.sources` and is not meant to be the common sources between `macos` and `linux` either, so even having one include the other might not be the way forward. But I agree, it's a good solution for now and we should see how far it gets us.

luhenry:

akoeplinger:
@luhenry we should bump the corlib version for this change

akoeplinger:
I do think it'd make sense for `unix_*.sources` to be a common source between macos_ and linux_ since afaik that's how it's also used in corefx?

luhenry:
@monojenkins commit csproj

luhenry:
@akoeplinger for `unix` as a common between `linux` and `macos`: if `unix` is used for platforms other than `linux`, `macos` and `win32`, then there will be some platform-specific files in the `unix_*.sources` files ([`Process.UnknownUnix.cs`](https://github.com/mono/corefx/blob/master/src/System.Diagnostics.Process/src/System/Diagnostics/Process.UnknownUnix.cs) for example). This is not the case yet so we are ok for now, but as soon as it will be the case then it will break this assumption that `unix` is the common parts between `linux` and `macos`.

akoeplinger:

luhenry:
@monojenkins commit csproj

Issue #7167, opened 2018-02-21
Alpine Linux build is broken


akoeplinger:
## Steps to Reproduce

1. Run build on Alpine Linux
2. The build runs very slowly during the `make` step (~4 hrs instead of 35mins) and then runs into all the timeouts of test steps so it's essentially broken

This started with https://jenkins.mono-project.com/job/test-mono-mainline-docker/label=docker-alpine/573/ which makes these commits suspect:

~1.  \[System\] System.Net.* tests from corefx (+RemoteExecutorTestBase) ([commit](https://github.com/mono/mono/commit/e9e17e069e6949600a382341e8888d7486553f21))~
~2.  IOS SDKs (# 6257) ([commit](https://github.com/mono/mono/commit/a2407ac76a50cc1b5fdfbea83adb135d9be40356))~
~3.  Remove a few braces that are contrary to our prescribed style. (# 6245) ([commit](https://github.com/mono/mono/commit/891941ecbe9640b180f73096b8b230c6f48bf11b))~
~4.  \[threads\] Ensure we remove the thread only once (# 6194) ([commit](https://github.com/mono/mono/commit/f6bf1d427b29ba00211f047081e492c56d06b41c))~
5.  \[threadpool\] Replace parked_threads condition variable by a semaphore ([commit](https://github.com/mono/mono/commit/5f5c5e97a08f7086d7c18af37352c4a03dc4c0d1))
~6.  \[runtime\] Fixes issue when compiling with -no\_weak\_imports (# 6276) ([commit](https://github.com/mono/mono/commit/6ca7b346cd3fc52318caccc182f35b8425439a07))~
~7.  Minor cleanup split out and expanded from another larger change. (# 6263) ([commit](https://github.com/mono/mono/commit/fba6bfc9549fb8686576cca20e33df4d01fb2d1a))~
~8.  This list is almost sorted -- so repair the sorting and add comment. ([commit](https://github.com/mono/mono/commit/11e18c2a5d6ed7d17184ad4d5e0606ab1e7ec334))~
~9.  Only call \`waitpid\` for processes that Mono tracks, rather than all ([commit](https://github.com/mono/mono/commit/8c9c37c45bb3302f917fe345004a5076c21d5ba1))~
~10.  cooperative suspend babystep wrt ([commit](https://github.com/mono/mono/commit/e0b1d6651915c395eca637ed172399354ee0fa17))~
[...]

PR #9025, opened 2018-06-05
Make mono_coop_sem_timedwait interruptable.

### LINKED ISSUE #9000: untime-interp/thread6.exe test is flaky
## Steps to Reproduce

1.  Submit a PR.

### Current Behavior

Tests fail.
For example:

Flaky test.

MonoTests.runtime-interp.thread6.exe_timedout

https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-armhf/5285/testReport/junit/MonoTests/runtime-interp/thread6_exe_timedout/

                                                MESSAGE:
                                                Thread 1 started
Count: 0
cought exception level 3 
cought exception level 2 STATETEST
System.Threading.ThreadAbortException: Thread was being aborted.
  at Tests.ThreadStart1 () <0xf5307510 + 0x000aa> in <b5e56f684da84c808720a8b25d32b5e6>:0 
cought exception level 1 
System.Threading.ThreadAbortException: Thread was being aborted.
  at Tests.ThreadStart1 () <0xf5307510 + 0x00118> in <b5e56f684da84c808720a8b25d32b5e6>:0 
end
Working thread abort
Retained thread abort exception
[New LWP 2428]
[New LWP 2434]
[New LWP 2813]
[New LWP 2847]
[New LWP 5130]
[New LWP 8997]
[N

### Expected Behavior

Tests pass.

[x] macOS
[x] Linux
[x] Windows

jaykrell:
This addresses regression from 5f5c5e9 [threadpool] Replace parked_threads condition variable by a semaphore (#6222)

While that change simplifies the code, it caused regression.
In particular, it appears there is no longer coordination between shutdown
and the threadpool, leading to hangs at shutdown, of a bounded random
but significant duration. i.e. because threadpool workers do a random wait,
and then check for shutdown in progress. Previously the wait was interrupted.

The hang is often long enough to cause mono/tests/thread6.exe to fail CI.
The hang is usually very obvious running mono/tests/thread6.exe, even if it is within
the threshold allowed by CI.
The hang is long enough to make building mono many times slower on Alpine/muslc.
  I'm guessing due to different random number generation?
  I'm guessing this is the same issue.
  But the problems with mono/tests/thread6.exe are severe enough.

This should fix #9000 and #7167.
The problem is not specific to Alpine/muslc, it is just more noticable there.

### DIFF
--- a/mono/utils/mono-coop-semaphore.c
+++ b/mono/utils/mono-coop-semaphore.c
@@ -1,0 +1,41 @@
+/**
+ * \file
+ */
+
+#include <config.h>
+#include "mono/utils/mono-coop-semaphore.h"
+#include "mono/utils/mono-threads.h"
+
+static void
+mono_coop_sem_interrupt (gpointer sem)
+{
+	mono_os_sem_post (&((MonoCoopSem*)sem)->s);
+}
+
+MonoSemTimedwaitRet
+mono_coop_sem_timedwait (MonoCoopSem *sem, guint timeout_ms, MonoSemFlags flags)
+{
+	MonoSemTimedwaitRet res;
+	gboolean const alertable = (flags & MONO_SEM_FLAGS_ALERTABLE) != 0;
+	gboolean interrupted = FALSE;
+
+	if (alertable) {
+		mono_thread_info_install_interrupt (mono_coop_sem_interrupt, sem, &interrupted);
+		if (interrupted)
+			return MONO_SEM_TIMEDWAIT_RET_ALERTED;
+	}
+
+	MONO_ENTER_GC_SAFE;
+
+	res = mono_os_sem_timedwait (&sem->s, timeout_ms, flags);
+
+	MONO_EXIT_GC_SAFE;
+
+	if (alertable)
+		mono_thread_info_uninstall_interrupt (&interrupted);
+
+	if (interrupted)
+		return MONO_SEM_TIMEDWAIT_RET_ALERTED;
+
+	return res;
+}
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -646,6 +646,7 @@ TESTS_CS_SRC=		\
 	generic-unloading.2.cs	\
 	namedmutex-destroy-race.cs	\
 	thread6.cs	\
+	thread7.cs	\
 	appdomain-threadpool-unload.cs	\
 	process-unref-race.cs	\
 	bug-46661.cs	\
--- a/mono/tests/thread7.cs
+++ b/mono/tests/thread7.cs
@@ -1,0 +1,58 @@
+//
+// Subset of thread6.cs, just to watch for the hang at end.
+// Note this test does not indicate success or failure well,
+// it just runs quickly or slowly.
+//
+using System;
+using System.Threading;
+
+public class Tests {
+
+	public static int Main() {
+		return test_0_regress_4413();
+	}
+
+	public static int test_0_regress_4413 () {
+		// Check that thread abort exceptions originating in another thread are not automatically rethrown
+		object o = new object ();
+		Thread t = null;
+		bool waiting = false;
+		Action a = delegate () {
+			t = Thread.CurrentThread;
+			while (true) {
+				lock (o) {
+					if (waiting) {
+						Monitor.Pulse (o);
+						break;
+					}
+				}
+
+				Thread.Sleep (10);
+			}
+			while (true) {
+				Thread.Sleep (1000);
+			}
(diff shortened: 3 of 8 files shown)


jaykrell:
Need to update Windows project files..

jaykrell:
Hm, thread7.exe still timed out. Maybe there is an interpreter problem?

jaykrell:
Maybe look closer here.
Make sure it implemented alertable.
Maybe an interpreter problem.

https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/6255/parsed_console/log.html

```
Thread 1 (Thread 0x7f31d2e6b740 (LWP 63969)):
#0  0x00007f31d233615f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x000055fff8a37e65 in mono_os_cond_wait (mutex=0x55fff90776b8, cond=0x55fff90776e0) at mono-os-mutex.h:173
#2  mono_os_cond_timedwait (cond=cond@entry=0x55fff90776e0, mutex=mutex@entry=0x55fff90776b8, timeout_ms=timeout_ms@entry=4294967295) at mono-os-mutex.c:32
#3  0x000055fff89967d5 in mono_coop_cond_timedwait (timeout_ms=4294967295, mutex=0x55fff90776b8, cond=0x55fff90776e0) at ../../mono/utils/mono-coop-mutex.h:102
#4  mono_w32handle_timedwait_signal_naked (alerted=0x7ffc7c588d30, poll=0, timeout=4294967295, mutex=0x55fff90776b8, cond=0x55fff90776e0) at w32handle.c:646
#5  mono_w32handle_timedwait_signal_handle (handle_data=0x55fff90776a8, timeout=timeout@entry=4294967295, alerted=alerted@entry=0x7ffc7c588d30, poll=0) at w32handle.c:761
#6  0x000055fff8997a3f in mono_w32handle_wait_one (handle=handle@entry=0x55fff90776a8, timeout=timeout@entry=4294967295, alertable=alertable@entry=1) at w32handle.c:869
#7  0x000055fff89815a7 in mono_threadpool_end_invoke (ares=ares@entry=0x7f31d192c720, 
```

jaykrell:
That code is pretty convincing.
`mono_w32handle_timedwait_signal_handle` handles alertable.

jaykrell:
It seems there was a failure still, however there was a maybe critical typo and the repro/norepro are pretty cut and dry, interp or not.

jaykrell:
Locally this consistently helps. In CI it still seems flaky. :(

jaykrell:

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
You can see in the failures the worker threads are still running, so this is the problem..still pondering it.

mono/utils/mono-coop-semaphore.c
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#include <config.h>
+#include "mono/utils/mono-coop-semaphore.h"
+#include "mono/utils/mono-threads.h"
+
+static void
+mono_coop_sem_interrupt (gpointer sem)
+{
+	mono_os_sem_post (&((MonoCoopSem*)sem)->s);
```
BrzVlad: This doesn't guarantee that it will wake up our waiter. It can wake up any other thread waiting on this semaphore and you can't really broadcast on a semaphore, it would also screw up the semaphore count. For these reasons, for example, I used a condvar in the monitor code instead of a semaphore.

jaykrell: Interesting.
What if we undo 5f5c5e97a08f7086d7c18af37352c4a03dc4c0d1 but then
move the interruptibility into the coop_cond layer?

I was also debugging to see if there is was some other regression.
Like, @ludovic's hints were around the abort vs. suspend signal
and SA_RESTART and we should get back EINTR.

I see we do sometimes get EINTR, on Linux, but it doesn't work out for
reasons I don't yet know.

Mac is different. Signals are not used.

luhenry:
As discussed today, this is not the way to fix the bug you are looking at. This is also very racy and is breaking the abstraction these functions are putting in place.

jaykrell:
Linux bisect seems to be getting to the same change in December. It didn’t, I thought earlier, probably because of too few tries and random numbers. Still confirming..

jaykrell:
There are a fair number of uses of registering interrupts, how this code use. All suspicious or just some?

PR #9027, opened 2018-06-05
[Coop] Cleanup ahead of converting ves_icall_System_Threading_Thread_…

jaykrell:
…GetAbortExceptionState.

 - serialize_object and deserialize_object are almost identical and
   can be implemented on top of the same function.
 - Error propagation was a little wierd and had FIXME comments.
   It involved MonoError, MonoException, and gboolean *failed, and
   some conversions. Converge on mostly MonoError, with fewer conversions,
   albeit still sometimes holding a MonoException.

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -4396,8 +4396,7 @@ mono_runtime_try_run_main (MonoMethod *method, int argc, char* argv[],
 }
 
 static MonoObject*
-serialize_or_deserialize_object (MonoObject *obj, MonoError *error,
-	const gchar *method_name, MonoMethod **method)
+serialize_or_deserialize_object (MonoObject *obj, const gchar *method_name, MonoMethod **method, MonoError *error)
 {
 	if (!*method) {
 		MonoClass *klass = mono_class_get_remoting_services_class ();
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

mono/metadata/object.c
```diff
@@ -4395,76 +4395,43 @@ mono_runtime_try_run_main (MonoMethod *method, int argc, char* argv[],
 	return mono_runtime_try_exec_main (method, args, exc);
 }
 
-
 static MonoObject*
-serialize_object (MonoObject *obj, gboolean *failure, MonoObject **exc)
+serialize_or_deserialize_object (MonoObject *obj, MonoError *error,
+	const gchar *method_name, MonoMethod **method)
 {
```
vargaz: error is usually passed as the last argument.

mono/metadata/object.c
```diff
@@ -4395,76 +4395,43 @@ mono_runtime_try_run_main (MonoMethod *method, int argc, char* argv[],
...
 		MonoClass *klass = mono_class_get_remoting_services_class ();
-		serialize_method = mono_class_get_method_from_name (klass, "SerializeCallData", -1);
+		*method = mono_class_get_method_from_name (klass, method_name, -1);
 	}
 
-	if (!serialize_method) {
-		*failure = TRUE;
+	if (!*method) {
+		mono_error_set_exception_instance (error, NULL);
 		return NULL;
```
vargaz: What will this do with a NULL argument ? I think this could be just an assert.

mono/metadata/object.c
```diff
@@ -4395,76 +4395,43 @@ mono_runtime_try_run_main (MonoMethod *method, int argc, char* argv[],
...
 	}
 
-	g_assert (!mono_class_is_marshalbyref (mono_object_class (obj)));
-
```
vargaz: This could stay.

mono/metadata/object.c
jaykrell: I was kinda appending parameters, albeit having replaced failure/exc with error; ok.

mono/metadata/object.c
jaykrell: I'm emulating the previous behavior I believe -- where it'd set *failure = TRUE and leave exc == NULL and caller would mono_error_set_exception_instance.

mono/metadata/object.c
```diff
@@ -4395,76 +4395,43 @@ mono_runtime_try_run_main (MonoMethod *method, int argc, char* argv[],
...
-	array = mono_runtime_try_invoke (serialize_method, NULL, params, exc, error);
-	if (*exc == NULL && !mono_error_ok (error))
-		*exc = (MonoObject*) mono_error_convert_to_exception (error); /* FIXME convert serialize_object to MonoError */
-	else
-		mono_error_cleanup (error);
+	MonoException *exc = NULL;
+	void *params [ ] = { obj };
+	MonoObject *result = mono_runtime_try_invoke (*method, NULL, params, (MonoObject**)&exc, error);
+	if (!is_ok (error) && exc)
```
jaykrell: This is off -- remove the exclamation point.

mono/metadata/object.c
jaykrell: I think "it works" -- the error handling infrastructure will take it -- but maybe indeed assert is preferred.

jaykrell:
@monojenkins build  Linux ARMv7 Interpreter

jaykrell:

PR #9029, opened 2018-06-05
[interp] Transition mono_interp_transform_method () to use MonoError.

vargaz:


mono/mini/interp/transform.c
```diff
@@ -4944,6 +4935,6 @@ mono_interp_transform_method (InterpMethod *imethod, ThreadContext *context, Int
 	}
 	mono_os_mutex_unlock (&calc_section);
 
-	return NULL;
+	return;
```
jaykrell: Coud remove line.

lewurm:
LGTM. why though?

Issue #9030, opened 2018-06-05
[Mac] Use copyfile API to copy files


migueldeicaza:
Since MacOS 10.5 there is an API to copy files and directories recursively.   These APIs are capable of copying the metadata along with the file contents.

(man copyfile, also new APIs here: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/APFS_Guide/ToolsandAPIs/ToolsandAPIs.html#//apple_ref/doc/uid/TP40016999-CH7-DontLinkElementID_11)

Additionally, these APIs are accelerated when running on top of Apple File System.

Therzok:
Might be worth taking a look at this in MSBuild? - to prevent lots of time spent on copying files to output directory (cc @radical), which is most of the time in NOP builds (40s on MonoDevelop's sln)

migueldeicaza:
I also have a managed implementation that I can share.

radical:
msbuild will automatically get the benefit via `File.Copy`, which #9038 is updating to use this API.

migueldeicaza:
Here is my C# implementation, you can use it if you want as either a NuGet package, or just copy the C# file into the place you need it, it is self-contained and tiny:

https://github.com/migueldeicaza/Darwin.CopyFile

PR #9038, opened 2018-06-05
[MacOS] On Mac, use the copyfile API to copy files.

migueldeicaza:
Copyfile is a high level API that does all the heavy lifting for
copying files and does this in a Mac-correct way.   We configure it
to copy both the contents as well as the metadata and extended attributes
in the file.

Additionally, this uses COPYFILE_CLONE which will attempt to use the
APFS clone capabilities for fast copies.

On my Mac with APFS, copying a 5 gig file (Xcode) goes from taking
7.2 seconds to copy on an SSD to 0.5 seconds.

Fixes: https://github.com/mono/mono/issues/9030


mono/metadata/w32file-unix.c
```diff
@@ -14,6 +14,12 @@
 #ifdef HAVE_SYS_STATVFS_H
 #include <sys/statvfs.h>
 #endif
+#ifdef HAVE_COPYFILE_H
+#include <copyfile.h>
+#  if !defined(COPYFILE_CLONE)
```
luhenry: `#ifndef` and no indentation, for consistency with rest of file.

mono/metadata/w32file-unix.c
```diff
@@ -2359,7 +2365,17 @@ CopyFile (const gunichar2 *name, const gunichar2 *dest_name, gboolean fail_if_ex
 		
 		return(FALSE);
 	}
-	
+
+#if HAVE_COPYFILE_H
```
luhenry: `#ifdef` to be consistent with https://github.com/mono/mono/pull/9038/files#diff-e3902264b6c6cb537e9eec9149668e83R17

mono/metadata/w32file-unix.c
```diff
@@ -14,6 +14,12 @@
 #ifdef HAVE_SYS_STATVFS_H
 #include <sys/statvfs.h>
 #endif
+#ifdef HAVE_COPYFILE_H
+#include <copyfile.h>
+#  if !defined(COPYFILE_CLONE)
+#    #define COPYFILE_CLONE 0
```
lewurm: why is this needed?

mono/metadata/w32file-unix.c
migueldeicaza: No - that is a bad style that someone went and added in recent years.   Undo that disaster please.

mono/metadata/w32file-unix.c
migueldeicaza: Older MacOS versions do not have COPYFILE_CLONE

mono/metadata/w32file-unix.c
migueldeicaza: The top of the file uses #ifdef, the rest of the file uses #if.

Wont make this change.

mono/metadata/w32file-unix.c
jaykrell: Can you unconditionally define it to the correct value?
Or you want 0 (incorrect value?) as no-op on old systems?

jaykrell: Correct value is #define COPYFILE_CLONE          (1<<24)
but intent might be to make it a no-op downlevel.

luhenry:
@migueldeicaza this seems to break quite a few tests on macOS:
 - https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/11085/testReport/
 - https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/11092/testReport/

mono/metadata/w32file-unix.c
migueldeicaza: Good idea

migueldeicaza:
Submitting a new patch, to prevent the breakage in IsolatedStorage, the subtle semantics of Copy and should fix the MSBuild issue.

mono/metadata/w32file-unix.c
```diff
@@ -2359,7 +2365,17 @@ CopyFile (const gunichar2 *name, const gunichar2 *dest_name, gboolean fail_if_ex
 		
 		return(FALSE);
 	}
-	
+
+#if HAVE_COPYFILE_H
+	ret = copyfile (utf8_src, utf8_dest, NULL, COPYFILE_ALL | COPYFILE_CLONE | (fail_if_exists ? COPYFILE_EXCL : 0));
```
filipnavara: This won't work because COPYFILE_CLONE implies COPYFILE_EXCL (https://opensource.apple.com/source/copyfile/copyfile-146.30.2/copyfile.c.auto.html). One way to solve that is to fallback on `if (ret != 0 && errno == EEXIST && ! fail_if_exists)` to non-cloning variant. I wanted to check why COPYFILE_UNLINK didn't work, but the build results are gone now :-/

filipnavara: Nvm, I see it now - same file name for src and dest is broken for COPYFILE_UNLINK.

PR #9039, opened 2018-06-05
[runtime] Test GetPid for Mono.Posix

alexanderkyte:


mcs/class/Mono.Posix/Test/Mono.Unix.Native/StdlibTest.cs
```diff
@@ -29,6 +29,15 @@ public void Handler (int sn)
 			}
 		}
 
+
+		[Test]
+		public void GetPid ()
+		{
+			var currentPID = Mono.Unix.Native.Syscall.getpid();
+			Console.WriteLine ("CurrentPID is {0}", currentPID);
```
akoeplinger: please remove this unnecessary logging, we'll get the value in the assertion message anyway.

mcs/class/Mono.Posix/Test/Mono.Unix.Native/StdlibTest.cs
```diff
@@ -29,6 +29,15 @@ public void Handler (int sn)
...
+			Assert.AreNotEqual (currentPID, 0);
```
akoeplinger: the arguments need to be switched, it's `(expected, actual)` not the other way round :)

might make sense to use `Assert.Greater(currentPID, 0)` so we verify we don't get a negative value too.

jaykrell: Who says negative values are invalid?
http://pubs.opengroup.org/onlinepubs/007904875/functions/getpid.html ? no.
mac man posix? no.

jaykrell: 0 is valid too. There is no reserved value.

akoeplinger: Right but the fact that every OS we care about returns positive PIDs and uses PID0 for scheduling/system processes makes me confident in this being fine for testing code :)

luhenry:
@alexanderkyte what's the progress on this PR?

luhenry:
@monojenkins build

PR #9042, opened 2018-06-06
Fix/mitigate 5-60 second hang at shutdown (mono/tests/thread6.exe).

### LINKED ISSUE #9000: untime-interp/thread6.exe test is flaky

jaykrell:
Don't wait 5-60 seconds for work in threadpool on aborted threads.
This fixes #9000 and likely #7167.

Aborting already aborted threads is skipped, but interrupts on aborted threads are not.
This regressed with change 5f5c5e9 [threadpool] Replace parked_threads condition variable by a semaphore (#6222)

The skipping of syscall abort in aborted threads is circa here:
```
request_thread_abort:
	if (thread->state & (ThreadState_AbortRequested | ThreadState_Stopped))
	{
		UNLOCK_THREAD (thread);
		return FALSE;
	}
```

Otherwise there is a wait for a random duration between 5 and 60 seconds in mono/tests/thread6.cs,
which aborts a threadpool worker thread -- probably an inappropriate thing to do,
threadpool threads being a shared resource, however with the right amount of synchronization,
one can probably be sure to only abort one's own work, and the threadpool will survive by
spinning up additional threads, or in this case, shutting down in the expected rapid way.

Possibly there should be a notion of canceling a threadpool work item or task, instead
of the entire thread. Even if it is safe, it is defeating the efficiency of a threadpool.

There might be another fix involving doing more of the abort processing of aborted threads.

### DIFF
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -525,7 +525,7 @@ worker_try_create (void)
 	ERROR_DECL (error);
 	MonoInternalThread *thread;
 	gint64 current_ticks;
-	gint32 now;
+	gint32 now = 0;
 	ThreadPoolWorkerCounter counter;
 
 	if (mono_runtime_is_shutting_down ())
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -367,7 +367,9 @@ worker_park (void)
 	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_THREADPOOL, "[%p] worker parking",
 		GUINT_TO_POINTER (MONO_NATIVE_THREAD_ID_TO_UINT (mono_native_thread_id_get ())));
 
-	if (!mono_runtime_is_shutting_down () && !mono_thread_test_state (mono_thread_internal_current (), ThreadState_AbortRequested)) {
+	MonoInternalThread * const thread = mono_thread_internal_current ();
+
+	if (!mono_runtime_is_shutting_down ()) {
 		static gpointer rand_handle = NULL;
 		ThreadPoolWorkerCounter counter;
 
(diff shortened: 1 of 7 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
4 instead of expected 3 failures on FullAOT+LLVM, let's see if flake or not.

jaykrell:

jaykrell:
And then might have regressed  MonoTests.runtime.appdomain-unload.exe_timedout -- two failures of that. A compromise might be to shorten the wait if the thread is aborted.

jaykrell:
Or to interrupt the syscall as suggested. That'd handle other kinds of waits w/o having to check for abort everywhere and reason about it.

jaykrell:
Or check MONO_THREAD_FLAG_APPDOMAIN_ABORT.

jaykrell:
Notice that threadpool threads are treated as a special case here. They are "background" threads. So analogs to arbitrary syscalls are not clear.

jaykrell:

jaykrell:
`mono_thread_interruption_checkpoint_bool`?

jaykrell:

jaykrell:
Locally, without a fix, JIT and interp hang the same.
Locally, with a fix, JIT and interp are both fixed.
In CI the fix still leaves interp failing.

jaykrell:
@monojenkins build failed

lewurm:
superseded by https://github.com/mono/mono/pull/9163

PR #9054, opened 2018-06-07
Unblock signal when adding mono signal handler

mfilippov:
On Unix, if an application block some signals all child processes run with exec would inherit its block. To avoid this problem I add signal into signal set to unblock when adding a handler for a signal.

### DIFF
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -401,35 +401,46 @@ mono_runtime_posix_install_handlers (void)
 {
 
 	sigset_t signal_set;
-
-	if (mini_get_debug_options ()->handle_sigint)
+	sigemptyset (&signal_set);
+	if (mini_get_debug_options ()->handle_sigint) {
 		add_signal_handler (SIGINT, mono_sigint_signal_handler, SA_RESTART);
+		sigaddset (&signal_set, SIGINT);
+	}
 
 	add_signal_handler (SIGFPE, mono_sigfpe_signal_handler, 0);
+	sigaddset (&signal_set, SIGFPE);
 	add_signal_handler (SIGQUIT, sigquit_signal_handler, SA_RESTART);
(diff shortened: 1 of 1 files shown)


mono/mini/mini-posix.c
```diff
@@ -401,27 +401,37 @@ mono_runtime_posix_install_handlers (void)
...
 
 	/* it seems to have become a common bug for some programs that run as parents
 	 * of many processes to block signal delivery for real time signals.
 	 * We try to detect and work around their breakage here.
 	 */
-	sigemptyset (&signal_set);
 	if (mono_gc_get_suspend_signal () != -1)
 		sigaddset (&signal_set, mono_gc_get_suspend_signal ());
 	if (mono_gc_get_restart_signal () != -1)
 		sigaddset (&signal_set, mono_gc_get_restart_signal ());
```
lewurm: not really related to your changes, but we already do that here:
https://github.com/mono/mono/blob/006d6ce09734e8b555a53324c21bbaab3f1ebcb4/mono/utils/mono-threads-posix-signals.c#L250-L263
I guess it doesn't hurt though.

mono/mini/mini-posix.c
```diff
@@ -401,27 +401,37 @@ mono_runtime_posix_install_handlers (void)
...
 	sigaddset (&signal_set, SIGCHLD);
+	sigaddset (&signal_set, SIGPIPE);
+	sigaddset (&signal_set, SIGABRT);
+	sigaddset (&signal_set, SIGSEGV);
```
lewurm: do you mind moving that further down for the sake of keeping the same pattern?

mfilippov: Yes, I try don't touch order of sigprocmask and add_handler.

mfilippov:
@lewurm done

alexanderkyte:
Looks great to me, thank you @mfilippov

PR #9061, opened 2018-06-07
library.make: track changes in *.sources

EgorBo:
Currently if we change for example `corlib.dll.sources` and then execute `make` - the `make` won't rebuild the lib, because it depends only on `net_4_x_corlib.dll.sources` and doesn't depend on all `#include` inside it.
As a temporarily solution we can depend on all *.sources files in the directory so a change in any of them will trigger rebuild. Unfortunately changes in un-related profiles will also trigger the rebuild but it's better than nothing for now I guess. For better experience we should fetch all #includes recursively for the current profile

### DIFF
--- a/mcs/build/library.make
+++ b/mcs/build/library.make
@@ -295,8 +295,8 @@ endif
 # The library
 
 # If the directory contains the per profile include file, generate list file.
-PROFILE_sources := $(firstword $(if $(PROFILE_PLATFORM),$(wildcard $(PROFILE_PLATFORM)_$(PROFILE)_$(LIBRARY).sources)) $(wildcard $(PROFILE)_$(LIBRARY).sources) $(wildcard $(LIBRARY).sources))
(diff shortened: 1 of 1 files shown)


mcs/build/library.make
```diff
@@ -295,8 +295,8 @@ endif
 # The library
 
 # If the directory contains the per profile include file, generate list file.
-PROFILE_sources := $(firstword $(if $(PROFILE_PLATFORM),$(wildcard $(PROFILE_PLATFORM)_$(PROFILE)_$(LIBRARY).sources)) $(wildcard $(PROFILE)_$(LIBRARY).sources) $(wildcard $(LIBRARY).sources))
-PROFILE_excludes = $(firstword $(if $(PROFILE_PLATFORM),$(wildcard $(PROFILE_PLATFORM)_$(PROFILE)_$(LIBRARY).exclude.sources)) $(wildcard $(PROFILE)_$(LIBRARY).exclude.sources))
+# TODO: depend on all *.sources for now and figure out how to list only needed files later
+PROFILE_sources = $(shell ls *.sources)
```
akoeplinger: instead of `shell ls` use the `wildcard` makefile function.

Issue #8707, opened 2018-05-14
Invalidate bockbuild package cache after a few days


akoeplinger:
We had a few cases (like new Span APIs on core types) where a build break was disguised by the package cache in bockbuild and we only found out some time later.

@alexischr and me discussed this on and off and the best short term fix we can think of is to introduce a way to invalidate these caches after X days so we notice these breaks earlier.

PR #9068, opened 2018-06-07
Update bockbuild for artifact expiration

alexischr:
Fixes https://github.com/mono/mono/issues/8707


alexischr:
@monojenkins build pkg

PR #9081, opened 2018-06-08
cleanup: remove redundant condition

chipitsine:
issue detected by cppcheck

[mono/metadata/threads.c:3993] -> [mono/metadata/threads.c:3991]:
(warning) Either the condition 'if(location&&location->source_file)'
is redundant or there is possible null pointer dereference: location.

condition is redundant indeed, it is nested in "if (location) {"


luhenry:
@monojenkins build

PR #9082, opened 2018-06-08
Reading of Metadata and pdb blobs via Soft debugger protocol

nerzhulart:
This PR allows debugger clients to retrieve metadata and pdbs as blobs via soft debugger protocol. Debugger may use own metadata/symbol reader which makes possible to extract more information than *Mirror objects provide.
Added commands to get TypeMirror and MethodMirror from assembly by metadata token

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs
@@ -3,6 +3,7 @@ using System.Reflection;
 using Mono.Debugger;
 using Mono.Cecil;
 using System.Collections.Generic;
+using System.IO;
 
 namespace Mono.Debugger.Soft
 {
@@ -15,8 +16,13 @@ namespace Mono.Debugger.Soft
 		AssemblyName aname;
 		AssemblyDefinition meta;
 		AppDomainMirror domain;
+		byte[] metadata_blob;
+		bool? isDynamic;
+		byte[] pdb_blob;
 		Dictionary<string, long> typeCacheIgnoreCase = new Dictionary<string, long> (StringComparer.InvariantCultureIgnoreCase);
 		Dictionary<string, long> typeCache = new Dictionary<string, long> ();
+		Dictionary<uint, long> tokenTypeCache = new Dictionary<uint, long> ();
+		Dictionary<uint, long> tokenMethodCache = new Dictionary<uint, long> ();
 
 		internal AssemblyMirror (VirtualMachine vm, long id) : base (vm, id) {
 		}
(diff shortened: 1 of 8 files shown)


vargaz:
Will review.
In general, it is better to obtain these on the client side, since these files are huge, and transferring them over (possibly slow) links is not good for performance.

shalupov:
@vargaz Generally yes, but if don't have them at the client side it turns into usability nightmare

nerzhulart:
@vargaz Generally it's impossible to get metadata, e.g. when you attached to a remote machine or device. So we want to have unified way to read metadata/symbols, because Mirror API is very poor. We've checked the performance - e.g. transferring of mscorlib (which are quite huge - about 4 mb) takes about 10ms on our machines when debugging locally, so I guess it's won't be a bottleneck. Also debugger may read files when they're accessible locally and ask protocol when not

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs
```diff
@@ -128,5 +141,68 @@ public TypeMirror GetType (String name, Boolean throwOnError)
...
+
+				isDynamic = vm.conn.Assembly_IsDynamic(id);
+				return isDynamic.Value;
+			}
+		}	
+		
+		public bool HasPdb {
+			get {			
+				return GetPdbBlob().Length > 0;
+			}
```
vargaz: A method named HasSomething probably shouldn't load mb-s of data from the network.

mono/mini/debugger-agent.c
```diff
@@ -394,7 +395,12 @@ typedef enum {
 	CMD_ASSEMBLY_GET_OBJECT = 4,
 	CMD_ASSEMBLY_GET_TYPE = 5,
 	CMD_ASSEMBLY_GET_NAME = 6,
-	CMD_ASSEMBLY_GET_DOMAIN = 7
+	CMD_ASSEMBLY_GET_DOMAIN = 7,
+    CMD_ASSEMBLY_GET_METADATA_BLOB = 8,
+    CMD_ASSEMBLY_GET_IS_DYNAMIC = 9,
```
vargaz: space vs tabs problem.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/AssemblyMirror.cs
```diff
@@ -15,8 +16,13 @@ public class AssemblyMirror : Mirror
 		AssemblyName aname;
 		AssemblyDefinition meta;
 		AppDomainMirror domain;
+		byte[] metadata_blob;
+		bool? isDynamic;
+		byte[] pdb_blob;
 		Dictionary<string, long> typeCacheIgnoreCase = new Dictionary<string, long> (StringComparer.InvariantCultureIgnoreCase);
 		Dictionary<string, long> typeCache = new Dictionary<string, long> ();
+		Dictionary<uint, long> tokenTypeCache = new Dictionary<uint, long>();
+		Dictionary<uint, long> tokenMethodCache = new Dictionary<uint, long>();
```
vargaz: Please use the mono coding conventions i.e. space before '(' both in the c# and c code.

vargaz:
Looks ok otherwise.

Iliya-usov:
@alexanderkyte @vargaz Fixed

marek-safar:
approve

marek-safar:
@monojenkins commit apidiff

marek-safar:
Thank you very much for the contribution!

vargaz:
This caused a regression since parts of Mono.Debugger.Soft check Metadata against null. This PR will fix it:
https://github.com/mono/mono/pull/9346

PR #9087, opened 2018-06-09
Fix up behaviour of verbose method with arguments

Therzok:
Given a method spec which looks like:
`MainClass:Do(int,int,int)`

Via f024e820a025aacc1cf80eb82a8aee4552c3ac11 the code would match 3 methods:
`MainClass:Do(int`, `int`, `int)`, crashing the runtime.

Accept the multiple method matches via `;`, rather than `,`.

Now, the behaviour is changed so:
`MainClass:Do(int,int,int);MainClass:Do(int,int)`

Get parsed as:
`MainClass:Do(int,int,int)`
`MainClass:Do(int,int)`

## Notes:

Fixing the behaviour so it would actually support commas, it would mean not to use g_splitenv, which would complicate the PR.

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -2029,12 +2029,12 @@ testbundle: console.exe
 
 test-verbose: verbose.exe
 	@echo "Testing MONO_VERBOSE_METHOD"
-	$(Q) MONO_VERBOSE_METHOD='Method' $(RUNTIME) verbose.exe
-	$(Q) MONO_VERBOSE_METHOD='Method(int)' $(RUNTIME) verbose.exe
-	$(Q) MONO_VERBOSE_METHOD='N1.Test:Method' $(RUNTIME) verbose.exe
-	$(Q) MONO_VERBOSE_METHOD='Test:Method' $(RUNTIME) verbose.exe
-	$(Q) MONO_VERBOSE_METHOD='Method(int,int)' $(RUNTIME) verbose.exe
-	$(Q) MONO_VERBOSE_METHOD='Method(int,string[])' $(RUNTIME) verbose.exe
+	$(Q) MONO_VERBOSE_METHOD='*:Method' $(RUNTIME) --compile-all verbose.exe
+	$(Q) MONO_VERBOSE_METHOD='*:Method(int)' $(RUNTIME) --compile-all verbose.exe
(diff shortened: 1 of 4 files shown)


Therzok:
Old behaviour was just crashing the runtime. How do I go about unit testing this?

marek-safar:
You add e.g. `testverbose` target to `mono/tests/Makefile.am`

Therzok:
Tests added.

Therzok:
They do not test the output, just all the various combinations of inputs for no crash.

mono/tests/Makefile.am
```diff
@@ -666,7 +667,8 @@ TESTS_CS_SRC=		\
 	tailcall-generic-cast-cs.cs \
 	tailcall-interface.cs \
 	bug-60843.cs	\
-	nested_type_visibility.cs
+	nested_type_visibility.cs \
+	verbose.cs
```
marek-safar: you are missing the file

Therzok: Added

mono/tests/Makefile.am
```diff
@@ -1627,6 +1629,15 @@ testbundle: console.exe
 	@$(with_mono_path) MONO_CFG_DIR=$(mono_cfg_dir) ./a.out >> mkbundle.stdout
 	@- rm -rf a.out
 
+test-verbose: verbose.exe
+	@echo "Testing MONO_VERBOSE_METHOD"
+	$(Q) MONO_VERBOSE_METHOD='*:Method' $(RUNTIME) --compile-all verbose.exe
```
luhenry: How do you verify that the output is correct? I would suggest having a `verbose.exe` test which will launch a subprocess with the appropriate `MONO_VERBOSE_METHOD` env variable defined and which verifies the output of the subprocess. You can then run this `verbose.exe` as part of the `test-jit` command, allowing it to integrate with NUnit.

Therzok: Sure, I'll take a look over the weekend.

Therzok:
Hope I got it right now.

Therzok:
Hmmm, running the test suite locally (`make test-jit`) passes

Therzok:
@monojenkins build failed

Therzok:
@vargaz @luhenry PTAL

Therzok:
AArch64 seems unrelated, not sure about Windows failures.

Therzok:
@marek-safar ping

PR #9090, opened 2018-06-09
Modernize Windows random number generation.

jaykrell:
- Less code.
 - Less chance for failure.
 - Not seedable.
 - Requires Windows 7 or newer. Er. Keep Vista working for Unity.


jaykrell:
@monojenkins build WIndows i386

jaykrell:
@monojenkins build WIndows x64

jaykrell:
@monojenkins build failed

marek-safar:
@lateralusX could you please review this

jaykrell:
Cygwin makefile needs update here.

lateralusX:
@jaykrell Could you fix the Cygwin build so we could get all tests running? Could also be good to check that we have test coverage on this area.

jaykrell:
I will update cygwin.
Test coverage can be seen here: https://github.com/mono/mono/pull/9014

jaykrell:
I mean -- that that PR has test failures, shows that we have tests here. (still to figure out that one..)

mono/utils/mono-rand-windows.c
```diff
@@ -7,63 +7,33 @@
...
-	if (!CryptAcquireContext (&provider, NULL, NULL, PROV_INTEL_SEC, CRYPT_VERIFY_CONTEXT)) {
-		/* not a PIII or no drivers available, use default RSA CSP */
-		if (!CryptAcquireContext (&provider, NULL, NULL, PROV_RSA_FULL, CRYPT_VERIFY_CONTEXT)) {
-			/* exception will be thrown in managed code */
-			provider = 0;
-		}
+	while (buffer_size > 0) {
+		ULONG const size = (ULONG)MIN (buffer_size, ULONG_MAX);
+		long const status = BCryptGenRandom (NULL, buffer, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+		if (status < 0)
```
lateralusX: Check status using BCRYPT_SUCCESS instead?

mono/utils/mono-rand-windows.c
```diff
@@ -163,6 +92,5 @@ mono_rand_try_get_bytes (gpointer *handle, guchar *buffer, gint buffer_size, Mon
 void
 mono_rand_close (gpointer handle)
 {
-	mono_rand_win_close_provider ((MONO_WIN32_CRYPT_PROVIDER_HANDLE) handle);
```
lateralusX: Maybe assert that the handle value is the same as returned from mono_rand_init?

mono/utils/mono-rand-windows.c
```diff
@@ -122,37 +72,16 @@ mono_rand_init (guchar *seed, gint seed_size)
 gboolean
 mono_rand_try_get_bytes (gpointer *handle, guchar *buffer, gint buffer_size, MonoError *error)
 {
-	MONO_WIN32_CRYPT_PROVIDER_HANDLE provider;
+	g_assert (buffer || !buffer_size);
```
lateralusX: Maybe assert that the handle value is the same as returned from mono_rand_init?

mono/utils/mono-rand-windows.c
```diff
@@ -7,63 +7,33 @@
...
-	/* We first try to use the Intel PIII RNG if drivers are present */
-	if (!CryptAcquireContext (&provider, NULL, NULL, PROV_INTEL_SEC, CRYPT_VERIFY_CONTEXT)) {
-		/* not a PIII or no drivers available, use default RSA CSP */
-		if (!CryptAcquireContext (&provider, NULL, NULL, PROV_RSA_FULL, CRYPT_VERIFY_CONTEXT)) {
-			/* exception will be thrown in managed code */
-			provider = 0;
-		}
+	while (buffer_size > 0) {
+		ULONG const size = (ULONG)MIN (buffer_size, ULONG_MAX);
+		long const status = BCryptGenRandom (NULL, buffer, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
```
lateralusX: Will this take us up to Windows 7 when using BCryptGenRandom and BCRYPT_USE_SYSTEM_PREFERRED_RNG? Looking at the MSDN documentation it says:

> Windows Vista:  This flag is not supported.

I believe Mono should still supporting Vista and WINVER and _WIN32_WINNT is set to 0x0600 so we need to double check this since Unity might still have dependencies on Vista and forward.

lateralusX: Missed the last line in the PR description, so taking this to Win7 might not be possible unless we check with Unity. Another alternative would be to take the implementation currently under uwp as a start, should work under Vista.

mono/utils/mono-rand-windows.c
jaykrell: No need.

mono/utils/mono-rand-windows.c
jaykrell: No need.

jaykrell:
https://unity3d.com/unity/system-requirements suggests this is ok.

mono/utils/mono-rand-windows.c
lateralusX: But clearer and more future proof since the purpose of that macro in the bcrypt.h header file is to check for success/failures of bcrypt API calls.

mono/utils/mono-rand-windows.c
jaykrell: https://unity3d.com/unity/system-requirements

mono/utils/mono-rand-windows.c
lateralusX: Maybe just as a precautions if the methods get called with incorrect handle value since we know what it should be.

mono/utils/mono-rand-windows.c
lateralusX: Jupp, that's why we are still on Vista with mono.

lateralusX: But we could probably still retire the old CAPI implementation and use the implementation from the UWP source file?

mono/utils/mono-rand-windows.c
jaykrell: It will never change. There is no future proofing required.
This is based on NTSTATUS/NT_SUCCESS which hasn't changed since circa 1989. I was nervous about Cygwin headers being complete. But it is marginally clearer yes, marginally more obscure. I'll use it.

mono/utils/mono-rand-windows.c
jaykrell: Agreed I'll take the UWP version. It is pretty close to this. Keep Vista going as Unity requires.

mono/utils/mono-rand-windows.c
jaykrell: It can also be 0. And eventually it'll be ignored. But ok.

You think anyone depends on seedability? I think it is ignored on newer OS but I'll check the docs again.

jaykrell: In terms of the provider value, I think some defense-in-depth has value, and some is pointless.
In this case, pretty pointless. But we can only assert 0 or constant for win7 or newer, which won't be active for now. For Vista support, the value is dynamic.

jaykrell:
Let's see how https://github.com/mono/mono/issues/9102 goes before commiting?

lateralusX:
OK, let's wait and see decision on #9102.

mono/utils/mono-rand-windows.c
```diff
@@ -7,63 +7,20 @@
...
 #include <glib.h>
+#ifdef HOST_WIN32
 #include "mono-error.h"
 #include "mono-error-internals.h"
 #include "mono-rand.h"
-
-#if defined(HOST_WIN32)
 #include <windows.h>
-#include "mono/utils/mono-rand-windows-internals.h"
+#include <brypt.h>
```
lateralusX: Build failure due to misspelled include, should be bcrypt.h.

jaykrell:
Note we can commit this either way. It is a matter of reviewing both versions or not.

jaykrell:
Might conflict with https://github.com/mono/mono/pull/9249

jaykrell:

jaykrell:

jaykrell:

PR #8994, opened 2018-06-01
Import System.Security.Cryptography.Pkcs from CoreFX.

filipnavara:
Requires CoreFX bump and the attached CoreFX patch. I'll submit the relevant CoreFX patches, but I'd appreciate a review. The fully managed implementation is used on all systems for now.

Contributes to https://github.com/mono/mono/issues/7589

PR #9092, opened 2018-06-09
DO NOT MERGE: Import System.Security.Cryptography.Pkcs from CoreFX.

filipnavara:
This is version of https://github.com/mono/mono/pull/8994 with changed references to external modules to test builds on Jenkins.


filipnavara:
@monojenkins build

PR #9093, opened 2018-06-09
Powershell was not working for me and is slow; replace.

jaykrell:


jaykrell:
@monojenkins build failed

lateralusX:
Also, the time spend in these scripts compared to the whole build is a small fraction of time. In worst case where nothing needs to be rebuild it's ~1.5 seconds and as soon as we need to actually build something, the time spend running the powershell script will be marginalized.

jaykrell:
I believe I emulated the existing logic completely:
  1. If config.h does not contain include cygconfig.h, then copy it to cygconfig.h.
  2. Form a temp file from winconfig.h plus versioning in configure.ac
  3. If temp file differs from config.h, copy it to config.h.
  4. Form version.h with constant string, also only if isn't already formed correctly.

The build was failing because the version string wasn't getting replaced. It had the pound signs.

When I ran winsetup.bat manually to debug, it always failed with access denied.
But eventually that seemed to fix itself and it'd work run manually.
I don't know what happened in the build though, just that the version string still had the pound signs in it.

There might also be a missing error check in the old version and the new one. Since build should not have run with the pound signs intact.

lateralusX:
So this sounds more like a problem with the configure step to replace the strings done in the bat file than the powershell scripts. I see that those steps, setting the versions is not run with -executionpolicy bypass flag as the rest of the scripts from the -bat file, maybe that what's caused access denied on your system since by default executionpolicy is restricted. So maybe just setting those commands to use that parameter will solve the problem?

jaykrell:
I haven't yet been convinced to ever use Powershell for anything but yes that is a good theory.
The idea of a pipline of objects is nice, but the design/syntax seems otherwise very flawed.

lateralusX:
I believe you could argue both ways based on background and experience, a lot of people uses powershell and probably think it is way better than .bat files. However, since you already put in effort to rewrite things even though this probably would just be a smaller fix in the existing implementation, I'm not favor the one over the other in this case as long as we don't get regression bugs due to this change. @akoeplinger, any objections going back to .bat files for this particular build step?

msvc/winsetup.bat
```diff
@@ -1,35 +1,65 @@
...
-	ECHO #define FULL_VERSION "Visual Studio built mono"> %VERSION_H%
-)
+set MONO_CORLIB_VERSION=1%MONO_VERSION_MAJOR%%MONO_VERSION_MINOR%%MONO_VERSION_PATCH%%MONO_CORLIB_COUNTER%
 
-ECHO Successfully setup Mono configuration headers.
+:: Remove every define VERSION from config.h and add what we want.
+findstr /v /r /i /c:"#define..*VERSION" %win_config_h% > %monotemp%
+echo #define PACKAGE_VERSION "%MONO_VERSION%" >> %monotemp%
+echo #define VERSION "%MONO_VERSION%" >> %monotemp%
+echo #define MONO_CORLIB_VERSION %MONO_CORLIB_VERSION% >> %monotemp%
```
akoeplinger: it probably won't matter too much but this places the defines outside of the `#endif` in winconfig.h

msvc/winsetup.bat
```diff
@@ -1,35 +1,65 @@
...
-IF NOT %ERRORLEVEL% == 0 (
-	ECHO Configure %VERSION_H%
-	ECHO #define FULL_VERSION "Visual Studio built mono"> %VERSION_H%
-)
+set MONO_CORLIB_VERSION=1%MONO_VERSION_MAJOR%%MONO_VERSION_MINOR%%MONO_VERSION_PATCH%%MONO_CORLIB_COUNTER%
 
-ECHO Successfully setup Mono configuration headers.
+:: Remove every define VERSION from config.h and add what we want.
+findstr /v /r /i /c:"#define..*VERSION" %win_config_h% > %monotemp%
+echo #define PACKAGE_VERSION "%MONO_VERSION%" >> %monotemp%
```
akoeplinger: we should fix the other PACKAGE_ defines as well in winconfig.h source, they currently show libgc values :P

msvc/winsetup.bat
```diff
@@ -1,35 +1,65 @@
...
+:: If the file is different, replace it.
+fc %monotemp% %config_h% >nul || move /y %monotemp% %config_h%
+del %monotemp% 2>nul
+
+echo #define FULL_VERSION "Visual Studio built mono" > %monotemp%
+fc %monotemp% %version_h% || move /y %monotemp% %version_h%
+del %monotemp% 2>nul
+
+:: Show environment variables that start mono.
+set MONO
```
akoeplinger: why do we need this? and I guess the comment should say "start _with_ MONO"

jaykrell: Just logging, not needed.

msvc/winsetup.bat
jaykrell: As long as all the defines are the same, it is ok. You can duplicate defines w/o error or warning.
Since this is Windows-specific file, can also throw in pragma once.

msvc/winsetup.bat
jaykrell: Agreed.

winconfig.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
lateralusX: This file will be used by gcc if we run make again after building with visual studio. I believe this #pragma is supported by gcc, but might need verification (or moved in under the msvc specific part of the file).

msvc/winsetup.bat
```diff
@@ -1,35 +1,65 @@
...
-	ECHO Configure %VERSION_H%
-	ECHO #define FULL_VERSION "Visual Studio built mono"> %VERSION_H%
-)
+set MONO_CORLIB_VERSION=1%MONO_VERSION_MAJOR%%MONO_VERSION_MINOR%%MONO_VERSION_PATCH%%MONO_CORLIB_COUNTER%
 
-ECHO Successfully setup Mono configuration headers.
+:: Remove every define VERSION from config.h and add what we want.
+findstr /v /r /i /c:"#define..*VERSION" %win_config_h% > %monotemp%
+echo #define PACKAGE_VERSION "%MONO_VERSION%" >> %monotemp%
+echo #define VERSION "%MONO_VERSION%" >> %monotemp%
```
akoeplinger: did you check this gets written as `#define VERSION "1.2.3.4"` into the .h file, i.e. it includes the quotes?

lateralusX:
I got problems running this .bat file on a winconfig.h that include other defines including VERSION. There could be more defines in mono using that pattern, so I believe this needs to be a little more robust covering thoose scenarios as well.

winconfig.h
lateralusX: Just verified on local build, seems to be fine under gcc as well.

msvc/winsetup.bat
lateralusX: It will actually be used by cygwin gcc build when doing make again after build from within Visual Studio, just verified that scenario and the define duplication doesn't generate any additonal warnings or errors, so this should be fine.

jaykrell:
I’ll narrow the findstr & double check the quotes.

jaykrell:
Actually time/date can be used portably, in a bit.

jaykrell:
@monojenkins build failed

PR #9095, opened 2018-06-10
Bump corefx (mono/corefx/pull/79)

EgorBo:
With https://github.com/mono/corefx/pull/79

cc @filipnavara


NattyNarwhal:
@marek-safar see #9017 before merging for the CMake checks needed ported to configure.ac; there's not a lot as the others were already checked by Mono

PR #9098, opened 2018-06-11
[dist] Ensure gensources.cs is included in tarballs

directhex:
```
06:34:59 DEBUG: make[8]: *** No rule to make target `../../build/gensources.cs', needed by `../../build/gensources.exe'.  Stop.
```


akoeplinger:
gensources.sh is still used for tests.make so we need to include it.

lewurm:

PR #9103, opened 2018-06-12
Fix canon names not being returned from mono_get_address_info

### LINKED ISSUE #8935: Dns.GetHostEntry no longer returns hostname
## Dns.GetHostEntry no longer returns hostname

Original [bug](https://bugzilla.xamarin.com/show_bug.cgi?id=29279) claimed that it broke in Mono 4.0.
I've tested on 5.8.0.22 (Gentoo) and can reproduce the issue.

    // Works for any IP you wish to test
    // IP for google.com
    var google = Dns.GetHostEntry("216.58.216.14");
    Console.WriteLine("Hostname: " + google.HostName);
    // outputs: "216.58.216.14" instead of "google.com"

This is a particular problem in my use case for an IRCd where DNS lookups of IPs are common for bans (like DNSBLs) or allowing/disallowing ISPs or services.

kg:
This fixes #8935.

How do we typically write automated tests for this? It relies on a known canon name for a given IP address, which seems like it could produce test flakiness between the need for working DNS and the need for the address<->cname pairing to remain constant.

### DIFF
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -75,7 +75,7 @@ mono_get_address_info (const char *hostname, int port, int flags, MonoAddressInf
 /* Some ancient libc don't define AI_ADDRCONFIG */
 #ifdef AI_ADDRCONFIG
 	if (flags & MONO_HINT_CONFIGURED_ONLY)
-		hints.ai_flags = AI_ADDRCONFIG;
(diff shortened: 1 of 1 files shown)


luhenry:

luhenry:
@monojenkins Linux x64 Hybrid Suspend

PR #9104, opened 2018-06-12
Preserve Win32 error in mono_thread_info_uninstall_interrupt.

jaykrell:
An alternate more complete fix for http://github.com/Unity-Technologies/mono/commit/b6f1ba89f7ab285a0990e696e78798f7cf18dafb

### DIFF
--- a/mono/utils/mono-threads.c
+++ b/mono/utils/mono-threads.c
@@ -1598,6 +1598,9 @@ mono_thread_info_install_interrupt (void (*callback) (gpointer data), gpointer d
 void
 mono_thread_info_uninstall_interrupt (gboolean *interrupted)
 {
+#ifdef HOST_WIN32
+	DWORD const win32error = GetLastError ();
+#endif
 	MonoThreadInfo *info;
 	MonoThreadInfoInterruptToken *previous_token;
 
@@ -1621,6 +1624,9 @@ mono_thread_info_uninstall_interrupt (gboolean *interrupted)
 
 	THREADS_INTERRUPT_DEBUG ("interrupt uninstall  tid %p previous_token %p interrupted %s\n",
 		mono_thread_info_get_tid (info), previous_token, *interrupted ? "TRUE" : "FALSE");
+#ifdef HOST_WIN32
+	SetLastError (win32error);
+#endif
 }
 
 static MonoThreadInfoInterruptToken*


jaykrell:
cc @joncham

jaykrell:
Case-by-case basis is kinda expensive to determine.

https://github.com/Unity-Technologies/mono/commit/b6f1ba89f7ab285a0990e696e78798f7cf18dafb has two.

jaykrell:
Another way to put this, lasterror/errno should not be changed in a success path.

jaykrell:
Agreed but preserving it in functions that have no failure mode, known to be used in error/exit paths, also helps modularity and composability with a more tactical change.

Agreed errno/GetLastError are horrible constructs and bottlenecking all uses as shown escape the mess. Functions that return the actual error are much better, and don't break reentrance.
We'd want to essentially remove/change all uses of errno/GetLastError.

In C++ other projects successfully sprinkle in:
struct PreserveLastError_t
{
DWORD err;
PreserveLastError_t() : err(GetLastError()) { }
// reduce spurious write breakpoints
~PreserveLastError_t() { if (GetLastError() != err) SetLastError(err); }
};

and throw in a local of that.

So that is 3+ choices.
1. Fix just mono_thread_info_uninstall_interrupt. This PR.
2. Wrap all errno uses.
3. Fix case by case uses of mono_thread_info_uninstall_interrupt. ReadFile and WriteFile to start.
4. Identify which part of mono_thread_info_uninstall_interrupt is actually the problem and fix it -- TlsGetValue (use __declspec(thread) if not already), g_free?

luhenry:
Unfortunately we don't have C++ so we can't use what you describe.

Adding this construct in all functions - yes there isn't only `mono_thread_info_uninstall_interrupt` that _might_ have this issue and all functions need to be audited - that we need can lead to 2 paths:
 - we "fix" functions as we find bugs, that's going to take years and will never be complete since we might just add a call to another function that doesn't do the trick
 - we "fix" most functions in the runtime ala coop handles, and that's going to be a loss of time and have a negative impact on performance.

The set of functions where we care about the error value is very limited and the call stack is very shallow, so fixing it by wrapping and saving the error there is much easier and has a much more limited set of required changes.

jaykrell:
The lack of C++ here is ok not a big deal. It is just a tweak upon the other approaches, not a separate approach, and still have the exact same remaining decisions -- as to where to use it.

This must not be in all functions -- some return errno -- and I didn't mean to imply that.

It is really about error/exit/cleanup paths.
Things that cannot fail.
g_free and close are good candidates.
Yes I know close can sort of fail, but it only upon bugs, not e.g. low memory, and the error is rarely/never checked, and there is nothing to be done about it.

It is a case by case basis either way.
Either places that require errno preservation, or places that should preserve errno, because they cannot fail and tend to be called in error/exit paths that require it.

Having g_free/close always preserve errno...hard to say if that is good or bad really.
Somewhat bad for perf, though they are already slow (syscall, locking, etc.).
Somewhat good in that that tends to be where it is needed.

I'll fix the two for now, but I still think, functions that cannot fail, and are in error/exit paths, are kinda obligated to do this. A flip side is Rodrigo's point, which is kinda the same as yours -- capture where you need it -- could be a wrapper, or not, but still kinda same thing. Just that after capture -- put it back, or return it separately? Existing flow would be to put it back.

jaykrell:
Will do something like https://github.com/Unity-Technologies/mono/commit/b6f1ba89f7ab285a0990e696e78798f7cf18dafb instead.

jaykrell:
Notice we do have a good precedent here like @kumpera showed:

```gboolean
mono_w32file_move (const gunichar2 *path, const gunichar2 *dest, gint32 *error)
{
	gboolean result;

	MONO_ENTER_GC_SAFE;

	result = MoveFile (path, dest);
	if (!result)
		*error = GetLastError ();

	MONO_EXIT_GC_SAFE;

	return result;
}
```

jaykrell:
I think this pattern works out easily/well, just a few functions, and follows a partial precedent. I'll try.

PR #9105, opened 2018-06-12
Fix quoting of BMPString names in X501Name.

filipnavara:


marek-safar:
This breaks existing System tests

filipnavara:
Working on a fix. The set of characters that need quoting is incorrect, so a slash is included in it and causes the other tests to fail.

PR #9110, opened 2018-06-12
Save a coop/handle frame per PR feedback.

jaykrell:


jaykrell:
The calculus of skipping coop/frames is not as obvious as it initially seemed by the way.
1. Returning a handle, does not imply the caller has a frame, as the caller could be ignoring it. This is why an out parameter is better than a traditional return value. And once you avoid return values, the coop/handle programming loses some of it advantages over over approaches.

2. The caller could be in a loop. One of the points of a function call and its separate frame is to get handle allocation out of loops. This is another problem in the coop/handle programming model.

It is probably ok in this case, but again, it is not as obvious as initially indicated.

jaykrell:
@monojenkins build OS X x64

PR #9111, opened 2018-06-12
Opcodes.def is no longer generated and there are reports of breaks.

jaykrell:


lambdageek:
Why do we have `opcodes.def` and `cil-opcodes.xml`? Why are they out of sync?

This PR promotes `opcodes.def` to be a source of truth, but doesn't do anything to demote `cil-opcodes.xml`.

Would it be better instead to update `cil-opcodes.xml` to bring it back in sync with `opcodes.def` and to re-generate `opcodes.def` from the XML file?

jaykrell:
They are out of sync for multiple reasons.
 - I edited opcodes.def weeks ago.
 - I was expecting this not to work, but it has been.
 - I didn't think we wanted the Perl dependency.

The PR hedges bets. Is the xml valuable and we should repair it? Or not particularly and just delete it?

jaykrell:
Ok, I think I can fix it quickly/easily. Question still not really answered as to what is preferred.

lambdageek:
I don't know if the XML file is used, but it is installed in `<prefix>/share/mono-2.0/cil/`, so someone somewhere may use it.

If we don't/can't get rid of it, I would prefer to continue having a single source file and to generate the others from the source.

jaykrell:
Consider no longer installing it.
In .NET Core. the equivalent is I believe the master.
If we generate it, consider always generating it -- remove opcode.def from git, and output it to the output directory in an out of tree build.
Here I have updated the generation of it.

luhenry:
Since `cil-opcodes.xml` is being shipped, we are not going to simply remove it and hope it doesn't break anyone. So instead of persisting in breaking the synchronization, update it  with wathever changes you made to `opcodes.def` and fix the synchronization.

As for perl, we do not want _more_ of perl, but we are not going to rewrite every existing uses of perl for the sake of removing the dependency on it.

jaykrell:
And should we remove opcode.def from git? So it is always generated?

jaykrell:
We don't have the Perl dependency fully installed in CI.

```
perl ./make-opcodes-def.pl ./cil-opcodes.xml opcode.def.tmp
Can't locate XML/Parser.pm in @INC (you may need to install the XML::Parser module) (@INC contains: /etc/perl
```

I now recall seeing that earlie -- like on my machine, fixed with cpan install something.

jaykrell:
So -- our perl dependency is messed up -- our CI machines do not have it adequately installed. Fix the machines or drop the dependency or other?

akoeplinger:
@jaykrell opcode.def is already in .gitignore so it shouldn't be checked into to the repository: https://github.com/mono/mono/blob/272d70cfdc636b64f712f0ab0e6d4fad636469e4/mono/cil/.gitignore#L3

You likely edited it locally only.

akoeplinger:
Wait sorry, it _is_ there in the repo despite the .gitignore (looks like it was added even before the .gitignore was added) ...

akoeplinger:
So yeah this does look fine but we should make the following changes:

1. Remove opcode.def from .gitignore (since we expect it to be checked in)
2. Edit the comment at the top of opcode.def to be `/* GENERATED FILE, DO NOT EDIT. Edit cil-opcodes.xml instead and run "make opcode.def" to regenerate. */`

jaykrell:
This doesn't seem sufficient. But maybe I'm confused.
Witness that my current PR does break in CI because CI doesn't have the perl dependencies.
Perhaps the small missing piece is remove opcode.def from all. Let me try that.

jaykrell:
That doesn't seem to work because of EXTRA_DIST. I can try CI but not sure it'll work.

jaykrell:
I'm pretty certain this will fail. We either need to:
 1. Give up on generating the file. Can still install the xml for whatever people are doing.
 2. Update CI to have the Perl XML Parser.
 3. Remove opcode.def build rule from Makefile.am and leave it "further on the side", make-opcode-def.sh or such.

mono/cil/opcode.mk
```diff
@@ -0,0 +1,6 @@
+# opcode.def is not built by default.
```
luhenry: This doesn't need to be in a different file, and has always worked as is. Remove this file and revert https://github.com/mono/mono/pull/9111/files#diff-e60505016a74c1c1b451e54c1d999a98L3 and https://github.com/mono/mono/pull/9111/files#diff-e60505016a74c1c1b451e54c1d999a98L13

mono/cil/make-opcodes-def.pl
```diff
@@ -29,34 +29,35 @@
 
 sub handle_opcode {
     my ($parser, $elem, %attrs) = @_;
-    my ($name, $input, $output, $args, $o1, $o2, $flow, $uname, $count, $ff);
+    my ($count);
```
luhenry: The noise of useless changes in this file just drowns the signal of the changes which are actually useful.

jaykrell: The changes are of varying value.
It is broadly acknowledged as better to initialize variables at point of declaration than to leave them uninitialized. So valuable that C99 gained that feature among little else.
It is easier for the human reader, post-diff, to follow the data/control flow, when the resulting data/control flow is smaller.
That trades off with the viewing of the diff to get to the better state.

mono/cil/opcode.mk
jaykrell: This is quite unclear to me -- the resulting dependency on timestamps, and the fact that CI does not have the tools installed to actually generate the file. CI failed earlier as a result. But let's try again. Maybe the timestamps were messed up before and can be reliably controlled.
It seems fragile to me, to ever depend on timestamps.

mono/cil/Makefile.am
```diff
@@ -1,5 +1,10 @@
 include $(top_srcdir)/mk/common.mk
 
+all:
```
jaykrell: I guess line is not needed either. Will remove.

mono/cil/make-opcodes-def.pl
luhenry: I understand the change, that's not the problem. I am just pointing out that the signal of what actually _needed_ to be fixed is drowned into the noise of what you _wanted_ to change.

akoeplinger: I agree with @luhenry's recommendation but on the other hand I feel like we can accept the current state as is without needing to go through another revision/build cycle :)

jaykrell:
@monojenkins build failed

jaykrell:

Issue #9089, opened 2018-06-09
WaitHandle.WaitAny returns invalid result


EgorBo:
## Steps to Reproduce
The following code should print "0" (array index of the object that satisfied the wait)
```csharp
using System;
using System.Threading;

namespace ConsoleApplication11
{
    internal class Program
    {
        public static void Main()
        {         
            ManualResetEvent[] wh = new ManualResetEvent[2];
            wh[0] = new ManualResetEvent(true);
            wh[1] = wh[0];

            Console.WriteLine(WaitHandle.WaitAny(wh));
        }      
    }   
}
```

### Current Behavior

The code prints:
```
2147483647
```

### Expected Behavior
Should print `0` (just like .NET Core 2.1 does)

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

## UPD:
Also, `WaitAll` instead of `WaitAny` in this case should throw `ArgumentException` (mono doesn't throw).

jaykrell:
I believe we malfunction on all the platforms, even Windows somewhat.

jaykrell:
Fix sitting in PR for months. https://github.com/mono/mono/pull/9114

jaykrell:
Fix sitting in PR for months. #9114

marek-safar:
@lateralusX could you please review the fix

PR #9114, opened 2018-06-13
[w32handle] Fix duplicated handles for WaitAny and WaitAll

jaykrell:
Fixes https://github.com/mono/mono/issues/9089


mono/tests/wait-dup-all.cs
```diff
@@ -0,0 +1,26 @@
+/*
+https://github.com/mono/mono/issues/9089
+Duplication is ok for WaitAny, exception for WaitAll.
+System.DuplicateWaitObjectException: Duplicate objects in argument.
+*/
+using System;
+using System.Threading;
+
+class Program
```
kumpera: Those tests would be better included in the corlib test suite. mono/tests is better for tests that must be standalone or require custom process setup.

mono/metadata/threads.c
```diff
@@ -2085,16 +2086,59 @@ ves_icall_System_Threading_WaitHandle_Wait_internal (gpointer *handles, gint32 n
...
-			: mono_w32handle_convert_wait_ret (mono_win32_wait_for_single_object_ex (handles [0], timeoutLeft, TRUE), 1);
+		win32WaitResult = (numhandles != 1)
+			? mono_win32_wait_for_multiple_objects_ex (numhandles, handles, waitall, timeoutLeft, TRUE)
+			: mono_win32_wait_for_single_object_ex (handles [0], timeoutLeft, TRUE);
 		MONO_EXIT_GC_SAFE;
+		// Duplicate detection here is based on CoreCLR.
+		// In particular, it is triggered by an error, and it is handled-based
+		// instead of object-based.
+		// See coreclr/src/vm/threads.cpp CheckForDuplicateHandles and its use.
+		if (win32WaitResult == WAIT_FAILED
```
kumpera: I'm not 100% sure of whether we need to keep this weird split on error detection.
Can't we just have mono_win32_wait_for_multiple_objects_ex return a `MONO_W32HANDLE_WAIT_RET_FAILED_DUPLICATES`?

@luhenry what do you think?

jaykrell: The existing win32 layer is very thin & so I left it alone after brief consideration, but yes that might be a better structure.

mono/tests/wait-dup-all.cs
jaykrell: This area of the tree is much easier for me to work with, which I realize is not the best reason. Editing the makefile, editing-existing/adding-new/running tests, all much easier here.

mono/metadata/threads.c
kumpera: Tests in mono/tests takes a LOT longer to run compared to the unit tests in corlib.

Plus if you're making changes that are easily visible to C#, you should be running that test suite anyways.

jaykrell:
There is a bigger problem here on Win32 actually.

```
MonoOSEventWaitRet
mono_os_event_wait_multiple (MonoOSEvent **events, gsize nevents, gboolean waitall, guint32 timeout, gboolean alertable)
{
...
	else if (res == WAIT_FAILED)
		g_error ("%s: mono_thread_win32_wait_multiple_handle failed with error %d", __func__, 
}
```

We either need to revise this layer, or do the check earlier.

mono/metadata/w32handle.c
```diff
@@ -939,29 +947,46 @@ mono_w32handle_wait_multiple (gpointer *handles, gsize nhandles, gboolean waital
...
+			} else {
+				// There is are duplicates. Remove them -- in-place in
+				// order to return the lowest signaled, equal to the caller's
+				// indices, and ease the exit path's dereference.
+				for (i = 0; i < nhandles; ++i) {
+					for (gsize j = i + 1; j < nhandles; ++j) {
+						if (handles_data [i] == handles_data [j]) {
+							mono_w32handle_unref (handles_data [j]);
+							handles_data [j] = NULL;
+						}
```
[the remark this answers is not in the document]
jaykrell: Actually that data isn't sorted, and operating on the sorted data is tricky, so the code is a bit inefficient but about what it should be. Duplicates will be rare. Worst case is 64x64 but usually much less. The initial check for duplicates will tend to make this entirely skipped, esp. since mono didn't handle this before.

jaykrell:
Ok, I moved the duplicate checking down a later. I used optional MonoError* instead of a new disposition. Optionality feels much safer than letting the new code run in other paths, even if it is rare and an error path, it still has new non-required side effects.

I also moved the tests as requested but haven't managed to run them yet. Running stuff in that part of the tree requires building a bunch more stuff that I haven't succeeded i yet -- trying again from the root, which is known not to work well but might leave enough success behind it.

jaykrell:
[comment was in wrong PR]
Note these are two slightly different layers:
`mono_win32_wait_for_multiple_objects_ex`
`mono_w32handle_wait_multiple`

The second one does not exist on Windows, and contains the first.
Nevertheless, they are quite similar, and adding a MonoError* to both should work.

jaykrell:
Three failures. All interpreter. Merits more investigation.

jaykrell:
oh, wrong PR. maybe flakes. Maybe. Must look closer.

jaykrell:
Need to fix windows build.

jaykrell:
@monojenkins build failed

mono/metadata/w32handle.c
```diff
@@ -131,6 +130,8 @@ mono_w32handle_unlock_signal_mutex (void)
 void
 mono_w32handle_lock (MonoW32Handle *handle_data)
 {
+	if (!handle_data)
```
luhenry: This cannot transparently fail. Check in the caller that we do not lock a invalid handle.

mono/metadata/w32handle.c
```diff
@@ -143,6 +144,8 @@ mono_w32handle_trylock (MonoW32Handle *handle_data)
 void
 mono_w32handle_unlock (MonoW32Handle *handle_data)
 {
+	if (!handle_data)
```
luhenry: This cannot transparently fail. Check in the caller that we do not unlock a invalid handle.

mono/metadata/w32handle.c
```diff
@@ -418,6 +421,9 @@ mono_w32handle_ref_core (MonoW32Handle *handle_data)
 static gboolean
 mono_w32handle_unref_core (MonoW32Handle *handle_data)
 {
+	if (!handle_data)
```
luhenry: This cannot transparently fail. Check in the caller that we do not unref a invalid handle.

mono/utils/mono-error-internals.h
```diff
@@ -241,6 +241,9 @@ mono_error_set_null_reference (MonoError *error)
 	mono_error_set_generic_error (error, "System", "NullReferenceException", "");
 }
 
+// FIXME static inline failed to link for Windows
+#define mono_error_set_duplicate_wait_object(error) (mono_error_set_generic_error (error, "System", "DuplicateWaitObjectException", "Duplicate objects in argument."))
```
luhenry: Have that function like the rest of the other functions: the declaration in this file and the implementation in `mono-error.c`

jaykrell: I'll try that. The link error was surprising.

mono/metadata/w32handle.c
jaykrell: Ignoring null is not failing. Let me see  if it works, there might have been another problem.

luhenry: It fails to do the expected thing, thus breaking assumptions.

jaykrell:
@monojenkins build failed

jaykrell:
Oh, the windows break was  probably lack of include.
Lack of prototype isn't enough to error.

mono/utils/mono-error-internals.h
jaykrell: Problem was missing the include. I not used to the C behavior of allowing to call unprototyped functions and error not until link time but I'll remember it now. At least on Windows -- maybe there is a warning we can/should elevate for that, like the other compilers.

jaykrell:
@monojenkins build failed

jaykrell:
Related-seeming windows failure

jaykrell:
Maybe previous structure of fix better but will look more.

jaykrell:
@monojenkins build failed

mono/metadata/w32handle.c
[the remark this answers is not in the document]
luhenry: You have 2 choices here to avoid the inner-loop:
 - if growing upward (from i = 1 to nhandles), you can just set the `handles_data_sorted [i - 1]` to `NULL` so if `handles_data_sorted [i + 1]` is also a duplicate you haven't changed `handles_data_sorted [i]`
 - if growing downward (from i = nhandles - 1 to 0), you can still set `handles_data_sorted [i]` to `NULL` when comparing it to `handles_data_sorted [i + 1]` so when looking at `handles_data_sorted [i - 1]`, `handles_data_sorted [i]` haven't been changed.

I would go for the first one since it's the minimal change and the result is the same.

jaykrell:
That does not make sense to me. 

The loop is nested in order to find all duplicates of all handles, once we find there are any. Not just the duplicates of the first duplicate found.

Nulling the sorted data does not work, because wait must return the lowest signaled index, against the original unsorted handles. Nulling the sorted data does not help so easily. You’d have to map back to the unsorted data, which is possible but does not seem desirable.

mono/metadata/w32handle.c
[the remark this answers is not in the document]
luhenry: I missed the fact that you are checking `handles_data_sorted` but modifying `handles_data`. Extracting the duplication removal would certainly help in making the code clearer. In simpler terms: please move https://github.com/jaykrell/mono/blob/4493213cd1cc722603e838eb1e5b9977d404ff91/mono/metadata/w32handle.c?utf8=✓#L964-L988 after https://github.com/mono/mono/pull/9114/files#diff-81316870ec3e337f2fd9f434dfbd0a4cR962 loop, using a `duplicated` variable for example.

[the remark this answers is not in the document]
jaykrell: There is really no need to move it. We can do the "action" right after the "detection". There is only one instance of each so no need to flow multiple  "detections" into the one "action". If a variable is introduced, the human reader has to go searching for all the reads and writes, only to find one of each and wonder why the code is split apart. The comments explain everything pretty well here I think.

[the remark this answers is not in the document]
luhenry: Well, I didn't. What makes sense to you might not makes sense to other people and vice-versa.

jaykrell:
@monojenkins build failed

jaykrell:
ok now?

jaykrell:
CI passed. Approve?

jaykrell:
@monojenkins build failed

jaykrell:
Approve?

jaykrell:
Review?

jaykrell:
Review?

jaykrell:
Approve?

jaykrell:
Approve?

mono/metadata/w32handle.c
```diff
@@ -895,14 +902,75 @@ mono_w32handle_wait_one (gpointer handle, guint32 timeout, gboolean alertable)
...
 
+static MonoW32Handle*
+mono_w32handle_has_duplicates (MonoW32Handle *handles [ ], gsize nhandles)
+{
+	if (nhandles < 2 || nhandles > MONO_W32HANDLE_MAXIMUM_WAIT_OBJECTS)
+		return NULL;
+
+	MonoW32Handle *sorted [MONO_W32HANDLE_MAXIMUM_WAIT_OBJECTS]; // 64
+	memcpy (sorted, handles, nhandles * sizeof (handles[0]));
+	qsort (sorted, nhandles, sizeof (gpointer), g_direct_equal);
```
lateralusX: Just for clarity and consistency, `qsort (sorted, nhandles, sizeof  (sorted[0]), g_direct_equal);`

mono/utils/mono-os-wait-win32.c
```diff
@@ -121,29 +123,48 @@ mono_win32_wait_for_single_object_ex (HANDLE handle, DWORD timeout, BOOL alertab
...
 
+	// This is not perfect, but it is the best you can do in usermode and matches CoreCLR.
+	// i.e. handle-based instead of object-based.
+
+	if (result == WAIT_FAILED && waitAll && error &&
+			count > 1 && count <= MAXIMUM_WAIT_OBJECTS
+			&& GetLastError () == ERROR_INVALID_PARAMETER) {
+		gpointer handles_sorted [MAXIMUM_WAIT_OBJECTS]; // 64
+		memcpy (handles_sorted, handles, count * sizeof (handles [0]));
+		qsort (handles_sorted, count, sizeof (gpointer), g_direct_equal);
```
lateralusX: Just for clarity and consistency, `qsort (handles_sorted, count, sizeof  (handles_sorted [0]), g_direct_equal);`

jaykrell:

sourceskyboxer:
And llvm looks failed since "}" error if you see if make compilation throws error. Please fix it!

jaykrell:
Please provide more details. I don't see the error.

PR #9115, opened 2018-06-13
Quash countless warnings until/unless other work is done to address them.

jaykrell:


jaykrell:
Alternatively, it should be easy to fix most of the warnings, or quash them individually.

lateralusX:
Had an ambition several years ago to fix the warnings since several of these actually turns up on other platforms as well, but other platforms already build with either lower warning level or not enable all types of warnings. When we did a test building on a different platform (believe it was macos) with similar warning level as we have on Windows we got into a similar situation (wall of warnings in the build log) and then others believed we should just leave it as is. I didn't disable the warnings back then since I still had the ambition to get around and fix them at some point and disable and hiding them will make it even less likely to fix them going forward. This was however two years ago and no initiative has been taken to get the work done since (this is not just a Windows issue, the issues are on other platforms as well, just hidden). I'm fine to disable them and then maybe we could pick up one by one fixing them over time.

jaykrell:
I'm sure I could fix all of them but I am not confident the PRs would be accepted.

Issue #9109, opened 2018-06-12
[interp][corlib] Assert during corlib test suite.


kumpera:
Found here: https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/6492/parsed_console/log.html

Looks like this: 
```
***** MonoTests.System.Reflection.Emit.TypeBuilderTest.TypeResolveGenericInstances
***** MonoTests.System.Reflection.Emit.TypeBuilderTest.TypeWithFieldRVAWorksUnderSgen
* Assertion at mono-internal-hash.c:126, condition `0' not met
...

Thread 3 (Thread 0x7f7efe5fb700 (LWP 111601)):
#0  0x00007f7f018b7b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00005571c4b83ba1 in mono_handle_native_crash (signal=0x5571c4dd87c6 "SIGABRT", ctx=<optimized out>, info=<optimized out>) at mini-exceptions.c:3156
#2  <signal handler called>
#3  0x00007f7f01323fff in raise () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007f7f0132542a in abort () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00005571c4da05b4 in mono_log_write_logfile (log_domain=<optimized out>, level=<optimized out>, hdr=<optimized out>, message=0x7f7ef801ad10 "* Assertion at mono-internal-hash.c:126, condition `0' not met\n") at mono-log-common.c:135
#6  0x00005571c4db4a2e in monoeg_g_logstr (msg=<optimized out>, log_level=G_LOG_LEVEL_ERROR, log_domain=0x0) at goutput.c:117
#7  monoeg_g_logv_nofree (log_domain=log_domain@entry=0x0, log_level=log_level@entry=G_LOG_LEVEL_ERROR, format=format@entry=0x5571c4dbe098 "* Assertion at %s:%d, condition `%s' not met\n", args=args@entry=0x7f7efe5fab28) at goutput.c:128
#8  0x00005571c4db4e36 in monoeg_assertion_message (format=format@entry=0x5571c4dbe098 "* Assertion at %s:%d, condition `%s' not met\n") at goutput.c:163
#9  0x00005571c4da09dc in mono_internal_hash_table_remove (table=table@entry=0x5571c6c746e8, key=key@entry=0x5571cab59810) at mono-internal-hash.c:126
[...]

PR #9116, opened 2018-06-13
[interp] check existence of InterpMethod before removing it from interp code hastable

lewurm:
When creating dynamic methods, we put those in a queue so they can be collected later via `mono_jit_free_method ()`. the interpreter part was assuming, as soon as we allocate a dynamic method, it will be executed eventually and therefore an InterpMethod must exist in `info.interp_code_hash`, but this isn't necessarily the case.

Fixes https://github.com/mono/mono/issues/9109


lewurm:
@monojenkins squash

lewurm:

lewurm:

lewurm:

PR #9125, opened 2018-06-13
[SDKS,Android] Install cmake when provisioning Android SDK

grendello:
Android NDK supports cmake in addition to the older make-based ndk-build as the
build system and Xamarin.Android is in the process of switching the build of its
native bits to cmake. The reason for this switch is that certain parts of
Xamarin.Android (libmonodroid) are built for both the host system and
the Android target. The host system build currently uses hand-crafted code
directly in .csproj to build the native bits while the Android build uses
ndk-build. This results in duplication of code, compiler parameters are
specified in two locations - it is easy to introduce incompatibilities, bug
fixes may be applied only in one scenario etc.

Using cmake allows us to use a single build "script" for all the targets, thus
making the build more reliable. Android SDK has a cmake package which contains
full support for NDK builds and allows us to easily use cmake across all the
supported platforms without having to install cmake on the host system.

This commit adds the cmake package to the Android SDK provisiong process and,
alas, at the same time adds a workaround to properly unpack the package. Unlike
other Android SDK packages, cmake doesn't unpack to a single subdirectory of the
zip root, instead all its files and directories are placed directly in the zip
root. This requires yet another workaround to install the package properly.

### DIFF
--- a/sdks/builds/unzip-android-archive.sh
+++ b/sdks/builds/unzip-android-archive.sh
@@ -15,14 +15,33 @@ trap cleanup 0
 #    $1: path to source archive
 #    $2: destination directory
 
-# This weird syntax is necessary because some zip archives do not contain a separate
-# entry for the root directory but merely a collection of its subdirectories (for instance
-# platform-tools). The very first entry in the archive is retrieved, then its path is read and
-# finally the part up to the first '/' of the path is retrieved as the root directory of
-# the archive. With platform-tools the last part is necessary because otherwise the root directory
-# would end up being reported as `platform-tools/adb` as this is the first entry in the archive and
-# there's no `platform-tools/` entry
-ZIP_ROOT_DIR=$(unzip -qql "$1" | head -n1 | tr -s ' ' | cut -d' ' -f5- | tr '/' ' ' | cut -d' ' -f1)
+#
+# Special case archives - archives which have no single root directory when unpacked and thus
+#                         the workaround below doesn't work for them.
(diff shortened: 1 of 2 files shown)


sdks/builds/unzip-android-archive.sh
```diff
@@ -15,14 +15,31 @@ trap cleanup 0
...
-# the archive. With platform-tools the last part is necessary because otherwise the root directory
-# would end up being reported as `platform-tools/adb` as this is the first entry in the archive and
-# there's no `platform-tools/` entry
-ZIP_ROOT_DIR=$(unzip -qql "$1" | head -n1 | tr -s ' ' | cut -d' ' -f5- | tr '/' ' ' | cut -d' ' -f1)
+#
+# Special case archives - archives which have no single root directory when unpacked and thus
+#                         the workaround below doesn't work for them.
+# Entry format: grep regex matching the END of path in $1 (no trailing $ necessary)
+#
+SPECIAL_CASE_ARCHIVES="cmake-.*\.zip"
```
jonpryor: For "future maintenance" purposes, I would suggest doing:

```sh
SPECIAL_CASE_ARCHIVES="
    cmake-.*\.zip
    "
```

This will make it "obvious" where to put new entries.

grendello:
The failing checks weren't related to this PR. @luhenry OK to CP this to `2018-02`?

luhenry:

luhenry:

luhenry:

PR #9130, opened 2018-06-14
Replace two loops with one loop.

jaykrell:
Cut the code in roughly half, depending on if the optimizer saw through it.


jaykrell:
Noticed working on bugsweek.

kumpera:
The new code is doing some cryptic boolean magic, the two existing loops a lot simpler to understand even without comments. 👎

marek-safar:
Closing as such changes make Mono code even less attractive for external contributors

jaykrell:
It isn't cryptic really. You see the two loops, almost identical, you wonder why the duplication, you see there is an easy to derive commonality; they just have reversed starting value and reversed termination condition and value, but still checking the same conditions over the same data.

Issue #9031, opened 2018-06-05
HttpWebRequest.EndGetRequestStream throws wrong exception on abort


filipnavara:
## Steps to Reproduce

1. Issue a HTTP request using HttpWebRequest class.
2. Abort the request using HttpWebRequest.Abort after HttpWebRequest.BeginGetRequestStream was called.
3. Call HttpWebRequest.EndGetRequestStream

### Current Behavior

System.Threading.Tasks.TaskCanceledException exception is thrown. Note that these APIs are present in .NET 2.0, where TaskCanceledException didn't even exist yet. Thus applications cannot be aware of this particular exception.

### Expected Behavior

WebException with status WebExceptionStatus.RequestCanceled is thrown.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 5.12.0.226 (2018-02/9824e260f56 Mon Apr 30 18:45:52 EDT 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-master/8b1520c8aae)
	GC:            sgen (concurrent by default)

## Stacktrace

```
:35:52 PM   System.Threading.Tasks.TaskCanceledException: A task was canceled.
5:35:52 PM   at System.Net.HttpWebRequest+<RunWithTimeoutWorker>d__241`1[T].MoveNext () [0x000b5] in /Library/Frameworks/Xamarin.Mac.framework/Versions/4.5.0.280/src/Xamarin.Mac/mcs/class/System/System.Net/HttpWebRequest.cs:952 
5:35:52 PM --- End of stack trace from previous location where exception was thrown ---
5:35:52 PM   at System.Net.HttpWebRequest.EndGetRequestStream (System.IAsyncResult asyncResult) [0x00018] in /Library/Frameworks/Xamarin.Mac.framework/Versions/4.5.0.280/src/Xamarin.Mac/mcs/class/System/System.Net/HttpWebRequest.cs:903 
[...]

PR #9131, opened 2018-06-14
Fix issue #9031

kg:
Fixes issue #9031 by converting task cancellation into a request aborted exception if the Aborted flag is set when we see a cancellation. Also re-enabled the relevant test mentioned by the issue (it seems to pass now with this fix)

### DIFF
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -899,6 +899,11 @@ namespace System.Net
 
 			try {
 				return TaskToApm.End<Stream> (asyncResult);
+			} catch (System.Threading.Tasks.TaskCanceledException tce) {
+				if (Aborted)
+					throw CreateRequestAbortedException ();
+				else
+					throw FlattenException (tce);
 			} catch (Exception e) {
 				throw FlattenException (e);
 			}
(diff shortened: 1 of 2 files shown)


mcs/class/System/System.Net/HttpWebRequest.cs
```diff
@@ -899,6 +899,11 @@ public override Stream EndGetRequestStream (IAsyncResult asyncResult)
 
 			try {
 				return TaskToApm.End<Stream> (asyncResult);
+			} catch (System.Threading.Tasks.TaskCanceledException tce) {
```
baulig: We are using `OperationCanceledException` in most of the other code - if you have a specific reason to use `TaskCanceledException` instead, then please remove the ugly namespace (add `using System.Threading.Tasks` if necessary).

mcs/class/System/System.Net/HttpWebRequest.cs
```diff
@@ -1194,6 +1199,11 @@ public override WebResponse EndGetResponse (IAsyncResult asyncResult)
 
 			try {
 				return TaskToApm.End<HttpWebResponse> (asyncResult);
+			} catch (System.Threading.Tasks.TaskCanceledException tce) {
```
baulig: Same here.

mcs/class/System/Test/System.Net/HttpWebRequestTest.cs
```diff
@@ -567,7 +567,7 @@ public void EndGetRequestStream_AsyncResult_Null ()
 		}
 
 		[Test]
-		[Category ("NotWorking")] // do not get consistent result on MS
```
baulig: I'm not sure whether we should re-enable this, it has the potential to cause problems.  I will add something to the web-tests once I've done cleaning up some other things.  A correct and reliable way of testing this is to wait until the request has actually started - which requires the web-tests' instrumentation framework.

baulig:
I gave this some love during bug week and will have a new PR shortly.

PR #9134, opened 2018-06-14
[corlib] Use String from CoreCLR

marek-safar:


mcs/class/corlib/Test/System/StringTest.cs
```diff
@@ -348,15 +348,7 @@ public unsafe void Constructor7_Value_Invalid ()
...
-		try {
-			new String ((sbyte*) null, 0, 0);
-			Assert.Fail ("#A1");
-		} catch (ArgumentNullException ex) {
-			Assert.AreEqual (typeof (ArgumentNullException), ex.GetType (), "#A2");
-			Assert.IsNull (ex.InnerException, "#A3");
-			Assert.IsNotNull (ex.Message, "#A4");
-			Assert.AreEqual ("value", ex.ParamName, "#A5");
-		}
+		Assert.AreEqual (string.Empty, new String ((sbyte*) null, 0, 0));
```
luhenry: This means we are not expecting an `ArgumentNullException` anymore, correct?

mcs/class/referencesource/mscorlib/system/throwhelper.cs
```diff
@@ -593,7 +593,8 @@ internal enum ExceptionArgument {
         byteOffset,
         minimumBufferSize,
         offset,
-        values
+        values,
+        comparisonType
```
luhenry: You could have `comparisonType,` to avoid the 2-line changes next time you need to add a value to this enum.

mcs/class/corlib/Test/System/StringTest.cs
marek-safar: correct

PR #9137, opened 2018-06-14
[code hashtable] change remove function so that it does not fail on removing a non-existing element

lewurm:
saves us one lookup operation

### DIFF
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -2473,7 +2473,7 @@ static void
 mono_jit_free_method (MonoDomain *domain, MonoMethod *method)
 {
 	MonoJitDynamicMethodInfo *ji;
-	gboolean destroy = TRUE;
+	gboolean destroy = TRUE, removed;
 	GHashTableIter iter;
 	MonoJumpList *jlist;
 	MonoJitDomainInfo *info = domain_jit_info (domain);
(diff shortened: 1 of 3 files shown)


mono/mini/mini-runtime.c
```diff
@@ -2483,8 +2483,7 @@ mono_jit_free_method (MonoDomain *domain, MonoMethod *method)
 	if (mono_use_interpreter) {
 		mono_domain_jit_code_hash_lock (domain);
 		/* InterpMethod is allocated in the domain mempool */
-		if (mono_internal_hash_table_lookup (&info->interp_code_hash, method))
-			mono_internal_hash_table_remove (&info->interp_code_hash, method);
+		mono_internal_hash_table_remove (&info->interp_code_hash, method);
```
kumpera: Shouldn't we add a g_assert to the other call to mono_internal_hash_table_remove?

lewurm: fair enough. updated!

lewurm:
@monojenkins squash

mono/mini/mini-runtime.c
```diff
@@ -2482,9 +2482,9 @@ mono_jit_free_method (MonoDomain *domain, MonoMethod *method)
 
 	if (mono_use_interpreter) {
 		mono_domain_jit_code_hash_lock (domain);
-		/* InterpMethod is allocated in the domain mempool */
-		if (mono_internal_hash_table_lookup (&info->interp_code_hash, method))
-			mono_internal_hash_table_remove (&info->interp_code_hash, method);
+		/* InterpMethod is allocated in the domain mempool. We might haven't
```
luhenry: `We might not have`

PR #9138, opened 2018-06-14
Fix IndexOutOfRangeException: No adapter found for index when getting network information

mderoy:
GetAllNetworkInterfaces (correctly) returns the loopback, however GetAdaptersInfo does not return information about the loopback 
msdn https://msdn.microsoft.com/en-us/library/windows/desktop/aa365917(v=vs.85).aspx 

The GetAdaptersInfo and GetInterfaceInfo functions do not return information about the IPv4 loopback interface. Information on the loopback interface is returned by the GetIpAddrTable function)

Instead of querying data from GetAdaptersInfo, instead use data from Win32_IP_ADAPTER_ADDRESSES

I've manually tested the output of the script below against .NetFW; If theres a better way to test networking changes to windows please tell me.

```
public void Test()
{
	var interfaces = NetworkInterface.GetAllNetworkInterfaces();
	foreach (NetworkInterface ni in interfaces)
	{
		var ipv4Properties = ni.GetIPProperties().GetIPv4Properties();
		UnityEngine.Debug.Log(ni.Name + " " + ipv4Properties.Index);
		var isDHCP = ipv4Properties.IsDhcpEnabled;
		UnityEngine.Debug.Log("isDHCP " + isDHCP);
		var iswins = ipv4Properties.UsesWins;
		UnityEngine.Debug.Log("UsesWins " + iswins);
		var dhcpserveraddrcol = ni.GetIPProperties().DhcpServerAddresses;
		foreach (var dhcp in dhcpserveraddrcol)
			UnityEngine.Debug.Log("dhcpserver " + dhcp.Address);
		var gatewayipaddressinformationcol = ni.GetIPProperties().GatewayAddresses;
		foreach (var gateway in gatewayipaddressinformationcol)
			UnityEngine.Debug.Log("gatewayserver " + gateway.Address);
		var unicastipaddresinformationcol = ni.GetIPProperties().UnicastAddresses;
		foreach (var unicast in unicastipaddresinformationcol)
			UnityEngine.Debug.Log("unicastserver " + unicast.Address);
		var winsipaddresscol = ni.GetIPProperties().WinsServersAddresses;
		foreach (var ftw in winsipaddresscol)
			UnityEngine.Debug.Log("winsserver " + ftw.Address);
	}
}

### DIFF
--- a/mcs/class/System/System.Net.NetworkInformation/Win32IPAddressCollection.cs
+++ b/mcs/class/System/System.Net.NetworkInformation/Win32IPAddressCollection.cs
@@ -79,6 +79,25 @@ namespace System.Net.NetworkInformation {
 			return c;
 		}
 
+		public static Win32IPAddressCollection FromSocketAddress (Win32_SOCKET_ADDRESS addr)
+		{
+			Win32IPAddressCollection c = new Win32IPAddressCollection ();
+			if (addr.Sockaddr != IntPtr.Zero)
+				c.InternalAdd (addr.GetIPAddress ());
+			return c;
+		}
+
+		public static Win32IPAddressCollection FromWinsServer (IntPtr ptr)
+		{
+			Win32IPAddressCollection c = new Win32IPAddressCollection ();
+			Win32_IP_ADAPTER_WINS_SERVER_ADDRESS a;
+			for (IntPtr p = ptr; p != IntPtr.Zero; p = a.Next) {
+				a = (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS) Marshal.PtrToStructure (p, typeof (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS));
+				c.InternalAdd (a.Address.GetIPAddress ());
+			}
+			return c;
+		}
+
 		void AddSubsequentlyString (IntPtr head)
 		{
 			Win32_IP_ADDR_STRING a;
--- a/mcs/class/System/System.Net.NetworkInformation/Win32IPInterfaceProperties.cs
+++ b/mcs/class/System/System.Net.NetworkInformation/Win32IPInterfaceProperties.cs
@@ -44,13 +44,11 @@ namespace System.Net.NetworkInformation {
 
 		public override IPv4InterfaceProperties GetIPv4Properties ()
 		{
-			Win32_IP_ADAPTER_INFO v4info = Win32NetworkInterface2.GetAdapterInfoByIndex (mib4.Index);
-			return new Win32IPv4InterfaceProperties (v4info, mib4);
+			return new Win32IPv4InterfaceProperties (addr, mib4);
 		}
 
 		public override IPv6InterfaceProperties GetIPv6Properties ()
 		{
-			Win32_IP_ADAPTER_INFO v6info = Win32NetworkInterface2.GetAdapterInfoByIndex (mib6.Index);
 			return new Win32IPv6InterfaceProperties (mib6);
 		}
 
@@ -74,10 +72,9 @@ namespace System.Net.NetworkInformation {
 
 		public override IPAddressCollection DhcpServerAddresses {
 			get {
-				Win32_IP_ADAPTER_INFO v4info = Win32NetworkInterface2.GetAdapterInfoByIndex (mib4.Index);
 				// FIXME: should ipv6 DhcpServer be considered?
 				try {
-					return new Win32IPAddressCollection (v4info.DhcpServer);
+					return Win32IPAddressCollection.FromSocketAddress (addr.Dhcpv4Server);
 				} catch (IndexOutOfRangeException) {
 					return Win32IPAddressCollection.Empty;
 				}
@@ -96,25 +93,14 @@ namespace System.Net.NetworkInformation {
 			get {
 				var col = new GatewayIPAddressInformationCollection ();
 				try {
-					Win32_IP_ADAPTER_INFO v4info = Win32NetworkInterface2.GetAdapterInfoByIndex (mib4.Index);
 					// FIXME: should ipv6 DhcpServer be considered?
-
-					var a = v4info.GatewayList;
-					if (!String.IsNullOrEmpty (a.IpAddress)) {
-						col.InternalAdd(new SystemGatewayIPAddressInformation(IPAddress.Parse (a.IpAddress)));
-						AddSubsequently (a.Next, col);
+					Win32_IP_ADAPTER_GATEWAY_ADDRESS a;
+					for (IntPtr p = addr.FirstGatewayAddress; p != IntPtr.Zero; p = a.Next) {
+						a = (Win32_IP_ADAPTER_GATEWAY_ADDRESS) Marshal.PtrToStructure (p, typeof (Win32_IP_ADAPTER_GATEWAY_ADDRESS));
+						col.InternalAdd (new SystemGatewayIPAddressInformation (a.Address.GetIPAddress ()));
 					}
 				} catch (IndexOutOfRangeException) {}
 				return col;
-			}
-		}
-
-		static void AddSubsequently (IntPtr head, GatewayIPAddressInformationCollection col)
-		{
-			Win32_IP_ADDR_STRING a;
-			for (IntPtr p = head; p != IntPtr.Zero; p = a.Next) {
-				a = (Win32_IP_ADDR_STRING) Marshal.PtrToStructure (p, typeof (Win32_IP_ADDR_STRING));
-				col.InternalAdd (new SystemGatewayIPAddressInformation (IPAddress.Parse (a.IpAddress)));
 			}
 		}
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by akoeplinger** [comment]
@lateralusX could you please take a look?

File: mcs/class/System/System.Net.NetworkInformation/IPAddressCollection.cs [hunk 1]
```diff
@@ -84,6 +84,25 @@ public static Win32IPAddressCollection FromDnsServer (IntPtr ptr)
...
+				c.InternalAdd (addr.GetIPAddress ());
+			return c;
+		}
+
+		public static Win32IPAddressCollection FromWinsServer (IntPtr ptr)
+		{
+			Win32IPAddressCollection c = new Win32IPAddressCollection ();
+			Win32_IP_ADAPTER_WINS_SERVER_ADDRESS a;
+			for (IntPtr p = ptr; p != IntPtr.Zero; p = a.Next) {
+				a = (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS) Marshal.PtrToStructure (p, typeof (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS));
```
lateralusX: Maybe do some sanity check that we at least have the expected struct length (included as a field in the unmanaged struct)?

mcs/class/System/System.Net.NetworkInformation/IPInterfaceProperties.cs
```diff
@@ -393,28 +390,17 @@ static IPAddressInformationCollection Win32FromAnycast (IntPtr ptr)
...
-					Win32_IP_ADAPTER_INFO v4info = Win32NetworkInterface2.GetAdapterInfoByIndex (mib4.Index);
 					// FIXME: should ipv6 DhcpServer be considered?
-
-					var a = v4info.GatewayList;
-					if (!String.IsNullOrEmpty (a.IpAddress)) {
-						col.InternalAdd(new SystemGatewayIPAddressInformation(IPAddress.Parse (a.IpAddress)));
-						AddSubsequently (a.Next, col);
+					Win32_IP_ADAPTER_GATEWAY_ADDRESS a;
+					for (IntPtr p = addr.FirstGatewayAddress; p != IntPtr.Zero; p = a.Next) {
+						a = (Win32_IP_ADAPTER_GATEWAY_ADDRESS) Marshal.PtrToStructure (p, typeof (Win32_IP_ADAPTER_GATEWAY_ADDRESS));
```
lateralusX: Maybe do some sanity check that we at least have the expected struct length (included as a field in the unmanaged struct)?

mcs/class/System/System.Net.NetworkInformation/IPAddressCollection.cs
mderoy: Did you want something like this?
```
		public static Win32IPAddressCollection FromWinsServer (IntPtr ptr)
		{
			Win32IPAddressCollection c = new Win32IPAddressCollection ();
			Win32_IP_ADAPTER_WINS_SERVER_ADDRESS a;
			for (IntPtr p = ptr; p != IntPtr.Zero; p = a.Next) {
				a = (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS) Marshal.PtrToStructure (p, typeof (Win32_IP_ADAPTER_WINS_SERVER_ADDRESS));
				if(a.LengthFlags.Length != Marshal.SizeOf(typeof(Win32_IP_ADAPTER_WINS_SERVER_ADDRESS))) 
						return new Win32IPAddressCollection ();

				c.InternalAdd (a.Address.GetIPAddress ());
			}
			return c;
		}
```
Then if any of our list of Win32_IP_ADAPTER_WINS_SERVER_ADDRESS are junk than we'll return an empty address collection.

mcs/class/System/System.Net.NetworkInformation/IPInterfaceProperties.cs
mderoy: Same comment above would apply to this function as well

mcs/class/System/System.Net.NetworkInformation/NetworkInterface.cs
```diff
@@ -465,9 +465,13 @@ class Win32NetworkInterfaceAPI : NetworkInterfaceFactory
 			{
 				IntPtr ptr = IntPtr.Zero;
 				int len = 0;
-				GetAdaptersAddresses (0, 0, IntPtr.Zero, ptr, ref len);
+				uint flags = Win32_IP_ADAPTER_ADDRESSES.GAA_FLAG_INCLUDE_WINS_INFO | Win32_IP_ADAPTER_ADDRESSES.GAA_FLAG_INCLUDE_GATEWAYS;
+				GetAdaptersAddresses (0, flags, IntPtr.Zero, ptr, ref len);
+				if (Marshal.SizeOf (typeof (Win32_IP_ADAPTER_ADDRESSES)) > len)
+					throw new NetworkInformationException ();
+
 				ptr = Marshal.AllocHGlobal(len);
```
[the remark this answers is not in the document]
mderoy: It looks like the current Status Quo is that we leak a little every call.

mcs/class/System/System.Net.NetworkInformation/NetworkInterface.cs
```diff
@@ -465,9 +465,13 @@ class Win32NetworkInterfaceAPI : NetworkInterfaceFactory
 			{
 				IntPtr ptr = IntPtr.Zero;
 				int len = 0;
-				GetAdaptersAddresses (0, 0, IntPtr.Zero, ptr, ref len);
+				uint flags = Win32_IP_ADAPTER_ADDRESSES.GAA_FLAG_INCLUDE_WINS_INFO | Win32_IP_ADAPTER_ADDRESSES.GAA_FLAG_INCLUDE_GATEWAYS;
+				GetAdaptersAddresses (0, flags, IntPtr.Zero, ptr, ref len);
+				if (Marshal.SizeOf (typeof (Win32_IP_ADAPTER_ADDRESSES)) > len)
```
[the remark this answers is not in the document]
mderoy: Are you suggesting checking the length on the Alignment structure in Win32_IP_Adapter_Addresses and comparing that against the size of the managed struct to make sure we are on a windows version that has all the fields we need? (I guess this is similar to the sanity checks above)

mderoy:
@lateralusX Any response to the above questions? I've been waiting before looking at this PR again.

mcs/class/System/System.Net.NetworkInformation/NetworkInterface.cs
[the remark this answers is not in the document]
lateralusX: OK

mcs/class/System/System.Net.NetworkInformation/NetworkInterface.cs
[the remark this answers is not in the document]
lateralusX: Yes that was the idea (to check each converted item) since the above check just make sure there is at least one full Win32_IP_ADAPTER_ADDRESSES returned, but if there are multiple adapters returned (where each Win32_IP_ADAPTER_ADDRESSES is smaller than expected since we are on an earlier Windows version) it won't detect that condition. I did a check in corefx and it doesn't look like they do this additional checks though so maybe it is overkill. I'm fine leaving it as is.

mcs/class/System/System.Net.NetworkInformation/IPAddressCollection.cs
lateralusX: So the check should be that we don't get a a.LengthFlags.Length that is smaller than what we expect (meaning that we are on an earlier unsupported version of Windows). I did a check in corefx and it doesn't look like they do this additional checks though so maybe it is overkill. I'm fine leaving it as is.

mcs/class/System/System.Net.NetworkInformation/IPInterfaceProperties.cs
lateralusX: OK, same comment as above.

lateralusX:
@mderoy Sorry for delay (was on vacation), the additional checks was just an extra safety net to make sure we detect if we run on unsupported Windows version, but maybe this is overkill since we move up to Win7 as minimal supported platform anyways. Just took a quick look at corefx as well and they seemed to marshal the structs without additional sanity checks as well, so I'm fine leaving it as is. Just rebase changes and make sure we have pass on all tests and I believe we are good to go.

mderoy:
This may not be necessary to upstream anymore...it would seem a lot of the code has now been pulled out (I assume replaced with corefx) I'll download a nightly windows mono and see if I can still reproduce the issue.

mderoy:
Turns out the files were either renamed, or the code extracted into "Win32" versions of these. The issue still exists on a recent nightly. I've done the rebase I need to build and test this locally

mderoy:
Built and tested the sample code in the description locally after the rebase. The results seem the same between .net and mono @lateralusX

lateralusX:
@monojenkins build failed

mderoy:
@lateralusX I'm seeing "All checks have passed" could you point to me what failed so I can look at it?

lateralusX:
That was before I did the rebuild of this PR (build failure of Linux builds). Now with all passed we are good to go!

PR #9150, opened 2018-06-14
[Linux,XA build] Fix building of Windows cross-compilers on Ubuntu Linux

grendello:
Ubuntu Linux (possibly others too, but only Ubuntu is tested thus the
limitation) don't need to build or use `mxe`, they instead use mingw packaged
for the distribution, thus making the build faster.

This commit detects the Linux flavor in use and, if it's Ubuntu, configures the
build so that system mingw is used instead of mxe.

### DIFF
--- a/sdks/builds/determine-linux-flavor.sh
+++ b/sdks/builds/determine-linux-flavor.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+    FLAVOR=$NAME
+elif [ -n "$(which lsb_release)" ]; then
+    FLAVOR=$(lsb_release -si)
+elif [ -f /etc/lsb-release ]; then
+    . /etc/lsb-release
+    FLAVOR=$DISTRIB_ID
+elif [ -f /etc/debian_version ]; then
+    FLAVOR=Debian
+else
+    FLAVOR=$(uname -s)
+fi
+
+echo $FLAVOR
--- a/sdks/builds/mxe.mk
+++ b/sdks/builds/mxe.mk
@@ -1,4 +1,15 @@
+.PHONY: provision-mxe
 
+ifeq ($(UNAME),Linux)
+LINUX_FLAVOR=$(shell ./determine-linux-flavor.sh)
+endif
+
+ifeq ($(LINUX_FLAVOR),Ubuntu)
+MXE_PREFIX=/usr
+
+provision-mxe:
+	@echo $(LINUX_FLAVOR) Linux does not require mxe provisioning. mingw from packages is used instead
+else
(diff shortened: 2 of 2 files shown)


sdks/builds/Makefile
```diff
@@ -14,6 +14,11 @@ UNAME=$(shell uname)
 ifneq ($(UNAME),Darwin)
 # iOS requires Xcode to be available, and Xcode is only available on macOS
 DISABLE_IOS=1
+
+ifeq ($(UNAME),Linux)
+LINUX_FLAVOR=$(shell ./determine-linux-flavor.sh)
```
luhenry: Since this is only useful for `mxe`, put it in `mxe.mk` instead.

sdks/builds/determine-linux-flavor.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+if [ -f /etc/os-release ]; then
```
luhenry: Why isn't `lsb_release -s -i` sufficient?

luhenry:
Also, my understanding was that mxe is only useful on macOS since all the other linux can just install the corresponding gcc.

sdks/builds/determine-linux-flavor.sh
grendello: because it may not exist on some systems. LSB is being "phased out"

sdks/builds/Makefile
grendello: For now, but would it be bad to keep it in the top level file where it can be exported for all/any other bits which may need special treatment?

grendello:
> Also, my understanding was that mxe is only useful on macOS since all the other linux can just install the corresponding gcc.

That might be, but I didn't test it on other Linux distributions and I'd prefer to be conservative here and use it only where I'm sure it works. It's easy to update the code if other distros have mingw available, if such need arises

sdks/builds/Makefile
luhenry: We can always move it back here if it's needed.

sdks/builds/determine-linux-flavor.sh
luhenry: Is XA supporting any distribution which doesn't have `lsb_release`?

grendello: I honestly don't know... We support "Linux" in general - I'd like to stay as open with the support as possible, but I don't want to make assumptions about something working or not on other distributions than the ones I (or somebody else) tested.

grendello:
@luhenry can we CP this to the `2018-*` branches as well?

luhenry:

luhenry:

luhenry:

Issue #6947, opened 2018-02-12
Intermittent failure of process-stress-3.exe test


alexischr:
Example: https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot/7760/testReport/junit/MonoTests/test-process-stress/process_stress_3_exe/

lewurm:
also happened here: https://jenkins.mono-project.com/job/test-mono-pull-request-amd64/11301/

PR #9152, opened 2018-06-14
[tests] reduce IO load in process-stress-3

lewurm:
The spawned process by the test iterates through the whole mono repo,
including submodules. On a loaded machine with a already slow I/O
device, 10 seconds could be a tight time budget. Instead of increasing
the timeout, let's try to reduce the generated output.

Fixes https://github.com/mono/mono/issues/6947 hopefully.


lewurm:
@monojenkins squash

PR #9157, opened 2018-06-15
Don't call mono_thread_info_current_unchecked for non-alertable waits.

jaykrell:


jaykrell:
Note these are two slightly different layers:
`mono_win32_wait_for_multiple_objects_ex`
`mono_w32handle_wait_multiple`

The second one does not exist on Windows, and contains the first.
Nevertheless, they are quite similar, and adding a MonoError* to both should work.

PR #9158, opened 2018-06-15
[Utils] Add return from infinity loops in MonoConcurrentHashTable

InHavk:
For situations when current key and NULL are not present in keys of hashtable

For the users, this looks like 100% CPU.
backtrace in gdb show following backtraces:
```
Thread 13 (Thread 0x7ffa838ff700 (LWP 5424)):
#0  0x00000000006e7bc2 in mono_conc_hashtable_lookup (hash_table=0x1ea03f0, key=key@entry=0x7ffa788687d0) at mono-conc-hashtable.c:174
#1  0x000000000042f9c8 in mono_jit_runtime_invoke (method=0x7ffa788687d0, obj=0x0, params=0x7ffa838f88c0, exc=0x0, error=0x7ffa838f89c0) at mini-runtime.c:2668
#2  0x00000000005faa9e in do_runtime_invoke (method=0x7ffa788687d0, obj=<optimized out>, params=<optimized out>, exc=<optimized out>, error=0x7ffa838f89c0) at object.c:2922
#3  0x00000000006060f5 in mono_runtime_try_invoke_array (method=method@entry=0x7ffa788687d0, obj=obj@entry=0x0, params=params@entry=0x7ffb02b36930, exc=exc@entry=0x0, error=error@entry=0x7ffa838f89c0) at object.c:5261
#4  0x0000000000606490 in mono_runtime_invoke_array_checked (method=method@entry=0x7ffa788687d0, obj=obj@entry=0x0, params=params@entry=0x7ffb02b36930, error=error@entry=0x7ffa838f89c0) at object.c:5139
#5  0x00000000005ba3da in ves_icall_InternalInvoke (method=0x7ffb02b36908, this_arg=0x0, params=0x7ffb02b36930, exc=0x7ffa838f8b30) at icall.c:3392
#6  0x0000000041a4822a in ?? ()
#7  0x00007ffb02acb5a0 in ?? ()
#8  0x00007ffb03ed14b0 in ?? ()
#9  0x00007ffb02acb510 in ?? ()
#10 0x0000000000000000 in ?? ()
```

The memory dump analysis provided information that `kvs` does not have NULL in keys. Only keys and TOMBSTONE.

Similar in description problems:
- https://stackoverflow.com/questions/50400431/mono-100-cpu-spin-in-mono-conc-hashtable-lookup
- https://github.com/xplicit/HyperFastCgi/issues/86

### DIFF
--- a/mono/utils/mono-conc-hashtable.c
+++ b/mono/utils/mono-conc-hashtable.c
@@ -159,7 +159,7 @@ mono_conc_hashtable_lookup (MonoConcurrentHashTable *hash_table, gpointer key)
 {
 	MonoThreadHazardPointers* hp;
 	conc_table *table;
-	int hash, i, table_mask;
+	int hash, i, table_mask, first_i;
 	key_value_pair *kvs;
 	hash = mix_hash (hash_table->hash_func (key));
 	hp = mono_hazard_pointer_get ();
@@ -169,6 +169,7 @@ retry:
 	table_mask = table->table_size - 1;
 	kvs = table->kvs;
 	i = hash & table_mask;
+	first_i = i;
 
 	if (G_LIKELY (!hash_table->equal_func)) {
 		while (kvs [i].key) {
@@ -182,6 +183,9 @@ retry:
 				return value;
 			}
 			i = (i + 1) & table_mask;
+			if (i == first_i) {
+				break;
+			}
 		}
 	} else {
 		GEqualFunc equal = hash_table->equal_func;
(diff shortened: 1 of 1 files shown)


kumpera:
I’ll take a look at this issue later today. Looks pretty serious. We have a second implementation that is similar and would need to be fixed as well.

alexanderkyte:
@jaykrell the characteristic that differentiates this hash table from a normal GHashTable is the lockless reads. I think it matters to us.

kumpera:
Hey @InHavk,

Thanks for the PR, but I believe a better solution it to scrub those tombstones during removal as this will ensure lookup remains efficient.

I implemented it here: https://github.com/mono/mono/pull/9176

PR #9162, opened 2018-06-15
Improve certificate type detection and exception messages on import

filipnavara:
One unit test is removed as part of the commit because it is not consistent with documentation and CoreFX behaviour on non-Windows platforms (https://github.com/dotnet/corefx/issues/30416).

The main goal of this patch is to actually get meaningful error messages when constructing X509Certificate2 from a byte array. Previously PKCS#12/PFX certificate import could fail due to unimplemented features (such as ECDsa support) or bugs. The certificate import would then be retried as PKCS#7 and the original error message was lost.


filipnavara:
I know, but there are still pieces of code that don't use AppleTLS / BTLS. This is just a stepping stone to make the APIs report correct errors and let me understand what pieces are broken and need further attention.

baulig:
Well, I really appreciate your contribution and fully agree that it's a helpful stepping stone in the right direction.  Just wanted to let you know that this code might eventually be replaced, so you don't feel like you wasted your time if you see these files replaced with CoreFX in 6-months from now.

filipnavara:
I'd be more than happy if that happens :)

marek-safar:
@monojenkins commit apidiff

PR #9163, opened 2018-06-15
Fix non-interpreter 5-60 second hang at shutdown of thread6 (and thread7).

jaykrell:
(replacement for https://github.com/mono/mono/pull/9042 if it goes well)
regressed back in December with commit 5f5c5e97a08f7086d7c18af37352c4a03dc4c0d1

Because if there is an uncleared AbortRequest, suspend will be skipped, but the interrupt callback will still be called.

The skip is here:
```
request_thread_abort:
	if (thread->state & (ThreadState_AbortRequested | ThreadState_Stopped))
		return FALSE;
```

The interrupt callback anyway is, well, I debugged it a few days ago and don't see it now, would have to step through it again.

This is also the change that slowed down Alpine/muslc to be unusable. This change should be tested there too to see if it fixes it. The correlation between test/thread6 and anything else is unclear -- i.e. is thread abort so common, in the build?

### DIFF
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -477,6 +477,15 @@ worker_thread (gpointer unused)
 		if (mono_thread_interruption_checkpoint_bool ())
 			continue;
 
+		// If a worker thread is in its native top, not running managed code,
+		// there is no point in raising thread abort, and no code will clear
+		// the abort request. As such, the subsequent timedwait, would
+		// not be interrupted at runtime shutdown, because an abort is already requested.
+		// Clear the abort request.
+		// This avoids a shutdown hang in tests thread6 and thread7.
+		if (thread->state & ThreadState_AbortRequested)
+			mono_thread_internal_reset_abort (thread);
+
 		if (!work_item_try_pop ()) {
 			gboolean timeout;
 
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -487,9 +487,7 @@ worker_thread (gpointer unused)
 			mono_thread_internal_reset_abort (thread);
 
 		if (!work_item_try_pop ()) {
-			gboolean timeout;
-
-			timeout = worker_park ();
+			gboolean const timeout = worker_park ();
 			if (timeout)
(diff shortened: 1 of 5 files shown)


jaykrell:
This is how used to work, which does not agree with my earlier analysis.
Because request_thread_abort should have returned FALSE and this would be skipped too.

```
-> 370 		if (!mono_runtime_is_shutting_down ())
   371 			return;
   372 	
   373 		if (!mono_refcount_tryinc (&worker))
Target 0: (mono) stopped.
(lldb) bt
* thread #1, name = 'tid_307', queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001002a5bec mono`worker_wait_interrupt(unused=0x0000000000000000) at threadpool-worker-default.c:370
    frame #1: 0x00000001004fe1f1 mono`mono_thread_info_finish_interrupt(token=0x0000000100908080) at mono-threads.c:1687
    frame #2: 0x00000001003a93ae mono`async_abort_internal(thread=0x0000000100dd0650, install_async_abort=1) at threads.c:5290
    frame #3: 0x00000001003a9747 mono`mono_thread_internal_abort(thread=0x0000000100dd0650, appdomain_unload=0) at threads.c:2543
    frame #4: 0x00000001003aba86 mono`abort_threads(key=0x000070000fc43000, value=0x0000000100dd0650, user=0x00007ffeefbff188) at threads.c:3438

```

jaykrell:
To correct some analysis: In the old working state, what interrupted the aborted thread, was the interrupt callback indeed, but not on that thread. It is skipped on the aborted thread, but broadcast from another worker thread. Limiting the worker thread pool count to 1 would probably break the old code as well.

jaykrell:

jaykrell:

jaykrell:

jaykrell:
Two failures of finalizer test on Linux Aarch64.

jaykrell:

jaykrell:
Could be this exposing what was flaky.

jaykrell:
Interpreter still fails with this fix. :(

BrzVlad:
@jaykrell I can look into the interpreter hangs

Issue #9088, opened 2018-06-09
MutatingMethods_MutationsAffectOriginal test fails (imported from CoreFX)


EgorBo:
## Steps to Reproduce

The following test imported from CoreFX fails on mono (and works fine on .NET Core 2.1):

```csharp
using System;

namespace ConsoleApplication11
{
    internal class Program
    {
        public static void Main()
        {
            MutatingStruct? ms = new MutatingStruct() { Value = 1 };

            for (int i = 1; i <= 4; i++)
            {
                AssertEqual(i.ToString(), ms.Value.ToString());
                AssertEqual(i, ms.Value.Value);

                AssertEqual(i.ToString(), ms.ToString());
                AssertEqual(i + 1, ms.Value.Value);
            }

            for (int i = 5; i <= 6; i++)
            {
                ms.Value.Equals(new object());
                AssertEqual(i, ms.Value.Value);

                ms.Equals(new object());
                AssertEqual(i + 1, ms.Value.Value);
            }
        }

        private static void AssertEqual(object o1, object o2)
        {
            if (!o1.Equals(o2))
                throw new Exception($"{o1} != {o2}");
        }

        private struct MutatingStruct
        {
            public int Value;
            public override string ToString() => Value++.ToString();
            public override bool Equals(object obj) => Value++.Equals(null);
            public override int GetHashCode() => Value++.GetHashCode();
        }
    }   
}
```

### Current Behavior
```
Unhandled Exception:
System.Exception: 6 != 5
```

### Expected Behavior

Exception is not thrown (.NET Core 2.1)

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

marek-safar:
@vargaz please add it to your bugs week list

vargaz:
ms.Equals(new object());
doesn't seem to call MutatingStruct.Equals ().

vargaz:
2 problems here:
ms.Equals(new object());
returns false without calling ms.Value.Equals, since object () is not the same type.

it should be changed to:
                ms.Equals(new MutatingStruct() { Value = 1 });

Also, Nullable.Equals () does:
			return other.value.Equals (value);
this should be changed to
value.Equals (other.value);

lewurm:
FWIW, the interp has the same result. it's a problem in the BCL imho.

jaykrell:
Mono BCL that is?

vargaz:
Should we enable the original test ?

marek-safar:
[...]

PR #9165, opened 2018-06-15
[System] align Nullable.Equals behaviour with .NET core

lewurm:
Our `Nullable` implementation is slightly different regarding `Equals ()`.

mono:
https://github.com/mono/mono/blob/028ff0d34c8e8cbe1e819e49962775841972fc62/mcs/class/corlib/System/Nullable.cs#L111-L130

corefx:
https://github.com/mono/corefx/blob/ab816c3efdb1aedb4baa1b8d2fec8ee2f8e15683/src/Common/src/CoreLib/System/Nullable.cs#L62-L67

Fixes https://github.com/mono/mono/issues/9088

### DIFF
--- a/mcs/class/corlib/System/Nullable.cs
+++ b/mcs/class/corlib/System/Nullable.cs
@@ -110,23 +110,11 @@ namespace System
 
 		public override bool Equals (object other)
 		{
+			if (!has_value)
+				return other == null;
 			if (other == null)
-				return has_value == false;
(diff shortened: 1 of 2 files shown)


akoeplinger:
Should we close this in favor of https://github.com/mono/mono/pull/9165 ?

marek-safar:
I think it's ok to land it now as it also has a test, then #9181 will just delete the implementation file

PR #9168, opened 2018-06-15
Detect the private key type from the supplied OID instead of guessing from the content.

filipnavara:

### DIFF
--- a/mcs/class/Mono.Security/Mono.Security.X509/PKCS12.cs
+++ b/mcs/class/Mono.Security/Mono.Security.X509/PKCS12.cs
@@ -743,17 +743,17 @@ namespace Mono.Security.X509 {
 			byte[] privateKey = pki.PrivateKey;
 			try {
 				switch (pki.Algorithm) {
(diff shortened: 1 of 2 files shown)


filipnavara:
@monojenkins build failed

mcs/class/Mono.Security/Mono.Security.X509/PKCS12.cs
```diff
@@ -741,22 +741,26 @@ private DSAParameters GetExistingParameters (out bool found)
 		private void AddPrivateKey (PKCS8.PrivateKeyInfo pki) 
 		{
 			byte[] privateKey = pki.PrivateKey;
-			switch (privateKey [0]) {
-				case 0x02:
+			try {
+				switch (pki.Algorithm) {
+				case "1.2.840.113549.1.1.1": // Oids.RsaRsa
```
marek-safar: Would be better to use existing constants from https://github.com/mono/mono/blob/master/mcs/class/Mono.Security/Mono.Security.X509/X509Certificate.cs#L73 instead of comments (just make them internal)

PR #9170, opened 2018-06-15
[2018-06][loader] LoadFrom of problematic images should reprobe (correctly)

### LINKED ISSUE #8726: Loading denied assemblies via `Assembly.LoadFrom` can cause issues, eg. for VSMac

lambdageek:
This reverts commit a2bed67b6bef1bed79dc347cc8ed147096a07987
which itself reverted 9ad5255338386a22e7a861e8bcc7c195dc591b99

Additionally cherrypick 60f06e6ca26c28f874f67503cb58e7fb76095713: refonly load
of problematic image is allowed.

Fixes #8726

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -346,6 +346,10 @@ static MonoAssembly*
 mono_assembly_load_full_gac_base_default (MonoAssemblyName *aname, const char *basedir, MonoAssemblyContextKind asmctx, MonoImageOpenStatus *status);
 static MonoAssembly*
 mono_assembly_load_full_nodomain (MonoAssemblyName *aname, MonoAssemblyContextKind asmctx, MonoImageOpenStatus *status);
+static MonoAssembly*
+chain_redirections_loadfrom (MonoImage *image, MonoImageOpenStatus *status);
+static MonoAssembly*
+mono_problematic_image_reprobe (MonoImage *image, MonoImageOpenStatus *status);
 
 static MonoBoolean
 mono_assembly_is_in_gac (const gchar *filanem);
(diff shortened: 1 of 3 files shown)


lambdageek:
@monojenkins build pkg

lambdageek:
@monojenkins build failed

PR #9176, opened 2018-06-15
[runtime] Implement tombstone scrubbing for our concurrent hashtables.

kumpera:
Tombstones must be scrubbed as they can end up taking up all space of a hashtable and make lookup fail.

### DIFF
--- a/mono/metadata/mono-conc-hash.c
+++ b/mono/metadata/mono-conc-hash.c
@@ -32,7 +32,8 @@ struct _MonoConcGHashTable {
 	volatile conc_table *table; /* goes to HP0 */
 	GHashFunc hash_func;
 	GEqualFunc equal_func;
-	int element_count;
+	int element_count; //KVP + tombstones
+	int tombstone_count; //just tombstones
 	int overflow_count;
 	GDestroyNotify key_destroy_func;
 	GDestroyNotify value_destroy_func;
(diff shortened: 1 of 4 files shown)


alexanderkyte:
What happens if someone tries a lookup while the mutator threads are between ```set_key_to_tombstone``` and ```check_table_size``` in mono_conc_g_hash_table_remove? 

Won't they still get a hazard pointer to a hash table that lacks NULL or the current key? 

And then the table resize will happen on the mutator thread while the lookup code circles indefinitely in that while loop.

marek-safar:
@kumpera please backport down to 2018-02 or is the fix too risky?

kumpera:
@alexanderkyte search will work just fine because `mono_conc_g_hash_table_remove` doesn't increase the number of non-null elements in the hashtable.

The only function that increases the number of non-null elements is `mono_conc_g_hash_table_insert`, which checks before insert. This means that a removal right after the insert that sets `element_count` to `element_count` would trigger an expansion rehash.

This means that this patch reduces the effective load factor by 7% in the worst case (which the initial size of 16) but exponentially less as the table grows - it's .1% smaller at 32k size.

kumpera:
@marek-safar this part of the code base is really tricky, I'd wait at least a week in master before considering backporting.

kumpera:
failures look unrelated.

kumpera:
@monojenkins build failed

luhenry:
Is this a fix for a particular issue/bug? If not, I wouldn't want to backport it further than `2018-06` as it has a high chance of silently breaking things.

kumpera:
@luhenry it's a better fix for #9158
I spend some time today trying to write a test case

kumpera:
Disabled test under full-aot as it requires SRE.

PR #9178, opened 2018-06-15
[2018-04] Bump bockbuild to fix environment expansion bug

akoeplinger:
Brings in https://github.com/mono/bockbuild/pull/67.

This fixes the msbuild build problem where it'd crash with `error MSB4242: The SDK resolver "NuGetSdkResolver" failed to run. Unexpected failure reading NuGet.Config. Path: '@/external/bockbuild/builds/msbuild-15/artifacts/toolset/restore.proj/$HOME/.config/NuGet/NuGet.Config'.`.

The reason is that bockbuild exported the `XDG_CONFIG_HOME` environment variable with the value `$HOME/.config`. The problem is that the `$HOME` in there is never expanded so it just stays as the literal value. 

`XDG_CONFIG_HOME` is later used by e.g. `Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData)` which is what nuget uses to construct the path to NuGet.config - hence why we end up with the weird path.

We now just expand the variables in bockbuild before we pass them on.

Fixes https://github.com/mono/mono/issues/9179


akoeplinger:
@monojenkins build pkg

PR #9180, opened 2018-06-16
System.Diagnostics.Process improvements for AIX

NattyNarwhal:
AIX has a procfs with raw structures congruent with Solaris. For i,
which lacks procfs, fall back to a libc call.

This doesn't solve the Process weirdness that bungs up the test
suite, but I hope it's a start.

(I expect a lot of "fun" once CoreFX happens there...)


NattyNarwhal:
probable things to add: proclib cases to avoid even more usage of `/proc`

NattyNarwhal:
Should be ready to review. LMK if there's needed changes.

PR #9182, opened 2018-06-16
Reduce calls to mono_target_pagesize.

jaykrell:


jaykrell:

PR #9183, opened 2018-06-16
Make coop handles typesafe.

jaykrell:
Make coop handles typesafe.

Type-safe handles are a struct with a pointer to pointer.
The only operations allowed on them are the functions/macros in handle.h, and assignment
from same handle type to same handle type (which cannot be prohibited in C, but are ok anyway).

Type-unsafe handles are a pointer to a struct with a pointer.
Besides the type-safe operations, these can also be:
1. compared to NULL, instead of only MONO_HANDLE_IS_NULL
2. assigned from NULL, instead of only a handle
3. MONO_HANDLE_NEW (T) from anything, instead of only a T*
4. MONO_HANDLE_CAST from anything, instead of only another handle type
5. assigned from any void*, at least in C
6. Cast from any handle type to any handle type, without using MONO_HANDLE_CAST.
7. Cast from any handle type to any pointer type and vice versa, such as incorrect unboxing.
8. mono_object_class (handle), instead of mono_handle_class

None of those operations were likely intended and some are crashing sort of bugs,
esp. the last two, and possibly the use of NULL (NULL_HANDLE is not NULL, it is dereferencable,
yielding NULL).

There is no runtime cost.
NOTE: Running this code depends on the ABI to pass/return a struct
with a pointer the same as a pointer. This is tied in with
marshaling. If this is not the case, turn off type-safety, perhaps per-OS per-CPU. If it fails on all CPUs, make it a compile-only option. It is expected to actually work everywhere.

### DIFF
--- a/mono/eglib/goutput.c
+++ b/mono/eglib/goutput.c
@@ -187,7 +187,7 @@ to_android_priority (GLogLevelFlags log_level)
 #define LOG_MESSAGE_MAX_LEN 4096
 
 static void
-android_log_line (gint log_priority, const gchar *log_domain, gchar *log_message, gint log_len)
+android_log_line (gint log_priority, const gchar *log_domain, const gchar *log_message, gint log_len)
 {
 	gchar log_buf [LOG_MESSAGE_MAX_LEN];
 
@@ -204,7 +204,7 @@ static void
 android_log (gint log_priority, const gchar *log_domain, const gchar *log_message)
 {
 	gint log_message_len, log_message_p_len;
-	gchar *log_message_p;
+	const gchar *log_message_p;
 
 	log_message_len = strlen (log_message);
 	if (log_message_len <= LOG_MESSAGE_MAX_LEN) {
@@ -213,7 +213,7 @@ android_log (gint log_priority, const gchar *log_domain, const gchar *log_messag
 	}
 
 	for (log_message_p = log_message; log_message_p < log_message + log_message_len;) {
-		gchar *p = strstr (log_message_p, "\n");
+		const gchar *p = strstr (log_message_p, "\n");
 		if (p == NULL) {
 			/* There is no more "\n". */
 			android_log_line (log_priority, log_domain, log_message_p, LOG_MESSAGE_MAX_LEN - 1);
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -491,7 +491,7 @@ mono_domain_create_appdomain_checked (char *friendly_name, char *configuration_f
 	MonoDomain *result = NULL;
 
 	MonoClass *klass = mono_class_load_from_name (mono_defaults.corlib, "System", "AppDomainSetup");
-	MonoAppDomainSetupHandle setup = (MonoAppDomainSetupHandle)mono_object_new_handle (mono_domain_get (), klass, error);
+	MonoAppDomainSetupHandle setup = MONO_HANDLE_CAST (MonoAppDomainSetup, mono_object_new_handle (mono_domain_get (), klass, error));
 	goto_if_nok (error, leave);
 	MonoStringHandle config_file;
 	if (configuration_file != NULL) {
@@ -560,15 +560,15 @@ copy_app_domain_setup (MonoDomain *domain, MonoAppDomainSetupHandle setup, MonoE
 	caller_domain = mono_domain_get ();
 	ads_class = mono_class_load_from_name (mono_defaults.corlib, "System", "AppDomainSetup");
 
-	MonoAppDomainSetupHandle copy = (MonoAppDomainSetupHandle)mono_object_new_handle(domain, ads_class, error);
+	MonoAppDomainSetupHandle copy = MONO_HANDLE_CAST (MonoAppDomainSetup, mono_object_new_handle(domain, ads_class, error));
 	goto_if_nok (error, leave);
 
 	mono_domain_set_internal (domain);
 
-#define XCOPY_FIELD(dst,field,src,error)				\
+#define XCOPY_FIELD(type, dst, field, src, error)			\
 	do {								\
-		MonoObjectHandle src_val = MONO_HANDLE_NEW_GET (MonoObject, (src), field); \
-		MonoObjectHandle copied_val = mono_marshal_xdomain_copy_value_handle (src_val, error); \
+		TYPED_HANDLE_NAME (type) src_val = MONO_HANDLE_NEW_GET (type, (src), field); \
+		TYPED_HANDLE_NAME (type) copied_val = MONO_HANDLE_CAST (type, mono_marshal_xdomain_copy_value_handle (MONO_HANDLE_CAST (MonoObject, src_val), error)); \
 		goto_if_nok (error, leave);					\
 		MONO_HANDLE_SET ((dst),field,copied_val);		\
 	} while (0)
@@ -578,26 +578,26 @@ copy_app_domain_setup (MonoDomain *domain, MonoAppDomainSetupHandle setup, MonoE
 			MONO_HANDLE_SETVAL ((dst), field, type, MONO_HANDLE_GETVAL ((src),field)); \
 		} while (0)
 
-	XCOPY_FIELD (copy, application_base, setup, error);
-	XCOPY_FIELD (copy, application_name, setup, error);
-	XCOPY_FIELD (copy, cache_path, setup, error);
-	XCOPY_FIELD (copy, configuration_file, setup, error);
-	XCOPY_FIELD (copy, dynamic_base, setup, error);
-	XCOPY_FIELD (copy, license_file, setup, error);
-	XCOPY_FIELD (copy, private_bin_path, setup, error);
-	XCOPY_FIELD (copy, private_bin_path_probe, setup, error);
-	XCOPY_FIELD (copy, shadow_copy_directories, setup, error);
-	XCOPY_FIELD (copy, shadow_copy_files, setup, error);
+	XCOPY_FIELD (MonoString, copy, application_base, setup, error);
+	XCOPY_FIELD (MonoString, copy, application_name, setup, error);
+	XCOPY_FIELD (MonoString, copy, cache_path, setup, error);
+	XCOPY_FIELD (MonoString, copy, configuration_file, setup, error);
+	XCOPY_FIELD (MonoString, copy, dynamic_base, setup, error);
+	XCOPY_FIELD (MonoString, copy, license_file, setup, error);
+	XCOPY_FIELD (MonoString, copy, private_bin_path, setup, error);
+	XCOPY_FIELD (MonoString, copy, private_bin_path_probe, setup, error);
+	XCOPY_FIELD (MonoString, copy, shadow_copy_directories, setup, error);
+	XCOPY_FIELD (MonoString, copy, shadow_copy_files, setup, error);
 	COPY_VAL (copy, publisher_policy, MonoBoolean, setup);
 	COPY_VAL (copy, path_changed, MonoBoolean, setup);
 	COPY_VAL (copy, loader_optimization, int, setup);
 	COPY_VAL (copy, disallow_binding_redirects, MonoBoolean, setup);
 	COPY_VAL (copy, disallow_code_downloads, MonoBoolean, setup);
-	XCOPY_FIELD (copy, domain_initializer_args, setup, error);
+	XCOPY_FIELD (MonoArray, copy, domain_initializer_args, setup, error);
 	COPY_VAL (copy, disallow_appbase_probe, MonoBoolean, setup);
-	XCOPY_FIELD (copy, application_trust, setup, error);
-	XCOPY_FIELD (copy, configuration_bytes, setup, error);
-	XCOPY_FIELD (copy, serialized_non_primitives, setup, error);
+	XCOPY_FIELD (MonoObject, copy, application_trust, setup, error);
+	XCOPY_FIELD (MonoArray, copy, configuration_bytes, setup, error);
+	XCOPY_FIELD (MonoArray, copy, serialized_non_primitives, setup, error);
 
 #undef XCOPY_FIELD
 #undef COPY_VAL
@@ -624,7 +624,7 @@ mono_domain_create_appdomain_internal (char *friendly_name, MonoAppDomainSetupHa
 	/* FIXME: pin all those objects */
 	data = mono_domain_create();
 
-	MonoAppDomainHandle ad = (MonoAppDomainHandle)mono_object_new_handle (data, adclass, error);
+	MonoAppDomainHandle ad = MONO_HANDLE_CAST (MonoAppDomain, mono_object_new_handle (data, adclass, error));
 	goto_if_nok (error, leave);
 	MONO_HANDLE_SETVAL (ad, data, MonoDomain*, data);
 	data->domain = MONO_HANDLE_RAW (ad);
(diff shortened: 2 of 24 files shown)


jaykrell:
Windows gcc broken due to pointer type mismatches. For this PR don’t enable the warning/error for windows.

mono/metadata/handle.h
```diff
@@ -338,23 +406,81 @@ Handle macros/functions
...
+
+// Structs cannot be cast to structs.
+// Therefore, cast the function pointer to change its return type.
+#define MONO_HANDLE_CAST(type, value) \
+	(((TYPED_HANDLE_NAME (type) (*) (gpointer)) \
+		mono_remove_function_type_to_suppress_gcc_warning (mono_handle_cast)) ((value).__raw))
+
+// With Visual C++ compiling as C, the type of a ternary expression
+// yielding two unrelated non-void pointers is the type of the first, plus a warning.
+// This can be used to simulate gcc typeof extension.
```
lambdageek: Wow.

Do you think it's worth making this a more general utility somewhere in `utils/` and calling it here?  Might make the definition of `MONO_HANDLE_RAW` a bit cleaner.

mono/metadata/marshal.c
```diff
@@ -6223,19 +6223,29 @@ MonoObjectHandle
...
+#if MONO_TYPE_SAFE_HANDLES
+	// This is the only use of mono_handle_new_interior.
+	// It lacks the macro support of mono_handle_new.
+	//return *(MonoObjectHandle*)&rawh; // violates gcc -fstrict-aliasing
+	// Legal with gcc -fstrict-aliasing:
+	union { MonoRawHandle rawh; MonoObjectHandle objh; } type_pun = { rawh };
+	return type_pun.objh;
 #else
-	return mono_handle_new_interior (rawobj);
+	return rawh;
```
lambdageek: You know, I think `mono_icall_handle_new_interior` has the wrong return type.  It should probably be `MonoRawHandle` - `mono_handle_new_interior` doesn't really return a handle to a MonoObject.  It returns a handle to a pointer into the middle of some valuetype.

```
MonoRawHandle mono_icall_handle_new_interior (gpointer rawptr);
```

lambdageek:
Even if we find out from @vargaz that on all architectures that we care about `struct { gpointer *p} h;` and `struct { gpointer p; } *h` are passed the same way as args and return values, we should still keep at least one build going with `MONO_TYPE_SAFE_HANDLES` turned off.

mono/metadata/handle.h
```diff
@@ -24,6 +24,36 @@
...
+// None of those operations were likely intended.
+// Completely incidentally, the type-safe system is a little more expression-oriented than
+// the statement-based type-unsafe system, and might be valid C++ where type-unsafe is not (untested).
+
+// FIXME Do this only on checked builds? Or certain architectures?
+// There is not runtime cost.
+// NOTE: Running this code depends on the ABI to pass a struct
+// with a pointer the same as a pointer. This is tied in with
+// marshaling. If this is not the case, turn off type-safety, perhaps per-OS per-CPU.
+#if !defined (TARGET_X86) || defined (_WIN32) || defined (__MACH__)
```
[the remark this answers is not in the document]
luhenry: +1

mono/eglib/goutput.c
```diff
@@ -187,7 +187,7 @@ to_android_priority (GLogLevelFlags log_level)
 #define LOG_MESSAGE_MAX_LEN 4096
 
 static void
-android_log_line (gint log_priority, const gchar *log_domain, gchar *log_message, gint log_len)
+android_log_line (gint log_priority, const gchar *log_domain, const gchar *log_message, gint log_len)
```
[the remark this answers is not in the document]
luhenry: +1

mono/eglib/goutput.c
```diff
@@ -204,7 +204,7 @@ static void
 android_log (gint log_priority, const gchar *log_domain, const gchar *log_message)
 {
 	gint log_message_len, log_message_p_len;
-	gchar *log_message_p;
+	const gchar *log_message_p;
```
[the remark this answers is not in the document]
luhenry: +1

mono/eglib/goutput.c
```diff
@@ -213,7 +213,7 @@ android_log (gint log_priority, const gchar *log_domain, const gchar *log_messag
 	}
 
 	for (log_message_p = log_message; log_message_p < log_message + log_message_len;) {
-		gchar *p = strstr (log_message_p, "\n");
+		const gchar *p = strstr (log_message_p, "\n");
```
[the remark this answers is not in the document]
luhenry: +1

mono/metadata/appdomain.c
```diff
@@ -560,44 +560,42 @@ copy_app_domain_setup (MonoDomain *domain, MonoAppDomainSetupHandle setup, MonoE
...
-	XCOPY_FIELD (copy, application_name, setup, error);
-	XCOPY_FIELD (copy, cache_path, setup, error);
-	XCOPY_FIELD (copy, configuration_file, setup, error);
-	XCOPY_FIELD (copy, dynamic_base, setup, error);
-	XCOPY_FIELD (copy, license_file, setup, error);
-	XCOPY_FIELD (copy, private_bin_path, setup, error);
-	XCOPY_FIELD (copy, private_bin_path_probe, setup, error);
-	XCOPY_FIELD (copy, shadow_copy_directories, setup, error);
-	XCOPY_FIELD (copy, shadow_copy_files, setup, error);
+		MONO_HANDLE_SETVAL ((dst), field, type, MONO_HANDLE_GETVAL ((src), field))
```
luhenry: Unrelated change.

mono/metadata/handle.h
```diff
@@ -196,10 +224,22 @@ Icall macros
 	mono_stack_mark_record_size (__info, &__mark, __FUNCTION__);	\
 	mono_stack_mark_pop (__info, &__mark);
 
+#if MONO_TYPE_SAFE_HANDLES
+
+// Structs cannot be cast to structs.
+// Therefore, cast the function pointer to change its return type.
+#define CLEAR_ICALL_FRAME_VALUE(type, handle)				\
+	((mono_stack_mark_record_size (__info, &__mark, __FUNCTION__)),	\
```
luhenry: Use `do { } while (0)` instead of a LISP style abuse of paranthesis.

mono/metadata/handle.h
```diff
@@ -312,23 +363,40 @@ Handle macros/functions
  * For example, TYPED_HANDLE_DECL(MonoObject) (see below) expands to:
  *
  * typedef struct {
+#if MONO_TYPE_SAFE_HANDLES
```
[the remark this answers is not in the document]
luhenry: +1

mono/metadata/handle.h
```diff
@@ -312,23 +363,40 @@ Handle macros/functions
...
 	typedef TYPED_HANDLE_PAYLOAD_NAME (TYPE) * TYPED_OUT_HANDLE_NAME (TYPE)
+#endif
+
 /*
  * TYPED_VALUE_HANDLE_DECL(SomeType):
  *   Expands to a decl for handles to SomeType (which is a managed valuetype (likely a struct) of some sort) and to an internal payload struct.
  * For example TYPED_HANDLE_DECL(MonoMethodInfo) expands to:
  *
  * typedef struct {
+#if MONO_TYPE_SAFE_HANDLES
```
luhenry: Fix comment, add ` * ` at beginning of the line.

mono/eglib/goutput.c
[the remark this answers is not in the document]
jaykrell: The error to enforce handle-typesafety triggers here.

mono/eglib/goutput.c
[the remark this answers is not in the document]
jaykrell: The error to enforce handle-typesafety triggers here.

mono/eglib/goutput.c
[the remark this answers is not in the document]
jaykrell: The error to enforce handle-typesafety triggers here.

mono/metadata/handle.h
jaykrell: These have different meaning. Parens around expressions. do/while for statements. Use statements only where required. Prefer expressions for composability.
In at least one place I believe the composibility was required, i.e. because the type could not flow through a void*.

mono/metadata/handle.h
jaykrell: Sure.

mono/eglib/goutput.c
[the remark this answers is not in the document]
jaykrell: The error that enforced typesafety of handles also does not like this.

mono/eglib/goutput.c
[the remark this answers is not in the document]
jaykrell: The error that enforces typesafety of handles also does not like this.

mono/metadata/handle.h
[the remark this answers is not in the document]
jaykrell: We use these symbols too but ok.

mono/metadata/handle.h
jaykrell: I'm uneasy about this. I didn't find it in documentation. Documentation just says ternary does complicated type manipulation and discourages its  use. I was trying to think of how to use ternary to "preserve" a type in one place and apply it another and discovered it. I was hoping void* and T* would yield T*, but it yields void*. So I tried T1* and T2*. Turning off type-safety of handles when using Visual C++ is another choice. And C++ should allow doing this properly.

Or maybe it is wonderful??

Any on/off of type-safety will mean some builds succeed where others do not which is always unfortunate.

mono/metadata/marshal.c
jaykrell: You mean change MonoRawHandle to gpointer? Same thing but ok. Separate change?

mono/metadata/handle.h
jaykrell: Parens are not LISP style abuse. They are C style abuse. They are required to keep macro parameters separate from their surroundings. C++ fixes this with inline functions/templates.

Sometimes, alternatively, you can use temporaries, but this has drawbacks.

Once you have  temporaries, you lose the composibility of expressions, and you need to be able to state the types of the temporaries, which is sometimes difficult. There might be other problems. I never really thought deeply about it, or maybe long ago forgot details of the lesson, because everyone (except perhaps mono) just prefers expressions within macros, accepts that macros are poor, and therefore prefers neither.

jaykrell: In particular, CLEAR_ICALL_FRAME_VALUE needs to be an expression in the typesafe variant so 
 the type can transfer -- composibility. The type-unsafe variant does look off. Let me look again to see what changed.

mono/metadata/marshal.c
lambdageek: No I mean declare `MonoRawHandle mono_icall_handle_new_interior (gpointer rawobj);` and then implement it as

```c
MonoRawHandle
mono_icall_handle_new_interior (gpointer rawobj)
{
#ifdef MONO_HANDLE_TRACK_OWNER
    return mono_handle_new_interior (rawobj, "<marshal args>");
#else
    return mono_handle_new_interior (rawobj);
#endif
}
```

The only reason you have to jump through hoops here is because the icall is declared with a poor (incorrect) choice of return type.  So just change the return type.

mono/metadata/handle.h
jaykrell: I didn't change the type-unsafe one. It was already wierd, multiple-statements w/o  do/while:
This isn't generally anyone's preference but I left it alone. It is  used within a larger do/while and  is fairly private so ok.

```#define CLEAR_ICALL_FRAME_VALUE(RESULT, HANDLE)				\
	mono_stack_mark_record_size (__info, &__mark, __FUNCTION__);	\
	(RESULT) = mono_stack_mark_pop_value (__info, &__mark, (HANDLE));
```

mono/metadata/handle.h
jaykrell: This is required though to have MONO_HANDLE_RAW both support NULL_HANDLE, and not  double evaluate. Historically it null deferences on NULL_HANDLE.

I found this because I was removing the double null checks in  favor of existing MONO_HANDLE_IS_NULL -- which we could require gate all possibly-null MONO_HANDLE_RAWs.

mono/metadata/marshal.c
jaykrell: Got it. MonoRawHandle should probably be gpointer or gpointer* then.
gpointer* is more correct, gpointer is more opaque.

mono/metadata/appdomain.c
jaykrell: Ok.

mono/metadata/handle.h
```diff
@@ -24,6 +24,36 @@
...
+// None of those operations were likely intended.
+// Completely incidentally, the type-safe system is a little more expression-oriented than
+// the statement-based type-unsafe system, and might be valid C++ where type-unsafe is not (untested).
+
+// FIXME Do this only on checked builds? Or certain architectures?
+// There is not runtime cost.
+// NOTE: Running this code depends on the ABI to pass a struct
+// with a pointer the same as a pointer. This is tied in with
+// marshaling. If this is not the case, turn off type-safety, perhaps per-OS per-CPU.
+#if !defined (TARGET_X86) || defined (HOST_WIN32) || defined (HOST_DARWIN)
```
jaykrell: ABIs should be read up on to confirm the Linux divergence. Or look at small cod e samples. Or confirm this was the problem, etc.

mono/metadata/handle.h
jaykrell: Given the type parameter, it might be possible to keep it a statement. Though that would not seem preferable.

luhenry: In one case `CLEAR_ICALL_FRAME_VALUE` sets the `RESULT` value and in the other it returns it, this is using the same symbol to do 2 different things, which makes it harder to read. I understand in one case - the type safe case - we need to pass a type to actually be type safe, but changing from statement to expression is purely because of your preference for the latter. Nothing forces you to use a expression over a statement, and you could simply - as you also pointed out - use a statement with an added parameter. I will not block this PR over that but stop confusing your preferences over required differences, it's not the first time and it's becoming wearing.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

rjvdboon:
The change in `configure.ac` causes the `autoconf.sh`/`configure` command to fail when using gcc 4.x (tested 4.4.7 to 4.9.2 on RHEL 6.8).
```
checking for inet_aton... no
checking for main in -lpthread... no
checking pthread.h usability... no
checking pthread.h presence... yes
configure: WARNING: pthread.h: present but cannot be compiled
configure: WARNING: pthread.h:     check for missing prerequisite headers?
configure: WARNING: pthread.h: see the Autoconf documentation
configure: WARNING: pthread.h:     section "Present But Cannot Be Compiled"
configure: WARNING: pthread.h: proceeding with the preprocessor's result
configure: WARNING: pthread.h: in the future, the compiler will take precedence
configure: WARNING:     ## ------------------------------------------------------ ##
configure: WARNING:     ## Report this to https://github.com/mono/mono/issues/new ##
configure: WARNING:     ## ------------------------------------------------------ ##
checking for pthread.h... yes
checking pthread_np.h usability... no
checking pthread_np.h presence... no
checking for pthread_np.h... no
checking for pthread_mutex_timedlock... no
checking for pthread_getattr_np... no
checking for pthread_attr_get_np... no
checking for pthread_setname_np... no
checking for pthread_cond_timedwait_relative_np... no
checking for pthread_kill... no
checking for PTHREAD_MUTEX_RECURSIVE... no
configure: error: Posix system lacks support for recursive mutexes
```
Nearly all of the `... no` turn to `... yes` when I revert the change to `configure.ac`.

luhenry:
@jaykrell revert for now and fix later, or fix now?

jaykrell:
I think I can fix it shortly.  Obviously if we revert,  just revert configure.ac.

jaykrell:
https://github.com/mono/mono/pull/9251

Tested on Centos6/gcc 4.4.6, and MacHighSierra. Verified config.log success/failure looking for  "incompatible".

PR #9188, opened 2018-06-17
Fix handle type violations and null deref checking for null.

jaykrell:


jaykrell:
Redundant with https://github.com/mono/mono/commit/475a69f7c56a3827a4eafe9fed4151b6feed05a4, closing.

Issue #7249, opened 2018-02-23
Fast Span<T>


marek-safar:
We bundle portable (slow) version of Span<T> which is using [`Pinnable<T>`](https://github.com/dotnet/corefx/blob/1885cf8e6590fc7c4f48ef22a368beeb4fc1ee59/src/System.Memory/src/System/Pinnable.cs) type to do all the memory operations which is not as efficient as it could be.

We'd like to switch to CoreClr [fast](https://github.com/dotnet/coreclr/blob/6b27fa740ad44962093c5af4e6060011ec454a46/src/mscorlib/shared/System/Span.Fast.cs) implementation  which is using [`ByReference<T>`](https://github.com/dotnet/coreclr/blob/6b27fa740ad44962093c5af4e6060011ec454a46/src/mscorlib/src/System/ByReference.cs) as JIT intristic around ref T fields.

Same applies to `ReadOnlySpan<T>`.

Therzok:
Would this mean that the NuGet version of Span would be switched to the BCL (Fast) supplied one? Or would it still be available only in modern API targets?

Therzok:
Ah, redundant question, I'll just check if the type is TypeForwarded

marek-safar:
@vargaz please look at the pending PR, it has the managed pieces wired-up but needs runtime icalls or a different way of implementing them. The build fails due to that right now.

PR #9204, opened 2018-06-18
[corlib] Default to fast Span<T>

marek-safar:
Part of #7249


vargaz:
This needs a rebase.

vargaz:
Seems to be missing Buffer:Memmove<T>, also the definition of the non generic version of memmove doesn't seem right:

                internal static unsafe void Memmove (byte *dest, byte *src, uint len)
                {
                        Memcpy (dest, src, (int) len);
                }

                internal static void Memmove<T>(ref T destination, ref T source, nuint elementCount)
                {
                        throw new NotImplementedException ("Buffer::Memmove<T>");
                }

marek-safar:
@vargaz correct, not sure if we should do ical or try to adopt quite ugly managed implementation from coreclr

vargaz:
The build doesn't define BIT64, and some code like Span.Clear () depends on it being the correct value:

               SpanHelpers.ClearWithReferences(ref Unsafe.As<T, IntPtr>(ref _pointer.Value), (nuint)_length * (nuint)(Unsafe.SizeOf<T>() / sizeof(nuint)));

vargaz:
The build seems to work now but there are a lot of class lib failures.

mcs/class/System/System.Net/HttpWebRequest.cs
```diff
@@ -828,7 +828,7 @@ void AddRange (string rangeSpecifier, long from, long to)
 		WebOperation SendRequest (bool redirecting, BufferOffsetSize writeBuffer, CancellationToken cancellationToken)
 		{
 			lock (locker) {
-				WebConnection.Debug ($"HWR SEND REQUEST: Req={ID} requestSent={requestSent} redirecting={redirecting}");
+				WebConnection.Debug ($"HWR SEND REQUEST: Req={ID} requestSent={requestSent} actualUri={actualUri} redirecting={redirecting}");
```
kg: Minor issue but won't $"..." generate garbage and burn cycles string formatting even if debug messages are suppressed? Wouldn't it be better to do "... {0} ...", x style instead

akoeplinger: In general yes but `WebConnection.Debug()` is marked as `[Conditional ("MONO_WEB_DEBUG")]` so the compiler will not even emit it into the output assembly.

vargaz:
The 32 bit failures are (mostly) caused by the BIT64 define when compiling corlib, some coreclr code depends on a correct type for nuint:

       public void Clear()
        {
            if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
            {
                SpanHelpers.ClearWithReferences(ref Unsafe.As<T, IntPtr>(ref _pointer.Value), (nuint)_length * (nuint)(Unsafe.SizeOf<T>() / sizeof(nuint)));
            }
            else
            {
                SpanHelpers.ClearWithoutReferences(ref Unsafe.As<T, byte>(ref _pointer.Value), (nuint)_length * (nuint)Unsafe.SizeOf<T>());
            }
        }

marek-safar:

marek-safar:
@monojenkins backport 2018-06

PR #9208, opened 2018-06-18
[aot modes] rename FullAOT + interp and add interp-only mode

lewurm:
usually we don't rename constants that are available via public API, but in this case `MONO_AOT_MODE_INTERP` wasn't available that long yet, so I argue we can still do it 😉


kumpera:
This PR missed the man page.

lewurm:
see https://github.com/mono/mono/pull/9220

PR #9212, opened 2018-06-18
[System]: `HttpWebRequest` now throws `WebExceptionStatus.RequestCanceled` on abort.  #9031.

baulig:
Cleanup the internal `HttpWebRequest.RunWithTimeout()` function and use
`GetWebException()` everywhere to throw `WebException` with
`WebExceptionStatus.RequestCanceled` on abort.

New tests for this will be added to the web-tests by extending the existing
`Xamarin.WebTests.HttpInstrumentationTests.AbortDuringHandshake` to check the
synchronous and begin/end async APIs as well, using both `GET` and `POST`.

See https://github.com/xamarin/web-tests/blob/master/Xamarin.WebTests.Tests/Xamarin.WebTests.HttpInstrumentationTests/AbortDuringHandshake.cs.

Fixes issue #9031.

### DIFF
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -434,7 +434,6 @@ namespace System.Net
 		}
 
 		public string Host {
-
 			get {
 				Uri uri = hostUri ?? Address;
 				return (hostUri == null || !hostHasPort) && Address.IsDefaultPort ?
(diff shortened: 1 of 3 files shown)


marek-safar:
@baulig this breaks existing tests

baulig:
The expected behavior in `HttpClientTest.Send_Transfer_Encoding_Custom` is wrong (matching the currently broken behavior in `HttpWebRequest`).

I'm going to fix both `HttpWebRequest` and that test to correctly throw `InvalidOperationException` instead of `ProtocolException`.

baulig:
Should be all fixed now.  See also the newly added test https://github.com/xamarin/web-tests/commit/20fc54fb3677c5a4f635b6164d2c22d1de61bcd8.

mcs/class/System/System.Net/HttpWebRequest.cs
```diff
@@ -1184,6 +1167,20 @@ public override IAsyncResult BeginGetResponse (AsyncCallback callback, object st
 			if (Aborted)
 				throw CreateRequestAbortedException ();
 
+			string transferEncoding = TransferEncoding;
+			if (!sendChunked && transferEncoding != null && transferEncoding.Trim () != "") {
+				/*
```
baulig: This section here https://github.com/baulig/mono/blob/02c139772082bd708b510c0732bda27bc7bc936b/mcs/class/System/System.Net/HttpWebRequest.cs#L660 already matches the CoreFX behavior.

We can also get rid of the `Trim()` calls once we have switched `HttpClient` to the CoreFX version and also use the header code from CoreFX.

mcs/class/System/System.Net/HttpWebRequest.cs
```diff
@@ -967,29 +948,26 @@ public override Task<Stream> GetRequestStreamAsync ()
...
+			var workerTask = func (cts.Token);
+			return RunWithTimeoutWorker (workerTask, timeout, Abort, () => Aborted, cts);
 		}
 
 		async Task<HttpWebResponse> MyGetResponseAsync (CancellationToken cancellationToken)
 		{
 			if (Aborted)
 				throw CreateRequestAbortedException ();
 
-			if (method == null)
```
baulig: This is impossible, we assign a default value to the field and have null checks in all accessors.

baulig:
You can run that newly added test on Windows by doing

```
$ msbuild Xamarin.WebTests.DotNet.sln
$ .\DotNet\Xamarin.WebTests.DotNet\bin\Debug\Xamarin.WebTests.DotNet.exe --features=+RecentlyFixed --select=HttpClientInstrumentation.CustomTransferEncoding
```

akoeplinger:
@baulig please make a manual backport PR for 2018-04.

PR #9214, opened 2018-06-19
[mscorlib] Conditional string ctors dependency

marek-safar:


marek-safar:
@monojenkins commit csproj

PR #9219, opened 2018-06-19
Refine typesafe coop handles to not cast function pointers.

jaykrell:


jaykrell:
@monojenkins build failed

mono/metadata/handle.h
```diff
@@ -432,7 +432,7 @@ typedef struct _MonoTypeofCastHelper *MonoTypeofCastHelper; // a pointer type un
 #ifndef MONO_HANDLE_TRACK_OWNER
 
 #define MONO_HANDLE_NEW(type, object) \
```
jaykrell: There are missing parens elsewhere i.e. non-typesafe handles, but seemingly we prefer it that way???

jaykrell:

PR #9220, opened 2018-06-19
[interp] add different API to enable interpreter in embedding scenario

lewurm:
Squezzing the interpreter into `mono_jit_set_aot_mode` was hackish and
doesn't scale. `mono_runtime_use_interpreter` is orthogonal to
`mono_jit_set_aot_mode`. For example, if we want to add HybridAOT +
interpreter we don't need to add another AOT mode with this approach.

At the moment calling those two functions requires a certain order,
which we will get rid of in the future.

### DIFF
--- a/mono/mini/driver.c
+++ b/mono/mini/driver.c
@@ -1759,7 +1759,7 @@ apply_root_domain_configuration_file_bindings (MonoDomain *domain, char *root_do
 static void
 mono_enable_interp (const char *opts)
 {
-	mono_runtime_set_execution_mode (MONO_EE_MODE_INTERP);
+	mono_runtime_use_interpreter (TRUE);
 	if (opts)
 		mono_interp_opts_string = opts;
 
@@ -1988,7 +1988,8 @@ mono_main (int argc, char* argv[])
 		} else if (strcmp (argv [i], "--hybrid-aot") == 0) {
 			mono_jit_set_aot_mode (MONO_AOT_MODE_HYBRID);
 		} else if (strcmp (argv [i], "--full-aot-interp") == 0) {
-			mono_jit_set_aot_mode (MONO_AOT_MODE_INTERP);
+			mono_jit_set_aot_mode (MONO_AOT_MODE_FULL);
+			mono_runtime_use_interpreter (TRUE);
 		} else if (strcmp (argv [i], "--print-vtable") == 0) {
 			mono_print_vtable = TRUE;
 		} else if (strcmp (argv [i], "--stats") == 0) {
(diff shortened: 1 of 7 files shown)


mono/mini/driver.c
```diff
@@ -2578,14 +2579,14 @@ mono_jit_set_aot_only (gboolean val)
...
-	memset (&mono_ee_features, 0, sizeof (mono_ee_features));
-
 	switch (mode) {
 	case MONO_AOT_MODE_LLVMONLY:
 		mono_aot_only = TRUE;
 		mono_llvm_only = TRUE;
 
+#ifndef TARGET_WASM
 		mono_ee_features.use_aot_trampolines = TRUE;
+#endif
```
lewurm: what do you think about that @kumpera ? would you prefer to have another AOT mode so we can get rid of the `ifdef`?

kumpera:
I don't like this, I think a set of execution modes instead of a bucket of options much more manageable for us.
If `mono_jit_set_aot_mode` is problematic for the interp. Deprecate it (and revert your recent breaking changes to it) and introduce a `mono_runtime_set_execution_mode` that subsumes it all.

BTW, if we go with this. I suggest we clearly document that the 2 APIs are runtime incompatible that mixing them would assert. IE:

```
static int mode_api_used;

void mono_runtime_set_execution_mode(int mode) {
if (mode_api_used == 2) g_error ("can't mix api modes");
mode_api_used = 1;
}

void mono_jit_set_aot_mode (int mode) {
if (mode_api_used == 1) g_error ("can't mix api modes");
mode_api_used = 2;
}
```

mono/mini/driver.c
```diff
@@ -1988,7 +1988,8 @@ mono_main (int argc, char* argv[])
 		} else if (strcmp (argv [i], "--hybrid-aot") == 0) {
 			mono_jit_set_aot_mode (MONO_AOT_MODE_HYBRID);
 		} else if (strcmp (argv [i], "--full-aot-interp") == 0) {
```
luhenry: I would love to take this occasion to remove this command line parameter which seems like a hack to me while it's equivalent to `--full-aot --interpreter`, but that's not required for this PR.

luhenry:
@kumpera the reasoning behind having 2 APIs is the following: you want to be able to mix and match any AOT mode with the dynamic EE you want (JIT or Interpreter). Interpreter only makes sense on platforms where the JIT is not available (since it's faster and supports more features), so having to enable the Interpreter on demand makes more sense for now. We will want the same approach for the command lines where using the JIT or the Interpreter will be based on passing `--interp`, and that will be separate from which AOT mode we run with (`--hybrid-aot`, `--full-aot`, `-O=-aot`). A `mono_runtime_set_execution_mode` where you pass `MONO_AOT_MODE_... | MONO_EE_MODE_INTER_...` would fix the issue at hand as well and would need to be a superset of `mono_jit_set_aot_mode`.

kumpera:
I think the salad bar approach of execution configuration the way to lots of non working configurations and bugs.

PR #9221, opened 2018-06-19
Enable incompatible-pointer warnings on Windows/gcc like the other gcc.

jaykrell:
And suppress new warnings from typesafe coop handles.
And suppress other warnings.

### DIFF
--- a/mono/mini/mini-amd64.h
+++ b/mono/mini/mini-amd64.h
@@ -203,7 +203,7 @@ typedef struct MonoCompileArch {
 	gint32 async_point_count;
 	gpointer vret_addr_loc;
 #ifdef HOST_WIN32
-	gpointer unwindinfo;
+	struct _UNWIND_INFO* unwindinfo;
 #endif
 	gpointer seq_point_info_var;
 	gpointer ss_trigger_page_var;
@@ -584,8 +584,10 @@ mono_arch_unwindinfo_free_unwind_info (PUNWIND_INFO unwind_info);
 guint
 mono_arch_unwindinfo_init_method_unwind_info (gpointer cfg);
 
+#ifdef MONO_ARCH_HAVE_UNWIND_TABLE
 void
-mono_arch_unwindinfo_install_method_unwind_info (gpointer *monoui, gpointer code, guint code_size);
+mono_arch_unwindinfo_install_method_unwind_info (PUNWIND_INFO *monoui, gpointer code, guint code_size);
+#endif
 
 void
 mono_arch_unwindinfo_install_tramp_unwind_info (GSList *unwind_ops, gpointer code, guint code_size);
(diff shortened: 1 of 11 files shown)


jaykrell:

winconfig.h
```diff
@@ -713,6 +713,7 @@
 #pragma warning(disable:4273) // inconsistent dll linkage
 #pragma warning(disable:4293) // shift count negative or too big, undefined behavior
 #pragma warning(disable:4305) // truncation from 'double' to 'float'
+#pragma warning(disable:4312: // 'type cast': conversion from 'MonoNativeThreadId' to 'gpointer' of greater size
```
akoeplinger: the trailing `:` looks suspicious :)

jaykrell: oops copy/paste

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
review?

mono/mini/mini-amd64.h
```diff
@@ -584,8 +584,10 @@ mono_arch_unwindinfo_free_unwind_info (PUNWIND_INFO unwind_info);
 guint
 mono_arch_unwindinfo_init_method_unwind_info (gpointer cfg);
 
+#ifdef MONO_ARCH_HAVE_UNWIND_TABLE
```
lateralusX: Is this #ifdef needed? The define is set above and all of the methods in this block are enabled so nothing special about this method.

jaykrell: Maybe not.

jaykrell: LPDWORD is a leftover from 16bit systems with "long" pointers and has no place in today's code.

jaykrell: I think you are right. Let me check and maybe remove it. You just can't use that type in portable code.

jaykrell: You also use it right below.

jaykrell: I see. "Right below" after the endifs, in which it is defined.

jaykrell: I was also surprised you typedef UNWIND_INFO and UNWIND_CODE yourself vs. from windows.h. windows.h does have a big problem, granted -- it doesn't allow for cross-compilation or cross-dumping (link /dump). The types in windows.h should be AMD64_UNWIND_INFO and then conditionally typedefed to "native" UNWIND_INFO. Many folks are forced to duplicate them because of this, alas, i.e. all cross-compilers and cross-linkers, i.e. all compilers and linkers, except for the occasional JIT-only.

jaykrell: Lately people are even advising against the "P" and favoring DWORD* over PDWORD. The "P" is also kind of a holdover, from NT style and I suspect it dates to more verbose language like Pascal or Modula or PL/1 where the "P" was really a tremendous shortening over like REF DWORD or DWORD POINTER; TYPE PDWORD = REF DWORD or such. But in C it is just trading one character for one character, two tokens for one token.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
Btw, the pointer can be avoided entirely, via copy in/out of the value, or usually just copy out, being mindful of null.

PR #9230, opened 2018-06-20
Add LLVM AOT/Full AOT support on Windows x64.

lateralusX:
Uses mono LLVM master branch, build using cmake and VisualStudio 2015/2017. Commit adds support for AOT/Full AOT LLVM codegen for Windows x64 Visual Studio build mono runtime. Normal configure/make using cygwin won’t support LLVM and will trigger an error if configured.

In order to build using LLVM first mono LLVM master branch must be build using static linked LLVM configuration (dynamic loading can be added later), NOTE mono LLVM build needs to include the following commit, https://github.com/mono/llvm/commit/9b92b4b87607e137266f84dc307181b8842fe54a in order to successfully build on Windows x64.

Build mono LLVM branch using the following cmake configuration:

-DCMAKE_BUILD_TYPE=Release
-DLLVM_ENABLE_ZLIB=OFF
-DLLVM_TARGETS_TO_BUILD="X86"
-DCMAKE_CROSSCOMPILING=False
-DCMAKE_SYSTEM_NAME=Windows

Use cmake with -G "Visual Studio 14 2015 Win64" to generate VS 2015 x64 targets. NOTE, if mono runtime is build using VS 2017, then LLVM should be build using the same VS version using -G "Visual Studio 15 2017 Win64"

To enable LLVM build in mono.sln, set MONO_ENABLE_LLVM to “true” and make sure MONO_LLVM_INSTALL_DIR_PREFIX points to the install directory used in LLVM build above, default msvc/dist/llvm. Above LLVM properties can also be set if build using msbuild.

A sample LLVM configure script has been included in msvc/llvm-cmake-config.bat.

### DIFF
--- a/mono/mini/llvm-jit.h
+++ b/mono/mini/llvm-jit.h
@@ -16,7 +16,9 @@
 #include "llvm-c/Core.h"
 #include "llvm-c/ExecutionEngine.h"
 
+#ifdef HAVE_UNWIND_H
 #include <unwind.h>
+#endif
 
 /* These can't go into mini-<ARCH>.h since thats not included into llvm-jit.cpp */
 #if defined(TARGET_AMD64) && defined(TARGET_OSX)
--- a/configure.ac
+++ b/configure.ac
@@ -3736,11 +3736,11 @@ if test "x$enable_llvm" = "xdefault"; then
 fi
 
 if test "x$enable_llvm" = "xyes"; then
+   if test x$host_win32 = xyes; then
+         AC_MSG_ERROR([LLVM for Windows only supported using Visual Studio build mono runtime, see http://www.mono-project.com/docs/advanced/mono-llvm/ for more details.])
+   fi
    if test "x$with_llvm" != "x"; then
    	  LLVM_CONFIG=$with_llvm/bin/llvm-config
-   	  if test x$host_win32 = xyes; then
-   	  	LLVM_CONFIG=$LLVM_CONFIG.exe
-   	  fi
 	  if test ! -x $LLVM_CONFIG; then
 	  	 AC_MSG_ERROR([LLVM executable $LLVM_CONFIG not found.])
       fi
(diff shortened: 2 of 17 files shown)


luhenry:
@lateralusX when you say "mono LLVM master branch", do you mean `mono/llvm:master`? We switched to `mono/llvm:release_60` for Unix platforms (at least on `mono:master`), so it would be useful - even if not required - for windows and other platforms to track the same LLVM.

lateralusX:
@luhenry Yes, its currently mono/llvm:master. I was pointed to start using that branch by @vargaz in order to get things working. Maybe I could upgrade to use later LLVM branch ones things has been merged? I expect most of the mono specific changes to work as is, just need to port the fixes I currently have in mono/llvm:master to never LLVM branch and rebuild/retest to see if we have the same pass rate on later LLVM branches as well.

luhenry:
@lateralusX we already do depend on `mono/llvm:release_60` in `mono:master`, but again tracking the same exact LLVM version between Windows and other platforms is not _required_ even if _much easier_ to maintain (if we test our LLVM integration on non-Windows, we would have a much lower confidence in not breaking Windows). 

For some context, what's the current existing status of LLVM on Windows? Non-existent and not tested, or still a bit better?

lateralusX:
@luhenry, before this PR, non-existent. This PR enables LLVM on Windows x64 since that is the only target we currently support full AOT.

lateralusX:
Looks like the stability issue on "Linux x64 Hybrid Suspend" affects many different PR's and doesn't look related to changes done in this PR.

lateralusX:
@vargaz OK with changes done in this PR?

mono/mini/llvm-jit.h
```diff
@@ -16,7 +16,9 @@
 #include "llvm-c/Core.h"
 #include "llvm-c/ExecutionEngine.h"
 
+#ifdef HAVE_UNWIND_H
 #include <unwind.h>
+#endif
```
vargaz: Shouldn't this be in a .c or .cpp file ?

directhex:
For my own reference: `MONO_LLVM_INSTALL_DIR_PREFIX` should be e.g. `c:\sources\mono\mcvc\Debug\` and the trailing slash is important

directhex:
Hm, or not. I'm clearly missing something.

directhex:
aha! missing step is building INSTALL.vcxproj, which installs into dist/llvm, then pass dist/llvm, as MONO_LLVM_INSTALL_DIR_PREFIX

directhex:
@lateralusX 

1) This requires manual setting of --aot=llvm-path=c:\foo\ right? it doesn't seem to use llc/opt out of the same directory as mono-sgen.exe if present

2) This is using `clang.exe` to compile the final assembly (`#define AS_NAME "clang.exe"`). Where is that clang.exe supposed to come from? We aren't building one. The experimental clang/c2 compiler from MSVC2017?

lateralusX:
@directhex 

1. By default mono just runs llc/opt so it need to be located either in current WD, PATH or by setting llvm-path.

2. Yes, this is the same for normal full AOT on Windows. We use clang/c2 from the VS install, either 2015 or 2017 version. Clang also needs to be available in the PATH of the command prompt running the command since mono just runs clang.exe. I believe you should be able to use toolprefix as well on command line ti specify the path where clang.exe is located.

Since we use the VS2015 or VS2017 toolchain as part of the full AOT process the best way is to run the command under a corresponding VS command prompt since that will make right version of link.exe and potential needed c-runtime and Win SDK visible for the toolchain. NOTE, clang.exe is currently not setup as VS normal build prompts, so step 2 is still needed.

I have a task to add build instructions to the official documentation, but didn't have the time to do it yet and there are a couple of smaller LLVM tasks left to do as well. I have added a couple of helper scripts used when running mini regression tests on Windows using JIT as well as Full AOT + LLVM. If you look inside the scripts (especially the ones setting up one of the toolchains, clang-vs2015-toolchain.bat or clang-vs2017-toolchain.bat). There is also some info in the PR itself, https://github.com/mono/mono/pull/9252.

Issue #7589, opened 2018-03-13
CoreFX System.Security


marek-safar:
One of few assemblies where we didn't even migrated to referencesource. We are missing some functionality and our code is not very linker friendly. There are also reported security vulnerabilities we didn't act on.

marek-safar:
This one will require multiple steps. It's better to start with small types first before doing any large ones.

filipnavara:
I think System.Security.Cryptography.Pkcs should be checked off, X509Certificates is now tracked in issue #9216.

marek-safar:
@filipnavara System.Security assembly has additional X509Certificates types which I track here (similarly for mscorlib)

filipnavara:
@marek-safar My bad, I missed that. However, CoreFX doesn't implement these types at the moment. They are Windows-specific and all functions referencing them were wiped-out with `throw new PlatformNotSupportedException()`.

I once tried to implement `X509Certificate2UI.DisplayCertificate` overloads, but eventually I abandoned the effort. On macOS the underlying APIs require the application to be properly initialized Cocoa application. Moreover the window handle is dependent on the System.Windows.Forms backend (at least for the common usage), which currently lacks the Cocoa implementation altogether. Essentially the API is non-portable unless you constrain it with specific technologies. The only sensible implementation is by spawning a separate process that will implement the UI. Not sure if that's worth the effort, but I can contribute code for that if desired...

marek-safar:
I agree the *UI API is just poor Windows API, we will keep the minimalistic implementation we have for now.

PR #9236, opened 2018-06-20
CoreFX import for Pkcs9SigningTime type

MaximLipnin:
as part of #7589


marek-safar:
@MaximLipnin could you comment on the disabled test or include the test failures

MaximLipnin:
Existing implementation has [explicit checks](https://github.com/mono/mono/blob/master/mcs/class/System.Security/System.Security.Cryptography.Pkcs/Pkcs9SigningTime.cs#L101-L105) for valid period. Ignored tests verify this logic. Running these tests against CoreFX code of Pkcs9SigningTime  (which doesn't have period checks) on MacOS leads to failures because some expected exception are not thrown.
Net framework and CoreFX  on windows throw exceptions for these edge cases.
CoreFX on MacOS doesn't throw any exception for these edge cases.

If it's still actual then possibly there is some issue in supporting validity period for non-windows platform.

MaximLipnin:
https://github.com/dotnet/corefx/issues/30578

filipnavara:
The explict checks should be there for the reasons mentioned in dotnet/corefx#30578. The only question is whether they should be done by Pkcs9SigningTime or AsnWriter. I lean towards the later, but it's ultimately up to the CoreFX guys to decide.

MaximLipnin:
@marek-safar Could you please cherry-pick https://github.com/dotnet/corefx/pull/30844 ?

marek-safar:
@MaximLipnin do we still need Ignore tests change?

filipnavara:
The test should pass now, but the time-specific tests are now in CoreFX too, so they should probably be removed from Mono. Eventually CoreFX may start supporting the GENERAL-TIME representation (and thus years outside of the 1950-2049 range) and the test there would reflect that.

marek-safar:
@monojenkins commit csproj

MaximLipnin:
Thanks for the review!

PR #9237, opened 2018-06-20
[runtime] Move accessors for hidden-yet-public debugger agent types into debugger-agent.h

alexanderkyte:
The code implementations of these functions live in debugger-agent.c.

Even though it makes the header seem more like it's a machine-generated way to wrap the
datatypes than the meaning-carrying MONO_API header, this will
allow for code which calls these functions to be moved between .c files
without having to change the header files.

### DIFF
--- a/mono/mini/debugger-agent.h
+++ b/mono/mini/debugger-agent.h
@@ -31,6 +31,22 @@ struct _MonoDebuggerCallbacks {
 };
 
 typedef struct _DebuggerTlsData DebuggerTlsData;
+typedef enum _MonoDebuggerThreadState MonoDebuggerThreadState;
+
+MonoGHashTable *
+mono_debugger_get_thread_states (void);
+
+gboolean
+mono_debugger_is_disconnected (void);
+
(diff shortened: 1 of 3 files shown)


alexanderkyte:
An alternative change would be to move DebuggerTls into debugger-engine or another type. This would be more future-proof.

alexanderkyte:
@kumpera

mono/mini/debugger-agent.h
```diff
@@ -31,6 +31,22 @@ struct _MonoDebuggerCallbacks {
 };
 
 typedef struct _DebuggerTlsData DebuggerTlsData;
+typedef enum _MonoDebuggerThreadState MonoDebuggerThreadState;
```
luhenry: You could just declare the enum here.

alexanderkyte: I can do that, but @kumpera in channel sounded like he wanted to make these changes another way. If this is still destined for merge, I'm happy to make that change.

PR #9238, opened 2018-06-20
[mini] Implement delegate tramp info using gsharing context.

kumpera:


mono/mini/patch-info.h
```diff
@@ -65,3 +65,4 @@ PATCH_INFO(SET_TLS_TRAMP, "set_tls_tramp")
 PATCH_INFO(JIT_ICALL_ADDR_NOCALL, "jit_icall_addr_nocall")
 PATCH_INFO(PROFILER_ALLOCATION_COUNT, "profiler_allocation_count")
 PATCH_INFO(PROFILER_CLAUSE_COUNT, "profiler_clause_count")
+PATCH_INFO(DELE_TRAMP, "delegate_trampoline_info")
```
vargaz: Missing newline. Also, it could be called DEL_TRAMP or DELEGATE_TRAMP, DELE is not used as an abbrev currently.

vargaz: Also, the aot file format version should be bumped in aot-runtime.h.

kumpera: I think I need to update AOT patch serialization for this.

vargaz: Yes

PR #9239, opened 2018-06-20
Make gensources.cs fail on missing files or zero-match wildcards

kg:
The gensources.cs currently checked in to master doesn't fail on missing files by default because it didn't work at the time. This turns that behavior on. It is currently only in my other in-progress branch so I'm moving it here.


kg:
The interp and OS X x64 failures look like flakes. Assuming the other checks pass, is it OK to merge this?

PR #9246, opened 2018-06-20
[w32socket] Remove useless interrupt callback

luhenry:
Thread A calls `Accept()`. Thread B wants to abort Thread A:
- Thread A installs the interrupt callback `abort_syscall`
- Thread B begins an async abort of Thread A, and ends up in `async_abort_internal`
  - It will call `mono_thread_info_safe_suspend_and_run` and after that succeeds, it will get back the interrupt token that Thread A installed, and then Thread B will call `abort_syscall` on Thread A.
  - but before it does that, Thread A first ends up in `suspend_sync` which will end up trying to suspend the thread and since `interrupt_kernel` is `TRUE`, it will even break syscalls.
  - So by the time Thread B returns to `async_abort_internal` and calls the callback,  Thread A has already started being aborted, so we would just abort it for a second time for no reason.

### DIFF
--- a/mono/metadata/w32socket.c
+++ b/mono/metadata/w32socket.c
@@ -215,12 +215,6 @@ mono_w32socket_close (SOCKET sock)
 
 #endif /* HOST_WIN32 */
 
-static void
-abort_syscall (gpointer data)
-{
-	mono_thread_info_abort_socket_syscall_for_close ((MonoNativeThreadId) (gsize) data);
-}
-
 static gint32
 convert_family (MonoAddressFamily mono_family)
 {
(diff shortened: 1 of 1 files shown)


lambdageek:
@luhenry I don't entirely understand what the problem is.  Is it something like this:

Thread A calls `Accept()`. Thread B wants to abort Thread A:
- Thread A installs the interrupt callback `abort_syscall`
- Thread B begins an async abort of Thread A, and ends up in `async_abort_internal`
  - It will call `mono_thread_info_safe_suspend_and_run` and after that succeeds, it will get back the interrupt token that Thread A installed, and then Thread B will call `abort_syscall` on Thread A.
  - but before it does that, Thread A first ends up in `suspend_sync` which will end up trying to suspend the thread and since `interrupt_kernel` is `TRUE`, it will even break syscalls.
  - So by the time Thread B returns to `async_abort_internal` and calls the callback,  Thread A has already started being aborted, so we would just abort it for a second time for no reason.

Is that right?

luhenry:
@lambdageek exactly! Do you mind idea I use your explanation in my commit message? :D

mono/metadata/w32socket.c
```diff
@@ -2812,15 +2631,52 @@ mono_network_cleanup (void)
...
+ * break a given thread from socket syscalls.
+ *
+ * This only exists because linux won't fail a call to connect
+ * if the underlying is closed.
+ *
+ * TODO We should cleanup and unify this with the other syscall abort
+ * facility.
+ */
 void
 ves_icall_cancel_blocking_socket_operation (MonoThreadObjectHandle thread, MonoError *error)
```
jaykrell: I imagine mono_thread_info_abort_socket_syscall_for_close was in threads.c because it is duplicating so much of its gnarly internals. Perhaps that should not be changed?

But then, more so, can this instead be more like:

```
static SuspendThreadResult
mono_thread_abort_syscall_callback (THREAD_INFO_TYPE *info, gpointer user_data)
{
	return MonoResumeThread;
}

void	
mono_thread_abort_syscall (MonoNativeThreadId tid)	
{
	if (tid == mono_native_thread_id_get ())
		return;
	mono_thread_info_safe_suspend_and_run (id, TRUE/*interrupt kernel*/, mono_thread_abort_syscall_callback, NULL);
}
```
?

Or, why is this not just calling ves_icall_System_Threading_Thread_Interrupt_internal ()?
With an intervening mono_thread_set_state/mono_thread_clr_state (thread, ThreadState_WaitSleepJoin);
Why is it so different than other interruptable kernel waits or syscalls?

The comment about Linux seems off.
It is only common sense that a close on one thread will not interrupt
the kernel on another when it has already taken a reference to the underlying data.
Doing a close to implement this would be dangerous.

Nobody tries to interrupt other waits or I/O with  close, so should they not interrupt socket functions that way either.

luhenry:

luhenry:

Issue #9102, opened 2018-06-12
Bump minimum requirement to Windows 7.


jaykrell:
The source currently tacitly supports Windows Vista.
We should increment this to Windows 7.

i.e. here:
```
$ grep 600 *ac win*h
configure.ac:		CPPFLAGS="$CPPFLAGS -DWINVER=0x0600 -D_WIN32_WINNT=0x0600 -D_WIN32_IE=0x0501 -D_UNICODE -DUNICODE -DWIN32_THREADS -DFD_SETSIZE=1024"
winconfig.h:#if _WIN32_WINNT < 0x0600
winconfig.h:#endif /* _WIN32_WINNT < 0x0600 */
```

jaykrell:
Change 600 to 601.

jaykrell:
I'm not sure about the 501.

marek-safar:
We should align the minimum Windows requirement with netcore, see https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md

jaykrell:
Yeah, oops, it looks like Unity still runs on Vista so I guess close or delay this a while.
The code I was going to make require Win7, I can keep back at Vista.

netcore requires Win7 though.

akoeplinger:
We should really talk to Unity about whether they still need Vista. It's out of Microsoft support since April 2017 and we could silently call Win7+ APIs already since corefx assumes that as the baseline.

marek-safar:
@joncham could you confirm the Unity minimal OS requirements going forward?

akoeplinger:
Here's one example: [DriveInfo.Windows.cs#L65](https://github.com/dotnet/corefx/blob/ceda383ca0c165dbcc274f34476538051bf42f8d/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.Windows.cs#L65) uses `SetThreadErrorMode()` which is a Win7+ API

jaykrell:
Here is another Windows 7 dependency:
https://github.com/dotnet/corefx/blob/96eddf39539879915e5bf570825e78af37149154/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.Windows.cs#L15

We don't use this code currently, and Vista is easy to support with a few more lines.

jaykrell:
Indeed!

akoeplinger:
Done with https://github.com/mono/mono/pull/9568.

PR #9249, opened 2018-06-21
[corlib]: Cleanup random number generation and use CoreFX implementation where possible.

baulig:
We now use the CoreFX implementation of both the cryptographic and the non-cryptographic
random number generator on all platforms where this is supported.

* Mac and Windows use the CoreFX PAL for both RNGs.

* Linux/Unix uses the CoreFX PAL for the non-crypto RNG and we have a custom native
  implementation of the crypto RNG (using boringssl, they use openssl).

* We have a fallback implementation using `RNGCryptoServiceProvider` (like our [`Guid.FastNewGuidArray`](https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Guid.cs#L66)).

* `HashCode` is now enabled in all profiles.

* This removes the distinction between the `basic` and `net_4_x` profiles, so we can
  cleanup the .sources files quite a bit.

* Add `PLATFORMS = macos linux win32 unix` to the `basic` profile.

* Add fallback implementation using `RNGCryptoServiceProvider` (like our [`Guid.FastNewGuidArray`](https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Guid.cs#L66)) and use in on profiles which don't have a native implementation in CoreFX.

* We now use the fallback implementation in `corlib.dll.sources` - platforms which support
  on of the PAL versions need to explicitly override and remove the fallback via a `.exclude` file.


baulig:
Requires CoreFX merge of https://github.com/mono/corefx/pull/89.

mcs/class/corlib/win32_xammac_net_4_5_corlib.dll.sources
```diff
@@ -0,0 +1 @@
+#include win32_net_4_x_corlib.dll.sources
```
marek-safar: What is the `win32_xammac_net_4_5_corlib.dll.sources` file for?

mcs/class/corlib/corlib.dll.sources
```diff
@@ -1543,7 +1542,10 @@ corefx/SR.cs
 corefx/CompareInfo.cs
 corefx/MemoryExtensions.cs
 corefx/GlobalizationMode.cs
+corefx/Interop.GetRandomBytes.cs
```
marek-safar: I think it might be more maintainable don't add this to common sources but add to the profiles where needed instead. It should be only few and for example, for monodroid we want to switch to native anyway.

(same with RandomNumberGen.ratorImplementation.Mono.cs)

mcs/class/corlib/monotouch_corlib.dll.exclude.sources
```diff
@@ -1,3 +1,4 @@
+corefx/RandomNumberGeneratorImplementation.Mono.cs
```
marek-safar: Why is corefx/Interop.GetRandomBytes.Mono.cs needed in this profile?

mcs/class/corlib/win32_corlib.dll.exclude.sources
```diff
@@ -0,0 +1,2 @@
+corefx/Interop.GetRandomBytes.cs
+corefx/RandomNumberGeneratorImplementation.Mono.cs
```
marek-safar: Are you missing Interop.GetRandomBytes.Mono.cs ?

mcs/class/corlib/corlib_xtest.dll.sources
```diff
@@ -146,6 +146,7 @@
 
 # System.Runtime
 # TODO: Many more to be included
+../../../external/corefx/src/System.Runtime/tests/System/HashCodeTests.netcoreapp.cs
```
marek-safar: Please also include

`../../../external/corefx/src/System.Security.Cryptography.Algorithms/tests/RandomNumberGeneratorTests*.cs` as we are switching the implementation

mcs/build/profiles/basic.make
```diff
@@ -9,6 +9,8 @@ use_monolite := $(wildcard $(monolite_flag))
 
 MONOLITE_MSCORLIB = $(monolite_path)/mscorlib.dll
 
+PLATFORMS = macos linux win32 unix
```
luhenry: This is it what you want to do. Same issue as https://github.com/mono/mono/pull/8984#discussion_r192228194

mcs/class/corlib/corefx/Interop.GetRandomBytes.Mono.cs
```diff
@@ -0,0 +1,29 @@
...
+			lock (_rngAccess) {
+				if (_rng == null)
+					_rng = new RNGCryptoServiceProvider ();
+				_rng.GetBytes (buffer);
+			}
+		}
+
+		internal static unsafe void GetRandomBytes (byte* buffer, int length)
+		{
+			var bytes = new byte[length];
```
luhenry: To avoid creating a new array and copy around, it would be worthwhile to pass the byte* to the RandomNumberGenerator, even if it means adding an internal API

mcs/class/corlib/win32_xammac_net_4_5_corlib.dll.sources
baulig: Should go away

mcs/class/corlib/monotouch_corlib.dll.exclude.sources
baulig: This is the .exclude.

mcs/class/corlib/win32_corlib.dll.exclude.sources
baulig: No.

mcs/class/corlib/corefx/Interop.GetRandomBytes.Mono.cs
baulig: Would be / might be - but this is unrelated to this particular issue and it may not be worth the time anyways.

baulig:
This is now fixed and should land.

akoeplinger:
Looks like it breaks the build on iOS:

```
ld: weak import of symbol '_CCRandomGenerateBytes' not supported because of option: -no_weak_imports for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[5]: *** [Makefile:1593: libmono-apple-crypto.la] Error 1
```

and mcs:

```
../../../external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(17,24): error CS0213: You cannot use the fixed statement to take the address of an already fixed expression
../../../external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(17,36): error CS0029: Cannot implicitly convert type `System.Span<byte>' to `byte*'
../../../external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(37,24): error CS0213: You cannot use the fixed statement to take the address of an already fixed expression
../../../external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(37,36): error CS0029: Cannot implicitly convert type `System.Span<byte>' to `byte*'
```

mono/metadata/Makefile.am
```diff
@@ -185,6 +185,16 @@ libmono_system_native_la_SOURCES = \
 
 libmono_system_native_la_CFLAGS = -I$(abs_top_srcdir)/external/corefx/src/Native/Unix/Common -I$(abs_top_srcdir)/external/corefx/src/Native/Unix/System.Native -Wno-typedef-redefinition
 
+if HOST_DARWIN
+lib_LTLIBRARIES += libmono-apple-crypto.la
```
akoeplinger: I don't feel strongly about it but should we name this `libmono-system-security-cryptography-native-apple.la`?

It'd better match the corefx naming and also what we did with `System.Native -> libmono-system-native.la`

akoeplinger:

akoeplinger:
The tarball build fails with:

```
21:02:08 CSC     [basic-unix] mscorlib.dll
21:02:14 /mnt/jenkins/workspace/build-source-tarball-mono-pullrequest/external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(17,42): error CS7036: There is no argument given that corresponds to the required formal parameter 'count' of 'RandomNumberGeneratorImplementation.GetBytes(byte[], int, int)'
21:02:14 /mnt/jenkins/workspace/build-source-tarball-mono-pullrequest/external/corefx/src/System.Security.Cryptography.Algorithms/src/Internal/Cryptography/RandomNumberGeneratorImplementation.cs(37,42): error CS7036: There is no argument given that corresponds to the required formal parameter 'count' of 'RandomNumberGeneratorImplementation.GetBytes(byte[], int, int)'
```

Presumably this is because of the basic.make change to add PLATFORMS, could you check out @luhenry's suggestion in https://github.com/mono/mono/pull/8984#discussion_r192228194 ?

baulig:
@akoeplinger The `CCRandomGenerateBytes` is because CoreFX does not have any OS version checks in their PAL and it's not supported in all versions of iOS.

I switched iOS to the fallback implementation - and it will keep using the fallback implementation until someone from the iOS team has provided a recommended way of dealing with those version checks.

baulig:
@akoeplinger I believe the tarball failure was due to incorrect unix profiles.

baulig:
The idea behind this Pull Request was to use the CoreFX PAL for the two different Random Number Generators.  Unfortunately, the amount of work that's required to finish this by far exceeds my worst nightmares.  I am leaving this open to give folks who are interested in it the chance to finish it.

As it turns out, there are several major challenges that need to be mastered first:

- The CoreFX PAL does not support all platforms that we do.  There needs to be some authoritative decision about what to do with those.  It's not too difficult in terms of this PR due to the fallback mechanics - but it has the potential of hurting us badly in the long-run in other projects.
- The CoreFX PAL uses native APIs that do not exist in all versions of the OS that we support - for instance `CCRandomGenerateBytes` is not available in older versions of iOS.  We need to come up with a plan on how to deal with these situations.
- Compiling corlib in a platform-specific way turned out to be a much larger challenge than I originally anticipated. 

However, my primary objective was to bring `HttpClient` from CoreFX, not to spend several weeks reworking RNGs, so I will now essentially abandon this and make the code that I need work with the fallback implementation - using Mono's existing `RNGCryptoServiceProvider`.

baulig:
Let's see this as an opportunity to get things right, figure out all these problems and find the right solutions to them.

Me "abandoning" this for `HttpClient` and using the fallback implementation just means that it will unblock my future work, thus giving us the time we need to get this right on all our platforms - and once that's done, we can then quickly switch out the implementation.

I simply do not want to be blocked on this anymore where this issue essentially prevents me from doing any further work on `HttpClient`.

baulig:
Moving forward, my plan is to create a new PR which adds the fallback implementation for all platforms and uses it in the places where I need it.   Then switching out the implementation - aka finishing and landing this PR - just means editing the .sources files.

jaykrell:
https://github.com/mono/mono/issues/9102#issuecomment-400895780 is relevant here.

mcs/build/profiles/basic.make
luhenry: This review is not staled. This is not the way to go, fix it.

baulig:
The plan is to split this into smaller, more manageable pieces to unblock both me and the runtime team from other work.

Issue #9216, opened 2018-06-19
CoreFX System.Security.Cryptography.X509Certificates namespace


marek-safar:
## Types in System.dll to import

- [ ] CertificateRequest
- [ ] OpenFlags
- [ ] PublicKey
- [ ] StoreLocation
- [ ] StoreName
- [ ] X500DistinguishedName
- [ ] X500DistinguishedNameFlags
- [ ] X509BasicConstraintsExtension
- [ ] X509Certificate2
- [ ] X509Certificate2Collection
- [ ] X509Certificate2Enumerator
- [ ] X509CertificateCollection
- [ ] X509Chain
- [ ] X509ChainElement
- [ ] X509ChainElementCollection
- [ ] X509ChainElementEnumerator
- [ ] X509ChainPolicy
- [ ] X509ChainStatus
- [ ] X509ChainStatusFlags
- [ ] X509EnhancedKeyUsageExtension
- [ ] X509Extension
- [ ] X509ExtensionCollection
- [ ] X509ExtensionEnumerator
- [ ] X509FindType
- [ ] X509IncludeOption
- [ ] X509KeyUsageExtension
- [ ] X509KeyUsageFlags
- [ ] X509NameType
- [ ] X509RevocationFlag
- [ ] X509RevocationMode
- [ ] X509Store
- [ ] X509SubjectKeyIdentifierExtension
- [ ] X509SubjectKeyIdentifierHashAlgorithm
- [ ] X509VerificationFlags

## Required Dependencies
- [ ] AppleCrypto PAL #9417
- [ ] SafePasswordHandle #9416

baulig:
**IMPORTANT**  Prior to working on any of these classes, please talk to me (Martin Baulig) first, so we can discuss a few things and make sure it is heading into the right direction.

Most importantly, we need to make a plan on how to proceed before writing any code.  This will require some research and discussion among the team as well.

- [ ] CoreFX and OpenSSL
CoreFX currently uses OpenSSL, which is unacceptable.
Can we either convince them to switch to boringssl or can we port their openssl PAL to boringssl?
- [ ] Which PAL do we want to use on iOS and what is our plan about missing features?  Several of the native crypto and certificate APIs are not supported on iOS, do we want to throw `PlatformNotSupportedException` or do we need some workaround?
- [ ] Does CoreFX have a native PAL for X.509 extensions for all platforms that we care about?  Do we need a managed implementation for some platforms?  If so, could we possibly contribute that to CoreFX?

[...]

PR #9250, opened 2018-06-21
WIP: Use AsnEncodedData, X.509 extension and X.500 name classes from CoreFX.

filipnavara:
Required CoreFX changes:
- [x] Trailing delimiter compatibility for X500DistinguishedName (https://github.com/dotnet/corefx/pull/30572)
- [x] T61 string support (https://github.com/dotnet/corefx/issues/27466, https://github.com/dotnet/corefx/pull/30572)
- [x] X509EnhancedKeyUsageExtension compatibility issue (https://github.com/dotnet/corefx/issues/30559, https://github.com/dotnet/corefx/pull/30574)

Part of #9216


filipnavara:
Reminder to myself: Enable xunit tests.

baulig:
Well, I really wish you had talked to me prior to working on this, so we could have discussed a few things.

There are certainly some things in this pull request that I really like, but there is also other stuff that needs to be reworked before it can go in.  I also would have preferred to split the individual issues into smaller pieces - for instance all the `X500DistinguishedName` changes should have gone into their own PR without being intermixed with certificate extension code.

Before we move forward on this, we should come up with a plan on what needs to be done and how we want to do this.  Now that you started to work on this is probably a good time to write down such a plan.  If I remember correctly, then there is an issue open about this somewhere - I'll try to find it and update it with what I think needs to be done then get back to you.

baulig:
https://github.com/mono/mono/issues/9216

filipnavara:
I knew there's plenty of issues with this patch. That's one of the reasons why it's marked as WIP. I actually have the whole X509Certificate & friends ported over from CoreFX as a proof of concept. Unfortunately there are many issues with the CoreFX code and I am slowly going through them. It's hard to dissect the patches into small pieces that work by itself. The X509 code on macOS fails roughly 100-150 of the Mono unit tests.  I will be happy to split out the useful stuff into individual PRs and cooperate on anything that needs to be done.

filipnavara:
I have my own list of things to discuss/address and I will share them in the issue as soon as I get back to my laptop. Thanks for the time to write it all down.

mcs/class/System/Mono.Btls/X509CertificateImplBtls.cs
```diff
@@ -285,10 +285,7 @@ void MustFallback ()
 				ThrowIfContextInvalid ();
 				if (issuerName == null) {
 					using (var xname = x509.GetIssuerName ()) {
-						var encoding = xname.GetRawData (false);
-						var canonEncoding = xname.GetRawData (true);
-						var name = MonoBtlsUtils.FormatName (xname, true, ", ", true);
-						issuerName = new X500DistinguishedName (encoding, canonEncoding, name);
+						issuerName = new X500DistinguishedName (xname.GetRawData (false));
 					}
```
baulig: If we go down this path, then `MonoBtlsUtils.FormatName()` should be removed completely and all its other uses replaced with your new code.   However, we should be careful about comparing different `X500DistinguishedName` and avoid managed parsing.

For BTLS specifically, I was actually planning on replacing that `MonoBtlsUtils.FormatName()` method with a native boringssl based implementation for performance reasons.

mcs/class/System/Internal.Cryptography/AsnFormatter.Managed.cs
```diff
@@ -0,0 +1,303 @@
...
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Internal.Cryptography
```
baulig: Prior to adding a new managed implementation to Mono, we should talk to the CoreFX team and ask them whether they might be interested in adding this to their PAL.

Should we then eventually decide to put it into Mono, then we should try to avoid namespaces to could possibly conflict with future CoreFX work.  To make it consistent with other pieces, a namespace such as for instance `Mono.Cryptography`  is probably best, considering that most of our internal code in `System.dll` is already in namespaces that start with `Mono.`

mcs/class/System/Internal.Cryptography/X509Pal.cs
```diff
@@ -0,0 +1,61 @@
...
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+using System;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Internal.Cryptography.Pal
+{
+	// Temporary until real X509Pal from CoreFX could be used.
```
baulig: We should separate the `X500DistinguishedName` from the general `X509Certificate` code.  A general `X509Pal` class might not be possible because the code needs to be integrated with our TLS Provider framework.

As a general rule, we should ask ourselves whether or not a particular class needs to hold any native resources such as pointers to some native object.

mcs/class/System/Internal.Cryptography/X509Pal.cs
```diff
@@ -0,0 +1,61 @@
...
+	internal sealed partial class X509Pal
+	{
+		public static ManagedX509Pal Instance = new ManagedX509Pal ();
+
+		private X509Pal ()
+		{
+		}
+	
+		public partial class ManagedX509Pal : ManagedX509ExtensionProcessor
```
baulig: Let's make this an internal top-level class and find a better name for it.  We are currently using names such as for instance `X509ChainImpl`, but I wouldn't mind transitioning to a new "Pal" based naming scheme.

Where does `ManagedX509ExtensionProcessor` come from, is that a new CoreFX type?

mcs/class/System/System.Security.Cryptography.X509Certificates/X509Certificate2ImplMono.cs
```diff
@@ -62,6 +62,8 @@ internal class X509Certificate2ImplMono : X509Certificate2Impl
```
baulig: I am not sure whether it's a good idea to make changes to this class as it is going to be retired once we have removed the Legacy TLS Provider.

Adding new features to it is also a bit problematic in the sense that it would commit us to also supporting these in the new implementation.

mcs/class/System/System.Security.Cryptography.X509Certificates/X509ChainImplMono.cs
```diff
@@ -580,6 +580,62 @@ private void Process (int n)
 
 			// TODO 6.1.3.f - verify explict_policy > 0 if valid_policy_tree != NULL
 		}
+
```
baulig: Please do not put valuable new code into a class that is scheduled for retirement.

Since you have already added a new internal PAL-based class for `X500DistinguishedName`, that would be the perfect place for these two methods and make them public in that internal class.  I definitely see value in having a new managed implementation of this, it just shouldn't be hidden in here but find a home where it can be reused in other code as well.

mcs/class/System/common.sources
```diff
@@ -147,18 +147,28 @@ System.Security.Authentication.ExtendedProtection.Configuration/ExtendedProtecti
...
 System.Security.Authentication.ExtendedProtection.Configuration/ServiceNameElementCollection.cs
 
-System.Security.Cryptography/AsnEncodedData.cs
-System.Security.Cryptography/AsnEncodedDataCollection.cs
-System.Security.Cryptography/AsnEncodedDataEnumerator.cs
+../../../external/corefx/src/Common/src/System/Security/Cryptography/DerEncoder.cs
+../../../external/corefx/src/Common/src/System/Security/Cryptography/DerSequenceReader.cs
+../../../external/corefx/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs
+../../../external/corefx/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Oids.cs
+../../../external/corefx/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/ManagedX509ExtensionProcessor.cs
```
baulig: I haven't looked at their PAL yet, but you are referencing a directory called `Pal.Unix` from `common.sources`.

If these classes are platform-specific, then they need to go into the relevant platform-specific `.sources` files and we need to make sure to have an implementation for our platforms.

Can we use this `Pal.Unix` on iOS and Android and what about Windows?

mcs/class/System/Internal.Cryptography/X509Pal.cs
filipnavara: This basically boils down to whether we want to support only one implementation per platform or not. CoreFX strictly uses AppleTLS on macOS, OpenSSL on Linux and native implementation on Windows. Mono currently builds both BoringSSL and AppleTLS code on macOS and switches them at runtime. I have an open CoreFX issue on making the code a little bit more flexible so both options are possible (https://github.com/dotnet/corefx/issues/30606).

This was just a placeholder class to supplement whatever platform-specific implementation CoreFX uses now and default to the fully managed one. In this dependant PR - https://github.com/dotnet/corefx/pull/30572 - we agreed to drop the OpenSSL-specific code for X500 name decoding, so that part of PAL interface may soon be gone too.

mcs/class/System/Internal.Cryptography/X509Pal.cs
filipnavara: The X509Pal name is referenced in the CoreFX code, the ManagedX509Pal class is just temporary one to make this small piece of code compile without the full platform specific backends.

ManagedX509ExtensionProcessor comes from CoreFX and implements all the X509Extension code on all non-Windows platforms (but works on Windows too).

mcs/class/System/Mono.Btls/X509CertificateImplBtls.cs
filipnavara: Unfortunately I don't understand this piece of code well enough. Apparently the CanonicalName is used only in X509ChainMonoImpl. I started with the (wrong) assumption that X509ChainBtlsImpl is used with BTLS and thus it's ok to remove.

mcs/class/System/System.Security.Cryptography.X509Certificates/X509Certificate2ImplMono.cs
filipnavara: The change is moving code from internal X509ExtensionCollection constructor directly to the only place where it was used. The constructor is not present in CoreFX code.

mcs/class/System/System.Security.Cryptography.X509Certificates/X509ChainImplMono.cs
filipnavara: It is not a new code. The code is implementation that was formerly present in X500DistinguishedName, but used only in this one place. I moved it over.

mcs/class/System/common.sources
filipnavara: It is located in Pal.Unix, but it is actually a fully managed implementation that works on all the platforms. On CoreFX it's used on all non-Windows platforms and a separate implementation is provided for Windows. It would make sense to eventually use the CoreFX Windows implementation on Windows, but it's not strictly necessary.

mcs/class/System/Mono.Btls/X509CertificateImplBtls.cs
baulig: Yeah, that code is really complicated.  No worries, I can help you with that.

mcs/class/System/Internal.Cryptography/X509Pal.cs
baulig: The primary reason why we need both implementations at the same time is that it would otherwise become nearly impossible to test certain things.  For instance, triggering a TLS Renegotiation is not supported in boringssl, so the only way of properly testing renegotiation in the BTLS client-side code is by using AppleTLS on the server-side.  In addition to that, I also don't want the BTLS code to bit-rot and the best way of preventing it is by having it in active use.  Most of us, including myself, are working on Macs most of the time and not on Unix/Linux - so the best way of keeping it in active use is by having it built on Mac with the option to quickly and easily switch.

mcs/class/System/common.sources
baulig: Oh cool!  Maybe we could suggest to them to rename it into something like `Pal.Managed` to make it more obvious that it's not just for Unix?

mcs/class/System/Internal.Cryptography/X509Pal.cs
filipnavara: Definitely makes sense to me. I also work mostly on Mac these days and I test with both the Apple and BTLS code bases. For me a runtime switch is always a nice option, but it requires CoreFX changes. I am slowly going through process of upstreaming my local CoreFX changes. I explored this one too, but ultimately I finished it only for the X509* classes and the macOS backend and it now sits on a stale branch. I will wait for feedback from CoreFX people and revisit it later.

mcs/class/System/common.sources
marek-safar: yeah, that code should not be under Pal.Unix/ in corefx

mcs/class/System/Internal.Cryptography/OidLookup.Managed.cs
```diff
@@ -0,0 +1,76 @@
...
+		private static string NativeFriendlyNameToOid(string friendlyName, OidGroup oidGroup, bool fallBackToAllGroups)
+		{
+			string oid;
+			if (s_extraFriendlyNameToOid.TryGetValue (friendlyName, out oid)) {
+				return oid;
+			}
+			return null;
+		}
+
+		private static readonly Dictionary<string, string> s_extraFriendlyNameToOid =
```
marek-safar: Use switch statement instead

mcs/class/System/System.Security.Cryptography.X509Certificates/X509Certificate2ImplMono.cs
marek-safar: It's fine to move it here to simplify the migration

mcs/class/System/common.sources
filipnavara: It's kind of a common occasion in the X509 code that a managed implementation is present in the Pal.Unix / Pal.OSX directory. As part of https://github.com/dotnet/corefx/issues/30606 I would probably propose to rename couple of files to represent more accurately what it is supposed to do.

mcs/class/System/Internal.Cryptography/OidLookup.Managed.cs
marek-safar: Yeah, OID changes will be probably easier/faster to merge

mcs/class/System/System.Security.Cryptography.X509Certificates/X509ChainImplMono.cs
baulig: Well, it looks like it's general-purpose code that could possibly be reused elsewhere and I also want to retire `X509ChainImplMono` soon.

filipnavara: It does look general purpose, but it also looks like the equality is not very well defined - at least for this particular implementation. Right now the method is not used anywhere else and CoreFX implements the chain in the PAL in a different way.

filipnavara: Once X509ChainImplMono (& the rest of the legacy code) gets retired then this patch will shrink considerably.

baulig:
I started to give this some love by splitting their `Internal/Cryptography/Pal.OSX/CertificatePal.cs` into a managed part that we can reuse and the actual PAL: https://github.com/dotnet/corefx/pull/30707.

For the `corlib.dll` versus `System.dll` problem, we are already doing some kind of dependency injection for BTLS, but not for AppleTLS (because there is no native API that we could use for `X509Certificate2`) - so maybe that's something that we could also use for AppleTLS.

mcs/class/System/Internal.Cryptography/AsnFormatter.Managed.cs
baulig: We should eventually land this.

mcs/class/System/Internal.Cryptography/X509Pal.cs
baulig: Could you possibly take this out of this PR?  I have a better idea for this in my current working copy and splitting it up into pieces will make it easier.

mcs/class/System/Mono.Btls/X509CertificateImplBtls.cs
baulig: This code is gone in my working tree.

mcs/class/System/Mono.Btls/X509CertificateImplBtls.cs
```diff
@@ -334,10 +331,7 @@ void MustFallback ()
 				ThrowIfContextInvalid ();
 				if (subjectName == null) {
 					using (var xname = x509.GetSubjectName ()) {
```
baulig: Same, I'm using the managed CoreFX PAL for this, so we don't have to worry about this.

mcs/class/System/Mono.Net.Security/SystemCertificateValidator.cs
```diff
@@ -356,8 +356,7 @@ static bool CheckCertificateUsage (X509Certificate2 cert)
 				// last chance - try with older (deprecated) Netscape extensions
 				X509Extension ext = cert.Extensions ["2.16.840.1.113730.1.1"];
 				if (ext != null) {
```
baulig: Let's land this.

mcs/class/System/System.Security.Cryptography.X509Certificates/X509BasicConstraintsExtension.cs
```diff
@@ -1,236 +0,0 @@
-//
```
baulig: We cannot replace these extension classes with the CoreFX version until we have the full PAL support.

mcs/class/System/System.Security.Cryptography.X509Certificates/X509Certificate2ImplMono.cs
baulig: I have essentially completely replaced the `X509Certificate2ImplMono` class in my latest working tree.

filipnavara:
I tried to rebase the branch against master, but broke it in the process. A fix will be coming soon. :) I'll see if I can extract the "good bits" into separate PRs then.

filipnavara:
@monojenkins build failed

filipnavara:
@monojenkins build failed

marek-safar:
I am closing this (we can reopen later if this becomes relevant again)

PR #9251, opened 2018-06-21
Werror at the end of configure.ac because autoconf requires warnings …

jaykrell:
…to be success -- fix for gcc 4.x.


rjvdboon:
Works for me again with this PR. Thanks!

Issue #9234, opened 2018-06-20
[System]: Renegotiation in AppleTls requires OS X 10.12+


MSylvia:
This failure also happens on 2018-06/3ee5a5cc3d9

## Steps to Reproduce

1. Install mono 5.16.0.5 (2018-06/7627a5f9eeb)

2. Compile sample
```C#
// Build with: csc curl.cs  -r:System.Net.Http.dll /langversion:7.1
using System;
using System.IO;
using System.Net.Http;
using System.Threading.Tasks;

using TTask = System.Threading.Tasks.Task;


class App {
	public static async Task Main (string[] urls)
	{
		var tasks   = new TTask [urls.Length];
		using (var client = new HttpClient ()) {
			client.Timeout = TimeSpan.FromHours (3);
			for (int i = 0; i < urls.Length; ++i) {
				tasks [i] = DownloadFile (client, urls [i], Path.Combine (Directory.GetCurrentDirectory(), Path.GetFileName (urls [i])));
			}
			TTask.WaitAll (tasks);
		}		
	}

	static async TTask DownloadFile (HttpClient client, string uri, string destinationFile)
	{
		if (File.Exists (destinationFile)) {
			Console.WriteLine ($"Skipping uri '{uri}' as destination file already exists '{destinationFile}'.");
			return;
		}
		var dp       = Path.GetDirectoryName (destinationFile);
		var dn       = Path.GetFileName (destinationFile);
		var tempPath = Path.Combine (dp, "." + dn + ".download");
		Directory.CreateDirectory(dp);

		Console.WriteLine ($"Downloading `{uri}` to `{tempPath}`.");
		try {
			using (var s = await client.GetStreamAsync (uri))
			using (var o = File.OpenWrite (tempPath)) {
				await s.CopyToAsync (o);
			}
			Console.WriteLine ($"mv '{tempPath}' '{destinationFile}'.");
			File.Move (tempPath, destinationFile);
		}
		catch (Exception e) {
			Console.Error.WriteLine ("Unable to download URL `{0}` to `{1}`: {2}", uri, destinationFile, e.Message);
			Console.Error.WriteLine (e);
		}
	}
}
```
3. Run sample with argument
```
mono --debug curl.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
```
4. Crash:
```
Mono.Security.Interface.TlsException: Unknown Secure Transport error `-50'.
```

### Current Behavior

Crashes with error:
```
[...]

PR #9254, opened 2018-06-21
[System]: Renegotiation in AppleTls requires OS X 10.12+.  #9234.

baulig:
Fixes https://github.com/mono/mono/issues/9234

### DIFF
--- a/mcs/class/System/Mono.AppleTls/AppleTlsContext.cs
+++ b/mcs/class/System/Mono.AppleTls/AppleTlsContext.cs
@@ -341,9 +341,11 @@ namespace Mono.AppleTls
 				PeerDomainName = ServerName;
 			}
 
-			if (Options.AllowRenegotiation)
+			if (Options.AllowRenegotiation && Environment.OSVersion.Version >= MinRenegotiationVersion)
 				SetSessionOption (SslSessionOption.AllowRenegotiation, true);
 		}
+
+		static readonly Version MinRenegotiationVersion = new Version (16, 6);
 
 		void InitializeSession ()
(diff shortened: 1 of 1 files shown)


baulig:
@monojenkins backport 2018-06

mcs/class/System/Mono.AppleTls/AppleTlsContext.cs
```diff
@@ -341,10 +341,12 @@ void InitializeConnection ()
 				PeerDomainName = ServerName;
 			}
 
-			if (Options.AllowRenegotiation)
+			if (Options.AllowRenegotiation && Environment.OSVersion.Version >= MinRenegotiationVersion)
```
akoeplinger: will this work on iOS etc? I'm not sure which Environment.OSVersion we're getting there and the API is listed as iOS 10.0+, tvOS 10.0+, watchOS 3.0+.

shouldn't we just call CanRenegotiate? or is that just used on the server-side?

marek-safar: I think we need better "checking" method which does include all MONOTOUCH profiles. The option is supported https://developer.apple.com/documentation/security/sslsessionoption/ksslsessionoptionallowrenegotiation?language=objc

XI does it like https://github.com/xamarin/xamarin-macios/blob/54101252aca7e15a9e4fd04e98d1bfb49e411f05/src/IOSurface/IOSurfacePropertyKey.cs#L34-L43

akoeplinger:
@monojenkins backport 2018-06

PR #9261, opened 2018-06-22
Remove 3 unused incorrectly written functions.

jaykrell:
g_string_prepend does not update string->len.

g_string_insert does not update string->len, and copies the wrong number of characters around,
	and does not allow for overlapping parameters

g_string_erase does not always update string->len.

### DIFF
--- a/mono/eglib/eglib-remap.h
+++ b/mono/eglib/eglib-remap.h
@@ -94,7 +94,6 @@
 #define g_list_length monoeg_g_list_length
 #define g_list_nth monoeg_g_list_nth
 #define g_list_nth_data monoeg_g_list_nth_data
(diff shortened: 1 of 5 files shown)


mono/eglib/eglib-remap.h
```diff
@@ -94,7 +94,6 @@
 #define g_list_length monoeg_g_list_length
 #define g_list_nth monoeg_g_list_nth
 #define g_list_nth_data monoeg_g_list_nth_data
-#define g_list_prepend monoeg_g_list_prepend
```
lambdageek: I think you meant to get rid of `g_string_prepend`?

jaykrell: Right oops thank you.

jaykrell:
Review?

PR #9273, opened 2018-06-22
[runtime] Fix debugger dumping when unused

alexanderkyte:
Somewhere in the daily rebasing to edit the PR, I lost the correct version of the early-exit. This fixes our native crash when we did not use the debugger. It's important to merge, or we'll crash before MERP/gdb dumping happens.


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #9275, opened 2018-06-22
Disable handle type-safety for PowerPC, S390X, SPARC, MIPS.

jaykrell:


jaykrell:
(sticky Mac keyboard gave two spaces..)

lambdageek:

NattyNarwhal:
For the record: seems to unbreak the build on aix/ppc64

FWIW, is the type safe handle mode something that would need to be worked on down the road for the other targets and disabling is a transitional thing, or is it intended to be optional?

jaykrell:
It is probably indefinitely optional. 

It’d be nice to teach the marshaler more, or never return them by value, but it could easily be asis a very long time. Notice that even Linux/x86 does not have it, so you are in good company.

jaykrell:
Another vision I have here, is that the JIT<=>native types are different than the native<=>native types.
Either in C or in C++, with or without marshaling changes -- that is -- type safe handles do work on all platforms today at the native<=>native part, it is JIT<=>native that has the ABI sensitivity.

For example, we could have type:ObjectHandleFromJit and macro:RETURN_OBJ_TO_JIT and type:ObjectHandleToJit.

Or we could adapt the marshaling.

jaykrell:
@monojenkins build

Issue #9195, opened 2018-06-18
tailcall-rgctxb-static.exe fails on FullAOT armel/armhf runs


akoeplinger:
## Steps to Reproduce

1. Run tailcall-rgctxb-static.exe test on FullAOT armel or armhf (not arm64!)
2. It fails

### Current Behavior

It fails.

### Expected Behavior

No failure.

## On which platforms did you notice this

[ ] macOS
[ x ] Linux
[ ] Windows

**Version Used**: master

akoeplinger:
@jaykrell it looks like this started with https://github.com/mono/mono/commit/6f590e214ec9d8628dbd9d86d58fb142bb3241c3, can you please take a look?

jaykrell:
arm32 is quite challenged tailcall-wise and always will be, and we should  be willing to disable tests.
But if it works for JIT, then it should  be viable for AOT.

jaykrell:
ps: don't be surprised at arm32 vs. arm64 divergence -- arm32 is quite challenged, arm64 is not

jaykrell:
This test also fails armhf JIT if you pass -O=gsharedvt. So the FullAOT failure is not surprising.
This test does also fail FullAOT with -O=-gsharedvt, which is surprising.
The test pretty deliberately focuses on this case, which ARM32 does have trouble with.
The choices are roughly:
1. disable test
2. disable gsharedvt by default, or selectively

jaykrell:
Need to understand why this uses gsharedvt.

PR #9276, opened 2018-06-23
Disable tailcall-rgctxb-static.exe for ARM.

jaykrell:
Depending on flags it cannot easily work, i.e. gsharedvt.
https://github.com/mono/mono/issues/9195

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -955,6 +955,7 @@ PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail.exe
 PLATFORM_DISABLED_TESTS += tailcall-interface-conservestack.exe
 PLATFORM_DISABLED_TESTS += ivtail1.exe
 PLATFORM_DISABLED_TESTS += tailcall/coreclr/JIT/Directed/tailcall/tailcall.exe  # interface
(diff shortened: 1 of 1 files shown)


jaykrell:
Could see about making it work without gsharedvt however. Is anyone likely to benefit from that? Would we consider turning off gsharedvt by default in fullaot? Or removing gsharedvt entirely?

jaykrell:
Relevant-seeming code..as to why AOT is different than JIT..need more investigation...mini.c:

```        if (compile_aot) {
                if (is_open_method (method)) {
                        try_generic_shared = TRUE;
                        method_is_gshared = TRUE;
                } else {
                        try_generic_shared = FALSE;
                }
                g_assert (opts & MONO_OPT_GSHARED);
```

jaykrell:
That assert is trivial to make fail btw, just by saying like mono  -O=-gshared -aot=full  foo.exe and some generics in foo.exe. I kinda think it shouldn't be there, but it did get me looking here..

jaykrell:
Let's not commit this until mono -O=-gsharedvt -aot failure is better understood.
The tests do run with  -O=gsharedvt -aot however.

jaykrell:
Seems not needed now.

PR #9279, opened 2018-06-23
Cleanup some confusing code.

jaykrell:
Had additional restructure not been ok, at least typedefing the function pointer
  would have helped.

Add const for efficiency -- make the copy-on-write faults count for more and
  not take extra readonly data with them.

Remove unused and actually confusing parameter -- it is just zeroed memory
and almost cannot be used for anything (granted, could hash it).

### DIFF
--- a/mono/metadata/w32handle.c
+++ b/mono/metadata/w32handle.c
@@ -33,7 +33,7 @@ struct _MonoW32HandleSlot {
 };
 
 static MonoW32HandleCapability handle_caps [MONO_W32TYPE_COUNT];
-static MonoW32HandleOps *handle_ops [MONO_W32TYPE_COUNT];
+static MonoW32HandleOps const *handle_ops [MONO_W32TYPE_COUNT];
 
 static MonoW32HandleSlot *handles_slots_first;
 static MonoW32HandleSlot *handles_slots_last;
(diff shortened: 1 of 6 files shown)


jaykrell:
review?

mono/metadata/w32handle.c
```diff
@@ -471,10 +471,7 @@ w32handle_destroy (MonoW32Handle *handle_data)
 
 	mono_coop_mutex_unlock (&scan_mutex);
 
-	close_func = _wapi_handle_ops_get_close_func (type);
-	if (close_func != NULL) {
-		close_func (handle_data, handle_specific);
-	}
+	mono_w32handle_ops_close (type, handle_specific);
```
lambdageek: I like the cleanup around replacing `_wapi_handle_ops_get_close_func` (which returns the `close` callback) by `mono_w32handle_ops_close` which just invokes the `close` callback.

I'm a little surprised that `handle_data` is really not needed by any close operation - it's got stuff in there that seems like it might be useful - but apparently not.

https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mono/metadata/w32handle.h#L39-L47

jaykrell: Note that it is zeroed before the callback. You can’t do much with it, and there is approx. only one callback. You could hash the address, is about all.

PR #9281, opened 2018-06-24
[jit] Constant folding for some Math operations on doubles

migueldeicaza:
This adds support for constant folding of additions and
multiplications of doubles as well as constant folding of Math.Sin,
Math.Cos, Math.Abs and Math.Sqrt.

Inspired by some changes that went into RyuJit:

https://github.com/AndreyAkinshin/aakinshin.net/blob/master/web/_posts/en/2015/ryujit-rc-and-constant-folding.md


mono/mini/intrinsics.c
```diff
@@ -1235,7 +1278,7 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		ins = llvm_emit_inst_for_method (cfg, cmethod, fsig, args, in_corlib);
 		if (ins)
 			return ins;
-	}
+	}
```
vargaz: Extra whitespace.

mono/mini/intrinsics.c
```diff
@@ -1139,6 +1139,49 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
...
+						result = cos (source);
+						break;
+					case OP_SIN:
+						result = sin (source);
+						break;
+					case OP_ABS:
+						result = fabs (source);
+						break;
+					}
+					*dest = result;
```
vargaz: Hopefully these will return the same value on the host and target machines when cross compiling.

migueldeicaza: They are constants, so they should be the same.

I wanted to get your opinion generally on the code so far that adds the capability.

vargaz: There could be precision problems between the implementations on the different platforms. Also, some functions are not defined for some values, so some error handling is needed here.
You might need something like:
https://en.wikipedia.org/wiki/GNU_MPFR

migueldeicaza: You mean for the constant folding of floating point values, or the math operations that are being inlined?    I will check with RyuJIT is doing.

vargaz:
Both probably.

mono/mini/intrinsics.c
```diff
@@ -1139,6 +1139,49 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		 * all inputs:
 		 * http://everything2.com/?node_id=1051618
 		 */
+		if (cmethod->klass == mono_defaults.math_class){
```
kumpera: This should be done in cprop

migueldeicaza:
I forgot where we ended up on the internal discussion on this.

As far as I can tell, the current patch as it is should be ok.

lewurm:
The comment by Zoltan isn't addressed yet: https://github.com/mono/mono/pull/9281#discussion_r197654275
Alternatively, we could disable this optimization for AOT.

kumpera:
This needs to address the precision issue due to us intrinsifying sin/cos to use x86 transcendental instructions.

migueldeicaza:
@kumpera please clarify what this precision issue is

migueldeicaza:
@lewurm I do not see @vargaz comment on that link, nor on the commit page - or you mean the whitespace?

lewurm:
@migueldeicaza hrm, you have to click here to make my link work (https://github.com/mono/mono/pull/9281#discussion_r197654275 )
screen shot 2018-09-19 at 19 38 33


github UI bug?

kumpera:
@migueldeicaza https://xamarinhq.slack.com/archives/D03CJ4CFZ/p1530796386000366

migueldeicaza:
Short version: while there are differences in the result between using the standard library implementation of these math functions and the FPU version of it, the difference in result are (a) expected and within the accepted range of differences for math operations (Epsilon), and (b) making this change is within the parameters that other .NET implementations have done

Let me summarize the information that I gathered from the chat discussions that we had back in July.   It was a discussion that we did not continue as we went on vacations.

The implementation of functions like `Math.Sin` takes different forms on different runtimes and operating systems.   Sometimes they can use the FPU instruction, sometimes the C library implementation, the following table shows what is used (found out by testing every one):

|Runtime            |Uses         |
|-------------------|-------------|
|Mono (everywhere)  |FPU intrinsic|
|Mono (everywhere/LLVM)  |LLVM-specific, on Intel this happens to be library|
|.NET Core (Windows)|FPU intrinsic|
|.NET Core (Mac)    |library      |
|.NET Windows       |library      |

The challenge is that some of the FPU instructions (like fsin, fcos) produce different results than the methods found in the system library that implements sin/cos due to differences in how the functionality is implemented [1].

What the table above illustrates is that .NET Core has a different behavior between Windows and Linux., yet .NET Core on Mac is compatible with .NET Framework on Windows.   Today, Mono is compatible with Mono on Windows.

.NET Framework on Windows used to use FPU intrinsics, and at some point move to the library implementation, as explained here:

https://github.com/dotnet/coreclr/blob/13c4ab0742ee7f80fb1cb98bef95e106b0f9c83a/src/classlibnative/float/floatdouble.cpp#L12

## Mono Consistency ##

There is an additional dimension, which is that Mono only uses the intrinsics on x86, x86-64 which means that Mono itself is not consistent with its own results based on the application platform, as we only use intrinsics on x86/x86-64.

The other dimension happens when using LLVM, which turns the llvm intrinsic for sin/cos into a library call.  Which makes Mono compatible with .NET Core on Mac, and .NET on Windows.

## How different are the results? ##

I decided to explore what the difference in the results were, as this might be a problem affecting more than use, so for any case where the outputs of FPU/sin and libc/sin did not match, I printed out the difference.

The difference are numbers in the 1.11022302462516E-16 to 5.55111512312578E-17 which look like this:
0.0000000000000000111022302462516.

Not only do they look very small to me, but then I realized that the largest of those two numbers (ie, the worst case scenario) happens to be the very definition of Epsilon for 64-bit double values: https://en.wikibooks.org/wiki/Floating_Point/Epsilon (section: Determining Epsilon).   

## What does Clang do with these calls in the C source code? ##

When clang is faced with a math operation like `sin(CONSTANT)`, it inlines the value computed by `sin(CONSTANT)` as computed on the host into the generated output.

## My take ##

My take is that given that I could not find anyone complaining about the change that took place at some point in the .NET Framework, that this has not been raised as a problem between Core across platforms (likely early to tell), that we have different behaviors already, that the error is 

## Appendix ##

[1] Given that Mono uses fsin/fcos to implement Math.Sin, Math.Cos, the following sample shows how Math.Sin differs from sin (x):

```
using System;
// using System.Runtime.CompilerServices;


class Driver {
    [DllImport("libc")]
    static extern double sin (double v);

    static void Try (double d) {
        if (Math.Sin (d) != sin (d)) {
            Console.WriteLine (d);
            Environment.Exit (-1);
        }
    }
    static void Main () {
        double d = 0;
        for (int i = 0; i < 1000 * 1000 * 10; ++i) {
            Try (d);
            d += 0.0001;
        }
        Console.WriteLine (d);
    }
}
```

[2] 
```
using System;

class Driver {
    [DllImport("libc")]
    static extern double sin (double v);
	static int total, fmtfail;
	
    static void Try (double d) {
	    Console.WriteLine ("{3} {0}-> Math.Sin ()={1} sin()={2}", d, Math.Sin (d), sin (d), Math.Sin (d) != sin (d) ? "F" : "O");
	    if (Math.Sin (d) != sin (d)) 
		total++;
		
	    if (Math.Sin (d).ToString () != sin (d).ToString ()){
		    Console.WriteLine ("FMTFAIL");
		    fmtfail++;
	    }
		
    }
    static void Main () {
        double d = 0;
        for (int i = 0; i < 1000 * 1000 * 10; ++i) {
            Try (d);
            d += 0.0001;
        }
        Console.WriteLine ("{0} failed {1} fmtfail {2}", d, total, fmtfail);
    }
}
```

migueldeicaza:
@monojenkins build failed

luhenry:
@monojenkins build failed

luhenry:
@monojenkins build

lewurm:
fixed a build issue. getting this failure on my machine:

```
1) TestDoubleRound (MonoTests.System.MathTest.TestDoubleRound)
     #4
  Expected: True
  But was:  False

  at MonoTests.System.MathTest.TestDoubleRound () [0x00087] in /Users/lewurm/work/mono-miguel-constant-folding/mcs/class/corlib/Test/System/MathTest.cs:874
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Users/lewurm/work/mono-miguel-constant-folding/mcs/class/corlib/System.Reflection/MonoMethod.cs:325
```

https://github.com/mono/mono/blob/141e1e48a5a8a5ec559b665fe34ba878d8dcb9a1/mcs/class/corlib/Test/System/MathTest.cs#L874

lewurm:
`OP_ROUND` wasn't implemented, fixed it.

lewurm:
couple more failures on x86:
https://github.com/mono/mono/blob/141e1e48a5a8a5ec559b665fe34ba878d8dcb9a1/mcs/class/corlib/Test/System/MathTest.cs#L45
https://github.com/mono/mono/blob/141e1e48a5a8a5ec559b665fe34ba878d8dcb9a1/mcs/class/corlib/Test/System/MathTest.cs#L315-L316
and
```
Test run: image=/Users/builder/jenkins/workspace/test-mono-pull-request-i386-osx/mono/mini/basic-math.exe, opts=peephole,branch,inline,cfold,consprop,copyprop,deadce,linears,cmov,intrins,loop,aot,exception,float32,gshared,simd,alias-analysis
Test 'test_0_sqrt_precision_and_spill' failed result (got 1, expected 0).
Results: total tests: 27, failed: 1, cfailed: 0 (pass: 96.30%)
```

PR #9283, opened 2018-06-24
[aot] Pass the soft-float argument to llc properly, the -soft-float option doesn't seem to exist anymore.

vargaz:


vargaz:
Fixes the armel lane on
https://jenkins.mono-project.com/job/test-mono-mainline-linux-fullaot+llvm/

PR #9293, opened 2018-06-25
Add support for dynamic symbol resolve for all JIT unwind info registration functions on Windows

lateralusX:
Using dynamic symbol resolve (LoadLibrary/GetProcAddress) for all unwind info related registration functions enables building the same code on all Windows platforms where JIT is build and included.

### DIFF
--- a/mono/mini/exceptions-amd64.c
+++ b/mono/mini/exceptions-amd64.c
@@ -1086,6 +1086,10 @@ static SRWLOCK g_dynamic_function_table_lock = SRWLOCK_INIT;
 
 // Module handle used when explicit loading ntdll.
 static HMODULE g_ntdll;
+static HMODULE g_kernel32dll;
+
+static RtlInstallFunctionTableCallbackPtr g_rtl_install_function_table_callback;
+static RtlDeleteFunctionTablePtr g_rtl_delete_function_table;
 
 // If Win8 or Win2012Server or later, use growable function tables instead
 // of callbacks. Callback solution will still be fallback on older systems.
@@ -1109,6 +1113,8 @@ init_table_no_lock (void)
 	if (g_dyn_func_table_inited == FALSE) {
 		g_assert_checked (g_dynamic_function_table_begin == NULL);
 		g_assert_checked (g_dynamic_function_table_end == NULL);
+		g_assert_checked (g_rtl_install_function_table_callback == NULL);
+		g_assert_checked (g_rtl_delete_function_table == NULL);
 		g_assert_checked (g_rtl_add_growable_function_table == NULL);
 		g_assert_checked (g_rtl_grow_function_table == NULL);
 		g_assert_checked (g_rtl_delete_growable_function_table == NULL);
@@ -1120,6 +1126,14 @@ init_table_no_lock (void)
 			g_rtl_add_growable_function_table = (RtlAddGrowableFunctionTablePtr)GetProcAddress (g_ntdll, "RtlAddGrowableFunctionTable");
 			g_rtl_grow_function_table = (RtlGrowFunctionTablePtr)GetProcAddress (g_ntdll, "RtlGrowFunctionTable");
 			g_rtl_delete_growable_function_table = (RtlDeleteGrowableFunctionTablePtr)GetProcAddress (g_ntdll, "RtlDeleteGrowableFunctionTable");
+		}
+
+		// Fallback on systems not having RtlAddGrowableFunctionTable.
+		if (g_rtl_add_growable_function_table == NULL) {
+			if (GetModuleHandleEx (0, TEXT ("kernel32.dll"), &g_kernel32dll) == TRUE) {
+				g_rtl_install_function_table_callback = (RtlInstallFunctionTableCallbackPtr)GetProcAddress (g_kernel32dll, "RtlInstallFunctionTableCallback");
+				g_rtl_delete_function_table = (RtlDeleteFunctionTablePtr)GetProcAddress (g_kernel32dll, "RtlDeleteFunctionTable");
+			}
 		}
 
 		g_dyn_func_table_inited = TRUE;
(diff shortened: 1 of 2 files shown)


mono/mini/exceptions-amd64.c
```diff
@@ -1122,6 +1128,14 @@ init_table_no_lock (void)
 			g_rtl_delete_growable_function_table = (RtlDeleteGrowableFunctionTablePtr)GetProcAddress (g_ntdll, "RtlDeleteGrowableFunctionTable");
 		}
 
+		// Fallback on systems not having RtlAddGrowableFunctionTable.
+		if (g_rtl_add_growable_function_table == NULL) {
+			if (GetModuleHandleEx (0, TEXT ("kernel32.dll"), &g_kernel32dll) == TRUE) {
```
jaykrell: Comparison to TRUE is a mistake. The success values are non-zero, not specifically  TRUE. Just say `if (GetMod..())` not `if (GetMod...() == TRUE)`

Is anything really ever in kernel32.dll? Is anything ever missing from ntdll.dll? I think you can just use ntdll.dll.

jaykrell: Also, this TEXT business is a Win9x compat layer. Nobody needs that any  longer.
Say `GetModuleHandleExW(0, L"kernel32.dll")...`

mono/mini/exceptions-amd64.c
```diff
@@ -1167,6 +1181,14 @@ terminate_table_no_lock (void)
 			g_ntdll = NULL;
 		}
 
+		g_rtl_delete_function_table = NULL;
+		g_rtl_install_function_table_callback = NULL;
+
+		if (g_kernel32dll != NULL) {
+			FreeLibrary (g_kernel32dll);
```
jaykrell: Never FreeLibrary kernel32.dll. It doesn't do anything but waste code and cycles.

mono/mini/exceptions-amd64.c
jaykrell: Ok, well MSDN, does say kernel32.dll. Still probably just a forwarder to ntdll and just use ntdll. "Rtl" basically means "ntdll".

jaykrell: Another option to strongly consider:
 The v1 (XP) functions are present on all systems, in ntdll.lib.
 Just build a custom ntdll.lib to bind to them directly -- no LoadLibrary/GetProcAddress then.

mono/mini/exceptions-amd64.c
jaykrell: Also remove the FreeLibrary of ntdll.dll.
Imagine if unloading these .dlls really did anything.
ntdll.dll is loaded specially by the kernel.
ntdll.dll, the .exe, and direct and indirect dependencies of the .exe are special and not unloadable. They do not actually maintain a reference count.

lateralusX: Just following the documentation of the GetModuleHandleEx API:

From MSDN:

"dwFlags [in]

This parameter can be zero or one or more of the following values. If the module's reference count is incremented, the caller must use the FreeLibrary function to decrement the reference count when the module handle is no longer needed."

So not doing this for kernel32.dll is based on internal information not described in the above documentation. I also know that this is really not needed in practice, due to internal implementation details, just following the public API documentation of the API.

No big deal, keeping it won't harm, I'm fine to remove it as well.

mono/mini/exceptions-amd64.c
lateralusX: Will fix the compare against TRUE to none zero.

jaykrell:
I suggest more like this:

```
typedef struct _MonoDelayload {
	PROC* ptr;
	PCSTR name;
} MonoDelayload;

const static MonoDelayload function_table_functions [ ] = {
	{ (PROC*)&g_rtl_install_function_table_callback, "RtlInstallFunctionTableCallback" },
	{ (PROC*)&g_rtl_delete_function_table, "RtlDeleteFunctionTable" },
	{ (PROC*)&g_rtl_delete_growable_function_table, "RtlDeleteGrowableFunctionTable" },
	{ (PROC*)&g_rtl_grow_function_table, "RtlGrowFunctionTable" },
	{ (PROC*)&g_rtl_add_growable_function_table, "RtlAddGrowableFunctionTable" },
};

static void
init_table_no_lock (void)
// This function is essentially idempotent, cannot fail, can be called any number of times,
// and requires no locking.
{
	gboolean missing = FALSE;
	for (int i = 0; !missing < G_N_ELEMS (functions); ++i)
		missing = !*function_table_functions [i].ptr;
	if (!missing)
		return;
	const static PCSTR dllnames [ ] = {
		L"ntdll.dll",
#ifdef WIN32_KERNEL32_NAME
		WIN32_KERNEL32_NAME
#else
		L"kernel32.dll"
#endif
	};
	HMODULE dlls [G_N_ELEMENTS (dllnames)] = { 0 };
	for (int i = 0; i < G_N_ELEMENTS (dlls); ++i) {
		if (!dlls [i])
#ifdef UWP // or whatever
			GetModuleHandleExW (0, dllnames [i], &dlls [i]);
#else
			dlls [i] = LoadLibraryW (dllnames [i]);
			g_assert (dlls [i]);
#endif
	}
	g_assert (dlls [0]);
	for (int i = 0; !missing < G_N_ELEMENTS (function_table_functions); ++i) {
		MonoDelayload const * const function = &function_table_functions [i];
		if (*function->ptr)
			continue;
		for (int j = 0; j < G_N_ELEMENTS (dlls); ++j) {
			HMODULE const dll = dlls [j];
			if (!dll)
				continue;
			PROC const proc = GetProcAddress (dll, function->name);
			// Do not write NULL to avoid race with writes of non-NULL.
			if (proc)
				*function->ptr = proc;
		}
	}
	// Make sure we got enough.
	// Windows 8 and newer have all the functions.
	// pre-Windows 8 does not have growable function tables.
	g_assert (g_rtl_install_function_table_callback && g_rtl_delete_function_table);
	g_assert (g_rtl_add_growable_function_table == g_rtl_delete_growable_function_table);
	g_assert (g_rtl_add_growable_function_table == g_rtl_grow_function_table);
}

static void
terminate_table_no_lock (void)
{
	// Free all list elements and the list.
	for (GList *l = g_dynamic_function_table_begin; l; l = l->next)
		g_free (l->data);
	g_list_free (g_dynamic_function_table_begin);
	g_dynamic_function_table_begin = NULL;
	g_dynamic_function_table_end = NULL;
}
```

lateralusX:
Could of course be done that way as well using a data driven approach, but I believe what we currently have in this commit is good enough for this use case.

PR #9266, opened 2018-06-22
[2018-04] unbreak interp-only mode in Xamarin.iOS

lewurm:
Context: https://github.com/xamarin/xamarin-macios/pull/3856#issuecomment-398930387

This PR
* backports #8838 (already exists on `master` and `2018-06`)
* fixes an AOT compiler regression for interp support due to 67e995c6e27eb9a66b352ab0cdf923373c72eceb (this needs to be forward ported to `master` and `2018-06`)
* adding two workarounds to force interp-only execution. This is fine for this branch, because we aren't there yet regarding mixed mode execution, but I need to work on a better solution for `master`/`2018-06`.

Once approved, I'm going to propagate fixes to the other branches.

PR #9295, opened 2018-06-25
[mini] make interp work with AOT image generated with aot=interp

lewurm:
[aot-compiler] use mono_aot_mode_is_interp instead of directly using the flag

---------
[interp] add FULL_AOT flags to AOT image when `--aot=interp` is used

fixes regression from 67e995c6e27eb9a66b352ab0cdf923373c72eceb

---------
partial forward port of https://github.com/mono/mono/pull/9266

needs backporting to `2018-06`


lewurm:
@monojenkins build failed

mono/mini/aot-compiler.c
```diff
@@ -6950,7 +6950,7 @@ emit_trampolines (MonoAotCompile *acfg)
 	int tramp_type;
 #endif
 
-	if ((!mono_aot_mode_is_full (&acfg->aot_opts) || acfg->aot_opts.llvm_only) && !acfg->aot_opts.interp)
+	if ((!mono_aot_mode_is_full (&acfg->aot_opts) || acfg->aot_opts.llvm_only) && !mono_aot_mode_is_interp (&acfg->aot_opts))
 		return;
```
vargaz: The two mean exactly the same, the current one is jut easier to type.

mono/mini/aot-compiler.c
```diff
@@ -7879,7 +7879,7 @@ compile_method (MonoAotCompile *acfg, MonoMethod *method)
 	 * does not need to support them by creating a fake GOT etc.
 	 */
 	flags = JIT_FLAG_AOT;
-	if (mono_aot_mode_is_full (&acfg->aot_opts))
+	if (mono_aot_mode_is_full (&acfg->aot_opts) || mono_aot_mode_is_interp (&acfg->aot_opts))
 		flags = (JitFlags)(flags | JIT_FLAG_FULL_AOT);
```
lewurm: this change is not needed imho, need to test it.

mono/mini/aot-compiler.c
```diff
@@ -12630,7 +12630,7 @@ mono_compile_assembly (MonoAssembly *ass, guint32 opts, const char *aot_options,
 		}
 	}
 
-	if (mono_aot_mode_is_full (&acfg->aot_opts)) {
+	if (mono_aot_mode_is_full (&acfg->aot_opts) || mono_aot_mode_is_interp (&acfg->aot_opts)) {
 		acfg->flags = (MonoAotFileFlags)(acfg->flags | MONO_AOT_FILE_FLAG_FULL_AOT);
```
lewurm: @vargaz you mentioned on Slack that we shouldn't use this this for `--aot=interp`. 

I'll look into adding `MONO_AOT_FILE_FLAG_INTERP`

lewurm:
updated PR. It now contains changes to make the interpreter work with an AOT images that is generated with `--aot=interp` only (instead of `--aot=full,interp`). It depends on https://github.com/mono/mono/pull/9385

@vargaz please review.

lewurm:
@monojenkins build

vargaz:
There is already an array in the aot image which can be used to check whenever a method was AOT compiled, can that be used instead of the 'nextra_methods' ?

lewurm:
@vargaz which array do you mean?

vargaz:
amodule->methods will contain a -1 for methods which are missing.

lewurm:
`amodule->methods` builds an index via `nmethods` which already includes wrappers (aka. "extra methods). Say you have an AOT image with 100 wrappers and 0 regular methods, `amodule->methods` can't be used to determine if a method is included in the image when `mono_metadata_token_index (method->token) - 1` provides something that is `<100`.

vargaz:
Looks ok then. It might be easier to add a MONO_AOT_FILE_FLAG_HAS_NORMAL_METHODS or something.

lewurm:
we might want to be able to AOT a subset of methods of a given DLL in the future, so I think this is approach is better than a flag.

vargaz:
If we do that then we need to emit the method_addresses array for them with most entries -1.

lewurm:
True, but there could be still tricky scenarios like that:
* compile 10 methods
* highest token index is 30
* have 100 wrappers

so `amodules->methods` will be filled with 130 entries (20 being -1), you still have to check that method index somehow. It's the same problem we are trying to solve now.

Not sure if this is a constructed example, but I could imagine to run into this kind of issues with very small DLLs.

vargaz:
I think with the current code, you can't really add a subset of 'normal' methods, since method_index == token, so you either add all of them, or none. By 'adding' i mean adding it to the various tables, it doesn't have to be compiled.

lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2018-06

PR #9297, opened 2018-06-25
Use System.Security.Cryptography.Oid* from CoreFX.

filipnavara:
Extracted from #9250 (needs .csproj update).


mcs/class/System/Internal.Cryptography/OidLookup.Managed.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+
+namespace Internal.Cryptography
+{
+	/// <summary>
+	/// Fully managed implementation of OidLookup for the commonly supported Oids.
+	/// </summary>
+	internal static partial class OidLookup
```
filipnavara: This needs to be in the same namespace as the partial class in CoreFX. It lists the OIDs that are missing in OidLookup.cs in CoreFX compared to Mono prior to this patch. It may be worth it to move the missing oids to the generic OidLookup and submit that as PR to CoreFX, which would at the same time render their OidLookup.OSX.cs obsolete.

filipnavara: Apparently the Windows names that are preserved in OidLookup.cs are only those that are not localized in different languages. The expectation is that if the names could differ on Windows between languages they could also differ between platforms. So, short term it makes some sense to keep a Mono-specific implementation. Long term it makes sense to switch to the OidLookup.[OSX/Unix/Windows].cs along with the full PAL layer and remove the unit tests that depend on the names. Thoughs?

filipnavara:
@monojenkins commit csproj

filipnavara:
cc @baulig

mcs/class/System/common.sources
```diff
@@ -768,9 +767,12 @@ ReferenceSources/Win32Exception.cs
 ../referencesource/System/security/system/security/Authentication/ExtendedProtection/TokenBinding.cs
 ../referencesource/System/security/system/security/Authentication/ExtendedProtection/ServiceNameCollection.cs
 
-../referencesource/System/security/system/security/cryptography/oid.cs
-
-../referencesource/System/security/system/security/cryptography/x509/x509utils.cs
+../../../external/corefx/src/System.Security.Cryptography.Encoding/src/System/Security/Cryptography/OidGroup.cs
```
marek-safar: Please start including the tests as well

marek-safar:
@monojenkins build failed

mcs/class/System/common.sources
filipnavara: Will add them, running a local build now before I push it to GIT.

filipnavara:
@monojenkins build failed

filipnavara:
@monojenkins build failed

Issue #6921, opened 2018-02-09
Hybrid Suspend


marek-safar:
# Goal

As a way to reduce the risk of bing-bang change, we are approaching the project of transitioning to cooperative suspend by a middle ground approach: hybrid suspend. It consist in using cooperative suspend for runtime and managed code, and preemptive suspend for any other kind of code (external libraries, OS, etc.). Thus, preemptive suspend is used only for code which is not controlled by Mono, while cooperative suspend is only used for code which is fully controlled by Mono. This forces us to make sure we transition to the right state, to ensure we use the proper mechanism of suspending.

A _hopefully_ exhaustive list of where these transitions are needed is:
 1. managed-to-native wrappers
 2. native-to-managed wrappers
 3. MONO_API functions
 4. trampolines
 5. external libraries calls (libc, btls, etc.)

Points 1. and 2. are quite easy to achieve, and are already working for supporting cooperative suspend on watchOS. Some of 3., 4. and 5. have already been done for watchOS and Mono's CI for cooperative suspend. Unfortunately, 3. and 4. are not complete enough to fully support VSfM, XI/XM and XA, or any other embedding project.

We want to switch to hybrid suspend by default by `2018-08`, and to do so we have identified some necessary steps to ship, and some longer-term improvement steps.

# Tasks

## For `2018-08`

 - [x] Implement hybrid suspend (native -> preemptive; runtime+managed -> cooperative) - #8068
 - [x] `MONO_API` functions and trampolines
   - List functions loaded from XI/XM and XA, and functions which are public API: https://gist.github.com/luhenry/f450d55128deddcedc2e27852ab6dc23 /cc @luhenry
   - Ensure transition in the above functions and trampolines
 - [ ] Performance: measure impact of hybrid suspend + fix major regressions
   - Some possible measurements
     - time-to-suspend
     - wall clock time: benchmarks, VSfM (startup time, new projects, new file, etc.)
 - [x] Documentation: how to disable, what's needed to transition embedders, how to use diagnostics
 - [ ] Timeout time-to-suspend on CI
 - [x] Diagnostics: `mono_trace` in `mono-threads-debug.h`
 - [ ] Resolve failing tests
   - [x] `abort-cctor.exe` #7599 
   - [x] `bug-58782-plain-throw.exe` and `bug-58782-capture-and-throw.exe` #5837
[...]

PR #9299, opened 2018-06-25
[runtime] Second batch of GC Unsafe thread state transitions for the MONO_API

lambdageek:
For hybrid suspend we need all entry points into the runtime from native code to be annotated with transitions to GC Unsafe mode.  This is the second set of changes after #8847. The overall project is #6921 

One sizeable change here is marking `mono_class_get_method_from_name` as external only - and updating the runtime to use `mono_class_get_method_from_name_checked`.


luhenry:

luhenry:

luhenry:

lambdageek:
@monojenkins build failed

PR #9305, opened 2018-06-25
[wasm] Small set of changes that are wasm specific.

kumpera:


akoeplinger:
@kumpera https://github.com/GoogleChromeLabs/jsvu#supported-engines seems to imply Linux support, do we need to upgrade to a later version?

kumpera:
@akoeplinger perhaps that would fix it, let me try locally.

kumpera:
@akoeplinger you're right. With 1.3.1 linux should work.

Issue #8127, opened 2018-04-09
No BTLS in Windows builds


directhex:
Add support for BTLS on Windows targeting both cygwin/mingw and msvc build runtimes.

- [x] Build BTLS as part of cygwin/mingw build. #9306.
- [x] Validate pass on basic BTLS tests using cygwin/mingw build (NOTE, by default CI uses msvc build runtime to run tests). #9306.
- [x] Build BTLS as part of msvc build. #11100.
- [x] Do a better integration between autoconf and msvc build. Implement logic to pickup build parameters from autoconf step (in this case BTLS but should be other configs as well) and add logic to build, use and link external repro's as part of msvc build without additional configuration. #11100.
- [x] Add support for assembler BTLS builds on cygwin and msvc builds. #11234.
- [x] Update Mono Windows build instructions. #11343, https://github.com/mono/website/pull/411
- [x] Update documentation related to additional build dependencies when build using BTLS. #11343, https://github.com/mono/website/pull/411

akoeplinger:
I assume you can tell cmake to output VS project files and then build that?

directhex:
Maybe! Will be exploring the whole thing this week.

baulig:
What's the status of this?  Not having BTLS on Windows is blocking the Legacy TLS removal (https://github.com/mono/mono/issues/8559).

directhex:
I was told to drop it, in favour of totally imminent work on implementing a Windows SSL back-end (presumably from Netcore)

directhex:
I was told to un-drop this.

Right now, my main blocker is the `curve25519` module.

Like many BTLS modules, this has an optimized assembly implementation. Unlike most, this was added by Google in the form of raw assembly (`.S`), rather than using an assembly-generating stub script (`.pl`)

Issue 1 is CMake skipping it entirely from the build, causing failures down the line. This is fixed by copying the `.S` file to a `.asm` file, as the extension for Assembly on Windows is .asm, and the CMake vcxproj generation will only add assembly rules for `.asm`, not `.S`

Next is the issue that the `.S` file is strictly in GNU Assembly format, and doesn't build with NASM/YASM.

I'm going to give building a no-assembly build a shot.

directhex:
[...]

PR #9306, opened 2018-06-25
Build BTLS on Windows, cygwin/mingw.

baulig:
Part of #8127

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3905,7 +3905,7 @@ case "$host" in
 			;;
 		  mingw*|cygwin*)
 			have_visibility_hidden=no
-			BTLS_SUPPORTED=yes
+			BTLS_SUPPORTED=no
 			BTLS_PLATFORM=i386
 			;;
 		  haiku*)
@@ -3955,7 +3955,7 @@ case "$host" in
 			BTLS_PLATFORM=x86_64
 			;;
 		  mingw*|cygwin*)
-			BTLS_SUPPORTED=yes
+			BTLS_SUPPORTED=no
 			BTLS_PLATFORM=x86_64
 			;;
 		esac


filipnavara:

marek-safar:
@monojenkins build failed

filipnavara:
Note that even if the build gets fixed it is still missing the Windows x86 changes. It should probably be fixed within one PR.

filipnavara:
The build seems to fail because the non-CygWin CMake is given a CygWin path.

marek-safar:
@lateralusX could you help with this?

lateralusX:
I can take a look at this during the week. Another side note, maybe we should include a msvc build of BTLS since the msvc build is what's actually get's tested on Jenkins, not the cygwin build?

directhex:
@lateralusX hopefully that's not too onerous - we have some understanding of mix-and-match of CMake & MSBuild now, from your work with LLVM

lateralusX:
@baulig, @directhex Did some local experimentation around BTLS cygwin build. I believe the best way is to install cmake as part of the cygwin install and use the cygwin version when doing cygwin build. When doing BTLS VS build we should use Window installed cmake version. Ones cygwin cmake was installed the build completed on x64 Windows:

lateralusX:
Pushed commit fixing btls cygwin x86, x64. NOTE this won't build on CI until build boots have cygwin cmake installed. I have updated the build documentation with that change,

https://github.com/mono/website/commit/bbb27729f216584e69a5cc496c757179360b91df#diff-6ff39b0903ab2de908c0926d1369c0e3

@akoeplinger Could we get cmake onto the jenkins Windows build boots?

lateralusX:
Next step is to look at doing the VS build of BTLS as well, since all tests will run using the Visual Studio build mono and not the cygwin build.

lateralusX:
Rebased on latest master. Build and validated cygwin/mingw build BTLS on both x86 and x64 Windows using simple TLS test. Verified that BTLS shared library gets loaded and that we get failures from BTLS provider when for example getting certificate validation errors.

lateralusX:
Failures looks unrelated and occurs on multiple other builds. Approve/Merge?

directhex:
Adding to the MSI is an assigned task for me & @joewstroman for this sprint.

lateralusX:
@directhex, Right now, BTLS is not default for Windows builds unless you pass --enabled-btls to autogen.sh, so nothing will be build when using default build configs at the moment. Regarding the dllmap, I tried to fix the default naming and use mono-btls-shared in dllimport's since normally the mono loader will prefix module name with lib if it fails to load using regular name, but there where other errors related to the naming on some platforms. One option is to use libmono-btls-shared for Windows build as well, since we only include one build in the msi anyways (just the msvc build and not mingw builds), correct? Since cygwin/mingw build using libmono-btls-shared.dll I believe the best is to use the same naming on Windows build for this module as well to reduce additional complexity.

PR #9307, opened 2018-06-26
[WIP] [jit] Add new OP_EXTEND/OP_INSERT opcodes to represent loading/storing fields of vtypes. Extend the alias analysis pass to convert loads/stores to the new opcodes. Only enabled for Span<T>/ReadOnlySpan<T> for now.

vargaz:
Part of #7249

### DIFF
--- a/mono/mini/alias-analysis.c
+++ b/mono/mini/alias-analysis.c
@@ -150,19 +150,17 @@ lower_store_imm (MonoCompile *cfg, MonoInst *store, MonoInst *ldaddr)
 	return TRUE;
 }
 
-static gboolean
-lower_memory_access (MonoCompile *cfg)
+/*
+ * compute_defs:
+ *
+ * Find all vregs whose only definition is an ldaddr opcode.
+ */
+static G_GNUC_UNUSED void
+compute_defs (MonoCompile *cfg, MonoInst **defs)
 {
 	MonoBasicBlock *bb;
-	MonoInst *ins, *tmp;
-	gboolean needs_dce = FALSE;
-	GHashTable *addr_loads = g_hash_table_new (NULL, NULL);
-	MonoInst **defs;
+	MonoInst *ins;
 
-	/*
-	 * Find all vregs whose only definition is an ldaddr opcode.
-	 */
-	defs = g_malloc0 (sizeof (MonoInst*) * cfg->next_vreg);
 	for (bb = cfg->bb_entry; bb; bb = bb->next_bb) {
 		for (ins = bb->code; ins; ins = ins->next) {
 			if (ins->dreg != -1) {
@@ -194,6 +192,20 @@ lower_memory_access (MonoCompile *cfg)
 			}
 		}
 	}
+}
+
+static gboolean
+lower_memory_access (MonoCompile *cfg)
+{
+	MonoBasicBlock *bb;
+	MonoInst *ins, *tmp;
+	gboolean needs_dce = FALSE;
+	GHashTable *addr_loads = g_hash_table_new (NULL, NULL);
+	MonoInst **defs;
+
+	defs = g_malloc0 (sizeof (MonoInst*) * cfg->next_vreg);
+	if (COMPILE_LLVM (cfg))
+		compute_defs (cfg, defs);
 
 	for (bb = cfg->bb_entry; bb; bb = bb->next_bb) {
 		g_hash_table_remove_all (addr_loads);
@@ -349,6 +361,8 @@ mini_is_scalar_repl_class (MonoClass *klass)
 		return FALSE;
 	if (mono_class_get_field_count (klass) != 3)
 		/* Fast span */
+		return FALSE;
+	if (mini_is_gsharedvt_klass (klass))
 		return FALSE;
 	return TRUE;
 }
--- a/mono/mini/alias-analysis.c
+++ b/mono/mini/alias-analysis.c
@@ -471,6 +471,293 @@ done:
 		mono_print_code (cfg, "AFTER ALIAS_ANALYSIS");
 }
 
+typedef struct {
+	MonoInst *var;
+	MonoClass *klass;
+	MonoClassField *fields;
+	int fcount;
+	MonoInst **scalars;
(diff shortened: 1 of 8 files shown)


mono/mini/mini-ops.h
```diff
@@ -185,6 +185,15 @@ MINI_OP(OP_FMOVE,	"fmove", FREG, FREG, NONE)
 MINI_OP(OP_VMOVE,   "vmove", VREG, VREG, NONE)
 MINI_OP(OP_RMOVE,	"rmove", FREG, FREG, NONE)
 
+/* Load the field at inst_offset from the vtype sreg1 */
+/* inst_imm is the field index of the field loaded */
+MINI_OP(OP_EXTRACTI4, "extracti4", IREG, VREG, NONE)
+MINI_OP(OP_EXTRACTI, "extracti", IREG, VREG, NONE)
+
+/* Set the field at inst_offset in the vtype sreg1 from sreg2, store the result into dreg */
+MINI_OP(OP_INSERTI4, "inserti4", VREG, VREG, IREG)
```
kumpera: I tried this design in the past and its interaction with CPROP is really tricky unless you force something like you did, which is disallow `dreg != sreg1`.

This PR lacks the code that generates them.

mono/mini/alias-analysis.c
```diff
@@ -157,7 +157,44 @@ lower_memory_access (MonoCompile *cfg)
...
+			if (defs [i] == GINT_TO_POINTER (-1) || defs [i]->opcode != OP_LDADDR)
+				defs [i] = NULL;
+		}
+	}
+
+	for (int i = 0; i < cfg->next_vreg; ++i) {
+		if (defs [i]) {
+			ins = defs [i];
+			MonoInst *var = (MonoInst*)ins->inst_p0;
+			if (var->flags & MONO_INST_VOLATILE) {
```
kumpera: This must check for MONO_INST_INDIRECT as well. But this would only makes sense if we do it after we do extract/insert conversion.

This needs to be tested with other sources of indirection such as passing/return from functions and taking a ref to a span.

mono/mini/alias-analysis.c
```diff
@@ -208,17 +245,41 @@ lower_memory_access (MonoCompile *cfg)
...
 				if (tmp) {
-					if (cfg->verbose_level > 2) { printf ("Found candidate load:"); mono_print_ins (ins); }
-					if (lower_load (cfg, ins, tmp)) {
-						needs_dce = TRUE;
-						/* Try to propagate known aliases if an OP_MOVE was inserted */
-						goto handle_instruction;
+					MonoInst *var = (MonoInst*)tmp->inst_p0;
+					if (MONO_TYPE_ISSTRUCT (var->inst_vtype) && ins->opcode != OP_LOADV_MEMBASE && mini_is_scalar_repl_class (var->klass)) {
+						/* Loading a field of a vtype */
+						if (ins->opcode == OP_LOADI4_MEMBASE || ins->opcode == OP_LOAD_MEMBASE) {
```
kumpera: How about changing this to call `mono_load_membase_to_extract` and check if opcoce is `> 0`?

mono/mini/alias-analysis.c
```diff
@@ -208,17 +245,41 @@ lower_memory_access (MonoCompile *cfg)
...
+				// FIXME: Unify the two kinds of lookups
+				tmp = defs [ins->sreg1];
 				if (tmp) {
-					if (cfg->verbose_level > 2) { printf ("Found candidate load:"); mono_print_ins (ins); }
-					if (lower_load (cfg, ins, tmp)) {
-						needs_dce = TRUE;
-						/* Try to propagate known aliases if an OP_MOVE was inserted */
-						goto handle_instruction;
+					MonoInst *var = (MonoInst*)tmp->inst_p0;
+					if (MONO_TYPE_ISSTRUCT (var->inst_vtype) && ins->opcode != OP_LOADV_MEMBASE && mini_is_scalar_repl_class (var->klass)) {
```
kumpera: I assume there's no need to handle nested structs?

mono/mini/decompose.c
```diff
@@ -1459,6 +1459,40 @@ mono_decompose_vtype_opts (MonoCompile *cfg)
...
+					break;
+				}
+				case OP_INSERTI4:
+				case OP_INSERTI: {
+					MonoInst *var, *addr;
+
+					if (COMPILE_LLVM (cfg) && !mini_is_gsharedvt_klass (ins->klass))
+						break;
+
+					g_assert (ins->dreg == ins->sreg1);
```
kumpera: Don't we need to add safegards to cprop and other passes to make sure they won't break this?

mono/mini/mini-ops.h
vargaz: The code will be added later. Probably only going to enable this for llvm, since it has a similar insert instruction so the conversion is trivial.

mono/mini/alias-analysis.c
vargaz: Extract is going to be translated back to a load in the vtype decomposition pass. So this is just a temporary instruction used to get rid of ldaddr opcodes, making the vtype non-indirect.

mono/mini/alias-analysis.c
vargaz: Not right now since we only support Span<T>.

mono/mini/decompose.c
vargaz: Will probably only use this for llvm.

mono/mini/alias-analysis.c
kumpera: The issue is that MONO_INST_INDIRECT has the actual effect you're trying to catch here. I did cleanup usage of MONO_INST_VOLATILE a few years ago to only mean volatile and not "address taken".

vargaz: This code tries to compute vregs whose only definition of OP_LDADDR, not anything more, so we can do ldaddr+load -> extract transformations.

vargaz:
This doesn't seem to actually generate better code with llvm, closing.

PR #9309, opened 2018-06-26
[Coop] Some conversion of custom-attrs.c.

jaykrell:


mono/metadata/object-internals.h
```diff
@@ -2019,4 +2042,9 @@ mono_class_get_virtual_method (MonoClass *klass, MonoMethod *method, gboolean is
 MonoStringHandle
 mono_string_empty_handle (MonoDomain *domain);
 
+// FIXME value can be a raw pointer
+// We probably need mono_field_set_value_handle_ref and mono_field_set_value_handle_value
+void
+mono_field_set_value_handle (MonoObjectHandle obj, MonoClassField *field, void *value);
```
lambdageek: If this signature is going to be wrong (because `void *value` could be a raw ptr to managed), let's not add it - just put inline its body at the use site in `custom-attr.c` and leave a FIXME.

mono/metadata/object.c
```diff
@@ -6241,6 +6248,13 @@ mono_array_new_specific_checked (MonoVTable *vtable, uintptr_t n, MonoError *err
 	return (MonoArray*)o;
 }
 
+
+MonoArrayHandle
+mono_array_new_specific_handle (MonoVTable *vtable, uintptr_t n, MonoError *error)
+{ // FIXMEcoop
```
lambdageek: What needs to be fixed here?

Also nit: I don't think comments on the same line as a brace is Mono coding style.

mono/metadata/custom-attrs.c
```diff
@@ -767,14 +767,16 @@ decode_blob_value_checked (const char *ptr, const char *endp, guint32 *value_out
 	return is_ok (error);
 }
 
-static MonoObject*
-create_custom_attr (MonoImage *image, MonoMethod *method, const guchar *data, guint32 len, MonoError *error)
+static void
+create_custom_attr (MonoObjectHandle *result, MonoArrayHandle *result_array, int result_index, MonoImage *image, MonoMethod *method, const guchar *data, guint32 len, MonoError *error)
```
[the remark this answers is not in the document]
jaykrell: 1. I don't understand the "out" stuff. Just comments-via-naming, but always the same thing? ok.
2. I put the results first like assignment/memcpy, and this is replacing assignment, but ok.
3. ok. I think I didn't realize at first I'd have both.
more generally: I don't like extra single-use functions, and having to name them, but again this could be because I only later found the second user...and I have to check if it was also in a loop, and I never came up with good names for the loop helpers, so this avoids coming up with more similar names. I can try again here.

mono/metadata/custom-attrs.c
```diff
@@ -1193,34 +1197,27 @@ create_custom_attr_data_handle (MonoImage *image, MonoCustomAttrEntry *cattr, Mo
...
 	MonoReflectionAssemblyHandle assm = mono_assembly_get_object_handle (domain, image->assembly, error);
 	goto_if_nok (error, fail);
-	params [0] = ctor_obj;
+	params [0] = MONO_HANDLE_RAW (ctor_obj);
 	params [1] = MONO_HANDLE_RAW (assm);
-	params [2] = (gpointer)&cattr->data;
+	params [2] = &cattr->data;
 	params [3] = &cattr->data_size;
 
 	mono_runtime_invoke_checked (ctor, MONO_HANDLE_RAW (attr), params, error);
```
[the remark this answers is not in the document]
jaykrell: Good catch. invoke is kinda a large unfinished area..and I'm still a little nervous about what I did wrt "try" and the exception.

mono/metadata/object-internals.h
```diff
@@ -94,11 +94,20 @@
 		__arr = mono_array_new_specific_checked (__vtable, (size), (error)); \
 	__arr; })
 
+/* eclass should be a run-time constant */
```
jaykrell: I'd like to fix this to remove the gcc extension but not now.

mono/metadata/object-internals.h
jaykrell: ok. I'm not sure my comment is correct, but ok either way.

jaykrell: ps: do we need fixme anyway? I think there are like two phases here: The large obvious phase we are now in, and the later also maybe large mop-up driven by Centrinel. I think Centrinel will superset any few fixmes I leave.

mono/metadata/object.c
jaykrell: 1. Use of underlying raw pointer, maybe? Though see previous point -- Centrinel should guide is eventually, once the obvious pieces are reduced.
2. Maybe it could/should be? Plus it is a fixme so ideally gets deleted.

jaykrell: so ok I'll just remove it

mono/metadata/custom-attrs.c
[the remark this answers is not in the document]
jaykrell: btw 3 isn't right I think -- both can be non-null, but they don't tend to be. Anyway let me see if I can make it a little better.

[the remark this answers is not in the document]
jaykrell: This should clear up a lot.
```
static void
create_custom_attr_into_array (MonoImage *image, MonoMethod *method, const guchar *data,
	guint32 len, MonoArrayHandle array, int index, MonoError *error)
{
	// This function serves to avoid creating handles in a loop.
	HANDLE_FUNCTION_ENTER ();
	MonoObjectHandle attr = create_custom_attr (image, method, data, len, error);
	MONO_HANDLE_ARRAY_SETREF (array, index, attr);
	HANDLE_FUNCTION_RETURN ();
}
```

mono/metadata/custom-attrs.c
```diff
@@ -823,18 +826,12 @@ create_custom_attr (MonoImage *image, MonoMethod *method, const guchar *data, gu
 	}
 
 	named = p;
-	attr = mono_object_new_checked (mono_domain_get (), method->klass, error);
+	attr = mono_object_new_handle (mono_domain_get (), method->klass, error);
 	goto_if_nok (error, fail);
 
-	MonoObject *exc = NULL;
-	mono_runtime_try_invoke (method, attr, params, &exc, error);
+	(void)mono_runtime_try_invoke_handle (method, attr, params, error);
```
jaykrell: I'm a little nervous about this transform, that I recently started. Maybe restore the exception parameter. So far I *believe* it is equivalent, and we can how that holds up upon further conversion. Everyone checks error and then exc/set_exception_instance, so set_set_exception_instance in invoke. I didn't survey all users, just the ones I have changed.

mono/metadata/custom-attrs.c
```diff
@@ -1173,9 +1182,11 @@ ves_icall_System_Reflection_CustomAttributeData_ResolveArgumentsInternal (MonoRe
 	mono_error_set_pending_exception (error);
 }
 
-static MonoObjectHandle
-create_custom_attr_data_handle (MonoImage *image, MonoCustomAttrEntry *cattr, MonoError *error)
+static void
+create_custom_attr_data (MonoArrayHandle result, int index, MonoImage *image, MonoCustomAttrEntry *cattr, MonoError *error)
```
jaykrell: Append "into_array" to name?

jaykrell: (and put array last, before error?)

mono/metadata/custom-attrs.c
lambdageek: Yea the form with both `exc` and `error` was useful while we still had places where we called invoke but didn't have a way of threading `error` out, while `exc` was used in some manner.  We're probably in a better place now.

mono/metadata/custom-attrs.c
```diff
@@ -258,6 +258,7 @@ load_cattr_type_object_with_header (MonoImage *image, MonoType *t, const char *p
 	return load_cattr_type_object (image, t, p, boundp, end, error, slen);
 }
 
+// FIXMEcoop This appears to be polymorphic between managed pointers and unmanaged.
```
lambdageek: Yea, the return type on this function is a mess.

mono/metadata/object.c
```diff
@@ -6241,6 +6241,14 @@ mono_array_new_specific_checked (MonoVTable *vtable, uintptr_t n, MonoError *err
 	return (MonoArray*)o;
 }
 
+
+MonoArrayHandle
+mono_array_new_specific_handle (MonoVTable *vtable, uintptr_t n, MonoError *error)
+{
+	// FIXMEcoop invert relationship with mono_array_new_specific_checked
```
lambdageek: :+1:

jaykrell:
@monojenkins build failed

jaykrell:
Still waiting on arm...

jaykrell:
Overnight & still waiting on arm.

jaykrell:
Strange armv5 error.
```
/out:../../class/lib/net_4_x-linux/corlib.unsafe.dll.tmp
Assembling '../../class/corlib/System.Runtime.CompilerServices/Unsafe.il' , no listing file, to dll --> '../../class/lib/net_4_x-linux/corlib.unsafe.dll.tmp'

../../class/corlib/System.Runtime.CompilerServices/Unsafe.il : Warning -- Reference to undeclared extern assembly 'mscorlib', adding.

Unhandled Exception:
System.IO.DirectoryNotFoundException: Could not find a part of the path "../../class/lib/net_4_x-linux/corlib.unsafe.dll.tmp".
  at System.IO.File.Delete (System.String path) [0x00058] in <5f4eca14fe1d43f491a6984fb0c964bf>:0 
  at Mono.ILASM.Driver+DriverMain.Run () [0x00113] in <b3bc10d17e1e496abffabfb45cfbd15c>:0 
  at Mono.ILASM.Driver.Main (System.String[] args) [0x00015] in <b3bc10d17e1e496abffabfb45cfbd15c>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.IO.DirectoryNotFoundException: Could not find a part of the path "../../class/lib/net_4_x-linux/corlib.unsafe.dll.tmp".
  at System.IO.File.Delete (System.String path) [0x00058] in <5f4eca14fe1d43f491a6984fb0c964bf>:0 
  at Mono.ILASM.Driver+DriverMain.Run () [0x00113] in <b3bc10d17e1e496abffabfb45cfbd15c>:0 
  at Mono.ILASM.Driver.Main (System.String[] args) [0x00015] in <b3bc10d17e1e496abffabfb45cfbd15c>:0 
mkdir -p -- ../../class/lib/net_4_x-linux/
```

jaykrell:
@monojenkins build ARMv5

jaykrell:

jaykrell:
rebased to master, there was a minor merge conflict

jaykrell:
@monojenkins build Linux x64

jaykrell:
Overnight again not long enough for arm.

jaykrell:

jaykrell:

jaykrell:
@monojenkins squash

jaykrell:
@monojenkins build failed

jaykrell:

PR #9313, opened 2018-06-26
[sdks] Download then fallback to build MXE

luhenry:


luhenry:
This is not going to work because MXE binaries are not relocatable and thus needs to be built per-machine.

PR #9316, opened 2018-06-26
[tailcall] Split up large tests for fine grained enable/disable, instead of large scale disable (F# and interface).

jaykrell:
Centralize tailcall debugprint.
Tweak tests to account for x86 frame alignment -- this way they pass, and in the real world, we might consume a little stack instead of zero, but not a lot. The test checks for zero.

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -696,6 +696,163 @@ if HOST_DARWIN
 TESTS_CS_SRC += bug-8417.cs
 endif
 
+if CROSS_COMPILING # FIXME build vs. host vs. target
+TAILCALL_INTERFACE_CONSERVESTACK_IL=
+TAILCALL_FSHARP_DEEPTAIL_IL=
+else
+if HOST_WASM # FIXME build problem
+TAILCALL_INTERFACE_CONSERVESTACK_IL=
+TAILCALL_FSHARP_DEEPTAIL_IL=
+else
+if HOST_WIN32 # FIXME build problem
+TAILCALL_INTERFACE_CONSERVESTACK_IL=
+TAILCALL_FSHARP_DEEPTAIL_IL=
+else
+# These tests are produced by splitting up tailcall-interface-conservestack.il.
+TAILCALL_INTERFACE_CONSERVESTACK_IL=\
+	tailcall/interface-conservestack/1.il \
+	tailcall/interface-conservestack/10.il \
+	tailcall/interface-conservestack/11.il \
+	tailcall/interface-conservestack/12.il \
+	tailcall/interface-conservestack/13.il \
+	tailcall/interface-conservestack/14.il \
+	tailcall/interface-conservestack/15.il \
+	tailcall/interface-conservestack/16.il \
+	tailcall/interface-conservestack/17.il \
+	tailcall/interface-conservestack/18.il \
+	tailcall/interface-conservestack/19.il \
+	tailcall/interface-conservestack/2.il \
+	tailcall/interface-conservestack/20.il \
+	tailcall/interface-conservestack/21.il \
+	tailcall/interface-conservestack/22.il \
+	tailcall/interface-conservestack/23.il \
+	tailcall/interface-conservestack/24.il \
+	tailcall/interface-conservestack/25.il \
+	tailcall/interface-conservestack/26.il \
+	tailcall/interface-conservestack/27.il \
+	tailcall/interface-conservestack/28.il \
+	tailcall/interface-conservestack/29.il \
+	tailcall/interface-conservestack/3.il \
+	tailcall/interface-conservestack/30.il \
+	tailcall/interface-conservestack/31.il \
+	tailcall/interface-conservestack/32.il \
+	tailcall/interface-conservestack/33.il \
+	tailcall/interface-conservestack/34.il \
+	tailcall/interface-conservestack/35.il \
+	tailcall/interface-conservestack/36.il \
+	tailcall/interface-conservestack/37.il \
+	tailcall/interface-conservestack/38.il \
+	tailcall/interface-conservestack/39.il \
+	tailcall/interface-conservestack/4.il \
+	tailcall/interface-conservestack/40.il \
+	tailcall/interface-conservestack/41.il \
+	tailcall/interface-conservestack/42.il \
+	tailcall/interface-conservestack/43.il \
+	tailcall/interface-conservestack/44.il \
+	tailcall/interface-conservestack/45.il \
+	tailcall/interface-conservestack/46.il \
+	tailcall/interface-conservestack/47.il \
+	tailcall/interface-conservestack/48.il \
+	tailcall/interface-conservestack/49.il \
+	tailcall/interface-conservestack/5.il \
+	tailcall/interface-conservestack/50.il \
+	tailcall/interface-conservestack/51.il \
+	tailcall/interface-conservestack/52.il \
+	tailcall/interface-conservestack/53.il \
+	tailcall/interface-conservestack/6.il \
+	tailcall/interface-conservestack/7.il \
+	tailcall/interface-conservestack/8.il \
+	tailcall/interface-conservestack/9.il
+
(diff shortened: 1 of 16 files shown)


jaykrell:
Very relevant failures. So far not reproing locally but still trying.

jaykrell:
Let's split up the F# tests here, and then fine tune the enable/disable, and then drill into the disabled. The failures are gsharedvt related, at least, but that doesn't make sense. The types should be statically discovered and instantiated and gsharedvt not used. Investigate more..

jaykrell:
This is working well as intended -- splitting up the tests and then most pass, a few fail, some failures are understood (gsharedvt), some merit more investigation (x86). Still waiting on the pass/fail lists for arm..

mono/tests/Makefile.am
```diff
@@ -912,7 +1009,106 @@ PLATFORM_DISABLED_TESTS += tailcall/3.exe
 PLATFORM_DISABLED_TESTS += tailcall/4.exe
 PLATFORM_DISABLED_TESTS += tailcall/8273.exe
 PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail.exe
+PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail-Seq.filter-length1.exe
+PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail-Seq.filter-length2.exe
```
vargaz: Could this be written in fewer lines ? The other places too.

vargaz:
Also, would it be possible to move the decomposed tests into a subdir ? There are like a 100 files.

mono/tests/Makefile.am
jaykrell: It is this way for easy add/delete/merge. Adding a test is adding a line, removing a test is removing a line. No backslashes or word wrap to contend with.

jaykrell:
Separate dir -- yes. I prefer that but much earlier @luhenry didn't seem to (he didn't really want the coreclr stuff as it is, but it was easier this way).
Tailcall/ or tailcall/x and tailcall/y or don't care as long as at least something/?
Probably just tailcall/fsharp-deeptail and tailcall/interface-conservestack or somesuch.

The "deeptail" and "conservestack" mean, like, vs. just at least running correctly.
deeptail means stack overflow for failure, though I go back and forth on recursion depth.
Some tests can detect stack consumption, some cannot. Perhaps not the best naming.
There were regressions of correctness at some point, which is why some tests run "just for correctness" w/o caring about stack use.

mono/tests/Makefile.am
jaykrell: The number of lines here should go down as well. I think I have commented out disable all for ARM, and disable all for PowerPC. Those should decrease. Oh and disable all for interpreter, granted. The *.il existing style is one per line, but with merge conflicts at the end due to lack of trailing slash. I believe ease of edit/merge is valuable, but it does conflict with reading quickly, glossing over stuff.

Maybe we can if-in the .il lines? That'd be a net reduction for certain.
We also maybe use include to shrink the file. But I'm not certain. At least the disables should reduce.

jaykrell:
I've moved everything into subdirs tailcall/fsharp-deeptail/ and tailcall/interface-conservestack.
Reset all of the disabled lists, except what failed today. So that'll be smaller and hopefully not fully grow back. Will look at the x86 breaks.
Waiting for initial arm/arm64 lists from CI.
Will CI PowerPC also on next round.

jaykrell:
x86 failures:
One is because called function does take more parameters than caller, which we don't support.

Some or all of the rest are related to frame alignment. The test checks for zero stack used, but 8 bytes were being used. I adjusted the tests -- take address of parameter instead of local, that helps some, maybe all, we'll see. The test also prints stack used in error message now.

The FullAOT failures are all believed to be due to gsharedvt.

Still waiting on ARM. :(

jaykrell:
I also centralized the debugprinting.

jaykrell:
Still waiting for ARM. Probably will have to redisable a very small number of tests due to extra_arg.

mono/tests/split-tailcall-interface-conservestack.cpp
```diff
@@ -0,0 +1,66 @@
+/*
```
luhenry: Remove this file.

mono/tests/tailcall/split-fsharp.cpp
```diff
@@ -0,0 +1,151 @@
+/*
```
luhenry: Remove this file.

mono/tests/split-tailcall-interface-conservestack.cpp
jaykrell: This file is critical to maintenance of the tests. I have used it repeatedly as I edit the tests and it saved me a ton of time.

mono/tests/tailcall/split-fsharp.cpp
jaykrell: This file is critical to maintenance of the tests. I have used it repeatedly as I edit the tests and it saved me a ton of time. I tried writing in Perl but gave up.

jaykrell:
x86 passing with the test tweaks.
Overnight not long enough to run arm ci.

jaykrell:
A few of the tests had the same names, so the last duplicate overwrote the others.
That is fixed locally, but it'd be good to get ARM results before resetting CI for another 18+ hours.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Looking good now. Review?

mono/tests/tailcall/split-fsharp.cpp
vargaz: Its imho we already have test generating files like tests/gen-runtime-invoke.cs.

jaykrell:
@monojenkins build failed

jaykrell:
I agree commiting derived files is generally bad. I’ll have to work the makefile & ensure the C++ works on centos6, or skip the tests there. And make sure it works out-of-tree. 

Some of the splitting is maybe redundant now, like all the interface tests maybe fail on arm but pass on the rest.

jaykrell:
Shouldn’t be too difficult, I can try.

luhenry:
Write the c++ in c#, that’s available everywhere we need it to be :wink:

jaykrell:
Makefile not working yet.

jaykrell:
@monojenkins Linux  WebAssembly

jaykrell:

jaykrell:

jaykrell:

jaykrell:

mono/tests/tailcall/split-fsharp.cpp
```diff
@@ -0,0 +1,174 @@
...
+		suffix.push_back (line);
+
+	for (tests_t::const_iterator t = tests.begin(); t != tests.end(); ++t)
+	{
+		if (t->name.length() == 0)
+			continue;
+		//printf("%s\n", t->name.c_str());
+		ofstream output(("tailcall/fsharp-deeptail/" + t->name + ".il").c_str());
+		for (strings_t::const_iterator a = prefix.begin(); a != prefix.end(); ++a)
+			output << a->c_str () << endl;
```
jaykrell: c_str() not likely needed here

lewurm:
Sorry for the strong language, but why the heck do we have C++ files now?!  This introduces another dependency that we don't want (we sort of need it for LLVM, but this is optional). This was sneakily introduced after the initial PR approval. For the future, if you do larger changes on your PR _after_ it was approved, please dismiss the review and ask for review again.

Can you please rewrite those generators in C# as suggested in https://github.com/mono/mono/pull/9316#issuecomment-402266442 ?

vargaz:
The file is not part of the build, its just used to generate the original versions of tests, so its not a big problem.

jaykrell:
It is used now.  @luhenry kinda suggested that -- if keeping file, then use it. And then the rewrite. I took only the first suggestion.

PR #9334, opened 2018-06-27
Fix a couple of failing DYNCALL scenarios on Windows x64.

lateralusX:
This PR fixes a couple of issues preventing DYNCALL to work in a couple of scenarios on Windows x64.

Fix incorrect float parameter handling on Winx64 DYNCALL:

The Winx64 calling convention handles parameters in a way that was not correctly implemented in some DYNCALL use cases. When using a method signature like this (double, int, double), the Winx64 ABI dictates that the first parameter goes into first floating point parameter register (XMM0), second parameter goes into second general register (RDX) and third floating point goes into third floating point
parameter register (XMM2). So above call should use registers XMM0, RDX and XMM2, but current DYNCALL implementation will just use next floating-point register, like XMM0, RDX and XMM1 and this will load wrong parameter registers as expected by callee.

For an example https://msdn.microsoft.com/en-us/library/zthk2dkh.aspx#Anchor_2

Fix makes sure mono_arch_start_dyn_call uses the already evaluated registry information for floating point parameters. Currently it assumes floating point registers are assigned in a linear fashion and that works for System V ABI, but not for Winx64’s ABI.

Commit also adds a couple of tests into aot-tests.cs that works with mixed signatures
together with DYNCALL.

Fix DYNCALL value type parameters to follow Windows x64 ABI:

Correctly handle use cases when parameter storage is ArgValuetypeAddInIReg or ArgValuetypeAddrOnStack. This is used for value types not fitting into registers on Windows x64.

Enable aot-tests.exe mini regression tests previously excluded using !FULLAOT-AMD64:

After fixes issues in DYNCALL Windows x64 ABI all aot-tests.exe previously excluded using !FULLAOT-AMD64 can now be reenabled since outstanding issues have been fixed. There is one new test still failing on Windows x64, test_42_arm64_dyncall_vtypebyval. Turns out that this was caused by problems to pass float/double parameters using DYNCALL when spilling over to stack (not using registers). There was also a problem with ArgValuetypeAddrOnStack on Windows x64 when spilling over to the stack that is now also fixed.

Added more tests into aot-tests.cs for the above scenarios as well.

### DIFF
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -2500,6 +2500,8 @@ dyn_call_supported (MonoMethodSignature *sig, CallInfo *cinfo)
 		case ArgInFloatSSEReg:
 		case ArgInDoubleSSEReg:
 		case ArgValuetypeInReg:
+		case ArgValuetypeAddrInIReg:
+		case ArgValuetypeAddrOnStack:
 		case ArgOnStack:
 			break;
 		default:
@@ -2640,8 +2642,11 @@ mono_arch_start_dyn_call (MonoDynCallInfo *info, gpointer **args, guint8 *ret, g
 		ArgInfo *ainfo = &dinfo->cinfo->args [pindex + sig->hasthis];
 		int slot;
 
-		if (ainfo->storage == ArgOnStack) {
+		if (ainfo->storage == ArgOnStack || ainfo->storage == ArgValuetypeAddrOnStack) {
 			slot = PARAM_REGS + (ainfo->offset / sizeof (mgreg_t));
+		} else if (ainfo->storage == ArgValuetypeAddrInIReg) {
+			g_assert (ainfo->pair_storage [0] == ArgInIReg && ainfo->pair_storage [1] == ArgNone);
+			slot = general_param_reg_to_index [ainfo->pair_regs [0]];
 		} else if (ainfo->storage == ArgInFloatSSEReg || ainfo->storage == ArgInDoubleSSEReg) {
 			slot = float_param_reg_to_index [ainfo->reg];
 		} else {
@@ -2752,6 +2757,12 @@ mono_arch_start_dyn_call (MonoDynCallInfo *info, gpointer **args, guint8 *ret, g
 						break;
 					}
 				}
+				break;
+			case ArgValuetypeAddrInIReg:
+			case ArgValuetypeAddrOnStack:
+				// In DYNCALL use case value types are already copied when included in parameter array.
+				// Currently no need to make an extra temporary value type on stack for this use case.
+				p->regs [slot] = (mgreg_t)arg;
 				break;
 			case ArgOnStack:
 				for (i = 0; i < ainfo->arg_size / 8; ++i)
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -2544,7 +2544,8 @@ mono_arch_dyn_call_prepare (MonoMethodSignature *sig)
 		ArgInfo *ainfo = &cinfo->args [i];
 		switch (ainfo->storage) {
 		case ArgOnStack:
-			info->nstack_args = MAX (info->nstack_args, ainfo->offset + (ainfo->arg_size / 8));
+		case ArgValuetypeAddrOnStack:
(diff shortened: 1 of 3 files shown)


mono/mini/mini-amd64.c
```diff
@@ -2600,15 +2602,19 @@ mono_arch_start_dyn_call (MonoDynCallInfo *info, gpointer **args, guint8 *ret, g
...
+
 	static gboolean param_reg_to_index_inited;
 
 	if (!param_reg_to_index_inited) {
 		for (i = 0; i < PARAM_REGS; ++i)
-			param_reg_to_index [param_regs [i]] = i;
+			general_param_reg_to_index [param_regs[i]] = i;
+		for (i = 0; i < FLOAT_PARAM_REGS; ++i)
+			float_param_reg_to_index [float_param_regs[i]] = i;
 		mono_memory_barrier ();
```
jaykrell: I recently learned, that this is not actually thread-safe.
The rule of thumb, is that one barrier is seldom/never enough. They come in pairs -- on the reader and the writer.
At least for ARM. So it'll only matter here, I guess, for an ARM[64]-hosted AMD64-targeted compiler which will be rare. I was surprised and still have trouble getting my head around it.

Can this data therefore be const/statically initialized?

lewurm: there is `mono_mini_arch_lock ()` available in this file

jaykrell: In particular, having reread the compiler developer's notes on a similar matter,
you need a barrier between reading param_reg_to_index_inited, and reading what it guards.

Like this:

```
if (!param_reg_to_index_inited) {
		for (i = 0; i < PARAM_REGS; ++i)
			general_param_reg_to_index [param_regs[i]] = i;
		for (i = 0; i < FLOAT_PARAM_REGS; ++i)
			float_param_reg_to_index [float_param_regs[i]] = i;
		mono_memory_barrier ();
		param_reg_to_index_inited = 1;
	}
else {
		mono_memory_barrier (); // << add this
}
```

Otherwise you can see uninitialized data.

We should probably use a "once" -- which indeed the Win32/arm version
has a barrier every time.

C++ initialized static locals now offer the best functionality: terse, thread safe, no barrier.
They do use a thread local under the covers, which is a bit unfortunate.

lateralusX: Jepp, this was old code behaving like this so I didn't change it (added two years ago by this commit https://github.com/mono/mono/commit/a58c2ef804779be70832c6b0e092fbca1edaebf1). C++ has acquire/release semantics together with sequential consistent memory order to correctly order load/stores on different architectures and since C++11 made static locals thread safe, this would have been fine, on C++11, but now we have C and this is code running on AMD64 targets (part of the runtime and not the compiler), but in order to close this theoretical initialization race I can extend the code with the proposed memory barrier.

jaykrell: C11 atomics should be easy to use and super efficient here (on non-arm at least).

jaykrell: Maybe not with Visual C++ though?

lateralusX: We have mono atomic's implementation that could be an option instead of barriers. Moving to static initialization could also be an option, but I think original author didn't want to manual maintain that mapping in the first place (but it should be static per ABI and should not change once set). I could give it a try as an alternative, but first I need to add support for stack parameters in DYNCALL as part of this PR (that why it's still work in progress).

jaykrell: Mono atomics are probably not the same thing or as efficient, as C11 atomics (or C++ atomics<>)
I understand the desire to generate the data, the code might appear smaller than the generated data, but race conditions aren't worthwhile.
Sometimes I write a program to generate the data (printf the source) and check that in.
I wonder if we are misusing barriers elsewhere in the runtime.
Work in progress understood.

lateralusX: @vargaz Any opinions related to the above discussion?

vargaz: The code in https://github.com/mono/mono/pull/9334#discussion_r198606590 looks correct to me.

lateralusX: Added memory barrier as outline above.

lateralusX:
@vargaz Still OK with the fix in last commit as well (added after your approval of the PR)?

PR #9346, opened 2018-06-27
[sdb] Add a AssemblyMirror.GetMetadata () method to read the metadata from the debuggee, make AssemblyMirror.Metadata return it, or null if its not set/read yet.

vargaz:
Fixes a regression introduced by b26a41cc3ee5fbee01430b8e4f31e687ef90b7db. Parts of Mono.Debugger.Soft would check
AssemblyMirror.Metadata against null, and that would throw an exception when the debuggee was using an older protocol version.


vargaz:
Another solution would be to check whenever metadata_blob is set in AssemblyMirror.Metadata, so there is no need to add an additional method.

sailro:
Adding more context here: after testing M.D.S. against Unity (with protocol version support up to 2.46), when accessing custom attributes the code is indirectly using the Metadata properties.

So given the code used unconditionally a method requiring 2.47, it failed (CheckProtocolVersion ), and the receiver thread died.

Then we were stuck afterwards waiting for request responses.

This PR fixes the issue we hit.

marek-safar:
@vargaz this breaks debugger tests

akoeplinger:
@monojenkins commit apidiff

PR #9347, opened 2018-06-27
[coop] mono_coop_cond_signal should switch to GC Safe thread state

lambdageek:
1. Switch to GC Safe thread state in `mono_coop_cond_signal` and `mono_coop_cond_broadcast` because at least on Linux/glibc the underlying implementation can block on a futex (specifically in `__condvar_quiesce_and_switch_g1` in `glibc/ntpl/pthread_cond_common.c`).  If you'd like details:
   * glibc NTPL description of the implementation of pthread condition variable signaling [(pthread_cond_wait.c:193 on sourceware.org)](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_cond_wait.c;h=3e1105418210288ef21c37730a59dc8bc619ab3b;hb=HEAD#l193) 
   * glibc NTPL implementation of `__condvar_quiesce_and_switch_g1` [(pthread_cond_common.c:341 on sourceware.org)](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_cond_common.c;h=8e425eb01eceabec64e2bad3cfeb2ec51b6f6d72;hb=HEAD#l341) it can block at [(pthread_cond_common.c:412 on sourceware.org)](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_cond_common.c;h=8e425eb01eceabec64e2bad3cfeb2ec51b6f6d72;hb=HEAD#l412)
2. make `mono_pmip` work again under coop and hybrid suspend inside GDB/LLDB (it used to call `mono_method_full_name` which would try to switch GC Unsafe thread state and then assert if the developer unwittingly called it from a thread that isn't currently attached).

### DIFF
--- a/mono/utils/mono-coop-mutex.h
+++ b/mono/utils/mono-coop-mutex.h
@@ -109,13 +109,27 @@ mono_coop_cond_timedwait (MonoCoopCond *cond, MonoCoopMutex *mutex, guint32 time
 static inline void
 mono_coop_cond_signal (MonoCoopCond *cond)
 {
+	/*
+	 * On glibc using NTPL (ie Linux with an underlying futex), signaling a
+	 * condition variable can block in the __condvar_quiesce_and_switch_g1
+	 * operation. So switch to GC Safe mode here.
+	 */
+	MONO_ENTER_GC_SAFE;
 	mono_os_cond_signal (&cond->c);
+	MONO_EXIT_GC_SAFE;
 }
 
 static inline void
 mono_coop_cond_broadcast (MonoCoopCond *cond)
 {
+	/*
+	 * On glibc using NTPL (ie Linux with an underlying futex), signaling a
+	 * condition variable can block in the __condvar_quiesce_and_switch_g1
+	 * operation. So switch to GC Safe mode here.
+	 */
+	MONO_ENTER_GC_SAFE;
 	mono_os_cond_broadcast (&cond->c);
+	MONO_EXIT_GC_SAFE;
 }
 
 G_END_DECLS
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -2144,13 +2144,13 @@ ves_icall_System_Threading_WaitHandle_SignalAndWait_Internal (gpointer toSignal,
 
 	mono_thread_set_state (thread, ThreadState_WaitSleepJoin);
 	
-	MONO_ENTER_GC_SAFE;
 #ifdef HOST_WIN32
+	MONO_ENTER_GC_SAFE;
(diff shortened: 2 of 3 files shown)


jaykrell:
Does this fix Alpine and is it a better fix for the thread6/7 hang at shutdown? (see f3319f793562ba245859bc9d3a5f8d4c0f9f0953)

mono/utils/mono-coop-mutex.h
```diff
@@ -109,13 +109,27 @@ mono_coop_cond_timedwait (MonoCoopCond *cond, MonoCoopMutex *mutex, guint32 time
 static inline void
 mono_coop_cond_signal (MonoCoopCond *cond)
 {
+	/*
+	 * On glibc using NTPL (ie Linux with an underlying futex), signaling a
+	 * condition variable can block in the __condvar_quiesce_and_switch_g1
+	 * operation.  So switch to GC Safe mode here.
```
jaykrell: Modern style is one space after periods.

jaykrell: And this seems reasonable even with less commentary -- "syscalls can block" -- klnda just a blanket/blind good assumption. But ok.

lewurm:
> it used to call mono_method_full_name which would try to switch GC Unsafe thread state 

I call `mono_method_full_name` quite often during debugging sessions. Does that mean with COOP I shouldn't use it (or only carefully)? Does that apply for every function annotated with `MONO_API`?

lambdageek:
@lewurm As long as you're calling it from a thread that's attached to the runtime, it's fine.  I accidentally called it from some random thread - a GC worker or something libtest.c created.

mono/utils/mono-coop-mutex.h
jaykrell: While it isn't likely, it could be they haven't been called in a while, are paged out, and just calling them requires paging them in. Just crossing from one page of code to another, without even a call, can page fault and be slow. But we can't do much about that, except maybe page-align functions and keep them under page in size, limiting the fault to their start.

jaykrell:
Oh I guess nevermind, it's a sem_timedwait here:

https://github.com/mono/mono/commit/f3319f793562ba245859bc9d3a5f8d4c0f9f0953#diff-ce0350ac89da817c904c38a837fc7b23L391

This still needs attention but it isn't related to your PR.

Maybe the fix is around here, to not skip aborting when abort already requested..
https://github.com/mono/mono/blob/master/mono/metadata/threads.c#L2482

lambdageek:
@monojenkins build failed

Issue #9023, opened 2018-06-04
Two ValueType properties initialization inline fails in mono wasm


jeromelaban:
The following code fails on mono-wasm:

```csharp
    public static class Class1
    {
		static void Main(string[] args)
		{
			Console.WriteLine("Start");

			// This fails
			Item1 = Item2 = MyValueType.Value1;

			// This works
			// Item2 = MyValueType.Value1;
			// Item1 = MyValueType.Value1;

			Console.WriteLine("Done");
		}

		public static MyValueType Item1 { get; set; }
		public static MyValueType Item2 { get; set; }
	}

	public struct MyValueType
	{
		public static readonly MyValueType Value1 = new MyValueType();
	}
``` 

Initializing the two fields in two lines works. If the MyValueType is a class, it works.

### Current Behavior
```
* Assertion at /Users/builder/jenkins/workspace/test-mono-mainline-webassembly/label/highsierra/mono/metadata/class-init.c:4107, condition `klass' not met
mono.js:1 
uno-bootstrap.js:108 abort(). Build with -s ASSERTIONS=1 for more info.
```

### Expected Behavior

The runtime does not assert.

**Version Used**:

https://jenkins.mono-project.com/view/All/job/test-mono-mainline-wasm/319/label=ubuntu-1804-amd64/Azure/

### Complete repro sample
[EnumException.zip](https://github.com/mono/mono/files/2069746/EnumException.zip)

Build and host the dist folder of the bin folder.

jeromelaban:
Here's another scenario that fails similarly: 

```csharp
    public static class Class1
    {
		static void Main(string[] args)
		{
			Console.WriteLine("Start");

			new Test().Run();

			Console.WriteLine("Done");
		}
	}

	public class Test
	{
		private DateTime _timeSampleEditedFirst = DateTime.MinValue;
		private DateTime _timeSampleEditedLast = DateTime.MinValue;

		public void Run()
		{
			if (_timeSampleEditedFirst != DateTime.MinValue &&
				_timeSampleEditedLast != DateTime.MinValue)
			{
				Console.WriteLine($"min value");
			}
			else
			{
				Console.WriteLine($"not min value");
			}

			// works 
			//_timeSampleEditedFirst = DateTime.MinValue;
			//_timeSampleEditedLast = DateTime.MinValue;

			// Fails
			_timeSampleEditedFirst = _timeSampleEditedLast = DateTime.MinValue;
		}
	}
```

lewurm:
[...]

PR #9348, opened 2018-06-27
[interp] fix scope issue regarding `klass` var

lewurm:
`klass` exists in the main loop, but we defined `klass` again inside the
block, so the outer `klass` never had the right value.

Fixes https://github.com/mono/mono/issues/9023


lewurm:

lewurm:

PR #9349, opened 2018-06-27
[corlib]: Remove the old legacy implementation of `X509Certificate.GetName()` and `GetIssuerName()`

baulig:
`X509Certificate.GetName()` and `GetIssuerName()` are `[Obsolete]` since forever and CoreFX switched to just returning
`Subject` / `Issuer` two years ago.

We still have the old .NET 1.x behavior in Mono, but getting rid of it makes us compatible with
CoreFX, making it easier to eventually adapt part of their PAL.

Replace the corresponding `GetSubjectName()` and `GetIssuerName()` methods in `X509CertificateImpl`
with `Subject` and `Issuer` properties, in preparation of `X509CertificateImpl` implementing CoreFX's
`ICertificatePal`.


baulig:
Obsoleted by latest upstream CoreFX changes and "Blue".

PR #9368, opened 2018-06-28
[build] Skips RemoteExecutor bases tests on monodroid

marek-safar:
The proper implementation will be written later

/cc @grendello


marek-safar:
@monojenkins backport 2018-06

marek-safar:
@monojenkins backport 2018-04

akoeplinger:
do we have a tracking issue for enabling the proper support?

marek-safar:
We don't as we don't own the runner either but will check if XA folks made an issue for this or not

PR #9369, opened 2018-06-28
[corlib]: Cleanup `X509Helper` initialization and import.

baulig:
This is currently only used for BTLS.

* `X509Helper.EnsureNativeHelper`: New internal method to ensure the native helper is initialized.

  On Mobile, we call `MonoTlsProviderFactory.Initialize()` (which calls `X509Helper2.Initialize()`) during system startup, so the `nativeHelper` is always initialized.
  
  On Desktop, using either `SslStream` or `X509Certificate2` will initialize it as well.  Add a reflection-based fallback which will call `X509Helper2.Initialize()` in System.dll if necessary.
  
* `X509Helper.InitFromCertificate`: Call `EnsureNativeHelper()` then use `ImportCore()`.

* `X509Helper.InitCore`: Try to use the native helper if possible, so we can avoid the managed PEM -> DER conversion.

* `INativeCertificateHelper.Import()`: Add new overload only taking a `byte[]`.  The implementation will automatically convert PEM to DER using its native backend if possible.  Since BTLS supports both import formats, it will simply pass the correct format type to the native backend.

Since `X509Helper2.Initialize()` is called independently from the choice of TLS Provider (including Legacy TLS), this will allow us to move all the fallback code from corlib into System.

### DIFF
--- a/mcs/class/System/System.Security.Cryptography.X509Certificates/X509Helper2.cs
+++ b/mcs/class/System/System.Security.Cryptography.X509Certificates/X509Helper2.cs
@@ -48,6 +48,7 @@ using Mono.Btls;
 
 using System.IO;
 using System.Text;
+using System.Threading;
 
 namespace System.Security.Cryptography.X509Certificates
 {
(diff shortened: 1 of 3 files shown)


baulig:
build

baulig:
I'm going to make a few more changes to make sure this also works with AppleTls and then test on Mobile.

baulig:
I am going to replace this with a general-purpose dependency injector, so we can reuse it for other purposes like for instance crypto classes and things like `X500DistinguishedName`.

PR #9374, opened 2018-06-29
Combine mono_cominterop_get_com_interface and ves_icall_System_Runtime_InteropServices_Marshal_GetIUnknownForObjectInternal which are nearly identical.

jaykrell:


mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
...
+			return ((MonoComInteropProxy*)real_proxy)->com_object->iunknown;
 		void* com_itf = cominterop_get_interface_checked (((MonoComInteropProxy*)real_proxy)->com_object, ic, error);
 		return com_itf;
 	}
 	else {
+		if (icall)
+			ic = mono_class_get_iunknown_class ();
 		void* ccw_entry = cominterop_get_ccw_checked (object, ic, error);
+		if (icall)
+			mono_error_set_pending_exception (error);
```
lambdageek: I think it's better to have the call to `mono_error_set_pending_exception` in`ves_icall_System_Runtime_InteropServices_Marshal_GetIUnknownForObjectInternal` not in `..._internal`. It will make the icall more like others and will eventually make it easier to convert it to handles.

mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
 			return NULL;
 		klass = mono_object_class (object);
 		if (!mono_class_is_transparent_proxy (klass)) {
+			if (icall)
+				g_assert_not_reached ();
```
lambdageek: I think it's better to `g_assert (mono_class_is_transparent_proxy (klass))` in `..._internal` and to do the dynamic check in `mono_cominterop_get_com_interface`.

mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
...
 			mono_error_set_invalid_operation (error, "Class is not transparent");
 			return NULL;
 		}
 
 		real_proxy = ((MonoTransparentProxy*)object)->rp;
 		if (!real_proxy) {
+			if (icall)
+				g_assert_not_reached ();
```
lambdageek: Same.  Put `g_assert (real_proxy)` in `..._internal` and the dynamic check in `mono_cominterop_get_com_interface`

mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
...
 			mono_error_set_invalid_operation (error, "RealProxy is null");
 			return NULL;
 		}
 
 		klass = mono_object_class (real_proxy);
 		if (klass != mono_class_get_interop_proxy_class ()) {
+			if (icall)
+				g_assert_not_reached ();
```
lambdageek: Same.  Assert in `..._internal` and move the check to the caller.

mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
...
 			mono_error_set_invalid_operation (error, "Object is not a proxy");
 			return NULL;
 		}
 
 		if (!((MonoComInteropProxy*)real_proxy)->com_object) {
+			if (icall)
+				g_assert_not_reached ();
```
lambdageek: Same. Assert here, move check to caller.

mono/metadata/cominterop.c
```diff
@@ -3699,32 +3669,46 @@ mono_cominterop_get_com_interface (MonoObject *object, MonoClass *ic, MonoError
...
 			mono_error_set_invalid_operation (error, "Proxy points to null COM object");
 			return NULL;
 		}
 
+		if (icall)
+			return ((MonoComInteropProxy*)real_proxy)->com_object->iunknown;
```
lambdageek: I don't understand what this is doing.  But I guess that's what the old icall code used to do too.  So ok.

mono/metadata/cominterop.c
jaykrell: I think it should be asis to perfectly preserve the old meaning. Coop conversion will remove these 2 lines either way.

mono/metadata/cominterop.c
jaykrell: I don’t understand this suggestion. I formed it, here and elsewhere, to preserve exact same meaning & maximum reuse and I think it is pretty clear. Nothing is reordered for example. It is possible some subtle semantic changes are ok, like errors instead of not_reached, but I erred totally toward preserving semantics. Error_init could also be shared.

mono/metadata/cominterop.c
jaykrell: mono_error_set_pending_exception could be moved to caller since it is last, and the other branch of the arm cannot cause it to do anything. Though asis is slightly more obviously semantic preserving, and these two lines should go away soon either way.

The other not_reached vs. error cannot be so obviously or "smally" moved as I understand.
For example, we could return an out "invalid_operation string" for caller to not_reached or mono_error_set_invalid_operation. Another characteristic of the current one is an easier to read diff where at least one path has no change, just conditional adds that won't occur.

mono/metadata/cominterop.c
lambdageek: Hm. well fair enough.

What about

```c
g_assert_message (!icall, "real_proxy->com_object can't be NULL in COM icall");
```

or something similar to replace `if (!icall) g_assert_not_reached()` with a slightly more informative crash message.

jaykrell: Agreed. g_assert_not_reached should be banned, along with g_assert(FALSE). Let me give it a try.

jaykrell: g_assert_not_reached has other problems in that Visual C++ will generate bad code sometimes when it is hit, like a switch default (__assume(0) that is). Maybe gcc/clang similar. assert is at least a well defined stop that the compiler does not overoptimize without its own proof, instead of trusting the programmer.

jaykrell:
@monojenkins build Linux x64

jaykrell:
(Rebased to master.)

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

PR #9375, opened 2018-06-29
CoreFX import for ProtectedData type (Windows part)

MaximLipnin:
As part of #7589 

- CoreFX import only for Windows platform (no support for non-Windows platforms, see https://github.com/dotnet/corefx/issues/22510)
- left Mono managed implementation for non-Windows platforms
- removed DataProtectionScope parameter value check (no such a check in CoreFX and NET Framework, see https://github.com/dotnet/corefx/pull/30726) and couple of related unit tests
- added CoreFX xunit tests


filipnavara:
Is there still any point in having NativeDapiProtection.cs?

MaximLipnin:
no, I will delete it, thanks

mcs/class/System.Security/win32_System.Security.dll.sources
```diff
@@ -0,0 +1,11 @@
+#include System.Security.dll.sources
```
marek-safar: Please update also winaot profile

MaximLipnin:
Seems that xunit tests for ProtectedData need to be extracted to some separate test profiles for win32/winaot.

MaximLipnin:
At the moment I don't know what exactly test profile naming should be used.

MaximLipnin:
https://github.com/mono/mono/issues/9395

filipnavara:
At least on macOS the tests that failed show an actual problem with the managed implementation. I don't necessarily think it has to be solved within this PR, but there should probably be an issue tracking it. The path that the functions try to access on macOS is protected and read-only, maybe "/usr/local" instead of "/usr" would work.

marek-safar:
@monojenkins commit csproj

PR #9377, opened 2018-06-29
[2018-04] [interp] fix scope issue regarding `klass` var

lewurm:
`klass` exists in the main loop, but we defined `klass` again inside the
block, so the outer `klass` never had the right value.

Fixes https://github.com/mono/mono/issues/9023

Backport of https://github.com/mono/mono/pull/9348#issuecomment-401305916


lewurm:
@monojenkins squash

lewurm:
@monojenkins build failed

PR #9378, opened 2018-06-29
[interp] remove unused macros

lewurm:
we use different macros to use computed gotos:
https://github.com/mono/mono/blob/431e7e7fd0d962480f7cd48b3224d4fff41b70ee/mono/mini/interp/interp.c#L2487-L2504


lewurm:
@monojenkins squash

PR #9379, opened 2018-06-29
Reduce sensitivity to thread local win32 last error.

jaykrell:
It is too easily trashed, e.g. by mono_thread_info_uninstall_interrupt.

See https://github.com/Unity-Technologies/mono/commit/b6f1ba89f7ab285a0990e696e78798f7cf18dafb
Which this should be an alterate fix for, as discussed in https://github.com/mono/mono/pull/9104.

This is a small change, but ultimately we should be able to thread
gint32 *win32error around enough, such as to eliminate the thread local on Unix.
It doesn't appear to take much. Thread locals, it turns out, are not a good mechanism.
They are sensitive to loss as addressed here. They are slower to access than regular globals
or locals. They do have the advantage of "automatic threading through a callgraph", as long
as you are very careful with their use (https://github.com/Unity-Technologies/mono/commit/b6f1ba89f7ab285a0990e696e78798f7cf18dafb shows the problem).

As well we could/should move the GetLastError calls down one layer on Win32,
from w32file.c to w32file-win32.c.

But this change addresses only ReadFile/WriteFile.

Also remove unnecessary void** vs. void* cast.
Also combine Unix mono_w32file_read and mono_w32file_write.

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -1799,7 +1799,7 @@ shadow_copy_create_ini (const char *shadow, const char *filename)
 	guint16 *u16_ini;
 	gboolean result;
 	guint32 n;
-	HANDLE *handle;
+	HANDLE handle;
 	gchar *full_path;
 
 	dir_name = g_path_get_dirname (shadow);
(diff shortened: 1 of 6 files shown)


jaykrell:
review?

PR #9381, opened 2018-06-29
Add platform customization support for post native crash handler and native crash info.

lateralusX:
Add support for easier platform customization of native crash info output.

Add support to replace implementation of a post native crash handler behavior if needed (used on none desktop Windows platforms at the moment).


mono/mini/mini-exceptions.c
```diff
@@ -3179,6 +3135,73 @@ mono_handle_native_crash (const char *signal, void *ctx, MONO_SIG_HANDLER_INFO_T
...
+#if defined (HOST_ANDROID)
+		exit (-1);
+#else
+		abort ();
+#endif
+	}
+}
+#else
+extern void
+mono_dump_native_crash_info (const char *signal, void *ctx, MONO_SIG_HANDLER_INFO_TYPE *info);
```
kumpera: There's no reason why we can't keep all implementations symmetric and move the unix code to mini-posix.

lateralusX: Will do, doing same thing as done to mini-windows.c, move to mini-posix.c?

kumpera: Yes, we don't need to have an impl or a forward decl in a source file.

lateralusX:
Failing test in Linux x64 FullAOT + LLVM seems unrelated and happens on many PR's. BTW, that failure served as a regression tests that the native crash dump logging is still working as before :)

PR #9385, opened 2018-06-29
[mini] respect additional flags when caching runtime invoke wrappers

lewurm:
depending on `virtual_` and `need_direct_wrapper` we generate slightly
different IL. The interpreter always set `need_direct_wrapper`, but the
JIT doesn't, so this gets problemantic with mixed mode. We need to be
careful when we cache those results.

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -2510,6 +2510,55 @@ runtime_invoke_signature_equal (MonoMethodSignature *sig1, MonoMethodSignature *
 		return mono_metadata_signature_equal (sig1, sig2);
 }
 
+struct _MonoWrapperMethodCacheKey {
+	MonoMethod *method;
+	gboolean virtual_;
+	gboolean need_direct_wrapper;
+};
+
+struct _MonoWrapperSignatureCacheKey {
+	MonoMethodSignature *signature;
+	gboolean virtual_;
+	gboolean need_direct_wrapper;
+};
+
+typedef struct _MonoWrapperMethodCacheKey MonoWrapperMethodCacheKey;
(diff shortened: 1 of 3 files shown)


mono/metadata/marshal.c
```diff
@@ -2510,6 +2510,55 @@ runtime_invoke_signature_equal (MonoMethodSignature *sig1, MonoMethodSignature *
...
+	gboolean need_direct_wrapper;
+};
+
+typedef struct _MonoWrapperMethodCacheKey MonoWrapperMethodCacheKey;
+typedef struct _MonoWrapperSignatureCacheKey MonoWrapperSignatureCacheKey;
+
+static guint
+wrapper_cache_method_key_hash (MonoWrapperMethodCacheKey *key)
+{
+	return mono_aligned_addr_hash (key->method) ^ ((!!key->virtual_) << 18) ^ ((!!key->need_direct_wrapper) << 19);
```
lewurm: does this look okay?

vargaz:
Might be easier to just add another cache.

lewurm:
@vargaz I tried that, but it made the code harder to follow. This might be more verbose, but it's easier to understand imho.

BrzVlad:
@lewurm Why do we cache signatures based on virtual/need_direct_wrapper. Don't these flags matter only for the code generation ?

lewurm:
@BrzVlad they produce different IL:
https://github.com/mono/mono/blob/fe3ac71532e04b68601c840224b118dc28cfac8c/mono/metadata/marshal-ilgen.c#L1441-L1448

BrzVlad:
I'm talking about the signature cache not the wrapper method cache

lewurm:
@monojenkins squash

lewurm:

PR #9389, opened 2018-06-30
Extend "auto tailcall" functionality similar to explicit tail.call.

jaykrell:
Do not merge.

Experimentally extend "auto tailcall" functionality similar to explicit tail.call.

1. But not if function takes address of anything -- very difficult to improve.
2. Not if function has any exception handling -- should be easy to improve.

Tailcall interacts badly with inlining however.
You don't really want to sacrifice either for either.
Inlining is a crucial optimization.

Should obviate previous auto tailcall code, but doesn't yet.

Initial test is the F# test but with tail. prefixes removed -- and it still gets tailcalled, good.

It would also be good to see through some extra IL around calls, like call nop br ret and call stloc locloc ret; written but not working yet.


jaykrell:
This is a sensitive area but hopefully monotonic improvement is possible.

jaykrell:
https://github.com/mono/mono/pull/9620 instead

PR #9397, opened 2018-07-02
[ci] Add -mmacosx-version-min to OSX builds

akoeplinger:
We're targeting 10.7 in the bockbuild package build, let's do it in testing builds too.


akoeplinger:
OSX builds passed, merging.

PR #9400, opened 2018-07-02
autoconf and CI infrastructure for hybrid suspend

lambdageek:
1. If `configure` enables SGen, print the suspend policy (preemptive, cooperative, hybrid) in the summary at the end.
2. Add a way for `configure.ac` to enable hybrid suspend by default based on presets, or arch, etc.  But do not enable it by default on any platform.
3. add a `preemptive-suspend` CI tag that explicitly disables coop and hybrid suspend; explicitly disable hybrid suspend when enabling coop suspend using CI tags.
4. Add `mono_threads_suspend_policy_name()` function that returns a description of the current suspend policy (subject to configure flags and environment variables)
5. Print suspend policy in `mono --version` and MERP error reporting mechanisms.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1104,6 +1104,7 @@ with_wasm_default=no
 
 with_bitcode_default=no
 enable_cooperative_suspend_default=no
+enable_hybrid_suspend_default=no
 
 INVARIANT_AOT_OPTIONS=nimt-trampolines=2000,ntrampolines=10000,nrgctx-fetch-trampolines=256,ngsharedvt-trampolines=4000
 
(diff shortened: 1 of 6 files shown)


configure.ac
```diff
@@ -5681,6 +5697,7 @@ echo "
 	Host:	       $host
 	Target:	       $target
 	GC:	       $gc_msg 
+	$thread_suspend_msg
```
akoeplinger: should we add this to the `gc_msg` instead? it already mentions whether sgen is concurrent, so e.g. something like this:

```
GC:            sgen (concurrent by default, hybrid suspend)
```

If we do that we should also fix up mono_gc_get_description() in sgen-mono.c to include this so it shows up for mono --version.

lambdageek: Well the suspend mechanism is used by more than just SGen: it's also used by the debugger, by the managed suspend API, the thread abort machinery, etc.

Really pretty much the only thing that doesn't use it (AFAIK) is Boehm, when it's enabled, because it used its own version of preemptive suspend.

akoeplinger: Ok :)

lambdageek:
Added suspend policy to `mono --version` and to the MERP output. /cc @alexanderkyte 

Also refactored the logic for deciding on what suspend policy we'll be using so that it's all in one function.  Should be NFC. /cc @luhenry

alexischr:
Thank you for this, I was about to file an issue :) 

Unrelated to this PR, but the environmental variable `MONO_ENABLE_COOP_SUSPEND` is not in the man page, and `MONO_ENABLE_HYBRID_SUSPEND` is in neither the man page or the README.

PR #9401, opened 2018-07-02
[runtime] Create flight recorder logger backend

alexanderkyte:
This backend creates a shared ring buffer library and creates infrastructure to manipulate it in a dedicated thread. The runtime and this dedicated thread communicate over a lockless queue, preventing the need for a global reader/writer lock on the logger.


alexanderkyte:
I have a ton of printfs I would like to leave in until CI tells me that everything is correct. Please nobody merge this until I remove those printfs.

mono/utils/mono-threads.c
```diff
@@ -791,6 +791,82 @@ mono_thread_info_set_flags (MonoThreadInfoFlags flags)
...
+
+	while (TRUE) {
+		gpointer old_read = mono_atomic_cas_ptr ((gpointer *) &init_callbacks, MONO_END_INIT_CB, (gpointer) old);
+		if (old == old_read)
+			break;
+		else
+			old = old_read;
+	}
+	if (old == MONO_END_INIT_CB) {
+		g_warning ("Global threads inited twice");
```
alexrp: We should probably `g_error` here as multiple calls to `mono_thread_info_init` would have pretty disastrous results.

alexanderkyte: Thanks, will fix. And it does. I was doing it accidentally for a while and the crashes were weird TLS variable issues that all looked like totally unrelated things

mono/utils/mono-threads.c
```diff
@@ -1835,3 +1911,181 @@ mono_thread_info_get_tools_data (void)
...
+	MonoUtilityThread *thread = g_malloc0 (sizeof (MonoUtilityThread));
+	size_t entry_size = offsetof (UtilityThreadQueueEntry, payload) + payload_size;
+
+	thread->message_block_size = mono_pagesize ();
+	thread->payload_size = payload_size;
+	thread->callbacks = *callbacks;
+
+	mono_lock_free_queue_init (&thread->work_queue);
+	mono_lock_free_allocator_init_size_class (&thread->message_size_class, entry_size, thread->message_block_size);
+	mono_lock_free_allocator_init_allocator (&thread->message_allocator, &thread->message_size_class, MONO_MEM_ACCOUNT_OTHER);
```
alexrp: The memory accounting type should be a parameter to this function.

alexanderkyte: What other memory type does it make sense to call the memory of a log statement?

mono/utils/mono-threads.c
```diff
@@ -1835,3 +1911,181 @@ mono_thread_info_get_tools_data (void)
...
+	mono_thread_hazardous_try_free (entry, free_queue_entry);
+
+	return TRUE;
+}
+
+static void *
+utility_thread (void *arg)
+{
+	mono_thread_info_wait_inited ();
+	mono_thread_info_attach ();
```
alexrp: For the profiler to use this API, there needs to also be an early init callback that's invoked before `mono_thread_info_attach` is called.

mono/utils/mono-threads.c
```diff
@@ -1835,3 +1911,181 @@ mono_thread_info_get_tools_data (void)
...
+	MonoSemType sem;
+	mono_os_sem_init (&sem, 0);
+
+	UtilityThreadQueueEntry *entry = mono_lock_free_alloc (&thread->message_allocator);
+	gboolean done;
+
+	entry->finished = &done;
+	entry->response_sem = &sem;
+	entry->thread = thread;
+	memcpy (entry->payload, message, thread->payload_size);
```
lambdageek: I'm a little bit confused about the memory management of this utility thread mechanism.

So you send it messages - it will allocate its own copy and then queue that for the utility thread to process.  When the queue processes an entry, it will pass a pointer to the copy that it owns to the `command` callback but then it doesn't free the entry.  And the `handle_command` that you wrote for the flight recorder doesn't free the entry that's passed to it either.

So who is freeing the copy of the message when it's pulled off the queue?

mono/utils/mono-threads.c
```diff
@@ -1835,3 +1911,181 @@ mono_thread_info_get_tools_data (void)
...
+
+	MonoUtilityThread *thread = (MonoUtilityThread *) arg;
+	thread->callbacks.init (&thread->state_ptr);
+
+	while (mono_atomic_load_i32 (&thread->run_thread)) {
+		MONO_ENTER_GC_SAFE;
+		mono_os_sem_timedwait (&thread->work_queue_sem, 1000, MONO_SEM_FLAGS_NONE);
+		MONO_EXIT_GC_SAFE;
+		utility_thread_handle_inbox (thread);
+	}
```
alexrp: The profiler also needs a callback here. Either it has to be a separate callback or the command callback needs a Boolean parameter indicating whether the inbox is being drained on shutdown or if it's just the normal message handling loop.

mono/utils/mono-threads.c
alexanderkyte: I wonder if it would be safe to just move the init call up and to say that you can't use certain primitives in the init function

mono/utils/mono-threads.c
alexanderkyte: The message is passed to the hazard pointer freeing mechanism. The payload blob is just a static memory blob that we memcpy into. There's no dynamic memory allocation outside of the queue, after we allocate our ring buffer's memory array.

mono/utils/mono-threads.c
```diff
@@ -1835,3 +1911,181 @@ mono_thread_info_get_tools_data (void)
...
+	mono_os_sem_post (&thread->work_queue_sem);
+}
+
+void
+mono_utility_thread_send (MonoUtilityThread *thread, gpointer message)
+{
+	int small_id = mono_thread_info_get_small_id ();
+	if (small_id < 0) {
+#if 1
+		fprintf (stderr, "Dropping message send because thread not attached yet\n");
```
alexrp: Shouldn't these be `g_error`s?

mono/utils/mono-threads.c
alexanderkyte: utility_thread_handle_inbox pops the message off of the queue, passes the pointer to the static memory to the flight recorder so it can memcpy out of it, and then frees the message.

mono/utils/mono-threads.h
```diff
@@ -710,4 +719,39 @@ typedef void (*background_job_cb)(void);
...
+	MonoNativeThreadId thread_id;
+
+	MonoLockFreeQueue work_queue;
+	MonoSemType work_queue_sem;
+	gboolean run_thread;
+
+	MonoLockFreeAllocator message_allocator;
+	MonoLockFreeAllocSizeClass message_size_class;
+
+	unsigned int message_block_size;
```
alexrp: These should be `size_t` judging by the rest of the code.

mono/utils/mono-threads.h
```diff
@@ -710,4 +719,39 @@ typedef void (*background_job_cb)(void);
 void mono_threads_schedule_background_job (background_job_cb cb);
 #endif
 
+typedef struct {
```
alexrp: I feel like this whole abstraction should probably be moved to a separate file, say, `mono-utility-thread.h`.

mono/utils/mono-threads.c
alexanderkyte: I was worried about increasing all the data motion, but in a cachegrind/ubuntu benchmark it was much faster than even a single cache miss from pointers so data traffic was kind of a moot point.

mono/utils/mono-threads.h
alexanderkyte: Yeah, I can do that

mono/utils/mono-threads.c
alexrp: If the profiler were to use this abstraction for its internal threads (which it can, with a few minor changes that I pointed out here), it would be passing `MONO_MEM_ACCOUNT_PROFILER`.

mono/utils/mono-threads.c
lambdageek: Ah ok. I missed that the entry is variably-sized.

So if the message owned some allocated memory, the call to send would transfer ownership to the queue, and the `command` callback would be responsible for deallocating.

mono/utils/mono-threads.c
alexrp: The issue is that the profiler needs to do stuff both before and after `mono_thread_info_attach` is called, during utility thread init.

Although, I don't recall if it is safe to call `mono_thread_info_attach` on an already-attached thread, but if it is, I suppose moving the init callback earlier and having the profiler just call `mono_thread_info_attach` as appropriate in its callback would be an OK solution too.

mono/utils/mono-threads.c
alexanderkyte: Yeah, that memory lifetime contract would be between the sender and the command callback.

alexrp:
For the purposes of this PR, feel free to ignore the comments about using the utility thread API in the profiler, btw. Changing the API to fit the profiler a bit better is more of a nice to have and can always be done later down the line.

mono/utils/mono-threads.h
```diff
@@ -710,4 +719,39 @@ typedef void (*background_job_cb)(void);
...
+
+	unsigned int message_block_size;
+	unsigned int payload_size;
+
+	gpointer state_ptr;
+	MonoUtilityThreadCallbacks callbacks;
+} MonoUtilityThread;
+
+MonoUtilityThread *
+mono_utility_thread_launch (size_t payload_size, MonoUtilityThreadCallbacks *callbacks);
```
alexrp: These would also need to be marked `MONO_PROFILER_API` for the profiler to use them.

mono/utils/mono-threads.c
alexanderkyte: g_error will try to go through the logger, so I've tried to avoid using them in this machinery. I can figure something out if people would prefer to see that here.

lewurm: `g_printerr`?  so it works for embedders too.

alexanderkyte:
@alexrp I believe that I addressed all of your requests. Could you review this?

mono/utils/mono-utility-thread.c
```diff
@@ -0,0 +1,192 @@
...
+	if (thread->callbacks.early_init)
+		thread->callbacks.early_init (&thread->state_ptr);
+
+	mono_thread_info_wait_inited ();
+	mono_thread_info_attach ();
+
+	thread->callbacks.init (&thread->state_ptr);
+
+	while (mono_atomic_load_i32 (&thread->run_thread)) {
+		MONO_ENTER_GC_SAFE;
```
lambdageek: I don't think this is necessary.  And actually may not be right under coop or hybrid suspend - the utility threads is created with `mono_native_thread_create`, so it's not initially attached to the runtime.  That means it doesn't have a `MonoThreadInfo *info`, which means that:
1. the GC doesn't know about it, won't stop it
2. thread state transitions functions will assert because `info` will be `NULL`.

There are a couple of options:
1. just get rid of the thread state transition macros here, and keep using the OS semaphors etc.  The downside is that you have to be vigilant not to call anything from the utility thread that may attach it to the runtime.
2. Create it as an internal thread with `mono_thread_create_internal`, like the profiler does, and set `MONO_THREAD_FLAG_DONT_MANAGE` so that it's not paused by the GC, and then use coop concurrency primitives.

alexanderkyte: The utility thread is attached to the runtime here: https://github.com/mono/mono/pull/9401/files#diff-34d834b8d69724f49f21964be1e1258dR61

alexanderkyte:
Requested change made

alexanderkyte:
@monojenkins build failed

mono/utils/mono-flight-recorder.h
```diff
@@ -0,0 +1,66 @@
...
+
+typedef struct {
+	MonoFlightRecorderHeader header;
+	gpointer payload [MONO_ZERO_LEN_ARRAY]; // We have a variably-sized payload
+} MonoFlightRecorderItem;
+
+typedef struct {
+	intptr_t cursor; // Signed, for sentinel value of -1
+	size_t max_count; // Maximum number of items in logger
+	size_t payload_size; // Size of data reserved for logging message
```
lewurm: any reason why you don't use `gint` and `gsize` here?

mono/utils/mono-flight-recorder.c
```diff
@@ -0,0 +1,153 @@
...
+	g_assert (recorder->max_count > 0);
+
+	iter->recorder = recorder;
+	if (recorder->cursor == MONO_FLIGHT_RECORDER_SENTINEL) {
+		iter->lowest_index = MONO_FLIGHT_RECORDER_SENTINEL; 
+		iter->highest_index = MONO_FLIGHT_RECORDER_SENTINEL;
+	} else if (recorder->cursor >= recorder->max_count) {
+		// Ring buffer has wrapped around
+		// So the item *after* the highest index is the lowest index
+		iter->highest_index = (recorder->cursor + 1) % recorder->max_count;
```
lewurm: why `+1`?

mono/utils/mono-flight-recorder.c
```diff
@@ -0,0 +1,153 @@
...
+		iter->lowest_index = (iter->highest_index + 1) % recorder->max_count;
+	} else {
+		iter->lowest_index = 0;
+		iter->highest_index = recorder->cursor + 1;
```
lewurm: why `+1`?

mono/utils/mono-log-flight-recorder.c
```diff
@@ -0,0 +1,225 @@
...
+#if MONO_PRINT_DROPPED_MESSAGES
+		fprintf (stderr, "Dropping message because thread killed:\n\t%s\n", message);
+#endif
+		return;
+	}
+
+	LogQueueEntry entry;
+	entry.command = MONO_FLIGHT_RECORDER_APPEND;
+	g_snprintf ((gchar *) &entry.message.message, MAX_RECORDER_MSG_LEN, message);
+	mono_utility_thread_send (logger_thread, &entry);
```
lewurm: maybe I'm missing something obvious here, but why can't we append the message here into the ring buffer with the lockless stuff you do in the other thread? having a dedicated thread for this feels over-engineered.

mono/utils/mono-flight-recorder.c
alexanderkyte: The cursor points to the last written record.

mono/utils/mono-flight-recorder.c
alexanderkyte: The cursor points to the last written record.

mono/utils/mono-flight-recorder.h
alexanderkyte: I don't see that much usage of gsize in the codebase compared to size_t (693 vs 1235 occurrences in /mono). I didn't think we should be trying to use it. I can make that change though. 

The gint change does make more sense. I used cursor as an actual pointer before, I should have changed this after I squashed my changes. Thanks for catching this.

mono/utils/mono-log-flight-recorder.c
alexanderkyte: What lockless stuff? With the lockless queue? Making a lockless ring buffer with high scalability would be harder, I think. This is fairly fast and keeps things simple by reusing existing components. The ring buffer data structure itself is not lockless.

alexanderkyte: I examined existing algorithms for lockless ring buffers, and found that the amount of memory allocated and freed was too much for us. I didn't want to add some big performance sink, just for a logger.

alexanderkyte: If you consider the design of the profiler, and the necessity of similar threads with lockless queues in order to log into a data structure without slowing down mono, I think that this design looks a lot less radical. I went with a solution we've proven works elsewhere in the runtime, using the same APIs as that other implementation.

alexanderkyte:
So I can run the webassembly CI script on my mac and my linux server.

alexanderkyte:
The failure seen here doesn't reproduce for me.

jaykrell:
Imho the glib types are pointless.
Even the GNOME develpers seem to think so.

We should imho use

size_t unsigned pointer-sized integer
	This is only smaller than pointer on systems where pointer can be truncated to it.

ptrdiff_t signed pointer-sized integer
	This is only smaller than pointer on systems where pointer can be truncated to it.

int 32bit signed integer
 This is only 16bits on obsolete systems.

typedef unsigned uint or uint_t: 32bit unsigned integer
 This is only 16bits on obsolete systems.

long: useless in portable code, or equivalent to ptrdiff_t on Unix
unsigned long: ditto (size_t)

Issue #9407, opened 2018-07-02
Hybrid suspend deadlock taking the marshal mutex


lambdageek:
Seen on Hybrid suspend CI https://jenkins.mono-project.com/job/test-mono-pull-request-hybrid-suspend/2170/parsed_console/log.html

```
  Id   Target Id         Frame 
* 1    Thread 0x7f88e70fd740 (LWP 59080) "mono" 0x00007f88e60383d6 in sigsuspend () from /lib/x86_64-linux-gnu/libc.so.6
  2    Thread 0x7f88e5bff700 (LWP 59127) "SGen worker" 0x00007f88e65cbb3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
  3    Thread 0x7f88e33e1700 (LWP 59136) "Finalizer" 0x00007f88e65ca536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
  4    Thread 0x7f88e1d4b700 (LWP 59384) "mono" 0x00007f88e60383d6 in sigsuspend () from /lib/x86_64-linux-gnu/libc.so.6
  5    Thread 0x7f88e1b4a700 (LWP 59385) "Thread Pool Wor" 0x00007f88e65caf5c in __lll_lock_wait () from /lib/x86_64-linux-gnu/libpthread.so.0
  6    Thread 0x7f88e1949700 (LWP 59386) "Thread Pool Wor" 0x00007f88e60383d6 in sigsuspend () from /lib/x86_64-linux-gnu/libc.so.6
  7    Thread 0x7f88e1744700 (LWP 62887) "Timer-Scheduler" 0x00007f88e60383d6 in sigsuspend () from /lib/x86_64-linux-gnu/libc.so.6
  8    Thread 0x7f88cbd6d700 (LWP 62888) "Domain unloader" 0x00007f88e65ca536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0

Thread 8 (Thread 0x7f88cbd6d700 (LWP 62888)):
#0  0x00007f88e65ca536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007f88e65ca5e4 in __new_sem_wait_slow.constprop.0 () from /lib/x86_64-linux-gnu/libpthread.so.0
[...]

PR #9408, opened 2018-07-02
[marshal] Use a MonoCoopMutex for marshal_mutex

lambdageek:
Fixes https://github.com/mono/mono/issues/9407

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -76,9 +76,9 @@ enum {
  * The marshal lock is a non-recursive complex lock that sits below the domain lock in the
  * runtime locking latice. Which means it can take simple locks suck as the image lock.
  */
-#define mono_marshal_lock() mono_locks_os_acquire (&marshal_mutex, MarshalLock)
-#define mono_marshal_unlock() mono_locks_os_release (&marshal_mutex, MarshalLock)
-static mono_mutex_t marshal_mutex;
+#define mono_marshal_lock() mono_locks_coop_acquire (&marshal_mutex, MarshalLock)
+#define mono_marshal_unlock() mono_locks_coop_release (&marshal_mutex, MarshalLock)
+static MonoCoopMutex marshal_mutex;
 static gboolean marshal_mutex_initialized;
 
 static MonoNativeTlsKey last_error_tls_id;
@@ -217,7 +217,7 @@ mono_marshal_init (void)
 
 	if (!module_initialized) {
 		module_initialized = TRUE;
-		mono_os_mutex_init_recursive (&marshal_mutex);
+		mono_coop_mutex_init_recursive (&marshal_mutex);
 		marshal_mutex_initialized = TRUE;
 
 		register_icall (ves_icall_System_Threading_Thread_ResetAbort, "ves_icall_System_Threading_Thread_ResetAbort", "void", TRUE);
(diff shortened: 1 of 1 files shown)


vargaz:
Why is this needed ? The marshal lock is relatively low level lock used to protect hash table accesses etc, how can it block ?

lambdageek:
@vargaz There's a stack trace of a deadlock in #9407.  Under hybrid suspend, GC s-t-w hangs if Thread A locks the mutex then steps into GC Safe code and is preemptively suspended while Thread B thread is in GC Unsafe (cooperatively suspended) code and also tries to lock and blocks.

If the hash table is protected by an OS mutex and the code does a `g_hashtable_foreach` and the callback is something that runs in GC Unsafe mode, that's pretty much guaranteed to be a deadlock under hybrid suspend.

OS mutexes should only be used if it might be taken by threads that aren't attached, or we're doing explicit GC Safe/Unsafe transitions by hand, or we can guarantee that there will be no GC Unsafe->GC Safe transitions while the mutex is locked.  Everywhere else it should be a coop mutex.

vargaz:
Coop mutexes are kinda heavyweight, so we should avoid them if possible. In this case, the problem is that the code calls mono_method_signature () which is an api function so it enters unsafe mode, it should call some internal version which doesn't do that.

lambdageek:
I think for this specific mutex we can fix it by pulling the call `mono_method_signature` so we get the signature before taking the `marshal_lock` in `mono_marshal_free_dynamic_wrappers`.

But I don't think your suggestion about `mono_method_signature` is right.   The runtime is mostly in GC Unsafe mode. The fact that `mono_method_signature` is a wrapper that does a Unsafe->Unsafe transition is mostly irrelevant - that just happens to be the first place where we do a thread state poll and coop-suspend the thread.  The implementation of `mono_method_signature_checked` calls a ton of other methods that can poll and coop-suspend the thread.

We should make `mono_threads_enter_gc_safe_region_unbalanced_with_info` cheaper on non-bitcode platforms, but I still think most mutexes in the runtime should be coop.

But this kind of thing is going to come up again. All non-coop locks are suspect.

vargaz:
Alternatively, the problem is that the code which calls it is in gc safe mode, probably the aot runtime code should  transition to gc unsafe mode at the beginning.

lambdageek:
@vargaz 
1. Yea I need to look a bit closer at that AOT runtime code
2. But I think this problem would still show up if all threads were in GC Unsafe.  The issue really is that if you're Thread A holding an OS mutex and you call anything that may checkpoint you will be suspended while holding a mutex.  If any other Thread B, also in Unsafe, tries to take the same mutex it will block and prevent suspend from finishing.  So Thread A is already suspend and is holding the mutex.  Thread B is waiting for the mutex.  And the Suspend Initiator is waiting for B to suspend.  B will never  wake up.  The suspend initiator will never wake A so the mutex will never be unlocked.  Everyone is stuck.
3. For curiosity's sake, I just tried auditing `marshal.c` to see what it would take to keep the marshal mutex using an OS mutex, and it will require changes to `mono_ftnptr_to_delegate_handle`, `delegate_hash_table_add`, `get_cache` (because it's called with `mono_metadata_signature_equal` as `equal_func`), `lookup_string_ctor_signature`, `mono_marshal_get_runtime_invoke_dynamic`, `mono_marshal_get_synchronized_wrapper`, `mono_marhsal_free_dynamic_wrappers` because they all call back into the runtime with the mutex locked.  Some of them would require pretty substantial reorganization.

PR #9411, opened 2018-07-03
[sdb] Avoid calling vm.GetThread () in the event callback, it could c…

vargaz:
…all into the debuggee leading to a deadlock if the thread mirror object has not been created yet.


lewurm:
@monojenkins squash

vargaz:
@monojenkins build failed

PR #9412, opened 2018-07-03
[sdks] Restore the meaning of the IGNORE_PACKAGE_LLVM option.

vargaz:


lewurm:
@monojenkins build failed

PR #9419, opened 2018-07-04
[runtime] Map memory used by code managers with MAP_JIT on osx.

vargaz:


vargaz:
Fixes https://github.com/mono/mono/issues/9210.

akoeplinger:
Looks like it breaks on OSX.

mono/utils/mono-mmap.c
```diff
@@ -225,8 +249,11 @@ mono_valloc (void *addr, size_t length, int flags, MonoMemAccountType type)
 	if (flags & MONO_MMAP_32BIT)
 		mflags |= MAP_32BIT;
 #if defined(__APPLE__) && defined(MAP_JIT)
-	if (flags & MONO_MMAP_JIT)
-		mflags |= MAP_JIT;
+	if (flags & MONO_MMAP_JIT) {
+		if (get_darwin_version () >= 18) {
```
lewurm: can you define a constant for 18?

vargaz: DARWIN_18 ?

mono/utils/mono-mmap.c
```diff
@@ -197,6 +200,30 @@ prot_from_flags (int flags)
...
+
+#define DARWIN_VERSION_MOJAVE 18
+
+static guint32
+get_darwin_version (void)
+{
+	static guint32 version;
+	static gboolean res_set;
+
+	if (!res_set) {
```
jaykrell: I realize this feature is yet x86-only, but this does not look thread safe on ARM.
Please consider pthread_once for initialization instead of rolling our own, which is surprisingly non trivial to get correct.

vargaz: This is actually thread safe, we are storing to 'version' which is a pointer and pointer sized writes are atomic even on arm.

jaykrell: The specific missing thread safety is that barriers typically must come in pairs to have effect. The reader of res_set == TRUE, can still see uninitialized version.

jaykrell: The atomicity of the write does not make this thread safe. The reader can see uninitialized data. The reader can see version==0 despite res_set==TRUE.

jaykrell: The problem is that version can be read before res_set. The compiler and/or processor can reorder the reads.

The barrier on the write does not inhibit reordering of the reads.

One solution is a barrier between reading res_set and reading version.
Another solution is let pthread_once handle everything.
Another solution is probably to merge res_seat and version into one variable, i.e. "version 0" or (-1 or some other value) be the uninitialized value, instead of having the separate variable.

jaykrell: So this doesn't seem thread safe on x86 either.

vargaz:
@monojenkins build failed

mono/utils/mono-mmap.c
```diff
@@ -197,6 +200,29 @@ prot_from_flags (int flags)
...
+	/* This doesn't need locking */
+	if (!version) {
+      char str[256] = {0};
+      size_t size = sizeof(str);
+      int err = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
+	  g_assert (err == 0);
+	  err = sscanf (str, "%d", &version);
+	  g_assert (err == 1);
+	  g_assert (version > 0);
+	}
```
lewurm: sorry to be this guy, but the indent is off in this function 🙂

vargaz:
@monojenkins build failed

PR #9427, opened 2018-07-04
[ios sdk] add interp-only support

lewurm:
depends on https://github.com/mono/mono/pull/9295

should be backported to `2018-06`

### DIFF
--- a/sdks/ios/README.md
+++ b/sdks/ios/README.md
@@ -68,3 +68,7 @@ seem to read the device logs, only the graphical Console app can.
 * Use a publish-subscribe pattern by uploading test results to some cloud service like
 Azure EventHub. This only requires client side internet access on the device and
 the test harness.
+
+## Notes
+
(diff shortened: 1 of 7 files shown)


sdks/ios/README.md
```diff
@@ -68,3 +68,7 @@ seem to read the device logs, only the graphical Console app can.
 * Use a publish-subscribe pattern by uploading test results to some cloud service like
 Azure EventHub. This only requires client side internet access on the device and
 the test harness.
+
+## Notes
+
+Bernhard needed this patch to make it work on his machine: https://gist.github.com/lewurm/7711187deadb57bae97a4d4be4df563b
```
vargaz: This should probably not be part of the README.

vargaz: Then it should cleaned up and included here, not referring to a gist somewhere.

lewurm: We should probably generate `sdks/ios/runtime/Entitlements.xcent`, but let's do that in another PR.

lewurm:
@monojenkins squash

PR #8626, opened 2018-05-08
[corlib] CoreFX sources for System namespace

EgorBo:
Brings CoreFX and CoreRT sources for primitive types inside corlib (and other types under System namespace)

*ADS:* I use this tool to inspect all files I replace https://github.com/EgorBo/MonoMergeHelper 🙂 

Part of #7737

PR #9429, opened 2018-07-05
Fix invalid preprocessor check for BE that broke System.Decimal

NattyNarwhal:
This bug has been lurking for four years, being introduced in commit
5b2bb8c9017ab563d7ec819cf249b209c770423e. `BYTE_ORDER != G_BYTE_ORDER`.

Not thoroughly tested, but `System.Decimal` from csharp seems to return
sane results again. (related: is z or AIX CI available for PR checks?
if not, what test suite is `System.Decimal` under?)

Also remove my shoddy workaround from a few months ago, it got the
BCL building, but was incorrect.

cc @nealef (Linux/s390x might have been affected, but maybe not, if glib and glibc had congruent `BYTE_ORDER` and `G_BYTE_ORDER`....)

### DIFF
--- a/mono/metadata/decimal-ms.c
+++ b/mono/metadata/decimal-ms.c
@@ -84,7 +84,7 @@ static const uint32_t ten_to_ten_div_4 = 2500000000U;
 typedef union {
 	uint64_t int64;
 	struct {
-#if BYTE_ORDER == G_BIG_ENDIAN
+#if G_BYTE_ORDER == G_BIG_ENDIAN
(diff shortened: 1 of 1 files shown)


EgorBo:
@NattyNarwhal I am currently porting Decimal from CoreRT - https://github.com/mono/mono/pull/8626

EgorBo:
@NattyNarwhal the new decimal in CoreRT is fully managed implementation (except number-to-string stuff via snprintf)

PR #9430, opened 2018-07-05
[Coop] Mostly convert System.Array.

jaykrell:


mono/metadata/icall-def.h
```diff
@@ -151,17 +151,17 @@ ICALL(ARGI_3, "IntGetNextArgType",                mono_ArgIterator_IntGetNextArg
...
 ICALL(ARRAY_5, "GetGenericValueImpl", ves_icall_System_Array_GetGenericValueImpl)
-ICALL(ARRAY_6, "GetLength",        ves_icall_System_Array_GetLength)
-ICALL(ARRAY_15, "GetLongLength",        ves_icall_System_Array_GetLongLength)
-ICALL(ARRAY_7, "GetLowerBound",    ves_icall_System_Array_GetLowerBound)
-ICALL(ARRAY_8, "GetRank",          ves_icall_System_Array_GetRank)
-ICALL(ARRAY_9, "GetValue",         ves_icall_System_Array_GetValue)
-ICALL(ARRAY_10, "GetValueImpl",     ves_icall_System_Array_GetValueImpl)
+HANDLES(ICALL(ARRAY_6, "GetLength",        ves_icall_System_Array_GetLength))
+HANDLES(ICALL(ARRAY_15, "GetLongLength",   ves_icall_System_Array_GetLongLength))
+HANDLES(ICALL(ARRAY_7, "GetLowerBound",    ves_icall_System_Array_GetLowerBound))
```
lambdageek: `GetLength` and `GetLowerBound` are sometimes replaced by the JIT with intrinsics. https://github.com/mono/mono/blob/1ba158386dfd8fbba3c2d8a60da77dc4189c676a/mono/mini/intrinsics.c#L345-L346

Also there's this bit in `method-to-ir.c` which looks like it will skip making a wrapper for `System.Array` icalls... 
https://github.com/mono/mono/blob/1ba158386dfd8fbba3c2d8a60da77dc4189c676a/mono/mini/method-to-ir.c#L7736-L7744

So I'm a little suspicious about adding handles here.  Can we actually get to this slow path, and if so, do the arguments look right?

mono/metadata/icall.c
```diff
@@ -153,84 +153,73 @@ mono_icall_get_file_path_prefix (const gchar *path)
...
+ves_icall_System_Array_GetValueImpl (MonoArrayHandle array, guint32 pos, MonoError *error)
 {
-	ERROR_DECL (error);
-	MonoClass *ac;
-	gint32 esize;
-	gpointer *ea;
-	MonoObject *result = NULL;
-
-	ac = (MonoClass *)arr->obj.vtable->klass;
+	MonoVTable *array_vtable = MONO_HANDLE_GETVAL (array, obj.vtable);
```
lambdageek: `array_vtable` seems to only be used to get the domain and klass of `array`. So just use `mono_handle_class (array)` and `MONO_HANDLE_DOMAIN (array)`, respectively.

mono/metadata/icall.c
```diff
@@ -153,84 +153,73 @@ mono_icall_get_file_path_prefix (const gchar *path)
...
+ves_icall_System_Array_GetValue (MonoArrayHandle arr, MonoArrayHandle indices, MonoError *error)
 {
-	MonoClass *ac, *ic;
-	MonoArray *io;
-	gint32 i, pos, *ind;
-	ERROR_DECL (error);
+	MONO_CHECK_ARG_NULL_HANDLE (indices, NULL_HANDLE);
 
-	MONO_CHECK_ARG_NULL (idxs, NULL);
+	MonoClass *indices_class = MONO_HANDLE_GETVAL (indices, obj.vtable->klass);
```
lambdageek: `mono_handle_class (indices)`

mono/metadata/icall.c
```diff
@@ -651,6 +639,8 @@ ves_icall_System_Array_CreateInstanceImpl (MonoReflectionType *type, MonoArray *
 	return array;
 }
 
+
+// FIXME Combine ves_icall_System_Array_CreateInstanceImpl64 and ves_icall_System_Array_CreateInstanceImpl
```
lambdageek: `CreateInstanceImpl64` doesn't seem to have a declaration (or callers) on the managed side. Just delete it.

mono/metadata/icall.c
```diff
@@ -707,81 +697,54 @@ ves_icall_System_Array_CreateInstanceImpl64 (MonoReflectionType *type, MonoArray
...
-		return 0;
-	}
-	
-	if (arr->bounds == NULL)
-		length = arr->max_length;
-	else
-		length = arr->bounds [dimension].length;
-
-#ifdef MONO_BIG_ARRAYS
+	uintptr_t const length = ves_icall_System_Array_GetLongLength (arr, dimension, error);
```
lambdageek: (Just my personal preference) I don't really like this pattern (calling  an icall implementation from inside another function in the runtime).  I'd make a static helper and call it from both of the `ves_icall` functions.  I think there will be value in having icall implementations that are as mechanical as possible and delegate all the real work to helper functions.

mono/metadata/object-internals.h
```diff
@@ -25,7 +25,7 @@
 #define MONO_CHECK_ARG(arg, expr, retval) do {				\
 	if (G_UNLIKELY (!(expr)))					\
 	{								\
-		if (arg) {} /* check if the name exists */		\
+		(void)(arg); /* check if the name exists */		\
```
lambdageek: (bikeshedding) why not `if (0) { (void)(arg); }` ?

jaykrell: Seems like extra unnecessary evaluation.

mono/metadata/icall-def.h
jaykrell: Agreeed slow path might be difficult to reach. I could add reflection-based tests.

jaykrell: Direct icalls I believe currently disabled.

lambdageek: @jaykrell sorry I highlighted too much code to try to show context and then didn't actually call out the suspicious bits.  It's not the direct calls. I meant the first condition `cmethod->iflags & METHOD_IMPL_ATTRIBUTE_INTERNAL_CALL && `mini_class_is_system_array (cmethod->klass)`  will cause us to skip the third one (`fsig->pinvoke`) where the call to `mono_marshal_get_native_wrapper` is the bit that actually adds the wrapper around the icall invocation. 

So if I'm reading this right, icalls in `System.Array` will never get a wrapper.

lambdageek: Also I  don't fully understand this part of `method-to-ir.c`.  I think we first add the wrapper for the icall, and then replace the call instruction with an intrinsic.

lambdageek: I think it's probably difficult to write a test that reliably ends up on the slow path (we do a lot to try to make arrays fast), but it's worth doing a local test (maybe some printf debugging or flipping some bits with the debugger to bypass the intrinsification) just to verify that the args on the slow path get the expected wrappers.

vargaz:
Arrays are very perf critical so have to be really careful here not to add too much overhead.

jaykrell:
`fixed` might be good on some of this.

mono/metadata/icall.c
```diff
@@ -707,84 +729,61 @@ ves_icall_System_Array_CreateInstanceImpl64 (MonoReflectionType *type, MonoArray
...
+	return MONO_HANDLE_GETVAL (arr, bounds) ? MONO_HANDLE_GETVAL (arr, bounds [dimension].length)
+						: MONO_HANDLE_GETVAL (arr, max_length);
 }
 
 ICALL_EXPORT gint32
-ves_icall_System_Array_GetLowerBound (MonoArray *arr, gint32 dimension)
+ves_icall_System_Array_GetLowerBound (MonoArrayHandle arr, gint32 dimension, MonoError *error)
 {
-	gint32 rank = m_class_get_rank (mono_object_class (&arr->obj));
+	icallarray_print ("%s arr:%p dimension:%d\n", __func__, arr, (int)dimension);
```
lambdageek: So do you get output for each of the converted icalls when you run the new tests? And it all looks reasonable?

mono/tests/coop-array.mak
```diff
@@ -0,0 +1,27 @@
+# This file is not used in automated tests, but is useful for unautomated testing.
+
```
lambdageek: I'd like to see a comment here about making sure `MONO_DEBUG_ICALLARRAY` is defined in the source and `MONO_LOG_MASK=icallarray` is set.  Otherwise it's not obvious what is expected from running this manually.

jaykrell:
I think should iterate the testing a bit more.

jaykrell:
I will try also ldftn testing.

jaykrell:
`ves_icall_System_Array_ClearInternal`
`ves_icall_System_Array_GetValueImpl`
at least are not removed by JIT. But more testing is reasonable.

filipnavara:
Apparently the "debug_icallarray = true" line in the last commit managed to produce more than 2 Gb of logs and clogged Jenkins macOS builds. While this is definitely an infrastructure issue it may be worth keeping in mind in order not to trigger it again.

mono/tests/coop-array.mak
```diff
@@ -0,0 +1,27 @@
+# This file is not used in automated tests, but is useful for unautomated testing.
```
akoeplinger: I don't see why we need this file, why is the normal runtime test infrastructure not sufficient?

jaykrell: I thought it was useful for AOT but it isn't. Perhaps it isn't better than Makefile.am but I did use it again tonight.

jaykrell:
@monojenkins build Linux i386

mono/tests/coop-array.mak
lambdageek: Maybe just add a new target to Makefile.am that focuses just on this set of tests?

mono/metadata/object-internals.h
jaykrell: 1. Is it good or bad that the old form is invalid for type safe handles?
I forsee it being valid in C++ though, to replace MONO_HANDLE_IS_NULL.

2. I misread your comment. I thought you were asking vs. what was there.
Your ```if (0) { (void)(arg); }`` is nice.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

PR #9431, opened 2018-07-05
[Coop] Partial conversion of COM interop.

jaykrell:
Esp. but not entirely unimplemented stuff.


mono/metadata/cominterop.c
```diff
@@ -1884,26 +1879,22 @@ ves_icall_Mono_Interop_ComInteropProxy_FindProxy (gpointer pUnk, MonoError *erro
...
-		return NULL;
+	MonoCCW * const ccw = verify ? (MonoCCW *)g_hash_table_lookup (ccw_interface_hash, ccw_entry) : ccw_entry->ccw;
+	g_assert (verify || ccw);
+	return ccw ? mono_gchandle_get_target_handle (ccw->gc_handle) : NULL_HANDLE;
+}
+
+static MonoObject*
+cominterop_get_ccw_object (MonoCCWInterface* ccw_entry, gboolean verify)
+{
+	return MONO_HANDLE_RAW (cominterop_get_ccw_object_handle (ccw_entry, verify));
```
lambdageek: This is going to leak handles.   Note that this function is (sometimes, but not always) used as a dynamically-registered icall:
https://github.com/mono/mono/blob/master/mono/metadata/cominterop.c#L634

So I would suggest:
1. Make a new entrypoint `static ves_icall_cominterop_get_ccw_object` and use that exclusively as a dynamically registered icall.
2. Keep the `cominterop_get_ccw_object` for now until its remaining caller(s) are comverted to use the handle-ized version.
3. In both (1) and (2) add handle frame entry/exit macros so that it doesn't leak the coop handles.

mono/metadata/icall.c
```diff
@@ -7948,82 +7912,64 @@ ves_icall_Mono_TlsProviderFactory_IsBtlsSupported (MonoError *error)
...
 {
-	ERROR_DECL (error);
 	mono_error_set_not_implemented (error, "System.Runtime.InteropServices.WindowsRuntime.UnsafeNativeMethods.RoOriginateLanguageException internal call is not implemented.");
-	mono_error_set_pending_exception (error);
 	return FALSE;
 }
 
 ICALL_EXPORT void
-ves_icall_System_Runtime_InteropServices_WindowsRuntime_UnsafeNativeMethods_RoReportUnhandledError (MonoObject* oerr)
+ves_icall_System_Runtime_InteropServices_WindowsRuntime_UnsafeNativeMethods_RoReportUnhandledError (MonoObjectHandle oerr, MonoError *error)
```
lambdageek: These are silly.  Here's what we have in `mcs/class/corlib/System.Runtime.InteropServices.WindowsRuntime/UnsafeNativeMethods.cs`:

```csharp
#if !DISABLE_COM
		public static extern void RoReportUnhandledError(IRestrictedErrorInfo error);
#else
		public static void RoReportUnhandledError(IRestrictedErrorInfo error) {
			throw new NotImplementedException ();
		}
#endif
```

(And I don't see a Win32 specific C impl for these, in Mono, either.  So they just... don't do anything, ever).

IMO, we should drop these icalls and just do the NIE throw from managed, always.

jaykrell:
Agreed on leaky handles. I thought about this but misjudged it. Silly: It’d be nice to remove ifdef from bcl & have it be shared.

mono/metadata/cominterop.c
```diff
@@ -1884,26 +1879,28 @@ ves_icall_Mono_Interop_ComInteropProxy_FindProxy (gpointer pUnk, MonoError *erro
...
-		ccw = ccw_entry->ccw;
-		g_assert (ccw);
-	}
-	if (ccw)
-		return mono_gchandle_get_target (ccw->gc_handle);
-	else
-		return NULL;
+	MonoCCW * const ccw = verify ? (MonoCCW *)g_hash_table_lookup (ccw_interface_hash, ccw_entry) : ccw_entry->ccw;
+	g_assert (verify || ccw);
+	return ccw ? mono_gchandle_get_targe (ccw->gc_handle) : NULL;
```
lambdageek: typo. ("`targe`")

jaykrell:
Relevant test failures.

jaykrell:
@monojenkins build failed

PR #9432, opened 2018-07-05
Restructure NetworkInformation classes to avoid full AOT compile errors.

lateralusX:
When doing full AOT of System.dll on Windows and other platforms building runtime without HOST_DARWIN or HOST_BSD we will get an ugly error, but compilation will still succeed leaving the user confused around the result.

The following error is printed:

--------------------------------------

cant resolve internal call to
"System.Net.NetworkInformation.MacOsIPInterfaceProperties::ParseRouteInfo_internal"
(tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: C:\github\mono\msvc\build\fullaot\winaot\system.dll

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.
cant resolve internal call to
"System.Net.NetworkInformation.MacOsIPInterfaceProperties::ParseRouteInfo_internal"
(tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: C:\github\mono\msvc\build\fullaot\winaot\system.dll

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.

--------------------------------------

The problem is that MacOs/BSD specific NetworkInformation classes are build into System.dll on all platforms. One way to solve this is to start splitting the different platforms specific implementations into different source files so we could start to include/exclude sources per platform and profile in a nicer way instead of just ifdef:ing inside current source files. This strategy also matches how corefx have structured their NetworkInformation sources making it easier in the future to bring mono’s NetworkInformation implementation closer to corefx. This PR implements this strategy.

Due to the fact that we don’t have platform specific configs for all profiles (like basic and build) we can’t exclude all platform specific sources from common.sources in System and move them to platform specific profiles. Ones we have that full capability we could drop the #if WIN_PLATFORM in all the Win32* files and exclude them from common.sources and only include them in the corresponding
win32 specific profiles. The same strategy could then be applied for other platform specific profiles as well.

For Windows specific profiles, like win32_net_4_x_System.dll and winaot_System.dll we use the .exclude.sources to remove the unix/linux/macos NetworkInformation sources, so with this commit we will get a System.dll that only includes the Win32 specific NetworkInformation classes and none of the unix/linux/macos, this will make sure the compile error during full AOT won’t happen since there is no classes left including the icall not available on a Windows mono runtime.

Other platforms will have the same classes in System.dll as before for now but the source code is now structured making it possible to include/exclude platform specific source files for other platforms as well, when needed.


lateralusX:
Errors in "Project Files Diff" and "Linux x64" are due to changes in system.csproj files and that is expected since a number of new files was added and some files that ended up empty removed from the csproj, matching changes done to the common.sources file.

PR #9433, opened 2018-07-05
[interp] Use _fast variants of array functions are the former because tons slower for being part of the runtime API.

kumpera:


lewurm:
@monojenkins build failed

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #9435, opened 2018-07-05
Remove unused ves_icall_System_Array_CreateInstanceImpl64.

jaykrell:


vargaz:
There is no need to bump corlib for adding/removing icalls.

jaykrell:
Checking first. Do we ever append "64" to names. I don't see it.

jaykrell:
https://msdn.microsoft.com/en-us/library/1z8w3at5(v=vs.110).aspx ?

lambdageek:
The upshot is the native icall implementation is not used and is just taking up space in the icall-def.h table.

jaykrell:
Review?

PR #9436, opened 2018-07-05
[checked-build] don't collect state transition backtraces on abort_blocking

lambdageek:
for the same reason as the other cases in this function - if we're in `RUNNING`
or `ASYNC_SUSPEND_REQUESTED` under hybrid suspend we're supposed to cooperatively
suspend.  If we call `backtrace()`, we might block waiting for the low-level
dynamic linker mutex on Linux which may be held by a thread that was
preemptively-suspended already.


lambdageek:
@monojenkins squash

PR #9447, opened 2018-07-06
Add GC Unsafe thread state transitions to MONO_API functions - third batch

lambdageek:
Part of #6921, this is a third PR that adds thread state transitions to switch to GC Unsafe state when embedders call API functions from outside of Mono.


lambdageek:
Coop and hybrid failures in `install_eh_callback.exe` and `pinvoke3.exe` are related.

lambdageek:
Rebased on top of #9502 which should fix the coop and hybrid failures.

luhenry:

luhenry:
@monojenkins build Linux x64

luhenry:

PR #9454, opened 2018-07-08
System.Net.NetworkInformation.NetworkInterface backend for AIX

NattyNarwhal:
Yes, I do know of a possible refactor, but why not get this working? This will enable `System.Net.Sockets.Socket` to depend IPv4/IPv6 are working, and thus, fix a LOT of issues on CI; a bunch of test suites have one or two things that depend on loopback, which is currently busted unless you nop out Socket's protocol check.

Also made a note that the Mac version of the backend is actually likelier to work under other OSes. Structures are kinda ugly, but they work; it took a whole morning and night beating them into submission.

Untested under PASE, though I've marked notes for self to what's different. (My guess: it works, but not as well as it should - but good enough so that Socket starts working again.)

And hey, the Mac backend can't do this!

```
csharp> NetworkInterface.GetAllNetworkInterfaces().Select(x => $"The MTU of {x.Name} is {x.GetIPProperties().GetIPv4Properties().Mtu}");
{ "The MTU of en0 is 1500", "The MTU of lo0 is 16896" }   
```


marek-safar:
I don't think we should get this in before with switch to CoreFX as it'd make it quite harder for us as the implementation will look quite different

NattyNarwhal:
I was somewhat expecting that, even if this would get through the period until then. I can probably wait for whenever the CoreFX conversion for this part happens, and then graft the `SIOCGIFCONF` backend afterwards?

NattyNarwhal:
Just curious, what's with the merge? Seems like it changed lots of lines. I also thought the CoreFX version was coming soon, and I was told it would be better to wait for that?

akoeplinger:
@NattyNarwhal I was just resolving the conflict to get a new build, System.csproj is auto-generated anyway. Just trying to see whether we should get this in or close the PR, there's no point in keeping a PR open for months if it won't be merged :)

NattyNarwhal:
I think it'd depend on how soon the CoreFX version arrives. I've skimmed the C side of it, so I think I know what to do there when it comes; it's just that I'm not touching CoreFX unless I have to. (mix of can't test/slightly more complicated process)

akoeplinger:
I'm not aware of any in-progress PRs so it might take a little. Even if we delete this code in a month it's better than having it just sit around ;)

NattyNarwhal:
Yeah, I think it'd turn a lot of tests on AIX green.

akoeplinger:
@monojenkins commit csproj

PR #9457, opened 2018-07-09
Enable SIMD optimization on Windows x64.

lateralusX:
When implementing Windows x64 value type ABI mono’s SIMD optimization didn’t handle the case when passed arguments uses ArgValuetypeAddrInIReg/ArgValuetypeAddrOnStack so optimization was disabled on Windows x64.

This commit fixes mono SIMD support for ABI’s passing value types using ArgValuetypeAddrInIReg/ArgValuetypeAddrOnStack on Windows x64. It does this as part of regular value type decomposition, including a pass decomposing SIMD intrinsic ending up as arguments of type OP_ARG into correct load/store of the value type reference before argument can be used by SIMD instructions.

All JIT and full AOT regression tests pass with enabled SIMD optimization on Windows x64. Running this optimization on some vector intensive applications showed some great performance increase. Just running this on basic-simd.exe regression shows:

Elapsed time: 0.040471 secs (no optimizations at all)
Elapsed time: 0.015756 secs (only simd and intrins optimization)

So in this small, isolated test, the optimization indicates a 2.5x performance increase.


lateralusX:
@vargaz OK with changes?

mono/mini/mini.h
```diff
@@ -2674,11 +2680,16 @@ enum {
 };
 
 const char *mono_arch_xregname (int reg);
+guint32     mono_arch_cpu_enumerate_simd_versions (void);
+
+#ifdef MONO_ARCH_SIMD_INTRINSICS
 void        mono_simd_simplify_indirection (MonoCompile *cfg);
+void        mono_simd_decompose_intrinsic (MonoCompile *cfg, MonoBasicBlock *bb, MonoInst *ins);
```
kumpera: This is really bad naming, mono_simd_decompose_intrinsic and mono_simd_decompose_intrinsics differs by a single letter and do different things.

mono/mini/simd-intrinsics.c
```diff
@@ -817,6 +817,96 @@ mono_simd_simplify_indirection (MonoCompile *cfg)
...
+			store_ins->type = STACK_VTYPE;
+			store_ins->sreg1 = ins->dreg;
+			store_ins->dreg = varload_ins->dreg;
+			mono_bblock_insert_after_ins (bb, varload_ins, store_ins);
+		}
+	}
+}
+
+void
+mono_simd_decompose_intrinsics (MonoCompile *cfg)
```
kumpera: This function is not called anywhere.

It should probably check for cfg->opts and if simd intrinsics were actually used or not.

mono/mini/method-to-ir.c
```diff
@@ -9929,13 +9933,15 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 
 					MONO_EMIT_NULL_CHECK (cfg, sp [0]->dreg, foffset > mono_target_pagesize ());
 
+#ifdef MONO_ARCH_SIMD_INTRINSICS
```
kumpera: This should be changed to use MONO_CLASS_IS_SIMD instead of using an ifdef

mono/mini/simd-intrinsics.c
```diff
@@ -817,6 +817,96 @@ mono_simd_simplify_indirection (MonoCompile *cfg)
...
+		*sreg = load_ins->dreg;
+	}
+}
+
+/*
+* Windows x64 value type ABI uses reg/stack references (ArgValuetypeAddrInIReg/ArgValuetypeAddrOnStack)
+* for function arguments. When using SIMD intrinsics arguments optimized into OP_ARG needs to be decomposed
+* into correspondig SIMD LOADX/STOREX instructions.
+*/
+#if defined(TARGET_WIN32) && defined(TARGET_AMD64)
```
kumpera: Move this to mini-amd64.h and give it a symbolic name instead of this?

PR #9459, opened 2018-07-09
Remove unused KeyAgreeKeyChoice enum

MaximLipnin:
I [didn't find](https://github.com/mono/mono/search?q=KeyAgreeKeyChoice&unscoped_q=KeyAgreeKeyChoice) any use of `KeyAgreeKeyChoice` enum in Mono.

CoreFX also doesn't contain it (there is only [message in api compat baseline](https://github.com/dotnet/corefx/blob/3442edac91d958b76a3bd6cab0e00fe376a3353a/src/shims/ApiCompatBaseline.netcoreapp.netfx461.txt#L428)). The similar thing here is [CmsKeyAgreeKeyChoice](https://github.com/dotnet/corefx/search?q=CmsKeyAgreeKeyChoice&unscoped_q=CmsKeyAgreeKeyChoice)

NetFX has only declaration (https://referencesource.microsoft.com/#q=KeyAgreeKeyChoice ).

As part of #7589.


filipnavara:
It is not used anywhere, but unfortunately it is part of an API contract of .NET Framework and as such it should stay.

filipnavara:
CoreFX dropped it deliberately because it was not used anywhere. I can find the issue referencing it if necessary.

MaximLipnin:
So in similar cases I also need to look at something like https://docs.microsoft.com/en-us/dotnet/api/?view=netframework-4.7.2&term=KeyAgreeKeyChoice to check that there is no API compatibility issue, right?

filipnavara:
Yes. Also the "public" decoration in the Reference Source is a red flag. And if everything fails then the PR checks also do the verification against https://github.com/mono/reference-assemblies, so it will fail there as well.

MaximLipnin:
So that's my lesson, thanks. I assume that the PR should be closed.

PR #9465, opened 2018-07-09
[mdoc] bump to 5.7.2 release

joelmartinez:
This bumps to the `api-doc-tools` repo (ie. mdoc) to the latest [5.7.2 release](https://github.com/mono/api-doc-tools/releases/tag/tag-mdoc-5.7.2)


joelmartinez:
hmm ... ok, it seems that there's some difference in the way the mono project is building mdoc/monodoc from what the root repo is doing:

```
external/api-doc-tools/monodoc/Monodoc/storage/ZipStorage.cs(106,23): error CS1061: 'ZipOutputStream' does not contain a definition for 'UseZip64' and no extension method 'UseZip64' accepting a first argument of type 'ZipOutputStream' could be found (are you missing a using directive or an assembly reference?)
/home/builder/jenkins/workspace/test-mono-pull-request-coop-arm64/external/api-doc-tools/monodoc/Monodoc/storage/ZipStorage.cs(106,34): error CS0103: The name 'UseZip64' does not exist in the current context
```

need to figure out what's going on here

marek-safar:
We build against in-tree ICSharpCode.SharpZipLib.dll

joelmartinez:
@marek-safar ahh I see ... so what is the process for updating that version of SharpZipLib to a version that includes the `UseZip64` property? I introduced the use of that to preserve backwards compatibility: https://github.com/mono/api-doc-tools/issues/215

marek-safar:
Hmm, that won't be easy. Do you need that property or the fact that it's missing means it's false by default?

joelmartinez:
@marek-safar right, if you look at the original issue (https://github.com/icsharpcode/SharpZipLib/issues/131#issue-169874004), the backwards incompatibility was intentional in the original library, and the workaround for those who needs backwards compat is to set `UseZip64` to Off.

I still have some deployment scenarios where this compatibility comes into play.

joelmartinez:
@marek-safar I've updated the submodule to include the following commit: https://github.com/mono/api-doc-tools/commit/21606e5993111da069afd2081bab3b3dd4e963b8

joelmartinez:
@marek-safar Hmm, strange error:

```
make[9]: *** No rule to make target '../../../external/api-doc-tools/mdoc/Mono.Documentation/monodocer.cs', needed by '../../class/lib/net_4_x-linux/mdoc.exe'.  Stop.
```
`monodocer.cs` was renamed in the repository ... is this being built by a different makefile than the one in the `api-doc-tools` repository?

joelmartinez:
@marek-safar ok, looking at the repo, looks like there's another .csproj, and it's [generated](https://github.com/mono/mono/commit/e271c293ee7a364a853a30f8574830ea84e0481a#diff-3f16c8e588abf7ff58abd27ded7acfaf) by something (presumably the `mdoc.exe.sources` file)?

Also, I couldn't find a `net_4_x-linux/mdoc.exe` makefile target anywhere in the mono repo.

Basically ... just looking for some guidance on how to do this "right" :)

marek-safar:
Correct, you need to update https://github.com/mono/mono/blob/master/mcs/tools/mdoc/mdoc.exe.sources

steveisok:
@marek-safar do we still want to take this?

marek-safar:
yes but this need more work

lewing:
@joelmartinez is this still something you have interest in or is it orphaned at this point?

joelmartinez:
@lewing ... yes, I very much have an interest in it; I just have not had any bandwidth to follow up. mdoc has been iterating regularly (github and nuget releases), but the difference in project build structure has made integration back into the mono project difficult. As a result, mdoc still effectively ships with an "old" version of mdoc.

akoeplinger:
Replaced by https://github.com/mono/mono/pull/16035

PR #9469, opened 2018-07-09
[runtime] Print tids in hex in mono_threads_summarize

lambdageek:
Match the rest of our debug output that prints MonoNativeThreadId in hex

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -5983,7 +5983,7 @@ mono_threads_summarize (MonoContext *ctx, gchar **out, MonoStackHash *hashes)
 
 	MonoNativeThreadId current = mono_native_thread_id_get ();
 
-	MOSTLY_ASYNC_SAFE_PRINTF("Entering thread summarizer from %zu\n", current);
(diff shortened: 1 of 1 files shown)


lambdageek:
Example of slightly unpleasant output:
```
Assertion failure in thread 0x7fff948dc380 due to: Received cookie 0x0 but expected 0x7fd345801c00

Last 2 state transitions: (most recent first)
[ABORT_BLOCKING] RUNNING -> RUNNING (0) 0 at:
[ABORT_BLOCKING] RUNNING -> RUNNING (0) 0 at:

Entering thread summarizer from 140735685706624
Entering thread summarizer from 123145451487232
Self-reporting for thread 123145451487232. Registered summarizing thread right now is 123145451487232
Assertion failure in thread 0x700008e44000 due to: object.c:1868: Expected GC Unsafe mode but was in STATE_BLOCKING state
```

140735685706624 is 0x7fff948dc380 and 123145451487232 is 0x700008e44000 so it'd be nice to have everything match up without pulling out a hex converter

lambdageek:
@monojenkins squash

vargaz:
@monojenkins build failed

PR #9472, opened 2018-07-10
Introducing a general-purpose dependency injector to use `System` code from within `corlib`.

baulig:
## Dependency Injector

* `corlib/Mono/DependencyInjector.cs`: this is corlib's entry point.

* `corlib/Mono/ISystemDependencyProvider.cs`: this interface is implemented in `System.dll`.

* `System/Mono/SystemDependencyProvider.cs`: the `System.dll` side of it.

To use this, we need to call `SystemDependencyProvider.Initialize()` during system startup on Mobile
(this is already done for Android, but still needs to be done for XI/XM).

For Desktop, this will be triggered by a few `System.dll` based APIs such as for instance `SslStream`
or `X509Certificate2` - and the corlib side can use reflection if needed.

The corlib side should use `DependencyInjector.GetSystemDependency()`.

## X509Certificate - corlib side

We previously had a mechanic called `INativeCertificateHelper` (which unfortunately was not really
working as intended) to use the BTLS code (which lives in System) in corlib.  This has now been
replaced to use the new dependency injector.

The actual implementation as well as all the platform conditionals have now been moved into `System.dll`
where they can take advantage of existing PAL code.

This also removes all `Mono.Security.X509` usage from the `System.Security.Cryptography.X509Certificates`
namespace (it is still used by some of the other crypto classes, so we can't remove it yet).

* `Mono/ISystemCertificateProvider.cs`: this interface is now hooked up via the new dependency injector and replaces `INativeCertificateHelper`.

* `System.Security.Cryptography.X509Certificates/X509Helper.cs`: this file has been completely reworked and is now just a tiny stub around `ISystemCertificateProvider` and will eventually go away completely.  The entire implementation has been moved into `System.dll`.

* `System.Security.Cryptography.X509Certificates/X509CertificateImplMono.cs`: removed, we now use `X509Certificate2ImplMono` from System via `ISystemCertificateProvider`.

* `System.Security.Cryptography.X509Certificates/X509CertificateImplApple.cs`: moved into `System.dll` and slightly reworked there.

* `System.Security.Cryptography.X509Certificates/X509Helper.Apple.cs`: removed, we do not have any platform-specific code here anymore; the new `ISystemCertificateProvider` provides this for us.

## X509Certificate - System side

The actual implementation as well as all the platform-specific code now lives in `System.dll`.

* `Mono/SystemCertificateProvider.cs`: the `System.dll` implementation of `ISystemCertificateProvider`. We do the platform / TLS Provider checks in here, then construct a custom PAL class.

* `Mono/X509Pal.cs`: New internal static class.

* `Mono/X509PalImpl.cs`: New abstract internal class.  It contains some common code, including the fallback implementation.

* `Mono/X509PalImpl.Mono.cs`: Fallback PAL, used by platforms without SslStream / MonoTlsProvider.

* `Mono.Btls/X509PalImpl.Btls.cs`: BTLS implementation.

* `Mono.AppleTls/X509PalImpl.Apple.cs`: We are using the `MONO_FEATURE_APPLE_X509` conditional in here, so it can be used without the actual Apple TLS code.

* `Mono.AppleTls/X509CertificateImplApple.cs`: moved here from corlib.

* `Mono.AppleTls/*.cs`: Use the `X509CertificateImplApple(IntPtr)` constructor everywhere instead of `X509Certificate(IntPtr)` to make sure we are using the "correct" PAL.


mcs/class/corlib/Mono/ISystemDependencyProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
```
marek-safar: Clean up the usings

mcs/class/corlib/Mono/DependencyInjector.cs
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+#if !MOBILE
+		static ISystemDependencyProvider ReflectionLoad ()
+		{
+			var assembly = Assembly.Load ("System");
+			if (assembly == null)
+				return null;
+
+			var type = assembly.GetType ("Mono.SystemDependencyProvider");
```
marek-safar: Use `Type.GetType` to help linker to catch the dependency

mcs/class/corlib/Mono/DependencyInjector.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+	static class DependencyInjector
+	{
+		/*
+		 * Allows us to use code from `System.dll` in `mscorlib.dll`.
+		 * 
+		 */
+		internal static ISystemDependencyProvider SystemProvider {
+			get {
+				lock (locker) {
```
marek-safar: Reduce the lock scope

mcs/class/corlib/Mono/DependencyInjector.cs
```diff
@@ -0,0 +1,92 @@
...
+					if (systemDependency != null)
+						return systemDependency;
+
+#if MOBILE
+					// This can never happen because Mobile initializes this during system startup.
+					throw new PlatformNotSupportedException ("Cannot get `ISystemDependencyProvider`.");
+#else
+					systemDependency = ReflectionLoad ();
```
marek-safar: Would be better to if-def ReflectionLoad

mcs/class/corlib/Mono/DependencyInjector.cs
```diff
@@ -0,0 +1,92 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using System.Threading;
+using System.Threading.Tasks;
```
marek-safar: Why do you need Tasks?

mcs/class/System/Mono.AppleTls/X509CertificateImplApple.cs
```diff
@@ -1,11 +1,22 @@
-#if MONO_FEATURE_APPLETLS || MONO_FEATURE_APPLE_X509
+#if SECURITY_DEP && (MONO_FEATURE_APPLETLS || MONO_FEATURE_APPLE_X509)
```
marek-safar: SECURITY_DEP is always defined

mcs/class/System/Mono.AppleTls/X509PalImpl.Apple.cs
```diff
@@ -0,0 +1,66 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+#if SECURITY_DEP && (MONO_FEATURE_APPLETLS || MONO_FEATURE_APPLE_X509)
```
marek-safar: SECURITY_DEP is always defined

mcs/class/System/Mono.Btls/X509PalImpl.Btls.cs
```diff
@@ -0,0 +1,78 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+#if SECURITY_DEP && MONO_FEATURE_BTLS
```
marek-safar: SECURITY_DEP is always defined

mcs/class/System/Mono/SystemCertificateProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+#if SECURITY_DEP
```
marek-safar: SECURITY_DEP is always defined

mcs/class/System/Mono/SystemDependencyProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+
+		public SystemCertificateProvider CertificateProvider {
+			get;
+		}
+
+		public X509PalImpl X509Pal => CertificateProvider?.X509Pal;
+
+		SystemDependencyProvider ()
+		{
+#if SECURITY_DEP
```
marek-safar: SECURITY_DEP is always defined

mcs/class/System/Mono/SystemDependencyProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+				Interlocked.CompareExchange (ref instance, new SystemDependencyProvider (), null);
+		}
+
+		ISystemCertificateProvider ISystemDependencyProvider.CertificateProvider => CertificateProvider;
+
+		public SystemCertificateProvider CertificateProvider {
+			get;
+		}
+
+		public X509PalImpl X509Pal => CertificateProvider?.X509Pal;
```
marek-safar: Unnecessary ?.

mcs/class/System/Mono/X509PalImpl.Mono.cs
```diff
@@ -0,0 +1,59 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+#if SECURITY_DEP
```
marek-safar: SECURITY_DEP is always defined

mcs/class/corlib/Mono/DependencyInjector.cs
baulig: How do you want to reduce it?

baulig:
I am aware of the failing tests.  We currently have these issues:

- `X509CertificateTest.Certificate8` - This is an old bug in BTLS that we are now triggering, you can get the same result without this patch by using `X509Certificate2` and using `MONO_TLS_PROVIDER=btls`.  I have already fixed this in a separate PR, including a newly added test.
- `X509CertificateTest.Certificate7` - The test attempt to parse an invalid certificate, it is currently disabled on Mac (see comment in the test).  I need to investigate what's happening here.
- `X509CertificateTest.Certificate13` - This is a very poorly written test that needs to be either removed or rewritten.  The test has a bunch of `Assert.*` in it, but the entire test method is wrapped in `[ExpectedException (typeof (CryptographicException))]` - so essentially _something_ in that test method is supposed to throw.  That _something_ needs to be identified, then the expected exception clause narrowed around it.  BTLS fails here because the certificate has an unknown public key type.

mcs/class/corlib/Mono/DependencyInjector.cs
marek-safar: lock just the ReflectionLoad

filipnavara:
I think that at the moment it makes sense to throw `PlatformNotSupportedException` in `X509Certificate(IntPtr)`. However, in long term it would probably make sense to adopt the CoreFX approach and ensure that only one TLS/certificate implementation is used. CoreFX decides the one at compile time, we can do it at runtime. At the moment CoreFX supports the Handle overloads only on macOS and Windows anyway.

`X509CertificateTest.Certificate13` - I have a suspicion that something is not handled in BTLS quite right. Note the comments above the test - "has NO parameters (shared group is present in issuer certificate)". I already started looking into it, but haven't got too far.

baulig:
`X509CertificateTest.Certificate13` is an error test - we are simply throwing the wrong exception type.  Just fixed that in a separate PR, but I'm going to cherry-pick everything on top of here as well.

baulig:
@monojenkins build failed

baulig:

baulig:
@monojenkins commit csproj

marek-safar:
I don't think this should have been merged when it breaks on iOS. 

@baulig could you address that

alexischr:
@marek-safar @baulig is aware of the failures and asked to get it in as it is easier for him to work on the follow up

PR #9473, opened 2018-07-10
[sdk] Do not fix android platform-tools version. Make it configurable.

atsushieno:
mono/sdks setup for Android specifies the fixed version of platform-tools of 27.0.1, which brings incompatibility with the latest Android emulator which needs adb protocol version 40. It is only available at platform-tools 28.0.0. We specify 28.0.0 in our xamarin-android/build-tools/android-toolchains setup, but it was never respected when we set up android-toolchains. The reason was here.


lewurm:
should we backport this?

atsushieno:
We seem to be using 2018-02 so yes there please.

lewurm:

lewurm:

lewurm:

PR #9476, opened 2018-07-10
Fix race condition in WaitAnyWithSecondMutexAbandoned test.

### LINKED ISSUE #8708: WaitHandleTest.WaitAnyWithSecondMutexAbandoned is flaky on Windows
## Steps to Reproduce

1. Run WaitHandleTest.WaitAnyWithSecondMutexAbandoned on Windows
2. It sometimes fails, example:
      - https://jenkins.mono-project.com/job/z/label=w32/6928/testReport/
      - https://jenkins.mono-project.com/job/z/label=w64/6941/testReport/
      - https://jenkins.mono-project.com/job/z/label=w32/6945/testReport/

### Current Behavior

It sometimes fails with:

```
System.ApplicationException : Mutex is not owned

  at System.Threading.Mutex.ReleaseMutex () [0x00018] in D:\j\workspace\z\label\w32\mcs\class\corlib\System.Threading\Mutex.cs:207 
  at (wrapper remoting-invoke-with-check) System.Threading.Mutex.ReleaseMutex()
  at MonoTests.System.Threading.WaitHandleTest.WaitAnyWithSecondMutexAbandoned () [0x00170] in D:\j\workspace\z\label\w32\mcs\class\corlib\Test\System.Threading\WaitHandleTest.cs:526 
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in D:\j\workspace\z\label\w32\mcs\class\corlib\System.Reflection\MonoMethod.cs:305 
```

### Expected Behavior

No failure.

[ ] macOS
[ ] Linux
[ X ] Windows

**Version Used**: master

lateralusX:
There is a small race condition during the completion of a native thread join call. During that period of time the tid is no longer included on the internal list tracking joinable threads so a thread that will join on the tid while another thread (like the finalizer thread) is waiting on native join to complete for the same tid, will cause the managed Thread.Join call to complete before the native join call has completed. This race could cause issues on some OS:es that clear's up some thread resources, like abandoned mutexes when the thread has exited. This race is hit by WaitAnyWithSecondMutexAbandoned since the call to Thread.Join will return before the thread owning the mutex has terminated meaning that it doesn't get ownership of the abandoned mutex as assumed by the test.

Fix makes sure Thread.Join won't complete until native thread join is complete. Increasing the join timeouts in the test also eliminates a timeout error making it harder to hit the problematic code path, primarily during reproduction in the debugger.

Fixes #8708

### DIFF
--- a/mcs/class/corlib/Test/System.Threading/WaitHandleTest.cs
+++ b/mcs/class/corlib/Test/System.Threading/WaitHandleTest.cs
@@ -409,7 +409,7 @@ namespace MonoTests.System.Threading {
 					m.WaitOne ();
 				});
 				thread1.Start ();
-				thread1.Join (1000);
+				Assert.IsTrue (thread1.Join (Timeout.Infinite), "thread1.Join");
 				try {
 					m.WaitOne ();
 					Assert.Fail ("Expected AbandonedMutexException");
@@ -421,7 +421,7 @@ namespace MonoTests.System.Threading {
 					signalled = m.WaitOne (100);
 				});
 				thread2.Start ();
-				thread2.Join (1000);
+				Assert.IsTrue (thread2.Join (Timeout.Infinite), "thread2.Join");
 				Assert.IsFalse (signalled);
 
 				// Since this thread owns the Mutex releasing it shouldn't fail
@@ -489,7 +489,7 @@ namespace MonoTests.System.Threading {
 						m1.ReleaseMutex ();
 					});
 					thread1.Start ();
-					thread1.Join (1000);
+					Assert.IsTrue (thread1.Join (Timeout.Infinite), "thread1.Join");
 					thread2.Start ();
 					while (!mainProceed) {
 						Thread.Sleep (10);
@@ -502,7 +502,7 @@ namespace MonoTests.System.Threading {
 						Assert.AreEqual (m2, e.Mutex);
 					} finally {
 						thread2Proceed = true;
-						thread2.Join (1000);
+						Assert.IsTrue (thread2.Join (Timeout.Infinite), "thread2.Join");
 					}
 
 					// Current thread should own the second Mutex now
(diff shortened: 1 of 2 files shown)


mcs/class/corlib/Test/System.Threading/WaitHandleTest.cs
```diff
@@ -409,7 +409,7 @@ public void WaitOneWithAbandonedMutex ()
 					m.WaitOne ();
 				});
 				thread1.Start ();
-				thread1.Join (1000);
+				thread1.Join (Timeout.Infinite);
```
marek-safar: This is not right, it should be Assert.IsTrue (thread1.Join (<reasonable timeout>));

lateralusX:
Will investigate problems with coop suspend, have at least a theory on a scenario that could occur in coop mode.

lateralusX:
Current failures are due to unrelated problems on AArch64 builds. At least the updated PR resolved the regressions previously identified on coop and hybrid suspend. Would still like to run it on the AArch architectures before approve/merge.

jaykrell:
Rebase to master and arm64 problems will go away.

mono/metadata/threads.c
```diff
@@ -5608,6 +5619,39 @@ mono_threads_add_joinable_runtime_thread (gpointer thread_info)
...
+static void
+threads_add_pending_native_thread_join_call_nolock (gpointer tid)
+{
+	if (!pending_native_thread_join_calls)
+		pending_native_thread_join_calls = g_hash_table_new (NULL, NULL);
+
+	gpointer orig_key;
+	gpointer value;
+
+	if (!g_hash_table_lookup_extended (pending_native_thread_join_calls, tid, &orig_key, &value))
```
kumpera: Why is this using g_hash_table_lookup_extended when it's not doing anything with orig_key and value?

lateralusX: g_hash_table_lookup_extended handles cases when the table includes a value that is 0 or NULL (since it returns a gboolean instead of the value). Using regular g_hash_table_lookup doesn't distinct between a value of 0 or NULL and not existing at all in the table. The same pattern is used in several places in threads.c to handle cases where tid could be 0.

lateralusX:
Failures on Linux x64 FullAOT+LLVM and OS X x64 IOS SDK seems unrelated and appears on many PR's.

PR #9477, opened 2018-07-10
[mcs] Implements awaited expression check for indirections

marek-safar:


lewurm:
could you add a test case?

marek-safar:
@monojenkins backport 2018-06

PR #9479, opened 2018-07-10
Fix issue where GetCheckedIPs filters out addresses, when the AddressFamily is Unspecified

mderoy:
GetCheckedIPs filters out endpoints that do not match the remote endpoint address family, however in the case the remote endpoint is unspecified, we shouldn't do this filtering. Filtering was added in https://github.com/mono/mono/pull/8041

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -1172,6 +1172,10 @@ namespace System.Net.Sockets
 			DnsEndPoint dep = e.RemoteEndPoint as DnsEndPoint;
 			if (dep != null) {
 				addresses = Dns.GetHostAddresses (dep.Host);
+
+				if (dep.AddressFamily == AddressFamily.Unspecified)
+					return true;
+
 				int last_valid = 0;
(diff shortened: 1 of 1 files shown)


mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -1172,6 +1172,10 @@ bool GetCheckedIPs (SocketAsyncEventArgs e, out IPAddress [] addresses)
 			DnsEndPoint dep = e.RemoteEndPoint as DnsEndPoint;
 			if (dep != null) {
 				addresses = Dns.GetHostAddresses (dep.Host);
+
+				if (dep.AddressFamily == AddressFamily.Unspecified)
```
marek-safar: This could be done even before `Dns.GetHostAddresses`

mderoy: @marek-safar This can't be done before Dns.GetHostAddresses, as that populates the addresses field that we want to return. by returning out after the call to Dns.GetHostAddresses, we return the full list of addresses instead of filtering out unmatching address below.

PR #9480, opened 2018-07-10
[sdks] fix build for watchOS and tvOS

lewurm:
fixes regressions introduced by 5e9af502c052 and 2a8fcbd9fe5560de0d4d57d2da6cbb855f284f86. Also adding build targets to CI so we avoid breakages in the future.

parts of this PR need to be backported to `2018-06`

### DIFF
--- a/scripts/ci/run-jenkins.sh
+++ b/scripts/ci/run-jenkins.sh
@@ -89,8 +89,11 @@ if [[ ${CI_TAGS} == *'product-sdks-ios'* ]];
 	   echo "DISABLE_ANDROID=1" > sdks/Make.config
 	   echo "DISABLE_WASM=1" >> sdks/Make.config
 	   export device_test_suites="Mono.Runtime.Tests System.Core"
-	   ${TESTCMD} --label=runtimes --timeout=60m --fatal make -j4 -C sdks/builds package-ios-sim64 package-ios-target64
-	   ${TESTCMD} --label=cross --timeout=60m --fatal make -j4 -C sdks/builds package-ios-cross64 package-ios-cross32
+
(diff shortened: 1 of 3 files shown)


scripts/ci/run-jenkins.sh
```diff
@@ -89,8 +89,15 @@ if [[ ${CI_TAGS} == *'product-sdks-ios'* ]];
...
 	   echo "DISABLE_WASM=1" >> sdks/Make.config
 	   export device_test_suites="Mono.Runtime.Tests System.Core"
-	   ${TESTCMD} --label=runtimes --timeout=60m --fatal make -j4 -C sdks/builds package-ios-sim64 package-ios-target64
-	   ${TESTCMD} --label=cross --timeout=60m --fatal make -j4 -C sdks/builds package-ios-cross64 package-ios-cross32
+
+	   export sim_targets="sim64 sim32 simtv simwatch"
+	   export runtime_targets="target64 target32 targettv targetwatch"
+	   export cross_targets="cross64 cross32 crosswatch"
+
+	   for arch in ${sim_targets}; do ${TESTCMD} --label=build-sim-runtime-${arch} --timeout=20m --fatal make -j4 -C sdks/builds package-ios-${arch}; done
```
vargaz: This turns off parallel builds.

lewurm:
here a decent diff for the 2nd commit (via `git log --minimal`):
https://gist.github.com/lewurm/adffe388f8bb1236bbf38aefb8d1da22

scripts/ci/run-jenkins.sh
lewurm: hmm, I would argue that this is preferable so we can easily spot breakages. but I don't have strong opinions on that.

lewurm: fixed

lewurm:
@monojenkins rebase

Issue #7295, opened 2018-02-26
[coop] Crash in DebuggerTests.SingleStepping


kumpera:
Found during CI.

Looks like this:
```

***** MonoTests.DebuggerTests.SetIPSingleStep
***** MonoTests.DebuggerTests.SimpleBreakpoint
***** MonoTests.DebuggerTests.SingleStepping
Stacktrace:

  at <unknown> <0xffffffff>
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Threading.Monitor.Monitor_wait (object,int) [0x00008] in <998b9a650d584d368d7f2e3f56fce105>:0
  at (wrapper managed-to-native) System.Threading.Monitor.Monitor_wait (object,int) [0x00008] in <998b9a650d584d368d7f2e3f56fce105>:0
  at System.Threading.Monitor.ObjWait (bool,int,object) [0x0002f] in /mnt/jenkins/workspace/test-mono-pull-request-coop/mcs/class/corlib/System.Threading/Monitor.cs:83
...

Thread 1 (Thread 0x14c3d19fa780 (LWP 63952)):
#0  0x000014c3d0ec2ed9 in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00000000004b8a1c in mono_handle_native_crash (signal=0x707fb6 "SIGABRT", ctx=<optimized out>, info=<optimized out>) at mini-exceptions.c:2722
#2  <signal handler called>
#3  0x000014c3d0ebf404 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
#4  0x00000000006d05c5 in mono_os_cond_wait (mutex=0x2535d20, cond=0x2535d48) at mono-os-mutex.h:173
#5  mono_os_cond_timedwait (cond=cond@entry=0x2535d48, mutex=mutex@entry=0x2535d20, timeout_ms=timeout_ms@entry=4294967295) at mono-os-mutex.c:32
#6  0x000000000063c445 in mono_coop_cond_timedwait (timeout_ms=4294967295, mutex=0x2535d20, cond=0x2535d48) at ../../mono/utils/mono-coop-mutex.h:102
#7  mono_w32handle_timedwait_signal_naked (alerted=0x7ffeee6daa58, poll=0, timeout=4294967295, mutex=0x2535d20, cond=0x2535d48) at w32handle.c:652
[...]

PR #9486, opened 2018-07-10
[runtime] Take ownership of thread mutex before freeing it

alexanderkyte:
In this crash:  https://github.com/mono/mono/issues/7295 we see that the mutex is freed while it is being held by another thread. In order to avoid this situation, I CAS in a loop until we can win and get ownership of the pointer.

Once we have it, nobody else should be able to get a new reference to it. We lock it, so we wait until other threads unlock it. We then unlock it, then we can free it. Our CAS prevents another thread from locking it after we unlock it.


mono/metadata/threads.c
```diff
@@ -1603,9 +1603,25 @@ ves_icall_System_Threading_InternalThread_Thread_free_internal (MonoInternalThre
...
+		gpointer new_old_val = mono_atomic_cas_ptr ((gpointer *)&this_obj->synch_cs, NULL /* set */, old_val /* compare */);
+		if (new_old_val == old_val)
+			break;
+		else
+			old_val = new_old_val;
+	}
+
+	if (old_val) {
+		MonoCoopMutex *synch_cs = (MonoCoopMutex *) old_val;
 		this_obj->synch_cs = NULL;
```
lambdageek: Why do you still need this assignment?  didn't the CAS just set `this->synch_cs` to NULL?

jaykrell:
This is not convincing.
Any number of threads can read the lock into a register and proceed toward the lock function, before you CAS it, and then they can each take the lock after you CAS it, interleaved arbitrarily with the freer.

Restated: Nulling the field does not affect in-progress uses of it, that have already read it.
They can take the lock after you CAS and lock and unlock, and before you free.

mono/metadata/threads.c
alexanderkyte: You're right

alexanderkyte:
@jaykrell I see what you're saying. Having pointed that out, there also needs to be memory fences at a couple of synchronization points in here. I am concerned with a design flaw here as well. The code that uses this will re-initialize a thread after it's been freed (ensure_synch_cs_set). If we actually have the above race, we aren't going to be freeing anything if our freeing thread comes before the racing thread. The free always has to be last. This motivates a work-stealing system, where the last writer to leave is the one that frees the unreferenced thread.

alexanderkyte:
The ABA problem occurs here too, where you can have two different locks being used to coordinate the same memory. This has a plain old data race.

jaykrell:
Right, I was going to point out the reinitialization case. Ultimately shutdown has to be very coordinated, to avoid reinit and race. Perhaps it already is. Except #7295 suggests otherwise?

alexanderkyte:
Yeah, I have observed three free-while-locked races so far. The flight recorder output made it clear what was happening. The race really is between VM shutdown and concurrent thread freeing. This isn't a debugger bug.

jaykrell:
Here is another angle: uninitialized and shutdown maybe have to look different.
1. Either threads can only proceed uninitialized => initialized => shutdown.
2. Or they can loop uninitialized => initialized => uninitialized.
2 can be valid but must be a deliberate decision.
If 2 is not valid, then, like, anywhere that initializes on demand, also needs to check for shutdown, and error out. Shutdown probably has a started/finished aspect as well -- you indicate it started atomically. This is maybe what the NT "rundown protection" locks are all about, that I never quite understood.

alexanderkyte:
This reinitialization bug has to be hiding a lot of managed threading bugs.

alexanderkyte:
`create_internal_thread_object` sets our synchronization (initializes it) and the free icall is called only from the finalizer for the object. I think that unless we want to start changing the CLR semantics of objects coming back to life, we should probably not reinit.

jaykrell:
Sorry if I implied either way -- I don't know.

alexanderkyte:
I guess the only answer is to make it correct and see if that breaks then

alexanderkyte:
@jaykrell @lambdageek @luhenry could I have a review of my new attempt to fix this?

mono/metadata/threads.c
```diff
@@ -95,7 +95,8 @@ extern int tkill (pid_t tid, int signal);
 
 #define SPIN_UNLOCK(i) i = 0
 
-#define LOCK_THREAD(thread) lock_thread((thread))
+#define TRY_LOCK_THREAD(thread) try_lock_thread((thread))
```
jaykrell: Macros need parens on the outside, but the parameter only needs one set, not two

mono/metadata/threads.c
```diff
@@ -453,7 +454,44 @@ thread_get_tid (MonoInternalThread *thread)
...
+// Freer:
+// - set finalized state
+// - check if any waiters
+// - if not, free. If so, let waiters finish.
+//
+// Note that the ordering of fetching the waiters and the freeing lock must be
+// inverted between the two to prevent state going wrong between the read and the validity check.
+//
+
+#define INTERNAL_THREAD_TOMBSTONE GINT_TO_POINTER (-1)
```
jaykrell: Need parens around  macro.

mono/metadata/threads.c
```diff
@@ -464,35 +502,122 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+
+static inline void
+increment_wait_queue (MonoInternalThread *thread)
+{
+	// We will have a pointer to the lock in our registers, need to warn
+	// finalizer to wait for us to finish.
+	mono_atomic_inc_i32 (&thread->num_waiters);
+}
+
 static inline void
```
jaykrell: No need to say inline. Let the compiler decide.

mono/metadata/threads.c
```diff
@@ -464,35 +502,122 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	MonoCoopMutex *lock = get_lock (thread, FALSE);
+	wait_on_lock (thread, lock);
 }
 
 static inline void
 unlock_thread (MonoInternalThread *thread)
 {
-	mono_coop_mutex_unlock (thread->synch_cs);
+	gpointer lock = mono_atomic_load_ptr ((volatile gpointer *) &thread->synch_cs);
+	if (G_LIKELY(lock != INTERNAL_THREAD_TOMBSTONE))
```
jaykrell: space after G_LIKELY

mono/metadata/threads.c
alexanderkyte: I'll fix the UNLOCK_THREAD beneath this as well then

mono/metadata/threads.c
```diff
@@ -95,7 +95,8 @@ extern int tkill (pid_t tid, int signal);
 
 #define SPIN_UNLOCK(i) i = 0
 
-#define LOCK_THREAD(thread) lock_thread((thread))
+#define TRY_LOCK_THREAD(thread) try_lock_thread((thread))
+#define ASSERT_LOCK_THREAD(thread) assert_lock_thread((thread))
```
jaykrell: Macros need one set of parens on the outside, and only one on params.

mono/metadata/threads.c
```diff
@@ -453,7 +454,44 @@ thread_get_tid (MonoInternalThread *thread)
...
+// partition usually seen in networks. The threads before the split will use one lock to
+// synchronize access to data, threads after will use another lock, and both will concurrently
+// dirty the same memory.
+//
+// Instead we needed to transition a thread from an initialized to a non-initializable state. Furthermore
+// we needed to ensure that locking attempts in progress finish before we free the lock. We do that by tracking
+// the number of waiters.
+//
+// Locker:
+// - inc num waiters
```
jaykrell: Feels like a lock around the lock or such.

alexanderkyte: More of a reference count

alexanderkyte: But a count of the number of threads with incoming pointers, not the number of objects

mono/metadata/threads.c
```diff
@@ -464,35 +502,122 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+}
+
+static gboolean
+try_lock_thread (MonoInternalThread *thread)
+{
+	MonoCoopMutex *lock = get_lock (thread, FALSE);
+	if (lock)
+		wait_on_lock (thread, lock);
+
+	return (lock != NULL);
```
jaykrell: no need for parens

mono/metadata/threads.c
alexanderkyte: The comments here should help clarify that for future readers: https://github.com/mono/mono/pull/9486/files#diff-f1a9287559d3aa620b058619f47f00c9R584

alexanderkyte:

configure.ac
```diff
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=16
+MONO_CORLIB_COUNTER=17
```
jaykrell: It is already 17 in master.
And I have 18 here https://github.com/mono/mono/pull/9503.
So please use 19.
Too bad it isn't an automatic hash of something. :(

jaykrell: I think whoever gets to master first, the other will have to rebuild monolite yet again. But we can proceed racing in parallel..

mcs/class/corlib/System.Threading/Thread.cs
```diff
@@ -89,6 +89,8 @@ sealed class InternalThread : CriticalFinalizerObject {
 		private IntPtr owned_mutex;
 		private IntPtr suspended_event;
 		private int self_suspended;
+		private int num_waiters;
+		private int freeing;
```
jaykrell: I'd kind of like to see byte used for booleans instead of int32.

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	MonoCoopMutex *synch_cs = thread->synch_cs;
+	thread->synch_cs = INTERNAL_THREAD_TOMBSTONE;
+	mono_memory_barrier ();
+
+	mono_coop_mutex_destroy (synch_cs);
+	g_free (synch_cs);
+}
+
+static void
+cleanup_thread_lock (MonoInternalThread *thread)
```
jaykrell: This isn't used.

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+static void
+wait_on_lock (MonoInternalThread *thread, MonoCoopMutex *lock)
+{
+	mono_coop_mutex_lock (lock);
+	decrement_wait_queue (thread);
+}
+
+static MonoCoopMutex *
+get_lock (MonoInternalThread *thread, gboolean assert_fine)
+{
```
jaykrell: Should this first check freeing, and return null?

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+		wait_on_lock (thread, lock);
+
+	return lock != NULL;
+}
 
-	mono_coop_mutex_lock (thread->synch_cs);
+static void
+assert_lock_thread (MonoInternalThread *thread)
+{
+	MonoCoopMutex *lock = get_lock (thread, FALSE);
```
jaykrell: No need for temporary..

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	wait_on_lock (thread, lock);
 }
 
-static inline void
+static void
```
jaykrell: How about the idea that the locker gets the lock and unlocks that lock?
Does that help? Otherwise it kinda looks like I can lock but then not unlock.

configure.ac
jaykrell: I've ceded 18. You can have it. I'm using 20 and 21.
But 17 is in master and stomping it here I think will break things.
Might need to make another PR to rebuild that.

jaykrell:
There are multiple things I find unconvincing here, so I tried to do it myself.

```
union lock_flags
{
	uintptr_t i;
	struct {
		uint count : sizeof (intptr_t) * 8 - 1;
		uint free : 1;
	};
};

gboolean
try_lock(thread)
{
	// inc count and read free
	while (1) {
		lock_flags flags;
		flags.i = atomic read thread->flags.i;
		if (flags.free)
			return false;
		lock_flags old_flags = flags;
		flags.count += 1;
		if (cas(&thread->flags.i, old_flags.i, flags.i) == old_flags.i)
			break;
	}

	// At this point it could be set to free, but we have
	// a count and we will lock and unlock. Nobody will
	// free out from under us.

	underlying_lock(thread->lock)
	return true
}

void
lock(thread)
{
	gboolean success = try_lock(thread);
	g_assert (success);
}

void
cleanup_lock(thread)
{
	if (!try_lock(thread))
		return;

	// free = 1 count unchanged
	while (1) {
		lock_flags flags;
		flags.i = atomic read thread->flags.i;
		if (flags.free)
			break;
		lock_flags old_flags = flags;
		flags.free = 1;
		if (cas(&thread->flags.i, old_flags.i, flags.i) == oldflags.i)
			break;
	}

	unlock(local_lock);
}

void
unlock (thread)
{
	// dec count and read free
	lock_flags flags;
	while (1) {
		flags.i = atomic read thread->flags.i;
		lock_flags old_flags = flags;
		g_assert (flags.count > 0);
		flags.count -= 1;
		if (cas(&thread->flags.i, old_flags.i, flags.i) == old_flags.i)
			break;
	}

	underlying_unlock (thread->lock);
	if (flags.free && flags.count == 0) {
		free (thread->lock);
		thread->lock = null; // or -1 or anything or leave it alone
	}
}
```

??

mcs/class/corlib/System.Threading/Thread.cs
lambdageek: Just use `bool` in C# and `mono_bool` in the matching struct in native.  (C# bool is 1 byte)

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+		internal_thread_free (thread);
+}
+
+static void
+decrement_wait_queue (MonoInternalThread *thread)
+{
+	gint32 waiters = mono_atomic_dec_i32 (&thread->num_waiters);
+	if (waiters == 0) {
+		gboolean freeing = mono_atomic_load_i32 (&thread->freeing);
+		if (freeing)
```
lambdageek: What guarantees that there are no new waiters at this point?

kumpera:
This looks like a big and important change. 
I'd like to take a good look at this PR and the comment, but I doubt I'll have time this week.
Could we hold merging it (if you decide so) until later next week?

mono/metadata/threads.c
alexanderkyte: The ordering of events ensures that new waiters will not be added after the thread enters "freeing" mode.

mono/metadata/threads.c
alexanderkyte: It already does? I'm not sure I follow. 
https://github.com/mono/mono/pull/9486/files#diff-f1a9287559d3aa620b058619f47f00c9R580

mono/metadata/threads.c
alexanderkyte: You can always unlock. 

```
-	mono_coop_mutex_unlock (thread->synch_cs);
+	gpointer lock = mono_atomic_load_ptr ((volatile gpointer *) &thread->synch_cs);
+	if (G_LIKELY (lock != INTERNAL_THREAD_TOMBSTONE))
+		mono_coop_mutex_unlock (lock);
```

If you unlock at the wrong time, you'll just exit silently

jaykrell:
I still find my prototype pseudo code more convincing.
I totally based it on yours, but I merged the count and the boolean freeing into one atomic variable. That way I believe there is never a need to temporarily increment waiters, only to decrement it. And some states/transitions of unclear correctness are thereby avoided.

jaykrell:
The point of the union is to get cas to work. So that you never inc waiters tentatively and backup off and dec it. Now, all incs of waiters stick -- the inc leads to a lock and unlock. Even if a free is being attempted at about the same time.

It doesn't change managed much or non-obviously -- just use an IntPtr there, to replace the two variables.

mono/metadata/threads.c
```diff
@@ -464,35 +502,123 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	gboolean freeing = mono_atomic_load_i32 (&thread->freeing);
+
+	if (G_UNLIKELY(freeing)) {
+		if (assert_fine)
+			g_error ("Attempting to lock a thread that has been finalized.");
+
+		// We lied, we actually are not going to be joining the queue
+		// of threads that have a reference to the lock and will attempt to
+		// use it.
+		decrement_wait_queue (thread);
```
jaykrell: This is the part that my union+cas removes.

jaykrell:
IntPtr maybe not -- it needs to be same size on all platforms?
Maybe use long managed, and add an int64_t to the union also.

Or decide that a 31 bit wait is enough and ints all around. I don't quickly buy that and feel more confident with more bits, and it is plenty cheap.

mono/metadata/threads.c
```diff
@@ -464,35 +502,122 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	wait_on_lock (thread, get_lock (thread, FALSE));
 }
 
-static inline void
+static void
 unlock_thread (MonoInternalThread *thread)
 {
-	mono_coop_mutex_unlock (thread->synch_cs);
+	gpointer lock = mono_atomic_load_ptr ((volatile gpointer *) &thread->synch_cs);
+	if (G_LIKELY (lock != INTERNAL_THREAD_TOMBSTONE))
```
jaykrell: This state doesn't occur in my proposal either. Every lock must be balanced by an unlock. The "waiter" count is kept incremented until unlock, rather than where you decrement it right after lock.

alexanderkyte: I was attempting to minimize visible changes in threading behavior. You're right though, changing a refcount at unlock time is more in alignment with expected behavior

jaykrell: I'm going for correctness here.

alexanderkyte:
@monojenkins build tarball

mono/metadata/object-internals.h
```diff
@@ -429,6 +429,16 @@ typedef enum {
...
+typedef struct {
+		gboolean freeing : 1; // Either 0x1 or 0x0
+		gint32 waiters : 31;
+} MonoInternalThreadLockFlags;
+
+typedef union {
+	gint32 placeholder;
+	MonoInternalThreadLockFlags value;
+} MonoInternalThreadLockFlagBox;
+
```
jaykrell:
```
typedef struct {
+		gboolean freeing : 1; // Either 0x1 or 0x0
+		gint32 waiters : 31;
+} MonoInternalThreadLockFlags;
+
+typedef union {
+	gint32 placeholder;
+	MonoInternalThreadLockFlags value;
+} MonoInternalThreadLockFlagBox;

This should be combined into:

union {
  gint32 i;
  struct {
    gboolean freeing : 1:
    gint32 waiters : 31;
} MonoInternalThreadLockFlags;

but I question if 31 bits is convincingly enough on a 64bit system.
And the uncertainty is not worth it.
So like I had, plus expansion for same-size on all systems.

native:

union {
  gsize i;
  gint64 same_size_on_all_systems;
  struct {
    gboolean freeing : 1:
    gsize waiters :  (sizeof (gsize) * 8) - 1;
} MonoInternalThreadLockFlags;


managed:
  long thread_lock_flags;
```

alexanderkyte: The combination makes it really unwieldy to work with the memory from the callers due to C's unwillingness to cast from an int to a struct.

mono/metadata/threads.c
```diff
@@ -464,35 +502,128 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+
+static void
+cleanup_thread_lock (MonoInternalThread *thread)
+{
+	mono_memory_barrier ();
+	g_assert (sizeof (MonoInternalThreadLockFlags) == sizeof (gint32));
+
+	gint32 flags = mono_atomic_load_i32 (&thread->lock_flags.placeholder);
+	MonoInternalThreadLockFlags *rich_flags = (MonoInternalThreadLockFlags *) &flags;
+
```
jaykrell: No.
```
MonoInternalLockFlags flags;
flags.i = mono_atomic_load_i32(&thread->lock_flags.i);
```

alexanderkyte: That's a really awkward interface. I guess I can do it like that.

mono/metadata/threads.c
```diff
@@ -464,35 +502,128 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+	while (rich_flags->freeing != TRUE) {
```
jaykrell: Do not compare to TRUE.
``while (flags.freeing)``` or ``while (!flags.freeing)```

mono/metadata/threads.c
```diff
@@ -464,35 +502,128 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+static void
+increment_wait_queue (MonoInternalThread *thread)
+{
+	return change_wait_queue (thread, 1);
+}
+
+static MonoCoopMutex *
+get_lock (MonoInternalThread *thread, gboolean assert_fine)
 {
+	increment_wait_queue (thread);
```
jaykrell: You need to see if it successfully incremented, or if it saw `freeing`.

mono/metadata/threads.c
```diff
@@ -464,35 +502,128 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+			flags = seen_old_flags;
+	}
+
+	if (rich_flags->waiters == 0 && rich_flags->freeing)
+		internal_thread_free (thread);
+}
+
+static void
+decrement_wait_queue (MonoInternalThread *thread)
+{
```
jaykrell: C does not allow `return` with `void`. This won't compile. C++ does though.

mono/metadata/object-internals.h
jaykrell: No it doesn't. I guess I have not been clear enough.
```
MonoInternalThreadLockFlags flags;
flags.freeing = true
flags.i  = cas(&thread->flags.i);
```

etc.

mono/metadata/threads.c
jaykrell: It is not more awkward than pointers and casting.

alexanderkyte: Casting is visually a side-effect-free operation while assigning to locals to me feels like it should be more than just a value unwrapping step. If this were going to be used elsewhere, I would wrap it with a macro or something, but this change only impacts these 3 functions so it's fine.

alexanderkyte: Picture how too many of these in a large method would completely destroy the ability of someone quickly scanning through the code to figure out what inputs and external state determines the control flow

jaykrell:
Look again at my psuedo code. It is pretty close to how this should be, if agree on the algorithm.
Using int diff to share code with +1 and -1 is good. You could also pass in whether or not to set free, so there is just one function that loops until it can make a change, or stops when free is true. (but be careful, as unlock should decrement even if it sees freeing, I think, if that is possibl -- well, freeing is set with the lock held actually, and it is an exclusive lock..).

mono/metadata/threads.c
jaykrell: There is union and atomic operations, locks, and threads, plenty of warning signs.

jaykrell:
Look at the threadpool for similar:

```
typedef union {
	struct {
		gint16 starting; /* starting, but not yet in worker_callback */
		gint16 working; /* executing worker_callback */
	} _;
	gint32 as_gint32;
} ThreadPoolCounter;

#define COUNTER_ATOMIC(var,block) \
	do { \
		ThreadPoolCounter __old; \
		do { \
			(var) = __old = COUNTER_READ (); \
			{ block; } \
			if (!(counter._.starting >= 0)) \
				g_error ("%s: counter._.starting = %d, but should be >= 0", __func__, counter._.starting); \
			if (!(counter._.working >= 0)) \
				g_error ("%s: counter._.working = %d, but should be >= 0", __func__, counter._.working); \
		} while (mono_atomic_cas_i32 (&threadpool.counters.as_gint32, (var).as_gint32, __old.as_gint32) != __old.as_gint32); \
	} while (0)
```

alexanderkyte:
@monojenkins build tarball

jaykrell:
But you can write a function instead of a macro, as you did, good.
Point means -- people do this. It is little ugly but it buys correctness and reasonable simplicity.

mono/metadata/threads.c
```diff
@@ -464,35 +502,133 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+			break;
+		else
+			flags.placeholder = seen_old_flags.placeholder;
+	}
+
+	if (flags.waiters == 0 && flags.freeing)
+		internal_thread_free (thread);
+	else if (!thread->synch_cs)
 		ensure_synch_cs_set (thread);
+}
```
jaykrell: You should maybe assert that diff is +1 or -1 and the logic at the end should vary based on it.
Only check waiting/freeing if diff==-1, only check sync_cs if diff==+1.
It is probably correct asis though.

This still seems off. In my proposal I decremened and set free at the same time.
I did it inside the lock, then unlocked, then freed the lock.
Perhaps other orders are ok.
I only had one call to internal_thread_free though which seems better.

I would name `placeholder` as `i` or `integer` or `atomic_integer` perhaps.
People often calls AsUint32 or AsSizeT -- i.e. its type. It is the "union views as typeX".

alexanderkyte: If you decrement before you unlock the lock then can't someone else free the lock before we unlock?

jaykrell: Ok that last comment wasn't entirely  accurate.
It looks like I split setting free and decrement.
Still the structure here seems a little different.

alexanderkyte:
@monojenkins build tarball

alexanderkyte:

configure.ac
```diff
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=24
+MONO_CORLIB_COUNTER=25
```
jaykrell: This is out  of date again and damaging other PRs.

alexanderkyte:
@monojenkins build failed

alexanderkyte:
@monojenkins build tarball

alexanderkyte:

alexanderkyte:
@monojenkins build failed

alexanderkyte:
@luhenry recommended that I factor out this logic into an unmanaged "safe handle" in our last 1-on-1. I am going to do that here.

alexanderkyte:

alexanderkyte:
@monojenkins build failed

- seems to have not picked up the tarball

mono/utils/mono-safe-handle.c
```diff
@@ -0,0 +1,167 @@
...
+
+static MonoCoopMutex *
+get_lock (MonoSafeHandleNative *handle, gboolean assert_fine)
+{
+	increment_wait_queue (handle);
+
+	MonoCoopMutex *ptr = (MonoCoopMutex *) mono_atomic_load_ptr ((volatile gpointer *) &handle->synch_cs);
+
+	gboolean fine = ptr && (ptr != INTERNAL_HANDLE_TOMBSTONE);
+	if (assert_fine && !fine)
```
jaykrell: These two lines can be combined -- i.e. just `g_assert` or `g_assertf`.

mono/utils/mono-safe-handle.h
```diff
@@ -0,0 +1,42 @@
...
+ * (C) 2018 Microsoft, Inc.
+ *
+ */
+#ifndef __MONO_UTILS_SAFE_HANDLE__
+#define __MONO_UTILS_SAFE_HANDLE__
+
+#include "mono/utils/mono-coop-mutex.h"
+
+typedef union {
+	gint64 mem_start;
```
jaykrell: I'd call it i or integer or as_int or asint.
AsInt32 and such is the pattern in other projects

Why did we go with gint64 vs. gsize? Want same size across BCL builds? Or just easier to reason about fixed size? Or worried about 32bit overflow on 32bit system?

jaykrell: In particular 64bit CAS operations are cheap on modern 32bit x86, but what about 32bit ARM?

mono/utils/mono-safe-handle.c
```diff
@@ -0,0 +1,167 @@
...
+{
+	MonoCoopMutex *lock = (MonoCoopMutex *) mono_atomic_load_ptr ((volatile gpointer *) &handle->synch_cs);
+	mono_coop_mutex_unlock (lock);
+
+	// Now the lock can be freed
+	decrement_wait_queue (handle);
+
+	mono_memory_barrier ();
+	MonoSafeHandleNativeState flags;
+	flags.mem_start = mono_atomic_load_i64 (&handle->lock_flags.mem_start);
```
jaykrell: In my version, which is failing assert for some reason, I put this read inside the lock. Which I still feel more comfortable about. How can multiple threads not see this result concurrently?

mono/utils/mono-safe-handle.c
```diff
@@ -0,0 +1,167 @@
...
+ensure_synch_cs_set (MonoSafeHandleNative *handle)
+{
+	MonoCoopMutex *synch_cs;
+
+	if (handle->synch_cs != NULL) {
+		return;
+	}
+
+	synch_cs = g_new0 (MonoCoopMutex, 1);
+	mono_coop_mutex_init_recursive (synch_cs);
```
jaykrell: Need barrier after `mono_coop_mutex_init_recursive`, so its affects are done before the atomic_cas_ptr? Or is one or the other already a full barrier?

In either case..I have to ask in slack...there is a theory that there are two missing barriers here, but maybe not.

jaykrell: I thought about it -- I suggest a barrier between the init and the cas.
So the init will have finished before the cas, so another thread doesn't see a pointer that isn't initialized.

jaykrell: Er, and before the return, to avoid prefetching from lock. Let me double  check.

mono/utils/mono-safe-handle.c
jaykrell: This is maybe protected by atomic swap in of tombstone?

mono/metadata/object-internals.h
```diff
@@ -451,7 +452,9 @@ struct _MonoInternalThread {
 	gpointer appdomain_refs;
 	/* This is modified using atomic ops, so keep it a gint32 */
 	gint32 __interruption_requested;
-	MonoCoopMutex *synch_cs;
+	MonoSafeHandleNative synch_handle;
+	//MonoCoopMutex *synch_cs;
```
kumpera: Dead comments, should be removed.

mcs/class/corlib/System.Threading/Thread.cs
```diff
@@ -72,6 +72,7 @@ sealed class InternalThread : CriticalFinalizerObject {
 		private IntPtr appdomain_refs;
 		private int interruption_requested;
 		private IntPtr synch_cs;
```
kumpera: SyncCS is gone in the native struct.

mono/metadata/threads.c
```diff
@@ -907,14 +870,11 @@ mono_thread_detach_internal (MonoInternalThread *thread)
...
 	 * The shutting_down flag is not always set, so we can't assert on it.
 	 */
-	if (thread->synch_cs)
-		LOCK_THREAD (thread);
-
-	thread->state |= ThreadState_Stopped;
-	thread->state &= ~ThreadState_Background;
-
-	if (thread->synch_cs)
+	if (TRY_LOCK_THREAD (thread)) {
```
kumpera: This looks wrong, the thread state must be always updated.

mono/utils/mono-safe-handle.c
```diff
@@ -0,0 +1,167 @@
...
+}
+
+static void
+change_wait_queue (MonoSafeHandleNative *handle, int diff)
+{
+	g_assert (diff == 1 || diff == -1);
+
+	mono_memory_barrier ();
+	MonoSafeHandleNativeState flags;
+	flags.mem_start = mono_atomic_load_i64 (&handle->lock_flags.mem_start);
```
kumpera: The runtime cannot use 64bits cas.

jaykrell: That was my point -- gsize/ptr should be ok though, right? i.e. 64bit is ok on 64bit, right? 32bit on any.

mono/metadata/threads.c
jaykrell: I was kinda wondering about try vs. assert also, in my attempt I asserted all and was going to see where it failed.

jaykrell:
What is a reference queue?

kumpera:
It's a GC data structure that correctly deals with unreachability.
https://github.com/mono/mono/blob/master/mono/metadata/object.h#L373

PR #9487, opened 2018-07-10
cominterop cleanup ahead of coop conversion.

jaykrell:


mono/metadata/cominterop.c
```diff
@@ -573,22 +573,16 @@ cominterop_get_interface_checked (MonoComObject* obj, MonoClass* ic, MonoError *
  * Returns: the COM interface requested
  */
 static gpointer
-cominterop_get_interface (MonoComObject *obj, MonoClass *ic, gboolean throw_exception)
+cominterop_get_interface (MonoComObject *obj, MonoClass *ic)
```
lambdageek: (bikeshed) rename to `ves_icall_cominterop_get_interface` so it's a little bit easier to tell that it's a JIT icall

jaykrell:
The const lets us remove a cast which is good.

jaykrell:
ok, I might put in more.. or less.. the whitespace is maybe large/obnoxious. The rest is small/reasonable. I wasn't going to commit w/ just the earlier approval, unless I rebase back to that, which for overnight might be a good idea.

mono/metadata/cominterop.c
```diff
@@ -573,22 +626,16 @@ cominterop_get_interface_checked (MonoComObject* obj, MonoClass* ic, MonoError *
...
-			mono_error_set_pending_exception (error);
-			return NULL;
-		} else {
-			mono_error_cleanup (error);
-		}
-	}
-
-	if (throw_exception)
+		mono_error_set_pending_exception (error);
+		itf = NULL;
```
jaykrell: This is incorrect or at least unclear to me now.
It should use goto exit, and then leave alone the if (throw_exception) g_assert(itf).
Point being -- single point of exit, because we will have function enter/exit, but don't change more than that.

mono/metadata/cominterop.c
```diff
@@ -3522,10 +3500,10 @@ mono_marshal_safearray_set_value (gpointer safearray, gpointer indices, gpointer
 static void
 mono_marshal_safearray_set_value (gpointer safearray, gpointer indices, gpointer value)
```
lambdageek: Isn't there a one-line difference between the WIN32 and non-WIN32 version of this function? As long as you're changing it, maybe pull the ifdef into the body so that the common parts are shared?

mono/metadata/cominterop.c
```diff
@@ -3681,7 +3635,7 @@ ves_icall_System_Runtime_InteropServices_Marshal_PtrToStringBSTR (gpointer ptr)
 gpointer
 ves_icall_System_Runtime_InteropServices_Marshal_StringToBSTR (MonoString* ptr)
 {
-	return mono_string_to_bstr(ptr);
+	return mono_string_to_bstr (ptr);
```
lambdageek: This is silly. why do we bother doing `register_icall (mono_string_to_bstr, ...)` when we already have a perfectly good icall that does exactly the same thing.

It would be nice if we could make `mono_string_to_bstr` just a utility function and replace uses of it as an icall by this one.

Is there some `#ifdef` logic that i'm missing why that's not possible?

jaykrell:
I have a lot of cleanup but should probably take this first set of steps first. I’ll undo the whitespace & address the other stuff.

mono/metadata/cominterop.c
```diff
@@ -352,11 +406,11 @@ cominterop_get_com_slot_begin (MonoClass* klass)
 static MonoClass*
 cominterop_get_method_interface (MonoMethod* method)
 {
-	ERROR_DECL (error);
```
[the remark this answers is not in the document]
jaykrell: We do scope down ERROR_DECL a fair amount but I've undone it for now.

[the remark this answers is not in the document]
jaykrell: The lack of ves_icall_ prefixes on JIT calls is widespread also, I'm afraid.

Very small part of it:```
	register_icall (cominterop_get_method_interface, "cominterop_get_method_interface", "ptr ptr", FALSE);
	register_icall (cominterop_get_function_pointer, "cominterop_get_function_pointer", "ptr ptr int32", FALSE);
	register_icall (cominterop_object_is_rcw, "cominterop_object_is_rcw", "int32 object", FALSE);
	register_icall (cominterop_get_ccw, "cominterop_get_ccw", "ptr object ptr", FALSE);
	register_icall (cominterop_get_ccw_object, "cominterop_get_ccw_object", "object ptr int32", FALSE);
	register_icall (cominterop_get_hresult_for_exception, "cominterop_get_hresult_for_exception", "int32 object", FALSE);
	register_icall (cominterop_get_interface, "cominterop_get_interface", "ptr object ptr", FALSE);

	register_icall (mono_string_to_bstr, "mono_string_to_bstr", "ptr obj", FALSE);
	register_icall (mono_string_from_bstr_icall, "mono_string_from_bstr_icall", "obj ptr", FALSE);
	register_icall (mono_free_bstr, "mono_free_bstr", "void ptr", FALSE);
	register_icall (cominterop_type_from_handle, "cominterop_type_from_handle", "object ptr", FALSE);

	/* SAFEARRAY marshalling */
	register_icall (mono_marshal_safearray_begin, "mono_marshal_safearray_begin", "int32 ptr ptr ptr ptr ptr int32", FALSE);
	register_icall (mono_marshal_safearray_get_value, "mono_marshal_safearray_get_value", "ptr ptr ptr", FALSE);
	register_icall (mono_marshal_safearray_next, "mono_marshal_safearray_next", "int32 ptr ptr", FALSE);
	register_icall (mono_marshal_safearray_end, "mono_marshal_safearray_end", "void ptr ptr", FALSE);
	register_icall (mono_marshal_safearray_create, "mono_marshal_safearray_create", "int32 object ptr ptr ptr", FALSE);
	register_icall (mono_marshal_safearray_set_value, "mono_marshal_safearray_set_value", "void ptr ptr ptr", FALSE);
	register_icall (mono_marshal_safearray_free_indices, "mono_marshal_safearray_free_indices", "void ptr", FALSE);
```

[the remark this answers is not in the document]
lambdageek: @jaykrell Sure. And there's no need to gratuitously rename them.  But if you're already changing a function, might as well fixup the name.

mono/metadata/cominterop.c
jaykrell: Good catch.

jaykrell: Sort of.

jaykrell:

mono/metadata/cominterop.c
jaykrell: Sorry, no renames here yet. Given change of interface though..not a bad idea.

I'd like icall-def and jit to be named differently, I think.
Because jit is apparently stuck in old style.
mono_jit_icall_foo ?

I'd suggest every single JIT icall get such a name, even if it is just one line to call some other function, i.e. when there are no pointers.

If people are worried about the cost of the (tail) call, which is legitimate given elf semantics and cross-module calls, use a define, or an inline in a header, yes.

jaykrell:
CI passes.

PR #9488, opened 2018-07-10
[System]: Fix `mono_btls_x509_get_public_key()` implementation.

baulig:
* Copy corlib's `X509CertificateTest.Certificate8` into System, call it
  `Corlib_Test_Certificate8` and use `X509Certificate2`.

The test in corlib does not expose the problem because it would use the
fallback implementation.


mcs/class/System/Test/System.Security.Cryptography.X509Certificates/X509Certificate2Test.cs
```diff
@@ -41,6 +43,23 @@
 namespace MonoTests.System.Security.Cryptography.X509Certificates {
 	[TestFixture]
 	public class X509Certificate2Test {
+
+		CultureInfo oldcult;
```
marek-safar: Use http://nunit.org/docs/2.4.8/setCulture.html instead

mcs/class/System/Test/System.Security.Cryptography.X509Certificates/X509Certificate2Test.cs
```diff
@@ -1686,6 +1705,81 @@ public void T61String ()
...
+			//Assert.AreEqual (tsa, x509.ToString (true), "ToString(true)");
+			Assert.AreEqual ("CN=Armed Forces Root, OU=DoD, O=U.S. Government, C=US", x509.Issuer, "Issuer");
+			Assert.AreEqual ("CN=Armed Forces Root, OU=DoD, O=U.S. Government, C=US", x509.Subject, "Subject");
+		}
+
+		string ToString (byte[] data)
+		{
+			StringBuilder sb = new StringBuilder ();
+			for (int i = 0; i < data.Length; i++)
+				sb.Append (data[i].ToString ("X2"));
```
Therzok: `sb.AppendFormat ("{0:X2}", data[i])`? Not sure if `AppendFormat` just does `string.Format` before appending, but it might not allocate a string for each hex byte.

mcs/class/System/Test/System.Security.Cryptography.X509Certificates/X509Certificate2Test.cs
baulig: Interesting!  That must be a new addition - all the tests in corlib are using the setup/teardown mechanic.  Just fixed it.

mcs/class/System/Test/System.Security.Cryptography.X509Certificates/X509Certificate2Test.cs
Therzok: Ah, didn't see it was a test file. 👍

nealef:
A client has a program that fails in 5.16 but works with the nightly preview code:

System.Net.WebException: Error: TrustFailure (A call to SSPI failed, see inner exception.) ---> System.Security.Authentication.AuthenticationException: A call to SSPI failed, see inner exception. ---> Mono.Btls.MonoBtlsException: Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED

I wonder if this fix would help. I can't see anything else that may explain the failure at 5.16 but working with 5.17.

PR #9491, opened 2018-07-11
Fix race conditions in init_com_provider_ms.

jaykrell:


mono/metadata/cominterop.c
```diff
@@ -2685,8 +2685,12 @@ init_com_provider_ms (void)
 	MonoDl *module = NULL;
 	const char* scope = "liboleaut32.so";
 
-	if (initialized)
+	if (initialized) {
+		// Barrier here prevents reads of sys_alloc_string_len_ms etc.
+		// from being reordered before initialized.
+		mono_memory_barrier ();
```
lewurm: I don't understand why this barrier is needed. the other one looks okay

jaykrell: The comment explains it. Barriers come in pairs usually.
Without this one, the globals protected by initialized can be read before it, and before it is written.

lewurm: if `initialized` is true I don't get why you care about preventing reads for those globals. should this barrier go above the first read of `initialized`?

jaykrell: Without a barrier, the reads of other relevant variables can be reordered to before `initialized` is true. The barrier must go after reading initialized, so that reads of others are not moved earlier.

These do not have to be full barriers though.

It goes like this:

```
use_foo()
{
init_foo();
foo(); 
}

init_foo()
{
 if (initialized)
   return;
 foo = ...
barrier(); // new but insufficient 
initialized = true;
}
```

but `use_foo` can read `foo` before it is written, unless there is the second barrier.

It could as well have been written:

```
use_foo()
{
local_foo = foo;
init_foo();
local_foo(); 
}
```

A barrier amidst the writes does not force an ordering of the reads.

jaykrell: A barrier before reading initialized does not help. That still allows all the reads to be reordered among each other.

There are other solutions that aren't as slow.

pthread_once might be a little faster, might, might not.

C++11 static initializers use thread locals under the covers to avoid barriers in the fast path on ARM. The underlying implementation is quite tricky, but all the runtimes have it now, and the way to use it is very easy.

C11 atomics might also help.

jaykrell:
See here https://github.com/mono/mono/pull/9419

mono/metadata/cominterop.c
lewurm: I see, that makes sense. Thanks for the explanation!

jaykrell:
@monojenkins build failed

PR #9494, opened 2018-07-11
Avoid thread local access in non-error case.

jaykrell:


jaykrell:
@monojenkins build failed

PR #9495, opened 2018-07-11
ves_icall_System_Threading_Thread_ResetAbort vs. mono_thread_internal_reset_abort.

jaykrell:
ves_icall_System_Threading_Thread_ResetAbort vs. mono_thread_internal_reset_abort?

Expand locking in first to resemble second.
Add comments wondering about their differences.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -2553,6 +2553,9 @@ mono_thread_internal_abort (MonoInternalThread *thread, gboolean appdomain_unloa
 void
 ves_icall_System_Threading_Thread_ResetAbort (MonoThreadObjectHandle this_obj, MonoError *error)
 {
+	// FIXME Why are ves_icall_System_Threading_Thread_ResetAbort and mono_thread_internal_reset_abort so different?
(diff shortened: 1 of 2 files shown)


mono/metadata/threads.c
```diff
@@ -2553,6 +2553,9 @@ mono_thread_internal_abort (MonoInternalThread *thread, gboolean appdomain_unloa
 void
 ves_icall_System_Threading_Thread_ResetAbort (MonoThreadObjectHandle this_obj, MonoError *error)
 {
+	// FIXME Why are ves_icall_System_Threading_Thread_ResetAbort and mono_thread_internal_reset_abort so different?
+	// FIXME? this_obj is unused but passed by reference source.
```
lambdageek: `System.Threading.Thread.ResetAbortNative` is private so we could make it static, I suppose...

jaykrell:
Should native ignore appdomain abort?
Should managed check `if (thread->abort_exc)`?
Should managed extend its locking to more resemble native? (Or share that code?)

jaykrell:

PR #9517, opened 2018-07-12
3/3: Remove System.Diagnostics.Private namespace.

filipnavara:
Depends on https://github.com/mono/corert/pull/18 and https://github.com/mono/corefx/pull/96. As long as the implementation of System.Diagnostics.Debug is private (not public, not internal) in corlib there should be no namespace or class conflicts.

### DIFF
--- a/mcs/class/corlib/corert/Debug.cs
+++ b/mcs/class/corlib/corert/Debug.cs
@@ -1,4 +1,4 @@
-namespace System.Diagnostics.Private
+namespace System.Diagnostics
 {
 	static partial class Debug
 	{


marek-safar:
There will be because we have IVT to System

filipnavara:
IVT applies only to internal classes.

filipnavara:
I thought IVT applies only to internal classes. This one is not internal. I tried a full build on macOS and it succeeded.

filipnavara:
Eh, spoke too soon... I tried clean rebuild and it failed somewhere :-/

filipnavara:
NVM, this won't work.

filipnavara:
Interestingly it didn't fail to build mscorlib and System, but it failed for System.Core. Essentially it fails for assemblies that both link System and at the same time are IVT of mscorlib. It would be possible to remove IVT from mscorlib towards System.Core, but I don't like that since it's likely to bite back in the future (and this IVT exists on NetFX too).

There's a way to make the compiler happy by linking against a fake facade assembly generated from mscorlib, but that sounds quite clunky too.

Issue #8340, opened 2018-04-19
Issue with black text on console, possibly ncurses


grbd:
Hi,
I've recently noticed an issue with mono producing black text when running things like nuget
I'm using the latest version of gentoo and I suspect this might be related to ncurses

## Steps to Reproduce

Let's say I try to run the latest nuget with an invalid option
```
nuget --invalidoption 
```
This should show a single line of red text giving an error
followed by a stack trace, but what I actually end up with is just a load of unreadable lines of black

To get around this I need to prefix the command 
```
TERM=xterm nuget --invalidoption 
```

I think dotnet core had a similar problem with https://github.com/dotnet/corefx/issues/26966
I thought this might be fixed with https://github.com/mono/mono/issues/6752
but that doesn't seem to have fixed the problem

### On which platforms did you notice this

  * [* ] Linux - gentoo latest
  * **Version Used**: mono 5.10.1.48

kasper3:
Fot donet core, maybe try `dotnet nuget --invalidoption`

PiotrZierhoffer:
I can confirm this on mono 5.12.0.226. Unfortunately it makes console applications that use colors in their output quite unusable.

While this is not fixed, does anyone know if it's safe to set TERM=xterm by default on modern Linux distros? Are you aware of any distributions that lack this config by default?

lewurm:
@PiotrZierhoffer setting `TERM=xterm` is safe for pretty much any distribution out there.

PR #9520, opened 2018-07-12
[TermInfo] fix reading integer value for new file format

lewurm:
16bit vs. 32bit only applies for offsets, not for actual entries in the
file format.

Fixes https://github.com/mono/mono/issues/8340

before / after:
screen shot 2018-07-12 at 19 08 19

### DIFF
--- a/mcs/class/corlib/System/TermInfoReader.cs
+++ b/mcs/class/corlib/System/TermInfoReader.cs
@@ -173,7 +173,7 @@ namespace System {
 				offset++;
 
 			offset += ((int) number) * intOffset;
-			return GetInteger (buffer, offset);
+			return GetInt16 (buffer, offset);
 		}
 
 		public string Get (TermInfoStrings tstr)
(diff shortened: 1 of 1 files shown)


lewurm:
@monojenkins build failed

lewurm:
@monojenkins build failed

luhenry:
Any relatively **easy** way we can have a test for that?

lewurm:
@luhenry we could do something similar as corefx: put terminfo files in the repository and just try to parse them (without exception). That would be an regression test for the initial fix; I'm not sure how to test this regression _easily_. I can add mentioned regression test later if desirable.

akoeplinger:
Let's do this in a followup PR since we need to backport this ;)

lewurm:
we probably can reuse those tests with some modifications: https://github.com/dotnet/corefx/pull/27109/files#diff-90eedc08931731780b901ec0454101deR61

lewurm:
@monojenkins backport 2018-06

lewurm:
@monojenkins backport 2018-04

lewurm:
@monojenkins backport 2018-02

lewurm:
@monojenkins backport 2017-12

PR #9521, opened 2018-07-12
Fix DriveInfo.IsReady always returning true

mderoy:
We received a bug report, that driveinfo.isready would return true when the user did not have a disk in their CD drive, thus causing them an I/O exception when they go to check the free space of the drive. After investigating, I found this MonoTODO, and an unimplemented IsReady function. I've applied the same logic as CoreFX does. Checking that the directory of the drive exists. This means when there is no cd in the drive, isready will return false, but if there is a cd in the drive, isready will return true.

coreFX code v
https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.DriveInfo/src/System/IO/DriveInfo.cs


marek-safar:
@monojenkins build failed

akoeplinger:
@monojenkins commit apidiff

PR #9523, opened 2018-07-12
[mmap] flock() the image that we want to open

lambdageek:
Try to fail eagerly if someone writes over an assembly that we're going to mmap.

Update: this isn't going to work.  advisory locks don't prevent someone from doing anything if they aren't also following the locking protocol.

### DIFF
--- a/mono/utils/mono-filemap.c
+++ b/mono/utils/mono-filemap.c
@@ -20,6 +20,7 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
+#include <sys/file.h>
 #include <stdlib.h>
 #include <stdio.h>
 
--- a/mono/utils/mono-filemap.c
+++ b/mono/utils/mono-filemap.c
@@ -72,7 +72,7 @@ mono_file_map_close (MonoFileMap *fmap)
 #ifdef WIN32
 	return fclose ((FILE*)fmap);
 #else
-	flock (fd, LOCK_UN);
+	flock (mono_file_map_fd (fmap), LOCK_UN);
 	return close (mono_file_map_fd (fmap));
 #endif
 }
--- a/mono/utils/mono-filemap.c
+++ b/mono/utils/mono-filemap.c
@@ -14,6 +14,7 @@
 #if HAVE_SYS_STAT_H
 #include <sys/stat.h>
 #endif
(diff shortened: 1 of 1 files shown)


lambdageek:
**DO NOT MERGE**
Doing an experiment.

lambdageek:
@monojenkins build deb

mono/utils/mono-filemap.c
```diff
@@ -40,6 +42,8 @@ mono_file_map_open (const char* name)
 	int fd = open (name, O_RDONLY);
 	if (fd < 0)
 		return NULL;
+	if (flock (fd, LOCK_EX) < 0)
```
jaykrell: Reading a file should only get shared access.

lambdageek:
Did a bunch of reading. An advisory lock is apparently the wrong thing for this anyway. If we really want to bail out in a nice way if someone modifies an assembly that we already `mmap`ed, this isn't going to be a viable approach.

migueldeicaza:
What is the purpose of this patch?   I don’t understand what this is trying to
Solve.

lambdageek:
@migueldeicaza 
So when Mono loads an assembly, it will use `mmap()` to map it into the process' memory and then the `MonoImage` will have pointers into that memory region.  Things like `MonoImage:raw_metadata` or `MonoImage:string_heap` are just pointers into the mapped pages.

On Linux, but not on Mac, we've observed that if Mono loaded some assembly (/path/Foo.dll) and then that file is modified, we end up with a `MonoImage` with bad pointers because it seems the pages are unmapped.

The effect in Mono is random crashes when we happen to access something in the `MonoImage` - for example if we need to access the assembly name (which is a pointer into the string heap of the MonoImage) for example if we're iterating over all the loaded assemblies of an AppDomain to check if we've loaded something before or not.

What we would like (which this patch doesn't do correctly): some way for Mono either to prevent or to be notified when another process is modifying a loaded assembly so that we can fail in a predictable way (ideally with a good error message) instead of randomly segfaulting in unrelated code.

lambdageek:
@migueldeicaza The next thing I was going to try was to use inotify to listen for file modifications for every loaded assembly.

lambdageek:
This wasn't the right approach. An advisory lock won't do.

PR #9528, opened 2018-07-13
When csc fails, for any reason, retry once with -parallel-.

jaykrell:
This might give us higher ARM CI success rates, and might be much slower.
This will double up actual errors for incorrect C# source.

### DIFF
--- a/mcs/build/executable.make
+++ b/mcs/build/executable.make
@@ -138,8 +138,14 @@ $(topdir)/class/lib/$(PROFILE)/.stamp: | $(topdir)/class/lib/$(PROFILE)-$(HOST_P
 	$(if $(filter $(HOST_PLATFORM),$(BUILD_PLATFORM)),$(if $(filter $(BUILD_PLATFORM),win32),CYGWIN=winsymlinks:nativestrict) ln -s $(abspath $(topdir)/class/lib/$(PROFILE)-$(BUILD_PLATFORM)) $(abspath $(topdir)/class/lib/$(PROFILE)))
 endif
 
+# If csc fails, try again not parallel.
 $(build_lib): $(BUILT_SOURCES) $(EXTRA_SOURCES) $(response) $(build_libdir:=/.stamp)
+ifndef CSC_RETRY_SERIAL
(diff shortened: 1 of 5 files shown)


jaykrell:
ARM64 CI failure rate is quite high lately.

For example:

https://github.com/mono/mono/pull/9527/files

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64-fullaot/1871/parsed_console/log.html

https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-aarch64/6259/parsed_console/log.html

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64/11640/parsed_console/log.html

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64-fullaot/1801/parsed_console/log.html

jaykrell:
It *appears* to have retried.
It *appears* the failure is in cil-stringreplacer.exe, not csc.exe.
But I could certainly be wrong about anything.
The switches are a little confusing: -parallel or -parallel- or -parallel+? I used -parallel-.

jaykrell:
This does not address the current severe problems, but might still be reasonable.
See also https://github.com/mono/mono/pull/9529.

marek-safar:
I know it's a hack but perhaps we should just always run in -parallel- mode on ARM for now

@akoeplinger thoughts?

jaykrell:
It is reportedly quite slow which is why I put it in only upon failure.

jaykrell:
The gsharedvt comment is meant for elsewhere, an issue, will move later.

jaykrell:
Thinking about this more, is the failure path in csc to reliably crash, or could it appear to succeed but generate different/incorrect code?

akoeplinger:
@marek-safar I have no objection to running with parallel- always for ARM, but we should measure the time increase.

Are we confident the issue is fixed by `parallel-` ?

lewurm:
I don't have any numbers handy, but running with `parallel-` is _significantly_ slower. I wouldn't recommend that.

jaykrell:
The underlying bug is fixed for x86/amd64, but we still fail, because:
 1. ARM memory model.
 2. Using system mono.

marek-safar:
I'm closing this one as this is not a real solution, we cannot say everyone including other teams to keep compiling until it works.

Issue #8731, opened 2018-05-16
build failure on arm64 chromebook


jaykrell:
## Steps to Reproduce

1. ARM64 Acer Chromebook R 13, 4GB RAM, 32GB storage -- under $400 at Amazon
2. Install Ubuntu Xenial chroot with "crouton"; easier done than said
3. git clone
4. configure
5. make

### Current Behavior

It always null derefs building mscorlib.dll. Despite many retries.

### Expected Behavior

Build and pass all tests, repeatedly.

## On which platforms did you notice this

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:

Head master from around May 15 2018, and also a few weeks ago.
Since I was cleaning up method_to_ir I got an idea where to put a fix/workaround.

## Stacktrace

I didn't save it sorry.

Overkill workaround:

I realize this is overkill. I can try to use the stlr instruction instead of dmb. On my arm64 chromebook, building mscorlib.dll would fail 100% with quite a number of tries.

This "fixes" it:

```(xenial)jay@localhost:/s/mono$ git diff
diff --git a/mono/mini/method-to-ir.c b/mono/mini/method-to-ir.c
index 9ff4dc8..42c5c55 100644
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -8681,7 +8681,7 @@ calli_end:
                        CHECK_STACK (2);
                        sp -= 2;
 
-                       if (ins_flag & MONO_INST_VOLATILE) {
+                       if (il_op == CEE_STIND_REF) {
                                /* Volatile stores have release semantics, see 12.6.7 in Ecma 335 */
                                mini_emit_memory_barrier (cfg, MONO_MEMORY_BARRIER_REL);
                        }
@@ -9461,6 +9461,9 @@ calli_end:
                        MonoInst *store_val = NULL;
                        MonoInst *thread_ins;
 
+                        if (il_op == CEE_STFLD || il_op == CEE_STSFLD)
+                                mini_emit_memory_barrier (cfg, MONO_MEMORY_BARRIER_REL);
+
                        is_instance = (il_op == CEE_LDFLD || il_op == CEE_LDFLDA || il_op == CEE_STFLD);
                        if (is_instance) {
                                if (il_op == CEE_STFLD) {
```

Better will be to try the stlr instruction whenever storing a reference.
And maybe optimize some of them away. I'll try.

marek-safar:
Overkill workaround:

[...]

PR #9529, opened 2018-07-13
Update ARM memory model.

jaykrell:
This has been seen to fix the test case in https://github.com/mono/mono/issues/8731 on ARM64.
I didn't try ARM32. Here it is duplicated:

```
using System;
using System.Threading;

class A
{
class obj { public int i; }

static int N = 100 * 100 * 100;
static obj[] reserve = new obj [N];
static obj[] ready = new obj [N];
static long writes;
static volatile int barrier;

static void Assert(bool expr)
{
 if (expr) return;
 ++barrier;
 Console.WriteLine($"assertion failed {writes}");
 Environment.Exit(1);
}

static void ThreadMain()
{
 for (int i = 0; i < N; ++i)
 {
  var o = ready[i];
  Assert(o == null || o.i == 1);
  if (i == N - 1 && o != null)
   return;
 }
}

static void Main()
{
 for (int i = 0; i < N; ++i)
  reserve[i] = new obj();
 while (true)
 {
  for (int i = 0; i < N; ++i)
    reserve[i].i = 0;
  for (int i = 0; i < N; ++i)
    ready[i] = null;
   ++barrier;
   var thread = new Thread (ThreadMain);
   thread.Start ();
   for (int i = 0; i < N; ++i)
   {
     var o = reserve[i];
     ++writes;
     o.i = 1;
     ready[i] = o;
   }
   thread.Join ();
  }
}
}
```

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -8937,6 +8937,11 @@ calli_end:
 		case MONO_CEE_STIND_I: {
 			sp -= 2;
 
+			// Workaround weak ARM memory model.
+#if defined (HOST_ARM) || defined (HOST_ARM64)
+			if (il_op == CEE_STIND_REF)
+				ins_flag |= MONO_INST_VOLATILE;
+#endif
 			if (ins_flag & MONO_INST_VOLATILE) {
 				/* Volatile stores have release semantics, see 12.6.7 in Ecma 335 */
 				mini_emit_memory_barrier (cfg, MONO_MEMORY_BARRIER_REL);
@@ -9732,6 +9737,12 @@ calli_end:
 				}
 			}
 
+			// Workaround weak ARM memory model.
+#if defined (HOST_ARM) || defined (HOST_ARM64)
+			if (store_val && store_val->type == STACK_OBJ)
+				ins_flag |= MONO_INST_VOLATILE;
+#endif
+
 			if (method->wrapper_type != MONO_WRAPPER_NONE) {
 				field = (MonoClassField *)mono_method_get_wrapper_data (method, token);
 				klass = field->parent;
@@ -10138,7 +10149,7 @@ calli_end:
 
 			/* Generate IR to do the actual load/store operation */
 
-			if ((il_op == MONO_CEE_STFLD || il_op == MONO_CEE_STSFLD) && (ins_flag & MONO_INST_VOLATILE)) {
+			if (store_val && (ins_flag & MONO_INST_VOLATILE)) {
 				/* Volatile stores have release semantics, see 12.6.7 in Ecma 335 */
 				mini_emit_memory_barrier (cfg, MONO_MEMORY_BARRIER_REL);
 			}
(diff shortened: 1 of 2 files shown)


jaykrell:
Still fails. I finally got VPN working, will see if I can get any stack or such.
Not clear if it is csc or ilstringreplacer. csc is usual but it looks like ilstringreplacer.

jaykrell:
This has been seen to fix the test case in https://github.com/mono/mono/issues/8731 on ARM64.
I didn't try ARM32. Here it is duplicated:

```
using System;
using System.Threading;

class A
{
class obj { public int i; }

static int N = 100 * 100 * 100;
static obj[] reserve = new obj [N];
static obj[] ready = new obj [N];
static long writes;
static volatile int barrier;

static void Assert(bool expr)
{
 if (expr) return;
 ++barrier;
 Console.WriteLine($"assertion failed {writes}");
 Environment.Exit(1);
}

static void ThreadMain()
{
 for (int i = 0; i < N; ++i)
 {
  var o = ready[i];
  Assert(o == null || o.i == 1);
  if (i == N - 1 && o != null)
   return;
 }
}

static void Main()
{
 for (int i = 0; i < N; ++i)
  reserve[i] = new obj();
 while (true)
 {
  for (int i = 0; i < N; ++i)
    reserve[i].i = 0;
  for (int i = 0; i < N; ++i)
    ready[i] = null;
   ++barrier;
   var thread = new Thread (ThreadMain);
   thread.Start ();
   for (int i = 0; i < N; ++i)
   {
     var o = reserve[i];
     ++writes;
     o.i = 1;
     ready[i] = o;
   }
   thread.Join ();
  }
}
}
```

jaykrell:
This does not fix the current severe problems, but very well might still be reasonable.
It does address the test case that is believed to resemble real world code.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build Linux x64 mcs

jaykrell:

jaykrell:
&monojenkins build failed

jaykrell:
@monojenkins build failed

mono/mini/method-to-ir.c
```diff
@@ -8937,6 +8937,11 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 		case MONO_CEE_STIND_I: {
 			sp -= 2;
 
+			// Workaround weak ARM memory model.
+#if defined (HOST_ARM) || defined (HOST_ARM64)
```
lewurm: shouldn't this be `#if defined (TARGET_ARM) || defined (TARGET_ARM64)`?

mono/mini/method-to-ir.c
```diff
@@ -10138,7 +10149,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 
 			/* Generate IR to do the actual load/store operation */
 
-			if ((il_op == MONO_CEE_STFLD || il_op == MONO_CEE_STSFLD) && (ins_flag & MONO_INST_VOLATILE)) {
+			if (store_val && (ins_flag & MONO_INST_VOLATILE)) {
```
lewurm: either I don't understand this change or it's unrelated to what this PR is about. could you elaborate?

mono/mini/method-to-ir.c
jaykrell: Yes.

jaykrell:
Still 100% failure rate on arm64 chromebooks. :(

PR #9530, opened 2018-07-13
Wrap some more SIMD code in MONO_ARCH_SIMD_INTRINSICS define.

lateralusX:


jaykrell:
Prefer runtime-if over preprocessor-ifdef, so that some day, one mono binary/compiler, can have multiple targets, selected at runtime (command line flag), i.e. for AOT.

But I'm guilty too and I don't know if we have commited to this. I wish we would.

mono/mini/simd-intrinsics.c
```diff
@@ -1580,7 +1580,11 @@ simd_intrinsic_emit_ctor (const SimdIntrinsic *intrinsic, MonoCompile *cfg, Mono
 {
 	MonoInst *ins = NULL;
 	int i, addr_reg;
+#if defined(TARGET_WIN32) && defined(TARGET_AMD64)
```
kumpera: Why is this only needed on windows amd64?

mono/mini/mini.h
```diff
@@ -1364,7 +1364,9 @@ typedef struct {
 	guint            r4fp : 1;
 	guint            llvm_only : 1;
 	guint            domainvar_inited : 1;
+#ifdef MONO_ARCH_SIMD_INTRINSICS
```
kumpera: Makes no sense to throw ifdefs for his.

mono/mini/mini.h
```diff
@@ -1538,11 +1540,13 @@ typedef enum {
 	MONO_CFG_HAS_TYPE_CHECK = 1 << 9
 } MonoCompileFlags;
 
+#ifdef MONO_ARCH_SIMD_INTRINSICS
```
kumpera: What ifdefing an enum out buys us?

mono/mini/method-to-ir.c
```diff
@@ -6421,12 +6421,17 @@ emit_setret (MonoCompile *cfg, MonoInst *val)
 		} else {
 			EMIT_NEW_RETLOADA (cfg, ret_addr);
 
+#ifdef MONO_ARCH_SIMD_INTRINSICS
```
kumpera: This change should not cause the code to behave differently.

Simplified from mini.h: 

```
#ifdef MONO_ARCH_SIMD_INTRINSICS
#define MONO_CLASS_IS_SIMD(cfg, klass) (((cfg)->opt & MONO_OPT_SIMD) && m_class_is_simd_type (klass))
#else
#define MONO_CLASS_IS_SIMD(cfg, klass) (0)
#endif
```

Meaning if `MONO_ARCH_SIMD_INTRINSICS` is not defined, the existing code will always execute the else branch and emit the same code as proposed by this change.

Additionally, 8dd6baee6e7f change to method-to-ir.c should use MONO_CLASS_IS_SIMD instead of adding a `#ifdef MONO_ARCH_SIMD_INTRINSICS`

mono/mini/mini.h
jaykrell: There is a possible rationale, of minimizing what you declare, to only what works. It helps people  avoid later errors. This is like producing slimmed down BCL metadata so calling nonexistant functions fails to autocomplete/compile instead throwing a runtime exception.

Personally I strive to minimize ifdefs and there are way too many in mono and the wider world.
In order to have fewer variations to compile and perhaps to link, so more functionality can be had from fewer binaries, unless performance suffers much.

mono/mini/mini.h
jaykrell: Maybe similar to the enum. I can see the point, but I prefer fewer ifdefs, instead of the slight optimization or slightly increased static checking. (while meanwhile I do like static checking, so maybe that is contradictory).

lateralusX:
This PR was initially done as a test to reduce some of the SIMD changes done in 8dd6bae on platforms not using SIMD optimizations while tracking down a problem on ARM64 that later turned out to be related to the span checkins. Since we have pass on all SIMD tests (and the ARM64 problem was identified to be caused by a different change) this PR doesn't add much value anymore so I will just close it and leave it.

Issue #9531, opened 2018-07-13
crash handler is blocked on lock


lewurm:
## Steps to Reproduce

```csharp
using System;

public class CrashMe {
        public static unsafe void Main (string []args) {
                int *p = (int *) 0x51337;
                *p = 0;
        }
}
```

```
$ mcs /unsafe crashme.cs
$ ./mono/mini/mono-sgen crashme.exe
Stacktrace:

Memory around native instruction pointer (0x10bcf0f9c):
0x10bcf0f8c  00 00 00 00 48 83 ec 08 b8 37 13 05 00 48 63 c0  ....H....7...Hc.
0x10bcf0f9c  c7 00 00 00 00 00 48 83 c4 08 c3 00 00 00 00 00  ......H.........
0x10bcf0fac  00 00 00 00 55 48 8b ec 48 83 ec 60 4c 89 6d e8  ....UH..H..`L.m.
0x10bcf0fbc  4c 89 75 f0 4c 89 7d f8 4c 8b f6 48 89 55 b0 4c  L.u.L.}.L..H.U.L

Native stacktrace:

```
attaching via `lldb` from another terminal:

```
(lldb) bt
warning: could not execute support code to read Objective-C class data in the process. This may reduce the quality of type information available.
* thread #1, name = 'tid_307', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
  * frame #0: 0x00007fff5377beda libsystem_pthread.dylib`pthread_mutex_trylock
    frame #1: 0x000000010b640f75 mono-sgen`mono_os_mutex_trylock(mutex=0x0000000000000018) at mono-os-mutex.h:109
    frame #2: 0x000000010b63fec5 mono-sgen`mono_coop_mutex_lock(mutex=0x0000000000000018) at mono-coop-mutex.h:51
    frame #3: 0x000000010b640648 mono-sgen`mono_debugger_state(writer=0x000000010bd86de8) at debugger-state-machine.c:271
    frame #4: 0x000000010b640ecb mono-sgen`mono_debugger_state_str at debugger-state-machine.c:403
    frame #5: 0x000000010b5da63c mono-sgen`dump_native_stacktrace(signal="SIGSEGV", ctx=0x000000010bd87f48) at mini-posix.c:1007
[...]

PR #9543, opened 2018-07-13
[arm64] Make OP_CHECK_THIS read only a byte, to prevent unaligned access errors.

vargaz:


marek-safar:
@monojenkins backport 2018-06

jaykrell:
Other architectures?

lewurm:
Let me hijack this PR to do a recap of the Slack conversation that lead to this fix.

We started to see ARM64 lanes fail to pretty reliably in `cil-stringreplacer` after the PR https://github.com/mono/mono/pull/9204 for `Span<T>` had been merged into `master`. Actually, this failure already showed up on some CI runs of this PR (e.g. https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-aarch64/6250/parsed_console/log.html ), but it didn't seem to be reliable enough yet and probably had been falsely identified as a roslyn crash. To be fair, error handling was also broken at this time due to https://github.com/mono/mono/issues/9531 so there was some additional factor that helped the crash to sneak through.

Additionally this crash needed some special conditions in the environment to be true; it wasn't reproducible on our "dev arm64" box. Thanks to @akoeplinger for providing proper steps to reproduce the exact environment that we have on CI.

Once reproduced, it was pretty clear this is a `Span<T>` related issue:
```
Thread 1 "mono" received signal SIGSEGV, Segmentation fault.
0x0000007fac841de4 in ?? ()
(gdb) bt
#0  0x0000007fac841de4 in ?? ()
#1  0x0000007faaa64120 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) p mono_pmip (0x0000007fac841de4)
$1 = 0x5555def000 " System.SpanHelpers:SequenceCompareTo<char> (char&,int,char&,int) {0x5555df0510} + 0x54 (0x7fac841d90 0x7fac841e60) [0x555590ed80 - cil-stringreplacer.exe]"
(gdb) x/10i $pc
=> 0x7fac841de4:        ldr     x30, [x0]
   0x7fac841de8:        ldrh    w0, [x0]
   0x7fac841dec:        sub     w22, w0, w1
(gdb) p/x $x0
$4 = 0x7fae52bffa
```
So what happens is that `ldr` tries to do a word load from an _unaligned_ address (`0x7fae52bffa`) which causes a segfault. This is what this PR fixes: It changes the word load to a byte load which doesn't have this behaviour (side note: `LR` is a alias for `x30`).

kumpera:
@lewurm @vargaz could you explain how come `this` can be an unaligned pointer?
It's counter intuitive to what one would normally assume about managed object pointers.

lewurm:
@kumpera it's unsafe code: https://github.com/mono/corefx/blob/df455a492907085c0bf9dd8c92de161e602251fe/src/Common/src/CoreLib/System/SpanHelpers.Char.cs#L20

jaykrell:
I had the same question, and still do -- do/can we use "this" for arbitrary other things?

lewurm:
it's the guard from inlining a method via `callvirt`: https://gist.github.com/jaykrell/d3fe055780a1be1b7c2bed844b969665#file-arm64-cil-stringreplacer-fast-span-L85-L96

kumpera:
It makes sense. `this` can be the byref to a valuetype. And if that's a `byte` or `char`, it's ok to be unaligned.

PR #9552, opened 2018-07-14
[Coop] Mostly convert System.GC and System.Runtime.InteropServices.GC.Handle.

jaykrell:


mono/metadata/icall.c
```diff
@@ -8425,3 +8425,26 @@ mono_register_jit_icall (gconstpointer func, const char *name, MonoMethodSignatu
 	return mono_register_jit_icall_full (func, name, sig, no_wrapper, NULL);
 }
 
+ICALL_EXPORT int
```
jaykrell: One wonders if this can overflow.

jaykrell: It follows MSDN though https://msdn.microsoft.com/en-us/library/system.gc.collectioncount(v=vs.110).aspx.

jaykrell:
@monojenkins build failed

jaykrell:
Familiar Rosyln and FullAOT+LLVM crashes.

jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/gc.c
```diff
@@ -622,70 +629,73 @@ ves_icall_System_GC_WaitForPendingFinalizers (void)
...
+ves_icall_System_GC_register_ephemeron_array (MonoObjectHandle array, MonoError *error)
 {
-#ifdef HAVE_SGEN_GC
-	if (!mono_gc_ephemeron_array_add (array)) {
-		mono_set_pending_exception (mono_object_domain (array)->out_of_memory_ex);
-		return;
-	}
-#endif
+	if (!mono_gc_ephemeron_array_add (MONO_HANDLE_RAW (array)))
+		mono_set_pending_exception (MONO_HANDLE_DOMAIN (array)->out_of_memory_ex);
```
lambdageek: Could drop the use of `mono_set_pending_exception` in favor of `mono_error_set_exception_instance (error, ...); return;`

mono/metadata/gc.c
```diff
@@ -622,70 +629,73 @@ ves_icall_System_GC_WaitForPendingFinalizers (void)
...
-ves_icall_System_GC_register_ephemeron_array (MonoObject *array)
+ves_icall_System_GC_register_ephemeron_array (MonoObjectHandle array, MonoError *error)
 {
-#ifdef HAVE_SGEN_GC
-	if (!mono_gc_ephemeron_array_add (array)) {
-		mono_set_pending_exception (mono_object_domain (array)->out_of_memory_ex);
-		return;
-	}
-#endif
+	if (!mono_gc_ephemeron_array_add (MONO_HANDLE_RAW (array)))
```
lambdageek: Why not just keep the #ifdef here instead of adding a stub to Boehm?

mono/metadata/object-internals.h
```diff
@@ -46,7 +46,7 @@
 
 /* Use this as MONO_CHECK_ARG_NULL_HANDLE (arg,) in functions returning void */
 #define MONO_CHECK_ARG_NULL_HANDLE(arg, retval) do { 		\
-	if (MONO_HANDLE_IS_NULL (arg))				\
+	if (G_UNLIKELY (MONO_HANDLE_IS_NULL (arg)))		\
```
jaykrell: "likely" has bounced here, somewhat by accident.
Arguably it should  be in MONO_HANDLE_IS_NULL.

mono/metadata/object-internals.h
```diff
@@ -55,7 +55,7 @@
 
 /* Use this as MONO_CHECK_NULL (arg,) in functions returning void */
 #define MONO_CHECK_NULL(arg, retval) do { 			\
```
jaykrell: This macro only has one user at this point.

mono/metadata/object-internals.h
lambdageek: I don't think it makes sense for `likely`/`unlikey` to be part of `MONO_HANDLE_IS_NULL` - different circumstances would have different likelihoods

mono/metadata/gc.c
jaykrell: ifdef is bad and to be reduced. Variation generally better at link time or runtime (command line switch), in order to compile and/or link fewer times. Unless for extreme perf reasons.

mono/metadata/object-internals.h
jaykrell: Agreed, though barely. Pretty few paths will have l likely(null), I think.  But I understand, everything is context relative.

mono/metadata/gc.c
jaykrell: Good point, I'll try.

jaykrell:

PR #9556, opened 2018-07-14
Add support for an output path in mono-api-diff

mattleibow:
Rather than always writing to the std out, provide an option to write to a file. This makes things easier for tooling by not requiring `RedirectStandardOutput = true`


mattleibow:
This makes it a bit harder as we will have to move the files around and add a reference to mono.options… Right now this folder is all set up to build mono-api-info.exe. This appears to be just a single file.

Is that an option?

mattleibow:
let me get on it then :)

mattleibow:
@marek-safar I moved the mono-api-diff tool into its own folder: https://github.com/mono/mono/pull/9563

mattleibow:
I think this is ready to be merged from my side. I tested the app with my projects, and it works as expected.

marek-safar:
@monojenkins build failed

PR #9563, opened 2018-07-14
[mono-api-diff] Moved mono-api-diff into its own tools directory

mattleibow:
Once this is merged, I will re-do #9556 with Mono.Options


packaging/MacSDK/packaging/resources/whitelist.txt
```diff
@@ -77,6 +77,7 @@ mono32
 mono64
 mono-boehm
 mono-api-info
+mono-api-diff
```
marek-safar: I don't think we want to install the tool at this point

scripts/Makefile.am
```diff
@@ -89,6 +89,7 @@ scripts_4_0 = \
 	mod$(SCRIPT_SUFFIX)			\
 	monolinker$(SCRIPT_SUFFIX)		\
 	mono-api-info$(SCRIPT_SUFFIX) 		\
+	mono-api-diff$(SCRIPT_SUFFIX) 		\
```
marek-safar: Same as above

mcs/tools/mono-api-diff/mono-api-diff.sln
```diff
@@ -0,0 +1,17 @@
+
```
marek-safar: This file is redundant

PR #9572, opened 2018-07-16
[corlib] Moves private Debug class in mscorlib to System.Diagnostics namespace

marek-safar:
This is to help with CoreFX and CoreRT integration which have Debug class implementation in mscorlib.


filipnavara:
Hmm, shouldn't Mono.Security link against non-reference assemblies? Or at least link against reference assemblies for both `mscorlib` and `System`? Mixing it seems wrong...

I thought that `System` references `Mono.Security` and thus I am quite surprised to see the circular reference.

filipnavara:
Eureka! Looks like the remaining failures are just different line numbers in "symbolicate" output, which is to be expected.

marek-safar:
@monojenkins commit csproj

PR #9587, opened 2018-07-16
GetLastError does not warrant changing GC mode. It is one instruction, …

jaykrell:
…not a syscall, and call/ret.

Getting/setting thread locals generally is not either, except for the rare expansion path (unless
we switch to __declspec(thread) which is never a syscall and is much faster.)

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -4940,7 +4940,6 @@ mono_marshal_set_last_error (void)
 	/* This icall is called just after a P/Invoke call before the P/Invoke
 	 * wrapper transitions the runtime back to running mode. */
 #ifdef WIN32
-	MONO_REQ_GC_SAFE_MODE;
 	mono_native_tls_set_value (last_error_tls_id, GINT_TO_POINTER (GetLastError ()));
 #else
 	mono_native_tls_set_value (last_error_tls_id, GINT_TO_POINTER (errno));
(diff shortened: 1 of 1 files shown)


lambdageek:
I don't understand this PR.  `MONO_REQ_GC_UNSAFE_MODE` is not _changing_ the thread state. it's _asserting_ the thread state.

And all blocking syscalls should be done from GC Safe state.

What's the problem?

alexischr:
@lambdageek @jaykrell pointed out on another place that GetLastError() has not been an actual API call in a very long time and is just thread-local access. So they are redundant assertions, but not slow GC transitions as Jay may have expected.

lambdageek:
I see. well that probably does simplify how these two function can be called.  Fair enough. 👍

jaykrell:
I exaggerated, it is two instructions (and call and ret):

\bin\amd64\cdb cmd

Microsoft (R) Windows Debugger Version 10.0.16299.91 AMD64

0:000> u  kernelbase!GetLastError
KERNELBASE!GetLastError:
00007ffa`2a4ba300 65488b042530000000 mov   rax,qword ptr gs:[30h]
00007ffa`2a4ba309 8b4068          mov     eax,dword ptr [rax+68h]
00007ffa`2a4ba30c c3              ret

but that's just bad optimization.
It could be mov eax, gs:68.

TlsGetValue and/or TlsSetValue do have sometimes "expansion" paths that can heap allocate, and therein lock and/or syscall. But if we switch to __declspec(thread) (i.e. __thread), that goes away. PR maybe soon.

jaykrell:
The second case is not SetLastError/GetLastError.

As such, I’d like https://github.com/mono/mono/pull/9594#issuecomment-406127885 to go first, to avoid the rare expansion path of TLS.

Also, I think the Posix path is mischaracerized. It likely has rare expansion paths too. Like the very first get or set, depending on some factors.

PR #9588, opened 2018-07-16
[marshal] Remove redundant GC-Safe assertions

alexischr:
/cc @jaykrell


filipnavara:
Duplicate of #9587? :)

alexischr:
@filipnavara yup :)

PR #9590, opened 2018-07-16
Use __declspec(thread) with Visual C++, and __thread with Win32/gcc,

jaykrell:
instead of much slower TlsGetValue / TlsSetValue.

### DIFF
--- a/mono/mini/trace.c
+++ b/mono/mini/trace.c
@@ -60,6 +60,7 @@ MonoCallSpec *mono_trace_set_options (const char *options)
 	return &trace_spec;
 }
 
+// comment only
 static
 #ifdef HAVE_KW_THREAD
 __thread 


jaykrell:
Wierd all around. I didn't close it. The errors don't make sense. I pushed an update with  no change due to  the nonsense errors.

kumpera:
This is meant to be an empty PR? Sure, go for it. Just make sure to fully understand all issues documented in mono-tls.c become committing to it.

jaykrell:
The "real" PR  failed to compile  in wierd says so I'm baselining an empty PR first.
Building locally with msbuild/msvc worked, and the PR failed with msvc.

The main "problems" with  __declspec(thread):

1. Don't work often pre-Vista -- only in the  .exe and static dependents of it.

2. Different cross-dll semantics. I.e. tls keys can be passed across dlls and lifetime not tied to a dll. __declspec(thread) lifetime is tied to a .dll. I've seen that bite one project. Usually people tie TLS keys to .dlls anyway.

3. They aren't fiber local. A different and contradictory change that I would reluctantly agree with is to  use FlsAlloc instead. __declspec(fiber) would be ideal, but it doesn't exist.
We do actually sometimes  use FlsAlloc, but only sometimes. The inconsistency seems bad.

```
/*
* These APIs were added back in Windows SDK 14393. Let's redirect them to
* Fls* APIs on older SDKs just like Windows 8.1 headers do
*/
#define TlsGetValue FlsGetValue
#define TlsSetValue FlsSetValue
#define TlsFree FlsFree
#endif
#endif
```

We do have other uses of TlsAlloc that should probably be __declspec(thread) but I left them alone for now.

PR #9595, opened 2018-07-16
[Coop] Convert 3 more functions and strengthen some types to be not just gpointer.

jaykrell:
i.e. to see in debugger and for static type checking

### DIFF
--- a/mono/metadata/cominterop.c
+++ b/mono/metadata/cominterop.c
@@ -116,13 +116,6 @@ enum {
 #define mono_cominterop_unlock() mono_os_mutex_unlock (&cominterop_mutex)
 static mono_mutex_t cominterop_mutex;
 
-/* STDCALL on windows, CDECL everywhere else to work with XPCOM and MainWin COM */
-#ifdef HOST_WIN32
-#define STDCALL __stdcall
-#else
-#define STDCALL
-#endif
-
 GENERATE_GET_CLASS_WITH_CACHE (interop_proxy, "Mono.Interop", "ComInteropProxy")
 GENERATE_GET_CLASS_WITH_CACHE (idispatch,     "Mono.Interop", "IDispatch")
 GENERATE_GET_CLASS_WITH_CACHE (iunknown,      "Mono.Interop", "IUnknown")
@@ -143,7 +136,7 @@ typedef struct {
 	guint32 gc_handle;
 	GHashTable* vtable_hash;
 #ifdef  HOST_WIN32
-	gpointer free_marshaler;
+	MonoIUnknown *free_marshaler; // actually IMarshal
 #endif
 } MonoCCW;
 
@@ -572,7 +565,7 @@ cominterop_get_interface_checked (MonoComObjectHandle obj, MonoClass* ic, MonoEr
 	guint8 iid [16];
 	gboolean const found = cominterop_class_guid (ic, iid);
 	g_assert (found);
-	int const hr = ves_icall_System_Runtime_InteropServices_Marshal_QueryInterfaceInternal (MONO_HANDLE_GETVAL (obj, iunknown), iid, &itf);
+	int const hr = ves_icall_System_Runtime_InteropServices_Marshal_QueryInterfaceInternal (MONO_HANDLE_GETVAL (obj, iunknown), iid, &itf, error);
 	if (hr < 0) {
 		g_assert (!itf);
 		cominterop_set_hr_error (error, hr);
@@ -1580,13 +1573,6 @@ mono_cominterop_emit_marshal_com_interface (EmitMarshalContext *m, int argnum,
 	return conv_arg;
 }
 
-typedef struct
-{
-	int (STDCALL *QueryInterface)(gpointer pUnk, gconstpointer riid, gpointer* ppv);
-	int (STDCALL *AddRef)(gpointer pUnk);
-	int (STDCALL *Release)(gpointer pUnk);
-} MonoIUnknown;
-
 #define MONO_S_OK 0x00000000L
 #define MONO_E_NOINTERFACE 0x80004002L
 #define MONO_E_NOTIMPL 0x80004001L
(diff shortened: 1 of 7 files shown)


mono/metadata/cominterop.c
```diff
@@ -2854,6 +2846,7 @@ mono_string_from_bstr_checked (mono_bstr_const bstr, MonoError *error)
 		gunichar2* utf16 = g_ucs4_to_utf16 ((const gunichar *)bstr, sys_string_len_ms (bstr), NULL, &written, NULL);
 		MonoStringHandle res = mono_string_new_utf16_handle (mono_domain_get (), utf16, written, error);
 		g_free (utf16);
+		return res;
```
jaykrell: Accidental recent regression.

jaykrell:
@monojenkins Linux x64 Hybrid Suspend

jaykrell:

mono/metadata/cominterop.c
```diff
@@ -1595,24 +1581,30 @@ typedef struct
...
+	return pUnk->vtable->AddRef(pUnk);
 }
 
 int
-ves_icall_System_Runtime_InteropServices_Marshal_QueryInterfaceInternal (gpointer pUnk, gconstpointer riid, gpointer* ppv)
+ves_icall_System_Runtime_InteropServices_Marshal_QueryInterfaceInternal (MonoIUnknown *pUnk, gconstpointer riid, gpointer* ppv, MonoError *error)
 {
 	g_assert (pUnk);
-	return (*(MonoIUnknown**)pUnk)->QueryInterface(pUnk, riid, ppv);
+	return pUnk->vtable->QueryInterface(pUnk, riid, ppv);
```
lambdageek: bikeshedding: I'd love a static `mono_IUnknown_QueryInterface` that's called from here and from `cominterop_get_interface_checked` so that we can have fewer situations where icalls are used as utility functions inside the runtime.

mono/metadata/marshal.h
```diff
@@ -761,13 +761,16 @@ MonoBoolean
...
 gint32
-ves_icall_System_Runtime_InteropServices_Marshal_ReleaseComObjectInternal (MonoObject* object);
+ves_icall_System_Runtime_InteropServices_Marshal_ReleaseComObjectInternal (MonoObjectHandle object, MonoError *error);
 
 MonoObjectHandle
 ves_icall_System_ComObject_CreateRCW (MonoReflectionTypeHandle ref_type, MonoError *error);
 
 void
-ves_icall_System_ComObject_ReleaseInterfaces(MonoComObject* obj);
+mono_System_ComObject_ReleaseInterfaces (MonoComObjectHandle obj);
```
lambdageek: I'm curious why this one is global while `mono_IUnknown_Release` is static, and an accessor for `QueryInterface` shouldn't exist at all.

mono/utils/mono-threads.h
```diff
@@ -177,7 +177,7 @@ typedef enum {
 	MONO_THREAD_INFO_FLAGS_NO_SAMPLE = 2,
 } MonoThreadInfoFlags;
 
-typedef struct {
+typedef struct _MonoThreadInfo {
```
lambdageek: 👍

mono/metadata/threads-types.h
```diff
@@ -289,7 +289,7 @@ gboolean
 mono_thread_create_checked (MonoDomain *domain, gpointer func, gpointer arg, MonoError *error);
 
 /* Can't include utils/mono-threads.h because of the THREAD_INFO_TYPE wizardry */
-void mono_threads_add_joinable_runtime_thread (gpointer thread_info);
+void mono_threads_add_joinable_runtime_thread (MonoThreadInfo *thread_info);
```
lambdageek: comment above this line about `THREAD_INFO_TYPE` seems relevant - this should probably be `struct _MonoThreadInfo*` (because `MonoThreadInfo` typedef is not necessarily visible every time this header is included).

mono/metadata/marshal.h
jaykrell: Could be maybe more consistent. I can look again.

mono/metadata/threads-types.h
jaykrell: Yeah I was kinda expecting it not to compile but it did.

jaykrell:
@monojenkins build failed

mono/metadata/threads-types.h
lambdageek: Yea we really need a fragile header CI lane.  

`gcc -Wall -Werror -DHEADER=threads-types.h -c wrapper.c`
wrapper.c:
```c
#include "config.h"
#include HEADER
/* include twice to verify that there are include guards */
#include HEADER
int main () { return 0; }
```

mono/metadata/cominterop.c
jaykrell: There were minor variations that made it not obvious how to unify internal calls and icall.
They were related to:
 - check null?
 - assert null? The assert is redundant though with the C# caller. But people are usually ok-or-better with simple safe fast redundant asserts, as long as the code isn't overrun with them.
 - use return value?
I thought I ended up with some unification but maybe more could be had.
I had to invent some names too, like mono_com_foo or mono_cominterop_foo or mono_IUknown_foo. IUnknown_Release and such are macros in the C versions of midl output. I believe they commit the sin of double evaluation (#define Foo_Bar(foo) ((foo)->vtable->bar(foo)) so I went with mono_IUnknown_foo.

Anyway I can look again.

jaykrell: Ok I think I see your  point.
ves_AddRef and ves_QueryInterface have internal versions, for Release does not.
Probably based on what used/needed.

jaykrell: and ves don't call ves, understood.

jaykrell: and non-ves calling ves; ves only called from C#/icall-def.h

mono/metadata/marshal.h
jaykrell: I added all 3 helpers now. I think one is only used once, or zero. But there was room for consistency, agreed.

jaykrell:
Rebasing to master due to merge conflict.

mono/metadata/threads-types.h
jaykrell: I think the code is ok and the comment is wrong. I will remove the comment.
The header is included and does have an include guard, and the typedef is always visible.

The header does have varying definitions, gross, but I don't see how it affects this line.

This might  be somewhat or completely improvable with C++ with overloads or derivatin (SgenThreadInfo from ThreadInfo), at least for input parameters, not return values -- there are some of each.

mono/metadata/threads-types.h
```diff
@@ -288,8 +288,7 @@ void mono_threads_perform_thread_dump (void);
 gboolean
 mono_thread_create_checked (MonoDomain *domain, gpointer func, gpointer arg, MonoError *error);
 
-/* Can't include utils/mono-threads.h because of the THREAD_INFO_TYPE wizardry */
-void mono_threads_add_joinable_runtime_thread (gpointer thread_info);
+void mono_threads_add_joinable_runtime_thread (MonoThreadInfo *thread_info);
```
lambdageek: Still seems to me like this should be `struct _MonoThreadInfo`.  Otherwise aren't we depending on `mono/utils/mono-threads.h` to be included first?

jaykrell: Yes. It is included first.

jaykrell:
@monojenkins build Linux x64

mono/metadata/object-internals.h
```diff
@@ -335,9 +335,31 @@ typedef struct {
...
+struct _MonoIUnknownVTable
+{
+	int (STDCALL *QueryInterface)(MonoIUnknown *pUnk, gconstpointer riid, gpointer* ppv);
+	int (STDCALL *AddRef)(MonoIUnknown *pUnk);
+	int (STDCALL *Release)(MonoIUnknown *pUnk);
+};
+
+struct _MonoIUnknown
+{
+	MonoIUnknownVTable const * const vtable;
```
jaykrell: This toplevel/rightmost const should be removed, for obscure reason.

It is rarely really useful in a struct.
Some recent shipping versions of Visual C++ have a bug where it leads to dynamic initialization instead of static initialization.

PR #9596, opened 2018-07-16
[Coop] Remove unused mono_object_to_string_checked.

jaykrell:


jaykrell:

lambdageek:
`System.IO.IOException : System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached.` in the AArch64 and AArch64 Coop failures is not related.

How did we end up with such a low limit, anyway?

jaykrell:
@monojenkins build failed

Issue #6886, opened 2018-02-08
Update csproj generator to produce output which has no makefile pre-requirements


marek-safar:
- [x] mcs/build/common/Consts.cs.in

For it, we could make a common project which does the transformation and fills needed version info and make it a dependency of mscorlib to run before any project

PR #9598, opened 2018-07-16
Auto-generate Consts.cs on Windows when building from Visual Studio

kg:
This adds a simple script for generating Consts.cs that can run on Windows, and makes it a dependency of corlib.csproj in the BCL solution file. From my initial testing this will autogenerate Consts.cs if it's missing and then corlib will successfully build.
This (mostly?) addresses issue https://github.com/mono/mono/issues/6886.

This currently has an awkward dependency on ```winsetup.bat``` and I modified it slightly to make it clearer whether the batch file succeeded. Tagging @jaykrell since he wrote winsetup.


msvc/scripts/genconsts.cs
```diff
@@ -0,0 +1,98 @@
+using System;
```
jaykrell: 1. It might be less code to read configure.ac?
2. More so, whatever lines we parse from it, probably should be in separate file or files.

kg: My thinking was that whatever winsetup produces should match whatever genconsts produces. Are you suggesting making winsetup produce a file to read from instead of parsing stdout? That seems reasonable

jaykrell: I was thinking of extracting lines from configure.ac into, say, monoversion.txt.
Have one small file for everyone to read and easily parse.
But maybe not worth it. Maybe configure.ac is too strict and must have those lines.

Your idea is also good though, similar, but less.

I just kinda think that "files" tend to have more stable formats than "stdout".

For example, it should be possible to debug, by removing `echo off`, without breaking stuff.

But what you have is ok. I can debug just by running the thing directly.

jaykrell: Also, your comment said you generate the file if it doesn't exist. That sounds wrong.  You generate it if it is out of date or incorrect. Such as, worst but common case, generate it always, compare to what you have,  and if different, move it over. Hopefully that isn't slow.

jaykrell: Like this in winsetup.bat:

`:: If the file is different, replace it.
fc %monotemp% %config_h% >nul || move /y %monotemp% %config_h%`

but in a more elegant efficient  C# way. (The cmd way is inefficient in that it creates an extra process, for example.)

jaykrell: I was also thinking -- what I really said -- is you could read configure.ac and duplicate winsetup.bat probably in far fewer lines. But that does lose synchronicity, granted. More code to achieve sharing logic is better than less code and duplicated logic. But you also fixup the number, so the logic doesn't match. Anyway, this is much ado about little.

msvc/scripts/genconsts.cs
```diff
@@ -0,0 +1,98 @@
...
+            return 1;
+        }
+
+        using (winsetupProcess) {
+            var outputBuffer = new StringBuilder ();
+
+            winsetupProcess.OutputDataReceived += (s, e) => {
+                outputBuffer.AppendLine (e.Data);
+            };
+            winsetupProcess.BeginOutputReadLine ();
```
jaykrell: You should also be aware, at least for future, that when you run stuff,  and don't read both stderr and stdout, that you risk deadlock. I've seen it many times. In this case you are probably ok, probably no writes to stderr.

In native code it is quite messy. C#/BCL might be easier.
In  native code you have two choices for deadlock-free reading of a process's output.

1. You need to read stdout and stderr on separate threads or 
2. You need to set stdout==stderr, i.e. redirect them to the same handle.

Again, I've seen this deadlock countless times, but it tended to be in C++ and script.exe/jscript, perhaps not C#.

msvc/scripts/genconsts.cs
```diff
@@ -0,0 +1,98 @@
...
+                Console.Error.WriteLine ("Failed running winsetup.bat");
+                Console.Write (output);
+                return winsetupProcess.ExitCode;
+            } else {
+                var m = Regex.Match (output, "MONO_VERSION=([0-9.]+)");
+                if (!m.Success)
+                    return 1;
+                monoVersion = m.Groups[1].Value;
+
+                // HACK: winsetup.bat produces N.N.N instead of N.N.N.N like configure.ac,
```
jaykrell: Maybe add another line or two to winsetup.bat to remove the hack? i.e. to output another line with the data you need?

kg: Yeah, if that makes sense to you. I'm not sure why the batch file isn't generated, so I thought you might know.

msvc/scripts/genconsts.cs
```diff
@@ -0,0 +1,98 @@
...
+                Console.WriteLine ($"MONO_VERSION={monoVersion}");
+                m = Regex.Match (output, "MONO_CORLIB_VERSION=([0-9]+)");
+                if (!m.Success)
+                    return 1;
+                monoCorlibVersion = m.Groups[1].Value;
+                Console.WriteLine ($"MONO_CORLIB_VERSION={monoCorlibVersion}");
+            }
+        }
+
+        var constsDirectory = Path.Combine (executableDirectory, "..", "..", "mcs", "build", "common");
```
jaykrell: Is this navigating from project output to source tree? Can output be placed arbitrarily elsewhere, like on another drive? I'd hope so but I don't know/ Is there a parameter you can get from msbuild that is more correct in the face of that?

msvc/scripts/genconsts.cs
kg: It should be OK because I didn't redirect stderr, I think.

msvc/scripts/genconsts.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        var resultPath = Path.GetFullPath (Path.Combine (constsDirectory, "Consts.cs"));
+        var templateText = File.ReadAllText (constsTemplatePath);
+
+        var resultText = templateText.Replace ("@MONO_VERSION@", monoVersion)
+            .Replace ("@MONO_CORLIB_VERSION@", monoCorlibVersion);
+
+        if (File.Exists (resultPath)) {
+            var existingText = File.ReadAllText (resultPath);
+            if (existingText.Trim () == resultText.Trim ()) {
```
jaykrell: Good, this is what I meant.

msvc/scripts/genconsts.cs
jaykrell: I have another idea.
This C# program just runs winsetup.bat and extracts one line, into a file with a "template" aroun that? Can you just augment winsetup.bat to do all this work? It'd only be like 2 or 3 more lines and remove the C# entirely?

msvc/scripts/genconsts.cs
jaykrell: This is why btw, it is to mimic:
```
config.h:#define PACKAGE_VERSION "5.17.0"
config.h:#define VERSION "5.17.0"
```

jaykrell:
```
$ /s/mono/mono/mini/mono --version
Mono JIT compiler version 5.17.0 (cominterop.3/4660b820c27 Mon Jul 16 15:22:46 PDT 2018)

jaymac1:mono jay$ which mono
/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono
jaymac1:mono jay$ mono --version
Mono JIT compiler version 5.10.1.47 (2017-12/8eb8f7d5e74 Fri Apr 13 20:18:12 EDT 2018)
```

kg:
@monojenkins update csproj

kg:
@monojenkins commit csproj

Issue #6992, opened 2018-02-13
BCL size regression checks


marek-safar:
We need to start tracking BCL size regressions for various reasons but mainly to see what corefx imports do. The reasonable way to do it is to have the samples submodule which we use with the linker and then measure any size regression of linked BCL assemblies on every PR.

We care mostly about mobile profiles and the check should have a size tolerance or something similar not to alarm on every commit

PR #9618, opened 2018-07-17
[linker] Add BCL size regression testing

akoeplinger:
We're now tracking the size of the BCL after it was processed by the linker across various profiles we care about (currently net_4_x, monotouch and monodroid).

The way it works is that we keep a file with size measurements in the repo and during CI we run the linker against the current state and compare.

Right now we're simply running the linker tests as they are a good approximation of various usages (this PR additionally adds more comprehensive linker tests we had in XI).

If changes are found there's a report to check where the increase/decrease came from.
The changes can be applied via `@monojenkins commit sizediff` similar to existing csproj and apidiff usages.

Fixes https://github.com/mono/mono/issues/6992


mcs/tools/linker/Makefile
```diff
@@ -41,12 +49,48 @@ TEST_EXEC = MONO_PATH=$(LINKER_OUTPUT) $(RUNTIME) $(RUNTIME_FLAGS) --debug -O=-a
...
-System/test-%.exe mscorlib/test-%.exe:
+System/test-%.exe mscorlib/test-%.exe link-all/link-%.exe link-sdk/link-%.exe:
 	@rm -rf $(LINKER_OUTPUT)
 	@mkdir $(LINKER_OUTPUT)
 	@echo Testing $@
 	$(LINKER) -a Tests/$@
 	$(TEST_EXEC) $(LINKER_OUTPUT)/$(@F)
 	@rm -rf $(LINKER_OUTPUT)
 
+BCL_ASSEMBLIES_CORE=mscorlib.dll System.dll System.Core.dll System.Xml.dll
```
marek-safar: Why do we need the separation?

mcs/tools/linker/Makefile
```diff
@@ -7,6 +7,8 @@ PROGRAM = monolinker.exe
 LIB_REFS = System System.Core System.Xml Mono.Cecil
 
 TEST_CASES := \
+	link-all/link-all.cs \
```
marek-safar: Let's not add these tests in this form

marek-safar:
I'm getting jquery and css content security issues on the output

```
Refused to load the stylesheet '' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' xamjenkinsartifact.blob.core.windows.net cdnjs.cloudflare.com fonts.googleapis.com fonts.gstatic.com". Note that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.
```

mcs/tools/linker/Makefile
akoeplinger: Purely cosmetical so that the .csv and diff shows those assemblies first since they're the ones which have data and are the most interesting. Otherwise they're intermingled between a bunch of 0,0,0,0 values for assemblies which are completely linked away.

mcs/tools/linker/Makefile
akoeplinger: so just build/link them, but not run the actual tests right?

akoeplinger:
The css/js issue should be fixed (might need a cache refresh in your browser).

marek-safar:
@akoeplinger the linker tests are causing crashes on full aot lanes. Let's disable them in this PR

akoeplinger:
@monojenkins commit bclsize

akoeplinger:
@monojenkins commit bclsize

PR #9620, opened 2018-07-18
Extend "auto tailcall" functionality similar to explicit tail.call.

jaykrell:
Many cases that support tail.call, should be able to tailcall w/o the tail. prefix.
The logic should be largely the same but with some tweaks.

Details:
If a function takes the address of any local or parameter, do not auto.tail.
Including fields of local/parameter valuetypes.  Or just any addresses of any fields, if we don't know if the struct is local.
It'd take sophisticated/impossible analysis to see if these escape, and to know if their escape is required by a non-explicit tailcall. Later we could try escape analysis -- which has other uses like converting heap to stack (like go does, if I understand correctly).

Other than call/ret, some other simple common sequences are tail callable:

call ; nop ; ret
call ; st n ; ld n; ret
call ; br x ; x: st n ; ld n; ret
and a few combinations, i.e. nops, one unconditional matching st/ld, unconditional branches, ending at ret
Perhaps this is best done at IR level and not IL level.
Perhaps csc /optimize is common and moots this, but these are common patterns w/o csc /optimize.
Currently not working, just call/ret.

Autotailcall should not likely reduce inlining (hard to say precisely -- they are both optimizations).
Explicit tailcall should not impact inlining as much either.
In particular, the interplay of inlining and tailcall affects non-guaranteed stackwalk, and the ability of the callee stacked parameters to "fit" in the caller stacked parameters.
Inlining can help or hurt the ability to tailcall.

The tests here can be based on the prior tailcall case, but w/ tail. removed.
However, the use of taking-address-of-local-or-parameter to determine if a tailcall is performed cannot be done. Instead we must just recurse deeply and stack overflow or not.
Some tests already fit this, some do not.

As well, this opens up C# tests. We were previously dependent on IL tests (or F# tests) because csc never outputs tail. prefix. Whatever C# the IL tests were based on, we can use as tests here -- but without address-taken, just deep recursion.


mono/mini/mini-x86.c
```diff
@@ -5872,6 +5872,13 @@ mono_arch_get_patch_offset (guint8 *code)
 		/* mov <REG>, imm */
 		return 1;
 	else {
+		guchar bytes [6] = { code [0] };
```
jaykrell: I was debugging something, don't remember what.

mono/mini/method-to-ir.c
```diff
@@ -6175,6 +6175,58 @@ mono_is_supported_tailcall_helper (gboolean value, const char *svalue)
...
+		case MONO_CEE_LDLOCA:
+		case MONO_CEE_LDLOCA_S:
+		case MONO_CEE_LDELEMA: // of a fixed local array?
+		// FIXME check if call is within clause, vs. entire function
+		case MONO_CEE_ENDFINALLY:
+		case MONO_CEE_ENDFILTER:
+		case MONO_CEE_LEAVE:
+		case MONO_CEE_LEAVE_S:
+		case MONO_CEE_RETHROW:
+		case MONO_CEE_THROW:
```
jaykrell: throw should actually be ok

jaykrell: and rethrow I think

mono/tests/Makefile.am
```diff
@@ -1125,26 +1200,29 @@ PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndM
...
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_byte_.Method1_int_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_byte_.Method1_string_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_int_.Method1_DateTime_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_int_.Method1_byte_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_int_.Method1_int_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_int_.Method1_string_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_string_.Method1_DateTime_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_string_.Method1_byte_.exe
-PLATFORM_DISABLED_TESTS += tailcall/fsharp-deeptail/TailCallLoopGenericClassAndMethodAbstractInterface_string_.Method1_int_.exe
+# ditto
```
jaykrell: Looks like bad merge, arm-specific.

jaykrell:
32bit x86 failing horribly, but amd64 doing well locally.

mono/mini/method-to-ir.c
```diff
@@ -6560,6 +6613,124 @@ mono_opcode_decode (guchar *ip, guint op_size, MonoOpcodeEnum il_op, MonoOpcodeP
...
+	gboolean br = FALSE;
+
+	MonoMethodHeader * const header = mono_method_get_header_checked (method, &cfg->error);
+	g_assert (header);
+	guchar * const start = (guchar*)header->code;
+	///*uchar * const*/ end = start + header->code_size;
+	g_assert (end == start + header->code_size);
+	MonoOpcodeEnum il_op = MonoOpcodeEnum_Invalid;
+
+	for (guchar *next_ip = ip; ip >= start && ip < end; ip = next_ip) {
```
jaykrell: This is why I made method_to_ir more data driven -- I wanted to write additional passes.

mono/mini/method-to-ir.c
```diff
@@ -6586,6 +6768,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 		   guint inline_offset, gboolean is_virtual_call)
 {
 	ERROR_DECL (error);
+	int safe_auto_tailcall_cache = 0;
```
jaykrell: Can this be safely/efficiently/more-efficiently be moved into MonoMethod?
1. 2 bits is enough, but might not be atomic, so int?
To some extend non-atomicity is ok, in a race, multiple threads could recompute it, but I worry if it is bitfield, we could have read/update/write and it could tear other bits. Is an int too much to add to every MonoMethod?

2. Can a MonoMethod be compiled multiple times concurrently and then race at commit time?
If not, then a 2 bit bitfield suffices.
However opacity would suggest a typedef and a typedef could be only as small as a char.
Since it is every MonoMethod, I'm nervous.

3. The reason to put in MonoMethod is, well, it doesn't matter currently, but if we can tailcall within an inlined method, that calculation can be reused across multiple inlinings and non-inlined version. Currently I never tailcall while inlining, but that should go away.

jaykrell:
@monojenkins build Linux x64

jaykrell:

jaykrell:

PR #9630, opened 2018-07-18
Add GC Unsafe thread state transitions to MONO_API functions - Fourth Edition

lambdageek:
Part of #6921, this is a fourth PR that adds thread state transitions to switch to GC Unsafe state when embedders call API functions from outside of Mono.

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -4417,6 +4417,16 @@ mono_class_num_events (MonoClass *klass)
 MonoClassField*
 mono_class_get_fields (MonoClass* klass, gpointer *iter)
 {
+	MonoClassField *result;
+	MONO_ENTER_GC_UNSAFE;
+	result = mono_class_get_fields_internal (klass, iter);
+	MONO_EXIT_GC_UNSAFE;
+	return result;
+}
+
+MonoClassField*
+mono_class_get_fields_internal (MonoClass *klass, gpointer *iter)
+{
 	MonoClassField* field;
 	if (!iter)
 		return NULL;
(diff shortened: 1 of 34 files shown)


mono/metadata/class.c
```diff
@@ -4416,6 +4439,16 @@ mono_class_num_events (MonoClass *klass)
  */
 MonoClassField*
 mono_class_get_fields (MonoClass* klass, gpointer *iter)
+{
+	MonoClassField *result;
+	MONO_ENTER_GC_UNSAFE;
+	result = mono_class_get_fields_internal (klass, iter);
```
lewurm: why not `m_class_get_fields`?

lambdageek: Not sure what you're asking:
- I'm trying to keep the `m_class_get_*` namespace just for simple field accessors
- There's already `m_class_get_fields` - it directly returns a pointer to an array of `MonoClassField` elements; `mono_class_get_fields` is an iterator.

lambdageek:
Oops. Shouldn't have done `mono_dllmap_insert`, looks like.  Broke the coop build.

mono/metadata/class.c
lewurm: Ah okay. I noticed this inconsistency: usually it's something like
* `mono_class_get_namespace` -> GC_UNSAFE wrapper + `m_class_get_name_space`

here the pattern is different
* `mono_class_get_fields` -> GC_UNSAFE wrapper + `mono_class_get_fields_internal` ???

I guess the problem is that `mono_class_get_fields` should be really `mono_class_get_fielditerator` or something like that, but I understand we can't easily change that name. With your explanation it makes sense.

PR #9641, opened 2018-07-19
[Mono.Options] Properly handle nested quoting when parsing arguments …

jstedfast:
…from a file

Needed for vsts bug #650253

### DIFF
--- a/mcs/class/Mono.Options/Test/Mono.Options/ResponseFileTest.cs
+++ b/mcs/class/Mono.Options/Test/Mono.Options/ResponseFileTest.cs
@@ -1,0 +1,81 @@
+//
+// ResponseFileTest.cs
+//
+// Authors:
+//  Jeffrey Stedfast <jestedfa@microsoft.com>
+//
+// Copyright (C) 2018 Microsoft (http://www.microsoft.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+#if NDESK_OPTIONS
+using NDesk.Options;
+#else
+using Mono.Options;
+#endif
+
+using Cadenza.Collections.Tests;
+
+using NUnit.Framework;
+
+#if NDESK_OPTIONS
+namespace Tests.NDesk.Options
+#else
+namespace MonoTests.Mono.Options
+#endif
+{
+	[TestFixture]
+	public class ResponseFileTest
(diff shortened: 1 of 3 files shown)


marek-safar:
@jstedfast could you please add a test for this?

jstedfast:
@marek-safar what's the best way to provide a test case? Should I just attach the "response-file.rsp" that is getting loaded by mtouch using Mono.Options?

marek-safar:
See existing tests at `class/Mono.Options/Test/Mono.Options/`

jstedfast:
Ok, unit test added :)

mcs/class/Mono.Options/Test/Mono.Options/ResponseFileTest.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace MonoTests.Mono.Options
+#endif
+{
+	[TestFixture]
+	public class ResponseFileTest
+	{
+		[Test]
+		public void ResponseFileParser ()
+		{
+            const string text = "--arg1=value1\n\"--arg2=\\\"value2 contains nested quoting\\\"\"\n\"--arg3=value3\"\n";
```
rolfbjarne: There's a mix of spaces and tabs in this file.

marek-safar:
Interestingly, the windows tests fail as the file is still locked at Delete

jstedfast:
@marek-safar Perhaps the problem was because the code did not dispose the TextReader?

If so, hopefully my latest commit will fix that.

jstedfast:
Reverted the commit that disposed the TextReader as the `close` argument to `GetArguments()` already handled that.

luhenry:
@monojenkins build failed

jonpryor:
@marek-safar: Conceptual question: since `GetArguments()` was cribbed from mcs' `LoadArgs()`, should mcs be updated as well?  (Or is mcs "dead" and thus this question is moot?)

mcs/class/Mono.Options/Test/Mono.Options/ResponseFileTest.cs
```diff
@@ -0,0 +1,82 @@
...
+			IEnumerable<string> replacement;
+
+			File.WriteAllText (path, text);
+
+			try {
+				Assert.IsTrue (source.GetArguments ("@" + path, out replacement));
+			} catch {
+				throw;
+			} finally {
+				File.Delete (path);
```
jonpryor: The reason this [fails on Windows](
https://github.com/mono/mono/pull/9641#issuecomment-407397877) is because `replacement` is a [C# iterator *into* the file `path`](https://github.com/mono/mono/blob/d7337904202dd0c59b98dac3473d9fc5da39ab34/mcs/class/Mono.Options/Mono.Options/Options.cs#L638-L691), and the iterator won't close `path` until after `replacement` has been fully enumerated.

In retrospect this is a Bad Design™, because it means `replacement` can't be iterated over more than once, i.e. this won't work:

```csharp
if (replacement.Any ()) {
    foreach (var r in replacement) {
        // code is never reached!
    }
}
```

There are two plausible fixes:

 1. Change `ResponseFileSource.GetArguments()` so that `replacement` is fully iterated:

    ```csharp
    replacement = ArgumentSource.GetArgumentsFromFile (value.Substring (1)).ToArray ();
    ```

 2. Update this unit test so that `File.Delete(path)` happens *after* the `replacement.ToList()` invocation, line 74.

jonpryor:
I have one conceptual problem with this PR: it's infuriating.

Related: https://github.com/xamarin/xamarin-android/blob/38bc7d0d7705b580c8e0941affb3954b7de7d602/src/Xamarin.Android.Build.Tasks/Tasks/JavaCompileToolTask.cs#L78-L93

`javac` behaves as this PR would do, meaning that *all Windows paths must be escaped*, because -- you know -- Windows paths use `\` as the directory separator character.

This complicates the "understanding" of response files on Windows.  For *some* tools, you can use the *actual filesystem path* (what a concept!):

```
--whatever-flag C:\Users\foo\filename.txt
```

But with `javac` -- and *everything using Mono.Options after this patch* -- you *can't* use the actual filesystem path, it must instead be escaped!

```
--whatever-flag C:\\Users\\foo\\filename.txt
```

This is *incredibly* annoying.  On the plus side, this is *slightly* better than `javac`, in that `<tab>` won't be magically inserted into filenames.  Instead, you'll just "lose" the directory separator character on Windows, getting the "crazy" path of `C:Usersfoofilename.txt`.  (That'll make a *wonderfully* useful error message </sarcasm>.)

I am *very* tempted to just unilaterally *reject* this patch.  It would *break* ***all*** existing response files which currently work on Windows.

mcs/class/Mono.Options/Test/Mono.Options/ResponseFileTest.cs
marek-safar: @jstedfast could you update the test?

jstedfast:
The original issue was https://devdiv.visualstudio.com/DevDiv/_workitems/edit/650253

luhenry:
@marek-safar what's up with that now?

jstedfast:
Since the original issue was worked around in a different way, XI no longer needs this patch.

Since @jonpryor says that it will break XA, I'm in favor of just dropping this.

marek-safar:
Thanks, @jstedfast. Closing the PR.

PR #9642, opened 2018-07-19
[MacOS] On Mac, use the copyfile API to copy files

filipnavara:
Rebase of #9038 with added fix for the failing unit tests. Note that there's a race condition in the check between `source file == destination file` and the `copyfile` call, but a similar race condition already exists there in the current code except the outcome is different.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -521,7 +521,7 @@ AC_SUBST(export_ldflags)
 AC_PROG_LD
 AC_PROG_LD_GNU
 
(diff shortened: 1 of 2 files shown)


filipnavara:
Please don't merge this yet even if it passes the tests. It breaks IS_PORTABILITY_SET code path. Apparently that code path is not tested at all by the automated unit tests or it miraculously passes.

filipnavara:
@monojenkins build failed

filipnavara:
I am now happy enough with the patch :)

lewurm:
@monojenkins squash

migueldeicaza:
Do we want to consider backporting this to other branches?   Not the one for 15.8, but would be very nice for 15.9 and we will have ample time to test.

PR #9643, opened 2018-07-19
[interp] fix stack mismatch warning around ldlen and a C compiler warning

lewurm:
removes this warning in e.g. mcs-tests:

> Object.CallSite.Target: Store local stack type mismatch 0 1
> Object.CallSite.Target: Store local stack type mismatch 0 1
> Object.CallSite.Target: Store local stack type mismatch 0 1
> Object.CallSite.Target: Store local stack type mismatch 0 1
> Object.CallSite.Target: Store local stack type mismatch 0 1
> Object.CallSite.Target: Store local stack type mismatch 0 1

and fixes those warnings in the build

> interp/interp.c:3924:13: warning: unused variable 'valuetype_this' [-Wunused-variable]
>                         stackval valuetype_this;
>                                  ^
> interp/interp.c:3926:13: warning: unused variable 'retval' [-Wunused-variable]
>                         stackval retval;
>                                  ^
> 2 warnings generated.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3921,9 +3921,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, guint16 *st
 		MINT_IN_CASE(MINT_INTRINS_BYREFERENCE_CTOR) {
 			MonoClass *newobj_class;
 			MonoMethodSignature *csig;
-			stackval valuetype_this;
 			guint32 token;
-			stackval retval;
 
 			frame->ip = ip;
 			token = * (guint16 *)(ip + 1);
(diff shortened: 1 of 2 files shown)


lewurm:
@BrzVlad you are right, previous attempt was wrong. how about now?  `MINT_LDLEN` uses `mono_array_length_fast ()` which in the end depends on `MONO_BIG_ARRAYS`, so I don't think we should always load it as `int32`; instead we should set the stack type depending on `MONO_BIG_ARRAYS`.

BrzVlad:
@lewurm oh. big arrays are enabled by default on 64bit. Yeah we should do something about that. But I'd prefer to not use additional conv instructions since they add overhead and this is perf sensitive code

lewurm:
@monojenkins squash

lewurm:
fwiw, it was merged because the only failing step for linux x64 was the newly introduced "bclsizediff": https://jenkins.mono-project.com/job/test-mono-pull-request-amd64/16268/

PR #9650, opened 2018-07-19
[corlib]: Cleanup `X509Certificate` implementation and sync it with CoreFx.

baulig:
Completely refactored the `X509Certificate` implementation to make it based very closely on CoreFX.

We cannot fully switch to their implementation yet because we are missing things and they also have
a few issues in their implementation.  However, this will make it a lot easier to do this at a later time.

The internal `X509CertificateImpl` has been cleaned up as well, using properties instead of methods in
most places and the `ToString()` implementation is now shared between all platforms.

* `X509Certificate.CreateFromSignedFile()`: in CoreFX, this is the same as `CreateFromCertFile()`.  Our old implementation was incomplete and did not verify the signature - we even had a test where loading a certificate with an invalid signature was supposed to work; this test has been disabled because we're now throwing an exception.

* `X509CertificateImpl.KeyAlgorithmParameters`: this is supposed to throw `CryptographicException` on error, see `Certificate13` in `X509CertificateTest.cs`.

* We now match CoreFX's behavior regarding serialization - while the `X509Certificate` type is still marked as `[Serializable]`, we now throw `PlatformNotSupportedException` when attempting to serialize or deserialize it.


nealef:
Any plans for this to be backported to 2018-06?

baulig:
In general, backports should be used for critical bug fixes, though there is certainly room for discussion when it comes to fixing major annoyances.

However, this current work on the certificate code is essentially rolling out new features, it has cross dependencies on other ongoing BCL work (including CoreFX changes) while also doing major code refactoring.

From my point of view, I would draw the line with the Dependency Injector - from this point forward, the changes are getting too complex and have too many cross dependencies to be considered for backporting.

filipnavara:
I don't think this should be back-ported. As much I as I would like to have access to the new code as soon as possible this is just a small part of a much bigger task.

nealef:
We have a client using a build based off 2018-06 so I was just curious if I would end up having to add this or not. The least that gets added the better :-)

baulig:
Updated to reflect latest CoreFX changes regarding the obsolete `GetName()` / `GetIssuerName()` and their fix for serial numbers (see https://github.com/dotnet/corefx/pull/31268 and https://github.com/dotnet/corefx/pull/31269).

mcs/class/corlib/System.Security.Cryptography.X509Certificates/X509Certificate.cs
```diff
@@ -28,275 +28,682 @@
...
+{
 	// References:
 	// a.	Internet X.509 Public Key Infrastructure Certificate and CRL Profile
 	//	http://www.ietf.org/rfc/rfc3280.txt
-	
+
 	// LAMESPEC: the MSDN docs always talks about X509v3 certificates
 	// and/or Authenticode certs. However this class works with older
 	// X509v1 certificates and non-authenticode (code signing) certs.
 	[Serializable]
```
marek-safar: Let's keep the signature same for mobile and !mobile

baulig:

mcs/class/Mono.Btls.Interface/Mono.Btls.Interface/BtlsX509.cs
```diff
@@ -87,7 +87,10 @@ public byte[] GetPublicKeyData ()
 
 		public byte[] GetSerialNumber (bool mono_style)
 		{
-			return Instance.GetSerialNumber (mono_style);
+			var serial = Instance.GetSerialNumber (mono_style);
+			if (mono_style)
+				Array.Reverse (serial);
+			return serial;
```
filipnavara: What does `mono_style` mean? I was under the impression that it means "little endian", but then I don't understand why it would be both passed to the underlying implementation and then `Array.Reverse`d.

baulig: `MonoBtlsX509.GetSerialNumber(false)` returns the raw data as they're stored in the certificate.

`MonoBtlsX509.GetSerialNumber(true)` returns it in the format expected by CoreFX's PAL - which is big-endian plus padding.  This is not a direct reverse of the first and we have a test somewhere which exposes the difference.

The `Array.Reverse()` here in `Mono.Btls.Interface` (which is used by the web-tests) was added because the internal implementation (in `MonoBtlsX509`) switched from returning little-ending to big-ending (to match CoreFX PAL) - to avoid regressing the web-tests with a behavior change in the "semi-public" `Mono.Btls.Interface` code.

baulig: We could also rename `mono_style` into `dotnet_style` as it is now the same.

akoeplinger:

PR #9651, opened 2018-07-20
Use `SafePasswordHandle` in the internal certificate APIs.

baulig:
* `X509Helper`, `X509CertificateImpl`, `X509Certificate2Impl` and `ISystemCertificateProvider` now use `SafePasswordHandle` instead of strings.

* `X509Certificate`: actually implement `SecurePassword` based APIs using `SafePasswordHandle`.

* `mono_btls_pkcs12_import()`: when passed an empty password, try both the empty string and null.


marek-safar:

PR #9652, opened 2018-07-20
[CI] Add `corlib-btls` step to run the X509Certificate tests in corlib with BTLS as well.

baulig:
After the dependency injector changes, the `X509Certificate` code in corlib now uses the implementation from `System.dll`, including the platform-specific switch via `MONO_TLS_PROVIDER` / `MonoTlsProviderFactory`.

This adds a new `X509Certificates2` category to the corlib tests and runs them with `MONO_TLS_PROVIDER=btls` on OS X (where the default is applets).


marek-safar:
I thought BTLS on macos is not the combination we want/need to support

baulig:
It always been fully supported on OS X - and it fact this has been a strict requirement right from the start and we also rely on AppleTls/BTLS cross-testing for a few scenarios in the web-tests.  We have therefore, always run the System tests with both AppleTls and BTLS on CI and nothing has changed in this regard.

However, the dependency injector shuffled a bunch of code around and I don't want to expose possible untested code paths.

PR #9661, opened 2018-07-20
[mini] return gracefully from compile_special if a method can not be compiled

lewurm:
the caller will check `error` anyway. It allows the machinery to try the
interpreter instead.


lewurm:
@monojenkins rebase

lewurm:
@monojenkins backport 2018-06

PR #9665, opened 2018-07-20
Move generated resource files from System and System.Security into corlib.

baulig:
* Make `corlib` internals visible to `System.Security` on all profiles, not just monotouch watch.
  
* Moves all generated `SR` files and their generation logic from `System` and `System.Security` into `corlib`.  Since `corlib` internals are visible to these assemblies, we don't need to duplicate the resource strings.

* Cleanup the `RESX_RESOURCE_STRING` is `corlib` and sort them alphabetically.

* Add the internal `Internal.Cryptography.Helpers` helper class from CoreFX to `corlib` (currently unused, but "Blue" will need it).


baulig:
I took the `../../../external/corefx/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs` addition out of "Blue" and added a local copy there, so we don't depend on this.

"Code Ultramarine" (coming shortly) then actually use the real file.

baulig:

baulig:

marek-safar:

PR #9666, opened 2018-07-20
[coop] Only take a snapshot of the stack on watchOS and tvOS

lambdageek:
Elsewhere we have a usable MonoContext so this is just unnecessary work.

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -166,6 +166,7 @@ return_stack_ptr (gpointer *i)
 static void
 copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
 {
+#if defined (HOST_WATCHOS) || defined (HOST_TVOS)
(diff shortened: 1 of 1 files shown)


mono/utils/mono-threads-coop.c
```diff
@@ -166,6 +166,7 @@ return_stack_ptr (gpointer *i)
 static void
 copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
 {
+#if defined (HOST_WATCHOS) || defined (HOST_TVOS)
```
jaykrell: Can we use runtime if against constants? i.e. so this code continues get type checked on regular builds?

lambdageek: That's a good point.  Actually the stack copying / scanning code  should probably also run on a regular basis on a more mundane platform so that we get at least some reassurance that it's working.

I'll rework the PR.

lambdageek:
This wasn't the right approach. I'll revisit later.

PR #9667, opened 2018-07-21
A related take on using cleaned up threads. 

jaykrell:
Based on findings and work by @alexanderkyte.
This is my variation of https://github.com/mono/mono/pull/9486.
It might need more work, i.e. if the new assert triggers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=25
+MONO_CORLIB_COUNTER=30
 MONO_CORLIB_VERSION=`printf "1%02d%02d%02d%03d" $MONO_VERSION_MAJOR $MONO_VERSION_MINOR 0 $MONO_CORLIB_COUNTER`
 
 AC_DEFINE_UNQUOTED(MONO_CORLIB_VERSION,$MONO_CORLIB_VERSION,[Version of the corlib-runtime interface])
--- a/mono/metadata/object-internals.h
+++ b/mono/metadata/object-internals.h
@@ -429,12 +429,20 @@ typedef enum {
 
 struct _MonoThreadInfo;
 
+typedef union _MonoInternalThreadLockFlags {
+	gpointer i;
+	struct {
+		gboolean free : 1;
+		gsize count : (sizeof (gpointer) * 8) - 1;
+	};
+} MonoInternalThreadLockFlags;
+
 struct _MonoInternalThread {
 	MonoObject  obj;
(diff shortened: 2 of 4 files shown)


jaykrell:

mono/metadata/threads.c
```diff
@@ -1603,18 +1692,14 @@ ves_icall_System_Threading_InternalThread_Thread_free_internal (MonoInternalThre
 	CloseHandle (this_obj->native_handle);
 #endif
 
-	if (this_obj->synch_cs) {
-		MonoCoopMutex *synch_cs = this_obj->synch_cs;
-		this_obj->synch_cs = NULL;
-		mono_coop_mutex_destroy (synch_cs);
-		g_free (synch_cs);
-	}
+	// This can initialize on-demand.
```
jaykrell: This comment should be up on the lock_thread (which wasn't there at the time).

configure.ac
```diff
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=25
+MONO_CORLIB_COUNTER=30
```
jaykrell: This is just because master monolite 25 is currently broken. This isn't really needed.

mono/metadata/object-internals.h
```diff
@@ -429,12 +429,20 @@ typedef enum {
 
 struct _MonoThreadInfo;
 
+typedef union _MonoInternalThreadLockFlags {
+	gpointer i;
```
jaykrell: I wanted to use gsize here, but the available atomics..

mono/metadata/threads.c
```diff
@@ -472,27 +472,113 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
-static inline void
+static void
 unlock_thread (MonoInternalThread *thread)
 {
+	// dec count and read free
+	MonoInternalThreadLockFlags flags;
+	flags.i = mono_atomic_load_ptr (&thread->lock_flags.i);
+	while (1) {
+		MonoInternalThreadLockFlags old_flags = flags;
+		g_assertf (flags.count > 0, "count:%d free:%d", flags.count, flags.free);
```
jaykrell: I only elabored this to `f` because I had typoed struct vs. union..

mono/metadata/threads.c
```diff
@@ -472,27 +472,113 @@ static void ensure_synch_cs_set (MonoInternalThread *thread)
...
+		old.i = mono_atomic_cas_ptr (&thread->lock_flags.i, flags.i, old_flags.i);
+		if (old.i == old_flags.i)
+			break;
+		flags = old;
+	}
+
 	mono_coop_mutex_unlock (thread->synch_cs);
+
+	if (flags.free && flags.count == 0) {
+		MonoCoopMutex *synch_cs = thread->synch_cs;
```
jaykrell: Consider cas or swap here, rare path, shouldn't be needed. In fact should be safe to leave the non-null old value.

jaykrell:

PR #9669, opened 2018-07-21
[Coop] Convert System.Environment and System.IO.DriveInfo, dial back HANDLES() for NOHANDLES().

jaykrell:


vargaz:
A lot of these use no handles, so it would be easier and faster to just mark them with NOHANDLES in icall-def.h.

jaykrell:
I know -- and they have no error paths -- and I'm torn on the matter.

I'd kinda like to see HANDLES() on every line.

To whatever extent a signature has no handles, marshal-ilgen should figure that out pretty quickly and generate the same code. And the C code won't change. But MonoError* will get added, which might also be useless.

NOHANDLES it kinda not enforced, but we do have the Centrinel checking and the more important goal, than wrapping in HANDLES() is elimination of raw pointers, granted.

vargaz:
HANDLES() has an overhead even if the the signature uses no handles, mono error needs to be initialized and checked, both are done by calling into the runtime, i.e. mono_icall_start ()/mono_icall_end () which do other stuff as well like saving stack state etc.

jaykrell:
mono_stack_mark_init looks like about 8 instructions ("counting the arrows", one mov each).
The marshal-ilgen.c code could optimize that, if there are no handles, i.e. to a different begin/end.
It could also warn, have the user (me) check for use of the MonoError and suggest NOHANDLES.
I acknowledge there is the MonoError stack allocation and zeroing, and checking.
I assume we have to check the thread for pending, and the MonoError if present.

I also think ultimately some of this code will fall away with CoreFX imports, i.e. "all managed" but maybe only some of it (i.e. esp. and less interesting the Windows paths).

I'd still rather the consistency, as regarded-as-dumb as that might be.

mono_stack_mark_pop looks like another 8, and a barrier, hm.
Handles used to have barriers for assignment, I wonder if this is similarly leftover.

Another concern I have here is simply the cost of native-to-native function calls, in the face of ELF-interposability. ves_icall_foo => mono_foo might cost beyond regular mono_foo. Even Windows could have a jmp. or worse on x86 (the mismatched MonoError parameter).
Maybe we turn off those rules though and mono-to-mono is not interposable.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build failed

jaykrell:
Ok, I may have missed some spots, but I dialed back unnecessary HANDLES() in this PR, and preexisting, and added some NOHANDLES.

jaykrell:

jaykrell:
Flake?

Another https://jenkins.mono-project.com/job/test-mono-pull-request-amd64/16524/parsed_console/log.html

```
runtimemetadataversion:v4.0.30319 -d:MONO_FEATURE_CONSOLE     -target:library -out:../../class/lib/basic/mscorlib.dll  @./../../build/deps/_basic__corlib.dll.response

Unhandled Exception:
System.ExecutionEngineException: System.NullReferenceException: Object reference not set to an instance of an object ---> System.NullReferenceException: Object reference not set to an instance of an object
   --- End of inner exception stack trace ---
  at System.Environment.FailFast (System.String message, System.Exception exception) [0x00007] in <71d8ad678db34313b7f718a414dfcb25>:0 
  at Microsoft.CodeAnalysis.FailFast.OnFatalException (System.Exception exception) [0x00038] in <b28b918e025b48209f28967b441780d7>:0 
  at Microsoft.CodeAnalysis.FatalError.Report (System.Exception exception, System.Action`1[T] handler) [0x00080] in <b28b918e025b48209f28967b441780d7>:0 
```

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

jaykrell:
This needs a merge/rebase, but review?

jaykrell:
This will also conflict with the externC work but ok.

jaykrell:
@monojenkins build deb with  monolite

jaykrell:

jaykrell:
@monojenkins build

jaykrell:
oops, meant build failed, alas

jaykrell:
Was good, now needs rebase.

jaykrell:

akoeplinger:
@monojenkins build failed

jaykrell:
I'm gradually picking this PR apart into smaller ones..

PR #9670, opened 2018-07-21
Use msbuild project reference to establish dependency on genconsts instead of solution dependencies

kg:
Using solution dependencies in ```bcl.sln``` seems flaky and seems like it might not establish the full ordering we need to ensure that ```Consts.cs``` exists before we build things that require it. Let's try using project references (where ```corlib.dll``` 'depends' on ```genconsts.exe```) instead. This should also insert the dependency for any project that includes Consts.cs instead of just corlib.

This PR also makes update-solution-files actually fail if ```genconsts.exe``` fails to build because it was driving me mad.

Part of #6886


kg:
cc @rainersigwald: You suggested using project references this way, right? Does it look like I did it correctly or did you mean something else? It seems to work in my testing.

marek-safar:
Why do we have to add it to every project and adding it to mscorlib is not enough?

msvc/scripts/genproj.cs
```diff
@@ -884,6 +883,19 @@ string groupConditional
 			}
 		).ToList ();
 
+		if (commonFiles.Any (f => f.EndsWith("build\\common\\Consts.cs"))) {
+			var genconstsRelativePath = "$(SolutionDir)\\msvc\\scripts\\genconsts.csproj";
+			result.Append ($"  <ItemGroup>{NewLine}");
+			result.Append ($"    <ProjectReference Include=\"{genconstsRelativePath}\">{NewLine}");
+			result.Append ($"      <Name>genconsts</Name>");
+			result.Append ($"      <Project>{SlnGenerator.genconsts_csproj_guid}</Project>");
```
rainersigwald: `Name` and `Project` are generally not required by command-line builds, if that's relevant to you here; they could potentially be dropped--but there's no need to do so since you already have the information.

marek-safar:
Yeah, I am not sure which way is less painful. This was the msbuild will have to work extra hard to remove the duplication as we want the project to build only once.

The Const.cs is used mostly by libraries and I'd need to double check but it's possible we don't use it for any tool.

kg:
@monojenkins commit csproj

kg:
@monojenkins commit csproj

kg:
@marek-safar I experimented but I can't find a way to omit the dependency from non-corlib projects without breaking tools. We have some exes we build and run at build time that depend on Consts.cs. Do you want me to put in some sort of one-off workaround where I only generate the dependency for corlib-or-executables instead of for any project that uses Consts? I'd rather keep it simple but I can probably build that logic in.

PR #9671, opened 2018-07-21
Use MONO_KEYWORD_THREAD for int last_error and GSList* load_type_info.

jaykrell:
The macros here should probably be used for the others.


jaykrell:
@monojenkins build failed

jaykrell:

PR #9672, opened 2018-07-21
gcc/win32 emulates __thread, which is not clearly worthwhile.

jaykrell:


jaykrell:

PR #9675, opened 2018-07-21
[runtime] Return the result properly from mono_exception_walk_trace ().

vargaz:


lewurm:
@monojenkins squash

lewurm:

PR #9677, opened 2018-07-21
[runtime] Add a drive type entry for the Paragon NTFS drive on macos.

vargaz:


jaykrell:
Is there an easier to maintain way? Can/should we assume unknown types are fixed? Does it matter much either way?

vargaz:
@monojenkins build failed

Issue #9621, opened 2018-07-18
64 bit cross compilers targeting 32 bit platforms


marek-safar:
We'll have to come up with a solution to update Mono cross-compilers to handle the bitness difference, i.e. generate 32bits code from a 64bits process.
 
The main motivation is because Apple is planning to remove 32bits support from macOS.
 
This affects Xamarin.iOS tooling since the current Mono AOT cross-compiler needs to be an 32bits executable to produce 32bits code, e.g. ARMv7[s][k]. This is a problem because we’ll still require 32bits support for watchOS and iOS. The similar situation could happen with Android.
 
Apple confirmed that macOS 10.15 will not ship with 32bits slices of any system libraries. Running the watchOS simulator (and likely old iOS ones) will work since it is (or will be) shipped as a self-contained package (the 32bits libraries won’t be for macOS).

## Milestone I

- [x] Update Mono sources where needed to have the configurable target size
- [x] Check the output binary is same when cross compiling as with 32 bits 

## Milestone II
- [x] Produce cross compiler output as part of SDK target for iOS
- [x] Produce cross compiler output as part of SDK target for Android

## Milestone III
- [ ] Figure out Xamarin.iOS integration
- [ ] Figure out Xamarin.Android integration
- [x] switch to python offset tool.

vargaz:
Short documentation:
- any code which refers to host sizes/offsets and can influence generated code needs to use target sizes/offsets. I.e. sizeof(gpointer) should be sizeof(target_mgreg_t), etc.
- code which depends on the sizes/offsets being the host ones like allocation, corlib version checks, need to be disabled.

lewurm:
Related to this issue:
```console
$ file Xamarin.iOS.framework/Versions/git/LLVM*/bin/llc
Xamarin.iOS.framework/Versions/git/LLVM/bin/llc:   Mach-O 64-bit executable x86_64
Xamarin.iOS.framework/Versions/git/LLVM36/bin/llc: Mach-O executable i386
```
Alternatively we could try to get rid of LLVM36.

vargaz:
The 32 bit executables are not really needed, they are only there to simplify some build scripts.

vargaz:
So now we have ios-cross32-64 targets in builds/. They should be drop-in replacements for the normal 32 bit cross compilers. Also, these cross compilers are used by wasm, so they are somehow tested.

PR #9678, opened 2018-07-21
Cross32 changes.

vargaz:
Part of #9621

### DIFF
--- a/mono/metadata/abi-details.h
+++ b/mono/metadata/abi-details.h
@@ -61,4 +61,6 @@ enum {
 #endif
 #endif
 
+#define MONO_SIZEOF_MonoObject (2 * MONO_SIZEOF_gpointer)
+
 #endif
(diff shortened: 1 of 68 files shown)


kumpera:
Hey Zoltan, can you update the man page and —AOT-help?

And can you start a page on the website explaining the dos and donts to make this work?

migueldeicaza:
Folks, before we embark on these disruptive changes to the runtime, I want to consider cheaper alternatives to this.   For example, using QEmu/Userland to run the 32-bit binaries in 64-bit mode or using the WebAssembly work to run inside WebAssembly.

jaykrell:
Suggest to put the name "target" in somewhere. target_sizeof, target_offsetof, target_mono_foo maybe.

Many but not all compilers are structured this way and "fixing" would enable many other scenarios, that may or may not be interesting. I.e. arbitrary cross AOT. Like people get from all their C and C++ compilers.

Lack of distinction between host and target in a compiler is kinda a short term hack.

vargaz:
Was merged in other PRs.

PR #9680, opened 2018-07-22
[Coop]  Conversion of CreateProcess and ShellExecute.

jaykrell:
… native/managed interface.

Replace them with long parameter lists.
This will ease 1. Using "fixed" 2. Trimming the native/managed interface.
Presently the managed/native interface is inflated by essentially the managed/managed interface.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,7 +40,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This can be reset to 0 when Mono's version number is bumped
 # since it's part of the corlib version (the prefix '1' in the full
 # version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=25
+MONO_CORLIB_COUNTER=33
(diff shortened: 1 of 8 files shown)


jaykrell:

vargaz:
This looks kinda ugly, we put parameters in structures to avoid these long parameter lists.

jaykrell:
It is ugly either way. Parameters in struct have 3 disadvantages: 1 native/managed must be slightly more councident (similar to parameters though), 2 can’t “fixed” the fields 3 can’t trim unused ones

jaykrell:
Specifically this enabled “fixing” the strings & it will allow trimming.

jaykrell:
@monojenkins build failed

jaykrell:
This is also incomplete (do not merge). The trimming isn’t done. The parameter list also eliminates the managedness/handle of the first parameter. 

The code is not performance sensitive, but I’d like to get it simpler. Maybe move the env var handling to managed (can’t fixed array of strings).

jaykrell:
@monojenkins build failed

jaykrell:
Will try more conventional conversion.

jaykrell:
Will probably do this differently (almost done).

Issue #9532, opened 2018-07-13
Mono.Unix.Native.Syscall.sys_strerror_r SEGFAULT on musl libc


frebib:
## Steps to Reproduce

1. On any Alpine Linux/musl based OS (`docker run -ti --rm spritsail/alpine-sdk`)
2. `apk add mono` (from `edge/testing` repository)
3. `csc <the file below>`
```c#
using System;
using Mono.Unix;

namespace Mono.Unix {
    public class UnixTest {
        public static void Main() {
            Console.WriteLine("EINVAL: " + UnixMarshal.GetErrorDescription(Native.Errno.EINVAL));
        }
    }
}
```
4. `mono <the file above>.exe`
5. Crash!

## Cause of the issue

The Mono installation I am running is built from source. Observing the compilation of the `support/` libraries, namely `libMonoPosixHelper.so` with `cd support; make V=1`.
The offending source `errno.c` which contains the definition for the native half of `Mono.Unix.Native.Syscall.sys_strerror_r` compiles as such:

```
gcc -DHAVE_CONFIG_H -I. -I..  -I../mono/eglib -I../mono/eglib -I.. -Os -fomit-frame-pointer -DGC_LINUX_THREADS -D_GNU_SOURCE -D_REENTRANT -DUSE_MMAP -DUSE_MUNMAP -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length  -DUSE_COMPILER_TLS  -Os -fomit-frame-pointer -fno-strict-aliasing -std=gnu99 -fno-strict-aliasing -fwrapv -DMONO_DLL_EXPORT -Wno-unused-but-set-variable -g -Wall -Wunused -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes  -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wno-cast-qual -Wwrite-strings -Wno-switch -Wno-switch-enum -Wno-unused-value -Wno-attributes -Wno-format-zero-length -mno-tls-direct-seg-refs -Werror-implicit-function-declaration -c -o errno.lo errno.c
```

[...]

PR #9691, opened 2018-07-23
Fix strerror_r test macro (issue #9532)

frebib:
According to strerror_r(3):
```
    The XSI-compliant version is provided if:
    (_POSIX_C_SOURCE >= 200112L) && !  _GNU_SOURCE
    Otherwise, the GNU-specific version is provided.
```

Fixes https://github.com/mono/mono/issues/9532

Signed-off-by: Joe Groocock <me@frebib.net>

### DIFF
--- a/support/errno.c
+++ b/support/errno.c
@@ -41,7 +41,7 @@ Mono_Posix_Stdlib_SetLastError (int error_number)
  * we assume that the XPG version is present.
  */
 
-#ifdef _GNU_SOURCE
+#if (_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE
 #define mph_min(x,y) ((x) <= (y) ? (x) : (y))
 
 /* If you pass an invalid errno value to glibc 2.3.2's strerror_r, you get


frebib:
I wonder if this should actually use the exact macro test from the manpage, instead of the one I found suggested in this post from ~18 months ago? The difference being the additional ` || _XOPEN_SOURCE >= 600` check which appears to have since been removed.

I'm not massively familiar with these macros but I'll do a bit of digging to see what _should_ be used.

lewurm:
@monojenkins approve

frebib:
Any reasons the CI is failing? I highly doubt it's my doing

lewurm:
hey @frebib sorry for the delay, the CI results are looking good.  could you reopen the PR please?

frebib:
I noticed that whilst looking at this earlier I seem to have inverted the check. Even though it worked on Alpine, I wanna test it against glibc

frebib:
@lewurm After thinking about it, this is definitely not a fix.  Aside from the fact that it is straight-up inverted and therefore incorrect, it avoids the underlying problem that `_GNU_SOURCE` is defined even on systems that _don't_ use a GNU libc implementation.
I feel like this should be discussed further because realistically the fix is a lot more complex and will likely cause more issues.

PR #9697, opened 2018-07-23
Remove excess code -- strings are already nul terminated.

jaykrell:
For coop, this loop needs to be a separate function. This change reduces that. Or so I thought -- that isn't true, but the change is still ok.


jaykrell:
@monojenkins build failed

PR #9699, opened 2018-07-23
[Coop] Convert CreateProcess and ShellExecute.

jaykrell:


mono/metadata/w32process-win32.c
```diff
@@ -202,131 +199,141 @@ process_quote_path (const gchar *path)
...
-			/* null-separated */
-			len += sizeof (gunichar2);
+		for (gsize i = 0; i < array_length; i++) {
+			MONO_HANDLE_ARRAY_GETREF (var, array, i);
+			len += mono_string_handle_length (var) + 1; // nul-separated
 		}
-		/* null-terminated */
-		len += sizeof (gunichar2);
-
-		env_vars = ptr = g_new0 (gunichar2, len);
```
jaykrell: Code was multiplying by sizeof array element twice.

mono/metadata/w32process-win32.c
```diff
@@ -202,131 +199,141 @@ process_quote_path (const gchar *path)
...
 	
-	if (process_get_shell_arguments (proc_start_info, &cmd) == FALSE) {
+	if (process_get_shell_arguments (coop, &cmd, error) == FALSE) {
 		process_info->pid = -ERROR_FILE_NOT_FOUND;
-		return FALSE;
+		ret = FALSE;
+		goto exit;
 	}
 
 	if (process_info->env_variables) {
```
jaykrell: Is a raw pointer ok if just checking for null?

jaykrell: Could the compiler cache it on the line that makes a handle out of it?

mono/metadata/w32process-unix.c
```diff
@@ -1913,25 +1911,28 @@ process_create (const gunichar2 *appname, const gunichar2 *cmdline,
 	 * new process inherits the same environment.
 	 */
 	if (process_info->env_variables) {
-		gint i;
-		MonoString *var;
+		MonoArrayHandle array = MONO_HANDLE_NEW (MonoArray, process_info->env_variables);
```
lambdageek: I see `MonoW32ProcessInfo` is a managed struct that's been pinned and passed by ref.

mono/metadata/w32process-unix.c
```diff
@@ -2091,32 +2085,30 @@ process_create (const gunichar2 *appname, const gunichar2 *cmdline,
...
 MonoBoolean
-ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStartInfo *proc_start_info, MonoW32ProcessInfo *process_info)
+ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStartInfoHandle proc_start_info, MonoW32ProcessInfo *process_info, MonoError *error)
 {
-	const gunichar2 *lpFile;
-	const gunichar2 *lpParameters;
-	const gunichar2 *lpDirectory;
-	gunichar2 *args;
+	MonoCreateProcessCoop coop;
+	mono_createprocess_coop_init (&coop, proc_start_info, process_info);
```
lambdageek: I like how this works

mono/metadata/w32process-win32.c
lambdageek: I think this is okay.

mono/metadata/w32process-unix.c
```diff
@@ -47,14 +47,14 @@
 #include <utime.h>
 #endif
 
+#include <mono/metadata/object-internals.h>
```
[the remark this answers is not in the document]
jaykrell: Yes, I needed TYPED_DECL in w32process.h.

mono/metadata/w32process-unix.c
jaykrell: Not quite. I pinned its constituents.

mono/metadata/w32process-unix.c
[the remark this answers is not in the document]
jaykrell: Yes. w32process.h could not used TYPED_DECL for a coop handle otherwise.

mono/metadata/w32process-unix.c
lambdageek: Yes - this seems fine - the managed ref constituents need to be accessed either by pinning or via handles, as you've done.  I was just surprised to see this use of `MONO_HANDLE_NEW` - I don't think we often pass struct like this.

jaykrell: Imho we should revise the managed/native interface here, but it isn't performance sensitive and there are many fish to fry. Remove the two structs and pass the constituents imho, `fixed` the strings, `handles()` the array, ref the two actual out  parameters, pass the other values directly, etc.

mono/metadata/w32process-unix.c
jaykrell: Indeed. The approximate code would have been duplicated 4 times otherwise.

mono/metadata/w32process.c
```diff
@@ -672,3 +672,48 @@ ves_icall_System_Diagnostics_Process_GetProcessData (int pid, gint32 data_type,
...
+		*length = mono_string_handle_length (in_coophandle);
+	}
+}
+
+void
+mono_createprocess_coop_init (MonoCreateProcessCoop *coop, MonoW32ProcessStartInfoHandle proc_start_info, MonoW32ProcessInfo *process_info)
+{
+	memset (coop, 0, sizeof (*coop));
+
+#define PIN_STRING(h, x) (mono_pin_string (h, &coop->coophandle.x, &coop->x, &coop->length.x, &coop->gchandle.x))
```
jaykrell: Probably should be `handle` and `field` or `name`.

mono/metadata/w32process.h
```diff
@@ -53,21 +55,68 @@ typedef struct
...
+	MonoObject *unused_envVars;
+
 } MonoW32ProcessStartInfo;
 
+TYPED_HANDLE_DECL (MonoW32ProcessStartInfo);
+
+typedef uint32_t gchandle_t; // FIXME use this more, make it typesafe.
+
+typedef struct _MonoCreateProcessCoop {
+	gunichar2 *filename;
```
jaykrell: We could have a struct of ptr/handle/gchandle/length, but elsewhere we treat the  gchandles as an array.

mono/metadata/w32process.h
```diff
@@ -100,11 +149,11 @@ void
 ves_icall_System_Diagnostics_FileVersionInfo_GetVersionInfo_internal (MonoObject *this_obj, MonoString *filename);
 
 MonoBoolean
-ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStartInfo *proc_start_info, MonoW32ProcessInfo *process_handle);
+ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStartInfoHandle proc_start_info_handle, MonoW32ProcessInfo *process_info, MonoError *error);
 
 MonoBoolean
-ves_icall_System_Diagnostics_Process_CreateProcess_internal (MonoW32ProcessStartInfo *proc_start_info, gpointer stdin_handle,
-	gpointer stdout_handle, gpointer stderr_handle, MonoW32ProcessInfo *process_handle);
+ves_icall_System_Diagnostics_Process_CreateProcess_internal (MonoW32ProcessStartInfoHandle proc_start_info_handle, gpointer stdin_handle,
```
jaykrell: I never ended up with separate handle/ptr variables, so the parameter can be renamed to not say `handle`.

jaykrell: And above, and maybe elsewhere. i.e. I was going to pin or fix it, but didn't.

PR #9701, opened 2018-07-24
Be sure always assign out parameter.

jaykrell:


mono/metadata/icall.c
```diff
@@ -1135,6 +1135,8 @@ ves_icall_System_ValueType_InternalGetHashCode (MonoObject *this_obj, MonoArray
 ICALL_EXPORT MonoBoolean
 ves_icall_System_ValueType_Equals (MonoObject *this_obj, MonoObject *that, MonoArray **fields)
 {
+	*fields = NULL;
+
```
vargaz: This should come after the declarations imho.

jaykrell: Ok, but it still has to move up from where it was.

PR #9703, opened 2018-07-24
[Coop] Convert System.TypedReference (part handles(), part nohandles())

jaykrell:


mono/metadata/icall.c
```diff
@@ -7526,18 +7526,17 @@ mono_ArgIterator_IntGetNextArgType (MonoArgIterator *iter)
...
-	MonoObjectHandle result = typed_reference_to_object (tref, error);
-	mono_error_set_pending_exception (error);
-	ICALL_RETURN_OBJ (result);
+	return typed_reference_to_object (tref, error);
 }
 
 ICALL_EXPORT MonoTypedRef
 mono_TypedReference_MakeTypedReferenceInternal (MonoObject *target, MonoArray *fields)
 {
+	MONO_REQ_GC_UNSAFE_MODE; // because of "p", so not worth converting target and  fields
```
jaykrell: Two spaces here due to Mac keyboard problem.

mono/metadata/icall.c
```diff
@@ -7524,18 +7524,17 @@ mono_ArgIterator_IntGetNextArgType (MonoArgIterator *iter)
...
-	MonoObjectHandle result = typed_reference_to_object (tref, error);
-	mono_error_set_pending_exception (error);
-	ICALL_RETURN_OBJ (result);
+	return typed_reference_to_object (tref, error);
 }
 
 ICALL_EXPORT MonoTypedRef
 mono_TypedReference_MakeTypedReferenceInternal (MonoObject *target, MonoArray *fields)
 {
+	MONO_REQ_GC_UNSAFE_MODE; // because of "p", so not worth converting target and fields
```
lambdageek: `MONO_REQ_GC_UNSAFE_MODE` means "the thread must be in the `RUNNING` thread state" - it doesn't have anything to do with garbage collection, directly.  In particular it doesn't mean "Do not perform garbage collection while in this scope" - which is what you intended, I think.

jaykrell:
I want to disallow GC or at least precise/moving GC. How?

lambdageek:
> I want to disallow GC or at least precise/moving GC. How?

I don't think we have anything like that defined.

`MONO_PREPARE_GC_CRITICAL_REGION`/`MONO_FINISH_GC_CRITICAL_REGION` + `MONO_REQ_GC_CRITICAL` might be close but I don't think it has the right semantics (it's unused so i'm only going by the comment in `checked-build.h`).

jaykrell:
How about I do nothing code-wise, and just a comment for now. Small progress (esp. the other function).

jaykrell:
@monojenkins build failed

jaykrell:
arm64 flaked

jaykrell:
Review?

jaykrell:
Review?

mono/metadata/icall-def.h
```diff
@@ -1057,8 +1057,10 @@ HANDLES(ICALL(TYPE_1, "internal_from_handle", ves_icall_System_Type_internal_fro
 HANDLES(ICALL(TYPE_2, "internal_from_name", ves_icall_System_Type_internal_from_name))
 
 ICALL_TYPE(TYPEDR, "System.TypedReference", TYPEDR_1)
-ICALL(TYPEDR_1, "InternalToObject",	mono_TypedReference_ToObject)
-ICALL(TYPEDR_2, "MakeTypedReferenceInternal", mono_TypedReference_MakeTypedReferenceInternal)
+HANDLES(ICALL(TYPEDR_1, "InternalToObject",	mono_TypedReference_ToObject))
+
+// Has raw pointers but not easily convertable at this time.
+NOHANDLES(ICALL(TYPEDR_2, "MakeTypedReferenceInternal", mono_TypedReference_MakeTypedReferenceInternal))
```
lambdageek: I don't think we should use NOHANDLES on functions that still require our attention.

mono/metadata/icall.c
```diff
@@ -7524,23 +7524,21 @@ mono_ArgIterator_IntGetNextArgType (MonoArgIterator *iter)
...
-	MonoObjectHandle result = typed_reference_to_object (tref, error);
-	mono_error_set_pending_exception (error);
-	ICALL_RETURN_OBJ (result);
+	return typed_reference_to_object (tref, error);
 }
 
 ICALL_EXPORT MonoTypedRef
 mono_TypedReference_MakeTypedReferenceInternal (MonoObject *target, MonoArray *fields)
 {
+	guint8 *p = NULL; // Inhibits coop conversion at this time, so not worth converting target and fields.
```
lambdageek: Needs a `// FIXME` somewhere around here.

mono/metadata/icall.c
```diff
@@ -7524,23 +7524,21 @@ mono_ArgIterator_IntGetNextArgType (MonoArgIterator *iter)
...
-ICALL_EXPORT MonoObject*
-mono_TypedReference_ToObject (MonoTypedRef* tref)
+ICALL_EXPORT MonoObjectHandle
+mono_TypedReference_ToObject (MonoTypedRef* tref, MonoError *error)
 {
-	ICALL_ENTRY ();
-	MonoObjectHandle result = typed_reference_to_object (tref, error);
-	mono_error_set_pending_exception (error);
-	ICALL_RETURN_OBJ (result);
+	return typed_reference_to_object (tref, error);
```
lambdageek: This looks ok.

mono/metadata/icall.c
jaykrell: ok

mono/metadata/icall-def.h
jaykrell: ok yeah I was thinking that tood.

jaykrell: I removed the marker, edited the command, and big bonus: mostly converted the function. The `p` variable remains.

jaykrell:

jaykrell:
Better now?

jaykrell:
It appears `MonoTypedRef` is defined as:

```
    public struct TypedReference
    {
        RuntimeTypeHandle type;
        private IntPtr Value;
        private IntPtr Type;
}
```

and ```
struct RuntimeTypeHandle { IntPtr value; }
```

so returning that by value *should* be ok.
However I'm open to other.
In particular, maybe we don't want to be passing/returning structs by value.

jaykrell:
The undo is just temporary to unblock myself while master is broken.

jaykrell:

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build Linux x64

PR #9713, opened 2018-07-24
Add support for Blazor custom interop support.

kjpou1:
This adds additional ability to perform unmarshalled calls for advanced but high-performance shared-memory interop.
* Note: * This is only safe in Blazor's case because of its exact usage patterns.

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -76,13 +76,13 @@ debug/:
 release/:
 	mkdir -p $@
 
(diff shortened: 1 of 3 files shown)


sdks/wasm/Makefile
```diff
@@ -181,6 +181,7 @@ package: build
 	cp $(TOP)/sdks/out/wasm-interp/lib/libmonosgen-2.0.a tmp/
 	cp library_mono.js tmp/
 	cp binding_support.js tmp/
+	cp dotnet_support.js tmp/
```
kumpera: The dotnet_suppport.js file is missing.

kumpera:

PR #9715, opened 2018-07-24
[System]: Remove two `X509Certificate2Test` tests.

baulig:
Both `Certificate_11_Properties` and `Certificate_14_Properties` use an invalid
certificate that violates X.690 11.8.2 (UTCTime under a DER encoding).

This will no longer be supported once we switch to CoreFX's PAL code.

See also https://github.com/dotnet/corefx/pull/31293.


baulig:

PR #9716, opened 2018-07-24
Fix issue where get offset could return 0 when the type is uninitialized

mderoy:
We have some example code that gets all the fields via reflection
```foreach (var field in type.GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public))```
Then later when we try to get the field offset, the type hasn't been initialized (unless your C# code does something earlier that initializes the type).
type.GetFields calls mono_class_get_fields_lazy, which doesn't setup all the field info (in this case the offset)

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -4754,6 +4754,7 @@ mono_field_get_flags (MonoClassField *field)
 guint32
 mono_field_get_offset (MonoClassField *field)
 {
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

marek-safar:
Any chance to include your repro as the test?

marek-safar:
Sadly we don't have any embedding testing in place.

PR #9720, opened 2018-07-24
BigStep: drop some LLVM in there

lambdageek:
It's doing something...

but TestSimpleRet says
```
1) TestSimpleRet (MonoTests.Mono.CompilerInterface.ICompilerTests.TestSimpleRet)
     Expected: 195
  But was:  85
```

Where `85` is `0x55` which is `push rbp`.

Unfortunately if I uncomment
```csharp
			// InstalledRuntimeCode irc = runtimeInfo.InstallCompilationResult (result, nativeCode);

			// runtimeInfo.ExecuteInstalledMethod (irc);
```

Instead of running, it crashes in native while calling `mono_install_compilation_result`


lambdageek:
Ugh... wrong repo. sorry.

PR #6695, opened 2018-01-27
A few steps toward C++: Fix some code, build system, extern "C" wrappings

jaykrell:

PR #9722, opened 2018-07-25
[Cplusplus] work in progress do not merge

jaykrell:
This is a merge/rebase from the  initial port back in January or February.


jaykrell:
Needs https://github.com/dotnet/corefx/pull/31342
or https://github.com/mono/corefx/pull/106

jaykrell:
Continuation of https://github.com/mono/mono/pull/6695.

filipnavara:
I do see the value in making Mono embeddable from C++ code (ie. `extern "C" `), but quite frankly I do not see the value of making some code compilable by C++ just for the sake of it (such as the external System.Native code).

jaykrell:
Unfortunately this was a large rebase from an earlier unmerged PR, which at this point is very merge-conflict-laden.

1. As such, besides being large and tedious, it was also largely partially automated, which makes me nervous.

2. @lambdageek also suggested following a kind of challenge/response/proof pattern where we run CI with configure -enable-cplusplus and use the errors there to motivate the changes. That CI is setup, albeit Linux/amd64 only. There are Centos6-specific changes, and surely arm-specific, Windows-specific, Mac-specific, etc. Those can be covered with later CI or mop-up or piggy backing on top of CI-demonstrated patterns. The bulk of the changes are in portable code, at least.

Either way, the reluctant point, is that I will probably reapply all/many these changes from scratch.

In particular, the splitting of declaration from initialization is too hard to quickly read I suspect and must be carefully mechanically manually applied using copy/paste, not merge/rebase.

This PR should stand a while however,  successfully building (soon) across the matrix.

The voluminous extern "C" part is not so worrisome, easy to review.

ALTERNATIVELY We could merge this after careful tedious expensive review.
Also probably moving changes around among commits for easier reading.

Also mini isn't finished being ported. sgen and metadata have. Not sure about utils. So that is needed.

jaykrell:
moving on to https://github.com/mono/mono/pull/9893 due to all the merging

PR #9724, opened 2018-07-25
[runtime] Refactor cattr decoding.

vargaz:
Return objects from load_cattr_value () using an additional out parameter. Add a variant of
mono_reflection_create_custom_attr_data_args () which can only decode a subset of cattr types, but
it doesn't allocate any objects. Use this in marshal.c/marshal-ilgen.c to avoid allocating objects
while decoding FixedBufferAttribute and UnmanagedFunctionPointerAttribute.

### DIFF
--- a/mono/metadata/custom-attrs.c
+++ b/mono/metadata/custom-attrs.c
@@ -220,11 +220,20 @@ load_cattr_enum_type (MonoImage *image, const char *p, const char *boundp, const
 	return mono_class_from_mono_type (t);
 }
 
-static MonoReflectionType*
-load_cattr_type_object (MonoImage *image, MonoType *t, const char *p, const char *boundp, const char **end, MonoError *error, guint32 *slen)
+static MonoType*
+load_cattr_type (MonoImage *image, MonoType *t, gboolean header, const char *p, const char *boundp, const char **end, MonoError *error, guint32 *slen)
 {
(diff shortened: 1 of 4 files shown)


vargaz:

mono/metadata/custom-attrs.c
```diff
@@ -1107,6 +1148,155 @@ mono_reflection_create_custom_attr_data_args (MonoImage *image, MonoMethod *meth
...
 
+/*
+ * mono_reflection_create_custom_attr_data_args_noalloc:
+ *
+ * Same as mono_reflection_create_custom_attr_data_args_noalloc but allocate no managed objects, return values
+ * using C arrays. Only usable for cattrs with primitive/type arguments.
+ * TYPED_ARGS, NAMED_ARGS, and NAMED_ARG_INFO should be freed using g_free ().
+ */
+void
+mono_reflection_create_custom_attr_data_args_noalloc (MonoImage *image, MonoMethod *method, const guchar *data, guint32 len,
```
jaykrell: This can either this call `mono_reflection_create_custom_attr_data_args` or can `mono_reflection_create_custom_attr_data_args` call this, or can they  be one function with more parameters?

vargaz:
It could be, but it would look ugly, there are too many differences.

lewurm:
@monojenkins build failed

PR #9733, opened 2018-07-25
Remove assert when retrieving value of string with invalid chars

brianradunity:
This assert was causing Unity to crash when trying to view a local
string variable in the debugger with invalid characters.


lewurm:
@monojenkins build failed

PR #9734, opened 2018-07-25
"Code Egyptian Blue".

baulig:
Move all the recent CoreFX additions to `System.Security.dll` into `System.dll` and add type-forwarders.


filipnavara:
I don't think this is necessarily a good idea. The S.S.C.Pkcs classes are quite standalone and separated from the rest of the code. Moving the AsnReader/AsnWriter to System would make sense though since it can be shared and it will eventually replace DerEncoder/DerSequenceReader used in the CoreFX X509 code.

The new PKCS12 API would also make sense in System if you plan to use it as a fallback parser on macOS for the problematic use cases (pfx with no password).

baulig:
I plan on completely replacing the old managed parser with it - though pfx with no password will made to hard fail on all platforms (I'm going to suggest that to the corefx guys).

filipnavara:
What is the point in that? Is there something that is not covered by BTLS, OpenSSL, Apple Crypto or Windows API (which are already part of Mono or CoreFX)? My main concern is that I want X509Chain and X509Store working and I don't see how managed parser fits into the big picture if you would still have to convert it to some native implementation in the end. I see how some managed parsing could be used to cover some edge cases, but that currently feels way down the road.

filipnavara:
> though pfx with no password will made to hard fail on all platforms (I'm going to suggest that to the corefx guys).

Not a good idea IMHO. First, CoreFX on Windows should not regress compared to NetFX. Second, I believe that Mono/CoreFX should support at least what the native tools on the underlying platform do. Bonus points if it can actually support more for cross-platform compatibility. 

/Putting my application developer hat on/

In our application we expose access to both the system certificate store and our own store. We also let users import certificates into our store. I would have hard time explaining our users that there are certificates that can be imported into the system just fine but fail when they try to import them through the application.

PR #9738, opened 2018-07-26
[Coop] A bunch of nohandles().

jaykrell:


mono/metadata/icall-def.h
```diff
@@ -970,34 +970,34 @@ HANDLES(ICALL(THREAD_23, "SuspendInternal", ves_icall_System_Threading_Thread_Su
...
+NOHANDLES(ICALL(THREAD_43, "VolatileWrite(intptr&,intptr)", ves_icall_System_Threading_Thread_VolatileWriteIntPtr))
+NOHANDLES(ICALL(THREAD_44, "VolatileWrite(long&,long)", ves_icall_System_Threading_Thread_VolatileWrite8))
+NOHANDLES(ICALL(THREAD_45, "VolatileWrite(object&,object)", ves_icall_System_Threading_Thread_VolatileWriteObject))
+NOHANDLES(ICALL(THREAD_46, "VolatileWrite(sbyte&,sbyte)", ves_icall_System_Threading_Thread_VolatileWrite1))
+NOHANDLES(ICALL(THREAD_47, "VolatileWrite(single&,single)", ves_icall_System_Threading_Thread_VolatileWriteFloat))
+NOHANDLES(ICALL(THREAD_48, "VolatileWrite(uint&,uint)", ves_icall_System_Threading_Thread_VolatileWrite4))
+NOHANDLES(ICALL(THREAD_49, "VolatileWrite(uint16&,uint16)", ves_icall_System_Threading_Thread_VolatileWrite2))
+NOHANDLES(ICALL(THREAD_50, "VolatileWrite(uintptr&,uintptr)", ves_icall_System_Threading_Thread_VolatileWriteIntPtr))
+NOHANDLES(ICALL(THREAD_51, "VolatileWrite(ulong&,ulong)", ves_icall_System_Threading_Thread_VolatileWrite8))
+NOHANDLES(ICALL(THREAD_9, "YieldInternal", mono_threads_platform_yield))
```
lambdageek: I don't understand this change.  Why not keep `ves_icall_System_Threading_Thread_Yield`

jaykrell: Zoltan @vargaz had pointed out in another PR that NOHANDLES often suffices and is more efficient. So I have started rolling back HANDLES. There are a bunch where it wasn't really needed.  And can be added as needed -- Centrinel being the higher authority, not just sprinking HANDLES everywhere.

lambdageek: @jaykrell sorry, I didn't mean "why did we add `NOHANDLES()` here" - that's fine.  what I was wondering about is why we're exposing a low-level function like `mono_threads_platform_*` directly as an icall instead of going through a `ves_icall_*`.

jaykrell: I worry about ELF interposability and lack of LTO/LTCG so skip ahead to the lowest level function possible/easy -- don't want even ELF overhead or tailcalls. Micro concerns all around granted.

lambdageek: Serious question because i'm curious: is there a benefit from yielding a bit faster?

jaykrell: Just arbitrary target of mechanical transform.

jaykrell:

mono/metadata/w32file.c
```diff
@@ -689,19 +689,19 @@ ves_icall_System_IO_MonoIO_SetFileTime (HANDLE handle, gint64 creation_time,
 }
 
 HANDLE 
-ves_icall_System_IO_MonoIO_get_ConsoleOutput ()
+ves_icall_System_IO_MonoIO_get_ConsoleOutput (void)
```
jaykrell: These functions could be eliminated too.

jaykrell:

PR #9742, opened 2018-07-26
[Coop] Convert ves_icall_MonoField_GetFieldOffset.

jaykrell:


jaykrell:
Review? This might need cross work, but since it is an icall, probably not (runtime, not compiler).

PR #9743, opened 2018-07-26
[Coop] Convert ves_icall_System_Reflection_Module_ResolveStringToken.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/icall.c
```diff
@@ -6004,36 +6004,34 @@ ves_icall_System_Reflection_Module_ResolveMethodToken (MonoImage *image, guint32
...
 		*resolve_error = ResolveTokenError_BadTable;
-		return NULL;
+		return NULL_HANDLE_STRING;
 	}
 
 	if (image_is_dynamic (image)) {
-		MonoString * result = (MonoString *)mono_lookup_dynamic_token_class (image, token, FALSE, NULL, NULL, error);
-		mono_error_cleanup (error);
+		ERROR_DECL (ignore_inner_error); // FIXME?
+		MonoStringHandle result = MONO_HANDLE_NEW (MonoString, mono_lookup_dynamic_token_class (image, token, FALSE, NULL, NULL, ignore_inner_error));
```
lambdageek: This could use a FIXME note - `mono_lookup_dynamic_token_class` ultimately ends up in `mono_reflection_lookup_dynamic_token` which returns managed objects.

jaykrell:
oops

jaykrell:
I'm not sure this was "officially" ready in the sense I usually expect, but should be ok.
It was approved. I didn't see if CI was finished or ask sheriffs to override. It is a small safe change.

alexischr:
@jaykrell It had a tiny breaking typo that I fixed on master. While fixing it I noticed that https://github.com/mono/mono/pull/9743/files#diff-55205fbc5be02187cdfe6d77547ab70fR7338 looks like it should be `*domain` being used

jaykrell:
@alexischr You are right. It won't matter for the initial caller, which is passing the same thing, but that was certainly the intent, good catch.

PR #9749, opened 2018-07-26
Use System.Security.Policy.Publisher from corefx (they essentially stubbed it out).

baulig:
Fixes the XUnit 2.4 problem.


alexischr:
@monojenkins commit apidiff

alexischr:
@monojenkins commit csproj

mcs/class/corlib/System.Security.Policy/Publisher.cs
```diff
@@ -38,7 +38,8 @@ namespace System.Security.Policy {
 	public sealed class Publisher :
 		EvidenceBase,
 		IIdentityPermissionFactory, IBuiltInEvidence {
-	
+
+		[NonSerialized]
```
marek-safar: Why did you change file which is no longer used?

baulig: Left-over from testing, before replacing the file.

marek-safar: Could you delete the no longer used files

PR #8654, opened 2018-05-10
Enable mono-state-based telemetry for crashing on CI

alexanderkyte:
Everything seemed to be working before on CI before. Let's add the jenkins hooks for the crash files with this PR too.

PR #9755, opened 2018-07-26
[runtime] Disable telemetry on the sdks

alexanderkyte:

### DIFF
--- a/sdks/builds/runtime.mk
+++ b/sdks/builds/runtime.mk
@@ -48,9 +48,13 @@ __$(1)_CONFIGURE_ENVIRONMENT = \
 	$(if $$(_$(1)_LDFLAGS),LDFLAGS="$$(_$(1)_LDFLAGS)") \
 	$$(_$(1)_CONFIGURE_ENVIRONMENT)
 
+__$(1)_CONFIGURE_FLAGS = \
+	--with-structured_crashes=no \
+	$$(_$(1)_CONFIGURE_FLAGS)
+
 .stamp-$(1)-$$(CONFIGURATION)-configure: $$(TOP)/configure .stamp-$(1)-toolchain
 	mkdir -p $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)
-	$(if $$(_$(1)_PATH),PATH="$$$$PATH:$$(_$(1)_PATH)") ./wrap-configure.sh $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION) $(abspath $(TOP)/configure) $$(_$(1)_AC_VARS) $$(__$(1)_CONFIGURE_ENVIRONMENT) $$(_$(1)_CONFIGURE_FLAGS)
+	$(if $$(_$(1)_PATH),PATH="$$$$PATH:$$(_$(1)_PATH)") ./wrap-configure.sh $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION) $(abspath $(TOP)/configure) $$(_$(1)_AC_VARS) $$(__$(1)_CONFIGURE_ENVIRONMENT) $$(__$(1)_CONFIGURE_FLAGS)
 	touch $$@
 
 .PHONY: .stamp-$(1)-configure


marek-safar:
Please backport down to 2018-04

alexanderkyte:
@monojenkins backport 2018-06

alexanderkyte:
@monojenkins backport 2018-04

sdks/builds/runtime.mk
```diff
@@ -48,6 +48,10 @@ __$(1)_CONFIGURE_ENVIRONMENT = \
 	$(if $$(_$(1)_LDFLAGS),LDFLAGS="$$(_$(1)_LDFLAGS)") \
 	$$(_$(1)_CONFIGURE_ENVIRONMENT)
 
+__$(1)_CONFIGURE_FLAGS = \
```
luhenry: Where is this variable used?

jaykrell: $ git grep _CONFIGURE_FLAGS

```
sdks/builds/runtime.mk: $(if $$(_$(1)_PATH),PATH="$$$$PATH:$$(_$(1)_PATH)") ./wrap-configure.sh $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION) $(abspath $(TOP)/configure) $$(_$(1)_AC_VARS) $$(__$(1)_CONFIGURE_ENVIRONMENT) $$(_$(1)_CONFIGURE_FLAGS)
```

?

sdks/builds/runtime.mk
```diff
@@ -48,6 +48,10 @@ __$(1)_CONFIGURE_ENVIRONMENT = \
...
+	--with-structured_crashes=no \
+	$$(_$(1)_CONFIGURE_FLAGS)
+
 .stamp-$(1)-$$(CONFIGURATION)-configure: $$(TOP)/configure .stamp-$(1)-toolchain
 	mkdir -p $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)
 	$(if $$(_$(1)_PATH),PATH="$$$$PATH:$$(_$(1)_PATH)") ./wrap-configure.sh $$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION) $(abspath $(TOP)/configure) $$(_$(1)_AC_VARS) $$(__$(1)_CONFIGURE_ENVIRONMENT) $$(_$(1)_CONFIGURE_FLAGS)
```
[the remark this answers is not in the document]
alexanderkyte: Thanks!

alexanderkyte:
@lewurm what would the advantages of that be? I don't have many opinions either way. Is it easy to identify that we're building for android from within `mono/utils`?

lewurm:
I don't have strong opinion on it either. putting that stuff in SDKs is probably fine. at some point we should probably figure out how to write such crash reports on android/ios/wasm too, since they can be useful.

@alexanderkyte I'm a bit confused about this whole PR though: isn't crash dumping happening on macOS only anyway? https://github.com/mono/mono/blob/801e3f360fdc3d1358ebf41eaa309eeb5ec6f577/mono/mini/mini-posix.c#L1035-L1062

alexanderkyte:
That ``` #elif defined(TARGET_OSX)``` seems a bit heavy-handed yeah. Though it appears that TARGET_OSX is set to true when android is being built, because android was running into this being enabled on CI.

alexanderkyte:
https://github.com/mono/mono/pull/8654 should fix that for the enabling case, but this PR should be enough for the disabling case.

lewurm:
> Though it appears that TARGET_OSX is set to true when android is being built, because android was running into this being enabled on CI.

@alexanderkyte that doesn't make sense to me. are you sure? do you have some jenkins log?

marek-safar:
@monojenkins backport 2018-06

marek-safar:
@monojenkins backport 2018-04

marek-safar:
I think putting this to SDKs is the right setup. We don't want SDKs products to use our MERP (not event for MS based apps)

lewurm:
@marek-safar this PR doesn't do anything, see https://xamarinhq.slack.com/archives/C03CELHQQ/p1532978454000046 and following messages.

It might be relevant once https://github.com/mono/mono/pull/8654 is merged. We definitely don't need to backport this PR.

marek-safar:
How can we then completely disable MERP for all SDKs ?

alexanderkyte:
I'm closing this, because this concern was addressed in this PR: https://github.com/mono/mono/pull/8654

PR #9757, opened 2018-07-26
[corlib] Disable comparer xunit tests because we are still running re…

marek-safar:
…ferencesource version on mobile


marek-safar:
@monojenkins backport 2018-06

PR #9758, opened 2018-07-26
[sdks][CI] Commit canonical autoconf cache files

alexischr:
This allows for faster first-builds (such as in pull requests) and prevents host configuration mismatch problems. We use explicit minimum versions so these aren't very variable anyway


jaykrell:
Autoconf is often overkill agreed.

lewurm:
@rolfbjarne @spouliot could you take a look please?

rolfbjarne:
@lewurm if these config files are to be used when building xamarin-macios' mono runtimes, they won't work, because we use a versioned Xcode path (`/Applications/Xcode94.app` for instance). These files all use the generic `/Applications/Xcode.app`.

alexischr:
@rolfbjarne I updated the paths to point to Xcode94.app, and RE linked the path on the Mono SDK bots. Joining XCode version and path usage with macios would also be a good thing :)

marek-safar:
I am not sure the pain from out of date thing is worth the time-saving. How much do we save 10 secs on bots?

alexischr:
Yeah @rolfbjarne 's example shows that this might not work at all. It would have saved significant time, as there are 4 configures and each one seems to take over 2 mins (they are timed in the log). The cache use is broken on CI so it runs without it, but locally I think it goes down to 20 secs, so a 6-7 min save per build

vargaz:
I don't think its a good idea to cache these files, without checking that the inputs used to generate them are actually the same.

luhenry:
We do not want to do that without a guarantee that they will be updated automatically whenever they need to be. On top of that, configuration time is not that bad compared to building, and can (should) be done concurrently across all targets.

Issue #9649, opened 2018-07-19
[RFC] Tail recursion should work better in C#, not just F#.


jaykrell:
## Steps to Reproduce
Use the functional programming idiom of mutually recursive functions using tail calls.

Do it in C#.

Provide very large input.

### Current Behavior
Stack overflow.

### Expected Behavior
No stack overflow.

Workarounds:
 - Use F# instead. It will generate tail. prefixes and
   Mono will handle many cases. C# compilers don't generate tail. prefix.

 - Use loops inside of a single function.
   This is less idiomatic perhaps.

 - Or a function calling itself, and non-default optimization.
   Mono already handles some cases, but not many, and not by default.

 - Use CoreCLR instead of Mono. They do these classic transforms already.

 - Use C++ instead of C#. These are classic transforms widely implemented.

Something like:
  https://github.com/mono/mono/pull/9620 but it is not done.

Risks/caveats:
  arm32 is problematic even with tail., and lack of tail. is no better
  Specifically with generics and interface calls.
  So there is somewhat of a processor-portability problem.

  Local addresses could escape and be dependend on, broken by this.
   Overly conservative heuristic is in place already in the PR to mitigate that completely.
  Similar to CoreCLR but probably less sophisticated.

  Stackwalk sensitivity, esp. in test code. This can be seen in the PR already.
  There should perhaps be a way to indicate a function cannot be tailcalled from.
  I have something. The PR is too hacky.

  Debuggability -- frames are lost in the stackwalk. Plenty remain.
  Not clear if this is significant.
  Can alter based on --debug but not clear that is good, because it can be a works/doesn't work situation, and debugging therefore break stuff.

jaykrell:
There is performance to be gained here also. Less stack provides for better locality. Jmp is faster than call/ret. The way we form parameters/jmp can be more instructions. There is code size hit due to repeated epilogues. I can try a microbenchmark.

lambdageek:
Two options (not necessarily good ones...):
1. Do TCO aggressively.  But allow a custom attribute on caller methods that says to retain the frame when debugging:
   ```csharp
      [DebugRetainStackFrame]
      int Caller () {
          if (b)
              return Callee (); /* no TCO if debugging */
          else
              return 1 + Foo (); /* no TCO ever - not a tailcall */
      }
      int Callee () { ... }
   ```
2. Onlly do TCO when explicitly requested using a custom attribute: 
   ```csharp
   [RequestTailCall]
   int Caller () {
       if (b)
[...]

PR #9759, opened 2018-07-27
Reduce BCL/test sensitivity to tailcall optimizations,

jaykrell:
only where it is fairly obvious what the problem is.

There are other problems but they aren't understood.

In particular code likes to walk stack to precisely verify
it is "as expected" or to report data based on its caller.

One unusual API fails to take a parameter but instead the parameter
is the calling class, called from .cctor.

### DIFF
--- a/mcs/class/corlib/Test/System.Reflection.Emit/DynamicMethodTest.cs
+++ b/mcs/class/corlib/Test/System.Reflection.Emit/DynamicMethodTest.cs
@@ -475,6 +475,9 @@ namespace MonoTests.System.Reflection.Emit
 	        DynamicMethod dm = new DynamicMethod("Foo", typeof(object), null);
 	        ILGenerator ilgen = dm.GetILGenerator();
 	        ilgen.Emit(OpCodes.Call, typeof(MethodBase).GetMethod("GetCurrentMethod"));
+		// Call twice and pop to avoid tailcall optimization.
+	        ilgen.Emit(OpCodes.Call, typeof(MethodBase).GetMethod("GetCurrentMethod"));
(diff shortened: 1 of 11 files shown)


jaykrell:
https://github.com/mono/mono/issues/9649

mcs/class/corlib/Test/System.Reflection.Emit/DynamicMethodTest.cs
```diff
@@ -475,6 +475,9 @@ public void GetCurrentMethodWorksWithDynamicMethods ()
 	        DynamicMethod dm = new DynamicMethod("Foo", typeof(object), null);
 	        ILGenerator ilgen = dm.GetILGenerator();
 	        ilgen.Emit(OpCodes.Call, typeof(MethodBase).GetMethod("GetCurrentMethod"));
+		// Call twice and pop to avoid tailcall optimization.
+	        ilgen.Emit(OpCodes.Call, typeof(MethodBase).GetMethod("GetCurrentMethod"));
```
jaykrell: If it works, this might be better of call to noinline NoTailcall like the others.
As well, it might be reasonable for JIT to special case GetCurrentMethod -- it already does sometimes.

mono/tests/exception18.cs
```diff
@@ -38,6 +38,19 @@ public static void Main ()
 				throw new Exception (String.Format("Exception carried {0} frames along with it when it should have reported one.", frames));
 		}
 
+		/*
```
jaykrell: I don't think test code should rely on such BCL details, tailcall or otherwise.

PR #9761, opened 2018-07-27
Tighten up some types from void* to the correct type.

jaykrell:


jaykrell:
@monojenkins build AArch64

jaykrell:
@monojenkins build ARMv5

jaykrell:
@monojenkins build ARMv7

mono/utils/mono-threads.h
```diff
@@ -153,6 +153,8 @@ enum {
 	ASYNC_SUSPEND_STATE_INDEX = 1,
 };
 
+struct HandleStack;
+struct MonoJitTlsData;
```
lambdageek: should we follow the underscore convention for these struct tags?

jaykrell: Maybe, maybe not.
Eventually you just have C++ "struct foo" and the underscores kinda get in the way -- they end up duplicated.
For C it is a good convention, it implies private -- users should say `Foo` not `struct _Foo`.

jaykrell:
Unrelated CI failures.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

mono/utils/mono-threads.h
jaykrell: bikeshedding++:


I suggest the ugliness of the eventual hypothetical C++ outweighs the C convention, even if the C++ does not come to pass. But it is arguable either way. You could remove the underscores later. You could leave them in the C++. The names are also reserved for the C/C++ implementer but *everyone* ignores that part and gets away with it.
The underscores are part of the C++ ABI, the name mangling, but that doesn't clearly  matter either. You can strive for an externC ABI, or you can break the C++ ABI. The intersection of C++ ABI and compatibility is approximately zero. glibc has broken those sorts of things.

I believe Mono follows both conventions, which isn't to say anyone agreed to it.

Anyway, independent of the leading underscore, I believe this is a good coding style. It has cons but mostly pros:

pros:
  can see things w/o cast in debugger
  can pass around  the pointers among clients w/o cast; can stop writing the little wrapper macros and functions to do the  casting (!) (which arguably takes a popular breakpoint target away)
  The struct is still usually totally opaque to client, as it is not declared where they see it, typically. So just as good as a gpointer.

cons:
  Revealing something is a struct is a certain limited transparency. You are revealing it isn't a  union or an integer. But yet, really, it could be. The struct is just a  "token" type, and internally  you can cast. Albeit, you are limited then to the size. If you wanted 64bits on 32bit platform, the struct* isn't good  -- in that case, pass a struct by value that has a longlong, or however many  opaque bits you need. gpointer has same problem.

   There is still no provision for a hiearchy of  types.  Imagine we had object, string, and animal. They'd each be separate uncovertable types. If they were all opaque integer or void*, they'd be freely convertable, perhaps too freely.

 That it is a struct foo* is a cat somewhat out of the bag. You then have to be careful to guard the struct declaration and prevent its accidental misuse. A gpointer is harder to accidentally abuse as it requires a cast. The  struct might be revealed by accident as it is often just an include away. Though in a really nicely formed system, the struct is only revealed in a .c file, not any foo-internals.h. That is good. Imagine threads-types.h only had extra functions, and the types were only  in threads.c

jaykrell:
Really I just want to get these things in. There is infinite quibbling possible, but this is a good and well known style.

jaykrell:
Review?

mono/utils/mono-tls.h
```diff
@@ -85,11 +85,11 @@ gint32 mono_tls_get_tls_offset (MonoTlsKey key);
 gpointer mono_tls_get_tls_getter (MonoTlsKey key, gboolean name);
 gpointer mono_tls_get_tls_setter (MonoTlsKey key, gboolean name);
 
-gpointer mono_tls_get_thread (void);
-gpointer mono_tls_get_jit_tls (void);
-gpointer mono_tls_get_domain (void);
-gpointer mono_tls_get_sgen_thread_info (void);
-gpointer mono_tls_get_lmf_addr (void);
+struct _MonoInternalThread *mono_tls_get_thread (void);
```
kumpera: This is exposing structs from other subsystems here.
The casts should be on the use site instead.

mono/utils/mono-threads.h
```diff
@@ -153,6 +153,8 @@ enum {
 	ASYNC_SUSPEND_STATE_INDEX = 1,
 };
 
+struct HandleStack;
```
kumpera: Those are types in metadata and sgen, they should not be in utils at all.

jaykrell: Forward declaring a struct doesn’t really expose anything. It is still opaque. We expose the presence of the pointers in the thread already.

mono/utils/mono-tls.h
jaykrell: It doesn’t expose what they have, just that they exist & their name. They remain opaque. You can now see them more easily in debugger.

mono/utils/mono-threads.h
kumpera: This is not the style we have in mono and this is violating basic layering.

jaykrell: The layering is preserved as the types are opaque.
The layering is already similarly violated/preserved by putting the pointers in utils, albeit void*. Right? Why is this worse than that?
It helps debugging and cleans up the C code. Seems like a win.

jaykrell:
On the underscores there are probably countless cases of each.
For example, no underscore: https://github.com/mono/mono/pull/9884/files#diff-16a2ebe4e225a1a41bc787173d1b6933L15

luhenry:
As pointed out in https://github.com/mono/mono/pull/9761#pullrequestreview-142375477, exposing the types is not acceptable. If you want this to get merged, you'll have to fix it.

jaykrell:
This does not expose any types. Their existence only. There are many advantages to this approach. Type-safety, debugablity, readability, writ ability — can see the type in the declaration, users who have the revealed type can use it without cast, debugger can see it without cast. I can bring it up in slack instead.

PR #9769, opened 2018-07-27
Changed http: to https:

RemiBou:
I changed all the link for going to the https version as there is no redirect in some of them (.net foundation and monoproject)


migueldeicaza:
Some of these link changes should not be made until we actually fix mono-project.com certificates

RemiBou:
Only the mono-project.com is bugged but www.mono-project.com is fine

PR #9775, opened 2018-07-27
Add check for a generic method when invoking in debugger

brianradunity:
VSTU will try to invoke a generic (uninflated) method in certain
situations, which was causing the Mono debugger to assert and crash
Unity.  Added a check for this situation before the invoke so the
debugger can return an error to the client. Previously, this case
wasn't caught until the method was in JIT compilation, which will
assert when this situation is detected.


mono/mini/debugger-agent.c
```diff
@@ -5965,7 +5965,11 @@ do_invoke_method (DebuggerTlsData *tls, Buffer *buf, InvokeData *invoke, guint8
 		this_buf = (guint8 *)g_alloca (mono_class_instance_size (m->klass));
 	else
 		this_buf = (guint8 *)g_alloca (sizeof (MonoObject*));
-	if (m_class_is_valuetype (m->klass) && (m->flags & METHOD_ATTRIBUTE_STATIC)) {
+
+	if (m->is_generic) {
+        DEBUG_PRINTF (1, "[%p] Error: Attemtping to invoke uninflated generic method %s.\n", (gpointer)(gsize)mono_native_thread_id_get (), mono_method_full_name (m, TRUE));
```
vargaz: The indentation seems to be off.

vargaz:
Looks ok otherwise.

Issue #9776, opened 2018-07-27
Flaky IO.Pipes.Tests.AnonymousPipeTest_*


jaykrell:
See these on a few PRs this morning.

## Steps to Reproduce

1. Submit a PR.

### Current Behavior

Fails.
For example:

https://github.com/mono/mono/pull/9768

https://jenkins.mono-project.com/job/test-mono-pull-request-i386/16378/#showFailuresLink

```
	Test Result (3 failures / ±0)
.IO.Pipes.Tests.AnonymousPipeTest_Specific.ClonedServer_ActsAsOriginalServer
.IO.Pipes.Tests.AnonymousPipeTest_Write_ServerIn_ClientOut.WriteToPipeWithClosedPartner_Throws_IOException
.IO.Pipes.Tests.AnonymousPipeTest_Write_ServerOut_ClientIn.WriteToPipeWithClosedPartner_Throws_IOException
```

### Expected Behavior

Success.

filipnavara:
Uh, the failure are actually because some API is missing in the PAL:

```
      ---- System.EntryPointNotFoundException : SystemNative_ConvertErrorPlatformToPal
      Stack Trace:
          at (wrapper managed-to-native) Interop+Sys.ConvertErrorPlatformToPal(int)
```

I will check it tonight unless someone beats me to it.

filipnavara:
I think it's broken since https://github.com/mono/mono/commit/1e9be63015ee88ec2e7ca6d92c22904902989440#diff-0ff844e784e15d2542b71c85d2c0133e. There's `if !TARGET_WIN32` in the Makefile that prevents inclusion of many PAL files. /cc @alexischr

alexischr:
@filipnavara Sorry about that, I'm not sure how I missed that. I'll fix it up

jaykrell:
Eek, how did that get merged?

filipnavara:
@alexischr Sorry. I may have misread the diff. Nevertheless it looks like some configuration doesn't include correct PAL files.

filipnavara:
@alexischr BTW, I think the problem is that ENABLE_MONODROID (and same for MONOTOUCH) means "enable the profiles along with the other ones", not "building for Android/iOS exclusively".

PR #9780, opened 2018-07-27
Revert "[System.Native] Build a reduced library on Android and iOS"

jaykrell:
This reverts commit 1e9be63015ee88ec2e7ca6d92c22904902989440.

It doesn't appear to been run through CI and does appear to be breaking things.
https://github.com/mono/mono/issues/9776


filipnavara:
Thanks much for the effort and help! Let's see if this one goes through with the CI.

jaykrell:
Good point. If this works, then abandon the other.

PR #9782, opened 2018-07-27
[cxx] Add extern "C" { } to many headers.

jaykrell:
Sometimes G_BEGIN/END_DECLS, sometimes MONO_BEGIN/END_DECLS, if they work,
else just plain extern "C".

- Predictable portable names for dlopen/dlsym (LoadLibrary/GetProcAddress).
- Letting C and C++ interoperate if/while the tree
  undergoes a transition from C to C++, including
  if some code stays as C indefinitely.
- Some might later be reverted, esp. outside of the dlopen/dlsym (LoadLibrary/GetProcAddress)
  interface or embedding interface.

Also add to some .c files, in particular for functions that do not have prototypes in .h files.

Adding it to all .c files is also an idea depending on policy,
i.e. to inhibit overloading.

### DIFF
--- a/libgc/include/gc.h
+++ b/libgc/include/gc.h
@@ -42,11 +42,6 @@
 #   define GC_CONST
 #  endif
 
-# ifdef __cplusplus
-    extern "C" {
-# endif
-
-
 /* Define word and signed_word to be unsigned and signed types of the 	*/
 /* size as char * or void *.  There seems to be no way to do this	*/
 /* even semi-portably.  The following is probably no better/worse 	*/
@@ -64,6 +59,10 @@
   #include <stdint.h>
   typedef unsigned __int64 GC_word;
   typedef __int64 GC_signed_word;
+#endif
+
+#ifdef __cplusplus
+extern "C" {
 #endif
 
 /* Public read-only variables */
@@ -501,7 +500,17 @@ GC_API GC_PTR GC_malloc_atomic_ignore_off_page GC_PROTO((size_t lb));
 #endif
 
 #if defined(__linux__) || defined(__GLIBC__)
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
 # include <features.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 # if (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 1 || __GLIBC__ > 2) \
      && !defined(__ia64__)
 #   ifndef GC_HAVE_BUILTIN_BACKTRACE
@@ -965,7 +974,17 @@ GC_API void (*GC_is_visible_print_proc)
 
 #if !defined(GC_USE_LD_WRAP) && \
     (defined(GC_PTHREADS) || defined(GC_SOLARIS_THREADS) || defined(GC_DARWIN_THREADS) || defined(GC_MACOSX_THREADS))
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
 # include "gc_pthread_redirects.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #endif
 
 # if defined(PCR) || defined(GC_SOLARIS_THREADS) || \
@@ -985,7 +1004,16 @@ extern void GC_thr_init(void);	/* Needed for Solaris/X86	*/
 #endif /* THREADS && !SRC_M3 */
 
 #if defined(GC_WIN32_THREADS) && !defined(__CYGWIN32__) && !defined(__CYGWIN__)
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
 # include <windows.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 # ifdef GC_INSIDE_DLL
    BOOL WINAPI GC_DllMain(HINSTANCE inst, ULONG reason, LPVOID reserved);
@@ -1081,6 +1109,10 @@ extern void GC_thr_init(void);	/* Needed for Solaris/X86	*/
     GC_API void GC_win32_free_heap ();
 #endif
 
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
 #if ( defined(_AMIGA) && !defined(GC_AMIGA_MAKINGLIB) )
   /* Allocation really goes through GC_amiga_allocwrapper_do */
 # include "gc_amiga_redirects.h"
@@ -1088,10 +1120,6 @@ extern void GC_thr_init(void);	/* Needed for Solaris/X86	*/
 
 #if defined(GC_REDIRECT_TO_LOCAL) && !defined(GC_LOCAL_ALLOC_H)
 #  include  "gc_local_alloc.h"
-#endif
-
-#ifdef __cplusplus
-    }  /* end of extern "C" */
 #endif
 
 #endif /* _GC_H */
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -54,6 +54,15 @@
 #define G_END_DECLS
 #endif
 
+// These macros are duplicated in mono-compiler.h, mono-publib.h, glib.h for maximum reach and consistency.
+#ifdef  __cplusplus
+#define MONO_BEGIN_DECLS  extern "C" {
+#define MONO_END_DECLS    }
+#else
+#define MONO_BEGIN_DECLS /* nothing */
+#define MONO_END_DECLS   /* nothing */
+#endif
+
 G_BEGIN_DECLS
 
 /*
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -46,21 +46,14 @@
 
 #define __EGLIB_X11 1
 
+// FIXMEcxx in future G_BEGIN_DECLS should always be empty, and/or remove all uses,
+// and extern "C" added to MONO_API instead.
 #ifdef  __cplusplus
 #define G_BEGIN_DECLS  extern "C" {
 #define G_END_DECLS    }
 #else
-#define G_BEGIN_DECLS
-#define G_END_DECLS
-#endif
-
-// These macros are duplicated in mono-compiler.h, mono-publib.h, glib.h for maximum reach and consistency.
-#ifdef  __cplusplus
-#define MONO_BEGIN_DECLS  extern "C" {
-#define MONO_END_DECLS    }
-#else
-#define MONO_BEGIN_DECLS /* nothing */
-#define MONO_END_DECLS   /* nothing */
+#define G_BEGIN_DECLS  /* nothing */
+#define G_END_DECLS    /* nothing */
 #endif
 
 G_BEGIN_DECLS
--- a/mono/metadata/icall-table.c
+++ b/mono/metadata/icall-table.c
@@ -36,7 +36,7 @@
 #include <mono/utils/mono-publib.h>
 #include <mono/utils/bsearch.h>
 
-MONO_BEGIN_DECLS // lack of prototypes
+G_BEGIN_DECLS // lack of prototypes
 
 /*
  * icall.c defines a lot of icalls as static, to avoid having to add prototypes for
@@ -446,7 +446,7 @@ lookup_icall_symbol (gpointer func)
 #endif
 }
 
-MONO_END_DECLS // lack of prototypes
+G_END_DECLS // lack of prototypes
 
 void
 mono_icall_table_init (void)
--- a/mono/metadata/icall-table.c
+++ b/mono/metadata/icall-table.c
@@ -36,6 +36,8 @@
 #include <mono/utils/mono-publib.h>
 #include <mono/utils/bsearch.h>
 
+G_BEGIN_DECLS // lack of prototypes
+
 /*
  * icall.c defines a lot of icalls as static, to avoid having to add prototypes for
  * them, just don't include any mono headers and emit dummy prototypes.
@@ -443,6 +445,8 @@ lookup_icall_symbol (gpointer func)
 	return NULL;
 #endif
 }
+
+G_END_DECLS // lack of prototypes
 
 void
 mono_icall_table_init (void)
--- a/mono/metadata/icall-table.c
+++ b/mono/metadata/icall-table.c
@@ -36,7 +36,7 @@
 #include <mono/utils/mono-publib.h>
 #include <mono/utils/bsearch.h>
 
-G_BEGIN_DECLS // lack of prototypes
+MONO_BEGIN_DECLS // lack of prototypes
 
 /*
  * icall.c defines a lot of icalls as static, to avoid having to add prototypes for
@@ -446,7 +446,7 @@ lookup_icall_symbol (gpointer func)
 #endif
 }
 
-G_END_DECLS // lack of prototypes
+MONO_END_DECLS // lack of prototypes
 
 void
 mono_icall_table_init (void)
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -6669,7 +6669,9 @@ ves_icall_System_Environment_GetEnvironmentVariable_native (const gchar *utf8_na
  * arm-apple-darwin9.  We'll manually define the symbol on Apple as it does
  * in fact exist on all implementations (so far) 
  */
+G_BEGIN_DECLS
 gchar ***_NSGetEnviron(void);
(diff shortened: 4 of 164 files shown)


jaykrell:
Failures are unrelated, master was broken. Review?

jaykrell:
@monojenkins build  deb with monolite

jaykrell:

jaykrell:
ExternC can also be put into MONO_API to handle dlsym and embedding.

libgc/include/gc.h
```diff
@@ -66,6 +61,10 @@
   typedef __int64 GC_signed_word;
 #endif
 
+#ifdef __cplusplus
```
luhenry: We have no interest in compiling libgc in C++. Revert all the changes in the `libgc/` folder.

mono/sgen/sgen-minor-scan-object.h
```diff
@@ -9,6 +9,10 @@
  * Licensed under the MIT license. See LICENSE file in the project root for full license information.
  */
 
+#ifdef __cplusplus
```
luhenry: Use `G_BEGIN_DECLS` in all of `sgen/`

mono/utils/bsearch.h
```diff
@@ -9,6 +9,10 @@
 
 #include "mono/utils/mono-compiler.h"
 
+#ifdef __cplusplus
```
luhenry: Use `G_BEGIN_DECLS` in all places you do that.

libgc/include/gc.h
jaykrell: The libgc changes are needed for the sake of libgc callers.

mono/sgen/sgen-minor-scan-object.h
jaykrell: I tried and it does not always work. G_ for example isn’t for public headers. I can try again.

mono/utils/bsearch.h
jaykrell: I tried. It is not always available. I can try more.

libgc/include/gc.h
jaykrell: Mainline libgc already has all these changes, but presumably we are never updating.

mono/sgen/sgen-minor-scan-object.h
jaykrell: If we are to be consistent, I think MONO_ is the only choice. I.e. no G_ in public headers. This will likely require adding some defines which I hadn’t done.

jaykrell: extern “C” w/o macro is also a good option for consistency. Very portable no matter includes & defines.

jaykrell: I appears I already use `G_BEGIN_DECLS` in all of sgen. Ok?
Similarly mini.
utils has a mix.
metadata has a mix.

Is consistency within each of the 5-ish directories (utils, metadata, mini, libgc, sgen) enough, or you require consistency across all 5?

libgc changes are needed (the pointer type  changes are negotiable, but they benefit the callers, and match mainline; the extern C is required, unless maybe libgc becomes C++ which was not my  direction).

Public headers exists.

So consistency across all 5 would be no macro or MONO_.

I can try MONO_ more consistently in utils and metadata.

I believe extern "C" a little outside of libgc was also deliberate, as the macros weren't defined in some places, and I didn't change that.

We could endevour for MONO_ to be defined more and used more; we could settle on it everywhere.

jaykrell:
So for example, mono/utils/bsearch.h and mono/utils/bsearch.c do not have G_BEGIN_DECLS or MONO_BEGIN_DECLS available to it without additional change. I can attempt that change.

jaykrell:
sysmath.h is another one.
It includes glib.h but not mono*h before declarations, so G_ there instead of MONO_.
I was making the smaller change.

Anyway, I will duplicate MONO_BEGIN_DECLS in mono-compiler.h, mono-publib.h, glib.h so enable using it more. One could add or reorder includes to achieve this but that seems unnecessarily risky,  unless, perhaps, we have infrastructure to ensure binaries are absolutely unchanged by a change -- which is essentially a goal here -- making code valid C++ should not change the results compiled as C at all.

I will endeavor that any additions I made in mono and utils be MONO_.
It appears mini and sgen are consistently G_, and you proscribed using G_ in places, so that should probably remain We can move to MONO_ or not.

There is arguably a difference between MONO_ and G_.
MONO_ is public. G_ is not.
G_ could stop being extern "C". MONO_ cannot, for C embedders.
Bug really this strikes at the other point of not making this change at all, but putting extern "C" in MONO_API. Not MONO_BEGIN_DECLS but MONO_API.

It depends on a few things. How reliablie is MONO_API? In particular, are there embedders using static lib who have no dependency on such marking, but can call any function?

Really I just want to move on here.
Blanket extern "C" is too useful as a temporary-albeit-large-and-noisy step, maybe a permanent step, to get too hung up on it while we have nothing to show for it all anyway. Let's get the tree all valid C++ (metadata, sgen, mini, utils), compile as C++, test it, try to use some C++, commit to C++, and then after all that hypothetical, come back to the extern "C" matter. But yes I'll see about consistency in the short term.

mono/eglib/glib.h
```diff
@@ -54,6 +54,15 @@
 #define G_END_DECLS
 #endif
 
+// These macros are duplicated in mono-compiler.h, mono-publib.h, glib.h for maximum reach and consistency.
```
luhenry: Revert that, it has nothing to do here.

mono/metadata/property-bag.h
```diff
@@ -12,6 +12,10 @@
 
 #include <mono/utils/mono-compiler.h>
 
+#ifdef __cplusplus
```
luhenry: Use appropriate `G_BEGIN_DECLS` or `MONO_BEGIN_DECLS`.

mono/sgen/sgen-minor-scan-object.h
luhenry: This still needs to be fixed. I am not going to go over all the ones that needs fixing simply because you can very clearly see it in this PR. And no, not all of sgen is using `G_BEGIN_DECLS`, I assumed that would be pretty easy to see and fix.

mono/utils/mono-compiler.h
```diff
@@ -5,6 +5,15 @@
 #ifndef __UTILS_MONO_COMPILER_H__
 #define __UTILS_MONO_COMPILER_H__
 
+// These macros are duplicated in mono-compiler.h, mono-publib.h, glib.h for maximum reach and consistency.
```
luhenry: Why the duplication?

mono/eglib/glib.h
luhenry: Seems pretty straightforward to me: `MONO_BEGIN_DECLS` in public headers, `G_BEGIN_DECLS` everywhere else. We already do that for a few types. I am pretty optimistic that's not too complex.

mono/utils/mono-compiler.h
jaykrell: The comment indicates but maybe inadequately— the headers are not always included and adding or reordering includes is unwarranted risk/change.

mono/eglib/glib.h
jaykrell: It is a rule, not terrible, can establish its own consistency, but when you suggested consistency, to me that meant something even simpler and more consistent — use just one of G_ or MONO_ w/o requiring to partition into public/private headers or be sensitive to such partitioning. I mean, you know, the PR already had a mix, which you objected to. I thought the idea was drive down/out the mix, not have a specific mix. To the casual observer this consistency will not be evident, it will just look like another inconsistency. But the importance of casual observers is unclear, granted.

jaykrell:
Sgen appears consistent -- all G_.
I was wrong about mini, it does (now) have G_ and MONO_, and it looks like I added both.
Anyway, with the rule now clearly stated I can try to follow it.

jaykrell:
sgen: What I missed in searching out the inconsistency, is I was looking for G_ and MONO_ but missed the non-macroed extern "C".

jaykrell:
The alternative solution of putting extern "C in MONO_API, and having G_BEGIN_DECLS do nothing, gets promisingly pretty far. (MONO_BEGIN_DECLS left alone, arguable).

It fails for the following reasons:

MONO_API is missing in a few places like `mono_file_map_error`.
glib is still C.

Many icalls fail to link. Putting extern "C" in ICALL_EXPORT didn't work -- it is maybe underused.
Underused, for example `ves_icall_System_Threading_Monitor_Monitor_pulse_all.

Lots of warnings about the resulting `extern extern "C"` -- we should just remove all those meaningless externs. It looks like there are just a few.

jaykrell:
Should be much better now. I still didn't follow the new-to-me rules when existing defines/includes do not allow them. This is rare, about 5 cases. I didn't think it worth adding/reordering includes/defines.

libgc is still as it was, though we could adopt G_BEGIN_DECLS for it also.

Also a future story might dispense with both G_BEGIN_DECLS and MONO_BEGIN_DECLS, and rely only on MONO_API. A quick prototype of that was promising, but I don't think it is worth the time presently.

jaykrell:
@monojenkins build failed

mono/eglib/glib.h
```diff
@@ -46,12 +46,14 @@
 
 #define __EGLIB_X11 1
 
+// FIXMEcxx in future G_BEGIN_DECLS should always be empty, and/or remove all uses,
```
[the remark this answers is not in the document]
jaykrell: I only partly understand. Is it that, we don’t require MONO_API, for static lib embedders? A quick prototype suggests it almost works, but this is with a limited set of embedders.

Still: MONO_BEGIN_DECLS should work, and only drop G_BEGIN_DECLS.

Anyway ok. The PR retains the commentary, code need not.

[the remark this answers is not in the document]
jaykrell: Yes you said that, MONO_API  missing. Retaining MONO_BEGIN_DECLS but not G_BEGIN_DECLS should work, another day.

mono/metadata/icall-table.c
```diff
@@ -36,6 +36,8 @@
 #include <mono/utils/mono-publib.h>
 #include <mono/utils/bsearch.h>
 
+G_BEGIN_DECLS // lack of prototypes
```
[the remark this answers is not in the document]
jaykrell: You mean add all the missing prototypes? Ok, understood. There are choices: in some header? All together in .c? Right before each function? First is more usual style. Last limits exposure.

mono/metadata/icall.c
```diff
@@ -107,6 +107,8 @@
 #include <sys/utsname.h>
 #endif
 
+G_BEGIN_DECLS // lack of prototypes
```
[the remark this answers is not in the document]
jaykrell: I think that is what I did but will double check. It might not matter any longer though & these can dropped. I.e. once the tree is enough C++, there is no need for C/C++ interop, like between mini & metadata. So let me just try dropping these. I do have a test bed in which all this code actually compiles & links. The changes are not hypothetical, or at least were not in the first place, but might no longer be needed.

mono/metadata/property-bag.h
```diff
@@ -12,6 +12,11 @@
 
 #include <mono/utils/mono-compiler.h>
 
+// Neither MONO_BEGIN_DECLS nor G_BEGIN_DECLS is available here. Fallback to what works.
```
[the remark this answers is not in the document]
jaykrell: That will likely work but I didn’t want that level of change here. Adding includes can break things subtley sometimes.

mono/metadata/icall-table.c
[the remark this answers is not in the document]
luhenry: Make it C++ compatible. If it's local functions/definitions then just do it in this file, if it's not local then in a shared header. Which exactly needs definition in this file?

mono/metadata/property-bag.h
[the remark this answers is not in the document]
luhenry: Did you try, or just assumed?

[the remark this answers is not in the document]
jaykrell: Did not try in this case. Test matrix is large. Been burned before by seemingly safe changes. Granted it is too much paralysis & adding includes maybe just always try w/o resistance.

jaykrell:
The problem in icalls is two part.
1. They are missing ICALL_EXPORT often.
2. We generate an incorrect prototype for them (void foo(void)). That can be replaced with extern "C", or make visible correct prototypes.
There are around 725 affected functions, i.e. either add 725 prototypes or 725 ICALL_EXPORT.
In some?many cases the prototypes already exist, but the headers just aren't included, for example sysmath.h.

Then, MONO_API and MONO_BEGIN_DECLS should say extern "C", but G_BEGIN_DECLS can go away -- except presently in glib.

luhenry:
How about `#define ICALL_EXPORT extern "C"`? You can then add the missing `ICALL_EXPORT`, harmless change.

jaykrell:
I have the other approach *almost* working and discovered some more somewhat ugliness that was in its way and covered up by C/externC.

 - icall-table.c add a few includes and copy a bunch of prototypes from icall.c
  And then remove the incorrect prototypes.
  This is pretty much goodness even in C, if people don't mind "extra" prototypes.

 - anywhere there is THREAD_INFO_TYPE, put BEGIN/END_DECLS around just those functions -- these declarations/definitions do not agree; maybe a better fix in future, such as taking gpointer and casting in the implementation.

 - if a function is in two headers and MONO_API in only one, make it MONO_API in both; or at least extern "C" in both; this does occur; maybe remove the duplication in future

 - if MONO_API is used on data -- bad idea but it occurs -- then add MONO_API_DATA instead which builds in the `extern`. And/or require they have MONO_BEGIN_DECLS / MONO_END_DECLS around them, which wasn't clearly required. This is for the opcodes array for example. Really data should be behind accessor functions, even if it is constant data, so that delayload is viable on Windows.

 - anywhere there is MONO_API  extern or extern MONO_API_EXTERNAL_ONLY functions or such, remove extern -- there is a warning otherwise for `extern extern`.

 - put extern "C" on MONO_API

 - use MONO_BEGIN_DECLS / MONO_END_DECLS in all  public headers -- same as before, but relatively  few headers

 - remove all G_BEGIN_DECLS / G_END_DECLS; it belongs only in very select places; still in eglib since I haven't ported that

 - move one enum from icall.c to a header so icall-table.c can use it.

The main downside, and you please confirm/strengthen/weaken this, is if embedders use static libs and call functions not in a public header and/or not marked MONO_API.

  Anything in a public header or with MONO_API is fully supported unchanged with this approach.

  Anything not marked MONO_API and in internal header, becomes mangled, and somewhat broken from an an embedding point of view.

I was strongly considering finishing this and submitting it as a separate PR to replace this one.

The PR in this change was the right first step months ago, when it was concievable-to-me the C++ conversion would kinda be commited and enabled, but gradually, motivating C/C++ interop among mini/utils/metadata/sgen. At this point, I think the conversion will happen "rapidly" (months later) and no matter the pace, not expect it to link until late in the process. So we only need interop with embedders, not ourselves.

Unless, as I acknowledge, embedders really call just any random function, without markings, in any header. If we are open to that level of abuse, well, that is a tough position and I don't think we are in it, as we have not been maintaining that unreasonable level of compatibility.

jaykrell:
So, we do have "a lot" of G_BEGIN_DECLS in master. Well, 36, not really a lot. Why?

mono/eglib/glib.h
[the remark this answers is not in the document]
jaykrell: All functions not MONO_API -- but  such functions are at least in public headers and within MONO_BEGIN_DECLS / MONO_END_DECLS, right?? Or people reach into the non-public headers? And if so, how can ever remove anything?

This was bugging me so I got it to work, but I really need  to get onto the other compilation errors, as there is still a lot to merge there. This maybe was not the best to try to move over earlier.

luhenry:
The only 2 files there seem to be a problem with are `icall.c` and `icall-table.c` and from your explanation it seems like they are just not C++ compatible.

This also begs the question of why should all the functions be `extern "C"`. As you already proposed, we could have `MONO_API` and `ICALL_EXPORT` contain `extern "C"` and decorate the public headers with appropriate `MONO_BEGIN_DECLS` as needed. https://github.com/mono/mono/pull/9782 seems to fill up this purpose.

jaykrell:
https://github.com/mono/mono/pull/9814 instead.

PR #9785, opened 2018-07-28
[cxx] C++ does not allow goto around initialization.

jaykrell:
The result would be undefined/uninitialized -- presumably that is what we get in C,
not as safe as it seems.

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -2383,7 +2383,8 @@ mono_class_proxy_vtable (MonoDomain *domain, MonoRemoteClass *remote_class, Mono
 
 	/* initialize vtable */
 	mono_class_setup_vtable (klass);
-	MonoMethod **klass_vtable = m_class_get_vtable (klass);
+	MonoMethod **klass_vtable;
+	klass_vtable = m_class_get_vtable (klass);
 	for (i = 0; i < m_class_get_vtable_size (klass); ++i) {
 		MonoMethod *cm;
 		    
@@ -4604,7 +4605,8 @@ make_transparent_proxy (MonoObjectHandle obj, MonoError *error)
 	MonoDomain *domain = mono_domain_get ();
 	MonoRealProxyHandle real_proxy = MONO_HANDLE_CAST (MonoRealProxy, mono_object_new_handle (domain, mono_defaults.real_proxy_class, error));
 	goto_if_nok (error, return_null);
-	MonoReflectionTypeHandle reflection_type = mono_type_get_object_handle (domain, m_class_get_byval_arg (mono_handle_class (obj)), error);
+	MonoReflectionTypeHandle reflection_type;
+	reflection_type = mono_type_get_object_handle (domain, m_class_get_byval_arg (mono_handle_class (obj)), error);
 	goto_if_nok (error, return_null);
 
 	MONO_HANDLE_SET (real_proxy, class_to_proxy, reflection_type);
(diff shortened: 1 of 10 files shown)


mono/metadata/object.c
```diff
@@ -4649,6 +4651,11 @@ create_unhandled_exception_eventargs (MonoObjectHandle exc, MonoError *error)
 {
 	MONO_REQ_GC_UNSAFE_MODE;
 
+	if (0) { // This is at the top instead of the bottom to avoid goto around initialization.
```
jaykrell: This is a different solution than the rest.

mono/metadata/appdomain.c
```diff
@@ -530,7 +530,8 @@ mono_domain_create_appdomain_checked (char *friendly_name, char *configuration_f
 	}
 	MONO_HANDLE_SET (setup, configuration_file, config_file);
 
-	MonoAppDomainHandle ad = mono_domain_create_appdomain_internal (friendly_name, setup, error);
+	MonoAppDomainHandle ad;
+	ad = mono_domain_create_appdomain_internal (friendly_name, setup, error);
 	goto_if_nok (error, leave);
```
vargaz: Whats the difference between these two lines ?

jaykrell: Cannot goto around initialization in C++. First form is illegal in C++, because there is a goto that bypasses it, leaving the variable uninitialized (hypothetically — the code is just illegal and will not compile). It is uninitialized in the second case, and in the first case for C, but C++ requires the more explicit lack of initialization.

I.e. the style is dubious but valid in C. I.e. what if you use the variables at the goto target? The code suggests they are initialized but they are not. C++ disallows it.

jaykrell:
The code is still not great, but the goal so far is no semantic change. Valid C, valid C++, same semantics as before either way.

mono/metadata/appdomain.c
vargaz: I mean there is no goto between the two lines.

jaykrell: Maybe I edited an extra line, but in general there is a goto from before to after.

jaykrell:
```
$ cat /s/1.cpp
void f1()
{
 goto L1;
 int i = 0;
L1:
 return;
}
$ clang++ -c /s/1.cpp
/s/1.cpp:3:2: error: cannot jump from this goto statement to its label
 goto L1;
 ^
/s/1.cpp:4:6: note: jump bypasses variable initialization
 int i = 0;
     ^
1 error generated.
```

mono/metadata/appdomain.c
jaykrell: Alternate solution is move locals to the top, perhaps initialized with zeros, or not -- C89 style.
Or goto start of function instead of end, as I did once or twice (once in this PR, once in another I think).
All three work.
Eventually we can use "early return" i.e. with  destructors, but not yet.

These are all good solutions.

jaykrell: Moving the handles to top in particular  is good, so they can have  constructors -- which will otherwise bring back the same error. Currently they are uninitialized and that is allowed..

jaykrell:
@monojenkins build Linux i386

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build Linux AArch Coop Suspend

jaykrell:

mono/metadata/object.c
kumpera: This looks like dead code, rather stay with the extra line than this.

jaykrell: It isn't dead because there is a label but I'm ok either way -- and people do not like the other way either..though the other way does work more and maybe, consistency. Let me update then..

jaykrell: The old code also had a return before it, so similarly dead, but ok, I'm updating.

jaykrell:
Had a small merge conflict with mono/master, resolved.

jaykrell:
Gotta rebase/merge..

jaykrell:
Just waiting for Linux ARMv7 Interpreter.

PR #9787, opened 2018-07-28
Change corlib version to a string, so that you can use uuidgen (or a git hash or github PR).

jaykrell:


configure.ac
```diff
@@ -32,18 +32,28 @@ MONO_VERSION_MAJOR=`echo $VERSION | cut -d . -f 1`
...
+# names, birthdays, branch names -- or use funny characters that
+# will not interoperate with C, C++, sh, make, etc. -- spaces, percents, stars, dollars, parens, etc.
 #
-# This can be reset to 0 when Mono's version number is bumped
-# since it's part of the corlib version (the prefix '1' in the full
-# version number is to ensure the number isn't treated as octal in C)
-MONO_CORLIB_COUNTER=26
-MONO_CORLIB_VERSION=`printf "1%02d%02d%02d%03d" $MONO_VERSION_MAJOR $MONO_VERSION_MINOR 0 $MONO_CORLIB_COUNTER`
+MONO_CORLIB_UUID=07F6F369B78D4D709FE90C0EA002A970
+MONO_CORLIB_VERSION=`printf "1%02d%02d%02d%s" $MONO_VERSION_MAJOR $MONO_VERSION_MINOR 0 $MONO_CORLIB_UUID`
```
akoeplinger: we can simplify this even more: the corlib version doesn't need to include the Mono version. It only needs to be unique which is already covered by the uuid.

I suggest we just do `MONO_CORLIB_VERSION=07F6F369B78D4D709FE90C0EA002A970`

configure.ac
```diff
@@ -32,18 +32,28 @@ MONO_VERSION_MAJOR=`echo $VERSION | cut -d . -f 1`
...
 # of classes the runtime knows about, changing icall signature or
-# semantics etc), increment this variable.
+# semantics etc), change this variable.
+#
+# This must be unique relative to corlib interface/semantics.
+#
+# Use uuidgen or make up a really reliably unique string, such as a git hash
+# or github PR number. Do not use anything personally identifiable or not particularly unique --
+# names, birthdays, branch names -- or use funny characters that
+# will not interoperate with C, C++, sh, make, etc. -- spaces, percents, stars, dollars, parens, etc.
```
akoeplinger: I'd remove all the trivia about PII/birthdays etc. and just mention the uuidgen command you need to generate the string in the right format.

mono/metadata/appdomain.c
```diff
@@ -379,27 +384,29 @@ mono_get_corlib_version (void)
...
-	int version = mono_get_corlib_version ();
-	if (version != MONO_CORLIB_VERSION)
-		return g_strdup_printf ("expected corlib version %d, found %d.", MONO_CORLIB_VERSION, version);
+	char *version = mono_get_corlib_version ();
+	if (!version) {
+		result = g_strdup_printf ("expected corlib string (%s) but not found or not string", MONO_CORLIB_VERSION);
+		goto exit;
+	}
+	if (strcmp (version, MONO_CORLIB_VERSION) != 0) {
+		result = g_strdup_printf ("expected corlib version %s, found %s.", MONO_CORLIB_VERSION, version);
```
akoeplinger: Let's improve the error message while we're here, something like: "The runtime didn't find the mscorlib.dll it expected. Expected interface version %s but found %s. Check that your runtime and class libraries are matching."

vargaz:
This will make it impossible to know which way the incompatibility is, i.e. whenever corlib is too new or too old.

akoeplinger:
That's true but does it matter that much?

kg:
We could always just add a simple time component to the front of the id, like y-m-d + uuid. The odds of the monolite changing multiple times per day are not very high, and we'll still have uniqueness in that case, but we'll be able to easily identify whether a given version is old or not. For precise time/date we can search for the ID in the repo history.

jaykrell:
There is no such thing as “old” or “new”, just matching or not. The version check was and is exact, not ordered.

mono/metadata/appdomain.c
jaykrell: Agreed.

configure.ac
jaykrell: Ok.

jaykrell:
The guid does include a high precision timestamp fyi. Or at least they used to. But it isn’t obvious from sight. I originally left the version parts in for humans who like to see them and that still seems reasonable appeasement, albeit unnecessary, granted.

scripts/update-corlib-version
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/perl
+
+use warnings FATAL => 'all';
```
luhenry: Use bash.

mono/metadata/appdomain.c
```diff
@@ -379,27 +384,32 @@ mono_get_corlib_version (void)
 const char*
 mono_check_corlib_version (void)
 {
-	const char* res;
+	char *result = NULL;
```
luhenry: There is a reason for these 2 functions to be separated. Revert it.

jaykrell: It just seems like extra layers to read through and nobody else uses the inner layer. It is unnecessary exposure, even if static. But ok.

scripts/update-corlib-version
jaykrell: Bash is to be avoided imho. Python is good choice though.

jaykrell: Every time I try to do anything in Bourne shell, many times across many years, I always come away disappointed and resolved not to waste more time on it.
It feels like a pile of nonsense and not an actually designed language.
This time the same.

For example:

```
!/usr/bin/bash

uuid=`uuidgen`

while IFS= read -r line; do
	printf "%s\n" "${line//CORLIB_VERSION=.+/CORLIB_VERSION=$uuid}"
done
```

Close, but the .+ part of the regular expression doesn't work.

It is noticably slower than Perl, probably due to applying the regexp on every line instead of en masse once over the entire file.

Attempt to run regexp just once is faster when the regexp doesn't match but hangs and cannot control-c or z. Perhaps it kinda loops.

```
#!/bin/bash

contents=`cat configure.ac`

uuid=`uuidgen`

printf "%s\n" "${contents//CORLIB_VERSION=/CORLIB_VERSION=$uuid}"
```

Perhaps we can agree on Python?
But the Python 2.x vs. 3.x situation suggests Python developers do not take compatibility seriously enough.

Anyway, if Perl isn't acceptable, I wrote it in C also. That is easier and more portable than bash.

akoeplinger: We definitely don't want a C version and even the bash version is debatable. 

I don't think we need a helper script for essentially running `uuidgen` and copy/pasting a string to a file, you can do that manually in 2 seconds.

Add a comment to configure.ac instead, something along the lines of:

```
# Use the uuidgen command to create a unique string
MONO_CORLIB_VERSION=07F6F369B78D4D709FE90C0EA002A970
```

jaykrell: I think command + copy/paste is a bit much to ask, when the automation is so easy. As an obvious side-benefit, the automation is now easy. It was kinda impossible before, because I would do such things as try to increment past concurrent PRs.

jaykrell:

mono/metadata/appdomain.c
jaykrell: What is the reason? It should be in comments in the code. As the code stands, it seems like extra unnecesary functionality to provide both functions.

scripts/update-corlib-version
jaykrell: The C version is because Perl was objected to. I agree it is borderline. However, consider that you might want this is part of a nightly or weekly automation as well. i.e. 1. Otherwise we are promulgating binaries, which while built by infrastructure, derive from temporary private branches, not actual current merged master 2. Because the behavioral vs. calling interface variable is perhaps subtle and people won't abide by it, so rebuild every so often just in case.

jaykrell:
@monojenkins build failed

jaykrell:
Good?

jaykrell:
Debian package: ‘’m09:57:10 // Include does not exist: /build/mono-snapshot-pull-request-9787-2018.07.29+09.47.25-2018.07.29+09.47.25/mcs/class/System.Core/pipes_pns.sources
09:57:10 // gensources produced 1 error(s) and a set of 422 filename(s)
09:57:10 // Invoked with ‘’’

scripts/update-corlib-version
jaykrell: I wrong about infra needing to increment/change. Like today, it can rebuild and overwrite.

luhenry: Bash doesn't introduce a dependency. We don't want more dependencies than we actually need. Pretty easy to apply, especially in this case where all we want to do is invoke an executable. And as pointed by @akoeplinger, even easier: don't have a script but document what's needed.

jaykrell: The dependency is optional as well.

jaykrell:

jaykrell:
@monojenkins build failed

akoeplinger:
Looks good, thank you!

jaykrell:
I don't really see the difference between one optional line of Perl in a comment, vs. 8 in a file.
Well, the difference is the one line doesn't work on native Windows but the 8 did, alas.
I'll add another comment for Windows support.

PR #9789, opened 2018-07-28
[mini] Increase inline cost linearly to number of calls.

nsivov:
The issue is that for large enough methods it is relatively easy to get an overflow of 'inline_costs', that will trigger negative value check later in this function, leading to 'Method too complex' exception. The reason for that obviously is that every next call get increasing penalty all the time, as 10 * (n - 1), and that accumulates fast enough when thousands of calls are involved.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -83,6 +83,7 @@
 #include "mini-runtime.h"
 
 #define BRANCH_COST 10
+#define CALL_COST 10
 /* Used for the JIT */
 #define INLINE_LENGTH_LIMIT 20
 /* Used to LLVM JIT */
@@ -7555,7 +7556,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 				UNVERIFIED;
 			}
 
-			inline_costs += 10 * num_calls++;
+			inline_costs += CALL_COST * MIN(10, num_calls++);
 
 			/*
 			 * Making generic calls out of gsharedvt methods.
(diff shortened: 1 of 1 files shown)


jaykrell:
What was the original intent? I.e. while large methods should work, avoid the overflow, good catch, was the intent increasing cost per call or not? Using longlong would be another fix if so, or checking for large value at each increment instead letting it keep climbing.

nsivov:
I don't really know what was the idea originally, according to blame it always worked like this. But yes, another option is to use larger int type, which will be more invasive, or we can keep existing type, but add check on every increment so it never goes over some factor of existing threshold of 60.

Personally I don't understand why earlier calls increase penalty of next ones, but I could be missing a bigger picture of course.

jaykrell:
I agree it is suspicious. Perhaps the intent was actually how you have it.

lewurm:
@monojenkins build

lewurm:
while it looks okay, this will change perf behaviour. not sure if we want that. let's see what CI says.

lewurm:
@kumpera @vargaz what do you guys think?

vargaz:
Can't we just make inline_costs an int64 ?

kumpera:
We should increase the inline cost based on number of calls, but we should use a ceiling.
I'd say we use something like `10 * MIN(10, num_calls++)`.

nsivov:
@vargaz there is two issue here, one is the policy on how to increment the cost, and another is about comparing resulting cost to some threshold. Those are separate, overflow that I get could certainly be fixed with larger int type, but personally I don't think it's pretty, I'd rather move threshold condition somewhere closer to where cost is incremented.

@kumpera making sure I understood, do you mean counting num_calls first, and then updating the cost once? Also, why would it be a cost of 10 * 10 for num_calls == 0?

kumpera:
@nsivov what I mean is to keep the logic of counting the number of calls but limit the cost of calls we care for to 10. IE, the cost of calls are between 10 and 100.

nsivov:
Alright, will do.

nsivov:
@kumpera, pushed the update, please take a look when you got time.

nsivov:
Thank you.

PR #9793, opened 2018-07-29
[mono-api-tools] Make the tools work as libraries

mattleibow:
Right now, the set of mono-api-tools only works as console apps. This PR will make them work as normal dlls so they can be consumed by other apps and libraries. They will still work as .exes, so nothing visibly changes for existing users.

The motivation for this is for some build process or third party tool to take an assembly and generate an .xml doc from it. Since the writer is in-memory, the tool can then parse the .xml and do whatever it needs to. The same goes for the diff tools - if the diff happens in-memory, then the tool can process it without having to save and then load from the file system.

## Libraries
 - [x] mono-api-info -> `Mono.ApiTools.ApiInfo`
 - [x] mono-api-diff -> `Mono.ApiTools.ApiDiff`
 - [x] mono-api-html -> `Mono.ApiTools.ApiDiffFormatted`

## Major changes
 - the existing `Driver` classes are now just thin wrappers to parse the args and pass it on to the generators
 - the "options" are now abstracted into config classes to avoid methods with numerous parameters
 - the APIs have new support for reading assembly and info `Stream`s instead of just file paths
 - added compiler flags to control the namespace and type visibility to make the library APIs nicer
 - added compiler flag to remove the `Driver` class and `Mono.Options` dependencies
 - the mono-api-info tool has some more options to allow it to run on libraries, and ignore dependencies that it cannot find - this makes it easier to use so that the entire dependency chain does not have to be loaded just to output the current assembly API

## Compiler defines
 - `EXCLUDE_DRIVER` will remove the console entry point and remove the dependency of Mono.Options
 - `USE_MONO_API_TOOLS_NAMESPACE` will switch the namespace to `Mono.ApiTools` so that libraries can have a nice API
 - `USE_INTERNAL_VISIBILITY` will make any non-essential, `public` APIs turn `internal`

## Review points
 - are there any variables/fields that are still static?
 - are there any exposed types/members that need to be hidden?
 - are the new names for the APIs and config classes OK?
 - are my whitespaces according to the guidelines?
 - did I go horribly wrong?


mcs/tools/corcompare/AssemblyResolver.cs
```diff
@@ -32,9 +32,13 @@
 
 using Mono.Cecil;
 
-namespace GuiCompare {
+#if USE_MONO_API_TOOLS_NAMESPACE
+namespace Mono.ApiTools {
+#else
+namespace CorCompare {
```
mattleibow: There appears to be a weird namespace difference here. I changed it to match the rest, but I can revert this.

mcs/tools/corcompare/Util.cs
```diff
@@ -3,15 +3,17 @@
...
-using GuiCompare;
-
+#if USE_MONO_API_TOOLS_NAMESPACE
+namespace Mono.ApiTools {
+#else
 namespace CorCompare {
+#endif
 
-	static class TypeHelper {
+	class TypeHelper {
```
mattleibow: This class was made to be an instance class.

mcs/tools/corcompare/mono-api-info.cs
```diff
@@ -22,18 +22,20 @@
...
-{
-	public class Driver
+#if USE_MONO_API_TOOLS_NAMESPACE
+namespace Mono.ApiTools {
+#else
+namespace CorCompare {
+#endif
+
+#if !EXCLUDE_DRIVER
+	class Driver
```
mattleibow: The driver can be excluded for library builds.

mcs/tools/corcompare/mono-api-info.cs
```diff
@@ -211,10 +260,12 @@ AssemblyDefinition LoadAssembly (string assembly)
 	abstract class BaseData
 	{
 		protected XmlWriter writer;
+		protected ApiInfoConfig config;
 
-		protected BaseData (XmlWriter writer)
+		protected BaseData (XmlWriter writer, ApiInfoConfig config)
 		{
 			this.writer = writer;
+			this.config = config;
```
mattleibow: We do have to pass the config object around.

mcs/tools/mono-api-diff/mono-api-diff.cs
```diff
@@ -32,14 +35,14 @@ static int Main (string [] args)
 			string output = null;
 			List<string> extra = null;
 
-			var options = new OptionSet {
+			var options = new Mono.Options.OptionSet {
```
mattleibow: I didn't want to include another `#if` just to flag the using.

mcs/tools/mono-api-html/ApiChange.cs
```diff
@@ -4,20 +4,29 @@
...
-	public class ApiChange
+#if USE_MONO_API_TOOLS_NAMESPACE
+namespace Mono.ApiTools {
+#else
+namespace Xamarin.ApiDiff {
+#endif
+
+#if !USE_INTERNAL_VISIBILITY
+	public
+#endif
```
mattleibow: Because this tool has all it's types public, someone may be using it somewhere. So, I just have a flag for making it internal for those who don't want to see them.

mcs/tools/mono-api-html/ApiDiff.cs
```diff
@@ -37,127 +37,98 @@
...
+		public Formatter Formatter { get; set; }
+		public string Assembly { get; set; }
+		public string Namespace { get; set; }
+		public string Type { get; set; }
+		public string BaseType { get; set; }
+		public string Parent { get; set; }
+		public int Indent { get; set; }
+		public List<Regex> IgnoreAdded { get; } = new List<Regex> ();
+		public List<Regex> IgnoreNew { get; } = new List<Regex> ();
+		public List<Regex> IgnoreRemoved { get; } = new List<Regex> ();
```
mattleibow: Using newer C# readonly properties

mcs/tools/mono-api-html/ApiDiff.cs
```diff
@@ -218,4 +172,99 @@ public static int Main (string[] args)
...
 		}
 	}
+#endif
+
+	public enum ApiDiffFormatter {
+		Html,
+		Markdown,
+	}
+
+	public class ApiDiffFormattedConfig {
```
mattleibow: The `ApiDiffFormattedConfig` type is new and different from `State` in that `State` also contains temporary data during the actual processing.

I _can_ merge the types and make things internal and all that, but this way there is a clear separation - and this config is never modified by the library.

mcs/tools/corcompare/AssemblyResolver.cs
akoeplinger: I think we could change the namespace to `Mono.ApiTools` everywhere and remove the define, the .exe's shouldn't be loaded via reflection so nobody should be broken by the namespace change.

mcs/tools/mono-api-diff/mono-api-diff.cs
```diff
@@ -53,35 +56,92 @@ static int Main (string [] args)
...
 			XmlNode node = doc.SelectSingleNode ("/assemblies/assembly");
 			XMLAssembly result = new XMLAssembly ();
+#if !EXCLUDE_DRIVER
 			try {
+#endif
 				result.LoadData (node);
+#if !EXCLUDE_DRIVER
 			} catch (Exception e) {
-				Console.Error.WriteLine ("Error loading {0}: {1}\n{2}", file, e.Message, e);
+				Console.Error.WriteLine ("Error loading {0}: {1}\n{2}", (stream as FileStream)?.Name ?? "<stream>", e.Message, e);
```
akoeplinger: What do you think about abstracting the output handling into a class that either writes to`Console.Error`/`Console.Out` (for .exe) or a StringBuilder (for .dll)?

That way you can get ahold of the log messages even in a library build and we don't need to litter the code with `EXCLUDE_DRIVER` ifdefs.

mcs/tools/mono-api-html/ApiChange.cs
akoeplinger: I did a quick GitHub search across the Mono and Xamarin organization and everyone seems to be using the whole tool and not just pieces so we might be able to just make this internal all the time.

mcs/tools/corcompare/mono-api-info.cs
```diff
@@ -22,107 +22,224 @@
 using Mono.Cecil.Cil;
 using System.IO;
 
-namespace CorCompare
-{
-	public class Driver
+namespace Mono.ApiTools {
+
+#if !EXCLUDE_DRIVER
```
mattleibow: The driver classes (console app part) can still be excluded for libraries

mcs/tools/corcompare/mono-api-info.cs
```diff
@@ -196,25 +317,27 @@ public void DoOutput ()
...
-					return TypeHelper.Resolver.ResolveFile (assembly);
+			if (File.Exists (assembly))
+				return state.TypeHelper.Resolver.ResolveFile (assembly);
 
-				return TypeHelper.Resolver.Resolve (AssemblyNameReference.Parse (assembly), new ReaderParameters ());
-			} catch (Exception e) {
-				Console.WriteLine (e);
-				return null;
-			}
+			return state.TypeHelper.Resolver.Resolve (AssemblyNameReference.Parse (assembly), new ReaderParameters ());
```
mattleibow: No use catching the exception here. If this assembly cannot be loaded, then the very assembly we are trying to generate info for is not loaded. If the assembly is not loaded, nothing will be generated. If nothing is generated, then why do we even need this tool? If we wanted something to be generated, then throw an error so we know it failed. #Logic

mcs/tools/mono-api-diff/mono-api-diff.cs
```diff
@@ -50,38 +49,86 @@ static int Main (string [] args)
...
-			doc.Load (File.OpenRead (file));
+			doc.Load (stream);
 
 			XmlNode node = doc.SelectSingleNode ("/assemblies/assembly");
 			XMLAssembly result = new XMLAssembly ();
-			try {
-				result.LoadData (node);
-			} catch (Exception e) {
-				Console.Error.WriteLine ("Error loading {0}: {1}\n{2}", file, e.Message, e);
-				Environment.Exit (1);
```
mattleibow: What is the point of catching the exception and then quitting anyway? Just throw and let the caller handle it - chances are they will do a better job.

mcs/tools/mono-api-diff/mono-api-diff.cs
```diff
@@ -451,7 +498,9 @@ public override void LoadData (XmlNode node)
 			}
 
 			if (atts == null || atts.Name != "namespaces") {
-				Console.Error.WriteLine ("Warning: no namespaces found!");
+#if !EXCLUDE_DRIVER
+				Console.Error.WriteLine (@"Warning: no namespaces found for {name}");
+#endif
```
mattleibow: Leaving this here as there are only 2 instances of writing to the console - and these aren't very important messages at all.

mattleibow: In this case, I didn't want to do some hectic abstraction or some advance mechanism just for two boring messages that are probably going to be ignored anyway since they are potentially valid cases. Sure, and assembly without namespaces is weird, but my just be an empty assembly. And sure, an empty namespace is pointless, but is it really bad?

mattleibow:
@akoeplinger, made some changes.

mattleibow:
@akoeplinger I am playing around with this extra flag to remove the need for downloading an entire dependency tree when you just want to see if you have removed any methods on YOUR type: https://github.com/mattleibow/mono/pull/5

mattleibow:
@marek-safar I have been testing this PR: https://github.com/mattleibow/mono/pull/5 I think it adds some useful things and a bugfix.

 - Adds support for optionally ignoring inherited interfaces  
   This is useful when we are trying to generate the API from what is in the assembly, and not what comes with a framework/library
 - Adds support for optionally ignoring resolution errors  
   This is again useful for when we only care about the current assembly, and not where or what comes with the framework. This also helps with libraries so that all the dependencies don't need to be provided
 - Fixed a bug where nested interfaces would be removed/excluded from the inheritance tree

akoeplinger:
@mattleibow love it, thanks for making these improvements!

PR #9794, opened 2018-07-29
[cxx] Make main be C, because cannot link with -xc++, it treats

jaykrell:
the libraries as C++ (i.e. C++ textual source instead of binary archives).
Renaming main to .cpp would help a little.
Main is small so shouldn't be a big deal and be further factored if needed/wanted.

This can be revisited much later but is definitely the path of progress.
i.e. it actually lets the rest of mini/sgen/utils/metadata compile as C++ and link,
leaving just main.c as C.

Other than renaming, having foocpp.cpp #include foo.c works, and preserves history better,
albeit is "surprising".

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -182,19 +182,21 @@ endif
 mono_boehm_SOURCES = \
 	main.c
 
-mono_CFLAGS = $(AM_CFLAGS)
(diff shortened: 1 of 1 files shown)


jaykrell:
Required CI lines are green.

jaykrell:
@monojenkins build failed

jaykrell:
This is probably redundant now, need to check.

jaykrell:
Redundant with merged https://github.com/mono/mono/pull/10115.

PR #9795, opened 2018-07-29
[cxx] Better types at the mono_gc_alloc layer per https://github.com/…

jaykrell:
…mono/mono/pull/6695#discussion-diff-167398359L5601.

This allows for fewer casts from void* to non-void* elsewehere.


jaykrell:
This change is out of sequence, closing for now, will try later.

PR #9799, opened 2018-07-29
[cxx] Port coop handles to valid C++.

jaykrell:
In particular the conversion from void* to non-void*, with the
non-void* type being contextual (the function return type), was a problem.

There are already two forms of coop handles -- type-safe and type-unsafe.

In converting to C++ we are therefore faced with the prospect of creating
two more variations. For now, create just one more, type-unsafe.

This otherwise contributes to errors all around the tree, i.e. anywhere
one is returning a coop handle as a raw pointer.
i.e. all uses of `HANDLE_FUNCTION_RETURN_OBJ` and `ICALL_RETURN_OBJ`

Also C++ requires initialization of consts.
You cannot say:
	```const int i; // zero```
You have to say:
	```const int i = 0;```

Also void* to non-void* cast in `MONO_ARRAY_HANDLE_PIN` as required by C++.

This is probably the biggest seeming change, in conversion to valid C+, and at first glance people will think this is unwarranted premature improvement by taking advantage of C++.
However 1. It is mostly under ifdef so barely touches C
2. Something is required. We have to cast from void* to non-void*, and the non-void* in question is not directly in hand, in C++98. It isn't passed to the macros. It is 1. the return type of the function we are returning from, which is not capturable and 2. A type associated with a parameter to the macros -- which is what this leverages. With C++11 we could use `auto x = MONO_HANDLE_RAW (handle)`, and the majority of the change would fall away. auto is a notable virtue of C++11, but I do not want C++11 language dependencies at this point, nor to argue the merits/demerits of `auto` specifically.

### DIFF
--- a/mono/metadata/handle.c
+++ b/mono/metadata/handle.c
@@ -94,7 +94,7 @@ free_handle_chunk (HandleChunk *chunk)
 	g_free (chunk);
 }
 
-const MonoObjectHandle mono_null_value_handle;
+const MonoObjectHandle mono_null_value_handle = { 0 };
 
 #define THIS_IS_AN_OK_NUMBER_OF_HANDLES 100
 
--- a/mono/metadata/handle.h
+++ b/mono/metadata/handle.h
@@ -405,7 +405,6 @@ MONO_HANDLE_TYPECHECK_FOR (TYPE) (TYPE *a)			\
 }
 
 #else
-
 #define TYPED_HANDLE_DECL(TYPE)						\
 	typedef struct { TYPE *__raw; } TYPED_HANDLE_PAYLOAD_NAME (TYPE) ; \
 	typedef TYPED_HANDLE_PAYLOAD_NAME (TYPE) * TYPED_HANDLE_NAME (TYPE); \
--- a/mono/metadata/handle.h
+++ b/mono/metadata/handle.h
@@ -54,7 +54,6 @@
 #define MONO_TYPE_SAFE_HANDLES 0 // PowerPC, S390X, SPARC, MIPS, Linux/x86, BSD/x86, etc.
 #endif
 
-G_BEGIN_DECLS
 
 /*
 Handle stack.
@@ -245,9 +244,9 @@ Icall macros
 // Return a raw pointer from coop handle.
 #define HANDLE_FUNCTION_RETURN_OBJ(HANDLE)			\
 	do {							\
-		void* __result = MONO_HANDLE_RAW (HANDLE);	\
+		MONO_HANDLE_AUTO __result = MONO_HANDLE_RAW (HANDLE);	\
 		CLEAR_ICALL_FRAME;				\
-		return mono_typed_handle_cast_voidptr ((HANDLE), __result); \
+		return __result;				\
 	} while (0); } while (0);
 
 #if MONO_TYPE_SAFE_HANDLES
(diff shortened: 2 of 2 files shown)


jaykrell:
@monojenkins build

jaykrell:

jaykrell:

jaykrell:
That is not what this is really doing. This is making the code compile successfully as C++. It isn't possible without ifdef cplusplus or a more invasive change. Because of the void* to non-void* cast without knowing what the non-void* is.

jaykrell:
The goal is to make the tree be valid C++ and valid C at the same time, with minimal change.
I don't know how else to do this.

jaykrell:

jaykrell:
Review? (CLA??)

jaykrell:
Again, the goal is simultaneously valid correct C & C++ and I do not know how to this otherwise. This isn’t taking advantage of C++ or making things better because C++ is available. It is making everything able to compile & link as C++. Consider the use of extern “C” already in the code. That is also C++.

kumpera:
@migueldeicaza what do you want us to do here? This PR uses C++ features

lambdageek:
Okay, so let's say we allow a bit of C++ to make uses of `HANDLE_FUNCTION_RETURN_OBJ (someHandle)` valid C++.

I would argue that:
```cplusplus
template<typename T>
struct TypedHandle {
   T *_raw;
}
```
is _not_ the C++ that we want:
1. it makes every instantiation `TypedHandle<MonoString>` and `TypedHandle<MonoObject>` unrelated to each other - they do not derive from a common representation, for example.
2. All we wanted was to fixup one function and instead we get a whole new representation.

What about:
```cplusplus
   template <typename Handle>
   decltype (Handle().__raw) get_raw_ptr (Handle h) { return h.__raw; }
```

Needs C++11, granted, but it localizes the need for C++ just to the one problem we're trying to solve without imposing a new data structure already.

jaykrell:
The C++ we want is a future matter. I'm just aiming for something that works.
If we are ok with C++11 at least in the short term, I can try auto instead, that might suffice.
I was aiming for C++98.

The unrelatedness TypedHandle<String> and TypedHandle<MonoObject> is the same as in C.

jaykrell:
Better now? I'm just hoping that C++11 doesn't kill C++.

jaykrell:
I like auto, as long as C++11 is ok. I just don't want movement toward C++11 to wreck movement toward C++.

migueldeicaza:
At least the current version of this patch seems like it compiled both with C and C++ compilers, so I do not see a problem with this.

And generally, I am supportive of using C++11 features for the C++ code paths.

jaykrell:
The original was C++98 with “a few”
lines of C++ under ifdef. The latest is C++11 with “one” line. There may be another C++98 approach. But it sounds moot.

jaykrell:
The fix for void* to non-void* provides g_cast which can be sprinkled around, and works here, and works with C++98. So let's hold off on this PR.

jaykrell:
Though some parts of this PR will stand, e.g. the zero initailization.

jaykrell:
https://github.com/mono/mono/pull/10027/files is  an alternative to some of this, and does a lot more.

jaykrell:
Ok, given the desire to narrow the alternative, let's go back to this one. Which I still need to rebase/merge.

jaykrell:
This should be redundant with master now.

jaykrell:
Mostly redundant with master now, but we should use `auto`, and revisit the other `g_cast` here.

PR #9802, opened 2018-07-30
[cxx] CentOS 6 gcc 4.4 and Android stdint.h/inttypes.h does not give everything without asking.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
Review?

mono/eglib/glib.h
```diff
@@ -1,5 +1,17 @@
 #ifndef __GLIB_H
 #define __GLIB_H
+
+// Ask stdint.h for the full C99 features even for C++98 CentOS 6 g++ 4.4.
+#if defined(__cplusplus) && !defined(__STDC_LIMIT_MACROS)
```
lambdageek: I think this whole block needs to be wrapped in `#if defined (__cplusplus) && (cplusplus <= 199711L)`
Because in C++11 this is no longer necessary (?) https://stackoverflow.com/a/26523878/570682

jaykrell: I’ll have to try centos6 again.

jaykrell: It is needed, for the centos6/gcc4.4 C++ prerelease.
Though include cstdint maybe the nicer workaround.
That is where the described compiler-supplied workaround is.

jaykrell:
No that workaround is incomplete, it doesn't provide the printf macros.
Let's go with the original.

jaykrell:
Actually yes. __cplusplus is 1 on gcc4.4 with compiling C++0x.

jaykrell:
And original insufficient, nothing found yet.
`sgen-mono-ilgen.c:321: error: ‘INT32_MAX’ was not declared in this scope`

jaykrell:
https://github.com/mono/mono/pull/10044 has all the Centos6/g++4.4 fixes including making this work, and be narrower than this.

jaykrell:
Eh, apparently this is all correct behavior, and Android does it too. So the original is needed.
I've seen breaks on other platforms that this was fixing.

See
 $HOME/android-toolchain/toolchains/armeabi-v7a-clang/sysroot/usr/include/inttypes.h  $HOME/android-toolchain/toolchains/armeabi-v7a-clang/sysroot/usr/include/stdint.h

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X i386

PR #9807, opened 2018-07-30
[jit] Clean up the handling of the lower/higher parts of long values on 32 bit platforms.

vargaz:
Instead of accessing them using the inst_ls_word/inst_ms_word macros which use the union in MonoInst.data, access them normally using
MonoInst.Data.i8const.

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -210,7 +210,7 @@ mini_llvm_ins_info[] = {
 #undef MINI_OP3
 
 #if SIZEOF_VOID_P == 4
-#define GET_LONG_IMM(ins) (((guint64)(ins)->inst_ms_word << 32) | (guint64)(guint32)(ins)->inst_ls_word)
+#define GET_LONG_IMM(ins) ((ins)->inst_l)
 #else
 #define GET_LONG_IMM(ins) ((ins)->inst_imm)
 #endif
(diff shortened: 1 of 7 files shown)


mono/mini/mini.h
```diff
@@ -851,6 +849,20 @@ enum {
...
+	return (guint32)(ins->data.i8const & 0xffffffff);
+}
+
+/* Return the higher 32 bits of the 64 bit immediate in INS */
+static inline guint32
+ins_get_l_high (MonoInst *ins)
+{
+	return (guint32)((ins->data.i8const >> 32) & 0xffffffff);
+}
+
```
lewurm: inst_ls_word/inst_ms_word respected byte order, this doesn't

mono/mini/mini-llvm.c
```diff
@@ -210,7 +210,7 @@ mini_llvm_ins_info[] = {
 #undef MINI_OP3
 
 #if SIZEOF_VOID_P == 4
-#define GET_LONG_IMM(ins) (((guint64)(ins)->inst_ms_word << 32) | (guint64)(guint32)(ins)->inst_ls_word)
+#define GET_LONG_IMM(ins) ((ins)->inst_l)
```
lewurm: we could use the same macro for 32 bit and 64 bit, right?

mono/mini/mini.h
vargaz: They needed byte order because the i8const field is in a union with the data.op[0]/op[1] fields, this is not needed any more, since we directly read i8const.

PR #9809, opened 2018-07-30
[runtime] Make collect_threads helper pin MonoInternalThread objects

alexanderkyte:
This fix is supposed to fix a big performance issue exposed by MERP. The preexisting race doesn't appear to be causing any instability currently, but it looks like it could if hit wrong.

== 
Edit: Elaboration

The function collect_threads returns a pointer to the thread, not a handle. We do pin already. This reference is freed when the thread is detached from the runtime. There is no guarantee that the unmanaged functions interacting with these threads will not run concurrently with threads being detached. In fact, one can observe that being the case when using MERP. 

When we being a dump, we collect a list of active threads. We walk them, and signal each thread one-by-one. While dumping one thread, it is not uncommon to find that a couple of other threads die. We run into a timeout while waiting for a thread on about 6-7 threads when dumping VS4M.

When one reruns collect_threads, we don't see that thread anymore.

Now there are a few problems here. Inside of MERP, I want to be able to check the MonoInternalThread to check the thread state so I can abort early if the thread has died. It'll be much faster than what I'm doing now. But when I do so, I run into some crashes due to memory unsafety. The MonoInternalThread has been freed I'm pretty sure, because the pinned reference is no longer pinned. This is the race I am discussing. 

I think that a good first step here is to make collect_threads pin the threads in question so that racing detachment doesn't lead to memory unsafety.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -3758,7 +3758,7 @@ get_thread_dump (MonoThreadInfo *info, gpointer ud)
 
 typedef struct {
 	int nthreads, max_threads;
-	MonoInternalThread **threads;
+	guint32 *threads;
 } CollectThreadsUserData;
 
 static void
@@ -3768,7 +3768,7 @@ collect_thread (gpointer key, gpointer value, gpointer user)
 	MonoInternalThread *thread = (MonoInternalThread *)value;
 
 	if (ud->nthreads < ud->max_threads)
-		ud->threads [ud->nthreads ++] = thread;
+		ud->threads [ud->nthreads ++] = mono_gchandle_new (&thread->obj, TRUE);
 }
 
 /*
@@ -3776,7 +3776,7 @@ collect_thread (gpointer key, gpointer value, gpointer user)
  * THREADS should be an array allocated on the stack.
  */
 static int
-collect_threads (MonoInternalThread **thread_array, int max_threads)
+collect_threads (guint32 *thread_handles, int max_threads)
 {
 	CollectThreadsUserData ud;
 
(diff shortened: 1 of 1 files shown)


alexanderkyte:
@monojenkins build pkg

mono/metadata/threads.c
```diff
@@ -5907,32 +5916,42 @@ static size_t num_threads_summarized = 0;
...
+	guint32 thread_array [128];
 	int nthreads = collect_threads (thread_array, 128);
+	gboolean success = FALSE;
 
 	for (int i=0; i < nthreads; i++) {
-		MonoNativeThreadId tid = thread_get_tid (thread_array [i]);
-		if (tid == id)
-			return thread_array [i];
+		guint32 handle = thread_array [i];
+		MonoInternalThread *thread = (MonoInternalThread *) mono_gchandle_get_target (handle);
```
jaykrell: Can new code use coop?

jaykrell: I suppose in the face of pinning, it doesn't matter.

kumpera:
Hey Kyte,

Can you explain the perf issue and the race? I don’t see how this PR addresses  any of that

kumpera:
In the case of pinning, I’d be curious to understand why a handle would be needed as we're (or supposed to be) storing to the stack, that will keep those objects alive and pinned

alexanderkyte:
@kumpera I elaborated a bit more on how this fix/bug is related to the issue in MERP

alexanderkyte:
@monojenkins build failed

kumpera:
MonoInternalThread is managed memory,so it can’t simply be freed.

How does a stack reference not help here? And why does your patch requires a pinning handle?

PR #9810, opened 2018-07-30
Fix tarballs again

directhex:


mcs/class/System.Core/Makefile
```diff
@@ -4,6 +4,8 @@ include ../../build/rules.make
 
 LIBRARY = System.Core.dll
 
+EXTRA_DIST += pipes_pns.sources
```
akoeplinger: this shouldn't be necessary, I fixed it this morning: #9803

mcs/class/Makefile
```diff
@@ -410,8 +410,14 @@ DISTFILES = \
...
-	test-helpers/NetworkHelpers.cs	\
-	test-helpers/SocketResponder.cs	\
+	test-helpers/AdminHelper.cs \
+	test-helpers/Configuration.Http.cs \
+	test-helpers/NetworkHelpers.cs \
+	test-helpers/NunitHelpers.cs \
+	test-helpers/PlatformDetection.cs \
+	test-helpers/RemoteExecutorTestBase.Mobile.cs \
+	test-helpers/RemoteExecutorTestBase.Mono.cs \
+	test-helpers/SocketResponder.cs \
```
akoeplinger: I thought we dist all the .cs files from .sources? Maybe that doesn't work for tests?

directhex: I don't think DISTFILES/EXTRA_DIST can traverse _up_ a directory tree? mcs/build/tests.make assigns a full list o' files to TEST_FILES by parsing the relevant .sources files, but I think those files in test-helpers were only being included by being referenced elsewhere (e.g. the two above), not via the .sources parsing

```
directhex@flame:~/Projects/mono/mcs/class$ sed -e '/^$$/d' -e 's,^../,,' -e '/^\#.*$$/d' -et -e 's,^,Test/,' corlib/corlib_test.dll.sources | grep ^\\.\\.
../../../mono/mini/TestHelpers.cs
../test-helpers/NunitHelpers.cs
```

PR #7046, opened 2018-02-14
[coop] Enable by default

luhenry:
Part of #6921

PR #9811, opened 2018-07-30
[coop] Turn hybrid suspend on by default on desktop platforms

lambdageek:
Use hybrid suspend by default on x86 and amd64 OSX, Windows and Linux.

Pass `--disable-hybrid-suspend` to configure to disable at build time, or set
the enrivonment variable `MONO_THREADS_SUSPEND=preemptive` when running Mono to use preemptive suspend (the previous default).

Pass `--disable-hybrid-suspend --enable-cooperative-suspend` to configure
cooperative suspend at build time.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4997,6 +4997,31 @@ AC_ARG_WITH(lazy_gc_thread_creation, [  --with-lazy-gc-thread-creation=yes|no   
 	fi
 ], [with_lazy_gc_thread_creation=no])
 
+dnl *****************************
+dnl *** Thread suspend policy ***
+dnl *****************************
+
+dnl Set a default hybrid or cooperative suspend on some platforms
+
+dnl Coop default is set by the bitcode preset.
+
+dnl If coop isn't on by default, maybe hybrid should be?
+if test x$enable_cooperative_suspend_default != xyes; then
+	case $HOST in
+	X86 | AMD64)
+		dnl Some host/target confusion, there's no host_osx (and
+		dnl host_darwin would be true on iOS not just macOS).
+		if test x$target_osx = xyes; then
+			enable_hybrid_suspend_default=yes
+		elif test x$host_linux = xyes -o x$host_win32 = xyes; then
+			enable_hybrid_suspend_default=yes
+		fi
+		;;
+	esac
+fi
+
+dnl Now check if there were flags overriding the defaults
+
 AC_ARG_WITH(cooperative_gc,        [  --with-cooperative-gc=yes|no        Enable cooperative stop-the-world garbage collection (sgen only) (defaults to no)], [AC_MSG_WARN([--with-cooperative-gc is deprecated, use --enable-cooperative-suspend instead])], [with_cooperative_gc=default])
 AC_ARG_ENABLE(cooperative_suspend, [  --enable-cooperative-suspend      Enable cooperative stop-the-world garbage collection (sgen only) (defaults to no)], [], [enable_cooperative_suspend=default])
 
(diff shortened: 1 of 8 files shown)


lambdageek:
We should add an x64 Linux (and OSX?) preemptive suspend CI lane after this is merged so we can keep testing the previous configuration.  (There's already a `CI_TAGS` keyword for it - `CI_TAGS=preemptive-suspend`)

lambdageek:
@luhenry Updated to explicitly pass `--disable` flags for the sdks

lambdageek:
Now I'm second-guessing myself - do I actually want to check the HOST not the TARGET? this isn't an (AOT) compiler related feature so I think checking HOST is more correct.

lambdageek:
Could I get another review of this?  I switched the configure logic to look at `HOST` and `host_linux` and `host_win32`  (but `target_osx` - see comment in code).

I don't think it matters (since this isn't an AOT thing), but I'd love some feedback.

jaykrell:
Good point. Host should be correct. "runtimes" only have a host. I mean, even if the AOT compiler itself used a garbage collector, and configurable suspension, "host" would be correct.

jaykrell:
Unless the compiler had to inject safe points into the AOT code depending on suspension mechanism, then target could be relevant.

configure.ac
```diff
@@ -5044,6 +5044,31 @@ AC_ARG_WITH(lazy_gc_thread_creation, [  --with-lazy-gc-thread-creation=yes|no
...
+dnl *****************************
+
+dnl Set a default hybrid or cooperative suspend on some platforms
+
+dnl Coop default is set by the bitcode preset.
+
+dnl If coop isn't on by default, maybe hybrid should be?
+if test x$enable_cooperative_suspend_default != xyes; then
+	case $HOST in
+	X86 | AMD64)
```
jaykrell: I'm not sure processor is relevant..but I guess ok, it disambiguates darwin, and we don't have Windows/arm support. What about Linux/arm?

jaykrell: FIXME?  add host_macos, host_ios, and either remove host_darwin or make it clear host_darwin is maco | ios? And never say "osx"?

jaykrell: Does coop work w/o coop handles finished? We just treat transition from jit=>native differently based on if it is a handles() transition?

lambdageek: The processor check is specifically because we only want to turn this on on Intel for now - it feels like there are more deadlocks to chase down on linux on ARM.

The coop handles (plus precise managed stack scanning) would allow us to skip the "copy stackdata" step when a thread self-suspends and to stop scanning the native stack conservatively.  So not having them doesn't preclude using hybrid suspend today.

lambdageek:
@monojenkins build failed

lambdageek:
The interp failures seem related: we don't have GC Unsafe transitions in `interp_entry`.

@luhenry @lewurm What do you think: switch interp lane to use preemptive suspend for now and fixup later, or hold this PR until interp is ready?

luhenry:
@lambdageek disabling on CI will not fix the Interpreter by default, so either we always disable Hybrid Suspend while running with the interpreter, either we fix the Interpreter. I would go for the latter one, but I don't have a good idea of how long it's going to take.

lambdageek:
@luhenry The interpreter doesn't have safepoints right now - https://github.com/mono/mono/pull/7046#issuecomment-365764273.  I'm not sure if there's a usecase for interp + hybrid on desktop.  So I would prefer to go with a warning and a switch to preemptive if the interpreter and hybrid or coop are specified simultaneously.

luhenry:
@lambdageek sounds good to me, let's do that.

lambdageek:
Updated to always switch to preemptive suspend if using the interpreter.

lambdageek:
Pushed a fix for sgen-new-threads-collect failures on OSX - need to take the join lock in GC Safe mode.

PR #9812, opened 2018-07-30
[cxx] [do not merge] now with less extern "C"

jaykrell:

PR #9814, opened 2018-07-31
[cxx] Add extern "C" to MONO_API. And ICALL_EXPORT. Add missing ICALL_EXPORT.

jaykrell:
- Public headers should be wrapped in it (all but one are).
 - MONO_API should include it.

But it is not needed "everywhere".

Placement of "extern" therefore tweaked.
  - Remove from functions where it is meaningless and gets in the way.
  - Mark data with MONO_API_DATA instead of MONO_API.

[not yet]Sprinkle in a few G_BEGIN_DECLS for monodis and where THREAD_INFO_TYPE makes prototypes incorrect.

include/add prototypes for all icalls. [undone per PR, possibly a misunderstanding]

The problem would be, if embedders manage to use functions that are neither MONO_API nor
in public headers.

Some parts of the larger solution are omited here in the hopes of finding better options.
In particular, compiling all of monodis as C++ so we don't need some extern "C" for it,
and THREAD_INFO_TYPE at least incoming being gpointer so declaration and definition match.

This is based on https://github.com/mono/mono/pull/9812/commits/40315ee475bdf89c40189707a8cca86b3f6d91c8
and is an alternative to https://github.com/mono/mono/pull/9782.

We also can/should prune the preexisting G_BEGIN_DECLS / G_END_DECLS uses.


jaykrell:
So, gcc doesn't like extern  "C" between attributes, as occurs with:
`MONO_RT_EXTERNAL_ONLY MONO_API`
clang is ok with it. Solution is put them the other way around:
`MONO_API MONO_RT_EXTERNAL_ONLY`
I will try that. If that fails, or there  is another reason for the ordering, probably give up,

mono/utils/mono-publib.h
```diff
@@ -59,6 +59,27 @@ typedef unsigned __int64	uint64_t;
...
 #include <stdlib.h>
 
+// 1.cpp:4:12: warning: duplicate 'extern' declaration specifier
+//      [-Wduplicate-decl-specifier]
+// extern "C" extern int c;
+//
+// Therefore, remove extern on functions as always meaningless/redundant,
+// and provide MONO_API_DATA for data, that always has one and only one extern.
+
+#ifdef __cplusplus
```
luhenry: You should define a `MONO_EXTERN_C extern "C"` and use that at https://github.com/mono/mono/pull/9814/files#diff-bd8dbfad6e5c0280586ab06f8513939dR84 instead to avoid repetition.

mono/utils/mono-threads-mach-helper.c
```diff
@@ -24,6 +24,12 @@
 */
 void mono_threads_init_dead_letter (void);
 void mono_threads_install_dead_letter (void);
+
+// mono_thread_info_detach is MONO_API elsewhere, so needs to at least be extern "C" here.
```
luhenry: Just `#include <glib.h>`

mono/metadata/icalls.h
```diff
@@ -0,0 +1,805 @@
...
+#include "utils/mono-proclib.h"
+#include "utils/mono-time.h"
+#include "w32event.h"
+#include "w32file.h"
+#include "w32mutex.h"
+#include "w32process.h"
+#include "w32semaphore.h"
+#include "w32socket.h"
+
+ICALL_EXPORT MonoReflectionAssemblyHandle
```
lambdageek: So there's now a declaration for every icall, right?

If so, why can't we just use the standard `#ifdef __cplusplus`/`export "C" {` mechanism (in its `G_BEGIN_DECLS`) for this whole header instead of mucking about trying to get `ICALL_EXPORT` past GCC's parser.

jaykrell: Yes we could do that. You lose typesafe linkage, but on these functions that is probably ok, or even required (the taking address in the table would be ambiguous and fail to compile).

jaykrell: (You'd have to cast to disambiguate, which is kinda horrible at first, not sure there is a good way.)

mono/utils/mono-publib.h
jaykrell: Yes I considered that but didn't want to add to the public symbol surface area. Ok. Will do.

jaykrell:
I discovered a wrinkle later.

gcc (but not clang) doesn’t like externC amidst attributes, only before them.

So I had to change `MONO_RT_EXTERNAL_ONLY MONO_API` to `MONO_API MONO_RT_EXTERNAL_ONLY`

https://github.com/mono/mono/pull/9812/commits/85c4527fa2a9c84c3dae34578da9030103be094c

mono/utils/mono-threads-mach-helper.c
jaykrell: Relucant, but  ok.

mono/metadata/icalls.h
jaykrell: Then ICALL_EXPORT needs to be extern "C", and have to fix all the missing ones.
Will try.

jaykrell:
https://github.com/mono/mono/pull/9812 is test bed for this change -- actually compiling as C++.

mono/metadata/icalls.h
jaykrell: I kinda still would prefer to minimize extern "C". It seems more "pure" to avoid it. But ok it is mostly done locally.

jaykrell:
There is some inconsistent formating in the `MONO_API MONO_RT_EXTERNAL_ONLY` stuff, as there was previously.

jaykrell:
Should be better now -- I filled in tons of ICALL_EXPORT. At least testing C++ on Mac.
I moved ICALL_EXPORT to glib.h rather than include class-internals more.
I suggest we don't shy from piling stuff into glib.h as perhaps our base-most internal header, after config.h.
We could still provide the missing prototypes.
Or, we could remove some other prototypes, perhaps.
We could still make ICALL_EXPORT be empty, if we provide the missing prototypes, which are in there in an earlier commit.

mono/utils/mono-threads-mach-helper.c
luhenry: Since it's MONO_API, why not simply putting `MONO_API void mono_thread_info_detach (void);`?

jaykrell: This part of the code is really cautious in what it includes but I can try adding publib/compiler if you insist. Recall you had to push me even to add glib.h and I was reluctant to do that.

jaykrell: And I agree what I have is dubious -- like, we declare two different ways, what is the result?

jaykrell:
Actually, adding `MONO_API` here on `mono_thread_detach` is a little wierd too.
I mean -- it is an extra unnecessary change, maybe?
Making it extern "C" is the point here. Changing any import/export is not.

mono/utils/mono-threads-mach-helper.c
```diff
@@ -24,7 +24,8 @@
 */
 void mono_threads_init_dead_letter (void);
 void mono_threads_install_dead_letter (void);
-void mono_thread_info_detach (void);
+MONO_API void
```
jaykrell: This may be too much unwarranted change, vs. just G_BEGIN_DECLS / G_END_DECLS.

jaykrell: Well..implementation is elsewhere, so I guess this isn't really a change in import/export.

jaykrell:
Not sure we prefer icall prototypes or not.

jaykrell:
So, the signatures could actually be icall-def.h, and the marshaling code could probably be statically generated via preprocessor macros. There are ancient techniques here. That should probably be considered/attempted, another day. They’d possibly ease making typesafe handles portable, and applying C++ to the resulting non-jit-interop handles. But probably for another day.

jaykrell:
This PR needs a new title.

jaykrell:

jaykrell:

jaykrell:
Good now? Awaiting a larger conversation?

jaykrell:
(want the icall prototypes back?)

luhenry:
It breaks the iOS SDK build though. It's worth investigating why.

jaykrell:
Good catch, thank you, will iterate, doesn't look like a big deal.
```
sers/builder/jenkins/workspace/test-mono-pull-request-amd64-osx-products-sdks-ios/mono/utils/mono-time.h:21:8: note: 'mono_msec_ticks' declared here
gint64 mono_msec_ticks (void);
       ^
/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx-products-sdks-ios/mono/utils/mono-time.h:43:13: warning: this function declaration is not a prototype [-Wstrict-prototypes]
        w->start = mono_100ns_ticks ();
                   ^
/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx-products-sdks-ios/mono/utils/mono-time.c:206:1: error: conflicting types for 'mono_100ns_ticks'
mono_100ns_ticks (void)
^
/Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx-products-sdks-ios/mono/utils/mono-time.h:43:13: note: previous implicit declaration is here
        w->start = mono_100ns_ticks ();
 ```

**Comment by jaykrell** [comment]
Maybe an old build from non-latest changes?

**Comment by jaykrell** [comment]

**Comment by jaykrell** [comment]
Still failed, that is wierd. I addressed this exact problem earlier. But it was a gcc thing, oh. I not sure I can/should commit this yet.

**Comment by jaykrell** [comment]
Oh.  MONO_API isn't defined, or is defined such as to not work out.

**Comment by jaykrell** [comment]
Obvious and mysterious at same time.
MONO_API isn't defined.
But where is ENABLE_ICALL_EXPORT getting defined?
Move/copy publib.h to glib.h?
Just add include of publib?

Well, first try local repro and look at config.h and try to figure it out more.
Don't want to just paper over a mystery.

**Comment by jaykrell** [comment]
I'm also probably going to push a nop commit to force a rebuild, still maybe for some reason running old stuff.

**Comment by jaykrell** [comment]

**Comment by jaykrell** [comment]
So, configure -enable-icall-export I probably broke. Not as much as might appear, as macros can be defined out of order, but still somewhat. I have to look at history/slack to see what it is/was for and if it is needed. Fixing it..move/copy more stuff to glib.h?

**Comment by jaykrell** [comment]
More of same. Will see what I can do tonight. Probably undo previous and include publib.h more. Leave icall_export in glib.h.

**Comment by directhex** [comment]
`21:57:00 
make[3]: *** No rule to make target 'icalls.h', needed by 'distdir'.  Stop.`

**Comment by jaykrell** [comment]
That is related to what I thought was an older version.

**Comment by jaykrell** [comment]

**Comment by directhex** [comment]
You added icalls.h to common_sources: https://github.com/jaykrell/mono/blob/1fe207e010afa1f6d9da7fe3e26b18c15b2cfee6/mono/metadata/Makefile.am#L267

But haven't added a file by that name (or a rule to create one), that I can see

File: mono/eglib/glib.h [hunk 5]
```diff
@@ -1145,6 +1145,20 @@ glong     g_utf8_pointer_to_offset (const gchar *str, const gchar *pos);
 #define G_HAVE_API_SUPPORT(x) (x)
 #define G_UNSUPPORTED_API "%s:%d: '%s' not supported.", __FILE__, __LINE__
 #define g_unsupported_api(name) G_STMT_START { g_warning (G_UNSUPPORTED_API, name); } G_STMT_END
+
+#ifdef ENABLE_ICALL_EXPORT
+#pragma GCC diagnostic ignored "-Wmissing-prototypes"
+#define ICALL_DECL_EXPORT MONO_API
```
luhenry: This will work with a strike of luck though, since you are effectively using a `#define` that you can only hope will be defined later, thus introducing a intrinsic dependency between `glib.h` and `mono-compiler.h` (where `MONO_API` is defined IIRC). Why moving these `ICALL_DECL_EXPORT` and `ICALL_EXPORT` to `glib.h` where it has nothing to do?

Defining it in `icall.h` which can then be included everywhere you need it will fix the issue at hand here.

jaykrell: It isn't quite luck, in that we check, so if it isn't defined later we fail (at least if configured correctly).

It was previously in class-internals.h. I wanted it somewhere where it'd be widely available, but not public, and not add much new exposure, i.e. not add more include class-internals.h.

glib.h is probably the most widely available non-public header. There is no `icall.h`. Do you mean `icalls.h` that I had earlier? Or introduce a new file icall.h that include publib please these lines?

jaykrell: I’ll add icall.h and a bunch of includes. Utils since it is needes there, unless we moving them all to metadata.

jaykrell:

mono/utils/mono-icall.h
```diff
@@ -0,0 +1,20 @@
+/**
+ * \file
+ */
+
+#ifndef __UTILS_MONO_ICALL_H__
+#define __UTILS_MONO_ICALL_H__
+
+#include <mono/utils/mono-publib.h>
+
+#ifdef ENABLE_ICALL_EXPORT
```
luhenry: Unfortunately, icalls have nothing to do in `utils/` since they are here to support managed code, and in `utils/` we tend to avoid things that support managed code directly. I understand you did that because some code in `utils/` depends on `ICALL_EXPORT`.

Move this file into `metadata/icalls.h`, and move the icalls that depends on `ICALL_EXPORT` to `metadata/`. Following are the only icalls I could identify that are in `utils/` and that needs to be moved to `metadata/`:
 - `ves_icall_System_Threading_Thread_SystemMaxStackSize`: move this function to `threads.c`/`thread-types.h` and create a function `mono_thread_info_get_system_max_stack_size` which is declared in `mono-threads.h` (in place of the icall) and implemented in `mono-threads-{posix,...}.c`
 - `mono_threads_platform_yield`: add a `ves_icall_System_Threading_Thread_Yield` function  in `threads.c`/`thread-types.h` that simply calls `mono_threads_platform_yield`.

jaykrell: I believe mini also has icalls, so utils is shared place for it.

jaykrell:
Is the layering, that is already violated, worth fixing? Worth anything? Does it help the system hold together soundly or is it arbitrary?

jaykrell:
Those are one question. If the layering helps the system be correct & maintainable, then it is worth fixing.

mono/utils/mono-icall.h
luhenry: `mini/` can depend on `metadata/` and `utils`, `metadata/` can depend on `utils/` only, `utils/` cannot depend on anything else. That's excluding other directories, but you get the idea. So `utils/` is not the shared place for that, and it has to be moved to `metadata/`

jaykrell: I was unsure of the mini vs. metadata layering. Great, that is easy then, as stated.

jaykrell:
This “in `utils/` we tend to avoid things that support managed code directly.” I wish was a little more defined. “Tend to” is gray, tend a little more here? Tend a little less there? Does it matter if it is only a “tendency” and not strict? Mono-threads contains sgen and cooohandle pieces for efficiency, for example. A problem, a mess, or perfectly ok?

luhenry:
Tends to as "let's not make the situation worst than it is". The case in this PR doesn't warrant the icall in `utils/` and fixing it is easy, so we fix it.

jaykrell:
There is another solution that @vargaz implied, recording it here in the open -- any header that contains an icall -- there are about 30 -- wrap the entire header in extern "C" (`G_BEGIN_DECLS` / `G_END_DECLS`). Instead of specific symbols.

I thought accurate per-function marking was kinda the intent so did that, but indeed it isn't really needed.

Longer term, the preciseness of extern C markings has various pros/cons, i.e. inhibit overloading or not, but short term it doesn't matter much.

jaykrell:

PR #9815, opened 2018-07-31
[cxx] various  fixes in mono/metadata

jaykrell:
casts between int and enum
casts from void* to non-void*
goto around initialization
remove designated initializers
fix typesafe coop handle creation


mono/metadata/class.c
```diff
@@ -424,8 +424,8 @@ mono_type_get_name_recurse (MonoType *type, GString *str, gboolean is_recursed,
 		}
 		const char *klass_name = m_class_get_name (klass);
 		if (format == MONO_TYPE_NAME_FORMAT_IL) {
-			char *s = strchr (klass_name, '`');
-			int len = s ? s - klass_name : strlen (klass_name);
+			const char *s = strchr (klass_name, '`');
+			gssize len = s ? (s - klass_name) : (gssize)strlen (klass_name);
```
jaykrell: Second line didn't need change. First line did.

mono/metadata/console-unix.c
```diff
@@ -218,13 +218,13 @@ tty_teardown (void)
...
 {
-	static MonoMethod *System_Console_DoConsoleCancelEventBackground_method = ((gpointer)-1);
+	static MonoMethod *System_Console_DoConsoleCancelEventBackground_method = (MonoMethod*)-1;
 	ERROR_DECL (error);
 
 	if (mono_defaults.console_class == NULL)
 		return;
 
-	if (System_Console_DoConsoleCancelEventBackground_method == ((gpointer)-1)) {
+	if (System_Console_DoConsoleCancelEventBackground_method == (gpointer)-1) {
```
jaykrell: Again, first needed change, second did not.

mono/metadata/icall.c
```diff
@@ -3576,7 +3586,7 @@ read_enum_value (const char *mem, int type)
 }
 
 static void
-write_enum_value (char *mem, int type, guint64 value)
+write_enum_value (void *mem, int type, guint64 value)
```
jaykrell: Safer way to  write this `void* mem_param` and then local `char* mem = (char*)mem_param`, i.e. in case any math or deref or pass off to other function occurs, but this should be ok.

mono/metadata/opcodes.h
```diff
@@ -20,21 +20,23 @@ MONO_BEGIN_DECLS
 
 #define MONO_CUSTOM_PREFIX 0xf0
 
+// FIXME duplication
```
jaykrell: Push into a new separate header? mono-opcodenum.h? Or alter includes? Or ok to duplicate? Temporarily?

mono/metadata/w32socket-unix.c
```diff
@@ -1143,7 +1144,7 @@ mono_w32socket_ioctl (SOCKET sock, gint32 command, gchar *input, gint inputlen,
 {
 	SocketHandle *sockethandle;
 	gint ret;
-	gchar *buffer;
+	gpointer buffer;
```
jaykrell: Perhaps not the safest change to make it work but should  be ok.

mono/metadata/opcodes.h
```diff
@@ -20,21 +20,23 @@ MONO_BEGIN_DECLS
...
+#ifndef MonoOpcodeEnum
+#define MonoOpcodeEnum MonoOpcodeEnum
+
 #define OPDEF(a,b,c,d,e,f,g,h,i,j) \
 	MONO_ ## a,
 
-#ifndef __MONO_METADATA_BASIC_BLOCK_H__
```
lambdageek: I don't understand what this was for - the comment in `mono-basic-block.h` says 
```c
/*This function is here because opcodes.h is a public header*/

#ifndef __MONO_METADATA_OPCODES_H__
typedef enum _MonoOpcodeEnum MonoOpcodeEnum;
#endif

int
mono_opcode_value_and_size (const unsigned char **ip, const unsigned char *end, MonoOpcodeEnum *value);
```

Which sounds like they just want to `#include <opcodes.h>` in `mono-basic-block.h` but couldn't for some reason.

lambdageek: Oh, but also: I don't understand why it's necessary to change this file for this PR.

jaykrell: 1. Cannot forward reference an enum, alas.
2. I think maybe no common header, or varying include order. This is the smaller but maybe not prettiest way. I can try adding  an include.
Could use ints more also, granted.

lambdageek: ah, no forward ref for enum makes sense.

mono/metadata/mono-basic-block.h
```diff
@@ -31,8 +31,21 @@ mono_basic_block_free (MonoSimpleBasicBlock *bb);
 
 /*This function is here because opcodes.h is a public header*/
```
lambdageek: This comment is a bit orphaned now.  I think it should go with `mono_opcode_value_and_size`, below.

jaykrell:
Resolved merge conflict with mono/master.

jaykrell:
@monojenkins build OS X x64

jaykrell:

jaykrell:
@monojenkins build OS X x64

PR #9816, opened 2018-07-31
In corlib updating documentation, improve Windows support that regres…

jaykrell:
…sed in commit 28dddea4dbd705557e8975f9cd56218afe87f36f.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,7 +41,8 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This must be unique relative to corlib interface/semantics.
 #
 # Generate it with uuidgen. For example:
-# perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=`uuidgen`/" configure.ac
+# Unix:    perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=`uuidgen`/" configure.ac
+# Windows: for /f %a in ('uuidgen') do perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=%a/" configure.ac
 #
 # There is no ordering of corlib versions, no old or new,
 # the runtime expects an exact match.


configure.ac
```diff
@@ -41,7 +41,8 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This must be unique relative to corlib interface/semantics.
 #
 # Generate it with uuidgen. For example:
-# perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=`uuidgen`/" configure.ac
+# Unix:    perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=`uuidgen`/" configure.ac
+# Windows: for /f %a in ('uuidgen') do perl -pi.bak -e "s/^MONO_CORLIB_VERSION=\S+/MONO_CORLIB_VERSION=%a/" configure.ac
```
akoeplinger: How about `sed -i "" "s/^MONO_CORLIB_VERSION=.*/MONO_CORLIB_VERSION=$(uuidgen)/" configure.ac` ?

That should work the same on Unix and Windows.

jaykrell: I'm assuming Perl is common enough, and cmd.
I wouldn't assume that of sed or bash.
This was another advantage of the C version imho: cl scripts/foo.c && foo.exe configure.ac.
Though, again, I think assuming Python or Perl on Windows is reasonable, more so than sed/bash. I guess csc or mcs similarly -- C:\WINDOWS\microsoft.net\Framework\v2.0.50727\csc.exe is pretty prevalent. I should write a few lines of C# some day..

akoeplinger: If you're building mono on Windows today you need `sed` anyway since it's used inside of configure.ac already :)

jaykrell: The msbuild stuff, reads configure.ac, but doesn't use autoconf/sed/make etc.
Granted..people revising this line, and working on Windows, probably zero.
As well, it is safe to just increment any of the digits manually. It is most unlikely to collide.
It is just that the 6 lines of Perl in a separate file would have worked asis on Windows. I don't understand what was so bad about it. Anyway I'll try to move on.

PR #9817, opened 2018-07-31
[cxx] make more of metadata valid C++, casts, const, goto around init, enum vs. int etc.

jaykrell:


jaykrell:

jaykrell:
Closing temporarily while casts are reworked.

PR #9818, opened 2018-07-31
[cxx] g_get_assertion_message returns const; do not write over it

jaykrell:
We could alternatively return non-const, but the overwrite seems gratuitious, assuming the length fits in 31 bits.


jaykrell:
Review?

PR #9821, opened 2018-07-31
[cxx] Port complex math to small portable C and C++.

jaykrell:
In file included from threadpool-worker-default.c:28:
../../mono/utils/mono-complex.h:67:8: warning: inline variables are a C++17 extension [-Wc++17-extensions]
static inline
       ^
../../mono/utils/mono-complex.h:68:15: error: expected ';' after top level declarator
double_complex mono_double_complex_make(gdouble re, gdouble im)
              ^
              ;


jaykrell:
1. Concerns about C++ runtime dependency here? That's why I held back on Android but welcome more facts/opinions. I'm kinda operating in FUD territory on this one, and that is mostly ok, avoid the dependency, but not as easy here.
2. Does the "old" code compile asis on Android as C++? I need to try building the SDKs.
3. Anyone want to rewrite it w/o complex/imaginary?

jaykrell:
Let's not commit this PR until more understood. It could take feedback/facts/opinions, but the present lack of facts is a bit unconvincing.

jaykrell:
Maybe the right thing here is just leave this code as C?
To avoid any libc++/libstdc++ dependency?
And of course ensuring it interfaces with the rest of the world via C?

jaykrell:
LGTM.

jaykrell:
LGTM. Anyone else?

mono/utils/mono-complex.h
```diff
@@ -60,38 +70,4 @@ double_complex mono_double_complex_sub(double_complex left, double_complex right
...
-}
-
-static inline
-double_complex mono_double_complex_sub(double_complex left, double_complex right)
-{
-	return left - right;
-}
-
-#endif
+#include "../support/libm/complex.c"
```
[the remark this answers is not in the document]
jaykrell: And separate file due to license. On one line of code. :(

mono/metadata/Makefile.am
```diff
@@ -64,10 +64,6 @@ endif
 
 glib_libs = $(top_builddir)/mono/eglib/libeglib.la
 
-if HOST_ANDROID
```
lambdageek: I don't understand why we're getting rid of something that we apparently needed for android - and the commit message doesn't really explain it.

jaykrell: Commit message says "portable", i.e. one thing for all platfroms.

There were three functions there: `creal, `cimag`, `cabs`.
`creal`, `cimag` are in mono-complex.h now, one line each, inlined, `cabs` is left in this file which is not deleted, but included for inlining. Another choice is no inlining and remove the if android and build the file for all platforms. But one line, should be inlined.

mono/utils/mono-complex.h
```diff
@@ -11,24 +11,34 @@
...
+	double_complex a = { re, im };
+	return a;
+}
+
+static inline double creal (double_complex c)
+{
+	return c.real;
+}
+
+static inline double cimag (double_complex c)
```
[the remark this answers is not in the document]
jaykrell: Could also be macros.

mono/metadata/Makefile.am
jaykrell: Is not clear enough?

lambdageek:
Okay so in summary:
* We will never use the C99 `complex double` type.
* We will always use the `struct mono_double_complex` defined in `utils/mono-complex.h`
* We will redefine `double_complex` (our own thing) and `cabs`, `creal`, `cimag` (C99 standard names) to use the mono versions whenever the `utils/mono-complex.h` header is included.
* Previously on Android we used to use an implementation of complex math that looks like it was based on some FreeBSD code.  Now Android is no different than any other platform and we'll just use the `mono_double_complex` struct.

Is that accurate?  If yes, I'm ready to approve.

mono/utils/mono-complex.h
```diff
@@ -11,24 +11,34 @@
...
-#endif
-
 #define _USE_MATH_DEFINES // needed by MSVC to define math constants
 #include <math.h>
 
-#ifdef _MSC_VER
+#define double_complex	mono_double_complex
+#define creal 		mono_double_complex_creal
+#define cimag 		mono_double_complex_cimag
+#define cabs 		mono_double_complex_cabs
```
lambdageek: We talked off-GitHub and @jaykrell will try to remove these #defines and update `threadpool-worker-default.c` to use `mono_double_complex_creal` etc.  Otherwise, LGTM.

jaykrell:
So much line damage now, hard to read, let me try again.

jaykrell:
I read some complex math lessons and they lined up with the preexisting msvc code but didn't check it all.

jaykrell:

jaykrell:

PR #9822, opened 2018-07-31
[cxx] threadpool-worker-default.c other than complex math: casts on f…

jaykrell:
…unctions and ints/enums


jaykrell:

PR #9824, opened 2018-07-31
Fix System.Drawing on 64-bit macOS.

filipnavara:
Carbon implementation now throws exception instead of crashing in the mismatched p/invoke calls.
Cocoa implementation now support 64-bit mode and uses the handle of `NSView` that is passed in instead of relying on `NSGraphicsContext.CurrentContext`. It also fixes handling of flipped views.

Requires up to date libgdiplus.


filipnavara:
@monojenkins build failed

PR #9826, opened 2018-07-31
Fix race condition where parallel builds will continue before we are actually done installing monolite

kg:
Right now we have a race condition where parallel make on CI will barge ahead and start building things before monolite+basic-profile-check have completed, which results in some weird behavior - gensources silently fails to run, etc. Testing some possible fixes for this.

### DIFF
--- a/mcs/build/profiles/basic.make
+++ b/mcs/build/profiles/basic.make
@@ -117,7 +117,6 @@ do-profile-check-monolite: $(depsdir)/.stamp
 endif
 
 $(PROFILE_EXE): $(topdir)/build/common/basic-profile-check.cs $(GENSOURCES_CS)
-	echo "Starting profile check and gensources build"
 	$(MAKE) $(MAKE_Q) -C $(topdir)/packages
 	$(BOOTSTRAP_MCS) /warn:0 /noconfig /langversion:latest /r:System.dll /r:mscorlib.dll /out:$@.tmp $<
 	$(BOOTSTRAP_MCS) /noconfig /langversion:latest /r:mscorlib.dll /r:System.dll /r:System.Core.dll /out:$(GENSOURCES_EXE).tmp $(GENSOURCES_CS)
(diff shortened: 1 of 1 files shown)


kg:

mcs/build/profiles/basic.make
```diff
@@ -118,8 +118,12 @@ endif
 
 $(PROFILE_EXE): $(topdir)/build/common/basic-profile-check.cs $(GENSOURCES_CS)
 	$(MAKE) $(MAKE_Q) -C $(topdir)/packages
-	$(BOOTSTRAP_MCS) /warn:0 /noconfig /langversion:latest /r:System.dll /r:mscorlib.dll /out:$@ $<
-	$(BOOTSTRAP_MCS) /noconfig /langversion:latest /r:mscorlib.dll /r:System.dll /r:System.Core.dll /out:$(GENSOURCES_EXE) $(GENSOURCES_CS)
+	$(BOOTSTRAP_MCS) /warn:0 /noconfig /langversion:latest /r:System.dll /r:mscorlib.dll /out:$@.tmp $<
+	$(BOOTSTRAP_MCS) /noconfig /langversion:latest /r:mscorlib.dll /r:System.dll /r:System.Core.dll /out:$(GENSOURCES_EXE).tmp $(GENSOURCES_CS)
+	- rm $(GENSOURCES_EXE)
+	mv $(GENSOURCES_EXE).tmp $(GENSOURCES_EXE)
+	- rm $@
```
marek-safar: Why not to use `mv -f` ?

kg: wasn't familiar with it, but I suspect it's not atomic so it would reintroduce the race (at a lower chance of happening)

marek-safar: Are you saying `mv -f` is not atomic but `rm + mv` is atomic?

kg: Only a single mv is atomic, so I'm not certain whether ```mv -f```'s observed behavior will do what I want. I can do more reading about it if you want but the manpage did not seem to make any guarantees - this isn't something googleable but maybe i can read through the source to gnu mv

marek-safar: It's not worth it, let's get this in

kg: I read through the source code and it looks like -f is just implemented as rm + mv so I'll switch to that.

PR #9829, opened 2018-07-31
Avoid all Roslyn dependencies in --with-csc=mcs builds

directhex:
A major driver for MCS_MODE is so distributions can do a 100% from-source build, distros will strip Roslyn from those, so we shouldn't need it there when not using it.


lewurm:
why was this closed?

directhex:
@marek-safar wasn't interested in the change set, on the basis that `--with-csc=mcs` builds are not intended to create a Roslyn-free distribution (just not use Roslyn during build)

directhex:
There are other relates changes I didn't put in this PR, on that basis, like making xbuild default to mcs/vbnc on mcs-mode builds,

PR #9831, opened 2018-07-31
Fix mcs mode aot

directhex:
We should avoid all dependency on Roslyn, when doing a --with-csc=mcs build, as a major use-case for this is Linux distributions which will need to strip out Roslyn for from-source builds.


marek-safar:
The intention for mcs mode was to use mcs to build Mono not to ship mcs only setup

PR #9840, opened 2018-08-01
[runtime] Cross32 changes.

vargaz:


vargaz:
First set of cross32 changes.

vargaz:
Note that the code already contains all the ifdefs, these changes just change some of them to be target specific.

configure.ac
```diff
@@ -4099,15 +4099,14 @@ if test "x$host" != "x$target"; then
    enable_mcs_build=no
    enable_support_build=no
    BTLS_SUPPORTED=no
+   # Can't use tls, since it depends on the runtime detection of tls offsets
+   # in mono-compiler.h
+   with_tls=pthread
```
filipnavara: Care to elaborate a bit on this? I don't see anything like that in `mono-compiler.h`. Moreover, I was convinced that the TLS offset detection was fully removed from the code base already.

vargaz:
@monojenkins Linux WebAssembly

vargaz:

vargaz:

configure.ac
filipnavara: Good point, I didn't notice you just moved it around. The part of code from mono-compiler.h was removed in PR #4014. I believe these lines are obsolete now, but I would like some confirmation before making a PR to remove them (or alternatively, to update the comment). @BrzVlad?

PR #9843, opened 2018-08-01
Truncate MachineName (and UserDomainName) at first period

### LINKED ISSUE #9839: Behaviour difference between Mono and CoreCLR
```
using System;

namespace envrinoment_test
{
    class Program
    {
        {
			Console.WriteLine(Environment.UserDomainName);
			Console.WriteLine(Environment.MachineName);
        }
    }
}
```

## Steps to Reproduce

1. Given the code above, run this under both dotnet and mono
2. Dotnet prints:
```
MININT-8D14Q3D
MININT-8D14Q3D
```
3. Mono prints:
```
MININT-8D14Q3D.redmond.corp.microsoft.com
MININT-8D14Q3D.redmond.corp.microsoft.com
```

They should be matching, not sure what full framework CLR prints

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 5.17.0.84 (pull-request-8951/ab1aafc1d80 Fri Jun 15 11:11:39 EDT 2018)

NattyNarwhal:
Unix machines often put their FQDN as the host name, and domain
name is used for things such as yp. As such, truncate the FQDN to
just the hostname, as other Unix utilities seem to do.

Matches the .NET Core and .NET Framework behaviour, and MS seems to
require this for a future project. Fixes #9839.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -6542,7 +6542,7 @@ mono_icall_get_machine_name (MonoError *error)
 #if !defined(DISABLE_SOCKETS)
 	MonoStringHandle result;
 	char *buf;
-	int n;
+	int n, i;
 #if defined _SC_HOST_NAME_MAX
 	n = sysconf (_SC_HOST_NAME_MAX);
 	if (n == -1)
@@ -6552,6 +6552,13 @@ mono_icall_get_machine_name (MonoError *error)
 	
 	if (gethostname (buf, n) == 0){
 		buf [n] = 0;
+		// try truncating the string at the first dot
+		for (i = 0; i < n; i++) {
+			if (buf[i] == '.') {
+				buf[i] = '\0';
+				break;
+			}
+		}
 		result = mono_string_new_handle (mono_domain_get (), buf, error);
 	} else
 		result = MONO_HANDLE_CAST (MonoString, NULL_HANDLE);


NattyNarwhal:
cc @Therzok

Therzok:
cc @jsuarezruiz

Therzok:
@monojenkins build pkg

Therzok:
Thanks for this! It works well!

Maybe an unit test could be added to ensure it doesn't contain dots?

NattyNarwhal:
Unit test added.

mcs/class/corlib/Test/System/EnvironmentTest.cs
```diff
@@ -231,6 +231,13 @@ public void SetEnvironmentVariable ()
 			Environment.SetEnvironmentVariable ("A3", "\0");
 			Assert.IsNull (Environment.GetEnvironmentVariables ()["A3"]);
 		}
+
+		[Test] // github issue #9839
+		public void MachineNameIsNotFullyQualifiedDomainName()
+		{
+			Assert.IsNotNull (Environment.MachineName);
+			Assert.AreEqual(Environment.MachineName.IndexOf("."), -1);
```
Therzok: Nit: missing spaces before parens here and in the method definition.

Nit: the expected parameter (-1) should be first.

akoeplinger:
Looks good to me but I'm wondering about the compatibility impact of this, is there a legitimate scenario where this could break users?

jaykrell:
Future project?

Therzok:
@jaykrell VSMac TFS support

NattyNarwhal:
But that function doesn't take length?

Therzok:
@jsuarezruiz 15.8?

akoeplinger:
@Therzok @jsuarezruiz any update? this will be in the upcoming 2018-10 branch (Mono 5.20) right now.

jsuarezruiz:
15.8 will be nice.

akoeplinger:

akoeplinger:

slaingod:
For anyone else who got screwed over by this change (amazon linux instances tend to have hostnames that contain periods, etc) :
csharp> System.Net.Dns.GetHostName()                          
"vm.ctos.dev.01"                        
seems to be a reasonable alternative (TBD if it performs similarly or needs to be cached). Now to go change Environment.MachineName in a few hundred places in code...
Oh, right I can't really, I have no doubt that plenty of nuget packages use Environment.MachineName and haven't tested this scenario. Let's see if we can fix this with reflection or a hook. 

Can we add an AppDomain setting for this?

Feels like the plan should have been to try and get netCore to change to allow perfectly legit hostnames in linux, rather than what was done here.

NattyNarwhal:
I actually agree here that the Core/Framework version is short-signed, even if this commit aims to make it more compatible. It (seems to me) makes more sense on Windows where FQDN and domain are likely to be joined at the hip via AD, but less so on Unix systems.

slaingod:
Well, I will shut up on this a little. I was actually able to convince my managers to fix it on their end with clients, and that the RFC said hostnames with dots meant FQDN, period. Once I explained the nuget package minefield and the thousands of the places in the code that directly or indirectly relied on machine name matching the hostname. Sorry for the noise!

PR #9846, opened 2018-08-01
[msbuild] Bump to point to HEAD of mono-2018-04 msbuild  …

radical:
.. once, msbuild has an updated branch for 2018-06/master, then this
will point to that.

This also picks up fix for https://github.com/xamarin/xamarin-android/issues/1845 .


radical:

radical:
All the msbuild tests passed. We didn't get the results files though, but that should be unrelated to this PR.

radical:
hm that will be fixed with the upcoming big master update.

PR #9859, opened 2018-08-02
Ensure cil-stringreplacer is built and used to post-process mscorlib when building using msbuild

kg:
A recent change caused msbuild/visual studio builds to break due to a name collision in mscorlib. On linux this name collision is harmless because cil-stringreplacer erases the name collision by renaming the type. Because cil-stringreplacer was entirely missing from our msbuild support, the build would fail. This PR adds a manually constructed csproj file for cil-stringreplacer, then ensures that it builds and is used to post-process mscorlib.dll.

Part of #6886


kg:
@monojenkins build failed

marek-safar:
@akoeplinger could we enable the VS lane for every PR (as optional) once this lands

kg:
@monojenkins commit csproj

Issue #6949, opened 2018-02-12
Mono fails to build on Solaris.


jaykrell:
Mono fails to build on Solaris (x86).
The problems are not complicated from what I've seen.

To get recurse pthread mutexes, you need -D_XPG6.
-std=c99 is somehow missed and that is needed with -D_XPG6, so add that in a solaris section of configure.ac. (It could be that -std=c99 is enough actually).
There is some problem with dtrace and the code not being PIC-enough. Could just disable dtrace on Solaris but I'm sure the right fix isn't difficult either.

vargaz:
We don't support solaris, its community supported.

sourceskyboxer:
Wait @vargaz "We don't support solaris, its community supported."

Why do you say that? But hello check headers or sources have "(xxxxxxx)_solaris.h" after starting "make". I saw that. Please do not lie me! I have caught that.

Thanks Please support for Solaris OS now! Many users are arguing.... Same FreeBSD too

sourceskyboxer:
Check @vargaz You're liar! Check [website](http://www.mono-project.com/docs/advanced/runtime/)!

Thanks! It looks like official.

jaykrell:
Look at Boehm here too. Modern boehm just uses pthreads. Our uses Solaris threads or something ancient and irrelevant, I think.

PR #9860, opened 2018-08-02
Use argv for cli_launcher on OSes that don't return full path for process names; fix mono_w32process_get_path on Free/NetBSD

NattyNarwhal:
Should fix process-leak test on AIX, maybe more. Also tested that it fixes the
same behaviour on FreeBSD. Other OSes may be affected; consult and test with the
files `w32process-unix-*.c`.

From comments:

```c
/*
 * XXX: verify if other OSes need it; many platforms seem to have it so that
 * mono_w32process_get_path -> mono_w32process_get_name, and the name is not
 * necessarily a path instead of just the program name
 */

/*
 * mono_w32process_get_path on these can only return a name, not a path;
 * which may not be good for us if the mono command name isn't on $PATH,
 * like in CI scenarios. chances are argv based is fine if we inherited
 * the environment variables.
 */
```

### DIFF
--- a/mono/metadata/w32process-unix-bsd.c
+++ b/mono/metadata/w32process-unix-bsd.c
@@ -99,17 +99,18 @@ mono_w32process_get_path (pid_t pid)
 #else
 	gsize path_len = PATH_MAX + 1;
 	gchar path [PATH_MAX + 1];
-	gint mib [3];
+	gint mib [4];
+	mib [0] = CTL_KERN;
 #if defined (__NetBSD__)
-	mib [0] = KERN_PROC_ARGS;
-	mib [1] = -1
-	mib [2] = KERN_PROC_PATHNAME;
+	mib [1] = KERN_PROC_ARGS;
+	mib [2] = pid;
+	mib [3] = KERN_PROC_PATHNAME;
 #else // FreeBSD
-	mib [0] = KERN_PROC;
-	mib [1] = KERN_PROC_PATHNAME;
-	mib [2] = -1;
+	mib [1] = KERN_PROC;
+	mib [2] = KERN_PROC_PATHNAME;
+	mib [3] = pid;
 #endif
-	if (sysctl (mib, 3, path, &path_len, NULL, 0) < 0) {
+	if (sysctl (mib, 4, path, &path_len, NULL, 0) < 0) {
 		mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_PROCESS, "%s: sysctl() failed: %d", __func__, errno);
 		return NULL;
 	} else {
--- a/mono/metadata/w32process-unix-bsd.c
+++ b/mono/metadata/w32process-unix-bsd.c
@@ -93,7 +93,29 @@ retry:
 gchar*
 mono_w32process_get_path (pid_t pid)
 {
+#if defined (__OpenBSD__)
+	// No KERN_PROC_PATHNAME on OpenBSD
 	return mono_w32process_get_name (pid);
+#else
+	gsize path_len = PATH_MAX + 1;
+	gchar path [PATH_MAX + 1];
+	gint mib [3];
+#if defined (__NetBSD__)
+	mib [0] = KERN_PROC_ARGS;
+	mib [1] = -1
+	mib [2] = KERN_PROC_PATHNAME;
+#else // FreeBSD
+	mib [0] = KERN_PROC;
+	mib [1] = KERN_PROC_PATHNAME;
+	mib [2] = -1;
+#endif
+	if (sysctl (mib, 3, path, &path_len, NULL, 0) < 0) {
+		mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_PROCESS, "%s: sysctl() failed: %d", __func__, errno);
+		return NULL;
+	} else {
+		return g_strdup (path);
+	}
+#endif
 }
 
 static gint
--- a/mono/mini/driver.c
+++ b/mono/mini/driver.c
@@ -2231,7 +2231,20 @@ mono_main (int argc, char* argv[])
 		return 1;
 	}
 
-#if !defined(HOST_WIN32) && defined(HAVE_UNISTD_H)
+/*
+ * XXX: verify if other OSes need it; many platforms seem to have it so that
+ * mono_w32process_get_path -> mono_w32process_get_name, and the name is not
+ * necessarily a path instead of just the program name
+ */
+#if defined (_AIX) || defined (__FreeBSD__)
+	/*
+	 * mono_w32process_get_path on these can only return a name, not a path;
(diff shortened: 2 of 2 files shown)


jbeich:
FWIW, [FreeBSD](https://github.com/freebsd/freebsd/commit/5aec008257f77f80c182f1f37a6086260fb53a46/sys/kern/kern_proc.c), [DragonFly](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/f849311b84a9d80fed3dcd8b0ed2a1c9bd46baf9), [NetBSD](https://github.com/NetBSD/src/commit/6526421a92938d1252abdc8710c2e0b3928d6380) support `KERN_PROC_PATHNAME` ([example](https://github.com/freebsd/freebsd-ports/blob/770f0abb35f8/devel/android-tools-adb/files/patch-base_file.cpp)) which returns full path.

NattyNarwhal:
Seems fixing that could be done in `w32process-unix-bsd.c` then, and remove that ifdef for FreeBSD. I'll poke at it.

mono/metadata/w32process-unix-bsd.c
```diff
@@ -93,7 +93,29 @@ mono_w32process_get_name (pid_t pid)
...
+#if defined (__OpenBSD__)
+	// No KERN_PROC_PATHNAME on OpenBSD
 	return mono_w32process_get_name (pid);
+#else
+	gsize path_len = PATH_MAX + 1;
+	gchar path [PATH_MAX + 1];
+	gint mib [3];
+#if defined (__NetBSD__)
+	mib [0] = KERN_PROC_ARGS;
+	mib [1] = -1
```
jbeich: `-1` is for self (saving `getpid()` call), use `pid` argument here...

mono/metadata/w32process-unix-bsd.c
```diff
@@ -93,7 +93,29 @@ mono_w32process_get_name (pid_t pid)
...
+	mib [2] = KERN_PROC_PATHNAME;
+#else // FreeBSD
+	mib [0] = KERN_PROC;
+	mib [1] = KERN_PROC_PATHNAME;
+	mib [2] = -1;
```
jbeich: ...and here.

mono/metadata/w32process-unix-bsd.c
```diff
@@ -93,7 +93,29 @@ mono_w32process_get_name (pid_t pid)
...
 {
+#if defined (__OpenBSD__)
+	// No KERN_PROC_PATHNAME on OpenBSD
 	return mono_w32process_get_name (pid);
+#else
+	gsize path_len = PATH_MAX + 1;
+	gchar path [PATH_MAX + 1];
+	gint mib [3];
+#if defined (__NetBSD__)
+	mib [0] = KERN_PROC_ARGS;
```
jbeich: `mib [0]` should be `CTL_KERN` for both NetBSD and FreeBSD/DragonFly.

NattyNarwhal: Does that need to bump the MIB further?

NattyNarwhal: like?:

```c
        gint mib [4];
        mib [0] = CTL_KERN;
#if defined (__NetBSD__)
        mib [1] = KERN_PROC_ARGS;
        mib [2] = pid;
        mib [3] = KERN_PROC_PATHNAME;
#else // FreeBSD
        mib [1] = KERN_PROC;
        mib [2] = KERN_PROC_PATHNAME;
        mib [3] = pid;
#endif
        if (sysctl (mib, 4, path, &path_len, NULL, 0) < 0) {
```

mono/mini/driver.c
```diff
@@ -2231,7 +2231,20 @@ mono_main (int argc, char* argv[])
 		return 1;
 	}
 
-#if !defined(HOST_WIN32) && defined(HAVE_UNISTD_H)
+/*
+ * XXX: verify if other OSes need it; many platforms seem to have it so that
+ * mono_w32process_get_path -> mono_w32process_get_name, and the name is not
+ * necessarily a path instead of just the program name
+ */
+#if defined (_AIX)
```
jbeich: On Solaris `/proc/%d/path/a.out` returns full path. AIX probably uses`/proc/%d/object/a.out` instead (like old Solaris versions).

NattyNarwhal: they do have the file itself, but not its path, and it's not any kind of link either:

```
aix$ ls -l /proc/7733404/object/a.out
-rwxr-xr-x 1 root system 810463 Jun  5 08:05 /proc/7733404/object/a.out
aix$ file /proc/7733404/object/a.out
/proc/7733404/object/a.out: executable (RISC System/6000) or object module not stripped
```

and Solaris, for good measure:

```
calvin@hako:~$ ls -l /proc/7829/object/a.out
-r-xr-xr-x   1 root     bin      3236820 Mar 24 09:16 /proc/7829/object/a.out
calvin@hako:~$ file /proc/7829/object/a.out
/proc/7829/object/a.out:        ELF 32-bit LSB executable 80386 Version 1 [FPU], dynamically linked, not stripped
```

(FWIW, on AIX, the internal kernel structures for processes keep track of inode, for presumably what procfs is doing, but it doesn't expose that in userland APIs beyond what procfs is doing, though kernel modules can peer into it....)

granted, you could still launch from that executable in procfs, but it'd seem dirty IMHO; if I did do that though, i'll change this def to PASE and OpenBSD (plus maybe Haiku/Linux/non-x86 Darwin, since those backends do the same thing for path/name?)

jaykrell: Our build is quite broken on Solaris, I tried recently..

NattyNarwhal: if I can get this admin of the Solaris system I'm using to install Sun Studio (and/or GCC) whenever he gets the time, I'll see what I can do WRT that....

jaykrell: It is easy to try Solaris/x86 and Solaris/amd64 in VMware, etc.
I have access to SPARC systems, but..it isn't a priority.

jaykrell: https://github.com/mono/mono/issues/6949

NattyNarwhal: looks like Linux has congruent path == name where name is equiv to the full path, but only if the first two attempts at reading procfs succeed, otherwise you just get the name (not 

so in that case, the OSes that would suffer without this are AIX/PASE, OpenBSD, Solaris, Haiku, and non-x86 Darwin; and I've only tested AIX and fixed FreeBSD (as well as NetBSD), so... (i'll consider testing Solaris whenever the guy gets back to me; Haiku depends on me getting it in a less broken state, etc.; so the others can be added when/if the need arises for them, especially if proper path fetching could be added)

lambdageek:

lewurm:
@monojenkins squash

krytarowski:
NetBSD part looks fine.

PR #9862, opened 2018-08-02
[runtime] Cache culture locale name in managed

alexanderkyte:
Every different locale getter in unmanaged attempts to
cache the result from this icall. Some of them fail to do so in a manner
that is correct.

CFLocaleCopyCurrent (); in get_darwin_locale ends up being called
thousands of times. In an app with a few dozen Console.WriteLines, it's easy to see
18mb of leaked memory through this.

Rather than trying to fix them all in unmanaged in a perfect manner, I
thought it better to also avoid the icall overhead and to cache it in
C#.

Without this fix, a memory traffic stress test (debian shootout's pidigits) leads to 66% of
memory pressure being due to this  icall. With this fix, it is 4.3%.

If this caching is wrong for correctness reasons (we expect it to change?),
I would like to point out that we have similar caching done in unmanaged
and that we should then bring both into a correct design.

### DIFF
--- a/mcs/class/corlib/System.Globalization/CultureInfo.cs
+++ b/mcs/class/corlib/System.Globalization/CultureInfo.cs
@@ -146,17 +146,26 @@ namespace System.Globalization
 			}
 		}
 
+		private static string current_culture_name;
+
 		internal static CultureInfo ConstructCurrentCulture ()
 		{
 			if (default_current_culture != null)
 				return default_current_culture;
 
-			var locale_name = get_current_locale_name ();
+			string locale_name;
+			if (current_culture_name != null)
+				locale_name = current_culture_name;
+			else
+				locale_name = get_current_locale_name ();
+
 			CultureInfo ci = null;
 
 			if (locale_name != null) {
 				try {
 					ci = CreateSpecificCulture (locale_name);
+					if (current_culture_name == null)
+						current_culture_name = locale_name;
 				} catch {
 				}
 			}


mcs/class/corlib/System.Globalization/CultureInfo.cs
```diff
@@ -146,17 +146,26 @@ struct Data {
...
+			if (current_culture_name != null)
+				locale_name = current_culture_name;
+			else
+				locale_name = get_current_locale_name ();
+
 			CultureInfo ci = null;
 
 			if (locale_name != null) {
 				try {
 					ci = CreateSpecificCulture (locale_name);
```
Therzok: we're always reconstructing `ci` even though current_culture_name is set?

alexanderkyte: Storing a single CultureInfo left me with the choice of making it readonly or not, 
https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.readonly(v=vs.110).aspx

If I didn't, then I was concerned that there was some way to abuse this singleton we return. If I did make it readonly, that might break existing code that depends on it not being readonly. 

I picked a conservative choice. If you have assurances that it's safe to pick one of the above policies, I'd love to make a singleton.

alexanderkyte: I do not know why we clear the cached objects while leaving the cached strings in unmanaged. I'm not sure what that accomplishes.

akoeplinger: I'm confused about two things:

1. Why is `default_current_culture` getting cleared? I can't find any callers of `ClearCachedData()`.
2. Why is `ConstructCurrentCulture()` even being called more than once. The only caller is `nativeInitCultureAccessors` and that in turn shouldn't be called more than once per thread.

PR #9877, opened 2018-08-03
Remove pragma GCC diagnostic ignored "-Wmissing-prototypes".

jaykrell:
1. I just expanded its reach.
2. It doesn't seem needed anywhere. Where it was intended to matter, we generate incorrect prototypes. I did some testing on Linux with configure export-icall.


jaykrell:

jaykrell:
I'm seeing the warnings now, granted, with other changes. Somewhat silly warning imho anyway, should only occur on callers.

PR #9880, opened 2018-08-04
[cxx] Prefer gpointer/gconstpointer over char*, guint8* for untyped data -- fewer casts generally.

jaykrell:
And more C++ compatible -- conversion to gpointer/gconstpointer is generally already valid, no need to cast, whereas converting to char*, guint8* often requires casts.

Sometimes it is inconvenient because you want to do pointer math.


jaykrell:
Close but doesn’t build yet.

jaykrell:
Didn't mean to close but ok.

jaykrell:
Replaced by https://github.com/mono/mono/pull/9881.

PR #9881, opened 2018-08-04
[cxx] Prefer gpointer/gconstpointer over char*, guint8* for untyped data -- fewer casts generally.

jaykrell:
This  tends to be more C++ compatible, and really is in this case.

### DIFF
--- a/mono/metadata/dynamic-image-internals.h
+++ b/mono/metadata/dynamic-image-internals.h
@@ -46,7 +46,7 @@ MonoDynamicImage*
 mono_dynamic_image_create (MonoDynamicAssembly *assembly, char *assembly_name, char *module_name);
 
 guint32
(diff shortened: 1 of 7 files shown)


jaykrell:
Requires some non-local analysis to check that we don't depend on sizeof(*p). Too expensive?

jaykrell:
Ok?

jaykrell:
This makes C++ port less messy but I workaround it with more casts.

jaykrell:
This isn't required for C++ conversion, but helps a little, and seems good anyway.
Otherwise probably more casts.

jaykrell:
Review?

jaykrell:
Review?

jaykrell:
This probably redundant now, let's wait on the other.

jaykrell:
Likely redundant with https://github.com/mono/mono/pull/10211
wlll rebase to master once it it is on.

PR #9885, opened 2018-08-05
Fix overallocate by factor of four -- this was len * sizeof(guint8*) instead of len * sizeof(guint8), which devolves here to just alloca(len).

jaykrell:


jaykrell:
@monojenkins build OS X i386

mono/mini/mini-x86.c
```diff
@@ -597,7 +597,7 @@ mono_arch_get_argument_info (MonoMethodSignature *csig, int param_count, MonoJit
 
 	/* Avoid g_malloc as it is not signal safe */
 	len = sizeof (CallInfo) + (sizeof (ArgInfo) * (csig->param_count + 1));
-	cinfo = (CallInfo*)g_newa (guint8*, len);
+	cinfo = (CallInfo*)alloca (len);
 	memset (cinfo, 0, len);
```
vargaz: This should use g_alloca ().

jaykrell:
Oh `g_alloca`? I read `g_newa`. Let's see. `g_newa` uses `alloca` directly though.

jaykrell:
`g_alloca` is just `alloca`. What's the point?
`#define g_alloca(size)		alloca (size)`

jaykrell:
We use alloca directly roughly twice as often as g_alloca.
```
cd mono
$ git grep g_alloca . | wc -l
      24
$ git grep \)alloca . | wc -l
      43
```

jaykrell:
Better? I submitted a separate PR to change many `alloca` to `g_newa`.

jaykrell:

PR #9886, opened 2018-08-05
[cxx] Fix enum/int mixup.

jaykrell:


mono/utils/mono-threads-coop.h
```diff
@@ -53,14 +53,12 @@ mono_threads_safepoint (void)
 		mono_threads_state_poll ();
 }
 
-/* -1 and 0 also used:
- * -1 means uninitialized
- * 0 means unset
- */
 typedef enum {
+	MONO_THREADS_SUSPEND_UNINITIALIZED   = -1,
```
jaykrell: Ideally uninitialized would be 0, but code depends on otherwise so leave it alone.

mono/utils/mono-threads-coop.c
```diff
@@ -574,13 +574,13 @@ threads_suspend_policy_getenv (void)
 	return policy;
 }
 
-static MonoThreadsSuspendPolicy threads_suspend_policy = -1;
+static MonoThreadsSuspendPolicy threads_suspend_policy = MONO_THREADS_SUSPEND_UNINITIALIZED;
```
jaykrell: Ideally this would not be initialized -- 0 would be correct, but that doesn't work currently for understood reasons.

jaykrell:
The second part I believe I kept correct but I see what you mean about public/private.

mono/utils/mono-threads-coop.c
```diff
@@ -574,13 +573,13 @@ threads_suspend_policy_getenv (void)
 	return policy;
 }
 
-static MonoThreadsSuspendPolicy threads_suspend_policy = -1;
+static MonoThreadsSuspendPolicy threads_suspend_policy = (MonoThreadsSuspendPolicy)-1;
```
lambdageek: What if we declare `static int threads_suspend_policy = -1;` and only cast to `MonoThreadsSuspendPolicy` on the return from `mono_threads_suspend_policy` once we know we definitely have a valid value?

jaykrell: Yeah you could switch the types the other way around, I didn't think of that.
I still like 0 for uninitialized if you are entertaining other change. It takes no room in the executable where non-zero takes room.

lambdageek:
Yea I think using 0 instead of -1 everywhere would be fine.  I was making a distinction between unset and uninitialized that's doesn't seem necessary.  If the global is set to zero we probe the various functions and whichever one returns non-zero that's the policy to use - at which point the global gets set to a valid enum value.

Thanks for helping to clean this up!

jaykrell:
The two values are ok, I just want the global to be initialized to zero, on principle, it is a cheap file size optimization, as “bss” is not stored in the file.

jaykrell:
I think I see what you mean though, one value probably suffice. I can try that.

mono/utils/mono-threads-coop.c
```diff
@@ -508,18 +508,16 @@ mono_threads_assert_gc_unsafe_region (void)
...
 threads_suspend_policy_default (void)
 {
 #if defined (ENABLE_COOP_SUSPEND)
 	return MONO_THREADS_SUSPEND_FULL_COOP;
-#else
-#if defined (ENABLE_HYBRID_SUSPEND)
+#elif defined (ENABLE_HYBRID_SUSPEND)
 	return MONO_THREADS_SUSPEND_HYBRID;
-#else
+#elif
```
lambdageek: this seems like an unintended change. does `#elif` without a condition do anything?

jaykrell: oops

mono/utils/mono-threads-coop.c
```diff
@@ -574,34 +572,26 @@ threads_suspend_policy_getenv (void)
...
-			policy = env_compat_policy;
-		else
-			policy = MONO_THREADS_SUSPEND_FULL_PREEMPTIVE;
-		
-		g_assert (policy > 0);
-		threads_suspend_policy = policy;
+		   (policy = threads_suspend_policy_getenv ())
+		|| (policy = threads_suspend_policy_default ())
+		|| (policy = threads_suspend_policy_getenv_compat ())
+		|| (policy = MONO_THREADS_SUSPEND_FULL_PREEMPTIVE);
```
lambdageek: I like that this tries the assortment of `thread_suspend_policy_*` functions one at a time.
Using short-circuiting `||` instead of `if` is a little unusual for Mono, but alright.

I think in Mono we usually we put the binary operator trailing on the line above:
```c
  a ||
  b ||
  c;
```
but I don't see anything about that in the coding guidelines.

jaykrell: Both exist in the wider world. || on left is advertised as more obviously a continuation line. || on the right is popular, and lines up the first line on the left w/o extra artificial indent. clang-tidy for  the win some day...

PR #9888, opened 2018-08-05
[cxx] Strengthen some types away from gpointer.

jaykrell:


jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

mono/mini/mini-amd64.h
```diff
@@ -210,16 +210,16 @@ typedef struct MonoCompileArch {
...
+	CallInfo *cinfo;
 	gint32 async_point_count;
-	gpointer vret_addr_loc;
+	MonoInst *vret_addr_loc;
+	MonoInst *seq_point_info_var;
+	MonoInst *ss_tramp_var;
+	MonoInst *bp_tramp_var;
+	MonoInst *lmf_var;
 #ifdef HOST_WIN32
 	struct _UNWIND_INFO* unwindinfo;
```
jaykrell: Also moved the "variable" part to the end, so more things are more the same, i.e. the offsets of the common parts are the same on all platforms.

jaykrell:

jaykrell:

jaykrell:
Review?

jaykrell:
Review?

jaykrell:
We could keep/add casts instead but this is *definitely* the way to go. It cleans up the code & makes it easier to understand & maintain & debug.

jaykrell:
Review?

mono/mini/mini-amd64.c
```diff
@@ -1443,7 +1443,7 @@ mono_arch_compute_omit_fp (MonoCompile *cfg)
 
 	if (!cfg->arch.cinfo)
 		cfg->arch.cinfo = get_call_info (cfg->mempool, sig);
-	cinfo = (CallInfo *)cfg->arch.cinfo;
+	cinfo = cfg->arch.cinfo;
```
vargaz: If we forward declare CallInfo, then why are these casts needed ?

jaykrell: The intent (did I mess up?) is to remove casts, not add them.
And then to make it more valid C++ in the process -- but better C either way imho.

vargaz: Misread it.

jaykrell: Oh -- perhaps, you mean -- we could cut out local variables in the process? Yes we probably could.

PR #9894, opened 2018-08-06
Replace many uses of alloca with g_newa.

jaykrell:


jaykrell:
Triggered by this, which suggests `g_alloca` over `alloca`, though we use `alloca` a lot.

https://github.com/mono/mono/pull/9885#issuecomment-410584742

jaykrell:
@monojenkins build  OS X i386

jaykrell:
@monojenkins build OS X i386

PR #9896, opened 2018-08-06
Remove gratuitous ifdef and variable-length array.

jaykrell:
We might as well just have one more portable form here.


jaykrell:
I surveyed `alloca` uses due to https://github.com/mono/mono/pull/9885#discussion_r207768933.

jaykrell:

PR #9911, opened 2018-08-06
[runtime] Speed-up MERP timeout

alexanderkyte:
Now that https://github.com/mono/mono/pull/9809 is merged so this won't dereference freed objects, I added a speed increase of the timeout. 

Making a test branch first to get a package to make sure this fix works. Don't merge until I've confirmed so.


alexanderkyte:
@monojenkins build Failed

PR #9914, opened 2018-08-06
[runtime] Add a reference queue cleanup for MonoInternalThread

alexanderkyte:
Alternate method of fixing: 
https://github.com/mono/mono/pull/9486

### DIFF
--- a/mono/metadata/gc.c
+++ b/mono/metadata/gc.c
@@ -46,6 +46,7 @@
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
 #include <mono/utils/mono-os-wait.h>
+#include <mono/utils/mono-lazy-init.h>
 
 #ifndef HOST_WIN32
 #include <pthread.h>
@@ -70,6 +71,7 @@ gchar **mono_do_not_finalize_class_names ;
 #define mono_finalizer_unlock() mono_coop_mutex_unlock (&finalizer_mutex)
 static MonoCoopMutex finalizer_mutex;
 static MonoCoopMutex reference_queue_mutex;
+static mono_lazy_init_t reference_queue_mutex_inited = MONO_LAZY_INIT_STATUS_NOT_INITIALIZED;
 
 static GSList *domains_to_finalize;
 
@@ -943,11 +945,17 @@ mono_gc_init_finalizer_thread (void)
 	mono_error_assert_ok (error);
 }
 
+static void
+reference_queue_mutex_init (void)
+{
+	mono_coop_mutex_init_recursive (&reference_queue_mutex);
+}
+
 void
 mono_gc_init (void)
 {
+	mono_lazy_initialize (&reference_queue_mutex_inited, reference_queue_mutex_init);
 	mono_coop_mutex_init_recursive (&finalizer_mutex);
-	mono_coop_mutex_init_recursive (&reference_queue_mutex);
 
 	mono_counters_register ("Minor GC collections", MONO_COUNTER_GC | MONO_COUNTER_INT, &mono_gc_stats.minor_gc_count);
 	mono_counters_register ("Major GC collections", MONO_COUNTER_GC | MONO_COUNTER_INT, &mono_gc_stats.major_gc_count);
@@ -1212,6 +1220,7 @@ mono_gc_reference_queue_new (mono_reference_queue_callback callback)
 	MonoReferenceQueue *res = g_new0 (MonoReferenceQueue, 1);
 	res->callback = callback;
 
+	mono_lazy_initialize (&reference_queue_mutex_inited, reference_queue_mutex_init);
 	mono_coop_mutex_lock (&reference_queue_mutex);
 	res->next = ref_queues;
 	ref_queues = res;
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -148,6 +148,9 @@ static GHashTable *contexts = NULL;
 /* Cleanup queue for contexts. */
 static MonoReferenceQueue *context_queue;
 
+/* Cleanup queue for threads. */
+static MonoReferenceQueue *thread_queue;
+
 /*
  * Threads which are starting up and they are not in the 'threads' hash yet.
  * When mono_thread_attach_internal is called for a thread, it will be removed from this hash table.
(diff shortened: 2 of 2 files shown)


mono/metadata/threads.c
```diff
@@ -453,14 +456,30 @@ thread_get_tid (MonoInternalThread *thread)
...
+ensure_synch_cs_set (MonoInternalThread *thread)
 {
 	MonoCoopMutex *synch_cs;
 
 	if (thread->synch_cs != NULL) {
 		return;
 	}
 
+	if (!thread_queue) {
+		gpointer temp_queue = mono_gc_reference_queue_new (free_synch_cs);
```
kumpera: It's fine to allocate this queue on startup.

kumpera: alternatively, use mono_lazy.

alexanderkyte:
It was breaking because the thread startup code runs before `mono_gc_init`

mono/metadata/gc.c
```diff
@@ -70,6 +70,7 @@ gchar **mono_do_not_finalize_class_names ;
 #define mono_finalizer_unlock() mono_coop_mutex_unlock (&finalizer_mutex)
 static MonoCoopMutex finalizer_mutex;
 static MonoCoopMutex reference_queue_mutex;
+static gint32 reference_queue_mutex_inited = 0x0;
```
jaykrell: No need to init.

mono/metadata/gc.c
```diff
@@ -943,11 +944,20 @@ mono_gc_init_finalizer_thread (void)
 	mono_error_assert_ok (error);
 }
 
+void
+reference_queue_mutex_init (void)
+{
+	if (mono_atomic_cas_i32 ((gint32 *)&reference_queue_mutex_inited, 0x1, 0x0) == 0x0) {
+		// If we are the initializer
```
jaykrell: No need for `0x`.

jaykrell: Why the cast? The type is already `gint32`.

jaykrell: What happens to threads that don't get the cas, they just fall through and use it? Please use `pthread_once`, err, the mono wrapper of it (Win32 has an analog.

mono/metadata/gc.c
alexanderkyte: There is a need to init. The reference queue relies on the mutex.

jaykrell: There is no need for this zeroing. It is how variables are already initialized. Like the finalizer_mutex, reference_queue_mutex right above.

mono/metadata/threads.c
```diff
@@ -1612,6 +1612,11 @@ ves_icall_System_Threading_InternalThread_Thread_free_internal (MonoInternalThre
 	CloseHandle (this_obj->native_handle);
 #endif
 
+	// Taken care of by reference queue, but we should
+	// zero it out
+	if (this_obj->synch_cs)
```
jaykrell: No need for the `if`, just null it unconditionally.

mono/metadata/threads.c
```diff
@@ -888,19 +904,19 @@ mono_thread_detach_internal (MonoInternalThread *thread)
...
-	 * thread->synch_cs can be NULL if this was called after
-	 * ves_icall_System_Threading_InternalThread_Thread_free_internal.
-	 * This can happen only during shutdown.
-	 * The shutting_down flag is not always set, so we can't assert on it.
+	 * This should be alive until after the reference queue runs the
+	 * post-free cleanup function
 	 */
-	if (thread->synch_cs)
-		LOCK_THREAD (thread);
+	while (TRUE) {
```
jaykrell: Are there paths that take the lock to cover a state change, and more, and read it multiple times under lock and expect it not to change, and this would introduce  race conditions with those? I'd have to survey the code..

mono/metadata/gc.c
jaykrell: We should seriously consider that reference_queue_mutex is a pointer,  its nullness is a lack of init, and cas the ptr,  and  avoid  the separate boolean. The separate booleans tend to create a subtle race condition that people do not see. Merging the init-ness into the state tends to fix it.

mono/metadata/gc.c
jaykrell: If reference_queue_mutex is a pointer, the strongly consider check null, new local, cas it with global, free if race lost. This cuts out some subtle race conditions. Also barrier between init and cas maybe, maybe.

mono/metadata/threads.c
alexanderkyte: I would rather not depend on that. This seems conservative, and remains fast in the correct case.

jaykrell: This is not as conservative as it may seem, as it may create race conditions.  Let me check the code..

jaykrell: For example,  `mono_thread_clear_and_set_state` and `mono_thread_test_and_set_state` do read/modify/write with lock. How does that interoperate with interlocked changes? You cannot easily/generally have multiple locking protocols for the same data.

jaykrell: `mono_thread_suspend` => `async_suspend_internal` similarly.

mono/metadata/gc.c
```diff
@@ -70,6 +71,7 @@ gchar **mono_do_not_finalize_class_names ;
 #define mono_finalizer_unlock() mono_coop_mutex_unlock (&finalizer_mutex)
 static MonoCoopMutex finalizer_mutex;
 static MonoCoopMutex reference_queue_mutex;
+static mono_lazy_init_t reference_queue_mutex_inited = MONO_LAZY_INIT_STATUS_NOT_INITIALIZED;
```
jaykrell: bikeshed: I would drop the `ed` but ok.

mono/metadata/threads.c
alexanderkyte: Worth noting is that before this PR was landed, the control flow would have been to change this enum without locking at all. This would have run after the finalizer, which would have already freed it.

alexanderkyte: I think that we should do the opposite, and create helper libraries for our fancy locking patterns that can be proven correct in isolation.

alexanderkyte: And ban use of mutex primitives in most code

alexanderkyte: I think we've been doing just that, slowly

alexanderkyte:
@monojenkins build failed

PR #9915, opened 2018-08-06
[runtime] Make -Wreturn-type into an error.

vargaz:


lambdageek:
All the autoconf tests are failing... not sure if that's because it should be `-Werror=return-type` or because all the autoconf tests trigger the warning.

PR #9916, opened 2018-08-06
[threads] switch to GC Safe around pthread calls in mono_thread_internal_set_priority

lambdageek:
Observed pthread_getschedparam taking a lock on OSX

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -673,7 +673,9 @@ mono_thread_internal_set_priority (MonoInternalThread *internal, MonoThreadPrior
 
 	g_assert (internal->native_handle);
 
+	MONO_ENTER_GC_SAFE;
 	res = SetThreadPriority (internal->native_handle, priority - 2);
+	MONO_EXIT_GC_SAFE;
 	if (!res)
 		g_error ("%s: SetThreadPriority failed, error %d", __func__, GetLastError ());
 #else /* HOST_WIN32 */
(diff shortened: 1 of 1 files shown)


vargaz:
Is it ok to take a lock if its only taken for a bounded amount of time ?

lambdageek:
@vargaz yes, that's fine as long as the lock is taken consistently either from GC Safe or from GC Unsafe thread state.

The issue is when a thread in GC Safe and another in GC Unsafe thread state both want the same lock. 
 If the GC Safe (preemptively-suspending) thread is preemptively suspended while holding a lock and if the other thread attempts to take the same lock while in GC Unsafe (cooperatively-suspending) thread state, we will get a deadlock between the suspend initiator and the 2 threads contending the lock.

lambdageek:

mono/metadata/threads.c
```diff
@@ -670,12 +670,16 @@ mono_thread_internal_set_priority (MonoInternalThread *internal, MonoThreadPrior
 
 #ifdef HOST_WIN32
 	BOOL res;
+	DWORD last_error;
 
 	g_assert (internal->native_handle);
 
+	MONO_ENTER_GC_SAFE;
 	res = SetThreadPriority (internal->native_handle, priority - 2);
+	last_error = GetLastError ();
```
jaykrell: One could avoid the call if res, but it is incredibly cheap, so ok.
SetThreadPriority is a "slow" syscall, GetLastError is not.

We should have ```inline int-or-enum  MonoToWin32ThreadPriority(int-or-enum a) { return a - 2;}```

I added a comment about the `2` somewhere, probably in `cxx`.

PR #9918, opened 2018-08-06
[mono-api-tools] Split properly for cross-platform

mattleibow:
When creating a diff on windows, the lines are split on the `\n` so the `\r\n` results in multiple lines:

**START SNIP**
Obsoleted methods:

```diff
 [Obsolete ("Use AddRoundRect instead.")]

 public void AddRoundedRect (SKRect rect, float rx, float ry, SKPathDirection dir);
```
**END SNIP**

This PR makes sure to use the platform line endings everywhere. When adding line endings, use the same that `StringBuilder` uses - platform endings. This makes it easy to split - just use the platform as that is what everyone is using.

### DIFF
--- a/mcs/tools/mono-api-html/MarkdownFormatter.cs
+++ b/mcs/tools/mono-api-html/MarkdownFormatter.cs
@@ -191,7 +191,7 @@ namespace Mono.ApiTools {
 
 		public override void Diff (TextWriter output, ApiChange apichange)
 		{
-			foreach (var line in apichange.Member.ToString ().Split (new[] { "\r\n", "\n" }, 0))) {
+			foreach (var line in apichange.Member.ToString ().Split (new[] { Environment.NewLine }, 0)) {
 				if (line.Contains ("+++")) {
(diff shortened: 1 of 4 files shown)


mcs/tools/mono-api-html/MarkdownFormatter.cs
```diff
@@ -191,7 +191,7 @@ public override void DiffRemoval (StringBuilder output, string text, bool breaki
 
 		public override void Diff (TextWriter output, ApiChange apichange)
 		{
-			foreach (var line in apichange.Member.ToString ().Split ('\n')) {
+			foreach (var line in apichange.Member.ToString ().Split (new[] { "\r\n", "\n" }, 0))) {
```
mattleibow: Here I split on both, but technically I could split on `Environment.NewLine` since the way this property is created is with a `StringBuilder` and the `AppendLine` method.

luhenry:
@spouliot can you please take a look at it?

PR #9919, opened 2018-08-06
[sdks] Rename LLVM_VERSION make var to LLVM_RELEASE, LLVM_VERSION is already used by versions.mk.

vargaz:


vargaz:
@monojenkins build failed

Issue #9751, opened 2018-07-26
Mono - Using Integrated Security with SQL server no longer works


NickDufour:
This bug has already been reported on IOS and MAC.  Before mono 5.10, we were able to connect to a SQL server with integrated security.  Since Mono 5.10, we are no longer able to do so.  Connecting directly to SQL Server (with or without integrated security) is not a good approach and we know that.  However, it was requested by many of our clients.  Now many of our client cannot connect to their SQL server.

## Steps to Reproduce

1.  Open a SQL connection using the sample code below

using (SqlConnection connection = new SqlConnection("data source=\<host\>\\\\<instance\>;User ID=\<domain\>\\\\<user\>; Password=\<password\>;database=\<database\>;Integrated Security=True;Connect Timeout=30"))
{
    connection.Open();
}

### Current Behavior

An exception is thrown

### Expected Behavior

It should connect like it used to.

## On which platforms did you notice this

Android Oreo (8.1)

**Version Used**: Mono 5.10 Commit ea8a24b1

## Stacktrace

```
Array cannot be null.
Parameter name: bytes
  at System.Text.Encoding.GetString (System.Byte[] bytes) [0x00003] in /Users/builder/jenkins/workspace/xamarin-android-d15-7/xamarin-android/external/mono/mcs/class/referencesource/mscorlib/system/text/encoding.cs:1527 
  at System.Data.SqlClient.SNI.SNIProxy.GenSspiClientContext (System.Data.SqlClient.SNI.SspiClientContextStatus sspiClientContextStatus, System.Byte[] receivedBuff, System.Byte[]& sendBuff, System.Byte[] serverName) [0x0005d] in /Users/builder/jenkins/workspace/xamarin-android-d15-7/xamarin-android/external/mono/external/corefx/src/System.Data.SqlClient/src/System/Data/SqlClient/SNI/SNIProxy.cs:111 
  at System.Data.SqlClient.SNI.TdsParserStateObjectManaged.GenerateSspiClientContext (System.Byte[] receivedBuff, System.UInt32 receivedLength, System.Byte[]& sendBuff, System.UInt32& sendLength, System.Byte[] _sniSpnBuffer) [0x00000] in /Users/builder/jenkins/workspace/xamarin-android-d15-7/xamarin-android/external/mono/external/corefx/src/System.Data.SqlClient/src/System/Data/SqlClient/TdsParserStateObjectManaged.cs:241 
[...]

Issue #9028, opened 2018-06-05
Mono to SQL Server using integrated security no longer works on iOS


MathieuMorrissette:
We used to be able to connect to an SQL Server  using integrated security by specifying the username and password in the connection string. It is no longer working on mono 5.10.0

I know there are no security benefits to use integrated security on a Unix platform as it is supposed to be password-less but some of my app users were using this feature and are now left unable to connect to their SQL server. Has the support been dropped?

## Steps to Reproduce

1.  Connect to an SQL Server using the code sample below.

```
using (var connection = new SqlConnection("data source=<host>\\<instance>;User ID=<domain>\\<user>; Password=<password>;database=<database>;Integrated Security=SSPI;"))
{
    connection.Open();
}
```

### Current Behavior

Throws an Exception.

### Expected Behavior

Connects just like with Mono 5.8.1 or maybe its just not supported anymore?

## On which platforms did you notice this

iOS, (macOS, android) also have the same issue.

**Version Used**:
Mono 5.10.1.57 (2017-12/ea8a24b1bbf)

## Stacktrace

```
Specified method is not supported.
  at Microsoft.Win32.SafeHandles.SafeGssNameHandle.CreatePrincipal (System.String name) <0x103ce280c + 0x00030> in <3809544b44834bdc9f70d632bd5493e1#00c21f382d13682289c1f5e2f971daa6>:0
  at System.Net.Security.SafeDeleteNegoContext..ctor (System.Net.Security.SafeFreeNegoCredentials credential, System.String targetName) <0x103b9cd84 + 0x0005c> in <3809544b44834bdc9f70d632bd5493e1#00c21f382d13682289c1f5e2f971daa6>:0
  at System.Net.Security.NegotiateStreamPal.EstablishSecurityContext (System.Net.Security.SafeFreeNegoCredentials credential, System.Net.Security.SafeDeleteContext& context, System.String targetName, System.Net.ContextFlagsPal inFlags, System.Net.Security.SecurityBuffer inputBuffer, System.Net.Security.SecurityBuffer outputBuffer, System.Net.ContextFlagsPal& outFlags) <0x103b9c534 + 0x0007f> in <3809544b44834bdc9f70d632bd5493e1#00c21f382d13682289c1f5e2f971daa6>:0
[...]

PR #9922, opened 2018-08-06
[System.Data] use PAL_gssapi.c for SSPI

EgorBo:
PAL_gssapi.c was converted to C recently in corefx but since our fork is not up to date I copied `PAL_gssapi.c` and `PAL_gssapi.h` into mono/metadata (I guess I should just cherry-pick those files to our fork).
For System.Data it will allow users to connect to sql servers using SSPI (Security Support Provider Interface).
See https://github.com/mono/mono/issues/9028 and https://github.com/mono/mono/issues/9751
on macOS (and iOS) it uses built-in GSS.framework.
on Linux it requires an additional package (`krb`) to be installed (see .NET Core prerequisites, e.g. https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x#ubuntu). 
Unfortunately it's not installed on our CI yet.


EgorBo:
@monojenkins build failed

akoeplinger:
@monojenkins build

mcs/class/System.Data/Makefile
```diff
@@ -25,6 +25,10 @@ ifdef MONO_FEATURE_APPLETLS
 LIB_MCS_FLAGS += -d:MONO_FEATURE_APPLETLS
 endif
 
+ifdef ENABLE_GSS
+LIB_MCS_FLAGS += -d:ENABLE_GSS
```
marek-safar: What is this for?

EgorBo: if GSS is not enabled then we do not include [DllImport] and use stubs (like it was before). As far as I remember "dead" DllImports cause problems for AOT, see https://github.com/mono/corefx/blob/master/src/Common/src/Interop/Unix/System.Net.Security.Native/Interop.NetSecurityNative.cs#L15

EgorBo:
@monojenkins build failed

mcs/class/System.Data/Makefile
marek-safar: ENABLE_GSS won't be set for BCL builds, you need to do it manually via `mcs/build/config.make`

marek-safar:
@monojenkins backport 2018-08

marek-safar:
@monojenkins commit csproj

marek-safar:
@alexischr I have merged this in. The name unification will come as part of Martin renaming work.

PR #9924, opened 2018-08-07
Add nuget restore to initial targets for cil-stringreplacer so it will build by default using command-line msbuild

kg:
Right now cil-stringreplacer only builds right by default in VS, you have to manually perform a restore from the command line (so that broke CI). This makes restore happen by default in both modes.


kg:
I've been informed that this is not a reliable solution so we shouldn't land it (https://gist.github.com/jeffkl/37b14b0601bafbdc9462afa8b288fadc#restoring-nuget-packages-before-building)

PR #9926, opened 2018-08-07
[linker]: Bump linker to pickup the new PreserveDependencyInjectorStep.

baulig:


baulig:
@marek-safar I believe you solved this problem in some other way, so we can close this?

marek-safar:
Yes, it should just work by using the general purpose attributes

PR #9930, opened 2018-08-07
Fix a few deficients in mono_path_filename_in_basedir.

jaykrell:
The main one is that strstr is too loose.
Add tests.
Case insensitivity is really a function of filesystem, not host operating system, and the precise casing rules are not available (NTFS format stores upcase table).
Make the Win32 variant testable on Unix.


mono/utils/mono-path.c
```diff
@@ -177,27 +177,129 @@ mono_path_resolve_symlinks (const char *path)
...
+	size_t i = *length;
+	while (i > 0 && mono_path_char_is_separator (path [i - 1]))
+		i -= 1;
+	*length = i;
+}
+
+static gboolean
+mono_path_char_is_lowercase (char ch)
+{
+	return mono_host_win32 && ch >= 'a' && ch <= 'z';
```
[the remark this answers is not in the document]
jaykrell: File system case sensitivity is not locale sensitive. That would fail miserably. There is multi-user persistent sorted data structures. Locale sensitivity would break them. Upcase table is stored per volume by format & not the same through OS versions, but close.

There is an approximation in win32, but using 8 bit characters seems maybe far from correct in the first place. I can check.
CompareString with invariant culture or somesuch?

Yes we could depend on canonicalization wrt slashes.

lambdageek:
Thanks for that explanation of why locales are the wrong thing.  Makes sense!

I think `mono_path_filename_in_basedir` should assume its inputs have been through `mono_path_canonicalize` and defer the separator normalization to it.

As far as the case manipulation, I would prefer that we use `g_ascii_toupper`, `g_ascii_strncasecmp` etc, if possible.

lambdageek:
Hm.  a little bit surprised by the failing loader tests - LoadFile should throw an exception if given relative path names.  I made a new issue https://github.com/mono/mono/issues/9943

jaykrell:
I think I'll just remove case insensitivity.
Test will call canonicalize.
Function in question will assume it.
Test will not alter case.

mono/utils/mono-path.c
```diff
@@ -177,27 +177,129 @@ mono_path_resolve_symlinks (const char *path)
 #endif
 }
 
+#ifndef mono_host_win32
+#ifdef HOST_WIN32
+#define mono_host_win32 TRUE
```
luhenry: Just use `HOST_WIN32` around the code. There are only 3 uses of `mono_host_win32` so it will be easy I am sure.

jaykrell: I changed to `HOST_WIN32`. The test deliberately does not use `ifdef`, in order to get win32 and non-win32 coverage, on win32 and non-win32 platforms; without affecting the performance of optimized mainline builds -- `HOST_WIN32` still being a constant `#define`, `0` or `1` there, but not `#ifdef`.

jaykrell:
@monojenkins build failed

jaykrell:
@lambdageek old PR, in better shape now, including finding that the inputs are not as specified, so previous behavior less predictable.

mono/unit-tests/test-path.c
```diff
@@ -0,0 +1,79 @@
+/*
+ * test-path.c
+ */
+
+#include "config.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include "glib.h"
+
+gboolean mono_host_win32;
```
luhenry: Remove this hack and replace it by #ifdef HOST_WIN32

mono/utils/mono-path.c
```diff
@@ -177,27 +177,132 @@ mono_path_resolve_symlinks (const char *path)
 #endif
 }
 
+#ifndef HOST_WIN32
```
luhenry: Remove this hack and replace it by #ifdef HOST_WIN32

jaykrell: This is a good technique to enable testing Win32 on Posix and vice versa.
The mainline code is as optimized as usual while a separate test variant enables more/easier testing.

PR #9945, opened 2018-08-07
Change platform specific sources overrides from hostplatform_libname to hostplatform_defaultprofile_libname

kg:
The commit https://github.com/mono/mono/commit/f602125d107b0966f8045670326124ad988291cc accidentally introduced a (hostplatform)_(libname) style sources file, making it get used for any scenario where the host platform was win32 and we didn't have a more specific sources file. We didn't notice this hazard when adding support for the hostplatform style so this PR changes the format to (hostplatform)_defaultprofile_(libname) to make it explicit that the sources file is intended to behave this way. The hazard is partly a result of us already naming common included sources files in a similar pattern, like win32_foo.sources, and then including them from a bunch of profiles.

This PR also updates the 'load entire folder' path used by genproj to check for any uses of the hazardous hostplatform style that were not also includes, and I didn't find any. I suspect we might have introduced one intentionally, but I don't have any way to identify those. @baulig do you know if there are any cases where we wanted to use a win32_xxx as both an include and a direct sources file so I can fix them?

### DIFF
--- a/mcs/build/gensources.cs
+++ b/mcs/build/gensources.cs
@@ -68,7 +68,7 @@ public static class Program {
             Console.Error.WriteLine ("--stdout");
             Console.Error.WriteLine ("  Writes results to standard output (omit outputFileName if you use this)");
             Console.Error.WriteLine ("--strict");
-            Console.Error.WriteLine ("  Produces an error exit code if files or directories are invalid/missing");
+            Console.Error.WriteLine ("  Produces an error exit code if files or directories are invalid/missing or other warnings occur");
             Console.Error.WriteLine ("--basedir:<dir>");
             Console.Error.WriteLine ("  Sets the base directory when reading a single sources/exclusions pair (default is the directory containing the sources file)");
             return 1;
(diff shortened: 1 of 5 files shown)


marek-safar:
@monojenkins build failed

kg:
Looks like System.Security depended on the ambiguous behavior on win32... gonna have to figure that out.

kg:
@monojenkins build failed

kg:
@monojenkins build failed

kg:
@monojenkins commit csproj

kg:
@monojenkins build Linux x64

kg:
@monojenkins build Linux x64

kg:
@monojenkins commit csproj

kg:
@monojenkins commit csproj

kg:
@monojenkins build Linux x64 Coop Suspend and Linux ARMv7

kg:

kg:

kg:
@monojenkins commit csproj

kg:
@monojenkins build failed

Issue #9935, opened 2018-08-07
tests/src/baseservices/threading/readerwriterlockslim/upgrader.cs CoreCLR acceptance-test fails


akoeplinger:
## Steps to Reproduce

1. Run third party acceptance tests, via `make -C acceptance-tests coreclr-runtest-basic`
2. Observe tests/src/baseservices/threading/readerwriterlockslim/upgrader.exe test failing

This was most probably introduced by https://github.com/mono/mono/pull/9811. /cc @lambdageek 

### Current Behavior

```
MESSAGE:
Main thread EnterUpgradeableReadLock !
    @Thread 0 start 
    @Thread 0 Signal 
    @Thread 1 start 
    @Thread 1 Signal 
    @Thread 2 start 
    @Thread 2 Signal 
    @Thread 3 start 
    @Thread 3 Signal 
    @Thread 4 start 
    @Thread 4 Signal 
    @Thread 5 start 
    @Thread 5 Signal 
    @Thread 6 start 
    @Thread 6 Signal 
    @Thread 7 start 
    @Thread 7 Signal 
    @Thread 8 start 
    @Thread 8 Signal 
    @Thread 9 start 
    ev.wait 
    @Thread 9 Signal 
    @Main thread cannot enter Write Lock! 
    Test Passed! 
* Assertion at exception.c:108, condition `is_ok (error)' not met, function:mono_exception_new_by_name_domain, (null) assembly:/mnt/jenkins/workspace/test-mono-pull-request-acceptance-tests/mcs/class/lib/net_4_x-linux/mscorlib.dll type:ThreadAbortException member:(null)


+++++++++++++++++++
STACK TRACE:
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Threading.Thread.JoinInternal (System.Threading.Thread,int) [0x00017] in <29c7e323d116485c94ee214ef79b56c5>:0
  at System.Threading.Thread.Join () [0x00000] in <29c7e323d116485c94ee214ef79b56c5>:0
  at Upgrader.ThreadMethod (object) [0x00072] in <cbf71e0b78aa403a83e66a28fdebbd56>:0
  at System.Threading.ThreadHelper.ThreadStart_Context (object) [0x00025] in <29c7e323d116485c94ee214ef79b56c5>:0
[...]

PR #9948, opened 2018-08-07
[coop] Thread.JoinInternal should not call mono_error_set_pending_exception

lambdageek:
The icall was converted to use `HANDLES()`, so the icall wrapper will set the
pending exception from the `MonoError`.  If the icall does it, too, the second
use (in the wrapper) will assert that the `MonoError` is being reused after
cleanup without being reinitialized.

Also add an assertion in `mono_error_get_message` that we're not trying to get a
message from a cleaned up `MonoError`.

Fixes https://github.com/mono/mono/issues/9935

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -2047,10 +2047,10 @@ ves_icall_System_Threading_Thread_Join_internal (MonoThreadObjectHandle thread_h
 
 	mono_thread_clr_state (cur_thread, ThreadState_WaitSleepJoin);
 
-	mono_error_set_pending_exception (error);
-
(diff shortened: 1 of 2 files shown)


jaykrell:
I thought redundant setting of error was ok -- well..on the same call/thread, I wasn't thinking about the same error. Maybe need to audit for more of these. :(
And I still suspect the ban on implicit reuse is too strict, but ok.

jaykrell:
I think icall wrapper should just accept and ignore a cleaned up error..hoping the cleaner-upper was correct, that it either set the error, or ignored it.

I didn't realize setting the thread's pending error cleaned up the error..but it makes sense..as well..maybe there we should just reinit..

lambdageek:

lambdageek:

PR #9949, opened 2018-08-08
[cxx jit x86] Make x86-specific parts of mini valid C++

jaykrell:
Cast from void* to non-void*.
Cast from function pointer to void*.
Cast pointer to gsize, and then to enum.
Cast int to enum.

### DIFF
--- a/mono/mini/exceptions-x86.c
+++ b/mono/mini/exceptions-x86.c
@@ -86,7 +86,7 @@ mono_win32_get_handle_stackoverflow (void)
 		return start;
 
 	/* restore_contect (void *sigctx) */
-	start = code = mono_global_codeman_reserve (128);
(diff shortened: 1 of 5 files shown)


jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build  OS X i386

jaykrell:
@monojenkins build OS X i386

jaykrell:
cc @vargaz

jaykrell:
I think I can eliminate the addition of casts on a fixed set allocating functions to an indefinite set of types. Will try this morning. So maybe a lot less line damage/noise.

jaykrell:
Temporarily closing, trying another approach.

PR #9952, opened 2018-08-08
Move to generating msbuild choose elements to get if-else selection behavior for sources in projects so that we don't get erroneous duplicate files in cases where there are both profile and host platform criteria

kg:
A recent commit revealed that in cases where we select based on a mix of host platform and profile, genproj csproj files can end up with duplicate sources because the existing ```<ItemGroup Condition=``` approach could make multiple groups match for a given compile when we really just want one.

This PR changes to generating a cascade of msbuild ```<Choose>``` elements, which give if-else selection to ensure that we only ever build a single set of files.


kg:
This is currently causing errors in VS2017, so I need to dig into that some

akoeplinger:
@monojenkins commit csproj

PR #9953, opened 2018-08-08
[cxx] [boehm] extern "C" and interface void*  instead of char* for use by  C++

jaykrell:
Add extern "C" correctly to boehm so it can be C, and used by C++.
Change char* to void*  for use by C++ callers.

This tracks mainline but just barely.

https://github.com/ivmai/bdwgc/blob/ae3d724f3645af119e1b2c2cefbb71c97d467546/mark.c#L1337

https://github.com/ivmai/bdwgc/commit/f4fbef0f7f0e3446856d2177535080158e16e403#diff-74e77959a62b8a560274f69789f50a94R1623

https://github.com/ivmai/bdwgc/commit/a825a2d612e749c08c24fca0f11eb2a523ce3b6b

etc.

### DIFF
--- a/libgc/mark.c
+++ b/libgc/mark.c
@@ -1191,11 +1191,11 @@ void GC_mark_init()
  * Should only be used if there is no possibility of mark stack
  * overflow.
  */
-void GC_push_all(bottom, top)
-ptr_t bottom;
-ptr_t top;
+void GC_push_all(void* void_bottom, void* void_top)
 {
-    register word length;
+    ptr_t bottom = (ptr_t)void_bottom;
+    ptr_t top = (ptr_t)void_top;
+    word length;
     
     bottom = (ptr_t)(((word) bottom + ALIGNMENT-1) & ~(ALIGNMENT-1));
     top = (ptr_t)(((word) top) & ~(ALIGNMENT-1));
--- a/libgc/mark.c
+++ b/libgc/mark.c
@@ -1191,7 +1191,9 @@ void GC_mark_init()
  * Should only be used if there is no possibility of mark stack
  * overflow.
  */
-void GC_push_all(void* void_bottom, void* void_top)
+void GC_push_all(void_bottom, void_top)
+void* void_bottom;
+void* void_top;
 {
     ptr_t bottom = (ptr_t)void_bottom;
     ptr_t top = (ptr_t)void_top;
--- a/libgc/include/gc.h
+++ b/libgc/include/gc.h
@@ -42,11 +42,6 @@
 #   define GC_CONST
 #  endif
 
-# ifdef __cplusplus
-    extern "C" {
-# endif
-
-
 /* Define word and signed_word to be unsigned and signed types of the 	*/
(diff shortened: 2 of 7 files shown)


jaykrell:
@monojenkins build OS X x64

jaykrell:

jaykrell:
Review?

lambdageek:
Seems alright.

How come https://github.com/ivmai/bdwgc/commit/a825a2d612e749c08c24fca0f11eb2a523ce3b6b is much more invasive than what we have here?  Is any of that stuff needed for correctness?

We don't have any CI of Mono using Boehm, as far as I know.  Have you tried running some of the tests locally for a sanity check?

jaykrell:
Good questions.

This was done mostly separately from mainline, not specifically subsetted.

Their goal is larger: Be all valid C++. Ours is only to serve C++ callers.
Maybe I overshot that, not sure. I will explain.

I did this work in January 2018. Some memory is lost.

Need to wrap declarations, but not includes, in externC. Each header owns doing this for itself.
If I did that within implementation also, that is overshoot, could be undone, but is also ok.

I have broken boehm in the past in CI, at least building it, so there is some coverage, at least building. 
(I had/do locally used configure -disable-boehm, so broke only in CI.)

I did not run tests.

jaykrell:
Looking at mainline, briefly, you can see they fixed keywords (new) and cast void* to non-void*.

I have no idea generally how we diverge. They are under very active maintenance, we are not here.

jaykrell:
Just waiting for Linux ARMv7..

libgc/mark.c
```diff
@@ -1191,11 +1191,11 @@ void GC_mark_init()
  * Should only be used if there is no possibility of mark stack
  * overflow.
  */
-void GC_push_all(bottom, top)
-ptr_t bottom;
-ptr_t top;
+void GC_push_all(void* void_bottom, void* void_top)
```
luhenry: I understand you might find it very ugly (we all do), but keep the old C parameters declaration.

jaykrell: The types, or syntax, or both?

jaykrell: This matches mainline and unity fwiw.

luhenry:
```
void GC_push_all(bottom, top)
void* bottom;
void* top;
```

jaykrell: The types matter, a little, but I can undo that if you insist. The syntax does not matter at all, looks fine either way.

jaykrell: Ok, no problem. It was char* before, is the real point.

jaykrell:
ok?

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build  OS X i386

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X i386

PR #9955, opened 2018-08-08
[cxx] Remove two excess toplevel consts in structs, one a C++ problem, one very obscure/hypothetical.

jaykrell:


mono/metadata/object-internals.h
```diff
@@ -356,7 +356,7 @@ struct _MonoIUnknownVTable
 
 struct _MonoIUnknown
 {
-	MonoIUnknownVTable const * const vtable;
+	const MonoIUnknownVTable *vtable;
```
jaykrell: Hypothetical problem.

mono/utils/mono-merp.c
```diff
@@ -97,7 +97,7 @@ typedef struct {
 	const char *osVersion; 
 	int uiLidArg; // Application LCID 
 
-	const char systemModel [100];
+	char systemModel [100];
```
jaykrell: Real problem.

jaykrell:

jaykrell:
@monojenkins build Linux x64

jaykrell:

jaykrell:
Review?

mono/utils/mono-merp.c
lambdageek: What's the problem here? all uses of systemModel are either okay with `const` or cast constness away.

I'm not opposed to this change, but I'm curious what it's fixing.

jaykrell: Error message is in the commit. Having const members requires initializing them, once — or multiple times and casting away const as you say. There may be another fix.

jaykrell: I’m really torn on this. More const seems good, even if interface-only. Lack of these errors in general suggests it is rare. And not idiomatic. But still good. Rare != bad, often the opposite. New/better implies rare.

The secondary reason is a recent msvc compiler regression and will not stand long, did ship, does not hit us, probably never will.

PR #9968, opened 2018-08-08
[Facades] Adds System.Threading.Tasks.Extensions for mobile profiles

marek-safar:


marek-safar:
@akoeplinger why? that table is used by desktop loader only and the facade is mobile only

akoeplinger:
Ok, I found a few of the other mobile-only facades in there but e.g. System.Memory.dll and System.Drawing.Common.dll is missing. Would be nice to have everything in there even if it's not needed just for consistency :)

marek-safar:
@monojenkins backport 2018-06

marek-safar:
@monojenkins backport 2018-08

akoeplinger:

onovotny:
This is causing active pain with users today. Getting these facades out cannot wait until the next VS release, IMO, as 15.8 just shipped. It's too long to wait.

I really think this has to be in a 15.8.x servicing release.

marek-safar:
@monojenkins backport 2018-04

ahmedalejo:
> @monojenkins backport 2018-04

Hi @marek-safar how could we try out this backport?
currently using Xamarin.Android.SDK v9.0.0.19

marek-safar:
@ahmedalejo I don't we have any available build with the fix for Xamarin.Android because the issue was not reported

@jonpryor can give you better estimate when this is going to be fixed for XA

ahmedalejo:
> @ahmedalejo I don't we have any available build with the fix for Xamarin.Android because the issue was not reported
> 
> @jonpryor can give you better estimate when this is going to be fixed for XA

Thanks for this quick feedback much appreciated

PR #9969, opened 2018-08-08
[metadata] Add nullfs to list of recognized filesystems

akoeplinger:
Noticed this today when the corlib test which checks for unknown filesystems failed. Turned out it was because of a .dmg I had mounted:

```
$ mount | grep nullfs
/Volumes/Skype Meetings App/Skype Meetings App.app on /private/var/folders/cs/kt8x25p14g9dlk_p192rd7th0000gn/T/AppTranslocation/957CCB05-C113-4897-9C80-700C2CE4809E (nullfs, local, nodev, nosuid, read-only, noowners, quarantine, nobrowse, mounted by alexander)
```

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -4415,6 +4415,7 @@ static _wapi_drive_type _wapi_drive_types[] = {
 	{ DRIVE_CDROM, "cddafs" },
 	{ DRIVE_CDROM, "cd9660" },
 	{ DRIVE_RAMDISK, "devfs" },
+	{ DRIVE_RAMDISK, "nullfs" },
 	{ DRIVE_FIXED, "exfat" },
 	{ DRIVE_RAMDISK, "fdesc" },
 	{ DRIVE_REMOTE, "ftp" },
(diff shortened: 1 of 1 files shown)


jaykrell:
Is there a more auto-maintaining way to do this? It seems to be small bug farm.

akoeplinger:
@jaykrell nothing off the top of my mind. It's also mostly a cosmetic issue (the C# API will just return RamDisk now instead of Unknown), except for our tests.

jaykrell:
Hm. At least on Darwin statfs returns a string but indeed looks not on Linux.

migueldeicaza:
Found another MacOS file system "acfs", networked file system

jaykrell:
I don’t think we need this table for mac. Statfs. I’ll look later..

akoeplinger:
I'm going to merge this now, we can update later if we find a better way to get the values :)

NattyNarwhal:
Would statvfs on Darwin return a string in one of its members? AIX is in a similarish boat; not helped by PASE having different values on the type ints, so I just used a string member in statvfs instead of enumeration of values based on the type ints.

Issue #9947, opened 2018-08-07
AOT images compiled with safepoints should be usable with MONO_THREADS_SUSPEND=preemptive


lambdageek:
If Mono is compiled with hybrid or cooperative suspend on by default, AOT images for mscorlib.dll and other assemblies will be created with safepoints enabled.

If you then run `MONO_THREADS_SUSPEND=preemptive mono Foo.exe` you will get:
```
* Assertion at mini-runtime.c:1405, condition `mono_threads_are_safepoints_enabled ()' not met
```

We should allow the AOT images with safepoints to be usable in a runtime with safepoints disabled.

This affects `master` and `2018-08`

PR #9974, opened 2018-08-08
[mini] Don't assert on MONO_PATCH_INFO_GC_SAFE_POINT_FLAG if coop is disabled

lambdageek:
If an AOT library is loaded that was compiled with safepoints, but currently
we're running with preemptive suspend (e.g. `MONO_THREADS_SUSPEND=preemptive` is
set), don't assert.  `mono_polling_required` will always be false, so the
safepoint will be ignored.

Fixes https://github.com/mono/mono/issues/9947


lambdageek:

lewurm:
@monojenkins squash

lambdageek:
@monojenkins build failed

PR #9954, opened 2018-08-08
[cxx][jit][arm] Make ARM32-specific JIT code valid C++.

jaykrell:
Lots of casts
One type change (local variable guint8*  tramp to gpointer)
Rewrite one key macro (ARM_EMIT).
Note this done with Linux/gcc. Android/clang might be different/stricter.

PR #9982, opened 2018-08-08
[cxx] ARM_EMIT w/o address-taken

jaykrell:
Alternative to https://github.com/mono/mono/pull/9954#discussion-diff-208670369R24.


jaykrell:
@monojenkins build failed

vargaz:
What exactly is the problem with the original version ?

jaykrell:
Assignment from void* to non-void*, where p is updated at the end.

jaykrell:
Temporarily closing, trying another approach.

jaykrell:
This is a fine approach.
Or we can use `g_cast` from https://github.com/mono/mono/pull/10027.

PR #9994, opened 2018-08-09
[cxx] Make glib valid C++.

jaykrell:
Mostly cast from void* to non-void*.
Also cast int to enum.
Remove register which is a warning currently, maybe error later.

Build support.

### DIFF
--- a/mono/eglib/Makefile.am
+++ b/mono/eglib/Makefile.am
@@ -53,6 +53,7 @@ libeglib_la_SOURCES = \
 	unicode-data.h	\
 	$(os_files)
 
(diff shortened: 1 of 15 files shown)


jaykrell:
@monojenkins build OS X i386

jaykrell:
I can probably avoid most of the casts from void* to non-void*.

jaykrell:
Closing temporarily while I try to significantly reduce the diff.

jaykrell:
These changes are all ok to me, but a lot can go away if https://github.com/mono/mono/pull/10027 is accepted. I don't think the call-site casts are so terrible but right now the jury is out.

PR #9999, opened 2018-08-09
[wasm] First round of changes for the AOT compiler to handle WASM.

kumpera:


configure.ac
```diff
@@ -3891,6 +3892,20 @@ BTLS_SUPPORTED=no
 with_tls=pthread
 target_mach=no
 
+elif test "x$enable_wasm_aot" = "xyes"; then
```
vargaz: Why can't we use the normal cross compiling code in configure.ac ?

kumpera: Because autoconf doesn't accept any wasm triplet and faking it as a linux variant made the changes even worse.

kumpera:
@monojenkins build failed

PR #10007, opened 2018-08-09
Add lock around SynchronizedCollection<T>.Add()

gregmac:
This fixes a possible ArgumentOutOfRange exception when `SynchronizedCollection<T>.Remove(T)` is called simultaneously with `SynchronizedCollection<T>.Add(T)`.

Previously reported in https://bugzilla.xamarin.com/show_bug.cgi?id=43447

Note that the `IList.Add(object)` implementation is locked and doesn't have this problem, so the following usage is effectively a workaround for this bug:

    var collection = new SynchronizedCollection<int>();
    //....
    ((IList)collection).Add(1);

### DIFF
--- a/mcs/class/System.ServiceModel/System.Collections.Generic/SynchronizedCollection.cs
+++ b/mcs/class/System.ServiceModel/System.Collections.Generic/SynchronizedCollection.cs
@@ -105,7 +105,9 @@ namespace System.Collections.Generic
 
 		public void Add (T item)
 		{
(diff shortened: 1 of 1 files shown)


akoeplinger:
Thanks!

This looks good but I went a step further and replaced the whole SynchronizedCollection with the referencesource implementation which does have the locking in `Add()` too: https://github.com/mono/mono/pull/10021 😄

marek-safar:
Closing in favour of #10021

PR #10008, opened 2018-08-09
[w32file] Fix usage of copyfile

luhenry:
When passing COPYFILE_CLONE, copyfile will not follow a symbolic link on the src file. This is a problem since it breaks File.Copy assumption which, when passed a symbolic link, follows the symbolic link instead of copying the symbolic link itself. The bug is visible as soon as you use `<Copy />` with msbuild and you try to access the copied file with `<Touch />` for example

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -16,9 +16,6 @@
 #endif
 #ifdef HAVE_COPYFILE_H
 #include <copyfile.h>
-#  if !defined(COPYFILE_CLONE)
-#    define COPYFILE_CLONE (1 << 24)
-#  endif
 #endif
 #if defined(HAVE_SYS_STATFS_H)
 #include <sys/statfs.h>
@@ -2421,7 +2418,7 @@ CopyFile (const gunichar2 *name, const gunichar2 *dest_name, gboolean fail_if_ex
 		}
 	}
 
-	ret = _wapi_copyfile (utf8_src, utf8_dest, NULL, COPYFILE_ALL | COPYFILE_CLONE | (fail_if_exists ? COPYFILE_EXCL : COPYFILE_UNLINK));
+	ret = _wapi_copyfile (utf8_src, utf8_dest, NULL, COPYFILE_ALL | COPYFILE_ACL | COPYFILE_STAT | COPYFILE_XATTR | (fail_if_exists ? COPYFILE_EXCL : COPYFILE_UNLINK));
 	g_free (utf8_src);
 	g_free (utf8_dest);
 	if (ret != 0) {


luhenry:

luhenry:

luhenry:

luhenry:

luhenry:
@filipnavara could you please have a look at that since you authored https://github.com/mono/mono/pull/9642

filipnavara:
It is definitely an issue, but removing COPYFILE_CLONE kinda defeats the whole purpose of using `copyfile` in the first place. There's no unit test covering this specific use case, which would be really nice to have. I am fine with reverting my PR or applying your patch for now, but a proper fix would be to detect and handle symlinks and either follow them or drop COPYFILE_CLONE in that specific case. There's already some `stat` called earlier in the function, so it should not be too hard to detect the case.

filipnavara:
Apparently it used to work correctly until Apple "fixed" it in 10.13.2:

> This change was intentional as following symlinks allows you to clone directories (which is disallowed due to potential system slowdown in apfs). We will update the man page to reflect this.
>
> This is not a bug in clonefileat(), this was an intentional change in copyfile itself.

I was looking into using `clonefile` directly to get around all the `copyfile` quirks. However, the API is only available on newer macOS versions and thus hard to consume if down-level compatibility is desired with the same binaries.

luhenry:
We should be able to have a check for `clonefile` in our `configure.ac`, use it if available, fallback to `copyfile` (without `COPYFILE_CLONE`) otherwise. I am pretty sure `copyfile` is still going to be faster than our handrolled `read`+`write` just because of the absence of userspace<->kernelspace memory copy.

filipnavara:
`copyfile` does not really help performance-based compared to read+write. The whole code is open source (https://opensource.apple.com/source/copyfile/copyfile-146.30.2/) and the actual copying is still just read+write. In addition it can do all the nice things like copying extended attributes, but that's not why it was used in the first place.

`clonefile` is not a silver bullet, but macOS now uses apfs on all system drives, so it will work there. The hope was to avoid the copy overhead during msbuild runs that do a lot of copying. Using `configure` to check for it is not an option unfortunately. Mono distributes single package for all macOS versions, so it would have to be used through dynamic loading and with proper fallbacks.

filipnavara:
I'm working on a PR with `clonefile`. If everything goes smooth I will submit it later today.

luhenry:
I definitely agree `clonefile` would be better than `copyfile`, that's a given. What I meant is `clonefile` might not be working everywhere, so we still **require** the fallback to `copyfile`. I'll take a look at https://github.com/mono/mono/pull/10020 right now.

PR #10015, opened 2018-08-10
Don't Merge - Cleanup after an Exception to prevent major memory leak

### LINKED ISSUE #8689: Memory leak in HttpWebRequest / HttpWebResponse when using a WebProxy
This is a follow-on of sorts to bug [#7356](https://github.com/mono/mono/issues/7356) . The test is essentially the same thing except through a WebProxy.

## Steps to Reproduce

1. Use Ubuntu 16.04. Install mono-devel, mono-utils. mono-profiler, whatever is necessary to have the compiler and profiler. I doubt this is specific to the Linux version, however.  You will also need a proxy server through which to make the web requests.  Replace **yourproxyip** and **yourproxyport** withthe proper values for your proxy server.
2. The program below simply loops over a set of popular websites and gets their homepages through a proxy server, using HttpWebRequest and HttpWebResponse. Compile and run it under the profiler, e.g.
```
mono --gc=sgen --profile=log:heapshot=50gc TestMemGrow.exe
```
3. Let it run for several hours (3-4, or more). After that, to view the profiler output enter:

```
mprof-report output.mlpd > out.txt
grep System.Byte out.txt
(grep -F "System.String" | grep -Fv "System.String[]") < out.txt
```

You should see  lists of allocations of System.Byte[] and System.String that grow increasingly larger. You can leave the program running and re-execute those commands, to see the continued growth of memory usage as the profiler adds more heap shots to its data file.  In my testing the heapshots came about 10 minutes apart.

I left in the GC calls from my prior bug, and this example writes the web page data to a file.  It also writes exceptions to a file exceptions.txt, because it is possible (I really don't know) that it is the exception handling leaking memory, so that information might be useful.

To execute this code you will need a proxy server through which you make the web requests.  Fill in your proxy server's IP and port in the appropriate place in the code.

**CODE:**

```C#
using System;
using System.IO;
using System.Net;
using System.Runtime;

namespace TestMemGrow
{
    class Program
    {
        static void Main()
        {
		List<string> websites = new List<string>() { "http://www.amazon.com", "http://www.google.com", "http://www.yahoo.com", "http://www.ebay.com", "http://www.overstock.com",
                "http://www.marketwatch.com", "http://www.bn.com", "http://www.newegg.com", "http://www.wsj.com", "http://www.arstechnica.com", "http://www.slashdot.org",
                "http://www.mediaite.com", "http://www.disqus.com", "http://www.twitter.com", "http://www.snap.com", "http://www.facebook.com", "http://www.usps.com",
                "http://www.ups.com", "http://www.techcrunch.com", "http://www.oracle.com", "http://www.java.com", "http://www.apple.com", "http://www.microsoft.com",
                "http://www.ibm.com", "http://www.dell.com", "http://www.asus.com", "http://www.gigabyte.com", "http://www.intel.com", "http://www.crucial.com",
                "http://www.westerndigital.com", "http://www.samsung.com", "http://www.sandisk.com", "http://www.brother.com", "http://www.hp.com",
                "http://www.msn.com", "http://www.disney.com", "http://www.nintendo.com", "http://www.twitter.com", "http://www.youtube.com",
                "http://www.instagram.com", "http://www.linkedin.com", "http://www.wordpress.org", "http://www.pinterest.com", "http://www.wikipedia.org",
                "http://www.blogspot.com", "http://www.adobe.com", "http://www.tumblr.com", "http://www.vimeo.com", "http://www.flickr.com", "http://www.godaddy.com",
                "http://www.buydomains.com", "http://www.reddit.com", "http://www.w3.org","http://www.nytimes.com", "http://www.statcounter.com",
                "http://www.weebly.com","http://www.blogger.com","http://www.github.com", "http://www.jimdo.com", "http://www.myspace.com",
                "http://www.mozilla.org", "http://www.gravatar.com", "http://www.theguardian.com", "http://www.bluehost.com", "http://www.cnn.com", "http://www.foxnews.com",
                "http://www.msnbc.com", "http://www.wix.com", "http://www.paypal.com","http://www.stumbleupon.com", "http://www.digg.com","http://www.huffingtonpost.com",
                "http://www.feedburner.com", "http://www.imdb.com","http://www.yelp.com","http://www.dropbox.com", "http://www.baidu.com","http://www.washingtonpost.com",
                "http://www.slideshare.net","http://www.etsy.com","http://www.telegraph.co.uk", "http://www.about.com", "http://www.bing.com", "http://www.latimes.com",
                "http://www.tripadvisor.com","http://www.opera.com", "http://www.live.com", "http://www.wired.com", "http://www.bandcamp.com"};
			StreamWriter sw = new StreamWriter("exceptions.txt");

            while (true)
            {
                foreach (string website in websites)
                {
                    try
                    {
                        GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
                        GC.Collect();
			Console.WriteLine(DateTime.Now.ToString("HH:mm:ss") + " Getting " + website);
                        HttpWebRequest request = (HttpWebRequest)WebRequest.Create(website);
			WebProxy myProxy = new WebProxy("yourproxyip", yourproxyport);
			request.Proxy = myProxy;
                        HttpWebResponse response = request.GetResponse() as HttpWebResponse;
                        StreamReader reader = new StreamReader(response.GetResponseStream());
                        File.WriteAllText("florb.txt", reader.ReadToEnd());
                        reader.Close();
                        response.Close();
                    }
                    catch (Exception ex)
                    {
			string exString = ex.ToString();
			exString = exString.Substring(0, Math.Min(150, exString.Length));
                        Console.WriteLine("Caught exception . . . " + exString);
			sw.WriteLine("EXCEPTION\r\n" + ex.ToString() + "\r\n");
			sw.Flush();
                    }
                }
            }
        }
    }
}
```

### Current Behavior

Memory allocations of System.Byte[] and System.String increase apparently without limit.

### Expected Behavior

The garbage collector would clean up these allocations.

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.12.0.226 (tarball Thu May  3 09:48:32 UTC 2018)
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
```

## Other Matters

I am not at all certain whether this leak is caused by the use of a WebProxy itself, or by the increased number of exceptions, particularly WebExceptions, that occur when using a WebProxy to make the requests.

nealef:
This is an attempt to fix #8689. When an exception happens during the WebRequest there is no response generated but GetException is called in MonoBtlsContext to gather the stacktrace etc. That is formed into a message that is used to throw an exception. However, the stacktrace artifacts etc are left hanging around. Very quickly, using the test case shown in the #8689 issue memory use grows substantially. In addition the read and write buffers used by MobileAuthenicatedSteam are left hanging around as well. This patch will, when an exception is thrown, clean up these artifacts by disposing of things like MonoBtlsSsl, MonoBtlsSslCtx, MonoBtlsBio ... I'm not comfortable with use of Dispose() to do the cleanup but the Close methods of Ssl etc are protected so can't be performed in the same place as the Dispose().

Also, the size of the read and write buffers are 16834 and 16384 which seems like a transposition error for the 16834 case. 

I also wonder that even for successful HttpWebRequests and responses whether some cleanup is required when KeepAlive property is not set otherwise those same artifacts will stay alive (but because we don't have lots of stacktraces) the growth in memory footprint is not as extensive. For connections that are kept alive the Close() method should probably do the same cleanup.

This is a first pass effort designed to generate discussion on the issue which has become critical for a project we are looking to get into production.

### DIFF
--- a/mcs/class/System/Mono.Btls/MonoBtlsContext.cs
+++ b/mcs/class/System/Mono.Btls/MonoBtlsContext.cs
@@ -160,7 +160,7 @@ namespace Mono.Btls
 			}
 		}
 
-		static Exception GetException (MonoBtlsSslError status)
+               private Exception GetException (MonoBtlsSslError status)
 		{
 			string file;
 			int line;
@@ -179,6 +179,7 @@ namespace Mono.Btls
 				message = string.Format ("{0} {1}\n  at {2}:{3}", status, text, file, line);
 			else
 				message = string.Format ("{0} {1}", status, text);
+                       Dispose (true);
 			return new MonoBtlsException (message);
 		}
 
--- a/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
@@ -393,8 +393,8 @@ namespace Mono.Net.Security
 			}
 
                        if (result.Error != null) {
-                               readBuffer = null;
-                               writeBuffer = null;
+				readBuffer = null;
+				writeBuffer = null;
 				result.Error.Throw ();
                        }
 		}
--- a/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
@@ -392,8 +392,11 @@ namespace Mono.Net.Security
 				}
 			}
 
-			if (result.Error != null)
+                       if (result.Error != null) {
+                               readBuffer = null;
+                               writeBuffer = null;
 				result.Error.Throw ();
+                       }
 		}
 
 		protected abstract MobileTlsContext CreateContext (MonoSslAuthenticationOptions options);


nealef:
Has anyone had time to review this PR?

baulig:
I have a very bad feeling about this and am afraid we might cause more harm than good.

The underlying issue is that whatever is calling this code and handling the exception upstream does not clean up the SslStream.  Adding some code that will call Dispose() when the exception is _thrown_ will only hide that, not fix it.  And I would expect that you still have that very same problem when using AppleTls.

baulig:
We should figure out where that exception is being caught / handled and why that code does not close the connection and dispose everything.

nealef:
> I have a very bad feeling about this and am afraid we might cause more harm than good

I had this feeling also, but looking at the Dispose() being invoked it's cleaning up things like ssl, bio, bioerr, and certificate stuff that should have no visibility. I would've explicitly disposed these individually but they are not visible from that class.  Once GetException() has done its thing then all the stuff that was used to construct the error message is no longer needed and neither are the ssl context and session related things which get setup with every HttpWebRequest request() operation.

In MonoBtlsContext.cs the relevant Dispose method includes:

```
Dispose (ref ssl);
Dispose (ref ctx);
Dispose (ref remoteCertificate);
Dispose (ref nativeServerCertificate);
Dispose (ref nativeClientCertificate);
Dispose (ref clientCertificate);
Dispose (ref bio);
Dispose (ref errbio);
```

Also, isn't HttpWebRequest a one-shot unless keepalive is set? That is, even when successful the resources need to be cleaned up after the response is received? When I run the test case when we are successful there is still a small growth in memory use which would seem to correspond to the ssl and certificate objects created for that request not being cleaned up. 

> We should figure out where that exception is being caught / handled and why that code does not close the connection and dispose everything.

I think that the exception finally gets caught by the class that started the async task so I am not sure how it would request a cleanup.

nealef:
> Let me have a look at this before merging this.

Absolutely! That's why I flagged it as Don't Merge as I thought it would take a lot more consideration given its impact.

baulig:
HttpWebRequest shouldn't close the connection as that's handled by the ServicePoint.

Sorry for not replying earlier to this, I was busy with stuff and then went on vacation from which I just came back.  I'll have a look at this asap.

nealef:
No worries. I know how busy things have been and it's also summer. 

My solution was based on an incomplete understanding of the architecture of an HttpWebRequest.

akoeplinger:
Replaced by https://github.com/mono/mono/pull/10254

Issue #8775, opened 2018-05-19
Timeout in HttpWebRequest.GetResponse() when called repeatedly


AlexRixhardson:
## Steps to Reproduce

1. Create a web app which calls HttpWebRequest.GetResponse() repeatedly
2. Run app on Mono 5.12 + Apache + CentOS 7 configuration

### Current Behavior

Calls to GetResponse() will eventually stop working and will fail with:

```
The operation has timed out., Stack Trace:   at System.Net.HttpWebRequest+<RunWithTimeoutWorker>d__244`1[T].MoveNext () [0x000ba] in <93075ab5c03242e39aa9261622b472b8>:0 
--- End of stack trace from previous location where exception was thrown ---
  at System.Net.HttpWebRequest.GetResponse () [0x00013] in <93075ab5c03242e39aa9261622b472b8>:0 
```

App does not recover. Restart of the Mono process is required.

### Expected Behavior

GetResponse() method not to fail if called repeatedly

### Sample Code

The following code must be repeatedly called for the bug to start occurring:

```
            Uri myUri = new Uri(...);
            WebRequest myWebRequest = HttpWebRequest.Create(myUri);
            HttpWebRequest request = (HttpWebRequest)myWebRequest;

            request.Method = "GET";
            request.ContentType = "application/json; charset=utf-8";

            using (Stream s = request.GetResponse().GetResponseStream())
            {
                using (StreamReader sr = new StreamReader(s))
                {
                    var jsonData = sr.ReadToEnd();
                    return;
                }
            }
```

## On which platforms did you notice this

[ ] macOS
[X ] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.12.0.233 (tarball Tue May  8 09:28:02 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          supported, not enabled.
        GC:            sgen (concurrent by default)

```

## Stacktrace

```
The operation has timed out., Stack Trace:   at System.Net.HttpWebRequest+<RunWithTimeoutWorker>d__244`1[T].MoveNext () [0x000ba] in <93075ab5c03242e39aa9261622b472b8>:0 
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #10017, opened 2018-08-10
fix timeout when poll fail with the error EINTR

EncoreJiang:
#8775 
when os is busy, calling syscall poll() will return EINTR error. In the current source code, the calculation of timeout of retrying poll() is incorrect. I supposed a better way to solve this problem.


luhenry:
@monojenkins build

luhenry:
@monojenkins build failed

PR #10019, opened 2018-08-10
[cxx] everything, now with much less line damage (many casts removed)

jaykrell:


jaykrell:
@monojenkins build

jaykrell:
Much less churn now, but something is broken maybe as a result.

PR #10020, opened 2018-08-10
Use clonefile instead of copyfile on macOS.

filipnavara:
Alternative to #10008.

Note that `copyfile` is completely user-mode API implemented in `/usr/lib/system/libcopyfile.dylib`. As you can see in the [source code](https://opensource.apple.com/source/copyfile/copyfile-146.30.2/copyfile.c.auto.html) in the `copyfile_data` method it does exactly the same read+write loop as Mono and offers no performance benefits by itself.

Someone in Apple decided to break the `copyfile` API in macOS 10.13.2 by cloning simlinks instead of targets despite the fact that passed in flags do not include `COPYFILE_NOFOLLOW_SRC`. The reason for the change is also completely bogus - we don't support copying directories, so let's not follow symlinks because they could point to directories. Well, just before calling `clonefile` they check if the source is a link, so they could have skipped cloning symlinks instead.

Since the broken `copyfile` is apparently in the wild already the only way to fix it is to drop using it for cloning. And since cloning was the only performance benefit of the API it makes sense to drop its usage entirely.

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -2338,28 +2338,25 @@ write_file (gint src_fd, gint dest_fd, struct stat *st_src, gboolean report_erro
 static int
 _wapi_clonefile(const char *from, const char *to, int flags)
 {
-	if (clonefile_ptr != NULL) {
-		gchar *located_from, *located_to;
-		int ret;
+	gchar *located_from, *located_to;
+	int ret;
 
-		located_from = mono_portability_find_file (from, FALSE);
-		located_to = mono_portability_find_file (to, FALSE);
+	g_assert (clonefile_ptr != NULL);
 
-		MONO_ENTER_GC_SAFE;
-		ret = clonefile_ptr (
-			located_from == NULL ? from : located_from,
-			located_to == NULL ? to : located_to,
-			flags);
-		MONO_EXIT_GC_SAFE;
+	located_from = mono_portability_find_file (from, FALSE);
+	located_to = mono_portability_find_file (to, FALSE);
 
-		g_free (located_from);
-		g_free (located_to);
+	MONO_ENTER_GC_SAFE;
+	ret = clonefile_ptr (
+		located_from == NULL ? from : located_from,
+		located_to == NULL ? to : located_to,
+		flags);
+	MONO_EXIT_GC_SAFE;
 
-		return ret;
-	} else {
-		errno = ENOTSUP;
-		return -1;
-	}
+	g_free (located_from);
+	g_free (located_to);
+
+	return ret;
 }
 #endif
 
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -14,12 +14,6 @@
 #ifdef HAVE_SYS_STATVFS_H
 #include <sys/statvfs.h>
 #endif
-#ifdef HAVE_COPYFILE_H
-#include <copyfile.h>
-#  if !defined(COPYFILE_CLONE)
-#    define COPYFILE_CLONE (1 << 24)
-#  endif
-#endif
 #if defined(HAVE_SYS_STATFS_H)
 #include <sys/statfs.h>
 #endif
@@ -42,6 +36,9 @@
 #include <sys/time.h>
 #ifdef HAVE_DIRENT_H
 # include <dirent.h>
+#endif
+#if __APPLE__
+#include <dlfcn.h>
 #endif
 
 #include "w32file.h"
@@ -111,6 +108,11 @@ static MonoCoopMutex file_share_mutex;
 
 static GHashTable *finds;
 static MonoCoopMutex finds_mutex;
+
+#if __APPLE__
+typedef int (*clonefile_fn) (const char *from, const char *to, int flags);
+static clonefile_fn clonefile_ptr;
+#endif
 
 static void
 time_t_to_filetime (time_t timeval, FILETIME *filetime)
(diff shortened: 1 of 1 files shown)


filipnavara:
I went through the build failures and they seem to be unrelated.

mono/metadata/w32file-unix.c
```diff
@@ -2460,21 +2434,75 @@ CopyFile (const gunichar2 *name, const gunichar2 *dest_name, gboolean fail_if_ex
...
+				mono_w32error_set_last (ERROR_ACCESS_DENIED);
+				return (FALSE);
+			}
+
+			_wapi_unlink (utf8_dest);
+		}
+#endif
+	}
+
+#if __APPLE__
```
kumpera: this should use the HOST_ defines

mono/metadata/w32file-unix.c
```diff
@@ -4885,6 +4912,10 @@ mono_w32file_init (void)
 	finds = g_hash_table_new_full (g_direct_hash, g_direct_equal, NULL, finds_remove);
 	mono_coop_mutex_init (&finds_mutex);
 
+#if __APPLE__
```
kumpera: This is should use HOST_ defines.

mono/metadata/w32file-unix.c
```diff
@@ -4885,6 +4912,10 @@ mono_w32file_init (void)
...
+	clonefile_ptr = (clonefile_fn)dlsym (RTLD_DEFAULT, "clonefile");
```
kumpera: AFAICT this won't work on iOS or watchOS. @rolfbjarne can you give us a more authoritative answer here?

mono/metadata/w32file-unix.c
```diff
@@ -43,6 +37,9 @@
 #ifdef HAVE_DIRENT_H
 # include <dirent.h>
 #endif
+#if __APPLE__
```
kumpera: Shouldn't this be a HAVE_DLFCN_H define?

mono/metadata/w32file-unix.c
kumpera: We must not use RTLD_DEFAULT as it will cause us all sorts of trouble with embeders and fullAOT scenarios.

mono/metadata/w32file-unix.c
filipnavara: Technically yes, but the assumption was that all Apple platforms have it anyway.

mono/metadata/w32file-unix.c
filipnavara: I'm fine with changing it, but there were already other pieces of code using __APPLE__, so I just followed whatever was already there.

mono/metadata/w32file-unix.c
filipnavara: I'd appreciate any suggestions on how to fix it. I assumed RTLD_DEFAULT is ok since it's a system API (libc, or libsystem respectively).

mono/metadata/w32file-unix.c
filipnavara: HOST_DARWIN?

mono/metadata/w32file-unix.c
filipnavara: Something like
```c
libc_handle = dlopen("/usr/lib/libc.dylib", RTLD_NOW);
clonefile_ptr = (clonefile_fn)dlsym (libc_handle, "clonefile");
```
?
How about iOS? I'd be fine with supporting macOS only if `dlsym` is a problem.

mono/metadata/w32file-unix.c
```diff
@@ -2331,27 +2334,32 @@ write_file (gint src_fd, gint dest_fd, struct stat *st_src, gboolean report_erro
...
 
-#if HAVE_COPYFILE_H
+#if HOST_DARWIN
 static int
-_wapi_copyfile(const char *from, const char *to, copyfile_state_t state, copyfile_flags_t flags)
+_wapi_clonefile(const char *from, const char *to, int flags)
 {
-	gchar *located_from, *located_to;
-	int ret;
+	if (clonefile_ptr != NULL) {
```
luhenry: Since you check `clonefile_ptr ` before calling `_wapi_clonefile`, you can just `g_assert(clonefile_ptr)` here.

mono/metadata/w32file-unix.c
```diff
@@ -2331,27 +2334,32 @@ write_file (gint src_fd, gint dest_fd, struct stat *st_src, gboolean report_erro
...
 
-	MONO_ENTER_GC_SAFE;
-	ret = copyfile (
-		located_from == NULL ? from : located_from,
-		located_to == NULL ? to : located_to,
-		state, flags);
-	MONO_EXIT_GC_SAFE;
+		MONO_ENTER_GC_SAFE;
+		ret = clonefile_ptr (
+			located_from == NULL ? from : located_from,
```
jaykrell: To avoid worrying about precedence I like to reverse this:
`foo ? foo : other`

But I realize this is how it was.

filipnavara: I prefer to keep it unchanged to keep the diff smaller. I'll remember your advice when writing new code.

filipnavara:
@monojenkins build failed

luhenry:
@kumpera could you please take another look at it?

mono/metadata/w32file-unix.c
```diff
@@ -4885,6 +4910,12 @@ mono_w32file_init (void)
 	finds = g_hash_table_new_full (g_direct_hash, g_direct_equal, NULL, finds_remove);
 	mono_coop_mutex_init (&finds_mutex);
 
+#if HOST_DARWIN
+	libc_handle = dlopen ("/usr/lib/libc.dylib", 0);
+	g_assert (libc_handle);
+	clonefile_ptr = (clonefile_fn)dlsym (RTLD_DEFAULT, "clonefile");
```
kumpera: We're not using `libc_handle` I don't get why are we dlopen'ing it.

filipnavara: Uh, my fault, should be used instead of RTLD_DEFAULT

luhenry:

luhenry:

PR #10021, opened 2018-08-10
[System.ServiceModel] Replace SynchronizedCollection with referencesource

akoeplinger:
This fixes a bug in the old implementation where we didn't have proper locking in `Add()`.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=43447
Closes https://github.com/mono/mono/pull/10007


akoeplinger:

PR #10022, opened 2018-08-10
Remove unnecessary Mono.Security references from corlib.

filipnavara:


mcs/class/corlib/System.Security.Policy/Evidence.cs
```diff
@@ -260,16 +260,14 @@ static internal Evidence GetDefaultHostEvidence (Assembly a)
...
-					if (x509.GetHashCode () != 0) {
-						e.AddHost (new Publisher (x509));
-					}
+				try {
+					X509Certificate x509 = X509Certificate.CreateFromSignedFile (a.Location);
+					e.AddHost (new Publisher (x509));
+				}
+				catch (CryptographicException) {
 				}
 			}
```
filipnavara: The comment about X509Certificate.CreateFromSignedFile is totally bogus. AuthenticodeDeformatter checks the signature and the result is ignored. X509Certificate.CreateFromSignedFile is supposed to only load the certificate, but not validate its chain, so in theory it's a weaker verification.

X509Certificate.CreateFromSignedFile is currently non-working in master, in the same way as it is broken in CoreFX. That's a separate issue though.

marek-safar:
This fails to build for all profiles

filipnavara:
@marek-safar Working on it, but it actually fails for monotouch_watch only now.

filipnavara:
@monojenkins build Linux x64

filipnavara:
The remaining build failures are unrelated to this PR.

filipnavara:
@monojenkins build failed

marek-safar:
@monojenkins commit csproj

mcs/class/corlib/System.Security.Policy/Evidence.cs
baulig: Let's submit that to CoreFX as we're now using their implementation.

filipnavara: @baulig I have a patch queue for CoreFX. It's a bit overflowing due to all the vacations of other people and some blocking issues, but I am aware of this and I do plan to fix it.

PR #10026, opened 2018-08-10
Fix s390x build broken by incorrect specification of the msgfi instuc…

nealef:
Fix s390x build broken by incorrect specification of the msgfi instruction used in MUL_IMM type operations. The instruction had been encoded as its 32-bit counterpart (msfi). 

In addition, the s390x microcode makes the mono_strength_reduction_division unnecessary so this can be bypassed.

The change to basic.make is just to avoid error messages when basic-profile-check.exe hasn't been built yet.


mono/mini/local-propagation.c
```diff
@@ -368,12 +368,14 @@ mono_strength_reduction_ins (MonoCompile *cfg, MonoInst *ins, const char **spec)
 		}
 		break;
 	}
+#ifndef __s390__ // s390x microcode takes care of optimizing divide operations
 	case OP_IDIV_UN_IMM:
```
vargaz: We try to avoid platform specific ifdefs in this code. Why is this needed ?

mcs/build/profiles/basic.make
```diff
@@ -122,7 +122,7 @@ $(PROFILE_EXE): $(topdir)/build/common/basic-profile-check.cs $(GENSOURCES_CS)
 	$(BOOTSTRAP_MCS) /noconfig /langversion:latest /r:mscorlib.dll /r:System.dll /r:System.Core.dll /out:$(GENSOURCES_EXE).tmp $(GENSOURCES_CS)
 	- rm $(GENSOURCES_EXE)
 	mv $(GENSOURCES_EXE).tmp $(GENSOURCES_EXE)
-	- rm $@
+	- rm -f $@
```
vargaz: Unrelated change.

nealef:
The s390x processor microcode optimizes divide instructions in much the same way that the mono_strength_reduction_division() whereas this results in 5-6 instructions. So this optimization becomes a deoptimization.

mono/mini/local-propagation.c
jaykrell: Would a runtime check be ok, like cfg->arch->optimize_divide?
i.e. is it preprocessor specifically, or any platform-conditionality, or unabstracted platform-conditionality?

jaykrell: (personally I don't like preprocessor, because it breaks future configure/compile/link once to get target-aot-all)

nealef:
How could we do something similar as:

if (!COMPILE_LLVM (cfg))

that’s used there now for platform-specific instances?

vargaz:
In this case add a flag to MonoBackend, add a MONO_ARCH_ define, and set the backend flag based on the define in init_backend () in mini.c.

nealef:
﻿Will do.

nealef:
Found an earlier instance in local_propagation of ifdef of s390. I will replace it with same logic.

mono/mini/mini.h
```diff
@@ -1150,6 +1150,7 @@ typedef struct {
 	guint            no_unaligned_access : 1;
 	guint            disable_div_with_mul : 1;
 	guint            explicit_null_checks : 1;
+	guint            optimized_div : 1;
```
jaykrell: I suggest `gboolean`, but ok.

mono/mini/local-propagation.c
```diff
@@ -383,10 +383,8 @@ mono_strength_reduction_ins (MonoCompile *cfg, MonoInst *ins, const char **spec)
 			ins->opcode = OP_ICONST;
 			MONO_INST_NULLIFY_SREGS (ins);
 			ins->inst_c0 = 0;
-#if __s390__
```
jaykrell: if you do something like this again, in any codebase, move the if the closing brace and have no else. It is much easier to read that way.

PR #10027, opened 2018-08-10
[cxx] [glib] [coop] Auto void* to non-void*, int vs. enum, and other glib fixes.

jaykrell:
Mainly this deals with a large number of void* to non-void* conversions w/o line damage all around.
 They can be fixed individually and that was done and working but was not popular.
It also provides infrastructure for int vs. enum problem, and uses it some. This  is a much smaller problem in terms of line damage but this drives churn  further down.
And void*/non-void* is addressed in handle.h.

Basically, apply `g_cast` liberally around compilation errors.

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -54,6 +54,248 @@
 #define G_END_DECLS
 #endif
 
+#ifdef __cplusplus
+
+#define g_cast_t monoeg_g_cast_t // in case not inlined (see eglib-remap.h)
+
+// Mostly-internal type to support g_cast(), but also can be used in containers.
+template <typename TFrom> // TFrom is almost always void*
+struct g_cast_t
+{
+private:
+	TFrom x;
+public:
+	explicit g_cast_t (TFrom y) : x(y)
+	{
+	}
+
+	g_cast_t () // Not really intended but has a user.
+	{
+	}
+
+	g_cast_t (const g_cast_t& y) : x(y.x)
+	{
+	}
+
+	template <typename TTo>
+	operator TTo()
+	{
+		return (TTo)x;
+	}
+
+	g_cast_t& operator= (TFrom y)
+	{
+		x = y;
+		return *this;
+	}
+
+	template <typename T>
+	bool operator != (T y) // e.g. compare to NULL
+	{
+		return x != (TFrom)y;
+	}
+
+	template <typename T>
+	bool operator == (T y) // e.g. compare to NULL
+	{
+		return x == (TFrom)y;
+	}
+
+	template <typename T>
+	bool operator < (T y)
+	{
+		return x < (TFrom)y;
+	}
+
+	template <typename T>
+	bool operator <= (T y)
+	{
+		return x <= (TFrom)y;
+	}
+
+	template <typename T>
+	bool operator > (T y)
+	{
+		return x > (TFrom)y;
+	}
+
+	template <typename T>
+	bool operator >= (T y)
+	{
+		return x >= (TFrom)y;
+	}
+};
+
+// Member function compares g_cast_t to other, global compares other to g_cast_t,
+// i.e. which parameter is first vs. second.
+template <typename T1, typename T2>
+inline bool operator == (T1 x, g_cast_t<T2> y) // e.g. compare to NULL
+{
+	return (T2)x == y.x;
+}
+
+template <typename T1, typename T2>
+inline bool operator != (T1 x, g_cast_t<T2> y) // e.g. compare to NULL
+{
+	return (T2)x != y.x;
+}
+
+template <typename T1, typename T2>
+inline bool operator < (T1 x, g_cast_t<T2> y) // e.g. compare to NULL
+{
+	return (T2)x < y.x;
+}
+
+template <typename T1, typename T2>
+inline bool operator <= (T1 x, g_cast_t<T2> y)
+{
+	return (T2)x <= y.x;
+}
+
+template <typename T1, typename T2>
+inline bool operator > (T1 x, g_cast_t<T2> y)
+{
+	return (T2)x > y.x;
+}
+
+template <typename T1, typename T2>
+inline bool operator >= (T1 x, g_cast_t<T2> y)
+{
+	return (T2)x >= y.x;
+}
+
+/*
+Apply g_cast() around expressions that C will convert
+to whatever is on the left of an assignment or return.
+
+The overwhelming most common case is void* to non-void*
+such as from malloc. Most of these can be handled
+in central places, such as:
+
+#define malloc(x) g_cast (malloc (x))
+
+The pattern can also be applied to container of void* such as GSList.
+
+*/
+
+// C++11:
+//define g_cast(x) (g_cast_t<decltype(x)>(x))
+
+// C++98:
+#define g_cast monoeg_g_cast // in case not inlined (see eglib-remap.h)
+
+template <typename TFrom>
+inline g_cast_t<TFrom>
+g_cast (TFrom from)
+{
+	return g_cast_t<TFrom> (from);
+}
+
+#else
+
+// FIXME? Parens are omitted to preserve prior meaning.
+#define g_cast(x) x
+
+#endif
+
+#ifdef __cplusplus
+
+/*
+Provide for math on enums.
+This alleviates a fair number of casts in porting C to C++.
+*/
+#define G_ENUM_FUNCTIONS(Enum)			\
+extern "C++" { /* in case within extern "C" */	\
+inline Enum					\
+operator~ (Enum a)				\
+{						\
+	return (Enum)~(int)a;			\
+}						\
+						\
+inline Enum					\
+operator| (Enum a, Enum b)			\
+{						\
+	return (Enum)((int)a | (int)b);		\
+}						\
+						\
+inline Enum					\
+operator& (Enum a, Enum b)			\
+{						\
+	return (Enum)((int)a & (int)b);		\
+}						\
+						\
+inline Enum&					\
+operator|= (Enum& a, Enum b)			\
+{						\
+	return a = (Enum)((int)a | (int)b);	\
+}						\
+						\
+inline Enum&					\
+operator&= (Enum& a, Enum b)			\
+{						\
+	return a = (Enum)((int)a & (int)b);	\
+}						\
+						\
+inline Enum					\
+operator^ (Enum a, Enum b)			\
+{						\
+	return (Enum)((int)a ^ (int)b);		\
+}						\
+						\
+inline Enum					\
+operator- (Enum a, Enum b)			\
+{						\
+	return (Enum)((int)a - (int)b);		\
+}						\
+						\
+inline Enum					\
+operator+ (Enum a, Enum b)			\
+{						\
+	return (Enum)((int)a + (int)b);		\
+}						\
+						\
+inline Enum					\
+operator+ (Enum a, int b)			\
+{						\
+	return (Enum)((int)a + b);		\
+}						\
+						\
+inline Enum					\
+operator+ (Enum a, unsigned b)			\
+{						\
+	return (Enum)((unsigned long)a + b);	\
+}						\
+						\
+inline Enum					\
+operator+ (Enum a, unsigned long b)		\
+{						\
+	return (Enum)((unsigned long)a + b);	\
+}						\
+						\
+inline Enum					\
+operator+ (Enum a, unsigned long long b)	\
+{						\
+	return (Enum)((unsigned long long)a + b); \
+}						\
+} /* extern "C++" */				\
+
+#else
+
+#define G_ENUM_FUNCTIONS(Enum) /* nothing */
+
+#endif
+
+#ifdef __cplusplus
+
+// Overload strcmp to support common usage.
+inline int
+strcmp (const void *a, const void* b)
+{
+	return strcmp ((const char*)a, (const char*)b);
+}
+
+#endif
+
 G_BEGIN_DECLS
 
 /*
@@ -161,7 +403,7 @@ gpointer g_try_realloc (gpointer obj, gsize size);
 
 #define g_memmove(dest,src,len) memmove (dest, src, len)
 #define g_renew(struct_type, mem, n_structs) ((struct_type*)g_realloc (mem, sizeof (struct_type) * n_structs))
-#define g_alloca(size)		alloca (size)
+#define g_alloca(size)		(g_cast (alloca (size)))
 
 gpointer g_memdup (gconstpointer mem, guint byte_size);
 static inline gchar   *g_strdup (const gchar *str) { if (str) { return (gchar*) g_memdup (str, (guint)strlen (str) + 1); } return NULL; }
@@ -329,7 +571,13 @@ typedef void     (*GFreeFunc)      (gpointer       data);
  */
 typedef struct _GSList GSList;
 struct _GSList {
+#ifdef __cplusplus
+	// Allow for conversion of list->data to anything, in context.
+	// This saves adding casts to turn valid C into valid C++.
+	g_cast_t<gpointer> data;
+#else
 	gpointer data;
+#endif
 	GSList *next;
 };
 
@@ -382,7 +630,13 @@ gpointer g_slist_nth_data     (GSList	     *list,
 
 typedef struct _GList GList;
 struct _GList {
+#ifdef __cplusplus
+  // Allow for conversion of list->data to anything, in context.
+  // This saves adding casts to turn valid C into valid C++.
+  g_cast_t<gpointer> data;
+#else
   gpointer data;
+#endif
   GList *next;
   GList *prev;
 };
@@ -517,7 +771,10 @@ void    g_array_set_size          (GArray *array, gint length);
 
 #define g_array_append_val(a,v)   (g_array_append_vals((a),&(v),1))
 #define g_array_insert_val(a,i,v) (g_array_insert_vals((a),(i),&(v),1))
-#define g_array_index(a,t,i)      *(t*)(((a)->data) + sizeof(t) * (i))
+
+// FIXME This can probably be g_cast_ref that provides operator&.
+#define g_array_index_ref(a,t,i) (*(t*)(((a)->data) + sizeof(t) * (i))) // for taking address
+#define g_array_index(a,t,i)      (g_cast (g_array_index_ref ((a), t, (i))))
 
 /*
  * QSort
@@ -548,7 +805,7 @@ void       g_ptr_array_set_size           (GPtrArray *array, gint length);
 gpointer  *g_ptr_array_free               (GPtrArray *array, gboolean free_seg);
 void       g_ptr_array_foreach            (GPtrArray *array, GFunc func, gpointer user_data);
 guint      g_ptr_array_capacity           (GPtrArray *array);
-#define    g_ptr_array_index(array,index) (array)->pdata[(index)]
+#define    g_ptr_array_index(array,index) (g_cast ((array)->pdata[(index)]))
 
 /*
  * Queues
@@ -589,6 +846,8 @@ typedef enum {
 	
 	G_LOG_LEVEL_MASK              = ~(G_LOG_FLAG_RECURSION | G_LOG_FLAG_FATAL)
 } GLogLevelFlags;
+
+G_ENUM_FUNCTIONS (GLogLevelFlags)
 
 void           g_printv               (const gchar *format, va_list args);
 void           g_print                (const gchar *format, ...);
@@ -922,6 +1181,7 @@ typedef enum {
 	G_FILE_TEST_EXISTS = 1 << 4
 } GFileTest;
 
+G_ENUM_FUNCTIONS (GFileTest)
 
 gboolean   g_file_set_contents (const gchar *filename, const gchar *contents, gssize length, GError **gerror);
 gboolean   g_file_get_contents (const gchar *filename, gchar **contents, gsize *length, GError **gerror);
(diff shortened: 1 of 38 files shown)


mono/eglib/gunicode.c
```diff
@@ -88,7 +88,7 @@ g_unichar_type (gunichar c)
 	/* since the argument is UTF-16, we cannot check beyond FFFF */
 
 	/* It should match any of above */
-	return 0;
+	return g_cast (0); // G_UNICODE_CONTROL
```
lambdageek: This seems like an abuse of `g_cast`.  You have to write something here and all the types are known, so just write the explicit cast.

mono/eglib/glib.h
```diff
@@ -1148,4 +1408,24 @@ glong     g_utf8_pointer_to_offset (const gchar *str, const gchar *pos);
...
+#undef g_memdup
+#undef g_hash_table_lookup
+#define g_malloc(x) (g_cast (monoeg_malloc (x)))
+#define g_realloc(obj, size) (g_cast (monoeg_realloc ((obj), (size))))
+#define g_malloc0(x) (g_cast (monoeg_malloc0 (x)))
+#define g_calloc(x, y) (g_cast (monoeg_g_calloc ((x), (y))))
+#define g_try_malloc(x) (g_cast (monoeg_try_malloc (x)))
+#define g_try_realloc(obj, size) (g_cast (monoeg_try_realloc ((obj), (size))))
+#define g_memdup(mem, size) (g_cast (monoeg_g_memdup ((mem), (size))))
+#define g_hash_table_lookup(hash, key) (g_cast (monoeg_g_hash_table_lookup ((hash), (key))))
```
jaykrell: See for example

https://github.com/mono/mono/pull/10019/files#diff-98cf043c47e9ac369f935d88ea2c0e5dR517

https://github.com/mono/mono/pull/10019/files#diff-34ddcfc99bebbe2859c2bafcd20a2ac4R1320

for the slightly larger use.

mono/eglib/glib.h
```diff
@@ -329,7 +571,13 @@ typedef void     (*GFreeFunc)      (gpointer       data);
  */
 typedef struct _GSList GSList;
 struct _GSList {
+#ifdef __cplusplus
+	// Allow for conversion of list->data to anything, in context.
+	// This saves adding casts to turn valid C into valid C++.
+	g_cast_t<gpointer> data;
```
lambdageek: This makes `_GSList` not a POD anymore.  Is this where the non-`explicit` default constructor for `g_cast_t<>` is used?

mono/eglib/gqsort.c
```diff
@@ -40,14 +40,14 @@ typedef struct _QSortStack {
 	size_t count;
 } QSortStack;
 
-#define QSORT_PUSH(sp, a, c) (sp->array = a, sp->count = c, sp++)
+#define QSORT_PUSH(sp, a, c) (sp->array = g_cast (a), sp->count = c, sp++)
```
lambdageek: the result type is known, can just use a normal explicit cast to `(char*)`

mono/eglib/gqsort.c
```diff
@@ -40,14 +40,14 @@ typedef struct _QSortStack {
...
 #define QSORT_POP(sp, a, c) (sp--, a = sp->array, c = sp->count)
 
 #define SWAPTYPE(TYPE, a, b) {              \
-	long __n = size / sizeof (TYPE);    \
-	register TYPE *__a = (TYPE *) (a);  \
-	register TYPE *__b = (TYPE *) (b);  \
-	register TYPE t;                    \
+	gssize __n = size / sizeof (TYPE);  \
+	TYPE *__a = (TYPE *) (a);           \
```
lambdageek: What's wrong with `register`?

mono/eglib/gqsort.c
```diff
@@ -69,7 +69,7 @@ void
 g_qsort_with_data (gpointer base, size_t nmemb, size_t size, GCompareDataFunc compare, gpointer user_data)
 {
 	QSortStack stack[STACK_SIZE], *sp;
-	register char *i, *k, *mid;
+	char *i, *k, *mid;
```
lambdageek: Is removing `register` motivated by the C++ project or unrelated?

mono/eglib/gunicode.c
```diff
@@ -62,7 +62,7 @@ g_unichar_type (gunichar c)
 			continue;
 		if (unicode_category_ranges [i].end <= cp)
 			continue;
-		return unicode_category [i] [cp - unicode_category_ranges [i].start];
+		return g_cast (unicode_category [i] [cp - unicode_category_ranges [i].start]);
```
lambdageek: The result type is known - just use an explicit cast to `(GUnicodeType)`

mono/eglib/glib.h
```diff
@@ -54,6 +54,248 @@
...
+struct g_cast_t
+{
+private:
+	TFrom x;
+public:
+	explicit g_cast_t (TFrom y) : x(y)
+	{
+	}
+
+	g_cast_t () // Not really intended but has a user.
```
lambdageek: If I understand what this is for (to add `g_cast_t<>` data members to previously C-compatible structs), I think we're better off without it.

mono/eglib/glib.h
```diff
@@ -54,6 +54,248 @@
...
+	bool operator >= (T y)
+	{
+		return x >= (TFrom)y;
+	}
+};
+
+// Member function compares g_cast_t to other, global compares other to g_cast_t,
+// i.e. which parameter is first vs. second.
+template <typename T1, typename T2>
+inline bool operator == (T1 x, g_cast_t<T2> y) // e.g. compare to NULL
```
lambdageek: First: `x` is `private`, so these all need to be `friend`s.

Second: this is a really really bad idea.

You're essentially throwing away all type safety (because `T1` and `T2` may have nothing to do with each other) so this will mask a ton of unintended bugs:

```cplusplus
int* f (void)
{
  return 0;
}

int
main ()
{
  g_cast_t<int*> z = g_cast ((int*)0);
  std::printf ("%d\n", f () == z);
  std::printf ("%d\n", f == z);
}
```

So in this case I accidentally wrote `f` instead of `f()` on the last line, and the answer changes from always true to always false without a warning (even with `-Wall -Wextra`).

mono/eglib/glib.h
jaykrell: This is not where it is used. There is a local variable out in the tree. I tried other fixed.

mono/eglib/gqsort.c
jaykrell: g_cast becomes its own philosophy: no change to the C.

mono/eglib/glib.h
jaykrell: ‘Private’ was a very late addition. Probably a mistake. Let’s just remove it.

mono/eglib/gqsort.c
jaykrell: But ok either way.

mono/eglib/glib.h
lambdageek: I'm confused/surprised how there can be a local variable of `g_cast_t<>` type. It's a brand new type.

mono/eglib/gqsort.c
jaykrell: ‘Register’ is a C++ warning at least, maybe error now or in future.

mono/eglib/gqsort.c
jaykrell: C++ warning due to deprecation.

mono/eglib/glib.h
jaykrell: Local of ‘GSList’ or ‘GList’.

mono/eglib/glib.h
jaykrell: Type safety is not a goal here yet. Limiting line damage and making the code valid C & C++ & with same meaning.

But if you want TFrom only void* (or MonoObject*? Or any pointer?) and TTo only a pointer, then yes, agreed, I can try.

mono/eglib/glib.h
jaykrell: It avoids casts in uses of lists. For actual line damage reduction not to provide a well designed type. Same thing with most of the operators, including both orders. They might not all be used, but quite a few are. Including ordering, not just equality.

mono/eglib/gunicode.c
jaykrell: “No change to C.” Useful approach?

mono/eglib/glib.h
jaykrell: But this also can help with function pointers fyi. But maybe fix those otherwise.

jaykrell:
Higher level obvious question: add casts everywhere? It is tractable. I had it all done for x86/amd64/arm/arm64. Lots of green CI lanes.

mono/eglib/gqsort.c
lambdageek: if we really want no change to C, we should do:
```
#if defined (__cplusplus)
#define MONO_CXX_CAST(T,X) ((T)(X))
#else
#define MONO_CXX_CAST(T,X) (X)
#endif
```

Using `g_cast` in cases where we can use a simpler workaround seems unnecessary.

lambdageek:
@jaykrell The goal of `g_cast_t<>` should not be to eliminate "line damage", IMO.  It should be to solve a very specific problem that we have: there are a few places where the existing C code casts from `void*` to some non-`void*` type that isn't written down.  (ie the return type in the `HANDLE_FUNCTION_RETURN_OBJ` macro).  We should only use `g_cast_t<>` when we have no other choice.  It shouldn't have any operators or default- or copy-constructors.  It should solve one single problem for which we have no other workaround.

mono/eglib/glib.h
jaykrell: POD-ness should not affect layout. Perhaps ABI of pass/return-by-value, which should be rare/zero. We can cast in list uses instead. It isn’t as much as allocators. Also C/C++ should be pretty rare, that traffic in glib, if we convert utils/metadata/sgen/mini, as I have.

mono/eglib/gqsort.c
jaykrell: ‘g_cast’ is also more mechanical consisent easier to read write. And most of all, easy to find later & improve. Explicit casts will be harder to find. So maybe a separate macro as you show? Some of these probably ok indefinitely but some (not necessarily here) less so. Let me fix enum and sort and we can look at others later.

jaykrell:
There is workarounds for all of it, and a matter of degree. Auto is good for handle, and casts everywhere else, imho. 

But the casts drew objection, esp. for allocations.

I also think we could view this as temporary. Agree to cast in the long term? It is one solid tedious day & I already did it, but it is noisy, & we are not (yet) commited to C++.

jaykrell:
I can try also to address the cases that need operators, but you might be surprised — comparison to ‘NULL’ gets here, and its type was ‘long’. That was why I first added operator== and !=. After I added them, I have no record of number of later uses.

jaykrell:
The goal really was exactly line damage reduction. For handle, auto is good trivial C++11, or there are other options. Probably everywhere else knows the type and a local cast is easy, albeit very frequent.

jaykrell:
Ok. I'm down to:

```
#define g_cast monoeg_g_cast // in case not inlined (see eglib-remap.h)

// g_cast converts void* to T*.
// e.g. #define malloc(x) (g_cast (malloc (x)))
// FIXME It used to do more. Rename?

struct g_cast
{
private:
	void * const x;
public:
	explicit g_cast (void *y) : x(y) { }
	g_cast (g_cast&& y) : x(y.x) { }
	g_cast () = delete;
	g_cast (const g_cast& y) = delete;

	template <typename TTo>
	operator TTo* () const
	{
		return (TTo*)x;
	}
};

#else

// FIXME? Parens are omitted to preserve prior meaning.
#define g_cast(x) x

#endif
```

The rvalue constructor is used by two places using ternary operator.
This can be fixed if needed, by inserting casts or not using ternary. It is probably ok.
List traversal, TLS, various trampolines, hashtable lookup, I all changed back to casts.
Leaving this just for "allocators", more than is in this PR.

Will update this PR later.

jaykrell:
See https://github.com/mono/mono/commit/d0c43ac207358e6faf451a0c56b39cbe340f6ebe
for an idea of what line damage this traded for.

mono/eglib/gqsort.c
jaykrell: `long` there was insidious -- having the same-ish meaning on all but one platform -- win64.

mono/eglib/gqsort.c
jaykrell: I'm not against `MONO_CXX_CAST` but really we can probably add casts in C.

jaykrell:
The level of type unsafety was pretty deliberate. You request it by saying “cast” or by using a list of void*. To shoehorn/bootstrap C into C++. One might argue though that was an opening for worse, and I used it as such, from any to any. It now (soon) narrower, from void* to pointer so I think will be more satisfactory.

mono/eglib/glist.c
```diff
@@ -306,7 +306,7 @@ gpointer
 g_list_nth_data (GList *list, guint n)
 {
 	GList *node = g_list_nth (list, n);
-	return node ? node->data : NULL;
+	return node ? node->data : (gpointer)NULL;
```
jaykrell: Leftover, undo.

mono/eglib/gslist.c
```diff
@@ -347,7 +347,7 @@ gpointer
 g_slist_nth_data (GSList *list, guint n)
 {
 	GSList *node = g_slist_nth (list, n);
-	return node ? node->data : NULL;
+	return node ? node->data : (gpointer)NULL;
```
jaykrell: Leftover from earlier, undo.

mono/mini/mini-amd64.c
```diff
@@ -7562,7 +7562,7 @@ mono_arch_emit_exceptions (MonoCompile *cfg)
 #endif /* DISABLE_JIT */
 
 void*
-mono_arch_instrument_prolog (MonoCompile *cfg, void *func, void *p, gboolean enable_arguments)
+(mono_arch_instrument_prolog) (MonoCompile *cfg, void *func, void *p, gboolean enable_arguments)
```
jaykrell: Stretching the definition of "allocator", perhaps, but it a void*-returning function that users assign result to non-void* (guint8*). We have a mix of void* and guint8*.

jaykrell:
Fyi, some of the operators come into play when you expand this to hashtable lookup:

```
custom-attrs.c:2578:86: error: invalid operands to binary expression
      ('monoeg_g_cast' and 'long')
  ...g_hash_table_lookup (image->weak_field_indexes, GINT_TO_POINTER (field_idx)) != NULL;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~
```

"NULL" is of type long.

Longer term, the hashtables can templatized and only return an exact more derived type, and then no need for casting.

jaykrell:
Also limit of gcc4.4/Cento6 is quick to reach here.

jaykrell:

jaykrell:

jaykrell:
@monojenkins build OS X x64

jaykrell:
Should be better now.

This stuff is a bit touchy, in that if you extend it to hashtable lookup, you need more functions, like compare to long. And if you add much, you exceed g++4.4 C++11 support.

But this should be ok.

The rvalue constructor can be removed if the two ternary operators out in the tree are tweaked, to cast or not use ternary.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build Linux i386

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X x64

lambdageek:
I like this version of `g_cast_t<>` much better.  it does a lot fewer things.

mono/eglib/glib.h
```diff
@@ -54,6 +54,122 @@
...
+// g_cast converts void* to T*.
+// e.g. #define malloc(x) (g_cast (malloc (x)))
+// FIXME It used to do more. Rename?
+struct g_cast
+{
+private:
+	void * const x;
+public:
+	explicit g_cast (void *y) : x(y) { }
+	g_cast (g_cast&& y) : x(y.x) { }
```
jaykrell: This line can be changed to the mildly preferred `= delete` if two occurences of ternary operator in the tree gain a cast or lose the ternary operator.

lambdageek: Let's fix the ternaries and put `= delete` here.

mono/mini/mini-amd64.c
lambdageek: This function is also defined in `mini-sparc.c`, `mini-s390x.c`, `mini-mips.c`, etc.

My preference would be not to fixup `mono_arch_instrument_prolog`/epilog in this PR.  Let's just get `g_cast_t<>` merged in the cases where we need it.

mono/eglib/gmodule.h
```diff
@@ -20,6 +20,9 @@ typedef enum {
 	G_MODULE_BIND_LOCAL = 0x02,
 	G_MODULE_BIND_MASK = 0x03
 } GModuleFlags;
+
+G_ENUM_FUNCTIONS (GModuleFlags)
```
lambdageek: I wonder if each of the uses of `G_ENUM_FUNCTIONS` should come with a "fixme" note.

jaykrell:
One subtlety I missed that Aleksey pointed out, is "which operators?".
Windows.h doesn't provide +/-/++, "only bit stuff".
Though there is the a & (a - 1) idiom like to see if only one bit set or something.

Maybe we have too many here. Maybe it doesn't matter.
In C all the operators are available. We use +/- for wait results I think, and ++ for stop-the-world phases. Though could get casts though, if we want fewer operators.

Everything can get casts, really.
Also ints might be appropriate.
Though, I an see where this goes -- ints are too weak. You want your own little ranges/units of numbers for different things, and don't accidentally mix them. Inches can be added to feet, but not to square inches, etc. stop-the-world phase is different than a wait result.

jaykrell:
See https://github.com/mono/mono/commit/3a3607faa8827a2ac994cd7385508b2beb10f11a for some of the flip side of the enum operators.

mono/eglib/glib.h
```diff
@@ -518,6 +554,7 @@ void    g_array_set_size          (GArray *array, gint length);
 #define g_array_append_val(a,v)   (g_array_append_vals((a),&(v),1))
 #define g_array_insert_val(a,i,v) (g_array_insert_vals((a),(i),&(v),1))
 #define g_array_index(a,t,i)      *(t*)(((a)->data) + sizeof(t) * (i))
+//FIXME previous missing parens
```
luhenry: Remove that.

mono/eglib/glib.h
```diff
@@ -549,6 +586,7 @@ gpointer  *g_ptr_array_free               (GPtrArray *array, gboolean free_seg);
 void       g_ptr_array_foreach            (GPtrArray *array, GFunc func, gpointer user_data);
 guint      g_ptr_array_capacity           (GPtrArray *array);
 #define    g_ptr_array_index(array,index) (array)->pdata[(index)]
+//FIXME previous missing parens
```
luhenry: Remove that.

mono/eglib/glib.h
```diff
@@ -1148,4 +1186,22 @@ glong     g_utf8_pointer_to_offset (const gchar *str, const gchar *pos);
...
+// Macros do not recurse, so naming function and macro the same is ok.
+// However these are also already macros.
+#undef g_malloc
+#undef g_realloc
+#undef g_malloc0
+#undef g_calloc
+#undef g_try_malloc
+#undef g_try_realloc
+#undef g_memdup
+#define g_malloc(x) (g_cast (monoeg_malloc (x)))
```
luhenry: Just declare that at https://github.com/mono/mono/pull/10027/files#diff-b8c6c2ba344450884373fb00f33685adR186 instead, and rename the `g_malloc` function to `g_malloc_internal` if it collides.

mono/metadata/handle.h
```diff
@@ -245,7 +245,7 @@ Icall macros
 	do {							\
 		void* __result = MONO_HANDLE_RAW (HANDLE);	\
 		CLEAR_ICALL_FRAME;				\
-		return __result;				\
+		return g_cast (__result);			\
```
lambdageek: @luhenry this use of `g_cast` in `HANDLE_FUNCTION_RETURN_OBJ` is difficult to avoid unless we change the macro to take a return type.

Instead of this:

```c
MONO_RT_EXTERNAL_ONLY
MonoString *
some_api_function (const char *s)
{
  HANDLE_FUNCTION_ENTER();
  MonoStringHandle h = mono_string_new_handle (s);
  HANDLE_FUNCTION_RETURN_OBJ (h);
}
```

We would have to write

```c
MonoString *
some_api_function (const char *s)
{
  HANDLE_FUNCTION_ENTER ();
  MonoStringHandle h = mono_string_new_handle (s);
  HANDLE_FUNCTION_RETURN_OBJ (MonoString, h);
}
```

mono/eglib/glib.h
jaykrell: And then breakpoints on `g_malloc` no longer work.

mono/metadata/handle.h
jaykrell: This one can use `auto` instead. Two of the handle.h changes can be `auto`, the other one I didn't look at closely. Because the type usually matches from `MONO_HANDLE_RAW`.

jaykrell:
I'm torn on autocast or not. I originally had everything working with explicit casts. But there were many and it bothered people.

mono/eglib/glib.h
jaykrell: Er, breakpoint is `monoeg_malloc`. I guess just break on `malloc`, though sometimes the extra layer is nice, i.e. to break only on mono's calls.

jaykrell:

Issue #9959, opened 2018-08-08
sgen-regular-ms-simple.sgen-new-threads-collect.exe_timedout and sgen-regular-ms-simple.sgen-new-threads-collect.exe_timedout


jaykrell:
## Steps to Reproduce

1. Submit a PR.

### Current Behavior

Failures, for example:

```

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/13230/


MonoTests.test-sgen-regular-ms-simple.sgen-new-threads-collect.exe_timedout


https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/13230/testReport/junit/MonoTests/test-sgen-regular-ms-simple/sgen_new_threads_collect_exe_timedout/



MonoTests.test-sgen-regular-ms-simple.sgen-new-threads-collect.exe_timedout

Failing for the past 1 build (Since Unstable#13230 )
Took 15 min.
Stacktrace
                                                MESSAGE:
                                                (lldb) command source -s 0 '/var/folders/z4/ymx__0095rlclmf1p5pscfy00000gn/T/tmp11ee83a0.tmp'
Executing commands in '/var/folders/z4/ymx__0095rlclmf1p5pscfy00000gn/T/tmp11ee83a0.tmp'.
(lldb) process attach --pid 57872

                                                +++++++++++++++++++
                                                STACK TRACE:
                                                error: attach failed: unable to attach

MonoTests.test-sgen-regular-ms-conc-simple.sgen-new-threads-collect.exe_timedout


https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/13230/testReport/junit/MonoTests/test-sgen-regular-ms-conc-simple/sgen_new_threads_collect_exe_timedout/


MonoTests.test-sgen-regular-ms-conc-simple.sgen-new-threads-collect.exe_timedout

Failing for the past 1 build (Since Unstable#13230 )
Took 15 min.
Stacktrace
                                                MESSAGE:
                                                (lldb) command source -s 0 '/var/folders/z4/ymx__0095rlclmf1p5pscfy00000gn/T/tmp33173c59.tmp'
Executing commands in '/var/folders/z4/ymx__0095rlclmf1p5pscfy00000gn/T/tmp33173c59.tmp'.
(lldb) process attach --pid 75237

                                                +++++++++++++++++++
                                                STACK TRACE:
                                                error: attach failed: unable to attach

```

### Expected Behavior

Success.

## On which platforms did you notice this

[...]

Issue #9921, opened 2018-08-06
Hybrid/coop suspend: We need a wrapper for malloc


lambdageek:
# Summary of the problem

We don't know what third-party code will take locks.  Where third-party includes things like "libc" and "the kernel".

In order for hybrid and cooperative suspend to work correctly without deadlock, logically suspended threads (that is, threads that the suspend initiator perceives as suspended) must not prevent any running thread from being able to reach a safepoint.

One way to prevent a running thread from being able to reach a safepoint is by taking a lock that the running thread needs.

We don't know where the locks are.  They are inside `malloc` and `free`, for example.

If thread 1 is the GC initiator, thread 2 is suspended while calling `malloc`, and thread 3 is running (and would be cooperatively suspended - that is, only when it reaches a safepoint) and also calls `malloc` we will get a deadlock: 

* thread 1 is waiting for thread 3 to suspend;
* thread 2 is waiting for thread 1 to wake it up;
* thread 3 is waiting for the lock held by thread 2.

# Possible solutions

## Option 1: always call via glib; wrap all library calls in GC Safe transitions

Never call `malloc`, always call `g_malloc`.  Same for every other 3rd party function that Mono calls directly from inside the runtime.  (P/Invokes automatically get GC Safe transitions in the wrapper - nothing to do there.)

```c
  inline void* g_malloc (size_t size)
  {
    void *result;
    MONO_ENTER_GC_SAFE;
    result = malloc (size);
    MONO_EXIT_GC_SAFE;
    return result;
  }
```

### Issues

* We will need some way to automatically check that we never call outside the runtime.  Maybe by doing a special build with a dynamically linked glib and without linking metadata/, mini/ and utils/ with libc.
* AOT Compiler and early startup code: the `MONO_ENTER_GC_SAFE` macro asserts if the current thread isn't attached.  We'll need variants `MONO_TRY_ENTER_GC_SAFE` that are usable when the thread state machinery isn't initialized yet (in which case it should be a no-op instead of an assertion).
* Unknown how expensive this will be.  It's every library and syscall in Mono...

## Option 2: Two-phase hybrid suspend

Motivating observation: this is a problem because we preemptively suspended some threads before we cooperatively suspend some others.  The cooperative suspend should get a chance to happen first.

The proposed solution is outlined in a comment below https://github.com/mono/mono/issues/9921#issuecomment-411897101

## Option 3: ???

We should brainstorm some other options.

[...]

PR #10028, opened 2018-08-10
[ci] Disable sgen-new-threads-collect.exe until #9959 is addressed

lambdageek:
The test times out due to an issue in hyrbid suspend.  Turn off the test
temporarily while we work on a solution.

https://github.com/mono/mono/issues/9959
https://github.com/mono/mono/issues/9921


lambdageek:

akoeplinger:
Going to merge this right away to unblock PRs.

Issue #9989, opened 2018-08-08
Incorrect rounding for double-precision number that is 0.5 + epsilon


athas:
## Steps to Reproduce

Use this program, which constructs the smallest double that is larger than 0.5 and rounds it.

```
using System;

class MainClass {
    const string formatter = "{0,25:E16}{1,23:X16}";
    public static void DoubleToLongBits( double argument )
    {
        long longValue;
        longValue = BitConverter.DoubleToInt64Bits( argument );

        // Display the resulting long in hexadecimal.
        Console.WriteLine( formatter, argument, longValue );
    }

  public static void Main (string[] args) {
    double x = 0.5;
    double y = 0.5000000000000001;
    DoubleToLongBits(x);
    DoubleToLongBits(y);
    DoubleToLongBits(Math.Round(y));
  }
}
```

### Current Behavior

0.5000000000000001 is rounded to 0.  The output is:

```
  5.0000000000000000E-001       3FE0000000000000
  5.0000000000000011E-001       3FE0000000000001
  0.0000000000000000E+000       0000000000000000
```

### Expected Behavior

0.5000000000000001 should be rounded to 1.  This is also what the CLR does (as well as Python, C, and all other languages I tried).

## On which platforms did you notice this

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:  5.14.0.177 (also [4.0.4.0
](https://repl.it/repls/DualFailingWireframes))

Issue #7737, opened 2018-03-21
CoreFX System namespace


marek-safar:
The namespace contains a lot of types and some are complex to replace.

## Milestone I

- [x] AccessViolationException
- [x] Action
- [x] Action`1-8
- [x] AggregateException
- [x] AppDomainUnloadedException
- [x] ApplicationException
- [x] ApplicationId
- [x] ArgumentException
- [x] ArgumentNullException
- [x] ArgumentOutOfRangeException
- [x] ArithmeticException
- [x] Array
- [x] ArraySegment`1
- [x] ArrayTypeMismatchException
- [x] AssemblyLoadEventArgs
- [x] AssemblyLoadEventHandler
- [x] AsyncCallback
- [x] AttributeTargets
- [x] AttributeUsageAttribute
- [x] BadImageFormatException
- [x] Base64FormattingOptions
- [x] BitConverter
- [x] Boolean
- [x] Byte
- [x] CannotUnloadAppDomainException
- [x] Char
- [x] CharEnumerator
- [x] CLSCompliantAttribute
- [x] ConsoleCancelEventArgs
- [x] ConsoleCancelEventHandler
- [x] ConsoleColor
- [x] ConsoleKey
- [x] ConsoleKeyInfo
- [x] ConsoleModifiers
- [x] ConsoleSpecialKey
- [x] Convert
- [x] DataMisalignedException
- [x] DateTime
- [x] DateTimeKind
- [x] DateTimeOffset
- [x] DayOfWeek
- [x] DBNull
- [x] Decimal
- [x] DivideByZeroException
- [x] DllNotFoundException
- [x] Double
- [x] DuplicateWaitObjectException
- [x] EntryPointNotFoundException
- [x] EventArgs
- [x] EventHandler
- [x] EventHandler`1
- [x] ExecutionEngineException
- [x] FieldAccessException
- [x] FlagsAttribute
- [x] FormatException
- [x] FormattableString
- [x] Func`1-8
- [x] Guid
- [x] HashCode
- [x] IAppDomainSetup
- [x] IAsyncResult
- [x] ICloneable
- [x] IComparable
- [x] IComparable`1
- [x] IConvertible
- [x] ICustomFormatter
- [x] IDisposable
- [x] IEquatable`1
- [x] IFormatProvider
- [x] IFormattable
- [x] IndexOutOfRangeException
- [x] InsufficientExecutionStackException
- [x] InsufficientMemoryException
- [x] Int16
- [x] Int32
[...]

PR #10033, opened 2018-08-10
[corlib] Math from CoreFX

EgorBo:
Replaces reference source implementation with CoreFX
(Math.CoreCLR is copied from coreclr repo).
Also +2k math tests from CoreFX.
There are a couple of differences, e.g.:
```csharp
Console.WriteLine(Math.Pow(double.NaN, 0));
Console.WriteLine(Math.Pow(1, double.NaN));
Console.WriteLine(Math.Pow(-1, double.PositiveInfinity));
Console.WriteLine(Math.Pow(-1, double.NegativeInfinity));
```
All print `1` on .NET Core but `NaN` on .NET 4.7 and old mono (before these changes).
But @tannergooding proved it's correct behavior https://gitter.im/dotnet/corefx?at=5b6e072d5762ee4fe55bc313 (IEEE 754:2008 spec - 9.2.1: Special Values)

`Math.Round` is now managed, but it has `[Intrinsics]` so I guess it's implemented using SSE41.
Also, fixes https://github.com/mono/mono/issues/9989

Part of #7737


tannergooding:
> Math.Round is now managed, but it has [Intrinsics] so I guess it's implemented using SSE41.

Yes, `Math.Round`, `Math.Ceiling`, and `Math.Floor` all use SSE4.1, on supported hardware.

EgorBo:
@monojenkins build failed

mono/metadata/icall-def.h
```diff
@@ -400,28 +400,32 @@ ICALL_TYPE(IOSELECTOR, "System.IOSelector", IOSELECTOR_1)
...
-NOHANDLES(ICALL(MATH_10, "Log10", ves_icall_System_Math_Log10))
-NOHANDLES(ICALL(MATH_11, "Pow", ves_icall_System_Math_Pow))
-NOHANDLES(ICALL(MATH_12, "Round", ves_icall_System_Math_Round))
-NOHANDLES(ICALL(MATH_14, "Sin", ves_icall_System_Math_Sin))
-NOHANDLES(ICALL(MATH_15, "Sinh", ves_icall_System_Math_Sinh))
-NOHANDLES(ICALL(MATH_22, "SplitFractionDouble", ves_icall_System_Math_SplitFractionDouble))
-NOHANDLES(ICALL(MATH_16, "Sqrt", ves_icall_System_Math_Sqrt))
-NOHANDLES(ICALL(MATH_17, "Tan", ves_icall_System_Math_Tan))
-NOHANDLES(ICALL(MATH_18, "Tanh", ves_icall_System_Math_Tanh))
+ICALL_TYPE(MATH, "System.Math", MATH_1)
```
kumpera: This is unwanted line damage

mono/metadata/sysmath.c
```diff
@@ -108,27 +54,27 @@ ves_icall_System_Math_Sin (gdouble x)
 }
 
 gdouble 
-ves_icall_System_Math_Cos (gdouble x)
+ves_icall_System_Math_Cbrt (gdouble x)
```
kumpera: did you sort the functions by name?

kumpera:
`Math.Round is now managed, but it has [Intrinsics] so I guess it's implemented using SSE41.`
Since mono is not CoreCLR, it's not. Plus, SSE41 is X86 only. This PR is a perf regression.

mono/metadata/sysmath.c
EgorBo: um.. yes, the list looks sorted (also no warrnings about the sort order)

EgorBo:
@kumpera intrinsics are implemented here https://github.com/mono/mono/pull/10039
but I guess yes we should keep this method native.

mono/metadata/icall-def.h
EgorBo: could you please clarify why?

jaykrell: I believe @kumpera wants a smaller edit here. If you are inserting functions, you can extend the numbering scheme, e.g. insert "17a" between "17" and "18".

EgorBo:
a small benchmark:
image
Abs is super fast because it was not inline-friendly in ref-source implementation.

EgorBo:
@monojenkins build failed

kumpera:
@EgorBo `[Intrinsic]` is a custom attribute ignored by mono. You can have an icall as fallback for lack of an intrinsic.

mono/metadata/sysmath.c
kumpera: FYI, we don't do this sort of code shuffling in runtime code. We favor smaller diffs.

mono/metadata/icall-def.h
kumpera: Jay is on spot. Larger changes that make it hard to review and reduce the value of history for no  good reason are to be avoided.

EgorBo:
@kumpera yeah I understand that it is just a hint for user that a method is intrinsics (btw, for some reason `Math.Ceiling` and `Math.Floor` are not marked with it in CoreFX). Ok, since we can have an intrinsic on top of icall I guess it makes sense to rollback it.

mono/metadata/icall-def.h
EgorBo: got it, now the diff is smaller, but looks a bit weird :)

mono/metadata/sysmath.c
EgorBo: ah, sorry, I thought you mean icall-defs order. Re-order in this file was made by a mistake

EgorBo:
@monojenkins build failed

EgorBo:
@monojenkins build failed

EgorBo:

mcs/class/corlib/Test/System/MathTest.cs
```diff
@@ -499,15 +499,15 @@ public void TestPow ()
 			Assert.IsTrue (double.IsNaN (Math.Pow (             double.NaN, double.NegativeInfinity)), "#2");
 			Assert.IsTrue (double.IsNaN (Math.Pow (             double.NaN,                      -2)), "#2");
 			Assert.IsTrue (double.IsNaN (Math.Pow (             double.NaN,                      -1)), "#3");
-			Assert.IsTrue (double.IsNaN (Math.Pow (             double.NaN,                       0)), "#4");
+			Assert.IsFalse (double.IsNaN (Math.Pow (             double.NaN,                       0)), "#4");
```
luhenry: Doesn't this break on top of .NET Framework?

EgorBo: it does, but matches .NET Core and the IEEE 754 spec, see https://github.com/dotnet/coreclr/issues/9807

EgorBo: or did you mean the case where we could run our tests on top of .NET Framework on Windows? AFAIR nobody maintains it and it doesn't work.

marek-safar:

akoeplinger:
@EgorBo This caused two failures in acceptance-tests:

```
coreclr/tests/src/JIT/Directed/intrinsic/pow/pow1.exe:

MESSAGE:
x: 1, y: NaN, Pow(x,y): 1
FAILED
```

```
coreclr/tests/src/CoreMangLib/cti/system/math/mathpow.exe:

MESSAGE:
Beginning test case MathPow at 8/13/2018 5:21:32 PM
Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro

[Positive]
Beginning scenario: PosTest1: Verify Pow when one of args is NaN .
Beginning scenario: PosTest2: Verify Pow when second arg is zero .
Beginning scenario: PosTest3: Verify Pow(NegativeInfinity, < 0) .
Beginning scenario: PosTest4: Verify Pow(NegativeInfinity, positive odd int) .
Beginning scenario: PosTest5: Verify Pow(NegativeInfinity, positive non-odd int) .
Beginning scenario: PosTest6: Verify Pow(< 0, non-int) .
Beginning scenario: PosTest7: Verify Pow(-1, NegativeInfinity) .
ERROR!!!-007.1: Return value is wrong!
Beginning scenario: PosTest8: Verify Pow(-1<x<1, NegativeInfinity) .
Beginning scenario: PosTest9: Verify Pow(-1<x<1, PositiveInfinity) .
Beginning scenario: PosTest10: Verify Pow(>1, NegativeInfinity) .
Beginning scenario: PosTest11: Verify Pow(>1, PositiveInfinity) .
Beginning scenario: PosTest12: Verify Pow( 0, < 0) .
Beginning scenario: PosTest13: Verify Pow( 0, > 0) .
Beginning scenario: PosTest14: Verify Pow( 1, any value) .
Beginning scenario: PosTest15: Verify Pow( PositiveInfinity, < 0) .
Beginning scenario: PosTest16: Verify Pow( PositiveInfinity, > 0) .

Ending test case at 8/13/2018 5:21:32 PM
FAIL

```

PR #10042, opened 2018-08-11
[cxx] Better types at the mono_gc_alloc layer to reduce casts elsewhere.

jaykrell:


jaykrell:

jaykrell:
Specifically, at least, all the handle allocators expect MonoObject*, MonoString*, MonoArray* here, and that is very reasonable of them. Or they can cast. And few other places.

This is also likely hiding raw managed pointers from Centrinel as `void*`, though the common cases of turning them right into the handles, is ok.

PR #10053, opened 2018-08-12
[cxx] Don't use overflowing constants, to avoid warning/error in some configurations.

jaykrell:
These are the ones all 0x80000000 or higher.
And remove an excess const.
For consistenty/ease, we could cast them all.
Or change the type to unsigned long (and cast back at use to preserve meaning).

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -4427,12 +4427,21 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 typedef struct {
 	guint32 drive_type;
 #if __linux__
-	const long fstypeid;
+	// http://man7.org/linux/man-pages/man2/statfs.2.html
+	//
+	// The __fsword_t type used for various fields in the statfs structure
+	// definition is a glibc internal type, not intended for public use.
+	// This leaves the programmer in a bit of a conundrum when trying to
+	// copy or compare these fields to local variables in a program.  Using
+	// unsigned int for such variables suffices on most systems.
+	//
+	// Let's hope "most" is enough, and that it works with other libc.
+	unsigned fstypeid;
 #endif
 	const gchar* fstype;
 } _wapi_drive_type;
 
-static _wapi_drive_type _wapi_drive_types[] = {
+static const _wapi_drive_type _wapi_drive_types[] = {
 #if HOST_DARWIN
 	{ DRIVE_REMOTE, "afp" },
 	{ DRIVE_REMOTE, "afpfs" },
@@ -4583,9 +4592,9 @@ static _wapi_drive_type _wapi_drive_types[] = {
 };
 
 #if __linux__
-static guint32 _wapi_get_drive_type(long f_type)
+static guint32 _wapi_get_drive_type(unsigned f_type)
 {
-	_wapi_drive_type *current;
+	const _wapi_drive_type *current;
 
 	current = &_wapi_drive_types[0];
 	while (current->drive_type != DRIVE_UNKNOWN) {
@@ -4599,7 +4608,7 @@ static guint32 _wapi_get_drive_type(long f_type)
 #else
 static guint32 _wapi_get_drive_type(const gchar* fstype)
 {
-	_wapi_drive_type *current;
+	const _wapi_drive_type *current;
 
 	current = &_wapi_drive_types[0];
 	while (current->drive_type != DRIVE_UNKNOWN) {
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X x64

jaykrell:
Using `long` here is also dubious, it varies per-target, should probably be int/int32/uint/uint32.

mono/metadata/w32file-unix.c
```diff
@@ -4402,7 +4402,7 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 typedef struct {
 	guint32 drive_type;
 #if __linux__
-	const long fstypeid;
+	long fstypeid;
```
lambdageek: From http://man7.org/linux/man-pages/man2/statfs.2.html#NOTES
> The __fsword_t type used for various fields in the statfs structure
>       definition is a glibc internal type, not intended for public use.
>       This leaves the programmer in a bit of a conundrum when trying to
>       copy or compare these fields to local variables in a program.  Using
>       unsigned int for such variables suffices on most systems.

So maybe we should just use `unsigned long` here?

jaykrell: So..this is gnarly stuff actually. We have to have some value we store, and then later compare it to some value we get back. We need to ensure that sign extension works the same.

I think the best options are:
1. global: _fsword_t 

2. global: almost anything, int, int32, uint, uint32, long (casted), ulong, etc.
But don't use the value directly.
Have a local struct statfs, store the global into statfs.f_type, and then compare to the other statfs.f_type. Having the full struct statfs and storing into a temporary is a C form of casting to `decltype`.

3. Or what I showed, just cast, but put an #if around so it is for 32bit long, in order to not change semantics.

mono/metadata/w32file-unix.c
```diff
@@ -4558,13 +4563,18 @@ static _wapi_drive_type _wapi_drive_types[] = {
...
+	const _wapi_drive_type *current;
 
 	current = &_wapi_drive_types[0];
 	while (current->drive_type != DRIVE_UNKNOWN) {
-		if (current->fstypeid == f_type)
+		// f_type is the platform-specific internal __statfs_word, which we
+		// prefer not to utter, but still use a temporary with exact matching type.
+		// We don't want to be surprised by varying zero-extension vs. sign-extension.
+		struct statfs stat2;
+		stat2.f_type = current->fstypeid;
```
jaykrell: This line is still imperfect. We are assuming the conversion of unsigned long to f_type is lossless.

I have one more idea here, and that is to never store our data.
Instead of the array of constants, we can generate an if/else ladder or switch.
So it'd look like switch (stat1->f_type) case XX.

However, that makes me think we should just __statfs_word as a lesser evil.

I'm going to read the header/manpage/bing again. This shouldn't be so difficult.

jaykrell: The man page acknowledges our exact situation:

"conundrum when trying to copy or compare these fields to local variables in a program"

and says "Using unsigned int for such variables suffices on most systems."

"most systems"?

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins help

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build AArch64 FullAOT

jaykrell:

jaykrell:

jaykrell:
@monojenkins Linux AArch64 Coop Suspend

jaykrell:
@monojenkins OS X i386

jaykrell:
@monojenkins OS X x64

jaykrell:

jaykrell:

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X x64

PR #10057, opened 2018-08-13
Strengthen the type of mono_gc_get_range_copy_func -- return a typed function pointer.

jaykrell:
instead of a void*. And add const on the source of a memcpy.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build OS X i386

PR #10060, opened 2018-08-13
Minor corlib cleanups

filipnavara:
- Remove two unused types from corlib (DecoupledTask is not used since migrating to StreamReader/StreamWriter from referencesources).
- Fix indentation in System.Security.Policy/Site.cs which mixed tabs and spaces.
- Fix comment in Test/System.Diagnostics/DebuggerTypeProxyAttribute.cs (to avoid false results when using grep on the source code).


akoeplinger:
@monojenkins commit csproj

Issue #9566, opened 2018-07-14
flaky runtime.finalizer-exception.exe (Assertion at unwind.c:640, condition `cfa_reg != -1' not met)


jaykrell:
## Steps to Reproduce

1. Submit a PR. 

### Current Behavior

Failures, for example:

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64/11720/

MonoTests.runtime.finalizer-exception.exe

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64/11720/testReport/junit/MonoTests/runtime/finalizer_exception_exe/

```
Failing for the past 1 build (Since Unstable#11720 )
Took 0.38 sec.
add description
Stacktrace
                                                MESSAGE:
                                                caught
* Assertion at unwind.c:640, condition `cfa_reg != -1' not met

* Assertion at mini-exceptions.c:1191, condition `unwind_options == MONO_UNWIND_NONE' not met

                                                +++++++++++++++++++
                                                STACK TRACE:
                                                Stacktrace:
```

For this we probably could use some EH stress that forces
an exception at every instruction or every call, or every call and dereference.

### Expected Behavior

Success.

jaykrell:
Another instance:

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64/11707/

MonoTests.test-unhandled-exception-2-255-with-managed-handler.unhandled-exception-2.exe

https://jenkins.mono-project.com/job/test-mono-pull-request-arm64/11707/testReport/junit/MonoTests/test-unhandled-exception-2-255-with-managed-handler/unhandled_exception_2_exe/

DaanTacoSchool:
not a 100% sure this is the correct thread, but the error in the title still appears. i can find a lot about it in unity threads, but for me it occurs in VS2017 w/ Xamarin forms. in the android emulator. see the [link](https://forums.xamarin.com/discussion/69205/ios-crash-error?) for details

BrzVlad:
@DaanTacoSchool This error is not really actionable from the assertion only, we would need a reproduction case. In the meantime, for the next release, I'll include additional logging when this error shows so we can hopefully debug it more easily.

DaanTacoSchool:
I Cannot supply a reproduction case as i dont know what causes this. It
happens on all installed emulators. (ranging from api 26-28 and ram
512mb-2gb, even the heap size and renderes are varying). this occurred
since android studio was installed alongside visual studio (which should be
[...]

PR #10071, opened 2018-08-13
mono_thread_info_set_is_async_context assert that setting to false is correct, i.e. no nesting

jaykrell:
mono_thread_info_set_is_async_context assert that setting to false is correct, i.e. no nesting
Possibly related to https://github.com/mono/mono/issues/9566

### DIFF
--- a/mono/utils/mono-threads.c
+++ b/mono/utils/mono-threads.c
@@ -1373,8 +1373,18 @@ mono_thread_info_set_is_async_context (gboolean async_context)
 {
 	MonoThreadInfo *info = mono_thread_info_current ();
 
-	if (info)
+	if (info) {
+		// If this assert fails, that means there is recursion and/or
+		// concurrency, such that setting async_context to FALSE
+		// that all the callers do after this, is incorrect,
+		// and this should instead be incremented/decremented.
+		//
+		// As the value is only accessed via current(), that
+		// limits the case to recursion, but increment/decrement
+		// is still fast and correct and simple.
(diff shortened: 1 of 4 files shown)


BrzVlad:
Does this happen in the first place ? I don't see the use of recursiveness for this. Personally I'd rather have an assert just in case code will change.

jaykrell:
I don't know.
This is where reading the code got me wrt that bugweek bug.
The pattern in the code seems suspicious and the pattern in the PR seems very safe.

jaykrell:
It's not a recursive thing necessarily, it is what state is the thread in when we call this.

jaykrell:

mono/utils/mono-threads.c
```diff
@@ -1373,8 +1373,18 @@ mono_thread_info_set_is_async_context (gboolean async_context)
...
-	if (info)
+	if (info) {
+		// If this assert fails, that means there is recursion and/or
+		// concurrency, such that setting async_context to FALSE
+		// that all the callers do after this, is incorrect,
+		// and this should instead be incremented/decremented.
+		//
+		// As the value is only accessed via current(), that
+		// limits the case to recursion, but increment/decrement
+		// is still fast and correct and simple.
```
jaykrell: Ugh I edited the comment wrong. It used to mention save/set/restore..

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X x64

jaykrell:

PR #10084, opened 2018-08-13
[runtime] Make mini-llvm use only mono_aot_can_dedup

alexanderkyte:
LLVM was using the helper for the abandoned prototype linkonce work that
much predated when I picked up the project.

This is a problem because it can lead to aot-compiler thinking a method
is dedupable, while llvm thinking it's not. When this happens, the
symbol does not have the Assembly baked into it's name to ensure
uniqueness.


kumpera:
@monojenkins build Windows 64

kumpera:

PR #10088, opened 2018-08-14
Detect Arm/Arm64 in RuntimeInformation

NattyNarwhal:
* For ProcessArchitecture, use the DllMap arch of the runtime.

* For OSArchitecture, make an educated guess combining ProcessArchitecture and Environment.Is64BitOperatingSystem.

* For non-Intel/ARM architectures, (supported by Mono and not on the chopping block are PPC32/64, z, and WebAssembly) please dogpile onto dotnet/corefx#30706.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=42403

### DIFF
--- a/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
+++ b/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
@@ -35,6 +35,9 @@ namespace System.Runtime.InteropServices
 {
 	public static class RuntimeInformation
 	{
+		[DllImport("__Internal", CharSet = CharSet.Auto)]
+		static extern string mono_config_get_cpu ();
+
 		public static string FrameworkDescription {
 			get {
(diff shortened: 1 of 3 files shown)


mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
```diff
@@ -35,6 +35,9 @@ namespace System.Runtime.InteropServices
 {
 	public static class RuntimeInformation
 	{
+		[DllImport("__Internal", CharSet = CharSet.Auto)]
```
akoeplinger: Just remembered that we can't use `__Internal` as it would end up not working in embedded cases like Xamarin.Android because it searches for the method in the host which embeds Mono.

You'll need to add a new icall for this.

NattyNarwhal:
I do have the suggestion of using DllMap OS name/uname as the default case fallback for OSPlatform, but that's out of scope for this PR.

akoeplinger:
Thanks for the contribution! ❤️

Issue #9452, opened 2018-07-07
Runtime uses multiple instances of custom marshaller for marshalling single argument


rokups:
## Steps to Reproduce

Code reproducing issue:
```cs
struct MarshalTest : ICustomMarshaler
{
    private IntPtr _stored;
    public static ICustomMarshaler GetInstance(string _)
    {
        return new MarshalTest();        // This is important
    }

    public void CleanUpManagedData(object ManagedObj)
    {
    }

    public void CleanUpNativeData(IntPtr pNativeData)
    {
        System.Console.WriteLine($"CleanUpNativeData: _stored = {_stored}, pNativeData = {pNativeData}, identity = {GetHashCode()}");
    }

    public int GetNativeDataSize()
    {
        return -1;
    }

    public IntPtr MarshalManagedToNative(object ManagedObj)
    {
        _stored = (IntPtr)int.Parse((string)ManagedObj);
        System.Console.WriteLine($"MarshalManagedToNative: _stored = {_stored}, returning = {_stored}, identity = {GetHashCode()}");
        return _stored;
    }

    public object MarshalNativeToManaged(IntPtr pNativeData)
    {
        System.Console.WriteLine($"MarshalNativeToManaged: {pNativeData}");
        return pNativeData.ToString();
    }
}

[DllImport("NativeLibrary", CallingConvention = CallingConvention.Cdecl)]
private static extern void TestMarshalling([param: MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(MarshalTest))]ref string p);

public static void Main(string[] args)
{
    var param = "1";
    TestMarshalling(ref param);
}
```
```cpp
EXPORT_API void TestMarshalling(void** ptr)
{
    *ptr = (void*)2;
}
```

### Current Behavior

mono 5.12.0 prints:

    MarshalManagedToNative: _stored = 1, returning = 1, identity = 1192725433
    MarshalNativeToManaged: 2
    CleanUpNativeData: _stored = 0, pNativeData = 2, identity = 1192725432

Mono calls `MarshalManagedToNative()` on one instance of the struct and then calls `CleanUpNativeData()` on a new instance. Due to this behaviour no state (like `_stored`) can be used in custom marshaller.

### Expected Behavior

.net framework 4.7 prints:

    MarshalManagedToNative: _stored = 1, returning = 1, identity = 346948957
    MarshalNativeToManaged: 2
[...]

PR #10090, opened 2018-08-14
Fix issue #9452: Add a cache for CustomMarshaler instances

kg:
This PR moves some of our CustomMarshaler logic into managed code and implements a cache keyed on type+cookie to match .NET Framework & .NET Core's (not documented) behavior. Fixes issue #9452.

Pending:
* Add tests
* Figure out how to emit a type token or Type into the IL, because right now doing that causes a silent verification failure and results in an empty method body with no debugging info
* Add error handling for various GetInstance failure scenarios
* Verify the way .NET Framework and .NET Core handle GetInstance failure scenarios (what exception type, what message, etc)
* Eliminate some of the duplicate get_instance calls from the original code because there's not really a good reason for them


vargaz:
I think the marshallers should be cached during a call, not globally, i.e. if a ref parameter has a marshaller, the same instance should be used for both the in and out directions.

kg:
I agree, but that is not how .NET Framework/.NET Core behave. If you mean don't call get_instance multiple times, yeah, that is also on my todo list - forgot to include it. I'm not sure why we don't DUP them in every case already.

vargaz:
Does .net really reuse the same custom marshaller instance between calls ?

kg:
Yes, when I ran my test case on windows and linux netcore it used the same instance across all calls, cached based on type+cookie pair. The test case is in the issue

mono/metadata/marshal.c
```diff
@@ -6210,11 +6210,11 @@ mono_icall_handle_new_interior (gpointer rawobj)
 }
 
 MonoObject*
-mono_marshal_get_type_object (MonoType *type)
+mono_marshal_get_type_object (MonoClass *klass)
```
kumpera: This requires a bump to BCL or AOT

kg:
Will I need to do more than bump monolite?

kumpera:
No, bumping it should be more than enough.

akoeplinger:
@monojenkins build failed

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,84 @@ internal static void SetLastWin32Error (int error)
 			throw new NotSupportedException();
 		}
 #endif
+
+		internal struct MarshalerInstanceKey {
+			public Type Type;
+			public string Cookie;
+
+			public override int GetHashCode () {
+				int result = 0;
```
marek-safar: `int result = Type?.GetHashCode () ?? 0;` would look nicer or use HashCode utility

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,84 @@ internal static void SetLastWin32Error (int error)
...
+			public bool Equals (MarshalerInstanceKey lhs, MarshalerInstanceKey rhs) {
+				return (lhs.Type == rhs.Type) && (lhs.Cookie == rhs.Cookie);
+			}
+
+			public int GetHashCode (MarshalerInstanceKey key) {
+				return key.GetHashCode ();
+			}
+		}
+
+		internal static Dictionary<MarshalerInstanceKey, ICustomMarshaler> MarshalerInstanceCache =
```
marek-safar: This should be lazy initialized

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,84 @@ internal static void SetLastWin32Error (int error)
 			throw new NotSupportedException();
 		}
 #endif
+
+		internal struct MarshalerInstanceKey {
```
marek-safar: Why not use ValueTuple?

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,84 @@ internal static void SetLastWin32Error (int error)
...
+
+			if (!gotExistingInstance) {
+				MonoMethod getInstanceMethod;
+				try {
+					getInstanceMethod = (MonoMethod)type.GetMethod (
+						"GetInstance", BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.InvokeMethod,
+						null, new Type[] { typeof(string) }, null
+					);
+				} catch (AmbiguousMatchException) {
+					throw new ApplicationException ($"Custom marshaler '{type.FullName}' implements multiple static GetInstance methods that take a single string parameter.");
```
marek-safar: ApplicationException is not right exception for BCL code

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,84 @@ internal static void SetLastWin32Error (int error)
...
+			ICustomMarshaler result;
+			bool gotExistingInstance;
+			lock (MarshalerInstanceCache)
+				gotExistingInstance = MarshalerInstanceCache.TryGetValue (key, out result);
+
+			if (!gotExistingInstance) {
+				MonoMethod getInstanceMethod;
+				try {
+					getInstanceMethod = (MonoMethod)type.GetMethod (
+						"GetInstance", BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.InvokeMethod,
```
marek-safar: Please add test that all this work with linker enabled to mcs/tools/linker/Tests

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
kg: It's what this code throws in other implementations

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
kg: Didn't think of it. It's in corlib in all our profiles?

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
kg: Is lazy initialization trivially thread safe? I guess I could guard it with a separate lock but I wouldn't be able to lazy initialize that. If there's somewhere else in the BCL we do this I can just copy that pattern

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
marek-safar: It should be

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
marek-safar: oh, really. that's strange

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
marek-safar: Just use `LazyInitializer.EnsureInitialized`

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
kg: I added linker tests but the linker appears to eat custom marshalers entirely. I think this means we need to fix the linker. For now I guess I should whitelist the marshaler classes?

marek-safar: Whitelist Marshall class is too aggressive. I can look at the linker tweaks so just add the test as disabled

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,71 @@ internal static void SetLastWin32Error (int error)
...
+
+		internal static Dictionary<(Type, string), ICustomMarshaler> MarshalerInstanceCache = null;
+		internal static object MarshalerInstanceCacheLock = new object ();
+
+		internal static ICustomMarshaler GetCustomMarshalerInstance (Type type, string cookie) {
+			var key = (type, cookie);
+
+			ICustomMarshaler result;
+			bool gotExistingInstance;
+			lock (MarshalerInstanceCacheLock) {
```
marek-safar: The lock is redundant for LazyInitializer

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1897,5 +1897,72 @@ internal static void SetLastWin32Error (int error)
...
+					throw new ApplicationException ($"Custom marshaler '{type.FullName}' does not implement a static GetInstance method that takes a single string parameter and returns an ICustomMarshaler.");
+				}
+
+				Exception exc;
+				try {
+					result = (ICustomMarshaler)getInstanceMethod.InternalInvoke (null, new object[] { cookie }, out exc);
+				} catch (Exception e) {
+					// FIXME: mscorlib's legacyUnhandledExceptionPolicy is apparently 1, 
+					//  so exceptions are thrown instead of being passed through the outparam
+					exc = e;
```
marek-safar: Just doing `throw e;` would be same as capture + throw

kg: doesn't that trash the stack?

marek-safar: yes, `throw e;` resets the stack trace whereas `throw;` doesn't

kg: I wanted to try to preserve and propagate the original stack, because that is what .net framework does. I noticed .Throw does not do that, but it at least includes the original stack (below the current stack), which is better than nothing.

marek-safar: ok, let's move the Throw () into the catch section

kg: If the icall is changed to actually write the exception into the var, the throw won't get used and then we'll blow away the stack. I guess I could put it in both places?

marek-safar: ok, I missed that part

kg:

kg:

kg:
@monojenkins build failed

kg:

kg:
@monojenkins build OS X i386

kg:

PR #10105, opened 2018-08-14
[glib] About half of glib does not use config.h.

jaykrell:


jaykrell:
I mean, I understand how it got this way, but this isn't how it is now meant to be, right?
We are supposed to be able to say ifdef host_win32 in these parts (glib), right?

vargaz:
Yes, we should probably include config.h in eglib-config.h.

jaykrell:
Huh this might actually cause problems, esp. the later form where config.h is put into glib. Let me try again with it in the .c files.

jaykrell:
Putting the include in glib.h itself seemed to cause major breakage, so back to putting it in the .c files.

jaykrell:

jaykrell:

jaykrell:

jaykrell:
I had to workaround it, using `#ifdef _WIN32` instead of `#ifdef HOST_WIN32` in `glib.h`.

However, `glib.h` itself I believe cannot (yet) assume `config.h`, as I think there are other users that don't use `config.h`. i.e. when I put `config.h` in `glib.h`, stuff broke. So I left in the workaround for now. More to do.

PR #10118, opened 2018-08-15
[cxx] Char literals are now chars, not ints (sizeof changes) and indexing by them warns. Cast.

jaykrell:
(Really just split this out because github is slow linking to within large diff).


jaykrell:
@monojenkins build  OS X i386

jaykrell:
@monojenkins build OS X i386

jaykrell:
Review?

Issue #9870, opened 2018-08-03
iOS: 32-bit device unit test failure: [FAIL] FSharpTest.SprintfTest :   Expected: True


rolfbjarne:
## Steps to Reproduce

1. Checkout & build the `mono-2018-06` branch from xamarin-macios
2. Open tests/tests.sln
3. Select the fsharp project as the startup project, and select a Debug32|iPhone configuration (this failure does not happen in 64-bit).
4. Execute and run all the tests

### Current Behavior

    [FAIL] FSharpTest.SprintfTest :   Expected: True
        But was:  False
		  at fsharp.FSharpTest.SprintfTest () [0x00052] in /work/maccore/mono-master/xamarin-macios/tests/fsharp/FSharpTests.fs:34 
		  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)

improving the test a bit gets me a better assert message:

    [FAIL] FSharpTest.SprintfTest :   String lengths are both 24. Strings differ at index 10.
      Expected: "1111 2222 3333 4444 5555"
      But was:  "1111 2222 4444 3333 5555"
      ---------------------^
    at fsharp.FSharpTest.SprintfTest () [0x00044] in /work/maccore/mono-master/xamarin-macios/tests/fsharp/FSharpTests.fs:33 
    at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)

### Expected Behavior

No test failure.

## On which platforms did you notice this

[x] iOS
[ ] macOS
[ ] Linux
[ ] Windows

vargaz:
Did this use to work ?

vargaz:
Which was the last version where this was tested and it succeeded ?

jaykrell:
Does this require device or repros in emulator?
Simple desktop/amd64 test passes.

rolfbjarne:
@jaykrell it requires an iOS device capable of running 32-bit code

@vargaz it passes in d15-7.

the same thing happens with watchOS on device:

    FSharpTest
    	[PASS] FSharpTest.Func
    	[FAIL] FSharpTest.SprintfTest :   Expected string length 24 but was 32. Strings differ at index 5.
      Expected: "1111 2222 3333 4444 5555"
      But was:  "1111 57822332 2222 57822332 5555"

PR #10121, opened 2018-08-15
[jit] Disable tail calls from gsharedvt code for now.

vargaz:
Fixes https://github.com/mono/mono/issues/9870.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -6254,6 +6254,7 @@ is_supported_tailcall (MonoCompile *cfg, const guint8 *ip, MonoMethod *method, M
 		// Interface method dispatch has the same problem (imt_arg).
 
 		|| IS_NOT_SUPPORTED_TAILCALL (extra_arg && !cfg->backend->have_volatile_non_param_register)
+			   || IS_NOT_SUPPORTED_TAILCALL (cfg->gsharedvt)
 		) {
 		tailcall_calli = FALSE;
 		tailcall = FALSE;


jaykrell:
These all have wrappers anyway, so aren't fully tailcalls anyway?
i.e. any tests impacted?

marek-safar:
@monojenkins backport 2018-08

marek-safar:
@monojenkins backport 2018-06

marek-safar:
@monojenkins rebase

marek-safar:
@jaykrell this breaks `MonoTests.runtime.tailcall/interface-conservestack/20.exe` tests for FullAOT. Should we disable it as well?

jaykrell:
Sure. I'm kinda surprised it breaks anything, because..well..gsharedvt is wierd and already caused brief. See here in Makefile.am:

```
# Some, probably all, of these tests use gsharedvt.
# gsharedvt is very much at odds with tailcall, and they fail as a result.
# RegularAOT and HybridAOT should be ok.
#
# The tests might be tweakable but then they will less resemble real world code.
#
# FullAOT should probably be changed to intersect tailcall signatures called
# and all generic signatures provided, and compile more functions.
# As well, gsharedvt can probably be a tailcall sometimes, like if parameters
# fit in registers and there is no out conversion. Maybe it already does that.
#
# gsharedvt should be viable when only wrapping "in" but not "out",
# as long as "in" does not expand parameter stack size, which it should never do.
if MONO_ARCH_GSHAREDVT_SUPPORTED
PROFILE_DISABLED_TESTS += tailcall-interface-conservestack.exe
PROFILE_DISABLED_TESTS += tailcall/fsharp-deeptail.exe
PROFILE_DISABLED_TESTS += tailcall/interface-conservestack/26.exe
```

jaykrell:
I can get it, assuming this is the entire problem.

jaykrell:
https://github.com/mono/mono/pull/10323 Let's give CI a chance?

PR #10133, opened 2018-08-15
[runtime] Add pipeline to convert managed exception into telemetry format

alexanderkyte:
This enables a correctly-configured runtime to convert unhandled exceptions into telemetry-formatted dumps. 

Future readers: The tests / example usage for the MS-Telemetry-side interop are in the commit message for 

```
[runtime] Add test for managed telemetry reporting 
```
at
https://github.com/mono/mono/pull/10133/commits/97a7af3f9733c29e49183247e866d8ad06601c2f

### DIFF
--- a/mcs/class/corlib/Mono/Runtime.cs
+++ b/mcs/class/corlib/Mono/Runtime.cs
@@ -81,12 +81,63 @@ namespace Mono {
 			return true;
 		}
 
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern void ExceptionToState (Exception exc, out string payload, out IntPtr portable_hash, out IntPtr unportable_hash);
+
 #if !MOBILE 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		static extern void DisableMicrosoftTelemetry (IntPtr appBundleID, IntPtr appSignature, IntPtr appVersion, IntPtr merpGUIPath);
+		static extern void DisableMicrosoftTelemetry ();
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		static extern void EnableMicrosoftTelemetry_internal (IntPtr appBundleID, IntPtr appSignature, IntPtr appVersion, IntPtr merpGUIPath, IntPtr eventType, IntPtr appPath);
+
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern void SendMicrosoftTelemetry_internal (IntPtr payload, IntPtr portable_hash, IntPtr unportable_hash);
+
+		static void SendMicrosoftTelemetry (string payload_str, IntPtr portable_hash, IntPtr unportable_hash)
+		{
+			if (RuntimeInformation.IsOSPlatform (OSPlatform.OSX)) {
+				using (var payload_chars = RuntimeMarshal.MarshalString (payload_str))
+				{
+					SendMicrosoftTelemetry_internal (payload_chars.Value, portable_hash, unportable_hash);
+				}
+			} else {
+				throw new PlatformNotSupportedException("Merp support is currently only supported on OSX.");
+			}
+		}
+
+		// Usage: 
+		//
+		// catch (Exception exc) {
+		//   var monoType = Type.GetType ("Mono.Runtime", false);
+		//   var m = monoType.GetMethod("SendExceptionToTelemetry", BindingFlags.NonPublic | BindingFlags.Static);
+		//   m.Invoke(null, new object[] { exc });
+		// }
+		static void SendExceptionToTelemetry (Exception exc)
+		{
+			string payload_str;
+			IntPtr portable_hash;
+			IntPtr unportable_hash;
+			ExceptionToState (exc, out payload_str, out portable_hash, out unportable_hash);
+			SendMicrosoftTelemetry (payload_str, portable_hash, unportable_hash);
+		}
+
+		static void EnableMicrosoftTelemetry (string appBundleID_str, string appSignature_str, string appVersion_str, string merpGUIPath_str, string eventType_str, string appPath_str)
+		{
+			if (RuntimeInformation.IsOSPlatform (OSPlatform.OSX)) {
+				using (var appBundleID_chars = RuntimeMarshal.MarshalString (appBundleID_str))
+				using (var appSignature_chars = RuntimeMarshal.MarshalString (appSignature_str))
+				using (var appVersion_chars = RuntimeMarshal.MarshalString (appVersion_str))
+				using (var merpGUIPath_chars = RuntimeMarshal.MarshalString (merpGUIPath_str))
+				using (var eventType_chars = RuntimeMarshal.MarshalString (eventType_str))
+				using (var appPath_chars = RuntimeMarshal.MarshalString (appPath_str))
+				{
+					EnableMicrosoftTelemetry_internal (appBundleID_chars.Value, appSignature_chars.Value, appVersion_chars.Value, merpGUIPath_chars.Value, eventType_chars.Value, appPath_chars.Value);
+				}
+			} else {
+				throw new PlatformNotSupportedException("Merp support is currently only supported on OSX.");
+			}
+		}
 
 		static void EnableMicrosoftTelemetry (string appBundleID_str, string appSignature_str, string appVersion_str, string merpGUIPath_str, string eventType_str, string appPath_str)
 		{
--- a/mcs/class/corlib/Mono/Runtime.cs
+++ b/mcs/class/corlib/Mono/Runtime.cs
@@ -82,7 +82,18 @@ namespace Mono {
 		}
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		static extern void ExceptionToState (Exception exc, out string payload, out IntPtr portable_hash, out IntPtr unportable_hash);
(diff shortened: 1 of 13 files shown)


alexanderkyte:
People of interest with this PR who are not in the reviewers list (non-exhaustive list):
@migueldeicaza 
@sgmunn 
@slluis

mono/metadata/threads.c
```diff
@@ -5993,7 +5993,7 @@ mono_threads_summarize_one (MonoThreadSummary *out, MonoContext *ctx)
 
 	MonoInternalThread *thread = (MonoInternalThread *) mono_gchandle_get_target (handle);
```
jaykrell: Why? The previous form is resilient to changing type, assuming the entire thing should be zeroed.

mono/mini/mini-exceptions.c
```diff
@@ -1448,6 +1464,22 @@ mono_summarize_stack (MonoDomain *domain, MonoThreadSummary *out, MonoContext *c
 }
 #endif
 
+static void
+mono_summarize_exception (MonoException *exc, MonoThreadSummary *out)
+{
+	memset (out, 0, sizeof (MonoThreadSummary));
```
jaykrell: Suggest `sizeof (*out)` in case out change changes type.

mono/mini/mini-exceptions.c
```diff
@@ -1448,6 +1464,22 @@ mono_summarize_stack (MonoDomain *domain, MonoThreadSummary *out, MonoContext *c
...
+
+	MonoSummarizeUserData data;
```
jaykrell: Suggest `memset (&data, 0, sizeof (data))` in case fields are added, for which zero is a safe default.

mono/utils/mono-state.c
```diff
@@ -466,24 +463,53 @@ mono_native_state_add_epilogue (JsonWriter *writer)
...
-	return writer.text->str;
+	return mono_native_state_emit (&writer);
 }
 
 void
 mono_summarize_native_state_add_thread (MonoThreadSummary *thread, MonoContext *ctx)
 {
-	mono_native_state_add_thread (&writer, thread, ctx);
+	
+	static gboolean not_first_thread = FALSE;
```
jaykrell: No need to say `= FALSE` here.

mono/metadata/icall.c
```diff
@@ -5809,6 +5810,54 @@ ves_icall_Mono_Runtime_EnableMicrosoftTelemetry (char *appBundleID, char *appSig
...
+ICALL_EXPORT void
+ves_icall_Mono_Runtime_SendMicrosoftTelemetry (char *payload, unsigned long portable_hash, unsigned long unportable_hash, MonoError *error)
+{
+#ifdef TARGET_OSX
+	if (!mono_merp_enabled ())
+		g_error ("Cannot send telemetry without registering parameters first");
+
+	pid_t crashed_pid = getpid ();
+
+	MonoStackHash hashes;
```
jaykrell: Worth a `memset (&hashes, 0, sizeof (hashes))` in case it grows?

mono/metadata/threads.c
alexanderkyte: I can revert this change if people really prefer the previous one, but as a reader I find the `*out` form produces code that is much less readable. For short functions like this, you can just visually check out the variable type, but reminders of types become more and more useful to code reviewers as the code gets longer.

I also prefer it because the old code is really not resilient to a changing type signature. If you make `out` into a `gpointer *`, this program will subtilely break. 

Now one can say that a programmer who makes that change will of course update the memset. I favor defensive programming in general though. C makes it easy enough to shoot yourself in the foot without me adding more footguns.

mono/metadata/icall.c
```diff
@@ -5809,6 +5810,54 @@ ves_icall_Mono_Runtime_EnableMicrosoftTelemetry (char *appBundleID, char *appSig
 #endif
 }
 
+ICALL_EXPORT MonoStringHandle
+ves_icall_Mono_Runtime_ExceptionToState (MonoExceptionHandle exc_handle, unsigned long *portable_hash_out, unsigned long *unportable_hash_out, MonoError *error)
```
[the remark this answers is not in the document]
alexanderkyte: Thanks for spotting this, I was debugging all the icall out parameter marshaling stuff using machine primitives and I completely forgot to switch back to glib types once it was working.

mono/utils/mono-state.c
alexanderkyte: No, but I think it makes it more readable in general to people who aren't familiar with C memory specifics. Every time I merge code with a static variable without an initializer, someone asks "are you sure this memory gets zeroed and it's not just garbage?"

You're 100% correct in machine semantics. I don't disagree there.

mono/metadata/threads.c
jaykrell: The less type-dependent form is generally considered more resilient and readable and defensive, but there are subtle arguments either way.

The problem with naming the type, is the reader has to double check if it is in fact the correct type. And the writer has to notice to update it.

Changing out to gpointer* works, depending on intent. You'd be nulling the pointer.
Try to use strong types instead of gpointer, as well.

mono/metadata/icall.c
alexanderkyte: Github doesn't show a line diff, but I fixed this.

mono/mini/mini-exceptions.c
alexanderkyte: Github doesn't show a line diff, but I fixed this.

mono/utils/mono-state.c
jaykrell: People need to know this detail in C to read much code.
It is a widespread idiom.

Another thing to consider, is to give the variable a longer `mono_`-prefixed name and put it at global scope. That way I can easily ask the debugger to see its value even if I'm not in the function. There is a scoping tradeoff though, no best answer.

alexanderkyte:

mono/metadata/icall.c
jaykrell: Would still prefer `sizeof (hashes)` though.

mono/utils/mono-merp.c
```diff
@@ -514,19 +519,20 @@ mono_wer_template (MERPStruct *merp)
 }
 
 void
-mono_merp_invoke (const intptr_t crashed_pid, const char *signal, const char *non_param_data, MonoStackHash *hashes, char *version)
+mono_merp_invoke (const intptr_t crashed_pid, const char *signal, const char *non_param_data, MonoStackHash *hashes)
 {
 	MERPStruct merp;
 	memset (&merp, 0, sizeof (merp));
```
jaykrell: See `sizeof (variable)` here, not `sizeof (type)`, good.

mono/utils/mono-state.c
```diff
@@ -466,24 +463,53 @@ mono_native_state_add_epilogue (JsonWriter *writer)
...
+{
+	mono_native_state_add_epilogue (writer);
+	return writer->text->str;
+}
+
+char *
+mono_native_state_free (JsonWriter *writer, gboolean free_data)
+{
+	mono_native_state_add_epilogue (writer);
+	gchar *output = NULL;
```
jaykrell: nit: `gchar` is historical, `char` in new code.

alexanderkyte:
Note to self: backport to 2018-04
@slluis do we want this on 2018-02? It's a big change.

alexanderkyte:
Looks like the interpreter test is having some problems. When I check on the unmanaged and managed sides, the `out` parameter doesn't seem to match up. I think I might be seeing a legitimate issue @lewurm

alexanderkyte:
Identified the issue as specific to interpreter but unrelated to icall mechanism. Fix pushed to this PR.

alexanderkyte:
Failures so far are all in the `runtime` suite, which doesn't hit the added code paths.

alexanderkyte:
@monojenkins build failed

luhenry:

luhenry:

luhenry:

luhenry:

luhenry:
@alexanderkyte could you please take care of backporting manually?

alexanderkyte:
I'm on it

alexanderkyte:
@luhenry can I get a review/approval on the other 3 backport PRs?

mcs/class/corlib/Mono/Runtime.cs
```diff
@@ -81,13 +81,57 @@ public static bool SetGCAllowSynchronousMajor (bool flag)
 			return true;
 		}
 
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
```
marek-safar: Why is it not guarded with `!MOBILE` when it used only from there?

mcs/class/corlib/Mono/Runtime.cs
```diff
@@ -81,13 +81,56 @@ public static bool SetGCAllowSynchronousMajor (bool flag)
...
+
+		static void SendMicrosoftTelemetry (string payload_str, ulong portable_hash, ulong unportable_hash)
+		{
+			if (RuntimeInformation.IsOSPlatform (OSPlatform.OSX)) {
+				using (var payload_chars = RuntimeMarshal.MarshalString (payload_str))
+				{
+					SendMicrosoftTelemetry_internal (payload_chars.Value, portable_hash, unportable_hash);
+				}
+			} else {
+				throw new PlatformNotSupportedException("Merp support is currently only supported on OSX.");
```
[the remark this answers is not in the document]
marek-safar: Do we really want to throw here?

PR #10136, opened 2018-08-15
[wasm] Enough scafold to enable AOT work with the SDKs

kumpera:


kumpera:
Fixed comments from @jeromelaban

kumpera:
Blocked on #10153

kumpera:
@monojenkins build OS X i386

mono/mini/objects.cs
```diff
@@ -1389,6 +1389,7 @@ sealed class Bar2 : Foo2 {
 		return 1.4e-45f;
 	}
 
+#if !NO_BITCODE
 	[Category ("!BITCODE")] // bug #59953
```
vargaz: Why is this disabled if it already has a !BITCODE category ?

kumpera: It was crashing the AOT compiler, I'll remove it once its fixed.

PR #10139, opened 2018-08-15
[sdks] Add a XCODE32_DIR make variable and use it to compile the 32 bit cross compilers.

vargaz:


luhenry:
As far as I looked at it with @VincentDondain this is not needed. @rolfbjarne could you please show us where this is required in xamarin-macios?

vargaz:
We do set CC/CXX:
_ios-$(1)_CC=$$(CCACHE) $$(_ios_$(1)_PLATFORM_BIN)/clang
_ios-$(1)_CXX=$$(CCACHE) $$(_ios_$(1)_PLATFORM_BIN)/clang++

I tested this, and it works.

luhenry:
@vargaz then why cannot we more simply pass the specific XCODE_DIR when invoking package-ios-cross32 for example?

vargaz:
That would work too, this version just makes it a bit easier to pass the argument, since the caller doesn't have to pass 2 kinds of arguments, it also makes things clearer that some builds require a 32 bit ready xcode.

luhenry:
I don't really like the complication, since it will make it harder for us to track.

luhenry:
It was fixed in XI directly: https://github.com/xamarin/xamarin-macios/blob/master/builds/Makefile#L1568

PR #10147, opened 2018-08-16
[cxx] fmod is overloaded, assign to local to disambiguate.

jaykrell:
```
$ cat fmod-broken.cpp

#include <math.h>

void * pmod = fmod;

$ cc -c fmod-broken.cpp
fmod-broken.cpp:3:15: error: address of overloaded function 'fmod' does not match required type 'void'
void * pmod = fmod;
              ^~~~

1 error generated.

$ cat fmod-fixed.cpp

#include <math.h>

typedef double (*fmod_t)(double, double);

fmod_t pmod = fmod;

$ cc -c fmod-fixed.cpp
```


jaykrell:
Wrapper function w/o overloads would also work:

```
#include <math.h>

double mono_fmod(double a, double b)
{
 return fmod(a, b);
}

void * pmod = mono_fmod;
```

jaykrell:

jaykrell:

Issue #9941, opened 2018-08-07
WASM: Use BrowserHttpMessageHandler when available


SteveSandersonMS:
As just discussed with @kumpera, @marek-safar, @kjpou1

Rodrigo's proposal is that the Mono WebAssembly BCL can use reflection to check at runtime whether the assembly `Microsoft.AspNetCore.Blazor.Browser` is loadable, and if so, load the type `Microsoft.AspNetCore.Blazor.Browser.Http.BrowserHttpMessageHandler` from it, and use that as the default `System.Net.Http.HttpMessageHandler` subclass for `HttpClient`. It doesn't take any constructor params, so you can just `Activator.CreateInstance(thatType)` it.

For reference, the sources for `BrowserHttpMessageHandler` are at https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor.Browser/Http/BrowserHttpMessageHandler.cs

More difficult alternative: if you want, Mono could take the `BrowserHttpMessageHandler` sources and use them directly in the Mono WebAssembly BCL. However if you follow this approach, it will necessitate having a reference from the BCL to [the `Mono.WebAssembly.Interop` package](https://www.nuget.org/packages/Mono.WebAssembly.Interop/) (or @kjpou1's version of that assembly, if it's now ready to use). This is fine with me, but it's a more involved solution than Rodrigo's proposal above to reference `BrowserHttpMessageHandler` dynamically via reflection. It also still would only work in the context of Blazor, unless you also extract the [corresponding JS-side code](https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor.Browser.JS/src/Services/Http.ts) and merge that into Mono WebAssembly too. Plus you'd want to get all the tests and have a way of running them too. So I think in the short term at least, the simple dynamic-reflection-loading technique may be much better!

cc @danroth @rynowak

Suchiman:
Related to #7933

jeromelaban:
This direction is a bit too tied to a specific framework/library implementation. Could the implementation be based on something like this instead: 

https://github.com/mono/mono/blob/0bcbe39b148bb498742fc68416f8293ccd350fb6/mcs/class/System.Net.Http/System.Net.Http/HttpClient.android.cs#L18

[...]

PR #10152, opened 2018-08-16
[wasm] Modify HttpClient to use a default handler provider.

kjpou1:
- Add a private static variable to HttpClient called `Func<HttpMessageHandler> GetHttpMessageHandler` that will be used to obtain the default message handler and can be set for example as follows:

```
		var httpMessageHandler = typeof(HttpClient).GetField("GetHttpMessageHandler", 
                            BindingFlags.Static | 
                            BindingFlags.NonPublic);

                httpMessageHandler.SetValue(null, (Func<HttpClientHandler>) (() => {
			return new MyHttpClientHandler ();
		}));
```

- If a handler is not set or returns null the default implementation HttpClientHandler will be returned.

- Add HttpClient.wasm.cs implementation to handle the default HttpClient
- Add simple test to the wasm test harness bindings-test.
- Reference https://github.com/mono/mono/issues/9941

### DIFF
--- a/mcs/class/System.Net.Http/Makefile
+++ b/mcs/class/System.Net.Http/Makefile
@@ -10,9 +10,6 @@ LIB_MCS_FLAGS = $(EXTRA_LIB_MCS_FLAGS)
 ifeq (monodroid,$(PROFILE))
 LIB_MCS_FLAGS += -d:XAMARIN_MODERN
 endif
-ifeq (wasm,$(PROFILE))
-LIB_MCS_FLAGS += -d:WASM
-endif
 
 TEST_LIB_REFS = System System.Core
 TEST_MCS_FLAGS =
--- a/mcs/class/System.Net.Http/Makefile
+++ b/mcs/class/System.Net.Http/Makefile
@@ -10,6 +10,9 @@ LIB_MCS_FLAGS = $(EXTRA_LIB_MCS_FLAGS)
 ifeq (monodroid,$(PROFILE))
 LIB_MCS_FLAGS += -d:XAMARIN_MODERN
 endif
+ifeq (wasm,$(PROFILE))
+LIB_MCS_FLAGS += -d:WASM
+endif
 
 TEST_LIB_REFS = System System.Core
 TEST_MCS_FLAGS =
--- a/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
+++ b/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
@@ -1,33 +1,36 @@
 using System;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
-namespace System.Net.Http {
-	public partial class HttpClient {
+namespace System.Net.Http
+{
+    public partial class HttpClient
+    {
 
         private static Func<HttpMessageHandler> GetHttpMessageHandler;
 
-		public HttpClient ()
-			: this (GetDefaultHandler (), true)
-		{
-		}
+        public HttpClient()
+            : this(GetDefaultHandler(), true)
+        {
+        }
 
-		static HttpMessageHandler GetDefaultHandler ()
-		{
+        static HttpMessageHandler GetDefaultHandler()
+        {
 
             if (GetHttpMessageHandler == null)
                 return GetFallback("No custom HttpClientHandler registered");
 
             var handler = GetHttpMessageHandler();
-			if (handler == null)
-			    return GetFallback ($"Custom HttpMessageHandler is not valid");
+            if (handler == null)
+                return GetFallback($"Custom HttpMessageHandler is not valid");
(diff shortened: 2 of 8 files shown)


mcs/class/System.Net.Http/Makefile
```diff
@@ -10,6 +10,9 @@ LIB_MCS_FLAGS = $(EXTRA_LIB_MCS_FLAGS)
 ifeq (monodroid,$(PROFILE))
 LIB_MCS_FLAGS += -d:XAMARIN_MODERN
 endif
+ifeq (wasm,$(PROFILE))
```
kumpera: This should not be needed, we add `-d:WASM` in `mcs/build/profiles/wasm.make`.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Reflection;
+
+namespace System.Net.Http {
+	public partial class HttpClient {
+
+        public static Func<HttpMessageHandler> GetHttpMessageHandler;
```
kumpera: This should be private. We cannot add public members to BCL types.

sdks/wasm/test.js
```diff
@@ -73,7 +73,7 @@ var Module = {
 	},
 };
 
-var assemblies = [ "mscorlib.dll", "System.dll", "System.Core.dll", "Mono.Security.dll", "main.exe", "nunitlite.dll", "mini_tests.dll", "wasm_corlib_test.dll", "wasm_System_test.dll", "wasm_System.Core_test.dll", "binding_tests.dll" ];
+var assemblies = [ "mscorlib.dll", "System.dll", "System.Core.dll", "System.Net.Http.dll", "Mono.Security.dll", "main.exe", "nunitlite.dll", "mini_tests.dll", "wasm_corlib_test.dll", "wasm_System_test.dll", "wasm_System.Core_test.dll", "binding_tests.dll" ];
```
kumpera: UG, I need to take the time to replace this fugly piece of code to use runtime.g.js :(

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kumpera: Oh! Forgot one thing. We'll need to add GetHttpMessageHandler to the linker xml for System.Net.Http.
@akoeplinger can you help us with this?

kjpou1: How does the variable get set if it is private?

mcs/class/System.Net.Http/Makefile
kjpou1: Indeed it does work.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kumpera: It would be set by reflection.

kjpou1: :-)

kjpou1:
@monojenkins build OS X i386

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
akoeplinger: you can add the `[PreserveDependency ("GetHttpMessageHandler", "System.Net.Http.HttpClient")]` attribute to GetDefaultHandler().

kumpera:
@baulig have you read the associated issue? It covers your questions.

kumpera:
@monojenkins build failed

baulig:
Well, what I really do not like about this is that now several of our platforms each have their own mechanism for providing a default handler instead of this being handled in unified way.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kjpou1: Thanks @akoeplinger

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -0,0 +1,33 @@
+using System;
```
marek-safar: Please use consistent spacing

kjpou1: Cleaned up

kjpou1:
@monojenkins build failed

kjpou1:
Thanks @marek-safar

masudanmurmu:
supported

Issue #7087, opened 2018-02-16
Calling TcpListener.Create(int port) on multiple threads can fail


Tragetaschen:
## Description
I have two ThreadPool threads (`Task.Run(…)`) that both call `TcpListener.Create(int port)` and then `Start()` with different ports each. Sometimes, these two calls fail with
```
System.Net.Sockets.SocketException (0x80004005): Protocol option not supported
  at System.Net.Sockets.Socket.GetSocketOption (System.Net.Sockets.SocketOptionLevel optionLevel, System.Net.Sockets.SocketOptionName optionName) [0x00020] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.Socket.get_DualMode () [0x0001a] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.Socket.get_IsDualMode () [0x0000a] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.Socket.RemapIPEndPoint (System.Net.IPEndPoint input) [0x00000] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.Socket.Bind (System.Net.EndPoint localEP) [0x0001e] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.TcpListener.Start (System.Int32 backlog) [0x00044] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.TcpListener.Start () [0x00000] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
```
and
```
System.Net.Sockets.SocketException (0x80004005): Protocol option not supported
  at System.Net.Sockets.Socket.SetSocketOption (System.Net.Sockets.SocketOptionLevel optionLevel, System.Net.Sockets.SocketOptionName optionName, System.Int32 optionValue) [0x0002f] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.Socket.set_DualMode (System.Boolean value) [0x0001a] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
  at System.Net.Sockets.TcpListener.Create (System.Int32 port) [0x0002a] in <a6842b19ae8449b7a2fd38fe764bcc1c>:0
```
respectively.

If that happened, every call to `TcpListener.Create(int port)` will throw the second exception.

[...]

PR #10158, opened 2018-08-16
[runtime] Synchronize calls to getprotobyname() which is thread-unsafe

alexischr:
Fixes https://github.com/mono/mono/issues/7087

### DIFF
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -33,6 +33,8 @@
 #include <mono/utils/networking.h>
 #include <mono/utils/mono-threads-coop.h>
 
+static mono_mutex_t networking_mutex;
+
 static void*
 get_address_from_sockaddr (struct sockaddr *sa)
 {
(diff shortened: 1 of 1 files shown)


jaykrell:
There must be a way without having private custom locks? So that uncoordinated code in the process can work. You know, what about all the other non-mono code in the process that needs this information? It will race with mono just the same.

jaykrell:
`getprotobyname_r`

alexischr:
@jaykrell I understand, and explored other options and were not sufficient/even less sufficient. There are multiple paths to `fetch_protocol` which calls `getprotobyname` differently, and all of them can be called from any managed thread via a Socket internal call.

getprotobyname_r is the right way to go though, thanks :) If it exists in all our targets I will do that.

jaykrell:
Android getprotobyname appears to always fail, but not certain.

I think you need to autoconfigure or ifdef it.
getprotobyname_r on Linux.
getprotobyname w/o lock on Darwin, or w/ superfluous lock for ease.
getprotobyname w/o lock on Windows it appears, or w/ superfluous lock for ease.
It'd be good to confirm that Android just fails here, or add a lock of uncertain.

Key is that that `_r` should be used if available, and where it isn't available, it seems the code was already ok.

jaykrell:
getprotobyname_r is also on Solaris and OpenBSD.
Use `_r` if present, autoconf, else probably non-`_r` is ok (Darwin, Windows).

alexischr:
Superseded by https://github.com/mono/mono/pull/10163 for now

PR #10160, opened 2018-08-16
[cxx] Remove most G_BEGIN_DECLS but add back to make things work.

jaykrell:
This is a little messy but motivated.

As much as things are compiled as C++ and are not public, extern "C" is not needed.
It is ok, but not needed. We *could* leave all historical G_BEGIN_DECLS in for now.

In the current build support, you cannot say:
  cc -c -xc++ lib.c
  cc -xc++ main.c lib.o

The compiler driver interprets the .o file as text and fails.

So we have:
  cc -c -xc++ lib.c
  cc main.c lib.o

As such, main remains in C for now.
We can drain main into maincpp.cpp or rename, but not now.

For this reason, there remains some C.

Also I didn't port the tests yet. And they use internals.

Also wasm has some known and unknown bindings that depend on function names.
So externC is applied a bit liberally to have WebAssembly pass CI.

We play games where prototypes and functions do not match.
This is with THREAD_INFO_TYPE and MonoGCDescriptor.
There are other solutions, but extern "C" seems ok for now.

We declare OS APIs like ARM cache flush, mmap, sendfile.
We pretty much must provide extern "C" for those.
Hypothetically they could have C++ bindings but it is not likely.

### DIFF
--- a/mono/metadata/gc-internals.h
+++ b/mono/metadata/gc-internals.h
@@ -160,7 +160,9 @@ gboolean mono_gc_user_markers_supported (void);
  * size bytes will be available from the returned address (ie, descr
  * must not be stored in the returned memory)
  */
+G_BEGIN_DECLS // FIXMEcxx? Probably because MonoGCDescriptor varies.
 MonoObject* mono_gc_alloc_fixed      (size_t size, MonoGCDescriptor descr, MonoGCRootSource source, void *key, const char *msg);
+G_END_DECLS // FIXMEcxx?
 void  mono_gc_free_fixed             (void* addr);
 
 /* make sure the gchandle was allocated for an object in domain */
@@ -437,11 +439,15 @@ void mono_gc_register_altstack (gpointer stack, gint32 stack_size, gpointer alts
 
 gboolean mono_gc_is_critical_method (MonoMethod *method);
 
+G_BEGIN_DECLS // FIXMEcxx THREAD_INFO_TYPE varies between declaration and definition
+
 gpointer mono_gc_thread_attach (THREAD_INFO_TYPE *info);
 
 void mono_gc_thread_detach_with_lock (THREAD_INFO_TYPE *info);
 
 gboolean mono_gc_thread_in_critical_region (THREAD_INFO_TYPE *info);
+
+G_END_DECLS // FIXMEcxx THREAD_INFO_TYPE varies between declaration and definition
 
 /* If set, print debugging messages around finalizers. */
 extern gboolean mono_log_finalizers;
--- a/mono/metadata/w32file-win32.c
+++ b/mono/metadata/w32file-win32.c
@@ -370,6 +370,8 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 	return result;
 }
 
+G_BEGIN_DECLS // FIXMEcxx wasm
+
 gboolean
 mono_w32file_get_volume_information (const gunichar2 *path, gunichar2 *volumename, gint volumesize, gint *outserial, gint *maxcomp, gint *fsflags, gunichar2 *fsbuffer, gint fsbuffersize)
 {
@@ -379,6 +381,8 @@ mono_w32file_get_volume_information (const gunichar2 *path, gunichar2 *volumenam
 	MONO_EXIT_GC_SAFE;
 	return res;
 }
+
+G_END_DECLS
 
 #if HAVE_API_SUPPORT_WIN32_MOVE_FILE
 gboolean
--- a/mono/metadata/w32file-win32.c
+++ b/mono/metadata/w32file-win32.c
@@ -370,8 +370,6 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 	return result;
 }
 
-G_BEGIN_DECLS // FIXMEcxx wasm
-
 gboolean
 mono_w32file_get_volume_information (const gunichar2 *path, gunichar2 *volumename, gint volumesize, gint *outserial, gint *maxcomp, gint *fsflags, gunichar2 *fsbuffer, gint fsbuffersize)
 {
@@ -381,8 +379,6 @@ mono_w32file_get_volume_information (const gunichar2 *path, gunichar2 *volumenam
 	MONO_EXIT_GC_SAFE;
 	return res;
 }
-
-G_END_DECLS
 
 #if HAVE_API_SUPPORT_WIN32_MOVE_FILE
 gboolean
--- a/mono/utils/mono-threads.h
+++ b/mono/utils/mono-threads.h
@@ -71,6 +71,8 @@ typedef gsize (*MonoThreadStart)(gpointer);
 
 #endif /* #ifdef HOST_WIN32 */
 
+G_BEGIN_DECLS // FIXMEcxx for wasm
+
 #ifndef MONO_INFINITE_WAIT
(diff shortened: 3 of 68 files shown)


jaykrell:

jaykrell:

jaykrell:

jaykrell:
There are several ways to approach this PR.

1. Accept it now asis. This my preference but there are other options.

2. Accept it later asis. Fix all the compilation errors first, and then link.

3. Deal with some/many/all of the FIXMEs.

 3a. Don't call them FIXMEs, but leave them asis.

 3b. Port monodis, pedump, tests to fully C++.

 3c. Decide to leave monodis, pedump, tests as C indefinitely. Not a terrible situation
     even in a hypothetical very C++ world.

 3d. Determine the exact WebAssembly dependencies on externC.
     Some is obvious/known already, some is not.
     Some of the externC is obviously needed, some I was just
     sprinkling it around until the tests passed -- and are the tests enough?

 3e. The THREAD_INFO_TYPE / SgenDescriptor stuff.

4. Independent of the FIXME resolution, strive for a simpler
   pattern of tending to wrap files, not functions.
   This obviously would not be done 100%, i.e. for OS APIs,
   but for mono.

5. While the entire PR cannot be abandoned, i.e. we need the OS APIs
   wrapped, THREAD_INFO_TYPE, WebAssembly, dealt with, there is
   the idea of leaving in any preexisting unnecessary externC. Just to
   to arbitrarily reduce/delay churn. At least for a while, i.e.
   at least while the tree also compiles as C.

6. Even in a "very C++" world, externC *might* be a desirable
   way to enforce hypothetical policies like "no overloading".
   That would be the route of wrapping every .c file in externC as well, not just .h.
   That loses typesafe linkage, which is arguably ok.

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build OS X i386

mono/metadata/w32file-win32.c
```diff
@@ -370,6 +370,8 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 	return result;
 }
 
+G_BEGIN_DECLS // FIXMEcxx wasm
```
luhenry: Oh, so wasm is using win32 now?

jaykrell: mono_w32file_get_volume_information "has to" be externC, so I marked them all.
"has to" is a probably exaggeration though and could be avoided.

luhenry: How about you mark it in the header like all other functions that need to be marked as `extern "C"`?

jaykrell:
Ok.

1. The thing with wasm though, is that it has some "bindings", where the functions occur in JScript. Kinda like LoadLibrary/GetProcAddress. Some of them are obvious and I made progress with externC on them. However there were still test failures until I put in more externC, I added it until tests passed. I don't know which was really needed.

2. The thing with pedump/monodis/tests is not making the code C++ compatible per se, it is:
  a. Build system support for linking/main.
 We cannot currently say cc -xc++ main.c lib.o
 It treats lib.o as textual source. We need to do one of a few things:
    Rename main.c to main.cpp.
     Leave main.c alone but have main.cpp #include it.
    Move main.c content to maincpp.cpp, leaving main.c just one line.
 I considered sprinkling externC a safer short term tactic,but I can try the others.

b. Having some stuff as C outside the runtime might be good for some, not much, but some, validation of the C embedding API.

I don't think adding MONO_API is the right solution but maybe.
i.e. expose stuff, only for pedump/monodis/tests? So for example they can be dynamically linked?

mono/metadata/w32file-win32.c
jaykrell: "has to" because I was getting all lanes to compile and link and pass all tests.  A bit rushed perhaps. wasm was the only lane with failures, non-obvious test failures, but worked compiled as C, and sprinkling externC around fixed it. Some of the need for externC is obvious but some is not.
It occurs here: https://github.com/mono/mono/blob/master/mono/mini/mini-wasm.c#L580
Yes marking only in header should suffice. I'll try that.

jaykrell:
That isn’t a hack at all. This is wrapping the file, but not includes. Everywhere there is an include, you have to go back to “regular”, #include, and restart “decls”.

Otherwise requires navigating the ifdefs which can be tricky and hard to read. This way is mechanical and easier to understand. Perhaps add a comment though?

jaykrell:
Let see about actually building pedump, monodis, and tests as C++.
And try to track down the wasm problem.
One thing I'm confused on, the OS APIs, those should be G_DECLS, or ifdef __cplusplus extern "C"?
Also I missed _chkstk.

jaykrell:
Ok, the linked PR addresses monodis/monograph/pedump/unit-tests, and will clean this up a bunch.

I think I have a better handle on wasm and will cleanup as a result.

glib.h needs wrapping to compile/link main.c as C which I think is ok.

And there remains the THREAD_INFO_TYPE / SgenDescriptor problem, for which there are multiple other solutions but I'm not sure they are preferable at this time.

Also I missed chkstk.

Let me adjust the PR a bit now..

jaykrell:
If https://github.com/mono/mono/pull/10293 (Fix typesafe-linkage problem due to varying MonoGCDescriptor) and/or https://github.com/mono/mono/pull/10294 (Rework mono-threads.h THREAD_INFO_TYPE parameterization to not violate C++ typesafe linkage) or something like them is accepted, then the FIXMEs can be driven much down further.

mono/metadata/gc-internals.h
```diff
@@ -437,12 +439,16 @@ void mono_gc_register_altstack (gpointer stack, gint32 stack_size, gpointer alts
 
 gboolean mono_gc_is_critical_method (MonoMethod *method);
 
+G_BEGIN_DECLS // FIXMEcxx THREAD_INFO_TYPE varies between declaration and definition
```
jaykrell: This can be removed if e.g. https://github.com/mono/mono/pull/10297 is accepted.

mono/utils/mono-threads.h
```diff
@@ -101,6 +101,40 @@ code ended up looking horrible. So we use this cute little hack. The idea is tha
...
+2. outputs
+
+void*
+mono_threads_out (void);
+
+// Macros do not recurse.
+#define mono_threads_out() \
+	((THREAD_INFO_TYPE*)mono_threads_out ())
+
+For now we use extern "C" instead.
```
jaykrell: https://github.com/mono/mono/pull/10294 offers a solution w/o externC.
There are other solutions.

jaykrell:
Advise? ExternC is much reduced, dependent on some other PRs, and can be much further reduced if linked PRs are accepted.

jaykrell:
The blanketing of glib.h with G_BEGIN_DECLS / G_END_DECLS  can be further reduced via https://github.com/mono/mono/pull/10342
https://github.com/mono/mono/pull/10341

Though one more change is needed.

jaykrell:
The  "last" piece of externC cleanup is https://github.com/mono/mono/pull/10345.
Except for the wasm driver which I'd just as soon leave alone for now.
If all of those go in, then this PR can reduce to just "OS" and "wasm", which I agree is good. We can fix wasm later.

jaykrell:
Also need https://github.com/mono/mono/pull/10368 to reduce externC in glib.

Issue #10031, opened 2018-08-10
unhandled-exception-2.exe crashes crash reporting


alexanderkyte:
There is a unhandled-exception-2.exe crash on CI that seems to hit the assertion at `mini-exceptions.c:1196`. 

I need to investigate this for bugs week. If the domain isn't set, stack walking shouldn't die spectacularly. It should indicate failure to the caller.

alexanderkyte:
@akoeplinger @lambdageek please include any information you know about this that I haven't mentioned.

akoeplinger:
It seems to occur on unhandled-exception-4.exe too ([see Jenkins PR](https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/13540/testReport/junit/MonoTests/test-unhandled-exception-2-255-without-managed-handler/unhandled_exception_4_exe/)), albeit way more infrequently than with -2.exe. Disabled it as well.

alexanderkyte:
It sounds like @lambdageek is debugging this one as well. We agreed that it seemed to be a race between `mono_thread_interruption_checkpoint_void` and getting the info block to pass to `mono_thread_suspend_all_other_threads`

lambdageek:
@alexanderkyte Thanks for helping to track this down! Take a look at #10161 for an attempted fix.

PR #10161, opened 2018-08-16
[coop] Check async suspend status in ReqSuspendInitSuspendBlocking case

lambdageek:
So the issue is that calling `begin_suspend_for_blocking_thread` under hybrid
suspend will perform a preemptive suspension of the victim thread.

The preemptive suspend doesn't always succeed.
In particular, `thread_state_init_from_sigctx` or `thread_state_init_from_handle`
could return FALSE and `info->suspend_can_continue` will be set to FALSE.
When the `thread_state_init_*` functions return FALSE, they also set the
`MonoThreadUnwindState:valid` to FALSE.

So it is important to check the value of `MonoThreadInfo:suspend_can_continue`
before proceeding with a suspension.

As a result in `suspend_sync` in the `ReqSuspendInitSuspendRunning` case we call
`check_async_suspend` which checks (for preemptively suspended threads) whether
`suspend_can_continue` is true.  If it isn't, we resume the victim thread and
loop once more in `suspend_sync_nolock` and try again.

We didn't have a `check_async_suspend` call in the `ReqSuspendInitSuspendBlocking`
case, so a thread could be considered suspended even if we couldn't capture a
valid `MonoThreadUnwindState` for it.

That breaks the assertion the call to `mono_thread_info_get_last_managed` in
`async_suspend_critical` which needs a valid `MonoThreadUnwindState`.

Fixes https://github.com/mono/mono/issues/10031

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -1499,9 +1499,6 @@ CI_PR_DISABLED_TESTS = \
 # FIXME Hybrid Suspend - see https://github.com/mono/mono/issues/9959
 CI_PR_DISABLED_TESTS += sgen-new-threads-collect.exe
 
-# see https://github.com/mono/mono/issues/10031
-CI_PR_DISABLED_TESTS += unhandled-exception-2.exe unhandled-exception-4.exe
-
 # appdomain-threadpool-unload.exe creates 100 appdomains, takes too long with llvm
 LLVM_DISABLED_TESTS = \
 	finally_block_ending_in_dead_bb.exe \
--- a/mono/utils/mono-threads-mach.c
+++ b/mono/utils/mono-threads-mach.c
@@ -24,6 +24,7 @@
 #include <mono/utils/mono-compiler.h>
 #include <mono/utils/mono-threads.h>
 #include <mono/utils/hazard-pointer.h>
+#include <mono/utils/mono-threads-coop.h>
 #include <mono/utils/mono-threads-debug.h>
 
 void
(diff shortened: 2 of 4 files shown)


lambdageek:

lambdageek:
Linux x64 Coop suspend lane hit this assert https://github.com/mono/mono/blob/c114ff59d96baba4479361b2679b7de602517877/mono/utils/mono-threads.c#L181 which is related.  Needs more work.

lambdageek:
OSX (x86 and amd64) has a new crash in `main-returns-background-change.exe`.  Happens pretty consistently on CI (this PR has been stuck for a few days now) but rarely on my machine.  Investigating...

lambdageek:
@luhenry @kumpera Could you take a look at the new change - it's the OSX and win32 companion to https://github.com/mono/mono/commit/006d6ce09734e8b555a53324c21bbaab3f1ebcb4

First the suspend initiator switches the victim from `BLOCKING` to `BLOCKING_SUSPEND_REQUESTED`; and then victim runs and self-suspends to `BLOCKING_SELF_SUSPENDED`.  Finally when the initiator runs `finish_async_suspend` in `mono_threads_suspend_begin_async_suspend` it will find out that async lost the race with self-suspend.  This is okay, and suspend can continue provided that we undo the `SuspendThread` (or `thread_suspend`).  The victim will still be self-suspended and waiting for a resume.

---

This should fix the `main-returns-background-change.exe` failure in this PR.  The reason it started failing is because this PR was making BLOCKING threads check the `info->suspend_can_continue` value and restarting the entire suspend.  That ended up giving the background thread in the test case just enough time to get to a bad exit.

PR #10163, opened 2018-08-17
[runtime] Use getprotobyname_r() where needed and available

alexischr:
Fixes https://github.com/mono/mono/issues/7087

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2287,6 +2287,23 @@ if test x$host_win32 = xno; then
 	])
 
 	dnl **********************************
+	dnl *** Check for getprotobyname_r ***
+	dnl **********************************
+	AC_MSG_CHECKING(for getprotobyname_r)
+		AC_TRY_LINK([
+		#include <stdio.h>
+		#include <netdb.h>
+	], [
+		getprotobyname_r(NULL, NULL, NULL, 0, NULL);
+	], [
+		# Yes, we have it...
+		AC_MSG_RESULT(yes)
+		AC_DEFINE(HAVE_GETPROTOBYNAME_R, 1, [Have getprotobyname_r])
+	], [
+		AC_MSG_RESULT(no)
+	])
+
+	dnl **********************************
 	dnl *** Check for getnameinfo ***
 	dnl **********************************
 	AC_MSG_CHECKING(for getnameinfo)
--- a/mono/utils/networking-posix.c
+++ b/mono/utils/networking-posix.c
@@ -127,7 +127,24 @@ mono_get_address_info (const char *hostname, int port, int flags, MonoAddressInf
 
 #endif
 
-#ifdef HAVE_GETPROTOBYNAME
+#if defined(__linux__) && defined(HAVE_GETPROTOBYNAME_R)
+
+static int
+fetch_protocol (const char *proto_name, int *cache, int *proto, int default_val)
+{
+	if (!*cache) {
+		struct protoent protoent_buf = { 0 };
+		struct protoent *pent = NULL;
+		char buf[1024];
+
+		getprotobyname_r (proto_name, &protoent_buf, buf, 1024, &pent);
+		*proto = pent ? pent->p_proto : default_val;
+		*cache = 1;
+	}
+	return *proto;
+}
+
+#elif HAVE_GETPROTOBYNAME
 
 static int
 fetch_protocol (const char *proto_name, int *cache, int *proto, int default_val)
@@ -141,6 +158,8 @@ fetch_protocol (const char *proto_name, int *cache, int *proto, int default_val)
 	}
 	return *proto;
 }
+
+#endif
 
 int
 mono_networking_get_tcp_protocol (void)
@@ -162,8 +181,6 @@ mono_networking_get_ipv6_protocol (void)
 	static int cache, proto;
 	return fetch_protocol ("ipv6", &cache, &proto, 41); //41 is SOL_IPV6 on linux
 }
-
-#endif
 
 #if defined (HAVE_SIOCGIFCONF)
 


jaykrell:
So, looking closer, the function varies in signature/semantics on other systems.
You do pass a certain number of parameters, to maybe ok asis.
   I'd be nervouse that the C compiler accepts almost anything.
Or maybe just ifdef host_linux.
Or maybe ifdef every system and #error otherwise (ios/windows/osx/android/linux..)
You know, I would want the solaris/openbsd/netbsd/freebsd ports to pay special attention here.

mono/utils/networking-posix.c
```diff
@@ -127,7 +127,24 @@ mono_get_address_info (const char *hostname, int port, int flags, MonoAddressInf
...
+
+static int
+fetch_protocol (const char *proto_name, int *cache, int *proto, int default_val)
+{
+	if (!*cache) {
+		struct protoent protoent_buf;
+		struct protoent *pent;
+		char buf[1024];
+
+		getprotobyname_r (proto_name, &protoent_buf, buf, 1024, &pent);
```
jaykrell: Check the error? Initialize pent?

alexischr: Error doesn't need to be checked because pent will be set to NULL as per https://linux.die.net/man/3/getprotobyname_r

alexischr:
@jaykrell can you link to the information that you're looking about semantics? It's already not going to be used on Windows (I didn't add it to the windows side of `configure.ac`), and FreeBSD is also thread-local-based.

alexischr:
@jaykrell I see OpenBSD's version is just different (https://man.openbsd.org/getprotoent.3), but also has a different signature, so it should just not be detected by autoconf. C compiler accepting it anyway sounds like a obscure concern, but I am sure you have stories :)

If this form is truly linux-specific I can ifdef, but also don't want to accidentally ignore this on some platforms

jaykrell:
Solaris reentrant form with different signature: https://docs.oracle.com/cd/E86824_01/html/E54774/getprotoent-r-3socket.html

OpenBSD ditto: https://man.openbsd.org/getprotoent.3

NetBSD appears no good. Should double check its source and submit fix..

Agreed about FreeBSD.

mono/utils/networking-posix.c
jaykrell: Yuck but ok. I prefer to initialize locals and check errors, and am reluctant to trust documentation when not trusting it is cheap.

krytarowski:
Please do not use these _r functions on NetBSD, they are internal, but due to some mistake symbols went public, without being exposed in public headers.

There are like three or more different implementations (function prototypes) in the wild of these functions.

alexischr:
@krytarowski Thanks for the info! This is now used only for Linux.

mono/utils/networking-posix.c
```diff
@@ -127,7 +127,24 @@ mono_get_address_info (const char *hostname, int port, int flags, MonoAddressInf
 
 #endif
 
-#ifdef HAVE_GETPROTOBYNAME
+#if defined(__linux__) && defined(HAVE_GETPROTOBYNAME_R)
```
jaykrell: Probably could just be ifdef linux, but ok.

configure.ac
```diff
@@ -2286,6 +2286,23 @@ if test x$host_win32 = xno; then
 		AC_MSG_RESULT(no)
 	])
 
+	dnl **********************************
```
jaykrell: Does this work with cross compile?

jaykrell:
The C compiler esp. as run by autoconf accepts a lot of incorrect code, with warnings, and autoconf is then prone to making incorrect decisions. For example prototype checking is often weak, and you can call unprototyped functions. So for example, your autoconf could easily find the existant non-prototyped NetBSD function. I compile some of the autoconf checks as C++ to workaround this, a good idea.

When I said porters should be aware, I meant they should receive #error so they can fix it when they port. OpenBSD for example. So the idea is #ifdef every system we know and understand, and #error otherwise.

jaykrell:
I.e. don’t even use autoconf. Ifded linux use _r. Ifdef freebsd or darwin or android or win use regular. Else #error.

krytarowski:
It's still possible to detect exact function prototype in the implementation, but given that it's dedicated for Linux only, it might not be worth the effort.

alexischr:
@jaykrell I am not going add #error here. It changes this PR from "Linux bugfix" to "possibly breaks platforms that were happily working with the standard `getprotobyname`, which may very well be safe in that environment". We can follow-up

jaykrell:
It also would highlight race condition on NetBSD. Problem is, difficult to
autoconf for the thread local. Could maybe have autoconf create two threads and see if they get the same address, to detect the race.

krytarowski:
For NetBSD it's enough to call it inside a critical section and store the result in a thread local storage.

jaykrell:
Which critical section? There could be multiple such pieces of code each with their own critical section.

krytarowski:
Just use `getprotobyname`(3) through a thread-safe wrapper. I defer the decision on implementation detail, but I see no issue of making it functional through `fetch_protocol()`.

jaykrell:
Such a wrapper does not suffice. The process could have multiple such wrappers and the result is still racy.

krytarowski:
Specifying `#error` for POSIX interface and requiring extensions is controversial. Personally, I think it's even better to go for a workaround with manual synchronization for every platform and do not introduce any `ifdefs`.

jaykrell:
Imagine there are multiple wrappers. 

Nothing forces everything in the process to use the mono wrapper.

They will have separate locks and race each other. 

#error is better than a known (NetBSD) or suspected (other, not Windows, Android, Darwin, Linux) race condition. Race conditions are worse than controversial.

krytarowski:
But we do maintain the code in mono, aren't we? Do you mean that someone might write custom plugins in C?

jaykrell:
Other code in the process, other than mono, could be calling getprotobyname. Or multiple statically linked monos.

krytarowski:
I lost the track a bit, but I sounds arcane to me that someone links multiple static monos into a single binary... but even if someone does it, in static binaries TLS is usually not functional.

krytarowski:
I can wish best luck to maintain multiple garbage collectors running concurrently... in other words this sounds like no issue. Over the years nobody complained and for some controversial decision breaking a platform is odd.

jaykrell:
Statically linking mono does not preclude threads and thread locals from working. 

It is a little esoteric to have multiple monos but it follows from statically linking mono in multiple projects, that just happen to be used in the same process.

Mono & non-mono is less esoteric.

Minimal Posix compliance is not a reason to have thread-unsafety. 

Every implementer except maybe NetBSD recognized this and fixed it.

jaykrell:
Multiple garbage collectors is a thing, indeed, recognized. 

Processes can be running Java and/or JavaScript and/or C# etc. 

There becomes a need to coordinate across them as to
if the overall OS is under memory pressure. Windows has such a mechanism that CLR uses. 

It is needed just to coordinate multiple garbage collectors across multiple processes, not just multiple within a process.

jaykrell:
Lack of complaints does not equal lack of race conditions. Notice someone did find the bug on Linux.

krytarowski:
If something is not broken, there is no need for a fix. Please prompt POSIX group and convince them that their specification is broken.

NetBSD uses this implementation at least since 1983 and POSIX threads for ~20 years and frankly 'nobody' noticed.

If someone is mixing projects and doing it wrongly, it's also not a bug of NetBSD and not a bug of mono. Every user in mono shall use the wrapper.

krytarowski:
If you try to convince me that concurrent g/c is a typical thing (well recognized), I think the whole discussion goes surreal (I would like to see Stop-The-World running concurrently, especially since mono used to use libgc...). Of course there is an option for some classes of g/c that could be independent from threading etc.. but well - this discussion goes to nonsense.

jaykrell:
Java & C# in same process on Android is normal, not surreal. All the languages with GC interop but generally do not share runtime/gc (with a few exceptions like ironpython/iron

jaykrell:
Here is another project hitting the problem I described where autoconf does the wrong thing, on this function, on NetBSD. We are protected here by the ifdef linux.

https://github.com/the-tcpdump-group/libpcap/commit/ec29cc36139d2498cdddf90cde5d02ea84f6af55

krytarowski:
If there would be a good will, there are always some workarounds, such as installing a static version of getprotobyname() linked into mono that is the wrapper... Setting a bar so low (practically to just a priori rejection) for a de facto removal of an OS support is not playing nicely with the community. I'm detaching from this thread.

jaykrell:
Agreed static linking a copy of that function can workaround. A little heavy perhaps.

I was thinking from the point of view of packages/ports. 
The NetBSD porter could fix it for NetBSD. Less good for people that expect mainline to just work out of the box.

 Anyway we should take it up in a NetBSD forum to use
a thread local (first to check the implementation) or provide a reentrant version, like every other OS.

jaykrell:
@monojenkins build failed

alexischr:
@monojenkins build failed

NattyNarwhal:
FWIW, AIX's `getprotobyname_r` seems congruent with OpenBSD's at a quick skim of the manual, at least based on prototype.

alexischr:
@monojenkins build

alexischr:
@monojenkins build

Issue #9318, opened 2018-06-26
watchOS: "fastcc" not supported with Xcode 10's clang


rolfbjarne:
Repro:

* Download Xcode 10 beta 2, and configure Visual Studio for Mac to use it. Restart Visual Studio for Mac.
* Create a new iOS app and add a watchOS extension (or download & open [nofastcc.zip](https://github.com/mono/mono/files/2138071/nofastcc.zip))
* Change to a Release|Device configuration and build.
* Install Xamarin.iOS from the xcode10 branch - either built from source or a package from CI, which are available as GitHub statuses here: https://github.com/xamarin/xamarin-macios/commits/xcode10 (I'm not sure this is required, believe this bug should reproduce with any version of Xamarin.iOS, but I haven't tested it)

Result:

> error MT5209 : Native linking error : fatal error: error in backend: Unsupported calling convention

Complete build log: https://gist.github.com/rolfbjarne/d664ac9440da1e127e97d9779a7bafa2

I opened a radar with Apple about this, and they said:

--- 

This issue behaves as intended based on the following:

The Mono developers decided to use "fastcc". This has never been a supported configuration but happened to "work" before clang-1000 because Clang is compiled with assertions turned off. 

The check was turned into a true error on clang-1000, leading to this failure. 

---

So we need to stop using fastcc.

This is a blocker for our Xcode 10 support, since it breaks our watchOS story completely.

luhenry:
@vargaz could you please take a look at that? Thank you.

marek-safar:
@spouliot what is the dev releases alignment for our xcode 10 support?

vargaz:
Where is the error message about using 'fastcc' ? We are using preserveall_cc, we can get rid of that if needed.

vargaz:
preserve_allcc is a calling convention saving all registers, we use it around calls to init functions so they don't clobber argument registers. Disabling it is going to make the code slower/bigger, but it will still work.

spouliot:
@marek-safar we need the PR backported to `2017-12`, that's what's used in stable 15.7, to fix watchOS support on our next previews. Thanks!

vargaz:
Note that previous xcode versions probably didn't support this calling convention they just didn't assert on it so this will not actually make things slower/bigger.

luhenry:
@spouliot done at https://github.com/mono/mono/pull/9355

GLjackyvaswani:
[...]

PR #9518, opened 2018-07-12
[llvm] Avoid using the preserveall calling convention in another plac…

vargaz:
…e on watchos.

Fixes https://github.com/mono/mono/issues/9318.

PR #10165, opened 2018-08-17
[2018-02] [llvm] Avoid using the preserveall calling convention in another place on watchos. (#9518)

vargaz:
[llvm] Avoid using the preserveall calling convention in another plac…

…e on watchos.

Fixes https://github.com/mono/mono/issues/9318.

backport of https://github.com/mono/mono/pull/9518


lewurm:
this was already backported https://github.com/mono/mono/pull/9578

this PR is wrong anyway? `set_call_preserveall_cc` is already defined above your diff.

Issue #9925, opened 2018-08-07
Further specify the CultureInfo memory lifetimes


alexanderkyte:
Apple's Instruments measured a many-megabyte memory leak associated with the CultureInfo. I fixed this by caching the string. Instruments confirmed that it now only makes a single icall (which on OSX unfortunately allocates a full 16mb because of how the native API is). 

A straightforward reading of how the API is meant to work suggests that my change shouldn't have fixed anything!

For bugs week, I should add some assertions around how this API works and I should measure the memory used. That instruments claimed that a good 15k calls were being made during the BinaryTrees debian shootout benchmark with a Console.WriteLine in the Task tells me that something is really off if it's off. 

Maybe it's us, maybe it's instruments. We should know why though. 

See https://github.com/mono/mono/pull/9862

alexanderkyte:
When I revert ae4bd539443fe194cdc2de648bba00f62824b9b4

![When I revert the commit](https://i.imgur.com/pG6N0BJ.jpg)

alexanderkyte:
If I put an fprintf in the icall, it only prints once. If I set an lldb breakpoint on the icall, or use instruments, it fires constantly. That part of this is clearly due to bugs in instruments. 

What's strange to me is that the commit we merged to cache this string changes what instruments sees. I'm not sure if we're seeing misreporting or if the code is executing differently when it's ptraced.  

I'm going to diagnose why the leak here is reported to be so big. That one seems actionable.

alexanderkyte:
Run under instruments, it prints many times. I would wager there's a lack of visibility of the default string pointer because of broken memory fencing on instruments. This benchmark basically makes Tasks in a loop. 

```
	while (i <= iterations)
					{
						Console.WriteLine("iter {0}", i);
						if (safeDept > 18)
						{
							var split = new[]
							{
								Task.Run(() => (TreeNode.BottomUpTree(safeDept)).ItemCheck()),
								Task.Run(() => (TreeNode.BottomUpTree(safeDept)).ItemCheck())
							};

							i += 2;
							Task.WaitAll(split);
							check += split[0].Result + split[1].Result;
						} else {
							check += (TreeNode.BottomUpTree(safeDept)).ItemCheck();
							i++;
						}
					}
```

And it's this accidental stress on Console.WriteLine that exposed this to me before in Instruments.

alexanderkyte:
[...]

PR #10172, opened 2018-08-17
 Revert "[runtime] Cache culture locale name in managed"

alexanderkyte:


alexanderkyte:
https://github.com/mono/mono/issues/9925

PR #10173, opened 2018-08-17
[runtime] Add option to always use posix locale on darwin

alexanderkyte:


alexanderkyte:
Supports https://github.com/mono/mono/issues/9925

kumpera:
Under what scenarios should the posix locale differ from the Darwin one?

mono/metadata/locales.c
```diff
@@ -559,7 +559,8 @@ get_current_locale_name (void)
 #ifdef HOST_WIN32
 	locale = g_win32_getlocale ();
 #elif defined (__APPLE__)	
-	locale = get_darwin_locale ();
+	if (!g_hasenv ("MONO_DARWIN_FAST_LOCALE"))
```
akoeplinger: I'd name this `MONO_DARWIN_PREFER_POSIX_LOCALE`.

We'll also need to document it in the man page.

migueldeicaza: The POSIX locale is based on environment variables and might not be set when launching outside of a shell (launchd, the open command, iOS apps).

migueldeicaza:
@kumpera The POSIX locale is based on environment variables and might not be set when launching outside of a shell (launchd, the open command, iOS apps).

PR #10178, opened 2018-08-18
[cxx] Wrap fmod in mono_fmod for stronger future-proofing.

jaykrell:
Apparently the standard or implementers would be within their rights to change it to:
double fmod(double, double, double = 1);

which would break this. I don't think they can get away with it.
Taking address is discouraged, so that such evolution can occur compatibilty.


jaykrell:
@monojenkins build

Issue #8871, opened 2018-05-25
Reusing `SocketAsyncEventArgs` for multiple calls to the static `Socket.Connect()`


baulig:
Using the static `Socket.Connect()` multiple times with the same `SocketAsyncEventArgs` throws

```System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException: Operation already in progress
  at System.Net.Sockets.SocketAsyncEventArgs.SetLastOperation (System.Net.Sockets.SocketAsyncOperation op) [0x00033] in /Workspace/mono-master/mcs/class/System/System.Net.Sockets/SocketAsyncEventArgs.cs:196 
  at System.Net.Sockets.SocketAsyncEventArgs.StartOperationWrapperConnect (System.Net.Sockets.MultipleConnectAsync args) [0x00000] in /Workspace/mono-master/mcs/class/System/System.Net.Sockets/SocketAsyncEventArgs.cs:245 
  at System.Net.Sockets.Socket.ConnectAsync (System.Net.Sockets.SocketType socketType, System.Net.Sockets.ProtocolType protocolType, System.Net.Sockets.SocketAsyncEventArgs e) [0x0007f] in /Workspace/mono-master/mcs/class/referencesource/System/net/System/Net/Sockets/Socket.cs:7877 
  at X.TestStaticConnect (System.Net.Sockets.SocketAsyncEventArgs e, System.String host, System.Int32 port) [0x00067] in <a3d5ba1db6c141f58c821248794a8daa>:0 
  at X.TestMultipleConnect (System.Boolean reuse) [0x00124] in <a3d5ba1db6c141f58c821248794a8daa>:0 
   --- End of inner exception stack trace ---
```

Simple example:

```	
	static async Task TestMultipleConnect (bool reuse)
	{
		var port = 8888;
		var socket = new Socket (AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
		socket.Bind (new IPEndPoint (IPAddress.Loopback, port));
		socket.Listen (3);

		ThreadPool.QueueUserWorkItem (_ => {
			var accepted = socket.Accept ();
			Console.WriteLine ($"ACCEPTED: {accepted.RemoteEndPoint}");

			accepted.Close ();

			var accepted2 = socket.Accept ();
			Console.WriteLine ($"ACCEPTED #1: {accepted2.RemoteEndPoint}");
		});

		var e = new SocketAsyncEventArgs ();
[...]

PR #10181, opened 2018-08-20
[System] Release SocketAsyncEventArgs when Socket.ConnectAsync finish immediately

somdoron:
Socket.ConnectAsync doesn't release the SocketAsyncEventArgs in case of exception in the internal call or immediate result.
This make the SocketAsyncEventArgs unusable.
This fix release SocketAsyncEventArgs in case of internal exception (like no network).

Might be related to #8871

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -926,6 +926,7 @@ namespace System.Net.Sockets
 					return false;
 				}
 			} catch (Exception exc) {
+				e.in_progress = 0;
 				e.socket_async_result.Complete (exc, true);
 				return false;
 			}


akoeplinger:
@monojenkins build

akoeplinger:
It seems this breaks a test:

```
SocketTest.ConnectAsyncUnhandledEx:

Error Message
  ConnectedAsync timeout
  Expected: True
  But was:  False

Stacktrace
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in /mnt/jenkins/workspace/test-mono-pull-request-amd64/mcs/class/corlib/System.Reflection/MonoMethod.cs:305 
```

EgorBo:
Related: https://github.com/mono/mono/pull/8913

mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -922,10 +922,12 @@ public bool ConnectAsync (SocketAsyncEventArgs e)
 				}
 
 				if (ares.IsCompleted && ares.CompletedSynchronously) {
+					Interlocked.Exchange (ref e.in_progress, 0);
```
EgorBo: Does it make sense to use Interlocked here?

somdoron: I will change it.

somdoron:
@EgorBo @akoeplinger I think there is another issue here than what I fixed.

I had a case when the ConnectAsync returned false, in .Net when ConnectAsync returns false the complete function is not being called, as in the docs:

> Returns true if the I/O operation is pending. The Completed event on the e parameter will be raised upon completion of the operation.

> Returns false if the I/O operation completed synchronously. In this case, The Completed event on the e parameter *will not be raised* and the e object passed as a parameter may be examined immediately after the method call returns to retrieve the result of the operation.

Following .Net docs (my library is also crossed platform) I assumed the same behavior from Mono and reused the SocketAsyncEventArgs immediately without waiting for the Completed event.

However, because of the call to `e.socket_async_result.Complete (exc, true);` in the `catch` block, the Completed event is being raised while still returning false.

Suggested fixes:
- Return true from the exception handling of ConnectAsync (and maybe other async funtions)
- Return false and not raising the Completed event when the function complete synchronously. This also means to change the failing test.

PR #10193, opened 2018-08-20
[sdks] Fix iOS SDK recent regression.

kumpera:


akoeplinger:
iOS build passed, merging.

Issue #9660, opened 2018-07-20
CoreFX System.Reflection namespace


marek-safar:
## Types and tests in mscorlib System.Reflection namespace to import 

- [ ] AmbiguousMatchException
- [ ] Assembly
- [ ] AssemblyAlgorithmIdAttribute
- [ ] AssemblyCompanyAttribute
- [ ] AssemblyConfigurationAttribute
- [ ] AssemblyContentType
- [ ] AssemblyCopyrightAttribute
- [ ] AssemblyCultureAttribute
- [ ] AssemblyDefaultAliasAttribute
- [ ] AssemblyDelaySignAttribute
- [ ] AssemblyDescriptionAttribute
- [ ] AssemblyFileVersionAttribute
- [ ] AssemblyFlagsAttribute
- [ ] AssemblyInformationalVersionAttribute
- [ ] AssemblyKeyFileAttribute
- [ ] AssemblyKeyNameAttribute
- [ ] AssemblyMetadataAttribute
- [ ] AssemblyName
- [ ] AssemblyNameFlags
- [ ] AssemblyNameProxy
- [ ] AssemblyProductAttribute
- [ ] AssemblySignatureKeyAttribute
- [ ] AssemblyTitleAttribute
- [ ] AssemblyTrademarkAttribute
- [ ] AssemblyVersionAttribute
- [ ] Binder
- [ ] BindingFlags
- [ ] CallingConventions
- [ ] ConstructorInfo
- [ ] CustomAttributeData
- [ ] CustomAttributeExtensions
- [ ] CustomAttributeFormatException
- [ ] CustomAttributeNamedArgument
- [ ] CustomAttributeTypedArgument
- [ ] DefaultMemberAttribute
- [ ] EventAttributes
- [ ] EventInfo
- [ ] ExceptionHandlingClause
- [ ] ExceptionHandlingClauseOptions
- [ ] FieldAttributes
- [ ] FieldInfo
- [ ] GenericParameterAttributes
- [ ] ICustomAttributeProvider
- [ ] ImageFileMachine
- [ ] InterfaceMapping
- [ ] IntrospectionExtensions
- [ ] InvalidFilterCriteriaException
- [ ] IReflect
- [ ] IReflectableType
- [ ] LocalVariableInfo
- [ ] ManifestResourceInfo
- [ ] MemberFilter
- [ ] MemberInfo
- [ ] MemberTypes
- [ ] MethodAttributes
- [ ] MethodBase
- [ ] MethodBody
[...]

PR #10194, opened 2018-08-20
[System.Reflection] CoreFX import fot Pointer type

MaximLipnin:
Part of #9660

### DIFF
--- a/mcs/class/corlib/corert/Type.cs
+++ b/mcs/class/corlib/corert/Type.cs
@@ -1,0 +1,7 @@
+namespace System
+{
+    partial class Type
+    {
+        public bool IsRuntimeImplemented() => true;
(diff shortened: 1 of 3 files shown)


mcs/class/corlib/corert/Type.cs
```diff
@@ -0,0 +1,7 @@
+namespace System
+{
+    partial class Type
+    {
+        public bool IsRuntimeImplemented() => true;
```
akoeplinger: this should be internal

MaximLipnin: addressed

akoeplinger:

MaximLipnin:
Thanks for review!

Issue #9380, opened 2018-06-29
System.Native integration for mobile profiles


marek-safar:
We need to make it work for monotouch and monodroid to enable further CoreFX which are System.Native dependant

PR #10196, opened 2018-08-20
[System.Native] Build a reduced library on Android and iOS

alexischr:
https://github.com/mono/mono/issues/9380

The original PR had a typo (missed ` \` at the end of a line causing a prematurely terminated list of files in `metadata/Makefile.am`), so it had to be reverted

### DIFF
--- a/mono/metadata/Makefile.am
+++ b/mono/metadata/Makefile.am
@@ -148,30 +148,24 @@ libmono_ilgen_la_LIBADD = $(glib_libs) ../utils/libmonoutils.la ../sgen/libmonos
 endif
 endif
 
-if !HOST_WIN32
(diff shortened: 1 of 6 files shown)


luhenry:
This PR has been merged without review while it's clearly a feature. It also contains a mistake: https://github.com/mono/mono/pull/10196/files#diff-65ca3a126a25b819a824e0397c36650aR186 has changed `HOST_WIN32` to `TARGET_WIN32` which breaks the cross-compiler case.

alexischr:
@luhenry the original PR (https://github.com/mono/mono/pull/9575) was approved and this one only fixed a single-letter typo from that one. My mistake is in the original PR though, I will follow up to fix.

PR #10205, opened 2018-08-20
Fix return values in error paths of mono_w32socket_transmit_file.

jaykrell:


jaykrell:

jaykrell:

PR #10210, opened 2018-08-21
[llvm] Clean llvm/llvm_config.mk on configure since it can become stale.

vargaz:


jaykrell:
Can it be made correct intead, i.e. by configure?
And do a move-if-change (from temp), to avoid unnecessary timestamp-based invalidation?

vargaz:
Regenerating it is cheap.

PR #10211, opened 2018-08-21
[cxx] Favor gpointer for weakly typed data, vs. guint8*, gchar*, etc. to reduce casts.

jaykrell:


jaykrell:
One might argue that the type is "array of (pointer to) bytes", or even in some cases, deliberately characters, and that "gpointer" is too weak. If so, I'm ok adding more casts.

jaykrell:

jaykrell:

PR #10212, opened 2018-08-21
[coop] Static coop wrappers.

jaykrell:
Populate icall-def.h with function signatures, including in/out/inout/ref/ptr nuances, and generate wrappers statically with C preprocessor.

This then allows typesafe handles for all platforms.
And the marshal-ilgen code removed.
And hypothetically C++ member functions on handles.

Probably should transform the former marshal-ilgen logic into an offline verification pass, at least done once, due to the in/out/inout/ref/ptr semantics not reflected in the C types.


vargaz:
This looks pretty ugly, what does it save us ?

jaykrell:
If this work is finished, the code for uses_handles in marshal-ilgen.c could be removed.

jaykrell:
Note this [do not merge] -- I'd like to see it work for more functions.

jaykrell:
Oh nevermind, ios must ilgen and then AOT.

mono/metadata/icall-table.h
```diff
@@ -24,4 +24,165 @@ mono_install_icall_table_callbacks (MonoIcallTableCallbacks *cb);
...
+#define MONO_IF_OBJECT_gint64(x) /* nothing */
+#define MONO_IF_OBJECT_gpointer(x) /* nothing */
+#define MONO_IF_OBJECT_MonoBoolean(x) /* nothing */
+#define MONO_IF_OBJECT_MonoObject(x) x
+#define MONO_IF_OBJECT_MonoString(x) x
+#define MONO_IF_OBJECT_MonoW32ProcessStartInfo(x) x
+#define MONO_IF_OBJECT_MonoW32ProcessInfo(x) /* TODO */
+
+// All icall parameter and return types need a handler macro here.
+#define MONO_IF_NOT_OBJECT(type, x) MONO_IF_NOT_OBJECT_##type (x)
```
lambdageek: If this is going to go forward, I think this `MONO_IF_XXX` idea won't scale well - need some case/switch macros, not just booleans.

Because there are actually 5 different ways to pass arguments to icalls - just "object/not object" is not enough:
https://github.com/mono/mono/blob/d8a5d891be97c333fa13b6ca2918a5c0a829c729/mono/metadata/marshal-ilgen.c#L6160-L6176

1. No wrapping - roughly `MONO_IF_NOT_OBJECT` - this is for primitives.
2. Wrap object - roughly `MONO_IF_OBJECT` - this is pointers to managed objects.
3. Wrap object inout - for reference types passed by `ref` - we copy back the value from the handle when the icall returns
4. Wrap object out - for reference types passed by `out` - we initialize the handle with null (this is needed because the `out` param usually has garbage in it, but GC scanning of coop handles is precise, so garbage leads to crashes).
5. Valuetype ref - for value types passed by `out` or `ref` - we pin the address of the valuetype using an interior coop handle (this is pretty much the only place where we use them) and then pass the _raw pointer_ to managed.  This is for things like `MonoW32ProcessStartInfo` and the various reflection `Info` structs, and so on.

lambdageek:
I wonder if instead of trying to infer how to marshal based on the type, maybe it's better to wrap the type in a keyword:
``` HANDLE2 (IC_2, ves_icall_my_icall, OBJ (MonoObject), 3, (VAL (guint32), OBJ (MonoString), OBJOUT (MonoArray)))```
and then you can do token-pasting for each arg to make `MONO_MARSHAL_OBJ`, `MONO_MARSHAL_OBJOUT`, `MONO_MARSHAL_VAL` macros.

I guess the downside is that everyone writing an icall will have to be a little bit of an expect on calling conventions.  But that's true already since to use the `HANDLES()` macro today, you have to know the correct signature for your unmanaged declaration.

**Comment by lambdageek** [comment]
Another option is to ditch `icall-def.h` and write a data file in some easy to consume format and generate `.c` and `.h` files using a more powerful generator.

**Comment by vargaz** [comment]
The wrappers also have to do stuff like throw exceptions etc., so they cannot be expressed in c code.

**Comment by jaykrell** [comment]
@lambageek Assuming 2 cases are most common, the others can be handled with occasional typedefs. The macros are driven by type names, not types.

Handle- function could also do some work, leaving automation for just overwhelmingly common cases, if overwhelming,  & if wrapper isn’t wrong, just insufficient.

I like cpp, if it suffices. I suspect it does here. But other ok.

Can you point at examples of each?

@vargaz Does the set_pending_exception not cover that?

**Comment by vargaz** [comment]
It also has to set up support for stack walks by pushing an LMF frame, which is done by JITted code.

**Comment by jaykrell** [comment]
Maybe the wrapping in how to pass is better. It could interleave with the type parameters probably, more flexible and easier to understand than wrapping. I can try that, if people like the general idea.

Can LMF be established in C?

Are these wrappers more efficient, assuming equally optimizing compilers? Or equivalent?

Is ilgen kinda a good hammer or bad hammer? I’m less comfortable with dynamic stuff, but maybe just me, and once you have a JIT, just use it?

**Comment by jaykrell** [comment]
I guess wrapping types, is a little better, if the 2 cases are overwhelmingly common. The default wrapper would be nothing. The other wrappers could just generate other names and fall through.

I need to understand all the ways to marshal, really. I’ll study it more later, since already explained.

**Comment by jaykrell** [comment]
1. Having the signatures here, might let us improve static checking.
Though that still isn't easy, involved matching up type-driven macros and the C# metadata.
2. It will let us have correct prototypes, though externC is still desirable, when we export icalls.
3. It should help generate wrappers for the icalls that are not in icall-def.h, the ones called by the JIT but not by C#. Though another solution there would be to declare things in C# I think.
4. We can optimize away the frames here, I have that. But the dynamic one could too, or maybe does.

Could probably bootstrap all of this from the metadata, as well.

**Comment by jaykrell** [comment]
The point isn't optimization, though the static wrappers would probably be smaller, and involve viewer managed/native transitions (once per icall, and no longer once per managed parameter, i.e. the calls mono_handle_new would be only on the native side).

This  also provides for typesafe handles on all platforms -- removing the understanding of coop handles from marshal-ilgen.c entirely. That should  be possible otherwise but is tricky (teaching marshal-ilgen.c to marshal the handles as structs, I tried a bit).

I understand the in/out/inout stuff now and have a local fix.
I kinda understand the generic case too. In either case, I have this locally refactored to switch generally on the name of the parameter and return types, instead of their object-ness.

**Comment by jaykrell** [comment]
Relevant link  errors in SDK build. Will fix later.

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx-products-sdks-ios/5275/parsed_console/log.html

File: mono/metadata/marshal-ilgen.c [hunk 2]
```diff
@@ -6141,168 +6141,14 @@ emit_create_string_hack_ilgen (MonoMethodBuilder *mb, MonoMethodSignature *csig,
 	mono_mb_emit_byte (mb, CEE_RET);
 }
 
-/* How the arguments of an icall should be wrapped */
```
jaykrell: These comments should be preserved somewhere. I.e. in icall-table.h.

mono/metadata/handle-decl.h
```diff
@@ -36,16 +36,7 @@
...
-// There is not runtime cost.
-// NOTE: Running this code depends on the ABI to pass a struct
-// with a pointer the same as a pointer. This is tied in with
-// marshaling. If this is not the case, turn off type-safety, perhaps per-OS per-CPU.
-#if defined (HOST_DARWIN) || defined (HOST_WIN32) || defined (HOST_ARM64) || defined (HOST_ARM) || defined (HOST_AMD64)
-#define MONO_TYPE_SAFE_HANDLES 1
-#else
-#define MONO_TYPE_SAFE_HANDLES 0 // PowerPC, S390X, SPARC, MIPS, Linux/x86, BSD/x86, etc.
-#endif
+// marshal-ilgen.c did not know how to marshal type unsafe handles. And so they are gone now.
```
jaykrell: Comment is backwards. Maybe further rewrite comments.

mono/metadata/handle.h
```diff
@@ -500,8 +458,6 @@ TYPED_HANDLE_DECL (MonoObject);
 TYPED_HANDLE_DECL (MonoException);
 TYPED_HANDLE_DECL (MonoAppContext);
 
-#if MONO_TYPE_SAFE_HANDLES
-
 // Structs cannot be cast to structs.
```
jaykrell: Comment is outdated, that technique didn’t work with wasm.

mono/metadata/icall-table.c
```diff
@@ -32,10 +32,21 @@
...
 #endif
 
-#include <mono/metadata/icall-table.h>
 #include <mono/utils/mono-publib.h>
 #include <mono/utils/bsearch.h>
 #include <mono/metadata/icalls.h>
+#include "handle-decl.h"
+
+/* Baked typed handles we all want */
+TYPED_HANDLE_DECL (MonoString);
```
jaykrell: I think these were in handle-decl.h. Now duplicated. Back to handle-decl.h?

mono/metadata/icall-table.h
```diff
@@ -24,4 +24,479 @@ mono_install_icall_table_callbacks (MonoIcallTableCallbacks *cb);
...
+typedef WSABUF *WSABUF_ptr;
+
+#ifndef MonoResolveTokenError
+#define MonoResolveTokenError MonoResolveTokenError
+typedef gint32 MonoResolveTokenError;
+#endif
+
+typedef gint32 MonoProtocolType;
+
+typedef const char *const_char_ref;
```
jaykrell: Some of these are not used.

mono/metadata/handle-decl.h
```diff
@@ -65,26 +57,14 @@ Handle macros/functions
  *
  * For example, TYPED_HANDLE_DECL(MonoObject) (see below) expands to:
  *
- * #if MONO_TYPE_SAFE_HANDLES
```
jaykrell: Backwards.

jaykrell:
It looks like the movement of types  around might not be needed. I'm experimenting w/o that.

jaykrell:
All the type movement should instead just get the includes, and *one* type movement demonstrated here: https://github.com/mono/mono/pull/11019

jaykrell:
This is fairly redundant with master now since https://github.com/mono/mono/pull/11294.

Issue #9973, opened 2018-08-08
Update framework facades list used by loader


marek-safar:
We have added few assemblies recently but didn't update https://github.com/mono/mono/blob/e8c3dc775bf4a9c6e5f313702fcdd94fc38fd21d/mono/metadata/assembly.c#L97 accordingly (at least we should document mobile only additions)

Related PR https://github.com/mono/mono/pull/9968

PR #10214, opened 2018-08-21
Update framework assemblies list in assembly.c

akoeplinger:
Fixes https://github.com/mono/mono/issues/9973


akoeplinger:
@monojenkins build pkg

akoeplinger:
@monojenkins build pkg

mono/metadata/assembly.c
```diff
@@ -254,8 +270,8 @@ static const AssemblyVersionMap framework_assemblies [] = {
 	FACADE_ASSEMBLY ("System.Security.Cryptography.Encryption.Aes"),
 	FACADE_ASSEMBLY ("System.Security.Cryptography.Encryption.ECDiffieHellman"),
 	FACADE_ASSEMBLY ("System.Security.Cryptography.Encryption.ECDsa"),
-	FACADE_ASSEMBLY ("System.Security.Cryptography.Encryption.Hashing"),
-	FACADE_ASSEMBLY ("System.Security.Cryptography.Encryption.Hashing.Algorithms"),
+	FACADE_ASSEMBLY ("System.Security.Cryptography.Hashing"),
+	FACADE_ASSEMBLY ("System.Security.Cryptography.Hashing.Algorithms"),
```
lambdageek: what happened here?  Were the ones with `Encryption` in the name incorrect?

akoeplinger: yup, it's the namespace, not the assembly name ;)

akoeplinger:
@monojenkins build pkg

mono/metadata/assembly.c
```diff
@@ -310,6 +339,7 @@ static const AssemblyVersionMap framework_assemblies [] = {
 	FACADE_ASSEMBLY ("System.Xml.XmlSerializer"),
 	FACADE_ASSEMBLY ("System.Xml.Xsl.Primitives"),
 	{"WindowsBase", 3},
+	{"cscompmgd", 0},
```
marek-safar: I don't think we should add assemblies we know we'll never add/support

akoeplinger: @marek-safar https://github.com/mono/mono/blob/master/mcs/class/Cscompmgd/Makefile 😄

marek-safar: This was just one example, there are more like System.Web.RegularExpressions etc

akoeplinger: I only added assemblies which we're building as part of net_4_x. I think we should add them all for consistency regardless of whether they are "useful".

PR #10224, opened 2018-08-21
[interp] fix ftnptr_eh_callback

lewurm:
please look at the single commits. here's a `git log` copy/paste:

---------------------------------

[exceptions] ftnptr_eh_callback test that stresses n2m transitions.

on Xamarin.iOS it happens that the ftnptr_eh_callback calls into managed code, that is yet another native-to-managed transition.

----------------------------------

[exceptions] pop lmf stack before executing ftnptr_eh_callback

----------------------------------

[interp] always set LMF for interp_in wrapper.

partial revert of a33cbb873c487490f2a3c271b24c70c1c3fc0a2c

----------------------------------

[mini] ignore LMF if it isn't initialized

interp_in wrapper serves two purposes:

1) native to interp transition: usually this transition doesn't need a
    LMF marker, but in a ftnptr_eh_callback scenario it can be required.
    see ./mono/tests/install_eh_callback.cs
2) jitted code to interp. LMF marker is used to signal EH the transition.

The change of this commit is needed if we have a wrapper of kind 1), and it's called from native code by a thread that isn't attached to the runtime yet. Therefore LMF won't be initialized yet.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -334,6 +334,33 @@ mono_interp_get_imethod (MonoDomain *domain, MonoMethod *method, MonoError *erro
 	return rtm;
 }
 
+#if defined (MONO_CROSS_COMPILE) || defined (HOST_WASM) || defined (_MSC_VER)
+#define INTERP_PUSH_LMF_WITH_CTX_BODY(ext, exit_addr) /* do nothing */
+#elif defined (MONO_ARCH_HAS_MONO_CONTEXT)
+#define INTERP_PUSH_LMF_WITH_CTX_BODY(ext, exit_addr) \
+	MONO_CONTEXT_GET_CURRENT ((ext).ctx); \
+	MONO_CONTEXT_SET_IP (&(ext).ctx, (exit_addr)); \
+	mono_arch_do_ip_adjustment (&(ext).ctx);
+#else
+#define INTERP_PUSH_LMF_WITH_CTX_BODY(ext, exit_addr) g_error ("requires working mono-context");
+#endif
+
+/* INTERP_PUSH_LMF_WITH_CTX:
+ *
+ * same as interp_push_lmf, but retrieving and attaching MonoContext to it.
+ * This is needed to resume into the interp when the exception is thrown from
+ * native code (see ./mono/tests/install_eh_callback.exe).
+ *
+ * This must be a macro in order to retrieve the right register values for
+ * MonoContext.
+ */
+#define INTERP_PUSH_LMF_WITH_CTX(frame, ext, exit_addr) \
+	memset (&(ext), 0, sizeof (MonoLMFExt)); \
+	(ext).kind = MONO_LMFEXT_INTERP_EXIT_WITH_CTX; \
+	(ext).interp_exit_data = (frame); \
+	INTERP_PUSH_LMF_WITH_CTX_BODY ((ext), (exit_addr)); \
+	mono_push_lmf (&(ext));
+
 /*
  * interp_push_lmf:
  *
@@ -346,7 +373,7 @@ static void
 interp_push_lmf (MonoLMFExt *ext, InterpFrame *frame)
 {
 	memset (ext, 0, sizeof (MonoLMFExt));
-	ext->interp_exit = TRUE;
+	ext->kind = MONO_LMFEXT_INTERP_EXIT;
 	ext->interp_exit_data = frame;
 
 	mono_push_lmf (ext);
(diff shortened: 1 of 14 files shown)


BrzVlad:
@lewurm The interp lmf_wrapper was added so you can unwind directly into the calling code (with the register state at the site), so it's main purpose is to save the register context. I don't see how this is used for the same purpose here. I'm guessing it just adds a dummy lmf structure that is for whatever reason expected somewhere for the EH to continue unwinding ?

mono/mini/method-to-ir.c
```diff
@@ -1740,11 +1740,14 @@ emit_push_lmf (MonoCompile *cfg)
 	/*
 	 * Emit IR to push the LMF:
 	 * lmf_addr = <lmf_addr from tls>
-	 * lmf->lmf_addr = lmf_addr
-	 * lmf->prev_lmf = *lmf_addr
-	 * *lmf_addr = lmf
+	 * if (lmf_addr != NULL) {
+	 *   lmf->lmf_addr = lmf_addr
```
vargaz: Why is this needed ? The lmf should only be null if the thread is not attached to the JIT.

lewurm: the `interp_in` wrapper is called _before_ the `native-to-managed` wrapper which would to do the attaching. if it's called from an unattached thread, we don't care about the LMF in this wrapper; in other scenarios we might do.

vargaz: Not sure why its called before, but if that really happens, then the null check should be emitted only in interp_in wrappers, not everywhere.

mono/mini/interp/interp.c
```diff
@@ -334,6 +334,28 @@ mono_interp_get_imethod (MonoDomain *domain, MonoMethod *method, MonoError *erro
...
+
+static MONO_NEVER_INLINE void
+interp_push_lmf_with_ctx (MonoLMFExt *ext, InterpFrame *frame)
+{
+	memset (ext, 0, sizeof (MonoLMFExt));
+	ext->interp_exit = 2;
+	ext->interp_exit_data = frame;
+
+#ifdef MONO_INIT_CONTEXT_FROM_CURRENT
+	MONO_INIT_CONTEXT_FROM_CURRENT (&ext->ctx);
```
BrzVlad: @lewurm I don't understand the point of this define path

lewurm: it's copy/paste from other places that are using `MONO_INIT_CONTEXT_FROM_FUNC`. we could probably remove everything that mentions `MONO_INIT_CONTEXT_FROM_CURRENT` from the code base, no architecture seems to define it.

lewurm:
My previous attempt was wrong: Previously, I tried to sneak in a `MonoContext` via an LMF wrapper on the native2interp transition. That worked by chance for this testcase, but is wrong and broke the debugger. As @BrzVlad pointed out to me on Slack, we should rather inject `MonoContext` on the interp2native transition which is what the updated PR does. We could have done it with the LMF wrapper as well, but due to `MINT_CALLI_NAT_FAST` (a fast path optimization for common signatures of icalls/pinvokes) we would need to handle a couple different signatures. I tried to avoid that with my solution.

The testcase is unchanged, so I keep @lambdageek's review for that. @BrzVlad @vargaz please have a look at the native bits.

mono/mini/interp/interp.c
```diff
@@ -334,6 +334,33 @@ mono_interp_get_imethod (MonoDomain *domain, MonoMethod *method, MonoError *erro
 	return rtm;
 }
 
+#ifdef MONO_ARCH_HAS_MONO_CONTEXT
+#define INTERP_PUSH_LMF_WITH_CTX_BODY(ext, exit_addr) \
+	MONO_CONTEXT_GET_CURRENT ((ext).ctx); \
+	MONO_CONTEXT_SET_IP (&(ext).ctx, (exit_addr)); \
+	mono_arch_do_ip_adjustment (&(ext).ctx);
```
BrzVlad: @lewurm I'm assuming the only purpuse of ip adjustment is for EH to undo it, right ?

lewurm: correct

lambdageek:
Discussed with @lewurm - the coop failures are related to the new test but not to the code changes.  We should take this PR and fix the coop breakage separately. (it's a checked build failure - the cookies on the stack got out of sync)

lewurm:
removed the test case from this PR and moved it to https://github.com/mono/mono/pull/10337

Here are some CI runs where the interpreter changes worked for mentioned test case:
* https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/9285/
* https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-aarch64/7751/
* https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-armhf/7465/
* https://jenkins.mono-project.com/job/test-mono-pull-request-wasm/4228/
* https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx-products-sdks-ios/4247/

Failing coop jobs:
* https://jenkins.mono-project.com/job/test-mono-pull-request-coop-arm64/3592/
* https://jenkins.mono-project.com/job/test-mono-pull-request-coop/7310/

lewurm:
@monojenkins rebase

lewurm:
@monojenkins backport 2018-06

lewurm:
@monojenkins backport 2018-08

PR #10231, opened 2018-08-22
[cxx] int/enum alternative -- typedef to int,

jaykrell:
This is really good for source compat,
pretty good for semantic compat, and not
so good for debugging and typesafety.

Perfect semantic compat, it turns out, is impossible.
The underlying type of an enum varies
in C vs. C++ and per compiler.

Things like ((enum A)-1) < 0 are not compiler-portable
or C vs. C++-portable.

In particular, msvc errs toward int,
clang can be unsigned, clang++ toward int.
We do not have the same types across compilers,
so varying types for C++ seems ok.


jaykrell:
I think we are going to go with operator overloads instead of this.
I will present that PR also.

jaykrell:
This is an option though and keep it in mind for a bit.

jaykrell:
https://github.com/mono/mono/pull/10274 instead.

Issue #10235, opened 2018-08-22
Mono’s System.Windows.Forms implementation has new bugs in V5.14


oldtomdevel:
The release notes for V5.14 state that "Filip Navara has contributed significant improvements to Mono’s System.Windows.Forms implementation."
However, there are some functions that used to work fine till V5.12 but do not work any more in V5.14.
In my WinForms Applications the following functions are broken now due to the changes in the library:

1. If a font is assigned to a Form or directly to a TextBox or NumericUpDown when the form is not yet shown, the control does not always take over that assignment.
2. A ComboBox with DropDownStyle=DropDownList and a lot of entries does not show correctly the automatic vertical scroll bar.
3. There are layout problems in a TableLayoutPanel i.e. wrong positioning of controls.

See attached example. 

[Test.zip](https://github.com/mono/mono/files/2309955/Test.zip)

Execute the attached exe on Mono 5.14 and on Mono 5.12. Compare the results:
1. font of NumericUpDown (row #0) and font of TextBox (row #2).
2. scrollBar in ComboBox (row #1).
3. position of TextBox within TableLayoutPanel (row #2).

The behaviour of V5.12 was okay (and more or less as in Windows .Net), the bahaviour of V5.14 causes problems.

filipnavara:
I will look into it.

filipnavara:
Ok, I could easily reproduce 1 and 2. I don't see what's wrong with the table layout. Could you elaborate on that? On my macOS test setup it looks identical on 5.12 and 5.16 (which should not be different from 5.14 afaik).

oldtomdevel:
I have attached a pic to show how it looks like on my system.

As one can expect from inspecting the source code,
the TextBox in the last line should be vertically centered
in the last line.
But in fact it is positioned too much towards the bottom.
This observation is true since Mono 5.14.
![wrong](https://user-images.githubusercontent.com/21215858/44519237-c4cb5800-a6cc-11e8-9bf2-1584ea871e7c.png)

filipnavara:
I do see the text box looks a little bit lower. However, when I resize the form it remains in the relative vertical position to the "label3" and vertically centred.

oldtomdevel:
The TextBox is lower than it should be.

When the form is resized, the relative position of TextBox to label3 remains the same.
That is, when making the form bigger, the absolute position error is the same but the relative position error within the row decreases (at some size of the form you might get the wrong impression that everything were okay).

[...]

PR #10237, opened 2018-08-22
System.Windows.Forms:  Fix changing TextBox font before the Control handle has been created

filipnavara:
Fixes one regression reported in #10235.


filipnavara:
Yes, it is, it would need to access the internal structures... However, there are already simple UI test apps for manual testing, so I can add it there.

filipnavara:
I'll just merge it all into one PR and add the test there. All the changes are 1 to 5 liners and it will be easier to backport it.

PR #10239, opened 2018-08-22
[RFS][SDKs] Add configure script and update top level Makefile to use it.

kumpera:
This still needs the iOS / Android targets to append to `TARGETS` but we should make it so that the standard `all, build, package, test, clean` targets work just fine in a well configured SDKs setup.

### DIFF
--- a/sdks/configure
+++ b/sdks/configure
@@ -1,0 +1,80 @@
+#!/bin/sh -e
+
+function show_help () {
+cat <<EOL
+Usage: configure [options]
+    -h, --help       Displays this help
+    --ios            Enable IOS build (disabled by default)
+    --android        Enable Android build (disabled by default)
+    --wasm           Enable WebAssembly build (disabled by default)
+	--desktop        Enable Desktop build (disabled by default)
+    --disable-bcl    Disable BLC build (enabled by default)
+EOL
+}
+
+CONFIGURED_FILE=Make.local
+
+rm -f $CONFIGURED_FILE
+
+echo "# Configure arguments: $*" >> $CONFIGURED_FILE
+
+disable_ios=true
+disable_android=true
+disable_wasm=true
+disable_desktop=true
+disable_bcl=false
+
+
+while test x$1 != x; do
+	case $1 in
+	--ios)
+		$disable_ios = false
+		shift
+		;;
+	--android)
+		$disable_android = false
+		shift
+		;;
+	--wasm)
+		$disable_wasm = false
+		shift
+		;;
+	--desktop)
+		$disable_desktop = false
+		shift
+		;;
+	--disable-bcl)
+		$disable_bcl = false
+		shift
+		;;
+	--help|-h)
+		show_help
+		exit 0
+		;;
+	*)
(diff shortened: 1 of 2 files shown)


kumpera:
cc @vargaz @luhenry

sdks/configure
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh -e
+
+function show_help () {
+cat <<EOL
+Usage: configure [options]
+    -h, --help       Displays this help
+    --ios            Enable IOS build (disabled by default)
+    --android        Enable Android build (disabled by default)
+    --wasm           Enable WebAssembly build (disabled by default)
+	--desktop        Enable Desktop build (disabled by default)
```
luhenry: Mismatching indentation.

sdks/configure
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh -e
+
+function show_help () {
+cat <<EOL
+Usage: configure [options]
+    -h, --help       Displays this help
+    --ios            Enable IOS build (disabled by default)
```
luhenry: Should we use `--enable-{ios,android,wasm}` to be consistent with `--disable-bcl`?

jaykrell:
I wonder if configure is a bad thing really, to be avoided.
Instead configuration is implied by make target or directory, and all configurations are always available.
Once you have configure, you then want to cd and build in a different tree, for example, to keep the configures separate.

sdks/configure
```diff
@@ -0,0 +1,80 @@
...
+    --android        Enable Android build (disabled by default)
+    --wasm           Enable WebAssembly build (disabled by default)
+	--desktop        Enable Desktop build (disabled by default)
+    --disable-bcl    Disable BLC build (enabled by default)
```
jaykrell: `BLC` vs. `BCL`.

jaykrell:
Well, at least it isn't autoconf, so maybe configure isn't so bad.

kumpera:
@jaykrell the default is to build everything plus there are specific make targets for specific configurations.

What the configure script does is disable everything and only enable what one asks for. This is so one can use the catch-all targets without waiting forever.

Take wasm, for example.

Using explicit targets I'd need to build `package-wasm-cross package-wasm-runtime`. Again and again and again as I work on it.

With the configure script I'd do `./configure --enable-wasm` and the only target I need to care is `package`.

The other nice property of this is that this knowledge of target is transferable. Say I wanted to work on Android, I don't need to know the specific android targets I need, just that `./configure --enable-android` would pick them for me.

jaykrell:
I think the point of configure or Make.config is/should-be to shorten command lines and that is valuable.

I would request, if possible/easy, something like, that changing configure does not require a clean build, nor an expensive incremental build (unless I never built that configuration before).

That configure just change where plain `make` with fairly short command line, cd's to and/or which targets it builds. Not that if effects, like, the contents of output files.

Perhaps that is already the direction.

kumpera:
That is already the direction.

Right it works because it just filters which builds to trigger, but if we end up adding higher level options (IE build with heavy assertions) this might not work and we'll have to reassess the situation.

PR #10246, opened 2018-08-22
[mini] Don't regen Makefile.am if llvm/llvm_config.mk changes

lambdageek:
The llvm/llvm_config.mk makefile fragment just includes make (not automake)
variable definitions.  So there is no need to regenerate the mini/Makefile.am
file if the fragment updates - the include will pick it up.


lambdageek:

PR #10036, opened 2018-08-10
[cxx] Make some code valid C++, esp. adding casts and esp. not touching allocators.

jaykrell:

PR #10252, opened 2018-08-23
[cxx] Make some code valid C++, esp. adding casts and esp. not touching allocators. alternative to 10036

jaykrell:
This is an alternative to https://github.com/mono/mono/pull/10036.
Instead of macros, or casts, it uses templates.
The line damage of the casts does not seem good to me and I prefer templates or macros.

### DIFF
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -1068,9 +1068,6 @@ MonoJitICallInfo *
 mono_register_jit_icall_full (gconstpointer func, const char *name, MonoMethodSignature *sig, gboolean no_wrapper, const char *c_symbol);
 
 #ifdef __cplusplus
-extern "C++" // in case of surrounding extern "C"
-{
-
 template <typename T>
 inline MonoJitICallInfo *
 mono_register_jit_icall (T func, const char *name, MonoMethodSignature *sig, gboolean is_save)
@@ -1084,8 +1081,6 @@ mono_register_jit_icall_full (T func, const char *name, MonoMethodSignature *sig
 {
 	return mono_register_jit_icall_full ((gconstpointer)func, name, sig, no_wrapper, c_symbol);
 }
-
-} // extern "C++"
 #endif // __cplusplus
 
 void
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -1067,6 +1067,27 @@ mono_register_jit_icall (gconstpointer func, const char *name, MonoMethodSignatu
 MonoJitICallInfo *
 mono_register_jit_icall_full (gconstpointer func, const char *name, MonoMethodSignature *sig, gboolean no_wrapper, const char *c_symbol);
 
+#ifdef __cplusplus
+extern "C++" // in case of surrounding extern "C"
+{
+
+template <typename T>
+inline MonoJitICallInfo *
+mono_register_jit_icall (T func, const char *name, MonoMethodSignature *sig, gboolean is_save)
+{
+	return mono_register_jit_icall ((gconstpointer)func, name, sig, is_save);
+}
+
+template <typename T>
+inline MonoJitICallInfo *
+mono_register_jit_icall_full (T func, const char *name, MonoMethodSignature *sig, gboolean no_wrapper, const char *c_symbol)
+{
+	return mono_register_jit_icall_full ((gconstpointer)func, name, sig, no_wrapper, c_symbol);
+}
+
+} // extern "C++"
+#endif // __cplusplus
+
 void
 mono_register_jit_icall_wrapper (MonoJitICallInfo *info, gconstpointer wrapper);
 
--- a/mono/metadata/threads-types.h
+++ b/mono/metadata/threads-types.h
@@ -44,7 +44,8 @@ typedef enum {
 } MonoThreadApartmentState;
 
 typedef enum {
-// These values match Windows, but are offset by 2.
+// These values match System.Threading.ThreadPriority.
+// These values match System.Diagnostics.ThreadPriorityLevel and Windows, but are offset by 2.
 	MONO_THREAD_PRIORITY_LOWEST       = 0,
 	MONO_THREAD_PRIORITY_BELOW_NORMAL = 1,
 	MONO_THREAD_PRIORITY_NORMAL       = 2,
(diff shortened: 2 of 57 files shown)


mono/metadata/threads-types.h
```diff
@@ -117,7 +119,7 @@ ICALL_EXPORT
 int ves_icall_System_Threading_Thread_GetPriority (MonoThreadObjectHandle this_obj, MonoError *error);
 
 ICALL_EXPORT
-void ves_icall_System_Threading_Thread_SetPriority (MonoThreadObjectHandle this_obj, int priority, MonoError *error);
+void ves_icall_System_Threading_Thread_SetPriority (MonoThreadObjectHandle this_obj, MonoThreadPriority priority, MonoError *error);
```
luhenry: You have no guarantee on the size of the enum in Native, while managed will always pass an int. Revert that change.

jaykrell:
`enum` will be `int` or `unsigned` here but ok.

jaykrell:
Another tradeoff is `typedef MonoThreadPriority int`.
It loses typesafety and debugging, but works, is compatible and kinda looks nice like the previous.
C already had no typesafety here. But I'll just put it back.

jaykrell:
Also `loader.h` don't touch, move to `loader-internals.h` or other.

jaykrell:

jaykrell:

jaykrell:
@monojenkins build OS X i386

mono/mini/aot-runtime.c
```diff
@@ -1583,7 +1583,7 @@ aot_cache_load_module (MonoAssembly *assembly, char **aot_name)
 		if (!home)
 			return NULL;
 		cache_dir = g_strdup_printf ("%s/Library/Caches/mono/aot-cache", home);
-		if (!g_file_test (cache_dir, G_FILE_TEST_EXISTS|G_FILE_TEST_IS_DIR))
+		if (!g_file_test (cache_dir, (GFileTest)(G_FILE_TEST_EXISTS|G_FILE_TEST_IS_DIR)))
```
jaykrell: This cast is redundant with #10274 or #10231.

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build failed

jaykrell:
Good now?

jaykrell:
@luhenry ok now?

jaykrell:
@luhenry ok now?

jaykrell:
Review?

mono/metadata/class-internals.h
```diff
@@ -1067,6 +1067,22 @@ mono_register_jit_icall (gconstpointer func, const char *name, MonoMethodSignatu
 MonoJitICallInfo *
 mono_register_jit_icall_full (gconstpointer func, const char *name, MonoMethodSignature *sig, gboolean no_wrapper, const char *c_symbol);
 
+#ifdef __cplusplus
+template <typename T>
+inline MonoJitICallInfo *
+mono_register_jit_icall (T func, const char *name, MonoMethodSignature *sig, gboolean is_save)
```
lambdageek: One day, maybe, we can make this more typesafe than the C version:

```c++
template <typename T>
inline typename std::enable_if<std::is_function <T>::value, MonoJitICallInfo*>::type
mono_register_jit_icall (T func, const char *name, MonoMethodSignature *sig, gboolean is_save)
{
   return mono_register_jit_icall ((gconstpointer)func, name, sig, is_save);
}
```

mono/mini/aot-runtime.c
```diff
@@ -3114,7 +3114,7 @@ decode_exception_debug_info (MonoAotModule *amodule, MonoDomain *domain,
 			}
 		}
 
-		flags |= JIT_INFO_HAS_UNWIND_INFO;
+		flags = (MonoJitInfoFlags)(flags | JIT_INFO_HAS_UNWIND_INFO);
```
lambdageek: still needed after `G_ENUM_FUNCTIONS (MonoJitInfoFlags)` ?

mono/metadata/class-internals.h
jaykrell: Agreed but not today.
Does it end up being templates all the way down, or is this the correct C++ approach?

mono/mini/aot-runtime.c
jaykrell: Good catch. Unlikely still needed.

jaykrell:
Undo the two lines right?

mono/metadata/class-internals.h
lambdageek: in the libraries that i've seen, sooner or later the templates cease and you end up at some untyped storage layer where everything's a `char*` (and a destructor).

So for something that's essentially a big lookup table of heterogeneous functions this is probably close to the right place to cut down to a `gconstpointer`.

lambdageek:
No need to cleanup the two casts now (look at that all-green :heavy_check_mark: CI!) - the casts aren't _wrong_, after all.

mono/mini/aot-runtime.c
jaykrell: https://github.com/mono/mono/issues/10319 (can undo this line later)

mono/mini/aot-runtime.c
jaykrell: https://github.com/mono/mono/issues/10319 (can undo this line later)

jaykrell:
@lambdageek says elsewhere: "Oops... #10252 loader-internals.h needs to go into some csproj so it gets picked up on windows."

Yes, but why did it pass CI then?

jaykrell:
Working on loader-internals.h right away though. I know where you mean.

jaykrell:
Tarball build will be broken, again.

jaykrell:
er, tarball maybe ok, it is in Makefile.am.in.  phew

jaykrell:
Hm. Maybe not a problem?
```
jaymac1:msvc jay$ grep loader.h *
libmonoruntime-common.targets:    <ClInclude Include="$(MonoSourceLocation)\mono\metadata\loader.h" />
libmonoruntime-common.targets.filters:    <ClInclude Include="$(MonoSourceLocation)\mono\metadata\loader.h">
$ grep  internal  *proj*
```

jaykrell:
Grep for `ClInclude`, some found, but not much.  So I think ok.

jaykrell:
Oh, not proj, but common.targets.

PR #10254, opened 2018-08-23
[System]: `MonoTlsStream` is now `IDisposable` and cleans up on error.

### LINKED ISSUE #8689: Memory leak in HttpWebRequest / HttpWebResponse when using a WebProxy

baulig:
When the TLS Handshake fails in `MonoTlsStream.CreateStream()`, then we must dispose
the `SslStream` to avoid a memory leak.

In addition to this, `MonoTlsStream` now implements `IDisposable` and is disposes
in `WebConnection.CloseSocket()`.

Fixes #8689

### DIFF
--- a/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
@@ -157,20 +157,12 @@ namespace Mono.Net.Security
 #endif
 		}
 
-		void Dispose (bool disposing)
+		public void Dispose ()
 		{
-			if (Interlocked.CompareExchange (ref disposed, 1, 0) != 0)
-				return;
-
 			if (sslStream != null) {
 				sslStream.Dispose ();
 				sslStream = null;
 			}
-		}
-
-		public void Dispose ()
-		{
-			Dispose (true);
 		}
 	}
 }
--- a/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
@@ -73,7 +73,6 @@ namespace Mono.Net.Security
 #endif
 
 		WebExceptionStatus status;
-		int disposed;
 
 		internal WebExceptionStatus ExceptionStatus {
 			get { return status; }
--- a/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
@@ -137,7 +137,7 @@ namespace Mono.Net.Security
 					request.ServicePoint.UpdateClientCertificate (sslStream.InternalLocalCertificate);
 				else {
 					request.ServicePoint.UpdateClientCertificate (null);
-					sslStream?.Dispose ();
+					sslStream.Dispose ();
 					sslStream = null;
 				}
 			}
--- a/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
@@ -50,7 +50,7 @@ using System.Security.Cryptography;
 
 namespace Mono.Net.Security
 {
-	class MonoTlsStream
+	class MonoTlsStream : IDisposable
 	{
 #if SECURITY_DEP
 		readonly MonoTlsProvider provider;
@@ -73,6 +73,7 @@ namespace Mono.Net.Security
 #endif
 
 		WebExceptionStatus status;
+		int disposed;
 
 		internal WebExceptionStatus ExceptionStatus {
 			get { return status; }
@@ -136,6 +137,7 @@ namespace Mono.Net.Security
 					request.ServicePoint.UpdateClientCertificate (sslStream.InternalLocalCertificate);
 				else {
 					request.ServicePoint.UpdateClientCertificate (null);
+					sslStream?.Dispose ();
 					sslStream = null;
 				}
 			}
@@ -153,6 +155,22 @@ namespace Mono.Net.Security
 #else
 			throw new PlatformNotSupportedException (EXCEPTION_MESSAGE);
 #endif
+		}
+
+		void Dispose (bool disposing)
+		{
+			if (Interlocked.CompareExchange (ref disposed, 1, 0) != 0)
+				return;
+
+			if (sslStream != null) {
+				sslStream.Dispose ();
+				sslStream = null;
+			}
+		}
+
+		public void Dispose ()
+		{
+			Dispose (true);
 		}
 	}
 }
--- a/mcs/class/System/Makefile
+++ b/mcs/class/System/Makefile
@@ -41,6 +41,8 @@ USE_XTEST_REMOTE_EXECUTOR = YES
 XTEST_LIB_REFS = System System.Core System.Net Facades/System.Threading.Tasks Facades/System.Runtime.InteropServices.RuntimeInformation System.Net.Http
 LIB_MCS_FLAGS = -d:COREFX -d:CONFIGURATION_2_0 -d:XML_DEP -d:SECURITY_DEP $(REFERENCE_SOURCES_FLAGS) -unsafe $(RESOURCE_FILES:%=-resource:%) -nowarn:436
 
+LIB_MCS_FLAGS += -d:MONO_WEB_DEBUG -d:MONO_TLS_DEBUG
+
 ifndef NO_MONO_SECURITY
 MONO_SECURITY := Mono.Security
 LIB_MCS_FLAGS += -d:MONO_SECURITY_ALIAS
--- a/mcs/class/System/Makefile
+++ b/mcs/class/System/Makefile
@@ -41,8 +41,6 @@ USE_XTEST_REMOTE_EXECUTOR = YES
 XTEST_LIB_REFS = System System.Core System.Net Facades/System.Threading.Tasks Facades/System.Runtime.InteropServices.RuntimeInformation System.Net.Http
 LIB_MCS_FLAGS = -d:COREFX -d:CONFIGURATION_2_0 -d:XML_DEP -d:SECURITY_DEP $(REFERENCE_SOURCES_FLAGS) -unsafe $(RESOURCE_FILES:%=-resource:%) -nowarn:436
 
-LIB_MCS_FLAGS += -d:MONO_WEB_DEBUG -d:MONO_TLS_DEBUG
-
 ifndef NO_MONO_SECURITY
 MONO_SECURITY := Mono.Security
 LIB_MCS_FLAGS += -d:MONO_SECURITY_ALIAS
--- a/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
@@ -77,7 +77,7 @@ namespace Mono.Net.Security
 			Settings = settings;
 			Provider = provider;
 
-			readBuffer = new BufferOffsetSize2 (16834);
+			readBuffer = new BufferOffsetSize2 (16384);
 			writeBuffer = new BufferOffsetSize2 (16384);
 			operation = Operation.None;
 		}
--- a/mcs/class/System/System.Net/WebConnection.cs
+++ b/mcs/class/System/System.Net/WebConnection.cs
@@ -412,6 +412,13 @@ namespace System.Net
 					networkStream = null;
 				}
 
+				if (monoTlsStream != null) {
+					try {
+						monoTlsStream.Dispose ();
+					} catch { }
+					monoTlsStream = null;
+				}
+
 				if (socket != null) {
 					try {
 						socket.Dispose ();
--- a/mcs/class/System/System.Net/WebConnection.cs
+++ b/mcs/class/System/System.Net/WebConnection.cs
@@ -405,6 +405,7 @@ namespace System.Net
 		void CloseSocket ()
 		{
 			lock (this) {
+				Debug ($"WC CLOSE SOCKET: Cnc={ID} NS={networkStream} TLS={monoTlsStream}");
 				if (networkStream != null) {
 					try {
 						networkStream.Dispose ();
--- a/mcs/class/System/System.Net/WebReadStream.cs
+++ b/mcs/class/System/System.Net/WebReadStream.cs
@@ -41,7 +41,7 @@ namespace System.Net
 		}
 
 #if MONO_WEB_DEBUG
-		internal string ME => $"WRS({GetType ().Name}:Op={operation.ID})";
+		internal string ME => $"WRS({GetType ().Name}:Op={Operation.ID})";
 #else
 		internal string ME => null;
 #endif


mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
```diff
@@ -154,5 +156,21 @@ internal async Task<Stream> CreateStream (WebConnectionTunnel tunnel, Cancellati
 			throw new PlatformNotSupportedException (EXCEPTION_MESSAGE);
 #endif
 		}
+
+		void Dispose (bool disposing)
+		{
+			if (Interlocked.CompareExchange (ref disposed, 1, 0) != 0)
```
akoeplinger: do we expect this to be called by multiple threads or why are you doing this vs. a simple `if (disposed) return;` check?

mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
```diff
@@ -136,6 +137,7 @@ internal async Task<Stream> CreateStream (WebConnectionTunnel tunnel, Cancellati
 					request.ServicePoint.UpdateClientCertificate (sslStream.InternalLocalCertificate);
 				else {
 					request.ServicePoint.UpdateClientCertificate (null);
+					sslStream?.Dispose ();
```
akoeplinger: how can sslStream be null here?

nealef:
That will also fix the readBuffer and writeBuffer objects? (And were they supposed to be different sizes? 16834 looks like a typo)

mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
baulig: I'm using this as a pattern in the other related files - does it hurt to have the interlocked here?

mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
```diff
@@ -154,5 +156,21 @@ internal async Task<Stream> CreateStream (WebConnectionTunnel tunnel, Cancellati
...
+
+			if (sslStream != null) {
+				sslStream.Dispose ();
+				sslStream = null;
+			}
+		}
+
+		public void Dispose ()
+		{
+			Dispose (true);
```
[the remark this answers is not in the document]
marek-safar: That's not needed when there is no finalizer but at the same time we don't need the argument either

baulig:
@nealef I believe the 16834 was a type, just fixed that in the code.
@marek-safar @akoeplinger I also cleaned up the Dispose() a bit.

nealef:
Thanks!


@nealef<https://github.com/nealef> I believe the 16834 was a type, just fixed that in the code.
@marek-safar<https://github.com/marek-safar> @akoeplinger<https://github.com/akoeplinger> I also cleaned up the Dispose() a bit.

mcs/class/System/Mono.Net.Security/MonoTlsStream.cs
```diff
@@ -73,6 +73,7 @@ class MonoTlsStream
 #endif
 
 		WebExceptionStatus status;
+		int disposed;
```
marek-safar: This is not used anywhere

akoeplinger:
how far back should we backport this?

nealef:
2018-06 please.


-------- Original message --------

nealef:
BTW, I did my testing against 2018-06 and it worked as expected.

nealef:
Individually this fix and the ChainStatus fix appear to work. Now that both fixes are part of 2018-06 I cloned and checked out this branch and rebuilt. Now when I run the ChainStatus test case things break:

```
System.Net.WebException: A call to SSPI failed, see inner exception. ---> System.Security.Authentication.AuthenticationException: A call to SSPI failed, see inner exception. ---> System.InvalidOperationException: Operation is not valid due to the current state of the object.
  at Mono.Net.Security.MobileAuthenticatedStream.InnerRead (System.Boolean sync, System.Int32 requestedSize, System.Threading.CancellationToken cancellationToken) [0x00055] in <41884af155f44b798e04496ffbf2e844>:0 
  at Mono.Net.Security.AsyncProtocolRequest.InnerRead (System.Threading.CancellationToken cancellationToken) [0x000ac] in <41884af155f44b798e04496ffbf2e844>:0 
  at Mono.Net.Security.AsyncProtocolRequest.ProcessOperation (System.Threading.CancellationToken cancellationToken) [0x00093] in <41884af155f44b798e04496ffbf2e844>:0 
  at Mono.Net.Security.AsyncProtocolRequest.StartOperation (System.Threading.CancellationToken cancellationToken) [0x0008b] in <41884af155f44b798e04496ffbf2e844>:0 
   --- End of inner exception stack trace ---
  at Mono.Net.Security.MobileAuthenticatedStream.StartOperation (Mono.Net.Security.MobileAuthenticatedStream+OperationType type, Mono.Net.Security.AsyncProtocolRequest asyncRequest, System.Threading.CancellationToken cancellationToken) [0x0024b] in <41884af155f44b798e04496ffbf2e844>:0 
  at System.Net.WebResponseStream.InitReadAsync (System.Threading.CancellationToken cancellationToken) [0x000db] in <41884af155f44b798e04496ffbf2e844>:0 
  at System.Net.WebOperation.Run () [0x001d9] in <41884af155f44b798e04496ffbf2e844>:0 
  at System.Net.WebCompletionSource`1[T].WaitForCompletion () [0x00094] in <41884af155f44b798e04496ffbf2e844>:0 
  at System.Net.HttpWebRequest.MyGetResponseAsync (System.Threading.CancellationToken cancellationToken) [0x0027f] in <41884af155f44b798e04496ffbf2e844>:0 
   --- End of inner exception stack trace ---
  at System.Net.HttpWebRequest.RunWithTimeoutWorker[T] (System.Threading.Tasks.Task`1[TResult] workerTask, System.Int32 timeout, System.Action abort, System.Func`1[TResult] aborted, System.Threading.CancellationTokenSource cts) [0x000c5] in <41884af155f44b798e04496ffbf2e844>:0 
  at System.Net.HttpWebRequest.GetResponse () [0x00016] in <41884af155f44b798e04496ffbf2e844>:0 
  at ConsoleApplication3.TestHttpsRequest (System.String url, System.Boolean addCert) [0x0004e] in <d81caa8705284c178c7e0ddab8649693>:0 
  at ConsoleApplication3.TestCase4 () [0x00023] in <d81caa8705284c178c7e0ddab8649693>:0 
```
It's as if the initial certificate error is noticed but not that our callback is giving the go ahead for the session to proceed but the rug has been pulled out from under it.

nealef:
I note that this problem has been reported elsewhere so I am not sure it is these fixes that are responsible. I did some MONO_TLS_DEBUG tracing and saw the critical piece is:
```
MobileAuthenticatedStream(1): AsyncReadRequest(1:75): ProcessOperation - read inner: 10917
MobileAuthenticatedStream(1): InnerRead: False 0 5515 10869 10917
Remaining: 10869 Requsted: 10917 Min: 10869
MobileAuthenticatedStream(1): InnerRead done: 10869 10869 - 10869
MobileAuthenticatedStream(1): InnerRead done: [BufferOffsetSize: 0 16384] - 10869 10869
MobileAuthenticatedStream(1): AsyncReadRequest(1:76): ProcessOperation - read inner done: 10917 - 10869
MobileAuthenticatedStream(1): AsyncReadRequest(1:77): ProcessOperation - read inner: 48
MobileAuthenticatedStream(1): InnerRead: False 0 16384 0 48
Remaining: 0 Requested: 48 Min: 0
```
Is seems we are told we need to read 10917 bytes but our read only returns 10869 so we go back and try and read the remaining 48 bytes but the readBuffer.Remaining value is 0 and the logic is:
```
                         var len = System.Math.Min (readBuffer.Remaining, requestedSize);
                        if (len == 0)
                                throw new InvalidOperationException ();
```
(This is in ```Mono.Net.Security/MobileAuthenticatedStream.cs```). The trace above shows the values and result of the System.Math.Min() function.

I think I found the problem but I don't know why. In ```MobileAuthenticatedStream.cs``` we used to set readBuffer to be of size 16834. In a recent push it was set to 16384 to match writeBuffer thinking it was simply a typo. If I return it to 16834 everything works again!

```
MobileAuthenticatedStream(1): AsyncReadRequest(1:75): ProcessOperation - read inner: 10917
MobileAuthenticatedStream(1): InnerRead: False 0 5515 11319 10917
Remaining: 11319 Requsted: 10917 Min: 10917
MobileAuthenticatedStream(1): InnerRead done: 11319 10917 - 10917
MobileAuthenticatedStream(1): InnerRead done: [BufferOffsetSize: 0 16432] - 10917 10917
MobileAuthenticatedStream(1): AsyncReadRequest(1:76): ProcessOperation - read inner done: 10917 - 10917
MobileAuthenticatedStream(1): AsyncReadRequest(1:77): ProcessOperation run: Continue
```

nealef:
Referencing mono/mono#10371

nealef:
It appears that during the exchange we are requesting 16432 bytes but the buffer only holds 16384 so there is still 48 remaining. However, before we go for that remaining 48 we ask for 10917 but only get 10869. I put a Console.Error.WriteLine() in the get_Remaining property and the other message comes just before the System.Math.Min() in InnerRead.
```
Remaining: 16384 Requested: 16432 Min: 16384                          
Length: 16384 Offset: 0 Size: 0                                      
Length: 16384 Offset: 0 Size: 5515                                   
Length: 16384 Offset: 0 Size: 5515                                   
Length: 16384 Offset: 0 Size: 5515                                   
Length: 16384 Offset: 0 Size: 5515                                   
Remaining: 10869 Requested: 10917 Min: 10869                          
Length: 16384 Offset: 0 Size: 5515                                   
Length: 16384 Offset: 0 Size: 16384                                  
Length: 16384 Offset: 0 Size: 16384                                  
Length: 16384 Offset: 0 Size: 16384                                  
Length: 16384 Offset: 0 Size: 16384                                  
Remaining: 0 Requested: 48 Min: 0
```
When the buffer is 16834 we are able to satisfy the first of those requests and proceed on:
```
Remaining: 16834 Requested: 16432 Min: 16432
Length: 16834 Offset: 0 Size: 0
Length: 16834 Offset: 0 Size: 5515
Length: 16834 Offset: 0 Size: 5515
Length: 16834 Offset: 0 Size: 5515
Length: 16834 Offset: 0 Size: 5515
Remaining: 11319 Requested: 10917 Min: 10917
Length: 16834 Offset: 0 Size: 5515
Length: 16834 Offset: 0 Size: 13795
Length: 16834 Offset: 0 Size: 13795
Length: 16834 Offset: 0 Size: 13795
Length: 16834 Offset: 0 Size: 13795
```

nealef:
If ret < requested then should multiple reads be done until that request is fulfilled?

PR #10266, opened 2018-08-23
[wasm] Another batch of AOT work. Now we crash when running AOT'd code!

kumpera:
IMPORTANT NOTE: This PR includes changes to the offsets setup, so the iOS SDK lane is required to be working.


mono/metadata/abi-details.h
```diff
@@ -23,29 +23,33 @@
  * is given by the --with-cross-offsets= configure argument.
  */
```
vargaz: Why is this needed ?

kumpera: Because I decided I don't want to add an offsets file for wasm. We discussed a thousand times about getting rid of those annoying thing and start to support a retargetable runtime.

This PR includes the beginning of that. It only does for alignment info and it only has info for wasm. I hope we eventually include all our targets in abi.c (or whatever we decide to use instead) and the offset files stop being a thing.

mono/metadata/abi.c
```diff
@@ -0,0 +1,105 @@
...
+	static gboolean inited;
+	/*
+	 * OMG This is RACY! Like SUPER racy!
+	 * That's true, *but* given the values are constant and we do safe publication, it's harmless.
+	 */
+	if (!inited) {
+		init_abi_detail ();
+		mono_memory_barrier ();
+		inited = TRUE;
+	}
```
jaykrell: Barriers come in pairs. This is missing the barrier on the reader.
The reader can pre-read the variables that are attempted to be guarded by init the barrier. The barrier keeps the writes ordered (enough) but not the reads.

Suggest pthread_once or mono-wrapped analog.

or put in `else { barrier(); }`

mono/mini/aot-runtime.c
```diff
@@ -2760,10 +2763,14 @@ compute_llvm_code_range (MonoAotModule *amodule, guint8 **code_start, guint8 **c
 static gboolean
 is_llvm_code (MonoAotModule *amodule, guint8 *code)
 {
+#if HOST_WASM
```
jaykrell: In the interest of the "retargetable runtime", can these be `if` instead of `#if`?

mono/metadata/abi-details.h
vargaz: Alignment determines offsets/sizes, so hardcoding just the offsets will not work.

kumpera: I'm fully aware of it.
This is what I needed so far, the rest will come later as part of the wasm AOT work.

mono/mini/aot-runtime.c
kumpera: Those are workarounds for now to unblock further work. I don't know what we'll end up with.
I rather not have any of those at all.

mono/metadata/abi.c
kumpera: Fair enough. So I stopped being lazy and went with static initialization which does address this issue.

kumpera:

kumpera:

kumpera:
@monojenkins build failed

kumpera:
@monojenkins build failed

kumpera:

PR #10269, opened 2018-08-23
Strengthen several field members and TLS and function returns to not be void*.

jaykrell:
The components are already coupled by virtue of where the data already is.

Pointers to opaque structs do not represent an increased dependency/coupling.
The types are as opaque as a void*.

Nothing about the structs is revealed except their name, and even this, if necessary, unlikely, could revert to casts to some other type.

The result is slightly cleaner, typesafe, easier to understand, harder to misuse the pointers, viewable  in debugger w/o casts, etc.


jaykrell:
@monojenkins build Linux x64

mono/metadata/metadata.h
```diff
@@ -23,8 +23,6 @@ MONO_BEGIN_DECLS
 
 #define MONO_CLASS_IS_IMPORT(c) ((mono_class_get_flags (c) & TYPE_ATTRIBUTE_IMPORT))
 
-typedef struct _MonoDomain MonoDomain;
```
jaykrell: Any header with a removed typedef should gain an include.

mono/metadata/object-internals.h
```diff
@@ -489,7 +490,7 @@ struct _MonoInternalThread {
 
 struct _MonoThread {
 	MonoObject obj;
-	struct _MonoInternalThread *internal_thread;
```
jaykrell: Any header switching from its own forward declare to rely on typedef should gain an include.

PR #10274, opened 2018-08-24
[cxx] int vs. enum -- bit overload version

jaykrell:
There are at least three or so approaches to this.

1. Sprinkle around casts.
2. Overload operators.
2a. Bit-only (| & ^ |= &= ^=)   This is this PR and seemingly the consensus.
2b.  Also math (+ - ++ -- == !=)
3. typedef to int instead (https://github.com/mono/mono/pull/10231)

They all have tradeoffs.

1. Casts: The most line damage. Not huge, but the most.

2. Overloads: New idiom to some but an old idiom. Supported in this way by windows.h even.
  This is the most "C++y solution".

3. int: Highest level of source compat and semantic compat. Least line damage (just
change the declarations). Least debuggable and least typesafe.

In terms of semantic compat, even in C the guarantees aren't particularly strong.
Some enums will be int, some unsigned, depending on the values and the compiler.
Perhaps even other types, but I doubt it.

I observe C++ compilers don't tend toward unsigned the same way, so switching
to C++ will on some platforms lose unsignedness. Given we didn't have it portably
anyway, should be ok.

A typesafe enum has the properties that enum A cannot be silently assigned to enum B (ok,
there might be a warning), or from an integer.

A typesafe enum, bitfield or otherwise, can/will be shown symbolically by a debugger.
 (lldb does not handle bitfields, only strict enums)

A C enum might be debuggable, w/o typesafety.

A C enum can be freely mixed with other enums and integers.

A C++ enum is more restricted.

Using typedef int, is a maybe surprising option, but does provide the closest meaning to C enum.

Per guidance, this PR also changes one enum to #defines + int.
It might be reasonable to do similar for others.
It might be reasonable to provide more operators (see for example https://github.com/mono/mono/pull/10027/commits/648ed594c16f77954e16fed9ecfeddfdd6e49bcb#diff-b8c6c2ba344450884373fb00f33685adR245).

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -109,6 +109,13 @@ public:
 
 #ifdef __cplusplus
 
+// G++4.4 breaks opeq below without this.
+#if defined  (__GNUC__) || defined  (__clang__)
+#define G_MAY_ALIAS  __attribute__((__may_alias__))
+#else
+#define G_MAY_ALIAS /* nothing */
+#endif
+
 // Provide for bit operations on enums, but not all integer operations.
 // This alleviates a fair number of casts in porting C to C++.
 
@@ -121,22 +128,24 @@ template <> struct g_size_to_int<2> { typedef int16_t type; };
 template <> struct g_size_to_int<4> { typedef int32_t type; };
 template <> struct g_size_to_int<8> { typedef int64_t type; };
 
-template <typename T>
-using g_size_to_int_t = typename g_size_to_int <sizeof (T)>::type;
+// g++4.4 does not accept:
+//template <typename T>
+//using g_size_to_int_t = typename g_size_to_int <sizeof (T)>::type;
+#define g_size_to_int_t(x) g_size_to_int <sizeof (x)>::type
 
 #define G_ENUM_BINOP(Enum, op, opeq) 		\
 inline Enum					\
 operator op (Enum a, Enum b)			\
 {						\
-	typedef g_size_to_int_t <Enum> type; 	\
+	typedef g_size_to_int_t (Enum) type; 	\
 	return static_cast<Enum>(static_cast<type>(a) op static_cast<type>(b)); \
 }						\
 						\
 inline Enum&					\
 operator opeq (Enum& a, Enum b)			\
 {						\
-	typedef g_size_to_int_t <Enum> type; 	\
-	return (Enum&)((type&)(a) opeq static_cast<type>(b)); \
+	typedef g_size_to_int_t (Enum) G_MAY_ALIAS type; \
+	return (Enum&)((type&)a opeq b); 	\
 }						\
 
 #define G_ENUM_FUNCTIONS(Enum)			\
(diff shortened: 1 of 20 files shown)


mono/eglib/glib.h
```diff
@@ -107,6 +107,56 @@ struct g_cast
...
+template <> struct mono_size_to_int<8> { typedef int64_t T; };
+
+#define MONO_SIZE_TO_INT(x) mono_size_to_int <sizeof (x)>::T
+
+#define G_ENUM_BINOP(Enum, op, opeq) 		\
+inline Enum					\
+operator op (Enum a, Enum b)			\
+{						\
+	typedef MONO_SIZE_TO_INT (a) T;		\
+	return (Enum)((T)a op (T)b);		\
```
lambdageek: These casts could be `static_cast<>`.

mono/eglib/glib.h
```diff
@@ -107,6 +107,56 @@ struct g_cast
 
 #endif
 
+#ifdef __cplusplus
+
+// Provide for bit operations on enums, but not all integer operations.
+// This alleviates a fair number of casts in porting C to C++.
+
+// Forward declare template with no generic implementation.
+template <size_t> struct mono_size_to_int;
```
lambdageek: bikeshed: since this is in glib, maybe a `g` prefix or namespace?

mono/eglib/glib.h
```diff
@@ -107,6 +107,56 @@ struct g_cast
...
+
+// Template specializations.
+template <> struct mono_size_to_int<1> { typedef int8_t T; };
```
lambdageek: C++ type traits libraries usually define the result type with name `type`, not `T`.

mono/eglib/glib.h
jaykrell: Oops, yes.

mono/eglib/glib.h
jaykrell: I was shielding developers from the verbosity, and think it is reasonable, but will fix.

mono/eglib/glib.h
jaykrell: I was erring toward less C++-y,  but yes, will fix.

jaykrell:
I think `make_int` maybe too terse -- make an `int` from what? Thus `size_to_int`. Otherwise agreed, will attempt.

jaykrell:
I also don't like the `.hpp` extension. It is commonly used, and commonly not used.
Files can have a mix of C and C++. I realize this is solidly C++, except, for the `#else`.
I don't foresee us renaming any files, except maybe main.c, for build reasons.
I do use `.cpp` in new code, but still don't like `.h`.

jaykrell:
`gtype_traits.h`? Well, I also don't like overengineering file organization. I like fewer files. I had  everything in `util.h` and `util.c`. Which is common, people do it, and people dislike it also.

jaykrell:
Is this still ok, with the post-review changes?

jaykrell:
@monojenkins build Linux i386

jaykrell:
Looks like need to update .proj file.

jaykrell:
@monojenkins build ARMv5

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X x64

PR #10283, opened 2018-08-25
Configure eglib/test/Makefile.

jaykrell:


jaykrell:
https://github.com/mono/mono/pull/10274 instead.

PR #10284, opened 2018-08-25
[cxx] Port libtest to C++.  @monojenkins skip

jaykrell:
@monojenkins skip
This also reduces need for extern "C" in runtime.

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -2561,7 +2561,7 @@ test-unhandled-exception: unhandled-exception-test-runner.2.exe
 safehandle.2.exe winx64structs.exe thunks.exe pinvoke3.exe pinvoke2.exe pinvoke-2.2.exe pinvoke17.exe pinvoke13.exe \
 	pinvoke11.exe pinvoke_ppcs.exe pinvoke_ppci.exe pinvoke_ppcf.exe pinvoke_ppcd.exe pinvoke_ppcc.exe pinvoke.exe \
 	marshalbool.exe marshal9.exe marshal5.exe marshal.exe handleref.exe cominterop.exe bug-Xamarin-5278.exe \
-	bug-58782-plain-throw.exe bug-58782-capture-and-throw.exe install_eh_callback.exe: libtest.la
+	bug-58782-plain-throw.exe bug-58782-capture-and-throw.exe install_eh_callback.exe: libtest.la libtest-c.la
 
 event-get.2.exe$(PLATFORM_AOT_SUFFIX): event-il.exe$(PLATFORM_AOT_SUFFIX)
 event-get.2.exe: event-il.exe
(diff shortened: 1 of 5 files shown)


mono/tests/Makefile.am
```diff
@@ -2942,22 +2942,29 @@ EXTRA_DIST += test_lldb.py test-lldb.cs
...
 libtest_la_LDFLAGS=-no-undefined -rpath `pwd` -Wl,--kill-at
 else
+libtest_la_CFLAGS=@CXX_ADD_CFLAGS@
 libtest_la_LDFLAGS = -rpath `pwd`
 endif
 libtest_la_SOURCES = libtest.c
-libtest_la_LIBADD = $(top_builddir)/mono/eglib/libeglib.la
+libtest_la_LIBADD = $(top_builddir)/mono/eglib/libeglib.la libtest-c.la
+
+libtest_c_la_CFLAGS = @CXX_REMOVE_CFLAGS@
```
lambdageek: I'm confused by the CFLAGS here: I thought `CXX_REMOVE_CFLAGS` was just a set of flags that might be present in a `CFLAGS` that must be dropped if we're adding the `CXX_ADD_CFLAGS`.  What does it mean to pass just the remove flags as the cflags for a library?

lambdageek: oh, I guess you remove the `CXX_REMOVE_FLAGS` on line `2950`.  and so here you add them back? But what if one of the REMOVE_FLAGS wasn't present on the original CFLAGS?

jaykrell: It is convoluted I agree. It evolved, and maybe can evolve again. I'm not sure.

CFLAGS contains flags that invalid for C++.
So we have to "remove" them.
And then "add" some flags to become C++.
But if some libraries within a directory remain  as C, which I did not exect, then we have to add back what we removed.

The expected usage model was just, as a directory is ported, add:
```CFLAGS := $(filter-out @CXX_REMOVE_CFLAGS@, @CFLAGS@) @CXX_ADD_CFLAGS@```

but directories that are a mix of C and C++ aren't so clean.

Comment (jaykrell) [reply mono/tests/Makefile.am]: Good question. Everything in REMOVE should be present in CFLAGS, as configure.ac computes them.

**Comment by jaykrell** [comment]
Ok?

File: mono/tests/Makefile.am [hunk 2]
```diff
@@ -2942,22 +2942,29 @@ EXTRA_DIST += test_lldb.py test-lldb.cs
 test-lldb: test-lldb.exe
 	python test_lldb.py $(JITTEST_PROG)
 
-noinst_LTLIBRARIES = libtest.la
+noinst_LTLIBRARIES = libtest.la libtest-c.la
```
jaykrell: This could instead be a regular non-libtool library, and below LIBADD the .o files.

mono/tests/libtest-c.c
```diff
@@ -0,0 +1,92 @@
...
+#define LIBTEST_API MONO_EXTERN_C
+#endif
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wc++-compat"
+#endif
+
+/*
+* Standard C and C++ doesn't allow empty structs, empty structs will always have a size of 1 byte.
```
jaykrell: This comment is copy/pasted, and inaccurate.

jaykrell:
We have decided that more extern "C" is ok, to cut off the graph of C++ conversion, so this PR is now probably optional. I will look what extern "C" is required, shouldn't really matter what it is.

jaykrell:
I also like this PR kinda highlighted wierd stuff and its wierd tests, made it even slightly wierder.

luhenry:
Not porting tests for now.

PR #10286, opened 2018-08-25
Bump build requirement to 10.9 mavericks

### LINKED ISSUE #9581: Update macOS minimal version

kg:
Fixes #9581

How do I check that I have the right llvm triple for 10.9? As far as I can tell they aren't documented anywhere and the recommended approach is 'run llvm on a machine that matches the desired triple', which is not really an option I have. There were at least two darwin releases for 10.9.


marek-safar:
@monojenkins build failed

alexischr:
@monojenkins build pkg

baulig:
Once this has landed, we need to back port to `2018-08` and then also do some changes in `xamarin-macios`.

kg:
@marek-safar @alexischr does that mean I need to adjust this so it only changes the minimum?

alexischr:
No I think it’s fine, just clarifying against the title that the intent was to raise the target version

akoeplinger:
@monojenkins backport 2018-08

PR #10288, opened 2018-08-25
Remove gratituious ifdef -- just have one portable form -- alloca and no variable length arrays.

jaykrell:


jaykrell:

jaykrell:

PR #10293, opened 2018-08-26
[cxx] Fix typesafe-linkage problem due to varying MonoGCDescriptor.

jaykrell:
Allow mono_gc_alloc_fixed subset outside of metadata (mini/tasklets.c), without breaking due to C++ typesafe linkage.
    
This enables avoiding externC workaround here:
https://github.com/mono/mono/pull/10160/files#diff-bd3a3d5c2bfd6af46373b3844de9c1d0R163


jaykrell:

jaykrell:
We compile/link it and this is fixing actual build problems. Whether or not it runs, I don't know.

PR #10294, opened 2018-08-26
[cxx] Rework mono-threads.h THREAD_INFO_TYPE parameterization to not violate C++ typesafe linkage.

jaykrell:
i.e. declaration and implementation must match.
Thus eliminating externC workaround here:
 https://github.com/mono/mono/pull/10160/files#diff-6fa4afbec40d6b93ebd74cf63b06fe93R607
 and many times in same header.


jaykrell:

jaykrell:
Review?

mono/utils/mono-threads.h
```diff
@@ -751,4 +797,84 @@ typedef void (*background_job_cb)(void);
...
+// That will require changing all the explicit accesses to the "named base".
+// SgenThreadInfo * info;
+// info->base.foo => info-foo, and can be only one foo between base and derived.
+// This at least solves the input casts, which is most of the problem.
+
+// Output casts.
+#define mono_thread_info_attach(info)		((THREAD_INFO_TYPE*)mono_thread_info_attach ())
+#define mono_thread_info_current()		((THREAD_INFO_TYPE*)mono_thread_info_current ())
+#define mono_thread_info_current_unchecked()	((THREAD_INFO_TYPE*)mono_thread_info_current_unchecked ())
+#define mono_thread_info_lookup(id) 		((THREAD_INFO_TYPE*)mono_thread_info_lookup (id))
```
jaykrell: These macros can all be functions, if that is more palatable.
It is obvious for the "input" form, just overload.

The "output" form goes like:
```
static inline THREAD_INFO_TYPE*
mono_thread_info_lookup (id, THREAD_INFO_TYPE *  = 0 )
{
	return (THREAD_INFO_TYPE*)mono_thread_info_lookup (id, (MonoThreadInfo*)0);
}
```

Notice the defaulted parameter.
The main functions would get this too though, otherwise is ambiguous.

jaykrell:
Offline proposal is try removing `THREAD_INFO_TYPE` completely and see how many casts result.

PR #10297, opened 2018-08-26
[cxx] Slightly rework gc-internals.h THREAD_INFO_TYPE parameterization

jaykrell:
to not break w/ C++ typesafe linkage. All three GC implementations provide both overloads of three functions. This avoids the need for externC workaround here:
 https://github.com/mono/mono/pull/10160/files#diff-bd3a3d5c2bfd6af46373b3844de9c1d0R442


jaykrell:
Review?

jaykrell:
Offline proposal is try removing `THREAD_INFO_TYPE` completely and see how many casts result.

PR #9981, opened 2018-08-08
[sdb] Disable an assert to avoid breaking debugging.

vargaz:
Fixes https://github.com/mono/mono/issues/8848.
Fixes https://github.com/mono/mono/issues/9262.
Fixes https://github.com/mono/mono/issues/8627.

PR #10315, opened 2018-08-27
[2018-02] [sdb] Disable an assert to avoid breaking debugging.

luhenry:
Backport of https://github.com/mono/mono/pull/9981.


luhenry:
See https://devdiv.visualstudio.com/DevDiv/Xamarin%20VS%20for%20Mac/_workitems/edit/660993/?triage=true

PR #10316, opened 2018-08-27
[mini] fix memory dumping on devices

lewurm:
also, guard for `IP == NULL`: I haven't seen it in the wild, but when messing with EH I have seen it happen and then we have a segfault in the signal handler, which is... bad 😄 

we should backport this to `2018-06` and `2018-08`.


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2018-06

lewurm:
@monojenkins backport 2018-08

PR #10320, opened 2018-08-28
Show dot-folders in folder browser

HebaruSan:
## Problem

Mono's `FolderBrowserDialog` always hides folders with names that start with ".".

Some important folders have such names, for example `.local/share/Steam/SteamApps/common/`, or `.local/share/MyApp/`.

This behavior could be seen as a sensible default if there was a way for developers to control it, since these folders are traditionally treated as "hidden." However, there is no way to control this behavior; the check is hard-coded, and so some of the folders that exist on disk and may be important to display simply never appear. If an application needs to allow a user to choose a folder with a name that starts with ".", or a folder *inside* a folder with a name that starts with ".", alternatives to the standard `FolderBrowserDialog` class must be pursued, such as a text box for manually editing a path.

Showing hidden folders may be awkward in some use cases, but hiding them is much worse when they're needed.

## Changes

This filter is now removed. Folders that start with "." now appear in the folder browser.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/FolderBrowserDialog.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/FolderBrowserDialog.cs
@@ -615,9 +615,6 @@ namespace System.Windows.Forms {
 				ArrayList folders = vfs.GetFoldersOnly ();
 				
 				foreach (FSEntry fsentry in folders) {
-					if (fsentry.Name.StartsWith ("."))
-						continue;
-					
 					FBTreeNode child = new FBTreeNode (fsentry.Name);
 					child.Tag = fsentry.FullName;
 					child.RealPath = fsentry.RealName == null ? fsentry.FullName : fsentry.RealName;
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

/cc @filipnavara

filipnavara:
I am not sure that this is such a great idea. I do agree that there should be a way to access the hidden folders.

The managed implementation mimics the Windows file open dialog. In Windows you can type the path into the file name field to open it and that works whether it is hidden or not. I didn't check if the Mono implementation allows that, but my hunch is that it could be broken.

Similar mechanism work on other systems too. On macOS the default behavior is also to hide hidden folders, but one can press CMD+SHIFT+G and enter a path directly in the same way as Windows allows it.

filipnavara:
I just realized that it affects only `FolderBrowserDialog` and not `OpenFileDialog`/`SaveFileDialog`. That makes it trickier. I am not aware of any way to enter a path directly there. On the other hand it makes it inconsistent with the other dialogs.

HebaruSan:
Hmm, `FolderBrowserDialog` and the `FileDialog`s are already pretty dramatically inconsistent; simple tree vs list box, text combobox, "places" sidebar, various other comboboxes, even vertical vs horizontal layout of buttons, etc. And `FileDialog` doesn't have a functional problem currently; as noted, you can already access hidden folders with it by typing their names.

FileDialog

While I agree that an overhaul of `FolderBrowserDialog` would be great to have, I'm not comfortable attempting that, and I still think there's value in correcting a problem with the current implementation.

steveisok:
@filipnavara Should this PR be closed?  Your suggestion seems reasonable, but I don't want to hold a PR open if the implementation is going to be different.

steveisok:
OK, I'm closing this one.  Please consult our [old pull request policy](https://github.com/mono/mono/wiki/Pull-Request-Guide#old-pull-request-policy) for more info.

filipnavara:
Sorry, it fell off my radar. For the record I would be fine if we introduced an environment variable to force showing the dot-files until a better solution is available.

PR #10323, opened 2018-08-28
Disable tailcall of gsharedvt.

jaykrell:
It doesn't really work fully anyway -- some stack is conserved, but the wrappers still consume stack, largely defeating the purpose.

Better would be to avoid gsharedvt by instantiating the generic methods.


marek-safar:
@monojenkins backport 2018-08

jaykrell:

jaykrell:

jaykrell:
The failing test implies it did work usefully, so this should be looked at further in master.
But disabling for now is ok.

Issue #10054, opened 2018-08-12
DeflateStream behaviour is different on Mono


JimBobSquarePants:
This issue has been compiled from information gathered by @Splamy in an issue raised on the ImageSharp issue tracker affecting our ability to decode png files.

https://github.com/SixLabors/ImageSharp/issues/598

`DeflateStream` exhibits slightly different behavior in net/dotnet and Mono. The following code will execute differently:
 
## Steps to Reproduce

```csharp
using System;
using System.IO;
using System.IO.Compression;
using System.Text;

namespace Bug1
{
    internal class Program
    {
        public static void Main(string[] args)
        {
            // "HelloWorld" as UTF8/DeflateStream(...,CompressionMode.Compress)
            var buffer = new byte[] { 243, 72, 205, 201, 201, 15, 207, 47, 202, 73, 1, 0 };
            var mem = new MemoryStream(buffer);
            var chu = new ChunkedReader(mem);
            var def = new DeflateStream(chu, CompressionMode.Decompress);
            
            var buffer2 = new byte[4096];
            int read2 = 0;

            chu.limit = 3;
            read2 += def.Read(buffer2, read2, buffer2.Length - read2);
            chu.limit = 100;
            read2 += def.Read(buffer2, read2, buffer2.Length - read2);

            var res = Encoding.UTF8.GetString(buffer2, 0, read2);
            Console.WriteLine(res);
        }
    }   
    
    public class ChunkedReader : Stream
    {
        private int sum = 0;
        public int limit = 0;
        private string name;
        private Stream baseStream;
        
        public ChunkedReader(Stream baseStream, string name = null)
        {
            this.baseStream = baseStream;
            this.name = name;
        }

        public override void Flush()
        {
            this.baseStream.Flush();
        }

        public override int Read(byte[] buffer, int offset, int count)
        {
            int read;
            read = this.baseStream.Read(buffer, offset, Math.Min(count, this.limit));
            sum += read;
            this.limit -= read;
            Console.WriteLine("{3}> Sum: {0}, Read {1}, Req: {2}", this.sum, read, count, name ?? "XXX");
            return read;
        }

        public override long Seek(long offset, SeekOrigin origin)
        {
            return this.baseStream.Seek(offset, origin);
        }

        public override void SetLength(long value)
        {
            this.baseStream.SetLength(value);
        }

        public override void Write(byte[] buffer, int offset, int count)
        {
            this.baseStream.Write(buffer, offset, count);
        }

[...]

PR #10329, opened 2018-08-28
Fix DeflateStream handling of zero-length reads from underlying stream.

filipnavara:
Fixes issue #10054.

### DIFF
--- a/support/zlib-helper.c
+++ b/support/zlib-helper.c
@@ -208,7 +208,9 @@ ReadZStream (ZStream *stream, guchar *buffer, gint length)
 			stream->eof = TRUE;
 			break;
 		} else if (status == Z_BUF_ERROR && stream->total_in == zs->total_in) {
-			stream->eof = TRUE;
+			if (zs->avail_in != 0) {
+				stream->eof = TRUE;
+			}
 			break;
 		} else if (status != Z_OK) {
 			return status;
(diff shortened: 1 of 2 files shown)


filipnavara:
@monojenkins build failed

filipnavara:

support/zlib-helper.c
```diff
@@ -208,7 +208,9 @@ ReadZStream (ZStream *stream, guchar *buffer, gint length)
 			stream->eof = TRUE;
 			break;
 		} else if (status == Z_BUF_ERROR && stream->total_in == zs->total_in) {
-			stream->eof = TRUE;
+			if (zs->avail_in != 0) {
```
marek-safar: I am not sure this is this right place. Could we skip everything when the count is 0 at https://github.com/mono/mono/blob/198150688b7087450ba67df2e2437dd747bab5aa/support/zlib-helper.c#L188 ?

filipnavara: We cannot and there are unit tests for it. There could be partially decoded bytes from previous call that didn't fit in the output buffer. Thus even reading 0 bytes from input stream can produce some output.

filipnavara: To further elaborate: Lot of the code, including `stream->eof` and https://github.com/mono/mono/blob/198150688b7087450ba67df2e2437dd747bab5aa/support/zlib-helper.c#L203-204 are just unnecessary shortcuts and micro-optimizations. This was the least invasive place I could find to solve the issue.

marek-safar:
@monojenkins backport 2018-08

PR #10337, opened 2018-08-28
[exceptions] ftnptr_eh_callback test that stresses n2m transitions

lewurm:
on Xamarin.iOS it happens that the ftnptr_eh_callback calls into managed
code, that is yet another native-to-managed transition.

This fails coop right now, see https://github.com/mono/mono/pull/10224#issuecomment-416646880

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -7841,9 +7841,10 @@ mono_get_eh_callbacks (void)
 void
 mono_raise_exception (MonoException *ex) 
 {
-	MONO_ENTER_GC_UNSAFE;
+	/* raise_exception doesn't return, so the transition to GC Unsafe is unbalanced */
+	MONO_STACKDATA (stackdata);
+	mono_threads_enter_gc_unsafe_region_unbalanced_with_info (mono_thread_info_current (), &stackdata);
 	mono_raise_exception_deprecated (ex);
-	MONO_EXIT_GC_UNSAFE;
 }
 
 /*
--- a/mono/tests/install_eh_callback.cs
+++ b/mono/tests/install_eh_callback.cs
@@ -10,8 +10,15 @@ public class Tests {
 
 	[DllImport ("libtest")]
 	public static extern void mono_test_setjmp_and_call (VoidVoidDelegate del, out IntPtr handle);
+
+	[DllImport ("libtest")]
+	public static extern void mono_test_setup_ftnptr_eh_callback (VoidVoidDelegate del, VoidHandleHandleOutDelegate inside_eh_callback);
+
+	[DllImport ("libtest")]
+	public static extern void mono_test_cleanup_ftptr_eh_callback ();
 	
 	public delegate void VoidVoidDelegate ();
+	public delegate void VoidHandleHandleOutDelegate (uint handle, out int exception_handle);
 
 	public class SpecialExn : Exception {
 	}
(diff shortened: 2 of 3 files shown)


lambdageek:
One good theory that @lewurm had is that `mono_raise_exception` shouldn't be using `MONO_ENTER_GC_UNSAFE`/`MONO_EXIT_GC_UNSAFE` but rather it should be using the `_unbalanced` thread state transition function.

And indeed:
1. It's not expected to return, so the balancing `EXIT` will never be reached
2. We don't have any uses of `mono_raise_exception` (or any tests of it) since it's marked `MONO_RT_EXTERNAL_ONLY`.

lambdageek:
Pushed a fix to `mono_raise_exception`.

This is ready for review now.

mono/metadata/object.c
```diff
@@ -7841,9 +7841,10 @@ mono_get_eh_callbacks (void)
 void
 mono_raise_exception (MonoException *ex) 
 {
-	MONO_ENTER_GC_UNSAFE;
+	/* raise_exception doesn't return, so the transition to GC Unsafe is unbalanced */
+	MONO_STACKDATA (stackdata);
+	mono_threads_enter_gc_unsafe_region_unbalanced_with_info (mono_thread_info_current (), &stackdata);
```
lewurm: I tried something similar, using `mono_threads_enter_gc_unsafe_region_unbalanced ()` which basically expands to `mono_threads_enter_gc_unsafe_region_unbalanced_with_info (mono_thread_info_current_unchecked (), &stackdata)`.

so it's `mono_thread_info_current ()` vs. `mono_thread_info_current_unchecked ()`. Any reason why we don't use `mono_thread_info_current ()` in `mono_threads_enter_gc_unsafe_region_unbalanced ()`?

lewurm:
one failure on aarch64 coop:
```
threadpool-exceptions2.exe
=============== threadpool-exceptions2.exe.stdout ===============
* Assertion at os-event-unix.c:82, condition `event' not met

Entering thread summarizer from 7f904d91d0
Entering thread summarizer from 7f906d91d0
Self-reporting for thread 7f906d91d0. Registered summarizing thread right now is 7f906d91d0
Assertion failure in thread 0x7f906d91d0 due to: object.c:1875: Expected GC Unsafe mode but was in STATE_BLOCKING state
Last 2 state transitions: (most recent first)
[DONE_BLOCKING] STATE_BLOCKING -> RUNNING (0) 0 at:
[DO_BLOCKING] RUNNING -> STATE_BLOCKING (0) 0 at:
	at /home/builder/jenkins/workspace/test-mono-pull-request-coop-arm64/mono/mini/mono(+0x2b34b0) [0x55615334b0]


=============== threadpool-exceptions2.exe.stderr ===============
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Environment.Exit (int) <0x00007>
  at Test.OnUnhandledException (object,System.UnhandledExceptionEventArgs) [0x00053] in <e414e82e22a94d9bad5891e9b2c25251>:0
  at (wrapper runtime-invoke) <Module>.runtime_invoke_void__this___object_object (object,intptr,intptr,intptr) [0x0001e] in <c1da1f60e92846d58342963eb0f3d26e>:0
/proc/self/maps:
```

not sure if related, let me retry.

lewurm:

mono/metadata/object.c
lambdageek: ISTR `mono_threads_enter_gc_unsafe_region_unbalanced` (with the `_unchecked` thread info) can be called by products on detached threads in some circumstance.

But the real reason is because there's a dozen of these functions and I always forget about them.

lewurm:
@monojenkins rebase

PR #10341, opened 2018-08-29
[cxx] Port support to C++. I wasn't going to go here, but it uses glib

jaykrell:
and therefore leads to sprinkling or blanketing externC there.

This lets us cut down the externC in glib further.
Alternatively, only about four files in support use glib, those could be made
into a separate library.

Ideally some of the `void*` would be `FILE*` or `DIR*` or `void (*MonoSignalHandler)(int)`
but the .h file says not to edit, so `void*` remains.

Care/caution is taken also to mark pretty much everything as extern "C".
Much already was. nm used to verify.

Most of zlib here is not used on most platforms.
Consider compiling it anyway.
Makefile.am was temporarily hacked up for that purpose.

### DIFF
--- a/support/Makefile.am
+++ b/support/Makefile.am
@@ -16,6 +16,8 @@ AM_CPPFLAGS =					\
 
 glib_libs = $(top_builddir)/mono/eglib/libeglib.la
 
+CFLAGS := $(filter-out @CXX_REMOVE_CFLAGS@, @CFLAGS@) @CXX_ADD_CFLAGS@
+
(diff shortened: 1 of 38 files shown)


jaykrell:
Maybe split up compilation of the lib, so that most of the files remain as C, and then  smush the objs back into the same libs. 

Or maybe just some externC in glib.h. Esp. for now.

jaykrell:
There  are at least two serous alternatives to this, and this one is not likely what we chose.

jaykrell:
Not doing this. Too many preferred alternatives.

Issue #10319, opened 2018-08-27
C++ worsenings/improvements


jaykrell:
# The conversion has some classes of compromises.

1.  The code errs toward one codebase that is valid C and C++, erring away from #ifdef __cplusplus.
If/once the tree is commited to C++, C++ idioms can be used, instead of a common C/C++ subset.

2. Time and churn rate. Maybe given more/faster churn/time, other alteratives would be used.

3. Time and decision making time. What I mean, is it *possible*, people were busy with other priorities and let things through, in the interest of neither slowing down  C++ conversion nor other proirities, but with some breaking time, maybe other decisions would have been made.
So, this is anything people disagreed with, or anything that seems a little off.

# function-macro-same-name

There are multiple uses of this.
Let's break it down.

# function-macro-same-name-voidp-to-non-voidp

You can find these in the tree by looking for `g_cast`.

The decision taken was, `#define foo(x) g_cast(foo(x))`.
The alternatives to this include:
1. Rename the function to `foo_internal`.
I consider this inferior because it breaks breakpoints.
Opinions might vary.

2. Given time and acceptable churn rate, sprinkle in all the casts.

3. Templatized the allocators?

# function-macro-same-name-thread-info-type

This has not been commited yet so is speculative.
THREAD_INFO_TYPE leads to prototypes mismatching implementation.
There is a PR (not yet commited) which does:
define foo foo
to workaround.

There are two sets of functions here -- with inputs and outputs.
There are far more input functions.
For inputs, deriving SgenThreadInfo from MonoThreadInfo seems the better solution, but more churn in the short term.

These can be found mostly by looking for THREAD_INFO_TYPE.

# enum-vs-int.

Enum vs. int will be covered up in two ways.
Overloading operators, and adding casts.
Overloading operators is imho ok.
Adding casts could be reduced by replacing some enums, just some, with define and int.

Also some of the casts were added because we have duplicated enums.
The right fix for this is https://github.com/mono/mono/pull/6700

# coop handles

Coop handles underwent very slight change -- use of `g_cast.`
Perhaps this can be improved.

jaykrell:
[goto-around-init]

before:
```
void f()
{
  goto b;
  int a = 1;
b:
  return;
```

after:
```
void f()
{
  goto b;
  int a;
   a = 1;
b:
  return;
```

Nobody likes this.
It is sprinkled around now and not  easy to find.
There are a few alternatives.
 - move locals to the top, initialized or not
[...]

PR #10342, opened 2018-08-29
[cxx] Port profilers to C++. @monojenkins skip

jaykrell:
@monojenkins skip
 [cxx] Port profilers to C++.
    This alleviates need for externC in glib.

### DIFF
--- a/mono/profiler/Makefile.am
+++ b/mono/profiler/Makefile.am
@@ -12,6 +12,7 @@ if !DISABLE_LIBRARIES
 if !DISABLE_PROFILER
 
 bin_PROGRAMS = mprof-report
+noinst_LIBRARIES = libmprof-report.a
 
 if HAVE_VTUNE
 vtune_libs = \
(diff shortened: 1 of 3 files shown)


mono/profiler/log.c
```diff
@@ -3301,7 +3313,8 @@ handle_writer_queue_entry (void)
 		if (!entry->methods)
 			goto no_methods;
 
-		gboolean wrote_methods = FALSE;
+		gboolean wrote_methods;
+		wrote_methods = FALSE;
```
alexrp: These initialization changes make the code rather ugly. Why are they necessary?

jaykrell: Cannot goto around initialization in C++.

jaykrell: https://github.com/mono/mono/issues/10319#issuecomment-416873012

jaykrell:
@alexrp @lambdageek we have decided to use extern "C" more in the runtime/glib.h, to workaround these non-core uses of non-public not-even-internal APIs, so this PR can be considered probably very optional this point. If you are ok with it, ok. If you don't want, ok too.

jaykrell:
If we do switch to C++, we'll probably want to take this PR, so that we can use C++ freely in the internal mono headers that the profiler includes.

I'm trying C++ in handle.h/handle-decl.h but I have to keep C working because of the profiler.

Or move whatever profiler uses to other/public headers. Public headers have to remain C.

kumpera:
If the provider is using internal headers, it’s fair game to convert them to c++.

kumpera:
I mean profilers

jaykrell:
Reopen and approve?

jaykrell:
@alexrp?

mono/profiler/Makefile.am
```diff
@@ -111,39 +112,51 @@ endif
...
 endif
 
-mprof_report_SOURCES = mprof-report.c
-mprof_report_LDADD = $(glib_libs) $(zlib_dep)
+mprof_report_SOURCES =
+mprof_report_LDADD = libmprof_report_a-mprof-report.$(OBJEXT) $(glib_libs) $(zlib_dep)
 mprof_report_LDFLAGS = -no-undefined
 
+libmprof_report_a_SOURCES = mprof-report.c
+libmprof_report_a_CFLAGS = @CXX_ADD_CFLAGS@
```
alexrp: It's not immediately obvious why this needs to be a static library just to change its `CFLAGS`.

mono/profiler/log.c
```diff
@@ -16,7 +16,18 @@
 #include <mono/metadata/class-internals.h>
 #include <mono/metadata/debug-helpers.h>
 #include <mono/metadata/loader.h>
+#if 0 //FIXME
 #include <mono/metadata/loader-internals.h>
+#else
+#ifdef __cplusplus
+template <typename T>
+inline void
+mono_add_internal_call (const char *name, T method)
```
alexrp: Just modify the `ADD_ICALL` macro to do the cast; no need for a template hack IMO.

alexrp: And we are certain that, on the targets where we care about runtime size, those templates don't result in a bunch of pointless instantiations increasing the size of the final library?

mono/profiler/Makefile.am
alexrp: Is this explained in some other place in our makefiles? If so, a comment here pointing to that explanation would be good.

mono/profiler/log.c
jaykrell: Yes. It is a one line inline function, that casts and calls the original.

If we templatized “deeply”, and didn’t inline, most?all systems combine identical functions (Microsoft linker got this feature circa 1996 and it isn’t alone.)

jaykrell:
Reopen. Otherwise github not seeing updates.

jaykrell:
Github would not let me reopen, so I opened another. https://github.com/mono/mono/pull/10886

PR #10343, opened 2018-08-29
Cleanup confusing code in Makefile.am.in. We don't compile/link "mono".

jaykrell:
mono is a symlink.
So mono_CFLAGS is just an intermediate, equivalent to AM_CFLAGS, plus the "REMOVE_CXX_FLAGS", which is
 a mistake, borne of me thinking mono_CFLAGS is [primary=mono]_CFLAGS like it might appear, but it isn't.
AM_CFLAGS already contains JEMALLOC_CFLAGS.
Adding it to mono_CFLAGS isn't needed.
Adding @CXX_REMOVE_CFLAGS@ to mono_CFLAGS then was unintended, leading to many warnings  with g+4.4.
So this should fix those warnings and make the code just a tiny bit simpler.

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -182,8 +182,6 @@ endif
 mono_boehm_SOURCES = \
 	main.c
 
-mono_CFLAGS = $(AM_CFLAGS) @CXX_REMOVE_CFLAGS@
-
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build Linux x64

jaykrell:
Redundant with merged https://github.com/mono/mono/pull/10345.

PR #10348, opened 2018-08-29
Fix the bug with multiline TextBox which contains /t characters in the middle of text.

FrozenShade666:
I don't know if there is a reported bug for that, I just found it and fixed it.

There is wrong cursor placement after the \t character. Example:

textBox1.Multiline = true;
textBox1.Text = "aa\taa\r\nbb\tbb\r\n";
Point pt = textBox1.GetPositionFromCharIndex(4); //pt.X is wrong for index 4 or any other char after \t in the line.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/Line.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/Line.cs
@@ -528,7 +528,7 @@ namespace System.Windows.Forms
 					size = tag.SizeOfPosition (g, pos);
 					w = size.Width;
 					newWidth = widths[pos] + w;
-					add_width += w;
+					if (text[pos] == '\t') add_width += w;
 				}
 
 				if (Char.IsWhiteSpace (text[pos]))
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/Line.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/Line.cs
@@ -472,6 +472,7 @@ namespace System.Windows.Forms
 			int wrap_pos;
 			int prev_height;
 			int prev_ascent;
+			float add_width;
 
 			pos = 0;
 			len = this.text.Length;
@@ -493,6 +494,7 @@ namespace System.Windows.Forms
 			wrapped = false;
 
 			wrap_pos = 0;
+			add_width = 0;
 
 			while (pos < len) {
 
@@ -517,14 +519,16 @@ namespace System.Windows.Forms
 				{
 					// MeasureText doesn't measure trailing spaces, so we do the best we can for those
 					// in the else branch.
+					// It doesn't measure /t characters either, we need to add it manually with add_width.
 					size = TextBoxTextRenderer.MeasureText (g, text.ToString (0, pos + 1), tag.Font);
-					newWidth = widths[0] + size.Width;
+					newWidth = widths[0] + size.Width + add_width;
 				}
 				else
 				{
 					size = tag.SizeOfPosition (g, pos);
 					w = size.Width;
 					newWidth = widths[pos] + w;
+					add_width += w;
 				}
 
 				if (Char.IsWhiteSpace (text[pos]))


marek-safar:
approve

FrozenShade666:
Looks like those 4 failures looks unrelated to code changes.

filipnavara:
Shouldn't it rather be fixed in `TextBoxTextRenderer`? The code looks like it should be able to handle it and so there's possibly some deeper problem involved.

filipnavara:
I wrote a test code for this and came to the conclusion that this is actually caused by a chain of bugs:
- `TextBoxTextRenderer` has implementation that uses `TextRenderer` and one that uses `Graphics.Draw/MeasureString`. The `TextRenderer` version is missing the `TextFormatFlags.ExpandTabs` flag. A similar thing could be missing for the other version, but I didn't check.
- `TextFormatFlags.ExpandTabs` is actually not implemented, so the tabs are never counted.

```
			var size1 = TextRenderer.MeasureText("aa\tbb", SystemFonts.MessageBoxFont, Size.Empty, TextFormatFlags.NoPadding | TextFormatFlags.NoPrefix | TextFormatFlags.ExpandTabs);
			var size2 = TextRenderer.MeasureText("aabb", SystemFonts.MessageBoxFont, Size.Empty, TextFormatFlags.NoPadding | TextFormatFlags.NoPrefix | TextFormatFlags.ExpandTabs);
			Console.WriteLine(size1);
			Console.WriteLine(size2);
```

produces `{Width=69, Height=15} {Width=33, Height=15}` on NetFX, but `{Width=32, Height=17 {Width=32, Height=17}` on Mono.

On which platform did you originally reproduce the problem? If it was non-Windows I will look at the `Graphics.MeasureString` code path too...

FrozenShade666:
I found the problem on both Linux and Windows.

TextBoxTextRenderer.MeasureText uses g.MeasureString or TextRenderer.MeasureTextInternal (it is platform dependant). Both are used by a lot of other controls, which I don't even want to touch - I'm new here and don't want to cause troubles by going too deep and changing code which I don't know. For example - Label does not include \t in its print/measure, and this is ok on both Mono and .NET.

filipnavara:
Don't get me wrong. I am more than happy that these small issues are getting fixed. I just want to understand first whether it is a genuine fix or a workaround of some other issue. My concern is that the TextBox rendering code is already quite complicated and any special handling makes it harder to maintain. It is not immediately obvious that this patch doesn't get things out-of-sync with regards to measuring code (Line.cs) and the actual drawing (Document in TextControl.cs).

FrozenShade666:
Looks like g.MeasureString ignores \t on Windows.
```
            StringFormat sf_nonprinting;
            sf_nonprinting = new StringFormat(StringFormat.GenericTypographic);
            sf_nonprinting.Trimming = StringTrimming.None;
            sf_nonprinting.FormatFlags = StringFormatFlags.DisplayFormatControl;
            sf_nonprinting.HotkeyPrefix = HotkeyPrefix.None;

            SizeF s1 = e.Graphics.MeasureString("aa\taa", textBox1.Font, 10000, sf_nonprinting);
            SizeF s2 = e.Graphics.MeasureString("aaaa", textBox1.Font, 10000, sf_nonprinting);
```
s1 = {Width = 37.8266563 Height = 19.241209}
s2 = {Width = 37.81835 Height = 19.2412071}

StringFormat settings are taken from TextBoxTextRenderer and there is no setting for \t characters.

So, TextBoxTextRenderer.MeasureText uses code that ignores \t for at least one platform.

FrozenShade666:
Any chance to merge it?

luhenry:
@filipnavara does this looks good to go to you?

filipnavara:
I guess it's fine. I will do some testing later, but it should not block this from merging.

FrozenShade666:
@monojenkins build failed

FrozenShade666:

FrozenShade666:

FrozenShade666:
@filipnavara any progress?

filipnavara:
I didn't get the time to write all the unit tests that I wanted to. However I checked that the code for calculating the lines/tags placements is shared in such a way that this change should not break the drawing.

filipnavara:

akoeplinger:

PR #10359, opened 2018-08-29
[WIP] Micro baby steps towards making m2i modular.

kumpera:
This PR is about exploring the space for improving modularity of the JIT.

It started with me trying to move the CEE_CALL code to a different file, but after adding the 15th argument to the signature it became clear we need other improvements first.

Our JIT needs more structure to its passes and analyzes with clear input/outputs. This means, among other things,
not storing temporary pass state in MonoCompile - method_to_ir does it, a lot.

This PR does some nano cleanup around that. I working to expand it to basic block formation, but that will require more than code shuffling as it requires a better representation for EH regions.

This commit series can be though of high-risk since it will touch the ugly spaguetti innards of the JIT.

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -431,6 +431,7 @@ common_sources = \
 	trace.c			\
 	trace.h			\
 	patch-info.h		\
+	method-state.h	\
 	mini-ops.h		\
 	mini-arch.h		\
 	dominators.c		\
--- a/mono/mini/method-state.h
+++ b/mono/mini/method-state.h
@@ -1,0 +1,69 @@
+#ifndef __MONO_METHOD_STATE_H__
+#define __MONO_METHOD_STATE_H__
+
+/*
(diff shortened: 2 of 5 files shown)


lewurm:
💯 on your suggested mission.

Keep in mind that a lot of the `CEE_CALL` logic is duplicated in the interpreter (`./mono/mini/interp/transform.c`), there is some opportunity to unify it 🙂

vargaz:
The reason all these fields are stored in cfg is to avoid having to pass a bunch of parameters to every function, i.e. being able to refactor the jit code more easily. Most passes don't store their information in cfg, only method-to-ir does, like cfg->cbb.

kumpera:
I'd argue all those global fields in MonoCompile makes refactoring a fucking nightmare actually.

There's so much hidden state and unknown invariants spread around that code that any changes are hilariously hard.

If anything, CEE_CALL is the proof of how bad the current design is. That code is a 1k spaghetti-monster case statement that is anything but easy to modify or refactor.

PR #10367, opened 2018-08-29
[2018-08] Bump target requirement to OSX 10.9 Mavericks

### LINKED ISSUE #9581: Update macOS minimal version

akoeplinger:
Fixes #9581
Backport of #10286


akoeplinger:
@monojenkins build pkg

PR #10369, opened 2018-08-29
Follow up to bump target requirement to OSX 10.9 Mavericks

### LINKED ISSUE #9581: Update macOS minimal version

akoeplinger:
After another grep with different search terms I found a few more matches.

Follow up to #10286 
Fixes #9581


akoeplinger:
@monojenkins build pkg

alexischr:

akoeplinger:
@alexischr I already added this to https://github.com/mono/mono/pull/10367 :)

PR #10375, opened 2018-08-29
[cxx] Eliminate THREAD_INFO_TYPE by suggestion, to fix typesafe linkage. @monojenkins skip

jaykrell:
@monojenkins skip
This is a suggested alternative to https://github.com/mono/mono/pull/10294 and https://github.com/mono/mono/pull/10297.
As `THREAD_INFO_TYPE` wasn't covering up all that much.
Remove a hack instead of pile on?

In future we can still derive `SgenThreadInfo` from `MonoThreadInfo`, which will further clean this up.
We'd also win just by having both `stack_end` and `stack_start` in `MonoThreadInfo`, instead of one in `mono` and one in `sgen`.

A few lines of this are optional, where we have `info` and `mono_thread_info`, I replaced `info->client_info.info` with `mono_thread_info->`.

Also, the `FOREACH` macros could still be factored on `THREAD_INFO_TYPE`.
And the extern "C" functions could be as well.
I chose to eliminate `THREAD_INFO_TYPE` entirely, rather than just where it causes typesafe linkage violations, but there is flexibility here.

Also strengthened type in `checked-build.{c,h]` from `void*` to `MonoThreadInfo*`. (was `void*` and `THREAD_INFO_TYPE`).

### DIFF
--- a/mono/metadata/sgen-mono.c
+++ b/mono/metadata/sgen-mono.c
@@ -73,9 +73,9 @@ static gboolean
 ptr_on_stack (void *ptr)
 {
 	gpointer stack_start = &stack_start;
-	SgenThreadInfo *info = mono_thread_info_current ();
+	MonoThreadInfo *info = mono_thread_info_current ();
 
-	if (ptr >= stack_start && ptr < (gpointer)info->client_info.info.stack_end)
+	if (ptr >= stack_start && ptr < (gpointer)info->stack_end)
 		return TRUE;
 	return FALSE;
 }
@@ -836,7 +836,7 @@ mono_gc_clear_domain (MonoDomain * domain)
 	sgen_clear_nursery_fragments ();
 
 	FOREACH_THREAD_ALL (info) {
-		mono_handle_stack_free_domain ((HandleStack*)info->client_info.info.handle_stack, domain);
+		mono_handle_stack_free_domain ((HandleStack*)info->handle_stack, domain);
 	} FOREACH_THREAD_END
 
 	if (sgen_mono_xdomain_checks && domain != mono_get_root_domain ()) {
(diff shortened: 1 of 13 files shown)


jaykrell:
This will much cleaner if/when `SgenThreadInfo` derives from `MonoThreadInfo`.

mono/metadata/sgen-mono.c
```diff
@@ -2162,37 +2166,38 @@ sgen_client_scan_thread_data (void *start_nursery, void *end_nursery, gboolean p
 	return;
 #endif
 
-	FOREACH_THREAD_EXCLUDE (info, MONO_THREAD_INFO_FLAGS_NO_GC) {
+	FOREACH_THREAD_EXCLUDE (mono_thread_info, MONO_THREAD_INFO_FLAGS_NO_GC) {
+		SgenThreadInfo *info = (SgenThreadInfo*)mono_thread_info;
```
kumpera: IDK about having two variables aliasing to the same value.
I think we should stick with a single one and emit casts in the very few places that needs it.
As you mentioned on slack, moving stack_start would fix most of this.

jaykrell: The problem with moving stack_start is I have to touch all the uses. ok?
Also could keep FOREACH factored on type, that might help. I can try.

jaykrell:
I think having `info` and `mono_thread_info` is ok, but the selective replacement isn't well motivated, granted, so then `mono_thread_info` would just exist to be assigned to `info`, which is ok if that is all, or can try to remove it.
Let me try just factoring `FOREACH` and see how it looks. Else cast.

jaykrell:
https://github.com/mono/mono/pull/10403 is a likely alternative to this for the time being, as was https://github.com/mono/mono/pull/10160 before it.

jaykrell:
I can try this again, but it has lots its urgency due to https://github.com/mono/mono/pull/10403.

lambdageek:
We went with #10403

PR #10378, opened 2018-08-30
[cxx] Port MonoPosixHelper to C++, except zlib, because there is some glib use,

jaykrell:
and this reduces extern "C" therein.
zlib is K&R and does not use glib.
This is an alternative to https://github.com/mono/mono/pull/10341.
And addresses this comment https://github.com/mono/mono/pull/10160#pullrequestreview-148551783.

There are other alternatives.
 - This compiles more of the code as C++ than uses glib; make a more precise division instead.
 - Mark some/all of glib.h as extern "C" -- and leave this directory alone. (https://github.com/mono/mono/pull/10379)

extern "C" is applied a bit liberally here, probably redundantly, as  it is mostly already present and in .h files.

### DIFF
--- a/support/Makefile.am
+++ b/support/Makefile.am
@@ -1,16 +1,14 @@
-if HOST_WIN32
-SUPPORT=
-else
-SUPPORT=libMonoSupportW.la
-endif
-
 if BUILD_SUPPORT
-lib_LTLIBRARIES = 				\
-	libMonoPosixHelper.la			\
-	$(SUPPORT)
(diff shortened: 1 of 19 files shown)


jaykrell:
We have decided to use extern "C" more, to cover up this and others.
This PR is ok, but very optional now.

jaykrell:
@jonpryor Do you own this? We are less interested in this now, but if you want it, ok.

PR #10379, opened 2018-08-30
[rfc] [cxx] Mark the 14 functions used by MonoPosixHelper as extern "C".

jaykrell:
This is an alternative to:
https://github.com/mono/mono/pull/10378
https://github.com/mono/mono/pull/10341

There are more:
We could mark the entire glib.h as extern "C", or slightly larger chunks.

This would enable us to leave MonoPosixHelper completely untouched.

I prefer closer to https://github.com/mono/mono/pull/10378, so that, e.g. we might in future templatize the hashtable, but presenting this data to decision making.

Even the templatization fantasy could be with new/alternate types though, possibly reusing existing, possibly under ifdef __cplusplus. Or we could convert MonoPosixHelper later anyway.

For further reference,  profilers use:
```
g_calloc
g_free
g_malloc
g_realloc
```

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -264,12 +264,22 @@ typedef guint32 gunichar;
 #define G_GUINT32_FORMAT PRIu32
 #define G_GINT32_FORMAT PRIi32
 
+#ifdef __cplusplus
+#define G_USED_BY_MONO_POSIX_HELPER extern "C"
+#else
+#define G_USED_BY_MONO_POSIX_HELPER /* nothing */
+#endif
+
 /*
  * Allocation
  */
+G_USED_BY_MONO_POSIX_HELPER
 void g_free (void *ptr);
+G_USED_BY_MONO_POSIX_HELPER
 gpointer g_realloc (gpointer obj, gsize size);
+G_USED_BY_MONO_POSIX_HELPER
 gpointer g_malloc (gsize x);
+G_USED_BY_MONO_POSIX_HELPER
 gpointer g_malloc0 (gsize x);
 gpointer g_calloc (gsize n, gsize x);
 gpointer g_try_malloc (gsize x);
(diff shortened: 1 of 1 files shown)


kumpera:
The fact that monoposixhelper uses implementation details of mono like that is really bad, it's fragile.
It would be so much better if we change it not to depend on those internal symbols at all.
Looking at the symbols used, I see 3 groups: ghashtable, glog and g_malloc.

glog can be addressed by exposing a new API entry point in ùtils/mono-logger.h` that does so. We'd expose it as logging under a new "user" category.
g_free -> mono_free
g_malloc -> add `mono_malloc` to `utils/mono-publib.h`
ghashtable could be statically linked into it, I suppose. This is the more complicate bit I'm not 100% sold on how to address it.

What do you think?

lambdageek:
I think we're facing two separate tasks:
1. Now that `eglib` is pulled in closer to Mono proper, how do we untangle `libMonoPosixHelper`'s Mono dependency?  @kumpera 's suggestion is one way forward.
2. Can we build the core of Mono with a C++ compiler?

This PR should only address (2).  Which it does.  I see no reason to build `support/` with C++ - so I like this PR better than #10378 (and much more than #10341

kumpera:
@lambdageek my suggestion is to address (2) by way of (1) as it's a net positive change regardless.

jaykrell:
If the dependency is via static linking (yes), and/or we build it in our tree (yes), then how worried about this fragility are we really? I can only stir up and put back to sleep so many sleeping dogs. Leave this one alone?

I'm really inclined to port to C++, but not zlib, and/or sprinkle or blanket extern "C", and move on.

kumpera:
libMonoPosixHelper is both statically and dynamically linked into mono.
It's built together with mono but maintained by a different set of ppl

jaykrell:
We are going to do this, but I don't think in this form. I will amend this PR, or make another (because this is also about libtest and profilers).
Either `G_BEGIN_DECL` or `G_EXTERN_C`.
I guess I favor `G_EXTERN_C` for "function at a time".

We failed to address @kumpera concerns in the meeting of @lambdageek and @luhenry, but one could argue that is a separate concern??

jaykrell:
The PR itself is redundant with merged https://github.com/mono/mono/pull/10403. The issue raised remains open.

PR #10391, opened 2018-08-30
Revert "[2018-06] [System]: `MonoTlsStream` is now `IDisposable` and cleans up on error.

baulig:
Revert PR #10336.

This reverts commit 3bb7817ffd8c92458e875a179c62f26c390c9e0e.

### DIFF
--- a/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
+++ b/mcs/class/System/Mono.Net.Security/MobileAuthenticatedStream.cs
@@ -77,7 +77,7 @@ namespace Mono.Net.Security
 			Settings = settings;
 			Provider = provider;
 
-			readBuffer = new BufferOffsetSize2 (16384);
+			readBuffer = new BufferOffsetSize2 (16834);
 			writeBuffer = new BufferOffsetSize2 (16384);
 			operation = Operation.None;
 		}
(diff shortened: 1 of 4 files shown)


nealef:
Rather than revert can you just set readBuffer back to 16834?

baulig:
That's not going to fix it completely.   We have decided to revert in 2018-06, but try to fix in master and 2018-08.  We can later put it back into 2018-06 once we have a better understanding of what was causing the problems.

nealef:
Were my comments in the PR accurate: The problem seems to occur because the length of the data requested is larger than the size of readBuffer and there appears to be no logic in OnRead to read the remainder?
```
                int OnRead (IntPtr data, int dataLength, out int wantMore)
                {
                        bool wantMoreBool;
                        int index = 0, remaining = dataLength;
                        var buffer = new byte[dataLength];
                        while (remaining > 0) {
                                var ret = backend.Read (buffer, 0, dataLength, out wantMoreBool);
                                wantMore = wantMoreBool ? 1 : 0;
                                if (ret <= 0)
                                        return ret;
                                 Marshal.Copy (buffer, index, data, ret);
                                 index += ret;
                                 remaining -= ret;
                        }
                        return index;
                }
```
Either that or check that the data length requested is greater than the size of readBuffer and reallocate readBuffer to be that size?

baulig:
Yeah, your comments were accurate.  The problem is that after changing the size to 16384, the read buffer was not large enough to hold a full TLS Record (which is 16384 bytes plus header unencrypted).

PR #10392, opened 2018-08-30
Revert "An implementation of the ChainStatus property for Btls. (#10216).

baulig:
Revert #10216.

This reverts commit a51f37f6c59d51ace1b09f66b016067465a8b29b.


baulig:
We can also instead back port https://github.com/mono/mono/pull/10396.

PR #10134, opened 2018-08-15
An implementation of the ChainStatus property for Btls

nealef:
An implementation of the ChainStatus property for Btls. When we execute the certificate verification callback we just don't test for success and set status11. Now we get the error (if any) and map the error code to those defined by ChainStatusFlags and add it to a list of X509ChainStatus objects for that chain. Now when get_ChainStatus is called we convert that list to an array of X509ChainStatus and return it.

PR #10396, opened 2018-08-30
[System]: Fix `MonoBtlsProvider.CheckValidationResult ()` logic.

baulig:
See my comments in https://github.com/mono/mono/pull/10134 for details.

The problem was that
1. It is wrong to assume that `chain != null`.  In fact, it likely won't be if `wantsChain == false`.
2. Similarly, the `storeCtx` should only be used if `storeCtx != NULL && wantsChain == true`.
3. It was missing `MonoBtlsX509Error.HOSTNAME_MISMATCH`.
4. I reformatted and re-indented the method to match Mono's style.


baulig:
@monojenkins backport 2018-08

baulig:
@monojenkins 2018-06

baulig:
@monojenkins backport 2018-06

Issue #10376, opened 2018-08-29
SDKs development configuration improvements


kumpera:
This is a list of small improvements I think we could do to the SDKs to improve our quality-of-life when using it as a way to work on non-desktop targets:

- [ ] Add a configure-style script (being discussed in #10239)
- [ ] An option to disable llvm usage by ios/android/wasm *1
- [ ] Audit and document all external properties that can be used to control the SDKs behavior *2 *3
- [ ] Ensure that configuring for nothing builds nothing
- [ ] Add high level targets in `sdks` that go together with `configure` to build/test the whole thing
- [ ] Transition CI to use the mono-team-friendly targets as a way to ensure they don't regress
- [x] Cleanup the llvm targets so they have consistent output directories *4 - https://github.com/mono/mono/pull/10445 https://github.com/mono/mono/pull/10427 https://github.com/mono/mono/pull/10398

Notes:
- 1 This would mean no AOT under wasm, we can debate later
- 2 By audit I mean ensure positive flags don't start as negatives, I.E. IGNORE_PACKAGE_LLVM
- 3 Bonus point if all of them are integrated into configure
- 4 It's confusing that wasm uses `ios-llvm32` and `make package llvm-lvm32` leaves an empty `out/llvm-llvm32`

kumpera:
cc @luhenry @vargaz 
I'm volunteering for doing all of this work and whatever else comes up, btw.

kumpera:
This is from a discussion with @alexischr on #mono-community.

If we use `--enable-llvm` instead of `--with-llvm=` to sprinkle llvm into our build, it uses `llvm/llvm.mk` directly, which leaves build artifacts in that place.

Maybe instead of the convoluted relation between this and the SDKs we could just have it ask the SDKs for the right llvm and be done with it.

luhenry:
I am right now at work to unify a bit more how we provision LLVM for iOS and Android. I would make what's necessary for WASM to use it, so I would need to understand better how WASM uses it.

For the `llvm/build/mk` vs `sdks/builds/llvm.mk` question, it was decided when we did it for the latter to depend on the former (and not the other way around), so the source of truth for how Mono builds LLVM is in `llvm/build.mk` which is easier to use than `sdks/builds/` from the Mono sources. That's mostly because we do **not** want to have automake involved in `sdks/` while we still need to pass parameters from `configure.ac` to `llvm/Makefile.am`.

[...]

PR #10398, opened 2018-08-30
[2018-06] [sdks] Add cross-{arm,arm64,x86,x86_64}{,-win} runtimes for XA + Unify provisioning of LLVM for both Android and iOS.

luhenry:


luhenry:
Fixes some of https://github.com/mono/mono/issues/10376.

PR #10401, opened 2018-08-31
[cxx] Add missing ICALL_EXPORTs, esp. but not only if excess G_BEGIN_DECLS is removed.

jaykrell:


jaykrell:
@monojenkins build  OS X x64

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build OS X x64

PR #10218, opened 2018-08-21
[2018-06][aot] Fix win32 build

luhenry:

PR #10402, opened 2018-08-31
[2018-06] [runtime] Fix mxe+release_60 llvm build

alexanderkyte:
This PR contains the commits from https://github.com/mono/mono/pull/10218 . 

Comments there may be relevant to future readers of this PR.


configure.ac
```diff
@@ -3827,14 +3827,16 @@ if test "x$enable_llvm" = "xyes"; then
...
+       LLVM_SYSTEM_LIBS="-lshell32 -lpsapi -limagehlp -ldbghelp -lm -lz -luuid -lole32"
        LLVM_LIBS="-lLLVMLTO -lLLVMObjCARCOpts -lLLVMLinker -lLLVMipo -lLLVMVectorize -lLLVMBitWriter \
-         -lLLVMTableGen -lLLVMDebugInfo -lLLVMOption -lLLVMSelectionDAG -lLLVMAsmPrinter \
-         -lLLVMMCDisassembler -lLLVMJIT -lLLVMAnalysis -lLLVMTarget \
-         -lLLVMIRReader -lLLVMAsmParser -lLLVMLineEditor -lLLVMMCAnalysis -lLLVMInstrumentation \
-         -lLLVMInterpreter -lLLVMCodeGen -lLLVMScalarOpts -lLLVMInstCombine -lLLVMTransformUtils \
-         -lLLVMipa -lLLVMAnalysis -lLLVMProfileData -lLLVMMCJIT -lLLVMTarget -lLLVMRuntimeDyld \
-         -lLLVMObject -lLLVMMCParser -lLLVMBitReader -lLLVMExecutionEngine -lLLVMMC -lLLVMCore \
+         -lLLVMTableGen  -lLLVMOption -lLLVMSelectionDAG -lLLVMAsmPrinter \
+         -lLLVMARMAsmPrinter -lLLVMX86AsmParser \
```
luhenry: Why are these 2 needed when we already pass them at https://github.com/mono/mono/pull/10402/commits/8cd8c3ee378a5a474b235016a26617924e3cdb75#diff-67e997bcfdac55191033d57a16d1408aR3758 and https://github.com/mono/mono/pull/10402/commits/8cd8c3ee378a5a474b235016a26617924e3cdb75#diff-67e997bcfdac55191033d57a16d1408aR3754 respectively?

luhenry:
@lateralusX could you please confirm it works for your use case as well? That's for `2018-06`, so we should figure out if we will want that in `2018-08` and `master` as well.

lateralusX:
@luhenry, @alexanderkyte I have successfully build release_60 using mingw and linked towards mingw build mono with some adjusted linker flags as well, will create a PR during the day for that change in master. I have just tested manual build LLVM release (not using MXE) but with similar cmake flags.

lateralusX:
#10402

configure.ac
alexanderkyte: These defined were found by removing the flags that llvm-config didn't support, looking for the missing symbols, grepping for their definitions in the `.a` files, and including their names here. I'm not sure why our llvm blob had undefined references to these symbols if we were passing them above. The above seems to be host-specific, I saw the relevant failure when building MinGW x64.

luhenry:
@alexanderkyte I merged `mono:2018-06` into your PR and enabled the builds of `android-cross-{arm,arm64,x86,x86_64}-win` so we can verify your PR on CI.

luhenry:
So it doesn't work completely, the XA build later segfaults when running `"/Users/ludovic/android-toolchain/mxe-b9cbb53/bin/x86_64-w64-mingw32.static-strip" -S "/Users/ludovic/Xamarin/xamarin-android/bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/cross-arm64.exe"`.

That looks awfully related to this changes since it fails to strip the `cross-arm64.exe` Windows executable in which the LLVM libraries were linked. Stripping `cross-arm.exe` works though.

lateralusX:
@alexanderkyte Just checked a sample method of doing AOT compile of basic.exe with linked LLVM x86_64 cross compiler and the code looked OK.

alexanderkyte:
@luhenry yeah, those were the linking issues I had mentioned before. I did some tracing of it and it seems to just segfault with a modest stack depth. I fed the binary into otool, objdump, nm, and other binary-manipulating programs and it managed to show me all of the import/export tables without problem. 

I did a cursory analysis of that strip binary and all of the data that it touches seems to be asserted to be correctly formed by the OSX-side tools when I run them on the binary.

alexanderkyte:
@luhenry @lateralusX:
I added a few system lib flags, is there any chance that those are coming from the wrong place?

PR #10406, opened 2018-08-31
Kill sln dependencies

kg:
We currently use .sln level dependencies to order builds, but this is kind of fragile and seemingly unreliable. This PR moves to generating project references for all dependencies, and also changes some flags to make sure dependencies are fully resolved before we run build events. This should fix a culevel problem when using command-line msbuild.

With these changes, building bcl.sln with msbuild.exe finally works again...

Contributes to #6886


akoeplinger:
@monojenkins commit csproj

PR #10423, opened 2018-08-31
Cleanup w32socket-unix.c. Strictly cosmetic. @monojenkins skip

jaykrell:
@monojenkins skip    (low priority)


luhenry:
Strictly cosmetic so quite useless. All this is going away with the import of Socket from corefx, so I wouldn't spend any time on making it prettier.

PR #10426, opened 2018-08-31
[Mono.Options] Add `CommandSet.GetCompletions()`

jonpryor:
Add a `Mono.Options.CommandSet.GetCompletions()` method, which returns
all possible commands which match a given "completion prefix".

This is intended for e.g. bash completion support.


alexischr:
@monojenkins commit apidiff

PR #10428, opened 2018-08-31
Remove unused mono_pages_not_faulted, autoconf for mlock instead of mincore.

jaykrell:


support/sys-mman.c
```diff
@@ -89,7 +89,7 @@ Mono_Posix_Syscall_msync (void *start, mph_size_t len, int flags)
 int
 Mono_Posix_Syscall_mlock (void *start, mph_size_t len)
 {
-#if !defined(HAVE_MINCORE)
+#if !defined(HAVE_MLOCK)
```
jaykrell: Or __HAIKU__ like munlock?

NattyNarwhal: i'd recommend just testing for mlock/munlock, and that should cover Haiku

jaykrell:
Just one test for both, eh? Splitting them up would kinda be nonsense?
I’ll do that.

jaykrell:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #10429, opened 2018-08-31
Move the PreBuild dependency property later in the file so targets can't override it

kg:


kg:
@monojenkins build Linux x64

kg:
@monojenkins commit csproj

PR #10434, opened 2018-09-01
[Facades] Clean up no longer needed references

marek-safar:
and build more facades in parallel


akoeplinger:
@monojenkins commit csproj

PR #10438, opened 2018-09-01
[cxx][wip] 4

jaykrell:
This should match the combination of a few PRs to master, plus hardcoding -enable-cxx in  configure.ac.

This is the real test bed, as building locally I only usually do Mac, and the Jenkins job is only Linux/x64.


jaykrell:
@monojenkins build failed

jaykrell:

PR #10442, opened 2018-09-02
[cxx] Add ReleaseCxx and DebugCxx configurations to MSBuild project files.

jaykrell:
And then sometimes use /TP to compile source files as C++.
All “editing” done via Visual Studio GUI (and then undo a version change from 14
to 15). Note that not all project files use /TP, but all have ReleaseCxx
and DebugCxx configurations (i.e. sometimes they are just the same as
Release and Debug). Perhaps that can be improved, though in
in the long run it won't matter.

This is the build system support for using Visual C++, compiling as C++.


jaykrell:
@monojenkins build failed

lambdageek:
@lateralusX @ntherning Could you review this for us?

luhenry:
We do not have an existing concept of Debug and Release builds for the runtime, and have had no use of it so far. We do not _even_ want to have separate builds for debug and release.

@jaykrell if there is no **clear case scenario** for that feature, this is not going to go in (ie technical reasons won't do it)

jaykrell:
There is already Release and Debug there.
I duplicated them and used this to find the problems compiling with Visual C++.
Otherwise there is no coverage with Visual C++.
This is how we compile using Visual C++.
See here https://github.com/mono/mono/pull/10438/commits/68cabfe480caaa2ab3754cc14b7c624bc86dd681

lateralusX:
Having both Debug/Release configurations in VS projects is standard and something that is very useful when developing/debugging the mono runtime on Windows. For the official mono Windows builds we only build and ship the release configuration, but for some platforms we actually build and ship both since it makes it much easier for developers to for example debug the embedding of the mono runtime inside their own VS debug builds. Current mono VS solution also makes the delta between the Debug/Release small to maintain and major differences are coming from the original Debug/Release VS templates in the first place. So there is a lot of value to have both debug and release configurations in the VS mono runtime builds.

Having that said we might not need to duplicate the configurations (as pointed out in this PR) and instead use global properties/config to apply the same affect, but reduce the number of configurations to maintain.

@jaykrell Is the only thing needed to build all source files using C++ compiler (in this case passing /TD)? No other changes to compiler/linker flags etc?

jaykrell:
@lateralusX Correct /TP. Not on all projects, but intended on on sgen, metadata, mini, utils, glib.
And some others like monodis -- maybe I missed some.
P is for PlusPlus.
Lowercase p applies to one file, capital P applies to all files (or all subsequent files?).

A command line property can work, however $(Configuration) feeds into the output directory, kinda by stylistic default as well and multiple side-by-side per-configuration output directories can be helpful for comparing them (and using up diskspace, granted).

I also don't know how custom conditionals end up in terms of IDE parsing and editing -- do they work?
I kinda let the IDE guide me here.
I admit, I didn't try much, custom properties, started trying to learn msbuild, before realizing this easy good solution.

Eventually this PR will be mostly undone, and /TP either dropped or folded into the regular configurations.
But for now I think having multiple configurations is useful and cheap.

lateralusX:
@jaykrell Custom global settings are evaluated when set in the IDE and can be passed on command line to MS build. We already use similar approach to build different variations of bohem/sgen gc, static/dynamic runtime, LLVM support etc, without exploding the number of configurations to maintain in the projects and some of them will also make sure they end up in different build folders for example mono sgen will end up C:\github\mono\msvc\build\sgen\x64\bin\Debug\ to have separate build folders. I agree it is not as visible as the build configurations but much more lightweight and easier to maintain. I believe the C++ support could be done similar introducing a change that is just a couple of lines instead of above and when we decide that the C++ will be the default it will be much easier to switch. I could test it out if you want to see if things works as expected. All needed changes already in master?

jaykrell:
@lateralusX I believe everything is in master, except https://github.com/mono/mono/pull/10444, still awaiting CI to complete. Some of this just a few minutes ago.

jaykrell:
I understand the MSBuild command line, but how does one request the setting within the IDE, and what does authoring within the IDE look like, I wonder?

I wouldn't want to start the IDE up, for example, with different environment variables, to switch configurations, though I know that works.

Does "batch build" in the IDE allow enumerating property values to build multiple?

lateralusX:
Will look at this tomorrow.

jaykrell:
I suggest we go with what I did. It was quick & easy & follows prior convention & works & has multiple advantages over other conventions, & anything we do here either way is temporary.

lateralusX:
I believe the temporary nature of this should try to make smallest possible change. I have an example in https://github.com/mono/mono/pull/10487 that alters around 70 line instead of 5000+and will enable C++ compiler for selected projects based on a global property in mono.props, building into separate build folders. This change uses similar convention as already used when configuring GC, dynamic/static builds, LLVM etc and is easy to alter/remove ones we make C++ compiler default for all projects in the future.

PR #10443, opened 2018-09-02
[cxx] Adapt typesafe coop handles for standard/Visual C++.

jaykrell:
W/o gcc/clang language extension `__typeof__`. This is not meant to be
long term idiomatic C++, just enough to get us by for now.

The solution does not quite work for type unsafe handles, because
checking for this == NULL is not defined. A non-member function
would be needed.

C++11 decltype could be used instead at the MONO_TYPEOF_CAST level,
but this solution is trivial C++98.

### DIFF
--- a/mono/metadata/handle.h
+++ b/mono/metadata/handle.h
@@ -384,7 +384,7 @@ Handle macros/functions
 
 #if MONO_TYPE_SAFE_HANDLES
 #define TYPED_HANDLE_DECL(TYPE)							\
-	typedef struct TYPED_HANDLE_NAME (TYPE) {				\
+	typedef struct {							\
 		MONO_IF_CPLUSPLUS (						\
 			MONO_ALWAYS_INLINE					\
 			TYPE * GetRaw () { return __raw ? *__raw : NULL; }	\
--- a/mono/metadata/handle.h
+++ b/mono/metadata/handle.h
@@ -376,11 +376,23 @@ Handle macros/functions
  * #endif
  */
 
+#ifdef __cplusplus
+#define MONO_IF_CPLUSPLUS(x) x
+#else
+#define MONO_IF_CPLUSPLUS(x) /* nothing */
+#endif
+
 #if MONO_TYPE_SAFE_HANDLES
 #define TYPED_HANDLE_DECL(TYPE)							\
-	typedef struct { TYPE **__raw; } TYPED_HANDLE_PAYLOAD_NAME (TYPE),	\
-					 TYPED_HANDLE_NAME (TYPE),		\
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

mono/metadata/handle.h
```diff
@@ -376,11 +376,23 @@ Handle macros/functions
...
+#define MONO_IF_CPLUSPLUS(x) /* nothing */
+#endif
+
 #if MONO_TYPE_SAFE_HANDLES
 #define TYPED_HANDLE_DECL(TYPE)							\
-	typedef struct { TYPE **__raw; } TYPED_HANDLE_PAYLOAD_NAME (TYPE),	\
-					 TYPED_HANDLE_NAME (TYPE),		\
-					 TYPED_OUT_HANDLE_NAME (TYPE);		\
+	typedef struct {							\
+		MONO_IF_CPLUSPLUS (						\
```
lambdageek: If we wanted a non-member, we could do something like this: 
```c++
#define TYPED_HANDLE_DECL(TYPE) \
  typedef struct { \
    MONO_IF_CPLUSPLUS (typedef TYPE raw_type); \
    TYPE **__raw; \
  } ...

namespace mono {
  namespace handle {
    template <typename H>
    typename H::raw_type* get_raw (const H& h) {
       return h.__raw ? *__raw : NULL;
    }
  }
}

#define MONO_HANDLE_RAW (handle) (mono::handle::get_raw (handle))
```

jaykrell:
I would do it like:
https://github.com/mono/mono/commit/09666f3297e42ce5aea53a755dcdbc03c1c56820
```
#ifdef __cplusplus
extern "C++"
{
#if safe

template <typename T>
struct TypedHandle { T** __raw; }; // unchanged representation

template <typename T>
inline T*
mono_typed_handle_get_raw (TypedHandle<T> handle)
{
	return handle.raw ? *handle.raw : NULL;
}

#else

template <typename T>
struct TypedHandle { T* __raw; };

template <typename T>
inline T*
mono_typed_handle_get_raw (TypedHandle<T>* handle)
{
	return handle ? handle->raw : NULL;
}

#endif

}
```

jaykrell:
@monojenkins build OS X x64

lambdageek:
yea good point.

jaykrell:

jaykrell:
Yes. The existing code is not valid C++.
It uses a gcc/clang extension, and takes advantage of a quirk in Visual C++ when compiling as C, that is not allowed as C++. There are errors on every invocation of `MONO_HANDLE_RAW`.

https://github.com/mono/mono/blob/master/mono/metadata/handle.h#L419

Issue #9996, opened 2018-08-09
The type initializer for 'System.Data.Entity.Core.Objects.ELinq.ExpressionConverter' threw an exception. (regression in Mono 5.14)


alex-mazzariol:
## Steps to Reproduce

1. Have a project for .NET Framework 4.7 with EntityFramework 6.2 (downloaded from NuGet) connecting to a database
2. Try an operation with the database
3. Operation initializes ELinq.ExpressionConverter, which results in ArgumentNullException

A test-case has been attached: [mono-elinq-test.zip](https://github.com/mono/mono/files/2273250/mono-elinq-test.zip)

## Steps to Use the Test Case

1. Open the solution with any recent MonoDevelop (e.g. 7.5) and let it restore solution packages
2. Correct the ConnectionString property that is set in `DbConnFact.cs` to point it to a valid running PostgreSQL instance, with a valid username and password. Note that a correct execution will create a database with the same name as the user connecting, so maybe use a disposable instance (i.e. properly exposed container?)
3. Compile the solution. There are no specific problems during this stage: either msbuild from mono 5.10, 5.12 or 5.14 have been successfully used without difference.
4. Test with mono docker image 5.14: `docker run -it --rm -v /[solution-directory]/mono-elinq-test/bin/Debug/:/test mono:5.14 /bin/sh` and then type `mono /test/mono-elinq-test.exe` to check the crash.
5. Test with mono docker image 5.12: `docker run -it --rm -v /[solution-directory]/mono-elinq-test/bin/Debug/:/test mono:5.12 /bin/sh` and then type `mono /test/mono-elinq-test.exe` to check the successful execution.

### Current Behavior
As of Mono 5.12.0.301 EntityFramework 6.2 did work successfully, but since 5.14.0.177 the type initializer throws an ArgumentNullException - full stack trace appended.

### Expected Behavior
The operation does not produce any exception.

## On which platforms did you notice this

[ ] macOS
[X] Linux (CentOS 7 fully updated)
[ ] Windows
[X] official Docker image (https://hub.docker.com/_/mono/)

**Version Used**:
5.12.0.301 (without problems), 5.14.0.177 (with problem)

## Stacktrace

```
<System.TypeInitializationException: The type initializer for 'System.Data.Entity.Core.Objects.ELinq.ExpressionConverter' threw an exception. ---> System.TypeInitializationException: The type initializer for 'MethodCallTranslator' threw an exception. ---> System.ArgumentNullException: Value cannot be null.
[...]

PR #10452, opened 2018-09-03
[corlib] Bring back String.Concat with __arglist

EgorBo:
Fixes https://github.com/mono/mono/issues/9996 
Tested repro project from the link on a local PostgresSQL server and EntityFramework packages.

See https://github.com/aspnet/EntityFramework6/blob/master/src/EntityFramework/Core/Objects/ELinq/MethodCallTranslator.cs#L1482-L1497


jaykrell:
Test case?

EgorBo:
@jaykrell Added!

marek-safar:
@monojenkins commit apidiff

PR #10466, opened 2018-09-04
Avoid deadlock during GC. Check for GC_DONT_GC environment variable a…

joncham:
…t startup rather than during collection with world stopped since g_hasenv may take OS locks and encounter deadlock.

### DIFF
--- a/mono/metadata/boehm-gc.c
+++ b/mono/metadata/boehm-gc.c
@@ -58,6 +58,7 @@ void *pthread_get_stackaddr_np(pthread_t);
 #define MIN_BOEHM_MAX_HEAP_SIZE (MIN_BOEHM_MAX_HEAP_SIZE_IN_MB << 20)
 
 static gboolean gc_initialized = FALSE;
+static gboolean gc_dont_gc_env = FALSE;
 static mono_mutex_t mono_gc_lock;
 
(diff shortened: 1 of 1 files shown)


jaykrell:
Suggest g_hasenv or new function take a pointer to a "cache" (or rather, a store-once, use-forever-in-the-process canonical storage), and find all calls in mono, and initialize them all up front. Unless someone protests and says they set/clear it during the runtime lifetime.

mono/metadata/boehm-gc.c
```diff
@@ -1388,7 +1392,7 @@ mono_gc_is_moving (void)
 gboolean
 mono_gc_is_disabled (void)
 {
-	if (GC_dont_gc || g_hasenv ("GC_DONT_GC"))
+	if (GC_dont_gc || gc_dont_gc_env)
```
jaykrell: Function can just be one line `return GC_dont_gc || gc_dont_gc_env;`'

joncham:
@jaykrell could we not have a `g_init` or something that we call on startup which caches env once internally? Saves callers all being changed.

jaykrell:
Yes, a variety of tactics would work.

luhenry:
@joncham changing the mechanism of `g_hasenv` could be a great followup PR, but for the sake of context, let's not do that in this one.

PR #10468, opened 2018-09-04
[runtime] Fix stored exception throws

alexanderkyte:
On cooperative suspend, the wrapper frames capture the exception
in a MonoError, which then must be thrown. The problem with treating
it as a CIL THROW opcode is that every time you THROW, you completely
reset the stacktrace's collectin of saved frames.

The RETHROW opcode will force the JIT to rely on an in-flight exception,
that in our case has been taken our of the in-flight context.

In order to enable the runtime to store and multiplex in-flight
exceptions as MonoErrors, without losing throw context, I made these
places of THROW-ing into places where we directly make a JIT icall that
rethrows the exception in question. On each platform, we were already
using very similar control flow when we translated to a "throw".

The most impactful difference on the generated code is that an internal
runtime function that got "rethrow = FALSE" for this exception will now
see "rethrow = TRUE", and not zero out the pointers to the trace_ips.

I did a few prototypes here, and all of the other methods of doing this
either required very large churn in the codegen backend, or would leave
a dirty global state that would undermine the ability of the MonoError
to keep multiple ones in-flight.

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -1173,6 +1173,11 @@ emit_thread_interrupt_checkpoint_call (MonoMethodBuilder *mb, gpointer checkpoin
 	mono_mb_emit_byte (mb, CEE_DUP);
 	pos_noex = mono_mb_emit_branch (mb, CEE_BRFALSE);
 
+	mono_mb_emit_byte (mb, CEE_DUP);
+	mono_mb_emit_ldflda (mb, MONO_STRUCT_OFFSET (MonoException, caught_in_unmanaged));
+	mono_mb_emit_byte (mb, CEE_LDC_I4_1);
+	mono_mb_emit_byte (mb, CEE_STIND_I4);
+
 	mono_mb_emit_byte (mb, MONO_CUSTOM_PREFIX);
 	mono_mb_emit_byte (mb, CEE_MONO_RETHROW);
 
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -1172,7 +1172,10 @@ emit_thread_interrupt_checkpoint_call (MonoMethodBuilder *mb, gpointer checkpoin
 	/* Throw the exception returned by the checkpoint function, if any */
 	mono_mb_emit_byte (mb, CEE_DUP);
 	pos_noex = mono_mb_emit_branch (mb, CEE_BRFALSE);
-	mono_mb_emit_byte (mb, CEE_THROW);
+
+	mono_mb_emit_byte (mb, MONO_CUSTOM_PREFIX);
+	mono_mb_emit_byte (mb, CEE_MONO_RETHROW);
+
 	mono_mb_patch_branch (mb, pos_noex);
 	mono_mb_emit_byte (mb, CEE_POP);
 	
--- a/mono/mini/interp/transform.c
+++ b/mono/mini/interp/transform.c
@@ -4321,6 +4321,12 @@ generate (MonoMethod *method, MonoMethodHeader *header, InterpMethod *rtm, unsig
 		case MONO_CUSTOM_PREFIX:
 			++td->ip;
 		        switch (*td->ip) {
+				case CEE_MONO_RETHROW:
+					CHECK_STACK (td, 1);
+					SIMPLE_OP (td, MINT_MONO_RETHROW);
+					td->sp = td->stack;
+					break;
+
 				case CEE_MONO_LD_DELEGATE_METHOD_PTR:
 					--td->sp;
 					td->ip += 1;
--- a/mono/mini/tramp-amd64.c
+++ b/mono/mini/tramp-amd64.c
@@ -520,9 +520,9 @@ mono_arch_create_generic_trampoline (MonoTrampolineType tramp_type, MonoTrampInf
 	 */
 	if (aot) {
 		/* Not really a jit icall */
-		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "throw_exception_addr");
+		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "rethrow_exception_addr");
 	} else {
-		amd64_mov_reg_imm (code, AMD64_R11, (guint8*)mono_get_throw_exception_addr ());
+		amd64_mov_reg_imm (code, AMD64_R11, (guint8*)mono_get_rethrow_exception_addr ());
 	}
 	amd64_mov_reg_membase (code, AMD64_R11, AMD64_R11, 0, sizeof(gpointer));
 	amd64_mov_reg_reg (code, AMD64_ARG_REG1, AMD64_RAX, sizeof(mgreg_t));
--- a/mono/mini/tramp-amd64.c
+++ b/mono/mini/tramp-amd64.c
@@ -520,9 +520,9 @@ mono_arch_create_generic_trampoline (MonoTrampolineType tramp_type, MonoTrampInf
 	 */
 	if (aot) {
 		/* Not really a jit icall */
-		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "rethrow_exception_addr");
+		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "rethrow_preserve_exception_addr");
 	} else {
-		amd64_mov_reg_imm (code, AMD64_R11, (guint8*)mono_get_rethrow_exception_addr ());
+		amd64_mov_reg_imm (code, AMD64_R11, (guint8*)mono_get_rethrow_preserve_exception_addr ());
 	}
 	amd64_mov_reg_membase (code, AMD64_R11, AMD64_R11, 0, sizeof(gpointer));
 	amd64_mov_reg_reg (code, AMD64_ARG_REG1, AMD64_RAX, sizeof(mgreg_t));
--- a/configure.ac
+++ b/configure.ac
@@ -46,7 +46,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # There is no ordering of corlib versions, no old or new,
 # the runtime expects an exact match.
 #
-MONO_CORLIB_VERSION=BEAF66F4-BA36-44C5-974E-83C6F627DF8C
+MONO_CORLIB_VERSION=E06DCE2D-7852-4BBA-AD9F-54D67EEF1FF9
 
 #
 # Put a quoted #define in config.h.
--- a/mcs/class/referencesource/mscorlib/system/exception.cs
+++ b/mcs/class/referencesource/mscorlib/system/exception.cs
@@ -978,6 +978,8 @@ namespace System {
 
         // Mono addition: Used on iPhone
         IntPtr[] native_trace_ips;
+
+        int caught_in_unmanaged;
 #endif
 
     // See clr\src\vm\excep.h's EXCEPTION_COMPLUS definition:
--- a/mono/cil/cil-opcodes.xml
+++ b/mono/cil/cil-opcodes.xml
@@ -323,4 +323,5 @@
 <opcode name="mono_get_rgctx_arg" input="Pop0" output="PushI" args="InlineNone" o1="0xF0" o2="0x1C" flow="next" />
 <opcode name="mono_ldptr_profiler_allocation_count" input="Pop0" output="PushI" args="InlineNone" o1="0xF0" o2="0x1D" flow="next" />
 <opcode name="mono_ld_delegate_method_ptr" input="Pop1" output="PushI" args="InlineNone" o1="0xF0" o2="0x1E" flow="next" />
+<opcode name="mono_rethrow" input="PopRef" output="Push0" args="InlineNone" o1="0xF0" o2="0x1F" flow="throw" type="Objmodel" />
 </opdesc>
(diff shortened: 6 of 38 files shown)


alexanderkyte:
This got me much better stacktraces locally. I tried to test in a number of different configurations / platforms, but lets see if CI catches anything I didn't.

alexanderkyte:
An audit of my s390x changes by the community would be greatly appreciated. I believe that's @NattyNarwhal right now?

nealef:
That’d be me.

alexanderkyte:
Thank you @nealef

nealef:
What you’ve done is now load R1 with the result of calling mono_get_rethrow_exception_addr:

S390_SET (buf, s390_r1, (guint *)mono_get_rethrow_exception_addr ());

We then go back one stack frame:

s390_aghi (buf, STK_BASE, sizeof(trampStack_t));

Then load the value stored at the address previously loaded:

s390_lg   (buf, s390_r1, 0, s390_r1, 0);

Which we branch to after restoring the preserved registers from the stack

s390_lmg  (buf, s390_r6, s390_r14, STK_BASE, S390_REG_SAVE_OFFSET);
s390_br   (buf, s390_r1);

mono/metadata/marshal-ilgen.c
```diff
@@ -1171,7 +1171,8 @@ emit_thread_interrupt_checkpoint_call (MonoMethodBuilder *mb, gpointer checkpoin
 	/* Throw the exception returned by the checkpoint function, if any */
 	mono_mb_emit_byte (mb, CEE_DUP);
 	pos_noex = mono_mb_emit_branch (mb, CEE_BRFALSE);
-	mono_mb_emit_byte (mb, CEE_THROW);
+	mono_mb_emit_icall (mb, mono_marshal_rethrow_exception);
```
lambdageek: Is it okay to rethrow an exception that was never thrown?  Some of the code paths in `mono_thread_execute_interruption` return brand-new interruptions (e.g. `mono_exception_new_thread_interrupted` in the `thread_interrupt_requested` case).

In other cases `ThreadState_AbortRequested` we re-fly an existing exception (I can't quite tell if the same abort exception gets used multiple times - I think it's possible with some creative use of `ResetAbort`).

mono/metadata/threads.c
```diff
@@ -4992,7 +4992,11 @@ mono_thread_interruption_checkpoint_request (gboolean bypass_abort_protection)
 	if (!bypass_abort_protection && !mono_thread_current ()->pending_exception && is_running_protected_wrapper ())
 		return NULL;
 
-	return mono_thread_execute_interruption_ptr ();
+	MonoException *ptr = mono_thread_execute_interruption_ptr ();
+	if (ptr != NULL)
+		fprintf (stderr, "break %s %d\n", __FILE__, __LINE__);
```
lambdageek: leftover debugging output

mono/metadata/marshal-ilgen.c
alexanderkyte: The `mini` helper in question is used to resume throwing exceptions that were already taken out of the in-flight status. 

This code path seemed very hard to actually hit when testing. The other jit icall I added takes 99.9% of the rethrows. Do you think this is more llvm-specific? Some comments might be in order if that's the case.

mono/metadata/threads.c
alexanderkyte: Oh! Thanks

mono/mini/tramp-amd64.c
```diff
@@ -520,9 +520,9 @@ mono_arch_create_generic_trampoline (MonoTrampolineType tramp_type, MonoTrampInf
 	 */
 	if (aot) {
 		/* Not really a jit icall */
-		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "throw_exception_addr");
+		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, "rethrow_exception_addr");
```
lambdageek: I think you need a matching change in the AOT runtime. 
 Somewhere near https://github.com/mono/mono/blob/a7b9128bf12b50a9d2b10f2de15756311d7b9457/mono/mini/aot-runtime.c#L5310-L5311

alexanderkyte: Interesting. It looks like you're right, but I am surprised this didn't cause any failures on my end when doing a fullaot build.

mono/metadata/marshal-ilgen.c
lambdageek: I don't think it's LLVM specific.  probably since we AOT corlib under normal circumstances you'd only hit this in the early bootstrapping profiles or for the handful of icalls that aren't in corlib.

alexanderkyte:
@nealef in theory if the old icall worked, the new one should as well because they have the same type signature, but it definitely stresses what dependencies the exception throwing mechanisms take.

mono/metadata/marshal-ilgen.c
```diff
@@ -1171,7 +1171,15 @@ emit_thread_interrupt_checkpoint_call (MonoMethodBuilder *mb, gpointer checkpoin
...
-	mono_mb_emit_byte (mb, CEE_THROW);
+
+	// This codepath seems to be very rarely hit when using the non-LLVM
+	// backend. I think that the aot compiler passes on the BCL for the default
+	// profiles probably capture this a bit. 
+	//
+	// See calls to mono_get_rethrow_exception_addr in /mini to see the more
+	// common path
+	mono_mb_emit_icall (mb, mono_marshal_rethrow_exception);
+
```
lewurm: why can't we just `mono_mb_emit_byte (mb, CEE_RETHROW);` instead?

lewurm:
I would love to see a testcase for this

alexanderkyte:
@lewurm in theory every test is a test case for this, because this fixes stacktrace truncation on CI

mono/metadata/marshal-ilgen.c
alexanderkyte: The exception isn't in flight when we're here. Any attempt to hack it would need to ensure that codegen does not require ECMA behavior, but that it can identify when we're in this case. We need a rethrow-like behavior, while putting a brand new exception into the "current thrown exception" state.

alexanderkyte:
I think I could unit test this if I put a throw behind a ton of invokes and made sure the stack trace wasn't being cut off

mono/metadata/marshal-ilgen.c
lewurm: I might missing something here: What does "in flight" mean? 

I don't understand the difference: Calling `mono_marshal_rethrow_exception ()` ends up the same as when `OP_RETHROW` is lowered. See AMD64 for example: https://github.com/mono/mono/blob/1d8954083f7da0d83be04abe201a3e833eb8021f/mono/mini/mini-amd64.c#L5010-L5016

lewurm:
https://github.com/mono/mono/blob/master/mono/tests/exception18.cs does some testing of stacktraces

alexanderkyte:
@monojenkins build Failed

alexanderkyte:
@monojenkins build Failed

alexanderkyte:
@monojenkins build failed

mono/metadata/marshal-ilgen.c
lambdageek: @lewurm I didn't understand either, but I went and looked at 
https://github.com/mono/mono/blob/8fe989cb70a54d4c46f27de2229895fef5db5635/mono/mini/method-to-ir.c#L11712-L11732

The IL `rethrow` opcode (`CEE_RETHROW`) doesn't take an argument.  it rethrows "the current exception" (because it's supposed to only be used in a catch clause).  By contrast the IR rethrow opcode (`OP_RETHROW`) takes an exception as an operand.

Since marshal-ilgen is writing IL, we can't just use `CEE_RETHROW`.

lewurm: ah, that makes sense, thanks @lambdageek. I think we could instead introduce for example `CEE_MONO_RETHROW` (as far as I understand now, that's what @alexanderkyte meant in his previous comment). But at least an explanatory comment in the source code for that would be nice 🙂

alexanderkyte:
I'm seeing an earlier/fixed typo still popping up on the bots. Hmm

alexanderkyte:
Hmm, it seems the output format on fullaot is different. I will have to compare with .net framework and fix that.

alexanderkyte:
On .net framework:

```
https://gist.github.com/alexanderkyte/d685f97bcc332489a23bc9160ba7f4cb
```

On the JIT:

```
https://gist.github.com/alexanderkyte/cd75570e76a0b8183c95eb1b0fe8793c
```

On fullaot:

```
https://gist.github.com/alexanderkyte/4dcb36cbfdf1488a5125da7ea4c40480
```

Hmm, fullaot seems to be working on my mac

alexanderkyte:
The windows+mono stacktrace seems a bit off though. Not sure if that's to be expected. Investigating

mono/metadata/marshal-ilgen.c
```diff
@@ -1193,6 +1201,16 @@ emit_thread_force_interrupt_checkpoint (MonoMethodBuilder *mb)
 	emit_thread_interrupt_checkpoint_call (mb, (gpointer)mono_thread_force_interruption_checkpoint_noraise);
 }
 
+MonoException *
+mono_marshal_rethrow_exception (MonoException *exc)
+{
+	mono_get_eh_callbacks ()->mono_reraise_exception (exc);
+
```
vargaz: I think we should add a new CEE_MONO_RETHROW opcode instead of this. mono_reraise_exception () is a trampoline which throws an exception and it can't be called from native code.

alexanderkyte:
The commit `[runtime] Fix off by one with stacktrace rethrow` may change stacktrace layout in other cases where exceptions are caught. I have compared it against the output from the interpreter/.net framework through VS and it seems to match up.

Let's see if any other runtime tests fail.

alexanderkyte:
Semantically what I'm saying there is "if we are rethrowing an exception, don't exclude the frame that caught/rethrew the exception from the stack". By excluding the last item from the ips array, we were removing the catching frame. 

When an exception is caught, the frame that catches it typically is not included in the stacktrace. This isn't the case for a rethrow though. 

Let me create a small test that shows how we were broken before / to compare with .net framework on windows.

mono/mini/interp/transform.c
```diff
@@ -4321,6 +4321,12 @@ generate (MonoMethod *method, MonoMethodHeader *header, InterpMethod *rtm, unsig
 		case MONO_CUSTOM_PREFIX:
 			++td->ip;
 		        switch (*td->ip) {
+				case CEE_MONO_RETHROW:
+					CHECK_STACK (td, 1);
+					SIMPLE_OP (td, MINT_THROW);
```
luhenry: `MINT_MONO_RETHROW`

alexanderkyte: Hmm, it seemed to be working locally. Let me make sure this is being run during the tests.

alexanderkyte: It was being used but apparently MINT_THROW doesn't break the Invoke stack trace. Not sure if that's an interpreter bug.

alexanderkyte:
Having some trouble reproducing these failures from fresh clones.

alexanderkyte:
@lewurm I'm pretty sure the mono_get_rethrow_exception_addr  stuff is necessary for AOT?

alexanderkyte:

alexanderkyte:
@monojenkins build failed

alexanderkyte:
Fullaot needs a small fix by the look of it. Testing one out now.

alexanderkyte:
Linux WebAssembly is complaining about clang++ being missing.

alexanderkyte:
```sh: 1: clang++: not found
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.MidEast.dll.so' failed
AOT of image class/lib/testing_aot_full/I18N.MidEast.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.MidEast.dll.so] Error 1
make[6]: *** Waiting for unfinished jobs....
sh: 1: clang++: not found
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.Other.dll.so' failed
AOT of image class/lib/testing_aot_full/I18N.Other.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.Other.dll.so] Error 1
sh: 1: clang++: not found
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.West.dll.so' failed
AOT of image class/lib/testing_aot_full/I18N.West.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.West.dll.so] Error 1
sh: 1: clang++: not found
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.dll.so' failed
AOT of image class/lib/testing_aot_full/I18N.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.dll.so] Error 1
sh: 1: clang++: not found
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.CJK.dll.so' failed
AOT of image class/lib/testing_aot_full/I18N.CJK.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.CJK.dll.so] Error 1
sh: 1: build/rules.make:199: recipe for target 'class/lib/testing_aot_full/I18N.Rare.dll.so' failed
clang++: not found
AOT of image class/lib/testing_aot_full/I18N.Rare.dll failed.
make[6]: *** [class/lib/testing_aot_full/I18N.Rare.dll.so] Error 1
sh: 1: clang++: not found
AOT of image class/lib/testing_aot_full/Microsoft.CSharp.dll failed.
make[6]: *** [class/lib/testing_aot_full/Microsoft.CSharp.dll.so] Error 1
build/rules.make:199: recipe for target 'class/lib/testing_aot_full/Microsoft.CSharp.dll.so' failed
sh: 1: clang++: not found``` Bitcode ^

alexanderkyte:
Building the fullaot_llvm preset on softiron-1 now, seems to be proceeding without problem.

alexanderkyte:
I'm able to build this on fullaot+llvm on softiron-1 and it passes the corlib test suite.

PR #10477, opened 2018-09-05
[corlib] Don't decrement ConditionalWeakTable::size on remove.

kumpera:
We can't decrement the size on removal as we use tombstones.
When using tombstones, the invariant to keep is on the number of null entries as those
are used as search stop-marks.

Without this change, it's possible that a long series of add/remove operations could leave the array with no null entries,
just tombstones and filled entries. In that case, search would degenerate into a linear search that would look into all array elements.

This change fixes worst-case perf of not-found searches to be bound to the load factor.

Additionally, clear the array to null entries, otherwise we fall into the above trap even more easily.


marek-safar:
@monojenkins build failed

marek-safar:
@monojenkins backport 2018-08

PR #10480, opened 2018-09-05
[coop] Make mono_jit_thread_attach external only and switch to GC Safe

lambdageek:
`mono_jit_thread_attach` had two uses:
* in preemptive suspend it was used for in all native-to-managed wrappers to attach a thread to the runtime and to switch the active domain;
* it is also used by embedders to attach foreign threads to the runtime.  There is no corresponding detach operation. For example, Xamarin.Mac uses this function in this way.

With hybrid suspend the first use isn't needed (hybrid and coop suspend use `mono_threads_attach_coop`/`mono_threads_detach_coop` in the n2m wrappers).

The second use, however is still present.  However in this case we actually want to attach the thread in BLOCKING (GC Safe) mode since we know the thread is foreign and isn't running any managed code yet.  When it calls a Mono API function or when it calls a managed delegate through a trampoline, we'll switch to GC Unsafe (and then switch back to GC Safe when the managed or mono API function returns).

One place where this is useful is when Xamarin.Mac users call managed delegates from Grand Central Dispatch worker threads - after the task runs, we need to leave the idle GCD worker in GC Safe so that if Mono performs a GC, we don't try to wait for the worker to cooperatively suspend itself.

Also bump the AOT file format so we don't have any stale AOT references to `mono_jit_thread_attach` from AOTd managed code.

**NOTE**: PR against 2018-08.  Sorry. I'll do a matching `master` PR soon.

### DIFF
--- a/mono/metadata/threads-types.h
+++ b/mono/metadata/threads-types.h
@@ -455,7 +455,7 @@ MONO_API void
 mono_threads_detach_coop (gpointer cookie, gpointer *dummy);
 
 MonoDomain*
-mono_threads_attach_coop_internal (MonoDomain *domain, gpointer *cookie, MonoStackData *stackdata);
+mono_threads_attach_coop_internal (MonoDomain *domain, gpointer *cookie, MonoStackData *stackdata, gboolean attach_external_gc_safe);
 
 void
 mono_threads_detach_coop_internal (MonoDomain *orig_domain, gpointer cookie, MonoStackData *stackdata);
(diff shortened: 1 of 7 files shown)


lambdageek:
@luhenry I agree.  Do you think getting rid of the JIT_ATTACH opcodes is still possible for 2018-08, or should that be done on master only?

lambdageek:
Will try the more ambitious approach in another PR.

PR #10485, opened 2018-09-06
Fix forceBuild not working for projects with unknown platform names (i.e. jay)

kg:
This broke at some point and I didn't notice because it's a behavior divergence between msbuild and devenv (I was busy fixing msbuild-only bugs). As a result of some old broken logic in genproj (my fault probably but I forget), jay.vcxproj was not being set to build and VS would happily skip it even though some .csproj files depended on it, and as a result VS tried to build them and failed.

This should fix it but I'm submitting the PR since make -j + update-solution-files takes an eternity on my machine.

### DIFF
--- a/msvc/scripts/genproj.cs
+++ b/msvc/scripts/genproj.cs
@@ -141,11 +141,11 @@ public class SlnGenerator {
 		var fallbackProfileNames = new List<string> ();
 
 		foreach (var profile in profiles) {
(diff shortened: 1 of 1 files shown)


kg:
@monojenkins build Linux x64

kg:
@monojenkins build failed

luhenry:

Issue #10495, opened 2018-09-06
Recent regression: ```ProvidedField has invalid vtable method slot 20 with method none``` 


dsyme:
We're seeing a failure in Mono 5.19.0.290 (mono-weekly tag on travis) that was passing in all previous Mono. It looks like a regression has slipped in 

I'm seeing this in CI for project https://github.com/fsprojects/FSharp.TypeProviders.SDK/ 

![image](https://user-images.githubusercontent.com/7204669/45174585-2b8b5e00-b203-11e8-95ce-1325588b5584.png)

We CI that repo using mono-weekly to make sure we see mono regressions before they are too late.

```ProvidedField has invalid vtable method slot 20 with method none``` 

## Steps to Reproduce

1. Clone https://github.com/fsprojects/FSharp.TypeProviders.SDK/
2. set USE_MSBUILD=1  (not sure if it's relevant)
3.  `./build.sh Pack`

### Current Behavior

```

/home/travis/build/fsprojects/FSharp.TypeProviders.SDK/examples/BasicProvider.Tests/BasicProvider.Tests.fs(12,39): error FS3033: The type provider 'ProviderImplementation.BasicErasingProvider' reported an error: Type ProviderImplementation.ProvidedTypes.ProvidedField has invalid vtable method slot 20 with method none [/home/travis/build/fsprojects/FSharp.TypeProviders.SDK/examples/BasicProvider.Tests/BasicProvider.Tests.fsproj]
```

### Expected Behavior

Tests pass

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

See above  - 5.19.0.290

lambdageek:
@MaximLipnin `ProvidedField` in the error message here is a custom `FieldInfo` subclass.  Could this be related to your recent CoreFX System.Reflection work?

lambdageek:
So my guess is that this is a regression since #10370 related to https://github.com/mono/mono/pull/10370/files#diff-d734a56e2e46fe93fa6696019ba1a9b0R62
That line adds a new abstract `GetFieldOffset` method to `FieldInfo` which is not present in .NET Standard.

@MaximLipnin that makes it impossible for custom FieldInfo subclasses to be created by downstream projects.

In this case it's a runtime error, rather than compile time error because this project builds against .NET Standard (which doesn't have the `GetFieldOffset` internal method), but then runs against Mono.

MaximLipnin:
[...]

PR #10497, opened 2018-09-06
[System.Reflection] Make GetFieldOffset method virtual to fix issue with custom FieldInfo type.

MaximLipnin:
Fixes https://github.com/mono/mono/issues/10495

### DIFF
--- a/mcs/class/corlib/System.Reflection.Emit/FieldOnTypeBuilderInst.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/FieldOnTypeBuilderInst.cs
@@ -128,11 +128,6 @@ namespace System.Reflection.Emit
 			throw new NotSupportedException ();
 		}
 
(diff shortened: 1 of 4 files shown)


dsyme:
@MaximLipnin I tried 5.19.0.290 and still see a failure - should that version have had the fix?  Thanks!

MaximLipnin:
@dsyme It looks like that the tarball used in the build is a bit old and doesn't include the fix that was merged on September, 7th.
```
$ mono --version
Mono JIT compiler version 5.19.0.290 (tarball Tue Sep  4 11:00:10 UTC 2018)
```
Maybe @directhex has any track of commit/version for tarballs.

directhex:
Mapping at https://jenkins.mono-project.com/view/Releases/job/build-package-osx-mono/job/master/

Looks like a misconfiguration on some new hardware was causing nightly packages to fail. Resolved that.

Issue #10441, opened 2018-09-02
FullAOT LLVM  Linux x64 one flaky failure.


jaykrell:
FullAOT LLVM  Linux x64 has one flaky failure.

Here is how far I've gotten it, reduced:

```
using System;
using System.Linq;
using System.Collections.Generic;

internal static class AsParallelHelper
{
	internal static ParallelQuery<T> AsReallyParallel<T> (this IEnumerable<T> source)
	{
		return source.AsParallel ().WithExecutionMode (ParallelExecutionMode.ForceParallelism);
	}
}

public class ParallelEnumerableTests
{
	public static void Main(string[] args)
	{
		int num = 100;
		Tuple<int, int>[] source = Enumerable.Range (0, num).Select ((i) => Tuple.Create (i / 10, i)).ToArray ();

		var actual = source.AsReallyParallel ().GroupBy ((e) => e.Item1);
		foreach (var group in actual) { }
	}
}
```

jaykrell:
Here is the relevant stack, which is elusive -- you have to add excess fields to System.Linq.Parallel.Wrapper to see it.

```
  at System.Buffer.Memcpy (System.Byte* dest, System.Byte* src, System.Int32 len) <0x7f0e247b87c0 + 0x00040> in <b0ac880de1b74e779047cdc849d8159b#9863251a1fd577ad9f1585c3dc141e1a>:0 
  at System.String.memcpy (System.Byte* dest, System.Byte* src, System.Int32 size) <0x7f0e247904c0 + 0x0002c> in <b0ac880de1b74e779047cdc849d8159b#9863251a1fd577ad9f1585c3dc141e1a>:0 
  at System.Linq.Parallel.WrapperEqualityComparer`1[T].GetHashCode (System.Linq.Parallel.Wrapper`1[T] x) [0x00000] in <b4141d4f0046412bbd28c6d2dc4c5d53>:0 
  at (wrapper unknown) System.Object.gsharedvt_out()
  at System.Linq.Parallel.HashLookup`2[TKey,TValue].GetKeyHashCode (TKey key) [0x00000] in <b4141d4f0046412bbd28c6d2dc4c5d53>:0 
  at (wrapper unknown) System.Object.gsharedvt_in()
  at (wrapper unknown) System.Object.gsharedvt_out()
[...]

Issue #9046, opened 2018-06-07
ParallelEnumerableTests.TestGroupBy flaky test


jaykrell:
## Steps to Reproduce

1.  Submit a PR.

### Current Behavior

e.g.

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot+llvm/1438/#showFailuresLink

ParallelEnumerableTests.TestGroupBy

```
Regression
ParallelEnumerableTests.TestGroupBy

Failing for the past 1 build (Since Unstable#1438 )
Took 5 sec.
Stacktrace
                                                MESSAGE:
                                                System.AggregateException : One or more errors occurred.
  ----> System.NullReferenceException : Object reference not set to an instance of an object
                                                +++++++++++++++++++
                                                STACK TRACE:
                                                  at System.Linq.Parallel.QueryTaskGroupState.QueryEnd (System.Boolean userInitiatedDispose) [0x00081] in <d07252c7308d4678b2bfc6a9a28e5c4c>:0 
  at System.Linq.Parallel.AsynchronousChannelMergeEnumerator`1[T].MoveNextSlowPath () <0x7f74638a1ff0 + 0x0049a> in <d07252c7308d4678b2bfc6a9a28e5c4c#17884142f13783db20e8a3d8a2ed0edc>:0 
  at System.Linq.Parallel.AsynchronousChannelMergeEnumerator`1[T].MoveNext () <0x7f74638a1f40 + 0x00044> in <d07252c7308d4678b2bfc6a9a28e5c4c#17884142f13783db20e8a3d8a2ed0edc>:0 
  at System.Linq.Parallel.QueryOpeningEnumerator`1[TOutput].MoveNext () <0x7f74638ae7c0 + 0x00039> in <d07252c7308d4678b2bfc6a9a28e5c4c#17884142f13783db20e8a3d8a2ed0edc>:0 
  at MonoTests.System.Linq.ParallelEnumerableTests+<>c__DisplayClass44_0.<TestGroupBy>b__1 () [0x000e8] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-fullaot+llvm/mcs/class/System.Core/Test/System.Linq/ParallelEnumerableTests.cs:606 
  at MonoTests.System.Linq.ParallelEnumerableTests.TestGroupBy () <0x7f746875dfc0 + 0x0014b> in <0dc1ddc4fbf9442a824faf89ee7def8b#17884142f13783db20e8a3d8a2ed0edc>:0 
[...]

PR #10502, opened 2018-09-07
Fix MonoTests.System.Linq.ParallelEnumerableTests.TestGroupBy frequently crashing on FullAOT/LLVM/Linux/amd64 CI.

jaykrell:
https://github.com/mono/mono/issues/10441
https://github.com/mono/mono/issues/9554
https://github.com/mono/mono/issues/9046

See https://github.com/mono/mono/commit/8dd12eab7fa2f2fac79cad5c751be7c2795683d4.

Note that I don't have a detailed analysis of the codegen before/after bad/good.
This is based on a lot of guessing and experimenting, and the experimenting is conclusive.
I had a greater than 10% failure rate without this, and 100% passing with.
It could be this covers up something else.
gsharedvt is obviously involved with the repro.
I don't have an explanation for the flakiness, or the Linux-specificity.
But the repro rates were high without this and zero with this, so this fixes it.
We can also bear this out in time with CI.
I looked into preserving more registers, i.e. in the nearby safepoint callout.
That didn't seem to work, but if it did, would explain flakiness.
 Oh, I also hacked JIT to not emit safe points and that didn't help, so that should rule that out.
Removing parallel from the test also made it not repro.

The way gsharedvt works, I doubt anyone will mind this pessimization.
In the test, we call memcpy frequently to copy 4 bytes. Actual calls. For a constant 4 bytes. With LLVM. To compute the hash of an int. You can see it in some of the repros.

It is still reasonable to suspect this is a problem of register preservation.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -6956,11 +6956,18 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 		/* Allocate locals */
 		locals_var = mono_compile_create_var (cfg, mono_get_int_type (), OP_LOCAL);
 		/* prevent it from being register allocated */
-		locals_var->flags |= MONO_INST_VOLATILE;
+		//locals_var->flags |= MONO_INST_VOLATILE;
 		cfg->gsharedvt_locals_var = locals_var;
 
 		dreg = alloc_ireg (cfg);
 		MONO_EMIT_NEW_LOAD_MEMBASE_OP (cfg, OP_LOADI4_MEMBASE, dreg, var->dreg, MONO_STRUCT_OFFSET (MonoGSharedVtMethodRuntimeInfo, locals_size));
+
+		// FIXME: Increase locals_size by 1, to probably cover up some miscalculation elsewhere,
+		// that causes locals to be corrupt. Inconsistent test case is
+		// MonoTests.System.Linq.ParallelEnumerableTests.TestGroupBy on Linux/amd64/FullAOT/LLVM.
+		// Adding logging to gsharedvt memcpy calls should uncover it. Obvious candidate of
+		// cleaning up instantiate_info failed.
+		MONO_EMIT_NEW_BIALU_IMM (cfg, OP_ADD_IMM, dreg, dreg, 1);
 
 		MONO_INST_NEW (cfg, ins, OP_LOCALLOC);
 		ins->dreg = locals_var->dreg;
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -6956,7 +6956,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 		/* Allocate locals */
 		locals_var = mono_compile_create_var (cfg, mono_get_int_type (), OP_LOCAL);
 		/* prevent it from being register allocated */
-		//locals_var->flags |= MONO_INST_VOLATILE;
+		locals_var->flags |= MONO_INST_VOLATILE;
 		cfg->gsharedvt_locals_var = locals_var;
 
 		dreg = alloc_ireg (cfg);


jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build  failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build  failed

kumpera:
gsharedvt performance matters as customers do hit it in practice.
Without a clear understanding of what this changes does and what's the perf impact, it's hard to review & approve it.

jaykrell:
Maybe avoid enregistration on any address taken, i.e. any that are passed as parameters, i.e. still enregister in leaves.
The original intent of the “indirect” flag, that got changed to volatile?

kumpera:
Please measure the perf impact and check how codegen changes.

I understand that the empirical result of better stability are promising, but we must have some understanding of why. Otherwise we might just have disguised the bug behind slower code without actually fixing it.

kumpera:
git grep is your friend for this sort of questions.
Using MONO_INST_INDIRECT would be wrong here because at no place we take the address of `locals_var`.

jaykrell:
Git grep understood.
Ah--subtle. This thing, I  assume, need to read more, is all the gsharedvt locals combined. So they are all address-taken, but the pointer to them is not, understood. The pointer to them should not need to be volatile either. Maybe something in its use can be pessimized less. Let me read more..

vargaz:
I don't think this will actually fix the problem, it will probably just make it happen less often.

jaykrell:
I don't know. It is certainly a lot less often.
Maybe worth an experiment to have two variables, one volatile, one not, and insert checks at each use that they are equal? Find if someone is trashing the register? It always see it is correctly put in a nonvolatile register.

The codegen difference is about as expected -- the variable does go from enregistered to stacked, and there are a small number of additional uses.

jaykrell:
So, just allocating an extra 16 bytes makes the repro go away too. Ugh. Volatile is ok.
Nonvolatile might have been winning merely by virtue of adding 8 bytes.

jaykrell:
Ok, this is still not my proudest moment, and it is still a little hacky, but it makes more sense, is less pessimization, and retains the empirical goodness of the other fix.

Increase locals size by one.

I think with a bit more digging I could find the overwrite and/or fix the  size calculation.
I poked around where I thought relevant instantiate_info, but failed.

vargaz:
Shouldn't MonoGSharedVtMethodRuntimeInfo->locals_size be increased instead ?

jaykrell:
I tried increasings its computation and that didn't seem to work, oddly. I can look more. It seems it is in instantiate_info. I can try again, later. I realize  the +1  is really going to be +4 or +8 or more, due to alignment.

jaykrell:
Maybe a race in `mono_class_fill_runtime_generic_context`, but have to not make it deadlock either.

jaykrell:
Since this is difficult to debug, and any apparent fix has been large and nonsensical, and there is a reliable enough repro, I went in another direction  -- twiddling the `disable_llvm` flag.

With *a lot* of binary search and experimentation I have found that compiling `System_Security_Cryptography_ECCurve_get_Oid` with mini  instead of mono, just this one function, moves from non-working to working. I have run the experiment both with all functions as LLVM and toggling this one, or no functions as LLVM and toggling this. I have also toggled other functions in groups or one at a time (i.e. binary search). It is just this function.

The repro takes under a second. Failure rate is >10% when it fails, so the results are pretty conclusive. Iteration time is slow as I have to  re-AOT corlib, System, System.Core, Mono.Security.

This makes no sense. It does not even run.

My only ideas are maybe this function is related by having same unwind info -- but the repro does not  hit mono_error_prepare_exception, or that compiling it triggers some memory corruption and incorrect code.

Similar theory on the other fixes -- they don't really make sense, but cause some perturbation of the code.

I'd like to dig further as to the difference in the binaries, and maybe reduce the difference further.

The results here are both very consistent and nonsensical at the same time.

jaykrell:
More data.
1.  Repro rate on physical 4 core Linux laptop is much lower. Like 5% or less (1 in 20 or less). But still repros.
2. I was able to get it under gdb+ipt and step backwards some. But this reverse debugging does not show memory or registers.

Independent of that, it seems like a gsharedvt_trampoline is being skipped.
Usually there is an out/in pair, and a struct goes from ref to value and back to ref.
But the repros all show only trampoline_out on the stack, and printf debugging shows the value to ref marshal not occuring. This, if true, explains it. But then how does this occur?

There is a  bunch of code I do not understand:
`lookup_or_register_info`, `mono_method_lookup_or_register_info`, `alloc_rgctx_array`, the loop in `fill_runtime_generic_context`

Is it racy? Can we see uninitialized (zeroed) data, or data from another call?

jaykrell:
It would be interesting therefore to run it on MacOS in a VM to see if repro rate comes up.
To take away the Linux-specificity of it.
(VMware allows Mac-on-Mac VMs.)

jaykrell:
This now appears to be problems/raceconditions in `mini_jit_info_table_find`.

Sometimes it fails to find:

`System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT`

therefore  we skip putting a gsharedvt trampoline around it.

jaykrell:
Replaced by https://github.com/mono/mono/pull/10615.

PR #10509, opened 2018-09-07
[WIP] Detect endianess in runtime (BitConverter.IsLittleEndian)

EgorBo:
This PR also expects https://github.com/mono/corefx/blob/master/src/Common/src/CoreLib/System/BitConverter.cs#L25 to be unassigned (to avoid a redundant static cctor).

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -188,6 +188,7 @@ create_domain_objects (MonoDomain *domain)
 	MonoStringHandle arg;
 	MonoVTable *string_vt;
 	MonoClassField *string_empty_fld;
+	MonoClassField *bitconverter_isle_fld;
 
 	if (domain != old_domain) {
 		mono_thread_push_appdomain_ref (domain);
@@ -208,6 +209,14 @@ create_domain_objects (MonoDomain *domain)
 	mono_error_assert_ok (error);
 	mono_field_static_set_value (string_vt, string_empty_fld, empty_str);
 	domain->empty_string = empty_str;
+
+	bitconverter_isle_fld = mono_class_get_field_from_name_full (mono_defaults.bitconverter_class, "IsLittleEndian", NULL);
+	if (bitconverter_isle_fld) {
+		gint x = 1;
+		MonoBoolean is_le = *((char*) &x); // 00 00 00 01 for little-endian and 01 00 00 00 for big-endian
+		MonoVTable *bitconverter_vt = mono_class_vtable_checked (domain, mono_defaults.bitconverter_class, error);
+		mono_field_static_set_value (bitconverter_vt, bitconverter_isle_fld, &is_le);
+	}
 
 	/*
 	 * Create an instance early since we can't do it when there is no memory.
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -941,6 +941,7 @@ typedef struct {
 	MonoClass *threadpool_wait_callback_class;
 	MonoMethod *threadpool_perform_wait_callback_method;
 	MonoClass *console_class;
+	MonoClass *bitconverter_class;
 } MonoDefaults;
 
 #ifdef DISABLE_REMOTING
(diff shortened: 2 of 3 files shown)


mono/metadata/appdomain.c
```diff
@@ -209,6 +210,14 @@ create_domain_objects (MonoDomain *domain)
 	mono_field_static_set_value (string_vt, string_empty_fld, empty_str);
 	domain->empty_string = empty_str;
 
+	bitconverter_isle_fld = mono_class_get_field_from_name_full (mono_defaults.bitconverter_class, "IsLittleEndian", NULL);
+	if (bitconverter_isle_fld) {
+		gint x = 1;
+		MonoBoolean is_le = *((char*) &x); // 00 00 00 01 for little-endian and 01 00 00 00 for big-endian
```
kumpera: Just use the TRUE/FALSE constants.

mono/metadata/class-internals.h
```diff
@@ -941,6 +941,7 @@ typedef struct {
 	MonoClass *threadpool_wait_callback_class;
 	MonoMethod *threadpool_perform_wait_callback_method;
 	MonoClass *console_class;
+	MonoClass *bitconverter_class;
```
kumpera: We shouldn't add classes here unless they are accessed from multiple places during execution.

mono/metadata/appdomain.c
jaykrell: x should be const/static. Or use G_ENDIAN.

mono/metadata/appdomain.c
```diff
@@ -210,12 +209,16 @@ create_domain_objects (MonoDomain *domain)
...
-		MonoBoolean is_le = *((char*) &x); // 00 00 00 01 for little-endian and 01 00 00 00 for big-endian
-		MonoVTable *bitconverter_vt = mono_class_vtable_checked (domain, mono_defaults.bitconverter_class, error);
-		mono_field_static_set_value (bitconverter_vt, bitconverter_isle_fld, &is_le);
+	/* Initialize BitConverter.IsLittleEndian */
+	MonoClass *bitconverter_class = mono_class_try_load_from_name (mono_defaults.corlib, "System", "BitConverter")
+	if (bitconverter_class) {
+		MonoClassField *bitconverter_isle_fld = mono_class_get_field_from_name_full (bitconverter_class, "IsLittleEndian", NULL);
+		if (bitconverter_isle_fld) {
+			gint x = 1;
+			MonoBoolean is_le = G_BYTE_ORDER == G_BIG_ENDIAN ? FALSE : TRUE;
```
jaykrell: `const static MonoBoolean is_le = (G_BYTE_ORDER != G_BIG_ENDIAN);`
And fix `mono_copy_value` and `mono_field_static_set_value` to take `const void*`  or  `gconstpointer`.

vargaz:
It would be better to do this in such a way that its only done when the bitconverter class is actually used.

vargaz:
What I mean is that this change means one more class is loaded at startup, which slows down startup, even if the app never uses the bitconverter class.

jaykrell:
Can you do more like what I said?
 - Leave the C# alone. Let the cctor exist  and run. On all platforms.
 - Intrinsify the field load, similar to how intrinsify some functions.

The we way intrinsify in general, I assume  does not cause classes to load. It should be pretty easy.

As well, didn't we find, that intrinsification isn't needed for the optimization anyway, that code JITed after the cctor, gets the constant? At least sometimes?

vargaz:
First the proper value needs to be computed somewhere, then the JIT/interpreter can optimize it away.

vargaz:
https://github.com/mono/corefx/pull/134

vargaz:
https://github.com/mono/mono/pull/10531

jaykrell:
There is some question as to if #10531 is needed, but yes that's what I meant.
i.e. some experiments show cctor written values get propagated w/o special casing, but not always.

PR #10513, opened 2018-09-07
Bump the corefx submodule to latest master

joshpeterson:
I'm doing this to pick up changes from https://github.com/mono/corefx/pull/132


joshpeterson:
I'm unsure why 6 checks failed. Is there anything else I need to do to have this merged?

marek-safar:
Another commit with the bump to same revision was merged instead.

PR #10515, opened 2018-09-07
[cxx] Jenkins support for Windows C++.

jaykrell:


scripts/ci/run-jenkins.sh
```diff
@@ -178,11 +179,11 @@ fi
 if [[ ${CI_TAGS} == *'win-i386'* ]];
     then
 	# only build boehm on w32 (only windows platform supporting boehm).
-    ${TESTCMD} --label=make-msvc --timeout=60m --fatal /cygdrive/c/Program\ Files\ \(x86\)/MSBuild/14.0/Bin/MSBuild.exe /p:PlatformToolset=v140 /p:Platform=${PLATFORM} /p:Configuration=Release /p:MONO_TARGET_GC=boehm msvc/mono.sln
+     ${TESTCMD} --label=make-msvc --timeout=60m --fatal /cygdrive/c/Program\ Files\ \(x86\)/MSBuild/14.0/Bin/MSBuild.exe /p:PlatformToolset=v140 /p:Platform=${PLATFORM} /p:Configuration=Release ${MSBUILD_CXX} /p:MONO_TARGET_GC=boehm msvc/mono.sln
```
akoeplinger: extra space

jaykrell:
@lambdageek That should be addressed by https://github.com/mono/mono/pull/10520.

lateralusX:
Should we switch just to the C++ builds now? Looks like this change will only build one version on CI and run all tests using that version. I believed that we intended to just do the build for now to catch errors related to the port and make the switch later? What is the strategy on other platforms related to the switch to C++?

jaykrell:
There are many viable choices, but for now one configuration per CI run is ok.
I realize I pushed for the separate directories. And it still isn't a bad idea.
The Windows build works better than Unix/automake in this regard.
Unix/automake don't have per-configuration output directories  and there is this unfortunate thing, that being better (Windows) isn't super worthwhile when everything else is worse.

lateralusX:
So just to be clear, we will still keep the C compiler as default build on CI for now, correct? Do we plan to setup specific lanes for the C++ build on Windows CI as well or will this only be run on specific CI requests?

jaykrell:
C is default for now. I'm not sure how our C++ testing will proceed but I think by request, like "monojenkins build Linux x64 C++" and then maybe later separate lanes.

lambdageek:
I actually changed the Linux C++ lane to run on all PRs now (but it's not required) over the weekend.

But in any case, C is the default.  C++ is just to prevent regressions while we figure the next steps.

PR #10516, opened 2018-09-07
DO NOT MERGE explicitly bad C++ code

lambdageek:
Testing that the C++ build fails on bad C++

@monojenkins skip

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -55,6 +55,23 @@
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
 
+#ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


lambdageek:

mono/metadata/object.c
```diff
@@ -55,6 +55,23 @@
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
 
+#ifdef __cplusplus
+class Testing
```
jaykrell: I would have just said `#error` here.

lambdageek:
Ok, good. It failed in the correct manner.

PR #10537, opened 2018-09-10
[System.Reflection] CoreFX import for PropertyInfo type.

MaximLipnin:
Part of #9660.

The changes:
- included corefx version of PropertyInfo type;
- included related xunit tests from corefx System.Reflection;
- moved `GetPropertyFromHandle `method to MonoProperty type;
- moved `internal_from_handle_type `extern method and related icall to MonoProperty type;
- fixed some edge cases where exception is expected;
- removed obsolete System.Reflection/PropertyInfo.cs.


MaximLipnin:

mcs/class/corlib/System.Reflection/MonoMethod.cs
```diff
@@ -296,6 +296,15 @@ internal override int GetParametersCount ()
 		[DebuggerStepThrough]
 		public override Object Invoke (Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) 
 		{
+			if (!IsStatic) {
+				if (!DeclaringType.IsInstanceOfType (obj)) {
```
EgorBo: I'd rewrite these conditions to 
```csharp
if (!IsStatic) {
    if (obj == null) 
        throw new TargetException ("Non-static method requires a target.");
    if (!DeclaringType.IsInstanceOfType (obj))
        throw new TargetException ("Object does not match target type.");
}
```
so !IsStatic and obj == null won't need to call that icall `IsInstanceOfType `

kumpera: That's a pointless optimization as the success path will have the same amount of checks.

marek-safar:
@monojenkins build failed

marek-safar:

MaximLipnin:
Thanks for review!

PR #10542, opened 2018-09-10
[exception handling] do not clear LMF when stack pointer equals to LMF address

lewurm:
Otherwise we could clear a LMF entry from an upper frame when `new_ctx`
is from a frame of an optimized function that doesn't do any stack
allocation.

Fixes `make -C mono/mini mixedcheck` with `CFLAGS=-O2` on macOS (= clang).


lewurm:
let's see if it breaks any other configuration.

BrzVlad:
@lewurm I think we shouldn't inline methods that push lmf. Is MONO_NEVER_INLINE not working ?

PR #10545, opened 2018-09-11
Drop JIT attach: use coop attach in mini, aot and interp

lambdageek:
1. Always use `mono_threads_attach_coop`/`mono_threads_detach_coop` instead of `mono_jit_thread_attach`/`mono_domain_set`
2. Drop the mono JIT_ATTACH opcode
3. Drop the interp JIT_ATTACH opcodes; add a coop attach opcode.  (This is necessary because the usual interp icall machinery expects the TLS values to already be initialized - precisely the thing that `mono_threads_attach_coop` is supposed to do).
4. Mark `mono_jit_thread_attach` external only.
   This is still used by Xamarin.Mac and perhaps other embedders.  However now that it is not called inside mono, we can always attach threads in GC Safe mode - if they call a `MONO_API` function or a managed delegate we'll switch to GC Unsafe at that entry point.
   This fixes hybrid suspend deadlocks when using Xamarin.Mac - when running managed tasks on Grand Central Dispatch, we need to leave the thread pool workers in GC Safe mode when they're idle.
5. Drop `MONO_PATCH_INFO_JIT_THREAD_ATTACH`; bump AOT file format.


mono/mini/method-to-ir.c
```diff
@@ -11196,80 +11196,6 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			MONO_ADD_INS (cfg->cbb, ins);
 			break;
 		}
-		case MONO_CEE_MONO_JIT_ATTACH: {
-			g_assert (method->wrapper_type != MONO_WRAPPER_NONE);
```
vargaz: This was optimized code which avoided calling into the runtime if not needed.
Also, it handled the case when calling into the runtime would go though the trampoline code which expected the runtime to be already attached.

luhenry: This is not compatible with cooperative and hybrid suspend which are becoming defaults. And since we want to generate coop-aware code even when the runtime is not configured for cooperative/hybrid (in the AOT case for example), we need this change. Also, the existing `mono_threads_attach_coop` are compatible with the preemptive case as well.

mono/metadata/marshal-ilgen.c
```diff
@@ -5879,10 +5879,7 @@ emit_managed_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethodSignature *invoke_s
 	mono_mb_emit_byte (mb, CEE_CONV_U4);
 	mono_mb_emit_stloc (mb, ex_local);
 
-	if (!mono_threads_is_blocking_transition_enabled ()) {
-		mono_mb_emit_byte (mb, MONO_CUSTOM_PREFIX);
-		mono_mb_emit_byte (mb, CEE_MONO_JIT_ATTACH);
-	} else {
+	{
```
luhenry: Let's just fix the indentation.

mono/metadata/marshal-ilgen.c
```diff
@@ -6036,10 +6033,7 @@ emit_managed_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethodSignature *invoke_s
 		}
 	}
 
-	if (!mono_threads_is_blocking_transition_enabled ()) {
-		mono_mb_emit_byte (mb, MONO_CUSTOM_PREFIX);
-		mono_mb_emit_byte (mb, CEE_MONO_JIT_DETACH);
-	} else {
+	{
```
luhenry: Let's just fix the indentation.

mono/mini/interp/interp.c
```diff
@@ -5088,28 +5088,12 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, guint16 *st
 			mono_memory_barrier ();
 			MINT_IN_BREAK;
 		}
-		MINT_IN_CASE(MINT_MONO_JIT_ATTACH) {
+		MINT_IN_CASE(MINT_MONO_THREADS_ATTACH_COOP) {
```
luhenry: AFAIU it doesn't seem we need this opcode: since this opcode is generated when calling `mono_threads_attach_coop ` and all this opcode does is calling `mono_threads_attach_coop `. Why can't we let the icall go through directly through the existing icall calling code?

lambdageek: The existing icall calling code - in `do_icall` expects the thread to already be attached (specifically it wants to manipulate the LMF of the current thread).  I tried it as a regular icall and got crashes.  This was simpler.

luhenry: Sounds good 👍

lambdageek:

mono/mini/mini-runtime.c
```diff
@@ -996,6 +1009,7 @@ mono_thread_attach_cb (intptr_t tid, gpointer stack_start)
 	thread = mono_thread_info_current_unchecked ();
 	if (thread)
 		thread->jit_data = jit_tls;
+	mini_get_interp_callbacks ()->thread_attach_cb ();
```
lewurm: why is that needed? we usually lazily initialize `ThreadContext` in the interpreter.

lambdageek: Not sure. I could try without.  I'm trying to follow what `MINT_MONO_JIT_ATTACH` used to do - which is a call to `set_context` after attaching to set `jit_tls->interp_context`.  Couldn't really tell from the code if that would always happen on its own otherwise.

lambdageek: If I stub out the callback, `bug-58782-capture-and-throw.exe` fails with

```
* Assertion at interp/interp.c:5693, condition `context' not met
* thread #4, name = 'tid_1107', stop reason = signal SIGABRT
  * frame #0: 0x00007fff6be4ab66 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6c015080 libsystem_pthread.dylib`pthread_kill + 333
    frame #2: 0x00007fff6bda61ae libsystem_c.dylib`abort + 127
    frame #3: 0x00000001002eeb40 mono`mono_log_write_logfile(log_domain=<unavailable>, level=<unavailable>, hdr=<unavailable>, message=<unavailable>) at mono-log-common.c:135 [opt]
    frame #4: 0x000000010030bf22 mono`monoeg_assertion_message [inlined] monoeg_g_logstr(log_domain=<unavailable>, log_level=G_LOG_LEVEL_ERROR, msg=<unavailable>) at goutput.c:117 [opt]
    frame #5: 0x000000010030bef8 mono`monoeg_assertion_message [inlined] monoeg_g_logv_nofree(log_domain=<unavailable>, log_level=G_LOG_LEVEL_ERROR, format=<unavailable>, args=0x0000003000000020) at goutput.c:128 [opt]
    frame #6: 0x000000010030bee4 mono`monoeg_assertion_message(format=<unavailable>) at goutput.c:163 [opt]
    frame #7: 0x0000000100111c85 mono`interp_set_resume_state(jit_tls=<unavailable>, ex=<unavailable>, ei=<unavailable>, interp_frame=<unavailable>, handler_ip=<unavailable>) at interp.c:5693 [opt]
    frame #8: 0x00000001000b400b mono`mono_handle_exception_internal(ctx=<unavailable>, obj=0x0000000102802030, resume=<unavailable>, out_ji=<unavailable>) at mini-exceptions.c:2568 [opt]
    frame #9: 0x00000001000b3359 mono`mono_handle_exception(ctx=<unavailable>, void_obj=<unavailable>) at mini-exceptions.c:2720 [opt]
    frame #10: 0x0000000100121d3a mono`interp_throw(context=0x00000001010015c0, ex=0x0000000102802030, frame=0x000070000eed5758, ip=0x0000000101814194, rethrow=0) at interp.c:734 [opt]
    frame #11: 0x0000000100117975 mono`interp_exec_method_full(frame=<unavailable>, context=0x00000001010015c0, start_with_ip=<unavailable>, filter_exception=0x0000000102802030, exit_at_finally=-1, base_frame=0x0000000000000000) at interp.c:4084 [opt]
    frame #12: 0x0000000100113bd8 mono`interp_exec_method_full(frame=<unavailable>, context=0x00000001010015c0, start_with_ip=<unavailable>, filter_exception=0x0000000000000000, exit_at_finally=-1, base_frame=0x0000000000000000) at interp.c:3012 [opt]
    frame #13: 0x0000000100113bd8 mono`interp_exec_method_full(frame=<unavailable>, context=0x00000001010015c0, start_with_ip=<unavailable>, filter_exception=0x00000001003e5030, exit_at_finally=-1, base_frame=0x0000000000000000) at interp.c:3012 [opt]
    frame #14: 0x0000000100112644 mono`interp_entry(data=<unavailable>) at interp.c:1749 [opt]
    frame #15: 0x00000001001240fc mono`interp_entry_static_0(rmethod=<unavailable>) at interp.c:2301 [opt]
    frame #16: 0x00000001016e9c99
    frame #17: 0x00007fff6c012661 libsystem_pthread.dylib`_pthread_body + 340
    frame #18: 0x00007fff6c01250d libsystem_pthread.dylib`_pthread_start + 377
    frame #19: 0x00007fff6c011bf9 libsystem_pthread.dylib`thread_start + 13
```

So it looks like without the attach callback we never set the interpreter callback on the natively-created thread.

lewurm: I'll take a look tomorrow

lewurm: pushed a fix

lambdageek: Thanks! I'll update this PR. Could you backport to 2018-08 too

lewurm: already happened 🙂

lewurm:

lewurm:

lewurm:

lambdageek:
This is going to merge poorly with #10561 (static `set_context` function used before it's declared), so I'm going to rebase on `master`, cleanup some of the other fixups in this PR and force-push an update.

luhenry:
@lambdageek just saw your message https://github.com/mono/mono/pull/10545#issuecomment-420747918 😞 I'll approve the PR you submit to fix it right away.

PR #10547, opened 2018-09-11
[bcl] Removes Mono.CSharp from most mobile profiles

marek-safar:
It's no longer needed as a dependency of Microsoft.CSharp and it has
no stable API


marek-safar:
@monojenkins commit csproj

PR #10553, opened 2018-09-11
Make enum values obvious.

jaykrell:


lambdageek:
Why?

jaykrell:
So when I read the header I can know their values quickly.
More readable code. Less writable or mergable, but readable wins.

jaykrell:
I can only eyeball a few lines to count them. 30 is too many to eyeball.

vargaz:
Adding values in the middle will be painful though.

jaykrell:
There are multiple options for adding values:
 1 don't add them in the middle
 2 renumber carefully to preserve readability at the cost of writability
 3 if you really need to add to the middle and renumbering is really difficult, undo this change then

jaykrell:
@monojenkins build  failed

jaykrell:

jaykrell:

jaykrell:

Issue #10244, opened 2018-08-22
[wasm] Make test.js use generated driver and bindings.


kumpera:
Right now sdks/wasm/test.js uses custom code to deal with loading mono. We should replace this with using a generated runtime.js.

Plus, deployment shouldn't depend on us cp'ing test.js to the packaged directory. packager should suppport a `--assets` option that takes non-dotnet deps.

PR #10558, opened 2018-09-12
[WASM] Make test-suite use generated driver and bindings.

kjpou1:
- bindings_tests.dll uses the generated bindings.dll.
- Modify packager to accept a template parameter that overrides the default `runtime.g.js`.
- Modify the tests to call runtime.js in the test-suite directory
- The runtime.js is automatically created in the test-suite library so no need to do extra copies.
- Add -asset option that allows an extra asset to be copied to the output direcotry.
- resolves https://github.com/mono/mono/issues/10244


kjpou1:
@kumpera there were merge conflicts and will fix this today.

kjpou1:
@kumpera This should be ready now after the build checks pass

PR #10563, opened 2018-09-12
[wasm] Pass ac_cv_func_shm_open_working_with_mmap=no to the wasm runtime's configure. Fixes compilation on linux.

vargaz:


kumpera:
Why can't we fix configure instead of this hack?

vargaz:
We could, its just consistent with how other platforms do it.

Issue #10544, opened 2018-09-10
[REGRESSION][master] Getting custom attributes+AttributeType for a `DllImport` method throws NRE


radical:
Failed output:
```
attr.Type: System.Runtime.InteropServices.PreserveSigAttribute

Unhandled Exception:
System.NullReferenceException: Object reference not set to an instance of an object
  at System.Reflection.CustomAttributeData.get_AttributeType () [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/master/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Reflection/CustomAttributeData.cs:147
  at MainClass.Main (System.String[] args) [0x00052] in <a90fc1674b9b42588ce3846543c65a1e>:0
```

Expected output:
```
attr.Type: System.Runtime.InteropServices.PreserveSigAttribute
attr.Type: System.Runtime.InteropServices.FieldOffsetAttribute
```

test case:
```
using System;
using System.Reflection;
using System.Runtime.InteropServices;

public class MainClass
{
	public static void Main(string[] args)
	{
		var mi = typeof(MainClass).FindMembers(MemberTypes.Method, BindingFlags.Static | BindingFlags.Public, (m, criteria) => m.Name == "pinvoke", null);

		foreach (CustomAttributeData data in ((MethodInfo)(mi[0])).CustomAttributes)
		{
			Console.WriteLine($"attr.Type: {data.AttributeType}");
		}
	}

	[DllImport("libc")]
	public static extern void pinvoke();
}
```

This is because of the `CustomAttributeData` being passed a null `ctorInfo` in the .ctor . And that seems to stem from [this](https://github.com/mono/mono/blob/master/mcs/class/corlib/System.Reflection/MonoMethod.cs#L462) where it is trying to find a constructor for `FieldOffsetAttribute` that takes a `string`, but there seems to be only a .ctor taking an `int`. This came from https://github.com/mono/mono/commit/7454c2de2b1e6e4b7da937e91f39d32867801d48 .

This works fine with 2018-06.

marek-safar:
@MaximLipnin it looks like predefined ctor handling broke

MaximLipnin:
It looks like that some changes related to pseudo custom attribute data were not so correct initially.
Mentioned test case for CoreFX and NetFX returns
```
attr.Type: System.Runtime.InteropServices.DllImportAttribute
attr.Type: System.Runtime.InteropServices.PreserveSigAttribute
```
[...]

PR #10568, opened 2018-09-12
[System.Reflection] Fix the issue with constructing DllImportAttribute data

MaximLipnin:
Fixes https://github.com/mono/mono/issues/10544

### DIFF
--- a/mcs/class/corlib/System.Reflection/MonoMethod.cs
+++ b/mcs/class/corlib/System.Reflection/MonoMethod.cs
@@ -33,6 +33,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using InteropServicesCallingConvention = System.Runtime.InteropServices.CallingConvention;
 using System.Runtime.Serialization;
 #if !FULL_AOT_RUNTIME
 using System.Reflection.Emit;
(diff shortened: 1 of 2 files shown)


mcs/class/corlib/System.Reflection/MonoMethod.cs
```diff
@@ -455,18 +456,92 @@ internal CustomAttributeData[] GetPseudoCustomAttributesData ()
...
 
+        private CustomAttributeData GetDllImportAttributeData ()
+        {
+			if ((Attributes & MethodAttributes.PinvokeImpl) == 0)
+				return null;
+
+			string entryPoint, dllName = null;
+			PInvokeAttributes flags = 0;
+
+			((MonoMethod)this).GetPInvoke (out flags, out entryPoint, out dllName);
```
marek-safar: Why do you need this cast?

mcs/class/corlib/System.Reflection/MonoMethod.cs
```diff
@@ -455,18 +456,92 @@ internal CustomAttributeData[] GetPseudoCustomAttributesData ()
...
+
+			CharSet charSet = CharSet.None;
```
marek-safar: This should be set in default case

mcs/class/corlib/System.Reflection/MonoMethod.cs
```diff
@@ -455,18 +456,92 @@ internal CustomAttributeData[] GetPseudoCustomAttributesData ()
...
+				charSet = CharSet.Unicode; 
+				break;
+			case PInvokeAttributes.CharSetAuto: 
+				charSet = CharSet.Auto; 
+				break;
+			// Invalid: default to CharSet.None
+			default: break;
+			}
+
+			var callingConvention = InteropServicesCallingConvention.Cdecl;
```
marek-safar: This should be set in default case

mcs/class/corlib/System.Reflection/MonoMethod.cs
MaximLipnin: No need

mcs/class/corlib/System.Reflection/MonoMethod.cs
MaximLipnin: done

mcs/class/corlib/System.Reflection/MonoMethod.cs
MaximLipnin: done

mcs/class/corlib/System.Reflection/MonoMethod.cs
```diff
@@ -455,18 +456,96 @@ internal CustomAttributeData[] GetPseudoCustomAttributesData ()
...
+			bool preserveSig = (GetMethodImplementationFlags () & MethodImplAttributes.PreserveSig) != 0;
+
+			var ctorArgs = new CustomAttributeTypedArgument [] { 
+				new CustomAttributeTypedArgument (typeof (string), dllName),
+			};
+
+			var attrType = typeof (DllImportAttribute); 
+
+			var namedArgs = new CustomAttributeNamedArgument [] { 
+				new CustomAttributeNamedArgument (attrType.GetField ("EntryPoint"), entryPoint),
```
marek-safar: Hmm, this will break with linker (will update that in separate PR)

PR #10579, opened 2018-09-12
[interp] move thread attach to interpreter entry

lewurm:
We want to attach a thread before any TLS operation is done by the
interpreter. In the JIT this happens inside the native-to-managed
wrapper via `mono_threads_attach_coop ()`. Unfortunately, this is too
late for the interpreter as it already requires a working TLS before
even calling into the wrapper.

Also, fixup naming of locals in the native-to-managed wrapper.

needs backport to `2018-08`


lewurm:
@monojenkins backport 2018-08

PR #10593, opened 2018-09-13
[amd64] Increase max size of OP_LMUL_IMM.

vargaz:


jaykrell:

PR #10595, opened 2018-09-13
[System]: Make sure `HttpWebRequest` observes exceptions on timeout.  #10488.

### LINKED ISSUE #10488: HttpWebRequest throws an uncatchable task exception when processing requests simultaneously with timeout specified
## Steps to Reproduce

Just open up a new project and copy and paste the following code to replicate the error.

```c#
using System;
using System.Net;
using System.IO;
using System.Threading;
using System.Threading.Tasks;

namespace TestProgram
{
    class MainClass
    {
		static bool close_program = false;

		{
			//Create a catch all Task exception handler
			//This will catch the eventual task exception
			TaskScheduler.UnobservedTaskException += (s, e) =>
                LastExceptionHandler(e.Exception, "TaskScheduler.UnobservedTaskException");

			Console.WriteLine("This program contains 2 types of exceptions");
			Console.WriteLine("One that is catchable, the other that is not. The latter will be displayed here.");
			Console.WriteLine("Now wait for the uncatchable exception (less than 60 seconds in most cases), do not press any key.");
			int counter = 0;
			while(true){
				if(Console.KeyAvailable){
					break;
				}
				Thread.Sleep(1000);
				counter++;

                //We will run 2 queries to the same URL in an alternate task
				Task.Run(() => QueryHttp());
				Task.Run(() => QueryHttp());

				Console.WriteLine(counter + " Second(s)");
				if(close_program == true){
					break;
				}
			}
        }

        public static string QueryHttp()
		{
			try
			{
				ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls;
                
				HttpWebRequest request = (HttpWebRequest)WebRequest.Create("https://httpstat.us/200?sleep=3000");
				request.Credentials = System.Net.CredentialCache.DefaultCredentials;
				request.Timeout = 1000;
                            
				using (HttpWebResponse response = (HttpWebResponse)request.GetResponse())
				{
					using (StreamReader respreader = new StreamReader(response.GetResponseStream()))
					{
						return respreader.ReadToEnd();
					}
				}
			}
			catch (Exception)
			{
				//This exception will normally be a timeout exception which is catchable
			}
			return "";
		}

		public static void LastExceptionHandler(Exception e, string source)
        {
			//This will catch the eventual task exception
			Console.WriteLine("Uncatchable thread exception: " + e);
			close_program = true;
        }
    }
}
```

### Current Behavior
This code throws a task exception by method **MyGetResponseAsync** found within HttpWebRequest. The exception here is uncatchable due to the fact that the MyGetReponseAsync is a task created by BeginGetResponse and there is no current way (I'm aware of) to add an exception handler to the MyGetResponseAsync task. So when the exception is thrown, it bubbles to the global task exception handler if one is set.

### Expected Behavior
That the exception created by MyGetResponseAsync should be catchable outside of global exception handlers. The exception itself is probably caused by multithreading issues with HttpWebRequest (as it only happens when multiple requests are sent around the same time) or the threadpool used by the created task.

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:
This error happens also in the most recent version of Mono as well.

```
Mono JIT compiler version 5.12.0.226 (tarball Thu May  3 09:42:09 UTC 2018)
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
```

## Stacktrace

```
N/A - Run the code to see the stacktrace
```

baulig:
`HttpWebRequest.RunWithTimeoutWorker()` needs to observe the worker task's exception
on timeout to prevent it from being thrown via the `TaskScheduler.UnobservedTaskException`
event.  Fixes #10488.

### DIFF
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -949,7 +949,7 @@ namespace System.Net
 #pragma warning disable 4014
 				// Make sure the workerTask's Exception is actually observed.
 				// Fixes https://github.com/mono/mono/issues/10488.
-				workerTask.ContinueWith (t => t.Exception?.GetHashCode ());
+				workerTask.ContinueWith (t => t.Exception?.GetHashCode (), TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.OnlyOnCanceled);
 #pragma warning restore 4014
 				throw new WebException (SR.net_timeout, WebExceptionStatus.Timeout);
 			} catch (Exception ex) {
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -946,6 +946,11 @@ namespace System.Net
 				} catch {
 					// Ignore; we report the timeout.
 				}
+#pragma warning disable 4014
+				// Make sure the workerTask's Exception is actually observed.
+				// Fixes https://github.com/mono/mono/issues/10488.
+				workerTask.ContinueWith (t => t.Exception?.GetHashCode ());
+#pragma warning restore 4014
 				throw new WebException (SR.net_timeout, WebExceptionStatus.Timeout);
 			} catch (Exception ex) {
 				throw GetWebException (ex, aborted ());
--- a/mcs/class/System/System.Net/HttpWebRequest.cs
+++ b/mcs/class/System/System.Net/HttpWebRequest.cs
@@ -949,7 +949,7 @@ namespace System.Net
 #pragma warning disable 4014
 				// Make sure the workerTask's Exception is actually observed.
 				// Fixes https://github.com/mono/mono/issues/10488.
-				workerTask.ContinueWith (t => t.Exception?.GetHashCode (), TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.OnlyOnCanceled);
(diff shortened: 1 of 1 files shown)


baulig:
@monojenkins build failed

marek-safar:
My bad this combination is not allowed, it throws invalid arguments but I did some testing and `OnlyOnFaulted` should be enough.

This works fine

```c#
	internal class Program {

		public static void Main ()
		{
			TaskScheduler.UnobservedTaskException += delegate {
				Console.WriteLine ("crash!!");
			};

			Foo ();

			Thread.Sleep (2000);

			GC.Collect ();
			GC.WaitForPendingFinalizers ();
			GC.Collect ();
		}

		static void Foo ()
		{
			var cc = new CancellationTokenSource ();
			cc.Cancel ();
			Task originalTask = Task.Run (() => throw new Exception (), cc.Token);
			Task continuationTask = originalTask.ContinueWith (t => t.Exception.GetHashCode (), TaskContinuationOptions.OnlyOnFaulted);

			Thread.Sleep (2000);
		}
	}
```

marek-safar:
@monojenkins backport 2018-08

filipnavara:
Thanks for fixing this, we have been suffering from this issue as well!

marek-safar:
@monojenkins backport 2018-06

shnaz:
@marek-safar Any idea when this fix is going to be released?

marek-safar:
The fix will be available with VS16 Preview 2

PR #10613, opened 2018-09-14
Bump roslyn to 2.8.2 + changes to make tests compatible with mono

kg:


kg:
Unresolved questions @marek-safar et al:
* How do I get mono into the PATH for the ```build.sh``` roslyn script? This is a problem we have everywhere in our makefiles and I still don't know how to solve it consistently. I have to manually ```PATH=~/.mono/bin:$PATH``` any time I run 'make' in our tree and that seems like it will be true here on CI as well.
* Using roslyn's script to run tests is 'correct' but their test runner is *probably* not compatible with ours. What's the right way to address that?
* How can we mechanically verify that changes to this infrastructure don't quietly break our support for running the roslyn tests? With how this is set up it feels very easy to accidentally turn a hundred roslyn test failures into a zero exit code.

fwiw, all the roslyn tests appear to pass locally for me w/master mono + my roslyn branch.

kg:
@monojenkins build Linux x64

marek-safar:
>How do I get mono into the PATH for the build.sh roslyn script? This is a problem we have everywhere in our makefiles and I still don't know how to solve it consistently. I have to manually PATH=~/.mono/bin:$PATH any time I run 'make' in our tree and that seems like it will be true here on CI as well.

What we usually do is we have a script/shell variable e.g. `RUNTIME` which we set when it's not set. In this case you'd probably need to update Roslyn script to use it everywhere.

>Using roslyn's script to run tests is 'correct' but their test runner is probably not compatible with ours. What's the right way to address that?

We should try to fix their runner. Ideally, the same runner should be used because it's testing the same thing just with different Mono version (we could add specific options for us though)

>How can we mechanically verify that changes to this infrastructure don't quietly break our support for running the roslyn tests? With how this is set up it feels very easy to accidentally turn a hundred roslyn test failures into a zero exit code.

The roslyn test xunit results should be uploaded to be part of our tests matrix. We could check there the output xml size of file we are going to copy is > min-size.

marek-safar:
@monojenkins build

PR #10615, opened 2018-09-14
Fix MonoTests.System.Linq.ParallelEnumerableTests.TestGroupBy (Linux/amd64/FullAOT/LLVM).

jaykrell:
mono_aot_get_unbox_trampoline: amd64 compute unbox trampoline size correctly via runtime disassembly.
This fixes: MonoTests.System.Linq.ParallelEnumerableTests.TestGroupBy frequently crashing on FullAOT/LLVM/Linux/amd64 CI.

https://github.com/mono/mono/issues/10441
https://github.com/mono/mono/issues/9554
https://github.com/mono/mono/issues/9046

Race condition because the code varies in what actually runs.
Sometimes System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT is called,
sometimes not.

FullAOT+LLVM for reasons obvious from the fix.

Very sensitive to exact code presumably because the bug depends on whether the first
unbox trampoline has the same size as the unbox trampoline for System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT?
For example, this could be the only unbox trampoline or it could be first.

For example System_Security_Cryptography_ECCurve_get_Oid gets the first unbox trampoline and compiling it with mini or LLVM makes a difference.

Guessing here:
When the first unbox trampoline is 9 bytes in size and
System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT
is 6 bytes in size, then System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT's
unbox trampoline, which comes right before System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT
is thought to be 9 bytes also, which overlaps System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT.

So the binary search finds that, which goes wrong somehow -- mono_aot_find_jit_info returns NULL (for reasons not debugged).

And then we fail to wrap System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT
with a gsharedvt trampoline.

In the normal course when things work, calling System_Linq_Parallel_WrapperEqualityComparer_1_T_GSHAREDVT_GetHashCode_System_Linq_Parallel_Wrapper_1_T_GSHAREDVT
goes through both gsharedvt_out_trampoline and gsharedvt_trampoline.

The first turns a byref into a byvalue, the second turns it back into a byref.
When the second is skipped, we end up treating the value as a pointer and SIGSEGV.
This can be seen using https://github.com/mono/mono/pull/10582.

mini amd64 unbox trampolines are always 9 bytes.
LLVM amd64 unbox trampolines hypothetically vary 6 or 9, but are probably always 6.

There maybe other fixes here.

1. The change to use jmp instead of db 0xe9 might be have been more about
consistency than necessacity -- necessary other places.  (https://github.com/mono/mono/commit/2617698743a1#r30505555)

2. It might be possible to force the assembler to output 0xe9.
GNU assembler supports jmp.d32 to force a 32bit displacement opcode 0xE9.
A solution portable to other assemblers is not known.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -9192,7 +9192,10 @@ static void
 emit_code (MonoAotCompile *acfg)
 {
 	int oindex, i, prev_index;
+#ifndef TARGET_AMD64
+	// See mono_aot_get_unbox_trampoline.
 	gboolean saved_unbox_info = FALSE;
+#endif
 	char symbol [MAX_SYMBOL_SIZE];
 
 	if (acfg->aot_opts.llvm_only)
@@ -9293,6 +9296,7 @@ emit_code (MonoAotCompile *acfg)
 			if (acfg->thumb_mixed && cfg->compile_llvm)
 				emit_set_arm_mode (acfg);
 
+#ifndef TARGET_AMD64
 			if (!saved_unbox_info) {
 				char user_symbol [128];
 				GSList *unwind_ops;
@@ -9309,6 +9313,7 @@ emit_code (MonoAotCompile *acfg)
 
 				saved_unbox_info = TRUE;
 			}
+#endif
 		}
 
 		if (cfg->compile_llvm) {
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -2152,14 +2152,13 @@ arch_emit_specific_trampoline (MonoAotCompile *acfg, int offset, int *tramp_size
  *
  *   Emit code for the unbox trampoline for METHOD used in the full-aot case.
  * CALL_TARGET is the symbol pointing to the native code of METHOD.
+ *
+ * See mono_aot_get_unbox_trampoline.
  */
 static void
 arch_emit_unbox_trampoline (MonoAotCompile *acfg, MonoCompile *cfg, MonoMethod *method, const char *call_target)
 {
 #if defined(TARGET_AMD64)
-
-	// See mono_aot_get_unbox_trampoline.
-
 	guint8 buf [32];
 	guint8 *code;
 	int this_reg;
@@ -9192,10 +9191,7 @@ static void
 emit_code (MonoAotCompile *acfg)
 {
 	int oindex, i, prev_index;
-#ifndef TARGET_AMD64
-	// See mono_aot_get_unbox_trampoline.
-	gboolean saved_unbox_info = FALSE;
-#endif
+	gboolean saved_unbox_info = FALSE; // See mono_aot_get_unbox_trampoline.
 	char symbol [MAX_SYMBOL_SIZE];
 
 	if (acfg->aot_opts.llvm_only)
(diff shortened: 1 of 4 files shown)


mono/mini/aot-runtime.c
```diff
@@ -5820,13 +5820,43 @@ mono_aot_get_unbox_trampoline (MonoMethod *method)
...
 		mono_tramp_info_free (tinfo);
 		return FALSE;
 	}
 
-	tinfo->code_size = *(guint32*)symbol_addr;
+	guint32 const code_size = *(guint32*)symbol_addr;
+	tinfo->code_size = code_size;
+
+#ifdef HOST_AMD64
+
```
vargaz: Put this in tramp-amd64.c in a function named mono_arch_get_unbox_trampoline_size () or mono_amd64_get_unbox_trampoline_size (). Looks good otherwise.

vargaz:
Note that it would be simpler if we could force the assembler to always emit the 32 bit version of the jump, it can be done with gnu assembler using a pseudo prefix:
{disp32} jmp foo
But apparently, clang doesn't support it.

mono/mini/aot-compiler.c
```diff
@@ -2157,6 +2157,9 @@ static void
 arch_emit_unbox_trampoline (MonoAotCompile *acfg, MonoCompile *cfg, MonoMethod *method, const char *call_target)
 {
 #if defined(TARGET_AMD64)
+
```
vargaz: It will be simpler to keep the existing code, and add:
#ifdef AMD64
  size = 5;
#endif
There will be some dead code, but it doesn't matter.

jaykrell: I'm not sure what you mean. It kinda makes most sense in aot-runtime.c but you did place your  comment in aot-compiler.c -- deliberately?

So how about https://github.com/mono/mono/commit/2343b8b13d6086372804f99199bb0ed1156a5090?

Runtime is unchanged, compiler outputs a constant 5.

Before I had ifdef host_amd64, which is fine as an ifdef.
Now target_amd64, which is less good.

jaykrell:

jaykrell:
@monojenkins build Linux i386

jaykrell:

marek-safar:
@monojenkins backport 2018-08

Issue #10475, opened 2018-09-05
ConditionalWeakTable throws `NotImplementedException` on `GetEnumerator()`


DavidKarlas:
## Steps to Reproduce

I don't have exact steps, I was debugging VSfM and exception was thrown when I had enabled to catch exceptions... I imagine this can cause memory leaks when using TLS connections since things are not cleaned up because of exception.

### Current Behavior

Throws exception.

### Expected Behavior

Doesn't throw exception but instead returns valid entries.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

5.16.0.112

## Stacktrace

```
mscorlib.dll!System.Runtime.CompilerServices.ConditionalWeakTable<byte[][],object>.System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<TKey,TValue>>.GetEnumerator() in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-06/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs:380
mscorlib.dll!System.Buffers.TlsOverPerCoreLockedStacksArrayPool<byte>.Trim() in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-06/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs:260
mscorlib.dll!System.Buffers.TlsOverPerCoreLockedStacksArrayPool<byte>.Gen2GcCallbackFunc(System.Buffers.TlsOverPerCoreLockedStacksArrayPool<byte> target) in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-06/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs:282
mscorlib.dll!System.Gen2GcCallback.Finalize() in /Users/builder/jenkins/workspace/build-package-osx-mono/2018-06/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/CoreLib/System/Gen2GcCallback.cs:58
```

filipnavara:
cc @EgorBo, who added it while porting CoreFX code.

EgorBo:
[...]

PR #10616, opened 2018-09-14
Implement IEnumerable for ConditionalWeakTable

EgorBo:
Fixes: https://github.com/mono/mono/issues/10475
I copied coreclr's implementation and slightly modified

### DIFF
--- a/mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
+++ b/mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
@@ -31,6 +31,8 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
 
 namespace System.Runtime.CompilerServices
 {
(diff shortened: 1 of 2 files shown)


kumpera:
This change is blatantly wrong in the face of table mutation and GC. How did we not catch it during code review?

marek-safar:
What exactly is wrong and how do we fix it?

kumpera:
The table does compaction which reorders table elements, so the counting setup doesn't work in that case. The fix is to either snapshot, out-of-place compact or somehow remove the need for compaction.

I take back the GC issue, it should work as is.

EgorBo:
so is it ok or there is an issue? the enumerator enumerates inside a `lock` and it takes into account that values might be not compacted as far as I understand. `Rehash` also is wrapped with `lock`

kumpera:
@EgorBo what if compaction happens in between two calls to MoveNext?

EgorBo:
@kumpera by compaction you mean `Rehash`/`RehashWithoutResize`? if so - it won't happen until enumeration is ended.

EgorBo:
@kumpera oh sorry, I am wrong

kumpera:
@EgorBo that's a complicated problem, cuz finalization is inherently racy by itself.
CoreFX doesn't need compaction as much cuz they use separate chaining while we do open addressing, which requires compacting for search correctness.

Issue #10555, opened 2018-09-11
CustomAttribute's typed argument value has incorrect type for enum arguments


radical:
Test case:

```c#
using System;
using System.Reflection;

namespace dm_play
{
    public enum Levels { one, two, three }

    [AttributeUsage(AttributeTargets.Method)]
    public class TestAttribute : Attribute
    {
        Levels this_level;
        public TestAttribute(Levels level) => this_level = level;
    }

    public class MainClass
    {
        public static void Main(string[] args)
        {
            var mi = typeof(MainClass).FindMembers(MemberTypes.Method, BindingFlags.Instance | BindingFlags.Public, (m, criteria) => m.Name == nameof(MethodWithAttribute), null);

            foreach (CustomAttributeData cad in ((MethodInfo)(mi[0])).CustomAttributes) {
                Console.WriteLine($"attr.Type: {cad.AttributeType}");

                foreach (CustomAttributeTypedArgument cata in cad.ConstructorArguments)
                    Console.WriteLine($"-> cata: {cata.ArgumentType}, val: {cata.Value}, type of val: {cata.Value.GetType()}");
            }
        }

        [TestAttribute(Levels.two)]
        public void MethodWithAttribute() => Console.WriteLine("Hello");
    }

}
```

Expected output: (works with 5.16.0)
```
$ mono dm_play.exe
attr.Type: dm_play.TestAttribute
-> cata: dm_play.Levels, val: two, type of val: dm_play.Levels
```

Actual output: (with master)
```
 $ ~/misc/bin/mono dm_play.exe
attr.Type: dm_play.TestAttribute
-> cata: dm_play.Levels, val: 1, type of val: System.Int32
```

The `CustomAttributeTypedArgument.ArgumentType` is the same, but the `CustomAttributeTypedArgument.Value` has different types:
5.16.x: `dm_play.Levels`
master: `System.Int32`

The commit that introduced this behavior: https://github.com/mono/mono/commit/faaa9a41c66c31708b47bbd55d65de9f24c85aff

This breaks msbuild tests being run with xunit.

cc: @marek-safar @MaximLipnin

MaximLipnin:
I'm not sure if it's regression because System.Reflection namespace are moving to CoreFX sources so it's correct to assume that we should follow the CoreFX behavior in general.
The mentioned test case has the same output on CoreFX and NetFX as for actual result above:
`cata: dm_play.Levels, val: 1, type of val: System.Int32`

marek-safar:
[...]

PR #10619, opened 2018-09-14
[Reflection] Use a type of actual argument in case when custom attribute ctor parameter is of System.Object type

MaximLipnin:
Fixes https://github.com/mono/mono/issues/10555


luhenry:
@monojenkins build failed

MaximLipnin:
@monojenkins build failed

Issue #9507, opened 2018-07-12
isinst for unused array type fails


vargaz:
## Steps to Reproduce

Run this:
```csharp
using System;

public struct TestTableScalarSse2<T>
{
        public T[] inArray1;

        public TestTableScalarSse2(int i) {
            inArray1 = new T[16];

			if (inArray1 is double[])
				Console.WriteLine ("OK.");
		}
}

public class Tests
{
	public static void Main () {
		var table = new TestTableScalarSse2<ulong> (1);
	}
}
```
### Current Behavior

```
Unhandled Exception:
System.NullReferenceException: Object reference not set to an instance of an object
  at TestTableScalarSse2`1[T]..ctor (System.Int32 i) [0x0000e] in <0d150afdf55c4bb6b6168c448b599f2c>:0 
  at Tests.Main () [0x00001] in <0d150afdf55c4bb6b6168c448b599f2c>:0 
```

## On which platforms did you notice this

[x ] macOS
[ ] Linux
[ ] Windows

**Version Used**:
Master/5.12

This seems to happen because using the  double[] type in code doesn't seem to initialize the Double class anymore, i.e. mono_class_init () is not called on it.

PR #10626, opened 2018-09-14
[class] Ensure 'cast_class' is initialized for the type parameters of a generic class

alexischr:
[class] Ensure 'cast_class' is initialized for the type
 parameters of a generic class.

Fixes https://github.com/mono/mono/issues/9507

Background: `Monoclass:cast_class` is used internally to allow for some interfaces to be assignable to each other if they share a "storage class" instead of the class's element class. This is used to i.e. implement `IList<ulong>` being assignable from `IList<long>` (`IList<ulong>` has `cast_class=long`)

1) Consider a `GenericClass<T>` with a `T[] field`. On instantiation with e.g. `GenericClass<ulong>` ,  the `ulong[]` array class will be generated. As above, `ulong[]` has `cast_class=long`
2) When `mono_class_init()` is called on an array class, its `element_class` is also initialized.
3) If `GenericClass<T>:Method()` uses any type/subtype checking (e.g.  `if (field is double[])`  ), those may attempt to use `cast_class`, which is not ensured as initialized.

This is hard to trigger: Usually `cast_class==element_class` holds; the `cast_class` can be already initialized by happenstance; and C#'s `is` is evaluated at compile-time if possible. Practically this showed up as a NULL dereference in JITted code.

This solution may also work but have only tested locally:
```
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -4159,10 +4159,16 @@ mono_class_init (MonoClass *klass)
 	klass_byval_arg = m_class_get_byval_arg (klass);
 	if (klass_byval_arg->type == MONO_TYPE_ARRAY || klass_byval_arg->type == MONO_TYPE_SZARRAY) {
 		MonoClass *element_class = klass->element_class;
+		MonoClass *cast_class = klass->cast_class;
+
 		if (!element_class->inited)
 			mono_class_init (element_class);
 		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array element class"))
 			goto leave;
+		if (!cast_class->inited)
+			mono_class_init (cast_class);
+		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array cast class"))
+			goto leave;
 	}

 	UnlockedIncrement (&mono_stats.initialized_class_count);
```

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -4176,6 +4176,22 @@ mono_class_init (MonoClass *klass)
 
 		if (MONO_CLASS_IS_INTERFACE_INTERNAL (klass))
 			mono_class_setup_interface_id (klass);
+
+		MonoGenericInst *ginstance = mono_class_get_generic_class (klass)->context.class_inst;
+		for (guint x = 0; x < ginstance->type_argc; x++) {
+			if (mono_type_is_generic_parameter (ginstance->type_argv [x]))
+				break;
+
+			MonoClass *type_param = mono_class_from_mono_type_internal (ginstance->type_argv [x]);
+
+			if (!m_class_get_cast_class (type_param))
+				break;
+
+			if (m_class_get_cast_class (type_param) == m_class_get_element_class (type_param))
+				break; // element_class is taken care of
+
+			mono_class_init (m_class_get_cast_class (type_param));
+		}
 	}
 
 	if (klass->parent && !klass->parent->inited)
--- a/mono/tests/bug-gh-9507.cs
+++ b/mono/tests/bug-gh-9507.cs
@@ -1,0 +1,21 @@
+using type_with_special_array_cast = System.UInt64; // MonoClass<ulong[]>::cast_class = long
+using other_type = System.Double;
+
+public class TestIsInst<T>
+{
+	public T[] array;
+
+	public TestIsInst() {
+		array = new T[16];
+
+		if (array is other_type[]) // should not crash or throw NullReferenceException
+			throw new System.Exception("Unreachable");
+	}
+}
+
+public class Bug9507  // https://github.com/mono/mono/issues/9507
+{
+	public static void Main () {
+		var table = new TestIsInst<type_with_special_array_cast> ();
+	}
+}
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -692,8 +692,9 @@ TESTS_CS_SRC=		\
 	nested_type_visibility.cs	\
 	dynamic-method-churn.cs	\
 	verbose.cs \
-	generic-unmanaged-constraint.cs	\
-	bug-10834.cs
+	generic-unmanaged-constraint.cs \
+	bug-10834.cs \
+	bug-gh-9507.cs
 
 # some tests fail to compile on mcs
 if CSC_IS_ROSLYN


mono/tests/bug-gh-9507.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+public struct TestIsInst<T>
+{
+        public T[] inArray1;
+
+        public TestIsInst(int i) {
```
luhenry: Fix indentation to respect rest of codebase. It might be a test but it's always kinder on the eye to have code formatting consistency.

mono/metadata/class-init.c
```diff
@@ -4176,6 +4176,19 @@ mono_class_init (MonoClass *klass)
...
+		for (guint x = 0; x < ginstance->type_argc; x++) {
+			if (mono_type_is_generic_parameter (ginstance->type_argv [x]))
+				break;
+
+			MonoClass *type_param = mono_class_from_mono_type_internal (ginstance->type_argv [x]);
+
+			if (m_class_get_cast_class (type_param) == m_class_get_element_class (type_param))
+				break; // element_class is taken care of
+			if (!m_class_is_inited (m_class_get_cast_class (type_param)))
+				mono_class_init (m_class_get_cast_class (type_param));
```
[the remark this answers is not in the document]
alexischr: @vargaz I noticed, but the pattern is used a lot (e.g. in `mono_object_handle_isinst` and `mono_class_compute_gc_descriptor`). I read it as a short-circuits on hot code (1 comparison vs. a call and two more checks). `if (!m_class_is_inited (m_class_get_cast_class (type_param)))` will usually be true.

alexischr:
The failures were fixed and the `if` removed. @vargaz could you re-review?

mono/tests/bug-gh-9507.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public struct TestIsInst<T>
+{
+    public T[] inArray1;
+
+    public TestIsInst(int i) {
+        inArray1 = new T[16];
+
+		if (inArray1 is double[])
+			Console.WriteLine ("OK.");
```
lambdageek: this test is a little mysterious.  I guess the `is double[]` is what's important here?  the Console Write is misleading then.  maybe `throw new Exception ("Unexpected")` ?

Also maybe a comment with a clue of how this relates to `MonoClass:cast_class`?

alexischr: Yeah I copy pasted the test from the issue, I'll make it nicer. I'm writing a nicer commit message too. I'll resolve when they're ready

mono/tests/bug-gh-9507.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+public struct TestIsInst<T>
+{
+    public T[] inArray1;
+
+    public TestIsInst(int i) {
+        inArray1 = new T[16];
+
+		if (inArray1 is double[])
```
luhenry: Fix indentation as well. Again, tests are just like code. Thanks!

alexischr: Sorry, new editor :/

alexischr:
Let's not merge if https://github.com/mono/mono/pull/11268 works as it is a simpler patch. However it may leave the problem unresolved for some Enums and Nullables (e.g. I could not test the new C# 7.3 enum generic constraint)

alexischr:
https://github.com/mono/mono/pull/11268 worked and was merged instead

PR #10631, opened 2018-09-15
Garbage Collect assemblies on domain unload

### LINKED ISSUE #9658: Assemblies with mutual references have incorrect refcount
If assemblies `First.dll` and `Second.dll` reference each other, the `MonoAssembly:ref_count` will be wrong each time they are added to a new `AppDomain`.  Unloading the AppDomains will incorrectly retain references and the assemblies will still be loaded at shutdown.

## Steps to Reproduce

1. Download 
[mutually-recursive.zip](https://github.com/mono/mono/files/2214107/mutually-recursive.zip)

2. `make` 
3. `make run`

### Current Behavior

The end of a log from `make run`:
```
Mono: Unloading domain second dom[0x7fd10053c600], assembly mscorlib[0x7fd1007043d0], ref_count=6
Mono: Unloading domain second dom[0x7fd10053c600], assembly Mono.Security[0x7fd100410cc0], ref_count=2
Mono: Unloading domain second dom[0x7fd10053c600], assembly Main[0x7fd10050c5f0], ref_count=2
Mono: Unloading domain second dom[0x7fd10053c600], assembly First[0x7fd1007170b0], ref_count=4
Mono: Unloading domain second dom[0x7fd10053c600], assembly Second[0x7fd100718060], ref_count=4
Mono: Unloading domain Main.exe[0x7fd100509e10], assembly mscorlib[0x7fd1007043d0], ref_count=5
Mono: Unloading domain Main.exe[0x7fd100509e10], assembly Main[0x7fd10050c5f0], ref_count=1
Mono: Unloading assembly Main [0x7fd10050c5f0].
Mono: Unloading domain Main.exe[0x7fd100509e10], assembly First[0x7fd1007170b0], ref_count=3
Mono: Unloading domain Main.exe[0x7fd100509e10], assembly Second[0x7fd100718060], ref_count=3
Mono: Unloading domain Main.exe[0x7fd100509e10], assembly Mono.Security[0x7fd100410cc0], ref_count=1
Mono: Unloading assembly Mono.Security [0x7fd100410cc0].
Mono: Unloading image /Library/Frameworks/Mono.framework/Versions/5.8.0/lib/mono/gac/Mono.Security/4.0.0.0__0738eb9f132ed756/Mono.Security.dll [0x7fd10081d600].
Mono: Unloading image /Users/alekseyk/work/mutually-recursive/src/Main.exe [0x7fd10180a400].
Mono: Assembly image '/Users/alekseyk/work/mutually-recursive/src/First.dll' still loaded at shutdown.
Mono: Assembly image '/Library/Frameworks/Mono.framework/Versions/5.8.0/lib/mono/4.5/mscorlib.dll' still loaded at shutdown.
Mono: Assembly image '/Users/alekseyk/work/mutually-recursive/src/Second.dll' still loaded at shutdown.
```

Note that `First.dll` and `Second.dll` still have `ref_count`s of 3 at unload.  And are still loaded at shutdown.

### Expected Behavior

Unloading the `AppDomain` should decrement the refcounts for the mutually-referencing assemblies.

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

`master`.  Also 5.8.0.109.  It's been like this for a long time - it's more acute now because it's common to compile against reference assemblies and to accidentally introduce mutual references in the concrete implementation assemblies.

https://github.com/mono/mono/blob/e359121b82e5488bbf721f3fd48c0509b1fc4221/mono/metadata/metadata-internals.h#L116-L125

lambdageek:
This fixes #9658 - cyclic references among assemblies are not properly counted and assemblies are not unloaded at domain unload or app shutdown.

----

## Introducing a pin count ##

In `metadata-internals.h` a comment says:
> The number of appdomains which have this assembly loaded plus the number of 
> assemblies referencing this assembly through an entry in their `image->references`
> arrays. The latter is needed because entries in the `image->references` array
> might point to assemblies which are only loaded in some appdomains, and without
> the additional reference, they can be freed at any time.
> The `ref_count` is initially 0.

Note that this is broken for assemblies with cyclic references which
will never be unloaded even if every domain that referenced them is
unloaded.

To fix that, we keep a second `pin_count` which excludes image
references, but includes domain references.

The invariant is that `pin_count <= ref_count`.
* If `pin_count > 0`, we cannot free the assembly.
* If the `pin_count == 0`, the assembly may be garbage:
    - if `ref_count == 0` the assembly actually is garbage and we can free it eagerly.
    - if `ref_count > 0` we don't know if it's reachable or if it's part of a cycle.

We periodically (at domain unload) run a tracing GC that attempt to
free the cycles.

**WONTFIX**: There is a problem here for embedders: `mono_assembly_open()` and
other APIs that don't reference a domain will hand back assemblies
without incrementing their refcounts.  So once we start running a GC
on domain unload, the assembly will go away.  This was already a
problem, however, for embedders if there were multiple threads and
one of them called `mono_assembly_close()` that was open in both.

### Pre-pinning assemblies for the runtime ### 

There is a race here for Mono: If one thread tries to
open an assembly while another is running the GC, the fresh assembly
will be added to loaded_assemblies before its pinning count could be
incremented.  that means the GC may consider it garbage since it
hasn't been added to a domain or an `image->references` yet.  We add a flag on assembly open
asking to pre-pin the fresh assembly and then to transfer the pin to the image
or the domain (or drop it if two threads race to open and one
loses).

---

## GC Implementation ##

It's a mark-sweep collector.  We keep a `gc_mark` bit in `MonoAssembly` to indicate that the marking phase has seen an assembly.

We use a queue for assemblies that have been marked as reachable from the roots, but whose references haven't been scanned yet.

The roots are elements of `loaded_assemblies` with `pin_count > 0`.

After we're done marking, we collect all assemblies that haven't been marked into a free list.

An issue is that `mono_assembly_close_except_pools` would call `mono_image_close_except_pools` which recursively closes the referenced assemblies.  We don't want that - the free list is not in dependency order, so if references are recursively closed when they're in the free list, we may see the same `MonoAssembly*` after it has already been freed.

So instead we do two passes.

First pass. For each assembly in the free list we:
- remove it from `loaded_assemblies`
- decrement the refcount of each referenced assembly without closing it
- set all references to NULL, breaking links between assemblies.

Second pass.  Force-close each assembly in the free list.

---

## Add `exe_image` as a GC root

Problem: Mono keeps a reference to the main executable assembly in `exe_image` and calls `mono_image_close` on it at shutdown.

This happens after we already GC the root domain, so the `MonoAssembly` for the `MonoImage` exe image already went away.  And so did the references.  Because as far as the `MonoAssembly` was concerned it has a single pin - from the root domain.

So we add `exe_image` as an additional scannable root, and make sure to run another GC in `mono_close_exe_image`

---

## Problems ##

1.  I don't understand `coree.c` and it seems to be doing stuff with `exe_image`.  I guess Windows might break in some way.

2. It feels slow.   Corlib test seemed like it took noticeably longer than without GC.  This could just be because nunit is an aggressive (ab)user of domains.

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -1253,12 +1253,24 @@ assemblyref_public_tok_checked (MonoImage *image, guint32 key_index, guint32 fla
  * invoked.
  */
 void
-mono_assembly_addref (MonoAssembly *assembly, gboolean pinning)
+mono_assembly_addref (MonoAssembly *assembly)
+{
+	mono_assembly_addref_nopin (assembly);
+}
+
+void
+mono_assembly_addref_pin (MonoAssembly *assembly)
 {
 	mono_atomic_inc_i32 (&assembly->ref_count);
 	/* Order matters: increment pincount second */
-	if (pinning)
-		mono_atomic_inc_i32 (&assembly->pin_count);
+	mono_atomic_inc_i32 (&assembly->pin_count);
+	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_ASSEMBLY, "Addref %s ref_count = %d (pin_count = %d)", assembly->aname.name, assembly->ref_count, assembly->pin_count);
+}
+
+void
+mono_assembly_addref_nopin (MonoAssembly *assembly)
+{
+	mono_atomic_inc_i32 (&assembly->ref_count);
 	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_ASSEMBLY, "Addref %s ref_count = %d (pin_count = %d)", assembly->aname.name, assembly->ref_count, assembly->pin_count);
 }
 
@@ -1481,7 +1493,7 @@ static void
 mono_assembly_pinref (MonoAssembly *reference)
 {
 	if (reference && reference != (MonoAssembly*)REFERENCE_MISSING)
-		mono_assembly_addref (reference, TRUE);
+		mono_assembly_addref_pin (reference);
 }
 
 static MonoAssembly*
@@ -1741,7 +1753,7 @@ mono_assembly_load_reference (MonoImage *image, int index)
 
 	if (!image->references [index]) {
 		if (reference != REFERENCE_MISSING){
-			mono_assembly_addref (reference, FALSE);
+			mono_assembly_addref_nopin (reference);
 			/* decrement pin_count from load_reference_by_aname_xxx, above */
 			mono_assembly_dropref (reference, TRUE);
 			if (image->assembly)
@@ -1759,7 +1771,7 @@ mono_assembly_load_reference (MonoImage *image, int index)
 
 	if (image->references [index] != reference) {
 		/* Somebody loaded it before us */
-		mono_assembly_close_internal (reference, TRUE);
+		mono_assembly_close_unpin (reference);
 	}
 }
 
@@ -2857,7 +2869,7 @@ mono_assembly_request_load_from (MonoImage *image, const char *fname,
 		mono_image_addref (mono_defaults.corlib);
 		*status = MONO_IMAGE_OK;
 		if (req->want_pinned)
-			mono_assembly_addref (mono_defaults.corlib->assembly, TRUE);
+			mono_assembly_addref_pin (mono_defaults.corlib->assembly);
 		return mono_defaults.corlib->assembly;
 	}
 
@@ -2881,7 +2893,7 @@ mono_assembly_request_load_from (MonoImage *image, const char *fname,
 			*status = MONO_IMAGE_OK;
 			/* FIXME: this is a race.  Need to ask the search hook to pin for us. */
 			if (req->want_pinned)
-				mono_assembly_addref (ass2, TRUE);
+				mono_assembly_addref_pin (ass2);
 			return ass2;
 		}
 	}
@@ -2929,7 +2941,7 @@ mono_assembly_request_load_from (MonoImage *image, const char *fname,
 		/* Pin ass2 before unlocking so it doesn't get GC'd */
 		ass2 = image->assembly;
 		if (req->want_pinned)
-			mono_assembly_addref (ass2, TRUE);
+			mono_assembly_addref_pin (ass2);
 		mono_assemblies_unlock ();
 		g_free (ass);
 		g_free (base_dir);
@@ -2950,7 +2962,7 @@ mono_assembly_request_load_from (MonoImage *image, const char *fname,
 	/* Unconditionally pin the new assembly.  We want it to survive the call to mono_assembly_load_hook ().
 	 * If it turns out we didn't want it pinned, we will unpin it after the hooks run.
 	 */
-	mono_assembly_addref (ass, TRUE);
+	mono_assembly_addref_pin (ass);
 	mono_assemblies_unlock ();
 
 #ifdef HOST_WIN32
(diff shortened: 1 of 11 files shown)


lambdageek:
This is blocked on a few things right now:
- [ ] consistent AOT failures on CI
- [ ] resolve the assembly open / GC race within the runtime. (runtime callers of `mono_assembly_load_from_predicate` should get a pre-pinned assembly that they're responsible for decrementing after attaching the assembly to a domain or to some `MonoImage:references`)

lambdageek:
@monojenkins build OSX x64 C++

lambdageek:

mono/metadata/assembly.c
```diff
@@ -1460,14 +1468,18 @@ load_reference_by_aname_refonly_asmctx (MonoAssemblyName *aname, MonoAssembly *a
...
 			req.basedir = assm->basedir;
 			reference = mono_assembly_request_byname (aname, &req, status);
 		} else {
 			reference = mono_assembly_loaded_full (aname, TRUE);
-			if (!reference)
+			mono_assembly_pinref (reference);
+			if (!reference) {
 				/* Try a postload search hook */
 				reference = mono_assembly_invoke_search_hook_internal (aname, assm, TRUE, TRUE);
+				mono_assembly_pinref (reference);
```
kumpera: Won't this introduce a race condition as well? We have to increment while holding the lock coordinating unload.

mono/metadata/assembly.c
```diff
@@ -1495,11 +1507,13 @@ load_reference_by_aname_default_asmctx (MonoAssemblyName *aname, MonoAssembly *a
 		MonoAssemblyByNameRequest req;
 		memset (&req, 0, sizeof (req));
 		req.request.asmctx = MONO_ASMCTX_DEFAULT;
```
kumpera: There seem to be a big share part of initing `MonoAssemblyByNameRequest`.
How about we extract it to a function?

mono/metadata/assembly.c
```diff
@@ -1540,6 +1555,7 @@ load_reference_by_aname_individual_asmctx (MonoAssemblyName *aname, MonoAssembly
 	aname = mono_assembly_apply_binding (aname, &maped_name_pp);
 
 	reference = mono_assembly_loaded_full (aname, FALSE);
+	mono_assembly_pinref (reference);
```
kumpera: Why won't this race with the unloader since it's happening outside of the unload coordinating lock?

mono/metadata/assembly.c
```diff
@@ -1222,9 +1227,21 @@ assemblyref_public_tok_checked (MonoImage *image, guint32 key_index, guint32 fla
  * invoked.
  */
 void
-mono_assembly_addref (MonoAssembly *assembly)
+mono_assembly_addref (MonoAssembly *assembly, gboolean pinning)
```
kumpera: Replace the gboolean with an enum or create a pair of functions (addref_pin / addref_nopin). This gboolean is not informative.

lambdageek:
Thanks for the review!

In addition to the helpful suggestions, there is a persistent failure on Windows due to crashes in `mini-generic-sharing.c` that happens when the GC is unloading several images simultaneously that it determined to be garbage:  the issue is that the GC unloads the images in arbitrary order which doesn't respect the invariant that generic sharing wants - that the generic instances go away before or at the same time as the GTD.  I think the solution is to teach the image unload hooks about an entire `MonoImageSet` unloading at the same time, not just a single `MonoImage`.

Not sure why the crash only happens on windows - the mempools ought to already be deallocated on other platforms, too.

lambdageek:
Just to update the status of this PR... we're still blocked on the issue that I describe in https://github.com/mono/mono/pull/10631#issuecomment-431385792.

> there is a persistent failure on Windows due to crashes in mini-generic-sharing.c that happens when the GC is unloading several images simultaneously that it determined to be garbage: the issue is that the GC unloads the images in arbitrary order which doesn't respect the invariant that generic sharing wants - that the generic instances go away before or at the same time as the GTD. I think the solution is to teach the image unload hooks about an entire MonoImageSet unloading at the same time, not just a single MonoImage.

  At this point this PR is relative low-priority work for me, so if anyone wants to take it over, please feel free.  I don't think it would suffice merely to relax the invariant in `mini-generic-sharing.c` - we actually need to teach the runtime about simultaneously unloading multiple images.  The assertion is probably masking other issues.

PR #10632, opened 2018-09-15
Fix s390x Outarg_VT

nealef:
OUTARG_VT processing was incorrect - the structure wasn't being used properly on the stack such that things were being overwritten. 

This fix is still incomplete but I'd like to have something for people to examine and comment upon. It fails in the following test case:

```
using System;
using System.Text;

public class main{
        static void Main (string[] args)
        {
                decimal x = 100m;
                decimal y = 500m;
                var z = 1 / (y / x);
                Console.WriteLine("{0} {1} {2}",x,y,z);
        }
}
```

Where it gives the results:
```
100 5 0.2
```

If I change the division to:
```
var z = y / x;
```

It gives the expected result of
```
100 500 5
```


luhenry:
@nealef feel free to remove the `do-not-merge` label whenever you think the fix is complete. Thank you!

PR #10637, opened 2018-09-16
[wasm] Compile the runtime with --disable-icall-tables and explicitly link in libmono-icall-table.a. This will allow linking out unused icalls in the future in full-aot mode.

vargaz:


lewurm:

Issue #10508, opened 2018-09-07
thread-suspend-suspended.exe and thread-suspend-selfsuspended.exe fail a lot on Windows i386


akoeplinger:
## Steps to Reproduce

1. Run thread-suspend-suspended.exe and thread-suspend-selfsuspended.exe tests on 32bit Windows (it doesn't happen on 64bit!)
2. The tests time out very often.

Might be related to the recent switch to hybrid suspend.

### Current Behavior

The tests time out after 5 minutes. Examples:

- https://jenkins.mono-project.com/job/z/label=w32/7844/testReport/
- https://jenkins.mono-project.com/job/z/label=w32/7843/testReport
- https://jenkins.mono-project.com/job/z/label=w32/7841/testReport
- https://jenkins.mono-project.com/job/z/label=w32/7838/testReport/
- https://jenkins.mono-project.com/job/z/label=w32/7837/testReport

### Expected Behavior

No failure.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[ X ] Windows i386 (not x64!)

**Version Used**: master

/cc @lateralusX

lateralusX:
Looks like first failure of one of these tests occured here (might not be directly related though):

https://jenkins.mono-project.com/job/z/label=w32/7657/

Will try to reproduce this in the debugger during the week.

lateralusX:
Still no local repro, but investigated the code heavily under debugger and could force a couple of scenarios that could (at least in theory) happen under wow64 related to thread suspend failing. This scenario is then not handled by sgen (and very uncommon and only observed under wow64) but could get us into either a deadlock or asserts when doing the GC in scenarios like the once's hit by the test. Will need to spend some more time investigate this a little deeper in the debugger, but looks like this boils down to a potential case when mono_threads_suspend_begin_async_suspend returns FALSE (could happen for various resons under wow64).

lateralusX:
[...]

PR #10646, opened 2018-09-17
Fix scheduling problem related to thread-suspend-suspended and thread-suspend-selfsuspended.

lateralusX:
thread-suspend-suspended and thread-suspend-selfsuspended tests has showed long execution times on low core Windows machines running wow64 (x86 binary on x64 OS). It looks like an OS related issue where ResumeThread Win32 API takes a lot longer time to complete when running a test scenario like thread-suspend-suspended and this additional time adds up over many iterations pushing it over to timeout.

NOTE, the same behavior has been identified running on .NET framework, so it's not a mono specific issue and only reproduce under wow64 on low core machines, like the Jenkins Windows builders that only have 2 cores.

Since the purpose of the test is to run GC when t1 is suspended, doing this more deterministic will fix the ResumeThread issue under wow64 on low core machines as well as making the tests more deterministic, always having t1 in a suspended state when running GC (that will test the suspend suspended scenario).

Full investigation can be found in, #10508.


lateralusX:
@luhenry, review?

Issue #9604, opened 2018-07-17
Reflection ResolveMember crashes on invalid input


DReichl:
When passing a non-existing metadata token (of type `MemberRef` though) to the `ResolveMember` method of `Module`, Mono crashes with a SIGABRT (details below).

The expected behavior is that an `ArgumentOutOfRangeException` is thrown, see
[Module.ResolveMember Method (Int32) (System.Reflection)](https://msdn.microsoft.com/en-us/library/2xafyzy1.aspx)

Example program:

```csharp
using System;
using System.Reflection;

namespace ResolveMemberBug
{
	public static class Program
	{
		public static void Main(string[] args)
		{
			try
			{
				Module m = typeof(Program).Module;
				m.ResolveMember(0x0A00F000);
			}
			catch(ArgumentOutOfRangeException)
			{
				Console.WriteLine("Catched ArgumentOutOfRangeException.");
			}
		}
	}
}
```

This should print "Catched ArgumentOutOfRangeException." instead of crashing.

Tested with Mono 5.12.0.226.

## Stacktrace

```
* Assertion at metadata.c:1117, condition `idx < t->rows' not met

Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.Reflection.Module.ResolveMemberToken (intptr,int,intptr[],intptr[],System.Reflection.ResolveTokenError&) [0x0000b] in <71d8ad678db34313b7f718a414dfcb25>:0
  at System.Reflection.MonoModule.ResolveMember (int,System.Type[],System.Type[]) [0x00015] in <71d8ad678db34313b7f718a414dfcb25>:0
  at System.Reflection.Module.ResolveMember (int) [0x00000] in <71d8ad678db34313b7f718a414dfcb25>:0
  at ResolveMemberBug.Program.Main (string[]) [0x0000f] in <14f68449117d4ab089871682c5d52ce5>:0
  at (wrapper runtime-invoke) <Module>.runtime_invoke_void_object (object,intptr,intptr,intptr) [0x0004e] in <14f68449117d4ab089871682c5d52ce5>:0
/proc/self/maps:
4083a000-4084a000 rwxp 00000000 00:00 0 
41335000-41345000 rwxp 00000000 00:00 0 
55e65100d000-55e65142e000 r-xp 00000000 08:01 791829                     /usr/bin/mono-sgen
55e65162d000-55e651634000 r--p 00420000 08:01 791829                     /usr/bin/mono-sgen
[...]

PR #10648, opened 2018-09-17
[corlib] Throw AOORE for invalid input in Module.ResolveMember

EgorBo:
Fixes: https://github.com/mono/mono/issues/9604 

`ResolveMember` crashes here https://github.com/mono/mono/blob/master/mono/metadata/metadata.c#L1158-L1159 (comes from [mono_memberref_is_method](https://github.com/mono/mono/blob/362d1d94e11ff0a44bac7aa6830a4783d9c70610/mono/metadata/icall.c#L6297))

All other resolvers such as [module_resolve_field_token](https://github.com/mono/mono/blob/362d1d94e11ff0a44bac7aa6830a4783d9c70610/mono/metadata/icall.c#L6239-L6242) have this kind of check for out-of-bounds.

### DIFF
--- a/mcs/class/corlib/Test/System.Reflection/ModuleTest.cs
+++ b/mcs/class/corlib/Test/System.Reflection/ModuleTest.cs
@@ -318,6 +318,13 @@ public class ModuleTest
 	}
 
 	[Test]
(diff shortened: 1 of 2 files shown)


EgorBo:
@monojenkins build failed

PR #10667, opened 2018-09-18
[bcl] Avoid retaining System.Globalization.CultureInfo.CreateCulture () in mscorlib.xml, its not longer used by the runtime.

vargaz:


vargaz:
@monojenkins build failed

PR #10681, opened 2018-09-19
[Coop] Convert dynamic-image.c.

jaykrell:
See https://jenkins.mono-project.com/job/test-mono-mainline-staticanalysis/Centrinel_20Report/.

Perhaps some unnecessary change here, the out parameter vs. boolean. If so I can reduce it.


jaykrell:

jaykrell:

PR #10686, opened 2018-09-19
 Some cleanup in appdomain.c.

jaykrell:
Noticed some duplication of exit paths doing coop conversion for example.

### DIFF
--- a/mono/eglib/gstr.c
+++ b/mono/eglib/gstr.c
@@ -914,22 +914,17 @@ g_utf16_asciiz_equal (const gunichar2 *utf16, const char *ascii)
 	}
 }
 
-gchar *
-g_strdelimit (gchar *string, const gchar *delimiters, gchar new_delimiter)
+void
+g_strdelimit (gchar *string, char delimiter, char new_delimiter)
 {
-	gchar *ptr;
+	char *ptr;
 
-	g_return_val_if_fail (string != NULL, NULL);
-
-	if (delimiters == NULL)
-		delimiters = G_STR_DELIMITERS;
+	g_return_if_fail (string != NULL);
 
 	for (ptr = string; *ptr; ptr++) {
-		if (strchr (delimiters, *ptr))
+		if (delimiter == *ptr)
 			*ptr = new_delimiter;
 	}
-	
-	return string;
 }
 
 gsize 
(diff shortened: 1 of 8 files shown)


mono/eglib/gstr.c
```diff
@@ -914,22 +914,17 @@ g_utf16_asciiz_equal (const gunichar2 *utf16, const char *ascii)
 	}
 }
 
-gchar *
-g_strdelimit (gchar *string, const gchar *delimiters, gchar new_delimiter)
+void
```
luhenry: There is now no way to verify if the call succeeded. What's the point of this change?

jaykrell: Nobody checks the return value, and it is only ever the parameter returned back to the caller unchanged, and it isn't really failure, it is just don't replace what isn't present, and what caller doesn't assume is present -- doesn't seem worthwhile.

jaykrell:
@monojenkins build Linux x64

mono/metadata/appdomain.c
```diff
@@ -1513,49 +1503,24 @@ set_domain_search_path (MonoDomain *domain)
...
-		 * domain->search_path [0] = NULL;
-		*/
-		mono_domain_assemblies_unlock (domain);
-		return;
-	}
-
-	if (domain->search_path)
-		g_strfreev (domain->search_path);
+	g_strfreev (domain->search_path);
+	domain->search_path = NULL;
```
jaykrell: Previously this left a dangling pointer in the error path soon after.

jaykrell:

mono/eglib/gstr.c
lambdageek: @luhenry, I agree with @jaykrell  - I don't think a non-`void` return type serves any purpose here.

luhenry: On one hand the return indeed doesn't serve a purpose right now, but since it's actually used to return if there was an error or not, it could be used by callers.

jaykrell:
I'm not saying it started here, but this PR demonstrates a regression in Linux AMD64 FullAOT LLVM. I'll investigate a bit more, for when it started, etc.

```
mono_log_write_logfile (log_domain=<optimized out>, level=<optimized out>, hdr=<optimized out>, message=0x5601f0ba5dc0 "* Assertion at mini-llvm.c:4564, condition `GET_LONG_IMM (ins) == ins->inst_l' not met\n") at mono-log-common.c:135
```

PR #10691, opened 2018-09-19
[wasm] Cross compile the AOT compiler for windows.

kumpera:


sdks/builds/runtime.mk
```diff
@@ -142,7 +142,7 @@ _cross-runtime_$(1)_CONFIGURE_FLAGS= \
 
 .stamp-$(1)-$$(CONFIGURATION)-configure: | $$(if $$(IGNORE_PROVISION_LLVM),,provision-$(5))
 
-$$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)/$(3).h: .stamp-$(1)-$$(CONFIGURATION)-configure $$(TOP)/tools/offsets-tool/MonoAotOffsetsDumper.exe | configure-$(4)
+$$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)/$(3).h: .stamp-$(4)-$$(CONFIGURATION)-configure .stamp-$(1)-$$(CONFIGURATION)-configure $$(TOP)/tools/offsets-tool/MonoAotOffsetsDumper.exe | configure-$(4)
```
luhenry: This is unnecessary: the chain on dependency should be `configure-$(4)` -> `.stamp-$(4)-configure` -> `.stamp-$(4)-$$(CONFIGURATION)-configure`. You can see that at https://github.com/mono/mono/blob/a4d0e665e190e551b043558d7c1d26bf5bcf705a/sdks/builds/Makefile#L88 and https://github.com/mono/mono/blob/a4d0e665e190e551b043558d7c1d26bf5bcf705a/sdks/builds/runtime.mk#L70 respectively. If that chain of dependency is broken for the `wasm` targets, the `wasm` targets should be themselves be fixed.

sdks/builds/wasm.mk
```diff
@@ -38,25 +38,31 @@ $(TOP)/sdks/builds/toolchains/emsdk:
 .stamp-wasm-runtime-toolchain: .stamp-wasm-toolchain
 	touch $@
 
-.stamp-wasm-runtime-configure: $(TOP)/configure
-	mkdir -p $(TOP)/sdks/builds/wasm-runtime
-	cd $(TOP)/sdks/builds/wasm-runtime && source $(TOP)/sdks/builds/toolchains/emsdk/emsdk_env.sh && CFLAGS="-Os -g" emconfigure $(TOP)/configure $(WASM_RUNTIME_AC_VARS) $(WASM_RUNTIME_CONFIGURE_FLAGS)
+.stamp-wasm-runtime-$(CONFIGURATION)-configure: $(TOP)/configure
```
luhenry: Add the following to fix https://github.com/mono/mono/pull/10691/files#diff-77ef5cf8324c86c1370f3550e8eeffd2R145:
```
.PHONY: .stamp-wasm-runtime-configure
.stamp-wasm-runtime-configure: .stamp-wasm-runtime-$(CONFIGURATION)-configure
```

This duplicates the logic at https://github.com/mono/mono/blob/a4d0e665e190e551b043558d7c1d26bf5bcf705a/sdks/builds/runtime.mk#L70

sdks/builds/wasm.mk
```diff
@@ -108,3 +114,66 @@ clean-wasm-cross:
...
+# Parameters
+#  $(1): target
+#  $(2): host arch
+#  $(3): target arch
+#  $(4): device target
+#  $(5): llvm
+#  $(6): offsets dumper abi
+define WasmCrossMXETemplate
+
+_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$(EMSCRIPTEN_SDK_DIR)/emscripten/$(EMSCRIPTEN_VERSION)"
```
luhenry: Should we have a `--gen-wasm` to duplicate the logic of `--gen-ios` and `--gen-android`?

sdks/builds/wasm.mk
```diff
@@ -108,3 +114,66 @@ clean-wasm-cross:
...
+	--with-llvm=$(TOP)/sdks/out/llvm-llvmwin32 \
+	--enable-minimal=appdomains,com,remoting \
+	--disable-boehm \
+	--disable-mcs-build \
+	--disable-nls \
+	--enable-maintainer-mode \
+	--with-tls=pthread
+
+.stamp-wasm-$(1)-$$(CONFIGURATION)-configure: | $(if $(IGNORE_PROVISION_MXE),,provision-mxe)
+	mkdir -p $(TOP)/sdks/builds/wasm-runtime-$(CONFIGURATION)
```
luhenry: Why is that needed? I assume you need that because you require the MonoAotOffsetDumper to look at an existing folder. If so that should already be taken care of by the following chain of dependency: `build-wasm-cross-win` -> `$(TOP)/sdks/builds/wasm-cross-win-$(CONFIGURATION)/wasm32-unknown-unknown.h` -> `configure-wasm-runtime` -> `.stamp-wasm-runtime-configure` -> `.stamp-wasm-runtime-$(CONFIGURATION)-configure`. If you ran into an error, it might be related to https://github.com/mono/mono/pull/10691/files#diff-77ef5cf8324c86c1370f3550e8eeffd2R145

sdks/builds/wasm.mk
```diff
@@ -108,3 +114,66 @@ clean-wasm-cross:
...
+#  $(4): device target
+#  $(5): llvm
+#  $(6): offsets dumper abi
+define WasmCrossMXETemplate
+
+_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$(EMSCRIPTEN_SDK_DIR)/emscripten/$(EMSCRIPTEN_VERSION)"
+
+_wasm-$(1)_PATH=$$(MXE_PREFIX)/bin
+
+_wasm-$(1)_AR=$$(MXE_PREFIX)/bin/$(2)-w64-mingw32$$(if $$(filter $(UNAME),Darwin),.static)-ar
```
luhenry: Use `$$(UNAME)` instead of `$(UNAME)`, it's a typo in the original `android.mk`.

sdks/builds/wasm.mk
```diff
@@ -108,3 +114,66 @@ clean-wasm-cross:
...
+	--with-tls=pthread
+
+.stamp-wasm-$(1)-$$(CONFIGURATION)-configure: | $(if $(IGNORE_PROVISION_MXE),,provision-mxe)
+	mkdir -p $(TOP)/sdks/builds/wasm-runtime-$(CONFIGURATION)
+
+$$(eval $$(call CrossRuntimeTemplate,wasm-$(1),$(2)-w64-mingw32$$(if $$(filter $(UNAME),Darwin),.static),$(3)-unknown-none,$(4),$(5),$(6)))
+
+endef
+
+$(eval $(call WasmCrossMXETemplate,cross-win,i686,wasm32,wasm-runtime,llvm-llvmwin32,wasm32-unknown-unknown))
```
luhenry: Add `ifeq ($(DISABLE_WASM_CROSS),)` otherwise you might still have access to the target while `mxe` is not enabled.

sdks/builds/wasm.mk
kumpera: No idea, this was done by @vargaz

sdks/builds/wasm.mk
vargaz: The cross compiler needs the target runtime to be configured since it uses the generated header files from there.

sdks/builds/runtime.mk
kumpera: Ok, I'll change to be `.stamp-$(4)-configure`.
This dep is needed otherwise building the win cross compiler by itself doesn't work.

kumpera: This dep is requires because the offsets file is generated by the configure of the runtime configuration.

IE, `configure-wasm-runtime-release` produces the offset needed by `build-wasm-crosswin-release`

sdks/builds/wasm.mk
kumpera: Ok, added

sdks/builds/wasm.mk
luhenry: Yep, and that's done by the chain of dependency I described in https://github.com/mono/mono/pull/10691#discussion_r218862796

luhenry: Remove that `mkdir -p ...` command, it's already taken cared of by the chain of dependency.

sdks/builds/runtime.mk
```diff
@@ -142,7 +142,7 @@ _cross-runtime_$(1)_CONFIGURE_FLAGS= \
 
 .stamp-$(1)-$$(CONFIGURATION)-configure: | $$(if $$(IGNORE_PROVISION_LLVM),,provision-$(5))
 
-$$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)/$(3).h: .stamp-$(4)-$$(CONFIGURATION)-configure .stamp-$(1)-$$(CONFIGURATION)-configure $$(TOP)/tools/offsets-tool/MonoAotOffsetsDumper.exe | configure-$(4)
+$$(TOP)/sdks/builds/$(1)-$$(CONFIGURATION)/$(3).h: .stamp-$(4)-configure .stamp-$(1)-$$(CONFIGURATION)-configure $$(TOP)/tools/offsets-tool/MonoAotOffsetsDumper.exe | configure-$(4)
```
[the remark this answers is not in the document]
kumpera: Well, for some reason `configure-$(4)` is not doing what's supposed to because without this extra dependency `make configure-wasm-cross-win` fails because `wasm-runtime-release` was not configured.

This was after applying all the other changes you proposed.

sdks/Make.config.sample
```diff
@@ -1,22 +0,0 @@
-#This is a sample file of all options that can be used by a Make.config file.
```
kumpera: Why are we deleting this file?

sdks/builds/wasm.mk
```diff
@@ -25,88 +38,120 @@ WASM_RUNTIME_CONFIGURE_FLAGS = \
...
+	$(MAKE) -C $(TOP)/sdks/builds/wasm-runtime-$(CONFIGURATION)/mono install
 
 .PHONY: clean-wasm-runtime
 clean-wasm-runtime:
-	rm -rf .stamp-wasm-runtime-configure $(TOP)/sdks/builds/wasm-runtime $(TOP)/sdks/builds/wasm-runtime.config.cache $(TOP)/sdks/out/wasm-runtime
+	rm -rf .stamp-wasm-runtime-toolchain .stamp-wasm-runtime-$(CONFIGURATION)-configure $(TOP)/sdks/builds/wasm-runtime-$(CONFIGURATION) $(TOP)/sdks/builds/wasm-runtime-$(CONFIGURATION).config.cache $(TOP)/sdks/out/wasm-runtime-$(CONFIGURATION)
 
 TARGETS += wasm-runtime
 
-UNAME := $(shell uname -s)
```
kumpera: How come we don't need this?

sdks/Make.config.sample
luhenry: That was part of your initial commit.

sdks/builds/wasm.mk
luhenry: It's defined at https://github.com/mono/mono/blob/6d5bf96f1c7f01c223beb9363359a117ae8a505f/sdks/builds/Makefile#L13

sdks/Make.config.sample
kumpera: err, then there's something wrong with it.

luhenry: Ok, let me revert it then.

PR #10700, opened 2018-09-19
DO NOT MERGE - Rework OpenSSL TLS Provider

nealef:
Incorporate several large fixes/enhancem…ents including certificate renegotiation.

Add OpenSSL as a TLS provider:
- Basically copy the Btls code and do some renaming of files, methods, and variables
- Reuse some boringssl code to provide a few extra methods
- Set this as a configurable option
- Enable it to be selected via MONO_TLS_PROVIDER

Btls suffers from a couple of problems:
- Lack of big endian support
- Unable to take advantage of hardware acceleration like that available on s390x:
-- s390x has instruction support for most of the symmetrical ciphers
-- s390x has a co-processor for the asymmetrical ciphers
-- This support is available via the openssl libraries

Current deficiency:
- No support for the PKCS12_get_key_and_certs API

I have added the "do-not-merge" label so that it can be tested in different environments. As such I am looking for suggestions how the test suite could be enabled to test btls and/or openssl.


marek-safar:
This is very heavy copy&paste code even with sections for Mobile and I am worried how maintainable this will be in the long term.

Secondly what openssl version does this work with?

NattyNarwhal:
As the maintainer for a big endian platform, I'm interested in this. As Mono catches up with upstream CoreFX and BoringTLS, BoringTLS will likely be harder and harder to keep portable. OpenSSL could be an appropriate substitute in this case.

nealef:
I totally understand your reasoning. Part of what prompted me from generating this PR was to explore such issues. I don't think I could have articulated what I was trying to do without code to instantiate it. The duplication of Mono.Btls was just to keep my hands off existing working code.

I wonder if using what I think you were alluding to would be to keep the code in mono/openssl as having both provider code in the one set of modules would result in a fair bit of #ifdefing as the semantics for some of the APIs is different as well as the header files. Then in the individual ```MonoBtlsXXXX.cs/X509BtlsXXXX.cs``` files use generic names (```mono_tls_xxx``` instead of either mono_btls/mono_openssl) and select the .so/dll containing the appropriate provider's APIs at runtime startup - this would mean changing mono/btls and mono/openssl code entry point names.

I am happy to wait until you have a design in mind and then to collaborate as best I can.

baulig:
Well, what I dislike most about this is the duplication of the managed code - which will make it both hard to see where the two providers are different and also maintain it long-term.  For instance, the TLS Provider framework is somewhat interconnected with the X509Certificate code (which will undergo fundamental changes).

migueldeicaza:
To bring the gitter conversation here.

The BTLS/OpenSSL code currently P/Invoke into a support library, without being able to diff the C# code between BTLS and OpenSSL it is difficult to tell if there is any substance or if it was just a matter of renaming/adjusting types, files and p/invokes.

Assuming that the different is only the above, one thing that we could do is to make the OpenSSL C runtime that we P/Invoke into have the same API as the BTLS P/Invokeable surface.

Then, we could add this to `$prefix/etc/mono/config`

```
<dllmap dll=“libmono-btls-shared” target=“libmono-openssl-shared” cpu=“s390x”/>
```

And of course users could do the same.

Stretch goal: make it prettier by renaming symbols from "btls" in the above library into "ssl" so that we remove a level of surprise.

nealef:
> Stretch goal: make it prettier by renaming symbols from "btls" in the above library into "ssl" so that we remove a level of surprise.

Or tls just to be pedantique ;-)

migueldeicaza:
Sure - but you did not answer my main concern/question ;-)

nealef:
I will provide a diff shortly. I think the answer is going to be sweet-f-a.

nealef:
Looking at the diff, there is nothing of consequence. BoringSSL has more error codes though:

```
+       enum MonoBtlsSslError
        {
                None = 0,
                Ssl = 1,
                WantRead = 2,
                WantWrite = 3,
                WantX509Lookup = 4,
                Syscall = 5,
                ZeroReturn = 6,
                WantConnect = 7,
-               WantAccept = 8
+               WantAccept = 8,
+               WantChannelIdLookup = 9,
+               PendingSession = 11,
+               PendingCertificate = 12,
+               WantPrivateKeyOperation = 13
        }
```

baulig:
After having a second look at this and also reading the comments, I am very unhappy with the current situation and would like to at least discuss a few issues before landing this.

The `Mono.Btls.Interface` assembly is only used by the web-tests and as its [README.md](https://github.com/mono/mono/blob/master/mcs/class/Mono.Btls.Interface/README.md) states it's not stable API and should not be used by third parties.  Adding a copy for OpenSSL only makes sense if we'd also update the web-tests to use it - but then the question arises why we cannot just update the existing one.  Let's just not worry about this, as I am maintaining the web-tests, I can take care of that at some point.  For the moment, I would just like to avoid increasing this assembly's usable and size.

About the code duplication, I disagree with Miguel here - adding a new TLS Provider is not an "add-on" or "plugin" because there is no "stable connection" via a public API.  In fact, the entire `Mono.Btls` namespace is built on top of internal APIs _which are going to change in very near future_.  Similarly, the assumption that this could be maintained solely as a community project is wrong because my upcoming changes are essentially ripping away parts of the foundation underneath its feet.

The internal APIs (especially `Mono.Net.Security` and `X509CertificateImpl`) that the TLS Providers use were never intended to be stable and backwards-compatible - even the public `Mono.Security.Interface` APIs should not be considered to last for all eternity.

Within the next ~3 months, major refactoring changes are going to happen with the X509Certificate code and most of the internal classes like `X509CertificateImpl` will undergo major rework.  Some of that work already exists in some of my work branches.  Several internal helper classes such as `X509Helper2` will go away completely.  Others will receive a major overhaul (`X509CertificateImpl`).  The code in `AppleTlsProvider` that deals with certificates will be completely replaced.  Its `MonoBtlsProvider` and `MonoBtlsContext` counterparts will receive majors changes as well.  The entire `X509CertificateImplBtls` class will be almost completely rewritten - with most of its code being moved into a new `X509CertificateImplUnix`.

I don't want to break your code to the point that it won't even compile anymore each time I'm making changes to the internal certificate code - but me having to edit three TLS Providers instead of just two isn't ideal either.  We do not need to address these problems now and I am also not strictly against landing this PR as-is either - it's just that I am worried and concerned about the long-term maintainability.

However, I have some ideas on how we can fix all these issues while also avoiding those "areas of risk" that @migueldeicaza was worried about.

baulig:
First of all, will the distinction between using boringssl or openssl be done at compile-time or at run-time?  Will the class libraries contain both managed implementations or just one - selected at compile time?

Depending on the answer of this question, we have two options - two separate TLS Providers with a shared abstract base - or partial classes.  Neither of these two options should cause any major risk of breaking existing code because we'll just be shuffling some code around.

I would like all code to be shared between the boringssl and openssl versions that either implements or accesses any of the following internal APIs (which can essentially be summarized as anything that deals with certificates, the certificate store or TLS Provider framework):

- `Mono.Net.Security.MobileTlsContext`
- `Mono.Net.Security.ChainValidationHelper`
- `Mono.Security.Interface.MonoTlsProvider`
- `System.Security.Cryptography.X509CertificateImpl`
- `System.Security.Cryptography.X509Certificate2Impl`
- `System.Security.Cryptography.X509ChainImpl`
- `System.Security.Cryptography.X509PalImpl`

We can either introduce new abstract base classes or use partial classes.

About the native code, I don't really have any strong feelings either way and pretty much need to trust your judgment @nealef about how different the two implementations are.  We can certainly rename all of the native functions - I also don't have any pending changes in this regard, so it won't conflict with any of my ongoing work.

For the managed side, could both implementations use the same native API or would that need to be separate?  If it needs to be separate, then could we possibly move everything that needs to access the native API into partial class parts?

I am just thinking that if we could introduce a few abstract base classes (or partial class parts) and move everything that has any "outside connections" into those, then we could actually keep most of the implementation separate while still allowing me to freely change those "outside connections" without the risk of possibly breaking anything.

How does this sound?

baulig:
Would it be possible to share the `Mono.Btls.MonoBtlsObject` base class (and possibly the error handling as well) between both providers?

baulig:
Speaking of renaming things - this is just some idea - we could actually rename things into something like this:

- Mono.Tls
  - Mono.Tls.Apple
  - Mono.Tls.Unix
    - Mono.Tls.OpenSsl
    - Mono.Tls.BoringSsl

The internal certificate code could then follow this pattern.

How would you like this @migueldeicaza?

lewing:
I'm closing this because of inactivity, conflicts, and the do-not-merge label.  Please reopen if the intention is to continue work here.

nealef:
I would like to reopen this discussion and come to agreement that if providing openssl is do-able then how it should be slotted in. I'd like to use the last comment by @baulig as the starting point. I've left things alone as I was busy with other s390x issues and the comments in this issue from @migueldeicaza and @baulig regarding corefx and refactoring. I am hoping that things are settled enough to progress to make openssl a community supported option.

akoeplinger:
@nealef please open a new issue to make sure this doesn't get lost (the PR can't be reopened since the fork it was created from was deleted).

PR #10703, opened 2018-09-19
[corlib] Fix possible race condition in ConditionalWeakTable

EgorBo:
Addresses https://github.com/mono/mono/pull/10616#issuecomment-422924261 
Alternative solution is to give a copy of the whole array to the enumerator

cc @kumpera

### DIFF
--- a/mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
+++ b/mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
@@ -376,112 +376,21 @@ namespace System.Runtime.CompilerServices
 			}
 		}
 
-		// IEnumerable implementation was copied from CoreCLR
 		IEnumerator<KeyValuePair<TKey, TValue>> IEnumerable<KeyValuePair<TKey, TValue>>.GetEnumerator ()
 		{
+			var list = new List<KeyValuePair<TKey, TValue>> (data.Length);
+			var tombstone = GC.EPHEMERON_TOMBSTONE;
 			lock (_lock)
 			{
-				return size == 0 ?
-					((IEnumerable<KeyValuePair<TKey, TValue>>)Array.Empty<KeyValuePair<TKey, TValue>>()).GetEnumerator() :
-					new Enumerator(this);
+				foreach (var eph in this.data) 
+				{
+					if (eph.key != null && eph.key != tombstone)
+						list.Add (new KeyValuePair<TKey, TValue> ((TKey) eph.key, (TValue) eph.value));
+				}
 			}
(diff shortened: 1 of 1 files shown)


mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
```diff
@@ -132,6 +133,17 @@ private void Rehash ()
 
 			uint newLength = (uint)HashHelpers.GetPrime (((int)(size / LOAD_FACTOR) << 1) | 1);
 
+			if (activeEnumeratorRefCount > 0)
+			{
+				// do not rehash if there are active enumerators
+				// Resize if there is no any free slot
+				if (size == data.Length)
+				{
+					Array.Resize<Ephemeron> (ref data, (int) newLength);
```
kumpera: You can't resize a hashtable without rehashing it.

kumpera: Additionally, Ephemeron arrays are special and must be registered with the runtime in order to produce the desired effect.

luhenry:
@kumpera could you please re-review?

lewurm:
maybe @BrzVlad can review it too

mcs/class/corlib/System.Runtime.CompilerServices/ConditionalWeakTable.cs
```diff
@@ -376,112 +376,21 @@ internal ICollection<TValue> Values
...
-				get
+				foreach (var eph in this.data) 
 				{
-					if (_currentIndex < 0)
-					{
-						ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumOpCantHappen();
-					}
-					return _current;
+					if (eph.key != null && eph.key != tombstone)
+						list.Add (new KeyValuePair<TKey, TValue> ((TKey) eph.key, (TValue) eph.value));
```
kumpera: This will produce a correct enumeration of key-value pair, but will break the conditional liveness offered by CWT.
IE, as long as the enumerator remains alive, nothing in the CWT will be collectible.

marek-safar:
@BrzVlad could you approve/reject the fix?

BrzVlad:
I'm not super comfortable with having a random enumerator object keep all objects in the table alive. Having said that, do we even need this change in the first place ? The documentation for IEnumerable says that the behavior is undefined if the table is changed and it is the responsibility of the caller to synchronize with the mutators

steveisok:
@BrzVlad should I reject the PR then?

BrzVlad:
I think we can close this for now.

PR #10711, opened 2018-09-20
[SDKS] Build host runtime with correct bitness

grendello:
When building 32-bit binaries on a 64-bit system and vice versa, it is necessary
to pass the `-m32` or `-m64` compiler flags, respectively. Failure to do so will
result not only in linking problems (different binary architecture) but also in
problems when using some system headers which are sensitive to the architecture.
For instance, `ucontext.h` contains register set definitions for different
architectures and building Mono runtime with a 64-bit compiler targetting 32-bit
architecture results in this error:

```
   .../mono/mono/utils/mono-sigcontext.h:173:74: error: ‘REG_EIP’ undeclared (first use in this function); did you mean ‘REG_RIP’?
      #define UCONTEXT_REG_EIP(ctx) (((ucontext_t*)(ctx))->uc_mcontext.gregs [REG_EIP])
```
The issue here is that the compiler defines the `__x86_64__` macro which causes
`ucontext.h` to choose declarations for amd64 (which does not use `REG_EIP` but
has `REG_RIP` instead) while Mono runtime assumes its building for a 32-bit host
and uses 32-bit register names. Passing `-m32` to the compiler will fix the
issue as now the compiler will define the `__i386__` macro and all parties will
agree where they are and what they want.

This commit uses a rather baroque statement when constructing compiler flags to
build the Mono runtime to check whether the host triple targets 32 or 64-bit
host and adds the appropriate flag.

### DIFF
--- a/sdks/builds/runtime.mk
+++ b/sdks/builds/runtime.mk
@@ -25,10 +25,20 @@
 #  _$(1)_PATH
 define RuntimeTemplate
 
-_runtime_$(1)_CFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CFLAGS) $$($(1)_CFLAGS)
-_runtime_$(1)_CXXFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CXXFLAGS) $$($(1)_CXXFLAGS)
-_runtime_$(1)_CPPFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CPPFLAGS) $$($(1)_CPPFLAGS)
-_runtime_$(1)_CXXCPPFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CXXCPPFLAGS) $$($(1)_CXXCPPFLAGS)
+_runtime_$(1)_BITNESS=$$(if $$(findstring i686,$(2)),-m32)
+
+ifeq ($$(_runtime_$(1)_BITNESS),)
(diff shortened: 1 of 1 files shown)


grendello:
@luhenry I see your point, however - is it a problem also when building iOS compilers? I'm asking because the cross compilers, for some reason, all build fine on macOS?

grendello:
@kumpera do we build cross compilers on ARM machines too?

sdks/builds/runtime.mk
```diff
@@ -25,10 +25,20 @@
 #  _$(1)_PATH
 define RuntimeTemplate
 
-_runtime_$(1)_CFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CFLAGS) $$($(1)_CFLAGS)
-_runtime_$(1)_CXXFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CXXFLAGS) $$($(1)_CXXFLAGS)
-_runtime_$(1)_CPPFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CPPFLAGS) $$($(1)_CPPFLAGS)
-_runtime_$(1)_CXXCPPFLAGS=$(if $(RELEASE),-O2 -g,-O0 -ggdb3 -fno-omit-frame-pointer) $$(_$(1)_CXXCPPFLAGS) $$($(1)_CXXCPPFLAGS)
+_runtime_$(1)_BITNESS=$$(if $$(findstring i686,$(2)),-m32)
```
jaykrell: Can/should all this be in configure.ac instead?

jaykrell:
Most likely,  can hardcode -m32 for all 32bit x86 hosts and -m64 for all amd64 hosts — no need to check if compiler accepts it, in configure.ac.

luhenry:

luhenry:

luhenry:

PR #10718, opened 2018-09-20
[mcs] Adds sourcelink option (as unsupported)

marek-safar:


lewurm:
shouldn't `mcs` at least print a warning that the option is not supported?

marek-safar:
nah, maybe in different PR ;-)

PR #10724, opened 2018-09-20
DO NOT MERGE - Alternative approach to OpenSSL TLS Provider

nealef:
A different approach to adding openSSL (#10700)  as a TLS provider. In this change-set:

- The existing structure is unchanged from the current BTLS implementation
- There is no separate Mono.OpenSSL directory and thus no copy of the contents of Mono.Btls with name changes
- mono/openssl contains all the OpenSSL calls with analogs in the mono/btls directory
- The APIs in both btls/openssl are called mono_tls_xxxx. These changes are made to the functions in mono/{btls,openssl} and all the c# source that references the backend routines
- MONO_FEATURE_TLS is defined instead of MONO_FEATURE_BTLS. It gets set when either HAVE_BTLS or HAVE_OPENSSL are defined. Similarly, a HAVE_TLS is defined if either of those previous symbols are defined
- To use the openSSL provider the &lt;dllmap&gt; directive is used to load the openssl shared library in lieu of the default btls dll/.so
- This means the provider is selectable at runtime

If this approach proves to be acceptable then Mono.Btls etc. could also be renamed to something more generic.

The danger of this approach is the assumption that BTLS APIs will remain relatively in-sync with OpenSSL. 

I have only given this code a cursory testing but am submitting it to contribute the discussion raised in #10700.


baulig:
I will give this a detailed review shortly, but really like this approach.  Thank you so much for refactoring this, I really appreciate all the hard work that you put into this!

nealef:
> I will give this a detailed review shortly, but really like this approach. Thank you so much for refactoring this, I really appreciate all the hard work that you put into this!

👍

nealef:
Mutually exclusive in what sense?

NattyNarwhal:
In the sense that you can't have both the BTLS and OpenSSL providers?

nealef:
Yes. It will default to btls but if you add the dllmap directive to the config you will use the OpenSSL provider. So it's a runtime decision that lasts for the duration of that process.

mcs/class/System/Mono.Net.Security/MonoTlsProviderFactory.cs
```diff
@@ -292,9 +292,9 @@ static void PopulateProviders ()
 #endif
 
 
-#if MONO_FEATURE_BTLS
+#if MONO_FEATURE_TLS
 		[MethodImpl (MethodImplOptions.InternalCall)]
-		internal extern static bool IsBtlsSupported ();
+		internal extern static bool IsTlsSupported ();
```
baulig: This is one of the examples where this rename becomes more than just a cosmetic issue - since this would completely change the meaning of what the function does, the new name is unacceptable in this context.

mcs/class/System/Makefile
```diff
@@ -85,11 +85,19 @@ ifdef ONLY_APPLETLS
 LIB_MCS_FLAGS += -d:ONLY_APPLETLS
 endif
 
+X=$(shell echo "Checking")
 ifndef PROFILE_DISABLE_BTLS
 ifdef HAVE_BTLS
```
baulig: Similarly to my comment to `MonoTlsProviderFactory`, this rename would alter meaning and should therefore be considered unacceptable.

configure.ac
```diff
@@ -5227,12 +5235,69 @@ if test "x$enable_btls" = "xyes"; then
 	AC_SUBST(BTLS_CMAKE_ARGS)
 
 	AC_DEFINE(HAVE_BTLS, 1, [BoringTls is supported])
+	AC_DEFINE(HAVE_TLS, 1, [Unix Tls is supported])
```
baulig: This needs to be either renamed into something like `HAVE_UNIX_TLS` or not renamed at all.  I don't have any strong preference on the new name (although I don't understand why it needs to be renamed at all) as long as it clearly reflects the meaning of the constant and does not clash with any "global" things.  Since the name `HAVE_TLS` would suggest to mean _all_ TLS implementations, it cannot be used in this context.

NattyNarwhal: Not to mention being confusable for thread-local storage...

configure.ac
```diff
@@ -5227,12 +5235,69 @@ if test "x$enable_btls" = "xyes"; then
...
+dnl ****************
+dnl ***  OpenSSL ***
+dnl ****************
+
+AC_ARG_ENABLE(openssl, [  --disable-openssl             Disable the OpenSSL provider], enable_openssl=$enableval, enable_openssl=$OPENSSL_SUPPORTED)
+
+AM_CONDITIONAL(OPENSSL, test x$enable_openssl = xyes)
+
+if test "x$enable_openssl" = "xyes"; then
+	AC_PATH_PROG(CMAKE, [cmake], [no], [$PATH:/Applications/CMake.app/Contents/bin:/usr/local/bin])
```
baulig: This conflicts with other existing cmake check (including the one for BTLS) and needs to be moved out of that conditional.

configure.ac
```diff
@@ -5227,12 +5235,69 @@ if test "x$enable_btls" = "xyes"; then
...
+	AC_SUBST(OPENSSL_ARCH)
+	AC_SUBST(OPENSSL_CFLAGS)
+	AC_SUBST(OPENSSL_LDFLAGS)
+	AC_SUBST(OPENSSL_PLATFORM)
+	AC_SUBST(OPENSSL_CMAKE_ARGS)
+
+	AC_DEFINE(HAVE_OPENSSL, 1, [OpenSSL is supported])
+	if test "x$enable_btls" = "xno"; then
+		AC_DEFINE(HAVE_TLS, 1, [Unix TLS is supported])
+	fi
```
baulig: See my other comment - we cannot use `HAVE_TLS` in this context.

nealef:
Are you suggesting it should be:

  *   isUnixTlsSupported (); or,
  *   We need both isBtlsSupported() and isOpenSSLSupported()?

In mcs/class/System/Mono.Net.Security/MonoTlsProviderFactory.cs<https://github.com/mono/mono/pull/10724#discussion_r221045947>:

>              [MethodImpl (MethodImplOptions.InternalCall)]

-              internal extern static bool IsBtlsSupported ();

+              internal extern static bool IsTlsSupported ();

This is one of the examples where this rename becomes more than just a cosmetic issue - since this would completely change the meaning of what the function does, the new name is unacceptable in this context.

nealef:
I am not sure what you mean. Does this section need moving to somewhere else in the file?

In configure.ac<https://github.com/mono/mono/pull/10724#discussion_r221049116>:

>  else

        enable_btls=no

 fi



 AM_CONDITIONAL(BTLS_ANDROID, test x$btls_android = xyes)



+dnl ****************

+dnl ***  OpenSSL ***

+dnl ****************

+

+AC_ARG_ENABLE(openssl, [  --disable-openssl             Disable the OpenSSL provider], enable_openssl=$enableval, enable_openssl=$OPENSSL_SUPPORTED)

+

+AM_CONDITIONAL(OPENSSL, test x$enable_openssl = xyes)

+

+if test "x$enable_openssl" = "xyes"; then

+       AC_PATH_PROG(CMAKE, [cmake], [no], [$PATH:/Applications/CMake.app/Contents/bin:/usr/local/bin])

This conflicts with other existing cmake check (including the one for BTLS) and needs to be moved out of that conditional.

baulig:
There is an existing check for `cmake` in the BTLS section of `configure.ac`.

We should do something like

    need_cmake=no

    ...

    if test "x$enable_btls" = "xyes"; then
        .....
        need_cmake=yes
    fi
    
    ....

    if test "x$enable_openssl" = "xyes"; then
        ....
        need_cmake=yes
    fi

    ...

    if test x$need_cmake = xyes; then
       	<check for cmake>
    fi

baulig:
After giving this a more detailed review, I also don't see any value in having the duplicate copy of the native files because they are pretty much all identical except for a few trivial differences.

Here's a detailed comparison:

- `x509-store-ctx`: trivial
- `x509-store`: trivial
- `x509-chain`: trivial, one `size_t` versus `ssize_t`.
- `x509-crl`: same
- `x509-lookup-mono`: identical
- `x509-lookup`: identical
- `x509-name`: identical
- `x509-revoked`: identical
- `x509-verify-param`: identical
- `time64` - identical; should probably move the other time-related function from `btls-util.c` in here.
- `pkcs12` - trivial `size_t` and refcount - what happened to `mono_btls_pkcs12_import` in the openssl version?
- `key` - identical
- `error` - identical
- `bio` - trivial
- `util`- put the new `CBS_*` and `BN_*` into a new file, then all that's left are trivial header changes.
- `btls-ssl.c` -  trivial; `STACK_OF`, `SSL_set_min_version()` / `SSL_set_max_version()`, different `mono_btls_ssl_set_renegotiate_mode()` and `mono_btls_ssl_renegotiate_pending()` implementations.
- `btls-ssl-ctx.c` - different `mono_btls_ssl_ctx_new()` implementation, trivial `mono_btls_ssl_ctx_free()` change, possibly broken `cert_verify_callback()` change (why are we silently accepting self-signed certificates here?), `SSL_CTX_set_min_version()` / `SSL_CTX_set_max_version()`, `STACK_OF` in `mono_btls_ssl_ctx_is_cipher_supported()`, different `mono_btls_ssl_ctx_set_ciphers()`.

Okay so to summarize - there is absolutely no need for a duplicate native implementation, this is only making it a lot more complicated to see what has actually changed while giving us no real benefit.  We should also keep the existing names in the native code.

For all files marked as "identical" above, the openssl version can just simply go away.  For "trivial", the openssl difference should be conditionally added to the btls file.

For a compile-time conditional, use `OPENSSL_IS_BORINGSSL` - this is the official `#define` from boringssl (defined in `include/openssl/base.h`).

This leaves us with the following differences:

- there are a few lines where we cast an `int` to `size_t` in one version and to `ssize_t` in the other - can we pick one for both?
- `openssl-util.c` has a bunch of `CBS_` and `BN_` functions added - did you write them or where did they come from?
- what about `SSL_set_min_version()` and friends - are those missing in openssl?
- the renegotiation changes are trivial enough to be done by a conditional.
- what is this code in `cert_verify_callback()` doing:

```
        ret = X509_verify_cert (storeCtx);
        if (ret != 1) {
                err = X509_STORE_CTX_get_error(storeCtx);
                if (err == X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN)
                        ret = 1;
        }
```

- `mono_btls_ssl_ctx_free` has `ctx->ctx = NULL;` added - I believe this should also be in the boringssl version.
- Regarding `mono_btls_ssl_ctx_is_cipher_supported()` and `mono_btls_ssl_ctx_set_ciphers()` - I planned to retire these in BTLS because the user shouldn't really interfere with cipher selection, so we do not really need to add an openssl version of these.
- In `mono_btls_ssl_ctx_new()`, we need to make sure that we have a reasonable list of ciphers and protocol versions.

baulig:
Is there any value in having a runtime switch between the two implementations or could we just make this a compile-time switch instead?

nealef:
Trivial at the moment but as with all (effectively) forked products the divergences will only increase. Given the choice of #ifdefing and two very similar code bases I’d opt for the latter.

.

steveisok:
@baulig Please re-visit and update where you think this is going.  @nealef @NattyNarwhal same thing.

PR #10740, opened 2018-09-21
[wasm] Add --regression suppport for wasm.

kumpera:


mono/mini/driver.c
```diff
@@ -350,6 +351,92 @@ domain_dump_native_code (MonoDomain *domain) {
...
+
+	if (strncmp (method->name, "test_", 5) != 0)
+		return FALSE;
+
+	if (interp) {
+		static gboolean filter_method_init = FALSE;
+		static const char *filter_method = NULL;
+
+		if (!filter_method_init) {
+			filter_method = g_getenv ("INTERP_FILTER_METHOD");
```
lewurm: (can be done later) we should rename it to `REGRESSION_FILTER_METHOD`. it's not really interpreter specific

kumpera: notes! I'll address that as I add support for running individual tests with wasm.

luhenry:

mono/mini/gshared.cs
```diff
@@ -1350,6 +1350,8 @@ class Class6 : IFace6 {
 
 	// FIXME: The runtime-invoke wrapper used by BeginInvoke is not found
 	[Category ("!FULLAOT")]
+	// FIXME: Wasm is single threaded and can't handle blocking waits
+	[Category ("!WASM")]
```
jaykrell: Another time/never: Should be a characteristic, not a specific platform: `[Category ("Threads")]` or Multithreaded or  HasBlockingWaits, etc.

mono/mini/driver.c
```diff
@@ -411,7 +411,7 @@ method_should_be_regression_tested (MonoMethod *method, gboolean interp)
 		if (!is_ok (error))
 			continue;
 
-		char *utf8_str = typed_args[0]; //this points into image memory that is constant
+		char *utf8_str = (char*)(void*)typed_args[0]; //this points into image memory that is constant
```
jaykrell: No need for the `(void*)` but ok.

PR #10750, opened 2018-09-22
[llvm] Remove an incorrect assert.

vargaz:


lewurm:
@monojenkins backport 2018-08

Issue #10598, opened 2018-09-13
Many scary "Your mono runtime and class libraries are out of sync" mono warnings when building for device


rolfbjarne:
_From @chamons on August 9, 2018 21:42_

### Steps to Reproduce

1. Install Alpha Channel
2. New Single View iOS App
3. Build for generic device and review output.

```
    /Applications/Xcode94.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang  -framework UIKit -framework Foundation -weak_framework CFNetwork -Xlinker -rpath -Xlinker @executable_path /Users/donblas/Projects/HelloIOS/obj/iPhone/Debug/mtouch-cache/arm64/Xamarin.iOS.dll.o /Library/Frameworks/Xamarin.iOS.framework/Versions/11.14.0.11/SDKs/MonoTouch.iphoneos.sdk/usr/lib/libmonosgen-2.0.dylib /Library/Frameworks/Xamarin.iOS.framework/Versions/11.14.0.11/SDKs/MonoTouch.iphoneos.sdk/usr/lib/libxamarin-debug.dylib -lz -liconv -gdwarf-2 -std=c99 -I/Library/Frameworks/Xamarin.iOS.framework/Versions/11.14.0.11/SDKs/MonoTouch.iphoneos.sdk/usr/include -isysroot /Applications/Xcode94.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.4.sdk -Qunused-arguments -miphoneos-version-min=11.4 -arch arm64 -shared -read_only_relocs suppress -install_name @rpath/libXamarin.iOS.dll.dylib -fapplication-extension -o /Users/donblas/Projects/HelloIOS/obj/iPhone/Debug/mtouch-cache/arm64/libXamarin.iOS.dll.dylib -DDEBUG
    Mono Ahead of Time compiler - compiling assembly /Users/donblas/Projects/HelloIOS/obj/iPhone/Debug/mtouch-cache/3-Build/mscorlib.dll
    AOTID 3585DBA3-DF9F-C621-1525-BBCEFD2C8102
    cant resolve internal call to "System.Runtime.Remoting.Proxies.RealProxy::InternalGetProxyType" (tested without signature also)
    
    Your mono runtime and class libraries are out of sync.
    The out of sync library is: /Users/donblas/Projects/HelloIOS/obj/iPhone/Debug/mtouch-cache/3-Build/mscorlib.dll
    
[...]

PR #10760, opened 2018-09-23
[runtime] Avoid warning about missing icalls when AOTing.

vargaz:
Fixes https://github.com/mono/mono/issues/10598.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -8291,7 +8291,7 @@ no_icall_table (void)
  * If the method is not found, warns and returns NULL.
  */
 gpointer
-mono_lookup_internal_call_full (MonoMethod *method, mono_bool *uses_handles)
+mono_lookup_internal_call_full (MonoMethod *method, gboolean warn_on_missing, mono_bool *uses_handles)
 {
 	char *sigstart;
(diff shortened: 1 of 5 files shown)


lewurm:
`warn_on_missing` is never true?

vargaz:
Fixed.

mono/metadata/icall.c
```diff
@@ -8291,7 +8291,7 @@ no_icall_table (void)
  * If the method is not found, warns and returns NULL.
  */
 gpointer
-mono_lookup_internal_call_full (MonoMethod *method, mono_bool *uses_handles)
+mono_lookup_internal_call_full2 (MonoMethod *method, gboolean warn_on_missing, mono_bool *uses_handles)
```
luhenry: Since `mono_lookup_internal_call_full` is now dead-code (it's not `MONO_API` and nothing else use it in the runtime AFAICT), why not simply adding the parameter to `mono_lookup_internal_call_full`?

vargaz:
It was in loader.h which is a public header. Is that just an oversight ?

marek-safar:
@monojenkins backport 2018-10

Issue #6555, opened 2018-01-17
Process.WaitForExit() gets slower the larger your open file limit (ulimit -n) is on Linux


akoeplinger:
## Steps to Reproduce

1. Compile this: `csc test.cs`:

```csharp
public static class Program {
    public static int Main () {
        for (int i = 0; i < 100; i++) {
            var p = System.Diagnostics.Process.Start ("echo", "hello");
            if (!p.WaitForExit (10000)) return 1;
        }
        return 0;
    }
}
```
2. `$ (ulimit -n 1000; time mono test.exe)`
```
...
real	0m0.402s
user	0m0.028s
sys	0m0.184s
```
3. `$ (ulimit -n 100000; time mono test.exe)`
```
...
real	0m1.277s
user	0m0.380s
sys	0m0.592s
```
4. `$ (ulimit -n 1000000; time mono test.exe)`
```
...
real	0m8.758s
user	0m3.424s
sys	0m4.424s
```

## Current Behavior

It gets slower the higher your ulimit -n is.

Note: you may need to change your system settings (`/etc/security/limits.conf`) to allow higher limits.

## Expected Behavior

Not getting slower.

### On which platforms did you notice this

[ ] macOS
[ X ] Linux, Ubuntu 14.04/16.04
[ ] Windows

**Version Used**: master, 5.8.088 and 4.2.1 (so either a very old bug or something outside our control)

This was the root cause behind an issue (https://github.com/mono/mono/pull/6537) we had on Jenkins because the Azure Linux VM builders had ulimit -n set to 1048576.

migueldeicaza:
This might be caused by the need of the child process to close the file descriptors on fork, to prevent the child from accessing the parent's descriptors.

What we could do is set close-on-exec flag on any file descriptor we open, to avoid paying this price at fork/exec time.

lewurm:
relevant code that Miguel mentioned:

https://github.com/mono/mono/blob/5839d7b43447e3ff6ca080fae0b03b56c67eb5e4/mono/metadata/w32process-unix.c#L1987-L1991

lewurm:
I should point out that the `strace` log mentioned by @akoeplinger in https://github.com/mono/mono/pull/6537#issuecomment-358150946 explicitly filtered out syscalls to `close`, because it spammed the log *a lot*, so yeah, that might be it 😅

luhenry:
I would definitely prefer the close-on-exec flag over what we have today. Let me work on a PR for that.

[...]

PR #10761, opened 2018-09-23
Mitigation for spawn FD closing taking forever with a big FD limit

NattyNarwhal:
On systems with a large file descriptor limit, Mono takes a very
long time to spawn a new process; with informal testing on the AIX
CI builder, (with a POWER7) it took ~30 minutes. This is obviously
very undesirable, but hand our is forced - libraries the user might
be calling could be creating non-CLOEXEC files we're unaware of. As
such, we started from the FD limit and worked our way down until we
hit stdio. Using APIs such as posix_spawn aren't desirable as we do
some fiddling with the child before we exec; and even then, closing
FDs with posix_spawn relies on non-standard file actions like
Solaris' addclosefrom not present on many systems. (All of this is
unnecessary on Windows, of course.)

This presents an alternative way (currently only implemented on AIX
but with notes how for other platforms) to try to close the child's
loose FDs before exec; by trying to get the highest number FD in
use, then work our way down. In the event we can't, we simply fall
back to the old logic.

See #6555 for a discussion and the initial problem being mitigated.

----

Basic testing seems to work; even with an "unlimited" FD limit, it works - the returned FD seems congruent with procfs, and it starts on a small number that takes no time at all. Implementing this for other systems would be quite nice; maybe I'll try to poke how for FreeBSD. However, as suggested in the comments, perhaps straight-up enumeration might also be desirable instead, in which case, the close logic would need a slight rewrite for that.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,8 @@ case "$host" in
 		with_sgen_default_concurrent=yes
 		;;
 	*-*-*freebsd*)
+		dnl For close_my_fds
+		LDFLAGS="$LDFLAGS -lutil"
 		if test "x$PTHREAD_CFLAGS" = "x"; then
 			CPPFLAGS="$CPPFLAGS -DGC_FREEBSD_THREADS"
 			libmono_cflags=
@@ -197,7 +199,7 @@ case "$host" in
 			LDFLAGS="$LDFLAGS $PTHREAD_LIBS -L/usr/local/lib"
 			libmono_ldflags="$PTHREAD_LIBS"
 		fi
-		CPPFLAGS="$CPPFLAGS -DHOST_BSD"
+		CPPFLAGS="$CPPFLAGS -DHOST_BSD -D_WITH_GETLINE"
 		need_link_unlink=yes
 		AC_DEFINE(PTHREAD_POINTER_ID, 1, [pthread is a pointer])
 		libdl=
--- a/mono/metadata/w32process-unix.c
+++ b/mono/metadata/w32process-unix.c
@@ -47,6 +47,11 @@
 #include <utime.h>
 #endif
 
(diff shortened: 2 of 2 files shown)


NattyNarwhal:
For my notes on implementing this on FreeBSD, I need to check if it returns the maximum number, but:

* https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/sys/sys/sysctl.h#L867 (see: that last MIB, plus the FD one)
* https://github.com/freebsd/freebsd/blob/master/sys/sys/user.h#L121 (the proc struct isn't seemingly too useful; the FD one might be)

Even if we can't simply get the maximum FD number in use, just having a table of the valid FDs could be useful to enumerate and just close those. (In which case, this would become a multi-strategy close later, depending on compile platform.)

luhenry:
@monojenkins squash

luhenry:
@monojenkins build failed

NattyNarwhal:
This does work as-is, but perhaps over the weekend I'll try to refactor it to attempt different strategies and implement the FreeBSD one:

* On some platforms, we can simply get a list of all FDs in use (most different code path, may require like, walking directories if Linux works how I think it does)

* On some platforms, get the whole FD table, and do what we do similar, but skip null FDs (similar to current code path)

* On some platforms, get the highest FD number in use and start enumeration from there, failing that, use the maximum possible (the current code path)

NattyNarwhal:
I've implemented a FreeBSD and Linux strategy. macOS is unimplemented as I lack a system to test with.

NattyNarwhal:
Unsure if failures are related; AArch64 was failing on a process one, but seems to be spurious threading related; amd64 interpreter is unrelated it seems

PR #10762, opened 2018-09-23
[profiler] Avoid native crash with GUID-less assemblies

ztzg:
Without this patch, Mono crashes while writing AOT profiling data at shutdown (when instructed to do so via `--profile=aot:output=out.aotprof`) if any of the sampled methods belong to a dynamic image, which doesn't have a GUID.

This happens with e.g. the `DynamicProxyGenAssembly2` image:

> DynamicProxyGenAssembly2 is a temporary assembly built by mocking systems that use CastleProxy like Moq or NSubsitute. It is generated when the mock is needed and disposed of after the tests are finished.

(cf. https://stackoverflow.com/a/17179923)

### DIFF
--- a/mono/profiler/aot.c
+++ b/mono/profiler/aot.c
@@ -242,6 +242,12 @@ add_image (MonoProfiler *prof, MonoImage *image)
 	if (id)
 		return id - 1;
 
+	// Dynamic images don't have a GUID set.  Moreover, we won't
+	// have a chance to AOT them.  (But perhaps they should be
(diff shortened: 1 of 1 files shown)


luhenry:
@monojenkins build

luhenry:
@vargaz can you please take a look?

luhenry:
@monojenkins squash

luhenry:
@monojenkins build failed

PR #10763, opened 2018-09-23
[WIP][corlib] Managed NumberToDouble and DoubleToNumber implementations from CoreCLR

EgorBo:
This PR brings fully managed implementation for number formatting & parsing from CoreCLR (recently merged, see https://github.com/dotnet/coreclr/pull/19999 and https://github.com/dotnet/coreclr/pull/20080). We recently switched to CoreRT implementation based on C# + icall to `snprintf`/`ecvt_s` but it turned out it's quite slow. Here is a couple of benchmarks from **dotnet/performance** repo to see the difference between our current implementation and this fully-managed one:
image
_(TODO: double.Parse and LLVM results)_
PS: I copied files from dotnet/coreclr to "coreclr" folder but since these files were moved to shared in coreclr I guess I can move them to mono/corefx


mcs/class/corlib/Test/System/DoubleTest.cs
```diff
@@ -470,9 +470,9 @@ public void CustomFormatHelper ()
 			list.Add (new Element (-1234d, "0.00", "-1234.00"));
 			list.Add (new Element (-1234d, "00.00", "-1234.00"));
 			list.Add (new Element (-1234d, "#,#", "-1,234"));
-			list.Add (new Element (-1234d, "#,,", ""));
-			list.Add (new Element (-1234d, "#,,,", ""));
-			list.Add (new Element (-1234d, "#,##0,,", "0"));
+			list.Add (new Element (-1234d, "#,,", "-"));
```
EgorBo: smells like a bug? https://github.com/dotnet/coreclr/pull/20080#issuecomment-423836133

PR #10765, opened 2018-09-24
mono-complex cleanup: remove macros and rename

jaykrell:


luhenry:
What's the reason for this change?

jaykrell:
When I first ported the complex math to C++, the macros to rename were decided against. 

But the line damage was a bit much. 

This follows through with that decision. Perhaps just do that.

jaykrell:
Otherwise it is to follow mono naming style.

jaykrell:
@monojenkins build failed

PR #10764, opened 2018-09-24
[cxx] Rework C++ linkage to C99 trunc, isnan, isinf, etc. and some cleanup.

jaykrell:
Android was giving quite some problems otherwise, and WebAssembly less so.

PR #10766, opened 2018-09-24
Remove hacks.h.

jaykrell:
Looks kinda no-opish, except isunordered.
These ifdef/define games confuse things because these are functions in C++, not macros.
The only likely difference is HPUX64 + non-gcc.
Which would be IA64 or HPPA64.
The C99 enabling is likely redundant with compiling for C99.
Part of https://github.com/mono/mono/pull/10764.


jaykrell:

jaykrell:
There was one alive looking part, isordered, for 64bit HP-UX, when not using gcc. Which would imply HPPA64 or IA64. So ok to delete for now and probably forever?

jaykrell:
Also the define C99 could be kept but probably not needed. I guessing it dates to not compiling overall as C99.

lewurm:
it's a leftover from 2001.

PR #10571, opened 2018-09-12
[Facades] Adds System.Buffers mobile facade

marek-safar:
and sets up facades without corresponding nuget _._ for conflict
resolution done by msbuild

PR #10771, opened 2018-09-24
Add System.Buffers facade to framework assemblies list

akoeplinger:
It's a new facade added in https://github.com/mono/mono/pull/10571


marek-safar:
I am not sure we should do it, it required non-facade handling for !MOBILE

akoeplinger:
It won't have an effect if said assembly isn't part of the shipped framework facades, I checked this with @lambdageek when I did https://github.com/mono/mono/pull/10214.

marek-safar:
Why do we need it then? Mobile does not have GAC so it's no-op there as well

akoeplinger:
Because we'll forget to add it to this list once we enable it for !MOBILE in a few months down the road :)

PR #10778, opened 2018-09-24
[sdks] Add a 64 bit arm32 cross compiler build.

vargaz:
Part of #9621


luhenry:
Is that in the hope of replacing `cross32`? If so shouldn't we more simply replace `cross32` to be built as a 64bit executable?

vargaz:
We need the two to coexist for a while to make it easier to track down regressions and to allow users to fall back to the old one in case of a regression.

PR #10781, opened 2018-09-24
Update roslyn automated tests

kg:
Roslyn submodule was updated to enable more automated tests. This bumps our submodule revision, and uses roslyn's script to run tests instead of ours.

@monojenkins skip

Part of #6991

### DIFF
--- a/acceptance-tests/SUBMODULES.json
+++ b/acceptance-tests/SUBMODULES.json
@@ -2,7 +2,7 @@
   {
     "name": "roslyn", 
     "url": "git://github.com/kg/roslyn.git",
-    "rev": "4e8e106956c62501cb2248e8610e47fe96a36cd0",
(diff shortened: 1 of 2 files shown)


acceptance-tests/SUBMODULES.json
```diff
@@ -1,8 +1,8 @@
 [
   {
     "name": "roslyn", 
-    "url": "git://github.com/mono/roslyn.git",
-    "rev": "380eec412868464a132cdcd3a45ef457fab6e060",
+    "url": "git@github.com:mono/roslyn.git",
```
akoeplinger: please revert this back to git:// so it can be cloned anonymously.

kg: Ohhh, so that's the difference. Thanks, will do

PR #10790, opened 2018-09-25
Fix parameter problem in test_op_il_seq_point.sh.

lateralusX:
Looks like: https://github.com/mono/mono/commit/e8029e1474b701184840f6e0abea51966d048831 caused issues in the parameters for this test since caller was not updated accordingly.

This fixes so that the caller pass needed additional argument + simplify the AOT use case for Windows since different tested versions will need different toolchain versions that in turn will cause additional complexity. Since the test doesn't validate the compile/link of AOT module (and don't handle errors)
anyways, Windows versions of this test will just generate assembler output.

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -808,8 +808,8 @@ check-seq-points:
 else
 check-seq-points: mono $(regtests)
 	rm -f TestResult-op_il_seq_point.xml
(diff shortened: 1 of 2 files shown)


lateralusX:
Looks like failure in Linux ARMv5 is occurring for several other PR's and not related to this change.

PR #10809, opened 2018-09-25
[System] Add FSEvent FileSystemWatcher to 'monotouch' BCL profile

alexischr:


alexischr:
@monojenkins build failed

alexischr:
@monojenkins build failed

marek-safar:
@monojenkins build failed

marek-safar:
@baulig this just for master so it should be safe to land, it won't interfere with existing branches

marek-safar:
@alexischr please rebase it

alexischr:
@monojenkins build failed

alexischr:
@monojenkins build failed

luhenry:
@alexischr it's not benign on master since we do build and run tests on iOS. Also just postponing the native work until it bits us in a release branch integration is exactly what's already slowing down the `2018-10` integration. So let's not repeat the same mistake, and make sure the native bits are coming right along the managed bits.

luhenry:
Just discussed offline: right now, the FSW throws `PlatformNotSupportedException` on `monotouch`, so this PR is actually going to bring us closer to getting it to work, rather than regressing. So :+1: on my end. Thanks for the clarification @alexischr

alexischr:
@monojenkins commit csproj

alexischr:
Failures are on `upload-to-sentry` step (cc @alexanderkyte)

PR #10812, opened 2018-09-26
[bcl] Mark ActivationServices.CreateProxyForType () with 'remoting' in the linker descriptor.

vargaz:


marek-safar:
I think we should also update runtime to exclude `create_proxy_for_type_method` field for `#ifndef DISABLE_REMOTING`

vargaz:
@monojenkins build failed

PR #10817, opened 2018-09-26
[sdks] Debian Linux doesn't need to build MXE

grendello:
**When/if this PR is accepted, could we backport it to the `2018-{04,06,08}` branches as well? It fixes XA build on our PR bots, thanks!**

Debian and its derivatives include full mingw chain (both 64 and 32-bit) and so
they don't need to build the toolchain at all. However, the current detection of
Linux flavor takes into account only Ubuntu and all other distros are configured
to *require* MXE but they do *not* build them.

This commit fixes detection of distros which provide the mingw toolchain in a
way that makes it easy to extend the set of compatible distributions as needed.
In order to mark a distribution as one with MinGW one has to add the `:Distro:`
string to the `LINUX_WITH_MINGW` make variable in `mxe.mk`. The `Distro` string
is whatever the `determine-linux-flavor.sh` script returns.

### DIFF
--- a/sdks/builds/mxe.mk
+++ b/sdks/builds/mxe.mk
@@ -3,7 +3,10 @@ ifeq ($(UNAME),Linux)
 LINUX_FLAVOR=$(shell ./determine-linux-flavor.sh)
 endif
 
-ifeq ($(LINUX_FLAVOR),Ubuntu)
+LINUX_WITH_MINGW=:Ubuntu:,:Debian:
+LINUX_HAS_MINGW=$(if $(findstring :$(LINUX_FLAVOR):,$(LINUX_WITH_MINGW)),yes)
+
+ifeq ($(LINUX_HAS_MINGW),yes)
 MXE_PREFIX=/usr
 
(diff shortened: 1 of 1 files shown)


luhenry:

luhenry:

luhenry:

directhex:
The coop failure was:

```
1) StackTraceInNative (MonoTests.DebuggerTests.StackTraceInNative)
   TearDown : Mono.Debugger.Soft.VMDisconnectedException : Exception of type 'Mono.Debugger.Soft.VMDisconnectedException' was thrown.
--TearDown
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /mnt/jenkins/workspace/test-mono-pull-request-coop/mcs/class/corlib/System.Reflection/MonoMethod.cs:324 
```

Not related.

PR #9873, opened 2018-08-03
[build] Switch to netstandard based xunit

marek-safar:
/cc @baulig

PR #10835, opened 2018-09-26
Add Mono.Android facade

EgorBo:
Since xunit tests (e.g. monodroid_corlib_xunit.dll) requires netstandard lib now (see https://github.com/mono/mono/pull/9873) and the netstandard requires Mono.Android on monodroid profile (Mono.Android implements System.Drawing) I am adding this dummy facade in order to successfully compile mono-runtimes on XA.

### DIFF
--- a/mcs/class/Facades/Mono.Android/AssemblyInfo.cs
+++ b/mcs/class/Facades/Mono.Android/AssemblyInfo.cs
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2013 Xamarin Inc. (http://www.xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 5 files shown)


EgorBo:
@monojenkins backport 2018-08

EgorBo:
^ in order to test XA-2018-08

marek-safar:
Let's fix this properly
- move Mono.Android to https://github.com/mono/reference-assemblies/tree/master/build/monodroid 
- use correct assembly version and public key
- add type skeletons only for types we need (should be just few)
- mark the assembly [ReferenceAssembly]
- Remove all EXTERNAL_FACADE_DRAWING_REFERENCE hacks for monodroid in the build
- Update netstandard Makefile to always reference Mono.Android reference for monodroid profile

PR #10857, opened 2018-09-27
[wip] Do not merge; C++ coop handle prototype.

jaykrell:
This is expected to fail all C lanes, and even some C++ lanes (Linux/32bit x86).

The gist of this is to change types but not line by line code.

That is, given:

```
MonoFoo* p;
p->foo
```

where p->foo could be a read, or a write, of a "value" or a managed pointer, the "only" changes should be:
1. Change p to be `MonoFooHandle` or `MonoHandle<MonoFoo>`.
2. `p.New (...)` (or historical `MONO_HANDLE_NEW` is ok).
3. Change managed pointers within p to be MonoPtr() (or other name).
   Rewriting the handle type is not sufficient for significant gains.
   Continuing to require "all the macros" seems not a big win.
   i.e. only enabling early return, or removing type parameters from the macros,
   are other lesser goals.
   This PR shows a significant but incomplete amount of this.
   As this is applied, non-coop code needs GetRaw sprinkled in, or conversion,
   or cover it up by adding silent conversion in MonoPtr -- remove that to
   find places that need coop work.
4. Add `HANDLE_FUNCTION_ENTER`, same as before.
5. Watch out for loops, same as before.
6. `MONO_RETURN_HANDLE` (or its predecessor `HANDLE_FUNCTION_RETURN_REF`).
   i.e. allocation in higher frame remains not easily covered up
7. Functions that return non-handles can just use regular return, no macro.
   Early return works, for handles and non-handles.
8. Add overloads of various functions, as before, that accept handles, and either pass them to
   through as raw, or invert/rewrite -- same/similar to before.
   This same as before, not huge, perhaps made easier now.

Previously what was also needed was:
9. No early return in functions with frames. Return always with macros.
10. Change every dereference of p.

Both of which are significant.

As well, not covered here, is more esoteric handle uses, like arrays and MonoClassField.
Only the most common:
	`MONO_HANDLE_GETVAL`
	`MONO_HANDLE_SETVAL`
	`MONO_HANDLE_SET`
	`MONO_HANDLE_IS_NULL` (and new `MONO_HANDLE_BOOL`)
	`MONO_HANDLE_GET`

become trivial, so far.

All preexisting macros work. All prior coop-friendly code requires no revisit,
though changes could be dialed back.

While these macros are written to be trivial, the old forms also work.

Typesafe handles are required, and this should be not difficult to fix.
The marshaling should change to marshal a struct containing a pointer.

The underlying representation and runtime remains unchanged and fairly directly reused.
There is room for improvement here, i.e. handles should be in frames, not heap,
but this is a separate concern.

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -867,8 +867,8 @@ GUnicodeBreakType   g_unichar_break_type (gunichar c);
 #endif
 
 #if defined(__GNUC__) && (__GNUC__ > 2)
-#define G_LIKELY(expr) (__builtin_expect ((expr) != 0, 1))
-#define G_UNLIKELY(expr) (__builtin_expect ((expr) != 0, 0))
+#define G_LIKELY(expr) (__builtin_expect (!!(expr), 1))
+#define G_UNLIKELY(expr) (__builtin_expect (!!(expr), 0))
 #else
 #define G_LIKELY(x) (x)
 #define G_UNLIKELY(x) (x)
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -1312,7 +1312,9 @@ struct g_ptr
 {
 	T* p;
 
-	void detach () { T* q = p; p = 0; return q; }
+	T* get() { return p; } // e.g. printf ("%p");
+
+	T* detach () { T* q = p; p = 0; return q; }
 
 	void cleanup () { g_free (detach ()); }
 
@@ -1320,13 +1322,15 @@ struct g_ptr
 
 	~g_ptr () { cleanup (); }
 
-	operator T* () { return p; }
+	operator T* () { return get (); }
+
+	g_ptr (T* q = 0) : p (q) { }
 
 	g_ptr& operator = (T* q)
 	{
 		if (p == q)
 			return *this;
-		cleanup();
+		cleanup ();
 		p = q;
 		return *this;
 	}
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -17,6 +17,14 @@
 #endif
 #endif // __cplusplus
 
+#ifdef __GNUC__
+#define G_ALWAYS_INLINE __attribute__ ((__always_inline__))
+#elif defined (_MSC_VER)
+#define G_ALWAYS_INLINE __forceinline
+#else
+#define G_ALWAYS_INLINE
+#endif
+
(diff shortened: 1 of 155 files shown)


mono/metadata/handle-decl.h
```diff
@@ -88,16 +161,237 @@ Handle macros/functions
...
+
+	// hopefully used sparingly, but e.g. printf ("%p", x.get ());
+	// printf ("%p", x) will work on some ABIs/compilers but probably not all.
+	T* get () { return p; }
+
+	//FIXME This is used in non-coop-converted code.
+	// It might also be a good printf("%p") piece.
+	T* GetRaw() { return p; }
+
+	struct OperatorAmpersandResult
```
jaykrell: I thought I had removed this. Try removing it.

mono/metadata/handle-decl.h
```diff
@@ -88,16 +161,237 @@ Handle macros/functions
...
+
+/******************************************************************************************
+This is important.
+******************************************************************************************/
+	explicit operator bool () const { return !!p; }
+	MonoHandle<T> operator -> () { return NewHandle<T> (); }
+
+	//MonoPtr& operator = (std::nullptr_t) { p = 0; return *this; }
+	MonoPtr& operator = (MonoPtr<T> q) { p = q.p; return *this; }
+	MonoPtr& operator = (T* q) { p = q; return *this; }
```
jaykrell: Barriers missing here, i.e. use `MONO_OBJECT_SETREF` here.

mono/eglib/glib.h
```diff
@@ -1305,4 +1313,39 @@ G_END_DECLS
 #define g_try_realloc(obj, size) (g_cast (monoeg_try_realloc ((obj), (size))))
 #define g_memdup(mem, size) (g_cast (monoeg_g_memdup ((mem), (size))))
 
+#ifdef __cplusplus
+
+template <typename T>
+struct g_ptr
```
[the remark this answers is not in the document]
jaykrell: Even if the size didn't match, for local variables, ok.
There not likely any runtime dependency, ok.
There is a C++11 dependency, granted.
It is good and bad to reveal how to write easy templates within mono.
This is another form of template to show how to write, granted.

jaykrell:
Closing this as github is confused and not keeping up.

jaykrell:
https://github.com/mono/mono/pull/10867

PR #10860, opened 2018-09-27
[corlib] Import System.Text.Encoding from CoreFX

alexischr:
Merge with https://github.com/mono/corefx/pull/142

Part of #7247


NattyNarwhal:
I've become a little antsy around imports like this due to CoreFX endianness handling being compile-time (yet the define isn't defined?) which badly breaks monolite-on-BE scenarios; I did a skim, however, and it does seem alright, the only endian behaviour I can see is using LE by default for UTF-16/32.

PR #10867, opened 2018-09-27
[cxx][coop] work in progress do not merge prototyping

jaykrell:
This is expected to fail all C lanes, and even some C++ lanes (Linux/32bit x86).

The gist of this is to change types but not line by line code.

That is, given:

```
MonoFoo* p;
p->foo
```

where p->foo could be a read, or a write, of a "value" or a managed pointer, the "only" changes should be:
1. Change p to be `MonoFooHandle` or `MonoHandle<MonoFoo>`.
2. `p.New (...)` (or historical `MONO_HANDLE_NEW` is ok).
3. Change managed pointers within p to be MonoPtr() (or other name).
   Rewriting the handle type is not sufficient for significant gains.
   Continuing to require "all the macros" seems not a big win.
   i.e. only enabling early return, or removing type parameters from the macros,
   are other lesser goals.
   This PR shows a significant but incomplete amount of this.
   As this is applied, non-coop code needs GetRaw sprinkled in, or conversion,
   or cover it up by adding silent conversion in MonoPtr -- remove that to
   find places that need coop work.
4. Add `HANDLE_FUNCTION_ENTER`, same as before.
5. Watch out for loops, same as before.
6. `MONO_RETURN_HANDLE` (or its predecessor `HANDLE_FUNCTION_RETURN_REF`).
   i.e. allocation in higher frame remains not easily covered up
7. Functions that return non-handles can just use regular return, no macro.
   Early return works, for handles and non-handles.
8. Add overloads of various functions, as before, that accept handles, and either pass them to
   through as raw, or invert/rewrite -- same/similar to before.
   This same as before, not huge, perhaps made easier now.

Previously what was also needed was:
9. No early return in functions with frames. Return always with macros.
10. Change every dereference of p.

Both of which are significant.

As well, not covered here, is more esoteric handle uses, like arrays and MonoClassField.
Only the most common:
	`MONO_HANDLE_GETVAL`
	`MONO_HANDLE_SETVAL`
	`MONO_HANDLE_SET`
	`MONO_HANDLE_IS_NULL` (and new `MONO_HANDLE_BOOL`)
	`MONO_HANDLE_GET`

become trivial, so far.

All preexisting macros work. All prior coop-friendly code requires no revisit,
though changes could be dialed back.

While these macros are written to be trivial, the old forms also work.

Typesafe handles are required, and this should be not difficult to fix.
The marshaling should change to marshal a struct containing a pointer.

The underlying representation and runtime remains unchanged and fairly directly reused.
There is room for improvement here, i.e. handles should be in frames, not heap,
but this is a separate concern.

===

So, the above was achieved, however things like `a->b->c` would quietly create handles.
As would `f (handle->managed)`.
This was deemed dubious for multiple reasons.
  1. Hides cost.
  2. Hides handle creation within loops and within functions w/o their own frames.

So a slightly less automatic programming model is now provided.
In this model `managed->value` read/write is still friendly, but `managed->managed` will error, forcing programmer to write either `managed->managed.GetRaw ()` to keep existing code working asis coop-unfriendly, or `managed->managed.NewHandle ()`. Programmer can decide to store the handle for reuse, and to place `HANDLE_FUNCTION_ENTER` as needed. Note that `HANDLE_FUNCTION_ENTER` is now about scopes, not functions, so can be placed within loops, multiple times in nested loops, etc. -- with care taken not to store a handle outsides its scope.

The new model depends on, similar to the earlier, "annotation" of managed fields with what is presently called `MonoPtr()`. It might not be named well. For example, it doesn't mean "free with mono_free". It really means `ManagedPointerInStruct` or `GCAlreadySeesThisRootButNotIfYouCopyFromIt`.

Much such annotation is present here, but not all.

As the annotation is added, existing code is deliberately broken, and typically `GetRaw()` or `NewHandle()` is inserted.

Line by line visitation of code is still required, however the required edits are significantly easier and smaller and more idiomatic. C++ overloads are also brought to bear to limit edits -- functions taking pointers and handles can have the same name.

There are a few other niceties now provided.
Comparison to null and other pointers is macro-free and handle-free.
You can say:
`if (handle->managed)`

and `if (handle->managed == NULL)`

and if, if not already, should allow `if (handle1->managed1 != handle2->managed2)`

Things like array length and maybe access are more convenient, one can go from:

`mono_array_length (pointer->array)` (or `mono_array_length_fast`)

right to:
`mono_array_length (handle->array)`

===

I have a C++ prototype of coop handles where you basically don't need the macros.

You can just say handle->field.
None of the get/set stuff.

However it has some caveats:
 1. Requires typesafe handles.
 2. Requires coop handles be pinned.
 3. Passes null as the object start to the write barrier, which isn't used presently.

Viable or too flawed?

There are smaller improvements to ease coop conversion possible, w/o those caveats.
For example, early/multiple return could be allowed. As well as handle creation in loops (same thing).
MONO_HANDLE_NEW wouldn't need the type parameter.
We could use overloads to avoid having to rename every function call.

I also accidentally made a way to get compiler errors for any raw pointer use,
instead of depending on separate static analysis.

### DIFF
--- a/mono/metadata/sre.c
+++ b/mono/metadata/sre.c
@@ -50,7 +50,7 @@
 static GENERATE_GET_CLASS_WITH_CACHE (marshal_as_attribute, "System.Runtime.InteropServices", "MarshalAsAttribute");
 static GENERATE_GET_CLASS_WITH_CACHE (module_builder, "System.Reflection.Emit", "ModuleBuilder");
 
-static char* string_to_utf8_image_raw (MonoImage *image, MonoString *s, MonoError *error);
+static char* string_to_utf8_image_raw (MonoImage *image, MonoPtr <MonoString>& s, MonoError *error);
 
 #ifndef DISABLE_REFLECTION_EMIT
 static guint32 mono_image_get_sighelper_token (MonoDynamicImage *assembly, MonoReflectionSigHelperHandle helper, MonoError *error);
@@ -96,15 +96,11 @@ mono_reflection_emit_init (void)
 }
 
 char*
-string_to_utf8_image_raw (MonoImage *image, MonoString *s_raw, MonoError *error)
+string_to_utf8_image_raw (MonoImage *image, MonoPtr<MonoString>& s, MonoError *error)
 {
-	/* FIXME all callers to string_to_utf8_image_raw should use handles */
 	HANDLE_FUNCTION_ENTER ();
-	char* result = NULL;
 	error_init (error);
-	MONO_HANDLE_DCL (MonoString, s);
-	result = mono_string_to_utf8_image (image, s, error);
-	HANDLE_FUNCTION_RETURN_VAL (result);
+	return mono_string_to_utf8_image (image, s.NewHandle (), error);
 }
 
 static char*
@@ -193,11 +189,7 @@ static char*
 image_strdup (MonoImage *image, const char *s)
 {
 	MONO_REQ_GC_NEUTRAL_MODE;
-
-	if (image)
-		return mono_image_strdup (image, s);
-	else
-		return g_strdup (s);
+	return image ? mono_image_strdup (image, s) : g_strdup (s);
 }
 #endif
 
@@ -479,7 +471,7 @@ mono_reflection_methodbuilder_from_method_builder (ReflectionMethodBuilder *rmb,
 	memset (rmb, 0, sizeof (ReflectionMethodBuilder));
 
 	rmb->ilgen = mb->ilgen;
-	rmb->rtype = (MonoReflectionType*)mb->rtype;
+	rmb->rtype = (MonoReflectionType*)mb->rtype.GetRaw ();
 	return_val_if_nok (error, FALSE);
 	rmb->parameters = mb->parameters;
 	rmb->generic_params = mb->generic_params;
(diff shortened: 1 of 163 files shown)


mono/metadata/sre.c
```diff
@@ -96,15 +96,11 @@ mono_reflection_emit_init (void)
...
+string_to_utf8_image_raw (MonoImage *image, MonoPtr<MonoString>& s, MonoError *error)
 {
-	/* FIXME all callers to string_to_utf8_image_raw should use handles */
 	HANDLE_FUNCTION_ENTER ();
-	char* result = NULL;
 	error_init (error);
-	MONO_HANDLE_DCL (MonoString, s);
-	result = mono_string_to_utf8_image (image, s, error);
-	HANDLE_FUNCTION_RETURN_VAL (result);
+	return mono_string_to_utf8_image (image, s.NewHandle (), error);
```
jaykrell: It counts for something when a function drops from 6 lines to 3, with no loss of clarity. It is still 3 too many though.

luhenry:
@monojenkins build

jaykrell:
Some cleanup coming, was going to rebuild soon.

jaykrell:
Relevant failure

*** MonoTests.System.AppDomainTest.Load_EmptyString
***** MonoTests.System.AppDomainTest.Load_PartialVersion
* Assertion at handle.c:341, condition `mono_object_domain (elem->o) != domain' not met
The

jaykrell:
@monojenkins build Linux x64

jaykrell:

jaykrell:
Windows has build breaks, that can be fixed, if going forward.
Proposed fix is move the coop wrappers from IL to C, so that IL/C++ interface remains C and extern "C".

PR #10873, opened 2018-09-28
[tests] Reenable bug-80392.2.exe

akoeplinger:
According to @kumpera it should be fixed by https://github.com/mono/mono/pull/10793.


akoeplinger:
Running a few builds to make sure it works.

@monojenkins build OS X x64
@monojenkins build OS X x86

Issue #9729, opened 2018-07-25
Proposal: environment variable + configure flag to gate experiments


lambdageek:
## Summary ##

Let's add `MONO_EXPERIMENTS=*comma-separated-keywords*` and `./configure --with-experiments=*command-separated-keywords*` which have no stability/backward compatibility or deprecation guarantees.

## Motivation ##

We have a lot of development going on in Mono on features that could use regular testing on CI, or with brave customers, but which are not ready for prime time.

We've occasionally gated these features with one-off environment variables: `MONO_ENABLE_COOP` was one, for example.  What we've done in the past is add the variable and then document it in the man page.  And it's been understood that customers may begin to rely on the environment variable, so retiring the environment variable or changing its semantics requires ample notification and a period of transition.

But we should test earlier - before we're ready to commit to supporting something.

Other features only make sense on some platforms, but ones that are harder for us to test.  For example #9666: we only really need to copy stack data on watchOS and tvOS, but we don't test Mono on those platforms in nightly CI, so it would be useful to periodically run a CI configuration on Linux/amd64 or some other common platform for sanity checking.

## Proposal ##

We add an environment variable and a configure flag that can be used to selectively enable experiments.  The variable and flag will have no stability or deprecation guarantees.

The environment variable and configure flag should use the same keywords and the environment variable should take priority over the configure flags. 

(Optional: maybe every flag also has a "no-flag" in the environment variable to turn _off_ something that the configure flag turned on?)

Each experiment flag should have an open GitHub Issue associated with it.  The issue is closed when the experiment is retired.

jaykrell:
Agreed, except, where affordable, either:
1. Runtime ifs not preprocessor ifs, so we can test one binary with runtime switches.
2. Or maybe a compromise, just one preprocessor if to gate all experimental code, and then runtime switches to enable/disable individual experiments?

I really don't like to configure and build so many times, and imagine a customer is willing to experiment. He shouldn't have to reconfigure/rebuild.

I realize it is messy -- anything here is messy. Either a mess of pound-if or regular-if. Configure does have an advantage of tunnelling through settings most automatically, i.e. if can't place the command line switch in the pipeline. So, lots of options make sense and have advantages and disadvantages.

There is also risk of runtime-if polluting mainline behavior, not just bloat.

Another idea is to make it cheap/easy to duplicate CI for an alternative branch, including all the way to debian packages and monodevelop. Then experiments would go in branches instead of mainline.
[...]

Issue #10877, opened 2018-09-28
Disable experiment: null


lambdageek:
This is a tracking issue to disable the Mono experiment `"null"`.

----

The `"null"` experiment is just an example.  This issue is an example of what a tracking issue for an experiment may look like - each experiment should have an associated issue that will be closed when the experiment is removed.

There are generally three possible fates for an experiment:
1. The experiment becomes part of mainline Mono
2. The experiment is changed into an optional feature guarded by a regular `configure` flag or runtime setting.
3. The experiment is removed from Mono.

jaykrell:
2b. The experiment changes into a runtime setting via command line or environment variable.

PR #10878, opened 2018-09-28
Add ability to add experiments without backward compatibility or support guarantees

lambdageek:
Implement the idea from #9729

* A `configure` flag `--enable-experiment=LIST` to define compile-time constants for experiments
* Support for a `MONO_EXPERIMENT=LIST` environment variable and `mono_experiment_enabled` test function.

The idea is that experiments are experimental, unsupported and are not guaranteed to be backward compatible.  For small experiments a compile time check is enough.  Larger experiments can live behind the feature flag.

Each experiment also lists a GitHub tracking issue to disable and remove the experiment - either the experiment graduates to a mainline feature or it becomes gated by a regular `configure` flag or some other opt-in mechanism, or it is dropped.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5043,6 +5043,26 @@ AM_CONDITIONAL([ENABLE_HYBRID_SUSPEND], [test x$enable_hybrid_suspend != xno])
 
 dnl End of thread suspend policy
 
+dnl ***************************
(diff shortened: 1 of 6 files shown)


lambdageek:
See #10877 for an example of an issue tracking an experiment.

PR #10881, opened 2018-09-28
[cxx]Fix pointer truncation errors in cross builds.

jaykrell:
We have *apparent to compilers* but not actual pointer truncations.

They error in cross C++ builds when compiler host/pointers is 64bit and target/mgreg_t is 32bit,  which is a CI scenario (just in the past week?). These also build a runtime, which isn’t run.

1. Selectively change mgreg_t to host_mgreg_t? Reasonable, I had it, but a bit large/noisy.  I got scared of the codegen interaction, but that is probably nothing  actually.

2. ifndef MONO_CROSS_COMPILE? Reasonable, similar in size to option 3.
We do this generally but miss some. Inhibits a multi-configure, but there are already
many barriers to that.

3. Insert casts, i.e. pointers to/from gsize/gssize. Redundant but still correct if 1 or 2 are done. Clearly no-op on non-cross builds, and very safe.

4. Maybe a compiler flag, I didn’t check.

Option 3 casts is this PR.

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -2905,7 +2905,7 @@ decode_llvm_mono_eh_frame (MonoAotModule *amodule, MonoDomain *domain, MonoJitIn
 
 	if (is_thumb_code (amodule, code_start))
 		/* Clear thumb flag */
-		code_start = (guint8*)(((mgreg_t)code_start) & ~1);
+		code_start = (guint8*)(((gssize)code_start) & ~1);
 
 	fde = amodule->mono_eh_frame + table [(pos * 2) + 1];	
 	/* This won't overflow because there is +1 entry in the table */
@@ -2976,10 +2976,10 @@ decode_llvm_mono_eh_frame (MonoAotModule *amodule, MonoDomain *domain, MonoJitIn
 		jei->clause_index = clause_index;
 
 		if (is_thumb_code (amodule, (guint8 *)jei->try_start)) {
-			jei->try_start = (void*)((mgreg_t)jei->try_start & ~1);
-			jei->try_end = (void*)((mgreg_t)jei->try_end & ~1);
+			jei->try_start = (void*)((gssize)jei->try_start & ~1);
+			jei->try_end = (void*)((gssize)jei->try_end & ~1);
 			/* Make sure we transition to thumb when a handler starts */
-			jei->handler_start = (void*)((mgreg_t)jei->handler_start + 1);
+			jei->handler_start = (void*)((gssize)jei->handler_start + 1);
 		}
 	}
 
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -4975,7 +4975,7 @@ find_aot_module (guint8 *code)
 }
 
 void
-mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, mgreg_t *regs, guint8 *addr)
+mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, host_mgreg_t *regs, guint8 *addr)
 {
 	MonoAotModule *amodule;
 
@@ -5175,7 +5175,7 @@ mono_aot_get_plt_entry (guint8 *code)
  *   Return the PLT info offset belonging to the plt entry called by CODE.
  */
 guint32
-mono_aot_get_plt_info_offset (mgreg_t *regs, guint8 *code)
+mono_aot_get_plt_info_offset (host_mgreg_t *regs, guint8 *code)
 {
 	guint8 *plt_entry = mono_aot_get_plt_entry (code);
 
@@ -6178,7 +6178,7 @@ mono_aot_plt_resolve (gpointer aot_module, guint32 plt_info_offset, guint8 *code
 }
 
 void
-mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, mgreg_t *regs, guint8 *addr)
+mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, host_mgreg_t *regs, guint8 *addr)
 {
 }
 
@@ -6191,7 +6191,7 @@ mono_aot_get_method_from_vt_slot (MonoDomain *domain, MonoVTable *vtable, int sl
 }
 
 guint32
-mono_aot_get_plt_info_offset (mgreg_t *regs, guint8 *code)
+mono_aot_get_plt_info_offset (host_mgreg_t *regs, guint8 *code)
 {
 	g_assert_not_reached ();
 
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -4975,7 +4975,7 @@ find_aot_module (guint8 *code)
 }
 
 void
-mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, host_mgreg_t *regs, guint8 *addr)
+mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, mgreg_t *regs, guint8 *addr)
 {
 	MonoAotModule *amodule;
 
@@ -5175,7 +5175,7 @@ mono_aot_get_plt_entry (guint8 *code)
  *   Return the PLT info offset belonging to the plt entry called by CODE.
  */
 guint32
-mono_aot_get_plt_info_offset (host_mgreg_t *regs, guint8 *code)
+mono_aot_get_plt_info_offset (mgreg_t *regs, guint8 *code)
 {
 	guint8 *plt_entry = mono_aot_get_plt_entry (code);
 
@@ -6178,7 +6178,7 @@ mono_aot_plt_resolve (gpointer aot_module, guint32 plt_info_offset, guint8 *code
 }
 
 void
-mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, host_mgreg_t *regs, guint8 *addr)
+mono_aot_patch_plt_entry (guint8 *code, guint8 *plt_entry, gpointer *got, mgreg_t *regs, guint8 *addr)
 {
 }
 
@@ -6191,7 +6191,7 @@ mono_aot_get_method_from_vt_slot (MonoDomain *domain, MonoVTable *vtable, int sl
 }
 
 guint32
-mono_aot_get_plt_info_offset (host_mgreg_t *regs, guint8 *code)
+mono_aot_get_plt_info_offset (mgreg_t *regs, guint8 *code)
 {
 	g_assert_not_reached ();
 
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -137,7 +137,7 @@ typedef struct
 	MonoInterpFrameHandle interp_frame;
 	gpointer frame_addr;
 	int flags;
-	mgreg_t *reg_locations [MONO_MAX_IREGS];
+	host_mgreg_t *reg_locations [MONO_MAX_IREGS];
 	/*
 	 * Whenever ctx is set. This is FALSE for the last frame of running threads, since
 	 * the frame can become invalid.
@@ -3172,7 +3172,7 @@ process_frame (StackFrameInfo *info, MonoContext *ctx, gpointer user_data)
 	frame->interp_frame = info->interp_frame;
 	frame->frame_addr = info->frame_addr;
 	if (info->reg_locations)
-		memcpy (frame->reg_locations, info->reg_locations, MONO_MAX_IREGS * sizeof (mgreg_t*));
+		memcpy (frame->reg_locations, info->reg_locations, MONO_MAX_IREGS * sizeof (host_mgreg_t*));
 	if (ctx) {
 		frame->ctx = *ctx;
 		frame->has_ctx = TRUE;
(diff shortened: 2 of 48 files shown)


jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++ 
@monojenkins build Linux WebAssembly C++

jaykrell:
Alternatively, truncate the pointer with a cast, knowing it will never actually  run, when it is a truncation.

jaykrell:
csc crash

jaykrell:

jaykrell:

mono/mini/mini-ppc.c
```diff
@@ -5769,11 +5769,11 @@ mono_arch_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMetho
 	return NULL;
 }
 
-mgreg_t
+gssize
```
jaykrell: This is dubious for x32 and amd64_32, will truncate.
In as much as we are looking at context for pointers, should be ok.
Not right if using debugger to look at registers in code doing 64bit integer math on in 32bit address space.

But we are mixing up host and target, and will require more change to straighten out, to support systems where registers and pointers are different size, x32 and arm64_32.

mono/mini/debugger-agent.c
```diff
@@ -6066,7 +6066,7 @@ do_invoke_method (DebuggerTlsData *tls, Buffer *buf, InvokeData *invoke, guint8
 				NOT_IMPLEMENTED;
 
 			if (sig->params [i]->byref) {
-				arg_buf [i] = (guint8 *)g_alloca (sizeof (mgreg_t));
+				arg_buf [i] = (guint8 *)g_alloca (sizeof (host_mgreg_t));
 				*(gpointer*)arg_buf [i] = args [i];
```
jaykrell: Very dangerous, if sizeof(pointer) > sizeof(host_mgreg) -- assert somewhere? Very unlikely.

mono/utils/mono-machine.h
```diff
@@ -20,6 +20,20 @@
 #include "config.h"
 #include <glib.h>
 
+#if defined (__x86_64__) || defined (__arm64__) || defined (__aarch64__)
+// Note: This is independent of __ILP32__ and __LP64__ and is not
+// the same thing as HOST_AMD64 and HOST_ARM64. It is more like
+// defined (HOST_X32) || defined (HOST_ARM6432).
```
jaykrell: It might be the same as HOST_AMD64/ARM64, as X32/ARM6432 might also define those. It is a bit hypothetical right now but might be non-hypothetical soon.

For now we could just use gssize unconditionally here for host_mgreg_t.
We might also consider having unsigned  and signed registers.

jaykrell:
csc  crashed

jaykrell:
@monojenkins build Linux x64

jaykrell:

mono/mini/method-to-ir.c
```diff
@@ -3606,7 +3606,7 @@ mini_emit_check_array_type (MonoCompile *cfg, MonoInst *obj, MonoClass *array_cl
 				mono_cfg_set_exception (cfg, MONO_EXCEPTION_MONO_ERROR);
 				return;
 			}
-			MONO_EMIT_NEW_BIALU_IMM (cfg, OP_COMPARE_IMM, -1, vtable_reg, vtable);
+			MONO_EMIT_NEW_BIALU_IMM (cfg, OP_COMPARE_IMM, -1, vtable_reg, (gssize)vtable);
```
jaykrell: Should probably be unsigned `gsize` but no difference as long as exactly sized.

mono/mini/debugger-agent.c
```diff
@@ -3172,7 +3172,7 @@ process_frame (StackFrameInfo *info, MonoContext *ctx, gpointer user_data)
 	frame->interp_frame = info->interp_frame;
 	frame->frame_addr = info->frame_addr;
 	if (info->reg_locations)
-		memcpy (frame->reg_locations, info->reg_locations, MONO_MAX_IREGS * sizeof (mgreg_t*));
+		memcpy (frame->reg_locations, info->reg_locations, MONO_MAX_IREGS * sizeof (host_mgreg_t*));
```
jaykrell: Not really a change.

mono/mini/aot-runtime.c
```diff
@@ -2976,10 +2976,10 @@ decode_llvm_mono_eh_frame (MonoAotModule *amodule, MonoDomain *domain, MonoJitIn
 		jei->clause_index = clause_index;
 
 		if (is_thumb_code (amodule, (guint8 *)jei->try_start)) {
-			jei->try_start = (void*)((mgreg_t)jei->try_start & ~1);
-			jei->try_end = (void*)((mgreg_t)jei->try_end & ~1);
+			jei->try_start = (void*)((gssize)jei->try_start & ~1);
+			jei->try_end = (void*)((gssize)jei->try_end & ~1);
 			/* Make sure we transition to thumb when a handler starts */
-			jei->handler_start = (void*)((mgreg_t)jei->handler_start + 1);
+			jei->handler_start = (void*)((gssize)jei->handler_start + 1);
```
jaykrell: Use of a signed type maybe wierd, but it is what was previously happening.

mono/mini/mini-arm.c
```diff
@@ -3085,8 +3085,8 @@ mono_arch_finish_dyn_call (MonoDynCallInfo *info, guint8 *buf)
 	DynCallArgs *p = (DynCallArgs*)buf;
```
jaykrell: Could use `HostDynCallArgs`  and `TargetDynCallArgs`.

vargaz: A lot of this code doesn't run during AOTing, so it doesn't need host/target.

jaykrell: In terms of my change, right, I think the errors all describe things that will not happen.

But I suggest either my changes, or ifdef cross_compile out a bunch of code.
The compiler doesn't know the code is dead and I'm not sure what else to do about it. Maybe we won't use C++ and C will ignore them. But presently I need SDKs to work with C++ to validate C++.

In terms of DynCallArgs, about same thing, agreed. It seems a bit off though.
If we had HostDynCallArgs and TargetDynCallArgs, we could cast the constituents less, just to fix these errors, in code that doesn't run.

If we ever have a multiple-configuration runtime, that both crosses word size and runs natively, then something larger is needed.

mono/mini/mini-arm.c
```diff
@@ -3096,7 +3096,7 @@ mono_arch_finish_dyn_call (MonoDynCallInfo *info, guint8 *buf)
 	case MONO_TYPE_I:
 	case MONO_TYPE_U:
 	case MONO_TYPE_PTR:
-		*(gpointer*)ret = (gpointer)res;
+		*(gpointer*)ret = (gpointer)(host_mgreg_t)res;
```
jaykrell: Redundant now with change to local res type.

jaykrell:

jaykrell:

mono/mini/method-to-ir.c
jaykrell: As long as exactly sized here, and not further extended to a larget type, both  of which should be the  case.

jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++
@monojenkins build Linux WebAssembly C++

jaykrell:
Basically, everywhere I added a cast, should be able to `#ifdef MONO_CROSS_COMPILE` it out.

jaykrell:
The lane(s) that was fixing are temporarly gone (OS X x64 iOS SDK C++).

jaykrell:

jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++

mono/utils/mono-machine.h
```diff
@@ -20,6 +20,25 @@
...
+#else
+typedef gssize host_mgreg_t;
+#endif
+
+// SIZEOF_REGISTER is target. mgreg_t is usually target, sometimes host.
+// There is a mismatch in cross (AOT) compilers, and the
+// never executed JIT and runtime truncate pointers.
+// When casting to/from pointers, use gsize or gssize instead of mgreg_t.
+// When dealing with register context, use host_mgreg_t instead of mgreg_t.
+// Over conversion to host_mgreg_t causes cross compilers to generate incorrect code,
```
jaykrell: This comment is probably inaccurate and should be removed/rewritten.

jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
@monojenkins build Android SDK C++
@monojenkins build iOS SDK C++

mono/utils/mono-context.h
```diff
@@ -811,18 +811,18 @@ typedef struct MonoContext {
 #include <mono/arch/mips/mips-codegen.h>
 
 typedef struct {
-	mgreg_t	    sc_pc;
-	mgreg_t		sc_regs [32];
+	mgreg_t	sc_pc;
+	mgreg_t	sc_regs [32];
```
jaykrell: Accidental  line damage.

jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
Let's try close and reopen to get the SDK builds.

jaykrell:
Reopen to get SDK builds.

jaykrell:
@monojenkins build OS X x64 Android SDK C++
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
@monojenkins build  Linux WebAssembly
@monojenkins build  Linux  ARMv7 Interpreter

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
monojenkins build OS X x64 Android SDK C++

jaykrell:
Geez hit wrong button. :(

jaykrell:
@monojenkins build OS X x64 Android SDK C++

jaykrell:
@monojenkins build OS X x64 Android SDK C++

jaykrell:
monojenkins build OS X x64 Android SDK C++

akoeplinger:
@monojenkins build OS X x64 Android SDK C++

jaykrell:
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
@monojenkin OS X x64 iOS SDK C++

jaykrell:
@monojenkins build OS X x64 iOS SDK C\+\+

jaykrell:

jaykrell:
@monojenkins build OS X x64 iOS SDK  C++

jaykrell:
@monojenkins OS X x64 iOS SDK  C++

jaykrell:
Review?

vargaz:
Using host_mgreg_t would simplify a lot of the
(mgreg_t)(gssize) casts.

jaykrell:
I'll make a separate PR and you decide between them? The applications for host_mgreg_t run pretty far and wide, for better & worse. Basically earlier in this PR. Anyway I'll make a separate PR.

It runs from MonoContext to all its direct and indirect getters/setters, except for AOT compiler.
There is also some trampoline construction that I wasn't 100% sure about, but is probably runtime, not compiler, so should be host_mgreg_t.

Thing is, there is some potential for regressing cross AOT compiler and I didn't want to risk that, but I can do it again and you can decide. It shouldn't be long.

vargaz:
It is indeed a bit difficult to find which code is host and which is target because everything is mixed together. There are about 600 references to mgreg_t in the current code, we either have to rewrite them to use either host_mgreg_t/target_mgreg_t, or keep mgreg_t to point to either the host or the target, and rewrite the other one. Not sure which is more work.

vargaz:
I guess since we already have a target_mgreg_t, we should use that for target specific stuff, and define mgreg_t to be the host one.

jaykrell:
It can be done in multiple steps...but with that risk of the cross AOT compiler.
I think I have a handle on a bunch that are obviously host..and you will have to double check.
And there are a few I'm unsure of..safe to keep as target.
When I did this exercise before, it shifted from a 90%/10% mix to an 80%/20% mix.

MonoContext is host. I think.
exception stuff (exceptions*c) is host. I think.
JIT is host but I'm not going to touch that generally, because it is next to AOT often.
There is trampoline formation which I suspect is host, but probably won't touch it all. get_arch_trampoline stuff.

vs. this PR which casts pointers to gsize and leaves all reg_t alone.. anyway I'll take a quick shot at it, and if it is too scary/incorrect/large, maybe just go with the casting for the short term. The casting becomes redundant but not incorrect when the reg_t is fixed.

It is also funny, if think about it, because, we know reg can be larger than pointer.
   Zero extend or sign extend? I think zero, but a kernel/address-space-supplier/page-table-setter-upper could decide otherwise.

jaykrell:
@monojenkins build failed

jaykrell:
This will likely be closed/unmerged in deference to a larger change, but holding it open slightly longer.

jaykrell:
This is redundant with master now, and master is better -- this often casted to gsize and then to mgreg_t, whereas master casts to gsize and then host_mgreg_t.

i.e. if a cross compiler could actually also run native, this PR had bugs, master is better now.

PR #10885, opened 2018-09-29
[jit] Profiler argument handling changes

vargaz:


alexrp:
This looks to be missing changes to `mini_profiler_context_get_this` and `mini_profiler_context_get_argument` to use the `args` field when it's set.

Also, I wonder if we should just switch to this approach regardless of target, so that we can avoid going through debug info when accessing arguments at least (accessing locals, i.e. `mini_profiler_context_get_local`, could be dealt with later).

vargaz:
That would be nice, the generated code is probably smaller than the code which saves the whole context and platform independent.

vargaz:
As for mini_profiler_context_get_local (), I'm not sure how is that supposed to be used, the locals are not very relevant at method entry and exit.

mono/mini/mini-profiler.c
```diff
@@ -260,12 +289,7 @@ mini_profiler_context_get_this (MonoProfilerCallContext *ctx)
 	if (ctx->interp_frame)
 		return memdup_with_type (mini_get_interp_callbacks ()->frame_get_this (ctx->interp_frame), m_class_get_this_arg (ctx->method->klass));
 
-	MonoDebugMethodJitInfo *info = mono_debug_find_method (ctx->method, mono_domain_get ());
-
-	if (!info)
-		return NULL;
-
-	return get_variable_buffer (info, info->this_var, &ctx->context);
+	return ctx->args [0];
```
alexrp: This has to use `memdup_with_type` (as in the interpreter case just above) as the returned buffer will be explicitly freed by users of `mono_profiler_call_context_get_this`.

See the API docs here: https://github.com/mono/mono/blob/master/mono/metadata/profiler.c#L661

mono/mini/mini-profiler.c
```diff
@@ -279,12 +303,7 @@ mini_profiler_context_get_argument (MonoProfilerCallContext *ctx, guint32 pos)
 	if (ctx->interp_frame)
 		return memdup_with_type (mini_get_interp_callbacks ()->frame_get_arg (ctx->interp_frame, pos), sig->params [pos]);
 
-	MonoDebugMethodJitInfo *info = mono_debug_find_method (ctx->method, mono_domain_get ());
-
-	if (!info)
-		return NULL;
-
-	return get_variable_buffer (info, &info->params [pos], &ctx->context);
+	return ctx->args [sig->hasthis + pos];
```
alexrp: Same here.

alexrp:
The note about LLVM should be removed in this doc comment: https://github.com/mono/mono/blob/master/mono/metadata/profiler.c#L638

alexrp:
> As for mini_profiler_context_get_local (), I'm not sure how is that supposed to be used, the locals are not very relevant at method entry and exit.

The idea was that if some interesting information happens to be computed locally in a method but isn't explicitly returned or stored anywhere, it would still be accessible through that function (at method exit) without having to modify framework sources.

vargaz:
The problem with that is that locals have liveness ranges, their locations can be shared, etc., so its not possible in general to access them at method exit without disabling a bunch of optimizations.

alexrp:
We do the same thing as sdb when doing call context instrumentation here: https://github.com/mono/mono/blob/master/mono/mini/mini.c#L3319

luhenry:
@monojenkins squash

lewurm:
@monojenkins build failed

lewurm:
@monojenkins build failed

lewurm:
there are some issues with C++:

```
trace.c: In function 'void mono_trace_enter_method(MonoMethod*, MonoProfilerCallContext*)':
trace.c:164:20: error: invalid conversion from 'void*' to 'MonoObject* {aka _MonoObject*}' [-fpermissive]
    MonoObject *o = *(void**)this_buf;
                    ^~~~~~~~~~~~~~~~~
trace.c: In function 'void mono_trace_leave_method(MonoMethod*, MonoProfilerCallContext*)':
trace.c:344:19: error: invalid conversion from 'gpointer {aka void*}' to 'MonoObject* {aka _MonoObject*}' [-fpermissive]
   MonoObject *o = *arg_in_stack_slot (buf, gpointer);
                   ^
trace.c:401:1: error: jump to label 'unlock' [-fpermissive]
 unlock:
 ^~~~~~
trace.c:316:10: note:   from here
     goto unlock;
          ^~~~~~
trace.c:323:11: note:   crosses initialization of 'void* buf'
  gpointer buf = mini_profiler_context_get_result (ctx);
           ^~~
```

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #10886, opened 2018-09-29
[cxx] Port profilers to C++.

jaykrell:


mono/profiler/log.c
```diff
@@ -1933,10 +1933,10 @@ method_filter (MonoProfiler *prof, MonoMethod *method)
 	    !mono_callspec_eval (method, &log_config.callspec))
 		return MONO_PROFILER_CALL_INSTRUMENTATION_NONE;
 
-	return MONO_PROFILER_CALL_INSTRUMENTATION_ENTER |
+	return (MonoProfilerCallInstrumentationFlags)(MONO_PROFILER_CALL_INSTRUMENTATION_ENTER |
 	       MONO_PROFILER_CALL_INSTRUMENTATION_LEAVE |
```
alexrp: Should align these, or maybe move the first value to the next line.

luhenry:
We are not interested in doing that for now.

jaykrell:
*If* use C++ in mono internal headers, we'll either want this, or reduce profilers use of them. For example, coop handles must somewhat support C due to profilers.

Issue #10838, opened 2018-09-27
Delegate members


jaredpar:
## Steps to Reproduce

Consider the following code in Program.cs:

``` csharp
using System.Linq;

class Program
{
    delegate int D1();

    static void Main()
    {
        foreach (var member in typeof(D1).GetMembers(
            System.Reflection.BindingFlags.DeclaredOnly |
            System.Reflection.BindingFlags.Public |
            System.Reflection.BindingFlags.Instance).OrderBy(m=>m.MetadataToken))
        {
            System.Console.WriteLine(member.ToString());
        }
    }
}
```

Then do the following:

1. `mcs Program.cs`
1. `mono Program.exe`

The output for the delegate constructor will be:

> Void .ctor(Object, IntPtr)

The expected output is:

> Void .ctor(System.Object, IntPtr)

Note: this is breaking the C# emit test [Microsoft.CodeAnalysis.CSharp.UnitTests.CodeGen.CodeGenClosureLambdaTests.OrderOfDelegateMembers](https://github.com/dotnet/roslyn/blob/5bedd8b880ae7123791fd1b77a319c42f3100aac/src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenClosureLambdaTests.cs#L4248)

## On which platforms did you notice this

Ubuntu 18.04 using Mono 5.19.0.209

jaredpar:
Further testing shows this is not just delegate signatures, appears to happen on all signatures.

marek-safar:
@MaximLipnin could you look into this one about ToString implementation

PR #10901, opened 2018-10-01
[Reflection] Fix the issue with ToString method in MonoCMethod type

MaximLipnin:
Fixes https://github.com/mono/mono/issues/10838.

The new behavior is similar to coreclr's one: https://github.com/dotnet/coreclr/blob/af4ec7c89d0192ad14392da04e8c097da8ec9e48/src/System.Private.CoreLib/src/System/Reflection/RuntimeConstructorInfo.cs#L145

### DIFF
--- a/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
+++ b/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
@@ -993,7 +993,7 @@ namespace MonoTests.System.Reflection
 			else
 				Assert.IsTrue (foundExpectedType, "#2-4");
 		}
-#endif	
+#endif
 	}
 	
 	// Helper class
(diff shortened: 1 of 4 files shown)


mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
```diff
@@ -994,6 +994,16 @@ public void TestLocalVariableTypes ()
...
 		}
 #endif
+
+		delegate int D1 ();
+
+		[Test]
+		public void Issue10838 ()
+		{
+			var flags = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance;
+			var members = typeof (D1).GetMembers (flags).OrderBy (m => m.MetadataToken);
```
lambdageek: Can we just get the '.ctor' method directly? Why should this test depend on the order of methods (and their tokens) in a class?

MaximLipnin: It's from original repro sample.

marek-safar: Tweak the test, order by MetadataToken is not reliable

MaximLipnin:
@monojenkins build failed

PR #10903, opened 2018-10-01
[sdks] Switch the invocation of the offsets too to 64 bit on osx.

vargaz:


vargaz:
@monojenkins build failed

Issue #10863, opened 2018-09-27
Possible GC hang while building msbuild with mono/2018-08 and mono/master


radical:
Mono version: `Mono JIT compiler version 5.18.0.103 (2018-08/56bed71f118 Tue Sep 18 07:06:42 EDT 2018)`

Steps:

0. This seemed to help with reproducing the issue: `nuget locals all -Clear`, but not always.
1. Build msbuild from `xplat-master` branch (HEAD: 17218b611a97315eaef9fa4a8add5ef049a2ff8e) with
    `$ make`
2. This gets stuck at the bootstrapped build while restoring packages. This is not reproducible every time though.

I attached to the process with lldb and ran `monobt all` and got https://gist.github.com/radical/b8130fcb125714737d5515f53b4c5031 .
IIUC, some of the threads are waiting `*Threading.Wait*`, and some seem to be waiting for the gc to resume them. Thread #19 seems to be
the one which causes the gc to kick in:

```
  thread #19
    frame #0: 0x00007fff7e735246 libsystem_kernel.dylib`semaphore_wait_trap + 10
    frame #1: 0x0000000108f1c678 mono`mono_os_sem_timedwait + 344
    frame #2: 0x0000000108f1bec5 mono`mono_threads_wait_pending_operations + 101
    frame #3: 0x0000000108ebff93 mono`sgen_client_stop_world + 563
    frame #4: 0x0000000108ed32d6 mono`sgen_stop_world + 70
    frame #5: 0x0000000108ed005d mono`sgen_perform_collection + 157
    frame #6: 0x0000000108ec5a40 mono`sgen_alloc_obj_nolock + 928
    frame #7: 0x0000000108ec2697 mono`mono_gc_alloc_vector + 119
    frame #8: 0x109176fd9 (wrapper managed-to-native) object:__icall_wrapper_mono_gc_alloc_vector (intptr,intptr,intptr) [{0x7fcbfd002958} + 0x89]  (0x109176f50 0x109177055) [0x7fcbfbd0dd60 - MSBuild.dll]
    frame #9: 0x000000010b14b6ad mscorlib.dll.dylib`System_Collections_Generic_Dictionary_2_TKey_REF_TValue_REF_Initialize_int + 157
    frame #10: 0x000000010b1496f9 mscorlib.dll.dylib`System_Collections_Generic_Dictionary_2_TKey_REF_TValue_REF__ctor_int_System_Collections_Generic_IEqualityComparer_1_TKey_REF + 73
[...]

PR #10910, opened 2018-10-01
[w32handle] Add coop state transitions in mono_w32handle_lock_handles

lambdageek:
If another thread locked at least one of the handles, and there is a
cooperative GC STW happening, this function will loop forever.

Add GC Safe transitions around the call to sleep to give this thread a chance
to suspend.

Fixes https://github.com/mono/mono/issues/10863


lambdageek:

luhenry:

PR #10913, opened 2018-10-02
[sdks] Ensure LLVM is cloned before trying to download

luhenry:
It would get the wrong LLVM hash otherwise.


luhenry:

luhenry:

PR #10921, opened 2018-10-02
[sdks/ios] Build libMonoPosixHelper for device architectures, so that zlib-helper.o is created.

rolfbjarne:
xamarin-macios currently builds zlib-helper.o using mono's source code.

It would be preferrable to use a version built by mono, and since
zlib-helper.o is a byproduct of building libMonoPosixHelper, we need to build
libMonoPosixHelper (we already build it for simulator architectures), even
though we don't need/use libMonoPosixHelper itself.

### DIFF
--- a/sdks/builds/ios.mk
+++ b/sdks/builds/ios.mk
@@ -39,6 +39,8 @@ _ios-$(1)_CXX=$$(CCACHE) $$(PLATFORM_BIN)/clang++
 
 _ios-$(1)_AC_VARS= \
 	ac_cv_c_bigendian=no \
+	ac_cv_func_fstatat=no \
+	ac_cv_func_readlinkat=no \
 	ac_cv_func_getpwuid_r=no \
 	ac_cv_func_posix_getpwuid_r=yes \
 	ac_cv_header_curses_h=no \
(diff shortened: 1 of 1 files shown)


sdks/builds/ios.mk
```diff
@@ -39,6 +39,8 @@ _ios-$(1)_CXX=$$(CCACHE) $$(PLATFORM_BIN)/clang++
 
 _ios-$(1)_AC_VARS= \
 	ac_cv_c_bigendian=no \
+	ac_cv_func_fstatat=no \
+	ac_cv_func_readlinkat=no \
```
jaykrell: Can we move this  stuff to configure.ac?

rolfbjarne: IIRC no, because configure checks pass. The problem is that configure verifies that the function exists in the current SDK, but that doesn't mean the function existed in the earliest OS version we support, and at link time we end up with errors about functions that must be weakly linked (note that weak linking isn't the solution, because then the error would just happen at runtime instead, on older Ones - the solution is to avoid the function altogether).

jaykrell: You can still hardcode it in configure.ac. Or ifdef it in the code. For people that run configure and not SDK build.

rolfbjarne: I just copied how we do it for simulator builds: https://github.com/mono/mono/blob/3084cf9922d23ef731967440c92e97562e555443/sdks/builds/ios.mk#L185-L186

I'm fine with changing the logic, but I think it should be done everywhere, which makes it out of scope for this PR.

luhenry: @rolfbjarne that’s good as-is

luhenry:

luhenry:

luhenry:

PR #10928, opened 2018-10-02
[bcl] Put xunit tests in mcs/class/lib/PROFILE/tests too

akoeplinger:
So they are next to their NUnit counterparts.

Move xunit remote executor app to tests dir too and compile it when
compiling the test assembly instead of doing it before running the test.

We need to use an absolute path for the xunit remote executor and it
needs to be in Windows format on Windows since we pass it as an env var
to the xunit tests.


mcs/build/tests.make
```diff
@@ -73,20 +76,21 @@ ifndef NO_BUILD
...
 test_assembly_dep = $(the_assembly)
 endif
-tests_CLEAN_FILES += $(ASSEMBLY:$(ASSEMBLY_EXT)=_test*.dll) $(ASSEMBLY:$(ASSEMBLY_EXT)=_test*.pdb) $(test_response) $(test_makefrag)
+tests_CLEAN_FILES += $(test_lib_output) $(test_lib_output:$(ASSEMBLY_EXT)=.pdb) $(test_response) $(test_makefrag)
 
 xtest_sourcefile = $(PROFILE_PLATFORM)_$(PROFILE)_$(ASSEMBLY:$(ASSEMBLY_EXT)=_xtest.dll.sources)
 xtest_sourcefile_excludes = $(PROFILE)_$(ASSEMBLY:$(ASSEMBLY_EXT)=_xtest.dll.exclude.sources)
 
 xunit_test_lib = $(PROFILE)_$(ASSEMBLY:$(ASSEMBLY_EXT)=_xunit-test.dll)
+xtest_lib_output = $(test_lib_dir)/$(xunit_test_lib)
```
marek-safar: We should move XTEST_REMOTE_EXECUTOR there as well

luhenry:

luhenry:

luhenry:

PR #10931, opened 2018-10-02
[metadata] Compute interface offsets of ginst by inflating the GTD instead of doing it from the grounds up.

### LINKED ISSUE #10837: Incorrect interface dispatch on generics
## Steps to Reproduce

Add the following code to Program.cs

``` csharp
using System;
interface Interface<T>
{
    void Method();
}
class Base<T> : Interface<T>
{
    void Interface<T>.Method() { Console.WriteLine("Base.Method()"); }
}
class Derived<U, V> : Base<int>, Interface<U>
{
    void Interface<U>.Method() { Console.WriteLine("Derived`2.Method()"); }
}
class Derived : Derived<int, string>, Interface<string>, Interface<int>
{
    void Interface<string>.Method() { Console.WriteLine("Derived.Method()"); }
}
class Program
{
    public static void Main()
    {
        Interface<int> j = new Derived();
        j.Method();
    }
}
```

Then 

1. `mcs Program.cs`
1. `mono Program.exe`

Expected Output: "Derived`2.Method()"
Actual Output: "Base.Method()"

Here is a sharp lab demo of the expected behavior: [sharp lab demo](https://sharplab.io/#v2:EYLgHgbALANALiATgVwHYB8ACAGABJgRgG4BYAKAEtU4BTRAMwEMBjG3ASWrqdYB4AVAHzkA3uVwT8UXAFkacABYB7ACYAKAJSkyAX3KYATLgBCjAM40Bg3CA5cGLS0NHjJmaZ1oO+QgHRzFVU1cEXwCAE41ACJTC395ZXUNKK1cPV19IwAROgoANxoVXgBVGFwANWtbWMsqOEEyz25HEuEyMTJJKTsvHktiwXjApJCwyKicxHzCgAMDIcTNFKI08nTDXEnplRtN3IKiurLCbAae5r4Ts6bvWuo2jq73c9veK4WgjVHCca2Dj6Sy1WGTIG0IAHYXJ03ABmMIQboyRhUTSuCSPLqSG59Xh1awAK1wAF5cKgaAB3PZTA6abSYyT4gG0tHA9JAA)

Note: this is a simplification of the C# emit test [Microsoft.CodeAnalysis.CSharp.UnitTests.CodeGen.CodeGenExplicitImplementationTests.TestExplicitImplementationInBaseGenericType2](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Test/Emit/CodeGen/CodeGenExplicitImplementationTests.cs#L758)

Ubuntu 18.04 using Mono 5.19.0.209

kumpera:
Say you have;
```c#
interface IBar<T> { T Bla(); }
class Foo<A,B>: IBar<A>, IBar<B> {
        int IBar<int>.Bla)() { ... }
}
```

The GTD of Foo will have the following interface offsets table:
```
4: IBar<A>
5: IBar<B>
```

With the existing code, if you have Foo<int,int> the interface offsets table will look like this:
```
4: IBar<int>
```

The problem with that is it will produce a hole in the vtable and subclasses of Foo<int,int> will have not on slot 5.

That would be just wasteful, unless you have this:

```c#
class Sub: Foo<int,int>, IBar<int> {
        int IBar<int>.Bla)() { ...  }
}
```

IOW, a subclass that overrides the interface implementation in Foo. The problem is that it's not always reliable to which iface offset
calls to `IBar<int>` will go through and thus it's possible to have calls made against instances of `Sub` to execute the code from its base
class.

The solution is to inflate the interface offsets table from the gtd as that will preserve the number of entries and will cause the
rest of the vtable layout code to fill the duplicated slots correctly.

Fixes #10837

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -1789,7 +1789,7 @@ setup_interface_offsets (MonoClass *klass, int cur_slot, gboolean overwrite)
 			}
 
 			if (!inflated->interface_id)
-				mono_class_init (inflated);
+				mono_class_setup_interface_id (inflated);
 
 			co_pair [i].ic = inflated;
 			co_pair [i].offset = gklass->interface_offsets_packed [i];
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -1731,6 +1731,20 @@ mono_class_interface_match (const uint8_t *bitmap, int id)
 }
 #endif
 
+typedef struct {
+	MonoClass *ic;
+	int offset;
+} ClassAndOffset;
+
+static int
+compare_by_interface_id (const void *a, const void *b)
+{
+	const ClassAndOffset *ca = (const ClassAndOffset*)a;
+	const ClassAndOffset *cb = (const ClassAndOffset*)b;
+
+	return ca->ic->interface_id - cb->ic->interface_id;
+}
+
 /*
  * Return -1 on failure and set klass->has_failure and store a MonoErrorBoxed with the details.
  * LOCKING: Acquires the loader lock.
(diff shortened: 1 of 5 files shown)


kumpera:
@monojenkins build failed

luhenry:
@lambdageek could you please take a look?

lambdageek:
Looks ok.
I'm going to try an SRE example - I can never predict what it will do. I'll push a test case if it looks like it behaves distinctly enough.

kumpera:
The last commit was specifically to make SRE work

marek-safar:
@monojenkins build failed

marek-safar:
@kumpera could you resolve the conflicts?

marek-safar:
@monojenkins build failed

marek-safar:
@jaykrell could you find out why is this failing on Windows only?

/cc @luhenry

lambdageek:
Not sure if this is totally related, but `qsort` on windows is not a stable sort (the C standard doesn't require it to be stable) - elements with equal keys aren't guaranteed to stay in the same relative order.

As best I can tell from testing, `gqsort_with_data` is also not stable (it's stable below 7 elements when it runs insertion sort; but the real qsort core of it isn't stable).

The easiest way to get things stable is just to add a third field to `ClassAndOffset` that records the insertion order and use that as a secondary key when `a->ic->interface_id == b->ic->interface_id`.

I'm going to do that now.


---

The following program prints different results on Windows and OSX:

```c
#include <stdio.h>
#include <stdlib.h>

typedef struct foo {
	char *s;
	unsigned int i;
} foo;

int
compare_by_id (const void *p, const void *q)
{
	const foo *a = (const foo*)p;
	const foo *b = (const foo*)q;

	return a->i - b->i;
}

int
main (void)
{
	foo f[] = {
		{"abc", 2}, {"pqr", 2}, {"xyz", 1},
	};
	int n = sizeof (f) / sizeof (foo);

	qsort (&f[0], n, sizeof (foo), &compare_by_id);

	for (int i = 0; i < n; i++){
		printf ("%d: {\"%s\", %d}\n", i, f[i].s, f[i].i);
	}
}
```

OSX:
```
0: {"xyz", 1}
1: {"abc", 2}
2: {"pqr", 2}
```

Windows:
```
0: {"xyz", 1}
1: {"pqr", 2}
2: {"abc", 2}
```

I'm going to try replacing

PR #10944, opened 2018-10-02
Fix default colors for DataGridView

HebaruSan:
## Problem

`DataGridView`'s default colors are currently as follows:

| Cell | Background | Foreground |
| --- | --- | --- |
| Header | `SystemColors.Control` | `SystemColors.WindowText` |
| Regular row | `SystemColors.Window` | `SystemColors.ControlText` |

For a typical theme, this is fine because both foreground colors are black.

For a dark theme, the `Control` colors are light-on-dark while the `Window` colors are black-on-white, so this results in black-on-dark for the header and light-on-white for the rows.

image

## Changes

Now `ControlText` is used as the foreground color for `Control`, and `WindowText` is used as the foreground color for `Window`.

image


HebaruSan:
Updated the tests. They were just looking for the old values, which are wrong as explained above.

akoeplinger:
@monojenkins build

akoeplinger:
@filipnavara do you have any thoughts on this?

PR #10948, opened 2018-10-03
[icalls] Replace gboolean and mono_bool by MonoBoolean.

jaykrell:
Though one of these is an enum, use gint32.

The enums MonoResolveTokenError and MonoProtocolType are also used in icalls, but that is left alone.

There's actually a bunch more of this I found later.
Not sure we want to fix them all, or none, etc. Maybe just leave it alone.


jaykrell:
@monojenkins build failed

PR #10950, opened 2018-10-03
[build] Build security tools to profile dir

marek-safar:


mcs/tools/security/Makefile
```diff
@@ -66,4 +66,4 @@ permview.exe: permview.cs
 	$(CSCOMPILE) $^ $(HELPER_SOURCES) -r:$(topdir)/class/lib/$(PROFILE)/mscorlib.dll -r:$(topdir)/class/lib/$(PROFILE)/Mono.Cecil.dll
 
 %.exe: %.cs $(HELPER_SOURCES)
```
akoeplinger: hmm we'll either need to update the target to be `$(topdir)/class/lib/$(PROFILE)/%.exe` or use some stamp file logic otherwise make won't recognize the output artifacts, breaking incremental build.

PR #10952, opened 2018-10-03
[cxx] Fix pointer truncation errors in cross builds. try changing title.

jaykrell:
New PR as old one wasn't running the relevant CI.

We have *apparent to compilers* but not actual pointer truncations.

They error in cross C++ builds when compiler host/pointers is 64bit and target/mgreg_t is 32bit,  which is a CI scenario (just in the past week?). These also build a runtime, which isn’t run.

1. Selectively change mgreg_t to host_mgreg_t? Reasonable, I had it, but a bit large/noisy.  I got scared of the codegen interaction, but that is probably nothing  actually.

2. ifndef MONO_CROSS_COMPILE? Reasonable, similar in size to option 3.
We do this generally but miss some. Inhibits a multi-configure, but there are already
many barriers to that.

3. Insert casts, i.e. pointers to/from gsize/gssize. Redundant but still correct if 1 or 2 are done. Clearly no-op on non-cross builds, and very safe.

4. Maybe a compiler flag, I didn’t check -- would have to worry about multiple compilers.

Option 3 casts is this PR.

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -2905,7 +2905,7 @@ decode_llvm_mono_eh_frame (MonoAotModule *amodule, MonoDomain *domain, MonoJitIn
 
 	if (is_thumb_code (amodule, code_start))
 		/* Clear thumb flag */
-		code_start = (guint8*)(((mgreg_t)code_start) & ~1);
+		code_start = (guint8*)(((gssize)code_start) & ~1);
 
 	fde = amodule->mono_eh_frame + table [(pos * 2) + 1];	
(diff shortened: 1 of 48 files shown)


jaykrell:
@monojenkins build

jaykrell:
@monojenkins build OS X x64 Android SDK C++

jaykrell:
@monojenkins build OS X x64 iOS SDK C++

jaykrell:
@monojenkins build OS X x64 Android SDK C++

jaykrell:
Extra PR was to try to get the lanes to run, but they aren't, so closing.

Issue #10945, opened 2018-10-02
Serialization error in nightly build


jaredpar:
## Steps to Reproduce
This was found attempting to get the C# semantic tests running on Mono

1. Clone https://github.com/jaredpar/roslyn
1. Check out the branch `repro/mono-serialize`
1. Make sure mono is on your path
1. Run `./build.sh --restore --build`
1. Run `./build/scripts/test.sh Debug mono Microsoft.CodeAnalysis.CSharp.Symbo` 

### Actual Behavior
This will run the Symbol unit tests only and eventually Mono will throw an exception trying to call `ITestCase.Traits` across `AppDomain` instances. The error is a `NullReferenceException` during serialization 

### Expected Behavior
The xunit process runs to completion 

## On which platforms did you notice this

This was discovered on Ubuntu 18.04. I'm using the latest nightly mono as described by the Download page

https://www.mono-project.com/download/nightly/#download-lin

```
Mono JIT compiler version 5.19.0.290 (tarball Tue Sep  4 10:59:48 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(4)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

## Stacktrace and full details 

```
    [FATAL ERROR] System.NullReferenceException
      System.NullReferenceException : Object reference not set to an instance of an object
      Stack Trace:
        
        
        Server stack trace: 
          at System.Runtime.Serialization.FormatterServices.GetParentTypes (System.RuntimeType parentType, System.RuntimeType[]& parentTypes, System.Int32& parentTypeCount) [0x0001f] in <5a872f306a874e34bfe4796f739b7324>:0 
          at System.Runtime.Serialization.FormatterServices.InternalGetSerializableMembers (System.RuntimeType type) [0x00089] in <5a872f306a874e34bfe4796f739b7324>:0 
          at System.Runtime.Serialization.FormatterServices+<>c__DisplayClass9_0.<GetSerializableMembers>b__0 (System.Runtime.Serialization.MemberHolder _) [0x00000] in <5a872f306a874e34bfe4796f739b7324>:0 
[...]

PR #10954, opened 2018-10-03
Fix null reference exception in GetParentTypes when it starts at (or reaches?) System.Object

kg:
This is the surface level issue in https://github.com/mono/mono/issues/10945. Fixing it reveals another separate issue. I'm not sure how or why we're hitting it, but the code does appear to be incorrect, and it doesn't look like we ever modified formatterservices.cs, so I think it was always broken.

### DIFF
--- a/mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs
+++ b/mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs
@@ -146,7 +146,11 @@ namespace System.Runtime.Serialization {
             /*out*/ parentTypeCount = 0;
             bool unique = true;
             RuntimeType objectType = (RuntimeType)typeof(object);
-            for (RuntimeType t1 = parentType; t1 != objectType; t1 = (RuntimeType)t1.BaseType)
+            if (parentType == null)
+                throw new ArgumentNullException("parentType");
+            if (parentType == objectType)
+                return true;
+            for (RuntimeType t1 = parentType; (t1 != objectType) && (t1 != null); t1 = (RuntimeType)t1.BaseType)
             {
                 if (t1.IsInterface) continue;
                 string t1Name = t1.Name;


kg:
@monojenkins build failed

akoeplinger:
can you add a test for this? we should also upstream the fix to corefx so if we switch this class we're not regressing.

kg:
I have no way to manually reproduce this. It has something to do with custom modifiers on types and the bad types can't be created using reflection or S.R.E. Roslyn is hitting it with automated tests again (I can't repro that either)

mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs
```diff
@@ -146,7 +146,11 @@ static FormatterServices()
             /*out*/ parentTypeCount = 0;
             bool unique = true;
             RuntimeType objectType = (RuntimeType)typeof(object);
-            for (RuntimeType t1 = parentType; t1 != objectType; t1 = (RuntimeType)t1.BaseType)
+            if (parentType == null)
```
marek-safar: This is redundant check, https://github.com/mono/mono/blob/ae45caafd9fdbc40fc97d103ca0b1cb046a2e73a/mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs#L108 already checks for null

mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs
```diff
@@ -146,7 +146,11 @@ static FormatterServices()
...
+                throw new ArgumentNullException("parentType");
+            if (parentType == objectType)
```
marek-safar: Same as above, already checked at https://github.com/mono/mono/blob/ae45caafd9fdbc40fc97d103ca0b1cb046a2e73a/mcs/class/referencesource/mscorlib/system/runtime/serialization/formatterservices.cs#L108

Issue #10957, opened 2018-10-03
[wasm] [bindings] Support for strongly typed marshalling of C# enums.


kjpou1:
JavaScript does not have the concept of Enums like TypeScript.  Strings are used to represent enums.  A strongly type contract between C# values and the string representations used by a lot of the JavaScript API's should exist.

Example:

DOM InputElement type -> [InputElementType](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-type#Form_<input>_types)

The InputElement types are represented by strings.  button, checkbox, color, date, datetime-local etc....

These strings could be represented and marshalled back and forth by defining an `Export` data contract attribute.

PR #10958, opened 2018-10-03
[wasm][bindings] Add support for strongly typed marshalling of C# enums.

kjpou1:
JavaScript does not have the concept of Enums like TypeScript.  Strings are used to represent enums.  This PR allows a strongly type contract between C# values and the string representations used by a lot of the JavaScript API's.

Example:

DOM InputElement type ->  [InputElement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-type#Form_<input>_types)

The InputElement types are represented by strings.  button, checkbox, color, date, datetime-local etc....

These string can be represented and marshalled back and forth by defining an enum as shown below with an `Export` data contract attribute.

```
    public enum InputElementType
    {
        [Export(EnumValue = ConvertEnum.ToLower)]
        Button,
        [Export(EnumValue = ConvertEnum.ToLower)]
        CheckBox,
        [Export(EnumValue = ConvertEnum.ToLower)]
        Color,
        [Export(EnumValue = ConvertEnum.ToLower)]
        Date,
        [Export("datetime-local")]
        DateTimeLocal,
   }

```

The enum conversion allows controlling the marshalling:

- Default - Marshals the enum name as is
- ToLower - Marshals the enum name to lower case string
- ToUpper - Marshals the enum name to upper case string
- Numeric - Marshals the enum numeric values.

Marshalling from C# to JavaScript using the SDK Bindings should be made as easy and pain free as possible.

Marshalling the string values from JavaScript to C# will need a little help.  The SDK Bindings Runtime provides a method to transport the string value to a strongly typed enum.

```
   WebAssembly.Runtime.EnumFromExportContract(typeof(InputElementType), inputElementType);
```

resolves: https://github.com/mono/mono/issues/10957


kjpou1:
@monojenkins build Linux WebAssembly C++

PR #10965, opened 2018-10-03
[mscorlib] System.IO.Path, add support for systems that use volume paths (under NDA)

migueldeicaza:
Workbooks test:

```
(IsVolumePrefix (""), IsVolumePrefix ("fo"), IsVolumePrefix ("a"), IsVolumePrefix ("a:"), IsVolumePrefix ("delta?"))
=> (false, false, false, false, false)

(IsVolumePrefix ("a:/"), IsVolumePrefix ("a:/aasd"), IsVolumePrefix ("mount:/etc"))
=>
(true, true, true)
```

### DIFF
--- a/mcs/class/corlib/System.IO/Path.cs
+++ b/mcs/class/corlib/System.IO/Path.cs
@@ -62,9 +62,7 @@ namespace System.IO {
 
 		internal static readonly char[] PathSeparatorChars;
 		private static readonly bool dirEqualsVolume;
-#if HAS_VOLUME_PREFIXES
-		internal static readonly bool volumePaths;
-#endif
+
 		// class methods
 		public static string ChangeExtension (string path, string extension)
 		{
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Good catch, will check on device.

migueldeicaza:
Confirmed, this is how this OS behaves.

jspohr-ubi:
I think there's another issue with these kinds of paths in Path.GetDirectoryName().

- `Path.GetDirectoryName("foo:/bar")` returns `"foo:"` (but should return `"foo:/"`)
- `Path.GetDirectoryName("foo:/")` returns `"foo:"` (but should probably return `null` (?))

It should never return `"foo:"`, because without the trailing slash that's not a valid path.

lewing:
@migueldeicaza is this something we still want/need?

lewing:
Closing due to conflicts and lack of response.  Please reopen if this is premature.

PR #10977, opened 2018-10-04
[tests] Remove magic which prepends MonoTests to NUnit test selection

akoeplinger:
When passing FIXTURE or TESTNAME to "make run-test" we automatically
prepended `MonoTests.` to the value since all of our NUnit tests
start with that to save you from typing a few chars.

Now that we have xUnit tests as well this is confusing since it doesn't
happen there. Better to remove it for consistency.


marek-safar:
Please update also https://www.mono-project.com/community/contributing/test-suite/

PR #10987, opened 2018-10-04
[runtime] Fix malformed string selection in exceptions-ppc.c

alexanderkyte:


akoeplinger:
Merging since we have no PPC on PRs anyway and would just waste builders :)

akoeplinger:
ohhh right, totally forgot about that one 😆

PR #11003, opened 2018-10-05
Adding check for empty string to correctly get constructors in case of "*" search criteria

MaximLipnin:
The following code
```
MemberInfo[] members = typeof(MyClassWithCtor).GetMember("*", MemberTypes.Constructor, BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
```
returns an empty array because `*` passed as search criteria becomes unexpected empty string later.
But `.ctor` or `.ctor*` works as expected.

The fix is to add a check for empty string.


MaximLipnin:
The original bug was caught by [TypeTests_GetMember.TestMemberTypeCombos](https://github.com/mono/corefx/blob/411def5f278956f3c514c02ab76d9bb89428fef7/src/System.Runtime/tests/System/Reflection/TypeTests.GetMember.cs#L100) xunit test.

MaximLipnin:
Thanks for review!

PR #11008, opened 2018-10-05
[sdks] Add cmake toolchain file for LLVM MXE builds

luhenry:
We couldn't compile `llvmwin{32,64}` on Linux because we would try to use `{i686,x86_64}-w64-mingw-cmake` but it wouldn't exist.


luhenry:
@monojenkins build Linux x64 Android SDK

PR #11010, opened 2018-10-05
[llvm] Add as a submodule

luhenry:
This reduces the time it takes to build the SDKs since we do not have to clone LLVM anytime you clean the repository. We will pay the price once - when we first download the submodule - but we won't have to pay for it every time we switch branch or clean the repo.


vargaz:
Why do we have to clone the llvm repo since a prebuilt one is supposed to be downloaded most of the time ?

vargaz:
Also, the llvm repo is 900mb.

luhenry:
@vargaz we already had this discussion https://xamarinhq.slack.com/archives/C03CELHQQ/p1536594867000100. The reason is we can't have a fallback to building without having cloned **before** trying to download. By having LLVM asa submodule, since you have LLVM as a submodule, you don't need to care about cloning it before trying to download.

vargaz:
Can't we clone the repo on demand when the download fails ?

luhenry:
@vargaz https://xamarinhq.slack.com/archives/C03CELHQQ/p1536271859000100

PR #11011, opened 2018-10-05
[wasm] Unbreak the debugger and harden it against bad assemblies.

kumpera:


kumpera:
@monojenkins build failed

PR #10532, opened 2018-09-10
[crash] Crash Reporter V2

alexanderkyte:
This change set represents a major set of changes to the crash reporting mechanism. 

1) Coordination involving mandatory timeouts and signals have been changed greatly, to instead use semaphores and to end waiting as soon as possible. 

2) Dumping of unmanaged symbols outside of mono has been made possible. Embedders that want to get information on frames outside of mono-sgen can register the library with an icall. This carefully preserves both privacy (opt-in) and resiliency (can't crash/lose more info when misused).

3) The dumper has moved from a workflow that kills the runtime after a long delay into one that can dump individual threads or the entire runtime in a prompt manner. The runtime doesn't die at the end of the dump, but allows threads to resume what they were doing and returns to the caller. 

4) This has allowed for the addition of a number of tests that stress both parallel and concurrent dumping of threads. These run fairly quickly, and leave the runtime in a state to continue testing. 

5) With this lightweight testing setup created, a long burn-in testing period was done. Places we were using locks or trying to malloc resulted in crashes or hangs. After making the needed changes, it managed to run the dumping tests in a loop for a whole weekend on a Linux x64 configuration without any problems.

This has resulted in a dumper with dramatically improved stability and performance.

PR #11012, opened 2018-10-06
Typesafe coop handles for all platforms.

jaykrell:
This is an alternative to https://github.com/mono/mono/pull/10212.

 - Removes the dependency of some ABIs passing a pointer and a pointer in a struct the same.
 - Therefore gives all platforms the same type safety.
 - *If* we move to C++, allows us to have member functions in coop handles.
 -  Having the coop signatures available in C can provide  benefits, though we should  probably generate them from reading metadata -- though they won't match w/o more changes, since we mixup types in several places, like IntPtr vs. pointers.

Unlike previous proposal and prototype, it retains the dynamic wrapper generation in marshal-ilgen.c and the extra native/managed roundtrips they perform.

There is less/zero sensitivity in this version to getting correct the nuances of the parameter marshaling, however this does retain an attempt to state them correctly. The previous version crashed in one test on Windows  so there is probably a mistake somewhere there, but it doesn't really matter.


jaykrell:
Huh, the very platform intended to be helped, Linux/x86, fails. Needs investigation.

jaykrell:
@monojenkins build

jaykrell:
It looks like the movement of types  around might not be needed. I'm experimenting w/o that.

jaykrell:
All the type movement should instead just get the includes, and *one* type movement demonstrated here: https://github.com/mono/mono/pull/11019

jaykrell:
Interpreter failures is known problem in master.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build WebAssembly

jaykrell:
@monojenkins build WebAssembly

jaykrell:
@monojenkins build WebAssembly

mono/metadata/icall-table.h
```diff
@@ -24,4 +24,352 @@ mono_install_icall_table_callbacks (MonoIcallTableCallbacks *cb);
...
+typedef MonoObjectHandle MonoObjectOutHandle;
+typedef MonoObjectHandle MonoObjectInOutHandle;
+typedef MonoArrayHandle MonoArrayOutHandle;
+typedef MonoArrayHandle MonoArrayInOutHandle;
+typedef MonoString MonoStringOut;
+typedef MonoStringHandle MonoStringOutHandle;
+typedef MonoReflectionModule MonoReflectionModuleOut;
+typedef MonoReflectionModuleHandle MonoReflectionModuleOutHandle;
+
+// How the arguments and return value of an icall should be wrapped.
```
jaykrell: Maybe too much duplication of comments from marshal-ilgen.c.
In earlier version, this was a comment move, not a comment copy.

mono/metadata/icall-table.h
```diff
@@ -24,4 +24,352 @@ mono_install_icall_table_callbacks (MonoIcallTableCallbacks *cb);
...
+#define MONO_HANDLE_DECLARE_RAW(id, name, func, rettype, n, argtypes)	\
+ICALL_EXPORT MONO_HANDLE_TYPE_RAW (rettype)				\
+func ## _raw ( MONO_HANDLE_FOREACH_ARG_RAW_ ## n argtypes MONO_HANDLE_COMMA_ ## n MonoError *error) \
+
+// Implement ves_icall_foo_raw over ves_icall_foo.
+// Raw handles are converted to/from typed handles and the rest is passed through.
+
+#define MONO_HANDLE_IMPLEMENT_MAYBE(cond, id, name, func, rettype, n, argtypes)	\
+										\
+MONO_HANDLE_DECLARE_RAW (id, name, func, rettype, n, argtypes)			\
```
jaykrell: "raw" can mean "raw pointer" or "type unsafe handle". Maybe a different name..

mono/metadata/icall.c
```diff
@@ -2209,6 +2207,13 @@ ves_icall_MonoField_SetValueInternal (MonoReflectionFieldHandle field, MonoObjec
 		mono_gchandle_free (value_gchandle);
 }
 
+ICALL_EXPORT void
```
jaykrell: More flexibility in the macros might fix this, i.e. "table only".

mono/metadata/marshal-ilgen.c
```diff
@@ -6405,7 +6405,7 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 			// }
 			mono_mb_emit_byte (mb, CEE_DUP);
 			int pos = mono_mb_emit_branch (mb, CEE_BRFALSE);
-			mono_mb_emit_ldflda (mb, MONO_HANDLE_PAYLOAD_OFFSET (MonoObject));
+			mono_mb_emit_ldflda (mb, MONO_HANDLE_OFFSET (MonoObject));
```
lambdageek: this is really the offset of the handle content in a `MonoRawHandle` - it doesn't have much to do with `MonoObjectHandle`.

mono/metadata/icall-def.h
```diff
@@ -1059,29 +1102,29 @@ NOHANDLES(ICALL(VOLATILE_25, "Write(uintptr&,uintptr)", ves_icall_System_Threadi
...
+HANDLES(WEBIC_3, "GetUnmanagedResourcesPtr", ves_icall_get_resources_ptr, MonoBoolean, 3, (MonoReflectionAssembly, gpointer_ref, gint32_ref))
 
 #ifndef DISABLE_COM
 ICALL_TYPE(COMOBJ, "System.__ComObject", COMOBJ_1)
-HANDLES(ICALL(COMOBJ_1, "CreateRCW", ves_icall_System_ComObject_CreateRCW))
-HANDLES(ICALL(COMOBJ_2, "GetInterfaceInternal", ves_icall_System_ComObject_GetInterfaceInternal))
-HANDLES(ICALL(COMOBJ_3, "ReleaseInterfaces", ves_icall_System_ComObject_ReleaseInterfaces))
+HANDLES(COMOBJ_1, "CreateRCW", ves_icall_System_ComObject_CreateRCW, MonoObject, 1, (MonoReflectionType))
+HANDLES(COMOBJ_2, "GetInterfaceInternal", ves_icall_System_ComObject_GetInterfaceInternal, gpointer, 3, (MonoComObject, MonoReflectionType, MonoBoolean))
+HANDLES(COMOBJ_3, "ReleaseInterfaces", ves_icall_System_ComObject_ReleaseInterfaces, void, 1, (MonoComObject))
```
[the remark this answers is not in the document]
jaykrell: And that is kinda the thing..more about the other PR. This one is maybe self-checking.
I will do a line count check and check the diff in this file for comments. Sometimes I commented out problematic lines, to fix later -- but usually I didn't because I was afraid I'd forget to.

At some point this might be generated from metadata.

mono/metadata/marshal-ilgen.c
jaykrell: I was tempted just to say `0` but avoided it.
I guess this was "broken" by typesafe handles in the first place, i.e. it is talking about the wrong level of indirection? Could probably make a struct for that ("Payload" again) and only fix up MONO_HANDLE_RAW. I can try that.

jaykrell: Not sure this is worthwhile. Unless maybe we have two of these? Offset of _raw within handle, and offset of another _raw within payload?

jaykrell:
This is as much success as you can expect from CI currently.

jaykrell:
@kyte this will merge conflict with your long standing almost finished PR, again. Maybe you win race though.

jaykrell:
Will conflict with https://github.com/mono/mono/pull/10532.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build OS X i386

jaykrell:
Failures were:
https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15414/
Test collection for System.Runtime.InteropServices.HandleCollectorTests.System.Runtime.InteropServices.HandleCollectorTests.TestHandleCollector

https://jenkins.mono-project.com/job/test-mono-pull-request-armhf/17122/
MonoTests.test-sgen-regular-ms-conc-simple.sgen-new-threads-dont-join-stw.exe

mono/metadata/icall-def.h
```diff
@@ -43,14 +43,26 @@
  * boilerplate when manipulating managed objects from runtime code and to use MonoError for
  * threading exceptions out to managed callerrs:
  *
- * HANDLES(ICALL(icallid, methodname, cfuncptr))
+ * HANDLES(icallid, methodname, cfuncptr, return-type, number-of-parmeters, (parameter-types))
```
kumpera:
```suggestion
 * HANDLES(icallid, methodname, cfuncptr, return-type, number-of-parameters, (parameter-types))
```

mono/metadata/icall-def.h
```diff
@@ -43,14 +43,26 @@
...
+ * types:
+ *   managed types are just MonoObject, MonoString, etc. * and Handle are implied.
```
kumpera: It took me a bit to get what `* and Handle are implied.` means.

kumpera:
```suggestion
 *   managed types are just MonoObject, MonoString, etc. `*` and Handle prefix are appended automatically.
```

mono/metadata/icall-def.h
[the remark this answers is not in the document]
kumpera: It's not self-checking as we don't do managed signature matching.

mono/metadata/icall-table.h
kumpera: Then let's not duplicate it.

mono/metadata/icall.c
```diff
@@ -2797,6 +2802,62 @@ ves_icall_reflection_get_token (MonoObjectHandle obj, MonoError *error)
 	return mono_reflection_get_token_checked (obj, error);
 }
 
+ICALL_EXPORT guint32
+ves_icall_reflection_get_token1 (MonoObjectHandle obj, MonoError *error)
```
kumpera: Why do we need all this function duplication?

mono/metadata/marshal-ilgen.c
```diff
@@ -6220,6 +6220,26 @@ signature_param_uses_handles (MonoMethodSignature *sig, MonoMethodSignature *gen
...
+	// with a pointer to a pointer. For this code, in order to ease
+	// ABI concerns, it is a pointer to a struct containing a pointer.
+	//
+	// The same level of indirection either way.
+	//
+	// Generated C code converts between the types. The JIT doesn't
+	// have to pass/accept structs by value.
+	//
+	// In either case, we dereference twice and both offsets are 0.
+	mono_mb_emit_ldflda (mb, MONO_HANDLE_OFFSET (MonoObject));
```
kumpera: I don't understand this, we said we'd switch to use an opaque `void*` for marshaling handles yet we're using the code for structs.

lambdageek: Yea the comment and the code are wrong. There are no structs here.

This could just needs to implement: `(MonoObject*)*((gpointer*)raw_handle))` - take the raw handle, dereference it, and cast the result to a `MonoObject*`.

mono/metadata/icall.c
jaykrell: The native function is exposed with multiple managed names, and each managed name produces a wrapper. There are duplicate wrappers, one per managed name, with name based on the native name. An additional feature of the macros might be to reuse a wrapper with a specified name.

mono/metadata/icall-table.h
jaykrell: Agreed.

mono/metadata/icall-def.h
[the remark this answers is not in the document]
jaykrell: There are three pieces, the managed signature, the wrapper, and the native implementation.
Previously, the first two were strongly related and checked, the last two not.
Now it is basically the other way, the last two are seen together and checked, the first two not.
There was and is room for improvement.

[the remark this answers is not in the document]
jaykrell: Either I should do that checking, automated, every build, and then safely bulk migrate this scheme to more like the other PR, where all the coop icalls drop the managed/native transitions to the ideal 1. And work differently than embedding, but an ok tradeoff imho.

Or we should check and migrate them piecemeal.

Either way this PR is a useful stepping stone.

mono/metadata/marshal-ilgen.c
jaykrell: There is meant to be no change to representation at this time -- i.e. marshal the old way, a pointer to a struct containing a pointer. However, the struct is no longer used or declared anywhere else, so this code just leaves kind of a note and does things oddly, in case the struct later changes (which granted, does not make this code keep working).

We could either declare the old Payload struct and use it only here.
Or just treat this as a void** and deref once -- I think the first instruction is effectively just a nop -- a type-changing move.

I will try just removing the first instruction, and remove the rest.

mono/metadata/icall.c
jaykrell: Multiple icalls have the same underlying native implementation.
The wrapper is named based on that implementation.

`System.Arbitrary1.foo` => `ves_foo_raw` => `ves_foo`
`System.Arbitrary2.foo` => `ves_foo_raw` => `ves_foo`

but that produces two (identical) ves_foo_raws.
I could probably alter their names based on __LINE__ or, the enum that is the first parameter to ICALL(), or give the ICALL macro an ability to reuse another icall's wrapper.

jaykrell: Or declare them no `NOHANDLES()` but and call the `_raw`. That is slightly abstraction violating but might be pretty good otherwise.

jaykrell: That won't work -- I confused raw pointers and raw handles.

jaykrell: I guess HANDLES_NO_WRAPPER.
marshal-ilgen is still going to produce duplicates.
If/when we move to the marshal-ilgen-less solution though, it works.

jaykrell:
Had to rebase to master due to merge conflicts, not sure what is the right approach there, as it invalidates PR. Maybe put it off until PR approved?

jaykrell:
All CI lanes passed!

mono/metadata/icall.c
```diff
@@ -7466,6 +7464,18 @@ ves_icall_System_Configuration_DefaultConfig_get_machine_config_path (MonoError
 	return mcpath;
 }
 
+ICALL_EXPORT MonoStringHandle
+ves_icall_System_Environment_GetMachineConfigPath (MonoError *error)
```
jaykrell: This is still excess and can reuse a wrapper. And some others.

jaykrell:
@monojenkins build Linux x64

jaykrell:
ok now?

PR #11017, opened 2018-10-06
[icalls] Change gboolean to MonoBoolean.

jaykrell:


jaykrell:
Returning a gboolean isn't a terrible pattern. It'll be 32bits instead of 8, which is safe. The upper bits of the return register are just as volatile as the lower bits.
Accepting 32 on the theory that ABIs support it isn't terrible either. Would need to examine things to be sure.

However neither are the convention and the idea of the convention is to be strict.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
Linux x64 Interpreter failure is known and unrelated.

jaykrell:

PR #11019, opened 2018-10-07
[icalls] Declare all icalls correctly.

jaykrell:
It is hacky to have to avoid includes in order to have conflicting/incorrect declarations.

And hacky to use extern "C" when the reason is to have incorrect prototypes.

It is arguably hacky to suppress the gcc warning -- granted, the functions are never called, and merely their addresses taken and put into void* -- arguably gcc's warning is too strict and should only trigger for calls or taking address into a non-void*.

Arguably the declarations should be moved around the tree.

Perhaps we should omit the prototypes from C++.


lambdageek:
Questions:

1. Why do we have to have both `icall.h` and `icalls.h`?
2. Does this mean we should remove icall declarations from other headers?

jaykrell:
icalls.h is preexisting and provides support for declaring icalls.
I needed a new file so called it icall.h. It declares the icalls.
We do not need to move the declarations -- this is functions that were maybe not declared anywhere else (or maybe I didn't look enough -- tedious/not automated). We could.
1. Do what I show. It suffices and is improvement.
2. Put the new ones elsewhere.
3. Put the old ones in new place.

jaykrell:
icall.h is needed in far more places than icalls,h, as long as the declarations are spread around.
icall.h is only needed by icall.c and icall-table.c.
There are very few references to icalls within the runtime -- implementation and table building.

lambdageek:
I see. Thanks.

I guess the only one that stands out now is `mono_cpu_count`.  Can we make a proper `ves_icall_System_Environment_get_ProcessorCount` icall that delegates to `mono_cpu_count`?  Then everything in icall.h will be uniform.

vargaz:
We might want to name it something different like icall-decl.h to avoid confusion.

jaykrell:
`mono_cpu_count` sure.

jaykrell:
Rename, sure.

jaykrell:
Must have mismerged with master, as it was building. Will fix.

jaykrell:
Better?

jaykrell:
Interpreter failures is known problem in master.

PR #11027, opened 2018-10-08
[Reflection] Unify exception param name

MaximLipnin:
This change unifies exception param name value which is expected in https://github.com/mono/corefx/pull/147


MaximLipnin:
Thanks for review!

Issue #10024, opened 2018-08-10
Array ctor with negative length throws TargetInvocationException instead of OverflowException


MaximLipnin:
## Steps to Reproduce
Execute the following snippet 
`typeof(object[]).GetTypeInfo().DeclaredConstructors.ToArray()[0].Invoke(new object[] { -1 });
`
or [this test](https://github.com/mono/corefx/blob/master/src/System.Reflection/tests/ConstructorInfoTests.cs#L96) 

### Current Behavior
The code throws TargetInvocationException
### Expected Behavior
The code throws OverflowException as it occurs on .Net Core FX
## On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows

**Version Used**:

## Stacktrace
Result of unit test execution:

    System.Reflection.Tests.ConstructorInfoTests.Invoke_OneDimensionalArray_NegativeLengths_ThrowsOverflowException [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.OverflowException)
      Actual:   typeof(System.Reflection.TargetInvocationException): Exception has been thrown by the target of an invocation.
      ---- System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
      -------- System.OverflowException : 
      Stack Trace:
          at System.Reflection.MonoCMethod.InternalInvoke (System.Object obj, System.Object[] parameters, System.Boolean wrapExceptions) [0x00012] in /Users/admin/dev/mono/mcs/class/corlib/System.Reflection/MonoMethod.cs:704 
          at System.Reflection.MonoCMethod.DoInvoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0007a] in /Users/admin/dev/mono/mcs/class/corlib/System.Reflection/MonoMethod.cs:688 
          at System.Reflection.MonoCMethod.Invoke (System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in /Users/admin/dev/mono/mcs/class/corlib/System.Reflection/MonoMethod.cs:720 
          at System.Reflection.ConstructorInfo.Invoke (System.Object[] parameters) [0x00000] in /Users/admin/dev/mono/external/corefx/src/Common/src/CoreLib/System/Reflection/ConstructorInfo.cs:18 
[...]

PR #11028, opened 2018-10-08
[Reflection] Handle overflow exception on constructor invocation

MaximLipnin:
It allows to re-enable  [Invoke_OneDimensionalArray_NegativeLengths_ThrowsOverflowException](https://github.com/mono/corefx/blob/master/src/System.Reflection/tests/ConstructorInfoTests.cs#L96) test and enable [TestInvoke_1DArrayWithNegativeLength](https://github.com/mono/corefx/blob/master/src/System.Runtime/tests/System/Reflection/ConstructorInfoTests.cs#L79) test.

Fixes https://github.com/mono/mono/issues/10024

### DIFF
--- a/mcs/class/corlib/System.Reflection/MonoMethod.cs
+++ b/mcs/class/corlib/System.Reflection/MonoMethod.cs
@@ -325,6 +325,8 @@ namespace System.Reflection {
 				} catch (ThreadAbortException) {
 					throw;
 #if MOBILE
+				} catch (OverflowException) {
+					throw;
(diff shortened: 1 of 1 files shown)


lambdageek:
What's the rationale for `#if MOBILE` - why do you sometimes catch `OverflowException` only under `MOBILE` and sometimes everywhere?

MaximLipnin:
Thanks @lambdageek, it shouldn't be so, I'll fix it.

MaximLipnin:
Thanks for review!

PR #11030, opened 2018-10-08
Depend on all *_test.dll.sources for `make run-test`

EgorBo:
Also, do not rebuild main library if test.sources were changed

### DIFF
--- a/mcs/build/tests.make
+++ b/mcs/build/tests.make
@@ -272,7 +272,7 @@ endif
 
 # This handles .excludes/.sources pairs, as well as resolving the
 # includes that occur in .sources files
-$(test_response_preprocessed): $(test_sourcefile) $(wildcard $(test_sourcefile_excludes)) $(GENSOURCES_CS)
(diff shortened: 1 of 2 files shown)


mcs/build/tests.make
```diff
@@ -272,7 +272,7 @@ endif
 
 # This handles .excludes/.sources pairs, as well as resolving the
 # includes that occur in .sources files
-$(test_response_preprocessed): $(test_sourcefile) $(wildcard $(test_sourcefile_excludes)) $(GENSOURCES_CS)
+$(test_response_preprocessed): $(test_sourcefile) $(wildcard *_test.dll.sources) $(wildcard *_test.dll.exclude.sources) $(GENSOURCES_CS)
```
akoeplinger: it should be *test.dll.sources here as well so we catch xunit too

EgorBo: @akoeplinger `$(test_response_preprocessed)` is only for nunit so we don't need xunit files here. Xunit already has this fix.

akoeplinger: ah right, I overlooked that part

PR #11033, opened 2018-10-08
Fix Mono.Debugger.Soft to use test exe's from different directory

akoeplinger:
And make it work for the monodroid profile.


mcs/class/Mono.Debugger.Soft/Test/dtest.cs
```diff
@@ -362,7 +369,7 @@ public static void trace_ss_recursive_chaotic (List<RecursiveChaoticPoint> trace
 	[SetUp]
 	public void SetUp () {
 		ThreadMirror.NativeTransitions = false;
-		Start (new string [] { "dtest-app.exe" });
+		Start (new string [] { Path.Combine (this_assembly_path, "dtest-app.exe") });
 	}
```
vargaz: Can this be put into a static variable so its easier to write ?

vargaz:
Looks ok otherwise.

mcs/class/Mono.Debugger.Soft/Test/dtest.cs
```diff
@@ -47,6 +47,15 @@ public class DebuggerTests
 	public static string runtime = Environment.GetEnvironmentVariable ("DBG_RUNTIME");
 	public static string runtime_args = Environment.GetEnvironmentVariable ("DBG_RUNTIME_ARGS");
 	public static string agent_args = Environment.GetEnvironmentVariable ("DBG_AGENT_ARGS");
+#if MONODROID_TEST
```
marek-safar: Why not just MOBILE or MONODROID ?

sdks/android/Makefile
```diff
@@ -192,14 +192,14 @@ package: app/assets/asm/monodroid_mini_test.dll
 
 ## Package Mono.Debugger.Soft test assemblies
 
-.PHONY: $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.exe $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.pdb
-$(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.exe $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.pdb:
-	$(MAKE) PROFILE='monodroid' -C $(TOP)/mcs/class/Mono.Debugger.Soft dtest-app.exe
+.PHONY: $(TOP)/mcs/class/lib/monodroid_tools/tests/dtest-app.exe $(TOP)/mcs/class/lib/monodroid_tools/tests/dtest-app.pdb
```
luhenry: `dtest-app.exe` should come from `monodroid` (since we run `dtest-app.exe` on the device, not the host).

sdks/desktop/Makefile
```diff
@@ -11,5 +11,5 @@ $(CURDIR)/net_4_x_Mono.Debugger.Soft_test.dll: $(TOP)/mcs/class/lib/net_4_x/test
 
 .PHONY: $(TOP)/mcs/class/lib/net_4_x/tests/net_4_x_Mono.Debugger.Soft_test.dll
 $(TOP)/mcs/class/lib/net_4_x/tests/net_4_x_Mono.Debugger.Soft_test.dll:
-	$(MAKE) TEST_MCS_FLAGS="-d:MONODROID_TEST" PROFILE='net_4_x' -C $(TOP)/mcs/class/Mono.Debugger.Soft test
+	$(MAKE) PROFILE='net_4_x' -C $(TOP)/mcs/class/Mono.Debugger.Soft test
```
luhenry: By removing `-d:MONODROID_TEST`, you are breaking https://github.com/mono/mono/pull/11033/files#diff-b3be8767d253feba4c1f59f296284053R52. If that's intentional you should remove it there as well.

sdks/android/Makefile
```diff
@@ -192,14 +192,14 @@ package: app/assets/asm/monodroid_mini_test.dll
 
 ## Package Mono.Debugger.Soft test assemblies
 
-.PHONY: $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.exe $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.pdb
-$(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.exe $(TOP)/mcs/class/Mono.Debugger.Soft/dtest-app.pdb:
-	$(MAKE) PROFILE='monodroid' -C $(TOP)/mcs/class/Mono.Debugger.Soft dtest-app.exe
+.PHONY: $(TOP)/mcs/class/lib/monodroid/tests/dtest-app.exe $(TOP)/mcs/class/lib/monodroid/tests/dtest-app.pdb
+$(TOP)/mcs/class/lib/monodroid/tests/dtest-app.exe $(TOP)/mcs/class/lib/monodroid/tests/dtest-app.pdb:
+	$(MAKE) PROFILE='monodroid_tools' -C $(TOP)/mcs/class/Mono.Debugger.Soft test
```
luhenry: This is broken: you are building the `monodroid_tools` profile but picking the `monodroid` one.

sdks/desktop/Makefile
luhenry: This hasn't been addressed.

sdks/android/Makefile
akoeplinger: That's what you told me you want to happen. dtest-app.exe should be built against the monodroid profile since it'll run on the device.

I added the necessary magic to the Mono.Debugger.Soft Makefile so that this will do the right thing.

akoeplinger:
@luhenry we discussed this in person and I thought we resolved all points, that's why I dismissed the review :)

sdks/android/Makefile
luhenry: There is confusion here: we need `dtest-app.exe` to come from `monodroid`, and `Mono.Debugger.Soft` to come from `monodroid_tools`, nothing else, nothing more. Here you are "building" in `monodroid_tools` but copying from `monodroid` which goes against any proper assumption. Also, it doesn't seem warranted from what I understand from https://github.com/mono/mono/pull/11033/files#diff-ea8be71a4e802c7bf5e99a96fc5a2ff0R11 since `dtest-app.exe` is a very simple "console" app which doesn't rely on anything else than `mscorlib` and `System` - which *are* available in `monodroid`. `Mono.Debugger.Soft` is, as we indeed discussed, a different beast since it depends on `Mono.Cecil` and others, which warrants that we build it in `monodroid_tools`.

PR #11042, opened 2018-10-09
[icall] Add InOut handles. Use InOut handles and Out handles more correctly.

jaykrell:
This is only an unenforced naming convention.


jaykrell:

mono/metadata/icall.c
```diff
@@ -8704,7 +8704,7 @@ ves_icall_System_Threading_Thread_SystemMaxStackSize (void)
 	return mono_thread_info_get_system_max_stack_size ();
 }
 
-ICALL_EXPORT gboolean
+ICALL_EXPORT MonoBoolean
```
rjvdboon: Is this change intended? Looks like part of #11017.

jaykrell: Yeah, it was missed there by accident. It is a separate thing, but it is correct and intended.

PR #11060, opened 2018-10-09
Port update_submodules.sh to update_submodules.cmd.

jaykrell:
It actually doesn't look quite right, but does mimic update_submodules.sh.

### DIFF
--- a/scripts/update_submodules.cmd
+++ b/scripts/update_submodules.cmd
@@ -1,0 +1,17 @@
+@if not exist .git goto :eof
+
+git submodule update --init --recursive && @goto :ok
+git submodule init && git submodule update --recursive && @goto :ok
+
+:: This looks wrong but is what update_submodules.sh does.
+git submodule init && git submodule update && @goto :error
+
+@echo Git submodules could not be updated. Compilation will fail.
+@exit /b 1
+
+:error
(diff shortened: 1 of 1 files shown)


scripts/update_submodules.cmd
```diff
@@ -0,0 +1,17 @@
+@if not exist .git goto :eof
+
+git submodule update --init --recursive && @goto :ok
+git submodule init && git submodule update --recursive && @goto :ok
+
+:: This looks wrong but is what update_submodules.sh does.
+git submodule init && git submodule update && @goto :error
```
akoeplinger: it's a best effort thing which is just there for very old git versions which supported neither `git submodule update --init --recursive` nor `... init  && ... update --recursive`.

I think we can remove it.

jaykrell: I thought it was just there for developer convenience.

jaykrell: On the first comment: I thought it was just there for developer convenience.

akoeplinger: Nah it's getting run by the root Makefile at the start of the build: https://github.com/mono/mono/blob/feaf8ee391a09fbff2137905ac7139f9bacc99fa/Makefile.am#L21

That's why you probably never needed to manually pull in the submodule changes ;)

PR #11084, opened 2018-10-11
[2018-10] Bump xunit

akoeplinger:
/cc @luhenry this should fix your issue


akoeplinger:
@monojenkins commit apidiff

Issue #11091, opened 2018-10-11
CI accepts failure as success


jaykrell:
1. Submit a PR
2. Expect success to mean things are working.

Counter example:

https://github.com/mono/mono/pull/11017

https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15230/

```
Checking unintended native code changes in iltests.exe with AOT
Unable to compile method 'int Tests:calli_sig_check_2 ()' due to: 'Invalid IL code in Tests:calli_sig_check_2 (): IL_000b: calli     0x11000044

 assembly:<unknown assembly> type:<unknown type> member:(null)'.
Memory around native instruction pointer (0xa17f0496):
0xa17f0486  8e 05 ff e2 c3 90 b8 48 01 08 00 e8 72 1f 00 00  .......H....r...
0xa17f0496  73 0e e8 00 00 00 00 5a 8b 92 87 1b 8e 05 ff e2  s......Z........
0xa17f04a6  c3 90 b8 4c 01 04 00 e8 56 1f 00 00 73 0e e8 00  ...L....V...s...
0xa17f04b6  00 00 00 5a 8b 92 6b 1b 8e 05 ff e2 c3 90 b8 49  ...Z..k........I

Native stacktrace:

	0   mono                                0x001fe0a9 mono_dump_native_crash_info + 809
	1   mono                                0x001aa4a3 mono_handle_native_crash + 179
	2   mono                                0x001fd389 sigabrt_signal_handler + 153
	3   libsystem_platform.dylib            0xa18cae5b _sigtramp + 43
	4   ???                                 0xffffffff 0x0 + 4294967295
```

I suspect I regressed this (the assertion failure, not CI missing it) back in May but I have to dig more.

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15206/parsed_console/log.html

search for "due to:"

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15066/parsed_console/log.html

search for "due to:"

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15055/parsed_console/log.html

search for "due to:"

lewurm:
[...]

PR #11092, opened 2018-10-11
[ci] run regression tests with AOT

lewurm:


akoeplinger:
Looks like it fails on Linux x86 / macOS x86 and Windows.

```
Mono Ahead of Time compiler - compiling assembly /mnt/jenkins/workspace/test-mono-pull-request-i386/mono/mini/iltests.exe
AOTID A23F364D-172E-95C4-BDB9-8A033A17186F
* Assertion: should not be reached at mini-x86.c:5878
...
```

On windows it can't find clang.exe ;)

lewurm:
It's expected to fail on i386 linux/osx, see https://github.com/mono/mono/issues/11091 and https://github.com/mono/mono/pull/11089.

For windows, let us not run there for now. I'll disable it.

jaykrell:
Is my understanding correct: The intent here, is that this stuff already runs, but in a way that failures are hidden from the automation. So run them again, earlier? (or later, ok), in a way that isn't hidden? 

Wasteful of course, and would be nice to surface the errors from the other run, or maybe have CI search out assertion failures, but ok. The duplication is likely small.

lewurm:
It's a bit duplication, but hard to share: `test_op_il_seq_point.sh` either uses `--aot` or `--compile-all` to trigger a compilation with the JIT for a specific method and extracts some information. It also runs the JIT with slightly different configuration than default (`MONO_DEBUG=no-compact-seq-points,single-imm-size `).

While I agree that it would be good to fix `test_op_il_seq_point.sh` to do proper error propagation, we should also run the AOT mini regression tests, which this PR does (that makes fixing `test_op_il_seq_point.sh` kinda not necessary anymore 😄 ).

akoeplinger:
@monojenkins build Linux i386
@monojenkins build OS X i386

lewurm:
@monojenkins squash

akoeplinger:
@lewurm this seems to still fail on Linux/OSX i386, can you try merging master?

lewurm:
There is some problem on linux/macos i386. I'll look into it.

lewurm:
@monojenkins build failed

PR #11100, opened 2018-10-11
Build BTLS on Windows, Visual Studio.

lateralusX:
PR setup BTLS build on Windows using Visual Studio msvc compiler. By default, BTLS builds will be disabled on Windows but can be enabled by passing --enable-btls when running ./autogen.sh after #9306 have been merged. Normal mono.sln Visual Studio build will look at configuration and decide if BTLS should be build based on that information (in original config.h). mono.sln have been extended with a new utility project and .bat file taking care of the complete BTLS build following the build configuration currently used in Visual Studio. This way the BTLS build will be an integrated build step part of the overall mono Visual Studio build. BTLS build has been setup to work under all Visual Studio mono build configuration, x86/x64 Debug/Release and build artifacts will end up under standard build tree and output directories.

PR also fixes several mismatches in mono BTLS code related to MONO_API in order to correctly build under msvc compiler and following the same pattern as used by all other mono runtime usage of MONO_API. PR also adjust a couple of warnings in mono’s BTLS code.

Since BTLS build relies on cmake, when enabling BTLS for Windows Visual Studio build, a Windows cmake version must be found or the build will fail. Build also needs to locate a working git version since it needs to make sure mono BTLS submodule has been correctly updated before building. NOTE, since the mono Visual Studio build is done before Cygwin build we can’t depend on the normal submodule update logic to take care of this.

In order to do a full BTLS build supporting assembler optimized encrypt/decrypt/hash functions the Visual Studio cmake generator can not be used. It also adds a number of additional prerequisites like yasm, ninja and perl to the build infrastructure. In order to take this in steps, this PR builds without assembler support making it possible to use cmake Visual Studio generators. Future PR’s will add support for assembler optimized BTLS build if the correct tools can be found on the build system.


jaykrell:
msvc has different consistency requirements around `MONO_API` than non-msvc.
i.e. between .h and .c.

baulig:
@directhex @jaykrell does any of you know something about building with msvc?  The code changes look good to me, but I don't know enough about this build stuff.

jaykrell:
I'd just leave it to Johan, unless you insist that an expert author it, and a second expert review it.

MONO_API is extern "C" and exported from .dll (__declspec(dllexport)), and whatever the wacky ELF stuff.
The exported from .dll business should be a .def file, really, but whatever, mono does it differently.
If you use it, and again I recommend it against it entirely, then use it consistently in .h and .c.
Unix compilers might not care for consistency, but msvc does.

jaykrell:
Oh maybe it isn't quite that -- maybe msvc wants it on .h at least, can be also on .c.
Unix doesn't care between them -- code used either.
So use .h at least.

msvc/build-external-btls.bat
```diff
@@ -0,0 +1,230 @@
+@echo off
```
jaykrell: Would prefer portable Python over any cmd but ok.

msvc/winsetup.bat
```diff
@@ -55,19 +54,33 @@ for /f "tokens=*" %%a in ('findstr /b /c:MONO_CORLIB_VERSION= %CONFIGURE_AC%') d
...
+for /f "tokens=*" %%a in ('findstr /i /r /c:".*#define.*LLVM.*" %CYG_CONFIG_H%') do (
+	echo %%a >> %CONFIG_H_TEMP%
+)
+
+echo #endif >> %CONFIG_H_TEMP%
 
 :: If the file is different, replace it.
-fc %monotemp% %config_h% >nul || move /y %monotemp% %config_h%
-del %monotemp% 2>nul
+fc %CONFIG_H_TEMP% %CONFIG_H% >nul 2>&1 || move /y %CONFIG_H_TEMP% %CONFIG_H%
```
jaykrell: No need for caps.

marek-safar:
@lateralusX is it still do-not-merge ?

lateralusX:
Just hold on a little with the merge for now, will get library name change into this PR first.

lateralusX:
Failures are unrelated and happens on other PR's.

PR #11101, opened 2018-10-11
[wasm] Add support for Module.locateFile to runtime loading

kjpou1:
See issue mono#11099 for more information.

resolves issue: mono#11099

The PR also includes a change to the `packager.exe`.

- Add --copy parameter to packager 

- copy parameter

```

        --copy=always|ifnewer        Set the type of copy to perform
                              'always' overwrites the file if it exists.
                              'ifnewer' copies or overwrites the file if modified or size is different.

```

- Fix wasm build target in do-runtime target.


kumpera:
This PR includes a bunch of changes from @vargaz to fix the build. I suppose he'll review those.

I don't get exactly what it accomplishes so I'll have to research first.
My only concern is the interaction it has with `mono_load_runtime_and_bcl` parameters. Do they become redundant/confusing to users?

PR #11102, opened 2018-10-11
[corlib] Import System.IO.IOException from CoreFX

alexischr:


alexischr:

mcs/class/referencesource/mscorlib/system/io/fileloadexception.cs
```diff
@@ -1,214 +0,0 @@
-// ==++==
```
marek-safar: Don't delete referencesource files now (makes patching them easier)

alexischr:

PR #11103, opened 2018-10-11
[sdks] Use curl to download llvm instead of wget.

rolfbjarne:
Since macOS doesn't ship wget, only curl works out-of-the-box.


rolfbjarne:
Looks like this PR is already out of date.

PR #11077, opened 2018-10-10
Replace mono_object_get_class with mono_object_class to avoid gc transitions.

jaykrell:

PR #11110, opened 2018-10-11
[2018-06] Replace mono_object_get_class with mono_object_class to avoid gc transitions.

lambdageek:
Backport #11077 to `2018-06`

### DIFF
--- a/mono/metadata/boehm-gc.c
+++ b/mono/metadata/boehm-gc.c
@@ -1611,7 +1611,7 @@ test_toggleref_callback (MonoObject *obj)
 	MonoToggleRefStatus status = MONO_TOGGLE_REF_DROP;
 
 	if (!mono_toggleref_test_field) {
-		mono_toggleref_test_field = mono_class_get_field_from_name (mono_object_get_class (obj), "__test");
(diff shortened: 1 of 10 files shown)


jaykrell:
Should we, maybe, revert “all”of this in master, and then combine the changes and put them back? I mean, to avoid the merge conflicts in backports? Or just manually port when they occur?

lambdageek:
So the interesting thing that happened here is that 2018-06 had more instances of `mono_object_get_class` than master - I had to actively write some code.  So I suspect no matter what we do, we will have to manually intervene every now and then.

Fortunately since we're marking the symbols with "external only", we'll get compile problems in the backport PRs if there's anything missing.

luhenry:
Why is that required in that branch?

luhenry:
@lambdageek not even that, more that it's a release branch so backporting something to it carries quite some risk. Is the risk worth it in this case? If it's only a performance problem which is very limited in scope and impact, I would say it isn't, but I'll let you be the judge of that.

lambdageek:
Good point.

Not worth it.

PR #11111, opened 2018-10-11
[reflection] Implement RuntimeType.IsByRefLike

### LINKED ISSUE #11086: Type.IsByRefLike throws NotSupportedException
## Steps to Reproduce

```csharp
int test = 10;
Console.WriteLine (test.GetType ().IsByRefLike);
```

### Current Behavior

```
   System.NotSupportedException : Derived classes must provide an implementation.
  at System.Type.get_IsByRefLike () [0x00000] in /Users/alexander/dev/mono/mcs/class/corlib/ReferenceSources/Type.cs:42
```

### Expected Behavior

No exception and `false` for types like int.

[ X ] macOS
[ X ] Linux
[ X ] Windows

**Version Used**: master

lambdageek:
Fixes #11086

- [x] Add a byreflike bit to `MonoClass`
- [x] Set bit when we create a class from metadata; or when we create a ginst
- runtime checks
   - [x] no byreflike arrays
   - [x] byreflike only allowed as a field iff it's an instance field and contained in another byreflike struct
   - [x] no boxing
   - [x] can't instantiate generic params by ByRefLike structs
- [x] reflection icall
- [x] Add `[PreserveDependencyAttribute (".ctor()", "System.Runtime.CompilerServices.IsByRefLikeAttribute")]` 
- [x] See https://github.com/mono/mono/issues/11220 - ~Run tests from https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/Type/TypeTests.netcoreapp.cs#L208-L213~


marek-safar:
@MichalStrehovsky could you help us here?

lambdageek:
Is it just "has `System.Runtime.CompilerServices.IsByRefLikeAttribute`"?

MichalStrehovsky:
@lambdageek yup, that's correct. The C# 7 language compilers emit this attribute on `ref struct`.

The runtime does a bit more with it:
* It enforces that a byref-like type can only be used as a field type if the owning type of the field is also byref-like and if it's an instance field
* It enforces that byref-like types can never be boxed and cannot be elements of an array.

Byref-like types are things like `Span<T>` or `TypedReference`.

mcs/class/corlib/System/RuntimeTypeHandle.cs
```diff
@@ -258,5 +258,7 @@ internal static bool IsSubclassOf (RuntimeType childType, RuntimeType baseType)
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		internal extern static bool is_subclass_of (IntPtr childType, IntPtr baseType);
 
+		[MethodImplAttribute(MethodImplOptions.InternalCall)]
```
marek-safar: Please add `[PreserveDependencyAttribute (".ctor()", "System.Runtime.CompilerServices. IsByRefLikeAttribute"` for linker tracking

lambdageek:
Okay, this is ready for review.  Runtime and JIT changes.

lambdageek:
Resolved merge conflicts.

mono/metadata/class-init.c
```diff
@@ -701,6 +716,30 @@ mono_generic_class_setup_parent (MonoClass *klass, MonoClass *gtd)
 	mono_loader_unlock ();
 }
 
+struct HasIsByrefLikeUD {
```
jaykrell: nit:  typedef to HasIsByrefLikeUD w/o struct

mono/metadata/icall-def.h
```diff
@@ -894,6 +894,7 @@ HANDLES(RTH_9, "HasInstantiation", ves_icall_RuntimeTypeHandle_HasInstantiation,
 HANDLES(RTH_20, "HasReferences", ves_icall_RuntimeTypeHandle_HasReferences, MonoBoolean, 1, (MonoReflectionType))
 HANDLES(RTH_10, "IsArray", ves_icall_RuntimeTypeHandle_IsArray, MonoBoolean, 1, (MonoReflectionType))
 HANDLES(RTH_11, "IsByRef", ves_icall_RuntimeTypeHandle_IsByRef, MonoBoolean, 1, (MonoReflectionType))
+HANDLES(RTH_21, "IsByRefLike", ves_icall_RuntimeTypeHandle_IsByRefLike, MonoBoolean, 1, (MonoReflectionType))
```
jaykrell: I thought it'd have to be `RTH_11a`, but if it works, ok.

mono/mini/method-to-ir.c
```diff
@@ -3893,6 +3893,12 @@ mini_emit_box (MonoCompile *cfg, MonoInst *val, MonoClass *klass, int context_us
 {
 	MonoInst *alloc, *ins;
 
+	if (G_UNLIKELY (m_class_is_byreflike (klass))) {
+		mono_error_set_bad_image (&cfg->error, m_class_get_image (cfg->method->klass), "Cannot box IsByRefLike type '%s.%s'", m_class_get_name_space (klass), m_class_get_name (klass));
+		mono_cfg_set_exception (cfg, MONO_EXCEPTION_MONO_ERROR);
```
jaykrell: Tests for all the error paths?

mono/metadata/icall-def.h
marek-safar: This requires mscorlib version bump (otherwise LGTM)

lambdageek:
Ok, I'm going to add the version bump and try to add some IL negative tests, if possible.

lambdageek:
Hm. tests are good.  - I expected the IL equivalent of`G<R> o = new G<R>()` where `R` is a `ref struct` to throw some kind of exception, but it didn't.  Investigating...

mono/metadata/custom-attrs.c
```diff
@@ -2397,6 +2411,23 @@ mono_assembly_metadata_foreach_custom_attr (MonoAssembly *assembly, MonoAssembly
...
 
+	metadata_foreach_custom_attr_from_index (image, idx, func, user_data);
+}
+
+/**
+ * iterate over the custom attributes that belong to the given index and call func, passing the 
+ *  assembly ref (if any) and the namespace and name of the custom attribute.
+ *
+ * Everything is done using low-level metadata APIs, so it is fafe to use
+ * during assembly loading and class initialization.
```
vargaz: Typo.

lambdageek:
Now with some IL tests.

Had to add a new check to `mono_class_create_generic_inst` to fail classes that are invalid instantiations.

The issue is that in the metadata parser we create a `MonoGenericContext` separately from the generic type definition that it will be attached to, and then bring them together right in the parser without going through `mono_class_inflate_generic_type_checked`. 

This happens for IL locals that are generic instances, we basically create those `MonoType`s without going through the inflation machinery and consequently miss sanity-checking the instantiation.

So now we'll do it when we actually try to make a `MonoClass` out of the `MonoType`.

(Note: checking in the parser didn't work because it broke an existing test.  we can't be too eager to fail.)

lambdageek:

PR #11112, opened 2018-10-11
Replace MONO_CLASS_IS_INTERFACE  with MONO_CLASS_IS_INTERFACE_INTERNAL, remove gc mode transitions.

jaykrell:
`MONO_CLASS_IS_INTERFACE` uses `mono_class_get_type`.
`MONO_CLASS_IS_INTERFACE_INTERNAL` uses `m_class_get_byval_arg` instead.
Despite the much different name, the difference is only GC mode transition.

There is no other runtime user of `mono_class_get_type`, so `mono_class_get_type_internal` not introduced. If needed, either introduce `mono_class_get_type_internal`, or use `m_class_get_byval_arg` instead.

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -1137,7 +1137,7 @@ make_generic_param_class (MonoGenericParam *param)
 			;
 
 	pos = 0;
-	if ((count > 0) && !MONO_CLASS_IS_INTERFACE (pinfo->constraints [0])) {
+	if ((count > 0) && !MONO_CLASS_IS_INTERFACE_INTERNAL (pinfo->constraints [0])) {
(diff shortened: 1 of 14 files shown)


jaykrell:
Um. Or should we just remove GC mode transitions in `mono_class_get_type`?

lambdageek:
(IMO) yes. the runtime should not depend on the embedding API.

jaykrell:
Agreed.

jaykrell:
The thing about types, is..they should be opaque. Like JNI. The ability to provide functions however hinging largely on the  actual  internal details.

jaykrell:
@monojenkins build failed

lambdageek:

Issue #10318, opened 2018-08-27
Coop: deadlock in threadpool-exceptions1.exe


lambdageek:
The test `threadpool-exceptions1.exe` may deadlock when using cooperative suspend.

The test in question has two threads, one (Thread 4) executing
`Environment.Exit (1)`, the other (Thread 1) is executing `Monitor.Wait (object)`.

Stack traces:
```
Thread 4 (Thread 0x7f84b5748700 (LWP 18447)):
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f84b8e07023 in __GI___pthread_mutex_lock (
    mutex=mutex@entry=0x55c15e3f9540) at ../nptl/pthread_mutex_lock.c:78
#2  0x000055c15c5d2b3d in mono_os_mutex_lock (mutex=0x55c15e3f9540)
    at ../../mono/utils/mono-os-mutex.h:99
#3  mono_coop_mutex_lock (mutex=0x55c15e3f9540)
    at ../../mono/utils/mono-coop-mutex.h:56
#4  signal_monitor (mon_untyped=0x55c15e3f91e0) at monitor.c:771
#5  0x000055c15c64d0d7 in mono_thread_info_finish_interrupt (
    token=0x55c15e3f9500) at mono-threads.c:1857
#6  0x000055c15c57e634 in async_suspend_internal (thread=0x7f84b8780130, 
    interrupt=1) at threads.c:5425
#7  0x000055c15c581070 in mono_thread_suspend_all_other_threads ()
    at threads.c:3720
#8  0x000055c15c51bf96 in ves_icall_System_Environment_Exit (result=0)
    at icall.c:6902
#9  0x00000000407e13bd in ?? ()
#10 0x00007f84b57479e0 in ?? ()
#11 0x000055c15e3eb8f8 in ?? ()
#12 0x00007f84b8000510 in ?? ()
#13 0x00007f84b57479e0 in ?? ()
#14 0x00007f84b8000500 in ?? ()
#15 0x00007f84ac002c20 in ?? ()
#16 0x00007f84b5747730 in ?? ()
#17 0x00007f84b5747690 in ?? ()
#18 0x00007f84b5747730 in ?? ()
#19 0x00000000407e123b in ?? ()
#20 0x000055c15e3edc48 in ?? ()
#21 0x00007f84b5747940 in ?? ()
#22 0x00007f84b57479f0 in ?? ()
[...]

PR #11113, opened 2018-10-11
Remove GC mode transitions from mono_array_*, and replace with mono_array_*_internal.

jaykrell:


mono/metadata/object-internals.h
```diff
@@ -168,7 +168,7 @@ struct _MonoString {
 
 // Equivalent to mono_array_addr_with_size, except:
 // 1. A macro instead of a function.
-// 2. No GC enter/exit unsafe transition.
+// 2. mono_array_addr_with_size could, but does not, do GC mode transitions.
```
jaykrell: The macro vs. function means the types  of the parameters are "generic".
Or we could cast them. Or review the few uses. Or change it to an inline function.
Or make an _internal form that is inline function, for extreme parity.

jaykrell: It looks like the only? two uses had type parity already.

jaykrell: For size to be a signed int is kinda bogus, should be unsigned size_t.

jaykrell:
oops

jaykrell:
This is messed up for multiple reasons.
1.  Changing function to macro messes with the types subtely.
2. There are several macros that pile on to this.

So the solution is
1. Replace function with new function, inline or not, not preexisting macro.
2. And all the macros with new macros.

jaykrell:
Linux x64 C++ failure  is https://github.com/mono/mono/issues/10318 occurred here.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
Might need to poll sometimes in loops.

jaykrell:
Likely subsumed by  https://github.com/mono/mono/pull/11188.

jaykrell:

jaykrell:
Redundant with master now.

PR #11119, opened 2018-10-12
[DONT MERGE] MARTIN TEST #2

baulig:
@monojenkins skip

Just testing the "OS X x64 iOS SDK" lane over night on Jenkins.  There are two PR's both based on the latest mono/master commit (same commit hash), this one has my changes cherry-picked on top, the other one does not.

If the one that's mono/master without my changes passes the "OS X x64 iOS SDK" build and this one timeouts, then we know that it's most likely an issue in my code.


baulig:

PR #11121, opened 2018-10-12
mono_class_init and GC mode

jaykrell:
Add GC mode transitions to mono_class_init.
Provide  mono_class_init_internal that still lacks them.
Convert all runtime callers from mono_class_init to mono_class_init_internal.
Mark mono_class_init as external only.
Finally stake a claim to external_only and move/add external-only functions there -- just one to start.


mono/metadata/class-internals.h
```diff
@@ -1447,6 +1447,9 @@ mono_class_contextbound_bit_offset (int* byte_offset_out, guint8* mask_out);
 gboolean
 mono_class_init_checked (MonoClass *klass, MonoError *error);
 
+gboolean
+mono_class_init_internal (MonoClass *klass);
```
lambdageek: Please move this to `class-init.h` - all class setup/initialization entrypoints ought to go there.

jaykrell:
@monojenkins build failed

jaykrell:
If it fixes things, this should include external-only instead of compile it separately.
Like subsumed by  https://github.com/mono/mono/pull/11188.

jaykrell:
Also might want more holistically transition class.h, like https://github.com/mono/mono/pull/11188 does object.h.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/class-init.h
```diff
@@ -10,6 +10,9 @@
 #include <mono/metadata/metadata.h>
 #include <mono/metadata/class-internals.h>
 
+gboolean
+mono_class_init_internal (MonoClass *klass);
```
lambdageek: Should be after `MONO_BEGIN_DECLS`.

jaykrell: This was deliberate. It is internal and doesn't need `extern C`.

jaykrell: Many cases of `G_BEGIN_DECLS` and `MONO_BEGIN_DECLS` are suspicious and should be removed, including this one. The header is not public.
But removing them is considered excess change, esp. when we aren't yet commited to C++, and does indeed carry risk -- of needing to add back some `G_EXTERN_C` as found and needed.

jaykrell: Arguably, as a replacement for a prior function, it should extern C, to maintain consistenty but that doesn't seem super well motivated. Is consistency the reason, or other?

lambdageek:
On second thought let's just delete `MONO_BEGIN_DECLS`/`MONO_END_DECLS` in class-init.h entirely.  they're not needed for an internal header.

jaykrell:
Granted, maybe the placement is wierd. Can move it when externC is removed.

PR #11122, opened 2018-10-12
Remove GC mode transitions from mono_array_length, and replace with mono_array_length_internal.

jaykrell:


jaykrell:
appdomain-serialize-exception.exe timeout, missing GC transitions?

mono/metadata/object-internals.h
```diff
@@ -164,7 +164,13 @@ struct _MonoString {
...
 #define mono_string_length_fast(s) ((s)->length)
 
-#define mono_array_length_fast(array) ((array)->max_length)
+/**
+ * mono_array_length_internal:
+ * \param array a \c MonoArray*
+ * \returns the total number of elements in the array. This works for
+ * both vectors and multidimensional arrays.
+ */
+#define mono_array_length_internal(array) ((array)->max_length)
```
lambdageek: Why rename it?

jaykrell: Consistency and it was little used.

jaykrell: Consistency and it was little used, but ok either way.

jaykrell:
@monojenkins build failed

jaykrell:
I'm wondering if we should insert safepoints in the new functions, and therefore leave fast as it was, its name and no safepoint.

Issue #11021, opened 2018-10-07
Mono fails to build on FreeBSD/PowerPC64


lenoil98:
I understand that PowerPC64 is not officially supported on FreeBSD. I can successfully build 32-bit Mono on FreeBSD/PowerPC.  The 64-bit build progresses fairly well.  However I get a SIGILL when building.  Below is an excerpt from the build log:

```
System.Net/HttpConnection.cs(57,19): warning CS0414: The private field `System.Net.HttpConnection.cert' is assigned but its value is never used
System.Net/HttpWebRequest.cs(86,10): warning CS0414: The private field `System.Net.HttpWebRequest.initialMethod' is assigned but its value is never used
System.Net/WebConnection.cs(196,25): warning CS0649: Field `System.Net.WebConnection.ID' is never assigned to, and will always have its default value `0'
System.Net/WebConnectionStream.cs(46,10): warning CS0414: The private field `System.Net.WebConnectionStream.locker' is assigned but its value is never used
System.Net/WebConnectionStream.cs(49,17): warning CS0649: Field `System.Net.WebConnectionStream.IgnoreIOErrors' is never assigned to, and will always have its default value `false'
System.Net/WebOperation.cs(65,25): warning CS0649: Field `System.Net.WebOperation.ID' is never assigned to, and will always have its default value `0'
System.Net/WebOperation.cs(66,10): warning CS0649: Field `System.Net.WebOperation.ME' is never assigned to, and will always have its default value `null'
System.Net/WebRequestStream.cs(49,28): warning CS0649: Field `System.Net.WebRequestStream.ME' is never assigned to, and will always have its default value `null'
System.Net/WebResponseStream.cs(78,28): warning CS0649: Field `System.Net.WebResponseStream.ME' is never assigned to, and will always have its default value `null'
Compilation succeeded - 45 warning(s)
MCS     [basic] System.Xml.dll
MCS     [basic] System.Security.dll
../../../external/corefx/src/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/KeyInfoRetrievalMethod.cs(78,24): warning CS0219: The variable `retrievalMethodElement' is assigned but its value is never used
System.Security.Cryptography.Xml/SignedXml.cs(391,11): warning CS0219: The variable `hashvalue' is assigned but its value is never used
Compilation succeeded - 2 warning(s)
MCS     [basic] System.Core.dll
[...]

PR #11127, opened 2018-10-12
[ppc] use G_BYTE_ORDER in order to define patch_load_sequence

lewurm:
Fixes https://github.com/mono/mono/issues/11021


lewurm:
@monojenkins squash

PR #11133, opened 2018-10-12
[crash] Bump crash reporter static memory size

### LINKED ISSUE #11130: Ran out of space to create crash dump json blob, and no stack.
## Steps to Reproduce

1.  Submit a PR.

### Current Behavior

Failures, and errors within the failure reports.

https://jenkins.mono-project.com/job/test-mono-pull-request-armhf/17122/testReport/junit/MonoTests/test-sgen-regular-ms-conc-simple/sgen_new_threads_dont_join_stw_exe/

```
one 1 iterations
Pkilling 0xf36ff450 from 0xf33ff450
Pkilling 0xf7a23450 from 0xf33ff450
Pkilling 0xf7cf35d0 from 0xf33ff450
Pkilling 0xf32ff450 from 0xf33ff450
Pkilling 0xf31ff450 from 0xf33ff450
Pkilling 0xf3aff450 from 0xf33ff450
Pkilling 0xf30ff450 from 0xf33ff450
Pkilling 0xf2fff450 from 0xf33ff450
Pkilling 0xf2eff450 from 0xf33ff450
Pkilling 0xf37ff450 from 0xf33ff450
Entering thread summarizer pause from 0xf33ff450
Finished thread summarizer pause from 0xf33ff450.
domain required for stack walk
Ran out of space to create crash dump json blob.
```

What space? File system space? Fixed size buffer?
And there is no stack as we used to often get.

### Expected Behavior

Success, or at least better errors.

alexanderkyte:
Fixes #11130

### DIFF
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -26,7 +26,7 @@ extern GCStats mono_gc_stats;
 #include <mach/task_info.h>
 #endif
 
-#define MONO_MAX_SUMMARY_LEN 2500
+#define MONO_MAX_SUMMARY_LEN 500000
 static gchar output_dump_str [MONO_MAX_SUMMARY_LEN];
 
 static JsonWriter writer;
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -61,7 +61,7 @@ static void assert_has_space (void)
 		return;
 
 	if (static_gstr.allocated_len - static_gstr.len < margin)
-		g_error ("Ran out of space to create crash dump json blob.");
+		g_error ("Ran out of memory to create crash dump json blob. Current state:\n%s\n", static_gstr.str);
 }
 
 static void


vargaz:
This will increase memory usage by 500k ?

alexanderkyte:
I would love to be able to allocate that large a chunk at runtime through another means, but I'm running from a signal handler. File IO or mmap is a possible alternative, but async safety is a concern.

jaykrell:
Open/read/write/mmap should be ok in signal handler — as long as no heap allocation to form full path. At least on Unix.

Do the devices page? I.e. is this address space or physical memory? Allocating address space is pretty cheap.

jaykrell:
Can also allocate with mmap, when needed. VirtualAlloc on Windows.

alexanderkyte:
Yeah, we agreed on mmap in channel. Let's hope it's stable enough.

jaykrell:
Please realize and remember that large global buffers are super cheap on many, maybe almost all, platforms. Address space only.

jaykrell:
Certainly Mac and Windows and Linux desktop environments. Anything with paging. You only pay for pages you touch, when you touch them.

mono/mini/mini-posix.c
```diff
@@ -1023,15 +1026,27 @@ dump_native_stacktrace (const char *signal, void *ctx)
 
 			if (!leave) {
 				mono_sigctx_to_monoctx (ctx, &mctx);
+
+				gchar *scratch_space = NULL;
+#if HAVE_SYS_MMAN_H
+				const int max_summary_len = 500000;
+				scratch_space = (gchar *) mmap (NULL, max_summary_len, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
```
lewurm: I suggest to use `mono_valloc` instead.

alexrp:
> Let's hope it's stable enough.

Just for the record, the GC, profiler, and probably many other parts of the runtime literally wouldn't be able to function if `mmap` from async context (signal handler, world stopped, etc) wasn't safe. It's definitely something you can rely on.

mono/mini/mini-posix.c
alexanderkyte: mono_valloc has ```	BEGIN_CRITICAL_SECTION;
``` logic. It was mentioned in channel that this function takes a lock. Looking at it now, I think mono_valloc might be okay.

alexanderkyte:
Following up on conversation in slack, I'm going to use the lazily-initialized static array. This 500k memory increase won't be realized on any machine with an MMU and OS paging, not until we try to write into it.

PR #11140, opened 2018-10-12
[2018-10] [sdks] Build tests for BCL targets

luhenry:


luhenry:
@monojenkins build failed

PR #11142, opened 2018-10-12
[crash] Split unmanaged method pointers into method ip + offset

alexanderkyte:
The VS4M team is making a tool that makes a symbolication lookup table. This enables symbolication without access to the relevant binary. 

For unmanaged, it's much harder to precompute things because we were presenting the instruction address. From the address, any lookup needs to find the function enclosing this interior pointer.

To make it easier, this reports the address of the native method as well as the offset. A lookup table can simply key on the function start address with this now. 

ie:

  ``` 
      {
        "native_address" : "0x1003a6340",
        "native_offset" : "0x00046",
        "native_module" : "mono"
      }
```

 ```
$ nm mini/mono-sgen | grep '1003a6340' 

00000001003a6340 t _mono_runtime_run_main_checked

```

```
$ nm mini/mono-sgen | grep ' t ' | sort -n

00000001000a4e90 t _alloc_ireg
00000001000a4eb0 t _decompose_long_opcode
00000001000aae90 t _mono_decompose_long_opts
00000001000aaea0 t _mono_decompose_vtype_opts
00000001000ae130 t _m_class_get_byval_arg
00000001000ae150 t _alloc_dreg
............
```

### DIFF
--- a/mono/metadata/threads-types.h
+++ b/mono/metadata/threads-types.h
@@ -561,6 +561,7 @@ typedef struct {
 	} managed_data;
 	struct {
 		intptr_t ip;
+		gint32 offset;
 		const char *module;
 		gboolean is_trampoline;
 		gboolean has_name;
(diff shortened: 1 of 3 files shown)


alexanderkyte:
@monojenkins build failed

PR #11169, opened 2018-10-15
`make dist` fixes for 8d8fd97dbfc6f4213e0fe9fca7574808dd8496d3

directhex:


mcs/class/Makefile
```diff
@@ -469,14 +469,14 @@ lib/$(monolite_dir)/Facades:
 
 $(monolite_files): | lib/$(monolite_dir)
 $(monolite_files): | lib/$(monolite_dir)/Facades
-$(monolite_files): lib/$(monolite_dir)/%: lib/build-$(HOST_PLATFORM)/%
```
akoeplinger: this doesn't look right, we do want a monolite with the platform-specific BCL files

mcs/class/Makefile
```diff
@@ -469,14 +469,14 @@ lib/$(monolite_dir)/Facades:
...
+$(monolite_files): lib/$(monolite_dir)/%: lib/build/%
```
marek-safar: This does not seem to be right, why did you have to change it?

akoeplinger:
Found the root cause, will be fixed by https://github.com/mono/mono/pull/11185

PR #11174, opened 2018-10-15
[2018-06] [sdks] futimens and futimes symbols are missing on anything earlier than 10.13

luhenry:
This breaks when LLVM is built on 10.13+ and the runtimes are built on 10.12.


luhenry:

sdks/builds/llvm.mk
```diff
@@ -67,6 +67,7 @@ define LLVMTemplate
 _llvm-$(1)_CMAKE_FLAGS = \
 	$$(llvm_CMAKE_FLAGS) \
 	-DCMAKE_INSTALL_PREFIX=$$(TOP)/sdks/out/llvm-$(1) \
+	-DHAVE_FUTIMENS=0 -DHAVE_FUTIMES=0 \
```
akoeplinger: fwiw `futimes` does exist on older macOS versions.

PR #11176, opened 2018-10-15
[2018-08] [sdks] futimens symbols are missing on anything earlier than 10.13

luhenry:
This breaks when LLVM is built on 10.13+ and the runtimes are built on 10.12.


luhenry:

Issue #11138, opened 2018-10-12
Hybrid suspend: mono_add_internal_call is broken


lambdageek:
- Embedders can use `mono_add_internal_call` to register an icall.
- Hybrid suspend runs icalls in GC Unsafe (cooperative suspended) mode.
- Embedders don't know that and rely on being able to call blocking operations.

(fixable complication: the runtime also calls `mono_add_internal_call` - we should redirect those to a new entrypoint)

We should run embedder icalls in GC Safe (preemptively suspended) mode under hybrid suspend.

lambdageek:
For example XM `xamarin_release_managed_ref` calls `xamarin_framework_peer_lock(); xamarin_framework_peer_unlock ();` 
https://github.com/xamarin/xamarin-macios/blob/f4dbb6840ae6d908523ae69f5b5a01c75228fc88/runtime/runtime.m#L1998-L1999

PR #11180, opened 2018-10-15
[2018-08][marshal] Emit GC Safe transitions around foreign internal calls.

lambdageek:
* Mark `mono_add_internal_call` external only.  Add `mono_add_internal_call_internal` and `mono_add_internal_call_with_flags` for the runtime.
   - icalls added using the external-only API are "foreign".
 
* Add coop transitions around `mono_gc_toggleref_add` (and mark it external only  - it's unused in the runtime).
* Change icall wrappers to call foreign icalls in GC Safe mode.
   - foreign icalls are not coop-aware, so if they perform a blocking operation under hybrid suspend they will hang indefinitely - this is bug #11138. With this change, we run them in GC Safe mode and rely on the runtime API to transition back to GC Unsafe (same as other embedder code).
   - this will require changes to watchOS support in xamarin-macios.

----

This is the `2018-08` version - `master` version is coming soon (the icall internals changed a bit, so those bits wouldn't port cleanly anyway).

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -8223,6 +8223,7 @@ ves_icall_System_IO_LogcatTextWriter_Log (const char *appname, gint32 level, con
 static MonoIcallTableCallbacks icall_table;
 static mono_mutex_t icall_mutex;
 static GHashTable *icall_hash = NULL;
+static GHashTable *icall_hash_foreign = NULL;
 static GHashTable *jit_icall_hash_name = NULL;
 static GHashTable *jit_icall_hash_addr = NULL;
 
@@ -8240,6 +8241,7 @@ mono_icall_init (void)
 	mono_icall_table_init ();
 #endif
 	icall_hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+	icall_hash_foreign = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
 	mono_os_mutex_init (&icall_mutex);
 }
 
@@ -8259,6 +8261,7 @@ void
 mono_icall_cleanup (void)
 {
 	g_hash_table_destroy (icall_hash);
+	g_hash_table_destroy (icall_hash_foreign);
 	g_hash_table_destroy (jit_icall_hash_name);
 	g_hash_table_destroy (jit_icall_hash_addr);
 	mono_os_mutex_destroy (&icall_mutex);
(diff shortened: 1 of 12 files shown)


mono/metadata/icall.c
```diff
@@ -8314,6 +8349,15 @@ mono_lookup_internal_call_full (MonoMethod *method, mono_bool *uses_handles)
 		mono_icall_unlock ();
 		return res;
 	}
+	res = g_hash_table_lookup (icall_hash_foreign, mname);
+	if (res) {
+		if (foreign)
+			*foreign = TRUE;
+		g_free (classname);
```
jaykrell: Ideally one exit path but ok.

mono/metadata/marshal-ilgen.c
```diff
@@ -1673,6 +1674,103 @@ mono_mb_emit_auto_layout_exception (MonoMethodBuilder *mb, MonoClass *klass)
 	mono_mb_emit_exception_marshal_directive (mb, msg);
 }
 
+typedef struct EmitGCSafeTransitionBuilder {
+	MonoMethodBuilder *mb;
+	gboolean func_param;
+	int coop_gc_stack_dummy;
+	int coop_gc_var;
+#ifndef DISABLE_COM
```
jaykrell: Probably not worth ifdef but ok.

mono/metadata/marshal-ilgen.c
```diff
@@ -1673,6 +1674,103 @@ mono_mb_emit_auto_layout_exception (MonoMethodBuilder *mb, MonoClass *klass)
...
+	if (mono_threads_is_blocking_transition_enabled ()) {
+		builder->mb = mb;
+		builder->func_param = func_param;
+		builder->coop_gc_stack_dummy = -1;
+		builder->coop_gc_var = -1;
+#ifndef DISABLE_COM
+		builder->coop_cominterop_fnptr = -1;
+#endif
+		return TRUE;
+	} else
```
jaykrell: else not needed, but ok.

mono/metadata/marshal-ilgen.c
```diff
@@ -1694,16 +1792,16 @@ emit_native_wrapper_ilgen (MonoImage *image, MonoMethodBuilder *mb, MonoMethodSi
 	MonoClass *klass;
 	int i, argnum, *tmp_locals;
 	int type, param_shift = 0;
-	int coop_gc_stack_dummy, coop_gc_var;
-#ifndef DISABLE_COM
-	int coop_cominterop_fnptr;
-#endif
+	gboolean need_gc_safe = FALSE;
+	GCSafeTransitionBuilder gc_safe_transition_builder;
```
jaykrell: memset this early on?

mono/mini/mini-runtime.c
```diff
@@ -4463,17 +4464,17 @@ mini_init (const char *filename, const char *runtime_version)
 static void
 register_icalls (void)
 {
-	mono_add_internal_call ("System.Diagnostics.StackFrame::get_frame_info",
+	mono_add_internal_call_internal ("System.Diagnostics.StackFrame::get_frame_info",
```
jaykrell: double internal sounds confusing

mono/metadata/marshal-ilgen.c
lambdageek: it's initialized by `gc_safe_transition_builder_init`, if needed.

PR #11183, opened 2018-10-15
[marshal] Emit GC Safe transitions around foreign internal calls.

lambdageek:
* Mark `mono_add_internal_call` external only.  Add `mono_add_internal_call_internal` and `mono_add_internal_call_with_flags` for the runtime.
   - icalls added using the external-only API are "foreign".
 
* Add coop transitions around `mono_gc_toggleref_add` (and mark it external only  - it's unused in the runtime).
* Change icall wrappers to call foreign icalls in GC Safe mode.
   - foreign icalls are not coop-aware, so if they perform a blocking operation under hybrid suspend they will hang indefinitely - this is bug #11138. With this change, we run them in GC Safe mode and rely on the runtime API to transition back to GC Unsafe (same as other embedder code).
   - this will require changes to watchOS support in xamarin-macios.

----

This is the `master` version.  `2018-08` backport is #11180

### DIFF
--- a/mono/metadata/icall-internals.h
+++ b/mono/metadata/icall-internals.h
@@ -66,6 +66,12 @@ mono_icall_drive_info_get_drive_type (MonoString *root_path_name);
 #endif  /* !G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT) */
 
 void*
-mono_lookup_internal_call_full (MonoMethod *method, gboolean warn_on_missing, mono_bool *uses_handles);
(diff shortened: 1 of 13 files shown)


lambdageek:
@kumpera thanks. Added `mono_dangerous_add_raw_internal_call`

lambdageek:
The C++ builds have been failing pretty consistently here.  Not sure why.  Locally, OS X x64 C++ builds fine.

lewurm:
@monojenkins build failed

lewurm:
@monojenkins build failed

marek-safar:
@lambdageek is this the version which should go to 2018-10?

PR #11188, opened 2018-10-16
[coop] Transition all functions in object.h to MONO_RT_EXTERNAL_ONLY

jaykrell:
In general this is achieved by renaming mono_foo to mono_foo_internal,
and mono_foo calls mono_foo_internal, possibly adding a GC mode change
to mono_foo and removing from mono_foo_internal.

Going forward though the wrapping could diverge further. For example, all the _internal functions could only accept/return coop handles, whereas currently they handle a lot of raw pointers.

This gives us leeway to limit gc mode changes to embedders,
and more generally to separate the runtime from the embedding API.

### DIFF
--- a/mono/metadata/domain.c
+++ b/mono/metadata/domain.c
@@ -227,15 +227,6 @@ mono_install_free_domain_hook (MonoFreeDomainFunc func)
 	free_domain_hook = func;
 }
 
-/**
- * mono_string_equal:
- * \param s1 First string to compare
- * \param s2 Second string to compare
- *
- * Compares two \c MonoString* instances ordinally for equality.
- *
- * \returns FALSE if the strings differ.
- */
 gboolean
 mono_string_equal_internal (MonoString *s1, MonoString *s2)
 {
@@ -250,19 +241,21 @@ mono_string_equal_internal (MonoString *s1, MonoString *s2)
 	return memcmp (mono_string_chars_internal (s1), mono_string_chars_internal (s2), l1 * 2) == 0;
 }
 
+/**
+ * mono_string_equal:
+ * \param s1 First string to compare
+ * \param s2 Second string to compare
+ *
+ * Compares two \c MonoString* instances ordinally for equality.
+ *
+ * \returns FALSE if the strings differ.
+ */
 gboolean
 mono_string_equal (MonoString *s1, MonoString *s2)
 {
 	MONO_EXTERNAL_ONLY (gboolean, mono_string_equal_internal (s1, s2));
 }
 
-/**
- * mono_string_hash:
- * \param s the string to hash
- *
- * Compute the hash for a \c MonoString*
- * \returns the hash for the string.
- */
 guint
 mono_string_hash_internal (MonoString *s)
 {
@@ -278,6 +271,13 @@ mono_string_hash_internal (MonoString *s)
 	return h;	
 }
 
+/**
+ * mono_string_hash:
+ * \param s the string to hash
+ *
+ * Compute the hash for a \c MonoString*
+ * \returns the hash for the string.
+ */
 guint
 mono_string_hash (MonoString *s)
 {
--- a/mono/metadata/domain.c
+++ b/mono/metadata/domain.c
@@ -50,6 +50,7 @@
 #include <mono/metadata/profiler-private.h>
 #include <mono/metadata/coree.h>
 #include <mono/utils/mono-experiments.h>
+#include "external-only.h"
 
 //#define DEBUG_DOMAIN_UNLOAD 1
 
@@ -63,7 +64,7 @@
 } while (FALSE)
 
 #define GET_APPCONTEXT() (mono_thread_internal_current ()->current_appcontext)
-#define SET_APPCONTEXT(x) MONO_OBJECT_SETREF (mono_thread_internal_current (), current_appcontext, (x))
+#define SET_APPCONTEXT(x) MONO_OBJECT_SETREF_INTERNAL (mono_thread_internal_current (), current_appcontext, (x))
 
 static guint16 appdomain_list_size = 0;
 static guint16 appdomain_next = 0;
@@ -236,17 +237,23 @@ mono_install_free_domain_hook (MonoFreeDomainFunc func)
  * \returns FALSE if the strings differ.
  */
 gboolean
-mono_string_equal (MonoString *s1, MonoString *s2)
+mono_string_equal_internal (MonoString *s1, MonoString *s2)
 {
-	int l1 = mono_string_length (s1);
-	int l2 = mono_string_length (s2);
+	int l1 = mono_string_length_internal (s1);
+	int l2 = mono_string_length_internal (s2);
 
 	if (s1 == s2)
 		return TRUE;
 	if (l1 != l2)
 		return FALSE;
 
-	return memcmp (mono_string_chars (s1), mono_string_chars (s2), l1 * 2) == 0; 
+	return memcmp (mono_string_chars_internal (s1), mono_string_chars_internal (s2), l1 * 2) == 0;
+}
+
+gboolean
+mono_string_equal (MonoString *s1, MonoString *s2)
+{
+	MONO_EXTERNAL_ONLY (gboolean, mono_string_equal_internal (s1, s2));
 }
 
 /**
@@ -257,10 +264,10 @@ mono_string_equal (MonoString *s1, MonoString *s2)
  * \returns the hash for the string.
  */
 guint
-mono_string_hash (MonoString *s)
+mono_string_hash_internal (MonoString *s)
 {
-	const gunichar2 *p = mono_string_chars (s);
-	int i, len = mono_string_length (s);
+	const gunichar2 *p = mono_string_chars_internal (s);
+	int i, len = mono_string_length_internal (s);
 	guint h = 0;
 
 	for (i = 0; i < len; i++) {
@@ -269,6 +276,12 @@ mono_string_hash (MonoString *s)
 	}
 
 	return h;	
+}
+
+guint
+mono_string_hash (MonoString *s)
+{
+	MONO_EXTERNAL_ONLY (guint, mono_string_hash_internal (s));
 }
 
 static gboolean
@@ -423,14 +436,14 @@ mono_domain_create (void)
 	domain->mp = mono_mempool_new ();
 	domain->code_mp = mono_code_manager_new ();
 	domain->lock_free_mp = lock_free_mempool_new ();
-	domain->env = mono_g_hash_table_new_type ((GHashFunc)mono_string_hash, (GCompareFunc)mono_string_equal, MONO_HASH_KEY_VALUE_GC, MONO_ROOT_SOURCE_DOMAIN, domain, "Domain Environment Variable Table");
+	domain->env = mono_g_hash_table_new_type ((GHashFunc)mono_string_hash_internal, (GCompareFunc)mono_string_equal_internal, MONO_HASH_KEY_VALUE_GC, MONO_ROOT_SOURCE_DOMAIN, domain, "Domain Environment Variable Table");
 	domain->domain_assemblies = NULL;
 	domain->assembly_bindings = NULL;
 	domain->assembly_bindings_parsed = FALSE;
 	domain->class_vtable_array = g_ptr_array_new ();
 	domain->proxy_vtable_hash = g_hash_table_new ((GHashFunc)mono_ptrarray_hash, (GCompareFunc)mono_ptrarray_equal);
 	mono_jit_code_hash_init (&domain->jit_code_hash);
-	domain->ldstr_table = mono_g_hash_table_new_type ((GHashFunc)mono_string_hash, (GCompareFunc)mono_string_equal, MONO_HASH_KEY_VALUE_GC, MONO_ROOT_SOURCE_DOMAIN, domain, "Domain String Pool Table");
+	domain->ldstr_table = mono_g_hash_table_new_type ((GHashFunc)mono_string_hash_internal, (GCompareFunc)mono_string_equal_internal, MONO_HASH_KEY_VALUE_GC, MONO_ROOT_SOURCE_DOMAIN, domain, "Domain String Pool Table");
 	domain->num_jit_info_table_duplicates = 0;
 	domain->jit_info_table = mono_jit_info_table_new (domain);
 	domain->jit_info_free_queue = NULL;
@@ -923,7 +936,7 @@ mono_domain_set_internal_with_options (MonoDomain *domain, gboolean migrate_exce
 			return;
 
 		g_assert (thread->abort_exc->object.vtable->domain != domain);
-		MONO_OBJECT_SETREF (thread, abort_exc, mono_get_exception_thread_abort ());
+		MONO_OBJECT_SETREF_INTERNAL (thread, abort_exc, mono_get_exception_thread_abort ());
 		g_assert (thread->abort_exc->object.vtable->domain == domain);
 	}
 }
--- a/mono/metadata/external-only.c
+++ b/mono/metadata/external-only.c
@@ -17,19 +17,20 @@
 #include "marshal.h"
 #include "object.h"
 
-/* GC handles support
+/**
+ * mono_gchandle_new:
+ * \param obj managed object to get a handle for
+ * \param pinned whether the object should be pinned
+ * This returns a handle that wraps the object, this is used to keep a
+ * reference to a managed object from the unmanaged world and preventing the
+ * object from being disposed.
  *
- * A handle can be created to refer to a managed object and either prevent it
- * from being garbage collected or moved or to be able to know if it has been
- * collected or not (weak references).
- * mono_gchandle_new () is used to prevent an object from being garbage collected
- * until mono_gchandle_free() is called. Use a TRUE value for the pinned argument to
- * prevent the object from being moved (this should be avoided as much as possible
- * and this should be used only for shorts periods of time or performance will suffer).
- * To create a weakref use mono_gchandle_new_weakref (): track_resurrection should
- * usually be false (see the GC docs for more details).
- * mono_gchandle_get_target () can be used to get the object referenced by both kinds
- * of handle: for a weakref handle, if an object has been collected, it will return NULL.
+ * If \p pinned is false the address of the object can not be obtained, if it is
+ * true the address of the object can be obtained.  This will also pin the
+ * object so it will not be possible by a moving garbage collector to move the
+ * object.
+ *
+ * \returns a handle that can be used to access the object from unmanaged code.
  */
 uint32_t
 mono_gchandle_new (MonoObject *obj, mono_bool pinned)
@@ -37,18 +38,57 @@ mono_gchandle_new (MonoObject *obj, mono_bool pinned)
 	MONO_EXTERNAL_ONLY (uint32_t, mono_gchandle_new_internal (obj, pinned));
 }
 
+/**
+ * mono_gchandle_new_weakref:
+ * \param obj managed object to get a handle for
+ * \param track_resurrection Determines how long to track the object, if this is set to TRUE, the object is tracked after finalization, if FALSE, the object is only tracked up until the point of finalization.
+ *
+ * This returns a weak handle that wraps the object, this is used to
+ * keep a reference to a managed object from the unmanaged world.
+ * Unlike the \c mono_gchandle_new_internal the object can be reclaimed by the
+ * garbage collector.  In this case the value of the GCHandle will be
+ * set to zero.
+ *
+ * If \p track_resurrection is TRUE the object will be tracked through
+ * finalization and if the object is resurrected during the execution
+ * of the finalizer, then the returned weakref will continue to hold
+ * a reference to the object.   If \p track_resurrection is FALSE, then
+ * the weak reference's target will become NULL as soon as the object
+ * is passed on to the finalizer.
+ *
+ * \returns a handle that can be used to access the object from
+ * unmanaged code.
+ */
 uint32_t
 mono_gchandle_new_weakref (MonoObject *obj, mono_bool track_resurrection)
 {
 	MONO_EXTERNAL_ONLY (uint32_t, mono_gchandle_new_weakref_internal (obj, track_resurrection));
 }
 
+/**
+ * mono_gchandle_get_target:
+ * \param gchandle a GCHandle's handle.
+ *
+ * The handle was previously created by calling \c mono_gchandle_new_internal or
+ * \c mono_gchandle_new_weakref.
+ *
+ * \returns a pointer to the \c MonoObject* represented by the handle or
+ * NULL for a collected object if using a weakref handle.
+ */
 MonoObject*
 mono_gchandle_get_target (uint32_t gchandle)
 {
 	MONO_EXTERNAL_ONLY (MonoObject*, mono_gchandle_get_target_internal (gchandle));
 }
 
+/**
+ * mono_gchandle_free:
+ * \param gchandle a GCHandle's handle.
+ *
+ * Frees the \p gchandle handle.  If there are no outstanding
+ * references, the garbage collector can reclaim the memory of the
+ * object wrapped.
+ */
 void
 mono_gchandle_free (uint32_t gchandle)
 {
--- a/mono/metadata/external-only.c
+++ b/mono/metadata/external-only.c
@@ -69,7 +69,7 @@ mono_gchandle_new_weakref (MonoObject *obj, mono_bool track_resurrection)
  * mono_gchandle_get_target:
  * \param gchandle a GCHandle's handle.
  *
- * The handle was previously created by calling \c mono_gchandle_new_internal or
+ * The handle was previously created by calling \c mono_gchandle_new or
  * \c mono_gchandle_new_weakref.
  *
  * \returns a pointer to the \c MonoObject* represented by the handle or
--- a/mono/metadata/external-only.c
+++ b/mono/metadata/external-only.c
@@ -6,18 +6,100 @@
  * Copyright 2018 Microsoft
  * Licensed under the MIT license. See LICENSE file in the project root for full license information.
  */
+
+// FIXME In order to confirm this is all extern_only,
+// a variant of the runtime should be linked without it.
+
 #include "config.h"
 #include "class-internals.h"
 #include "object-internals.h"
+#include "class-init.h"
+#include "marshal.h"
+#include "object.h"
 
-/**
- * mono_array_length:
- * \param array a \c MonoArray*
- * \returns the total number of elements in the array. This works for
- * both vectors and multidimensional arrays.
+/* GC handles support
+ *
+ * A handle can be created to refer to a managed object and either prevent it
+ * from being garbage collected or moved or to be able to know if it has been
+ * collected or not (weak references).
+ * mono_gchandle_new () is used to prevent an object from being garbage collected
+ * until mono_gchandle_free() is called. Use a TRUE value for the pinned argument to
+ * prevent the object from being moved (this should be avoided as much as possible
+ * and this should be used only for shorts periods of time or performance will suffer).
+ * To create a weakref use mono_gchandle_new_weakref (): track_resurrection should
+ * usually be false (see the GC docs for more details).
+ * mono_gchandle_get_target () can be used to get the object referenced by both kinds
+ * of handle: for a weakref handle, if an object has been collected, it will return NULL.
  */
-uintptr_t
-mono_array_length (MonoArray *array)
+uint32_t
+mono_gchandle_new (MonoObject *obj, mono_bool pinned)
 {
-	return mono_array_length_internal (array);
+	MONO_EXTERNAL_ONLY (uint32_t, mono_gchandle_new_internal (obj, pinned));
+}
+
+uint32_t
+mono_gchandle_new_weakref (MonoObject *obj, mono_bool track_resurrection)
+{
+	MONO_EXTERNAL_ONLY (uint32_t, mono_gchandle_new_weakref_internal (obj, track_resurrection));
+}
+
+MonoObject*
+mono_gchandle_get_target (uint32_t gchandle)
+{
+	MONO_EXTERNAL_ONLY (MonoObject*, mono_gchandle_get_target_internal (gchandle));
+}
+
+void
+mono_gchandle_free (uint32_t gchandle)
+{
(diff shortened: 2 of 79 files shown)


mono/metadata/object.h
```diff
@@ -230,7 +230,7 @@ mono_monitor_enter           (MonoObject *obj);
 MONO_API void
 mono_monitor_enter_v4        (MonoObject *obj, char *lock_taken);
 
-MONO_API unsigned int
+MONO_API MONO_RT_EXTERNAL_ONLY unsigned
```
luhenry: Revert that to `unsigned int`

mono/metadata/object.h
```diff
@@ -175,13 +175,13 @@ mono_string_from_utf16	    (/*const*/ mono_unichar2 *data);
 MONO_API MONO_RT_EXTERNAL_ONLY MonoString *
 mono_string_from_utf32	    (/*const*/ mono_unichar4 *data);
 
-MONO_API mono_bool
+MONO_API MONO_RT_EXTERNAL_ONLY mono_bool
 mono_string_equal           (MonoString *s1, MonoString *s2);
 
-MONO_API unsigned int
+MONO_API MONO_RT_EXTERNAL_ONLY unsigned
```
luhenry: Revert that to `unsigned int`.

mono/metadata/object.h
```diff
@@ -69,11 +69,11 @@ typedef void	    (*MonoMainThreadFunc)    (void* user_data);
...
 	} while (0)
 
-MONO_API mono_unichar2 *mono_string_chars  (MonoString *s);
-MONO_API int            mono_string_length (MonoString *s);
+MONO_API MONO_RT_EXTERNAL_ONLY mono_unichar2 *mono_string_chars  (MonoString *s);
+MONO_API MONO_RT_EXTERNAL_ONLY int            mono_string_length (MonoString *s);
 
 MONO_API MONO_RT_EXTERNAL_ONLY MonoObject *
-mono_object_new		    (MonoDomain *domain, MonoClass *klass);
+mono_object_new (MonoDomain *domain, MonoClass *klass);
```
luhenry: Revert that.

mono/metadata/object.h
jaykrell: Same thing.

jaykrell:
Something wrong with profiler tests.

jaykrell:
Crash on this PR was apparently the hint to resolve csc crashes.

https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/11028/parsed_console/log.html

```Thread 20 (Thread 0x7f82c3fff700 (LWP 25840)):
#0  0x00007f8323c95b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00005588a228e9a6 in dump_native_stacktrace (ctx=ctx@entry=0x7f8319163ac0, signal=0x5588a2472b60 "SIGSEGV") at mini-posix.c:1092
#2  0x00005588a228eb20 in mono_dump_native_crash_info (signal=signal@entry=0x5588a2472b60 "SIGSEGV", ctx=ctx@entry=0x7f8319163ac0, info=info@entry=0x7f8319163bf0) at mini-posix.c:1115
#3  0x00005588a2225575 in mono_handle_native_crash (signal=signal@entry=0x5588a2472b60 "SIGSEGV", ctx=ctx@entry=0x7f8319163ac0, info=info@entry=0x7f8319163bf0) at mini-exceptions.c:3196
#4  0x00005588a21a7374 in mono_sigsegv_signal_handler_debug (_dummy=11, _info=0x7f8319163bf0, context=0x7f8319163ac0, debug_fault_addr=0xfcb36) at mini-runtime.c:3527
#5  <signal handler called>
#6  0x00007f8320072432 in System_Reflection_Metadata_BlobBuilder__ctor_int (this=..., capacity=1024) from /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mcs/class/lib/build/System.Reflection.Metadata.dll.so
#7  0x00007f832039282f in Microsoft_Cci_PooledBlobBuilder__ctor_int (this=..., size=1024) from /mnt/jenkins/workspace/test-mono-pull-req```

**Comment by lambdageek** [comment]
Could you explain why `object.c` includes `external-only.c` - why can't we compile `external-only.c` as a regular object file?

**Comment by jaykrell** [comment]
It seems to break things. It makes sense that it breaks things on Windows. I thought Unix would work.

On Windows, the normal static linker, functions marked __declspec(dllexport) are only exported, if you otherwise get the linker to bring in anything from the .obj, and then it sees the directives.

.def files do not have this problem and I have recommended them repeatedly.
With .def  files, the linker must resolve the symbol and export it, not just export it if it is found.

I thought Unix linkers were more eager, so was surprised at the breakage, but anyway, you  have to do it this way for Windows.

Another option is to list the .o/.obj on the link command line in mini, instead of burying it in a lib.

There is also a  very new switch to the Windows linker /wholearchive that makes it more Unix-like.

(and I forgot to update the project files again)

**Comment by jaykrell** [comment]
Profiler test failure not reproing locally.

File: mono/metadata/object.c [hunk 4]
```diff
@@ -55,6 +55,9 @@
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
 
+// Export do not work if compiled separately.
+#include "external-only.c"
```
luhenry: Well, I guess you will have to revert some of the breaking changes then 😉

jaykrell:
@monojenkins build Linux i386

jaykrell:
@monojenkins build Linux x64

jaykrell:
hm same failures with reduced change

mono/metadata/object.c
```diff
@@ -2834,19 +2852,23 @@ mono_upgrade_remote_class (MonoDomain *domain, MonoObjectHandle proxy_object, Mo
  * the instance of a callvirt of \p method.
  */
 MonoMethod*
-mono_object_get_virtual_method (MonoObject *obj_raw, MonoMethod *method)
+mono_object_get_virtual_method_internal (MonoObject *obj_raw, MonoMethod *method)
 {
 	HANDLE_FUNCTION_ENTER ();
 	MonoMethod *result;
-	MONO_ENTER_GC_UNSAFE;
```
jaykrell: Perhaps this PR should not have any GC mode change changes, i.e. do them in a separate PR.

mono/metadata/object.c
```diff
@@ -2834,19 +2852,23 @@ mono_upgrade_remote_class (MonoDomain *domain, MonoObjectHandle proxy_object, Mo
...
 	ERROR_DECL (error);
 	MONO_HANDLE_DCL (MonoObject, obj);
 	result = mono_object_handle_get_virtual_method (obj, method, error);
 	mono_error_assert_ok (error);
-	MONO_EXIT_GC_UNSAFE;
```
jaykrell: Perhaps this PR should not have any GC mode change changes, i.e. do them in a separate PR.

mono/metadata/object.c
```diff
@@ -3309,19 +3331,21 @@ mono_copy_value (MonoType *type, void *dest, void *value, int deref_pointer)
  * The value must be in the native format of the field type. 
  */
 void
-mono_field_set_value (MonoObject *obj, MonoClassField *field, void *value)
+mono_field_set_value_internal (MonoObject *obj, MonoClassField *field, void *value)
 {
-	MONO_ENTER_GC_UNSAFE;
```
jaykrell: Perhaps this PR should not have any GC mode change changes, i.e. do them in a separate PR.

mono/metadata/object.c
```diff
@@ -3333,17 +3357,15 @@ mono_field_set_value (MonoObject *obj, MonoClassField *field, void *value)
  * The value must be in the native format of the field type. 
  */
 void
-mono_field_static_set_value (MonoVTable *vt, MonoClassField *field, void *value)
+mono_field_static_set_value_internal (MonoVTable *vt, MonoClassField *field, void *value)
 {
-	MONO_ENTER_GC_UNSAFE;
```
jaykrell: Perhaps this PR should not have any GC mode change changes, i.e. do them in a separate PR.

mono/metadata/object.c
```diff
@@ -3421,15 +3447,13 @@ mono_field_get_addr (MonoObject *obj, MonoVTable *vt, MonoClassField *field)
...
  * int i;
  *
- * mono_field_get_value (obj, int_field, &i);
+ * mono_field_get_value_internal (obj, int_field, &i);
  * </pre>
  */
 void
 mono_field_get_value (MonoObject *obj, MonoClassField *field, void *value)
 {
-	MONO_ENTER_GC_UNSAFE;
```
jaykrell: Perhaps this PR should not have any GC mode change changes, i.e. do them in a separate PR.

jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build Linux x64

jaykrell:
CI looking promising now.

mono/metadata/object.c
```diff
@@ -54,6 +54,13 @@
 #include "cominterop.h"
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
+#include "external-only.h"
+#include "monitor.h"
+
+// Exports do not work portably/reliably if compiled separately.
+// They must be in .o/.obj files otherwise referenced.
+// .def files work better in this regard.
+#include "external-only.c"
```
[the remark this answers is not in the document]
lambdageek: > [Exports] must be in .o/.obj files otherwise referenced.

I have no idea what this means.

lambdageek: Discussed with Jay.  `#include "external-only.c"` can go away.  We don't quite have the problem like in https://github.com/mono/mono/pull/11188/files#r226004001 because we don't build intermediate `.lib` on Windows.  We should revisit using `.def` files another day.

jaykrell:
@zoltan since you raised the issue?

`#define` vs. edit?
`m_foo`?
`mi_foo`?

jaykrell:
@monojenkins build  Linux i386

jaykrell:
The Unix linker is behaving how I described the Windows linker works.
The functions in external-only.c are not found by the test, unless the include is restored.

lambdageek:
> The Unix linker is behaving how I described the Windows linker works.
> The functions in external-only.c are not found by the test, unless the include is restored.

But object.c is in the same libraries as external-only.c  (isn't it?) how can that possibly change anything?

jaykrell:
Because we happen to call stuff in object.c.

jaykrell:
I'm *hoping* `#include "external-only.c"` is acceptable to reviewers but I am reluctant to *assume* it.

jaykrell:
Is the object.c factor clear?

The linker does not crack open .a files eagerly, and read of all of their constituent .o files, to export everything proactively.

It only reads the .o files within .a files if they resolve unresolved symbols.

Listing extra .a files on the linker command line, or putting extra .o files in .a files, does not cause stuff to be exported.

Putting extra stuff in object.c does make a difference, because we already require symbols from object.c or other reason. The linker has to resolve the entry points (main, DllMain, and .def file) and references from .o files given directly on the command line (not in .a files), and from there there is a graph walk.

It might be different for executable vs. shared libraries.
It isn't for Windows linker -- Windows linker has a simpler adequate more consistent model,  executables and shared libraries are indistinguishable.

Or there is an "entire archive" functionality that kicks in sometimes.
I thought it was always on Unix but clearly not.
I can build a small example on Mac like the above Windows and it behaves like Windows, and this test using dlsym against external_only symbols demonstrates it.

lambdageek:
Thanks for clarifying, Jay. Frustrating that we have to play these games with the linker, but it sounds like we don't have any options that are less involved. I don't mind the include now.

jaykrell:
There was MONO_EXTERNAL_ONLY_GC_UNSAFE.
It served to somewhat automate the tedium of ENTER_GC_UNSAFE / EXIT_GC_UNSAFE.
Then, we don't always switch the modes, so there is MONO_EXTERNAL_ONLY w/o the GC_UNSAFE.

This serves to be easily move back and forth to have the GC mode change or not, and it serves to mark the function as external_only.

These are minor things and they can be changed.
All of the macros could go away.

The reason for VOID vs. regular is because, without additional known machinery,  the macro has to know to introduce a local to assign to and return.

As to external-only.c, again, if the runtime does not reference a symbol within a source/.o file, then exports within that source/.o file are not exported. Unless we use .def files. Or move the .o files out of the .a files and link to them directly.  Or move the symbols to files that do have used symbols. include kinda does that last step, while maintaining a source organization that isolates external_only symbols -- if that is even desirable.

If we don't wish to separate them, then they can go to gc.c or such.

These particular symbols, if placed near what they replace, would be duplicated approx three times each -- null-gc.c, boehm-gc.c sgen-gc.c.

We needed it before, or not, *if* there were source files that no used symbols and only external_only symbols, or if we didn't notice that external_only symbols were disappearing.
One of those conditions was probably always true -- by leaving the functions where they were, they'd be in source files with used symbols and therefore still get exported.

It is quite easy to miss the loss of external_only symbols in  CI in fact.
We only notice them here, because there is one test that uses dlopen/dlsym to get to them, rather than calling them more directly. If you call them directly, you get a compiler error and you move to approved symbols.

Perhaps the mistake is in creating the external-only.c file, which doesn't work well, but seemed like a good idea.

As well, there is the pattern of replacing mono_foo with mono_foo_internal.
There are several options here and I am open to several.
We could m_foo, mi_foo, macros, etc.

jaykrell:
Anything else needed here?

Should I just move the external-only.c content to gc.c?
  Or continue in that direction in a later PR and move stuff to external-only.c?
  It had been said in the past that it might be good to move external_only code out somehow, so I figured I'd go ahead with that. Initially more was moved here but I dialed back due to 1. harder to read diff 2. test failures.

Remove some/all the macros?

jaykrell:
@monojenkins build OS X i386

jaykrell:
Restated: We did not previously have any entirely unused source files. External-only.c is a new type of file that way.

jaykrell:
ok?

jaykrell:
Speaking to @kumpera, he deferred to @lambdageek.
Given no changes since @lambdageek approval, seems ok then.

mono/metadata/external-only.c
```diff
@@ -6,18 +6,100 @@
...
+ * prevent the object from being moved (this should be avoided as much as possible
+ * and this should be used only for shorts periods of time or performance will suffer).
+ * To create a weakref use mono_gchandle_new_weakref (): track_resurrection should
+ * usually be false (see the GC docs for more details).
+ * mono_gchandle_get_target () can be used to get the object referenced by both kinds
+ * of handle: for a weakref handle, if an object has been collected, it will return NULL.
  */
-uintptr_t
-mono_array_length (MonoArray *array)
+uint32_t
```
lambdageek: Hm.  you should probably copy the API doc comments from `sgen-mono.c`

mono/metadata/object.c
```diff
@@ -54,6 +54,13 @@
 #include "cominterop.h"
 #include <mono/utils/w32api.h>
 #include <mono/utils/unlocked.h>
+#include "external-only.h"
+#include "monitor.h"
+
+// Exports do not work portably/reliably if compiled separately.
+// They must be in .o/.obj files otherwise referenced.
+// .def files work better in this regard.
```
lambdageek: Maybe it doesn't hurt to be a bit more verbose here.  Something like:

```
// If an object file full of exported functions that are not called anywhere is included in
// an intermediate static library, and the library is in turn linked into another binary or library, the
// external only object fill will be dropped.  As a workaround, we include "external-only.c" here.  Another
// workaround on Windows would be to use .def files to specify exports.  On Unix, we would need to link
// external-only.o directly into the final binary or shared library.
```

jaykrell: Unix has .def files or equivalent also.

jaykrell: And listing .o files on link command line works on Windows. I'll wordsmith a bit.

mono/metadata/external-only.c
```diff
@@ -6,18 +6,140 @@
...
+mono_gchandle_new_weakref (MonoObject *obj, mono_bool track_resurrection)
+{
+	MONO_EXTERNAL_ONLY (uint32_t, mono_gchandle_new_weakref_internal (obj, track_resurrection));
+}
+
+/**
+ * mono_gchandle_get_target:
+ * \param gchandle a GCHandle's handle.
+ *
+ * The handle was previously created by calling \c mono_gchandle_new_internal or
```
lambdageek:
```suggestion
 * The handle was previously created by calling \c mono_gchandle_new or
```

mono/metadata/domain.c
```diff
@@ -271,6 +278,12 @@ mono_string_hash (MonoString *s)
 	return h;	
 }
 
+guint
+mono_string_hash (MonoString *s)
```
lambdageek: Lost api doc comment.

mono/metadata/domain.c
```diff
@@ -236,17 +237,23 @@ mono_install_free_domain_hook (MonoFreeDomainFunc func)
...
 		return TRUE;
 	if (l1 != l2)
 		return FALSE;
 
-	return memcmp (mono_string_chars (s1), mono_string_chars (s2), l1 * 2) == 0; 
+	return memcmp (mono_string_chars_internal (s1), mono_string_chars_internal (s2), l1 * 2) == 0;
+}
+
+gboolean
+mono_string_equal (MonoString *s1, MonoString *s2)
```
lambdageek: lost API doc comment.

mono/metadata/handle.h
```diff
@@ -614,24 +606,12 @@ mono_gchandle_from_handle (MonoObjectHandle handle, mono_bool pinned);
...
 
-static inline gboolean
-mono_gchandle_target_equal (uint32_t gchandle, MonoObjectHandle equal)
-{
-	// This function serves to reduce coop handle creation.
-	MONO_HANDLE_SUPPRESS_SCOPE (1);
-	return mono_gchandle_get_target (gchandle) == MONO_HANDLE_RAW (equal);
-}
+gboolean
+mono_gchandle_target_equal (uint32_t gchandle, MonoObjectHandle equal);
```
lambdageek: Why? this isn't a public header

lambdageek: You can just update this inline function (and the others in handle.h to call the _internal functions.

mono/metadata/monitor.c
```diff
@@ -634,6 +635,12 @@ mono_object_hash (MonoObject* obj)
 #endif
 }
 
+int
+mono_object_hash (MonoObject* obj)
```
lambdageek: api doc comment.

mono/metadata/object.c
```diff
@@ -2834,19 +2856,23 @@ mono_upgrade_remote_class (MonoDomain *domain, MonoObjectHandle proxy_object, Mo
...
 	ERROR_DECL (error);
 	MONO_HANDLE_DCL (MonoObject, obj);
 	result = mono_object_handle_get_virtual_method (obj, method, error);
 	mono_error_assert_ok (error);
-	MONO_EXIT_GC_UNSAFE;
 	HANDLE_FUNCTION_RETURN_VAL (result);
 }
 
+MonoMethod*
+mono_object_get_virtual_method (MonoObject *obj, MonoMethod *method)
```
lambdageek: api doc comment.

mono/metadata/object.c
```diff
@@ -3309,19 +3335,21 @@ mono_copy_value (MonoType *type, void *dest, void *value, int deref_pointer)
...
+		return;
 
 	dest = (char*)obj + field->offset;
 	mono_copy_value (field->type, dest, value, FALSE);
-leave:
-	MONO_EXIT_GC_UNSAFE;
+}
+
+void
+mono_field_set_value (MonoObject *obj, MonoClassField *field, void *value)
```
lambdageek: api doc comment.

mono/metadata/object.c
```diff
@@ -3357,8 +3383,12 @@ mono_field_static_set_value (MonoVTable *vt, MonoClassField *field, void *value)
 		dest = (char*)mono_vtable_get_static_field_data (vt) + field->offset;
 	}
 	mono_copy_value (field->type, dest, value, FALSE);
-leave:
-	MONO_EXIT_GC_UNSAFE;
+}
+
+void
+mono_field_static_set_value (MonoVTable *vt, MonoClassField *field, void *value)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -3421,15 +3451,13 @@ mono_field_get_addr (MonoObject *obj, MonoVTable *vt, MonoClassField *field)
  * <pre>
  * int i;
  *
- * mono_field_get_value (obj, int_field, &i);
+ * mono_field_get_value_internal (obj, int_field, &i);
```
lambdageek:
```suggestion
 * mono_field_get_value (obj, int_field, &i);
```

mono/metadata/object.c
```diff
@@ -4067,6 +4103,12 @@ mono_get_delegate_invoke (MonoClass *klass)
 	return result;
 }
 
+MonoMethod*
+mono_get_delegate_invoke (MonoClass *klass)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -4106,6 +4148,12 @@ mono_get_delegate_begin_invoke (MonoClass *klass)
 	return result;
 }
 
+MonoMethod*
+mono_get_delegate_begin_invoke (MonoClass *klass)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -4145,6 +4193,12 @@ mono_get_delegate_end_invoke (MonoClass *klass)
 	return result;
 }
 
+MonoMethod*
+mono_get_delegate_end_invoke (MonoClass *klass)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -4796,6 +4850,12 @@ mono_unhandled_exception (MonoObject *exc_raw)
 	HANDLE_FUNCTION_RETURN ();
 }
 
+void
+mono_unhandled_exception (MonoObject *exc)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6363,17 +6422,23 @@ mono_string_empty_wrapper (void)
...
-mono_string_empty (MonoDomain *domain)
+mono_string_empty_internal (MonoDomain *domain)
 {
 	g_assert (domain);
 	g_assert (domain->empty_string);
 	return domain->empty_string;
 }
 
+MonoString*
+mono_string_empty (MonoDomain *domain)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6720,14 +6781,15 @@ mono_string_new_wtf8_len_checked (MonoDomain *domain, const char *text, guint le
...
-	MonoString *res;
-	MONO_ENTER_GC_UNSAFE;
+	return mono_string_new_internal (mono_domain_get (), text);
+}
 
-	res = mono_string_new_internal (mono_domain_get (), text);
-	MONO_EXIT_GC_UNSAFE;
-	return res;
+MonoString*
+mono_string_new_wrapper (const char *text)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6890,11 +6952,17 @@ mono_value_box_checked (MonoDomain *domain, MonoClass *klass, gpointer value, Mo
...
+mono_value_copy_internal (gpointer dest, gpointer src, MonoClass *klass)
 {
 	MONO_REQ_GC_UNSAFE_MODE;
 
-	mono_gc_wbarrier_value_copy (dest, src, 1, klass);
+	mono_gc_wbarrier_value_copy_internal (dest, src, 1, klass);
+}
+
+void
+mono_value_copy (gpointer dest, gpointer src, MonoClass *klass)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6908,36 +6976,54 @@ mono_value_copy (gpointer dest, gpointer src, MonoClass *klass)
...
 
 	int size = mono_array_element_size (dest->obj.vtable->klass);
 	char *d = mono_array_addr_with_size_fast (dest, size, dest_idx);
 	g_assert (size == mono_class_value_size (m_class_get_element_class (mono_object_class (dest)), NULL));
-	mono_gc_wbarrier_value_copy (d, src, count, m_class_get_element_class (mono_object_class (dest)));
+	mono_gc_wbarrier_value_copy_internal (d, src, count, m_class_get_element_class (mono_object_class (dest)));
+}
+
+void
+mono_value_copy_array (MonoArray *dest, int dest_idx, void* src, int count)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6908,36 +6976,54 @@ mono_value_copy (gpointer dest, gpointer src, MonoClass *klass)
...
-mono_object_get_domain (MonoObject *obj)
+mono_object_get_domain_internal (MonoObject *obj)
 {
 	MONO_REQ_GC_UNSAFE_MODE;
 
 	return mono_object_domain (obj);
 }
 
+MonoDomain*
+mono_object_get_domain (MonoObject *obj)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6981,6 +7063,12 @@ mono_object_get_size (MonoObject* o)
 	}
 }
 
+unsigned
+mono_object_get_size (MonoObject *o)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -6990,22 +7078,23 @@ mono_object_get_size (MonoObject* o)
...
 	/* add assert for valuetypes? */
 	g_assert (m_class_is_valuetype (mono_object_class (obj)));
-	res = mono_object_get_data (obj);
-	MONO_EXIT_GC_UNSAFE;
-	return res;
+	return mono_object_get_data (obj);
+}
+
+void*
+mono_object_unbox (MonoObject *obj)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -7275,6 +7363,12 @@ mono_string_is_interned (MonoString *o)
 	return result;
 }
 
+MonoString*
+mono_string_is_interned (MonoString *str)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -7555,13 +7647,19 @@ mono_string_to_utf8_checked (MonoString *s, MonoError *error)
 	if (!s->length)
 		return g_strdup ("");
 
-	return mono_utf16_to_utf8 (mono_string_chars (s), s->length, error);
+	return mono_utf16_to_utf8 (mono_string_chars_internal (s), s->length, error);
+}
+
+char*
+mono_string_to_utf8_checked (MonoString *string_obj, MonoError *error)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -7620,19 +7718,25 @@ mono_string_to_utf16 (MonoString *s)
...
 		as [length] = 0;
 		if (length)
-			memcpy (as, mono_string_chars(s), length * sizeof (*as));
+			memcpy (as, mono_string_chars_internal (s), length * sizeof (*as));
 	}
 	return as;
 }
 
+mono_unichar2*
+mono_string_to_utf16 (MonoString *string_obj)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -7642,6 +7746,12 @@ mono_string_to_utf32 (MonoString *s)
 	return g_utf16_to_ucs4 (s->chars, s->length, NULL, NULL, NULL);
 }
 
+mono_unichar4*
+mono_string_to_utf32 (MonoString *string_obj)
```
lambdageek: api doc comment

mono/metadata/object.c
```diff
@@ -7808,14 +7918,20 @@ mono_get_eh_callbacks (void)
...
+mono_raise_exception_internal (MonoException *ex) 
 {
 	/* raise_exception doesn't return, so the transition to GC Unsafe is unbalanced */
 	MONO_STACKDATA (stackdata);
 	mono_threads_enter_gc_unsafe_region_unbalanced_with_info (mono_thread_info_current (), &stackdata);
 	mono_raise_exception_deprecated (ex);
 }
 
+void
+mono_raise_exception (MonoException *ex)
```
lambdageek: api doc comment (IMPORTANT!)

mono/metadata/object.c
```diff
@@ -8342,6 +8458,12 @@ mono_print_unhandled_exception (MonoObject *exc)
 		g_free (message);
 }
 
+void
+mono_print_unhandled_exception (MonoObject *exc)
```
lambdageek: api doc comment

lambdageek:
I only commented on about 25 of the external-only functions in object.c but in general: if you add an _internal version of a function, keep the API doc comment together with the non-_internal version - that's the one that will be the API documentation.

mono/metadata/domain.c
jaykrell: Agreed, will fix.

mono/metadata/handle.h
jaykrell: Cannot include object-internals.h without causing more breaks, and didn't work them through.

mono/metadata/monitor.c
jaykrell: Agreed, will fix.a

mono/metadata/object.c
jaykrell: Agreed, will fix.

mono/metadata/object.c
jaykrell: Agreed, will fix.

mono/metadata/object.c
jaykrell: Agreed, will fix.

jaykrell:
Will duplicate (or move?) all the doc comments.

mono/metadata/handle.h
lambdageek: Ah ok.  gotcha.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build  Linux WebAssembly
@monojenkins build  Linux OS X x64 C++

jaykrell:

jaykrell:
@monojenkins build Linux OS X x64 C++

jaykrell:

PR #11196, opened 2018-10-16
Fix MessageBox text color for dark themes

HebaruSan:
## Problem

`MessageBox` displays as black-on-dark in a dark theme:

image

This can be difficult to read.

## Cause

The text is drawn with a hard coded color of `Color.Black`. This assumes that the background under the text will be light to contrast with black.

## Changes

Now the color is `SystemColors.ControlText` instead, to match the background of the popup.

In a normal theme this will still be black.


akoeplinger:
@monojenkins approve

Issue #11202, opened 2018-10-16
Mac installer does not always clear out old files


sandyarmstrong:
1. Install some build of Mono 5.16 with an `msbuild` that lacks https://github.com/Microsoft/msbuild/pull/3239 .
2. Note that `/Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/SdkResolvers/NuGet.MSBuildSdkResolver/` exists.
3. Install Mono 5.16.0.173.

### Expected results

`/Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/SdkResolvers/NuGet.MSBuildSdkResolver/` has been removed.

### Actual results

`/Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/SdkResolvers/NuGet.MSBuildSdkResolver/` still exists.

If you `sudo rm -rf /Library/Framework/Mono.framework` and then reinstall Mono 5.16.0.173, the `NuGet.MSBuildSdkResolver` directory will be gone, as it should be.

---

Installations should be deterministic. Consider modifying your preinstall script to completely delete `/Library/Frameworks/Mono.framework` each time. This is what we did for Workbooks.

sandyarmstrong:
At the very least, let's make sure that if I am installing some build of 5.16.0, `Library/Frameworks/Mono.framework/Versions/5.16.0` always gets cleared out.

PR #11203, opened 2018-10-16
[pkg] Add preinstall which removes existing Mono of the same version

akoeplinger:
Fixes https://github.com/mono/mono/issues/11202


akoeplinger:
@monojenkins build pkg

akoeplinger:
@luhenry it should be 5.16.0 since we're using the same variable to calculate the `Versions/X.X.X` path in the first place

akoeplinger:
I verified this contains the correct version string in the preinstall and it blows away the dir before installing so any changes made to the dir are cleaned too.

akoeplinger:
@monojenkins backport 2018-10
@monojenkins backport 2018-08

akoeplinger:
@monojenkins backport 2018-06

Issue #11146, opened 2018-10-13
Unhandled exception at the end of MonoDevelop.UnitTesting.Tests test run


kg:
At the end of test runs for the MonoDevelop.UnitTesting.Tests assembly these unhandled exceptions show up in the log:

```
2018-10-11T17:20:36.4689180Z Unhandled Exception:
2018-10-11T17:20:36.4689290Z System.ArgumentException: is negative
2018-10-11T17:20:36.4689370Z Parameter name: count
2018-10-11T17:20:36.4690230Z   at (wrapper managed-to-native) System.Buffer.InternalBlockCopy(System.Array,int,System.Array,int,int)
2018-10-11T17:20:36.4691200Z   at System.IO.StreamWriter.Write (System.Char[] buffer, System.Int32 index, System.Int32 count) [0x00094] in 
2018-10-11T17:20:36.4692330Z   at System.IO.TextWriter.WriteLine (System.String value) [0x00070] in 
2018-10-11T17:20:36.4693490Z   at System.IO.TextWriter.WriteLine (System.String format, System.Object arg0, System.Object arg1, System.Object arg2) [0x00000] in 
2018-10-11T17:20:36.4694280Z   at MonoDevelop.Core.Logging.FileLogger.Log (MonoDevelop.Core.Logging.LogLevel level, System.String message) [0x00051] in 
2018-10-11T17:20:36.4694670Z   at MonoDevelop.Core.LoggingService.Log (MonoDevelop.Core.Logging.LogLevel level, System.String message) [0x00022] in 
2018-10-11T17:20:36.4694960Z   at MonoDevelop.Core.LoggingService.LogError (System.String message) [0x00000] in 
2018-10-11T17:20:36.4695280Z   at MonoDevelop.DesignerSupport.ToolboxService.GetFileItems (MonoDevelop.DesignerSupport.Toolbox.LoaderContext ctx, System.String fileName) [0x0008e] in 
2018-10-11T17:20:36.4695880Z   at MonoDevelop.DesignerSupport.ToolboxService+<>c__DisplayClass20_0.<RegisterDefaultToolboxProvider>b__0 (System.Object <p0>) [0x0008b] in 
2018-10-11T17:20:36.4697190Z   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context (System.Object state) [0x0000d] in 
[...]

PR #11205, opened 2018-10-17
[corlib] Ensure that Console Streams are thread-safe

alexischr:
The Console class wraps the standard system stream reader/writers on construction so as to expose them as thread-safe. However, the Set-Out/In/Err methods in referencesource do not do the same when they replace default targets.

CoreFX code does: https://github.com/mono/corefx/blob/1748f6492e8b96842f76e1a9e7b5d7e8ccfca480/src/System.Console/src/System/Console.cs#L412-L444

Possible fix for https://github.com/mono/mono/issues/11146

### DIFF
--- a/mcs/class/corlib/System/Console.cs
+++ b/mcs/class/corlib/System/Console.cs
@@ -243,7 +243,7 @@ namespace System
 			if (newError == null)
 				throw new ArgumentNullException ("newError");
 
-			stderr = newError;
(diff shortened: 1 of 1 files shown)


marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2018-08

Issue #7246, opened 2018-02-23
CoreFX System.IO corlib namespace


marek-safar:
This one is quite tricky as we couldn't even migrate to reference source for most of the types but at the same time important as it got several performance optimizations and also includes new Span<T> APIs

## Milestone I

- [x] BufferedStream
- [x] DirectoryNotFoundException
- [x] DriveNotFoundException
- [x] DriveType
- [x] EndOfStreamException
- [x] FileAccess
- [x] FileAttributes
- [x] FileInfo
- [x] FileLoadException
- [x] FileMode
- [x] FileNotFoundException
- [x] FileOptions
- [x] FileShare
- [x] FileSystemInfo
- [x] IOException
- [x] MemoryStream
- [x] PathTooLongException
- [x] SearchOption
- [x] SeekOrigin
- [x] Stream
- [x] StreamReader
- [x] StreamWriter
- [x] StringReader
- [x] StringWriter
- [x] TextReader
- [x] TextWriter
- [x] UnmanagedMemoryAccessor
- [x] UnmanagedMemoryStream

## Milestone II
- [x] Directory
- [x] DirectoryInfo
- [x] File

## Future
- [ ] BinaryReader
- [ ] BinaryWriter
- [ ] DriveInfo
- [ ] FileStream
- [ ] Path

PR #11210, opened 2018-10-17
[corlib] Import System.IO.UnmanagedMemoryStream from CoreFX

alexischr:
Part of #7246


alexischr:

Issue #11123, opened 2018-10-12
Type.IsSubclassOf throws TypeLoadException


rjvdboon:
## Steps to Reproduce

[file.txt](https://github.com/mono/mono/files/2473140/file.txt)

1. Save the attached file as `file.cs`
2. Compile A.dll, B.dll, and Prog.exe as specified in the header of the file.
3. Run `mono Prog.exe`, observe no exception
4. rm A.dll
5. Run `mono Prog.exe` again, should have no exception, has a TypeLoadException.

While the steps to reproduce obviously cause runtime failure, this is also a problem in more legitimate scenarios, see Stacktrace section below for the exception we get in our testsuite.

### Current Behavior

On Mono 5.16.0.179 I get the following output with a TypeLoadException
```
.
Assembly mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
  contains 0 types
Assembly Prog, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
  contains 0 types
Assembly B, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
Caught UNexpected: System.TypeLoadException: Could not load type of field 'B.Two.Derived:myOne' (0) due to: Could not load file or assembly 'A, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
  at (wrapper managed-to-native) System.RuntimeTypeHandle.is_subclass_of(intptr,intptr)
  at System.RuntimeTypeHandle.IsSubclassOf (System.RuntimeType childType, System.RuntimeType baseType) [0x00016] in <406b9a7e0e504a4289a76df450bd6eef>:0
  at System.RuntimeType.IsSubclassOf (System.Type type) [0x00020] in <406b9a7e0e504a4289a76df450bd6eef>:0
  at Program.C+<>c__DisplayClass0_0.<Main>b__0 (System.Type t) [0x00000] in <3daa97e0a49f4911953cec2221f49a83>:0
  at System.Linq.Enumerable+WhereArrayIterator`1[TSource].GetCount (System.Boolean onlyIfCheap) [0x0001a] in <b7a52c68d5a5436ab599e3e3f31ab4b2>:0
  at System.Linq.Enumerable.Count[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00029] in <b7a52c68d5a5436ab599e3e3f31ab4b2>:0
[...]

PR #11227, opened 2018-10-17
[runtime] Don't init classes in ves_icall_RuntimeTypeHandle_is_subclass_of

lambdageek:
We must not get a TLE if referenced types are in an assembly that can't be
loaded.

Fixes https://github.com/mono/mono/issues/11123

TODO:
- [ ] Add test case.  (It's a bit annoying because we need to compile against an assembly that should be missing at execution time)


marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2018-08

marek-safar:
@monojenkins backport 2018-06

marek-safar:
@lambdageek could you please backport it manually?

for the test case, why don't compile it against reference assembly?

PR #11234, opened 2018-10-18
Windows BTLS assembler build support (cygwin/msvc).

lateralusX:
Add support for BTLS assembler build on Windows (cygwin/msvc). Building BTLS with assembler support on Windows requires several additional build tools (except cmake) for both cygwin/msvc builds.
On cygwin, perl and yasm will be needed and on msvc, ninja, perl and yasm. On msvc builds BTLS doesn’t support Visual Studio generator when building with assembler support, so this commit adds support for ninja build system covering that scenario. This commit adds support for auto detection of needed build tools and enable assembler builds if all needed tools have been identified as available. Commit also fixes the execution of msbuild on CI to just use the original PATH variable instead of the cygwin PATH to make sure correct version of identified build tools gets used (Windows version of git, ninja, perl and yasm). Both cygwin and msvc build will use nasm compiler (through use of yasm) since generated assembler won’t compile with mingw’s gcc version.

PR also include fix for BTLS password encoding on Windows that cause test failures when running on CI.


lateralusX:
All tests pass on CI using BTLS on Windows using BTLS build without assembler optimization.

lateralusX:
@baulig approve/merge?

lateralusX:
@baulig approve/merge?

lateralusX:
@baulig approve/merge?

lateralusX:
@baulig approve/merge?

lateralusX:
@baulig approve/merge?

Issue #11187, opened 2018-10-16
[wasm] [packager] Duplicate assemblies in `config.js` file_list: [ ... ]


kjpou1:
The `packager.exe` program allows duplicate assembly names.

Current behavior:

When `packager.exe` is executed against a `dll` dependent assemblies may be found multiple times while traversing the dependencies.  `packager.exe` does not verify if the assembly is already in the list of assemblies to load when writing the `config.js` file.  This creates a list of assemblies that will include the assembly name multiple times in `config.js` file_list: [] when passed to `MONO.mono_load_runtime_and_bcl`.

Expected:

`packager.exe` should not write duplicate assembly entries to the list of assemblies to load when writing the `config.js` file_list: [].

PR #11236, opened 2018-10-18
[wasm] [packager] Fix Duplicate assemblies in `config.js` file_list: [ ... ]

kjpou1:
packager.exe should not write duplicate assembly entries to the list of assemblies to load when writing the config.js file_list: [].

resolves: https://github.com/mono/mono/issues/11187


kjpou1:

Issue #11239, opened 2018-10-18
[wasm] Sample program in wasm sdks does not run.


kjpou1:
The sample is missing the loading of the `mono-config.js` file in the `sample.html`.

PR #11242, opened 2018-10-18
[wasm] Fix Sample.html so sample will run

kjpou1:
- Add `mono-config.js` reference to html file.

resolves: https://github.com/mono/mono/issues/11239

- packager.exe
  - Replaced the File Delete and Copy with a Copy if new call.
  - re-add the generation of a skeleton  `runtime.js` file if one can not be found.


vargaz:
@monojenkins build failed

PR #11246, opened 2018-10-18
[System] Don't throw NIE for ServicePoint.ConnectionLeaseTimeout and .ReceiveBufferSize

akoeplinger:
Instead return a no-op value. This is the same behavior as .NET Core.

Fixes https://github.com/dotnet/standard/issues/642


akoeplinger:
@luhenry this is also what .NET Core does and it came up in https://github.com/dotnet/standard/issues/642 that we're throwing an exception here. I'm not sure if we can even implement this on all platforms but at least we'd be consistent with .NET Core.

akoeplinger:
@monojenkins commit apidiff

PR #11253, opened 2018-10-18
[sdks] Package what's missing for XA

luhenry:
They require the additional files:
 - mono/eglib/eglib-config.h: that's to build `libmonodroid`
 - mcs/build/Consts.cs: that's to build `Mono.Posix` and `Mono.Data.Sqlite`


luhenry:
@akoeplinger removing the dependency is in the work. This is a shortcut to get the existing archive to work for 2018-08.

luhenry:

Issue #11132, opened 2018-10-12
HandleCollectorTests flaky


jaykrell:
## Steps to Reproduce

1. Submit a PR.

### Current Behavior

Failures for example:

https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/15414/testReport/junit/Test%20collection%20for%20System.Runtime.InteropServices/HandleCollectorTests/System_Runtime_InteropServices_HandleCollectorTests_TestHandleCollector/

```
Test collection for System.Runtime.InteropServices.HandleCollectorTests.System.Runtime.InteropServices.HandleCollectorTests.TestHandleCollector

Failing for the past 1 build (Since Unstable#15414 )
Took 27 ms.
Error Message
Low limit handle did not trigger a GC\nExpected: True\nActual:   False
Stacktrace
  at System.Runtime.InteropServices.HandleCollectorTests.TestHandleCollector () [0x00055] in /Users/builder/jenkins/workspace/test-mono-pull-request-i386-osx/external/corefx/src/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/HandleCollectorTests.cs:136 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Users/builder/jenkins/workspace/test-mono-pull-request-i386-osx/mcs/class/corlib/System.Reflection/MonoMethod.cs:324 
```

### Expected Behavior

Success.

jaykrell:
https://jenkins.mono-project.com/job/x/14676/testReport/junit/Test%20collection%20for%20System.Runtime.InteropServices/HandleCollectorTests/System_Runtime_InteropServices_HandleCollectorTests_TestHandleCollector/

lambdageek:
Another instance https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/15725/testReport/junit/Test%20collection%20for%20System.Runtime.InteropServices/HandleCollectorTests/System_Runtime_InteropServices_HandleCollectorTests_TestHandleCollector/

akoeplinger:
[...]

PR #11268, opened 2018-10-19
[class] Ensure 'cast_class' is initialized for array type classes.

alexischr:
Fixes https://github.com/mono/mono/issues/9507

Background: Monoclass:cast_class is used internally to allow for some interfaces to be assignable to each other if they share a "storage class" instead of the class's element class. This is used to i.e. implement IList<ulong> being assignable from IList<long> (IList<ulong> has cast_class=long)

Consider a GenericClass<T> with a T[] field. On instantiation with e.g. GenericClass<ulong> , the ulong[] array class will be generated. As above, ulong[] has cast_class=long
When mono_class_init() is called on an array class, its element_class is also initialized.
If GenericClass<T>:Method() uses any type/subtype checking (e.g. if (field is double[]) ), those may attempt to use cast_class, which is not ensured as initialized.

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -4159,9 +4159,15 @@ mono_class_init (MonoClass *klass)
 	klass_byval_arg = m_class_get_byval_arg (klass);
 	if (klass_byval_arg->type == MONO_TYPE_ARRAY || klass_byval_arg->type == MONO_TYPE_SZARRAY) {
 		MonoClass *element_class = klass->element_class;
+		MonoClass *cast_class = klass->cast_class;
+
 		if (!element_class->inited) 
 			mono_class_init (element_class);
 		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array element class"))
+			goto leave;
+		if (!cast_class->inited)
+			mono_class_init (cast_class);
+		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array cast class"))
 			goto leave;
 	}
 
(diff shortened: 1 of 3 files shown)


lambdageek:
If it works, I like this one better than #10626 - initializing `cast_class` right when the array type is initialized is less surprising than initializing a `cast_class` of a generic instantiation argument.

kumpera:
Yeah, I agree that initing the element class makes more sense.

mono/metadata/class-init.c
```diff
@@ -4159,10 +4159,16 @@ mono_class_init (MonoClass *klass)
...
 		MonoClass *element_class = klass->element_class;
+		MonoClass *cast_class = klass->cast_class;
+
 		if (!element_class->inited) 
 			mono_class_init (element_class);
 		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array element class"))
 			goto leave;
+		if (!cast_class->inited)
+			mono_class_init (cast_class);
+		if (mono_class_set_type_load_failure_causedby_class (klass, element_class, "Could not load array cast class"))
```
lambdageek:
```suggestion
		if (mono_class_set_type_load_failure_causedby_class (klass, cast_class, "Could not load array cast class"))
```

alexischr:
Failures are unrelated flaky tests. Tests will now restart as I accept the commit for the typo. @lambdageek @kumpera I will merge if we're good

lambdageek:
OSX failure in System.Runtime.InteropServices.HandleCollectorTests.System.Runtime.InteropServices.HandleCollectorTests.TestHandleCollector is #11132

PR #11273, opened 2018-10-19
[wasm] Don't set the assembly load path as we no longer use mmap-base…

kumpera:
…d assembly loading.


lewurm:
superseded by https://github.com/mono/mono/pull/11291 ?

PR #11281, opened 2018-10-19
[runtime] Fix mismatched visibility modifiers on class Unsafe

alexanderkyte:
This brings it into alignment with the visibility in Unsafe.il


marek-safar:
This type is intentionally private

PR #11287, opened 2018-10-19
Really fix ToolStrip background

HebaruSan:
## Problem

Even after #10942, menus and toolstrips still look awful with a dark theme:

toolstrip

## Previous attempt

#10942 was a first stab in the dark at what was going on here, before I managed to get mono to compile for testing. I was hopeful that it would address this, but the problem was different.

## Another red herring

Later I thought the problem might be that `SystemColors.ButtonFace.IsSystemColor` was `false`. It looked like this might have caused the runtime to interpret `ButtonFace`'s special 168 value as an RGB code instead of a system color marker.

This fix had no effect (see below for true cause), but I'll include it anyway; the lower bound of the upper system colors block needs to decrease from 170 to 168 to allow `ButtonFace` and `ButtonHighlight` to be considered system colors:

https://github.com/mono/mono/blob/44db64c620c2c957a49b3f5a69cb09c19c1c8c1a/mcs/class/System.Drawing/System.Drawing/KnownColor.cs#L200-L203

## Actual cause

The toolstrip color comes from `ProfessionalColorTable.ToolStripGradientBegin` and `ProfessionalColorTable.ToolStripGradientEnd`. The latter color is hard coded to a light shade, but the former defaults to `SystemColors.ButtonFace`.

These are the colors that are loaded from the GTK system theme:

https://github.com/mono/mono/blob/b104565ccfde477fc7b6308f2d3e4efff0af5848/mcs/class/System.Windows.Forms/System.Windows.Forms/X11DesktopColors.cs#L111-L116

Note that `ButtonFace` is not among them. This means that it uses a compile time default, and is not a good choice for the background of a control. `ToolStrip` is using it anyway, but other controls use `SystemColors.Control` instead.

## Changes

There are two obvious ways to fix this:

1. Use `SystemColors.Control` instead of `SystemColors.ButtonFace` in `ProfessionalColorTable`
2. Load `SystemColors.ButtonFace` from the system theme

This pull request does both. In addition, it also sets the other end of the toolstrip gradients to `SystemColors.ControlLight` instead of a hard coded value. The choice of `ControlLight` was made because in those cases where alternate hard coded values were available for both sides of the gradient, the "other" side was consistently lighter. Now it looks better:

image

### DIFF
--- a/mcs/class/System.Drawing/System.Drawing/Color.cs
+++ b/mcs/class/System.Drawing/System.Drawing/Color.cs
@@ -160,7 +160,7 @@ namespace System.Drawing
 			} else {
 				c = new Color ();
 				c.state = (short) (ColorType.ARGB | ColorType.Known | ColorType.Named);
-				if ((n < 27) || (n > 169))
+				if ((n < 27) || (n >= 168))
 					c.state |= (short) ColorType.System;
 				c.Value = KnownColors.ArgbValues [n];
 			}
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/ProfessionalColorTable.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/ProfessionalColorTable.cs
@@ -217,32 +217,32 @@ namespace System.Windows.Forms
 					menu_item_selected_gradient_begin = Color.FromArgb (182, 189, 210);
 					menu_item_selected_gradient_end = Color.FromArgb (182, 189, 210);
 
-					menu_strip_gradient_begin = SystemColors.ButtonFace;
+					menu_strip_gradient_begin = SystemColors.Control;
 					menu_strip_gradient_end = Color.FromArgb (246, 245, 244);
 
 					overflow_button_gradient_begin = Color.FromArgb (225, 222, 217);
 					overflow_button_gradient_end = SystemColors.ButtonShadow;
 					overflow_button_gradient_middle = Color.FromArgb (216, 213, 206);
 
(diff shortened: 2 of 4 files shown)


akoeplinger:
@monojenkins build

filipnavara:
I'd prefer to fix #2 to be honest. It's likely that applications will use the ButtonFace color internally and still end up with light colors.

The mixing of system and predefined colors for gradients should probably be fixed too, but I am not sure about the best approach. Either it could be some gradient between two system colors (Control and ControlLight/ControlDark), no gradient at all, or some lightened/darkened color calculated from a system color.

In any case this has to be tested with KeePass to ensure that it does not break anything there.

HebaruSan:
Fair enough, I'll start a new pull request for the other approach and make them mutually exclusive. Or would you prefer both in this PR, since the suggestion about using ControlLight/Dark for the gradient would make the most sense if the other side was Control?

KeePass looks identical before and after this change. I'm guessing they have their own custom renderer to draw their toolstrips rather than using the default one:

image

NattyNarwhal:
Yeah, KeePass was IIRC using the professional renderer but custom, to do some kind of highlight effect that looked like VS2008, but they just changed it to something that looks like rebar-based rendering/the system theme?

HebaruSan:
OK, pushed a new commit that 

1. Loads `ButtonFace` from the theme
2. Uses `SystemColors.ControlLight` as the other end of the gradients

Description and screenshots updated.

akoeplinger:
@monojenkins build

PR #11290, opened 2018-10-20
[aot] Bitcode fixes.

vargaz:


mono/mini/aot-compiler.c
```diff
@@ -4172,7 +4172,7 @@ add_wrappers (MonoAotCompile *acfg)
 
 			/* Create simplified signatures which match the signature used by the gsharedvt out wrappers */
 			for (variants = 0; variants < 4; ++variants) {
-				for (i = 0; i < 16; ++i) {
+				for (i = 0; i < 40; ++i) {
```
alexanderkyte: What is the source of this magic number? A comment or assignment to a named variable would be nice.

PR #11293, opened 2018-10-20
Make icall handle parameters pointers to locals.

jaykrell:
Per discussion, handles for icall parameters allocated in managed (but not native),
can just be pointers to locals or parameters.

Currently the managed stack is scanned conservatively.
When this changes, there will be metadata provided to find these.

Note that we can just use address of volatile parameters, not volatile locals.
For ObjOut/ObjInOut, these might point then to native frames containing pointers, but the pointers
to pointers in managed frame should suffice. If not, volatile locals are about as good and easily implemented.

This brings coop icall cost from 3+handle managed/native transitions down to 3.
  (Sometimes handle==0, i.e. for icalls that just want MonoError parameter.)
This can be further reduced, i.e. to one.

FIXME: Tests that exercise every icall, coop and otherwise.
FIXME: Gcstress.

Collateral damage: The one rarely run assert that handles are not out of scope cannot be easily implemented so is removed.

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6170,10 +6170,6 @@ typedef enum {
 
 typedef struct {
 	IcallHandlesWrap wrap;
-	/* if wrap is NONE or OBJ or VALUETYPE_REF, this is not meaningful.
-	   if wrap is OBJ_INOUT it's the local var that holds the MonoObjectHandle.
-	*/
-	int handle;
 }  IcallHandlesLocal;
 
 /*
@@ -6242,7 +6238,6 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 	int thread_info_var = -1, stack_mark_var = -1, error_var = -1;
 	MonoMethodSignature *call_sig = csig;
 	gboolean uses_handles = FALSE;
-	gboolean save_handles_to_locals = FALSE;
 	IcallHandlesLocal *handles_locals = NULL;
 	MonoMethodSignature *sig = mono_method_signature_internal (method);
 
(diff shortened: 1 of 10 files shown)


jaykrell:
CI all green in one attempt but let's get a full rebuild to exercise big change.

jaykrell:
@monojenkins build

jaykrell:
csc still crashes, because system mono.

jaykrell:

mono/metadata/marshal-ilgen.c
```diff
@@ -6274,17 +6271,15 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
 				case ICALL_HANDLES_WRAP_OBJ_INOUT:
 				case ICALL_HANDLES_WRAP_OBJ_OUT:
-					ret->params [i] = mono_class_get_byref_type (mono_class_from_mono_type_internal (csig->params[i]));
-					if (w == ICALL_HANDLES_WRAP_OBJ_OUT || w == ICALL_HANDLES_WRAP_OBJ_INOUT)
-						save_handles_to_locals = TRUE;
-					break;
-				case ICALL_HANDLES_WRAP_NONE:
 				case ICALL_HANDLES_WRAP_VALUETYPE_REF:
 					ret->params [i] = csig->params [i];
 					break;
```
jaykrell: These lines maybe wrong, correct before? What difference it makes?

jaykrell:
2 totalky clean CI runs except for system mono csc crashing once.

jaykrell:
Hopefully superceded by https://github.com/mono/mono/pull/11294.

jaykrell:
Redundant with master now.

PR #11294, opened 2018-10-20
Stack-based icall parameter handles and only one managed/native transition per icall.

jaykrell:
This builds on the stack-based icall parameter handles in https://github.com/mono/mono/pull/11293, and also drives the managed/native transitions to the ideal one.

Managed doesn't need the handle frame -- native can do it.
And native can also do the MonoError handling.

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -80,7 +80,7 @@ struct g_cast
 private:
 	void * const x;
 public:
-	explicit g_cast (void *y) : x(y) { }
+	explicit g_cast (void volatile *y) : x((void*)y) { }
 	// Lack of rvalue constructor inhibits ternary operator.
 	// Either don't use ternary, or cast each side.
 	// sa = (salen <= 128) ? g_alloca (salen) : g_malloc (salen);
--- a/mono/metadata/icall-table.h
+++ b/mono/metadata/icall-table.h
@@ -229,15 +229,15 @@ typedef MonoReflectionModuleHandle MonoReflectionModuleOutHandle;
 #define MONO_HANDLE_DO2(macro_prefix, type) MONO_HANDLE_DO3 (macro_prefix, type)
 #define MONO_HANDLE_DO(macro_prefix, type)  MONO_HANDLE_DO2 (macro_prefix, MONO_HANDLE_TYPE_WRAP_ ## type)
 
-#define MONO_HANDLE_RETURN_BEGIN(type)				MONO_HANDLE_DO (MONO_HANDLE_RETURN_BEGIN_, type)
-#define MONO_HANDLE_RETURN_BEGIN_Void				/* nothing */
-#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_NONE   	return
-#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_OBJ		return
+#define MONO_HANDLE_RETURN_BEGIN(type)				MONO_HANDLE_DO (MONO_HANDLE_RETURN_BEGIN_, type) (type)
+#define MONO_HANDLE_RETURN_BEGIN_Void(type)			/* nothing */
+#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_NONE(type)  type icall_result =
+#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_OBJ(type)   type ## Handle icall_result =
 
 #define MONO_HANDLE_RETURN_END(type)				MONO_HANDLE_DO (MONO_HANDLE_RETURN_END_, type);
-#define MONO_HANDLE_RETURN_END_Void				/* nothing */
-#define MONO_HANDLE_RETURN_END_ICALL_HANDLES_WRAP_NONE   	/* nothing */
-#define MONO_HANDLE_RETURN_END_ICALL_HANDLES_WRAP_OBJ		.__raw
+#define MONO_HANDLE_RETURN_END_Void				HANDLE_FUNCTION_RETURN ()
+#define MONO_HANDLE_RETURN_END_ICALL_HANDLES_WRAP_NONE   	HANDLE_FUNCTION_RETURN_VAL (icall_result)
+#define MONO_HANDLE_RETURN_END_ICALL_HANDLES_WRAP_OBJ		HANDLE_FUNCTION_RETURN_OBJ (icall_result)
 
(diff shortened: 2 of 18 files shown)


jaykrell:
Very small number of failures. Relevant?

jaykrell:

jaykrell:

jaykrell:
@monojenkins build Linux Aarch64

jaykrell:
@monojenkins build Linux Aarch64 Coop Suspend

jaykrell:
Let's get a full rebuild to exercise big change.

jaykrell:
@monojenkins build

mono/metadata/marshal-ilgen.c
```diff
@@ -6255,74 +6234,45 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
-					if (w == ICALL_HANDLES_WRAP_OBJ_OUT || w == ICALL_HANDLES_WRAP_OBJ_INOUT)
-						save_handles_to_locals = TRUE;
-					break;
 				case ICALL_HANDLES_WRAP_NONE:
+				case ICALL_HANDLES_WRAP_OBJ_OUT:
+				case ICALL_HANDLES_WRAP_OBJ_INOUT:
 				case ICALL_HANDLES_WRAP_VALUETYPE_REF:
 					ret->params [i] = csig->params [i];
 					break;
+				case ICALL_HANDLES_WRAP_OBJ:
```
jaykrell: These lines maybe wrong, correct before? What difference it makes?

jaykrell:
The consistent failure was due to the test assuming precise GC. Workaround is to zero just the right stack location.

jaykrell:
What reproed and fixed locally did not suffice for CI, so hunt around a bit more broadly.
Locally I searched the obvious candidates to narrow it down. CI will take more time for a narrower solution. Fix the test?

mono/utils/mono-error.h
```diff
@@ -57,7 +57,7 @@ typedef union _MonoError {
 	struct {
 		uint16_t error_code;
 		uint16_t private_flags; /*DON'T TOUCH */
-		void *hidden_1 [12]; /*DON'T TOUCH */
+		void *hidden_1 [9]; /*DON'T TOUCH */
```
kumpera: This is a breaking change, since the struct is public, its size must remain constant.

jaykrell: Oh because it in public embedding API? Agreed, will restore.

jaykrell:

jaykrell:
@lambdageek @kumpera review?

mono/eglib/glib.h
```diff
@@ -80,7 +80,7 @@ struct g_cast
 private:
 	void * const x;
 public:
-	explicit g_cast (void *y) : x(y) { }
+	explicit g_cast (void volatile *y) : x((void*)y) { }
```
jaykrell: Might want to move the cast elsewhere and leave this alone.

mono/metadata/icall-table.h
```diff
@@ -229,15 +229,15 @@ typedef MonoReflectionModuleHandle MonoReflectionModuleOutHandle;
 #define MONO_HANDLE_DO2(macro_prefix, type) MONO_HANDLE_DO3 (macro_prefix, type)
 #define MONO_HANDLE_DO(macro_prefix, type)  MONO_HANDLE_DO2 (macro_prefix, MONO_HANDLE_TYPE_WRAP_ ## type)
 
-#define MONO_HANDLE_RETURN_BEGIN(type)				MONO_HANDLE_DO (MONO_HANDLE_RETURN_BEGIN_, type)
-#define MONO_HANDLE_RETURN_BEGIN_Void				/* nothing */
-#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_NONE   	return
-#define MONO_HANDLE_RETURN_BEGIN_ICALL_HANDLES_WRAP_OBJ		return
+#define MONO_HANDLE_RETURN_BEGIN(type)				MONO_HANDLE_DO (MONO_HANDLE_RETURN_BEGIN_, type) (type)
```
jaykrell: Maybe not a great name for it? And the next one.

jaykrell:
@monojenkins build Linux x64

mono/metadata/marshal-ilgen.c
```diff
@@ -6340,148 +6336,103 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 			mono_error_assert_ok (error);
 		}
 
-		/* Add a MonoError argument and figure out which args need to be wrapped in handles */
+		// Add a MonoError argument (due to a fragile test external/coreclr/tests/src/CoreMangLib/cti/system/weakreference/weakreferenceisaliveb.exe),
```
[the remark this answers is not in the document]
jaykrell: Call f1, ret, call f2. Gc during f2 sees f1’s locals. Conservative gc.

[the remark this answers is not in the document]
lambdageek: I see. because there's still junk on the stack that `f2` didn't overwrite yet.

jaykrell:
I not sure efficiency & public go together. There is so much angst about coop icall perf, so we should be willing to have a more efficient internal form to get over that.

That said, the fast path of error check can be in IL.

That said, for interpreter, the less IL, the better.

jaykrell:
The point of volatile, is, not in registers. But no barriers. Taking address is almost as strong. Except whole program compilation could see through it. Also for valuetype_ref, the address is dead & could easily be removed by an optimizer.

jaykrell:
“Second pair of eyes” @kumpera ?

jaykrell:
Could be ‘icall_begin’ and ‘icall_end’ against opaque context.

jaykrell:

mono/metadata/marshal-ilgen.c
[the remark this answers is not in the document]
jaykrell: "yet" is an understatement -- not every path through a function writes to all of its locals.

jaykrell:
The “handle frame is setup in native”, yes.

jaykrell:
Is that ok? Compensate by having native use a different stack pointer for it?

jaykrell:
I think it is ok since other code does this -- older conversions and JIT icalls.

jaykrell:
Most of this PR has stood the test of its brief time, but the use of MonoMethodHeader has proven slightly wrong, has a race condition, when same method compiled multiple times concurrently.

PR #11299, opened 2018-10-20
Interior pointer handles are not passed in the way this code imagines

jaykrell:
and so the inefficient assert is next to impossible to hit.

Interior pointers are handled by retaining a handle in the managed wrapper,
and passing a raw pointer to native.

Native never uses the interior pointer handles, other than to create
them on behalf of marshal-ilgen and would have to
go extremely far out of its way to get to them (reaching into the handle stack).

I suppose this is also checking for use of out of scope handles.
But we don't do that in HANDLE_RAW for example.

### DIFF
--- a/mono/metadata/handle.c
+++ b/mono/metadata/handle.c
@@ -104,37 +104,6 @@ chunk_element (HandleChunk *chunk, int idx)
 	return &chunk->elems[idx];
 }
 
-static HandleChunkElem*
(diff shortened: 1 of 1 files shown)


jaykrell:
I guess this is more about out of scope handles than interior handles.
It is odd that this is the only place we check for this, and there isn't a helper ever sprinkled around like `mono_assert_valid_handle().`

PR #11301, opened 2018-10-20
Draw menu background with SystemColors.Menu to fix dark themes

HebaruSan:
## Problem

Dropdown menus have light text on a light background in a dark theme:

image

## Cause

The background is rendered here:

https://github.com/mono/mono/blob/4a6b11577fe06a58e6326b1c3a3bb58cfee6185c/mcs/class/System.Windows.Forms/System.Windows.Forms/ToolStripProfessionalRenderer.cs#L372-L375

`this.ColorTable.ToolStripDropDownBackground` is set to either `SystemColors.Window` or various hard coded shades of near-white in `ProfessionalColorTable`.

## Changes

Now that color comes from `SystemColors.Menu`, defined as "[Gets a Color structure that is the color of a menu's background](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.systemcolors)", which looks better in a dark theme:

image

It still looks fine in a normal theme as well:

image

KeePass is not affected by this change, presumably thanks to its custom toolstrip renderer.


akoeplinger:
@monojenkins build

akoeplinger:
Sorry for the delay, thanks for your contribution!

PR #11305, opened 2018-10-22
[ci] Enable ccache on linux builds.

vargaz:


vargaz:
@monojenkins build

vargaz:
@monojenkins build

lewurm:
is that useful? given that CI jobs run inside a chroot

scripts/ci/run-jenkins.sh
```diff
@@ -94,6 +94,15 @@ then
 	wget -qO- https://download.mono-project.com/test/new-certs.tgz| tar zx -C ~/.config/.mono/
 fi
 
+
+if [[ ${CI_TAGS} == *'linux-'* ]];
```
akoeplinger: I'd prefer if we made this an explicit `CI_TAGS == *'ccache'*` so we don't turn it on everywhere (i.e. it will probably cause issues on jobs that already override the compiler like the static analysis ones).

akoeplinger: Another thing we need to solve is where to put the ccache cache, cause as far as I know it gets put into $HOME by default which would be cleaned up after the build when the chroot is destroyed.

@directhex can help us figure out a location that is preserved.

vargaz:
Its just an experiment, we could put the cache outside the chroot.

scripts/ci/run-jenkins.sh
directhex: We can add a bind-mounted directory to the job config. Android uses this to share android-toolchain between different jobs.

vargaz:
ccache has a CCACHE_DIR env variable which can be set to a dir outside the chroot or a binddir.

vargaz:
@monojenkins build failed

scripts/ci/run-jenkins.sh
```diff
@@ -109,6 +109,15 @@ then
 	wget -qO- https://download.mono-project.com/test/new-certs.tgz| tar zx -C ~/.config/.mono/
 fi
 
+if [[ ${CI_TAGS} == *'ccache'* ]];
+then
+	# CCACHE_DIR should be set to a directory outside the chroot which holds the cache
+	CCACHE=`which ccache`
+	if [ "$CCACHE" ]; then
+		export EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} CC='ccache gcc' CXX='ccache g++'"
```
luhenry: Export the `CC` and `CXX` variables instead, similar to https://github.com/mono/mono/pull/11305/files#diff-4847efbf06c7d2913d13c1a50c2ae0b8L22

vargaz: Done.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

Issue #11302, opened 2018-10-21
`ParameterInfo.Get{Required,Optional}CustomModifiers()` returns cmods in different order than the CLR


stakx:
There have been some improvements in Mono 5.16 regarding custom modifiers. One of them is that `ParameterInfo.GetRequiredCustomModifiers()` and `ParameterInfo.GetOptionalCustomModifiers()` now report cmods when previously they didn't see them at all.

There's still one notable difference to the CLR's behavior: Mono returns the cmods in the same order as they appear in the signatures, while the CLR returns them in reversed order.

## Steps to Reproduce

Compile and run the following console program:

```csharp
using System;
using System.Reflection;
using System.Reflection.Emit;

public sealed class A { }
public sealed class B { }

class Program
{
    static void Main()
    {
        var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName("A"), AssemblyBuilderAccess.RunAndSave);
        var moduleBuilder = assemblyBuilder.DefineDynamicModule("A", "A.dll");
        var typeBuilder = moduleBuilder.DefineType("C", TypeAttributes.Class);
        var methodBuilder = typeBuilder.DefineMethod(
            "M",
            MethodAttributes.Public,
            CallingConventions.HasThis,
            typeof(void),
            null,
            null,
            new[] { typeof(int) },
            new[] { new[] { typeof(A), typeof(B), typeof(A), typeof(A) } },
            null);
        methodBuilder.GetILGenerator().Emit(OpCodes.Ret);

        var type = typeBuilder.CreateType();
        var method = type.GetMethod("M");
        var parameter = method.GetParameters()[0];
        foreach (var modreq in parameter.GetRequiredCustomModifiers())
        {
            Console.WriteLine(modreq);
        }

        assemblyBuilder.Save("A.dll");
    }
}
```

Observe the console output.

### Current Behavior

The program prints the following four lines:

```
A
B
A
A
```

That is, `ParameterInfo.GetRequiredCustomModifiers()` returns the cmods exactly like they were emitted.

### Expected Behavior

Assuming that Mono should behave exactly like the CLR, it should print:

```
A
A
B
A
```

`ParameterInfo.GetRequiredCustomModifiers()` on the CLR returns the modreqs in reversed order. (Same is true for modopts btw.) 

## On which platforms did you notice this

[ ] macOS
[X] Linux (Debian GNU/Linux 9 (stretch))
[ ] Windows

**Version Used**:

```
[...]

PR #11308, opened 2018-10-22
[Reflection] Fix ParameterInfo's custom modifiers order

MaximLipnin:
Mono returns custom modifiers in the same order as they appear in the signatures. NetFX/CoreFX returns them in reversed order. So we just reverse the order to follow NetFX/CoreFX behavior.

Fixes https://github.com/mono/mono/issues/11302

### DIFF
--- a/mcs/class/corlib/System.Reflection/MonoParameterInfo.cs
+++ b/mcs/class/corlib/System.Reflection/MonoParameterInfo.cs
@@ -315,15 +315,6 @@ namespace System.Reflection
 			return new MonoParameterInfo (type, member, marshalAs);
 		}
 
-		private Type[] GetCustomModifiers (bool optional) {
-			Type[] types = GetTypeModifiers (optional);
-			if (types == null)
-				return Type.EmptyTypes;
-			// Mono returns modifiers in the same order as they appear in the signatures.
-			// NetFX/CoreFX returns them in reversed order.
-			// So we just reverse the order to follow NetFX/CoreFX behavior.
-			Array.Reverse (types);
-			return types;			
-		}
+		private Type[] GetCustomModifiers (bool optional) => GetTypeModifiers (optional) ?? Type.EmptyTypes;
 	}
 }
--- a/mcs/class/corlib/System.Reflection/MonoParameterInfo.cs
+++ b/mcs/class/corlib/System.Reflection/MonoParameterInfo.cs
@@ -214,14 +214,7 @@ namespace System.Reflection
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		internal extern int GetMetadataToken ();
 
-		public
-		override
-		Type[] GetOptionalCustomModifiers () {
-			Type[] types = GetTypeModifiers (true);
-			if (types == null)
-				return Type.EmptyTypes;
-			return types;
-		}
+		public override Type[] GetOptionalCustomModifiers () => GetCustomModifiers (true);
 
 		internal object[] GetPseudoCustomAttributes () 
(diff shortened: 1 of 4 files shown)


mcs/class/corlib/System.Reflection/MonoParameterInfo.cs
```diff
@@ -327,6 +313,17 @@ internal static ParameterInfo New (ParameterInfo pinfo, MemberInfo member)
...
+		}
+
+		private Type[] GetCustomModifiers (bool optional) {
+			Type[] types = GetTypeModifiers (optional);
+			if (types == null)
+				return Type.EmptyTypes;
+			// Mono returns modifiers in the same order as they appear in the signatures.
+			// NetFX/CoreFX returns them in reversed order.
+			// So we just reverse the order to follow NetFX/CoreFX behavior.
+			Array.Reverse (types);
```
marek-safar: We built the array in the loop so just do it there instead. Please also add tests for Field and Property custom modifiers

MaximLipnin: Seems that custom modifiers functionality does't work properly for Field and Property. 
The following code has empty output on Mono (NetFX returns both required and optional modifiers in expected order)
```
using System;
using System.Reflection;
using System.Reflection.Emit;

public sealed class A { }
public sealed class B { }
public sealed class C { }
public sealed class D { }
class Program
{
    static void Main()
    {
        var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName("A"), AssemblyBuilderAccess.RunAndSave);
        var moduleBuilder = assemblyBuilder.DefineDynamicModule("A", "A.dll");
        var typeBuilder = moduleBuilder.DefineType("C", TypeAttributes.Class);
        var fieldBuilder = typeBuilder.DefineField(
            "F",
            typeof(string),
            new[] { typeof(A), typeof(B), typeof(A), typeof(A) },
            new[] { typeof(C), typeof(D), typeof(C), typeof(C) },
            FieldAttributes.Public);

        var type = typeBuilder.CreateType();
        var field = type.GetField("F");
        foreach (var modreq in field.GetRequiredCustomModifiers())
        {
            Console.WriteLine(modreq);
        }

        foreach (var modopt in field.GetOptionalCustomModifiers())
        {
            Console.WriteLine(modopt);
        }
    }
}
```

MaximLipnin:
> to add a second test that ensures it is only ParameterInfo.Get...CustomModifiers() that reverses order

@stakx Do you have an appropriate small test case to add ?

MaximLipnin:
https://github.com/mono/mono/issues/11320 "[Reflection] Get{Required/Optional}CustomModifiers methods don't return modifiers for Field and Property"

stakx:
@MaximLipnin - Not right away. I'll need to think about how this could be done. One would have to get at the parameter's signature, ideally... possibly via `ParameterInfo.MetadataToken` and `Module.ResolveSignature` somehow?

MaximLipnin:
@stakx anyway, thank you for participating!

akoeplinger:
Should we backport this fix?

lambdageek:
> Consider: The test that's being added here would also pass if it were Reflection.Emit that performed the reversing... but that would be wrong: Cmod order in the emit call and in the produced signature should match.

@stakx good thinking.   That is actually happening.  I have a fix in #13320. (It's difficult to make a bulletproof test without external dependencies - refemit, metadata parsing, and the reflection API are involved - if any two of the three of them are wrong, we'll get a passing test. )

PR #11312, opened 2018-10-22
[tests] Regression test for #11123

lambdageek:
Loading a type that has a field from an assembly that cannot be loaded should
still leave the type usable for some operations such as IsSubclassOf.

Regression test for #11123


lambdageek:
@monojenkins build failed

PR #11321, opened 2018-10-23
[aot] add unbox_arbitrary_trampoline for fullaot/interp mixed mode

lewurm:
On a virtual call where the receiver is a value type, the value type is
boxed at the caller's site, but must be unboxed before passed to the
callee. In FullAOT this happens with the help of the `unbox_trampoline`,
which is attached to the AOT'd assembly for every virtual method with a
value type as receiver, so we have a specific `unbox_trampoline` for
each such `MonoMethod`.

In mixed mode it can happen that an FullAOT'd method is calling such a
method which is _not_ FullAOT compiled, therefore it has to be executed
by the interpreter. We have to insert a trampoline between AOT caller
and interp callee (being a native-to-interp entry), that does the
unboxing and has a configurable target address at run-time. So that:

  AOT caller -> unbox_arbitrary_trampoline -> native-to-interp entry

Note on testing: Tested on three different configurations
* amd64/macos with --full-aot-interp and `make -C mcs/class/corlib TESTNAME=MonoTests.System.ArrayTest.BinarySearch_ArgPassingOrder`. This will be covered soon by https://github.com/mono/mono/pull/11307.
* arm64/linux, same as above.
* arm64/iOS which has a different code path due to trampoline pages being used. Tested with `monotouch-tests` mixed mode (i.e. mscorlib.dll FullAOT, rest is interpreter).

armv7 isn't implemented yet.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -215,6 +215,7 @@ typedef struct MonoAotOptions {
 	int ngsharedvt_arg_trampolines;
 	int nftnptr_arg_trampolines;
 	int nrgctx_fetch_trampolines;
+	int nunbox_arbitrary_trampolines;
 	gboolean print_skipped_methods;
 	gboolean stats;
 	gboolean verbose;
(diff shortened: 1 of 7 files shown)


marek-safar:
@vargaz could you please review

mono/mini/aot-compiler.c
```diff
@@ -10477,12 +10547,14 @@ emit_aot_file_info (MonoAotCompile *acfg, MonoAotFileInfo *info)
 		symbols [sindex ++] = "imt_trampolines";
 		symbols [sindex ++] = "gsharedvt_arg_trampolines";
 		symbols [sindex ++] = "ftnptr_arg_trampolines";
+		symbols [sindex ++] = "unbox_arbitrary_trampolines";
 	} else {
```
vargaz: The llvm version needs to be changed too in emit_aot_file_info () in mini-llvm.c.

vargaz:
Looks ok otherwise.

mono/mini/aot-compiler.c
lewurm: thanks, updated.

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

mono/mini/aot-compiler.c
```diff
@@ -7618,6 +7744,8 @@ mono_aot_parse_options (const char *aot_options, MonoAotOptions *opts)
 			printf ("    nrgctx-trampolines=\n");
 			printf ("    nrgctx-fetch-trampolines=\n");
 			printf ("    ngsharedvt-trampolines=\n");
+			printf ("    nftnptr-arg-trampolines=\n");-
```
lewurm: fixed by https://github.com/mono/mono/pull/11360
thanks @jaykrell

lewurm:
FYI, it was merged because `Test collection for System.Reflection.Tests.GetTypeTests.System.Reflection.Tests.GetTypeTests.GetType` is an unrelated failure

lewurm:
@monojenkins backport 2018-06

lewurm:
@monojenkins backport 2018-08

lewurm:
@monojenkins backport 2018-10

Issue #11303, opened 2018-10-21
`ParameterInfo.[Raw]DefaultValue` returns `Missing.Value` for non-null default values of `DateTime?` and `decimal?` parameters


stakx:
This is (kind of) a continuation of #8597 (which has been fixed in Mono 5.16).

## Steps to Reproduce

Compile and run the following program:

```csharp
using System;
using System.Linq;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using System.Reflection;

class Program
{
    public void M1(decimal? arg = 12.345M) {}
    public void M2([Optional, DecimalConstant(1,2,3,4,5)] decimal? arg) {}
    public void M3(decimal? arg = null) {}
    public void M4([Optional, DateTimeConstant(1L)] DateTime? arg) {}
    public void M5(DateTime? arg = null) {}

    static void Main()
    {
        foreach (var method in typeof(Program).GetMethods().Where(m => m.Name.StartsWith("M")))
        {
            Console.WriteLine("{0}: {1}", method.Name, method.GetParameters()[0].RawDefaultValue);
        }
    }
}
```

(It doesn't appear to make a difference whether you use `.DefaultValue` or `.RawDefaultValue`.)

Observe the program's console output.

### Current Behavior

The program prints the following:

```
M1: System.Reflection.Missing
M2: System.Reflection.Missing
M3:
M4: System.Reflection.Missing
M5:
```

### Expected Behavior

The program should produce the same output as the CLR:

```
M1: 12.345
M2: -5534023223830852403.7
M3:
M4: 01/01/0001 00:00:00
M5:
```

## On which platforms did you notice this

[ ] macOS
[X] Linux (Debian GNU/Linux 9 (stretch))
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.16.0.179 (tarball Thu Oct  4 10:24:32 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          yes(3.6.0svn-mono-/)
        GC:            sgen (concurrent by default)
```

/cc @lambdageek

marek-safar:
@MaximLipnin could you check if we are still failing on this one?

marek-safar:
[...]

PR #11330, opened 2018-10-23
Fixing ParameterInfo.RawDefaultValue for non-null default values of DateTime?` and decimal? 

MaximLipnin:
Adding checks for `DateTime?` and `decimal?` to `DefaultValue` property getter will allow to retrieve correct values instead of  System.Reflection.Missing

Fixes https://github.com/mono/mono/issues/11303


MaximLipnin:
@monojenkins build failed

marek-safar:
@monojenkins backport 2018-10

PR #11331, opened 2018-10-23
[System.Net.Http] Comment out dead code.

rolfbjarne:
Comment out code to convert NSUrlError to WebExceptionStatus, because the
result isn't used. Looking through git history it seems the resulting
WebExceptionStatus was never used.

Fixes this warning:

    HttpClientEx.cs(83,8): warning CS0219: The variable 'webExceptionStatus' is assigned but its value is never used


marek-safar:
@rolfbjarne is it worth keeping that comments in?

rolfbjarne:
@marek-safar I'm not sure, it seemed like the code could be useful at some point if we decided to make the mapping work again, but I don't really care either way, whatever any of you say is fine with me.

PR #11338, opened 2018-10-24
[2018-08] Replace mono_object_get_class with mono_object_class to avoid gc transitions.

jaykrell:
Replacement for #11104.
Backport of #11077.


jaykrell:

PR #11340, opened 2018-10-24
[SDKS, LLVM] Disable detection and use of libxml2

grendello:
**Please backport to the following branches: `2018-10`, `2018-08`**

Mono doesn't use libxml2 and neither do the LLVM bits used by Mono when building
the AOT cross-compilers. Presence of `-lxml2` on the command line breaks 32-bit
builds on 64-bit Linux hosts:

    libtool: link: cc -shared  -fPIC -DPIC  -Wl,--whole-archive ./.libs/libmini.a ./.libs/libmono-ee-interp.a ./.libs/libmono-dbg.a ../../mono/metadata/.libs/libmonoruntimesgen.a ../../mono/sgen/.libs/libmonosgen.a ../../mono/utils/.libs/libmonoutils.a ../../mono/eglib/.libs/libeglib.a -Wl,--no-whole-archive  -lLLVMBitWriter -lLLVMAnalysis -lLLVMProfileData -lLLVMObject -lLLVMMCParser -lLLVMMC -lLLVMBitReader -lLLVMCore -lLLVMBinaryFormat -lLLVMSupport -lLLVMDemangle -L$HOME/xamarin-android/external/mono/sdks/out/llvm-llvm32/lib -lz -ltinfo -lxml2 -lstdc++ -lm -lrt -ldl -lpthread  -O0 -ggdb3 -m32 -g -mno-tls-direct-seg-refs -Wl,-export-dynamic   -Wl,-soname -Wl,libmonosgen-2.0.so.1 -o .libs/libmonosgen-2.0.so.1.0.0
    mv -f .deps/libmini_la-aot-runtime-wasm.Tpo .deps/libmini_la-aot-runtime-wasm.Plo
    /usr/bin/ld: skipping incompatible //usr/lib/x86_64-linux-gnu/libxml2.so when searching for -lxml2
    /usr/bin/ld: skipping incompatible //usr/lib/x86_64-linux-gnu/libxml2.a when searching for -lxml2
    /usr/bin/ld: cannot find -lxml2
    collect2: error: ld returned 1 exit status
    make[7]: *** [Makefile:1495: libmonosgen-2.0.la] Error 1

Installation of `libxml2-dev` (on Ubuntu/Debian + friends) for the x86
architecture (or x86_64 architecture on 32-bit host) would be possible but then
it would break the 64-bit build because the 64-bit version of the library would
be absent. It would be a problem if Mono+LLVM actually used xml2, but since it
doesn't we can just disable the feature completely.


sdks/builds/llvm.mk
```diff
@@ -48,8 +48,10 @@ endif
 #  $(1): target
 define LLVMTemplate
 
+# -DLLVM_ENABLE_LIBXML2=Off is needed because xml2 is not used and it breaks 32-bit builds on 64-bit Linux hosts
 _llvm-$(1)_CMAKE_ARGS = \
-	$$(llvm-$(1)_CMAKE_ARGS)
+	$$(llvm-$(1)_CMAKE_ARGS) \
+	-DLLVM_ENABLE_LIBXML2=Off \
```
luhenry: Move that to a new line in between the following lines: https://github.com/mono/mono/blob/05bc22015c9cfedfb6faead1fe3aac7e918e498a/llvm/build.mk#L32-L33

akoeplinger:
Cherry-picked manually to 2018-10 and -08

PR #11342, opened 2018-10-24
[corlib] Import FileSystemInfo and family from CoreFX

alexischr:
Includes the following from System.IO:

* FileSystemInfo
* FileInfo
* DirectoryInfo
* Directory
* FileSystem
* Path `Join()` Span API

and all of System.IO.Enumeration namespace.

Adds about 1600 System.IO tests

Behavior changes:
* `File.Replace` now replaces files even if they are read-only (using the `rename` syscall on Unix)
* Directory enumeration is no longer stable in the way that many Mono tests expect
* Some changes in when/whether some representations of directories contain a trailing separator character
* Changes in which exceptions are thrown (sometimes these are platform-dependent on CoreFX)

Issues:
* https://github.com/dotnet/corefx/issues/33486
* https://github.com/dotnet/corefx/issues/33490

Part of #7246


alexischr:
@monojenkins build

alexischr:
@monojenkins build

mcs/class/corlib/System.IO.Enumeration/FileSystemEnumerator.NotSupported.cs
```diff
@@ -0,0 +1,23 @@
+using System.Collections;
+
+namespace System.IO.Enumeration
+{
+	partial class FileSystemEnumerator<TResult>
```
marek-safar: This does not match the API of the reference class

```
    public abstract class FileSystemEnumerator<TResult> : Runtime.ConstrainedExecution.CriticalFinalizerObject, Collections.Generic.IEnumerator<TResult>
    {
        public FileSystemEnumerator(string directory, EnumerationOptions options = null) { }

        protected virtual bool ShouldIncludeEntry(ref FileSystemEntry entry) { throw null; }
        protected virtual bool ShouldRecurseIntoEntry(ref FileSystemEntry entry) { throw null; }
        protected abstract TResult TransformEntry(ref FileSystemEntry entry);
        protected virtual void OnDirectoryFinished(ReadOnlySpan<char> directory) { throw null; }
        protected virtual bool ContinueOnError(int error) { throw null; }

        public TResult Current { get { throw null; } }
        object System.Collections.IEnumerator.Current { get { throw null; } }
        public bool MoveNext() { throw null; }
        public void Reset() { throw null; }
        public void Dispose() { throw null; }
        protected virtual void Dispose(bool disposing) { throw null; }
    }
```

mcs/class/corlib/System.IO/FileSystem.NotSupported.cs
```diff
@@ -0,0 +1,32 @@
+namespace System.IO
+{
+	internal static partial class FileSystem
```
marek-safar: Why is it internal?

mcs/class/corlib/System.IO/Path.cs
```diff
@@ -971,6 +972,183 @@ public static ReadOnlySpan<char> GetFileName(ReadOnlySpan<char> path)
             return path;
         }
 
+		public static string Join(ReadOnlySpan<char> path1, ReadOnlySpan<char> path2)
```
marek-safar: Add a marker where this was copied from

mcs/class/corlib/System.IO/FileSystem.NotSupported.cs
alexischr: This is an internal helper class, see https://github.com/dotnet/corefx/blob/e0cb81d4df9e6279cff5be3659750f1140d977e0/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs for example

mcs/class/corlib/System.IO.Enumeration/FileSystemEnumerator.NotSupported.cs
alexischr: The API is provided by https://github.com/mono/corefx/blob/806496e9a4608b2a5a91c7fa0d3fe79cde616d44/src/System.IO.FileSystem/src/System/IO/Enumeration/FileSystemEnumerator.cs which is included in all corlib sources.

mcs/class/corlib/System.IO/Path.cs
alexischr: This is already in a `#region Copied from corefx` (Added region to File.cs change too)

mcs/class/corlib/corefx/MonoLinqHelper.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+
+namespace System.IO
+{
+	internal static class MonoLinqHelper {
```
marek-safar: You don't need this we already have `EnumerableHelpers`

mcs/class/corlib/Test/System.IO/DirectoryTest.cs
```diff
@@ -1113,7 +1109,7 @@ public void SetLastWriteTimeException2 ()
 	}
 	
 	[Test]
-	[ExpectedException(typeof(FileNotFoundException))]
+	[ExpectedException()]
```
marek-safar: Why was the type removed?

mcs/class/corlib/System.IO/FileSystem.NotSupported.cs
```diff
@@ -0,0 +1,32 @@
...
+	{
+		public static void CopyFile(string sourceFullPath, string destFullPath, bool overwrite) => throw new PlatformNotSupportedException ();
+		public static void ReplaceFile(string sourceFullPath, string destFullPath, string destBackupFullPath, bool ignoreMetadataErrors) => throw new PlatformNotSupportedException ();
+		public static void MoveFile(string sourceFullPath, string destFullPath) => throw new PlatformNotSupportedException ();
+		public static void DeleteFile(string fullPath) => throw new PlatformNotSupportedException ();
+		public static void CreateDirectory(string fullPath) => throw new PlatformNotSupportedException ();
+		public static void MoveDirectory(string sourceFullPath, string destFullPath) => throw new PlatformNotSupportedException ();
+		public static void RemoveDirectory(string fullPath, bool recursive) => throw new PlatformNotSupportedException ();
+		public static bool DirectoryExists(string fullPath) {
+			MonoIOError error;
```
marek-safar: Why is this method special?

mcs/class/corlib/System.IO.Enumeration/FileSystemEnumerator.NotSupported.cs
marek-safar: I don't understand this is NSE version of the API but with slightly different shape

mcs/class/corlib/corefx/MonoLinqHelper.cs
```diff
@@ -3,30 +3,11 @@
 
 namespace System.IO
 {
-	internal static class MonoLinqHelper {
+	internal static class MonoLinqHelper
```
marek-safar: Why do you need this wrapper?

alexischr: Several files expect a ToArray extension to `IEnumerable<T>` (e.g. at https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Directory.cs#L146) and that comes from LINQ (https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.toarray?view=netframework-4.7.2)

alexischr:
@marek-safar I used the "Resolve Conversation" option wrong and some of my comments did not post. Regarding `FileSystemEnumerator.NotSupported.cs`: it is used only to allow 
https://github.com/mono/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Enumeration/FileSystemEnumerator.cs to compile, which is the platform-independent CoreFX partial file and also implements the full API. The corefx file is included everywhere (at `corlib.dll.sources` level)

mcs/class/corlib/Test/System.IO/DirectoryTest.cs
alexischr: These now throw `DirectoryNotFoundException` on Unix, and `FileNotFoundException` on Windows

mcs/class/Facades/System.IO.FileSystem.AccessControl/System.IO.FileSystem.AccessControl.dll.sources
```diff
@@ -1,4 +1,3 @@
 TypeForwarders.cs
 AssemblyInfo.cs
```
marek-safar: We need to replace it with type forwareded

alexischr:
@monojenkins build failed

alexischr:

marek-safar:
@alexischr I think the mcs failures are related to this change

alexischr:
@monojenkins build failed

alexischr:

alexischr:

mcs/class/corlib/System.IO/FileSystem.NotSupported.cs
alexischr: It was used by gensources while on the basic profile, so the method needed to work. Change removed as basic profile is no more

alexischr:
@monojenkins commit csproj

alexischr:

alexischr:

alexischr:
@marek-safar can you approve/merge?

marek-safar:
@monojenkins commit apidiff

PR #11343, opened 2018-10-24
Add option to integrate msvc build into regular make build.

lateralusX:
Add support to perform msvc mono runtime build as part of regular make build. By default, msvc build will not automatically but can be enabled using --enable-msvc argument passed to autogen.sh. Doing msvc build as part of regular make will automatically set MONO_EXECUTABLE to build msvc build Mono runtime used in BCL and Test build as well as runtime used when running tests. Build msvc binaries will also be part of make install, if enabled.

Running build using --enable-msvc removes the need to do separate msvc build (normally before running make) and setup MONO_EXECUTABLE to matching build binary later picked up by BCL/Test builds and test execution. Using --enable-msvc automatically makes sure a matching msvc runtime gets build and used based on used --host parameter. Doing a full Windows build, including msvc build can now be done like this:

./autogen.sh --prefix=$PREFIX --host=x86_64-w64-mingw32 --enable-msvc --disable-boehm
make -j4


configure.ac
```diff
@@ -5319,7 +5319,37 @@ AC_DEFINE(MONO_INSIDE_RUNTIME,1, [Disable banned functions from being used by th
 mono_build_root=`pwd`
 AC_SUBST(mono_build_root)
 
-mono_runtime=mono/mini/mono
+AC_ARG_WITH(msvc, [  --with-msvc=yes,no      Build with Visual Studio (msvc) (defaults to no)],[],[with_msvc=no])
```
luhenry: Use `AC_ARG_ENABLE` so you have `--enable-msvc`. We overuse the `--with-*` which is not originally meant for a yes/no parameter. Then propagate that `ENABLE_MSVC` variable instead of `WITH_MSVC`.

lateralusX: OK, will change. Original used --enable-msvc at a start but then I looked for other patterns that seemed to be using --with-*.

lateralusX:
@monojenkins build failed

Issue #11346, opened 2018-10-24
[wasm] sample docs update


kjpou1:
The wasm sample docs are out of date after recent changes.

Issue #11344, opened 2018-10-24
[wasm] Regression with packager.exe and distribution build


kjpou1:
- Can not resolve bcl nor frameworks assemblies from mono build .zip distribution

   ```
   Unhandled Exception:
   System.Exception: Could not resolve mscorlib
     at Driver.Resolve (System.String asm_name, Driver+AssemblyKind& kind) [0x00064] in 
   <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Import (System.String ra, Driver+AssemblyKind kind) [0x000df] in 
   <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Run (System.String[] args) [0x00409] in <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Main (System.String[] args) [0x00006] in <af613a8cb37f45edb522c09b096debb2>:0
   [ERROR] FATAL UNHANDLED EXCEPTION: System.Exception: Could not resolve mscorlib
     at Driver.Resolve (System.String asm_name, Driver+AssemblyKind& kind) [0x00064] in 
   <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Import (System.String ra, Driver+AssemblyKind kind) [0x000df] in 
   <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Run (System.String[] args) [0x00409] in <af613a8cb37f45edb522c09b096debb2>:0
     at Driver.Main (System.String[] args) [0x00006] in <af613a8cb37f45edb522c09b096debb2>:0
   ```

- The regression comes from the following lines of `packager.exe`:

   ``` csharp
        //are we working from the tree?
        if (sdkdir != null) {
            framework_prefix = tool_prefix; //all framework assemblies are currently side built to packager.exe
            bcl_prefix = Path.Combine (sdkdir, "wasm-bcl/wasm");
        } else if (Directory.Exists (Path.Combine (tool_prefix, "../out/wasm-bcl/wasm"))) {
            framework_prefix = tool_prefix; //all framework assemblies are currently side built to packager.exe
            bcl_prefix = Path.Combine (tool_prefix, "../out/wasm-bcl/wasm");
            sdkdir = Path.Combine (tool_prefix, "../out");
        } else {
            framework_prefix = Path.Combine (tool_prefix, "framework");
            bcl_prefix = Path.Combine (tool_prefix, "wasm-bcl/wasm");
            sdkdir = tool_prefix;
        }
[...]

PR #11347, opened 2018-10-24
[wasm] Wasm sample docs

kjpou1:
Create docs directory and update sample documentation.

- Add getting started docs directory
- Add section of obtaining the mono distribution.

resolves: https://github.com/mono/mono/issues/11346

Fix distribution package from wasm build to have the same structure as `packager.exe` is expecting.

resolves: https://github.com/mono/mono/issues/11344


kjpou1:
@monojenkins build failed

PR #11348, opened 2018-10-24
[runtime] Fixes threadpool tests not to rely on arbitrary sleep timeout

marek-safar:

### DIFF
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -436,10 +436,8 @@ TESTS_CS_SRC=		\
 	threadpool.cs		\
 	threadpool1.cs		\
 	threadpool-exceptions1.cs \
-	threadpool-exceptions2.cs \
 	threadpool-exceptions3.cs \
 	threadpool-exceptions4.cs \
-	threadpool-exceptions5.cs \
 	threadpool-exceptions6.cs \
 	base-definition.cs	\
 	bug-27420.cs		\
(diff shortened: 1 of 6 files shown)


akoeplinger:
@monojenkins build failed

mono/tests/Makefile.am
```diff
@@ -2866,7 +2864,9 @@ TESTS_UNHANDLED_EXCEPTION_255_SRC =	\
 	unhandled-exception-5.cs	\
 	unhandled-exception-6.cs	\
 	unhandled-exception-7.cs	\
-	unhandled-exception-8.cs
+	unhandled-exception-8.cs	\
+	threadpool-exceptions2.cs	\
```
luhenry: This means the `threadpool-exceptions2.cs` and `threadpool-exceptions5.cs` tests are expected to have exit code = 255? That sounds wrong.

mono/tests/threadpool-exceptions2.cs
```diff
@@ -2,38 +2,46 @@
...
+			return;
+		}
+
+		if (!e.IsTerminating) {
+			Environment.Exit (4);
+			return;
 		}
-		Environment.Exit (return_value);
+
+		mre.Set ();
```
luhenry: Why not having the same `Environment.Exit (0)` as the other tests?

mono/tests/threadpool-exceptions5.cs
```diff
@@ -2,37 +2,47 @@
...
-		lock (monitor) {
-			Monitor.Wait (monitor);
-		}
-		Thread.Sleep (1000);
-		return 1;
+
+		if (!mre.WaitOne (10000))
+			return 2;
+
+		GC.Collect ();
```
luhenry: What's the need for these 2 new function calls?

mono/tests/threadpool-exceptions2.cs
```diff
@@ -2,38 +2,46 @@
...
-		lock (monitor) {
-			Monitor.Wait (monitor);
-		}
-		Thread.Sleep (1000);
-		return 1;
+
+		if (!mre.WaitOne (10000))
+			return 2;
+
+		GC.Collect ();
```
luhenry: What's the need for these 2 new function calls?

mono/tests/Makefile.am
marek-safar: Yes, that's expected. They are testing the unhandled process exit caused by user exception

mono/tests/threadpool-exceptions2.cs
marek-safar: They are not strictly necessary, could remove them but don't hurt either but it make the repro more reliable

mono/tests/threadpool-exceptions2.cs
marek-safar: Because this is exit path for unhandled exception

mono/tests/threadpool-exceptions5.cs
marek-safar: same as above, could remove then but they help to ensure there is no other side effect

PR #11352, opened 2018-10-24
[build] Add more tools for monodroid to reduce sources dependency

marek-safar:


akoeplinger:
@monojenkins commit csproj

PR #11358, opened 2018-10-24
Remove excess in icall-decl.h.

jaykrell:
The extra ves_icall_get_method_info breaks WebAssembly/C++ for understood reasons.


jaykrell:
More complete fix is https://github.com/mono/mono/pull/11359.

PR #11359, opened 2018-10-24
[cxx] Fix WebAssembly/C++ (ICALL_EXPORT)

jaykrell:


jaykrell:

jaykrell:

jaykrell:

lambdageek:
@jaykrell could you explain why `ICALL_EXPORT` on definitions for icalls that are marked with `HANDLES()` is incorrect? What goes wrong, exactly?

PR looks ok - I just want to understand what to avoid in the future.

jaykrell:
Subject to configuration, ICALL_EXPORT is extern "C and exported.
For HANDLES() icalls, that is the wrappers, only.
The  typed-handled functions are extern C++  and not exported.
We could apply  it to the other functions, but we'd have to do so consistently.

For some configurations, ICALL_EXPORT is nothing, in which case it can be applied arbitrarily.

I suppose we could make it extern "C" unconditionally, that doesn't hurt.
And a separate symbol for icall-table.h and icall-decl.h to use for exporting?

jaykrell:

jaykrell:
What goes wrong is consistency.
1. ICALL_EXPORT was missing in some raw exports.
2. ICALL_EXPORT was inconsistent otherwise. icall-table.h does not apply it to the typed implementations.
So I think the typed implementations were sometimes extern "C", sometimes extern "C++".
All the raw implementations, i.e. sysmath.c, could be left alone. They were consistent and ok.

However, if everything is declared in icall-def.h with a signature or icall-decl.h, then ICALL_EXPORT becomes redundant (or incorrect) elsewhere.

jaykrell:

PR #11366, opened 2018-10-25
[System.Core]: Stub out `Interop.Sys.GetEUid` and `SetEUid`.

baulig:
This is a temporary fix; the native side still needs to be ported from c++ and tested.


baulig:
@monojenkins backport 2018-10

baulig:
@monojenkins commit csproj

PR #11370, opened 2018-10-25
[2018-10][corlib] Adds back accidentally removed legacy COM API to Binder

marek-safar:


marek-safar:

PR #11371, opened 2018-10-25
[corlib] Tweak more wasm defines

marek-safar:


marek-safar:
@monojenkins commit csproj

PR #11372, opened 2018-10-25
A couple of missing mono_class_from_mono_type_internal changes

nealef:
Two instances were missed when mono_class_from_mono_type was changed to mono_class_from_mono_type_internal


lewurm:
@monojenkins squash

PR #11377, opened 2018-10-25
[build] Enable netstandard facade in the "build" profile

akoeplinger:
It is necessary for a helix tool which I'm building in this profile and it relies on NS2.0.

Adding full NS2.0 would require building System.Transactions.dll, System.Runtime.Serialization.dll and System.Data.dll too but those in turn rely on a whole bunch of other junk that I don't want to bring
into the build profile. 

Therefore I just added stubs for the types directly into netstandard.dll.


akoeplinger:
@monojenkins commit csproj

PR #11379, opened 2018-10-25
[marshal] Delete unused locals in emit_managed_wrapper_ilgen

lambdageek:
`ex_local` and `e_local` were used prior to
https://github.com/mono/mono/commit/b20d5ce1e5d69512242dd42ae6ff1afc2483c16d
in order to store an exception raised by the managed method before passing it
to `ftnptr_eh_callback`.  After that commit, we invoke the callback from the
stack unwinder, rather than from the method wrapper.  As a result both of the
locals are now dead code.


luhenry:
@monojenkins squash

luhenry:
@monojenkins build failed

PR #11384, opened 2018-10-25
[Mono.Profiler.Log] Use full path for temporary .mlpd's

akoeplinger:
So they don't depend on the working dir.

### DIFF
--- a/mcs/class/Mono.Profiler.Log/Test/ProfilerTestRun.cs
+++ b/mcs/class/Mono.Profiler.Log/Test/ProfilerTestRun.cs
@@ -46,7 +46,7 @@ namespace MonoTests.Mono.Profiler.Log {
 			_currentProcess = Process.GetCurrentProcess();
 			Name = name;
 			Options = options;
-			_output = $"test-{_id++}.mlpd";
+			_output = Path.GetFullPath ($"test-{_id++}.mlpd");
 		}
 
 		public void Run (Action<IReadOnlyList<LogEvent>> action)


mcs/class/Mono.Profiler.Log/Test/ProfilerTestRun.cs
```diff
@@ -37,16 +37,28 @@ public override void VisitBefore (LogEvent ev)
...
+		{
+			_testAssemblyPath = Path.Combine (Path.GetDirectoryName (System.Reflection.Assembly.GetExecutingAssembly ().Location), "log-profiler-test.exe");
+			_currentProcess = Process.GetCurrentProcess();
+			_tempDirectory = Path.Combine (Path.GetTempPath(), Path.GetRandomFileName ());
+			Directory.CreateDirectory (_tempDirectory);
+		}
+
+		~ProfilerTestRun()
+		{
+			File.Delete (_output);
```
alexrp: Given that profiler test failures are often hard to reproduce, I prefer to keep the MLPDs around when possible to help diagnose a failure. This is mostly useful locally, so maybe we could delete the files only on CI, if doing so is actually necessary?

mcs/class/Mono.Profiler.Log/Test/ProfilerTestRun.cs
```diff
@@ -84,4 +96,4 @@ IReadOnlyList<LogEvent> ParseFile ()
 			return visitor.Events;
 		}
 	}
-}
+}
```
alexrp: line damage

mcs/class/Mono.Profiler.Log/Test/ProfilerTestRun.cs
alexrp: For the same reasons, I'm not sure it's a good idea to just throw these files in a random/unknowable directory in `/tmp`.

akoeplinger: Hmm, now that I think about it more I'm not sure I really need the files to be outside the working directory. This was part of my work on getting the tests running in the Helix environment but I probably only need to make `_output` a full instead of relative path. I'll try and come back :)

akoeplinger: Yup confirmed, using Path.GetFullPath() is enough. Updated.

alexrp:
The change to move those static field initializations to a static constructor was good regardless, though.

PR #11387, opened 2018-10-26
Be clear about null objects and 0 gchandles.

jaykrell:


jaykrell:

jaykrell:
@monojenkins build failed

mono/metadata/boehm-gc.c
```diff
@@ -1785,6 +1785,8 @@ handle_data_grow (HandleData *handles, gboolean track)
 static guint32
 alloc_handle (HandleData *handles, MonoObject *obj, gboolean track)
 {
+	if (!obj)
```
kumpera: We can't do that because gchandles are mutable cells.
It's valid to alloc a handle to `null` then set its value to something else and if we canonicalize this case, we'll break when someone mutates it.

Issue #11392, opened 2018-10-26
appdomain-threadpool-unload


marek-safar:
## Steps to Reproduce

1. run mono/tests/appdomain-threadpool-unload.exe test
2. observe flaky behavior

### Current Behavior

Random crashes
- https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx/15963/testReport/junit/MonoTests/runtime/appdomain_threadpool_unload_exe/

```
mono(61025,0x700008faf000) malloc: *** error for object 0x7fd9d966d1d0: double free
*** set a breakpoint in malloc_error_break to debug
Stacktrace:

  at <unknown> <0xffffffff>
  at System.AppDomain.Load (string,System.Security.Policy.Evidence,bool) [0x00021] in <678443b6ac354440ad26f56c500de5de>:0
  at System.AppDomain.Load (string) [0x00000] in <678443b6ac354440ad26f56c500de5de>:0
  at (wrapper remoting-invoke-with-check) System.AppDomain.Load (string) [0x00032] in <678443b6ac354440ad26f56c500de5de>:0
  at System.Reflection.Assembly.Load (string) [0x00005] in <678443b6ac354440ad26f56c500de5de>:0
  at System.Runtime.Serialization.Formatters.Binary.ObjectReader.FastBindToType (string,string) [0x0005a] in <678443b6ac354440ad26f56c500de5de>:0
  at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Bind (string,string) [0x0001b] in <678443b6ac354440ad26f56c500de5de>:0
  at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType (System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo,string) [0x00067] in <678443b6ac354440ad26f56c500de5de>:0
  at System.Runtime.Serialization.Formatters.Binary.ObjectMap..ctor (string,string[],System.Runtime.Serialization.Formatters.Binary.BinaryTypeEnum[],object[],int[],System.Runtime.Serialization.Formatters.Binary.ObjectReader,int,System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo,System.Runtime.Serialization.Formatters.Binary.SizedArray) [0x00060] in <678443b6ac354440ad26f56c500de5de>:0
[...]

Issue #10803, opened 2018-09-25
appdomain-threadpool-unload.exe timeout Linux amd64 interpreter


jaykrell:
## Steps to Reproduce

1. Submit a PR.

### Current Behavior

e.g.

https://github.com/mono/mono/pull/10733

Linux amd64 Interpreter.

https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/10123/parsed_console/log.html

appdomain-threadpool-unload.exe timed out (300s)

1 test(s) timed out

Timed out test(s):

appdomain-threadpool-unload.exe
=============== appdomain-threadpool-unload.exe.stdout ===============
.........................
.........................
.........................

Id Target Id Frame

1 Thread 0x7f0404d47740 (LWP 10039) "mono" 0x00007f040421215f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
2 Thread 0x7f04037ff700 (LWP 10139) "SGen worker" 0x00007f040421215f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
3 Thread 0x7f0401bb5700 (LWP 10140) "Finalizer" 0x00007f0404214536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
4 Thread 0x7f040163d700 (LWP 10150) "mono" 0x00007f0404214536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
5 Thread 0x7f040143c700 (LWP 10152) "Thread Pool Wor" 0x00007f0404214536 in do_futex_wait.constprop () from /lib/x86_64-linux-gnu/libpthread.so.0
6 Thread 0x7f040123b700 (LWP 10157) "Thread Pool Wor" 0x00007f040421215f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/x86_64-linux-gnu/libpthread.so.0
7 Thread 0x7f040103a700 (LWP 10165) "Thread Pool Wor" 0x00007f0404214f5c in __lll_lock_wait () from /lib/x86_64-linux-gnu/libpthread.so.0
8 Thread 0x7f0400e39700 (LWP 10172) "Thread Pool Wor" 0x00007f0404214f5c in __lll_lock_wait () from /lib/x86_64-linux-gnu/libpthread.so.0
[...]

PR #11393, opened 2018-10-26
[runtime] Disables crashing appdomain-threadpool-unload.exe test

marek-safar:
https://github.com/mono/mono/issues/11392


jaykrell:
It also hangs:
https://github.com/mono/mono/issues/10803

PR #11401, opened 2018-10-26
[corlib] Replace few more simple types with corefx

marek-safar:


akoeplinger:

PR #11405, opened 2018-10-27
Native code frequently fails to honor the definitely-write semantics of out parameters, so use ref instead.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins buid Linux AArch64

jaykrell:

lewurm:
does that fix a specific failure?

jaykrell:
We should just never have this incorrectness or fragility. It is kinda a safety violation.

akoeplinger:
shouldn't we rather fix the native code to make sure it always writes a value into the parameters?

jaykrell:
Fixing native side is ok. Resulting efficiency is essentially identical. I’d prefer to discourage out parameters as generally error-prone.

jaykrell:
Out parameters are an optimization of ref, with good static checking, but only checked if you stay in C#. Not so good with unchecked C.

akoeplinger:
A quick grep shows we have about 800 other instances in mcs/class where we use `out` in a p/invoke so it'd take quite a while to clean them all up (and using `out` does document the intended data flow better).

I'm inclined to still take this however given you already invested the work, unless others object?

vargaz:
Note that c# locals are are zero initialized, so not setting them in native code is not very dangerous.

jaykrell:
I see your point— the “in” part of the “ref” is either “useful” or “initialized”. There is also disagreement as to if uninitialized integers are unsafe— they are still valid integers within the type system.

jaykrell:
@vargaz really?

vargaz:
I mean .net IL variables are zero initialized if the method has the initlocals flag set, which all c# methods do.

jaykrell:
That surprises me & all but moots this. I thought there was a more efficient thing where the C# compiler enforces definite assignment. Perhaps it comes to not trusting the IL & not being able to efficiently analyze at run time.
Corner case remains native should write for bad IL callers.

PR #11407, opened 2018-10-27
[Coop] Convert ves_icall_System_Environment_GetEnvironmentVariableNames.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins buid OS X x64

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build Mac OS X x64

jaykrell:
@monojenkins build OS X x64

mono/metadata/icall.c
```diff
@@ -6892,34 +6892,85 @@ ves_icall_System_Environment_GetCommandLineArgs (MonoError *error)
 }
 
 #ifndef HOST_WIN32
-static MonoArray *
-mono_icall_get_environment_variable_names (MonoError *error)
+
+static void
+mono_new_string_utf8_to_array (MonoArrayHandle array, gsize index,
```
luhenry: Why having this function in this file while `mono_new_string_utf16_to_array` is in `object.c`? How about you try to keep it consistent: either both functions in `object.c` or both functions here.

jaykrell: I think it was because of ifdef win32, because of this being the only use. But yeah, both in object.c not bad idea.

jaykrell: That is, in a larger change, there was maybe more uses of the one function & only one use of the other. Anyway, I can put them both in object.c.

jaykrell:
csc using system mono crashed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X x64

jaykrell:
Maybe those  new functions  should be called `_assign`.

mono/metadata/icall.c
```diff
@@ -6891,53 +6891,83 @@ ves_icall_System_Environment_GetCommandLineArgs (MonoError *error)
...
-	return_val_if_nok (error, NULL);
+	if (env_strings) {
+		n = 0;
+		WCHAR const *env_string = env_strings;
+		while (*env_string) {
+			/* weird case that MS seems to skip (per drive letter current working directory) */
+			if (*env_string != '=') {
+				WCHAR const * const equal_str = wcschr (env_string, '=');
+				g_assert (equal_str);
+				mono_string_new_utf16_to_handle (str, domain, env_string, (gsize)(equal_str - env_string), error);
```
luhenry: There is no need to introduce this new concept of `_to_handle`, expect if you come up with clear benchmark results pointing otherwise. Is it a backed-by-data problem, or is it something you just came up with? _I understand it will avoid allocating handles, no need to lecture on that_

For sake of getting this PR in, revert to using `mono_string_new_utf16_handle`, and submit another PR with this specific change, and we can discuss it there.

mono/metadata/icall.c
```diff
@@ -6891,53 +6891,83 @@ ves_icall_System_Environment_GetCommandLineArgs (MonoError *error)
...
+	return_val_if_nok (error, NULL_HANDLE_ARRAY);
 
 	n = 0;
 	for (e = environ; *e != 0; ++ e) {
-		parts = g_strsplit (*e, "=", 2);
-		if (*parts != 0) {
-			str = mono_string_new_checked (domain, *parts, error);
+		char **parts = g_strsplit (*e, "=", 2);
+		if (*parts) {
+			mono_string_new_utf8_to_handle (str, domain, *parts, strlen (*parts), error);
```
luhenry: Same as https://github.com/mono/mono/pull/11407/files#diff-eba0fd0052bf3ec9dd60f2f3e90f19e1R6927, revert to using `mono_string_new_utf8_handle`.

mono/metadata/icall.c
jaykrell: Assigning through out parameters is an old concept.

`to_array` is a new concept I just came up with and is not wearing well as I do more conversions, i.e. in custom-attr.c, which lead me to revisit it here.

I've had to repeatedly split up loops into multiple functions, making the code messier and slower.  It took me a while to realize the better approach.
The idiom of `_to_handle`, or better name,  `_assign`, avoids that repeatedly.

luhenry:
@jaykrell what's the status on that?

jaykrell:
@lambdageek agrees with `_to_handle` ("somewhat sympathetic, with reservations").
https://xamarinhq.slack.com/archives/C03CELHQQ/p1540836731087600
I suggest calling it `_assign`.

The original `mono_string_new_utf8_handle` and `mono_string_new_utf16_handle` are unacceptable because they are creating handles in loops.

I'll rename it `_assign` and resolve the merge conflict.

jaykrell:
Diff worth reducing otherwise though, too hard for me to read.

jaykrell:
@monojenkins build failed

jaykrell:

PR #11409, opened 2018-10-27
mono_signature_get_managed_fmt_string does not exist, remove prototype.

jaykrell:


jaykrell:

PR #11410, opened 2018-10-27
[w32file] Restrict mono_w32file_get_volume_information interface to its actual implementation and usage.

jaykrell:
On Unix and Wasm, mono_w32file_get_volume_information only implements last parameter.
On all platforms, mono_w32file_get_volume_information is only asked for its last parameter.
So replace mono_w32file_get_volume_information with mono_w32file_get_file_system_type that
only takes that parameter.


jaykrell:
system mono crashed in csc

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X i386

PR #11413, opened 2018-10-27
[wasm] Bitcode code generation improvements

vargaz:


vargaz:
@monojenkins build failed

mono/mini/mini-llvm.c
```diff
@@ -2477,17 +2507,49 @@ emit_get_method (MonoLLVMModule *module)
...
 	LLVMBuilderRef builder = LLVMCreateBuilder ();
+	LLVMValueRef table = NULL;
 	char *name;
 	int i;
+	gboolean emit_table = FALSE;
+
+#ifdef TARGET_WASM
+	/*
+	 * Emit a table of functions instead of a switch statement,
+	 * its very efficient on wasm. This might be usable on
```
jeromelaban: In which way is this more efficient ?

mono/mini/aot-runtime.c
```diff
@@ -4548,9 +4546,10 @@ static void
 init_llvmonly_method (MonoAotModule *amodule, guint32 method_index, MonoMethod *method, MonoClass *init_class, MonoGenericContext *context)
 {
 	ERROR_DECL (error);
+	gboolean res;
 
-	init_method (amodule, method_index, method, init_class, context, error);
-	if (!is_ok (error)) {
+	res = init_method (amodule, method_index, method, init_class, context, error);
```
luhenry: Shouldn't we guarantee that if `res != TRUE`, then `error` is always set to something not-ok?

luhenry:
@monojenkins build failed

mono/mini/mini-llvm.c
vargaz: On wasm, each function pointer is an int and it doesn't require runtime linking, so the whole thing is going to be an array of ints.

vargaz:
@monojenkins build failed

PR #11435, opened 2018-10-29
[crash] Fix merp invocation

alexanderkyte:
The change which made the handler return (rather than crashing the
runtime) did not consider this control flow path that causes the sigterm
signal handler to trigger the merp dumper if we return.

We now always return.

This was previously so that a random sigterm could be used to trigger a
dump. This isn't really used, and won't work right now anyways
(we don't register it until right before we dump). My fix is to remove
it, and to make the signal handler only work as part of the dumper
machinery.


alexanderkyte:
Forward-port of https://github.com/mono/mono/pull/11436

alexanderkyte:
@monojenkins backport 2018-08
@monojenkins backport 2018-10

Issue #11385, opened 2018-10-25
MonoTests.DebuggerTests.Crash test is flaky


akoeplinger:
## Steps to Reproduce

1. Run Mono.Debugger.Soft tests
2. Observe flaky behavior of MonoTests.DebuggerTests.Crash

### Current Behavior

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64/20338/
https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-cxx/1872/

```
Mono.Debugger.Soft.VMDisconnectedException : Exception of type 'Mono.Debugger.Soft.VMDisconnectedException' was thrown.

  at Mono.Debugger.Soft.Connection.disconnected_check () [0x00018] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs:1323 
  at Mono.Debugger.Soft.Connection.SendReceive (Mono.Debugger.Soft.Connection+CommandSet command_set, System.Int32 command, Mono.Debugger.Soft.Connection+PacketWriter packet) [0x00009] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs:1602 
  at Mono.Debugger.Soft.Connection.SendReceive (Mono.Debugger.Soft.Connection+CommandSet command_set, System.Int32 command) [0x00000] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs:1644 
  at Mono.Debugger.Soft.Connection.VM_Dispose () [0x00000] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs:1749 
  at Mono.Debugger.Soft.VirtualMachine.Detach () [0x00000] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs:139 
  at MonoTests.DebuggerTests.Crash () [0x0006c] in /mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mcs/class/Mono.Debugger.Soft/Test/dtest.cs:2480 
[...]

PR #11438, opened 2018-10-29
[crash] Catch exception for Debugger Crash test, reenable

alexanderkyte:
This might be enough if the problem was with the uncaught exceptions.

For continued flake risks, can we have a tighter bound on how often this is flaking? We should be able to handle it with some retries. If it's flaking more often than that, a post-send pause might need to be enabled on the debuggee side.

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Test/dtest.cs
+++ b/mcs/class/Mono.Debugger.Soft/Test/dtest.cs
@@ -2456,7 +2456,6 @@ public class DebuggerTests
 
 	[Test]
 	[Category("NotOnWindows")]
-	[Ignore("https://github.com/mono/mono/issues/11385")]
 	public void Crash () {
 		bool success = false;
 
@@ -2476,13 +2475,17 @@ public class DebuggerTests
 
 				break;
 			}
-		} finally {
-			try {
-				vm.Detach ();
-			} finally {
-				vm = null;
-			}
+		} catch (VMDisconnectedException exc) {
+			Assert.Fail ("Got VMDisconnectedException rather than crash event");
 		}
+
+		try {
+			vm.Detach ();
+		} catch (VMDisconnectedException exc) {
+			// This is fine
+		}
+
+		vm = null;
 
 		if (!success)
 			Assert.Fail ("Didn't get crash event");


alexanderkyte:
@akoeplinger you filed https://github.com/mono/mono/issues/11385 . I see two failures linked there. I've re-enabled the test here but changed it to catch the exception in question. 

If we've got an issue with the flake rate being too high, there are some heuristics we can tune here to improve the reliability of the handler. There's a scalar "wait" argument passed to the dumper that's `0` right now that we can bump.

vargaz:
As this is a newly added test, it should not be flaky.

alexanderkyte:
@vargaz I agree with you. But the question is how often is it happening. If we see a very rare failure, I think that reflects an unavoidable instability in doing something like this while crashing. If it's happening more often, that means we need to make the mechanism more reliable.

vargaz:
Since the crash happens in a very controlled environment, the test should be reliably in theory.

alexanderkyte:
I think we're depending on the debugger sending the event before the socket is closed, which doesn't happen in some cases. The problem is that there's a lot of delayed queueing of event sending. 

I think I can throw together something that will deterministically keep the VM alive until that socket can be closed and things can shut down.

alexanderkyte:
I've been trying to reproduce on many machines for more than 14 hours and I haven't seen one test failure of this type. Using x64, x32, and armv7

mcs/class/Mono.Debugger.Soft/Test/dtest.cs
```diff
@@ -2476,14 +2475,18 @@ Event step_out_await (string method, Event e)
...
 				break;
 			}
-		} finally {
-			try {
-				vm.Detach ();
-			} finally {
-				vm = null;
-			}
+		} catch (VMDisconnectedException exc) {
+			Assert.Fail ("Got VMDisconnectedException rather than crash event");
```
luhenry: It's better to let the exception go through in that case, so we know what's the root failure.

You should keep part of the previously existing code, while still adding what you have here, as followed:
```
		} finally {
			try {
				vm.Detach ();

			} catch (VMDisconnectedException) {
				// This is fine
			} finally {
				vm = null;
			}
		}

alexanderkyte:
I want to re-enable this test after https://github.com/mono/mono/pull/12125 is merged and green.

alexanderkyte:
Closing because https://github.com/mono/mono/pull/12304 merged

PR #11445, opened 2018-10-30
[System.Core] Adds ~28k tests

marek-safar:


EgorBo:
@marek-safar @akoeplinger   hm.. I think these tests are already exist in corlib_xtest

marek-safar:
@EgorBo will remove them from mscorlib

PR #11447, opened 2018-10-30
[interp] fix definition of MINT_CKNULL_N

lewurm:
So that the debug output renders the operand of `MINT_CKNULL_N` correctly.


lewurm:
@monojenkins squash

PR #11466, opened 2018-10-30
[coop] Refactor/reuse mono_value_box_handle/mono_value_box_checked and reduce raw pointers.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/object-internals.h
```diff
@@ -2160,6 +2160,8 @@ mono_string_empty_handle (MonoDomain *domain);
 gpointer
 mono_object_get_data (MonoObject *o);
 
+#define mono_get_data_unsafe(handle) ((guint8*)MONO_HANDLE_RAW (handle) + MONO_ABI_SIZEOF (MonoObject))
```
jaykrell: Not clear what type this ought to return.
1. void*
2. char*
3. uint8*?
4. typeof(MONO_HANDLE_RAW(handle))?
5. uintptr


Current guint8* just bleeds from implementation.

lambdageek: nit: mention `valuetype` somewhere in the name? Don't we have some function that does this already?

For C, probably `void*` made the most sense; in C++, `char*` is probably what we'll want.

mono/metadata/object-internals.h
```diff
@@ -2331,4 +2333,7 @@ mono_gc_wbarrier_value_copy_internal (void* dest, /*const*/ void* src, int count
 void
 mono_gc_wbarrier_object_copy_internal (MonoObject* obj, MonoObject *src);
 
+#define MONO_BEGIN_UNSUSPENDABLE /* nothing yet */
```
lambdageek: I think these will:
- be names `MONO_ENTER_NO_SAFEPOINTS`/`MONO_EXIT_NO_SAFEPOINTS`
- live in `mono/utils/mono-threads-api.h`

mono/metadata/object-internals.h
jaykrell: The other is right above `mono_object_get_data`. Nothing about `valuetype`. Or `unbox`ing.
I find `get_data` pretty non specific, but it is there.
I could say `mono_handle_get_data`, but `handle` is supposed to be somewhat superfluous. So is `object`. Hey, what else is mono dealing with, but objects and handles all over the place?

I'm flexible here.

Really in C++ -- overload, it just remains called the same old `mono_object_get_data` -- *and less churn* and no loss of clarity -- you don't really care if it is object or handle, it just works and does essentially the same thing.

luhenry: Why the need to diverge so much from the existing `mono_object_get_data`? Neither the return type, nor the name is equivalent. How about you try to make it more consistent?

mono/metadata/object-internals.h
```diff
@@ -2160,6 +2160,8 @@ mono_string_empty_handle (MonoDomain *domain);
 gpointer
 mono_object_get_data (MonoObject *o);
 
+#define mono_get_data_unsafe(handle) ((void*)((guint8*)MONO_HANDLE_RAW (handle) + MONO_ABI_SIZEOF (MonoObject)))
```
luhenry: Why the need to diverge so much from the existing mono_object_get_data? Neither the return type, nor the name is equivalent. How about you try to make it more consistent?

mono/metadata/object-internals.h
jaykrell: Return type good point, I raised that, but missed the consistency point, and is already fixed, except for the exact name of the same type, I can fix that too.

On the name, well, change "object" to "handle"? I considered that, but there was the prior goal of not adding "handle" to names as well, because these are the new better names and shouldn't be warty. Perhaps we have done that enough though -- lots of functions with "handle" in name, and saying nothing is too vague, even with "object" or "handle" it seems vague, and this is now worse, so I'll do that.

Ultimately with C++ the name can be identical via overloading and that would be ideal presumably in terms of name choice, assuming the original name is good, and reducing churn, but we aren't here.

jaykrell:
There remains the deliberate inconsistency of appending `_unsafe` to name.
Alternative is to pin.
I believe that is our general new pattern -- either pin or label things unsafe.

jaykrell:
@monojenkins build failed

mono/utils/mono-threads-api.h
```diff
@@ -157,6 +158,10 @@ mono_threads_exit_no_safepoints_region (const char *func);
 		if (mono_threads_are_safepoints_enabled ())			\
 			mono_threads_exit_no_safepoints_region (__func__);	\
 	} while (0)
+=======
+#define MONO_ENTER_NO_SAFEPOINTS /* nothing yet */
+#define MONO_EXIT_NO_SAFEPOINTS /* nothing yet */
+>>>>>>> Line up better with https://github.com/mono/mono/pull/11476
```
lewurm: merging went wrong

Issue #11465, opened 2018-10-30
Threading: implement MONO_ENTER_NO_SAFEPOINTS/MONO_EXIT_NO_SAFEPOINTS


lambdageek:
# Summary #

Let's add a scope that causes a runtime assertion if the thread attempt to cooperatively suspend while in that region, or if it attempts to switch to GC Safe mode.

# Motivation #

We often have places in the code where we need to do something to a managed object while in GC Unsafe where we "know" that a cooperative safepoint won't occur.  For example we may temporarily take a raw pointer to a MonoObject for a valuetype that has no reference types in it and memcpy it somewhere.  We could take a gchandle to pin the object, memcpy and then unpin.  Or we could wrap the entire block in a `MONO_ENTER_NO_SAFEPOINTS`/`MONO_EXIT_NO_SAFEPOINTS` pair which would assert at runtime if we later added a call to `mono_object_new` in that block (or anything similar).

# Implementation idea #

Take a bit in the thread state machine.  add transitions that set the bit only from `RUNNING` state.  Disallow transitions from `RUNNING`+`UNINTERRUPTIBLE` to `BLOCKING`.  Transition from `RUNNING`+`UNINTERRUPTIBLE` to `ASYNC_SUSPEND_REQUESTED`+`UNINTERRUPTIBLE` (ie - suspend initiator asks victim thread to suspsend - and the victim is expected to suspend next time it reaches a safepoint) is allowed.  but polling from `ASYNC_SUSPEND_REQUESTED`+`UNINTERRUPTIBLE` would assert.

**Tricky bit** don't let an `UNINTERRUPTIBLE` thread become the GC Initiator, either! so if the current thread is going to stop the world, it better not be uninterruptible.

(This only makes sense if we're not using preemptive suspend - so define the macros as a no-op if preemptive suspend is enabled.)

jaykrell:
nits: NO_SUSPEND  is shorter. Not sure if banning a verb or describing a region is preferred.
END will be `}` if C++.

lambdageek:
or maybe even `MONO_BEGIN_NO_SAFEPOINTS`/`MONO_END_NO_SAFEPOINTS` to be completely direct about it.

I think it's similar to `ENTER_CRITICAL_REGION`, but the mechanisms are different - critical regions are about async suspension - we're guaranteeing that the thread won't be interrupted by the suspend machinery.   `NO_SAFEPOINTS` is about cooperative suspension.

[...]

PR #11476, opened 2018-10-30
Implement MONO_ENTER_NO_SAFEPOINTS/MONO_EXIT_NO_SAFEPOINTS

lambdageek:
Implement the idea from #11465: if a region of code is marked with this pair of macros, it:
- must be in GC Unsafe
- must not safepoint
- must not call the managed allocator
- must not transition to GC Safe
- must not nest a `MONO_ENTER_NO_SAFEPOINTS`/`MONO_EXIT_NO_SAFEPOINTS` pair

In return, the region gets to dereference managed pointers without having to pin the memory.
If writing, still have to use write barriers.

----

Best to review this one commit at a time, I think.

See #11466 for an example of how this may be used.


mono/metadata/sgen-stw.c
```diff
@@ -253,6 +253,9 @@ sgen_unified_suspend_stop_world (void)
 {
 	int sleep_duration = -1;
 
+	// we can't lead STW if we promised not to safepoint.
+	g_assert (!mono_thread_info_wont_safepoint (mono_thread_info_current ()));
```
jaykrell: Nit: Contradictions should be avoided. It should be `will_not` or `no`.

mono/utils/mono-threads-coop.c
```diff
@@ -129,6 +129,9 @@ mono_threads_state_poll_with_info (MonoThreadInfo *info)
 
 	THREADS_SUSPEND_DEBUG ("FINISH SELF SUSPEND OF %p\n", mono_thread_info_get_tid (info));
 
+	/* Fast fail if wont_safept is set */
+	g_assert (!(info->thread_state & THREAD_SUSPEND_WONT_SAFEPT_MASK));
+
```
jaykrell: Nit: More hard to read abbreviations. `WONT` and `SAFEPT` are in poor form imho.

mono/utils/mono-threads-state-machine.c
```diff
@@ -27,13 +27,20 @@ get_thread_suspend_count (int thread_state)
 	return (thread_state & THREAD_SUSPEND_COUNT_MASK) >> THREAD_SUSPEND_COUNT_SHIFT;
 }
 
+static inline gboolean
+get_thread_wont_safept (int thread_state)
+{
+	return (thread_state & THREAD_SUSPEND_WONT_SAFEPT_MASK) >> THREAD_SUSPEND_WONT_SAFEPT_SHIFT;
```
jaykrell: nits: `wont` and `safept` should be spelled out, and seems like a good place for a bitfield.

mono/utils/mono-threads-state-machine.c
```diff
@@ -27,13 +27,20 @@ get_thread_suspend_count (int thread_state)
...
+}
+
 static inline int
-build_thread_state (int thread_state, int suspend_count) 
+build_thread_state (int thread_state, int suspend_count, gboolean wont_safept) 
 {
 	g_assert (suspend_count >= 0 && suspend_count <= THREAD_SUSPEND_COUNT_MAX);
 	g_assert (thread_state >= 0 && thread_state <= STATE_MAX);
+	g_assert (wont_safept == 0 || wont_safept == 1);
```
jaykrell: Nit: Assert kinda redundant with `gboolean`, but I understand, it isn't quite.
If we ever get C++ `bool`...

mono/utils/mono-threads-state-machine.c
```diff
@@ -90,14 +104,16 @@ check_thread_state (MonoThreadInfo* info)
...
-trace_state_change_with_func (const char *transition, MonoThreadInfo *info, int cur_raw_state, int next_state, int suspend_count_delta, const char *func)
+trace_state_change_with_func (const char *transition, MonoThreadInfo *info, int cur_raw_state, int next_state, gboolean next_wont_safept, int suspend_count_delta, const char *func)
 {
 	check_thread_state (info);
-	THREADS_STATE_MACHINE_DEBUG ("[%s][%p] %s -> %s (%d -> %d) %s\n",
+	THREADS_STATE_MACHINE_DEBUG ("[%s][%p] %s %s -> %s %s (%d -> %d) %s\n",
 		transition,
 		mono_thread_info_get_tid (info),
 		state_name (get_thread_state (cur_raw_state)),
+		(get_thread_wont_safept (cur_raw_state) ? "X" : "."),
```
jaykrell: nit: Could be `char` instead of `char*`.

mono/utils/mono-threads-state-machine.c
```diff
@@ -206,37 +230,42 @@ MonoRequestSuspendResult
...
 	case STATE_BLOCKING:
 		if (!(suspend_count == 0))
 			mono_fatal_with_history ("suspend_count = %d, but should be == 0", suspend_count);
-		if (mono_atomic_cas_i32 (&info->thread_state, build_thread_state (STATE_BLOCKING_SUSPEND_REQUESTED, 1), raw_state) != raw_state)
+		if (wont_safept)
+			mono_fatal_with_history ("wont_safept = TRUE, but should be FALSE");
+		if (mono_atomic_cas_i32 (&info->thread_state, build_thread_state (STATE_BLOCKING_SUSPEND_REQUESTED, 1, wont_safept), raw_state) != raw_state)
 			goto retry_state_change;
-		trace_state_change ("SUSPEND_INIT_REQUESTED", info, raw_state, STATE_BLOCKING_SUSPEND_REQUESTED, 1);
+		trace_state_change ("SUSPEND_INIT_REQUESTED", info, raw_state, STATE_BLOCKING_SUSPEND_REQUESTED, wont_safept, 1);
```
jaykrell: Contradictions really are ugly. `wont` is actually a word with a different meaning.

mono/utils/mono-threads.h
```diff
@@ -714,6 +719,8 @@ int mono_thread_info_suspend_count (THREAD_INFO_TYPE *info);
 int mono_thread_info_current_state (THREAD_INFO_TYPE *info);
 const char* mono_thread_state_name (int state);
 gboolean mono_thread_is_gc_unsafe_mode (void);
+G_EXTERN_C // due to THREAD_INFO_TYPE varying
```
jaykrell: Ludovic wants `G_EXTERN_C` on same line going forward.
I guess I should have eliminated `THREAD_INFO_TYPE` by now.

mono/utils/mono-threads-api.h
```diff
@@ -140,6 +140,22 @@ For further explanation of what can and can't be done in GC unsafe mode:
...
+mono_threads_exit_no_safepoints_region (const char *func);
+
+#define MONO_ENTER_NO_SAFEPOINTS					\
+	do {								\
+		if (mono_threads_are_safepoints_enabled ())		\
+			mono_threads_enter_no_safepoints_region (__func__)
+
+#define MONO_EXIT_NO_SAFEPOINTS							\
+		if (mono_threads_are_safepoints_enabled ())			\
+			mono_threads_exit_no_safepoints_region (__func__);	\
```
jaykrell: I wonder if this even worth a macro.
I get that the braces force a balance.
But one could just call mono_threads_enter/exit_no_safepoints_region () and avoid the ugly.
In C++ the end macro would just a brace -- destructor.

mono/utils/mono-threads-coop.c
```diff
@@ -707,3 +707,17 @@ mono_threads_coop_end_global_suspend (void)
 	if (mono_threads_are_safepoints_enabled ())
 		mono_polling_required = 0;
 }
+
+void
+mono_threads_enter_no_safepoints_region (const char *func)
+{
+	MONO_REQ_GC_UNSAFE_MODE;
+	mono_threads_transition_begin_wont_safept (mono_thread_info_current (), func);
```
jaykrell: So. In a separate PR, I suggest we have a global non-thread local `mono_thread_info_current_var`, pointer, const, always null. You pass that. At least in some macros, maybe not here.

A function needing the data, if it gets null, will call `mono_thread_info_current()`

Macros like HANDLE_FUNCTION_BEGIN that call `mono_thread_info_current()` anyway, will store in a local named `mono_thread_info_current_var`.

Therefore TLS accesses will be optimized down.

If/as perf sensitive functions are found, pass the local as a parameter.

mono/utils/mono-threads-coop.c
lambdageek: going to `s/wont_safept/no_safepoints/ig`

mono/utils/mono-threads-state-machine.c
lambdageek: Maybe `no_safepoints = !!no_safepoints; // ensure it's 0 or 1`?
I'm paranoid about bitmasks and `gboolean` not really being a single bit.

(also renaming wont_safept to no_safepoints as elsewhere)

mono/utils/mono-threads-state-machine.c
lambdageek: i don't like `can_safepoint` because the bit applies to many states where safepointing just doesn't make sense:  `STATE_SELF_SUSPENDED | can_safepoint`, for example, just confuses me - it can't really safepoint - it's not even running.

mono/utils/mono-threads-api.h
lambdageek: there are a couple of reasons for the macro:
1. I want to block scope these things.
2. if we decide we want to allow nesting no-safepoints scopes, in C, I'm going to need to declare a var to propagate some state from the enter to the exit.
3. consistency with the other `MONO_ENTER_*`/`MONO_EXIT_*` macros.

mono/utils/mono-threads-coop.c
lambdageek: ok, I'll update to pass the thread info and reconcile with #11487 once it's in.

lambdageek:
@monojenkins build failed

mono/utils/mono-threads.h
luhenry: @jaykrell I never said that as a general rule, I certainly said to keep consistent with the surrounding context though  :wink:

mono/utils/mono-threads-api.h
jaykrell: All three reasons are really: C needs them, C++ does not.

PR #11487, opened 2018-10-31
Reduce TLS accesses.

jaykrell:
Functions with HANDLE_FUNCTION_ENTER, only call mono_thread_info_current once,
instead of 1 + per-MONO_HANDLE_NEW.

This can be extended in future in multiple potential ways.
 1 Macro split from HANDLE_FUNCTION_ENTER.
   This is easiest/no downside, i.e. for functions that do not HANDLE_FUNCTION_ENTER
   but that do multiple MONO_HANDLE_NEW, call other macro
   just to mono_thread_info_current () into local.
 2 Pass TLS as separate parameter. Probably not actually, see 4.
 3 Store TLS in handles -- too large? Probably not actually, see next.
 4 Store TLS in MonoError. i.e. MonoError is "call context", chained
   up to "thread context".

### DIFF
--- a/mono/metadata/handle.c
+++ b/mono/metadata/handle.c
@@ -155,15 +155,17 @@ mono_handle_chunk_leak_check (HandleStack *handles) {
 }
 #endif
 
+extern MonoThreadInfo * const mono_thread_info_current_var = NULL;
+
 /* Actual handles implementation */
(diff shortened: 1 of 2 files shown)


jaykrell:
@monojenkins build failed

PR #11490, opened 2018-10-31
[runtime] allocate volatile_{locals,args} bitset in image mempool

lewurm:
Fixes https://github.com/mono/mono/issues/11488, an regression introduced by https://github.com/mono/mono/pull/11294

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6373,11 +6373,20 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 				case ICALL_HANDLES_WRAP_OBJ_OUT:
 					// FIXME better type
 					local = mono_mb_add_local (mb, mono_get_object_type ());
-					mono_bitset_set_safe (&mb->volatile_locals, local);
+
+					if (!mb->volatile_locals) {
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
+						mb->volatile_locals = mono_bitset_mem_new (mem, csig->param_count + 1, 0);
+					}
+					mono_bitset_set (mb->volatile_locals, local);
 					break;
 				case ICALL_HANDLES_WRAP_VALUETYPE_REF:
 				case ICALL_HANDLES_WRAP_OBJ:
-					mono_bitset_set_safe (&mb->volatile_args, i);
+					if (!mb->volatile_args) {
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
+						mb->volatile_args = mono_bitset_mem_new (mem, csig->param_count + 1, 0);
+					}
+					mono_bitset_set (mb->volatile_args, i);
 					break;
 				case ICALL_HANDLES_WRAP_NONE:
 					break;
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6375,7 +6375,7 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 					local = mono_mb_add_local (mb, mono_get_object_type ());
 
 					if (!mb->volatile_locals) {
-						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), mono_bitset_alloc_size (csig->param_count + 1));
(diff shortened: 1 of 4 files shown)


mono/mini/mini.c
```diff
@@ -2079,12 +2079,6 @@ mono_compile_create_vars (MonoCompile *cfg)
 		mono_apply_volatile (cfg->locals [i], header->volatile_locals, i);
 	}
 
-	mono_bitset_free (header->volatile_locals);
```
jaykrell: I was going to grow the header to hold the bits, and copy them here, had even written that in the initial work, but this should also work.

mono/metadata/marshal-ilgen.c
```diff
@@ -6373,11 +6373,20 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
+						mb->volatile_locals = mono_bitset_mem_new (mem, csig->param_count + 1, 0);
+					}
+					mono_bitset_set (mb->volatile_locals, local);
 					break;
 				case ICALL_HANDLES_WRAP_VALUETYPE_REF:
 				case ICALL_HANDLES_WRAP_OBJ:
-					mono_bitset_set_safe (&mb->volatile_args, i);
+					if (!mb->volatile_args) {
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
```
jaykrell: Size should be `mono_bitset_alloc_size (csig->param_count + 1)`.

mono/metadata/marshal-ilgen.c
```diff
@@ -6373,11 +6373,20 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 				case ICALL_HANDLES_WRAP_OBJ_OUT:
 					// FIXME better type
 					local = mono_mb_add_local (mb, mono_get_object_type ());
-					mono_bitset_set_safe (&mb->volatile_locals, local);
+
+					if (!mb->volatile_locals) {
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), csig->param_count + 1);
```
jaykrell: Size should be mono_bitset_alloc_size (csig->param_count + 1).

jaykrell:
`mono_bitset_resize` also becomes dead.
Though those functions aren't wrong in general and could be preserved.
They just assume a certain allocation strategy which isn't always correct.
You could also set the MONO_BITSET_DONT_FREE flag.

jaykrell:
I'll have an alterate PR here shortly..but yours maybe simpler.

mono/mini/mini.c
lewurm: growing for this usecase is not necessary, just makes it more complicated than it needs to be.

lewurm:
`MONO_BITSET_DONT_FREE` is set by `mono_bitset_mem_new ()`.

let's remove `mono_bitset_resize ()` too, since it uses `mono_bitset_new ()` (therefore `malloc ()`) to resize the bitset, which is not what you want when you use a mempool for example.

jaykrell:
Can this occur in a dynamic image?
I mean, in https://github.com/mono/mono/pull/11491, I took the approach of making the bitset lifetime equal the header lifetime, which would seem to cover an additional scenario?

mono/metadata/marshal-ilgen.c
```diff
@@ -6373,11 +6373,20 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), mono_bitset_alloc_size (csig->param_count + 1, 0));
+						mb->volatile_locals = mono_bitset_mem_new (mem, csig->param_count + 1, 0);
+					}
+					mono_bitset_set (mb->volatile_locals, local);
 					break;
 				case ICALL_HANDLES_WRAP_VALUETYPE_REF:
 				case ICALL_HANDLES_WRAP_OBJ:
-					mono_bitset_set_safe (&mb->volatile_args, i);
+					if (!mb->volatile_args) {
+						char *mem = (char *)mono_image_alloc0 (get_method_image (method), mono_bitset_alloc_size (csig->param_count + 1, 0));
```
jaykrell: Char* should just be gpointer, and no casts.

lewurm: https://github.com/mono/mono/pull/11499

lewurm:
@jaykrell what would be the issue with dynamic images?

PR #11493, opened 2018-10-31
"Mono.Native - Halloween Edition"

baulig:
In preparation of the upcoming FSW support, `Mono.Native` now supports icall registration
and has a test added for this.  This doesn't actually hook up FSW yet on iOS, but all the
preparations have been made as well as some tests for it.

* When using Bitcode (iOS only), we now link against the Mono runtime because we cannot have
  any undefined symbols.  Without Bitcode, the `mtouch` tool will do the linking for us.

* Add `-framework GSS` if needed.

* Removed `common/MonoNativeConfig.cs` - the test harness in `xamarin-macios/tests/xharness`
  now takes care of that.

* Include a few more `System.Native` files on IOS.

* `mono_native_initialize()`: New native method to initialize the module and register any
  necessary icalls.  We register a new `Mono.MonoNativePlatform::IncrementInternalCounter`
  icall for testing purposes.

* `mono_native_is_initialized()`: New native method for testing.

* `MonoNativePlatform.Initialize()`, `IsInitialized()`: Managed side of these.

* `MonoNativePlatform.IncrementInternalCounter()`: New icall, which is registered by
  `mono_native_initialize()` - this is used by the test suite to make sure the icall
  registration works.

* `NativePlatformTest`: Removed compat / unified check, see above.  Add tests to call
  the other managed APIs via reflection.


mcs/class/corlib/win32_net_4_x_corlib.dll.exclude.sources
```diff
@@ -6,6 +6,9 @@ corert/ThreadPoolBoundHandle.platformnotsupported.cs
 corefx/Interop.GetRandomBytes.Mono.cs
 corefx/Mono.SafePasswordHandle.Unix.cs
 ../../../external/corefx/src/System.Security.Cryptography.X509Certificates/src/Microsoft/Win32/SafeHandles/SafePasswordHandle.Unix.cs
+Mono/NativePlatformTest.cs
```
baulig: This should be
```
Mono/MonoNativePlatform.cs
Mono/MonoNativePlatformType.cs
```

mcs/class/corlib/win32_net_4_x_corlib.dll.exclude.sources
```diff
@@ -6,6 +6,9 @@ corert/ThreadPoolBoundHandle.platformnotsupported.cs
...
+../Mono/MonoNativePlatform.cs
```
baulig:
```suggestion
```

mcs/class/corlib/win32_net_4_x_corlib.dll.exclude.sources
```diff
@@ -6,6 +6,9 @@ corert/ThreadPoolBoundHandle.platformnotsupported.cs
...
+../Mono/MonoNativePlatformType.cs
```
baulig:
```suggestion
```

baulig:
@monojenkins commit csproj

baulig:
@monojenkins backport 2018-10

PR #11497, opened 2018-11-01
Possible cleanup of MONO_ENTER_NO_SAFEPOINTS / MONO_EXIT_NO_SAFEPOINTS.

jaykrell:


mono/utils/mono-threads-state-machine.c
```diff
@@ -742,17 +743,22 @@ Set the no_safepoints flag on an executing GC Unsafe thread.
...
 (Future: We could augment this function to return a return value that tells the
 thread to poll and retry the transition since if we enter here in the
 ASYNC_SUSPEND_REQUESTED state).
+
+This function is internal.
+The public interface is MONO_ENTER_NO_SAFEPOINTS.
  */
 void
-mono_threads_transition_begin_no_safepoints (MonoThreadInfo *info, const char *func)
+mono_threads_enter_no_safepoints_slow (MonoThreadInfo *info, const char *func)
```
lambdageek: Don't rename this.  We need to keep a clear distinction between the thread transition functions (the policy). and the layer that combines the policy with the mechanism.
In the case of `mono_threads_enter_no_safepoints_slow`:
1. it should be in `mono-threads-coop.c`
2. it should call `mono_threads_transition_begin_no_safepoints`
3. based on the return value (which is trivial)
4. it should execute the policy (which also trivial and empty in this case).
5. `mono_threads_transition_begin_no_safepoints` should take a non-null `MonoThreadInfo* info` argument - just like the rest of the state machine transition functions.

Let's keep this uniform.


Same for mono_exit_

mono/utils/mono-threads-state-machine.c
```diff
@@ -789,17 +795,22 @@ Unset the no_safepoints flag on an executing GC Unsafe thread.
...
 (Future: We could augment this function to perform the transition and then
 return a return value that tells the thread to poll (and safepoint) if we enter
 here in the ASYNC_SUSPEND_REQUESTED state).
+
+This function is internal.
+The public interface is MONO_EXIT_NO_SAFEPOINTS.
  */
 void
-mono_threads_transition_end_no_safepoints (MonoThreadInfo *info, const char *func)
+mono_threads_exit_no_safepoints (MonoThreadInfo *info, const char *func)
```
lambdageek: Same as enter_no_safepoints.  Keep the thread state transition functions as is.  Make a new `mono_threads_exit_no_safepoints` that ensures a non-null `MonoThreadInfo*` and calls this function.

mono/utils/mono-threads-coop.h
```diff
@@ -107,4 +107,36 @@ mono_threads_enter_gc_unsafe_region_unbalanced_with_info (THREAD_INFO_TYPE *info
...
+// that have raw pointers, but have otherwise
+// been converted to be cooperative suspend / precise
+// stack scan correct.
+//
+// Such spans of code will neither initiate a cooperative
+// suspend nor cooperate with such a request from another thread.
+// Therefore extending pause times under memory pressure,
+// or exhausting memory only due to refusal to collect.
+// Therefore such spans of code should be rare and small.
+#define MONO_ENTER_NO_SAFEPOINTS						\
```
lambdageek: One reason I had these in `mono-threads-api.h` is because that's where they logically belong along with the other user-facing thread transitions (enter/exit GC Safe/Unsafe).  I think there's a benefit to keeping the entire API in one place.

lambdageek:
I think the `mono_thread_info_var` plan, though.  I was going to do it myself last night, but ran out of time when I ran into a problem: `mono-threads-api.h` can't see it.

PR #11508, opened 2018-11-01
Improve *.sources and *.exclude.sources handling 

EgorBo:
Unifies algorithms to find correct *.sources files (and *.exclude.sources):

Example (profile=net_4_x, platform=win32):

take `win32_net_4_x_corlib_test.dll.sources`
else take `net_4_x_corlib_test.dll.sources`
else take `win32_corlib_test.dll.sources`
else take `corlib_test.dll.sources`

same for eclude.sources and xunit.

Fixes https://github.com/mono/mono/issues/11515


marek-safar:
@EgorBo this will probably need some file adjustments as well

EgorBo:
@monojenkins build

baulig:
I believe you also need to modify the gensources tool to strip the `.sources` extension - or remove it from the variable.

kg:
I suggest teaching gensources to strip the extension, it keeps the makefile simpler. Ideally the makefile would not have to do this at all

baulig:
Taking over.

PR #11512, opened 2018-11-01
[corlib] Update aot testing profile to run more tests

marek-safar:


mcs/class/System.Core/Makefile
```diff
@@ -31,7 +31,7 @@ ifneq (2.1, $(FRAMEWORK_VERSION))
 LIB_MCS_FLAGS += -d:NET_3_5
 endif
 
-MOBILE_NATIVE_INTERP :=
+MOBILE_NATIVE_INTERP := $(filter testing_aot_full_interp, $(PROFILE))
```
lewurm: interesting. for Xamarin.iOS, does it mean they are going to need to build two different BCLs? since the interpreter is optional.

marek-safar:
@monojenkins commit csproj

PR #10472, opened 2018-09-05
Handle ref returns in runtime invoke wrappers.

joncham:
Indirectly load ref return values in invoke wrappers and return by value.
Don't share wrappers with ref returns since indirect load is used.
Add soft debugger support for methods with ref returns.

PR #11514, opened 2018-11-02
Fix ref return for structs in runtime invoke wrapper.

joncham:
Use mono_mb_emit_op for CEE_LDOBJ to ensure wrapper data is added for
indirect load of a ref return value as method-to-ir expects this for wrappers.

Follow up to https://github.com/mono/mono/pull/10472


luhenry:

PR #11519, opened 2018-11-02
[Coop] Convert System_IO_DriveInfo.

jaykrell:


mono/metadata/icall.c
```diff
@@ -7330,35 +7326,29 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
...
+ves_icall_System_IO_DriveInfo_GetDiskFreeSpace (const gunichar2 *path_name, gint32 path_name_length, guint64 *free_bytes_avail,
 						guint64 *total_number_of_bytes, guint64 *total_number_of_free_bytes,
 						gint32 *error)
 {
-	gboolean result;
-
 	*error = ERROR_SUCCESS;
 
-	result = mono_w32file_get_disk_free_space (mono_string_chars_internal (path_name), free_bytes_avail, total_number_of_bytes, total_number_of_free_bytes);
+	if (free_bytes_avail)
```
luhenry: There is no need to check for `NULL`.

mono/metadata/icall.c
```diff
@@ -7330,35 +7326,29 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
...
+		*free_bytes_avail = (guint64)-1;
+
+	if (total_number_of_bytes)
```
luhenry: There is no need to check for `NULL`.

mono/metadata/icall.c
```diff
@@ -7330,35 +7326,29 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
...
+		*total_number_of_bytes = (guint64)-1;
+
+	if (total_number_of_free_bytes)
```
luhenry: There is no need to check for `NULL`.

mono/metadata/w32file-unix.c
```diff
@@ -4333,7 +4334,7 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 		utf8_path_name = g_strdup (g_get_current_dir());
 		if (utf8_path_name == NULL) {
 			mono_w32error_set_last (ERROR_DIRECTORY);
-			return(FALSE);
+			return FALSE;
```
luhenry: Line damage

mono/metadata/w32file-unix.c
```diff
@@ -4342,7 +4343,7 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 			mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_FILE, "%s: unicode conversion returned NULL", __func__);
 
 			mono_w32error_set_last (ERROR_INVALID_NAME);
-			return(FALSE);
+			return FALSE;
```
luhenry: Line damage

mono/metadata/w32file-unix.c
```diff
@@ -4371,17 +4372,15 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 	if (ret == -1) {
 		_wapi_set_last_error_from_errno ();
 		mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_FILE, "%s: statvfs failed: %s", __func__, g_strerror (errno));
-		return(FALSE);
+		return FALSE;
```
luhenry: Line damage

mono/metadata/w32file-unix.c
```diff
@@ -4371,17 +4372,15 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
...
 	}
 
 	/* total number of free bytes for non-root */
 	if (free_bytes_avail != NULL) {
-		if (isreadonly) {
+		if (isreadonly)
```
luhenry: Line damage

mono/metadata/w32file-unix.c
```diff
@@ -4391,35 +4390,15 @@ mono_w32file_get_disk_free_space (const gunichar2 *path_name, guint64 *free_byte
 
 	/* total number of bytes available for root */
 	if (total_number_of_free_bytes != NULL) {
-		if (isreadonly) {
+		if (isreadonly)
```
luhenry: Line damage

mono/metadata/w32file-win32.c
```diff
@@ -347,35 +347,38 @@ mono_w32file_create_pipe (gpointer *readpipe, gpointer *writepipe, guint32 size)
...
-	if (result) {
-		if (free_bytes_avail)
-			*free_bytes_avail = wapi_free_bytes_avail.QuadPart;
-		if (total_number_of_bytes)
-			*total_number_of_bytes = wapi_total_number_of_bytes.QuadPart;
-		if (total_number_of_free_bytes)
-			*total_number_of_free_bytes = wapi_total_number_of_free_bytes.QuadPart;
-	}
+
+	if (free_bytes_avail)
```
luhenry: No need to check for `NULL`. You can even more simply assert on it, like `g_assert (free_bytes_avail);`.

mono/metadata/w32file-win32.c
```diff
@@ -347,35 +347,38 @@ mono_w32file_create_pipe (gpointer *readpipe, gpointer *writepipe, guint32 size)
...
+		*free_bytes_avail = wapi_free_bytes_avail.QuadPart;
+
+	if (total_number_of_bytes)
```
luhenry: No need to check for `NULL`. You can even more simply assert on it.

mono/metadata/w32file-win32.c
```diff
@@ -347,35 +347,38 @@ mono_w32file_create_pipe (gpointer *readpipe, gpointer *writepipe, guint32 size)
...
+		*total_number_of_bytes = wapi_total_number_of_bytes.QuadPart;
+
+	if (total_number_of_free_bytes)
```
luhenry: No need to check for `NULL`. You can even more simply assert on it.

mcs/class/corlib/System.IO/DriveInfo.cs
```diff
@@ -169,14 +169,39 @@ public override string ToString ()
 		}
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		extern static bool GetDiskFreeSpaceInternal (string pathName, out ulong freeBytesAvail,
+		unsafe extern internal static bool GetDiskFreeSpaceInternal (char *pathName, int pathName_length, out ulong freeBytesAvail,
```
luhenry: There is no need for that to be `internal`.

mcs/class/corlib/System.IO/DriveInfo.cs
```diff
@@ -169,14 +169,39 @@ public override string ToString ()
...
 							     out ulong totalNumberOfBytes, out ulong totalNumberOfFreeBytes,
 							     out MonoIOError error);
 
+		unsafe static bool GetDiskFreeSpaceInternal (string pathName, out ulong freeBytesAvail,
```
luhenry: Why isn't that one `internal` anymore? If it doesn't need to be then that's actually better, otherwise you are just breaking the API.

mcs/class/corlib/System.IO/DriveInfo.cs
```diff
@@ -169,14 +169,39 @@ public override string ToString ()
...
+						     out ulong totalNumberOfBytes, out ulong totalNumberOfFreeBytes,
+						     out MonoIOError error)
+		{
+			fixed (char *fixed_pathName = pathName)
```
luhenry: Use proper `{}` around the `fixed` content.

mono/eglib/gstr.c
```diff
@@ -1056,3 +1056,18 @@ g_strnfill (gsize length, gchar fill_char)
...
 }
+
+size_t
+g_utf16_len (const gunichar2 *a)
+{
+#ifdef G_OS_WIN32
+	return wcslen (a);
+#else
+	size_t length = 0;
+	while (*a) {
```
luhenry: `while (a [length]) length++;`

mcs/class/corlib/System.IO/DriveInfo.cs
jaykrell: I didn’t intend to change existing access. Will double check.

mono/metadata/icall.c
jaykrell: Following nearby patterns but ok.

mono/metadata/w32file-win32.c
jaykrell: Nulls checks preexisting but ok.

mcs/class/corlib/System.IO/DriveInfo.cs
jaykrell: I thought folllowing precedent but ok.

mcs/class/corlib/System.IO/DriveInfo.cs
jaykrell: There is no access change here, just adding unsafe.

jaykrell:

mcs/class/corlib/System.IO/DriveInfo.cs
jaykrell: Because the default is the more restrictive private? Agreed.

mono/metadata/icall.c
jaykrell: I removed all the related null checks then and added asserts.

mono/metadata/icall.c
jaykrell: I removed all the related null checks then and added asserts.

mono/metadata/w32file-unix.c
jaykrell: Inconsistent style restored.

mcs/class/corlib/System.IO/DriveInfo.cs
jaykrell: I did another pass on the access and I think they are all correct -- no change  in access to existing signatures, and new signatures are all defaulted to private.

Also, if existing icall had not "Internal" in its name, I added that, when changing the signature from string to const char*.
But no name/access change to existing signatures.

I'm a little nervous about the resulting name overloading and icall-table.c but suspect it is ok.
i.e. where before we might have had one `[icall]FooInternal(string)`, just one `FooInternal` we now have two `[icall]FooInternal(const char*)` and `[not-icall]FooInternal(string)`.

If that is a problem, obvious two suggestions are record signature in icall-def.h as we sometimes do, or further rename like to `FooNative`.

mono/metadata/w32file-win32.c
jaykrell: Removed new/existing null checks and added g_asserts. Note that g_assert is larger than a null check. But also fails a bit more gracefully.

mono/metadata/w32file-unix.c
jaykrell: Inconsistent style restored, everywhere, I think.

mono/metadata/icall.c
jaykrell: Generally removed new/existing null checks and added g_assert.

mono/metadata/icall-def.h
```diff
@@ -366,9 +366,9 @@ ICALL(REGINF_2, "construct_internal_region_from_name", ves_icall_System_Globaliz
 
 #ifndef PLATFORM_NO_DRIVEINFO
 ICALL_TYPE(IODRIVEINFO, "System.IO.DriveInfo", IODRIVEINFO_1)
-ICALL(IODRIVEINFO_1, "GetDiskFreeSpaceInternal", ves_icall_System_IO_DriveInfo_GetDiskFreeSpace)
-ICALL(IODRIVEINFO_2, "GetDriveFormat", ves_icall_System_IO_DriveInfo_GetDriveFormat)
-ICALL(IODRIVEINFO_3, "GetDriveTypeInternal", ves_icall_System_IO_DriveInfo_GetDriveType)
+NOHANDLES(ICALL(IODRIVEINFO_1, "GetDiskFreeSpaceInternal", ves_icall_System_IO_DriveInfo_GetDiskFreeSpace))
+HANDLES(IODRIVEINFO_2, "GetDriveFormatInternal", ves_icall_System_IO_DriveInfo_GetDriveFormat, MonoString, 2, (const_gunichar2_ptr, int))
```
jaykrell: Name change here not strictly needed.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build  failed

jaykrell:
@monojenkins build

PR #11554, opened 2018-11-05
[mini] do not emit safepoints for methods without calls and loops

lewurm:
with this optimization, running Roslyn with mono on
`mono/tests/pinvoke11.cs` compiles:

* 2019 methods without safepoints, and
* 9443 methods with safepoints.


lewurm:

lewurm:
@monojenkins squash

PR #11555, opened 2018-11-05
[2018-10] Fix ref return for structs in runtime invoke wrapper.

luhenry:
Use mono_mb_emit_op for CEE_LDOBJ to ensure wrapper data is added for
indirect load of a ref return value as method-to-ir expects this for wrappers.

Backport https://github.com/mono/mono/pull/11514

/cc @joncham


lewurm:
@monojenkins squash

PR #11556, opened 2018-11-05
Don't override default LLVM's opt & llc args via llvmopts-add and llvmllc-add commands

EgorBo:
Currently we can pass custom arguments to LLVM's `opt` and `llc` via:
```
mono --aot=llvm,llvmopts="...",llvmllc="..."
```
but those completely override the default arguments mono uses (some of them are required).
E.g. lets say I need to add `-foo` to both opt and llc, I should do:

```
mono --aot=llvm,llvmllc="**-foo** -march=x86-64 -mattr=sse4.1 -asm-verbose=false 
-disable-gnu-eh-frame -enable-mono-eh-frame -mono-eh-frame-symbol=_mono_aot_p_eh_frame 
-disable-tail-calls -no-x86-call-frame-opt -relocation-model=pic 
-filetype=obj,llvmopts="**-foo** -O2 -disable-tail-calls"
```

So this PR introduces new commands `llvmopts-add` and `llvmllc-add` to be able to simply do:
`mono --aot=llvm,llvmopts-add="-foo",llvmllc-add="-foo"`

PS: let me know if you don't like the `-add` suffix

/cc: @migueldeicaza

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,5 @@ tmpinst-dir.stamp
 msvc/scripts/inputs/
 extensions-config.h
 *.mlpd
+llvm/build/
+llvm/usr/
--- a/man/mono.1
+++ b/man/mono.1
@@ -215,14 +215,14 @@ and
 options. This feature is experimental.
 .TP
 .I llvmopts=[options]
-Use this option to override (or add via `llvmopts-add`) the built-in set of flags passed to the
+Use this option to add more flags to the built-in set of flags passed to the
 LLVM optimizer.   The list of possible flags that can be passed can be
 obtained by calling the bundled 
 .I opt
 program that comes with Mono.
 .TP
 .I llvmllc=[options]
-Use this option to override (or add via `llvmllc-add`) the built-in set of flags passed to the
+Use this option to add more flags to the built-in set of flags passed to the
 LLVM static compiler (llc).   The list of possible flags that can be passed can be
 obtained by calling the bundled 
 .I llc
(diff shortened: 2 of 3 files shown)


EgorBo:
Tested:
`$ mono --aot=llvm,llvmopts-add="-O3",llvmllc-add="-O3"`

Prints:

Executing opt: "opt" -f -O2 -disable-tail-calls -O3 -o "mono_aot_8aZM0g/temp.opt.bc" "mono_aot_8aZM0g/temp.bc"
Executing llc: "llc"  -march=x86-64 -mattr=sse4.1 -asm-verbose=false -disable-gnu-eh-frame -enable-mono-eh-frame -mono-eh-frame-symbol=_mono_aot_p_eh_frame -disable-tail-calls -no-x86-call-frame-opt -relocation-model=pic -filetype=obj -O3 -o "mono_aot_8aZM0g/temp-llvm.o" "mono_aot_8aZM0g/temp.opt.bc"

luhenry:
Why don't we more simply change `llvmopts` and `llvmllc` to always be additive instead of replacing?

EgorBo:
@luhenry we can if it's not a breaking change for users and we don't need to override the default arguments

vargaz:
I don't think we ever need to override the default arguments, most of those are needed to make things work.

luhenry:
@EgorBo we could more simply put it after we declare our parameters, so the user can override what we pass. AFAIU that's how the `opt` and `llc` command line parameters works anyway: later parameters override previous ones.

EgorBo:
@luhenry @vargaz made it additive by default. Do you mind if I add llvm/build and llvm/usr to gitignore? (generated by `--enable-llvm`)

luhenry:
@monojenkins squash

lewurm:
@monojenkins build failed

PR #11562, opened 2018-11-05
[tests] Make sure we pack all resources along with the tests

luhenry:


akoeplinger:
Why is this necessary? All of the resources should be embedded inside of the test assemblies with my recent changes. Did you mean to target this against an older branch?

luhenry:
@akoeplinger have all your changes been back ported to `2018-10` as well? If so, XA still seems to be missing some pieces, even on the `xamarin-android:mono-2018-10` integration branch. What's still necessary on XA's side is https://github.com/xamarin/xamarin-android/blob/30faa85d6da06d1d703c1fb789d60bc678bbfda6/src/mono-runtimes/mono-runtimes.targets#L325-L327.

I can come up with a list of the failing tests if you want, so that might point you to the list of resources we are missing.

PR #11565, opened 2018-11-06
[icalls] remove DISABLE_POLICY_EVIDENCE

lewurm:
Originally it was introduced with https://github.com/mono/mono/commit/db3591bb93dfe913c39ff73b95194effb88c880a . It seems like only the icall definitions were disabled for device runtimes (which aren't shipped there anyway), however the icalls themselves are _included_ in the device runtime. Note that the definitions are available to the AOT compiler.

Not sure what the rationale was behind this. Since the original change is very old, I think it just happened to be like that today. When running with the interpreter we need the icalls definitions at run-time available too (via the optional icall table that is linked in, so no additional cost for the fullAOT runtime due to this change).

Also remove `HANDLES_MAYBE` since it isn't needed anymore.

I think eventually we want to have `--enable-minimal=policy_evidence` that _also_ excludes the actual icalls, but this will need changes on the BCL side as well.


jaykrell:
Should we put `#if` within the bodies of the icalls then?
And yes, that is definitely cleaner either way, thank you -- and hopefully w/o size regression.
The build was somewhat obscurely broken unless I structured just this way, but I didn't try changing ios.mk.

jaykrell:
@lambdageek might be interested too.

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2018-10

lewurm:
@monojenkins backport 2018-08

lewurm:
@monojenkins backport 2018-06

PR #11573, opened 2018-11-06
Update to latest head to pull in System.Xml.dll file name fix in vbc.rsp

slide:
At the request of @akoeplinger, I am submitting this PR to update roslyn-binaries to pull in a fix for vbc.rsp where the case of System.Xml was incorrect. An upstream patch against roslyn was created as well (https://github.com/dotnet/roslyn/pull/29246)


akoeplinger:
@monojenkins build

PR #11547, opened 2018-11-04
[sdks] Exit with non-zero exit code if the simulator test fails on ios.

vargaz:

PR #11575, opened 2018-11-06
[ios sdk] Package resource assemblies into app bundle

akoeplinger:
.NET resource assemblies live in subdirectories named after the culture of the resource. We need to package these assemblies too.

Addresses https://github.com/mono/mono/pull/11547#issuecomment-435688914


akoeplinger:
Passed the `OS X x64 iOS SDK` job, merging to unblock another PR.

PR #11577, opened 2018-11-06
[Reflection] CoreFX import for Module type

MaximLipnin:
Part of #9660.

The changes:
- included corefx version of Module type;
- enabled related xunit tests from corefx System.Reflection and System.Runtime locations;
- enabled PointerTests fixture which some Module tests depend on;
- made Mono implemented Module class partial; removed duplicated code and left only Mono internal stuff;
- updated corlib make-file to process TestModule.il required by some Module tests (but they will be disabled due to runtime crash).


lambdageek:
I wonder if we should move the internal calls, such as
https://github.com/mono/mono/blob/22143c7916d8f0b2095aeed4d1c4bf0c8cbfedc6/mcs/class/corlib/System.Reflection/Module.cs#L238-L257
from Mono's `System.Reflection/Module.cs` to `MonoModule.cs`

Maybe we should make that a cleanup task for the future.

lambdageek:
Or perhaps, given the build errors, it ought to happen now.  And perhaps the private fields that are accessed from the runtime ought to be on MonoModule.  I think all concrete instances are `MonoModule`s?

vargaz:
Will look into that.

MaximLipnin:
https://github.com/mono/mono/pull/11595 needs to be merged to fix a bug with parsing the nunit xml from xunit.

akoeplinger:
@monojenkins build

MaximLipnin:
@monojenkins build failed

mcs/class/corlib/Makefile
```diff
@@ -133,8 +133,13 @@ DEFAULT_REFERENCES =
 
 TEST_LIB_REFS = System.Core System
 
+TestModule.dll: ../../../external/corefx/src/System.Runtime/tests/TestModule/System.Reflection.TestModule.il
```
marek-safar: This need to use profile specific path, e.g. `$(test_lib_dir)/System.Reflection.TestModule.dll:`

mcs/class/corlib/System.Reflection/Module.cs
```diff
@@ -246,199 +134,18 @@ internal virtual Guid GetModuleVersionId ()
...
-			get { return GetCustomAttributesData (); }
-		}
-
 		internal virtual IntPtr GetImpl () {
 			throw CreateNIE ();
 		}
+
+		static Exception CreateNIE ()
+		{
+			return new NotImplementedException ("Derived classes must implement it");
```
marek-safar: Use existing `throw NotImplemented.ByDesign;`

mcs/class/corlib/System.Reflection/Module.cs
```diff
@@ -43,97 +43,17 @@ internal enum ResolveTokenError {
...
-		Module () {
-		}
-
-		public ModuleHandle ModuleHandle {
-			get {
-				return new ModuleHandle (GetImpl ());
-			}
-		}
-
+	public abstract partial class Module : ISerializable, ICustomAttributeProvider {
```
marek-safar: Could be just simple `partial class Module` there rest is copied from corefx part

marek-safar:
The change is ok but there is a strangely consistent mono_threads_suspend_begin_async_suspend issue on macOS only. @lambdageek @luhenry could you check what this is about?

MaximLipnin:
Does it have a chance to be merged?

luhenry:
@monojenkins build failed

lambdageek:
I'm going to spend a couple of hours to try to repro the osx suspend issue - I don't think I've seen this kind of failure before.

lambdageek:
Tried to get osx64 to fail locally a half-dozen times or so.  No luck.  Looks like it passed the most recent run on CI, too.

marek-safar:

MaximLipnin:
Thanks for review!

PR #11580, opened 2018-11-06
[2018-06] [ios] remove DISABLE_POLICY_EVIDENCE

lewurm:
Originally it was introduced with db3591b
It seems like only the icall definitions were disabled for device
runtimes (which aren't shipped there anyway), however the icalls
themselves are _included_ in the device runtime. Not sure what the
rationale was behind this. Since the original change is very old, I
think it just happened to be like that today.  When running with the
interpreter we need the definitions at runtime too (via the optional
icall table that is linked in, so no additional cost for the fullAOT
runtime due to this change).

Backport of https://github.com/mono/mono/pull/11565


lewurm:
@monojenkins backport 2018-08

lewurm:
@monojenkins backport 2018-10

lewurm:
@monojenkins squash

jaykrell:
Should the `#if` move to within the implementations of the calls?

jaykrell:
I suppose the implementation is pretty small anyway, it just calls back out to managed.

lewurm:
@jaykrell quoting from the original PR on master:
> I think eventually we want to have --enable-minimal=policy_evidence that also excludes the actual icalls, but this will need changes on the BCL side as well.

luhenry:
@monojenkins build failed

PR #11583, opened 2018-11-06
[android sdk] Package resource assemblies into app bundle

akoeplinger:
Follow up to https://github.com/mono/mono/pull/11575 for android

### DIFF
--- a/sdks/android/app/src/main/java/org/mono/android/AndroidRunner.java
+++ b/sdks/android/app/src/main/java/org/mono/android/AndroidRunner.java
@@ -15,6 +15,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.FileNotFoundException;
 
 import org.mono.android.AndroidTestRunner.R;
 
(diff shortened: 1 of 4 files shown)


sdks/android/app/src/main/java/org/mono/android/AndroidRunner.java
```diff
@@ -72,8 +73,19 @@ void copyAssetDir (AssetManager am, String path, String outpath) {
 			for (int i = 0; i < res.length; ++i) {
 				String fromFile = path + "/" + res [i];
 				String toFile = outpath + "/" + res [i];
+
+				InputStream fromStream;
```
luhenry: I would rather check for the type of the file, instead of relying on an exception being thrown. Something like `File.isDirectory` [1] and `File.isFile` [1] would be most helpful here.

[1] https://docs.oracle.com/javase/7/docs/api/java/io/File.html

akoeplinger: @luhenry that was my first idea but the only thing we get from the AssetManager is a stream and `File.isDirectory` actually tries to verify if the path exists _on disk_ which isn't the case here:

```
public boolean isDirectory()

Returns:
true if and only if the file denoted by this abstract pathname exists and is a directory; false otherwise
```

PR #11590, opened 2018-11-07
[wasm] Move bundled linker to different location

marek-safar:
Right now it's in same location as wasm implementation assemblies
which means they are used for execution but they are not .NET desktop
compatible.

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -121,12 +121,6 @@ sample.dll: Simple.Dependency.dll sample.cs WebAssembly.Bindings.dll WebAssembly
 
 OPTIONS_CS = $(TOP)/mcs/class/Mono.Options/Mono.Options/Options.cs
 
-$(WASM_BCL_DIR)/monolinker.exe: $(TOP)/mcs/class/lib/wasm_tools/monolinker.exe
-	cp $< $@
-
-$(WASM_BCL_DIR)/Mono.Cecil.dll: Mono.Cecil.dll
-	cp $< $@
-
 Mono.Cecil.dll: $(TOP)/mcs/class/lib/wasm_tools/Mono.Cecil.dll
 	cp $< $@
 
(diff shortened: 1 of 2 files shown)


sdks/wasm/Makefile
```diff
@@ -17,6 +17,7 @@ DRIVER_CONF=release
 EMSCRIPTEN_SDKDIR=$(TOP)/sdks/builds/toolchains/emsdk
 EMCC=source $(TOP)/sdks/builds/toolchains/emsdk/emsdk_env.sh && emcc
 WASM_BCL_DIR=$(TOP)/sdks/out/wasm-bcl/wasm
+WASM_TOOLS_DIR=$(TOP)/sdks/out/wasm-tools
```
luhenry: This clearly isn't going to work: this `$(TOP)/sdks/out/wasm-bcl/wasm` is pretty much the copy/pasting of `mcs/class/lib/wasm`, so simply "creating" a new `$(TOP)/sdks/out/wasm-tools` to which nothing is ever copied or generated is not good.

What is it exactly you are trying to achieve here?

sdks/wasm/Makefile
```diff
@@ -121,12 +122,15 @@ sample.dll: Simple.Dependency.dll sample.cs WebAssembly.Bindings.dll WebAssembly
...
 
-$(WASM_BCL_DIR)/monolinker.exe: $(TOP)/mcs/class/lib/wasm_tools/monolinker.exe
+$(WASM_TOOLS_DIR)/monolinker.exe: $(TOP)/mcs/class/lib/wasm_tools/monolinker.exe | $(WASM_TOOLS_DIR)
 	cp $< $@
 
-$(WASM_BCL_DIR)/Mono.Cecil.dll: Mono.Cecil.dll
+$(WASM_TOOLS_DIR)/Mono.Cecil.dll: Mono.Cecil.dll | $(WASM_TOOLS_DIR)
 	cp $< $@
 
+$(WASM_TOOLS_DIR):
```
luhenry: This is just **not** the way to go, you do _not_ create a folder in `sdks/out` from `sdks/wasm`. How do you make sure the archive contains it otherwise?

Since you **already** package `wasm_tools` profile in `$(TOP)/sdks/out/wasm-bcl/wasm_tools`, you can _way more simply_ define `WASM_TOOLS_DIR=$(TOP)/sdks/out/wasm-bcl/wasm_tools`. That's of course if you care to fit in the existing model.

marek-safar:
@luhenry thanks for the review

PR #11591, opened 2018-11-07
[2018-10] Bump corefx

marek-safar:


marek-safar:
@monojenkins commit apidiff

marek-safar:
@akoeplinger good question, I think not as that should be only fallback path

akoeplinger:
should we backport the type move just in case or is that going to open a different can of worms?

PR #11599, opened 2018-11-07
[sdks] Compile support/{zlib-helper,nl}.c so XI/XM/XA does not have to cherry-pick them

luhenry:


mono/metadata/icall.c
```diff
@@ -7334,6 +7334,55 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
 	return mono_string_new_handle (mono_domain_get (), g_get_tmp_dir (), error);
 }
 
+#if defined(ENABLE_MONODROID) || defined(ENABLE_MONOTOUCH)
+
+gpointer
+CreateZStream (gint32 compress, MonoBoolean gzip, gpointer feeder, gpointer data);
```
lambdageek: `extern "C"`

mono/metadata/Makefile.am
```diff
@@ -90,6 +90,43 @@ if !ENABLE_ILGEN
 ilgen_libraries = libmono-ilgen.la
 endif
 
+if ENABLE_MONOTOUCH
+platform_sources += ../../support/zlib-helper.c
```
lambdageek: we probably need a `platform_c_sources` and some CFLAGS stuff to make sure it's compiled without the `CXX_CFLAGS` when building Mono with `--enable-cxx`  --- `support/` is not going to be ported to C++.

mono/metadata/icall-decl.h
```diff
@@ -179,5 +179,12 @@ ICALL_EXPORT void ves_icall_System_Runtime_RuntimeImports_Memmove (guint8*, guin
 ICALL_EXPORT void ves_icall_System_Runtime_RuntimeImports_Memmove_wbarrier (guint8*, guint8*, guint, MonoType*);
 ICALL_EXPORT void ves_icall_System_Runtime_RuntimeImports_ZeroMemory (guint8*, guint);
 ICALL_EXPORT void ves_icall_System_Runtime_RuntimeImports_ecvt_s(char*, size_t, double, int, int*, int*);
+#if defined(ENABLE_MONODROID) || defined(ENABLE_MONOTOUCH)
```
lambdageek: would be better to have a feature macro, instead of platforms.

mono/metadata/icall.c
```diff
@@ -7334,6 +7334,55 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
...
+ves_icall_System_IO_Compression_DeflateStreamNative_CreateZStream (gint32 compress, MonoBoolean gzip, gpointer feeder, gpointer data)
+{
+	return CreateZStream (compress, gzip, feeder, data);
+}
+
+gint32
+CloseZStream (gpointer stream);
+
+gint32
+ves_icall_System_IO_Compression_DeflateStreamNative_CloseZStream (gpointer stream)
```
lambdageek: Could these all be PInvokes? is there any reason to add icalls?

luhenry: p/invoke on XA/XI is inherently broken because `__Internal` points to `libmonodroid` (and XI's equivalent), not `libmonosgen-2.0`. So `InternalCall` are just the sure way to hit the right target.

mono/metadata/icall-decl.h
luhenry: We are only interested in having the for the `monodroid`/`monotouch{_tv,_watch}` profiles (as per the BCL), so the feature macro is kinda already `ENABLE_MONODROID`/`ENABLE_MONOTOUCH`.

PR #11602, opened 2018-11-07
Fix wasm mono showing print dialog twice on error

### LINKED ISSUE #11596: [wasm] Abort in browser causes print dialog to show
When there is an abort from webassembly in browser it causes 2 browser print dialogs to open.

It is coming from the two lines in [wasm_logger]( https://github.com/mono/mono/blob/master/sdks/wasm/driver.c#L92-L93) from `driver.c`

NattyNarwhal:
Untested. Fixes #11596.


vargaz:
@monojenkins build failed

PR #11611, opened 2018-11-08
[runtime] Add __attribute__(format) to glib string formatting functions, fix -Wformat warnings/errors exposed by the change.

vargaz:


mono/mini/interp/interp.c
```diff
@@ -897,6 +897,8 @@ ves_array_get (InterpFrame *frame, stackval *sp, stackval *retval, MonoMethodSig
 static gpointer
 ves_array_element_address (InterpFrame *frame, MonoClass *required_type, MonoArray *ao, stackval *sp, gboolean needs_typecheck)
 {
+	printf ("D: %p\n", ao);
```
marek-safar: some left over?

mono/mini/aot-compiler.c
```diff
@@ -9039,9 +9039,13 @@ append_mangled_context (GString *str, MonoGenericContext *context)
 	g_free (res);
 }
 
+static int count;
```
rjvdboon: what is the purpose of this count? It seems it is only increased in append_mangled_wrapper, but never used. If it is needed, is there a more descriptive name? (It clashes with a lot of count variables used in other methods)

lambdageek:
We already added `MONO_ATTR_FORMAT_PRINTF`.  Do we need a separate one for GLib? Should we redefine the Mono one in terms of G?

mono/mini/aot-compiler.c
vargaz: Fixed.

vargaz:
mono-compiler.h doesn't include glib, and its included into libgc/ so it would be complicated to change that.

luhenry:
@monojenkins squash

PR #11616, opened 2018-11-08
[wasm] Clean up leaks in the WasmHttpMessageHandler.

kjpou1:
- Most of the leaks present themselves during a Cancel of a streaming request.

### DIFF
--- a/sdks/wasm/WasmHttpMessageHandler.cs
+++ b/sdks/wasm/WasmHttpMessageHandler.cs
@@ -129,18 +129,23 @@ namespace WebAssembly.Net.Http.HttpClient
                 }
 
                 JSObject abortController = null;
+                JSObject signal = null;
+                WasmHttpReadStream wasmHttpReadStream = null;
+
                 CancellationTokenRegistration abortRegistration = default(CancellationTokenRegistration);
                 if (cancellationToken.CanBeCanceled)
                 {
                     abortController = (JSObject)global.Invoke("__mono_wasm_abortcontroller_hook__");
-                    var signal = abortController.GetObjectProperty("signal");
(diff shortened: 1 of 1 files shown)


sdks/wasm/WasmHttpMessageHandler.cs
```diff
@@ -196,6 +201,13 @@ private async Task doFetch(TaskCompletionSource<HttpResponseMessage> tcs, HttpRe
                 }
 
                 tcs.SetResult(httpresponse);
+
+                // Do not remove the following line of code.  The httpresponse is used in the lambda above when parsing the Headers.
+                // if a local is captured (used) by a lambda it becomes heap memory as we translate them into fields on an object.
+                // If we do not null the field out it will not be GC'd
```
Suchiman: I still don't agree on this. Yes, the `httpresponse` will become part of the closure, but the `httpresponse` will be eligible for collection once the lambda goes out of scope (which it should have been, even before we get to this point).
If the `httpresponse` gets kept alive indefinitely, this means we're leaking at least the closure.

kjpou1: Hello Robin.  The closure is what is leaking.  It is being marshalled to JavaScript and being allocated for use on the JS side.  We do not have a way, at this time, to know that JS is finished with the closure so we will need to manually free it.  The local `httpresponse` stays allocated as well.  We will need to add a mechanism to free that marshalled Action.

Suchiman: Ah okay so it's a known limitation :+1:, thanks for the explanation.

kjpou1:
@monojenkins build failed

PR #11620, opened 2018-11-08
[crashing] attempt mono_threads_summarize () after forking

lewurm:
If `mono_threads_summarize ()` hangs for whatever reason, it still gives us the output by lldb/gdb

### DIFF
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -995,43 +995,6 @@ dump_native_stacktrace (const char *signal, void *ctx)
 		gchar *output = NULL;
 		MonoStackHash hashes;
 
-#ifndef DISABLE_CRASH_REPORTING
-		MonoContext mctx;
-		if (ctx) {
-			gboolean leave = FALSE;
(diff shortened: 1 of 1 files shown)


alexanderkyte:
We discussed this on slack before but let me summarize here. 

I tried this before, and it was often the case that lldb would not have started by the time the dump began and they’d race and signals would be caught. 

It looks like that could still happen here. If you can add some tests that it doesn’t happen, this could work. 

A plan that I mentioned in channel which would fix this problem in a more reliable manner is to:

1) Make sure double faults (getting a SIGSEGV while in the crash reporter) results in lldb being invoked. 

2) Make the crash reporter start a thread that sends it a sigsegv if it hasn’t finished by a certain time. Make sure not to send the sigterm to this supervisor thread. 

I think that we can accomplish each with minimal code.

alexanderkyte:
I created https://github.com/mono/mono/pull/11636 in an attempt to ensure that these dumps appear

lewurm:
closing in favor of https://github.com/mono/mono/pull/11636

Issue #7859, opened 2018-03-28
Reference assemblies need to include private struct fields


jkpubsrc:
Hi!

I just accidentally wrote

    Size size = size;

The compiler did not complain: It assigned an empty `Size` structure to `size`. Is this a bug? According to my understanding this should be: `size` is being assigned from an uninitialized `size`.

vargaz:
Please attach a complete testcase.

jkpubsrc:
You're kidding me, right?

```python
using System;
using System.Drawing;
using System.Windows.Forms;

namespace MyTest
{
	class MainClass
	{
		public static void Main()
		{
			Size size = size;
		}
	}
}
```

vargaz:
Can't reproduce this with mono 5.10:

with csc:

bug8.cs(11,16): error CS0165: Use of unassigned local variable 'size'

with mcs:

bug8.cs(11,9): warning CS1717: Assignment made to same variable; did you mean to assign something else?
bug8.cs(11,16): error CS0165: Use of unassigned local variable `size'

Therzok:
I seem to be able to reproduce this from VSMac inside a console app/using msbuild, not csc directly.

Therzok:
[compilerbug.zip](https://github.com/mono/mono/files/1864518/compilerbug.zip)

lambdageek:
Building against `4.7-api/System.Drawing.dll` seems to silence the compiler error.

If I instead build against `4.5/System.Drawing.dll`, I get:
```
mono /Library/Frameworks/Mono.framework/Versions/5.13.0/lib/mono/4.5/cs
c.exe /noconfig /nostdlib+ /platform:x86 /define:DEBUG /reference:/Library/Fram
eworks/Mono.framework/Versions/5.13.0/lib/mono/4.7-api/mscorlib.dll /reference:
/Library/Frameworks/Mono.framework/Versions/5.13.0/lib/mono/4.7-api/System.Core
.dll /reference:/Library/Frameworks/Mono.framework/Versions/5.13.0/lib/mono/4.7
-api/System.dll /reference:/Library/Frameworks/Mono.framework/Versions/5.13.0/l
ib/mono/4.5/System.Drawing.dll  /out:obj/x86/Debug/compilerbug.exe /target:exe 
 Program.cs
Microsoft (R) Visual C# Compiler version 2.7.0.62620 (e873e693)
Copyright (C) Microsoft Corporation. All rights reserved.

[...]

PR #11628, opened 2018-11-09
Bumping reference assemblies generated with latest GenAPI

henricm:
This PR bumps the reference assemblies submodule to bring in private dummy fields in structs and should fix  mono/mono#7859.


akoeplinger:
@monojenkins backport 2018-10

PR #11629, opened 2018-11-09
Updating GenAPI binaries

henricm:
Bumping api-snapshot submodule which brings in a newly
built GenAPI binaries along with a rename of the exe
to Microsoft.DotNet.GenAPI.exe


akoeplinger:
@monojenkins commit apidiff

PR #11630, opened 2018-11-09
Fix starting process with empty environment incorrectly inherits all variables

### LINKED ISSUE #8766: Starting process with empty ProcessStartInfo.Environment incorrectly inherits all variables
## Steps to Reproduce

Simple `mcs <file.cs>; mono <file.exe>`.

For macOS/Linux:
```csharp
class Program
{
    {
        var process = new System.Diagnostics.Process();
        process.StartInfo.FileName = "env";
        process.StartInfo.UseShellExecute = false;

        process.StartInfo.Environment.Clear();
        //process.StartInfo.Environment["DUMMY"] = "";

        process.Start();
    }
}
```

Windows:
```csharp
class Program
{
    {
        var process = new System.Diagnostics.Process();
        process.StartInfo.FileName = "cmd";
        process.StartInfo.Arguments = "/c set";
        process.StartInfo.UseShellExecute = false;

        process.StartInfo.Environment.Clear();
        //process.StartInfo.Environment["DUMMY"] = "";

        process.Start();
    }
}
```

### Current Behavior

On Mono 5.x `ProcessStartInfo.Environment.Clear()` correctly clears the collection, but calling `Process.Start()` incorrectly inherits the parent process's environment variables. Running the above unmodified code on macOS with Mono 5.x incorrectly outputs all environment variables, if you uncomment the line setting a dummy environment variable the program will output a single environment variable.

As expected the `ProcessStartInfo.EnvironmentVariables` property also exhibits the same behaviour.

Sidenote: .NET Framework on Windows doesn't completely clear the environment variables (at least in the above example), but it does not inherit everything, it is likely that `cmd` is just adding the 3 environment variables below before running `set`, we could create an `env` clone for Windows but there really isn't any need:

```
COMSPEC=C:\Windows\SysWOW64\cmd.exe
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC
PROMPT=$P$G
```

### Expected Behavior

On Mono versions prior to 5.x (and on .NET Framework 4.6.1 on Windows) clearing the collection clears the collection, and starting the process doesn't inherit the parent's environment variables, the example program unmodified above on macOS should output nothing.

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

- Mono 4.6.2.7 doesn't have the problem
- Mono 5.4.1.6 has the problem

## Possible Fix

The problem appears to be caused by `procInfo.envVariables` being left `null` in `Process.StartWithCreateProcess` when there are no environment variables as `envVariables` is lazily created rather than being created immediately inside `if (startInfo.HaveEnvVars)`. This was broken by the changes in 31f90b9db62f84357c97caa4ccb3327dc116c754.

jonorossi:
This pull request fixes starting a process with an empty environment variable collection in `ProcessStartInfo`. As per the issue it fixes a regression that it used to work but currently an empty environment variable collection works like it was never touched.

I also made a minor fix to remove the `StringBuilder` because a simple `string.Concat` does the job without the `StringBuilder` allocation and overhead.

I've included unit tests. Please let me know if you'd like any changes.

Fixes #8766

### DIFF
--- a/mcs/class/System/System.Diagnostics/Process.cs
+++ b/mcs/class/System/System.Diagnostics/Process.cs
@@ -691,22 +691,15 @@ namespace System.Diagnostics
 
 			if (startInfo.HaveEnvVars) {
 				List<string> envVariables = new List<string> ();
-				StringBuilder sb = null;
 
 				foreach (DictionaryEntry de in startInfo.EnvironmentVariables) {
 					if (de.Value == null)
 						continue;
 
-					if (sb == null)
-						sb = new StringBuilder ();
-					else
-						sb.Clear ();
-
-					sb.Append ((string) de.Key);
-					sb.Append ('=');
-					sb.Append ((string) de.Value);
-
-					envVariables.Add (sb.ToString ());
+					envVariables.Add (string.Concat (
+						(string) de.Key,
+						"=",
+						(string) de.Value));
 				}
 
 				procInfo.envVariables = envVariables.ToArray ();
--- a/mcs/class/System/System.Diagnostics/Process.cs
+++ b/mcs/class/System/System.Diagnostics/Process.cs
@@ -690,15 +690,12 @@ namespace System.Diagnostics
 			var procInfo = new ProcInfo ();
 
 			if (startInfo.HaveEnvVars) {
-				List<string> envVariables = null;
+				List<string> envVariables = new List<string> ();
 				StringBuilder sb = null;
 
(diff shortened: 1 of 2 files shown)


luhenry:
@monojenkins squash

luhenry:
@monojenkins build

luhenry:
Looks great, thank you!

marek-safar:
@monojenkins backport 2018-10

luhenry:

PR #11636, opened 2018-11-10
[crash] Force lldb/gdb traces to appear during hangs

alexanderkyte:
Going to put this through some tests offline with sleeps introduced to make sure the mechanism works.


alexanderkyte:
This should accomplish the goal stated here: https://github.com/mono/mono/pull/11620#pullrequestreview-173650084 but should retain the stability of the current dump ordering.

mono/metadata/threads.c
```diff
@@ -6118,6 +6123,69 @@ typedef struct {
...
+{
+	sleep (TIMEOUT_CRASH_REPORTER_FATAL);
+
+	// If we haven't been SIGKILL'ed yet, we signal our parent
+	// and then exit
+#ifdef HAVE_PTHREAD_KILL
+	MOSTLY_ASYNC_SAFE_PRINTF("Crash Reporter has timed out, sending SIGSEGV\n");
+	kill (state->pid, SIGSEGV);
+#else
+	g_error ("pthread_kill () is not supported by this platform");
```
lewurm: `s/pthread_kill/kill/`

mono/metadata/threads.c
```diff
@@ -6118,6 +6123,69 @@ typedef struct {
...
+	if (pid != 0)
+		state->supervisor_pid = pid;
+
+	return pid;
+}
+
+static void
+summarizer_supervisor_end (SummarizerSupervisorState *state)
+{
+	kill (state->supervisor_pid, SIGKILL);
```
lewurm: guard it with `HAVE_KILL`

mono/metadata/threads.c
```diff
@@ -6118,6 +6123,69 @@ typedef struct {
...
+
+	// Accessed on same thread that sets it.
+	int status;
+	waitpid (state->supervisor_pid, &status, 0);
```
lewurm: is `waitpid(2)` available on all platforms? do we need some `HAVE_*` guard?

mono/metadata/threads.c
```diff
@@ -6118,6 +6123,69 @@ typedef struct {
 	gboolean silent; // print to stdout
 } SummarizerGlobalState;
 
+typedef struct {
+	MonoSemType supervisor;
+	gint32 done;
```
lewurm: not used?

alexanderkyte:
Making a note here: VS4M wants the supervisor thread to produce a file, so it can detect when the crash reporter has failed. 

@slluis 
My plan was to make it touch a file that is named either by an environment variable or is in the working directory. 

Anything more detailed about the desired contents or is it sufficient just to create an empty file? I could write a timestamp or something. That seems useful for disambiguation maybe.

lewurm:
@alexanderkyte could you please address my comments, so we can get this puppy merged?

mono/metadata/threads.c
```diff
@@ -6118,6 +6123,69 @@ typedef struct {
...
+	state->pid = getpid();
+
+	/*
+	* glibc fork acquires some locks, so if the crash happened inside malloc/free,
+	* it will deadlock. Call the syscall directly instead.
+	*/
+#if defined(HOST_ANDROID)
+	/* SYS_fork is defined to be __NR_fork which is not defined in some ndk versions */
+	g_assert_not_reached ();
+#elif !defined(HOST_DARWIN) && defined(SYS_fork)
```
vargaz: Shouldn't this return 0 on unsupported arches instead of asserting which produces another fault ?

alexanderkyte: I'm going to just make sure this code isn't defined in this case.

alexanderkyte: This assertion isn't compiled in now, but I want to leave this in here because it explains why the big #if right above it says what it says.

mono/metadata/threads.c
alexanderkyte: Added

mono/metadata/threads.c
alexanderkyte: Done

alexanderkyte: Done

mono/metadata/threads.c
alexanderkyte: Done

alexanderkyte:
Looks like I broke it, will fix.

alexanderkyte:
Ah, it needs the configure.ac check for waitpid too

alexanderkyte:

alexanderkyte:

Issue #8711, opened 2018-05-14
Running into an assertion: mono/utils/memfuncs.c:152, condition `unaligned_bytes (src) == 0' not met


danroth27:
_From @discostu105 on May 9, 2018 0:7_

I have a little prototype where I try to disassemble .NET assemblies in the browser with Mono.

The code is here: https://github.com/discostu105/bilspy

However; I am running into a Mono assertation. Not sure if this is a blazor / mono / webassembly bug, or my own.

Error: 
```
WASM: * Assertion at /Users/builder/jenkins/workspace/test-mono-mainline-webassembly/label/highsierra/mono/utils/memfuncs.c:152, 
condition `unaligned_bytes (src) == 0' not met MonoPlatform.ts:191 
WASM: mono.js:1 Uncaught abort(). Build with -s ASSERTIONS=1 for more info.`
```
Screenshot of the problem (from Chrome devtools): https://twitter.com/discostu105/status/992663078981525504

How to reproduce:
 * Build and Run https://github.com/discostu105/bilspy (it's blazor 0.3)
 * "Choose File": `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\Accessibility.dll` (it does not really matter which assembly exactly, but there are a couple of different errors for different assemblies).
 * In F12 dev tools you should see the error mentioned.

Again, I am not totally sure this isn't my error, but I am kind of stuck with that kind of error.

_Copied from original issue: aspnet/Blazor#800_

danroth27:
_From @SteveSandersonMS on May 11, 2018 22:26_

This is an interesting looking project! I hope it's successful.

Since the code is blowing up at a low level inside Mono, would you be able to report your repro steps to https://github.com/mono/mono/? They'll need to know exactly which third-party library you're calling, and which particular calls trigger the Mono error. Be sure to note that it's a WebAssembly-specific problem.

kumpera:
@vargaz

vargaz:
I can reproduce it.

Entroper:
I don't know if it's the same or similar problem, but I found this issue while googling my own error:

`WASM: * Assertion at /mnt/jenkins/workspace/test-mono-mainline-wasm/label/ubuntu-1804-amd64/mono/mini/wasm_m2n_invoke.g.h:547, condition `0' not met`

roozbehid:
So I get here by searching but I am having similar problem with cannot handle cookie ILI. I think it is because I want to send a long (enum) from cs code into c code.

[...]

PR #11646, opened 2018-11-12
[wasm] add ILI signature to managed2native transitions

lewurm:
Mentioned here https://github.com/mono/mono/issues/8711#issuecomment-437764245


lewurm:
@monojenkins squash

lewurm:
@monojenkins build failed

Issue #9099, opened 2018-06-11
Helix Queues for building and testing


marek-safar:
Due to growing number of tests and scaling reasons, we want to experiment with switching to Helix as the backend execution system for our CI system.

For getting started about sending Helix Jobs, see https://github.com/dotnet/core-eng/blob/master/Documentation/Tools/Sample-Job-Sender/Helix-SDK-Job-Sender.md

### Major motivation factors
- Faster execution due to better scaling
- Easier parallelism
- Less maintenance on our side
- Large pool of bots in the future

# Milestone I
- [x] Port one of our Azure base lanes (~both build and~ tests)
  - [x] Add Makefile targets to package mono runtime + BCL + test assemblies into a .zip file which can be uploaded to Helix
  - [x] Integrate into Jenkins so we trigger a Helix test run after the build
  - [x] Enable xunit xml output for nunitlite (either directly or via separate converter tool)
  - [x] Clean up the BCL test assemblies to not rely on having access to the source tree (e.g. move dependencies into embedded resources that are extracted to some temp file during the test)
  - [x] Add the runtime tests to the .zip file testing setup
  - [x] Add other one-off tests (check-seq-points, profiler, mcs, mdoc, linker) to the .zip file testing setup
- [x] Mission Control integration for the tests results #7039
	- [x] Convert runtime tests to run test-runner with configuration options managed by the runner instead of Makefile
 
# Milestone II
- [ ] Convert acceptance tests into separate queues
- [ ] Update all Azure lanes
- [ ] Switch Mac lanes
- [ ] Switch ARM lanes

# Known issues

- [ ] When submitting to the "offical" MC/Helix queues the test results won't be publicly available, you need to be in the Microsoft org to see it.
- [ ] MC only supports build numbers in float format (e.g. `1.2345` but not `1.2.3.4`) for official builds because they use it in the UI to sort/compare those numbers
- [ ] No way to abort a currently executing Helix test run, this could make the PR auto-abort on updates less useful.

# List of PRs

- [x] https://github.com/mono/NUnitLite/pull/14 and https://github.com/mono/mono/pull/10788
- [x] https://github.com/mono/mono/pull/10777
- [x] https://github.com/mono/mono/pull/10787
- [x] https://github.com/mono/mono/pull/10789
- [x] https://github.com/mono/mono/pull/10862
- [x] https://github.com/mono/mono/pull/10912
- [x] https://github.com/mono/mono/pull/10928
[...]

PR #11647, opened 2018-11-12
Add tools/scripts for running tests in Helix environment

akoeplinger:
- Adding a new client tool for sending jobs which uses the Helix SDK to create jobs in the Helix environment.
- Add targets to bundle tests and add a script to run them in Helix
- Workaround an issue with System.Web tests in the Helix environment due to FileSystemWatcher

https://github.com/mono/mono/issues/9099

### DIFF
--- a/mcs/tools/mono-helix-client/HelixBase.cs
+++ b/mcs/tools/mono-helix-client/HelixBase.cs
@@ -1,0 +1,155 @@
+//
+// HelixBase.cs
+//
+// Authors:
+//	Alexander Köplinger  <alkpli@microsoft.com>
+//
+// Copyright (C) 2018 Microsoft
+//
+// Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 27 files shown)


mcs/tools/mono-helix-client/HelixBase.cs
```diff
@@ -0,0 +1,155 @@
...
+        bool success = false;
+        bool printedMcLink = false;
+
+        Console.WriteLine ($"Waiting for job '{correlationId}' to finish...");
+
+        int sleepTime = 0;
+
+        while (true)
+        {
+            await Task.Delay (sleepTime);
```
baulig:
```suggestion
            await Task.Delay (sleepTime).ConfigureAwait (false);
```

akoeplinger: This is only used in a console app so there's no synchronisation context and so the ConfigureAwait isn't needed. :)

Issue #8378, opened 2018-04-20
WebAssembly: Debugger support


kumpera:
This issue tracks the pending work to get the debugger to work:

MVP:
- [x] Set and Hit Breakpoints
- [x] Handle page reload
- [x] Fix/workaround asp.net crashes/races when WASM payload is parsed
- [x] Unify debugging engine with sdb
- [x] Single stepping
- [x] Build MonoProxy as a library and ship it on mono-wasm.zip
- [x] Move proxy code to mono repo
- [x] BP manipulation (enable/disable/remove)
- [x] C# Async stepping
- [x] Object inspection (this variable, fields, properties, etc)
- [ ] objectGroup support (debugger object lifecycle, so long sessions are possible)
- [x] automated testing
- [x] Release + no debugger build of mono
- [x] Error handling (invalid BP request)

V2:
- [ ] Firefox and Edge support
- [x] js async debug integration
- [ ] watches
- [x] console eval
- [ ] high performance json forwarding

kumpera:
With https://github.com/mono/mono/pull/9339 we got bp, stepping and unification done.

w8w8w8:
could you find a way to resolve this problem?
https://github.com/mono/mono/issues/9350

kumpera:
Issues:
https://github.com/mono/mono/issues/9405
https://github.com/mono/mono/issues/9406

kumpera:
Quality of life things:

- Copy the discover feature from the Blazor prototype. It would simplify our OOTB debugger story by removing the URL port rewrite nonsense.
- Validate the downloaded assemblies against what's loaded in the webapp. I frequently forget to reload it after rebuilding locally. 🤦‍♂️

kumpera:
With #9626 we get the proxy code in the mono repo and the proxy lib shipping with mono-wasm.zip

kumpera:
Missing from that list: object inspection of maps and arrays

kumpera:
Object/frame inspection has 3 arguments we currently ignore: ownProperties, accessorPropertiesOnly, generatePreview

kumpera:
When describing fields, we need to handle boxed objects.

DierkDroth:
I'm trying to understand where we're at regarding WASM support for Blazor and would have thought that this very feature would be important for Blazor on WASM. Given that Blazor is supposed to have WASM support by May 2020 I find it concerning that there hasn't been much progress recently on this particular issue. 

Could anyone please shed some light? What am I missing?

[...]

PR #11649, opened 2018-11-12
[wasm] Add object inspection to the debugger.

kumpera:
https://github.com/mono/mono/issues/8378


thaystg:
I haven't seen this PR before and I did the same thing about object inspection.
Thanks @kumpera, I will use it to inspire myself to write the debugger tests :).

Issue #8430, opened 2018-04-24
Fails to compile without AOT (enable-minimal=profiler,debug,aot)


hmargies:
Mono fails to compile when AOT is disabled (enable-minimal=profiler,debug,aot).

## Steps to Reproduce

```
sh autogen.sh
./configure --enable-minimal=profiler,debug,aot
make
```

## Current Behavior

The function ```mono_aot_get_weak_field_indexes``` is used in mini-runtime.c and implemented in ```aot-runtime.c``` but not compiled in due to ```DISABLE_AOT``` being defined.

```
  CC       mono_boehm-main.o
  CCLD     mono-boehm
./.libs/libmini.a(libmini_la-mini-runtime.o): In function `mini_init':
/home/henry/Development/test/mono/mono/mini/mini-runtime.c:4041: undefined reference to `mono_aot_get_weak_field_indexes'
collect2: error: ld returned 1 exit status
make[4]: *** [mono-boehm] Error 1
make[4]: Leaving directory `/home/henry/Development/test/mono/mono/mini'
make[3]: *** [all] Error 2
make[3]: Leaving directory `/home/henry/Development/test/mono/mono/mini'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `/home/henry/Development/test/mono/mono'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/home/henry/Development/test/mono'
make: *** [all] Error 2
```

## On which platforms did you notice this

[ ] macOS
[X ] Linux
[ ] Windows

**Version Used**: 5.10.1.47 and also master from GitHub

mikaelj:
Any progress?

drgroot:
Seem to have the same problem with version 5.12.0.226 on Linux.
Reopen?

lewurm:
it wasn't backported to the release branches.

@marek-safar @luhenry is it okay to backport it to 2018-02 still? https://github.com/mono/mono/pull/9394

jbatt33:
Not sure if this is fixed in 5.16 release against ARM builds.  x86 appears to build fine with AOT disabled, but receiving the following error from the ARM build:

"
mono_aot_get_ftnptr_arg_trampoline
"

```

/libmonogc.a/stubborn.o .libs/libmonoboehm-2.0.lax/libmonogc.a/typd_mlc.o .libs/libmonoboehm-2.0.lax/libmonogc.a/win32_threads.o
[...]

PR #11653, opened 2018-11-12
[arm] fix --enable-minimal=aot build

lewurm:
mentioned here: https://github.com/mono/mono/issues/8430#issuecomment-437041383


mikaelj:
Which release is this in? It's not in 5.20.1.19.

lewurm:
It will be in 6.0

PR #11659, opened 2018-11-13
[wasm] add IIIIIIIIIIIII signature to managed2native transitions

roozbehid:
For my project I needed this 12 parameters calling and I also added 11, 10 and 9 parameters calling.

It might sound strange but I was really calling somebody else library which had 12 parameters calling. It is also not that strange if you look at CreateWindowEx in Win32!

I did count all those i's probably four time to make sure I did not make a mistake :D


lewurm:
@monojenkins build

lewurm:
@monojenkins build

lewurm:
Thank you for your contribution @roozbehid 👍

PR #11660, opened 2018-11-13
[wasm][binding] Allow the cleanup of transparent/bridged marshalled objects causing leaks.

kjpou1:
- Example: When a delegate is marshalled to JS it is allocated so that it is not GC'd.  Since we do not know when JS is finished with the delegate we need a way, for right now, to tell the Runtime to de-allocate the object and remove the transparent/bridged instance from JS as well.  The side-effect of not doing this is that any local that is captured (used) by a lambda became heap memory and were allocated as fields on the object.  Since the delegate is never a candidate for GC the local that was captured was never released either.

- Fix an error with Windows Edge and Windows IE that threw an exception `Object doesn't support this action` on certain host objects.  This only affected Windows Edge and IE when using the JavaScript delete operator to clean up an object after being disposed.

- Cleanup marshalled Tasks after completion.  The marshalled tasks and their continuation objects were never un-allocated thus not allowing them to be GC'd.

### DIFF
--- a/sdks/wasm/WasmHttpMessageHandler.cs
+++ b/sdks/wasm/WasmHttpMessageHandler.cs
@@ -203,13 +203,19 @@ namespace WebAssembly.Net.Http.HttpClient
                         ((JSObject)other).Dispose();
                     });
 
-                    respHeaders.Invoke("forEach", foreachAction);
+                    try
+                    {
 
-                    // Do not remove the following line of code.  The httpresponse is used in the lambda above when parsing the Headers.
-                    // if a local is captured (used) by a lambda it becomes heap memory as we translate them into fields on an object.
-                    // The foreachAction is allocated when marshalled to JavaScript.  Since we do not know when JS is finished with the
-                    // Action we need to tell the Runtime to de-allocate the object and remove the instance from JS as well.
-                    WebAssembly.Runtime.FreeObject(foreachAction);
+                        respHeaders.Invoke("forEach", foreachAction);
+                    }
+                    finally
+                    {
+                        // Do not remove the following line of code.  The httpresponse is used in the lambda above when parsing the Headers.
+                        // if a local is captured (used) by a lambda it becomes heap memory as we translate them into fields on an object.
+                        // The foreachAction is allocated when marshalled to JavaScript.  Since we do not know when JS is finished with the
+                        // Action we need to tell the Runtime to de-allocate the object and remove the instance from JS as well.
+                        WebAssembly.Runtime.FreeObject(foreachAction);
+                    }
 
                 }
 
--- a/sdks/wasm/WasmHttpMessageHandler.cs
+++ b/sdks/wasm/WasmHttpMessageHandler.cs
@@ -193,7 +193,7 @@ namespace WebAssembly.Net.Http.HttpClient
                 {
                     // Here we invoke the forEach on the headers object
                     // Note: the Action takes 3 objects and not two.  The other seems to be the Header object.
-                    respHeaders.Invoke("forEach", new Action<object, object, object>((value, name, other) =>
+                    var foreachAction = new Action<object, object, object>((value, name, other) =>
                     {
 
                         if (!httpresponse.Headers.TryAddWithoutValidation((string)name, (string)value))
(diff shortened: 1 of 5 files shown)


sdks/wasm/bindings.cs
```diff
@@ -119,20 +124,29 @@ static int UnBindRawJSObjectAndFree(int gcHandle)
...
                 }
 
-
                 obj.Dispose();
 
                 var gCHandle = obj.Handle;
                 obj.Handle.Free();
                 obj.JSHandle = -1;
                 obj.RawObject = null;
+                obj = null;
```
kg: Why are these null assignments before return needed? Are they working around a runtime bug?

sdks/wasm/binding_support.js
```diff
@@ -693,8 +695,27 @@ var BindingSupportLib = {
 				var gc_handle = obj.__mono_gchandle__;
 				if (typeof gc_handle  !== "undefined") {
 					this.wasm_unbind_js_obj_and_free(js_id);
-					delete obj.__mono_gchandle__;
-					delete obj.__mono_jshandle__;
+
+					// Windows Edge and Windows IE throws an exception `Object doesn't support this action` on
+					// certain host objects.
+					try {
+						delete obj.__mono_gchandle__;
```
kg: given the need for an exception handler and that we can just do ```x.__mono_gchandle__ !== undefined``` to cover all cases, should we even bother using ```delete```? It deoptimizes code in some runtimes as well.

sdks/wasm/bindings.cs
kjpou1: Just wanted to make sure things were clean.

sdks/wasm/binding_support.js
kjpou1: So we just set `obj.__mono_gchandle__ = undefined` instead of trying to delete?

kjpou1: Did not know it deoptimized code in some runtimes.  Good to know.

kg: Yeah there are specific primitives (the delete statement is one of them) that make runtimes unhappy because they can change object shapes and stuff. It gets nasty since key/value pairs on JS objects are ordered, so if you do ```var temp = a.x; delete a.x; a.x = temp;``` you've actually mutated the object in an observable way.

try/catch and the 'arguments' object are two other ones that sometimes cause deopts.

In this case I think writing undefined to the fields should be good enough, and it eliminates the need for that special case. delete is only (imo) necessary if you're using the object's keys themselves for some important reason (like a dictionary) instead of just the values.

kjpou1: Will do and thanks for the explanation.

sdks/wasm/sample.cs
```diff
@@ -38,22 +38,13 @@ static void Main(string[] args)
...
                 BaseApiUrl = (string)location.GetObjectProperty("origin");
             }
 
-            window = null;
-
-            // Setup the HttpClientHandler
-            httpMessageHandler.SetValue(null, (Func<HttpMessageHandler>)(() => {
-                return new WasmHttpMessageHandler();
-            }));
+            window.Dispose();
```
marek-safar: Should use using as well

sdks/wasm/bindings.cs
```diff
@@ -591,7 +609,10 @@ public void SetObjectProperty(string expr, object value, bool createIfNotExists
         protected void FreeHandle()
         {
 
-            Runtime.InvokeJS("BINDING.mono_wasm_free_handle(" + JSHandle + ");");
+            int exception = 0;
```
marek-safar: No need to initialize it

sdks/wasm/bindings.cs
```diff
@@ -119,20 +124,26 @@ static int UnBindRawJSObjectAndFree(int gcHandle)
...
                 return (int)(IntPtr)gCHandle;
             }
-
             return 0;
 
         }
 
+        public static void FreeObject (object obj)
+        {
+            int exception = 0;
```
marek-safar: Does not need to be initialized

sdks/wasm/WasmHttpMessageHandler.cs
```diff
@@ -193,25 +193,28 @@ private async Task doFetch(TaskCompletionSource<HttpResponseMessage> tcs, HttpRe
...
-                    ));
+                    });
+
+                    respHeaders.Invoke("forEach", foreachAction);
+
+                    // Do not remove the following line of code.  The httpresponse is used in the lambda above when parsing the Headers.
+                    // if a local is captured (used) by a lambda it becomes heap memory as we translate them into fields on an object.
+                    // The foreachAction is allocated when marshalled to JavaScript.  Since we do not know when JS is finished with the
+                    // Action we need to tell the Runtime to de-allocate the object and remove the instance from JS as well.
+                    WebAssembly.Runtime.FreeObject(foreachAction);
```
marek-safar: This should be in `finally` section

sdks/wasm/sample.cs
kjpou1: That should have been removed altogether but `using` is better.  Will change it.

kjpou1:
@monojenkins build failed

PR #11671, opened 2018-11-13
[System.Data]: GSS is not available on tvOS and watchOS.

baulig:


baulig:
@monojenkins commit csproj

PR #11673, opened 2018-11-14
[configure] Pass -stdlib=libc++ to clang on darwin to build C++ code

lambdageek:
On iOS 6.0 (`-miphoneos-version-min=6.0`), this is needed to get modern C++11
headers like `<type_traits>`.  It does not seem to harm anything to
over-approximate and pass it to clang always when building on Darwin.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1021,6 +1021,13 @@ AC_ARG_ENABLE(cxx, [  --enable-cxx   compile some code as C++])
 CXXFLAGS_COMMON=' -std=gnu++0x -fno-exceptions -fno-rtti '
 # "c++0x" instead of C++11, for compat with Centos6/gcc4.4
 
+# -stdlib=libc++ is needed by clang for iOS 6.0 (-miphoneos-version-min=6.0)
+# to support C++11 headers but it does not seem to harm elsewhere, so over-approximate
+# and add it whenever we're running clang on Darwin.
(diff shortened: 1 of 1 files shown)


lambdageek:
> I assume that doesn’t link in libc++ (i.e. add it as a dependency) correct?

As best I can tell it does not.

Experiment:

```
$ clang -xc++ -stdlib=libc++ -fno-rtti -fno-exceptions -std=gnu++0x -c bar.c
$ clang -stdlib=libc++ -o bar  bar.o
$ otool -L ./bar
./bar:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
$ clang++ -stdlib=libc++ -o bar bar.o
$ otool -L ./bar
./bar:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.4)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
```

So `-stdlib=libc++` just affects include and lib paths, but doesn't imply `-lc++`.

lambdageek:
coop suspend failure in sentry-telemetry-upload step is known. WebAssembly C++ fails elsewhere, too.

akoeplinger:
@lambdageek is there an issue tracking the sentry+coop failure?

The one I looked at was just the normal left-over crash.json files which should be cleaned up now. I've restarted the job.

akoeplinger:
@monojenkins squash

PR #11677, opened 2018-11-14
[2018-10][System.Data]: GSS is not available on tvOS and watchOS.

baulig:
Manual backport of https://github.com/mono/mono/pull/11671.


akoeplinger:
@monojenkins commit csproj

PR #11694, opened 2018-11-15
[wcf] do not assume that message body is always Wrapped

madsjohnsen:
Currently, setting IsWrapped=false on a MessageContract leads to an error in serialization due to an empty element being present where the wrapper element would have been, instead of it being gone entirely.
The serialization then fails because the empty string is not valid localName (which of course is fine).


luhenry:
@monojenkins build

akoeplinger:

PR #11695, opened 2018-11-15
[bcl] Remove unused DISABLE_CAS_USE and AGCLR define

akoeplinger:
Stumbled over it while grepping for #11679, thought I'd just clean it up.


akoeplinger:
@monojenkins commit csproj

PR #11696, opened 2018-11-15
[upload-to-sentry] Skip reading assemblies without mono_crash.json and silence no symbols error

akoeplinger:
The previous code read all the .dll and .exe files even when there actually wasn't any mono_crash.*.json file to process.

In the happy case of no crash file this cuts the time from 60s to 5s.

We can also use more efficient patterns in Directory.GetFiles() instead of getting _all_ files and filtering in-memory.

Lastly we also ignore printing a message when a debug symbols file wasn't found for an assembly, this is an expected case for a lot of our assemblies.
Using InMemory Cecil reading shaves another 20s off.

### DIFF
--- a/mcs/tools/upload-to-sentry/UploadToSentry.cs
+++ b/mcs/tools/upload-to-sentry/UploadToSentry.cs
@@ -381,14 +381,14 @@ namespace UploadToSentry
 			this.codebase = assemblies;
 		}
 
-		static IEnumerable<string> GetAssemblies (string fileRoot) {
+		static string[] GetAssemblies (string fileRoot) {
 			var dlls = Directory.GetFiles (fileRoot, "*.dll", SearchOption.AllDirectories);
 			var exes = Directory.GetFiles (fileRoot, "*.exe", SearchOption.AllDirectories);
 
-			return Enumerable.Concat (dlls, exes);
+			return dlls.Concat (exes).ToArray ();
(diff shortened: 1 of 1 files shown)


mcs/tools/upload-to-sentry/UploadToSentry.cs
```diff
@@ -411,12 +414,17 @@ public static void Main (string[] args)
 
 			var dsn = new Dsn(url);
 
+			var files = GetFiles (fileRoot);
+
+			if (files.Count () == 0)
```
marek-safar: This could be expensive operation, try to avoid it

akoeplinger: How? `Directory.GetFiles()` already returns a fully materialized list so I don't see how this could expensive.

Therzok: Sadly, it is not. Array does not implement `ICollection`/`ICollection<T>`, so it does not get fast paths.

akoeplinger: I can switch to `.Empty()` if it makes you feel better :)

marek-safar: Please change the GetFiles to return the string array and use Length here instead

PR #11700, opened 2018-11-15
[crash] Remove duplicated il_offset

alexanderkyte:
This code is duplicated completely from the code directly above it (github doesn't show in this diff)


akoeplinger:

Issue #11403, opened 2018-10-26
error: Thread 0x700005a5a000 in unexpected state STATE_BLOCKING with PEEK_BLOCKING_SUSPEND_REQUESTED


spouliot:
## Steps to Reproduce

1. Checkout and build `mono-2018-08` branch of `xamarin-macios` prior to commit xamarin/xamarin-macios@08813eff2d20fbd669c28eab2402ff2f745d73d5 (later in the branch, hybrid suspend is disabled)
2. Run Xamarin.Mac `xammac` unified tests on Mojave (10.14)

https://jenkins.internalx.com/blue/rest/organizations/jenkins/pipelines/macios/branches/PR-4596/runs/44/nodes/184/steps/516/log/?start=0

### Current Behavior

```
***** InitWithDataTest
2018-10-26 14:08:25.578 xammac_tests[16420:1047220] error: Thread 0x700005a5a000 in unexpected state STATE_BLOCKING with PEEK_BLOCKING_SUSPEND_REQUESTED
...
```

Unless it's related to a recent bump it's likely that it's a random, hard to reproduce failure. I'll update the issue if this happens again on our bots.

### Expected Behavior

No crash

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

mono 2018-08 integration branch w/xamarin-ios

## Stacktrace

```
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) System.GC.InternalCollect (int) [0x00014] in <a221268c5d8f45a695639c16463540f8>:0
  at System.GC.Collect () [0x00000] in /Library/Frameworks/Xamarin.Mac.framework/Versions/5.5.1.181/src/Xamarin.Mac/mcs/class/referencesource/mscorlib/system/gc.cs:248
  at MonoTouchFixtures.Foundation.ThreadTest.InitWithDataTest () [0x00044] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/monotouch-test/Foundation/ThreadTest.cs:68
  at (wrapper runtime-invoke) object.runtime_invoke_void__this__ (object,intptr,intptr,intptr) [0x00020] in <a221268c5d8f45a695639c16463540f8>:0
  at <unknown> <0xffffffff>
[...]

Issue #10722, opened 2018-09-20
Hybrid suspend: Thread 00006484 in unexpected state RUNNING with PEEK_BLOCKING_SUSPEND_REQUESTED


lambdageek:
The hybrid suspend `sgen-stw` code assumes that if a restart was required it's enough to run just the mop-up phase of two-phase suspend.  This is incorrect: a thread could be in a critical region in GC Safe and it would then resume into BLOCKING.  So we need the full two-phase suspend.

## Steps to Reproduce

1.  add `ENABLE_HYBRID_SUSPEND` to `winconfig.h` and implement a dummy `SAVE_REGS_ON_STACK` (until #10684 is fixed)
2.  build an msvc win32 release Mono runtime
3.  run `thread-suspend-suspend.exe`

### Current Behavior

```
Thread 00006484 in unexpected state RUNNING with PEEK_BLOCKING_SUSPEND_REQUESTED
```

### Expected Behavior

Test exits with exit code 0.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[x] Windows

**Version Used**:

master (I used 5bfe7a5)

lambdageek:
This is actually more general than just a windows problem: `mono_thread_info_in_critical_location` in `sgen_unified_suspend_stop_world` could return TRUE for GC Safe suspended threads on other platforms too - for example if the thread's SP is outside the normal stack bounds:
https://github.com/mono/mono/blob/c2113246b90a29113ca536e3cf89c5b0424a94e5/mono/utils/mono-threads.c#L1148-L1150

lambdageek:
The solution is to run both phases of suspend after a restart.

lambdageek:
One alternative might be to have a "pulse" thread state transition that resumes a thread right back into a "suspend requested" state - RUNNING threads into "async_suspend_requested" and onward to a cooperative suspend; BLOCKING threads into "blocking_suspend_requested" and then they will either self suspend at a GC Safe -> GC Unsafe transition, or else we'll preempt them in the next mop-up phase.

We would then only need to restart the whole world if any RUNNING thread got pulsed.

lambdageek:
We implemented 'pulse' transitions (see the PRs above).  Haven't seen similar crashes since then.  Closing.

PR #11708, opened 2018-11-15
[threading] Implement 'pulse' transition for two-phase STW

lambdageek:
This addresses the symptoms (ie invalid thread state transition assertions) from #11403 and #10722, at least for GC Safe threads.

With non-hybrid suspend, a pulse is just a resume.

With hybrid suspend, a `pulse` is a combination of a resume of a preemptively
suspended GC Unsafe thread and a request that it suspend again - as if we
atomically resumed the thread and then applied the initial phase of two-phase
suspend again.  If the now-running thread comes to a GC Safe -> GC Unsafe
transition it will suspend.  If not, we will preemptively suspend it again
during the mop-up phase.

The important thing is that we don't want the thread to transition to GC Unsafe
again: if it does so, it may try to take a lock that some other
async-suspended thread is holding, and it will never reach a safepoint ---
exactly the situation that two-phase suspend is supposed to solve.

----

This doesn't fix the underlying issue in #11403 - that we sometimes see a `MonoThreadInfo*` that relates to a GCD worker thread that isn't running (because the OS reuses a `pthread_t`).  But it does mean that we will resume it without asserting and then try to suspend it again once it's running and has non-zero stack and doesn't look like it's in a critical region anymore.

It's a fix for #10722 assuming that waiting for a resume after a self-suspend doesn't end up in what looks like a WOW64 API call - that would look like a GC Unsafe thread in a critical region which we would attempt to pulse. For now `mono_threads_transition_request_pulse` will assert if it sees a GC Unsafe thread.

### DIFF
--- a/mono/utils/mono-threads.c
+++ b/mono/utils/mono-threads.c
@@ -978,6 +978,25 @@ mono_thread_info_core_resume (MonoThreadInfo *info)
 	return res;
 }
 
+/*
+ *   Current thread must hold the global_suspend_semaphore.
+ *   The given MonoThreadInfo* is a suspended thread.
+ *   Must be using hybrid suspend.
+ */
+static gboolean
+mono_thread_info_core_pulse (MonoThreadInfo *info)
+{
+	gboolean res = FALSE;
+
+	switch (mono_threads_transition_request_pulse (info)) {
+	case PulseInitAsyncPulse:
+		resume_async_suspended (info);
+		res = TRUE;
+		break;
+	}
+	return res;
+}
+
 gboolean
 mono_thread_info_resume (MonoNativeThreadId tid)
 {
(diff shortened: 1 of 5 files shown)


mono/utils/mono-threads.c
```diff
@@ -1117,6 +1136,21 @@ mono_thread_info_begin_resume (MonoThreadInfo *info)
...
+mono_thread_info_begin_pulse_resume_and_request_suspension (MonoThreadInfo *info)
+{
+	/* For two-phase suspend, we want to atomically resume the thread and
+	 * request that it try to cooperatively suspend again.  Specifically,
+	 * we really don't want it to transition from GC Safe to GC Unsafe
+	 * because we then it could (in GC Unsafe) try to take a lock that's
+	 * held by another preemptively-suspended thread, essentially
+	 * recreating the same problem that two-phase suspend intends to
+	 * fix. */
+	if (!mono_threads_is_hybrid_suspension_enabled ())
```
luhenry: Shouldn't this be `if (! mono_threads_is_hybrid_suspension_enabled() && ! mono_threads_is_cooperative_suspension_enabled())`?

lambdageek: It should be a (non-existent) `mono_threads_is_multiphase_stw_enabled()`: we only need to pulse if STW requires two-phase suspension.  If there's only a single phase, resuming is fine.

lambdageek:
Added `mono_threads_is_multiphase_stw_enabled`

luhenry:
@monojenkins build failed

lambdageek:

PR #11724, opened 2018-11-16
[wasm] Add the declaring types assembly name to the mangled name for instances during dedup. Fixes failures when two assemblies have a generic type with the same name.

vargaz:


mono/mini/aot-compiler.c
```diff
@@ -9052,7 +9052,7 @@ append_mangled_method (GString *s, MonoMethod *method)
 		g_assert (imethod->context.class_inst != NULL || imethod->context.method_inst != NULL);
 
 		append_mangled_context (s, &imethod->context);
-		g_string_append_printf (s, "_declared_by_");
+		g_string_append_printf (s, "_declared_by_%s_", imethod->declaring->klass->image->assembly->aname.name);
```
jeromelaban: can there be issues if the assembly name is containing dots or some other characters ?

akoeplinger:
The `Linux x64 Checked Private Types Build` fails with 

```
aot-compiler.c: In function 'append_mangled_method':
aot-compiler.c:9055:75: error: dereferencing pointer to incomplete type 'MonoClass {aka struct _MonoClass}'
   g_string_append_printf (s, "_declared_by_%s_", imethod->declaring->klass->image->assembly->aname.name);
                                                                           ^~
```

vargaz:
@monojenkins build failed

Issue #11684, opened 2018-11-14
Type.GetType doesn't return null for "System.Int32[,*,]"


MaximLipnin:
## Steps to Reproduce

```
using System;

class Program
{
	static void Main()
	{
		Type t = Type.GetType("System.Int32[,*,]", throwOnError: false, ignoreCase: false);
		Console.WriteLine($"t is null : {t == null}");
	}
}
```

### Current Behavior
The output is:
`t is null : False`

### Expected Behavior

The output should be:
`t is null : True`

## On which platforms did you notice this

[X] macOS
[ ] Linux
[ ] Windows

**Version Used**:

5.16

## Stacktrace

```
Please paste the stack trace here if available.
```

lambdageek:
This is a bug in the unmanaged type name parser.  The managed type name parser correctly reports the problem.
https://github.com/mono/mono/blob/1076d746b9caa447981825dfa3117ee7341642fe/mcs/class/corlib/System/TypeSpec.cs#L561-L562
We're missing the corresponding check in reflection.c `_mono_reflection_parse_type`

----

Test case for the managed parser is:
```csharp
using System;

class Program
{
	static void Main()
	{
		string s = "System.Int32[,*,]";
		Func<System.Reflection.AssemblyName,System.Reflection.Assembly> assemblyResolver =
			(asmName) => {
			return AppDomain.CurrentDomain.Load (asmName);
		};
		Func<System.Reflection.Assembly,string,bool,Type> typeResolver =
			(assembly, typeName, ignoreCase) => {
			return assembly.GetType (typeName, ignoreCase);
		};
		Type t = Type.GetType (s, assemblyResolver, typeResolver);
	}
}
```

PR #11725, opened 2018-11-16
[reflection] Parse error if rank > 1 array contains *

lambdageek:
Can either have a multi-dimensional array type or an array with a lower
bound, not both.  (eg `System.Int32[,,]` vs `System.Int32[*]`, but not `System.Int32[*,]`)

Fixes https://github.com/mono/mono/issues/11684


luhenry:
@lambdageek could you please bump to `corefx:master` in the PR as well, now that the test has been re-enabled.

lambdageek:
Test failure is related.  Parsing of `System.String[*]` got broken somehow...

lambdageek:
@monojenkins build failed

PR #11726, opened 2018-11-16
[runtime] Skip upload when tool fails to build

alexanderkyte:


luhenry:
@monojenkins build failed

PR #11729, opened 2018-11-17
Do not emit safepoints if method is marked with NoSafepoints

EgorBo:
This API Proposal adds a kind of "I know what I am doing" attribute to disable safepoint insertion for methods.
Currently we insert them into (bounded) loops even if the optimizer can safely unroll them.

Example:
```csharp
static int Test()
{
    int sum = 0;
    for (int i = 0; i < 5; i++)
        sum += i;
    return sum;
}
```
Generates:
```asm
_ConsoleApp31_Program_Test:
	pushq	%rbx
	movq	2280(%rip), %rbx
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x13>
	callq	189 <plt__jit_icall_mono_threads_state_poll>
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x1e>
	callq	178 <plt__jit_icall_mono_threads_state_poll>
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x29>
	callq	167 <plt__jit_icall_mono_threads_state_poll>
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x34>
	callq	156 <plt__jit_icall_mono_threads_state_poll>
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x3f>
	callq	145 <plt__jit_icall_mono_threads_state_poll>
	cmpq	$0, (%rbx)
	je	5 <_ConsoleApp31_Program_Test+0x4a>
	callq	134 <plt__jit_icall_mono_threads_state_poll>
	movl	$10, %eax
	popq	%rbx
```

`[MethodImpl(MethodImplOptions.NoSafepoints)]` makes it
```asm
_ConsoleApp31_Program_Test:
	movl	$10, %eax
	retq
```

PS: I use 0x400 because of https://github.com/dotnet/corefx/pull/32322
PS2: This particular case can be fixed in runtime if it is able to detect if a loop is bounded or not but it'd be nice to have this flag anyway

PS3: This is what .NET Core 3.0 generates (doesn't unroll the loop):
```asm
Program.Test()
  xor     eax,eax
  xor     edx,edx
M00_L00  
  add     eax,edx
  inc     edx
  cmp     edx,5
  jl      M00_L00
  ret
```


luhenry:
This may be a step in the right direction, but I would largely favor an optimisation at LLVM level, or even at our JIT level. We should explore what LLVM safepoint/statepoint API has to offer us in terms of easing LLVM’s optimization work.

marek-safar:
I agree we should explore this path only if LLVM capabilities are not sufficient to help us here

PR #11739, opened 2018-11-20
[System]: Bring `System.Diagnostics.DiagnosticsSource` from CoreFX.

baulig:


mcs/class/System/common.sources
```diff
@@ -890,3 +891,11 @@ ReferenceSources/Win32Exception.cs
 
 ../../../external/corefx/src/System.Text.RegularExpressions/src/System/Collections/Generic/ValueListBuilder.Pop.cs
 ../../../external/corefx/src/System.Text.RegularExpressions/src/System/Collections/HashtableExtensions.cs
+
+../../../external/corefx/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
```
marek-safar: How many entry points we need from this API? We cannot make it public and we don't care about this on mobile profiles

baulig:
Okay, turns out that we can stub this out and then just disable the tests:
https://github.com/baulig/mono/commit/cf2572229d9c9437610d466844d5ea114584e41c.

PR #11740, opened 2018-11-20
[System]: Bring two internal helper classes from CoreFX.

baulig:


mcs/class/System/common.sources
```diff
@@ -790,8 +790,10 @@ ReferenceSources/Win32Exception.cs
 ../../../external/corefx/src/Common/src/System/IO/DelegatingStream.cs
 ../../../external/corefx/src/Common/src/System/IO/ChunkedMemoryStream.cs
 ../../../external/corefx/src/Common/src/System/IO/PathInternal.CaseSensitivity.cs
+../../../external/corefx/src/Common/src/System/IO/ReadOnlyMemoryStream.cs
 
 ../../../external/corefx/src/Common/src/System/Net/HttpStatusDescription.cs
+../../../external/corefx/src/Common/src/System/Net/HttpVersionInternal.cs
```
marek-safar: Why do we need that when we already have identical `../../../external/corefx/src/System.Net.Primitives/src/System/Net/HttpVersion.cs`

PR #11746, opened 2018-11-20
[bcl] Use the UTF8 encoding on WASM by default, similarly to ios.

vargaz:


akoeplinger:
@monojenkins commit csproj

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

Issue #11479, opened 2018-10-31
Loading assembly generated using tlbimp.exe give error in mono runtime


pieandcakes:
When running code generated using tlbimp.exe in mono, we get an error even though code executes just fine.

Mono runtime version: 5.0.14.177

## Repro
```
using System;
using System.Runtime.CompilerServices;

namespace ConsoleApp15
{
    class Program
    {
        static void Main(string[] args)
        {
            ITest test = (ITest)new Test();
            test.Test();
        }
    }

    interface ITest
    {
        // This attribute is not necessary, but tlbimp.exe produces it
        [MethodImpl(MethodImplOptions.PreserveSig | MethodImplOptions.InternalCall)]
        void Test();
    }

    class Test : ITest
    {
        void ITest.Test()
        {
            Console.WriteLine("Test!");
        }
    }
}
```
## Error

```
cant resolve internal call to "ConsoleApp15.ITest::Test" (tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: /src/andrew/ConsoleApp15.exe

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.
Test!
```

pieandcakes:
@andrewcrawley

pieandcakes:
Rolling back to 5.0.1.1 does not generate this error anymore.

lambdageek:
Can we just ignore `MethodImplOptions.InternalCall` on an abstract method?

I tried an example with a DIM, which I assume is why we have a regression, and it comes out as:
```
.method public virtual hidebysig newslot 
           instance default void M ()  cil managed internalcall 
```
So I think that would work.

(By contrast, a normal interface method is 
```
    .method public virtual hidebysig newslot abstract 
           instance default void N ()  cil managed internalcall 
```
)

PR #11755, opened 2018-11-21
[loader] ignore 'internalcall' impl attribute on 'abstract' methods

lambdageek:
Otherwise we emit a noisy warning when doing a lookup of the method.

Fixes https://github.com/mono/mono/issues/11479


lambdageek:
@vargaz this feels like a hack, but we check for `METHOD_IMPL_ATTRIBUTE_INTERNAL_CALL` all over the place and I wasn't sure how to ignore it consistently on abstract methods otherwise.

marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2018-08

Issue #11765, opened 2018-11-23
watchOS extensions crash at launch with 'Failed to load AOT module 'mscorlib' while running in aot-only mode: not compiled with safepoints.'


rolfbjarne:
## Steps to Reproduce

1. Build & install the mono-2018-08 branch of xamarin-macios.
1. Create new iOS app and add a watchOS App + Extension
2. Run on a watch device.

### Current Behavior

Crashes at launch with:

    Failed to load AOT module 'mscorlib' while running in aot-only mode: not compiled with safepoints.

### Expected Behavior

No crash

**Version Used**:

https://github.com/xamarin/xamarin-macios/commit/74b966e403427a5154e45528301f6d3848326684

## Stacktrace

    (lldb) bt
    * thread #1, name = 'tid_303', queue = 'com.apple.main-thread', stop reason = signal SIGABRT
      * frame #0: 0x1c6eda1c libsystem_kernel.dylib`__pthread_kill + 8
        frame #1: 0x1c75e55a libsystem_pthread.dylib`pthread_kill + 278
        frame #2: 0x1c661dca libsystem_c.dylib`abort + 106
        frame #3: 0x0128c22e link all`mono_log_write_asl(log_domain=<unavailable>, level=<unavailable>, hdr=<unavailable>, message=<unavailable>) at mono-log-darwin.c:41 [opt]
        frame #4: 0x0129b870 link all`monoeg_g_log [inlined] monoeg_g_logstr(log_domain=0x00000000, log_level=G_LOG_LEVEL_ERROR, msg=<unavailable>) at goutput.c:117 [opt]
        frame #5: 0x0129b84c link all`monoeg_g_log [inlined] monoeg_g_logv_nofree(log_domain=0x00000000, log_level=G_LOG_LEVEL_ERROR, format=<unavailable>, args=<unavailable>) at goutput.c:128 [opt]
        frame #6: 0x0129b83a link all`monoeg_g_log [inlined] monoeg_g_logv(log_domain=0x00000000, log_level=G_LOG_LEVEL_ERROR, format=<unavailable>, args=<unavailable>) at goutput.c:135 [opt]
        frame #7: 0x0129b83a link all`monoeg_g_log(log_domain=<unavailable>, log_level=<unavailable>, format=<unavailable>) at goutput.c:144 [opt]
        frame #8: 0x01162ec0 link all`load_aot_module(assembly=<unavailable>, user_data=<unavailable>) at aot-runtime.c:2179 [opt]
[...]

PR #11770, opened 2018-11-23
[ios sdk] outline suspend related flags from default flags

lewurm:
The crosscompiler template also picks up the flags from the device runtime, that is:

```
ios-targetwatch_CONFIGURE_FLAGS = [...]
```

will be passed to the crosscompiler runtime as well. Therefore we had something like this:

```console
$ ./configure [...] --disable-cooperative-suspend --disable-hybrid-suspend --disable-crash-reporting --enable-cooperative-suspend [...]
```

with this PR it looks like this:

```console
$ ./configure [...] --disable-crash-reporting --enable-cooperative-suspend [...]
```

Should fix: https://github.com/mono/mono/issues/11765

### DIFF
--- a/sdks/builds/ios.mk
+++ b/sdks/builds/ios.mk
@@ -97,8 +97,6 @@ _ios-$(1)_CONFIGURE_FLAGS = \
 	--with-tls=pthread \
 	--without-ikvm-native \
 	--without-sigaltstack \
-	--disable-cooperative-suspend \
-	--disable-hybrid-suspend \
 	--disable-crash-reporting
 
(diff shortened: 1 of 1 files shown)


lewurm:
@rolfbjarne I would appreciate if you would test this patch

rolfbjarne:
@lewurm yes, it's set:

```
$ grep ENABLE_COOP_SUSPEND sdks/builds/ios-targetwatch-release/config.h
#define ENABLE_COOP_SUSPEND 1
```

I can test the PR next week.

PR #11778, opened 2018-11-26
[WIP]: Move `NegotiateStreamPal` from `System.Data` into `System`.

baulig:
Make `System` internals visible to `System.Data` and move some of the internal
GSS sources into `System`.

FIXME: only done unix so far, still need to edit Windows sources.


baulig:
@monojenkins commit csproj

PR #11780, opened 2018-11-26
Fixes #11779

### LINKED ISSUE #11779: Marshaling an interface with a base interface fails on Remoting
## Steps to Reproduce

1. Marshal an object via `System.Runtime.RemotingServices.Marshal` as an interface that implements another interface
2. Send the `ObjRef` for that object to a remote server
3. Try to unmarshal as the base interface

The below is just example code, it is not a specific reproduction case.

Shared:
```csharp
public interface IClient
{
}

public interface IClient2
{
}
```

Client:
```csharp
public class Client : IClient2 {
   public void SomeMethod() {
       ObjRef obj = RemotingServices.Marshal(this, $"Client.{_id}", typeof(IClient2));
       _server.Register(obj);
   }   
}
```

Server:
```csharp
public class Server {
    public void Register(ObjRef obj) {
        IClient client = RemotingServices.Unmarshal(obj);
    }
}
```

### Current Behavior

Mono does not send the base interface across in the `TypeInfo` for the `ObjRef` while MS.NET does. It looks like the Mono implementation of TypeInfo only adds the actual interface when the type IS an interface (https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mcs/class/corlib/System.Runtime.Remoting/TypeInfo.cs#L49). If I set a break point in the server application a client that is MS.NET has two interfaces in the `ImplementedInterfaces` (the base interface and the interface that I marshal), whereas a Mono client only has the interface that I marshal.

### Expected Behavior

MS.NET allows unmarshaling as the base interface type, so Mono should do this as well. 

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

```
mono --version
Mono JIT compiler version 5.16.0.179 (tarball Thu Oct  4 10:20:38 UTC 2018)
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          yes(3.6.0svn-mono-/)
```

slide:
This adds any base interfaces to the remoting TypeInfo to match the
behavior of MS.NET

### DIFF
--- a/mcs/class/System.Runtime.Remoting/Test/RemotingServicesTest.cs
+++ b/mcs/class/System.Runtime.Remoting/Test/RemotingServicesTest.cs
@@ -553,6 +553,39 @@ namespace MonoTests.Remoting
 			}
 		}
 
+		public interface IMarshalTest
+		{
+			void DoSomething();
+		}
+
+		public interface ITest2 : IMarshalTest
+		{
+			void DoSomethingElse();
+		}
+
+		public class TestClass : MarshalByRefObject, ITest2
+		{
+			public void DoSomething()
+			{
+			}
+
+			public void DoSomethingElse()
+			{
+			}
+		}
+
+		[Test]
+		public void MarshalBaseInterfaces()
+		{
+			TestClass test = new TestClass();
+			ObjRef obj = RemotingServices.Marshal(test, "TestClass", typeof(ITest2));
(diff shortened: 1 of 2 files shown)


slide:
I don't think the test that failed on `Windows x64 C++` is related to my change.

marek-safar:

directhex:
@monojenkins build deb

PR #11784, opened 2018-11-26
[2018-06][mini] use AOT trampolines in interp mixed mode

lewurm:
Backport of #11767


lewurm:
@monojenkins build OS X i386

PR #11786, opened 2018-11-26
Windows x64 LLVM 6.0 support.

lateralusX:


mono/mini/mini-amd64.c
```diff
@@ -2715,10 +2715,7 @@ mono_arch_start_dyn_call (MonoDynCallInfo *info, gpointer **args, guint8 *ret, g
 	pindex = 0;
 
 	/* Stored after the stack arguments */
-	nullable_buffer = (guint8*)&(p->regs [PARAM_REGS + (dinfo->nstack_args * sizeof (mgreg_t))]);
```
kg: was the old array indexer here a bug? just curious, since it looks like a bug fix

lateralusX: Correct, this was a bug, so for a use case currently just hit on Windows this caused stack corruption and crash, https://github.com/mono/mono/pull/11786/commits/6d0af3c44dd322700ce6f3240bd80f4d9a0f2b7d

mono/sgen/sgen-pinning-stats.c
```diff
@@ -220,7 +220,7 @@ sgen_pin_stats_report (void)
 		int i;
 		mono_gc_printf (sgen_gc_debug_file, "%-50s", name);
 		for (i = 0; i < PIN_TYPE_MAX; ++i)
-			mono_gc_printf (sgen_gc_debug_file, "  %10ld", pinned_entry->num_pins [i]);
+			mono_gc_printf (sgen_gc_debug_file, "  %10zd", pinned_entry->num_pins [i]);
```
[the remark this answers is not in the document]
lateralusX: Format already used in several places around the source code and should be an addition to C99 and mandatory in C11, that's why I picked it. I believe the only compiler that still struggles is mingw at this point, but it already has problems with a bunch of other formats in various files, so fixing mingw will need to make an overhaul at several places in source code. Having that said I agree with more portable way and since it already limited to %ld in current code cast to long should give same result without warning on VS2015.

mono/sgen/sgen-pinning-stats.c
```diff
@@ -229,7 +229,7 @@ sgen_pin_stats_report (void)
 		mono_gc_printf (sgen_gc_debug_file, "%-50s  %10ld\n", name, remset_entry->num_remsets);
 	} SGEN_HASH_TABLE_FOREACH_END;
 
-	mono_gc_printf (sgen_gc_debug_file, "\nTotal bytes pinned from stack: %ld  static: %ld  other: %ld\n",
+	mono_gc_printf (sgen_gc_debug_file, "\nTotal bytes pinned from stack: %zd  static: %zd  other: %zd\n",
```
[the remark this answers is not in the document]
lateralusX: Comment same as above.

mono/sgen/sgen-los.c
```diff
@@ -716,9 +716,10 @@ LOSObject*
 sgen_los_header_for_object (GCObject *data)
 {
 #if _MSC_VER
```
jaykrell: Upon further inspection.
The msvc code should be the only code.
The other thing should be offsetof from stddef.h.

mono/mini/interp/interp.c
```diff
@@ -74,6 +74,10 @@
 #include <mono/mini/mini-arm.h>
 #endif
 
+#ifdef _MSC_VER
+#pragma warning(disable:4102)
```
jaykrell: Please add a comment as to what the warning is. Numbers are cryptic.

mono/mini/mini-llvm-cpp.cpp
```diff
@@ -21,6 +21,11 @@
 // possible
 //
 
+#ifdef _MSC_VER
+#pragma warning(disable:4141)
```
jaykrell: Please add a comment as to what the warnings are. Or tell me just to use the documentation every time and gradually memorize them.

mono/sgen/sgen-gc.h
```diff
@@ -1111,7 +1111,7 @@ void sgen_env_var_error (const char *env_var, const char *fallback, const char *
 
 /* Utilities */
 
-void sgen_qsort (void *array, size_t count, size_t element_size, int (*compare) (const void*, const void*));
+void sgen_qsort (void *const array, const size_t count, const size_t element_size, int (*compare) (const void*, const void*));
```
jaykrell: Top level const is rare on parameters.
Some of these are toplevel, some are not.
I suggest:
`void sgen_qsort(const void* array, size_t, size_t, and then what you have)`
or equivalently
`void sgen_qsort(void const * array, size_t, size_t, and then what you have)`

mono/sgen/sgen-pointer-queue.c
```diff
@@ -39,7 +39,7 @@ realloc_queue (SgenPointerQueue *queue)
 	sgen_free_internal_dynamic (queue->data, sizeof (void*) * queue->size, queue->mem_type);
 	queue->data = new_data;
 	queue->size = new_size;
-	SGEN_LOG (4, "Reallocated pointer queue to size: %lu", new_size);
+	SGEN_LOG (4, "Reallocated pointer queue to size: %zu", new_size);
```
jaykrell: %z is not portable to msvc runtime.

mono/sgen/sgen-pointer-queue.c
```diff
@@ -107,7 +107,7 @@ sgen_pointer_queue_sort_uniq (SgenPointerQueue *queue)
 	void **start, **cur, **end;
 	/* sort and uniq pin_queue: we just sort and we let the rest discard multiple values */
 	/* it may be better to keep ranges of pinned memory instead of individually pinning objects */
-	SGEN_LOG (5, "Sorting pointer queue, size: %lu", queue->next_slot);
+	SGEN_LOG (5, "Sorting pointer queue, size: %zu", queue->next_slot);
```
jaykrell: %z is not portable to msvc runtime.

mono/sgen/sgen-pointer-queue.c
```diff
@@ -119,7 +119,7 @@ sgen_pointer_queue_sort_uniq (SgenPointerQueue *queue)
 		start++;
 	};
 	queue->next_slot = start - queue->data;
-	SGEN_LOG (5, "Pointer queue reduced to size: %lu", queue->next_slot);
+	SGEN_LOG (5, "Pointer queue reduced to size: %zu", queue->next_slot);
```
jaykrell: %z is not portable to msvc runtime.

jaykrell: %z is not portable to msvc runtime.

mono/sgen/sgen-pointer-queue.c
lateralusX: This is portable for msvc versions we support VS2015+ and its VS2015 triggering warning around use of %lu for size_t so fine there, but it still has issues on mingw. Since original is %lu it should be OK to type cast to unsigned long to not bring up additional warnings on mingw.

mono/sgen/sgen-pointer-queue.c
lateralusX: This is portable for msvc versions we support VS2015+ and its VS2015 triggering warning around use of %lu for size_t so fine there, but it still has issues on mingw. Since original is %lu it should be OK to type cast to unsigned long to not bring up additional warnings on mingw.

mono/metadata/w32file-win32.c
```diff
@@ -375,7 +375,7 @@ mono_w32file_get_file_system_type (const gunichar2 *path, gunichar2 *fsbuffer, g
 {
 	gboolean res;
 	MONO_ENTER_GC_SAFE;
-	res = GetVolumeInformationW (path, NULL, NULL, NULL, NULL, NULL, fsbuffer, fsbuffersize);
+	res = GetVolumeInformationW (path, NULL, 0, NULL, NULL, NULL, fsbuffer, fsbuffersize);
```
jaykrell: Plain `0` would work across the board as well.

mono/mini/mini-llvm.c
```diff
@@ -5236,7 +5240,7 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 
 			rhs = LLVMBuildSExt (builder, convert (ctx, rhs, LLVMInt32Type ()), LLVMInt64Type (), "");
 
-			v1 = LLVMBuildMul (builder, convert (ctx, rhs, IntPtrType ()), LLVMConstInt (IntPtrType (), (1 << ins->backend.shift_amount), FALSE), "");
+			v1 = LLVMBuildMul (builder, convert (ctx, rhs, IntPtrType ()), LLVMConstInt (IntPtrType (), ((unsigned long long)1 << ins->backend.shift_amount), FALSE), "");
```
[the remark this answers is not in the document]
lateralusX: Correct, could be done either way.

PR #11814, opened 2018-11-27
Fix race condition in XmlCharType.Instance

ww898:
Fix race condition in `XmlCharType.Instance` because:
```
if ( s_CharProperties == null ) {
    InitInstance();
}
```

### DIFF
--- a/mcs/class/referencesource/System.Xml/System/Xml/XmlCharType.cs
+++ b/mcs/class/referencesource/System.Xml/System/Xml/XmlCharType.cs
@@ -445,23 +445,25 @@ namespace System.Xml {
                 }
 
                 byte[] chProps = new byte[CharPropertiesSize];
-                s_CharProperties = chProps;
 
-                SetProperties( s_Whitespace,  fWhitespace );
-                SetProperties( s_LetterXml4e, fLetter );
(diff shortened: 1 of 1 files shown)


EgorBo:
Probably it makes sense to just replace the ref-source implementation with corefx:
https://github.com/mono/mono/blob/master/mcs/class/System.XML/common.sources#L371
to be
`../../../external/corefx/src/System.Private.Xml/src/System/Xml/XmlCharType.cs`

if it doesn't have this issue of course (from the quick glance - it doesn't).

ww898:
We got `XmlReader` failures with probability 2% in our case. It's too big for use because more then 1000 builds broken in our ecosystem. After apply this fix in our mono branch no failures any more. Could you please approve the fix before switching to corefx version? It's very important for us.

This condition doesn't work properly for us: https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mcs/class/referencesource/System.Xml/System/Xml/Core/XmlTextReaderImpl.cs#L3443

marek-safar:
build

marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2018-08

PR #11817, opened 2018-11-27
[mini] Add mini_target_safepoints_enabled

lambdageek:
Allows us to emit safepoints for the target even if the host doesn't need them.

Always emit safepoints when targeting watchOS. This should address https://github.com/mono/mono/issues/11765: due to https://github.com/mono/monodevelop/pull/6327, Mono was not emitting safepoints in the watchOS cross-compiler since the host had `MONO_THREADS_SUSPEND=preemptive` set.


lambdageek:
This isn't totally wrong, but I need to rethink it.  We should also decide in `metadata/marshal-ilgen.c` whether to emit blocking transitions in wrappers based on target settings, not the host.

So this shouldn't go in mini, it should go in metadata.

PR #11818, opened 2018-11-28
[runtime] Bundle mscorlib in mono

alexanderkyte:
I put this together in 20 minutes and did topical testing on my VPS. Seems to work there for simple hello world apps, thought I'd put it through the testing paces on CI.

This optimization was discussed in channel and I thought I'd throw together the proof of concept before it slipped away. If we're always AOTing mscorlib, why not link it to mono in the most optimized way? 

Making sure that we're using direct-icalls seemed like a good start.

A nice workflow improvement made by this is that MONO_PATH doesn't need to be set for the binary to work for simple applications. In addition, the corlib version mismatch problem that often leads to strange local bugs cannot happen because each mono will be closely bundled with it's corlib.

### DIFF
--- a/mcs/class/aot-compiler/Makefile
+++ b/mcs/class/aot-compiler/Makefile
@@ -30,6 +30,7 @@ csc_SCI_image = $(images_dir)/System.Collections.Immutable.dll$(PLATFORM_AOT_SUF
 
 mscorlib_dll = $(the_libdir)/mscorlib.dll
 mscorlib_aot_image = $(mscorlib_dll)$(PLATFORM_AOT_SUFFIX)
+runtime_with_mscorlib = $(the_libdir)/mono
 
 # The $(dir $(RUNTIME)) is necessary to get path to the mono binary in case when we cross-compile
 # or just compile from a different directory than the top source dir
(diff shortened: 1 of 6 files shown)


mcs/tools/mkbundle/mkbundle.cs
```diff
@@ -1085,8 +1089,10 @@ static void GenerateBundles (List<string> files)
 			tc.WriteLine ("\n}\n");
 			tc.WriteLine ("#endif\n");
 
-
-			tc.WriteLine ("static char *image_name = \"{0}\";", prog);
+			if (no_managed_entry_point)
+				tc.WriteLine ("static char *image_name = NULL;", prog);
```
jaykrell: Could/should just be `tc.WriteLine("static char* image_name");` The extra `prog` even runs successfully?

alexanderkyte: Oh good catch, yeah that was a copy-paste error.

alexanderkyte:
5076 failures, looks like the change really broke something

jaykrell:
Aren’t there various mscorlib to bundle, for various parts/times of the build?

jaykrell:
Maybe make/copy a new convenience mono. Or have the bundle favor the env var, if it doesn’t already.

alexanderkyte:
@jaykrell I specifically only changed the aot-compiler steps that work over the default profile, so we shouldn't have to worry about this change leading to breakage in the other profiles. 

`mono/mini/mono-sgen` remains a valid binary.

alexanderkyte:
A brief glance makes it seem like the problem has to do with missing config files. I think the problem is that mkbundle will just init with null if we don't pass one, and then ignores the later `--config` flag

mcs/class/aot-compiler/Makefile
```diff
@@ -72,6 +73,26 @@ $(csc_SRM_image): $(csc_SRM_dll) $(runtime_dep)
 $(csc_SCI_image): $(csc_SCI_dll) $(runtime_dep)
 	$(Q_AOT) MONO_PATH='$(the_libdir)' > $(LOG_FILE) 2>&1 $(RUNTIME) --aot=bind-to-runtime-version$(profile_arg),outfile=$(csc_SCI_image) --debug $(csc_SCI_dll) || (cat $(LOG_FILE); exit 1)
 
+MKBUNDLE_EXE = $(topdir)/class/lib/$(PROFILE)/mkbundle.exe
```
marek-safar: Please don't do it this way, add it to build tools https://github.com/mono/mono/blob/master/mcs/tools/Makefile#L54 instead

mcs/class/aot-compiler/Makefile
```diff
@@ -97,7 +118,13 @@ endif
 
 ifeq ($(PROFILE), $(DEFAULT_PROFILE))
 
-IMAGES = $(mscorlib_aot_image) $(mcs_aot_image) $(CSC_IMAGES)
+ifndef DISABLE_BUNDLE_CORLIB
+CORLIB_IMAGE = $(runtime_with_mscorlib)
```
marek-safar: This is confusing to read, just use `IMAGE` and change the assignment below to IMAGES += $(mcs_aot_image) $(CSC_IMAGES)

mcs/tools/mkbundle/template_common.inc
```diff
@@ -18,8 +18,6 @@ void initialize_mono_api (const BundleMonoAPI *info)
 	mono_api.mono_register_machine_config = info->mono_register_machine_config;
 }
 
-#ifdef USE_COMPRESSED_ASSEMBLY
```
marek-safar: Why did you undo this change?

alexanderkyte: If you use mkbundle and disable USE_COMPRESSED_ASSEMBLY, it doesn't build currently. This change removed too much code. I intend to find the commit that made the problem and fix it correctly, this was a prototype to see if I could get it working well without investing much in it.

I intend to circle back and fix this for my next sprint, as I want to also benchmark the impact of this change and see if some knobs should be tuned. We probably want more than just direct-icalls with the normal AOT options if we're special casing mscorlib.

mcs/class/aot-compiler/Makefile
alexanderkyte: Will do

alexanderkyte:
The .NET 5 BCL transition will make this rather interesting. Closing because it didn't seem to have much desire to be merged in the current incarnation. This is a rather small set of changes so I expect it to be good documentation for the effort that does this (if it gets revived). Good comments; looks like this served it's purpose to explore the idea and to get feedback about the route we'd like to take to accomplish it.

PR #11823, opened 2018-11-28
[runtime] Fix return of uninit value

alexanderkyte:
We have a return on [line 3608](https://github.com/mono/mono/blob/892cad1dab3db16c269ae22b0d68b0bd1d5c0668/mono/metadata/object.c#L3608) that doesn't have any assignment to the variable "o" before that. Since it's stack memory, it will be garbage.

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -3527,7 +3527,7 @@ mono_field_get_value_object_checked (MonoDomain *domain, MonoClassField *field, 
 
 	error_init (error);
 
-	MonoObject *o;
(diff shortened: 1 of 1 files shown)


alexanderkyte:
Ah, I see that's not the case. clang was complaining about this deref but I see an assignment right above it.

jaykrell:
That isn't obvious, but, yeah initializing locals is a good idea. Not initializing them in a seldom worthwhile micro optimization imho.

alexanderkyte:
@jaykrell fair point, we should probably make this change either way.

PR #11829, opened 2018-11-28
[System] Refactor tests to use system-assigned ports whenever possible.

rolfbjarne:
Refactor tests to use system-assigned ports whenever possible, so that we
avoid the potential race condition in NetworkHelpers.FindFreePort.

This should make it much less probable to run into spurious "Address already
in use" exceptions.

In some cases I kept the call to NetworkHelpers.FindFreePort:

* When the test is not binding/listening on a port, but instead trying to
  connect to an inexistent port (for tests that verifies error handling /
  exceptions). In this case there is still a race condition (if a test tries
  to verify that a connection fails, it may unexpectedly succeed instead).

* Sometimes the test doesn't actually use the port (the test is verifying
  something very different, for instance a ArgumentNullExceptions, which are
  thrown before the port is used).

* Some API doesn't allow for requesting a system assigned port:

    * HttpListener: in this case I elected to re-try a few times in case of
      "Address already in use" exceptions.

    * The 'UdpClient (string, port)' constructor (other constructors work
      fine): these turned out a bit complicated to work around (or would
      modify the test more than would be reasonable), so for these we still
      have to hope for the best.

Hopefully fixes https://github.com/xamarin/maccore/issues/604.

### DIFF
--- a/mcs/class/System/Test/System.Net/HttpListener2Test.cs
+++ b/mcs/class/System/Test/System.Net/HttpListener2Test.cs
@@ -62,21 +62,53 @@ namespace MonoTests.System.Net {
 			}
 		}
 
-		public static HttpListener CreateAndStartListener (string prefix)
+		public static HttpListener CreateAndStartListener (string host, int port, string path, AuthenticationSchemes? authSchemes = null, Action<HttpListener> initializer = null)
 		{
+			return CreateAndStartListener (host, path, port, authSchemes, initializer);
+		}
+
+		public static HttpListener CreateAndStartListener (string host, string path, int port, AuthenticationSchemes? authSchemes = null, Action<HttpListener> initializer = null)
+		{
+			var prefix = host + port + path;
 			HttpListener listener = new HttpListener ();
+			if (initializer != null)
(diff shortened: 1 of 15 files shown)


mcs/class/System/Test/System.Net/HttpListener2Test.cs
```diff
@@ -62,23 +62,55 @@ public Socket GetSocket ()
 			}
 		}
 
-		public static HttpListener CreateAndStartListener (string prefix)
+		public static HttpListener CreateAndStartListener (string host, int port, string path, AuthenticationSchemes? authSchemes = null, Action<HttpListener> initializer = null)
```
akoeplinger: please move this and the other methods to `NetworkHelpers` so unrelated test classes don't need to call into `HttpListener2Test`.

This will also fix the Android SDK build since for some reason we exclude HttpListener2Test.cs in the monodroid profile.

rolfbjarne: @akoeplinger thanks, I've changed it as requested.

PR #11831, opened 2018-11-28
[System] Implement a helper function to create unique temporary directories and use it in tests.

rolfbjarne:
While working on PR #11829 I stress-tested the tests by running multiple test
runs in parallel, and it turned out that many tests were not parallel-safe,
because they used the same temporary file names and ended up stomping on
eachother.

So I implemented a helper method that creates a guaranteed unique temporary
directory, and I've changed all tests to use this new helper method.


mcs/class/test-helpers/PathHelpers.cs
```diff
@@ -0,0 +1,109 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace MonoTests.Helpers {
+
+	public static class PathHelpers
```
akoeplinger: I'd prefer using [CoreFX TempDirectory.cs](https://github.com/dotnet/corefx/blob/24db8490eb6bdeeb2c4353d5b7f9b786546cfa2b/src/Common/tests/System/IO/TempDirectory.cs) instead (either directly or if you really care about including the assembly name then copy the source and adapt it).

rolfbjarne: Is it the IDisposable pattern of TempDirectory you prefer?

That would require more changes to the tests.

I'd also have to copy & adapt the source, since operates on a "hope for no duplicate directories" approach (it's not parallel safe).

mcs/class/test-helpers/PathHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+	{
+		static string root;
+		static uint last_number;
+
+		static PathHelpers ()
+		{
+			// Create a root temporary directory that is to be the base of all other temporary directories we create.
+			// Contains the assembly name to make the name somewhat intelligible,
+			// and make it include the PID so that parallel execution won't stomp on eachother.
+			root = Path.Combine (Path.GetTempPath (), Assembly.GetExecutingAssembly ().GetName ().Name + "_" + Process.GetCurrentProcess ().Id);
```
jaykrell: Just using a guid would work well.

mcs/class/test-helpers/PathHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+		{
+			string name;
+			var calling_method = new StackFrame (1).GetMethod ();
+			if (calling_method != null) {
+				name = calling_method.DeclaringType.FullName + "_" + calling_method.Name;
+			} else {
+				name = "unknown-test";
+			}
+
+			// We store the last number used to create a unique directory, to
```
jaykrell: Guid would be much simpler. Possibly removing dashes and braces.

mcs/class/test-helpers/PathHelpers.cs
jaykrell: Please consider just generating a guid.  Much less code that way. No need to check for  existance, etc. It is what they were meant for.

akoeplinger: @rolfbjarne the IDisposable and not doing all the special magic code that Jay seems to dislike too. Alternatively use `Path.GetRandomFileName()` (it just returns a unique path despite the name) instead of hand rolling the implementation.

Why does the IDisposable pattern need changes to the tests?

jaykrell: Very random does sound promising/adequate/simple but I still like guids.

rolfbjarne: > Why does the IDisposable pattern need changes to the tests?

Using the IDisposable pattern means disposing the TempDirectories, which means the test code needs to be changed to dispose them (adding `using`s, which probably means code will end up re-indented, which means a lot of whitespace changes, etc - nothing groundbreaking, but the diff would end up bigger than what it's today).

mcs/class/test-helpers/PathHelpers.cs
rolfbjarne: Guid would be simpler, but I don't really like filling /tmp with randomly named stuff. It makes me wonder why it's there and who created it (but not enough to actually figure out the why and who, which just leaves me annoyed).

That said, I won't be running the System tests very often, so I'm fine with whatever you think is best.

jaykrell: You can prefix the guid with something indicative and constant.

mcs/class/test-helpers/PathHelpers.cs
akoeplinger: @rolfbjarne you don't need to explicitly dispose the temp directory, it'll be GC'd once it's no longer used.

rolfbjarne:
@akoeplinger I've copied and modified TempDirectory, and changed the code to use it.
@jaykrell I've changed the code to use Guid instead, prefixed with the test type+method name.

PR #11833, opened 2018-11-28
[coop] Separate host and target suspend policies

lambdageek:
Always set target suspend policy to coop when cross-compiling to watchOS.

Target policy affects two places:
- safepoint generation in AOT in mini
- blocking transition insertion in marshal-ilgen

Everything else (primarily metadata) cares about host policy (same as before).

There's one tricky bit:  if `MONO_ENABLE_BLOCKING_TRANSITION` env var is set, it affects both host and target, for backward compatibility.  (although this variable wasn't widely publicized or used in production, afaik).

This addresses #11765 which was caused by mono/monodevelop#6327 where `MONO_THREADS_SUSPEND=preemptive` affected both host policy and the watchOS cross compiler target.  Now the env var is just about the host.

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -577,8 +577,8 @@ threads_suspend_policy_getenv (void)
 
 static char threads_suspend_policy;
 
-static MonoThreadsSuspendPolicy
-mono_threads_suspend_policy (void)
+MonoThreadsSuspendPolicy
+mono_threads_host_suspend_policy (void)
 {
 	int policy = threads_suspend_policy;
 	if (G_UNLIKELY (policy == 0)) {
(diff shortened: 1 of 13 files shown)


vargaz:
We should avoid safepoints at least on wasm which has no async gc.

luhenry:
@vargaz this is a much smaller problem, and it would be straightforward to not generate safepoints on wasm (the `if (mono_threads_polling) mono_threads_state_poll()` code)., but still allow us to unify the paths in, for example, `marshal-ilgen.c`.

lambdageek:
> but still allow us to unify the paths in, for example, marshal-ilgen.c

I don't understand this part: wasm doesn't need thread state transitions either.

mono/utils/mono-threads-coop.c
```diff
@@ -640,40 +639,66 @@ mono_threads_suspend_policy_name (void)
...
+	static int inenv = -1;
+	if (G_UNLIKELY (inenv == -1))
+		inenv = g_hasenv ("MONO_ENABLE_BLOCKING_TRANSITION");
+	return inenv;
+}
+
+static gboolean
+blocking_transition_from_policy (MonoThreadsSuspendPolicy p)
+{
+	switch (mono_threads_host_suspend_policy ()) {
```
lambdageek:
```suggestion
	switch (p) {
```

lambdageek:
This is going to be reworked - blocking transitions and safepoints on most targets (except wasm).

PR #11836, opened 2018-11-29
[2018-06] [interp] Compile method for errors when needed

lewurm:
Partial backport of https://github.com/mono/mono/pull/10095

This change is also needed for the following situation: On a native-to-managed transition the interpreter executes the `NATIVE_TO_MANAGED` wrapper. This wrapper makes sure that the thread is attached. However, doing `transform` requires an attached thread in some cases (e.g. when you run with `--debug`). This commit makes sure that we transform the `InterpMethod` when the thread is attached, that is, when the entry point for the native-to-managed transition is requested.

FYI: This is already in `2018-08` and up.

### DIFF
--- a/mono/mini/ee.h
+++ b/mono/mini/ee.h
@@ -15,7 +15,7 @@
 #ifndef __MONO_EE_H__
 #define __MONO_EE_H__
 
-#define MONO_EE_API_VERSION 0x4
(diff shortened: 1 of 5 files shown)


lewurm:
@monojenkins build OS X x64

PR #11661, opened 2018-11-13
[corlib] Make all Encoding types serializable

EgorBo:
Fixes https://github.com/mono/mono/issues/11529 
Fixes https://github.com/mono/mono/issues/11663

PR #11842, opened 2018-11-29
[2018-10] Fix Encoding serialization issue

EgorBo:
Manual backport of #11661

### DIFF
--- a/mcs/class/corlib/Test/System.Text/EncodingTest.cs
+++ b/mcs/class/corlib/Test/System.Text/EncodingTest.cs
@@ -27,8 +27,10 @@
 //
 
 using System;
+using System.IO;
 using System.Text;
-
+using System.Linq;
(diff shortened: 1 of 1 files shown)


mcs/class/corlib/Test/System.Text/EncodingTest.cs
```diff
@@ -128,5 +130,32 @@ public void EncodingName ()
 		{
 			Assert.AreEqual ("Unicode (UTF-8)", Encoding.UTF8.EncodingName);
 		}
+
+		[Test] // https://github.com/mono/mono/issues/11663
+		[Category ("MobileNotWorking")] // Custom Encodings are not included by default on mobile platforms
```
akoeplinger: this one should work fine on mobile, at least according to the master builds. it only uses ASCII encoding which is available everywhere.

marek-safar:
@monojenkins commit apidiff

PR #11851, opened 2018-11-29
[Mono.Debugger.Soft] Make exposing Cecil metadata an option

jbevain:
Not all downstream users of the Mono.Debugger.Soft library are using the feature where mirrors provide direct access to the Cecil metadata if the `AssemblyMirror` has been fed a Cecil `AssemblyDefinition`.

This PR introduces a define to opt-in to use the Cecil metadata, otherwise falls back to exposing System.Reflection.Emit opcodes in the Instruction API.

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Makefile
+++ b/mcs/class/Mono.Debugger.Soft/Makefile
@@ -5,7 +5,7 @@ LIBRARY = Mono.Debugger.Soft.dll
 LIBRARY_SNK = ../mono.snk
 
 LIB_REFS = System Mono.Cecil System.Core
-LIB_MCS_FLAGS = /unsafe -D:MONO_DATACONVERTER_STATIC_METHODS /publicsign
+LIB_MCS_FLAGS = /unsafe -D:MONO_DATACONVERTER_STATIC_METHODS -D:ENABLE_CECIL /publicsign
 KEYFILE = $(LIBRARY_SNK)
 
 TEST_LIB_REFS = Mono.Cecil System System.Core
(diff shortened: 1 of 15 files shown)


vargaz:
Why does this have to be behind an ifdef ?

marek-safar:
@jbevain what is the intention with the define? why you want a version which has no Cecil dependency?

jbevain:
@marek-safar yes that is correct, we're not using this feature and would like to avoid taking a dependency we don't use.

jbevain:
@vargaz would you rather remove the feature completely? That would be a big API break, and some downstreams like `mono/debugger-libs` may be using it.

I figured this way people can choose to take the dependency or not.

@marek-safar , @vargaz , would you prefer to have the Cecil dependency opt-out instead of opt-in?

vargaz:
What problem having this dependency causes ? Its just an assembly which has to be present, its a desktop assembly, it shouldn't be a problem.

jbevain:
@vargaz we're just trying to cut down on our dependencies. Inside Visual Studio, everything has to run in one app domain, and we've seen conflicts of assemblies like Cecil between plugins.

To address this we have our own Cecil with its own strong name, but we're not using anywhere and would like to not have to ship it. Shipping assemblies as part of VS means a bunch of bookkeeping for each assembly, so we'd gladly remove one that we're not using.

vargaz:
Its ok, it just looks a bit ugly. Also, the define could simpler like ENABLE_CECIL or something.

jbevain:
@vargaz here you go :)

marek-safar:
@monojenkins commit csproj

PR #11862, opened 2018-11-30
Fix warnings and race condition in time keeping.

jaykrell:
```
/s/mono/mono/utils/mono-time.c:23:34: warning: unused variable 's_TimebaseInfo' [-Wunused-variable]
static mach_timebase_info_data_t s_TimebaseInfo;

/s/mono/mono/utils/mono-time.c:112:1: warning: unused function 'get_boot_time' [-Wunused-function]
```

### DIFF
--- a/mono/utils/mono-time.c
+++ b/mono/utils/mono-time.c
@@ -20,7 +20,6 @@
 
 #if HAVE_MACH_ABSOLUTE_TIME
 #include <mach/mach_time.h>
-static mach_timebase_info_data_t s_TimebaseInfo;
 #endif
 
 #define MTICKS_PER_SEC (10 * 1000 * 1000)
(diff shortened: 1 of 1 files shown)


mono/utils/mono-time.c
```diff
@@ -109,6 +108,7 @@ mono_100ns_datetime (void)
 
 #include <time.h>
 
+#if 0
```
lambdageek: Just delete it.   It should've been removed as part of 06a4273d73cfa732d911bdaf50f752c025a0665c

jaykrell: Ok.

jaykrell:
The struct is only two uint32. The  barrier could be eliminated, at least on 64bit platforms, by doing a 64bit write to it.

jaykrell:
@monojenkins build failed

Issue #11430, opened 2018-10-29
.NET Standard 2.1 support


marek-safar:
Umbrella ticket tracking work on full .NET Standard 2.1 support. The changes and enhancements to the standard have started landing to dotnet/standard master but the .NET Standard 2.1 has not been finalized.

Most of the missing APIs will show up automatically as we progress on [importing CoreFX code](#6845).

The planning document is available at https://github.com/dotnet/standard/tree/master/docs/planning/netstandard-2.1 

## Namespaces with significant differences on desktop (net_4_x)
- [x] System.Reflection
- [x] System.Reflection.Emit
- [x] System.Drawing
- [x] System.IO
- [x] System.IO.Compression with Brotli #11431
- [x] System.Security.Cryptography
- [x] System.Net

## Mobile profiles changes
- [x] System.Xml.Serialization
- [x] System.Drawing convertors
- [x] DbProviderFactories
- [x] ECDiffieHellman

## Runtime required changes
- [ ] https://github.com/mono/mono/issues/8397

The announcement is available at https://blogs.msdn.microsoft.com/dotnet/2018/11/05/announcing-net-standard-2-1/

marek-safar:
Status check https://gist.github.com/marek-safar/acc256181fa4ef5a74f3c36d4c2a783c

marek-safar:
Status check https://gist.githubusercontent.com/marek-safar/b322a00f68f45e2d154a8d3cc9b32745/raw/e56378308a16665281b334a7bd7f377ae490832c/gistfile1.txt

danielmeza:
Hi, I wish to know the state of this, where I can find info about the release or integration?

marek-safar:
Unfortunately .NET 2.1 is still moving target but our goal is to have it available fully in Mono 6.4 which aligns with dev16.3.

danielmeza:
Si we have this on the 16.3 preview right now?

borgdylan:
What about IAsyncEnumerable that is also in NS 2.1?

danielmeza:
Currently in android it work but the project still use by default the 7.3 version o c#, so do yo need to edit the .csproj file and put the version manually to 8.0 or to latest.

aaronfranke:
What is the status of this? How do you use .NET Standard 2.1 with Mono? This comment on this issue indicates that you can't: https://github.com/dotnet/standard/issues/1689#issuecomment-579550027

rvalitov:
[...]

PR #11871, opened 2018-11-30
[corlib] Make by-ref struct actually ref now C# supports them.

marek-safar:
This is technically breaking change but the types are already specialized
in C# compiler to be ref structs therefore the breaking change is more
theoretical than practical

Part of #11430


marek-safar:
@monojenkins apidiff

akoeplinger:
@monojenkins commit apidiff

lambdageek:
Nice.

I hardcoded `RuntimeArgumentHandle` and `TypedReference` as `IsByRefLike` in the runtime.  Can we drop that now, or will that break things?

marek-safar:
@lambdageek yes, that can be removed now

mcs/class/corlib/System/RuntimeArgumentHandle.cs
```diff
@@ -31,13 +31,9 @@
...
 //
 
-using System;
-using System.Runtime.InteropServices;
-
 namespace System
 {
-	[ComVisible (true)]
-	public struct RuntimeArgumentHandle
+	public ref struct RuntimeArgumentHandle
```
benaadams: Could also be `readonly`?
```csharp
public readonly ref struct RuntimeArgumentHandle
{
    internal readonly IntPtr args;
}
```

Issue #11809, opened 2018-11-27
AM_EXTRA_RECURSIVE_TARGETS requires AM 1.13; breaks CentOS6


directhex:
```
05:27:41 DEBUG: configure.ac:23: error: possibly undefined macro: AM_EXTRA_RECURSIVE_TARGETS
05:27:41 DEBUG:       If this token and others are legitimate, please use m4_pattern_allow.
05:27:41 DEBUG:       See the Autoconf documentation.
05:27:41 DEBUG: **Error**: autoconf failed.
```

Introduced in 5d2525527ef8 and c4087410008d

There's already special-case handling for stuff like this in our AM_INIT_AUTOMAKE

akoeplinger:
Do you think we need to find a workaround or is it fine to bump our requirements to AM1.13 ?

directhex:
Automake 1.13 is problematic for RHEL/CentOS 6.

We should have discussions before dropping 6. 8 is in beta though.

PR #11877, opened 2018-11-30
[configure.ac] Only use AM_EXTRA_RECURSIVE_TARGETS when available

akoeplinger:
It's only available in automake 1.13+. Since we only use it for tests
we can just make it optional on systems with earlier automake.

Fixes https://github.com/mono/mono/issues/11809


akoeplinger:
Merging to unblock the centos build.

PR #11855, opened 2018-11-30
[crash] Add crash reporter logger

alexanderkyte:

PR #11884, opened 2018-12-01
[crash] Switch crash reporting writers to no-allocate + validate json

alexanderkyte:


alexanderkyte:

alexanderkyte:

alexanderkyte:
@monojenkins build failed

mono/utils/mono-merp.c
```diff
@@ -206,95 +217,60 @@ parse_exception_type (const char *signal)
 	g_error ("Merp doesn't know how to handle %s\n", signal);
 }
 
-static gchar *
-mono_encode_merp_params (MERPStruct *merp)
-{
-	GString *output = g_string_new ("");
+#define FORMAT_TEXT_TO_FILE(handle, buff, sizeof_buff, ...) do { \
```
luhenry: `MOSTLY_ASYNC_FPRINTF`? So it more closely match the name `MOST_ASYNC_PRINTF`, and so you can even reimplement the latter in term of the former.

mono/utils/mono-merp.c
```diff
@@ -206,95 +217,60 @@ parse_exception_type (const char *signal)
...
-	g_string_append_printf (output, "ApplicationBundleId: %s\n", merp->bundleIDArg);
-	g_string_append_printf (output, "ApplicationVersion: %s\n", merp->versionArg);
+static int merp_file_permissions = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH;
 
-	g_string_append_printf (output, "ApplicationBitness: %s\n", get_merp_bitness (merp->archArg));
+static gboolean
+mono_merp_write_params (MERPStruct *merp)
+{
+	int handle = g_open (merp->merpFilePath, O_TRUNC | O_WRONLY | O_CREAT, merp_file_permissions);
+	if (handle == -1) {
```
luhenry: `g_assertf (handle != -1, "Could not open MERP file at %s", merp->merpFilePath);` to trigger a double fault.

mono/utils/mono-merp.c
```diff
@@ -206,95 +217,60 @@ parse_exception_type (const char *signal)
...
-			fprintf (stderr, "Crashing Dump File:\n####\n%s\n####\n", crashLog);
-		if (werXml != NULL)
-			fprintf (stderr, "Crashing XML WER File:\n####\n%s\n####\n", werXmlPath);
-	}
-
-	// // Create process to launch merp gui application
-	const char *argvOpen[] = {"/usr/bin/open", "-a", config.merpGUIPath, NULL};
-	int status = posix_spawn(NULL, "/usr/bin/open", NULL, NULL, (char *const*)(argvOpen), NULL);
-	g_assertf (status == 0, "Could not start the Microsoft Error Reporting client (at %s). Error code: %d\n", config.merpGUIPath, status);
+	if (status != 0)
```
luhenry: Revert to `g_assert (status == 0);`.

mono/utils/mono-merp.c
```diff
@@ -357,162 +331,112 @@ mono_init_merp (const intptr_t crashed_pid, const char *signal, MonoStackHash *h
...
-
-	mono_json_writer_indent (&writer);
-	mono_json_writer_object_key(&writer, "ExceptionType:");
-	mono_json_writer_printf (&writer, "\"%s\",\n", get_merp_exctype (merp->exceptionArg));
-
-	mono_json_writer_indent (&writer);
-	mono_json_writer_object_key(&writer, "StackChecksum:");
-	mono_json_writer_printf (&writer, "\"0x%x\",\n", merp->hashes.offset_free_hash);
+	int handle = g_open (merp->crashLogPath, O_TRUNC | O_WRONLY | O_CREAT, merp_file_permissions);
+	if (handle == -1) {
```
luhenry: `g_assertf (handle != -1, "Could not open crash log file at %s", merp->merpFilePath);`

mono/utils/mono-merp.c
```diff
@@ -357,162 +331,112 @@ mono_init_merp (const intptr_t crashed_pid, const char *signal, MonoStackHash *h
...
+mono_write_wer_template (MERPStruct *merp)
 {
 	// Note about missing ProcessInformation block: we have no PID that makes sense
 	// and when mono is embedded and used to run functions without an entry point,
 	// there is no image that would make any semantic sense to send either. 
 	// It's a nuanced problem, each way we can run mono would need a separate fix.
 
-	GString *output = g_string_new ("");
+	int handle = g_open (merp->werXmlPath, O_WRONLY | O_CREAT | O_TRUNC, merp_file_permissions);
+	if (handle == -1) {
```
luhenry: `g_assertf (handle != -1, "Could not open WER XML file at %s", merp->merpFilePath);`

mono/utils/mono-state.c
```diff
@@ -26,263 +31,280 @@ extern GCStats mono_gc_stats;
...
-	static_gstr.len = 0;
-	static_gstr.allocated_len = len;
-	static_gstr.str = output_dump_str;
+	// Not using static, exit
+	if (writer->allocated_len == 0)
+		return;
 
-	writer.indent = 0;
-	writer.text = &static_gstr;
+	if (writer->allocated_len - writer->len < margin)
```
luhenry: `g_assertf`

mono/utils/mono-state.c
```diff
@@ -637,24 +647,21 @@ mono_crash_dump (const char *jsonFile, MonoStackHash *hashes)
...
-			if (ftell (fp) == 0) {
-				fwrite (jsonFile, size, 1, fp);
-				success = TRUE;
-			}
+		char name [100]; 
+		name [0] = '\0';
+		g_snprintf (name, sizeof (name), "mono_crash.%" PRIx64 ".%d.json", hashes->offset_free_hash, increment);
+
+		int handle = g_open (name, O_WRONLY | O_CREAT | O_EXCL);
+		if (handle == -1) {
```
luhenry: `g_assertf(…)` for the double fault.

mono/utils/mono-state.c
```diff
@@ -31,266 +31,280 @@ extern GCStats mono_gc_stats;
 #include <mach/task_info.h>
 #endif
 
+#define FORMAT_APPEND_STATIC(writer, ...) do { \
```
luhenry: Why a macro? This looks largely used like `mono_json_writer_printf`, so how about `mono_state_writer_printf` as a function.

mono/metadata/icall-def.h
```diff
@@ -137,7 +137,7 @@ ICALL_TYPE(RUNTIME, "Mono.Runtime", RUNTIME_1)
 HANDLES(RUNTIME_1, "DisableMicrosoftTelemetry", ves_icall_Mono_Runtime_DisableMicrosoftTelemetry, void, 0, ())
 HANDLES(RUNTIME_15, "DumpStateSingle_internal", ves_icall_Mono_Runtime_DumpStateSingle, MonoString, 2, (guint64_ref, guint64_ref))
 HANDLES(RUNTIME_16, "DumpStateTotal_internal", ves_icall_Mono_Runtime_DumpStateTotal, MonoString, 2, (guint64_ref, guint64_ref))
-HANDLES(RUNTIME_2, "EnableMicrosoftTelemetry_internal", ves_icall_Mono_Runtime_EnableMicrosoftTelemetry, void, 6, (char_ptr, char_ptr, char_ptr, char_ptr, char_ptr, char_ptr))
+HANDLES(RUNTIME_2, "EnableMicrosoftTelemetry_internal", ves_icall_Mono_Runtime_EnableMicrosoftTelemetry, void, 7, (char_ptr, char_ptr, char_ptr, char_ptr, char_ptr, char_ptr, char_ptr))
```
jaykrell: Maybe `const_char_ptr`?

mono/utils/mono-state.c
```diff
@@ -26,263 +31,280 @@ extern GCStats mono_gc_stats;
...
+
+	assert_has_space (writer);
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "IP");
+	FORMAT_APPEND_STATIC(writer, "\"%p\",\n", (gpointer) MONO_CONTEXT_GET_IP (ctx));
+
+	assert_has_space (writer);
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "SP");
+	FORMAT_APPEND_STATIC(writer, "\"%p\",\n", (gpointer) MONO_CONTEXT_GET_SP (ctx));
```
jaykrell: Consider a different approach -- sprintf is a slow, type unsafe, overly general function. Very popular, but still slow, unsafe, overly  general. The slowness doesn't matter in crashing, but the rest maybe.
```
json_append_char  (writer, '"');
json_append_pointer  (writer, (gpointer) MONO_CONTEXT_GET_SP (ctx));
json_append_char  (writer, '"');
```

mono/utils/mono-state.c
jaykrell: Agreed.

One thing gained by a macro is the assert is reported in-situ. But agreed -- stack trace doubles for assert context, most people don't think about assert context anyway, and  the primary reason to favor macros over function is "syntactic structure within macro", like the error handling and error_init macros. Also almost-guaranteed inlining, which doesn't apply here.  Or to double as C++ templates, which doesn't apply here.

glib has something like this -- g_string_append_printf -- that should  be reusable. It might need some "policy" flag to inhibit realloc.

mono/utils/mono-merp.c
alexanderkyte: Yeah, this is probably a critical failure. I'll add that.

mono/utils/mono-state.c
alexanderkyte: @jaykrell aren't you still going to end up needing to use format strings for all the "%x" "%p" "0x%llx" values? Or are you suggesting manual formatting of all of these strings?

mono/metadata/icall-def.h
jaykrell: Right, they can all change.

mono/utils/mono-state.c
jaykrell: That is one approach, and it does add a sort of type safety, but no, you can write a bunch of small functions that are simpler and faster.

```
#include <stdio.h>
#include <stddef.h>
typedef int gboolean;
#define TRUE 1
#define FALSE 0
typedef unsigned long long uint64;
typedef long long int64;

void
mono_strrev (char* buf, size_t len)
{
 if (len < 2) return;
 size_t j = len - 1;
 for (size_t i = 0; i < j; ++i, --j)
 {
   char c = buf [i];
   buf [i] = buf [j];
   buf [j] = c;
 }
}

int
mono_sprintf_int (char* buf, uint64 u, unsigned base, gboolean upper, unsigned width)
{
int len = 0;
do
{
 buf [len++] = (upper ? "0123456789ABCDEF" : "0123456789abcdef") [u % base];
 width -= (width > 0);
 u /= base;
} while (u);
while (width--)
  buf [len++] = '0';
 mono_strrev (buf, len);
 return len;
}

int
mono_sprintf_u (char* buf, uint64 u)
{
 return mono_sprintf_int (buf, u, 10, FALSE, 0);
}

int
mono_sprintf_d(char* buf, int64 i)
{
 uint64 u;
 int neg = (i < 0);
 if (neg) {
    *buf++ = '-';
    u = 1 + (uint64)-(i + 1); // avoid negating the most negative number
 }
 else u = (uint64)i;
 return neg + mono_sprintf_int (buf, u, 10, FALSE, 0);
}

int
mono_sprintf_x (char* buf, uint64 u)
{
 return mono_sprintf_int (buf, u, 16, FALSE, 0);
}

int
mono_sprintf_X (char* buf, uint64 u)
{
 return mono_sprintf_int (buf, u, 16, TRUE, 0);
}

int
mono_sprintf_8x (char* buf, uint64 u)
{
 return mono_sprintf_int (buf, u, 16, FALSE, 8);
}


int main()
{
char  buf[99];

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_int (buf, 0x123, 16, 0, 0)] = 0;
printf("0x%s\n", buf);

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_int (buf, 123, 10, 0, 0)] = 0;
printf("%s\n", buf);

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_int (buf, 0x123a, 16, 1, 16)] = 0;
printf("%s\n", buf);

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_int (buf, 0x1234ABCD1234ABCD, 16, 1, 16)] = 0;
printf("%s\n", buf);

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_d (buf, -123)] = 0;
printf("%s\n", buf);

memset (buf,  'X', sizeof (buf));
buf [mono_sprintf_u (buf, 123)] = 0;
printf("%s\n", buf);
}

```

jaykrell: sprintf also has reentrancy problems, at least on some  platforms.
On Visual  C++ it deals with locales and locks, and %f does a heap allocation.
Perhaps the non-Windows versions are more reentrant.

mono/utils/mono-state.c
alexanderkyte: This is called from an icall. It seems a bit much to crash the runtime in that case.

alexanderkyte: Also this isn't what we want because we are doing this loop to try to make file names that may already be taken.

alexanderkyte: In practice, there are often multiple mono_crash files with the same stack hash

alexanderkyte:
I need to fix handling of open. Fix incoming, testing it now.

alexanderkyte:
The test can be made to try to open merp. ```+               var merpGUIPath = "/Applications/Visual\\ Studio.app/Contents/SharedSupport/Microsoft\\ Error\\ Reporting.app";
``` 

But I have been unable to get the .app to start for a while now. I've been reporting this to the VS4M team. It crashes on launch ever since the last major release. This is the case even when the merp .folder is cleaned out.

File: mono/utils/mono-threads-debug.h [hunk 11]
```diff
@@ -6,10 +6,15 @@
 #define __MONO_UTILS_MONO_THREADS_DEBUG_H__
 
 /* Logging - enable them below if you need specific logging for the category you need */
+#define MOSTLY_ASYNC_SAFE_FPRINTF(handle, buff, sizeof_buff, ...) do { \
+	buff [0] = '\0'; \
```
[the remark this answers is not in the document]
alexanderkyte: I made this change. When I first made the function, that buffer argument was used in a much more interesting way. You're right that there's no need to keep that flexibility with how we're using it in this PR now.

alexanderkyte:

alexanderkyte:
If it were a function then the stack would be popped and pushed by sizeof(buff), a huge amount relative to what a good C compiler could generate for a series of those macro calls. 

But maybe that's a microoptimization.

alexanderkyte:
@monojenkins build failed

jaykrell:
Right it is micro optimization. The push and pop are a single add/subtract. (Cost a bit higher if function has more than 4K of locals, on NT).

alexanderkyte:
Why was the other function a macro in the first place then?

jaykrell:
I don't know. It is fine asis. Just something to consider in future.

alexanderkyte:
Let's see if this will need a manual backport along with https://github.com/mono/mono/pull/11855#issuecomment-444675126

alexanderkyte:

alexanderkyte:

Issue #11898, opened 2018-12-02
Windows i386 CI very high failure rate lately.


jaykrell:
Windows i386 CI is failing very frequently lately.
In  pinvoke2.exe.
Here is a reduced case that fails.
It is heap corruption -- well, mismatched allocate vs. free  -- different heap.
If you run under a debugger, it breaks in due to the heap corruption detected in free.

```
using System;
using System.Runtime.InteropServices;

public class Tests
{
	[DllImport ("libtest", EntryPoint="mono_test_asany")]
	public static extern int mono_test_asany ([MarshalAs (UnmanagedType.AsAny)] object o, int what);

	public static int Main (string[] args)
	{
		mono_test_asany ("ABC", 2);
		return 0;
	}
}
```

jaykrell:
There is a second problem sometimes in this test:

cant resolve internal call to "Tests/T2::MyClone" (tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: C:\s2\mono\mono\tests\pinvoke2.exe

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.
cant resolve internal call to "Tests/T2::MyClone" (tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: C:\s2\mono\mono\tests\pinvoke2.exe

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.
test_0_x86_single_double_struct_ret failed: got 1, expected 0
Regression tests: 97 ran, 1 failed in Tests

that I haven't debugged yet, which really might be what is failing in CI.

The allocater/free problem shows as a break into the debugger with free.
Perhaps outside of a debugger it is tolerated and no failure.

jaykrell:
Actually that is maybe just expected, not an error. I need to look more.

jaykrell:
Yes, indeed.

jaykrell:
So I think the problem is the allocate/free mismatch, fixed by https://github.com/mono/mono/pull/11899. Let's see how CI fairs with that.

jaykrell:
I can get gcc to agree with msvc by unioning in a long long.
mono still  doesn't agree, because  !pinvoke.
[...]

PR #9503, opened 2018-07-11
[Coop] Cominterop and marshal conversion.

jaykrell:

PR #11899, opened 2018-12-02
Fix https://github.com/mono/mono/issues/11898

jaykrell:
Fix https://github.com/mono/mono/issues/11898
This regressed six months ago at https://github.com/mono/mono/pull/9503/files#r238093574.

There are two identical sounding functions:

mono_string_handle_to_utf8:
  What we usually use. Sounds reasonable. Usually is reasonable.
  Allocates with malloc.

mono_string_to_utf8str_handle:
  What should be used here. Allocates with CoTaskMemAlloc.
  So it can be freed with marshal_free => CoTaskMemFree.

i.e. The allocator and the freeer need to agree.

### DIFF
--- a/mono/metadata/marshal-internals.h
+++ b/mono/metadata/marshal-internals.h
@@ -14,7 +14,7 @@ MonoObjectHandle
 mono_marshal_xdomain_copy_value_handle (MonoObjectHandle val, MonoError *error);
 
 // On Windows platform implementation of bellow methods are hosted in separate source file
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
Let's see it pass again.
@monojenkins build Windows  i386

jaykrell:

jaykrell:
Changed a bit since approval, in order to build on Unix, but still addressing the same problem.

jaykrell:
@monojenkins build failed

PR #11906, opened 2018-12-03
"Imperial Red": Bring HttpClient from CoreFx

baulig:
This brings `System.Net.Http` from CoreFx.

It is currently only enabled on Desktop Mac and Linux; Mobile is coming shortly.


baulig:
@monojenkins build failed

baulig:
@monojenkins build BCL Size Diff

baulig:
@monojenkins commit csproj

baulig:
@monojenkins commit csproj

baulig:
@monojenkins commit csproj

baulig:
@monojenkins build OS X i386

baulig:
We still need to land the API Diff changes before this can go in.

You can review them in a separate PR here: https://github.com/mono/api-snapshot/pull/4.

mcs/class/System.Net.Http/corefx/PlatformDetection.cs
```diff
@@ -0,0 +1,15 @@
+namespace System
```
marek-safar: Don't create a new file but use https://github.com/mono/mono/blob/master/mcs/class/test-helpers/PlatformDetection.cs instead

mcs/class/System.Net.Http/monodroid_System.Net.Http.dll.sources
```diff
@@ -1,2 +1,2 @@
-#include System.Net.Http.dll.sources
+#include legacy.sources
```
marek-safar: monotouch, monodroid and wasm have to use new version (it can be done in follow up PR)

runtime/mono-test.sh
```diff
@@ -22,6 +22,20 @@ echo "--------------------------------------------------------------------------
...
+    "Linux")
+        mkdir -p ~/.config/.mono/
+        wget -qO- https://download.mono-project.com/test/new-certs.tgz| tar zx -C ~/.config/.mono/
+        ls -l ~/.config
+        ls -l ~/.config/.mono
+        ls -laR ~/.config/.mono
+        ;;
+esac
+
+echo "MARTIN TEST #1"
```
marek-safar: Some logging leftover?

runtime/mono-test.sh
```diff
@@ -22,6 +22,20 @@ echo "--------------------------------------------------------------------------
 "${MONO_EXECUTABLE}" --version
 echo "---------------------------------------------------------------------------------------"
 
+echo "MARTIN TEST!"
```
marek-safar: Some logging leftover?

mcs/class/System.Net.Http/corefx/NetEventSource.Http.cs
```diff
@@ -0,0 +1,57 @@
+using System.Diagnostics.Tracing;
+using System.Net.Http;
+
+namespace System.Net
+{
+	internal sealed partial class NetEventSource : EventSource
+	{
+		public static readonly NetEventSource Log = new NetEventSource ();
+
+		public static new bool IsEnabled => false;
```
marek-safar: This should not be needed and could break linker

mcs/class/System.Net.Http/corefx/NetEventSource.Http.cs
```diff
@@ -0,0 +1,57 @@
+using System.Diagnostics.Tracing;
+using System.Net.Http;
+
+namespace System.Net
+{
+	internal sealed partial class NetEventSource : EventSource
```
marek-safar: Why is it partial?

mcs/class/System.Net.Http/corefx/Configuration.Certificates.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            private const string CertificatePassword = "testcertificate";
+            private const string TestDataFolder = "TestData";
+
+            private static readonly Mutex m;
+            private const int MutexTimeout = 120 * 1000;
+
+            static Certificates()
+            {
+                if (PlatformDetection.IsUap)
```
marek-safar: Seems to be copied from somewhere?

mcs/class/System.Net.Http/UnitTests/Makefile
```diff
@@ -0,0 +1,17 @@
+thisdir = class/System.Net.Http/UnitTests
```
marek-safar: Why this cannot be in System.Net.Http Makefile?

mcs/class/System.Net.Http/MonoWebRequestHandler.cs
```diff
@@ -0,0 +1,593 @@
...
+using System.Threading.Tasks;
+using System.Collections.Specialized;
+using System.Net.Http.Headers;
+using System.Net.Cache;
+using System.Net.Security;
+using System.Linq;
+
+namespace System.Net.Http
+{
+	class MonoWebRequestHandler : IMonoHttpClientHandler
```
marek-safar: This is the original HttpClientHandler, right? Can we please keep only a single copy of the file?

mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Collections.Generic;
+using System.Net.Security;
+using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Net.Http
+{
+	interface IMonoHttpClientHandler : IDisposable
```
marek-safar: What is the purpose of this once used only internal interface?

mcs/class/System.Net.Http/legacy.sources
```diff
@@ -1,11 +1,14 @@
...
+IMonoHttpClientHandler.cs
+MonoWebRequestHandler.cs
+HeaderUtils.Legacy.cs
 System.Net.Http/ByteArrayContent.cs
 System.Net.Http/ClientCertificateOption.cs
 System.Net.Http/DelegatingHandler.cs
 System.Net.Http/FormUrlEncodedContent.cs
 System.Net.Http/HttpClient.cs
-System.Net.Http/HttpClientHandler.cs
+# System.Net.Http/HttpClientHandler.cs
```
marek-safar: Leftover

mcs/class/System.Net.Http/corefx/PlatformDetection.cs
baulig: No, this is intentionally in a new file by design.

It contains a few definitions that I definitively do not want to see outside the very limited context of the `System.Net.Http` tests.  This includes especially `OpenSslVersion`, `SupportsX509Chain` and `SupportsCertRevocation` (which in turn uses `UsingBtls`).

I want to keep tight control over these and - to facilitate a potential future removal / replacement with a more robust mechanism - prevent them from spreading all over the place.  This is best achieved by keeping them in a separate file.

I added a comment to the file.

mcs/class/System.Net.Http/corefx/NetEventSource.Http.cs
baulig: The property is required - but we might be able to change it's definition into

```
public static NetEventSource Log => throw new PlatformNotSupportedException ();
```

Would that help?

mcs/class/System.Net.Http/corefx/Configuration.Certificates.cs
baulig: It is, but the implementation has been modified.  The entire `GetResourceData()` method is new as well as our way of bundling the files as embedded resources.

They require the files to be available on disk and have some PowerShell script to do that.

mcs/class/System.Net.Http/MonoWebRequestHandler.cs
baulig: It is - and I only see one copy of the file, where is the other one?

mcs/class/System.Net.Http/corefx/NetEventSource.Http.cs
baulig: Looks like it's passing the tests, so I'm going to change that.

baulig:
Addressed feedback and switched to live API Snapshot (this will now cause the "API Diff" step to fail).

mcs/class/System.Net.Http/UnitTests/Makefile
baulig: The `UnitTests` certainly cannot be in the top-level `Makefile`.  They are using a "mock" implementation and not the "real" one - so they need to explicitly *not* link against the `System.Net.Http` assembly; which means they need to be kept in a separate assembly from the other tests.

We could, in theory, move the `FunctionalTests` tests into the top-level Makefile, though I think it's cleaner to have them both in subdirectories.

baulig:
Regarding mobile, this is essentially blocked on xUnit test support being added to `sdks/ios` because without that, we wouldn't have a way of testing it.  I believe the iOS team is working on this and they have an open PR for it somewhere, which hasn't been merged yet.

We will also need to ensure that the custom handlers for mobile (like for instance `CFNetworkHandler`) still work with the new code.

mcs/class/System.Net.Http/corefx/PlatformDetection.cs
marek-safar: PlatformDetection is used for xunit tests only which are fixed at this point. I think it's more useful to have them in one place than spread everywhere.

mcs/class/System.Net.Http/UnitTests/Makefile
marek-safar: The tests are always in separate assembly, so it should not matter. Where do you build the mock assembly?

@akoeplinger will this work with helix/sdks packaging?

mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
marek-safar: I don't understand how using IMonoHttpClientHandler instead of MonoWebRequestHandler ensures that we do not pull in any dependencies from the old handler. It's just indirection with the same dependencies.

baulig: Look at what's in the `MonoWebRequestHandler` implementation that's not in the interface.  You won't find any reference to any of `HttpWebRequest`, `HttpWebResponse` or `ServicePointManager` in the interface - nor are any of these classes used by the `SocketsHttpHandler`.

So when using the new handler, the `System.Net.Http.dll` that we build won't reference any of the web stack from `System.dll` - only `System.Net.Http.WebRequest.dll` will do so.  This makes it easy for the linker to completely remove the web stack on mobile.

During some of my earlier experiments with mobile, the linker failed to remove the traditional web stack when I did something like this:

```
    readonly MonoWebRequestHandler  _monoHandler;
    readonly SocketsHttpHandler _socketsHttpHandler;
    ClientCertificateOption _clientCertificateOptions;

    internal HttpClientHandler (MonoWebRequestHandler  monoHandler)
    {
        if (monoHandler != null) {
            _monoHandler = monoHandler;
        } else {
            _socketsHttpHandler = new SocketsHttpHandler ();
            ClientCertificateOptions = ClientCertificateOption.Manual;
        }
    }
```

Setup like this, you have a public class (`HttpClientHandler`) with an internal constructor referencing some class containing

```
	internal virtual HttpWebRequest CreateWebRequest (HttpRequestMessage request)
```

Whereas with the interface, there won't be any reference to `HttpWebRequest` and friends at all, so the linker should see that there is some interface, but no class implementing it.

marek-safar: I think the IMonoHttpClientHandler has the potential to make the platform specific handling easier but you didn't implement it. You made webrequest only possibly optional but at the same time the socket one mandatory which means for  webrequest mode we cannot remove the new code.

I'd propose to finish this in the way that we have a single version of HttpClientHandler which only delegates using IMonoHttpClientHandler. Basically what you have today in HttpClientHandler.Legacy.cs, then add IMonoHttpClientHandler interface to all handlers we have that way we can support as many as we need for mobile/wasm with no size increases.

marek-safar:
@baulig Both XI and XA can run xunit tests already, so there is nothing blocking

baulig:
At the moment, the only way of testing this with XI / XM is to manually bump the `xamarin-macios` module to the latest Mono, hope that it builds, then run their tests via `xamarin-macios/tests`.

What I was referring to above is the planned adding of xUnit support to `mono/sdks` and then having them run automatically on CI.

Considering the huge overhead and time investment of just testing a single commit via the `xamarin-macios/tests` method, I believe it would make more sense to finish the `mono/sdks` approach instead - or at least investigate what it's current state is, who's currently working on it and what's blocking it's completion.

baulig:
To summarize the discussion that we had about the interface - the fundamental problem was that the `HttpClientHandler` does more than just calling either or two implementations; several of it's public API entry points had additional code for the socket handler case.  Since the sockets handler is the more "modern" implementation (that will eventually become the default) and the "legacy" implementation will eventually be removed - it does not make any sense to base an interface implementation on the "legacy" version.

The correct approach is to make the `HttpClientHandler` implementation use sockets handler semantics - and then design the interface around those.

After a closer look at those differences, it turns out that most of those were actually newer additions in the socket handler that were never implemented in the legacy handler - and in all those cases, the interface design should follow the new implementation.

- Several security related properties have been replaced by `SslClientAuthenticationOptions`.
- There is no `UseDefaultCredentials`; it's setting `Credentials` to either `CredentialCache.DefaultCredentials` or `null`.  However, this is also what `HttpWebRequest.UseDefaultCredentials` does, so we can safely remove the property.
- The new handler does not support `MaxRequestContentBufferSize` (the getter always returns 0, the setter does some error checking then discards the value); this required special handling.
- We need `CertificateHelper.GetEligibleClientCertificate ()`; this is the platform-specific implementation which we already include for unix / windows.

baulig:
I'm still not a big fan of having a Mono-specific interface (instead of having a switcher mechanism inside CoreFx), but this is now as good as it can go without requiring CoreFX changes.

There is now only one `HttpClientHandler` implementation and the same file is used in all profiles.  The default constructor has been moved out into a `partial` class.  This means that the `HttpClientHandler.cs` does not reference any of the handlers.

The legacy handler has been slightly adjusted to use above mentioned properties.

mcs/class/System.Net.Http/Test/MartinTest.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using MonoTests.Helpers;
+using NUnit.Framework;
+
+namespace MonoTests.System.Net.Http
+{
+	[TestFixture]
+	class MartinTest
```
marek-safar: Looks like some testing leftover

baulig: Somehow, these keep coming in again, thought I removed it.

baulig:

baulig:
All tests passed previously.  I have taken out the submodule changes - both api-diff and the accidental linker bump.

marek-safar:
@baulig you still need to bump corefx. It's failing because it's missing partial modifier change

baulig:
@monojenkins build

baulig:
Commit 8a1db94 was good, then bumped corefx.

baulig:
@monojenkins build failed

mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
lewing: That doesn't match my understanding. Looking at 
 https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.7.2#httpclient-and-net-core it appears the that HttpClientHandler is supposed continue to use the old request stack and that SocketsHttpHandler is supposed to be swapped in at the client level.

Also notice that calling `AppContext.SetSwitch("System.Net.Http.useSocketsHttpHandler", false);` is supposed to use System.Net.Http.WebRequest in netcore 2.1 forward

marek-safar:
@monojenkins build failed

marek-safar:
@baulig it still fails to build for xammac

marek-safar:
@monojenkins commit apidiff

PR #11909, opened 2018-12-03
[crash] Early exit if callbacks not set

alexanderkyte:
When a crash happens too early in initialization for the crash reporter to operate, exit gracefully

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -5740,28 +5740,30 @@ ves_icall_Mono_Runtime_ExceptionToState (MonoExceptionHandle exc_handle, guint64
 	MonoStringHandle result;
 
 #ifndef DISABLE_CRASH_REPORTING
-	// FIXME: Push handles down into mini/mini-exceptions.c
-	MonoException *exc = MONO_HANDLE_RAW (exc_handle);
-	MonoThreadSummary out;
-	mono_get_eh_callbacks ()->mono_summarize_exception (exc, &out);
+	if (mono_get_eh_callbacks ()->mono_summarize_exception) {
+		// FIXME: Push handles down into mini/mini-exceptions.c
+		MonoException *exc = MONO_HANDLE_RAW (exc_handle);
+		MonoThreadSummary out;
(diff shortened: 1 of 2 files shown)


mono/metadata/icall.c
```diff
@@ -5740,26 +5740,30 @@ ves_icall_Mono_Runtime_ExceptionToState (MonoExceptionHandle exc_handle, guint64
...
 #else
-	*portable_hash_out = 0;
-	*unportable_hash_out = 0;
-	result = mono_string_new_handle (mono_domain_get (), "", error);
+	{
+		*portable_hash_out = 0;
+		*unportable_hash_out = 0;
+		result = mono_string_new_handle (mono_domain_get (), "", error);
+	}
 #endif
```
lewurm: shouldn't the `#endif` move up and replace `#else`?

lewurm:
with this patch I got a hang previously during crash dumping:
```patch
diff --git a/mono/metadata/domain.c b/mono/metadata/domain.c
index 667f7875313..1a342b723ae 100644
--- a/mono/metadata/domain.c
+++ b/mono/metadata/domain.c
@@ -498,6 +498,8 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
 	const MonoRuntimeInfo* runtimes [G_N_ELEMENTS (supported_runtimes) + 1] = { NULL };
 	int n;
 
+	g_assert (0);
+
 #ifdef DEBUG_DOMAIN_UNLOAD
 	debug_domain_unload = TRUE;
 #endif
```
this PR fixes it 👍

mono/metadata/icall.c
alexanderkyte: There, I moved the return up into the if block, and moved it all into a single if, with a fallthrough for the else.

alexanderkyte:
The linux wasm failure: ```checking whether the C compiler works... no```

alexanderkyte:

lewurm:
was this backported to `2018-08` and `2018-10`?

PR #11912, opened 2018-12-03
Fix order of parameter processing. 

nealef:
New parameter addition was put in wrong spot in the parameter list. Also the offsets on the stack were incorrect as 32-bit values are placed in 64-bit aligned slots on s390x. Therefore, even though these are gbooleans the appear to be treated as gulongs.


lewurm:
@monojenkins squash

lewurm:

lewurm:

PR #11914, opened 2018-12-04
More mgreg_t to host_mgreg_t or target_mgreg_t conversion.

jaykrell:
Mostly target_mgreg_t.
Note that this is all optional, in the sense that cross compilers
never run their runtime. So mgreg_t is always target_mgreg_t, or
they are the same.
But agreement is to ban mgreg_t.


jaykrell:
@monojenkins build failed

PR #11865, opened 2018-11-30
Fix cross compile of libtest.

jaykrell:
```
    CC       libtest_la-libtest.lo
    <inline asm>:37:2: error: instruction requires: Not 64-bit mode
            popl %eax
            ^
```
Same error many times.

PR #11859, opened 2018-11-30
Fix a bunch of warnings.

jaykrell:
```
/s/mono/mono/mini/interp/transform.c:1666:15: warning: unused variable 'vt' [-Wunused-variable]
                MonoVTable *vt = mono_class_vtable_checked (domain, target_method->klass, error);
                            ^
/s/mono/mono/utils/mono-counters.c:547:55: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'gssize' (aka 'int') [-Wformat]
                fprintf (outfile, ENTRY_FMT "%zd\n", counter->name, *(gssize*)buffer);
                                             ~~~                    ^~~~~~~~~~~~~~~~
                                             %td

/s/mono/mono/sgen/sgen-debug.c:1066:72: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'int' [-Wformat]
        fprintf (heap_dump_file, "<occupied offset=\"%zd\" size=\"%zd\"/>\n", start - section_start, end - start);
                                                     ~~~                      ^~~~~~~~~~~~~~~~~~~~~
                                                     %d
/s/mono/mono/sgen/sgen-debug.c:1066:95: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'int' [-Wformat]
        fprintf (heap_dump_file, "<occupied offset=\"%zd\" size=\"%zd\"/>\n", start - section_start, end - start);
                                                                  ~~~                                ^~~~~~~~~~~
                                                                  %d
/s/mono/mono/sgen/sgen-debug.c:1139:4: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'mword' (aka 'unsigned int') [-Wformat]
                        safe_object_get_size (obj));
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~
/s/mono/mono/sgen/sgen-debug.c:37:30: note: expanded from macro 'safe_object_get_size'
                                ^
/s/mono/mono/sgen/sgen-memory-governor.c:309:5: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'unsigned int' [-Wformat]
                                entry->promoted_size / 1024,
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~

/s/mono/mono/sgen/sgen-debug.c:1139:4: warning: format specifies type 'ssize_t' (aka 'long') but the argument has type 'mword' (aka 'unsigned long long') [-Wformat]
                        safe_object_get_size (obj));
[...]

PR #11917, opened 2018-12-04
Adjust test (not runtime) to workaround Windows i386 GCC ABI bug.

jaykrell:
https://gcc.gnu.org/bugzilla/show_activity.cgi?id=88352

Investigating my own PR breakage..I ended up building libtest with gcc and the runtime with msvc.
A runtime built with gcc would have behaved the same. Gcc has an ABI bug, that mono does not and probably should not implement, unless maybe gcc refuses to fix it.

Given:

libtest.c:
typedef struct {
        double d;
 } SingleDoubleStruct;

SingleDoubleStruct mono_test_marshal_return_single_double_struct (void);

Visual C++ and mono expect the return value in edx:eax.
Gcc returns it in ST0.
Changing struct to union fixes it.

Mono "actually" expects return value in local via pointer
in extra parameter, but it generates a wrapper to bridge
its custom (SysV?) ABI to the NT ABI.

Adjust the test to avoid the gcc bug.
The runtime is unchanged.

### DIFF
--- a/mono/tests/libtest.c
+++ b/mono/tests/libtest.c
@@ -5616,20 +5616,24 @@ mono_test_marshal_return_lpwstr (void)
 	return res;
 }
 
-typedef struct {
+typedef
+#if defined (HOST_WIN32) && defined (HOST_X86) && defined (__GNUC__)
+// Workaround gcc ABI bug. It returns the struct in ST0 instead of edx:eax.
+// Mono and Visual C++ agree.
+union
+#else
+struct
+#endif
+{
 	double d;
 } SingleDoubleStruct;
 
 LIBTEST_API SingleDoubleStruct STDCALL
 mono_test_marshal_return_single_double_struct (void)
 {
-	SingleDoubleStruct res;
-
-	res.d = 3.0;
-
+	SingleDoubleStruct res = {3.0};
 	return res;
 }
-
 
 #ifndef TARGET_X86
 


lateralusX:
CI runs test using msvc build mono runtime, setting MONO_EXECUTABLE to the msvc build runtime when building BCL and running tests. msvc build libtest.dll should be build into the same output directory as msvc build runtime, so it should be the one picked up when running the tests. If it picks up the gcc build libtest.dll I believe this is more of a CI issue (if we could fix this ABI difference it's great) since I believe we have agreed in the past not to mix and match msvc/gcc binaries due to various ABI differences over the years. Did you observe CI picking up gcc libtest.dll causing this failure or something you discovered locally?

jaykrell:
I not sure what CI does. I built msvc runtime and mingw tests & ran test manually. I was of the impression we never build tests with msvc.
I think but have to double check CI log shows this test failing.

jaykrell:
@monojenkins build failed

jaykrell:
Is configure -enable-msvc the recent change?

jaykrell:
We can’t fix the ABI problem. It is a gcc bug. Maybe something very creative??

jaykrell:
The workaround seems ok. It allows testing msvc. It can also be union unconditionally. The ifdef kinda provides extra place for a comment but can be omitted.

jaykrell:
It looks like gcc did fix the bug, very recently.

jaykrell:
I don’t think mixing binaries is the problem. I.e. I don’t think, but could double check, that mono ever implements the gcc form. Plus that gcc fixed the bug so you’d need to be careful as to gcc version.

jaykrell:
Also not clear how to disable, one test, within pinvoke2.

Disable all of pinvoke2? That would work but seems like overkill in terms of disabling. Unless Windows/x86 matters so little.

Split into pinvoke2 and pinvoke2_disable_win32 where pinvoke2-disable-win32  has just this test? That would also work.

Or leave in the workaround? And change ifdef much later to resume testing on Gcc 8.1 or later?

lateralusX:
So have you verified that this is acutally causing a problem on CI? Looking at the latest master builds, https://jenkins.mono-project.com/view/All/job/z/, their are no failure on Win32 and one failure on x64, not related to pinvoke2.exe.

CI runs all tests and builds using msvc build but still building both gcc and msvc, but sets MONO_EXECUTABLE so msvc runtime will be picked up and used:

https://github.com/mono/mono/blob/master/scripts/ci/run-jenkins.sh#L54
https://github.com/mono/mono/blob/master/scripts/ci/run-jenkins.sh#L55

So unless we can prove that this is causing a test failure that we can't fix, we should not disable the test.

jaykrell:
It fails locally 100% for a known reason. And gcc is consistent and incorrect.
I am not in favor of disabling the test, but working around the gcc problem.
Perhaps  we do not test the gcc variation? Still, the workaround is ok imho.

Here is a CI failure though, different failures:

https://jenkins.mono-project.com/job/w/17996/testReport/junit/MonoTests/runtime/pinvoke2_exe/
```
test_0_native_thiscall failed: got 1, expected 0
test_0_managed_thiscall failed: got 1, expected 0
```

jaykrell:
So, CI is flaky here? And the reasons aren't yet known. So  maybe  indeed disable. But just the failing ones? Need a repro and debug this other one. Perhaps it is caused by the PR?? Not likely but almost anything is possible.

jaykrell:
That actually might be related to the PR. Let me check something.
The double in struct should be worked around imho.

jaykrell:
Ok that is PR related indeed. My mistake.

jaykrell:
And  still probably  a  gcc bug..but different.

jaykrell:
Or a test bug.. looking..

jaykrell:
My mistake here somewhat.
There were other problems -- debug C runtime vs. cotaskmemalloc so debug runtime always failed pinvoke2.exe test, CI probably doesn't test that, it is fixed now.
And double in struct also does fail 100%  and should be fixed, due to gcc bug, maybe CI doesn't test that either for some reason.

But the problems in CI were probably self induced and the likely fix is here:
https://github.com/mono/mono/pull/11865

jaykrell:
This also explains https://github.com/mono/mono/pull/11859 which had that change. :(

jaykrell:
I don’t know why CI doesn’t see this problem but it repros locally consistently and makes sense & the workaround is trivial and lets the test run & pass. I suggest we take it.

lateralusX:
CI doesn't run the gcc, only msvc build and in the past mixing gcc and msvc build libraries has not been a prioritized configuration, but since you invested time on this and have a solution that workarounds the problem, I agree we could take it unless it regress anything. I assume gcc -> gcc still works with this fix as well as msvc->gcc (the thing you tested) and that msvc -> msvc is untouched (whats running on CI)?

jaykrell:
Notice there is no runtime change here.
Only the test is changed.
User code that resembles the former test, using gcc, did not work and still does not work.
User code that resembles the new test, using gcc, did work and still works.
Msvc worked and continues to work, either form. (well -- only one is tested, but assuming..)

It is a bug in gcc and is not reasonable to workaround in the runtime, until/unless gcc refuses to fix the bug and locks in the behavior for all time/versions. We can wait and see on that. They fixed it for amd64 recently so maybe they will fix it for x86 too.

PR #11925, opened 2018-12-04
Fix more crash data truncation.

jaykrell:


jaykrell:
This is probably moot with master now.

Issue #10331, opened 2018-08-28
System.Drawing.Common Facade needs implementation


marek-safar:
Right now our implementation for mobile profiles is empty https://github.com/mono/mono/blob/master/mcs/class/Facades/System.Drawing.Common/TypeForwarders.cs but the System.Drawing.Common nuget forwards to them.

Neither XI or XA have the types available but we should include what we can from System.Drawing and throw PNSE for the rest.

filipnavara:
Should the same thing be done for System.Drawing.Primitives? Currently the facades are patched in the Xamarin.Mac/iOS builds, but since there are the reference assemblies for Xamarin products now it would be possible to move it to Mono repository.

PR #11931, opened 2018-12-04
iOS impl of System.Drawing.Common based on mono/sysdrawing-coregraphics (#10331)

ntherning:
iOS impl of System.Drawing.Common based on mono/sysdrawing-coregraphics (#10331)

Also adds the appropriate System.Drawing.Common typeforwards for desktop profiles which have a System.Drawing implementation.

**NOTE!** WIP, do not merge.


filipnavara:
Wow, interesting concept. If it goes through I would like to follow with Mac version. Makes me wonder about the technical concept used for referencing Xamarin.iOS, which was always the technical roadblock for us

marek-safar:
@filipnavara we use reference assembly with API description only see https://github.com/ntherning/reference-assemblies/commit/792ebd610808dcac64b5dbd19ad7afbbd51a17fb

filipnavara:
In theory, could the same thing be done for Xamarin.Mac? If yes, then that would be interesting also for hosting XplatUICocoa WinForms backend directly in the Mono repository.

ntherning:
Added `xammac` profile support.

I have compared the IL of the `System.Drawing.dll` assembly built from `mono/sysdrawing-coregraphics` when compiled against the real `Xamarin.iOS.dll`/`Xamarin.Mac.dll` and the ones from my `mono/reference-assemblies` fork and the IL is identical. So it looks like I have managed to include everything that is necessary.

Next I will verify that the samples from `mono/sysdrawing-coregraphics` work as expected when compiled and run against the `System.Drawing.Common.dll` facade rather than `System.Drawing.dll`.

ntherning:
@marek-safar Can I push my `mono/reference-assemblies` changes to a branch in that repo? Right now I have a fork but if I can get a branch in the upstream repo I don't have to refer to my fork in `.gitmodules` in this PR.

akoeplinger:
@ntherning yes go ahead.

ntherning:
@akoeplinger I don't seem to have write permissions to https://github.com/mono/reference-assemblies.

akoeplinger:
@ntherning I don't have admin permissions either so I can't grant you write access, please talk to release engineering.

ntherning:
I have run some of the samples from mono/sysdrawing-coregraphics against the System.Drawing.Common.dll facade rather than System.Drawing.dll on Mac and in the iOS simulator. The produce the same results as the original samples. The rendering in some of the samples is a bit wonky (e.g. canvas is too large, viewport is flipped) but as this is the same as in the original samples this is not due to my work in this PR.

marek-safar:
@monojenkins build failed

PR #11938, opened 2018-12-05
[jit] Rename the textual names of several instructions so they match the opcode names.

vargaz:


vargaz:
@monojenkins build failed

PR #11941, opened 2018-12-05
[Coop] Mostly convert 30+ JIT icalls (and embedding API).

jaykrell:
[Coop] Mostly convert 30+ icalls used via register_icall.
And infrastructure to support that.
The infrastructure is not as ideal as it could be.
In future the C preprocessor wrappers should be marshal-ilgen wrappers,
so the handles are more efficiently created.

Some of the conversion is incomplete and just goes back to raw pointers,
but mostly it is good.

Some of the conversion is by replacing previously manual conversion
with more automatic conversion, but most of it is new.

Some is embedding API, not JIT icalls.

### DIFF
--- a/mono/metadata/boehm-gc.c
+++ b/mono/metadata/boehm-gc.c
@@ -40,7 +40,7 @@
 #include <mono/utils/mono-counters.h>
 #include <mono/utils/mono-compiler.h>
 #include <mono/utils/unlocked.h>
-
+#include <mono/metadata/icall-decl.h>
 
 #if HAVE_BOEHM_GC
 
@@ -1177,7 +1177,7 @@ create_allocator (int atype, int tls_key, gboolean slowpath)
  always_slowpath:
 	if (atype == ATYPE_STRING) {
 		mono_mb_emit_ldarg (mb, 1);
-		mono_mb_emit_icall (mb, ves_icall_string_alloc);
+		mono_mb_emit_icall (mb, ves_icall_string_alloc_raw);
 	} else {
 		mono_mb_emit_ldarg (mb, 0);
 		mono_mb_emit_icall (mb, ves_icall_object_new_specific);
--- a/mono/metadata/boehm-gc.c
+++ b/mono/metadata/boehm-gc.c
@@ -1177,7 +1177,7 @@ create_allocator (int atype, int tls_key, gboolean slowpath)
  always_slowpath:
 	if (atype == ATYPE_STRING) {
 		mono_mb_emit_ldarg (mb, 1);
-		mono_mb_emit_icall (mb, ves_icall_string_alloc_raw);
+		mono_mb_emit_icall (mb, ves_icall_string_alloc);
 	} else {
 		mono_mb_emit_ldarg (mb, 0);
 		mono_mb_emit_icall (mb, ves_icall_object_new_specific);
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -401,7 +401,7 @@ emit_ptr_to_object_conv (MonoMethodBuilder *mb, MonoType *type, MonoMarshalConv 
 		mono_mb_emit_byte (mb, CEE_LDIND_REF);
 		mono_mb_emit_ldloc (mb, 0);
 		mono_mb_emit_icon (mb, mspec->data.array_data.num_elem);
-		mono_mb_emit_icall (mb, mono_byvalarray_to_byte_array_raw);
+		mono_mb_emit_icall (mb, mono_byvalarray_to_byte_array);
 		break;
 	}
 	case MONO_MARSHAL_CONV_STR_BYVALSTR: 
@@ -409,11 +409,11 @@ emit_ptr_to_object_conv (MonoMethodBuilder *mb, MonoType *type, MonoMarshalConv 
 			mono_mb_emit_ldloc (mb, 1);
 			mono_mb_emit_ldloc (mb, 0);
 			mono_mb_emit_icon (mb, mspec->data.array_data.num_elem);
-			mono_mb_emit_icall (mb, mono_string_from_byvalstr_raw);
+			mono_mb_emit_icall (mb, mono_string_from_byvalstr);
 		} else {
 			mono_mb_emit_ldloc (mb, 1);
 			mono_mb_emit_ldloc (mb, 0);
-			mono_mb_emit_icall (mb, ves_icall_string_new_wrapper_raw);
+			mono_mb_emit_icall (mb, ves_icall_string_new_wrapper);
 		}
 		mono_mb_emit_byte (mb, CEE_STIND_REF);		
 		break;
@@ -422,11 +422,11 @@ emit_ptr_to_object_conv (MonoMethodBuilder *mb, MonoType *type, MonoMarshalConv 
 			mono_mb_emit_ldloc (mb, 1);
 			mono_mb_emit_ldloc (mb, 0);
 			mono_mb_emit_icon (mb, mspec->data.array_data.num_elem);
-			mono_mb_emit_icall (mb, mono_string_from_byvalwstr_raw);
+			mono_mb_emit_icall (mb, mono_string_from_byvalwstr);
 		} else {
 			mono_mb_emit_ldloc (mb, 1);
 			mono_mb_emit_ldloc (mb, 0);
-			mono_mb_emit_icall (mb, ves_icall_mono_string_from_utf16_raw);
+			mono_mb_emit_icall (mb, ves_icall_mono_string_from_utf16);
 		}
 		mono_mb_emit_byte (mb, CEE_STIND_REF);		
 		break;		
@@ -435,9 +435,9 @@ emit_ptr_to_object_conv (MonoMethodBuilder *mb, MonoType *type, MonoMarshalConv 
 		mono_mb_emit_ldloc (mb, 0);
 		mono_mb_emit_byte (mb, CEE_LDIND_I);
 #ifdef TARGET_WIN32
-		mono_mb_emit_icall (mb, ves_icall_mono_string_from_utf16_raw);
+		mono_mb_emit_icall (mb, ves_icall_mono_string_from_utf16);
 #else
-		mono_mb_emit_icall (mb, ves_icall_string_new_wrapper_raw);
+		mono_mb_emit_icall (mb, ves_icall_string_new_wrapper);
 #endif
 		mono_mb_emit_byte (mb, CEE_STIND_REF);	
 		break;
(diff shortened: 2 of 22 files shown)


jaykrell:
1. Failures need to be looked at.
2. Perf needs to be considered, but there were already managed/native transitions, and this doesn't add any. It does add TLS use and coop handle allocate/free, which could be improved, but isn't improved here.

jaykrell:
Embedding API may be here too, not just managed/native transitions.

mono/metadata/boehm-gc.c
```diff
@@ -40,7 +40,7 @@
 #include <mono/utils/mono-counters.h>
 #include <mono/utils/mono-compiler.h>
 #include <mono/utils/unlocked.h>
-
+#include "icall-decl.h"
```
luhenry: Follow the convention _right_ above.

jaykrell:

jaykrell:
Too much extra diff here, will reduce. And maybe can do a more complete conversion.

mono/metadata/marshal.c
```diff
@@ -691,7 +675,7 @@ mono_array_to_lparray (MonoArray *array)
 		nativeArray = (void **)g_malloc (sizeof(gpointer) * nativeArraySize);
```
jaykrell: Consider `malloc0` so that `break` below does not leave uninitialized.
Or remove the `break`.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build  failed

jaykrell:

jaykrell:
@monojenkins build  Linux x64 Acceptance Tests

jaykrell:
@monojenkins build failed

lambdageek:
> There is something suspicious about the most/entire thing, that I realized later that I will call out explicitly. This is largely/all p/invoke marshalling, and for example, marshaling a string returns a raw pointer to the characters. This doesn't fit the usual framework of coop conversion. There maybe more similar problems. This PR helps, but ultimately might hit blockages.
I think marshaling a string needs to fixed it.

Yea, I think there's a little bit missing from marshalling.  I made a work item: https://github.com/mono/mono/issues/12165

mono/metadata/marshal-ilgen.c
```diff
@@ -401,19 +401,19 @@ emit_ptr_to_object_conv (MonoMethodBuilder *mb, MonoType *type, MonoMarshalConv
 		mono_mb_emit_byte (mb, CEE_LDIND_REF);
 		mono_mb_emit_ldloc (mb, 0);
 		mono_mb_emit_icon (mb, mspec->data.array_data.num_elem);
-		mono_mb_emit_icall (mb, mono_byvalarray_to_byte_array);
+		mono_mb_emit_icall (mb, mono_byvalarray_to_byte_array_raw);
```
lambdageek: I wonder about the usability of this approach.  So I say `MONO_HANDLE_REGISTER_ICALL (mono_byvalarray_to_byte_array, ...)` and then I get to define `mono_byvalarray_to_byte_array`, but then when I go to write a _call_ I have to write `mono_byvalarray_to_byte_array_raw` or else I will get mysterious failures.

Maybe it's better if `mono_byvalarray_to_byte_array` was the name of the wrapper, and the implementation went into `mono_byvalarray_to_byte_array_impl` or something like that.

At least in that case if I make a mistake and write my own definition of `mono_byvalarray_to_byte_array`, I will get either a compile or link-time error, which seems better than a crash at runtime.

Does that make sense?

mono/metadata/marshal.c
```diff
@@ -969,63 +923,62 @@ mono_string_builder_to_utf8 (MonoStringBuilder *sb)
...
-			g_assert (chunk->chunkOffset >= 0);
-			g_assertf (chunk->chunkLength + chunk->chunkOffset <= len, "A chunk in the StringBuilder had a length longer than expected from the offset.");
-			memcpy (str + chunk->chunkOffset, source, chunk->chunkLength * sizeof (gunichar2));
+			MONO_HANDLE_GET (chunkChars, chunk, chunkChars);
+			const int chunkOffset = MONO_HANDLE_GETVAL (chunk, chunkOffset);
+			g_assert (chunkOffset >= 0);
+			g_assertf ((chunkOffset + chunkLength) >= chunkLength, "integer overflow");
+			g_assertf ((chunkOffset + chunkLength) <= len, "A chunk in the StringBuilder had a length longer than expected from the offset.");
+			gchandle_t gchandle = 0;
+			memcpy (str + chunkOffset, MONO_ARRAY_HANDLE_PIN (chunkChars, gunichar2, 0, &gchandle), chunkLength * sizeof (gunichar2));
```
lambdageek: this (and a couple others, but especially this one, since it loops) could be `NO_SAFEPOINTS`

vargaz:
It would be better to declare some kind of constant and pass that to emit_icall () which would do a lookup, instead of the current weird code which does a reverse lookup from the c function to the icall info.

jaykrell:
Agreed with both.

1. The names can be either way, we could rename both even, and it is error prone asis agreed -- I repeatedly forgot to add _raw on the register, and it compiles either way.

2. Can each reference just be to a struct by pointer with all the information together?
No arrays? The arrays are linker-unfriendly, the linker cannot remove any icalls.
If it can't go together then yes an integer indexing into an array is the best lookup.
Can this wait? There is no change in this regard either way in this PR.

jaykrell:
I can do it, but it is kinda an inconsistent mess to have JITed code call foo_raw, that calls foo, but the JIT to emit calls to foo, that call foo_impl.
Perhaps switch the others.
But all the C code was implementing foo, and sometimes calling it.
It is true the emit to foo instead of foo_raw was error prone, I made that mistake a few times.

jaykrell:
Is this the lookup to avoid:

```
		case MONO_CEE_MONO_ICALL: {
			gpointer func;
			MonoJitICallInfo *info;

```


```
MonoJitICallInfo *
mono_find_jit_icall_by_addr (gconstpointer addr)
{
	MonoJitICallInfo *info;
	g_assert (jit_icall_hash_addr);

	mono_icall_lock ();
	info = (MonoJitICallInfo *)g_hash_table_lookup (jit_icall_hash_addr, (gpointer)addr);
	mono_icall_unlock ();

	return info;
}
```

jaykrell:
I replaced foo => foo_impl, foo_raw => foo.

I replaced the one looped case of pinning/gchandle with no_safepoints.

I haven't yet tried to remove the hash table uses but I think I understand.

With pinning the loop pins/unpins, maybe slow.
With no_safepoints we end up with an extended period of no gc. Maybe fast/rare enough?

I understand the raw-ness of the more direct icall calls, less marshaling, but wonder if just pinning all the toplevel parameters would be easier. It wouldn't help this case much, and, mostly coop handles aren't that difficult to deal with either granted.

jaykrell:
Also, removing these lookups help wasm? Completely fixing some problem or just a dent?

jaykrell:

jaykrell:
Is this correct and the intent?

Visit every use of MONO_PATCH_INFO_JIT_ICALL and MONO_PATCH_INFO_JIT_ICALL_ADDR and MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL. Their emission and handling.
Ignore MONO_PATCH_INFO_ICALL_ADDR_CALL.
The first three are register_jit_icall ("new" contents of icall-def.h here, and plenty more).

The last is metadata-driven icalls (prior contents of icall-def.h).
 Maybe this part can be improved too. Maybe.

The difference between MONO_PATCH_INFO_JIT_ICALL and MONO_PATCH_INFO_JIT_ICALL_ADDR is that MONO_PATCH_INFO_JIT_ICALL might have a wrapper, but MONO_PATCH_INFO_JIT_ICALL_ADDR never does. Is that all?

For the time being, retain all the hash tables and inserts into them, and the lookup functions.
However:
Provide alternative forms of mono_aot_get_direct_call_symbol, mono_aot_get_plt_symbol, etc., where the "data" is MonoJitICallInfo instead of a string.

Visit every call to register_jit_icall.
For every call, declare & allocate a static MonoJitICallInfo.
Introduce register_jit_icall_info. That takes the address of that thing.
And fills it in.
And puts it in the hash tables (at least for now).

Introduce forms of MONO_PATCH_INFO_JIT_ICALL and MONO_PATCH_INFO_JIT_ICALL_ADDR and MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL appended with "_INFO" on their ends.
(or _data or _desc; _data is almost my favorite, except it could imply emiting a reference to data, like instead of a function, so _info).
Instead of a string or code address, emit the address of the info.

If this is correct and successful, the hashtable reads and writes should actually go away.
And some but not many of the edits could go away.

jaykrell:
ps: *IF* I understand this, make it a separate PR? Can proceed somewhat in parallel with this PR, but ultimately would have a merge conflict.

jaykrell:
Perhaps this effort can be tied to this PR in some ways.
That is, perhaps the static data, can be const, and initialized by these macros.

Initially this work is only for icalls using handles, but the macros can have a form for handle-less/wrapper-less, or differently-wrapped icalls.

Was that the intent? i.e. don't convert all en masses, just take advantage of the churn on them to start?

jaykrell:
Removing/reducing the hash tables will take a while.

vargaz:
It was just an idea, its actually a lot of work that we probably don't want to do right now.

jaykrell:
I think this other change should go in a separate PR or a few:
 Eliminate heap allocation of `MonoJitICallInfo`.
 Reduce or eliminate hash table lookup of `MonoJitICallInfo`, via direct references to named globals.
  i.e. for `mono_foo_icall`, there should be `mono_foo_icall_jit_call_info`. register_icall and such will take the address of that.
 If hash table lookup eliminated, then eliminate the inserts and the the tables.

Every register_icall will gain a line in icall-def.h for this (or a different file).

Every helper register_icall, register_icall_with_wrapper, register_icall_no_wrapper, register_dyn_icall, etc. will gain a corresponding or replacement macro or macros, in order to be a readable diff, that represents the same small variety that there already is.

jaykrell:
Note that the registered icall set is not static. :(
This one stamps out new names and maybe new wrappers.
```
MonoJitICallInfo *
mono_get_array_new_va_icall (int rank)
{
	MonoMethodSignature *esig;
	char icall_name [256];
	char *name;
	MonoJitICallInfo *info;

	/* Need to register the icall so it gets an icall wrapper */
	sprintf (icall_name, "ves_array_new_va_%d", rank);

	mono_jit_lock ();
	info = mono_find_jit_icall_by_name (icall_name);
	if (info == NULL) {
		esig = mono_get_array_new_va_signature (rank);
		name = g_strdup (icall_name);
		info = mono_register_jit_icall_full (mono_array_new_va, name, esig, FALSE, NULL);
	}
	mono_jit_unlock ();

	return info;
}
```

jaykrell:
Regarding https://github.com/mono/mono/pull/11941#issuecomment-449421909,
https://github.com/mono/mono/pull/12188 is a good start at just not using the hashtables, but directly referencing the values instead of looking up keys, but for now leaving their contents alone, until/unless all lookups gone.

mono/metadata/marshal.c
```diff
@@ -828,46 +779,49 @@ mono_string_builder_new (int starting_string_length)
...
 {
-	gunichar2 *charDst = (gunichar2 *)sb->chunkChars->vector;
-	memcpy (charDst, text, sizeof (gunichar2) * string_len);
-
-	sb->chunkLength = string_len;
+	MonoArrayHandle chunkChars = MONO_HANDLE_NEW_GET (MonoArray, sb, chunkChars);
+	gchandle_t gchandle = 0;
+	memcpy (MONO_ARRAY_HANDLE_PIN (chunkChars, gunichar2, 0, &gchandle), text, sizeof (gunichar2) * string_len);
+	mono_gchandle_free_internal (gchandle);
+	MONO_HANDLE_SETRAW (sb, chunkLength, string_len);
```
jaykrell: This line incorrect, caught and fixed recently.

PR #11943, opened 2018-12-05
[wasm] Add browser test suite.

kjpou1:
The existing tests work with JavaScript Engines and do not have any way to test code that works only in the browser.   The JavaScript Engines do not have access to any browser API’s so functionality like the WasmHttpMessageHandler cannot be tested.

- Add a test suite that uses Headless Chrome and Node.js with Karma.
- Allows testing the browser API.


kjpou1:
@monojenkins build failed

sdks/wasm/tests/browser/http-spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+    it('RequestStream: should return size of Stream with streaming', (done) => {
+      //karmaHTML.httpspec.document gives the access to the Document object of 'http-spec.html' file
+      var _document = karmaHTML.httpspec.document;
+      
+      _document.Module.BINDING.call_static_method("[HttpTestSuite]TestSuite.Program:RequestStream", [true]).then(
+        (result) => 
+        {
+            //console.log("we are here: " + result);
+            assert.equal(result, 500000);
```
lewing: If we do something like this:
```suggestion
            var error = null;
            try {
              assert.equal(result, 500000);
            } catch (e) {
              error = e;
            }
            done (e);
```
I think the assertions will be properly surfaced to the test results

sdks/wasm/Makefile
```diff
@@ -233,7 +246,22 @@ run-all-%:
 	make -C . run-sm-$*
 	make -C . run-jsc-$*
 
+run-browser-tests: $(BROWSER_TEST)/.stamp-browser-test-suite
```
lewing: This target should probably be listed in ../README.md along with the others for wasm

sdks/wasm/tests/browser/http-spec.js
```diff
@@ -0,0 +1,168 @@
...
+            done();
+        },
+        (error) =>
```
lewing: Similarly something like
```suggestion
        (error) => done.fail(error)
```
might be good here

kg:
Is it feasible for us to do more of this testing using C#? I get the impression karma is an end-to-end integrated stack so we have no choice but to use js, but it's kind of a bummer to have an entire separate testing stack that needs to be manually integrated into our CI tooling.

sdks/wasm/tests/browser/http-spec.js
kjpou1: Will do

sdks/wasm/Makefile
kjpou1: I will be updating the ../README.md in a separate PR.  Was waiting on some feedback.

sdks/wasm/tests/browser/http-spec.js
kjpou1: Yes, that will be better.  Thanks for that.

kjpou1:
@kg We have most of the tests from the C# side.  The problem is that the WasmHttpMessageHandler needs access to the browser API's.  The current tests do not have that as they only test against the JavaScript engines.  Definitely can be discussed though.

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build

kjpou1:

Issue #11635, opened 2018-11-10
FileSystemWatcher problem in VSMac


Therzok:
## Steps to Reproduce

1. Create 200 FileSystemWatcher instances

### Current Behavior

One thread is created per FileSystemWatcher instance

### Expected Behavior

There should be something to prevent resource starvation in that scenario.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

## Stacktrace

```
Please paste the stack trace here if available.
```

## History

There is https://github.com/dotnet/corefx/issues/30600 open at corefx.

We need a short term solution of having a fix in Mono. The 2 suggested improvements in CoreFX repo work, but there's also a 3rd solution we've implemented in VSMac.

That is to use a PathTree structure that is modified every time a new FileSystemWatcher is requested on a path.

The tree emulates a filesystem tree into an actual in-memory tree. Every path sent in is processed and the following happens: every directory becomes a node, every path separator is an edge. Each node also tracks liveness, so we know whether a given directory is a node requested as a watcher directory or if it's been created just to be able to create the whole tree.

Then, the PathTree itself provides a really efficient algorithm to limit the number of directories to watch. The `Normalize` method will use a queue and a maximum number of nodes to watch.

The normalize method is a BFS that yields live nodes until the number of yielded nodes is smaller than the maximum one, or yields parent nodes when we would exceed that limit.

https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.FSW/PathTree.cs
https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.FSW/PathTreeNode.cs

Therzok:
Other solutions would be to make one thread run all the runloops for filesystemwatchers or have some kind of scheduler with more than 1 thread.

Or some way to use the FSEventStream overload which takes multiple root directories to watch.

PR #11945, opened 2018-12-05
Share CFRunLoop for FSEventStreams

MaximLipnin:
This is an attempt to solve the problem https://github.com/mono/mono/issues/11635 using the idea of sharing CFRunLoop for FSEventStreams.

For simplicity I've copied `external/corefx/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.OSX.cs` to `System/corefx/OSX`, added StaticWatcherRunLoopManager class, updated Start  and CancellationCallback methods.


marek-safar:
@MaximLipnin could you share any stats about the improvements, secondly we'll like to get this upstream so testing this locally is fine but the merge should go via corefx

MaximLipnin:
I'm closing it. Further work is in https://github.com/mono/mono/pull/12197

PR #11946, opened 2018-12-05
Windows x64 LLVM assert support + MSVC only for none cross compiler build.

lateralusX:
**Add support for LLVM asserts in internal LLVM build on Windows:**

Read and use of enable-llvm-asserts in autogen and include it in MSVC internal build LLVM.

**Only build LLVM using MSVC if not cross compiling:**

If not cross compiling, building runtime where host/target is Windows, only build and use LLVM as part of msvc build (only supported configuration), disabled on gcc build. This is needed so autogen.sh can be used to configure build using for example --enable-llvm, but only build on supported Windows configurations. NOTE, this will only be applied to none cross compiler builds since that LLVM build is currently x64, MSVC only.

Enforce 64-bit LLVM support when building using Visual Studio solution file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3814,6 +3814,27 @@ if test "x$enable_llvm" = "xdefault"; then
    enable_llvm=$enable_llvm_default
 fi
 
+enable_llvm_msvc_only="no"
+if test "x$enable_llvm" = "xyes"; then
+	if test "x$host_win32" = "xyes"; then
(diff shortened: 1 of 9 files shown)


lateralusX:
@luhenry Approve, merge?

directhex:
@luhenry please review this, I'm blocking on it.

PR #11947, opened 2018-12-05
[Win32] Enable BTLS in Windows builds

directhex:
Let's not merge this until we see what Jenkins thinks about the whole thing. There's a slight element of confusion around CMake - Cygwin CMake can't generate .vcxproj files, and which CMake you get given seems a bit non-obvious

### DIFF
--- a/packaging/Windows/defs/mono
+++ b/packaging/Windows/defs/mono
@@ -70,6 +70,9 @@ install()
 	cp -a msvc/build/boehm/Win32/bin/Release/mono-2.0-boehm.dll	${INSTALLDIR}/bin/
 	cp -a msvc/build/boehm/Win32/lib/Release/mono-2.0-boehm.lib	${INSTALLDIR}/lib/
 	cp -a msvc/build/boehm/Win32/bin/Release/mono-2.0-boehm.pdb	${INSTALLDIR}/lib/
+	if [ -e msvc/build/sgen/Win32/bin/Release/libmono-btls-shared.dll ]; then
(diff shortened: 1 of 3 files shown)


directhex:
Removing the do-not-merge label. I'm eyeballing the build logs of https://jenkins.mono-project.com/job/w/18030/console and https://jenkins.mono-project.com/job/x/16254/console, and I'm reasonably confident things are going well - `msbuild` built `libmono-btls-shared.dll`, Cygwin `make` built `Mono.Btls.Interface.dll`. The only real thing that hasn't been confirmed yet is whether `Mono.Btls.Interface.dll` is copied into the staging tree as expected by `make install`.

An undecided issue right now is how to seed the cert store. Not that we seeded the cert store at all on Windows, thinking about it :grimacing:

directhex:
Failures are unrelated.

@akoeplinger @lateralusX does this seem sane to you?

lateralusX:
If we would like to have full support for assembler optimization in CI BTLS build we will need to install ninja, perl and yasm, both for cygwin (if not already installed on CI) as well as in the environment seen by msbuild, this is documented here, https://www.mono-project.com/docs/compiling-mono/windows/#enable-btls-as-cryptographic-backend-for-windows-builds. I will change the stawberryperl into activeperl, so we should pick activeperl (since strawberryperl adds gcc to the path).

Current build works as is, but doesn't include the assembler optimization for some of the crytpo features in OpenSSL. since our build scripts needed to fallback to Visual Studio for BTLS build and BTLS build only support the optimization if ninja, perl and yasm can be found.

akoeplinger:
@lateralusX TBH I don't think crypto speed is the most pressing issue on Windows so I'd postpone that for now.

PR #11948, opened 2018-12-05
Fix regression with building PPC JIT

NattyNarwhal:
d5547b1929afa91944ba941d66a93767ca534a97 made an incorrect change to `mini-ppc.c` that wouldn't build; revert that file.


NattyNarwhal:
So should you/I make that opposite change instead, or is this OK?

vargaz:
I'd prefer the change the in the comment, it was just missing from the PR for some reason.

vargaz:
@monojenkins build failed

PR #11624, opened 2018-11-09
Add meaningful resource strings for System.Web.Services

mhutch:
Generated from mcs/class/referencesource/System.Web.Services/System.Web.Services.txt using resx2sr

Also adds support for txt files and custom class names to resx2sr

Fixes #11623

PR #11951, opened 2018-12-06
Revert "Add meaningful resource strings for System.Web.Services (#11624)"

baulig:
This reverts commit c2b6caf04d2c2ddac9e5d8e2a01edd9078668c85.


luhenry:
@baulig how about fixing it instead?

baulig:
I did not break it and since the author of that PR is refusing the do anything about it (after insulting me on Slack), I don't see particularly compelled to do anything about it.  But nevertheless, the tool is broken and it can't remain broken forever.

baulig:
So unless @mhutch fixes it, we will have to revert that PR.

PR #11952, opened 2018-12-06
Eliminate of mgreg_t in favor of target_mgreg_t and host_mgreg_t. And one minor fix.

jaykrell:
This is mostly cosmetic, but agreed to, since mono cross compilers never carry a runtime.
The most important thing is that x86 and arm backends use mgreg_t instead of gpointer, and MONO_ABI_SIZEOF instead of sizeof (for structs).

### DIFF
--- a/mono/utils/mono-machine.h
+++ b/mono/utils/mono-machine.h
@@ -31,18 +31,6 @@ typedef gssize host_mgreg_t;
 typedef gsize host_umgreg_t;
 #endif
 
-// SIZEOF_REGISTER is target. mgreg_t is usually target, sometimes host.
-// There is a mismatch in cross (AOT) compilers, and the
-// never executed JIT and runtime truncate pointers.
-// When casting to/from pointers, use gsize or gssize.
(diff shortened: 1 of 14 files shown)


jaykrell:
The completeness here can be guaged with  `git grep mgreg_t  | grep -v _mgreg_t` which shows essentially nothing.

mono/utils/mono-machine.h
```diff
@@ -38,9 +38,9 @@ typedef gsize host_umgreg_t;
 // When dealing with register context, use host_mgreg_t.
 // Or ifndef MONO_CROSS_COMPILE out runtime code.
 #if SIZEOF_REGISTER == 4
-typedef gint32 mgreg_t;
```
vargaz: This block can be removed.

jaykrell: Yeah, I wanted to preserve some part of the comment but was taking too long to rewrite it.
I had this on monojenkins squash so it went in asis (good or bad etiquette?).

PR #11964, opened 2018-12-07
PPC JIT optimizations (System.Math instruction inlining)

NattyNarwhal:
Optimizes a bunch of functions with expensive function call overhead into neat one or two instructions inlined. Optimizations are based on amd64 mini instruction emitting. Instructions used are still within the POWER5 optimizations applied.

`rcheck` in mini passes, and output seems sane from prodding code. Linux, PASE, and ppc32 are untested. Floating point I know can be.... delicate, but since amd64 tries to take the HW shortcuts, and IEEE floating point should be consistent, try this.

Functions affected:

* `System.Math`

  *  `Min (int32, int32)`

  *  `Min (uint32, uint32)`

  *  `Min (int64, int64)` (don't enable long int arith functions on ppc32 though, based on implications in mini for ppc32)

  *  `Min (uint64, uint64)`

  *  `Sqrt (double)`

  *  `Abs (double)`

  *  `Round (double)`

  *  `Truncate (double)`

  *  `Ceiling (double)`

  *  `Floor (double)`

* `System.MathF`

  *  `Sqrt (float)`

Other easy wins to add later: It'd be nice to add the double/float overloads for Min/Max working, (even if fsel is a bit wonky, compare-and-branch inlining could work?) integer modulo on POWER9, popcount/trailing zeroes on POWER9. (I don't have anything past POWER7 to test with.)

### DIFF
--- a/mono/arch/ppc/ppc-codegen.h
+++ b/mono/arch/ppc/ppc-codegen.h
@@ -754,6 +754,23 @@ my and Ximian's copyright to this code. ;)
 
 /* this marks the end of my work, ct */
 
+/* Introduced in Power ISA 2.02 (P4?) */
+#define ppc_frinx(c,D,B,Rc) ppc_emit32(c, (63 << 26) | (D << 21) | (0 << 16) | (B << 11) | (392 << 1) | Rc)
+#define ppc_frin(c,D,B) ppc_frinx(c,D,B,0)
+#define ppc_frind(c,D,B) ppc_frinx(c,D,B,1)
+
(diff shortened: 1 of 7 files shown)


marek-safar:
@lewurm could you please review

lewurm:

luhenry:
@NattyNarwhal is that still good to merge?

luhenry:
@monojenkins build

NattyNarwhal:
I think it should be fine, only caveat is I haven't tested it on Linux, nor have I tested on G5. I tried to check for the appropriate flags though, so...

PR #11967, opened 2018-12-07
[gensources]: Add platform/project logic to tests.

baulig:
The `gensources` tool now supports platform/specific `.sources` for nunit and xunit tests.

### DIFF
--- a/mcs/class/corlib/win32_net_4_x_corlib_test.dll.sources
+++ b/mcs/class/corlib/win32_net_4_x_corlib_test.dll.sources
@@ -1,2 +1,1 @@
 #include corlib_test.dll.sources
-#include unix_build_corlib.dll.sources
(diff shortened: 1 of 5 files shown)


baulig:
@monojenkins build

baulig:
@monojenkins build failed

mcs/class/corlib/win32_net_4_x_corlib_test.dll.sources
```diff
@@ -1,2 +1 @@
 #include corlib_test.dll.sources
-#include unix_build_corlib.dll.sources
```
baulig: This was added by accident in an earlier PR.  We need to take this out again because we now actually use the `win32_net_4_x_corlib_test.dll.sources` when building the tests.

baulig:
@monojenkins build

PR #11968, opened 2018-12-07
marshal.c cleanup

jaykrell:
- Reduce length computations of strings.
 - Handle zero sized StringBuilder correctly (initialize the first
   not just the second character).
 - Check for overflow correctly in StringBuilder -- include offset.
 - Fix prototype of mono_string_to_byvalstr (gpointer vs. char*).
 - Fix prototype of mono_string_to_byvalwstr (gpionter vs. gunichar2*).
 - Single point of exit in mono_string_builder_to_utf8.


jaykrell:
@monojenkins build failed

PR #11969, opened 2018-12-07
[resx2sr]: Giving this tool some love.

baulig:


baulig:
For future reference, this code here (in `System.Web.Services/Makefile`):

```
RESX_EXTRA_ARGUMENTS = \
	--in=corefx/SR.template.cs --name=System.Web.Services.Res --existing

RESX_RESOURCE_STRING = \
	../referencesource/System.Web.Services/System.Web.Services.txt
```

requires the `--in=` file to exist and it cannot be equivalent to the output file (because the output file is regenerated when the tool starts to run).

The `SR.template.cs` is not strictly required - it only serves as an example to test the `--existing` mode (because we currently do not have any option of testing that).

See also my comments at the top of the `resx2sr.cs` tool for details on testing.

baulig:
Also keep in mind that the tool has two modes of operation:

1.) The "standard" mode of operation is what we are using in `corlib`: we have a mixture of resource strings from the reference source, manually-generated strings and `.resx` files from corefx.  Here, we are using the `--in=` arguments to skip everything that's already in those first mentioned files.

2.) The "existing only" mode was used to update the resource strings for `System.Web.Services` - here we read an input file and replace all strings in it, but without adding any new key.

See my comments in the source code for details.

baulig:
@monojenkins commit csproj

PR #11970, opened 2018-12-07
[wasm] Allow creation of JavaScript Objects.

kjpou1:
- Allow creation of JavaScript objects without resorting to json parsing and creating hooks.
- Small cleanup of the bindings routines.
- Modify WasmHttpMessageHandler to use new routines and take out the hackish calls.


kjpou1:
@monojenkins build failed

kjpou1:

kjpou1:

sdks/wasm/bindings.cs
```diff
@@ -79,69 +104,68 @@ static int UnBindJSObject(int js_id)
...
-            else
-                bound_objects[js_id] = obj;
-
-            return (int)(IntPtr)obj.Handle;
-        }
-
-        static int UnBindJSObjectAndFree(int js_id)
+        static void UnBindJSObjectAndFree(int js_id)
         {
             if (bound_objects.ContainsKey(js_id))
```
lewing: Use TryGetValue like
```suggestion
            if (bound_objects.TryGetValue (js_id, out var obj)
```
To avoid an extra lookup here.  Also what is the new `bound_objects[js_id].RawObject = null;` looks redundant?

sdks/wasm/bindings.cs
```diff
@@ -79,69 +104,68 @@ static int UnBindJSObject(int js_id)
...
         }
 
         public static void FreeObject (object obj)
         {
-            int exception;
-            Runtime.ReleaseObject(obj, out exception);
-            if (exception != 0)
-                throw new JSException($"Error releasing object on (raw-obj)");
-   
+            if (raw_to_js.ContainsKey(obj))
```
lewing:
```suggestion
            if (raw_to_js.TryGetValue(obj, out JSObject jsobj))
```

sdks/wasm/bindings.cs
kjpou1: Yes you are correct with the TryGetValue.  Will make those changes.

If I do not null them out I get the following intermittently:

`Uncaught RuntimeError: memory access out of bounds`

sdks/wasm/bindings.cs
kjpou1: Here if I do the `TryGetValue` then I have to reference it again later with `raw_to_js[obj].RawObject = null;` anyway to circumvent an intermittent `Uncaught RuntimeError: memory access out of bounds` when GC runs.

kjpou1:
@monojenkins build failed

PR #11973, opened 2018-12-07
[Mono.Debugger.Soft] Fix reading custom attribute named arguments on old protocol versions

jbevain:
We were seeing random exceptions debugging an old protocol version, that would all happen after looking at Custom Attributes.

Turns out the client was always expecting the server to send a field id for named arguments in custom attributes. The server was fixed in https://github.com/mono/mono/commit/db0b932cd6c3c93976479ae3f6b5b2a885f681de without bumping the version, in between `2.11` and `2.12`

2.12 is then the only version in which can guarantee we can read named arguments for fields.

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -1665,7 +1665,15 @@ namespace Mono.Debugger.Soft
 					CattrNamedArgInfo arg = new CattrNamedArgInfo ();
 					int arg_type = r.ReadByte ();
 					arg.is_property = arg_type == 0x54;
-					arg.id = r.ReadId ();
+
+					// 2.12 is the only version we can guarantee the server will send a field id
+					// It was added in https://github.com/mono/mono/commit/db0b932cd6c3c93976479ae3f6b5b2a885f681de
+					// In between 2.11 and 2.12
+					if (arg.is_property)
+						arg.id = r.ReadId ();
+					else if (Version.AtLeast (2, 12))
+						arg.id = r.ReadId ();
+
 					arg.value = r.ReadValue ();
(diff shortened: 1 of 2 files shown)


mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
```diff
@@ -1665,7 +1665,16 @@ static String CommandString (CommandSet command_set, int command)
...
 					int arg_type = r.ReadByte ();
 					arg.is_property = arg_type == 0x54;
-					arg.id = r.ReadId ();
+
+					// 2.12 is the only version we can guarantee the server will send a field id
+					// It was added in https://github.com/mono/mono/commit/db0b932cd6c3c93976479ae3f6b5b2a885f681de
+					// In between 2.11 and 2.12
+					if (arg.is_property)  {
+						arg.id = r.ReadId ();
+					} else if (!arg.is_property && Version.AtLeast (2, 12)) {
```
vargaz: Extra braces. The !arg.is_property is not needed.

vargaz: Yes.

jbevain: Here you go sir.

sailro:
Please **do not merge yet**. We are currently testing this patch in our Unity workload CI

sailro:
@vargaz we find an issue and @jbevain pushed a fix. All good now on our side.

jbevain:
@vargaz we're good to merge our side, our test suite is green with this patch.

vargaz:
@monojenkins build failed

marek-safar:
@monojenkins backport 2018-12

marek-safar:
@monojenkins backport 2018-10

PR #11978, opened 2018-12-07
[2018-06] [crash] Force lldb/gdb traces to appear during hangs

alexanderkyte:
This PR creates a thread that will watch the crash reporter and send a fatal signal if it does not complete within a fixed timeout period. This fatal signal is interpreted as a nested fault by the crash reporter machinery, and it completes the (now non-failing) early exit and continues to dump.

Backport of #11636.

/cc @alexanderkyte


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #11989, opened 2018-12-09
for "build" config, import corlib common sources at unix instead

NattyNarwhal:
...of importing it at linux/macOS

this unbreaks the build on "unix" platforms like FreeBSD and AIX

(is this all needed files, and am I doing this right?)

### DIFF
--- a/mcs/class/corlib/linux_build_corlib.dll.sources
+++ b/mcs/class/corlib/linux_build_corlib.dll.sources
@@ -1,2 +1,1 @@
-#include corlib.dll.sources
 #include unix_build_corlib.dll.sources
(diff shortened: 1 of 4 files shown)


NattyNarwhal:
I think Alexis said he had a better idea, but that works for now. Need to get the CoreFX fix landed and hopefully we should be good.

jaykrell:
This will fix monojenkins build deb with monolite?
How to test that locally, just curious.

akoeplinger:
@jaykrell it should, watch https://jenkins.mono-project.com/job/build-source-tarball-mono/9881/ and if that turns green then monolite build will work too.

alexischr:
@NattyNarwhal nope that was exactly what I wanted to do (remove now-duplicate includes), thank you :)

PR #11991, opened 2018-12-09
[Coop] Convert System.String icalls.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/object.c
```diff
@@ -56,6 +56,7 @@
 #include <mono/utils/unlocked.h>
 #include "external-only.h"
 #include "monitor.h"
+typedef guint32 gchandle_t; // FIXME
```
lambdageek: This isn't going in.

mono/metadata/object.c
```diff
@@ -3726,13 +3727,19 @@ mono_get_constant_value_from_blob (MonoDomain* domain, MonoTypeEnum type, const
...
+
+	int result = -1;
+
 	if (!mono_metadata_read_constant_value (blob, type, value, error))
-		return -1;
+		goto exit;
 
 	if (type == MONO_TYPE_STRING)
-		*(gpointer*)value = mono_ldstr_metadata_sig (domain, *(const char**)value, error);
+		*(gpointer*)value = MONO_HANDLE_RAW (mono_ldstr_metadata_sig (domain, *(const char**)value, error));
```
lambdageek: This needs a FIXME (or perhaps hold off on changing `mono_get_constant_value_from_blob` until all of its callers are ready to change).

mono/metadata/object.c
```diff
@@ -7265,67 +7273,56 @@ mono_string_get_pinned (MonoString *str, MonoError *error)
...
+	const gsize size = MONO_SIZEOF_MONO_STRING + (length + 1) * sizeof (gunichar2);
+	MonoStringHandle news = MONO_HANDLE_CAST (MonoString, mono_gc_alloc_handle_pinned_obj (MONO_HANDLE_GETVAL (str, object.vtable), size));
+	if (!MONO_HANDLE_BOOL (news)) {
+		mono_error_set_out_of_memory (error, "Could not allocate %" G_GSIZE_FORMAT " bytes", size);
+		return news;
+	}
+	const gboolean pin = FALSE; // FIXME pin is slow and breaks profiler test.
+	gchandle_t str_gchandle = 0;
+	gchandle_t new_gchandle = 0;
+	memcpy (pin ? mono_string_handle_pin_chars (news, &new_gchandle)
```
lambdageek: no_safepoints instead of pinning?

mono/metadata/object-internals.h
```diff
@@ -1941,8 +1941,24 @@ mono_string_to_utf8_ignore (MonoString *s);
...
+ * \returns The interned string. On failure returns NULL and sets \p error
+ */
+#define mono_string_intern_checked(str, error) (mono_string_is_interned_lookup ((str), TRUE, (error)))
+
+/**
+ * mono_string_is_interned_internal:
+ * \param o String to probe
+ * \returns Whether the string has been interned.
+ */
+#define mono_string_is_interned_internal(str, error) (mono_string_is_interned_lookup ((str), FALSE, (error)))
```
lambdageek: shouldn't this be
```suggestion
 #define mono_string_is_interned_internal(str, error) (MONO_HANDLE_BOOL (mono_string_is_interned_lookup ((str), FALSE, (error))))
```

jaykrell: And  `ves_icall_System_String_InternalIsInterned`, at least first order.

mono/metadata/object.c
lambdageek: This isn't 
1. the right place for this typedef
2. the right PR to add this typedef

(also global handles are part of the API, so they'll --- at least in part --- need to stay as plain ints, so no destructors).

jaykrell: Appears unused in this file/PR. Will confirm/remove. I guess because of the no_safepoints vs. pin. Perhaps that is what you meant.  Agreed.

mono/metadata/object.c
jaykrell: Imho FIXMEcoop is best for now.

mono/metadata/object-internals.h
jaykrell: I agree is is confusing but this fits the prior pattern and the callers' prior and remaining requirements. Maybe rename somehow?

jaykrell: I agree is is confusing but this fits the prior pattern and the callers' prior and remaining requirements. Maybe rename somehow?

mono/metadata/object.c
jaykrell: They should have been pointer-sized for implementation flexibilty.
The name-for-documentation-purposes is somewhat useful internally even if stuck with the wrong type externally.
I added it, or expand its scope whenever I add pinning (it is present in some of the code), which was the case here but no longer.
But anyway no longer used here, removed.

mono/metadata/object.c
```diff
@@ -7281,16 +7280,15 @@ mono_string_get_pinned (MonoStringHandle str, MonoError *error)
...
 	}
-	const gboolean pin = FALSE; // FIXME pin is slow and breaks profiler test.
-	gchandle_t str_gchandle = 0;
-	gchandle_t new_gchandle = 0;
-	memcpy (pin ? mono_string_handle_pin_chars (news, &new_gchandle)
-		    : mono_string_chars_internal (MONO_HANDLE_RAW (news)),
-		pin ? mono_string_handle_pin_chars (str, &str_gchandle)
-		    : mono_string_chars_internal (MONO_HANDLE_RAW (str)),
+
+	MONO_ENTER_NO_SAFEPOINTS;
```
jaykrell: Static analysis needs to learn about this at some point, but plenty to do before then.

jaykrell: Actually it might just be suppress within the macro (and a scope). i.e. already  have trivial to  use building blocks.

jaykrell:
@monojenkins build  failed

jaykrell:

jaykrell:

jaykrell:

mono/metadata/object.c
jaykrell: I guess two names are warranted, maybe `gchandle_raw_t`  w/o destructor and `gchandle_t` with destructor. One can also play game of C and C++ reuse same name but probably bad idea.
(windows.h does this for example with REFIID and nobody seemed to mind except me -- pointer in C, reference in C++; also midl/com does this systematically, but still probably  not worthwhile, better to have the C  subset of  C++ mean the same thing and switching language of  compilation, when in the valid subset, not change meaning, except very subtely and harmlesslyl like our  templates and enum operators).

jaykrell:
@monojenkins build failed

mono/metadata/object.c
```diff
@@ -6585,25 +6593,30 @@ mono_string_new_size_checked (MonoDomain *domain, gint32 len, MonoError *error)
...
+	return_val_if_nok (error, NULL_HANDLE_STRING);
 
-	s = (MonoString *)mono_gc_alloc_string (vtable, size, len);
+	s = mono_gc_alloc_handle_string (vtable, size, len);
 
-	if (G_UNLIKELY (!s)) {
-		mono_error_set_out_of_memory (error, "Could not allocate %zd bytes", size);
-		return NULL;
-	}
+	if (G_UNLIKELY (MONO_HANDLE_IS_NULL (s)))
```
jaykrell: Consider pushing `MonoError` down a level but ok.

mono/metadata/handle.h
```diff
@@ -498,11 +498,12 @@ extern const MonoObjectHandle mono_null_value_handle;
 #define NULL_HANDLE_STRING (MONO_HANDLE_CAST (MonoString, NULL_HANDLE))
 #define NULL_HANDLE_ARRAY  (MONO_HANDLE_CAST (MonoArray,  NULL_HANDLE))
 
-static inline void
-mono_handle_assign (MonoObjectHandleOut dest, MonoObjectHandle src)
+static inline MonoObjectHandle
+mono_handle_assign_raw (MonoObjectHandle dest, void *src)
```
jaykrell: Restore `MonoObjectHandleOut` presumably.

jaykrell: Done.

jaykrell:
WebAssembly break is unrelated and fixed in master.

PR #11994, opened 2018-12-10
Inline mono_string_handle_length by making it a macro, usually.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
No good reason, granted.  Bad excuses: guaranteed inlining, safe because single evaluation of parameters. But no good reason.

PR #11997, opened 2018-12-10
Reduce lock hold times.

jaykrell:


jaykrell:
@monojenkins build failed

PR #12000, opened 2018-12-10
[corlib] Import System.IO.File from CoreFX

alexischr:
Contributes to #7246


marek-safar:
@alexischr please add corefx tests as well

marek-safar:
@monojenkins commit apidiff

PR #12003, opened 2018-12-10
[Coop] Mark System.ArgIterator icalls all NOHANDLES.

jaykrell:
If this is incorrect, static or compile-time checks will eventually catch it.


jaykrell:

jaykrell:

PR #12004, opened 2018-12-10
[Coop] Convert System.Buffer, with handle frames optimized away.

jaykrell:
That is, add a mechanism for icalls that usually do not create handles,
only want input handles.
MonoError is still provided, as that overhead is deemed small, but
we could provide an option to remove that also, i.e. for icalls that never fail.

### DIFF
--- a/mono/metadata/icall-table.h
+++ b/mono/metadata/icall-table.h
@@ -121,6 +121,7 @@ typedef MonoReflectionModuleHandle MonoReflectionModuleOutHandle;
 #define MONO_HANDLE_TYPE_WRAP_gint   			ICALL_HANDLES_WRAP_NONE
 #define MONO_HANDLE_TYPE_WRAP_gint32   			ICALL_HANDLES_WRAP_NONE
 #define MONO_HANDLE_TYPE_WRAP_gint64   			ICALL_HANDLES_WRAP_NONE
+#define MONO_HANDLE_TYPE_WRAP_gint8   			ICALL_HANDLES_WRAP_NONE
 #define MONO_HANDLE_TYPE_WRAP_gpointer   		ICALL_HANDLES_WRAP_NONE
 #define MONO_HANDLE_TYPE_WRAP_gconstpointer   		ICALL_HANDLES_WRAP_NONE
 #define MONO_HANDLE_TYPE_WRAP_gsize   			ICALL_HANDLES_WRAP_NONE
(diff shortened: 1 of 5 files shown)


jaykrell:

jaykrell:

mono/metadata/icall-table.h
```diff
@@ -236,10 +237,18 @@ typedef MonoReflectionModuleHandle MonoReflectionModuleOutHandle;
...
+#define MONO_HANDLE_RETURN_END_WITH_FRAME_ICALL_HANDLES_WRAP_NONE   	HANDLE_FUNCTION_RETURN_VAL (icall_result)
+#define MONO_HANDLE_RETURN_END_WITH_FRAME_ICALL_HANDLES_WRAP_OBJ	HANDLE_FUNCTION_RETURN_OBJ (icall_result)
+
+#define HANDLE_FUNCTION_ENTER_NO_FRAME()	/* nothing */
+#define HANDLE_FUNCTION_ENTER_WITH_FRAME() 	HANDLE_FUNCTION_ENTER()
+
+#define MONO_HANDLE_RETURN_END_NO_FRAME(type)			MONO_HANDLE_DO (MONO_HANDLE_RETURN_END_NO_FRAME_, type);
+#define MONO_HANDLE_RETURN_END_NO_FRAME_Void			return
+#define MONO_HANDLE_RETURN_END_NO_FRAME_ICALL_HANDLES_WRAP_NONE return icall_result
+#define MONO_HANDLE_RETURN_END_NO_FRAME_ICALL_HANDLES_WRAP_OBJ	error_returning_a_managed_type_requires_a_frame
```
jaykrell: Not exactly. Could reuse a parameter handle, and return MONO_HANDLE_RAW of it.

jaykrell:
@monojenkins build failed

jaykrell:
New precedent but yeah maybe. Could also just be “regular” as these seem very inefficient anyway? Need better functionality with system.array in managed? Can cast to byte[]? That’d seem good & might work, I’ll try.

jaykrell:
I like the icall-tables.h feature, could maybe apply to a bunch of others.

lambdageek:
I'm not convinced we need yet another variant of icall-def for something that doesn't seem particularly pricy: a handle stack mark is stored on the stack and consists of an int and a pointer that are saved/restored by an inline function.  We're saving a handful of instructions.

jaykrell:
People were always complaining about coop perf. But which part?
Handle allocation? Native/managed transitions? TLS access? Error setup/check?
Much better now and good enough to stop worrying?
This saves two TLS accesses and some non-inlined functions.
But the native/managed transitions are ideal now, and handle allocation sometimes much cheaper.

jaykrell:
```class A
{
	void f (System.Array b)
	{
		((byte[])b)[0] = (byte)0;
	}
}
```
compiles but doesn't seem as directly applicable as I was thinking.

jaykrell:
This does have error paths, so nohandles() might be considered overkill. Might.
The ideas compose -- handles() + nosafepoints or nohandles +nosafepoints.
Nosafepoints, once broached, might be worth using over a  slightly  larger range, and the old fashioned error handling not so evil.
I still think no_frame might be a good idea where applicable.

Arguably at this point no frame should be the default and only option and any icall wanting to create a handle can use enter/return and create its own frame where needed.

i.e. the frames used to be needed almost across the board, for very many input parameters, now they are only needed for outputs or temporaries -- many/all of which could be created on the managed size as well (reducing/eliminating natively-created handles is not a terrible idea imho).

jaykrell:
Inline function?
Not `mono_thread_info_current`.
`mono_native_tls_get_value` not really, could be better.
Not `mono_stack_mark_record_size`.
`mono_stack_mark_pop` yes. `mono_stack_mark_pop_value` no.
   (`pop` is the case for `no_frame` functions, granted).
`mono_memory_write_barrier` also.

It seems worth removing when function creates no handles, and when handles are created, seems worth using a ref or out parameter to push the handle creation to the managed side.
At which point, native would have far less handle creation of the enter/exit cost.

I can prototype this for a few functions, based on 1. The infrastructure here and 2. Moving handle creation to managed side (in the C# source specifically, implicitly).

jaykrell:
Therefore the `_assign_raw` idiom present in some other PRs helps a bunch then.
Basic underlying, but somewhat abstracted, idiom is that managed has volatile locals, passes their address, and native can assign raw pointers "through" them.

Issue #11998, opened 2018-12-10
Unnecessary exception: System.UnauthorizedAccessException: Access to the path "/Library/Frameworks/Mono.framework/Versions/5.18.0/etc/mono/registry/last-btime" is denied.


rolfbjarne:
An exception is thrown (and handled, so no actual ill effects at runtime) when the registry is accessed:

> System.UnauthorizedAccessException: Access to the path "/Library/Frameworks/Mono.framework/Versions/5.18.0/etc/mono/registry/last-btime" is denied.

This is annoying when:

* Breaking on all exceptions when debugging in VSfM
* Tracing exceptions (this exception confuses everybody that sees it the first time, thinking it's causing something to break).

## Steps to Reproduce

1. Test case:

```csharp
using System.Xml;

class C {
    static void Main ()
    {
            XmlDocument xml = new XmlDocument ();
            xml.LoadXml ("<xml></xml>");
    }
}
```

2. Compile:

```shell
$ csc test.cs
```

3. Run while tracing exceptions:

```
$ mono --trace=E:all test.exe
[0x7fffab6a0380:] EXCEPTION handling: System.UnauthorizedAccessException: Access to the path "/Library/Frameworks/Mono.framework/Versions/5.18.0/etc/mono/registry/last-btime" is denied.

"<unnamed thread>" tid=0x7fffab6a0380 this=0x109e3c130 , thread handle : 0x7ffc826094a0, state : not waiting
  at System.IO.FileStream..ctor (string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int,bool,System.IO.FileOptions) [0x0019e] in <699ddfd029d348d1a69fb1f9fc28d7cb>:0
  at System.IO.FileStream..ctor (string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int,System.IO.FileOptions,string,bool,bool,bool) [0x00000] in <699ddfd029d348d1a69fb1f9fc28d7cb>:0
  at (wrapper remoting-invoke-with-check) System.IO.FileStream..ctor (string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int,System.IO.FileOptions,string,bool,bool,bool) [0x00029] in <699ddfd029d348d1a69fb1f9fc28d7cb>:0
  at System.IO.StreamWriter.CreateFile (string,bool,bool) [0x0001c] in <699ddfd029d348d1a69fb1f9fc28d7cb>:0
[...]

PR #12008, opened 2018-12-10
[System.XML] Get rid of Registry in XmlReader

EgorBo:
Let's don't access Registry every time we use System.Xml for the first time
Fixes https://github.com/mono/mono/issues/11998
.NET Core doesn't have this option at all.
This legacy mode allows users to have old <4.5 behavior when XmlReader could send some web request for DTD urls, we even had a bug for it when `XmlDocument.Load` could crash if internet is turned off (see https://bugzilla.xamarin.com/show_bug.cgi?id=60621)


EgorBo:
@monojenkins build failed

lewurm:
@monojenkins build OS X x64

rolfbjarne:
@lewurm it's just a little bit annoying, but I've coped for years now, so I can cope for a few more months. IOW you don't have to backport for my sake :)

PR #12013, opened 2018-12-11
Fix race condition initializing verbose_method_names.

jaykrell:
verbose_method_names could be read before verbose_method_inited.
This is common form of race condition -- lock-free on demand initialization.

There a few ways to fix it.
 1. A barrier in the fast path -- making it less fast on ARM.
 2. Have a special uninitialized value instead of a separate boolean.
    This isn't always possible and isn't portable to Alpha.
 3. Something complicated involving a thread local, and no barrier.
    This is wrapped up nicely by some C++ implementations.
    http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2660.htm
    (Wrapped up nicely either way -- they could also do option 1,
    esp. on x86 where it is fast, or anything else invented.)
 4. pthread_once or Win32 InitOnceInitialize.
    For Win32 at least, this is equivalent to 1 or worse (two function
    calls and a barrier).

Here we can chose option 2.

Also remove some do-nothing nearby code.


jaykrell:

PR #12014, opened 2018-12-11
[Coop] Convert performance counters and performance counter categories.

jaykrell:
Due to the loops, and to keep the performance counter code structurally unchanged,
introduce string creators that use passed in handles.

The alternative is more churn and bigger slower code, involving
moving loop bodies into separate functions, with separate frames.

C++ can enable avoiding the functions separation, but the frame back and forth
would still occur.

Also cut down on strlen calls.
But keep the shared memory format unchanged to match .NET. (i.e. don't
put string lengths in it).


jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

mono/metadata/object-internals.h
```diff
@@ -1673,13 +1673,14 @@ mono_string_handle_length (MonoStringHandle s);
 
 #endif
 
+#define mono_string_chars_unsafe(s) (mono_string_chars_internal (MONO_HANDLE_RAW (s)))
```
jaykrell: This line needs more attention now or later.

jaykrell:

jaykrell:

jaykrell:
Needs @monojenkins build deb with monolite but  I cannot do it any longer.

vargaz:

jaykrell:
@monojenkins build failed

jaykrell:

akoeplinger:
@monojenkins whitelist

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/mono-perfcounters.c
```diff
@@ -1591,98 +1604,104 @@ mono_perfcounter_instance_exists (MonoString *instance, MonoString *category)
...
 	if (!is_ok (error)) {
-		res = NULL;
+		res = NULL_HANDLE_ARRAY;
 		goto leave;
 	}
 
 	for (i = 0; i < NUM_CATEGORIES; ++i) {
 		const CategoryDesc *cdesc = &predef_categories [i];
-		MonoString *name = mono_string_new_checked (domain, cdesc->name, error);
+		mono_string_new_utf8_assign (name, domain, cdesc->name, cdesc->name_length, error);
```
lambdageek: Could we use `HANDLE_LOOP_PREPARE` and make local frames instead of adding the `_assign` operations?

(although have to be careful about `goto_if_nok`)

mono/metadata/object.c
```diff
@@ -7574,25 +7597,47 @@ mono_utf16_to_utf8 (const gunichar2 *s, gsize slength, MonoError *error)
...
+char *
+mono_string_to_utf8len (MonoStringHandle s, gsize *utf8len, MonoError *error)
+{
+	*utf8len = 0;
+	if (MONO_HANDLE_IS_NULL (s))
+		return NULL;
 
-	return mono_utf16_to_utf8 (mono_string_chars_internal (s), s->length, error);
+	uint32_t gchandle = 0;
+	char *utf8 = mono_utf16_to_utf8len (mono_string_handle_pin_chars (s, &gchandle), mono_string_handle_length (s), utf8len, error);
```
lambdageek: could be done with `NO_SAFEPOINTS` now.

jaykrell: Agreed. I think this is an accidental leftover, having carried this commit along through several PRs.

mono/metadata/mono-perfcounters.c
jaykrell: It is possible. It is not my preference. The result is larger (even if not run) and probably slower (may not matter, but feels hypocritical esp. when managing performance counters), and possibly harder to reason about.

The `HANDLE_LOOP_PREPARE` step is optional actually -- it is an optimization, as the code can access the global, null, and then later get the TLS (in loop, granted, slower).

The real problem imho is placing the begin/end macros carefully, and minding the continue/break/return, as you indicate.

jaykrell:
`_assign` has other benefits -- you can move the handle creation to managed, where it is trivial, and pass those down. And cut down on the native handle creation further.

mono/metadata/mono-perfcounters.c
jaykrell: This also feels contradictory to the support afforded to arrays.
There the pattern is very much to create handle outside of loop, and get elements into it.
All types tend to be used in loops, not just arrays.

That said, I tried it out and it wasn't difficult, for me.
Assignment/reuse is a natural idiom. Granted, at odds with functional programming.

jaykrell: https://github.com/mono/mono/pull/12276 shows the other way. I still prefer to avoid the push/pop/allocate in loop.

PR #12018, opened 2018-12-11
no-op change to get install log for #10919 baseline

dsyme:
To help understand the problems with updating F# in #10919 we need a baseline log of the package build


dsyme:
@monojenkins build pkg

luhenry:
@dsyme feel free to re-open if you want to get the package again.

PR #12025, opened 2018-12-12
Remove toplevel const on culture-info.

jaykrell:
It is rare, not very useful, and very rarely harmful (recent released Visual C++ bug).
What is more typical is a pointer to a constant struct, or a toplevel constant value of the struct, not
a struct of constants. Const within the struct does make the others automatic, but again, is rare.
It also precludes ever having mutable instances of the type.

### DIFF
--- a/mono/metadata/culture-info.h
+++ b/mono/metadata/culture-info.h
@@ -27,111 +27,110 @@
 typedef guint16 stridx_t;
 
 typedef struct {
-	const stridx_t month_day_pattern;
-	const stridx_t am_designator;
-	const stridx_t pm_designator;
+	stridx_t month_day_pattern;
+	stridx_t am_designator;
+	stridx_t pm_designator;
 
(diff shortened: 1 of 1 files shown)


jaykrell:
In a recent release of Visual C++, statically initialized structs, with toplevel const on fields within them, compiled as C++, generated dynamic initializers instead of static ones.

Customer(s) noticed. The behavior is almost the same but not quite.
This toplevel const pattern is *extremely* rare -- rare enough that its implementation regressed.
The rareness of this pattern suggests it is also not useful and should be avoided.

Most need for this pattern can be mimiced by placing const on the instance.
i.e. `const static Foo foo = {  ... };`
Unless the desire is some fields are const and some are not.
However "some fields const" doesn't lead to the data being readonly, and shared across processes if not written, as a fully const/static struct is  -- so again, not all that useful.

Issue #6997, opened 2018-02-13
Random failure in MonoTests.DebuggerTests.StepOutAsync


kumpera:
Found during CI.

Failure looks like this:
```
1) StepOutAsync (MonoTests.DebuggerTests.StepOutAsync)
     Mono.Debugger.Soft.StepEvent
  Expected: True
  But was:  False

  at MonoTests.DebuggerTests.StepOutAsync () [0x00082] in /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mcs/class/Mono.Debugger.Soft/Test/dtest.cs:4305 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mcs/class/corlib/System.Reflection/MonoMethod.cs:305 
```

kumpera:
Found on https://github.com/mono/mono/pull/6958.

Linux x64 interpreter -> https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter/2953/
i386 OSX -> https://jenkins.mono-project.com/job/test-mono-pull-request-i386-osx/7812/

kumpera:
Found on https://github.com/mono/mono/pull/6018

All jobs did hit it.

luhenry:
It's very reproducible whenever I set the number of CPU of my machine to 1 (done in Instruments -> Preferences -> CPUs)

/cc @thaystg

PR #12032, opened 2018-12-12
[debugger] Fix test StepOutAsync

thaystg:
The problem with this flaky issue was the timing of processing of the async tasks, as it was executed in background threads it was not possible to reach the end of execution because all the foreground threads exit earlier.

The solution creates a TaskScheduler that ensure that the thread that will execute the task will be a foreground thread and then the test always works.

Fixes https://github.com/mono/mono/issues/6997


luhenry:
@monojenkins build failed

PR #12033, opened 2018-12-12
[wasm] Avoid dedup-ing StructureToPtr/PtrToStructure wrappers, its not neccessary, and it doesn't seem to work.

vargaz:


kjpou1:

jeromelaban:
For reference, this PR fixes messages like these when running under mono-wasm AOT: 

```
System.ExecutionEngineException: 
Attempting to JIT compile method 
'(wrapper other) void Uno.UI.Xaml.WindowManagerInterop/WindowManagerCreateContentParams:StructureToPtr (object,intptr,bool)' 
while running in aot-only mode. 
See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.
```

PR #12036, opened 2018-12-12
[runtime] Remove duplicate merp invoke line

alexanderkyte:
Looks like this broke when it got squashed down, or we overlooked it. We've got two calls where we expected one.


alexanderkyte:
Including in the existing backport trains

lewurm:
@monojenkins squash

PR #12042, opened 2018-12-13
Fix two leaks in error paths.

jaykrell:
These are the kinds of things a bare minimum of C++ -- destructors -- will make trivial to get correct with smaller/cleaner code.

A strict C convention would help too -- only ever a single point of function exit, at least for functions that "allocate any resources" (or more generally have "invariants to restore" -- like a function that increments some counter upon entry and decrements upon exit).

"Things you find if you read the code while converting to Coop."


jaykrell:

PR #12023, opened 2018-12-11
[Coop] Convert System.Globalization.{CalendarData,CultureData,CultureInfo,RegionInfo} icalls.

jaykrell:

PR #12043, opened 2018-12-13
Fix MonoSortKey native/managed layout mismatch.

jaykrell:
MonoSortKey native:
https://github.com/mono/mono/blob/f019810b12ba130b7fed1cd41bb2eaed04ececef/mono/metadata/object-internals.h#L700

SortKey managed:
https://github.com/mono/mono/blob/f019810b12ba130b7fed1cd41bb2eaed04ececef/mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs#L69

do not agree.

Memory is corrupted and crash if you set $MONO_DISABLE_MANAGED_COLLATION=yes documented here:
https://github.com/mono/mono/blob/f019810b12ba130b7fed1cd41bb2eaed04ececef/man/mono.1#L1161

Fix is to remove the native code and replace by trivial managed.

Found testing coop-conversion of System.Globalization. https://github.com/mono/mono/pull/12023

### DIFF
--- a/mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
+++ b/mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
@@ -77,6 +77,11 @@ namespace System.Globalization
 			this.lcid = lcid;
 			this.source = source;
 			this.options = opt;
+			var source_length = source.Length;
+			var k = new byte [source_length];
+			for (int i = 0; i < source_length; ++i)
+				k [i] = (byte)source [i];
+			this.key = k;
 		}
 
 		internal SortKey (int lcid, string source, byte [] buffer, CompareOptions opt,
--- a/mono/metadata/locales.c
+++ b/mono/metadata/locales.c
@@ -740,7 +740,8 @@ int ves_icall_System_Globalization_CompareInfo_internal_compare (MonoCompareInfo
 					 options));
 }
 
-void ves_icall_System_Globalization_CompareInfo_assign_sortkey (MonoCompareInfo *this_obj, MonoSortKey *key, MonoString *source, gint32 options)
+MonoArray*
+ves_icall_System_Globalization_CompareInfo_assign_sortkey (MonoString *source)
 {
 	ERROR_DECL (error);
 	MonoArray *arr;
@@ -751,13 +752,13 @@ void ves_icall_System_Globalization_CompareInfo_assign_sortkey (MonoCompareInfo 
 	arr=mono_array_new_checked (mono_domain_get (), mono_get_byte_class (),
 				    keylen, error);
 	if (mono_error_set_pending_exception (error))
-		return;
+		return NULL;
 
 	for(i=0; i<keylen; i++) {
 		mono_array_set_internal (arr, guint8, i, mono_string_chars_internal (source)[i]);
 	}
-	
-	MONO_OBJECT_SETREF_INTERNAL (key, key, arr);
+
+	return arr;
 }
 
 int ves_icall_System_Globalization_CompareInfo_internal_index (MonoCompareInfo *this_obj, MonoString *source, gint32 sindex, gint32 count, MonoString *value, gint32 options, MonoBoolean first)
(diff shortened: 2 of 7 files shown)


jaykrell:

jaykrell:

jaykrell:

mono/metadata/locales.c
```diff
@@ -740,7 +740,8 @@ int ves_icall_System_Globalization_CompareInfo_internal_compare (MonoCompareInfo
 					 options));
 }
 
-void ves_icall_System_Globalization_CompareInfo_assign_sortkey (MonoCompareInfo *this_obj, MonoSortKey *key, MonoString *source, gint32 options)
+MonoArray*
+ves_icall_System_Globalization_CompareInfo_assign_sortkey (MonoString *source)
 {
 	ERROR_DECL (error);
```
vargaz: This doesn't seem to do anything useful anymore, it converts a string to its bytes.

jaykrell: Yeah you are right -- just an identity. Just rewrite it in all managed?

There are a few confusing things.
In particular, I thought I saw a similar crash in an unrelated build.
Like maybe monolite got updated with the change, and maybe we depend on the data being corrupted?
Maybe pretend I didn't see anything confusing and proceed.

Removing the icall and just doing it in managed seems reasonable. Let me try that.

jaykrell: Identity, but truncating guint16 to guint8, granted.

jaykrell: Ugh.
```
ReferenceSources/CompareInfo.cs(85,8): error CS0122: 'SortKey.key' is inaccessible due to its protection level
make[8]: *** [../../class/lib/build-macos/mscorlib.dll] Error 1
```

jaykrell:

jaykrell:

mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
```diff
@@ -89,6 +98,7 @@ internal SortKey (int lcid, string source, CompareOptions opt)
 			this.source = source;
 			this.key = buffer;
 			this.options = opt;
+			this.key = CreateKey (source);
```
jaykrell: oops

jaykrell:

mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
```diff
@@ -77,6 +77,11 @@ internal SortKey (int lcid, string source, CompareOptions opt)
 			this.lcid = lcid;
 			this.source = source;
 			this.options = opt;
+			var source_length = source.Length;
+			var k = new byte [source_length];
+			for (int i = 0; i < source_length; ++i)
+				k [i] = (byte)source [i];
```
lewurm: shorter?
```csharp
this.key = new byte [source.Length];
for (int i = 0; i < source.Length; ++i)
  this.key [i] = (byte) source [i];
```

jaykrell: I don't trust every `.` to be a reload from memory. Do you?
I agree it works and means the same thing of course.

jaykrell:
corlib bump not clearly needed.
The new runtime does not work with old corlib. But could be made to, by leaving the code in.
New corlib does work with old runtime.

jaykrell:
@monojenkins squash

mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
lewurm: is it critical in terms of perf or raciness?

jaykrell: No but I prefer to keep it faster when easy.

jaykrell:

vargaz:
@monojenkins build failed

mcs/class/corlib/Mono.Globalization.Unicode/SortKey.cs
luhenry: Use the shorter version.

PR #12047, opened 2018-12-13
[Network] Fix CVE 2018-8292 on Mac OS X.

mandel-macaque:
This commit ensures that CVE 2018-8292 is fixed by not setting
auto-redirect in the first request done by CFNetwork. CFNetwork has the
same bug, but we can work around it by performing the first request,
checking the response and, if it is a rediret, follow it without the
Authorization headers.

PS: This fix should be ported to 2018-08, 2018-10 and 2018-12.

### DIFF
--- a/mcs/class/System.Net.Http/CFNetworkHandler.cs
+++ b/mcs/class/System.Net.Http/CFNetworkHandler.cs
@@ -269,7 +269,7 @@ namespace System.Net.Http
  					bucket.StreamCanBeDisposed = true;
 					// remove headers in a redirect for Authentication.
 					request.Headers.Authorization = null;
- 					var redirectResponse = await SendAsync (redirectRequest, cancellationToken, false).ConfigureAwait (false);
+ 					var redirectResponse = await SendAsync (request, cancellationToken, false).ConfigureAwait (false);
  					return redirectResponse;
  				}
  				return initialRequest;
--- a/mcs/class/System.Net.Http/CFNetworkHandler.cs
+++ b/mcs/class/System.Net.Http/CFNetworkHandler.cs
@@ -267,7 +267,7 @@ namespace System.Net.Http
  				var status = initialRequest.StatusCode;
  				if (IsRedirect (status) && allowAutoRedirect) {
  					bucket.StreamCanBeDisposed = true;
-					// remove headers in a redirec for Authentication.
+					// remove headers in a redirect for Authentication.
 					request.Headers.Authorization = null;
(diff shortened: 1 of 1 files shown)


luhenry:

mandel-macaque:
@luhenry I don't think that the failures are related to the change, but please confirm that :)

mcs/class/System.Net.Http/CFNetworkHandler.cs
```diff
@@ -173,6 +195,43 @@ CFHTTPMessage CreateWebRequestAsync (HttpRequestMessage request)
...
+		
+		static async Task<HttpRequestMessage> CreateHttpRequestMessageRedirectAsync (HttpRequestMessage req)
+		{
+			var str = req.RequestUri.OriginalString;
+			var clone = new HttpRequestMessage (req.Method, str);
+
+			// Copy the request's content (via a MemoryStream) into the cloned object
+			var ms = new MemoryStream ();
+			if (req.Content != null) {
+				await req.Content.CopyToAsync (ms).ConfigureAwait (false);
```
marek-safar: This does not look right. What if the content was already uploaded, why are we uploading it again?

mcs/class/System.Net.Http/CFNetworkHandler.cs
```diff
@@ -173,6 +195,43 @@ CFHTTPMessage CreateWebRequestAsync (HttpRequestMessage request)
...
+
+				// Copy the content headers
+				if (req.Content.Headers != null)
+					foreach (var h in req.Content.Headers)
+						clone.Content.Headers.Add (h.Key, h.Value);
+			}
+
+			clone.Version = req.Version;
+
+			foreach (KeyValuePair<string, object> prop in req.Properties) {
```
marek-safar: All this would be much better handled inside HttpRequestMessage. Perhaps just do it there and then remove `Authorization`

mandel-macaque: This is about the properties, but no the headers, am I confused. The ignore of the header can be found in line 226.

marek-safar: What I meant is doing the close from inside HttpRequestMessage would be much better because you'd not need to manually clone all individual fields

mandel-macaque: We can reuse the obj. Latests push should fix the issue.

mandel-macaque:

Issue #11714, opened 2018-11-16
Improve AOT-d assembly loading process on Android/mobile


grendello:
While loading an assembly, Mono looks for the corresponding AOT DSO in several locations. Some of those are specific to (in my case) Android and lookups are made by the Xamarin.Android runtime on Mono's request, but some locations used by Mono refer to **build host** locations which cannot and will not exist on the mobile device, for instance (full example log for `mscorlib` is [available here](https://gist.github.com/grendello/8e5a0aadd1f2492c2849feec024cb878)):

```
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Trying to load shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so'
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so' not found
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Trying to load shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so'
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so' not found
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Trying to load shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so'
11-15 15:23:23.651 30800 30800 I monodroid-assembly: Shared library '/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/sdks/out/android-arm64-v8a-debug/lib/mono/aot-cache/arm64/mscorlib.dll.so' not found
[...]

PR #12053, opened 2018-12-13
[runtime] Add opt-out for AOT file probing with MONO_AOT_MODE_NONE

alexanderkyte:
Fixes reported startup time issues with probing filesystem for native AOT code files on Android.

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -2034,6 +2034,9 @@ load_aot_module (MonoAssembly *assembly, gpointer user_data)
 	if (mono_compile_aot)
 		return;
 
+	if (mono_aot_mode == MONO_AOT_MODE_NONE)
+		return;
+
 	if (assembly->image->aot_module)
 		/* 
 		 * Already loaded. This can happen because the assembly loading code might invoke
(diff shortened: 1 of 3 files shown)


lewurm:
related https://github.com/mono/mono/issues/11714

lewurm:
Did you test this change?

One of the reported issues by the Android team was that _when_ using AOT, mono would still probe weird places (i.e. paths that only exist on the system where the runtime was _built_), so they would like an API to control the probing.

/cc @lambdageek

alexanderkyte:
@lewurm that is another bug fixed by https://github.com/mono/mono/pull/12056

alexanderkyte:
Hmm, loading *AOT* dylibs from weird paths seems like it might be one of the other symbols

alexanderkyte:
I see, for AOT it's this one https://github.com/mono/mono/pull/12058

alexanderkyte:
@monojenkins backport 2018-08
@monojenkins backport 2018-10
@monojenkins backport 2018-12

@vargaz @lewurm could one of you assert that the scope of negative impact of these changes is minimal? 

I am changing the enum label used for the default value while keeping the behavior of the default, in order to make it possible for someone to opt-out of AOT checking completely.

alexanderkyte:
@lewurm just made that change, making sure it builds locally before I make the PR. I'll cherry pick onto these PRs after it's in master.

PR #11691, opened 2018-11-15
[ios] initial watch64_32 support

lewurm:
```shell
$ file ios-targetwatch64_32-release/mono/mini/.libs/libmonosgen-2.0.dylib
ios-targetwatch64_32-release/mono/mini/.libs/libmonosgen-2.0.dylib: Mach-O dynamically linked shared library arm64_32_v8
```

Note, it doesn't add an AOT compiler yet.

PR #12055, opened 2018-12-13
[sdks] Add a XCODE32_DIR make variable and use it to compile the 32 bit cross compilers.

lewurm:
Forward port of https://github.com/mono/mono/pull/10139

Needed for `arm64_32` (https://github.com/mono/mono/pull/11691): We still need to build AOT compilers that are 32bit, which is only possible with Xcode9. However, `arm64_32` needs Xcode10.

### DIFF
--- a/sdks/versions.mk
+++ b/sdks/versions.mk
@@ -23,7 +23,7 @@ ANDROID_SDK_VERSION_x86_64?=21
 
 # iOS
 
-XCODE_DIR?=/Applications/Xcode101.app/Contents/Developer
+XCODE_DIR?=/Applications/Xcode.app/Contents/Developer
 # Xcode version used to compile 32 bit cross compilers
 XCODE32_DIR?=/Applications/Xcode94.app/Contents/Developer
 
--- a/sdks/versions.mk
+++ b/sdks/versions.mk
@@ -23,7 +23,9 @@ ANDROID_SDK_VERSION_x86_64?=21
 
 # iOS
 
-XCODE_DIR?=/Applications/Xcode.app/Contents/Developer
+XCODE_DIR?=/Applications/Xcode101.app/Contents/Developer
+# Xcode version used to compile 32 bit cross compilers
+XCODE32_DIR?=/Applications/Xcode94.app/Contents/Developer
 
 #MACOS_VERSION?=10.13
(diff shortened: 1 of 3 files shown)


sdks/versions.mk
```diff
@@ -23,7 +23,9 @@ ANDROID_SDK_VERSION_x86_64?=21
 
 # iOS
 
-XCODE_DIR?=/Applications/Xcode.app/Contents/Developer
+XCODE_DIR?=/Applications/Xcode101.app/Contents/Developer
+# Xcode version used to compile 32 bit cross compilers
+XCODE32_DIR?=/Applications/Xcode94.app/Contents/Developer
```
lewurm: This maps to the configuration we have on our CI machines: https://github.com/mono/mono/pull/11691#issuecomment-439178459

sdks/versions.mk
```diff
@@ -23,7 +23,9 @@ ANDROID_SDK_VERSION_x86_64?=21
 
 # iOS
 
-XCODE_DIR?=/Applications/Xcode.app/Contents/Developer
+XCODE_DIR?=/Applications/Xcode101.app/Contents/Developer
```
luhenry: I would keep the directory name to Xcode.app and follow the convention that it is always the latest stable Xcode (10.1 in the current situation)

lewurm: Okay, I've moved the settings needed for the bots to `run-jenkins.sh`. Is that better?

akoeplinger: what if we make `Xcode.app` point to the latest one and `Xcode32bit.app` to the one which can build x86 apps? then we don't need to hardcode version numbers in the scripts anywhere and just rely on bots being set up correctly.

lewurm: we can do that later. for now, if we would do that, it would break current master.

akoeplinger: Not if you add another symlink for the old name, but fair enough :)

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

akoeplinger:
Merging since this only affects the SDKs and those passed (the rest hit some Jenkins issue).

lewurm:
@monojenkins backport 2018-10
@monojenkins backport 2018-12

PR #12056, opened 2018-12-13
[runtime] Don't define build config dir on android

alexanderkyte:
Previously we were defining the build configuration dir on android. This directory exists on the host and not the target. 

It often appears in places like https://github.com/mono/mono/blob/master/mono/metadata/mono-config.c#L647 which causes us to probe files that will never exist.

### DIFF
--- a/mono/metadata/mono-config.c
+++ b/mono/metadata/mono-config.c
@@ -642,9 +642,14 @@ mono_config_for_assembly_internal (MonoImage *assembly)
 	g_free (cfg_name);
 
 	cfg_name = g_strdup_printf ("%s.config", mono_image_get_name (assembly));
+	const char *mono_cfg_dir = mono_get_config_dir ();
+	if (!mono_cfg_dir) {
+		g_free (cfg_name);
+		return;
+	}
 
 	for (i = 0; (aname = get_assembly_filename (assembly, i)) != NULL; ++i) {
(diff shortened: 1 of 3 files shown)


lewurm:
why is this needed?

alexanderkyte:
@lewurm, added a PR body

lewurm:
I don't understand how this helps. if `$mono_cfg_dir` is set to empty string, the code you have linked to https://github.com/mono/mono/blob/caf22d5f0b89b2968c322857ef4d0f00d9bb7ed0/mono/metadata/mono-config.c#L647-L649 would still try to look up `/mono/assemblies/$aname/$cfg_name`, right?

mono/metadata/mono-config.c
```diff
@@ -642,9 +642,10 @@ mono_config_for_assembly_internal (MonoImage *assembly)
 	g_free (cfg_name);
 
 	cfg_name = g_strdup_printf ("%s.config", mono_image_get_name (assembly));
+	const char *mono_cfg_dir = mono_get_config_dir ();
 
-	for (i = 0; (aname = get_assembly_filename (assembly, i)) != NULL; ++i) {
```
luhenry: Just check before https://github.com/mono/mono/pull/12056/files#diff-7f8a7c5a525275537e8961a9c56f8568R644, and do an early exit.

alexanderkyte: Addressed

PR #12058, opened 2018-12-13
[runtime] Don't probe aot cache on android

alexanderkyte:
These attempts result in a lot of attempts to access this (nonexistant) path on android during load


alexanderkyte:
@monojenkins backport 2018-08
@monojenkins backport 2018-10
@monojenkins backport 2018-12

At request of @jonpryor

PR #12059, opened 2018-12-13
Dispose tooltips

PaulBol:
I noticed that some controls do not dispose tooltips and timers. Depending on the underlying implementation these may use native resources. Therefore I'd suggest to dispose these with the parent control to not rely on the garbage collector to eventually release the resources.

Additionally, in the case of `ToolStrip.cs` `CloseToolTip` would first create the objects before turning them off.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/ToolStrip.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/ToolStrip.cs
@@ -744,10 +744,7 @@ namespace System.Windows.Forms
 					ToolStripManager.RemoveToolStrip (this);
 
 					if (tooltip_timer != null)
-					{
-						tooltip_timer.Stop();
 						tooltip_timer.Dispose();
-					}
 
 					if (tooltip_window != null)
 						tooltip_window.Dispose();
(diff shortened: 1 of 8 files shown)


mcs/class/System.Windows.Forms/System.Windows.Forms/ToolStrip.cs
```diff
@@ -743,6 +742,15 @@ protected override void Dispose (bool disposing)
 						this.overflow_button.drop_down.Dispose ();
 
 					ToolStripManager.RemoveToolStrip (this);
+
+					if (tooltip_timer != null)
+					{
+						tooltip_timer.Stop();
+						tooltip_timer.Dispose();
```
filipnavara: Is there any reason why you sometime call just Dispose and other times you call Stop and then Dispose? I would expect the Dispose to be sufficient.

PaulBol: That's true. `Dispose` includes `Stop`. In `ToolStrip.cs` I copied the `Stop` from `CloseToolTip` which previously was called. Let me try to update the PR.

lewurm:
@monojenkins build

lewurm:
@monojenkins approve

lewurm:
@monojenkins build

akoeplinger:
@monojenkins build

akoeplinger:

Issue #11489, opened 2018-10-31
IronPython 2.7.9 strange issue on mono 5.16


equiclouds:
I installed ironpython 2.7.9 on rasperberry pi, and code bellow:

```python
class test:
    def extend(self, a, *b):
        return a, b
def main():
    c1 = test().extend({},{})
    for i in range(0,100):
        c2 = test().extend({},{})
        if c1!=c2:
            print(i,c1,c2)
            return
main()
```

Randomly, after some times, the extend function b paramters will be valued with test() object...

```
root@raspberrypi:/tmp/tmp# ipy test.ipy 
(39, ({}, ({},)), ({}, <__main__.test instance at 0x000000000000006C>))
root@raspberrypi:/tmp/tmp# ipy test.ipy 
(82, ({}, ({},)), ({}, <__main__.test instance at 0x000000000000006C>))
```

The result above means after call test().extend fuctions 39 or 82 times, the test instance will pass to *b paremeter. While the code return None on X86 Devices.

Is mono compile differences for ARM and X86 caused the issue?

marek-safar:
What ARM is this and could you reproduce the issue it with previous Mono versions?

equiclouds:
@marek-safar My platform is RASPBERRY PI DESKTOP run on RASPBERRY PI 3B+:

```
Linux raspberrypi 4.14.71-v7+ #1145 SMP Fri Sep 21 15:38:35 BST 2018 armv7l GNU/Linux
```

I haven't test the code on previous Mono versions yet.

equiclouds:
@marek-safar My platform is RASPBERRY PI DESKTOP run on RASPBERRY PI 3B+, and I can't reproduce the issue it with mono 5.12.

lewurm:
@equiclouds with what mono version do you see this problem? could you try `export MONO_ENV_OPTIONS='-O=-aot'` and run your example code again?

lewurm:
I can reproduce with current master on armv7 (armhf). suggested workaround from my previous comment does _not_ work. will investigate

lewurm:
As a workaround: `MONO_ENV_OPTIONS=--interp` works, but it's significant slower.

lewurm:
```patch
diff --git a/mono/mini/mini-arm.c b/mono/mini/mini-arm.c
index b44ecd7..9f98aa7 100644
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
[...]

PR #12070, opened 2018-12-14
[Tailcall] [arm] [jit] Optimize and fix moving tailcall parameters.

jaykrell:
- Only move unaligned stack usage worth of parameters.
This is a simple optimization.

- When moving parameters for tailcall, and having pushed ip,
and adjusting the offset to account for that, split the adjustment
between sp and frame pointer, as frame pointer may or may not be sp.
This fixes https://github.com/mono/mono/issues/11489.

Here is the incorrect code (with the optimization applied):

`configure -prefix /i/1`

`MONO_VERBOSE_METHOD=*:CallSite.Target MONO_LOG_LEVEL=debug MONO_LOG_MASK=tailcall /i/1/bin/mono /usr/share/ironpython2.7/ipy.exe ./1.py `

There are many functions with this name.
I believe it is the last one printed, at least when it fails.

```
   0:	e92d49f0 	push	{r4, r5, r6, r7, r8, fp, lr} ; ok, mirrored later
   4:	e24dd05c 	sub	sp, sp, #92	; 0x5c ; ok, mirrored of later
 .
 .
 .

 274:	e595c00c 	ldr	ip, [r5, #12]   ; read function pointer into ip
 278:	e92d1000 	stmfd	sp!, {ip}       ; push ip, so sp needs adjustment by 4 
 27c:	e28bc078 	add	ip, fp, #120	; 92 from above + 7*4 for the pushes/pops
						; for some functions, previous instruction will
						; be add ip, sp, #120, and those already worked

 280:	e59de004 	ldr	lr, [sp, #4]    ; sp adjusted by 4, good
 284:	e58ce004 	str	lr, [ip, #4]    ; fp adjusted by 4 also, incorrect
```
Without the optimization, there was another ldr/str with offset=8, which should be ok,
it is just moving unused/uninitialize space.
```
 288:	e8bd1000 	ldmfd	sp!, {ip}          		; pop ip=function pointer
 28c:	e28bd05c 	add	sp, fp, #92	; 0x5c		; mirrors prolog
 290:	e8bd49f0 	pop	{r4, r5, r6, r7, r8, fp, lr}	; mirrors prolog
 294:	e12fff1c 	bx	ip				; and jump
```

Here is the corrected code:

```
   0:	e92d49f0 	push	{r4, r5, r6, r7, r8, fp, lr} ; same
   4:	e24dd05c 	sub	sp, sp, #92	; 0x5c ; same
 .
 .
 .
 240:	e596c00c 	ldr	ip, [r6, #12]     	; different register allocation, ok
 244:	e92d1000 	stmfd	sp!, {ip}		; same
 248:	e28bc078 	add	ip, fp, #120		; same
 24c:	e59de004 	ldr	lr, [sp, #4]		; same
 250:	e58ce000 	str	lr, [ip]		; <== the fix
 254:	e8bd1000 	ldmfd	sp!, {ip}		; same
 258:	e28bd05c 	add	sp, fp, #92	; 0x5c	; same
 25c:	e8bd49f0 	pop	{r4, r5, r6, r7, r8, fp, lr} ; same
 260:	e12fff1c 	bx	ip			; same
```

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -1624,10 +1624,10 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
 		add_general (&gr, &stack_size, &cinfo->sig_cookie, TRUE);
 	}
 
+	// Alignment causes tailcall to copy extra data unnecessarily.
 	DEBUG (g_print ("      stack size: %d (%d)\n", (stack_size + 15) & ~15, stack_size));
-	stack_size = ALIGN_TO (stack_size, MONO_ARCH_FRAME_ALIGNMENT);
-
-	cinfo->stack_usage = stack_size;
+	cinfo->unaligned_stack_usage = stack_size;
+	cinfo->stack_usage = ALIGN_TO (stack_size, MONO_ARCH_FRAME_ALIGNMENT);
 	return cinfo;
 }
 
@@ -2652,6 +2652,7 @@ mono_arch_emit_call (MonoCompile *cfg, MonoCallInst *call)
 
 	call->call_info = cinfo;
 	call->stack_usage = cinfo->stack_usage;
+	call->unaligned_stack_usage = cinfo->unaligned_stack_usage;
 }
 
 static void
@@ -5001,7 +5002,7 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 			gboolean const tailcall_reg = ins->opcode == OP_TAILCALL_REG;
 			MonoCallInst *call = (MonoCallInst*)ins;
 
-			max_len += call->stack_usage / sizeof (target_mgreg_t) * ins_get_size (OP_TAILCALL_PARAMETER);
+			max_len += call->unaligned_stack_usage / sizeof (target_mgreg_t) * ins_get_size (OP_TAILCALL_PARAMETER);
 
 			if (IS_HARD_FLOAT)
 				code = emit_float_args (cfg, call, code, &max_len, &offset);
@@ -5035,13 +5036,15 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 			 * Need to copy the arguments from the callee argument area to
 			 * the caller argument area, and pop the frame.
 			 */
-			if (call->stack_usage) {
+
+			if (call->unaligned_stack_usage) {
 				int i, prev_sp_offset = 0;
-				int saved_ip_offset = 0;
+				int saved_ip_offset_sp = 0;
+				int saved_ip_offset_fp = 0;
 
 				if (tailcall_membase || tailcall_reg) {
 					ARM_PUSH (code, 1 << ARMREG_IP);
-					saved_ip_offset = 4;
+					saved_ip_offset_sp = 4;
 				}
 
 				/* Compute size of saved registers restored below */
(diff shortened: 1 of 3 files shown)


mono/mini/mini-arm.c
```diff
@@ -5055,16 +5058,18 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
 
 				/* Copy arguments on the stack to our argument area */
 				// FIXME a fixed size memcpy is desirable here,
 				// at least for larger values of stack_usage.
-				for (i = 0; i < call->stack_usage; i += sizeof (target_mgreg_t)) {
-					ARM_LDR_IMM (code, ARMREG_LR, ARMREG_SP, i + saved_ip_offset);
-					ARM_STR_IMM (code, ARMREG_LR, ARMREG_IP, i + saved_ip_offset);
+				for (i = 0; i < call->unaligned_stack_usage; i += sizeof (target_mgreg_t)) {
+					ARM_LDR_IMM (code, ARMREG_LR, ARMREG_SP, i + saved_ip_offset_sp);
+					ARM_STR_IMM (code, ARMREG_LR, ARMREG_IP, i + saved_ip_offset_fp);
```
lewurm: `s/saved_ip_offset_fp/saved_ip_offset_ip`?

jaykrell: I know, it is confusing, but I meant that.
What to say that won't tell you what you already know?
There is the stack aka the frame. There are one or two pointers to it. The "floaty" one and "fixedy" one. Sometimes just one.
We are moving from the bottom of the floaty one, sp, to the top of the fixedy one, either sp or fp. As well, because we saved ip and were out of space, we pushed it, so we are adjusting for ip's saving..and the values are adjusting sp, and ip, ip as a temporary stand-in for fp, because, um, I'm not sure. Perhaps where it says ARMREG_IP it can actually say `cfg->frame_reg`, and then the entire push/pop and offseting can go away?

Let me think about that.

jaykrell: Ok, I think the problem with the alternative..it will often work, but the problem is `emit_big_add`.
`i + saved_ip_offset_xx` cannot be unboundedly sized, and we are using ARMREG_IP to aborb that.

jaykrell: Right. There is a 4k limit on parameter sizing:
```
mono_arch_tailcall_supported:
	// FIXME The limit here is that moving the parameters requires addressing the parameters
	// with 12bit (4K) immediate offsets. - 4 for TAILCALL_REG/MEMBASE
	res &= IS_SUPPORTED_TAILCALL (callee_info->stack_usage < (4096 - 4));
	res &= IS_SUPPORTED_TAILCALL (caller_info->stack_usage < (4096 - 4));
```

but what is being allowed here is huge amount of locals also.
If locals + params < 4k, then we could simplify, and that would be the overwhelming case.
It is ok asis, except for its error-proneness that caught me.

jaykrell: Another plan is to store the parameters there in the first place.
Being sure not to use that space otherwise.
But I think this is ok.

jaykrell:
My historical tailcall testing focused more on whether or not a tailcall occurred, and less on if the parameters were passed correctly. So more tests would be good. But not likely to happen today.

jaykrell:
Closing in dereference to https://github.com/mono/mono/pull/12079 that optimizes more in the very common case.
But if disagreed and don't want the optimization, close that and reopen this.
There is another even smaller alternative of course, of keeping the aligned copy too.

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build failed

lewurm:
superseded by https://github.com/mono/mono/pull/12701

jaykrell:
Not entirely superceded. There are small easy optimizations ignored. There is another PR with additional though. Partially superceded, agreed.

jaykrell:
The superceding PR also moves fetching of the call target. Why? That seems dangerous. Seems to have broken stuff as a result.

PR #12081, opened 2018-12-16
[runtime] Guard usage of unix functions that aren't posix

alexanderkyte:
Some of these functions aren't available on POSIX-compliant platforms that are not linux or OSX. This some checks for these functions. 

It also makes it easier to use DISABLE_SOCKETS, by adding an enable-minimal option for it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2022,6 +2022,9 @@ if test x$host_win32 = xno; then
 		AC_CHECK_FUNCS(getpwuid_r)
 	fi
 
+	AC_CHECK_FUNCS(readlink)
+	AC_CHECK_FUNCS(chmod)
+
 	AC_FUNC_STRERROR_R()
 
 	dnl ****************************************************************
@@ -2519,7 +2522,6 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(pthread_mutex_timedlock)
 	AC_CHECK_FUNCS(pthread_getattr_np pthread_attr_get_np pthread_getname_np pthread_setname_np pthread_cond_timedwait_relative_np)
 	AC_CHECK_FUNCS(pthread_kill)
-	AC_CHECK_FUNCS(readlink)
 	AC_MSG_CHECKING(for PTHREAD_MUTEX_RECURSIVE)
 	AC_TRY_COMPILE([ #include <pthread.h>], [
 		pthread_mutexattr_t attr;
--- a/configure.ac
+++ b/configure.ac
@@ -476,6 +476,11 @@ fi
 
 AC_SUBST(PLATFORM_AOT_SUFFIX)
 
+if test -z "$HOST_WASM_TRUE"; then :
+AC_DEFINE(HAVE_UTIME)
+AC_DEFINE(HAVE_UTIMES)
+fi
+
 ## PLATFORM_AOT_SUFFIX not so simple for windows :-)
 
 AC_CHECK_TOOL(CC, gcc, gcc)
--- a/configure.ac
+++ b/configure.ac
@@ -1492,7 +1492,7 @@ fi
 AC_ARG_ENABLE(minimal, [  --enable-minimal=LIST      drop support for LIST subsystems.
      LIST is a comma-separated list from: aot, profiler, decimal, pinvoke, debug, appdomains, verifier, 
      reflection_emit, reflection_emit_save, large_code, logging, com, ssa, generics, attach, jit, interpreter, simd, soft_debug, perfcounters, normalization, desktop_loader, shared_perfcounters, remoting,
-	 security, lldb, mdb, assert_messages, cleanup, sgen_marksweep_conc, sgen_split_nursery, sgen_gc_bridge, sgen_debug_helpers.],
+	 security, lldb, mdb, assert_messages, cleanup, sgen_marksweep_conc, sgen_split_nursery, sgen_gc_bridge, sgen_debug_helpers, sockets.],
 [
 	for feature in `echo "$enable_minimal" | sed -e "s/,/ /g"`; do
 		eval "mono_feature_disable_$feature='yes'"
@@ -1692,6 +1692,11 @@ fi
 if test "x$mono_feature_disable_sgen_debug_helpers" = "xyes"; then
 	AC_DEFINE(DISABLE_SGEN_DEBUG_HELPERS, 1, [Disable debug helpers in SGEN.])
 	AC_MSG_NOTICE([Disabled debug helpers in SGEN.])
+fi
+
+if test "x$mono_feature_disable_sockets" = "xyes"; then
+	AC_DEFINE(DISABLE_SOCKETS, 1, [Disable sockets])
+	AC_MSG_NOTICE([Disabled sockets])
 fi
 
 AC_ARG_ENABLE(executables, [  --disable-executables disable the build of the runtime executables], enable_executables=$enableval, enable_executables=yes)
--- a/configure.ac
+++ b/configure.ac
@@ -2039,6 +2039,7 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(utime utimes)
 
 	AC_CHECK_FUNCS(openlog closelog)
+	AC_CHECK_FUNCS(atexit)
 
 	AC_FUNC_STRERROR_R()
 
--- a/configure.ac
+++ b/configure.ac
@@ -2030,6 +2030,8 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(ftruncate)
 	AC_CHECK_FUNCS(msync)
 
+	AC_CHECK_FUNCS(gethostname)
+
 	AC_FUNC_STRERROR_R()
 
 	dnl ****************************************************************
--- a/configure.ac
+++ b/configure.ac
@@ -2027,6 +2027,9 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(lstat)
 	AC_CHECK_FUNCS(getdtablesize)
 
+	AC_CHECK_FUNCS(ftruncate)
+	AC_CHECK_FUNCS(msync)
+
 	AC_FUNC_STRERROR_R()
 
 	dnl ****************************************************************
--- a/configure.ac
+++ b/configure.ac
@@ -2038,6 +2038,8 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(gethostname)
 	AC_CHECK_FUNCS(utime utimes)
 
+	AC_CHECK_FUNCS(openlog closelog)
+
 	AC_FUNC_STRERROR_R()
 
 	dnl ****************************************************************
--- a/configure.ac
+++ b/configure.ac
@@ -2519,6 +2519,7 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNCS(pthread_mutex_timedlock)
 	AC_CHECK_FUNCS(pthread_getattr_np pthread_attr_get_np pthread_getname_np pthread_setname_np pthread_cond_timedwait_relative_np)
 	AC_CHECK_FUNCS(pthread_kill)
+	AC_CHECK_FUNCS(readlink)
 	AC_MSG_CHECKING(for PTHREAD_MUTEX_RECURSIVE)
 	AC_TRY_COMPILE([ #include <pthread.h>], [
 		pthread_mutexattr_t attr;
--- a/configure.ac
+++ b/configure.ac
@@ -2024,6 +2024,7 @@ if test x$host_win32 = xno; then
 
 	AC_CHECK_FUNCS(readlink)
 	AC_CHECK_FUNCS(chmod)
+	AC_CHECK_FUNCS(lstat)
 
 	AC_FUNC_STRERROR_R()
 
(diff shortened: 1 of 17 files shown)


mono/metadata/icall.c
```diff
@@ -6737,6 +6737,7 @@ mono_icall_get_machine_name (MonoError *error)
 	n = 512;
 	buf = (char*)g_malloc (n + 1);
 
+#if defined(HAVE_GETHOSTNAME)
```
jaykrell: Skip malloc?

configure.ac
```diff
@@ -1998,6 +2003,21 @@ if test x$host_win32 = xno; then
...
+
+	AC_CHECK_FUNCS(ftruncate)
+	AC_CHECK_FUNCS(msync)
+
+	AC_CHECK_FUNCS(gethostname getpeername)
+	AC_CHECK_FUNCS(geteuid)
+	AC_CHECK_FUNCS(utime utimes)
+
+	AC_CHECK_FUNCS(openlog closelog)
+	AC_CHECK_FUNCS(atexit)
```
jaykrell: Atexit is ANSI C89. Is any system really lacking it?

mono/eglib/gspawn.c
```diff
@@ -226,7 +226,7 @@ write_all (int fd, const void *vbuf, size_t n)
 	return nwritten;
 }
 
-#ifndef G_OS_WIN32
+#if !defined(G_OS_WIN32) && defined(HAVE_GETDTABLESIZE)
```
jaykrell: Whitespace.

mono/eglib/gfile-unix.c
```diff
@@ -58,11 +58,13 @@ g_file_test (const gchar *filename, GFileTest test)
 		if (access (filename, X_OK) == 0)
 			return TRUE;
 	}
+#ifdef HAVE_LSTAT)
```
jaykrell: Extra paren.

mono/metadata/assembly.c
```diff
@@ -997,8 +997,12 @@ mono_set_rootdir (void)
 	int  s;
 	char *str;
 
+#if defined(HAVE_READLINK)
```
jaykrell: Whitespace.

mono/metadata/assembly.c
```diff
@@ -1008,7 +1012,13 @@ mono_set_rootdir (void)
 
 	/* Solaris 10 style */
 	str = g_strdup_printf ("/proc/%d/path/a.out", getpid ());
+
+#if defined(HAVE_READLINK)
```
jaykrell: Whitespace.

mono/metadata/w32file-unix.c
```diff
@@ -351,21 +357,26 @@ _wapi_chmod (const gchar *pathname, mode_t mode)
...
+		ret = -1;
+#endif
 		g_free (located_filename);
 	}
 
 	return ret;
 }
 
-#ifndef HAVE_STRUCT_TIMEVAL
+#ifndef HAVE_UTIME HAVE_STRUCT_TIMEVAL
```
jaykrell: Doesn’t make sense. Maybe should be defined && defined?

mono/metadata/file-mmap-posix.c
```diff
@@ -489,8 +493,10 @@ mono_mmap_flush (void *mmap_handle, MonoError *error)
 {
 	MmapInstance *h = (MmapInstance *)mmap_handle;
 
+#ifdef HAVE_MSYNC
```
jaykrell: Are there systems with mmap but without msync? It is important sometimes to not silently drop semantics. Sometimes better to fail to compile. These ifdefs in general imply a latent untested combinatorial explosion. Not clearly worth filling the minefield.

mono/metadata/w32file-unix.c
```diff
@@ -525,6 +540,9 @@ _wapi_lstat (const gchar *path, struct stat *buf)
 		ret = lstat (located_filename, buf);
 		g_free (located_filename);
 	}
+#else
+	ret = -1;
+#endif
```
jaykrell: Certain nobody looks at the struct in the error path? Memset(0) it?

mono/metadata/w32process-unix-default.c
```diff
@@ -319,8 +321,11 @@ mono_w32process_get_modules (pid_t pid)
 		if (!g_ascii_isspace (*p)) {
 			continue;
 		}
-
+#if defined(MAJOR_IN_MKDEV) || defined(MAJOR_IN_SYSMACROS)
 		device = makedev ((int)maj_dev, (int)min_dev);
```
jaykrell: Whitespace.

mono/metadata/w32file-unix.c
```diff
@@ -3668,9 +3686,13 @@ mono_w32file_set_attributes (const gunichar2 *name, guint32 attrs)
 		if ((buf.st_mode & S_IROTH) != 0)
 			exec_mask |= S_IXOTH;
 
+#if defined(HAVE_CHMOD)
```
jaykrell: Space after paren.

mono/mini/main.c
```diff
@@ -159,7 +159,9 @@ bundle_save_library_initialize (void)
 	g_free (path);
 	if (bundled_dylibrary_directory == NULL)
 		return;
+#ifdef HAVE_ATEXIT
```
jaykrell: Atexit is ANSI C89 and should be as portable as malloc, sprintf, exit, strlen, etc.

mono/metadata/file-mmap-posix.c
```diff
@@ -286,7 +286,9 @@ open_file_map (const char *c_path, int input_fd, int mode, gint64 *capacity, int
 	}
 
 	if (result != 0 || *capacity > buf.st_size) {
+#ifdef HAVE_FTRUNCATE
 		int unused G_GNUC_UNUSED = ftruncate (fd, (off_t)*capacity);
+#endif
```
jaykrell: Else what?

mono/metadata/file-mmap-posix.c
```diff
@@ -358,7 +360,9 @@ open_memory_map (const char *c_mapName, int mode, gint64 *capacity, int access,
 		}
 
 		unlink (file_name);
+#ifdef HAVE_FTRUNCATE
 		unused = ftruncate (fd, (off_t)*capacity);
+#endif
```
jaykrell: Else what?

mono/metadata/mono-security.c
```diff
@@ -65,9 +65,13 @@
  */
 static size_t mono_sysconf (int name)
 {
+#ifdef HAVE_GETEUID
```
jaykrell: Incorrect ifdef. One needs to read posix & various manuals to see how portable each function is. It might be better to assume things are there when they usually are, despite posix. Otherwise future porters get to debug each of these silent changes.

mono/metadata/w32file-unix.c
```diff
@@ -351,21 +357,26 @@ _wapi_chmod (const gchar *pathname, mode_t mode)
 			return -1;
 		}
 
+#if defined(HAVE_CHMOD)
```
jaykrell: Whitespace. And chmod is in posix. Ifdef should not be needed & is therefore harmful. Some known system lacks it?

mono/metadata/assembly.c
jaykrell: Readlink is posix. Ifdef should not be needed.
http://pubs.opengroup.org/onlinepubs/9699919799

jaykrell:
If this is for an actual system please indicate so, even if the system cannot be identified.
Ideally, but it is expensive human time, make sure every missing function is gracefully handled.
I'd describe such a system as Posix-like and C89-like, but conformant to neither.
Maybe even put in comments jn the code, not just the PR. At least configure.ac -- "actually  missing in some system(s)".
Because allowing for arbitrary missing functions when standards require them seems dubious, unless there are actual systems like that.

NattyNarwhal:
I suspect it would be better to handle cases of missing functions as you see them in a port. When you compile for it, you can deal with the function in context with immediacy instead of ifdefing it out, and implement the workaround from there.

mono/utils/dlmalloc.c
```diff
@@ -1187,7 +1187,7 @@ int mspace_mallopt(int, int);
 #include <string.h>      /* for memset etc */
 #endif  /* LACKS_STRING_H */
 #if USE_BUILTIN_FFS
-#ifndef LACKS_STRINGS_H
+#ifndef LACKS_STRINGS_H && defined(HAVE_STRINGS_H)
```
jaykrell: I don't think `#ifndef` composes that way, but is merely ignoring the rest of the line.
It should be `#if !defined (LACKS_STRINGS) && defined (HAVE_STRINGS_H)`.

jaykrell: Or does not even compile.

jaykrell:
```
$ cat 1.c
#ifndef LACKS_STRINGS_H && defined(HAVE_STRINGS_H)
#endif
HB30-VM-Eklavya:mono4 jay$ gcc -c  1.c
1.c:1:25: warning: extra tokens at end of #ifndef directive [-Wextra-tokens]
#ifndef LACKS_STRINGS_H && defined(HAVE_STRINGS_H)
                        ^
                        //
1 warning generated.
```

mono/utils/mono-dl-posix.c
```diff
@@ -47,7 +47,11 @@ mono_dl_get_so_suffixes (void)
 int
 mono_dl_get_executable_path (char *buf, int buflen)
 {
+#if defined(HAVE_READLINK)
 	return readlink ("/proc/self/exe", buf, buflen - 1);
+#else
+	return -1;
```
jaykrell: Maybe set errno to ENOSYS or such? Looks like that wouldn't really do anything.
Or raise `#ifdef` out to caller? Seems reasonable, if obsessing over size.
Or have `#ifdef HAVE_MONO_DL_GET_EXECUTABLE_PATH` to abstract it? To let the caller decide what to do, if the function will always fail.

alexanderkyte:
@NattyNarwhal these *are* fixes from a port 😉

NattyNarwhal:
Now you have me curious as to what could be more vague with POSIX than AIX/PASE ;)

mono/metadata/file-mmap-posix.c
alexanderkyte: It doesn't seem like there would be any reason for there to be an else body. We silently discard the error there (the variable "unused"), and we don't seem to require that the file size truncation happen for correctness reasons. On the platforms without ftruncate we will simply not truncate.

alexanderkyte:
Wasm failures:
```

|
|
|
WASM: 
WASM: Tests run: 7767, Passed: 7723, Errors: 2, Failures: 0, Inconclusive: 1
WASM:   Not run: 41, Invalid: 0, Ignored: 41, Skipped: 0
WASM: Elapsed time: 00:09:29.6810000
WASM: 
WASM: Errors and Failures:
WASM: 
WASM: 1) LastWriteTime (MonoTests.System.IO.FileTest.LastWriteTime)
WASM:    System.IO.IOException : Invalid parameter
WASM:   at System.IO.File.SetLastWriteTime (System.String path, System.DateTime lastWriteTime) <0x7c50590 + 0x0005e> in <1e8b12bf6b694bdfa46faefda409538c>:0 
WASM:   at MonoTests.System.IO.FileTest.LastWriteTime () <0x7c50030 + 0x000a0> in <59f1c4d2a4dd448ebed1b6fdb3015ff3>:0 
WASM:   at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
WASM:   at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) <0x30d70c8 + 0x000ea> in <1e8b12bf6b694bdfa46faefda409538c>:0 
WASM: 
WASM: 
WASM: 2) LastWriteTimeSubMs (MonoTests.System.IO.FileTest.LastWriteTimeSubMs)
WASM:    System.IO.IOException : Invalid parameter
WASM:   at System.IO.File.SetLastWriteTime (System.String path, System.DateTime lastWriteTime) <0x7c50590 + 0x0005e> in <1e8b12bf6b694bdfa46faefda409538c>:0 
WASM:   at System.IO.File.SetLastWriteTimeUtc (System.String path, System.DateTime lastWriteTimeUtc) <0x7c51860 + 0x00020> in <1e8b12bf6b694bdfa46faefda409538c>:0 
WASM:   at MonoTests.System.IO.FileTest.LastWriteTimeSubMs () <0x7c511b8 + 0x00184> in <59f1c4d2a4dd448ebed1b6fdb3015ff3>:0 
WASM:   at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
WASM:   at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) <0x30d70c8 + 0x000ea> in <1e8b12bf6b694bdfa46faefda409538c>:0 
WASM: 
```

alexanderkyte:
That github diff looked really weird. I rebased on master, that looks better.

PR #12082, opened 2018-12-16
[runtime] Create some indirection for errno setting

alexanderkyte:
Some posix platforms require some logic around errno setting. Rather than hardcode any of that here now, I factored our setting into a single place. 

As an enhancement for today, this makes it easy to add some logging around setting it. 

It should all compile away to a no-op as this PR is.


mono/utils/mono-errno.h
```diff
@@ -0,0 +1,21 @@
...
+ * (C) 2018 Microsoft, Inc.
+ *
+ */
+
+#ifndef __MONO_ERRNO_H__
+#define __MONO_ERRNO_H__
+
+// Enough indirection to do something else here, or log
+
+#define SET_ERRNO(errno_val) do { \
```
jaykrell: Inline function? So I can set breakpoint? Expression? Getter not just setter? And no trailing semicolon. But good idea.

mono/utils/mono-threads-posix-signals.c
```diff
@@ -121,17 +122,17 @@ static void
...
 	info->signal = restart_signal_num;
-	errno = old_errno;
+	SET_ERRNO(old_errno);
 }
 
 static void
 suspend_signal_handler (int _dummy, siginfo_t *info, void *context)
 {
-	int old_errno = errno;
+	int old_SET_ERRNO(errno;
```
jaykrell: Oops.

mono/utils/mono-errno.h
jaykrell: And it should be rare enough, that a non-inline function should suffice. Errno is a thread local, and accesses might involve a function call, or two. On Windows, probably two, try stepping through it.

mono/btls/btls-bio.c
```diff
@@ -40,14 +41,14 @@ mono_read (BIO *bio, char *out, int outl)
 	ret = mono->read_func (mono->instance, out, outl, &wantMore);
 
 	if (ret < 0) {
-		errno = EIO;
+		SET_ERRNO(EIO)
```
alexrp: Missing space before opening parenthesis throughout.

mono/utils/mono-errno.h
jaykrell: i.e. one of:
`#define SET_ERRNO(x) (errno = (x))`

or:
```
void mono_set_errno(inline x); // .h  inline or not 
void mono_set_errno(int x) { errno = x; ] // .c
```

A macro that is well behaved (no multi-evaluation of parameters) imho could just be named like a function, though I not sure mono allows that.

I'd just as soon put `MONO_` or `mono_` prefix on, even if there no linkage (macro).
i.e. what if later it becomes a function (granted `#define SET_ERRNO mono_set_errno`).

NattyNarwhal:
> Some posix platforms

I'm curious which now - I thought every platform defined a sane errno macro/TL variable? (I do know errno also has to be a macro, per the C++ standard?)

FWIW, I know AIX is bone-headed in many ways, but there errno is just dereferencing the result of a function call.... once you define thread safety.

```c
#if defined(_THREAD_SAFE) || defined(_THREAD_SAFE_ERRNO)
/*
 * Per thread errno is provided by the threads provider. Both the extern int
 * and the per thread value must be maintained by the threads library.
 */
extern  int     *_Errno( void );
#define errno   (*_Errno())

#else

extern int errno;

#endif  /* _THREAD_SAFE || _THREAD_SAFE_ERRNO */
```

alexanderkyte:
When you have a flat/single-process OS, you have issues with setting errno directly. The set_errno function offers just enough indirection to make it sane.

mono/utils/mono-errno.h
lambdageek: (nit) Could you make the macro definition not include a `;` and make all the callers write `SET_ERRNO (whatever);` (with a trailing `;`)?

mono/utils/mono-threads-posix-signals.c
```diff
@@ -121,17 +122,17 @@ static void
 restart_signal_handler (int _dummy, siginfo_t *_info, void *context)
 {
 	MonoThreadInfo *info;
-	int old_errno = errno;
+	int old_SET_ERRNO(errno;
```
lambdageek: sed failure, looks like.

mono/utils/mono-errno.h
jaykrell: Or `#define SET_ERRNO(x) (errno = (x))`?

jaykrell: Or an inline function, or even a function? All should be ok. errno access is already big and slow (a function call), a function won't be much bigger and slower. Inline function would be no change.

alexanderkyte:
@monojenkins build failed

Issue #11940, opened 2018-12-05
[enhancement] Add C code backend through llvm-cbe


alexanderkyte:
People wanting to use mono with AOT but not wanting to work with LLVM IR (OpenCL-linking code, C code compiled by gcc for exotic platforms, scientific computing, exotic game consoles, etc) find themselves in a place where being able to AOT their .dlls into C code would be helpful. 

This stack isn't an uncommon one for this reason. Unity has their own IL2CPP. 

As a quick win, I wanted to see if I could get llvm to emit C code instead of using a new pipeline. Success was pretty quick to be had.

```
- Check out https://github.com/JuliaComputing/llvm-cbe into `external/llvm/projects`
- Edit `llvm/build.mk` to add llvm-cbe to the list of LLVM_TOOLS_TO_BUILD
- Make aot compiler save bitcode files. 
-- Pass each to llvm-dis. Pass output of that to llvm-cbe. Use `-o` flag
- Collect one C file per managed assembly 
- Link and compile it all together. (not done)
```

I was able to turn a HelloWorld.exe app into a C source file using this stack. 

I think that with a very small amount of effort, we could get mkbundle to produce a collection of C source files for use with the developer's existing C stack.

jaykrell:
I kinda thought a new from-scratch project but that is promising.

alexanderkyte:
This PR adds the backend and tool https://github.com/mono/llvm/pull/37

alexanderkyte:
```akyte@akyte-pc ~/mono/mono/tests (master*) $ cat exception19.cs
using System;

public class TestFinallyException {

	public static int Main (string[] args)
	{
		return args.Length;
	}

}
```

yields 
[test.c.txt](https://github.com/mono/mono/files/2651558/test.c.txt)
[test.ll.txt](https://github.com/mono/mono/files/2651559/test.ll.txt)

alexanderkyte:
it appears to reference 

https://github.com/JuliaLang/julia/blob/master/src/APInt-C.cpp
https://github.com/JuliaLang/julia/blob/master/src/APInt-C.h

but that seems easy enough to fix

alexanderkyte:
After removing it, I see some:

```
[...]

PR #12083, opened 2018-12-16
[runtime] Add some llvm-cbe automation

alexanderkyte:
Automates the steps named in https://github.com/mono/mono/issues/11940

Note that the emitted C currently has some issues compiling due to linking to some minor runtime code from Julia. I need to bring that over, or make it get those primitives from us. That'll probably happen in the `mono/llvm` repo.


alexanderkyte:
This is a small amount of automation around an experiment that isn't enabled right now. Anybody who wants to revive this code is welcome to contact me about how to do so. I'll do my best to assist you.

PR #12088, opened 2018-12-17
[wasm] Update build docs

kjpou1:

### DIFF
--- a/sdks/README.md
+++ b/sdks/README.md
@@ -58,7 +58,7 @@ mono$ make -C sdks/builds archive-mac [upcoming]
 
 ## WebAssembly
 
-First, ensure the `bcl`, `AOT` and `wasm-runtime` have been built and packaged in the `builds` directory:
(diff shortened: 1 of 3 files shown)


sdks/README.md
```diff
@@ -31,49 +31,50 @@ it in your system's package manager](https://github.com/ninja-build/ninja/wiki/P
 Copy `Make.config.sample` to `Make.config` and edit the new file to disable building the target you're not interested in.
 Unless you have a very particular need, the BCL build should be left enabled as it's needed by all test runners.
 
-## Targets
+## Building for XA/XI/XM and WebAssembly
 
-To build Mono for Android, iOS or WebAssembly, the build scripts can be found in `sdks/builds/`. There are 5 predefined targets that can be invoked:
```
vargaz: Why is this removed ? Its just some general comments which still seem to be true.

kjpou1: Although, a comment about the `sdks/builds/out` should be mentioned.

Issue #12060, opened 2018-12-13
Compiling on RHEL 6 or 7 with a non-default gcc ignores --disable-rpath and adds libstdc++


rjvdboon:
## Steps to Reproduce

1. Install an additional GCC on your system (side-by-side, so not replacing the default)
2. extract a mono release tarball
3. Set appropriate PATH, LD_LIBRARY_PATH, INCLUDE_PATH etc. to point to the additional GCC. (We use [Environment modules](https://en.wikipedia.org/wiki/Environment_Modules_(software))
4. `./configure --disable-rpath --prefix=...`
5. make
6. inspect mono-sgen / libmonosgen-2.0.so using objdump -x
7. observe two unexpected entries:
  a. RPATH set to the libdir of libstdc++ of the gcc version used.
  b. NEEDED libstdc++.so.6

When compiling using the system default gcc (i.e. the one installed in /usr/bin/gcc) the RPATH and NEEDED are not part of the mono binaries.

### Current Behavior

RPATH is set, while `--disable-rpath` was provided on the command line
NEEDED libstdc++

### Expected Behavior
RPATH is not set
libstdc++ is not needed.

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

Mono 5.16.0.220

RHEL6.9:
with the default: gcc 4.4.7 / binutils 2.21: no RPATH, no libstdc++ linked
with the non-default: gcc 4.8.4 / binutils 2.25: RPATH set, libstdc++ linked.

RHEL7.4:
with the default gcc 4.8.5 / binutils 2.25: no RPATH, no libstdc++ linked
with the non-default gcc 4.9.3 / binutils 2.2.5: RPATH set, libstdc++ linked.

The RPATH and libstdc++ are problematic for us as we pinvoke into native libraries. These libraries have a dependency on libstdc++ as well, but possibly a newer version. Now suddenly mono has an RPATH this means that the loader uses this RPATH to resolve libstdc++, and resolves wrong. The error thrown by mono on the pinvoke is "libxyz.so not found", which is a redherring for "one of the dependencies doesn't match, i.e. libstdc++".
We compiled with --disable-rpath to keep this under the applications control by allowing the application's scripts to set LD_LIBRARY_PATH, but apparently this flag is not honored by mono build anymore.
The dependency on libstdc++ is surprising as I understood that the recent refactorings have tried to prevent linking to libstdc++.

[...]

PR #12089, opened 2018-12-17
For the files that are always C++, for LLVM support, instead of compi…

jaykrell:
…ling foo.cpp,

instead compile foo.c, that includes foo.cpp, with -xc++.

This is to prevent a C++ library dependency -- even when mono is compiled as C, due to the LLVM support -- and fix https://github.com/mono/mono/issues/12060.

There might be another smaller way.
Let me try to understand this:
https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1041,38 +1041,51 @@ fi
 
 AC_SUBST(CXXFLAGS_COMMON)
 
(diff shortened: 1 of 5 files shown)


jaykrell:
Try one-liner inhttps://github.com/mono/mono/pull/12091 instead.

PR #12091, opened 2018-12-17
Avoid C++ runtime dependency despite LLVM .cpp files.

jaykrell:
This is much reduced form of https://github.com/mono/mono/pull/12089.
See https://github.com/mono/mono/issues/12060.
See https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html.

### DIFF
--- a/mono/mini/Makefile.am.in
+++ b/mono/mini/Makefile.am.in
@@ -707,6 +707,11 @@ endif
 libmini_la_SOURCES = $(common_sources) $(llvm_sources) $(llvm_runtime_sources) $(arch_sources) $(os_sources)
 libmini_la_CFLAGS = $(AM_CFLAGS) @CXX_ADD_CFLAGS@
 
+# Avoid C++ runtime dependency despite the LLVM .cpp files.
+# See https://github.com/mono/mono/issues/12060.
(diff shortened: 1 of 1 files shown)


lambdageek:
@monojenkins build

jaykrell:
I'm a little nervous this breaks some LLVM configurations. i.e. is there a dependency?
It certain seems right for when no LLVM configured though.
I tested configure -enable-llvm.
CI will test some.
I don't think CI will test LLVM JIT though, or if we still have a directly linked LLVM scenario (vs. loaded LLVM).

jaykrell:
@monojenkins build OS X i386

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lambdageek:
Took a shot at it https://github.com/mono/mono/pull/12266

Jay's idea about splitting up C and C++ sources in mini was the right one - messing with `..._LINK` is fragile (because you lose `..._LDFLAGS` if you're not careful)

PR #12101, opened 2018-12-18
[runtime] Add mode for direct static+dedup calls

alexanderkyte:
This only runs on bitcode, because those have self-initing functions. Haven't really tested this, should be useful for wasm. 

@vargaz, you spot anything off? It'll bloat the exports symbol table for the dedup module, but that can be stripped in theory. Don't really think it's possible to get around that while using the native linker. 

It doesn’t do direct calls for all calls, just adds direct calls to methods in the dedup module. I'm not really sure if we can mangle all the non-deduped stuff. Should be fine, just hasn't been tested before.

When I'm looking at optimizing LLVM stuff, I can look more into it. 

If anybody has a bit of free time, might be worth poking at. If the speedups from this experiment are noticeable, doing it for all calls might be worth it.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -196,6 +196,7 @@ typedef struct MonoAotOptions {
 	char *gen_msym_dir_path;
 	gboolean direct_pinvoke;
 	gboolean direct_icalls;
+	gboolean direct_cross_amod;
 	gboolean no_direct_calls;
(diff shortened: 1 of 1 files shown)


alexanderkyte:
Closing, @vargaz mentioned he was already tackling this issue in mini-llvm.c. 

We can reopen this if this approach is desired.

PR #12114, opened 2018-12-18
[debugger] Fix test ShouldCorrectlyStepOverOnExitFromArgsAfterStepInMethodParameter

### LINKED ISSUE #7326: Incorrect step over in debugger, test ShouldCorrectlyStepOverOnExitFromArgsAfterStepInMethodParameter
## Steps to Reproduce

1. Run test ShouldCorrectlyStepOverOnExitFromArgsAfterStepInMethodParameter

## Current Behavior
Test fail
## Expected Behavior
Test pass

thaystg:
The problem was in this line:
ss_nested_with_two_args(ss_nested_arg (), ss_nested_arg ());
The first step_into works correctly entering in the ss_nested_arg method, but the second one doesn't work.
The problem was because this flag MONO_SEQ_POINT_FLAG_NONEMPTY_STACK was set, so the debugger don't consider that it was a possible break.
The solution of the issue was not passing this flag in this situation, there was another place that does the same check that I did.
Fixes #7326


mono/mini/method-to-ir.c
```diff
@@ -7819,7 +7819,10 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			ins_flag = 0;
 			constrained_class = NULL;
 			if (need_seq_point)
-				emit_seq_point (cfg, method, next_ip, FALSE, TRUE);
+			{
+				gboolean sym_seq_point = sym_seq_points && mono_bitset_test_fast (seq_point_locs, ip - header->code);
+				emit_seq_point (cfg, method, next_ip, FALSE,((sp == stack_start) || !sym_seq_point));
+			}
```
vargaz: Shouldn't this be sp != stack_start ?

mono/mini/method-to-ir.c
```diff
@@ -7819,7 +7819,10 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			ins_flag = 0;
 			constrained_class = NULL;
 			if (need_seq_point)
-				emit_seq_point (cfg, method, next_ip, FALSE, TRUE);
+			{
```
vargaz: Use K&R style indentation.

mono/mini/method-to-ir.c
thaystg: Hi, Vargaz, maybe you are right, now I checked if they are nested calls in a different way. Thanks for the review.

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
```diff
@@ -581,11 +583,27 @@ public static void ss7_3 ()
 		ss_nested_with_two_args(ss_nested_arg (), ss_nested_arg ());
 	}
 
+	[MethodImplAttribute (MethodImplOptions.NoInlining)]
+	public static void ss_nested_with_three_args_wrapper () {
+		ss_nested_with_three_args(ss_nested_arg (), ss_nested_arg (), ss_nested_arg ());
```
lewurm: is there any order expected? should we rename it rather to `ss_nested_arg1`, `ss_nested_arg2` etc. and verify that in the test?

thaystg: Yes, yes, good idea, I will change the names!

vargaz:
Looks ok. How does coreclr handle this situation ?

thaystg:
> Looks ok. How does coreclr handle this situation ?

I didn't look at the source code because Ludovic said me that this wouldn't be productive. But the behavior in VS for Windows with coreclr is after the step out, you can do an step into in the next function in the same line. In VS for MAC, when you do an step out it goes directly to the next line so it's not possible to do an step into in the next function.

mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
thaystg: Done!

thaystg:
@monojenkins build failed

mfilippov:
@thaystg @vargaz Is it possible to backport this fix into 5.18 bugfix? It is a very painful problem in a debugger.

thaystg:
> @thaystg @vargaz Is it possible to backport this fix into 5.18 bugfix? It is a very painful problem in a debugger.

I'm new in the team and I'm not sure about which types of bugfix they backport, let's wait for @vargas opinion.

marek-safar:
This should go down to 2018-10, 2018-08 looks risky to me

thaystg:
@monojenkins backport 2018-10

mfilippov:
@marek-safar @thaystg Greate thanks!

thaystg:
@monojenkins backport 2018-12

PR #12115, opened 2018-12-18
[2018-08] Bump msbuild to track xplat-master

radical:


radical:

PR #12116, opened 2018-12-18
[coop] Add high level suspend policy accessors

lambdageek:
- mono_threads_suspend_policy_name
- mono_threads_suspend_policy_are_safepoints_enabled
- mono_threads_suspend_policy_is_multiphase_stw_enabled
- mono_threads_suspend_policy_is_blocking_transition_enabled

The normal accessors (mono_threads_are_safepoints_enabled, etc) just apply the
policy acessors to the current policy

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -577,7 +577,7 @@ threads_suspend_policy_getenv (void)
 
 static char threads_suspend_policy;
 
-static MonoThreadsSuspendPolicy
+MonoThreadsSuspendPolicy
 mono_threads_suspend_policy (void)
 {
 	int policy = threads_suspend_policy;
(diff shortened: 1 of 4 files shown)


lambdageek:
@monojenkins build

mono/utils/mono-threads-coop.c
```diff
@@ -640,6 +639,36 @@ mono_threads_suspend_policy_name (void)
 	}
 }
 
+static gboolean
+blocking_transition_getenv_compat (void)
+{
+	static int inenv = -1;
+	if (G_UNLIKELY (inenv == -1))
+		inenv = g_hasenv ("MONO_ENABLE_BLOCKING_TRANSITION");
```
luhenry: We should remove support for this env variable, it's not really useful and not even documented.

PR #12117, opened 2018-12-18
Handle MONO_TYPE_TYPEDBYREF return values

lewing:
MonoTests.System.TypedReferenceTest.MakeTypedReference was
aborting in wasm due to missing TypedReference support.


lewurm:
looks okay.  why was it closed?

PR #12118, opened 2018-12-18
Bump to latest SQLite for Mac bundle

directhex:
Bump Mac bundle from SQLite 3.9.2 to 3.26.0


directhex:
@monojenkins build pkg

akoeplinger:
I tested with the script from Duncan and sqlite no longer crashes with this new version.

@directhex do you know how far back we need to backport it?

directhex:
@akoeplinger whatever version is in the next 16.0 preview. But *NOT* what we use in 15.x

akoeplinger:
@monojenkins backport 2018-12
@monojenkins backport 2018-10
@monojenkins backport 2018-08

PR #12119, opened 2018-12-18
[coop] Always emit safepoints and blocking transitions, except on WASM

lambdageek:
Always emit safepoints and blocking transitions on all targets except WASM.

Addresses #11765.

---

This is an alternative to #11833 which was over-engineered.  The rationale is that we should always emit the machinery at compile time and just ignore it at runtime if the host suspend policy says so.

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -1706,17 +1706,18 @@ typedef struct EmitGCSafeTransitionBuilder {
 static gboolean
 gc_safe_transition_builder_init (GCSafeTransitionBuilder *builder, MonoMethodBuilder *mb, gboolean func_param)
 {
-	if (mono_threads_is_blocking_transition_enabled ()) {
-		builder->mb = mb;
-		builder->func_param = func_param;
-		builder->coop_gc_stack_dummy = -1;
-		builder->coop_gc_var = -1;
+	builder->mb = mb;
+	builder->func_param = func_param;
+	builder->coop_gc_stack_dummy = -1;
+	builder->coop_gc_var = -1;
 #ifndef DISABLE_COM
-		builder->coop_cominterop_fnptr = -1;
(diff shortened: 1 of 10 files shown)


vargaz:
Won't this lead to perf degradation on platforms which don't need safepoints right now ?

lambdageek:
@vargaz 

Safepoints:  the `mono_polling_required` global will always be false so the cost is a conditional branch to avoid the call to `mono_threads_state_poll`.

Thread state transitions: there will be an extra call to a known function that will immediately exit.

vargaz:
The call itself will force the compiler to kill all values in registers, so it will lead to a perf degradation even if the branch is always false. It will also increase code size somewhat.

luhenry:
@vargaz we already made our mind on that. Since we are moving to cooperatively suspend as the default, we require to generate safepoints. This approach just makes it much easier, and actually correct.

mono/metadata/marshal-ilgen.c
```diff
@@ -1706,17 +1706,17 @@ typedef struct EmitGCSafeTransitionBuilder {
...
 gc_safe_transition_builder_init (GCSafeTransitionBuilder *builder, MonoMethodBuilder *mb, gboolean func_param)
 {
-	if (mono_threads_is_blocking_transition_enabled ()) {
-		builder->mb = mb;
-		builder->func_param = func_param;
-		builder->coop_gc_stack_dummy = -1;
-		builder->coop_gc_var = -1;
+#if defined (TARGET_WASM)
+	return FALSE;
+#endif
```
luhenry: That will likely makes compiler complain. Using `#else` here would fix it.

vargaz:
We should only take the perf hit when coop suspend is going to be used at runtime which is not the case on ios afaik right now.

jaykrell:
Call should not kill all registers, “only” volatile ones.
This could be mitigated on some architectures by marking & making the call “preserveall” right?

Issue #11956, opened 2018-12-06
os-event-unix.c:82 assert(event) failure


jaykrell:
## Steps to Reproduce

1.  PR

### Current Behavior

https://jenkins.mono-project.com/job/test-mono-pull-request-coop/10660/testReport/junit/MonoTests/test-unhandled-exception-2-255-with-managed-handler/unhandled_exception_2_exe/

```
Thread 1 (Thread 0x7fc3f6123740 (LWP 21274)):
#0  0x00007fc3f55f1b3a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x000055d64e68c540 in dump_native_stacktrace (ctx=ctx@entry=0x7fffebab31c0, signal=0x55d64e8a2656 "SIGABRT") at mini-posix.c:1051
#2  0x000055d64e68c710 in mono_dump_native_crash_info (signal=signal@entry=0x55d64e8a2656 "SIGABRT", ctx=ctx@entry=0x7fffebab31c0, info=info@entry=0x7fffebab32f0) at mini-posix.c:1080
#3  0x000055d64e61e005 in mono_handle_native_crash (signal=0x55d64e8a2656 "SIGABRT", ctx=0x7fffebab31c0, info=0x7fffebab32f0) at mini-exceptions.c:3222
#4  <signal handler called>
#5  0x00007fc3f505dfff in raise () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007fc3f505f42a in abort () from /lib/x86_64-linux-gnu/libc.so.6
#7  0x000055d64e865214 in mono_log_write_logfile (log_domain=<optimized out>, level=<optimized out>, hdr=<optimized out>, message=0x55d64ffa4cb0 "* Assertion at os-event-unix.c:82, condition `event' not met\n") at mono-log-common.c:136
#8  0x000055d64e87df8e in monoeg_g_logstr (msg=<optimized out>, log_level=G_LOG_LEVEL_ERROR, log_domain=0x0) at goutput.c:117
#9  monoeg_g_logv_nofree (log_domain=log_domain@entry=0x0, log_level=log_level@entry=G_LOG_LEVEL_ERROR, format=format@entry=0x55d64e887070 "* Assertion at %s:%d, condition `%s' not met\n", args=args@entry=0x7fffebab3a18) at goutput.c:128
[...]

PR #12121, opened 2018-12-18
Allow urls with a blob scheme to pass validation in WASM

### LINKED ISSUE #11681: [wasm] Add support for blob: URLs
Please add support in WASM for fetching blob: URLs via HttpClient. (https://www.w3.org/TR/FileAPI/#url)

As far as I can tell, the [HttpRequestMessage](https://github.com/mono/mono/blob/6b858ef9d75a5aaea344529b640eeb30d4d78740/mcs/class/System.Net.Http/System.Net.Http/HttpRequestMessage.cs#L99) url validation is what is preventing this.

I made a [hacky demo](https://github.com/jburman/BlazorLocalFile/tree/master/src/BlazorLocalFile) with Blazor (by hooking in at browser's fetch() method) to show that files can be loaded in that way.

lewing:
Relax HttpMessageRequest url scheme validation for wasm so that urls with a blob scheme make it to the host backend. Fixes #11681

note: the browser tests fail when running chrome 71 in headless mode, they work in chrome 70 and 72

### DIFF
--- a/mcs/class/System.Net.Http/System.Net.Http/HttpRequestMessage.cs
+++ b/mcs/class/System.Net.Http/System.Net.Http/HttpRequestMessage.cs
@@ -101,6 +101,11 @@ namespace System.Net.Http
 			if (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)
 				return true;
 
+#if WASM
+			if (uri.Scheme == "blob")
+				return true;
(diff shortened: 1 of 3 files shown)


lewing:
@monojenkins build failed

akoeplinger:
The coop failure is unrelated (https://github.com/mono/mono/issues/11956), merging.

PR #12124, opened 2018-12-19
Skip TypedReferenceTest fixture for WASM as it causes a runtime abort

lewing:
The sdks/wasm run-all-corlib target currently aborts in the TypedReferenceTest fixture. Skip it in that case by adding it to the NotWasm category.

### DIFF
--- a/mcs/class/corlib/Test/System/TypedReferenceTest.cs
+++ b/mcs/class/corlib/Test/System/TypedReferenceTest.cs
@@ -34,6 +34,7 @@ namespace MonoTests.System
 {
 #if !MONODROID // Type load segfaults the runtime on ARM64 (https://gist.github.com/grendello/334d06c45376602a9afc)
 	[TestFixture]
+	[Category("NotWasm")]
 	public class TypedReferenceTest
 	{
 		struct TestFields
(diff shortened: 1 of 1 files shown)


lewing:
The monodroid ifdef seems wrong.  The comment makes it sound arm64 specific.  Is there a category that makes sense here?

mcs/class/corlib/Test/System/TypedReferenceTest.cs
```diff
@@ -34,6 +34,7 @@ namespace MonoTests.System
 {
 #if !MONODROID // Type load segfaults the runtime on ARM64 (https://gist.github.com/grendello/334d06c45376602a9afc)
 	[TestFixture]
+	[Category("NotWasm")]
```
alexischr: maybe add `// currently causes the runtime to abort`? (I can't find the "Suggested change" github function)

I will do the merge right after, no need to wait for all the checks

alexischr:

mcs/class/corlib/Test/System/TypedReferenceTest.cs
lewing: added the comment

PR #12129, opened 2018-12-19
[wasm] Run patch with -N so it doesn't fail if called more than once.

vargaz:


vargaz:
@monojenkins build failed

alexischr:

lewing:
@monojenkins build failed

PR #12137, opened 2018-12-19
[w32file-unix] Remove unnecessary GC-Safe transitions when calling ut…

alexischr:


jaykrell:
For that matter, why are they *not* present around the `access` in the nearby `mono_portability_find_file_internal`? Or all calls to `malloc`?
Or rather, pretty much all runtime exits, that might take a lock?

alexischr:
These calls do not return `EINTR` according to POSIX[1], Linux and FreeBSD (the Darwin docs seem to be down) and are meant to be really fast (nanosecond-resolution file timestamping, sometimes to current time); it's hard to have proof-positive they are not blocking on any platform as POSIX doesn't specify but it's hard to imagine them not being time-bound given the above. There is symlink resolution according to docs so this can't be constant-time, but the `ELOOP` error makes it clearly that this processing is also bound. 

What brought this change is just reviewing the code as the surrounding code needs to be transplanted to CoreFX and it'll be a bit nicer if these are removed first. The potential of bottlenecking here I think is important too: it's easy for build tools to loop through a directory setting timestamps and that would result in thousands of GC transitions (these functions are being used from managed-land System.IO via the icalls of the internal MonoIO class )

[1] http://pubs.opengroup.org/onlinepubs/007904875/functions/utimes.html, http://pubs.opengroup.org/onlinepubs/000095399/functions/utime.html

I will put the above in the PR description

alexischr:
@jaykrell that is the discussion of "what is a blocking call" and it is tricky, so I tried to argue my case deliberately :)

jaykrell:
They can likely block. Other processes can be accessing the files concurrently.
The precision of the timestamp and the lack of EINTR does not affect blockingness.
Also they could be going over the network for networked file systems.
The calls that take paths, instead of int filedes, are affectively doing an open.

jaykrell:
Agreed file enumerating could be doing this many times. Can the transition be lifted out of the loop?

jaykrell:
Maybe lack of EINTR is blockingness? i.e. anything that is slow, is interruptible/cancelable?
NT doesn't have EINTR but it does have cancelable I/O. I think these are exactly analogous.

jaykrell:
But there are certainly locks in syscalls that are not I/O.
Memory allocation for example can lock, and be slow under heavy load.

jaykrell:
I see point though, that open is kinda bounded and read/write can be quite large.

PR #12142, opened 2018-12-19
[runtime] Avoid typedref returns from icalls, wasm doesn't support them.

vargaz:


vargaz:

mono/metadata/icall.c
```diff
@@ -7743,45 +7736,42 @@ mono_ArgIterator_IntGetNextArgType (MonoArgIterator *iter)
...
+		MonoClassField *f;
+		MONO_HANDLE_ARRAY_GETVAL (f, fields, MonoClassField*, i);
 
-	for (i = 0; i < mono_array_length_internal (fields); ++i) {
-		f = mono_array_get_internal (fields, MonoReflectionField*, i);
 		g_assert (f);
 
 		if (i == 0)
-			p = (guint8*)target + f->field->offset;
+			p = (guint8*)MONO_HANDLE_RAW (target) + f->offset;
```
lambdageek: Using `MONO_HANDLE_RAW` in this loop isn't great.  I think (via asm loading) `mono_class_from_mono_type_internal` can end up invoking managed code, so we don't want to get the address from the handle too early.

How about:
 Change from `guint8 *p` to `intptr_t field_offset`.  In the loop just compute `field_offset` from all the `f->offset`s.
then at the end `res->value = (guint8*)MONO_HANDLE_RAW (target) + field_offset`.

vargaz: Fixed.

vargaz:
@monojenkins build failed

mcs/class/corlib/System/ArgIterator.cs
```diff
@@ -86,22 +86,30 @@ public TypedReference GetNextArg ()
...
+			TypedReference result = new TypedReference ();
+			unsafe {
+				IntGetNextArg (&result);
+			}
+			return result;
 		}
 
 		[MethodImpl (MethodImplOptions.InternalCall)]
-		extern TypedReference IntGetNextArg ();
+		extern unsafe void IntGetNextArg (void *res);
```
luhenry: Shouldn't this have used `ref TypedReference res` to make it "safe" code.

vargaz: The corefx/coreclr code uses this convention for some reason.

Issue #8922, opened 2018-05-29
Generic Error [GDI+ status: GenericError] in NativeWindow.WndProc


sancheolz:
## Steps to Reproduce

1.  Open sln in Monitor.zip or run this code

```
using System;
using System.Windows.Forms;

namespace Monitor
{
    internal static class Program
    {
        [STAThread]
        private static void Main(string[] args)
        {
                Application.EnableVisualStyles();
                Application.SetCompatibleTextRenderingDefault(false);
                Application.Run(new MainForm());
        }
    }

	public class MainForm : Form
    {
        public MainForm()
        {
            InitializeComponent();
        }
        private System.ComponentModel.IContainer components = null;

        protected override void Dispose(bool disposing)
        {
            if (disposing && (components != null))
            {
                components.Dispose();
            }
            base.Dispose(disposing);
        }

        private void InitializeComponent()
        {
            this.components = new System.ComponentModel.Container();
            this.tableLayoutPanel4 = new System.Windows.Forms.TableLayoutPanel();
            this.cbProfileSelect = new System.Windows.Forms.Button();
            this.flowLayoutPanel2 = new System.Windows.Forms.FlowLayoutPanel();
            this.btnLocatedOnTerritory = new System.Windows.Forms.Button();
            this.tableLayoutPanel4.SuspendLayout();
            this.flowLayoutPanel2.SuspendLayout();
            this.SuspendLayout();
            // 
            // tableLayoutPanel4
            // 
            this.tableLayoutPanel4.AutoSize = true;
            this.tableLayoutPanel4.Controls.Add(this.cbProfileSelect, 0, 0);
            this.tableLayoutPanel4.Size = new System.Drawing.Size(123, 45);
            // 
            // flowLayoutPanel2
            // 
            this.flowLayoutPanel2.AutoSize = true;
            this.flowLayoutPanel2.BorderStyle = System.Windows.Forms.BorderStyle.FixedSingle;
            this.flowLayoutPanel2.Controls.Add(this.btnLocatedOnTerritory);
            this.flowLayoutPanel2.Controls.Add(this.tableLayoutPanel4);
            this.flowLayoutPanel2.Location = new System.Drawing.Point(0, 0);
            this.flowLayoutPanel2.Size = new System.Drawing.Size(363, 1962);
            // 
            // MainForm
            // 
            this.Controls.Add(this.flowLayoutPanel2);
            this.Name = "MainForm";
            this.tableLayoutPanel4.ResumeLayout(false);
[...]

PR #12161, opened 2018-12-20
System.Windows.Forms: Fix incorrect layout for table layout nested in flow layout

filipnavara:
Fixes issue #8922. The flow layout code specifies `int.MaxValue` as width in `GetPreferredSize` call, which the table layout didn't handle properly and tried to use all the available space.

Relevant WinForms open source code that implements the same behavior:

https://github.com/dotnet/winforms/blob/e5606060fa95bdb40af9ddeceb7bea81535de291/src/System.Windows.Forms/src/System/Windows/Forms/Layout/FlowLayout.cs#L183-L188

https://github.com/dotnet/winforms/blob/e5606060fa95bdb40af9ddeceb7bea81535de291/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs#L700-L702

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/TableLayout.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms.Layout/TableLayout.cs
@@ -483,7 +483,7 @@ namespace System.Windows.Forms.Layout
 			CalculateColumnWidths (settings, actual_positions, max_colspan, settings.ColumnStyles, auto_size, column_widths, false);
 
 			// Calculate available width
-			int available_width = size.Width - (border_width * (columns + 1));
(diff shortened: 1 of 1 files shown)


marek-safar:
@monojenkins backport 2018-12

marek-safar:
@monojenkins backport 2018-10

sancheolz:
@monojenkins backport 2018-04

akoeplinger:
@sancheolz the 2018-04 (== Mono 5.14) release is done at this point and we won't touch it anymore.

sancheolz:
@akoeplinger we need this version because Remoting deserialization broken for DataSet tables in latest versions. We have legacy project which uses Remoting and Winforms. And gdi bug is crashes its application. Ok I will backport it by my own

PR #12178, opened 2018-12-22
[Coop] Convert and consolidate {remoting,cominterop}type_from_handle.

jaykrell:
mostly into mono_type_from_handle.
However keep them separate for registration/patching sake, since
will error if register twice, else need to arrange to register once.

### DIFF
--- a/mono/metadata/reflection.c
+++ b/mono/metadata/reflection.c
@@ -3092,19 +3092,10 @@ mono_class_from_mono_type_handle (MonoReflectionTypeHandle reftype)
 	return mono_class_from_mono_type_internal (MONO_HANDLE_RAW (reftype)->type);
 }
 
-// This is an icall, it will return NULL and set pending exception on failure.
-MonoReflectionType*
-mono_type_from_handle (MonoType *handle)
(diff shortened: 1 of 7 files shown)


jaykrell:
@monojenkins build failed

jaykrell:

mono/metadata/reflection.c
```diff
@@ -3090,3 +3090,20 @@ mono_class_from_mono_type_handle (MonoReflectionTypeHandle reftype)
 {
 	return mono_class_from_mono_type_internal (MONO_HANDLE_RAW (reftype)->type);
 }
+
+// This is an icall, it will return NULL and set pending exception on failure.
+MonoReflectionType*
+mono_type_from_handle (MonoType *handle)
```
lambdageek: So this is like `mono_type_get_object_{checked,handle}`, except that it also calls `mono_class_init_internal` - it would be better if it followed their naming scheme. (Also I'm not a fan of hiding pending exception setting in helper functions)

jaykrell: I agree the functions need work but first I wanted to combine them w/o change.

PR #12185, opened 2018-12-23
[build] Fixed csproj.tmpl for macos: Conflict in naming of HostPlatform.

galli-leo:
`macOS` was used for checking HostPlatform for OSX, but `darwin` is what the value was set to.
I have changed the template to set HostPlatform to `macOS` on OSX. This fixes the solution and csproj files for OSX. Before, there were a lot of files missing from projects, most notably System.Net did not have many files in there.

Let me know whether I should regenerate the csproj files or if that's done by the CI.


kg:
CI will regenerate them. Change looks good.

kg:
@monojenkins build

marek-safar:
@monojenkins commit csproj

PR #12188, opened 2018-12-23
Reduce cost and uses of JIT icall hashtables.

jaykrell:
This based  on request:
https://github.com/mono/mono/pull/11941#issuecomment-449157323

which really strikes to the point, other than not hashing function addresses, most/all uses of these hashtables don't need a hashtable at all. So this reduces the use of the hashtables significantly, while still maintaining their data unchanged, with  an eye toward either eliminating them entirely, or maybe devolving to a linked list or fixed array, for the remaining enumeration scenarios.

Statically allocate the data.
Reference the data directly often instead of looking up strings in hashtables -- could do more.

This incorporates one way to remove the dynamic icalls for array construction -- instantiate 50ish functions. Another way would be to use localalloc and pass pointer to parameters, instead of vararg.
Both of those are in separate PRs as well. Only one should be chosen. Both were implemented because the first is easier but then the PR asked for the second.


mono/mini/mini-llvm.c
```diff
@@ -3959,27 +3971,27 @@ emit_throw (EmitContext *ctx, MonoBasicBlock *bb, gboolean rethrow, LLVMValueRef
...
-			callee = get_callee (ctx, sig_to_llvm_sig (ctx, throw_sig), MONO_PATCH_INFO_JIT_ICALL, icall_name);
+			callee = get_callee (ctx, sig_to_llvm_sig (ctx, throw_sig), MONO_PATCH_INFO_JIT_ICALL_INFO, icall_info);
 		} else {
 			gpointer target;
 #ifdef TARGET_X86
 			/* 
 			 * LLVM doesn't push the exception argument, so we need a different
 			 * trampoline.
 			 */
-			target = resolve_patch (ctx->cfg, MONO_PATCH_INFO_JIT_ICALL, rethrow ? "llvm_rethrow_exception_trampoline" : "llvm_throw_exception_trampoline");
```
jaykrell: Given "llvm_foo", I change to "mono_llvm_foo". This is a problem if there was already mono_llvm_foo.

jaykrell:

jaykrell:

jaykrell:
Solution for `handle_array_new` is probably to stackalloc, put the params there, pass pointer to first, to have one function and no varargs. And enable compiling with LLVM as well.

jaykrell:
But don't localloc in loop.

vargaz:
You can add the localloc to the end of initlocals_bb.

jaykrell:
How about statically generating 30-ish little functions? I have that almost done.
I admit -- I got stuck and hadn't seen your hint.

jaykrell:
60-ish functions.
https://github.com/mono/mono/pull/12200 will make this simpler.

jaykrell:
`init_localsbb`

jaykrell:
Some functions, very few, are registered with multiple names.
Tying the info to the function instead of the name broke that. Will fix.

jaykrell:
Do I create struct type on-demand of the required size?

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
This works. Need to chose either
https://github.com/mono/mono/pull/12213
or https://github.com/mono/mono/pull/12200
to go with it, so the set of JIT icalls is static.
Currently 12200 is in here.

jaykrell:
I not able to remove/labels but reason for do not merge is gone
or mostly gone, as it works now.

mono/mini/dwarfwriter.c
```diff
@@ -1400,8 +1400,7 @@ disasm_ins (MonoMethod *method, const guchar *ip, const guint8 **endip)
 			MonoJitICallInfo *info;
 
 			token = read32 (ip + 2);
-			data = mono_method_get_wrapper_data (method, token);
-			info = mono_find_jit_icall_by_addr (data);
+			info = (MonoJitICallInfo*)mono_method_get_wrapper_data (method, token);
```
jaykrell: It'd be  furthermore nice to put the pointer in the IL, not a token.

mono/mini/interp/transform.c
```diff
@@ -4850,14 +4850,12 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
...
 					guint32 token;
-					gpointer func;
 					MonoJitICallInfo *info;
 					int icall_op;
 
 					token = read32 (td->ip + 1);
 					td->ip += 5;
-					func = mono_method_get_wrapper_data (method, token);
-					info = mono_find_jit_icall_by_addr (func);
+					info = (MonoJitICallInfo*)mono_method_get_wrapper_data (method, token);
```
jaykrell: It'd be  furthermore nice to put the pointer in the IL, not a token.

mono/mini/method-to-ir.c
```diff
@@ -10098,12 +10089,8 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 
 		case MONO_CEE_MONO_ICALL: {
 			g_assert (method->wrapper_type != MONO_WRAPPER_NONE);
-			gpointer func;
-			MonoJitICallInfo *info;
-
-			func = mono_method_get_wrapper_data (method, token);
-			info = mono_find_jit_icall_by_addr (func);
-			if (!info)
+			MonoJitICallInfo *info = (MonoJitICallInfo*)mono_method_get_wrapper_data (method, token);
```
jaykrell: It'd be  furthermore nice to put the pointer in the IL, not a token.

jaykrell:
@monojenkins build failed

jaykrell:
Please remove do-not-merge label.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This was ready to go but I was not able to remove the label.
The hashtables can mostly or perhaps completely be eliminated.
This mostly eliminated their use of them as a first step.

PR #12198, opened 2018-12-24
Cleanup mono_array_new_{1,2,3,4}.

jaykrell:
Mainly to remove the dynamic alloca but also combine common code.
If there is a perf concern then forceinline mono_array_new_n.

### DIFF
--- a/mono/mini/jit-icalls.c
+++ b/mono/mini/jit-icalls.c
@@ -737,151 +737,57 @@ mono_array_new_va (MonoMethod *cm, ...)
 	return arr;
 }
 
+static MonoArray *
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
That will not work without additional casting.
The types are chosen to do what the existing code was doing and to exactly match `mono_array_new_full_checked`.

```
mono/metadata/object-internals.h:mono_array_new_full_checked (MonoDomain *domain, MonoClass *array_class, uintptr_t *lengths, intptr_t *lower_bounds, MonoError *error);
```

`uintptr_t` and `gsize` are any of {`unsigned int`, `unsigned int`, `unsigned long long`}, but not necessarily the same type. intptr_t and gssize likewise.

Issue #12193, opened 2018-12-24
problems creating multi dimensional arrays


jaykrell:
Was developing tests for icalls.

```
using System;

class T
{
	static object f0 () { return new int[] {1,2}; }
	static object f1 () { return new int[,] {{2,3}}; }
	static object f2 () { return new int[,,] {{{3}}}; }
	static object f3 () { return new int[,,,] {{{{4}}}}; }
	static object f4 () { return new int[,,,,] {{{{{5}}}}}; }
	static object f5 () { return new int[,,,,,] {{{{{{6}}}}}}; }

	static int Main ()
	{
		Console.Write ("f0 "); f0 ();
		Console.Write ("f1 "); f1 ();
		Console.Write ("f2 "); f2 ();
		Console.Write ("f3 "); f3 ();
		Console.Write ("f4 "); f4 ();
		Console.Write ("f5"); f5 ();
		return 0;
	}
}
```

Works on desktop.
Mono:

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.OutOfMemoryException: Out of memory
  at (wrapper managed-to-native) System.Object.__icall_wrapper_ves_array_new_va_5(intptr,intptr,intptr,intptr,intptr,intptr)
  at T.f5 () [0x00001] in <f2c01a052b9046698116be6a6b64fb3e>:0 
  at T.Main () [0x00061] in <f2c01a052b9046698116be6a6b64fb3e>:0 
```

```
HB30-VM-Eklavya:tests jay$ mono -V
Mono JIT compiler version 5.23.0.68 (master/5870948f569 Wed Dec 12 19:21:08 EST 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
HB30-VM-Eklavya:tests jay$ which mono
/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono
```

Initializing arrays of higher rank is an odd path in the runtime, using JIT icalls in a unique way, and varargs.

jaykrell:
The results are inconsistent -- suggesting uninitialized data being used.

jaykrell:
`lengths` comes through as garbage at least the last one in f5.

[...]

PR #12200, opened 2018-12-24
Change array construction with rank>4 to not use varargs.

jaykrell:
This fixes https://github.com/mono/mono/issues/12193.

There are a few other choices.

 - Debug the perhaps incorrect varargs passing.

 - localalloc and store the pointers there and pass by address.
   Being sure not to do it in a loop.
   See https://github.com/mono/mono/issues/12193.

 - Not using varargs has other benefits.
   It does not fallback from LLVM to mini.
   It leads to JIT icalls being a static set.
   localalloc has those properties also.


jaykrell:
Relevant failures but fixable.

jaykrell:
Understood, will try that.

jaykrell:
I don't mind the "ugly" per se, but macros provide for slightly larger object code that does somewhat less (the 64 limit) and is probably no faster, and rarely used. localloc will be just as fast, more general, smaller. Just require churn at a different level.

jaykrell:
Very strange what this breaks..

jaykrell:

jaykrell:

jaykrell:

PR #12204, opened 2018-12-25
Remove string parsing, hashing, locking, and dynamic allocations from JIT icall signatures.

jaykrell:
Make them statically allocated, single-instanced at compile time, and almost statically initialized.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -109,6 +109,7 @@
 #endif
 #include "icall-decl.h"
 #include "mono/utils/mono-threads-coop.h"
+#include "mono/metadata/icall-signatures.h"
 
 //#define MONO_DEBUG_ICALLARRAY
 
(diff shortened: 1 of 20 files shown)


mono/metadata/icall.c
```diff
@@ -8523,111 +8524,105 @@ mono_lookup_icall_symbol (MonoMethod *m)
...
-	mono_image_lock (corlib);
-	res2 = (MonoMethodSignature *)g_hash_table_lookup (corlib->helper_signatures, sigstr);
-	if (res2)
-		res = res2; /*Value is allocated in the image pool*/
-	else
-		g_hash_table_insert (corlib->helper_signatures, (gpointer)sigstr, res);
-	mono_image_unlock (corlib);
+void
+mono_create_icall_signatures (void)
+{
```
jaykrell: A lock around this function might be a good idea.
But this, and the prior code, run very early, and there are no later mutations in the prior or current code -- the array_va stuff avoids this path for example, it has its own separate synchronized cache of its own signatures (which can go away if I get the other stuff working).
I think mono_init should have a giant lock.

jaykrell:
So much excess startup code and cost, the same strings handled repeatedly when none are needed at all.

PR #12210, opened 2018-12-26
Register two separate functions for mono_fconv_u4 and mono_fconv_u8

jaykrell:
instead of each function twice.

Elsewhere there is single instancing based on function address
and registering the same function twice breaks things.

What breaks is that due to the single instancing, based on address, the second name is skipped, whatever happens to be second in hash order. Sometimes that is ok, sometimes not.

### DIFF
--- a/mono/mini/jit-icalls.c
+++ b/mono/mini/jit-icalls.c
@@ -963,6 +963,17 @@ mono_fconv_u8 (double v)
 }
 
 guint64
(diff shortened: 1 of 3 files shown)


jaykrell:
It should be possible but a bit tricky to write a test to show the problem.

jaykrell:

alexischr:
Nice, been seeing this one for a while :)

PR #12217, opened 2018-12-28
[Debugger] Record time it took between steps.

### LINKED ISSUE #8460: [Debugger] Record time it took between steps
This is a feature request for SDB.

It would be nice for SDB to record the time between resume and suspend on debuggee side (to avoid overhead of debug machinery and tcp communication as much as possible). The value would be shown to the user via the debugger client.

It could be used as a poor man's profiler (showing the user time spent on each step).

The debugger suspend/step events should include time spent in user-code from breakpoint to breakpoint.

image

cc @kumpera @vargaz

thaystg:
I implemented a new message to avoid breaking the protocol. 
I used the MonoStopwatch to count the time between the steps, it was already used in the debugger-agent.
There is a new unit test to check this feature.
Fixes #8460


thaystg:
@monojenkins commit apidiff

thaystg:
@monojenkins build failed

alexischr:
@monojenkins commit apidiff

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
```diff
@@ -30,11 +30,20 @@ public class ThreadMirror : ObjectMirror
 			return frames;
 		}
 
+		public long ElapsedTime () {
+			long elapsedTime = GetElapsedTime ();
```
alexischr: How come this is split into also an internal method (ie. why not `return vm.conn.Thread_GetElapsedTime (id)`)?

thaystg: I don't have a good reason I was following the other implementations, but I really think that this is an unnecessary overhead. Should I change this and do a new PR?

jbevain:
@vargaz @thaystg @alexischr @marek-safar heads-up, this was merged without bumping the protocol version, introducing a potential binary protocol breakage.

This version of the server and client should be bumped.
The client should check that server has the supported version before calling the server to get the elapsed time.

marek-safar:
@thaystg could you address above ^

thaystg:
> @thaystg could you address above ^

Yes!

PR #12221, opened 2018-12-29
Add WebAssembly MSBuild targets

mhutch:


sdks/wasm/sdk/Mono.WebAssembly.Build/Mono.WebAssembly.Build.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <Description>Mono.WebAssembly build targets. This package is not intended to be referenced directly.</Description>
+    <BuildOutputTargetFolder>build\netstandard2.0</BuildOutputTargetFolder>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" PrivateAssets="all" />
+    <PackageReference Include="System.CodeDom" PrivateAssets="all" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="..\..\..\out\wasm-bcl\wasm_tools\monolinker.exe" Link="monolinker.exe" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath="$(BuildOutputTargetFolder)" />
```
kg: Is PreserveNewest what we want here? I think we want CopyAlways so that you get the correct, consistent set of outputs if you have a latest-master checkout and you revert to an older rev, etc

sdks/wasm/sdk/Mono.WebAssembly.Build/Mono.WebAssembly.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <ItemGroup Condition=" '$(EnableDefaultItems)' == 'true' And '$(EnableDefaultWebContentItems)' == 'true' ">
+    <_WasmItem Include="**\*.html;**\*.png;**\*.jpg;**\*.css;**\*.js" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <Content Include="@(_WasmItem)" />
+    <None Remove="@(_WasmItem)" Condition="'$(EnableDefaultNoneItems)' == 'true'" />
+    <_WasmItem Remove="@(_WasmItem)" />
```
kg: What's the purpose of this?

mhutch: Minor memory optimization, the items aren't needed for anything else.

sdks/wasm/sdk/Mono.WebAssembly.Build/Mono.WebAssembly.Build.csproj
mhutch: Good point, will fix

kjpou1:
@monojenkins build failed

sdks/wasm/sdk/Mono.WebAssembly.Build/WasmLinkAssemblies.cs
```diff
@@ -0,0 +1,162 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
marek-safar: Could we use existing ILLink tasks https://dotnet.myget.org/feed/dotnet-core/package/nuget/ILLink.Tasks instead of building a new one?

(it could be follow up PR)

sdks/wasm/sdk/NuGet.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+	<packageSources>
+		<add key="local" value="packages" />
```
kjpou1: This is giving the following error `The local source 'mono/sdks/wasm/sdk/packages' doesn't exist.`  Is there a way to create this automatically during the restore instead of manually?

sdks/wasm/sdk/Mono.WebAssembly.Build/WasmLinkAssemblies.cs
kjpou1: I am thinking Makayla created a new one because of the special linking of the WebAssembly Bindings and Http.

mhutch: Right. This is just temporary anyways, was going to replace it with a custom task that lets us wire in wasm-specific steps.

sdks/wasm/sdk/NuGet.config
mhutch: Weird, here do you get that error? I don't see it even with a clean build.

kjpou1: Humm, not getting this anymore.  Weird.

Issue #12235, opened 2019-01-02
Crash while profiling roslyn compiler server


kg:
Been trying to profile the compiler server but not having much luck. I get crashes like this sometimes:

```
* Assertion at mono-threads-state-machine.c:95, condition `suspend_count > 0' not met

Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) string.FastAllocateString (int) [0x00007] in <972dc028e30f4795ba71578315b30462>:0
  at string.Concat (string,string) [0x00029] in /home/kate/Projects/mono/external/corefx/src/Common/src/CoreLib/System/String.Manipulation.cs:276
  at string.Concat (object,object) [0x00014] in /home/kate/Projects/mono/external/corefx/src/Common/src/CoreLib/System/String.Manipulation.cs:56
  at Microsoft.CodeAnalysis.CSharp.Conversion.AssertTrivialConversion (Microsoft.CodeAnalysis.CSharp.ConversionKind) [0x00093] in /home/kate/Projects/roslyn/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversion.cs:206
  at Microsoft.CodeAnalysis.CSharp.Conversion.GetTrivialConversion (Microsoft.CodeAnalysis.CSharp.ConversionKind) [0x00001] in /home/kate/Projects/roslyn/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/Conversion.cs:211
  at Microsoft.CodeAnalysis.CSharp.ConversionsBase.FastClassifyConversion (Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbol,Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbol) [0x0001e] in /home/kate/Projects/roslyn/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs:378
  at Microsoft.CodeAnalysis.CSharp.ConversionsBase.ClassifyImplicitConversionFromExpression (Microsoft.CodeAnalysis.CSharp.BoundExpression,Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbol,System.Collections.Generic.HashSet`1<Microsoft.CodeAnalysis.DiagnosticInfo>&) [0x00062] in /home/kate/Projects/roslyn/src/Compilers/CSharp/Portable/Binder/Semantics/Conversions/ConversionsBase.cs:98
[...]

PR #12241, opened 2019-01-02
[coop] Make sure suspend_count is positive in suspend request states

lambdageek:
* Fix bug in `pulse` transition - the suspend count was lost and incorrectly
  swapped to 0.
* `finish_async_suspend` transition - add assertions that the suspend count is
  positive. Try to catch incorrect states sooner.

Fixes https://github.com/mono/mono/issues/12235

PR #12245, opened 2019-01-02
[2018-10] [coop] Make sure suspend_count is positive in suspend request states

lambdageek:
* Fix bug in pulse transition - the suspend count was lost and incorrectly
  swapped to 0.
* finish_async_suspend transition - add assertions that the suspend count is
  positive. Try to catch incorrect states sooner.

Fixes https://github.com/mono/mono/issues/12235

Backport of #12241 to `2018-10`


lewurm:

Issue #8927, opened 2018-05-29
Mixed mode bitcode+ interpreter


kumpera:
This issue tracks the technical problems and proposed solution

The problem with mixed-mode is that the interpreter is entered through a trampoline. As those are not available in bitcode, we use a function descriptor (address + extra arg) instead.

The issue with ftndescs is that the caller must know about it, meaning we must conditionally check for them at every call site and either do a direct call or setup the descriptor and make the indirect call.

alexanderkyte:
I think that the combination of a function to call and a context to pass is universal enough to describe any closure we'll see at runtime. If we need to support the most general call option, it doesn't make sense to say 

``` 
if (mode_1) {
    call_option_one (fn);
}
if (mode_2) {
     call_option_two (fn);
}
if (mode_3) {
ftn_desc = get_tls_val ("key");
most_general_call_option (fn, ftn_desc);
}

When we can put the indirection in the choice of the fn and the ftn_desc (closure state).

In more concrete terms, what I'm suggesting is that when we have mixed-mode interpreter calls enabled, we probably should just call everything through the same interface we call the interpreter calls through.

PR #12248, opened 2019-01-03
Bitcode mixed mode work

vargaz:
Add a new execution mode MONO_AOT_MODE_LLVMONLY_INTERP and make --llvmonly-interp use that. INTERP_LLVMONLY defaults to the interpreter, while LLVMONLY_INTERP defaults to aot.

Contributes to #8927


vargaz:
@monojenkins build failed

Issue #12225, opened 2018-12-30
Is WASM possible without FULL_AOT_RUNTIME


amieres:
Is it possible to compile mono Wasm without the FULL_AOT_RUNTIME flag?

The idea is to enable DefineDynamicAssembly.

migueldeicaza:
CC @lewing

PR #12254, opened 2019-01-03
Enables SRE for wasm

marek-safar:
Addressed also issues like #12225


mcs/class/System.Core/wasm_System.Core.dll.sources
```diff
@@ -1 +1,15 @@
-#include testing_aot_full_System.Core.dll.sources
+#include common_System.Core.dll.sources
+#include dynamic_System.Core.dll.sources
+#include pipes_pns.sources
+
+System.Security.Cryptography/AesCryptoServiceProvider.cs
```
vargaz: Unrelated ?

marek-safar: no, it's part of NS

marek-safar:
@monojenkins commit csproj

Issue #9009, opened 2018-06-03
[s390x] Fails to compile against recent glibc


directhex:
```
  CC       libmonoutils_la-os-event-unix.lo
In file included from mono-threads.h:15:0,
                 from mono-lazy-init.h:19,
                 from os-event-unix.c:14:
../../mono/utils/mono-stack-unwinding.h:116:14: error: field 'ctx' has incomplete type
  MonoContext ctx;
              ^~~
```

Presumably this is similar to https://github.com/mono/mono/commit/9cd9e3b0a13e1f0b757606e5966e1ce27722ed95 i.e. `ucontext` was removed from glibc: https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=commitdiff;h=251287734e89a52da3db682a8241eb6bccc050c9;hp=c86ed71d633c22d6f638576f7660c52a5f783d66

Changing `typedef struct ucontext MonoContext` to `typedef ucontext_t MonoContext` fixes the runtime build - but segfaults with no detail when the runtime tries to actually do anything.

Paging @nealef since nobody else will look at it

nealef:
From memory, MoniContext is identical to mcontext or u context and us defined in one of the header files in utils. I'll take a look at it tomorrow or later today if I can get a moment.


-------- Original message --------

akoeplinger:
Interestingly, this seems to have started working again recently: https://jenkins.mono-project.com/job/test-mono-mainline-community/label=centos-s390x/3096/

directhex:
```(sid_s390x-dchroot)directhex@zelenka:~/monoreal$ mono/mini/mono --debug mcs/class/lib/monolite-linux/1051700017/mcs.exe 
Segmentation fault
```

Still sorta boned. Debian Unstable, mono master, glibc 2.27

nealef:
I'll take a look. I wonder if the vector registers are now part of the context and that's causing grief. I don't have access to a Debian system though. Is it building on centos & sles systems?

directhex:
The runtime is compiling right now on the CI machines, the classlib isn't (not sure why).

As far as I can tell, the build failure in Debian started with glibc 2.26, which is newer than what's in CentOS/SLES

directhex:
If it helps:

```
(sid_s390x-dchroot)directhex@zelenka:~/monoreal$ gdb mono/mini/mono
GNU gdb (Debian 7.12-6+b2) 7.12.0.20161007-git
[...]

PR #12272, opened 2019-01-04
Cater for later glibc use of ucontext

nealef:
Later levels of glibc redefine struct ucontext to typedef struct ucontext_t. This PR will check the level of glibc and define MonoContext appropriately.


nealef:
See issue #9009 for further details

akoeplinger:
@monojenkins backport 2018-10

PR #12288, opened 2019-01-06
Rearrange the TcpStatus enum to match Linux

### LINKED ISSUE #6477: GetActiveTcpConnections->state returns wrong status
I am trying to monitor an active TCP connection with the following code, when running on windows the _log.logInfo function will simply return established but when running on Mono 4.2.1 it will return "Closed", while the connection is established according: netstat -anp --tcp. 

```
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 192.168.200.202:42685   192.168.200.218:18000   ESTABLISHED 23378/mono
```

This is the code that I am using:
```
IPGlobalProperties ipProperties = IPGlobalProperties.GetIPGlobalProperties();
TcpConnectionInformation[] tcpConnections = ipProperties.GetActiveTcpConnections().Where(x => x.LocalEndPoint.Equals(_client.Client.LocalEndPoint) && x.RemoteEndPoint.Equals(_client.Client.RemoteEndPoint)).ToArray();
if (tcpConnections != null && tcpConnections.Length > 0)
{
       TcpState stateOfConnection = tcpConnections.First().State;
      if (_connected != 2)
      {                              
           if (stateOfConnection != TcpState.Established)
           {
                _log.logInfo(tcpConnections.First().RemoteEndPoint.ToString() + "status:" + 
                tcpConnections.First().State.ToString());
                reConnect();
           }
      }
}
```

The output on windows when placing the log function above the if statement: 

```
2018-01-10 20:14:23.767245 | TCP/IP Connector -  : established
```

While the output on Linux is:

```
2018-01-10 20:14:23.767245 | TCP/IP Connector -  : closed
```

I found out that Mono uses the ```/proc/net/tcp``` to read out the TCP/IP connections 
When trying to monitor the connection on remote addr 192.168.200.218  /proc/net/tcp returns the following:

```  
sl  local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode
0: 0100007F:0CEA 00000000:0000 0A 00000000:00000000 00:00000000 00000000   111        0 1077650 1 0000000000000000 100 0 0 10 0
1: 00000000:008B 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 2815396 1 0000000000000000 100 0 0 10 0
2: 00000000:0016 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 17055 1 0000000000000000 100 0 0 10 0
3: 00000000:01BD 00000000:0000 0A 00000000:00000000 00:00000000 00000000     0        0 2815395 1 0000000000000000 100 0 0 10 0
4: 29CAA8C0:0016 DAC8A8C0:DBFB 01 00000080:00000000 01:00000015 00000000     0        0 2906153 3 0000000000000000 25 4 1 10 16
```
Here is number 4 the connection that I am trying to monitor(HEX DA C8 A8 C0  = decimal 192.168.200.218 ) the  state is  01 in the table.

 According the [kernel ](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/tcp_states.h?id=HEAD) state 01 should be established so I am wondering where the state closed is comming from?

BijanT:
The enum in Linux can be seen here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/tcp_states.h?id=HEAD
Fixes #6477

### DIFF
--- a/mcs/class/referencesource/System/net/System/Net/NetworkInformation/TcpState.cs
+++ b/mcs/class/referencesource/System/net/System/Net/NetworkInformation/TcpState.cs
@@ -7,17 +7,17 @@ namespace System.Net.NetworkInformation
     public enum TcpState
     {
         Unknown,
-        Closed,
-        Listen,
+        Established,
         SynSent,
         SynReceived,
-        Established,
         FinWait1,
         FinWait2,
-        CloseWait,
-        Closing,
-        LastAck,
         TimeWait,
+        Closed,
+        CloseWait,
+        LastAck,
+        Listen,
+        Closing,
         DeleteTcb
     }
  }


marek-safar:
We cannot do that it'd be breaking change. See the values assigned in https://docs.microsoft.com/en-us/dotnet/api/system.net.networkinformation.tcpstate?view=netframework-4.7.2

PR #12294, opened 2019-01-07
[aot] Delete the automatically created temporary directory when using --aot=llvm.

vargaz:


vargaz:
@monojenkins build failed

EgorBo:
I used them in order to see the generated LLVM IR, is there an alternative way? AFAIR it's not possible to override `-output` for `opt` via `llvmopts=""`

vargaz:
You can pass --aot=save-temps to avoid it being deleted.

PR #12305, opened 2019-01-07
[2018-10] Preserve old EnableMicrosoftTelemetry signature on backport

alexanderkyte:
Needed to avoid changing the signature of this function within a mono release


luhenry:
We don’t need that. We just need to fix the call in VSfM as part of the integration

PR #12314, opened 2019-01-08
[wasm] Fix build target errors 

kjpou1:
- Fix different errors when building via the nuget instead of locally.
- The nuget generated had a different structure so fixed the output generation to work for both local build as well as nuget build.
- Fix problem with the monolinker.exe not being located correctly.
- Move the monolinker.exe to tools directory within nuget.


kjpou1:

kjpou1:

kjpou1:

PR #12317, opened 2019-01-08
Fix warning and build errors triggered in win64 platform that doesn't define HAVE_CLASSIC_WINAPI_SUPPORT.

lateralusX:
Fix warnings and build errors when bumping Mono on Windows platforms not defining HAVE_CLASSIC_WINAPI_SUPPORT.

Several of the name changes fixes MSVC warnings where local variables hides global variables using the same name.

Number of adjustments not building on platforms not defining HAVE_CLASSIC_WINAPI_SUPPORT.

Fix a couple of build projects not setting dependencies on build-init.vcxproj. This could cause races in build order if projects depend on build initialization.

Added support to force external builds to use msbuild even if ninja is present.

Fix build error when doing MSVC build wihtout cygwin not having a config.h.


mono/metadata/mono-security-windows-uwp.c
```diff
@@ -41,8 +41,8 @@ ves_icall_System_Security_Principal_WindowsIdentity_GetRoles (gpointer token)
 {
 	// FIXME This is now supported by UWP.
 	ERROR_DECL (error); // FIXMEcoop
-	mono_security_win_not_supported ("GetTokenInformation", error);
-	mono_error_set_pending_exception (error); // FIXMEcoop
+	mono_security_win_not_supported ("GetTokenInformation", &error);
```
jaykrell: This doesn't look correct. It should not compile. `ERROR_DECL` makes `MonoError* error`.

mono/metadata/threads.c
```diff
@@ -2202,8 +2203,7 @@ ves_icall_System_Threading_WaitHandle_Wait_internal (gpointer *handles, gint32 n
 	return map_native_wait_result_to_managed (ret, numhandles);
 }
 
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
-
+#ifdef HAVE_API_SUPPORT_WIN32_SIGNAL_OBJECT_AND_WAIT
```
jaykrell: I believe this should be `if` not `ifdef`. The design was you can define 0 or 1.

mono/metadata/w32file-win32.c
```diff
@@ -515,6 +515,8 @@ mono_w32file_get_console_error (void)
 }
 #endif // HAVE_API_SUPPORT_WIN32_GET_STD_HANDLE
 
+#ifdef HAVE_API_SUPPORT_WIN32_GET_FILE_SIZE
```
jaykrell: I believe this should be `if` not `ifdef`. The design was you can define 0 or 1.

mono/metadata/w32subset.h
```diff
@@ -106,3 +112,39 @@
 #define HAVE_API_SUPPORT_WIN32_WAIT_FOR_INPUT_IDLE 1
 #endif
 #endif
+
+#ifndef HAVE_API_SUPPORT_WIN32_SIGNAL_OBJECT_AND_WAIT
```
jaykrell: The intent and prior pattern is 0 or 1. That way another file can define the symbols, not just omit them. Omission is ambiguous. 0 is not.

mono/utils/mono-context.h
```diff
@@ -271,6 +271,11 @@ struct sigcontext {
 #undef MONO_SIGNAL_USE_UCONTEXT_T
 #endif
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable:4324)
```
jaykrell: Add comment as to what the warning is?

mono/utils/mono-dl-windows.c
```diff
@@ -50,20 +51,22 @@ mono_dl_open_file (const char *file, int flags)
 	if (file) {
 		gunichar2* file_utf16 = g_utf8_to_utf16 (file, strlen (file), NULL, NULL, NULL);
 
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#ifdef HAVE_API_SUPPORT_WIN32_SET_ERROR_MODE
```
jaykrell: Should be `if` not `ifdef` so that it can be explicitly stated as not present, instead of ambiguous omission.

mono/utils/mono-os-wait-win32.c
```diff
@@ -168,8 +169,7 @@ mono_win32_wait_for_multiple_objects_ex (DWORD count, CONST HANDLE *handles, BOO
 	return result;
 }
 
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
-
+#ifdef HAVE_API_SUPPORT_WIN32_SIGNAL_OBJECT_AND_WAIT
```
jaykrell: Should be `if` not `ifdef` so that it can be explicitly stated as not present, instead of ambiguous omission.

mono/utils/mono-dl-windows.c
```diff
@@ -72,10 +75,10 @@ mono_dl_open_file (const char *file, int flags)
 		if (!hModule)
 			SetLastError (last_error);
 	} else {
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#ifdef HAVE_API_SUPPORT_WIN32_GET_MODULE_HANDLE
```
jaykrell: Should be `if` not `ifdef` so that it can be explicitly stated as not present, instead of ambiguous omission. Also should be `GetModuleHandleW`. Plain `GetModuleHandle` is a Win9x compatibility layer.

mono/utils/mono-dl-windows.c
```diff
@@ -50,20 +51,22 @@ mono_dl_open_file (const char *file, int flags)
...
 		gunichar2* file_utf16 = g_utf8_to_utf16 (file, strlen (file), NULL, NULL, NULL);
 
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#ifdef HAVE_API_SUPPORT_WIN32_SET_ERROR_MODE
 		guint last_sem = SetErrorMode (SEM_FAILCRITICALERRORS);
 #endif
 		guint32 last_error = 0;
 
-#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#ifdef HAVE_API_SUPPORT_WIN32_LOAD_LIBRARY
```
jaykrell: Should be if not ifdef so that it can be explicitly stated as not present, instead of ambiguous omission. Also should put `W` on functions, instead of using Win9x compat symbols.

mono/metadata/mono-security-windows-uwp.c
lateralusX: I does compile, strange, will look into why that happened.

mono/metadata/mono-security-windows-uwp.c
```diff
@@ -62,17 +62,20 @@ ves_icall_System_Security_Principal_WindowsImpersonationContext_SetCurrentToken
...
 }
 
 gint32
-mono_security_win_get_token_name (gpointer token, gunichar2 ** uniname, MonoError *error)
+mono_security_win_get_token_name (gpointer token, gunichar2 ** uniname)
 {
+	ERROR_DECL (error); // FIXMEcoop
 	// FIXME This is now supported by UWP.
-	mono_security_win_not_supported ("GetTokenInformation", error);
+	mono_security_win_not_supported ("GetTokenInformation", &error);
```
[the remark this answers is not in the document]
lateralusX: Any example showing how this could adopt alternative pattern? This implementation just followed the implementation of corresponding method in mono-security.c and mono-security-windows.c that also uses the mono_error_set_pending_exception for this method implementation, so in order for this to change I believe we need to adjust all the implementations so they will take a MonoError *error as input parameter, similar to other icalls in those file, correct?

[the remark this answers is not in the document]
jaykrell: These icalls might not be in the right form yet. I’ll look later.

mono/metadata/mono-security-windows-uwp.c
jaykrell: The C compilers are very lax that way, not surprising. There might be a warning.

mono/metadata/w32file-win32.c
jaykrell: Also can just use GetFileSizeEx directly. I.e. drop Win9x/NT4 compat. GetFileSizeEx was introduced in Win2000.

mono/metadata/mono-security-windows-uwp.c
lateralusX: So the above fails to compile under C++ compiler (as expected), but under C compiler it works, but will give incorrect runtime result. There is a warning around constness, warning C4090: 'function': different 'const' qualifiers, but that warning is currently disabled in winconfig.h so it didn't show up. The other warning that normally should popup is warning C4047: 'operator' : 'identifier1' differs in levels of indirection from 'identifier2' but that is currently also disabled in winconfig.h so that's why this didn't show up during compile. I believe we should see if we could re-enable and fix some of these warnings (or disable on a granular level) into the since they are level 1 warnings. Thanks for spotting this.

lateralusX:
Good to go?

mono/metadata/icall.c
```diff
@@ -6752,7 +6752,7 @@ ves_icall_System_Environment_get_UserName (MonoError *error)
 {
 	error_init (error);
 	/* using glib is more portable */
-	return mono_string_new_handle (mono_domain_get (), g_get_user_name (), error);
+	return g_get_user_name () != NULL ? mono_string_new_handle (mono_domain_get (), g_get_user_name (), error) : MONO_HANDLE_CAST (MonoString, NULL_HANDLE);
```
jaykrell: Comparison to null is optional, and adds operator precedence ambiguity.

jaykrell: There is also a shorthand for MONO_HANLE_CAST (MonoString, NULL_HANDLE) available, but that is only cosmetic, ok either way.

mono/metadata/w32file-win32.c
```diff
@@ -515,6 +515,8 @@ mono_w32file_get_console_error (void)
 }
 #endif // HAVE_API_SUPPORT_WIN32_GET_STD_HANDLE
 
+#if HAVE_API_SUPPORT_WIN32_GET_FILE_SIZE_EX
```
jaykrell: Can just GetFileSizeEx unconditionally, but ok either way.
i.e. GetFileSizeEx introduced in Windows 2000, no need to use GetFileSize. But it is ok.

lateralusX: The previous one was just GetFileSize, so then it made sense, moving up to GetFileSizeEx drops the need to guard it as pointed out, but minor detail that I believe we can keep for now.

mono/metadata/w32subset.h
```diff
@@ -14,95 +14,167 @@
 #ifndef HAVE_API_SUPPORT_WIN32_GET_COMPUTER_NAME
 #if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT) || G_HAVE_API_SUPPORT(HAVE_UWP_WINAPI_SUPPORT)
 #define HAVE_API_SUPPORT_WIN32_GET_COMPUTER_NAME 1
+#else
+#define HAVE_API_SUPPORT_WIN32_GET_COMPUTER_NAME 0
```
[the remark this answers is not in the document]
lateralusX: Yes the intention was, of not previously defined to something, make sure we define it to either 1 or 0 depending on platform support making sure we always have it defined after including w32subset.h

mono/metadata/icall.c
lateralusX: OK.

lateralusX:
@luhenry, possible for you to approve this as well, jaykrell's approval didn't unlock the merge capabilities.

mono/metadata/assembly.c
```diff
@@ -1911,13 +1911,13 @@ static AssemblyPreLoadHook *assembly_preload_hook = NULL;
 static AssemblyPreLoadHook *assembly_refonly_preload_hook = NULL;
 
 static MonoAssembly *
-invoke_assembly_preload_hook (MonoAssemblyName *aname, gchar **assemblies_path)
```
luhenry: Curiousity: why the need of changing the name?

jaykrell: Probably it masks a global. I think msvc maybe goes overboard in what it thinks is worth warning about. Lexical scoping is a deliberate feature, usually accepted as common sense but sometimes criticized.

lateralusX: Several of the name changes fixes MSVC warnings where local variables hides global variables, using the same name.

mono/metadata/icall.c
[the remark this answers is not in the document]
lateralusX: Will fix.

mono/metadata/icall.c
```diff
@@ -7120,7 +7120,7 @@ ves_icall_System_IO_DriveInfo_GetDriveFormat (const gunichar2 *path, gint32 path
 MonoStringHandle
 ves_icall_System_Environment_InternalGetHome (MonoError *error)
 {
-	return mono_string_new_handle (mono_domain_get (), g_get_home_dir (), error);
+	return g_get_home_dir () != NULL ? mono_string_new_handle (mono_domain_get (), g_get_home_dir (), error) : MONO_HANDLE_CAST (MonoString, NULL_HANDLE);
```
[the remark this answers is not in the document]
lateralusX: Will fix.

mono/metadata/mono-security-windows-uwp.c
```diff
@@ -62,17 +62,20 @@ ves_icall_System_Security_Principal_WindowsImpersonationContext_SetCurrentToken
...
-ves_icall_System_Security_Principal_WindowsImpersonationContext_RevertToSelf (MonoError *error))
+ves_icall_System_Security_Principal_WindowsImpersonationContext_RevertToSelf (MonoError *error)
 {
 	mono_security_win_not_supported ("RevertToSelf", error);
 	return FALSE;
 }
 
 gint32
-mono_security_win_get_token_name (gpointer token, gunichar2 ** uniname, MonoError *error)
+mono_security_win_get_token_name (gpointer token, gunichar2 ** uniname)
```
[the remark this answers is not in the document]
lateralusX: OK, will fix this the error is part of the original icall signature as well. So for future insights, are all icalls called with an additional MonoError * as last parameter or something that needs to be defined somewhere in order to be called that way? For example, ves_icall_System_Security_Principal_WindowsIdentity_GetRoles, that one currently don't have a MonoError * as last parameter, what would need to be done in order for that method to also get a MonoErrorr * from the caller?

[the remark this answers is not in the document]
luhenry: You would need to look at `icall-def.h` and any icalls defined with `HANDLES` will take a `MonoError*` as last parameter.

PR #12321, opened 2019-01-08
[System.Net.Http] Fix CFNetworkHandler when no data is sent

spouliot:
The handler receive an `Open` and the a `Close` event but never a
`DataAvailable` event.

So it _hangs_ if only a `200` status is received (since the headers
are not considered _data_) received from the web server.

This is because some initialization only occurs inside
`HandleHasBytesAvailableEvent` which is now called, on `Close`,
to ensure the initialization is done before completing the request.

This, along with a cancellation check, makes the test case from
https://github.com/xamarin/xamarin-macios/issues/5190 work properly
with this handler.

Fixes https://github.com/xamarin/xamarin-macios/issues/5289

### DIFF
--- a/mcs/class/System.Net.Http/CFNetworkHandler.cs
+++ b/mcs/class/System.Net.Http/CFNetworkHandler.cs
@@ -302,17 +302,19 @@ namespace System.Net.Http
 		void HandleClosedEvent (object sender, CFStream.StreamEventArgs e)
 		{
 			var stream = (CFHTTPStream)sender;
+			// might not have been called (e.g. no data) but initialize critical data
+			HandleHasBytesAvailableEvent (sender, e);
 			CloseStream (stream);
 		}
 
 		void CloseStream (CFHTTPStream stream)
 		{
-			StreamBucket bucket;
-			if (streamBuckets.TryGetValue (stream.Handle, out bucket)) {
-				bucket.Close ();
-				streamBuckets.Remove (stream.Handle);
+			lock (streamBuckets) {
+				if (streamBuckets.TryGetValue (stream.Handle, out var bucket)) {
+					bucket.Close ();
+					streamBuckets.Remove (stream.Handle);
+				}
 			}
-
 			stream.Close ();
 		}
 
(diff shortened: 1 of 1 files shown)


spouliot:
c.c. @mandel-macaque

mandel-macaque:
For me, it looks good and the explanation makes perfect sense, if we got a header only response this can occur +1 although I can't give you a proper one in the mono repo :(

mcs/class/System.Net.Http/CFNetworkHandler.cs
```diff
@@ -302,17 +302,19 @@ void HandleErrorEvent (object sender, CFStream.StreamEventArgs e)
...
 
 		void CloseStream (CFHTTPStream stream)
 		{
-			StreamBucket bucket;
-			if (streamBuckets.TryGetValue (stream.Handle, out bucket)) {
-				bucket.Close ();
-				streamBuckets.Remove (stream.Handle);
+			lock (streamBuckets) {
+				if (streamBuckets.TryGetValue (stream.Handle, out var bucket)) {
+					bucket.Close ();
```
marek-safar: It might be better to do the Close outside of lock

kochev:
Good day! We encountered the same problem. In test conditions, the problem is not reproduced. In production, we see about 3-5 crashes per day.
 Stack trace:
`CFNetworkHandler.HandleHasBytesAvailableEvent (System.Object sender, CoreFoundation.CFStream+StreamEventArgs e)
CFStream.OnHasBytesAvailableEvent (CoreFoundation.CFStream+StreamEventArgs args)
CFStream.OnCallback (CoreFoundation.CFStreamEventType type)
CFStream.OnCallback (System.IntPtr s, System.nint type, System.IntPtr info)
(wrapper native-to-managed) CoreFoundation.CFStream.OnCallback(intptr,System.nint,intptr)
(wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate)
UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName)
Application.Main (System.String[] args)
(wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate)
UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName)
Application.Main (System.String[] args)`

Is there any WA?
When will the fix go to the master branch?

PR #12326, opened 2019-01-09
[wasm] Cleanup Framework Bindings and HttpMessageHandler code.

kjpou1:
Note: This modifies the WebAssembly.Bindings and WebAssembly.Net.Http assemblies to `netstandard`.

General cleanup.

- move to framework folder to match package structure.  Simplifies packager code finding the framework libraries which is always problematic.
- create solution and project files
- split up the `bindings.cs`, now removed, into multiple files and reformat code
- Clean up xml documentation
- generate xml file documentation on build.
- modify make file to build the solution and modify tests to use framework output.
- modify `packager` to use the correct directory structure.
- modify `sdk` build targets.
- modify `package` target to package the correct assemblies along with source projects and solution.

Tested:

- JavaScript shell tests which run in jenkens
- Browser Tests
- AOT build and tests


kjpou1:

Issue #12093, opened 2018-12-17
IronPython regression - stack overflow


migueldeicaza:
## Steps to Reproduce

1.  Install Mono (locally I have 5.16.0.221) on Mac.
2. Run `ipy`
3. This produces a stack trace, rather than getting the IronPython prompt

### Current Behavior

Stack traced produced is here:

https://gist.github.com/migueldeicaza/5465c7efcbc73927f29e5484d50d20ec

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.16.0.221 (2018-06/b63e5378e38 Mon Nov 19 18:08:09 EST 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-release_60/0b3cb8ac12c)
	GC:            sgen (concurrent by default)
```
## Stacktrace

Added offline, as it is large:

https://gist.github.com/migueldeicaza/5465c7efcbc73927f29e5484d50d20ec

EgorBo:
Crashes here https://github.com/IronLanguages/ironpython2/blob/0bdf2567889df8b24593add9dbe8e9da5f830c93/Src/IronPython/Compiler/GeneratorRewriter.cs#L595 when node is `System.Linq.Expressions.TypedDynamicExpression2`

lewurm:
@jaykrell could this be a tailcall issue?

marek-safar:
This is actually .NET Core regression (change in behaviour in CoreFX). I made a fix for the code to behave link .NET again for Mono only.

PR #12335, opened 2019-01-09
Bump CoreFX to bring a fix for #12093

marek-safar:


marek-safar:
@monojenkins build failed

PR #12340, opened 2019-01-09
[2018-10] [debugger][backport] Fix test ShouldCorrectlyStepOverOnExitFromArgsAfterStepInMethodParameter (#12114)

thaystg:
Backport of #12114

The problem was in this line:
ss_nested_with_two_args(ss_nested_arg (), ss_nested_arg ());
The first step_into works correctly entering in the ss_nested_arg method, but the second one doesn't work.
The problem was because this flag MONO_SEQ_POINT_FLAG_NONEMPTY_STACK was set, so the debugger don't consider that it was a possible break.
The solution of the issue was create another flag to identify nested calls and ignore the  MONO_SEQ_POINT_FLAG_NONEMPTY_STACK.
2 new tests were added to test other nested calls.

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
+++ b/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
@@ -466,6 +466,8 @@ public class Tests : TestsBase, ITest2
 		ss_fp_clobber ();
 		ss_no_frames ();
 		ss_await ();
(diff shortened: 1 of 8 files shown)


thaystg:
@monojenkins build failed

lewurm:
@monojenkins backport 2018-12

lewurm:
@monojenkins squash

PR #12342, opened 2019-01-09
Fix re-entrant structured exception handling

joncham:
Store handled status in local that is passed into
handler rather than a single thread local value
which can be overwritten in case of re-entrance.

I encountered this in following scenario:
```
try {
    // cause null reference exception
} finally {
   // cause a structure exception that mono does not handle, for example call OutputDebugString
}
```

The above caused a problem as:
1. Mono is in the middle of processing an exception it will handle, so previous thread local variable `mono_win_chained_exception_needs_run` is `FALSE`
2. Finally clauses are called during structured exception handling
3. Benign structured exception is triggered in finally clause (pinvoke `OutputDebugString` without debugger attached). This causes an exception mono doesn't want to handle but other handler will.
4. In previous logic, the thread local `mono_win_chained_exception_needs_run` is now set to `TRUE`
5. Return from other structured exception handler
6. Return from mono structured exception handling of null reference.
7. In previous logic, `mono_win_chained_exception_needs_run` is now `TRUE` and `EXCEPTION_CONTINUE_SEARCH` is returned indicating mono did *not* handle the null reference exception leading to a crash.

Question: Is it okay to reuse the `_dummy` signal number parameter on windows for this purpose? Or should we introduce another parameter?

Modified PR of https://github.com/Unity-Technologies/mono/pull/1119/files

### DIFF
--- a/mono/mini/exceptions-amd64.c
+++ b/mono/mini/exceptions-amd64.c
@@ -56,7 +56,7 @@ LPTOP_LEVEL_EXCEPTION_FILTER mono_old_win_toplevel_exception_filter;
 void *mono_win_vectored_exception_handle;
 
 #define W32_SEH_HANDLE_EX(_ex) \
-	if (_ex##_handler) _ex##_handler(0, ep, ctx)
+	if (_ex##_handler) _ex##_handler(er->ExceptionCode, &info, ctx)
 
 static LONG CALLBACK seh_unhandled_exception_filter(EXCEPTION_POINTERS* ep)
 {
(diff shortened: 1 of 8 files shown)


joncham:
Build failed as I hacked at cherry-pick from other branch. Before I fix, can I get input on reusing `_dummy` variable or adding another one (for Windows only?)?

lateralusX:
Maybe we it should be named better than _dummy since it has different meanings on different platforms? One downside reusing could be the different in and in/out behaviors of the parameter on different platforms.

Could an option be to do a have a different _info type on windows that includes this additional parameter instead of reusing _dummy? That parameter is already different on different platforms and could be allocated on stack and passed by pointer when calling mono signal handlers, it would include pointer to current EXCEPTION_POINTERS as well as a good name of this additional parameter and then putting a better name on _dummy indicating that it's actually a signo and we could use that on windows as the exception code, that way it will at least be more unified and the specific parameters can go into the platform specific _info parameter.

joncham:
I will make these changes.

joncham:
Thoughts @lateralusX ? Also, I don't think CI error is due to my changes.

lateralusX:
I believe the change looks ok, I retriggered the failing platform to see if it was some temporary failure. Would it make sense to add a test covering this scenario? I could add that for you later or you add it as part of this PR.

joncham:
I can add test. What did you have in mind? Should I pinvoke `RaiseException` in a finally block of a test?

joncham:
On our side we were calling `OutputDebugStringW` in a finally block *without* a debugger attached. This seems to cause a structured exception handled by a system wide handler/logger.

lateralusX:
OK, possible to add similar Windows only test to CI just to make sure this don't regress going forward?

PR #12345, opened 2019-01-10
[threading] Try to fix the os-event-unix.c:82 assert(event) failure

### LINKED ISSUE #11956: os-event-unix.c:82 assert(event) failure

lambdageek:
This is an attempt to fix #11956, but it contains a **gross hack** and I would like some **help**.

## The good bit ##

Partly revert a29ad08 (#9914)

The problem is a race between `mono_thread_detach_internal` and
`mono_thread_suspend_all_other_threads`:

In `mono_thread_suspend_all_other_threads`, we use `LOCK_THREAD (thread)` (which
locks `MonoInternalThread:synch_cs`) and then test and update the `thread->state` and also
call `mono_os_reset_event (thread->suspended)` while holding the lock.

In `mono_thread_detach_interna`l, we do not lock the thread, and then CAS the
`thread->state` and then continue and set `thread->suspended = NULL`.

As a result we have a concurrency bug: in `suspend_all_other_threads` it's
possible to see both the old (non-Stopped) value of `thread->state` and the
new (`NULL`) value of `thread->suspended`.  Hence the crashes seen in
#11956

This PR switches back to using `LOCK_THREAD` in `mono_thread_detach_internal` when
manipulating the thread state.

As a result, `mono_thread_suspend_all_other_thread`s will see a consistent
`thread->state` and `thread->suspended`: when it takes the lock, if the thread is
not stopped, the suspended event is not NULL, or else it will see the thread is
Stopped and we will skip trying to suspend it.

## The hacky bit ## 

Set the finalizer's `MonoInternalThread:synch_cs` to NULL when finished.

This is a hack to work around the way that synch_cs is managed.

We use a reference queue to ensure that the `MonoInternalThread:synch_cs` mutex
is destroyed when the thread is detached, but it does not set the `MonoInternalThread:synch_cs`
to NULL (since it just has a pointer to the mutex, not the address of the field).

The problem is that reference queues are force-cleared when a domain is
destroyed, and `MonoInternalThread` for the finalizer thread is associated with
the root domain.  The root domain is finalized before the finalizer thread is
detached.

Consequently, when `mono_thread_detach_internal` is called for the finalizer
thread, it will try to lock a destroyed mutex, which leads to a crash.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -943,19 +943,26 @@ mono_thread_detach_internal (MonoInternalThread *thread)
 	thread->current_appcontext = NULL;
 
 	/*
+	 * thread->synch_cs can be NULL if this was called after
+	 * ves_icall_System_Threading_InternalThread_Thread_free_internal.
+	 * This can happen only during shutdown.
+	 * The shutting_down flag is not always set, so we can't assert on it.
 	 * This should be alive until after the reference queue runs the
 	 * post-free cleanup function
+	 *
+	 * The lock here also helps us synchronize with
+	 * mono_threads_suspend_all_other_threads.  We don't want to delete
+	 * thread->suspended if suspend_all_other_threads is holding the thread
+	 * lock.
 	 */
-	while (TRUE) {
-		guint32 old_state = thread->state;
+	if (thread->synch_cs)
+		LOCK_THREAD (thread);
 
-		guint32 new_state = old_state;
-		new_state |= ThreadState_Stopped;
-		new_state &= ~ThreadState_Background;
+	thread->state |= ThreadState_Stopped;
+	thread->state &= ~ThreadState_Background;
 
-		if (mono_atomic_cas_i32 ((gint32 *)&thread->state, new_state, old_state) == old_state)
-			break;
-	}
+	if (thread->synch_cs)
+		UNLOCK_THREAD (thread);
 
 	/*
 	An interruption request has leaked to cleanup. Adjust the global counter.
--- a/mono/metadata/gc.c
+++ b/mono/metadata/gc.c
@@ -946,6 +946,15 @@ finalizer_thread (gpointer unused)
 	mono_coop_cond_signal (&exited_cond);
 	mono_finalizer_unlock ();
 
+	/* HACK - by the time we're finished, the root domain was already
+	 * destroyed, so the reference queue that cleans up
+	 * MonoInternalThread:synch_cs was already cleared since the
+	 * finalizer's MonoInternalThread is in the root domain.
+	 *
+	 * (MonoInternalThread:synch_cs is accessed by mono_thread_detach_internal)
+	 */
+	mono_thread_internal_current ()->synch_cs = NULL;
+
 	return 0;
 }
 
--- a/mono/metadata/object-internals.h
+++ b/mono/metadata/object-internals.h
@@ -510,7 +510,7 @@ struct _MonoInternalThread {
 	gpointer unused3;
 	gunichar2  *name;
 	guint32	    name_len;
-	guint32	    state;
+	guint32	    state;      /* must be accessed while synch_cs is locked */
 	MonoException *abort_exc;
 	int abort_state_handle;
 	guint64 tid;	/* This is accessed as a gsize in the code (so it can hold a 64bit pointer on systems that need it), but needs to reserve 64 bits of space on all machines as it corresponds to a field in managed code */


mono/metadata/threads.c
```diff
@@ -943,19 +943,26 @@ mono_thread_detach_internal (MonoInternalThread *thread)
...
-		new_state |= ThreadState_Stopped;
-		new_state &= ~ThreadState_Background;
+	thread->state |= ThreadState_Stopped;
+	thread->state &= ~ThreadState_Background;
 
-		if (mono_atomic_cas_i32 ((gint32 *)&thread->state, new_state, old_state) == old_state)
-			break;
-	}
+	if (thread->synch_cs)
+		UNLOCK_THREAD (thread);
```
jaykrell: I still like to try to guarantee the cs stays valid until the unlock. By refcounting it. Or at least storing it in local. Granted, might already be otherwise guaranteed. You know, is second if redundant? Can it be false but first true? Or vice versa? That’d be pretty bad, if possible.

jaykrell: Also, if no cs, then interlock? What does happen if the cs not there?

lambdageek: I believe with the refqueue freeing of synch_cs, there is only one situation where we get here with `thread->synch_cs == NULL`  --- when we're detaching the finalizer thread.  In which case nothing else is running, so it's probably alright to change the finalizer thread's state without locking.  But yea these lifetimes are not great.

lambdageek:
I'm going to try something...

@jaykrell wrote:
> I still like to try to guarantee the cs stays valid until the unlock. By refcounting it. 

My current hypothesis is: The relationship between a managed `System.InternalThread` object and an actual executing machine thread is that either one can outlive the other.

Therefore we basically need a refcount of upto 2 on data that needs to live as long as either the thread is running or its InternalThread is alive.  And evidently `synch_cs` is one such piece of data.  So my plan is to make a refcounted long-lived thread data struct that starts out with a refcount of 2 and both detaching the machine thread and finalizing the InternalThread object would decrement the refcount.  Whichever one happens later gets to cleanup.

The reason for the prolonged lifetime as best I understand it:

1. `InternalThread` normally outlives the machine thread.  This can happen because when one thread starts another it can hold a reference to the fresh thread's `Thread` object which holds a reference to the `InternalThread`.  So after the machine thread is done, the older thread can query the state of the younger `Thread` object.  This is the normal situation.
2. During shutdown we can have the opposite situation: the `InternalThread` objects are collected and their finalizers are invoked (I think this happens during root domain finalization), but the machine threads are still running, and they may still return to `start_wrapper_internal`.  So in this case we have an `InternalThread` whose finalizer ran first and detach will run second.

luhenry:
The native threads keeps the InternalThread alive by keeping a GCHandle to the InternalThread and freeing it whenever we detach the thread (ie the native thread dies).

lambdageek:
@luhenry you're right the machine thread keeps the InternalThread alive.

But there is still an issue: whether InternalThread's finalizer ran or not. As best I can tell, domain unloading forcibly finalizes objects in the domain even if they're reachable.

So we need to divide up the cleanup between detaching of the machine thread and finalization of the InternalThread object.

lambdageek:
Going to close this and open a new PR with a better fix: https://github.com/mono/mono/pull/12391

PR #12355, opened 2019-01-10
 [swf] Fix dispose of panels in StatusBar

ermshiperete:
This change implements disposing of the panels of a status bar.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/StatusBar.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/StatusBar.cs
@@ -220,15 +220,22 @@ namespace System.Windows.Forms {
 		}
 
 		protected override void Dispose (bool disposing) {
-			if (!IsDisposed)
+			if (!IsDisposed && disposing)
 			{
-				if (disposing)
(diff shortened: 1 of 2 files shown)


mcs/class/System.Windows.Forms/System.Windows.Forms/StatusBar.cs
```diff
@@ -220,15 +220,22 @@ protected override void CreateHandle ()
...
+				if (tooltip_timer != null)
+					tooltip_timer.Dispose();
 
-					if (tooltip_window != null)
-						tooltip_window.Dispose();
+				if (tooltip_window != null)
+					tooltip_window.Dispose();
+
+				if (panels != null) {
+					var copiedPanels = new StatusBarPanel [panels.Count];
```
marek-safar: Why is this copy needed?

ermshiperete: Good question. I copied that part from the referencesource implementation. I guess it's theoretically possible that calling `panel.Dispose()` modifies `panels`.

PR #12361, opened 2019-01-10
[Mono.Native] Build on Android (for 2018-10 branch)

alexischr:


alexischr:
@monojenkins build failed

PR #12365, opened 2019-01-11
[configure] Small fix in sigaltstack() checking on cross-compiler case

alexischr:
Currently we print out that we are "assuming yes" for the condition but we don't actually follow the "yes" action here: https://github.com/mono/mono/blob/1291c08c1d07b97ed7087d15f4eca2538bd5c24f/configure.ac#L2678-L2679

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2577,7 +2577,8 @@ if test x$host_win32 = xno; then
 	if test "x$with_sigaltstack" != "xyes"; then
 		AC_MSG_RESULT(disabled)
 	elif test "x$cross_compiling" = "xyes"; then
-		AC_MSG_RESULT(cross compiling, assuming yes)
+		AC_MSG_RESULT(cross compiling, assuming no)
+		with_sigaltstack=no
 	else
 		AC_TRY_RUN([
 			#include <stdio.h>


marek-safar:
@alexischr please add description to this change

luhenry:
@monojenkins build failed

alexischr:
@lewurm that's fair especially since it doesn't build; then the change will be `assuming no` instead

lewurm:
@monojenkins build failed

configure.ac
```diff
@@ -2528,7 +2528,8 @@ if test x$host_win32 = xno; then
 	if test "x$with_tls" != "x__thread"; then
 		AC_MSG_RESULT(disabled)
 	elif test "x$cross_compiling" = "xyes"; then
-		AC_MSG_RESULT(cross compiling, assuming yes)
+		AC_MSG_RESULT(cross compiling, assuming no)
+		with_sigaltstack=no
```
luhenry: You are doing `with_sigaltstack` while checking for `__thread`. That looks very wrong.

alexanderkyte: Could you add some comments around here? I don't understand how the condition relates to the body

alexischr: Sorry, had put that in the wrong block

lewurm:
@monojenkins build

vargaz:
@monojenkins build failed

PR #12367, opened 2019-01-11
[wasm] Compile the wasm runtime using optimization.

vargaz:


vargaz:
@monojenkins build failed

lewurm:
https://github.com/mono/mono/blob/2f794e00b7e7c264488f624b788090e7b3e899bd/sdks/builds/wasm.mk#L62

I assume that `CFLAGs=-Os -g` gets overridden by the latter?  If so, we should remove that one here.

vargaz:
@monojenkins build failed

vargaz:
Will fix that.

PR #12372, opened 2019-01-11
Use LoadFrom instead LoadFile for load logger type.

mfilippov:
LoadFile lost assembly load-context. If the logger assembly has any dependencies placed in same folder logger can't load it. Main.cs using LoadFrom to load assembly but assembly already loaded in the cache after GetLoggerTypeName.
It is a regression in previous version mono(I'm not sure about certain version) didn't use cache in this case. We have some customer reports from Linux distro without msbuild(like Arch). They have a problem after  mono upgrade.


marek-safar:
approve

mfilippov:
@marek-safar Could you backport it into next stable mono version? We have a broken test with xbuild on mono 5.18.

EgorBo:

marek-safar:

marek-safar:

PR #12380, opened 2019-01-11
[wasm] Don't make aot-dummy.dll an output of the linker.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #12381, opened 2019-01-11
[runtime] Register the GC alloc icalls during startup instead of on-demand, since create_allocator () might not be called at runtime in llvmonly mode if the generated code directly calls its wrapper.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #12382, opened 2019-01-11
Fix mono conc hashtable lookup endless loop

hecklatcipsoft:
The tombstone scrubbing fix applied in the past did not work in every case.

Since the key assignment to an element added via mono_conc_hashmap_insert() happens before the tombstone check, the tombstone count remains the same even if a tombstone was replaced. 

In this case the table size check can decide to rehash to the same table size because assuming free spots will become available from not copying over the tombstones. But if all tombstones have been replaced with new elements the fill rate of the hashmap will effectively stay the same, and in the long run rehashing with the same hashmapsize finally leads to a completely filled hashmap.

In the case of a completely filled hashmap looking up an element, not contained in the hashmap, will result in an endless loop because the termination condition for lookups is finding a null key.

We continuously run into this issue with our server application.

Please include this fix into the main repository

### DIFF
--- a/mono/utils/mono-conc-hashtable.c
+++ b/mono/utils/mono-conc-hashtable.c
@@ -129,6 +129,8 @@ rehash_table (MonoConcurrentHashTable *hash_table, int multiplier)
 	mono_memory_barrier ();
 	hash_table->table = new_table;
 	hash_table->overflow_count = (int)(new_table->table_size * LOAD_FACTOR);
+	hash_table->element_count -= hash_table->tombstone_count;
+	hash_table->tombstone_count = 0;
 	conc_table_lf_free (old_table);
 }
 
@@ -141,7 +143,7 @@ check_table_size (MonoConcurrentHashTable *hash_table)
 			rehash_table (hash_table, 1);
 		else
 			rehash_table (hash_table, 2);
-  }
+	}
 }
 
 
@@ -351,14 +353,11 @@ mono_conc_hashtable_insert (MonoConcurrentHashTable *hash_table, gpointer key, g
 				kvs [i].value = value;
 				/* The write to values must happen after the write to keys */
 				mono_memory_barrier ();
-				
 				if (kvs [i].key == TOMBSTONE)
 					--hash_table->tombstone_count;
 				else
 					++hash_table->element_count;	
-				
 				kvs [i].key = key;
-				
 				return NULL;
 			}
 			if (key == kvs [i].key) {
@@ -374,14 +373,11 @@ mono_conc_hashtable_insert (MonoConcurrentHashTable *hash_table, gpointer key, g
 				kvs [i].value = value;
 				/* The write to values must happen after the write to keys */
 				mono_memory_barrier ();
-				
 				if (kvs [i].key == TOMBSTONE)
 					--hash_table->tombstone_count;
 				else
 					++hash_table->element_count;
-				
 				kvs [i].key = key;
-				
 				return NULL;
 			}
 			if (equal (key, kvs [i].key)) {
--- a/mono/utils/mono-conc-hashtable.c
+++ b/mono/utils/mono-conc-hashtable.c
@@ -141,7 +141,15 @@ check_table_size (MonoConcurrentHashTable *hash_table)
 			rehash_table (hash_table, 1);
 		else
 			rehash_table (hash_table, 2);
-	}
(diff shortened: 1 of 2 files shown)


lewurm:
@hecklatcipsoft Thanks for your contribution! In general the workflow is like this: open PR against `master` branch and once it's accepted we can backport it to the relevant release branches. Also, you could you please sign the CLA in the meanwhile?

@BrzVlad can you review this PR please?

hecklatcipsoft:
@lewurm : Sorry for the inconvenience. This is my first pull request.

Regarding the CLA: I investigated and fixed this issue during my working hours, and I'm still waiting for a signed statement from my employer. I will sign the CLA as soon as I hold the statement in my hands.

mono/utils/mono-conc-hashtable.c
```diff
@@ -351,12 +351,14 @@ mono_conc_hashtable_insert (MonoConcurrentHashTable *hash_table, gpointer key, g
...
 				mono_memory_barrier ();
-				kvs [i].key = key;
+				
 				if (kvs [i].key == TOMBSTONE)
 					--hash_table->tombstone_count;
 				else
 					++hash_table->element_count;	
-					
+				
+				kvs [i].key = key;
```
BrzVlad: @hecklatcipsoft From your description I see that the issue you are encountering comes from not resetting the tombstone / element_counts after the rehash. Isn't that fix enough ? In what case is this commit needed ?

hecklatcipsoft: HI @BrzVlad ,

Well it's just plain wrong to assign to the key before checking key == tombstone, because it never will be after assignment.

But yes both commits are fixes for scenarios where lookups of elements that are not contained in the hashmap can result in an endless loop. 

simplified example (assume hashmap size: 4):
insert #1  => insert #2  => insert #3 ( tombstone_count= 0, element_count =3)
remove #1  => remove #2 ( tombstone_count= 2, element_count =3 )
insert #1  => insert #2 ( rehash() will rehash to size 4, tombstone_count= 2, element_count =5 )
remove #1  => remove #2 ( tombstone_count= 4, element_count =5 )
insert #1  => insert #2 ( rehash() will rehash to size 4, tombstone_count= 4, element_count =7 )
insert #4 ( rehash() will rehash to size 4, tombstone_count= 4, element_count =8 )
lookup #5 ( endless loop )

The second fix will effectively fix some of the errors introduced by the key assignment bug (including the simple example above). But I'm not sure if a scenario exists where the erroneous tombstone count will introduce other problems. So to be on the save side I recommend merging the key assignment fix as well.

Best, Jochen

BrzVlad: Wow, sorry. I didn't even notice that it was checking for the previously assigned value. Can't believe anything worked

mono/utils/mono-conc-hashtable.c
```diff
@@ -141,7 +141,15 @@ check_table_size (MonoConcurrentHashTable *hash_table)
...
 			rehash_table (hash_table, 2);
-	}
+	
+
+    // now we have to reset the tombstone and element counters to the actual values
+    // because if we don't we will be doomed to finally fill up the entire hashmap 
+    // and the next failing lookup will toast our cpu in an endless cycle
+
+    hash_table->element_count -= hash_table->tombstone_count;
+    hash_table->tombstone_count = 0;
```
BrzVlad: @hecklatcipsoft These logically belong in rehash_table. There is also some whitespace damage here.

We need to set these also for the concurrent hashtable in https://github.com/mono/mono/blob/master/mono/metadata/mono-conc-hash.c#L167 . Feel free to add it here, or I can make a separate PR for it.

hecklatcipsoft:
@BrzVlad Ups, I totally missed that this was duplicated code. Thanks for the refix.

lewurm:
@monojenkins build

hecklatcipsoft:
@lewurm
I do have a written permission now, and signed the agreement.

lewurm:
@monojenkins backport master
@monojenkins backport 2019-02
@monojenkins backport 2018-12
@monojenkins backport 2018-10

lewurm:
@hecklatcipsoft great, thank you!

Unfortunately we don't accept changes for this release branch anymore, but I'll try to get them into the open release branches (and `master` as well).

PR #12387, opened 2019-01-11
[runtime] Fix jit-info.c memory consistency

alexanderkyte:
In the process of getting the tests added in https://github.com/mono/mono/pull/12125 green, I ran into a lot of data races and use-after-frees in jit-info.c . 

This PR has extracted them, so they can be considered separately.

PR #12125, opened 2018-12-19
[crash] Test and fix stability of native crash pipeline(s)

alexanderkyte:
The added test caught a lot of failures. It failed most of the time it was run before (90% of executions resulting in crashes in seq-point or jit info tables or in gdb dumper).

After these changes, I was able to run the suite 10k times on OSX and 10k times on Linux without seeing a failure. 

This brings some crash format changes too:

https://gist.github.com/alexanderkyte/72c12f0450513f079189e3e6561db502

I think this is a good idea. If we've got so many crash-time state reporters, having visual demarcation between them makes it easier to say "we crashed in the middle of doing X".

Please don't squash these changes, as they're rather varied, and would not offer a single semantic unit to be rebased around and reverted.

PR #12388, opened 2019-01-11
[runtime] Fix jemalloc support

alexanderkyte:
In the process of diagnosing 

https://github.com/mono/mono/pull/12125
and
https://github.com/mono/mono/pull/12387

I needed jemalloc to be fixed because instruments was too heavyweight and I needed the memory assertions (--enable-fill).

This fixes the build with --enable-jemalloc.


lewurm:
@monojenkins build failed

PR #12395, opened 2019-01-12
Align mcs's tuple deconstruction behavior with csc

### LINKED ISSUE #11334: mcs's behavior in tuple deconstruction differs from csc's
## Steps to Reproduce
```cs
using System;
 
public class Test
{
	public static void Main()
	{
		int a = 1;
		int b = 2;

		(a, b) = (b, a);
        
		Console.WriteLine($"a:{a} b:{b}");
	}
}
```

### Current Behavior
`a:2 b:2`

### Expected Behavior
`a:2 b:1`

- cf. https://github.com/dotnet/roslyn/blob/master/docs/features/deconstruction.md

[ ] macOS
[ ] Linux
[x] Windows

**Version Used**:
5.21.0 and 5.16.0

BijanT:
Before the code
(a, b) = (b, a);
was the equivalent of
a = b;
b = a;

now, it evaluates it to so something like
tmp1 = a;
tmp2 = b;
a = tmp2;
b = tmp1;

This change only applies to tuple literals because I don't think other tuples are able to overwrite themselves.

Fixes #11334

### DIFF
--- a/mcs/mcs/tuples.cs
+++ b/mcs/mcs/tuples.cs
@@ -447,6 +447,7 @@ namespace Mono.CSharp
 	{
 		Expression source;
 		List<Expression> targetExprs;
+		List<Expression> tempExprs;
 		List<BlockVariable> variables;
 		Expression instance;
 
@@ -455,6 +456,8 @@ namespace Mono.CSharp
 			this.source = source;
 			this.targetExprs = targetExprs;
 			this.loc = loc;
+
+			tempExprs = new List<Expression> ();
 		}
 
 		public TupleDeconstruct (List<BlockVariable> variables, Expression source, Location loc)
@@ -462,6 +465,8 @@ namespace Mono.CSharp
 			this.source = source;
 			this.variables = variables;
 			this.loc = loc;
+
+			tempExprs = new List<Expression> ();
 		}
 
 		public override Expression CreateExpressionTree (ResolveContext ec)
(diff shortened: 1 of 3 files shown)


marek-safar:
approve

marek-safar:
Please add relevant test to mcs/tests/test-tuple-??.cs

mcs/mcs/tuples.cs
```diff
@@ -538,10 +539,18 @@ protected override Expression DoResolve (ResolveContext rc)
...
+						var temp = new LocalTemporary (element_src.Type);
+						tempExprs.Add (new SimpleAssign (temp, element_src).Resolve (rc));
+						targetExprs [i] = new SimpleAssign (targetExprs[i], temp).Resolve (rc);
+					} else {
+						targetExprs [i] = new SimpleAssign (targetExprs [i], element_src).Resolve (rc);
+					}
 				}
 
 				eclass = ExprClass.Value;
+				targetExprs.InsertRange (0, tempExprs);
```
marek-safar: It should probably be better to have this as a separate field and not to add to targetExprs at all

mcs/mcs/tuples.cs
```diff
@@ -590,8 +599,17 @@ public override void EmitStatement (EmitContext ec)
 			if (instance != null)
 				((ExpressionStatement) source).EmitStatement (ec);
 			
-			foreach (ExpressionStatement expr in targetExprs)
+			foreach (ExpressionStatement expr in targetExprs) {
+				var temp = (expr as Assign)?.Target as LocalTemporary;
+				if(temp != null) {
```
marek-safar: nit: please following existing coding style

BijanT:
I have moved the tmp = source assignments to a separate field of the TupleDeconstruct class as suggested.
I have also fixed a failing regression test by trying to make the temporary variables type equal to the target's type, and if that fails, make the type the same as the source's type. After running "make run-test" on my machine, I saw that the only failing tests were due to increases in code size. I'm assuming this is fine because I don't know how this bug would be fixed without changing the output code.
Finally, I added a simple regression test.

steveisok:
@monojenkins build failed

lambdageek:

PR #12396, opened 2019-01-12
[crash] Remove missing LLVM_API_SUPPORT symbol usage

alexanderkyte:
According to @akoeplinger we are seeing a failure on CI:

```
mono-state.c:631:48: error: use of undeclared identifier 'LLVM_API_VERSION'
20:28:52         mono_state_writer_printf (writer, "\"%d\"\n", LLVM_API_VERSION);
```

I've removed that here. 

Note that this is the last item in the JSON list, so we have to preserve the lack of a trailing comma. That's why Github renders this diff like so.


alexanderkyte:
@monojenkins build pkg

PR #12405, opened 2019-01-14
[corlib] Add tracking exception for not object based type hierarchies

marek-safar:


marek-safar:
@monojenkins build failed

PR #12418, opened 2019-01-14
[wasm] Build target fixes

kjpou1:
- Add linker xml descriptor to WebAssembly.Bindings and WebAssembly.Net.Http
   - The XML linker descriptor will keep the full assembly for these assemblies during link.

- Fix Link Assemblies command line.
   - Miscellaneous fixes for how the assemblies are linked whether they are framework or facades.
   - Add the Facades directory to the resolve paths.


mhutch:
Doesn't this mean that `WebAssembly.Bindings` will _always_ be preserved, even with "link all"?

vargaz:
@monojenkins build failed

PR #12197, opened 2018-12-24
Fixing FileSystemWatcher problem in VSMac

MaximLipnin:
Fixes https://github.com/mono/mono/issues/11635.

Main changes are in https://github.com/mono/corefx/pull/225.

Some initial work was done in https://github.com/mono/mono/pull/11945

The idea is to share CFRunLoop for FSEventStreams.

PR #12431, opened 2019-01-15
[2018-12] Backport of changes related to FileSystemWatcher problem in VSMac

### LINKED ISSUE #11635: FileSystemWatcher problem in VSMac

MaximLipnin:
Backport of https://github.com/mono/mono/pull/12197 to 2018-12 branch.

This PR bumps corefx submodule to get changes from https://github.com/mono/corefx/pull/237.

Fixes #11635.


marek-safar:
@monojenkins commit apidiff

PR #6898, opened 2018-02-09
[System]: Set NetworkStream.ReadTimeout in WebConnection.InitConnection().

baulig:
Setting `ReadTimeout` in `WebResponseStream.ctor` throws `System.Net.Sockets.SocketException : Invalid arguments` on iOS.

Set it in `WebConnection.InitConnection()` immediately after establishing a connection, before we read/write anything.

PR #12432, opened 2019-01-15
Fixes VS4Mac Issue #6898

mkrueger:
https://github.com/mono/monodevelop/issues/6898

Not sure if it's the root of the problem but it surely fixes the dead lock on my machine.


marek-safar:
@alexischr why is this implemented still used on Mike's machine?

alexischr:
@mkrueger what version of Mono are you using where you saw this? Newer versions of Mono should be always using a different FileSystemWatcher on Linux (if not that's a bug)

mkrueger:
Was a pretty recent master build. See the original bug report for a trace.

PR #12452, opened 2019-01-16
[System.IO.Compression]: Correctly handle exceptions in inner stream.  #12205.

### LINKED ISSUE #12205: [bcl] DeflateNativeStream may leak SafeDeflateStreamHandle if an exception in the underlying stream occurs.
## Steps to Reproduce

1. Create a DeflateStream (e.g. via GzipStream or using the normal HttpWebRequest classes).
2. Cause the underlying stream to throw an error on write / read (e.g. when the other end of the NetworkingStream closes the connection)

Reproduction case: https://gist.github.com/Taloth/5a9e0e2408717ab464810984b645e250

Discovered by: @Taloth

### Current Behavior

The native zlib code calls `UnmanagedRead` / `UnmanagedWrite` which in turn calls `base_stream.Read` / `base_stream.Write`. If `base_stream` throws an exception during those calls, the exception will be propagated to unmanaged code. Since the unmanaged code does not understand exceptions, this causes a few issues. The most noticeable being, that `SafeDeflateStreamHandle` will not be disposed properly. This is because, the call to native zlib increases the ref count by one. If there is an exception in the managed code called by the native code, the ref count will not be decreased again. This leads to a ref count of 2 or more for the `SafeDeflateStreamHandle`. When disposing a `SafeHandle` its ref count must be 1. Hence, the `ReleaseHandle` method of `SafeDeflateStreamHandle` will never be called and hence the native resources will not be correctly freed. There could also be other possible side effects.

### Expected Behavior

`UnmanagedRead` / `UnmanagedWrite` catch exceptions and do not let them propagate through to native code. However, maybe `WriteZStream` / `ReadZStream` also need to catch exceptions. In the rare case of an exception that cannot be caught by `UnmanagedRead` / `UnmanagedWrite` (e.g. a ThreadAbortException), the `SafeDeflateStreamHandle` should still be properly disposed.

**Edit:** One could also rewrite the whole DeflateStream to only use the native zlib calls and not use any native "glue" like .NET does it. However, that would require significantly more work.

[x] macOS
[x] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.23.0 (fix/inner-stream-leaking/d5988ae8874 Tue 25 Dec 2018 02:44:51 CET)
        TLS:
        SIGSEGV:       altstack
        Notification:  kqueue
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        Suspend:       hybrid
```

I can make a PR if that's ok. Should be relatively easy to fix.

baulig:
DeflateStream now correctly handles exceptions thrown by it's inner stream and
rethrows them after returning from native code.

Fixes #12205.

### DIFF
--- a/mcs/class/System/System.IO.Compression/DeflateStream.cs
+++ b/mcs/class/System/System.IO.Compression/DeflateStream.cs
@@ -456,7 +456,6 @@ namespace System.IO.Compression
 			return self.UnmanagedWrite (buffer, length);
 		}
 
-		static int martinTest;
 
 		int UnmanagedWrite (IntPtr buffer, int length)
 		{
--- a/mcs/class/System/System.IO.Compression/DeflateStream.cs
+++ b/mcs/class/System/System.IO.Compression/DeflateStream.cs
@@ -349,6 +349,7 @@ namespace System.IO.Compression
 		GCHandle data;
 		bool disposed;
 		byte [] io_buffer;
+		Exception last_error;
 
 		private DeflateStreamNative ()
 		{
@@ -430,7 +431,15 @@ namespace System.IO.Compression
 				io_buffer = new byte [BufferSize];
 
 			int count = Math.Min (length, io_buffer.Length);
-			int n = base_stream.Read (io_buffer, 0, count);
+			int n;
+
+			try {
+				n = base_stream.Read (io_buffer, 0, count);
+			} catch (Exception ex) {
+				last_error = ex;
+				return -12;
+			}
+
 			if (n > 0)
 				Marshal.Copy (io_buffer, 0, buffer, n);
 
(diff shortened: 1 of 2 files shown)


mcs/class/System/System.IO.Compression/DeflateStream.cs
```diff
@@ -447,6 +456,8 @@ static int UnmanagedWrite (IntPtr buffer, int length, IntPtr data)
 			return self.UnmanagedWrite (buffer, length);
 		}
 
+		static int martinTest;
```
Therzok:
```suggestion
```

baulig: Oops, thanks a lot for spotting this!

PR #12454, opened 2019-01-16
[Android,AOT] Support dumping AOT offsets with Android NDK r17+

grendello:
The latest Android NDK releases no longer ship with per-platform C headers.
Instead they ship with a set of unified headers and require the presence of the
`__ANDROID_API__` macro in order to select the target platform.
The AOT offset dumper doesn't support this new layout and this commit changes
that by adding support for the unified headers (without removing support for the
older NDKs) thus fixing the following error when trying to dump the offsets with
newer NDK:

    Processing triple: aarch64-v8a-linux-android
    Error parsing '[...]/mono/mono/metadata/metadata-cross-helpers.c'
    [...]/mono/mono/metadata/metadata-cross-helpers.c(5,10): fatal: 'stdio.h' file not found
    Error parsing '[...]/mono/mono/mini/mini-cross-helpers.c'
    [...]/mono/mono/mini/mini-cross-helpers.c(7,10): fatal: 'stdio.h' file not found

### DIFF
--- a/tools/offsets-tool/MonoAotOffsetsDumper.cs
+++ b/tools/offsets-tool/MonoAotOffsetsDumper.cs
@@ -632,9 +632,46 @@ namespace CppSharp
 
             const int androidNdkApiLevel = 21;
 
+            string arch = GetArchFromTriple(target.Triple);
             var toolchainPath = Path.Combine(AndroidNdkPath, "platforms",
-                "android-" + androidNdkApiLevel, "arch-" + GetArchFromTriple(target.Triple),
+                "android-" + androidNdkApiLevel, "arch-" + arch,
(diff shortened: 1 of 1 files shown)


lewurm:
@akoeplinger @alexischr what's wrong with the `Linux ARMv7 Interepreter` status here? The job is done, but the status hasn't been reported (?)

akoeplinger:
According to the log:

```
org.kohsuke.github.HttpException: Server returned HTTP response code: 502, message: 'Bad Gateway' for URL: https://api.github.com/repos/mono/mono/statuses/ea2f4568d2207ad01c5cc1c74daf7a2f618a57b5
```
It's an issue on Github's end, merging.

PR #12459, opened 2019-01-16
[System]: Correctly dispose inner stream in `WebResponseStream`.  #11928.

### LINKED ISSUE #11928: InnerStream is not correctly Disposed for WebResponseStream, leading to potential leaking of DeflateStreamNative (and maybe others).
## Background
I maintain a popular app that uses mono to run under OSX and Linux: https://github.com/Radarr/Radarr.
When profiling some memory issues, we noticed, that some DeflateStreamNative instances are kept around forever. It seems as though they are leaked by the WebResponseStream. See below for more details.

## Steps to Reproduce
I haven't yet tried to create a minimal working example, however, these are the steps used in our app which leads to the leak:
```
            HttpWebResponse httpWebResponse = null;
            HttpWebRequest webRequest = null;
            try
            {
                webRequest = (HttpWebRequest) WebRequest.Create((Uri) url);

                // Deflate is not a standard and could break depending on implementation.
                // we should just stick with the more compatible Gzip
                //http://stackoverflow.com/questions/8490718/how-to-decompress-stream-deflated-with-java-util-zip-deflater-in-net
                webRequest.AutomaticDecompression = DecompressionMethods.GZip;

                webRequest.Method = "GET";
                webRequest.AllowAutoRedirect = false;

                try
                {
                    httpWebResponse = (HttpWebResponse) webRequest.GetResponse();
                }
                catch (WebException e)
                {
                    if (e.Status == WebExceptionStatus.SecureChannelFailure && OsInfo.IsWindows)
                    {
                        SecurityProtocolPolicy.DisableTls12();
                    }

                    httpWebResponse = (HttpWebResponse) e.Response;

                    if (httpWebResponse == null)
                    {
                        throw;
                    }
                }

                byte[] data = null;

                using (var responseStream = httpWebResponse.GetResponseStream())
                {
                    if (responseStream != null)
                    {
                        data = responseStream.ToBytes();
                    }
                }

                return data;
            }
            finally
            {
                webRequest = null;
                (httpWebResponse as IDisposable)?.Dispose();
                httpWebResponse = null;
            }
```

### Current Behavior

WebResponseStream only disposes of its innerStream here: https://github.com/mono/mono/blob/cf7b6cfd5f3bc653a4a139c299fb2d6006d49fa9/mcs/class/System/System.Net/WebResponseStream.cs#L181-L184

This lambda will only be executed, if the request is aborted, hence this will not happen with a successful request.

Normally, this wouldn't be a big deal, since the Garbage Collector will still clean up the stream afterwards. However, if the content is GZipped (as is often the case), innerStream will be a ContentDecodeStream which wraps a DeflateStream. DeflateStream creates a DeflateStreamNative which is not collected by the garbage collector due to the following code: https://github.com/mono/mono/blob/5187b0bdaaff56936d697b0753dad51e7498991d/mcs/class/System/System.IO.Compression/DeflateStream.cs#L360
This leads to DeflateStreamNative being kept alive forever. I haven't looked how many other streams are leaked due to this, but the stream wrapped by DeflateStreamNative is also kept alive.

### Expected Behavior

The innerStream is properly closed and no leakage occurs.

I have added innerStream?.Dispose() inside the Close_Internal method of WebResponseStream here: https://github.com/mono/mono/blob/cf7b6cfd5f3bc653a4a139c299fb2d6006d49fa9/mcs/class/System/System.Net/WebResponseStream.cs#L397

This seems to work fine, no error has occured so far and all DeflateStreamNative instances are correctly cleaned up.

Here are two profiling runs, both with heapshots at every 30seconds, one is with the fix one without:

https://galli.me/output_with_fix.mlpd

https://galli.me/output_no_fix.mlpd

Below is DeflateStreamNative and its paths to root (this instance for example was created at 20 seconds and was still alive by the time of the fourth snapshot at 2 minutes).

screen shot 2018-12-04 at 14 39 53

[x] macOS
[x] Linux
[ ] Windows

**Version Used**:
Latest master (freshly cloned today):

```
Mono JIT compiler version 5.21.0 (master/c99d2de68ef Tue  4 Dec 2018 13:39:02 CET)
        TLS:
        SIGSEGV:       altstack
        Notification:  kqueue
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        Suspend:       hybrid
```
## Stacktrace

Linked issue with some more discussion (also surrounding other possible leaks): https://github.com/Radarr/Radarr/issues/1580

baulig:
Fixes #11928.

### DIFF
--- a/mcs/class/System/System.Net/WebResponseStream.cs
+++ b/mcs/class/System/System.Net/WebResponseStream.cs
@@ -400,6 +400,8 @@ namespace System.Net
 				WebConnection.Debug ($"{ME} CLOSE #1: read_eof={read_eof} bufferedEntireContent={bufferedEntireContent}");
 				if (read_eof || bufferedEntireContent) {
 					disposed = true;
+					innerStream?.Dispose ();
+					innerStream = null;
 					Operation.Finish (true);
 				} else {
 					// If we have not read all the contents


marek-safar:
@monojenkins backport 2018-12

marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins build failed

PR #12460, opened 2019-01-17
[do not merge] Loader hacks for netcore

lambdageek:
The main thing is to add a preload hook to the loader that uses the 
 list of assemblies from `TRUSTED_PLATFORM_ASSEMBLIES` from `coreclr_initialize` to drive assembly loading.

Also had to stub out a bunch of classes that Mono really wants that I couldn't see how to disable, like `System.AppDomain` and `System.AppDomainSetup` and `System.Runtime.Remoting.Contexts.Context`

I also had to make a stub non-`abstract` `System.RuntimeType` and add the native fields to `System.Type` and `System.RuntimeType` so they match the layout of `MonoReflectionType` and `MonoReflectionMonoType`, respectively, in native.

----

There is a grab bag of stuff here, so best to look one commit at a time or else it won't make sense.

----

I run the sample with 
```
cd netcore/sample
make -C ../..           # this will fail when trying to AOT some assemblies --- I broke AOT
make -C ../../mcs/class/Sysem.Private.CoreLib            # this needs a working donet install some place
make prepare            # after the first run make link-mono is sufficient
MONO_LOG_LEVEL=debug MONO_LOG_MASK=asm lldb ./dotnet HelloWorld/bin/Debug/netcoreapp3.0/HelloWorld.dll
```

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -4106,7 +4106,6 @@ mono_assembly_load_from_gac (MonoAssemblyName *aname,  gchar *filename, MonoImag
 MonoAssembly*
 mono_assembly_load_corlib (const MonoRuntimeInfo *runtime, MonoImageOpenStatus *status)
 {
-	char *corlib_file;
 	MonoAssemblyName *aname;
 	MonoAssemblyOpenRequest req;
 	mono_assembly_request_prepare (&req.request, sizeof (req), MONO_ASMCTX_DEFAULT);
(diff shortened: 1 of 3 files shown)


lambdageek:
bits and pieces here are getting cherrypicked in separate PRs.

* #12477 
* #12503 
* #12504 

The main remaining piece is the parsing of coreclr_initialize trusted framework assemblies property and setting up a Mono assembly loader preload hook, as well as the initial loading of System.Private.CoreLib using the preload hook.

lambdageek:
The last unmerged piece of this PR is in #12519.

Closing.

PR #12464, opened 2019-01-17
[wasm] Fix crash with delegates.

kjpou1:
Internally the javascript bridge support uses `System.Delegate` and `System.Delegate.DynamicInvoke` to support delegate callbacks.  The linker, when executed over the assemblies, does not know about this code coverage and will strip out `System.Delegate.DynamicInvoke`.

To fix this:

- Add reference to `System.Delegate.DynamicInvoke` to the `WebAssembly.Bindings` linker xml file so the linker knows not to strip the code.
- Add detailed error messages in case something else goes wrong so the crash is not so cryptic.


kjpou1:
@monojenkins build failed

PR #12466, opened 2019-01-17
[2018-10 Android,AOT] Support dumping AOT offsets with Android NDK r17+ (#12454)

grendello:
Backport of: https://github.com/mono/mono/commit/eebe4b52857140ba9dea872c6249960273d222f2 from https://github.com/mono/mono/pull/12454

The latest Android NDK releases no longer ship with per-platform C headers.
Instead they ship with a set of unified headers and require the presence of the
`__ANDROID_API__` macro in order to select the target platform.
The AOT offset dumper doesn't support this new layout and this commit changes
that by adding support for the unified headers (without removing support for the
older NDKs) thus fixing the following error when trying to dump the offsets with
newer NDK:

    Processing triple: aarch64-v8a-linux-android
    Error parsing '[...]/mono/mono/metadata/metadata-cross-helpers.c'
    [...]/mono/mono/metadata/metadata-cross-helpers.c(5,10): fatal: 'stdio.h' file not found
    Error parsing '[...]/mono/mono/mini/mini-cross-helpers.c'
    [...]/mono/mono/mini/mini-cross-helpers.c(7,10): fatal: 'stdio.h' file not found

### DIFF
--- a/tools/offsets-tool/MonoAotOffsetsDumper.cs
+++ b/tools/offsets-tool/MonoAotOffsetsDumper.cs
@@ -632,9 +632,46 @@ namespace CppSharp
 
             const int androidNdkApiLevel = 21;
 
+            string arch = GetArchFromTriple(target.Triple);
             var toolchainPath = Path.Combine(AndroidNdkPath, "platforms",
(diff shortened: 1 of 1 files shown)


marek-safar:
@monojenkins backport 2018-12

PR #12476, opened 2019-01-17
Integration and bug fixes to enable use of roslyn compiler server

kg:
This PR still needs additional cleanup work to be ready to merge but it's getting close enough to be worth running tests and getting code review. The purpose of this PR is to make various bug fixes and infrastructure changes so that it is trivial to use the roslyn compiler server on a clean checkout.

Summary of the changes:
* configure.ac generates some additional variables in config.make that are used to configure the compiler server mode in make. You can then use the compiler server by passing ```ENABLE_COMPILER_SERVER=1``` to make or flipping the switch in config.make.
* config.make has two variables pointing to the compiler server version of roslyn. This lets you use a local build/checkout of roslyn but only when the compiler server is enabled. Once we update in-tree roslyn these vars can default to that. (The in-tree version of csc is not compatible with the compiler server.) The vars are named ```SERVER_CSC_LOCATION``` and ```VBCS_LOCATION```.
* config.make also has a variable that sets the name of the roslyn compiler server pipe. If you're trying to build multiple checkouts in parallel, or build in multiple user accounts, you'll need to use separate pipenames. You'll be able to set this from the command line once I tweak the config file with ```COMPILER_SERVER_PIPENAME=xxx```.
* basic_profile_check attempts to start the compiler server before any builds happen. I put it there because that was the only way I found to get it to always run before mcs builds. I spent a few hours trying to add a separate build target for this, and then gave up because makefiles. (I'd love to get suggestions on a better way to do this) Because it's nontrivial to track compiler server instances across builds, I just spawn a new one every time - this works fine because the compiler server has logic to detect an existing instance.
* I completely rewrote our implementation of Timer and its scheduler, because the existing one was burning an entire core at all times in the compiler server, mostly calling SortedList methods. The new implementation is probably not release-ready, but I focused on keeping it easy to audit and understand because attempts to preserve the existing logic kept breaking. Ideally we can just pull in corefx's timers here, but it looked non-trivial to do that.
* ReaderWriterLockSlim was checking Environment.ProcessorCount every time it was used, which in most configurations will result in a syscall to getpid() and another syscall to check our affinity. This was bottlenecking the compiler server really bad. I changed this to match most other threading code and use PlatformHelper.ProcessorCount instead.
* Our nunit.snk was corrupt (not sure how this happened) so I replaced it with the current nunit.snk from mainline nunit. Thanks to @jaredpar and other roslyn folks for troubleshooting this - it was causing problems for our msbuild infrastructure as well and I never figured that out.

Some work remains before the compiler server is good enough to use by default, which I will probably document here (or in tree? on a wiki?)


mcs/build/start-compiler-server.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+if [ -x "$(command -v mono)" ] ; then
+    if test -f $1; then
+        echo . > /tmp/vbcs-log.txt
+        RoslynCommandLineLogFile=/tmp/vbcs-log.txt mono --gc-params=nursery-size=64m $1 -pipename:$2 &
```
marek-safar: you should pass mono path from the build.make

lambdageek:
If I understand how this works, I think this subtly changes the final artifacts that we produce.

Currently, this is how I imagine our build stages:
1. Build the native bits
2. Build the `[build]` BCL profile and csc using the system mono, system BCL and the system csc
3. Build (for example) the `[net_4_x]` BCL profile using the in-tree mono binary, the `[build]` BCL and the `[build]` csc.

So the final net_4_x BCL profile we get was built using the in-tree compiler running on the in-tree BCL itself.

The new setup with the compile server is:
1. Build the native bits
2. Start the compile server running on the system Mono and the system BCL.
3. Build the `[build]` BCL profile and csc using system csc (calling out to the compile server)
4. Build the `[net_4_x]` BCL profile using the in-tree `[build]` csc running atop the in-tree mono and the `[build]` BCL but calling out to the same system compile server.

So the final net_4_x BCL that we get is built using the system compiler on the system BCL.

I don't know whether this is desirable or not, but I wanted to point out the difference.

(Please let me know if I misunderstood his this works).

kg:
Minor detail I forgot:
* For the compiler server to autostart, mono needs to be in your PATH (so a system mono will do, I just manually set MONO_PATH on invocation). We probably want to support monolite for this once monolite is compatible

@lambdageek You're correct that (if the compiler server is enabled) our outputs are now built using system mono. Everything is going through the compiler server if it's turned on. I'd like to change that so it's the same, though - pipe the same MONO through from make. Getting the right path for it is historically tricky due to how monolite is used, and in this case isn't possible yet because the compiler server is a basic-profile-check prerequisite (and we use basic-profile-check to decide whether to use monolite). The ideal solution would be a post-event that runs after basic-profile-check, but it didn't seem like make has that kind of primitive (wtb msbuild)

In-tree csc isn't used at all by the compiler server path, since it's built from in-tree roslyn. Once in-tree roslyn is updated we could use in-tree csc along with in-tree vbcs.

mcs/build/start-compiler-server.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+if [ -x "$(command -v mono)" ] ; then
+    if test -f $1; then
+        echo . > /tmp/vbcs-log.txt
```
marek-safar: should be just local file (not shared one)

kjpou1:

configure.ac
```diff
@@ -6307,7 +6307,9 @@ fi
       echo "BCL_OPTIMIZE = 1" >> $srcdir/$mcsdir/build/config.make
     fi
 
-    echo "CSC_LOCATION = $CSC_LOCATION" >> $srcdir/$mcsdir/build/config.make
+    echo "STANDALONE_CSC_LOCATION=$CSC_LOCATION" >> $srcdir/$mcsdir/build/config.make
```
kjpou1: @kg This looks like the culprit for the Linux WebAssembly build.

The `CSC_LOCATION` is being used in the wasm build.

https://github.com/mono/mono/blob/master/sdks/wasm/Makefile#L22

kg: Thanks.

kg:
@monojenkins build failed

mcs/build/start-compiler-server.sh
marek-safar: We could add it if it's the only missing

kg: How do I do that? I can do it then

kg:

kg:

kg:
@monojenkins build failed

kjpou1:

kg:

kg:
@monojenkins build failed

kg:
This is ready to merge if I can get a review.

PR #12495, opened 2019-01-18
[sdks] Add net_4_x profile to android-bcl

radekdoulik:
XA needs net_4_x BCL to run jnimarshalmethod-gen.exe tool on
Windows.

Context: https://github.com/xamarin/xamarin-android/pull/2454

### DIFF
--- a/sdks/builds/android.mk
+++ b/sdks/builds/android.mk
@@ -401,5 +401,5 @@ $(eval $(call AndroidCrossMXETemplate,cross-arm64-win,x86_64,aarch64-v8a,android
 $(eval $(call AndroidCrossMXETemplate,cross-x86-win,i686,i686,android-x86,llvm-llvmwin32,i686-none-linux-android))
 $(eval $(call AndroidCrossMXETemplate,cross-x86_64-win,x86_64,x86_64,android-x86_64,llvm-llvmwin64,x86_64-none-linux-android))
 
-$(eval $(call BclTemplate,android-bcl,monodroid monodroid_tools,monodroid))
+$(eval $(call BclTemplate,android-bcl,monodroid monodroid_tools net_4_x,monodroid))
 android_TARGETS += android-bcl


marek-safar:
Why is .NET Framework not enough like with other monodroid tools on Windows?

radekdoulik:
I am not sure whether we run any other tools on windows. What we run into was problems with paths. https://github.com/xamarin/xamarin-android/pull/2454#issuecomment-442012350

radekdoulik:
OK, so we run the other tools on Windows, just not with mono.

marek-safar:
Why cannot we run the tool same way as others

radekdoulik:
Because we need to use `AppDomain`.

We use it to load the XA assemblies and then use LINQ expressions to generate the code for the marshal methods.

sdks/builds/android.mk
```diff
@@ -401,5 +401,5 @@ $(eval $(call AndroidCrossMXETemplate,cross-arm64-win,x86_64,aarch64-v8a,android
 $(eval $(call AndroidCrossMXETemplate,cross-x86-win,i686,i686,android-x86,llvm-llvmwin32,i686-none-linux-android))
 $(eval $(call AndroidCrossMXETemplate,cross-x86_64-win,x86_64,x86_64,android-x86_64,llvm-llvmwin64,x86_64-none-linux-android))
 
-$(eval $(call BclTemplate,android-bcl,monodroid monodroid_tools,monodroid))
+$(eval $(call BclTemplate,android-bcl,monodroid monodroid_tools net_4_x,monodroid))
```
luhenry: Add whatever tool you need to the `monodroid_tools` profile instead.

radekdoulik: The tool (jnimarshalmethod-gen.exe) comes from Java.Interop. Here we need the BCL assemblies, which work correctly on Windows.

Would it make more sense if I add BCL to monodroid_tools profile, compared to adding the net_4_x profile here?

luhenry: `monodroid_tools` is the profile used by XA to build things it will need to ship, and which is not part of the `monodroid` profile. Or put differently: `monodroid` is to run on the device, `monodroid_tools` is to run on the host. So feel free to add anything you require to it, it's only used by XA.

akoeplinger:
Closing this for now, the tool needs to go into monodroid_tools instead of adding net_4_x. And build against .net reference assemblies if the intention is to run against .net framework on windows (which won't work with net_4_x).

PR #12503, opened 2019-01-18
[netcore] Small runtime changes

lambdageek:
1. For `--with-core=only`, disable AppDomains, SRE.Save and Remoting.
2. System.Private.CoreLib System.Array type has some constant fields - don't assert in the runtime that there are 0 fields - when initializing the array type.
3. Don't treat `AppDomainUnloadedException` specially in exception handling.  (And in particular, don't try to load AppDomainUnloadedException - it's not in System.Private.CoreLib.dll, it's in netstandard.dll, so it won't work).


lambdageek:

PR #12504, opened 2019-01-18
[netcore] System.Private.CoreLib changes to match runtime expectations.

lambdageek:
1. Add classes that the runtime depends on:
   * System.Reflection.RuntimeAssembly and System.Reflection.MonoAssembly
   * System.Mono.MonoDomain and System.Mono.MonoDomainSetup
2. Add fields to some classes so they match what `object-internals.h` expect on the native side.
3. Make a non-`abstract` `System.RuntimeType` stub.
   `System.Type` and `System.RuntimeType` need to have fields that match `object-internals.h` and also the runtime wants to create instances of this type and things go wrong if it's abstract.


mcs/class/System.Private.CoreLib/System/AppDomainSetup.cs
```diff
@@ -0,0 +1,38 @@
+using System.Runtime.InteropServices;
+
+namespace System
+{
+	[StructLayout (LayoutKind.Sequential)]
+	public sealed class AppDomainSetup
```
marek-safar: This type does really not exist in Core in this shape, we should not add it at all

mcs/class/System.Private.CoreLib/System.Runtime.Remoting.Contexts/Context.cs
```diff
@@ -0,0 +1,40 @@
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.Remoting.Contexts {
+
+	[StructLayout (LayoutKind.Sequential)]
+	/* FIXME: Mono: this was public in mscorlib */
+	internal class Context
```
marek-safar: Hmm, that's unfortunate we still need this type

mcs/class/System.Private.CoreLib/System/RuntimeType.cs
```diff
@@ -0,0 +1,236 @@
...
+	internal class MonoTypeInfo {
+		// this is the displayed form: special characters
+		// ,+*&*[]\ in the identifier portions of the names
+		// have been escaped with a leading backslash (\)
+		public string full_name;
+		public /*MonoCMethod*/ object default_ctor;
+	}
+
+
+	public class RuntimeType : TypeInfo
```
filipnavara: internal?

mcs/class/System.Private.CoreLib/System/AppDomain.cs
```diff
@@ -0,0 +1,28 @@
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace System {
+
+	[StructLayout (LayoutKind.Sequential)]
+	public sealed partial class AppDomain : MarshalByRefObject {
```
marek-safar: It definitely cannot be public as it'd conflict with CoreFX System.AppDomain. Could we kill this or at least rename this to something else?

mcs/class/System.Private.CoreLib/System/AppDomainSetup.cs
lambdageek: until domains are dead in the runtime, this can't go away.  Right now we create a root AppDomain and its AppDomainSetup

mcs/class/System.Private.CoreLib/System/AppDomain.cs
lambdageek: killing it will be a big project.  I'll see if we can give it a different name for netcore builds...

vargaz: We can name it 'MonoDomain' or 'RuntimeDomain'.

marek-safar: or move it do different namespace, please make it internal

lambdageek:
Moved domain stuff to `System.Mono.MonoDomain` and `System.Mono.MonoDomainSetup`.  also made `System.RuntimeType` `internal sealed` [like coreclr](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/RtType.cs#L51).

lambdageek:

PR #12506, opened 2019-01-18
[debugger][wasm]  Handle with multi-doc methods

thaystg:
When we try to start debugger we got a throw because (c)ctors in corlib has multi-doc methods.
With this implementation it's possible to handle with multi-doc methods.
I tested with the class CurrentSystemTimeZone, I instantiated an object of this type in my cs file, set a breakpoint on it and StepInto, then both files were opened in debugger:
- mcs/class/corlib/corefx/CurrentSystemTimeZone.cs line 50
- external/corefx/src/Common/src/CoreLib/System/CurrentSystemTimeZone.cs line 43, 45, 47, 48, 49 and 50
As it should be.


thaystg:
https://github.com/mono/mono/pull/12540 does the same thing.

PR #12511, opened 2019-01-18
[Debugger] Checking server version before call elapsed_time

thaystg:
The version of the server and client should be bumped to call the elapsed_time.
https://github.com/mono/mono/pull/12217

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -2119,7 +2119,7 @@ namespace Mono.Debugger.Soft
 		}
 
 		internal long Thread_GetElapsedTime (long id) {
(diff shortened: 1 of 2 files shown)


mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
```diff
@@ -2119,7 +2119,9 @@ internal void SetSocketTimeouts (int send_timeout, int receive_timeout, int keep
 		}
 
 		internal long Thread_GetElapsedTime (long id) {
-			return SendReceive (CommandSet.THREAD, (int)CmdThread.GET_ELAPSED_TIME, new PacketWriter ().WriteId (id)).ReadLong ();
+			if (Version.AtLeast (2, 50))
```
vargaz: This should be checked in ThreadMirror.ElapsedTime instead using vm.CheckProtocolVersion (), so calling the public method will fail.

PR #12516, opened 2019-01-18
[sdks] Make android sdk opt-out of AOT file probing

alexanderkyte:
The feature-set added here:

https://github.com/mono/mono/pull/12053

isn't tested by the sdks. This changes that.

### DIFF
--- a/sdks/android/app/src/main/c/runtime-bootstrap.c
+++ b/sdks/android/app/src/main/c/runtime-bootstrap.c
@@ -59,6 +59,27 @@ enum {
         MONO_DL_MASK  = 3
 };
 
+typedef enum {
+	/* Disables AOT mode */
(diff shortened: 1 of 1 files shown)


luhenry:
This is going to make it impossible to test AOT on our end, and this doesn't even verify if it's working properly or not.

We need to have the following:
1. a flag to enable that or not, it can be an environment variable or anything we can pass to Android
1. a test to verify that when setting `MONO_AOT_MODE_NONE`, we do not probe for AOT images.

lewurm:
> 2\. a test to verify that when setting `MONO_AOT_MODE_NONE`, we do not probe for AOT images.

which can be a test on Desktop, for simplicity.

luhenry:
@grendello this is in our own test runner in Mono, it's not used in any way by XA.

PR #12529, opened 2019-01-20
[wasm] Add a --il-strip option to strip IL code from assemblies in AOT mode.

vargaz:


sdks/wasm/packager.cs
```diff
@@ -610,6 +619,9 @@ class WasmOptions {
 		ninja.WriteLine ("  description = [IL-LINK]");
 		ninja.WriteLine ("rule aot-dummy");
 		ninja.WriteLine ("  command = echo > aot-dummy.cs; csc /out:$out /target:library aot-dummy.cs");
+		ninja.WriteLine ("rule ilstrip");
+		ninja.WriteLine ("  command = cp $in $out; mono-cil-strip $out");
```
kjpou1: Where is `mono-cil-strip`?

kjpou1: Shouldn't the `--il-strip` be added to the list of options as well?

vargaz: It is added by this line:
		AddFlag (p, new BoolFlag ("il-strip", "strip IL code from assemblies in AOT mode", opts.ILStrip, b => opts.ILStrip = b));

vargaz: mono-cil-strip is an existing mono tool.

kjpou1: Yes see that now.  I thought it was part of the `wasm-tools`.

kjpou1:
@monojenkins build failed

PR #12532, opened 2019-01-21
[wasm] Distribute wasm_tools.

vargaz:


sdks/wasm/Makefile
```diff
@@ -284,6 +284,7 @@ package: build build-dbg-proxy
 	mkdir tmp/wasm-bcl
 	mkdir tmp/wasm-bcl/wasm
```
kjpou1: Hey Zoltan, for the package an extra `mkdir tmp/wasm-bcl/wasm_tools` is needed.

Issue #11697, opened 2018-11-15
[ios] unify definition of HOST_APPLETVOS et al.


lewurm:
_Originally posted by @rolfbjarne in https://github.com/mono/mono/pull/11691/files#r233817187_

PR #12539, opened 2019-01-21
[tvOS] cleanup defines for tvOS

lewurm:
Fixes https://github.com/mono/mono/issues/11697

### DIFF
--- a/mono/utils/mono-threads-mach.c
+++ b/mono/utils/mono-threads-mach.c
@@ -33,7 +33,7 @@ mono_threads_suspend_init (void)
 	mono_threads_init_dead_letter ();
 }
 
-#if defined(HOST_WATCHOS) || defined(HOST_TVOS)
+#if defined(HOST_WATCHOS)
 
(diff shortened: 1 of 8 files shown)


mono/utils/mono-threads-mach.c
```diff
@@ -33,7 +33,7 @@ mono_threads_suspend_init (void)
 	mono_threads_init_dead_letter ();
 }
 
-#if defined(HOST_WATCHOS) || defined(HOST_TVOS)
+#if defined(HOST_WATCHOS)
```
lewurm: Note, the guarded code by `HOST_TVOS` wasn't really guarded _before_ as `HOST_TVOS` didn't exist. async suspend and friends are available on tvOS, so this was wrong.

lewurm:
I manually verified that this doesn't break Xamarin.iOS on tvOS devices, I didn't test simulator.  The change is only cosmetic, doesn't need a backport.

PR #12546, opened 2019-01-21
[netcore] Make `Assembly.Load (new AssemblyName (corlib_name + ".dll"))` work again

lambdageek:
Partly revert a change in https://github.com/mono/mono/pull/12478

CoreCLR treats `Assembly.Load (new AssemblyName ("System.Private.CoreLib.dll"))`
the same as `Assembly.Load (new AssemblyName ("System.Private.CoreLib"))`  (and
.NET Framework behaves likewise with "mscorlib.dll").

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -4303,9 +4303,10 @@ mono_assembly_load_full_gac_base_default (MonoAssemblyName *aname,
 	 * only requests, like a common reflection only assembly 
 	 */
 	gboolean name_is_corlib = strcmp (aname->name, MONO_ASSEMBLY_CORLIB_NAME) == 0;
-#ifndef NETCORE_LOADER
-	name_is_corlib |= strcmp (aname->name, "mscorlib.dll") == 0;
-#endif
+	/* Assembly.Load (new AssemblyName ("mscorlib.dll")) (respectively,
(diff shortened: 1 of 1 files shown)


mono/metadata/assembly.c
```diff
@@ -4304,7 +4304,7 @@ mono_assembly_load_full_gac_base_default (MonoAssemblyName *aname,
 	 */
 	gboolean name_is_corlib = strcmp (aname->name, MONO_ASSEMBLY_CORLIB_NAME) == 0;
 #ifndef NETCORE_LOADER
-	name_is_corlib |= strcmp (aname->name, "mscorlib.dll") == 0;
+	name_is_corlib = name_is_corlib || strcmp (aname->name, "mscorlib.dll") == 0;
```
marek-safar: Why is this needed for desktop version only?

lambdageek:
@marek-safar Thanks for asking about this.  Actually, looks like CoreCLR treats `Assembly.Load (new AssemblyName ("System.Private.CoreLib.dll"))` the same as `Assembly.Load (new AssemblyName ("System.Private.CoreLib"))`.  (Also checked that other assembly names, e.g. `"System.Runtime"` don't get this special treatment).

So I need to change this.

lambdageek:
Updated the PR

PR #12551, opened 2019-01-22
[wasm] AOT linkage and samples

kjpou1:
- Add two new AOT'd samples targets
- Fix a problem with void methods that are invoked causing memory access crash.
   - The problem stems from void methods returning a corrupted MonoObject when the method that is invoked is marked as `void`.
   - Internally the cwrap'd invoke method is defined as returning a `number` which internally will return the result from `mono_wasm_invoke_method` of the driver.  In the case of `void` methods this may contain invalid address reference causing a memory access crash if referenced.
- binding support fixes
   - Fix a few issues that crop up from time to time. Mostly affects running in an AOT'd application.
- Modify packager linkage for AOT'd code. 
   - Add new experimental command line option

      - --link-mode=`sdkonly|all`        Set the link type used for AOT. (EXPERIMENTAL)
         - `sdkonly` only link the Core libraries.
         - `all` link Core and User assemblies. **(default)**

   - Add the Facades directory to the resolve paths.
   - Treat `netstandard`, `WebAssembly.Bindings` and `WebAssembly.Net.Http` as core libraries.  `monolinker.exe` does not see `netstandard` as a core library.
- Update readme docs

### DIFF
--- a/sdks/wasm/packager.cs
+++ b/sdks/wasm/packager.cs
@@ -319,7 +319,7 @@ class Driver {
 		public bool Linker;
 		public bool LinkIcalls;
 		public bool ILStrip;
-		public bool Verbose;
+		public bool LinkerVerbose;
 	}
 
 	int Run (string[] args) {
@@ -352,7 +352,7 @@ class Driver {
 		var linkModeParm = "all";
 		var linkMode = LinkMode.All;
 		string coremode, usermode;
-		var verbose = false;
+		var linker_verbose = false;
 
 		var opts = new WasmOptions () {
 				AddBinding = true,
@@ -360,7 +360,7 @@ class Driver {
 				DebugRuntime = false,
 				Linker = false,
 				ILStrip = true,
-				Verbose = false
+				LinkerVerbose = false
 			};
 
 		var p = new OptionSet () {
@@ -391,7 +391,7 @@ class Driver {
 		AddFlag (p, new BoolFlag ("binding", "enable the binding engine", opts.AddBinding, b => opts.AddBinding = b));
 		AddFlag (p, new BoolFlag ("link-icalls", "link away unused icalls", opts.LinkIcalls, b => opts.LinkIcalls = b));
 		AddFlag (p, new BoolFlag ("il-strip", "strip IL code from assemblies in AOT mode", opts.ILStrip, b => opts.ILStrip = b));
-		AddFlag (p, new BoolFlag ("verbose", "set verbose option on link", opts.Verbose, b => opts.Verbose = b));
+		AddFlag (p, new BoolFlag ("linker-verbose", "set verbose option on linker", opts.LinkerVerbose, b => opts.LinkerVerbose = b));
 
 		var new_args = p.Parse (args).ToArray ();
 		foreach (var a in new_args) {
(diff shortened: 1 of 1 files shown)


vargaz:
I don't think we should have --link=none, it was always very problematic on ios, the executable size and compilation speed was too big, and the executable size frequently reached OS limits.

kjpou1:
Removed --link-mode=None.

sdks/wasm/packager.cs
```diff
@@ -758,15 +785,40 @@ class WasmOptions {
...
 			foreach (var assembly in root_assemblies) {
 				string filename = Path.GetFileName (assembly);
 				linker_args += $"-a linker-in/{filename} ";
 			}
-			foreach (var assembly in dontlink_assemblies.Keys) {
-				linker_args += $"-p copy {assembly} ";
+
+			// the linker does not consider these core by default
+			// add netstandard as a core library
+			linker_args += $"-p {coremode} netstandard ";
```
marek-safar: This should be fixed in linker

kjpou1: Marek.  We can add it here:  https://github.com/mono/linker/blob/47cd4809b68b1782914e0395aa0f744e80f9f344/linker/Linker/LinkContext.cs#L312

kjpou1:

sdks/wasm/packager.cs
kjpou1: https://github.com/mono/linker/pull/448

kjpou1:
@monojenkins build failed

sdks/wasm/packager.cs
kjpou1: The linker is bumped which includes the fix for `netstandard` as stated above : https://github.com/mono/mono/pull/12560

sdks/wasm/packager.cs
```diff
@@ -60,6 +60,8 @@ class Driver {
 
 	const string BINDINGS_ASM_NAME = "WebAssembly.Bindings";
 	const string BINDINGS_RUNTIME_CLASS_NAME = "WebAssembly.Runtime";
+	const string HTTP_ASM_NAME = "WebAssembly.Net.Http";	
+
```
vargaz: Extra empty line.

kjpou1: done

sdks/wasm/packager.cs
```diff
@@ -374,6 +391,7 @@ class WasmOptions {
 		AddFlag (p, new BoolFlag ("binding", "enable the binding engine", opts.AddBinding, b => opts.AddBinding = b));
 		AddFlag (p, new BoolFlag ("link-icalls", "link away unused icalls", opts.LinkIcalls, b => opts.LinkIcalls = b));
 		AddFlag (p, new BoolFlag ("il-strip", "strip IL code from assemblies in AOT mode", opts.ILStrip, b => opts.ILStrip = b));
+		AddFlag (p, new BoolFlag ("verbose", "set verbose option on link", opts.Verbose, b => opts.Verbose = b));
```
vargaz: This should be linker-version, or the comment should be changed to indicate it sets everything to verbose.

kjpou1: done

PR #12566, opened 2019-01-22
[bcl] Fix the dependencies for the corlib tests.

vargaz:


vargaz:
@monojenkins build failed

PR #12567, opened 2019-01-22
[runtime] Reorganize mono_dl_build_path () a bit, add a new case for netcore.

vargaz:
The netcore native libs are named System.Native.dylib etc. so they don't have the
lib prefix. Add a case for it.


vargaz:
@monojenkins build failed

PR #12565, opened 2019-01-22
[2018-08] Native Crash Stability Fix Batch

alexanderkyte:
This contains the commits from 

https://github.com/mono/mono/pull/12125
https://github.com/mono/mono/pull/12126
and
https://github.com/mono/mono/pull/12518

PR #12573, opened 2019-01-22
[runtime] Fix building older runtimes with newer system Mono

alexanderkyte:
What was happening was that this was building with a Mono.Cecil from
the build profile, but was trying to load the system Mono.Cecil.

Those versions mismatch when building 2018-08 with a nightly mono that is 5.23.0.

Rather than messing with the MONO_PATH and risking shadowing something
important there, I chose to copy the Mono.Cecil files into this
"lib/build/tmp" directory that we are using for the emitted executable.
It is loaded by the system Mono as the correct Mono.Cecil to use, as assemblies in the same folder have precedence.

This fixes the build.

### DIFF
--- a/mcs/tools/cil-stringreplacer/Makefile
+++ b/mcs/tools/cil-stringreplacer/Makefile
@@ -8,10 +8,28 @@ NO_INSTALL = yes
 TARGET_NET_REFERENCE = $(BOOTSTRAP_BIN_PROFILE)
 PROGRAM_USE_INTERMEDIATE_FILE = 1
 
+ifneq ($(PROFILE_RUNTIME),$(EXTERNAL_RUNTIME))
+
 LIB_REFS = System Mono.Cecil
 
 ifdef MCS_MODE
 LIB_REFS += Mono.Cecil.Mdb
-endif
+endif # MCS_MODE
+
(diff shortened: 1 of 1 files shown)


alexanderkyte:
Blocking building tarball to build/merge https://github.com/mono/mono/pull/12565

alexanderkyte:
So it looks like 2018-12 is already copying that file into that path: 
``` 
mcs/class/lib/build/tmp (29a2d7c67e3) $ ls
Mono.Cecil.dll
```

alexanderkyte:
https://gist.github.com/alexanderkyte/38166dbd11df01f8d87d5e50245a4938

alexanderkyte:
Okay, so it looks like this was supposed to have happened as part of the normal build. 

We build $(the_lib) which is ("mcs/class/lib/build") and copy into $(build_lib) which is ("/mcs/class/lib/build/tmp") . 

This change seems to have been made by https://github.com/mono/mono/commit/8d8fd97dbfc6f4213e0fe9fca7574808dd8496d3

alexanderkyte:
This commit is part of 2018-12 but not 2018-08 or 2018-10.

PR #12587, opened 2019-01-23
[runtime] Avoid calling mono_pe_file_time_date_stamp () for images loaded from memory.

vargaz:


mono/metadata/image.c
```diff
@@ -1457,6 +1457,7 @@ do_mono_image_open (const char *fname, MonoImageOpenStatus *status,
 	iinfo = g_new0 (MonoCLIImageInfo, 1);
 	image->image_info = iinfo;
 	image->name = mono_path_resolve_symlinks (fname);
+	image->filename = image->name;
```
lambdageek: Should we strdup here and free in `close_except_pools`?

Otherwise maybe a comment in the header that when filename is non-NULL it is always an alias for `MonoImage:name`.

vargaz:
@monojenkins build failed

PR #12588, opened 2019-01-23
[runtime] Fix mono_pe_file_map logging

alexanderkyte:
Since it's being used when the image is created, it actually doesn't need to be signal safe. This restores the previous behavior.


akoeplinger:
@monojenkins backport 2018-08
@monojenkins backport 2018-10
@monojenkins backport 2018-12

PR #12595, opened 2019-01-24
Fix handling invalid Windows proxy config

### LINKED ISSUE #10030: Bug: AutoWebProxyScriptEngine.InitializeRegistryGlobalProxy
InitializeRegistryGlobalProxy should not error out if ProxyEnable is true and ProxyServer is null. All other applications and services work when my registry has proxyenabled with no proxyserver key. The only test I tried is when using System.Net.WebRequest.Create() and InitializeRegistryGlobalProxy throws a null object exception which I believe is on this line of code:
https://github.com/mono/mono/blob/22c9616fe2d23d11d12b883cb39020a5495c0e16/mcs/class/System/ReferenceSources/AutoWebProxyScriptEngine.cs#L127.

This should get a null check or default value of an empty string e.g. :
1.) set strProxyServer default value to String.Empty or equivalent. 
string strProxyServer = (string)Microsoft.Win32.Registry.GetValue ("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings", "ProxyServer", String.Empty);
2.) null check for strProxyServer could be on the line 127 before checking strProxyServer.Contains("=").

witcher112:
This fixes `NullReferenceException` that was thrown if:
- `HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings\\ProxyEnable` was set to true (1)
- but `HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings\\ProxyServer` was set to null

Fixes #10030

### DIFF
--- a/mcs/class/System/ReferenceSources/AutoWebProxyScriptEngine.cs
+++ b/mcs/class/System/ReferenceSources/AutoWebProxyScriptEngine.cs
@@ -122,6 +122,11 @@ namespace System.Net
 				ArrayList al = new ArrayList ();
 				
 				string strProxyServer = (string)Microsoft.Win32.Registry.GetValue ("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings", "ProxyServer", null);
+				
+				if(strProxyServer == null) {
+					return null;
+				}
+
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

PR #12600, opened 2019-01-24
[wasm] Make cil-strip quiet.

vargaz:


vargaz:
@monojenkins squash

vargaz:
This will not work since older versions of mono-cil-strip don't support the -q argument.

PR #12601, opened 2019-01-24
Add support for Windows x64 winaot, winaot_llvm builds/tests on CI.

lateralusX:
Add support for building Windows x64 fullaot version of winaot profile that can be run on CI. Initial changes also add execution of mini regression tests on CI. More tests will be added going forward and when having full pass rate.

Add support for building Windows x64 fullaot+llvm version of winaot profile that can be run on CI. Initial changes also add execution of mini regression tests on CI. More tests will be added going forward and when having full pass rate.

In order to get deterministic build results and working build on CI, we need to setup a correct build environment for the MSVC builds. Since MSVC build is currently run from cygwin, there are environment problems, both general but also in our CI environment, since we add cygwin bin folder to system path
(needed by Jenkins plugin). When running more complex cmake builds like BTLS and LLVM, this will cause problems picking up incorrect tools, failing builds on CI.

Another problem is the AOT compilation using MSVC toolchain. In the past CI used some "hard coded paths" to make sure clang.exe and link.exe was found. This is not optimal and won't work when including LLVM as part of full AOT since link.exe will need access to more of the full build environment setup by VS development environments (like platform libraries).

In order to get a better build environment on CI this commit includes and enhance a couple of build scripts used to run msbuild.exe and mono-sgen.exe (as AOT compiler) in a VS development environment, isolate that build process from cygwin environment.

The build scripts will locate installed VS tooling, VS2015 build tools or VS dev env, VS2017 build tools or VS dev env and use corresponding development environment scripts as part of running msbuild.exe and mono-sgen.exe in AOT mode.

In order to make sure mono-sgen.exe as AOT compiler is run through these scripts an additional environment variable MONO_EXECUTABLE_WRAPPER is introduced and used in mono-wrapper when set. Without this indirection there will be problems on CI tests since the environment is using MONO_EXECUTABLE and pass that into for example test driver that run child processes of mono-sgen.exe, meaning that we can't use scripts for MONO_EXECUTABLE, but splitting the MONO_EXECUTABLE to represent mono runtime and have a MONO_EXECUTABLE_WRAPPER that could be a launch script solves this issue.

This commit also includes several smaller fixes made to the generic fullaot build infrastructure.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1106,7 +1106,7 @@ AC_ARG_WITH(unreal,              [  --with-unreal=yes,no                If you w
 AC_ARG_WITH(wasm,                [  --with-wasm=yes,no                  If you want to build the WebAssembly (defaults to no)], [], [with_wasm=default])
 
 
-AC_ARG_WITH(runtime-preset, [  --with-runtime-preset=net_4_x,all,aot,aot_llvm,hybridaot,hybridaot_llvm,fullaot,fullaot_llvm,bitcode,unreal,fullaotinterp,fullaotinterp_llvm  Which default profile to build (defaults to net_4_x)],  [], [with_runtime_preset=net_4_x])
+AC_ARG_WITH(runtime-preset, [  --with-runtime-preset=net_4_x,all,aot,aot_llvm,hybridaot,hybridaot_llvm,fullaot,fullaot_llvm,winaot,winaot_llvm,bitcode,unreal,fullaotinterp,fullaotinterp_llvm  Which default profile to build (defaults to net_4_x)],  [], [with_runtime_preset=net_4_x])
 AC_ARG_WITH(spectre-mitigation,             [  --with-spectre-mitigation=yes,no   If you want to build the runtime with compiler flags that enable Spectre mitigation (defaults to no)], [], [with_spectre_mitigation=default])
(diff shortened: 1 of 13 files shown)


lateralusX:
@luhenry Could we get this reviewed?

lateralusX:
Will make a change to look into using our testing_aot_full profile on CI instead of winaot.

lateralusX:
testing_aot_full currently doesn't work on Windows since there is no win32_* versions of sources files meaning that current version includes unix specific source files causing errors when running on Windows. If we are going to use testing_aot_full we will need to go through all sources files and create corresponding win32_testing_aot_full_* making sure we get the correct set of sources build for Windows.

configure.ac
```diff
@@ -1310,9 +1310,23 @@ elif test x$with_runtime_preset = xwinaot; then
    mono_feature_disable_reflection_emit='yes'
    mono_feature_disable_appdomains='yes'
 
-   AOT_BUILD_FLAGS="--runtime=mobile --aot=full,$INVARIANT_AOT_OPTIONS"
+   AOT_BUILD_FLAGS="--runtime=mobile -O=gsharedvt --aot=full,$INVARIANT_AOT_OPTIONS"
```
luhenry: That's surprising you need to pass that with `--aot=full`.

lateralusX: Just did the same as we had on the other full AOT builds, it should still be the default optimization for full AOT, so probably not needed on any of the full AOT test builds.

PR #12613, opened 2019-01-25
[netcore] Parse and use NATIVE_DLL_SEARCH_DIRECTORIES

lambdageek:
I `#ifdef`'d out a bunch of code in `loader.c` that's probing in all sorts of places that I think don't make sense in a netcore world.

It's possible that a MonoDlFallbackHandler is too low-level for this - in the debug trace I see that we look for AOT images in those directories too, for example - should we? I'm not sure.

if we don't like this, the next option is to factor `mono_lookup_pinvoke_call` into separate reusable pieces and hook in netcore in there somewhere.


lambdageek:
@monojenkins build failed

baulig:
From my perspective as somebody who is looking at this code for the very first time, I have a few questions.

In line 1279, there is `for (i = 0; i < 5; ++i)`, but the following `switch` statement only handles cases 0 to 3 (although there is a `default` case that's probably going to handle case 4).

But then after the `switch` there's a larger block of code, the `switch` contains a few conditional `continue` statements and the next block has a bunch of `is_absolute` and `module` conditionals.

Overall, it's quite difficult for me as an outsider to understand what this code does - and adding more `#ifdef` conditionals is certainly not going to help in this regard.

Maybe adding a few comments could help?

PR #12614, opened 2019-01-25
[aot] Print detailed stats only when --aot=stats is used. Avoid emitting 2 copies of preinited got slots in llvmonly mode.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #12622, opened 2019-01-25
Fixing crash where a generic enum is passed into get_shared_type. 

UnityAlex:
Also added an automated test to catch this scenario.

### DIFF
--- a/mono/mini/mini-generic-sharing.c
+++ b/mono/mini/mini-generic-sharing.c
@@ -3856,6 +3856,8 @@ get_shared_type (MonoType *t, MonoType *type)
 	ttype = type->type;
 	if (type->type == MONO_TYPE_VALUETYPE) {
 		ttype = mono_class_enum_basetype_internal (type->data.klass)->type;
+	} else if (type->type == MONO_TYPE_GENERICINST && m_class_is_enumtype(type->data.generic_class->container_class)) {
+		ttype = mono_class_enum_basetype_internal (mono_class_from_mono_type_internal (type))->type;
 	} else if (MONO_TYPE_IS_REFERENCE (type)) {
 		ttype = MONO_TYPE_OBJECT;
 	} else if (type->type == MONO_TYPE_VAR || type->type == MONO_TYPE_MVAR) {
--- a/mono/tests/async-generic-enum.cs
+++ b/mono/tests/async-generic-enum.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Threading.Tasks;
+
+
+public enum someEnum2 {
+    aaa,
+    bbb
+}
+class Tests {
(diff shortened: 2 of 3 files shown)


akoeplinger:
@monojenkins build

mono/mini/mini-generic-sharing.c
```diff
@@ -3856,6 +3856,8 @@ get_shared_type (MonoType *t, MonoType *type)
 	ttype = type->type;
 	if (type->type == MONO_TYPE_VALUETYPE) {
 		ttype = mono_class_enum_basetype_internal (type->data.klass)->type;
+	} else if (type->type == MONO_TYPE_GENERICINST && type->data.generic_class->container_class->enumtype) {
+		ttype = mono_class_enum_basetype_internal (mono_class_create_generic_inst (type->data.generic_class))->type;
 	} else if (MONO_TYPE_IS_REFERENCE (type)) {
```
vargaz: This should call mono_class_from_mono_type_internal () instead of mono_class_create_generic_inst ().

mono/tests/async-generic-enum.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public async Task<anEnum> ThrowExceptionTaskReturn(T val) {
+        for (int i = 0; i < 3; i++) {
+            Console.WriteLine("[ASY] " + (3 - i) + " " + System.Threading.Thread.CurrentThread.ManagedThreadId);
+            await Task.Delay(TimeSpan.FromSeconds(1));
+        }
+
+        var nulla = default(string[]);
+        // Causes exception to fire, walking the stack trace causes nullpointer exception bug
+        var vala = nulla[0];
```
UnityAlex: I think that triggering an exception this way might be causing a native EXC_BAD_ACCESS that's causing lastcrashlog.txt to be generated.

joncham: Perhaps instead of throwing exception would could just force a stacktrace via System.Diagnostics.StackTrace class?

marek-safar:
whitelist

mono/tests/async-generic-enum.cs
UnityAlex: Unfortunately the null reference exception doesn't occur when manually walking the stack with System.Diagnostics.StackTrace. It does however still reproduce if I call `throw new Exception();` instead of intentionally triggering a null reference exception.

luhenry:
@monojenkins build failed

marek-safar:
@vargaz could you re-review

marek-safar:
@monojenkins backport 2019-02

marek-safar:
@monojenkins backport 2018-12

marek-safar:
@monojenkins backport 2018-10

PR #12624, opened 2019-01-25
[bockbuild] Implement handling of symbolic links to directories

alexischr:
/cc @radical


alexischr:

radical:
My patch is similar (https://gist.github.com/radical/6f7f975dcac3e633f152e0ab2de44d52) but it was failing for me in  https://github.com/mono/bockbuild/blob/master/bockbuild/util/util.py#L264 called from https://github.com/mono/bockbuild/blob/master/bockbuild/package.py#L543 , for the symlink. My build with `with_dirs` failed because of something else and I haven't been able to try it after that.

alexischr:
Implemented with https://github.com/mono/bockbuild/commit/5d818c1af4d10c9325a40aa66dc0c994aa3e1781 instead, and the bockbuild submodule was already bumped to have this included.

PR #12625, opened 2019-01-25
[cxx] Set CXXLD and CCLD in configure

### LINKED ISSUE #12435: [cxx] Bypass automake linker detection for .cpp files
Once we begin writing Mono code in C++, we will be using a subset of C++ that does not require linking with the C++ runtime library  (libstdc++ or libc++).  As a result we will have to fight with the [automake automatic linker selection](https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html)

This is different from the existing work we've done:
* The existing work was to compile `.c` files with a c++ compiler but link with a C linker, without the C++ library.

   This doesn't step on automake's toes too much.  We just pass some extra CFLAGS to gcc or clang to pick the correct language.

* The new work is to feed `.cpp` files to a c++ compiler, but link with a C linker, without the C++ library.

   This is more annoying - just adding `.cpp` files to `target_SOURCES` will make `target_LINK` be `$(CXXLINK) args...`.  So we have to do something.

We could just set `target_LINK=$(CXXLINK) args...` ourselves, *but* `args...` is actually pretty finnicky - we have to make sure to pick up all the relevant (general and target specific) `_LDADD` and `_LIBADD` and so forth, all in the correct order.

It would be better if we didn't have to maintain `target_LINK` ourselves.  But it's really not clear how.  

One option, perhaps, is to write our new C++ code in `.c` files and continue lying to automake.  But this is unfriendly to new contributors, to editors that do syntax-highlighting based on file extension, etc...

Another option might be to set `CXXLINK` to `LINK` somewhere (configure.ac? in the makefile?).  But then we'd have to make sure to add `-lc++` or `-lstdc++` when we really do need the C++ library (for example if LLVM is enabled).

We need some better options.

lambdageek:
Essentialy bypass automake's automatic linker selection based on _SOURCES file
extensions.

The issue:
  1. Right now we need to use the C linker unless we're linking with LLVM which
     needs C++.
  2. Later when we start using C++ in Mono we will still want the C linker,
     because we will use a subset of C++ that doesn't depend on the C++ runtime
     library (because we don't want to break embedders that don't expect us to
     use C++).

So in either case which linker we use is a decision that we want to make
independent of what the file extensions look like.

So we set MONO_CXXLD in configure.  If LLVM is enabled, we use $CXX, otherwise
$CC.

Then in mono/mini/Makefile.am.in we set CCLD and CXXLD to MONO_CXXLD.  So
whatever automake decides we're going to use the linker that we want.

This partly reverts 988f4a2 which tried
another approach (which turned out not to work).

This should also address #12060

Closes #12435

---

Also fix a linking issue between llvm-jit.cpp and `mono_dl_symbol` and `mono_dl_close` when both `--enable-llvm` and `--enable-cxx` were used - in that case `llvm-jit.cpp` wanted C symbols but mono-dl was producing C++ symbols.  Now `mono_dl_symbol` and `mono_dl_close` are always C symbols.

### DIFF
--- a/mono/utils/mono-dl.c
+++ b/mono/utils/mono-dl.c
@@ -10,6 +10,7 @@
  * Licensed under the MIT license. See LICENSE file in the project root for full license information.
  */
 #include "config.h"
+#include "mono/utils/mono-compiler.h"
 #include "mono/utils/mono-dl.h"
 #include "mono/utils/mono-embed.h"
 #include "mono/utils/mono-path.h"
--- a/configure.ac
+++ b/configure.ac
@@ -6168,6 +6168,41 @@ AC_SUBST(MONO_NATIVE_PLATFORM_TYPE)
 AC_SUBST(MONO_NATIVE_PLATFORM_TYPE_COMPAT)
 AC_SUBST(MONO_NATIVE_PLATFORM_TYPE_UNIFIED)
 
+### C++ Linker selection
+#
+# We have to lie to autotools, sometimes.
+#
+# If we're building with llvm compiled into Mono, set MONO_CXXLD to CXX,
+# otherwise to CC.  We want C linking even if we're compiling with C++ enabled
+# because we do not want to link inthe C++ runtime library (-lstdc++ or -lc++),
+# except if we're linking with LLVM which relies on it.
+#
+# Explanation of the hack:
+#
+# According to
+# https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html,
+# automake chooses the linker that it used to compile a target based on the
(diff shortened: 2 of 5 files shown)


mono/utils/mono-dl.c
```diff
@@ -10,6 +10,7 @@
  * Licensed under the MIT license. See LICENSE file in the project root for full license information.
  */
 #include "config.h"
+#include "mono/utils/mono-compiler.h"
```
jaykrell: Does this belong in mono-dl.h?

jaykrell: Er, it is already there.

Issue #12535, opened 2019-01-21
The configured user limit (128) on the number of inotify instances has been reached.


mysteryx93:
After updating my Debian server to the latest MONO, the websites wouldn't load anymore (actually only 1 worked out of 5). Very basic websites with plain HTML, many pages have no code-behind or very little, and no database. It was working very well up to this point and I was keeping MONO up-to-date.

I suspect it may be a breaking change with the introduction of LLVM, as it is the first version that took a very long time to install. First time I installed, I got disconnected from the server and it interrupted the process. I rebooted, things were not working. I uninstalled mono-complete and mono-devel and reinstalled everything, and it went through -- but compilation took a long time. Websites still not working.

For now I reverted back to an older version of MONO until we can find what went wrong.

If it makes a difference, I'm using fastcgi mode where I start each server instance with

sudo -H -u www-data /usr/bin/fastcgi-mono-server4 /applications=/:/var/www/spiritualselftransformation/ /socket=tcp:127.0.0.1:9004

I'm running Debian Linux 8

```
System.IO.IOException
The configured user limit (128) on the number of inotify instances has been reached.

Description: HTTP 500.Error processing request.

Details: Non-web exception. Exception origin (name of application or object): System.
Exception stack trace:
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEvents () [0x0007b] in <5bf358e735be486487282a37cb3bce80>:0 
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEventsIfNotDisposed () [0x00019] in <5bf358e735be486487282a37cb3bce80>:0 
  at System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents (System.Boolean value) [0x0001d] in <5bf358e735be486487282a37cb3bce80>:0 
  at (wrapper remoting-invoke-with-check) System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents(bool)
  at System.IO.CoreFXFileSystemWatcherProxy+<>c.<StartDispatching>b__9_0 (System.IO.CoreFX.FileSystemWatcher internal_fsw, System.IO.FileSystemWatcher fsw) [0x00048] in <5bf358e735be486487282a37cb3bce80>:0 
[...]

PR #12627, opened 2019-01-25
[filewatcher] Fall back to default if a new instance cannot be created

alexischr:
Fixes https://github.com/mono/mono/issues/12535

### DIFF
--- a/mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
+++ b/mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
@@ -95,6 +95,20 @@ namespace MonoTests.System.IO
 				fw.Path = "*";
 			}
 		}
+
+		[Test]
+		[Category ("NotOnMac")] // creates resource exhaustion issues
+		public void LargeNumberOfInstances ()
+		{
+			using (var tmp = new TempDirectory ()) {
+				var watchers = new FileSystemWatcher [256];
+				for (int x = 0; x < watchers.Length; x++)
+				{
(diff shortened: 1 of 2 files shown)


marek-safar:
@alexischr seems to cause issues on osx

marek-safar:
@monojenkins build failed

marek-safar:
@monojenkins build failed

steveisok:
@monojenkins build failed

steveisok:
I tried to test this out locally to see if it would fail in the same spot.  Turns out it did.  I wanted to see if a reboot helped and that seemed to do the trick.  I was able to run through the system tests (where it's failing) successfully. 

@alexischr since this has been hanging for a while, can you spot why there would be a hang w/ this PR?

alexischr:
@steveisok I am back to working on this one. I am still not sure which resource is scarce in this case - or which other reason this fails on mac. It looks like it's a mutex failure, but the usual `pthread_mutex` debug line (or Mono's expected `g_error`) do not show up.

The test really should shut down all the FileSystemWatchers; it doesn't because that exposes a different bug, a shutdown race condition. For now, I will disable the test on Mac, and follow-up with a issue/fix for the rapid-shutdown problem.

alexischr:
@monojenkins build failed

alexischr:
There was a crash on System.dll on Windows, don't know if it's related (https://jenkins.mono-project.com/job/w2/791/parsed_console/log.html). I'll run again

alexischr:
@monojenkins build failed

PR #12628, opened 2019-01-25
[corlib] Import System.IO.File from CoreFX (with CoreFX 3.0/master code for subsecond timestamps)

alexischr:


mcs/class/corlib/Test/System.IO/FileTest.cs
```diff
@@ -356,7 +356,6 @@ public void Copy_SourceFileName_DoesNotExist ()
 				Assert.Fail ("#1");
 			} catch (FileNotFoundException ex) {
 				Assert.AreEqual (typeof (FileNotFoundException), ex.GetType (), "#2");
-				Assert.AreEqual ("doesnotexist", ex.FileName, "#3");
```
baulig: What is `ex.FileName`?

mcs/class/corlib/Test/System.IO/FileTest.cs
```diff
@@ -487,7 +486,6 @@ public void Delete_Directory_DoesNotExist ()
 
 			try {
 				File.Delete (path);
-				Assert.Fail ("#1");
```
baulig: I am not sure I quite understand the logic here - removing the `Assert.Fail` only makes sense if the previous `File.Delete` could possibly succeed - and keeping the `catch` only makes sense if it might possibly fail.  Is there a specific reason why this could happen?

Could you please also add some comment to the `try` cause why there shouldn't be an `Assert.Fail`?

mcs/class/corlib/Test/System.IO/FileTest.cs
```diff
@@ -2745,36 +2583,6 @@ void MoveTest (FileAccess acc, FileShare share, bool works)
 			}
 		}
 
-		[Test]
```
baulig: Why have these been removed?

PR #12643, opened 2019-01-27
Take text alignment for LinkLabels into account 

### LINKED ISSUE #12402: System.Windows.Forms: LinkArea of LinkLabels is shifted relative to the text
## Steps to Reproduce

1. Compile and run my attached test project
2. This means: Create a WinForm
3. Create a link label somewhere on the form (bug best visible if centered)
4. Have something like this (see my attached code):
```
this.linkLabel1.Text = "TestTextTest";
this.linkLabel1.Size = new System.Drawing.Size(100, 15);
this.linkLabel1.AutoSize = false;
this.linkLabel1.Location = new System.Drawing.Point(50, 50);
this.linkLabel1.TextAlign = System.Drawing.ContentAlignment.MiddleCenter;
this.linkLabel1.LinkArea = new System.Windows.Forms.LinkArea(0, 12);
```
So: A link label with a size bigger than the text would need, AutoSize false,
and a ~center-aligned text~ text alignment which is *not* top left.

### Current Behavior
The `LinkArea` begins at the start of the label, instead of where the text begins,
and thus ends too early. Same with focus rectangle.
Also the text is not drawn behind the focus/link area.

--> The link area is shifted relative to the text.

grafik
And with a `LinkArea` smaller than the text has chars:
grafik (makes it a bit clearer)

### Expected Behavior
The link area and the focus rectangle should be at the position of the text itself.

[ ] macOS
[x] Linux -> Kubuntu 18.04.1
[ ] Windows

**Version Used**:
mono 5.18.0.225
but already existing on mono 5.16

~linkLabelIssue.zip (MonoDevelop/VS project)
linkLabelIssue-exe.zip (compiled exe)~
For better sample code see the corresponding PR.

DasSkelett:
## Problem
`LinkLabels` ignore the text alignment for rendering the "hitbox" and the text color (= the `LinkArea`), and the focus rectangle. Also every text behind the end of this LinkArea/focus rectangle region is not painted at all.
This becomes visible if `AutoSize = false` and the `Size` is bigger than the text would need,
and if the text alignment is anything different than `TopLeft`.

## Cause
I think this is because the `piece.regions` are only translated regarding the padding, but neglecting the text alignment.

## Solution
Calculate the offset due to alignment and add it to the translation in `LinkLabel.CreateLinkPieces ()`

Fixes #12402 

## One more problem:
~Right now, a very tiny bit at the end of the label is still cut off (see my zipped and attached comparison pictures). Maybe `GetBounds ()` returns a value little bit smaller than it should be.
Can someone help?~

Also, maybe I took the wrong approach to this issue, but I don't know the inner mechanisms of mono.
So feel free to either advise me to change something, or change it yourself(s) :)

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/LinkLabel.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/LinkLabel.cs
@@ -668,71 +668,71 @@ namespace System.Windows.Forms
 										 PaddingClientRectangle,
 										 string_format);
 
-            // Get offset for different text alignments
-            float align_offset_x = 0F;
-            float align_offset_y = 0F;
-            
-            if (TextAlign != ContentAlignment.TopLeft) {
-                Region all_regions = new Region (new Rectangle ());
+			// Get offset for different text alignments
+			float align_offset_x = 0F;
+			float align_offset_y = 0F;
 
-                foreach (Region region in regions) {
-                    all_regions.Union (region);
-                }
+			if (TextAlign != ContentAlignment.TopLeft) {
+				Region all_regions = new Region (new Rectangle ());
 
-                Graphics graphics = CreateGraphics ();
+				foreach (Region region in regions) {
+					all_regions.Union (region);
+				}
 
-                if (TextAlign == ContentAlignment.TopCenter) {
-                    float text_width = all_regions.GetBounds(graphics).Width;
+				Graphics graphics = CreateGraphics ();
 
-                    align_offset_x = (ClientRectangle.Width / 2 - text_width  / 2);
-                }
-                if (TextAlign == ContentAlignment.TopRight) {
-                    float text_width = all_regions.GetBounds(graphics).Width;
+				if (TextAlign == ContentAlignment.TopCenter) {
+					float text_width = all_regions.GetBounds(graphics).Width;
 
-                    align_offset_x = (ClientRectangle.Width - text_width);
-                }
-                if (TextAlign == ContentAlignment.MiddleLeft) {
-                    float text_height = all_regions.GetBounds(graphics).Height;
+					align_offset_x = (ClientRectangle.Width / 2 - text_width  / 2);
+				}
+				if (TextAlign == ContentAlignment.TopRight) {
+					float text_width = all_regions.GetBounds(graphics).Width;
 
-                    align_offset_y = (ClientRectangle.Height / 2 - text_height / 2);
-                }
-                if (TextAlign == ContentAlignment.MiddleCenter) {
-                    float text_width  = all_regions.GetBounds(graphics).Width;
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

filipnavara:
I'll look into it. The change is restricted to a case that was never properly tested, so even if it was not entirely correct the impact would be very limited.

The formatting is little bit off (tabs vs. spaces). It would also help a lot if you could provide a simple test case. An example could be found at https://github.com/mono/mono/blob/master/mcs/class/System.Windows.Forms/samples/issue10235.cs.

mcs/class/System.Windows.Forms/System.Windows.Forms/LinkLabel.cs
```diff
@@ -668,9 +668,73 @@ private void CreateLinkPieces ()
 										 PaddingClientRectangle,
 										 string_format);
 
-			for (int i = 0; i < pieces.Length; i ++) {
+            // Get offset for different text alignments
```
filipnavara: Nit: TABs vs. spaces, the indentation is broken.

mcs/class/System.Windows.Forms/System.Windows.Forms/LinkLabel.cs
```diff
@@ -668,9 +668,73 @@ private void CreateLinkPieces ()
...
+                    align_offset_x = (ClientRectangle.Width - text_width);
+                }
+                if (TextAlign == ContentAlignment.MiddleLeft) {
+                    float text_height = all_regions.GetBounds(graphics).Height;
+
+                    align_offset_y = (ClientRectangle.Height / 2 - text_height / 2);
+                }
+                if (TextAlign == ContentAlignment.MiddleCenter) {
+                    float text_width  = all_regions.GetBounds(graphics).Width;
+                    float text_height = all_regions.GetBounds(graphics).Height;
```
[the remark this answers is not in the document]
filipnavara: Uh. nvm, the diff was off, so I see that you already did that...

DasSkelett:
Here is a zip with a example/test source code and compiled exe, as well as two before/after pics :)
[linkLabelIssue.zip](https://github.com/mono/mono/files/2802884/linkLabelIssue.zip)
Regarding the cut off pixel: I think that was just an illusion due to the small 't'-letter.

mcs/class/System.Windows.Forms/System.Windows.Forms/LinkLabel.cs
DasSkelett: Should be fixed now

filipnavara:
> Here is a zip with a example/test source code and compiled exe, as well as two before/after pics :)

Thanks, I will submit a PR to add it to the Mono repository. It helps with manual validation of complex code changes to collect these small sample projects.

PR #12647, opened 2019-01-28
Fix leak if mono_set_config_dir is called multiple times.

rolfbjarne:
Fix leak if mono_set_config_dir is called multiple times by freeing the
previously stored value.


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

Issue #12480, opened 2019-01-17
Windows socket hangs on WSAWaitForMultipleEvents in threaded environment


UnityAlex:
From my own research into this issue it appears that when sending and receiving threads are sharing the same socket there is a race condition that exists within alertable_socket_wait (in w32socket-w32.c). Specifically when the recv thread is already waiting in WSAWaitForMultipleEvents for more data and a send sharing the same socket uses WSAEventSelect in order to perform it's own wait.

According to: https://docs.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-wsaeventselect

`Issuing a WSAEventSelect for a socket cancels any previous WSAAsyncSelect or WSAEventSelect for the same socket and clears the internal network event record.`

## Steps to Reproduce

1. Open & build sample. File contains both client and server
2. Run server first by providing no arguments to built executable (will bind to localhost 8888)
3. Run client by running built executable with argument "client" supplied `mono.exe <program.exe> client`
4. Client and server will attempt to send messages to one another. Client's message reception thread will hang and stop receiving messages.

The client will attempt to stop running after 30 seconds. The server will attempt to stop running after 1 minute. The server will end successfully when it asks for any key to exit. The client's receiving thread will be waiting on WSAWaitForMultipleEvents even though the socket has been closed by the server. Reproducible only on Windows.

### Repro Code
```
using System;
using System.IO;
using System.Net.Sockets;
using System.Threading;
using System.Timers;

namespace WinSocketRaceCondition
{
    class Program
    {
        #region TCP
        private const string ipAddress = "127.0.0.1";

        private const int port = 8888;

        private Thread readerThread;
        private StreamReader streamReader;
        private StreamWriter streamWriter;
        private NetworkStream networkStream;
        #endregion

        #region Threading
        private bool threadRunning = false;

        Mutex mutex = new Mutex();
        #endregion

        #region Messages
        private const string EOL = "\n";

        private double messagesReceived = 0;
        private double messagesSent = 0;

        private bool logMessages = false;
        private bool run = true;
        private bool isServer = false;
        #endregion

        public Program(bool isClient)
        {
            if (isClient)
                StartClient();
            else
                StartServer();
        }

        void ReaderThread()
        {
            try
            {
                using (streamReader = new StreamReader(networkStream))
                {
                    string line, message, value;
                    while (threadRunning && !streamReader.EndOfStream)
                    {
                        if ((line = streamReader.ReadLine()) != null)
                        {
[...]

PR #12654, opened 2019-01-28
Windows support to abort blocking system calls.

lateralusX:
Windows support to abort blocking system calls has been limited since several of the API's currently used are not using overlapped IO and since system handles are exposed to clients, can't be changed to always work with overlapped IO, even if support could be added in the future, at least for sockets.

In order to add support to abort socket waits,  https://github.com/mono/mono/commit/53af53ae5923ccb205ea8e0ebebdbe0c0ee5de88, did an attempt to implement alertable socket support on Windows, but the implementation was racy and assume that only one thread works against the same socket handle, something that regressed a couple of scenarios, like the one described in https://github.com/mono/mono/issues/12480.

This PR adds optional support to break blocking socket calls using a different approach, leveraging the support we already have in Mono around interrupting/aborting system calls. On Windows this is currently done using APC, but that currently only works for alertable waits and the blocking socket implementation won't return to caller (even if it's using alertable wait behind the scene). Since the abort is currently needed when doing a Thread.Abort the PR includes an additional mechanism where our implementation of blocking system calls registers the current IO handle on current thread MonoThreadInternal structure while doing the blocking call. If that thread gets an abort APC and it's in a blocking IO call region, it will cancel the IO from the queued APC (running on the same thread currently blocking on the call) and that will make sure the thread will break out of any blocking IO call with WSAEINTR or similar error. This builds on the interruptable wait implementation already in place and since the abort logic runs on the same thread as the one register/unregister the IO handle (as part of queued APC) there is no race between the thread requesting an abort and the thread actually running the queued APC.

Since we don't always have full control of the socket/handle used in blocking IO calls, there is a small risk that the socket/handle is not opened to support overlapped IO. If that's the case, the blocking IO call is not doing an alertable wait meaning that it can't run the queued APC so it won't return. This scenario is mitigated by cancel all outstanding sync IO calls on target thread as part of issuing the abort. There is however a small chance that the thread is not yet blocked on sync IO when issuing the cancel request and then it won't be possible to abort the blocking call. NOTE, this is not a problem with blocking calls on socket/handle that uses overlapped IO since the execution of the APC guarantees that the blocking thread has entered an alertable wait as part of waiting for the blocked IO to complete.

### DIFF
--- a/mono/metadata/w32file-win32.c
+++ b/mono/metadata/w32file-win32.c
@@ -87,11 +87,16 @@ gboolean
 mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread, gint32 *win32error)
 {
 	gboolean res;
+	MonoThreadInfo *info = mono_thread_info_current ();
+
+	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
 	MONO_ENTER_GC_SAFE;
-	res = ReadFile (handle, buffer, numbytes, (PDWORD)bytesread, NULL);
+	res = ReadFile ((HANDLE)handle, buffer, numbytes, (PDWORD)bytesread, NULL);
 	if (!res)
 		*win32error = GetLastError ();
 	MONO_EXIT_GC_SAFE;
+	mono_win32_leave_blocking_io_call (info, (HANDLE)handle);
+
 	return res;
 }
 
@@ -99,11 +104,16 @@ gboolean
 mono_w32file_write (gpointer handle, gconstpointer buffer, guint32 numbytes, guint32 *byteswritten, gint32 *win32error)
 {
 	gboolean res;
+	MonoThreadInfo *info = mono_thread_info_current ();
+
+	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
 	MONO_ENTER_GC_SAFE;
-	res = WriteFile (handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
+	res = WriteFile ((HANDLE)handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
 	if (!res)
 		*win32error = GetLastError ();
 	MONO_EXIT_GC_SAFE;
+	mono_win32_leave_blocking_io_call (info, (HANDLE)handle);
+
 	return res;
 }
 
--- a/mono/metadata/w32socket-win32.c
+++ b/mono/metadata/w32socket-win32.c
@@ -38,281 +38,145 @@ mono_w32socket_cleanup (void)
 {
 }
 
-static gboolean set_blocking (SOCKET sock, gboolean block)
-{
-	u_long non_block = block ? 0 : 1;
-	return ioctlsocket (sock, FIONBIO, &non_block) != SOCKET_ERROR;
-}
-
-static DWORD get_socket_timeout (SOCKET sock, int optname)
-{
-	DWORD timeout = 0;
-	int optlen = sizeof (DWORD);
-	if (getsockopt (sock, SOL_SOCKET, optname, (char *)&timeout, &optlen) == SOCKET_ERROR) {
-		WSASetLastError (0);
-		return WSA_INFINITE;
-	}
-	if (timeout == 0)
-		timeout = WSA_INFINITE; // 0 means infinite
-	return timeout;
-}
-
-/*
-* Performs an alertable wait for the specified event (FD_ACCEPT_BIT,
-* FD_CONNECT_BIT, FD_READ_BIT, FD_WRITE_BIT) on the specified socket.
-* Returns TRUE if the event is fired without errors. Calls WSASetLastError()
-* with WSAEINTR and returns FALSE if the thread is alerted. If the event is
-* fired but with an error WSASetLastError() is called to set the error and the
-* function returns FALSE.
-*/
-static gboolean alertable_socket_wait (SOCKET sock, int event_bit)
-{
-	static char const * const EVENT_NAMES[] = { "FD_READ", "FD_WRITE", NULL /*FD_OOB*/, "FD_ACCEPT", "FD_CONNECT", "FD_CLOSE" };
-	gboolean success = FALSE;
-	int error = -1;
-	DWORD timeout = WSA_INFINITE;
-	if (event_bit == FD_READ_BIT || event_bit == FD_WRITE_BIT) {
-		timeout = get_socket_timeout (sock, event_bit == FD_READ_BIT ? SO_RCVTIMEO : SO_SNDTIMEO);
-	}
-	WSASetLastError (0);
-	WSAEVENT event = WSACreateEvent ();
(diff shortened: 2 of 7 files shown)


lateralusX:
All Windows x64 and i386 tests passed with abort syscall enabled. This won't be the default so updating and disabling option as default, retriggering tests.

lateralusX:
Fixes https://github.com/mono/mono/issues/12480.

mono/metadata/w32file-win32.c
```diff
@@ -87,23 +87,33 @@ gboolean
 mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread, gint32 *win32error)
 {
 	gboolean res;
+	MonoThreadInfo *info = mono_thread_info_current ();
+
+	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
 	MONO_ENTER_GC_SAFE;
-	res = ReadFile (handle, buffer, numbytes, (PDWORD)bytesread, NULL);
+	res = ReadFile ((HANDLE)handle, buffer, numbytes, (PDWORD)bytesread, NULL);
```
jaykrell: gpointer and HANDLE are *always* the same. Cast is not needed.

mono/metadata/w32file-win32.c
```diff
@@ -87,23 +87,33 @@ gboolean
...
 	if (!res)
 		*win32error = GetLastError ();
 	MONO_EXIT_GC_SAFE;
+	mono_win32_leave_blocking_io_call (info, (HANDLE)handle);
```
jaykrell: gpointer and HANDLE are *always* the same. Cast is not needed.

mono/metadata/w32file-win32.c
```diff
@@ -87,23 +87,33 @@ gboolean
...
 gboolean
 mono_w32file_write (gpointer handle, gconstpointer buffer, guint32 numbytes, guint32 *byteswritten, gint32 *win32error)
 {
 	gboolean res;
+	MonoThreadInfo *info = mono_thread_info_current ();
+
+	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
 	MONO_ENTER_GC_SAFE;
-	res = WriteFile (handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
+	res = WriteFile ((HANDLE)handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
```
jaykrell: gpointer and HANDLE are *always* the same. Cast is not needed.

mono/metadata/w32file-win32.c
```diff
@@ -87,23 +87,33 @@ gboolean
...
 	if (!res)
 		*win32error = GetLastError ();
 	MONO_EXIT_GC_SAFE;
+	mono_win32_leave_blocking_io_call (info, (HANDLE)handle);
```
jaykrell: gpointer and HANDLE are *always* the same. Cast is not needed.

mono/metadata/w32socket-win32.c
```diff
@@ -38,281 +38,145 @@ mono_w32socket_cleanup (void)
...
+{
+	return internal_w32socket_transmit_file (hSocket, hFile, lpTransmitBuffers, dwReserved, blocking) == 0 ? TRUE : FALSE;
+}
+
+gint
+mono_w32socket_disconnect (SOCKET sock, gboolean reuse)
+{
+	gint ret = SOCKET_ERROR;
+
+	ret = internal_w32socket_disconnect (sock, reuse, TRUE);
```
jaykrell: Combine this line with previous?

mono/utils/mono-threads-windows.c
```diff
@@ -18,6 +18,157 @@
...
 #include <limits.h>
 
+enum Win32APCInfo {
+	WIN32_APC_INFO_CLEARED = 0,
+	WIN32_APC_INFO_ALERTABLE_WAIT_SLOT = 1 << 0,
+	WIN32_APC_INFO_PENDING_INTERRUPT_SLOT = 1 << 1,
+	WIN32_APC_INFO_PENDING_ABORT_SLOT = 1 << 2
+};
+
+static inline void
```
jaykrell: Why inline? Let compiler decide?

mono/utils/mono-threads-windows.c
```diff
@@ -18,6 +18,157 @@
...
+mono_win32_enter_blocking_io_call (THREAD_INFO_TYPE *info, HANDLE io_handle)
+{
+	if (info)
+		enter_alertable_wait_ex (info, io_handle);
+}
+
+void
+mono_win32_leave_blocking_io_call (THREAD_INFO_TYPE *info, HANDLE io_handle)
+{
+	if (info)
```
jaykrell: Consider moving if to callee and then combining the functions.

mono/utils/mono-threads-windows.c
```diff
@@ -18,6 +18,157 @@
...
+	do {
+		old_apc_info = mono_atomic_load_i32 (&info->win32_apc_info);
+		if (old_apc_info & pending_apc_slot)
+			return;
+
+		new_apc_info = old_apc_info | pending_apc_slot;
+	} while (mono_atomic_cas_i32 (&info->win32_apc_info, new_apc_info, old_apc_info) != old_apc_info);
+
+	THREADS_INTERRUPT_DEBUG ("%06d - Interrupting/Aborting syscall in thread %06d", GetCurrentThreadId (), tid);
+	QueueUserAPC (apc_callback, native_thread_handle, (ULONG_PTR)NULL);
```
jaykrell: Can just say `0` instead of `(ULONG_PTR)NULL`.

mono/utils/mono-threads-windows.c
```diff
@@ -18,6 +18,157 @@
...
+suspend_abort_syscall (PVOID thread_info, HANDLE native_thread_handle, DWORD tid)
+{
+	request_interrupt (thread_info, native_thread_handle, WIN32_APC_INFO_PENDING_ABORT_SLOT, abort_apc, tid);
+
+#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+	// In case thread is blocked on sync IO preventing it from running above queued APC, cancel
+	// all outputstanding sync IO for target thread. If its not blocked on a sync IO request, below
+	// call will just fail and nothing will be canceled. If thread is waiting on overlapped IO,
+	// the queued APC will take care of cancel specific outstanding IO requests.
+	CancelSynchronousIo (native_thread_handle);
```
joncham: I am seeing a problem where `suspend_abort_syscall` is called from `mono_threads_suspend_begin_async_suspend`. However, the target thread was already suspended in `mono_threads_suspend_begin_async_suspend`. This seems to cause the call to `CancelSynchronousIo` to block/hang forever. It seems like we want to make the call to `CancelSynchronousIo ` either before suspending the thread or after resuming it, similar to mechanism in `mono_thread_info_install_interrupt` which calls after resuming.

PR #12661, opened 2019-01-28
[2018-12][debugger][backport] Debugger failure when field content is assigned by an unsafe value 

### LINKED ISSUE #12374: Unable to inspect ReadOnlySpan<char> in debugger
## Steps to Reproduce

1. Have an application with ReadOnlySpan being used
```
var span = "asdf".AsSpan();
```
2. Given a StructMirror invoke of `ReadOnlySpan<char>.ToString()`, it'll fail with an `INVALID_ARGUMENT` error code.
3. Maybe ref structs need to be specialized in the debugger when passed?

### Current Behavior

We should get the string.

### Expected Behavior

INVALID_ARGUMENT error code

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.18.0.223 (2018-08/3f799cc3c78 Wed Dec 19 22:12:52 EST 2018)
	TLS:
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       preemptive
```

## Stacktrace

```
```

thaystg:
Backport of #12394 
Debugger failure when a method is called in watChes and any field of the Class has the type of the field different of the value of the field because the assign was done in an unsafe way like this:

Now the datatypes check are not executed only when each field of the class is been checked before the invoke method..
_pinnable = Unsafe.As<Pinnable<T>>(array);
A unit test that reproduces the issue was created either.
Fixes #12374

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Makefile
+++ b/mcs/class/Mono.Debugger.Soft/Makefile
@@ -34,7 +34,7 @@ $(test_output_dir)/dtest-excfilter.exe: Test/dtest-excfilter.il | $(test_output_
 	$(ILASM) -out:$@ /exe /debug Test/dtest-excfilter.il
 
 $(test_output_dir)/dtest-app.exe: Test/dtest-app.cs $(TEST_HELPERS_SOURCES) | $(test_output_dir)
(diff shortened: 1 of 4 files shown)


thaystg:
@monojenkins build faile

thaystg:
@monojenkins build failed

PR #12666, opened 2019-01-28
Fix pipes not being enabled in all relevant versions of the build profile

kg:
I missed this in my compiler server PRs. Without it, the compiler server won't be able to open pipes. (It already works in every other profile).

### DIFF
--- a/mcs/class/System.Core/linux_build_System.Core.dll.sources
+++ b/mcs/class/System.Core/linux_build_System.Core.dll.sources
@@ -1,0 +1,1 @@
+#include unix_net_4_x_System.Core.dll.sources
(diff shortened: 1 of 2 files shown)


baulig:
I don't have any objections against this particular PR - the following is merely a general criticism of the general direction that we're heading into - and I would like to see this as a base of discussion to improve our long-term strategy.

1. There are _a lot_ of `.sources` files in this directory.
2. It is getting very hard to understand which files are enabled in which profile / on which platform.

In this specific context for instance - I have some trouble understanding what the difference between the "build" and "net_4_x" profile is - and especially how this is going to work when cross-compiling for the products.  Or what is going to happen on Windows as I don't see any Windows-specific file for the "build" profile.

That being said, I only wanted to remark this as general concerns and as a basis of discussions, not as criticism of this particular PR (of which I am going to give you a green checkmark on).

kg:
@baulig I agree. I'll make a note to see whether the sources files for this dir can be cleaned up. The complexity is why I made this mistake in the first place.

PR #12668, opened 2019-01-29
Throw ArgumentNullException if null method is passed to delegate cons…

joncham:
…tructor via invoke.

Fixes Unity issue: https://issuetracker.unity3d.com/issues/editor-crashes-with-g-logv-when-entering-play-mode-with-active-flowcanvas-script


marek-safar:
This also need interpreter update

/cc @lewurm

joncham:
I pushed change for interpreter as well.

vargaz:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #12672, opened 2019-01-29
Fix build errors for net_4_x bcl.sln build.

lateralusX:
bcl.sln fails to build 11 tools project due to error around missing type. This is due referencing system mscorlib.dll instead of our own build corlib that doesn't match the system mscorlib.dll causing build errors.

This commit makes sure the tools build --nostdlib and reference our corlib project.

### DIFF
--- a/mcs/ilasm/ilasm.csproj
+++ b/mcs/ilasm/ilasm.csproj
@@ -13,6 +13,7 @@
     <HostPlatform Condition=" '$(HostPlatform)' == '' and '$(OS)' == 'Unix'">linux</HostPlatform>
     <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
     <NoStdLib>False</NoStdLib>
+    <NoStdLib Condition=" '$(Platform)' == 'net_4_x' ">True</NoStdLib>
     <NoConfig>True</NoConfig>
     <AssemblyName>ilasm</AssemblyName>
     <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
@@ -157,6 +158,7 @@
   <!-- @ALL_SOURCES@ -->
   <!-- @COMMON_PROJECT_REFERENCES@ -->
   <ItemGroup Condition=" '$(Platform)' == 'net_4_x' ">
+    <ProjectReference Include="../class/corlib/corlib.csproj" />
     <ProjectReference Include="../class/PEAPI/PEAPI.csproj" />
     <ProjectReference Include="../class/System/System.csproj" />
     <ProjectReference Include="../class/Mono.Security/Mono.Security.csproj" />
--- a/mcs/mcs/mcs.csproj
+++ b/mcs/mcs/mcs.csproj
@@ -13,6 +13,7 @@
     <HostPlatform Condition=" '$(HostPlatform)' == '' and '$(OS)' == 'Unix'">linux</HostPlatform>
     <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
     <NoStdLib>False</NoStdLib>
+    <NoStdLib Condition=" '$(Platform)' == 'net_4_x' ">True</NoStdLib>
     <NoConfig>True</NoConfig>
     <AssemblyName>mcs</AssemblyName>
     <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
@@ -213,6 +214,7 @@
   <!-- @ALL_SOURCES@ -->
   <!-- @COMMON_PROJECT_REFERENCES@ -->
   <ItemGroup Condition=" '$(Platform)' == 'net_4_x' ">
+    <ProjectReference Include="../class/corlib/corlib.csproj" />
     <ProjectReference Include="../class/System.Core/System.Core.csproj" />
     <ProjectReference Include="../class/System.XML/System.Xml.csproj" />
     <ProjectReference Include="../class/System/System.csproj" />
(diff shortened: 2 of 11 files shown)


kg:
The reason tools reference system mscorlib is they won't run otherwise, which makes parts of the build process fail. Running exes that don't reference system mscorlib on windows produces a cryptic target framework error.

kg:
Also, these csproj files are machine generated. You need to edit the template.

marek-safar:
They are all .exe so it looks likes missing logic in the generator

lateralusX:
So is it OK to update the csproj files directly for .exe's or should something be added to the generator?

lateralusX:
I can look into what happened to the missing types that are currently in our corlib, but I believe they are in different assemblies on system framework compared to where we moved them.

kg:
You can't edit these by hand, the project files diff will reject them. You have to update the generator and/or templates, and then do the commit csproj monojenkins command to update them.

This isn't missing functionality in genproj, I added it on purpose. Without doing this it's impossible to compile the BCL without a full win32 install of mono in a known location.

If we're unable to build because of differences between desktop netframework mscorlib and mono mscorlib, we'll need to figure out a solution for that. Maybe we raise our target .net framework version.

lateralusX:
@kg, do you have any other PR that shows the progress updating template/generator so I know how it should be done?

lateralusX:
So the error we have is this:

kg:
https://github.com/mono/mono/commits/master/msvc/scripts/genproj.cs

If you have specific questions about how to change it let me know. It seems like the fix to this might just be a one-off change to our System so it has an alias pointing to the right place in System. Alternately reference the platform System instead of the one we built so that Stack is there.

lateralusX:
When building for example ilasm.exe using make build system for net_4_x it actually uses -nostdlib passing the net_4_x profile version of mscorlib.dll and system.dll. This will cause a hard dependency to our mscorlib, so that build version will not work against another framework:
 
`echo "CSC     [net_4_x-win32] ilasm.exe"; MONO_PATH="./../class/lib/build;$MONO_PATH" CSC_SDK_PATH_DISABLED= /cygdrive/c/github/mono-alt/runtime/mono-wrapper  --aot-path=/cygdrive/c/github/mono-alt/mcs/class/lib/build --gc-params=nursery-size=64m C:/github/mono-alt/external/roslyn-binaries/Microsoft.Net.Compilers/Microsoft.Net.Compilers.2.8.2/tools/csc.exe /codepage:65001 /nologo /noconfig /deterministic -d:HAS_MONO_SECURITY  -d:NET_4_0 -d:NET_4_5 -d:NET_4_6 -d:MONO -d:WIN_PLATFORM -nowarn:1699 -nostdlib /debug:portable -optimize /features:peverify-compat /langversion:latest -r:./../class/lib/net_4_x-win32/mscorlib.dll -r:./../class/lib/net_4_x-win32/PEAPI.dll -r:./../class/lib/net_4_x-win32/System.dll -r:./../class/lib/net_4_x-win32/Mono.Security.dll -target:exe -out:../class/lib/net_4_x-win32/ilasm.exe ILParser.cs  @./../build/deps/ilasm.exe.sources.response
`
This means that the ilasm.exe that is currently build from our net_4_x profile will only work against our net_4_x profile (the same happens for all except our build profile). So the question is then, shouldn't the build from bcl.sln give the same results as when building the same profile using make.


Since the tools should run under desktop framework as well as Mono, I believe the tools csprojs shouldn't build directly against our library projects (that will be Mono specific as showed by the above compile errors) but instead use the assemblies available from targeted framework or using the \external\binary-reference-assemblies. This should prevent future problems related to building against profile specific assemblies. As a simple start I could update the tool to make sure assemblies

lateralusX:
So if we expect the tools to work on both .net framework as well as Mono's net_4_x profile then we need to build against .net frameworks version of system assembly as well in order to get around compile errors.

kg:
I agree with your interpretation, I think we should always build tools without doing -nostdlib unless there's a reason to do otherwise.

marek-safar:
We don't expect tools built for net_4_x to work on .NET. What we usually do instead is to have separate csproj/sln for each tool.

lateralusX:
So to summarize (so I need how we should fix the build errors). When I build ilasm.exe (or any of the other tools) using bcl.sln solutions net_4_x configuration, what do we expect, a copy of ilasm.exe that only works under Mono's net_4_x profile or a copy that works on both .net framework as well as Mono net_4_x?

kg:
To restate, if we only want them to work against mono.exe (which is fine, I guess?) then it's impossible to actually build the BCL under visual studio, because our build relies on our tools.

marek-safar:
@lateralusX when building as part of net_4_x it's intended to run on with mono only

lateralusX:
@kg what parts of the BCL build can't be done under VS if we don't build the tools working against .net framework? Is it the dependency on genconsts and cil-stringreplacer that corlib build is depending on? 

Looks like they already reference the .net framework version of system for thoose tools, that's why most of the build works, since they won't hit the compile errors hit by other tools referencing net_4_x version of system., but .net frameworks mscorlib. These tools also have special handling in genproj.cs compared to the other tools currently hitting the compile errors.

So it sounds like we have a small variations on what the bcl.sln builds will produce compared to the make build. It also sounds like the bcl.sln would need something similar to the "build" profile used by the make system. That profile is build targeting binary-reference-assemblies for mscorlib and system and used when building parts of other profiles. Doing something similar is probably a bigger change to how the bcl.sln works and might force users to first build the "build" profile and then net_4_x in VS. Since the bcl.sln already have the assumption that the tools should build against .net framework mscorlib they quickest way to fix the build errors is to build all tools similar to how we build them under "build" profile, using reference assemblies for both mscorlib and system. This involves changes to genproj.cs to handle the case of .exe (already handled) and how they reference system. Would that be an acceptable fix to do?

Issue #12255, opened 2019-01-03
Update System.Data.Entity to handle new String overloads


marek-safar:
After adding new string overloads we need to update System.Data.Entity mapping table otherwise the new overloads won't be recognized leading to unhandled exceptions for the new best overloads

The code to be updates can be fount at https://github.com/mono/mono/blob/40ec59d65b22590c6a5f4ed065d9f0a04de3db39/mcs/class/referencesource/System.Data.Entity/System/Data/Objects/ELinq/MethodCallTranslator.cs#L972

Trim/TrimStart/TrimEnd are already failing but we should check all

EgorBo:
I wonder who uses this BCL assembly at all, when users run EF6 on top of mono  - they use EF's System.Data.Entity, don't they?

PR #12680, opened 2019-01-30
Update System.Data.Entity to handle new String overloads

MaximLipnin:
Relates to https://github.com/mono/mono/issues/12255


mcs/class/referencesource/System.Data.Entity/System/Data/Objects/ELinq/MethodCallTranslator.cs
```diff
@@ -1008,6 +1014,13 @@ internal IndexOfTranslator()
                 private static IEnumerable<MethodInfo> GetMethods()
                 {
                     yield return typeof(String).GetMethod("IndexOf", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(string) }, null);
+#if MONO
+                    yield return typeof(String).GetMethod("IndexOf", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(char) }, null);
```
marek-safar: Remove methods which take more than 1 argument, the code can handle only `1 argument for String.IndexOf`

MaximLipnin: Done

mcs/class/referencesource/System.Data.Entity/System/Data/Objects/ELinq/MethodCallTranslator.cs
```diff
@@ -1363,6 +1378,10 @@ internal TrimTranslator()
                 private static IEnumerable<MethodInfo> GetMethods()
                 {
                     yield return typeof(String).GetMethod("Trim", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(Char[]) }, null);
+#if MONO
+                    yield return typeof(String).GetMethod("Trim", BindingFlags.Public | BindingFlags.Instance, null, Type.EmptyTypes, null);
```
marek-safar: This will still fail because the expects argument to be an array, see `if (!IsEmptyArray(call.Arguments[0]))` check in base class

MaximLipnin: It will fail a bit earlier, during method translators initialization because `CanonicalFunctionDefaultTranslator` already contains this:

https://github.com/mono/mono/blob/9177dc18a0099251dcdc751939f1fec39c23b947/mcs/class/referencesource/System.Data.Entity/System/Data/Objects/ELinq/MethodCallTranslator.cs#L844

MaximLipnin:
Does `RuntimeMethodInfo.GetMethodFromHandleInternalType` method work correctly?https://github.com/mono/mono/blob/4e3264f975668b40db0a8e409d2576485d9f0b2b/mcs/class/corlib/ReferenceSources/MethodBase.cs#L80

It gets invoked during executing the code like this `Expression<Func<ObjectContext, string>> l = s => "asd".Trim();` and returns `{System.String Trim()}`

But for `TrimStart()` it returns `{System.String TrimStart(Char[])}`.

MaximLipnin:
`Trim` overloads:
- `Trim()` is considered as `Trim(new char[] {})` and seems it can be translated;
- `Trim(new char[] { 'c' })` will fail because only an empty char array as argument is supported;
- `Trim('c')` is considered as `Trim(new char[] { 'c' })` and will fail too.

Same for `TrimStart` and `TrimEnd`.

PR #12685, opened 2019-01-30
[bcl] Fix possible deadlock race in CSharpCodeGenerator

grendello:
`CSharpCodeGenerator` runs a compiler while redirecting its standard error and
standard output streams for further reporting. However, in doing that it fails
to read the standard output data completely and also fails to properly wait for
the standard error output to complete. Depending on environment it may lead to
the process using `CSharpCodeGenerator` hanging while the process has already
exited but there still is data to be read from standard output and standard
error.

This commit changes the code to not redirect standard output at all, as it
appears to be unused, and switches to implementation of reading standard error
that makes sure all the data is read before the process is closed.

The hope is that it may fix some of the timeouts/hangs we see on the
Xamarin.Android build bots as a portion of them hangs in test that use the
code generator.

### DIFF
--- a/mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
+++ b/mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
@@ -67,7 +67,8 @@ namespace Microsoft.CSharp
 
 			mcs.StartInfo.Arguments += BuildArgs (options, fileNames, _provOptions);
 
-			var mcsOutMutex = new Mutex ();
+			var stderr_completed = new ManualResetEvent (false);
+			var stdout_completed = new ManualResetEvent (false);
 /*		       
 			string monoPath = Environment.GetEnvironmentVariable ("MONO_PATH");
 			if (monoPath != null)
@@ -93,15 +94,19 @@ namespace Microsoft.CSharp
 
 			mcs.StartInfo.CreateNoWindow=true;
 			mcs.StartInfo.UseShellExecute=false;
+			mcs.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
 			mcs.StartInfo.RedirectStandardOutput=true;
 			mcs.StartInfo.RedirectStandardError=true;
 			mcs.ErrorDataReceived += new DataReceivedEventHandler ((sender, args) => {
-				if (args.Data != null) {
-					mcsOutMutex.WaitOne ();
+				if (args.Data != null)
 					results.Output.Add (args.Data);
-					mcsOutMutex.ReleaseMutex ();
-				}
+				else
+					stderr_completed.Set ();
 			});
+			mcs.OutputDataReceived += new DataReceivedEventHandler ((sender, args) => {
+					if (args.Data == null)
+						stdout_completed.Set ();
+				});
 
 			// Use same text decoder as mcs and not user set values in Console
 			mcs.StartInfo.StandardOutputEncoding =
@@ -125,8 +130,8 @@ namespace Microsoft.CSharp
 				
 				results.NativeCompilerReturnValue = mcs.ExitCode;
 			} finally {
-				mcs.CancelErrorRead ();
-				mcs.CancelOutputRead ();
+				stderr_completed.WaitOne (TimeSpan.FromSeconds (30));
+				stdout_completed.WaitOne (TimeSpan.FromSeconds (30));
 				mcs.Close();
 			}
 


kg:
The logic behind this fix makes sense, but I think a better approach is to simply not use the (chronically buggy and just extremely not good) readline APIs and instead do a sync ReadToEnd on both stdout and stderr on separate threads (threadpool should be fine). You might want to do ReadToEnd on stdout on the main thread, or continue to WaitForExit on the main thread with the stdout/stderr reads happening on workers. If threads are undesirable you can perhaps use 'async' and do async reads on the streams until they complete - maybe the BCL has taskified non-blocking ReadToEnd by now I'm not aware of.

Ignoring stdout as this does means it might dump into the console of whatever is using this class, which isn't great. It may not be safe to redirect stderr while not redirecting stdout, depending on how everything in your chain is written.

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
```diff
@@ -93,18 +93,16 @@ private CompilerResults FromFileBatch (CompilerParameters options, string[] file
 
 			mcs.StartInfo.CreateNoWindow=true;
 			mcs.StartInfo.UseShellExecute=false;
-			mcs.StartInfo.RedirectStandardOutput=true;
```
akoeplinger: We should keep this and instead do the same what we do for stderr (and just discard the data).

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
```diff
@@ -93,18 +93,16 @@ private CompilerResults FromFileBatch (CompilerParameters options, string[] file
...
+			mcs.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
 			mcs.StartInfo.RedirectStandardError=true;
 			mcs.ErrorDataReceived += new DataReceivedEventHandler ((sender, args) => {
-				if (args.Data != null) {
-					mcsOutMutex.WaitOne ();
+				if (args.Data != null)
 					results.Output.Add (args.Data);
```
akoeplinger: I looked into some of our homegrown test-runner implementations which also do Process.Start() and they `lock` inside of the event handler before appending to the stringbuilder, maybe add that here too?

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
```diff
@@ -119,14 +117,12 @@ private CompilerResults FromFileBatch (CompilerParameters options, string[] file
...
 			try {
-				mcs.BeginOutputReadLine ();
 				mcs.BeginErrorReadLine ();
 				mcs.WaitForExit();
 				
 				results.NativeCompilerReturnValue = mcs.ExitCode;
 			} finally {
-				mcs.CancelErrorRead ();
-				mcs.CancelOutputRead ();
+				stderr_completed.WaitOne (TimeSpan.FromSeconds (30));
```
akoeplinger: maybe make this 15 seconds instead? seems like a long enough time.

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
akoeplinger: Talking with @kg on Slack this might be the actual underlying cause of the bug - if an exception happens on the event handler the async read breaks.

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
grendello: Yes, that was my first impression too, but then the code was ignoring `stdout` and I don't really know what would/should we do with it? Just discard?

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
grendello: The lock inside the handler is not needed - it won't be called simultaneously on several threads.

mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
grendello: I'm erring on the side of safety here. In practice, it rarely ever takes more than 500ms. And it's better to be safe than sorry should we hit a situation when the delay is longer.

grendello:
> The logic behind this fix makes sense, but I think a better approach is to simply not use the (chronically buggy and just extremely not good) readline APIs and instead do a sync ReadToEnd on both stdout and stderr on separate threads (threadpool should be fine). You might want to do ReadToEnd on stdout on the main thread, or continue to WaitForExit on the main thread with the stdout/stderr reads happening on workers. If threads are undesirable you can perhaps use 'async' and do async reads on the streams until they complete - maybe the BCL has taskified non-blocking ReadToEnd by now I'm not aware of.

That's complicating it too much. I had this kind of code running in the Xamarin Installer for years now, it's foolproof, simple and easy to follow. You *really* don't want to introduce threading in process control - that's invitation (especially with the `Process` class) for disaster. Been there, done that :)


> 
> Ignoring stdout as this does means it might dump into the console of whatever is using this class, which isn't great. It may not be safe to redirect stderr while not redirecting stdout, depending on how everything in your chain is written.

As commented elsewhere, I simply didn't know what to do with the output and seeing that it was being ignored, I'd just removed the redirect. I'm not against capturing it and discarding, if you think this is the way we should go.

grendello:
Updated the PR with code to capture and ignore `stdout`

grendello:
build

akoeplinger:
@monojenkins backport 2018-12
@monojenkins backport 2018-10
@monojenkins backport 2018-08

PR #12694, opened 2019-01-31
add missing typeforwardes in facades

roozbehid:
These typeforwardes were required to run Winform windows .net application with mono


marek-safar:
approve

marek-safar:
The type forwarders should not be used for compilation and the type is actually in the official package either. Where did you get this error from?

roozbehid:
Actually I got the error during runtime. Running winforms application in
webassembly. When using timers in winforms.

marek-safar:
@roozbehid that's strange, could you share the bin or at least the file which requires this forwarding?

roozbehid:
@marek-safar Sure.

https://github.com/roozbehid/WasmWinforms/blob/c9aa86e07dd8a3f0dc2f0cfaba08b53fb847e70f/Samples/WinFormSample-FullNet/Form1.Designer.cs#L8

and

https://github.com/roozbehid/WasmWinforms/blob/c9aa86e07dd8a3f0dc2f0cfaba08b53fb847e70f/Samples/WinFormSample-FullNet/Form1.Designer.cs#L57

Both require those typeforwarders if not they will generate error during runtime.

This file was outcome of using Visual Studio form designer, no changes were made to the file.
I did compile the program using Visual Studio CSC, then used packager.exe to make necessary changes and then run it in the browser.

You can see the binary files and download them from [here](https://webassembly.blob.core.windows.net/$web?comp=list)

marek-safar:
I couldn't find any reference to ISupportInitialize or Container in attached binaries

roozbehid:
@marek-safar I am a bit confused.
Can you look into [this file](https://webassembly.blob.core.windows.net/$web/managed/WinFormSimple1CoreNuget.dll)
I opened it in ILSpy and I can see those references. If you go to ILSpy, go to Form1.InitializeComponent, you should see them.

and also as you can see from the code lines they are referenced. Can you explain more what should I exactly show?

marek-safar:
That's interesting, it references version 4.2.1 of System.ComponentModel.Primitives but I cannot find that version anywhere. Where did you get it from?

roozbehid:
> That's interesting, it references version 4.2.1 of System.ComponentModel.Primitives but I cannot find that version anywhere. Where did you get it from?

Could it be because I am compiling it with Visual Studio C#/Net Core? Because of Winforms editor that only works in Visual Studio I have to use it. 
So my workflow is doing the visual design in Visual Studio and compile it. Then I use packager.exe (from mono) to change references to mono assemblies. (or I think that is what packager does.)

I am doing all of these in Windows.

marek-safar:
Could you attach whole project to see where the references are coming from?

roozbehid:
> Could you attach whole project to see where the references are coming from?

This is project file
https://github.com/roozbehid/WasmWinforms/blob/master/Samples/WinFormSample-NetCore/WinFormSample-NetCore.csproj

if you look into here

https://github.com/roozbehid/WasmWinforms/tree/master/Samples

WinFormSample-FullNet has the source codes, but project file is in WinformSample-NetCore

akoeplinger:
The types are there in the assemblies on .NET Core:
- [System.ComponentModel.Container](https://github.com/dotnet/corefx/blob/8c5260061b11323dfd97fbab614d54402405513f/src/System.ComponentModel.TypeConverter/ref/System.ComponentModel.cs#L170)
- [System.ComponentModel.ISupportInitialize](https://github.com/dotnet/corefx/blob/8c5260061b11323dfd97fbab614d54402405513f/src/System.ComponentModel.Primitives/ref/System.ComponentModel.Primitives.cs#L218)

Your binary compiled on Windows with the .NET Core-based WinForms Designer in VS will have these references in it.

@marek-safar I think there are many more such cases (I hit these in the past too when trying to run a vanilla .NET Core console app on Mono), so should we just accept this one or make a full pass over all the .NET Core assemblies?

marek-safar:
ok, will get this in but I don't think we should do the full pass over all the .NET Core assemblies

marek-safar:
@monojenkins commit apidiff

PR #12712, opened 2019-01-31
[2018-04] [Network] Fix CVE 2018-8292 on Mac OS X. (#12153)

spouliot:
* [Network] Fix CVE 2018-8292 on Mac OS X.

This commit ensures that CVE 2018-8292 is fixed by not setting
auto-redirect in the first request done by CFNetwork. CFNetwork has the
same bug, but we can work around it by performing the first request,
checking the response and, if it is a rediret, follow it without the
Authorization headers.

* No need to close the obc, it can be reused.

* Small typo.

* Fix error while chaging the request :/

* Minor style changes.


spouliot:
backport to 2018-04 (it was not public when the package was produced)
but we need further fixes in that branch...

PR #12725, opened 2019-02-01
[bcl] Fix assembly identity of nunit/nunitlite

akoeplinger:
The new strong name key added in https://github.com/mono/mono/commit/acaa73b7026897f7e83a101a9bc331476b5ed169 changed the assembly identity which we can't do for shipping libraries.

Given that the old .snk is invalid (see https://github.com/dotnet/roslyn/issues/32502), we just sign with the extracted public key and /publicsign.

For MCS_MODE builds we instead use the mono.snk since mcs doesn't support /publicsign
and we don't really care about having the correct identity there.


akoeplinger:
@monojenkins commit csproj

akoeplinger:
merging to unblock the windows build

PR #12743, opened 2019-02-01
[bcl] Actually fix the nunit/nunitlite assembly identity

akoeplinger:
The solution from https://github.com/mono/mono/pull/12725 doesn't fully work since e.g. gacutil seems to verify the strong name.

Instead follow what we do for other keys where we don't have the private key: add them to the machine.config pubTokenMapping list.


akoeplinger:
Merging to unblock the build, I'll address any feedback separately.

PR #12764, opened 2019-02-04
[WIP] Change config paths to ~/.config/mono

aaronfranke:
Previously, a mixture of `~/.mono` and `~/.config/.mono` was used, which is confusing and does not follow the XDG directory standard. Configs should be folders in `~/.config` and not be hidden.

I'm far from a Mono expert. This probably breaks a lot of things as-is, so I've marked it as [WIP], but at least it compiles fine. I mostly did this with `grep` and find/replace. I figured I'd open a pull request rather than an issue, to make an attempt at solving this, but if something is broken here I have no idea how to fix it. Feel free to take these changes and commit them separately if you wish.

See also: [Directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), https://github.com/mono/monodevelop/issues/7037, [this video](https://www.youtube.com/watch?v=AFtfpluqv14), [this article](https://0x46.net/thoughts/2019/02/01/dotfile-madness/).

### DIFF
--- a/man/certmgr.1
+++ b/man/certmgr.1
@@ -106,10 +106,10 @@ Display help about this tool.
 The only safe way to interact with certificate stores is to use the certmgr
 tool. The current releases of Mono keeps all the user certificate stores in 
 separates directories under
-.I ~/.config/.mono/certs/
+.I ~/.config/mono/certs/
 .TP
 For example the trusted root certificates for a user would be kept under
-.I ~/.config/.mono/certs/Trust/
+.I ~/.config/mono/certs/Trust/
 .TP
 Certificates files are kept in DER (binary) format (extension .cer).
 .TP
@@ -122,7 +122,7 @@ The filenames either start with
 The rest of the filename is the base64-encoded value (tbp or ski).
 .TP
 Private key data is stored under
-.I ~/.config/.mono/keypairs/
+.I ~/.config/mono/keypairs/
 
 .SH EXAMPLES
 .TP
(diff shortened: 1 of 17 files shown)


NattyNarwhal:
This shouldn't hardcode paths specified by XDG (except for fallback), but instead use the appropriate environment variables.

aaronfranke:
Yes, though I do have some questions. For one, how should it appear in the docs? For another, in the C# scripts, `Environment.SpecialFolder.ApplicationData` versus `$XDG_CONFIG_HOME`? And there will need to be checks for whether or not `$XDG_CONFIG_HOME` exists, and I'm not familiar with how environment variables work in all of the APIs in use here (C and makefile).

migueldeicaza:
This breaks things, so I am afraid we can’t take this

aaronfranke:
@migueldeicaza Yes, but please make this change yourselves at some point.

aaronfranke:
@Lassebq Mono has been mostly deprecated, and the developers have shown no interest in fixing this issue.

Lassebq:
Deprecated? What is it replaced by? I don't think there's an alternative at the moment for running .NET Framework on Linux. Other than wine.

aaronfranke:
@Lassebq Microsoft started developing .NET Core in 2016, which is a rewrite from scratch that is cross platform from the start, instead of Mono which ports over Windows-focused .NET Framework APIs to non-Windows platforms. As of 2020, when .NET Core was (confusingly) rebranded to just ".NET" with the release of .NET 5 (no relation to the old .NET Framework 4.x codebase), the new .NET (Core) officially replaces both .NET Framework and Mono, deprecating both. You can download the new .NET (Core) for Linux here: https://dotnet.microsoft.com/en-us/download

Issue #12421, opened 2019-01-14
Task calls SynchronizationContext.Post on cancelling


AndreyFromGomel:
After updating Visual Studio from version 15.7 to 15.9 we noticed regression of our Android application. Investigation showed that for mono 5.10.1 and 5.14.0 following code works different:
```
        private async void StartButton_OnClicked(object sender, EventArgs e)
        {
            _tcs = new TaskCompletionSource<bool>();
            _cts = new CancellationTokenSource();
            Label.Text = "Started";

            _cts.Token.Register(() => _tcs.TrySetCanceled());
            try
            {
                await _tcs.Task;
            }
            catch (OperationCanceledException)
            {
                Label.Text += " -> OCE ";
                Console.WriteLine("Catching OperationCanceledException");
            }
        }

        private void StopButton_OnClicked(object sender, EventArgs e)
        {
            _cts.Cancel();
            Label.Text += " -> SBC ";
            Console.WriteLine("Stop button clicked");
        }
```
Further investigation revealed that mono 5.14.0 calls SynchronizationContext.Post method while 5.10.1 does not.
I created repository with POC: https://github.com/AndreyFromGomel/TestSyncContext

1. Launch
2. Click Start
3. Click Stop

Output for mono 5.14.0:
 ```
Stop button clicked
Catching OperationCanceledException
```
Output for mono 5.10.1:
 ```
Catching OperationCanceledException
Stop button clicked
```
The same behavior for Xamarin.iOS (12.2), Xamarin.Android (9.1) and Xamarin.Mac (5.2) with mono 5.14.0.
My question: is this expected behavior or this is issue and will be fixed in further mono releases?

EgorBo:
"SBC -> OCE" (mono 5.14) behavior looks more reasonable, doesn't it? (WPF+net45 behavior).
I am not sure what was the reason of this change

>Further investigation revealed that mono 5.14.0 calls SynchronizationContext.Post method while 5.10.1 does not.

How did you log that event, did you override `SynchronizationContext.SetSynchronizationContext`? I don't see any redundant Post after I click "Start" button.

AndreyFromGomel:
@EgorBo First of all thank you for your reply.
> "SBC -> OCE" (mono 5.14) behavior looks more reasonable, doesn't it? (WPF+net45 behavior).

I'm agree with you that it is more reasonable, but if we check this behavior on Windows Forms (.NET 4.6.1) we get "OCE -> SBC" (like mono 5.10.1).
[...]

PR #12774, opened 2019-02-04
[corlib] Do not clone SynchronizationContext in ExecutionContext.Capture

EgorBo:
Fixes https://github.com/mono/mono/issues/12421
When we switched to CoreFX/CoreRT sources (mostly CoreRT) for System.Threading we had to keep old reference-source based `ExecutionContext`, `CancellationToken` and `CancellationTokenSource`.  In the bug above the redundant `SynchronizationContext.Post` was triggered here:
https://github.com/mono/corert/blob/1b7d4a1e4bd79305905338827cc41eb8303364f0/src/System.Private.CoreLib/src/System/Threading/Tasks/TaskContinuation.cs#L404

it was expected to be "myCustomSyncContext == myCustomSyncContext" but instead it was "myCustomSyncContext == empty SynchronizationContext instance".
That empty instance was allocated in `CreateCopy` which I am trying to delete in this PR.
That `.CreateCopy()` always creates `SynchronizationContext` instance (see [here](https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/threading/synchronizationcontext.cs#L363-L368), but it's expected) so `==` is `false` in our case => it goes to the "Post" path instead of "Execute synchronously". 
PS: .NET Core sources do not call `CreateCopy anywhere.

### DIFF
--- a/mcs/class/corlib/Test/System.Threading/CancellationTokenSourceTest.cs
+++ b/mcs/class/corlib/Test/System.Threading/CancellationTokenSourceTest.cs
@@ -505,12 +505,16 @@ namespace MonoTests.System.Threading
 			var tcs = new TaskCompletionSource<bool> ();
 			ct.Token.Register (() => tcs.TrySetCanceled ());
 
+			bool taskIsCancelled = false;
 			Action awaitAction = async () => {
 					try { await tcs.Task; }
-					catch (OperationCanceledException) { }
+					catch (OperationCanceledException) { 
+						taskIsCancelled = true;
+					}
 				};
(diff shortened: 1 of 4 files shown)


marek-safar:
@monojenkins backport 2018-12

marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2019-02

marek-safar:
@monojenkins backport 2018-08

marek-safar:

PR #12795, opened 2019-02-05
[runtime] Refactor pinvoke probing

lambdageek:
Split up `mono_lookup_pinvoke_call` into several smaller static functions.  There is still more that should be done (the `for`-loop and `switch` for filename probing should also factor into smaller functions) but this is a first step.

There should be no functional change.

Also mark `mono_lookup_pinvoke_call` as external-only and switch the runtime to using `mono_lookup_pinvoke_call_internal`.  (This will make it easier to switch to a `MonoError` for errors, instead of returning a pair of strings).

### DIFF
--- a/mono/metadata/loader.c
+++ b/mono/metadata/loader.c
@@ -1200,6 +1200,11 @@ mono_lookup_pinvoke_call (MonoMethod *method, const char **exc_class, const char
 	return result;
 }
 
+static MonoDl*
+pinvoke_probe_for_module (MonoImage *image, const char*new_scope, const char *import, char **found_name_out, char **error_msg_out);
+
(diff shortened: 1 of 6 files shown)


mono/metadata/loader.c
```diff
@@ -1267,6 +1286,64 @@ mono_lookup_pinvoke_call (MonoMethod *method, const char **exc_class, const char
 			found_name = g_strdup (found_name);
 	}
 
+	if (!module) {
```
vargaz: Extra {}

lambdageek:
@monojenkins build failed

lambdageek:
@vargaz fixed.  review again?

PR #12801, opened 2019-02-05
[2018-10] Extend mono-native minimum version magic to CPPFLAGS/CXXFLAGS

akoeplinger:
We were only doing the `-mmacosx-min-version` replacement magic on CFLAGS and LDFLAGS for mono-native-compat and -unified, but we need to do the same for CPPFLAGS/CXXFLAGS too.

Fixes an issue where the libraries were targeting the wrong minimum OSX version when built from xamarin-macios since they pass the minimum version in CPPFLAGS instead of CFLAGS for OSX.


akoeplinger:
@monojenkins backport 2018-12
@monojenkins backport 2019-02
@monojenkins backport master

akoeplinger:
manually pushed to master with 04c71af56a0619bb4b617ec54b2e6ce81313de88

PR #12809, opened 2019-02-06
[WASM] debug documentation and sample updates

lewing:
Small updates to documentation and debug sample in sdks/wasm

### DIFF
--- a/sdks/wasm/README.md
+++ b/sdks/wasm/README.md
@@ -83,9 +83,9 @@ To experiment with the debugger, do the following steps:
 
 - When calling `packager.exe` pass the `-debug` argument to it.
 - Start Chrome with remote debugging enabled (IE `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome\ Canary --remote-debugging-port=9222`)
-- Run the proxy: `dotnet dbg-proxy/ProxyDriver.dll`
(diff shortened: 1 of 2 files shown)


sdks/wasm/README.md
```diff
@@ -83,9 +83,9 @@ To experiment with the debugger, do the following steps:
 
 - When calling `packager.exe` pass the `-debug` argument to it.
 - Start Chrome with remote debugging enabled (IE `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome\ Canary --remote-debugging-port=9222`)
-- Run the proxy: `dotnet dbg-proxy/ProxyDriver.dll`
+- Run the proxy: `dotnet run -p ProxyDriver/ProxyDriver.csproj`
```
kjpou1: Larry, this depends on where you are trying to run the debugger from.  If you are running from build directory then it is ProxyDriver.  Running from the package that is downloaded from the CI it is dbg-proxy https://github.com/mono/mono/blob/master/sdks/wasm/Makefile#L340-L343.  This should be changed as well.

lewing: That is a good point. Since essentially everything else in the README.md is about building and running from the source directory I think it makes sense that the debugger instructions work that way as well but I completely agree that we should document how to work with CI builds and make them as similar as practical.

lewing:
@monojenkins build failed

lewing:
@monojenkins build failed

Issue #12085, opened 2018-12-17
how to build mono runtime with interpreter mode on windows?


srxqds:
how to build mono runtime with interpreter mode ?
with "--enable-minial=jit" but the output also show:
"Engine: building and using JIT"

And it should rebuild bcl library (mscorlib.dll  or mcs.exe) when it is in interpreter mode or jit mode?
If it should rebuild and how to setting build with interpreter mode?

platform: windows
host target:x86-64

srxqds:
I call mono_jit_set_aot_mode (MONO_AOT_MODE_INTERP);
It will crash...

srxqds:
I have confused with blow three questions:

1.as the website: https://www.mono-project.com/docs/about-mono/supported-platforms/ show:
![image](https://user-images.githubusercontent.com/5187024/50120048-4c962d00-028f-11e9-805f-fa4bc772412d.png)
windows don't support interpreter mode?

2.if windows support interpreter mode how should i build mono runtime and bcl should rebuild ( it differ from jit ?)

srxqds:
I use this command build mono with cywin64:
./autogen.sh --prefix=$PREFIX --host=x86_64-w64-mingw32 --enable-msvc --disable-boehm --disable-jit --enable-minimal=jit

I found in cygconfig.h has:

![image](https://user-images.githubusercontent.com/5187024/50143333-4cc21700-02e7-11e9-9c2d-3870cfb869d1.png)

![image](https://user-images.githubusercontent.com/5187024/50143341-5186cb00-02e7-11e9-833a-b00add850c81.png)

but when I call make command, config.h will replace by winconfig.h so that the "--disable-jit --enable-minimal=jit" args not work?

how can I do it?

I want implement that: use all platfom (windows, android, ios) runing with interpreter mode

I call mono_jit_set_aot_mode (MONO_AOT_MODE_INTERP); it will crash!

can someone help me?

lateralusX:
No work has been done to port or test interpreter to Windows (at least not that I'm aware of). It's on the list of things to do.

The mingw gcc build still uses the flags used in configure step (replaced config.h includes cygconfig.h that is the original config file), so that build should at least build with expected flags, but since interpreter has not been validated or tested, there might be issues actually running it. For MSVC build, additional work needs to be done in order to preserve the flags. If you just want to test the MSVC build, you can manually add the needed flags directly into winconfig.h and make sure they end up in final config.h

lateralusX:
[...]

PR #12810, opened 2019-02-06
Fix Windows x64 full AOT runtime test failures.

lateralusX:
Set of fixes and changes needed to get pass on runtime tests under Windows x64 full AOT CI test suite. The PR includes a couple of fixes related to specific failures, but it also change how we handle winconfig.h for our MSVC build since it needs to pick up disabled defines from our configure step, if one has been run (still an optional step, since you can build without cygwin).

Our process of constructing a config.h from our winconfig.h template is getting more complex, so I move the logic from winsetup.bat into a msbuild step with a custom task, making it easier to write the code needed for this process to work + better integration into existing msbuild targets. The winsetup.bat is still kept (sine we have clients using it), but it will only run the msbuild target, so we only have one place where all the logic creating a config.h for MSVC build is located.

Since our winconfig.h was old and outdated, I also walked through all the defines in winconfig.h and compared them against corresponding config.h generated for mingw configuration and the reversed. I also dropped all things not defined in winconfig.h to make the template easier to maintain going forward.


lateralusX:
The preserve of flags from config.h into winconfig.h (the DISABLE_* defines) should also resolve parts of the issue related to this issue, https://github.com/mono/mono/issues/12085.

luhenry:
@monojenkins build OS X x65 iOS SDK

luhenry:

PR #12827, opened 2019-02-06
[wasm][debugger] Bug on attributes with value type string and content NULL

thaystg:
An attribute with a content that was a string null was creating two items in result json and it was throwing an exception in client side because it is expected a name and a value and not a name and two values.

### DIFF
--- a/mono/mini/mini-wasm-debugger.c
+++ b/mono/mini/mini-wasm-debugger.c
@@ -618,9 +618,9 @@ static gboolean describe_value(MonoType * type, gpointer addr)
 			break;
 		case MONO_TYPE_STRING: {
 			MonoString *str_obj = *(MonoString **)addr;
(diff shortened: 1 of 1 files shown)


mono/mini/mini-wasm-debugger.c
```diff
@@ -620,11 +620,12 @@ static gboolean describe_value(MonoType * type, gpointer addr)
 			MonoString *str_obj = *(MonoString **)addr;
 			if (!str_obj)
 				mono_wasm_add_string_var (NULL);
```
vargaz: Both sides of the if-then should have a {} if one of them has it.

thaystg: Done! :)

thaystg:
@monojenkins build failed

PR #12829, opened 2019-02-06
Grow StackSlotInfo array rather than asserting

joncham:
Fixes Unity issue: https://issuetracker.unity3d.com/issues/allocating-too-many-variables-in-a-method-leads-to-a-crash


vargaz:
@monojenkins build failed

PR #12830, opened 2019-02-06
[runtime] Use temp folders for merp tests

alexanderkyte:
The 2018-10 lane seemed to be failing some tests because the merp-json-valid.exe test left behind the crash log file. I fixed that in this PR, but also move the tests into their separate folders so they can run at the same time safely.

### DIFF
--- a/mono/tests/merp-json-valid.cs
+++ b/mono/tests/merp-json-valid.cs
@@ -19,7 +19,7 @@ class C
 	}
 
 	public static void 
-	JsonValidateMerp ()
+	JsonValidateMerp (string configDir)
 	{
 		var monoType = Type.GetType ("Mono.Runtime", false);
 		var m = monoType.GetMethod("EnableMicrosoftTelemetry", BindingFlags.NonPublic | BindingFlags.Static);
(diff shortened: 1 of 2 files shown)


mono/tests/merp-json-valid.cs
```diff
@@ -103,6 +100,13 @@ public static void
 	public static void Main ()
 	{
 		JsonValidateState ();
-		JsonValidateMerp ();
+
+		var configDir = "./merp-json-valid/";
+		Directory.CreateDirectory (configDir);
```
lambdageek: Probably doesn't matter yet, but I think in XA the tests can't write to the working directory so some day this may be a problem for them.

alexanderkyte: XA won't be using this set of tests in all likelihood, as they're focusing on the merp pipeline working correctly.

That is a separate issue that may need to be fixed for the crash reporter in general though. The default `mono_crash.*.json` files made by the crash reporter are put in the working directory.

alexanderkyte: Device crash reporting will probably need a per-platform modification, or a native callback of some sort.

alexanderkyte:
The windows lane here has failed due to:

```
MonoTests.runtime.block_guard_restore_aligment_on_exit.exe_timedout
```

alexanderkyte:
@monojenkins backport 2018-12
@monojenkins backport 2018-10

PR #12833, opened 2019-02-06
Update roslyn-binaries submodule to point to latest master revision

kg:
This updates our roslyn submodule so that the new csc-dim is available.


kg:
csc crashes without them. I don't know why. Do you want me to investigate? I have no idea where to start

akoeplinger:
Yes we need to find out why. We build these assemblies in mono so we can't just install those other random dlls.

kg:
We're not installing them, only using them if csc-dim is being used to build. The whole bcl builds fine this way without impacting anything else

akoeplinger:
I see, didn't realize this is only updating the DIM version. I assume we'll hit the same issue though once we bump our normal roslyn to a later version so we still need to investigate.

(I'm fine with merging this if we need it for something as long as we do the investigation).

marek-safar:
@akoeplinger we don't build System.Memory and we build System.Runtime.CompilerServices.Unsafe.dll only for runtime testing  and don't install it because we don't track the latest changes in this assembly

akoeplinger:
@marek-safar we do build System.Memory but looks like only for mobile, I got confused by that...

I'm still a bit worried (once this hits the real csc) that we're installing these two essentially random assemblies into `<prefix>/lib/mono/4.5/` next to csc.exe and who knows what that will affect.

marek-safar:
Yes but I don't think it should break things. We have been installing System.Collections.Immutable.dll in same location for long time with no issue reported

kg:
@monojenkins build failed

kg:

kg:
@monojenkins build failed

PR #12834, opened 2019-02-06
[ios] pass --enable-monotouch to cross compilers so icall defintions are visible to it

lewurm:
This fixes a regression introduced by https://github.com/mono/mono/commit/5187b0bdaaff56936d697b0753dad51e7498991d

The AOT compiler wasn't aware of the icalls since they are guarded by `ENABLE_MONOTOUCH`. Passing `--enable-monotouch` to configure of the AOT compiler enables that define, so it can properly resolve those icalls. However, to avoid `zlib` dependency for the cross compilers, we stub
them out.

Needs backport up to 2018-10.


akoeplinger:
@monojenkins backport 2018-10
@monojenkins backport 2018-12
@monojenkins backport 2019-02

vargaz:
Its also ok to add a zlib dependency imho.

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #12840, opened 2019-02-07
[wasm] Fix icall table linking, generate one table per assembly.

vargaz:


kjpou1:
@monojenkins build failed

PR #12850, opened 2019-02-07
[runtime] Refactor pinvoke probing, part 2

lambdageek:
1. Use `MonoError` to signal pinvoke probing failures. This is a bit complicated because:
   a. the old API uses another error mechanism
   b. the marshalling code needs to emit IL that will raise an exception given a `MonoError`.
2. Refactor the pinvoke library path transformation code a bit.  It still needs work, but at least it's now hopefully a little clearer that it's just doing some name mangling every time through the loop.

3. There's a `ENABLE_NETCORE` leak that I didn't fix - see the comment in the code.  Ultimately, I think I'll need to make dllmap lookups return copies of results by default, or else copy them at the use site, and then free at the use site.

### DIFF
--- a/mono/metadata/loader.c
+++ b/mono/metadata/loader.c
@@ -1348,6 +1348,8 @@ lookup_pinvoke_call_impl (MonoMethod *method, MonoLookupPInvokeStatus *status_ou
 	// FIXME: The dllmap remaps System.Native to mono-native
 	mono_dllmap_lookup (image, orig_scope, import, &new_scope, &import);
 #else
+	/* AK: FIXME: dllmap, above doesn't strdup the results, so these leak
+	 * since there's no free() */
 	new_scope = g_strdup (orig_scope);
 	import = g_strdup (import);
 #endif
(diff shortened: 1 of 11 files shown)


mono/metadata/loader.c
```diff
@@ -1414,13 +1479,55 @@ pinvoke_probe_for_module (MonoImage *image, const char*new_scope, const char *im
...
 		}
-		
+		changed = TRUE;
 		if (is_absolute) {
 			if (!dir_name)
 				dir_name = g_path_get_dirname (file_name);
 			if (!base_name)
 				base_name = g_path_get_basename (file_name);
 		}
+	} while (0);
```
lambdageek: There's a complicated bit here in an attempt to minimize line damage.
```
  do { ... 
    switch { cases ... } ;
    ... success code ...
  } while (0);
```

This is because the old code used `break` and `continue`.  `continue` means - nothing changed in this phase, so try a different name transformation.  `break` means - I modified the path in some way and we should probe there.

So since we don't have a `for` loop in this function anymore, we do a `do { ... } while (0)`.  The `break` breaks out of the `switch` and sets `changed = TRUE`, whereas the `continue` tries to run another loop iteration (which doesn't work because `(0)`) and bypasses setting `changed = TRUE`.  So after the loop `changed == TRUE` only in the cases where we hit a `break`, and not in the cases where we hit a `continue`.

Obviously we can make this cleaner, but at the cost of line damage.  Maybe worth it?

vargaz: There is already a lot of line damage, so just for go it.

lambdageek:
Failures look related. investigating...

mono/metadata/loader.c
```diff
@@ -1187,6 +1187,75 @@ is_absolute_path (const char *path)
...
+ * detail for backward compatability.
+ */
+typedef struct MonoLookupPInvokeStatus {
+	MonoLookupPInvokeErr err_code;
+	char *err_arg;
+} MonoLookupPInvokeStatus;
+
+static gpointer
+lookup_pinvoke_call_impl (MonoMethod *method, MonoLookupPInvokeStatus *status_out);;
+
```
vargaz: Extra ;

mono/metadata/loader.c
```diff
@@ -1187,6 +1187,75 @@ is_absolute_path (const char *path)
 	return g_path_is_absolute (path);
 }
 
+typedef enum MonoLookupPInvokeErr {
+	LOOKUP_PINVOKE_ERR_OK = 0, /* No error */
```
vargaz: This could just be typedef enum {.
Also, internal structures/functions can have a Mono prefix, but its not required.

lambdageek:
@lewurm or @BrzVlad review please - there's an interp change in here for properly setting up a child frame when callvirt calls this thing:
```csharp
	public class T2 : T {
		public override extern object MyClone ();
	}
```

mono/metadata/loader.c
lambdageek: ok. made it use a normal switch.

mono/mini/interp/interp.c
```diff
@@ -496,6 +496,10 @@ get_virtual_method (InterpMethod *imethod, MonoObject *obj)
 		mono_error_cleanup (error); /* FIXME: don't swallow the error */
 	}
 
+	if (virtual_method->iflags & METHOD_IMPL_ATTRIBUTE_INTERNAL_CALL) {
+		virtual_method = mono_marshal_get_native_wrapper (virtual_method, TRUE, FALSE);
+	}
+
```
lewurm: ufff, good catch 👍

mono/metadata/marshal.c
```diff
@@ -345,13 +345,11 @@ mono_delegate_to_ftnptr_impl (MonoDelegateHandle delegate, MonoError *error)
 	}
 
 	if (method->flags & METHOD_ATTRIBUTE_PINVOKE_IMPL) {
-		const char *exc_class, *exc_arg;
 		gpointer ftnptr;
 
-		ftnptr = mono_lookup_pinvoke_call_internal (method, &exc_class, &exc_arg);
+		ftnptr = mono_lookup_pinvoke_call_internal (method, error);
 		if (!ftnptr) {
```
luhenry: Could we just use `if (!is_ok(error))`, or doesn't `mono_lookup_pinvoke_call_internal` guarantee that it will only return NULL in case of error?

lambdageek: At the present, it doesn't have that guarantee.

PR #12851, opened 2019-02-07
Incorrect call to MONO_HANDLE_SETRAW causes random crash in sgen.

lateralusX:
mono_string_utf16_to_builder_copy incorrectly uses MONO_HANDLE_SETRAW to update the chunkLength field with a string length. Since MONO_HANDLE_SETRAW assumes the argument is a MonoObject reference this could lead to a dereference and since the value is a string length, that dereference will lead to a crash.

The problem happens in mono_gc_wbarrier_set_arrayref_internal where vtable is retrieved from value. The crash happens randomly since it is depending on the object being moved out of the nursery + sgen_binary_protocol_wbarrier not being eliminated.

Fix is to use MONO_HANDLE_SETVAL since the value is not a reference so no need to trigger a write barrier.


luhenry:
@monojenkins build failed

PR #12853, opened 2019-02-07
Fix OP_GET_LAST_ERROR codegen on Windows x64 full AOT.

lateralusX:
Current codegen incorrectly read back last error when doing p/invoke for full AOT. This caused many failed corlib tests ~600. It also turns out that the test for set/get last error related to p/invoke that we
have in pinvoke2.exe, is not run for the full AOT profile, due to other tests depending on none full AOT features. In order to make sure we run the test and not regress, test has been moved into its own runtime tests, running as part of full AOT testing.


mono/mini/mini-amd64.c
```diff
@@ -3833,14 +3833,16 @@ emit_setup_lmf (MonoCompile *cfg, guint8 *code, gint32 lmf_offset, int cfa_offse
...
+#define TEB_LAST_ERROR_SLOT 0x30
 #define TEB_LAST_ERROR_OFFSET 0x068
 
 static guint8*
 emit_get_last_error (guint8* code, int dreg)
 {
 	/* Threads last error value is located in TEB_LAST_ERROR_OFFSET. */
 	x86_prefix (code, X86_GS_PREFIX);
-	amd64_mov_reg_membase (code, dreg, TEB_LAST_ERROR_OFFSET, 0, sizeof (guint32));
+	amd64_mov_reg_mem (code, dreg, TEB_LAST_ERROR_SLOT, sizeof (gpointer));
```
jaykrell: Why this change? I see it matches GetLastError, but it looked correct and more optimal before.

PR #12856, opened 2019-02-07
[Registry] Prevent crash due to background thread facing problems in the file system

### LINKED ISSUE #12863: KeyHandler.DirtyTimeout can cause uncatchable exceptions
We are seeing failures in VS for Mac related to the registry key implementation in Mono.  This failure accounts for 35% of known crashes in VS for Mac 8.0 Preview 1 (so a very high priority to get a fix here).

The underlying issue is that the registry KeyHandler creates a 3 second timer to ensure we've written the registry key to disk: https://github.com/mono/mono/blob/master/mcs/class/corlib/Microsoft.Win32/UnixRegistryApi.cs#L687

If this timer is triggered and the user has not saved the registry key, KeyHandler will attempt to save this registry key on a thread pool worker thread.  If we hit any issue writing to that file (say another process is attempting to write at the same time), the mono runtime will throw an uncatchable exception on this threadpool thread, tearing down VS for Mac as a result.

Note that we don't control every piece of code in VS for Mac, so we can't just go through and be sure to manually Dispose every RegistryKey object.  (Though we will try to search the codebase and make fixes that we can find.)

There may not be a ton of good options here, but I think the fundamental issue is that we shouldn't tear down VS for Mac with an uncatchable exception when we cannot save a registry key.

## Steps to Reproduce

This is difficult.  Open a registry key, write to it, do not dispose of the key, and be sure the underlying file that mono uses to back the registry is in some way locked or unwritable.  This should trigger the scenario outlined.

### Current Behavior

We throw an IOException (or similar) on a threadpool thread, which we don't have a good way to handle.

### Expected Behavior

We (somehow) do not throw this exception.  I'm not sure if silently swallowing the exception is the right behavior (meaning we didn't actually update the registry), or infinitely retrying (could create perf issues), or some other resolution I haven't thought of.

[ X ] macOS
[ X ] Linux
[ ] Windows

Observed on macOS, pretty sure Linux has the same issue.

**Version Used**:

5.16.0.221 (2018-06/b63e5378e38)

## Stacktrace

```
mscorlib.dll!FileStream..ctor+259
mscorlib.dll!FileStream..ctor+0
mscorlib.dll!0+ffffffff (UNKNOWN METHOD TOKEN)
mscorlib.dll!File.Create+0
mscorlib.dll!File.Create+0
mscorlib.dll!KeyHandler.Save+214
mscorlib.dll!KeyHandler.Flush+1d
mscorlib.dll!KeyHandler.DirtyTimeout+0
mscorlib.dll!60034e2+7 (UNKNOWN METHOD TOKEN)
mscorlib.dll!QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem+15
mscorlib.dll!ThreadPoolWorkQueue.Dispatch+74
mscorlib.dll!_ThreadPoolWaitCallback.PerformWaitCallback+0

```

migueldeicaza:
This should fix the crash spotted by MonoDevelop/Visual Studio for Mac that happens when a plugin uses the Win32 Registry. 

Fixes #12863

### DIFF
--- a/mcs/class/corlib/Microsoft.Win32/UnixRegistryApi.cs
+++ b/mcs/class/corlib/Microsoft.Win32/UnixRegistryApi.cs
@@ -690,7 +690,12 @@ namespace Microsoft.Win32 {
 
 		public void DirtyTimeout (object state)
 		{
-			Flush ();
+			try {
+				Flush ();
+			} catch {
+				// This was identified as a crasher under some scenarios
+				// Internal MS issue: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/787119
+			}
 		}
 
 		public void Flush ()


marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2018-08

marek-safar:
@monojenkins backport 2018-12

mcs/class/corlib/Microsoft.Win32/UnixRegistryApi.cs
```diff
@@ -690,7 +690,12 @@ void SetDirty ()
 
 		public void DirtyTimeout (object state)
 		{
-			Flush ();
+			try {
+				Flush ();
```
leculver: We _may_ need a try/catch on the finalizer as well.  From reading the code, I actually don't think it's possible for the finalizer to run (since the object is rooted by the DirtyTimeout callback).  I don't think it would hurt to wrap `~KeyHandler` as well since we would have the same crashing issue there too if there's a race somewhere I didn't consider.

leculver: To be clear though:  We only see failures from VS for Mac in the DirtyTimeout callstack.  I've never seen it from the finalizer, so I'm 100% good with this change.  Thanks!

PR #12868, opened 2019-02-07
[2018-10] [corlib] Disable four tests in TypeTest

akoeplinger:
TypeTest.MakeGenericType_BadUserType
TypeTest.MakeGenericType_NestedUserDefinedType
TypeTest.MakeGenericType_UserDefinedType
TypeTest.TestMakeGenericType_UserDefinedType_DotNet20SP1

These tests pass when run with the interpreter on xamarin-macios.
They were disabled on master with https://github.com/mono/mono/commit/dc9bc9e0e6462739872117afc59e46fbc780d981

### DIFF
--- a/mcs/class/corlib/Test/System/TypeTest.cs
+++ b/mcs/class/corlib/Test/System/TypeTest.cs
@@ -3295,6 +3295,7 @@ namespace MonoTests.System
 			Assert.AreEqual ("System.Int32", t.FullName);
 		}
 
+/*
 		[Test]
 #if MONOTOUCH || FULL_AOT_RUNTIME
 		[ExpectedException (typeof (NotSupportedException))]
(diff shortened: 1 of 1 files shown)


akoeplinger:
@monojenkins backport 2018-12

VincentDondain:
@akoeplinger @marek-safar could we possibly back port that to 2018-08 too. This is what Xamarin.iOS d16-1 is using and we're trying to greenify it (:

lambdageek:

lambdageek:
I'm confused why these tests are commented out (on master): if they are expected to pass on all platforms now, shouldn't we run them?

akoeplinger:
@VincentDondain the backport PR to 2018-08 was merged.

@lambdageek I think the issue was that we currently have no good way to tell at runtime whether we have the interpreter so we'd hit a test failure in both cases (with interpreter we don't get the expected exception and if we remove that then the "normal" case fails).

Issue #12747, opened 2019-02-02
System.Reflection.Emit.ParameterBuilder.SetCustomAttribute() throws undocumented ArgumentOutOfRangeException


MCpiroman:
## Steps to Reproduce
Compile and run following program:
```
using System;
using System.Collections.Generic;
using System.Runtime.InteropServices;
using System.Reflection;
using System.Reflection.Emit;

class Program
{
    [DllImport("SomeLib")]
    private static extern void SomeMethod([MarshalAs(UnmanagedType.LPWStr)] string param);

    static void Main(string[] args)
    {
        var aName = new AssemblyName("TestAssembly");
        var assembly = AppDomain.CurrentDomain.DefineDynamicAssembly(aName, AssemblyBuilderAccess.RunAndSave);
        var module = assembly.DefineDynamicModule(aName.Name, aName.Name + ".dll");

        var someMethod = typeof(Program).GetMethod(nameof(SomeMethod), BindingFlags.Static | BindingFlags.NonPublic);

        var typeBuilder = module.DefineType("NewType" + module.ToString(), TypeAttributes.Class | TypeAttributes.Public);
        var methodBuilder = typeBuilder.DefineMethod("NewMethod", MethodAttributes.Public | MethodAttributes.HideBySig, typeof(void), new[] { typeof(string) });
        var il = methodBuilder.GetILGenerator();
        il.Emit(OpCodes.Ret);

        var param = someMethod.GetParameters()[0];
        var paramBuilder = methodBuilder.DefineParameter(1, param.Attributes, null);
        var attr = param.GetCustomAttribute<MarshalAsAttribute>();
        var attrCtor = typeof(MarshalAsAttribute).GetConstructor(new[] { typeof(UnmanagedType) });
        object[] attrCtorArgs = { attr.Value };
        var fieldArguments = new List<FieldInfo>();
        var fieldArgumentValues = new List<object>();
        foreach (var field in typeof(MarshalAsAttribute).GetFields(BindingFlags.Public | BindingFlags.Instance))
        {
            fieldArguments.Add(field);
            fieldArgumentValues.Add(field.GetValue(attr));
        }

        var attrBuilder = new CustomAttributeBuilder(attrCtor, attrCtorArgs, Array.Empty<PropertyInfo>(), Array.Empty<object>(),
            fieldArguments.ToArray(), fieldArgumentValues.ToArray());
        paramBuilder.SetCustomAttribute(attrBuilder);

[...]

PR #12872, opened 2019-02-07
[sre] Handle null values in MarshalAsAttribute CustomAttributeBuilder

lambdageek:
Normally strings are encoded as length-then-utf8 in ECMA-335.  But a null
String is encoded as just the byte 0xFF.

When we make a CustomAttributeBuilder for a MarshalAsAttribute and attach it to
some parameter we first encode the CAB as a byte blob and then decode it in
CAB.get_umarshal into a UnmanagedMarshal attribute.  But the decoding didn't
account for the possibility of null strings.

Fixes https://github.com/mono/mono/issues/12747


lambdageek:

lambdageek:

lambdageek:

PR #12882, opened 2019-02-08
[wasm] Add mixed tests to CI.

vargaz:


vargaz:
@monojenkins build failed

PR #12883, opened 2019-02-08
[llvmonly] Add a mini_llvmonly_throw_nullref_exception () jit icall to throw nullref exceptions, it occurs very frequently when explicit null checks are enabled.

vargaz:


lewurm:
What's the advantage of doing that way? Smaller code size?

vargaz:
Yes

vargaz:
@monojenkins build failed

PR #12894, opened 2019-02-08
[arm/bitcode] mark valuetype returns as such when dealing with pinvokes

lewurm:
On armv7 our calling convention happens to be the same as the platform
one. However, when using a armv7 cross compiler to generate bitcode that
targets a different platform (e.g. arm64_32) the calling convention can
be different (e.g. on arm64_32, a return buffer must be passed via r8).
Thus, mark it properly for LLVM so it can generate correct code.

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -2327,9 +2327,13 @@ mono_arch_get_llvm_call_info (MonoCompile *cfg, MonoMethodSignature *sig)
 	case RegTypeIRegPair:
 		break;
 	case RegTypeStructByAddr:
-		/* Vtype returned using a hidden argument */
-		linfo->ret.storage = LLVMArgVtypeRetAddr;
-		linfo->vret_arg_index = cinfo->vret_arg_index;
+		if (sig->pinvoke) {
+			linfo->ret.storage = LLVMArgVtypeByRef;
+		} else {
+			/* Vtype returned using a hidden argument */
+			linfo->ret.storage = LLVMArgVtypeRetAddr;
+			linfo->vret_arg_index = cinfo->vret_arg_index;
+		}
 		break;
 #if TARGET_WATCHOS
 	case RegTypeStructByVal:


mono/mini/mini-arm.c
```diff
@@ -2327,9 +2327,13 @@ mono_arch_get_llvm_call_info (MonoCompile *cfg, MonoMethodSignature *sig)
 	case RegTypeIRegPair:
 		break;
 	case RegTypeStructByAddr:
-		/* Vtype returned using a hidden argument */
-		linfo->ret.storage = LLVMArgVtypeRetAddr;
-		linfo->vret_arg_index = cinfo->vret_arg_index;
+		if (sig->pinvoke) {
+			linfo->ret.storage = LLVMArgVtypeByRef;
+		} else {
```
vargaz: Shouldn't this be done only for arm64_32 ?

lewurm: LLVM will translate it properly on armv7k as well

vargaz: What about normal arm linux ?

lewurm:
@monojenkins build Linux ARMv7 AOT+LLVM

lewurm:
@monojenkins build Linux ARMv7 Bitcode

mono/mini/mini-arm.c
lewurm: I didn't test that configuration, but I don't see why LLVM shouldn't be able to emit proper code for that. It will emit whatever is right for the target ABI which is not always compatible with our internal ABI, thus the case distinction.

vargaz: Sometimes it can emit it, sometimes it can't, this is the reason for all the LLVMArg cases, the front end has to know ABI details in a lot of cases.

vargaz: Unfortunately, there is no arm+llvm CI lane.

lewurm: I'll have a look

lewurm: verified on armhf linux, generated code looks the same.

luhenry:

luhenry:
@rolfbjarne do we want that for `xamarin-macios:master`?

rolfbjarne:
@luhenry I think it's fine if this is 2018-10+ only, we only need it with arm64_32, which only happens with 2018-10.

lewurm:
@monojenkins backport 2018-08-watchos

PR #12895, opened 2019-02-08
[ios] link mono-native correctly

lewurm:
Fixes this app startup issue

```
dyld: Library not loaded: /Users/lewurm/work/mono-watch4-ios-sdk/sdks/out/ios-target64-release/lib/libmono-native-unified.0.dylib
  Referenced from: /var/containers/Bundle/Application/9697C8F3-811C-4BF3-B1BC-1E6385673984/PublicStaging.app/test-Mono.Runtime.Tests
  Reason: image not found
Process 2269 exited with status = 0 (0x00000000) Terminated due to signal 6
```

### DIFF
--- a/sdks/ios/appbuilder/appbuilder.cs
+++ b/sdks/ios/appbuilder/appbuilder.cs
@@ -230,7 +230,7 @@ public class AppBuilder
 		ninja.WriteLine ("  command = clang -isysroot $sysroot -miphoneos-version-min=10.1 -arch arm64 -c -o $out $in");
 		ninja.WriteLine ("rule gen-exe");
 		ninja.WriteLine ("  command = mkdir $appdir");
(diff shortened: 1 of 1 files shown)


lewurm:
@monojenkins squash

sdks/ios/appbuilder/appbuilder.cs
```diff
@@ -230,7 +230,7 @@ public class AppBuilder
 		ninja.WriteLine ("  command = clang -isysroot $sysroot -miphoneos-version-min=10.1 -arch arm64 -c -o $out $in");
 		ninja.WriteLine ("rule gen-exe");
 		ninja.WriteLine ("  command = mkdir $appdir");
```
baulig: Looks fine to me provided that the `GSS` framework is actually available on all platforms that this tool uses (which I'm not sure about).

lewurm: iOS 64bit devices is the only target currently

PR #12921, opened 2019-02-09
Fix build and use System.Native on FreeBSD

NattyNarwhal:
* System headers sometimes need another one to support it; otherwise they break

* System.Native already builds on FreeBSD for CoreCLR and before the refactor in Mono, so nothing but build system changes. An Android definition was also fixed as a byproduct.


vargaz:
@monojenkins build failed

PR #12923, opened 2019-02-09
[netcore] Update to the latest version

marek-safar:
/cc @vargaz


vargaz:
@monojenkins build failed

PR #12932, opened 2019-02-11
[System.Private.CoreLib] Clean up EqualityComparer

marek-safar:


mcs/class/System.Private.CoreLib/System.Collections.Generic/EqualityComparer.cs
```diff
@@ -42,26 +42,16 @@ static EqualityComparer<T> CreateComparer ()
...
-        public override bool Equals(T x, T y) {
-            int x_final = System.Runtime.CompilerServices.JitHelpers.UnsafeEnumCast(x);
-            int y_final = System.Runtime.CompilerServices.JitHelpers.UnsafeEnumCast(y);
-            return x_final == y_final;
-        }
-		/*
-        public override int GetHashCode(T obj) {
-            int x_final = System.Runtime.CompilerServices.JitHelpers.UnsafeEnumCast(obj);
-            return x_final.GetHashCode();
-        }
```
vargaz: This is going to be needed, the version in shared/ does boxing.

marek-safar: Why? It calls it using constrained. !T

PR #12952, opened 2019-02-12
[Wasm] Proxy More of the DevTools protocol

lewing:
Much of the tooling around DevTools expects
actions under /json to return useful data
this adds adds some simple logic to proxy those
so that you can point tooling directly at the
proxied port and get responses you expect


kjpou1:
@monojenkins build failed

PR #12953, opened 2019-02-12
[Wasm] debug local source

lewing:
If we can locate the source locally use direct file: references because some of the Visual Studio tooling works much better in this case.


kjpou1:
@monojenkins build failed

Issue #8486, opened 2018-04-27
struct with 3 bytes not marshaled correctly with LLVM on watchOS


rolfbjarne:
## Steps to Reproduce

1. Download test case: [struct-bbb-bug.zip](https://github.com/mono/mono/files/1954877/struct-bbb-bug.zip)
2. Open solution, select the watch app as the startup project.
3. Select a Release|Device configuration.
3. Run.

### Current Behavior

Application output:

```
[...]
Assertion failed: (s.x == 1), function set_s, file set_s.c, line 10.
0 -70 32
Xamarin.Hosting: Process '571' exited.
```

The code (in InterfaceController.cs) calls a P/Invoke that takes a struct with 3 bytes, managed code passes a struct with the values [1;2;3], native code receives [0;-72;32].

Other failing tests indicate that structs with 3 shorts and structs with 2 bytes have the same problem (I didn't bother creating separate test cases, since it looks like the same issue). Other variations work fine (structs with 1 or 4 bytes, structs with 1, 2 or 4 shorts, structs with 1, 2, 3 or 4 ints, longs, floats or doubles).

### Expected Behavior

If running the debug configuration:

```
[...]
1 2 3
DONE
```

## On which platforms did you notice this

watchOS with LLVM.

**Version Used**:

xamarin-macios/master (xamarin/xamarin-macios@bf159969f9b5c72abe3d6959cf8c4840495b3f2b) which is mono's 2018-02 branch.

It does not look like a regression (I can reproduce with d15-7, d15-6 and d15-5. I didn't try with earlier versions).

lewurm:
that looks similar to https://github.com/mono/mono/issues/7449 which got fixed by https://github.com/mono/mono/pull/7691

maybe we have a similar issue with the llvm backend

marek-safar:
@vargaz could you confirm that?

marek-safar:
@vargaz what's the status of this issue?

rolfbjarne:
Updated test case that tests for more signatures (structs with 1/2/3 bytes/shorts): 
[struct-bbb-bug-v2.zip](https://github.com/mono/mono/files/2303018/struct-bbb-bug-v2.zip)

Result:

    S1: expected <1>, got <52736>
    S2: success
    S3: expected <1, 2, 3>, got <1, 2, 1>
    B1: expected <1>, got <0>
    B2: expected <1, 2>, got <0, -50>
    B3: expected <1, 2, 3>, got <0, -50, 38>
[...]

PR #12992, opened 2019-02-13
[bitcode] round up value type size for slot calculation

lewurm:
consider

```csharp
struct S3
{
    public ushort X;
    public ushort Y;
    public ushort Z;
}
```
this would be 3 * 2 = 6 bytes. Structs, however, are passed around as i32
arrays in LLVM IR between functions. Those, if such a struct is passed, are
split to `int32`s depending on the struct's size. The previous calulation was
wrongly reserving one such i32 slot instead of two for `S3` above.

Additionally increase slot size to i64 so that it aligns with a
slightly different ABI on arm64_32.

Fixes https://github.com/mono/mono/issues/8486

@rolfbjarne I've tested it on a arm64_32 device, please confirm that this fixes the issue on your armv7k device as well

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -2367,17 +2367,22 @@ mono_arch_get_llvm_call_info (MonoCompile *cfg, MonoMethodSignature *sig)
 		case RegTypeFP:
 			lainfo->storage = LLVMArgNormal;
 			break;
-		case RegTypeStructByVal:
+		case RegTypeStructByVal: {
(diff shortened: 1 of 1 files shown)


lewurm:

lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2018-12
@monojenkins backport 2018-10

@rolfbjarne do we need that in `2018-08` as well?

rolfbjarne:
@lewurm no, this bug has been in stable for a while now, and nobody's run into it, so I don't think we need to get it into 2018-08.

PR #12994, opened 2019-02-14
[System.Net.Http] Improve HttpClientTest to use the retry http listener

akoeplinger:
It was missed in https://github.com/mono/mono/pull/11829 so we were still manually creating HttpListeners which are not hardened against "Address already in use".

Should fix/improve https://github.com/xamarin/maccore/issues/1407.


akoeplinger:
@monojenkins backport 2019-02

PR #12999, opened 2019-02-14
Pass X509Certificate2 to the certificate validator.

baulig:
This makes us consistent with corefx; see also https://github.com/dotnet/corefx/pull/35306.

Using `X509Certificate` has become a problem because the `HttpClientHandler.ServerCertificateCustomValidationCallback` callback takes an `X509Certificate2` argument.


baulig:
@monojenkins backport 2019-02

PR #13007, opened 2019-02-14
[tests] Add the BitcodeNotSupported category to tests exercising non-bitcode-compatible API.

rolfbjarne:
We have APIs that use exception filter clauses, and we don't support that when
compiling to bitcode.

So mark any tests exercising such API accordingly, so that we can skip those
tests when running tests on targets that use bitcode (watchOS).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1221,6 +1221,7 @@ elif test x$with_runtime_preset = xbitcode; then
    enable_cooperative_suspend_default=yes
    # FIXME: use TEST_PROFILE=testing_aot_full
    TEST_PROFILE=testing_aot_bitcode
+   PLATFORM_TEST_HARNESS_EXCLUDES=BitcodeNotSupported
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -1239,6 +1240,7 @@ elif test x$with_runtime_preset = xbitcodeinterp; then
    enable_cooperative_suspend_default=yes
    # FIXME: use TEST_PROFILE=testing_aot_full_interp
    TEST_PROFILE=testing_aot_bitcode_interp
+   PLATFORM_TEST_HARNESS_EXCLUDES=BitcodeNotSupported
```
luhenry: @rolfbjarne as discussed on Slack, please remove that change as the interpreter supports exception filters so the tests will work in that configuration.

lewurm: I added that 🙂  follow-up discussion was that the BCL code contains exception filters.

PR #13019, opened 2019-02-14
[monodroid] Forward the Mono.Android.dll System.Drawing types from Sy…

alexischr:
…stem.Drawing.dll too, as that's where Xamarin.Forms.Core.dll (as used by Xamarin Android) expects them to be.


alexischr:
This is probably wrong, guided by `monodis` resolving a type to `System.Drawing` assembly instead of `netstandard`

PR #13021, opened 2019-02-14
[crash] Move dump_memory_around_ip after the native crash

alexanderkyte:
If someone jumps to a bogus address for some reason, the dump_memory_around_ip function will double-fault. On OSX, a double fault is often a hang. 

Moving this after the mission-critical reporting is helpful here.

### DIFF
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -1154,9 +1154,9 @@ mono_dump_native_crash_info (const char *signal, void *ctx, MONO_SIG_HANDLER_INF
 {
 	print_process_map ();
 
-	dump_memory_around_ip (ctx);
-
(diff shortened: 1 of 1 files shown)


baulig:
Stumbled upon this by accident when playing around with some stuff.

One way to reproduce this is to edit `mono/btls/btls-ssl-ctx.c` then replace
```SSL_CTX_set_cert_verify_callback (ptr->ctx, cert_verify_callback, ptr);```
with
```SSL_CTX_set_cert_verify_callback (ptr->ctx, 0xdeadbeaf, ptr);```

Then `MONO_TLS_PROVIDER=btls make -C mcs/class/System.Net.Http run-xunit-test` should trigger it.

alexanderkyte:
@monojenkins build failed

PR #12965, opened 2019-02-12
[2018-10] [monodroid] Fill-in reference stubs for System.Drawing types that are …

alexischr:
…currently 'forked' from Mono into Mono.Android.dll

The list of localy implemented System.Drawing types: https://github.com/xamarin/xamarin-android/blob/e0da1f152de84c4cc39983f2fda775ec98c08250/src/Mono.Android/Mono.Android.csproj#L311-L325

PR #13023, opened 2019-02-14
Bump reference assemblies to include Mono.Android update

alexischr:


alexischr:
Folded into https://github.com/mono/mono/pull/12965

PR #13033, opened 2019-02-15
[wasm][debugger] Evaluate property without getter implemented

### LINKED ISSUE #12964: [debugger][wasm] Assertion at mono/metadata/loader.c:2500, condition '<disabled>' not met
## Steps to Reproduce
This is the sample I'm using: 
- https://github.com/nventive/Uno.Xaml-Controls-Gallery
- You can change the runtime used there: https://github.com/nventive/Uno.Xaml-Controls-Gallery/blob/1c19aa424e0aa507cdf0d99d1fa308ce52d32a09/XamlControlsGallery.Wasm/XamlControlsGallery.Wasm.csproj#L13 it's the urls out of the Jenkins CI
- The extracted mono path will be in /tmp/mono-xxxx, it's not overwritten on each build
- The used debugger proxy is here `obj/Debug/netstandard2.0/wasm-debugger/mono-wasm-xxx` it's written only once if the folder is not present, you can change the `Mono.WebAssembly.DebuggerProxy.dll` at will.
- Launching the debugger is similar to blazor, with `Shift+Alt+D` with the instructions to follow that you're most certainly familiar with :)
- After that, for this specific runtime issue, put a breakpoint https://github.com/nventive/Uno/blob/d10ec0242b0343cf9ca8e35da9917b773d091144/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs#L1189, I'm clicking on items on the navigation menu, break and expand `instance` or `prevInstance`

And to build the sample, `msbuild /r` in the `XamlControlsGallery.Wasm` folder is enough, the resulting website is in `bin/Debug/netstandard2.0/dist`

And if you run under linux, you'll need to remove this otherwise it'll produce an AOT build

[ ] macOS
[ ] Linux
[X] Windows

thaystg:
If getter is not implemented in a property we don't want to show this property, in Vs the behaviour is the same.

Fixes #12964

### DIFF
--- a/mono/mini/mini-wasm-debugger.c
+++ b/mono/mini/mini-wasm-debugger.c
@@ -702,17 +702,19 @@ describe_object_properties (guint64 objectId)
 	iter = NULL;
 	while ((p = mono_class_get_properties (obj->vtable->klass, &iter))) {
 		DEBUG_PRINTF (2, "mono_class_get_properties - %s - %s\n", p->name, p->get->name);
-		mono_wasm_add_properties_var(p->name);
-		sig = mono_method_signature_internal (p->get);
-		res = mono_runtime_try_invoke (p->get, obj, NULL, &exc, error);
-		if (!mono_error_ok (error) && exc == NULL)
-			exc = (MonoObject*) mono_error_convert_to_exception (error);
-		if (exc)
-			describe_value (mono_get_object_type (), &exc);
-		else if (!m_class_is_valuetype (mono_object_class (res)))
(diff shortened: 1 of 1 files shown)


thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

luhenry:
@thaystg would there be a simple test you could add to cover this case?

thaystg:
@luhenry, yes, I did a simple test and saved it to put into wasm debugger tests. :)

PR #13044, opened 2019-02-16
Fix some compiler warnings

nsivov:
Compiled with gcc 8.2.0/x86_64-linux-gnu.

### DIFF
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -97,10 +97,12 @@ assert_not_reached_fn_ptr_calloc (gsize n, gsize x)
 	assert_not_reached_mem ("Attempted to call calloc during merp dump");
 	return NULL;
 }
+#endif /* defined(ENABLE_CHECKED_BUILD_CRASH_REPORTING) && defined (ENABLE_OVERRIDABLE_ALLOCATORS) */
 
 void
 mono_summarize_toggle_assertions (gboolean enable)
 {
+#if defined(ENABLE_CHECKED_BUILD_CRASH_REPORTING) && defined (ENABLE_OVERRIDABLE_ALLOCATORS)
 	static GMemVTable g_mem_vtable_backup;
 	static gboolean saved;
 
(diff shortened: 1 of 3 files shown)


mono/utils/mono-state.c
```diff
@@ -116,8 +116,13 @@ mono_summarize_toggle_assertions (gboolean enable)
 	}
 
 	mono_memory_barrier ();
-#endif
 }
```
baulig: What's the reason behind this change?  I think it looks more ugly when we have the same function defined both inside and outside the conditional.

mono/utils/mono-state.c
```diff
@@ -687,6 +692,7 @@ mono_native_state_add_thread (MonoStateWriter *writer, MonoThreadSummary *thread
 static void
 mono_native_state_add_ee_info  (MonoStateWriter *writer)
 {
+#ifndef MONO_PRIVATE_CRASHES
```
baulig: Where does this conditional come from?  This particular change also seems to modify behavior, so it's more than just fixing compiler warnings.

nsivov: As far as I can tell it comes from configure, controlled with --with-crash_privacy=yes,no. What behavior does this change modify? It turns existing function to empty body, and changes to call it unconditionally, to fix unused function warning.

mono/utils/mono-state.c
nsivov: The reason is to move unused *_not_reached_* functions under conditional block. I was thinking one block is better than two for the same condition, but surely I can keep function intact if you prefer.

mono/utils/mono-state.c
lewurm: this is fine

mono/utils/mono-state.c
lewurm: I agree with @baulig here, we use that style everywhere and even you do it in `mono_native_state_add_ee_info` below 🙂

lewurm:

lewurm:
@monojenkins squash

mono/utils/mono-state.c
nsivov: Sure, I'll keep that in mind. To me the difference was about having conditional expression in one place or in two, having it consolidated seemed better.

PR #13046, opened 2019-02-17
Remove leading . from script invocation

rootwyrm:
The leading `.` in the script invocation causes the tests to fail on multiple platforms by causing the required argument to not be passed correctly resulting in bad output. Removal of the leading `.` ensures correct operation on all platforms.


akoeplinger:
@monojenkins build

akoeplinger:
@monojenkins backport 2019-02

Issue #13002, opened 2019-02-14
Environment.OSVersion should detect the OS through the browser when running on WebAssembly


danroth27:
It looks like `Environment.OSVersion` returns "Unix 1.0.0.0" when running on wasm. Perhaps it should instead reflect the value of [window.navigator.platform`](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform), or whatever is the most reliable way to detect the OS from a browser?

danroth27:
@kjpou1

kjpou1:
@danroth27 I may have jumped the gun on this one.  The [Environment.OSVersion](https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=netframework-4.7.2) is actually made up of a [PlatformID enum](https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-4.7.2).    

Even if we did put some code in to try to parse the operating system the problem with this is that the [window.navigator.platform](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform) can be spoofed.  So it will not be reliable anyway.

danroth27:
> Even if we did put some code in to try to parse the operating system the problem with this is that the [window.navigator.platform](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform) can be incorrect. So it will not be reliable anyway.

Yes, I am aware of this, but I just thought it would be more useful than always returning "Unix 1.0.0.0"

NattyNarwhal:
See these Core issues: https://github.com/dotnet/corefx/issues/23989 https://github.com/dotnet/corefx/issues/30706

kjpou1:
I totally agree and the reason why I jumped on it so fast.

akoeplinger:
There's also https://github.com/dotnet/corefx/issues/27417 which is related.

mkArtakMSFT:
This is also being discussed https://github.com/dotnet/runtime/issues/33328

mkArtakMSFT:
@lewing do you expect to align Mono implementation with the decision in the above referenced issue in dotnet/runtime?

PR #13047, opened 2019-02-18
[wasm] Detect the OS through the browser.

kjpou1:
Resolves issue: https://github.com/mono/mono/issues/13002

### DIFF
--- a/sdks/wasm/library_mono.js
+++ b/sdks/wasm/library_mono.js
@@ -358,7 +358,33 @@ var MonoSupportLib = {
 	mono_wasm_fire_bp: function () {
 		console.log ("mono_wasm_fire_bp");
 		debugger;
-	}
+	},
+
+	// Win32S = 0,
+	// Win32Windows = 1,
+	// Win32NT = 2,
+	// WinCE = 3,
+	// Unix = 4,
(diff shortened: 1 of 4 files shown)


sdks/wasm/library_mono.js
```diff
@@ -358,7 +358,33 @@ var MonoSupportLib = {
...
+	// Win32NT = 2,
+	// WinCE = 3,
+	// Unix = 4,
+	// // Since NET 3.5 SP1 or silverlight
+	// Xbox = 5,
+	// MacOSX = 6,
+	mono_wasm_os_platform_id: function () {
+		if (ENVIRONMENT_IS_WEB) {
+			var platformenv = window.navigator.platform.toUpperCase();
+			if (platformenv.indexOf("MAC")>=0
```
kg: This kind of substring match is too fragile, I think you need to have whitespace at the beginning or end of the match patterns.

vargaz: Agree, I think we should return WebAssembly or something.

kjpou1: We can setup another PlatformId for WebAssembly.

kjpou1:
Issue opened here: https://github.com/dotnet/standard/issues/1073

PR #13050, opened 2019-02-18
[mcs] Remove doc about CAS security

marek-safar:


marek-safar:
@monojenkins backport 2019-02

PR #13054, opened 2019-02-19
[netcore] Fix GCHandles, the lowest bit is used by the GCHandle class to store information.

vargaz:


vargaz:
@monojenkins build failed

PR #13071, opened 2019-02-19
[runtime] Move mono_get_context_capture_method () from security-manager.c to object.c.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

mono/metadata/object.c
```diff
@@ -8008,6 +8008,25 @@ mono_wait_handle_get_handle (MonoWaitHandle *handle)
...
 
+/*
+ * Returns the MonoMethod to call to Capture the ExecutionContext.
+ */
+MonoMethod*
+mono_get_context_capture_method (void)
+{
+	static MonoMethod *method;
+
+	/* older corlib revisions won't have the class (nor the method) */
```
spouliot: that comment does not make sense anymore and the check was not moved, which is good :-)

PR #13091, opened 2019-02-20
[ci] Prefer gmake over make for SDKs build

akoeplinger:
Apple's `make` doesn't support the `--output-sync` option and Homebrew's `make` is installed as `gmake`.

Also fixed a few places in the repo where we directly invoked make instead of $(MAKE).

### DIFF
--- a/scripts/ci/run-jenkins.sh
+++ b/scripts/ci/run-jenkins.sh
@@ -17,6 +17,7 @@ helix_set_env_vars
 helix_send_build_start_event "build/source/$MONO_HELIX_TYPE/"
 
 make_timeout=300m
+gnumake=$(which gmake || which gnumake || which make)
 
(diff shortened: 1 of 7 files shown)


scripts/ci/run-jenkins.sh
```diff
@@ -17,6 +17,7 @@ helix_set_env_vars
 helix_send_build_start_event "build/source/$MONO_HELIX_TYPE/"
 
 make_timeout=300m
+gnumake=$(which gmake || which gnumake || which make)
```
vargaz: Shouldn't this be named MAKE ?

akoeplinger: I thought about it but feared it might conflict with `make`'s internal `$(MAKE)` variable.

vargaz: The two should be the same.

Issue #12157, opened 2018-12-20
new DriveInfo(Path.GetTempPath()) crashes on macOS


EgorBo:
```csharp
string path = Path.GetTempPath();
var driveInfo = new DriveInfo(path);
```

Crashes with
```
Unhandled Exception:
System.ArgumentException: The drive name does not exist
Parameter name: driveName
  at System.IO.DriveInfo..ctor (System.String driveName) [0x000b7] in /Users/builder/jenkins/workspace/build-package-osx-mono/master@2/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.IO/DriveInfo.cs:66 
  at defefef.MainClass.Main (System.String[] args) [0x00007] in /Users/egorb/Projects/kkook/defefef/Program.cs:11 
[ERROR] FATAL UNHANDLED EXCEPTION: System.ArgumentException: The drive name does not exist
Parameter name: driveName
  at System.IO.DriveInfo..ctor (System.String driveName) [0x000b7] in /Users/builder/jenkins/workspace/build-package-osx-mono/master@2/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.IO/DriveInfo.cs:66 
  at defefef.MainClass.Main (System.String[] args) [0x00007] in /Users/egorb/Projects/kkook/defefef/Program.cs:11 
```

But doesn't crash on .NET Core.
.NET Core uses this snippet in order to obtain `DriveInfo.DriveFormat` property in System.IO tests

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

mono-master, macOS

PR #13105, opened 2019-02-20
Try to extract valid drive name from DriveInfo ctor string argument

MaximLipnin:
It's possible to pass any string as an argument to DriveInfo ctor (e.g.  `new DriveInfo(Path.GetTempPath())`). Mono throws ArgumentException if there is no exactly the same drive name in the list of available drives.

I tried to handle this case by extracting any valid drive name from the array of available drives sorted by path in descending order.

Fixes https://github.com/mono/mono/issues/12157

### DIFF
--- a/mcs/class/corlib/System.IO/DriveInfo.cs
+++ b/mcs/class/corlib/System.IO/DriveInfo.cs
@@ -55,11 +55,11 @@ namespace System.IO {
 			}
 
 			DriveInfo [] drives = GetDrives ();
+			Array.Sort (drives, (DriveInfo di1, DriveInfo di2) => String.Compare (di2.path, di1.path, true));
 			foreach (DriveInfo d in drives){
-				if (d.path == driveName){
+				if (driveName.StartsWith (d.path, StringComparison.OrdinalIgnoreCase)){
 					this.path = d.path;
 					this.drive_format = d.drive_format;
-					this.path = d.path;
 					return;
 				}
 			}
--- a/mcs/class/corlib/System.IO/DriveInfo.cs
+++ b/mcs/class/corlib/System.IO/DriveInfo.cs
@@ -56,14 +56,16 @@ namespace System.IO {
 
 			DriveInfo [] drives = GetDrives ();
 			Array.Sort (drives, (DriveInfo di1, DriveInfo di2) => String.Compare (di2.path, di1.path, true));
+			var sb = new StringBuilder($"driveName: {driveName}; available drives: ");
 			foreach (DriveInfo d in drives){
+				sb.Append($"{d.path}, ");
 				if (driveName.StartsWith (d.path, StringComparison.OrdinalIgnoreCase)){
 					this.path = d.path;
 					this.drive_format = d.drive_format;
 					return;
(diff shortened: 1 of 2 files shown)


marek-safar:
This breaks existing test

mcs/class/corlib/System.IO/DriveInfo.cs
```diff
@@ -55,15 +55,17 @@ public DriveInfo (string driveName)
 			}
 
 			DriveInfo [] drives = GetDrives ();
+			Array.Sort (drives, (DriveInfo di1, DriveInfo di2) => String.Compare (di2.path, di1.path, true));
+			var sb = new StringBuilder($"driveName: {driveName}; available drives: ");
```
marek-safar: I don't think it's good idea to expose this in the exception

MaximLipnin: Oh, it's just for small diagnostic. I'll remove it.

mcs/class/corlib/Test/System.IO/DriveInfoTest.cs
```diff
@@ -54,7 +54,22 @@ public void Constructor ()
...
-			Assert.Throws<ArgumentException> (() => new DriveInfo ("/monodriveinfotest"));
+			Assert.Throws<ArgumentException> (() => new DriveInfo ("monodriveinfotest"));
+		}
+
+		[Test]
+		[Category ("NotWasm")] // it doesn't know about 'memfs' drive format
+		public void ConstructorGetsValidDriveFromNonDriveString ()
+		{
+			if (Environment.OSVersion.Platform != PlatformID.Win32NT && Environment.OSVersion.Platform != PlatformID.MacOSX)
+				Assert.Ignore ("Some Linux-hosted CI builders don't have '/' mounted, just testing Windows and MacOS for now.");
```
akoeplinger: @MaximLipnin `Environment.OSVersion.Platform` for historical reasons returns `Unix` on macOS so right now this only runs on Windows.

MaximLipnin: @akoeplinger Thanks for info. I tested it on my local macOS env so I assume it should pass on CI too in case of switching to "right" macOS platform id.

akoeplinger: @MaximLipnin yeah if you want to enable it on macOS I'd switch to the newer RuntimeInformation APIs:

```
if (!RuntimeInformation.IsOSPlatform (OSPlatform.Windows) && !RuntimeInformation.IsOSPlatform (OSPlatform.OSX))
  Assert.Ignore ("Some Linux-hosted CI builders don't have '/' mounted, just testing Windows and MacOS for now.");
```

MaximLipnin: Ok, I'll make this change.

MaximLipnin: Updated in https://github.com/mono/mono/pull/13205

PR #13117, opened 2019-02-21
[aot] Only the static interp trampolines/wrappers into the mscorlib AOT image.

vargaz:


lewurm:
we should also add this guard to `mono_arch_get_interp_to_native_trampoline` and `mono_arch_get_native_to_interp_trampoline`

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #13123, opened 2019-02-21
Mono.CSharp is not included in winaot build causing test failure in full AOT.

lateralusX:
Since it is not included it won't be full AOT:ed as part of build and failing Mono.CSharp tests at a later point in time, since that won't do full AOT of assembly.


lateralusX:
Will remove the Mono.CSharp from winaot test suite instead. Creating a new PR for that.

Issue #13030, opened 2019-02-15
Directory.GetLogicalDrives () missing SystemNative_GetAllMountPoints pal function


akoeplinger:
## Steps to Reproduce

1. Run `Directory.GetLogicalDrives ()`
2. Observe EntryPointNotFoundException

This is due to the corefx code that landed as part of https://github.com/mono/mono/pull/11342.

### Current Behavior

```
1) GetLogicalDrives (MonoTests.System.IO.DirectoryTest.GetLogicalDrives)
   System.EntryPointNotFoundException : SystemNative_GetAllMountPoints assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Interop+Sys.GetAllMountPoints(Interop/Sys/MountPointFound)
  at Interop+Sys.GetAllMountPoints () [0x00019] in /Users/alexander/dev/mono/external/corefx/src/Common/src/Interop/Unix/System.Native/Interop.MountPoints.cs:26
  at System.IO.DriveInfoInternal.GetLogicalDrives () [0x00000] in /Users/alexander/dev/mono/external/corefx/src/Common/src/System/IO/DriveInfoInternal.Unix.cs:13
  at System.IO.FileSystem.GetLogicalDrives () [0x00000] in /Users/alexander/dev/mono/external/corefx/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs:520
  at System.IO.Directory.GetLogicalDrives () [0x00000] in /Users/alexander/dev/mono/external/corefx/src/System.IO.FileSystem/src/System/IO/Directory.cs:303
```

### Expected Behavior

No exception.

## On which platforms did you notice this

[X ] macOS
[ ] Linux
[ ] Windows

**Version Used**: master

EgorBo:
Didn't we have a script that checked pinvokes?

akoeplinger:
I've never seen one but on XI it's kinda implicitly checked since all pinvokes are required to exist (I found it during 2019-02 integration).

I agree it would be nice to have!

PR #13125, opened 2019-02-21
Add pal_mount to System.Native

EgorBo:
Fixes https://github.com/mono/mono/issues/13030


EgorBo:
Android:
```
 error: implicit declaration of function 'setmntent' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    FILE* fp = setmntent("/proc/mounts", MNTOPT_RO);
```

mono/native/pal_mount_tmp.c
```diff
@@ -64,6 +64,9 @@ static int32_t GetMountInfo(MountPointFound onFound)
     }
 
     return result;
+#else
+    return 0;
```
akoeplinger: does this mean this API will return empty result on Android?

akoeplinger:
Please add tests for non-empty results for `Environment.GetDrives()` and `Directory.GetLogicalDrives()`.

`DriveInfo.GetDrives()` seems to be tested already.

mono/native/pal_mount_tmp.c
EgorBo: @akoeplinger as far as I understand, `setmntinfo` is not available in Android NDK

EgorBo: It was added in Android-20 api level so I guess the `configure` check should handle it?..

akoeplinger: Our previous code did somehow handle it though so we need to add that back to corefx and not regress functionality.

Here's the output I get from `Directory.GetLogicalDrives()` on XA 9.2:

```
/system
/vendor
/mnt/runtime/default/emulated
/mnt/runtime/read/emulated
/mnt/runtime/write/emulated
/cache
/data
/data/var/run/netns
/storage/emulated
```

akoeplinger: Oh and I think we should put this code into corefx's `pal_mount.c` file instead of copy-pasting it into mono.

EgorBo: That's why I put "_tmp_" in its name ;-) 
I guess it's better to use existing icall then, it's Android and AIX friendly.

mcs/class/corlib/Test/System/EnvironmentTest.cs
```diff
@@ -239,5 +239,11 @@ public void MachineNameIsNotFullyQualifiedDomainName ()
 			Assert.AreEqual (-1, Environment.MachineName.IndexOf("."));
 		}
 #endif
+
+		[Test] // https://github.com/mono/mono/issues/13030
+		public void GetLogicalDrivesNotEmpty ()
+		{
+			Assert.AreNotEqual (0, Directory.GetLogicalDrives ().Length);
```
akoeplinger: should be `Environment.GetLogicalDrives ()` here.

better would be to use `CollectionAssert.IsNotEmpty ()` so we get better message in case of failure.

EgorBo: Oh, forgot about CollectionAssert thanks

mcs/class/corlib/Test/System/EnvironmentTest.cs
```diff
@@ -239,5 +239,11 @@ public void MachineNameIsNotFullyQualifiedDomainName ()
 			Assert.AreEqual (-1, Environment.MachineName.IndexOf("."));
 		}
 #endif
+
+		[Test] // https://github.com/mono/mono/issues/13030
+		public void GetLogicalDrivesNotEmpty ()
+		{
+			CollectionAssert.IsNotEmpty (Directory.GetLogicalDrives ());
```
akoeplinger: this still needs to be `**Environment**.GetLogicalDrives()`

marek-safar:
@monojenkins backport 2019-02

akoeplinger:
@monojenkins commit csproj

PR #13130, opened 2019-02-21
[System.Private.CoreLib] Implement more APIs

marek-safar:


mono/metadata/icall-def.h
```diff
@@ -582,6 +582,14 @@ ICALL(OBJ_2, "InternalGetHashCode", mono_object_hash_internal)
 HANDLES(OBJ_3, "MemberwiseClone", ves_icall_System_Object_MemberwiseClone, MonoObject, 1, (MonoObject))
 
 ICALL_TYPE(ASSEM, "System.Reflection.Assembly", ASSEM_2)
+#if ENABLE_NETCORE
+HANDLES(ASSEM_2, "GetCallingAssembly", ves_icall_System_Reflection_Assembly_GetCallingAssembly, MonoReflectionAssembly, 0, ())
```
vargaz: The ASSEM_ constants don't need to be in order, just need to be unique, so only the extra icall needs to be in an #ifdef.

vargaz: The entries do need to be ordered by name though.

Issue #13082, opened 2019-02-20
mono --response does not work with UTF8 response files (or characters)


dellis1972:
Related to https://github.com/mono/mono/issues/13081.

As part of the Xamarin.Android effort to reduce the length of our command line arguments when calling the cross compilers we have tried to use `--response`. If a user has non-Ascii characters in the path and/or they use a utf8 formatted file the response file will not be parsed correctly.

## Steps to Reproduce

1. Download and extract 
[ResponseUnicode.zip](https://github.com/mono/mono/files/2884154/ResponseUnicode.zip)
2. Open a mono command prompt and navigate to when you extracted the file.
3. run `mono --response=resp.txt`

### Current Behaviour

If the `resp.txt` file is formatted as `ASCII` you will get the following error.
`Cannot open assembly 'Execütable\Program.exe': Illegal byte sequence.`

If the file is formatted as `utf8` you will get the following error
`Cannot open assembly 'ï»¿ExecÃ¼table\Program.exe': No such file or directory.`

### Expected Behaviour

What should happen is the `Program.exe` should run and print `Hello`

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[x] Windows

Mono JIT compiler version 5.18.0 (Visual Studio built mono)

lateralusX:
Did some investigation, if I remove the BOM in the UTF version it works as expected running the program using the response file as argument. I will add support to ignore the BOM mark if one exist in the file to make sure similar problems are avoided (since saving in UTF8 from editors tends to add that mark into the file). I will also implement logic to handle the parameters in the response file in similar way as we handle parameters in main method. That will most likely take care of the problems seen in https://github.com/mono/mono/issues/13081 as well.

lateralusX:
Assumption will be that the --response file must be encoded in UTF8 (or ASCII), no other encodings will be handled, at least not for now.

lateralusX:
Should be fixed by, https://github.com/mono/mono/pull/13132.

Issue #13081, opened 2019-02-20
`--response` does not handle path separators on Windows.


dellis1972:
As part of the Xamarin.Android effort to reduce the length of our command line arguments when calling the cross compilers we have tried to use `--response`. It seems to have a a few problems.
The first is that in order to use path separators within the response file you need to delimit them. 
So `c:\Foo\Bar.dll` would need to be written as `c:\\Foo\\Bar.dll`. While this is not a huge problem it does mean that we need to have some special processing between windows and Unix based systems. 

## Steps to Reproduce (Assumes you have latest stable mono for windows installed)

1. Download [ResponsePaths.zip](https://github.com/mono/mono/files/2884129/ResponsePaths.zip) and extract it on windows. 
2. Open a mono command prompt and navigate to when you extracted the file.
3. run `mono --response=resp.txt`

### Current Behaviour

Current you will get the following 

`Cannot open assembly 'ExecutableProgram.exe': No such file or directory.`

### Expected Behaviour

What should happen is the `Program.exe` should run and print `Hello`
## On which platforms did you notice this

[ ] macOS
[ ] Linux
[x] Windows

Mono JIT compiler version 5.18.0 (Visual Studio built mono)

dellis1972:
This might also be a problem on Mac given the following.
```
MININT-G351LDI:ResponseUnicode dean$ mono Execütable/Program.exe 
Hello
MININT-G351LDI:ResponseUnicode dean$ mono --response=resp.txt
Cannot open assembly '﻿Execütable/Program.exe': No such file or directory.
MININT-G351LDI:ResponseUnicode dean$ file resp.txt 
resp.txt: UTF-8 Unicode (with BOM) text
```

jaykrell:
Please consider trying to use forward slashes instead.
Win32 supports them.
c:/foo/bar.txt

etc.

Pasting them into an open file dialog might not work, but CreateFile, GetFileAttributes, etc., all support them (unless you use the \\? long path support).

lateralusX:
Should be fixed by https://github.com/mono/mono/pull/13132.

PR #13132, opened 2019-02-21
 Fix multiple issues with response file option on Windows.

lateralusX:
* If file is UTF8 encoded and include BOM mark, implementation failed to detect and remove BOM mark causing errors. Fix check for BOM mark and if found, remove it before passed to parsing. NOTE, current implementation only supports UTF8 (and ASCII) content in response file and won't detect or handle any other codepages.

* Since the content in --response file should mimic what should have been passed on command line, we should apply same parsing rules on Windows as done by the shell when running the arguments on the command line. Current implementation uses its own parsing that is not the same as used by the Windows shell, meaning that arguments can't be identical in response file compared to what's passed on command line. This commit fix parsing of the command line in the response file so its first parsed by the shell before added to internal argc, argv parameters (similar as done when passed to main function).

NOTE, most of the old code around mono_parse_options was just split to be reused by two different functions. No focus on rewriting any of the original parsing logic. Implemented specific parse methods for response file in order to minimize regression impact of change on Windows. On none windows platforms its a 1:1 between mono_parse_response_options and mono_parse_options but on Windows the behavior is different to mimic the command line arguments. Since mono_parse_options is used in a couple of other scenarios on Windows, the fix will make sure the original behavior of mono_parse_options is the same on Windows and only the new response file parsing will get the new ability to handle the arguments in the same way as the command line.

### DIFF
--- a/mono/mini/driver.c
+++ b/mono/mini/driver.c
@@ -2361,9 +2361,14 @@ mono_main (int argc, char* argv[])
 			response_options = response_content;
 
 			// Check for UTF8 BOM in file and remove if found.
-			if (response_content_len > 3 && response_content [0] == '\xef' && response_content [1] == '\xbb' && response_content [2] == '\xbf') {
+			if (response_content_len >= 3 && response_content [0] == '\xef' && response_content [1] == '\xbb' && response_content [2] == '\xbf') {
 				response_content_len -= 3;
 				response_options += 3;
+			}
+
+			if (response_content_len == 0) {
+				fprintf (stderr, "The specified response file is empty\n");
+				exit (1);
 			}
 
 			mono_parse_response_options (response_options, &argc, &argv, FALSE);
@@ -3041,6 +3046,28 @@ mono_win32_parse_options (const char *options, int *ref_argc, char **ref_argv []
 	GPtrArray *array = g_ptr_array_new ();
 	optionsw = g_utf8_to_utf16 (options, -1, NULL, NULL, NULL);
 	if (optionsw) {
+		gunichar2 *p;
+		gboolean in_quotes = FALSE;
+		gunichar2 quote_char = L'\0';
+		for (p = optionsw; *p; p++){
+			switch (*p){
+			case L'\n':
+				if (!in_quotes)
+					*p = L' ';
+				break;
+			case L'\'':
+			case L'"':
+				if (in_quotes) {
+					if (quote_char == *p)
+						in_quotes = FALSE;
+				} else {
+					in_quotes = TRUE;
+					quote_char = *p;
+				}
+				break;
+			}
+		}
+
 		argv = CommandLineToArgvW (optionsw, &argc);
 		if (argv) {
 			for (int i = 0; i < argc; i++)
--- a/mono/mini/driver.c
+++ b/mono/mini/driver.c
@@ -133,6 +133,7 @@ static const gint16 opt_names [] = {
 #define EXCLUDED_FROM_ALL (MONO_OPT_SHARED | MONO_OPT_PRECOMP | MONO_OPT_UNSAFE | MONO_OPT_GSHAREDVT)
 
 static char *mono_parse_options (const char *options, int *ref_argc, char **ref_argv [], gboolean prepend);
+static char *mono_parse_response_options (const char *options, int *ref_argc, char **ref_argv [], gboolean prepend);
 
 static guint32
 parse_optimizations (guint32 opt, const char* p, gboolean cpu_opts)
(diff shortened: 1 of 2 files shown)


lateralusX:
Should fix both https://github.com/mono/mono/issues/13082 and https://github.com/mono/mono/issues/13081.

mono/mini/driver.c
```diff
@@ -2348,15 +2349,25 @@ mono_main (int argc, char* argv[])
...
 				fprintf (stderr, "The specified response file can not be read\n");
 				exit (1);
 			}
-			mono_parse_options (text, &argc, &argv, FALSE);
-			g_free (text);
+
+			response_options = response_content;
+
+			// Check for UTF8 BOM in file and remove if found.
+			if (response_content_len > 3 && response_content [0] == '\xef' && response_content [1] == '\xbb' && response_content [2] == '\xbf') {
```
jaykrell: Shouldn't it be >=3 not >3? i.e. an empty file?

mono/mini/driver.c
```diff
@@ -2972,40 +3014,63 @@ mono_parse_options (const char *options, int *ref_argc, char **ref_argv [], gboo
...
+	gunichar2 **argv;
+	gunichar2 *optionsw;
+
+	if (!options)
+		return NULL;
+
+	GPtrArray *array = g_ptr_array_new ();
+	optionsw = g_utf8_to_utf16 (options, -1, NULL, NULL, NULL);
+	if (optionsw) {
+		argv = CommandLineToArgvW (optionsw, &argc);
```
jaykrell: Likely CommandLineToArgvW is different than the C runtime.
Maybe try to reuse C runtime logic instead?
Dependency on shell32.dll etc. is distasteful too imho.

mono/mini/driver.c
```diff
@@ -2920,21 +2931,52 @@ mono_parse_options_from (const char *options, int *ref_argc, char **ref_argv [])
...
-mono_parse_options (const char *options, int *ref_argc, char **ref_argv [], gboolean prepend)
+static void
+merge_parsed_options (GPtrArray *parsed_options, int *ref_argc, char **ref_argv [], gboolean prepend)
 {
 	int argc = *ref_argc;
 	char **argv = *ref_argv;
+
+	if (parsed_options->len > 0){
+		int new_argc = parsed_options->len + argc;
+		char **new_argv = g_new (char *, new_argc + 1);
```
jaykrell: Consider g_new0, but I realize it is redundant if you write to every array element.

mono/mini/driver.c
lateralusX: This will mimic the exact same behavior as you will get when passing the arguments to main method on Windows and that's was the whole point. The dependency already exists from other sources, so this is not a new introduced dependency.

mono/mini/driver.c
lateralusX: Thought about that as well, but its the same implementation as found in other places of the source code like, mono_config_parse_file_with_context, mono_domain_set_options_from_config and if file is empty things will still break, but agree that we could error case for file just including BOM mark.

mono/mini/driver.c
lateralusX: Just move of existing code and should be fine.

mono/mini/driver.c
jaykrell: I think it should be  >=3 and gracefully handle an otherwise empty file.

mono/mini/driver.c
jaykrell: This does not mimic main/wmain. That was my point. It is different regarding quotes. I'll have to write small samples.

lateralusX: What I mean was that it mimic how mono handles arguments on Windows since it doesn't use arguments passed to main/wmain.

jaykrell: Oh right, that. Yes that is dubious too. Quotes and spaces are a mess no matter what.
So, there another angle here to consider.
Once you are in a response file, the argv delimiters can be considered newlines and carriage returns instead of spaces. Spaces can be left therefore unquoted.
The "new" difficult to pass characters can become newlines/carriage returns, which are more rare than horizontal spaces.
I'm not sure if there is precedent for that or not. Certainly accepting newlines in response files is normal, the compiler/linker do, and build systems use it. I'm not sure that affects space/quote rules though.

lateralusX: True, I will make a try to handle newlines in the file, probably use the parsing logic already available to handle this inside/outside quotes and then let the CommadLineToArgvW take care of the rest.

lateralusX: Putting newline in the response file also exposed another issue on Windows and the way g_file_get_contents is implemented, need to adjust that one as well.

mono/mini/driver.c
lateralusX: Will fix.

mono/mini/driver.c
jaykrell: I did a quick experiment with cl.exe.
My hopes that newlines act as quotes were dashed.
You still need quotes around spaces. So it still a mess.
My assumption that newlines are skipped, or treated as whitespace separating parameters, were preserved.

I don't know how common it is, but some build systems definitely put newlines in compiler/linker response files. It makes the response files somewhat more readable, for those few of us that debug this sort of thing. Perhaps nobody will do that to mono.

Eventually or presumably already, VS and VS4Mac are consumers of this.
So what they want/do is important.

lateralusX: Latest commit seems to take care of most scenarios for newline in response file. Preserved when inside quotes, but replaced with whitespace when outside so that shell argument parsing will handle them correctly. Tried with various response file (one was 21 KB) that mixed new lines, white spaces etc and seemed to get expected results. For example (uses new line, tabs, space both inside outside quotes):

```
-argument1 "test
 some more"
-argument2 test2	  	  

 
-argument3 "some	more whitespaces	"
     

Execütable\Program.exe
```

Gets parsed to this (debugger output):

argv[0] L"-argument1"
argv[1] L"test\n some more"
argv[2] L"-argument2"
argv[3] L"test2"
argv[4] L"-argument3"
argv[5] L"some\tmore whitespaces\t"
argv[6] L"Execütable\\Program.exe"

lateralusX:
@monojenkins rebuild failed

luhenry:

Issue #10101, opened 2018-08-14
ref local of jagged array leads to InvalidProgramException


lambdageek:
If object `f` has a field `a` of type `int[][]`, then `ref int[] r = ref f.a[1];` leads to an InvalidProgramException

## Steps to Reproduce

1.  `csc Foo.cs`
2.  `mono Foo.exe`

Foo.cs:
```csharp
using System;

public class Foo {
	public int[][] a;

	public Foo () {
		a = new int[][]{
			new int[]{1,2,3}, 
			new int[]{4,5,6},
			new int[]{7,8,9}
		};
	}

	public static void Main ()
	{
		var f = new Foo ();

		Console.WriteLine ($"{f.a[1][1]}");

		ref int[] r = ref f.a[1];

		bar (ref r);

		Console.WriteLine ($"{f.a[1][1]}");
	}

	public static void bar (ref int[] x)
	{
		x = new int[]{11,12,13};
	}
}
```

### Current Behavior

```
Unhandled Exception:
System.InvalidProgramException: Invalid IL code in Foo:Main (): IL_0030: stloc.1   


[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidProgramException: Invalid IL code in Foo:Main (): IL_0030: stloc.1   
```

### Expected Behavior

```
5
12
```

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.18.0.20 (2018-08/64e9b1a9839 Mon Aug 13 14:14:00 EDT 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

Also fails with mono `master`.
Also fails with `2018-02`:
```
Mono JIT compiler version 5.12.0.301 (2018-02/4fe3280bba1 Fri Jul 20 08:25:42 EDT 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
[...]

PR #13134, opened 2019-02-21
[jit] Fix the class of the stack entry for ldelema, it should be the class not its element class.

vargaz:
Fixes https://github.com/mono/mono/issues/10101.

### DIFF
--- a/mono/mini/arrays.cs
+++ b/mono/mini/arrays.cs
@@ -525,7 +525,7 @@ class Tests
 		size = -1;
 		try {
 			res = new float [size];
-		} catch (OverflowException e) {
+		} catch (OverflowException) {
 
 		} catch (Exception) {
 			return 1;
@@ -536,7 +536,7 @@ class Tests
 		size = -2147483648;
 		try {
 			res = new float [size];
-		} catch (OverflowException e) {
+		} catch (OverflowException) {
 
 		} catch (Exception) {
 			return 3;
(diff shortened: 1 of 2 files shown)


mono/mini/method-to-ir.c
```diff
@@ -3976,7 +3976,7 @@ mini_emit_ldelema_1_ins (MonoCompile *cfg, MonoClass *klass, MonoInst *arr, Mono
 		static const int fast_log2 [] = { 1, 0, 1, -1, 2, -1, -1, -1, 3 };
 
 		EMIT_NEW_X86_LEA (cfg, ins, array_reg, index2_reg, fast_log2 [size], MONO_STRUCT_OFFSET (MonoArray, vector));
-		ins->klass = m_class_get_element_class (klass);
+		ins->klass = klass;
```
lambdageek: How precise does the class have to be in IR?  `klass` here is some array type, right?  So this should be a reference (or maybe ptr) to the array element type - or would that end up generating worse code down the line?

vargaz: Its mostly used for verification.

mono/mini/arrays.cs
```diff
@@ -840,8 +840,30 @@ private unsafe static void WritePtr (FooStruct *val, out FooStruct* ptr)
...
+		var f = new JaggedClass ();
+
+		ref int[] r = ref f.a[1];
+
+		return r [0];
+	}
 
+	public static int bar (ref int[] x) {
+		return x [0];
+	}
```
lewurm: `bar ()` is not used

vargaz:
@monojenkins build failed

PR #13135, opened 2019-02-21
[monodroid] Use CoreFX to implement System.Drawing types that used to be in Mono.Android.dll

alexischr:


alexischr:

mcs/class/Facades/System.Drawing.Common/TypeForwarders.cs
```diff
@@ -10,6 +10,9 @@
 // System.Drawing.Primitives/ref:
 namespace System.Drawing
 {
+
+#if !MONODROID
```
marek-safar: Why do you need the ifdef if the file is not used for MONODROID

alexischr: Not sure I understand, this file is included. monodroid profiles sources include `System.Drawing.Common.dll.sources` which includes `TypeForwarders.cs`

marek-safar: oh, I missed the #include

alexischr:

akoeplinger:
Judging from the API Diff we're still missing a few `TypeForwardedFromAttribute("Mono.Android, ...")` as it's only on `SystemColors` right now.

Merging to get my other PR rolling but we need to keep track of this :)

alexischr:
This commit https://github.com/mono/corefx/commit/68f67ad6ced8b74a65a1a5617a4c761bb203a25b was missing from corefx master so it will be picked up in the next bump. `2019-02` has it now too (via a new corefx `mono-2019-02` branch)

PR #13138, opened 2019-02-21
[aot] abort lookup early when a method isn't available in AOT image

lewurm:
Mixed mode depends on that.  Usually, compile_method of mini takes care
to request only existing methods from the AOT image, but in some cases,
e.g. from a trampoline, AOT runtime is quiered directly.

Contributes to https://github.com/xamarin/xamarin-macios/issues/5618

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -4134,6 +4134,11 @@ load_method (MonoDomain *domain, MonoAotModule *amodule, MonoImage *image, MonoM
 	}
 
 	if (!code) {
+		if (method_index < amodule->info.nmethods)
+			code = (guint8 *)amodule->methods [method_index];
+		else
+			return NULL;
+
 		/* JITted method */
 		if (amodule->methods [method_index] == GINT_TO_POINTER (-1)) {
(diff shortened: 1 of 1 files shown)


lewurm:
@monojenkins build failed

lewurm:
@vargaz @BrzVlad could I get a review on this please?

mono/mini/aot-runtime.c
```diff
@@ -4152,8 +4152,11 @@ load_method (MonoDomain *domain, MonoAotModule *amodule, MonoImage *image, MonoM
 			}
 			return NULL;
 		}
+
```
BrzVlad: Doing this check here, after we actually accessed the method entry above looks fishy. Also, shouldn't the calling code make sure to pass a correct method_index.

lewurm: good point, I moved the bound check before the access.

passing an invalid `method_index` is okay I think (see function comment).

lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2018-12
@monojenkins backport 2018-10

PR #13139, opened 2019-02-21
[aot] use unbox arbitrary trampolines for "slim" mscorlib

lewurm:
We need the recently introduced "unbox arbitrary" trampolines in one more
scenario: non-mscorlib AOT code calls a method from mscorlib with
interpreter. Before we just relied on the amodule not being available,
but mscorlib amodule is always present due to wrappers, trampolines,
etc.

Contributes to https://github.com/xamarin/xamarin-macios/issues/5618


lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2018-12
@monojenkins backport 2018-10

Issue #13113, opened 2019-02-21
Remoting error while running ML.net tests on mono Unix


Anipik:
I am using nightly build of mono https://www.mono-project.com/download/nightly/#download-lin
to build and run the tests of ML.net repo on mono unix.

## Steps to Reproduce
1. ./init-tools.sh build -Debug-netfx (change the target framework from net461 to net472) 
2. ./build-netfx -Debug-netfx
3. msbuild Microsoft.ML.sln /p:Configuration=Debug-netfx
4. mono /git/machinelearning/packages/microsoft.testplatform/15.9.0/tools/net451/Common7/IDE/Extensions/TestPlatform/vstest.console.exe /git/machinelearning/bin/AnyCPU.Debug-netfx/Microsoft.ML.OnnxTransformTest/net472//Microsoft.ML.OnnxTransformTest.dll

### Current Behavior

It results in a remoting exception

```C#

root@44e83878be48:/git/machinelearning# mono /git/machinelearning/packages/microsoft.testplatform/15.9.0/tools/net451/Common7/IDE/Extensions/TestPlatform/vstest.console.exe /git/machinelearning/bin/AnyCPU.Debug-netfx/Microsoft.ML.OnnxTransformTest/net472//Microsoft.ML.OnnxTransformTest.dll
Microsoft (R) Test Execution Command Line Tool Version 15.9.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v2.4.0 (64-bit Desktop .NET 4.0.30319.42000)
[xUnit.net 00:00:00.66]   Discovering: Microsoft.ML.OnnxTransformTest
[xUnit.net 00:00:00.83]   Discovered:  Microsoft.ML.OnnxTransformTest
[xUnit.net 00:00:00.85]   Starting:    Microsoft.ML.OnnxTransformTest
[xUnit.net 00:00:01.05] Microsoft.ML.OnnxTransformTest: Catastrophic failure: System.NullReferenceException: Object reference not set to an instance of an object


Server stack trace:
  at System.Runtime.Remoting.ClientIdentity.get_ClientProxy () [0x00000] in <6a5b9d74d49940609a90b467b33ee81e>:0
[...]

PR #13140, opened 2019-02-21
If a ClientIdentity has no ClientProxy then return null from the accessor instead of throwing

kg:
Experimental partial fix for issue #13113. From looking at the code that uses this API, I don't think it's expected for it to throw in this scenario, and it's easy to end up with an instance that doesn't have a WR at all. The code using it seems to handle scenarios where the ClientProxy is null so I think this may be the correct way for the code to work.


kg:
It doesn't fix it, but it reveals an underlying issue that I think is the actual root cause

kg:
@monojenkins build failed

PR #13142, opened 2019-02-22
Split `AppDomain.DoTypeResolve(object)` into two stongly-typed methods.

baulig:
* Instead of having one `AppDomain.DoTypeResolve(object)` that checks at runtime whether the argument is a `string` or a `TypeBuilder`, use two strongly-typed methods `DoTypeResolve(string)` and `DoTypeBuilderResolve(TypeBuilder)`.

* Add both to the `corlib/LinkerDescriptor/mscorlib.xml`.

* Adjust the code in `appdomain.c` accordingly.

* Bump corlib version.

This is part of a larger on-going project to improve the linker logic.

The rationale behind this is that the new `AppDomain.DoTypeResolve(string)` that is being called from `mono_domain_try_type_resolve_name()` should not reference `TypeBuilder`, to allow the linker to remove some of the `System.Reflection.Emit` types.

### DIFF
--- a/mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
@@ -791,7 +791,7 @@ namespace System.Reflection.Emit
 					if (!fb.IsStatic && (ft is TypeBuilder) && ft.IsValueType && (ft != this) && is_nested_in (ft)) {
 						TypeBuilder tb = (TypeBuilder)ft;
 						if (!tb.is_created) {
-							AppDomain.CurrentDomain.DoTypeResolve (tb);
+							AppDomain.CurrentDomain.DoTypeBuilderResolve (tb);
 							if (!tb.is_created) {
 								// FIXME: We should throw an exception here,
 								// but mcs expects that the type is created
--- a/configure.ac
+++ b/configure.ac
@@ -59,7 +59,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This line is parsed by tools besides autoconf, such as msvc/mono.winconfig.targets.
 # It should remain in the format they expect.
 #
-MONO_CORLIB_VERSION=4A9CEBCF-2544-4CB4-9F84-07A10C16BEEB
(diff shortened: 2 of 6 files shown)


mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
```diff
@@ -791,7 +791,7 @@ TypeInfo CreateTypeInfo ()
 					if (!fb.IsStatic && (ft is TypeBuilder) && ft.IsValueType && (ft != this) && is_nested_in (ft)) {
 						TypeBuilder tb = (TypeBuilder)ft;
 						if (!tb.is_created) {
-							AppDomain.CurrentDomain.DoTypeResolve (tb);
+							AppDomain.CurrentDomain.DoTypeBuilderResolve (tb);
```
lambdageek: Where is `DoTypeBuilderResolve` defined?

baulig: Oops, it looks like I forgot to include my `AppDomain.cs` changes.

baulig:
I originally planned on using two overloads of the same method, but then couldn't figure out how to do the overload resolution on the native side, so I went with two methods instead.  I'll update the commit message accordingly.

@lambdageek We don't need to urgently land this, but do you like the general idea of it?  I still have to work on the linker side of things, though.

lambdageek:
@baulig yea, the overall approach makes sense to me.

mcs/class/corlib/System.Reflection.Emit/TypeBuilder.cs
baulig: Should be fixed now.

lambdageek:

baulig:
@monojenkins build failed

baulig:

baulig:
@monojenkins build failed

baulig:

baulig:

baulig:
@monojenkins build Linux x64

baulig:
@monojenkins build Linux x64 c++

baulig:
@monojenkins Linux x64 C++

PR #13145, opened 2019-02-22
[wasm] Add browser test suite.

kjpou1:
- Runs a series of tests against the HttpClientHandler and blob.


kjpou1:
@monojenkins build failed

PR #13161, opened 2019-02-23
[corlib] Hardcode IsRunningOnWindows as false on WASM too.

vargaz:


vargaz:
@monojenkins build failed

PR #13164, opened 2019-02-24
Mark some corlib/System tests that hang runtime on AIX as non-working

NattyNarwhal:
Uses reflection instead of an external harness variable as recommended to me on Gitter a few weeks ago.

Might also be slightly overzealous on the exchange tests as well.

----

This should let the System and corlib tests run all the way through on Jenkins too, so the babysitter won't have to intervene.

### DIFF
--- a/mcs/class/System/Test/System.Diagnostics/ProcessTest.cs
+++ b/mcs/class/System/Test/System.Diagnostics/ProcessTest.cs
@@ -15,6 +15,7 @@ using System.IO;
 using System.Text;
 using System.Threading;
 using System.Runtime.InteropServices;
+using System.Reflection;
 
 using NUnit.Framework;
 
@@ -1035,6 +1036,18 @@ namespace MonoTests.System.Diagnostics
 		[NUnit.Framework.Category ("MobileNotWorking")]
 		public void StandardInputWrite ()
 		{
+			// Because we have no NotWorking category for AIX, use
+			// reflection to access the internal System.Platform
+			// as recommended. (XXX: RuntimeInformation instead?)
+			var platform = Assembly.Load("System")
+				?.GetType("System.Platform");
+			if ((bool)platform?.GetMethod("get_IsAix")?.Invoke(platform, null) == true
+				|| (bool)platform?.GetMethod("get_IsIBMi")?.Invoke(platform, null) == true)
+			{
+				// This test is broken on AIX because the fork child seems to become comatose.
+				Assert.Ignore ("Skipping on AIX/i");
+			}
+
 			var psi = GetEchoCrossPlatformStartInfo ();
 			psi.RedirectStandardInput = true;
 			psi.RedirectStandardOutput = true;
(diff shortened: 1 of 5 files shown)


NattyNarwhal:
hold this until I build this on x86, I suspect null traversal might be weird

NattyNarwhal:
Doesn't seem to affect tests on FreeBSD.

mcs/class/System/Test/System.Diagnostics/ProcessTest.cs
```diff
@@ -1035,6 +1036,18 @@ public void DisposeWithDisposedStreams ()
 		[NUnit.Framework.Category ("MobileNotWorking")]
 		public void StandardInputWrite ()
 		{
+			// Because we have no NotWorking category for AIX, use
+			// reflection to access the internal System.Platform
+			// as recommended. (XXX: RuntimeInformation instead?)
+			var platform = Assembly.Load("System")
+				?.GetType("System.Platform");
+			if ((bool)platform?.GetMethod("get_IsAix")?.Invoke(platform, null) == true
+				|| (bool)platform?.GetMethod("get_IsIBMi")?.Invoke(platform, null) == true)
```
akoeplinger: I think we should add AIX/IBMi to RuntimeInformation instead of doing this reflection magic.

```diff
diff --git a/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs b/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
index 7b11484ead0..494e1f5d700 100644
--- a/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
+++ b/mcs/class/corlib/System.Runtime.InteropServices.RuntimeInformation/RuntimeInformation.cs
@@ -62,6 +62,13 @@ namespace System.Runtime.InteropServices
                        case PlatformID.Unix:
                                return osPlatform == OSPlatform.Linux;
                        default:
+                               if (Platform.IsFreeBSD)
+                                       return osPlatform == OSPlatform.Create ("FREEBSD");
+                               if (Platform.IsIBMi)
+                                       return osPlatform == OSPlatform.Create ("IBMI");
+                               if (Platform.IsAix)
+                                       return osPlatform == OSPlatform.Create ("AIX");
+
                                return false;
                        }
 #endif
```

NattyNarwhal: Ugh, Platform is in System, so you still need to cross the reflection boundary for it. How about a uname based approach for Unix, if possible; or DllMap based? I'm thinking another icall....

akoeplinger: sorry, forgot about that 😭

how about we move `Platform` into corlib? we have InternalsVisibleTo from corlib -> System so this should work.

NattyNarwhal: Sure, that works too. Would you want to do that, or me?

akoeplinger: To sum up what we discussed on Gitter:

- Adding a new icall that gets the OS as used by the runtime for dllmap, similar to the existing RuntimeArchitecture.
- Use that for the existing IsOSPlatform implementation
- Cache both values in a static field so we don't need to the icall every time

NattyNarwhal:
Alright, icall added and tests fixed.

PR #13165, opened 2019-02-24
[System.Private.CoreLib] Implement more APIs

marek-safar:


mcs/class/System.Private.CoreLib/System.Runtime.CompilerServices/RuntimeHelpers.cs
```diff
@@ -1,7 +1,113 @@
...
+		{
+			return SufficientExecutionStack ();
+		}
+
+		public static void ExecuteCodeWithGuaranteedCleanup (TryCode code, CleanupCode backoutCode, Object userData)
+		{
+		}
+
+		public static void PrepareConstrainedRegions ()
+		{
```
filipnavara: It would be nice to have a comment that says that these methods are intentionally empty. Alternatively they could probably be moved to the shared CoreLib partition since the (empty) implementation will likely be identical across runtimes (https://github.com/dotnet/coreclr/issues/17903).

mcs/class/System.Private.CoreLib/System.Threading/EventWaitHandle.cs
```diff
@@ -1,15 +1,78 @@
...
+
+			createdNew = true;
+		}
+
+		static OpenExistingResult OpenExistingWorker (string name, out EventWaitHandle result)
+		{
+			throw new PlatformNotSupportedException (SR.PlatformNotSupported_NamedSynchronizationPrimitives);
+		}
+
+		internal static bool Set (SafeWaitHandle waitHandle) => throw new NotImplementedException ();
```
filipnavara: Would be nice to have this implemented, it is practically identical to the `Set` above.

mcs/class/System.Private.CoreLib/System.Runtime.CompilerServices/RuntimeHelpers.cs
marek-safar: I have the shared PR unification somewhere but the coverage is quite low (so low priority for now)

mono/metadata/icall-def.h
```diff
@@ -124,6 +124,11 @@ HANDLES(NATIVEMETHODS_9, "SetProcessWorkingSetSize", ves_icall_Microsoft_Win32_N
 HANDLES(NATIVEMETHODS_10, "TerminateProcess", ves_icall_Microsoft_Win32_NativeMethods_TerminateProcess, MonoBoolean, 2, (gpointer, gint32))
 HANDLES(NATIVEMETHODS_11, "WaitForInputIdle", ves_icall_Microsoft_Win32_NativeMethods_WaitForInputIdle, gint32, 2, (gpointer, gint32))
 
+#if ENABLE_NETCORE // && UNIX
```
akoeplinger: the comment seems unrelated

filipnavara: On non-Unix builds it P/Invokes directly to the Win32 API, so the comment makes sense.

PR #13173, opened 2019-02-25
Cancel sync IO on Windows when aborting thread.

lateralusX:
Using CancelIO/CancelIOEx or CancelSynchronousIo on a thread that is moving into a suspended state can cause these calls to hang (they shouldn’t but have been observed to be the case). Fix does the cancel of sync IO after the aborted thread has been resumed, but before releasing suspend locks making sure thread won't be suspend again until after cancel IO call has completed.


lateralusX:
Most likely fixing random hangs seen in block_guard_restore_aligment_on_exit runtime test as well since it seems to hit this code path (doing abort + console IO).

lateralusX:
Backport?

lateralusX:
@monojenkins rebuild failed

luhenry:
Please add tests covering the failure.

lateralusX:
Added two test testing different aspects of the abort blocking IO logic, one for a blocking accept on sockets (triggering the socket path) and the other on a blocking read on console (triggering the file IO path). Since the logic around aborting blocking IO calls is Windows specific the new tests has only been added to win32_net_4_x and winaot profiles.

Issue #13133, opened 2019-02-21
[perf] Major array copying performance difference 


alexanderkyte:
The test `System.Tests.Perf_Array.ArrayCopy2D`(https://github.com/dotnet/performance/blob/0c626c0de72e9e018bb02d0f92b97117098f1fee/src/benchmarks/micro/corefx/System.Runtime/Perf.Array.cs#L236) shows a big slowdown in our array copy compared to .NET core. 

- This takes `3284214.702 ns` on Mono
- This takes only `342.3348056 ns`  for .NET core. This is 00.0104% of the time it takes Mono.

A reproduction of this has been made with the executable here: https://gist.github.com/alexanderkyte/3cc2522a0b7391a1326dc65afcd4a3db

When profiled with Instruments, we see a slowdown due to some handle manipulation. It doesn't seem to be falling back to any intrinsic. 

![array2dslowdown](https://user-images.githubusercontent.com/700530/53198451-74d13d80-35ea-11e9-9c07-bed98bd6c6d1.jpeg)

lambdageek:
It looks like `System.Array.FastCopy` (called from `System.Array.Copy (Array sourceArray, int sourceIndex, Array destinationArray, int destinationIndex, int length)`) is returning FALSE here.  The test is a copies a rank 2 int32 array to another rank 2 int32 array.  That ought to work.

BrzVlad:
This looks fixed. The two runtimes are comparable in performance on the attached benchmark on linux.

PR #13181, opened 2019-02-25
[runtime] Use FastCopy on multidimensional arrays

alexanderkyte:
This fixes 1 of 2 identified issues found in: https://github.com/mono/mono/issues/13133

We need a faster slow path and we need to use the fast path for 2d and 3d arrays. 

This makes us use the fast path on 2d and 3d arrays. Our check for bounds was wrong because we were checking if the array is null. When we have a rank above 1, we will always have at least some bounds struct pointed to by the relevant pointer because we place an array there. Instead, we check that all of the bounds are zero/NULL. 

There's an immediate speed gain on the benchmark.

Using the same number of iterations:

- Before:
image

- After:
new_bench_array_fastpath_2d

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -730,8 +730,14 @@ ves_icall_System_Array_FastCopy (MonoArrayHandle source, int source_idx, MonoArr
 	if (src_vtable->rank != dest_vtable->rank)
 		return FALSE;
 
-	if (MONO_HANDLE_GETVAL (source, bounds) || MONO_HANDLE_GETVAL (dest, bounds))
-		return FALSE;
+	MonoArrayBounds *source_bounds = MONO_HANDLE_GETVAL (source, bounds);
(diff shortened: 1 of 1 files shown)


mono/metadata/icall.c
```diff
@@ -730,8 +730,14 @@ ves_icall_System_Array_FastCopy (MonoArrayHandle source, int source_idx, MonoArr
 	if (src_vtable->rank != dest_vtable->rank)
 		return FALSE;
 
-	if (MONO_HANDLE_GETVAL (source, bounds) || MONO_HANDLE_GETVAL (dest, bounds))
-		return FALSE;
+	MonoArrayBounds *source_bounds = MONO_HANDLE_GETVAL (source, bounds);
+	MonoArrayBounds *dest_bounds = MONO_HANDLE_GETVAL (dest, bounds);
+
+	for (int i=0; i < src_vtable->rank; i++) {
```
lewurm: being a bit pedantic here, but we do `for (int i = 0; [...])` usually

PR #13186, opened 2019-02-25
[runtime] Optimize array_set_value_impl slow path

alexanderkyte:
This PR makes our array_set_value_impl slow path much faster. Benchmarks included. 
It fixes the second of two problems identified here: https://github.com/mono/mono/issues/13133
 

Before:
image

After:
untitled

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -243,7 +243,7 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, 
 {
 	MonoClass *ac, *vc, *ec;
 	gint32 esize, vsize;
-	gpointer *ea, *va;
+	gpointer *ea = NULL, *va = NULL;
 
 	guint64 u64 = 0;
 	gint64 i64 = 0;
@@ -261,12 +261,15 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, 
 
 	ac = mono_handle_class (arr);
 	ec = m_class_get_element_class (ac);
-
 	esize = mono_array_element_size (ac);
-	ea = (gpointer*)mono_array_handle_pin_with_size (arr, esize, pos, &arr_gchandle);
 
 	if (mono_class_is_nullable (ec)) {
-		mono_nullable_init_from_handle ((guint8*)ea, value, ec);
+		MONO_ENTER_NO_SAFEPOINTS;
+		ea = (gpointer*) mono_array_addr_with_size_internal (MONO_HANDLE_RAW (arr_handle), esize, pos);
+		if (!MONO_HANDLE_IS_NULL (value_handle))
+			va = (gpointer*) mono_object_unbox_internal (MONO_HANDLE_RAW (value_handle));
+		mono_nullable_init_unboxed ((guint8*)ea, va, ec);
+		MONO_EXIT_NO_SAFEPOINTS;
 		goto leave;
 	}
 
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -239,7 +239,7 @@ ves_icall_System_Array_SetValueImpl (MonoArrayHandle arr, MonoObjectHandle value
 }
 
 static void
-array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, MonoError *error)
+array_set_value_impl (MonoArrayHandle arr_handle, MonoObjectHandle value_handle, guint32 pos, MonoError *error)
 {
 	MonoClass *ac, *vc, *ec;
 	gint32 esize, vsize;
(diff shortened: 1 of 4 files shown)


mono/metadata/icall.c
```diff
@@ -248,29 +248,32 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos,
 	guint64 u64 = 0;
 	gint64 i64 = 0;
 	gdouble r64 = 0;
+	gboolean castOk = FALSE;
 
-	uint32_t arr_gchandle = 0;
-	uint32_t value_gchandle = 0;
+	MONO_ENTER_NO_SAFEPOINTS;
```
lambdageek: I think this no safepoints region is too big.  until line 269 you don't need it.

There is a cost to "no safepoints" regions - they increase how long the GC has to wait for threads to suspend.  It is better to keep them small.

Additionally, the runtime will assert if try to switch to GC Safe, or invoke managed code, or try to allocate something while in a no safepoints region.

mono/metadata/icall.c
```diff
@@ -248,29 +248,32 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos,
...
+	ac = mono_handle_class (arr_handle);
 	ec = m_class_get_element_class (ac);
 
 	esize = mono_array_element_size (ac);
-	ea = (gpointer*)mono_array_handle_pin_with_size (arr, esize, pos, &arr_gchandle);
+	ea = (gpointer*) mono_array_addr_with_size_internal (arr, esize, pos);
 
 	if (mono_class_is_nullable (ec)) {
-		mono_nullable_init_from_handle ((guint8*)ea, value, ec);
+		mono_nullable_init_from_handle ((guint8*)ea, value_handle, ec);
```
lambdageek: It maybe worth modifying / making a variant of `mono_nullable_init_from_handle` that takes two handles and a destination offset.  (if you look inside `mono_nullable_init_from_handle`, you can see that it creates a global gchandle for `value_handle`, which defeats some of the speedup you're going for here)

mono/metadata/icall.c
```diff
@@ -248,29 +248,32 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos,
...
 	else
 		vc = NULL;
 
-	ac = mono_handle_class (arr);
+	ac = mono_handle_class (arr_handle);
 	ec = m_class_get_element_class (ac);
 
 	esize = mono_array_element_size (ac);
-	ea = (gpointer*)mono_array_handle_pin_with_size (arr, esize, pos, &arr_gchandle);
+	ea = (gpointer*) mono_array_addr_with_size_internal (arr, esize, pos);
```
lambdageek: maybe we can move this call closer to each place where `ea` is used and put smaller safepoint regions around each of those occurrences.

mono/metadata/icall.c
```diff
@@ -248,29 +248,32 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos,
...
 
 	if (mono_class_is_nullable (ec)) {
-		mono_nullable_init_from_handle ((guint8*)ea, value, ec);
+		mono_nullable_init_from_handle ((guint8*)ea, value_handle, ec);
 		goto leave;
 	}
 
-	if (MONO_HANDLE_IS_NULL (value)) {
+	if (MONO_HANDLE_IS_NULL (value_handle)) {
```
lambdageek: for example here you could do:
```c
if (MONO_HANDLE_IS_NULL (value_handle)) {
  MONO_ENTER_NO_SAFEPOINTS;
  ea = (gpointer*) mono_array_addr_with_size_internal (MONO_HANDLE_RAW (arr_handle), esize, pos);
  mono_gc_bzero_atomic (ea, esize);
  MONO_EXIT_NO_SAFEPOINTS;
  goto leave;
}
```

mono/metadata/icall.c
```diff
@@ -331,21 +334,17 @@ array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos,
...
 		if (!castOk)
 			INVALID_CAST;
-		MONO_HANDLE_ARRAY_SETREF (arr, pos, value);
+		MONO_HANDLE_ARRAY_SETREF (arr_handle, pos, value_handle);
 		goto leave;
 	}
 
 	if (castOk) {
-		va = (gpointer*)mono_object_handle_pin_unbox (value, &value_gchandle);
+		va = (gpointer*) mono_object_unbox_internal (value);
```
lambdageek: the contents of this `if (castOk) { ... }` block would make another good no safepoints region (together with an assignment to `ea` from `arr_handle`.

alexanderkyte:
The new times as of the no-safepoint-region-shortening changes.

 untitled

luhenry:
@monojenkins build failed

alexanderkyte:
Ah, I wasn't running those xunit tests locally when I confirmed corlib worked. Looks like I broke two of them. Will fix. Seems like I may have broken an error handling path around enum types.

mono/metadata/object.c
```diff
@@ -7197,6 +7228,24 @@ mono_object_handle_isinst_mbyref (MonoObjectHandle obj, MonoClass *klass, MonoEr
 
 	MonoObjectHandle result = MONO_HANDLE_NEW (MonoObject, NULL);
 
+	if (MONO_HANDLE_IS_NULL (obj))
+		goto leave;
+
+	gboolean success = mono_object_handle_isinst_mbyref_raw (obj, klass, error);
```
jaykrell: Compiles for C++?

alexanderkyte: Yeah, I saw the failure. And that's true, but I think that's a really wide `if` statement to put on one line.

alexanderkyte:
@monojenkins build failed

PR #13187, opened 2019-02-25
[corlib] Add a RuntimeTypeHandle::GetCorElementType () icall, use it to remove a few other small icalls.

vargaz:


mono/metadata/icall.c
```diff
@@ -2692,18 +2692,15 @@ ves_icall_RuntimeTypeHandle_GetBaseType (MonoReflectionTypeHandle ref_type, Mono
...
 
-MonoBoolean
-ves_icall_RuntimeTypeHandle_IsPrimitive (MonoReflectionTypeHandle ref_type, MonoError *error)
-{
-	MonoType *type = MONO_HANDLE_GETVAL (ref_type, type);
-	return (!type->byref && (((type->type >= MONO_TYPE_BOOLEAN) && (type->type <= MONO_TYPE_R8)) || (type->type == MONO_TYPE_I) || (type->type == MONO_TYPE_U)));
+	if (type->byref)
+		return MONO_TYPE_BYREF;
+	else
+		return (guint32)type->type;
```
lambdageek: Not sure if it matters, but https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/reflection/mdimport.cs doesn't have 0x55 `MONO_TYPE_ENUM`.   But I don't think we ever use that for any MonoType, right?

vargaz: I think its only used for custom attributes, but the managed code doesn't need it right now.

vargaz:

vargaz:
@monojenkins build failed

mono/metadata/icall.c
jaykrell: `return type->byref ? MONO_TYPE_BYREF : (guint32)type->type;`?

vargaz:
@monojenkins build failed

PR #13189, opened 2019-02-25
[corlib]: Use `Guid.NewGuid()` implementation from CoreFx.

baulig:
The corefx implementation of `Guid.NewGuid()` uses the internal `Interop.GetRandomBytes(byte*, int)`, which uses `RNGCryptoServiceProvider` directly without using `CryptoConfig`.


baulig:
@monojenkins commit csproj

baulig:
@monojenkins build failed

mcs/class/corlib/System/Guid.MonoTouch.cs
```diff
@@ -1,5 +1,3 @@
-#if MONOTOUCH && FULL_AOT_RUNTIME
```
marek-safar: We should rename the file

baulig: How about `Guid.CommonCrypto.cs`?

baulig: I moved it into the `CommonCrypto/` directory and renamed it into `CommonCrypto/Guid.cs`.

Issue #12422, opened 2019-01-14
Runtime incorrectly resolving overloads that differ by modopt 


jaredpar:
## Steps to Reproduce
This was found attempting to get the C# semantic tests running on Mono

1. Clone https://github.com/jaredpar/roslyn
1. Run `./build.sh --restore --build`
1. Run `cd src/Compilers/CSharp/Test/Emit`
1. Run `dotnet msbuild /t:Test /p:TestRuntime=Mono` 

### Actual Behavior
This will run the Emit unit tests and eventually Mono will fail executing the test ` TestCallMethodsWithLeastCustomModifiers`.  The output of the test will be 11 and the expectation is 51. 

Note: this behavior is flaky. It occurs roughly 5% of the time in our test runs. 

### Expected Behavior
The test passes

## On which platforms did you notice this

This was discovered on Ubuntu 18.04. I'm using the latest nightly mono as described by the Download page

https://www.mono-project.com/download/nightly/#download-lin

```
Mono JIT compiler version 5.19.0.290 (tarball Tue Sep  4 10:59:48 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(4)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

## Full Details

This is a test of the C# compilers behavior when invoking a method that has many overloads where the only difference is the number of `modopt` on each signature. The C# compiler will choose the one with the least number of `modopt` signatures (historical reasons). 

In this case the compiler is invoking the overload of `M` with the following signature:

``` il
  IL_0014:  callvirt   instance uint8 modopt([mscorlib]System.Runtime.CompilerServices.IsConst) [ModoptTests]Metadata.LeastModoptsWin::M(uint8,
                                                                                                                                         uint8)
```

The Mono runtime though is not resolving this correctly at runtime. Instead it's picking the overload with the following signature:

``` il
.method public hidebysig newslot virtual 
        instance uint8 modopt([mscorlib]System.Runtime.CompilerServices.IsConst) modopt([mscorlib]System.Runtime.CompilerServices.IsLong) 
        M(uint8 t,
          uint8 v) cil managed
```

[...]

PR #13191, opened 2019-02-26
[draft] custom modifier strict equiv checking and inflation

lambdageek:
**DRAFT WORK IN PROGRESS**

1. Always (if env sets `ALEKSEY_STRICT=1`) use strict (custom-modifier-respecting) equivalence checking for `MonoType`s
2. When you substitute a type for a generic parameter if both the type and the parameter have custom modifiers, you need to append the two sets of modifiers (Roslyn has tests that check for this).
3. Inflate generic parameters that appear in custom modifiers (thanks C++/CLI!)  ``void .CL1`1.Test(!0 modopt(System.Nullable`1<!0>))``

This required adding a second representation for `MonoTypeWithModifiers`.  In the normal *cmods* representation there's a single image and a list of modifiers is represented as a token in that image.  This is what you get from metadata.  In the aggregate *amods* representation, you have a just a bag of `MonoType*`s.

The aggregate representation is due to the need to inflate custom modifiers as in the last bullet point.  If it weren't for that, we could use a simpler representation: a list of pairs of `MonoImage*` image and `uint32_t` token.  But if we need to inflate, we get back types that don't have a token in some table.

---

Status:

1. There's probably leaks here - in a few places I'm pretty sure I create fresh help-allocated types and don't `free` them.
2. There's also probably types being allocated into the wrong images.   In cases where a type has custom modifiers from multiple other images, I think we probably need to allocate the type into a `MonoImageSet` mempool, not just into a MonoImage mempool.
3. It's passing the tests from #12422 and #10945 - in fact, it's also passing tests that are disabled in that tree, like `ModifiersWithConstructedType_01`.

Overall: not ready yet.  But I want to see if it broke anything.

--- 

Tasks:

- [x] Update all call sites of `mono_type_get_cmods` to also handle aggregate modifiers.
- [x] One instance of the above: update the AOT image format to save/restore aggregate modifiers.

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -72,6 +72,9 @@ static gboolean mono_class_is_subclass_of_internal (MonoClass *klass, MonoClass 
 GENERATE_GET_CLASS_WITH_CACHE (valuetype, "System", "ValueType")
 GENERATE_TRY_GET_CLASS_WITH_CACHE (handleref, "System.Runtime.InteropServices", "HandleRef")
 
+// define to print types whenever custom modifiers are appended during inflation
+#undef DEBUG_INFLATE_CMODS
+
 static
 MonoImage *
 mono_method_get_image (MonoMethod *method)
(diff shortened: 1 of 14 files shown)


lambdageek:
The cleaned up version of this stuff is #13320

PR #13196, opened 2019-02-26
[wasm][ci]Update to latest JavaScript (engine) Version Updater

kjpou1:
Update to version `1.4.0`


kjpou1:
Closing as the update relies on node.js >= 10.  We need to look at another approach so as to react more quickly to CI build problems such as engines not installing correctly.

PR #13200, opened 2019-02-26
Initial support for Fuchsia

migueldeicaza:
Compile like this:

if you have Zircon installed, with the prebuilts on "../zircon", use this to compile:
```
CC="`pwd`/../zircon/prebuilt/downloads/clang/bin/clang-9 -nostdlib -O2 -g -fdebug-prefix-map=/cvs/fuchsia/zircon=. -finline -fno-common -ffunction-sections -fdata-sections -no-canonical-prefixes -march=x86-64 -mcx16 --target=x86_64-fuchsia -fPIC -D_ALL_SOURCE=1 -DENABLE_DRIVER_TRACING=1 -fasynchronous-unwind-tables -fno-omit-frame-pointer  -std=c11 -I`pwd`/../zircon/build-x64-clang/sysroot/include/ -I`pwd`/../zircon/system/public" LDFLAGS="-L`pwd`/../zircon/build-x64-clang/sysroot/lib/ -L`pwd`/../zircon/build-x64-clang/sysroot/debug" ./configure --host=x86_64-google-fuchsia
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -402,7 +402,7 @@ case "$host" in
 		;;
 	*-*-fuchsia*)
 		AC_DEFINE(DISABLE_PORTABILITY,1,[Disable the io-portability layer])
-		AC_DEFINE(TARGET_FUCHSIA,1,[Targeting the Fuchsia platform])
+		AC_DEFINE(HOST_FUCHSIA,1,[Targeting the Fuchsia platform])
 		fuchsia=true
 		with_tls=pthread
 		with_sigaltstack=yes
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -400,6 +400,18 @@ case "$host" in
 		use_sigposix=yes
 		with_sigaltstack=no
 		;;
+	*-*-fuchsia*)
+		AC_DEFINE(DISABLE_PORTABILITY,1,[Disable the io-portability layer])
+		AC_DEFINE(TARGET_FUCHSIA,1,[Targeting the Fuchsia platform])
```
lambdageek: I think this should be `HOST_FUCHSIA`.  host is the machine mono runs on, target is the machine the aot compiler produces output for.

PR #13202, opened 2019-02-26
Revert "[Mono.Security]: minor `CryptoConvert` changes to make it more linker-friendly."

baulig:
Reverts mono/mono#13190

PR #13190, opened 2019-02-25
[Mono.Security]: minor `CryptoConvert` changes to make it more linker-friendly.

baulig:
* Add new internal `CryptoConvert.TryImportCapiPrivateKeyBlob()` method that's using `RSAManaged` directly to make it more linker-friendly.

* Use it in `AssemblyName.IsPublicKeyValid`.

PR #13204, opened 2019-02-26
[Mono.Security]: minor `CryptoConvert` changes to make it more linker-friendly.

baulig:
* Add new internal `CryptoConvert.TryImportCapiPrivateKeyBlob()` method that's using `RSAManaged` directly to make it more linker-friendly.

* Also add new internal `CryptoConvert.TryImportCapiPublicKeyBlob()`.

* Use it in `AssemblyName.IsPublicKeyValid`.


baulig:
Fixed version of https://github.com/mono/mono/pull/13190, which we had to revert in https://github.com/mono/mono/pull/13202.  For tests, see my comment in the revert PR https://github.com/mono/mono/pull/13202#issuecomment-467663535.

baulig:
@monojenkins build failed

PR #13210, opened 2019-02-27
Update output from console-output test.

lateralusX:
Current output from console-output test is just Error and Out (indicating the different console output streams). When seen in log files the error output string can be misleading. Fix makes it clearer that the output is originating from a test and not an “error”.


lateralusX:
@monojenkins rebuild failed

Issue #13224, opened 2019-02-28
iOS SDKs test app build doesn't work when not in sdks/ios directory


akoeplinger:
## Steps to Reproduce

1. Build iOS SDKs
2. Run `make -C sdks/ios build-ios-sim-corlib` from the mono repo root
3. It fails

### Current Behavior

```
Unhandled Exception:
System.IO.FileNotFoundException: Could not find file "/Users/alexander/dev/mono/runtime/Info.plist.in"
File name: '/Users/alexander/dev/mono/runtime/Info.plist.in'
  at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, System.Int32 bufferSize, System.Boolean anonymous, System.IO.FileOptions options) [0x0019e] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
  at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, System.Int32 bufferSize, System.IO.FileOptions options, System.String msgPath, System.Boolean bFromProxy, System.Boolean useLongPath, System.Boolean checkHost) [0x00000] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
  at (wrapper remoting-invoke-with-check) System.IO.FileStream..ctor(string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare,int,System.IO.FileOptions,string,bool,bool,bool)
  at System.IO.StreamReader..ctor (System.String path, System.Text.Encoding encoding, System.Boolean detectEncodingFromByteOrderMarks, System.Int32 bufferSize, System.Boolean checkHost) [0x00067] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
  at System.IO.StreamReader..ctor (System.String path, System.Text.Encoding encoding, System.Boolean detectEncodingFromByteOrderMarks, System.Int32 bufferSize) [0x00000] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
  at System.IO.StreamReader..ctor (System.String path, System.Boolean detectEncodingFromByteOrderMarks) [0x0000d] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
  at System.IO.StreamReader..ctor (System.String path) [0x00000] in <b60a4d3991c34f6d9d6ecb26e93ea0c3>:0
[...]

PR #13230, opened 2019-02-28
[sdks] Use abspath instead of PWD on iOS app build

akoeplinger:
PWD is the wrong directory when e.g. you're doing "make -C sdks/ios" from the repo root.

Fixes https://github.com/mono/mono/issues/13224

### DIFF
--- a/sdks/ios/Makefile
+++ b/sdks/ios/Makefile
@@ -77,8 +77,8 @@ TEST_ASSEMBLIES = $(BCL_DIR)/mscorlib.dll \
 	test-runner.exe
 
 build-ios-sim-%: appbuilder.exe test-runner.exe runtime/runtime $($*_ASSEMBLIES)
-	mono appbuilder.exe --target ios-sim64 --mono-sdkdir $(PWD)/../out --appdir $(PWD)/bin/ios-sim/test-$*.app --runtimedir $(PWD)/runtime --builddir obj/ios-sim/test-$*.app --sysroot $(SYSROOT) --signing-identity - --bundle-executable test-$* --bundle-identifier com.xamarin.mono.ios.test-$* --bundle-name test-$* $(patsubst %,-r %,$(TEST_ASSEMBLIES) $($*_ASSEMBLIES))
-	mkdir -p $(PWD)/bin/ios-sim/test-$*.app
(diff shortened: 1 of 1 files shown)


sdks/ios/Makefile
```diff
@@ -77,8 +77,8 @@ TEST_ASSEMBLIES = $(BCL_DIR)/mscorlib.dll \
 	test-runner.exe
 
 build-ios-sim-%: appbuilder.exe test-runner.exe runtime/runtime $($*_ASSEMBLIES)
-	mono appbuilder.exe --target ios-sim64 --mono-sdkdir $(PWD)/../out --appdir $(PWD)/bin/ios-sim/test-$*.app --runtimedir $(PWD)/runtime --builddir obj/ios-sim/test-$*.app --sysroot $(SYSROOT) --signing-identity - --bundle-executable test-$* --bundle-identifier com.xamarin.mono.ios.test-$* --bundle-name test-$* $(patsubst %,-r %,$(TEST_ASSEMBLIES) $($*_ASSEMBLIES))
-	mkdir -p $(PWD)/bin/ios-sim/test-$*.app
```
vargaz: make -C is supposed to change the working directory:
"Change to directory dir before reading the makefiles or doing anything else."

lewurm: not really. you can use `$(CURDIR)` instead:
```shell
$ pwd
/tmp
$ mkdir -p deep/dir/path
$ cat > deep/dir/path/Makefile
all:
	echo $(PWD)
	echo $(CURDIR)
$ make -C deep/dir/path
echo /tmp
/tmp
echo /private/tmp/deep/dir/path
/private/tmp/deep/dir/path
```

rolfbjarne: PWD is an environment variable, and it seems it's not updated automatically when the current directory changes.

`$(CURDIR)` is a make variable which reflects the current directory and could be used instead: `Set to the absolute pathname of the current working directory (after all -C options are processed, if any).`

akoeplinger: Yeah, I didn't try whether passing the relative dirs to `appbuilder.exe` works (since `make` changes the working dir) but `abspath` works and fixes the issue :)

PR #13245, opened 2019-03-01
[netcore] Implements threads.

vargaz:
There is only one thread object, since there are no domains.


mono/metadata/domain.c
```diff
@@ -709,8 +709,13 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
 	mono_defaults.thread_class = mono_class_load_from_name (
                 mono_defaults.corlib, "System.Threading", "Thread");
 
+#ifdef ENABLE_NETCORE
```
jaykrell: It'd be nice to avoid `ifdef` and have one runtime that worked either way.

mcs/class/System.Private.CoreLib/System.Threading/Thread.cs
```diff
@@ -4,16 +4,80 @@
...
 {
+	//
+	// Under netcore, there is only one thread object per thread
+	//
+	[StructLayout (LayoutKind.Sequential)]
 	partial class Thread
 	{
-		string name;
-		InternalThread internalThread;
+#pragma warning disable 169, 414, 649
```
marek-safar: We cannot do that, the shared copy will get fields as well and we cannot guarantee the order.

mcs/class/System.Private.CoreLib/System.Threading/Thread.cs
```diff
@@ -159,33 +226,97 @@ public static void SpinWait (int iterations)
...
 		}
 
 		public void Start (object parameter)
 		{
+			m_start_arg = parameter;
+			StartInternal (this);
+		}
+
+		// Called from the runtime
+		public void StartCallback () {
```
marek-safar: Should be internal

vargaz:
@monojenkins build failed

PR #13246, opened 2019-03-01
[mini] Inline Selector.GetHandle for tvOS and watchOS assemblies as well.

rolfbjarne:


luhenry:
@monojenkins backport 2018-10
@monojenkins backport 2019-02

Issue #8142, opened 2018-04-10
Awaiting DelegatingHandler base.SendAsync throws a Null exception


DanielCauser:
## Steps to Reproduce

[InnerHandler is null](https://github.com/mono/mono/blob/e90dee3d0ca28518af81b46c11c1ad1cdaea239d/mcs/class/System.Net.Http/System.Net.Http/DelegatingHandler.cs#L76)

1. Create a class that inherits from DelegatingHandler
2. Override the SendAsync method
3. Inside the overrided method, await for  base.SendAsync
4. It throws an exception

 https://github.com/TorontoMobileDevelopers/HttpTracer/blob/master/src/HttpTracer/HttpTracerHandler.cs#L47

<img width="1013" alt="screen shot 2018-04-09 at 9 55 59 pm" src="https://user-images.githubusercontent.com/7211075/38532039-f39a6c26-3c40-11e8-9c82-f5f8bb5a2691.png">

## Current Behavior

I can't await the base method, it is throwing an exception.

## Expected Behavior

I'm expecting to be able to await the base method and be able to work with the response

### On which platforms did you notice this

[x] macOS
[ ] Linux
[x] Windows

**Version Used**:

Mono Debugging for Visual Studio   4.9.11-pre (71eb098)
Support for debugging Mono processes with Visual Studio.

### Stacktrace

```
 at System.Net.Http.DelegatingHandler.SendAsync (System.Net.Http.HttpRequestMessage request, System.Threading.CancellationToken cancellationToken) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.9.1.24/src/Xamarin.iOS/mcs/class/System.Net.Http/System.Net.Http/DelegatingHandler.cs:76 
  at HttpTracer.HttpTracerHandler.<>n__0 (System.Net.Http.HttpRequestMessage request, System.Threading.CancellationToken cancellationToken) <0x1325e2950 + 0x0007a> in <45fa7ebf3ce14963aa21f1f83ec864d3>:0 
  at HttpTracer.HttpTracerHandler+<SendAsync>d__1.MoveNext () [0x0009a] in C:\HttpTracer\src\HttpTracer\HttpTracerHandler.cs:47 
```

marek-safar:
Although this is not ideal, it's inline with .net behaviour

marek-safar:
Oh, I was wrong .net actually throws here

`throw new InvalidOperationException(SR.net_http_handler_not_assigned);`

[...]

PR #13252, opened 2019-03-01
Throw InvalidOperationException if inner handler is not assigned in DelegatingHandler

MaximLipnin:
Mono uses DelegatingHandler class legacy implementation on Windows. The PR aligns its behavior with corefx implementation used on macOS/Unix.

Fixes https://github.com/mono/mono/issues/8142


MaximLipnin:
@monojenkins build failed

PR #13262, opened 2019-03-01
[crash] Detect when 0 attached threads, fix hang/double-fault.

alexanderkyte:
When we have 0 threads, most of the bookkeeping code goes a bit off the rails. Rather than hanging (OSX doesn't deliver double fault nowadays), this makes us exit cleanly. 

After: 

```
=================================================================
	Telemetry Dumper:
=================================================================
No threads attached to runtime.

=================================================================
	External Debugger Dump:
=================================================================
```

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -6430,6 +6430,13 @@ mono_threads_summarize_execute (MonoContext *ctx, gchar **out, MonoStackHash *ha
 	MonoNativeThreadId current = mono_native_thread_id_get ();
 	gboolean this_thread_controls = summarizer_state_init (&state, current, &current_idx);
 
+	if (state.nthreads == 0) {
+		if (!silent)
+			g_async_safe_printf("No threads attached to runtime.\n");
+		memset (&state, 0, sizeof (state));
+		return FALSE;
+	}
+
 	if (this_thread_controls) {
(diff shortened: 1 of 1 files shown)


Therzok:
Would this fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/802629 ?

alexanderkyte:
Perhaps. That crash seemed to be related to shutdown. Around shutdown we have a ton of references that may become null. Crash reporting for crashes on shutdown is problematic and will likely always be. 

If you’ve got a reproduction for that bug, please ping me on Slack and I can see what it’s breaking and make it fail gracefully. But a shutdown failure will probably remain a failure. 

For the record, it should be a double fault with a full lldb stacktrace. OSX has started turning double faults into hangs as of a recent release, which is why the situation is the way that it is.

alexanderkyte:
@monojenkins build failed

luhenry:
@alexanderkyte what's the status of this PR?

luhenry:
@alexanderkyte shouldn't that be backported to `2018-10` and `2019-02`?

alexanderkyte:
@monojenkins backport 2019-02

alexanderkyte:
@monojenkins backport 2018-10

PR #13287, opened 2019-03-04
[linker] Fix monotouch tests which need Xamarin.iOS.dll

akoeplinger:
The linker couldn't resolve Xamarin.iOS.dll which System.Net.Http references now.


akoeplinger:
@monojenkins backport 2019-02

PR #13288, opened 2019-03-04
[netcore] Ongoing work.

vargaz:
* Implement all CustomAttributeExtensions methods.
* Add stubs for __ComObject which is needed by the runtime.

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/__ComObject.cs
+++ b/mcs/class/System.Private.CoreLib/System/__ComObject.cs
@@ -1,0 +1,10 @@
+namespace System
+{
+	internal class __ComObject
+	{
+		__ComObject ()
+		{
+			throw new NotSupportedException ();
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


mcs/class/System.Private.CoreLib/System/__ComObject.cs
```diff
@@ -0,0 +1,10 @@
+namespace System
+{
+	internal class __ComObject
```
marek-safar: Is it really needed?

vargaz: The runtime asserts on it.

marek-safar: Can we remove it? Otherwise, we'll have to teach linker etc which might be even more work

vargaz: Its needed for COM support, we can get rid of it, but its more work than adding this one class.

PR #13305, opened 2019-03-04
Introducing `HttpClient.CreateDefaultHandler()`.

baulig:
* The default `HttpClient` constructor now calls a static method `CreateDefaultHandler()`;
   we provide a platform-specific implementation of it.

* Replaced `HeaderUtils` with `PlatformHelper` and removed the `LEGACY_HTTPCLIENT` conditional.

* Added a new `socketshandler.sources` file to keep the sources in one place.

* Use a reflection check in the tests instead of the `LEGACY_HTTPCLIENT` conditional.

* This allows us to also get rid of the `SOCKETSHTTPHANDLER` define.


baulig:
Requires corefx changes: https://github.com/mono/corefx/pull/264.

baulig:
@monojenkins commit csproj

baulig:
@monojenkins commit csproj

mcs/class/System.Net.Http/HttpClient.DefaultHandler.cs
```diff
@@ -0,0 +1,10 @@
+namespace System.Net.Http
```
marek-safar: Not sure default is the right word, each profile will have a different default. This one is I guess WebRequestHandler, right?

Suchiman:
Should mono maybe adopt the trick employed by corefx to use `HttpClientHandler` as a mere wrapper around the real `HttpMessageHandler`? https://github.com/dotnet/corefx/blob/fc9629cae7bfcda2196b24ba5127265efce1716a/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Unix.cs#L24-L37

That would allow code that constructs a `HttpClientHandler` manually for various reasons to work, e.g. https://github.com/dotnet/wcf/blob/7f0e6db26da8eaf790606547d08dfb2806165549/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs#L336-L385

I was thinking about this while doing http work in blazor where hardcoded calls to `HttpClientHandler` are unfortunately resulting in errors.

mcs/class/System.Net.Http/HttpClient.DefaultHandler.cs
baulig: There is exactly one "default" that is used by the parameterless ("default") constructor - or in other words, the default constructor needs to call "something".  This PR allows this "something" to be profile-specific.  Does this answer your question?

baulig:
> Should mono maybe adopt the trick employed by corefx to use `HttpClientHandler` as a mere wrapper around the real `HttpMessageHandler`? https://github.com/dotnet/corefx/blob/fc9629cae7bfcda2196b24ba5127265efce1716a/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Unix.cs#L24-L37
> 
> That would allow code that constructs a `HttpClientHandler` manually for various reasons to work, e.g. https://github.com/dotnet/wcf/blob/7f0e6db26da8eaf790606547d08dfb2806165549/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs#L336-L385
> 
> I was thinking about this while doing http work in blazor where hardcoded calls to `HttpClientHandler` are unfortunately resulting in errors.

We are already doing this.  The PR is merely about what is used by the parameterless "default" constructor.

Ideally, we should evaluate whether we could use `HttpClientHandler` for everything, but at the moment some of our mobile profiles (android, web assembly) use a custom `HttpMessageHandler`.

Look at for instance https://github.com/mono/mono/blob/master/mcs/class/System.Net.Http/System.Net.Http/HttpClient.android.cs or https://github.com/mono/mono/blob/master/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs.

akoeplinger:
@monojenkins commit csproj

PR #13307, opened 2019-03-05
[wasm][ci]Update to latest JavaScript (engine) Version Updater

kjpou1:
- The purpose of this is to be able to respond to CI incidents faster for instance specific engine versions when JavaScript engines do not update/install correctly.
- This will install from a forked version of the jsvu repo: https://github.com/GoogleChromeLabs/jsvu
- Forked version rests here:  https://github.com/kjpou1/jsvu/tree/mono-wasm
   - The forked version removes the Node >= 10.0 dependency but brings in the newest 1.4.0 version modifications.
- This should be temporary until we get another solution i.e. a managed version of the engine updater.


kjpou1:
@monojenkins build failed

lewing:
@monojenkins build failed

PR #13308, opened 2019-03-05
Get fully qualified name of ModuleBuilder

### LINKED ISSUE #8503: System.Reflection.Emit: ModuleBuilder.FullyQualifiedName does not return a fully qualified name
`ModuleBuilder.FullyQualifiedName` does not include an absolute module path as it does on the CLR. As per [the MSDN documentation for the `ModuleBuilder.FullyQualifiedName` property](https://msdn.microsoft.com/en-us/library/system.reflection.emit.modulebuilder.fullyqualifiedname.aspx), it should:

> Gets a String representing the fully qualified name **and path** to this module.

## Steps to reproduce

```csharp
using System;
using System.IO;
using System.Reflection;
using System.Reflection.Emit;

class Program
{
    static void Main()
    {
        var thisModule = MethodInfo.GetCurrentMethod().Module;

        Console.WriteLine(Path.IsPathRooted(thisModule.FullyQualifiedName));

        var assemblyBuilder = AppDomain.CurrentDomain.DefineDynamicAssembly(new AssemblyName("DynamicAssembly"), AssemblyBuilderAccess.RunAndSave);
        var moduleBuilder = assemblyBuilder.DefineDynamicModule("DynamicModule", "DynamicAssembly.dll");

        Console.WriteLine(Path.IsPathRooted(moduleBuilder.FullyQualifiedName));
    }
}
```

### Current behavior

When run with Mono, the program outputs:

```
True
False
```

That is, `ModuleBuilder.FullyQualifiedName` does not return something that includes an absolute file path.

### Expected behavior

The program should produce the same output as when run on Windows / the CLR:

```
True
True
```

That is, `ModuleBuilder.FullyQualifiedName` should return a fully qualified name (i.e. one including an absolute file path).

## Platform on which I noticed this

[ ] macOS
[X] Bash on Ubuntu on Windows subsystem of Windows 10 (Ubuntu 16.04.4 LTS)
[ ] Windows

**Version used**:

Mono 5.10.1.47

MaximLipnin:
Fixes #8503

### DIFF
--- a/mcs/class/corlib/System.Reflection.Emit/AssemblyBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/AssemblyBuilder.cs
@@ -917,7 +917,7 @@ namespace System.Reflection.Emit
 			ModuleBuilder mainModule = null;
 			if (modules != null) {
 				foreach (ModuleBuilder module in modules)
-					if (module.FullyQualifiedName == assemblyFileName)
+					if (module.FileName == assemblyFileName)
 						mainModule = module;
 			}
 			if (mainModule == null)
--- a/mcs/class/corlib/System.Reflection.Emit/ModuleBuilder.cs
+++ b/mcs/class/corlib/System.Reflection.Emit/ModuleBuilder.cs
@@ -177,7 +177,19 @@ namespace System.Reflection.Emit {
 			Console.Error.WriteLine ("WARNING: {0}", message);
 		}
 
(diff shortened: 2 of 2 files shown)


marek-safar:
@MaximLipnin this broke existing tests

MaximLipnin:
@marek-safar I'm looking at them

MaximLipnin:
@monojenkins build failed

PR #11789, opened 2018-11-26
 [safepoint] Use LLVM pass to insert safepoint 

luhenry:
Depends on https://github.com/mono/llvm/pull/35

PR #13313, opened 2019-03-05
[do-not-merge] Experiment with weak linkage 

alexanderkyte:
Let's get a CI build of this with tests. Local build of the iOS SDK failed right at the end for submodule reasons.


mono/mini/mini-llvm.c
```diff
@@ -2959,6 +2959,7 @@ emit_gc_safepoint_poll (MonoLLVMModule *module)
 	sig = LLVMFunctionType0 (LLVMVoidType (), FALSE);
 	func = mono_llvm_get_or_insert_gc_safepoint_poll (lmodule);
 	mono_llvm_add_func_attr (func, LLVM_ATTR_NO_UNWIND);
+	LLVMSetLinkage (func, LLVMWeakAnyLinkage);
```
lambdageek: I think we want `LLVMLinkOnceODRLinkage` https://llvm.org/docs/LangRef.html#linkage-types - we want all instances of the safepoint function to be considered equivalent for linking and only one copy kept (linkonce) and we want to allow inlining the body of it by assuming that all copies have the same implementation (odr)

alexanderkyte: Interesting, ODR to me seemed to be defined as "only merge if the implementations are identical". I wanted to always merge, assuming that they're always identical. 

From your link:

```
 Other languages, such as C++, ensure that only equivalent globals are ever merged (the “one definition rule” — “ODR”). Such languages can use the linkonce_odr and weak_odr linkage types to indicate that the global will only be merged with equivalent globals. These linkage types are otherwise the same as their non-odr versions.
```

And weak over linkonce was chosen because

```
weak
“weak” linkage has the same merging semantics as linkonce linkage, except that unreferenced globals with weak linkage may not be discarded. This is used for globals that are declared “weak” in C source code.
```

I didn't want to discard the function. Though I suppose it shouldn't matter, since it should always be referenced at least once in managed code.

alexanderkyte: I think that because we need to use a patch / GOT slot, two different copies of this function would reference two different GOT slots, and be identified as different. Therefore if we did linkonce_odr, it wouldn't dedup the two. It would just treat it like `private`

lambdageek: @alexanderkyte 
1. I don't think they actually compare function bodies.  the `odr` flavor means that llvm optimizations get to assume that if they ever sees another function of the same name then it _would_have the same body.  therefore the body that they see _now_ is the only one they could ever expect - therefore it's safe to inline it and perform other optimizations on it.  without `odr`, they _can't_ do any optimizations because the body they see now may not be the final one that will get used.
2. actually because of the got slots I think we want private linkage - use it once per translation unit (and inline it if it's profitable), but never use it across translation units.

alexanderkyte: With weak: 
```
make[6]: *** [class/lib/testing_aot_full/Microsoft.CSharp.dll.so] Error 1
LLVMSymbolizer: error reading file: No such file or directory
#0 0x000055e593198ac5 (opt+0x137cac5)
#1 0x000055e593196bfe (opt+0x137abfe)
#2 0x000055e593196d22 (opt+0x137ad22)
#3 0x00007fc9460460e0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x110e0)
#4 0x000055e592c32477 (opt+0xe16477)
#5 0x000055e593102a7c (opt+0x12e6a7c)
#6 0x000055e592c53512 (opt+0xe37512)
#7 0x000055e592c535a3 (opt+0xe375a3)
#8 0x000055e592c53eb4 (opt+0xe37eb4)
#9 0x000055e59218cf1d (opt+0x370f1d)
#10 0x00007fc94538f2e1 __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x202e1)
#11 0x000055e5921ecb8a (opt+0x3d0b8a)
Stack dump:
0.	Program arguments: opt -f -O2 -disable-tail-calls -place-safepoints -spp-all-backedges -o class/lib/testing_aot_full/Mono.CSharp.dll_bitcode_tmp/temp.opt.bc class/lib/testing_aot_full/Mono.CSharp.dll_bitcode_tmp/temp.bc 
1.	Running pass 'Function Pass Manager' on module 'class/lib/testing_aot_full/Mono.CSharp.dll_bitcode_tmp/temp.bc'.
2.	Running pass 'Place Safepoints' on function '@System_Reflection_Emit_AssemblyBuilderExtensions_AddResourceFile_System_Reflection_Emit_AssemblyBuilder_object__'
Segmentation fault (core dumped)
AOT of image class/lib/testing_aot_full/Mono.CSharp.dll failed.
Mono Ahead of Time compiler - compiling assembly /mnt/jenkins/workspace/test-mono-pull-request-amd64-fullaot+llvm/mcs/class/lib/testing_aot_full/Mono.CSharp.dll
AOTID 395C680F-9DB7-2BA0-3F59-A204CC6E875D
```

I'll try weak_odr though.

vargaz: You can try naming the safepoint function differently, or making it internal or something.

alexanderkyte:
WeakODR seems to be working. Pushed to https://github.com/mono/mono/pull/11789

PR #13315, opened 2019-03-05
Enables CRLF normalization for .proj and .sln files and also renormalizes them

Anipik:
We have now a shared folder with the dotnet repos and dotnet repos have crlf normalization enables.
This difference leads to conflicts while applying changes from the dotnet repos to mono.

Mono repo no longer requires the crlf normalization to be disabled. This rule was from before we were generating the csproj files automatically

The second commit is done by running ```git add --renormalize .```
 
cc @marek-safar @akoeplinger


kg:
Github's diff viewer is useless for answering this question: did you update the relevant templates that we use to generate the project files? Csproj.tmpl etc? We should normalize those too.

Anipik:
> did you update the relevant templates that we use to generate the project files

I just changed .git attributes file in the root to remove line disabling the normalization. I checked the other .gitattribute  file and it was not present in any of the other ones.

> Csproj.tmpl etc?

I don't know what needs to get done here ? is there any other .tmpl I need to edit ?

akoeplinger:
I have a change locally that fixes the templates, I'll open a PR shortly.

Anipik:
>  I have a change locally that fixes the templates, I'll open a PR shortly.

Thanks. should I close this pr ?

akoeplinger:
Yeah :)

akoeplinger:
https://github.com/mono/mono/pull/13317

PR #13106, opened 2019-02-20
Make `System.dll` internals visible to `Mono.Android`.

baulig:

PR #13316, opened 2019-03-05
[2018-10] Make `System.dll` internals visible to `Mono.Android`.

alexischr:
Backport of https://github.com/mono/mono/pull/13106


alexischr:
this needs an apidiff so let's not push this in

alexischr:
@monojenkins commit apidiff

PR #13317, opened 2019-03-05
[gitattributes] Do CRLF normalization on sln/proj files

akoeplinger:
They can be used with native line endings.

We now have a shared folder with the dotnet repos and they have CRLF normalization enabled.
This difference leads to conflicts while applying changes from the dotnet repos to mono.


bcl.sln
```diff
@@ -1,6 +1,6 @@
 
-Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
```
kg: Just checking - did you make sure this works in VS? The .sln parser is notorious for being very bad and subtly broken, things like dependencies get dropped based on whitespace

akoeplinger: You mean VSMac or VSWin? for the latter there should be no change as the .sln will be checked out with CRLF on Windows.

kg: OK, wasn't sure whether our sln parser was bug-for-bug compatible

PR #13318, opened 2019-03-05
HttpClient for Web Assembly

baulig:


baulig:
@monojenkins commit csproj

baulig:
This needs to land on top of https://github.com/mono/mono/pull/13305.

baulig:
@monojenkins build failed

baulig:
Requires https://github.com/mono/corefx/pull/266.

mcs/class/System.Net.Http/wasm_System.Net.Http.dll.sources
```diff
@@ -1,2 +1,2 @@
-#include legacy.sources
+#include socketshandler.sources
```
marek-safar: The idea is to have really only wasm handler here (no sockethandler at all) but I guess that can be cleaned up by @kjpou1

baulig: We aren't quite there yet, the socket handler cannot be removed at this point.

baulig:
@kjpou1 I have a few questions about this:

1. Under which circumstances can `HttpClient.GetFallback` (in [`HttpClient.wasm.cs:46`](https://github.com/mono/mono/blob/c96df417d1c36a2f5adac42a67bdcece566cb354/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs#L51)) be used?
2. In that same file, what is `GetHttpMessageHandler` and where is it used?
3. Is there a specific reason why `WasmHttpMessageHandler` derives from `HttpMessageHandler` (and not `HttpClientHandler`)?  Is this a user-visible type or could we possibly rename it into `WasmHttpClientHandler`, deriving from `HttpClientHandler`?
4. Is there a way of making the code in `framework/WebAssembly.Net.Http` implement an internal interface from our `System.Net.Http.dll`?  I am thinking of `IMonoHttpClientHandler`.

Ideally, I would like to have the code in `WebAssembly.Net.Http` implement the internal `IMonoHttpClientHandler` API (instead of `HttpClientHandler` or `HttpMessageHandler`), then we could nicely integrate it with the current mechanism.  What do you think about this?

mcs/class/System.Net.Http/wasm_System.Net.Http.dll.sources
baulig: See my question to @kjpou1 - it's that fallback handler that's preventing the sockets handler from being removed.

baulig:
Adding to my list of questions:

5. What is `HttpClientHandler` supposed to do?  At the moment, it's using the sockets handler, but I suppose that's wrong and won't work.

lewing:
The answers as I understand them (I'm sure @kjpou1 will correct me if I'm wrong):
1. never in practice
2. I think it is just there to avoid doing the reflection every time.  Which brings up the issue that now that we've enabled full SRE that code no longer functions in full AOT because Type.GetType("blah") can't discover the calling assembly.
3. Because HttpMessageHandler is the abstract base for the http handlers:
```
System.Net.Http.CFNetworkHandler
System.Net.Http.DelegatingHandler
System.Net.Http.HttpClientHandler
System.Net.Http.NSUrlSessionHandler
System.Net.Http.SocketsHttpHandler
System.Net.Http.WinHttpHandler
```
And HttpClient would pick the handler based on platform settings.  I'm not sure we actually support all the options on HttpMessageHandler let alone HttpClientHandler.
4. Yes if we pull in the js interop stuff or make it visible.
5. There is no native socket abstraction in wasm so nothing but WasmHttpMessageHandler will work at all right now.

kjpou1:
@baulig 

Larry pretty much summed up everything.

One issue is the `GetHttpMessageHandler` this was used by client code to register a callback to a custom implementation.  Instead of registering and creating a new one each time they could just register this callback and provide the implementation by just calling `new HttpClient()`.  The Blazor team asked for this so they could provide their own implementation.  Set the callback up during platform intialization They should stop using this with this release or maybe the next by replacing their implementation with the one provided.  Taking this out may break user code but they should also start to use our delivered implementation.

We do not support all the options `HttpMessageHandler` nor `HttpClientHandler` because the browser API does not.  

Their is no browser socket support so what we do is bind the browser `Fetch` API.  The only real way to do that is by implementing the abstract class `HttpMessageHandler`.

mcs/class/System.Net.Http/wasm_System.Net.Http.dll.sources
kjpou1: So the `static HttpMessageHandler GetFallback(string message)` is what is preventing that?

kjpou1:
Let me add to point 1) `Under which circumstances can HttpClient.GetFallback (in HttpClient.wasm.cs:46) be used?`

There are actually two scenarios where a default `HttpClient` can be created where the defined behavior would ideally be a `PlatformNotSupported` exception.  

Those two are in the case mono wasm is being executed:

- By a JavaScript Engine where there are no browser API's.  
- The second would be in a `Node.js` environment.  In this case we would need another binding program for the `Node.js` API.  An exception would be Electron where wasm is being executed in the Renderer.   

Are those viable right now, probably not, but wanted to throw it out there.

baulig:
Does the fallback handler need to return an actual _handler_ (that's throwing `PlatformNotSupportedException`) or can the `GetFallback ()` method _itself_ just throw `PlatformNotSupportedException`?

kjpou1:
Martin.  That is a good question and probably we would need more feedback from others but if I were personally using this I would prefer a handler that would throw `PlatformNotSupported` exceptions.  

Isn't that what the defined behavior is for dotnet in these cases?

baulig:
My biggest concern right now is, is the following code supposed to work on web assembly:

```
   HttpClientHandler handler = new HttpClientHandler();
   handler.Credentials = ....

   HttpClient client = new HttpClient(handler);
```

If no, then do we provide a stubbed-out (`PlatformNotSupportedException`) version of the `HttpClientHandler` class or just omit it completely?

If yes, then I'd need something that implements `IMonoHttpClientHandler` (though I could build that bridge inside the `System.Net.Http.dll` if we can't make that internal type available to wasm).

kjpou1:
Yes, that is suppose to work on wasm.  That is why the `GetHttpMessageHandler` was introduced so people would not need to do that every time they wanted an `HttpClient`.  For instance Blazor would or still does set the `GetHttpMessageHandler` during their platform initialization and that way devs would not need to create and pass their own handlers unless it was custom.

mcs/class/System.Net.Http/wasm_System.Net.Http.dll.sources
baulig: This method currently calls `new HttpClientHandler ()` - which is only available via the sockets handler.  We could of course return a stubbed-out (`PlatformNotSupportedException`) version of `HttpClientHandler`.

baulig:
Okay, I cleaned this up to make `HttpClientHandler` to (mostly) throw `PlatformNotSupportedException` while delegating it's `SendAsync ()` call to the wasm handler.

This means that you can now write

```
   HttpClientHandler handler = new HttpClientHandler();
   HttpClient client = new HttpClient(handler);
```

but accessing any properties on the `handler` instance will throw.

baulig:
@monojenkins commit csproj

mcs/class/System.Net.Http/HttpClientHandler.wasm.cs
```diff
@@ -0,0 +1,148 @@
...
+					wasmHandler.Dispose ();
+					wasmHandler = null;
+				}
+			}
+			base.Dispose (disposing);
+		}
+
+		const string EXCEPTION_MESSAGE = "System.Net.Http.HttpClientHandler is not supported on the current platform.";
+
+		public virtual bool SupportsAutomaticDecompression => throw new PlatformNotSupportedException (EXCEPTION_MESSAGE);
```
marek-safar: bool properties should all be `false` instead of throw

baulig: Done.

Issue #11350, opened 2018-10-24
Add test coverage for custom modifiers


marek-safar:
We should import recently added CoreCLR tests for custom modifiers. We fail to pass the tests on array custom modifiers at least.

https://github.com/dotnet/coreclr/pull/20546

lambdageek:
https://gist.github.com/lambdageek/bf268c3e037dacfafb0e1734dee0b397 Standalone repro based on https://github.com/dotnet/coreclr/pull/20546/files

The test mostly exercises virtual calls.  We get about a half dozen of them wrong.

Don't know yet if it's because vtable setup is long or if the JIT dispatching the calls to incorrect slots.  My bet is that it's the vtable setup.

lambdageek:
From #12422 we now know that sometimes even if the vtable is correct, method-to-ir will incorrectly translate a `callvirt` (when it looks up a token to get a MonoMethod the signature equivalence uses a type equivalence that is insufficiently strict about cmods).

So there are two steps to working on this bug:
1. turn on vtable debug printing and check whether the correct methods are going in the vtable.
2. If they are, this is a duplicate of 12422; if the vtable is incorrect, then it's a separate issue.

lambdageek:
I see improvement from #13191 - the only bit that's failing now is the arrayModopt2 test at the end (exit  code 302)

lambdageek:
Arrays are failing for a pretty serious reason: We actually can't represent an array whose elements are of a type that has custom modifiers.   For a MONO_TYPE_SZARRAY we just store a `MonoClass*` (which doesn't have any place to store custom modifiers), and for a MONO_TYPE_ARRAY, we store a `MonoTypeArray*` which also just saves a `MonoClass*`, rank and bounds.

(This also applies to pointer and fnptr types)

I don't know the best way to go here: I guess we need to carry custom modifiers in a `MonoClass` (maybe allow variable-sized `byval_arg`)?

Alternately, maybe MonoTypeArray should hold a MonoType not a MonoClass for the element type (but the struct is in the public API, so we'd have to do something messy).

This will not be a quick fix.

marek-safar:
@vargaz could you validate this^ proposal?

vargaz:
I think if mono_type_equal () returns FALSE for 2 types, then their corresponding MonoClass-es should be different too. I'm not sure fixing this is really worth it it seems like a lot of work for little gain.

lambdageek:
[...]

Issue #10565, opened 2018-09-12
[roslyn] Multiple CustomModifiersTests failures due to overridden methods not running


kg:
```
Microsoft.CodeAnalysis.CSharp.UnitTests.Symbols.CustomModifiersTests.ConcatModifiersAndByRefReturn_01 [FAIL]
      Roslyn.Test.Utilities.ExecutionException : 
      Execution failed for assembly '/tmp/RoslynTests'.
      Expected: Overridden
      Overridden P
      Actual:   
      
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
          at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in <af30111a119b4c33adf68bea3aff0284>:0
```

```
Microsoft.CodeAnalysis.CSharp.UnitTests.Symbols.CustomModifiersTests.ConcatModifiersAndByRefReturn_04 [FAIL]
      Roslyn.Test.Utilities.ExecutionException : 
      Execution failed for assembly '/tmp/RoslynTests'.
      Expected: Overridden
      Overridden P
      Actual:   
      
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
          at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in <af30111a119b4c33adf68bea3aff0284>:0 
```

```    
Microsoft.CodeAnalysis.CSharp.UnitTests.Symbols.CustomModifiersTests.ConcatModifiersAndByRef_02 [FAIL]
      Roslyn.Test.Utilities.ExecutionException : 
      Execution failed for assembly ''.
      Expected: Overridden
      Actual:   
      
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
[...]

Issue #10564, opened 2018-09-12
[roslyn] CustomModifiersTests.ConcatModifiers_03 does not work on Mono


kg:
```
    Microsoft.CodeAnalysis.CSharp.UnitTests.Symbols.CustomModifiersTests.ConcatModifiers_03 [FAIL]
      System.Runtime.Serialization.SerializationException : Type 'CL3' in Assembly '8c5f7036-342e-4da0-80d2-af0f521d8894, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' is not marked as serializable.
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
          at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in <af30111a119b4c33adf68bea3aff0284>:0  
```

lambdageek:
Standalone repro gist https://gist.github.com/lambdageek/982e57262a8dffe3e19fda6141a67e11

This appears to be passing with `master` ac5ab1a7efe

Output is:
```
Set Overridden
Get Overridden
```

(and monodis of the repro .exe shows the expected modopts on the property - but that part isn't really mono-related - just means Roslyn is working correctly)

lambdageek:
Hmm.  Actually now I'm not so sure.

1. The expected failure is a serialization exception of some sort - nothing to do with what my repro is doing.
2. The repro could be passing for the wrong reasons - namely - there is nothing ambiguous in the way the vtable is set up here - there's only one method of each name and if you drop the modifiers completely everything will still line up nicely.

I think we need to actually understand what the roslyn test is doing - I'm not sure where serialization comes into it at all.

lambdageek:
This is passing with #13320.

PR #13320, opened 2019-03-05
Custom modifier equivalence, duplication, appending and inflation

lambdageek:
This all stems from ECMA 335, 7.1.1:

> The CLI itself shall treat required and optional modifiers in the same
> manner. Two signatures that differ only by the addition of a custom
> modifier (required or optional) shall not be considered to match. Custom
> modifiers have no other effect on the operation of the VES.

This rule means that when comparing signatures, we have to treat two types that
differ only in their custom modifiers as distinct. In particular, we have to do
this when comparing signatures in order to fill in the vtable for a class.  And
moreover, we have to concatenate custom modifiers when doing generic
instantiation.  If a method in a generic type has signature `T modopt(IsConst)
Method()` and we instantiate T with `int32 modopt(IsLong)`, we are expected to
produce the signature `int32 modopt(IsConst) modopt(IsLong) Method()` (with the
custom modifiers on `int32` in that order).

So this PR implements the following operations:
1. custom modifier appending during `mono_metadata_type_dup_with_cmods`
2. custom modifier-respecting comparison in `do_mono_metadata_type_equal`

To do this, we actually have to introduce a second representation of custom
modifiers: in the normal representation, all the cmod type specs come from a
single image.  Whereas now, due to inflation of signatures, we can have a mix
of modifiers from different images.  So in the MonoAggregateModContainer
representation, we pair each typespec token with the image in which it is to be
resolved.

We keep the old representation (a single image to resolve the entire set of
typespec tokens) because it's natural when you're just reading from a single
image.  Also this representation is part of the embedding API.

---

### But that's not all ###

It is possible for generic parameters of a generic class to appear in the
_custom modifiers_ of a type in a method signature.  For example:
```
    T modopt(Nullable`1<T>) Method ()
```

In this case, when we inflate this signature, we need to inflate not only the
return type, but also the modifiers on a return type.

That means our old aggregate representation (where each modifier is a pair of a
MonoImage and a typespec token) is inadequate: we can inflate a signature and
get a custom modifier that doesn't appear as a typespec in any table of any
existing image.

This is the same situation as generic instances: as a program runs, it creates
generic instances that are not part of the source code.  So we solve it the
same way as with generic instances.

Our new `MonoAggregateModContainer` representation is a pointer into memory owned
by a MonoImageSet that contains an entire MonoType for each custom modifier.

The invariant is that the aggregate modifier container is allocated from a
mempool of a `MonoImageSet` that includes the images of all the modifiers.

In general, when inflating, we first create a provisional candidate
MonoAggregateModContainer on the stack and then look it up the image set that
should own it and then look for a "canonical" mod container in the
`MonoImageSet:aggergate_modifiers_cache`

---

Also had to bump the AOT image format, which now always reads back in using the aggregate representation.

---

This should address #10564, #10565, #12422, #10945 (again), #6936, and the parts of #11350 that **don't** deal with arrays.

---

TODO:

- [x] Add some representative test cases to this PR.
- [ ] Modify our acceptance tests to run the various disabled Roslyn tests.

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -849,24 +849,38 @@ inflate_generic_type (MonoImage *image, MonoType *type, MonoGenericContext *cont
 static MonoType*
 inflate_generic_custom_modifiers (MonoImage *image, const MonoType *type, MonoGenericContext *context, MonoError *error)
 {
+	MonoType *result = NULL;
 	g_assert (type->has_cmods);
 	int count = mono_type_custom_modifier_count (type);
 	gboolean changed = FALSE;
-	MonoType *new_type = g_alloca (mono_sizeof_type_with_mods (count, TRUE));
-	memcpy (new_type, type, MONO_SIZEOF_TYPE);
-	mono_type_with_mods_init (new_type, count, TRUE);
-	MonoAggregateModContainer *new_mods = mono_type_get_amods (new_type);
+
+	size_t aggregate_size = mono_sizeof_aggregate_modifiers (count);
+	MonoAggregateModContainer *candidate_mods = g_alloca (aggregate_size);
+	memset (candidate_mods, 0, aggregate_size);
+	candidate_mods->count = count;
 
 	for (int i = 0; i < count; ++i) {
 		gboolean required;
 		MonoType *cmod_old = mono_type_get_custom_modifier (type, i, &required, error);
-		return_val_if_nok (error, NULL);
-		MonoType *cmod_new = inflate_generic_type (NULL, cmod_old, context, error); /* FIXME leaks */
-		return_val_if_nok (error, NULL);
+		goto_if_nok (error, leave);
+		MonoType *cmod_new = inflate_generic_type (NULL, cmod_old, context, error);
+		goto_if_nok (error, leave);
 		if (cmod_new)
 			changed = TRUE;
-		new_mods->modifiers [i].required = required;
-		new_mods->modifiers [i].type = cmod_new ? cmod_new : cmod_old;
+		candidate_mods->modifiers [i].required = required;
+		candidate_mods->modifiers [i].type = cmod_new;
+	}
+
+	if (changed) {
+		/* if we're going to make a new type, fill in any modifiers that weren't affected by inflation with copies of the original values. */
+		for (int i = 0; i < count; ++i) {
+			if (candidate_mods->modifiers [i].type == NULL) {
+				candidate_mods->modifiers [i].type = mono_metadata_type_dup (NULL, mono_type_get_custom_modifier (type, i, NULL, error));
+					
+				/* it didn't error in the first loop, so should be ok now, too */
+				mono_error_assert_ok (error);
+			}
+		}
 	}
 #ifdef DEBUG_INFLATE_CMODS
 	if (changed) {
@@ -875,9 +889,23 @@ inflate_generic_custom_modifiers (MonoImage *image, const MonoType *type, MonoGe
 		g_free (full_name);
 	}
 #endif
-	if (!changed)
-		return NULL;
-	return mono_metadata_type_dup (image, new_type);
+	if (changed) {
+		MonoType *new_type = g_alloca (mono_sizeof_type_with_mods (count, TRUE));
+		/* first init just the non-modifier portion of new_type before populating the
+		 * new modifiers */
+		memcpy (new_type, type, MONO_SIZEOF_TYPE);
+		mono_type_with_mods_init (new_type, count, TRUE);
+		mono_type_set_amods (new_type, mono_metadata_get_canonical_aggregate_modifiers (candidate_mods));
+		result =  mono_metadata_type_dup (image, new_type);
+	}
+
+leave:
+	for (int i = 0; i < count; ++i) {
+		if (candidate_mods->modifiers [i].type)
+			mono_metadata_free_type (candidate_mods->modifiers [i].type);
+	}
+
+	return result;
 }
 
 MonoGenericContext *
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -854,6 +854,8 @@ inflate_generic_custom_modifiers (MonoImage *image, const MonoType *type, MonoGe
 	int count = mono_type_custom_modifier_count (type);
 	gboolean changed = FALSE;
 
+	/* Try not to blow up the stack. See comment on MONO_MAX_EXPECTED_CMODS. */
+	g_assert (count < MONO_MAX_EXPECTED_CMODS);
 	size_t aggregate_size = mono_sizeof_aggregate_modifiers (count);
 	MonoAggregateModContainer *candidate_mods = g_alloca (aggregate_size);
 	memset (candidate_mods, 0, aggregate_size);
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -72,6 +72,9 @@ static gboolean mono_class_is_subclass_of_internal (MonoClass *klass, MonoClass 
 GENERATE_GET_CLASS_WITH_CACHE (valuetype, "System", "ValueType")
 GENERATE_TRY_GET_CLASS_WITH_CACHE (handleref, "System.Runtime.InteropServices", "HandleRef")
 
+// define to print types whenever custom modifiers are appended during inflation
+#undef DEBUG_INFLATE_CMODS
+
 static
 MonoImage *
 mono_method_get_image (MonoMethod *method)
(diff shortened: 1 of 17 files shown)


lambdageek:
We also pass [`ModifiersWithConstructedType_01`](https://github.com/dotnet/roslyn/blob/defeca5d0e44915fc978cfd93d3de2e4a6d0dbf4/src/Compilers/CSharp/Test/Symbol/Symbols/CustomModifiersTests.cs#L2044) from the Roslyn test suite.  So we should make a PR against roslyn to enable that test.

lambdageek:
I'm doing a lot of cloning -  I would especially welcome sanity checking of all the memory management.

lambdageek:
@monojenkins build failed

mono/metadata/class.c
```diff
@@ -632,18 +635,40 @@ can_inflate_gparam_with (MonoGenericParam *gparam, MonoType *type)
...
 		int num = mono_type_get_generic_param_num (type);
 		MonoGenericInst *inst = context->method_inst;
-		if (!inst)
-			return NULL;
+		if (!inst) {
+			if (!changed)
+				return NULL;
+			else
+				return type;
+		}
```
jaykrell: `return changed ? type : NULL;`?

mono/metadata/class.c
```diff
@@ -688,16 +717,40 @@ inflate_generic_type (MonoImage *image, MonoType *type, MonoGenericContext *cont
...
 		}
 
+#ifdef DEBUG_INFLATE_CMODS
+		gboolean append_cmods;
+		append_cmods = FALSE;
+		if (type->has_cmods && inst->type_argv[num]->has_cmods) {
+			char *tname;
+			char *vname;
+			tname = mono_type_full_name (type);
+			vname = mono_type_full_name (inst->type_argv[num]);
```
jaykrell: `char* vname = vname = mono_type_full_name (inst->type_argv[num]);` etc.?

mono/metadata/class.c
```diff
@@ -760,18 +831,82 @@ inflate_generic_type (MonoImage *image, MonoType *type, MonoGenericContext *cont
...
 
+static MonoType*
+inflate_generic_custom_modifiers (MonoImage *image, const MonoType *type, MonoGenericContext *context, MonoError *error)
+{
+	MonoType *result = NULL;
+	g_assert (type->has_cmods);
+	int count = mono_type_custom_modifier_count (type);
+	gboolean changed = FALSE;
+	MonoType *new_type = g_alloca (mono_sizeof_type_with_mods (count, TRUE));
+	memcpy (new_type, type, MONO_SIZEOF_TYPE);
```
jaykrell: `*new_type = *type`?

jaykrell:
Wow.

mono/metadata/class.c
kg: Previously it needed to be an alloc + some copies because mono_types with cmods are bigger than sizeof(type)

lambdageek: yea this line probably merits a comment or a helper function - I am specifically just copying the "plain type" slice of `type` into `new_type`.  It's true that `*new_type = *type` would do the same thing, but it's perhaps even less evident than a memcpy.

jaykrell: It should be obvious, at least in C.

lambdageek:
Also reversed the order that we store custom modifiers.
If you write IL like this:
```
   int32 modopt (Foo) modopt (Bar)
```
both .NET Framework ilasm and Mono's ilasm will produce an assembly with this:
```
   cmod_opt [encoding of Bar] cmod_opt [encoding of Foo] I4
```
So when parsing, we should store it with Foo first, then Bar.

And `monodis` should display it in the same order that `ilasm` consumed it in.

lambdageek:
@monojenkins build failed

mono/metadata/class.c
```diff
@@ -688,16 +717,38 @@ inflate_generic_type (MonoImage *image, MonoType *type, MonoGenericContext *cont
 			return NULL;			
 		}
 
+#ifdef DEBUG_INFLATE_CMODS
```
alexanderkyte: Can we hook this up to some autogen setting or document where it's set? Does someone just add a #define somewhere? Could you add a commented-out copy of that define, so people know where to put it?

mono/metadata/class.c
```diff
@@ -760,18 +829,85 @@ inflate_generic_type (MonoImage *image, MonoType *type, MonoGenericContext *cont
...
+	}
+#ifdef DEBUG_INFLATE_CMODS
+	if (changed) {
+		char *full_name = mono_type_full_name ((MonoType*)type);
+		printf ("\n\n\tcustom modifier on '%s' affected by subsititution\n\n\n", full_name);
+		g_free (full_name);
+	}
+#endif
+	if (changed) {
+		MonoType *new_type = g_alloca (mono_sizeof_type_with_mods (count, TRUE));
```
alexanderkyte: Can you add some assertions around this type of allocation (around the maximum return value of "mono_sizeof_type_with_mods") so that if this explodes the alt stack or something, it won't be a big debugging session?

mono/metadata/icall.c
```diff
@@ -7928,9 +7930,12 @@ type_array_from_modifiers (MonoImage *image, MonoType *type, int optional, MonoE
 	res = mono_array_new_handle (domain, mono_defaults.systemtype_class, count, error);
 	goto_if_nok (error, fail);
 	count = 0;
-	for (i = cmods->count - 1; i >= 0; --i) {
-		if ((optional && !cmods->modifiers [i].required) || (!optional && cmods->modifiers [i].required)) {
-			if (!add_modifier_to_array (domain, image, &cmods->modifiers [i], res, count , error))
+	for (i = cmod_count - 1; i >= 0; --i) {
+		gboolean required;
```
alexanderkyte: How does this code get hit/tested? Might be good to document. It seems difficult to go from the type of breakage you'd see if this misbehaved back to this code.

mono/metadata/metadata-internals.h
```diff
@@ -36,14 +36,86 @@ struct _MonoType {
...
+ * in image1 and the int32 is in image2.)
+ *
+ * Moreover, we can't just store an image and a type token per modifier, because
+ * Roslyn and C++/CLI sometimes create modifiers that mention generic parameters that must be inflated, like:
+ *     void .CL1`1.Test(!0 modopt(System.Nullable`1<!0>))
+ * So we have to store a resolved MonoType*.
+ *
+ * Because the types come from different images, we allocate the aggregate
+ * custom modifiers container object in the mempool of a MonoImageSet to ensure
+ * that it doesn't have dangling image pointers.
```
alexanderkyte: Well-written. 👍

mono/metadata/metadata.c
```diff
@@ -2426,7 +2430,40 @@ dump_ginst (MonoGenericInst *ginst)
 	g_print (">");
 }*/
 
+static gboolean
+aggregate_modifiers_equal (gconstpointer ka, gconstpointer kb)
```
alexanderkyte: At some point all of this modifier stuff probably deserves it's own file, outside of metadata. The churn might be too much for this PR though.

mono/metadata/reflection.c
```diff
@@ -459,6 +459,12 @@ mono_type_get_object_checked (MonoDomain *domain, MonoType *type, MonoError *err
 	 */
 	type = m_class_get_byval_arg (klass)->byref == type->byref ? m_class_get_byval_arg (klass) : m_class_get_this_arg (klass);
 
+	// We don't want to return types with custom modifiers to the managed
+	// world since they're hard to distinguish from plain types using the
+	// reflection APIs, but they are not ReferenceEqual to the unadorned
+	// types.
+	g_assert (!type->has_cmods);
```
alexanderkyte: Is this a "temporarily we don't support this" or a "this will never work" scenario?

alexanderkyte:
Left a few comments. Pretty readable all-in-all. I would feel better about it as a reviewer if I knew we had more coverage of it. Any chance we know of a larger suite testing this in one of the other dotnet foundation repos?

vargaz:
Wouldn't be better to only use the new representation, and do some kind of conversion for the embedding api ?

mono/metadata/class.c
lambdageek: There's an `#undef DEBUG_INFLATE_CMODS` at the top of the file:
https://github.com/mono/mono/pull/13320/files#diff-f085caaf39e861140ff40b9ef0c516d9R76

mono/metadata/reflection.c
lambdageek: It's a "this is a bug in the runtime if it ever happens" scenario.

lambdageek:
> Wouldn't be better to only use the new representation, and do some kind of conversion for the
> embedding api ?

Yea, that's possible.  I thought there may be some benefit to keeping the old representation - it's a bit more compact.

lambdageek:
Addressed @alexanderkyte comments.

Added a new reflection test.

Fixed a bug where our SRE was emitting cmods in reverse order.  Undid the order-reversing from #11308.  The same test from that PR still works - in effect we shifted the cmod reversal from reflection query (the wrong place for it) to SRE (the right place for it).

lambdageek:
@vargaz on second thought, getting rid fo the old custom mod representation is tricky: it will complicate the medatada parsing code and possibly make us resolve types too early.

The old representation is just an image and an array of tokens and it's allocated with the MonoType.  So there's no extra memory to manage, and we don't have to resolve any of the types - it matches up pretty nicely with what's in the assemblies.

The new representation is a pool-allocated array of `MonoType*`.  So we'd have to turn each image+token into a type right in the metadata parser.  So things can go wrong earlier.  Which might be more difficult to ignore or to recover from.

I would prefer to leave things with the two representations, if we can.

PR #13322, opened 2019-03-06
[linker] Only preserve the m_value fields of the primitive types.

vargaz:


vargaz:
@monojenkins build failed

Issue #12182, opened 2018-12-23
LLVM AOT caches require SSE4.1-capable CPU


Nichtraucher:
## Steps to Reproduce

1. update to 5.18.0.225-0xamarin1+ubuntu1404b1
2. 
3. 

### Current Behavior

update fails due to unresolved dependencies and other errors (see terminal output below)

### Expected Behavior

errr...update goes through smoothly (as always)?

## On which platforms did you notice this

Linux (Linux Mint 17.2 Rafaela (4.4.0-141-generic #167~14.04.1-Ubuntu SMP Mon Dec 10 13:20:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux). I'm using the repositories provided by mono.

**Version Used**:
Mono JIT compiler version 5.18.0.225
TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       preemptive
	GC:            sgen (concurrent by default)

## Stacktrace
Terminal output here: https://pastebin.com/um5HptcP

I read in the chat that at least one other user seems to have the same/a similar issue, she posted her output here: https://0x0.st/sdBY.txt , hence my decision to post this report

any help is much appreciated!

guydavis:
Confirmed same problem with 5.18.0.225-0xamarin1+ubuntu1604b1

simeon4110:
I get stuck in a SIGKILL:


```
=================================================================
Got a SIGILL while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================

Aborted (core dumped)
E: installing Assembly /usr/lib/cli/nunit.core.interfaces-2.6.3/nunit.core.interfaces.dll failed
E: Installation of libnunit-core-interfaces2.6.3-cil with /usr/share/cli-common/runtimes.d/mono failed
* Installing 1 assembly from libnunit-framework2.6.3-cil into Mono
Stacktrace:

/proc/self/maps:
40900000-40920000 rwxp 00000000 00:00 0
41322000-41332000 rwxp 00000000 00:00 0
55da82ca9000-55da830f2000 r-xp 00000000 08:22 13895047                   /usr/bin/mono-sgen
55da832f1000-55da832f8000 r--p 00448000 08:22 13895047                   /usr/bin/mono-sgen
[...]

PR #13323, opened 2019-03-06
[AOT] add option to control arguments to llc and add fallback for x86

lewurm:
Fixes https://github.com/mono/mono/issues/12182

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -228,6 +228,7 @@ typedef struct MonoAotOptions {
 	char *instances_logfile_path;
 	char *logfile;
 	char *llvm_opts;
+	char *llvm_llc_opts;
 	char *llvm_llc;
 	gboolean dump_json;
 	gboolean profile_only;
@@ -1081,6 +1082,8 @@ arch_emit_unwind_info_sections (MonoAotCompile *acfg, const char *function_start
 static void
 arch_init (MonoAotCompile *acfg)
 {
+	gboolean has_custom_args = !!acfg->aot_opts.llvm_llc_opts;
+	char *custom_args = acfg->aot_opts.llvm_llc_opts;
 	acfg->llc_args = g_string_new ("");
 	acfg->as_args = g_string_new ("");
 	acfg->llvm_owriter_supported = TRUE;
@@ -1095,11 +1098,11 @@ arch_init (MonoAotCompile *acfg)
 	acfg->user_symbol_prefix = "";
 
 #if defined(TARGET_X86)
-	g_string_append (acfg->llc_args, " -march=x86 -mattr=sse4.1");
+	g_string_append_printf (acfg->llc_args, " -march=x86 %s", has_custom_args ? custom_args : "-mcpu=generic");
 #endif
 
 #if defined(TARGET_AMD64)
-	g_string_append (acfg->llc_args, " -march=x86-64 -mattr=sse4.1");
+	g_string_append_printf (acfg->llc_args, " -march=x86-64 %s", has_custom_args ? custom_args : "-mcpu=generic");
 	/* NOP */
 	acfg->align_pad_value = 0x90;
 #endif
@@ -1128,6 +1131,9 @@ arch_init (MonoAotCompile *acfg)
 
 	if (acfg->aot_opts.mtriple)
 		mono_arch_set_target (acfg->aot_opts.mtriple);
+
+	if (has_custom_args)
+		g_string_append_printf (acfg->llc_args, " %s", custom_args);
 #endif
 
 #ifdef TARGET_ARM64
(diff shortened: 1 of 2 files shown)


filipnavara:
Wouldn't this make the AOT binaries unusable if I transfer them to different machine with less capable CPU?

vargaz:
Yes, that is the reason we don't use it. Perhaps we can add an aot command line option, or just the existing mtriple option.

lewurm:
it isn't worse than before in that regard

filipnavara:
My CPU supports AVX2. With the added flag it would pick it up and stop working even on SSE4.1 CPUs. That looks worse to me.

lewurm:
That specific example, yes. In general it was broken in that way before, but I get your point 🙂 


I updated the PR, what do you think now?

@directhex it would mean you would need to update debian packaging so that it will pass `--aot=llvmllcopts=-mcpu=native` to the AOT compiler.

filipnavara:
Realistically SSE4.1 was available on Intel CPUs for 11 years (Core 2 / Penryn) and on AMD CPUs for 8 years (AMD FX / Opteron / Bulldozer) so the base target was already quite low. I personally don't mind changing it as long as it does not break binaries I distribute to customers.

On some targets like macOS the base level could certainly be higher because the supported macOS version has already pretty high bar on CPU requirements. Currently Mono runs on macOS Mavericks or newer and Apple officially supports macOS Sierra and newer. The only machine without SSE4.1 where macOS Mavericks might have been supported is some low configuration of iMac from 2007 (Merom architecture) and I am not even sure if that one was officially supported or not.

mono/mini/aot-compiler.c
```diff
@@ -7873,6 +7884,8 @@ mono_aot_parse_options (const char *aot_options, MonoAotOptions *opts)
 			opts->llvm_opts = g_strdup (arg + strlen ("llvmopts="));
 		} else if (str_begins_with (arg, "llvmllc=")){
 			opts->llvm_llc = g_strdup (arg + strlen ("llvmllc="));
+		} else if (str_begins_with (arg, "llvmllcopts=")) {
+			opts->llvm_llc_opts = g_strdup (arg + strlen ("llvmllcopts="));
```
vargaz: I would call this llcmopts. Alternative, we can add a mcpu/march argument similar to the mtuple argument.

directhex:
@lewurm that's easily changed

Bear in mind 1) we had a bunch of angry bug reports recently because SSE4.1 as a hard dependency for LLVM AOT broke lots of people 2) the Debian/Ubuntu packages generate AOT caches at install time (not shipped in packages) so we can be as aggressive in optimizing as we like, as long as it decides what to do at install time (i.e. yay for `-mnative`)

lewurm:
@directhex how far do you want this to be backported? 2018-10?

lewurm:
@monojenkins backport 2019-02

marek-safar:
@monojenkins backport 2018-10

lewurm:
@marek-safar: @directhex communicated (via Slack) that we should only backport it to `2019-02`. It requires changes in Linux packaging.

directhex:
So, what is the desired outcome of this work?

Right now, when you install a package with a `installmonoaot` manifest, this script gets executed:

https://github.com/mono/linux-packaging-mono/blob/b7e7a350b3d0dccf7287c72598be9aa8018d8031/debian/postinst-monoaot#L3-L25

So _IF_ `/proc/cpuinfo` contains `sse4_1`, and `/usr/lib/libmono-llvm.so.0` exists, it calls:
```
/usr/bin/mono --aot=try-llvm,temp-path=/tmp,outfile=/usr/lib/mono/aot-cache/${MONOARCH}/foo.dll.so` -O=all foo.dll
```

else it calls
```
/usr/bin/mono --aot=temp-path=/tmp,outfile=/usr/lib/mono/aot-cache/${MONOARCH}/foo.dll.so` -O=all foo.dll
```

What is the new desired behaviour? I guess the SSE4.1 check can be removed entirely? Are we comfortable that this will do the right thing on ARM etc?

lewurm:
The script should pass `--aot=llvmllcopts=-mcpu=native` on i386/amd64.

For arm/aarch64: I'm not sure. It looks like LLVM would support `-mcpu=native` for those targets as well, but I have no idea how well it works in practice.
https://github.com/llvm/llvm-project/blob/4218d45f6581fd508f493522bf514981d5c9ca2b/clang/lib/Driver/ToolChains/Arch/ARM.cpp#L570-L583
https://github.com/llvm/llvm-project/blob/4218d45f6581fd508f493522bf514981d5c9ca2b/clang/lib/Driver/ToolChains/Arch/AArch64.cpp#L39-L43

EgorBo:
We now have two ways of passing args to `llc`:
`--aot=llvm,llcopts=""`
and
`--aot=llvm,llvmllc=""`
maybe we should drop one? or why `llvmllc` was not changed to just override `-mcpu`?

lewurm:
@EgorBo I falsely assumed that `--aot=llvmllc=[...]` would specify the path of `llc` 🤦‍♂️  could you please review https://github.com/mono/mono/pull/13877 ?

PR #13324, opened 2019-03-06
Propagate type load errors out of IsAssignableFrom instead of erroneously returning false

kg:
This change makes the underlying cause of #13113 discoverable - a type load error was occurring and instead of surfacing the error we were returning false from IsAssignableFrom, causing mysterious failures in other code.


vargaz:
@monojenkins build failed

PR #13198, opened 2019-02-26
NS2.1 Sync new changes, mobile profiles

EgorBo:
Sync the latest ns changes with Mono (Index, Range, IAsyncDisposable, etc). Also some progress on mobile profiles (almost finished `monodroid`).

PR #9746, opened 2018-07-26
[System.Reflection] CoreFX import for MemberInfo

MaximLipnin:
Part of #9660.

The imported CoreFX types:
- MemberInfo;
~~AssemblyNameProxy;~~
~~Binder~~
~~Missing.~~

The changes:
- used MemberInfo type from CoreFX;
- moved MetadataToken property from MemberInfo to leaf classes; added unit tests.
- updated icall-def.h for related types;
- updated corlib version;
- added overridings for new HasSameMetadataDefinitionAs property to leaf classes; 
- included more MemberInfo xunit tests to verify HasSameMetadataDefinitionAs property;
- renamed MemberInfoExtensions class to DbLinqMemberInfoExtensions in System.Data.Linq assembly to avoid a conflict with the class imported to System.Reflection namespace.
- [acceptance-tests] removed AssemblyExtensions class which duplicates the same functionality introduced in System.Reflection.

PR #13325, opened 2019-03-06
 [Facades] Add typeforwarders for DispatchProxy and TypeExtensions

akoeplinger:
DispatchProxy lives in System.Core for all profiles since #13198.

TypeExtensions live in corlib since #9746.


akoeplinger:

PR #13326, opened 2019-03-06
Add System.Drawing converters to netstandard facade

EgorBo:
This PR adds missing System.Drawing converters thus it completes NS2.1 support (as of today) for all profiles except macOS\IOS specific ones: monotouch, tv, watch, xammac, xammac_net_4_5.


mcs/class/Facades/netstandard/TypeForwarders.cs
```diff
@@ -2446,7 +2446,7 @@
 [assembly:System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Diagnostics.Tracing.EventSourceCreatedEventArgs))]
 [assembly:System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Drawing.KnownColor))]
 [assembly:System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Data.Common.DbProviderFactories))]
-#if SYSTEM_DRAWING
+#if !NO_SYSTEM_DRAWING
```
marek-safar: I think this should point to System.Drawing.Common instead in all cases. @akoeplinger thoughts?

marek-safar:
@monojenkins commit csproj

PR #13333, opened 2019-03-06
threads: set mutex PRIO_INHERIT in attribute

kurt-vd:
When running multiple threads on a linux-rt kernel, with different priorities, the lack of priority-inheritance
makes the kernel make wrong scheduling choices.
This commit changes the system-wide mutex initializers.
This change protects our system against unexpected latencies in higher priority threads.


marek-safar:
approve

luhenry:
@monojenkins squash

luhenry:
@monojenkins build failed

PR #13334, opened 2019-03-06
[wasm] add mixed mode assemblies exclusion

jeromelaban:
Added a packager option to allow for mixed mode assemblies exclusion, using `--skip-aot-assemblies` where by default all provided assemblies are AOTed, except the ones provided by `--skip-aot-assemblies`.

This enables scenarios where lots of assemblies can be AOTed, and some need to be excluded, such as JSon.NET or System.Xml.

### DIFF
--- a/sdks/wasm/packager.cs
+++ b/sdks/wasm/packager.cs
@@ -126,6 +126,7 @@ class Driver {
 		Console.WriteLine ("\t\t              'ifnewer' copies or overwrites the file if modified or size is different.");
 		Console.WriteLine ("\t--profile=x     Enable the 'x' mono profiler.");
 		Console.WriteLine ("\t--aot-assemblies=x List of assemblies to AOT in AOT+INTERP mode.");
+		Console.WriteLine ("\t--skip-aot-assemblies=x List of assemblies to skip AOT in AOT+INTERP mode.");
 		Console.WriteLine ("\t--link-mode=sdkonly|all        Set the link type used for AOT. (EXPERIMENTAL)");
 		Console.WriteLine ("\t\t              'sdkonly' only link the Core libraries.");
 		Console.WriteLine ("\t\t              'all' link Core and User assemblies. (default)");
@@ -336,6 +337,7 @@ class Driver {
 		string sdkdir = null;
 		string emscripten_sdkdir = null;
 		var aot_assemblies = "";
+		var skip_aot_assemblies = "";
 		out_prefix = Environment.CurrentDirectory;
 		app_prefix = Environment.CurrentDirectory;
 		var deploy_prefix = "managed";
@@ -387,6 +389,7 @@ class Driver {
 				{ "profile=", s => profilers.Add (s) },
 				{ "copy=", s => copyTypeParm = s },
 				{ "aot-assemblies=", s => aot_assemblies = s },
+				{ "skip-aot-assemblies=", s => skip_aot_assemblies = s },
 				{ "link-mode=", s => linkModeParm = s },
 				{ "help", s => print_usage = true },
 					};
@@ -451,6 +454,12 @@ class Driver {
 				return 1;
 			}
 		}
+		if (skip_aot_assemblies != "") {
+			if (ee_mode != ExecMode.AotInterp) {
+				Console.Error.WriteLine ("The --skip-aot-assemblies= argument requires --aot-interp.");
+				return 1;
+			}
+		}
 		if (link_icalls && !enable_linker) {
 			Console.Error.WriteLine ("The --link-icalls option requires the --linker option.");
 			return 1;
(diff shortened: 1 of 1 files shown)


sdks/wasm/packager.cs
```diff
@@ -489,7 +498,7 @@ class WasmOptions {
 		}
 
 		if (enable_aot) {
-			var to_aot = new Dictionary<string, bool> ();
+			var to_aot = new Dictionary<string, bool> (StringComparer.OrdinalIgnoreCase);
```
kg: is it possible to load two assemblies named Foo.dll and FOO.dll? Is it reasonable to expect someone would do that? I'm okay with the answer being "don't do that" if they try to feed them to the linker, just wondering though

jeromelaban: That's a good question. The main reason I'm adding this is that assembly names are provided by user entry in my scenario, but I'm not sure how that's relevant. 

I should probably do that validation/sanitization upfront.

lambdageek: .NET Framework (and Mono, and Core (?)) treat assembly names as case insensitive (so if an assembly names FOO is already loaded, references to Foo will just give you back FOO again).  But Mono probes the filesystem in a case-preserving manner (so if Foo and FOO aren't loaded and you ask for FOO.dll, we'll look for that exact name and won't see Foo.dll).

You could, sometimes, maybe, load two assemblies with the same name if you use `Assembly.LoadFile` or `Assembly.Load (byte[])` - the guarantees about what will happen are pretty vague, however.

jeromelaban: I remember now. The reason I added this was for convenience, as the user needs to input assemblies to AOT manually (even if those are discovered through assemblies inspection) through the CLI. That's the same reason I added this (to skip AOT): 

https://github.com/mono/mono/pull/13334/files/828793614595bab0038181daeaf6494780230095#diff-6ada20ee030b151fa1dba678cfaf43e4R522

as it can be difficult to the user to know the exact casing of an assembly.

luhenry:
Instead of adding an option, it would be more user friendly to use `--aot-assemblies=-System.Core.dll` to interpret System.Core for example. This is the convention used in Xamarin.iOS.

jeromelaban:
@luhenry I chose to make two different lists because `--aot-assemblies` and `--skip-aot-assemblies` are mutually exclusive. Does adding one exclusion in Xamarin.iOS make all the other assemblies registered for AOT ?

lewurm:
@jeromelaban This is how `mtouch` (Xamarin.iOS) exposes it:
https://github.com/xamarin/xamarin-macios/blob/849355eaf2cbdbc4909c89570fd6713389f3355d/tools/mtouch/mtouch.cs#L1190-L1197

A few examples:
* `--interpreter=-mscorlib`: `mscorlib.dll` will be AOT'd, everything else runs in the interpreter
* `--interpreter=-System,-System.Core`: `System.dll` and `System.Core.dll` will be AOT'd, everything else runs in the interpreter
* `--interpreter=all`: everything runs in the interpreter
* `--interpreter=-all`: everything will be AOT'd. However, dynamically generated code (such as `System.Reflection.Emit`) will run in the interpreter.

It would be nice to have it consistent across products 🙂

steveisok:
@lewing does this just need to be re-reviewed / merge conflicts fixed?

lewing:
So the issue here seems to be that packager has a combination of mode switches and --aot-assemblies.  Perhaps the solution is to combine them into one list?

jeromelaban:
I'll try adjusting the list to use that format. 

It will be a mutually exclusive parameter with `--aot--assemblies`, where specifying assemblies in that list will disable the `--interpreter` parameter, to not be breaking.

jeromelaban:
Here's how it may work:

- `--aot`, `--aot-interp`, `--aot-assemblies` are removed
- `--interpreter` is added

The following cases are then possible:
- `--interpreter=all` enables the interpreter only mode (no flags currently)
- `--interpreter=-all` enables the Full AOT mode with the interpreter disabled (`--aot` flag currently)
- `--interpreter=-mscorlib,System` is the equivalent of `--aot-interp --aot-assemblies=mscorlib,System`
- `--interpreter=-all,System.Data` enables AOT for everything, and the interpreter for `System.Data`. This has no equivalent and that this PR tries to add.

Would that better match what Xamarin.iOS is doing ?

kg:
@jeromelaban Hi Jerome,

Did you need input from us (like @lewing or someone else) to figure out how to update this PR or did you just set it aside? It looks to me like you needed some info on xamarin.ios so we can try to track that down if you still want to get these changes into main-line.

lewing:
I'm going to close this for now.  We'll be making some other tooling changes soon.

PR #13338, opened 2019-03-07
HttpClient for Android.

baulig:


baulig:
@monojenkins commit csproj

baulig:
I pressed "Ready for review" too early - please hold off merging for a bit.

baulig:
This will require work on the Android side in their https://github.com/xamarin/xamarin-android/blob/master/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs, but it is good from the Mono side of things.

marek-safar:
@monojenkins commit apidiff

PR #13339, opened 2019-03-07
HttpClient for ios / mac.

baulig:


mcs/class/System.Net.Http/NSUrlSessionHandler.cs
```diff
@@ -592,7 +592,8 @@ class NSUrlSessionDataTaskStreamContent : StreamContent
 		{
 			Action disposed;
 
-			public NSUrlSessionDataTaskStreamContent (NSUrlSessionDataTaskStream source, Action onDisposed, CancellationToken token) : base (source, token)
+			[Obsolete ("FIXME: Please talk to Martin about this; see https://github.com/mono/mono/issues/12996.")]
```
marek-safar: This is private API used in this file only in our source code, if you need to fix it then please do it

baulig: That's why this is still a _Draft_ pull request ;-)

lewing:
@baulig does this still represent work in progress? or should it be closed?

baulig:
I still want to eventually finish this some day, but would then probably start on top of current master.  Closing for the moment.

PR #13341, opened 2019-03-07
Reduce number of classes in winaot mscorlib.

lateralusX:
Winaot profile should be a Windows AOT friendly profile close to mobile and consoles. At some point it was set to include the full net_4_x that cause problems on none desktop platforms. This fix reduce the set to only include classes similar to mobile and console profiles.


akoeplinger:
@monojenkins commit csproj

PR #13344, opened 2019-03-07
Correctly define thread API support for different Windows platforms.

lateralusX:
Switch mono-threads-windows.c to use win32subset.h and more fine grained control over API support. Needed to better control supported API surface on none desktop Windows platforms.


mono/utils/mono-threads-windows.c
```diff
@@ -286,10 +287,10 @@ mono_threads_suspend_begin_async_resume (MonoThreadInfo *info)
 		if (!res) {
 			return FALSE;
 		}
-	}
 #else
-	g_error ("Not implemented due to lack of SetThreadContext");	
+		g_error ("Not implemented due to lack of GetThreadContext/SetThreadContext");
```
jaykrell: This should only say SetThreadContext. And let's ask for it.

jaykrell: I mean, I see GetThreadContext is in UWP current docs and 10240 SDK.
Agreed SetThreadContext is missing.

jaykrell:
A few minor concerns here.

1. Prefer LoadLibraryW over GetModuleHandle.
But understood it might be disallowed by policy.
So maybe another `#if`?
Or at least a comment?

2. Either way, check LoadLibraryW / GetModuleHandleW
for NULL. On unusual systems w/o kernel32.dll,
you might get back null. Which is then valid
input to GetProcAddress, and means the .exe,
which is typically ok, but you don't really
mean to be running GetProcAddress against the exe.

Just really the usual rule: Check for errors.

3. wow64 check: The combination of `#if` and GetProcAddress seems odd.
If the `#if` is true, just call directly w/o GetProcAddress?
Or, remove the `#if` and always GetProcAddress?

Or maybe you want the `#if` for when you know GetProcAddress will fail?

4. Cache the result of GetProcAddress, success or failure?

5. GetThreadContext is in UWP but SetThreadContext. We should ask.

lateralusX:
Just fixed the defines in this PR, no ambition (at least in this PR) to rewrite the rest of the implementation (that is old). The comments above are mostly related to the existing code, not touched by this PR.

mono/utils/mono-threads-windows.c
lateralusX: Yes, since SetThreadContext is missing the complete block is ruled out, viewing them as a pair of needed API's.

jaykrell: The old version was correct. SetThreadContext is the problem. GetThreadContext is not.

mono/metadata/w32subset.h
```diff
@@ -178,3 +178,43 @@
 #define HAVE_API_SUPPORT_WIN32_GET_MODULE_HANDLE 0
 #endif
 #endif
+
+#ifndef HAVE_API_SUPPORT_WIN32_GET_THREAD_CONTEXT
```
jaykrell: I don't think this part is needed. GetThreadContext is always supported.

mono/metadata/w32subset.h
```diff
@@ -178,3 +178,43 @@
...
+
+#ifndef HAVE_API_SUPPORT_WIN32_CANCEL_SYNCHRONOUS_IO
+#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#define HAVE_API_SUPPORT_WIN32_CANCEL_SYNCHRONOUS_IO 1
+#else
+#define HAVE_API_SUPPORT_WIN32_CANCEL_SYNCHRONOUS_IO 0
+#endif
+#endif
+
+#ifndef HAVE_API_SUPPORT_WIN32_OPEN_THREAD
```
jaykrell: OpenThread is supported now. In order to work with 10240, guidance is declare it yourself. So then this isn't needed.

mono/utils/mono-threads-windows.c
lateralusX: Will keep since this we have users supporting older UWP versions and would like to use this file.

mono/metadata/w32subset.h
lateralusX: Was just put there for consistency, but agree, we could drop it.

mono/utils/mono-threads-windows.c
jaykrell: 10240 SDK has GetThreadContext.
SetThreadContext remains a problem though.

mono/metadata/w32subset.h
lateralusX: Will keep since this we have users supporting older UWP versions and would like to use this file.

lateralusX:
@luhenry approve/merge?

PR #13345, opened 2019-03-07
Better probing logic on Windows when Unicode Win32 symbols are used.

lateralusX:
Several of the corefx interop sources uses the Unicode names of Win32 symbols. Our current probing implementation on Windows will test the given name last when nomangle is not provided (default). On Windows that will be a large number of probes due to old stdcall name mangling rules. Fix makes sure we first test the symbol against its charset and the given symbol, before trying the stdcall name mangling rules. Will reduce the probing a lot on Windows for corefx Win32 interop classes.

### DIFF
--- a/mono/metadata/loader.c
+++ b/mono/metadata/loader.c
@@ -1688,7 +1688,7 @@ pinvoke_probe_for_symbol (MonoDl *module, MonoMethodPInvoke *piinfo, const char 
 
 	g_assert (error_msg_out);
 
-#ifdef TARGET_WIN32
+#ifdef HOST_WIN32
 	if (import && import [0] == '#' && isdigit (import [1])) {
 		char *end;
 		long id;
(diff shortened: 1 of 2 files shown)


mono/metadata/loader.c
```diff
@@ -1704,30 +1704,19 @@ pinvoke_probe_for_symbol (MonoDl *module, MonoMethodPInvoke *piinfo, const char
...
 		 * Search using a variety of mangled names
 		 */
-		for (mangle_charset = 0; mangle_charset <= 1; mangle_charset ++) {
-			for (mangle_stdcall = 0; mangle_stdcall <= 1; mangle_stdcall ++) {
-				gboolean need_param_count = FALSE;
+		for (int mangle_stdcall = 0; mangle_stdcall <= 1 && addr == NULL; mangle_stdcall++) {
 #ifdef TARGET_WIN32
-				if (mangle_stdcall > 0)
-					need_param_count = TRUE;
+			const int max_managle_param_count = (mangle_stdcall == 0) ? 0 : 256;
```
jaykrell: Is there no signature with a param count, that we can just probe for once?
If so, great, use that.
If not, can you allocate the buffer just once instead of in a loop?

This mangling should be limited not just to Windows, but Windows/x86.
I'm pretty sure WIN32 means any architecture Windows, not just 32bit.

jaykrell: Also would be good to put an `#if` around the loop bounds, or to terminate the second iteration right away.
i.e. be sure to iterate just once on all but Windows/x86.

Either like this:

```
#if win32 && x86
			for (mangle_stdcall = 0; mangle_stdcall <= 1; mangle_stdcall ++)
#else
			for (mangle_stdcall = 0; mangle_stdcall <= 0; mangle_stdcall ++)
#endif
{
```

or like this:

```
			for (mangle_stdcall = 0; mangle_stdcall <= 1; mangle_stdcall ++) { // no change this line 
#if !win32 || !x86
if (mangle_stdcall)
   break;
#endif 
```

lateralusX: Tried to not change the probing behavior around param count from original logic, but I agree that it should probably be applied to just 32-bit Windows.

jaykrell: I see you do use the signature for 0 params.
 But it looks like then the brute force path skips 0.

lateralusX: Signature mangling is only used when doing the std_call path and was kept as it was implemented originally.

jaykrell: I think it'd be best to start at -2 and have -2 and -1 both be special cases.
-2: unmangled; this will work the vast majority of the time, i.e. for the OS
-1: use the managed signature param count
>=0 the guessing that is present
And definitely this should be x86 only.
No other architecture, 32bit or 64bit, has multiple calling conventions or C mangling (ignoring amd64 vectorcall, granted).

jaykrell: Oh maybe the stdcall check flows from the caller and it is skipped then for amd64. I'd have to step through it. :(

lateralusX: Yes that implementation is a mess, so I didn't want to rewrite it to much :) I believe the Current change is solving the problem with unicode symbols and keep the old behavior intact.

lateralusX: I think we should keep changes to minimum, but agree that we should disable the parameter probing on 64-bit Windows.

lateralusX: Reduced parameter probing to Windows x86.

lateralusX:
@monojenkins rebuild failed

lateralusX:
@luhenry approve/merge?

PR #13348, opened 2019-03-07
Fix System.Decimal regression from .NET Core on big endian

NattyNarwhal:
Upstream must have rewritten, because the runtime checks were
replaced with the (not good for Mono) ifdefs.

Not throughly tested, but unbreaks build locally; waiting for CI
to come back with more results.

If it works, could someone get this upstream?

### DIFF
--- a/mcs/class/corlib/corert/Decimal.DecCalc.cs
+++ b/mcs/class/corlib/corert/Decimal.DecCalc.cs
@@ -21,11 +21,7 @@ namespace System
 
         internal int Scale => (byte)(flags >> ScaleShift);
 
-#if BIGENDIAN
-        private ulong Low64 => ((ulong)Mid << 32) | Low;
-#else
-        private ulong Low64 => (ulong)ulomidLE;
-#endif
+        private ulong Low64 => BitConverter.IsLittleEndian ? (ulong)ulomidLE : ((ulong)Mid << 32) | Low;
 
         private static ref DecCalc AsMutable(ref decimal d) => ref Unsafe.As<decimal, DecCalc>(ref d);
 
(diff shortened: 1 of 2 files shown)


EgorBo:
@monojenkins build failed

EgorBo:
@NattyNarwhal Thanks! Do you want me to test it on your BE machine via ssh?
I'll upstream it

NattyNarwhal:
@EgorBo please do!

mcs/class/corlib/corert/Decimal.DecCalc.cs
```diff
@@ -703,11 +705,7 @@ private static unsafe uint DivByConst(uint* result, uint hiRes, out uint quotien
 #else
                     // 32-bit RyuJIT doesn't convert 64-bit division by constant into multiplication by reciprocal. Do half-width divisions instead.
                     Debug.Assert(power <= ushort.MaxValue);
-#if BIGENDIAN
-                    const int low16 = 2, high16 = 0;
-#else
-                    const int low16 = 0, high16 = 2;
-#endif
+                    const int low16 = BitConverter.IsLittleEndian ? 0 : 2, high16 = BitConverter.IsLittleEndian ? 2 : 0;
```
EgorBo: hm.. does it compile? 😨

NattyNarwhal: it seemed to for me, but I suspect that may be because for all I know, BIT64 is defined

marek-safar: it's const (it cannot be initialized conditionally)

EgorBo: compiles for me inside Decimal too, kind of magic

NattyNarwhal: this doesn't compile when extracted; i'll de-const it, but I suspect Core might not like it being non-const

```
const.cs(5,49): error CS0133: The expression being assigned to `low16' must be a constant or default value
const.cs(6,50): error CS0133: The expression being assigned to `high16' must be a constant or default value
Compilation failed: 2 error(s), 0 warnings
```

```csharp
using System;

class c {
        public static void Main() {
                const int low16 = BitConverter.IsLittleEndian ? 0 : 2;
                const int high16 = BitConverter.IsLittleEndian ? 2 : 0;
        }
}
```

EgorBo: @NattyNarwhal it looks like .NET Core does, at least I can compile it locally (coreclr with the changes)

EgorBo: UPD: no it does not

EgorBo:
@monojenkins build failed

PR #13351, opened 2019-03-07
[2018-10][configure.ac] Fix configure checks for Mono.Native

akoeplinger:
When the configure checks were ported from configure.cmake in corefx a couple of mistakes happened:

- Using AC_TRY_COMPILE instead of AC_TRY_LINK:
  CMake's check_c_source_compiles() function compiles *and* links while
  AC_TRY_COMPILE only does the former. This hides potential linking issues
  which should fail the configure check.
- Using CFLAGS="-Werror -W....":
  This works in CMake but autoconf always appends the previously defined
  CFLAGS and since we set e.g. `-Wno-undefined` some configure checks
  would fail due to an unrelated warning being treated as an error.
  We need to only promot specific warnings to errors instead.

While looking at this I also found that we were setting CFLAGS after we already AC_SUBST() it, we need to do this very close to the end of configure.ac. Moved some code into that last block.

Finally, Android has special cases for sendfile and arc4random_buf where they're defined in the header but aren't actually there for certain API profiles. For now we disable arc4random_buf on Android and add a special check for sendfile.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3041,10 +3041,14 @@ if test x$host_win32 = xno; then
 	AC_CHECK_FUNC(ioctl,             [AC_DEFINE(HAVE_IOCTL, 1, [ioctl])])
 	AC_CHECK_FUNC(sched_getaffinity, [AC_DEFINE(HAVE_SCHED_GETAFFINITY, 1, [sched_getaffinity])])
 	AC_CHECK_FUNC(sched_setaffinity, [AC_DEFINE(HAVE_SCHED_SETAFFINITY, 1, [sched_setaffinity])])
+
+	if test "x$platform_android" != "xyes"; then
+		AC_CHECK_FUNC(arc4random_buf,    [AC_DEFINE(HAVE_ARC4RANDOM_BUF, 1, [arc4random_buf])])
+	fi
+
 	AC_CHECK_DECL(TIOCGWINSZ,        [AC_DEFINE(HAVE_TIOCGWINSZ, 1, [TIOCGWINSZ])], [], [[#include <sys/ioctl.h>]])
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -3418,6 +3411,7 @@ if test x$host_win32 = xno; then
 	])
 
 	AC_CHECK_DECL(IN_EXCL_UNLINK, [AC_DEFINE(HAVE_IN_EXCL_UNLINK, 1, [IN_EXCL_UNLINK])], [], [[#include <sys/inotify.h>]])
+	AC_CHECK_DECL(TCP_KEEPALIVE, [AC_DEFINE(HAVE_TCP_H_TCP_KEEPALIVE, 1, [TCP_KEEPALIVE])], [], [[#include <netinet/tcp.h>]])
```
alexischr: What is this line regarding?

akoeplinger: ah yeah, it's not really needed. it's an additional define that they added on the corefx side that was not ported to our configure.ac

marek-safar:
@akoeplinger @alexischr there is windows build failure

akoeplinger:
This is actually a legit failure that was uncovered by moving the `-Werror=return-type` further down so it is actually captured by AC_SUBST().

```
sgen-grep-binprot.c: In function 'is_always_match':
sgen-grep-binprot.c:185:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
sgen-grep-binprot.c: In function 'match_index':
sgen-grep-binprot.c:507:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
sgen-grep-binprot.c: In function 'is_vtable_match':
sgen-grep-binprot.c:570:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
```

alexischr:
I am concerned about fixing conf checks on this branch other than the ones problematic on Android (`sendfile` and `arc4random_buf`) but if this builds I'll take it :)

PR #13363, opened 2019-03-08
[corlib] Remove Interop.MountPoints.cs from .sources

akoeplinger:
We no longer need it since https://github.com/mono/mono/pull/13125 and having it included makes xamarin-macios still complain about the missing symbol.


akoeplinger:
@monojenkins backport 2019-02

akoeplinger:
@monojenkins commit csproj

PR #13366, opened 2019-03-08
Include mono-forward.h public header on Windows.

lateralusX:


lateralusX:
@monojenkins rebuild failed

PR #13367, opened 2019-03-08
Fix build warning in mono-error.h.

lateralusX:
Warning (at least on Windows) due to unamed struct:

C4201, nonstandard extension used : nameless struct/union

Since the header can be consumed by embedders this warning should be disabled.


lateralusX:
@monojenkins rebuild failed.

PR #13373, opened 2019-03-08
[debugger] Fixing two crashes while debugging an Android app.

thaystg:
-> Doing stepping into in VSM in some situation the frame saved in TLS is not synchronised with what is really been executed in the main thread. This makes the debugger-agent crashes because it tries to get variable info in some memory that is not available anymore.
	-> To fix it I forced the update of stack when CMD_THREAD_GET_FRAME_INFO is called.

-> Doing step over in Visual Studio for Windows, if you have the threads debugger window enabled, VSW calls frame_commands for each thread that is showed, and if the thread is not really_suspended it tries to get variable info from a memory that is not available anymore because the thread is not suspended yet.
	-> To fix it I don't send variable info of a frame if the thread is not really_suspended and doesn't have an async_state valid.

#Fixes 12843

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -3216,7 +3216,7 @@ compute_frame_info_from (MonoInternalThread *thread, DebuggerTlsData *tls, MonoT
 }
 
 static void
-compute_frame_info (MonoInternalThread *thread, DebuggerTlsData *tls)
+compute_frame_info (MonoInternalThread *thread, DebuggerTlsData *tls, gboolean force_update)
 {
 	ComputeFramesUserData user_data;
 	GSList *tmp;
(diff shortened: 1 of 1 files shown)


marek-safar:
@monojenkins build

thaystg:
@monojenkins build failed

mono/mini/debugger-agent.c
```diff
@@ -4227,7 +4227,7 @@ ss_calculate_framecount (void *the_tls, MonoContext *ctx, gboolean force_use_ctx
 
 	if (force_use_ctx || !tls->context.valid)
 		mono_thread_state_init_from_monoctx (&tls->context, ctx);
-	compute_frame_info (tls->thread, tls);
+	compute_frame_info (tls->thread, tls, FALSE);
```
alexanderkyte: Could you say why it's false or true for the last arg in a line comment here and elsewhere?

thaystg:
@monojenkins build failed

thaystg:
@monojenkins backport 2018-10

thaystg:
@monojenkins backport 2018-12

thaystg:
@monojenkins backport 2019-02

thaystg:
@monojenkins backport 2018-10

thaystg:
@monojenkins backport 2018-12

thaystg:
@monojenkins backport 2019-02

lambdageek:
@monojenkins backport 2018-08

PR #13376, opened 2019-03-09
[wasm] Remove all use of ContainsKey from js bindings

lewing:
A few places in in the JS Bindings were using Dictionary.ContainsKey then immediately looking up the object, use TryGetValue instead.

### DIFF
--- a/sdks/wasm/framework/src/WebAssembly.Bindings/Runtime.cs
+++ b/sdks/wasm/framework/src/WebAssembly.Bindings/Runtime.cs
@@ -79,9 +79,7 @@ namespace WebAssembly {
 		static int BindJSObject (int js_id)
 		{
 			JSObject obj;
-			if (bound_objects.ContainsKey (js_id))
-				obj = bound_objects [js_id];
-			else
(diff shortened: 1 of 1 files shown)


lewing:
@monojenkins build failed

lewing:
@monojenkins build failed

lewing:
@monojenkins build failed

kjpou1:
Hey guys

Can we hold off merging this until it is tested.  There were some problems a while back with replacing the ContainsKey with TryGetValue that was causing the object to be held onto with the Dictionary.

lewing:
Sure, this isn't urgent.  Do you remember where you were seeing the problem?

lewing:
@monojenkins build failed

kjpou1:
@monojenkins build

kjpou1:
> Sure, this isn't urgent. Do you remember where you were seeing the problem?

@lewing yes but I noticed after looking at the PR that non of the xxxFree methods were changed.  I went ahead and tested this anyway within the open TypedArray PR as well.  

It is approved.  👍

PR #12619, opened 2019-01-25
Switch mono webforms with refsrc webforms

steveisok:
There's a lot here... Hopefully it will be (relatively) easy enough to review.

It's important to note that this change makes System.Configuration point at corefx System.Configuration.ConfigurationManager.  Since System.Web relies heavily on System.Configuration and mono's version was a little different, I felt like this change was essential.  

Please note that there are some classes that are not in refsrc, but are needed.  For those, I just used decompiled and took what I needed.  It would be helpful to know who to pester to fill in these gaps.

Thanks for everyone's help.

PR #13383, opened 2019-03-10
Add missing refsrc ToolLocationHelper methods 

steveisok:
Relates to https://github.com/mono/mono/pull/12619

The methods added will be used by refsrc System.Web.Compliation.AssemblyResolver.  They really don't mean anything in a mono context, so that's why they return empty lists.


marek-safar:
@monojenkins commit apidiff

PR #13391, opened 2019-03-11
[netcore] Ongoing work.

vargaz:


mcs/class/System.Private.CoreLib/System.Collections.Generic/EqualityComparer.cs
```diff
@@ -47,6 +47,16 @@ static EqualityComparer<T> CreateComparer ()
 			
 			return new ObjectEqualityComparer<T>();
 		}
+
+		internal virtual int IndexOf (T[] array, T value, int startIndex, int count)
```
marek-safar: Why virtual?

mcs/class/System.Private.CoreLib/System.Runtime.CompilerServices/RuntimeHelpers.cs
```diff
@@ -68,10 +68,25 @@ public static void PrepareDelegate (Delegate d)
...
+			}
 		}
 
 		public static void PrepareMethod (RuntimeMethodHandle method, RuntimeTypeHandle[] instantiation)
 		{
+			if (method.IsNullHandle ())
+				throw new ArgumentException (SR.Argument_InvalidHandle);
+			unsafe {
+				int length;
+				IntPtr[] instantiations = RuntimeTypeHandle.CopyRuntimeTypeHandles (instantiation, out length);
```
marek-safar: nit: `out int length` instead

mcs/class/System.Private.CoreLib/System.Collections.Generic/EqualityComparer.cs
vargaz: So it can have optimized implementations later.

PR #13392, opened 2019-03-11
[netcore] Fix build.

vargaz:


mcs/class/System.Private.CoreLib/System/Buffer.cs
```diff
@@ -52,7 +52,7 @@ public static void BlockCopy (Array src, int srcOffset, Array dst, int dstOffset
 
 		static bool IsPrimitiveTypeArray (Array array)
 		{
-			throw new NotImplementedException ();
+			return array.GetType ().GetElementType ().IsPrimitive;
```
EgorBo: I wonder why .net core has to do a pinvoke for it, performance reasons?

vargaz: This code needs to be optimized later.

filipnavara: @EgorBo Yes, performance reasons, avoiding reflection. Also it's FCall, not P/Invoke, which usually points to something performance sensitive especially when the native code doesn't erect a frame using `HELPER_METHOD_FRAME_BEGIN_*`.

EgorBo: maybe we should document somewhere or use a special attribute to mark places where we put a naive implementation to revise it later?

PR #13399, opened 2019-03-11
Fix a possible crash while printing a warning

joshpeterson:
The format string was incorrect in this case, which leads to a crash
in Unity. Simply remove the unused `%s` format specifier here.

### DIFF
--- a/mono/metadata/threadpool-io.c
+++ b/mono/metadata/threadpool-io.c
@@ -198,7 +198,7 @@ selector_thread_wakeup_drain_pipes (void)
 			break;
 		if (received == SOCKET_ERROR) {
 			if (WSAGetLastError () != WSAEINTR && WSAGetLastError () != WSAEWOULDBLOCK)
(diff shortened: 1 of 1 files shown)


joshpeterson:
A side note: Is there any interest in using the `format` attribute from Clang and GCC to attempt to catch issues like this at compile time?

https://clang.llvm.org/docs/AttributeReference.html#format

vargaz:
We already use that in some places, i.e. G_ATTR_FORMAT_PRINTF in glib.h, but not in all functions which need it.

joshpeterson:
Ahh, interesting. We don't have that code in our fork yet.

vargaz:
@monojenkins build failed

joshpeterson:
I'm not sure why the "OS X x64 Android SDK" check is not green. Is there a way to re-run it? It has been "pending" for a while now.

directhex:
The build in question is this one: https://jenkins.mono-project.com/view/Pull%20requests/job/test-mono-pull-request-amd64-osx-products-sdks-android/9929

I'll try re-running the job.

directhex:

PR #13410, opened 2019-03-12
[netcore] Implement Type.GetType () overloads with assembly/type resolve delegates.

vargaz:
Instead of extending the runtime gettype implementation, port the type parsing code
to C#.


mcs/class/System.Private.CoreLib/System/Type.cs
```diff
@@ -104,4 +117,386 @@ internal virtual string FormatTypeName (bool serialization)
...
+            foreach (char c in name) {
+                if (Array.IndexOf<char> (SPECIAL_CHARS, c) >= 0)
+                    sb.Append ('\\');
+                sb.Append (c);
+            }
+
+			return sb.ToString ();
+        }
+
+		class ParsedName {
```
marek-safar: Wouldn't be better to be a struct or ValueTuple?

mcs/class/System.Private.CoreLib/System/Type.cs
```diff
@@ -104,4 +117,386 @@ internal virtual string FormatTypeName (bool serialization)
 			return !object.ReferenceEquals (left, right);
 		}
 	}
+
+	internal sealed class TypeNameParser
```
marek-safar: Make it static and move to separate file

mcs/class/System.Private.CoreLib/System/Type.cs
```diff
@@ -104,4 +117,386 @@ internal virtual string FormatTypeName (bool serialization)
...
+                if (Array.IndexOf<char> (SPECIAL_CHARS, c) >= 0)
+                    sb.Append ('\\');
+                sb.Append (c);
+            }
+
+			return sb.ToString ();
+        }
+
+		class ParsedName {
+			public List<string> Names;
```
marek-safar: Use `ReadonlySpan<char>` instead

mcs/class/System.Private.CoreLib/System/Type.cs
```diff
@@ -104,4 +117,386 @@ internal virtual string FormatTypeName (bool serialization)
...
+			public List<ParsedName> TypeArguments;
+			public List<int> Modifiers;
+			public string AssemblyName;
```
marek-safar: Use ReadonlySpan<char> instead

mcs/class/System.Private.CoreLib/System/Type.cs
vargaz: Those values are eventually passed to methods expecting a string, so a string needs to be created anyway.

mcs/class/System.Private.CoreLib/System/Type.cs
vargaz: Fixed.

mcs/class/System.Private.CoreLib/System/Type.cs
vargaz: We use null for error checking.

marek-safar: Use can still use nullable for that

mcs/class/System.Private.CoreLib/System/Type.cs
marek-safar: Not always, there is foreach conversion in EscapeTypeName

mcs/class/System.Private.CoreLib/System/Type.cs
vargaz: Thats doable, not sure its worth it, this is not really perf sensitive code.

mcs/class/System.Private.CoreLib/System/Type.cs
vargaz: That code only runs if the string contains characters to be escaped, which is pretty rare.

lambdageek:
We already have a managed type parser based on the C code: mcs/class/corlib/System/TypeSpec.cs.  I guess now we have 2.

marek-safar:
@lambdageek maybe we could have custom ALC context and handle that in managed code as well for core

PR #13414, opened 2019-03-12
Move drawing types for XI/XM/wasm/orbis/unreal to System.Drawing.Common

akoeplinger:
For XI/XM they were in Xamarin.iOS.dll/OpenTK-1.0.dll before.

For wasm/orbis/unreal we remove System.Drawing.dll from the profiles and
add the types to System.Drawing.Common.dll instead to align with the other
mobile profiles (only "net_4_x" and "build" still have System.Drawing.dll).


akoeplinger:

PR #13417, opened 2019-03-12
[corlib] Define FEATURE_DEFAULT_INTERFACES so RuntimeFeatures indicates DIMs are supported.

vargaz:


marek-safar:

PR #13433, opened 2019-03-13
[netcore] Ongoing work.

vargaz:


vargaz:
Note that most of the changes/fixes are inside an #ifdef ENABLE_NETCORE. This is to avoid breaking code which might depend on the old behavior.

mcs/class/System.Private.CoreLib/System.Collections.Generic/EqualityComparer.cs
```diff
@@ -48,6 +48,7 @@ static EqualityComparer<T> CreateComparer ()
 			return new ObjectEqualityComparer<T>();
 		}
 
+		[MonoTODO ("Add specialized versions")]
```
marek-safar: Don't use MonoTODO, just use normal comment

mcs/class/System.Private.CoreLib/System.Runtime.Loader/AssemblyLoadContext.cs
```diff
@@ -25,6 +26,7 @@ Assembly InternalLoadFromPath (string assemblyPath, string nativeImagePath)
 		{
 			StackCrawlMark stackMark = StackCrawlMark.LookForMyCaller;
 
+			assemblyPath = assemblyPath.Replace ('\\', Path.DirectorySeparatorChar);
```
marek-safar: Could you add a comment on why is it needed?

mcs/class/System.Private.CoreLib/System/Array.cs
```diff
@@ -112,7 +113,7 @@ public static void Copy (Array sourceArray, int sourceIndex, Array destinationAr
 
 			// re-ordered to avoid possible integer overflow
 			if (source_pos > sourceArray.Length - length)
-				throw new ArgumentException ("length");
+				throw new ArgumentException (SR.Arg_LongerThanSrcArray, "sourceArray");
```
marek-safar: use `nameof`

mcs/class/System.Private.CoreLib/System/Buffer.cs
```diff
@@ -20,11 +20,15 @@ public static void BlockCopy (Array src, int srcOffset, Array dst, int dstOffset
...
-				throw new ArgumentOutOfRangeException (SR.ArgumentOutOfRange_MustBeNonNegInt32, nameof (srcOffset));
+				throw new ArgumentOutOfRangeException (nameof (srcOffset), SR.ArgumentOutOfRange_MustBeNonNegInt32);
 			if (dstOffset < 0)
-				throw new ArgumentOutOfRangeException (SR.ArgumentOutOfRange_MustBeNonNegInt32, nameof (dstOffset));
+				throw new ArgumentOutOfRangeException (nameof (dstOffset), SR.ArgumentOutOfRange_MustBeNonNegInt32);
 			if (count < 0)
-				throw new ArgumentOutOfRangeException (SR.ArgumentOutOfRange_MustBeNonNegInt32, nameof (count));
+				throw new ArgumentOutOfRangeException (nameof (count), SR.ArgumentOutOfRange_MustBeNonNegInt32);
+			if (!(src is byte[]))
+				throw new ArgumentException (SR.Arg_MustBePrimArray, "src");
```
marek-safar: Use `nameof`

mcs/class/System.Private.CoreLib/System.Runtime.Loader/AssemblyLoadContext.cs
vargaz: Some tests pass in windows paths to it, not sure how else to handle it.

mcs/class/System.Private.CoreLib/System.Collections.Generic/EqualityComparer.cs
vargaz: Will use // MONOTODO: in the future, so it can be searched for.

mcs/class/System.Private.CoreLib/System.Runtime.Loader/AssemblyLoadContext.cs
marek-safar: Are they windows specific? Could you share an example?

vargaz: https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/Type/TypeTests.cs#L300

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #13437, opened 2019-03-13
Fix generating `.dSYM` for GTK assemblies

DavidKarlas:
Problem is that https://github.com/mono/bockbuild/blob/14535c270ea9f8d536efe6852700d8d1cb657c1f/packages/gtk%2B.py#L4 passes `gtk` for repository name to `GitHubPackage` which then uses this as package name. So when it's decided if `-g` should be added at https://github.com/mono/bockbuild/blob/5c2378112dc681cea0aa5c2447661fb13adbc46b/bockbuild/darwinprofile.py#L158-L159 it's false hence GTK assemblies don't get debug symbols.


directhex:
@monojenkins build pkg

directhex:
@monojenkins build failed

DavidKarlas:
@monojenkins build failed

directhex:
Gonna merge, the Linux ARM failure is unrelated

PR #13448, opened 2019-03-13
[llvm] Fix receiving rgctx argument in DIM methods, they receive the mrgctx even if they are not generic.

vargaz:


vargaz:
@monojenkins build failed

PR #13461, opened 2019-03-14
Fix missing llc options for Windows x64 full AOT + LLVM mini regression tests.

lateralusX:
Turns out that https://github.com/mono/mono/pull/13444 didn't solve the issue fully since mini regression tests are not using the AOT_BUILD_FLAGS variable set in configure script. The mini regression tests do however use the values in the INVARIANT_AOT_OPTION, so adding the llcopt into that will make sure the flag ends up in mini regression tests as well as other CI test suites.


lateralusX:
@monojenkins rebuild failed

PR #13469, opened 2019-03-15
[netcore] Disable runtime cleanup on netcore, coreclr does not cleanup either. Also disable aborting background threads, its no longer needed.

vargaz:

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -3722,6 +3722,11 @@ mono_thread_manage (void)
 		mono_thread_execute_interruption_void ();
 	}
 
+#ifndef ENABLE_NETCORE
+	/*
+	 * Under netcore, we don't abort any threads, just exit.
+	 * This is not a problem since we don't do runtime cleanup either.
+	 */
 	/* 
 	 * Remove everything but the finalizer thread and self.
 	 * Also abort all the background threads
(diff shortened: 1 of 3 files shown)


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

jaykrell:
Wow. this might fix the Alpine Linux problem too, a hang at shutdown, related to alternative libc.

vargaz:
This is a netcore only change.

jaykrell:
Understood. This might fix it for netcore.

marek-safar:
@lewurm/@lambdageek could you please review

mono/metadata/threads.c
```diff
@@ -3722,6 +3722,11 @@ mono_thread_manage (void)
 		mono_thread_execute_interruption_void ();
 	}
 
+#ifndef ENABLE_NETCORE
```
filipnavara: Is there a reason to guard this with ENABLE_NETCORE instead of DISABLE_CLEANUP?

jaykrell: That name is too short, but it is a valid point -- to compose out of a finer grained set of names, vs. one omnibus one. The fine grained ifdef raises the question as to what combinations of things work or are tested. But that does not necessarily bother people -- witness autoconf.

filipnavara: DISABLE_CLEANUP is already used at the other place in this commit, so it made me wonder whether all the combinations are actually necessary.

vargaz: Doesn't matter too much, DISABLE_CLEANUP is a --enable-minimal feature which can be used by anyone, while not aborting threads is  a part of netcore.

PR #13474, opened 2019-03-15
Couple of fixes to LLVM AOT compiler on Windows x64.

lateralusX:
- Enable MSVC objects on LLVM >= 6.00

- Always used PIC relocation-model on Windows x64 due to LLVM problem with
static triggering incorrect ADDR32 relocation entries in 64-bit objects.

- Add CodeView symbol format instead of DWARF for Windows x64.

- Fix LLVM outfile when doing static link to not end up in temp directory.

- Only add /DEFAULTLIB when using MSVC and not doing static compile. For static compile the needed libraries must be added when objects get's linked and since the needed libraries are c-runtime libraries, should be transparent for most applications consuming the object files.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -9509,6 +9509,11 @@ emit_llvm_file (MonoAotCompile *acfg)
 	if (acfg->aot_opts.mtriple)
 		g_string_append_printf (acfg->llc_args, " -mtriple=%s", acfg->aot_opts.mtriple);
 
+#if defined(TARGET_X86_64_WIN32_MSVC) && LLVM_API_VERSION >= 600
+	if (!acfg->aot_opts.mtriple)
+		g_string_append_printf (acfg->llc_args, " -mtriple=%s", "x86_64-pc-windows-msvc");
+#endif
+
 	g_string_append (acfg->llc_args, " -disable-gnu-eh-frame -enable-mono-eh-frame");
 
 	g_string_append_printf (acfg->llc_args, " -mono-eh-frame-symbol=%s%s", acfg->user_symbol_prefix, acfg->llvm_eh_frame_symbol);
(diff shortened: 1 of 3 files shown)


lateralusX:
@monojenkins rebuild failed

mono/mini/aot-compiler.c
```diff
@@ -9522,8 +9527,8 @@ emit_llvm_file (MonoAotCompile *acfg)
 	g_string_append_printf (acfg->llc_args, " -no-x86-call-frame-opt");
 #endif
 
-#if ( defined(TARGET_MACH) && defined(TARGET_ARM) ) || defined(TARGET_ORBIS)
-	/* ios requires PIC code now */
+#if ( defined(TARGET_MACH) && defined(TARGET_ARM) ) || defined(TARGET_ORBIS) || defined(TARGET_X86_64_WIN32_MSVC)
+
```
vargaz: Extra empty line.

mono/mini/aot-compiler.c
```diff
@@ -13471,7 +13482,11 @@ emit_aot_image (MonoAotCompile *acfg)
 			acfg->tmpbasename = g_build_filename (temp_path, "temp", NULL);
 			acfg->tmpfname = g_strdup_printf ("%s.s", acfg->tmpbasename);
 			acfg->llvm_sfile = g_strdup_printf ("%s-llvm.s", acfg->tmpbasename);
-			acfg->llvm_ofile = g_strdup_printf ("%s-llvm." AS_OBJECT_FILE_SUFFIX, acfg->tmpbasename);
+
+			if (acfg->aot_opts.static_link)
+				acfg->llvm_ofile = g_strdup_printf (acfg->aot_opts.llvm_outfile);
```
vargaz: Why is this needed ?

lateralusX: If I specify the out file for the LLVM object (--llvm-outfile), it will still be generated in the temp folder using the tempbase name. If I specify the out file for the none LLVM object (--outfile) it will be generated as specified by the argument. For none static build this is not a problem since the shared library will be the outfile (the rest will only be temporary artifacts anyway), but for static build, it should be placed as specified in llvm-outfile.

mono/mini/aot-compiler.c
lateralusX: Fixed.

lateralusX:
@monojenkins rebuild failed

lateralusX:
@vargaz good to go?

mono/mini/aot-compiler.c
vargaz: We are using these options in ios in static mode, wondering why that works then.

lateralusX: OK, that is strange since it will then end up in a temp folder using the default naming schema. Lets do like this, I can keep that change downstream for the platform that currently needs it and remove it from this PR for now to reduce risk of regression.

lateralusX: Removed that change from this PR.

lateralusX:
@monojenkins rebuild failed

lateralusX:
@vargaz anything more to adjust on this PR?

PR #13475, opened 2019-03-15
Fix a couple of build warnings under MSVC.

lateralusX:


mono/mini/aot-compiler.c
```diff
@@ -4832,7 +4832,9 @@ add_wrappers (MonoAotCompile *acfg)
 					g_assert (!strcmp (name, "ExportSymbol"));
 
 					/* load_cattr_value (), string case */
-					g_assert (*named != (char)0xff);
+MONO_DISABLE_WARNING (4310) // cast truncates constant value
+					g_assert (*named != (char)0xFF);
```
jaykrell: Maybe cast pointer to ‘(const unsigned char*)’?

jaykrell: Or guint8.

jaykrell: And then remove the char cast.

lateralusX: I experimented with similar casts in a different area a while back and got into some incorrect behaviors on some platform (don't recall what platform) failing some tests, so fallback to just disable the warning. I could of course retry the cast, but then I believe I should do it at the other places as well, since we at least have tests covering those code paths and failing if we don't get identical results on all platforms.

jaykrell: "jaykrell feel free to revisit every `MONO_DISABLE_WARNING` for a 'better' fix, that neither quashes warnings nor changes semantics on any platform"?

mono/mini/method-to-ir.c
```diff
@@ -1666,7 +1667,9 @@ mini_emit_runtime_constant (MonoCompile *cfg, MonoJumpInfoType patch_type, gpoin
 	MonoInst *ins;
 
 	if (cfg->compile_aot) {
+MONO_DISABLE_WARNING (4306) // 'type cast': conversion from 'MonoJumpInfoType' to 'MonoInst *' of greater size
```
jaykrell: `GINT_TO_POINTER` or `GUINT_TO_POINTER`?

PR #13476, opened 2019-03-15
[eglib] Remap g_utf16_len and g_utf16_len to monoeg_ prefix

akoeplinger:
They fail a xamarin-macios public symbols check otherwise (monoeg_ is whitelisted).


jaykrell:
Why g_printv 11 months w/o problem?
utf16_len 4 months, I guess no integration since then?

akoeplinger:
g_printv was explicitly whitelisted in the test, probably the previous integrator didn't know that we should do the remapping instead. g_utf16_len is new in this integration.

jaykrell:
Hopefully scrub the whitelist.

PR #13482, opened 2019-03-15
Windows full AOT build scripts update.

lateralusX:
* Only do clean/rebuild when an assembly changed.
* Add wrappers for different full AOT build configurations.
* Make scripts callable from different working directory.


lateralusX:
@monojenkins build failed

PR #13489, opened 2019-03-16
Cleanup mono_array_new_[1-4] to share code and remove alloca.

jaykrell:


marek-safar:
Thanks!

luhenry:
@marek-safar this PR had requested changes on the previous iteration https://github.com/mono/mono/pull/12198, and he didn't even bother changing anything. I would be more careful in accepting his changes next time, so we don't want to fall in the same trap as previously. Good luck! :)

jaykrell:
Reasons were given as to why not make the changes.

PR #13495, opened 2019-03-17
[netcore] Fix case when InlineData's type != ParameterType

EgorBo:
As you can see here https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/Reflection/AssemblyFlagsAttributeTests.cs#L19
InlineData contains enums and an integer, without this change that integer will cause a compilation error (runner.cs).

Found and tested against the latest CoreFX repo 
```
RUN: 6261
FAILURES: 6
SKIPPED: 184
```


vargaz:
Looks ok, please use the mono conventions (K&R indentation, spaces before braces).

EgorBo:
@vargaz fixed

PR #13509, opened 2019-03-18
[netcore] Ongoing work.

vargaz:


EgorBo:
`make all` builds only System.Runtime.Tests
maybe it should build all `TEST_SUITES` ?

PR #13514, opened 2019-03-18
[aot] Fix a few memory leaks

lambdageek:
1. aot-compiler: don't leak `MonoCustomAttrInfo`

2. aot-runtime: don't leak sigs when decoding wrapper references.  All of these cases we decode a sig (which uses malloc), and then call some method that immediately calls `mini_get_underlying_signature` which dups the sig, so the one from the decode isn't needed anymore.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -4547,7 +4547,6 @@ add_wrappers (MonoAotCompile *acfg)
 	for (i = 0; i < acfg->image->tables [MONO_TABLE_TYPEDEF].rows; ++i) {
 		ERROR_DECL (error);
 		MonoClass *klass;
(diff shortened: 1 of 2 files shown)


mono/mini/aot-compiler.c
```diff
@@ -4587,6 +4588,7 @@ add_wrappers (MonoAotCompile *acfg)
 					add_method (acfg, wrapper);
 					add_method (acfg, del_invoke);
 				}
+				g_free (cattr);
 			}
```
vargaz: This should be mono_custom_attrs_free ().

lambdageek:
Windows i386 System.Runtime.Remoting failures are unrelated - it's an IPC channel reuse flakiness.

PR #13517, opened 2019-03-18
[runtime] Add backwards-compatible signature for reflection-only telemetry icall

alexanderkyte:
This was requested so that icall signatures don't change. 

VS4M wants to be able to support multiple monos at once.

### DIFF
--- a/mcs/class/corlib/Mono/Runtime.cs
+++ b/mcs/class/corlib/Mono/Runtime.cs
@@ -147,6 +147,11 @@ namespace Mono {
 			SendMicrosoftTelemetry (payload_str, portable_hash, unportable_hash);
 		}
 
+		static void EnableMicrosoftTelemetry (string appBundleID_str, string appSignature_str, string appVersion_str, string merpGUIPath_str, string eventType_str, string appPath_str)
+		{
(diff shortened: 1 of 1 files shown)


alexanderkyte:
@Therzok

Therzok:
I asked if this was possible, simply due to the fact that it is the simpler and more efficient approach, taking both sides into account.

Therzok:
Looks great!

marek-safar:
@monojenkins build

akoeplinger:
@Therzok are you sure your code can deal with multiple overloads? I took a quick look and I think it'll throw.

Therzok:
Seems like it doesn't: https://github.com/xamarin/md-addins/blob/52d9134291a1eed3753d435c17ea27688c2fc076/Xamarin.Ide/Xamarin.IdeCustomization/Xamarin.IdeCustomization/XamarinCustomizer.cs#L82

Sigh.

alexanderkyte:
@Therzok do you still prefer to keep around a legacy icall with the same signature if that's the case, or would you rather we use the length-check-based method we have open in the other PR?

Therzok:
At this point, given we'll have to do parameter binding on the IDE side, anyway, don't think we need this.

Issue #13523, opened 2019-03-19
[WinForms] One can not send inter-process window messages


NVoronchev:
There is no Linux-equivalent of Win32 API methods like `EnumWindows`, `RegisterWindowMessage` and `PostMessage` to implement inter-process communication.

filipnavara:
Would be nice to actually have some context for this. What's your use-case where you'd use such functionality?

plaes:
> Would be nice to actually have some context for this. What's your use-case where you'd use such functionality?

Legacy applications? Even stubbed would be nice to have.

NVoronchev:
There were some discussion on this topic in the PR #13525.

PR #13525, opened 2019-03-19
[WinForms] Inter-process window messages (#13523)

NVoronchev:
Here I propose the Linux-equivalent of Win32 API methods `EnumWindows`, `RegisterWindowMessage` and `PostMessage`. This allows one to reserve window message and send it to all mono top-level windows.

Newly introduced assembly `Mono.PortabilityUtils` is a helper library. It can be potentially expanded by other methods.

Fixes #13153


filipnavara:
Firstly, the new project name is not very good. The functionality is tied to WinForms implementation and that's exactly the purpose of the original `Mono.WinformsSupport` project that was abandoned over the years. `Mono.WinformsSupport` used to expose the `XplatUI` interface for use in 3rd-party projects.

Secondly, I don't see much of a use for this. It would work only between two Mono applications (if at all), only on X Windows, and you'd still need to make changes to the application to use these APIs. The whole point of System.Windows.Forms in Mono is to run existing applications. There should be a compelling reason to add something new instead of using a proper cross-platform inter-process communication (such as named pipes).

NVoronchev:
Hi, @filipnavara!

I am engaged in porting an existed .Net-application to Linux. This app supports multi-instances launching. Each new instance notify all previously started ones. Window messages is common way to do it in Windows. One just need to override `WndProc` and send a message by means of `PostMessage`.

In Linux this feature is not available now. I have consider couple solutions:

- Pipes. It is a good solutions, but this ways lead to the following issue. Mono in Linux doesn't allows more than one server listen the single pipe. Hence each application should create its own pipe and client must find pipes opened by other instances. But .Net doesn't provide the proper way to do it. Windows recipe like `System.IO.Directory.EnumerateFiles(@"\\.\pipe\", "MyAppPipes-*")` doesn't work in Linux too. BTW, I guess, it is a good point to add to Mono a cross-platform method, which lists opened pipes.

- .Net-equivalent of Win32 native functions `EnumWindows`, `RegisterWindowMessage`, `PostMessage` in the dedicated assembly. This assembly will work correctly both in Linux and Windows, because of the class `XplatUIWin32` redirects this call to native functions.

- Direct usage of network sockets or shared memory seems too cumbersome.

The drawback of pipes method is some additional code in user application: one need to create async pipe listener. Meanwhile, to use window messages one only need to switch his application from native calls to .Net alternatives.

filipnavara:
My concern is that this really is not very versatile solution. Whatever you are implementing here could just as easily be done in the ported application where it will likely be more flexible and also easier to maintain.

The proposed changes are problematic because you are implementing only a narrow subset of the APIs that fit only your specific use case. It gives the false impression that the APIs would behave consistently across platforms which is just not the case.

Aside from that, my other concerns from previous comment apply. There's already `Mono.WinformsSupport`, which is THE layer for exposing `XplatUI` for applications that need to call the Win32-like APIs directly. If your solution were to be implemented it should stick to that as the external interface.

steveisok:
In general, I agree with @filipnavara's assessment.  I appreciate the contribution, but this is not something we can accept.

PR #13528, opened 2019-03-19
[wasm] Javascript Core Function

kjpou1:
Support JavaScript [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function).

- Create a new Function:
``` csharp
var sum = new Function("a", "b", "return a + b");
Console.WriteLine($"Sum: {sum.Call(null, 3, 5)}");
```
```
console output ---

Sum: 8
```

- Call
- Apply
- Bind

- Marshalling of Functions passed from JavaScript to C#

``` csharp
            var math = (JSObject)Runtime.GetGlobalObject("Math");
            var min = (Function)math.GetObjectProperty("min");
            Console.WriteLine($"Min: {min.Apply(null, new object[] { 5, 6, 2, 3, 7 })}");
```
```
console output ---

Min: 2
```

- Add tests to binding-tests.cs.

- Fix up the TypedArray tests also included.


sdks/wasm/driver.c
```diff
@@ -761,6 +761,18 @@ mono_wasm_get_obj_type (MonoObject *obj)
 	}
 }
 
+EMSCRIPTEN_KEEPALIVE int
+mono_wasm_is_simple_array (MonoObject *obj)
+{
```
vargaz: You can do if (a is Array arr) { if (arr.Rank == 1 && a.GetLowerdBound (0) == 0) ... }

kjpou1: @vargaz I can do `if (a is Array arr)` in a `.c` program?  That looks more like c-sharp.

vargaz: I mean the calling c# code can do that instead of adding an icall.

lewing:
@monojenkins build failed

kjpou1:
@monojenkins build failed

sdks/wasm/driver.c
kjpou1: Address icall in recent commit.

kjpou1:
@monojenkins build failed

PR #13529, opened 2019-03-19
Fix mixup of Global Free and CoTaskMem Alloc in Marshal's Secure String implementation.

lateralusX:
Current implementation uses AllocCoTaskMem for secure strings in SecureStringToGlobalAllocAnsi/Unicode, but ZeroFreeGlobalAllocAnsi/Unicode will always free using FreeHGlobal. This will cause memory leaks and corruption on platforms where AllocCoTaskMem and AllocHGlobal are backed by different allocators.


lateralusX:
@monojenkins rebuild failed

PR #13536, opened 2019-03-19
Fix handle leak in cominterop

blattersturm:
The function `cominterop_get_interface` in `mono/metadata/cominterop.c` had a number of changes made to it in 18b6846a3a2218525b7335a1f0515de93a1e44ca.

One of these seems to have missed a `return itf;` statement preceding `HANDLE_FUNCTION_RETURN_VAL`. Since `HANDLE_FUNCTION_RETURN_VAL` is meant to clean up the handle created in this function, this'd lead to a leaking handle.

This mainly hurt when running COM interface routines from finalizers, since the finalizer thread would end up owning a GC handle, and then hitting an assertion expecting the handle stack to be empty, but would be practically unnoticeable otherwise.


lewurm:
@monojenkins approve

vargaz:
@monojenkins build failed

jaykrell:
@monojenkins squash

PR #13539, opened 2019-03-19
[netcore] fix typo

EgorBo:
I could not run tests because of it (I guess the typo was made when gen-xunit-runner was rewritten to use Roslyn instead of CodeDom).


EgorBo:
oh, wait, it now compiles but is not 100% correct

PR #13543, opened 2019-03-20
[netcore] Initial MemberDataAttribute support

EgorBo:
This PR adds ~2200 tests to System.Runtime.Tests and ~1000 to System.Collections.Tests

### DIFF
--- a/netcore/gen-xunit-runner/Program.cs
+++ b/netcore/gen-xunit-runner/Program.cs
@@ -41,9 +41,7 @@ class MsgSink : IMessageSink {
 }
 
 class Program
-{ 
-
-	
+{
 	static int nskipped = 0;
 
 	// The template for the whole progeam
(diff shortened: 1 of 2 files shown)


netcore/gen-xunit-runner/Program.cs
```diff
@@ -315,6 +318,43 @@ static int Main (string[] args)
...
+							// only primitive types + string are supported at this moment
+							if (pType.IsPrimitive && notSupportedPrimitiveTypes.Contains (pType))
+								goto notsupported;
+							
+							if (pType != typeof (string) && pType != typeof (decimal))
+								goto notsupported;
+						}
+
+						foreach (object[] data in rows) {
+							foreach (object dataItem in data) {
```
vargaz: This is supposed to be called at runtime. Not sure whenever it matters.

EgorBo: yeah, this PR adds support only for simple cases (primitive literals, etc).

Issue #9664, opened 2018-07-20
DateTime.Now or Timezone information do not transition out of DST for all time zones


jbatt33:
### Note
Using Vatican time zone as an example. 

## Steps to Reproduce

1. Turn off NTP time synchronization for linux
> sudo timedatectl set-ntp off
2. Turn off any other network time synchronization services that may be active
3. Set the time zone on the server to the Vatican time zone
> sudo timedatectl set-timezone Europe/Vatican
4. Set the time of the OS to 1 hour and 1 minute before the reported DST Exit datetime of 10/28/2018 3:00:00 AM 
> sudo date 102801592018.00 
5. Run a mono application that prints the time.
6. Wait for an 1 hour and compare the application's time to the OS's time

### Current Behavior

The mono application will not transition out of DST for two hours past DST exit. 

#### OS Linux Run Output
```
$ date
Sun Oct 28 02:59:55 CEST 2018

$ date
Sun Oct 28 02:59:57 CEST 2018

$ date
Sun Oct 28 02:00:00 CET 2018 << System Time correctly drops out of DST

$ date
Sun Oct 28 02:00:02 CET 2018 
```
#### Before transition
```
-----------------------------------------------
DateTime.Now: 10/28/2018 2:59:58 AM
Cleared: DateTime.Now: 10/28/2018 2:59:58 AM
UTC+Offset: 10/28/2018 2:59:58 AM
636762923986794550 - Start: 3/25/2018 2:00:00 AM:636575400000000000 - End: 10/28/2018 3:00:00 AM:636762924000000000 - True
In DST
UTC: 10/28/2018 00:59:58 - Offset: 02:00:00
-----------------------------------------------
```

#### After transition
```
-----------------------------------------------
DateTime.Now: 10/28/2018 3:00:03 AM
Cleared: DateTime.Now: 10/28/2018 3:00:03 AM
UTC+Offset: 10/28/2018 3:00:03 AM
636762924036831780 - Start: 3/25/2018 2:00:00 AM:636575400000000000 - End: 10/28/2018 3:00:00 AM:636762924000000000 - True
Not in DST
UTC: 10/28/2018 01:00:03 - Offset: 02:00:00
-----------------------------------------------
```

#### The UTC+Offset calculation will be correct, but the DateTime.Now will still return the incorrect information. 
```
-----------------------------------------------
DateTime.Now: 10/28/2018 4:14:01 AM
Cleared: DateTime.Now: 10/28/2018 4:14:01 AM
[...]

PR #13553, opened 2019-03-20
Fix transition offset logic

MaximLipnin:
I'm not sure about the correctness of this workaround but it works on macOS for the repro from https://github.com/mono/mono/issues/9664#issuecomment-472260968

Transition bounds used by time zone info represent some local date-time but actually they are of Unspecified date-time kind. 
When we try to get transition offset we compare some current date-time converted to UTC with appropriate transition bounds.
The fix is to use UTC for both current date-time and transition bounds during comparing.

Fixes https://github.com/mono/mono/issues/9664


MaximLipnin:
@monojenkins build failed

MaximLipnin:
@monojenkins build failed

MaximLipnin:
Should it be closed due to merging the fix to master and 2019-02?

marek-safar:
@monojenkins backport 2018-10

Issue #13533, opened 2019-03-19
bcl.sln fails on running genconsts on Windows


blattersturm:
Ever since dfa2fe6d2b0988bcf66c707dd41c809def4bace0 it seems that winsetup.bat no longer has the output genconsts expects, leading to building of bcl.sln by itself to fail.

lateralusX:
Successfully build bcl.sln using https://github.com/mono/mono/pull/13554.

PR #13554, opened 2019-03-20
Adjust genconst to changes in winsetup.bat.

lateralusX:
When logic moved from winsetup.bat into an msbuild task that changed the console output of winsetup.bat causing problems for genconst to find MONO_VERSION and MONO_CORLIB_VERSION. There was also a potential problem where winsetup.bat couldn't find msbuild.exe if not included in PATH.

This fix makes sure winsetup.bat detect and setup a full msbuild environment before running msbuild.exe. It also change genconsts to look at the generated config.h for the correct values of MONO_VERSION and MONO_CORLIB_VERSION that will be used by the runtime builds.

Should fix https://github.com/mono/mono/issues/13533.

### DIFF
--- a/msvc/scripts/genconsts.cs
+++ b/msvc/scripts/genconsts.cs
@@ -50,7 +50,17 @@ public static class Program {
                 Console.Write (output);
                 return winsetupProcess.ExitCode;
             } else {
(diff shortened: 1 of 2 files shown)


lateralusX:
@monojenkins rebuild failed

PR #13559, opened 2019-03-20
[netcore] better MemberData support

EgorBo:
```
RUN: 21756 (+13000)
FAILURES: 90
SKIPPED: 200
```
Output for System.Runtime.Tests: https://gist.github.com/EgorBo/b4f94cea3df38f74b9315ad2ad576d56

For MemberData, gen-xunit-proj now generates:

```csharp
foreach (var row in SomeType.MemberDataMethod())
    typeof(SomeType).GetMethod("TestMethod").Invoke(o, (object[])row);
```


vargaz:
One of the reasons the codegen was changed to roslyn was that CodeDom couldn't generate foreach statements.

PR #13566, opened 2019-03-21
[netcore] Implement missing bits in ThreadPool

EgorBo:
Now I am able to run xunit tests directly.
I mostly copied missing bits from old threadpool.
(In theory we need to use ThreadPool.Portable.cs, am I right?)

Managed to run 29808 xunit tests directly without the gen-xunit-proj in System.Runtime.Tests,
it crashes on `PointerFieldGetValue` which I guess should be skipped on mono (see https://github.com/mono/corefx/blob/master/src/System.Runtime/tests/System/Reflection/PointerTests.cs#L128)

I think we need to upstream all mono-runtime specific [Skip]s from mono/corefx to dotnet/corefx

### DIFF
--- a/mcs/class/System.Private.CoreLib/System.Threading/ThreadPool.cs
+++ b/mcs/class/System.Private.CoreLib/System.Threading/ThreadPool.cs
@@ -4,7 +4,6 @@ namespace System.Threading
 {
 	partial class ThreadPool
 	{
-		[System.Security.SecurityCritical]
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		private static extern void InitializeVMTp(ref bool enableWorkerTracking);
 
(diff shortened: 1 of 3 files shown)


mono/metadata/domain.c
```diff
@@ -786,7 +786,6 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
 	mono_defaults.generic_ienumerator_class = mono_class_load_from_name (
 	        mono_defaults.corlib, "System.Collections.Generic", "IEnumerator`1");
 
-#ifndef ENABLE_NETCORE
 	ERROR_DECL (error);
```
EgorBo: had to enable it because runtime crashed somewhere trying to access unitialized threadpool_perform_wait_callback_method (and also added `_ThreadPoolWaitCallback` for that)

filipnavara:
> In theory we need to use ThreadPool.Portable.cs, am I right?

That's the plan. I started working on it, but there was more dependencies than I realized at first. I have most of them sorted out one way or another, but I couldn't decide which code should go to Mono, which code should go to System.Native in CoreRT and so on...

If it's a blocker or something you would like to work on I can share all code and perhaps you can take it over. I closed most of the PRs, but I had a version of the code that compiled on Mono.

mcs/class/System.Private.CoreLib/System.Threading/ThreadPool.cs
```diff
@@ -1,47 +1,138 @@
+using System.Runtime.CompilerServices;
+
 namespace System.Threading
 {
 	partial class ThreadPool
 	{
+		[System.Security.SecurityCritical]
```
marek-safar: Remove all SecurityCritical

PR #13568, opened 2019-03-21
[netcore] Add a python script to download the prebuilt corefx test assemblies.

vargaz:


vargaz:
@monojenkins build failed

PR #13571, opened 2019-03-21
Add UTC timezone in WinRT GetSystemTimeZones fallback implementation.

lateralusX:
Needed by tests when hitting fallback solution in TimeZoneInfo.WinRT.cs. Fix also sort the TimeZones similar to how its done in .NET Framework/Core.

### DIFF
--- a/mcs/class/corlib/System/TimeZoneInfo.WinRT.cs
+++ b/mcs/class/corlib/System/TimeZoneInfo.WinRT.cs
@@ -338,8 +338,17 @@ namespace System
 				// EnumDynamicTimeZoneInformation() might not be available.
 			}
 
-			if (result.Count == 0)
+			if (result.Count == 0) {
 				result.Add (Local);
+				result.Add (Utc);
+			}
+
+			result.Sort ((x, y) =>
+			{
+				int comparison = x.BaseUtcOffset.CompareTo(y.BaseUtcOffset);
+				return comparison == 0 ? string.CompareOrdinal(x.DisplayName, y.DisplayName) : comparison;
+			});
+
 			return result;
 		}
 	}


lateralusX:
@monojenkins rebuild failed

mcs/class/corlib/System/TimeZoneInfo.WinRT.cs
```diff
@@ -338,8 +338,17 @@ internal static List<TimeZoneInfo> GetSystemTimeZonesWinRTFallback ()
 				// EnumDynamicTimeZoneInformation() might not be available.
 			}
 
-			if (result.Count == 0)
+			if (result.Count == 0) {
 				result.Add (Local);
+				result.Add (Utc);
```
jaykrell: Should one or both of these always be added? Or is it best to have fewer/one and leave it alone if not zero?

lateralusX: Yes in case there are not time zones on the system (cases where EnumDynamicTimeZoneInformation is not available) we should add them to the result. This logic was already in place but it didn't add UTC, so  there are test failing due to not having a UTC time zone returned from GetSystemTimeZones.

jaykrell: My question is -- should the `if` be removed or modified?
Should Local and Utc always be present?
Or should Utc always be present?
Or does the dynamic case always add them, so if not zero, they are present anyway?

lateralusX: Yes, the dynamic case should already add them to the list (all test pass on system that supports that scenario), so this only handle the special case where no time zones have been added.

PR #13572, opened 2019-03-21
Increase spin-count on test to not end up in frequent sleeps on multicore machines.

lateralusX:
When running this test, ManualResetSlimTest on Windows on a machine with free cores during the test run, the SpinOnce's default value will be to low forcing the thread into frequent sleep(1) scenario, causing long execution times. This PR set count to 100 for the test (default 20) before SpinOnces will hit sleep(1. This will run the test much quicker on machines with idle cores.


lateralusX:
@monojenkins rebuild failed

PR #13579, opened 2019-03-21
Remove the parsing, hashing, locking, allocations associated with JIT icall signatures.

jaykrell:
Remove the parsing, hashing, locking, allocations associated with JIT icall signatures.
 Make them statically allocated, single-instanced at compile time, and almost
 statically initialized.

Some signatures are redundant.


jaykrell:
The bockbuild thing is an accident, let me try to fix. I don't know how to work with submodules. :(

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
And after this the hash table holding them can go away too, but one thing at a time.

jaykrell:
@lambdageek

mono/metadata/icall-signatures.h
```diff
@@ -0,0 +1,279 @@
...
+// mono_icall_sig_int_ptr_int_int_ptr_object
+// mono_icall_sig_object_ptr_int_int_int_int
+// mono_icall_sig_object_ptr_ptr_ptr_ptr_ptr
+// mono_icall_sig_ptr_ptr_int32_ptr_ptr_ptrref
+// mono_icall_sig_void_ptr_ptr_int32_ptr_ptrref
+// mono_icall_sig_int32_ptr_ptr_ptr_ptr_ptr_int32
+// mono_icall_sig_void_ptr_ptr_int32_ptr_ptrref_ptr_ptrref
+
+#define ICALL_SIGS				\
+ICALL_SIG (1, (object))				\
```
jaykrell: Depending on desired sort order, put count at the end.
These are sorted, and the expanded list is sorted "the same", but lacks the numbers so it looks messy.
The expanded list is just a comment, so they need not be sorted the same.

vargaz:
LGTM. The encoding should be fixed to use something other than a pointer into mono_defaults.

jaykrell:
Suggestions?
The easiest most obvious one, is a size_t that offsetof MonoDefaults.
That will be same *slightly* bloated size, but avoid the relocations and virtually identical logic.
Sufficient?

Emphasis on the only slight bloat -- it is per signature (parameter), not per icall.
And the number of signatures can be brought down later but e.g. making obj and object the same.

There is a way to get it down to a byte per character but I struggled more than a minute.
It'd be offsetof / sizeof(gpointer), and then padded with chars for alignment.

Other than not crashing and passing tests, maybe I should be able to see the correct signatures flowing somewhere?

Does anyone have the patience to double check all the signatures in the diff? It was mostly automated and I haven't yet had the time/patience to check each of the 200 lines. But I understand if that is insisted upon.

jaykrell:
One can also imagine an enum but it doesn't really help.
Unless there is some traffic in a stable form outside the runtime, which there is not.
The offset of instead of pointer is mild goodness. Trade roughly one instruction for a quite few hidden relocations. And leave open the possibility of mono_defaults changing from a value to a pointer if there is ever a need for that.

I should probably also assert that the results are all not null, i.e. this is late enough in initialization.

jaykrell:
I changed to offsetof(MonoDefaults).
It morally the same but avoids the relocations.
I made it a *little* more commented and verbose but essentially the same except for the pointer vs. offsetof, and still a bit obscure.

jaykrell:
Oh the subltey. The initializer storage double as the signature storage. So pointer-sized is all but required, and the problem with smaller wasn't alignment so much as initializing overwriting not-yet-initialized.
So `size_t` is a little risky and `gsize` or `uintptr_t` instead. And using a byte or enum is ok, if small stable values are desirable, but it should be stored in a `gsize` or `uintptr_t.

jaykrell:
@monojenkins squash

jaykrell:
@vargaz please specify what you want for the initializer.
It all-but needs to be pointer-sized.
There is presently no trafficking outside the runtime, so no current stability requirement.
An enum could be introduced.
The duplicity of e.g. `obj` vs. `object` bugs me but that can be fixed later.

jaykrell:
@monojenkins build failed

mono/metadata/icall.c
```diff
@@ -8733,111 +8734,113 @@ mono_lookup_icall_symbol (MonoMethod *m)
...
+ICALL_SIGS
+};
 
-	g_strfreev (parts);
+#undef ICALL_SIG
+#define ICALL_SIG(n, types) MonoMethodSignature * const ICALL_SIG_NAME (n, types) = &mono_icall_signatures.ICALL_SIG_NAME (n, types).sig;
+ICALL_SIGS
+#undef ICALL_SIG
 
-	mono_image_lock (corlib);
```
jaykrell: I wonder if the signatures should be further hashed/single-instanced based on their eventual non-string representation.
To some extent that can be done statically, `obj` vs. `object`.
But they historically were not in either case -- just based on input string -- so this is progress.

jaykrell:
@monojenkins build failed

vargaz:
I'd prefer an enum or something.

jaykrell:
It doesn’t really buy anything but if you insist.

The offetof is kinda self-describing. No switch and new ones more trivial to add.

Except the ptrref special case would be less special. Mono imho has too many switches and too little self-description (like just call a function pointer and don’t have to visit every use of the enum).

The downside is if there is a need for a stable persistent form. Enums do shine there.

jaykrell:
(I’ll take silence as insistence.)

jaykrell:
enum kinda is there. There are many cosmetically similar ways to do this.
Perhaps just one function like before, with the if/else ladder or switch/break.
The way I have it avoids the break per case, minor.
I still like the offsetof based approach.
enum is also more flexible too, like if the data came from various places.
There are 16 values and a max of 8, so we could also pack this into an int instead of array of intptr_t, but we need to pay for that space anyway.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/icall.c
```diff
@@ -8733,111 +8734,129 @@ mono_lookup_icall_symbol (MonoMethod *m)
...
+		mono_class_get_byref_type (mono_defaults.int_class), // ICALL_SIG_TYPE_ptrref
+		m_class_get_byval_arg (mono_defaults.string_class),	 // ICALL_SIG_TYPE_string
+		m_class_get_byval_arg (mono_defaults.uint16_class),	 // ICALL_SIG_TYPE_uint16
+		m_class_get_byval_arg (mono_defaults.uint32_class),	 // ICALL_SIG_TYPE_uint32
+		m_class_get_byval_arg (mono_defaults.byte_class),	 // ICALL_SIG_TYPE_uint8
+		m_class_get_byval_arg (mono_defaults.uint64_class),	 // ICALL_SIG_TYPE_ulong
+		m_class_get_byval_arg (mono_defaults.void_class),	 // ICALL_SIG_TYPE_void
+	};
+
+	MonoMethodSignature_a *sig = (MonoMethodSignature*)&mono_icall_signatures;
```
jaykrell: Does anyone understand the strict aliasing rules?
Is there a good/better way to write this?
Other than splitting up the storage?

jaykrell: I think the simplest fix would be:

`*(gsize*)(i ? &sig->params [i - 1] : &sig->ret) = (gsize)lookup [index];`

jaykrell:
@monojenkins build failed

mono/metadata/icall.c
```diff
@@ -8733,111 +8734,132 @@ mono_lookup_icall_symbol (MonoMethod *m)
...
+	while ((n = sig->param_count)) {
+		--sig->param_count; // remove ret
+		gsize_a *types = (gsize*)(sig + 1);
+		for (int i = 0; i < n; ++i) {
+			gsize index = *types++;
+			g_assert (index < G_N_ELEMENTS (lookup));
+			// Casts on next line are attempt to follow strict aliasing rules,
+			// to ensure reading from *types precedes writing
+			// to params [].
+			*(gsize*)(i ? &sig->params [i - 1] : &sig->ret) = (gsize)lookup [index];
```
jaykrell: The cast should order the reads/writes in this function,
but still leaves a hypothetical hole of other reads
via other types outside the function. Odds
are very low that it is a real problem though.

PR #13585, opened 2019-03-21
Improve exception report for native crashes on x64

filipnavara:
When running the `netcore` tests I accidentally copied the `System.Globalization.Native` library instead of making a symlink. This mistake lead to a native exception. Unfortunately the exception handling code didn't print the managed stack trace where the error was happening because the original exception context was not passed through the whole code chain. This fixes it to print nice managed stack trace and information from the context as well.


mono/mini/exceptions-amd64.c
```diff
@@ -887,7 +887,7 @@ mono_arch_handle_altstack_exception (void *sigctx, MONO_SIG_HANDLER_INFO_TYPE *s
 	MonoException *exc = NULL;
 	gpointer *sp;
 	int frame_size;
-	MonoContext *copied_ctx;
+	ucontext_t *copied_ctx;
```
lewurm: unfortunately, `ucontext_t` is not portable 😕 That's the whole reason why we do the whole exercise with `MonoContext`.

mono/mini/exceptions-amd64.c
```diff
@@ -900,15 +900,15 @@ mono_arch_handle_altstack_exception (void *sigctx, MONO_SIG_HANDLER_INFO_TYPE *s
...
 	frame_size += 15;
 	frame_size &= ~15;
 	sp = (gpointer *)(UCONTEXT_REG_RSP (sigctx) & ~15);
 	sp = (gpointer *)((char*)sp - frame_size);
-	copied_ctx = (MonoContext*)(sp + 4);
+	copied_ctx = (ucontext_t*)(sp + 4);
 	/* the arguments must be aligned */
 	sp [-1] = (gpointer)UCONTEXT_REG_RIP (sigctx);
-	mono_sigctx_to_monoctx (sigctx, copied_ctx);
+	memcpy (sigctx, copied_ctx, sizeof (ucontext_t));
```
lewurm: it's swapped, it should be `memcpy (copied_ctx, sigctx, sizeof (...))`

filipnavara: Of course, messed it up just before committing...

mono/mini/exceptions-amd64.c
filipnavara: I followed what was already done in the -ppc version of the code under the assumption that it's ok to do if it's limited to the platform specific file.

It was the least intrusive change. I do agree that it's probably better to propagate MonoContext, but I wanted to keep the change small.

filipnavara:

lewurm:
thanks!

PR #13587, opened 2019-03-21
Make test runner able to deal with whitespace patterns other than newline

directhex:
Helix wasn't running our runtime tests as expected, because the test runner expected newline-separated tests to run, and we were generating space-separated.

Following "be liberal in what you accept" philosophy, and to prevent this issue in the future, just deal with "whitespace separated" rather than "this specific whitespace and no other"

### DIFF
--- a/mono/tests/test-runner.cs
+++ b/mono/tests/test-runner.cs
@@ -205,8 +205,12 @@ public class TestRunner
 
 		if (!String.IsNullOrEmpty (inputFile)) {
 			foreach (string l in File.ReadAllLines (inputFile)) {
-				for (int r = 0; r < repeat; ++r)
-					tests.Add (l);
+				foreach (string m in l.Split (' ')) {
+					if (!String.IsNullOrEmpty (m)) {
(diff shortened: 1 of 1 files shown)


mono/tests/test-runner.cs
```diff
@@ -205,8 +205,12 @@ class TestInfo {
 
 		if (!String.IsNullOrEmpty (inputFile)) {
 			foreach (string l in File.ReadAllLines (inputFile)) {
-				for (int r = 0; r < repeat; ++r)
-					tests.Add (l);
+				foreach (string m in l.Split (' ')) {
+					if (!String.IsNullOrEmpty (m)) {
```
filipnavara: StringSplitOptions.RemoveEmptyEntries?

akoeplinger: I agree, but it's also not worth spinning another build as this is test code so we can be a little more lenient :)

akoeplinger:
I verified it's showing up correctly on Helix now, thank you!

akoeplinger:
@monojenkins backport 2019-02

PR #13591, opened 2019-03-21
[netcore] Implement missing bits in Mutex

EgorBo:
Needed for xunit console runner tool.

### DIFF
--- a/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
+++ b/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
@@ -169,7 +169,6 @@
       <Compile Include="System.Security\DynamicSecurityMethodAttribute.cs" />
       <Compile Include="System.Threading\EventWaitHandle.cs" />
       <Compile Include="System.Threading\Interlocked.cs" />
-      <Compile Include="..\corlib\System.IO\MonoIOError.cs" />
       <Compile Include="System.Threading\Mutex.cs" />
       <Compile Include="System.Threading\Overlapped.cs" />
       <Compile Include="System.Threading\PreAllocatedOverlapped.cs" />
@@ -215,6 +214,7 @@
 
       <Compile Include="..\corlib\System.Runtime.CompilerServices\PreserveDependencyAttribute.cs" />
 
+      <Compile Include="..\corlib\System.IO\MonoIOError.cs" />
       <Compile Include="..\corlib\System.Threading\Monitor.cs" />
       <Compile Include="..\corlib\System.Threading\RegisteredWaitHandle.cs" />
 
--- a/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
+++ b/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
@@ -169,6 +169,7 @@
       <Compile Include="System.Security\DynamicSecurityMethodAttribute.cs" />
       <Compile Include="System.Threading\EventWaitHandle.cs" />
       <Compile Include="System.Threading\Interlocked.cs" />
+      <Compile Include="..\corlib\System.IO\MonoIOError.cs" />
       <Compile Include="System.Threading\Mutex.cs" />
       <Compile Include="System.Threading\Overlapped.cs" />
       <Compile Include="System.Threading\PreAllocatedOverlapped.cs" />
--- a/mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
+++ b/mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
@@ -1,14 +1,80 @@
+using System.Runtime.CompilerServices;
+using System.IO;
+
 namespace System.Threading
 {
 	partial class Mutex
 	{
-		void CreateMutexCore (bool initiallyOwned, string name, out bool createdNew) => throw new NotImplementedException ();
-
-		static OpenExistingResult OpenExistingWorker (string name, out Mutex result) => throw new NotImplementedException ();
+		Mutex (IntPtr handle) => Handle = handle;
 
 		public void ReleaseMutex ()
 		{
-			throw new NotImplementedException ();
+			if (!ReleaseMutex_internal (Handle))
+				throw new ApplicationException ("Mutex is not owned");
 		}
+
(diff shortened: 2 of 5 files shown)


netcore/gen-xunit-runner/Program.cs
```diff
@@ -379,7 +379,7 @@ static int Main (string[] args)
 
 				var block = ParseText<BlockSyntax> (case_template
 					.Replace ("#MSG#", msg)
-					.Replace ("#NRUN#", test.MemberDataMethod == null ? "1" : ((IEnumerable<object []>) test.MemberDataMethod.Invoke (null, null)).Count ().ToString ()));
+					.Replace ("#NRUN#", test.MemberDataMethod == null ? "1" : ((IEnumerable) test.MemberDataMethod.Invoke (null, null)).Cast<object> ().Count ().ToString ()));
```
EgorBo: Fixes `Unhandled Exception: System.InvalidCastException: Unable to cast object of type '<CopyTo_Hidden_Data>d__19' to type 'System.Collections.Generic.IEnumerable`1[System.Object[]]'.`
when it's used for System.Collections.Tests

mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -169,6 +169,7 @@
       <Compile Include="System.Security\DynamicSecurityMethodAttribute.cs" />
       <Compile Include="System.Threading\EventWaitHandle.cs" />
       <Compile Include="System.Threading\Interlocked.cs" />
+      <Compile Include="..\corlib\System.IO\MonoIOError.cs" />
       <Compile Include="System.Threading\Mutex.cs" />
```
vargaz: There is a ..\corlib section in the csproj, it should go there.

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
```diff
@@ -1,14 +1,80 @@
...
+			fixed (char *fixed_name = name)
+				return CreateMutex_icall (initiallyOwned, fixed_name,
+					name?.Length ?? 0, out created);
+		}
+
+		static OpenExistingResult OpenExistingWorker (string name, out Mutex result) 
+		{
+			if (name == null)
+				throw new ArgumentNullException ("name");
+
```
vargaz: Use nameof (name)

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
```diff
@@ -1,14 +1,80 @@
...
-		void CreateMutexCore (bool initiallyOwned, string name, out bool createdNew) => throw new NotImplementedException ();
-
-		static OpenExistingResult OpenExistingWorker (string name, out Mutex result) => throw new NotImplementedException ();
+		Mutex (IntPtr handle) => Handle = handle;
 
 		public void ReleaseMutex ()
 		{
-			throw new NotImplementedException ();
+			if (!ReleaseMutex_internal (Handle))
+				throw new ApplicationException ("Mutex is not owned");
```
marek-safar: This is an unlikely right exception, what does core throw?

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
```diff
@@ -1,14 +1,80 @@
...
+				throw new ArgumentNullException (nameof (name));
+
+			result = null;
+			if ((name.Length == 0) ||
+				(name.Length > 260)) {
+				return OpenExistingResult.NameInvalid;
+			}
+			
+			MonoIOError error;
+			IntPtr handle = OpenMutex_internal (name, MutexRights.Modify, out error);
```
marek-safar: use `out MonoIOError error` instead

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
```diff
@@ -1,14 +1,80 @@
...
+			if (handle == (IntPtr) null) {
```
marek-safar: prefer `handle == IntPtr.Zero`

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
```diff
@@ -1,14 +1,80 @@
...
+
+		[MethodImplAttribute(MethodImplOptions.InternalCall)]
+		private unsafe static extern IntPtr OpenMutex_icall (char *name, int name_length, MutexRights rights, out MonoIOError error);
+
+		[MethodImplAttribute(MethodImplOptions.InternalCall)]
+		private static extern bool ReleaseMutex_internal (IntPtr handle);
+	}
+
+	[Flags]
+	public enum MutexRights
```
marek-safar: Change the icall not to require the value

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
vargaz: These are probably the same icalls used by the normal version.

mcs/class/System.Private.CoreLib/System.Threading/Mutex.cs
filipnavara: Surprisingly enough that's what CoreRT throws on Unix: `throw new ApplicationException(SR.Arg_SynchronizationLockException);`

filipnavara: ...and CoreCLR as well: https://github.com/dotnet/coreclr/blob/cf85f56e473b6cdc4a68e9758f57dc571e14c9cb/src/System.Private.CoreLib/shared/System/Threading/Mutex.Windows.cs#L95

marek-safar: @stephentoub could you please confirm this^ is intentional

stephentoub: It's intentional, unfortunately.  It's what netfx has used forever:
https://referencesource.microsoft.com/#mscorlib/system/threading/mutex.cs,425
and the cost/benefit to changing it at this point is not favorable.

PR #13593, opened 2019-03-21
[netcore] Download test assets from the dotnet-core blob feed.

vargaz:
Use them in gen-xunit-runner. Note that COREFX_ROOT is still required since the test assemblies reference assemblies from
the corefx sdk dir.

### DIFF
--- a/netcore/Makefile
+++ b/netcore/Makefile
@@ -94,3 +94,15 @@ xtest-%: prepare check-env
 		$(shell grep -v '^#\|^$$' excludes-$*.rsp) $(FIXTURE)
 
 xtest-all: $(addprefix xtest-, $(TEST_SUITES))
+
+FEED_BASE_URL = https://dotnetfeed.blob.core.windows.net/dotnet-core
+TEST_ASSETS_URL = https://dotnetfeed.blob.core.windows.net/dotnet-core/corefx-tests/4.6.0-preview4.19156.10/OSX.x64/netcoreapp/corefx-test-assets.xml
+
+corefx-test-assets.xml:
+	wget $(TEST_ASSETS_URL)
+
+dl-test-assets: corefx-test-assets.xml .stamp-dl-test-assets
+
+.stamp-dl-test-assets:
+	python dl-test-assets.py corefx-test-assets.xml $(FEED_BASE_URL) assets
+	touch $@
--- a/netcore/dl-test-assets.py
+++ b/netcore/dl-test-assets.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+
+import sys
+import subprocess
+import xml.etree.ElementTree as ET
+
+if len(sys.argv) < 4:
+    print("Usage: dl-test-assets.py <path to assets.xml> <base url> <output dir>")
+    sys.exit(1)
+
+infile_name = sys.argv [1]
+base_url = sys.argv [2]
+outdir = sys.argv [3];
+tree = ET.parse(infile_name)
+root = tree.getroot()
+
+for elem in root:
+    if elem.tag != "Blob":
+        continue
+    print elem.attrib ["Id"]
+    res = subprocess.call (["wget", "-N", "-P", outdir, base_url + "/" + elem.attrib ["Id"]])
+    if res != 0:
+        print ("Download failed.")
+        sys.exit (1)
+


netcore/Makefile
```diff
@@ -74,6 +72,17 @@ all: $(addprefix build-, $(TEST_SUITES))
 
 check: $(addprefix run-, $(TEST_SUITES))
 
-clean:
-	rm -rf sdk shared host dotnet tests LICENSE.txt ThirdPartyNotices.txt $(NETCORESDK_FILE)
+FEED_BASE_URL = https://dotnetfeed.blob.core.windows.net/dotnet-core
+TEST_ASSETS_URL = https://dotnetfeed.blob.core.windows.net/dotnet-core/corefx-tests/4.6.0-preview4.19156.10/OSX.x64/netcoreapp/corefx-test-assets.xml
```
filipnavara: s/4.6.0-preview4.19156.10/$(VERSION)/ ?

filipnavara: Oh, nvm, it uses different prefix for the version. It's hard to keep track of it.

vargaz: No idea what this version is and how it relates to netcore version.

filipnavara: I actually do have an idea, but this comment space is not big enough to write the description of the Maestro (buildtools) / Maestro++ (arcade) dependency management and how the SDK picks the correct assets. It's quite confusing and I keep bumping into it.

filipnavara: https://github.com/dotnet/corefx/blob/b5ac532676652673c4e72fcc99e734309a034f56/eng/Versions.props#L6-L10

filipnavara: Looks like there's something strange I don't get. Apparently the current assets are still 4.6.0 while I would expect them to be 4.7.0. Unfortunately the builds are now published only on the feeds and MyGet is no longer updated [properly] so it's difficult to figure out what is going on. I guess it's fine to hard code the version for now and figure this out later.

vargaz:
Note that there are some test failures with System.Runtime.Tests with this version.

netcore/Makefile
marek-safar: @ViktorHofer could you suggest a way to keep them in sync manually?

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

netcore/Makefile
ViktorHofer: Sorry I'm currently OOF till Monday next week. As the PR is merged it seems you were successful?

vargaz: We are downloading some version of the test assets, just not sure how to pick the correct version.

ViktorHofer: Either with a DARC subscription to corefx or with mirror-updates from coreclr. cc @jkotas

jkotas: I think you should use DARC subscription.

You want to be trailing corefx master, for both the CoreFX binaries and CoreFX tests. DARC subscription will auto-submit PR whenever there is a new good official CoreFX build to pick up.

>  Apparently the current assets are still 4.6.0 while I would expect them to be 4.7.0.

There are way too many different version numbers in CoreFX that you need a secret decoder ring to understand. The build version number is still 4.6.X for these.

ViktorHofer: A Darc subscription is definitely preferable but I'm not sure how much this would require this repository to be arcade enabled. @wtgodbe @safern @cchosta should know.

> a secret decoder ring to understand

We tried to align versions but we still have some differences. We should do a better job here in the future. cc @ericstj

safern: > A Darc subscription is definitely preferable but I'm not sure how much this would require this repository to be arcade enabled. @wtgodbe @safern @cchosta should know.

In order to setup a Darc subscription should be pretty simple. We just need to add a`eng` folder if it doesn't exist and throw a couple of files there, `Version.Details.xml` and `Versions.props` -- for example here are the ones in coreclr who uses this channel: 

https://github.com/dotnet/coreclr/blob/master/eng/Version.Details.xml#L14
https://github.com/dotnet/coreclr/blob/master/eng/Versions.props#L11

Setting up the subscription in maestro is fairly simple, just a matter of adding it to the database -- @jcagme can give more details on that.

ericstj: We have a ton of constraints on assembly versioning due to netstandard, oobing, desktop support, etc, etc.  It's difficult to wiggle much in this space but we could make an effort to rebase all the assembly versions for netcoreapp.

The specific 4.6 / 4.7 discrepancy is due to the switch to arcade mid cycle which reset the file-version algorithm.  At the time folks chose to bump the file-versions to keep them monotonically increasing while keeping the packages versioned at 4.6.  We could choose to fix that now.  The major.minor of the packages is arbitrary and meant to be 1:1 with a release wave.

ViktorHofer: @filipnavara is the version discrepancy currently causing pain? If yes could you please describe the scenarios.

@vargaz we need to create the subscription as shortly explained by @safern above and prepare the repo for it. I think it makes sense to create an issue for it where we can discuss this further.

filipnavara: >  is the version discrepancy currently causing pain? If yes could you please describe the scenarios.

Not in technical way. It is just confusing when trying to figure out what version should I look for and when manually verifying that my setup is correct (happens too often now without the DARC subscription).

PR #13581, opened 2019-03-21
[aot] Avoid an assert which can happen in normal usage.

vargaz:
Fixes https://github.com/mono/mono/issues/13455.

PR #13605, opened 2019-03-22
[aot] cleanup MonoError in add_extra_method_with_depth

lambdageek:
I missed something in the code review for #13581.  Need to cleanup MonoError if it's not ok.


lambdageek:
@akoeplinger @vargaz sorry... missed this in the code review.  (need to also backport to 2019-02 since #13604 has the same missing cleanup.

akoeplinger:
I pushed the fix to the 2019-02 PR.

PR #13606, opened 2019-03-22
[wasm] Include the corebindings.c in the package output

jeromelaban:
This update restores the ability to use the packager outside of the mono source tree.


kjpou1:
@monojenkins build failed

Issue #13597, opened 2019-03-22
Runtime Error when using span on LeetCode


imba-tjd:
## Steps to Reproduce

1. Open <https://leetcode.com/problems/longest-palindromic-substring/>.
2. Copy and Paste my following code.
3. Click <kbd>Run Code</kbd>

I don't think my code has an algorithm issue because changing span to string works.

```c#
public class Solution {
    public string LongestPalindrome(string s) // DP
    {
        ReadOnlySpan<char> ss = s.AsSpan();
        ReadOnlySpan<char> str = "";
        for (int i = 0; i < s.Length; i++)
        {
            var r = Expand(ss, i - 1, i + 1);
            if (r.Length > str.Length)
                str = r;
            r = Expand(ss, i, i + 1);
            if (r.Length > str.Length)
                str = r;
        }
        return str.ToString();
    }
    internal ReadOnlySpan<char> Expand(ReadOnlySpan<char> s, int from, int to)
    {
        while (from >= 0 && to <= s.Length - 1)
        {
            if (s[from] != s[to])
                break;
            from--;
            to++;
        }

        return from + 1 == to ? "" : s.Slice(from + 1, to - from - 1);
    }
}
```

### Current Behavior

Runtime Error

### Expected Behavior

Pass

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

## Stacktrace

![图片](https://user-images.githubusercontent.com/24759802/54805094-d26fad00-4cb0-11e9-9730-d6c321b04126.png)

The stacktrace text on LeetCode seems can't be copied. If you want to see it, you must do it yourself.

---

In case this is a LeetCode issue, I tried my code in Ubuntu vm. But this is my first time using mono, I encountered many errors like "couldn't found System.Memory" or "Could not load file or assembly System.Runtime.Loader".

**So if you can compile and xunit test pass my following code, just close this issue.**

<details>
<summary>Full Code</summary>

```c#
using System;
using Xunit;

namespace LongestPalindromicSubstring
{
    public interface ISolution { string LongestPalindrome(string s); }
    class Solution : ISolution
    {
        public string LongestPalindrome(string s) // BF
        {
            if (s.Length == 1)
                return s;
            var result = string.Empty.AsSpan();
            var span = s.AsSpan();
            for (int i = 0; i < s.Length - 1; i++)
                for (int j = i + 1; j <= s.Length; j++)
                {
                    var ss = span.Slice(i, j - i);
[...]

Issue #13452, opened 2019-03-14
[Linux] Mono (5.18.0.268) crashes in String:IndexOf(char value, int startIndex)


VirusFree:
## Steps to Reproduce
1.  Run sample app (provided in attachment) -> mono ConsoleApp1.exe

### Current Behavior
mono crashes and produces a mono_crash.json file (included in attachment)

### Expected Behavior
not crash mono

## On which platforms did you notice this

[  ] macOS
[X] Linux
[  ] Windows

**Version Used**:
Mono JIT compiler version 5.18.0.268 (tarball Tue Mar 12 14:15:15 UTC 2019)

## Stacktrace
          at System.SpanHelpers:IndexOf <0x000fc>
          at System.String:IndexOf <0x00072>
          at System.String:IndexOf <0x0003a>
          at ConsoleApp1.Program:DoTheThing <0x00052>
          at ConsoleApp1.Program:Main <0x0002a>
          at <Module>:runtime_invoke_void_object <0x000d8>
      Aborted (core dumped)

### Attachments
[BugReport.zip](https://github.com/mono/mono/files/2965702/BugReport.zip)

VirusFree:
It took some time, but I was able to narrow down the problem using a small and simple sample code.
The source and binary are also in the zip file, but basically this is it :

        static void Main(string[] args)
        {
            var str = "this is a string";
            Console.WriteLine(DoTheThing(str, 's'));
        }
        public static string DoTheThing(string source, char c, int n = 1, int StartIndex = 0)
        {
            int idx = StartIndex - 1;
            while (n != 0)
            {
                idx = source.IndexOf(c, idx + 1);
                --n;
            }
            return null;
        }

lewurm:
(1) What CPU are you running on? 
(2) Could you post the full crash log?
(3) Did you install mono from our Debian reprository?
(4) Could you try to run your sample with `MONO_ENV_OPTIONS=-O=-aot`?

VirusFree:
```
$ cat /etc/apt/sources.list.d/mono-official-stable.list
deb https://download.mono-project.com/repo/ubuntu stable-bionic main
```

## CPU info
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : Intel(R) Xeon(R) CPU E5-2676 v3 @ 2.40GHz
stepping        : 2
microcode       : 0x41
cpu MHz         : 2400.099
cache size      : 30720 KB
physical id     : 0
siblings        : 1
[...]

PR #13609, opened 2019-03-22
[Unsafe] Sanatize the computed offset

lewurm:
We do 64bit operations, but operate on 32bit values. We need to sanitize those values.

Consider:
```
(lldb) x/8i 0x40011357
    0x40011357: 0f 84 c7 00 00 00  je     0x40011424
    0x4001135d: 49 8d 7c 24 14     leaq   0x14(%r12), %rdi
->  0x40011362: 49 8b c6           movq   %r14, %rax
    0x40011365: 48 d1 e0           shlq   %rax
    0x40011368: 48 03 f8           addq   %rax, %rdi
    0x4001136b: 0f b7 74 24 20     movzwl 0x20(%rsp), %esi
    0x40011370: 49 8b d7           movq   %r15, %rdx
    0x40011373: e8 bc 00 00 00     callq  0x40011434
(lldb) register read r14 rax rdi
     r14 = 0x0000000100000000
     rax = 0x0000000000000010
     rdi = 0x00007ffff7f48144
(lldb) si
(lldb) si
(lldb) si
(lldb) register read r14 rax rdi
     r14 = 0x0000000100000000
     rax = 0x0000000200000000
     rdi = 0x00008001f7f48144
```
In this snippet `r14` has some garbage in the 32bit upper half, which unfortunately is taken into account for the multiplication (`shl`) and addition in order to compute the address (`rdi`).

Several factors why we didn't see it earlier:
1. It doesn't trigger on AOT, because when compiling mscorlib.dll, `Unsafe.dll` isn't available to the AOT compiler and thus inlining won't happen. We use AOT where possible in our CI. When doing a call the upper halfs are properly cleared.
2. It doesn't trigger on macOS due to different memory organization.  We don't end up with gargabe in the registers.
3. Recently we added intrinsics for Unsafe.* https://github.com/mono/mono/pull/12605/files#diff-89fe30e1e87c5db21be052ddb0c8e28dR329 It masks this specific problem. Funnily enough, the assumed IL for the `Add ()` intrinsics didn't match what we had before. Now it does :-)

Fixes https://github.com/mono/mono/issues/13452
Fixes https://github.com/mono/mono/issues/13597

### DIFF
--- a/mcs/class/corlib/System.Runtime.CompilerServices/Unsafe.il
+++ b/mcs/class/corlib/System.Runtime.CompilerServices/Unsafe.il
@@ -239,8 +239,8 @@
         ldarg.0
         ldarg.1
         sizeof !!T
-        conv.i
         mul
+        conv.i
         add
         ret
   }
@@ -252,8 +252,8 @@
         ldarg.0
         ldarg.1
         sizeof !!T
-        conv.i
         mul
+        conv.i
         add
         ret
   }
@@ -266,6 +266,7 @@
         ldarg.1
         sizeof !!T
         mul
+        conv.i
         add
         ret
   }


lewurm:
@marek-safar once approved, we should consider backporting this up to `2018-08`, it breaks trivial scenarios on Linux (see linked issues).

marek-safar:
@monojenkins backport 2018-08

marek-safar:
@monojenkins backport 2018-10

marek-safar:
@monojenkins backport 2019-02

akoeplinger:
@lewurm should we upstream this to https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il ?

lewurm:
not sure yet if this is the correct fix, please don't merge.

I think the IL, as it used to be, is correct:
```
  .method public hidebysig static !!T& Add<T>(!!T& source, int32 elementOffset) cil managed aggressiveinlining
  {
        .maxstack 3
        ldarg.0
        ldarg.1.     /* type int32 (STACK_I4) */
        sizeof !!T.  /* type unsigned int32 (STACK_I4) */
        conv.i.      /* forces conversation to native int (STACK_PTR) */
        mul          /* so we have STACK_I4 * STACK_PTR = STACK_PTR, 
                      * which is a "PMUL" in mini, that operates on 64 bit regs. */
        add
        ret
  }
}
```
operating on `native int` makes sense, as we want to compute an address here.  The problem is the register containing `elementOffset` contains some "garbage" in the upper half. However, that "garbage" is there from the start when `IndexOf` is called:
```
* thread #1, name = 'mono-sgen', stop reason = breakpoint 2.1
    frame #0: 0x0000000040011310
->  0x40011310: subq   $0x38, %rsp  // start of IndexOf (char, int, int)
    0x40011314: movq   %rbx, (%rsp)
    0x40011318: movq   %r12, 0x8(%rsp)
    0x4001131d: movq   %r14, 0x10(%rsp)
(lldb) register read $rdx
     rdx = 0x0000000100000000
```
so we miss to clear the register properly somewhere else I think.

lewurm:
This is the correct fix: https://github.com/mono/mono/pull/13642

Sorry for the confusion.

PR #13624, opened 2019-03-23
Remove const on struct fields.

jaykrell:
It is unusual and not useful.
This was rejected previously.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #13625, opened 2019-03-23
[interp] fix "This greater-than-or-equal-to-zero comparison of an unsigned value is always true" reported by Coverity

lewurm:


jaykrell:
Sometimes these kinds of warnings bug me. You know, maybe you were thinking of changing the type to signed int, and coded for either case. They are always about the exact current code, never about any hypothetical evolution, or other configurations/targets, which are impossible to predict in general, granted.

lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #13628, opened 2019-03-23
[netcore] Marshal.IsPinnable

EgorBo:
Needed for MSBuild.
Stream changes were copied from [CoreRT](https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/IO/Stream.CoreRT.cs).
IsPinnable - perhaps it should be an icall/intrinsics for better perf?
[CoreCLR impl for IsPinnable (InternalCall)](https://github.com/dotnet/coreclr/blob/1f3f474a13bdde1c5fecdf8cd9ce525dbe5df000/src/vm/marshalnative.cpp#L257)
[CoreRT impl for IsPinnable](https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L114) (and [MightBeBlittable](https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/Runtime/InteropServices/InteropExtensions.cs#L60))

### DIFF
--- a/mcs/class/System.Private.CoreLib/System.IO/Stream.cs
+++ b/mcs/class/System.Private.CoreLib/System.IO/Stream.cs
@@ -2,14 +2,8 @@ namespace System.IO
 {
 	partial class Stream
 	{
-		private bool HasOverriddenBeginEndRead ()
-		{
-			throw new NotImplementedException ();
-		}
+		bool HasOverriddenBeginEndRead () => true;
 
-		private bool HasOverriddenBeginEndWrite ()
-		{
-			throw new NotImplementedException ();
-		}
+		bool HasOverriddenBeginEndWrite () => true;
 	}
 }
--- a/mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/Marshal.cs
+++ b/mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/Marshal.cs
@@ -49,7 +49,7 @@ namespace System.Runtime.InteropServices
 
 		internal static bool IsPinnable (object obj)
 		{
-			return (obj == null) || RuntimeTypeHandle.HasReferences (obj.GetType () as RuntimeType));
+			return (obj == null) || RuntimeTypeHandle.HasReferences (obj.GetType () as RuntimeType);
 		}
 
 		// TODO: Should be called from Windows only code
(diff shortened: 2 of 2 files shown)


mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -49,7 +49,7 @@ internal static IntPtr AllocBSTR (int length)
 
 		internal static bool IsPinnable (object obj)
 		{
-			throw new NotImplementedException ();
+			return (o == null) || RuntimeTypeHandle.HasReferences (obj.GetType () as RuntimeType));
 		}
```
vargaz: its called 'obj' not 'o'.

mcs/class/System.Private.CoreLib/System.IO/Stream.cs
```diff
@@ -2,14 +2,8 @@ namespace System.IO
 {
 	partial class Stream
 	{
-		private bool HasOverriddenBeginEndRead ()
-		{
-			throw new NotImplementedException ();
-		}
+		bool HasOverriddenBeginEndRead () => true;
```
marek-safar: This is incorrect, please implement it correctly instead

mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/Marshal.cs
marek-safar: is it ever called with null?

EgorBo: @marek-safar why not? `gcHandle.Target = null;` or `GCHandle.Alloc(null, ...);`

mcs/class/System.Private.CoreLib/System.IO/Stream.cs
EgorBo: Well, it's correct, it's just won't use an optimization https://github.com/dotnet/coreclr/blob/d0cdf8813ef099deec3b6940e34f20d1bba054b8/src/System.Private.CoreLib/shared/System/IO/Stream.cs#L1255-L1260 (CoreRT ignores it too).

I can try to implement an icall to scan method table for overriden Begin/End Write/Read.

marek-safar: I'll merge this but please implement this so we can benefit from the optimization

PR #13632, opened 2019-03-24
threads: make static mutexes PRIO_INHERIT

kurt-vd:
This merge request follows my previous one.
I found that some static mutexes exist, that still cause priority inversion problems.
This merge request addresses those static mutexes.


jaykrell:
I understand this but some questions.

1. Ok to leave the static init also, in case constructor attribute not understood and ignored?
2. Should we use constructor or MSVC pattern more?
3. Is there a lock around the constructors? Probably.
4. What is behavior when there is no PRIO_INHERIT? i.e. currently targeted Android.
5. Is there a way to statically initialize but with PRIO_INHERIT?
6. We should imho statically initialize more, when there is no PRIO_INHERIT, e.g. Windows and currently targeted Android.

jaykrell:
@monojenkins build

kurt-vd:
>     1. Ok to leave the static init also, in case constructor attribute not understood and ignored?
Is there a need, i.e. which compiler does not support constructor
attributes?
> 
>     3. Is there a lock around the constructors? Probably.
AFAIK, constructors are run before you enter main().
Unless someone launches threads from constructors already, there is no
need for locks yet.
> 
>     4. What is behavior when there is no PRIO_INHERIT? i.e. currently targeted Android.
the last patch silently ignores ENOTSUP.
This was detected on Alpine linux, where the underlying libmusl c
library does not support PTHREAD_PRIO_INHERIT.
Android, with bionic, should either not define PTHREAD_PRIO_INHERIT,
or return ENOTSUP too, but I don't know if it does.

> 
>     5. Is there a way to statically initialize but with PRIO_INHERIT?
I would have liked that, but I didn't find one in glibc.
> 
>     6. We should imho statically initialize more, when there is no PRIO_INHERIT, e.g. Windows and currently targeted Android.
Creating 2 flows seems more complex to me.
I don't see why constructors are a problem either.

kurt-vd:
>     2. Should we use constructor or MSVC pattern more?
I don't exactly understand the question. Can you elaborate on 'MSVC
pattern'?

jaykrell:
MSVC is Microsoft Visual C++.
It does not support this syntax, but it does support this feature.
What you do it put a put a pointer to each constructor function, into a particular section.
All pointers in that section are accumulated together by the linker, and the C startup code calls each pointer in the resulting array.

Statically initialize for the sake of systems without PRIO_INHERIT, imho.
Yes it is two flows, but they aren't that complicated and static initialization is a nice efficiency.
Only problem is that they very smaller non-zero init when a later runtime init will occur is a small cost, an extremely small one -- the data init bytes in the excutable.

kurt-vd:
> Statically initialize for the sake of systems without PRIO_INHERIT, imho.
> Yes it is two flows, but they aren't that complicated and static initialization is a nice efficiency.
> Only problem is that they very smaller non-zero init when a later runtime init will occur is a small cost, an extremely small one -- the data init bytes in the excutable.
If I understand well, you'd want to use static initializers, and if PRIO_INHERIT is supported, then run the constructor over it. With a little documentation describing the less-intuitive approach, I agree that this may be the best approach.

lambdageek:
@monojenkins build

libgc/aix_irix_threads.c
```diff
@@ -116,6 +118,15 @@ pthread_mutex_t GC_suspend_lock = PTHREAD_MUTEX_INITIALIZER;
 pthread_cond_t GC_suspend_ack_cv = PTHREAD_COND_INITIALIZER;
 pthread_cond_t GC_continue_cv = PTHREAD_COND_INITIALIZER;
 
+#if HAS_PTHREAD_PRIO_INHERIT
+__attribute__((constructor))
+__attribute__((unused))
+static void init_aix_irix_mutex(void)
+{
+	mono_os_static_mutex_init(&GC_suspend_lock);
```
lambdageek: I think we shouldn't change the Boehm code.
And we shouldn't make Boehm depend on a Mono header.

mono/utils/mono-static-mutex.h
```diff
@@ -0,0 +1,80 @@
...
+#define __MONO_STATIC_MUTEX_H__
+
+#if !defined(HOST_WIN32)
+#include <pthread.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define st_error(FMT, ...) \
```
lambdageek: `mono-os-mutex.h` uses `g_error` here.  Is there some reason you can't use `g_error` or `g_assertf`?

mono/eglib/gmisc-unix.c
```diff
@@ -30,6 +30,7 @@
 #include <stdlib.h>
 #include <glib.h>
 #include <pthread.h>
+#include "../mono/utils/mono-static-mutex.h"
```
lambdageek: The layering here is wrong - we don't make eglib depend on mono/utils.  You should define everything in eglib and use it in utils, metadata, etc.

mono/utils/mono-static-mutex.h
```diff
@@ -0,0 +1,80 @@
...
+static inline int
+mono_os_static_mutex_init(pthread_mutex_t *mutex)
+{
+	mono_os_static_mutex_init_type(mutex, PTHREAD_MUTEX_DEFAULT);
+}
+
+/* the actual test whether to run the constructor.
+ * It may become more fine-grained later on
+ */
+#define HAS_PTHREAD_PRIO_INHERIT	defined(PTHREAD_PRIO_INHERIT)
```
lambdageek: We should have a proper autoconf test for `PTHREAD_PRIO_INHERIT`.

mono/utils/atomic.c
```diff
@@ -18,8 +18,18 @@
 #if defined (WAPI_NO_ATOMIC_ASM) || defined (BROKEN_64BIT_ATOMICS_INTRINSIC)
 
 #include <pthread.h>
+#include "mono/utils/mono-static-mutex.h"
 
-static pthread_mutex_t spin G_GNUC_UNUSED = PTHREAD_MUTEX_INITIALIZER;
+static pthread_mutex_t spin = PTHREAD_MUTEX_INITIALIZER;
```
lambdageek: Lost `G_GNUC_UNUSED`...

support/signal.c
```diff
@@ -298,6 +299,15 @@ default_handler (int signum)
 
 static pthread_mutex_t signals_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+#if HAS_PTHREAD_PRIO_INHERIT
+__attribute__((constructor))
+__attribute__((unused))
```
lambdageek: 1. This is `G_GNUC_UNUSED`
2. We should make `G_GNUC_CONSTRUCTOR`

mono/eglib/gmisc-unix.c
jaykrell: I suspect this is due to bringing back old deleted mono code.

support/signal.c
jaykrell: Or something that works with msvc also? There is a way.

mono/utils/mono-static-mutex.h
kurt-vd: This was my approach to let eglib not depend on mono code.

mono/eglib/gmisc-unix.c
kurt-vd: I'm not well aware of the eglib vs. mono seperation.
If I understand well, I should put the header in eglib, and use it in mono?

mono/utils/atomic.c
kurt-vd: oops, good catch.

PR #13633, opened 2019-03-25
Fix DNS Name Resolution Failure on OpenBSD (GitHub issue #8168)

### LINKED ISSUE #8168: DNS: Name Resolution Failure
## Steps to Reproduce

1.  run any app that need to make use of DNS
2.  I've modified a snippet of code i've found on other issue to get the status of:
```
 Console.WriteLine("Socket.OSSupportsIPv4: " + Socket.OSSupportsIPv4);
 Console.WriteLine("Socket.SupportsIPv4: " + Socket.SupportsIPv4);

 Console.WriteLine("Socket.OSSupportsIPv6: " + Socket.OSSupportsIPv6);
 Console.WriteLine("Socket.SupportsIPv6: " + Socket.SupportsIPv6);
```
    And do a DNS query:
   `var addresses = Dns.GetHostAddresses("www.google.com");  `  
3. 
    The result:
```
  Socket.OSSupportsIPv4: False
  Socket.SupportsIPv4: False
  Socket.OSSupportsIPv6: False
  Socket.SupportsIPv6: False

  Unhandled Exception:
  System.Net.Sockets.SocketException (0x80004005): Could not resolve host 'www.google.com'
```

## Current Behavior

DNS lookup doesn't return any value and an exception is trown.
In sonarr app as example: [v2.0.0.5153] System.Net.WebException: DNS Name Resolution Failure: 'services.sonarr.tv'

## Expected Behavior

DNS lookup should return the IP address of the hostname

[ ] macOS
[ ] Linux
[ ] Windows
[x] DragonFly BSD (Yes i know...)

**Version Used**:
5.10.0.16 and 5.2.0.215
Both build from source with some BSD patches applied. (Freebsd and DragonFly)

If i set those variables to true before the DNS lookup call i have sucess. 
```
   var field = typeof(System.Net.Sockets.Socket).GetField("s_SupportsIPv4", 
   BindingFlags.Static|BindingFlags.NonPublic);
   field.SetValue(null, true);
   Console.WriteLine("Socket.SupportsIPv4: " + Socket.SupportsIPv4);
   Console.WriteLine("Socket.OSSupportsIPv4: " + Socket.OSSupportsIPv4);
```
Result:

```
Socket.SupportsIPv4: True
Socket.OSSupportsIPv4: True
172.217.16.228
```

I understand that DragonflyBSD is not "supported" but i would appreciate some help...
  How does Mono sets those variables? (OssSupportsIPV4, OssSupportsIPV6, ...)
  It is possible for me to force those to "True" in a global Mono base, so that every app that rellies on it is capable of using it?

I belive i just need a small patch or dependencie to get over this issue.

Thanks a lot,

rfht:
This adds OpenBSD to the platforms to use MacOsNetworkInterfaceAPI, fixes #8168, should apply similarly to DragonFly BSD. Tested with mono-5.14.0.177 from ports on OpenBSD with OpenRA and SMAPI for StardewValley that both do DNS lookup and exhibited NameResolutionFailure that is resolved with this diff.

### DIFF
--- a/mcs/class/System/System.Net.NetworkInformation/UnixNetworkInterfaceFactory.cs
+++ b/mcs/class/System/System.Net.NetworkInformation/UnixNetworkInterfaceFactory.cs
@@ -43,7 +43,7 @@ namespace System.Net.NetworkInformation {
 			if (runningOnUnix) {
 				// XXX: OpenBSD and NetBSD too? It seems other platforms map closer to the Mac OS version than Linux,
 				// even if not exactly; it seems Linux and/or glibc are the different ones.
-				if (Platform.IsMacOS || Platform.IsFreeBSD)
+				if (Platform.IsMacOS || Platform.IsFreeBSD || Platform.IsOpenBSD)
 					return new MacOsNetworkInterfaceAPI ();
 
 				// XXX: IBM i would be better with its own API targetting Qp2getifaddrs
--- a/mcs/class/System/System/Platform.cs
+++ b/mcs/class/System/System/Platform.cs
@@ -50,6 +50,7 @@ namespace System {
 
 #else
 		static bool isFreeBSD;
+		static bool isOpenBSD;
 
 		[DllImport ("libc")]
 		static extern int uname (IntPtr buf);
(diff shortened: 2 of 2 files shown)


marek-safar:
approve

directhex:
I'm merging this. Marek's approved it, and we don't CI OpenBSD, so we'll trust the community :man_shrugging:

NattyNarwhal:
This check should probably be inverted, actually. Most operating systems that use `getifaddrs` use the BSD-style structure, not the glibc style structure.

PR #13636, opened 2019-03-25
[wasm] [packager] Remove the unnecessary call to add_bindings()

kjpou1:
when runtime.js is generated.

Reported by @jeromelaban on mono-community.


vargaz:
Why is this unnecessary ? The user code will do it instead ?

kjpou1:
@vargaz The call is not longer necessary.  It is taken care of internally.

lewing:
@monojenkins build failed

PR #13656, opened 2019-03-25
[netcore] Change loader behaviour to disable assembly remapping…

filipnavara:
…and allow loading newer assembly versions.

Allows running `./dotnet build sample/HelloWorld`.

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -4221,7 +4221,13 @@ mono_assembly_candidate_predicate_sn_same_name (MonoAssembly *candidate, gpointe
 gboolean
 exact_sn_match (MonoAssemblyName *wanted_name, MonoAssemblyName *candidate_name)
 {
+#if ENABLE_NETCORE
+	gboolean result = mono_assembly_names_equal_flags (wanted_name, candidate_name, MONO_ANAME_EQ_IGNORE_VERSION);
+	if (result && assembly_names_compare_versions (wanted_name, candidate_name, -1) > 0)
+		result = false;
+#else
 	gboolean result = mono_assembly_names_equal (wanted_name, candidate_name);
+#endif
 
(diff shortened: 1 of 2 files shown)


mono/metadata/assembly.c
```diff
@@ -4221,7 +4221,13 @@ mono_assembly_candidate_predicate_sn_same_name (MonoAssembly *candidate, gpointe
 gboolean
 exact_sn_match (MonoAssemblyName *wanted_name, MonoAssemblyName *candidate_name)
 {
+#if ENABLE_NETCORE
+	gboolean result = mono_assembly_names_equal_flags (wanted_name, candidate_name, MONO_ANAME_EQ_IGNORE_VERSION);
+	if (result && assembly_names_compare_versions (wanted_name, candidate_name, -1) > 0)
```
lambdageek: why is this necessary? isn't this sort of thing usually papered over with a binding redirect in an app.config file?

filipnavara: Not anymore afaik. Binding redirects no longer exist on CoreCLR, or at least they should not be necessary. I will try to find a link to docs.

filipnavara: https://github.com/dotnet/coreclr/issues/14263#issuecomment-335975464

(will hopefully come with more authoritative documentation later)

lambdageek: Thanks @filipnavara for the link to this quote from @terrajobst 
https://github.com/dotnet/coreclr/issues/14263#issuecomment-335975464
>  [...] as .NET Core has a better binding policy and will happily load the higher version. 

(in a discussion of not supporting binding redirects anymore).

terrajobst: That's correct. Binding redirects are a .NET Framework only concept. No other binder we shipped since then has this. The general policy that we have now is that equal & higher versions satisfy the reference. Lower versions still fail.

lambdageek:
@monojenkins build failed

PR #13657, opened 2019-03-25
[System.Private.CoreLib] Implement StackFrame initialization

marek-safar:


netcore/Makefile
```diff
@@ -10,6 +10,9 @@ $(NETCORESDK_FILE):
 build-sample:
 	dotnet build HelloWorld
 
+run-sample:
+	COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(VERSION)" sample/HelloWorld/bin/Debug/netcoreapp3.0/HelloWorld.dll
```
EgorBo: build-sample & run-sample can be replace with `./dotnet run --fx-version "$(VERSION)" --project sample/HelloWorld`

filipnavara: Didn't work for me without #13656. I agree it should be done eventually.

PR #13661, opened 2019-03-26
Cleanup mono_marshal_get_icall_wrapper.

jaykrell:
Every caller was consistently forming the same name based on callinfo->name
and passing callinfo->func and callinfo->sig.
Pass callinfo instead and do the common work in mono_marshal_get_icall_wrapper.

This has the following side effects:
 Do not form and free name when there is cache hit. This seems like a nice optimization.
 Allow name to be null when there is cache hit. This is unintended. We could assert
 if it is a problem.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build Linux x64

jaykrell:

PR #13664, opened 2019-03-26
Fix a few deficiencies in mono_path_filename_in_basedir.

jaykrell:
The main one is that strstr is way too loose. The logic can be arbitrarily wrong depending on user paths.
The comments did not reflect observed reality.
Add tests.
Make the Win32 variant testable on Unix (and vice versa).

Add more data to asserts.

Account for test reality -- filename is not absolute.

Use HOST_WIN32 per https://github.com/mono/mono/pull/9930#pullrequestreview-144919252.
Comment was unclear, and regular ifdef will not meet the goals of the test.

Basedir might also not be a full path.

This is based on an older uncommited PR.
The cross-platform testability construct here was not approved at the time.

### DIFF
--- a/mono/unit-tests/test-path.c
+++ b/mono/unit-tests/test-path.c
@@ -1,0 +1,79 @@
+/*
+ * test-path.c
+ */
+
+#include "config.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include "glib.h"
+
+gboolean mono_host_win32;
+#undef HOST_WIN32
+#define HOST_WIN32 mono_host_win32
+#include "utils/mono-path.c"
+
+static char*
+make_path (const char *a, int itrail, int slash, int upcase, int win32)
+{
+	mono_host_win32 = !!win32;
+	char trail [3] = {'/', '/'};
+	trail [itrail] = 0;
+	char *b = g_strdup_printf ("%s%s", a, trail);
+	if (win32) {
+		if (slash)
+			g_strdelimit (b, '/', '\\');
+		if (upcase)
+			g_strdelimit (b, 'f', 'F');
+	}
+	return b;
+}
+
+int
+main (void)
+{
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/unit-tests/test-path.c
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#include "config.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include "glib.h"
+
+gboolean mono_host_win32;
+#undef HOST_WIN32
+#define HOST_WIN32 mono_host_win32
```
lambdageek: As mentioned before https://github.com/mono/mono/pull/9930#discussion_r245003497 we're not going to take this hack.

jaykrell: Testability technique?

jaykrell:
Please consider that this is not a hack.

It is a legitimate technique to form a more testable but not less efficient codebase.

In general `#ifdef` should be weeded out like this.

This is also for example, how you could form a native JIT/runtime with the same efficiency as today, but also a cross-compiler that targets multiple/all targets in one configuration -- by moving `#ifdef` to be `if` against what is either a just as optimizable constant, or a global variable (or thread local, or function call) that varies.

This enables testing, at least unit testing, Windows-specific code paths without using Windows or an alternate build system.

It provides an excellent balance between testability, testability of each configuration on all configurations, and does not compromise at all the runtime performance.

`#ifdef` on the other hand is very rigid. Tests are strongly tied to configuration and there is no ability to exercise code unless configuration matches exactly.

I do not want to be so quick to sacrifice testability when it is so easy to achieve.

PR #13665, opened 2019-03-26
Fix NRE when combobox was disposed and Text set

sancheolz:
If combobox already was disposed and we setting Text field there will be NullReferenceException. On .net winforms there no excetions.

### DIFF
--- a/mcs/class/System.Windows.Forms/Test/System.Windows.Forms/ComboBoxTest.cs
+++ b/mcs/class/System.Windows.Forms/Test/System.Windows.Forms/ComboBoxTest.cs
@@ -1504,33 +1504,19 @@ namespace MonoTests.System.Windows.Forms
         [Test]
         public void SetTextAfterDisposeTest()
         {
-            try
-            {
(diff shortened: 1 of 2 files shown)


mcs/class/System.Windows.Forms/Test/System.Windows.Forms/ComboBoxTest.cs
```diff
@@ -1500,7 +1500,39 @@ public void SortTest ()
 			cmb.Items.Add (new ComboVal ("A"));
 			cmb.Sorted = true;
 		}
-	}
+
+        [Test]
+        public void SetTextAfterDisposeTest()
+        {
+            try
```
marek-safar: The try/catch is redundant

sancheolz: Fixed

sancheolz:
@marek-safar could you backport it to 2019-02

marek-safar:
@monojenkins backport 2019-02

PR #13666, opened 2019-03-26
[wip] Reduce heap allocations and hash lookups in JIT icall implementation.

jaykrell:
[Do not merge][work in progress]

Per https://github.com/mono/mono/pull/11941#issuecomment-449157323.
The hash inserts remain, since some hash lookups remain but it is a good start.

Replace MONO_PATCH_INFO_JIT_ICALL with MONO_PATCH_INFO_JIT_ICALL_INFO.
MONO_PATCH_INFO_JIT_ICALL remains only in asserts and can be later removed.
There remain other string-based icall patch enums, this is only one
of two or a few.

It might help to replace or augment the pointers with an enum, particularly for ilgen and aot, as those still use the hashes. It'd be almost as good as pointer, and in some places better -- ilgen and esp. aot.

I had almost finished the enum approach, but got hung up on ICALL_ADDR and ICALL_ADDR_NOCALL, which are not handled here anyway. That is, the more common ICALL should probably change to enum, leave ICALL_ADDR[_NOCALL] as strings, at least in ilgen/aot, maybe they can be enum in some places. Their problem is at some dynamic strings are formed like for vtable slots and other cases.

This was originally written a few months ago (at the time of the PR comment).

### DIFF
--- a/mono/metadata/Makefile.am
+++ b/mono/metadata/Makefile.am
@@ -386,6 +386,8 @@ common_sources = \
 	w32handle.c	\
 	w32error.h	\
 	reflection.c	\
+	register-icall-def.c \
+	register-icall-def.h \
 	dynamic-image.c	\
 	sre.c	\
 	sre-encode.c	\
--- a/mono/metadata/boehm-gc.c
+++ b/mono/metadata/boehm-gc.c
@@ -41,6 +41,7 @@
 #include <mono/utils/mono-compiler.h>
 #include <mono/utils/unlocked.h>
 #include <mono/metadata/icall-decl.h>
+#include "register-icall-def.h"
 
 #if HAVE_BOEHM_GC
 
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -577,12 +577,15 @@ typedef struct MonoCachedClassInfo {
 
 typedef struct {
 	const char *name;
+	// FIXME try making these regular gpointer
 	gconstpointer func;
 	gconstpointer wrapper;
 	gconstpointer trampoline;
 	MonoMethodSignature *sig;
 	const char *c_symbol;
 	MonoMethod *wrapper_method;
+	MonoMethod *interp_lmf_wrapper;
+	gboolean inited;
 } MonoJitICallInfo;
 
 void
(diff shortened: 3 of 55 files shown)


vargaz:
Imho the only changed needed is the replacement of function names/strings with enums.

jaykrell:
@vargaz 

This was still work-in-progress not quite ready.

enums: That is kinda what this does, except it uses pointers to structs instead of indexing an array.
If you use enum, you get almost the same affect, somewhat worse, somewhat better.
I can try that instead. The diff will be *extremely* similar.

The enum will index into an array of preallocated structs.

Most references will have one instruction more than this PR, but some references -- AOT save and load -- will be much improved, and the hashtables *might* fall further way.
There are sticking points around dynamically generated JIT icall names, like for ICALL_ADDR and vtable slots.

Let me get this passing all lanes, it is close, and make a new branch from it switching to enums. Again, it'll be almost the same, a little worse, a little better.

vargaz:
The main problem with current impl is that it uses strings/function pointers for lookup, not the hash tables, those are not really important imho. So replacing the strings with enums would get most of the benefit with less work and changes.

jaykrell:
If you replace the strings and code pointers with enums, you get to replace the hashtable with an array. Granted, an array you cannot remove elements from. But the excess is small, and hashtable would possibly waste more space anyway.

I think enum is similar work and change as pointer, and I think that is ok. I will try it.

I'd like to get this version working first. The enum change I believe will be a fairly small increment/decrement from this. We'll see.
I had it almost done but got confused and went back to struct pointer.

Anyway, we'll see.

vargaz:
No, you can still keep the hash, just index it with enums instead of strings.

jaykrell:
I agree that works but it seems, not great, and easy to do much better. 

Almost every enum will be inserted. You can just index an array instead. 

The storage will be static and the density 100% and the diff about the same.

There is a hangup around some generated jit icall names that i need to double check on.

There might be a “need” to keep some as strings. We’ll see.

jaykrell:
> There is a hangup...

These are the ones I was worried about:

```
		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
```

```
		code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("trampoline_func_%d", tramp_type));
```

But I think they aren't registered, hashed, or regular `MONO_PATCH_INFO_JIT_ICALL` w/o `ADDR`
so maybe somewhat irrelevant.

They do prevent `MONO_PATCH_INFO_JIT_ICALL_ADDR` from going fully enum-based
w/o more work, but let's limit to plain `MONO_PATCH_INFO_JIT_ICALL` to start.

Probably the tramp ones should just be fully enumerated, and the slot ones can be a separate patch type. Later..

jaykrell:
So, it actually gets confusing with the enum,
because for the most part, pointers work
just as well, and it gets confusing to decide which
to use, in most places.

For AOT read/write, enum.
For ilgen, enum (so you don't need the extra data).
For const static arrays, enum wins slightly,
4 bytes instead of 8 and no relocation.
For the hash, yes, I think enums and not string,
 I haven't gotten there.

But everywhere else, pointers.

And then, you store all the values
in one array, and you can always compute
the enum from a pointer by subtract from start.

So the result is a hybrid.
Generally pointers, except for size/stability matter.
The sizes of locals and parameters don't generally matter.

I was on my way converting to all enum,
but this confusion set in.

I'll see where it goes, but leaving as an explanation
in case it ends up unexpected.

There is also an argument to leave AOT read/write as string.
That way the AOT file format is more resilient to change.
You can add/remove/reorder the enums w/o bumping AOT file version.
Like now dynamic linkers traffic more in strings and less in ordinals.

jaykrell:
The hashtables are keyed by string and by code pointer (two hashtables).
Most of the lookups are by constant string. Those lookups can be removed entirely.

To the extent that any lookups remain, which is still some, the hashtables remain.
  It might be possible to eliminate them but this PR doesn't get that far.

AOT and ilgen could traffic in strings or integers/enums.
strings are slower, but allow AOT file format to evolve w/o revising the version number.
Hard to decide that.

There are multiple relevant patch_types, JIT_ICALL, JIT_ICALL_ADDR, and JIT__ICALL_ADDR_NOCALL.

The last is always one specific function, but implemented using a string and shares code with the others.

The second includes some dynamically generated strings -- per tramp_type and per-slot.
I believe two new patch types should be introduced to handle those -- JIT_ICALL_ADDR_TRAMP and JIT_ICALL_ADDR_LAZY_FETCH_SLOT. That would enable removing strings from here entirely.
The hash table for code pointers would remain. It is used by abs patches, which I have to look at more closely.

jaykrell:
@monojenkins build failed

jaykrell:
This is mostly in master now.

PR #13667, opened 2019-03-26
[netcore] Handle NATIVE_DLL_SEARCH_DIRECTORIES for p/invoke

filipnavara:
Extracted the good parts out of https://github.com/mono/mono/pull/12613 and fixed up enough so it does not crash.

### DIFF
--- a/mono/metadata/loader.c
+++ b/mono/metadata/loader.c
@@ -1601,52 +1601,54 @@ mono_set_pinvoke_search_directories (int dir_count, char **dirs)
 #endif
 
 static MonoDl*
+pinvoke_probe_for_module_in_directory (const char *mdirname, const char *file_name, char **found_name_out)
+{
+	void *iter = NULL;
+	char *full_name;
+	MonoDl* module = NULL;
+
+	while ((full_name = mono_dl_build_path (mdirname, file_name, &iter)) && module == NULL) {
+		char *error_msg;
+		module = cached_module_load (full_name, MONO_DL_LAZY, &error_msg);
+		if (!module) {
+			mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_DLLIMPORT, "DllImport error loading library '%s': '%s'.", full_name, error_msg);
+			g_free (error_msg);
+		} else {
+			*found_name_out = g_strdup (full_name);
+		}
+		g_free (full_name);
+	}
+
+	return module;
+}
+
+static MonoDl*
 pinvoke_probe_for_module_relative_directories (MonoImage *image, const char *file_name, char **found_name_out)
 {
-	char *full_name;
 	char *found_name = NULL;
 	MonoDl* module = NULL;
 
 	g_assert (found_name_out);
 
-	int j;
-	void *iter;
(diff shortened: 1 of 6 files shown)


filipnavara:
I'm not entirely convinced this is the correct approach. I also get a lot of load attempts like this:
```
Mono: netcore DllImport fallback handler: wanted '/Users/filipnavara/Documents/mono/netcore/shared/Microsoft.NETCore.App/3.0.0-preview4-27514-06/System.Private.CoreLib.dll.dylib'
Mono: netcore DllImport fallback handler: wanted '/usr/local/lib/mono/aot-cache/amd64/System.Private.CoreLib.dll.dylib'
```
...which suggests some funky fallback logic and also that the code paths get hit for managed assemblies.

lambdageek:
> which suggests that the code paths get hit for managed assemblies when looking up the AOT versions
Yea this is a little bit of an abuse of the DL fallback handler. It's used for all native libs, including AOT images.  I think it's more for platform support than for Mono itself.

It might be worth seeing if we can just have a netcore version of `pinvoke_probe_for_module` that uses the native paths from the host, instead of using the DL fallback.

filipnavara:
> It might be worth seeing if we can just have a netcore version of pinvoke_probe_for_module that uses the native paths from the host, instead of using the DL fallback.

I'll check how much effort would that be. Probably not too much.

filipnavara:
For reference, here's the CoreCLR algorithm: https://github.com/dotnet/coreclr/blob/2f4a7564beb6f8c1f4929504ee2221953d25fc39/src/vm/dllimport.cpp#L6570-L6777

The basic structure is similar to what Mono does, but details differ. They also loop over prefix/suffix variations and each iteration they run inner loop over possible paths. I am not sure if Mono handles "libNAME.so.VERSION" the same way they do. Some further aspects can be configured using the `DefaultDllImportSearchPaths` attribute, which is not implemented in Mono for sure.

I will try to move the NATIVE_DLL_SEARCH_DIRECTORIES logic to `pinvoke_probe_for_module_relative_directories`. Rest can be implemented later.

mono/metadata/loader.c
```diff
@@ -1595,10 +1609,41 @@ pinvoke_probe_for_module_relative_directories (MonoImage *image, const char *fil
...
+			}
+			g_free (full_name);
+		}
 
+		if (j >= pinvoke_search_directories_count) {
+			g_free (mdirname);
+		}
+	}
+#else
 			for (j = 0; j < 3; ++j) {
```
filipnavara: The indentation is off here. I intentionally left it that way to keep the diff tidy, but I can fix it on request.

filipnavara:
> Might be worth refactoring the code so that the inner while loop is a separate helper function. 

Agreed, will do.

filipnavara:
Rebased on top of master to fix conflicts.

filipnavara:
@marek-safar Friendly ping on this one. I have to rebase it almost every time `netcore/Makefile[.am]` gets changed. It removes one of the hacks necessary to run the `netcore` samples / unit tests.

PR #13673, opened 2019-03-26
Runtime nupkg generation

directhex:
I changed the thing, so we can generate mono.nupkg runtime packages in the format consumed in core-setup. This isn't as immediately usable as the current makefile's "just squash Mono into a .zip", until some changes are made on the Core end (e.g. supporting `libmonosgen-2.0` in lieu of `libcoreclr`), but it's a starting point.

### DIFF
--- a/netcore/runtime.nuspec
+++ b/netcore/runtime.nuspec
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>runtime.$RID$.Microsoft.NETCore.Runtime.Mono</id>
+        <version>$VERSION$</version>
+        <description>Internal implementation package not meant for direct consumption.  Please do not reference directly. 
+The Mono runtime, and the base library, called mscorlib. It includes the garbage collector, JIT compiler, base .NET data types and many low-level classes.</description>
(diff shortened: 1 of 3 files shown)


netcore/runtime.nuspec
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>runtime.$RID$.Microsoft.NETCore.Runtime.Mono</id>
+        <version>$VERSION$</version>
+        <description>Internal implementation package not meant for direct consumption.  Please do not reference directly. 
+The Mono runtime, and the base library, called mscorlib. It includes the garbage collector, JIT compiler, base .NET data types and many low-level classes.</description>
```
filipnavara: Nit: There's no mscorlib, only System.Private.CoreLib

directhex: I copy-pasted the description from the .NET Core runtime nupkg, which also has no mscorlib :laughing:

filipnavara: Oh yeah. It had it by the time it was written, someone just forgot to update it :)

directhex:
(rebase due to changes in master in the interim)

directhex:
Oh for the love of... rebasing again.

directhex:
None of these failures are related to my changes, so merging.

PR #13674, opened 2019-03-26
[netcore] Implement Array.Sort

EgorBo:
```
make xtest-System.Runtime.Tests FIXTURE="-method System.Tests.ArrayTests.*"
```

Before:
```
Total: 2045, Errors: 0, Failed: 366, Skipped: 0, Time: 1.220s
```

After:
```
Total: 2045, Errors: 0, Failed: 223, Skipped: 0, Time: 1.188s
```

Most of the failures in ArrayTests are (un)expected `ArrayTypeMismatchException` (in Array.Copy?)
Also, it seems it's possible to move a bit more pieces from Array.CoreCLR.cs to the shared.
E.g. `Array.Clear` and this one (without the fast-path).

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/Array.cs
+++ b/mcs/class/System.Private.CoreLib/System/Array.cs
@@ -406,14 +406,11 @@ namespace System
 			object[] objItems = null;
 			if (objKeys != null)
 				objItems = items as object[];
-			if (objKeys != null && (items == null || objItems != null))
-			{
-				SorterObjectArray sorter = new SorterObjectArray(objKeys, objItems, comparer);
+			if (objKeys != null && (items == null || objItems != null)) {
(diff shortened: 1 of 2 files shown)


mcs/class/System.Private.CoreLib/System/Array.cs
```diff
@@ -394,7 +394,28 @@ public void SetValue (object value, int index1, int index2, int index3)
...
+			/* TODO: CoreCLR optimizes this case via an internal call
+			if (comparer == Comparer.Default)
+			{
+				bool r = TrySZSort(keys, items, index, index + length - 1);
+				if (r)
+					return;
+			}*/
+
+			object[] objKeys = keys as object[];
+			object[] objItems = null;
```
vargaz: Are these arrays really object[] ?

PR #13684, opened 2019-03-27
[wasm]: Disable managed collation on wasm.

baulig:
* The `SimpleCollator` class now implements a new internal interface `ISimpleCollator`.
  In `CompareInfo`, we reference it by that interface everywhere except `GetCollator ()`,
  so we can `#if`-def it out.

  Later on, we will use the linker to do this dynamically with some sort of conditional,
  but even then the interface is required to limit references to the `SimpleCollator` type
  to one single method.

* Add `Mono.Globalization.Unicode/SimpleCollator.cs` to `wasm_corlib.dll.exclude.sources`.

* `corlib/Makefile`: remove collation resources on WASM.

* Added new `[Category ("ManagedCollator")]` to the relevant tests.


baulig:
@monojenkins commit csproj

kg:
wasm tests for linq ordering appear to be broken by this change.

mcs/class/corlib/ReferenceSources/CompareInfo.cs
```diff
@@ -33,34 +33,69 @@
...
 		// Maps culture IDs to SimpleCollator objects
-		static Dictionary<string, SimpleCollator> collators;
+		static Dictionary<string, ISimpleCollator> collators;
 		static bool managedCollation;
 		static bool managedCollationChecked;
 
 		static bool UseManagedCollation {
 			get {
+#pragma warning disable 162
+				if (!GlobalizationSupport.SupportsManagedCollation)
```
marek-safar: I don't think we need another property. Change this to be for WASM
```
const bool UseManagedCollation false;
```

and let compiler to remove the unreachable code (with warnings)

mcs/class/corlib/ReferenceSources/CompareInfo.cs
```diff
@@ -33,34 +33,69 @@
...
+#pragma warning restore 162
 			}
 		}
 
-		SimpleCollator GetCollator ()
+		ISimpleCollator GetCollator ()
 		{
+#pragma warning disable 162
+			// This is `const bool SupportsManagedCollation = false` on WASM.
+			if (!GlobalizationSupport.SupportsManagedCollation)
```
marek-safar: Should not be needed with the change above

baulig:
> wasm tests for linq ordering appear to be broken by this change.

I didn't run the System.Core tests yet.  They're intentionally failing and need to be manually disabled like I did with the corlib tests.

baulig:
@marek-safar Either the `#pragma warning disable 162` has to stay or the `LIBRARY_WARN_AS_ERROR = yes` in the `mcs/class/corlib/Makefile` has to go away (could be WASM conditionally).  Which one do you prefer?

baulig:
@marek-safar Turns out that if I change that into a const, then I have to either disable the ` LIBRARY_WARN_AS_ERROR` or add those `#pragma warning disable 162` to the entire file.

However, I just realized that only by turning it into a constant we would get the same output from the compiler as my linker module.

mcs/class/corlib/ReferenceSources/CompareInfo.cs
baulig: Done.

mcs/class/corlib/ReferenceSources/CompareInfo.cs
baulig: Unreachable code still needs to compile, so we can't remove this method or the interface without also `#if`-def'ing everything that calls it.

However, at least for the time being I would like to keep the compiled output as closely to what my linker module would yield as output.

marek-safar:
@baulig I am fine with `#pragma warning disable 162` (even at file level if it becomes too wide spread in the file)

baulig:

mcs/class/corlib/corefx/GlobalizationSupport.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Globalization
+{
+	static class GlobalizationSupport
```
marek-safar: Please remove this unreachable file

baulig: It is only unreachable on WASM.

marek-safar: sure but for the rest it's true so no-op

baulig:
@monojenkins commit csproj

kjpou1:
@monojenkins build failed

akoeplinger:
Failures are unrelated, merging.

PR #13687, opened 2019-03-27
Fix incorrect use of FreeHGlobal on Marshal UTF8 string test.

lateralusX:
Make sure the correct free method is called for strings allocated by StringToAllocatedMemoryUTF8 or StringToCoTaskMemUTF8, ZeroFreeCoTaskMemUTF8. Using incorrect free method will lead to heap corruption on platforms where CoTaskMem is backed by a different heap compared to HGlobal.

StringToAllocatedMemoryUTF8 has been renamed in corefx to StringToCoTaskMemUTF8, so switch to that method for allocation call.

### DIFF
--- a/mcs/class/corlib/Test/System.Runtime.InteropServices/MarshalTest.cs
+++ b/mcs/class/corlib/Test/System.Runtime.InteropServices/MarshalTest.cs
@@ -197,7 +197,7 @@ namespace MonoTests.System.Runtime.InteropServices
 				if (srcString.Contains("\0"))
 					continue;
 
-				IntPtr ptrString = Marshal.StringToAllocatedMemoryUTF8(srcString);
+				IntPtr ptrString = Marshal.StringToCoTaskMemUTF8(srcString);
 				string retString = Marshal.PtrToStringUTF8(ptrString);
 
 				Assert.AreEqual (srcString, retString, "#" + i);
@@ -206,7 +206,7 @@ namespace MonoTests.System.Runtime.InteropServices
 					string retString2 = Marshal.PtrToStringUTF8(ptrString, srcString.Length - 1);
 					Assert.AreEqual (srcString.Substring(0, srcString.Length - 1), retString2, "#s" + i);
 				}
-				Marshal.FreeHGlobal(ptrString);
+				Marshal.ZeroFreeCoTaskMemUTF8 (ptrString);
 			}			
 		}
 		


jaykrell:
I understand your change looks good and correct, but is there a cleanup function that reliably matches `StringToAllocatedMemoryUTF8`, i.e. for a one line instead of two line change? (But hey 100 lines if needed.)

jaykrell:
I mean, you know, otherwise `StringToAllocatedMemoryUTF8` is not usable.

jaykrell:
Also `ZeroFreeCoTaskMemUTF8` sounds both slow -- `Zero` -- and specific `UTF8`. Can we use a more generic and correct `CoTaskMemFree`?

jaykrell:
I see some is answered "StringToAllocatedMemoryUTF8 has been renamed in corefx to StringToCoTaskMemUTF8"

They just leave the old name around for some sort of compatibility, but it actually must never be used?

lateralusX:
That API has been renamed in corefx to StringToCoTaskMemUTF8 and have a matching free method. This is just fixing test, so perf is not an issue.

jaykrell:
The `Zero` in the freer suggests it zeros the memory which would be slow. Or maybe that is normal now and recommended? I also realize this is test code, and these allocators are also slower than gc allocation.

lateralusX:
Looking at corefx test they are using both ZeroFreeCoTaskMemUTF8 and CoTaskMemFree for strings allocated using StringToCoTaskMemUTF8, so both should work in this context.

jaykrell:
I suggest avoid `Zero` to demonstrate best perf practices. `Zero` is probably like `SecureZeroMemory` before the free, like for sensitive data, like decrypted data or passwords, etc.

And then I checked the documentation. There are around 4 `Zero` functions and all but this one are documented that way. So only use `Zero` if you have a reason. Otherwise use "regular".

jaykrell:
There are 6 functions like this.

jaykrell:
Actually the documentations suggests you have an allocator mismatch this way. You probably don't.
For most of the functions there are "SecureAlloc" and "ZeroFree" but sometimes the documentation is incomplete. Anyway, sorry, this is much ado about *very* little. But you shouldn't use `Zero` here,.

jaykrell:
(or at least the samples match things up that way), anyway sorry, too noisy here..

PR #13688, opened 2019-03-27
[profiler] Add duration=NUM parameter to AOT profiler

radekdoulik:
It instructs the AOT profiler to write the data after NUM seconds
measured from runtime initialization.

Fixes: https://github.com/xamarin/xamarin-android/issues/2813

### DIFF
--- a/mono/profiler/aot.c
+++ b/mono/profiler/aot.c
@@ -10,6 +10,7 @@
 
 #include "aot.h"
 
+#include <mono/metadata/object-internals.h>
 #include <mono/metadata/profiler.h>
 #include <mono/metadata/tokentype.h>
 #include <mono/metadata/tabledefs.h>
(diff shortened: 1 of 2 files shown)


marek-safar:
@lambdageek please review

mono/profiler/aot.c
```diff
@@ -34,6 +36,8 @@ struct _MonoProfiler {
 	char *outfile_name;
 	mono_mutex_t mutex;
 	gboolean verbose;
+	MonoNativeThreadId helper_thread;
```
lambdageek: this field seems unused.

vargaz:
I think the proper way to fix this is to have some kind of api to control the profiler.

radekdoulik:
I though about various way to achieve this, in the end that was the simplest way for us as we can just adjust the arguments. (we set it by `adb setprop` from command line)

For later usage from UI I was thinking about listening on a socket, similar to what log profiler does.

radekdoulik:

radekdoulik:

Issue #13104, opened 2019-02-20
[dotnet/performance] Benchmarking Results and Reproduction Steps


alexanderkyte:
# Running the Benchmarks:

This is the process for running `dotnet/performance` benchmarks on the Mono Runtime. This uses the default AOT configuration with LLVM enabled. This isn’t the fullaot profile used on mobile. 

## Todo for fullAOT-based tests

- (Completed) Get `dotnet/performance` tests compiling with netstandard2.0 runtime
-- https://gist.github.com/alexanderkyte/592a60fe15c0783485b699d2932ab5e2
- (Not yet completed) Get NS2.0 working with testing_aot_full profile (was working for microbenchmarks, has since broken.)

## Current `--aot=llvm` Process:

```bash
mkdir $HOME/perf-mono && cd $HOME/perf-mono
git clone git@github.com:dotnet/performance.git
git clone git@github.com:mono/mono.git
```

```bash
cd $HOME/perf-mono
cd mono
./autogen.sh --with-runtime-preset=aot_llvm --disable-boehm --enable-llvm=yes
make -j32
```

```bash
cd $HOME/perf-mono
cd performance/src/benchmarks/micro
```

Now you’re going to want to edit `common.props` to add `net462` to test mono's desktop AOT profile.

```xml
     <!-- The Python script can narrow down the TFMs to what the user has asked for -->
     <TargetFrameworks>$(PYTHON_SCRIPT_TARGET_FRAMEWORKS)</TargetFrameworks>
     <TargetFrameworks Condition="'$(TargetFrameworks)' == '' AND '$(OS)' == 'Windows_NT'">net461;netcoreapp2.0;netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">netcoreapp2.0;netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">netcoreapp2.0;netcoreapp2.1;netcoreapp2.2;netcoreapp3.0;net462</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <GenerateDocumentationFile>False</GenerateDocumentationFile>
     <WarningLevel>4</WarningLevel>
```

And then continue to execute

```bash
sudo dotnet restore
sudo msbuild /p:Configuration=Release MicroBenchmarks.sln
```

[...]

PR #13697, opened 2019-03-27
[llvm] Propagate nonnull attribute through LLVM IR

alexanderkyte:
## Summary

This change allows LLVM to identify unnecessary null checks. We mark GOT accesses (including those made by LDSTR) and new object calls as nonnull.

Since LLVM won't propagate this, we propagate from definition to usage, across casts, and from usage to definition (conditionally).

This enables it to remove a significant portion of some benchmarks. 

In real-world code, we can expect to see this remove the unnecessary null checks made by private methods. 

### C#:

Note that the constant strings are trivially non-null. This PR spots and propagates that. 

```

   static void ThrowIfNull(string s)
    {
        if (s == null)
            ThrowArgumentNullException();
    }

    static void ThrowArgumentNullException()
    {
        throw new ArgumentNullException();
    }

    static int Bench(string a, string b, string c, string d)
    {
        ThrowIfNull(a);
        ThrowIfNull(b);
        ThrowIfNull(c);
        ThrowIfNull(d);

        return a.Length + b.Length + c.Length + d.Length;
    }

    [Benchmark(Description = nameof(NoThrowInline))]
    public int Test() => Bench("a", "bc", "def", "ghij");

```

### Before:

```

define hidden monocc i32 @NoThrowInline_MainClass_Bench_string_string_string_string(i64*  %arg_a, i64* %arg_b, i64* %arg_c, i64*  %arg_d) #6 gc "mono" {
BB0:
  br label %INIT_BB1

INIT_BB1:                                         ; preds = %BB0
  br label %INITED_BB2

INITED_BB2:                                       ; preds = %INIT_BB1
  br label %BB3

BB3:                                              ; preds = %INITED_BB2
  br label %BB2

BB2:                                              ; preds = %BB3
  notail call monocc void @NoThrowInline_MainClass_ThrowIfNull_string(i64* %arg_a)
  notail call monocc void @NoThrowInline_MainClass_ThrowIfNull_string(i64* %arg_b)
  notail call monocc void @NoThrowInline_MainClass_ThrowIfNull_string(i64* %arg_c)
  notail call monocc void @NoThrowInline_MainClass_ThrowIfNull_string(i64* %arg_d)
  %0 = bitcast i64* %arg_a to i32*
  %1 = getelementptr i32, i32* %0, i32 4
  %t50 = load volatile i32, i32* %1
  %2 = bitcast i64* %arg_b to i32*
  %3 = getelementptr i32, i32* %2, i32 4
  %t52 = load volatile i32, i32* %3
  %t53 = add i32 %t50, %t52
  %4 = bitcast i64* %arg_c to i32*
  %5 = getelementptr i32, i32* %4, i32 4
  %t55 = load volatile i32, i32* %5
  %t56 = add i32 %t53, %t55
  %6 = bitcast i64* %arg_d to i32*
  %7 = getelementptr i32, i32* %6, i32 4
  %t58 = load volatile i32, i32* %7
  %t60 = add i32 %t56, %t58
  br label %BB1

BB1:                                              ; preds = %BB2
  ret i32 %t60
}

```

### After:

Note: safepoint in below code is added by backend, not part of this change

```

define hidden monocc i32 @NoThrowInline_MainClass_Bench_string_string_string_string(i64* nonnull %arg_a, i64* nonnull %arg_b, i64* nonnull %arg_c, i64* nonnull %arg_d) #6 gc "mono" {
BB0:
  %0 = getelementptr i64, i64* %arg_a, i64 2
  %1 = bitcast i64* %0 to i32*
  %t50 = load volatile i32, i32* %1, align 4
  %2 = getelementptr i64, i64* %arg_b, i64 2
  %3 = bitcast i64* %2 to i32*
  %t52 = load volatile i32, i32* %3, align 4
  %t53 = add i32 %t52, %t50
  %4 = getelementptr i64, i64* %arg_c, i64 2
  %5 = bitcast i64* %4 to i32*
  %t55 = load volatile i32, i32* %5, align 4
  %t56 = add i32 %t53, %t55
  %6 = getelementptr i64, i64* %arg_d, i64 2
  %7 = bitcast i64* %6 to i32*
  %t58 = load volatile i32, i32* %7, align 4
  %t60 = add i32 %t56, %t58
  %8 = load i64*, i64** getelementptr inbounds ([37 x i64*], [37 x i64*]* @mono_aot_NoThrowInline_llvm_got, i64 0, i64 7), align 8
  %9 = load i64, i64* %8, align 4
  %10 = icmp eq i64 %9, 0
  br i1 %10, label %gc.safepoint_poll.exit, label %gc.safepoint_poll.poll.i

gc.safepoint_poll.poll.i:                         ; preds = %BB0
  %11 = load void ()*, void ()** bitcast (i64** getelementptr inbounds ([37 x i64*], [37 x i64*]* @mono_aot_NoThrowInline_llvm_got, i64 0, i64 25) to void ()**), align 8
  call void %11() #8
  br label %gc.safepoint_poll.exit

gc.safepoint_poll.exit:                           ; preds = %BB0, %gc.safepoint_poll.poll.i
  ret i32 %t60
}

```

## Dependencies 

This depends on https://github.com/mono/linker/pull/528.

## TODO:

1. Add usage to Wasm so tested 
2. Add unit tests
3. Benchmark change with https://github.com/mono/mono/issues/13104

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -4354,7 +4354,7 @@ add_gc_wrappers (MonoAotCompile *acfg)
 }
 
 static gboolean
-contains_cannot_specialize_attribute (MonoCustomAttrInfo *cattr)
+contains_disable_reflection_attribute (MonoCustomAttrInfo *cattr)
 {
 	for (int i = 0; i < cattr->num_attrs; ++i) {
 		MonoCustomAttrEntry *attr = &cattr->attrs [i];
@@ -4400,7 +4400,7 @@ mono_aot_can_specialize (MonoMethod *method)
 	if (!is_ok (cattr_error)) {
 		mono_error_cleanup (cattr_error);
 		goto cleanup_false;
-	} else if (cattr && contains_cannot_specialize_attribute (cattr)) {
+	} else if (cattr && contains_disable_reflection_attribute (cattr)) {
 		goto cleanup_true;
 	}
 
@@ -4409,7 +4409,7 @@ mono_aot_can_specialize (MonoMethod *method)
 	if (!is_ok (cattr_error)) {
 		mono_error_cleanup (cattr_error);
 		goto cleanup_false;
-	} else if (cattr && contains_cannot_specialize_attribute (cattr)) {
+	} else if (cattr && contains_disable_reflection_attribute (cattr)) {
 		goto cleanup_true;
 	} else {
 		goto cleanup_false;
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -4353,6 +4353,53 @@ add_gc_wrappers (MonoAotCompile *acfg)
 	}
 }
 
+static gboolean enable_method_specialization;
+
+gboolean
+mono_aot_can_specialize (MonoMethod *method)
+{
+	// FIXME: Cache!?
+	if (!enable_method_specialization)
+		return FALSE;
+
+	if (!method )
+		return FALSE;
+
+	if (method->wrapper_type != MONO_WRAPPER_NONE)
+		return FALSE;
+
+	// If it's not private, we can't specialize
+	if ((method->flags & METHOD_ATTRIBUTE_MEMBER_ACCESS_MASK) != METHOD_ATTRIBUTE_PRIVATE)
+		return FALSE;
+
+	// If it has the attribute disabling the specialization, we can't specialize
+	ERROR_DECL (cattr_error);
+	MonoCustomAttrInfo *cattr = mono_custom_attrs_from_method_checked (method, cattr_error);
+	if (!is_ok (cattr_error)) {
+		mono_error_cleanup (cattr_error);
+		goto cleanup_false;
+	} else if (!cattr) {
+		goto cleanup_true;
+	}
+
+	for (int i = 0; i < cattr->num_attrs; ++i) {
+		if (!cattr->attrs [i].ctor)
+			continue;
+
+		if (!strcmp (m_class_get_name (cattr->attrs [i].ctor->klass), "Reflected")) {
+			goto cleanup_false;
+		}
+	}
+
+cleanup_true:
+	mono_custom_attrs_free (cattr);
+	return TRUE;
+
+cleanup_false:
+	mono_custom_attrs_free (cattr);
+	return FALSE;
+}
+
 static void
 add_wrappers (MonoAotCompile *acfg)
 {
@@ -7999,6 +8046,8 @@ mono_aot_parse_options (const char *aot_options, MonoAotOptions *opts)
 		// Intentionally undocumented: Used for internal debugging
 		} else if (str_begins_with (arg, "dump")) {
 			opts->dump_json = TRUE;
+		} else if (str_begins_with (arg, "opt-callee")) {
+			enable_method_specialization = TRUE;
 		} else if (str_begins_with (arg, "llvmonly")) {
 			opts->mode = MONO_AOT_MODE_FULL;
 			opts->llvm = TRUE;
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -4353,16 +4353,29 @@ add_gc_wrappers (MonoAotCompile *acfg)
 	}
 }
 
-static gboolean enable_method_specialization;
+static gboolean
+contains_cannot_specialize_attribute (MonoCustomAttrInfo *cattr)
+{
+	for (int i = 0; i < cattr->num_attrs; ++i) {
+		MonoCustomAttrEntry *attr = &cattr->attrs [i];
+
+		if (!attr->ctor)
+			return FALSE;
+
+		if (strcmp (m_class_get_name_space (attr->ctor->klass), "System.Runtime.CompilerServices"))
+			return FALSE;
+
+		if (strcmp (m_class_get_name (attr->ctor->klass), "DisablePrivateReflectionAttribute"))
+			return FALSE;
+	}
+
+	return TRUE;
+}
 
 gboolean
 mono_aot_can_specialize (MonoMethod *method)
 {
-	// FIXME: Cache!?
-	if (!enable_method_specialization)
-		return FALSE;
-
-	if (!method )
+	if (!method)
 		return FALSE;
 
 	if (method->wrapper_type != MONO_WRAPPER_NONE)
@@ -4373,31 +4386,44 @@ mono_aot_can_specialize (MonoMethod *method)
 		return FALSE;
 
 	// If it has the attribute disabling the specialization, we can't specialize
+	//
+	// Set by linker, indicates that the method can be found through reflection
+	// and that call-site specialization shouldn't be done.
+	//
+	// Important that this attribute is used for *nothing else*
+	//
+	// If future authors make use of it (to disable more optimizations),
+	// change this place to use a new attribute.
 	ERROR_DECL (cattr_error);
-	MonoCustomAttrInfo *cattr = mono_custom_attrs_from_method_checked (method, cattr_error);
+	MonoCustomAttrInfo *cattr = mono_custom_attrs_from_class_checked (method->klass, cattr_error);
+
 	if (!is_ok (cattr_error)) {
 		mono_error_cleanup (cattr_error);
 		goto cleanup_false;
-	} else if (!cattr) {
+	} else if (cattr && contains_cannot_specialize_attribute (cattr)) {
 		goto cleanup_true;
 	}
 
-	for (int i = 0; i < cattr->num_attrs; ++i) {
-		if (!cattr->attrs [i].ctor)
-			continue;
+	cattr = mono_custom_attrs_from_method_checked (method, cattr_error);
 
-		if (!strcmp (m_class_get_name (cattr->attrs [i].ctor->klass), "Reflected")) {
-			goto cleanup_false;
-		}
+	if (!is_ok (cattr_error)) {
+		mono_error_cleanup (cattr_error);
+		goto cleanup_false;
+	} else if (cattr && contains_cannot_specialize_attribute (cattr)) {
+		goto cleanup_true;
+	} else {
+		goto cleanup_false;
 	}
 
-cleanup_true:
-	mono_custom_attrs_free (cattr);
-	return TRUE;
-
 cleanup_false:
-	mono_custom_attrs_free (cattr);
+	if (cattr)
+		mono_custom_attrs_free (cattr);
 	return FALSE;
+
+cleanup_true:
+	if (cattr)
+		mono_custom_attrs_free (cattr);
+	return TRUE;
 }
 
 static void
@@ -8046,8 +8072,6 @@ mono_aot_parse_options (const char *aot_options, MonoAotOptions *opts)
 		// Intentionally undocumented: Used for internal debugging
 		} else if (str_begins_with (arg, "dump")) {
 			opts->dump_json = TRUE;
-		} else if (str_begins_with (arg, "opt-callee")) {
-			enable_method_specialization = TRUE;
 		} else if (str_begins_with (arg, "llvmonly")) {
 			opts->mode = MONO_AOT_MODE_FULL;
 			opts->llvm = TRUE;
--- a/mono/mini/mini-llvm-cpp.cpp
+++ b/mono/mini/mini-llvm-cpp.cpp
@@ -265,6 +265,19 @@ mono_llvm_set_func_nonnull_arg (LLVMValueRef func, int argNo)
 	unwrap<Function>(func)->addParamAttr (argNo, Attribute::NonNull);
 }
 
+LLVMValueRef *
+mono_llvm_call_args (LLVMValueRef calli)
+{
+	CallInst *call = unwrap<CallInst> (calli);
+	unsigned int numOperands = call->getNumArgOperands ();
+
+	LLVMValueRef *ret = g_malloc (sizeof (LLVMValueRef) * numOperands);
+	for (int i=0; i < numOperands; i++)
+		ret [i] = wrap (call->getArgOperand (i));
+
+	return ret;
+}
+
 void
 mono_llvm_set_call_notailcall (LLVMValueRef func)
 {
--- a/mono/mini/mini-llvm-cpp.cpp
+++ b/mono/mini/mini-llvm-cpp.cpp
@@ -254,6 +254,12 @@ mono_llvm_set_call_nonnull_arg (LLVMValueRef calli, int argNo)
 }
 
 void
+mono_llvm_set_call_nonnull_ret (LLVMValueRef calli)
+{
+	unwrap<CallInst>(calli)->addAttribute (AttributeList::ReturnIndex, Attribute::NonNull);
+}
+
+void
 mono_llvm_set_func_nonnull_arg (LLVMValueRef func, int argNo)
 {
 	unwrap<Function>(func)->addParamAttr (argNo, Attribute::NonNull);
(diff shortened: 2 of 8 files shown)


mono/mini/mini-llvm.c
```diff
@@ -3968,6 +3992,45 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
...
+		for (int i = 0; i < num_params; i++) {
+			LLVMValueRef argument = args [i];
+
+			// Add to list of tracked callers/callees
+			GList *usages = (GList *) g_hash_table_lookup (ctx->module->lvalue_to_lcalls, argument);
+			usages = g_list_prepend (usages, lcall);
+			g_hash_table_insert (ctx->module->lvalue_to_lcalls, argument, usages);
+
+			if (g_hash_table_lookup (ctx->module->module_nonnull, argument)) {
+				g_assert (i < LLVMGetNumArgOperands (lcall));
```
lambdageek: Do we support variadic functions?   how does llvm represent variadic functions?   wondering if this assert would ever be false for varargs.

alexanderkyte: We're only propagating through direct calls, which by nature have to have identical types for the call sites and the functions. 

Documentation about that added here:
https://github.com/mono/mono/pull/13697/commits/ada64db2dda63866dd92e8f6f9d77f83e42c0d7b

mono/mini/mini-llvm.c
```diff
@@ -3968,6 +3992,45 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
...
 	lcall = emit_call (ctx, bb, &builder, callee, args, LLVMCountParamTypes (llvm_sig));
 
+	if (mono_aot_can_specialize (call->method)) {
+		LLVMValueRef lmethod = callee;
+
+		int num_params = LLVMCountParams (lmethod);
+		GArray *call_site_union = (GArray *) g_hash_table_lookup (ctx->module->method_to_call_info, call->method);
+
+		if (!call_site_union) {
+			call_site_union = g_array_sized_new (FALSE, TRUE, sizeof (gboolean), num_params);
```
lambdageek:
```suggestion
			call_site_union = g_array_sized_new (FALSE, TRUE, sizeof (gint32), num_params);
```

You're using it as a counter, below.

mono/mini/mini-llvm.c
```diff
@@ -9253,8 +9323,12 @@ mono_llvm_create_aot_module (MonoAssembly *assembly, const char *global_prefix,
 	module->plt_entries = g_hash_table_new (g_str_hash, g_str_equal);
 	module->plt_entries_ji = g_hash_table_new (NULL, NULL);
 	module->direct_callables = g_hash_table_new (g_str_hash, g_str_equal);
-	module->method_to_lmethod = g_hash_table_new (NULL, NULL);
+	module->module_nonnull = g_hash_table_new (NULL, NULL);
 	module->idx_to_lmethod = g_hash_table_new (NULL, NULL);
+	module->method_to_lmethod = g_hash_table_new (NULL, NULL);
+	module->lmethod_to_method = g_hash_table_new (NULL, NULL);
+	module->method_to_call_info = g_hash_table_new (NULL, NULL);
```
lambdageek: Do we care, particularly that this hash table (and the arrays in its values) are never freed?

alexanderkyte: It looks like we're never freeing any of these hash tables. It's good not to leak a lot of memory on exit, but the lifetime of these are such that they would be freed right before shutdown.

mono/mini/mini-llvm.c
alexanderkyte: Oh yikes, missed that change when I changed the type of this. Good catch.

mono/mini/mini-llvm.c
```diff
@@ -3968,6 +3992,45 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
...
+				g_array_insert_val (call_site_union, i, zero);
+		}
+
+		for (int i = 0; i < num_params; i++) {
+			LLVMValueRef argument = args [i];
+
+			// Add to list of tracked callers/callees
+			GList *usages = (GList *) g_hash_table_lookup (ctx->module->lvalue_to_lcalls, argument);
+			usages = g_list_prepend (usages, lcall);
+			g_hash_table_insert (ctx->module->lvalue_to_lcalls, argument, usages);
```
lambdageek: I'm not convinced this line does anything if there are 2 or more uses of an argument.  `g_hash_table_insert` (as opposed to `g_hash_table_replace`) is a no-op if `argument` was already in the table.   And it clearly must have been because we just looked it up.

alexanderkyte: The official docs seem to suggest otherwise. 

https://developer.gnome.org/glib/stable/glib-Hash-Tables.html

```
g_hash_table_insert (GHashTable *hash_table,
                     gpointer key,
                     gpointer value);
```

```
If the key already exists in the GHashTable its *current value is replaced with the new value*. If you supplied a value_destroy_func when creating the GHashTable, the old value is freed using that function. If you supplied a key_destroy_func when creating the GHashTable, the passed key is freed using that function.
```

Does our eglib differ from glib here?

lambdageek: ah interesting.
```
#define g_hash_table_insert(h,k,v)    g_hash_table_insert_replace ((h),(k),(v),FALSE)
#define g_hash_table_replace(h,k,v)   g_hash_table_insert_replace ((h),(k),(v),TRUE)
```
so that last `gboolean replace` parameter is about replacing the _key_.  The value is always replaced.   this is controlling whether to replace the key in the case when you have a custom key equivalence function.

mcs/class/corlib/Mono/Codegen.cs
```diff
@@ -0,0 +1,16 @@
...
+//
+//
+
+using System;
+
+namespace Mono {
+	public static class Codegen
+	{
+		[AttributeUsage(AttributeTargets.All)]
+		public class Reflected : Attribute
```
lambdageek: Usually attributes are named `SomethingAttribute` and then they can be applied either as `[SomethingAttribute]` or as `[Something]`.

mono/mini/aot-compiler.c
```diff
@@ -4344,6 +4344,58 @@ add_gc_wrappers (MonoAotCompile *acfg)
...
+		mono_error_cleanup (cattr_error);
+		success = FALSE;
+	} else if (!cattr) {
+		success = TRUE;
+	} else {
+		for (int i = 0; i < cattr->num_attrs; ++i) {
+			if (!cattr->attrs [i].ctor)
+				continue;
+
+			if (!strcmp (m_class_get_name (cattr->attrs [i].ctor->klass), "Reflected")) {
```
lambdageek: Maybe check the namespace, too.  "Reflected" might be a pretty common name for attributes.

marek-safar: Why not to use existing System.Runtime.CompilerServices.MethodImpl attribute with NoOptimization flag instead which this logic should honour anyway?

alexanderkyte: Interesting. We don't seem to actually use that attribute at all right now outside of monodis.

https://gist.github.com/alexanderkyte/357ac0e85b37b078b17c511da154c473

I guess my concern was that by adding this attribute to a method, we would disable *other* optimizations. 

I can definitely re-use it if that's the case though.

alexanderkyte:
Wow Github really mis-orders commits

mono/mini/mini-llvm.c
```diff
@@ -9623,6 +9675,83 @@ emit_aot_file_info (MonoLLVMModule *module)
...
+						item->argument = call_argument;
+						queue = g_slist_prepend (queue, item);
+					}
+				}
+			}
+			g_free (operands);
+
+			g_hash_table_insert (module->method_to_call_info, callee_method, call_site_union);
+		}
+		g_free (calls);
```
lambdageek: This is only freeing the head of the list.  should be `g_slist_free`.

alexanderkyte: Oh that's right. This was just a C array before. Will fix.

alexanderkyte:
https://gist.github.com/alexanderkyte/84be1514a9c86c5223cd3ff3a2be265f

alexanderkyte:
Waiting on https://github.com/mono/linker/pull/506 to bump this

mono/mini/mini-llvm-cpp.cpp
```diff
@@ -241,10 +241,114 @@ mono_llvm_set_preserveall_cc (LLVMValueRef func)
...
+		return calli->hasRetAttr (Attribute::NonNull);
+	} else if (InvokeInst *calli = dyn_cast<InvokeInst> (val)) {
+		return calli->hasRetAttr (Attribute::NonNull);
+	} else if (LoadInst *loadi = dyn_cast<LoadInst> (val)) {
+		return loadi->getMetadata("nonnull") != nullptr; // nonnull <index>
+	} else if (Instruction *inst = dyn_cast<Instruction> (val)) {
+		// If not a load or a function argument, the only case for us to
+		// consider is that it's a bitcast. If so, recurse on what was casted.
+		if (inst->getOpcode () == LLVMBitCast)
+			return mono_llvm_is_nonnull (wrap (inst->getOperand (0)));
```
lambdageek: maybe a loop instead of recursion?

alexanderkyte: Yeah, this is basically tail recursion. Can do.

mono/mini/mini-llvm.c
```diff
@@ -9757,10 +9891,33 @@ mono_llvm_emit_aot_module (const char *filename, const char *cu_name)
...
 				if (lmethod && LLVMTypeOf (callee) == LLVMTypeOf (lmethod)) {
 					mono_llvm_replace_uses_of (callee, lmethod);
+
+					g_hash_table_insert (directly_called, lmethod, ji->data.method);
 					mono_aot_mark_unused_llvm_plt_entry (ji);
 				}
 			}
 		}
+
+		// After direct calls are all computed, specialize directly called
```
lambdageek: So there's an analysis pass that computes `directly_called`. And starting from this comment, this is kind of a global optimization pass over the current module.
Might be worth pulling it out into a separate function so that it's clear that (1) it's optional, (2) where to add another such optimization. 

It'd be nice also to document what kinds of LLVM IR manipulations would invalidate `directly_called` in case we do more global passes.

mono/mini/mini-llvm.c
```diff
@@ -9757,10 +9891,33 @@ mono_llvm_emit_aot_module (const char *filename, const char *cu_name)
...
+
+			GArray *call_site_union = (GArray *) g_hash_table_lookup (module->method_to_call_info, method);
+
+			if (call_site_union)
+				g_assert (call_site_union->len == LLVMCountParams (lmethod));
+
+			for (int i = 0; call_site_union && i < call_site_union->len; i++)
+				if (g_array_index (call_site_union, gint32, i) == 0)
+					mono_llvm_propagate_nonnull_func (directly_called, module, lmethod, i);
+		}
```
lambdageek: Is there any reason why you don't just queue up work items for every `lmethod, i` all at once and then kick off a single call to `mono_llvm_propagate_nonnull_func` that drains the queue?  Unless I'm missing something, `mono_llvm_propagate_nonnull_func` doesn't actually care about the identity of the root work item, right?

(I'm not sure if doing it that way would help/hinder anything.  So this comment might just be bikeshedding)

lambdageek:
How can we test this on CI?

alexanderkyte:
@lambdageek testing this on CI was the point of https://github.com/mono/mono/pull/13697/commits/2ce909a22577dac6ba3700d57c50f37bd84d5634

This will run as part of wasm.

I can add some logging/assertions around the state for post-crash debugging of this though, if you forsee that being an issue.

vargaz:
Are there any size bechmarks to see the impact of this change ?

mono/mini/mini-llvm.c
alexanderkyte: @lambdageek the only reason I didn't build up an intermediate queue was that I wanted to avoid the allocation. The traversal pattern that we see above will be necessary either way if we want the same guarantee of termination (we can only decrement a nullable ref-count to zero once, after which it's dead). 

If we need to do the same traversal, building up the intermediate data structure on the C heap seemed like it would slow things down. This way, we have the queue-fill and queue-drain operations interleaved, keeping the work queue within reasonable (hopefully) bounds. 

If you think that joining the traversal and the action makes it a bit hard to read here, I can isolate the two into different functions without having to change allocation/control flow.

alexanderkyte: I think I'll just move all of this logic into mono_llvm_propagate_nonnull_func, so the nested loops are easy to see

alexanderkyte:
@vargaz that's what I was hoping to get from CI

alexanderkyte:
@vargaz you asked for sizes for build-aot-corlib and build-aot-System.Core

They're shown in the below gist:
https://gist.github.com/alexanderkyte/4020d89156b7f07f842319bd34babf02

alexanderkyte:
@monojenkins build failed

alexanderkyte:
1) @vargaz identified that we weren't getting expected size savings for nullref elimination on wasm, looking into that
2) CI shows failures in the mini test suite (working elsewhere), need to investigate that failure, probably some edge case.

alexanderkyte:
@lambdageek the annotation had been disabled on bitcode for LDSTR specifically. I fixed it.

mono/mini/mini-llvm.c
```diff
@@ -5023,10 +5047,21 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
...
-					cmp = LLVMBuildICmp (builder, cond_to_llvm_cond [rel], lhs, LLVMConstNull (LLVMTypeOf (lhs)), "");
-				else
-					cmp = LLVMBuildICmp (builder, cond_to_llvm_cond [rel], convert (ctx, lhs, IntPtrType ()), LLVMConstInt (IntPtrType (), ins->inst_imm, FALSE), "");
+				LLVMIntPredicate llvm_pred = cond_to_llvm_cond [rel];
+				if (LLVMGetTypeKind (LLVMTypeOf (lhs)) == LLVMPointerTypeKind && ins->inst_imm == 0) {
+					// We are emitting a NULL check for a pointer
+					gboolean nonnull = mono_llvm_is_nonnull (lhs);
+
+					if (nonnull && llvm_pred == LLVMIntEQ)
+						cmp = LLVMConstInt (LLVMInt1Type (), FALSE, FALSE);
```
lambdageek: why do we have to do this ourselves?  does LLVM not evaluate a comparison of a nonnull ptr and 0 to a constant boolean on its own?

alexanderkyte: I observed it not to be spotting that simplification on it's own. When I'm looking for it, I see multiple null checks on wasm being collapsed into a single "if true" branch when I have the above simplification. Otherwise it doesn't happen.

Now what's strange is that if the variable being compared is a function argument with the nonnull attribute on it, LLVM will do the optimization. 

It may be something we could get LLVM to do. An advantage of emitting a simple "true" though is that the bitcode we give LLVM is going to be simpler. When we have explicit null checks, that could amount to a lot of optimizations we aren't forcing LLVM to do before it can accurately see the control flow. In aggregate it probably helps compilation time a little bit too.

alexanderkyte:
The wasm failures are seen here: https://github.com/mono/mono/pull/14116 

Since this bump alone doesn't change the main driver code, and this PR isn't passing any arguments to it, it looks like there's an unrelated regression with the linker.

alexanderkyte:
@monojenkins build failed

alexanderkyte:
Failures seen here have been bisected in 
https://github.com/mono/mono/pull/14126

and the issue is filed here:
https://github.com/mono/linker/issues/541

alexanderkyte:
https://github.com/mono/mono/pull/14126/commits/a8cd756995e2494dc2f15f48dc72517f2b71c637 

should unblock us

alexanderkyte:
@monojenkins build failed

akoeplinger:
@monojenkins commit csproj

PR #13698, opened 2019-03-27
[2019-02][corlib] Fix exception 'replacement' in IsolatedStorageFile.MoveFile()

alexischr:
This was caused by changes in File.Move coming from https://github.com/mono/mono/pull/12000

### DIFF
--- a/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
+++ b/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
@@ -609,8 +609,12 @@ namespace System.IO.IsolatedStorage {
 
 			try {
 				File.Copy (source_full_path, dest_full_path, overwrite);
-			} catch (IOException) {
-				throw new IsolatedStorageException ("Operation not allowed.");
+			}
+			catch (IOException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
+			}
+			catch (UnauthorizedAccessException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
 			}
 		}
 
@@ -872,8 +876,12 @@ namespace System.IO.IsolatedStorage {
 
 			try {
 				Directory.Move (src_full_path, dest_full_path);
-			} catch (IOException) {
-				throw new IsolatedStorageException ("Operation not allowed.");
+			}
+			catch (IOException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
+			}
+			catch (UnauthorizedAccessException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
 			}
 		}
 
@@ -904,8 +912,8 @@ namespace System.IO.IsolatedStorage {
 
 			try {
 				File.Move (source_full_path, dest_full_path);
-			} catch (IOException) {
-				throw new IsolatedStorageException ("Operation not allowed.");
+			} catch (UnauthorizedAccessException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
 			}
 		}
 


mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
```diff
@@ -890,7 +890,7 @@ public void MoveFile (string sourceFileName, string destinationFileName)
 				throw new ArgumentException ("An empty file name is not valid.", "destinationFileName");
 
 			CheckOpen ();
-
+ß
```
filipnavara: This looks unintentional ;)

mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
```diff
@@ -872,7 +872,7 @@ public void MoveDirectory (string sourceDirectoryName, string destinationDirecto
 
 			try {
 				Directory.Move (src_full_path, dest_full_path);
-			} catch (IOException) {
+			} catch (UnauthorizedAccessException) {
```
lambdageek: `Directory.Move` docs say that it can throw `IOException` under some circumstances, too.  So we probably want to catch both, actually.

Also do you think it makes sense to set the inner Exception of the IsolatedStorageException?  (what's the coreFX IsolatedStorageFile do?)

lambdageek: Yea, CoreFX calls this function, which does use the 2 arg IsolatedStorageException constructor:
https://github.com/dotnet/corefx/blob/1453863bc998b63024987e13b1e7a45659b452de/src/System.IO.IsolatedStorage/src/System/IO/IsolatedStorage/IsolatedStorageFile.cs#L704-L709

lambdageek:
Test failures (for example on Linux x64) are related:
```
/mnt/jenkins/workspace/test-mono-pull-request-amd64/mcs/class/corlib/Test/System.IO.IsolatedStorage/IsolatedStorageFileTest.cs:931 
```
Looks like some operations that didn't used to throw are now throwing.

alexischr:
Can someone please check my sanity?

* `IsolatedStorageFileTest.CopyFile` fails because https://github.com/mono/mono/blob/2019-02/mcs/class/corlib/Test/System.IO.IsolatedStorage/IsolatedStorageFileTest.cs#L931 throws an exception
* It's an `IsolatedStorageException` so the code change here is probably active and a `UnauthorizedAccessException` was caught and replaced
* The `UnauthorizedAccesException` was also going to flow through `IsolatedStorageFileTest.CopyFile` , there is no `try/catch` for it or an `ExpectedException` attribute. 

Why wasn't it failing before?

lambdageek:
We should put back the catch for `IOException` - locally I see this File.Copy failing with a sharing violation `IOException` due to https://github.com/mono/mono/blob/5f5c51f3ef40a91053b38cc4ef9673098bab2fe9/mcs/class/corlib/Test/System.IO.IsolatedStorage/IsolatedStorageFileTest.cs#L936 and that makes the remaining tests break.  I'm not sure why the line numbers differ on the bot.  but I've only seen it fail locally on this later line.


```patch
diff --git a/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs b/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
index a9f3f56ed8d..6807daf5a53 100644
--- a/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
+++ b/mcs/class/corlib/System.IO.IsolatedStorage/IsolatedStorageFile.cs
@@ -611,6 +611,8 @@ namespace System.IO.IsolatedStorage {
 				File.Copy (source_full_path, dest_full_path, overwrite);
 			} catch (UnauthorizedAccessException e) {
 				throw new IsolatedStorageException ("Operation not allowed.", e);
+			} catch (IOException e) {
+				throw new IsolatedStorageException ("Operation not allowed.", e);
 			}
 		}
 
-- 
```

marek-safar:
@monojenkins backport 2018-04

marek-safar:
@monojenkins backport master

lambdageek:

Issue #7845, opened 2018-03-27
Don't install monograph


lambdageek:
`monograph` (https://github.com/mono/mono/tree/master/tools/monograph) is a utility that dumps a callgraph or type hierarchy information from a given assembly (in graphviz dot format).

I'm not sure who this is useful for, but this seems like a tool that ought to be built on top of IKVM or Cecil and not by embedding Mono.

We ship it, but it hasn't worked for a while because we never test it.

```
mono_os_mutex_lock: pthread_mutex_lock failed with "Invalid argument" (22)
Abort trap: 6
```

Since noone complained, I assume noone uses it.  So we should stop shipping it.

lambdageek:
The runtime crash will be fixed by https://github.com/mono/mono/pull/7836, so we can ship something that won't crash anymore.  But we shouldn't ship it at all.

akoeplinger:
Alternatively we could add a test so we don't break it again ;)

marek-safar:
It should be fine to stop shipping it, it's really special dev tools with no maintenance

directhex:
It's been removed on Linux since 4.0.0 packages. https://github.com/mono/mono/pull/13699 for Windows/Mac

PR #13699, opened 2019-03-27
Delete monograph. Closes https://github.com/mono/mono/issues/7845

directhex:
As requested, remove monograph.

From the relevant Github Issue:

> monograph (https://github.com/mono/mono/tree/master/tools/monograph) is a utility that dumps a callgraph or type hierarchy information from a given assembly (in graphviz dot format).

> I'm not sure who this is useful for, but this seems like a tool that ought to be built on top of IKVM or Cecil and not by embedding Mono.

> We ship it, but it hasn't worked for a while because we never test it.

Definitely warrants a .msi build before merging, I'm not 100% confident `nano` is a good .sln editor


kg:
For the record can you briefly describe monograph in the PR so it shows up in the commit comment?

akoeplinger:
@monojenkins build failed

akoeplinger:
@monojenkins build msi

akoeplinger:
@monojenkins build pkg and msi

Issue #9267, opened 2018-06-22
Release tarballs should be compressed with xz


thiagomacieira:
Mono should release its sources in a `.tar.xz` format. Quick benchmark for the 5.12.0 release file:

Compression Format | Tarball size | Decompression time
-------------------|--------------|-------------------
BZip2 | 204 MB | 31.04s
Xz | 132 MB | 9.89s

akoeplinger:
@directhex what do you think about this?

directhex:
It's a 1-line change to configure.ac, plus however many hours to fix assumptions made in our relevant Jenkins jobs

directhex:
I'm not terribly keen on supporting both.

Are there any distributions which can't handle .tar.xz? We only care about tarballs on Linux & friends, so if there's a distribution adjacent to our support matrix which can't handle .xz then I could be persuaded to do both - otherwise I'd prefer to just switch to .xz wholesale.

Any Debian derivative with dpkg 1.15.5 (Nov 2009) is fine with it. Not sure on RPM

thiagomacieira:
There's a lot of software that has switched to only .xz (and even .7z for Windows), ditching the .gz and especially ditching the .bz2 files. Unlike the Lempel-Ziv family of decompressors, bunzip2 takes a lot of CPU time to work. It's not only more bandwidth consumed for the project, it's also time for the user. It's so bad that decompressing the 5.12.0 release tarball on-the-fly while downloading stalls the download (averaged 5 MB/s while a pure download saturated my 100 Mbit/s office connection).

If you want to hedge your bets, add .xz and keep the current compression (bzip2) for another year, then check your download stats.

And if you want to be bold, try zstd.

Tarball Size | Decompression time
-------------|-------------------
141 MB | 3.047s

Just a pure xxxcat of the compressed tarball:

Decompressor | xxxxcat time
--------------|---------------
bunzip2 | 35.86 s
xzcat | 10.86 s
zstdcat | 1.23 s

akoeplinger:
I've seen build scripts that do stuff like `wget https://download.mono-project.com/sources/mono/mono-$(VERSION).tar.bz2` so I'd like to avoid breaking them if we can avoid it.

@directhex is there a reason why we can't easily do both?

PR #13700, opened 2019-03-27
Switch `make dist` format from BZip2 to XZ.

directhex:
Closes https://github.com/mono/mono/issues/9267

This will likely require fixups in Linux packaging (which is the only place we consume tarballs instead of git)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,7 +16,7 @@ AC_CANONICAL_HOST
 # The extra brackets are to foil regex-based scans.
 m4_ifdef([_A][M_PROG_TAR],[_A][M_SET_OPTION([tar-ustar])])
 
-AM_INIT_AUTOMAKE([1.9 dist-bzip2 tar-ustar no-dist-gzip foreign subdir-objects]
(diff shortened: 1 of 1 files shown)


marek-safar:
@directhex how much "stuff" will this break?

directhex:
@marek-safar looks like 7 jobs depend on Mono being a .tar.bz2, and they just need light alteration to be less picky, e.g.

`rm -f mono*tar.bz2`

`echo "NEWTARBALL = http://origin-download.mono-project.com/sources/mono/${BRANCHDIR}/mono-${fourpartversion}.tar.bz2" > ENVIRONMENT`

`<sourceFiles>mono-*.tar.bz2</sourceFiles>`

I don't think anything in actual repos depend on it, since like I said, we use git not tarballs on non-Linux

directhex:
@marek-safar https://github.com/mono/website/pull/438

akoeplinger:
let's backport to 2019-02 so that Mono 6.0 is the first one with this change.

@monojenkins backport 2019-02

PR #13707, opened 2019-03-28
Fix a few deficiencies in mono_path_filename_in_basedir. Still with testability but will remove if not desired.

jaykrell:
The main one is that strstr is too loose.
Add tests.
Make the Win32 variant testable on Unix (and vice versa).

Add more data to asserts.

Account for test reality -- filename is not absolute.

Use HOST_WIN32 per https://github.com/mono/mono/pull/9930#pullrequestreview-144919252.
Comment was unclear, and regular ifdef will not meet the goals of the test.

Basedir might also not be a full path.

Regarding `#ifdef` vs. `if`:

Please consider that this is not a hack.

It is a legitimate technique to form a more testable but not less efficient codebase.

In general `#ifdef` should be weeded out like this.

This is also for example, how you could form a native JIT/runtime with the same efficiency as today, but also a cross-compiler that targets multiple/all targets in one configuration -- by moving `#ifdef` to be `if` against what is either a just as optimizable constant, or a global variable (or thread local, or function call) that varies.

This enables testing, at least unit testing, Windows-specific code paths without using Windows or an alternate build system.

It provides an excellent balance between testability, testability of each configuration on all configurations, and does not compromise at all the runtime performance.

`#ifdef` on the other hand is very rigid. Tests are strongly tied to configuration and there is no ability to exercise code unless configuration matches exactly.

I do not want to be so quick to sacrifice testability when it is so easy to achieve without sacrificing runtime performance.

If testability is really not desired then I will remove it.

### DIFF
--- a/mono/unit-tests/Makefile.am
+++ b/mono/unit-tests/Makefile.am
@@ -101,10 +101,17 @@ test_mono_string_CFLAGS = $(test-cflags) -DMAIN=test_mono_string_main
 test_mono_string_LDADD = $(test_ldadd) $(mini_libs) $(sgen_libs) libtestlib_la-test-mono-string.lo
 test_mono_string_LDFLAGS = $(test_ldflags)
 
+test_path_SOURCES = test-path.c
+test_path_CFLAGS = $(test-cflags)
+test_path_LDADD = $(test_ldadd) $(mini_libs) $(sgen_libs)
+test_path_LDFLAGS = $(test_ldflags)
+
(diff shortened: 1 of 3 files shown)


jaykrell:
I tried reopening other PR which would seem more "conversational" but didn't see a way.
This is just meant to discuss that one, really.

jaykrell:
@monojenkinds build  Linux AArch64

jaykrell:

jaykrell:
Possible network flake, will retry gradually.

jaykrell:
Ok I'll do it with reduced but non-zero testability.

PR #13719, opened 2019-03-28
Fix a few deficiencies in mono_path_filename_in_basedir. Mainly strstr.

jaykrell:
The main one is that strstr is too loose. The intent is prefix matching, not searching the entire string.
Add tests.
 Despite the prior intent and comments, the input is not normalized/absolute. Assuming so broke tests.
  Note that case sensitivity is a function of file system, not
  operating system, but a lot of code gets it wrong, so here does too.

The Win32 code paths could be tested on Unix and vice versa
w/o compromise. See https://github.com/mono/mono/pull/13707
That is removed here per PR feedback.

### DIFF
--- a/mono/unit-tests/test-path.c
+++ b/mono/unit-tests/test-path.c
@@ -1,0 +1,79 @@
+/*
+ * test-path.c
+ */
+
+#include "config.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include "glib.h"
+
+gboolean mono_host_win32;
+#undef HOST_WIN32
+#define HOST_WIN32 mono_host_win32
+#include "utils/mono-path.c"
+
+static char*
+make_path (const char *a, int itrail, int slash, int upcase, int win32)
+{
+	mono_host_win32 = !!win32;
+	char trail [3] = {'/', '/'};
+	trail [itrail] = 0;
+	char *b = g_strdup_printf ("%s%s", a, trail);
+	if (win32) {
+		if (slash)
+			g_strdelimit (b, '/', '\\');
+		if (upcase)
+			g_strdelimit (b, 'f', 'F');
+	}
+	return b;
+}
+
+int
+main (void)
+{
+	static const char * const bases [2] = {"/", "/foo"};
+	static const char * const files [3] = {"/foo", "/foo/bar", "/foob"};
+
+	static const gboolean result [2][3] = {
+		{ TRUE, FALSE, TRUE },
+		{ FALSE, TRUE, FALSE }
+	};
+
+	int i = 0;
+	gboolean const verbose = !!getenv("V");
+
+	for (int win32 = 0; win32 <= 1; ++win32) {
+		for (int upcase_file = 0; upcase_file <= 1; ++upcase_file) {
+			for (int upcase_base = 0; upcase_base <= 1; ++upcase_base) {
+				for (int itrail_base = 0; itrail_base <= 2; ++itrail_base) {
+					for (int itrail_file = 0; itrail_file <= 2; ++itrail_file) {
+						for (int ibase = 1; ibase < G_N_ELEMENTS (bases); ++ibase) {
+							for (int ifile = 0; ifile < G_N_ELEMENTS (files); ++ifile) {
+								for (int islash_base = 0; islash_base <= 1; ++islash_base) {
+									for (int islash_file = 0; islash_file <= 1; ++islash_file) {
+										char *base = make_path (bases [ibase], itrail_base, islash_base, upcase_base, win32);
+										char *file = make_path (files [ifile], itrail_file, islash_file, upcase_file, win32);
+										verbose && printf ("mono_path_filename_in_basedir (%s, %s)\n", file, base);
+										gboolean r = mono_path_filename_in_basedir (file, base);
+										verbose && printf ("mono_path_filename_in_basedir (%s, %s, win32:%d):%d\n", file, base, win32, r);
+										g_assertf (result [ibase][ifile] == r,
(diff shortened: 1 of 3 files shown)


mono/utils/mono-path.c
```diff
@@ -177,27 +177,137 @@ mono_path_resolve_symlinks (const char *path)
...
+	mono_path_remove_trailing_path_separators (filename, &filename_length);
+	mono_path_remove_trailing_path_separators (basedir, &basedir_length);
+
+	// basedir_length can be 0 at this point and that is ok.
+
+	if  (!filename_length
+			|| filename_length <= basedir_length
+			|| (basedir_length && !mono_path_equal (filename, basedir, basedir_length)))
+		return FALSE;
+
```
jaykrell: This entire logic might need a revisit/rewrite, while continuing to abide by tests.
For example, does setting basedir ="/" work? Perhaps not reasonable. While I love short paths and regularly run \foo.cmd, I don't fill \ or / with .dlls/.exes.

mono/utils/mono-path.c
```diff
@@ -177,27 +177,137 @@ mono_path_resolve_symlinks (const char *path)
...
+static gboolean
+mono_path_equal (const char *a, const char *b, size_t length)
+{
+#ifdef HOST_WIN32
+	size_t i = 0;
+	for (i = 0; i < length && mono_path_char_equal (a [i], b [i]); ++i) {
+		// nothing
+	}
+	return i == length;
```
jaykrell: Consider return false in the loop, true after it. Clarity of current form is debatable. It is an idiom but maybe a stretch for too many people and not worth while. One instance, no big deal. More important is widespread convention.

mono/utils/mono-path.c
```diff
@@ -177,27 +177,137 @@ mono_path_resolve_symlinks (const char *path)
...
+	// /foo/1/ is in /foo.
+	// /foo//1/ is in /foo.
+	// /foo//1// is in /foo.
+
+	// /foo is not in /foo.
+	// /foo/ is not in /foo.
+	// /foob is not in /foo.
+	// /foo/1/2 is not in /foo.
+
+	const char *after_base = &filename [basedir_length];
```
jaykrell: This is where the casual read gets confusing. Maybe a comment pointing out that is the length of one string indexed into the other, or such.

mono/utils/mono-path.c
```diff
@@ -177,27 +177,137 @@ mono_path_resolve_symlinks (const char *path)
...
+	size_t after_base_length = filename_length - basedir_length;
+	size_t skip_separators = mono_path_path_separator_length (after_base, after_base_length);
+	after_base += skip_separators;
+	after_base_length -= skip_separators;
+	return skip_separators && !mono_path_contains_separator (after_base, after_base_length);
```
jaykrell: Maybe it is obvious, or maybe a redundant comment: "There must have been at least one separator, and then after any non-separators, no more separators. i.e. /1/2/3 is in /1/2, but /1/2/3/4 is not." Maybe I'm just being slow.

mono/utils/mono-path.c
```diff
@@ -177,27 +177,137 @@ mono_path_resolve_symlinks (const char *path)
...
  *
  * Return \c TRUE if \p filename is in \p basedir
  *
- * Both paths must be absolute and using \c G_DIR_SEPARATOR for directory separators.
+ * Both paths should be absolute and be mostly normalized.
  * If the file is in a subdirectory of \p basedir, returns \c FALSE.
  * This function doesn't touch a filesystem, it looks solely at path names.
+ *
+ * In fact, filename might not be absolute, in which case, FALSE.
+ * Ditto basedir.
```
jaykrell: Needs more comments, as indicated later.
Maybe "immediate subdirectory".
/1/2/3 is in /1/2, but /1/2/3/4 is not.
I originally misunderstood the intent.

There are examples below but they aren't as clear as I'd like.
I think I'll do this right now..

mono/utils/mono-path.c
jaykrell: I've attempted to write clarifying comments but, it isn't all that easy.

mono/utils/mono-path.c
jaykrell: I've attempted to write clarifying comments but, it isn't all that easy.

jaykrell:
@monojenkins build Linux i386

mono/unit-tests/test-path.c
```diff
@@ -26,12 +26,12 @@ make_path (const char *a, int itrail, int slash, int upcase)
 int
 main (void)
 {
-	static const char * const bases [2] = {"/", "/foo"};
-	static const char * const files [3] = {"/foo", "/foo/bar", "/foob"};
+	static const char * const bases [2] = {"/", "/1"};
+	static const char * const files [6] = {"/1", "/1/2", "/1/2/3", "/12", "/2", "/2/2/"};
```
lambdageek: but now you lost the case sensitive tests on win32...

jaykrell: Oops good point. I'll go with a/b/c and such (and change the upcasing)

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build OS X x64

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build 
Linux x64 All Profiles Build

jaykrell:
@monojenkins build Linux x64 All Profiles Build

jaykrell:
@monojenkins build API diff

jaykrell:
@monojenkins build API Diff

PR #13727, opened 2019-03-28
[netcore] More work on fixing tests

marek-safar:


mcs/class/System.Private.CoreLib/System.Reflection.Emit/ModuleBuilder.cs
```diff
@@ -456,19 +455,15 @@ public override Type GetType (string className, bool throwOnError, bool ignoreCa
 		public void SetCustomAttribute( ConstructorInfo con, byte[] binaryAttribute) {
 			SetCustomAttribute (new CustomAttributeBuilder (con, binaryAttribute));
 		}
-
-		public ISymbolWriter GetSymWriter () {
-			return symbolWriter;
-		}
-
-		public ISymbolDocumentWriter DefineDocument (string url, Guid language, Guid languageVendor, Guid documentType)
+/*
```
vargaz: This method does exist and its public in coreclr's SRE code.

marek-safar: not really, it's implementation detail check the public API at https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Emit/ref/System.Reflection.Emit.cs#L260

netcore/Makefile.am
```diff
@@ -99,7 +60,7 @@ xtest-%: prepare check-env dl-test-assets
 	cd assets/extracted/$* && \
 	COMPlus_DebugWriteToStdErr=1 $(CURDIR)/./dotnet --fx-version "$(VERSION)" xunit.console.dll $*.dll \
 		-notrait category=outerloop -notrait category=nonosxtests -notrait category=failing \
-		-notrait category=nonnetcoreapptests -noappdomain -noshadow -parallel all -verbose \
```
EgorBo: Without `-verbose` it's sometimes hard to tell which test crashed runtime (just in case)

PR #13729, opened 2019-03-29
Add `[Category ("LinkerNotWorking")]` to some tests that don't work with the linker.

baulig:


baulig:
@monojenkins build failed

PR #13736, opened 2019-03-29
Fixed casing

MarcosCobena:
Hi guys,

Yesterday I updated the WASM SDK and noticed an error while using Uint16Array: there's a misspeling in the binding_support.js file.

I've modified it on final mono.js one and works as expected.

Thanks for your work,

PS: We're using above here: https://github.com/MarcosCobena/WebGL.NET

—Marcos

### DIFF
--- a/sdks/wasm/binding_support.js
+++ b/sdks/wasm/binding_support.js
@@ -417,7 +417,7 @@ var BindingSupportLib = {
 					newTypedArray = new Int16Array(end - begin);
 					break;
 				case 8: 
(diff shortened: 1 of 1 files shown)


kg:
Assuming this passes tests (I don't see any reason why it wouldn't), this looks fine to me. Thank you for the fix!

kg:
@monojenkins build

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

PR #13737, opened 2019-03-29
Make the DnsTest more resilient to multi-IP DNS results

filipnavara:
Sometimes Google DNS is/was returning "dns.google" for reverse query of 8.8.8.8 and 8.8.4.4 instead of the expected "google-public-dns-a.google.com" and "google-public-dns-b.google.com". Since "dns.google" resolves to both 8.8.8.8 and 8.8.4.4 in random order the round-trip tests for IP -> hostname -> IP were failing when they got the other IP. This changes the logic in the hostname -> IP check to check whether the expected IP is any one of the returned ones (ie. non necessarily the first one). 

/cc @akoeplinger @kg

### DIFF
--- a/mcs/class/System/Test/System.Net/DnsTest.cs
+++ b/mcs/class/System/Test/System.Net/DnsTest.cs
@@ -75,8 +75,7 @@ namespace MonoTests.System.Net
 			IAsyncResult async = Dns.BeginResolve (site1Dot, null, null);
 			IPHostEntry entry = Dns.EndResolve (async);
 			SubTestValidIPHostEntry (entry);
-			var ip = GetIPv4Address (entry);
-			Assert.AreEqual (site1Dot, ip.ToString ());
+			CheckIPv4Address (entry, IPAddress.Parse (site1Dot));
 		}
 
 		[Test]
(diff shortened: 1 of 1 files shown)


mcs/class/System/Test/System.Net/DnsTest.cs
```diff
@@ -239,21 +238,18 @@ public void GetHostByName ()
...
 		}
 
-		static IPAddress GetIPv4Address (IPHostEntry h)
+		static void CheckIPv4Address (IPHostEntry h, IPAddress a)
 		{
-			var al = h.AddressList.FirstOrDefault (x => x.AddressFamily == AddressFamily.InterNetwork);
-			if (al == null)
-				Assert.Ignore ("Could not resolve an IPv4 address as required by this test case, e.g. running on an IPv6 only network");
-			return al;
+			var al = h.AddressList.FirstOrDefault (x => x.AddressFamily == AddressFamily.InterNetwork && x.Equals (a));
```
filipnavara: I can probably change this just to `.Contains`.

kg: Yeah, that'd be good.

jaykrell:
Seems like we should be mocking..

kg:
@jaykrell I might otherwise agree but I believe the point of these tests is to exercise the system layer in addition to the API layer. The test just happened to make an invalid assumption about DNS and the nature of typical layered DNS caches hid it.

filipnavara:
"Windows x64" is stuck 👎, the rest is green on first run 👍

PR #13738, opened 2019-03-29
Add missing roslyn assemblies to mcs package makefile

kg:
Roslyn picked up some new dependencies that I failed to add to a Makefile containing a list of its dependencies. This may fix some scenarios (not covered by CI) that result in csc and/or msbuild not working.


kg:

akoeplinger:
Failing tests are due to the unrelated DNS test problem, merging.

lambdageek:
Is there some smoketest we should be running so that we can catch this sort of thing on PRs instead of on the per-commit packaging job?

kg:
We're missing multiple smoke tests, I'm going to make an issue about it

PR #13742, opened 2019-03-29
[System] Clear MONO_CFG_DIR before starting mcs in CodeDOM CSharpCodeGenerator

akoeplinger:
This showed up in the xamarin-macios 2019-02 integration.
When you used Xslt with a custom C# script in a Xamarin.Mac Full application we'd try to launch mcs via CodeDOM to compile the script.

This started failing in 2019-02 because System.Native (mono-native) wasn't found and we now need it for basic File IO, the reason is that XM sets MONO_CFG_DIR to point to the Contents/MonoBundle in the app bundle where it packages the global config file with dllmaps.

However the started Mono wasn't able to find mono-native since the config there uses a different name for the file.

To prevent that we now clear MONO_CFG_DIR so it uses the default value for the mono process we're starting (note that setting it to String.Empty doesn't work)

### DIFF
--- a/mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
+++ b/mcs/class/System/Microsoft.CSharp/CSharpCodeGenerator.cs
@@ -89,8 +89,11 @@ namespace Microsoft.CSharp
 			/*
 			 * reset MONO_GC_PARAMS - we are invoking compiler possibly with another GC that
 			 * may not handle some of the options causing compilation failure
+			 *
+			 * reset MONO_CFG_DIR - we don't want to propagate the current config to another mono
 			 */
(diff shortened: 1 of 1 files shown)


akoeplinger:
Hm, it breaks on Helix. Need to take a closer look.

alexischr:
Consider just using `--config` for the System.Native dllmap you need? I definitely had another moment in monodroid where removing `MONO_CFG_DIR` was too big of a change: https://github.com/xamarin/monodroid/commit/79c23a087b87e38db19ab0ca3214a653609a0ac7

akoeplinger:
@alexischr in this case I don't know which `--config` value I'd need to pass since it should just use whatever system mono/mcs uses as the default. I'll make this fix specific to `XAMMAC_4_5` since it should only be an issue there (we usually want `MONO_CFG_DIR` to flow to child processes).

PR #13745, opened 2019-03-29
[netcore] Guard for skipFrames overflow, update test exclusions

filipnavara:


mcs/class/System.Private.CoreLib/System.Diagnostics/StackTrace.cs
```diff
@@ -39,7 +39,7 @@ void InitializeForCurrentThread (int skipFrames, bool needFileInfo)
 
 			StackFrame sf;
 			var frames = new List<StackFrame> ();
-			while ((sf = new StackFrame (skipFrames, needFileInfo)) != null && sf.GetMethod () != null) {
+			while (skipFrames >= 0 && (sf = new StackFrame (skipFrames, needFileInfo)) != null && sf.GetMethod () != null) {
```
filipnavara: Note that the `new StackFrame(...) != null` doesn't make sense, but it was already in the code before.

marek-safar: Could you please fix that as well

filipnavara: Done

marek-safar:
Thank you!

PR #13746, opened 2019-03-29
[System.Net.Http] Move Mac/iOS specific HttpMessageHandler's to Xamarin.*.dll

akoeplinger:
We decided to move these types into Xamarin products since they can't be used without the underlying infrastructure anyway.


mcs/class/System.Net.Http/System.Net.Http/HttpClient.macios.cs
```diff
@@ -0,0 +1,17 @@
+using System;
```
marek-safar: nit: redundant

akoeplinger:

PR #13749, opened 2019-03-29
Revert roslyn bump to fix msbuild and some other scenarios

kg:
MSBuild and/or csc have broken in some configurations and scenarios as a result of the roslyn bump, so this PR backs it out entirely, returning us to the version of Roslyn we were previously using. DIM remains a current-ish version of Roslyn as before.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5821,8 +5821,7 @@ fi
 AC_SUBST(mono_runtime)
 AC_SUBST(mono_runtime_wrapper)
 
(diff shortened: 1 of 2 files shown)


Therzok:
VSMac also needs to be updated for Roslyn/MSBuild support to work with 16.0 (3.x) roslyn, as targets have been moved to Current:

https://github.com/mono/monodevelop/issues/7372

PR #13750, opened 2019-03-29
Add tests to ensure `make install` prefix is functional

directhex:
Adds 5 quick checks, on prefix-tests tag:
* Build a basic file with mcs
* Build a basic file with Roslyn
* AOT an assembly
* LLVM AOT an assembly
* LLVM JIT an assembly

This uses loadedllvm, since that's what we ship on Linux.

This should let us spot regressions affecting installed prefixes sooner

### DIFF
--- a/scripts/ci/run-jenkins.sh
+++ b/scripts/ci/run-jenkins.sh
@@ -54,7 +54,7 @@ if [[ ${CI_TAGS} == *'osx-i386'* ]]; then EXTRA_CFLAGS="$EXTRA_CFLAGS -m32 -arch
 if [[ ${CI_TAGS} == *'osx-amd64'* ]]; then EXTRA_CFLAGS="$EXTRA_CFLAGS -m64 -arch x86_64 -mmacosx-version-min=10.9"; EXTRA_LDFLAGS="$EXTRA_LDFLAGS -m64 -arch x86_64" EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --with-libgdiplus=/Library/Frameworks/Mono.framework/Versions/Current/lib/libgdiplus.dylib"; fi
 if [[ ${CI_TAGS} == *'win-i386'* ]]; then PLATFORM=Win32; EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --host=i686-w64-mingw32 --enable-btls"; export MONO_EXECUTABLE="${MONO_REPO_ROOT}/msvc/build/sgen/Win32/bin/Release/mono-sgen.exe"; fi
 if [[ ${CI_TAGS} == *'win-amd64'* ]]; then PLATFORM=x64; EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --host=x86_64-w64-mingw32 --disable-boehm --enable-btls"; export MONO_EXECUTABLE="${MONO_REPO_ROOT}/msvc/build/sgen/x64/bin/Release/mono-sgen.exe"; fi
-if [[ ${CI_TAGS} == *'prefix-tests'* ]]; then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --enable-loadedllvm --prefix=${MONO_REPO_ROOT}/tmp/monoprefix"; fi
+if [[ ${CI_TAGS} == *'make-install'* ]]; then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --enable-loadedllvm --prefix=${MONO_REPO_ROOT}/tmp/monoprefix"; fi
 
 if   [[ ${CI_TAGS} == *'coop-suspend'* ]];   then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --disable-hybrid-suspend --enable-cooperative-suspend";
 elif [[ ${CI_TAGS} == *'hybrid-suspend'* ]]; then EXTRA_CONF_FLAGS="${EXTRA_CONF_FLAGS} --enable-hybrid-suspend";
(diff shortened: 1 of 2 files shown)


directhex:

scripts/ci/run-jenkins.sh
```diff
@@ -356,6 +357,7 @@ elif [[ ${CI_TAGS} == *'interpreter'* ]];              then ${MONO_REPO_ROOT}/sc
 elif [[ ${CI_TAGS} == *'mcs-compiler'* ]];             then ${MONO_REPO_ROOT}/scripts/ci/run-test-mcs.sh;
 elif [[ ${CI_TAGS} == *'mac-sdk'* ]];                  then ${MONO_REPO_ROOT}/scripts/ci/run-test-mac-sdk.sh;
 elif [[ ${CI_TAGS} == *'helix-tests'* ]];              then ${MONO_REPO_ROOT}/scripts/ci/run-test-helix.sh;
+elif [[ ${CI_TAGS} == *'prefix-tests'* ]];             then ${MONO_REPO_ROOT}/scripts/ci/run-test-make-install.sh;
```
akoeplinger: call this `make-install` so it matches the .sh file (and I think it's a better name).

scripts/ci/run-test-make-install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+${TESTCMD} --label=make-install --timeout=30m make -w install
+mkdir -p ${MONO_REPO_ROOT}/tmp/
```
akoeplinger: this shouldn't be necessary, it was created by `make install`.

scripts/ci/run-test-make-install.sh
```diff
@@ -0,0 +1,13 @@
...
+echo "class H{static void Main(){System.Console.WriteLine(\"ok\");}}" > ${MONO_REPO_ROOT}/tmp/hello.cs
```
akoeplinger:
```suggestion
echo "class H { static void Main () { System.Console.WriteLine (\"Hello World: \" + System.DateTime.Now); } }" > ${MONO_REPO_ROOT}/tmp/hello.cs
```

scripts/ci/run-test-make-install.sh
```diff
@@ -0,0 +1,13 @@
...
+export LD_LIBRARY_PATH=${MONO_REPO_ROOT}/tmp/monoprefix/lib:${LD_LIBRARY_PATH}
```
akoeplinger: Instead of just setting LD_LIBRARY_PATH we should do:

```bash
MONO_PREFIX=${MONO_REPO_ROOT}/tmp/monoprefix
export DYLD_FALLBACK_LIBRARY_PATH=$MONO_PREFIX/lib:$DYLD_LIBRARY_FALLBACK_PATH
export LD_LIBRARY_PATH=$MONO_PREFIX/lib:$LD_LIBRARY_PATH
export PATH=$MONO_PREFIX/bin:$PATH
```

and then use `mcs`, `csc` etc.

scripts/ci/run-test-make-install.sh
```diff
@@ -0,0 +1,13 @@
...
+${TESTCMD} --label=check-prefix-mcs --timeout=1m ${MONO_REPO_ROOT}/tmp/monoprefix/bin/mcs /out:${MONO_REPO_ROOT}/tmp/hello.exe ${MONO_REPO_ROOT}/tmp/hello.cs 
+${TESTCMD} --label=check-prefix-roslyn --timeout=1m ${MONO_REPO_ROOT}/tmp/monoprefix/bin/csc /out:${MONO_REPO_ROOT}/tmp/hello.exe ${MONO_REPO_ROOT}/tmp/hello.cs
+${TESTCMD} --label=check-prefix-aot --timeout=1m ${MONO_REPO_ROOT}/tmp/monoprefix/bin/mono --aot ${MONO_REPO_ROOT}/tmp/hello.exe
+${TESTCMD} --label=check-prefix-llvmaot --timeout=1m ${MONO_REPO_ROOT}/tmp/monoprefix/bin/mono --aot=llvm,llvm-path=/usr/lib/mono/llvm/bin ${MONO_REPO_ROOT}/tmp/hello.exe
+${TESTCMD} --label=check-prefix-llvmjit --timeout=1m ${MONO_REPO_ROOT}/tmp/monoprefix/bin/mono --llvm ${MONO_REPO_ROOT}/tmp/hello.exe
+
+${MONO_REPO_ROOT}/scripts/ci/run-upload-sentry.sh
```
akoeplinger: we don't need this one here.

scripts/ci/run-test-make-install.sh
```diff
@@ -0,0 +1,15 @@
...
+MONO_PREFIX=${MONO_REPO_ROOT}/tmp/monoprefix
+export DYLD_FALLBACK_LIBRARY_PATH=$MONO_PREFIX/lib:$DYLD_LIBRARY_FALLBACK_PATH
+export LD_LIBRARY_PATH=$MONO_PREFIX/lib:$LD_LIBRARY_PATH
+export PATH=$MONO_PREFIX/bin:$PATH
+
+${TESTCMD} --label=check-prefix-mcs --timeout=1m mcs /out:${MONO_REPO_ROOT}/tmp/hello.exe ${MONO_REPO_ROOT}/tmp/hello.cs 
+${TESTCMD} --label=check-prefix-roslyn --timeout=1m csc /out:${MONO_REPO_ROOT}/tmp/hello.exe ${MONO_REPO_ROOT}/tmp/hello.cs
+${TESTCMD} --label=check-prefix-aot --timeout=1m mono --aot ${MONO_REPO_ROOT}/tmp/hello.exe
+${TESTCMD} --label=check-prefix-llvmaot --timeout=1m mono --aot=llvm,llvm-path=/usr/lib/mono/llvm/bin ${MONO_REPO_ROOT}/tmp/hello.exe
+${TESTCMD} --label=check-prefix-llvmjit --timeout=1m mono --llvm ${MONO_REPO_ROOT}/tmp/hello.exe
```
marek-safar: Could we add also `msbuild /help` or something like that?

kg: I think testing msbuild via /help won't achieve anything, it'd need to compile a .cs file. It's a good idea as a baseline to make sure we at least have the dll i guess

directhex: This PR doesn't address msbuild. A future change can add "build, install, and use msbuild" as a step.

PR #13756, opened 2019-03-31
[netcore] Fix few Array tests

EgorBo:
Fixes 7 System.Tests.ArrayTests

ArrayTests still has a lot of failing tests for `ArrayTypeMismatchException` (tests expect it when we throw nothing and vice versa).

Mostly these two:
```
System.Tests.ArrayTests.ConstrainedCopy_UnreliableConversion_ThrowsArrayTypeMismatchException
System.Tests.ArrayTests.Copy_SourceAndDestinationNeverConvertible_ThrowsArrayTypeMismatchException
```

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/Buffer.cs
+++ b/mcs/class/System.Private.CoreLib/System/Buffer.cs
@@ -17,7 +17,7 @@ namespace System
 			if (src == null)
 				throw new ArgumentNullException (nameof (src));
 			if (dest == null)
-				throw new ArgumentNullException (nameof (dest));
+				throw new ArgumentNullException ("dst");
 
 			if (srcOffset < 0)
 				throw new ArgumentOutOfRangeException (nameof (srcOffset), SR.ArgumentOutOfRange_MustBeNonNegInt32);
--- a/mono/metadata/exception.c
+++ b/mono/metadata/exception.c
@@ -689,7 +689,10 @@ mono_exception_new_argument_internal (const char *type, const char *arg, const c
 	MonoStringHandle arg_str = arg ? mono_string_new_handle (mono_domain_get (), arg, error) : NULL_HANDLE_STRING;
 	MonoStringHandle msg_str = msg ? mono_string_new_handle (mono_domain_get (), msg, error) : NULL_HANDLE_STRING;
 
-	return mono_exception_from_name_two_strings_checked (mono_get_corlib (), "System", type, msg_str, arg_str, error);
+	if (!strcmp (type, "ArgumentException"))
+		return mono_exception_from_name_two_strings_checked (mono_get_corlib (), "System", type, msg_str, arg_str, error);
+	else
+		return mono_exception_from_name_two_strings_checked (mono_get_corlib (), "System", type, arg_str, msg_str, error);
 #else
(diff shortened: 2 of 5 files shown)


mono/metadata/exception.c
```diff
@@ -711,7 +711,7 @@ mono_exception_new_argument (const char *arg, const char *msg, MonoError *error)
 MonoExceptionHandle
 mono_exception_new_argument_null (const char *arg, MonoError *error)
 {
-	return mono_exception_new_argument_internal ("ArgumentNullException", arg, NULL, error);
+	return mono_exception_new_argument_internal ("ArgumentNullException", NULL, arg, error);
```
EgorBo:
```csharp
ArgumentException(message, paramName)
ArgumentNullException(paramName, message)
```
reversed order

netcore/excludes-System.Runtime.Tests.rsp
```diff
@@ -36,14 +36,6 @@
...
 -nomethod System.Tests.UriMethodTests.GetComponents_Invalid
 
-# Remote test runner can't load System.Globalization.Native
--nomethod System.Text.Tests.StringBuilderTests.Test_Insert_Float
--nomethod System.Text.Tests.StringBuilderTests.Test_Insert_Double
--nomethod System.Text.Tests.StringBuilderTests.Test_Insert_Decimal
--nomethod System.Text.Tests.StringBuilderTests.Test_Append_Float
--nomethod System.Text.Tests.StringBuilderTests.Test_Append_Double
--nomethod System.Text.Tests.StringBuilderTests.Test_Append_Decimal
-
```
EgorBo: there were fixed by https://github.com/mono/mono/pull/13667

mono/metadata/exception.c
vargaz: The prototype is:
mono_exception_new_argument_internal (const char *type, const char *arg, const char *msg, MonoError *error);

EgorBo: `mono_exception_new_argument_internal` is used for both `ArgumentException` and `ArgumentNullException` and while it works for `ArgumentException` (Message is the first argument in its constructor) it doesn't work for `ArgumentNullException` where `Message` is the second argument. Thus ParamName is always null for ANE.

vargaz: This will hopefully fix it:
https://gist.github.com/vargaz/b37ec8c3b5743975db11a15e88ee64cf

mcs/class/System.Private.CoreLib/System/Buffer.cs
```diff
@@ -17,7 +17,7 @@ public static void BlockCopy (Array src, int srcOffset, Array dest, int dstOffse
 			if (src == null)
 				throw new ArgumentNullException (nameof (src));
 			if (dest == null)
-				throw new ArgumentNullException (nameof (dest));
+				throw new ArgumentNullException ("dst");
```
EgorBo: can't use nameof here, this method also throws "dest" in ArgumentException below (and there is a test that expects both "dst" and "dest" I guess it's a typo and should be fixed in the coreclr repo (and corefx test).

marek-safar: That sounds like a typo, could you propose the change?

EgorBo:
@monojenkins build failed

PR #12434, opened 2019-01-15
[sdks,mac] Remove dependency on MXE in favor of MinGW

grendello:
It turns out that homebrew now has a package for the MinGW gcc-based windows
cross compiler which is awesome as it allows us to skip building MXE and use the
latest version of the GCC suite pre-packaged for Mac. This commit removes almost
all traces of MXE (except for `mxe.mk` itself, because I don't know if any bot
or external tool, whatever, use targets in it) and adds the following brew
packages to SDK bot provisioning as well as to the `provision-mxe` target of
Mono SDKs:

  * mingw-w64
  * xamarin/xamarin-android-windeps/mingw-zlib

The latter package builds a Windows version of zlib using MinGW and it comes
from Xamarin's own homebrew tap at https://github.com/xamarin/homebrew-xamarin-android-windeps/blob/f4cc90845ff1953800d8d71035566a12d9b7aa24/mingw-zlib.rb

A few things need to be solved before this PR builds successfully:

- [x]  zlib must be built for Windows before LLVM build is attempted
   * Installed by `homebrew` from our own tap
- [x] something must be done about [32-bit builds](https://gist.github.com/grendello/f488ecd69475286154e18838c0441e8a) on Mojave (I have Mojave and I can't complete the build atm)
   * Solved by installing Apple command line tools for Xcode 9.4.1 and macOS 10.13

PR #13768, opened 2019-04-01
[2018-10][sdks,mac] Remove dependency on MXE in favor of MinGW

grendello:
It turns out that homebrew now has a package for the MinGW gcc-based windows
cross compiler which is awesome as it allows us to skip building MXE and use the
latest version of the GCC suite pre-packaged for Mac. This commit removes almost
all traces of MXE (except for `mxe.mk` itself, because I don't know if any bot
or external tool, whatever, use targets in it) and adds the following brew
packages to SDK bot provisioning as well as to the `provision-mxe` target of
Mono SDKs:

  * mingw-w64
  * xamarin/xamarin-android-windeps/mingw-zlib

The latter package builds a Windows version of zlib using MinGW and it comes
from Xamarin's own homebrew tap at https://github.com/xamarin/homebrew-xamarin-android-windeps/blob/f4cc90845ff1953800d8d71035566a12d9b7aa24/mingw-zlib.rb

Additionally, this commit adds a new `configure` flag: `--with-static-zlib=PATH`
which allows one to specify the static zlib archive to use when linking Mono.
The static archive supersedes the otherwise indicated or detected zlib. This is
the recommended mode of operation for MinGW builds as it avoids problems with
`zlib.dll` versions on the target machine.

Backport of #12434

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -556,6 +556,13 @@ if test x$have_zlib = xyes; then
    ])
 fi
 
+AC_ARG_WITH(static-zlib, [  --with-static-zlib=PATH    use the specified static zlib instead of -lz],[STATIC_ZLIB_PATH=$with_static_zlib],[STATIC_ZLIB_PATH=])
+if test "x$STATIC_ZLIB_PATH" != "x"; then
+	have_static_zlib=yes
+	AC_SUBST(STATIC_ZLIB_PATH)
+fi
+
+AM_CONDITIONAL(HAVE_STATIC_ZLIB, test x$have_static_zlib = xyes)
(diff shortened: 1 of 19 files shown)


akoeplinger:
@grendello 

```
make: *** No rule to make target 'package-android-host-mxe-Win32', needed by 'archive-android'.  Stop.
```

grendello:
It seems that to get the PR working we'd need to backport a few more commits from master, some of which bump submodules and make changes that are too extensive to justify the effort of porting to 2018-10. Thus I'm closing the PR and we'll wait for 2019-02 to have mxe-less Mono in Xamarin.Android :)

PR #13776, opened 2019-04-01
[netcore] Fixes for System.Reflection tests.

vargaz:


mcs/class/corlib/System.Reflection/AssemblyName.cs
```diff
@@ -151,11 +151,7 @@ internal AssemblyName (SerializationInfo si, StreamingContext sc)
 				if (codebase == null)
 					return null;
 
-#if NETCORE
-				throw new NotImplementedException ();
-#else
 				return Mono.Security.Uri.EscapeString (codebase, false, true, true);
```
filipnavara: I am not a big fan of bringing back any dependency to Mono.[In]Security. CoreCLR has the implementation in managed code, so it should probably be moved to shared CoreLib partition and reused:

https://github.com/dotnet/coreclr/blob/dcf9f36c21022b8fce374559ec5fea8d11a0852b/src/System.Private.CoreLib/src/System/Reflection/AssemblyName.cs#L396-L577

(CoreRT throws PlatformNotSupportedException)

vargaz: Its just one file, we can move it to corelib for now.

filipnavara: I guess it should not block this PR, but I wanted to bring it to attention. I can do the code shuffling between repositories later.

marek-safar: Yes, AssemblyName should be in shared/

netcore/Makefile.am
```diff
@@ -53,9 +53,10 @@ clean:
 # '-parallel none -verbose' for debug, `-parallel all` to run all tests in parallel
 # we need COREFX_ROOT only for xunit.console.deps.json and CoreFx.Private.TestUtilities.dll
 xtest-%: prepare check-env dl-test-assets
+	echo -n "***************** $* *********************"
 	ln -sf $(CURDIR)/sdk/$(NETCORESDK_VERSION)/System.Text.Encoding.CodePages.dll assets/extracted/$*/
 	cp $(COREFX_BINDIR)/runtime/netcoreapp-OSX-Debug-x64/xunit.console.deps.json assets/extracted/$*/
-	cp $(COREFX_BINDIR)/runtime/netcoreapp-OSX-Debug-x64/CoreFx.Private.TestUtilities.dll assets/extracted/$*/
+	cp $(COREFX_BINDIR)/runtime/netcoreapp-OSX-Debug-x64/{CoreFx.Private.TestUtilities.dll,Microsoft.DotNet.PlatformAbstractions.dll} assets/extracted/$*/
```
filipnavara: Why was this necessary? My CoreFX build doesn't have this file. It does exist in the SDK though: `./sdk/3.0.100-preview4-010759/Microsoft.DotNet.PlatformAbstractions.dll`.

vargaz: Most test suites seem to require it.

filipnavara: I run `xunitall` without it and didn't hit it yet, so there must be some difference between our setups.

vargaz: Without it, I get:
     System.TypeInitializationException : The type initializer for 'System.PlatformDetection' threw an exception.
      ---- System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.DotNet.PlatformAbstractions, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies.

filipnavara: Do you have some specific xunit-<something> test where you can reproduce it? I will try it again, locally. Also, what CoreFX version do you have in COREFX_ROOT?

EgorBo: yeah works for me without it, I suspect you need to update your local corefx repo
see https://github.com/dotnet/corefx/issues/28620 (was fixed a month ago)

Issue #13775, opened 2019-04-01
New Mono drop crashing in C# symbol tests


jaredpar:
## Steps to Reproduce

1. Clone https://github.com/dotnet/roslyn
1. Make sure mono is on your path
1. Run `git reset --hard be54c4a899c7d7bb6360f68243759f62c30e5408` (ensures your before we disabled the tests)
1. Run `./build.sh --restore --build`
1. Run `./build.sh --test --testMono` 

The instability here caused us to temporarily disable C# symbol tests on Mono. Here is our tracking issue

https://github.com/dotnet/roslyn/issues/34646

### Current Behavior
Eventually this will crash with a SIGSEGV. May have to run it a few times. The actual test which causes the crash will vary from run to run. 

### Expected Behavior
Expectation is this test will succeed. 

### Mono Version
From our apt log

> Setting up mono-devel (6.1.0.713-0nightly2+ubuntu1604b1

## Stacktrace

```

=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================
/proc/self/maps:
00400000-0087d000 r-xp 00000000 08:21 8522697                            /usr/bin/mono-sgen
00a7c000-00a83000 r--p 0047c000 08:21 8522697                            /usr/bin/mono-sgen
00a83000-00a89000 rw-p 00483000 08:21 8522697                            /usr/bin/mono-sgen
00a89000-00aa1000 rw-p 00000000 00:00 0 
01791000-070fd000 rw-p 00000000 00:00 0                                  [heap]
402e8000-402f8000 rwxp 00000000 00:00 0 
40443000-40506000 rwxp 00000000 00:00 0 
405b8000-405c8000 rwxp 00000000 00:00 0 
4091a000-4099a000 rwxp 00000000 00:00 0 
4099c000-40bbc000 rwxp 00000000 00:00 0 
40bc8000-40c68000 rwxp 00000000 00:00 0 
40d57000-40f2d000 rwxp 00000000 00:00 0 
40f38000-40fe8000 rwxp 00000000 00:00 0 
40ffe000-4100e000 rwxp 00000000 00:00 0 
41010000-410a0000 rwxp 00000000 00:00 0 
410bf000-411f2000 rwxp 00000000 00:00 0 
411fd000-4120d000 rwxp 00000000 00:00 0 
[...]

PR #13785, opened 2019-04-02
[metadata] Don't free_aggregate_modifiers in mono_metadata_free_type

lambdageek:
The aggregate modifiers container is owned by a `MonoImageSet`.  The aggregate modifiers may be shared by more than one `MonoType` that happen to need the same modifiers. (Similar to how generic instantiations are shared)

When the image set is cleaned up the aggregate modifier cleanup will free the types in the aggregate modifiers container (the call to `free_aggregate_modifiers` in `mono_metadata_clean_for_image`).

With this extra call to `free_aggregate_modifiers` in `mono_metadata_free_type` may cause the same memory to be freed twice.

Fixes https://github.com/mono/mono/issues/13775

### DIFF
--- a/mono/metadata/metadata.c
+++ b/mono/metadata/metadata.c
@@ -4053,9 +4053,6 @@ do_mono_metadata_parse_type (MonoType *type, MonoImage *m, MonoGenericContainer 
 void
 mono_metadata_free_type (MonoType *type)
 {
-	if (type->has_cmods && mono_type_is_aggregate_mods (type))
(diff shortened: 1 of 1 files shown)


lambdageek:
Need to backport to 2019-04

lambdageek:

lambdageek:
@monojenkins build failed

lambdageek:
@monojenkins build failed

lambdageek:
@monojenkins build failed

PR #13803, opened 2019-04-02
[llvm] Make GOT a typed struct

alexanderkyte:
Our LLVM IR is littered with bitcasts due to the GOT being currently an untyped array. 

This allows opt to remove those bitcasts, and to then see that we aren't crafting pointers from ints that may alias to other memory we're working with.

These conversions can pose trouble for LLVM's alias analysis, causing it to give up.


vargaz:
Looks good.

alexanderkyte:
@monojenkins build failed

alexanderkyte:
@monojenkins build failed

alexanderkyte:
@monojenkins build failed

Issue #13804, opened 2019-04-02
[MacSDK] pkg-config is 32-bit only


alexischr:
Something has changed in how pkg-config is built, and instead of the expected lipo-ed 32-bit/64-bit Mach-O executable, we only get the 32-bit build.

For example:

* Download the latest 2018-10 package: https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-10/242/15d050a79274d3ad80fb273b38affff000da37e5/MonoFramework-MDK-5.20.1.10.macos10.xamarin.universal.pkg

* Install

* `$ file /Library/Frameworks/Mono.framework/Versions/Current/bin/pkg-config`:
    * `/Library/Frameworks/Mono.framework/Versions/Current/bin/pkg-config: Mach-O executable i386`

The pkg-config tarball/version hasn't changed so this is possibly because of XCode updates on the bots.

akoeplinger:
Fixed by https://github.com/mono/mono/pull/13805

akoeplinger:
I just saw the same issue with this mono which was built after the fix:

```
Mono JIT compiler version 6.0.0.184 (2019-02/3a8f93b9a47 Thu Apr 18 09:17:09 EDT 2019)
```

alexischr:
I have so far been unable to reproduce. @directhex do you have any insight on this issue or how to debug it? I am still not sure - it may be a bockbuild issue or an issue with pkg-config's build.

directhex:
It seems to be doing the right thing, in the build logs?

e.g. 

```
07:23:14  pkg-config 0.27
07:23:14  
07:23:14  pkg-config->start_build:
07:23:14  	pkg-config 0.27 <http://pkg-config.freedesktop.org/releases/pkg-config-0.27.tar.gz> md5: 3a4c9feab14b6719afd8904945d9b4e4)
07:23:14  
07:23:14  pkg-config->do_build:
07:23:14  	Building (arch: darwin-32)
07:23:14  
[...]

PR #13805, opened 2019-04-02
[MacSDK] Bump bockbuild for fix to https://github.com/mono/mono/issue…

alexischr:
https://github.com/mono/mono/issues/13804


alexischr:

alexischr:
```

**Comment by alexischr** [comment]
```
$ file /Library/Frameworks/Mono.framework/Versions/Current/bin/pkg-config
/Library/Frameworks/Mono.framework/Versions/Current/bin/pkg-config: Mach-O 64-bit executable x86_64
```

akoeplinger:
I cherry-picked and pushed this fix down to 2018-10. @alexischr do you remember if we hit this in 2018-08 too?

akoeplinger:
I checked the last 2018-08 .pkg and it also had the issue so I cherry-picked the fix to 2018-08 too.

alexischr:
@akoeplinger sorry I was out - I still understand this issue as build host-configuration dependent, so this would need to be applied to all CI'ed branches (2018-08 included yea)

PR #13809, opened 2019-04-02
[netcore] Fix WaitHandle.SignalAndWait

EgorBo:
This PR fixes 6 failing tests in `System.Runtime.Tests` and 3 tests in `System.Threading.Tests`.
Mono (both "netcore" and regular one) didn't throw exceptions for these two cases:

1:
```csharp
var toSignal = new Mutex();
var toWaitOn = new AutoResetEvent(true);
WaitHandle.SignalAndWait(toSignal, toWaitOn);
// Expected: "ApplicationException: Attempt to release mutex not owned by caller."
// Actual: (no exception)
```
2:
```csharp
var toSignal = new Semaphore(initialCount: 5, maximumCount: 5);
var toWaitOn = new AutoResetEvent(initialState: false);
WaitHandle.SignalAndWait(toSignal, toWaitOn);
// Expected: InvalidOperationException: The WaitHandle cannot be signaled because it would exceed its maximum count.
// Actual: hangs
```


mono/metadata/icall-def.h
```diff
@@ -1226,13 +1226,8 @@ NOHANDLES(ICALL(VOLATILE_25, "Write(uintptr&,uintptr)", ves_icall_System_Threadi
 NOHANDLES(ICALL(VOLATILE_26, "Write(ulong&,ulong)", ves_icall_System_Threading_Volatile_Write8))
 
 ICALL_TYPE(WAITH, "System.Threading.WaitHandle", WAITH_1)
-#if ENABLE_NETCORE
-HANDLES(WAITH_1, "SignalAndWaitCore", ves_icall_System_Threading_WaitHandle_SignalAndWait_Internal, gint32, 3, (gpointer, gpointer, gint32))
-HANDLES(WAITH_2, "Wait_internal", ves_icall_System_Threading_WaitHandle_Wait_internal, gint32, 4, (gpointer_ptr, gint32, MonoBoolean, gint32))
-#else
 HANDLES(WAITH_1, "SignalAndWait_Internal", ves_icall_System_Threading_WaitHandle_SignalAndWait_Internal, gint32, 3, (gpointer, gpointer, gint32))
```
EgorBo: Since I decided to throw exceptions from C# side (just like WaitHandle.CoreCLR.cs does) I can drop this `#if ENABLE_NETCORE`.

EgorBo:

EgorBo:
> Since you modified the icall, shouldn't you modify mscorlib to account for the additional values?

nothing changed for regular mono, icalls are the same

filipnavara:
> nothing changed for regular mono, icalls are the same

You return new values from the icall and they are not transferred to exceptions in mscorlib.

EgorBo:

EgorBo:
@monojenkins build failed

EgorBo:
@monojenkins build failed

EgorBo:

EgorBo:
@monojenkins build failed

filipnavara:

PR #13811, opened 2019-04-03
[netcore] Improve element type checks in Array.Copy and Array.ConstrainedCopy

filipnavara:
Fixes 192 tests:
`System.Runtime.Tests  Total: 31196, Errors: 0, Failed: 138, Skipped: 0, Time: 30.231s`

Many of the Array tests are still failing though.

/cc @EgorBo

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/Array.cs
+++ b/mcs/class/System.Private.CoreLib/System/Array.cs
@@ -133,17 +133,20 @@ namespace System
 			if (dst_type.IsEnum)
 				dst_type = Enum.GetUnderlyingType (dst_type);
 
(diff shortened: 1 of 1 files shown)


EgorBo:
Nice! I was thinking of replacing our implementation with CoreRT's which is fully managed.
At least for `Copy` and `ConstraintCopy`

filipnavara:
> I was thinking of replacing our implementation with CoreRT's which is fully managed.

That's also an option, but it relies on `EEType` which is CoreRT specific. This was just small modification to existing code. It covers most of the cases, but some obscure ones are still not accounted for in `CanAssignArrayElement`.

PR #13812, opened 2019-04-03
[llvm] Fix JIT support and remove support for old llvm versions.

vargaz:


lewurm:
llvm 3.6 is still used on Xamarin.iOS for some targets, right?

vargaz:
Yes.

lewurm:
@vargaz the LLVM JIT fix looks good though

PR #13814, opened 2019-04-03
[thread] Print function name when asserting MonoThreadInfo is non-NULL

jaykrell:
[thread] Print function name when asserting MonoThreadInfo is non-NULL.

### DIFF
--- a/mono/utils/mono-threads.c
+++ b/mono/utils/mono-threads.c
@@ -512,7 +512,7 @@ unregister_thread (void *arg)
 	MonoThreadHandle* handle;
 
 	info = (MonoThreadInfo *) arg;
-	g_assert (info);
+	g_assertf (info, ""); // f includes __func__
 	g_assert (mono_thread_info_is_current (info));
 	g_assert (mono_thread_info_is_live (info));
 
@@ -631,7 +631,7 @@ mono_thread_info_current (void)
 
 	We cannot function after cleanup since there's no way to ensure what will happen.
 	*/
-	g_assert (info);
+	g_assertf (info, ""); // f includes __func__
 
 	/*We're looking up the current thread which will not be freed until we finish running, so no need to keep it on a HP */
 	mono_hazard_pointer_clear (mono_hazard_pointer_get (), 1);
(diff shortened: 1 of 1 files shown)


mono/utils/mono-threads.c
```diff
@@ -512,7 +512,7 @@ unregister_thread (void *arg)
 	MonoThreadHandle* handle;
 
 	info = (MonoThreadInfo *) arg;
-	g_assert (info);
+	g_assert (info && "unregister_thread");
 	g_assert (mono_thread_info_is_current (info));
```
vargaz: The assert already contains a line number which can be used to identify the location if the exact version of the runtime is known.

jaykrell: Agreed, but I didn't see version in the issue. Still could lean on users to report that, or derive it from "Unity Xyz."

luhenry: @jaykrell you yourself added `g_assertf`, why not using it? statement vs expression problem?

jaykrell: I think either is ok. ‘f’ is required for dynamic values. Non-f suffices for constants.

jaykrell: Maybe ‘&&’ is obscure, making reader evaluate the true/falseness of it? ‘f’ nicely separates expression checked from reported information?

jaykrell:
Could also put func in g_assert/f for l asserts.

mono/utils/mono-threads.c
jaykrell: I had forgotten that `g_assertf` already includes `__func__`.
Perhaps it should honor `DISABLE_ASSERT_MESSAGES`.

jaykrell:
@monojenkins build failed

steveisok:
@vargaz @lambdageek please review when you get a chance.

jaykrell:
@monojenkins build

jaykrell:
I'm not able to squash but can edit some things ahead of that, will try.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #13816, opened 2019-04-03
More linker work; refemit, security and remoting

baulig:


baulig:
@monojenkins build

mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
```diff
@@ -698,7 +698,8 @@ public override MethodInfo MakeGenericMethod (Type [] methodInstantiation)
 #if FULL_AOT_RUNTIME
 				throw new NotSupportedException ("User types are not supported under full aot");
 #else
-				return new MethodOnTypeBuilderInst (this, methodInstantiation);
+				if (AppDomain.IsReflectionEmitSupported)
```
marek-safar: Use existing RuntimeFeature.IsDynamicCodeSupported property

mcs/class/corlib/System.Runtime.Remoting.Activation/ActivationServices.cs
```diff
@@ -88,6 +88,8 @@ public static IMessage RemoteActivate (IConstructionCallMessage ctorCall)
 
 		public static object CreateProxyFromAttributes (Type type, object[] activationAttributes)
 		{
+			if (!AppDomain.IsRemotingSupported)
```
marek-safar: This does not need to be dynamic check

mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
baulig: Sure, didn't know that already existed.

mcs/class/corlib/System.Runtime.Remoting.Activation/ActivationServices.cs
baulig: What do you mean?

baulig: I researched this and that method must go away.  However, since it is not called from the runtime, we can remove the conditional here and instead protect the call sites in `rttype.cs`.

PR #13821, opened 2019-04-03
[netcore] Simplify runtime and tests synchronization

marek-safar:


netcore/Makefile.am
```diff
@@ -54,13 +61,12 @@ clean:
 # we need COREFX_ROOT only for xunit.console.deps.json and CoreFx.Private.TestUtilities.dll
 xtest-%: prepare check-env dl-test-assets
 	echo -n "***************** $* *********************"
-	ln -sf $(CURDIR)/sdk/$(NETCORESDK_VERSION)/System.Text.Encoding.CodePages.dll assets/extracted/$*/
-	ln -sf $(CURDIR)/shared/Microsoft.AspNetCore.App/3.0.0-preview4-19180-01/Newtonsoft.Json.dll assets/extracted/$*/
+	cp $(COREFX_BINDIR)/runtime/netcoreapp-OSX-Debug-x64/Newtonsoft.Json.dll assets/extracted/$*/
```
EgorBo: We don't have to copy System.Text.Encoding.CodePages.dll anymore?

marek-safar: not for the tests I tried. It's possible we still need but we need to figure out where to copy it from (we no longer download sdk)

PR #13794, opened 2019-04-02
Add MONO_PATCH_INFO_TRAMPOLINE_FUNC_ADDR and MONO_PATCH_INFO_SPECIFIC_TRAMPOLINE_LAZY_FETCH_ADDR.

jaykrell:
The larger goal here is to stop hashing JIT icalls, by string and by code address, as suggested a few months ago. Both the inserts into hashtables and the lookups.

Note this is about JIT icalls.
Not icalls declared in C#.

Reducing the lookups is easy, but complete elimination takes more work,
and I believe is possible and desirable.

JIT icalls are used with:
  `MONO_PATCH_INFO_JIT_ICALL` 
  `MONO_PATCH_INFO_JIT_ICALL_ADDR` 
  `MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL` 

There is a fixed set of JIT icalls.
Their strings can be replaced with an enum,
or pointers to fixed data structures, or pointers/offsets
into a fixed data structure.

Except for:
  `MONO_PATCH_INFO_JIT_ICALL_ADDR + sprintf ("trampoline_func_%d", tramp_type)`
  `MONO_PATCH_INFO_JIT_ICALL_ADDR + sprintf ("specific_trampoline_lazy_fetch_%u", slot)`

This change splits those uses into separate patch types, along with
the dynamic integer. Instead of forming the string.

As well, an important theory here is, while these were reusing
some JIT icall infrastructure, they were not using all
of it. These were not actually placed in the hashtables,
and if they were looked up, they were not found.

If that theory is in fact false, then this effort fails.

A separate do-not-merge PR exists to help test that theory (https://github.com/mono/mono/pull/13792
).

It is fairly obvious in the code, in that these JIT icalls
are not passed to any `register` function.

The int vs. uint nuance in this PR not clearly required, but follows
the prior code. They can probably all be int or uint.
As well we might consider guint32 instead of guint, if Mono is ever
to run with a 16bit integer.

PR #13792, opened 2019-04-02
[do not merge] Add some (inefficient) asserts to confirm my understan…

jaykrell:
…ding of the system to validate a change.

In particular:

```
Searching for '%u", slot'...
C:\s\mono2\mono\mini\tramp-amd64.c(718):  code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
C:\s\mono2\mono\mini\tramp-arm.c(699):  ji = mono_patch_info_list_prepend (ji, code - buf, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
C:\s\mono2\mono\mini\tramp-arm64.c(474):  code = mono_arm_emit_aotconst (&ji, code, buf, ARMREG_IP0, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
C:\s\mono2\mono\mini\tramp-mips.c(429):  ji = mono_patch_info_list_prepend (ji, code - buf, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
C:\s\mono2\mono\mini\tramp-ppc.c(641):  code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));
C:\s\mono2\mono\mini\tramp-x86.c(493):  code = mono_arch_emit_load_aotconst (buf, code, &ji, MONO_PATCH_INFO_JIT_ICALL_ADDR, g_strdup_printf ("specific_trampoline_lazy_fetch_%u", slot));


Searching for 'trampoline_func_'...
C:\s\mono2\mono\mini\aot-runtime.c(5382):				} else if (strstr (ji->data.name, "trampoline_func_") == ji->data.name) {
C:\s\mono2\mono\mini\aot-runtime.c(5383):					MonoTrampolineType tramp_type2 = (MonoTrampolineType)atoi (ji->data.name + strlen ("trampoline_func_"));
[...]

PR #13837, opened 2019-04-04
Add MONO_PATCH_INFO_TRAMPOLINE_FUNC_ADDR and MONO_PATCH_INFO_SPECIFIC_TRAMPOLINE_LAZY_FETCH_ADDR.

jaykrell:
Formerly identical fork of https://github.com/mono/mono/pull/13794
 to get a new workspace, to try to resolve AOT version error.
 That didn't work, so an attempt is also present to incrementally build nunit to fix the problem.

The larger goal here is to stop hashing JIT icalls, by string and by code address, as suggested a few months ago. Both the inserts into hashtables and the lookups.

Note this is about JIT icalls.
Not icalls declared in C#.

Reducing the lookups is easy, but complete elimination takes more work,
and I believe is possible and desirable.

JIT icalls are used with:
  `MONO_PATCH_INFO_JIT_ICALL` 
  `MONO_PATCH_INFO_JIT_ICALL_ADDR` 
  `MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL` 

There is a fixed set of JIT icalls.
Their strings can be replaced with an enum,
or pointers to fixed data structures, or pointers/offsets
into a fixed data structure.

Except for:
  `MONO_PATCH_INFO_JIT_ICALL_ADDR + sprintf ("trampoline_func_%d", tramp_type)`
  `MONO_PATCH_INFO_JIT_ICALL_ADDR + sprintf ("specific_trampoline_lazy_fetch_%u", slot)`

This change splits those uses into separate patch types, along with
the dynamic integer. Instead of forming the string.

As well, an important theory here is, while these were reusing
some JIT icall infrastructure, they were not using all
of it. These were not actually placed in the hashtables,
and if they were looked up, they were not found.

If that theory is in fact false, then this effort fails.

A separate do-not-merge PR exists to help test that theory (https://github.com/mono/mono/pull/13792
).

It is fairly obvious in the code, in that these JIT icalls
are not passed to any `register` function.

The int vs. uint nuance in this PR not clearly required, but follows
the prior code. They can probably all be int or uint.
As well we might consider guint32 instead of guint, if Mono is ever
to run with a 16bit integer.

### DIFF
--- a/mcs/build/tests.make
+++ b/mcs/build/tests.make
@@ -131,7 +131,8 @@ ifndef NO_TEST
 $(test_nunit_dep): $(topdir)/build/deps/nunit-$(PROFILE).stamp
 	@if test -f $@; then :; else rm -f $<; $(MAKE) $<; fi
 
-$(topdir)/build/deps/nunit-$(PROFILE).stamp:
+# aot-runtime.h due to MONO_AOT_FILE_VERSION.
+$(topdir)/build/deps/nunit-$(PROFILE).stamp: $(topdir)/../mono/mini/aot-runtime.h
 	cd ${topdir}/tools/nunit-lite && $(MAKE)
 	echo "stamp" >$@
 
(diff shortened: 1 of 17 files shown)


jaykrell:
Same Linux amd64 Bitcode error. Hm.

jaykrell:
Could also be a race in the build. Pretty hard to untangle all the makefiles. :(

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

vargaz:
@monojenkins build failed

jaykrell:
This variation no longer needed.

PR #13840, opened 2019-04-04
[netcore] Implement RuntimeImports.RhBulkMoveWithWriteBarrier

filipnavara:
Fixes 15 failures in System.Memory.Tests.

### DIFF
--- a/mono/metadata/icall-def.h
+++ b/mono/metadata/icall-def.h
@@ -933,7 +933,8 @@ HANDLES(REMSER_2, "IsTransparentProxy", ves_icall_IsTransparentProxy, MonoBoolea
 
 ICALL_TYPE(RUNIMPORT, "System.Runtime.RuntimeImports", RUNIMPORT_1)
 #if ENABLE_NETCORE
-NOHANDLES(ICALL(RUNIMPORT_1, "ZeroMemory", ves_icall_System_Runtime_RuntimeImports_ZeroMemory))
+NOHANDLES(ICALL(RUNIMPORT_1, "RhBulkMoveWithWriteBarrier", ves_icall_System_Runtime_RuntimeImports_Memmove))
+NOHANDLES(ICALL(RUNIMPORT_2, "ZeroMemory", ves_icall_System_Runtime_RuntimeImports_ZeroMemory))
 #else
 NOHANDLES(ICALL(RUNIMPORT_1, "Memmove", ves_icall_System_Runtime_RuntimeImports_Memmove))
 NOHANDLES(ICALL(RUNIMPORT_2, "Memmove_wbarrier", ves_icall_System_Runtime_RuntimeImports_Memmove_wbarrier))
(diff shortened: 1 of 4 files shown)


mono/metadata/icall-def.h
```diff
@@ -933,7 +933,8 @@ HANDLES(REMSER_2, "IsTransparentProxy", ves_icall_IsTransparentProxy, MonoBoolea
 
 ICALL_TYPE(RUNIMPORT, "System.Runtime.RuntimeImports", RUNIMPORT_1)
 #if ENABLE_NETCORE
-NOHANDLES(ICALL(RUNIMPORT_1, "ZeroMemory", ves_icall_System_Runtime_RuntimeImports_ZeroMemory))
+NOHANDLES(ICALL(RUNIMPORT_1, "RhBulkMoveWithWriteBarrier", ves_icall_System_Runtime_RuntimeImports_Memmove))
```
EgorBo: Will it insert a memory barrier like in .net core?

filipnavara: <strike>I looked what normal Mono builds do. They end up calling `ves_icall_System_Runtime_RuntimeImports_Memmove_wbarrier`. That follows down two paths:
- For references it goes to `mono_gc_wbarrier_arrayref_copy_internal` which in turn calls `mono_gc_memmove_aligned (dest_ptr, src_ptr, count * sizeof (gpointer));`.
- For value types it goes to `mono_gc_wbarrier_value_copy_internal` which in turn calls `mono_gc_memmove_atomic (dest, src, count * mono_class_value_size (klass, NULL));`

`mono_gc_memmove_atomic` calls `mono_gc_memmove_aligned` if the input parametes are aligned.

That lead me to believe that it will behave the same as normal Mono builds. I do not see an explicit memory barrier in that code path though.</strike>

Looks like I messed up my analysis (copied it above for reference) because I accidentally looked at Boehm GC code which was missing the write barriers. SGen code paths on normal Mono seem to have it correctly.

I'll try to fix it.

filipnavara:
@monojenkins build Linux i386

mono/metadata/icall-def.h
```diff
@@ -933,7 +933,8 @@ HANDLES(REMSER_2, "IsTransparentProxy", ves_icall_IsTransparentProxy, MonoBoolea
 
 ICALL_TYPE(RUNIMPORT, "System.Runtime.RuntimeImports", RUNIMPORT_1)
 #if ENABLE_NETCORE
-NOHANDLES(ICALL(RUNIMPORT_1, "ZeroMemory", ves_icall_System_Runtime_RuntimeImports_ZeroMemory))
+NOHANDLES(ICALL(RUNIMPORT_1, "RhBulkMoveWithWriteBarrier", ves_icall_System_Runtime_RuntimeImports_Memmove_wbarrier))
+NOHANDLES(ICALL(RUNIMPORT_2, "ZeroMemory", ves_icall_System_Runtime_RuntimeImports_ZeroMemory))
```
vargaz: The C function name should match the icall name.

filipnavara:

filipnavara:
@monojenkins build Linux x64

filipnavara:

filipnavara:

Issue #13755, opened 2019-03-31
Roslyn bump prerequisites and related issues


kg:
We recently attempted to bump our roslyn binaries to a recent version and ran into multiple problems that didn't appear in commit-blocking CI results and weren't obvious in local builds at all. Going to try to collect a list of them here - I'm guessing I'm missing some.

* Users run csc and msbuild in various ways, and each method has different behavior when it comes to resolving roslyn dependencies. An incorrect bump may break some of them but not others.
* The dependency tree for roslyn assemblies is invisible in tools like dnSpy because the dependencies do not appear to all be assembly references.
* Roslyn assemblies we use have a target framework of net472, which also makes it non-obvious that they might have new dependencies.
* We ship some facades for roslyn dependencies but it seems as if (?) some of them are not compatible
* Shipping some of the roslyn dependencies included in its DLL will cause issues as they will conflict with our facades and we deploy its DLLs directly with ours.
* msbuild directly relies on roslyn DLLs (and thus roslyn's DLL dependencies) but we appear to in some cases deploy msbuild directly next to our csc, so if we bump roslyn without bumping msbuild, msbuild will break.
* there are multiple ways to compile msbuild and the result does not appear to behave in the same way. this can make it difficult to identify or reproduce issues.
* csc-dim has its binaries deployed in a different fashion from regular csc, so a roslyn build we test as csc-dim may not work correctly once we promote it to csc.
* in general, from-source builds of mono and msbuild do not appear to replicate the user experience of most mono users in regards to roslyn. I install from a prefix and never observed any of the user-facing problems the bump created. (A from-source build also does not have msbuild at all.)

kg:
CCing some people with relevant expertise or information (though i don't think any of these problems are directly or indirectly your responsibility):
@directhex @marek-safar @akoeplinger

borgdylan:
@kg The latest msbuild form xplat-master seems to work with a pre-bump version of mono:
msbuild : 16.0.42-preview+g804bde742b (8b3a694c2c419582108588a7bd1133630a1b8077)
mono :  6.1.0 (master/92b435d6fe7 Tue 12 Mar 18:25:17 CET 2019) (92b435d6fe7c27593f791eb2bc8d77ba4b7fc17d)

steveisok:
I don't know if this is exclusive to VS for Mac / MD, but it probably should be noted along w/ this...

https://github.com/mono/mono/issues/13213

steveisok:
[...]

PR #13850, opened 2019-04-04
INTENTIONALLY BROKEN COMMIT: Verify that CI catches the scenario where we deploy roslyn without necessary dependencies

kg:
Using this PR to validate some followup from #13755 on CI.


kg:

kg:
OK, once we enabled it "prefix sanity checks" caught this.

PR #13852, opened 2019-04-04
Bump macsdk msbuild hash to msbuild that uses new roslyn, matching mono master

kg:
There's a comment in this script that says to 'fix run-test-mac-sdk.sh' but i can't find anything to fix in that script. Is that comment dead?


radical:
@kg That refers to the *version* number in `(self, 'mono', 'msbuild', '15', ` -- `15` here, which doesn't need to be changed.

radical:

radical:

kg:

kg:
@monojenkins build failed

akoeplinger:
@radical shouldn't we change the version number since we're shipping msbuild 16? (I know it's just a cosmetic fix since it only shows up in the Mac installer).

radical:
@akoeplinger Does it show up in the installer? I was under the impression that it is just used for creating staging directories by bockbuild.

PR #13853, opened 2019-04-04
[netcore] Avoid loading the globalization native lib in invariant mode.

vargaz:


EgorBo:
In theory we could port our non-ICU based implementation so .NET could have three modes: ICU, invariant and "mobile". 🙂

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #13855, opened 2019-04-04
[metadata] Fix resource leak, on error, we did not free the named arguments.  Coverity 1443346

migueldeicaza:
[metadata] Fix resource leak, on error, we did not free the named arguments. Coverity 1443346

### DIFF
--- a/mono/metadata/custom-attrs.c
+++ b/mono/metadata/custom-attrs.c
@@ -1013,9 +1013,9 @@ create_custom_attr_into_array (MonoImage *image, MonoMethod *method, const gucha
  * NAMED_ARG_INFO will contain information about the named arguments.
  */
 void
-mono_reflection_create_custom_attr_data_args (MonoImage *image, MonoMethod *method, const guchar *data, guint32 len, MonoArray **typed_args, MonoArray **named_args, CattrNamedArg **named_arg_info, MonoError *error)
+mono_reflection_create_custom_attr_data_args (MonoImage *image, MonoMethod *method, const guchar *data, guint32 len, MonoArray **typed_args_out, MonoArray **named_args_out, CattrNamedArg **named_arg_info, MonoError *error)
 {
-	MonoArray *typedargs, *namedargs;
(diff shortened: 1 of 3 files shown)


mono/metadata/custom-attrs.c
```diff
@@ -1316,6 +1316,7 @@ mono_reflection_create_custom_attr_data_args_noalloc (MonoImage *image, MonoMeth
 fail:
 	mono_error_set_generic_error (error, "System.Reflection", "CustomAttributeFormatException", "Binary format of the specified custom attribute was invalid.");
 	g_free (arginfo);
+	g_free (named_args);
```
akoeplinger: I don't see where we're allocating `named_args` in this function? isn't it the responsibility of the caller?

jaykrell: Presumably it is supposed to be `namedargs` without the underscore.
https://github.com/mono/mono/pull/13855/files#diff-7781be09cd2ddfae798d13ecb23a63c6R1224

I suggest, um, renaming the args. `out_namedargs` and `namedargs`?

jaykrell: And `namedargs` will need null initialization also.

jaykrell: Or put a star on it and ignore the rest I said, `g_free(*named_args);` that should do, but double check.

jaykrell: No the last point is bad. We don't write the out parameter until the very end of success. So the first part -- null initialize the local and add underscore to the free.

CoffeeFlux:
I think it's worth doing some variable naming throughout the file to avoid this kind of ambiguity and to ensure everything is properly NULL-initialized and freed. I'll push to this branch with that work and a fix for the specific issue coverity caught.

CoffeeFlux:
Wasn't totally sure on naming, but other stuff in the file seemed to use underscores so I just switched all the *args to that. Should be easier to follow the logic now.

CoffeeFlux:
@monojenkins build failed

PR #13860, opened 2019-04-04
[debugger-agent]: Print error if the socket can not be created

migueldeicaza:
And remove assert that probed the wrong value, Coverity 1307166


marek-safar:
@thaystg please review

PR #13866, opened 2019-04-04
[mono-dl] Don't do thread state transitions in mono_dl_fallback_register

lambdageek:
1. it doesn't touch the managed runtime

2. it may be called (arguably it ought to be called only) before the runtime is
initialized (and in particular before the current thread is attached).


lambdageek:
This has caused issues in the android designer.  need to backport to 2018-10

lambdageek:

lambdageek:
@monojenkins build failed

PR #13874, opened 2019-04-05
[netcore] Keep the xunit parallelization setting at default value

filipnavara:
Keep the xunit parallelization setting at default value to allow `Xunit.CollectionBehavior(DisableTestParallelization = true)` override to work. Fixes all failures in System.Xml.RW.XmlWriterApi.Tests.


EgorBo:
I think we can drop `-noappdomain -noshadow` as well (appdomains are noop for .net core, noshadow)

PR #13807, opened 2019-04-02
[netcore] System.Runtime.InteropServices test fixes.

vargaz:

PR #13876, opened 2019-04-05
[netcore] Fix Marshal.IsPinnable regression

filipnavara:
Fix regression from https://github.com/mono/mono/pull/13807/files#diff-e881a4d9491559282e7ae2021f8ad3a4R54. 

CoreCLR code for reference: https://github.com/dotnet/coreclr/blob/2f4a7564beb6f8c1f4929504ee2221953d25fc39/src/vm/marshalnative.cpp#L276-L278

Fixes `System.Tests.StringTests.Ctor_CharPtr_DoesNotAccessInvalidPage` in `System.Runtime.Tests`.

/cc @EgorBo

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -5576,6 +5576,8 @@ ves_icall_System_Runtime_InteropServices_Marshal_IsPinnableType (MonoReflectionT
 
 	if (m_class_get_rank (klass)) {
 		MonoClass *eklass = m_class_get_element_class (klass);
+		if (m_class_is_primitive (eklass))
+			return TRUE;		
 		return eklass != mono_defaults.object_class && m_class_is_blittable (eklass);
 	} else
 		return m_class_is_blittable (klass);


EgorBo:
So if I understand correctly IsPinnable should return true for:
- any value type (even if it's blittable) or array of those
- blittable classes - decimal, (more examples?)

filipnavara:
My understanding is that it should return `true` for:
- Any blittable type
- Any array of blittable types
- Any array of primitive types (covers bool[])
- String, Utf8String (handled separately in managed code in Mono)

It should not return `true` for:
- Primitive non-blittable type
- Value types that are not blittable (eg. contain pointers to other GC objects)

vargaz:
Those changes were made to fix some System.Runtime.Interop.Tests failures.

filipnavara:
> Those changes were made to fix some System.Runtime.Interop.Tests failures.

I don't think I broken any tests there by making this change. I'll re-run them when I get home.

filipnavara:
I get 7 failures in `System.Runtime.InteropServices.Tests` both with and without this PR. There's no visible regression in other tests either.

Issue #13891, opened 2019-04-06
MonoAssembly/MonoImage multi-domain (and ALC) correctness


lambdageek:
The goal of this project is to stop sharing `MonoImage` between multiple domains.  Our current design is wrong.

Rght now there's a subtle bug in how Mono images behave when they're loaded in multiple domains (and there will be more problems once we're support `AssemblyLoadContext`):
If an image `A` is loaded in two domains, it is possible for the references from `A` to be resolved in two different ways in the two domains, however Mono stores a single `MonoAssembly*` for each `image->references[i]`.

What we need to do:

### Done:
- [x] Create a new `MonoImageStorage` data structure that will be responsible for the mmap'd data of a loaded assembly. → ~#13892~ reverted; #13942 
   - [x] fixup the win32 `FreeLibrary` on image close that should be done in the storage destructor
   - [x] leave `MonoImage:raw_data` as an alias to the storage raw data - would fix the extensions build, too.

### `AssemblyLoadContext` in netcore:
- [x] Stop sharing `MonoImage`s across domain or ALC boundaries.  Cache `MonoImage` loading in `MonoAssemblyLoadContext` and leave all the raw_data sharing to `MonoImageStorage` #15527
   - [x] Implement default `MonoAssemblyLoadContext` (done in #15527)
   - [x] Implement custom assembly load contexts (partly in #15527, some APIs will need more work.).

### Future / Nice to have:
- [ ] (Mono framework) Some kind of `MonoAssemblyLoadContext` support in non-netcore mono.  This will have compromises because a lot of the existing Mono API is not ALC aware (in particular - many `MonoImage` loading APIs assume a global set of loaded images).

We will do this in several steps to split up the refactoring step (adding MonoImageStorage) from the behavioral change (unsharing MonoImages).

srxqds:
it can support multiple domain or alc in android or ios platfroms when this task complete?

PR #13892, opened 2019-04-06
[image] Create MonoImageStorage to own the image raw data

lambdageek:
Create a new data structure: MonoImageStorage.

It is an object that will have the responsibility for the raw data of a
MonoImage.  It has a string key and a refcount that is used to share a
MonoImageStorage between multiple MonoImage objects.

The reason we need this is because the current MonoAssembly/MonoImage design is
broken for multiple domains and (more evidently) it will be broken when we add
AssemblyLoadContext support.  The issue is that a MonoImage may be shared
between multiple domains (or ALCs), but it has a 'references' field which
points to a single other MonoAssembly.

This is a problem because the references of an image may be resolved
differently in different domains (or ALCs).

The eventual solution will be to stop sharing MonoImages based on path name (or
in the case of in-memory images based on a name made up from the address of the
byte blob).

However we still don't want to open the same data more than once (if we're on a
config where we don't have mmap, we malloc some memory and dump the data in
there - we don't want to do that multiple times for the same file).

So the solution is to create this MonoImageStorage object and make it possible
to share it based on path name, but don't give it any responsibilities except
for owning the memory.  It doesn't know anything about metadata or assemblies
or any of that stuff - it just owns a chunk of memory and knows when and how to
free it.

This commit just adds the MonoImageStorage object and wires it up in MonoImage
loading.  There should be no observable behavioral changes from this commit.

This is the first step of #13891

### DIFF
--- a/mono/metadata/image.c
+++ b/mono/metadata/image.c
@@ -113,8 +113,17 @@ static gboolean debug_assembly_unload = FALSE;
 
 #define mono_images_lock() if (mutex_inited) mono_os_mutex_lock (&images_mutex)
 #define mono_images_unlock() if (mutex_inited) mono_os_mutex_unlock (&images_mutex)
+#define mono_images_storage_lock() do { if (mutex_inited) mono_os_mutex_lock (&images_storage_mutex); } while (0)
+#define mono_images_storage_unlock() do { if (mutex_inited) mono_os_mutex_unlock (&images_storage_mutex); } while (0)
 static gboolean mutex_inited;
 static mono_mutex_t images_mutex;
+static mono_mutex_t images_storage_mutex;
+
+/* Maps string keys to MonoImageStorage values.
+ *
+ * The MonoImageStorage in the hash owns the key.
+ */
+static GHashTable *images_storage_hash;
 
 static void install_pe_loader (void);
 
@@ -177,7 +186,7 @@ mono_install_image_loader (const MonoImageLoader *loader)
 	image_loaders = g_slist_prepend (image_loaders, (MonoImageLoader*)loader);
 }
 
-/* returns offset relative to image->raw_data */
+/* returns offset relative to m_image_get_raw_data (image) */
 guint32
 mono_cli_rva_image_map (MonoImage *image, guint32 addr)
 {
@@ -193,7 +202,7 @@ mono_cli_rva_image_map (MonoImage *image, guint32 addr)
 		if ((addr >= tables->st_virtual_address) &&
 		    (addr < tables->st_virtual_address + tables->st_raw_data_size)){
 #ifdef HOST_WIN32
-			if (image->is_module_handle)
+			if (m_image_is_module_handle (image))
 				return addr;
 #endif
 			return addr - tables->st_virtual_address + tables->st_raw_data_ptr;
@@ -223,9 +232,9 @@ mono_image_rva_map (MonoImage *image, guint32 addr)
 	int i;
 
 #ifdef HOST_WIN32
-	if (image->is_module_handle) {
-		if (addr && addr < image->raw_data_len)
-			return image->raw_data + addr;
+	if (m_image_is_module_handle (image)) {
+		if (addr && addr < m_image_get_raw_data_len (image))
+			return m_image_get_raw_data (image) + addr;
 		else
 			return NULL;
 	}
@@ -254,7 +263,10 @@ mono_image_rva_map (MonoImage *image, guint32 addr)
 void
 mono_images_init (void)
 {
+	mono_os_mutex_init (&images_storage_mutex);
 	mono_os_mutex_init_recursive (&images_mutex);
+
+	images_storage_hash = g_hash_table_new (g_str_hash, g_str_equal);
 
 	int hash_idx;
 	for(hash_idx = 0; hash_idx < IMAGES_HASH_COUNT; hash_idx++)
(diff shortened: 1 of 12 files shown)


mono/metadata/image.c
```diff
@@ -254,8 +263,11 @@ mono_image_rva_map (MonoImage *image, guint32 addr)
 void
 mono_images_init (void)
 {
+	mono_os_mutex_init (&images_storage_mutex);
 	mono_os_mutex_init_recursive (&images_mutex);
 
+	images_storage_hash = g_hash_table_new (g_str_hash, g_str_equal);
```
jaykrell: Files can change under a running program? Path is not identity?

mono/metadata/image.c
```diff
@@ -1417,14 +1433,98 @@ do_mono_image_load (MonoImage *image, MonoImageOpenStatus *status,
...
+static void
+mono_image_storage_close (MonoImageStorage *storage)
+{
+	mono_refcount_dec (storage);
+}
+
+static MonoImageStorage *
+mono_image_storage_open (const char *fname)
+{
+	char *key = NULL;
```
jaykrell: nit: Can be combined with next line.

mono/metadata/image.c
```diff
@@ -1687,21 +1840,46 @@ mono_image_open_from_data (char *data, guint32 data_len, gboolean need_copy, Mon
...
 
 #ifdef HOST_WIN32
+static MonoImageStorage *
+mono_image_storage_open_from_module_handle (HMODULE module_handle, const char *fname, gboolean has_entry_point)
+{
+	char *key = g_strdup (fname);
+	MonoImageStorage *published_storage = NULL;
+	if (mono_image_storage_tryaddref (key, &published_storage)) {
+		g_free (key);
+		return published_storage;
```
jaykrell: nit: Would be nice to avoid `g_strdup` if already present.

mono/metadata/metadata-internals.h
```diff
@@ -306,16 +307,24 @@ typedef struct {
...
-	/*
-	 * This count is incremented during these situations:
-	 *   - An assembly references this MonoImage though its 'image' field
-	 *   - This MonoImage is present in the 'files' field of an image
-	 *   - This MonoImage is present in the 'modules' field of an image
-	 *   - A thread is holding a temporary reference to this MonoImage between
-	 *     calls to mono_image_open and mono_image_close ()
-	 */
-	int   ref_count;
+/* Represents the physical bytes (usually on disk, but could be in memory) for
```
jaykrell: "file system". "disks" are scarse.

mono/metadata/reflection.c
```diff
@@ -1260,7 +1260,7 @@ method_body_object_construct (MonoDomain *domain, MonoClass *unused_class, MonoM
 	if ((method->flags & METHOD_ATTRIBUTE_PINVOKE_IMPL) ||
 		(method->flags & METHOD_ATTRIBUTE_ABSTRACT) ||
 	    (method->iflags & METHOD_IMPL_ATTRIBUTE_INTERNAL_CALL) ||
-	    (image->raw_data && image->raw_data [1] != 'Z') ||
+	    (!image_is_dynamic (image) && m_image_get_raw_data (image) && m_image_get_raw_data (image) [1] != 'Z') ||
```
jaykrell: Was there supposed to be a check for `M` here too? I realize it wasn't there prior.

mono/metadata/metadata-internals.h
```diff
@@ -332,6 +341,20 @@ struct _MonoImage {
 	/* Module entry point is _CorDllMain. */
 	guint8 has_entry_point : 1;
 #endif
+} MonoImageStorage;
+
+struct _MonoImage {
+	/*
+	 * This count is incremented during these situations:
+	 *   - An assembly references this MonoImage though its 'image' field
```
jaykrell: nit: `through`

vargaz:
Keeping image->image_data etc. and initializing it from MonoImageStorage would reduce the diff.

vargaz:
Looks good.

mono/metadata/image.c
```diff
@@ -1435,24 +1535,79 @@ do_mono_image_open (const char *fname, MonoImageOpenStatus *status,
...
+		storage = other_storage;
+	}
+	return storage;
+}
+
+static MonoImageStorage *
+mono_image_storage_new_raw_data (char *datac, guint32 data_len, gboolean raw_data_allocated, const char *name)
+{
+	char *key = (name == NULL) ? g_strdup_printf ("data-%p", datac) : g_strdup (name);
+	MonoImageStorage *published_storage = NULL;
```
vargaz: Note that key might alias with a filename. So it might be better to use file:<filename> as the key for files.

jaykrell:
@monojenkins build failed

mono/metadata/image.c
```diff
@@ -2133,41 +2311,34 @@ mono_image_close_except_pools (MonoImage *image)
...
+				((char*)(ii->cli_sections [i]) <= ((char*)m_image_get_raw_data (image) + m_image_get_raw_data_len (image))))
 				ii->cli_sections [i] = NULL;
 
-		g_free (image->raw_data);
 	}
 
+#ifdef HOST_WIN32
+	mono_images_lock ();
+	if (m_image_is_module_handle (image) && !m_image_has_entry_point (image))
+		FreeLibrary ((HMODULE) m_image_get_raw_data (image));
```
lambdageek: Hm. Pretty sure I need to move this to the `MonoImageStorage` destructor.  Don't want to  call `FreeLibrary` if it's still shared with another image.

vargaz:
Merged by mistake, should we revert, or fix it ?

akoeplinger:
Reverted with https://github.com/mono/mono/pull/13916 as it also broke the mono-extensions build

lambdageek:
Thanks. I'll fix it up and open a new PR
1. move the `FreeLibrary` stuff to run in the storage destructor
2. Add back `MonoImage:raw_data` as an alias to `MonoImageStorage:raw_data` to make the diff smaller and to fixup the extensions build

PR #13893, opened 2019-04-06
Remove obsolete AOT relocations for code clarity, assuming no need to support old AOT images.

jaykrell:
It does not matter perf-wise, but makes the code clearer.
i.e. if attempting to replace strings with enums.


jaykrell:
Note that in the event the strings do appear in images, there is a catch all at the bottom that fails. So we don't need to assert or error where I removed the code.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Merge?

PR #13896, opened 2019-04-06
[netcore] Propagate System.Globalization.Invariant from runtimeconfig.json

filipnavara:
Propagate System.Globalization.Invariant from runtimeconfig.json to GlobalizationMode. It's temporary hack to make all Invariant.Tests pass. CoreCLR also uses special `CLRConfig` class and icall to access this property because it needs it very early, so this may end up quite ok way to accomplish the same.


vargaz:
This should probably check the env variable to avoid overwriting it so the user can still overwrite the value used by the config file.

filipnavara:
> This should probably check the env variable to avoid overwriting it so the user can still overwrite the value used by the config file.

That was my first thought too, but CoreCLR takes precedence with the runtimeconfig.json file, so I matched that behaviour.

PR #13897, opened 2019-04-07
[wasm] Generate the pinvoke tables.

vargaz:
Put the default non-linked version of the table into version control, usable by the interpreter only build.


sdks/wasm/pinvoke-tables-default.h
```diff
@@ -0,0 +1,58 @@
+typedef struct {
```
jaykrell: nit i.e. ignorable: Please add a comment saying this is generated, if it is, as I suspect. The generator should output the comment.

mcs/tools/wasm-tuner/tuner.cs
```diff
@@ -76,12 +89,94 @@ public class WasmTuner
 		}
 	}
 
+	void Usage () {
+		Console.WriteLine ("Usage: tuner.exe <arguments>");
+		Console.WriteLine ("Arguments:");
+		Console.WriteLine ("--gen-icall-table icall-table.json <assemblies>.");
+		Console.WriteLine ("--gen-pinvoke-table <list of native library names separated by commas> <assemblies>.");
+	}
```
kjpou1: The `Usage:` seems to suggest that one can specify both `--gen-icall-table` and `--gen-pinvoke-table` arguments but below it it seems to to only process `--gen-icall-table` or `--gen-pinvoke-table`.  Is that correct?

vargaz: Currently yes, later we want both of them so it only needs to run once.

vargaz:
@monojenkins build failed

sdks/wasm/pinvoke-tables-default.h
```diff
@@ -0,0 +1,59 @@
...
+{"SystemNative_Rename", SystemNative_Rename},
+{"SystemNative_RmDir", SystemNative_RmDir},
+{"SystemNative_Stat2", SystemNative_Stat2},
+{"SystemNative_LStat2", SystemNative_LStat2},
+{"SystemNative_UTime", SystemNative_UTime},
+{"SystemNative_UTimes", SystemNative_UTimes},
+{"SystemNative_Unlink", SystemNative_Unlink},
+{NULL, NULL}
+};
+static void *pinvoke_tables[] = { System_Native_imports,};
```
jaykrell: nit (again, ignorable), you already know: a type other than `void*` would be nice

sdks/wasm/packager.cs
```diff
@@ -721,6 +721,10 @@ class WasmOptions {
 			ninja.WriteLine ($"build $builddir/driver.c: cpifdiff {source_file}");
 			ninja.WriteLine ($"build $builddir/driver-gen.c: cpifdiff $builddir/driver-gen.c.in");
 
+			var pinvoke_file = Path.GetFullPath (Path.Combine (tool_prefix, "pinvoke-tables-default.h"));
+			ninja.WriteLine ($"build $builddir/pinvoke-tables-default.h: cpifdiff {pinvoke_file}");
+			driver_deps += $" $builddir/pinvoke-tables-default.h";
```
jaykrell: nit, ignorable: first dollar sign is presumably nop

mcs/tools/wasm-tuner/tuner.cs
```diff
@@ -76,12 +89,94 @@ public class WasmTuner
...
+			string symbol = module.Replace (".", "_") + "_imports";
+			Console.WriteLine ("static PinvokeImport " + symbol + " [] = {");
+			foreach (var pinvoke in pinvokes) {
+				if (pinvoke.Module == module)
+					Console.WriteLine ("{\"" + pinvoke.EntryPoint + "\", " + pinvoke.EntryPoint + "},");
+			}
+			Console.WriteLine ("{NULL, NULL}");
+			Console.WriteLine ("};");
+		}
+		Console.Write ("static void *pinvoke_tables[] = { ");
```
jaykrell: Nit, ignorable, you know, non-`void*` preferable, and `const`. `const` provides efficiency in most C/C++ but maybe not in wasm.

vargaz:
@monojenkins build failed

PR #13898, opened 2019-04-07
[netcore] Remove an assertion which is no longer needed.

vargaz:


vargaz:
@monojenkins build failed

PR #13901, opened 2019-04-07
[netcore] Allow reflection invocation of methods with ByRef return types

filipnavara:
CoreCLR allows reflection on methods with ByRef return types under certain conditions. The PR implementing it is https://github.com/dotnet/coreclr/pull/17639.

Mono already implemented most of it in the marshalling code but artificially blocked it in `ves_icall_InternalInvoke`. I changed the condition for the exception to match the condition in `emit_invoke_call` (marshal-ilgen.c). It's similar to the condition in CoreCLR, but it may not handle byref-like types correctly. It would have to fixed on both places if it turns out to be the case.

The remaining issue is that null references returned in ByRef are not checked and converted to correct exception. It should result in `NullReferenceException` with special message that doesn't get wrapped in `TargetInvocationException`. I put a TODO marker at a place where I think it should be handled, but it's a bit over my head to implement it.

/cc @vargaz

### DIFF
--- a/mcs/class/corlib/System.Reflection/RuntimePropertyInfo.cs
+++ b/mcs/class/corlib/System.Reflection/RuntimePropertyInfo.cs
@@ -377,7 +377,7 @@ namespace System.Reflection {
 					MethodInfo method = GetGetMethod (true);
 					if (method == null)
 						throw new ArgumentException ($"Get Method not found for '{Name}'");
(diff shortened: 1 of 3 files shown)


filipnavara:
@monojenkins build Linux AArch64 FullAOT+Interpreter+LLVM

filipnavara:

marek-safar:
Thanks!

PR #13906, opened 2019-04-07
[jit] Disable inlining for methods with [System.Security.DynamicSecurityMethod], its used to make methods which contain a StackCrawlMark.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #13914, opened 2019-04-08
[netcore] Fix marshalling of ref ptr return types

filipnavara:


mono/metadata/object.c
```diff
@@ -5566,12 +5566,22 @@ mono_runtime_try_invoke_array (MonoMethod *method, void *obj, MonoArray *params,
...
+				tmpret->byref = FALSE;
+				box_args [1] = mono_type_get_object_checked (mono_domain_get (), tmpret, error);
+				mono_metadata_free_type (tmpret);
+			} else {
+				box_args [1] = mono_type_get_object_checked (mono_domain_get (), sig->ret, error);
+			}
 			return_val_if_nok (error, NULL);
 
 			res = mono_runtime_try_invoke (box_method, NULL, box_args, &box_exc, error);
-			g_assert (box_exc == NULL);
```
vargaz: Why is this needed ?

filipnavara: It's not, but it was easier to debug a managed exception than a runtime assert with no message. I can revert this part of necessary.

vargaz: I think we assert because that method is not supposed to fail.

filipnavara: It fails only when something is internally inconsistent. I'll revert this one back to the assert.

vargaz:
@monojenkins build failed

filipnavara:
@monojenkins build Linux AArch64 FullAOT+Interpreter

PR #13915, opened 2019-04-08
[llvm] If a true/false branch of a conditional branch is marked as out-of-line, emit an llvm.expect call.

vargaz:


mono/mini/mini-llvm.c
```diff
@@ -4971,6 +4971,14 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 			CompRelation rel;
 			LLVMValueRef cmp, args [16];
 			gboolean likely = (ins->flags & MONO_INST_LIKELY) != 0;
+			gboolean unlikely = FALSE;
+
+			if (MONO_IS_COND_BRANCH_OP (ins->next)) {
```
lewurm: I think we should add `&& !likely`

lateralusX:
This will be valuable when inline our direct call wrappers, getting all exception handling out of line will make it easier to branch predict and pre-fetch.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #13917, opened 2019-04-08
[netcore] Do not crash the runtime in System.Reflection.Emit.ILGeneration tests

filipnavara:


vargaz:
I think we should exclude the crashing tests instead.

filipnavara:
> I think we should exclude the crashing tests instead.

They are excluded at the moment, but it's quite hard to debug them. Some of the inputs work properly, but not all.

The two functions on `MonoDomain` are called by the runtime. A proper fix would be to implement the callbacks on `AssemblyLoadContext`, but having something non-crashing makes it easier to estimate what is still missing/broken.

I prefer not to crash the runtime if the error can be communicated back through reasonable means. This is user input, not some internal inconsistency.

vargaz:
@monojenkins build failed

filipnavara:

filipnavara:

PR #13928, opened 2019-04-08
Use TLS_KEY_LMF_ADDR on Windows when pusing lmf frame.

lateralusX:
Use TLS_KEY_LMF_ADDR on Windows as done on all other platforms. No need to go through the struct hosted at TLS_KEY_JIT_TLS since the same address is already set in TLS_KEY_LMF_ADDR.


lateralusX:
@monojenkins rebuild failed

PR #13931, opened 2019-04-08
Add support for direct icalls using external functions not in icall-def.h.

lateralusX:
An icall declaration can now use the following attribute:

MonoDirectICallSymbolNameAttribute("MyNativeCFunction")

and if not already defined in icall-def.h and direct-icalls are used when AOT compile the assembly, a direct call will use symbol defined in attribute as the symbol for direct call. Using this attribute will also enable inlining for the specific icall wrapper generating inlined direct icalls to external defined native methods.

This feature is valuable when embedding the runtime and extend with icalls added using mono_add_internal_call. By applying the above attribute, it is possible to do AOT with direct-icalls and get static linked icalls without the need to extend and rebuild cross compiler in the process.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -5698,11 +5698,10 @@ gboolean
 mono_aot_direct_icalls_enabled_for_method (MonoCompile *cfg, MonoMethod *method)
 {
 	gboolean enable_icall = FALSE;
-	if (cfg->compile_aot) {
+	if (cfg->compile_aot)
 		enable_icall = lookup_external_icall_symbol_name_aot (method) ? TRUE : FALSE;
-	} else {
+	else
 		enable_icall = FALSE;
-	}
 
 	return enable_icall;
 }
@@ -13824,6 +13823,13 @@ mono_aot_is_shared_got_offset (int offset)
 
 int
 mono_compile_deferred_assemblies (guint32 opts, const char *aot_options, gpointer **aot_state)
+{
+	g_assert_not_reached ();
+	return 0;
+}
+
+gboolean
+mono_aot_direct_icalls_enabled_for_method (MonoCompile *cfg, MonoMethod *method)
 {
 	g_assert_not_reached ();
 	return 0;
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -289,6 +289,7 @@ typedef struct MonoAotCompile {
 	GHashTable *method_to_cfg;
 	GHashTable *token_info_hash;
 	GHashTable *method_to_pinvoke_import;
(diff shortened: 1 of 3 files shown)


mono/mini/aot-compiler.c
```diff
@@ -401,6 +402,11 @@ typedef struct {
 /* This points to the current acfg in LLVM mode */
 static MonoAotCompile *llvm_acfg;
 
+/* Cache of decodeded method external icall symbol names. */
```
jaykrell: s/decodeded/decoded/

lateralusX:
@monojenkins rebuild failed

lateralusX:
@vargaz Approve/Merge?

mono/mini/aot-compiler.c
```diff
@@ -12423,6 +12503,7 @@ acfg_create (MonoAssembly *ass, guint32 opts)
 	acfg->method_to_cfg = g_hash_table_new (NULL, NULL);
 	acfg->token_info_hash = g_hash_table_new_full (NULL, NULL, NULL, NULL);
 	acfg->method_to_pinvoke_import = g_hash_table_new_full (NULL, NULL, NULL, g_free);
+	acfg->method_to_external_icall_symbol_name = g_hash_table_new_full (NULL, NULL, NULL, g_free);
 	acfg->image_hash = g_hash_table_new (NULL, NULL);
```
vargaz: This should simply be a global hash then.

lateralusX: used the acfg to simplify ownership of the hash (alloc/free), but will be cached in global static during lifetime of acfg (similar to how we do with the llvm acfg), but agree, we could just move ownership out of acfg and put it in global hash, I guess we still want lifetime to be connected with the lifetime of acfg (or at least clean it)?

mono/mini/aot-compiler.c
```diff
@@ -5627,6 +5633,80 @@ add_generic_instances (MonoAotCompile *acfg)
...
+	return symbol_name;
+}
+
+gboolean
+mono_aot_direct_icalls_enabled_for_method (MonoCompile *cfg, MonoMethod *method)
+{
+	gboolean enable_icall = FALSE;
+	if (cfg->compile_aot) {
+		enable_icall = lookup_external_icall_symbol_name_aot (method) ? TRUE : FALSE;
+	} else {
```
vargaz: Extra {}

vargaz: This also needs a stub in the AOT disabled section at the end.

mono/mini/aot-compiler.c
lateralusX: Decided to keep current construction for now.

lateralusX:
@monojenkins rebuild failed

PR #13934, opened 2019-04-08
[netcore] Runtime nupkg fixes

directhex:
I know everyone hated the move to Autotools, but there was method to my madness.

We can now build `--with-core=only` on Linux or Mac with the same makefile, and have a functional `make nupkg` rule which attempts to produce a .nupkg of the runtime in the format consumed by core-setup. We should at least have something we can throw at CI to get binaries produced, and talk to the Core team about consuming them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -159,12 +159,15 @@ case "$host" in
 		fi
 		HOST_CC="gcc"
 		RID="win-x86"
+		CORETARGETS="/p:TargetsWindows=true"
+		COREARCH="x86"
 		# Boehm not supported on 64-bit Windows.
 		case "$host" in
 		x86_64-*-* | amd64-*-*)
 			support_boehm=no
 			with_gc=sgen
 			RID="win-x64"
+			COREARCH="x64"
 			;;
 		esac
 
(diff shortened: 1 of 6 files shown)


filipnavara:
The BCL builds still need to be fixed up on Linux/Windows to pass correct `Targets*` properties. I am slowly going through the Windows build to fix it up.

filipnavara:
I fixed up the Windows builds locally, but I run into more issues that need to be solved properly in the build system:
- BCL on Linux has to be built with /p:TargetsUnix=true and no /p:TargetsOSX=true
- BCL on Windows has to be buiilt with /p:TargetsWindows=true
- The .NET Runtime packages for Windows are packaged as .zip, not .tar.gz. That needs to be handled for the downloads and in the extraction step (tar -xf vs unzip)

Additionally, I run into few more issues on my Windows machine that may or may not be related to my build environment:
- I didn't have unzip installed in Cygwin to fix one of the steps above. This is probably just missing in my local installation, but should be mentioned in build step.
- I didn't have NuGet executable on the machine aside from one copy from old Mono. Not sure how to solve it, but options are a) using the built-in NuGet in `dotnet` b) downloading the one package directly using curl and extracting it with `unzip`.
- The runtime had a dependency on `libgcc_s_seh-1.dll` which didn't get copied over. It would be nice to track down where this dependency comes from and if it could be avoided.

directhex:
There's also the unresolved issue that all the test-related targets in netcore/Makefile are hardcoded to OSX. And it looks like there are upper/lowercase issues to make that a pain to fix.

filipnavara:
> There's also the unresolved issue that all the test-related targets in netcore/Makefile are hardcoded to OSX. 

True. I am waiting on some actionable decision on how the test assemblies should actually be downloaded and executed correctly. It's one big hack now and I refuse to slap workarounds on top of workarounds any further.

PR #13935, opened 2019-04-08
[System.Data] Remove unnecessary file from xunit tests

akoeplinger:
It contained DllImports to some Windows APIs which causes issues on xamarin-macios which expects all pinvokes to exist.

Turns out we don't even need that file.


akoeplinger:
Cherry-picked down to 2018-10.

PR #13942, opened 2019-04-09
[image] Create MonoImageStorage to own the image raw data

lambdageek:
Second try.

Summary compared to previous reverted PR:
1. `MonoImage:raw_data` is an alias for `MonoImageStorage:raw_data` so that we don't have to update all the places in the code that wrote `image->raw_data`.
2. Call `FreeLibrary` from `mono_image_storage_dtor` instead of from `mono_image_close_except_pools` - if the `MonoImageStorage` is shared between multiple `MonoImage`s, only call `FreeLibrary` when the last image is closed, not every time.  (Right now we don't share `MonoImageStorage`, that's part 2 of #13891)

---- 
Original summary

Create a new data structure: MonoImageStorage.

It is an object that will have the responsibility for the raw data of a
MonoImage.  It has a string key and a refcount that is used to share a
MonoImageStorage between multiple MonoImage objects.

The reason we need this is because the current MonoAssembly/MonoImage design is
broken for multiple domains and (more evidently) it will be broken when we add
AssemblyLoadContext support.  The issue is that a MonoImage may be shared
between multiple domains (or ALCs), but it has a 'references' field which
points to a single other MonoAssembly.

This is a problem because the references of an image may be resolved
differently in different domains (or ALCs).

The eventual solution will be to stop sharing MonoImages based on path name (or
in the case of in-memory images based on a name made up from the address of the
byte blob).

However we still don't want to open the same data more than once (if we're on a
config where we don't have mmap, we malloc some memory and dump the data in
there - we don't want to do that multiple times for the same file).

So the solution is to create this MonoImageStorage object and make it possible
to share it based on path name, but don't give it any responsibilities except
for owning the memory.  It doesn't know anything about metadata or assemblies
or any of that stuff - it just owns a chunk of memory and knows when and how to
free it.

This commit just adds the MonoImageStorage object and wires it up in MonoImage
loading.  There should be no observable behavioral changes from this commit.

This is the first step of #13891

### DIFF
--- a/mono/metadata/image.c
+++ b/mono/metadata/image.c
@@ -113,8 +113,17 @@ static gboolean debug_assembly_unload = FALSE;
 
 #define mono_images_lock() if (mutex_inited) mono_os_mutex_lock (&images_mutex)
 #define mono_images_unlock() if (mutex_inited) mono_os_mutex_unlock (&images_mutex)
+#define mono_images_storage_lock() do { if (mutex_inited) mono_os_mutex_lock (&images_storage_mutex); } while (0)
+#define mono_images_storage_unlock() do { if (mutex_inited) mono_os_mutex_unlock (&images_storage_mutex); } while (0)
 static gboolean mutex_inited;
 static mono_mutex_t images_mutex;
+static mono_mutex_t images_storage_mutex;
+
+/* Maps string keys to MonoImageStorage values.
+ *
+ * The MonoImageStorage in the hash owns the key.
+ */
+static GHashTable *images_storage_hash;
 
 static void install_pe_loader (void);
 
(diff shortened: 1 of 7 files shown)


mono/metadata/image.c
```diff
@@ -1435,24 +1553,80 @@ do_mono_image_open (const char *fname, MonoImageOpenStatus *status,
...
+		mono_image_storage_close (storage);
+		storage = other_storage;
+	}
+	return storage;
+}
+
+static MonoImageStorage *
+mono_image_storage_new_raw_data (char *datac, guint32 data_len, gboolean raw_data_allocated, const char *name)
+{
+	char *key = (name == NULL) ? g_strdup_printf ("data-%p", datac) : g_strdup (name);
```
lambdageek: @vargaz So on the last PR you suggested that we should add a prefix on `name` here so that it won't collide with the normal file based opens.  But I think we actually do want it to collide.

The way this function gets called with a non-NULL name is from `mono_assembly_open_from_bundle` in which case I think we want it all other attempts to load the same assembly to match.

lambdageek:
@monojenkins build pkg

marek-safar:
@vargaz could you please review this change

lambdageek:
@monojenkins build failed

PR #13943, opened 2019-04-09
[netcore] Make System.Private.CoreLib and host APIs compilable on Windows

filipnavara:
Fixes CLR hosting API and BCL code on Windows enough to run "Hello World". It's still not integrated correctly into the build system, https://github.com/mono/mono/pull/13934 fixes some of the issues, but not all.


mcs/class/System.Private.CoreLib/System/DateTime.Windows.cs
```diff
@@ -0,0 +1,188 @@
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace System
+{
+	public readonly partial struct DateTime
```
marek-safar: Please move this to shared instead (together with Interop changes)

mcs/class/System.Private.CoreLib/System.Globalization/GlobalizationMode.Unix.cs
```diff
@@ -0,0 +1,29 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Globalization
+{
+	internal static partial class GlobalizationMode
```
marek-safar: We try not to repeat the "main" type modifiers, so `partial class GlobalizationMode` should be fine

mcs/class/System.Private.CoreLib/System/DateTime.Windows.cs
filipnavara: I already tried that once. It was shot down for performance reasons (internal calls vs. pinvokes).

filipnavara: https://github.com/dotnet/coreclr/pull/21848

mcs/class/System.Private.CoreLib/System.Globalization/GlobalizationMode.Unix.cs
filipnavara: Ok, will update that.

mcs/class/System.Private.CoreLib/System/DateTime.Windows.cs
filipnavara: I can give it one more try and use `!CORECLR` like DateTime.Unix.cs ended-up.

filipnavara:
I've rebased this PR now that all changes were propagated through the CoreLib mirror. Now majority of the Windows specific code is in the shared partition.

filipnavara:
Build system is still broken for several reasons:
- NuGet command line tool is not distributed on Windows (fixed by #13989)
- `make prepare` expects the .NET Core runtime package to be packed by .tar.gz. It's easy to fix to account for it by using `.zip` and `unzip`
- Copying the Mono runtime over CoreCLR runtime doesn't work because on Windows the `lib` prefix is not used

PR #13946, opened 2019-04-09
Merge icall_hash and icall_hash_foreign into one.

lateralusX:
icalls registered with mono_add_internal_call, mono_add_internal_call_with_flags or mono_add_internal_call_internal ended up in two different hash tables depending if they were "foreign" or not. This caused double lookups in mono_lookup_internal_call_full. There are also use cases where it could be interesting to keep additional flags related to the icall, that could be used by codegen for various optimization.

This PR merge the two hash tables into one and add a flag describing icall details. The flag could later be extended with additional behaviors going forward if needed.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -8526,9 +8526,13 @@ ves_icall_System_IO_LogcatTextWriter_Log (const char *appname, gint32 level, con
 static MonoIcallTableCallbacks icall_table;
 static mono_mutex_t icall_mutex;
 static GHashTable *icall_hash = NULL;
-static GHashTable *icall_hash_foreign = NULL;
 static GHashTable *jit_icall_hash_name = NULL;
 static GHashTable *jit_icall_hash_addr = NULL;
+
+typedef struct _MonoIcallHashTableValue {
+	gconstpointer method;
+	guint32 flags;
+} MonoIcallHashTableValue;
 
 void
 mono_install_icall_table_callbacks (MonoIcallTableCallbacks *cb)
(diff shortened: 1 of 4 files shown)


mono/metadata/icall.c
```diff
@@ -8564,12 +8567,26 @@ void
...
 	g_hash_table_destroy (jit_icall_hash_addr);
 	mono_os_mutex_destroy (&icall_mutex);
 }
 
+static void
+add_internal_call_with_flags (const char *name, gconstpointer method, guint32 flags)
+{
+	mono_icall_lock ();
+
+	MonoIcallHashTableValue *value = g_new (MonoIcallHashTableValue, 1);
```
jaykrell: `g_new` should be before `mono_icall_lock`, to reduce lock hold times.

jaykrell: `g_strdup` should also be outside of lock.

mono/metadata/icall.c
```diff
@@ -8526,10 +8526,14 @@ ves_icall_System_IO_LogcatTextWriter_Log (const char *appname, gint32 level, con
 static MonoIcallTableCallbacks icall_table;
 static mono_mutex_t icall_mutex;
 static GHashTable *icall_hash = NULL;
-static GHashTable *icall_hash_foreign = NULL;
 static GHashTable *jit_icall_hash_name = NULL;
 static GHashTable *jit_icall_hash_addr = NULL;
 
+typedef struct _MonoIcallHashTableValue {
+	gconstpointer method;
```
jaykrell: If `const` is only going to be frequently cast away, consider removing it here.

lateralusX: hopefully it won't be needed at more places since any new usage should use gconstpointer, reason that it needs to be casted away was an old API, but that doesn't seem to be exposed so maybe I can keep it const a little longer. Will update.

mono/metadata/icall.c
lateralusX: Yes, the g_strdup was in the lock in old solution as well, but moving both out should definetly be done to reduce lock times. Will update.

PR #13948, opened 2019-04-09
TransactionScope now actually uses IsolationLevel that was specified …

### LINKED ISSUE #13945: TransactionScope ignores specified IsolationLevel and causes Serializable transactions to be used
TransactionScope accepts TransactionOptions but ignores them. This causes transactions to use Serializable isolation level (which is the default) even though ReadCommited was specified.
This was first reported in https://bugzilla.xamarin.com/show_bug.cgi?id=28053

It boils down to this code:
```
using (var tx = new TransactionScope(TransactionScopeOption.Required, new TransactionOptions {IsolationLevel = IsolationLevel.ReadCommitted}))
{
  // tx is actually Serializable, not ReadCommitted
}
```
## Steps to Reproduce

This is a sample program that uses transaction scope to open database transaction.

1. Download source code from https://gist.github.com/dmitryvk/6623484f7faff7452be2f5b5cfe2af35
2. Compile source code (`msbuild /t:Restore /t:Build`)
3. Execute. E.g.,
    `mono bin/Debug/net472/MonoTxScope.exe`
4. This program will output isolation level of a transaction that was actually used.

### Current Behavior

Program prints `Serializable`

### Expected Behavior

Program should print `ReadCommitted` (as it does when using .NET on Windows).

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.18.0.240 (tarball Thu Jan 17 08:37:29 UTC 2019)
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug 
        Interpreter:   yes
        LLVM:          yes(600)
        Suspend:       preemptive

```

dmitryvk:
…in TransactionOptions

Previously, using TransactionScope would always produce transactions
that use serializable isolation level even when user specified another
level (like ReadCommited).

fixes #13945

### DIFF
--- a/mcs/class/System.Transactions/System.Transactions/Transaction.cs
+++ b/mcs/class/System.Transactions/System.Transactions/Transaction.cs
@@ -69,10 +69,10 @@ namespace System.Transactions
 
 		internal IPromotableSinglePhaseNotification Pspe { get { return pspe; } }
 
-		internal Transaction ()
+		internal Transaction (IsolationLevel isolationLevel)
 		{
 			info = new TransactionInformation ();
-			level = IsolationLevel.Serializable;
+			level = isolationLevel;
 		}
 
 		internal Transaction (Transaction other)
--- a/mcs/class/System.Transactions/System.Transactions/CommittableTransaction.cs
+++ b/mcs/class/System.Transactions/System.Transactions/CommittableTransaction.cs
@@ -31,12 +31,14 @@ namespace System.Transactions
 		}
 
 		public CommittableTransaction (TimeSpan timeout)
+			: base (IsolationLevel.Serializable)
 		{
 			options = new TransactionOptions ();
 			options.Timeout = timeout;
 		}
 
 		public CommittableTransaction (TransactionOptions options)
+			: base (options.IsolationLevel)
 		{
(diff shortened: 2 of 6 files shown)


marek-safar:
build

mcs/class/System.Transactions/System.Transactions/Transaction.cs
```diff
@@ -69,10 +69,10 @@ right now
 
 		internal IPromotableSinglePhaseNotification Pspe { get { return pspe; } }
 
-		internal Transaction ()
+		internal Transaction (TransactionOptions options)
```
marek-safar: I think it'd be better to only pass IsolationLevel

dmitryvk: Yes, it's better this way

dmitryvk:
I made suggested change and rebased onto current master.

marek-safar:
approve

steveisok:
@monojenkins build failed

lewing:
@monojenkins build failed

PR #13949, opened 2019-04-09
Use 	DNS.GetHostAddressesAsync instead of sync version

### LINKED ISSUE #13163: Task: Use async DNS resolution in Socket and web stack code.
https://github.com/mono/mono/blob/b936c531018a354c92103cc9b7cf401231063725/mcs/class/System/System.Net.Sockets/Socket.cs#L975

It causes that `BeginConnect()` doesn't return immediately but returns until the name resolving completes, then unexpected blocking will happen in some async code calling this method.

MaximLipnin:
Fixes #13163

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -974,8 +974,8 @@ namespace System.Net.Sockets
 				throw new InvalidOperationException ();
 
 			var callback = new AsyncCallback ((result) => {
-				IPAddress[] hostAddresses = ((Task<IPAddress[]>)result).Result;
-				BeginConnect (hostAddresses, port, requestCallback, state);
+				var resultTask = ((Task<IPAddress[]>)result);
+				BeginConnect (resultTask.Result, port, requestCallback, resultTask.AsyncState);
 			});
 			return ConvertToApm<IPAddress[]> (Dns.GetHostAddressesAsync (host), callback, state);
 		}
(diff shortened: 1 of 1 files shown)


mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -972,7 +972,9 @@ public IAsyncResult BeginConnect (string host, int port, AsyncCallback requestCa
 			if (is_listening)
 				throw new InvalidOperationException ();
 
-			return BeginConnect (Dns.GetHostAddresses (host), port, requestCallback, state);
+			IPAddress[] hostAddresses = Dns.GetHostAddressesAsync (host).GetAwaiter ().GetResult ();
```
filipnavara: Unfortunately this doesn't really solve the problem. The `GetResult ()` call makes is synchronous again.

MaximLipnin: Thanks for help!

mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -972,7 +973,32 @@ public IAsyncResult BeginConnect (string host, int port, AsyncCallback requestCa
 			if (is_listening)
 				throw new InvalidOperationException ();
 
-			return BeginConnect (Dns.GetHostAddresses (host), port, requestCallback, state);
+			var callback = new AsyncCallback ((result) => {
+				IPAddress[] hostAddresses = ((Task<IPAddress[]>)result).Result;
+				BeginConnect (hostAddresses, port, requestCallback, state);
```
Therzok: to avoid capture, maybe just use `((Task<IPAddress[]>)result).AsyncState`

marek-safar: btw, this will still capture the context due to port and requestCallback parameters

PR #13950, opened 2019-04-09
Make nullable embedding API's public.

lateralusX:
Currently we have two nullable API's that will have value in the public embedding API (already requestsed by users), eliminating the needs to go from native->managed and back to get nullable information about a type.

mono_class_is_nullable
mono_class_get_nullable_param

This PR moves them from class-internals.h into class.h and mark them with MONO_API and when needed include GC UNSAFE/SAFE transition..

### DIFF
--- a/mono/metadata/class.h
+++ b/mono/metadata/class.h
@@ -292,6 +292,12 @@ mono_method_can_access_field (MonoMethod *method, MonoClassField *field);
 MONO_API mono_bool
 mono_method_can_access_method (MonoMethod *method, MonoMethod *called);
 
+MONO_API mono_bool
+mono_class_is_nullable (MonoClass *klass);
+
+MONO_API MONO_RT_EXTERNAL_ONLY MonoClass*
+mono_class_get_nullable_param (MonoClass *klass);
+
 MONO_END_DECLS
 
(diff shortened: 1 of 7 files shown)


mono/metadata/class.h
```diff
@@ -292,6 +292,12 @@ mono_method_can_access_field (MonoMethod *method, MonoClassField *field);
 MONO_API mono_bool
 mono_method_can_access_method (MonoMethod *method, MonoMethod *called);
 
+MONO_API mono_bool
+mono_class_is_nullable (MonoClass *klass);
+
+MONO_API MonoClass*
+mono_class_get_nullable_param (MonoClass *klass);
```
lambdageek: This one should probably be split up into an `_internal` version and a `MONO_RT_EXTERNAL_ONLY` `mono_class_get_nullable_param` which does a `MONO_ENTER_GC_UNSAFE`/`MONO_EXIT_GC_UNSAFE` transition.

`mono_class_from_mono_type_internal` can block in the runtime or access managed objects.

jaykrell: Shouldn't all new apis be split, even if the external one doesn't mode switch?
I.e. isn't the goal that mono itself never call an external api?

Issue #13639, opened 2019-03-25
mono-cil-strip crashes when called against the monotouch_Microsoft.CSharp_xunit-test.dll


mandel-macaque:
While integrating the test dlls with xamarin-macios, we started using device tests. When using the cil stip command we have the following failure:

`
Error: System.IndexOutOfRangeException: Index was outside the bounds of the array.
  at Mono.Cecil.Metadata.Utilities.ReadCompressedInteger (System.Byte[] data, System.Int32 pos, System.Int32& start) [0x0002f] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.Signatures.SignatureReader.ReadMarshalSig (System.Byte[] data) [0x00000] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.Signatures.SignatureReader.GetMarshalSig (System.UInt32 index) [0x00031] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.AggressiveReflectionReader.ReadMarshalSpecs () [0x00081] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.AggressiveReflectionReader.VisitTypeDefinitionCollection (Mono.Cecil.TypeDefinitionCollection types) [0x00055] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.ReflectionReader.VisitModuleDefinition (Mono.Cecil.ModuleDefinition mod) [0x00007] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.StructureReader.TerminateAssemblyDefinition (Mono.Cecil.AssemblyDefinition asm) [0x0002e] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.AssemblyDefinition.Accept (Mono.Cecil.IReflectionStructureVisitor visitor) [0x0001f] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.AssemblyFactory.GetAssembly (Mono.Cecil.Binary.ImageReader irv, System.Boolean manifestOnly) [0x00013] in <5ab86655477f49bf9e849e708f3dcbc9>:0
  at Mono.Cecil.AssemblyFactory.GetAssembly (Mono.Cecil.Binary.ImageReader reader) [0x00000] in <5ab86655477f49bf9e849e708f3dcbc9>:0
[...]

PR #13954, opened 2019-04-09
[cil-strip] Fix reading of undefined NativeType values in old Cecil

akoeplinger:
cil-strip crashed when processing this method: https://github.com/mono/corefx/blob/f54e64b341a7b114c3502e197b46670cacb28639/src/Microsoft.CSharp/tests/DefaultParameterTests.cs#L88

It uses `MarshalAs((UnmanagedType)2000)`, i.e. an undefined value for the native type. The old Cecil embedded in cil-strip tried to read more than one byte which blows up in this case.

Upstream Cecil just reads a single byte:
https://github.com/jbevain/cecil/blob/1b79b96d29f1e8e9ba832191b1b42ea9cb750d20/Mono.Cecil/AssemblyReader.cs#L3700-L3703

Fixes https://github.com/mono/mono/issues/13639

### DIFF
--- a/mcs/tools/cil-strip/Mono.Cecil.Signatures/SignatureReader.cs
+++ b/mcs/tools/cil-strip/Mono.Cecil.Signatures/SignatureReader.cs
@@ -926,8 +926,8 @@ namespace Mono.Cecil.Signatures {
 
 		MarshalSig ReadMarshalSig (byte [] data)
 		{
-			int start;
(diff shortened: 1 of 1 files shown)


marek-safar:
@monojenkins backport 2019-04

marek-safar:
@monojenkins backport 2019-02

PR #13959, opened 2019-04-09
[windows] Link libgcc statically on Windows

filipnavara:
Mono currently uses `__builtin_popcount[ll]` functions. On older processor architectures the popcount operations are implemented using software fallback in libgcc library. On MinGW this happens to pull in dependency on two dynamic libraries (libgcc_s_seh-1.dll and transitively libwinpthread-1.dll). These two libraries are together ~130 Kb of code that is pulled in because of ~12 bytes of one function. Linking statically to libgcc reduces the overhead to 4Kb in debug build (page size aligned) and simplifies the deployment because the resulting binary has no external dependencies.


lateralusX:
On MSVC build Mono runtime we have the ability to static link c-runtime as well as build mono runtime, but default is dynamic linking. For mingw build I'm fine doing static libgcc as default (doesn't look like we have configure option on other platforms for this either), we already have cross compiler building this way.

filipnavara:
Note that this is not the C runtime. It's just GCC helper library for software fallbacks and things like C++ exceptions. The resulting binaries still link to the Microsoft C runtime (msvcrt.dll by default).

filipnavara:

PR #13960, opened 2019-04-09
[netcore] Eliminate circular issue between link-mono and prepare targets

directhex:
As-is, `make runtime` and `make bcl` will fail until after `make prepare` has
been run, but `make prepare` calls `make link-mono` which depends on both
`make runtime` and `make bcl`

Break that circular dependency, so calling `make runtime` on its own works,
on a clean checkout

### DIFF
--- a/netcore/Makefile.am
+++ b/netcore/Makefile.am
@@ -41,12 +41,12 @@ run-sample:
 # COREHOST_TRACE=1 
 SHAREDRUNTIME := shared/Microsoft.NETCore.App/$(NETCOREAPP_VERSION)
 
-bcl:
+bcl: prepare-stamp
 	$(MAKE) -C ../mcs/build/ common/Consts.cs
 	$(MAKE) -C ../mcs/class/System.Private.CoreLib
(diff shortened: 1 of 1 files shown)


filipnavara:
Another approach would be to skip the copying from `runtime` and `bcl` and move it to separate targets. It would probably need some adjustments for my workflow (`make runtime bcl xtest-<whatever>`), but the upside would be that building runtime, bcl or the nuget would not require to download the additional .NET Core runtime packages.

directhex:
@filipnavara I don't disagree, but my aim has been to minimize workflow adjustments.

filipnavara:
> my aim has been to minimize workflow adjustments.

Sure. I think the change is improvement as-is.

Maybe the copying can be removed from `bcl` and `runtime` and `xtest*` targets can add dependency on `link-mono`. That way my workflow will still work with no change. Just a food for thought.

vargaz:
make runtime/make bcl are helper targets I use during development, they should not be used during CI etc. CI can just run the normal make build.

directhex:
Huh. So it does. That simplifies things a little.

PR #13992, opened 2019-04-11
[wasm] Hackish netcore support.

vargaz:


vargaz:
With some additional hacks, this can run a hello world.

vargaz:
In particular:
- run make build-interp-sample
- Copy the following to bin/interp-sample/managed:
 - mono's System.Private.CoreLib
 - a netcore hello.exe
 - netcore's
-  "System.Runtime.dll", "System.Console.dll", "System.Runtime.Extensions.dll", "System.Threading.dll", "System.Text.Encoding.Extensions.dll", "System.Resources.ResourceManager.dll"

Use this as mono-config.js:
```
config = {
 	vfs_prefix: "managed",
 	deploy_prefix: "managed",
 	enable_debugging: 0,
 	file_list: [ "hello.exe", "System.Private.CoreLib.dll", "System.Runtime.dll", "System.Console.dll", "System.Runtime.Extensions.dll", "System.Threading.dll", "System.Text.Encoding.Extensions.dll", "System.Resources.ResourceManager.dll" ],
}
```
Run with:
```
~/.jsvu/sm runtime.js --setenv=DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 --run hello.exe
```

mono/metadata/assembly.c
```diff
@@ -4176,6 +4176,18 @@ mono_assembly_load_corlib (const MonoRuntimeInfo *runtime, MonoImageOpenStatus *
...
+	if (!corlib) {
+		MonoAssemblyCandidatePredicate predicate = &mono_assembly_candidate_predicate_sn_same_name;
+		void* predicate_ud = aname;
+		MonoAssemblyOpenRequest req;
+		mono_assembly_request_prepare (&req.request, sizeof (req), MONO_ASMCTX_DEFAULT);
+		req.request.predicate = predicate;
+		req.request.predicate_ud = predicate_ud;
+
+		MonoImageOpenStatus status;
+		char *corlib_name = g_strdup_printf ("%s.dll", MONO_ASSEMBLY_CORLIB_NAME);
```
jaykrell: nit: `printf` is overkill for concating and possibly duplicating constant strings.

If you need a copy, to be later freed, you can `g_strdup (MONO_ASSEMBLY_CORLIB_NAME ".dll")`.
If you don't need a copy, you can `MONO_ASSEMBLY_CORLIB_NAME ".dll"`

mono/metadata/object-offsets.h
```diff
@@ -88,10 +88,17 @@ DECL_OFFSET(MonoDelegate, method_code)
 DECL_OFFSET(MonoDelegate, method_is_virtual)
 DECL_OFFSET(MonoDelegate, extra_arg)
 
+#ifdef ENABLE_NETCORE
```
jaykrell: nit: A `typedef` could clean this and the other up, albeit maybe less clear due to abstraction.
````
#ifdef ENABLE_NETCORE
typedef MonoThread MonoInternalThread;
#endif 
```

mono/metadata/sysmath.c
```diff
@@ -222,7 +222,12 @@ ves_icall_System_Math_ScaleB (gdouble x, gint32 n)
 gdouble
 ves_icall_System_Math_FusedMultiplyAdd (gdouble x, gdouble y, gdouble z)
 {
+#ifdef TARGET_WASM
+	/* Enabling these leads to error: undefined symbol: llvm_fma_f32 from esmcripten */
+	g_assert_not_reached ();
```
jaykrell: `return x * y + z`?

jaykrell: I thought it might be close enough. Agreed.

EgorBo: The float version of `FusedMultiplyAdd` can be emulated with doubles though:
`return ((float)((double)x) * ((double)y) + ((double)z));`

sdks/wasm/Makefile
```diff
@@ -96,19 +105,27 @@ debug/:
...
-release/.stamp-build: driver.o corebindings.o library_mono.js binding_support.js dotnet_support.js $(TOP)/sdks/out/wasm-runtime-release/lib/libmonosgen-2.0.a debug/.stamp-build | release/
+release/.stamp-build: driver.o corebindings.o library_mono.js binding_support.js dotnet_support.js $(TOP)/sdks/out/wasm-runtime-release/lib/libmonosgen-2.0.a $(MONO_LIB_DEPS) debug/.stamp-build | release/
 	$(EMCC) $(EMCC_FLAGS) -Oz --llvm-opts 2 --llvm-lto 1 --js-library library_mono.js --js-library binding_support.js --js-library dotnet_support.js driver.o corebindings.o $(MONO_LIBS) -o release/mono.js -s "EXPORTED_FUNCTIONS=['_putchar']"
 	touch $@
 
+netcore-System.Native.bc:
+	cd $(COREFX_ROOT)/src/Native && touch _version.c && cd Unix && mkdir -p build && cd build && source $(TOP)/sdks/builds/toolchains/emsdk/emsdk_env.sh && emcmake cmake "-DEMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES=1" "-DCMAKE_TOOLCHAIN_FILE=$(TOP)/sdks/builds/toolchains/emsdk/emscripten/1.38.29/cmake/Modules/Platform/Emscripten.cmake" -G Ninja -DCMAKE_BUILD_TYPE=Debug .. && ninja && cp System.Native/System.Native.so $(PWD)/$@
+
```
kjpou1: Can we put the `1.38.29` in a variable so it is easier to find and set?  Maybe this should be made global somewhere so that this uses the version that is used when provisioning.

steveisok:
@vargaz is this a to be continued type thing?  Or is it old and can be closed?

vargaz:
We will need it eventually.

PR #13998, opened 2019-04-11
[netcore] hand-craft a dist-recursive rule for System.Private.CoreLib

directhex:
`make dist` stopped erroring in #13961 but didn't actually work, because the relevant Makefile.am was not distributed, causing configure to fail (because the Makefile.am was cited as an input, unconditionally).

We can't rely on the dist-recusrive from library.make or build.make because this lib is in a weirdo format which doesn't comply with the requirements of those helpers, so craft a simple handwritten replacement for our special case

### DIFF
--- a/mcs/class/System.Private.CoreLib/Makefile.am
+++ b/mcs/class/System.Private.CoreLib/Makefile.am
@@ -2,10 +2,16 @@ thisdir = class/System.Private.CoreLib
 SUBDIRS =
 #include ../../build/rules.make
 
+dirs := $(dir $(wildcard */*))
+files := $(wildcard */*.cs)
+
 all-local:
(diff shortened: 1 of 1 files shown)


akoeplinger:
@monojenkins build deb

directhex:
@monojenkins build deb

directhex:
@monojenkins build deb

directhex:
@monojenkins build deb

directhex:
OK, that was exciting, lots of .xz related issues to work through on the "build deb" job.

@akoeplinger seems to be building OK from a tarball now: https://jenkins.mono-project.com/job/build-package-dpkg-mono-pullrequest/label=ubuntu-1404-amd64/410/console is well into the BCL, and not breaking in configure as before

PR #14011, opened 2019-04-12
[netcore] Smarter version numbering of nupkg

directhex:
`nuget pack runtime.nuspec -properties VERSION=6.3.0.208\;RID=linux-x64\;PLATFORM_AOT_SUFFIX=.so\;COREARCH=x64`

### DIFF
--- a/netcore/Makefile
+++ b/netcore/Makefile
@@ -35,6 +35,10 @@ URL := https://dotnetcli.blob.core.windows.net/dotnet/Runtime/$(NETCOREAPP_VERSI
 FEED_BASE_URL := https://dotnetfeed.blob.core.windows.net/dotnet-core
 TEST_ASSETS_URL = $(FEED_BASE_URL)/corefx-tests/$(NETCORETESTS_VERSION)/OSX.x64/netcoreapp/corefx-test-assets.xml
 
+# used to calculate exact version number for generating nupkg
+BLAME = $(shell git blame ../configure.ac | grep AC_INIT | cut -f1 -d' ')
+VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
+
 $(NETCORESDK_FILE):
 	curl $(URL) --output $(NETCORESDK_FILE)
 	$(UNZIPCMD) $(NETCORESDK_FILE)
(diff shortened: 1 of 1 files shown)


netcore/Makefile.am
```diff
@@ -57,7 +57,9 @@ link-mono:
 prepare: $(NETCORESDK_FILE) update-corefx link-mono
 
 nupkg:
-	nuget pack runtime.nuspec -properties VERSION=@VERSION@\;RID=@RID@\;PLATFORM_AOT_SUFFIX=@PLATFORM_AOT_SUFFIX@\;COREARCH=@COREARCH@
+BLAME = $(shell git blame $(top_srcdir)/configure.ac | grep AC_INIT | cut -f1 -d' ')
+VERSTUB = .$(shell git log $(BLAME)..HEAD --oneline | wc -l)
```
filipnavara: nit: `git rev-list --count $(BLAME)..HEAD`

I know there are other places in the build scripts doing this, so it's probably worth a separated patch just for that.

akoeplinger: Interesting, I've never seen variables being defined inside of a Makefile target. Is that an automake feature?

filipnavara: I do it in regular Makefiles, so I think it's quite universal. I never seem to remember the difference between `:=` and `=` though.

directhex:
You say that, and now I'm worried it's not doing the right thing, i.e. always evaluating the variables. I'll test more on Monday to ensure it's working as planned.

directhex:
`$(top_srcdir)` isn't populated without .am :roll_eyes: looking for an adequate replacement. Maybe just ..

directhex:
OK. This is working as expected in VSTS.

directhex:
@monojenkins build failed

directhex:
@monojenkins build failed

EgorBo:
Btw, the content of this nuget package is wrong
1) System.Private.CoreLib.dll should be under `runtimes/osx-x64/lib/netcoreapp3.0`
2) `libmonosgen-2.0.dylib` for now should be `libcoreclr.dylib` in the output dir

directhex:
@EgorBo We are not pretending to be CoreCLR in what we deliver to the Core team. It's on them to cleanly consume Mono. We need to be in a position to give them something they can start integrating.

EgorBo:
@directhex ah, ok

directhex:
@EgorBo I took the directory layout in the nupkg from e.g. runtime.rhel.6-x64.Microsoft.NETCore.Runtime.CoreCLR

PR #14013, opened 2019-04-12
[sdks] Include 'etc/mono/config'

alexischr:
Still not 100% this will be the final approach - while getting the config file from the archive takes care of the proper name of `libmono-native` as well as the libary suffix, this also uses `$mono_libdir` which I am not sure I can make work on the Xamarin Android side.


alexischr:
`Unhandled Exception:
System.DllNotFoundException: /Users/alexischr/xamarin-android/external/mono/sdks/out/android-host-Darwin-release/lib/../lib/libmono-native.dylib`

Unfortunately the archive build bakes the local `mono_libdir` in the runtime (not that it would have a way to know the correct path), and setting `mono` and `libmono-native.dylib` side-by-side did not work either. We may not even be allowed to bake any path at all, as this file would be re-bundled by XA in what I believe is their redistributable package.

PR #14016, opened 2019-04-12
Respect call convention modopts in native wrappers.

madewokherd:
MS's C++ compiler emits code that uses this.


madewokherd:
Hang on, this is going to override a proper calling convention if it's been set. I'll need to put a little more thought into this.

PR #14017, opened 2019-04-12
Respect call convention modopts in native wrappers.

madewokherd:
MS's C++ compiler emits code that uses this.


lewurm:
Looks good to me, but I'm no expert in this area.

@vargaz @lambdageek could you review please?

marek-safar:
@vargaz please review this

PR #14018, opened 2019-04-12
Protecting boxing a null value

### LINKED ISSUE #13969: (regression) SIGSEGV in RuntimeMethodInfo:InternalInvoke with xunit
Environment: 
Ubuntu 16.04 Azure Pipelines image with
`Mono JIT compiler version 5.18.0.268 (tarball Thu Feb 28 15:29:06 UTC 2019)`

was working completely fine with 
`Mono JIT compiler version 5.18.0.240 (tarball Wed Jan 16 09:13:03 UTC 2019)`

Repro:
```
git clone --recursive git@github.com:AvaloniaUI/Avalonia.git
cd Avalonia
sudo apt-get update && sudo apt-get install castxml
dotnet tool install --global Nuke.GlobalTool --version 0.12.3
nuke --target CiAzureLinux --configuration=Release
```

One of xunit test assemblies segfaults like this https://dev.azure.com/AvaloniaUI/AvaloniaUI/_build/results?buildId=1851&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=d9b15392-e4ce-5e4c-0c8c-b69645229181&lineStart=1132&lineEnd=1133&colStart=1&colEnd=1

I've also checked with mono built from master, still segfaults, but I'm also seeing messages like these:
```
ves_icall_TypeBuilder_create_runtime_class_raw USED 151 handles
ves_icall_TypeBuilder_create_runtime_class_raw USED 156 handles
ves_icall_TypeBuilder_create_runtime_class_raw USED 104 handles
ves_icall_TypeBuilder_create_runtime_class_raw USED 108 handles
ves_icall_TypeBuilder_create_runtime_class_raw USED 158 handles
```

Not sure if that's relevant.

Also reproduced on OSX with `MonoFramework-MDK-5.18.0.268.macos10.xamarin.universal.pkg`, works fine with 5.18.0.225.

thaystg:
Trying to box a null value crashes inside mono_gc_wbarrier_value_copy_internal. As described by the customer using version 5.18.0.240 the crash doesn't happen.

As I could debug using this version doesn't crash only because is ignoring the SIGSEGV, when I attach to LLDB I can see the same crash in both versions.
I protected to only add data to res_handle if there is a valid value.
Fixes #13969

### DIFF
--- a/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
+++ b/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
@@ -658,6 +658,15 @@ namespace MonoTests.System.Reflection
 			Assert.AreEqual (10, pi2.GetGetMethod ().Invoke (10, null));
 		}
 
+		[Test]
+		public void NullableTestsStatic ()
+		{
+			Nullable<Double> val = new Nullable<Double>(new Double());
+			MethodInfo mi = typeof (Nullable<Double>).GetMethod ("op_Implicit");
+            object obj = val;
+            mi.Invoke(null, new[] { obj });
+		}
+
 		public static void foo_generic<T> ()
 		{
 		}
--- a/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
+++ b/mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
@@ -663,8 +663,8 @@ namespace MonoTests.System.Reflection
 		{
 			Nullable<Double> val = new Nullable<Double>(new Double());
 			MethodInfo mi = typeof (Nullable<Double>).GetMethod ("op_Implicit");
-            object obj = val;
-            mi.Invoke(null, new[] { obj });
+			object obj = val;
+			mi.Invoke(null, new[] { obj });
 		}
 
 		public static void foo_generic<T> ()
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -5528,16 +5528,19 @@ mono_runtime_try_invoke_array (MonoMethod *method, void *obj, MonoArray *params,
 		return (MonoObject *)obj;
 	} else {
 		if (mono_class_is_nullable (method->klass)) {
-			MonoObject *nullable;
+			if (method->flags & METHOD_ATTRIBUTE_STATIC) {
+				obj = NULL;
+			} else {
+				MonoObject *nullable;
+				/* Convert the unboxed vtype into a Nullable structure */
+				nullable = mono_object_new_checked (mono_domain_get (), method->klass, error);
+				return_val_if_nok (error, NULL);
 
-			/* Convert the unboxed vtype into a Nullable structure */
-			nullable = mono_object_new_checked (mono_domain_get (), method->klass, error);
-			return_val_if_nok (error, NULL);
-
-			MonoObject *boxed = mono_value_box_checked (mono_domain_get (), m_class_get_cast_class (method->klass), obj, error);
-			return_val_if_nok (error, NULL);
-			mono_nullable_init ((guint8 *)mono_object_unbox_internal (nullable), boxed, method->klass);
-			obj = mono_object_unbox_internal (nullable);
+				MonoObject *boxed = mono_value_box_checked (mono_domain_get (), m_class_get_cast_class (method->klass), obj, error);
+				return_val_if_nok (error, NULL);
+				mono_nullable_init ((guint8 *)mono_object_unbox_internal (nullable), boxed, method->klass);
+				obj = mono_object_unbox_internal (nullable);
+			}
 		}
 
 		/* obj must be already unboxed if needed */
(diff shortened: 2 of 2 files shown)


mono/metadata/object.c
```diff
@@ -6946,37 +6946,38 @@ mono_value_box_handle (MonoDomain *domain, MonoClass *klass, gpointer value, Mon
...
-			*(guint16 *)(data) = *(guint16 *) value;
-			break;
-		case 4:
-			*(guint32 *)(data) = *(guint32 *) value;
-			break;
-		case 8:
-			*(guint64 *)(data) = *(guint64 *) value;
-			break;
-		default:
+	if (value != NULL) {
```
lambdageek: So if the source value is null we return a result with some default contents?  That seems easy to use incorrectly.  

Can we assert in here that `value != NULL` and fix the callers instead?  or maybe there is some managed exception that we should throw?

thaystg:
@monojenkins build failed

mcs/class/corlib/Test/System.Reflection/MethodInfoTest.cs
```diff
@@ -658,6 +658,15 @@ public void NullableTests ()
 			Assert.AreEqual (10, pi2.GetGetMethod ().Invoke (10, null));
 		}
 
+		[Test]
+		public void NullableTestsStatic ()
+		{
+			Nullable<Double> val = new Nullable<Double>(new Double());
+			MethodInfo mi = typeof (Nullable<Double>).GetMethod ("op_Implicit");
+            object obj = val;
```
lambdageek: Indentation

mono/metadata/object.c
```diff
@@ -6928,6 +6931,7 @@ mono_value_box_handle (MonoDomain *domain, MonoClass *klass, gpointer value, Mon
 	error_init (error);
 
 	g_assert (m_class_is_valuetype (klass));
+	g_assert (value != NULL);
```
lambdageek: 🎉

mono/metadata/object.c
```diff
@@ -6937,6 +6941,7 @@ mono_value_box_handle (MonoDomain *domain, MonoClass *klass, gpointer value, Mon
 	if (mono_class_is_nullable (klass))
 		return mono_nullable_box_handle (value, klass, error);
 
+
```
lambdageek: Stray whitespace change

mono/metadata/object.c
```diff
@@ -6946,7 +6951,7 @@ mono_value_box_handle (MonoDomain *domain, MonoClass *klass, gpointer value, Mon
 	return_val_if_nok (error, NULL_HANDLE);
 
 	size -= MONO_ABI_SIZEOF (MonoObject);
-
+
```
lambdageek: Stray whitespace

mono/metadata/object.c
```diff
@@ -6978,6 +6983,7 @@ mono_value_box_handle (MonoDomain *domain, MonoClass *klass, gpointer value, Mon
 #endif
 		MONO_EXIT_NO_SAFEPOINTS;
 	}
+
```
lambdageek: Stray whitespace.

mono/metadata/object.c
alexischr: @lambdageek  This assert would be nice to have on 2019-02. Maybe you have already proved me wrong but I still suspect `[MemberData]` getting into trouble using reflection to transform C# tuples to argument lists.

lambdageek: @alexischr Can you point us at an example of the tuples transformation? Sounds like another interesting test case.

lambdageek:

thaystg:
> @lambdageek backporting to 2018-10 failed, the patch results in conflicts:
> 
> ```
> Applying: Trying to box a null value crashes inside mono_gc_wbarrier_value_copy_internal. As described by the customer using version 5.18.0.240 the crash doesn't happen. As I could debug using this version doesn't crash only because is ignoring the SIGSEGV, when I attach to LLDB I can see the same crash in both versions. I protected to only add data to res_handle if there is a valid value. Fixes #13969
> Using index info to reconstruct a base tree...
> M	mono/metadata/object.c
> .git/rebase-apply/patch:40: trailing whitespace.
> 	if (value != NULL) { 
> warning: 1 line adds whitespace errors.
> Falling back to patching base and 3-way merge...
> Auto-merging mono/metadata/object.c
> CONFLICT (content): Merge conflict in mono/metadata/object.c
> error: Failed to merge in the changes.
> Patch failed at 0001 Trying to box a null value crashes inside mono_gc_wbarrier_value_copy_internal. As described by the customer using version 5.18.0.240 the crash doesn't happen. As I could debug using this version doesn't crash only because is ignoring the SIGSEGV, when I attach to LLDB I can see the same crash in both versions. I protected to only add data to res_handle if there is a valid value. Fixes #13969
> ```
> Please backport manually!

@lambdageek, should I do it manually?

thaystg:

lambdageek:
@thaystg yes, please make a 2018-10 backport.

it might be easier to cherrypick the commit if you first rewrite it on a local branch to squash all the changes together - I think the two older commits are probably conflicting with something on the 2018-10 branch.

thaystg:

mono/metadata/object.c
alexischr: @lambdageek Unfortunately that was back in the summer, I don't remember any details. Digging into the source brought it back (it was params object[]): https://github.com/xunit/xunit/commit/ff6ecaaf4df79f4266e7d2ab24493e49a5133031#diff-c6cbc959191da11fa205494c2731efe9L27 . Some of those methods look great for testing :)

thaystg:
@monjenkins build Linux AArch64 Coop Suspend

thaystg:
@monjenkins build Linux AArch64 Coop Suspend

thaystg:
@monojenkins build failed

PR #14026, opened 2019-04-14
[aot] Change how the context is computed in init_aot_method (). Instead of getting it from the caller, save it in the aot image and read it back. This will make it possible to emit direct calls to gshared methods later.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14028, opened 2019-04-14
Remove redundant check of llvm_acfg->aot_opts.direct_icalls.

jaykrell:
It is checked just a few lines earlier.


jaykrell:
@monojenkins build failed

PR #14032, opened 2019-04-14
[netcore] Do not inline if [DynamicSecurityMethod] exists

EgorBo:
I think it's just a typo made in https://github.com/mono/mono/pull/13906

I tried to run System.Runtime.Tests on top of the fresh master and I got:
```
Method 'void Xunit.DiaSession:.cctor ()' which contains a StackCrawlMark local variable must be decorated with [System.Security.DynamicSecurityMethod].
```
so I suspect one of the `Type.GetType` was inlined despite the `[DynamicSecurityMethod]` attribute.

However after this fix I am getting:
```
System.InvalidOperationException: Could not find/load any of the following assemblies: xunit.execution.dotnet.dll
```

**UPD**: Applying @filipnavara's fix for the `InvalidOperationException` (missing null-check in CompurePublickKey).

cc: @filipnavara


filipnavara:
LGTM.

The other error is caused by missing null check for `_publicKey` in `AssemblyName.ComputePublicKeyToken`.

PR #14039, opened 2019-04-15
DO NOT MERGE: Move EventSource to shared CoreLib partition

filipnavara:
Counter-part to https://github.com/dotnet/coreclr/pull/23925 (for build tests / cherry-pick into sync PR).


marek-safar:
@filipnavara shared changes need to go via the mirror (otherwise it will break on next sync)

filipnavara:
@marek-safar Totally understood. For the complex changes I submit PRs to all the mirrored repositories to test builds and to allow cherry-picking these PRs into the merge commits. In this case the Mono changes turned out to be quite small (just removing EventSource.cs and updating csproj), but the changes to other repositories required me to rework the original code few times.

filipnavara:
> otherwise it will break on next sync

Turns out few changes still didn't propagate through the mirror correctly to Mono repository. All the other repositories are already in sync (except for CoreRT missing a new line at the end of Utf8.cs). /cc @Anipik

PR #3486, opened 2016-08-30
[aot] change method table generation so that it works for larger assemblies on ARM

lewurm:
At the end of the AOT compilation, we dump a table containing all method
entrypoints using the toolchain linker to resolve the addresses. Previously
we used `bl` for that, however on larger assemblies the offset might
overflow the 24bit immediate (thus, the linker bailing out with
"Error: branch out of range"). Instead, using `ldr pc,=<label>` followed
by `.ltorg` emits the full 32bit address to a method. That takes 8 bytes
for each method instead of 4 bytes.

cc @vargaz @ludovic-henry

PR #14043, opened 2019-04-15
[aot] change method table generation so that it works for larger assemblies on ARM

lewurm:
At the end of the AOT compilation, we dump a table containing all method
entrypoints using the toolchain linker to resolve the addresses. Previously
we used `bl` for that, however on larger assemblies the offset might
overflow the 24bit immediate (thus, the linker bailing out with
"Error: branch out of range"). Instead, using `ldr pc,=<label>` followed
by `.ltorg` emits the full 32bit address to a method. That takes 8 bytes
for each method instead of 4 bytes.

( I couldn't reopen https://github.com/mono/mono/pull/3486 anymore)

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -1580,6 +1580,21 @@ arch_emit_direct_call (MonoAotCompile *acfg, const char *target, gboolean extern
 	g_assert_not_reached ();
 #endif
 }
+
+static void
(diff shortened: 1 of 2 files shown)


lewurm:
PR verified on Android by @grendello. We gonna need to backport this up to `2018-10`

lewurm:
@monojenkins squash

marek-safar:
@monojenkins backport 2019-04

marek-safar:
@monojenkins backport 2019-02

PR #14056, opened 2019-04-15
[netcore] fix some CustomAttribute related tests

EgorBo:
This fixes 12 tests in System.Runtime.Tests.
Basically the fix replicates https://github.com/dotnet/coreclr/blob/edffbd9c545aa9c1f6cb7ae048efdf5db13d3991/src/System.Private.CoreLib/src/System/Attribute.CoreCLR.cs#L457-L458

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/Attribute.cs
+++ b/mcs/class/System.Private.CoreLib/System/Attribute.cs
@@ -39,19 +39,7 @@ namespace System
         public static Attribute[] GetCustomAttributes (ParameterInfo element, bool inherit) => (Attribute[])MonoCustomAttrs.GetCustomAttributes (element, inherit);
         public static Attribute[] GetCustomAttributes (ParameterInfo element, Type attributeType) => (Attribute[])GetCustomAttributes ((ICustomAttributeProvider)element, attributeType, true);
         public static Attribute[] GetCustomAttributes (ParameterInfo element, Type attributeType, bool inherit) => (Attribute[])GetCustomAttributes ((ICustomAttributeProvider)element, attributeType, inherit);
-
-        internal static Attribute[] GetCustomAttributes (ICustomAttributeProvider element, Type attributeType, bool inherit)
-        {
-            if (attributeType == null)
-                throw new ArgumentNullException (nameof (attributeType));
-#if !NETCORE
-            if (!attributeType.IsSubclassOf (typeof (Attribute)) && attributeType != typeof (Attribute))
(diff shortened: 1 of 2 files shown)


mcs/class/corlib/System/MonoCustomAttrs.cs
```diff
@@ -149,7 +149,15 @@ internal static object[] GetCustomAttributes (ICustomAttributeProvider obj, Type
 			if (obj == null)
 				throw new ArgumentNullException ("obj");
 			if (attributeType == null)
-				throw new ArgumentNullException ("attributeType");	
+				throw new ArgumentNullException ("attributeType");
+			
+#if NETCORE
+			if (!attributeType.IsSubclassOf(typeof(MonoCustomAttrs)) &&
```
vargaz: Do we really need both of these classes ?

EgorBo: Yeah, `MonoCustomAttrs` check is redundant, removed.

mcs/class/System.Private.CoreLib/System/Attribute.cs
```diff
@@ -44,8 +44,10 @@ internal static Attribute[] GetCustomAttributes (ICustomAttributeProvider elemen
         {
             if (attributeType == null)
                 throw new ArgumentNullException (nameof (attributeType));
+#if !NETCORE
```
marek-safar: Why do you need this if-def?

EgorBo: @marek-safar it's just redundant for netcore - it calls `MonoCustomAttrs.GetCustomAttributes` where it checks for `Is Attribute or subclass` again.

marek-safar: This code is used only for netcore

EgorBo: Oh, indeed

filipnavara:
How many test failures do you get on `System.ComponentModel.Composition.Tests` now? That was the reason why I moved the check. CoreCLR also has it in Attribute class.

filipnavara:
`System.ComponentModel.Composition.Tests  Total: 1428, Errors: 0, Failed: 281, Skipped: 0, Time: 1.972s` ... it fixed 12 System.Runtime.Tests, but it broke 279 tests here. We should come up with a plan to run some subset of the test on CI.

PR #14057, opened 2019-04-15
[llvmonly] Fix a buffer overflow.

vargaz:


vargaz:
@monojenkins build failed

PR #14058, opened 2019-04-15
[2019-04][MacSDK] Bump F# to 4.5 (#10919)

alexischr:
Backport of https://github.com/mono/mono/pull/10919

* Use a custom prefix instead of trying to infer it, which doesn't work under bockbuild.

The following line is problematic in the scope of building the Mac SDK package (using bockbuild):

https://github.com/fsharp/fsharp/blob/ccfccbacfdb16b2ab143cf83f5ece20f738fd864/mono/config.make#L5

It's problematic because where `mono` lives in this case would be something like `/macsdk-staging-dir/bin/mono` which is not the actual prefix (prefix here being the final destination, `/Library/Frameworks/Mono.framework/...`) but a staging area that bockbuild uses (this is how we build on system prefixes without installing on CI)

For the F# build, bockbuild expects things to end up  in `/fsharp-staging-dir/$(prefix)` (because that's what most `make install`s do when given `DESTDIR=/fsharp-staging-dir/`). If things end up in _any other_ directory in `/fsharp-staging-dir/`  (like `/fsharp-staging-dir/macsdk-staging-dir` in our case) that would cause us to miss packaging those files (this has happened before, with packages that have broken or different use of `DESTDIR`). So there is an added write-protection for anywhere in `/fsharp-staging-dir` that *isn't*  `/fsharp-staging-dir/$(prefix)` , because a write-protection error, while obtuse, is much preferable to missing the fact that some files are not packaged (as would have happened here)

So the fix is to stop trying to infer the prefix, and return to providing it explicitly to the build system (this used to be the case via the `configure` step)

* Also, if we are doing an override of the 'make' command, we need to take care of passing DESTDIR ourselves too

* Since now we perform 'make install' on the make step, make the 'install' step into a no-operation.

* Look for sn where mono is, not in the final destination (which again, may not be the same)


alexischr:

PR #14063, opened 2019-04-15
[netcore] Fix couple of custom attribute tests

filipnavara:
This is fix for regressions from #14056 (System.ComponentModel.Composition.Tests) and #14050 (System.Runtime.Tests).

Essentially the issue is that `Attribute` methods (`GetCustomAttribute`, `GetCustomAttributes`, `IsDefined`) have to check if the `attributeType` is subclass of `Attribute` or `Attribute` itself. On the other hand other custom attribute methods should not check that because someone can ask for all custom attributes implementing some interface as the component model code does. All of these calls end up calling `MonoCustomAttrs` so the `attributeType` parameter checks have to be in `Attribute` class.

This highlights the need to get some CI working ASAP since we keep breaking it (myself included). I'm keeping this as draft until I run the full test suite on my machine to check for any regressions.

### DIFF
--- a/mcs/class/System.Private.CoreLib/System/Attribute.cs
+++ b/mcs/class/System.Private.CoreLib/System/Attribute.cs
@@ -6,6 +6,11 @@ namespace System
 	{
 		static Attribute GetAttr (ICustomAttributeProvider element, Type attributeType, bool inherit)
 		{
+            if (attributeType == null)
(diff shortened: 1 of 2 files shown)


filipnavara:
Results are in. `Total: 358542 Failed: 829` (master + PR + #14039), 13 tests fixed since last full run in #14050.

PR #14064, opened 2019-04-15
[netcore] Add linker descriptor for corelib.

vargaz:


vargaz:
@monojenkins build failed

PR #14076, opened 2019-04-16
mono_marshal_get_proxy_cancast register icalls before using them.

jaykrell:
Alternatively move the registration to mono_remoting_init.

This should fix a problem I see in other work, and *seems* like an existing race condition.

In particular, you can emit a function pointer, and then:

```
		case MONO_CEE_MONO_ICALL: {
			gpointer func;
			MonoJitICallInfo *info;

			if (!info)
				g_error ("Could not find icall address in wrapper %s", mono_method_full_name (method, 1));
			g_assert (info);
```

though I don't know why with my changes this always hits, and previously never.
Anyway, it seems like this is the original intent.

Moving the initializer earlier would make it easier to maintain correctly.

### DIFF
--- a/mono/metadata/remoting.c
+++ b/mono/metadata/remoting.c
@@ -173,13 +173,42 @@ mono_remoting_init (void)
 }
 
 static void
+register_icalls (void)
+{
+	static gboolean icalls_registered = FALSE;
+
+	mono_loader_lock ();
+
+	if (!icalls_registered) {
+		register_icall (type_from_handle, "type_from_handle", mono_icall_sig_object_ptr, FALSE);
+		register_icall (mono_marshal_set_domain_by_id, "mono_marshal_set_domain_by_id", mono_icall_sig_int32_int32_int32, FALSE);
+		register_icall (mono_marshal_check_domain_image, "mono_marshal_check_domain_image", mono_icall_sig_int32_int32_ptr, FALSE);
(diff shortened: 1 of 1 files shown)


jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/remoting.c
```diff
@@ -1902,6 +1909,9 @@ mono_marshal_get_proxy_cancast (MonoClass *klass)
 	mb->method->save_lmf = 1;
 
 #ifndef DISABLE_JIT
+
```
vargaz: Just call mono_remoting_marshal_init () here like in the other cases ?

jaykrell: yeah, I tried that. I had some failures, I could not easily explain, so I decided to try something semantically smaller.
You know, to skip the, I'm not sure what it is -- class loading?
I can try that again. I'll make a separate PR for that, so the success stands here in case that fails/is-rejected.

jaykrell:

jaykrell:
Redundant with https://github.com/mono/mono/pull/14097.

PR #14088, opened 2019-04-16
[llvmonly] Allow direct calls to inflated methods since they can initialize themselves correctly now.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14094, opened 2019-04-16
[corefx] Use fallbacks when link() fails

alexischr:
System.IO.Filesystem's MoveFile() and ReplaceFile() use link/unlink under some conditions to avoid a copy. This fixes two problems with that:

* MoveFile()'s fall-back does not recognize Android's error as non-critical, and bails out on the fall-back with an exception. From logs we know that 'strerror_r(android_error)' is 'Permission denied' so the missing error should be EACCES (we already check for EPERM)
* While MoveFile() was using a fall-back, ReplaceFile() wasn't.

Also improve a relevant test for more useful fault output.

### DIFF
--- a/mcs/class/corlib/Test/System.IO/FileTest.cs
+++ b/mcs/class/corlib/Test/System.IO/FileTest.cs
@@ -923,7 +923,7 @@ namespace MonoTests.System.IO
 			}
 			
 			Assert.IsTrue (File.Exists (bar), "#1");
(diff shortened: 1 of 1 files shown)


lambdageek:
This corefx change is https://github.com/mono/corefx/pull/275

alexischr:
This took care of the issues. I will merge for the integration, and open a PR upstream before bringing this to master and others.

PR #14097, opened 2019-04-17
mono_remoting_marshal_init in mono_marshal_get_proxy_cancast to register icalls.

jaykrell:
Alternative to https://github.com/mono/mono/pull/14076.

Alternatively move the registration to mono_remoting_init.

This should fix a problem I see in other work, and *seems* like an existing race condition.

In particular, you can emit a function pointer, and then:

```
		case MONO_CEE_MONO_ICALL: {
			gpointer func;
			MonoJitICallInfo *info;

			if (!info)
				g_error ("Could not find icall address in wrapper %s", mono_method_full_name (method, 1));
			g_assert (info);
```

though I don't know why with my changes this always hits, and previously never.
Anyway, it seems like this is the original intent.

Moving the initializer earlier would make it easier to maintain correctly.

### DIFF
--- a/mono/metadata/remoting.c
+++ b/mono/metadata/remoting.c
@@ -1902,6 +1902,9 @@ mono_marshal_get_proxy_cancast (MonoClass *klass)
 	mb->method->save_lmf = 1;
 
 #ifndef DISABLE_JIT
+
+	mono_remoting_marshal_init (); // register icalls
+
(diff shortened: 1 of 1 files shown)


jaykrell:

jaykrell:
@monojenkins build failed

Issue #12981, opened 2019-02-13
Transform.c:2856, condition `<disabled>' not met


jeromelaban:
When running a wasm binary under the interpreter with debugging enabled, the following error can occur:
```
* Assertion at /mnt/jenkins/workspace/test-mono-mainline-wasm/label/ubuntu-1804-amd64/mono/mini/interp/transform.c:2856, condition `<disabled>' not met
```

## Steps to Reproduce

1. In Visual Studio 2017/2019, open [this solution](https://github.com/nventive/Uno.Xaml-Controls-Gallery/blob/uno/XamlControlsGallery.Windows/XamlControlsGallery.sln) at commit https://github.com/nventive/Uno.Xaml-Controls-Gallery/commit/f1be090d0183f49c9ad697b90215e69324f99285.
2. Select the XamlControlsGallery.Wasm project
3. Ensure that the active configuration is `Debug`
4. Run the project with `Ctrl+F5`
5. Once the app has started click on any item in the left list

### Current Behavior
There's an assertion in the runtime's interpreter.

### Expected Behavior
A managed exception is raised. There's really an issue in this commit of the repo, where an NRE is raised.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows
[X] WebAssembly

**Version Used**: 5f65ce9f7dfa6404219c90b6b82ba586fabb2e1c

## Stacktrace

```
Please paste the stack trace here if available.
```

jeromelaban:
Here's a repro:
```
static void Main(string[] args) => MethodA(null);

internal static bool MethodA(object sender)
{
    if (sender == null) { return true; }
    return MethodA(sender);
}
```
the issue happens only in release configutation.

See here for the full repro: https://github.com/jeromelaban/Wasm.Samples/tree/master/Bug12981-InterpRecursion/BugInterpRecursion

kjpou1:
@jeromelaban Is this still an issue?

I have taken your sample above and created both a Debug version and a Release version here using full linkage and can not reproduce this.

Sample code from: https://github.com/jeromelaban/Wasm.Samples/tree/master/Bug12981-InterpRecursion/BugInterpRecursion

Release mode:

```
Loaded: System.Drawing.Common.dll
Loaded: WebAssembly.Net.WebSockets.dll
Loaded: System.Web.Services.dll
Loaded: WasmCoreObjects.dll
Loaded: WebAssembly.Bindings.dll
Loaded: WebAssembly.Net.Http.dll
[...]

PR #14098, opened 2019-04-17
[interp] check bounds when updating il_offset for seq points

lewurm:
It could be a tailcall that hasn't any opcodes after it, so just leave
it as-is in those cases.

Fixes https://github.com/mono/mono/issues/12981


BrzVlad:
I would have assumed that tailcall optimization should be disabled when under debugger since we are missing the seq point for exiting that method.

lewurm:
@BrzVlad updated.  could you have a look again please?

lewurm:
@monojenkins squash

PR #14100, opened 2019-04-17
[netcore] improve Windows support

EgorBo:
The default packages we recommend to install with cygwin don't contain any zip/unzip commands (`tar` doesn't support zip files) so I came up with a small python script 🙁 

The default values in the csproj allow to open System.Private.Corlib.csproj in an IDE without errors.

cc @filipnavara


netcore/unzip.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+import zipfile
+import sys
+
+print(sys.argv[1])
+zip_ref = zipfile.ZipFile(sys.argv[1], 'r')
+zip_ref.extractall()
+zip_ref.close()
```
jaykrell: Last line surely not needed.

mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -3,6 +3,15 @@
...
 <!--  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" /> -->
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <!-- Default values -->
+  <PropertyGroup Condition="'$(TargetsUnix)'=='' AND '$(TargetsOSX)'=='' AND '$(TargetsWindows)'==''">
+    <TargetsUnix>true</TargetsUnix>
+    <TargetsOSX>true</TargetsOSX>
+    <TargetsWindows>false</TargetsWindows>
+    <BuildArch>x64</BuildArch>
+    <OutputPath>bin/x64</OutputPath>
+    <FeaturePortableTimer>true</FeaturePortableTimer>
```
filipnavara: nit: This one can be defined unconditionally and removed from the Makefile.

jaykrell:
See also https://github.com/mono/website/pull/440

netcore/Makefile.am
```diff
@@ -8,14 +8,32 @@ NETCORETESTS_VERSION := 4.6.0-preview5.19209.12
 # Extracted MicrosoftNETCoreAppVersion from https://github.com/dotnet/coreclr/blob/master/eng/Versions.props#L13
 NETCOREAPP_VERSION := 3.0.0-preview5-27609-17
 
-NETCORESDK_FILE := dotnet-runtime-$(NETCOREAPP_VERSION)-@RID@.tar.gz
+if HOST_WIN32
+NETCORESDK_EXT = zip
+UNZIPCMD = python -c "import zipfile,sys; zipfile.ZipFile(sys.argv[1], 'r').extractall()"
```
jaykrell: I like the anonymous temporary. :)

jaykrell: Might as well use .zip+python as the one portable solution? :)

filipnavara: Only .tar.gz packages are available for Linux/macOS.

jaykrell: Ok. xz will be much smaller as well, than any of them.

mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
EgorBo: I'd leave makefile as is - as an example how to configure build. These default values are for OSX-64 only (also as an example)

filipnavara: I meant specifically the `<FeaturePortableTimer>true</FeaturePortableTimer>` line. It's used across all platforms. Both CoreRT and CoreCLR specify the `Feature*` definitions inside the .csproj and optionally condition them based on `Targets*` definition.

netcore/Makefile.am
filipnavara: @jaykrell True, but that's not our call. It downloads pre-packaged .NET Core runtime from https://github.com/dotnet/core-sdk/blob/master/README.md#installers-and-binaries and those are only available in one format per platform.

mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj
EgorBo: ah, ok, makes sense then

PR #14101, opened 2019-04-17
Add hybrid/cooperate suspend runtime support on Windows.

lateralusX:
Adding initial support for hybrid and cooperate suspend on Windows 32/64-bit platforms.

Fixes/adjustments needed in several areas:

Implementation/emulation of gcc's __builtin_unwind_init on MSVC needed by copy_stack_data function.

All wait methods have been extended with a cooperative alternative + handling interrupt under hybrid/cooperative suspend.

Methods that handles IO abort interruption on Windows (file, socket) have been updated to handle interrupt under hybrid/cooperative suspend.

Moved Windows implementations into w32handle.c instead of #ifdef on call site.

Reduced number of GC safe transition in a couple of Win32 file IO methods.

Added restore logic to Win32 last error when exiting GC safe mode. This is needed since we have code that depends on GetLastError that could be clobbered if exiting GC safe mode enter a wait. Also added last error restore logic to uninstall interrupt handler since it is commonly used right after exiting GC safe mode.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -2230,17 +2230,8 @@ ves_icall_System_Threading_WaitHandle_Wait_internal (gpointer *handles, gint32 n
 
 	for (;;) {
 
-#ifdef HOST_WIN32
-		MONO_ENTER_GC_SAFE;
-		DWORD const wait_result = (numhandles != 1)
-			? mono_win32_wait_for_multiple_objects_ex (numhandles, handles, waitall, timeoutLeft, TRUE, error)
-			: mono_win32_wait_for_single_object_ex (handles [0], timeoutLeft, TRUE);
-		ret = mono_w32handle_convert_wait_ret (wait_result, numhandles);
-		MONO_EXIT_GC_SAFE;
-#else
 		/* mono_w32handle_wait_multiple optimizes the case for numhandles == 1 */
 		ret = mono_w32handle_wait_multiple (handles, numhandles, waitall, timeoutLeft, TRUE, error);
-#endif /* HOST_WIN32 */
 
 		if (ret != MONO_W32HANDLE_WAIT_RET_ALERTED)
 			break;
@@ -2290,13 +2281,7 @@ ves_icall_System_Threading_WaitHandle_SignalAndWait_Internal (gpointer toSignal,
 
 	mono_thread_set_state (thread, ThreadState_WaitSleepJoin);
 	
-#ifdef HOST_WIN32
-	MONO_ENTER_GC_SAFE;
-	ret = mono_w32handle_convert_wait_ret (mono_win32_signal_object_and_wait (toSignal, toWait, ms, TRUE), 1);
-	MONO_EXIT_GC_SAFE;
-#else
 	ret = mono_w32handle_signal_and_wait (toSignal, toWait, ms, TRUE);
-#endif
 	
 	mono_thread_clr_state (thread, ThreadState_WaitSleepJoin);
 
@@ -4958,7 +4943,21 @@ mono_alloc_special_static_data_free (GHashTable *special_static_fields)
 }
 
 #ifdef HOST_WIN32
-static void CALLBACK dummy_apc (ULONG_PTR param)
+static void
+flush_thread_interrupt_queue (void)
+{
+	/* Consume pending APC calls for current thread.*/
+	/* Since this function get's called from interrupt handler it must use a direct */
+	/* Win32 API call and can't go through mono_coop_win32_wait_for_single_object_ex */
+	/* or it will detect a pending interrupt and not entering the wait call needed */
+	/* to consume pending APC's.*/
+	MONO_ENTER_GC_SAFE;
+	WaitForSingleObjectEx (GetCurrentThread (), 0, TRUE);
+	MONO_EXIT_GC_SAFE;
+}
+#else
+static void
+flush_thread_interrupt_queue (void)
 {
 }
 #endif
@@ -4997,12 +4996,7 @@ mono_thread_execute_interruption (MonoExceptionHandle *pexc)
 	MonoThreadObjectHandle sys_thread;
 	sys_thread = mono_thread_current_handle ();
 
-	/* this will consume pending APC calls */
-#ifdef HOST_WIN32
-	MONO_ENTER_GC_SAFE;
-	mono_win32_wait_for_single_object_ex (GetCurrentThread (), 0, TRUE);
-	MONO_EXIT_GC_SAFE;
-#endif
+	flush_thread_interrupt_queue ();
 
 	/* Clear the interrupted flag of the thread so it can wait again */
 	mono_thread_info_clear_self_interrupt ();
--- a/mono/metadata/w32file-win32.c
+++ b/mono/metadata/w32file-win32.c
@@ -58,7 +58,7 @@ mono_w32file_create(const gunichar2 *name, guint32 fileaccess, guint32 sharemode
 {
 	gpointer res;
 	MONO_ENTER_GC_SAFE;
-	res = CreateFile (name, fileaccess, sharemode, NULL, createmode, attrs, NULL);
+	res = CreateFileW (name, fileaccess, sharemode, NULL, createmode, attrs, NULL);
 	MONO_EXIT_GC_SAFE;
 	return res;
 }
@@ -78,9 +78,15 @@ mono_w32file_delete (const gunichar2 *name)
 {
 	gboolean res;
 	MONO_ENTER_GC_SAFE;
-	res = DeleteFile (name);
+	res = DeleteFileW (name);
 	MONO_EXIT_GC_SAFE;
 	return res;
+}
+
+// See win32_wait_interrupt_handler for details.
+static void
+win32_io_interrupt_handler (gpointer ignored)
+{
 }
 
 gboolean
@@ -88,14 +94,33 @@ mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *b
 {
 	gboolean res;
 	MonoThreadInfo *info = mono_thread_info_current ();
+	gboolean alerted = FALSE;
 
-	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
+	if (info) {
+		mono_thread_info_install_interrupt (win32_io_interrupt_handler, NULL, &alerted);
+		if (alerted) {
+			SetLastError (ERROR_OPERATION_ABORTED);
+			*win32error = ERROR_OPERATION_ABORTED;
+			return FALSE;
+		}
+		mono_win32_enter_blocking_io_call (info, handle);
+	}
+
 	MONO_ENTER_GC_SAFE;
-	res = ReadFile ((HANDLE)handle, buffer, numbytes, (PDWORD)bytesread, NULL);
+	if (info && mono_thread_info_is_interrupt_state (info)) {
+		res = FALSE;
+		SetLastError (ERROR_OPERATION_ABORTED);
+	} else {
+		res = ReadFile (handle, buffer, numbytes, (PDWORD)bytesread, NULL);
+	}
 	if (!res)
 		*win32error = GetLastError ();
 	MONO_EXIT_GC_SAFE;
-	mono_win32_leave_blocking_io_call (info, (HANDLE)handle);
+
+	if (info) {
+		mono_win32_leave_blocking_io_call (info, handle);
+		mono_thread_info_uninstall_interrupt (&alerted);
+	}
 
 	return res;
 }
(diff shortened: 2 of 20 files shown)


jaykrell:
Nice overall, just nit picking.

mono/metadata/w32file-win32.c
```diff
@@ -235,33 +282,25 @@ guint32
...
-	MONO_ENTER_GC_SAFE;
-	find_handle = FindFirstFile (name, &find_data);
-	MONO_EXIT_GC_SAFE;
-
-	if (find_handle == INVALID_HANDLE_VALUE)
-		return INVALID_FILE_ATTRIBUTES;
+	res = GetFileAttributes (name);
+	if (res == INVALID_FILE_ATTRIBUTES && GetLastError () == ERROR_SHARING_VIOLATION) {
+		find_handle = FindFirstFile (name, &find_data);
+		if (find_handle != INVALID_HANDLE_VALUE) {
```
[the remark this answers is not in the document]
lateralusX: Sure, just copied pasted the original implementation.

mono/metadata/w32file-win32.c
```diff
@@ -105,14 +132,34 @@ mono_w32file_write (gpointer handle, gconstpointer buffer, guint32 numbytes, gui
...
+	}
 
-	mono_win32_enter_blocking_io_call (info, (HANDLE)handle);
 	MONO_ENTER_GC_SAFE;
-	res = WriteFile ((HANDLE)handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
+	if (info && mono_thread_info_is_interrupt_state (info)) {
+		res = FALSE;
+		SetLastError (ERROR_OPERATION_ABORTED);
+	} else {
+		res = WriteFile ((HANDLE)handle, buffer, numbytes, (PDWORD)byteswritten, NULL);
```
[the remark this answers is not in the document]
lateralusX: Copy of original implementation, but sure we can drop the cast.

mono/metadata/threads.c
```diff
@@ -4958,8 +4943,17 @@ mono_alloc_special_static_data_free (GHashTable *special_static_fields)
 }
 
 #ifdef HOST_WIN32
-static void CALLBACK dummy_apc (ULONG_PTR param)
+static inline void
+flush_thread_interrupt_queue ()
+{
```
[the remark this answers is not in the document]
lateralusX: Will fix.

mono/utils/mono-os-wait-win32.c
```diff
@@ -14,53 +14,177 @@
...
+}
+
+DWORD
+mono_coop_win32_wait_for_single_object_ex (HANDLE handle, DWORD timeout, BOOL alertable)
+{
+	return win32_wait_for_single_object_ex (handle, timeout, alertable, TRUE);
+}
+
+static inline DWORD
+win32_wait_for_multiple_objects_ex_interrupt_checked (MonoThreadInfo *info, DWORD count, CONST HANDLE *handles, BOOL waitAll, DWORD timeout, BOOL alertable)
```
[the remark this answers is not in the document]
lateralusX: Old habbit...

mono/metadata/w32file-win32.c
```diff
@@ -83,19 +83,46 @@ mono_w32file_delete (const gunichar2 *name)
...
+	;
+}
+
 gboolean
 mono_w32file_read(gpointer handle, gpointer buffer, guint32 numbytes, guint32 *bytesread, gint32 *win32error)
 {
 	gboolean res;
 	MonoThreadInfo *info = mono_thread_info_current ();
+	gboolean alerted;
+
```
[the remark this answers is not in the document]
lateralusX: Me too, will fix.

lateralusX:
First run all Windows configs have pass. Next round will be hybrid suspend + review feedback.

mono/metadata/w32file-win32.c
```diff
@@ -83,19 +83,46 @@ mono_w32file_delete (const gunichar2 *name)
 	return res;
 }
 
+// See win32_wait_interrupt_handler for details.
+static void
+win32_io_interrupt_handler (gpointer ignored)
+{
+	;
```
jaykrell: No need for the line and semicolon.

mono/metadata/w32file-win32.c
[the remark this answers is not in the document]
jaykrell: `WIN32_FIND_DATA`, `GetFileAttributes` etc.

There are multiple contradictory schools of thought here.

1. This a Win9x abstraction layer. Obsolete. Use the NT names directly.
2. These are the nice names. They aren't right because of the abstraction layer, but you can still use them.
3. Code should be as ifdef-independent as possible, so use the NT names directly.
4. Maybe Win9x/A isn't so dead. Very recently A can even mean UTF8. It almost seems good and we can now forget the entire wchar 30 year adventure (unless we write an actual text renderer).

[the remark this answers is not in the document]
lateralusX: Agree, I started out using TCHAR together with Win32 in late 90's and got stuck to using the old API names since then. But since all our string parameters are currently explicit gunichar2 I agree that we should also be explicit in our API calls as well, using the W version of all methods instead of relying on compability macros.

mono/utils/mono-threads-coop.c
```diff
@@ -35,8 +35,9 @@
 #endif
 
 #ifdef _MSC_VER
-// TODO: Find MSVC replacement for __builtin_unwind_init
-#define SAVE_REGS_ON_STACK g_assert_not_reached ();
+// __builtin_unwind_init not available under MSVC but equivalent implementation is done using
+// copy_stack_data_internal_win32_wrapper.
```
jaykrell: This should probably be RtlVirtualUnwind, but let's not solve it today. :)

jaykrell: Oh maybe this is the topic today. I'll read more.

mono/utils/mono-threads-coop.c
```diff
@@ -197,6 +198,53 @@ copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
...
+
+// Defined in win64.asm
+G_EXTERN_C void
+copy_stack_data_internal_win32_wrapper (MonoThreadInfo *, MonoStackData *, MonoBuiltinUnwindInfo *, CopyStackDataFunc);
+#else
+typedef struct {
+	host_mgreg_t gregs [4];
+} MonoBuiltinUnwindInfo;
+
+__declspec(naked) void
```
jaykrell: Maybe an explicit `__cdecl` or `__fastcall` or `__stdcall`? Or maybe the compiler default __cdecl is inescapable w/o massive code churn. Just that most code is calling convention-agnostic, and this probably is not. I admit I'm not reading it closely yet.

jaykrell: Might be nice to use jmp_buf and setjmp instead. Might.
Or explicitly list what is volatile/non-volatile on NT/x86.
That documentation is difficult to find.

mono/utils/win64.asm
```diff
@@ -38,6 +38,36 @@ __mono_current_ip:
 
 mono_context_get_current endP
 
+copy_stack_data_internal_win32_wrapper PROC PUBLIC
+;rcx MonoThreadInfo *
+;rdx MonoStackData *
+;r8 MonoBuiltinUnwindInfo *
+;r9 CopyStackDataFunc
+
```
jaykrell: Probably should comment, that this is saving nonvolatile registers.

mono/utils/mono-threads-coop.c
```diff
@@ -197,6 +198,53 @@ copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
 	state->gc_stackdata_size = stackdata_size;
 }
 
+#ifdef _MSC_VER
+typedef void (*CopyStackDataFunc)(MonoThreadInfo *, MonoStackData *, gconstpointer, gconstpointer);
+
+#ifdef TARGET_AMD64
+typedef struct {
+	__m128d fregs [10];
```
jaykrell: Maybe an extra comment, that this is specifically nonvolatiles.

mono/utils/mono-threads-coop.c
```diff
@@ -197,6 +198,53 @@ copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
...
+{
+	__asm {
+		mov edx, dword ptr [esp + 0Ch]
+		mov dword ptr [edx + 00h], ebx
+		mov dword ptr [edx + 04h], esi
+		mov dword ptr [edx + 08h], edi
+		mov dword ptr [edx + 0Ch], ebp
+
+		mov edx, dword ptr [esp + 10h]
+		jmp edx
```
jaykrell: Maybe a comment, that func takes same number of params as copy_stack_data_internal_win32_wrapper.
Or maybe it is blindingly obviously -- you wouldn't tailcall otherwise.

Maybe just add a comment to the last line "tailcall". but again, maybe blindingly obvious.
And there sure is a nice set of simplifications  -- no frame setup or teardown.

mono/utils/win64.asm
```diff
@@ -38,6 +38,36 @@ __mono_current_ip:
...
+	mov qword ptr [r8 + 0A0h], rbx
+	mov qword ptr [r8 + 0A8h], rsi
+	mov qword ptr [r8 + 0B0h], rdi
+	mov qword ptr [r8 + 0B8h], r12
+	mov qword ptr [r8 + 0C0h], r13
+	mov qword ptr [r8 + 0C8h], r14
+	mov qword ptr [r8 + 0D0h], r15
+	mov qword ptr [r8 + 0D8h], rbp
+
+	jmp r9
```
jaykrell: maybe a comment "; tailcall" or maybe this is the most obvious line of assembly ever :)

jaykrell: Oh, and that you are passing through the parameters. All quite subtle and simple at the same time.

mono/utils/mono-threads.h
```diff
@@ -814,6 +814,26 @@ mono_win32_interrupt_wait (PVOID thread_info, HANDLE native_thread_handle, DWORD
...
 mono_win32_abort_blocking_io_call (THREAD_INFO_TYPE *info);
+
+#define W32_DEFINE_LAST_ERROR_RESTORE_POINT \
+	DWORD _last_error_restore_point = GetLastError ();
+
+#define W32_UPDATE_LAST_ERROR_RESTORE_POINT \
+	_last_error_restore_point = GetLastError ();
+
+#define W32_RESTORE_LAST_ERROR_FROM_RESTORE_POINT \
+	do { \
```
jaykrell: This could be an expression instead of statement, but ok.

mono/utils/win64.asm
jaykrell: Kinda seems like a job for RtlCaptureContext, but your code does seem smaller and sufficient.

mono/utils/mono-threads.h
```diff
@@ -814,6 +814,26 @@ mono_win32_interrupt_wait (PVOID thread_info, HANDLE native_thread_handle, DWORD
...
+		/* Only restore if changed to prevent unecessary writes. */ \
+		if (GetLastError () != _last_error_restore_point) \
+			SetLastError (_last_error_restore_point); \
+	} while (0)
+
+#else
+
+#define W32_DEFINE_LAST_ERROR_RESTORE_POINT
```
jaykrell: I like to put `/* nothing */` as the apparent value of the define, to help some naïve readers, but it is correct either way.

jaykrell:
Now just to six more CI lanes? :)

jaykrell:
Again very good, I just nit-picky.
Putting explicit `__cdecl` on the x86 asm is maybe the only significant point.
The compiling mono with default `__fastcall` or `__stdcall` might be catastrophic in other places anyway.

mono/utils/mono-threads-coop.c
lateralusX: The dumped registers are inline with what you get on mingw gcc x86 implementation of __builtin_unwind_init, that's the only documentation that I could find regarding that.

jaykrell: x86

https://docs.microsoft.com/en-us/cpp/cpp/argument-passing-and-naming-conventions?view=vs-2019


The compiler generates prolog and epilog code to save and restore the ESI, EDI, EBX, and EBP registers, if they are used in the function.

So those are the non-volatiles.
And it matches your code.

jaykrell: And amd64:

https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019

The registers RBX, RBP, RDI, RSI, RSP, R12, R13, R14, and R15 are considered nonvolatile and must be saved and restored by a function that uses them.

which matches your code, except for rsp, which is maybe special.

lateralusX: Jepp, all inline with __builtin_unwind_init implementation on Windows x86/x64.

lateralusX:
First run of hybrid suspend on CI for all Windows configuration is done!

**Windows x64**, all CI test pass on hybrid suspend.
**Windows x64 C++**, all CI test pass on hybrid suspend.

**Windows x64 FullAOT**, only one failing test case, timeout in finalizer-exit:

finalizer-exit.exe
=============== finalizer-exit.exe.stdout ===============
Set ID: foo
finalizer thread ID: foo

=============== finalizer-exit.exe.stderr ===============

**Windows i386**, 3 failures, all looks related to the same issue:

MonoTests.runtime.thread-suspend-suspended.exe
MonoTests.test-sgen-regular-ms-simple.sgen-new-threads-collect.exe
MonoTests.test-sgen-regular-ms-simple.finalizer-wait.exe

Cannot transition thread 000001A8 from SELF_SUSPENDED with REQUEST_PULSE

lateralusX:
Run with cooperate suspend on CI, all test pass.

lateralusX:
Analyzed failure in supend-self-suspended.exe on 32-bit Windows and the identified problem, Cannot transition thread 000001A8 from SELF_SUSPENDED with REQUEST_PULSE, could happened in the following scenarion (emulated in debugger).

* Have main thread run its start_gc.Set () call but freeze before entering GC Safe of its next wait, finished_gc.WaitOne ()
* Run GC thread and let it do first phase, this will wait for main thread to suspend.
* Switch to main thread and resume it, this will make it self suspend on enter GC safe. NOTE, entering GC safe will call into OS WaitForSingleObject to wait for resume.
* GC thread will now look at the threads and determine if its in a critical region or not. On wow64 this means doing some extra check, mono_threads_platform_in_critical_region, that could fail, especially in hybrid suspend since we have not suspended the thread in the case (it's self suspended or on its way towards self suspend, but still running code).

If it fails due to that check stw will move over to pulse resume the thread but since it is already self suspend, it will hit the assert. The fix is to not do the wow64 critical check, mono_threads_platform_in_critical_region, if thread is not preemptive suspended. This is needed since the call to GetThreadContext is not guaranteed to get a valid context on a running thread.

lateralusX:
Run with hybrid suspend on CI, all Windows test lanes pass.

lateralusX:
@monojenkins rebuild failed

PR #14106, opened 2019-04-17
[aot] Cosmetical follow-up for https://github.com/mono/mono/pull/14043

lewurm:
https://github.com/mono/mono/pull/14043

If we ever should use FullAOT on a non-Darwin/arm32 target then this is needed.


lewurm:
this is wrong, we really need to emit a call here.

PR #14107, opened 2019-04-17
Update documentation for --with-large-heap

jneedhamspkt:
This flag only applies to the Boehm GC, but it's not obvious until you dig down into the source code. I thought that an edit like this to the readme would help prevent future confusion for others, given that SGen the default.


marek-safar:
approve

marek-safar:
Thank you!

PR #14108, opened 2019-04-17
[netcore] Fix the tests.

vargaz:


filipnavara:
Would not there be the same issue with `Monitor.cs`? Both `GC.cs` and `Monitor.cs` have no partial-class counter-parts in the shared CoreLib partition. Thus `Monitor` is missing the proper `public` modifier too.

EgorBo:
@filipnavara actually Monitor has two files in System.Private.CorLib: https://github.com/mono/mono/blob/master/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj#L256 and https://github.com/mono/mono/blob/master/mcs/class/System.Private.CoreLib/System.Private.CoreLib.csproj#L200 (one of them has `public`)

filipnavara:
@EgorBo Oh, I looked only in the `netcore/System.Private.CoreLib/shared` folder. Nevertheless I think it's a bit dangerous to leave the modifiers out here because it's possible that the other one will go away.

vargaz:
@monojenkins build failed

PR #14113, opened 2019-04-17
[wasm] Fix the types of the generated function declarations in the pinvoke table.

vargaz:


lewing:
@monojenkins build failed

PR #14117, opened 2019-04-17
Fix start-compiler-server

kg:
This bit-rotted at some point.


kg:
@monojenkins build failed

PR #14120, opened 2019-04-18
[wasm][sdk] Fixes to the sdk packages

kjpou1:
The PR targets the sdk project build to make sure they run on windows and VS 2019

- Package the wasm bcl and wasm bcl Facades correctly.
- Move the build props and targets file to `netstandard2.0` build directory.
- Place `framework` .props file properly for build
- Place `runtime` .props file properly for build
- Modify the WasmLinkAssemblies Task to escape the windows path separator.  - There are problems executing `mononlinker.exe` on windows if the windows path separators are not escaped. 

**Note:**
* Tested on VS 2019 project on windows.
* VS Mac 2019 has problems executing these project types.  VS Mac 2017 did not have problems.

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -400,6 +400,7 @@ clean-sdk:
 	$(RM) -r sdk/**/obj
 	$(RM) -r sdk/**/**/bin
 	$(RM) -r sdk/**/**/obj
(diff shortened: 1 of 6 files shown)


kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

PR #14128, opened 2019-04-18
[netcore] Throw the correct exception for byref returns returning null, .net core throws a nullref.

vargaz:


marek-safar:
Ugh, this is quite ugly. @jkotas do you know what was the rationale to make this to throw NullReferenceException ?

EgorBo:
@marek-safar see https://github.com/dotnet/coreclr/issues/24087 and the related links

vargaz:
This makes sense since the method returns a null byref, and the exception is caused because Invoke () tries to convert it into a value.

mono/metadata/marshal-ilgen.c
```diff
@@ -1485,7 +1485,16 @@ emit_invoke_call (MonoMethodBuilder *mb, MonoMethod *method,
 	}
 
 	if (sig->ret->byref) {
+		int pos;
+
 		/* perform indirect load and return by value */
+#ifdef ENABLE_NETCORE
+		mono_mb_emit_byte (mb, CEE_DUP);
+		pos = mono_mb_emit_branch (mb, CEE_BRTRUE);
+		mono_mb_emit_exception_full (mb, "Mono", "NullByRefReturnException", NULL);
```
MichalStrehovsky: Will this `throw` break the user into the debugger if they have first chance exceptions turned on?

In .NET Native we had to make this return a special sentinel value that we check for and then do the throw with the correct exception because multiple first chance exceptions tend to be confusing to users.

filipnavara: > Will this throw break the user into the debugger if they have first chance exceptions turned on?

I think it will. If not here then definitely at the site in https://github.com/mono/mono/pull/14128/files#diff-04cafe842355fb36dd025dfc83aa13be that converts it to regular `NullReferenceException`.

> In .NET Native we had to make this return a special sentinel value that we check for and then do the throw with the correct exception because multiple first chance exceptions tend to be confusing to users.

Ha, I had exactly the same idea how to implement this.

vargaz:
@monojenkins build failed

PR #14132, opened 2019-04-18
[2019-02][monodroid] Remove File.Copy xUnit tests

alexischr:
a) CopyFileWithData_MemberData() is crashing the Xamarin .Android test runner

b) Some other failures that may be real issues (but also these are new tests on a new platform so not necessarily a regression)


alexischr:
-> https://github.com/mono/mono/pull/14133

PR #14133, opened 2019-04-19
[monodroid] Exclude System.IO.Tests.File_Copy_str_str CoreFX testsuite

lambdageek:
The test case CopyFileWithData_MemberData with input of size 1024 * 1024 * 10
causes the Xamarin.Android.Bcl_Tests to crash for some reason.  Temporarily
exclude this entire set of tests from running.

Also disable System.IO.Tests.FileInfo_CopyTo_str because it derives from the
above test class.


lambdageek:
This isn't how I tested it locally (I just edited `corlib_xtest.dll.soures` directly), but I think this ought to do it the "right" way.

We should also open an issue to turn this stuff back on once we understand why the 10MB copy test caused a crash on android.

marek-safar:
@monojenkins backport master

PR #14135, opened 2019-04-19
[aot] partial revert of #14043

lewurm:
We need to emit a real call here. On arm/android we would emit this now:
```asm
ldr pc, =label
.ltorg         # place address of `label` here.
```
which is a jump to `label`, not a call.

We need a different solution to make larger assemblies work, but for now
I revert it so the `2019-02` intergration is unblocked.

Partial revert of #14043


lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2019-04

PR #14136, opened 2019-04-19
[wasm] Fix NRE WebAssembly.Core.Array during initialization

kjpou1:
- Add test for WebAssembly.Core.Array
- Add tests for WebAssembly.Core.Function Call and Apply.


kjpou1:
@monojenkins build failed

PR #14145, opened 2019-04-20
Add CI test coverage for compiler server

kg:
Also updates start-compiler-server to be a bit better.

This test should validate:
That compiling culevel.exe with ENABLE_COMPILER_SERVER set or unset succeeds
Compiling culevel.exe with ENABLE_COMPILER_SERVER produces an entry in the compiler server log
Compiling culevel.exe with ENABLE_COMPILER_SERVER produces a compiler server log with a successful compile in it

This is not a perfectly robust test but I think it's probably good enough for now.

Needs a little manual verification before merging.


kg:
OK I think this is ready and we can make a lane for it.

scripts/ci/run-jenkins.sh
```diff
@@ -359,6 +359,7 @@ elif [[ ${CI_TAGS} == *'mcs-compiler'* ]];             then ${MONO_REPO_ROOT}/sc
 elif [[ ${CI_TAGS} == *'mac-sdk'* ]];                  then ${MONO_REPO_ROOT}/scripts/ci/run-test-mac-sdk.sh;
 elif [[ ${CI_TAGS} == *'helix-tests'* ]];              then ${MONO_REPO_ROOT}/scripts/ci/run-test-helix.sh;
 elif [[ ${CI_TAGS} == *'make-install'* ]];             then ${MONO_REPO_ROOT}/scripts/ci/run-test-make-install.sh;
+elif [[ ${CI_TAGS} == *'compiler-server'* ]];          then ${MONO_REPO_ROOT}/scripts/ci/run-test-compiler-server.sh;
 elif [[ ${CI_TAGS} == *'no-tests'* ]];                 then echo "Skipping tests.";
```
vargaz: Shouldn't this be compiler-server-test, so 'compiler-server' can be used to enable the compiler server on CI ?

kg: good idea, though right now the autogen/make vars are --enable-compiler-server and ENABLE_COMPILER_SERVER

kg:
@monojenkins build failed

scripts/ci/run-test-compiler-server.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
```
marek-safar: Do we need extra test step? I though we will enable it everywhere for normal build

kg: It's not enabled everywhere right now, and this also verifies that it works so it makes sense for it to be an extra test step

The compiler server is designed to silently fail in like a dozen scenarios so we'll always need a dedicated test for it

kg:
@monojenkins build test-mono-pull-request-amd64-compiler-server

kg:
@monojenkins build Linux x64 - compiler server

kg:
@monojenkins build compiler server

kg:
@monojenkins build compiler server

kg:
@monojenkins build compiler server

kg:
@monojenkins build compiler server

mcs/build/profiles/build.make
```diff
@@ -147,8 +147,8 @@ export VBCS_LOCATION
 
 start-compiler-server:
 	echo Attempting to start compiler server...
-	./build/start-compiler-server.sh '$(realpath $(topdir))' '$(realpath $(topdir)/build)/compiler-server.log' '$(COMPILER_SERVER_PIPENAME)'
+	$(topdir)/build/start-compiler-server.sh '$(realpath $(topdir))' '$(realpath $(topdir)/build)/compiler-server.log' '$(COMPILER_SERVER_PIPENAME)'
```
lewing: Isn't this reverting #14117

kg: this might be a bad line merge, i'll check. thanks.

lewing:
Is there anything that is prohibiting us from enabling the compiler server as part of the normal build?

kg:
> Is there anything that is prohibiting us from enabling the compiler server as part of the normal build?
Due to GC performance and other hard-to-identify issues, it's only faster than non-compiler-server at -j 1. At anything above -j 1 on my test machines it's break-even or slower

marek-safar:
@kg could you enable it everywhere in PR so we can see if CI will experience similar behaviour or not.

/cc @directhex @akoeplinger

mcs/build/profiles/build.make
lewing: What is the status on this?

kg: I genuinely don't know. I think topdir is correct and the bitrot was in the realpath part

kg:
I need to rebase this on top of enable-compiler-server now, so I made sure that's going to work. We'll have to wait for that one before we can land this CI coverage

lewing:
@monojenkins build failed

PR #14047, opened 2019-04-15
Work towards removing JIT icall hashing.

jaykrell:
Based on:
 https://github.com/mono/mono/pull/11941#issuecomment-449157323

Hash lookups can often/always be replaced
by directly referencing a struct,
or by indexing into a very dense array.

This PR is a stepping stone. The steps can be made smaller or larger.
There is definitely more to do.

PR #14149, opened 2019-04-21
Small step toward reduction/elimination of JIT icall hashing.

jaykrell:
Ultimate goal is to significantly reduce or maybe completely
eliminate the hashing of JIT icalls.

It is not a small change.

This is a small digestable albeit tedious step toward that.

The "names" of opcode emulation icalls are needed for
token pasting and/or struct fields. They are needed without quotes.

This PR removes the quotes.
And then for now, just always restores them.

This is extracted from https://github.com/mono/mono/pull/14047
as a step toward making it easier to review.

### DIFF
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -695,6 +695,11 @@ register_opcode_emulation (int opcode, const char *name, MonoMethodSignature *si
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #14150, opened 2019-04-21
Ignore NULL in mono_marshal_free_dynamic_wrappers and mono_free_method.

jaykrell:


jaykrell:
@monojenkins build failed

vargaz:
Why is this needed ? These functions should probably assert instead on NULL.

jaykrell:
"Needed", not strongly. I was going to introduce a call that "needed" one of them, but ended up not.
In general though, "free" functions can and should silently ignore null.
Many already do.
For example `free` and C++ `delete.
It isn't worth the risk of accidentally missing a check in the caller.

CoffeeFlux:
This seems reasonable to me. I think the main argument for not ignoring NULL would be that it could potentially catch weird, legitimate errors, but I don't think that's the responsibility of this kind of function, and by ignoring NULL you potentially avoid silly/harmless issues. Should do one or the other though: either assert or ignore.

jaykrell:
@monojenkins build failed

lambdageek:
@monojenkins build

jaykrell:
@monojenkins build failed

PR #14151, opened 2019-04-21
Small steps toward JIT icall hash reduction/elimination.

jaykrell:
This is related to https://github.com/mono/mono/pull/14047.

In particular it is useful to have mono_register_jit_icall
nearer the creation of trampolines, so we can specify the related data or enum.

Alternate would be to grow MonoTrampInfo to pass the data through.
 Which I hadn't thought of before and would be a way to
 wastefully decrease the diff.

This restructuring also lets us avoid memory allocation/free
of the trampoline list -- register inline with creation instead
of returning the list and then registering and freeing.

And while we are here, save some string allocs/copies.

### DIFF
--- a/mono/mini/exceptions-amd64.c
+++ b/mono/mini/exceptions-amd64.c
@@ -919,29 +919,45 @@ mono_arch_handle_altstack_exception (void *sigctx, MONO_SIG_HANDLER_INFO_TYPE *s
 }
 
 #ifndef DISABLE_JIT
+
+static
 GSList*
-mono_amd64_get_exception_trampolines (gboolean aot)
+mono_amd64_get_exception_trampoline (gboolean rethrow, gboolean corlib, gboolean llvm_abs, gboolean resume_unwind,
+				     const char *tramp_name, gboolean aot, gboolean reg, GSList* tramps)
 {
 	MonoTrampInfo *info;
+
+	(void)get_throw_trampoline (&info, rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, FALSE);
+
+	// Register the trampolines or build up a list.
+	if (reg) {
+		mono_register_jit_icall (info->code, tramp_name, NULL, TRUE);
+		mono_tramp_info_register (info, NULL);
+		return NULL;
+	}
+	return g_slist_prepend (tramps, info);
+}
+
+// Not taking a string directly here will be useful in removing JIT icall hashing.
+#define mono_amd64_get_exception_trampoline(rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, reg, tramps) \
+ (mono_amd64_get_exception_trampoline ((rethrow), (corlib), (llvm_abs), (resume_unwind), #tramp_name, (aot), (reg), (tramps)))
+
+GSList*
+mono_amd64_get_exception_trampolines (gboolean aot, gboolean reg)
+{
 	GSList *tramps = NULL;
 
 	/* LLVM needs different throw trampolines */
-	get_throw_trampoline (&info, FALSE, TRUE, FALSE, FALSE, "llvm_throw_corlib_exception_trampoline", aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
-
-	get_throw_trampoline (&info, FALSE, TRUE, TRUE, FALSE, "llvm_throw_corlib_exception_abs_trampoline", aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
-
-	get_throw_trampoline (&info, FALSE, TRUE, TRUE, TRUE, "llvm_resume_unwind_trampoline", aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
-
-	return tramps;
+	//                                            rethrow corlib llvm_abs resume_unwind
+	tramps = mono_amd64_get_exception_trampoline (  FALSE,  TRUE,   FALSE,        FALSE, llvm_throw_corlib_exception_trampoline,     aot, reg, tramps);
+	tramps = mono_amd64_get_exception_trampoline (  FALSE,  TRUE,    TRUE,        FALSE, llvm_throw_corlib_exception_abs_trampoline, aot, reg, tramps);
+	return   mono_amd64_get_exception_trampoline (  FALSE,  TRUE,    TRUE,         TRUE, llvm_resume_unwind_trampoline,	         aot, reg, tramps);
 }
 
 #else
 
 GSList*
-mono_amd64_get_exception_trampolines (gboolean aot)
+mono_amd64_get_exception_trampolines (gboolean aot, gboolean reg)
 {
 	g_assert_not_reached ();
 	return NULL;
@@ -952,7 +968,6 @@ mono_amd64_get_exception_trampolines (gboolean aot)
 void
 mono_arch_exceptions_init (void)
 {
-	GSList *tramps, *l;
 	gpointer tramp;
 
 	if (mono_ee_features.use_aot_trampolines) {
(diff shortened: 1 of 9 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/mini/exceptions-arm.c
```diff
@@ -385,28 +385,42 @@ mono_arch_get_throw_corlib_exception (MonoTrampInfo **info, gboolean aot)
...
 	MonoTrampInfo *info;
-	GSList *tramps = NULL;
 
-	/* LLVM uses the normal trampolines, but with a different name */
-	get_throw_trampoline (168, TRUE, FALSE, FALSE, FALSE, "llvm_throw_corlib_exception_trampoline", &info, aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
-	
-	get_throw_trampoline (168, TRUE, FALSE, TRUE, FALSE, "llvm_throw_corlib_exception_abs_trampoline", &info, aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
+	(void)get_throw_trampoline (168, corlib, rethrow, llvm, resume_unwind, tramp_name, &info, aot, FALSE);
```
kg: this should be a named constant instead of a magic number

mono/mini/exceptions-arm64.c
```diff
@@ -287,28 +287,40 @@ mono_arch_get_throw_corlib_exception (MonoTrampInfo **info, gboolean aot)
...
-	GSList *tramps = NULL;
 
-	/* LLVM uses the normal trampolines, but with a different name */
-	get_throw_trampoline (256, TRUE, FALSE, FALSE, FALSE, "llvm_throw_corlib_exception_trampoline", &info, aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
-	
-	get_throw_trampoline (256, TRUE, FALSE, TRUE, FALSE, "llvm_throw_corlib_exception_abs_trampoline", &info, aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
+	// Register the trampolines or build up a list.
+	(void)get_throw_trampoline (256, corlib, rethrow, llvm, resume_unwind, tramp_name, &info, aot, FALSE);
```
kg: named constant here too please

mono/mini/exceptions-amd64.c
```diff
@@ -919,29 +919,45 @@ mono_arch_handle_altstack_exception (void *sigctx, MONO_SIG_HANDLER_INFO_TYPE *s
...
 
-	/* LLVM needs different throw trampolines */
-	get_throw_trampoline (&info, FALSE, TRUE, FALSE, FALSE, "llvm_throw_corlib_exception_trampoline", aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
+	(void)get_throw_trampoline (&info, rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, FALSE);
 
-	get_throw_trampoline (&info, FALSE, TRUE, TRUE, FALSE, "llvm_throw_corlib_exception_abs_trampoline", aot, FALSE);
-	tramps = g_slist_prepend (tramps, info);
+	// Register the trampolines or build up a list.
+	if (reg) {
```
kg: Will anything bad happen if this is called for the same trampoline multiple times with reg set each time?

mono/mini/exceptions-arm.c
jaykrell: I don't disagree, but that is what it already was.

mono/mini/exceptions-arm64.c
jaykrell: Ditto. I don't disagree, but that is what it already was.

mono/mini/exceptions-amd64.c
jaykrell: I admit, I didn't really think about it. This change is meant to basically be a no-op, but it is preparation for another change, or rather whittling away at another change.

It is true though, this kinda increases/changes the exposure of something.
That previously was inlined elsewhere.

I am inlined not to worry about.

And more so, in the later change, we do check for repeated registration, of JIT icalls.
And they do occur. But I assert, in that other change, equivalency of the reptition.
And I worked through what repitition does occur.

I think currently, last registration wins, and that is ok.
As long as there is no race -- registration occurs before use.
Which at least amd64 tries to do.

jaykrell: See here for the direction this is going and please accept this as a small intermediate step that doesn't really change anything, but makes later diffs smaller:

https://github.com/mono/mono/pull/14047/files#diff-eba0fd0052bf3ec9dd60f2f3e90f19e1R9150

vargaz:
Don't understand what this is trying to achieve. The old code built up a list of trampoline infos, and either registered it, or emitted it for aot. What was the problem with that ?

jaykrell:
Yeah sorry I kinda answered that but could be clearer.

The point of this change is to reduce the diff/noise of a later larger change (https://github.com/mono/mono/pull/14047)

There is also micro optimization, but that is besides the point and I can undo.

The real point, is that later this line:

```
 #define mono_amd64_get_exception_trampoline(rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, reg, tramps) \ 
 (mono_amd64_get_exception_trampoline ((rethrow), (corlib), (llvm_abs), (resume_unwind), #tramp_name, (aot), (reg), (tramps))) 
```

can be changed to e.g., watch the `tramp_name` part:

```
#define mono_amd64_get_exception_trampoline(rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, reg, tramps) \ 
 (mono_amd64_get_exception_trampoline ((rethrow), (corlib), (llvm_abs), (resume_unwind), &mono_jit_icall_info.tramp_name, (aot), (reg), (tramps))) 
```

or:

```
#define mono_amd64_get_exception_trampoline(rethrow, corlib, llvm_abs, resume_unwind, tramp_name, aot, reg, tramps) \ 
 (mono_amd64_get_exception_trampoline ((rethrow), (corlib), (llvm_abs), (resume_unwind), MONO_JIT_ICALL_ ## tramp_name, (aot), (reg), (tramps))) 
```


Without any/much other change.

The point is to get the "names" and the JIT icall registration kinda at a different "distance" from each other in the source.

Basically every call to `register_jit_icall` "needs" a change, and the vast bulk of them, can be done
by making `register_jit_icall` into a macro -- to pass on the enum or struct field.

But these cannot be, w/o this transform.

It is breaking up the referenced PR into smaller pieces.
That only have value once they are all done (or actually, not until even more change).

It is "like" the destringing/restringing of the emu registration but larger.

jaykrell:
The micro optimization is:
 - reusing the name strings instead of an extra dup
 - running the registration "inline" instead of returning the list

The second part is kinda unavoidable though.
There is possibly another way to do this, smaller diff, by putting the enum in the structs in the returned list. I hadn't considered until later. I'd rather not make things "slower" albeit infinitesimally slower.

jaykrell:
And then to be clear, this change should not really change any behavior. It is setup for later optimization.

jaykrell:
Approve? I'll rebase to master and rebuild to keep it current.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

vargaz:
I think it would be much simpler to store the future jit icall id in MonoTrampInfo, i.e. pass the id to
get_throw_trampoline () which would store it in MonoTrampInfo, and you can retrieve it from there when
registering the icall.

jaykrell:
It is a little wasteful but it is a smaller diff, understood. I’ll try that — in a “fork” of the other PR. I’ll leave this open for now but imagine this other option will supercede it.

jaykrell:
Closing per alternate proposal.

PR #14155, opened 2019-04-22
Remove special casing

### LINKED ISSUE #11959: [class-init] Remove special handling of TypedReference and RuntimeArgumentHandle
In `class-init.c` when we are initializing the `MonoClass:is_byreflike` bit, we handle `TypedReference` and `RuntimeArgumentHandle` specially:

https://github.com/mono/mono/blob/ff04de942bfb4e0c3373e151de0c3f95cc429aae/mono/metadata/class-init.c#L668-L672

But since https://github.com/mono/mono/pull/11871 was merged, they are both just ordinary `ref struct` types with an `IsByRefLikeAttribute` custom attribute.  So the special-casing is no longer needed.

pedestrianlove:
Fixes #11959.
Remove special handling of TypedReference and RuntimeArgumentHandle.


mono/metadata/class-init.c
```diff
@@ -669,10 +669,8 @@ mono_class_create_from_typedef (MonoImage *image, guint32 type_token, MonoError
 		/* TypedReference and RuntimeArgumentHandle are byreflike by
 		 * definition. Otherwise, look for IsByRefLikeAttribute.
 		 */
-		if (mono_is_corlib_image (image) &&
-		    ((m_class_get_byval_arg (klass)->type == MONO_TYPE_TYPEDBYREF) ||
-		     (!strcmp (m_class_get_name_space (klass), "System") &&
-		      !strcmp (m_class_get_name (klass), "RuntimeArgumentHandle"))))
+		if (mono_is_corlib_image (image) ||
+		     !strcmp (m_class_get_name_space (klass), "System"))
 			klass->is_byreflike = 1;
```
vargaz: This whole if can be removed.

pedestrianlove: Thanks for your reply!
Sorry I didn't get it while editing.
So this part of the code can be completely deleted? 
Since the OP requested that only the special casing for TypedReference and RuntimeArgumentHandle should be deleted.

Regards,

lambdageek: Everything in this block is just checking for TypedReference and RuntimeArgumentHandle.  A class is identified by a combination of: the assembly that it is defined in (in this case we want it to be in corlib), its namespace and its name.   The code is just arranged like this so that cheap checks happen first.
https://github.com/mono/mono/blob/f1d1bd6084771fa039957b7d9193fe4f949b15be/mono/metadata/class-init.c#L672-L676

PS: also please update or delete this comment
https://github.com/mono/mono/blob/f1d1bd6084771fa039957b7d9193fe4f949b15be/mono/metadata/class-init.c#L669-L671

PPS: If you're wondering why TypedReference has a special constant in MonoTypeEnum it's because this part of Mono's type system is trying to follow the ECMA-335 standard (the document that defines the Common Language Interface), and they have a special token for TypedReference.  And so does Mono.

pedestrianlove: Hi there:
Sorry I was busy past few weeks. Can you please review the modified code to see if anything has gone wrong? (I deleted the if block and the comments also)
Great thanks!

lambdageek: Now a bit too much stuff is gone 😄 

Here's what the code used to say:   "if the class we're initializing is a valuetype, we will set the `byreflike` flag if one of two things is true:  (1) it's a TypedReference or RuntimeArgumentHandle; or (2) it has the IsByRefLike attribute."

Here's what the code says now: "if the class we're initializing is a valuetype, we will set the `byreflike` flag".

What we _want_ it to say: "if the class we're initializing is a valuetype, we will set the `byreflike` flag if it has the IsByRefLike attribute".

pedestrianlove: Sry I seems like an idiot here...(almost all things done in this PR is completed by your directions lol)
I tried to use the IsByRefLikeAttribute function to check for the state of `byreflike` flag.
Is it fine now?

Great thanks for your detailed reply.

lambdageek:
@monojenkins build

lambdageek:
@monojenkins build failed

pedestrianlove:
Is there anything I can do to get this PR merged? @lambdageek

PR #14157, opened 2019-04-22
[netcore] Restore Roslyn compilers from NuGet to allow up-to-date version to be used

filipnavara:


filipnavara:
Hmm, the CI runs the Makefiles in a way that don't hit the `update-roslyn` target. That would have to be fixed.

filipnavara:
It has been pointed out to me by @kg that there may actually be new enough Roslyn binaries in external/roslyn-binaries submodule. We may try to reuse that instead of the NuGets.

marek-safar:
@filipnavara could you please resolve the conflicts. I am fine with this approach It should make it easier for everyone to update (including CI) and we can move to Arcade later

filipnavara:
@marek-safar Sure.

marek-safar:
Thanks @filipnavara!

PR #14158, opened 2019-04-22
[sdks] Add mono-api-html.exe and mono-api-info.exe

alexischr:


sdks/builds/bcl.mk
```diff
@@ -43,11 +43,15 @@ build-$(1)-bcl: build-bcl
...
 	$$(if $(3),$$(MAKE) -C bcl -C runtime test xunit-test test_profiles="$(3)")
+	$$(MAKE) -C bcl -C mcs/tools/mono-api-html
+	$$(MAKE) -C bcl -C mcs/tools/corcompare
 
 .PHONY: package-$(1)-bcl
 package-$(1)-bcl:
 	$$(foreach profile,$(2), \
 		cp -R $$(TOP)/mcs/class/lib/$$(profile)/* $$(TOP)/sdks/out/$(1)-bcl/$$(profile);)
+	cp $$(TOP)/mcs/tools/mono-api-html/mono-api-html.exe $$(TOP)/sdks/out/$(1)/
+	cp $$(TOP)/mcs/tools/corcompare/mono-api-info.exe $$(TOP)/sdks/out/$(1)/
```
akoeplinger: We should add these tools to the `monodroid_tools` profile instead of building/copying them here.

alexischr: Oh that's even better, I forgot about that profile - we briefly discussed a new tools dir.

akoeplinger: Hmm they are already there: https://github.com/mono/mono/blob/aed21f3e4bcdc2f50640866affe808cc5c1423ed/mcs/tools/Makefile#L56

alexischr: I see, they are on master, not 2019-02. I'll double-check in the archive but maybe we're good here then :)

alexischr: Yep, as of master they are in `android-bcl/monodroid/monodroid_tools`, as is `pdb2mdb.exe`

PR #14167, opened 2019-04-22
[netcore] Convert netcore/Makefile.am to a normal makefile, get the configure variables from config.make.

vargaz:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6539,9 +6539,10 @@ fi
     echo "RUNTIME = $mono_build_root/runtime/mono-wrapper" >> $mcs_topdir/build/config.make
     echo "ILDISASM = $mono_build_root/runtime/monodis-wrapper" >> $mcs_topdir/build/config.make
     echo "JAY_CFLAGS = $JAY_CFLAGS" >> $mcs_topdir/build/config.make
-	echo "RID = $RID" >> $mcs_topdir/build/config.make
-	echo "COREARCH = $COREARCH" >> $mcs_topdir/build/config.make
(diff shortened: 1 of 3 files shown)


vargaz:
@monojenkins build failed

directhex:
```
cd /home/vsts/work/1/s/mcs && make --no-print-directory -s NO_DIR_CHECK=1 PROFILES='             netcore' CC='gcc' all-profiles
make[7]: *** No rule to make target 'all'.  Stop.
```

from the build on VSTS

vargaz:
Did that work before ? Its a non-required lane.

directhex:
It worked before, yes. It's the job ./scripts/ci/pipeline-netcore-runtime.yml which will be used to build artifacts for consumption by core-setup

directhex:
```
nuget pack runtime.nuspec -properties VERSION=\;RID=linux-x64\;PLATFORM_AOT_SUFFIX=.so\;COREARCH=x64
Attempting to build package from 'runtime.nuspec'.
Value cannot be null or an empty string.
Parameter name: value
```

Looks like the version parameter isn't getting filled in? That comes from the AC_INIT line (i.e. it populates `@VERSION@` in .am files, or `$VERSION` within configure.ac - I'd like to push in something like https://github.com/mono/mono/pull/14011 though, to get 4-part version numbers instead of 3.

directhex:
Oh, you force-pushed while I was writing my message :sweat_smile:

directhex:
OK, that passed. Do you want an immediate merge? Other lanes aren't really relevant or affected

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14169, opened 2019-04-23
Add roslyn bump instructions next to roslyn-binaries

kg:
@monojenkins skip

### DIFF
--- a/external/how-to-bump-roslyn-binaries.md
+++ b/external/how-to-bump-roslyn-binaries.md
@@ -1,0 +1,51 @@
+How to bump roslyn
+
+1. Pick a revision of roslyn to use. Ideally, use a revision that Visual Studio is using, but any revision should do.
+2. Make a new folder in roslyn-binaries for this revision. Copy over the contents of the *previous folder* into this folder.
+3. Make a temporary commit.
+4. Find that revision in the 'roslyn compilers' nuget feed: https://dotnet.myget.org/feed/roslyn/package/vsix/7922692f-f018-45e7-8f3f-d3b7c0262841
+5. Download the vsix file and unpack it (it's a ZIP file).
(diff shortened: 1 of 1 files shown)


external/how-to-bump-roslyn-binaries.md
```diff
@@ -0,0 +1,51 @@
...
+12. Once you commit the submodule bump to your local repository, you can do test builds. **If you do test builds without committing the bump, the build process will erase the bump.**
+13. Now we need to change the roslyn paths used by the build process - we're still using the older roslyn folder. We do this by changing the CSC_LOCATION and VBCS_LOCATION variables in configure.ac. They should be something like ```Microsoft.Net.Compilers/3.1.0/csc.exe```, change the version number in those paths, save, and make a local commit. Then do the full autogen, make, make install routine and do some basic tests. This should work.
+14. Next, make sure the bump didn't break csi. If you managed the .rsp files correctly in the last step, it should work. This has test coverage, so just pay attention to the automated tests for it.
+15. Ensure the 'Linux x64 - prefix sanity checks' lane passed on CI, even if it's currently disabled or not required. This is required for a bump. If it isn't visible on the PR statuses, manually trigger the lane using Jenkins.
```
jaykrell: "using Jenkins" do you mean @monojenkins something? Be more specific?

kg: By going to jenkins and triggering the lane. It's not monojenkins. Do you think I need to add a URL or something? I'm not sure we can link to that

jaykrell: Agreed about the uncertainty then. I don't know.
33 steps, ugh.

kg: I'm sure it will turn out that I missed at least one :)

Issue #14103, opened 2019-04-17
Crash building F# 4.5 on 32-bit ARM with Mono 6


directhex:
```
Build started 04/17/2019 13:36:58.
Project "/tmp/fsharp-4.5/src/fsharp/FSharp.Build-proto/FSharp.Build-proto.fsproj" on node 1 (default targets).
CallFsSrGen:
  Creating directory "obj/Proto/net40/".
  mono /tmp/fsharp-4.5/packages/FSharp.Compiler.Tools.4.1.27/tools/fsi.exe --exec "/tmp/fsharp-4.5/src/scripts/fssrgen.fsx" "/tmp/fsharp-4.5/src/fsharp/FSharp.Build/FSBuild.txt" "obj/Proto/net40/FSBuild.fs"  "obj/Proto/net40/FSBuild.resx"

  =================================================================
        Native Crash Reporting
  =================================================================
  /proc/self/maps:
  00412000-00706000 r-xp 00000000 08:02 551368                             /usr/bin/mono-sgen
  00715000-0071a000 r--p 002f3000 08:02 551368                             /usr/bin/mono-sgen
  0071a000-0071d000 rw-p 002f8000 08:02 551368                             /usr/bin/mono-sgen
  0071d000-0072a000 rw-p 00000000 00:00 0
  01a56000-02bfd000 rw-p 00000000 00:00 0                                  [heap]
  eb900000-eba00000 rw-p 00000000 00:00 0
  eba70000-ebd00000 rw-p 00000000 00:00 0
  ebd3c000-ec100000 rw-p 00000000 00:00 0
  ec117000-ec600000 rw-p 00000000 00:00 0
  ec67c000-ecf00000 rw-p 00000000 00:00 0
  ecff2000-ede00000 rw-p 00000000 00:00 0
  ede1c000-ee200000 rw-p 00000000 00:00 0
  ee28a000-ef100000 rw-p 00000000 00:00 0
  ef100000-ef121000 rw-p 00000000 00:00 0
  ef121000-ef200000 ---p 00000000 00:00 0
  ef2ff000-ef300000 ---p 00000000 00:00 0
  ef300000-ef400000 rwxp 00000000 00:00 0
  ef400000-ef443000 rw-p 00000000 00:00 0
  ef443000-ef500000 ---p 00000000 00:00 0
  ef598000-ef5ff000 rwxp 00000000 00:00 0
  ef5ff000-ef600000 ---p 00000000 00:00 0
[...]

PR #12701, opened 2019-01-31
[Tailcall] [arm] [jit] Fix moving tailcall parameters.

thaystg:
When moving parameters for tailcall, first copy the parameters using ARMREG_IP and after that get destination in ARMREG_IP, then it's not necessary the PUSH and POP to SP nor the saved_ip_offset.

Fixes #11489

PR #14178, opened 2019-04-23
Fix the tailcall part of https://github.com/mono/mono/issues/14103

jaykrell:
Fix the tailcall part of https://github.com/mono/mono/issues/14103
Does building F# totally succeed on other targets? I'll see.

There are 4 pieces to this.

1. Revert 257efc6374cbffb95d83dc9257b432d6ae3d80f6.
The reordering this change is scary and seemingly pointless.
The register could refer to the frame and the ordering
makes it point to random instead.

The call target must be fetched before moving stuff around in frame.

See https://github.com/mono/mono/pull/12701.

2. The original fix to the bug that was trying to fix but for some reason was rejected.
See https://github.com/mono/mono/pull/12070.

3. Optimization to stop moving padding around.
See https://github.com/mono/mono/pull/12070.

4. Optimization to use frame_reg directly when frame is not huge.
See https://github.com/mono/mono/pull/12079.

With this change, building F# gets much further.
And ultimiately gets a null deref.

If you further disable tailcall entirely for ARM,
it gets just as far.

Perhaps we should take just parts 1 and 2 for now.

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -5027,18 +5027,42 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 
 			code = realloc_code (cfg, max_len);
 
+			// For reg and membase, get destination in IP.
+
+			if (tailcall_reg) {
+				g_assert (ins->sreg1 > -1);
+				if (ins->sreg1 != ARMREG_IP)
+					ARM_MOV_REG_REG (code, ARMREG_IP, ins->sreg1);
+			} else if (tailcall_membase) {
+				g_assert (ins->sreg1 > -1);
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
Which “end” of frame are parameters with respect to the padding for alignment? I.e. in the change to avoid copying the padding, should it loop 0 to unaligned_size or (aligned_size - unaligned_size) to aligned_size? I coded the first before thinking more about it.

jaykrell:
Github shows the commits in the wrong order. The revert is first.

jaykrell:
Smaller fix instead.
https://github.com/mono/mono/pull/14187

PR #14180, opened 2019-04-23
[netcore] Don't wrap CustomAttributeNamedArgument into CustomAttributeNamedArgument

EgorBo:
Basically to remove this check https://github.com/mono/corert/blob/master/src/System.Private.CoreLib/src/System/Reflection/CustomAttributeNamedArgument.cs#L42-L48

(`CustomAttributeNamedArgument` is moving to shared in https://github.com/dotnet/coreclr/pull/23651 without this #if MONO).

/cc: @filipnavara

Also, rename `dest` to `dst` in BlockCopy to fix a failing test (see https://github.com/dotnet/coreclr/pull/23636)


netcore/Makefile.am
```diff
@@ -3,10 +3,10 @@
 #
 
 # Extracted MicrosoftPrivateCoreFxNETCoreAppVersion from https://github.com/dotnet/coreclr/blob/master/eng/Versions.props#L11
-NETCORETESTS_VERSION := 4.6.0-preview5.19209.12
+NETCORETESTS_VERSION := 4.6.0-preview5.19217.2
 
 # Extracted MicrosoftNETCoreAppVersion from https://github.com/dotnet/coreclr/blob/master/eng/Versions.props#L13
-NETCOREAPP_VERSION := 3.0.0-preview5-27609-17
+NETCOREAPP_VERSION := 3.0.0-preview5-27620-01
```
EgorBo: update corefx in order to fix failing System.Range test (`GetOffsetAndLength` doesn't return  `OffsetAndLength` anymore)

mono/metadata/custom-attrs.c
```diff
@@ -1381,7 +1381,11 @@ reflection_resolve_custom_attribute_data (MonoReflectionMethod *ref_method, Mono
 
 		typedarg = create_cattr_typed_arg (arginfo [i].type, obj, error);
 		goto_if_nok (error, leave);
+#if ENABLE_NETCORE
+		namedarg = create_cattr_named_arg (minfo, obj, error);
```
vargaz: This should skip the creation of typedarg as well.

PR #14186, opened 2019-04-23
[acceptance-tests] Run roslyn tests using in-tree mono instead of system mono

akoeplinger:
We were inadvertently running the roslyn tests against the system mono in CI.
Instead we now install mono to a custom prefix, prepend that to PATH and run
the roslyn tests against the correct mono we just built.


akoeplinger:
@monojenkins backport 2019-04
@monojenkins backport 2019-02

akoeplinger:
@monojenkins backport 2019-04
@monojenkins backport 2019-02

PR #14187, opened 2019-04-23
Fix arm32 tailcall_reg/tailcall_membase.

jaykrell:
Revert 257efc6374cbffb95d83dc9257b432d6ae3d80f6
and provide something much closer to what it was based on.

There have been many variations of this.

My original slightly bigger fix that adds a small optimization:
https://github.com/mono/mono/pull/12070

My original bigger fix with even more optimization:
https://github.com/mono/mono/pull/12079 

The broken fix that was merged:
https://github.com/mono/mono/pull/12701

A current bigger fix:
https://github.com/mono/mono/pull/14178 

This PR is a revert of #12701 and then closely resembles #12070, but the small optimization is removed, and the semantically same thing is stated with an attempt at increased clarity and more const and slightly less change.

This should help or fix https://github.com/mono/mono/issues/14103.

The basic problem with #12701 is it moved computation of the target, in IP (intra procedural scratch), relative to the movement of the parameters. This is quite risky because the computation of the target could depend on the parameters being in their "normal" location (esp. for tailcall_membase, the register could point to the frame; for tailcall_reg it seems safe). The code was previously structured conservatively to avoid breaking such a dependency.

The basic problem prior to #12701 was assumption that the frame register is the stack pointer, which is usually but not always true. So the change is to compute 0 or 4, two values, one for SP one for frame register, when moving the parameters.

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -5027,18 +5027,42 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 
 			code = realloc_code (cfg, max_len);
 
+			// For reg and membase, get destination in IP.
+
+			if (tailcall_reg) {
+				g_assert (ins->sreg1 > -1);
+				if (ins->sreg1 != ARMREG_IP)
+					ARM_MOV_REG_REG (code, ARMREG_IP, ins->sreg1);
+			} else if (tailcall_membase) {
+				g_assert (ins->sreg1 > -1);
+				if (!arm_is_imm12 (ins->inst_offset)) {
+					g_assert (ins->sreg1 != ARMREG_IP); // temp in emit_big_add
+					code = emit_big_add (code, ARMREG_IP, ins->sreg1, ins->inst_offset);
+					ARM_LDR_IMM (code, ARMREG_IP, ARMREG_IP, 0);
+				} else {
+					ARM_LDR_IMM (code, ARMREG_IP, ins->sreg1, ins->inst_offset);
+				}
+			}
+
 			/*
 			 * The stack looks like the following:
 			 * <caller argument area>
 			 * <saved regs etc>
 			 * <rest of frame>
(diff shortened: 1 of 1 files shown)


mono/mini/mini-arm.c
```diff
@@ -5056,12 +5056,27 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
+				if (save_ip)
 					ARM_PUSH (code, 1 << ARMREG_IP);
-					saved_ip_offset = 4;
-				}
+
+				// When moving stacked parameters from the bottom
+				// of the frame (sp) to the top of the frame (ip),
+				// account, 0 or 4, for the conditional save of IP.
+				const int offset_sp = save_ip ? 4 : 0;
+				const int offset_ip = (save_ip && (cfg->frame_reg == ARMREG_SP)) ? 4 : 0;
```
jaykrell: This is the important part, `frame_reg == SP`.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This essentially does fix the F# build.
There is still a stack overflow, but you can `ulimit -s` that way.
It manifests unclearly, as error 139 building unit tests.
If you peel the onion and cd and run a command, you get core dump.
If you peel and run under gdb, sigsegv.
Notice it is on a push.
And that sp offset by 4 is invalid.
Stack overflow.
Stack is 8192 -- k or bytes? Not sure.
Raise to 65536, and it works.

directhex:
Can we backport this (and/or dependent changes) to 2019-04 and 2019-02?

jaykrell:
Please first double check me that it fixes F#. Too much room for error
 to trust single observer.

directhex:
Yeah, I'll do that now

directhex:

directhex:

PR #14188, opened 2019-04-23
[netcore] Mangle .gitmodules on internal project to internal mirrors

directhex:
Let's see what VSTS thinks of my conditional


directhex:
I guess the yaml was at least syntactically correct? The job didn't fail. Mind you I'm not sure if VSTS uses YAML from PRs in PR builds.

PR #14191, opened 2019-04-24
[wasm][bindings] Refactor WebAssembly.Core object instantiation.

kjpou1:
Implement suggestion by Katelyn Gadd to use Symbol instead of a Map or WeakMap.

- Attach a special named Symbol for `wasm type` to the prototype property for the Core objects that we have bound.
- The Managed Type is stored in the Symbol property that will be passed to the managed side.
- Move the supported Core object identification into javascript.
- Runtime: the core object type is now passed into the Runtime from javascript binding support instead of being housed in managed code.
- Remove the use of tostring to identify the object type.

### DIFF
--- a/sdks/wasm/binding_support.js
+++ b/sdks/wasm/binding_support.js
@@ -7,7 +7,7 @@ var BindingSupportLib = {
 		mono_wasm_ref_counter: 0,
 		mono_wasm_free_list: [],
 		mono_wasm_marshal_enum_as_int: false,
-		mono_wasm_core_map: {},	
+		mono_wasm_core_map: null,	
 		mono_bindings_init: function (binding_asm) {
 			this.BINDING_ASM = binding_asm;
 		},
@@ -96,7 +96,7 @@ var BindingSupportLib = {
 			this.object_to_string = get_method ("ObjectToString");
 
 			this.object_to_enum = get_method ("ObjectToEnum");
-			BINDING.mono_wasm_core_map = new Map();
+			BINDING.mono_wasm_core_map = new WeakMap();
 			this.init = true;
 		},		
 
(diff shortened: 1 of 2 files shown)


kjpou1:
@monojenkins build failed

sdks/wasm/binding_support.js
```diff
@@ -726,6 +728,38 @@ var BindingSupportLib = {
...
+					case switchName === "Int32Array":
+					case switchName === "Uint32Array":
+					case switchName === "Float32Array":
+					case switchName === "Float64Array":
+					case switchName === "Function":
+					case switchName === "SharedArrayBuffer":
+						coreType = this.wasm_get_core_type(obj);
+						break;
+				}
+		  }
```
lewing: nit: indentation

kg: is there a specific reason why the switch block is written this way? it is by far not how things are typically written in JS

sdks/wasm/binding_support.js
```diff
@@ -94,7 +96,7 @@ var BindingSupportLib = {
 			this.object_to_string = get_method ("ObjectToString");
 
 			this.object_to_enum = get_method ("ObjectToEnum");
-
+			BINDING.mono_wasm_core_map = new Map();
```
kg: I think this should probably be a WeakMap

sdks/wasm/binding_support.js
```diff
@@ -6,7 +6,8 @@ var BindingSupportLib = {
 		mono_wasm_object_registry: [],
 		mono_wasm_ref_counter: 0,
 		mono_wasm_free_list: [],
-		mono_wasm_marshal_enum_as_int: false,	
+		mono_wasm_marshal_enum_as_int: false,
+		mono_wasm_core_map: {},
```
kg: If this is a Map (or WeakMap) normally, initializing it to {} is wrong and will break. Might as well init it to null.

sdks/wasm/binding_support.js
kjpou1: Addressed in recent commit

sdks/wasm/binding_support.js
kjpou1: Addressed in recent commit

sdks/wasm/binding_support.js
kjpou1: Arghh, absolutely correct I did not go back and change it correctly.  It was a leftover from a switch for instanceof which did not work reliably.
Addressed in recent commit.

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

lewing:
@monojenkins build failed

PR #14194, opened 2019-04-24
[RFC] Drop arm32 tailcall support.

jaykrell:
I don't like it, but it was suggested.
Perhaps this is more than was suggested, granted.

Probably also need to disable tests.


jaykrell:
I prefer https://github.com/mono/mono/pull/14187 instead.

PR #14199, opened 2019-04-24
[wasm][sdk] Add command line quoting and escaping to WasmLinkAssembli…

kjpou1:
…es ToolTask.

- This resolves a few issues when executing `monolinker.exe` from windows build.
  - `monolinker.exe` executed on windows requires the windows path to be quoted or exception was thrown.
  - `monolinker.exe` executed on windows throws `Legacy path not supported` exception if the the path separator `\` is not escaped `\\`.  Example: `C:\path\to\assembly` needs to be passed as `C:\\path\\to\\assembly` 

The two new modules `Glob.cs` and `ProcessArguments.cs` were added as a general approach of escaping to replace the simplified escaping of the path separators using `.Replace ("\\", "\\\\")}\"")`.  The routines will also add quoting of the parameters instead of relying on the quoting within the module itself.

### DIFF
--- a/sdks/wasm/sdk/Mono.WebAssembly.Build/Glob.cs
+++ b/sdks/wasm/sdk/Mono.WebAssembly.Build/Glob.cs
@@ -1,0 +1,300 @@
+﻿//
+// Author:
+//   Aaron Bockover <abock@xamarin.com>
+//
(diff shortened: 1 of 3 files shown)


kjpou1:
@radical Your comments should be address in the recent commits.

kjpou1:
@monojenkins build failed

PR #14201, opened 2019-04-24
[interp] fix op_explicit for cast from nfloat to nint and vice versa

lewurm:
Fixes https://github.com/xamarin/xamarin-macios/issues/5809

/cc @spouliot

### DIFF
--- a/mono/mini/interp/transform.c
+++ b/mono/mini/interp/transform.c
@@ -998,6 +998,7 @@ interp_handle_intrinsics (TransformData *td, MonoMethod *target_method, MonoMeth
 		} else if (!strcmp ("op_Implicit", tm ) || !strcmp ("op_Explicit", tm)) {
 			MonoType *src = csignature->params [0];
 			MonoType *dst = csignature->ret;
+			MonoClass *src_klass = mono_class_from_mono_type_internal (src);
 			int src_size = mini_magic_type_size (NULL, src);
 			int dst_size = mini_magic_type_size (NULL, dst);
 
@@ -1008,6 +1009,8 @@ interp_handle_intrinsics (TransformData *td, MonoMethod *target_method, MonoMeth
 				if (!mini_magic_is_int_type (src) || !mini_magic_is_int_type (dst)) {
 					if (mini_magic_is_int_type (src))
 						store_value_as_local = TRUE;
+					else if (mono_class_is_magic_float (src_klass))
+						store_value_as_local = TRUE;
 					else
 						return FALSE;
 				}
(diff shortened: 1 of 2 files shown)


lewurm:
@monojenkins build failed

mono/mini/interp/transform.c
```diff
@@ -1008,6 +1009,8 @@ interp_handle_intrinsics (TransformData *td, MonoMethod *target_method, MonoMeth
 				if (!mini_magic_is_int_type (src) || !mini_magic_is_int_type (dst)) {
 					if (mini_magic_is_int_type (src))
 						store_value_as_local = TRUE;
+					else if (mono_class_is_magic_float (src_klass))
```
BrzVlad: Why not use mini_magic_is_float_type  ?

mono/mini/interp/transform.c
```diff
@@ -1016,6 +1019,8 @@ interp_handle_intrinsics (TransformData *td, MonoMethod *target_method, MonoMeth
 				if (!mini_magic_is_float_type (src) || !mini_magic_is_float_type (dst)) {
 					if (mini_magic_is_float_type (src))
 						store_value_as_local = TRUE;
+					else if (mono_class_is_magic_int (src_klass))
```
BrzVlad: Why not use mini_magic_is_int_type  ?

BrzVlad:
I'm wondering why we do this store_value_as_local hack. I think we should refactor the code at some point, so that at execution time, the method knows the real primitive type for the native type, and this managed implementation passes the native type by value. This would mean that we would not have any reference of native types in interp.c. Does this make sense ?

mono/mini/interp/transform.c
lewurm: it accepts `r4`, `r8` and `nfloat`. here I really only wanna do that if it's `nfloat`.

lewurm:
The `store_value_as_local` hack is only for implicit/explicit conversions that are implemented in managed code, for everything else we pass by value. We could intrinsify all the conversions, but some are hard, e.g. `nint` -> `decimal`. About what references to natives types are you worried about in `interp.c`?

That said, I welcome any improvements in this area 🙂 By now, we capture all the scenarios used in Xamarin.iOS in `builtin-types.cs`.

BrzVlad:
My understanding is that we need to create that local and fetch the address because we handle native types as primitive types inside a method (aka we store the value in a stackval), but these methods that we call think that we are passing a random struct (therefore pass it by reference on the vtstack). If that is the case, I was suggesting that it might make everything easier to teach the method that it's going to receive actually a int/long instead of a nint etc.

lewurm:
@monojenkins backport 2019-04
@monojenkins backport 2019-02
@monojenkins backport 2018-10

PR #14203, opened 2019-04-24
[wasm][sdk] Fix warning NU1604, does not contain an inclusive lower b…

kjpou1:
…ound

- Set explicit version in `Sdk.props`


kg:
Are we sure setting an explicit version isn't what we want, regardless of warning? It's not clear to me

lewing:
@monojenkins build failed

kjpou1:
@kg comments addressed in recent commit.

kjpou1:
@monojenkins build failed

PR #14208, opened 2019-04-24
Getting rid of the `DISABLE_REMOTING` conditional.

baulig:
We do not need the `DISABLE_REMOTING` conditional in the BCL because we can use `FEATURE_REMOTING` instead.


baulig:
@monojenkins commit csproj

PR #14229, opened 2019-04-25
[runtime] Add a python version of the offsets tool.

vargaz:
Use it on osx, since the c# version is 32 bits which is going away, and the 64 bit version doesn't work.

### DIFF
--- a/sdks/builds/wasm.mk
+++ b/sdks/builds/wasm.mk
@@ -93,6 +93,11 @@ archive-wasm: package-wasm-runtime
 
 wasm_ARCHIVE += wasm-runtime-$(CONFIGURATION)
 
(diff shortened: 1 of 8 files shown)


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

sdks/builds/wasm.mk
```diff
@@ -93,6 +93,11 @@ archive-wasm: package-wasm-runtime
 
 wasm_ARCHIVE += wasm-runtime-$(CONFIGURATION)
 
+ifeq ($(UNAME),Darwin)
```
marek-safar: I think we should use it everywhere (even if it does not work yet)

PR #14230, opened 2019-04-25
Disable "unreachable code" warnings on wasm.

baulig:


baulig:
@monojenkins build failed

PR #14237, opened 2019-04-25
[Wasm] Add missing pinvoke-tables-default.h from build artifacts

jeromelaban:
This change restores the ability to build wasm AOT apps outside the mono source tree.


vargaz:
@monojenkins squash

vargaz:
@monojenkins build failed

PR #14238, opened 2019-04-25
[netcore] Avoid needing any submodules to build `--with-core=only`

directhex:
This change:

* avoids updating submodules during autogen, if `--with-core=only`
* avoids updating submodules during make, if `--with-core=only`
* Bundles in 3 small files from corert with System.Private.CoreLib, to avoid the need for the corert submodule

This enables us to build on the internal project without needing to mirror or mangle submodules at all, which is proving difficult (and it should also speed builds up)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,11 +19,17 @@ build_with_msvc =
 build_without_msvc = msvc
 endif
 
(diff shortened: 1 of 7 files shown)


vargaz:
@monojenkins build failed

PR #14242, opened 2019-04-26
Basic CI test coverage verifying that we can compile msbuild

kg:
This adds a basic CI pass that attempts to compile the latest revision of xplat-master and verifies that it produced an MSBuild.dll.

Naturally there's a mysterious UNSTABLE in there coming from no obvious location that isn't actually stopping the build from working.


kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

scripts/ci/run-test-msbuild.sh
```diff
@@ -0,0 +1,21 @@
...
+cd /tmp/xplat-master
+${TESTCMD} --label=checkout-xplat-master --timeout=5m git checkout xplat-master
+mkdir /tmp/xplat-master/tmp-links
+ln -T -s ${OLD_CWD}/runtime/mono-wrapper /tmp/xplat-master/tmp-links/mono
+
+OLD_PATH=${PATH}
+export PATH=/tmp/xplat-master/tmp-links/:${OLD_CWD}:${PATH}
+
+# This fails for some reason due to weird bugs in the msbuild infra, and I can't suppress it using testcmd?
+${TESTCMD} --label=compile-msbuild --timeout=15m ./eng/cibuild_bootstrapped_msbuild.sh --host_type mono --configuration Release -skip_tests /p:DisableNerdbankVersioning=true "/p:Projects=/tmp/xplat-master/src/MSBuild/MSBuild.csproj" /p:AssemblyVersion=15.1.0.0
```
rainersigwald:
```suggestion
${TESTCMD} --label=compile-msbuild --timeout=15m ./eng/cibuild_bootstrapped_msbuild.sh --host_type mono --configuration Release -skip_tests /p:DisableNerdbankVersioning=true "/p:Projects=/tmp/xplat-master/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj" /p:AssemblyVersion=15.1.0.0
```

rainersigwald: actually `-skip_tests` doesn't seem to work either--try this:

```suggestion
${TESTCMD} --label=compile-msbuild --timeout=15m ./eng/cibuild_bootstrapped_msbuild.sh --host_type mono --configuration Release /p:DisableNerdbankVersioning=true "/p:Projects=/tmp/xplat-master/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj" /p:AssemblyVersion=15.1.0.0
```

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build failed

scripts/ci/run-test-msbuild.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
+
+OLD_CWD=`pwd`
+TESTCMD=`realpath ${TESTCMD}`
+
+${TESTCMD} --label=remove-mono-temp-dir --timeout=1m rm -rf /tmp/xplat-master
+${TESTCMD} --label=autogen --timeout=15m ./autogen.sh --prefix=/tmp/mono-from-source
+${TESTCMD} --label=make --timeout=30m make PROFILE=net_4_x -j 2
```
akoeplinger: you don't need to recompile mono here, you can just pass the desired prefix to autogen.sh in run-jenkins.sh.

scripts/ci/run-test-msbuild.sh
```diff
@@ -0,0 +1,22 @@
...
+${TESTCMD} --label=cert-sync --timeout=15m cert-sync /etc/ssl/certs/ca-certificates.crt
+
+${TESTCMD} --label=remove-msbuild-temp-dir --timeout=1m rm -rf /tmp/xplat-master
+${TESTCMD} --label=clone-msbuild --timeout=5m git clone https://github.com/mono/msbuild.git /tmp/xplat-master
+cd /tmp/xplat-master
+${TESTCMD} --label=checkout-xplat-master --timeout=5m git checkout xplat-master
+
+${TESTCMD} --label=compile-msbuild --timeout=15m ./eng/cibuild_bootstrapped_msbuild.sh --host_type mono --configuration Release /p:DisableNerdbankVersioning=true "/p:Projects=/tmp/xplat-master/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj" /p:AssemblyVersion=15.1.0.0
+
+${TESTCMD} --label=check-for-dll --timeout=1m test -s /tmp/xplat-master/artifacts/bin/MSBuild/*-MONO/net472/MSBuild.dll
```
akoeplinger: I think we should move this whole msbuild check into `run-test-make-install.sh` since that script/job already tests something very similar.

scripts/ci/run-test-msbuild.sh
```diff
@@ -0,0 +1,22 @@
...
+${TESTCMD} --label=autogen --timeout=15m ./autogen.sh --prefix=/tmp/mono-from-source
+${TESTCMD} --label=make --timeout=30m make PROFILE=net_4_x -j 2
+${TESTCMD} --label=make-install --timeout=15m make install
+
+OLD_PATH=${PATH}
+export PATH=/tmp/mono-from-source/bin/:${OLD_PATH}
+${TESTCMD} --label=cert-sync --timeout=15m cert-sync /etc/ssl/certs/ca-certificates.crt
+
+${TESTCMD} --label=remove-msbuild-temp-dir --timeout=1m rm -rf /tmp/xplat-master
+${TESTCMD} --label=clone-msbuild --timeout=5m git clone https://github.com/mono/msbuild.git /tmp/xplat-master
```
akoeplinger: I wonder if we should grep the msbuild commit out of `packaging/MacSDK/msbuild.py` instead of always taking xplat-master?

scripts/ci/run-test-msbuild.sh
kg: That sounds good as long as it doesn't mean msbuild oranges cause that whole lane to get shut off

scripts/ci/run-test-msbuild.sh
kg: That seems like a good idea. We always want to be using the same msbuild on mac and linux, right?

directhex: They ought to always match. When they don't, it's usually because nobody told me to update Linux.

kg: It sounds like the right fix is to verify that the macsdk hash matches xplat-master head or matches linux-packaging?

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

scripts/ci/run-test-msbuild.sh
```diff
@@ -0,0 +1,39 @@
...
+echo Detected msbuild hash ${MSBUILD_REVISION} in MacSDK/msbuild.py
+
+${TESTCMD} --label=clone-msbuild --timeout=5m git clone https://github.com/mono/msbuild.git /tmp/xplat-master
+cd /tmp/xplat-master
+${TESTCMD} --label=checkout-xplat-master --timeout=5m git checkout ${MSBUILD_REVISION}
+
+${TESTCMD} --label=compile-msbuild --timeout=15m ./eng/cibuild_bootstrapped_msbuild.sh --host_type mono --configuration Release /p:DisableNerdbankVersioning=true /p:CreateBootstrap=true "/p:Projects=/tmp/xplat-master/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj" /p:AssemblyVersion=15.1.0.0
+
+MONO=/tmp/mono-from-source/bin/mono
+DLL_PATH=`realpath /tmp/xplat-master/artifacts/2/bin/MSBuild.Bootstrap/*-MONO/net472/MSBuild.dll`
```
radical: It would be much better to install msbuild to `/tmp/mono-from-source` and then use `DLL_PATH=/tmp/mono-from-source/lib/mono/msbuild/Current/bin/MSBuild.dll`.
msbuild is not just the stuff in that directory.

kg: OK, I'll work on trying to do that instead

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build Linux x64 - compile msbuild

kg:
@monojenkins build compile msbuild

scripts/ci/run-test-msbuild.sh
kg: Is there a way to do this and still install to a custom prefix? Needing someone to use run-jenkins to run the test script is undesirable

scripts/ci/run-test-msbuild.sh
kg: The test script does a proper install to the prefix and tests that, so I think that's enough. The old-style test is still in here but that's because it seems like some customers do it. If it breaks here that's a good reason to remove it from the test and update the documentation to clearly tell them not to do it. :) Maybe we should update it anyway?

kg:
@monojenkins build compile msbuild

kg:

scripts/ci/run-test-msbuild.sh
kg: I still don't understand what to do here

scripts/ci/run-test-msbuild.sh
kg: Is moving this test into make install a requirement? I'd rather it be separate but if you want it moved there I can do it

kg:
@monojenkins build failed

kg:
@monojenkins build failed

kg:
@monojenkins build compile msbuild

kg:
@monojenkins build compile msbuild

akoeplinger:
@monojenkins build compile msbuild

PR #14245, opened 2019-04-26
Add a few `RuntimeFeature.IsDynamicCodeSupported` conditionals.

baulig:


mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
```diff
@@ -714,7 +714,8 @@ public override MethodInfo MakeGenericMethod (Type [] methodInstantiation)
 #if FULL_AOT_RUNTIME
 				throw new NotSupportedException ("User types are not supported under full aot");
 #else
-				return new MethodOnTypeBuilderInst (this, methodInstantiation);
+				if (RuntimeFeature.IsDynamicCodeSupported)
```
marek-safar: I think it would be better to throw for `!RuntimeFeature.IsDynamicCodeSupported` and remove `#if FULL_AOT_RUNTIME`

```
throw new PlatfornNotSupportedException ("User types are not supported");
```

baulig:
Resurfaced as a draft version in https://github.com/mono/mono/pull/14284.

PR #14248, opened 2019-04-26
[sdks] Make sure GCC runtime libs are installed for MinGW targets

grendello:
When building with MinGW, the mono runtime DLL (`libmonosgen-2.0.dll`) is built
and linked with both `-shared` (provided by libtool) as well as
`-static-libgcc` (provided by Mono's `configure.ac`) however it doesn't have the
desired effect of linking static versions of all the GCC runtime libraries. The
reason is hidden in the MinGW's gcc spec files which, seeing `-shared` first on
the command line, prefer to link the library with `-lgcc -lgcc_s` which gives us
static `libgcc` but dynamic SEH (structured exception handling) library -
`libgcc_s_seh-1.dll`.
The easiest fix, implemented by this commit, is to simply package the runtime
library with Mono. This is done only for the mingw targets.

PR #14251, opened 2019-04-27
Remove libgcc_s_seh-1.dll dependency when targeting Win32 with gcc.

jaykrell:
Don't use popcount intrinsic.
Alternative but not exactly contradictory to https://github.com/mono/mono/pull/14248.

Note that this is a single inlinable instruction on x86,
if you assume at least SSE4 from circa 2007, which I am avoiding here.


grendello:
Definitely much better than #14248, thanks Jay!

alexischr:

alexischr:

PR #14253, opened 2019-04-27
[wasm] Add support for pinvokes in user assemblies.

vargaz:
The --pinvoke-libs=x argument to packager.exe can be used to generate pinvoke linking tables.
'x' is a list of library names used in DllImport declarations, i.e. --pinvoke-libs=libtest is used
to link with [DllImport ("libtest")].


vargaz:
@monojenkins build failed

PR #14256, opened 2019-04-28
Correct type and bitness for TARGETS request of X11 clipboard

### LINKED ISSUE #14255: Copying from Mono port of Windows Forms doesn't work in some programs on Linux
## Steps to Reproduce

1. Create a following program, compile it and run with Mono:
   ```csharp
   using System.Windows.Forms;
   
   namespace ConsoleApplication1
   {
       internal class Program
       {
           {
               var form = new Form();
               var textBox = new TextBox();
               form.Controls.Add(textBox);
               Application.Run(form);
           }
       }
   }
2. Enter any text into a text box, select it and press Ctrl-C (or other default keyboard shortcut for **Copy** action).

### Current Behavior

1. Try to paste the result:

   - into `gedit`: success
   - into `gnome-terminal`: failure
   - into any Swing-based Java application: failure

2. Try to call `xclip -selection clipboard -o -t TARGETS > targets.txt`: it will output 40 bytes of garbage instead of a list of newline-separated targets supported by Mono.

### Expected Behavior

1. Paste into `gnome-terminal` and into any Swing-based Java application should work by default.
2. `xclip -selection clipboard -o -t TARGETS` should output a list of newline-separated targets, not garbage.

- Linux (I was testing on a 64-bit Ubuntu 18.04)

**Version Used**:

Latest `master`, commit `90a91429ae75240b4d0e791a563a9d553be09d56`, and latest stable, version 5.20.1.19.

## Analysis

The information provided by Mono into X11 clipboard mechanism is wrong in two ways:
1. Mono claims to provide the information of type `TARGETS` instead of type `ATOM`, so `XGetWindowProperty` will fail if asked to get information of type `ATOM` when performing a paste action.
2. The corresponding Mono code seems to think that X11 `Atom` is 32-bit always, but it is not the case: `Atom` is 64-bit on a 64-bit system.

I will send a patch fixing the issues soon.

ForNeVeR:
Fixes #14255.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/XplatUIX11.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/XplatUIX11.cs
@@ -1808,27 +1808,27 @@ namespace System.Windows.Forms {
 
 					// Seems that some apps support asking for supported types
 					if (format_atom == TARGETS) {
-						int[]	atoms;
+						IntPtr[]	atoms;
 						int	atom_count;
 
-						atoms = new int[5];
+						atoms = new IntPtr[5];
 						atom_count = 0;
 
 						if (Clipboard.IsSourceText) {
-							atoms[atom_count++] = (int)Atom.XA_STRING;
-							atoms[atom_count++] = (int)OEMTEXT;
-							atoms[atom_count++] = (int)UTF8_STRING;
-							atoms[atom_count++] = (int)UTF16_STRING;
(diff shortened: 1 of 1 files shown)


marek-safar:
approve

ForNeVeR:
I can see that something's wrong with the test. Am I expected to diagnose the test failures? From a quick look, I can't see how they relates to the bugfix.

EgorBo:
@monojenkins build failed

marek-safar:
Thank you!

PR #14257, opened 2019-04-29
Add minimal COM support into WinAOT BCL profile and Windows Mono runtime.

lateralusX:
Technologies like SharpDX uses a minimal set of COM support from runtime and System.Runtime.InteropServices.Marshal:

AddRef
Release
QueryInterface

All these are low level wrappers around methods in IUknown interface and doesn't need any of the additional runtime COM support enabled to work.

Technologies like CoreRT includes these methods in their full AOT profile, meaning that SharpDX can run on CoreRT on Windows platforms. Mono's WinAOT profile and runtime currently don't, making it impossible  to use SharpDX.

This PR adds the needed methods into WinAOT profile and also make sure corresponding icalls always gets included on Windows build Mono runtime.


marek-safar:
@monojenkins commit csproj

PR #14259, opened 2019-04-29
[Wasm] Add support for P/Invoke through dynamic linking

jeromelaban:
- New wasm build artifact name `release-dynamic` which includes `-s MAIN_MODULE=1` and `-s EXPORT_ALL=1` based on [emscripten's linking docs](https://github.com/emscripten-core/emscripten/wiki/Linking). This is a new configuration to avoid paying the additional cost of exporting all symbols. The runtime goes from 1.8MB to 3.6MB with these options.
- The ability for P/Invoke to work when symbols are exported in an additional wasm file, included via `dynamicLibraries`, e.g. `Module.dynamicLibraries = ['./canvaskit.wasm']`
- dlopen for invalid assemblies is not working properly (https://github.com/emscripten-core/emscripten/issues/8511) and `System.Native` is worked-around explicitly for now.


mono/metadata/Makefile.am
```diff
@@ -176,7 +176,7 @@ libmono_icall_table_la_SOURCES = \
 libmono_icall_table_la_CFLAGS = $(SGEN_DEFINES) @CXX_ADD_CFLAGS@
 libmono_icall_table_la_LDFLAGS = $(libmonoldflags)
 if BITCODE
-libmono_icall_table_la_LIBADD = $(glib_libs) ../utils/libmonoutils.la ../sgen/libmonosgen.la libmonoruntimesgen.la
+libmono_icall_table_la_LIBADD =
 endif
```
vargaz: These were added for the other bitcode config (watchos probably). So it should be:
if BITCODE
if WASM
..
else
...
endif
endif

mono/metadata/Makefile.am
```diff
@@ -196,7 +196,7 @@ libmono_ilgen_la_SOURCES = \
 libmono_ilgen_la_CFLAGS = $(SGEN_DEFINES) @CXX_ADD_CFLAGS@
 libmono_ilgen_la_LDFLAGS = $(libmonoldflags)
 if BITCODE
-libmono_ilgen_la_LIBADD = $(glib_libs) ../utils/libmonoutils.la ../sgen/libmonosgen.la libmonoruntimesgen.la
+libmono_ilgen_la_LIBADD = 
 endif
```
vargaz: Ditto.

mono/utils/mono-dl-wasm.c
```diff
@@ -45,7 +46,7 @@ mono_dl_get_system_dir (void)
 void*
 mono_dl_lookup_symbol (MonoDl *module, const char *name)
 {
-	return NULL;
+	return dlsym(module->handle, name);
 }
```
vargaz: These should use mono calling conventions (space after ( ).

jeromelaban: I'll try to add the .editorconfig file so this does not happen anymore... you've got to be tired of those 
 :)

mono/mini/m2n-gen.cs
```diff
@@ -37,6 +37,10 @@ class Driver {
 		"VIIIIIIII",
 		"VIIIIIIIII",
 		"VIIIIIIIIII",
+		"VIIIIIIIIIII",
+		"VIIIIIIIIIIII",
+		"VIIIIIIIIIIIII",
+		"VIIIIIIIIIIIIII",
 		"I",
```
vargaz: The wasm_m2n_invoke.g.h should be generated as well, using 
csc m2n-gen.cs
mono m2n-gen.exe > wasm_m2n_invoke.g.h

jeromelaban:
So this PR can't be merged as-is because of the dlopen call that I included, which makes [emscripten call abort](https://github.com/emscripten-core/emscripten/blob/bb48a639398136e09b9903d1d2522f76537a2d1e/src/library.js#L2049) when `MAIN_MODULE` is not defined.

To be able to have an optional runtime, and avoid having to do `-s EXPORT_ALL=1` for the `release` runtime, there are two options that I can see:
- either adding a function (built here)[https://github.com/mono/mono/pull/14259/files#diff-2f161238e8367fb46cf04ce102f524beR117] in all three configurations
- or enable `-s MAIN_MODULE=1` in all configurations, with only `-s EXPORT_ALL=1` in `release-dynamic`

I'll make some tests to determine the size of `mono.wasm` with only `-s MAIN_MODULE=1` as it does seems to include a lot of js code.

vargaz:
We already have a pinvoke callback in driver.c called wasm_dl_load ()/wasm_dl_symbol () maybe the dlopen() etc. can be added there.

jeromelaban:
Well, adding MAIN_MODULE=1 makes mono.wasm release go from 1.8MB to 2.6MB, so that's not an option. I'll look for altering the driver, and most probably build it for each configuration to alter dlopen's behavior.

jeromelaban:
@vargaz would something like this be acceptable ?

```c
mono_dl_fallback_register (gboolean allow_dlopen, MonoDlFallbackLoad load_func, MonoDlFallbackSymbol symbol_func, MonoDlFallbackClose close_func, void *user_data)
```

so dlopen is controled from there, or another method that would be in `mono-dl-wasm.c` like `mono_dl_disable_dlopen()` ?

vargaz:
Can't wasm_dl_load () in driver.c call dlopen () if some define is set ?

jeromelaban:
Ah I see. You'd redirect dlopen from there, I'll try that, thanks!

jeromelaban:
I need to make another change to avoid dlopen ordering issues.

jeromelaban:
The Linux WebAssembly tests did not pass because they ran out of time (20 minutes). They pass locally on my machine in a container built using the same parameters the CI uses, in 13 minutes. Could it be that the build machine running it is too slow now that the number of tests has grown and that the timeouts should be larger ?

vargaz:
@monojenkins build failed

lewing:
@monojenkins build failed

jeromelaban:
Good, the wasm test suite passed, the android one is an app crash on android with debugger test suite it seems.

lewing:

lewing:
@monojenkins build failed

lewing:
@jeromelaban unfortunately this was stuck while I landed the m2n changes.  Can you rebase and regenerate?

jeromelaban:
Thanks for the notice. FWIW, your m2n changes may have a positive performance impact for dllimport calls. Trampolines are called very often, particularly in sqlite and skia.

lewing:
Thanks for the update! Yeah, it turns out the m2n stuff had measurable impact in several cases.

lewing:

lewing:

lewing:

PR #14260, opened 2019-04-29
tests and improvements for GetIDsOfNames

madewokherd:


madewokherd:
I guess the test needs to be disabled on platforms/configurations that don't support COM interfaces.

madewokherd:
`cant resolve internal call to "System.Runtime.InteropServices.Marshal::GetIUnknownForObjectInternal(object)" (tested without signature also)

Your mono runtime and class libraries are out of sync.`

I'm confused by this failure, as I didn't change any internal call mappings.

marek-safar:
@lateralusX could you please review this?

/cc @lambdageek

mono/metadata/cominterop.c
```diff
@@ -540,6 +541,30 @@ cominterop_com_visible (MonoClass* klass)
...
 
+static guint32
+cominterop_class_iface_type (MonoClass* klass)
+{
+	ERROR_DECL (error);
+	static MonoClass *ClassInterfaceAttribute = NULL;
+	MonoCustomAttrInfo *cinfo;
+	guint32 result = 2;
+
+	if (!ClassInterfaceAttribute)
```
lateralusX: Does this initialization of static need to be thread safe? maybe better to do it in mono_cominterop_init.

madewokherd: I don't see any documented threading or locking requirement in mono_class_load_from_name. Maybe there is one that I'm not aware of. I do see there's a GENERATE_GET_CLASS_WITH_CACHE macro that seems designed for this purpose, so it's probably better to use that.

mono/metadata/cominterop.c
```diff
@@ -540,6 +541,30 @@ cominterop_com_visible (MonoClass* klass)
 
 }
 
+static guint32
+cominterop_class_iface_type (MonoClass* klass)
+{
+	ERROR_DECL (error);
+	static MonoClass *ClassInterfaceAttribute = NULL;
+	MonoCustomAttrInfo *cinfo;
+	guint32 result = 2;
```
lateralusX: What is this magic number, AutoDual? Maybe use a define or something to describe the meaning of the value.

madewokherd: Yes, it's ClassInterfaceType.AutoDual. I can add a define.

mono/metadata/cominterop.c
```diff
@@ -2589,7 +2622,9 @@ cominterop_ccw_queryinterface_impl (MonoCCWInterface* ccwe, const guint8* riid,
 			for (i = 0; i < ifaces->len; ++i) {
 				MonoClass *ic = NULL;
 				ic = (MonoClass *)g_ptr_array_index (ifaces, i);
-				if (cominterop_class_guid_equal (riid, ic)) {
+				if (handling_idispatch
+					? cominterop_get_com_slot_begin (ic) == 7
```
lateralusX: Maybe use a define for magic number, giving it a little more context.

mono/metadata/cominterop.c
```diff
@@ -2660,47 +2695,65 @@ cominterop_ccw_get_ids_of_names_impl (MonoCCWInterface* ccwe, gpointer riid,
...
 	for (i=0; i < cNames; i++) {
 		methodname = mono_unicode_to_external (rgszNames[i]);
 
-		method = mono_class_get_method_from_name_checked(klass, methodname, -1, 0, error);
-		if (method && is_ok (error)) {
+		if (mono_class_is_interface (iface))
+			pos = 0;
+		else
+			/* TODO: methods on Object should be properly included? */
+			pos = 4;
```
lateralusX: Magic number, either document what it means or use a define for it.

mono/metadata/cominterop.c
```diff
@@ -2660,47 +2695,65 @@ cominterop_ccw_get_ids_of_names_impl (MonoCCWInterface* ccwe, gpointer riid,
...
+
+		for (j = 0; j < methodcount; j++) {
+			method = m_class_get_methods (iface) [j];
```
lateralusX: Better to do m_class_get_methods outside of loop.

mono/metadata/cominterop.c
madewokherd: I don't know what it means, it was found via testing. My theory is that there are 4 Object methods that were assigned id's. Ideally, I'd figure out what those are and assign them that way, but I'm not sure how to find them.

mono/metadata/cominterop.c
```diff
@@ -2660,47 +2695,65 @@ cominterop_ccw_get_ids_of_names_impl (MonoCCWInterface* ccwe, gpointer riid,
...
+		object = mono_gchandle_get_target_internal (ccw->gc_handle);
+		g_assert (object);
+		iface = mono_object_class (object);
+	}
+
+	mono_class_setup_methods (iface);
+	methodcount = mono_class_get_method_count (iface);
+
 	for (i=0; i < cNames; i++) {
 		methodname = mono_unicode_to_external (rgszNames[i]);
```
lateralusX: Is this memory freed?

madewokherd: Can we free it? It seems to be potentially used by COM interop code for the lifetime of the class.

mono/metadata/cominterop.c
```diff
@@ -2660,47 +2695,65 @@ cominterop_ccw_get_ids_of_names_impl (MonoCCWInterface* ccwe, gpointer riid,
...
-				MonoObject *result = mono_custom_attrs_get_attr_checked (cinfo, ComDispIdAttribute, error);
+				MonoDispIdAttribute *result = (MonoDispIdAttribute*)mono_custom_attrs_get_attr_checked (cinfo, ComDispIdAttribute, error);
 				mono_error_assert_ok (error); /*FIXME proper error handling*/;
 
 				if (result)
-					rgDispId[i] = *(gint32*)mono_object_unbox_internal (result);
+					rgDispId[i] = result->dispid;
 				else
-					rgDispId[i] = (gint32)method->token;
+					rgDispId[i] = 0x60020000 + pos;
```
lateralusX: What does this value mean?

madewokherd: I don't know, it was found by testing. This value is commonly used in IDispatch definitions in Windows headers for sequential method id's, in a similar way to what .NET is doing, but I couldn't find a define or documentation for it.

mono/metadata/cominterop.c
lateralusX: Since this is a static maybe it is better to but it together with other statics in this file and init it as part of mono_cominterop_init. That will  make sure it is setup during startup and not during runtime. If that's not possible you might need to do something similar as done int mono_class_try_get_com_object_class. However I see that similar pattern is used on some other places as well so this source is not consistent so maybe it should be adjusted to follow similar pattern as used in other sources getting access to corlib classes.

mono/metadata/cominterop.c
lateralusX: methodname is a local, don't see where it is stored for he lifetime of the class?  Looking at the comments for mono_unicode_to_external it says:

"Callers must free the returned string."

lateralusX:
Possible to give a little context and background of the changes in the PR description?

madewokherd:
Not sure what sort of information is needed as context and background. I'm hoping to use dotnet core winforms in wine-mono, and it implements dispatch interfaces to receive events from mshtml/WebBrowser controls. Technically GetIDsOfNames isn't needed for this, but I thought it'd be a good way to get started without diving straight into the more challenging Invoke function.

lateralusX:
Just describe that use case in the PR description so it will be easier to understand and find in the future.

madewokherd:
From testing .NET, it turns out that System.Object has 4 methods with dispid's: ToString(0), Equals(0x60020001), GetHashCode(0x60020002), and GetType(0x60020003). I don't know why ToString has a dispid of 0 or why it's only those 4 methods, but it does explain why class dispid's start at 0x60020004.

I might have to rethink the approach.

nsivov:
FWIW dispid 0 is special (DISPID_VALUE), making ToString() "default" from scripting point of view.

madewokherd:
I see I need to consider object inheritance, but I'm running into a problem with this. The class's methods array contains the '.ctor' method, which shouldn't be exported to COM interfaces. I think we need some sort of filtering of methods exposed to COM, and we can't safely use the methods array for this because it's used for other things.

madewokherd:
I guess I can do the filtering for IDispatch specifically. We can revisit it if we find that COM interface vtables are wrong.

madewokherd:
Updated PR, requesting new review.

lateralusX:
@monojenkins rebuild failed

mono/tests/idispatch.cs
```diff
@@ -0,0 +1,160 @@
...
+
+	public static bool check_dispid (object obj, string iid, string name, int expected) {
+		IntPtr disp = get_interface (obj, iid);
+
+		int actual = getidofname (disp, name);
+
+		Marshal.Release (disp);
+
+		if (actual != expected)
+		{
```
lateralusX: Not following formatting rules for if statments.

madewokherd:
`Unregistered icall 'System.Runtime.InteropServices.Marshal:GetIUnknownForObjectInternal (object)'`
If this doesn't exist on AArch64 Full AOT then we do need to disable the test. I'm not sure how to correctly disable it for the platforms that lack this icall.

madewokherd:
We should disable the test when DISABLE_COM is 1. Is there a reason we don't check for that directly in the tests makefile?

akoeplinger:
(@lateralusX is currently on vacation so I'm looking at this PR for now)

@madewokherd DISABLE_COM is currently only available in config.h, not as a Makefile symbol. You'd need to add an `AM_CONDITIONAL (DISABLE_COM, test x$mono_feature_disable_com = xyes)` to configure.ac after this: https://github.com/mono/mono/blob/d34381f72e0a04febdb6d6ea084377d0d635e15f/configure.ac#L1789

I agree it would be cleaner than adding the test to each PROFILE_DISABLED_TESTS so would be great if you could change that :)

PR #14270, opened 2019-04-29
Reverting the path change

Anipik:
Helps to fix the final difference between the mono and coreclr shared folder


marek-safar:
@monojenkins build

PR #14284, opened 2019-04-30
Add a few `RuntimeFeature.IsDynamicCodeSupported` conditionals.

baulig:
Move the `RuntimeFeature.IsDynamicCodeSupported` and `IsDynamicCodeCompiled` conditionals over here from the corefx module and add them in a few places where `TypeBuilder` is used.

### DIFF
--- a/mcs/class/corlib/LinkerDescriptor/mscorlib.xml
+++ b/mcs/class/corlib/LinkerDescriptor/mscorlib.xml
@@ -7,7 +7,7 @@
 			<!-- appdomain.c: mono_domain_try_type_resolve -->
 			<method name="DoTypeResolve" />
 			<!-- appdomain.c: mono_domain_try_type_builder_resolve -->
-			<method name="DoTypeBuilderResolve" />
+			<method name="DoTypeBuilderResolve" feature="sre" />
 			<!-- appdomain.c: mono_try_assembly_resolve -->
 			<method name="DoAssemblyResolve" />
 			<!-- appdomain.c: mono_domain_fire_assembly_load -->
--- a/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
+++ b/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
@@ -711,12 +711,11 @@ namespace System.Reflection {
 			}
 
 			if (hasUserType) {
-#if FULL_AOT_RUNTIME
-				throw new NotSupportedException ("User types are not supported under full aot");
-#else
+#if !FULL_AOT_RUNTIME
 				if (RuntimeFeature.IsDynamicCodeSupported)
 					return new MethodOnTypeBuilderInst (this, methodInstantiation);
 #endif
+				throw new NotSupportedException ("User types are not supported under full aot");
 			}
 
 			MethodInfo ret = MakeGenericMethod_impl (methodInstantiation);
--- a/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
+++ b/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
@@ -714,7 +714,8 @@ namespace System.Reflection {
 #if FULL_AOT_RUNTIME
 				throw new NotSupportedException ("User types are not supported under full aot");
 #else
-				return new MethodOnTypeBuilderInst (this, methodInstantiation);
+				if (RuntimeFeature.IsDynamicCodeSupported)
+					return new MethodOnTypeBuilderInst (this, methodInstantiation);
 #endif
 			}
 
--- a/mcs/class/corlib/corefx/RuntimeFeature.cs
+++ b/mcs/class/corlib/corefx/RuntimeFeature.cs
@@ -1,0 +1,9 @@
+namespace System.Runtime.CompilerServices
+{
+	partial class RuntimeFeature
+	{
+		// https://github.com/dotnet/coreclr/blob/397aaccb5104844998c3bcf6e9245cc81127e1e2/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeFeature.CoreCLR.cs#L9-L10
+		public static bool IsDynamicCodeSupported => true;
+		public static bool IsDynamicCodeCompiled => true;
+	}
+}
--- a/mcs/class/referencesource/mscorlib/system/activator.cs
+++ b/mcs/class/referencesource/mscorlib/system/activator.cs
@@ -74,7 +74,7 @@ namespace System {
                 throw new ArgumentNullException("type");
             Contract.EndContractBlock();
 #if !FULL_AOT_RUNTIME
-            if (type is System.Reflection.Emit.TypeBuilder)
+            if (RuntimeFeature.IsDynamicCodeSupported && type is System.Reflection.Emit.TypeBuilder)
                 throw new NotSupportedException(Environment.GetResourceString("NotSupported_CreateInstanceWithTypeBuilder"));
 #endif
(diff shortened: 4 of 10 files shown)


mcs/class/corlib/LinkerDescriptor/mscorlib.xml
```diff
@@ -365,6 +365,10 @@
 		
 		<!-- domain.c: mono_defaults.methodhandle_class -->
 		<type fullname="System.RuntimeMethodHandle" preserve="fields" />
+
+		<type fullname="System.RuntimeType">
+			<method name="MakeTypeBuilderInstantiation" action="throw" feature="sre" />
+		</type>
```
baulig: This needs my new linker PR: https://github.com/mono/linker/pull/555.

mcs/class/referencesource/mscorlib/system/rttype.cs
```diff
@@ -4333,6 +4335,13 @@ public override Type MakeGenericType(Type[] instantiation)
             return ret;
         }
```
baulig: I'm still trying to figure out why using constants did not previous remove this.

baulig:
@monojenkins commit csproj

mcs/class/corlib/corefx/RuntimeFeature.cs
```diff
@@ -0,0 +1,14 @@
+namespace System.Runtime.CompilerServices
+{
+	partial class RuntimeFeature
+	{
+		// https://github.com/dotnet/coreclr/blob/397aaccb5104844998c3bcf6e9245cc81127e1e2/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeFeature.CoreCLR.cs#L9-L10
+#if WASM || (FULL_AOT_RUNTIME && !MONOTOUCH)
+		public const bool IsDynamicCodeSupported = false;
+		public const bool IsDynamicCodeCompiled = false;
```
lewurm: I'm a bit confused by the intention of this PR.
* does `WASM` imply `FULL_AOT_RUNTIME`?
* why is `FULL_AOT_RUNTIME && ...` needed? every usage of those two constants are guarded by `!FULL_AOT_RUNTIME`.

WebAssembly _can_ be shipped with the interpreter, so why would dynamic code execution be disabled?
* note that WASM _can_ have an interpreter available, so dynamic code execution

marek-safar: That's right, Martin we discussed that the fields can only be overwritten by linker for wasm/mobile

baulig: The conditional in it's current form is supposed to mean `WASM` and all the `testing_aot_*` profiles.  I added the `MONOTOUCH` because without it the API Diff broke.

baulig: Note that the constants are strictly required - you cannot use properties and the linker - at least not without additional linker work.

baulig:
@lewurm @marek-safar This is blocking me (especially the corefx bump and csproj) and we need to land this as quickly as possible.

However, we do not need to have a discussion about the correct values of those properties right now, so I changed them back to their old values from corefx.

Could you please review again?

mcs/class/corlib/corefx/RuntimeFeature.cs
baulig: See my comment below; I took out those changes, so we can move along with this one.

baulig:
@monojenkins build

lewurm:
fine for me, I was just curious 🙂  someone of the BCL team has to approve it

baulig:
@monojenkins build failed

mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
```diff
@@ -714,7 +714,8 @@ public override MethodInfo MakeGenericMethod (Type [] methodInstantiation)
 #if FULL_AOT_RUNTIME
 				throw new NotSupportedException ("User types are not supported under full aot");
```
marek-safar: As commented else change this to
```c#
#if !FULL_AOT_RUNTIME
				if (RuntimeFeature.IsDynamicCodeSupported)
					return new MethodOnTypeBuilderInst (this, methodInstantiation);
#endif

				throw new NotSupportedException ("User types are not supported");      
```

mcs/class/referencesource/mscorlib/system/activator.cs
```diff
@@ -74,7 +74,7 @@ private Activator()
                 throw new ArgumentNullException("type");
             Contract.EndContractBlock();
 #if !FULL_AOT_RUNTIME
-            if (type is System.Reflection.Emit.TypeBuilder)
+            if (RuntimeFeature.IsDynamicCodeSupported && type is System.Reflection.Emit.TypeBuilder)
```
marek-safar: We could also fix the linker not to mark for `is` (but that can be done later)

mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs
baulig: Oh sorry, I completely forgot this one.  It's now fixed.

mcs/class/referencesource/mscorlib/system/activator.cs
baulig: That would not work, it's not that trivial ;-)

The only way to make this work _inside_ an active method is to use dead-code elimination - including variable elimination, which again means that you need to adjust all `ldloc` / `stloc` instructions in the method.   That's not as scary as it may sound - and my new module already does this.

To use the existing linker with some minor changes (like for instance my abandoned draft https://github.com/mono/linker/pull/555, with `return false` added) is to move the `is` into a separate helper method and then make the linker rewrite that.

So for instance something like

```
static bool IsTypeBuilder (Type type) => type is TypeBuilder;
```

and then some XML description to tell the linker to turn that into `return false`:

```
<method name="IsTypeBuilder" action="return-false" />
```

baulig: Btw. I spent quite some time researching exactly this `x is y` situation and the tricky part is to distinguish between the trivial cases and those which require flow analysis and dead-code elimination.

This particular example here is simple enough that it could be rewritten - but detecting that is not easy.

In IL, the `isinst Foo` leaves an instance of `Foo` on the stack (but we're going to remove `Foo`, so it's a non-existing type).  The trivial case is something like this (`static bool IsWeakInstanceOf<T> (object obj) => obj is T`):

```
        IL_0000:  ldarg.0 
        IL_0001:  isinst !!0
        IL_0006:  ldnull 
        IL_0007:  cgt.un 
        IL_0009:  ret 
```

But if the `isinst` is followed by a `stloc`, `call` or `ret`, then we might be in trouble.

During my research, I've seen some cases in complex corlib methods where all my detection logic failed, so I only used it in a few cases that we explicitly whitelisted.

marek-safar: It's actually quite trivial, we just don't mark types as instantiated when used from is instruction. That way they will be marked but with no dependencies

baulig: Yeah mean keeping the `TypeBuilder` type, but not it's dependencies?

marek-safar: yeah, not even the ctors or fields (linker has this infrastructure in place already)

baulig:

baulig:

baulig:
@monojenkins build failed

Issue #14247, opened 2019-04-26
Assertion on 32-bit iOS device with thumb code


rolfbjarne:
## Steps to Reproduce

1. Build xamarin-macios/master as of today (for instance https://github.com/xamarin/xamarin-macios/commit/6c3674658aa39e95bbe83cb6a7d5dfccfea7dd3d)
2. `cd tests && make runner` to launch the test runner in the browser
3. Run the `monotouch-test/iOS Unified 32-bits - device/Release: UseThumb` test.

This requires an iOS device that supports 32-bit (i.e. with iOS 10.3 or earlier).

I currently don't know when this regressed, since we very recently started to run monotouch-test in thumb mode.

### Current Behavior

Assert at launch.

This shows in the device log:

> monotouchtest[1024] <Notice>: * Assertion at ../../../../../mono/mini/tramp-arm.c:1130, condition `(t1 >> 11) == 0x1e' not met

[Application Output](https://gist.github.com/rolfbjarne/6942c80096a5c593c2eebc6e6a2c130c)
[Crash report](https://gist.github.com/rolfbjarne/0e748f6687707d71b394ceaa615f7aa1)

### Expected Behavior

No crash

marek-safar:
@vargaz could you please look into this regression

vargaz:
What seems to happen is that llvm's llc doesn't generate thumb code even if we ask it to:
/Users/vargaz/mt/master/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/LLVM36/bin/llc"  -march=arm -mattr=+v7 -mattr=+vfp2,-neon,+d16 -asm-verbose=false -mtriple=thumbv7-ios -disable-gnu-eh-frame -enable-mono-eh-frame -mono-eh-frame-symbol=_mono_aot_mscorlib_eh_frame -relocation-model=pic -filetype=obj -o "/Users/vargaz/mt/master/xamarin-macios/tests/xharness/tmp-test-dir/monotouch-test1099/obj/iPhone/Release32-unified/mtouch-cache/armv7/mscorlib.dll-llvm.o" "/Users/vargaz/mt/master/xamarin-macios/tests/xharness/tmp-test-dir/monotouch-test1099/obj/iPhone/Release32-unified/mtouch-cache/armv7/mscorlib.dll.s.opt.bc

The -mtriple=thumb7-ios flag is supposed to force the generation of thumb code.

(__TEXT,__text) section
[...]

PR #14288, opened 2019-04-30
[aot] Avoid passing -march=arm to llc when using thumb, it forces arm32 code generation.

vargaz:
Fixes https://github.com/mono/mono/issues/14247.


vargaz:
@monojenkins backport 2019-02

lewurm:
we should untangle this code with some `ifdef TARGET_IOS`, `ifdef TARGET_LINUX`, etc because mixing that stuff makes it a maintenance nightmare.

jaykrell:
@lewurm but runtime `if` better than compile-time `#if` or `#ifdef` so one compiler can target multiple.

vargaz:
@monojenkins build failed

marek-safar:
@monojenkins backport 2019-04

Issue #14261, opened 2019-04-29
LLVM Generates Wrong IndexOf Span code


alexanderkyte:
This issue is a duplicate of https://github.com/mono/mono/issues/14143 . Making this issue to better track the history of this.

Some deep digging of the TechEmpower benchmark shows that the observed cause happens when we mis-parse the HTTP header. This happens because the offsets returned by the IndexOf operator to split the header by newlines doesn't work as expected. It returns  the index of the last newline, at offset 193 in the crashing reproduction. 

A gist with the relevant Span fixture is linked with executable instructions here: https://gist.github.com/alexanderkyte/d7bc0222cbcf7135f2513f99556ddd80

alexanderkyte:
LLVM IR for it here: https://gist.github.com/alexanderkyte/2734b8c98fbf0c191ace45b17de10875

alexanderkyte:
And this performs as expected when using fullAOT without llvm:

```
akyte@Alexanders-MacBook-Pro-3 ~/Projects/SpanCrash/SpanCrash/bin/Debug (master*) $ rm -rf mono_aot* *dylib* ; MONO_VERBOSE_METHOD="ConstStrLen" MONO_EXECUTABLE="/Users/akyte/mono_llvm_opt/mono/mini/mono-sgen --" MONO_PATH=/Users/akyte/mono_llvm_opt/mcs/class/lib/net_4_x /Users/akyte/mono_llvm_opt/runtime/mono-wrapper --aot-path=/Users/akyte/mono_llvm_opt/mcs/class/lib/net_4_x --aot=keep-temps SpanCrash.exe
Mono Ahead of Time compiler - compiling assembly /Users/akyte/Projects/SpanCrash/SpanCrash/bin/Debug/SpanCrash.exe
AOTID 6CF9A9C8-7DBA-99C9-CCED-7730AF6C7B16
Compiled: 4/4
Executing the native assembler: "clang"  -c -x assembler -o /var/folders/00/v8q_733j1hd8br7nxhxs41d80000gn/T/mono_aot_xm0HTG.o /var/folders/00/v8q_733j1hd8br7nxhxs41d80000gn/T/mono_aot_xm0HTG
Executing the native linker: clang --shared -o /Users/akyte/Projects/SpanCrash/SpanCrash/bin/Debug/SpanCrash.exe.dylib.tmp  /var/folders/00/v8q_733j1hd8br7nxhxs41d80000gn/T/mono_aot_xm0HTG.o 
[...]

Issue #14143, opened 2019-04-19
Techempower Platform Plaintext Hangs on Every Request using LLVM JIT


brianrob:
When running the .NET techempower platform plaintext benchmark, every request hangs when running using the LLVM JIT.  Things work properly using the built-in JIT.

## Steps to Reproduce

1. Build and run a container using this dockerfile: https://gist.github.com/brianrob/ba8214e89aef5406eaa8746ed83bf5f0#file-dockerfile.  This will start the ASP.NET service.
2. Exec into the container and run the following command to trigger a request:
```
curl --header "Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" --header "Connection: keep-alive" http://localhost:8080/plaintext
```

### Current Behavior

The HTTP request initiated by curl hangs.

### Expected Behavior

The HTTP request initiated by curl succeeds and returns "Hello, World!"

## On which platforms did you notice this

[ ] macOS
[ X ] Linux
[ ] Windows

**Version Used**:
```
Mono JIT compiler version 6.3.0.202 (tarball Fri Apr 12 18:28:05 UTC 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          yes(610)
        Suspend:       hybrid
        GC:            sgen (concurrent by default)
```
## Stacktrace

I've captured crash logs during the hang by sending a SIGABRT signal to the process running with LLVM JIT.  I've also captured crash logs after a successful request when using the built-in JIT:

[LLVM-JIT](https://gist.github.com/brianrob/ba8214e89aef5406eaa8746ed83bf5f0#file-llvm-jit-crashlog)
[BuiltIn-JIT](https://gist.github.com/brianrob/ba8214e89aef5406eaa8746ed83bf5f0#file-builtin-jit-crashlog)

This doesn’t look like a classic deadlock in that there are a bunch of threads backed up behind a single lock and one doing a bunch of work, or attempting to take another lock while holding the original lock.  In comparing the differences, I noticed that there are the same number of threads in both cases, and that all but one of the threads are stopped at the same location.  The difference is that in the good case, the newest thread is a parked worker thread, waiting for more work:
 ```
[...]

PR #14292, opened 2019-04-30
[llvm] Widen vector equality (OP_PCMPEQx)

alexanderkyte:
Fixes https://github.com/mono/mono/issues/14261
Fixes https://github.com/mono/mono/issues/14143

This was causing an issue when trying to inline
System.Numerics.Vectors.Vector`1<byte>:Equals into System.Span.IndexOf
from System.Memory.dll as provided by nuget. I believe that it is
comparing the result of a SIMD operation on two SIMD registers there with a
SIMD register full of zero bytes.

When we produce the Equals call (no inline), we spill the SIMD <4 x i32>
vector to the stack in the caller, and we then later load it
with the widened type used here <16 x i8>.

When we inline, we end up making the args into XMOV opcodes,
and then doing the "pcmpeqb" mini opcode with the return values
from those operations. This becomes a register to register no-op
copy, and the original <4 x i32> gets compared against the <16 x i8>
type:

```
  %116 = icmp eq <16 x i8> zeroinitializer, <4 x i32> %115
```

This does not go the way we want it to. It reliably leads to IndexOf
breaking and only finding the desired byte in the portion not processed
by the SIMD code (in the last few elements of the array).

This fix checks if the input dimensions differ, and does a conversion.
This produces the following IR:

```
  %114 = bitcast <4 x i32> %99 to <16 x i8>
  %115 = icmp eq <16 x i8> %114, %113
  %116 = sext <16 x i1> %115 to <16 x i8>
  %117 = icmp eq <16 x i8> zeroinitializer, %116
```

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -6723,26 +6723,30 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 		case OP_PCMPEQB:
 		case OP_PCMPEQW:
 		case OP_PCMPEQD:
-		case OP_PCMPEQQ: {
+		case OP_PCMPEQQ:
+		case OP_PCMPGTB: {
 			LLVMValueRef pcmp;
 			LLVMTypeRef retType;
+			int cmpOp;
+
+			if (ins->opcode == OP_PCMPGTB)
+				cmpOp = LLVMIntSGT;
+			else
+				cmpOp = LLVMIntEQ;
 
 			if (LLVMTypeOf (lhs) == LLVMTypeOf (rhs)) {
-				pcmp = LLVMBuildICmp (builder, LLVMIntEQ, lhs, rhs, "");
+				pcmp = LLVMBuildICmp (builder, cmpOp, lhs, rhs, "");
 				retType = LLVMTypeOf (lhs);
 			} else {
 				LLVMTypeRef flatType = LLVMVectorType (LLVMInt8Type (), 16);
 				LLVMValueRef flatRHS = convert (ctx, rhs, flatType);
 				LLVMValueRef flatLHS = convert (ctx, lhs, flatType);
-				pcmp = LLVMBuildICmp (builder, LLVMIntEQ, flatLHS, flatRHS, "");
+
+				pcmp = LLVMBuildICmp (builder, cmpOp, flatLHS, flatRHS, "");
 				retType = flatType;
 			}
 
 			values [ins->dreg] = LLVMBuildSExt (builder, pcmp, retType, "");
-			break;
-		}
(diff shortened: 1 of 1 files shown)


vargaz:
Looks ok but we should probably try to fix the underlying irregular IR. This will work for EQ opcodes, but not for GT etc.

jaykrell:
"Irregular underlying IR"? Meaning, this level should not have to `convert`, it should be presented with already matching types? All backends should expect that?

vargaz:
Where to we do the 16x8 load ?

alexanderkyte:
image

This is the caller-side change when we do the inline. 

The callee is:

```
; Function Attrs: uwtable
define monocc void @"System.Numerics.Vector:Equals<byte> (System.Numerics.Vector`1<16 x i8><byte>,System.Numerics.Vector`1<16 x i8><byte>)"(i64 %vret, i64 %arg_left, i64, i64 %arg_right, i64) #0 {
BB0:
  %2 = alloca <16 x i8>, align 16
  %3 = alloca <16 x i8>, align 16
  %4 = bitcast <16 x i8>* %3 to i8*
  %5 = getelementptr i8, i8* %4, i32 0
  %6 = bitcast i8* %5 to i64*
  store i64 %arg_left, i64* %6
  %7 = getelementptr i8, i8* %4, i32 8
  %8 = bitcast i8* %7 to i64*
  store i64 %0, i64* %8
  %9 = load <16 x i8>, <16 x i8>* %3
  %10 = bitcast <16 x i8>* %2 to i8*
  %11 = getelementptr i8, i8* %10, i32 0
  %12 = bitcast i8* %11 to i64*
  store i64 %arg_right, i64* %12
  %13 = getelementptr i8, i8* %10, i32 8
  %14 = bitcast i8* %13 to i64*
  store i64 %1, i64* %14
  %15 = load <16 x i8>, <16 x i8>* %2
  br label %BB3

BB3:                                              ; preds = %BB0
  br label %BB2

BB2:                                              ; preds = %BB3
  %16 = icmp eq <16 x i8> %9, %15
  %17 = sext <16 x i1> %16 to <16 x i8>
  %18 = add i64 %vret, 0
  %19 = inttoptr i64 %18 to <16 x i8>*
  store <16 x i8> %17, <16 x i8>* %19, align 1
  br label %BB1

BB1:                                              ; preds = %BB2
  ret void
}
```

alexanderkyte:
@vargaz I think the same policy should work for the greater than opcode. After analyzing what we do with the non-inlined `OP_PCMPGTB` code, it looks like the same vector dimension shifting happens there.

vargaz:
@monojenkins build failed

PR #14294, opened 2019-05-01
[netcore] clean up excludes-System.Runtime.Tests.rsp

EgorBo:
A lot of tests will be marked as "do not run on mono" in CoreFX (will submit a PR later).
Added comments for the active issues.
```
System.Runtime.Tests  Total: 32298, Errors: 0, Failed: 0, Skipped: 2, Time: 25.421s
```


netcore/excludes-System.Runtime.Tests.rsp
```diff
@@ -40,12 +41,6 @@
 # while the test passes, xunit fails in the end if it's enabled
 -nomethod System.Tests.Types.VoidTests.IsByRef_Get_ReturnsExpected
 
-# crashes runtime (Assertion at method-to-ir.c:12303, condition `load_opcode != OP_LOADV_MEMBASE' not met)
--nomethod System.Text.Unicode.Tests.Utf16UtilityTests.GetIndexOfFirstInvalidUtf16Sequence_WithSmallValidBuffers
--nomethod System.Text.Unicode.Tests.Utf16UtilityTests.GetIndexOfFirstInvalidUtf16Sequence_WithSmallInvalidBuffers
--nomethod System.Text.Unicode.Tests.Utf8Tests.ToChars_ValidBuffers
--nomethod System.Text.Unicode.Tests.Utf8Tests.ToBytes_WithLargeValidBuffers
```
filipnavara: When did this get fixed? I have Mono from few days ago and it was still failing on those tests.

EgorBo: Not sure, btw I've updated corefx version (tests) and can't reproduce these failures anymore

mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/NativeLibrary.cs
```diff
@@ -1,12 +1,24 @@
...
 	{
+		[DllImport ("libdl")]
+		private static extern IntPtr dlopen (string libName, int flags);
+		
+		private const int RTLD_LAZY = 0x001;
+		
 		static IntPtr LoadLibraryByName (string libraryName, Assembly assembly, DllImportSearchPath? searchPath, bool throwOnError) => throw new NotImplementedException ();
 
-		static IntPtr LoadFromPath (string libraryName, bool throwOnError) => throw new NotImplementedException ();
+		static IntPtr LoadFromPath (string libraryName, bool throwOnError) {
```
EgorBo: Had to implement it because `PlatformDetection` in corefx now uses it to load some gdi library: https://github.com/dotnet/corefx/blob/master/src/CoreFx.Private.TestUtilities/src/System/PlatformDetection.Unix.cs#L66-L78

mono/metadata/runtime.c
```diff
@@ -55,9 +55,23 @@ static void
...
+	MonoMethod *procexit_method;
+
+	appcontext_class = mono_class_try_load_from_name (mono_defaults.corlib, "System", "AppContext");
+	g_assert (appcontext_class);
+	
+	procexit_method = mono_class_get_method_from_name_checked (appcontext_class, "OnProcessExit", 0, 0, error);
+	g_assert (procexit_method);
+	
+	mono_runtime_try_invoke (procexit_method, NULL, NULL, &exc, error);
+#else
```
EgorBo: CoreCLR calls AppContext.OnProcessExit on exiting (OnProcessExit fires ProcessExit event, forwards it to AppDomain then).

EgorBo:
@marek-safar @vargaz Could you please review? 
I'll delete all the tests I listed in `# TODO: ignore in CoreFX for mono runtime:` in a separate PR once I "ignore" them directly in CoreFX repo.
BTW, there is a failing test which tries to load Type from a string:
`"System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]"`
Our mscorlib is `0.0.0.0` are we going to change the version to be 4.0.0.0?

mcs/class/System.Private.CoreLib/System.Runtime.InteropServices/NativeLibrary.cs
filipnavara: Implementing it this way will break the Windows builds. :-/

EgorBo: oh I always forget 😄copy-pasted the Unix impl for tests

filipnavara: I am fine if it remains unimplemented (ie. always returns error) but I would prefer not to break the builds ;)

filipnavara:
> Our mscorlib is 0.0.0.0 are we going to change the version to be 4.0.0.0?

Isn't that `mscorlib` facade from CoreFX?

netcore/excludes-System.Runtime.Tests.rsp
filipnavara: I'll retest it and report back. It should not hold up this PR though.

PR #14298, opened 2019-05-01
ld -Bsymbolic so dynamic linker does not get between mono and itself.

jaykrell:
With this change, within libmonosgen.so, calls from
‘mono_x’ to ‘mono_y’ are made directly instead of going through ‘mono_y@plt’.

This is the same as the mono executable already and the same as Windows.
Mac/iOS might also already be this way?
This definitely changes Linux.

No other shared object, nor the executable can therefore replace ‘mono_y’. Does anyone do that?

It is then much more understood where locks are, making coop and crash reporting more correct and reliable.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4940,14 +4940,15 @@ AC_SEARCH_LIBS(dlopen, dl)
 old_ldflags="${LDFLAGS}"
 # GNU specific option, this confuses IBM ld, but do offer alternatives when possible
 if test $lt_cv_prog_gnu_ld = yes; then
-	LDFLAGS="${LDFLAGS} -Wl,-export-dynamic"
+	LDFLAGS="${LDFLAGS} -Wl,-export-dynamic,-Bsymbolic"
 else
 	case $host in
 	*-*-aix*|*-*-os400*)
+		LDFLAGS="${LDFLAGS} -bsymbolic"
 		;;
 	*)
 		dnl Try to use export-dynamic anyways
-		LDFLAGS="${LDFLAGS} -Wl,-export-dynamic"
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
I tested on Mac and I don't think this changes it. The default is already equivalent to this.
I assume iOS is the same.
Mac has -interposable that seems to be the opposite but I couldn't get that to work yet -- we don't want it anyway.
I tentatively conclude then that this is a good change for Linux, that makes it more like other platforms.

jaykrell:
Maybe need to check Android also. Could be different than Linux.

jaykrell:
Not having checked Android, it is hard to imagine this is a bad change.
Some platforms are already this way -- Windows, OSX, probably iOS, Linux executable (but not .so).
Making things this way is generally a very good thing.
Consistency is somewhat valuable.
The only downside is the loss of interposability, that is already spotty, and arguably interposabilty is quite undesirable anyway.


Once this is then, I want to remove function pointer linkage between metadata and mini and just use direct calls. The function pointers presently serve a purpose, because getting them is interposable but calling them is not. This will make calling directly not interposable either and the pointers can go away.

vargaz:
Hopefully this won't break anything.

NattyNarwhal:
@jaykrell I think the dynamic linker on AIX is particular when it's using svr4 style dynamic linking. I actually attempt to deal with that here: https://gist.github.com/NattyNarwhal/b2d8ad27d40a7aef5be4b321493792ef

PR #14303, opened 2019-05-01
WIP [bcl] Introduce new 'monodroid_designer' profile.

alexischr:
We will be building and including this profile so that the Xamarin.Android Designer can use it as part of its simulation of  a monodroid environment.

The Designer used to load the monodroid mscorlib.dll (and other assemblies), and these used to be only platform-independent code. This is no longer the case with our BCL, so we need to bundle separate BCL builds for each Xamarin.Android Designer platform.


mcs/build/profiles/monodroid_designer.make
```diff
@@ -0,0 +1,36 @@
+#! -*- makefile -*-
```
alexischr: This is a carbon copy of `monodroid.make`

mcs/class/Makefile
```diff
@@ -424,6 +424,9 @@ wasm_tools_PARALLEL_SUBDIRS := $(wasm_tools_dirs_parallel)
 netcore_SUBDIRS := System.Private.CoreLib
 netcore_PARALLEL_SUBDIRS :=
 
+monodroid_designer_SUBDIRS := $(mobile_common_dirs)
+monodroid_designer_PARALLEL_SUBDIRS := $(monodroid_dirs_parallel)
```
alexischr: This could be "minimized", using our knowledge that not all assemblies are platform-independent and some can be re-used by the designer from the monodroid profile. But that will be brittle.

jaykrell:
“and these used to be only platform-independent code. “

Fix that? :(

mcs/build/profiles/monodroid_designer.make
alexischr: Changes were made: 

* The API_BIN_PROFILE was changed to 4.7.1 in order to find `System.Configuration`
* The `PLATFORM_MACOS` conditional was added to allow for a proper macos build

lambdageek:
https://github.com/mono/corefx/compare/5fe0364915df0f3568d14fb4c36213f7658b296d...4fe560b6a9b0a5c64ee9ed41215b4560f398ab93#diff-3d1d44833fd15bcac36bfe87e5d29146R19
That `if` should be `#if`

marek-safar:
I am not sure this is addressing the issue after all. Could any of the designed folks tell us what code do they want (@garuma @alanmcgovern )? Reading some of the conversation the build should be windows specific so the config file needs to reflect that.

Further, where will they get the correct native System.Native from? What assembly loader mode will this use desktop or mobile?

lambdageek:
> the build should be windows specific so the config file needs to reflect that.

@marek-safar  we address that by overriding `PROFILE_PLATFORM` in #14307 when we cross-compile the `monodroid_designer` profile for `win32` here: https://github.com/mono/mono/pull/14307/commits/fd4b8e094aeb51055691dd61b8ce3b3d13dc5bbc#diff-afc068418723845a82626f3774799052R99

alexischr:
We will be providing a Windows-hosted monodroid BCL instead: https://github.com/mono/mono/pull/14424

PR #14307, opened 2019-05-01
[sdks] Build a win32 monodroid_designer BCL into the android archive

lambdageek:
WIP.  Depends on #14303 

Just the most recent commit (with changes in the `sdks/builds/` directory) is part of this PR proper.

There's a couple of hacks in there:
1. `-$$(MAKE)` on the `build-custom-$(1)-bcl-cross-$(2):` should be changed to remove the `-` after all the assemblies build correctly.  right now it just collects whatever manages to build - as a proof of concept.
2. The same step ignores the test assemblies.  We should either get rid of the tests argument, or else really build them.
3. Right now we build the `bcl-cross-$(platform)` builds out of the same `bcl/` tree as the normal `BclTemplate`.  It would be better not to do that in case we accidentally depend on previous builds.

/cc @alexischr @akoeplinger


sdks/builds/android.mk
```diff
@@ -351,4 +351,5 @@ $(eval $(call AndroidCrossMXETemplate,cross-arm64-win,x86_64,aarch64-v8a,arm64-v
 $(eval $(call AndroidCrossMXETemplate,cross-x86-win,i686,i686,x86,llvm-llvmwin32,i686-none-linux-android))
 $(eval $(call AndroidCrossMXETemplate,cross-x86_64-win,x86_64,x86_64,x86_64,llvm-llvmwin64,x86_64-none-linux-android))
 
-$(eval $(call BclTemplate,android,monodroid monodroid_tools,monodroid monodroid_tools))
+$(eval $(call BclTemplate,android,monodroid monodroid_tools monodroid_designer,monodroid monodroid_tools))
```
alexischr: You want to remove `monodroid_designer` from this line then, and I'll remove it on my PR too

lambdageek:
Got a [android-release archive (660mb zip)](https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx-products-sdks-android/11535/Azure/processDownloadRequest/11535/android-release-Darwin-53312669dfd30e9350a0874d108de3141f858d14.zip)  that has a `android-bcl-cross-win32/monodroid_designer/mscorlib.dll` in it that does not have a System.Native dependency

Other assemblies for the profile still need work.

jonpryor:
Naming request: could we call this new profile `monodroid_desktop` and not `monodroid_designer`?  The Xamarin.Android team would like to use these new profile assemblies on desktop for various utilities.

lambdageek:
Instead of trying to cross-compile the monodroid_designer profile, we're going to work on making SDK infrastructure runnable on windows (#14333 )

PR #14309, opened 2019-05-02
[wasm] generate icall_trampoline_dispatch as nested branches

lewing:
Just an experiment to replace the generated list of strcmp calls with nested ifs.


vargaz:
Note that this function is not called very often in theory.

lewing:
yeah, I don't expect much real impact I just wanted to put this somewhere

mono/mini/m2n-gen.cs
```diff
@@ -174,7 +176,54 @@ class Driver {
...
+	}
+
+	static void WriteDispatch (string [] items)
+	{
+		Array.Sort (items);
+		Console.WriteLine ("static void\nicall_trampoline_dispatch (const char *cookie, void *target_func, InterpMethodArguments *margs)");
+		Console.WriteLine ("{");
+		Partition (items, 0);
+		Console.WriteLine ("\tprintf (\"CANNOT HANDLE COOKIE %s\\n\", cookie);");
+		Console.WriteLine ("\tg_assert (0);");
```
lewurm: This should use `g_error ("...")`. In general, `printf` doesn't work as expected on every platform so we use `g_print` instead. `g_error` will use it accordingly.

lewing:

lewing:

vargaz:
Turns out this function is called often, since instead of returning a function pointer, its actually makes the call.

PR #14315, opened 2019-05-02
[corlib] Avoid a runtime crash on new ArgIterator(new RuntimeArgumentHandle()).

vargaz:


EgorBo:
`System.Tests.ArgIteratorTests.ArgIterator_Throws_PlatformNotSupportedException` still fails but I guess mostly because CoreFX expects `ArgIterator` to throw PNSE for all members on Unix. https://github.com/dotnet/corefx/blob/0a751af9afe28042a720411925e6e6c55bdc10cc/src/System.Runtime/tests/System/ArgIteratorTests.netcoreapp.cs#L91 so `PlatformDetection.IsArgIteratorNotSupported` should return false on Mono

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14319, opened 2019-05-02
[wasm] Define wasm_strdup when setting options

lewing:
fix mono_wasm_set_runtime_options so that it can be used outside of the test app


lewing:
@monojenkins build failed

PR #14323, opened 2019-05-02
Throw ObjectDisposedException instead of crashing when Cairo.Context handle == IntPtr.Zero

leculver:
@marek-safar Can you please help find a dev to shepherd this change through?  It has **not** been built or tested...unfortunately I'm not quite set up for full mono development.  It should also be code reviewed by someone in this space.

We are seeing instances where VS for Mac is crashing with a SIGSEGV when
calling into certain Cairo.Context functions.  Looking at the code, it's clear
that it's possible to have Context.handle == IntPtr.Zero after disposing the
object.

From research, it looks like passing a nullptr to raw cairo functions can
result in a SIGSEGV:
    https://bugs.launchpad.net/ubuntu/+source/compiz/+bug/93807

This change adds check to see if Conext.handle == IntPtr.Zero in Context
painting, font, and path methods.  This will turn uncatchable SIGSEGVs
into ObjectDisposedException.

### DIFF
--- a/mcs/class/Mono.Cairo/Mono.Cairo/Context.cs
+++ b/mcs/class/Mono.Cairo/Mono.Cairo/Context.cs
@@ -313,9 +313,14 @@ namespace Cairo {
 
 		public void SetTarget (Surface target)
 		{
+			IntPtr newHandle = NativeMethods.cairo_create (target.Handle);
+			if (newHandle == IntPtr.Zero)
+				throw new InvalidOperationException ("Surface had an invalid handle.");
+
 			if (handle != IntPtr.Zero)
 				NativeMethods.cairo_destroy (handle);
-			handle = NativeMethods.cairo_create (target.Handle);
+
+			handle = newHandle;
 		}
 
 		[Obsolete("Use GetScaledFont/SetScaledFont")]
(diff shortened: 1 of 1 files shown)


marek-safar:
I think we should fix also the places where null/IntPtr.Zero can be assigned not to incorrectly throw DisposedException.

Looking at the code that should be only in ctor and SetTarget method. Could you add the IntPtr.Zero check there as well?

mcs/class/Mono.Cairo/Mono.Cairo/Context.cs
```diff
@@ -49,7 +49,7 @@ public class Graphics : Context {
 
 	public class Context : IDisposable
 	{
-		IntPtr handle = IntPtr.Zero;
+		IntPtr handle;
```
leculver: This is a no-op since objects are nulled on creation.  I will add this to the PR for clarity though.

leculver:
@marek-safar Hmm, I'm not sure I'm understanding.

		public void SetTarget (Surface target)
		{
			if (handle != IntPtr.Zero)
				NativeMethods.cairo_destroy (handle);
			handle = NativeMethods.cairo_create (target.Handle);
		}

What change would you propose here?

marek-safar:
We could throw InvalidOperationException when `NativeMethods.cairo_create (target.Handle);` returns IntPtr.Zero

marek-safar:
@monojenkins backport 2019-04

marek-safar:
@monojenkins backport 2019-02

PR #14325, opened 2019-05-03
Fix light on light ListViewGroup headers on dark theme

HebaruSan:
## Problem

If you use a ListViewGroup with a dark theme, the text is light on light:

image

## Cause

It's using `SystemBrushes.ControlText`, which is not guaranteed to contrast with the background, which is `SystemColor.Window` and contrasts with `WindowText`.

## Changes

On Windows these are drawn in a nice shade of blue:

image

So now mono will use `SystemBrushes.Highlight` to mimic that nice blue:

image

I think `Highlight` will contrast well with `Window`, and also look better in a dark theme.

(I was not able to get my locally compiled mono to obey a dark theme, but I'm 100% sure that's not caused by this change. If there are any known quirks related to that, I'd love some tips on how to fix it.)


marek-safar:
approve

steveisok:
@monojenkins build failed

PR #14334, opened 2019-05-03
[netcore] Publish to BAR too

directhex:
This should make our builds available in the BAR/Maestro++

Relatedly, this also changes the job to a matrix one, because that was required for the Arcade template I'm using, so it should be easier to add non-x86-Linux to the build now.


directhex:
@monojenkins build failed

directhex:
@monojenkins build failed

PR #14336, opened 2019-05-03
[netcore] Fix HideDetectionHappensBeforeBindingFlagChecks

EgorBo:
```csharp
class Base
{
    public int MyProp1 { get; }
    public int MyProp2 { get; }
}

class Derived : Base
{
    private new int MyProp1 { get; }
}
```
```csharp
typeof(Derived).GetProperty("MyProp1") -- should be NULL
typeof(Derived).GetProperty("MyProp2") -- should not be NULL
```
(in Mono both properties are not null)

I copied this piece of logic from CoreCLR, e.g.:
1) https://github.com/dotnet/coreclr/blob/7a24a538cd265993e5864179f51781398c28ecdf/src/System.Private.CoreLib/src/System/RtType.cs#L2022-L2053
2) https://github.com/dotnet/coreclr/blob/7a24a538cd265993e5864179f51781398c28ecdf/src/System.Private.CoreLib/src/System/RtType.cs#L241-L244

In Mono `RuntimePropertyInfo. BindingFlags` was always `0` for some reason.


EgorBo:
@monojenkins build failed

PR #14340, opened 2019-05-03
[wasm] Add --link-descriptor arg to packager

lewing:
Allow a --link-descriptor arg to supply a linker.xml configuration file


lewing:
@monojenkins build failed

PR #14348, opened 2019-05-04
[wasm] Add a function to return the interp trampoline instead of invoking it. Not used yet.

vargaz:

### DIFF
--- a/mono/mini/aot-runtime-wasm.c
+++ b/mono/mini/aot-runtime-wasm.c
@@ -90,6 +90,26 @@ get_long_arg (InterpMethodArguments *margs, int idx)
 
 #include "wasm_m2n_invoke.g.h"
 
+gpointer
+mono_wasm_get_interp_to_native_trampoline (MonoMethodSignature *sig)
+{
+	char cookie [32];
+	int c_count;
+
+	c_count = sig->param_count + sig->hasthis + 1;
+	g_assert (c_count < sizeof (cookie)); //ensure we don't overflow the local
+
+	cookie [0] = type_to_c (sig->ret);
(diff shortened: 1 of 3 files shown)


vargaz:
The interpreter needs to be modified to create some kind of preprocessed data structure to invoke functions of a given signature, this is the first step towards that.

mono/mini/aot-runtime-wasm.c
```diff
@@ -90,6 +90,26 @@ get_long_arg (InterpMethodArguments *margs, int idx)
...
 #include "wasm_m2n_invoke.g.h"
 
+gpointer
+mono_wasm_get_interp_to_native_trampoline (MonoMethodSignature *sig)
+{
+	char cookie [32];
+	int c_count;
+
+	c_count = sig->param_count + sig->hasthis + 1;
+	g_assert (c_count < sizeof (cookie)); //ensure we don't overflow the local
```
jaykrell: alloca it instead, or alloca if >32?

lewurm:
do I understand correctly that you want instead of this
```
interp_main_loop
  -> icall_trampoline_dispatch (cookie, target_func, margs)
  -> wasm_invoke_dd (target_func, margs)
```
have something that directly calls the `wasm_invoke`, so
```
interp_main_loop
  -> wasm_invoke_dd (target_func, margs)
```
that is, moving the `icall_trampoline_dispatch` to the transform phase?

vargaz:
Something like that, need to create a per-signature data structure and embed it into the interpreter code stream.

lewing:
@vargaz do you want to keep this open?

PR #14277, opened 2019-04-30
[do not merge] Work towards removing JIT icall hashing.

jaykrell:
More gradual work is requested.

Based on:
 https://github.com/mono/mono/pull/11941#issuecomment-449157323

Hash lookups can often/always be replaced
by directly referencing a struct,
or by indexing into a very dense array.

This PR is a stepping stone. The steps can be made smaller or larger.
There is definitely more to do.

This is https://github.com/mono/mono/pull/14047
but with https://github.com/mono/mono/pull/14151#issuecomment-487832605
used to reduce the diff.

This does not yet reduce the hash inserts/lookups, but it is progress toward that.

PR #14359, opened 2019-05-06
Introduce enum and static data for JIT icalls, that are currently hashed

jaykrell:
by string and address and use dynamic allocation.

These will be used shortly.


jaykrell:
Part of https://github.com/mono/mono/pull/14277.

jaykrell:
@monojenkins build failed

mono/metadata/class-internals.h
```diff
@@ -1469,6 +1469,347 @@ mono_class_init_checked (MonoClass *klass, MonoError *error);
 MonoType*
 mono_class_enum_basetype_internal (MonoClass *klass);
 
+// Eventually, changes within MONO_JIT_ICALLS might require revising MONO_AOT_FILE_VERSION.
+// i.e. if AOT persists enum/int instead of string. For now, this is not the case.
```
vargaz: These should go to the previous jit-icall-reg.h file, and that file should be included into this file.

jaykrell: Ok. I didn't like the name of it, but a separate file I do like.
I'll do jit-icall-reg.h as you say and it can renamed if that was just quick typing.
The include will probably at the end "like this".

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/jit-icall-reg.h
```diff
@@ -0,0 +1,353 @@
...
+ *
+ * Author:
+ *   Jay Krell (jaykrell@microsoft.com)
+ *
+ * Copyright 2019 Microsoft
+ * Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+// No include guard needed.
+
```
vargaz: You can add an include guard anyway for consistency.

PR #14361, opened 2019-05-06
Make direct calls when looking up in icall tables, instead of using function pointers (subject to ifdef).

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
- the ifdef do not motivate the function pointers 
- this *somewhat* depends on ld -Bsymbolic becuase without it, the dynamic linker interception is moved to initialization instead of runtime, which might matter in terms of when random locking is inserted

jaykrell:
Probably broke webassembly. Will investigate/fix later.

jaykrell:
@monojenkins build WebAssembly

jaykrell:

vargaz:
The reason for the function pointers is to be able to put the icall table into a separate library which can be linked with the normal libmonosgen by calling the register_ function.

jaykrell:
You don’t need function pointers for that, IF that separation is known at link time and the functions are either in the same binary or not in the same. If they are sometimes in the same binary, but can be optionally replaced, then writable pointers are needed.
The compiler doesn’t need this decision known, the call can look the same either way. But the static linker does need to know. Except under some situations i.e. Linux .so functions can still ve replaced even if in same binary.

jaykrell:
`mono_install_icall_table_callbacks` is not exported, so I'm confused.
And still the WebAssembly break I'll look at.

jaykrell:
If it was exported, if we do have replacable function pointers, then we should take a pointer to a constant struct of function pointers and remember the pointer -- no copy, no writability.

jaykrell:
driver.c aha/duh. I limit my search to mono directory, missed sdk.

jaykrell:
Are the icall tables meant to be replacable by anyone, or just by one of the two implementations within mono? I mean, anyone who statically links to the runtime? I don't think it is allowed via dynamic linkage due to symbol visibility. Granted, that is easy to change if you are willing to relink the runtime.

jaykrell:
So, the caller of mono_icall_table_lookup doesn't know if the static link contains any mono_icall_table_lookup or not, leading to
```
error: undefined symbol: mono_icall_table_lookup
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
```
but perhaps it can know.

jaykrell:
I guess this is borderline or actually not doable, without further ifdef.
Actual separate shared libraries are built, and presumably used, and must be shared.
I will probably close after the current CI iteration.

jaykrell:
@monojenkins build failed

jaykrell:
Giving up for now.
It looks like wasm sometimes/always has the tables builtin, but they get replaced anyway.
Likely the separate libraries can be static-only, which should help, but I waffled on that.
I need to be able to run/test/debug this locally more. Or at least open a draft PR to not send GitHub notifications.

PR #14362, opened 2019-05-06
[arm] fix armv7k regression on struct passing

lewurm:
We can't use `slotsize=8` on `armv7k` as it leads to stack corruption.

Note that `mtouch` already calls the AOT compiler with `--aot=mtriple=arm64_32-ios`, so no changes are required there.

Regression introduced by https://github.com/mono/mono/pull/12992

Fixes https://github.com/mono/mono/issues/13454

### DIFF
--- a/mono/mini/mini-arm.c
+++ b/mono/mini/mini-arm.c
@@ -88,6 +88,8 @@ static gboolean thumb2_supported = FALSE;
  */
 static gboolean eabi_supported = FALSE;
 
+static gboolean arm64_32_abi = FALSE;
+
 /* 
  * Whenever to use the iphone ABI extensions:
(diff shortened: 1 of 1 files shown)


mono/mini/mini-arm.c
```diff
@@ -7468,6 +7468,9 @@ mono_arch_set_target (char *mtriple)
 	}
 	if (strstr (mtriple, "gnueabi"))
 		eabi_supported = TRUE;
+
+	if (strstr (mtriple, "arm64_32"))
```
jaykrell: Cosmetic: Is “abi” useful in this identifier?

lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2019-04

PR #14368, opened 2019-05-06
[debugger] Deadlock while debugging a process

thaystg:
A deadlock happens when during a process_suspend there is a pending invoke and then when it's calling the invoke there is another call to process_suspend in the same thread.

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -2776,6 +2776,8 @@ try_process_suspend (void *the_tls, MonoContext *ctx)
 		/* Fastpath during invokes, see in process_suspend () */
 		if (suspend_count - tls->resume_count == 0)
 			return FALSE;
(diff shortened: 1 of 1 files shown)


thaystg:
@monojenkins build failed

alanmcgovern:
@marek-safar @vargaz 

I think this is worth putting in the release branch(es). This issue is quite common but random, and with the testcase i provided @thaystg the debugger deadlocked within 10 seconds of use every single time I tried it.

lewurm:
@monojenkins backport 2019-04
@monojenkins backport 2019-02
@monojenkins backport 2018-10

alanmcgovern:
2018-08 is the current build we're using on desktop. can we backport to there too?

thaystg:
@monojenkins backport 2018-08

thaystg:
@monojenkins backport 2018-08

PR #14369, opened 2019-05-06
[netcore] Set of workarounds for ArrayTests

EgorBo:
This PR enables >200 tests in System.Runtime.Tests.
 1.
```csharp
public enum SByteEnum : sbyte
{
    MinusTwo = -2,
    Zero = 0,
    Five = 5
}

SByteEnum[] arr = new SByteEnum[3];
arr.SetValue((sbyte) 1, 0);  // Expected: InvalidCast!
arr.CopyTo(new sbyte[3], 0); // Expected: success
```
Mono used to allow both (looks more logical).

2.
```csharp
Array sourceArray = new object[] { null };
Array destinationArray = new int?[] { 0 };
Array.Copy(sourceArray, 0, destinationArray, 0, 1);
```
Mono used to throw Exception here (that's why I added `Nullable.GetUnderlyingType`).

3.
In .NET Core it's allowed to copy array of `ushort` to `char` (`DefaultBinder.CanChangePrimitive` returns false for such conversion, but returns true for `char` to `ushort`).

4. if `#nullable enable` is defined in one of the `partial` file - it should be defined everywhere else (didn't compile without it with some weird error).

```
=== TEST EXECUTION SUMMARY ===
   System.Runtime.Tests  Total: 32502, Errors: 0, Failed: 0, Skipped: 2, Time: 25.303s
```

### DIFF
--- a/mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
+++ b/mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
@@ -1,5 +1,7 @@
 using System.Runtime.CompilerServices;
 
+#nullable enable
+
 namespace System.Collections.Generic
 {
 	partial class Comparer<T>
--- a/mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
+++ b/mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
@@ -1,6 +1,6 @@
-using System.Runtime.CompilerServices;
-
 #nullable enable
+
+using System.Runtime.CompilerServices;
 
 namespace System.Collections.Generic
 {
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -154,7 +154,7 @@ icallarray_print (const char *format, ...)
 static GENERATE_GET_CLASS_WITH_CACHE (module, "System.Reflection", "Module")
 
 static void
-array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, MonoError *error, gboolean strict);
+array_set_value_impl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, gboolean strict, MonoError *error);
 
 static MonoArrayHandle
 type_array_from_modifiers (MonoImage *image, MonoType *type, int optional, MonoError *error);
@@ -252,7 +252,7 @@ ves_icall_System_Array_GetValue (MonoArrayHandle arr, MonoArrayHandle indices, M
 void
 ves_icall_System_Array_SetValueImpl (MonoArrayHandle arr, MonoObjectHandle value, guint32 pos, MonoError *error)
 {
-	array_set_value_impl (arr, value, pos, error, FALSE);
+	array_set_value_impl (arr, value, pos, FALSE, error);
 }
 
 static inline void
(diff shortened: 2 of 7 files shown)


mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
```diff
@@ -1,5 +1,7 @@
 using System.Runtime.CompilerServices;
 
+#nullable enable
```
marek-safar: nit: should be the first code line in the file. I don't think we have to add it now as the feature will can be enabled globally for the project soon

mono/metadata/icall.c
```diff
@@ -409,10 +411,18 @@ array_set_value_impl (MonoArrayHandle arr_handle, MonoObjectHandle value_handle,
 
 	vsize = mono_class_value_size (vc, NULL);
 
-	if (et == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (ec)->data.klass))
+	gboolean et_isenum = et == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (ec)->data.klass);
+	gboolean vt_isenum = vt == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (vc)->data.klass);
```
jaykrell:
```suggestion
	gboolean vt_isenum;
	vt_isenum = vt == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (vc)->data.klass);
```

mono/metadata/icall.c
```diff
@@ -409,10 +411,18 @@ array_set_value_impl (MonoArrayHandle arr_handle, MonoObjectHandle value_handle,
 
 	vsize = mono_class_value_size (vc, NULL);
 
-	if (et == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (ec)->data.klass))
+	gboolean et_isenum = et == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (ec)->data.klass);
```
jaykrell:
```suggestion
	gboolean et_isenum;
	et_isenum = et == MONO_TYPE_VALUETYPE && m_class_is_enumtype (m_class_get_byval_arg (ec)->data.klass);
```

mono/metadata/icall.c
```diff
@@ -263,7 +263,7 @@ set_invalid_cast (MonoError *error, MonoClass *src_class, MonoClass *dst_class)
 }
 
 static void
-array_set_value_impl (MonoArrayHandle arr_handle, MonoObjectHandle value_handle, guint32 pos, MonoError *error)
```
jaykrell: `error` usually goes last, and might even need.

mono/metadata/icall.c
```diff
@@ -312,18 +312,20 @@ array_set_value_impl (MonoArrayHandle arr_handle, MonoObjectHandle value_handle,
 
 #ifdef ENABLE_NETCORE
 #define WIDENING_MSG NULL
```
jaykrell: Suggest local variable instead of macros.

mcs/class/System.Private.CoreLib/System.Collections.Generic/Comparer.cs
EgorBo: without it I get `error CS0265: Partial declarations of 'ArraySortHelper<T>' have inconsistent constraints for type parameter 'T'` when I try to build S.P.Corlib. (I am building it with .net core 3.0.100-preview6-011673)

EgorBo: i mean yeah I understand that a global switch will help but the error text is weird

mono/metadata/icall.c
EgorBo: WIDENING_MSG is then used in a different macro

EgorBo:
@monojenkins build failed

jaykrell:
There are C++ errors.

EgorBo:
@monojenkins build failed

EgorBo:
@monojenkins build failed

PR #14370, opened 2019-05-07
[wasm] Add a mixed target for sample

lewing:

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -197,6 +197,13 @@ build-aot-sample: packager.exe hello.exe
 	mono --debug packager.exe --emscripten-sdkdir=$(EMSCRIPTEN_SDKDIR) --mono-sdkdir=$(TOP)/sdks/out -appdir=bin/aot-sample --nobinding --builddir=obj/aot-sample --aot --template=runtime-tests.js hello.exe
 	ninja -v -C obj/aot-sample
 
+build-mixed-sample: packager.exe hello.exe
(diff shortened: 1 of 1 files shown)


sdks/wasm/Makefile
```diff
@@ -197,6 +197,13 @@ build-aot-sample: packager.exe hello.exe
 	mono --debug packager.exe --emscripten-sdkdir=$(EMSCRIPTEN_SDKDIR) --mono-sdkdir=$(TOP)/sdks/out -appdir=bin/aot-sample --nobinding --builddir=obj/aot-sample --aot --template=runtime-tests.js hello.exe
 	ninja -v -C obj/aot-sample
 
+build-mixed-sample: packager.exe hello.exe
+	mono --debug packager.exe --emscripten-sdkdir=$(EMSCRIPTEN_SDKDIR) --aot-interp --aot-assemblies=mscorlib --mono-sdkdir=$(TOP)/sdks/out -appdir=bin/mixed-sample --nobinding --builddir=obj/mixed-sample --template=runtime-tests.js hello.exe
```
kg: minor nit: do we need to quote emscripten-sdkdir, given how historically quirky emscripten is? I don't consider it a blocker since we already pass the argument this way elsewhere, but I'd wonder if this handles folders with spaces in them

lewing: that's a valid point, easy enough to test

lewing: turns out even the basic targets don't work with spaces in the mono directory name

kg: It sounds like maybe we should fix that in a separate PR or file an issue then

lewing: #14371

lewing:

PR #14375, opened 2019-05-07
Mutex memory leak on sgen GC (nursery, major).

lateralusX:
During GC we allocate a gray queue protected by a mutex, sgen_gray_object_queue_init. This mutex is never destroyed when clearing up the object containing it, sgen_gray_object_queue_dispose, so it will leak. Overtime this could cause OOM on some systems depending on how memory related to system primitives are handled and partitioned.


BrzVlad:
@monojenkins backport 2019-02
@monojenkins backport 2019-04

PR #14378, opened 2019-05-07
Update SecurityProtocolType to support Tls 1.3

woutermeuwis:

### DIFF
--- a/mcs/class/Mono.Security/Mono.Security.Protocol.Tls/SecurityProtocolType.cs
+++ b/mcs/class/Mono.Security/Mono.Security.Protocol.Tls/SecurityProtocolType.cs
@@ -39,8 +39,9 @@ namespace Mono.Security.Protocol.Tls
 		Default = -1073741824,
 		Ssl2	= 12,
 		Ssl3	= 48,
-		Tls		= 192,
+		Tls	= 192,
 		Tls11	= 768,
 		Tls12	= 3072,
+		Tls13	= 12288
 	}
 }


spouliot:
It's the `System.Net` (not `Mono.Security`) copy of the enum that needs the update.

The file you want to update is `external/corefx/src/System.Net.ServicePoint/src/System/Net/SecurityProtocolType.cs` which is in the corefx repo (and a more recent version, than the current revision used by mono, might already have the value).

SamuelDebruyn:
@spouliot It seems to be already updated, to which commit should the CoreFX submodule be updated?

spouliot:
@SamuelDebruyn bumping the revision of `corefx` has a lot of implications. 
@marek-safar might be able to tell you when the next bumps will happen (and if they go over the one that include the change you're looking for)

marek-safar:
As @spouliot said the correct place to update the value is in mono/corefx repo, please make PR there with suggested change

woutermeuwis:
> As @spouliot said the correct place to update the value is in mono/corefx repo, please make PR there with suggested change

@marek-safar As mentioned above by @SamuelDebruyn it is already updated in there (november 2018). So in order to get those changes we would need the version bumped up. Can you give a timeline on when to expect this?

marek-safar:
I still don't see them in https://github.com/mono/corefx/blob/master/src/System.Net.ServicePoint/src/System/Net/SecurityProtocolType.cs

akoeplinger:
Should be fixed by https://github.com/mono/mono/pull/14660

hiradyazdan:
@akoeplinger is this PR merged for latest mono and available on mono runtime 4.0.3? If so, I'm curious as to whether this would be available on latest Unity as well?

akoeplinger:
@hiradyazdan it's available in Mono 6.4.0.190 and later. I don't know when it will show up in Unity, you need to ask them :)

PR #14385, opened 2019-05-07
Don't apply MONO_ATTR_FORMAT_PRINTF to mono_error_set_argument

directhex:
The constructor for mono_error_set_argument lacks a `const char *msg_format` so clearly we shouldn't be applying formatting from an entirely unrelated parameter.

This was flagged by `-Werror=format-security`

### DIFF
--- a/mono/utils/mono-error-internals.h
+++ b/mono/utils/mono-error-internals.h
@@ -164,7 +164,7 @@ void
 mono_error_set_argument_format (MonoError *error, const char *argument, const char *msg_format, ...) MONO_ATTR_FORMAT_PRINTF(3,4);
 
 void
-mono_error_set_argument (MonoError *error, const char *argument, const char *msg, ...) MONO_ATTR_FORMAT_PRINTF(3,4);
+mono_error_set_argument (MonoError *error, const char *argument, const char *msg);
 
 void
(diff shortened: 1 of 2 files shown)


directhex:
@monojenkins build failed

mono/utils/mono-error-internals.h
```diff
@@ -164,7 +164,7 @@ void
 mono_error_set_argument_format (MonoError *error, const char *argument, const char *msg_format, ...) MONO_ATTR_FORMAT_PRINTF(3,4);
 
 void
-mono_error_set_argument (MonoError *error, const char *argument, const char *msg, ...) MONO_ATTR_FORMAT_PRINTF(3,4);
+mono_error_set_argument (MonoError *error, const char *argument, const char *msg, ...);
```
lambdageek: It shoudln't be variadic, either.  
```suggestion
mono_error_set_argument (MonoError *error, const char *argument, const char *msg);
```
Also need to update the definition to remove `...`
https://github.com/mono/mono/blob/bf7faa00ad7590ee7aebf3e06a43b1439bdaac64/mono/utils/mono-error.c#L546

directhex:
@monojenkins build failed

Issue #13006, opened 2019-02-14
[2018-08] [watchOS/LLVM] Test crash 


rolfbjarne:
## Steps to Reproduce

1. Clone & build the 2018-10 integration branch: https://github.com/xamarin/xamarin-macios/pull/4904
2. Run xharness, and run the `dont link / watchOS - device / Release` test.

### Current Behavior

Test crash:

```
Thread 0 name:  tid_303  Dispatch queue: com.apple.main-thread
Thread 0 Crashed:
0   dont link                     	0x003793b0 dont_link_DontLink_DontLinkRegressionTests_ProcessStart_NotSupported + 1250
1   dont link                     	0x009b07fa mscorlib_wrapper_runtime_invoke_object_runtime_invoke_sig_void_intptr_intptr_object_intptr_intptr_intptr + 92
2   dont link                     	0x009cc424 mscorlib_wrapper_unknown_object_gsharedvt_out_sig_intptr_0 + 0
```

Full crash report: https://gist.github.com/rolfbjarne/e93b9f0db3e03b358516a7858ec717a7

Full test run output: https://gist.github.com/rolfbjarne/9ce418682da72f22beedc9734ce11ecd

### Expected Behavior

No crash (the test is supposed to catch a PlatformNotSupportedException, so my assumption is that something went wrong either when throwing the exception, or when catching it).

## On which platforms did you notice this

[x] watchOS
[ ] Linux
[ ] Windows

**Version Used**:

https://github.com/xamarin/xamarin-macios/pull/4904

lewurm:
I'm on it, but I've troubles reproducing it:
https://gitter.im/xamarin/xamarin-macios?at=5c657aae604f233ab60bee3a

rolfbjarne:
Interestingly I can't reproduce this using arm64_32 (from https://github.com/xamarin/xamarin-macios/pull/5465), but it reproduces using armv7k from that same PR (which means this is not something new in 2018-10, since that branch is based on 2018-08).

I wonder if this is yet another example of the app being too big for armv7k, causing random behavior (although in this case the problem is 100% reproducible for me on armv7k).

vargaz:
I'd still suggest removing dontlink on watchos.

rolfbjarne:
> I wonder if this is yet another example of the app being too big for armv7k

Actually it's not. I enabled the linker for the dontlink project, and it still crashes (in a different place though). Crash report is quite unhelpful:

```
[...]

PR #14397, opened 2019-05-07
[w32mutex] use handle stack in ReleaseMutex

lewurm:
Fixes this on watchOS with llvmonly:
```
MutexTest
warning: Handle 0x1624bc0c (object = 0x2ce4120) (allocated from "../../../../../mono/metadata/threads.c:2022") is leaking.

* thread %1
  * frame %0: 0x0183a3e0 mscorlibtests`mono_pmip(ip=0x01a2e191) at mini-runtime.c:243
    frame %1: 0x0189f3a6 mscorlibtests`mono_handle_new(obj=0x01e505f8, info=0x1624e600, owner="../../../../../mono/metadata/gc.c:1337") at handle.c:186:2
    frame %2: 0x0189e398 mscorlibtests`mono_gc_alloc_handle_string(vtable=0x16a52760, size=50, len=18) at gc.c:1337:9
    frame %3: 0x018fa91e mscorlibtests`mono_string_new_size_handle(domain=0x15e14e60, len=18, error=0x01ce4180) at object.c:6658:6
    frame %4: 0x018f304c mscorlibtests`mono_string_new_size_checked(domain=0x15e14e60, length=18, error=0x01ce4180) at object.c:6670:2
    frame %5: 0x018fa6d6 mscorlibtests`mono_string_new_utf16_checked(domain=0x15e14e60, text=0x02c5980d, len=18, error=0x01ce4180) at object.c:6551:6
    frame %6: 0x018fa726 mscorlibtests`mono_string_new_utf16_handle(domain=0x15e14e60, text=0x02c5980d, len=18, error=0x01ce4180) at object.c:6569:9
    frame %7: 0x018f651e mscorlibtests`mono_ldstr_metadata_sig(domain=0x15e14e60, sig="M", error=0x01ce4180) at object.c:7525:23
    frame %8: 0x018fc0f0 mscorlibtests`mono_ldstr_checked(domain=0x15e14e60, image=0x16a4c400, idx=70660, error=0x01ce4180) at object.c:7477:8
    frame %9: 0x018224ce mscorlibtests`mono_helper_ldstr_mscorlib(idx=70660) at jit-icalls.c:1172:23
    frame %10: 0x00925efe mscorlibtests`aot_wrapper_icall_mono_helper_ldstr_mscorlib + 68
    frame %11: 0x0042a60e mscorlibtests`mscorlib_System_Threading_Mutex_ReleaseMutex + 194
    frame %12: 0x01232f36 mscorlibtests`mscorlibtests1_MonoTests_System_Threading_MutexTest_DoubleRelease + 394
    frame %13: 0x009456ca mscorlibtests`aot_wrapper_gsharedvt_out_sig_pinvoke_void_this_ + 28
    frame %14: 0x0091728a mscorlibtests`mscorlib_wrapper_runtime_invoke_object_runtime_invoke_sig_void_intptr_intptr_object_intptr_intptr_intptr + 312
    frame %15: 0x018438ca mscorlibtests`mono_llvmonly_runtime_invoke(method=0x1632ecd8, info=0x17b94ec0, obj=0x01e4fe70, params=0x00000000, exc=0x01ce4550, error=0x01ce48a4) at mini-runtime.c:3000:2
```
Contributes to https://github.com/mono/mono/issues/13006

### DIFF
--- a/mono/metadata/w32mutex-unix.c
+++ b/mono/metadata/w32mutex-unix.c
@@ -385,14 +385,16 @@ ves_icall_System_Threading_Mutex_ReleaseMutex_internal (gpointer handle)
 	if (!mono_w32handle_lookup_and_ref (handle, &handle_data)) {
 		g_warning ("%s: unkown handle %p", __func__, handle);
 		mono_w32error_set_last (ERROR_INVALID_HANDLE);
-		return FALSE;
+		ret = FALSE;
+		goto exit;
 	}
 
 	if (handle_data->type != MONO_W32TYPE_MUTEX && handle_data->type != MONO_W32TYPE_NAMEDMUTEX) {
 		g_warning ("%s: unknown mutex handle %p", __func__, handle);
 		mono_w32error_set_last (ERROR_INVALID_HANDLE);
 		mono_w32handle_unref (handle_data);
-		return FALSE;
+		ret = FALSE;
+		goto exit;
 	}
 
 	mutex_handle = (MonoW32HandleMutex*) handle_data->specific;
@@ -432,6 +434,7 @@ ves_icall_System_Threading_Mutex_ReleaseMutex_internal (gpointer handle)
 	mono_w32handle_unlock (handle_data);
 	mono_w32handle_unref (handle_data);
 
+exit:
 	HANDLE_FUNCTION_RETURN_VAL (ret);
 }
 
--- a/mono/metadata/w32mutex-unix.c
+++ b/mono/metadata/w32mutex-unix.c
@@ -380,6 +380,8 @@ ves_icall_System_Threading_Mutex_ReleaseMutex_internal (gpointer handle)
 	pthread_t tid;
 	gboolean ret;
 
+	HANDLE_FUNCTION_ENTER ();
+
 	if (!mono_w32handle_lookup_and_ref (handle, &handle_data)) {
 		g_warning ("%s: unkown handle %p", __func__, handle);
 		mono_w32error_set_last (ERROR_INVALID_HANDLE);
@@ -430,7 +432,7 @@ ves_icall_System_Threading_Mutex_ReleaseMutex_internal (gpointer handle)
 	mono_w32handle_unlock (handle_data);
 	mono_w32handle_unref (handle_data);
 
-	return ret;
+	HANDLE_FUNCTION_RETURN_VAL (ret);
 }
 
 gpointer


mono/metadata/w32mutex-unix.c
```diff
@@ -380,6 +380,8 @@ ves_icall_System_Threading_Mutex_ReleaseMutex_internal (gpointer handle)
 	pthread_t tid;
 	gboolean ret;
 
+	HANDLE_FUNCTION_ENTER ();
```
lambdageek: There are a couple of places inside `ReleaseMutex_internal` where it exits early using `return`.  Those need to change - wrapping a function in `HANDLE_FUNCTION_` means that it has to be single-exit.

lewurm: thanks, updated!  Let me know if we have proper macros for it, I'm a bit lost in that world 🙂

jaykrell:
I also don’t understand why these are platform-specific problems.

jaykrell:
ReleaseMutex should also not do an allocation, of memory, of strings, of handles.

lewurm:
`mono_thread_internal_current_handle` does the allocation.  not sure if it's supposed to do that.

lewurm:
@monojenkins build failed

jaykrell:
My use of the word "should" confuses people, sorry. "ought". "my preference", "probably not the existing code, maybe 'someone' can 'fix' it".

The `mono_thread_internal_current_handle` thing is annoying, stuck between efficiency, consistency of use, long standing implementation details etc. The internal thread is pinned. A handle to it is overkill. This arguably is an implementation detail, subject to change.

In fact the internal thread I think goes away under ifdef, and maybe its replacement is not in fact pinned?
I have to check that.

You know, releasing a lock, is not something any code is prepared to deal with failure from. Not even code that survives low memory. Perhaps some day Mono could run under low memory.

Acquiring lock arguably is fallable, and should return whatever is needed to release it, i.e. the handle. But then the interface is clumsy, what is this weird implementation detail (internal thread handle) leaking through into AcquireMutex...

But maybe wrong GitHub PR? This one is a string. Let me see..

jaykrell:
Oh it is the error path.

```
		[ReliabilityContractAttribute (Consistency.WillNotCorruptState, Cer.MayFail)]
		public void ReleaseMutex() {
			bool success = ReleaseMutex_internal(Handle);
			if (!success) {
				throw new ApplicationException ("Mutex is not owned");
			}
		}
```
I wonder if that can even be implemented reliably cross-platform, and if it should be failfast w/o raising an exception.
Releasing a lock is not really something code can be prepared to fail.

jaykrell:
Not just the error path. But the internal thread handle as stated. Looking more..

jaykrell:
Looks like both parts of threads are pinned:

```
static MonoThread*
create_thread_object (MonoDomain *domain, MonoInternalThread *internal)
{
#ifdef ENABLE_NETCORE
	MONO_OBJECT_SETREF_INTERNAL (internal, internal_thread, internal);
	return internal;
#else
	MonoThread *thread;
	MonoVTable *vtable;
	ERROR_DECL (error);

	vtable = mono_class_vtable_checked (domain, mono_defaults.thread_class, error);
	mono_error_assert_ok (error);

	thread = (MonoThread*)mono_object_new_mature (vtable, error);
	/* only possible failure mode is OOM, from which we don't expect to recover. */
	mono_error_assert_ok (error);

	MONO_OBJECT_SETREF_INTERNAL (thread, internal_thread, internal);

	return thread;
#endif
}
```

I'll PR taking advantage of this. There are arguments either way.

jaykrell:
```

static MonoInternalThread*
create_internal_thread_object (void)
{
	ERROR_DECL (error);
	MonoInternalThread *thread;
	MonoVTable *vt;

	vt = mono_class_vtable_checked (mono_get_root_domain (), mono_defaults.internal_thread_class, error);
	mono_error_assert_ok (error);
	thread = (MonoInternalThread*) mono_object_new_mature (vt, error);
	/* only possible failure mode is OOM, from which we don't exect to recover */
	mono_error_assert_ok (error);

	init_internal_thread_object (thread);

	return thread;
}
```

lewurm:
Let's go with https://github.com/mono/mono/pull/14403

PR #14403, opened 2019-05-08
Do not allocate handles in ReleaseMutex.

jaykrell:
Take advantage of thread/internal_thread maturity/pinning.

There are arguments for and against this.

It is arguably better to use handles everywhere.
But it is dubious to require an allocation to release a lock.
Releasing a lock should be infallible.
And surviving low memory shouldn't be a lost cause, imho..

This diff could also be viewed as an opportunistic cycle/size optimization
and be placed under ifdef -- ifdef THREAD_ALWAYS_PINNED orsuch.

This is an alternative to https://github.com/mono/mono/pull/14397.

Contributes to https://github.com/mono/mono/issues/13006

### DIFF
--- a/mono/metadata/w32mutex-unix.c
+++ b/mono/metadata/w32mutex-unix.c
@@ -37,27 +37,31 @@ static gpointer
 mono_w32mutex_open (const char* utf8_name, gint32 rights G_GNUC_UNUSED, gint32 *win32error);
 
 static void
-thread_own_mutex (MonoInternalThreadHandle internal, gpointer handle, MonoW32Handle *handle_data)
+thread_own_mutex (MonoInternalThread *internal, gpointer handle, MonoW32Handle *handle_data)
+// Thread and InternalThread are pinned/mature.
+// Take advantage of that and do not use handles here.
 {
 	/* if we are not on the current thread, there is a
 	 * race condition when allocating internal->owned_mutexes */
-	g_assert (mono_thread_internal_is_current_handle (internal));
+	g_assert (mono_thread_internal_is_current (internal));
 
-	if (!MONO_HANDLE_GETVAL (internal, owned_mutexes))
-		MONO_HANDLE_SETVAL(internal, owned_mutexes, GPtrArray*, g_ptr_array_new ());
+	if (!internal->owned_mutexes)
+		internal->owned_mutexes = g_ptr_array_new ();
 
-	g_ptr_array_add (MONO_HANDLE_GETVAL (internal, owned_mutexes), mono_w32handle_duplicate (handle_data));
+	g_ptr_array_add (internal->owned_mutexes, mono_w32handle_duplicate (handle_data));
 }
 
 static void
-thread_disown_mutex (MonoInternalThreadHandle internal, gpointer handle)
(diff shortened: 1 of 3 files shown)


jaykrell:
i.e. largely undoing my https://github.com/mono/mono/commit/5614bdddce19016dc875c6f8cd004ac02df18f65

jaykrell:
Completely undoing it actually.

lewurm:
Thanks Jay.  Tested it on the watch, and works.  I'll let @lambdageek review it.

jaykrell:
It is debatable either way. I'm on the fence.
There is a certain purity and future-proofing to the old change and your fix.
Thread/InternalThread could be unpinned and they'd work. My alternative depends on that not happening.

@lambdageek has espoused the purity in the past and I have gone along with it.
threads.c often does a "shallow" coop conversion and comments as such, depending on this pinning, and arguably it was a temporary shortcut. Arguably.

jaykrell:
@monojenkins build failed

mono/metadata/w32mutex-unix.c
```diff
@@ -37,27 +37,31 @@ static gpointer
 mono_w32mutex_open (const char* utf8_name, gint32 rights G_GNUC_UNUSED, gint32 *win32error);
 
 static void
-thread_own_mutex (MonoInternalThreadHandle internal, gpointer handle, MonoW32Handle *handle_data)
+thread_own_mutex (MonoInternalThread *internal, gpointer handle, MonoW32Handle *handle_data)
+// Thread and InternalThread are pinned/mature.
```
lambdageek: don't put comments between the parameters and the `{`

We usually put a `/** apidoc */` comment before the declaration and mention `\p internal` or `\c MonoInternalThread`.

mono/metadata/w32mutex-unix.c
```diff
@@ -37,27 +37,31 @@ static gpointer
...
+		internal->owned_mutexes = g_ptr_array_new ();
 
-	g_ptr_array_add (MONO_HANDLE_GETVAL (internal, owned_mutexes), mono_w32handle_duplicate (handle_data));
+	g_ptr_array_add (internal->owned_mutexes, mono_w32handle_duplicate (handle_data));
 }
 
 static void
-thread_disown_mutex (MonoInternalThreadHandle internal, gpointer handle)
+thread_disown_mutex (MonoInternalThread *internal, gpointer handle)
+// Thread and InternalThread are pinned/mature.
```
lambdageek: no comments between parameters and `{`

mono/metadata/w32mutex-unix.c
```diff
@@ -476,24 +474,22 @@ mono_w32mutex_open (const char* utf8_name, gint32 rights G_GNUC_UNUSED, gint32 *
 }
 
 void
-mono_w32mutex_abandon (MonoInternalThread *internal_raw)
+mono_w32mutex_abandon (MonoInternalThread *internal)
+// Thread and InternalThread are pinned/mature.
```
lambdageek: no comments between parameter list and `{'`

lewurm:
@monojenkins backport 2019-02
@monojenkins backport 2019-04

PR #14408, opened 2019-05-08
[gitignore] Add crash dump files, .dylib.dSYM directories, and test-path

CoffeeFlux:
Wasn't sure where in the file to best put the .dylib.dSYM stuff, so suggestions are welcome


lewurm:
@monojenkins squash

PR #14415, opened 2019-05-08
[llvm] Implement direct cross-amodule calls

alexanderkyte:
After much debugging, I've got an initial prototype for direct cross-aot-module calls. 

## Big Picture
Right now, every AOT'ed method call from one assembly to another goes through a layer of indirection.

This indirection leads to a number of performance issues. The need to lazily initialize references is a drain on latency and startup time, and repeated indirection calls introduces a need to keep many, many function addresses kept in L1/L2 cache in order to not continually stall on the access to the function pointer.

This PR tries to introduce direct calls. Doing so is very simple, we only need to change the linkage of a symbol. The hard parts come from figuring out which methods cannot be called directly, and failing gracefully.

## Linking Options
Since we're emitting direct OS-visible calls between AOT modules, we have some flexibility with how the OS links them together. We will want to use this flexibility in some places. 

On the desktop (dynamic linking) AOT scenario, I am using clang's ability to resolve undefined references at runtime. The external symbol name referenced by the native binary made by the AOT compile will be undefined. When we call it, the OS will look for the actual symbol and will patch it into process such that the symbol references it. I'm sure that this could be faster, but it seems pretty quick.

In the static scenario, all of these direct calls can simply be linked away into references to the actual function address.

## Tech Debt
Right now we're not using this infrastructure as much as can be used.

This disables a lot of possible calls for the reason that they are `beforefieldinit`. When allowed to directly call, there's a ton of crashes due to what seems like clearly calling the wrong callee or passing the wrong args. 

There's good statistic tracking to watch and fix these in later PRs though.

## Stats 
The following table is emitted with `--aot=stats` :

```

Call indirection stats:
    Direct Calls Made: 4
    Indirection Cause: Wrappers Count: 317
    Indirection Cause: Static Ctor Count: 149
    Indirection Cause: BeginInvoke Count: 2
    Indirection Cause: Duplicated Count: 6871
    Indirection Cause: BeforeFieldInit Count: 12734
    Indirection Cause: PrivateImpl Count: 2

```

## Caching

The idea behind the caching is to amortize the compilation. We have inescapable wasted work if we want to preserve parallel AOTing of all assemblies in the assembly reference graph.

It's important to be able to shadow-compile (see if we'd fail to compile by doing IR transformation steps without emitting code) an assembly to be able to see what we fail to compile. Otherwise, we force churn on the embedder that is everything from AOT dependency ordering to concurrency-vs-wasted-work tradeoffs. These are easier to handle on our end, from within the runtime. AOT interface churn can be very painful, as it involves having to set up dependencies between the emitted AOT files that mirror the dependencies between the source Assemblies. 

We therefore shadow-compile Assemblies when they're referenced *and* when they're normally compiled. 

This means that if System's nunit test assembly has a reference to mscorlib.dll, and mscorlib.dll has been AOT'ed before System, we will use the cached failure file. 

It also means that if two test assemblies reference the same mscorlib.dll, the AOT of both of them won't have to duplicate the step of checking which compilation failures will happen in mscorlib.dll

The caching amortizes the shadow compilation, providing theoretic running-time bounds that are much better than the combinatorial explosion of every AOT step having to re-AOT every method in every assembly that is referenced.

## Testing

Build any AOT profile with LLVM and you should see this enabled. CI here will test this. This is disabled on Linux in the non static linking scenario, but Wasm will enable it and OSX will run with it by default (using the clang runtime dynamic loading mechanism).

The idea is to have it enabled by default on the desktop, static scenarios, wasm, etc. This is tested by the BCL tests, and minor mistakes lead to easy-to-spot failures (missing symbol lookups or calling the wrong callee and crashing after stack corruption).

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -13380,7 +13380,7 @@ mono_setup_direct_external_call_state (MonoAotCompile *acfg, MonoAotState **glob
 	gpointer value;
 	g_hash_table_iter_init (&iter, (*astate)->exported_method_failures);
 	while (g_hash_table_iter_next (&iter, &key, &value))
-		g_hash_table_insert (acfg->imported_method_failures, g_strdup (key), value);
+		g_hash_table_insert (acfg->imported_method_failures, g_strdup ((char *) key), value);
 
 	// Clear out exports table for a fresh iteration
 	GHashTable *old = (*astate)->exported_method_failures;
@@ -13546,7 +13546,7 @@ mono_compile_assemblies (MonoDomain *domain, char **argv, int argc, guint32 opts
 			continue;
 		}
 
-		int res = mono_compile_assembly (assem, opts, aot_options, (gpointer) &aot_state);
+		int res = mono_compile_assembly (assem, opts, aot_options, (gpointer **) &aot_state);
 
 		if (!aot_opts.disable_direct_external_calls)
 			mono_write_callee_failures (aot_state->exported_method_failures, assem->image);
@@ -13569,7 +13569,7 @@ mono_compile_assemblies (MonoDomain *domain, char **argv, int argc, guint32 opts
 	while (g_hash_table_iter_next (&iter, (gpointer *) &assem, NULL)) {
 		g_assert (!aot_state->collecting_callee_failures);
 
-		int res = mono_compile_assembly (assem, opts, aot_options, (gpointer) &aot_state);
+		int res = mono_compile_assembly (assem, opts, aot_options, (gpointer **) &aot_state);
 
 		if (!aot_opts.disable_direct_external_calls)
 			mono_write_callee_failures (aot_state->exported_method_failures, assem->image);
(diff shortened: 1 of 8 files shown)


alexanderkyte:
When actually logged at a methhod-by-method level and put through the test of dealing with large apps, this PR seems to run into a pretty familiar issue. 

We'd like to have a deny-list rather than an allow-list. The deny-list is much smaller in the worst case. This above incarnation of this uses that policy. 

That doesn't actually work. There are a myriad of edge cases that can cause us to not see a method at compile-time that we think we expect to have seen when compiling the embedder. This implies that an allow-list would be better. There is a branch here (https://github.com/alexanderkyte/mono/tree/direct_cross_debug) which makes that change. 

In practice, it seems that far fewer methods are safe to call externally on most platforms. In many cases, we only allow ten to twelve instances out of thousands. Some of those are definitely callable, but the route forward for this PR involves a rather time-consuming exploration of the direct calls which break versus ones that currently work out of the box. 

Bootstrapping this in a less "all or nothing" correctness way is difficult. Being able to enable the methods call-by-call is doable and helpful for figuring out which calls are safe (and currently done in the linked development branch) but right now makes a rather poor use of time due to the latency of the AOT build. It's difficult to work on while multitasking as well. For that reason it was put down in order to tackle more pressing things.

PR #14418, opened 2019-05-09
[netcore] Enable MSVC builds of Mono w/ ENABLE_NETCORE

filipnavara:
/cc @directhex @lateralusX

### DIFF
--- a/mono/mini/main-core.c
+++ b/mono/mini/main-core.c
@@ -20,7 +20,7 @@
 #pragma comment(linker, "/export:coreclr_shutdown_2=_coreclr_shutdown_2@12")
 #pragma comment(linker, "/export:coreclr_create_delegate=_coreclr_create_delegate@24")
 #undef MONO_API
-#define MONO_API
+#define MONO_API MONO_EXTERN_C
 #endif
 
 MONO_API int STDAPICALLTYPE coreclr_initialize (const char* exePath, const char* appDomainFriendlyName,
--- a/mono/mini/main-core.c
+++ b/mono/mini/main-core.c
@@ -10,7 +10,7 @@
 #define STDAPICALLTYPE
 #endif
 
-#ifdef _MSC_VER && TARGET_WIN32
+#if defined(_MSC_VER) && defined(HOST_WIN32) && defined(HOST_X86)
 // Ensure that the exported symbols are not decorated and that only one set is exported
 #pragma comment(linker, "/export:coreclr_initialize=_coreclr_initialize@28")
 #pragma comment(linker, "/export:coreclr_execute_assembly=_coreclr_execute_assembly@24")
--- a/mono/mini/main-core.c
+++ b/mono/mini/main-core.c
@@ -10,6 +10,16 @@
 #define STDAPICALLTYPE
 #endif
 
+#ifdef _MSC_VER && TARGET_WIN32
+// Ensure that the exported symbols are not decorated and that only one set is exported
+#pragma comment(linker, "/export:coreclr_initialize=_coreclr_initialize@28")
+#pragma comment(linker, "/export:coreclr_execute_assembly=_coreclr_execute_assembly@24")
+#pragma comment(linker, "/export:coreclr_shutdown_2=_coreclr_shutdown_2@12")
+#pragma comment(linker, "/export:coreclr_create_delegate=_coreclr_create_delegate@24")
+#undef MONO_API
+#define MONO_API
+#endif
+
 MONO_API int STDAPICALLTYPE coreclr_initialize (const char* exePath, const char* appDomainFriendlyName,
 	int propertyCount, const char** propertyKeys, const char** propertyValues,
 	void** hostHandle, unsigned int* domainId);
--- a/mono/mini/main-core.c
+++ b/mono/mini/main-core.c
@@ -1,4 +1,7 @@
 #include <config.h>
+
+#if ENABLE_NETCORE
+
 #include "mini.h"
 #include "mini-runtime.h"
 #include <mono/metadata/assembly.h>
@@ -321,3 +324,5 @@ int STDAPICALLTYPE coreclr_create_delegate (void* hostHandle, unsigned int domai
 	g_error ("Not implemented");
 	return 0;
 }
+
+#endif // ENABLE_NETCORE
--- a/msvc/libmini.targets
+++ b/msvc/libmini.targets
@@ -6,4 +6,5 @@
   <Import Project="$(MSBuildThisFileDirectory)libmini-arch.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmini-interp.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmini-llvm.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)libmini-netcore.targets" />
 </Project>
--- a/msvc/build-init.vcxproj
+++ b/msvc/build-init.vcxproj
@@ -157,6 +157,7 @@
   <Target Name="AfterBuildWinSetup">
     <PropertyGroup>
       <_EnableDefines Condition="'$(MONO_ENABLE_LLVM)' == 'true'">$(_EnableDefines);ENABLE_LLVM;ENABLE_LLVM_RUNTIME</_EnableDefines>
+      <_EnableDefines Condition="'$(MONO_ENABLE_NETCORE)' == 'true'">$(_EnableDefines);ENABLE_NETCORE</_EnableDefines>
       <_HaveDefines Condition="'$(MONO_ENABLE_BTLS)' == 'true'">$(_HaveDefines);HAVE_BTLS</_HaveDefines>
       <_EnableDefines>$(_EnableDefines.Trim(';'))</_EnableDefines>
       <_HaveDefines>$(_HaveDefines.Trim(';'))</_HaveDefines>
--- a/msvc/libmini-netcore.targets
+++ b/msvc/libmini-netcore.targets
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="netcore_sources">
+    <ClCompile Include="$(MonoSourceLocation)\mono\mini\main-core.c" />
+  </ItemGroup>
+</Project>
--- a/msvc/libmini-netcore.targets
+++ b/msvc/libmini-netcore.targets
@@ -1,13 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
-    <ExcludeNetCoreFromWindowsBuild Condition="'$(MONO_ENABLE_NETCORE)'=='true'">false</ExcludeNetCoreFromWindowsBuild>
-    <ExcludeNetCoreFromWindowsBuild Condition="'$(MONO_ENABLE_NETCORE)'=='' Or '$(MONO_ENABLE_NETCORE)'!='true'">true</ExcludeNetCoreFromWindowsBuild>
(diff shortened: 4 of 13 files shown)


lateralusX:
More comments in a while, but first thing, we moved away from def files, and since the methods are public use Mono_API instead to export them when build as part of DLL.

jaykrell:
.def files are preferred imho.

The only wrinkle is that you need to transform (or maintain) the list into what each linker requires.

As long as the imports/exports are all functions, not data, you can remove the decorations, or make them be cosmetic only, and building static .lib and .dll is only one compilation.

Also consumers do not need declspec(dllimport) on functions, unless they are comparing/hashing function pointers.

declspec(dllexport) is imho really just for C++ mangled exports, which are to be avoided anyway.
Nobody wants to maintain them since they are OS and CPU specific, but some people do.

filipnavara:
> we moved away from def files, and since the methods are public use Mono_API instead to export them when build as part of DLL.

They are MONO_API already. The problem is that they are STDCALL and need to be exported without the decoration (`_` prefix and `@<param size>` suffix). There's no other good method to achieve that as far as I know (short of `/EXPORT` parameters for linker or `#pragma`s for linker in the actual source code).

jaykrell:
That is exactly one of the reasons .def files are favored.

Look at kernel32.dll's ability to serve LoadLibrary/GetProcAddress without the x86 __stdcall stuff but instead just portable names. .def files do that.

.def files also let you export from static .libraries without worry about this problem as much:
https://github.com/mono/mono/blob/c7697e06438ebb66d74e2ec20dd90d08773b2f4b/mono/metadata/object.c#L62

declspec(dllexport) from static .libs only kinda works -- you have get the relevant .objs referenced somehow. .def files force the reference.

jaykrell:
That last point is portable to gcc/ld also -- exported in unused .o files within .a are not exported.
You need something to more strongly get them exported, which that include is working around, or .def files, or presumably the gcc/ld equivalent.

msvc/libmini.targets
```diff
@@ -6,4 +6,5 @@
   <Import Project="$(MSBuildThisFileDirectory)libmini-arch.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmini-interp.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmini-llvm.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)libmini-netcore.targets" />
```
filipnavara: I forgot to make this conditional...somehow. What is the preferred way?

lateralusX:
Is the net core define also in config.h after running configure step? If so I recommend that you do something similar to the LLVM flag, keep the property in mono.props but don't set define on the projects, instead look in build-init.vcxproj, you also add the define in the white list of supported defines in mono.winconfig.targets, this way you can either set the property in msbuild, using IDE ro pick it up from Mono configuration step.

lateralusX:
Regarding def file, so the switch was made to make it easier for non Windows devs to export (since the missmatch regulary caused errors and diffs in exposed API surface) making is similar to other platforms, just using MONO_API to get a method exported on all platforms.

filipnavara:
> Is the net core define also in config.h after running configure step?

Yes

> If so I recommend that you do something similar to the LLVM flag, keep the property in mono.props but don't set define on the projects, instead look in build-init.vcxproj, you also add the define in the white list of supported defines in mono.winconfig.targets, this way you can either set the property in msbuild, using IDE or pick it up from Mono configuration step.

Ok, will do. I already have msbuild and IDE working. I'll try the ./configure path.

> Regarding def file, so the switch was made to make it easier for non Windows devs to export (since the missmatch regulary caused errors and diffs in exposed API surface) making is similar to other platforms, just using MONO_API to get a method exported on all platforms.

It already works on the MinGW builds correctly, both for native compilation and cross-compilation. The .def file is there only for the MSVC to drop the stdcall decorations.

lateralusX:
OK, I understand the requirement and it will only be 32-bit only and it is also possible to use both .def file and declspec export, so it should work in parallell with existing export mechanism, maybe just have the def file on the 32-bit build configurations?

filipnavara:
> OK, I understand the requirement and it will only be 32-bit only it is also possible to use both .def file and declspec export, so it should work in parallell with existing export mechanism, maybe just have the def file on the 32-bit build configurations?

Yep. I am trying the `#pragma` approach as an alternative to keep it close to the source code and avoid the .def file.

filipnavara:
So it's either the .def file or something like this in `main-core.c`:

```c
#ifdef _MSC_VER && TARGET_WIN32
#pragma comment(linker, "/export:coreclr_initialize=_coreclr_initialize@28")
#pragma comment(linker, "/export:coreclr_execute_assembly=_coreclr_execute_assembly@24")
#pragma comment(linker, "/export:coreclr_shutdown_2=_coreclr_shutdown_2@12")
#pragma comment(linker, "/export:coreclr_create_delegate=_coreclr_create_delegate@24")
#undef MONO_API
#define MONO_API
#endif
```

Any preferences?

lateralusX:
Since we are using MONO_API's and no def files and its 32-bit msvc only, I like the pragma approach.

lateralusX:
But you should use HOST_X86 instead TARGET_WIN32 is not 32-bit windows, its used in places where we generated code for WIN32 target, or like this:

#if HOST_WIN32 && HOST_X86

meaning that the thing we build will run on a windows 32-bit host.

filipnavara:
I am still not sure how the `./configure --with-core=only` path is supposed to work. It will generate `config.h` with `ENABLE_NETCORE` defined but I don't see how would the MSVC build get executed with `-p:MONO_ENABLE_NETCORE=true` (and I couldn't trace a similar LLVM code).

lateralusX:
So its picked up when you build mono.sln, it will run build-init.vcxproj that will, take a copy of config.h from configure into cygconfig.h and then setup the config.h that is compatible with msvc. The process doing that is taking the template winconfig.h and then apply a set of defines picked up from cygconfig.h or explicitly specified.

lateralusX:
So if you do configure and then just build mono.sln (no need for the property) you should get an msvc config.h that includes your define.

lateralusX:
So end result should be, if you don't use config or would like to force just when doing msvc build, use property in msbuild or IDE. If you just do config and build, it will be used if included in config step.

filipnavara:
Ah, found it. The missing bit in the logic is in `mono.external.targets`.

lateralusX:
Did you really need it in mono.external.targets? All defines white listed in mono.winconfig.targets and found in cygconfig.h should be included in config.h, didn't that happen? Don't say it can't be done this way but we pick up other defines from config step just using the white list. If this is just to change to include the file to build when using configure step, maybe we should apply a pattern used elsewhere in Mono, use define in source file to include/exclude depending on defines in config.h.

filipnavara:
> If this is just to change to include the file to build when using configure step, maybe we should apply a pattern used elsewhere in Mono, use define in source file to include/exclude depending on defines in config.h.

It is exactly because of that. I anticipate that the set of compiled files for netcore and non-netcore builds may differ more in the future. I'm not sure how far you can scale the `#ifdef` approach before it becomes unreadable.

lateralusX:
OK, the #ifdef approach is quite heavily used in Mono source files to include/exclude "features". I'm fine either way, but doing the #ifdef for this could remove the need for the external + additional target in vscxproj's for now. If we need to scale this in future and it becomes a problem, then maybe we could revisit.

filipnavara:
I am fine with it either way. I mostly tried to mimic the existing non-MSVC setup:

https://github.com/mono/mono/blob/f48b0a70c3ae4418775594200e29856dd2dd7668/mono/mini/Makefile.am.in#L428-L433

It was introduced in #12416 by @marek-safar.

lateralusX:
If we could do the #ifdef for configure builds to simplify this PR and I can (when time) have this in mind to setup a generic approach, sounds ok?

directhex:
@monojenkins build failed

mono/mini/main-core.c
```diff
@@ -10,6 +13,16 @@
...
 #endif
 
+#if defined(_MSC_VER) && defined(HOST_WIN32) && defined(HOST_X86)
+// Ensure that the exported symbols are not decorated and that only one set is exported
+#pragma comment(linker, "/export:coreclr_initialize=_coreclr_initialize@28")
+#pragma comment(linker, "/export:coreclr_execute_assembly=_coreclr_execute_assembly@24")
+#pragma comment(linker, "/export:coreclr_shutdown_2=_coreclr_shutdown_2@12")
+#pragma comment(linker, "/export:coreclr_create_delegate=_coreclr_create_delegate@24")
+#undef MONO_API
+#define MONO_API
```
jaykrell: They might still need to be extern "C"?

filipnavara: Ah, right.

directhex:
@monojenkins build failed

jaykrell:
Pragma has the same problem with exports from static libraries. GNU ld -export-list not almost like .def and one file could feed both?

filipnavara:
> Pragma has the same problem with exports from static libraries. GNU ld -export-list not almost like .def and one file could feed both?

In this particular case only the dynamic library is a concern. There's no use for static library in this configuration (yet). The approach with .def actually still exported all the symbols, just fixed the decoration for these STDCALL ones, so if necessary we can revive it later.

msvc/libmini.targets
kg: if you didn't know, the convention is to put them in an itemgroup with a Condition= on it

filipnavara: That's what I would normally do. However, in this context it is problematic. The variable you'd test in the condition could be set from three different places: a) MSBuild command line b) mono.props file (and VS IDE by proxy) c) by reverse lookup from config.h generated by `./configure` script. The first two are easy to handle, but the last one is late-bound and currently obtained in very complex way.

kg: Can you define late-bound for me? I'd like to help you solve this in a clear and understandable way.

jaykrell:
The problem isn't mono as a static library, it is mono dynamic library built out of a set of static libraries.

Imagine you static monometadata.lib and static monomini.lib and static monoutil.lib and then link them to produce monoruntime.dll.

The pragmas can work, but not in a particularly predictable way.

Just __declspec(dllexport) can work, but not in a particularly predictable way.

If an exported function is not used by mono itself, and is in a source file that contains only functions not actually used by mono, but are exported for others to use, then they won't be exported with the pragma 

I'm pretty sure, with declspec for sure, but a .def file will export them, as it counts as a reference from the linker's point of view.

Smaller projects don't run into this, because they don't split themselves up into static libs.

Larger projects do.

Mono is actually in this exact form at least on non-Windows platforms.

I worked out it with that include, which is somewhat by-chance, somewhat reliable.
I include into a file "known" to have used-by-mono symbols.

msvc/libmini.targets
filipnavara: It's an output of a task parameter. I replicated the whole logic used for LLVM in https://github.com/mono/mono/pull/14418/commits/bab126ce29b01ba72b70dca5c375cade523ad964, but it is cumbersome and in the end I reverted it at @lateralusX's request.

I do know ways how to solve it more elegantly but they are way out of scope for this change where I just wanted to get the basic build working. I didn't understand why a simple `Condition="'$(MONO_ENABLE_NETCORE)' == 'true'"` would not work but after figuring out the whole logic in `mono.external.targets` I came to the conclusion that trying to rework it at this point would result in too big change.

directhex:
@monojenkins build failed

filipnavara:
@jaykrell In this case the problem is specific to a single dynamic library, MSVC, x86-only and ENABLE_NETCORE build, all at once. In any other case it uses the usual `MONO_API` code (ie. `__declspec(dllexport)` or equivalent) or the file is not used at all. STDCALL and its decoration is Win32 specific, it doesn't exist on any other platform, including Win64.

The pragmas merely work to produce the proper aliases in one of the dynamic libraries that is consumed by external code. Redefining `MONO_API` is done only to avoid exporting the symbols twice.

I am not big fan of it, but the only other viable alternative is the .def file which turns out to be a horror to get done right for all the possible configurations (or, more specifically, for the `./configure` + MSVC builds).

PR #14420, opened 2019-05-09
[packaging] Add some missing .pdb's to Windows MSI

akoeplinger:


akoeplinger:
@monojenkins build msi

akoeplinger:
@monojenkins build pkg and msi

akoeplinger:
@directhex they belong next to the assembly. I'm not really sure why we copy the pdb to the lib/ dir (probably leftover from the old .msi)

PR #14425, opened 2019-05-09
Add log profiler support to Mono MSVC build.

lateralusX:
Several fixes to the log profiler in order to work on Windows.

* Add new dynamic library project to build log profiler on Windows MSVC.
* Fix dependecy between log profiler and Mono.
* Rewrite command pipe logic on Windows since it's not supported to select
  on both sockets and pipes.
* Several additional socket fixes needed for log profiler to work on Windows.

NOTE, this makes sure we have a log profiler that can be started/stopped and at least generate some basic output. Still needs more testing on Windows since there might be other parts of the profiler not working as expected (since it has not been run for quite some time).

### DIFF
--- a/mono/profiler/log.c
+++ b/mono/profiler/log.c
@@ -4,6 +4,7 @@
  * Authors:
  *   Paolo Molaro (lupus@ximian.com)
  *   Alex Rønne Petersen (alexrp@xamarin.com)
+ *   Johan Lorensson (lateralusx.github@gmail.com)
  *
  * Copyright 2010 Novell, Inc (http://www.novell.com)
  * Copyright 2011 Xamarin Inc (http://www.xamarin.com)
@@ -60,13 +61,26 @@
 #if defined(__APPLE__)
 #include <mach/mach_time.h>
 #endif
+#ifndef HOST_WIN32
 #include <netinet/in.h>
+#endif
 #ifdef HAVE_SYS_MMAN_H
 #include <sys/mman.h>
 #endif
+#ifndef HOST_WIN32
 #include <sys/socket.h>
+#endif
 #if defined (HAVE_SYS_ZLIB)
 #include <zlib.h>
+#endif
+
+#ifdef HOST_WIN32
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#endif
+
+#ifndef HOST_WIN32
+#define HAVE_COMMAND_PIPES 1
 #endif
 
 // Statistics for internal profiler data structures.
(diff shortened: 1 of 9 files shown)


mono/profiler/log.c
```diff
@@ -2862,13 +2881,63 @@ cleanup_reusable_samples (void)
...
+				int max_fd = -1;
+
+				FD_ZERO (&wfds);
+				FD_SET (client_socket, &wfds);
+
+				/*
+				* Include timeout to prevent hanging on connect call.
+				*/
+				struct timeval tv = { .tv_sec = 1, .tv_usec = 0 };
+				select (client_socket + 1, NULL, &wfds, NULL, &tv);
```
jaykrell: Designated initializers generally removed from tree for  C++-compat.

jaykrell: WaitForMultipleObjects likely could use pipes & sockets together.

lateralusX: To make that work reliable you would need to move over to WSA API's and then you will need to use event's an implement that logic together with network events to get something similar. Didn't want to go in to much depth doing that, just fixing this command side channel, only used during shutdown, only passing a command that is represented as a char. Current change will keep most of the logic as is.

lateralusX:
@monojenkins rebuild failed

mono/profiler/log.c
lateralusX: The initializer was a copy from the other select in this source, so we should probably fix both. Maybe that could be done in separate PR.

jaykrell: Fair enough. I have ported this directory to C++ at least once before..

jaykrell: Btw I don't think you need to use WSA for wait, I think just regular WaitForMultipleObjects will do.
Regular sockets are in kernel and it just works.
Sockets can be implemented I guess via a usermode provider, but those are then wrapped up by a kernel driver to restore kernel wait functions to working. As I understand.

jaykrell: But fair enough to both points.

jaykrell:
Notice the implementation of `WSAWaitForMultipleEvents`:.
```
C:\s>\bin\amd64\cdb curl

0:000> u ws2_32!WSAWaitForMultipleEvents

WS2_32!WSAWaitForMultipleEvents:
00007ffd`00443e90 48ff2579200300  jmp     qword ptr [WS2_32!_imp_WaitForMultipleObjectsEx (00007ffd`00475f10)]
```

lateralusX:
I was not primarly thinking about the wait method but setting up similar pattern using it, most documentation points to WSASelectEvent + WSAEnumNetworkEvents for this or switch to IO completion ports. Mixing that with anonymous pipes was not worth the effort and didn't want to change it that much (and I wanted to limit impact on other platforms as well).

akoeplinger:
@lateralusX I guess we should add the new library to the Windows MSI ?

lateralusX:
@akoeplinger  I believe we could, file is called, mono-profiler-log.dll.

Issue #14376, opened 2019-05-07
Split more public API's into public/internal called during init to avoid additional GC unsafe enter/exit calls.


lateralusX:
During work, porting hybrid/cooperate suspend over to Windows, I looked at the number of GC unsafe enter/exit just happening during runtime init. Turns out that there is quite a few of those transitions steaming from the fact that we call a couple of Mono public API's, like (there are probably more):

mono_g_hash_table_new_type
mono_g_hash_table_insert
mono_class_is_subclass_of
mono_domain_set
mono_config_for_assembly

In similar situation we split a public API into a public version doing the GC unsafe transition and an internal function just doing a check included in checked build that current thread is already in GC unsafe mode. We should investigate if we could apply the same pattern to at least some of the above methods and switch to calling the internal method during boot if we already know that the thread is in GC unsafe mode.

jaykrell:
Right, I did not finish that work.

lambdageek:
@CoffeeFlux This is another good task to do to get familiar with the code in the runtime.

PR #14427, opened 2019-05-09
[coop] Transition various public APIs into an external/internal form to avoid unnecessary GC-unsafe calls

CoffeeFlux:
Relevant issue: #14376

There are potentially more functions this can apply to, but this covers the ones listed in the issue and an extra one I found in the process.

Functions were either broken off into an external-only and internal version, or such a pair already existed and the internal version was moved from the source file to a separate `-internals` header, similarly to in 3c74df566776f468366d96dd164dd1a6b3d780cc.

This gets slightly complicated with `mono_domain_set` and `mono_domain_set_internal`, both of which are currently public (thanks to 4beada28924879ede55f3bf8b3f5b798d0f4637e) and have different signatures. Additionally, the latter does not actually include the gc-unsafe transition despite being public. 
I've opted to introduce `mono_domain_set_fast` as the 'internal' version of `mono_domain_set`, to add the proper gc-unsafe transition to `mono_domain_set_internal`, and then to migrate internal usage to `mono_domain_set_internal_with_options`. As part of that I've marked `mono_domain_set_internal` as external only, which is unfortunate naming-wise but I believe the best option here without introducing API breakage.


mono/metadata/appdomain.c
```diff
@@ -987,12 +989,21 @@ mono_domain_owns_vtable_slot (MonoDomain *domain, gpointer vtable_slot)
 gboolean
 mono_domain_set (MonoDomain *domain, gboolean force)
 {
+	gboolean result;
+	MONO_ENTER_GC_UNSAFE;
```
lambdageek: I think the external-only `mono_domain_set` should still do the `if (!force && ...)` stuff before entering GC Unsafe mode.  And then delegate directly to `mono_domain_set_internal_with_options`.

ie: `mono_domain_set` and `mono_domain_set_fast` are identical except the latter does a GC transition in the case where we've actually got to do some work.

mono/metadata/mono-hash-internals.h
```diff
@@ -0,0 +1,15 @@
...
+
+#ifndef __MONO_G_HASH_INTERNALS_H__
+#define __MONO_G_HASH_INTERNALS_H__
+
+#include "mono/metadata/mono-hash.h"
+#include "mono/metadata/mono-gc.h"
+
+MonoGHashTable *
+mono_g_hash_table_new_type_internal (GHashFunc hash_func, GEqualFunc key_equal_func, MonoGHashGCType type, MonoGCRootSource source, void *key, const char *msg);
+void mono_g_hash_table_insert_internal (MonoGHashTable *h, gpointer k, gpointer v);
```
lambdageek: `void` should be on its own line.

mono/metadata/mono-config-internals.h
```diff
@@ -0,0 +1,12 @@
+/**
+ * \file
+ */
+
+#ifndef __MONO_METADATA_CONFIG_INTERNALS_H__
+#define __MONO_METADATA_CONFIG_INTERNALS_H__
+
+#include "mono/metadata/mono-config.h"
+
+void mono_config_for_assembly_internal (MonoImage *assembly);
```
lambdageek: `void` should be on its own line

mono/metadata/appdomain-internals.h
```diff
@@ -0,0 +1,12 @@
+/**
+ * \file
+ */
+
+#ifndef _MONO_METADATA_APPDOMAIN_INTERNALS_H_
+#define _MONO_METADATA_APPDOMAIN_INTERNALS_H_
+
+#include "mono/metadata/appdomain.h"
+
+mono_bool mono_domain_set_fast (MonoDomain *domain, mono_bool force);
```
lambdageek: could be `gboolean` here.

mono/metadata/mono-hash.c
```diff
@@ -481,10 +481,17 @@ void
 mono_g_hash_table_insert (MonoGHashTable *h, gpointer k, gpointer v)
```
jaykrell: A later PR will move this to `external-only.c`, and build a runtime variant without it?

lambdageek: @jaykrell my fault. forgot we have that now.
@CoffeeFlux let's move these.

mono/metadata/appdomain.h
```diff
@@ -82,10 +82,10 @@ mono_domain_get_id         (MonoDomain *domain);
 MONO_API const char *
 mono_domain_get_friendly_name (MonoDomain *domain);
 
-MONO_API mono_bool
+MONO_API MONO_RT_EXTERNAL_ONLY mono_bool
 mono_domain_set            (MonoDomain *domain, mono_bool force);
 
-MONO_API void
+MONO_API MONO_RT_EXTERNAL_ONLY void
```
jaykrell: “Internal” is public? :(

CoffeeFlux: I mentioned this in the PR description but yeah, seems someone just made it public alongside everything else in the file in an older commit... It makes for really awkward naming, but its behavior isn't even the same as `mono_domain_set` and I'm trying to avoid API breakage.

jaykrell: Avoid breakage definitely.
Stuff will get broken, but at least by trying not to, less is broken.

mono/metadata/mono-hash.c
CoffeeFlux: everything should now be moved over, and it Builds For Me(tm)

CoffeeFlux:
While migrating things to `external-only.c` I noticed a couple more functions in `class.c` that already have the split made but where the internal version isn't in the internals header: `mono_class_get_nullable_param` and `mono_class_from_mono_type`. This probably shouldn't be merged until I get a chance to look at those.

vargaz:
The new file could be called 'embedding-api.c'.

CoffeeFlux:
Both those functions don't seem to be called externally, so with the latest commit I think this should be ready for review/merger once it passes CI.

jaykrell:
Externally doesn’t mean dis or profiler but completely outside mono repo, like unity or stuff unknown to us. Third party (or fourth depending on how you count.)
 Possibly visualstudiomac or xamarinios or xamarinandroid but hopefully not.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Given that the assertion was failing on CI [here](https://github.com/mono/mono/blob/ac2f871e90e42b15e1c623d9392f08b329574acb/mono/metadata/threads.c#L6109), I think it's simply not valid - this function can be called while in state BLOCKING. I've reverted the previous commit and removed the assertion as a result. I could split this off into a third function if deemed necessary, but I don't think it's worth the complexity.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #14431, opened 2019-05-09
Reduce malloc in JIT icall registration.

jaykrell:
There is a little more to do (in a later PR).

Need to decide here, pointer or enum. Either work.
Very minor difference either way.

Make duplicate registration checking stricter.

### DIFF
--- a/mono/metadata/class-internals.h
+++ b/mono/metadata/class-internals.h
@@ -583,6 +583,7 @@ typedef struct {
 	MonoMethodSignature *sig;
 	const char *c_symbol;
 	MonoMethod *wrapper_method;
+	gboolean inited;
 } MonoJitICallInfo;
 
 void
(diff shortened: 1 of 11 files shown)


jaykrell:
Actually I think this should have the duplicate checking.
Because the semantics are otherwise very subtely and unnecessarily altered.
When a function is reused, the old code would make a whole new info.
The new way unions the infos.

jaykrell:
@monojenkins build failed

jaykrell:
Very tempting to increase this PR.
The next steps are finishing off all registration, there isn't much left.
And the switching hash table enumeration to be array enumeration.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/class-internals.h
```diff
@@ -1067,28 +1068,64 @@ MONO_API gboolean
 mono_metadata_load_generic_param_constraints_checked (MonoImage *image, guint32 token,
 					      MonoGenericContainer *container, MonoError *error);
 
+// Tbis is the "real" function for registering JIT icalls. All others are one line wrappers that call it,
```
rjvdboon: Nit: typo: Tbis should be This

jaykrell: There is another round of changes in a later PR, maybe this PR (depending on overall feedback). I'll fix it then, so to not upset the CI green and almost green status.

mono/metadata/icall.c
```diff
@@ -9135,52 +9136,80 @@ mono_register_jit_icall_wrapper (MonoJitICallInfo *info, gconstpointer wrapper)
...
-		jit_icall_hash_name = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, g_free);
+		jit_icall_hash_name = g_hash_table_new_full (g_str_hash, g_str_equal, NULL, mono_jit_icall_info_free);
 		jit_icall_hash_addr = g_hash_table_new (NULL, NULL);
 	}
 
-	if ((info = (MonoJitICallInfo *)g_hash_table_lookup (jit_icall_hash_name, name))) {
-		g_warning ("jit icall already defined \"%s\" \"%s\" %p %p\n", name, info->name, func, info->func);
-		g_assert_not_reached ();
+	// Do not allow duplicate registration, either name or function or info,
+	// except the function mono_no_trampolines is reused.
```
vargaz: Which functions are registered multiple times ?

mono/metadata/marshal.c
```diff
@@ -117,8 +117,14 @@ get_method_image (MonoMethod *method)
 	return m_class_get_image (method->klass);
 }
 
+// func is an identifier, that names a function, and is also in jit-icall-reg.h,
+// and therefore a field in mono_jit_icall_info and can be token pasted into an enum value.
+//
+// The name of func must be linkable for AOT, for example g_free does not work (monoeg_g_free instead),
+// nor does the C++ overload fmod (mono_fmod instead). These functions therefore
+// must be extern "C".
 #define register_icall(func, sig, no_wrapper) \
```
vargaz: Not sure this block of comments need to be repeated multiple times.

mono/metadata/icall.c
jaykrell: There were 3 or 4 but now maybe just 1. There were opcode emulations registered w/ and w/o wrappers, the 2 get_lmf_addr functions, & no_trampoline registered repeatedly.
The 2 lmf functions were probably 2 functions, that I made 1. Compiler/linker could still make them the same even if I resplit them. Gcc won’t do that becuase they are address-taken but msvc still might.
I can report back on the opcode emulation. I hit that back in January when first did this change.

mono/metadata/marshal.c
jaykrell: Agreed it is annoying and not sure what to do. 

If they were identical, move to header, but they aren’t.

And what to document vs. read the code.

People follow contradictory styles — literate programming where comments are like a book and no need to read the code, or code is eminently readable and comments are for the rare hard to understand code, to be avoided.

mono/metadata/icall.c
jaykrell: https://github.com/mono/mono/pull/12210/files

was the duplicate opcode emulation registration.
I wonder if this is the right/ideal fix — maybe just one name and never wrapper.

Issue #14311, opened 2019-05-02
Unable to get ANY proper ssl certificate validation (TLS 1.2) in mono.


Ub1Catcrush:
**The core issue is we cannot get a proper ssl certificate validation (TLS 1.2) running.
The attached program is to show current issues with SSL certificate validation in Mono.**
============================
[MonoSSLTest.zip](https://github.com/mono/mono/files/3137792/MonoSSLTest.zip)

Everything was tried on windows 10 with VS 2017 15.9.7.
used mono was is at state ofl 21.4.2019 with BTLS enabled.
The tryout is on console but our target platform is an embedded AOT mono on W64/XB1/PS4/Switch.

**1. Micosoft SDK**
---------------------
On microsofts windows sdk validating a certificate works flawlessly

A log for system validation can be found in _examplelogs\ssl_test_windows_system.log_
[ssl_test_windows_system.log](https://github.com/mono/mono/files/3137802/ssl_test_windows_system.log)

**2. Micosoft SDK with OpenSSL net (github openssl-net with 32bit 1.0.2r windows libs"):**
---------------------
On microsofts windows sdk also this way to validate a certificate works flawlessly

A log for openssl validation can be found in _examplelogs\ssl_test_windows_openssl.log_
[ssl_test_windows_openssl.log](https://github.com/mono/mono/files/3137801/ssl_test_windows_openssl.log)

**3. MONO legacy:**
---------------------

It simply quits with and exception during AuthenticateAsClient:

```
IOException The authentication or decryption has failed.
INNER IOException The authentication or decryption has failed.
  at Mono.Security.Protocol.Tls.SslStreamBase.EndRead (System.IAsyncResult asyncResult) [0x0004b] in <2ba552739e354a84bf029a3cd0dcbc6d>:0 
  at Mono.Net.Security.Private.LegacySslStream.EndAuthenticateAsClient (System.IAsyncResult asyncResult) [0x0000e] in <5c715dc7cd084e1b9a6e9162f2fcbdc0>:0 
  at Mono.Net.Security.Private.LegacySslStream.AuthenticateAsClient (System.String targetHost, System.Security.Cryptography.X509Certificates.X509CertificateCollection clientCertificates, System.Security.Authentication.SslProtocols enabledSslProtocols, System.Boolean checkCertificateRevocation) [0x0000e] in <5c715dc7cd084e1b9a6e9162f2fcbdc0>:0 
[...]

PR #14435, opened 2019-05-10
Bump boringssl to include fix for Windows path's in lookup dirs.

lateralusX:
Fixes mono/mono#14311.

Current boringssl implementation uses ':' as path list separator, meaning that Windows paths doesn't work. By default our user and machine store paths on Windows are located here:

Default User Store Path:
C:\Users[current user]\AppData\Roaming.mono\new-certs\Trust

Default Machine Store Path:
C:\ProgramData.mono\new-certs\Trust

Fix is to use (as done in openssl repro) separators depending on platform.

The default OPENSSLDIR was also not accounting for none Unix platforms in boringssl repro, this sets the default values on Windows as you get when running configure in openssl.


lateralusX:

lateralusX:
@monojenkins rebuild failed

jaykrell:
Consider using the OS-provided SSL.
schannel I think.
https://docs.microsoft.com/en-us/windows/desktop/com/schannel

PR #14460, opened 2019-05-12
configure.ac: bail out if C++ is not found

### LINKED ISSUE #14195: 5.20.1.19 configure fails if enable-cxx but no C++ compiler found
## Steps to Reproduce

./configure --enable-cxx

### Current Behavior

No C++ available => CXX command = "no", and subsequent test for size_t fails.

    checking if size_t is unsigned long... no
    checking if size_t is unsigned int... no
    checking if size_t is unsigned __int64... no
    checking if size_t is unsigned long long... no
    ./configure: line 421: test: int: integer expression expected
    configure: error: Unable to determine size_t among unsigned long
    ./configure: line 310: return: int: numeric argument required
    ./configure: line 320: exit: int: numeric argument required

### Expected Behavior

./configure should either fail early when an invalid C++ compiler was found (i.e. "no" - which in itself is a bug?), or tolerate that a C++ compiler doesn't exist and ignore enable-cxx.

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

5.20.1.19

ffontaine:
Since version 5.16.0.179 and d8af775, mono requires the availability of
a working C++ compiler to get the exact type of size_t.

So bail out if CXX is set to no by AC_PROG_CXX and remove comment about
automake/libtool being broken as mono is now really requiring C++

Fix #14195

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -573,12 +573,11 @@ dnl We should use AM_PROG_AS, but it's not available on automake/aclocal 1.4
 AC_SUBST(CCAS)
 AC_SUBST(CCASFLAGS)
 
-# AC_PROG_CXX helpfully sets CXX to g++ even if no c++ compiler is found so check
+# With old versions of autoconf, AC_PROG_CXX helpfully sets CXX to g++ even if no c++ compiler is found so check
 # GXX instead. See http://lists.gnu.org/archive/html/bug-autoconf/2002-04/msg00056.html
-if test "x$CXX" = "xg++"; then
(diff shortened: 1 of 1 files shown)


akoeplinger:
Looks like you raced with @angeloc in https://github.com/mono/mono/pull/14461 :)

Since that one already got some build results I'll close this here but thank you for your contribution! ❤️

PR #14461, opened 2019-05-12
configure: check for c++ availability

angeloc:
Since commit d8af775, c++ compiler is required to compile.
This patch checks if the c++ compiler is available as soon as possible
and returns an error in case it is not found.

Fixes: https://github.com/mono/mono/issues/14195

Signed-off-by: Angelo Compagnucci <angelo@amarulasolutions.com>


lambdageek:
@monojenkins build

PR #14473, opened 2019-05-14
[do not merge] Work towards removing JIT icall hashing.

jaykrell:
leftover from previous large PR, see if it is small now, else continue to make it incremental
also fix any merging


jaykrell:
bad merge, but has some stuff to come back to later

PR #14474, opened 2019-05-14
Work towards removing JIT icall hashing.

jaykrell:


jaykrell:
@monojenkins build failed

vargaz:
I'd prefer using the newly added enums to identify the icalls instead of using &mono_jit_icall_info everywhere.

jaykrell:
Ok I can do that. Retroactively a bit probably too.

jaykrell:
And yes, this was a good time to make that decision before it gets further.

I will probably hold on the retroactive part, unless you want to weigh on a naming matter:

recent:
`mono_register_jit_icall_info (MonoJitICallInfo *info,...`

1. leave alone
2. `mono_register_jit_icall_info (MonoJitICallId, …)`
3. `mono_register_jit_icall_id (MonoJitICallId, …)`
4. `mono_register_jit_icall_index (MonoJitICallId, …)`
5. other
6. later


I think renaming the function is cleanest, but can be later.
The function would then gain as its first line to form the address -- as will occur other places.

jaykrell:
Typesafety might suffer with the enums.
But it is hard to argue against:
1. Micro optimization. Enums are relocation-free and 32bits. Occasionally cheaply manually relocated.
2. Enums allow indexing into multiple tables, if the data is split up.
Enums better model database primary keys.
Pointers better model single-table database.

jaykrell:
2 or so pointers remain, should wait for a later PR -- where we put enums in the generated IL.
And the older register w/ info remains, and that should be fairly easy to change, just limiting PR size really.

mono/metadata/marshal-ilgen.c
```diff
@@ -589,82 +589,82 @@ conv_to_icall (MonoMarshalConv conv, int *ind_store_type)
 	*ind_store_type = CEE_STIND_I;
 	switch (conv) {
 	case MONO_MARSHAL_CONV_STR_LPWSTR:
-		return (gpointer)mono_marshal_string_to_utf16;
```
jaykrell: GitHub presented the diff of the other form more readably. :(

jaykrell:
@monojenkins build failed

mono/metadata/method-builder-ilgen.c
```diff
@@ -544,10 +544,16 @@ mono_mb_emit_native_call (MonoMethodBuilder *mb, MonoMethodSignature *sig, gpoin
 }
 
 void
-mono_mb_emit_icall (MonoMethodBuilder *mb, gpointer func)
+mono_mb_emit_icall_id (MonoMethodBuilder *mb, MonoJitICallId jit_icall_id)
 {
```
vargaz: The old function name can stay, it would reduce the diff.

jaykrell: I'll try that. Elsewhere we need a macro with the old name to reduce diff, and occasional use of function for less common but still recursing cases. I'll check.

jaykrell: Maybe but I don't think so.
I kept the name as a macro for the common case -- when there is a constant identifier and not a variable, but then some cases cannot use that, and call the new function.
The ratio is about 25%,.

jay@jaykrell-tp2:/mnt/c/s/mono2/mono$ grep  mono_mb_emit_icall_id */*.{c,h} |wc -l
28
jay@jaykrell-tp2:/mnt/c/s/mono2/mono$ grep  mono_mb_emit_icall\  */*.{c,h} |wc -l
98

vargaz: Makes sense.

jaykrell:
@monojenkins build failed

PR #14475, opened 2019-05-14
[netcore] Exclude some unused icalls

filipnavara:


filipnavara:
Build failure on ARMv7 is unrelated and was happening for quite some time: https://github.com/mono/mono/issues/14486

filipnavara:
> I am not against this change but I think it'd be even more beneficial to have netcore specific icall source file instead

In general I agree. There should also be a build step that checks the native icall list against the managed one in CoreLib. The reason I went ahead with this approach was mainly the changes in `locale.{c,h}` and `object-internals.h` where incorrect object layouts were exposed to native code.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

jaykrell:
I don't think ifdef is the way unless the old code is really doing to be deleted, but:
https://github.com/mono/mono/pull/14489

PR #14476, opened 2019-05-14
[do not merge] Work toward eliminating *.md -- first make the generated C more readable.

jaykrell:


jaykrell:
Promising but not in good shape and not sure it can go much further if it was working.

Issue #14234, opened 2019-04-25
SqliteMembershipProvider initialization bug (Failed to map path '/').


musikhin:
## Steps to Reproduce

Please find the file attached with the steps to reproduce the bug.
[Steps to reproduce.pdf](https://github.com/mono/mono/files/3118316/Steps.to.reproduce.pdf)

### Current Behavior

In case when the asp.net application hosted not under the root location we get an error "Failed to map path '/'" at the first application run.

### Expected Behavior

Possibility to launch the asp.net application at the custom location without any errors =)

Needs to replace the 142 code line in SqliteMembershipProvider with 
m_MachineKey = (MachineKeySection)WebConfigurationManager.GetSection("system.web/machineKey");

## On which platforms did you notice this

[ ] macOS
[x] Linux (CentOS 7.4 + apache 2.4.6)
[ ] Windows

**Version Used**:
5.16.0.220.
5.20.1.19

## Stacktrace

  at System.Web.HttpRequest.MapPath (System.String virtualPath, System.String baseVirtualDir, System.Boolean allowCrossAppMapping) [0x0011a] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
  at System.Web.HttpRequest.MapPath (System.String virtualPath) [0x00012] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
  at System.Web.Hosting.HostingEnvironment.MapPath (System.String virtualPath) [0x00033] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
  at System.Web.Hosting.DefaultVirtualPathProvider.DirectoryExists (System.String virtualDir) [0x00013] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
  at System.Web.Configuration.WebConfigurationManager.FindWebConfig (System.String path, System.Boolean& inAnotherApp) [0x00019] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
  at System.Web.Configuration.WebConfigurationManager.OpenWebConfiguration (System.String path, System.String site, System.String locationSubPath, System.String server, System.String userName, System.String password, System.Boolean fweb) [0x0001d] in <979c6d3c1df54a4b9abbd8b5ecbe7cbb>:0 
[...]

PR #14479, opened 2019-05-14
 Fixes #issue-14234

musikhin:
Dear team Mono,

Please find below the fix of issue #14234 .  On the stage of the initialization of the SqliteMembershipProvider the existed code always produced an error "Failed to map path '/'" in case if the ASP.NET application is placed not at the root path. Detailed step by step description of the problem is provided in the issue description.

The bug was that was implemented the call of the function with wrong parameters list. So the fix was in changing the function call. The same function call is also used in SqlMembershipProvider.cs.

Thanks for your job.

Regards,
Alex

### DIFF
--- a/mcs/class/System.Web/System.Web.Security/SqliteMembershipProvider.cs
+++ b/mcs/class/System.Web/System.Web.Security/SqliteMembershipProvider.cs
@@ -139,7 +139,7 @@ namespace System.Web.Security
 			}
 
 			// Get encryption and decryption key information from the configuration.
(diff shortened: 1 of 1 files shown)


musikhin:
Dear @marek-safari, could you please explain what should I do to start code review? This fix is business critical for us. 

Currently, we make these changes manually and rebuild Mono for each version.

marek-safar:
build

steveisok:
I have no idea why the original code tried to map path anything.  LGTM

lewurm:
@monojenkins squash

PR #14481, opened 2019-05-15
Removing some remoting and multi-appdomain APIs on wasm and the AOT test profiles.

baulig:
Build / profile changes:
------------------------

* `build/profiles/testing_aot_common.make`: define `DISABLE_REMOTING = yes` and
  `NO_MULTIPLE_APPDOMAINS = yes`.

* `mcs/class/corlib/Makefile`: check `DISABLE_REMOTING` when setting
  `REFERENCE_SOURCES_FLAGS` (and don't define `FEATURE_REMOTING`, `MONO_COM`
   and `FEATURE_COMINTEROP`).

* added new `testing_aot_common_corlib.dll.exclude.sources`.  Several types in
  the `System.Runtime.Remoting.Messaging` namespace are used by the runtime in
  non-remoting scenarios, so those need to be preserved.
  For all other remoting namespaces, we only need a few stripped down stub versions
  of the some of the types; those have been added to a new `legacy` directory, so
  we can wildcard-exclude all the files here.

* added new `testing_aot_common_corlib.dll.sources`; this currently only contains
  `legacy/*.cs` to list above mentioned stub versions.

* added new `testing_aot_common_corlib_test.dll.exclude.sources` as a common test
  excludes file.

API Changes (conditional to `DISABLE_REMOTING`):
------------------------------------------------

* `mcs/class/corlib/legacy/`: add stub-versions for some of the remoting-related
  classes that are used by the runtime (so we can't completely remove those types):

  - `System.MarshalByRefObject`.
  - `System.Runtime.Remoting.IRemotingTypeInfo`.
  - `System.Runtime.Remoting.ObjectHandle`.
  - `System.Runtime.Remoting.RemotingServices`.
  - `System.Runtime.Remoting.Activation.ActivationServices`.
  - `System.Runtime.Remoting.Contexts.Context`.
  - `System.Runtime.Remoting.Messaging.ILogicalThreadAffinative`.
  - `System.Runtime.Remoting.Messaging.LogicalCallContext`.
  - `System.Runtime.Remoting.Proxies.TransparentProxy`.
  - `System.Runtime.Remoting.Proxies.RealProxy`.

  some of these have fields that need to be preserved; these are marked with
  appropriate `#region` pragmas.

All changes below are conditional to `DISABLE_REMOTING`.

* `System.Runtime.Remoting.Messaging.AsyncResult`: this class is used by some of
  the async code, so we can't completely remove it.  However, we can remove the
  `MonoMethodMessage call_message` field when `DISABLE_REMOTING`.

* `System.Runtime.Remoting.Messaging.MonoMethodMessage`: this class is also used
  by some runtime code and thus needs to be preserved. However, when `DISABLE_REMOTING`
  is defined, then we make some changes to it:
  - remove all iterfaces (`IMethodCallMessage`, `IMethodReturnMessage`, and `IInternalMessage`).
  - `Properties` throws `PlatformNotSupportedException`.

* `System.Runtime.InteropServices.Marshal`: disable some COM-related code.

* `System.Threading.Thread`: remove `CurrentContext` property.

* `System.Threading.ExecutionContext`: adjust conditional logic; don't define
  `FEATURE_REMOTING` if `DISABLE_REMOTING`.

* `System.Activator`: add `DISABLE_REMOTING` to the existing
   `#if FEATURE_REMOTING || MOBILE_LEGACY` conditional in this class.

* `System.AppDomain`: since we only have one domain, `DefaultDomain` will always
   return the root domain; comment out the call to `RemotingServices.GetDomainProxy`
   as well as some internal methods.

* `System.__ComObject`: stub out on `DISABLE_REMOTING` as well as `!FULL_AOT_RUNTIME`.

* `System.Runtime.Serialization.Formatters.Binary.BinaryFormatter`: adjust `IFormatter`
  versus `IRemotingFormatter` interface logic to use the former when `DISABLE_REMOTING`.

* `System.Runtime.Serialization.Formatters.Binary.BinaryObjectReader`: adjust existing
  `#if FEATURE_REMOTING || MOBILE_LEGACY` conditional to add `DISABLE_REMOTING` to it.

* `System.Diagnostics.CorrelationManager`: comment out some `CallContext.LogicalGetData()`
  and `CallContext.LogicalSetData`.

* `System.Diagnostics.TraceEventCache`: likewise.

Test changes:
-------------

* `mono/tests/Makefile.am`: conditionally disable some remoting tests.

* `corlib/Test/System.Runtime.Serialization/SerializationTest.cs`: conditionally
   disable all tests requiring remoting.

* `corlib/Test/System/ActivatorTest.cs`: same; trying to preserve as much as possible
  in this test.

* added some `DISABLE_REMOTING` and `DISABLE_SECURITY` conditionals to some of the tests.


baulig:
@monojenkins build failed

baulig:
@monojenkins build failed

baulig:
@monojenkins build failed

baulig:
@monojenkins build failed

mcs/class/corlib/Test/System.Reflection/FieldInfoTest.cs
```diff
@@ -1371,6 +1371,7 @@ public void SetValueDirect ()
 			Assert.AreEqual ("B", fields.str);
 		}
 
+#if !DISABLE_REMOTING
```
filipnavara: In general, is there reason to exclude these tests with #if as opposed to using `Category ("Remoting")` and excluding that by filter passed to NUnit?

baulig:
@monojenkins commit csproj

mcs/class/corlib/Test/System.Reflection/FieldInfoTest.cs
baulig: My general policy is that passing incorrect include / exclude parameters to NUnit should be "safe" in the way that you'd get failing tests, but not crash the test runner with an assertion in native code.

I tried to use `[Category]` attributes as much as possible in this PR, but there are two situations here where `#if` conditionals were required:

1.) Code that wouldn't even compile due to remoting APIs being removed.
2.) Code that when run (by accident, ie. by passing incorrect flags to NUnit) would cause a native assertion.

The second situation happens when you build the Mono runtime with remoting support and then use a corlib that has certain types removed.

In theory, there shouldn't be any API entry points where this might happen, but I've seen it in a few places.

baulig:
@monojenkins build failed

PR #14484, opened 2019-05-15
Make genmdesc output much smaller.

jaykrell:
```
const MonoInstSpec mono_amd64_desc [] = {
  {{   0,    0,    0,    0,    0,    0}}, // 0 null entry
//  dest  src1  src2  src3   len  clob
// ----- ----- ----- ----  ----- -----
  {{   0,    0,    0,    0,    0,    0}}, // 1 nop
  {{   0,    0,    0,    0,    1,    0}}, // 2 hard_nop
  {{   0,    0,    0,    0,    2,    0}}, // 3 relaxed_nop
  {{   0,  'i',  'i',    0,    3,    0}}, // 4 compare
  {{   0,  'i',    0,    0, 0x0D,    0}}, // 5 compare_imm
  {{   0,  'f',  'f',    0, 0x0D,  'a'}}, // 6 fcompare
//{{   0,  'f',  'f',    0, 0x0D,  'a'}}, // 6 rcompare
//{{   0,  'i',  'i',    0,    3,    0}}, // 4 lcompare
//{{   0,  'i',  'i',    0,    3,    0}}, // 4 icompare
  {{   0,  'i',    0,    0,    8,    0}}, // 7 icompare_imm
//{{   0,  'i',    0,    0, 0x0D,    0}}, // 5 lcompare_imm
  {{ 'i',  'i',    0,    0, 0x60,    0}}, // 8 localloc
  {{ 'i',    0,    0,    0, 0x60,    0}}, // 9 localloc_imm
.
.
.
```

The preexisting indirection is used to reuse rows.
Otherwise the indirection was costing more it was saving -- the 2 bytes per row is more than the 6 bytes per skipped row. Perhaps the point wasn't compression but special values. Anyway, it can be easily pressed into service for compression w/o changing the reader. Assuming row pointer equality isn't a thing.


jaykrell:
@monojenkins build failed

mono/mini/genmdesc.py
```diff
@@ -146,53 +146,92 @@ def parse_input(infile, opcodes):
...
     sorted_opcodes.sort (key=lambda op: op.num)
 
     f.write ("/* File automatically generated by genmdesc.py, don't change */\n\n")
 
-    # Write desc table
-
-    f.write ("const MonoInstSpec mono_{0} [] = {{\n".format (symbol_name))
-    f.write ("  {{")
-
+    # Write desc table
```
jaykrell: previous PR managed to introduce wrong newline format, sorry

mono/mini/genmdesc.py
```diff
@@ -146,53 +146,92 @@ def parse_input(infile, opcodes):
...
+        f.write ("   0")
+        if i != MONO_INST_MAX - 1:
+            f.write (", ")
+    f.write ("}}, // 0 null entry\n")
+    idx = 1
+
+    # Write the heading ever few rows for readability.
+    heading_row = 40
+
+    # Very many of the lines in the table are the same,
```
jaykrell: There might be a better way to form this output, in particular an enum, where the enum names are basically the content smushed into an identifier.
That then might be approaching a hand-writable/readable C form.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
I have a Python-free C+11 const/reordered slightly altered cpu-amd64.md design, but it doesn't do this compression..

jaykrell:
By "reordered" I mean cpu-amd64.h doesn't have to change its ordering of lines or within lines. It'd only have to change `foo:bar` to `foo(bar),`. I'll try to get that in a working PR-able state. But it would depend on C++11 at least in one file, because of all the reordering that *.md depend on.

akoeplinger:

PR #14488, opened 2019-05-15
[Wasm] Expand support for dynamic linking

jeromelaban:
These changes update the `release-dynamic` configuration of the wasm runtime:

- Enhance the support for dynamically linked libraries, by including all available emscripten libraries. This makes the runtime a bit larger and enables support for C++ libraries.
- Add support for `IntPtr` in JavaScript bindings.
- Enable `addFunction` support to allow for wasm originating callbacks.

### DIFF
--- a/sdks/wasm/bindings-test.cs
+++ b/sdks/wasm/bindings-test.cs
@@ -51,6 +51,12 @@ public class TestClass {
 		intptr_val = i;
 	}
 
+	public static IntPtr mkintptr_val;
+	public static IntPtr InvokeMkIntPtr() {
+		intptr_val = (IntPtr)42;
+		return intptr_val;
+	}
+
 	public static object obj1;
 	public static object InvokeObj1(object obj)
 	{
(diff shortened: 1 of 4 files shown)


sdks/wasm/bindings-test.cs
```diff
@@ -42,10 +42,17 @@ public class TestClass {
 	}
 
 	public static int int_val;
-	public static void InvokeInt (int i) {
+    public static void InvokeInt(int i)
+	{
```
vargaz: Identation problem.

jeromelaban: Yes... Still battling with those. The convention in this file is not consistent, I'll make another PR to adjust it.

jeromelaban:
I'm working on the IntPtr part, some tests I added are still failing.

jeromelaban:
The boxing/unboxing from and to Javascript is now fixed with the latest commit.

jeromelaban:

jeromelaban:

sdks/wasm/framework/src/WebAssembly.Bindings/Runtime.cs
```diff
@@ -346,9 +346,13 @@ static string GetCallSignature (IntPtr method_handle)
 					res += "s";
 					break;
 				default:
-					if (t.IsValueType)
-						throw new Exception ("Can't handle VT arguments");
-					res += "o";
+					if (t == typeof(IntPtr)) { 
+ 						res += "i";
```
kg: non-commit-blocking question: Is this function hot? It looks like it could potentially get called frequently. Under those circumstances it might be better for this to use a stringbuilder or a fixed-size char[] buffer that gets converted into a string

jeromelaban: It's call during any kind of interop with Javascript. In the context of this PR, it's also used for callbacks. So yes, optimizations there will make sense. Currently, it does not appear sharply in the profiler, as what's called is generally a lot more costly.

jeromelaban: After a second thought, it is only be called when registering a method: 

https://github.com/jeromelaban/mono/blob/527abcd11521282476d53f101c3b39d6c46d2ae3/sdks/wasm/binding_support.js#L723

PR #14489, opened 2019-05-15
Initial icall-def.h / icall-def-netcore.h split.

jaykrell:
Doesn't seem great but maybe core drops away much more?
Maybe there should be icall-def-common.h? But the ifdef is interleaved.
Or maybe each section should be one or two files?


filipnavara:
Thanks. There's more to be pruned (all the System.IO and System.Globalization icalls for a start), but it can be done incrementally. Ideally we should have a way to cross-check the list against the CoreLib binary. It is the only managed code doing icalls on netcore.

jaykrell:
@monojenkins build failed

jaykrell:
Agreed this file should be generated or checked with managed metadata.
Though I not sure the callers of non-netcore are only corlib.
I also don't think ifdef should be used so heavily, before or after this change, but seemingly just me.
i.e. it shouldn't be difficult to build one native runtime that works with either managed runtime and changing configuration shouldn't require recompiling everything. It is micro optimization at high engineering cost imho.

filipnavara:
> Agreed this file should be generated or checked with managed metadata.

CoreCLR checks it in checked builds as part of CI. Generating it is probably not feasible.

> Though I not sure the callers of non-netcore are only corlib.

Correct. This is specific to netcore builds.

> I also don't think ifdef should be used so heavily, before or after this change, but seemingly just me.

Not just you :) However, I am mostly focusing on small incremental improvement and I try to follow the established code patterns.

> i.e. it shouldn't be difficult to build one native runtime that works with either managed runtime and changing configuration shouldn't require recompiling everything. It is micro optimization at high engineering cost imho.

The differences between netcore and non-netcore builds at moment are quite substential in terms of functionality. It may be feasible to converge the code. Depends on what the goals are in long term.

jaykrell:
@monojenkins build failed

jaykrell:
In general you can provide two (n) struct types and dynamically chose.
Even templatize common implementation.. It is rare to really need ifdef. It is generally an optimization, with significant engineering cost. You need it for host OS and sometimes host architecture.

jaykrell:
Now there is an implicit merge conflict here. I'll fix/redo if approval is granted in principle.

filipnavara:
> In general you can provide two (n) struct types and dynamically chose.

That's not the issue at hand here. The netcore builds move a LOT of code from the native runtime to the managed code. The managed code is shared with other .NET runtimes and it has different layout from the old Mono code. In the netcore builds the structures simply don't need to exist on the native side, along with any code using them. In fact, in many cases (like System.IO code) it is not even part of code shipping with the runtime (ie. not part of System.Private.CoreLib.dll).

Having the additional code on the native side and the definitions from old Mono does not _conflict_ with notcore. It just goes unused. That is a LOT of unused code. It is dangerous to have it there because it's not tested in normal cases but it can be accidentally (or deliberately, in case of security exploits) get executed and result in memory corruption. Likewise, someone, in a good faith can try to use some of the structures without realizing that they are not supposed to be used in the netcore configuration.

We should try to prevent any mistakes at compile time.

Shipping single runtime is feasible with some effort. Doing it safely would likely mean adapting large parts of the netcore managed code and deleting the unused native code altogether.

jaykrell:
I agree with maximal static validation, but not ifdef and multiple configuration.
ifdef is overused in general. I understand some code has to be omitted to make it build (ifdef host_win32 can be ok).

Configuration should be more at runtime and less at compile time.
As well, larger optional pieces can be split into separate libraries.

I can't stand compiling things n times, and being unable to test some code without another rebuild because it got omitted from some configuration. Making code harder to test also contributes to bugs.

Anyway I haven't done much about it either, but also follow patterns.
Maybe some day one aot compiler can target all targets. Different area than the icalls.

jaykrell:
I updated to master. I need to learn to build/test netcore though.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

filipnavara:
> As well, larger optional pieces can be split into separate libraries.

A good first step would be to keep the ENABLE_NETCORE exclusions/inclusions at file level. I also expect most of the differences to be in `mono/metadata` even though currently there are few in `mono/mini` too.

PR #14493, opened 2019-05-16
[wasm] Avoid sharing C icall functions between icalls, in wasm to avoid calling them with incorrect signatures.

vargaz:


lewing:
@monojenkins build failed

PR #14498, opened 2019-05-16
Fix Windows Mono LLVM build error on nightly master build.

lateralusX:


lateralusX:
@monojenkins rebuild failed

PR #14500, opened 2019-05-16
[netcore] Fix Math.ILogB corner cases

EgorBo:
This PR fixes 5 tests in `System.Runtime.Extensions.Tests` (MathTests)
It's basically a copy-paste from CoreCLR:
1) https://github.com/dotnet/coreclr/blob/bd7f8e7435b8cba0cca3a198f15fdc2e4764c2fc/src/pal/src/configure.cmake#L869-L892
2) https://github.com/dotnet/coreclr/blob/master/src/pal/src/cruntime/math.cpp#L876-L894

```
System.Runtime.Extensions.Tests  Total: 6466, Errors: 0, Failed: 11, Skipped: 2
```

### DIFF
--- a/mono/metadata/sysmath.c
+++ b/mono/metadata/sysmath.c
@@ -205,17 +205,11 @@ gint32
 ves_icall_System_Math_ILogB (gdouble x)
 {
 	int ret;
-#if !HAVE_COMPATIBLE_ILOGB0
-	if (x == 0.0)
+	if (FP_ILOGB0 != -2147483648 && x == 0.0)
 		ret = -2147483648;
-	else 
-#endif
-
-#if !HAVE_COMPATIBLE_ILOGBNAN
-	if (isnan(x))
+	else if (FP_ILOGBNAN != 2147483647 && isnan(x))
 		ret = 2147483647;
 	else
-#endif
 		ret = ilogb(x);
 	return ret;
 }
@@ -382,17 +376,11 @@ gint32
 ves_icall_System_MathF_ILogB (float x)
 {
 	int ret;
-#if !HAVE_COMPATIBLE_ILOGB0
-	if (x == 0.0f)
+	if (FP_ILOGB0 != -2147483648 && x == 0.0)
 		ret = -2147483648;
-	else 
-#endif
-
-#if !HAVE_COMPATIBLE_ILOGBNAN
-	if (isnan(x))
+	else if (FP_ILOGBNAN != 2147483647 && isnan(x))
 		ret = 2147483647;
 	else
-#endif
 		ret = ilogbf(x);
 	return ret;
 }
--- a/mono/metadata/sysmath.c
+++ b/mono/metadata/sysmath.c
@@ -204,7 +204,20 @@ ves_icall_System_Math_Ceiling (gdouble v)
 gint32
 ves_icall_System_Math_ILogB (gdouble x)
 {
-	return ilogb (x);
+	int ret;
+#if !HAVE_COMPATIBLE_ILOGB0
+	if (x == 0.0)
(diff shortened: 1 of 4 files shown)


netcore/excludes-System.Runtime.Extensions.Tests.rsp
```diff
@@ -0,0 +1,20 @@
...
+-nomethod System.Tests.AppDomainTests.FirstChanceException_Called
+
+# AssemblyLoadContext.InternalLoad is not implemented
+-nomethod System.Tests.AppDomainTests.LoadBytes
+
+# TODO: investigate
+-nomethod System.Tests.AppDomainTests.TypeResolve
+-nomethod System.Tests.AppDomainTests.GetAssemblies
+-nomethod System.Tests.AppDomainTests.ResourceResolve
+-nomethod System.Tests.AppDomainTests.AssemblyLoad
```
filipnavara: The events are not wired up from the native side. In .NET Core the `AppDomain` class forwards them to `AssemblyLoadContext`. Mono on the native side still expects `AppDomain` (renamed to `MonoDomain` for .NET Core) which implements none of the events: https://github.com/mono/mono/blob/7082845114f603b3e174c26d9e8fb7f667bc2a51/mcs/class/System.Private.CoreLib/Mono/MonoDomain.cs#L17-L37

mono/metadata/sysmath.c
```diff
@@ -204,7 +204,20 @@ ves_icall_System_Math_Ceiling (gdouble v)
 gint32
 ves_icall_System_Math_ILogB (gdouble x)
 {
-	return ilogb (x);
+	int ret;
+#if !HAVE_COMPATIBLE_ILOGB0
```
filipnavara: Kinda makes me wonder why FP_ILOGB0 is not tested directly here without all the `autoconf` magic. Apparently it's a constant.

mono/metadata/sysmath.c
```diff
@@ -204,7 +204,20 @@ ves_icall_System_Math_Ceiling (gdouble v)
...
+	if (x == 0.0)
+		ret = -2147483648;
+	else 
+#endif
+
+#if !HAVE_COMPATIBLE_ILOGBNAN
```
filipnavara: Same here for FP_ILOGBNAN.

mono/metadata/sysmath.c
EgorBo: hm.. makes sense indeed.

netcore/excludes-System.Runtime.Extensions.Tests.rsp
EgorBo: Ah ok, so it's basically the same issues as with FirstChanceException event

filipnavara: Yep. I was thinking about fixing it, but I didn't want to step on anyone's toes. I believe @lambdageek was doing some work on `AssemblyLoadContext`, perhaps not directly related to this?

lambdageek: Yea, it's slow going. I got bogged down in refactoring `MonoImage:references` (a list of mono assemblies) to work correctly with ALCs.  I'll try to handoff the work to someone else. but right now it's stuck in a refactoring swamp

filipnavara: This part seems to be quite isolated from the rest of `AssemblyLoadContext` in terms of dependencies. Aside from placing it in the right place in the native code it should not clash much with the rest of the refactoring.

marek-safar:
@lewurm @vargaz could you please review

PR #13947, opened 2019-04-09
[do not merge] Mostly eliminate JIT icall hashing by string and address.

jaykrell:
[do not merge] Almost eliminate JIT icall hashing by string and address.

But, the diff is too large, and fails on a number of lanes (and succeeds on quite a few).
It'll probably have to be split up into digestable, working pieces. And maybe never do some parts.

Succeeds on Mac and Linux, x86 and amd64.
Fails on Windows and arm and arm64, and some configurations.

PR #14507, opened 2019-05-17
Reduce JIT icall hash use.

jaykrell:
Reduce JIT icall hash use.

### DIFF
--- a/mono/metadata/jit-icall-reg.h
+++ b/mono/metadata/jit-icall-reg.h
@@ -350,3 +350,15 @@ mono_check_jit_icall_info (gconstpointer jit_icall_info)
 
 	return (MonoJitICallInfo*)jit_icall_info;
 }
+
+static inline MonoJitICallInfo*
+mono_jit_icall_info_at (int index)
+{
+	g_assert (index && (guint)index < MONO_JIT_ICALL_count);
+
+	MonoJitICallInfo *info = &mono_jit_icall_info.array [index];
+
+	g_assert (info->func);
+
+	return info;
+}
--- a/mono/metadata/jit-icall-reg.h
+++ b/mono/metadata/jit-icall-reg.h
@@ -342,7 +342,7 @@ static inline MonoJitICallInfo*
 mono_check_jit_icall_info (gconstpointer jit_icall_info)
 {
 	// All pointers to MonoJitICallInfo must be within mono_jit_icall_info.
-	// This will enable converting to and from a small integer or enum.
+	// This enables converting to and from a small integer or enum.
 	g_assert ((char*)jit_icall_info >= (char*)&mono_jit_icall_info && (char*)jit_icall_info < (char*)(&mono_jit_icall_info + 1));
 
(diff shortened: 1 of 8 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
Gradually working toward https://github.com/mono/mono/pull/13947.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Failure doesn't seem related. Hm.

mono/metadata/jit-icall-reg.h
```diff
@@ -350,3 +350,15 @@ mono_check_jit_icall_info (gconstpointer jit_icall_info)
 
 	return (MonoJitICallInfo*)jit_icall_info;
 }
+
+static inline MonoJitICallInfo*
+mono_jit_icall_info_at (int index)
+{
```
vargaz: I would call this mono_find_jit_icall_info (). It should also take an enum typed argument.

jaykrell:
std:;vector::at is checked operator [], though that might be obscure, and to me `find` implies more expense to me, like a linear or binary search. `get`? I'm ok either way.

I'm torn as to even use a function vs. direct use of mono_jit_icall_info.array.

The failure seems unrelated, but also specific to this PR. and not flaky.
Like a null pointer calling into loaded LLVM.
So I doubt I will go to master until I resolve that.

And yes I was torn about the `int` vs. `enum`.

`int` is convenient when reaching into a patch info, if I reuse `index. I'll try `enum`, and for patch_info, well, I could add the field, but then there is the read/write, which I guess I'll be pedantic about one way or the other -- like read/write a local int, and copy that to/from the enum. Maybe do a sizeof check to avoid the copy. Minor points though.

We will also have int or uint in the IL stream, in this PR. But ok can cast if needed.
 
In summary:
 - figure out the optional CI failure
 - rename the function to mono_find_jit_icall_info (or get)
 - change int to enum; cast if needed to make that work, hopefully not

mono/metadata/jit-icall-reg.h
jaykrell: Both done now. We'll see if C++ compiles. :)

vargaz:
@monojenkins build failed

jaykrell:
Failure is somehow related.  
I’ll try to repro locally.

jaykrell:
@monojenkins build failed

jaykrell:
The partial undo, or perhaps the retry, worked.
I will PR each little part of the undo I guess.
And maybe repro locally.

jaykrell:
I will go back before this series of PRs and double check this “llvm” vs. “mono_llvm” stuff.

I recall there is inconsistent naming but only really one set. 

I think we are ok to proceed. I hope CI exercises these paths. But I will look again.

jaykrell:
@monojenkins build failed

PR #14510, opened 2019-05-17
Close two race conditions.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/class-init.c
```diff
@@ -5594,18 +5591,14 @@ MonoClass *
...
-	static gboolean inited = FALSE;
 
-	if (!inited) {
-		klass.element_class = mono_defaults.int64_class;
-		klass.rank = 1;
-		klass.instance_size = MONO_SIZEOF_MONO_ARRAY;
-		klass.sizes.element_size = 8;
-		klass.size_inited = 1;
-		klass.name = "array_filler_type";
+	klass.element_class = mono_defaults.int64_class;
```
jaykrell: This pattern does bounce cache lines.
Alternative is to read/check/write each field, so that cache lines remain shared, not bouncing exclusive.
In this case the caller is doing on demand initialization also so it is ok.

jaykrell: C99 designated initializers seeming nice then. Only one of these is dynamic, and that could be addresses, perhaps, by adding an indirect bit.

PR #14511, opened 2019-05-17
Use ld -Bsymbolic when linking AOT.

jaykrell:
This keeps the dynamic loader a little out of the way, I.e. for intra .so calls, on some systems, e.g. Linux. On other systems it does nothing e.g. Apple and Windows -- they do this always. So this makes Linux more efficient, more like Apple and Windows.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -11897,10 +11897,10 @@ compile_asm (MonoAotCompile *acfg)
 
 #if defined(sparc)
 #define LD_NAME "ld"
-#define LD_OPTIONS "-shared -G"
+#define LD_OPTIONS "-shared -G -Bsymbolic"
 #elif defined(__ppc__) && defined(TARGET_MACH)
 #define LD_NAME "gcc"
-#define LD_OPTIONS "-dynamiclib"
+#define LD_OPTIONS "-dynamiclib -Wl,-Bsymbolic"
 #elif defined(TARGET_AMD64) && defined(TARGET_MACH)
 #define LD_NAME "clang"
 #define LD_OPTIONS "--shared"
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

alexanderkyte:
Looking into the implications of this option, it seems that there are some particularly difficult bugs that could be caused by using it:

```
https://software.intel.com/en-us/articles/performance-tools-for-software-developers-bsymbolic-can-cause-dangerous-side-effects
```

And yet, there's some evidence that it's already been widely used, and shouldn't be feared. 

https://stackoverflow.com/a/7295615 says:

```
Ubuntu uses it or used it by default, and it seems it causes conflicts with some packages. But overall it should work nicely I expect.
```

I wonder if we could maybe restrict the list of libraries this impacts, so that we don't risk doing something strange to third-party code linked against. That S.O. post mentions

```
--dynamic-list
```

which is documented as

```
    --dynamic-list=dynamic-list-file
           Specify the name of a dynamic list file to the linker.  This is
           typically used when creating shared libraries to specify a list of
           global symbols whose references shouldn't be bound to the
           definition within the shared library, or creating dynamically
           linked executables to specify a list of symbols which should be
           added to the symbol table in the executable.  This option is only
           meaningful on ELF platforms which support shared libraries.
``` .

Is this portable? Clang seems to use it with address sanitizer enough that whenever ldd breaks it, the sanitizer folks file a bug. Otherwise, it would be nice to have a "if you see this, it means this, and this option is to blame" in a comment somewhere. Otherwise this will be a good, long debugging session to diagnose.

jaykrell:
1. We are limiting it to what we link. libmonoruntime binds to libmonoruntime.
2. It makes it so you can't replace functions, which is usually right, rarely not. For example some mock/testing frameworks depends on the other way around. Apple works this way. Windows works this way. So it makes the various platforms more similar.

jaykrell:
Oh sorry that comment is off. It'd be System.dll.so binding to System.dll.so, System.Xml.so binding to System.Xml.dll.so, etc. and the user code binding to itself. I think the odds of the functions being usefully replacable by the dynamic linker are incredibly low. And it already works this way Apple and Microsoft systems.

jaykrell:
This has no impact on things we link against.
It only impacts the things we link ourselves.
It binds them to themselves, for symbols they provide.
It prevents "interposition" -- other .sos taking over symbols.
We don't put operator new in AOT output so that isn't relevant.

The default dynamic link model on Linux, is based on Solaris, is based on the static linker, where there is just one global pool of symbols, and the first one found is it.

This is generally not desired and has semantic and performance implications throughout.
Every non-file-level static symbol is subject to replacement at runtime, so must go through a pointer.
I imagine, but am not certain, every first-time call through non-file-level static involves some locked search, of all already loaded .sos.

For the most part, this strongly defies expectations.

It is sometimes useful. When you are going to load libfoo.so or libbar.so and you want some implementation from whichever loads.

On Windows this is usually much more explicit, via LoadLibrary/GetProcAddress (dlopen/dlsym).
But on Linux by default it is how all non-file-level-static symbols work.

It is also useful for mocking.
However there again, a common practice is to identify which symbols you actually might replace for testing, and have only them incur some optional indirection/slowdown.

The Linux defaults apply this broadly. Which really seems not what is generally intended or desirable.

-Bsymbolic inverts this. If you implement a function in an .so, and call it, then you get your implementation. No earlier load of an .so with that function can replace it.

jaykrell:
@monojenkins build failed

jaykrell:
Runtime already links this way. Change is to do it for AOT.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lambdageek:
yolo

Let's try it, I guess?  If it breaks the `2019-10` integration on XI or something else unexpected, we can revert.

PR #14513, opened 2019-05-17
[netcore] Merge all *.rsp files into CoreFX.issues.rsp

EgorBo:
To make it CI friendly.

E.g. CoreCLR's [CoreFX.issues.rsp](https://github.com/dotnet/coreclr/blob/master/tests/CoreFX/CoreFX.issues.rsp).

Also ignore all failing tests in all test assemblies (`make xtestall`)

/cc: @filipnavara


netcore/Makefile
```diff
@@ -129,48 +129,15 @@ xtest-%: prepare update-corefx-tests
 		-notrait category=outerloop -notrait category=nonosxtests -notrait category=failing \
 		-notrait category=nonnetcoreapptests -notrait category=nonmonotests \
 		-html ../../TestResult-$*.html -nunit ../../TestResult-$*-netcore-xunit.xml \
-		$(shell if [ -a $(CURDIR)/excludes-$*.rsp ]; then grep -v '^#\|^$$' $(CURDIR)/excludes-$*.rsp; fi;) \
+		$(shell if [ -a $(CURDIR)/CoreFX.issues.rsp ]; then grep -v '^#\|^$$' $(CURDIR)/CoreFX.issues.rsp; fi;) \
```
filipnavara: Can we specify it as response file now (`@CoreFX.issues.rsp`)? I don't know if the fix for non-lower-case file names was flowed back to Mono yet, but it may have been. Alternatively we can just name it `corefx.issues.rsp` for now.

netcore/Makefile
```diff
@@ -129,48 +129,15 @@ xtest-%: prepare update-corefx-tests
...
-	System.Runtime.Loader.Tests \
-	System.Runtime.Serialization.Formatters.Tests \
-	System.Threading.Overlapped.Tests \
-	System.Xml.Xsl.XslTransformApi.Tests
+EXCLUDED_COREFX_TESTS =
 
 xtestall: update-corefx-tests $(foreach workingtest, $(foreach test, $(wildcard corefx/tests/extracted/*), \
 		$(filter-out $(EXCLUDED_COREFX_TESTS), $(notdir $(test)))), $(addprefix xtest-, $(workingtest)))
 
-xtestpass: update-corefx-tests $(foreach workingtest, $(foreach test, $(wildcard corefx/tests/extracted/*), \
```
filipnavara: Are we abandoning `xtestpass` altogether now? The main reason for it was to maintain a list for the CI. I know it was never deployed, but the general idea was to have a known subset with 0 failures. If `xtestall` can fill that role I am fine with it.

netcore/Makefile
EgorBo: @filipnavara the idea is to move to the `RunScript.sh`. I am not sure if I can lowercase that file, don't remember where it's set on CI. @ViktorHofer can we have a custom name for it?

netcore/Makefile
EgorBo: @filipnavara yeah, I removed it because xtestall now should have 0 failures.

netcore/Makefile
filipnavara: I see. The properly cased file name may work now. We should test it, and perhaps bump CoreFX version if necessary.

EgorBo:
@monojenkins build failed

EgorBo:
@monojenkins build failed

PR #14517, opened 2019-05-17
prototype amd64 only genmdesc.py removal

jaykrell:


akoeplinger:

PR #14518, opened 2019-05-17
Make copy_stack_data support configurable.

lateralusX:
By default when called through macros like MONO_ENTER_GC_UNSAFE code will always dump current registers into stack space on hybrid/coop. The registers will be scanned but only when hitting a GC and when running in GC unsafe, that will only happen when we block in the enter or hit a safe point at a later point in time. There are currently a couple of frequent scenarios where we go into GC safe when thread won't be suspended and in that case the copy of the registers will only reflect the state when
entering GC unsafe.

Commits add option to enable/disable copy_stack_data support in configure step:

--enable-copy-stack-data, --disable-copy-stack-data.

copy_stack_data is enabled by default on all platforms except desktop platforms macos, linux and windows (both 32/64-bit), since these platforms already support full context as well as capabilities to scan the stack of a running thread.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1310,6 +1310,7 @@ with_wasm_default=no
 with_bitcode_default=no
 enable_cooperative_suspend_default=no
 enable_hybrid_suspend_default=no
+enable_copy_stack_data_default=yes
 
 # For the sake of clearer error messages, these numbers should all be different from each other.
 INVARIANT_AOT_OPTIONS=nimt-trampolines=2000,ntrampolines=10000,nrgctx-fetch-trampolines=256,ngsharedvt-trampolines=4400,nftnptr-arg-trampolines=4000
@@ -5626,6 +5627,38 @@ fi
 AM_CONDITIONAL([ENABLE_HYBRID_SUSPEND], [test x$enable_hybrid_suspend != xno])
 
 dnl End of thread suspend policy
+
+dnl *******************************
+dnl *** Copy stack data support ***
+dnl *******************************
+
+dnl Disable copy_stack_data on platforms supporting full context and
+dnl having access to full stack of running threads.
+if test x$enable_copy_stack_data_default != xno; then
+	case $HOST in
+	X86 | AMD64)
+		if test x$target_osx = xyes; then
+			dnl Some host/target confusion, there's no host_osx (and
+			dnl host_darwin would be true on iOS not just macOS).
+			enable_copy_stack_data_default=no
+		elif test x$host_linux = xyes -o x$host_win32 = xyes; then
+			enable_copy_stack_data_default=no
+		fi
+		;;
+	esac
+fi
+
+AC_ARG_ENABLE(copy_stack_data, [ --enable-copy-stack-data     Enable copy_stack_data feature for hybrid/cooperative suspend scenarios, (defaults to yes)], [], [enable_copy_stack_data=default])
+
+if test x$enable_copy_stack_data = xdefault; then
+   enable_copy_stack_data=$enable_copy_stack_data_default
+fi
+
+if test x$enable_copy_stack_data != xno ; then
+	AC_DEFINE(ENABLE_COPY_STACK_DATA,1,[Enable copy_stack_data feature for hybrid/cooperative suspend scenarios])
+fi
+
+dnl End of Copy stack data support
 
 dnl ***************************
 dnl *** feature experiments ***
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -34,6 +34,12 @@
 #include <mono/utils/mach-support.h>
 #endif
 
+/* On platforms that doesn't have full context support (or doesn't do conservative stack scan), use copy stack data */
+/* when entering safe/unsafe GC regions. For platforms with full context support (doing conservative stack scan), */
+/* there is already logic in place to take context before getting in a state where thread could be conservative */
+/* scanned by GC. Avoiding doing additional stack copy will increse performance when entering safe/unsafe regions */
+/* when running in hybrid/cooperative supspend mode. */
+#if defined (ENABLE_COPY_STACK_DATA)
 #ifdef _MSC_VER
 // __builtin_unwind_init not available under MSVC but equivalent implementation is done using
 // copy_stack_data_internal_win32_wrapper.
@@ -43,6 +49,9 @@
 #define SAVE_REGS_ON_STACK do {} while (0)
 #else 
 #define SAVE_REGS_ON_STACK __builtin_unwind_init ();
+#endif
+#else
+#define SAVE_REGS_ON_STACK do {} while (0)
 #endif
 
 volatile size_t mono_polling_required;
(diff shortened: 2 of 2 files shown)


lateralusX:
@monojenkins rebuild failed

jaykrell:
Please remind me why we ever have to save context upon function all? 
I would imagine you can get it upon preemptive suspend or store it upon cooperative suspend. Why proactively save it way ahead of potential suspend?

jaykrell:
Also, assuming pointers aren’t in floating point registers (?) amd64 nonvolatile integer context should be around 64 bytes, right? But rather than shrink it or obsess over floating point, why/when save it, except when actually suspending?

lateralusX:
So this is what's unclear to me as well and what I'm experimenting with in this PR. It is hard to understand why it is needed for all scenarios where we currently do it. On platforms where we can get context, we will get it in thread_state_init just before suspending (both for preemptive/hybrid/cooperate), on platforms where we can't get context, copy_stack_data is what will make sure we get it stored for later GC scanning. But question is why it is take as part of GC safe/unsafe transitions instead of when we actually hit safe points since that is when a GC can happen. There is the hybrid suspend case that is also in play, but in cases where we do hybrid suspend we most likely have support to get thread context as part of thread_stat_init.

Include of xmm registers in copy_stack_data is done since it does the same as corresponding gcc intrinsic that dumps all nonvolatile regs. We also had problems with this for preemptive suspend in the past due to memcpy implementations using xmm, but for hybrid/cooperative where we take context on specific known locations and I agree we could probably optimize to take a smaller snapshot on Windows (on System V ABI xmm regs are all volatile) where we control exactly what we store instead of being dependent on an intrinsic to do it. That will reduce the size of copied data on each call to copy_stack_data.

So on platforms where we can both take thread context as part of suspending (preemptive/hybrid/coop) + scanning stack conservative I don't see why it is needed (potentially just a left over from initial implementation that only targeted platforms lacking that support). For scenarios where we can't get context and try to do preciese stack scanning it could make sense to do it on transistions into unsafe/safe or we won't see references in registers, but again, in safe mode, code should not work with unpinned managed memory and in unsafe mode (where we can't do conservative stack scan) we should keep track of all references used on stack.

Looking at the history, copy_stack_data seems to been added when doing initial cooperate suspend implementation for Watch OS, so back then I believe hybrid suspend was not a use case, but hybrid suspend adds into the mix for safe mode, espacialy for compability with old embedding API, since it breaks the rules of safe mode only using pinned managed memory (code using embedding API is in safe mode, but due to the design of embedding API, use unpinned managed memory). Maybe this was also true for platforms using cooperative suspend + embedding API since it will have the same potential problems. Back then only these platforms used the cooperative support, meaning using copy_stack_data and that platform doesn't have context support or using conservative stack scanning. When expanding the use of the cooperative support (and hybrid) to other platforms that still does both context and conservative stack scanning the copy_stack_data implementation was just kept as is, getting us into this discussion.

Maybe the solution is to exclude the call to copy_stack_data on platforms that already have full context support and platforms always doing conservative stack scan (making sure we see all managed references on stack for threads using embedding API). Currently WASM doesn't support it at all, but all other platforms does something.

@lambdageek, anything to add around the history of copy_stack_data? Do you believe it could make sense to exclude use of copy_stack_data on platforms that already have full context support + doing conservaive stack scanning? Question then is what platforms that we should make copy_stack_data into a noop, all except WASM (already working that way), Watch OS, TV OS, or just do it for desktop platforms for now (Windows, OSX, Linux)?

lateralusX:
@monojenkins rebuild failed

lateralusX:
@lambdageek, OK?

mono/utils/mono-threads-coop.c
```diff
@@ -16,6 +16,15 @@
 #define _DARWIN_C_SOURCE
 #endif
 
+/* On platforms that doesn't have full context support (or doesn't do conservative stack scan), use copy stack data */
+/* when entering safe/unsafe GC regions. For platforms with full context support (doing conservative stack scan), */
+/* there is already logic in place to take context before getting in a state where thread could be conservative */
+/* scanned by GC. Avoiding doing additional stack copy will increse performance when entering safe/unsafe regions */
+/* when running in hybrid/cooperative supspend mode. */
+#if defined (HOST_IOS) || defined (HOST_WATCHOS) || defined (HOST_TVOS) || defined (HOST_ANDROID) || defined (HOST_WASM)
```
lambdageek: I'd like also `!defined (USE_COPY_STACK_DATA) &&` and then we should have a linux CI lane (perhaps the coop suspend lane) with `-DUSE_COPY_STACK_DATA` so that we get some coverage of the implementation.

mono/utils/mono-threads-coop.c
```diff
@@ -16,6 +16,15 @@
 #define _DARWIN_C_SOURCE
 #endif
 
+/* On platforms that doesn't have full context support (or doesn't do conservative stack scan), use copy stack data */
+/* when entering safe/unsafe GC regions. For platforms with full context support (doing conservative stack scan), */
+/* there is already logic in place to take context before getting in a state where thread could be conservative */
```
lambdageek: This is talking about `thread_state_init (&info->thread_saved_state [SELF_SUSPEND_STATE_INDEX])`?

lambdageek:
> anything to add around the history of copy_stack_data? 

Not much - it took me a very long time to understand this part of the code, so at the time when `copy_stack_data` was first implemented, I didn't really understand what was happening.

mono/utils/mono-threads-coop.c
lateralusX: Yes, as part of entering safe mode we will take threads context in mono_threads_enter_gc_safe_region_unbalanced_with_info.

mono/utils/mono-threads-coop.c
lateralusX: I could add that into the condition above and then set CFLAGS in run-jenkins.sh for selected CI configuration (like linux coop suspend CI lanes 32/64-bit).

lateralusX:
Changed the solution into a configuration step where currently all platforms except MacOS, Linux and Windows 32/64-bit will still be using copy_stack_data as before. NOTE, default behavior is only changed on desktop platforms. Since it is now a configuration option is will be easier to enable/disable the support going forward when/if needed.

configure.ac
```diff
@@ -5627,6 +5628,38 @@ AM_CONDITIONAL([ENABLE_HYBRID_SUSPEND], [test x$enable_hybrid_suspend != xno])
...
 
+dnl *******************************
+dnl *** Copy stack data support ***
+dnl *******************************
+
+dnl Disable copy_stack_data on platforms supporting full context and
+dnl having access to full stack of running threads.
+if test x$enable_copy_stack_data_default != xno; then
+	case $HOST in
+	X86 | AMD64)
```
jaykrell: It is more of Linux/Windows/Android/Wasm thing, or the combinations, eh?

lateralusX: This follows the same platforms as default for hybrid suspend. For this case copy stack data will be disabled by default for x86/amd64 linux, osx or windows hosts (note shouldn't include Android, WASM or IOS platforms). It always possible to force it using --enable-copy-stack-data configure parameter.

jaykrell: Android runs on x86 too. That is the thing.
Windows runs on arm and arm64, etc. (albeit Mono hasn't been ported to them).
It is confusing processor architecture for other things.

lateralusX: Windows is fine since we only have x86, amd64 ports at the moment and if we get more there are things that needs to be updated elsewhere. Android shouldn't have $host_linux set as part if its configuration step, so this should not trigger on Android.

PR #14524, opened 2019-05-18
Unless --try-llvm, make llvm_init failure fatal earlier/clearer instea…

jaykrell:
…d of later jumping to 0.

i.e. MONO_LLVM_INTERNAL and the CI Linux x64 - prefix sanity checks/check-prefix-llvmaot.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -13404,11 +13404,8 @@ mono_compile_assembly (MonoAssembly *ass, guint32 opts, const char *aot_options,
 		acfg->flags = (MonoAotFileFlags)(acfg->flags | MONO_AOT_FILE_FLAG_DEBUG);
 	}
 
-	if (acfg->aot_opts.try_llvm)
-		acfg->aot_opts.llvm = mini_llvm_init ();
+	if (mono_use_llvm || acfg->aot_opts.llvm || acfg->aot_opts.try_llvm) {
 
-	if (mono_use_llvm || acfg->aot_opts.llvm) {
-		acfg->llvm = TRUE;
 		acfg->aot_opts.asm_writer = TRUE;
 		acfg->flags = (MonoAotFileFlags)(acfg->flags | MONO_AOT_FILE_FLAG_WITH_LLVM);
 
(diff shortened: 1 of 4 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This should improve those CI “prefix” breaks that we saw yesterday & today.
And be correct anyway.

jaykrell:
It should print right before that.
In CI we see errors when we fail to mark functions as MONO_LLVM_INTERNAL.
That is what this is intended to diagnose a bit differently.
Otherwise, it is a warning and a later jmp to 0.
This should be tested with fault injection, granted.
I can do that within this PR or as a separate PR.

jaykrell:
Not very convincing:

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-prefix/1371/parsed_console/log.html

${TESTCMD} --label=check-prefix-llvmjit --timeout=1m mono --llvm ${MONO_REPO_ROOT}/tmp/hello.exe

```
mono: symbol lookup error: /mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/tmp/monoprefix/lib/libmono-llvm.so: undefined symbol: mono_get_jit_icall_info
/mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/scripts/ci/babysitter: Test suite terminated with code 127, and suite cannot report test case data. Halting.
```

Let me iterate a bit more..

jaykrell:
Ok, so debugging the injected fault, the load succeeds. Because, like, the file exists.

The unfound symbol isn't found till quite late, till it is used.

I suggest we should be using ld -z now, which will make this fail earlier.
I'll try that.

jaykrell:
Close and reopen to reset approval since it changed a lot.

jaykrell:
Previous approval invalidated since I debugged more.
The failures in question were JIT, not AOT, and there was no --try-llv switch for JIT.
Furthermore, without -z now/bind_at_load, which is another PR, dlopen succeeds anyway.

jaykrell:
This should be dismissed and revisited based on churn.

jaykrell:
I'm not sure I understand.
I understand there is no testing here of --try-llvm, but I not sure what a test should look like.
We can duplicate all/some --llvm tests to --try-llvm, but it is difficult to gauge their result.
We could see if they work, which could be for multiple reasons.
We could see if, given a bad LLVM path, they still work.
We could try to measure their perf, i.e. see if they are using LLVM or fell back or what.

I believe there is already AOT --try-llvm also not tested

imhameed:
Loaded LLVM support was removed in https://github.com/mono/mono/pull/18868.

PR #14529, opened 2019-05-20
Fix warning: unused variable has_custom_args in aot-compiler.c:arch_init

jaykrell:


mono/mini/aot-compiler.c
```diff
@@ -1103,6 +1102,10 @@ arch_init (MonoAotCompile *acfg)
 	acfg->llvm_label_prefix = "";
 	acfg->user_symbol_prefix = "";
 
+#if TARGET_X86 || TARGET_AMD64
```
lewurm: we usually write
```c
#if defined(TARGET_X86) || defined(TARGET_AMD64)
```

jaykrell: Maybe start using a newer terser equivalent form?
They are defined to be 1 and undefined values evaluate to 0.
In future hopefully `#if` changed to `if`, so we have multi-targeting AOT compiler.

lewurm: not disagreeing with you, but I prefer consistency across the code base

jaykrell:
@monojenkins build failed

PR #14532, opened 2019-05-20
Fix race condition and use less memory in mono_lookup_icall_symbol.

jaykrell:
Albeit a little slower due to indirection.
Also fix the chance of overflow in the comparison of pointers.
Subtraction is not safe.

### DIFF
--- a/mono/metadata/icall-table.c
+++ b/mono/metadata/icall-table.c
@@ -44,6 +44,8 @@
 #define HANDLES_REUSE_WRAPPER		HANDLES
 #define MONO_HANDLE_REGISTER_ICALL(...) /* nothing  */
 
+//#define TEST_ICALL_SYMBOL_MAP 1
+
 // Generate Icall_ constants
 enum {
 #define ICALL_TYPE(id,name,first)	/* nothing */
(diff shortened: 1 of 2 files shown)


jaykrell:
@monojenkins build failed

mono/metadata/icall-table.c
```diff
@@ -253,57 +253,66 @@ icall_table_lookup (MonoMethod *method, char *classname, char *methodname, char
...
-					symbols_sorted [i] = symbols_sorted [i + 1];
-					symbols_sorted [i + 1] = (const char*)tmp;
-					changed = TRUE;
-				}
-			}
-		}
-		inited = TRUE;
+#if ENABLE_ICALL_SYMBOL_MAP || TEST_ICALL_SYMBOL_MAP
+	typedef guint16 T;
+	const T N = G_N_ELEMENTS (icall_functions) - 1; // skip terminal null element
```
jaykrell: N should be of type gsize for the assert to work, without truncation.

mono/metadata/icall-table.c
```diff
@@ -44,6 +44,8 @@
 #define HANDLES_REUSE_WRAPPER		HANDLES
 #define MONO_HANDLE_REGISTER_ICALL(...) /* nothing  */
 
+//#define TEST_ICALL_SYMBOL_MAP 1
+
 // Generate Icall_ constants
```
baulig: Debugging left-over?

baulig:
I looked at this, but unfortunately don't understand the code.

jaykrell:
It is racy because it is doing on-demand initialization, which is often racy.

The rewrite changes the result to be one pointer, atomically swapped, which implies a full barrier and data dependency, so no race.

It could also be fixed by copying and sorting an array of pairs, and one pointer to that.

However in order to save memory I instead use indirect data that is an array of uint16 pointing into the original data. That is a slight memory vs. time tradeoff.

Yes there is debugging code left, under #if, that I prefer to leave, both as evidence that I tested it, and to make it somewhat but ideally testable in future.

The code also was using an unnecessary somewhat risky but probably ok here way to compare pointers, that I fixed.

As well the result is marginally smaller and faster because it was inlining something like bubblesort, now reuses qsort.

mono/metadata/icall-table.c
jaykrell: Yes but I prefer to leave it.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Will rebase to master since this sat a while.

akoeplinger:
@monojenkins build failed

jaykrell:
Been two weeks, rebase again?

jaykrell:
Code is pretty isolated so rebase probably not warranted.
Not even enabled in most builds.

PR #14533, opened 2019-05-20
Fix leak in mono_parse_options

Therzok:
When MONO_ENV_OPTIONS is null, we would leak the array and buffer


lewurm:
@monojenkins squash

vargaz:
@monojenkins build failed

Therzok:
@monojenkins build failed

Issue #14491, opened 2019-05-16
Add an easy way to print the version number by itself


aaronfranke:
### Current Behavior

If you type `mono --version`, you get an output that looks like this:

```
$ mono --version
Mono JIT compiler version 5.20.1.19 (tarball Thu Apr 11 09:02:17 UTC 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

### Expected Behavior

I would like the ability to do something like this:

```
$ mono --version-number
5.20.1.19
```

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

5.20.1.19

lewurm:
I like this idea, however keep in mind that it will take some while until such an option is useful in the real world, as older mono versions do not understand this flag.

https://github.com/mono/mono/pull/14538

PR #14538, opened 2019-05-20
[driver] add --version=number

lewurm:
Useful for third parties consuming the version number of mono.

Fixes https://github.com/mono/mono/issues/14491

### DIFF
--- a/man/mono.1
+++ b/man/mono.1
@@ -659,6 +659,9 @@ mscorlib and assemblies located on the global assembly cache.
 \fB-V\fR, \fB--version\fR
 Prints JIT version information (system configuration, release number
 and branch names if available). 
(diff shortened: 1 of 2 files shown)


marek-safar:
Please add documentation to the man page

akoeplinger:
How about we do `--version=number`? so we can add e.g. `--version=commit` etc. later on?

lewurm:
@akoeplinger good idea, updated PR.

aaronfranke:
Good idea with `--version=number`, the `--version-number` in my post was just an example.

In addition to `commit`, perhaps we could also have `--version=major` and `minor` etc to get only `5` or `20` from a version string such as `5.20.1.19`.

Issue #14469, opened 2019-05-13
mini-posix.c: result of fork syscall not checked, and Mono segfaults


Xyene:
## Steps to Reproduce

1. Patch fork to return `EAGAIN`, or set `RLIMIT_NPROC` appropriately.
2. Cause a crash (e.g. like the one in #14416)

### Current Behavior

`fork` returns -1, which is not checked but stored in `pid`. Mono then calls `waitpid(pid, ...)` in https://github.com/mono/mono/blob/e7011c780f676914f559f14f25e76c192bb2b0b2/mono/mini/mini-posix.c#L1111

This [has the effect](https://linux.die.net/man/2/waitpid) of "wait[ing] for any child process whose process group ID is equal to the absolute value of _pid_" &mdash; in this case, 1.

Mono then crashes with a SIGSEGV.

### Expected Behavior

Mono attempts to check the return value of `fork`, and gracefully falls back.

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 5.16.0.179 (tarball Thu Oct  4 10:24:32 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(3.6.0svn-mono-/)
	GC:            sgen (concurrent by default)
```

The relevant code is present in master, so I would assume the behaviour is the same there.

## Stacktrace

```
/proc/self/maps:
40000000-40020000 rwxp 00000000 00:00 0 
15554a3ac000-15554a3b0000 rw-p 00000000 00:00 0 
15554a3b4000-15554a3b8000 rw-p 00000000 00:00 0 
15554a3bc000-15554a3c0000 rw-p 00000000 00:00 0 
15554a3c4000-15554a3c8000 rw-p 00000000 00:00 0 
15554a3cc000-15554a3d0000 rw-p 00000000 00:00 0 
15554a3d4000-15554a3d8000 rw-p 00000000 00:00 0 
15554a3dc000-15554a3e0000 rw-p 00000000 00:00 0 
15554a3e4000-15554a3e8000 rw-p 00000000 00:00 0 
15554a3ec000-15554a3f0000 rw-p 00000000 00:00 0 
[...]

PR #14539, opened 2019-05-20
[crash] Don't waitpid() if fork() fails

alexanderkyte:
Fixes https://github.com/mono/mono/issues/14469

This is necessary to make Mono crash reporting resilient in the face of forking failures.


alexanderkyte:
@monojenkins build failed

Issue #14416, opened 2019-05-08
Mono native crash during startup does not exit with a nonzero code


Xyene:
## Steps to Reproduce

1. Create `Test.cs` with the following content:
```c#
using System;

namespace App
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello, World!");
        }
    }
}
```

2. Compile with `mcs Test.cs` to produce `Test.exe`.
3. `ulimit -Sv 28000`. Might need to vary the exact number to produce the erroneous behaviour.
4. Run `mono Test.exe`. This should cause the runtime to fail during initialization.

### Current Behavior
```
xyene@iorwerth:/code$ mono Test.exe
Could not allocate 8192 bytes
=================================================================
        Native Crash Reporting
=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================
/proc/self/maps:
4030f000-4031f000 rwxp 00000000 00:00 0
40e91000-40ea1000 rwxp 00000000 00:00 0
55a981869000-55a981cc1000 r-xp 00000000 08:00 16535                      /usr/bin/mono-sgen
55a981ec1000-55a981ec7000 r--p 00458000 08:00 16535                      /usr/bin/mono-sgen
55a981ec7000-55a981ecd000 rw-p 0045e000 08:00 16535                      /usr/bin/mono-sgen
55a981ecd000-55a981ee4000 rw-p 00000000 00:00 0
55a983013000-55a9830d3000 rw-p 00000000 00:00 0                          [heap]
7fd29a355000-7fd29a5c9000 r--p 00000000 08:00 619998                     /usr/lib/mono/gac/System/4.0.0.0__b77a5c561934e089/System.dll
7fd29a5c9000-7fd29a5ca000 ---p 00000000 00:00 0
7fd29a5ca000-7fd29a5cb000 rw-p 00000000 00:00 0
7fd29a5cb000-7fd29a5d3000 ---p 00000000 00:00 0
7fd29a5d3000-7fd29a7ca000 rw-p 00000000 00:00 0
7fd29a7ca000-7fd29ad7d000 r-xp 00000000 08:00 619838                     /usr/lib/mono/aot-cache/amd64/mscorlib.dll.so
7fd29ad7d000-7fd29af7c000 ---p 005b3000 08:00 619838                     /usr/lib/mono/aot-cache/amd64/mscorlib.dll.so
[...]

PR #14540, opened 2019-05-20
[crash] Handle OOM/blocked allocation during crash dumper

alexanderkyte:
The issues fixed below were identified by https://github.com/mono/mono/issues/14416 .

The typo is a cosmetic fix.

The other change uses g_error and not g_assert_not_reached. g_error goes through the already-modified path that will work correctly during double faults during the crash dumper. Otherwise, the double fault results in a zero/successful status code.

The reproduction for this involves preventing mono from creating any more allocations, using `ulimit`.


lewurm:
@monojenkins build failed

PR #14549, opened 2019-05-21
Add ability to use cooperative suspend in configure step.

lateralusX:
Needed in order to setup CI lanes using --enable-cooperative-suspend.

### DIFF
--- a/winconfig.h
+++ b/winconfig.h
@@ -35,10 +35,9 @@
 /* @ENABLE_DEFINES@ */
 /* End configure ENABLE_DEFINES picked up from cygconfig.h or other external source, if available */
 
-#if defined(ENABLE_HYBRID_SUSPEND) || defined(ENABLE_COOP_SUSPEND)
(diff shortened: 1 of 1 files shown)


winconfig.h
```diff
@@ -35,10 +35,9 @@
 /* @ENABLE_DEFINES@ */
 /* End configure ENABLE_DEFINES picked up from cygconfig.h or other external source, if available */
 
-#if defined(ENABLE_HYBRID_SUSPEND) || defined(ENABLE_COOP_SUSPEND)
-/* Windows MSVC builds defaults to preemptive suspend. Disable ENABLE_HYBRID_SUSPEND/ENABLE_COOP_SUSPEND defines. */
+#if defined(ENABLE_HYBRID_SUSPEND)
```
jaykrell: You don’t need ifdef foo before undef foo.
You can just undef foo unconditionally.

PR #14557, opened 2019-05-21
[netcore] Fix PointerTests

EgorBo:
This PR enables 87 tests in `System.Runtime.Tests`.

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -3374,7 +3374,11 @@ mono_field_set_value_internal (MonoObject *obj, MonoClassField *field, void *val
 		return;
 
 	dest = (char*)obj + field->offset;
+#if ENABLE_NETCORE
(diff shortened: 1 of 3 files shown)


mono/metadata/object.c
```diff
@@ -3374,7 +3374,11 @@ mono_field_set_value_internal (MonoObject *obj, MonoClassField *field, void *val
 		return;
 
 	dest = (char*)obj + field->offset;
+#if ENABLE_NETCORE
+	mono_copy_value (field->type, dest, value, value && field->type->type == MONO_TYPE_PTR);
```
EgorBo: the last argument is `int deref_pointer`

filipnavara:
For future reference, I am pretty sure this should be enabled on non-netcore builds too. The Mono test is incorrectly ignoring it because the two changed places cancel each other out.

PR #14564, opened 2019-05-22
[PowerPC] Fix compile error.

jaykrell:


akoeplinger:
Merging since we don't have this config on PR checks anyway and the fix is clear :)

jaykrell:

PR #14567, opened 2019-05-22
[loaded LLVM] mono_gchandle_get_target_internal needs MONO_LLVM_INTERNAL.

jaykrell:
Found by ld -z now
 https://github.com/mono/mono/pull/14562
 https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-prefix/1398/parsed_console/log.html

 instruction pointer is NULL, skip dumping/mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/scripts/ci/babysitter: Test suite terminated with code -6, and suite cannot report test case data. Halting.
 *** start: check-prefix-llvmjit
 llvm load failed: /mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/tmp/monoprefix/lib/libmono-llvm.so: undefined symbol: mono_gchandle_get_target_internal

Mono Warning: llvm support could not be loaded.

There could be more of these, I don't know if it finds them all at once or just one at a time.

PR #14568, opened 2019-05-22
[llvm] Remove unused default_mono_llvm_unhandled_exception.

jaykrell:
It references mono_gchandle_get_target_internal which is not marked MONO_LLVM_INTERNAL
so attempts to run it will fail.

This is caught by linking wit -z now (https://github.com/mono/mono/pull/14562).

Alternative to https://github.com/mono/mono/pull/14567.


jaykrell:
https://github.com/mono/mono/pull/14589 instead.

Issue #14585, opened 2019-05-22
[tests] Don't send HTTP requests to Google


lambdageek:
See for example https://github.com/mono/mono/pull/14541#issuecomment-494993168

The tests `HttpWebRequestTest.Sync` and `WebClientTest.DownloadTwice` send requests to google.com and if we run enough tests simultaneously, Google starts responding with `(429) Too Many Requests`.

It would be better if we could identify some friendly site for testing (or startup a local server) so that we can rely on a predictable response.

baulig:
The same pattern probably exists in far more tests.  We could for instance block all outgoing network traffic on one bot then run the tests to see which.

PR #14586, opened 2019-05-23
[bcl] Remove usages of mono-project.com/go-mono.com/google.com in tests

akoeplinger:
Use example.com and example.org instead.

Fixes https://github.com/mono/mono/issues/14585

Note that the majority of the things I replaced don't make an actual network request but I thought it'd be nice to be consistent.


marek-safar:
@monojenkins build failed

PR #14595, opened 2019-05-23
[tests] Add xunit.execution.dotnet.dll to tests directory

akoeplinger:
So it gets included in the sdks archive


marek-safar:
I am wondering does this 4.0 version work with 2.1 assemblies?

akoeplinger:
@marek-safar it's actually the netstandard version and we already used it for a while (just picking from the mono submodule in XI) so yes, it does work.

PR #14598, opened 2019-05-23
[llvm] Use CoreCLR GC interop configuration

alexanderkyte:
Since they're identical right now, we can use the coreclr configuration
to make our IR more easily analyzed by tools built around LLVM.

```
class CoreCLRGC : public GCStrategy {
public:
  CoreCLRGC() {
    UseStatepoints = true;
    // These options are all gc.root specific, we specify them so that the
    // gc.root lowering code doesn't run.
    InitRoots = false;
    NeededSafePoints = 0;
    UsesMetadata = false;
    CustomRoots = false;
  }

  Optional<bool> isGCManagedPointer(const Type *Ty) const override {
    // Method is only valid on pointer typed values.
    const PointerType *PT = cast<PointerType>(Ty);
    // We pick addrspace(1) as our GC managed heap.
    return (1 == PT->getAddressSpace());
  }
};

/// A GCStrategy for the Mono Runtime.
class MonoGC : public GCStrategy {
public:
  MonoGC() {
    UseStatepoints = true;
    // These options are all gc.root specific, we specify them so that the
    // gc.root lowering code doesn't run.
    InitRoots = false;
    NeededSafePoints = 0;
    UsesMetadata = false;
    CustomRoots = false;
  }
};
```

Source: https://github.com/mono/llvm/blob/64c0343537016c153894dc60316bd7b316b812c8/lib/CodeGen/BuiltinGCs.cpp#L105-L137

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -7732,7 +7732,7 @@ emit_method_inner (EmitContext *ctx)
 	}
 
 	if (!cfg->llvm_only && cfg->compile_aot && mono_threads_are_safepoints_enabled () && requires_safepoint)
-		LLVMSetGC (method, "mono");
+		LLVMSetGC (method, "coreclr");
 	LLVMSetLinkage (method, LLVMPrivateLinkage);
 
 	mono_llvm_add_func_attr (method, LLVM_ATTR_UW_TABLE);


jaykrell:
So much for not having statics with constructors? :(

alexanderkyte:
@jaykrell I'm not sure I follow. This is a configuration change that results in no functional change in the emitted IR, but simplifies our configuration a bit and makes our IR back into legal LLVM IR for vanilla LLVM.

alexanderkyte:
We forked LLVM and added a new configuration "based on" the coreclr one. In practice, our configuration is identical. This change takes advantage of it being identical.

jaykrell:
Sorry, I was referring to the nearby code, not you change. Hypothetically LLVM has no static constructors, and is cheap to load, but we have loaded LLVM because this isn't true. So looking at your change got me to see some static constructors.

lewurm:
@monojenkins squash

PR #14607, opened 2019-05-23
[build] Provide in-tree llvm to mono correctly

alexanderkyte:
Without this, the Mono build won't actually succeed the first time on a fresh run. The wrong opt/llc are used. The builder has to stop and export the directory here to make it work. This invalidates our instructions we've always been giving about building with the one-liner of autogen and make.

This is sidestepped on linux in practice with a patch: 

https://github.com/mono/linux-packaging-mono/blob/centos/llvm_llc_opt_default_path.patch

But I'd like to fix the OSX build and remove the need for a patch in order to build Mono with LLVM out of the box.

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -13281,7 +13281,8 @@ mono_compile_assembly (MonoAssembly *ass, guint32 opts, const char *aot_options,
 	acfg->aot_opts.nftnptr_arg_trampolines = 128;
 #endif
 	acfg->aot_opts.nunbox_arbitrary_trampolines = 256;
-	acfg->aot_opts.llvm_path = g_strdup ("");
+	if (!acfg->aot_opts.llvm_path)
+		acfg->aot_opts.llvm_path = g_strdup ("");
 	acfg->aot_opts.temp_path = g_strdup ("");
 #ifdef MONOTOUCH
 	acfg->aot_opts.use_trampolines_page = TRUE;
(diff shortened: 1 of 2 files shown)


mono/mini/aot-compiler.c
```diff
@@ -13281,7 +13281,8 @@ mono_compile_assembly (MonoAssembly *ass, guint32 opts, const char *aot_options,
 	acfg->aot_opts.nftnptr_arg_trampolines = 128;
 #endif
 	acfg->aot_opts.nunbox_arbitrary_trampolines = 256;
-	acfg->aot_opts.llvm_path = g_strdup ("");
+	if (!acfg->aot_opts.llvm_path)
+		acfg->aot_opts.llvm_path = g_strdup ("");
```
lewurm: so I assume `$MONO_LLVM_PATH_OPTION` should eventually end up in `acfg->aot_opts.llvm_path`. Where does that happen?

alexanderkyte: In this commit, we put it in the AOT_BUILD_FLAGS. Look at configure.ac.

https://github.com/mono/mono/pull/14607/files#diff-67e997bcfdac55191033d57a16d1408aR6633

alexanderkyte: It is the only way to conditionally provide it to in-tree invocations without including it in the shipped artifacts at all.

alexanderkyte:
@lewurm this is a log of how it works, complete with llvm invocations logged

https://gist.github.com/alexanderkyte/940f1188e6407479475a957794f62ea7

alexanderkyte:
Looks like @lateralusX caught that the Windows LLVM build wasn't covered by this. I didn't think we had in-tree Windows LLVM.

lateralusX:
https://github.com/mono/mono/pull/14694 should fix build error on Windows MSVC LLVM in-tree build.

Issue #14377, opened 2019-05-07
Document/fix functions currently not doing GC safe enter/exit around blocking OS API calls.


lateralusX:
During work porting hybrid/coop suspend over to Windows I looked over the GC safe enter/exit used through Mono runtime code. I found a couple of places where we don't do it, and after some further investigation I believed that most seemed fine (at least the same behavior on Windows/Unix), but we miss documentation around these calls why they are not using GC enter/exit safe. Below is just a dump of methods that doesn't use GC enter/exit (might be more) for various reasons when calling into OS API's that could block. We should at least document why they don't need or shouldn't do it or if there is no reason not do it, add GC safe transition around OS API calls that could block (for any reason):

mono_dl_open_file
mono_file_map_error
mono_file_map
mono_file_unmap
mono_mprotect
mono_coree_set_act_ctx
mono_load_coree
open_handle
mono_mmap_open_file
mono_mmap_close
mono_mmap_flush
mono_mmap_map
mono_mmap_unmap
mono_process_create_process
ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal
ves_icall_System_Diagnostics_Process_GetProcesses_internal
ves_icall_System_Threading_Semaphore_CreateSemaphore_icall
ves_icall_System_Threading_Semaphore_ReleaseSemaphore_internal
ves_icall_System_Threading_Semaphore_OpenSemaphore_icall

Another idea/thought, instead of documenting why it can't be used in various places, we could fix the GC enter/exit safe methods to be noops if called within a context where its currently not supported (during parts of init for example). If we could support that, then we could just add the transitions on all places needed and no need for additional documentation. Note, the GC enter/exit safe macros are changed/replaced by products, so the noop needs to go into underlying functions shared by all clients using Mono runtime.

lambdageek:
@CoffeeFlux This is another good task to get familiar with the code organization in the runtime.

CoffeeFlux:
After talking, it was decided that at least initially it would be best to evaluate and document all these functions, ensuring they behave consistently across platforms, and then evaluate the noop solution later down the line.

CoffeeFlux:
After discussion, we decided that going through and auditing whether these functions can safely transition, ensuring it's consistent across platforms, and if not documenting why is the best approach for now. The noop solution can be evaluated at a later date.

alexischr:
[...]

PR #14612, opened 2019-05-23
[coop] Add GC safe transitions around blocking OS API calls

CoffeeFlux:
If this was not possible due to them being called during runtime initialization, a short comment was left noting the specific reason why.

Fixes https://github.com/mono/mono/issues/14377

This definitely needs a careful review, in particular the w32-process* files. Those, along with the other platform-specific ones, could potentially benefit from a more thorough audit, but it's rather difficult to tell what is and isn't being called during initialization. I've tried to be thorough on everything mentioned in the issue.

This includes a couple other random fixes I noticed on the way, which I can split off into a separate PR if preferable.

### DIFF
--- a/mono/metadata/file-mmap-posix.c
+++ b/mono/metadata/file-mmap-posix.c
@@ -192,7 +192,7 @@ access_mode_to_unix (int access)
 }
 
 static int
-acess_to_mmap_flags (int access)
+access_to_mmap_flags (int access)
 {
 	switch (access) {
 	case MMAP_FILE_ACCESS_READ_WRITE:
@@ -526,7 +526,7 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
 	mmap_offset = align_down_to_page_size (offset);
 	eff_size += (offset - mmap_offset);
 	//FIXME translate some interesting errno values
-	res.address = mono_file_map ((size_t)eff_size, acess_to_mmap_flags (access), fh->fd, mmap_offset, &res.free_handle);
+	res.address = mono_file_map ((size_t)eff_size, access_to_mmap_flags (access), fh->fd, mmap_offset, &res.free_handle);
 	res.length = eff_size;
 
 	if (res.address) {
--- a/mono/metadata/file-mmap-posix.c
+++ b/mono/metadata/file-mmap-posix.c
@@ -236,10 +236,12 @@ open_file_map (const char *c_path, int input_fd, int mode, gint64 *capacity, int
 	MmapHandle *handle = NULL;
 	int result, fd;
 
+	MONO_ENTER_GC_SAFE;
 	if (c_path)
 		result = stat (c_path, &buf);
 	else
 		result = fstat (input_fd, &buf);
+	MONO_EXIT_GC_SAFE;
 
 	if (mode == FILE_MODE_TRUNCATE || mode == FILE_MODE_APPEND || mode == FILE_MODE_OPEN) {
 		if (result == -1) { //XXX translate errno?
@@ -275,10 +277,12 @@ open_file_map (const char *c_path, int input_fd, int mode, gint64 *capacity, int
 		}
 	}
 
+	MONO_ENTER_GC_SAFE;
 	if (c_path) //FIXME use io portability?
 		fd = open (c_path, file_mode_to_unix (mode) | access_mode_to_unix (access), DEFAULT_FILEMODE);
 	else
 		fd = dup (input_fd);
+	MONO_EXIT_GC_SAFE;
 
 	if (fd == -1) { //XXX translate errno?
 		*ioerror = COULD_NOT_OPEN;
@@ -353,13 +357,17 @@ open_memory_map (const char *c_mapName, int mode, gint64 *capacity, int access, 
 		strcpy (file_name, tmp_dir);
 		strcat (file_name, MONO_ANON_FILE_TEMPLATE);
 
+		MONO_ENTER_GC_SAFE;
 		fd = mkstemp (file_name);
+		MONO_EXIT_GC_SAFE;
 		if (fd == -1) {
 			*ioerror = COULD_NOT_MAP_MEMORY;
 			goto done;
 		}
 
+		MONO_ENTER_GC_SAFE;
 		unlink (file_name);
+		MONO_EXIT_GC_SAFE;
 #ifdef HAVE_FTRUNCATE
 		unused = ftruncate (fd, (off_t)*capacity);
 #endif
@@ -467,7 +475,9 @@ mono_mmap_close (void *mmap_handle, MonoError *error)
 			g_hash_table_remove (named_regions, handle->name);
 
 		g_free (handle->name);
+		MONO_ENTER_GC_SAFE;
 		close (handle->fd);
+		MONO_EXIT_GC_SAFE;
 		g_free (handle);
 	}
 	named_regions_unlock ();
@@ -480,7 +490,9 @@ mono_mmap_configure_inheritability (void *mmap_handle, gint32 inheritability, Mo
 	int fd, flags;
 
 	fd = h->fd;
+	MONO_ENTER_GC_SAFE;
 	flags = fcntl (fd, F_GETFD, 0);
+	MONO_EXIT_GC_SAFE;
 	if (inheritability)
 		flags &= ~FD_CLOEXEC;
 	else
@@ -495,7 +507,9 @@ mono_mmap_flush (void *mmap_handle, MonoError *error)
 
 #ifdef HAVE_MSYNC
 	if (h)
+		MONO_ENTER_GC_SAFE;
 		msync (h->address, h->length, MS_SYNC);
+		MONO_EXIT_GC_SAFE;
 #endif
 }
 
@@ -525,8 +539,10 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
 
 	mmap_offset = align_down_to_page_size (offset);
 	eff_size += (offset - mmap_offset);
+	MONO_ENTER_GC_SAFE;
 	//FIXME translate some interesting errno values
 	res.address = mono_file_map ((size_t)eff_size, access_to_mmap_flags (access), fh->fd, mmap_offset, &res.free_handle);
+	MONO_EXIT_GC_SAFE;
 	res.length = eff_size;
 
 	if (res.address) {
@@ -544,7 +560,9 @@ mono_mmap_unmap (void *mmap_handle, MonoError *error)
 	int res = 0;
 	MmapInstance *h = (MmapInstance *)mmap_handle;
 
+	MONO_ENTER_GC_SAFE;
 	res = mono_file_unmap (h->address, h->free_handle);
+	MONO_EXIT_GC_SAFE;
 
 	g_free (h);
 	return res == 0;
--- a/mono/metadata/file-mmap-windows.c
+++ b/mono/metadata/file-mmap-windows.c
@@ -320,11 +320,9 @@ void
 mono_mmap_configure_inheritability (void *mmap_handle, gint32 inheritability, MonoError *error)
 {
 	g_assert (mmap_handle);
-	MONO_ENTER_GC_SAFE;
 	if (!SetHandleInformation (mmap_handle, HANDLE_FLAG_INHERIT, inheritability ? HANDLE_FLAG_INHERIT : 0)) {
 		g_error ("mono_mmap_configure_inheritability: SetHandleInformation failed with error %d!", GetLastError ());
 	}
-	MONO_EXIT_GC_SAFE;
 }
 
 void
--- a/mono/metadata/file-mmap-windows.c
+++ b/mono/metadata/file-mmap-windows.c
@@ -157,7 +157,10 @@ open_handle (void *handle, const gunichar2 *mapName, gint mapName_length, int mo
 		}
 	} else {
 		FILE_STANDARD_INFO info;
-		if (!GetFileInformationByHandleEx (handle, FileStandardInfo, &info, sizeof (FILE_STANDARD_INFO))) {
+		MONO_ENTER_GC_SAFE;
+		gboolean getinfo_success = GetFileInformationByHandleEx (handle, FileStandardInfo, &info, sizeof (FILE_STANDARD_INFO));
+		MONO_EXIT_GC_SAFE;
+		if (!getinfo_success) {
 			*ioerror = convert_win32_error (GetLastError (), COULD_NOT_OPEN);
 			return NULL;
 		}
@@ -175,16 +178,22 @@ open_handle (void *handle, const gunichar2 *mapName, gint mapName_length, int mo
 	HANDLE result = NULL;
 
 	if (mode == FILE_MODE_CREATE_NEW || handle != INVALID_HANDLE_VALUE) {
+		MONO_ENTER_GC_SAFE;
 		result = CreateFileMappingW (handle, NULL, get_page_access (access) | options, (DWORD)(((guint64)*capacity) >> 32), (DWORD)*capacity, mapName);
+		MONO_EXIT_GC_SAFE;
 		if (result && GetLastError () == ERROR_ALREADY_EXISTS) {
+			MONO_ENTER_GC_SAFE;
 			CloseHandle (result);
+			MONO_EXIT_GC_SAFE;
 			result = NULL;
 			*ioerror = FILE_ALREADY_EXISTS;
 		} else if (!result && GetLastError () != NO_ERROR) {
 			*ioerror = convert_win32_error (GetLastError (), COULD_NOT_OPEN);
 		}
 	} else if (mode == FILE_MODE_OPEN || mode == FILE_MODE_OPEN_OR_CREATE && access == MMAP_FILE_ACCESS_WRITE) {
+		MONO_ENTER_GC_SAFE;
 		result = OpenFileMappingW (get_file_map_access (access), FALSE, mapName);
+		MONO_EXIT_GC_SAFE;
 		if (!result) {
 			if (mode == FILE_MODE_OPEN_OR_CREATE && GetLastError () == ERROR_FILE_NOT_FOUND) {
 				*ioerror = INVALID_FILE_MODE;
@@ -210,14 +219,18 @@ open_handle (void *handle, const gunichar2 *mapName, gint mapName_length, int mo
 		guint32 waitSleep = 0;
 
 		while (waitRetries > 0) {
+			MONO_ENTER_GC_SAFE;
 			result = CreateFileMappingW (handle, NULL, get_page_access (access) | options, (DWORD)(((guint64)*capacity) >> 32), (DWORD)*capacity, mapName);
+			MONO_EXIT_GC_SAFE;
 			if (result)
 				break;
 			if (GetLastError() != ERROR_ACCESS_DENIED) {
 				*ioerror = convert_win32_error (GetLastError (), COULD_NOT_OPEN);
 				break;
 			}
+			MONO_ENTER_GC_SAFE;
 			result = OpenFileMappingW (get_file_map_access (access), FALSE, mapName);
+			MONO_EXIT_GC_SAFE;
 			if (result)
 				break;
 			if (GetLastError () != ERROR_FILE_NOT_FOUND) {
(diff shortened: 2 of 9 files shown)


mono/metadata/file-mmap-posix.c
```diff
@@ -275,10 +277,12 @@ open_file_map (const char *c_path, int input_fd, int mode, gint64 *capacity, int
 		}
 	}
 
+	MONO_ENTER_GC_SAFE;
 	if (c_path) //FIXME use io portability?
 		fd = open (c_path, file_mode_to_unix (mode) | access_mode_to_unix (access), DEFAULT_FILEMODE);
 	else
 		fd = dup (input_fd);
```
jaykrell: `dup`?

mono/metadata/file-mmap-posix.c
```diff
@@ -480,7 +490,9 @@ mono_mmap_configure_inheritability (void *mmap_handle, gint32 inheritability, Mo
 	int fd, flags;
 
 	fd = h->fd;
+	MONO_ENTER_GC_SAFE;
 	flags = fcntl (fd, F_GETFD, 0);
```
jaykrell: Does this do I/o or just set flags?
Granted there could be a lock around those flags, but it should be brief, unles its holder is otherwise busy.

mono/metadata/file-mmap-posix.c
```diff
@@ -525,8 +539,10 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
 
 	mmap_offset = align_down_to_page_size (offset);
 	eff_size += (offset - mmap_offset);
+	MONO_ENTER_GC_SAFE;
 	//FIXME translate some interesting errno values
-	res.address = mono_file_map ((size_t)eff_size, acess_to_mmap_flags (access), fh->fd, mmap_offset, &res.free_handle);
```
jaykrell: Does this actually do any I/O?

mono/metadata/file-mmap-posix.c
```diff
@@ -544,7 +560,9 @@ mono_mmap_unmap (void *mmap_handle, MonoError *error)
 	int res = 0;
 	MmapInstance *h = (MmapInstance *)mmap_handle;
 
+	MONO_ENTER_GC_SAFE;
 	res = mono_file_unmap (h->address, h->free_handle);
```
jaykrell: Does this actually do any I/O?

mono/metadata/file-mmap-windows.c
```diff
@@ -175,16 +178,22 @@ open_handle (void *handle, const gunichar2 *mapName, gint mapName_length, int mo
...
 			CloseHandle (result);
+			MONO_EXIT_GC_SAFE;
 			result = NULL;
 			*ioerror = FILE_ALREADY_EXISTS;
 		} else if (!result && GetLastError () != NO_ERROR) {
 			*ioerror = convert_win32_error (GetLastError (), COULD_NOT_OPEN);
 		}
 	} else if (mode == FILE_MODE_OPEN || mode == FILE_MODE_OPEN_OR_CREATE && access == MMAP_FILE_ACCESS_WRITE) {
+		MONO_ENTER_GC_SAFE;
 		result = OpenFileMappingW (get_file_map_access (access), FALSE, mapName);
```
jaykrell: Surely this does no I/O. It opens an in-memory construct. But it'd have to use locks, briefly.

mono/metadata/file-mmap-windows.c
```diff
@@ -292,16 +311,20 @@ void
...
 	CloseHandle (mmap_handle);
+	MONO_EXIT_GC_SAFE;
 }
 
 void
 mono_mmap_configure_inheritability (void *mmap_handle, gint32 inheritability, MonoError *error)
 {
 	g_assert (mmap_handle);
+	MONO_ENTER_GC_SAFE;
 	if (!SetHandleInformation (mmap_handle, HANDLE_FLAG_INHERIT, inheritability ? HANDLE_FLAG_INHERIT : 0)) {
```
jaykrell: This does no I/O.

mono/metadata/file-mmap-windows.c
```diff
@@ -365,13 +395,17 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
...
 	void *address = MapViewOfFile (handle, get_file_map_access (access), (DWORD) (newOffset >> 32), (DWORD) newOffset, (SIZE_T) nativeSize);
+	MONO_EXIT_GC_SAFE;
 	if (!address)
 		return convert_win32_error (GetLastError (), COULD_NOT_MAP_MEMORY);
 
 	// Query the view for its size and allocation type
 	MEMORY_BASIC_INFORMATION viewInfo;
+	MONO_ENTER_GC_SAFE;
 	VirtualQuery (address, &viewInfo, sizeof (MEMORY_BASIC_INFORMATION));
+	MONO_EXIT_GC_SAFE;
```
jaykrell: This does no I/O. But does involve locks.

mono/metadata/file-mmap-windows.c
```diff
@@ -385,12 +419,16 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
 	// and size of the region of pages with matching attributes starting from base address.
 	// VirtualQueryEx: http://msdn.microsoft.com/en-us/library/windows/desktop/aa366907(v=vs.85).aspx
 	if (((viewInfo.State & MEM_RESERVE) != 0) || viewSize < (guint64) nativeSize) {
+		MONO_ENTER_GC_SAFE;
 		void *tempAddress = VirtualAlloc (address, nativeSize != 0 ? nativeSize : viewSize, MEM_COMMIT, get_page_access (access));
```
jaykrell: No I/O, just locks.

mono/metadata/file-mmap-windows.c
```diff
@@ -385,12 +419,16 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
...
+		MONO_EXIT_GC_SAFE;
 		if (!tempAddress) {
 			return convert_win32_error (GetLastError (), COULD_NOT_MAP_MEMORY);
 		}
 		// again query the view for its new size
+		MONO_ENTER_GC_SAFE;
 		VirtualQuery (address, &viewInfo, sizeof (MEMORY_BASIC_INFORMATION));
```
jaykrell: No I/O, just locks.

mono/metadata/w32semaphore-win32.c
```diff
@@ -25,7 +25,9 @@ gpointer
 ves_icall_System_Threading_Semaphore_CreateSemaphore_icall (gint32 initialCount, gint32 maximumCount,
 	const gunichar2 *name, gint32 name_length, gint32 *win32error, MonoError *error)
 { 
+	MONO_ENTER_GC_SAFE;
 	HANDLE sem = CreateSemaphoreW (NULL, initialCount, maximumCount, name);
```
jaykrell: This does no I/O and only barely uses locks.

mono/metadata/w32semaphore-win32.c
```diff
@@ -34,14 +36,18 @@ ves_icall_System_Threading_Semaphore_CreateSemaphore_icall (gint32 initialCount,
 MonoBoolean
 ves_icall_System_Threading_Semaphore_ReleaseSemaphore_internal (gpointer handle, gint32 releaseCount, gint32 *prevcount, MonoError *error)
 {
+	MONO_ENTER_GC_SAFE;
 	return ReleaseSemaphore (handle, releaseCount, (PLONG)prevcount);
```
jaykrell: Releasing a lock would seem "fast".

mono/utils/mono-mmap.c
```diff
@@ -453,6 +455,7 @@ mono_mprotect (void *addr, size_t length, int flags)
 #endif
 #endif
 	}
+	// No GC safe transition because this is called early in mini_init via mono_arch_init (with a few layers of indirection)
 	return mprotect (addr, length, prot);
```
jaykrell: Would seem to be free of I/O, but involve a lock (all virtual alloc/free/protect requires synchronization, like flushing multi-processor TLBs).

mono/metadata/file-mmap-posix.c
CoffeeFlux: I'm pretty sure this is one of the ones where I checked to make sure it used a lock internally. Not sure how brief that would be and whether it's worth the transition, so I included it for now.

lambdageek:
@jaykrell 
1. "no I/O, just locks" that's worth a GC Safe transition.  Can't block from GC Unsafe.
2. In general unless there's very good reason (ie, it shows up hot in profiling and we are really certain the implementation won't change) we should do thread state transitions around syscalls.

jaykrell:
I guess nevermind then, but ReleaseSemaphore, releasing a lock?

mono/metadata/file-mmap-posix.c
CoffeeFlux: I thiiiiiiink `dup` calls `fcntl` internally? I should probably verify one way or another.

mono/metadata/file-mmap-posix.c
CoffeeFlux: It calls mmap internally

jaykrell: Which doesn't do I/O usually. Reads are deferred until you touch memory. Writes are deferred until there is memory pressure. Except on unusual systems w/o paging which will do the reading up front.

CoffeeFlux: Even without I/O, it should use a bunch of locks, so again I've opted to leave the transition in just in case

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

mono/metadata/w32process-win32.c
```diff
@@ -81,7 +81,7 @@ ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStar
 	gboolean ret;
 
 	shellex.cbSize = sizeof(SHELLEXECUTEINFO);
-	shellex.fMask = (gulong)(SEE_MASK_FLAG_DDEWAIT | SEE_MASK_NOCLOSEPROCESS | SEE_MASK_UNICODE);
+	shellex.fMask = (gulong)(SEE_MASK_NOASYNC | SEE_MASK_NOCLOSEPROCESS | SEE_MASK_UNICODE);
```
lambdageek: Why the flag change?

CoffeeFlux: Commit description has the full info, but tldr it's the same value and the old one is deprecated

jaykrell: This is reasonable and Win32 stinks.
The documentation says they have the same value, but don't use one, use the other.

mono/metadata/w32process-win32.c
```diff
@@ -209,6 +214,7 @@ process_complete_path (const gunichar2 *appname, gchar **completed)
 	utf8appmemory = g_utf16_to_utf8 (appname, -1, NULL, NULL, NULL);
 	utf8app = process_unquote_application_name (utf8appmemory);
 
+	MONO_ENTER_GC_SAFE;
 	if (g_path_is_absolute (utf8app)) {
 		*completed = process_quote_path (utf8app);
 		result = TRUE;
```
baulig: Does this do any I/O?

mono/metadata/w32process-win32.c
```diff
@@ -230,6 +236,7 @@ process_complete_path (const gunichar2 *appname, gchar **completed)
 
 	*completed = process_quote_path (found);
 exit:
+	MONO_EXIT_GC_SAFE;
 	g_free (found);
```
baulig: We should be careful to make sure that this `exit` label cannot be reached by some code outside the `MONO_ENTER_GC_SAFE` - or is it safe to call `MONO_EXIT_GC_SAFE` twice?

mono/metadata/w32process-win32.c
jaykrell: The `g_find_program_in_path` part presumably. Could probably reduce the scope to that. Or enlarge it in caller.

jaykrell: And `g_file_test`. Maybe those functions should handle it themselves?

jaykrell: Hard to say if these thing should be narrow as possible, or broad.
Doing it the next level down catches all callers.
Doing it higher limits the transitions, if that matters.

mono/metadata/w32process-win32.c
```diff
@@ -99,7 +99,10 @@ ves_icall_System_Diagnostics_Process_ShellExecuteEx_internal (MonoW32ProcessStar
 	else
 		shellex.fMask = (gulong)(shellex.fMask | SEE_MASK_FLAG_NO_UI);
 
+	MONO_ENTER_GC_SAFE;
 	ret = ShellExecuteEx (&shellex);
+	MONO_EXIT_GC_SAFE;
+	
 	if (ret == FALSE) {
```
jaykrell: Consider moving GetLastError to before `MONO_EXIT_GC_SAFE`.

mono/metadata/file-mmap-windows.c
```diff
@@ -413,7 +454,10 @@ mono_mmap_unmap (void *mmap_handle, MonoError *error)
 
 	MmapInstance *h = (MmapInstance *) mmap_handle;
 
-	gboolean result = UnmapViewOfFile (h->address);
+	gboolean result;
+	MONO_ENTER_GC_SAFE;
+	result = UnmapViewOfFile (h->address);
+	MONO_EXIT_GC_SAFE;
 
 	g_free (h);
```
jaykrell: Doesn't `g_free` potentially lock also?

mono/metadata/w32process-win32.c
CoffeeFlux: You can't, really - the macro, when paired with a corresponding MONO_ENTER_GC_SAFE, actually expands into a while loop. With that said, at least in general I believe trying to transition to a state you're already in will throw an error.

mono/metadata/w32process-win32.c
lambdageek: I don't think we should do GC transitions inside the `g_*` functions - it kind of violates layering.  (but we could make coop-safe versions of them in `mono/utils/`)

mono/metadata/file-mmap-windows.c
```diff
@@ -385,12 +421,17 @@ mono_mmap_map (void *handle, gint64 offset, gint64 *size, int access, void **mma
...
 	// VirtualQueryEx: http://msdn.microsoft.com/en-us/library/windows/desktop/aa366907(v=vs.85).aspx
 	if (((viewInfo.State & MEM_RESERVE) != 0) || viewSize < (guint64) nativeSize) {
-		void *tempAddress = VirtualAlloc (address, nativeSize != 0 ? nativeSize : viewSize, MEM_COMMIT, get_page_access (access));
+		void *tempAddress;
+		MONO_ENTER_GC_SAFE;
+		tempAddress = VirtualAlloc (address, nativeSize != 0 ? nativeSize : viewSize, MEM_COMMIT, get_page_access (access));
+		MONO_EXIT_GC_SAFE;
 		if (!tempAddress) {
 			return convert_win32_error (GetLastError (), COULD_NOT_MAP_MEMORY);
 		}
```
jaykrell: Consider `GetLastError` before `MONO_EXIT_GC_SAFE`. I need to check if `MONO_EXIT_GC_SAFE` can touch last error. Perhaps not, in which case, nevermind.

mono/metadata/w32process-win32.c
jaykrell: It looks like it doesn't matter, nevermind.

mono/metadata/w32process-win32.c
jaykrell: `g_free` can surely lock, though it'll tend not to. It'll tend to just be an interlocked push onto a free list.

lambdageek: Agree. the scope on this one feels too big.

mono/metadata/w32process-win32.c
jaykrell: Have to decide if `g_*` really has meaning outside of mono runtime with coop-compatibility, maybe yes, maybe no, and, if they do, if they themselves shouldn't have two variations, vs. visiting/wrapping all the calls.

So, yes, agreed, but the layering purity might not serve any user, or might be slightly onerous to implement, but ok either way.

mono/metadata/file-mmap-windows.c
lambdageek: > Doesn't `g_free` potentially lock also?

We're... not really prepared to deal with malloc/free properly.

mono/metadata/file-mmap-windows.c
jaykrell: It looks like it is ok. `MONO_EXIT_GC_SAFE` appears to deliberately preserve lasterror.

mono/metadata/w32process-win32.c
CoffeeFlux: Yeah it hits a couple functions inside of it, and as Martin brought up elsewhere the heavy use of goto makes it important that we be sure the transition is always hit on both parts and because of the way it expands. I opted to just wrap this whole section rather than a subset so I didn't have to keep transitioning back and forth, but I can't hit just the two parts with the blocking functions because of the gotos. I could rewrite the whole thing, but opted for this instead.

mono/metadata/file-mmap-windows.c
jaykrell: Modulo a possible but in its first time initialization call..can be fixed separately.

mono/metadata/file-mmap-windows.c
jaykrell: Notice that just getting the policy, at least the first time, can `g_free`. So there is a possible recursion there, if implemented.

CoffeeFlux: Ergh, yeah I think so, though I think it'll almost always be super brief similar to ReleaseSemaphore?

mono/metadata/w32process-win32.c
CoffeeFlux: Looking at feedback on the other part, I'll just wrap the individual blocking calls and transition back and forth a few times.

baulig: How expensive are those `MONO_ENTER_GC_SAFE` transitions?

In general, I just don't like this goto pattern very much because it's easy for somebody at a later time either add a goto statement from outside the protected block - or a return statement.

Where is this method called from?  Maybe we should consider protecting the caller instead.

mono/metadata/file-mmap-windows.c
jaykrell: `g_free` at least on Windows, will tend not to lock.
There are a few imaginable things it could do.

1. It could free to an interlocked list based on the size of the block.
Very fast, no lock (aside from what the CPU does, or an InterlockedCompareExchange loop).

2. It could coalesece adjacent free blocks into larger blocks.
I'm not sure that requires locking, or if it happens, but it probably happens.
Splitting/coalescing seems like basic heap operations, unless somehow optimized away.

3. It could, with a lot of free'ing, actually VirtualFree a large block, incuring a syscall and more locking. Though I think by default, on Windows, this might never happen.
Even if it happens sometimes, it'd be less often than you might expect, like to avoid "hysterisis" where when you are on the boundary, it'd keep going back and forth, VirtualAlloc/VirtualFree, turning a small alloc/free loop into slow.

jaykrell: But anyway, it seems we are ignoring `malloc` / `free` for now.

baulig: How would `g_free` lock?

jaykrell: `g_free` and `g_malloc` are managing a per-process heap.
You can `malloc` on one thread, and `free` it on another.
At some conceptual level, it must lock, because it is managing a process-wide resource.

That said, it is an area with a lot of optimization.
There tend to be a few size-buckets and per-thread free lists.
Or interlocked process free lists.

As well, under heavy `malloc` (not necessarily `free`), where you are hitting new high water marks in terms of memory use, `malloc` must go to the kernel for more memory, `VirtualAlloc` or `mmap`. Which is a very heavyweight operation. It has to change page table entries. And flush per-processor page table caches (TLB entries).

mono/metadata/w32process-win32.c
jaykrell: `goto` is one of my favorite ways to refactor code to avoid duplication and bugs and simplify error handling and flow control.

Some people consider it evil but I find it is really quite noble.
Opinion is strongly mixed on the matter unfortunately. :(

The alternative is a mess of if/else.

Early return is just as good, if you have no resources to cleanup, invariants to restore, or you have C++ destructors. But that doesn't apply here, so `goto`.

mono/metadata/file-mmap-windows.c
CoffeeFlux: Are you planning to fix that bug yourself, or do I need to look at it? If the latter, can you point out exactly what the issue is?

jaykrell: I'll fix it. It is a very rare or perhaps nonexistent thing. Maybe I misread the code.
Just for your curiousity, see around `threads_suspend_policy_getenv` and `W32_DEFINE_LAST_ERROR_RESTORE_POINT`.

probably best to deal with it in `hasenv_obsolete`.


Consider broadening `W32_DEFINE_LAST_ERROR_RESTORE_POINT` to also preserve `errno`, on all systems.

jaykrell: Actually, fix it in the slow path of `mono_threads_suspend_policy` probably.

mono/metadata/w32process-win32.c
jaykrell: You mean, because each of the underlying calls might change to not need it, and the need could fall away completely? That would argue for pushing into the callee, except for the layering `g_` problem?

mono/metadata/file-mmap-windows.c
CoffeeFlux: Preserving `errno` would be pretty helpful, I didn't realize that wasn't the current behavior. I don't think it matters for this PR, but it's possible for the lack of that to have consequences elsewhere in the codebase.

CoffeeFlux:
@monojenkins build failed

mono/metadata/file-mmap-windows.c
jaykrell: https://github.com/mono/mono/pull/14705 is what I was referring to.

CoffeeFlux:
@monojenkins build failed

PR #14618, opened 2019-05-24
[jit] Also allow System.Runtime.CompilerServices.Unsafe intrinsics from the System.Runtime.CompilerServices assembly, not just corlib.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
The failures are relevant.

lewurm:
Is it safe to assume that the same intrinsic should be used for an arbitrary assembly which happens to have the same name?

vargaz:
Unfortunately, there is no way to identify the 'official' assembly, so yes.

vargaz:
@monojenkins build failed

PR #14619, opened 2019-05-24
[System.Net.NetworkInformation] Fork FreeBSD from MacOs

### LINKED ISSUE #13400: IPv6 support broken on FreeBSD because AF_INET6 is 28 on FreeBSD not 30 like it is on MacOS
The change is pretty simple.  FreeBSD uses MacOsNetworkInterfaceAPI which statically defines AF_INET6 to 30 but on FreeBSD AF_INET6 is 28, resulting in various places where mono checks if the OS supports IPv6 (Socket.OSSupportsIPv6) to fail even when IPv6 is supported.

I submitted a patch to FreeBSD here: https://bz-attachments.freebsd.org/attachment.cgi?id=201988, but this should probably be fixed in a much better way and probably not reproduce the entire MacOsNetworkInterface.cs.

## Steps to Reproduce

1. On FreeBSD do sharp -e 'new System.Net.WebClient ().DownloadString ("http://www.v6.facebook.com/status.php")'

### Current Behavior

Connections to IPv6 only hosts fails on FreeBSD

### Expected Behavior

Connections succeed.

[ ] macOS
[ ] Linux
[ ] Windows
[x] FreeBSD

**Version Used**:
5.10.1.57

## Stacktrace

```
```

cemeyer:
They are related, but not identical.  In particular, MacOS AF_INET6 is wrong
on FreeBSD and breaks IPv6 networking on FreeBSD.  Establish an inheritence
relationship with overrides for now; it's possible more differences will be
needed in the future.

This change is released under the MIT license.

Fixes #13400


mcs/class/System/System.Net.NetworkInformation/MacOsNetworkInterface.cs
```diff
@@ -36,9 +36,16 @@
 namespace System.Net.NetworkInformation {
 	internal class MacOsNetworkInterfaceAPI : UnixNetworkInterfaceAPI
 	{
-		const int AF_INET  = 2;
-		const int AF_INET6 = 30;
-		const int AF_LINK  = 18;
+		protected int AF_INET;
+		protected int AF_INET6;
```
marek-safar: Please change AF_INET6 only

marek-safar:
approve

mcs/class/System/System.Net.NetworkInformation/MacOsNetworkInterface.cs
filipnavara: ...and make it `readonly`.

cemeyer: @marek-safar Sure, will do.  @filipnavara If that's the right attribute, I'm happy to add it.  Thanks!

cemeyer:
Left `AF_INET` and `AF_LINK` `const`; added `readonly` attribute to `AF_INET6` and protected ctor in parent class to allow override by FreeBSD child class.

cemeyer:
Looks like the Linux x64 builder (CI) is out of disk space.

baulig:
@monojenkins commit csproj

marek-safar:
Thank you!

cemeyer:
Thanks! 😁

PR #14621, opened 2019-05-24
[wasm] fix most cases of numeric debugger display for wasm

lewing:
Fix the signededness of numeric types and use double
for the display of everthing since we are stuck with that
on the js side.

image

need to add tests


kg:
Now that I think about it, it's my understanding that v8 has bigint support (and Firefox is shipping it soon), so perhaps we could use that to visualize long and ulong?

kg:
@monojenkins build failed

lewing:
The debugger test harness is problematic right now and 64bit values are tricky so I've opened https://github.com/mono/mono/issues/15124 to track that and I'll land this as is because it contains critical fixes.

PR #14625, opened 2019-05-24
Cleanup: load_function_full -- avoid a string alloc/copy and cast.

jaykrell:


lewurm:
@monojenkins squash

jaykrell:
@monojenkins build failed

PR #14628, opened 2019-05-24
Fix interpreter mini regression test on Windows x64.

lateralusX:
First set of fixes to get the following tests working under interpreter on Windows x64:

* mono/mini richeck
* mono/mini mixedcheck
* mono/mini fullaotmixedcheck
* mono/tests winx64structs.exe

Fixes are primarily related to Windows x64 ABI support in interpreter. Fixes also makes sure our interpreter trampolines get unwind info making it possible to debug through the trampolines on Windows x64.

Several fixes related to value types was needed, also fixed support for return value types not fitting into registers, current implementation didn't return passed in value type address in return register as mandated by both System V and Win x64 ABI's.

Identified and fixed memory leak on every pinvoke from interpreter that needs stack. Since Windows x64 ABI always use 32 bytes of stack allocated by caller, the memory leak was hit on every pinvoke call.

### DIFF
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -1095,7 +1095,8 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
 static int
 arg_need_temp (ArgInfo *ainfo)
 {
-	if (ainfo->storage == ArgValuetypeInReg)
+	// Value types using one register doesn't need temp.
+	if (ainfo->storage == ArgValuetypeInReg && ainfo->nregs > 1)
 		return ainfo->nregs * sizeof (host_mgreg_t);
 	return 0;
 }
@@ -1110,11 +1111,27 @@ arg_get_storage (CallContext *ccontext, ArgInfo *ainfo)
 		case ArgInDoubleSSEReg:
 			return &ccontext->fregs [ainfo->reg];
 		case ArgOnStack:
+		case ArgValuetypeAddrOnStack:
 			return ccontext->stack + ainfo->offset;
 		case ArgValuetypeInReg:
 			// Empty struct
-			g_assert (!ainfo->nregs);
-			return NULL;
+			if (ainfo->nregs == 0)
+				return NULL;
+			// Value type using one register can be stored
+			// directly in its context gregs/fregs slot.
+			g_assert (ainfo->nregs == 1);
+			switch (ainfo->pair_storage [0]) {
+				case ArgInIReg:
+					return &ccontext->gregs [ainfo->pair_regs [0]];
+				case ArgInFloatSSEReg:
+				case ArgInDoubleSSEReg:
+					return &ccontext->fregs [ainfo->pair_regs [0]];
+				default:
+					g_assert_not_reached ();
+			}
+		case ArgValuetypeAddrInIReg:
+			g_assert (ainfo->pair_storage [0] == ArgInIReg && ainfo->pair_storage [1] == ArgNone);
+			return &ccontext->gregs [ainfo->pair_regs [0]];
 		default:
 			g_error ("Arg storage type not yet supported");
 	}
@@ -1195,16 +1212,33 @@ mono_arch_set_native_call_context_args (CallContext *ccontext, gpointer frame, M
 
 	for (int i = 0; i < sig->param_count; i++) {
 		ainfo = &cinfo->args [i];
-		int temp_size = arg_need_temp (ainfo);
 
-		if (temp_size)
-			storage = alloca (temp_size); // FIXME? alloca in a loop
-		else
-			storage = arg_get_storage (ccontext, ainfo);
+		switch (ainfo->storage) {
+			case ArgInIReg:
+			case ArgInFloatSSEReg:
+			case ArgInDoubleSSEReg:
+			case ArgOnStack:
+			case ArgValuetypeInReg: {
+				int temp_size = arg_need_temp (ainfo);
+				if (temp_size)
+					storage = alloca (temp_size); // FIXME? alloca in a loop
+				else
+					storage = arg_get_storage (ccontext, ainfo);
 
-		interp_cb->frame_arg_to_data ((MonoInterpFrameHandle)frame, sig, i, storage);
-		if (temp_size)
-			arg_set_val (ccontext, ainfo, storage);
+				interp_cb->frame_arg_to_data ((MonoInterpFrameHandle)frame, sig, i, storage);
+				if (temp_size)
+					arg_set_val (ccontext, ainfo, storage);
+				break;
+			}
+			case ArgValuetypeAddrInIReg:
+			case ArgValuetypeAddrOnStack: {
+				storage = arg_get_storage (ccontext, ainfo);
+				*(gpointer *)storage = interp_cb->frame_arg_to_storage (frame, sig, i);
+				break;
+			}
+			default:
+				g_assert_not_reached ();
+		}
 	}
 
 	g_free (cinfo);
@@ -1237,6 +1271,14 @@ mono_arch_set_native_call_context_ret (CallContext *ccontext, gpointer frame, Mo
 		if (temp_size)
 			arg_set_val (ccontext, ainfo, storage);
 	}
+#ifdef TARGET_WIN32
+	// Windows x64 ABI ainfo implementation includes info on how to return value type address.
+	// back to caller.
+	else {
+		storage = arg_get_storage (ccontext, ainfo);
+		*(gpointer *)storage = interp_cb->frame_arg_to_storage (frame, sig, -1);
+	}
+#endif
 
 	g_free (cinfo);
 }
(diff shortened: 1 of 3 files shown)


lewurm:
fullaotmixed fails on linux/amd64:
```
Thread 1 (Thread 0x7f28be5f5740 (LWP 55245)):
#0  0x00007f28bdac2b5a in waitpid () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00005568373605c3 in dump_native_stacktrace (mctx=mctx@entry=0x7ffcdd3d8f40, signal=0x5568375da218 "SIGABRT") at mini-posix.c:1110
#2  0x0000556837360711 in mono_dump_native_crash_info (signal=signal@entry=0x5568375da218 "SIGABRT", mctx=mctx@entry=0x7ffcdd3d8f40, info=info@entry=0x7ffcdd3d9230) at mini-posix.c:1158
#3  0x000055683730d449 in mono_handle_native_crash (signal=signal@entry=0x5568375da218 "SIGABRT", mctx=mctx@entry=0x7ffcdd3d8f40, info=info@entry=0x7ffcdd3d9230) at mini-exceptions.c:3324
#4  0x000055683735fa4f in sigabrt_signal_handler (_dummy=6, _info=0x7ffcdd3d9230, context=0x7ffcdd3d9100) at mini-posix.c:234
#5  <signal handler called>
#6  0x00007f28bd52efff in raise () from /lib/x86_64-linux-gnu/libc.so.6
#7  0x00007f28bd53042a in abort () from /lib/x86_64-linux-gnu/libc.so.6
#8  0x00005568375bc507 in monoeg_assert_abort () at goutput.c:57
#9  0x00005568375a2b79 in mono_log_write_logfile (log_domain=<optimized out>, level=<optimized out>, hdr=<optimized out>, message=0x556838c93490 "* Assertion at mini-amd64.c:1133, condition `ainfo->pair_storage [0] == ArgInIReg && ainfo->pair_storage [1] == ArgNone' not met\n") at mono-log-common.c:136
#10 0x00005568375bc42e in monoeg_g_logstr (msg=<optimized out>, log_level=G_LOG_LEVEL_ERROR, log_domain=0x0) at goutput.c:134
#11 monoeg_g_logv_nofree (log_domain=log_domain@entry=0x0, log_level=log_level@entry=G_LOG_LEVEL_ERROR, format=format@entry=0x5568375c55b0 "* Assertion at %s:%d, condition `%s' not met\n", args=args@entry=0x7ffcdd3d9998) at goutput.c:149
#12 0x00005568375bc900 in monoeg_g_logv_nofree (args=0x7ffcdd3d9998, format=0x5568375c55b0 "* Assertion at %s:%d, condition `%s' not met\n", log_level=G_LOG_LEVEL_ERROR, log_domain=0x0) at goutput.c:142
#13 monoeg_assertion_message (format=format@entry=0x5568375c55b0 "* Assertion at %s:%d, condition `%s' not met\n") at goutput.c:184
#14 0x00005568373325a1 in arg_get_storage (ccontext=ccontext@entry=0x7ffcdd3d9c40, ainfo=ainfo@entry=0x556838c933dc) at mini-amd64.c:1133
#15 0x00005568373356cb in mono_arch_set_native_call_context_ret (ccontext=ccontext@entry=0x7ffcdd3d9c40, frame=frame@entry=0x7ffcdd3d9b40, sig=sig@entry=0x556838c951d0) at mini-amd64.c:1274
#16 0x00005568373dc5f5 in interp_entry_from_trampoline (ccontext_untyped=<optimized out>, rmethod_untyped=0x556838c8ad38) at interp/interp.c:2549
#17 0x00007f28bacec607 in wrapper_other_object___interp_lmf_mono_interp_entry_from_trampoline_intptr_intptr () from /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mono/mini/fullaot-tmp/mscorlib.dll.so
#18 0x00007f28bad5b9cd in native_to_interp_trampoline () from /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mono/mini/fullaot-tmp/mscorlib.dll.so
#19 0x00007f28ba7fd4b1 in System_IO_Path_IsPathRooted_string () from /mnt/jenkins/workspace/test-mono-pull-request-interpreter/mono/mini/fullaot-tmp/mscorlib.dll.so
```

mono/mini/mini-amd64.c
```diff
@@ -1236,6 +1270,9 @@ mono_arch_set_native_call_context_ret (CallContext *ccontext, gpointer frame, Mo
 		interp_cb->frame_arg_to_data ((MonoInterpFrameHandle)frame, sig, -1, storage);
 		if (temp_size)
 			arg_set_val (ccontext, ainfo, storage);
+	} else {
+		storage = arg_get_storage (ccontext, ainfo);
+		*(gpointer *)storage = interp_cb->frame_arg_to_storage (frame, sig, -1);
```
lewurm: If I understand correctly from this conversation (https://gitter.im/mono/mono?at=5ce7bda4ad024978c61dee93) with @BrzVlad, this would duplicate work on System V ABI. We should `ifdef` Windows specific ABI

lateralusX: We could ifdef, but just to clarify the behavior, when using value types as return, caller allocates memory, pass address to memory in arg1 register and callee should return the same address in the return register. So it is jus a reference to the memory that is passed over, used directly by callee and then returned back to caller. The way we handle the ainfo and call info should already abstract this away, so this is already abstracted away in ainfo and the address is already copied into the frame on call and only extracted back and returned in correct register depending on ABI.

lateralusX:
Will fix the assert hit on Linux full AOT, related to System V ABI not setting all fields in ainfo for returned value type. If we make the logic to return value type ABI specific, this will be eliminated.

mono/mini/mini-amd64.c
```diff
@@ -1195,16 +1212,33 @@ mono_arch_set_native_call_context_args (CallContext *ccontext, gpointer frame, M
...
+				interp_cb->frame_arg_to_data ((MonoInterpFrameHandle)frame, sig, i, storage);
+				if (temp_size)
+					arg_set_val (ccontext, ainfo, storage);
+				break;
+			}
+			case ArgValuetypeAddrInIReg:
+			case ArgValuetypeAddrOnStack: {
+				storage = arg_get_storage (ccontext, ainfo);
+				*(gpointer *)storage = interp_cb->frame_arg_to_storage (frame, sig, i);
+				break;
```
BrzVlad: Could we maybe do it in the same fashion as on arm64 (by avoiding big switch and less diff) ?

mono/mini/mini-amd64.c
```diff
@@ -1259,15 +1301,33 @@ mono_arch_get_native_call_context_args (CallContext *ccontext, gpointer frame, M
...
+					storage = arg_get_storage (ccontext, ainfo);
+				}
+				interp_cb->data_to_frame_arg ((MonoInterpFrameHandle)frame, sig, i, storage);
+				break;
+			}
+			case ArgValuetypeAddrInIReg:
+			case ArgValuetypeAddrOnStack: {
+				storage = arg_get_storage (ccontext, ainfo);
+				interp_cb->data_to_frame_arg ((MonoInterpFrameHandle)frame, sig, i, *(gpointer *)storage);
+				break;
```
BrzVlad: same as above comment

mono/mini/tramp-amd64.c
```diff
@@ -937,13 +937,25 @@ mono_arch_get_interp_to_native_trampoline (MonoTrampInfo **info)
...
+	mono_add_unwind_op_offset (unwind_ops, code, start, AMD64_RIP, -cfa_offset);
 
 	amd64_push_reg (code, AMD64_RBP);
+	cfa_offset += sizeof (target_mgreg_t);
+	mono_add_unwind_op_def_cfa_offset (unwind_ops, code, start, cfa_offset);
+	mono_add_unwind_op_offset (unwind_ops, code, start, AMD64_RBP, -cfa_offset);
+
 	amd64_mov_reg_reg (code, AMD64_RBP, AMD64_RSP, sizeof (target_mgreg_t));
+	mono_add_unwind_op_def_cfa_reg (unwind_ops, code, start, AMD64_RBP);
+	mono_add_unwind_op_fp_alloc (unwind_ops, code, start, AMD64_RBP, 0);
```
BrzVlad: do you know what this unwind up actually does ?

mono/mini/tramp-amd64.c
```diff
@@ -1002,11 +1014,20 @@ mono_arch_get_interp_to_native_trampoline (MonoTrampInfo **info)
 	for (i = 0; i < FLOAT_RETURN_REGS; i++)
 		amd64_sse_movsd_membase_reg (code, AMD64_R11, MONO_STRUCT_OFFSET (CallContext, fregs) + i * sizeof (double), i);
 
-	amd64_mov_reg_reg (code, AMD64_RSP, AMD64_RBP, 8);
+#if TARGET_WIN32
+	amd64_lea_membase (code, AMD64_RSP, AMD64_RBP, 0);
```
BrzVlad: I don't understand why this needs to change, given that the prologue appears to stay the same.

mono/mini/tramp-amd64.c
lateralusX: Sets the frame pointer register + inform unwinder about offset. This is mainly used by Win x64 unwind op codes since this information is part of the Win x64 ABI.

mono/mini/tramp-amd64.c
lateralusX: Win x64 ABI only allows a couple of instructions in the epilog when using a frame pointer to correctly work with OS unwinder.

mono/mini/tramp-amd64.c
lateralusX: The fp_alloc is a noop on none Windows platforms.

mono/mini/mini-amd64.c
lateralusX: ok

mono/mini/mini-amd64.c
lateralusX: ok

mono/mini/tramp-amd64.c
jaykrell: @lateralusX is correct.
Epilogues must be detected via runtime disassembly and the disassembler is exceedingly minimal.

https://docs.microsoft.com/en-us/cpp/build/prolog-and-epilog?view=vs-2019

lateralusX:
@monojenkins rebuild failed

PR #14632, opened 2019-05-24
[MacSDK] Trigger redownload of Nuget 5.0.2

alexischr:
Some of the packages/bockbuild caches have somehow ended up with a Nuget 5.0.0 binary, even after the 5.0.2 bump. We suspect that the wrong binary was temporarily published on the 5.0.2 URL


alexischr:

akoeplinger:
@monojenkins build OS X Package

PR #14633, opened 2019-05-24
 [MacSDK] Trigger redownload of Nuget 5.0.2

alexischr:


akoeplinger:
@monojenkins build pkg

akoeplinger:
@monojenkins build pkg

akoeplinger:
@alexischr looks like nuget is still not redownloaded with the comment in the .py file..

akoeplinger:
We decided to just clean the bots instead.

PR #14643, opened 2019-05-25
Remove MONO_PATCH_INFO_METHOD_REL.

jaykrell:
It is not used, and it had a larger than pointer field
where a pointer was required, at least for 32bit hosted cross builds.

By "unused", I'm assuming nobody does math on these enums to forms ones, that I don't see via textual search.


jaykrell:
@monojenkins build failed

PR #14644, opened 2019-05-25
Remove JIT icall hashing by name and AOT JIT icall hashing.

jaykrell:
This time I did not rename `MONO_PATCH_INFO_JIT_ICALL_ADDR` and `MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL`.

Could go either way.
Could change them in a separate PR.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -8616,7 +8616,6 @@ ves_icall_System_IO_LogcatTextWriter_Log (const char *appname, gint32 level, con
 static const MonoIcallTableCallbacks *icall_table;
 static mono_mutex_t icall_mutex;
 static GHashTable *icall_hash = NULL;
-static GHashTable *jit_icall_hash_name = NULL;
 static GHashTable *jit_icall_hash_addr = NULL;
 
 typedef struct _MonoIcallHashTableValue {
@@ -8657,7 +8656,6 @@ void
 mono_icall_cleanup (void)
 {
 	g_hash_table_destroy (icall_hash);
-	g_hash_table_destroy (jit_icall_hash_name);
 	g_hash_table_destroy (jit_icall_hash_addr);
 	mono_os_mutex_destroy (&icall_mutex);
 }
@@ -9104,18 +9102,6 @@ mono_create_icall_signatures (void)
 }
 
 MonoJitICallInfo *
-mono_find_jit_icall_by_name (const char *name)
-{
-	MonoJitICallInfo *info;
-	g_assert (jit_icall_hash_name);
-
-	mono_icall_lock ();
-	info = (MonoJitICallInfo *)g_hash_table_lookup (jit_icall_hash_name, name);
-	mono_icall_unlock ();
-	return info;
-}
-
-MonoJitICallInfo *
 mono_find_jit_icall_by_addr (gconstpointer addr)
 {
 	MonoJitICallInfo *info;
@@ -9128,25 +9114,6 @@ mono_find_jit_icall_by_addr (gconstpointer addr)
 	return info;
 }
 
-/*
- * mono_lookup_jit_icall_symbol:
- *
- *   Given the jit icall NAME, returns its C symbol if possible, or NULL.
- */
-const char*
-mono_lookup_jit_icall_symbol (const char *name)
-{
-	MonoJitICallInfo *info;
-	const char *res = NULL;
-
-	mono_icall_lock ();
-	info = (MonoJitICallInfo *)g_hash_table_lookup (jit_icall_hash_name, name);
-	if (info)
-		res = info->c_symbol;
-	mono_icall_unlock ();
-	return res;
-}
-
 void
 mono_register_jit_icall_wrapper (MonoJitICallInfo *info, gconstpointer wrapper)
 {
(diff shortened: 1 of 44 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Windows is failing for lack of cmake buiding BTLS.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/icall.c
```diff
@@ -9175,40 +9142,48 @@ mono_register_jit_icall_info (MonoJitICallInfo *info, gconstpointer func, const
...
 
-	g_assertf (!info->name && !info->func, "%s", name);
+	g_hash_table_insert (jit_icall_hash_addr, (gpointer)func, info);
+
+	mono_icall_unlock ();
+
+	// Allow duplicate equivalent racy initialization.
+	// The checks are not solid due to races.
+
+	MonoJitICallInfo old = *info;
```
jaykrell: Unnecessary small copy.

mono/mini/aot-runtime.c
```diff
@@ -5282,12 +5275,21 @@ mono_create_ftnptr_malloc (guint8 *code)
...
 	/* No need for locking */
-	if (!aot_jit_icall_hash)
-		aot_jit_icall_hash = g_hash_table_new (g_str_hash, g_str_equal);
-	g_hash_table_insert (aot_jit_icall_hash, (char*)name, addr);
+
+	g_assert (jit_icall_info);
+	g_assert (name);
+	g_assert (addr);
+
+	MonoJitICallInfo old = *jit_icall_info;
```
jaykrell: Remove copy.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/icall.c
```diff
@@ -9175,40 +9142,47 @@ mono_register_jit_icall_info (MonoJitICallInfo *info, gconstpointer func, const
...
 			name, existing_info->name, func, existing_info->func, i);
 	}
 
-	g_assertf (!info->name && !info->func, "%s", name);
+	g_hash_table_insert (jit_icall_hash_addr, (gpointer)func, info);
+
+	mono_icall_unlock ();
+
+	// Allow duplicate equivalent racy initialization.
+	// The checks are not solid due to races.
```
jaykrell: Second line of comment pointless, remove.

jaykrell: Asserts excessive too. Thinking of reducing them or making them conditional.

mono/mini/aot-runtime.c
```diff
@@ -5282,12 +5275,19 @@ mono_create_ftnptr_malloc (guint8 *code)
...
  */
 void
-mono_aot_register_jit_icall (const char *name, gpointer addr)
+mono_aot_register_jit_icall_info (MonoJitICallInfo *info, const char *name, gpointer addr)
 {
 	/* No need for locking */
-	if (!aot_jit_icall_hash)
-		aot_jit_icall_hash = g_hash_table_new (g_str_hash, g_str_equal);
-	g_hash_table_insert (aot_jit_icall_hash, (char*)name, addr);
+
```
jaykrell: I suspect this entire function can be removed.

jaykrell: Or change the switch to use the data. But I think remove the function and the initialization of the data. I need to check. It might also be possible to split the enum into a separate enum. This doesn’t seem horrible though.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/mini/aot-runtime.c
```diff
@@ -5378,28 +5369,92 @@ load_function_full (MonoAotModule *amodule, const char *name, MonoTrampInfo **ou
...
 					target = mono_get_throw_exception_addr ();
-				} else if (!strcmp (ji->data.name, "rethrow_preserve_exception_addr")) {
+					break;
+				case MONO_JIT_ICALL_mono_rethrow_preserve_exception:
 					target = mono_get_rethrow_preserve_exception_addr ();
-				} else if (strstr (ji->data.name, "generic_trampoline_")) {
-					target = mono_aot_get_trampoline (ji->data.name);
-				} else if (aot_jit_icall_hash && g_hash_table_lookup (aot_jit_icall_hash, ji->data.name)) {
-					/* Registered by mono_arch_init () */
-					target = g_hash_table_lookup (aot_jit_icall_hash, ji->data.name);
```
vargaz: This hash was added to decouple the registration of these functions, i.e. to avoid the #ifdefs added below.

jaykrell: I wondered that.
Is it worth it?
How about create per-target functions housing this short code instead?
So no ifdef and similar/same decoupling?

mono/mini/aot-runtime.c
```diff
@@ -5404,57 +5404,9 @@ load_function_full (MonoAotModule *amodule, const char *name, MonoTrampInfo **ou
...
-					break;
-				case MONO_JIT_ICALL_mono_x86_start_gsharedvt_call:
-					target = (gpointer)mono_x86_start_gsharedvt_call;
-					break;
-#endif
 				default:
-					g_error ("Unknown relocation '%p'\n", ji->data.target);
+					target = mono_arch_load_function (ji->data.jit_icall_id);
+					g_assert (target, "Unknown relocation '%p'\n", ji->data.target);
 					target = NULL;
```
jaykrell: oops

jaykrell: A few other mistakes here, fixing.

jaykrell:
@monojenkins build failed

mono/mini/aot-runtime.c
```diff
@@ -5404,58 +5404,9 @@ load_function_full (MonoAotModule *amodule, const char *name, MonoTrampInfo **ou
...
-					break;
-				case MONO_JIT_ICALL_mono_x86_start_gsharedvt_call:
-					target = (gpointer)mono_x86_start_gsharedvt_call;
-					break;
-#endif
 				default:
-					g_error ("Unknown relocation '%p'\n", ji->data.target);
-					target = NULL;
+					target = mono_arch_load_function (ji->data.jit_icall_id);
+					g_assertf (target, "Unknown relocation '%p'\n", ji->data.target);
```
vargaz: Needs some kind of ifdef or needs to be added for all architectures.

jaykrell: I'm "rewriting" it..

jaykrell: It bugged me to lose the inlining. I think the function was provided for all architectures that support AOT, as well as PowerPC, but anyway, the rewrite is tweaked to handle architectures with no extra functions.

Perhaps I should have removed only one hash table in this PR and left the AOT one asis. I could do that, though still heading toward the same place, if ok.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/mini/aot-runtime.c
jaykrell: Another solution is put back the AOT registration, into the struct/array, and just index into that (`mono_find_jit_icall)`.

mono/mini/mini-x86.h
```diff
@@ -350,5 +350,10 @@ mono_x86_start_gsharedvt_call (GSharedVtCallInfo *info, gpointer *caller, gpoint
 CallInfo*
 mono_arch_get_call_info (MonoMemPool *mp, MonoMethodSignature *sig);
 
+#define MONO_ARCH_AOT_ICALLS                           \
```
vargaz: The previous approach was fine, why do these need to be inlined ?

jaykrell: Just keeping it optimal where easy.
You prefer the per-arch helper function, nested switch instead of one switch?
I can put that back and fill in the other architectures.

jaykrell: Registration is also not terrible, then in the switch just index into the array.

vargaz: I prefer the helper function or the registration, this is only run a few times on startup, there is no need to make it optimal.

jaykrell: I'm fine either way. The helper function could also be inline in the .h.
Anything is better than a hashtable imho.
Optimal is also smaller and pays recurring dividends, albeit small ones.

I think when I first did this I didn't realize the connections, so I had redundantly optimized registration, and then didn't use the registration.

The registration also doesn't need the strings.

Note that I removed the ifdef gsharedvt from the helpers, that was in registration.
It isn't really conditional, as the current system is configured.

jaykrell:
Might need the stub in mini-wasm.c. We'll see if CI says.

jaykrell:
@monojenkins build failed

jaykrell:
Yep. `/mnt/jenkins/workspace/test-mono-pull-request-wasm/mono/mini/aot-runtime.c:5394: undefined reference to `mono_arch_load_function'`

jaykrell:
commits after approval:
 - add wasm and mipsv stubs
 - squash it all into one

jaykrell:
@monojenkins build failed

jaykrell:
(accidentally put change in wrong PR, undone)

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Windows x64 FullAOT crashes and there are no symbols in the stack.
Let's see if retry works.

jaykrell:
@monojenkins build failed

jaykrell:
Could use symbolic Windows stacks for the flakes. :(

PR #14648, opened 2019-05-25
Readonly interp callbacks.

jaykrell:


jaykrell:
@monojenkins build failed

lewurm:
Why?

jaykrell:
Generally function pointers should be readonly. C++ vtables for example are. More readonly & less mutable is better. Maximize the utility of any copy-on-write faults by only taking data that might be written. Etc.

jaykrell:
Also notice that some of the function pointers were left uninitialized. Hopefully not called. Null is mildly better.

mono/mini/interp/interp.c
```diff
@@ -6651,6 +6653,41 @@ register_interp_stats (void)
...
+	interp_frame_get_arg,
+	interp_frame_get_local,
+	interp_frame_get_this,
+	interp_frame_arg_to_data,
+	interp_data_to_frame_arg,
+	interp_frame_arg_to_storage,
+	interp_frame_arg_set_storage,
+	interp_frame_get_parent,
+	interp_start_single_stepping,
+	interp_stop_single_stepping,
```
lewurm: I'm actually not a fan of this. What we had previously, gave us some "extra safety".  Here it's easy to get the ordering wrong.

jaykrell: It is nice to match the names up I agree.
 Structs in general cannot be edited lightly because of this sensitivity.
Designated initializers in C++ would help with this.
Readonlyness and static initialization are also valuable. You can't have everything.
Function pointers are sercurity attack vectors and read only ones are preferred.
Granted you can overwrite the data too.
As well as initializing all the fields -- though you could memset to to get that.
The original had uninitialized values.

Notice that typesafety does help, some.
The compiler won't let you initialize functoin pointers w/ functions whose signature does not match. But the signatures are not necesssarily varying, and some could match others.

jaykrell: Also despite the ordering requirement, this is a heavily used idiom.

lewurm: I agree, having uninitialized values for the stub case is problematic. Please submit a PR for that, but changing the structure of the code just because of that is not the way to go. This comes to my mind: https://tirania.org/blog/archive/2010/Dec-31.html

We have `MonoDebuggerCallbacks`, `MonoUtilityThreadCallbacks`, `MonoThreadInfoRuntimeCallbacks`, `MonoThreadInfoCallbacks`, `MonoCodeManagerCallbacks`, `MonoRuntimeCallbacks`, `MonoRuntimeExceptionHandlingCallbacks` and others that all follow the same pattern as `MonoEECallbacks`, so I'm reluctant to change that.  Please don't understand this as a call to change the others.

jaykrell: I was already intending to them them all, just one at a time.
The data really should be readonly.

jaykrell:
@lewurm how about if I ‘g_assert’ equality? It is bigger & repititive but maybe satisfies everyone? I’ll add that change here later. I can macroize it for no repitition.
Maybe only for the “larger” structs?
Also, really, a macro in the header might help. Same macro, for all providers, to init & assert.
Macro is optional implementation detail. Depending on sympathy for the obscure C preprocessor language.
Details-as-actual-code later.

jaykrell:
Two other options presented as PRs.
I realize the current code is also an option.
And there are other options.

jaykrell:
Closing in deference to https://github.com/mono/mono/pull/14717, presumably.

PR #14651, opened 2019-05-26
Introduce g_static_assert.

jaykrell:
ifdef __cplusplus
define g_static_assert(x) static_assert (x, "")
else
define g_static_assert(x) g_assert (x)
endif

It is, granted, C++11, not C++98, and some C variations.


jaykrell:
@monojenkins build failed

PR #14652, opened 2019-05-26
AIX/PASE integration improvements

NattyNarwhal:
A summary of the changes:

* Don't try `shm_open` on PASE, i doesn't support the syscall, and just emits SIGILL. This just creates noise with a corefile and dumps an entry in the SLIC diag logs. This is changeable if PASE supports it in the future.

* Enablement of SysV/SVR4 sonames, which behave like other non-AIX systems, mostly. These are still `ar` archives, but with a predictable name. Extend the loader to deal with these. (It's recommended that you this for `make install` purposes, as DllMaps without the predictable names could be annoying.)

* To go with this, work around issues with these sonames and AIX. You need to specify `OBJECT_MODE=64` in configure so that they properly built; however this presents some difficulty with how autotools and CMake handle detections. GCC on AIX doesn't use this flag, unless the rest of the AIX build tools, so you need a CFLAG, but those aren't propagated. The workaround is to try to put the cflag needed into the CC invocation, which is pretty nasty. Note that PASE does *not* need the workaround, because its GCC package detects the `OBJECT_MODE` variables and acts appropriately. Dumb, but it tries to prevent shooting yourself in the shoot without making it too ugly.

* System.Native is built with a special LDFLAG, otherwise it won't be able to find runtime symbols in an already resident-in-memory Mono when built with SVR4 sonames.

* Make ODBC handled in the configure script and hardcore the 2.x version number - even CentOS uses that. This works around with an issue that PASE had because the `libodbc.so` file is in a devel package, whereas the fully specified version `libodbc.so.2` is in a normal RPM. (This likely blocks on the mono/corefx ODBC patch to actually be useful though.)

Changes include lots of inline comments to highlight asininity.


configure.ac
```diff
@@ -4787,6 +4827,27 @@ jit_status="Building and using the JIT"
...
 case "$host" in
+     *-*-aix*)
+	dnl on AIX/PASE, shared libraries can be and usually are inside archives
+	dnl so, we specify them by libfoo.a(libfoo.so.0) for libtool's conventions,
+	dnl or libfoo.a(shr[_64].o) for the AIX system convention,
+	dnl or lib.so[.0](shr[_64].o) for libtool's hybrid convention
+	dnl we may hardcode 64-bit names at times, but we don't do 32-bit AIX
+	LIBC="libc.a(shr_64.o)"
+	# This is particularily nasty because we need to refer to libintl by version, which may not necessarily match what a system has.
+	INTL="libintl.a(libintl.so.8)"
```
jaykrell: Can any of the behaviors or versions be detected instead of relatively hard coding?

NattyNarwhal: I could try something based on parsing the output of ar, but that seems somewhat flaky. I don't think autotools has a readymade test for this kind of thing.

jaykrell: It seems like it should be easier, but probably anything is ok.

NattyNarwhal:
Sorry, force pushed because git was being dumb after a rebase.

PR #14656, opened 2019-05-27
Add missing new files to Makefile.am, fix tarball build

akoeplinger:
They were added in https://github.com/mono/mono/pull/14427

/cc @CoffeeFlux


akoeplinger:
@monojenkins build deb

PR #14659, opened 2019-05-27
Reduce stack size used by interpreter frame on Windows MSVC.

lateralusX:
Interpreter frame was 1248 bytes in size under Windows MSVC. This was almost double the size of other platforms ~670 bytes. After investigating different constructs in interpreter switch statement in interp_exec_method_full, it turns out that local declared errors couldn't be reused over different case statements and added ~100 bytes each. Switching local declared errors to shared one, reduce the stack size down to ~670 bytes under Windows MSVC.


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

vargaz:
Passing the error down from the caller would probably reduce this even further.

lewurm:
@monojenkins build failed

jaykrell:
MonoError also has more fields than are used, for compat. Maybe can reduce its size.

PR #14667, opened 2019-05-27
[debugger] Update client thread frames after SetIP.

### LINKED ISSUE #13408: [sdb] SetIP requires a resume/suspend to update client thread frames
See attached repro:

[reprosdb3.zip](https://github.com/mono/mono/files/2954674/reprosdb3.zip)

`build.sh && mono debugger.exe`

After a SetIP, we'd like to have the thread frames invalidated and the top frame pointing to the newly set IP where the execution will resume. This will avoid either storing the next IL offset or using a hidden breakpoint and resume to it to properly display the new location in the IDE.

thaystg:
Fixes #13408

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
@@ -168,6 +168,8 @@ namespace Mono.Debugger.Soft
 				throw new ArgumentNullException ("loc");
 			try {
 				vm.conn.Thread_SetIP (id, loc.Method.Id, loc.ILOffset);
+				InvalidateFrames ();
+				FetchFrames (true);
 			} catch (CommandException ex) {
 				if (ex.ErrorCode == ErrorCode.INVALID_ARGUMENT)
 					throw new ArgumentException ("loc doesn't refer to a location in the current method of this thread.", "loc");
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
@@ -168,8 +168,10 @@ namespace Mono.Debugger.Soft
 				throw new ArgumentNullException ("loc");
 			try {
 				vm.conn.Thread_SetIP (id, loc.Method.Id, loc.ILOffset);
-				InvalidateFrames ();
-				FetchFrames (true);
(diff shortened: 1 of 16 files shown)


thaystg:
@jbevain, is this that you are expecting?

jbevain:
@thaystg Thank you for looking into it. That looks good, we'd need to test it.

That also a change of behavior, so we need to bump the protocol version for this. This will allow us to fallback to the previous behavior.

thaystg:
> @thaystg Thank you for looking into it. That looks good, we'd need to test it.
> 
> That also a change of behavior, so we need to bump the protocol version for this. This will allow us to fallback to the previous behavior.

Okay, I'll bump the protocol version. Will you test it before merge? Do you need any help to do it?

lambdageek:
@thaystg can we add a test for this in Mono?  Looks like we have some SetIP tests in dtest.cs already - can we extend it to check that we get updated frames after SetIP.

thaystg:
> @thaystg can we add a test for this in Mono? Looks like we have some SetIP tests in dtest.cs already - can we extend it to check that we get updated frames after SetIP.

Yes!

mono/mini/debugger-agent.c
```diff
@@ -3247,6 +3247,15 @@ compute_frame_info (MonoInternalThread *thread, DebuggerTlsData *tls, gboolean f
 
 	DEBUG_PRINTF (1, "Frames for %p(tid=%lx):\n", thread, (glong)thread->tid);
 
+	if (tls->restore_state.valid && MONO_CONTEXT_GET_IP (&tls->context.ctx) != MONO_CONTEXT_GET_IP (&tls->restore_state.ctx)) {
```
jbevain: Don't we want to do this only if the client supports version `2.52` of the protocol.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ThreadMirror.cs
```diff
@@ -168,6 +168,8 @@ public static void FetchFrames(IList<ThreadMirror> threads)
 				throw new ArgumentNullException ("loc");
 			try {
 				vm.conn.Thread_SetIP (id, loc.Method.Id, loc.ILOffset);
+				InvalidateFrames ();
```
jbevain: Same here, that doesn't look like a no-op, we probably want to do this only if the server supports version `2.52`.

thaystg:
Any news @jbevain ? Should I transform it in a PR?

lewing:
@monojenkins build failed

lambdageek:
Looks like the SetIP test is failing when the interpreter is used.

Issue #14669, opened 2019-05-28
mempool allocate bug?


srxqds:
In the function **mono_mempool_alloc0** why not  reduce **pool->pos** with size if **G_UNLIKELY (pool->pos >= pool->end)** is true?

```c
gpointer
(mono_mempool_alloc0) (MonoMemPool *pool, guint size)
{
	gpointer rval;

	// For the fast path, repeat the first few lines of mono_mempool_alloc
	size = ALIGN_SIZE (size);
	rval = pool->pos;
	pool->pos = (guint8*)rval + size;

	// If that doesn't work fall back on mono_mempool_alloc to handle new chunk allocation
	if (G_UNLIKELY (pool->pos >= pool->end)) {
		rval = mono_mempool_alloc (pool, size);
	}
#ifdef TRACE_ALLOCATIONS
	else if (pool == mono_get_corlib ()->mempool) {
		mono_backtrace (size);
	}
#endif

	memset (rval, 0, size);
	return rval;
}

```

It may a bug of wasting memory?
I think it should add **pool->pos -= size;** after if statement.

```c
if (G_UNLIKELY (pool->pos >= pool->end)) {
                pool->pos -= size;
		rval = mono_mempool_alloc (pool, size);
	}
```

jaykrell:
It's not as bad as it seems.

If the allocation is not large, the memory is going to be lost anyway.

It is only wasted if this is a large `(>= MONO_MEMPOOL_PREFER_INDIVIDUAL_ALLOCATION_SIZE`) allocation but there is room for more small (`< MONO_MEMPOOL_PREFER_INDIVIDUAL_ALLOCATION_SIZE`) allocations.

It does seem to be what was intended though, and the non-0 case does this.

jaykrell:
Even that might not be true. There really might be nothing lost here at all.
But the code could be clearer.

PR #14671, opened 2019-05-28
Cleanup mono_mempool_alloc0 code duplication from mono_mempool_alloc

jaykrell:
per https://github.com/mono/mono/issues/14669


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
I understand the old form was trying to optimize, but I don't think it is worth the source bloat, and the compiler might do the same thing anyway.

PR #14672, opened 2019-05-28
Align mempool like Windows heap -- 2 pointers.

jaykrell:
Doesn't fix any known problem but seems like a good idea.


jaykrell:
@monojenkins build failed

PR #4878, opened 2017-05-16
[android] add hardware specific workaround for Nexus9 in armv7 mode

lewurm:
we got a couple of bug reports, all with the same failure:

https://bugzilla.xamarin.com/show_bug.cgi?id=44907
https://bugzilla.xamarin.com/show_bug.cgi?id=46482
https://bugzilla.xamarin.com/show_bug.cgi?id=51791

The bugs are private, therefore here what I wrote:

> Thank you Matthias, this was very helpful.
> ```
> >> 09-30 12:24:50.347: A/DEBUG(6211): pid: 6191, tid: 6209, name: Thread-4  >>> com.distech.x50.ui.droid <<<
> >> 09-30 12:24:50.347: A/DEBUG(6211): signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8
> >> 09-30 12:24:50.347: A/DEBUG(6211):     r0 d6b58728  r1 00004001  r2 00000000  r3 d6b58738
> >> 09-30 12:24:50.347: A/DEBUG(6211):     r4 00000008  r5 ea97427c  r6 00001f3c  r7 00000015
> >> 09-30 12:24:50.347: A/DEBUG(6211):     r8 00001f38  r9 00000015  sl d6b58728  fp d67fed90
> >> 09-30 12:24:50.347: A/DEBUG(6211):     ip ea9742c0  sp d67fed60  lr ea7a4504  pc ea716a58  cpsr 800e0010
> ```
> The crash happens on this assignment:
> https://github.com/mono/mono/blob/de1865dad5c0350f391fedcaa08f02f610530d3f/mono/mini/mini-generic-sharing.c#L418
> Here the according disassembly:
> https://gist.github.com/lewurm/b1094749027c9e5ea19fdc4fac7905a7
> 
> The crash happens in the last loop iteration (`r9=i`, `r7=slot`).  Looking
> at the machine code it just _cannot_ happen, which is confirmed by the C
> code as well: `*oti` successfully happens in the if check, but after
> returning from `alloc_oti()`, `oti` doesn't contain a valid address anymore.
> I suspect some weird hardware issue that fails to restore all registers
> properly from the stack.
> 
> [...]
> 
> Thanks again Matthias.  Unfortunately, I'm out of ideas, and I can't
> blame anything but the hardware. The situation we see is too weird. We
> segfault at offset `0xdda60: str sl, [r4]`, but really we should
[...]

PR #14691, opened 2019-05-29
Revert "[android] add hardware specific workaround for Nexus9 in armv7 mode"

lewurm:
This reverts commit ee90fc601e762dc31eeb90fc61b6305246ad5275 (also see https://github.com/mono/mono/pull/4878 ).

The device was released November 2014.  I think it's fair to assume it isn't widely used anymore.

### DIFF
--- a/mono/mini/mini-generic-sharing.c
+++ b/mono/mini/mini-generic-sharing.c
@@ -407,17 +407,7 @@ info_has_identity (MonoRgctxInfoType info_type)
 /*
  * LOCKING: loader lock
  */
-#if defined(HOST_ANDROID) && defined(TARGET_ARM)
-/* work around for HW bug on Nexus9 when running on armv7 */
-#ifdef __clang__
-static __attribute__ ((optnone)) void
-#else
-/* gcc */
-static __attribute__ ((optimize("O0"))) void
-#endif
-#else
 static void
-#endif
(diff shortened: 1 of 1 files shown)


jaykrell:
Obnoxious/curious/devil's-advocate: I wonder if gcc/clang workaround this and how late such bugs shipped (not just when they first shipped).


It looks like:
unsigned i;
I = (i << k) >> k;

is what the processor didn't handle, to clear upper k bits.
workaround I gather is to change the second operation to a rotate which portable C lacks.
The first shift will have produced the desired zeros, so then rotate and shift are the same.
Unclear what is the portable C workaround. Just turning off optimizations wouldn't seem to reliably avoid it. Maybe it is reading k from memory between operations?

lewurm:
@jaykrell I don't know where you see this `(i<< k) >> k` pattern in the affected code. Please have a look at the original PR, the theory back then was that the post-increment variant of `ldr` didn't execute correctly.

jaykrell:
It was from following some of the links. Agreed not evident here. Could be multiple bugs.

lewing:
@jonpryor do we still need this?

steveisok:
I say take it.

lewing:
@lewurm rebase this to force retesting?

lewurm:
@monojenkins squash

lewurm:
@monojenkins build failed

lewurm:

PR #14692, opened 2019-05-29
Read only function pointers (MonoThreadInfoRuntimeCallbacks).

jaykrell:
And inline getting the pointer.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Better now? A form of the other is less objected to. The order dependencies have been partially mitigated.

jaykrell:
Better? Order sensitivity mostly gone. Readonly is kinda valuable and typical.

mono/dis/main.c
```diff
@@ -1984,7 +1984,16 @@ thread_state_init (MonoThreadUnwindState *ctx)
...
+	static const MonoThreadInfoRuntimeCallbacks ticallbacks = {
+		NULL, // setup_async_callback
+		NULL, // thread_state_init_from_sigctx
+		NULL, // thread_state_init_from_handle
+		thread_state_init,
+	};
+	g_assert (ticallbacks.setup_async_callback == NULL &&
+		  ticallbacks.thread_state_init_from_sigctx == NULL &&
+		  ticallbacks.thread_state_init_from_handle == NULL &&
+		  ticallbacks.thread_state_init == thread_state_init);
```
BrzVlad: This assert looks like overkill

mono/utils/mono-threads.h
```diff
@@ -311,6 +311,13 @@ typedef struct {
 	void (*thread_state_init) (MonoThreadUnwindState *tctx);
 } MonoThreadInfoRuntimeCallbacks;
 
+#define MONO_INIT_THREAD_INFO_RUNTIME_CALLBACKS(prefix) { \
+	prefix ## _setup_async_callback,                  \
+	prefix ## _thread_state_init_from_sigctx,         \
+	prefix ## _thread_state_init_from_handle,         \
+	prefix ## _thread_state_init,                     \
+};
```
BrzVlad: I don't like these defines since they make looking for methods harder. Can't we just initialize the callbacks normally since it is used with only one prefix ?

PR #14694, opened 2019-05-29
Fix LLVM build break on Windows MSVC.

lateralusX:
https://github.com/mono/mono/pull/14607 broke LLVM build on Windows MSVC. This is just a quick fix falling back to old behavior on Windows MSVC. Since all LLVM tools will be available in the same folder as mono runtime binary there is no specific need to point out other build folders hosting LLVM tooling.


lambdageek:
@monojenkins build failed

lateralusX:
@jenkins rebuild failed

lateralusX:
@monojenkins rebuild failed

lateralusX:
@monojenkins rebuild failed

PR #14696, opened 2019-05-29
[netcore] Download and use local .NET Core SDK

EgorBo:
1) move "build System.Private.CorLib" logic to netcore/Makefile
2) download .net core runtime and SDK via `dotnet-install.sh` (TODO: windows) so we don't depend on system wide sdk anymore. The SDK will be located in `repo-root/.dotnet`

We can now also use `repo-root/.dotnet` folder to copy our runtime and System.Private.Corlib there to run tests etc (so we don't need to download a runtime in `netcore` folder)

/cc: @filipnavara

So in order to build a nupkg with runtime and mscorlib:
```
cd netcore
./build.sh -c Release -pack

or:

./autogen.sh --with-core=only
make -C netcore runtime -j10
make -C netcore bcl
make -C netcore nupkg
```


netcore/Makefile
```diff
@@ -47,6 +48,13 @@ TEST_ASSETS_URL = $(FEED_BASE_URL)/corefx-tests/$(NETCORETESTS_VERSION)/$(TESTS_
 BLAME = $(shell git blame ../configure.ac | grep AC_INIT | cut -f1 -d' ')
 VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
 
+all: runtime bcl
+
+LOCAL_DOTNET = ../.dotnet/dotnet
+$(LOCAL_DOTNET):
+	bash ../eng/common/dotnet-install.sh -v $(NETCOREAPP_VERSION) && \
```
steveisok: If we do something like:

```bash
bash -l -c ". ../eng/common/tools.sh && InitializeToolset"
```

We'll be running the same tools install as corefx does.

filipnavara:
I do like the approach in general but there was previously some opposition to it. It would be nice to ensure that `make runtime bcl xtest-<something>` still works. As far as I can see that's not the case now since the generated binaries are no longer copied to the runtime used for tests. It could likely be solved by adding `link-mono` dependency to `xtest*` targets.

EgorBo:
@filipnavara `prepare` rule is appended to `xtest*` and `run-sample`
so
```
make bcl
make runtime
make xtest-System.Runtime.Tests
```
will run the test using the new binaries

filipnavara:
Ah, I missed that dependency :-)

directhex:
We will need to pass a `-a` or `-architecture` flag, because this method ultimately uses https://dot.net/v1/dotnet-install.sh which suffers from https://github.com/microsoft/azure-pipelines-tasks/issues/10439 on ARM

directhex:
There are also changes to make to scripts/ci/pipeline-netcore-runtime.yml to make it stop trying to install its own dotnet in addition to here

netcore/Makefile
steveisok: Never-mind, it seems like there's some unknowns cherry picking arcade functions.

netcore/Makefile
```diff
@@ -47,6 +51,16 @@ TEST_ASSETS_URL = $(FEED_BASE_URL)/corefx-tests/$(NETCORETESTS_VERSION)/$(TESTS_
 BLAME = $(shell git blame ../configure.ac | grep AC_INIT | cut -f1 -d' ')
 VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
 
+all: runtime bcl
+
+../.dotnet/dotnet:
+	bash ../eng/common/dotnet-install.sh -v $(NETCOREAPP_VERSION)
+	bash ../.dotnet/dotnet-install.sh  -c master -i ../.dotnet
```
marek-safar: I am not sure this is the right way doing it, it'll install "random" master version. Shouldn't DARC sync take care of this dependency?

@ViktorHofer any inputs on that?

ViktorHofer: As discussed with @steveisok offline, you should not clal into `eng/common/dotnet-install.sh` but into the InitializeToolset function in tools.sh: `. ../eng/common/tools.sh && InitializeToolset`. This should install the version in the global.json file.

steveisok: @ViktorHofer, I ran into some additional issues I'd like @EgorBo to help validate.  Depending on the outcome, we can discuss at our weekly meeting.

EgorBo: @ViktorHofer  I am trying to call that function but getting:
```
/prj/mono/mono5/artifacts/toolset/restore.proj : error : Unable to find package Microsoft.DotNet.Arcade.Sdk with version (= 1.0.0-beta.19274.6)
/prj/mono/mono5/artifacts/toolset/restore.proj : error :   - Found 1422 version(s) in dotnet-core [ Nearest version: 1.0.0-dev.18424.13 ]
/prj/mono/mono5/artifacts/toolset/restore.proj : error :   - Found 0 version(s) in nuget.org
/prj/mono/mono5/artifacts/toolset/restore.proj : error MSB4236: The SDK 'Microsoft.DotNet.Arcade.Sdk' specified could not be found.
```
is there an additional nuget feed I should add to nuget.config?

ViktorHofer: > UPD updated version to 1.0.0-dev.18424.13 in the global.json

 ?

ViktorHofer: If `1.0.0-beta.19274.6` doesn't work, try it again with `1.0.0-beta.19278.1`. Weird...

EgorBo: @ViktorHofer both don't work for me, I have no idea why, I do see these versions in my Nuget Explorer

EgorBo: @ViktorHofer ah, `dotnet nuget locals all --clear` helped

ViktorHofer: To quickly validate if a package exists in a feed you can use https://dotnetfeed.blob.core.windows.net/dotnet-core/sleet.packageindex.json

netcore/Makefile
```diff
@@ -53,13 +56,8 @@ VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
...
 
-../.dotnet/dotnet:
-	bash ../eng/common/dotnet-install.sh -v $(NETCOREAPP_VERSION)
-	bash ../.dotnet/dotnet-install.sh  -c master -i ../.dotnet
-
-../.dotnet/dotnet.exe:
-	powershell -ExecutionPolicy ByPass -NoProfile -command ../eng/common/dotnet-install.ps1 -version $(NETCOREAPP_VERSION)
-	powershell -ExecutionPolicy ByPass -NoProfile -command ../.dotnet/dotnet-install.ps1 -c master -i ../.dotnet
+init-tools:
+	$(SCRIPT_EXECUTOR) init-tools.$(SCRIPT_EXTENSION)
```
marek-safar: Do we still need roslyn restore?

directhex:
https://github.com/mono/mono/blob/master/scripts/ci/pipeline-netcore-runtime.yml#L73 - add `/c/windows/System32/WindowsPowerShell/v1.0/` to PATH here to fix VSTS failure

directhex:
Also, remove the dotnet-sdk package from https://github.com/mono/mono/blob/master/scripts/ci/pipeline-netcore-runtime.yml#L39

And this entire step: https://github.com/mono/mono/blob/master/scripts/ci/pipeline-netcore-runtime.yml#L63-L69

netcore/Makefile
```diff
@@ -61,11 +73,11 @@ $(ASPNETCORESDK_FILE):
 
 update-corefx: corefx/.stamp-dl-corefx-$(NETCORETESTS_VERSION)
 
-corefx/.stamp-dl-corefx-$(NETCORETESTS_VERSION): corefx-restore.csproj
+corefx/.stamp-dl-corefx-$(NETCORETESTS_VERSION): corefx-restore.csproj init-tools
 	dotnet build corefx-restore.csproj --runtime $(RID) --packages corefx/packages -p:MicrosoftPrivateCoreFxNETCoreAppVersion=$(NETCORETESTS_VERSION) -p:OutputPath=corefx/restore/
```
steveisok: This will refer to the system wide dotnet since the call to init-tools.sh is in another bash process

netcore/Makefile
```diff
@@ -90,27 +102,28 @@ run-aspnet-sample: prepare
...
 
 link-mono:
 	cp ../mono/mini/.libs/libmonosgen-2.0$(PLATFORM_AOT_SUFFIX) $(SHAREDRUNTIME)/$(PLATFORM_AOT_PREFIX)coreclr$(PLATFORM_AOT_SUFFIX)
 	cp ../mcs/class/System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.{dll,pdb} $(SHAREDRUNTIME)
 
-prepare: $(ASPNETCORESDK_FILE) $(NETCORESDK_FILE) update-corefx update-roslyn link-mono
+prepare: init-tools $(ASPNETCORESDK_FILE) $(NETCORESDK_FILE) update-corefx update-roslyn link-mono
 
-nupkg:
+nupkg: all
```
directhex: This will enforce a GCC build of the runtime on Windows (not a big deal, but it will slow things down)

netcore/Makefile
directhex: environment is not preserved from init-tools, so this will use dotnet from PATH. Same for all raw calls to `dotnet` - we need to fully qualify those to use the path from `../dotnet/`

EgorBo: should be fixed now I guess

netcore/Makefile
EgorBo: ok makes sense

netcore/Makefile
```diff
@@ -88,12 +81,12 @@ corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
...
+	$(DOTNET) build sample/HelloWorld
 	MONO_ENV_OPTIONS="--debug" COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(NETCOREAPP_VERSION)" sample/HelloWorld/bin/Debug/netcoreapp3.0/HelloWorld.dll
 
 run-aspnet-sample: prepare
 	rm -rf sample/AspNetCore/{bin,obj}
-	dotnet publish sample/AspNetCore -c Debug -r $(RID)
+	$(DOTNET) publish sample/AspNetCore -c Debug -r $(RID)
 	cp ../mono/mini/.libs/libmonosgen-2.0$(PLATFORM_AOT_SUFFIX) sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/$(PLATFORM_AOT_PREFIX)coreclr$(PLATFORM_AOT_SUFFIX)
 	cp ../mcs/class/System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.dll sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/
 	COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(ASPNETCOREAPP_VERSION)" sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/AspNetCore.dll
```
steveisok: put $(DOTNET) here

netcore/Makefile
```diff
@@ -88,12 +81,12 @@ corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
 	touch $@
 
 run-sample: prepare
-	dotnet build sample/HelloWorld
+	$(DOTNET) build sample/HelloWorld
 	MONO_ENV_OPTIONS="--debug" COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(NETCOREAPP_VERSION)" sample/HelloWorld/bin/Debug/netcoreapp3.0/HelloWorld.dll
```
steveisok: Put $(DOTNET) here

netcore/Makefile
```diff
@@ -185,11 +178,11 @@ check-env:
...
-	cd gen-xunit-runner && dotnet run ../tests $(COREFX_ROOT) $(COREFX_BINDIR)/$*/$($*_PROFILE)/$*.dll $(XUNIT_FLAGS) @../CoreFX.issues.rsp $(XUNIT_ARGS)
+	cd gen-xunit-runner && $(DOTNET) run ../tests $(COREFX_ROOT) $(COREFX_BINDIR)/$*/$($*_PROFILE)/$*.dll $(XUNIT_FLAGS) @../CoreFX.issues.rsp $(XUNIT_ARGS)
 	cp $(COREFX_BINDIR)/RemoteExecutorConsoleApp/netstandard-Debug/RemoteExecutorConsoleApp.* tests/$*
 	cp $(COREFX_BINDIR)/System.Runtime.Tests/netcoreapp-Unix-Debug/TestLoadAssembly.dll tests/$*
 	sed -e 's/9.9.9/$(VERSION)/g' < tests/$*/RemoteExecutorConsoleApp.runtimeconfig.json > 2 && mv 2 tests/$*/RemoteExecutorConsoleApp.runtimeconfig.json
-	cd tests/$* && dotnet build
+	cd tests/$* && $(DOTNET) build
 
 run-%: check-env
 	cd tests/$* && MONO_PATH=bin/Debug/netcoreapp3.0 MONO_ENV_OPTIONS="--debug --explicit-null-checks" COMPlus_DebugWriteToStdErr=1 ../../dotnet bin/Debug/netcoreapp3.0/$*-runner.dll
```
steveisok: put $(DOTNET) here instead of ../../dotnet

netcore/Makefile
```diff
@@ -76,12 +81,12 @@ corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
 	touch $@
 
 run-sample: prepare
-	dotnet build sample/HelloWorld
+	$(DOTNET) build sample/HelloWorld
 	MONO_ENV_OPTIONS="--debug" COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(NETCOREAPP_VERSION)" sample/HelloWorld/bin/Debug/netcoreapp3.0/HelloWorld.dll
```
steveisok: put $(DOTNET) here in replace of ./dotnet

EgorBo: @steveisok `./dotnet` refers to a local runtime we download and where we replace clrjit and mscorlib. I guess we will get rid of it eventually but not now

netcore/Makefile
```diff
@@ -76,12 +81,12 @@ corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
...
 
 run-aspnet-sample: prepare
 	rm -rf sample/AspNetCore/{bin,obj}
-	dotnet publish sample/AspNetCore -c Debug -r $(RID)
+	$(DOTNET) publish sample/AspNetCore -c Debug -r $(RID)
 	cp ../mono/mini/.libs/libmonosgen-2.0$(PLATFORM_AOT_SUFFIX) sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/$(PLATFORM_AOT_PREFIX)coreclr$(PLATFORM_AOT_SUFFIX)
 	cp ../mcs/class/System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.dll sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/
 	COMPlus_DebugWriteToStdErr=1 ./dotnet --fx-version "$(ASPNETCOREAPP_VERSION)" sample/AspNetCore/bin/Debug/netcoreapp3.0/$(RID)/publish/AspNetCore.dll
```
steveisok: put $(DOTNET) here instead of ./dotnet

netcore/Makefile
steveisok: Ok... I was assuming we would need to refer to local dotnet across the board.

EgorBo: $(DOTNET) is a complete SDK+Runtime with version specified in global.json.
./dotnet is a runtime with version = NETCOREAPP_VERSION

netcore/Makefile
```diff
@@ -5,35 +5,38 @@ include ../mcs/build/config.make
...
 
-include roslyn.make
+ROSLYN_VERSION := 3.2.0-beta3-19279-03
 
 ifeq ($(HOST_PLATFORM),win32)
 NETCORESDK_EXT = zip
 UNZIPCMD = python -c "import zipfile,sys; zipfile.ZipFile(sys.argv[1], 'r').extractall()"
 XUNIT_FLAGS =
 TESTS_PLATFORM = Windows_NT.x64
+DOTNET = $(shell .\init-tools.ps1)
```
directhex: Make shells out to _bash_, so this should be in a format which MSYS2 bash will work with (e.g. maybe just `./init-tools.ps1`, otherwise the full path to `/foo/bar/powershell init-tools.ps1`?)

Right now it fails with:
```
make: Entering directory 'D:/a/1/s/netcore'
make: Makefile:74: pipe: No such file or directory
restore roslyn-restore.csproj -p:RoslynVersion=3.2.0-beta3-19279-03 --packages roslyn/packages -p:OutputPath=roslyn/restore/
restore: Unknown option -p:RoslynVersion=3.2.0-beta3-19279-03. Try restore --help for more info.
make: *** [Makefile:74: update-roslyn] Error 1
```

EgorBo: Windows support is not finished yet, `init-tools.ps1` is not even commited 🙂 but it all should work on macOS/Linux

directhex: Yes, OSX and Linux are building.

EgorBo: Windows should now work too.

eng/common/tools.ps1
```diff
@@ -93,10 +93,6 @@ function Exec-Process([string]$command, [string]$commandArgs) {
 }
 
 function InitializeDotNetCli([bool]$install) {
-  if (Test-Path variable:global:_DotNetInstallDir) {
```
marek-safar: Won't this be overwritten?

EgorBo: @marek-safar I don't know why but with this line if I change version in `global.json` - it simply returns old version and does nothing. And don't worry without this like it won't download SDK every hit.

EgorBo: cc @ViktorHofer

EgorBo: the .sh version works without changes as expected.

marek-safar: I guess you could upstream this change

directhex:
`/usr/bin/sh: powershell: command not found`

see https://github.com/mono/mono/pull/14696#issuecomment-497361466

@EgorBo

eng/common/tools.ps1
EgorBo: But not sure how, afair the whole eng/** folder is autogenerated (by darc?)

EgorBo:
@directhex oh, forgot about it. Thanks

eng/common/tools.ps1
marek-safar: I think it lives in the normal repo at https://github.com/dotnet/arcade

EgorBo:
Oops accidentally merged my Windows 10SDK/v142 changes, reverting...

directhex:
:shipit:

EgorBo:
@marek-safar done.

ViktorHofer:
I'll review this today.

EgorBo:
@monojenkins build failed

scripts/ci/pipeline-netcore-runtime.yml
```diff
@@ -60,20 +60,13 @@ jobs:
...
-
   - script: |
       %CD:~0,2%\msys64\usr\bin\bash -lc "PATH=/usr/bin:/mingw64/bin:/c/hostedtoolcache/windows/dncs/2.2.203/x64:/c/Program\ Files/Git/bin ./autogen.sh --host=x86_64-w64-mingw32 --with-core=only"
       CALL msvc\run-msbuild.bat build x86_64 release "/p:PlatformToolset=v140 /p:MONO_TARGET_GC=sgen"
-      %CD:~0,2%\msys64\usr\bin\bash -lc "PATH=/usr/bin:/mingw64/bin:/c/hostedtoolcache/windows/dncs/2.2.203/x64:/c/Program\ Files/Git/bin make -C netcore bcl"
+      %CD:~0,2%\msys64\usr\bin\bash -lc "PATH=/c/windows/System32/WindowsPowerShell/v1.0/:/usr/bin:/mingw64/bin:/c/hostedtoolcache/windows/dncs/2.2.203/x64:/c/Program\ Files/Git/bin make -C netcore bcl"
       MKDIR mono\mini\.libs\
       COPY msvc\build\sgen\x64\bin\Release\mono-2.0-sgen.dll mono\mini\.libs\monosgen-2.0.dll
-      %CD:~0,2%\msys64\usr\bin\bash -lc "PATH=/usr/bin:/mingw64/bin:/c/hostedtoolcache/windows/dncs/2.2.203/x64:/c/Program\ Files/Git/bin make -C netcore nupkg"
+      %CD:~0,2%\msys64\usr\bin\bash -lc "PATH=/c/windows/System32/WindowsPowerShell/v1.0/:/usr/bin:/mingw64/bin:/c/hostedtoolcache/windows/dncs/2.2.203/x64:/c/Program\ Files/Git/bin make -C netcore nupkg"
```
ViktorHofer: is this necessary? I thought that powershell is always in the path on Windows.

directhex: It's in Windows' path, but not in MSYS2's path - so anything inside our makefiles which calls `powershell` fails

netcore/build.sh
```diff
@@ -0,0 +1,76 @@
...
+
+CPU_COUNT=$(getconf _NPROCESSORS_ONLN || echo 4)
+
+# build mono runtime
+make runtime -j$(CPU_COUNT)
+
+# build System.Private.CoreLib (../mcs/class/System.Private.CoreLib)
+make bcl CORLIB_BUILD_FLAGS="$properties"
+
+make nupkg
```
akoeplinger: shouldn't this be under an if that checks whether `pack` is true?

EgorBo: Oops, fixed. thanks!

EgorBo:

EgorBo:

netcore/Makefile
```diff
@@ -47,6 +51,8 @@ TEST_ASSETS_URL = $(FEED_BASE_URL)/corefx-tests/$(NETCORETESTS_VERSION)/$(TESTS_
 BLAME = $(shell git blame ../configure.ac | grep AC_INIT | cut -f1 -d' ')
 VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
 
+all: runtime bcl
```
alexanderkyte: What is going to trigger this line? If we're expecting it to run because it's the first target, can we move it higher in the file or comment something to that effect. Otherwise a comment describing what uses the "all" entrypoint would be helpful for future navigators.

PR #14704, opened 2019-05-29
[crash] Move /proc/self/maps crash collection to created JSON blob

alexanderkyte:
Previously, we would simply print the mapping to stdout. 

Since the pointers that are embedded into crash reports will point into these native modules, providing the mapping to module names helps someone debugging using the JSON to track down where a pointer came from.

Example report: https://gist.github.com/alexanderkyte/578b4e2f82a0e71a169c37d8bc18a305

Because of the fact that this is sending file paths from the end-user's machine, I have disabled it when we have crash privacy enabled.

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -1123,6 +1123,12 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
 #else
 #define g_write write
 #endif
+#ifdef G_OS_WIN32
+#define g_read _read
+#else
+#define g_read read
+#endif
+
 #define g_fopen fopen
 #define g_lstat lstat
 #define g_rmdir rmdir
@@ -1133,10 +1139,33 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
 
 gchar *g_mkdtemp (gchar *tmpl);
 
-
 /*
  * Low-level write-based printing functions
  */
+
+static inline int
+g_async_safe_fgets (char *str, int num, int handle, gboolean *newline)
+{
+	memset (str, '\0', sizeof(char) * num);
+	// Make sure we always have null-termination
+	int without_padding = num - 1;
+	int i=0;
+	char scratch [2];
+	while (i < without_padding && g_read (handle, scratch, sizeof(char))) {
+		str [i] = scratch [0];
+		if (str [i] == '\n') {
+			str [i] = '\0';
+			*newline = TRUE;
+		}
+
+		if (str [i] == '\0')
+			break;
+		i++;
+	}
+
+	return i;
+}
+
 static inline gint
 g_async_safe_vfprintf (int handle, gchar const *format, va_list args)
 {
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -1146,7 +1146,7 @@ gchar *g_mkdtemp (gchar *tmpl);
 static inline int
 g_async_safe_fgets (char *str, int num, int handle, gboolean *newline)
 {
-	memset (str, '\0', sizeof(char) * num);
+	memset (str, 0, num);
 	// Make sure we always have null-termination
 	int without_padding = num - 1;
 	int i=0;
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -1157,9 +1157,13 @@ g_async_safe_fgets (char *str, int num, int handle, gboolean *newline)
 			str [i] = '\0';
 			*newline = TRUE;
 		}
+		
+		if (!isprint (str [i]))
+			str [i] = '\0';
 
 		if (str [i] == '\0')
 			break;
+
 		i++;
 	}
 
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -1147,12 +1147,11 @@ static inline int
 g_async_safe_fgets (char *str, int num, int handle, gboolean *newline)
 {
 	memset (str, 0, num);
-	// Make sure we always have null-termination
+	// Make sure we don't overwrite the last index so that we are
+	// guaranteed to be NULL-terminated
 	int without_padding = num - 1;
 	int i=0;
-	char scratch [2];
-	while (i < without_padding && g_read (handle, scratch, sizeof(char))) {
-		str [i] = scratch [0];
+	while (i < without_padding && g_read (handle, &str [i], sizeof(char))) {
 		if (str [i] == '\n') {
 			str [i] = '\0';
 			*newline = TRUE;
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -908,7 +908,7 @@ dump_memory_around_ip (MonoContext *mctx)
 static void
 print_process_map (void)
 {
-#ifdef __linux__
+#if defined(__linux__) && !defined(HOST_ANDROID)
 	FILE *fp = fopen ("/proc/self/maps", "r");
 	char line [256];
 
--- a/mono/mini/mini-posix.c
+++ b/mono/mini/mini-posix.c
@@ -906,37 +906,6 @@ dump_memory_around_ip (MonoContext *mctx)
 }
 
 static void
-print_process_map (void)
-{
-#if defined(__linux__) && !defined(HOST_ANDROID)
-	FILE *fp = fopen ("/proc/self/maps", "r");
-	char line [256];
-
-	if (fp == NULL) {
-		mono_runtime_printf_err ("no /proc/self/maps, not on linux?\n");
-		return;
-	}
-
-	mono_runtime_printf_err ("/proc/self/maps:");
-	const int max_lines = 25;
-	int i = 0;
-
-	while (fgets (line, sizeof (line), fp) && i++ < max_lines) {
-		// strip newline
-		size_t len = strlen (line);
-		if (len > 0 && line [len - 1] == '\n')
-			line [len - 1] = '\0';
-
-		mono_runtime_printf_err ("%s", line);
-	}
-
-	fclose (fp);
-#else
-	/* do nothing */
-#endif
-}
-
-static void
 assert_printer_callback (void)
 {
 	mono_dump_native_crash_info ("SIGABRT", NULL, NULL);
@@ -1153,8 +1122,6 @@ dump_native_stacktrace (const char *signal, MonoContext *mctx)
 void
 mono_dump_native_crash_info (const char *signal, MonoContext *mctx, MONO_SIG_HANDLER_INFO_TYPE *info)
 {
-	print_process_map ();
-
 	dump_native_stacktrace (signal, mctx);
 
 	dump_memory_around_ip (mctx);
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -933,6 +933,15 @@ static void
 mono_native_state_add_process_map (MonoStateWriter *writer)
 {
 #if defined(__linux__) && !defined(HOST_ANDROID)
+	static gint32 handle_in_use;
+	gint32 old = mono_atomic_cas_i32 (&handle_in_use, TRUE /* set */, FALSE /* compare */);
+	if (old != FALSE) {
+		g_async_safe_printf ("Warning: Couldn't open /proc/self/maps for concurrent reads, crash report will lack it.\n");
+		return;
+	}
+
+	mono_memory_barrier ();
+
 	int handle = g_open ("/proc/self/maps", O_RDONLY, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
 	if (handle == -1) {
 		g_async_safe_printf ("Couldn't find /proc/self/maps on Linux system. Continuing.");
@@ -981,6 +990,9 @@ mono_native_state_add_process_map (MonoStateWriter *writer)
 	mono_state_writer_printf(writer, "],\n");
 
 	close (handle);
+
+	mono_atomic_store_i32 (&handle_in_use, FALSE);
+	mono_memory_barrier ();
 #endif
 }
 
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -952,12 +952,17 @@ mono_native_state_add_process_map (MonoStateWriter *writer)
 		mono_state_writer_printf (writer, "\t\"");
 
 		while (TRUE) {
-			char line [25];
+			char line [10];
 			gboolean newline = FALSE;
 			int charsCopied = g_async_safe_fgets (line, sizeof (line), handle, &newline);
 
 			if (charsCopied == 0)
 				break;
+
+			for (int i=0; i < charsCopied; i++)
+				g_assert (isprint (line [i]));
+
+			g_assert (line [charsCopied] == '\0');
 
 			mono_state_writer_printf (writer, "%s", line);
 
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -933,15 +933,6 @@ static void
 mono_native_state_add_process_map (MonoStateWriter *writer)
 {
 #if defined(__linux__) && !defined(HOST_ANDROID)
-	static gint32 handle_in_use;
-	gint32 old = mono_atomic_cas_i32 (&handle_in_use, TRUE /* set */, FALSE /* compare */);
-	if (old != FALSE) {
-		g_async_safe_printf ("Warning: Couldn't open /proc/self/maps for concurrent reads, crash report will lack it.\n");
-		return;
-	}
-
-	mono_memory_barrier ();
-
 	int handle = g_open ("/proc/self/maps", O_RDONLY, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
 	if (handle == -1) {
 		g_async_safe_printf ("Couldn't find /proc/self/maps on Linux system. Continuing.");
@@ -990,9 +981,6 @@ mono_native_state_add_process_map (MonoStateWriter *writer)
 	mono_state_writer_printf(writer, "],\n");
 
 	close (handle);
-
-	mono_atomic_store_i32 (&handle_in_use, FALSE);
-	mono_memory_barrier ();
 #endif
 }
 
--- a/mono/utils/mono-state.c
+++ b/mono/utils/mono-state.c
@@ -927,6 +927,58 @@ mono_native_state_add_memory (MonoStateWriter *writer)
 	mono_state_writer_printf(writer, "},\n");
 }
 
+#define MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT 30
+
+static void
+mono_native_state_add_process_map (MonoStateWriter *writer)
+{
+#if defined(__linux__) && !defined(HOST_ANDROID)
+	int handle = g_open ("/proc/self/maps", O_RDONLY, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
+	if (handle == -1) {
+		g_async_safe_printf ("Couldn't find /proc/self/maps on Linux system. Continuing.");
+		return;
+	}
+
+	assert_has_space (writer);
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "process_map");
+	mono_state_writer_printf(writer, "[\n");
+
+	int mapping = 0;
+	while (mapping < MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT) {
+		if (mapping > 0)
+			mono_state_writer_printf (writer, "\",\n");
+
+		mono_state_writer_printf (writer, "\t\"");
+
+		while (TRUE) {
+			char line [25];
+			gboolean newline = FALSE;
+			int charsCopied = g_async_safe_fgets (line, sizeof (line), handle, &newline);
+
+			if (charsCopied == 0)
+				break;
+
+			mono_state_writer_printf (writer, "%s", line);
+
+			if (newline)
+				break;
+		}
+
+		mapping++;
+	}
+
+	if (mapping > 0)
+		mono_state_writer_printf (writer, "\"");
+
+	mono_state_writer_indent (writer);
+	writer->indent--;
+	mono_state_writer_printf(writer, "],\n");
+
+	close (handle);
+#endif
+}
+
 static void
 mono_native_state_add_prologue (MonoStateWriter *writer)
 {
@@ -960,6 +1012,10 @@ mono_native_state_add_prologue (MonoStateWriter *writer)
 
 		mono_state_writer_printf(writer, "\"%.*s\",\n", (int)length, assertion_msg);
 	}
+
+#ifndef MONO_PRIVATE_CRASHES
+	mono_native_state_add_process_map (writer);
+#endif
 
 	// Start threads array
 	assert_has_space (writer);
--- a/mono/eglib/eglib-remap.h
+++ b/mono/eglib/eglib-remap.h
@@ -133,6 +133,7 @@
 #define g_pattern_match_string monoeg_g_pattern_match_string
 #define g_pattern_spec_free monoeg_g_pattern_spec_free
 #define g_pattern_spec_new monoeg_g_pattern_spec_new
+#define g_async_safe_fgets monoeg_g_async_safe_fgets
 #define g_async_safe_fprintf monoeg_g_async_safe_fprintf
 #define g_async_safe_vfprintf monoeg_g_async_safe_vfprintf
 #define g_async_safe_printf monoeg_g_async_safe_printf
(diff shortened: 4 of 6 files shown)


alexanderkyte:
Should accomplish what you wanted @grendello

baulig:
What's the rationale behind this?

alexanderkyte:
@monojenkins build failed

alexanderkyte:
Discussing it with @grendello , may not want to merge.

alexanderkyte:
It sounds like this doesn't print useful info on android (prints the process group that doesn't contain Mono), and if it's not helpful then there's a stability price to running something like that along the crash reporting critical path. So there's an argument to merge. 

But it doesn't sound like there was a compelling crash that motivated it. So I think in the end, it's just a question of whether XA wants it.

lewurm:
We should disable it everywhere, not just Android.

Having the first 25 lines of `/proc/self/maps` isn't very useful anyway. Dumping the whole file is too noisy. Ideally we would add this information to the JSON file created by the crash reporter.

alexanderkyte:
I can just rip out print_process_map if nobody wants it?

@lewurm @vargaz @marek-safar @lambdageek ?

mono/mini/mini-posix.c
```diff
@@ -908,7 +908,7 @@ dump_memory_around_ip (MonoContext *mctx)
 static void
 print_process_map (void)
 {
-#ifdef __linux__
+#if defined(__linux__) && !defined(HOST_ANDROID)
 	FILE *fp = fopen ("/proc/self/maps", "r");
```
jaykrell: It could be open/read/close instead of fopen/fgets/fclose, right?
You just have to do the newline scanning and it is pretty easy.
You don't have to output a line at a time, you just count the newlines and truncate the last output when you saw enough.

marek-safar:
I agree that dumping only part of that information is not that useful

jaykrell:
Isn’t the main point to dump especially mono/libmono’s range, so that offsets can be computed & therefore symbols?

lewurm:
@alexanderkyte could you dump `/proc/self/maps` into the JSON file instead?  I know it's not a thing yet on Android, but eventually we'll get there I assume?

@jaykrell it depends. libmono address range definitely. sometimes heap. sometimes third party library that crashes.

alexanderkyte:
@lewurm 

Given that the output looks like:

```
[mono-rt] 12c00000-42c00000 rw-p 00000000 00:01 7005       /dev/ashmem/dalvik-main space (region space) (deleted)
[mono-rt] 6f1e2000-6f3c8000 rw-p 00000000 fc:00 7372       /data/dalvik-cache/x86/system@framework@boot.art
[mono-rt] 6f3c8000-6f485000 rw-p 00000000 fc:00 7373       /data/dalvik-cache/x86/system@framework@boot-core-libart.art
[mono-rt] 6f485000-6f4c2000 rw-p 00000000 fc:00 7374       /data/dalvik-cache/x86/system@framework@boot-conscrypt.art
[mono-rt] 6f4c2000-6f4ea000 rw-p 00000000 fc:00 7375       /data/dalvik-cache/x86/system@framework@boot-okhttp.art
[mono-rt] 6f4ea000-6f4ee000 rw-p 00000000 fc:00 7376       /data/dalvik-cache/x86/system@framework@boot-legacy-test.art
[mono-rt] 6f4ee000-6f520000 rw-p 00000000 fc:00 7377       /data/dalvik-cache/x86/system@framework@boot-bouncycastle.art
[mono-rt] 6f520000-6f54e000 rw-p 00000000 fc:00 7378       /data/dalvik-cache/x86/system@framework@boot-ext.art
[mono-rt] 6f54e000-6fa89000 rw-p 00000000 fc:00 7379       /data/dalvik-cache/x86/system@framework@boot-framework.art
[mono-rt] 6fa89000-6fad9000 rw-p 00000000 fc:00 7380       /data/dalvik-cache/x86/system@framework@boot-telephony-common.art
[mono-rt] 6fad9000-6fae1000 rw-p 00000000 fc:00 7381       /data/dalvik-cache/x86/system@framework@boot-voip-common.art
[mono-rt] 6fae1000-6fae8000 rw-p 00000000 fc:00 7382       /data/dalvik-cache/x86/system@framework@boot-ims-common.art
[mono-rt] 6fae8000-6fafd000 rw-p 00000000 fc:00 7383       /data/dalvik-cache/x86/system@framework@boot-apache-xml.art
[mono-rt] 6fafd000-6fb1b000 rw-p 00000000 fc:00 7384       /data/dalvik-cache/x86/system@framework@boot-org.apache.http.legacy.boot.art
[mono-rt] 6fb1b000-6fb1c000 rw-p 00000000 fc:00 7385       /data/dalvik-cache/x86/system@framework@boot-android.hidl.base-V1.0-java.art
[mono-rt] 6fb1c000-6fb1f000 rw-p 00000000 fc:00 7386       /data/dalvik-cache/x86/system@framework@boot-android.hidl.manager-V1.0-java.art
[mono-rt] 6fb1f000-6fcfe000 r--p 00000000 fd:01 1218       /system/framework/x86/boot.oat
[mono-rt] 6fcfe000-703af000 r-xp 001df000 fd:01 1218       /system/framework/x86/boot.oat
[mono-rt] 703af000-703b1000 rw-p 00000000 00:00 0          [anon:.bss]
[mono-rt] 703b1000-703b2000 r--p 00890000 fd:01 1218       /system/framework/x86/boot.oat
[mono-rt] 703b2000-703b3000 rw-p 00891000 fd:01 1218       /system/framework/x86/boot.oat
[mono-rt] 703b3000-7048c000 r--p 00000000 fd:01 1191       /system/framework/x86/boot-core-libart.oat
[mono-rt] 7048c000-70727000 r-xp 000d9000 fd:01 1191       /system/framework/x86/boot-core-libart.oat
[mono-rt] 70727000-70728000 rw-p 00000000 00:00 0          [anon:.bss]
[mono-rt] 70728000-70729000 r--p 00374000 fd:01 1191       /system/framework/x86/boot-core-libart.oat
[libc] Fatal signal 11 (SIGSEGV), code 1, fault addr 0x50 in tid 13237 (AsyncTask #3)
```

I have two concerns with the thought of that change. The first is GDPR (may be fine / not necessary to worry, but we should do some due diligence there). 

The second is that the freeform line that I can read from that file doesn't admit a lot of structure, and I'm not sure that parsing it should happen in the signal handler of a fatal crash. This is easy to handle just by baking the freeform line in at first, but I have to question how we expect people to use that data. Is there any chance that we could get similar information about endpoints using dladdr, as we currently use to match a native address up to a module?

lewurm:
I would just dump the whole file (`/proc/self/maps` is plain text) as string into the JSON. I don't know about GDPR 😕 wouldn't that be a problem already today?

alexanderkyte:
Dumping to stdout versus sending to crash reporting backends is the big difference there. I'll just put it behind the crash privacy preprocessor flag.

mono/utils/mono-state.c
```diff
@@ -927,6 +927,34 @@ mono_native_state_add_memory (MonoStateWriter *writer)
...
+	const int max_lines = 25;
+	char line [256];
+	int i = 0;
+
+	assert_has_space (writer);
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "process_map");
+	mono_state_writer_printf(writer, "[\n");
+
+	while (async_safe_fgets (line, sizeof (line), handle) && i++ < max_lines) {
```
lewurm: `max_lines` should go away, it should contain the whole file

mono/eglib/glib.h
```diff
@@ -1133,6 +1139,30 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
 
 gchar *g_mkdtemp (gchar *tmpl);
 
+static inline int
+g_async_safe_fgets (char *str, int num, int handle)
+{
+	int without_padding = num - 1;
+	int i=0;
+	char scratch [2];
+	while (i < without_padding && read (handle, scratch, 1)) {
```
lewurm: `s/read/g_read`?

mono/utils/mono-state.c
```diff
@@ -927,6 +927,34 @@ mono_native_state_add_memory (MonoStateWriter *writer)
 	mono_state_writer_printf(writer, "},\n");
 }
 
+static void
+mono_native_state_add_process_map (MonoStateWriter *writer)
+{
+#if defined(__linux__) && !defined(HOST_ANDROID)
```
lewurm: why can't we do this on `HOST_ANDROID`?

alexanderkyte: Does android have /proc? I'm not sure why it was requested to be removed from their dumps initially. I think it was just unexpected. 

Also should have pushed to a non-PR branch, but FYI I'm currently debugging this. Sorry if you reviewed unready code. Should be done soon, just ironing out some C problems. `read` is an annoying syscall sometimes.

mono/eglib/glib.h
alexanderkyte: Ah good catch

jaykrell: Really? Read is standard Unix. Just that it is _read on Windows, which isn't relevant.

mono/eglib/glib.h
```diff
@@ -1123,6 +1123,12 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
 #else
 #define g_write write
 #endif
+#ifdef G_OS_WIN32
+#define g_read g_assert_not_reached
+#else
+#define g_read read
+#endif
```
jaykrell: Is this needed?

mono/utils/mono-state.c
jaykrell: This looks bogus for other reasons. What happens if a line is longer or shorter than 256?
For longer, you need to keep reading and combine the buffers.
For shorter you have to keep part of the buffer.
If you just dump it as one big string you can just read/write in a loop with a fixed sized buffer.
You can still limit the line count easily.
Splitting lines requires complexity, like what you have but more.

jaykrell: The simplest fix is to seek backwards when you read past end of line. If that works.
If you do limit line count, then an array of pointers/lengths into one large buffer will help, if there is a reasonable size for the overall buffer.
Several other choices: You could have one large buffer and truncate at the last newline.
Split or not.
etc.

mono/eglib/glib.h
alexanderkyte: @jaykrell I think it's more of a greppability and consistency thing

mono/eglib/glib.h
alexanderkyte: I'll make it _read, I just haven't looked into Windows yet

mono/eglib/glib.h
jaykrell: I didn't think of that. I thought that is what the spaces are for.

mono/utils/mono-state.c
alexanderkyte: @jaykrell I was trying to maintain existing behavior, just baking the output we currently have into the JSON as an array of string lines instead. 

```
 /proc/[pid]/maps
              A file containing the currently mapped memory regions and
              their access permissions.  See mmap(2) for some further infor‐
              mation about memory mappings.

              Permission to access this file is governed by a ptrace access
              mode PTRACE_MODE_READ_FSCREDS check; see ptrace(2).

              The format of the file is:

    address           perms offset  dev   inode       pathname
    00400000-00452000 r-xp 00000000 08:02 173521      /usr/bin/dbus-daemon
    00651000-00652000 r--p 00051000 08:02 173521      /usr/bin/dbus-daemon
    00652000-00655000 rw-p 00052000 08:02 173521      /usr/bin/dbus-daemon
    00e03000-00e24000 rw-p 00000000 00:00 0           [heap]
    00e24000-011f7000 rw-p 00000000 00:00 0           [heap]
    ...
    35b1800000-35b1820000 r-xp 00000000 08:02 135522  /usr/lib64/ld-2.15.so
    35b1a1f000-35b1a20000 r--p 0001f000 08:02 135522  /usr/lib64/ld-2.15.so
    35b1a20000-35b1a21000 rw-p 00020000 08:02 135522  /usr/lib64/ld-2.15.so
    35b1a21000-35b1a22000 rw-p 00000000 00:00 0
    35b1c00000-35b1dac000 r-xp 00000000 08:02 135870  /usr/lib64/libc-2.15.so
    35b1dac000-35b1fac000 ---p 001ac000 08:02 135870  /usr/lib64/libc-2.15.so
    35b1fac000-35b1fb0000 r--p 001ac000 08:02 135870  /usr/lib64/libc-2.15.so
    35b1fb0000-35b1fb2000 rw-p 001b0000 08:02 135870  /usr/lib64/libc-2.15.so
    ...
    f2c6ff8c000-7f2c7078c000 rw-p 00000000 00:00 0    [stack:986]
    ...
    7fffb2c0d000-7fffb2c2e000 rw-p 00000000 00:00 0   [stack]
    7fffb2d48000-7fffb2d49000 r-xp 00000000 00:00 0   [vdso]

              The address field is the address space in the process that the
              mapping occupies.  The perms field is a set of permissions:

                  r = read
                  w = write
                  x = execute
                  s = shared
                  p = private (copy on write)

              The offset field is the offset into the file/whatever; dev is
              the device (major:minor); inode is the inode on that device.
              0 indicates that no inode is associated with the memory
              region, as would be the case with BSS (uninitialized data).

              The pathname field will usually be the file that is backing
              the mapping.  For ELF files, you can easily coordinate with
              the offset field by looking at the Offset field in the ELF
              program headers (readelf -l).

              There are additional helpful pseudo-paths:

                   [stack]
                          The initial process's (also known as the main
                          thread's) stack.

                   [stack:<tid>] (from Linux 3.4 to 4.4)
                          A thread's stack (where the <tid> is a thread ID).
                          It corresponds to the /proc/[pid]/task/[tid]/
                          path.  This field was removed in Linux 4.5, since
                          providing this information for a process with
                          large numbers of threads is expensive.

                   [vdso] The virtual dynamically linked shared object.  See
                          vdso(7).

                   [heap] The process's heap.

              If the pathname field is blank, this is an anonymous mapping
              as obtained via mmap(2).  There is no easy way to coordinate
              this back to a process's source, short of running it through
              gdb(1), strace(1), or similar.

              pathname is shown unescaped except for newline characters,
              which are replaced with an octal escape sequence.  As a
              result, it is not possible to determine whether the original
              pathname contained a newline character or the literal \e012
              character sequence.

              If the mapping is file-backed and the file has been deleted,
              the string " (deleted)" is appended to the pathname.  Note
              that this is ambiguous too.

              Under Linux 2.0, there is no field giving pathname.

```

If the line is longer than 256 it gets turned into a number of strings right now. It looks like the previous implementation was implicitly concatenating these multiple strings by failing to print the separators it didn't see directly in the input file stream. I'll replicate that behavior here.

alexanderkyte:
Wow, github really has issues displaying any interactive rebases if you don't follow with an aggressive ```git pull --rebase``` , will fix.

alexanderkyte:
Example execution from my machine:

https://gist.github.com/alexanderkyte/d833c2e6d83e8812b806a519e86875e5

alexanderkyte:
If I type `make check` from corlib, this fails. If I copy and paste the line run by the makefiles and only run the harness in question using `-test:Foo`, it succeeds

alexanderkyte:
It looks like it only breaks when run through Make, hmm.

alexanderkyte:
Ah wait, `-test:MonoTests` through make fails but `-test:MonoTests.System` through make succeeds. Seems likely to be an earlier test breaking shared VM state.

alexanderkyte:
A dump with the process map using #f651665 
https://gist.github.com/alexanderkyte/578b4e2f82a0e71a169c37d8bc18a305

mono/eglib/glib.h
```diff
@@ -1133,10 +1139,37 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
...
 
-
 /*
  * Low-level write-based printing functions
  */
+
+static inline int
+g_async_safe_fgets (char *str, int num, int handle, gboolean *newline)
+{
+	memset (str, '\0', sizeof(char) * num);
```
jaykrell: `'\0'` can just be 0.
In fact the parameter is `int`, not `char`.

`sizeof (char)` == 1, no need to multiply by it.

mono/eglib/glib.h
```diff
@@ -1133,10 +1139,37 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
...
+	// Make sure we always have null-termination
+	int without_padding = num - 1;
+	int i=0;
+	char scratch [2];
+	while (i < without_padding && g_read (handle, scratch, sizeof(char))) {
```
jaykrell: Reading one char at a time is very inefficient, but I see helps correctness elsewhere.

jaykrell: Suggest to do it with larger reads, but I know it takes more work.

mono/utils/mono-state.c
```diff
@@ -927,6 +927,75 @@ mono_native_state_add_memory (MonoStateWriter *writer)
...
+#define MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT 30
+
+static void
+mono_native_state_add_process_map (MonoStateWriter *writer)
+{
+#if defined(__linux__) && !defined(HOST_ANDROID)
+	static gint32 handle_in_use;
+	gint32 old = mono_atomic_cas_i32 (&handle_in_use, TRUE /* set */, FALSE /* compare */);
+	if (old != FALSE) {
+		g_async_safe_printf ("Warning: Couldn't open /proc/self/maps for concurrent reads, crash report will lack it.\n");
```
jaykrell: Why not concurrent? Because of crash reporting, not because of this code?

mono/utils/mono-state.c
```diff
@@ -927,6 +927,75 @@ mono_native_state_add_memory (MonoStateWriter *writer)
...
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "process_map");
+	mono_state_writer_printf(writer, "[\n");
+
+	int mapping = 0;
+	while (mapping < MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT) {
+		if (mapping > 0)
+			mono_state_writer_printf (writer, "\",\n");
+
+		mono_state_writer_printf (writer, "\t\"");
```
jaykrell: Can this occur at end of file w/o anything in the quotes?

jaykrell: Try to raise the line limit to fit it all?

mono/eglib/glib.h
alexanderkyte: It's also the case that making the size of the stack frame larger than a very small amount is dangerous here. The signal handler code runs in an alternate stack, which is much smaller. It's easy to overflow that stack and crash the crash reporter. It's simply not safe to allocate more scratch space, and reading one char at a time ensures that I never need to "put back" any characters.

mono/utils/mono-state.c
alexanderkyte: Because it leads to multiple threads trying to open, read from, and close from the same file at the same time. There are enough interesting issues there, that the most straightforward solution (one which doesn't push this interestingness to future debuggers) is just to refuse to do the file IO concurrently.

mono/eglib/glib.h
jaykrell: Reading one byte at a time is extremely slow. I know it is crash reporting. Hopefully only ever that. The caller’s buffer could be better used, along with a position into it.

jaykrell: Reading one byte at a time is extremely slow. I know it is crash reporting. Hopefully only ever that. The caller’s buffer could be better used, along with a position into it.

mono/utils/mono-state.c
jaykrell: Why can’t the file be opened and read more than once concurrently?

mono/utils/mono-state.c
alexanderkyte: The line limit is a limit on the size of the produced JSON. It was present in the initial dumper because I assumed it was possible to find apps that had an unreasonably large number of mappings. 

If this occurs at the end of the file, won't it just be an empty array element at the end? That seems like a fine edge case. 

[ "foo", "" ]

mono/eglib/glib.h
alexanderkyte: I can definitely read into the caller's array, but reading by more than one byte at a time would prevent the reading from stopping the minute it found any nonprintable characters, and the algorithm for skipping them would have to change. 

I could come up with some logic of overwriting the caller's buffer's terminators with '\0', but then printing that would have to involve "restarting" printing directly after the skipped character. It seems prone to risk of error. 

It should be noted that producing invalid JSON here means that the crash reporting can't be parsed, so this is mission critical. Also noted is that as slow as I'm sure it is, it doesn't seem to have any perceptible impact on the crash dumper. This code has some other major perf killers, such as global thread synchronization using signals.

mono/utils/mono-state.c
alexanderkyte: That's a good question. It seemed like it was breaking at one point, but that code had multiple bits of unsafe code.

Looking at the POSIX spec, it seems like any Linux kernel after 3.14 should handle that correctly. Let me enable it and see if it works.

mono/utils/mono-state.c
jaykrell: Fair enough.

mono/eglib/glib.h
alexanderkyte: I have changed this to use the caller's memory, but it still reads one char at a time. Let's wait until we see it being a perf problem before we try a less-reliable reading policy.

alexanderkyte:
@monojenkins build failed

mono/eglib/glib.h
```diff
@@ -1123,6 +1123,12 @@ gboolean   g_file_test (const gchar *filename, GFileTest test);
 #else
 #define g_write write
 #endif
+#ifdef G_OS_WIN32
+#define g_read _read
+#else
+#define g_read read
+#endif
```
jaykrell: On Unix `read` is a sycall, and therefore async-safe. On Windows it is not a syscall. So they are arguably quite differnent and maybe don't provide a portability layer over them, unless that portability layer is async-safe.

jaykrell: Async-safe only if you preserve `errno` around use, that is.

mono/utils/mono-state.c
```diff
@@ -927,6 +927,63 @@ mono_native_state_add_memory (MonoStateWriter *writer)
...
+		return;
+	}
+
+	assert_has_space (writer);
+	mono_state_writer_indent (writer);
+	mono_state_writer_object_key (writer, "process_map");
+	mono_state_writer_printf(writer, "[\n");
+
+	int mapping = 0;
+	while (mapping < MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT) {
```
lewurm: `MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT` shouldn't be there (or maybe something crazy like 9999). and it doesn't seem to handle EOF; what if `/proc/self/maps` has fewer than `MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT` entries?

alexanderkyte: So there's a couple points in there, let me handle them one by one.

``` 
MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT shouldn't be there (or maybe something crazy like 9999)
```

We had a line limit on the old printer, and it seemed even more important here to limit it, where the crash output writer doesn't have realloc right now. 

I think you can definitely bump the constant if you're not seeing enough info for MonoDevelop or other large apps, but it should be a constant (not a magic number in the source) and it should probably be a bit smaller than 9999. 

```
what if /proc/self/maps has fewer than MONO_CRASH_REPORTING_MAPPING_LINE_LIMIT entries?
```

That's the common case on my machine during testing, where it only prints a couple of lines for a "GC.GetGeneration" example with an assertion in the icall. It works fine for me. 

You should remember that it continues until `read` fails, in the underlying reader. So in this case, it just prints a few zero-length string entries at the end of the array. 

If you're concerned about the space usage there, feel free to make it end earlier. This seems like an edge case only hit by very small programs that fail very quickly, and it doesn't fail in a way that leads to invalid JSON or anything breaking.

lewurm: > We had a line limit on the old printer

In the beginning we didn't have the limit, but `/proc/self/maps` can be large, thus we limited for user ease. It effectively rendered it useless because you really want to have the full content of `maps` available in order to debug an issue... not the first few mappings. Thus I suggested to completely drop it.

Having a partial snapshot of `/proc/self/maps` in the JSON file is useless too. We really want a full mapping (or drop it from JSON too).

What causes `realloc` here? `mono_state_writer_printf (writer, "...")`?

alexanderkyte: I meant to say that it's possible to run out of space: https://github.com/mono/mono/blob/master/mono/utils/mono-state.c#L368

So we don't want to be fastidious everywhere else and then allow ourselves a few hundred lines of `/proc/self/maps`. There should probably be some kind of limit. Or maybe a bit of parsing.

PR #14710, opened 2019-05-30
Propagate error in mono_file_map_error on posix implementation

### LINKED ISSUE #14594: Make mono_file_map_error actually propagate the error on the posix implementation
The Windows implementation does this correctly, but for other platforms the code is simply: 
https://github.com/mono/mono/blob/master/mono/utils/mono-mmap.c#L396

The platforms should be consistent here, and the Windows behavior is the correct one.

It also might be possible to add the GC safe transition to mono_file_map but not mono_file_map_error after this is done.

CoffeeFlux:
Fixes #14594

This should make the behavior consistent with Windows. The commit adding the emscripten-related length check didn't specify the exact exception, so I just put something that's as descriptive as possible.

### DIFF
--- a/mono/utils/mono-mmap.c
+++ b/mono/utils/mono-mmap.c
@@ -364,6 +364,13 @@ mono_vfree (void *addr, size_t length, MonoMemAccountType type)
 void*
 mono_file_map (size_t length, int flags, int fd, guint64 offset, void **ret_handle)
 {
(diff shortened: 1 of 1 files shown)


CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #14716, opened 2019-05-30
Readonly interp callbacks option 2.

jaykrell:
Alternative to https://github.com/mono/mono/pull/14648.


mono/mini/interp/interp.c
```diff
@@ -2481,7 +2483,9 @@ interp_entry_general (gpointer this_arg, gpointer res, gpointer *args, gpointer
 	interp_entry (&data);
 }
 
-#ifdef MONO_ARCH_HAVE_INTERP_ENTRY_TRAMPOLINE
+#define interp_entry_from_trampoline NULL
```
jaykrell: This isn't ideal, in that, it "works" even if there is a function. It'll just replace it, if it is later.
Maybe better to provide a function with unreachable?

jaykrell:
Closing in deference to https://github.com/mono/mono/pull/14717, presumably.

PR #14717, opened 2019-05-30
Readonly interp callbacks.

jaykrell:
Alternative to https://github.com/mono/mono/pull/14648.

### DIFF
--- a/mono/mini/interp-stubs.c
+++ b/mono/mini/interp-stubs.c
@@ -146,6 +146,45 @@ stub_delegate_ctor (MonoObjectHandle this_obj, MonoObjectHandle target, gpointer
 	g_assert_not_reached ();
 }
 
+static void
+stub_entry_from_trampoline (gpointer ccontext, gpointer imethod)
+{
+	g_assert_not_reached ();
+}
+
+static void
+stub_to_native_trampoline (gpointer addr, gpointer ccontext)
+{
+	g_assert_not_reached ();
+}
+
+static void
+stub_frame_arg_to_data (MonoInterpFrameHandle frame, MonoMethodSignature *sig, int index, gpointer data)
+{
+	g_assert_not_reached ();
+}
+
+static void
+stub_data_to_frame_arg (MonoInterpFrameHandle frame, MonoMethodSignature *sig, int index, gpointer data)
+{
+	g_assert_not_reached ();
+}
+
+static gpointer
+stub_frame_arg_to_storage (MonoInterpFrameHandle frame, MonoMethodSignature *sig, int index)
+{
+	g_assert_not_reached ();
+	return NULL;
+}
+
+static void
+stub_frame_arg_set_storage (MonoInterpFrameHandle frame, MonoMethodSignature *sig, int index, gpointer storage)
+{
+	g_assert_not_reached ();
+}
+
+static const MonoEECallbacks mono_interp_stub_callbacks = MONO_INIT_EE_CALLBACKS (stub);
+
 void
 mono_interp_stub_init (void)
 {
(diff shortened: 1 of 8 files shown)


mono/mini/interp-stubs.c
```diff
@@ -149,32 +149,18 @@ stub_delegate_ctor (MonoObjectHandle this_obj, MonoObjectHandle target, gpointer
...
-	c.frame_get_jit_info = stub_frame_get_jit_info;
-	c.frame_get_ip = stub_frame_get_ip;
-	c.frame_get_arg = stub_frame_get_arg;
-	c.frame_get_local = stub_frame_get_local;
-	c.frame_get_this = stub_frame_get_this;
-	c.frame_get_parent = stub_frame_get_parent;
-	c.start_single_stepping = stub_start_single_stepping;
-	c.stop_single_stepping = stub_stop_single_stepping;
-	c.delegate_ctor = stub_delegate_ctor;
+#define stub_entry_from_trampoline NULL
```
jaykrell: This isn't ideal, in that, it "works" even if there is a function. It'll just replace it.
Maybe better to provide a function with unreachable?

mono/mini/interp/interp.c
```diff
@@ -2481,7 +2483,9 @@ interp_entry_general (gpointer this_arg, gpointer res, gpointer *args, gpointer
 	interp_entry (&data);
 }
 
-#ifdef MONO_ARCH_HAVE_INTERP_ENTRY_TRAMPOLINE
+#define interp_entry_from_trampoline NULL
```
jaykrell: This isn't ideal, in that, it "works" even if there is a function. It'll just replace it.
Maybe better to provide a function with unreachable?

jaykrell:
@monojenkins build failed

mono/mini/interp-stubs.c
lewurm: `#define stub_entry_from_trampoline g_assert_not_reached`?

lewurm:
It's slightly better, but still we have the problem of ordering. `const` buys us safety that we don't ever override a function callback. I think the ordering is more likely to happen and thus I want to be guarded against that mistake. Also, we _might_ want to be able to override function callbacks at some point in the future.

Thus I weak-reject this idea, but I won't block the PR if other people are in favor of it.

jaykrell:
const is for read only memory for efficiency and prevent overwrites by others.
If we think we might overwrite, then we can, later, either relax the const, or make a copy (we were copying before anyway).

The ability to deliberately do a partial overwrite might outweigh the principle of doing a full copy.
But deliberately overwritable structs of function pointers I think are relatively rare.
I realize it is best to design the interface correctly in v1, but I think the hypothetical quite unlikely.

This form largely but not completely guards against the ordering problem. There is still the duplication, but it only once (twice total vs. three total), and much nearer to the original, so arguably eye-ballable. 

Option 2 guards against the ordering more completely but with more repetition.
  I'm open to that if you prefer its better order guarding.
Option 1 had to get the ordering to match more times and further apart.

There is maybe other ways to reduce the ordering problem but maybe too ugly.
The two options I presented seemed a reasonable-to-me balance of simplicity and not too much with the C preprocessor.

For non-MONO_API maybe some day virtual functions. :)

mono/mini/interp-stubs.c
jaykrell: `g_assert_not_reached` is a macro, can't take its address, has to be wrapped in a function.
That is larger than just `NULL` but I think preferable.
Because if you actually write it twice, it will fail to compile.
Whereas the define will silently replace a function if it occurs after it (or error if it occurs before it).

It'd be nice to have a typed null constant variable like:

```
void * const a = 0; // but actual function pointer type
struct 
{
 void * b; // but actual type
} c = { a };
```

but that isn't legal.

So  I think provide the little extra stub functions, that `g_assert_not_reached`.
Or `g_assertf (0, "");` which includes the function name, but `g_assert_not_reached` does not.
I'll amend the PR as such shortly.

jaykrell:
@monojenkins build failed

jaykrell:
For the ordering then how about https://github.com/mono/mono/pull/14716?
This is a very widespread idiom in C though, w/ const, despite the ordering requirement, w/o switching to C++. Is C++ soon?

jaykrell:
I'd say the macro helps a lot though. The macro repetition is near what it is based on, and then the uses don't have to worry. I can try another form of the macro with a tighter binding.

jaykrell:
Maximizing readonlyiness, esp. of function pointers, really is a thing.
This pattern is widespread.
Nobody seems to mind the order dependentness.
The macro does help.
If you get the order wrong, you will often but not always get a type mismatch, which will be a warning, which could/should be turned into an error.

jaykrell:
How about https://github.com/mono/mono/pull/14648 then?

jaykrell:
Closed in deference to https://github.com/mono/mono/pull/14974, presumably.

jaykrell:
Different form in master now.

PR #14722, opened 2019-05-31
[wasm][bindings] Add support for marshaling DateTime and DateTimeOffset

kjpou1:
Add support for marshaling DateTime and DateTimeOffset between CLR and JavaScript.

- JavaScript Date object will create a new CLR DateTime object.
- CLR DateTime creates a new JavaScript Date object.
- CLR DateTimeOffset is represented as a string value.


sdks/wasm/binding_support.js
```diff
@@ -280,6 +287,10 @@ var BindingSupportLib = {
 				return this.get_task_and_bind (tcs, js_obj);
 			}
 
+			if (js_obj.constructor.name === "Date")
+				//return this.call_method(this.create_date_time, null, "dm", [ ((js_obj.getTime() * 10000) + 621355968000000000) - (js_obj.getTimezoneOffset() * 600000000) ]);
+				return this.call_method(this.create_date_time, null, "dm", [ (js_obj.getTime() * 10000) + 621355968000000000 ]);
```
kg: please make these magic numbers named variables or add a comment describing what each one is

kg: also numbers that big as JS literals may not work. That number looks too big for float64 to accurately represent

jaykrell: Remember there is an exponent.
Very large and very small numbers fit, just with limited precision.
Double has 53 mantissa bits. This seems to require around 41 or 45 for lossless.

```
\bin\amd64\cdb cmd
0:000> .formats 0n621355968000000000
Evaluate expression:
  Hex:     089f7ff5`f7b58 000
  Binary:  00001000 10011111 01111111 11110101 11110111 10110101 10000000 00000000
```

jaykrell: I don't know if JS literal parsing is that smart though.
Maybe you have to say 1.2e23 or such explicitly.
Or do split it up and math 621355968 * 1000000000.

jaykrell: Standardese is difficult for me.
http://www.ecma-international.org/ecma-262/9.0/index.html#sec-literals-numeric-literals
speaks of something about 20 digits, something..

kg: Thanks Jay. If you're sure it's fine then I'm OK with it being merged... but isn't the representation of DateTime and TimeSpan inside C# a long, not a double? It would be ideal if we could do this computation inside the wasm boundary so it stays as int64s.

kjpou1: Hey everyone

Thank you for the comments.  Katelyn I have already moved all of that into the managed side.  It should have been that way from the beginning but was looking at the offset calculation.  Will commit shortly.

kjpou1: @kg and @jaykrell 

Addressed comments in recent commit.

jaykrell: @kg I wasn't sure. The numbers fit, but to parse it as a literal would require slight cleverness.
The actual double exponent is a power of 2, not 10. So more clever.
Something like, have two doubles, representing the mantissa and a power of 10.
Power of 10 starts at 1.
If the mantissa is not zero and you see a zero, assume temporarily it is trailing zero, and multiple the power of 10 by 10, leave mantissa alone.
When you get a non-zero, multiple the mantissa by the power of 10 and reset the power of 10 to 1 (and add the new digit).
When you are all done, multiple mantissa times power of 10 exponent.
In this fashion, you could parse numbers that do not fit in int64.
I had never previously worked that through and don't know if anyone else has either. :)

kg: conversions involving a mix of double, int64, and decimal are always scary so hopefully we can avoid being scared here :)

jaykrell: Maybe I over think it.
Regular parsing, where you multiply by 10 and add character - '0' might just work.
Into a double from the start.
Each step will perturb the mantissa and exponent, 10 not being a power of 2, but they'll kinda keep trading off with each other and you'll never lose precision, and be able to losslessly parse numbers that don't fit in int64.

lewing:
@monojenkins build failed

PR #14723, opened 2019-05-31
Update Windows SDK to 10.0, PlatformToolset to v142

EgorBo:
To make it VS2019 friendly.
The changes are the result of "Retarget" command in vs2019 + search and replace for `=v140` to `=v142` for CI files.

https://www.mono-project.com/docs/compiling-mono/windows/ needs to be updated if we merge this.

### DIFF
--- a/llvm/build_llvm_msbuild.bat
+++ b/llvm/build_llvm_msbuild.bat
@@ -60,7 +60,7 @@ if not "%~1" == "" (
 )
 shift
 
-set "VS_ADDITIONAL_ARGUMENTS=/p:PlatformToolset=v140 /p:MONO_TARGET_GC=sgen"
+set "VS_ADDITIONAL_ARGUMENTS=/p:PlatformToolset=v142 /p:MONO_TARGET_GC=sgen"
 if not "%~1" == "" (
     set VS_ADDITIONAL_ARGUMENTS=%~1
 )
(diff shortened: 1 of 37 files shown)


EgorBo:
```
error MSB8020: The build tools for v142 (Platform Toolset = 'v142') cannot be found. To build using the v142 build tools, please install v142 build tools. 
```

kg:
> ```
> error MSB8020: The build tools for v142 (Platform Toolset = 'v142') cannot be found. To build using the v142 build tools, please install v142 build tools. 
> ```
That's a common problem with VS2017 that seems impossible to solve, so I just don't use that toolset. It's cursed (like VS updates and targeting in general)

EgorBo:
@kg can I update to v141 at least? (currently it's v140)

EgorBo:
However it's not necessary, vs2019 installer is able to install v140  (unlike the windows 8 sdk)

kg:
This would be fine if we tell people to use 2019 and not 2017. 141 might work, you'd have to test it. Which project is giving that error? I can try to repro.

lateralusX:
We discussed this a little in the past, moving up to VS2017 has been approved and decided to do, so maybe we should start to upgrade to VS2017 and pick toolchain and SDK that works with both 2017 and 2019 make it at least easy to retarget solution to VS2019 if needed?

steveisok:
I like @lateralusX's idea.  @EgorBo if more work / discussion is necessary, please close the PR and re-submit / re-open when it is ready.

jaykrell:
The VS project system seems to break everything with every release.
Imagine if make or C compiler broke things this often.
I suggest cmake or even nmake instead.

PR #14733, opened 2019-05-31
Add logic to object array typecheck to handle arrays of unmanaged pointers

### LINKED ISSUE #14555: [runtime] Pointers are not Objects
## Steps to Reproduce
```csharp
using System;

class Program
{
    static unsafe void Main()
    {
        Console.WriteLine(Test(new int*[0]));
    }

    static unsafe bool Test(Array a)
    {
        return a is object[];
    }
}
```
### Current Behavior

```
True
```

### Expected Behavior

```
False (CoreCLR)
```

[v] macOS
[ ] Linux
[ ] Windows

**Version Used**:

6.3.0.120

I guess the fix should be somewhere here https://github.com/mono/mono/blob/master/mono/mini/type-checking.c#L684-L689
Related xunit tests (netcore):
```
-nomethod System.Tests.ArrayTests.LastIndexOf_ArrayOfPointers_ThrowsNotSupportedException
-nomethod System.Tests.ArrayTests.Reverse_ArrayOfPointers_ThrowsNotSupportedException
-nomethod System.Tests.ArrayTests.IndexOf_ArrayOfPointers_ThrowsNotSupportedException
```

CoffeeFlux:
Fixes #14555 

The NULL check alone is not enough, because interfaces are also parented to NULL for some reason, and so I've moved `class_kind` out from the bitfield section to make it accessible from IR. The beginning of the struct is going to be padded to 32 bits anyway at the least, so this move should save space if anything.

Our type checking overall is not very robust, and in particular the supertype comparison doesn't work well for anything that isn't an object, so it's possible there are some other weird corner cases. As far as I can tell looking through the array of objects stuff though, we should have reasonable coverage with the existing tests + the stuff I added.


marek-safar:
There is still a failure on interpreter @lewurm can help you with that

lewurm:
>  And I bet we will need some fixes in mono_class_is_assignable_from_checked.

This should also fix the interpreter issue.

CoffeeFlux:
The behavior of `IsAssignableFrom` in CoreCLR does not appear to be correct, so I'm holding off on updating this until I get clarification, at which point I'll add additional tests and the interpreter fix.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #14760, opened 2019-06-03
Increase main thread stack size on Windows.

lateralusX:
On Windows default thread stack size has been 1MB for a long time (both 32/64-bit). When starting to use interpreter some of the tests include deep recursions that will hit stackoverflow on Windows(ackermann.exe as one example). This commit adjust the default reserved stack size to 8MB for debug builds and 1.5MB for release builds (same as coreclr), for main thread only. All other threads will still default to 1MB stack size (both debug/release).

### DIFF
--- a/mono/utils/mono-threads-windows.c
+++ b/mono/utils/mono-threads-windows.c
@@ -350,13 +350,19 @@ mono_threads_suspend_get_abort_signal (void)
 
 #if defined (HOST_WIN32)
 
+#define MONO_WIN32_DEFAULT_NATIVE_STACK_SIZE (1024 * 1024)
+
 gboolean
 mono_thread_platform_create_thread (MonoThreadStart thread_fn, gpointer thread_data, gsize* const stack_size, MonoNativeThreadId *tid)
 {
 	HANDLE result;
 	DWORD thread_id;
+	gsize set_stack_size = MONO_WIN32_DEFAULT_NATIVE_STACK_SIZE;
 
-	result = CreateThread (NULL, stack_size ? *stack_size : 0, (LPTHREAD_START_ROUTINE) thread_fn, thread_data, 0, &thread_id);
+	if (stack_size && *stack_size)
+		set_stack_size = *stack_size;
+
+	result = CreateThread (NULL, set_stack_size, (LPTHREAD_START_ROUTINE) thread_fn, thread_data, 0, &thread_id);
 	if (!result)
 		return FALSE;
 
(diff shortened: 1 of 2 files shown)


EgorBo:
Can be related: https://github.com/dotnet/coreclr/issues/21450

lateralusX:
We should implement similar property in order to override default stack size value at some point.

mono/utils/mono-threads-windows.c
```diff
@@ -350,13 +350,19 @@ mono_threads_suspend_get_abort_signal (void)
 
 #if defined (HOST_WIN32)
 
+#define MONO_WIN32_DEFAULT_NATIVE_STACK_SIZE 1024 * 1024
```
jaykrell: Parens?

msvc/mono.vcxproj
```diff
@@ -107,6 +107,7 @@
       <SubSystem>Console</SubSystem>
       <DataExecutionPrevention>
       </DataExecutionPrevention>
+      <StackReserveSize>0x800000</StackReserveSize>
```
jaykrell: The logic on Windows to determine stack size is very convoluted, involving this field, stuff in the registry, rounding, maybe hardcoded minimums/defaults, and the parameter to CreateThread.
I'd want to double check that this does not actually impact other threads.
For example if 0 is passed to CreateThread for stack size. Might that use this value?
As well, even if you want get closer to Linux before, it seems excessive.
Maybe 2MB would suffice?
As well this will have no impact on libmono.dll hosted by other .exes, arguably a good thing.

lateralusX: I have fixed all calls within runtime to default to 1MB if 0 is used as stack size, for other direct calls to CreateThread (using pinvoke/or other libraries) will default to process default set in PE header, if not explicitly specified in call to CreateThread. The PE header can be changed by using editbin if needed and it will only apply to mono.exe, so embedders loading the library or link static link will get the default specified for their binary, but runtime threads will still default to 1MB. In order to dynamically configure we would need a property or argument. I agree that 8MB might be a little excessive, picked that value since it was inline with other platforms default for main thread (not it is just reserved memory, not committed). Looking at our pthread implementation it picks a hard stack size for new created thread 1MB for 32-bit and 2MB for 64-bit threads, but default for main remains larger. I can run through some tests to see if we could set a lower limit, like 2MB, coreclr uses 0x180000 as their default process stack size.

lateralusX: We can split the default stack size config to be different between debug and release builds, debug needs to be at the higher default stack size, 8MB, for worst case test (ackermann.exe) to pass under interpreter, but release build could get down to as low as 0x180000 (same as coreclr default stack size), 1.5MB default stack size for main thread.

lateralusX:
@monojenkins rebuild failed

PR #14766, opened 2019-06-03
Change in process_complete_path exposed uninitialized variable error.

lateralusX:
https://github.com/mono/mono/commit/5d088cf0de7f3e50e3547dba361af4401e938dd4 exposed a problem with an uninitialized variable, only exposed on some build configurations, 32-bit Windows release builds.


lateralusX:
@monojenkins rebuild failed

PR #14777, opened 2019-06-03
Switch supported runtimes storage from an array to a list

### LINKED ISSUE #12141: Too many supportedRuntime tags in exe.config file cause problems on Windows
## Steps to Reproduce
1. Using a "Hello World type Console program set the relevant exe.config to the following on windows, using Production release Mono 5.16 ( 32 Bit and 64 Bit)
```
<?xml version="1.0"?>
<configuration>
  <startup>
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7" />
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" />
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.1" />
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0"/>
  </startup>
</runtime>
 

</configuration>
```

### Current Behavior
Application does not display any output, and does not launch.

### Expected Behavior
Application will Display "Hello World"

[ ] macOS
[ ] Linux
[X] Windows

**Version Used**:
(5.16.0.220)

CoffeeFlux:
Fixes #12141 

Switched to a list because in theory any number of tags can be present. Old behavior was to put them into an array of size 3 and zero the byte following them, which was causing the crash with more than two tags. There's no null check for the default runtime because that should never happen, and the list is reversed to preserve the order from the XML. Changing the signature of `get_runtimes_from_exe` should be fine because it's static and only called in the one place.

### DIFF
--- a/mono/metadata/domain.c
+++ b/mono/metadata/domain.c
@@ -122,7 +122,7 @@ static MonoFreeDomainFunc free_domain_hook;
 static MonoAotCacheConfig aot_cache_config;
 
 static GSList*
-get_runtimes_from_exe (const char *exe_file, MonoImage **exe_image, GSList *runtimes);
+get_runtimes_from_exe (const char *exe_file, MonoImage **exe_image);
 
 static const MonoRuntimeInfo*
 get_runtime_by_version (const char *version);
@@ -562,7 +562,7 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
 		 * that would mean it would be reloaded later. So instead, we save it to
 		 * exe_image, and close it during shutdown.
 		 */
-		runtimes = get_runtimes_from_exe (exe_filename, &exe_image, runtimes);
+		runtimes = get_runtimes_from_exe (exe_filename, &exe_image);
 #ifdef HOST_WIN32
 		if (!exe_image) {
 			exe_image = mono_assembly_open_from_bundle (exe_filename, NULL, FALSE);
@@ -1874,12 +1874,13 @@ get_runtime_by_version (const char *version)
 }
 
 static GSList*
-get_runtimes_from_exe (const char *file, MonoImage **out_image, GSList *runtimes)
+get_runtimes_from_exe (const char *file, MonoImage **out_image)
 {
 	AppConfigInfo* app_config;
 	char *version;
 	const MonoRuntimeInfo* runtime = NULL;
 	MonoImage *image = NULL;
+	GSList *runtimes = NULL;
 	
 	app_config = app_config_parse (file);
 	
--- a/mono/metadata/domain.c
+++ b/mono/metadata/domain.c
@@ -121,8 +121,8 @@ static MonoFreeDomainFunc free_domain_hook;
 /* AOT cache configuration */
 static MonoAotCacheConfig aot_cache_config;
 
-static void
-get_runtimes_from_exe (const char *exe_file, MonoImage **exe_image, const MonoRuntimeInfo** runtimes);
+static GSList*
+get_runtimes_from_exe (const char *exe_file, MonoImage **exe_image, GSList *runtimes);
 
 static const MonoRuntimeInfo*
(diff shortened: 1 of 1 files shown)


mono/metadata/domain.c
```diff
@@ -573,26 +572,27 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
...
+		runtimes = g_slist_prepend (runtimes, (gpointer)default_runtime);
 		g_print ("WARNING: The runtime version supported by this application is unavailable.\n");
 		g_print ("Using default runtime: %s\n", default_runtime->runtime_version); 
 	}
 
 	/* The selected runtime will be the first one for which there is a mscrolib.dll */
-	for (n = 0; runtimes [n] != NULL && ass == NULL; n++) {
-		current_runtime = runtimes [n];
+	GSList *tmp = runtimes;
+	while (tmp != NULL) {
```
Therzok: g_slist_foreach?

CoffeeFlux: I'm not sure a new function is really necessary considering how short this is, and it's not like C has anonymous functions. I can change it if you think there's a good reason to, though.

CoffeeFlux: Actually looking again, I think foreach would further complicate the logic thanks to the break, so I'm inclined to say it's definitely not the way to go.

Therzok: > I can change it if you think there's a good reason to, though.

Was just wondering. Yeah, I completely missed the break statement. Can't mark my own comment as resolved.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

mono/metadata/domain.c
```diff
@@ -563,7 +562,7 @@ mono_init_internal (const char *filename, const char *exe_filename, const char *
 		 * that would mean it would be reloaded later. So instead, we save it to
 		 * exe_image, and close it during shutdown.
 		 */
-		get_runtimes_from_exe (exe_filename, &exe_image, runtimes);
+		runtimes = get_runtimes_from_exe (exe_filename, &exe_image, runtimes);
```
lambdageek: `runtimes` on the RHS here is always NULL, right?

mono/metadata/domain.c
```diff
@@ -1888,26 +1888,26 @@ get_runtimes_from_exe (const char *file, MonoImage **out_image, const MonoRuntim
...
 			while (list != NULL) {
 				version = (char*) list->data;
 				runtime = get_runtime_by_version (version);
 				if (runtime != NULL)
-					runtimes [n++] = runtime;
+					runtimes = g_slist_prepend (runtimes, (gpointer)runtime);
 				list = g_slist_next (list);
 			}
-			runtimes [n] = NULL;
+			runtimes = g_slist_reverse (runtimes);
```
lambdageek: why?

also: `runtimes` starts out NULL when `get_runtimes_from_exe` is first called, right?  But if it wasn't - this would reverse the input portion of the list too, right?  That seems... undesirable.

CoffeeFlux: I wanted to preserve the order from the XML, and I believe prepending and then reversing is faster than appending a bunch of times. You're right about reversing the input portion, though I also agree that there's no reason for it to take an input runtimes at all, so I'm inclined to leave this once I fix that.

mono/metadata/domain.c
CoffeeFlux: Yeah, no reason to pass it at all. Will change.

mono/metadata/domain.c
lambdageek: Sounds good to me.

CoffeeFlux:
@monojenkins build failed

Issue #8397, opened 2018-04-21
Implement GC.GetAllocatedBytesForCurrentThread


marek-safar:
This is a new API as proposed [here](https://github.com/dotnet/corefx/issues/10157).

It should help us to better integrate with benchmark.net and possibly other tools.

ColeCLR implementation https://github.com/dotnet/coreclr/pull/6953

Lexcess:
Just to lend support to this issue. I've proposed that GetAllocatedBytesForCurrentThread be adopted in .Net Standard vNext here: https://github.com/dotnet/standard/issues/771 . Having a working implementation on Mono would really aid this effort and make runtime memory analysis significantly easier (both for tooling and in process monitoring).

naricc:
Implemented in PR https://github.com/mono/mono/pull/14779

PR #14779, opened 2019-06-03
Implement GC.GetAllocatedBytes

naricc:
Add GetAllocatedBytes() to the C# Thread class and implementation in the runtime. This method returns the total number of bytes allocated by the current thread during its lifetimes.

https://github.com/mono/mono/issues/8397


mcs/class/corlib/System.Threading/Thread.cs
```diff
@@ -321,6 +321,9 @@ static void SerializePrincipal (Thread th, IPrincipal value)
 			}
 		}
 
+        [MethodImplAttribute(MethodImplOptions.InternalCall)]
+        public static extern long GetAllocatedBytes();
```
EgorBo: There is already `GC.GetAllocatedBytesForCurrentThread()` for it. But with empty icall (https://github.com/mono/mono/issues/8397)

marek-safar: There is already existing API for this which is wired up to https://github.com/mono/mono/blob/e2d6361771b489706ab0264b30e8040f20060bd8/mono/metadata/icall.c#L9177, you need to hook your code there instead

mono/utils/mono-threads.h
```diff
@@ -278,6 +278,9 @@ typedef struct _MonoThreadInfo {
 	 * TODO support multiple values by multiple tools
 	 */
 	void *tools_data;
+
+	/* Total bytes allocated by this thread in its lifetime so far. */
+	gint64 total_bytes_allocated;
```
BrzVlad: I'd rather have this variable on the SgenThreadInfo and access it through the gc interface (see for example mono_gc_get_used_size used for GC.GetTotalMemory). We will need special computation anyway in the GC, since we can't just bump this thread local variable for every allocation due to perf reasons.

mono/sgen/sgen-alloc.c
```diff
@@ -246,6 +250,8 @@ sgen_alloc_obj_nolock (GCVTable vtable, size_t size)
 
 				zero_tlab_if_necessary (p, size);
 			} else {
+				THREAD_BYTES_ALLOCATED += TLAB_NEXT - TLAB_START - size;
```
BrzVlad: I see you are bumping the size once the previous tlab is full. This does not handle in real time the allocation inside the current tlab. We could do this by computing on the spot the difference between tlab_next and tlab_start (when asked for the alloc size).

BrzVlad:
This does not handle all allocation paths. You should look for all paths that alloc large objects (sgen_los_alloc_large_inner) and all paths that allocate directly inside the major heap (alloc_degraded, alloc_small_pinned_obj). It also doesn't handle the lock free fast path (sgen_try_alloc_obj_nolock)

mcs/class/corlib/System.Threading/Thread.cs
```diff
@@ -321,7 +321,7 @@ static void SerializePrincipal (Thread th, IPrincipal value)
 			}
 		}
 
-		[MethodImplAttribute(MethodImplOptions.InternalCall)]
+		[MethodImpl*ttribute(MethodImplOptions.InternalCall)]
```
lambdageek: Stray typo

mono/utils/mono-threads.h
naricc: Is there some where that defines the gc interface? I see that all the gc's implement a mono_gc_get_used_size(void); but is there somewhere that specifies this is part of the interface?

BrzVlad: Not really. You can define the prototype in gc-internals.h and have all gc's (boehm, null) implement that method. That's it

mono/sgen/sgen-alloc.c
naricc: > This does not handle all allocation paths. You should look for all paths that alloc large objects (sgen_los_alloc_large_inner) and all paths that allocate directly inside the major heap (alloc_degraded, alloc_small_pinned_obj). It also doesn't handle the lock free fast path (sgen_try_alloc_obj_nolock)

So, many of these allocator call other allocator functions. I am actually incrementing on some sgen_los_alloc_large_inner calls, just not inside the method. To avoid double counting, I adopt this convention: Only increment on the paths where we actually allocate, and not if allocation is deferred to another method, to avoid confusion. Except for the hot path, where we increment on retiring the TLAB.

naricc: Hmm, no, what I have suggested doesn't actually work well. Because, for example, when we allocate on promotion I would double count those if I push increments down into the mark sweep allocator. I'll just have to do it carefully and make sure I get everything with no double counts.

mono/sgen/sgen-alloc.c
```diff
@@ -63,10 +63,19 @@ static guint64 stat_bytes_alloced_los = 0;
 #define TLAB_TEMP_END	(__thread_info__->tlab_temp_end)
 #define TLAB_REAL_END	(__thread_info__->tlab_real_end)
 
+static void
```
marek-safar: Shouldn't this be `static inline` or something like that?

jaykrell: Let the compiler decide to inline or not.

mono/sgen/sgen-alloc.c
```diff
@@ -63,10 +63,19 @@ static guint64 stat_bytes_alloced_los = 0;
 #define TLAB_TEMP_END	(__thread_info__->tlab_temp_end)
 #define TLAB_REAL_END	(__thread_info__->tlab_real_end)
 
+static void 
+increment_thread_allocation_counter(size_t byte_size)
+{
+	SgenThreadInfo* info;
+	info = mono_thread_info_current ();
+	info->total_bytes_allocated += byte_size;
+}
```
jaykrell: `mono_thread_info_current ()->total_bytes_allocated += byte_size`

mono/sgen/sgen-alloc.c
```diff
@@ -63,10 +63,19 @@ static guint64 stat_bytes_alloced_los = 0;
...
+
 static GCObject*
 alloc_degraded (GCVTable vtable, size_t size, gboolean for_mature)
 {
 	GCObject *p;
+	increment_thread_allocation_counter(size);
```
jaykrell: nit here and elsewhere: space before paren

mono/sgen/sgen-alloc.c
```diff
@@ -246,6 +258,12 @@ sgen_alloc_obj_nolock (GCVTable vtable, size_t size)
 
 				zero_tlab_if_necessary (p, size);
 			} else {
+				/* Subtract size because the size of the object we are trying to allocate
+				 * will either be counted in the next TLAB,
+				 *  or in alloc_degraded.
+				 */
+				increment_thread_allocation_counter(TLAB_NEXT - TLAB_START - size);
```
jaykrell: Is the temporary subtraction visible from outside and significant?

BrzVlad: I don't understand this line. Why are we decrementing size here ? Also we should bump the size for the previous tlab only after we are done allocating from it.

BrzVlad:
We should increment the size only after the allocation was successful. In theory alloc functions can return NULL if allocation failed. Also, please pay attention to the coding guidelines, mainly space before ( on function calls.

mono/sgen/sgen-alloc.c
naricc: My understanding of the code: If we are in this case, we have already moved  TLAB_NEXT beyond the end of the TLAB by size. We are therefore not going to allocate in this TLAB, and shouldn't count the allocation here. Thus the subtraction. 

This current allocation will be satisfied by either allocating in the next TLAB or using alloc_degraded. In the first case, it gets counted when *that* TLAB is retired. In the second, it is counted directly in alloc_degraded. Either way, it shouldn't be counted for the TLAB we are retiring now, or it else it will be double counted.

This is a little confusing, which is why I wrote the comment. I'm not sure of a more clear way to do it.

@jaykrell The subtraction is not visible from outside.

BrzVlad: 1. We do TLAB_NEXT -= size above, so we are undo'ing this _invalid_ bump
2. Assume you do a degraded_alloc for this current size (and you already bumped the size for the entire tlab with this line) and then a smaller size is allocated next that fits in this still valid tlab. Wouldn't we then double count the tlab data ?
3. Now that I think about it, when doing a collection we will clear tlabs with allocated data that will not be accounted for (see sgen_clears_tlabs). We will need to bump data there.
4. This is getting a little bit complex. We should add a small test that allocates objects in a loop and checks that this counter returns valid data. Some considerations would be that on 32 bit the object header is 8 bytes, while on 64 bit it is 16 bytes, and all objects are 8 byte aligned (the allocated object size is multiple of 8, see SGEN_ALLOC_ALIGN)

Therzok:
Don't mind me, I'm just really hyped for being able to reliably get allocation data from benchmarks.

mcs/class/corlib/Test/System.GC/GCTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            	expectedSize += objectAllocators[r.Next(0, objectAllocators.Length) ]().ExpectedSize();
+            }
+
+            long bytesAfterAlloc = GC.GetAllocatedBytesForCurrentThread();
+
+			Assert.AreEqual(expectedSize, bytesAfterAlloc - bytesBeforeAlloc);
+        }
+
+	}
+}
```
BrzVlad: Messy table / spaces in this file

mcs/class/corlib/corlib_test.dll.sources
```diff
@@ -86,6 +86,7 @@ System/EnumTest.cs
 System/EnvironmentTest.cs
 System/ExceptionTest.cs
 System/FloatingPointFormatterTest.cs
+System.GC/GCtest.cs
```
BrzVlad: This is GCTest. You should try out changes locally : make check

mono/sgen/sgen-alloc.c
```diff
@@ -167,6 +174,7 @@ sgen_alloc_obj_nolock (GCVTable vtable, size_t size)
 	 */
 
 	if (real_size > SGEN_MAX_SMALL_OBJ_SIZE) {
+		increment_thread_allocation_counter (size);
```
BrzVlad: only if p non null

mono/sgen/sgen-alloc.c
```diff
@@ -460,6 +483,7 @@ GCObject*
 sgen_alloc_obj_mature (GCVTable vtable, size_t size)
 {
 	GCObject *res;
+	increment_thread_allocation_counter (size);
```
BrzVlad: after allocation, if res is non null

BrzVlad:
This was the last of my review. If tests pass after the compilation error is fixed, I'd say it looks good.

mcs/class/corlib/corlib_test.dll.sources
BrzVlad: I just noticed there is also a System/GCTest. You could simply move the entire test case in that file

mono/metadata/sgen-mono.c
```diff
@@ -2486,6 +2486,16 @@ mono_gc_get_los_limit (void)
 	return SGEN_MAX_SMALL_OBJ_SIZE;
 }
 
+int64_t
```
lewurm: `gint64`

lambdageek: > `gint64`

@lewurm Why? We're C99, aren't we?

lewurm: @lambdageek consistency with the other function signatures in `boehm-gc.c` and `null-gc.c`

naricc:
@monojenkins build failed

naricc:
@monojenkins build failed

lewing:
@monojenkins build failed

adamsitnik:
Hello everyone!

I would like to add support for it to BenchmarkDotNet. So far we had an if check for Mono and the MemoryDiagnoser was returning 0. 

How can I obtain an installer of the latest Mono preview with this method implemented?

marek-safar:
@adamsitnik pick any commit you like and at https://github.com/mono/mono/commits/master you look for the green check, then click on it and it will show links with the packages for the commit

PR #14782, opened 2019-06-03
[jit] Treat ternary operations as non-constant during constant evaluation and propagation

### LINKED ISSUE #13873: Optimization ssa and abcrem do not work on x86 architectures since mono 5.18
## Steps to Reproduce

1. Get at least Mono 5.18.0 (with e.g. Debian Stretch or Ubuntu 18.10 for x86 platforms)
2. Compile this small program: 
```csharp
using System.IO;
using System.Security.Cryptography;
using System.Security.Cryptography.X509Certificates;

namespace test
{
        class test
        {
                {
                        X509Certificate2 result=new X509Certificate2(new byte[]{1,2,3});
                }
        }
}
```
3. Start it via `mono -O=ssa test.exe` or with `mono -O=abcrem test.exe` on a x86 system and it will crash. 
It works fine on x64 platforms or with Mono version 5.16 or lower.

### Current Behavior

What is the current behavior?
Segfault:
```
* Assertion at ssa.c:847, condition `num_sregs <= 2' not met
```

### Expected Behavior
No output, no crash, exit code 0.

[ ] macOS
[X] Linux
[ ] Windows

**Version Used**:
5.18.0.240

## Stacktrace

```
Stacktrace:

  at <unknown> <0xffffffff>
  at Mono.SystemCertificateProvider.get_X509Pal () [0x00000] in <a9a08e39ba304bd0a84c49bd158dfc02>:0
  at Mono.SystemCertificateProvider.Import (byte[],Microsoft.Win32.SafeHandles.SafePasswordHandle,System.Security.Cryptography.X509Certificates.X509KeyStorageFlags,Mono.CertificateImportFlags) [0x0001b] in <a9a08e39ba304bd0a84c49bd158dfc02>:0
  at System.Security.Cryptography.X509Certificates.X509Certificate2.Import (byte[],string,System.Security.Cryptography.X509Certificates.X509KeyStorageFlags) [0x00017] in <a9a08e39ba304bd0a84c49bd158dfc02>:0
  at System.Security.Cryptography.X509Certificates.X509Certificate2..ctor (byte[]) [0x00011] in <a9a08e39ba304bd0a84c49bd158dfc02>:0
  at test.test.Main (string[]) [0x00000] in <309d78ccfdbb47ae94234858050d890a>:0
  at (wrapper runtime-invoke) <Module>.runtime_invoke_void_object (object,intptr,intptr,intptr) [0x0004e] in <309d78ccfdbb47ae94234858050d890a>:0
/proc/self/maps:
565fa000-56615000 r--p 00000000 fb:00 45270507                           /usr/bin/mono-sgen
56615000-56913000 r-xp 0001b000 fb:00 45270507                           /usr/bin/mono-sgen
56913000-56adf000 r--p 00319000 fb:00 45270507                           /usr/bin/mono-sgen
56ae0000-56ae4000 r--p 004e5000 fb:00 45270507                           /usr/bin/mono-sgen
56ae4000-56ae7000 rw-p 004e9000 fb:00 45270507                           /usr/bin/mono-sgen
56ae7000-56b6e000 rw-p 00000000 00:00 0 
58228000-582f5000 rw-p 00000000 00:00 0                                  [heap]
f4e9d000-f5137000 r--p 00000000 fb:00 48740685                           /usr/lib/mono/gac/System/4.0.0.0__b77a5c561934e089/System.dll
f5137000-f5138000 r--p 00000000 fb:00 48740849                           /usr/lib/mono/aot-cache/x86/mscorlib.dll.so
f5138000-f561b000 r-xp 00001000 fb:00 48740849                           /usr/lib/mono/aot-cache/x86/mscorlib.dll.so
f561b000-f5769000 r--p 004e4000 fb:00 48740849                           /usr/lib/mono/aot-cache/x86/mscorlib.dll.so
f5769000-f576a000 r--p 00631000 fb:00 48740849                           /usr/lib/mono/aot-cache/x86/mscorlib.dll.so
f576a000-f576b000 rw-p 00632000 fb:00 48740849                           /usr/lib/mono/aot-cache/x86/mscorlib.dll.so
f576b000-f577f000 rw-p 00000000 00:00 0 
f577f000-f5bff000 r--p 00000000 fb:00 48740627                           /usr/lib/mono/4.5/mscorlib.dll
f5bff000-f63ff000 rw-p 00000000 00:00 0 
f63ff000-f6400000 ---p 00000000 00:00 0 
f6400000-f7000000 rw-p 00000000 00:00 0 
f70ae000-f7100000 r--p 00000000 fb:00 48740695                           /usr/lib/mono/gac/Mono.Security/4.0.0.0__0738eb9f132ed756/Mono.Security.dll
f7100000-f7121000 rw-p 00000000 00:00 0 
f7121000-f7200000 ---p 00000000 00:00 0 
f721b000-f721c000 ---p 00000000 00:00 0 
f721c000-f721d000 rw-p 00000000 00:00 0 
f721d000-f7225000 ---p 00000000 00:00 0 
f7225000-f73a0000 rw-p 00000000 00:00 0 
Memory around native instruction pointer (0xf771ae39):
0xf771ae29  8d b4 26 00 00 00 00 51 52 55 89 e5 0f 34 cd 80  ..&....QRU...4..
0xf771ae39  5d 5a 59 c3 fc f9 ff ff 64 00 00 00 71 00 03 03  ]ZY.....d...q...
0xf771ae49  03 ef f9 ff ff 5a 00 00 00 72 00 03 03 03 ae fd  .....Z...r......
0xf771ae59  ff ff 50 00 00 00 71 00 03 03 03 a1 fd ff ff 46  ..P...q........F

Native stacktrace:

	mono(+0x10dbf1) [0x56707bf1]
	mono(+0x10df29) [0x56707f29]
	mono(+0xb4f79) [0x566aef79]
	mono(+0x10d164) [0x56707164]
	linux-gate.so.1(__kernel_rt_sigreturn+0) [0xf771ae20]
	linux-gate.so.1(__kernel_vsyscall+0x9) [0xf771ae39]
	/lib/i386-linux-gnu/libc.so.6(gsignal+0xc2) [0xf74212d2]
	/lib/i386-linux-gnu/libc.so.6(abort+0xf0) [0xf740b2b6]
	mono(+0x1ca9a) [0x56616a9a]
	mono(+0x30f0d0) [0x569090d0]
	mono(monoeg_assertion_message+0x27) [0x56909427]
	mono(+0x11dd7f) [0x56717d7f]
	mono(+0x11fc8e) [0x56719c8e]
	mono(+0x114cb2) [0x5670ecb2]
	mono(+0x115754) [0x5670f754]
	mono(+0x226c5) [0x5661c6c5]
	mono(+0xb8937) [0x566b2937]
	mono(+0xb9446) [0x566b3446]
	[0xf73e0188]
	[0x582c8b68]
	[0xf73a18a4]
	[0xf73a1560]
	[0xf73a1307]
	[0xf73a1068]
	[0xf73a1123]
	mono(+0x238a1) [0x5661d8a1]
Pkilling 0xf731bb40 from 0xf73f0700
Entering thread summarizer pause from 0xf73f0700
Finished thread summarizer pause from 0xf73f0700.
le.
```

imhameed:
Fixes #13873

### DIFF
--- a/mono/mini/basic.cs
+++ b/mono/mini/basic.cs
@@ -1428,12 +1428,4 @@ class Tests
 
 		return k == -32768 ? 0 : 1;
 	}
-
-	static int variable_with_constant_address;
-
-	public static int test_cfold_with_non_constant_ternary_op () {
-		variable_with_constant_address = 0;
-		var old = System.Threading.Interlocked.CompareExchange(ref variable_with_constant_address, 1, 0);
-		return old == 0 && variable_with_constant_address == 1 ? 0 : 1;
-	}
 }
--- a/mono/mini/basic.cs
+++ b/mono/mini/basic.cs
@@ -1428,4 +1428,12 @@ class Tests
 
 		return k == -32768 ? 0 : 1;
 	}
+
+	static int variable_with_constant_address;
+
+	public static int test_cfold_with_non_constant_ternary_op () {
+		variable_with_constant_address = 0;
+		var old = System.Threading.Interlocked.CompareExchange(ref variable_with_constant_address, 1, 0);
+		return old == 0 && variable_with_constant_address == 1 ? 0 : 1;
+	}
 }
--- a/mono/mini/ssa.c
+++ b/mono/mini/ssa.c
@@ -826,6 +826,10 @@ evaluate_ins (MonoCompile *cfg, MonoInst *ins, MonoInst **res, MonoInst **carray
 		return 2;
 
 	num_sregs = mono_inst_get_src_registers (ins, sregs);
+
+	if (num_sregs > 2)
+		return 2;
+
 	for (i = 0; i < MONO_MAX_SRC_REGS; ++i)
 		args [i] = NULL;
 	for (i = 0; i < num_sregs; ++i) {
--- a/mono/mini/objects.cs
+++ b/mono/mini/objects.cs
@@ -2001,6 +2001,13 @@ ncells ) {
 		return gh_11378_inner_3 (new Vec3(0, 2, -20));
 	}
 
+	static int variable_with_constant_address;
+
+	public static int test_0_cfold_with_non_constant_ternary_op () {
+		variable_with_constant_address = 0;
+		var old = System.Threading.Interlocked.CompareExchange(ref variable_with_constant_address, 1, 0);
+		return old == 0 && variable_with_constant_address == 1 ? 0 : 1;
+	}
 }
 
 #if __MOBILE__


mono/mini/basic.cs
```diff
@@ -1428,4 +1428,12 @@ public static int test_1_rem_un_cfold ()
...
 		return k == -32768 ? 0 : 1;
 	}
+
+	static int variable_with_constant_address;
+
+	public static int test_cfold_with_non_constant_ternary_op () {
+		variable_with_constant_address = 0;
+		var old = System.Threading.Interlocked.CompareExchange(ref variable_with_constant_address, 1, 0);
+		return old == 0 && variable_with_constant_address == 1 ? 0 : 1;
+	}
```
vargaz: This should go to objects.cs since it uses statics/calls. Also, it should be named test_<EXPECTED RETURN VALUE>_...

lewurm:
@monojenkins build failed

mono/mini/ssa.c
```diff
@@ -826,6 +826,10 @@ evaluate_ins (MonoCompile *cfg, MonoInst *ins, MonoInst **res, MonoInst **carray
 		return 2;
 
 	num_sregs = mono_inst_get_src_registers (ins, sregs);
+
+	if (num_sregs > 2)
+		return 2;
```
lewurm: I think this is a bit too hack-ish. There is no real reason to _not_ support constant folding instructions with `num_sregs = 3`. The reason why we assert on it currently is that we do not have any instruction with that property where we could constant fold.

Why do we hit the assert then? First of all let us figure out why this only hits on 32bit. If you look at the generated IR for your testcase, you get on amd64:
```
 il_seq_point il: 0x7
 i8const R21 <- [4361201612]
 iconst R22 <- [1]
 iconst R23 <- [0]
 atomic_cas_i4 R25 <- [R21 + 0x0] R22 R23
```
vs. x86:
```
 iconst R13 <- [11144108]
 iconst R14 <- [1]
 iconst R15 <- [0]
 atomic_cas_i4 R17 <- [R13 + 0x0] R14 R15
```

The assert triggers for `atomic_cas_i4` only on 32bit because `i8const` isn't properly detected as constant. In fact, we need to extend this https://github.com/mono/mono/blob/a76fc055cf1dc3173b6df1690d14723a13fbaf63/mono/mini/ssa.c#L819-L823
to also support `OP_I8CONST`
```c
      /* Short-circuit this */
      if (ins->opcode == OP_ICONST || ins->opcode == OP_I8CONST) {
          *res = ins;
          return 1;
      }
```
_then_ you get the same issue on 64bit. <sup>1</sup>

So the assert hits when processing `atomic_cas_i4 R25 <- [R21 + 0x0] R22 R23`. I think the loop below should be able to detect memory access on a constant address and set `const_args = FALSE` in such cases. If we ever add instructions with `num_sregs = 3` and a constant folding opportunity would appear, we would still hit the `g_assert (num_sregs <= 2);`, letting us know to do something about it. With your change we would silently skip over it.


<sup>1</sup> Unfortunately this seems to break other things _sigh_

alexanderkyte: Yikes. If any of this remains unfixed, this should be documented with comments in the source. Sounds like something that would produce difficult-to-discern causes if it crashed in the field. 

@lewurm what other things does it break? Maybe a separate PR could consider those.

imhameed: It is true that there's no reason not to support constant folding ternary instructions.

Some ternary operations could benefit from constant folding--the ones that come to mind are fused multiply-add and SIMD blends/shuffles. But I don't think a crashing assertion is appropriate for this, even if we add constant-foldable ternary instructions, because I'm not sure what "abstract" invariant is being violated here. For example, this code doesn't crash when it encounters 64-bit arithmetic operations (LMUL/LADD/etc.); it just silently ignores them.

One less hacky option would be to extend `mono_constant_fold_ins` to take a pointer to an array of `MonoInst *[MONO_MAX_SRC_REGS]`; `mono_constant_fold_ins` could then, for the time being, treat ternary operations the same way it treats 64-bit arithmetic operations (i.e., it could ignore them). The failing assertion could then be completely removed.

vargaz: We don't have ternary ops which could be cfold-ed right now, so this looks fine to me.

lewurm: > But I don't think a crashing assertion is appropriate for this, even if we add constant-foldable ternary instructions, because I'm not sure what "abstract" invariant is being violated here.

True, an assertion might be too... assertive 🙂  There's `g_warning ()`, but depending on the situation that might be too annoying in the long-term.

For the sake of fixing the bug we can go with the suggested change, but please add a comment to it.

We should follow up on the `I8CONST` thing though, looks like we miss some constant folding opportunities.

marek-safar:
@monojenkins backport 2019-06

marek-safar:
@monojenkins backport 2019-02

lewurm:
This wasn't quite ready to merge yet imho (see my comment). I guess I should have used the "request change" feature 😕

PR #14798, opened 2019-06-04
Conditionalize pthread_mutexattr_setprotocol for Android.

jaykrell:
Could autoconf instead.


lambdageek:
Autoconf version over in #14799

akoeplinger:
The autoconf version looks a bit more consistent with how we do other checks so let's go with that one, thanks :)

PR #14811, opened 2019-06-05
Fix interpreter failure in install_eh_callback.exe under Windows x64 MSVC.

lateralusX:
INTERP_PUSH_LMF_WITH_CTX_BODY is not fully supported under Windows MSVC since that compiler doesn't support taking address from a label, needed by current implementation to correctly resume after unwinding exceptions happening in native frames as tested by install_eh_callback.exe.

This fix add support to restore to a known location and use a flag indicating if we have been resumed to the location or just passing by for the first time. If we resume it will transfer control to specified label. This will mimic the behavior of taking the address of a label and use that address as the resume point.

I also considered switching to setjmp/longjmp but that would have had a bigger impact on all platforms and needed changes into the resume context machinery as well.

While doing the implementation I also found an interesting issue with MONO_CONTEXT_GET_CURRENT under Windows x64. On most other platforms it is implemented as inline assembler, but on MSVC x64 inline assembler is not supported so it is implemented as a function. Problem with that is how IP
was resolved, since it won't be resolved to the last instruction in the expanded macro, but to the IP within the mono_context_get_current. This has not had any visible effects in the past since it will just do the ret and get back to the call site, but when this was used in INTERP_PUSH_LMF_WITH_CTX_BODY
it caused some issues. This was fixed to make sure rsp and rip reflect location in caller frame and not within mono_context_get_current, similar to how it works on all platforms with inline assembler support.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -384,24 +384,33 @@ mono_interp_get_imethod (MonoDomain *domain, MonoMethod *method, MonoError *erro
 	return imethod;
 }
 
-#if defined (MONO_CROSS_COMPILE) || defined (HOST_WASM) || defined (_MSC_VER)
-#define INTERP_PUSH_LMF_WITH_CTX_BODY(ext, exit_addr) \
+#if defined (MONO_CROSS_COMPILE) || defined (HOST_WASM)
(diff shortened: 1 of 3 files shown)


lateralusX:
@monojenkins rebuild failed

Issue #14793, opened 2019-06-04
Mono-Native not working on MacOS Catalina


xDaKillerx:
## Steps to Reproduce

1. Just run an app . say Sonarr or Radarr
2. 
3. 

### Current Behavior

Opening the app doesnt work and on opening using terminal , results in this error 
```
exception inside UnhandledException handler: (null) assembly:/Library/Frameworks/Mono.framework/Versions/6.5.0/lib/mono/4.5/mscorlib.dll type:TypeInitializationException member:(null)

[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: The type initializer for 'System.Console' threw an exception. ---> System.TypeInitializationException: The type initializer for 'System.ConsoleDriver' threw an exception. ---> System.DllNotFoundException: /Library/Frameworks/Mono.framework/Versions/6.5.0/lib/../lib/libmono-native-compat.dylib assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Interop+Sys.Stat(byte&,Interop/Sys/FileStatus&)
  at Interop+Sys.Stat (System.ReadOnlySpan`1[T] path, Interop+Sys+FileStatus& output) [0x0001b] in /Users/builder/jenkins/workspace/build-package-osx-mono/master/external/bockbuild/builds/mono-x64/external/corefx/src/Common/src/Interop/Unix/System.Native/Interop.Stat.Span.cs:24 
  at System.IO.FileSystem.FileExists (System.ReadOnlySpan`1[T] fullPath, System.Int32 fileType, Interop+ErrorInfo& errorInfo) [0x00007] in /Users/builder/jenkins/workspace/build-package-osx-mono/master/external/bockbuild/builds/mono-x64/external/corefx/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs:445 
  at System.IO.FileSystem.DirectoryExists (System.ReadOnlySpan`1[T] fullPath, Interop+ErrorInfo& errorInfo) [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/master/external/bockbuild/builds/mono-x64/external/corefx/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs:419 
[...]

Issue #14792, opened 2019-06-04
[sdks][ios] Perform the appropriate cast of objc_msgSend and friends


mandel-macaque:
With the new Xcode 11 the objc_msgSend we have OBJC_OLD_DISPATCH_PROTOTYPES == 0 which enforces the rule that the dispatch functions must be cast to an appropriate function pointer type.

This means that we need to perform the cast in several methods, specially the ones found in mono/utils/mono-threads-mach-helper.c 

A workaround is to #define OBJC_OLD_DISPATCH_PROTOTYPES 1 but a much better approach is to find the calls and cast.

akoeplinger:
Undo the -DOBJC_OLD_DISPATCH_PROTOTYPES=1 workaround that was landed as part of #14820 once this is fixed

PR #14820, opened 2019-06-05
Fix build on macOS 10.15 Catalina + xcode 11

EgorBo:
1) CC_MD5_Init, CC_MD5_Update and CC_MD5_Final are deprecated.
```c
mono-md5.c:45:2: error: 'CC_MD5_Init' is deprecated: first deprecated in macOS 10.15 - This function is
      cryptographically broken and should not be used in security contexts. Clients should migrate to SHA256
      (or stronger). [-Werror,-Wdeprecated-declarations]
        CC_MD5_Init (ctx);
        ^
```

2) 
```
mono-threads-mach-helper.c:95:17: error: too many arguments to function call, expected 0, have 2
                objc_msgSend (value, release);
                ~~~~~~~~~~~~  ^~~~~~~~~~~~~~
```
XCode has a special switch `Enable Strict Checking of objc_msgSend Calls` which is now `true` by default? anyway it's fixed by `-DOBJC_OLD_DISPATCH_PROTOTYPES=1` flag

3) Fixes https://github.com/mono/mono/issues/14793
No idea how it worked on pre 10.15 without `-framework GSS` 🤔

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6112,6 +6112,10 @@ elif test x$target_osx = xyes; then
 	MONO_NATIVE_PLATFORM=macos
 	MONO_NATIVE_PLATFORM_TYPE="MONO_NATIVE_PLATFORM_TYPE_MACOS"
 
+	if test x$enable_gss = xyes; then
+		MONO_NATIVE_LDFLAGS="$MONO_NATIVE_LDFLAGS -framework GSS"
+	fi
+
 	AC_MONO_APPLE_AVAILABLE(mono_native_compat, [whether we need the compatibility layer],
 		[!(MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_12)])
 
--- a/configure.ac
+++ b/configure.ac
@@ -399,7 +399,7 @@ case "$host" in
 		parallel_mark="Disabled_Currently_Hangs_On_MacOSX"
 		host_darwin=yes
 		target_mach=yes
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -399,7 +399,7 @@ case "$host" in
 		parallel_mark="Disabled_Currently_Hangs_On_MacOSX"
 		host_darwin=yes
 		target_mach=yes
-		CPPFLAGS="$CPPFLAGS -D_THREAD_SAFE -DGC_MACOSX_THREADS -DUSE_MMAP -DUSE_MUNMAP"
+		CPPFLAGS="$CPPFLAGS -D_THREAD_SAFE -DGC_MACOSX_THREADS -DUSE_MMAP -DUSE_MUNMAP -DOBJC_OLD_DISPATCH_PROTOTYPES=1"
```
akoeplinger: the proper fix is going to land with https://github.com/mono/mono/issues/14792 :)

akoeplinger:
@monojenkins backport 2019-02
@monojenkins backport 2019-06

filipnavara:
> No idea how it worked on pre 10.15 without -framework GSS

It got pulled in through the dependencies. You can see it by running `DYLD_PRINT_LIBRARIES=1 csharp` and then trace it back.  On Mojave `/System/Library/Frameworks/GSS.framework/Versions/A/GSS` got loaded as dependency of `/usr/lib/libcups.2.dylib` and so on. Unlike Windows, macOS will resolve the symbols to whatever is already available so it was just happy that someone loaded GSS before libmono-native[-compat/-unified].

PR #14825, opened 2019-06-05
Make `CalendarData` more linker friendly.

baulig:
See https://github.com/mono/corefx/pull/298 for a description of the
technique that we're using here to make it more linker friendly.


baulig:
We need to bump CoreFX after it's PR has landed.

PR #14835, opened 2019-06-06
[netcore] Improve netcore/build.sh script

EgorBo:
Improve `netcore/build.sh` script:
1) Run `.autogen.sh` only once (creates `.configured` file as a sign - let me know if you don't like this file name)
2) Added `--rebuild` flag to run `.autogen.sh` despite the `.configured` file
3) Added `--skipnative` and `--skipmscorlib` for consistency with coreclr's build.sh
4) `xtestall` rule didn't run tests if it needed to download them first (so you had to run it twice)

Also, bump corefx and coreclr versions (to preview7) and ignore new failures in xtestall.

### DIFF
--- a/netcore/build.sh
+++ b/netcore/build.sh
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# Stop script if unbound variable found (use ${var:-} if intentional)
+set -u
+
+# Stop script if command returns non-zero exit code.
+# Prevents hidden errors caused by missing error code propagation.
+set -e
+
 usage()
 {
   echo "Common settings:"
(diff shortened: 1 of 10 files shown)


netcore/build.sh
```diff
@@ -61,24 +84,32 @@ while [[ $# > 0 ]]; do
...
 
-cd ..
-./autogen.sh --with-core=only
-cd netcore
+# run .././autogen.sh only once or if "--rebuild" argument is provided
+if [[ "$force_rebuild" == "true" || ! -f .configured ]]; then
+  cd ..
+  ./autogen.sh --with-core=only
+  cd netcore
+  touch .configured
```
steveisok: Should we isolate in the netcore folder?  I don't have an opinion other than that.

EgorBo: it's currently inside the netcore folder

steveisok: facepalm :-)

PR #14836, opened 2019-06-06
[mach] cast objc_msgSend{,Super} usages

### LINKED ISSUE #14792: [sdks][ios] Perform the appropriate cast of objc_msgSend and friends

lewurm:
Fixes #14792

### DIFF
--- a/mono/utils/mono-threads-mach-helper.c
+++ b/mono/utils/mono-threads-mach-helper.c
@@ -10,8 +10,6 @@
 
 #if defined(__MACH__)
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
Manually backported to 2019-06 and 2019-02.

Issue #14841, opened 2019-06-06
opt: Unknown command line argument '-place-safepoints' in XI 2019-06 integration


akoeplinger:
Failure from XI mtouch tests:

```
11:19:42.5262970 	Mono Ahead of Time compiler - compiling assembly /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/mtouch/bin/Debug/tmp-test-dir/Xamarin.Tests.BundlerTool.CreateTemporaryDirectory118/mtouch-test-cache/3-Build/System.Core.dll
11:19:42.5263120 	AOTID A7D5B5C4-EB0E-D304-60F7-BFACDAAF4C5B
11:19:42.5263620 	Executing opt: "/Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/LLVM36/bin/opt" -f -targetlibinfo -no-aa -basicaa -notti -instcombine -simplifycfg -inline-cost -inline -sroa -domtree -early-cse -lazy-value-info -correlated-propagation -simplifycfg -instcombine -simplifycfg -reassociate -domtree -loops -loop-simplify -lcssa -loop-rotate -licm -lcssa -loop-unswitch -instcombine -scalar-evolution -loop-simplify -lcssa -indvars -loop-idiom -loop-deletion -loop-unroll -memdep -gvn -memdep -memcpyopt -sccp -instcombine -lazy-value-info -correlated-propagation -domtree -memdep -adce -simplifycfg -instcombine -strip-dead-prototypes -domtree -verify -place-safepoints -spp-all-backedges -o "/Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/mtouch/bin/Debug/tmp-test-dir/Xamarin.Tests.BundlerTool.CreateTemporaryDirectory118/mtouch-test-cache/armv7/System.Core.dll.s.opt.bc" "/Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/mtouch/bin/Debug/tmp-test-dir/Xamarin.Tests.BundlerTool.CreateTemporaryDirectory118/mtouch-test-cache/armv7/System.Core.dll.s.bc"
[...]

PR #14852, opened 2019-06-06
[ios] switch cross32/crosswatch to be executed on x86_64 and use newer LLVM

lewurm:
Fixes https://github.com/mono/mono/issues/14841

Note that it breaks iOS 32bit on devices. The Apple Watch _might_ still works (only tested on `2019-02`).  It unblocks some build related issues on the Xamarin.iOS `2019-06` integration, and we have to do it anyway. I'll come back to it to fix device related issues.


akoeplinger:
@monojenkins backport 2019-06

PR #14853, opened 2019-06-06
Enable the compiler server by default

lewing:
Now that 2019-06 has branched, enable the compiler server by default


kg:
Seems fine. Once we merge this I'll send out another email. Do you feel like it's already discoverable how to turn it off, or should we add a .md somewhere?

marek-safar:
@monojenkins build failed

Issue #14152, opened 2019-04-21
Mono 5.20.1.19 / httpcfg on Linux fails with "Operation is not supported on this platform."


alvarengasv:
Registration of a certificate/port with httpcfg fails with mono 5.20.1.19 with error "Operation is not supported on this platform". The previous stable version worked correctly. If the certificate/port was registered with a previous mono version it is still usable with HttpListener after updating to 5.20.

## Steps to Reproduce
1. Create a certificate and private key
`makecert -r -sv test.pvk test.cer`
2. register the certificate with any port.
`httpcfg -add -port 9999 -cert test.cer -pvk test.pvk`

### Current Behavior
Fails with this output:
error loading certificate or private key [Operation is not supported on this platform.]

### Expected Behavior
Success

## On which platforms did you notice this

[ ] macOS
[x ] Linux
[ ] Windows

**Version Used**:
Mono JIT compiler version 5.20.1.19 (tarball Thu Apr 11 09:07:22 UTC 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          yes(600)
        Suspend:       hybrid
        GC:            sgen (concurrent by default)

blastagator:
Same problem here.  Anyone had any luck working around this?  I just updated to Ubuntu 19.04 and encountered this.

Martidjen:
Also seeing this issue. Started having issues with certificates after updating to mono 5.20.1.19. Tried to reinstall certificates but got the issue as described here.

blastagator:
If you really need this I was able to revert to my 18.10 build (VM backup) and do the following to downgrade (You should BACKUP your machine BEFORE attempting this, in case it goes wrong):

sudo vi /etc/apt/sources.list.d/<mono file.list>
	Change only line to:
	deb http://download.mono-project.com/repo/ubuntu stable-bionic/snapshots/5.18.1.3 main
sudo apt update
sudo apt-get install apt-show-versions
Run regex command to downgrade all:
	sudo apt-get install $(apt-show-versions | grep -P 'newer than version in archive' | awk -F: '{print $1"/bionic"}')

I was able to get back to 5.18.1.3 doing this.  I broke my machine once while trying to flesh this out, so you've been warned.

[...]

PR #14860, opened 2019-06-06
Implement RSACertificateExtensions.CopyWithPrivateKey().  #14152.

baulig:
Since `RSACertificateExtensions` lives in `System.Core` and only `mscorlib` internals (but not `System` internals) are visible to it, we need to use a little trick here:

* System.Security.Cryptography.X509Certificates.X509CertificateImpl:
  add new abstract methods CopyWithPrivateKey() and CreateCertificate().

* System.Security.Cryptography.X509Certificates.X509Certificate2Impl:
  sealed implement them here.

The `httpcfg` tool, only checks whether the key is valid, so we don't really need to call `CopyWithPrivateKey()` (which requires us to take a dependency on `System.Core`).

Fixes https://github.com/mono/mono/issues/14152.


baulig:
@monojenkins build failed

baulig:
@monojenkins backport 2019-02

baulig:
@monojenkins backport 2019-06

PR #14865, opened 2019-06-07
[wasm] Prototype for collecting aot profile data.

vargaz:


sdks/wasm/packager.cs
```diff
@@ -655,6 +655,8 @@ class WasmOptions {
 		}
 		if (ee_mode == ExecMode.AotInterp)
 			aot_args = "interp,";
+		if (ee_mode == ExecMode.Interp)
+			enable_zlib = true;
```
lewing: was this related?

mono/profiler/aot.c
```diff
@@ -488,9 +552,60 @@ prof_shutdown (MonoProfiler *prof)
...
+	if (prof->send_to) {
+		GHashTableIter iter;
+		gpointer id;
+		MonoImage *image;
+		MonoMethod *send_method = NULL;
+		MonoMethodSignature *sig;
+		ERROR_DECL (error);
+
+		g_hash_table_iter_init (&iter, prof->images);
+        while (g_hash_table_iter_next (&iter, (void**)&image, (void**)&id)) {
```
lewing: nit, indentation

sdks/wasm/packager.cs
vargaz: This is needed because the runtime is compiled against zlib, so in reality this option always needs to be enabled. In AOT mode, the native linker can get rid of the references to zlib, but not in interpreter mode.

lewing:
cc @radekdoulik

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14873, opened 2019-06-07
[Wasm] Fix memory override in mono_wasm_add_assembly

jeromelaban:
Fixed the real issue, where the malloc was not requesting the proper structure size.

~~This change avoids the inclusion of the emscripten `dlmalloc` library which seems to have stability issues (memory overwrite on allocation) using the release-dynamic configuration.~~

### DIFF
--- a/sdks/wasm/driver.c
+++ b/sdks/wasm/driver.c
@@ -149,7 +149,7 @@ mono_wasm_add_assembly (const char *name, const unsigned char *data, unsigned in
 		mono_register_symfile_for_assembly (new_name, data, size);
 		return;
 	}
-	WasmAssembly *entry = (WasmAssembly *)malloc(sizeof (MonoBundledAssembly));
+	WasmAssembly *entry = (WasmAssembly *)malloc(sizeof (WasmAssembly));
 	entry->assembly.name = m_strdup (name);
 	entry->assembly.data = data;
 	entry->assembly.size = size;


jeromelaban:
I renamed the PR to fix the real issue, the size of the malloc was incorrect, it was not related to any force imported library.

jeromelaban:

jeromelaban:
@monojenkins build OS X i386

jeromelaban:

sdks/wasm/driver.c
```diff
@@ -149,7 +149,7 @@ mono_wasm_add_assembly (const char *name, const unsigned char *data, unsigned in
 		mono_register_symfile_for_assembly (new_name, data, size);
 		return;
 	}
-	WasmAssembly *entry = (WasmAssembly *)malloc(sizeof (MonoBundledAssembly));
+	WasmAssembly *entry = (WasmAssembly *)malloc(sizeof (WasmAssembly));
```
jaykrell: We have `g_new` and `g_new0` for this.
I realize driver.c might not, but maybe copy them there if it doesn't.

jaykrell: Seems like a good reason for `auto` too -- state the type just once.

jeromelaban: glib.h is not available there, there may be a reason @vargaz knows about. Also, it does seem like C99 supports `auto`, is it preferred to use C++11 instead ?

jaykrell: Agreed -- we'd have to copy/paste/edit g_new and this is yet C.
They are just one line:

Searching for 'define g_new'...
```
C:\s\mono2\mono\eglib\glib.h(290):#define g_new(type,size)        ((type *) g_malloc (sizeof (type) * (size)))
C:\s\mono2\mono\eglib\glib.h(291):#define g_new0(type,size)       ((type *) g_malloc0 (sizeof (type)* (size)))
C:\s\mono2\mono\eglib\glib.h(292):#define g_newa(type,size)       ((type *) alloca (sizeof (type) * (size)))
```

jeromelaban: `g_malloc` also comes with glib, are there specifics in the `g_malloc` behavior, or a plain `malloc` is enough ?

jaykrell: You'd change the macro to plain `malloc` most likely.
  It seems the lesser evil, than to do w/o `g_new`?
Or maybe provide `g_malloc` that `abort ()` upon failure.
Or maybe preferably, check for null in `driver.c` even if Mono often does not.

lewing: This is a fine discussion for a separate PR.  I'm happy with landing this now

jaykrell: https://github.com/mono/mono/pull/14890

Issue #14730, opened 2019-05-31
Assertion at /Users/builder/jenkins/workspace/archive-mono/2019-02/mac/release/mono/utils/strenc.c:184, condition `utf8!=NULL' not met


rolfbjarne:
## Steps to Reproduce

Unfortunately I don't have a way to reproduce this, it's a random issue and it's hit twice today, with two different Xamarin.Mac tests.

Reference: https://github.com/xamarin/maccore/issues/1685

I have the crash info dumped to the terminal:

* https://gist.github.com/rolfbjarne/9c294a19c1a0fc9fd4883263a5cae1d3
* https://gist.github.com/rolfbjarne/56220c52d47db8bde35dfedbfc717386

and the native stack trace looks identical in both:

```
0x11001f3cf - test.app/Contents/MacOS/test : monoeg_assertion_message
0x110017dbe - test.app/Contents/MacOS/test : mono_unicode_to_external
0x1100090bf - test.app/Contents/MacOS/test : mono_pe_file_map
0x10ffa5bb1 - test.app/Contents/MacOS/test : mono_w32process_get_fileversion_info
0x10ffa7cc1 - test.app/Contents/MacOS/test : mono_w32process_get_fileversion
0x10ffa8a6b - test.app/Contents/MacOS/test : ves_icall_System_Diagnostics_Process_GetModules_internal
```

This is with xamarin-macios [from today](https://github.com/xamarin/xamarin-macios/commit/8e2bbe8d2dc8ddb74106504981ae58970a0263b1
); mono hash is bab25c0a80f7a9ad8f1f18be227119384e3ec0a0 (2019-02).

marek-safar:
@lambdageek any ideas about this?

lambdageek:
After #14879 and #15064 we should get better reports (including the raw hex bytes of the string that we're failing to convert) if this shows up again.

rolfbjarne:
We got a crash with the improved information:

```
...
16:33:30.7090040 	[PASS] ProcessTest.HasExitedCurrent
[...]

Issue #8986, opened 2018-05-31
RFC Consider dropping MONO_EXTERNAL_ENCODINGS and assume UTF8.


jaykrell:
## Steps to Reproduce

1. Notice mono_unicode_to_external, mono_utf8_from_external in the code.
2. Seems like a mess and maybe not consistent.
We use it for filenames, at least sometimes, for command lines, and for some security stuff. At least.
We often hardcode UTF8.

Is there a clear system and rationale to what is "external" vs. what is "UTF8", or should we just remove "external" and only use "UTF8"?

### Current Behavior

Sometimes we use "external", sometimes UTF8.
"external" defaults to UTF8, but has other behaviors.

### Expected Behavior

Clear code that works and is easy to  understand and maintain.
When I coop-convert, and move conversion to managed, I shouldn't be faced with this additional dilemma. Either reassert that "external" is good and leave it alone, or remove it all, or on a case by case basis, with a clear way to know each case.

## On which platforms did you notice this

[x] macOS
[x] Linux
[ ] Windows

marek-safar:
This is a fair point but I don't know the history behind that (cc @migueldeicaza @vargaz).

PR #14879, opened 2019-06-07
Propagate error in mono_unicode_to_external

CoffeeFlux:
The error propagation made sense in most places the function was called, and so they were updated accordingly. Notably, however, this removes the `g_assert` previous present in the function that was being hit as part of https://github.com/mono/mono/issues/14730. I've re-added it higher in the call stack along with a print of more diagnostic information to figure out exactly what's going on. It preserves existing behavior, so this is not going to cause additional new failures, and a later PR addressing the underlying cause of the bug will remove it.

Hopefully this time I've added the new file to the build system properly...

### DIFF
--- a/mono/metadata/w32file-unix.c
+++ b/mono/metadata/w32file-unix.c
@@ -48,6 +48,7 @@
 #include "w32file-unix-glob.h"
 #include "w32error.h"
 #include "fdhandle.h"
+#include "utils/mono-error-internals.h"
 #include "utils/mono-io-portability.h"
 #include "utils/mono-logger-internals.h"
 #include "utils/mono-os-mutex.h"
@@ -1948,7 +1949,7 @@ mono_w32file_create(const gunichar2 *name, guint32 fileaccess, guint32 sharemode
 	gchar *filename;
 	gint fd, ret;
 	struct stat statbuf;
-	GError *err = NULL;
+	ERROR_DECL (err);
 
 	if (attrs & FILE_ATTRIBUTE_TEMPORARY)
 		perms = 0600;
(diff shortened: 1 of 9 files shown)


jaykrell:
Runtime seems inconsistent on "external" vs. "utf8".
Perhaps external should just always be utf8.

CoffeeFlux:
@monojenkins build failed

mono/utils/strenc-internals.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _MONO_STRENC_INTERNALS_H_
+#define _MONO_STRENC_INTERNALS_H_
+
+#include <glib.h>
+
+gchar *mono_unicode_to_external_error (const gunichar2 *uni, GError **err);
```
lambdageek: I don't think we should expose `GError*` to the callers.

We should make `mono_unicode_to_external_checked (const guinchar2 *uni, MonoError *error)` and convert the `GError*` into a `MonoError` in its implementation.

jaykrell:
https://github.com/mono/mono/issues/8986

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

mono/metadata/w32file-unix.c
```diff
@@ -1947,6 +1949,7 @@ mono_w32file_create(const gunichar2 *name, guint32 fileaccess, guint32 sharemode
 	gchar *filename;
 	gint fd, ret;
 	struct stat statbuf;
+	ERROR_DECL (err);
```
lambdageek: nit: naming - we usually call the variable `error`. (here an in a few other places)

mono/utils/mono-proclib.c
```diff
@@ -1017,16 +1019,19 @@ mono_pe_file_map (gunichar2 *filename, gint32 *map_size, void **handle)
...
 
 	/* According to the MSDN docs, a search path is applied to
 	 * filename.  FIXME: implement this, for now just pass it
 	 * straight to open
 	 */
 
-	filename_ext = mono_unicode_to_external (filename);
+	filename_ext = mono_unicode_to_external_checked (filename, err);
+	// Assert added to diagnose https://github.com/mono/mono/issues/14730, remove after resolved
+	g_assertf (filename_ext != NULL, "%s: unicode conversion returned NULL; %s; 0x%hx", __func__, mono_error_get_message (err), filename);
```
lambdageek: I don't think this is quite what we want.   if `filename_ext` is null, this will print the conversion error message and just the first `short` from the address of `filename`.
I think we want something more like
```c
if (G_UNLIKELY (filename_ext == NULL)) {
    GString *raw_bytes = g_string_new (NULL);
    char *p = (char*)filename;
    while (*p != 0) {
        g_string_append_printf (raw_bytes, " %02X", (int)*p);
        p++;
    }
    g_assertf (filename_ext != NULL, "%s: unicode conversion returned NULL; %s; input was: %s", __func__, mono_error_get_message (err), raw_bytes->str);
    g_string_free (raw_bytes, TRUE);
}
```

mono/metadata/w32file-unix.c
CoffeeFlux: It seemed inconsistent so I just picked one - will switch over.

mono/utils/mono-proclib.c
CoffeeFlux: Yeah, whoops. Not sure what I was thinking there, but it's been updated.

CoffeeFlux:
@monojenkins build failed

lambdageek:

Issue #14843, opened 2019-06-06
LLVM extremely slow for unlinked Xamarin.iOS.dll in 2019-06


akoeplinger:
The "don't link" test application takes almost 25 minutes to compile in 2019-06 for the iOS 64 bit Device Release configuration, compared to about 3 minutes in 2019-02.

The bulk of the time is spent in Xamarin.iOS.dll.

2019-02:

```
14:59:29.1461680   Mono Ahead of Time compiler - compiling assembly /Users/alexander/dev/xamarin-macios/xamarin-macios-master/tests/xharness/tmp-test-dir/dont link587/obj/iPhone/Release64-unified/mtouch-cache/3-Build/Xamarin.iOS.dll (Aufgaben-ID: 112)
14:59:29.1462630   AOTID F8D95594-6050-30F7-D04C-E63BFD3C081E (Aufgaben-ID: 112)
14:59:29.1462820   Executing opt: "/Users/alexander/dev/xamarin-macios/xamarin-macios-master/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/LLVM/bin/opt" -f -O2 -disable-tail-calls -o "/Users/alexander/dev/xamarin-macios/xamarin-macios-master/tests/xharness/tmp-test-dir/dont link587/obj/iPhone/Release64-unified/mtouch-cache/arm64/Xamarin.iOS.dll.s.opt.bc" "/Users/alexander/dev/xamarin-macios/xamarin-macios-master/tests/xharness/tmp-test-dir/dont link587/obj/iPhone/Release64-unified/mtouch-cache/arm64/Xamarin.iOS.dll.s.bc" (Aufgaben-ID: 112)
[...]

PR #14880, opened 2019-06-07
[llvm] Fix disabling direct calls and inlining

alexanderkyte:
Previously we disabled too much runtime format when we disable direct calls. This fixes that.

By putting inlining behind a different flag than direct calls, you can tell if it's the direct call itself or the inlining it enables that leads to a specific bit of observed behavior.

### DIFF
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -3663,6 +3663,8 @@ mini_parse_debug_option (const char *option)
 		mini_debug_options.lldb = TRUE;
 	else if (!strcmp (option, "llvm-disable-self-init"))
 		mini_debug_options.llvm_disable_self_init = TRUE;
+	else if (!strcmp (option, "llvm-disable-inlining"))
+		mini_debug_options.llvm_disable_inlining = TRUE;
 	else if (!strcmp (option, "explicit-null-checks"))
 		mini_debug_options.explicit_null_checks = TRUE;
 	else if (!strcmp (option, "gen-seq-points"))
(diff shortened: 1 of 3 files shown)


akoeplinger:
I can confirm that with `llvm-disable-direct-calls` MONO_DEBUG option set it fixes the slow LLVM issue from https://github.com/mono/mono/issues/14843.

akoeplinger:
The latest version works without the `llvm-disable-direct-calls` option now :+1:

akoeplinger:

mono/mini/mini-runtime.c
```diff
@@ -3731,7 +3733,7 @@ mini_parse_debug_options (void)
...
+			fprintf (stderr, "Available options: 'handle-sigint', 'keep-delegates', 'reverse-pinvoke-exceptions', 'collect-pagefault-stats', 'break-on-unverified', 'no-gdb-backtrace', 'suspend-on-native-crash', 'suspend-on-sigsegv', 'suspend-on-exception', 'suspend-on-unhandled', 'dont-free-domains', 'dyn-runtime-invoke', 'gdb', 'explicit-null-checks', 'gen-seq-points', 'no-compact-seq-points', 'single-imm-size', 'init-stacks', 'casts', 'soft-breakpoints', 'check-pinvoke-callconv', 'use-fallback-tls', 'debug-domain-unload', 'partial-sharing', 'align-small-structs', 'native-debugger-break', 'thread-dump-dir=DIR', 'no-verbose-gdb', 'llvm_disable_inlining', 'llvm-disable-self-init'.\n");
```
jaykrell: It’d help diff/merge/readability if this line was split up.

alexanderkyte: Or kept in sync with the options list in a non-manual way

marek-safar:
@vargaz, please review

vargaz:
Shouldn't this be an option to llc/opt ?

alexanderkyte:
It's not an opt/llc argument because it's a mono-side feature to toggle on and off (whether to use the replace_all_uses_of at the end there). I think you meant as an option to "--aot", which is a good idea that was considered in channel when this feature was first merged. 

@vargaz the argument that was made in favor of MONO_DEBUG is that it's a bit harder to use it to check if the subsystem is working by toggling it on/off if you have to change args to the AOT compiler in the build files, rather than just exporting a symbol in the environment outside the build system, and having the child subprocess inherit it.

It's not all that much harder though. Could definitely also have the option to live in both ```--aot``` and MONO_DEBUG if you want that.

PR #14884, opened 2019-06-07
[2019-02] Implement RSACertificateExtensions.CopyWithPrivateKey().  #14152. (#14860).

baulig:
_Manual backport of https://github.com/mono/mono/pull/14860._

* Implement RSACertificateExtensions.CopyWithPrivateKey().  #14152.

* RSACertificateExtensions.CopyWithPrivateKey(): implement.

Since `RSACertificateExtensions` lives in `System.Core` and only `mscorlib`
internals (but not `System` internals) are visible to it, we need to use a
little trick here:

* System.Security.Cryptography.X509Certificates.X509CertificateImpl:
  add new abstract methods CopyWithPrivateKey() and CreateCertificate().

* System.Security.Cryptography.X509Certificates.X509Certificate2Impl:
  sealed implement them here.

The `httpcfg` tool, only checks whether the key is valid, so we don't really
need to call `CopyWithPrivateKey()` (which requires us to take a dependency
on `System.Core`).

* Do the same thing in HttpListener.

(cherry picked from commit ac908dd334e122e98e063a5c50a3ee66aa245cd7)

### DIFF
--- a/mcs/class/System.Core/System.Security.Cryptography.X509Certificates/RSACertificateExtensions.cs
+++ b/mcs/class/System.Core/System.Security.Cryptography.X509Certificates/RSACertificateExtensions.cs
@@ -30,7 +30,7 @@ namespace System.Security.Cryptography.X509Certificates
 {
 	public static class RSACertificateExtensions
 	{
-		public static RSA GetRSAPrivateKey(this X509Certificate2 certificate)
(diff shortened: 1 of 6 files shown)


akoeplinger:
@monojenkins commit apidiff

Issue #14848, opened 2019-06-06
Failed assertion crash in Mono threadpool-io.c when starting Couchbase Lite replication using CreateNLSocket on arm64-v8a Xamarin Android


bdgza:
## Steps to Reproduce

1. Open the attached solution "TravelSample" in Visual Studio 2019 for Mac / Visual Studio 2019, with Xamarin installed.
2. Run the TravelSample.Android project on a 64-bit Android device with Android 9 installed (I use a Samsung A8 (2018) with consistent results).
3. Tap the "SIGN IN" button.
4. Assertion fails and crash.

[mobile-travel-sample-arm64-v8a.zip](https://github.com/mono/mono/files/3262438/mobile-travel-sample-arm64-v8a.zip)

### Current Behavior

When running the Xamarin.Android app on a 64-bit capable device using arm64-v8a, when the Couchbase Lite replicator starts there is a hard crash in mono:
> Assertion at /Users/builder/jenkins/workspace/xamarin-android-d16-0/xamarin-android/external/mono/mono/metadata/threadpool-io.c:368, condition `fd >= 0’ not met

When the project only enables the armeabi-v7a everything performs normally. But 32-bit apps will no longer be served to 64-bit capable devices by the Google Play Store from 1 August 2019.

Confirmed to crash every time on Samsung A8 (2018) (Samsung SM-A530F) running latest Samsung security pack (Android 9 "Pie"). We are unable to confirm if it is related to Android Pie or only to 64-bit.

### Expected Behavior

The Couchbase Lite replicator should be able to start a replication under arm64-v8a on 64-bit capable Android devices without a crash in Mono.

## On which platforms did you notice this

[X] Xamarin.Android (64-bit devices, arm64-v8a)

**Version Used**: Mono 5.18.1.3 (2018-08/fdb26b0a445) (64-bit), Visual Studio 2019 for Mac 8.0.8, Xamarin.Android 9.2.3.0.
Also: Mono 5.16.0.221 (2018-06/b63e5378e38) (64-bit), Visual Studio 2017 for Mac 7.8.1, Xamarin.Android 9.1.8.0.
Also: Mono ??, Mono.Android 9.3.0.22, Visual Studio 2019 16.1.1, Xamarin.Android 9.3.0.22.

## Stacktrace

> [Mono] Searching for 'CreateNLSocket'.
> [Mono] Probing 'CreateNLSocket'.
> [Mono] Found as 'CreateNLSocket'.
[...]

PR #14888, opened 2019-06-07
[threadpool-io] Extra asserts to trackdown `fd >=0` assertion

jaykrell:
Diagnosability for #14848.

### DIFF
--- a/mono/metadata/threadpool-io.c
+++ b/mono/metadata/threadpool-io.c
@@ -82,11 +82,19 @@ typedef struct {
 
 typedef struct {
 	ThreadPoolIOUpdateType type;
+	// consider struct instead of union as paranoia against bugs
 	union {
 		ThreadPoolIOUpdate_Add add;
 		ThreadPoolIOUpdate_RemoveSocket remove_socket;
 		ThreadPoolIOUpdate_RemoveDomain remove_domain;
 	} data;
+
+	struct {
+		MonoDomain *domain;
+	} debug;
+
+	// consider a small log of writers
+	const char* last_writer;
 } ThreadPoolIOUpdate;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lambdageek:
We're probably not going to merge this (or at least not right away), but it may be useful to cherrypick this into a custom Xamarin.Android build to track down the underlying issue.

jaykrell:
It is asserts, not logging. Not to nitpick title, but would want to decide which is prefered.

jaykrell:
An in-memory log, ringbuffer, of the last n items, not a terrible idea.
And/or change union to struct, wasteful but probably not horrible.

mono/metadata/threadpool-io.c
```diff
@@ -651,20 +664,18 @@ mono_threadpool_io_remove_socket (int fd)
...
-	update = update_get_new ();
-	update->type = UPDATE_REMOVE_SOCKET;
-	update->data.add.fd = fd;
-	mono_memory_barrier (); /* Ensure this is safely published before we wake up the selector */
+	if (io_selector_running) {
 
-	selector_thread_wakeup ();
+		update = update_get_new (UPDATE_REMOVE_SOCKET, __func__);
+		g_assertf (fd >= 0, "%d", fd);
+		update->data.remove_socket.fd = fd;
```
lambdageek: Kind of lost in here, but we should take this change no matter what.
`update->data.add.fd = fd` to `update->data.remove_socket.fd = fd`

CoffeeFlux:
We tracked down that particular bug, so I don't think there's a need to merge this unless the refactoring changes are otherwise important.

jaykrell:
Agreed.

PR #14892, opened 2019-06-08
Small style cleanup.

jaykrell:
- split long lines for readability (opcode == foo || opcode == bar...)
 - merge short lines for searchability (define opdef)
 - un-else after return
 - foo = abc && def vs. foo = abc; foo &= def;
 - unsigned char to guint8 (or guchar is just as good)
 - remove not needed cast
 - args [n]; args [0] = x; args [1] = y vs. args [ ] = { x, y}
   (which does likely remove sequence points between x and y, ok
 - space before [ as is usual Mono style
 - no space at end of line

### DIFF
--- a/mono/mini/aot-runtime.c
+++ b/mono/mini/aot-runtime.c
@@ -174,12 +174,12 @@ static GHashTable *aot_modules;
 #define mono_aot_unlock() mono_os_mutex_unlock (&aot_mutex)
 static mono_mutex_t aot_mutex;
 
-/* 
+/*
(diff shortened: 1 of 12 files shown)


jaykrell:
@monojenkins build failed

lambdageek:
Very little of this is cleanup. It's mostly your personal preferred programing style.

jaykrell:
It is hard to read the long lines, harder to search the short lines, and not following mono style in several places.

jaykrell:
Also fixed sized arrays manually sized and over-sized at the same time.

PR #14894, opened 2019-06-08
[wasm] Incremental compilation support.

vargaz:


vargaz:
With these changes, a very minimal change to a hello world (like changing Hello to Bye) will build in 15s down from 30s for a full build.
10s of this is the final wasm linker, which seems to run some kind of optimization passes.

jaykrell:
@monojenkins build  Linux x64 C++

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #14897, opened 2019-06-08
Cleanup amd64 emit_call / emit_call_body -- last parameter always FALSE. And let compiler decide inline.

jaykrell:
And it is reasonable to expect that to remain the case, as C compilers
do not generate code as would occur with the last parameter = TRUE.

They allocate the room for their maximum outgoing parameters at function entry,
and keep a constant stack, except for functions that call _alloca,
and this code risks but not necessarily breaks ABI compliance.

In particular, the rsp adjustment either needs unwind codes, or needs
to only occur in a function with a frame pointer other than RSP.

And the uwind codes for adding to RSP are very tricky to impossible.
Normally you only ever subtract from RSP. There is an unwind code for that.
But there isn't really for add.
You might be able to simulate it with chained unwind, making these
regions look kinda like end of function.
But this is all most unlikely.

Also remove an inline. Let compiler decide.

### DIFF
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -3048,7 +3048,7 @@ mono_arch_finish_dyn_call (MonoDynCallInfo *info, guint8 *buf)
 
 #ifndef DISABLE_JIT
 static guint8*
-emit_call_body (MonoCompile *cfg, guint8 *code, MonoJumpInfoType patch_type, gconstpointer data)
+emit_call (MonoCompile *cfg, guint8 *code, MonoJumpInfoType patch_type, gconstpointer data)
 {
 	gboolean no_patch = FALSE;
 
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
Parameter was added in 2008 and not always FALSE then. Though not clear the use was correct then. Still looking..

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/mini/mini-amd64.c
```diff
@@ -3176,25 +3176,7 @@ emit_call_body (MonoCompile *cfg, guint8 *code, MonoJumpInfoType patch_type, gco
...
-		amd64_alu_reg_imm (code, X86_ADD, AMD64_RSP, 32);
-#endif
-
-	set_code_cursor (cfg, code);
-	
-	return code;
-}
-
-static inline int
+static int
```
jaykrell: Also removed inline, let compiler decide.

PR #14955, opened 2019-06-10
Another speculative fix for 'pkg-config' sometimes getting built as a…

alexischr:
… 32-bit package regardless of arch configuration

Mono issue: https://github.com/mono/mono/issues/13804

The hypothesis is that at reconfigure using the configure cache file, the lack of explicit setting of the host somehow breaks the build process.

The bump also includes a fix for package logs being too eagerly deleted which was impeding efforts to trace the process on CI.


alexischr:

alexischr:
@lambdageek could you review please?

Issue #14963, opened 2019-06-10
[netcore] Make System.SpanTests.SpanTests.MemoryExtensions_StaticReturningReadOnlySpan Pass


steveisok:
Fails the same as https://github.com/mono/mono/issues/14959

Only reason this issue exists is to track individual failures.

Issue #14961, opened 2019-06-10
[netcore] Make System.SpanTests.SpanTests.ReadOnlySpan_Property Pass


steveisok:
Fails in the same way as #14959

Only reason this issue exists is to track individual failures.

lewurm:
@steveisok @filipnavara is this fixed now?

filipnavara:
Yes, it is.

Issue #14960, opened 2019-06-10
[netcore] Make System.SpanTests.SpanTests.Span_Property Pass


steveisok:
Fails in the same way as https://github.com/mono/mono/issues/14959

Only reason this issue exists is to track individual failures.

lewurm:
@steveisok @filipnavara is this fixed now?

filipnavara:
Yes, it is.

PR #14967, opened 2019-06-10
[netcore] Update the checks for ByRef return type in reflection invocation

### LINKED ISSUE #14959: [netcore] Make System.SpanTests.SpanTests.MemoryMarshal_GenericStaticReturningSpan Pass

```
System.SpanTests.SpanTests.MemoryMarshal_GenericStaticReturningSpan [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.NotSupportedException)
      Actual:   typeof(System.Reflection.TargetInvocationException): Exception has been thrown by the target of an invocation.
      ---- System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
      -------- System.BadImageFormatException : Cannot box IsByRefLike type 'System.Span`1'
      Stack Trace:
           at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
           at System.SpanTests.SpanTests.<>c__DisplayClass3_0.<MemoryMarshal_GenericStaticReturningSpan>b__0()
        ----- Inner Stack Trace -----
           at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
           at System.SpanTests.SpanTests.<>c__DisplayClass3_0.<MemoryMarshal_GenericStaticReturningSpan>b__0()
        ----- Inner Stack Trace -----
```

Determine what's causing the problem and make the test pass.

filipnavara:
Update the checks for ByRef return type in reflection invocation to also check for IsByRefLike types. This was missed as part of #13901 and noted in the description.

Fixes #14959, #14960, #14961, #14963

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -3548,18 +3548,20 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
 		}
 	}
 
-	if (sig->ret->byref) {
 #if ENABLE_NETCORE
+	if (sig->ret->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
-		if (ret_byval->byref) {
+		if (ret_byval->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
 			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef to ByRefLike type via reflection"));
 			return NULL;
 		}
+	}
 #else
+	if (sig->ret->byref) {
 		mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef type via reflection"));
 		return NULL;
-#endif
 	}
+#endif
 
 	pcount = params? mono_array_length_internal (params): 0;
 	if (pcount != sig->param_count) {
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -3555,6 +3555,10 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
 #if ENABLE_NETCORE
 	if (sig->ret->byref) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
+		if (ret_byval->type == MONO_TYPE_VOID) {
+			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "ByRef to void return values are not supported in reflection invocation"));
+			return NULL;
+		}	
 		if (m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
 			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef to ByRefLike type via reflection"));
 			return NULL;
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -3548,10 +3548,14 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
 		}
 	}
 
+	if ((m->klass != NULL && m_class_is_byreflike (m->klass)) || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
+		mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method with stack pointers via reflection"));
+		return NULL;
+	}
 #if ENABLE_NETCORE
-	if (sig->ret->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
+	if (sig->ret->byref) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
-		if (ret_byval->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
+		if (m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
 			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef to ByRefLike type via reflection"));
 			return NULL;
 		}
--- a/mcs/tests/test-ref-07.cs
+++ b/mcs/tests/test-ref-07.cs
@@ -1,8 +1,13 @@
 // Compiler options: -langversion:latest
 
+public struct TestMain
+{
+	public static void Main () => Test.MainMethod();
+}
+
 public readonly ref partial struct Test
 {
-	public static void Main ()
+	public static void MainMethod ()
 	{
 		var m = new Test ();
 		m.Method ();
--- a/mcs/tests/ver-il-net_4_x.xml
+++ b/mcs/tests/ver-il-net_4_x.xml
@@ -73853,8 +73853,13 @@
     </type>
   </test>
   <test name="test-ref-07.cs">
-    <type name="Test">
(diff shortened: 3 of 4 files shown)


mono/metadata/icall.c
```diff
@@ -3548,18 +3548,20 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
...
+	if (sig->ret->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
-		if (ret_byval->byref) {
+		if (ret_byval->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
 			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef to ByRefLike type via reflection"));
 			return NULL;
 		}
+	}
 #else
+	if (sig->ret->byref) {
```
lambdageek: We check for ByRefLike for classic Mono, too.

```csharp
using System;

public ref struct Q {
	public int i;
}

public class Foo {
	static int v;

	public static void Main () {
		var mi = typeof(Foo).GetMethod ("Test");

		var o = mi.Invoke (null, null);
	}

	public static Q Test ()
	{
		Q q;
		q.i = 42;
		Console.WriteLine ("a");
		return q;
	}
}
```
(throws `BadImageFormatException`)

mono/metadata/icall.c
```diff
@@ -3548,18 +3548,20 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
 		}
 	}
 
-	if (sig->ret->byref) {
 #if ENABLE_NETCORE
+	if (sig->ret->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
-		if (ret_byval->byref) {
+		if (ret_byval->byref || m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
```
lambdageek: I'm a little bit confused about #13901 and https://github.com/dotnet/coreclr/pull/17639 - seems like the intention in CoreCLR is to forbid things like `ref Span<int>` (ByRef to ByRefLike as they say).

But we seem to be forbidding a wider class of things - for example it looks like we'll disallow `Span<int>`.

mono/metadata/icall.c
filipnavara: The behaviour was undefined under NetFX so I thought it is fine to leave it as is. I can change it to return the same exception if desired.

mono/metadata/icall.c
filipnavara: I don't think we disallow `Span<int>` unless I messed up the check. The idea is to box one ByRef/IsByRefLike but disallow any nested ones that would require more than one boxing. The failing tests were doing `Span<ref X>`.

mono/metadata/icall.c
filipnavara: To expand on my previous answer... NetFX didn't support `ref struct`s in the first place. Imposing the NetFX check on Mono could break existing code that was using `Span`s with reflection which happened to be working. We can either leave the code as-is and depend on the checks lower down the stack or we can enforce the full CoreFX behaviour even on regular Mono (which is more lenient than NetFX).

mono/metadata/icall.c
lambdageek: @filipnavara ah... I think `Span<ref int>` should be ruled out by disallowing instantiations with ByRef types...

in the case of `Span<int>`, `sig->ret->byref` is `False` and `m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret)))` is `True`.
But then `ret_byval` is `Span<int>` again (`m_class_get_byval_arg` is just the `Span<int>` again - not `int`,  I think like you expected).

So here's a sample:
<details>
<pre>
using System;

public ref struct Q {
	public int i;
}

public struct S {
	public int j;
}

public class Foo {
	static S s;

	public static void Main () {
		
		bool threw;

		threw = false;
		try {
			var mi1 = typeof(Foo).GetMethod ("Test1");
			var o1 = mi1.Invoke (null, null);
		} catch (NotSupportedException) { threw = true; }
		Console.WriteLine ($"mi1.Invoke: threw {threw} (expect False)");
			
		threw = false;
		try {
			var mi2 = typeof(Foo).GetMethod ("Test2");
			var o2 = mi2.Invoke (null, null);
		}  catch (NotSupportedException) { threw = true; }
		Console.WriteLine ($"mi1.Invoke: threw {threw} (expect False)");


		threw = false;
		try {
			var mi3 = typeof(Foo).GetMethod ("Test3");
			var o3 = mi3.Invoke (null, null);
		} catch (NotSupportedException) { threw = true; }
		Console.WriteLine ($"mi3.Invoke: threw {threw} (expect False)");

	}

	public static Q Test1 ()
	{
		Q q;
		q.i = 42;
		Console.WriteLine ("a");
		return q;
	}

	public static Span<int> Test2 ()
	{
		int[] a = new int[] {1, 4, 9, 16};
		return new Span<int>(a);
	}

	public static ref S Test3 ()
	{
		s.j = 196;
		return ref s;
	}

}
</pre>
</details>

It ends up printing:
```
mi1.Invoke: threw True (expect False)
mi1.Invoke: threw True (expect False)
mi3.Invoke: threw False (expect False)
```

lambdageek: For completeness, here's the same sample as https://github.com/mono/mono/pull/14967#discussion_r292610023 running on .NET Framework 4.8 (using VS2019 you need to add a reference to the System.Memory nuget to get a copy of `Span<T>`; but ref structs as a language feature are there already):

```
a
mi1.Invoke: threw False (expect True)
mi2.Invoke: threw False (expect True)
mi3.Invoke: threw True (expect False)
```

So... yea... at least we know that `#if ENABLE_NETCORE` is a crucial piece of the implementation here.

marek-safar:
@lambdageek are you ok with the change?

filipnavara:
I spoke to @lambdageek on Gitter on Gitter last week and he pointed out a flaw in my code that just accidentally happened to yield the correct result for the affected tests. I am on holiday now but I intend to revisit the PR next week when I get back unless someone looks into it earlier.

mono/metadata/icall.c
```diff
@@ -3548,18 +3548,24 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
 		}
 	}
 
-	if (sig->ret->byref) {
+	if ((m->klass != NULL && m_class_is_byreflike (m->klass)) || m_class_is_byreflike (mono_class_from_mono_type_internal (sig->ret))) {
+		mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method with stack pointers via reflection"));
+		return NULL;
+	}
```
filipnavara: This should be roughly equivalent to this condition:
https://github.com/dotnet/coreclr/blob/a64cb0a41a4ebeb4a61b8b0f1f2eeeff8dd539c9/src/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.cs#L54-L56

mono/metadata/icall.c
```diff
@@ -3548,18 +3548,24 @@ ves_icall_InternalInvoke (MonoReflectionMethod *method, MonoObject *this_arg, Mo
...
 #if ENABLE_NETCORE
+	if (sig->ret->byref) {
 		MonoType* ret_byval = m_class_get_byval_arg (mono_class_from_mono_type_internal (sig->ret));
-		if (ret_byval->byref) {
+		if (m_class_is_byreflike (mono_class_from_mono_type_internal (ret_byval))) {
 			mono_gc_wbarrier_generic_store_internal (exc, (MonoObject*) mono_exception_from_name_msg (mono_defaults.corlib, "System", "NotSupportedException", "Cannot invoke method returning ByRef to ByRefLike type via reflection"));
 			return NULL;
 		}
+	}
```
filipnavara: This should be roughly equivalent to this condition:
https://github.com/dotnet/coreclr/blob/a64cb0a41a4ebeb4a61b8b0f1f2eeeff8dd539c9/src/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.cs#L68-L72

`ref void` is not handled properly yet.

filipnavara:
@lambdageek PTAL.

filipnavara:
@marek-safar This is mergable now provided that the test failures are unrelated.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
They seem to be related, but I've removed the label.

filipnavara:
Ok, I will check them.

filipnavara:
The build failures are related, most of them are caused by https://github.com/mono/mono/blob/master/mcs/tests/test-ref-07.cs. I checked that it works just fine on .NET Core. The issue seems to be related to how Mono executes the Main method and it hits the check which forbids reflection calls on IsByRefLike structs.

Issue #12577, opened 2019-01-23
HttpClient does not respect Timeout


grendello:
_From @csuzw on December 17, 2018 9:15_

I believe this is the same/similar to this issue for Xamarin.iOS: https://github.com/xamarin/xamarin-macios/issues/5190 .  There is a hint in the original issue linked in that IOS issue that using the AndroidClientHandler and setting the ReadTimeout/ConnectTimeout is a workaround but I get the an exception about encoding 'identity' not being supported if I do this.

### Steps to Reproduce

1. Create a HttpClient and set Timeout to a value greater than 100s ( the default value).
2. Send a request that takes longer than 100s but less than the Timeout set.

### Expected Behavior

The request to return successfully

### Actual Behavior

The request times out after 100s.

### Version Information

Microsoft Visual Studio Enterprise 2017  (2)
Version 15.9.4
VisualStudio.15.Release/15.9.4+28307.222
Microsoft .NET Framework
Version 4.7.03190

Installed Version: Enterprise

Visual C++ 2017   00369-60000-00001-AA999
Microsoft Visual C++ 2017

Application Insights Tools for Visual Studio Package   8.14.11009.1
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2017   15.9.04012.0
ASP.NET and Web Tools 2017

ASP.NET Core Razor Language Services   15.8.31590
Provides languages services for ASP.NET Core Razor.

ASP.NET Web Frameworks and Tools 2017   5.2.60913.0
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   15.9.03024.0
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   15.9.02046.0
Azure Functions and Web Jobs Tools

C# Tools   2.10.0-beta2-63501-03+b9fb1610c87cccc8ceb74a770dba261a58e39c4a
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.1.49 (remotes/origin/d15-8@ee674f3)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

HgSccPackage   2.0.5
Mercurial source code control package

JavaScript Language Service   2.0
JavaScript Language Service

JavaScript Project System   2.0
JavaScript Project System

JavaScript UWP Project System   2.0
[...]

PR #14970, opened 2019-06-10
[System.Net.Http]: Add hack to the legacy HttpClient to pass down Timeout property. #12577.

baulig:
* `HttpClientHandler`, `IMonoHttpClientHandler`: add new internal `MonoSetTimeout (TimeSpan)` function.

* `HttpClient.SendAsyncWorker()`: if `handler` is `HttpClientHandler`, call the new internal
  `MonoSetTimeout()` function to pass down the timeout.

* `MonoWebRequestHandler.SendAsync()`: this is only used here to pass the timeout to the `HttpWebRequest`.

### DIFF
--- a/mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
+++ b/mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
@@ -80,6 +80,6 @@ namespace System.Net.Http
 		Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, CancellationToken cancellationToken);
 
 		// Only used by MonoWebRequestHandler and ignored by the other handlers.
-		void MonoSetTimeout (TimeSpan timeout);
+		void SetWebRequestTimeout (TimeSpan timeout);
 	}
 }
--- a/mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
+++ b/mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
@@ -78,5 +78,8 @@ namespace System.Net.Http
 		}
 
 		Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, CancellationToken cancellationToken);
+
+		// Only used by MonoWebRequestHandler and ignored by the other handlers.
+		void MonoSetTimeout (TimeSpan timeout);
 	}
 }
--- a/mcs/class/System.Net.Http/MonoWebRequestHandler.cs
+++ b/mcs/class/System.Net.Http/MonoWebRequestHandler.cs
@@ -67,6 +67,7 @@ namespace System.Net.Http
 		bool unsafeAuthenticatedConnectionSharing;
 		bool sentRequest;
 		string connectionGroupName;
+		TimeSpan? timeout;
 		bool disposed;
 
 		internal MonoWebRequestHandler ()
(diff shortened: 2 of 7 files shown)


mcs/class/System.Net.Http/System.Net.Http/HttpClient.cs
```diff
@@ -41,6 +41,7 @@ public partial class HttpClient : HttpMessageInvoker
 		CancellationTokenSource cts;
 		bool disposed;
 		HttpRequestHeaders headers;
+		HttpMessageHandler handler;
```
marek-safar: The base type already have the handler stored

mcs/class/System.Net.Http/MonoWebRequestHandler.cs
```diff
@@ -446,6 +447,9 @@ public async Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, Ca
 			var wrequest = CreateWebRequest (request);
 			HttpWebResponse wresponse = null;
 
+			if (timeout != null)
```
marek-safar: All setup calls are in CreateWebRequest I'd move it there too

mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
```diff
@@ -78,5 +78,8 @@ interface IMonoHttpClientHandler : IDisposable
 		}
 
 		Task<HttpResponseMessage> SendAsync (HttpRequestMessage request, CancellationToken cancellationToken);
+
+		// Only used by MonoWebRequestHandler and ignored by the other handlers.
+		void MonoSetTimeout (TimeSpan timeout);
```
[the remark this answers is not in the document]
baulig: The comment above it explains it.   We can use something like "Legacy" instead, but I strongly suggest keeping the prefix because it is a "Mono-only" method that will have no effect in the other handlers.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.cs
baulig: It is inaccessible there.  There choice we have is one of the following:

1. Add a private field to the legacy handler that's eventually getting removed anyways.
2. Add Mono-specific hack here: https://github.com/mono/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpMessageInvoker.cs#L16
3. Discuss with the CoreFX team why exactly 2. should be upstreamed

mcs/class/System.Net.Http/MonoWebRequestHandler.cs
baulig: Sure, good idea.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.cs
marek-safar: I think this code is used only for "legacy" mode where we have all sources in mcs, making the base field `private protected` should then just work

mcs/class/System.Net.Http/IMonoHttpClientHandler.cs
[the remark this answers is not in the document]
marek-safar: Use different prefix, it's not legacy either from the description it looks like it should be `SetWebRequestTimeout` or something like that

baulig:
@marek-safar Should be all done now, can you please review again?

baulig:
@monojenkins build failed

steveisok:
@monojenkins build failed

marek-safar:
@monojenkins backport 2019-02

marek-safar:
@monojenkins backport 2019-06

marek-safar:
@baulig please backport manually

PR #14849, opened 2019-06-06
[netcore] Run CoreFX tests on CI (macOS)

EgorBo:
Run all CoreFX tests on CI.
In order to see the test results:
1) Click on `mono-coreonly-CI` lane
2) Click on `View more details on Azure Pipelines` link
3) Click on `Tests` tab

Should look like this:
image

Command to re-run the lane: `/azp run`

PR #15015, opened 2019-06-12
[netcore] Enable CoreFX tests on CI for Linux

EgorBo:
See https://github.com/mono/mono/pull/14849#issuecomment-501151916


EgorBo:
@marek-safar should now run linux tests with 0 failures.
(and we need to fix the dllmap issue)

Issue #14940, opened 2019-06-10
[Wasm] WebAssembly.Net.Http.HttpClient fails under AOT


jeromelaban:
## Steps to Reproduce

1. https://github.com/jeromelaban/Wasm.Samples/tree/master/HttpClientRepro
2. `msbuild /r /bl /p:Configuration=Release`
3. Browse using the `server.py` in the `bin/Release/netstandard2.0/dist` folder.

### Current Behavior
The first call fails with `RuntimeError: index out of bounds`
The second call works, but raises a native assert.

```
Loaded: aot-dummy.dll mono.js:6838:16
initializing mono runtime mono.js:6844:17
>>mono_wasm_runtime_ready mono.js:6759:14
/mnt/jenkins/workspace/test-mono-mainline-wasm/label/ubuntu-1804-amd64/mono/metadata/icall.c:1632: mono.js:2181:13
﻿WebAssembly.JSException: RuntimeError: index out of bounds mono.js:2181:13
mono.js:2181:13
/mnt/jenkins/workspace/test-mono-mainline-wasm/label/ubuntu-1804-amd64/mono/metadata/icall.c:1632: mono.js:2181:13
[{"CountryCode":"USA","CountryName":"United States","CurrencyCode":"USD","CurrencyName":"Dollar","CurrencySymbol":"$"},…
```

### Expected Behavior

Both calls to the GetStringAsync work, without raisong runtime errors.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[ ] Windows
[X] WebAssembly

**Version Used**:

## Stacktrace

```
Please paste the stack trace here if available.
```

kjpou1:
Hello @jeromelaban 

Maybe I am doing something wrong on my end but not able to reproduce

Build with
```
iMac-Pro-2:HttpClientRepro Jimmy$ msbuild /r /bl /p:Configuration=Release
```

Output:
```
 Running 
[...]

PR #15018, opened 2019-06-12
[wasm][http] Fix premature disposal of object causing crash in HttpClient

kjpou1:
Fix problem with fetch response header iteration prematurely disposing of headers object.

- Modify `WasmHttpMessageHandler.cs` to use the `Headers` iterator instead of it's `forEach` method to iterate over the header values returned.
- Add regression tests.

Fixes: https://github.com/mono/mono/issues/14940


kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

lewing:
@monojenkins build failed

kjpou1:
@monojenkins build failed

kjpou1:
> From looking at this, I also don't understand why this is needed - how was the old code previously disposing objects and why do we need this private class instead of some more simple approach?

Will change this to simpler approach.

kjpou1:
@baulig 

Simplified the code.  Ready to take another look when you want.

baulig:
> @baulig
> 
> Simplified the code. Ready to take another look when you want.

Looks good :-)

lewing:
@monojenkins build failed

PR #15020, opened 2019-06-12
Add read barrier to mono_lazy_initialize().

jaykrell:
This is unfortunate.

The general pattern of:

```
if !initialized
  initialize the data, possibly with lots of locks and barriers
  initialized = TRUE
end
use the data
```
is racy, because "use the data" can be scheduled ahead of "if initialized".

"Barriers come in pairs" generally, and this was missing one.

It depends somewhat. If the data is all pointers, initialized
from null to non-null, and use includes dereferencing, which is a common
but not universal case, then it is ok on all non-Alpha processors, due
to "data dependency".

But if the data includes reading globals, then there is a race.

This is a bit of a slow down on fast paths on arm, and possibly
other architectures.

There are barrier-free ways to solve this, involving a thread local,
that should seriously be considered, and applied throughout the runtime.

The runtime has a lot of on-demand initialization and a lot looks racy.

See www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2660.htm (scroll all the way to the end).
Which would have to be adapted to be coop-friendly which should not be difficult.

Notice that the existing "lazy" mechanism is also not coop-friendly.
That is not changed by this PR.

### DIFF
--- a/mono/utils/mono-lazy-init.h
+++ b/mono/utils/mono-lazy-init.h
@@ -62,8 +62,10 @@ mono_lazy_initialize (mono_lazy_init_t *lazy_init, void (*initialize) (void))
 
 	status = *lazy_init;
 
-	if (status >= MONO_LAZY_INIT_STATUS_INITIALIZED)
+	if (status >= MONO_LAZY_INIT_STATUS_INITIALIZED) {
+		mono_memory_read_barrier ();
 		return status == MONO_LAZY_INIT_STATUS_INITIALIZED;
+	}
 	if (status == MONO_LAZY_INIT_STATUS_INITIALIZING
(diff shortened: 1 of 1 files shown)


jaykrell:
Note that it might be implemented as a full barrier.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
There is overlap between this and another.

I accidentally placesd the barrier in a different place, but it doesn't really matter.
This one places it only in the fast path.
The other places it in the fast and slow path.

jaykrell:
The other essentially is a super set of this.
If you want more smaller PRs fixing one thing at a time though..

jaykrell:
Maybe abandon this and just take https://github.com/mono/mono/pull/15126.

lewing:
Closing in favor of  #15126

PR #15038, opened 2019-06-12
[wasm] Simplify AOT profiling initialization.

vargaz:
Current steps:
- Call MONO.mono_wasm_init_aot_profiler (null) before calling MONO.mono_load_runtime_and_bcl ().
- Call WebAssembly.Runtime.StopProfile () from the c# code to stop profiling.
- The profile data will be in Module.aot_profile_data.


vargaz:
@monojenkins build failed

PR #15047, opened 2019-06-12
[eglib] Don't escape chars in argv parsing enclosed in single quotes

### LINKED ISSUE #14724: eglib g_shell_parse_argv() treats \ in ' quote differently from glib
eglib `g_shell_parse_argv()` currently treats a \ in a single quote (') as an escape character, while glib (and POSIX shell) will treat it as a normal character (everything except ' is treated as a normal character inside single quotes).

The following code shows the difference:

````
#include "glib.h"
#include <stdio.h>
/*
gcc test.c -o test-glib $(pkg-config --cflags --libs glib-2.0)
gcc test.c -o test-eglib -Imono/eglib/src mono/eglib/src/.libs/libeglib.a
*/
int main () {
  gint argc;
  gchar** argv;
  if (g_shell_parse_argv("' a \\\\ b '", &argc, &argv, 0)) {
    printf ("%d %s\n", argc, argv[0]);
  }
  return 0;
}
````

With glib this will output:
````
1  a \\ b 
````

With eglib it will output:
````
1  a \ b 
````

The effect of this is that backslashes have to be escaped even in single quotes when launching processes with `Process.Start()`. This behaviour should either be changed to match glib/shell, or the difference should be documented (changing the behaviour might break C# applications which rely on the current behaviour).

CoffeeFlux:
Fixes #14724 

This may need to be reverted if customers are relying on the existing behavior, but we can cross that bridge when we come to it.

The larger issue of incompatibility with .NET Core should probably be a separate issue that I'll open once I hear back from the Core folks, but I think I can mark the original issue closed once the single quote bug specifically is resolved.

### DIFF
--- a/mono/eglib/gshell.c
+++ b/mono/eglib/gshell.c
@@ -62,7 +62,7 @@ split_cmdline (const gchar *cmdline, GPtrArray *array, GError **gerror)
 					g_ptr_array_add (array, g_string_free (str, FALSE));
 					str = g_string_new ("");
 				}
-			} else if (c == '\\'){
+			} else if (c == '\\' && quote_char == '\"'){
 				escaped = TRUE;
 			} else 
 				g_string_append_c (str, c);


CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

Issue #15041, opened 2019-06-12
[wasm] diagnose aot the warning when loading types from bindings assemblies


lewing:
When using the http bindings the type lookups result in a warning when looking up the assemblies.

PR #15051, opened 2019-06-13
[wasm][bcl] Change how HttpClient finds wasm message handler

kjpou1:
Follow the XI/XA method of creating wasm's `HttpMessageHandler`

- Modify System.Net.Http Makefile to add a reference to reference assemblies.
- Make the `WebAssembly.Net.Http.HttpClient` visisble to `System.Net.Http`.
- Fixes stack overflow exception on any errors by removing the call to `GetFallback` and issuing `PlatformNotSupportedException` directly.
- Introduced a WebAssembly.RuntimeOptions

NOTE: Will not pass CI build until reference assembly exists.  PR: https://github.com/mono/reference-assemblies/pull/19

Closes https://github.com/mono/mono/issues/15059
Closes https://github.com/mono/mono/issues/15041


marek-safar:
@lewing I think we discussed that should follow XI/XA and hook the same way to remove all the reflection overhead

lewing:
The plan is still to remove the static variable but I think it makes sense to allow an an override so that we can allow experimentation outside the browser.  I'd also like to avoid the hard dep on the framework bindings there. I think we should probably expose some runtime options above the js bindings and handle the override there.

lewing:
So follow the ios pattern (with the default ctor).  Is android planning on sticking with the reflection bits?

marek-safar:
> Is android planning on sticking with the reflection bits

They just didn't catch up with the changes

sdks/wasm/framework/src/WebAssembly.Net.Http/WasmHttpMessageHandler.cs
```diff
@@ -2,11 +2,14 @@
...
 using System.Net;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using WebAssembly.Core;
 using WebAssembly.Host;
 
+[assembly: InternalsVisibleTo ("System.Net.Http")]
+
```
baulig: Why is this not strong-named?

We are using this in `System.dll`:
https://github.com/mono/mono/blob/e6e9d5f6eb5488d9f87ab39727d76fb2ca81702c/mcs/class/System/Assembly/AssemblyInfo.cs#L71

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -9,41 +9,14 @@ public partial class HttpClient
 
 #pragma warning disable 649
         private static Func<HttpMessageHandler> GetHttpMessageHandler;
-        // Used to skip the type lookup on later calls.
```
baulig: If you take this approach of making internals visible and then calling directly into `WasmHttpMessageHandler.GetHttpMessageHandler`, then you also do not need this `Func<HttpMessageHandler> GetHttpMessageHandler` anymore.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -9,41 +9,14 @@ public partial class HttpClient
...
-                var handler = ret as HttpMessageHandler;
-                if (handler == null)
-                    throw new PlatformNotSupportedException ($"{ret?.GetType()} is not a valid HttpMessageHandler");
-
-                return handler;
-
+                return WebAssembly.Net.Http.HttpClient.WasmHttpMessageHandler.GetHttpMessageHandler();
             }
             else
             {
```
baulig: Where is this used?

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -9,41 +9,14 @@ public partial class HttpClient
...
-                    throw new PlatformNotSupportedException ("WebAssembly returned no custom HttpClientHandler");
-
-                var handler = ret as HttpMessageHandler;
-                if (handler == null)
-                    throw new PlatformNotSupportedException ($"{ret?.GetType()} is not a valid HttpMessageHandler");
-
-                return handler;
-
+                return WebAssembly.Net.Http.HttpClient.WasmHttpMessageHandler.GetHttpMessageHandler();
             }
```
baulig: This also looks ugly.

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -16,39 +16,16 @@ internal static HttpMessageHandler CreateDefaultHandler()
```
baulig: How about

```cs
using WebAssembly.Net.Http.HttpClient;
```

at the top and then replace this entire body with

```cs
return WasmHttpMessageHandler.GetHttpMessageHandler ();
```

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kjpou1: Addressed in recent commit

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kjpou1: Addressed in recent commit

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kjpou1: This will be removed at a later date.  There are still users using this until we get them moved over.

kjpou1: > The plan is still to remove the static variable but I think it makes sense to allow an an override so that we can allow experimentation outside the browser.

sdks/wasm/framework/src/WebAssembly.Net.Http/WasmHttpMessageHandler.cs
kjpou1: Addressed in recent commits

kjpou1:
>  I think we should probably expose some runtime options above the js bindings and handle the override there.

Agreed

lewing:
@monojenkins build failed

mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
kjpou1: Introduced a `WebAssembly.RuntimeOptions` modeled after XI/XA.

kjpou1:
@lewing Introduced a `WebAssembly.RuntimeOptions` in previous commit.

Project still won't build in CI until some more parts are synced like the external api reference submodule.

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

PR #14870, opened 2019-06-07
Remove JIT icall hashing by address.

jaykrell:
Due to https://github.com/mono/mono/pull/11941#issuecomment-449157323.

 -  Work toward removing JIT icall hashing by address.
Mostly remove it, assert it isn't needed/used.

 - Actually remove JIT icall hashing by address.

PR #15056, opened 2019-06-13
Fix S390x compile since #14870 and a warning.

jaykrell:


lewing:
@monojenkins build failed

sharkcz:
and confirmation from me that s390x builds again OK

PR #15057, opened 2019-06-13
[netcore] Clean up netcore build

marek-safar:


mono/Makefile.am
```diff
@@ -16,6 +16,10 @@ if ENABLE_NETCORE
 btls_dirs = 
 endif
 
+if ENABLE_NETCORE
+SUBDIRS = eglib arch utils sgen metadata mini
+else
+
```
vargaz: Aren't the profilers needed ?

marek-safar: that's unclear, we'll need to hook into existing profilers setup for netcore

netcore/Makefile
```diff
@@ -51,7 +51,7 @@ TEST_ASSETS_URL = $(FEED_BASE_URL)/corefx-tests/$(NETCORETESTS_VERSION)/$(TESTS_
 BLAME = $(shell git blame ../configure.ac | grep AC_INIT | cut -f1 -d' ')
 VERSTUB = .$(shell git rev-list --count $(BLAME)..HEAD)
 
-all: runtime bcl
+all: bcl
```
EgorBo: why `runtime` was removed here?

steveisok: @marek-safar we wanted to be able to build netcore from scratch by just using netcore/build.sh.

marek-safar: It was causing circular build dependencies. I don't think we can build with just that script. You need to run `autogen` first and then `make`. Let's stay with this setup.

PR #15061, opened 2019-06-13
Deadlock in loader when using bundling.

lateralusX:
We can hit the following deadlock when loading assemblies (using reflection) together with bundling:

One thread calls mono_assembly_open_from_bundle, that will take assemblies lock and then try to load assembly. If this race with a different load of the same assemblies, one will lose, and that method will call mono_image_close that in turn will call unload hook (mono_class_unregister_image_generic_subclasses) that will take loader lock.

If we at the same time have a different thread that calls mono_class_create_from_typedef , it will take loader lock and then it might end up calling mono_assembly_load_reference that will take assemblies lock, but since that thread takes the locks in different order compare to first thread, they could deadlock.

Looking into the use of assemblies lock in mono_assembly_open_from_bundle it uses a static variable, bundles, that is not protected in other scenarios and only set by mono_register_bundled_assemblies, normally called during boot. The method called in the loop mono_image_open_from_data_internal is called at several places without locks (and locks internal using image lock), so that doesn't need to be protected. The rest is local data access and the only thing the lock could do is serialize the whole loop between threads (but that shouldn't be needed).

Fix will remove the locking and also eliminating the incorrect lock order causing the deadlock.

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -2168,14 +2168,12 @@ mono_assembly_open_from_bundle (const char *filename, MonoImageOpenStatus *statu
 	is_satellite = g_str_has_suffix (lowercase_filename, ".resources.dll");
 	g_free (lowercase_filename);
 	name = g_path_get_basename (filename);
-	mono_assemblies_lock ();
 	for (i = 0; !image && bundles [i]; ++i) {
 		if (strcmp (bundles [i]->name, is_satellite ? filename : name) == 0) {
 			image = mono_image_open_from_data_internal ((char*)bundles [i]->data, bundles [i]->size, FALSE, status, refonly, FALSE, name);
 			break;
 		}
 	}
-	mono_assemblies_unlock ();
 	if (image) {
 		mono_image_addref (image);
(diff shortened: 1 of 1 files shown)


mono/metadata/assembly.c
```diff
@@ -2168,14 +2168,12 @@ mono_assembly_open_from_bundle (const char *filename, MonoImageOpenStatus *statu
 	is_satellite = g_str_has_suffix (lowercase_filename, ".resources.dll");
 	g_free (lowercase_filename);
 	name = g_path_get_basename (filename);
-	mono_assemblies_lock ();
 	for (i = 0; !image && bundles [i]; ++i) {
 		if (strcmp (bundles [i]->name, is_satellite ? filename : name) == 0) {
 			image = mono_image_open_from_data_internal ((char*)bundles [i]->data, bundles [i]->size, FALSE, status, refonly, FALSE, name);
```
vargaz: I think we should keep the lock, and call the open function outside the lock.

lateralusX: Any specific reasons that comes to mind keeping it? Doing unlock/lock around mono_image_open_from_data_internal will not give us much more safety for the loop and access to bundles is not protected using a lock in other places in this source file.

vargaz: It looks like bundles are inited before runtime startup so no locking is needed.

lewing:
@monojenkins build failed

PR #15083, opened 2019-06-13
Add the compiler server log to the gitignore

kg:
I failed to do this when we first added it.


lewing:
@monojenkins build failed

lewing:
@monojenkins build failed

PR #15098, opened 2019-06-13
[api] Add a mono/metadata/profiler-legacy.h header

lambdageek:
Add a header with the legacy Mono profiler API and mark every function in it with `MONO_DEPRECATED`.

New code should use the API from mono/metadata/profiler.h.

---
No functional change. All these symbols were already exported by mono/metadata/profiler.c.  Now we just have a proper header for them.

### DIFF
--- a/mono/metadata/Makefile.am
+++ b/mono/metadata/Makefile.am
@@ -477,7 +477,6 @@ libmonoruntimeinclude_HEADERS = \
 	opcodes.h		\
 	profiler.h		\
 	profiler-events.h	\
-	profiler-legacy.h	\
 	reflection.h		\
 	row-indexes.h		\
 	sgen-bridge.h		\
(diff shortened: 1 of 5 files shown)


lambdageek:
Initially we thought Xamarin.Android wants this; but they're actually going to switch to the new profiler API.

So there's no longer a pressing need to do this.  But:
1. it was weird to have `MONO_API` declarations in a `.c` file
2. maybe other embedders still want this

We could also move the declarations to `profile-legacy.h` and then _not_ install that header.

mono/metadata/Makefile.am
```diff
@@ -476,6 +477,7 @@ libmonoruntimeinclude_HEADERS = \
 	opcodes.h		\
 	profiler.h		\
 	profiler-events.h	\
+	profiler-legacy.h	\
```
lambdageek:
```suggestion
```

alexrp:
IIRC, Unity was one of the embedders that relied on the legacy APIs when I decided to add them back (and then of course there was XA/XI/XM).

PR #15105, opened 2019-06-13
[mini] Move some MONO_API functions to public header

### LINKED ISSUE #15099: [Xamarin.Android] Missing Mono public symbol declarations in the headers shipped with the Mono archive
`Xamarin.Android` is switching away from `dlopen/dlsym` to link the Mono runtime directly and therefore we need to be able to include Mono header files that declare the functions we use. Most of them are already there, however there is a small set that isn't:

- [mono_aot_register_module](https://github.com/mono/mono/blob/e6e9d5f6eb5488d9f87ab39727d76fb2ca81702c/mono/mini/aot-runtime.h#L272)
- [mono_set_use_llvm](https://github.com/mono/mono/blob/e6e9d5f6eb5488d9f87ab39727d76fb2ca81702c/mono/mini/mini-runtime.c#L126) (perhaps it would be better to introduce a function to set the flag?)
- [mono_jit_thread_attach](https://github.com/mono/mono/blob/e6e9d5f6eb5488d9f87ab39727d76fb2ca81702c/mono/mini/mini-runtime.h#L455)

lambdageek:
And mark them all `MONO_RT_EXTERNAL_ONLY`.

`mono_set_use_llvm` is new - Xamarin.Android previously assigned to `mono_use_llvm` directly.

Fixes #15099

### DIFF
--- a/mono/mini/jit.h
+++ b/mono/mini/jit.h
@@ -100,6 +100,16 @@ mono_jit_parse_options     (int argc, char * argv[]);
 
 MONO_API char*       mono_get_runtime_build_info    (void);
 
+MONO_API MONO_RT_EXTERNAL_ONLY void
+mono_set_use_llvm (mono_bool use_llvm);
+
+MONO_API MONO_RT_EXTERNAL_ONLY void
+mono_aot_register_module (gpointer *aot_info);
+
+MONO_API MONO_RT_EXTERNAL_ONLY
+MonoDomain* mono_jit_thread_attach (MonoDomain *domain);
+
+
 MONO_END_DECLS
 
 #endif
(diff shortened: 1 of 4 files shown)


mono/mini/jit.h
```diff
@@ -100,6 +100,16 @@ mono_jit_parse_options     (int argc, char * argv[]);
 
 MONO_API char*       mono_get_runtime_build_info    (void);
 
+MONO_API MONO_RT_EXTERNAL_ONLY void
+mono_set_use_llvm (mono_bool use_llvm);
+
+MONO_API MONO_RT_EXTERNAL_ONLY void
+mono_aot_register_module (gpointer *aot_info);
```
lambdageek: this is wrong - shouldn't use `gpointer` in a public header.  will fix.

lambdageek:

Issue #14957, opened 2019-06-10
[bockbuild] Non-git sources (like Nuget) don't have their cache invalidated properly


alexischr:
(Opening issue here as we disabled issues from https://github.com/mono/bockbuild)

This issue mostly shows up at https://github.com/mono/mono/blob/master/packaging/MacSDK/nuget.py#L7 , where changing the URL does not trigger a re-processing of the package as expected.

Bockbuild uses a "buildstring" (saved as `artifacts/packagename.buildstring`) to decide if it will re-build a package or not. The nuget.py URL update does indeed trigger "needing rebuilding" by comparing the current buildstring with the one saved on the file by the previous build. (Buildstring example: my `artifacts/NuGet-darwin-universal.buildstring` has a single line: `NuGet 5.0.2 <https://dist.nuget.org/win-x86-commandline/v5.0.2/nuget.exe> md5: c8461c00db2b16f5ce4671423768bf24`)

However by that time, the fetch code that downloads the sources has already ran, and having found a file with the same name under `cache/nuget.exe`, gives up on re-downloading it.

https://github.com/mono/bockbuild/blob/master/bockbuild/package.py#L134 is where fetching (processing of `sources`) happens, and https://github.com/mono/bockbuild/blob/master/bockbuild/package.py#L304 is the code path that `nuget.py` follows.

Note that we should not just delete everything in `cache` if the source needs rebuilding; caching in bockbuild also includes git repositories often just need to just do a fetch and a checkout.

lewing:
ono JIT compiler version 6.0.0.283 (2019-02/240aa5b900f Wed Jun  5 17:25:57 EDT 2019) shipped with the wrong version of nuget.exe because of this bug

lewing:
see https://github.com/mono/mono/pull/15121

PR #15121, opened 2019-06-13
[2019-02][MacSDK] Trigger redownload of Nuget 5.1.0

radical:
Some of the packages/bockbuild caches have somehow ended up with a Nuget 5.0.2 binary, even after the 5.1.0 bump. Trying to reset cache here

### DIFF
--- a/packaging/MacSDK/nuget.py
+++ b/packaging/MacSDK/nuget.py
@@ -23,4 +23,4 @@ class NuGetBinary (Package):
             output.write(
                 'exec {0}/bin/mono $MONO_OPTIONS {1} "$@"\n'.format(self.staged_prefix, target))
         os.chmod(launcher, 0o755)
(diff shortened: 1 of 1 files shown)


radical:
@monojenkins OS X Extended

radical:

radical:

lewing:
@monojenkins build failed

lewing:
@radical do we need to do this in master/2019-06 as well?

radical:
The package from `OSX Package` build still has the old nuget!

radical:
@alexischr How can we force this to rebuild? Or should this be merged and then the jenkins build is the one that matters?

alexischr:
The problem is because of https://github.com/mono/mono/issues/14957, so this patch will not work either. We will clear the cache manually for now, and backport the fix patch.

Issue #14486, opened 2019-05-15
Execution context on thread pool threads is not properly reset


filipnavara:
The CI on ARM often fails with the following errors. It seems to be caused by usage of `AsyncLocal` to propagate values to thread pool threads. The threads are then not properly reset after the queued job is complete and it retains the previous values.

```
xUnit.net Console Runner v2.4.1 (32-bit Desktop .NET 4.7.2, runtime: 4.0.30319.42000)
  Discovering: net_4_x_corlib_xunit-test
  Discovered:  net_4_x_corlib_xunit-test
  Starting:    net_4_x_corlib_xunit-test
    System.Threading.ThreadPools.Tests.ThreadPoolTests.QueueUserWorkItem_PreferLocal_NullValidForState [FAIL]
      System.TypeInitializationException : The type initializer for 'System.Threading.ThreadPools.Tests.ThreadPoolTests' threw an exception.
      ---- System.AggregateException : One or more errors occurred. (Assert.True() Failure
      Expected: True
      Actual:   False)
      -------- Assert.True() Failure
      Expected: True
      Actual:   False
      Stack Trace:
          at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
          at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /home/builder/jenkins/workspace/test-mono-pull-request-armhf/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395 
        ----- Inner Stack Trace -----
          at System.Threading.Tests.ThreadTestHelpers+<>c__DisplayClass3_0.<CreateGuardedThread>b__1 () [0x00022] in /home/builder/jenkins/workspace/test-mono-pull-request-armhf/external/corefx/src/Common/tests/System/Threading/ThreadTestHelpers.cs:44 
          at System.Threading.Tests.ThreadTestHelpers+<>c__DisplayClass3_0.<CreateGuardedThread>b__2 () [0x00015] in /home/builder/jenkins/workspace/test-mono-pull-request-armhf/external/corefx/src/Common/tests/System/Threading/ThreadTestHelpers.cs:49 
[...]

PR #15126, opened 2019-06-14
Fix race conditions in mono_lazy_initialize.

jaykrell:
There are two things here.
One is a no-brainer. Fast and required.
Changing `status` to `*lazy_init`.
Without that, initialization claims failure in a race, when there was actually success.

One is a little controversial.
A slow down (added barrier in fast path).
Could be done in a more complex barrier-free way.
Isn't always a problem.
Details and other recommendations within.

This is a superset of https://github.com/mono/mono/pull/15020.

The small no-brainer part fixes:
https://github.com/mono/mono/issues/14486

Thanks to @MaximLipnin for the trivial repro.

### DIFF
--- a/mono/utils/mono-lazy-init.h
+++ b/mono/utils/mono-lazy-init.h
@@ -43,7 +43,7 @@
  *  - not be called concurrently (either 2+ initialize or 2+ cleanup, either initialize and cleanup)
  */
 
-typedef gint32 mono_lazy_init_t;
+typedef volatile gint32 mono_lazy_init_t;
 
 enum {
 	MONO_LAZY_INIT_STATUS_NOT_INITIALIZED,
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

lewurm:
looks okay to me, but I would like @BrzVlad to have a look at it

lambdageek:
Is there any test case that we could extract from #14486?

Also great job @MaximLipnin and @jaykrell tracking this down!

lambdageek:
Once @BrzVlad looks it over, we should backport this a bunch - I think to `2019-02` and maybe more @marek-safar

jaykrell:
Test case, yes, I thought of that, and @MaximLipnin’s test is good. Ran quick and 100% / 0% pass rate.

jaykrell:
Might want to split it up.

jaykrell:
More credit to @MaximLipnin, he made it trivial.

marek-safar:
@monojenkins backport 2019-06

marek-safar:
@monojenkins backport 2019-02

marek-safar:
Backported to all active branches

jaykrell:
@monojenkins build failed

lewing:

jaykrell:
To be clear, the need for the barrier is small-sometimes-zero, and a slowdown on a fast path. With a possible alternate non-slowdown solution.

The need to replace `status` with `*lazy_init` is the more important part, and not a slowdown.

Personally I don't like even small-sometimes-zero bugs, but it could be argued.
Maybe I PR that other idea and the slowdown can be gone.
Maybe the slowdown itself nobody worries about.

jaykrell:
I thought I was pretty clear in the original description.
I am in favor of this asis, for the time being, and try later the barrier-less solution.
There are even more options though.

On-demand initialization has essentially two subtle boolean factors.
So four cases.
It is tricky to explain.

1. Some on-demand initialization is safe to run more than once, concurrently. Each run will produce the same value, take any. Just try not to run it every single time. Just the first few.

But some initialization can only be run no more than once.

2. Some on-demand initialization leads to a "data dependency". The easiest way to understand that is, the initialization mallocs a pointer and initializes it.

But some initialization just writes to globals, no pointers.

These two variables allow more or less wiggle room in how something that wraps on demand initialization works.

The previous version of lazy_init, clearly had in mind only initialize once.
The previous version of lazy_init, probably accidentally, had in mind the data dependency case.

One could argue, that lazy_init (or on_demand_init) actually accept two booleans.
Not clear if callers would understand them and set them correctly though.

In the case where data_dependency == TRUE, which is the case sometimes, the new barrier is not needed.

In the case where run_multiple_ok == true, the implementation can take a more efficient approach.
And be coop-friendly for free (by not blocking).

The "based on C++" approach is data_dependency = false, run_multiple_ok = false, that is, it does handle the most demanding case, nobody has to understand how to set the booleans, and would require coop-friendly adaption.

Furthermore though, with run_no_more_than_once = false

lewing:
@monojenkins build failed

jaykrell:
I understand the barrier-free solution now. At least w/o cleanup, which is some of the uses.
We can probably also move some of the initialization to not be on-demand.

I'd still take this PR for now, and later do a combination of:
  - switch to not on-demand initialization; for example try out `__thread` instead `error_key_init`. And `mono_w32file_init` => `mono_fdhandle_register` => `mono_lazy_initialize` seems pointlessly lazy, just initialize up front.

And `mono_runtime_init_checked` => `mono_network_init` =>  `mono_fdhandle_register` seems pointlessly lazy also.

and `mono_gc_init` => `mono_lazy_initialize` doesn't seem motivated.

Unless some of these were going to gain cleanup also, perhaps. But even not clearly that.

  - deciding the fast path isn't so fast, so barrier ok; for example in `process_create`

lewing:
@monojenkins build failed

PR #15130, opened 2019-06-14
[netcore] Fix CI build script

EgorBo:
Also fix `make run-sample`

### DIFF
--- a/scripts/ci/pipeline-netcore-runtime.yml
+++ b/scripts/ci/pipeline-netcore-runtime.yml
@@ -87,6 +87,11 @@ jobs:
     condition: ne(variables['poolname'], 'Hosted VS2017')
 
   - bash: |
+      make -C netcore nupkg
+    displayName: 'Build nupkg'
+    condition: ne(variables['poolname'], 'Hosted VS2017')
+
+  - bash: |
       make -C netcore update-tests-corefx
     displayName: 'Download CoreFX Tests'
     condition: ne(variables['poolname'], 'Hosted VS2017')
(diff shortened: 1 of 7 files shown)


scripts/ci/pipeline-netcore-runtime.yml
```diff
@@ -82,10 +82,16 @@ jobs:
 
   - bash: |
       ./autogen.sh --with-core=only
-      make -j 4
+      make -C mono -j4
```
marek-safar: This is redundant, make builds everything

EgorBo: @marek-safar I don't know why but `make` from the root folder sometimes hangs (happened for @steveisok too afair)

marek-safar: does it still happen? I thought I fixed by removing that recursive target from netcore

EgorBo: happened for me 15 minutes ago :(

EgorBo: hm.. `build.sh` used to run xunit tests via ./run-step script (which could handle timeouts) now the macOS lane is dead cause `make -C netcore run-tests-corefx` can't do it.

directhex:
that's new. hello bot. where'd you come from? :grimacing:

directhex:
oh, the slash-azp. interesting

scripts/ci/pipeline-netcore-runtime.yml
steveisok: make -j <any number> seems to hang all the time for me.  Just plain make does not.

marek-safar: could you debug that, I tried 10 times in a row with clean configure with no hang with tip of master

PR #15131, opened 2019-06-14
[metadata] Move problematic image out-param usage

alexanderkyte:
The second assignment to the "problematic" image boolean out-parameter was in the wrong place. 

I've moved it down into register_image. 

I also enabled it here on Windows. As I recall, this is particularly relevant on Windows.


alexanderkyte:
@monojenkins build failed

PR #15149, opened 2019-06-14
[System.Private.CoreLib] Extract RuntimeType for easier refactoring

marek-safar:


EgorBo:
It seems it broke a lot of tests

marek-safar:
@EgorBo ouch, that's strange it was a file copy

PR #15196, opened 2019-06-16
Fix compiler warnings.

jaykrell:
unused variable: save_sigint
 unused variable: mono_environ
 unprototyped function: mono_marshal_xdomain_copy_value

### DIFF
--- a/mono/metadata/remoting.c
+++ b/mono/metadata/remoting.c
@@ -2131,6 +2131,8 @@ leave:
 	return result;
 }
 
+#ifndef DISABLE_REMOTING
+
 /* mono_marshal_xdomain_copy_value
  * Makes a copy of "val" suitable for the current domain.
  */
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

mono/metadata/w32process-unix.c
```diff
@@ -97,6 +97,7 @@
 /* define LOGDEBUG(...) g_message(__VA_ARGS__)  */
 
 /* The process' environment strings */
+#if defined (HAVE_FORK) && defined (HAVE_EXECVE)
```
jaykrell: Alternative is to remove `static`, and give it a more unique name (it occurs elsewhere, but also `static`).

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/remoting.c
```diff
@@ -2131,6 +2131,8 @@ mono_marshal_xdomain_copy_value_handle (MonoObjectHandle val, MonoError *error)
 	return result;
 }
 
+#ifndef DISABLE_REMOTING
```
lateralusX: We should probably fix so that prototype and definition signatures of mono_marshal_xdomain_copy_value_handle are the same as well.

jaykrell: I don't follow. You mean put `static` on the definition to match declaration? Yeah that'd be ok.

jaykrell: Ok, I did that. Not sure it is what you meant.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15204, opened 2019-06-17
Add cooperative suspend aware threads for hybrid suspend.

lateralusX:
When running in hybrid suspend and having threads in blocking mode, GC Safe mode, threads will still be preemptive suspended. This is done to make sure compatibility is kept with embedders since current embedding API is not coop aware. If, however the external attached threads promise to follow the coop
rules for a thread in Safe mode, it could still be possible to have them running, making sure they won't be preempted when running STW. Unlocking this possibility could have large performance wins if embedding threads runs a lot of unrelated native code on threads attached to runtime (game engines as an example).

This commit adds support for a new pair of API's where an external thread can mark itself as cooperative suspend aware. Doing that guarantees the runtime that it will follow all rules as dictated by cooperative suspend when a thread is in safe mode. The change only goes one way since this is a lifetime commitment for a thread. Once’s set, thread will be kept running when in safe mode.

### DIFF
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -763,3 +763,26 @@ mono_threads_exit_no_safepoints_region (const char *func)
 	MONO_REQ_GC_UNSAFE_MODE;
 	mono_threads_transition_end_no_safepoints (mono_thread_info_current (), func);
 }
+
+void
+mono_thread_set_coop_aware (void)
+{
+	MONO_ENTER_GC_UNSAFE;
+	MonoThreadInfo *info = mono_thread_info_current_unchecked ();
+	if (info)
+		mono_atomic_store_i32 (&(info->coop_aware_thread), TRUE);
+	MONO_EXIT_GC_UNSAFE;
+}
+
+mono_bool
+mono_thread_get_coop_aware (void)
+{
+	mono_bool result = FALSE;
+	MONO_ENTER_GC_UNSAFE;
+	MonoThreadInfo *info = mono_thread_info_current_unchecked ();
+	if (info)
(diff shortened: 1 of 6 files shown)


lateralusX:
@monojenkins rebuild failed

lateralusX:
@monojenkins rebuild failed

lambdageek:
@monojenkins build failed

mono/utils/mono-threads-coop.c
```diff
@@ -763,3 +763,26 @@ mono_threads_exit_no_safepoints_region (const char *func)
...
 	mono_threads_transition_end_no_safepoints (mono_thread_info_current (), func);
 }
+
+void
+mono_thread_set_coop_aware (void)
+{
+	MONO_ENTER_GC_UNSAFE;
+	MonoThreadInfo *info = mono_thread_info_current_unchecked ();
+	if (info)
+		mono_atomic_store_i32 (&(info->coop_aware_thread), TRUE);
```
lambdageek: I think we need a comment right here that it is critical that we do this store while in GC Unsafe mode.

Otherwise, if we allow the flag to be flipped from GC Safe, there could be a race if we're using signals for suspending: The suspend initiator could send a signal, but then the thread would switch to be coop aware right before it's preemptively suspended, and then I believe the resume code would use the cooperative resume instead of the signal-based preemptive resume.

lateralusX: Added comment:

/* NOTE, this flag should only be changed while in unsafe mode. */
/* It will make sure we won't get an async preemptive suspend */
/* request against this thread while in the process of changing the flag */
/* affecting the threads suspend/resume behavior. */

steveisok:
@lambdageek I'm assuming this is good to merge?

PR #15210, opened 2019-06-17
Encode "&" symbols in script for xml reading

coolshooter:
This allows to properly read "&" symbols in a script (more specifically, in a url, which is inside a script "src" attribute).


steveisok:
@coolshooter, thanks for the PR, but https://github.com/mono/mono/pull/12619 should land in the relative short term.  I believe your issue will be fixed then.

Issue #15073, opened 2019-06-13
[netcore] Make System.Reflection.Tests.AssemblyNameTests.FullName Pass


steveisok:
The test currently fails with: 

```
System.Reflection.Tests.AssemblyNameTests.FullName(name: "name", expectedName: "name") [FAIL]
      Expected
      name, publickeytoken=null == name
      or
      name, publickeytoken=null == name, culture=neutral, publickeytoken=null
      Expected: True
      Actual:   False
      Stack Trace:
           at System.Reflection.Tests.AssemblyNameTests.FullName(String name, String expectedName)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    System.Reflection.Tests.AssemblyNameTests.FullName(name: "NAME", expectedName: "NAME") [FAIL]
      Expected
      name, publickeytoken=null == name
      or
      name, publickeytoken=null == name, culture=neutral, publickeytoken=null
      Expected: True
      Actual:   False
      Stack Trace:
           at System.Reflection.Tests.AssemblyNameTests.FullName(String name, String expectedName)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    System.Reflection.Tests.AssemblyNameTests.FullName(name: "name with spaces", expectedName: "name with spaces") [FAIL]
      Expected
      name with spaces, publickeytoken=null == name with spaces
      or
      name with spaces, publickeytoken=null == name with spaces, culture=neutral, publickeytoken=null
      Expected: True
      Actual:   False
      Stack Trace:
           at System.Reflection.Tests.AssemblyNameTests.FullName(String name, String expectedName)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    System.Reflection.Tests.AssemblyNameTests.FullName(name: "𐀀", expectedName: "𐀀") [FAIL]
      Expected
      𐀀, publickeytoken=null == 𐀀
      or
      𐀀, publickeytoken=null == 𐀀, culture=neutral, publickeytoken=null
      Expected: True
      Actual:   False
      Stack Trace:
           at System.Reflection.Tests.AssemblyNameTests.FullName(String name, String expectedName)
[...]

PR #15212, opened 2019-06-17
[netcore] Don't default to an empty array for a null publicKey

CoffeeFlux:
Fixes https://github.com/mono/mono/issues/15073

I'm not particularly fond of the test here since I don't see any benefit to there being multiple acceptable behaviors, but I've just matched Core 3. Fixed `System.Reflection.Tests.AssemblyNameTests.SetPublicKeyToken_GetPublicKeyToken` for free in the process!


CoffeeFlux:
@monojenkins build failed

PR #15215, opened 2019-06-17
[netcore] Fix RuntimeType.cs

EgorBo:
I suspect some regions were deleted accidentally during https://github.com/mono/mono/pull/15149 so now we have > 3k of failing tests in `System.Runtime.Tests`.
I've tried to merge `rttype.cs` with `RuntimeType.cs` myself using IDE  (because I don't think I could just find what is missing in these huge files).
What I did:
1) Removed all irrelevant #if-defs (IDE rendered them in gray color), only left one `#if FULL_AOT_RUNTIME` and `_DEBUG` symbols.
2) merged rttype.cs with netcore/RuntimeType.cs
3) removed unused members

Let's see what the CI thinks

### DIFF
--- a/netcore/System.Private.CoreLib/src/System/RuntimeType.cs
+++ b/netcore/System.Private.CoreLib/src/System/RuntimeType.cs
@@ -2,134 +2,2382 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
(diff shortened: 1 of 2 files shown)


EgorBo:
Looks like the CI is happy, the linux lane has 0 failing tests.

marek-safar:
@EgorBo yeah!

PR #15216, opened 2019-06-17
[System.Private.CoreLib] Align on CA naming for easier sharing

marek-safar:


marek-safar:
@monojenkins build failed

PR #15205, opened 2019-06-17
[arm] use target_mgreg_t instead of plain pointer type in sdb breakpoint

lewurm:
Fixes Debug mode for iOS 32bit.

Contributes to https://github.com/mono/mono/issues/15058 & https://github.com/mono/mono/issues/9621

Diff before (left old cross32, right new cross32 running on amd64 host _before_ this fix):
Screenshot 2019-06-17 at 14 09 51

/cc @mandel-macaque

PR #15222, opened 2019-06-18
[mini] fix OP_SEQ_POINT of other backends regarding target pointer size

lewurm:
See https://github.com/mono/mono/pull/15205#issuecomment-502925368

Theoretically fixes cross compilation on a 32bit host to a 64bit target.


jaykrell:
`TARGET_SIZEOF_VOID_P` agreed is better, because it is one token, for the hypothetical single configure
ation that can target multiple targets. Currently we use a lot of each and agreed, equivalent.

jaykrell:
There is also a subtelty, that the code is in general confused about, between registers and pointers.
You know, arm6432 and x32.
Registers can hypothetically be larger than pointers.
Does arm6432 work, and correctly discern those?

lewurm:
@monojenkins squash

lewurm:
@monojenkins build failed

alexischr:
Failures are unrelated

Issue #15058, opened 2019-06-13
[2019-06][iOS] Several tests on iOS fail on a 32b device


mandel-macaque:
After upgrading to mono 2019-06 the following tests fail on a 32b iOS device in the following configurations:

* AssemblyBuildTarget: SDK framework (release) (Sample [run log](https://gist.github.com/mandel-macaque/b57168cef60004b2477d4f3744cb6fe6))(Sample [build log](https://gist.github.com/mandel-macaque/af234a06d95d554aa1cd028e08039980)) (Sample [crash report](https://gist.github.com/mandel-macaque/07c0a376e5485344c1a89010d0591784)).
* Release (Sample [run log](https://gist.github.com/mandel-macaque/c88a30c4886e19ff654edc2add275eb8))(Sample [build log](https://gist.github.com/mandel-macaque/c3b2079faa769960a657d468e54b928f)) Sample [crash report](https://gist.github.com/mandel-macaque/9be6abee6a70f6340c79c030cc590b65)).

And timeout in the following ones:

* AssemblyBuildTarget: dylib (debug) (Sample [rung log](https://gist.github.com/mandel-macaque/656d12d061d11d17cf1d540dfc84f5ee)) (Sample [build log](https://gist.github.com/mandel-macaque/8dd02eac9b60768a357de1e1f80d4ff4)).
* AssemblyBuildTarget: dylib (debug, profiling) (Sample [run log](https://gist.github.com/mandel-macaque/c2041d0ce1e21224bfb743b7cb9471db)) (Sample [build log](https://gist.github.com/mandel-macaque/112316e3c065010b1eec48595d4756f1)).
* AssemblyBuildTarget: SDK framework (debug) (Sample[ run log](https://gist.github.com/mandel-macaque/bd129f517bdc7aa839e9f03e8fbf8bda)) (Sample [build log](https://gist.github.com/mandel-macaque/89cb4aacc40d5aa6363943059a87445f)).
* AssemblyBuildTarget: SDK framework (debug, profiling) (Sample [run log](https://gist.github.com/mandel-macaque/304e3d8b8e613b9623b81ee5bc962a28)) (Sample [build log](https://gist.github.com/mandel-macaque/714c9d3e2c109f62949f04671295e979)).
[...]

PR #15223, opened 2019-06-18
[arm] fix some more host/target pointer size issues

lewurm:
As suggested by Jay in https://github.com/mono/mono/issues/15058#issuecomment-501995400

I haven't seen fixing it any specific issue, but we should still backport it to `2019-06` in order to avoid hard-to-debug regressions.


jaykrell:
I'm not sure about these actually. They might only be for JIT scenarios. In which case it doesn't matter. But perhaps it is clearer this way.

/cc @vargaz

```
C:\s\mono2\mono\mini\mini-runtime.c(4305):		mono_install_imt_trampoline_builder (mono_arch_build_imt_trampoline);
```

lewurm:
@monojenkins build failed

lewurm:
@monojenkins backport 2019-06

lewurm:
@monojenkins squash

PR #15224, opened 2019-06-18
Remove dead Boehm code.

jaykrell:
Unity looked similarly dead.

The key points making it dead were:

create_allocator:
  g_assert_not_reached() covering everything but local variables

mono_gc_get_managed_allocator:
  always returning NULL, killing a bunch of code
  after it

mono_gc_is_critical_method:
   looking at data that was never filled in

This should have no impact, not even on Boehm.
Boehm should continue working just the same.

It is nice because it is the few/only uses of CEE_MONO_TLS.


mono/metadata/boehm-gc.c
```diff
@@ -959,371 +959,9 @@ mono_gc_get_restart_signal (void)
...
-	int index_var, bytes_var, my_fl_var, my_entry_var;
-	guint32 no_freelist_branch, not_small_enough_branch = 0;
-	guint32 size_overflow_branch = 0;
-	MonoMethodBuilder *mb;
-	MonoMethod *res;
-	MonoMethodSignature *csig;
-	const char *name = NULL;
-	WrapperInfo *info;
-
-	g_assert_not_reached ();
```
jaykrell: Here.

mono/metadata/boehm-gc.c
```diff
@@ -959,371 +959,9 @@ mono_gc_get_restart_signal (void)
...
-MonoMethod*
-mono_gc_get_managed_allocator (MonoClass *klass, gboolean for_box, gboolean known_instance_size)
-{
-	int atype;
-
-	/*
-	 * Tls implementation changed, we jump to tls native getters/setters.
-	 * Is boehm managed allocator ok with this ? Do we even care ?
-	 */
-	return NULL;
```
jaykrell: Here.

mono/metadata/boehm-gc.c
```diff
@@ -959,371 +959,9 @@ mono_gc_get_restart_signal (void)
...
- *   Return a managed allocator method corresponding to allocator type ATYPE.
- */
-MonoMethod*
-mono_gc_get_managed_allocator_by_type (int atype, ManagedAllocatorVariant variant)
-{
-	MonoMethod *res;
-	gboolean slowpath = variant != MANAGED_ALLOCATOR_REGULAR;
-	MonoMethod **cache = slowpath ? slowpath_alloc_method_cache : alloc_method_cache;
-
-	return NULL;
```
jaykrell: Here.

jaykrell:
@monojenkins build failed

BrzVlad:
@joncham Are you aware that this code is dead ? I am assuming it has a preaty heavy perf impact and it wouldn't be too hard to fix (assuming it is only related to tls access)

jaykrell:
Leave it for restoration?

jaykrell:
How about instead I `#if 0` it out?

jaykrell:
@monojenkins build failed

lambdageek:
Assuming @joncham doesn't want this code, I'd prefer to delete rather than `#if 0`

jaykrell:
Agreed, and let me trim at the ‘#if’, will clean up later.

mono/metadata/boehm-gc.c
```diff
@@ -959,371 +959,9 @@ mono_gc_get_restart_signal (void)
 }
 
 #if defined(USE_COMPILER_TLS) && defined(__linux__) && (defined(__i386__) || defined(__x86_64__))
```
jaykrell: Remove from ‘#if’ to/including ‘#else’ instead.

jaykrell: Now that I have bigger screen -- this isn't terrible. Just a tiny excess remains.

jaykrell:
Pinged Jon here and in slack, no answer.

PR #15227, opened 2019-06-18
Fix 593980

bmerlet:


bmerlet:
Forgot to put a description in. Sorry. This change fixes #593980. The problem is that if settings have not been read or written before they are saved, nothing is written to the user.config file (instead of the default value of the settings). The fix is to force serialization of the settings, even if they have not been accessed.

steveisok:
@bmerlet Thanks for the PR.  We currently have this https://github.com/mono/mono/pull/12619 PR that should be merged in the relative short term.  It updates System.Configuration to the reference source implementation and should fix your issue.  When it lands and you find otherwise, feel free to open another PR / log an issue.

bmerlet:
After looking at #12619 PR, I fail to see how it is related to my PR. The set of files changed by #12619 does not even come near the file affected by my fix. Anybody care to explain.

steveisok:
Because the user.config flow was touched and for me, your case works as I would expect.  

If you would like to double check me, feel free to test against the PR

bmerlet:
OK, thanks. I pulled the PR 12619 and it does indeed fix my issue.

PR #15241, opened 2019-06-19
If0 dead Boehm code.

jaykrell:
Alternative to https://github.com/mono/mono/pull/15224.

PR #15226, opened 2019-06-18
Change MONO_PATCH_INFO_GET_TLS_TRAMP to MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL.

jaykrell:
The goal being fewer patch types for slight simplicity and efficiency.
There are still plenty of patch types to reason over however.

It was structured like this before, presumably, either because:
1. MONO_PATCH_INFO_JIT_ICALL_ADDR_NOCALL was string-based instead of enum-based or
2. NOCALL variant did not exist.
3. Perhaps the "enum overlay" is considered inelegant but I think it is ok. It was only possible before by parsing integers out of the strings, which other icalls did do (but no longer). Now that it is free of string parsing, I think it is ok.

PR #15245, opened 2019-06-19
Simplify/reduce CEE_MONO_TLS due their being only one live producer.

jaykrell:
There is a dead producer in Boehm. (see https://github.com/mono/mono/pull/15241)
This combined with https://github.com/mono/mono/pull/15226
will leave no remaining uses of `mono_tls_get_tls_getter`.


jaykrell:
Probably `MINT_MONO_TLS` should be renamed, and if there were multiple keys here, the switch should be in transform, to produce different interpreter opcodes. Maybe I do that now.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

BrzVlad:
I think this optimizes dead code

jaykrell:
Rare or dead?
I checked that `MANAGED_ALLOCATION` is defined, at least sometimes.
But I don't claim it is a hot path.
It'll be nice imho to remove `mono_tls_get_tls_getter`.
If it is truly dead, then remove it?

BrzVlad:
yes, it's dead. Managed allocator is only used by jit

jaykrell:
Ah, I see.
So ok if I just assert(false)/unreachable around here, and remove some stuff?
I'll PR that separately. You can weigh in.

jaykrell:
Instead https://github.com/mono/mono/pull/15256

jaykrell:
It isn’t dead. Let’s take this instead.

jaykrell:
@monojenkins build failed

Issue #15247, opened 2019-06-19
ves_icall_TypeBuilder_create_runtime_class_raw USED 132 handles


jaykrell:
## Steps to Reproduce

1.  Submit a PR.

### Current Behavior

https://jenkins.mono-project.com/job/z4/41/console
03:24:25.859 ves_icall_TypeBuilder_create_runtime_class_raw USED 132 handles

### Expected Behavior

Not sure.
Preferably no "warnings".
This might or might not be a bug.

## On which platforms did you notice this

Windows amd64 interpreter

jaykrell:
Looks like `reflection_setup_internal_class` in loop creates handles.

jaykrell:
This is fixed in master now.

PR #15248, opened 2019-06-19
[Coop] Establish handle frame in reflection_setup_internal_class.

jaykrell:
It is called in a loop by ves_icall_TypeBuilder_create_runtime_class.
See https://github.com/mono/mono/issues/15247.

### DIFF
--- a/mono/metadata/sre.c
+++ b/mono/metadata/sre.c
@@ -3834,17 +3834,20 @@ remove_instantiations_of_and_ensure_contents (gpointer key,
 static gboolean
 reflection_setup_internal_class (MonoReflectionTypeBuilderHandle ref_tb, MonoError *error)
 {
+	HANDLE_FUNCTION_ENTER ();
+
 	MonoReflectionModuleBuilderHandle module_ref = MONO_HANDLE_NEW_GET (MonoReflectionModuleBuilder, ref_tb, module);
 	GHashTable *unparented_classes = MONO_HANDLE_GETVAL(module_ref, unparented_classes);
+	gboolean ret_val;
 
 	if (unparented_classes) {
-		return reflection_setup_internal_class_internal (ref_tb, error);
+		ret_val = reflection_setup_internal_class_internal (ref_tb, error);
 	} else {
 		// If we're not being called recursively
 		unparented_classes = g_hash_table_new (NULL, NULL);
 		MONO_HANDLE_SETVAL (module_ref, unparented_classes, GHashTable *, unparented_classes);
 
-		gboolean ret_val = reflection_setup_internal_class_internal (ref_tb, error);
+		ret_val = reflection_setup_internal_class_internal (ref_tb, error);
 		mono_error_assert_ok (error);
 
 		// Fix the relationship between the created classes and their parents
@@ -3853,9 +3856,9 @@ reflection_setup_internal_class (MonoReflectionTypeBuilderHandle ref_tb, MonoErr
 
 		g_hash_table_destroy (unparented_classes);
 		MONO_HANDLE_SETVAL (module_ref, unparented_classes, GHashTable *, NULL);
-
-		return ret_val;
 	}
+
+	HANDLE_FUNCTION_RETURN_VAL (ret_val);
 }
 
 MonoReflectionTypeHandle


jaykrell:
Might want to more aggressively put frames within for loops.
Possibly leveraging C++ constructor/destructor.
Instead of depending on what they call to know or guard against being in a loop, also penalizing callers not in a loop.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lateralusX:
Looks like reflection_setup_internal_class_internal is also setting up its own frame, only called by reflection_setup_internal_class. Maybe it is better to move the frame setup/teardown into the caller? If we keep it as is then maybe we shouldn't do the frame setup in reflection_setup_internal_class_internal as well?

jaykrell:
Oops, that makes me think this fix isn't effective and the handle allocation is elsewhere.

jaykrell:
Oops, no, this is it, it makes sense, because this is "in between". But agreed, the frame in `ves_icall_TypeBuilder_create_runtime_class` is maybe redundant then.

lateralusX:
Since reflection_setup_internal_class is called from other methods already having their own frame setup, maybe we should setup the frame at call site to not affect other callers already having their own local frame setup and are not calling this method in a loop?

jaykrell:
It is subjective, but we almost never setup frames in loops. Only when the code churn is too tedious/annoying (look for `HANDLE_LOOP_PREPARE`, and then obviously along with `SETUP_ICALL_FRAME` / `CLEAR_ICALL_FRAME`)
We almost always do it only for functions.
Anyway I added another commit to do that.
We should probably chose just one tactic here.

lateralusX:
Another thing/thought, we don't have a way to "free" a local handle? The only way currently is to pop the frame? So no way for reflection_setup_internal_class to free the unused handle on return (that is cheaper than setting up a local frame)?

jaykrell:
I favor the function-based approach. but it is subjective.
We can also split the function into framed and frameless, or in_loop and not_in_loop, to avoid penalizing the other callers. The names would be wonky, but efficiency would be preserved...well..I guess handling it in the loop is the same efficiency, just not a common pattern.

There is likely yet another solution where a handle to reuse is allocated in the caller.
That would be nice, to avoid the repeated setup/teardown.

Better yet is for managed to pass a handle, since it constructs them super efficienctly, much faster than native (managed-constructed handles are just pointers to stack, native-constructed handles involve thread locals and an occasional malloc).

jaykrell:
Handles are kinda like alloca -- freed upon function exit -- using the macro.
It is doable at arbitrary points, like loops, but is rarely done.
We should really have managed pass down temporary handles for native to reuse, for much better efficiency.

jaykrell:
Btw, I was feeling a little "FUDdy" about the loop construct, but if you look at it, it is pretty obviously correct, since it is just reusing the very pieces of the common function construct.
Perhaps the name could be better -- I'd prefer push/pop.
Not clear if they were meant to be public.
@lambdageek might weigh in, but to me it is a pretty close tie.
To be clear, the PR now contains two solutions, redundant, not incorrect. We can take one or the other or both, or maybe something else. (There are ways to make it more optimal but I doubt we will).
As well even the loop *could* be deemed ok since it scales with user-defined type structure, which don't tend to be huge, right? Thought I doubt it.

lambdageek:
I would err on the side of error-proofing and correctness - setup handle frames in both `reflection_setup_internal_class` and `reflection_setup_internal_class_internal`.

I'm not, in general, a fan of `HANDLE_LOOP_PREPARE` and company: using handles correctly in a loop body means you can't use `break`, `continue`,`goto` or `return` (including our various `MonoError` macros that hide a `goto` or a `return`).

jaykrell:
Splitting the loop body into a separate function -- isn't that *kinda* what `reflection_setup_internal_class_internal` already is, part of the loop, not the entire thing, and has other callers, and was missing a frame?

Doesn't seem a little excessive to do both?

Caller can also pass in a temporary handle for reuse, to avoid the repeated frame setup/teardown and handle allocation.

lambdageek:
> Doesn't seem a little excessive to do both?

No: it is simple and correct.

jaykrell:
Functions taking care of themselves seems the simpler ideal.
And then ignore loops.
Anyway I modified the comments to explain both frames.

jaykrell:
i.e. even adding a frame to `reflection_setup_internal_class_internal`.

jaykrell:
Sorry I misread https://github.com/mono/mono/pull/15248#issuecomment-503718557.
Agreed, will do.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15225, opened 2019-06-18
Increase timeout for interpreter fullaotmixed regression.

lateralusX:
On Windows (partly due to slow shell execution) this test take longer to
complete than current given timeout.

https://jenkins.mono-project.com/job/z3/

PR #15251, opened 2019-06-19
Interpreter PR builds run too many test suites.

lateralusX:
Fix makes sure we don't run full test suite on PR builds or when targeting arm architectures. Current implementation will always run full test suite on PR builds for Windows and Linux. Full test suites will be run on master CI builds for none arm architectures.

### DIFF
--- a/scripts/ci/run-test-interpreter.sh
+++ b/scripts/ci/run-test-interpreter.sh
@@ -14,7 +14,21 @@ ${TESTCMD} --label=System.Core --timeout=160m make -w -C mcs/class/System.Core r
 ${TESTCMD} --label=mcs-tests --timeout=160m make -w -C mcs/tests run-test V=1;
 ${TESTCMD} --label=Mono.Debugger.Soft --timeout=5m make -w -C mcs/class/Mono.Debugger.Soft run-test V=1
 
-if [[ ${CI_TAGS} != *'pull-request'* ]] || [[ ${CI_TAGS} != *'arm'* ]]; then
+run_full_test_suite=1
+
+# Don't run full interpreter test suite on arm architectures during PR builds.
+# NOTE, full test suite will still be run on none PR builds.
+if [[ ${CI_TAGS} == *'pull-request'* ]] && [[ ${CI_TAGS} == *'arm'* ]]; then
+	run_full_test_suite=0
+fi
+
+# Don't run full interpreter test suite on Windows during PR builds.
+# NOTE, full test suite will still be run on none PR builds.
+if [[ ${CI_TAGS} == *'pull-request'* ]] && [[ ${CI_TAGS} == *'win-'* ]]; then
+	run_full_test_suite=0
+fi
+
+if [[ "$run_full_test_suite" -eq "1" ]]; then
 
 	${TESTCMD} --label=corlib-xunit --timeout=60m make -w -C mcs/class/corlib run-xunit-test
 	${TESTCMD} --label=System.XML --timeout=5m make -w -C mcs/class/System.XML run-test V=1


jaykrell:
You don't want master to test all of arm? i.e. just:

`if [[ ${CI_TAGS} != *'pull-request'* ]]`

lateralusX:
Correct master shouldn't run full test suite on arm.

scripts/ci/run-test-interpreter.sh
```diff
@@ -14,7 +14,7 @@ ${TESTCMD} --label=System.Core --timeout=160m make -w -C mcs/class/System.Core r
 ${TESTCMD} --label=mcs-tests --timeout=160m make -w -C mcs/tests run-test V=1;
 ${TESTCMD} --label=Mono.Debugger.Soft --timeout=5m make -w -C mcs/class/Mono.Debugger.Soft run-test V=1
 
-if [[ ${CI_TAGS} != *'pull-request'* ]] || [[ ${CI_TAGS} != *'arm'* ]]; then
+if [[ ${CI_TAGS} != *'pull-request'* ]] && [[ ${CI_TAGS} != *'arm'* ]]; then
```
lewurm: @BrzVlad do you remember what the intend was back then? https://github.com/mono/mono/pull/9648/files#diff-5b6edc3ae02a8f8ef6e0cc12afca7eafR17

BrzVlad: I'm confused. The code looks ok to me. The idea was to not run full arm on PRs because it takes too long. (Comparing on master builds amd64 vs arm => 1:15 vs 3:30). Arm builds are already always lagging behind on PRs, this would make it worse.

And why would this PR make never running full arm tests ?

jaykrell: Johan is leaving PRs alone, still "fast", but also disabling on master.
I'd'a'thought you'd want to test *sometimes* on master, like daily or weekly. ?

lateralusX: OK, lets reset this to know what should be supported. Right now we run full test suite on PR's for all platforms except arm.

Doesn't look like we have any arm interpreter CI lanes on CI master.

So if the intent is to run full PR test suite on amd64 platforms then the above is not correct, but then this can't be done on Windows (takes to long time to run, similar to arm).

So if we still want to run full PR test suite on Linux and full test suite on master for all platforms, then this PR needs to be adjusted to exclude Windows on PR builds.

BrzVlad: https://jenkins.mono-project.com//job/test-mono-interpreter/

lateralusX: Ahh, sorry didn't find that job when I search. I will adjust the PR to exclude Windows on PR builds similar to arm.

BrzVlad: maybe it would look better and clearer to add in the scripts stuff like : if (pr and arm) exit; if (windows) exit; ... then the running of all test suites. And add comment to small explanation of why we don't run so it is no confusion

lateralusX: Agree, will make the condition clearer as well.

jaykrell: I curious why not test Windows sometimes, like daily or weekly, not PRs?

lateralusX: Full test suite will be run on daily master builds for Windows as well.

BrzVlad:
@lateralusX Do you know why interp on windows is so lazy compared to interp on linux ? Can't think of anything

lateralusX:
MSVC doesn't support computed goto's, so that could be one reason. Some other things that are slow is the execution environment as such (cygwin), but that is similar for other test suites, but it hits the fullaotmixed regression tests more (https://github.com/mono/mono/pull/15225). WIndows build bots are also quite weak compared to other machines so that makes all Windows builds slower compared to other desktop platforms.

Looking at a run the biggest suites taking the longest time is:

corlib, 10 minutes.
compile-runtime-tests, ~15 minutes.
mcs-tests, ~50 minutes.
corlib-xunit, ~20 minutes.

So right now a PR build will still take around ~2.5 hours.

Running the rest of the test suite will add and additional ~1h to that time.

So maybe the split on what's run on PR's should be done a little different, but I thought it would be easier to just start with the same set we used on other platforms excluding tests from the suite. Maybe we should exclude the mcs-tests suite even for PR builds to bring down time even more.

I believe we should at some point look at the performance as well, at least for some of the tests that seems to have a different relationship between Windows and Linux compared to other test suites.

marek-safar:
For Windows we should just run them on Helix in parallel as we do for other profiles that will cut the time to the slowest test suite

lateralusX:
@monojenkins rebuild failed

PR #15256, opened 2019-06-19
Remove unused CEE_MONO_TLS from interpreter.

jaykrell:
This contributes to removing `mono_tls_get_tls_getter` (along with https://github.com/mono/mono/pull/15226).

Alternative to https://github.com/mono/mono/pull/15245.


jaykrell:
Oops will fix

BrzVlad:
let's just remove the code. It doesn't serve any purpose

jaykrell:
@monojenkins build failed

jaykrell:
Looks like not dead. Take other PR?
Or I can figure out why it is used 
and try to remove.

BrzVlad:
This code should not be reached in practice. This happens in this test because we consider one class jit only and calling the managed allocator is seen as call to external code, which is interpreted. In real scenarios I think this might only happen if you have interp + aot assemblies and you don't have corlib full aot-ed. (maybe @lewurm can think of various configurations)

Compiled code should always use compiled managed allocator, so it either should be jitted or, if not possible, should be included in the --aot=interp image. As I looked into the code right now, it looks like the managed allocators are included into the interp image, so any compiled code should either use this or jit it, and never interpret it since it is perf sensitive code.

jaykrell:
Can we take the other PR for now?
"Later" (maybe never) try to remove it?

PR #15269, opened 2019-06-19
Write to the same union member that is later read.

jaykrell:
It should not matter in this case, but it is the usual practise,
and not strictly valid otherwise.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15270, opened 2019-06-19
Zero threadpool updates before broadcast.

jaykrell:
This *seems* better, not sure it fixes anything.
It *seems* important, if the queue actually fills up.
Would the producers hang, or something else wake them?

### DIFF
--- a/mono/metadata/threadpool-io.c
+++ b/mono/metadata/threadpool-io.c
@@ -440,7 +440,7 @@ selector_thread (gpointer data)
 
 		if (threadpool_io->updates_size > 0) {
 			threadpool_io->updates_size = 0;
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
man page: "The pthread_cond_broadcast() function is used whenever the shared-variable state has been changed in a way that more than one thread can proceed with its task. "

and that isn't being done here, if the queue was full, it doesn't get emptied until after the broadcast.

But I think you are right, that broadcast does not release/relock the lock, so it doesn't matter, as long as the change is made before the lock is released.

I believe this code violates common pthread patterns though as a result.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
It contradicts every example/documentation I checked.

PR #9568, opened 2018-07-14
Update Windows requirement from Vista (6.0) to 7 (6.1).

jaykrell:
Still to see what _WIN32_IE should be, maybe omit it.

Part of #9102

PR #15279, opened 2019-06-20
Add basic implementation of Environment.FailFast

### LINKED ISSUE #14827: [netcore]Implement Environment.FailFast
See the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.environment.failfast?view=netcore-3.0) for full description.

Couple of call outs:
1.) The process should write out a crash dump before terminating.  It would be nice if we could use our crash reporter as I think it does that plus a bunch of other useful things.
2.) On windows, we should write to the event log.  I don't know what the behavior is on other platforms.

filipnavara:
Fixes #14827.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -7410,7 +7410,7 @@ ves_icall_System_Environment_FailFast (MonoStringHandle message, MonoExceptionHa
 	// information in the error dump that CoreCLR would. On Windows 7+ we should call
 	// RaiseFailFastException directly instead of relying on the C runtime doing it
 	// for us and pass it as much information as possible. On Windows 8+ we can also
-	// use the __failfast intrinsic.
+	// use the __fastfail intrinsic.
 	abort ();
 }
 
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -7409,7 +7409,8 @@ ves_icall_System_Environment_FailFast (MonoStringHandle message, MonoExceptionHa
 	// NOTE: While this does trigger WER on Windows it doesn't quite provide all the
 	// information in the error dump that CoreCLR would. On Windows 7+ we should call
 	// RaiseFailFastException directly instead of relying on the C runtime doing it
-	// for us and pass it as much information as possible.
+	// for us and pass it as much information as possible. On Windows 8+ we can also
+	// use the __failfast intrinsic.
 	abort ();
 }
 
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -7391,6 +7391,28 @@ ves_icall_System_Environment_Exit (int result)
 	exit (result);
 }
 
+void
+ves_icall_System_Environment_FailFast (MonoStringHandle message, MonoExceptionHandle exception, MonoStringHandle errorSource, MonoError *error)
+{
+	if (MONO_HANDLE_IS_NULL (message)) {
+		g_warning ("CLR: Managed code called FailFast without specifying a reason.");
+	} else {
+		char *msg = mono_string_handle_to_utf8 (message, error);
+		g_warning ("CLR: Managed code called FailFast, saying \"%s\"", msg);
+		g_free (msg);
+	}
+
+	if (!MONO_HANDLE_IS_NULL (exception)) {
+		mono_print_unhandled_exception_internal ((MonoObject *) MONO_HANDLE_RAW (exception));
+	}
+
+	// NOTE: While this does trigger WER on Windows it doesn't quite provide all the
+	// information in the error dump that CoreCLR would. On Windows 7+ we should call
+	// RaiseFailFastException directly instead of relying on the C runtime doing it
+	// for us and pass it as much information as possible.
+	abort ();
+}
+
 MonoStringHandle
 ves_icall_System_Environment_GetGacPath (MonoError *error)
 {
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -175,16 +175,6 @@
 # https://github.com/mono/mono/blob/18f0bf055eb2bfa8981062bdad3af2fb4bbc9be4/netcore/System.Private.CoreLib/shared/System/String.cs#L811-L821
 -nomethod System.Tests.StringTests.NormalizationTest
 
-# Environment.FailFast is not implemented
--nomethod System.Buffers.Text.Tests.RealFormatterTests.TestFormatterDouble_C20
--nomethod System.Buffers.Text.Tests.RealFormatterTests.TestFormatterSingle_C20
--nomethod System.Buffers.Text.Tests.RealFormatterTests.TestFormatterSingle_N20
--nomethod System.Buffers.Text.Tests.RealFormatterTests.TestFormatterSingle_P20
--nomethod System.Buffers.Text.Tests.RealFormatterTests.TestFormatterDouble_N20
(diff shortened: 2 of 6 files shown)


mono/metadata/icall.c
```diff
@@ -7391,6 +7391,28 @@ ves_icall_System_Environment_Exit (int result)
...
 
+void
+ves_icall_System_Environment_FailFast (MonoStringHandle message, MonoExceptionHandle exception, MonoStringHandle errorSource, MonoError *error)
+{
+	if (MONO_HANDLE_IS_NULL (message)) {
+		g_warning ("CLR: Managed code called FailFast without specifying a reason.");
+	} else {
+		char *msg = mono_string_handle_to_utf8 (message, error);
+		g_warning ("CLR: Managed code called FailFast, saying \"%s\"", msg);
+		g_free (msg);
```
jaykrell: On Windows please consider `%ls` and passing through the Unicode.
i.e. no alloc, copy, free.
For that matter, if `sizeof(wchar_t) == sizef(gunichar2)`, maybe something.

jaykrell: Also please consider a small, maybe 64 local buffer, init with 0 and copy the first 64 or length chars from message.
i.e. if a dump is made, and includes stack, but doesn't follow pointers -- I don't know.

netcore/System.Private.CoreLib/src/System/Environment.cs
```diff
@@ -54,18 +54,16 @@ partial class Environment
...
-			throw new NotImplementedException ();
+			FailFast (message, exception, null);
 		}
 
-		public static void FailFast (string message, Exception exception, string errorMessage)
-		{
-			throw new NotImplementedException ();
-		}
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		public extern static void FailFast (string message, Exception exception, string errorSource);
```
jaykrell: Does C# have default parameters? Can you say just one function:
```
public extern static void FailFast (string message, Exception exception = null, string errorSource = null);
```

mono/metadata/icall.c
filipnavara: Agreed on both points. I just wanted to have something working first, pass the CI and then improve upon that.

netcore/System.Private.CoreLib/src/System/Environment.cs
filipnavara: This is public API, the shape cannot be changed.

mono/metadata/icall.c
```diff
@@ -7391,6 +7391,28 @@ ves_icall_System_Environment_Exit (int result)
...
+
+	if (!MONO_HANDLE_IS_NULL (exception)) {
+		mono_print_unhandled_exception_internal ((MonoObject *) MONO_HANDLE_RAW (exception));
+	}
+
+	// NOTE: While this does trigger WER on Windows it doesn't quite provide all the
+	// information in the error dump that CoreCLR would. On Windows 7+ we should call
+	// RaiseFailFastException directly instead of relying on the C runtime doing it
+	// for us and pass it as much information as possible.
+	abort ();
```
jaykrell: Windows 7 is the minimum requirement for a while now. Feel free to depend on it.
Notice that importing code from CoreFX was making it so.
And Vista is reall quite old now.

mono/metadata/icall.c
jaykrell: Looks good despite all my nitpicking.

mono/metadata/icall.c
jaykrell: If we are going to mention things, mention Windows 8 __failfast. Maybe another year or two down the line that'll be it.

filipnavara: I knew it was minimum requirement for .NET Core but I was unsure about Mono. I wrote the relevant part in managed code but I need to translate it back to native code and test the dumps.

jaykrell: Almost a year ago.
https://github.com/mono/mono/pull/9568

netcore/CoreFX.issues.rsp
```diff
@@ -175,16 +175,6 @@
 # https://github.com/mono/mono/blob/18f0bf055eb2bfa8981062bdad3af2fb4bbc9be4/netcore/System.Private.CoreLib/shared/System/String.cs#L811-L821
 -nomethod System.Tests.StringTests.NormalizationTest
 
-# Environment.FailFast is not implemented
```
EgorBo: These tests are unrelated actually - they call FailFast because of Debug.Assert (in debug mode)
Something wrong with fp operations inside number formatting

filipnavara: Interesting. I am pretty sure they passed when I removed the exclusions, perhaps for unrelated reasons.

filipnavara:
```
  Discovering: System.Memory.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  System.Memory.Tests (found 2295 of 2347 test cases)
  Starting:    System.Memory.Tests (parallel test collections = on, max threads = 4)
  Finished:    System.Memory.Tests
=== TEST EXECUTION SUMMARY ===
   System.Memory.Tests  Total: 45383, Errors: 0, Failed: 0, Skipped: 0, Time: 26.037s
```

EgorBo: Did you run them in debug mode? (`make debug-bcl && MONO_ENV...`)

filipnavara: Nope... good point.

filipnavara: I can reproduce it now. I'll put the exclusions back and clarify the comment. Thanks!

mono/metadata/icall.c
```diff
@@ -7391,6 +7391,29 @@ ves_icall_System_Environment_Exit (int result)
...
+
+	if (!MONO_HANDLE_IS_NULL (exception)) {
+		mono_print_unhandled_exception_internal ((MonoObject *) MONO_HANDLE_RAW (exception));
+	}
+
+	// NOTE: While this does trigger WER on Windows it doesn't quite provide all the
+	// information in the error dump that CoreCLR would. On Windows 7+ we should call
+	// RaiseFailFastException directly instead of relying on the C runtime doing it
+	// for us and pass it as much information as possible. On Windows 8+ we can also
+	// use the __failfast intrinsic.
```
CoffeeFlux: I believe it's `__fastfail`, because consistent naming is overrated. Also, will `__fastfail` still trigger WER?

filipnavara: Damn, will fix. It is documented to trigger WER.

mono/metadata/icall.c
NattyNarwhal: Shouldn't this use `syslog` to emulate Event Log semantics on WIndows?

filipnavara: This is not the part that is supposed to be written to event log. The event log should be written roughly at the `abort ();` call. In fact it happens on Windows that way and you get link to the WER minidump if enabled.

Printing this message is largely supposed to offer some indication either through attached native debugger, or through system log if no debugger is attached. On CoreCLR this goes through `OutputDebugString` on Windows. `syslog` may or may not be a good substitute, not sure about it.

`g_warning` is not perfect but it gets the job done here (both developer and user can see the message) and it has the added bonus of being wired properly on Android builds.

lewing:
@monojenkins build failed

PR #15293, opened 2019-06-20
[wasm] Add some missing directories to the clean target

lewing:
Add the driver conf targets to clean


lewing:

PR #15296, opened 2019-06-20
[sdks] Avoid building everything on make all.

vargaz:

### DIFF
--- a/sdks/builds/Makefile
+++ b/sdks/builds/Makefile
@@ -1,8 +1,6 @@
 TOP=$(realpath $(CURDIR)/../..)
 -include $(TOP)/sdks/Make.config
 
-all: package
-
 MAKEFLAGS += --no-builtin-rules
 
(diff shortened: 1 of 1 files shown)


vargaz:
@monojenkins build failed

sdks/builds/Makefile
```diff
@@ -1,8 +1,6 @@
 TOP=$(realpath $(CURDIR)/../..)
 -include $(TOP)/sdks/Make.config
 
-all: package
```
lambdageek: We should have some other phony target as the first thing in the makefile so that `make` without any arguments prints a message and exit.

Right now I think it'll run the `configure-mono` target (which just runs autogen).

vargaz:
@monojenkins build failed

lewing:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #15305, opened 2019-06-21
[netcore] fix windows build and build-test-corefx command 

EgorBo:
Fixes `build-test-corefx` command (cc @MaximLipnin)
Also don't know why but in order to use the runtime we download on windows I have to chmod it.

### DIFF
--- a/netcore/Makefile
+++ b/netcore/Makefile
@@ -164,10 +164,11 @@ run-tests-corefx-%: prepare update-tests-corefx
 		$(XUNIT_FLAGS) @../../../../CoreFX.issues.rsp \
 		$(FIXTURE) || true
 
-# build a test assembly in COREFX_ROOT and copy it to corefx/tests/extracted
+# build a test assembly in COREFX_ROOT (path to a fully built corefx repo) and copy it to corefx/tests/extracted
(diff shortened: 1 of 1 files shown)


netcore/Makefile
```diff
@@ -164,9 +166,8 @@ run-tests-corefx-%: prepare update-tests-corefx
 
 # build a test assembly in COREFX_ROOT and copy it to corefx/tests/extracted
 # e.g. `make build-test-corefx-System.Runtime COREFX_ROOT=/prj/corefx-master`
-build-test-corefx-%: check-env
-	cd $(COREFX_ROOT)/src/$*/tests && $(DOTNET) build -o "$(CURDIR)/corefx/tests/extracted/tmp"
-	cp $(CURDIR)/corefx/tests/extracted/tmp/$*.Tests.{dll,pdb} $(CURDIR)/corefx/tests/extracted/$*.Tests/
-	rm -rf $(CURDIR)/corefx/tests/extracted/tmp
+build-test-corefx-%:
+	cd $(COREFX_ROOT)/src/$*/tests && $(DOTNET) msbuild /p:OutputPath=tmp
```
MaximLipnin: Will it trigger the whole CoreFX build in case of the first pull? If no - maybe does it make sense to add short comment that CoreFX full build is required?

PR #15356, opened 2019-06-21
[mini] Support tracing pointer return values.

zfigura:
Greetings; this is just a simple debugging fix, partially to improve a +wrapper trace I encountered, and partly to ensure I'm acquainted to the Mono patch submission process.


marek-safar:
approve

PR #15363, opened 2019-06-22
[metadata] Respect unmanaged calling conventions in function signatures.

zfigura:
According to ECMA-335 § II.15.3, calling conventions other than "default" and
"vararg" describe unmanaged methods. Accordingly they should be treated as
pinvokes so that that their calling conventions are respected.

### DIFF
--- a/mono/metadata/metadata.c
+++ b/mono/metadata/metadata.c
@@ -2251,6 +2251,19 @@ mono_metadata_parse_method_signature_full (MonoImage *m, MonoGenericContainer *c
 	method->call_convention = call_convention;
 	method->generic_param_count = gen_param_count;
 
+	switch (method->call_convention) {
+	case MONO_CALL_DEFAULT:
+	case MONO_CALL_VARARG:
+		method->pinvoke = 0;
+		break;
+	case MONO_CALL_C:
+	case MONO_CALL_STDCALL:
+	case MONO_CALL_THISCALL:
+	case MONO_CALL_FASTCALL:
+		method->pinvoke = 1;
+		break;
+	}
+
 	if (call_convention != 0xa) {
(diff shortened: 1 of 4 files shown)


lambdageek:
Thanks for the PR @zfigura !

@monojenkins build

lambdageek:
@monojenkins approve

vargaz:
This has a chance of regressing things.

zfigura:
> I'm a little bit concerned that for a method definition we'd set `pinvoke` twice - once when we parse the signature here, and again when we notice `METHOD_ATTRIBUTE_PINVOKE_IMPL` in `mono_method_signature_checked`, but that seems harmless.

Yep. It could certainly be set in mono_metadata_parse_signature_checked() instead, if that would make more sense. I'll have to resend anyway, since the build is failing on MSVC, so please let me know if that would be preferred.

zfigura:
I've sent a new patch, which should fix the build with MSVC. Please let me know if any other changes are needed.

CoffeeFlux:
@monojenkins build failed

zfigura:
Hello, just checking, is there anything I still need to do to help get this reviewed or committed?

lambdageek:
@zfigura Sorry about the delay. Nothing to do on your side.

I think this isn't going to be too risky.  We see the method signatures in two places: when there's an indirect call and the signature is written on an instruction; or when we're loading method tables and we need to do a direct call.

If it's a sig on an indirect call, this is only changing the behavior where the signature specifies one of the unmanaged calling conventions (otherwise the pinvoke bit is zero initialized, as before).

If it's a direct call, I believe `mono_marshal_get_native_wrapper` won't look at this bit anyway and will instead look at the calling convention attributes directly.  So no change in behavior there, I think.

So I think this is good to go.

zfigura:
Thanks! No worry about the delay, just wanted to make sure I wasn't missing something :-)

PR #15374, opened 2019-06-24
Increase max size of localloc and localloc_imm on amd64.

lateralusX:
We hit the following assert when doing full AOT of some assemblies:

wrong maximal instruction length of instruction localloc (expected 96, got 99)
Assertion: should not be reached at mono\mono\mini\mini-amd64.c:6801

Turns out that that size could be hit when a couple of things happens as part of lowering OP_LOCALLOC and OP_LOCALLOC_IMM:

sreg1 is not RAX, RCX or RDI
dreg is not RAX, RCX or RDI
cfg->param_area is bigger than 127

Fix is to increase max size of the instruction, from 96 -> 120 (same as x86).

### DIFF
--- a/mono/mini/cpu-amd64.md
+++ b/mono/mini/cpu-amd64.md
@@ -140,7 +140,7 @@ cgt: dest:c len:8
 cgt_un: dest:c len:8
 clt: dest:c len:8
 clt_un: dest:c len:8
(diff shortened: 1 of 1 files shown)


lateralusX:
@monojenkins rebuild failed

PR #15373, opened 2019-06-24
[netcore] Add xunit summary script

EgorBo:
```
make xunit-summary
```
it prints some compact xunit report - a similar python script is used by coreclr (or corefx):
```
=========================================================================================================
Common.Tests.dll                                              Total:1144    Failed:0       Time:9.2 sec
CoreFx.Private.TestUtilities.Tests.dll                        Total:87      Failed:0       Time:0.2 sec
Invariant.Tests.dll                                           Total:296     Failed:0       Time:0.4 sec
Microsoft.Bcl.AsyncInterfaces.Tests.dll                       Total:10      Failed:0       Time:0.2 sec
Microsoft.CSharp.Tests.dll                                    Total:4796    Failed:0       Time:2.6 sec
Microsoft.VisualBasic.Core.Tests.dll                          Total:21748   Failed:0       Time:3.2 sec
Microsoft.Win32.Primitives.Tests.dll                          Total:1       Failed:0       Time:0.2 sec
Microsoft.XmlSerializer.Generator.Tests.dll                   Total:83      Failed:0       Time:0.7 sec
System.AppContext.Tests.dll                                   Total:15      Failed:0       Time:0.1 sec
System.Buffers.Tests.dll                                      Total:78      Failed:0       Time:1.9 sec
....
System.Xml.XmlSchema.XmlSchemaValidatorApi.Tests.dll          Total:362     Failed:0       Time:0.6 sec
System.Xml.XmlSchemaSet.Tests.dll                             Total:482     Failed:0       Time:0.9 sec
System.Xml.XmlSerializer.ReflectionOnly.Tests.dll             Total:228     Failed:0       Time:1.7 sec
System.Xml.XmlSerializer.Tests.dll                            Total:230     Failed:0       Time:1.8 sec
System.Xml.Xsl.XslCompiledTransformApi.Tests.dll              Total:1621    Failed:0       Time:7.8 sec
System.Xml.Xsl.XslTransformApi.Tests.dll                      Total:1593    Failed:0       Time:4.7 sec
XsltCompiler.Tests.dll                                        Total:1       Failed:0       Time:0.1 sec
=========================================================================================================

Total test suits:     205
Total tests run:      387451
Total tests passed:   384547
Total tests failed:   108
Total tests skipped:  2796
Total duration:       15 min
```

PR #15380, opened 2019-06-24
[netcore] Fix build.sh

CoffeeFlux:
It appears to have accidentally been changed in #15215, and the script no longer works correctly as a result, so revert

### DIFF
--- a/netcore/build.sh
+++ b/netcore/build.sh
@@ -87,7 +87,10 @@ CPU_COUNT=$(getconf _NPROCESSORS_ONLN || echo 4)
 
 # run .././autogen.sh only once or if "--rebuild" argument is provided
 if [[ "$force_rebuild" == "true" || ! -f .configured ]]; then
-  cd .. && ./autogen.sh --with-core=only
+  cd ..
+  ./autogen.sh --with-core=only
+  make -j$CPU_COUNT
+  cd netcore
   touch .configured
 fi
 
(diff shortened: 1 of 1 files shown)


EgorBo:
@CoffeeFlux it's fixed in https://github.com/mono/mono/pull/15373 🙂

CoffeeFlux:
I'd personally rather have this fix be explicit and merged quickly than squashed in with an only tangentially related PR - mixing in unrelated build system changes is how this happened in the first place. Maybe merge this and then make the changes in a separate PR? I have questions/feedback related to them that I'd rather not hold your xunit PR up for, notably:
1. Is there a reason we're moving the `.configured` file up a level? If it's not being used outside of netcore scripts it seems most appropriate to leave it there.
2. If we are moving it up a level, the `.gitignore` and `/netcore/Makefile` need to be updated accordingly.

EgorBo:
If you want to merge this you need to fix the gitignore, I have no preferences on where it should be (initially it was inside netcore but was moved to the root by Marek's clean up PR afair).

CoffeeFlux:
If you mean https://github.com/mono/mono/commit/9010d2f290d28dc97ac276e0c0d32a2932b7550f#diff-38868e38d9f6469767765e84ab10b1e4, it doesn't look like the `.configured` file got moved. I initially noticed this issue because the gitignore doesn't currently work. I thiiiink I can just merge this.

EgorBo:
@CoffeeFlux oh, indeed, but please remove `make -j` from there - it builds runtime and mscorlib separately

CoffeeFlux:
Ooh I see. I've imitated what you did in #15373 then but left the .configured file in the netcore directory. If you're good with this @EgorBo then I'll merge.

CoffeeFlux:
@monojenkins build failed

lewing:

PR #15384, opened 2019-06-24
[aotprof-tool] Initial import

radekdoulik:
Imported https://github.com/radekdoulik/aotprofile-tool and updated it
to use Mono.Profiler.Log assembly


radekdoulik:
@monojenkins build failed

man/aotprof-tool.1
```diff
@@ -0,0 +1,91 @@
+.TH aotprof-tool 1
```
vargaz: Should a tool's name include the -tool suffix ?

lewing:
@monojenkins build failed

man/aotprof-tool.1
marek-safar: It should not, @radekdoulik could come up with a better name?

radekdoulik:
Bunch of the tools in mcs/tools have util suffix, so *aotprofutil* or *aprofutil* might work.

Other possible names:
*aprofcruncher*
*aprofblender*
*aprofninja*
*aprofinspector*

steveisok:
I'm awful w/ names.  I like aprofinspector

marek-safar:
I like the aprof prefix, perhaps simple `aprofutil` would work

radekdoulik:
OK, let use the short version then.

steveisok:
@monojenkins build failed

akoeplinger:
@monojenkins commit csproj

PR #15357, opened 2019-06-22
[wasm][debugger] Adding support to debug local vars in async methods

thaystg:
- Improving performance calling only once mono_wasm_get_var_info.
- Adding support to debug local vars in async methods. They are special in the way that local variables can be lifted to generated class fields.

Fixes #14608

PR #15385, opened 2019-06-24
[debugger][wasm] Local vars

thaystg:
Fixing side effect of #15357.

### DIFF
--- a/sdks/wasm/library_mono.js
+++ b/sdks/wasm/library_mono.js
@@ -41,14 +41,12 @@ var MonoSupportLib = {
 				this.mono_wasm_get_var_info = Module.cwrap ("mono_wasm_get_var_info", 'void', [ 'number', 'number', 'number']);
 
 			this.var_info = [];
-			var typedArray = new Int32Array(var_list.length);
-			for (let i=0; i<var_list.length; i++) {
-				typedArray[i] = var_list[i]
-			}
-			var numBytes = typedArray.length * typedArray.BYTES_PER_ELEMENT;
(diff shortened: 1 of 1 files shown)


lewing:
@monojenkins build failed

thaystg:
@monojenkins build failed

lewing:
@monojenkins build failed

kg:
I don't understand what this commit is doing. The old typed array is still there but there's an additional allocation and some copies now

thaystg:
> I don't understand what this commit is doing. The old typed array is still there but there's an additional allocation and some copies now

As I could read in a lot of places this is the right way to pass parameters between Javascript and C functions.
Here is an example:
https://github.com/Planeshifter/emscripten-examples/tree/master/01_PassingArrays
Another:
https://becominghuman.ai/passing-and-returning-webassembly-array-parameters-a0f572c65d97

Please, if you know any other way to do this, please let me know...

kg:
From how I understand it, can you remove the initial scratch typed array and instead just malloc a block of space in the wasm heap for your data, then copy the data directly there? the bits where you fill typedArray[i] from your source would instead write directly into the wasm heap. Then there's no temporary typed array and no need to copy from your temp array into the wasm heap.

thaystg:
> From how I understand it, can you remove the initial scratch typed array and instead just malloc a block of space in the wasm heap for your data, then copy the data directly there? the bits where you fill typedArray[i] from your source would instead write directly into the wasm heap. Then there's no temporary typed array and no need to copy from your temp array into the wasm heap.

Done...

thaystg:
@monojenkins build failed

lewing:

Issue #15306, opened 2019-06-21
RemoteInvoke()-based tests and System.Diagnostics.Process.Tests crash on Windows


EgorBo:
All corefx tests with RemoteTestExecutor inside crash runtime on Windows. Somewhere inside Process.Start (see the following stacktrace). This blocks us on running tests on CI for Windows.

## Steps to Reproduce

1. `make run-tests-corefx-System.Runtime.Tests` or better `make run-tests-corefx-System.Diagnostics.Process.Tests`

### Current Behavior

Segmenation fault:
```
System.Diagnostics.Tests.ProcessStartInfoTests.TestWorkingDirectoryPropertyInChildProcess [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object
      Stack Trace:
        C:\prj\mono\mono-e\netcore\System.Private.CoreLib\shared\System\Runtime\InteropServices\SafeHandle.cs(92,0): at System.Runtime.InteropServices.SafeHandle.Dispose()
        /_/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs(622,0): at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
        /_/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Win32.cs(26,0): at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
        /_/src/System.Diagnostics.Process/src/System/Diagnostics/Process.cs(1245,0): at System.Diagnostics.Process.Start()
        /_/src/System.Diagnostics.Process/src/System/Diagnostics/Process.cs(1288,0): at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
           at Microsoft.DotNet.RemoteExecutor.RemoteExecutor.Invoke(MethodInfo method, String[] args, RemoteInvokeOptions options, Boolean pasteArguments)
           at Microsoft.DotNet.RemoteExecutor.RemoteExecutor.Invoke(Func`2 method, String arg, RemoteInvokeOptions options)
        /_/src/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs(367,0): at System.Diagnostics.Tests.ProcessStartInfoTests.TestWorkingDirectoryPropertyInChildProcess()
[...]

PR #15392, opened 2019-06-25
Fix marshalling bugs for StringBuilder parameters

### LINKED ISSUE #15366: StringBuilder marshalling bug
## Steps to Reproduce

1. Compile:
```csharp
using System;
using System.Text;

class Program
{
    [DllImport("user32.dll", CharSet = CharSet.Unicode)]
    static extern uint CharLowerBuffW([In, Out] StringBuilder lpsz, uint cchLength);

    static void Main()
    {
        var sb = new StringBuilder().AppendLine("HELLO").AppendLine("WORLD").AppendLine("FROM MONO");
        CharLowerBuffW(sb, (uint)sb.Length);
        Console.WriteLine(sb); // sb is corrupted/freed at this step
    }
}
```

### Expected Behavior

```
hello
world
from mono
```

### Actual Behavior

Segfault
or sometimes StringBuilder contains some random chars at the end after marshalling.

[ ] macOS
[ ] Linux
[x] Windows

filipnavara:
Fixes #15366 and #15306.

- Fixes NULL termination of `StringBuilder` to UTF-16 marshalling. Additionally an extra NULL is added as defense-in-depth protection for incorrect handling in native calls. This extra NULL adds protection when trying to marshal the NULL-terminated string back and the native function doesn't null-terminate at the size of the buffer.
- Fixes marshalling back from UTF-16/UTF-8 buffers into chunked `StringBuilder`.
- Fixes missing `mono_marshal_free` call causing a memory leak when marshalling `StringBuilder` to UTF-16 and back.
- Skips marshalling back to `StringBuilder` when the parameter doesn't have `[Out]` attribute.

Unit tests are still TBD. Not sure where to put them and how to structure them. Any help would be appreciated. /cc @vargaz @EgorBo

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -5483,7 +5483,7 @@ emit_marshal_object_ilgen (EmitMarshalContext *m, int argnum, MonoType *t,
 				}
 
 				mono_mb_emit_byte (mb, CEE_STIND_REF);
-			} else {
+			} else if (t->attrs & PARAM_ATTRIBUTE_OUT || !(t->attrs & PARAM_ATTRIBUTE_IN)) {
 				mono_mb_emit_ldarg (mb, argnum);
 				mono_mb_emit_ldloc (mb, conv_arg);
 
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -758,11 +758,33 @@ mono_string_builder_new (int starting_string_length, MonoError *error)
 static void
 mono_string_utf16_to_builder_copy (MonoStringBuilderHandle sb, const gunichar2 *text, size_t string_len, MonoError *error)
 {
-	MonoArrayHandle chunkChars = MONO_HANDLE_NEW_GET (MonoArray, sb, chunkChars);
-	gchandle_t gchandle = 0;
-	memcpy (MONO_ARRAY_HANDLE_PIN (chunkChars, gunichar2, 0, &gchandle), text, sizeof (gunichar2) * string_len);
-	mono_gchandle_free_internal (gchandle);
-	MONO_HANDLE_SETVAL (sb, chunkLength, int, string_len);
+	MonoArrayHandle chunkChars = MONO_HANDLE_NEW (MonoArray, NULL);
+	MonoStringBuilderHandle chunk = MONO_HANDLE_NEW (MonoStringBuilder, MONO_HANDLE_RAW (sb));
+
+	guint capacity = mono_string_builder_capacity (sb);
+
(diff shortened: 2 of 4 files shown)


mono/metadata/marshal.c
```diff
@@ -1358,11 +1383,6 @@ mono_marshal_get_ptr_to_stringbuilder_conv (MonoMethodPInvoke *piinfo, MonoMarsh
 
 	switch (encoding) {
 	case MONO_NATIVE_LPWSTR:
-		/* 
-		 * mono_string_builder_to_utf16 does not allocate a 
-		 * new buffer, so no need to free it.
-		 */
-		*need_free = FALSE;
```
filipnavara: There are cases where we can skip the allocation in `mono_string_builder_to_utf16`. However, that is currently not done and hence the comment was wrong. I kept the `need_free` argument in case someone decides to optimize this code path later.

mono/metadata/marshal.c
```diff
@@ -758,11 +758,33 @@ mono_string_builder_new (int starting_string_length, MonoError *error)
...
+			int charsToCopy = MIN (string_len - chunkOffset, maxLength);
+			memcpy (MONO_HANDLE_RAW (chunkChars)->vector, text + chunkOffset, charsToCopy * sizeof (gunichar2));
+			MONO_HANDLE_SETVAL (chunk, chunkLength, int, charsToCopy);
+		} else {
+			MONO_HANDLE_SETVAL (chunk, chunkLength, int, 0);
+		}
+		MONO_HANDLE_GET (chunk, chunk, chunkPrevious);
+	} while (MONO_HANDLE_BOOL (chunk));
+
+	MONO_EXIT_NO_SAFEPOINTS;
```
filipnavara: This code is quite long and admittedly not easy to read. Alternative is to go in CoreCLR footsteps and rewrite the function to always spit out `StringBuilder` with single chunk.

mono/metadata/marshal-ilgen.c
```diff
@@ -5483,7 +5483,7 @@ emit_marshal_object_ilgen (EmitMarshalContext *m, int argnum, MonoType *t,
 				}
 
 				mono_mb_emit_byte (mb, CEE_STIND_REF);
-			} else {
+			} else if (t->attrs & PARAM_ATTRIBUTE_OUT) {
```
filipnavara: It is both unnecessary and functionally wrong to marshal back the buffer contents. TBD: Check other places if the check is missing somewhere else too.

CoffeeFlux:
@monojenkins build failed

marek-safar:
@monojenkins backport 2019-06

PR #15393, opened 2019-06-25
interlocked_mutex: 32bit only and not recursive

jaykrell:


jaykrell:

jaykrell:
Static init should suffice too. The lock is exceedingly small & rare.

PR #15395, opened 2019-06-25
Use C++11 thread locals.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
Wow, looks like this is no-go, due to 32bit Watch or something.
I'd'a'thunk C++11 is well supported.

jaykrell:
Does not seem to have the desired portability.

PR #15402, opened 2019-06-25
Support IPv6 requests in EndPointListener and HttpListenerRequest.

rmugridge:
Signed-off-by: Robin Mugridge <robin.mugridge@oxinst.com>

### DIFF
--- a/mcs/class/System/System.Net/HttpListenerRequest.cs
+++ b/mcs/class/System/System.Net/HttpListenerRequest.cs
@@ -213,7 +213,7 @@ namespace System.Net {
 			if (raw_uri != null)
 				host = raw_uri.Host;
 	
-			int colon = host.IndexOf (':');
+			int colon = host.LastIndexOf (':');
 			if (colon >= 0)
 				host = host.Substring (0, colon);
 
--- a/mcs/class/System/System.Net/EndPointListener.cs
+++ b/mcs/class/System/System.Net/EndPointListener.cs
@@ -296,7 +296,7 @@ namespace System.Net {
 		{
 			ArrayList current;
 			ArrayList future;
-			if (prefix.Host == "*") {
+			if ((prefix.Host == "*") || (prefix.Host == "::")) {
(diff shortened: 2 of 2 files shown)


steveisok:
@rmugridge Thanks for the PR.  I'd like to have this covered via tests before pushing it through.

https://github.com/mono/mono/blob/master/mcs/class/System/Test/System.Net/HttpListenerRequestTest.cs

I've started a modifying the tests to get an idea, but if you beat me to the punch by all means.

rmugridge:
@steveisok Apologies regarding the tests - it's my first PR up!

steveisok:
@rmugridge No problem!  I think it's great you submitted and very much appreciate it.  Like I sorta said before, the tests / test helpers are going to need some ipv6 tweaking for it to work.  It'll be learning experience for you and me ;-).

mcs/class/System/System.Net/HttpListenerRequest.cs
```diff
@@ -213,7 +213,7 @@ internal void FinishInitialization ()
 			if (raw_uri != null)
 				host = raw_uri.Host;
 	
-			int colon = host.IndexOf (':');
+			int colon = host.LastIndexOf (':');
```
baulig: This is incorrect and needs to be reworked.  You cannot parse IPv6 address with with a simple `IndexOf()`, it is much more complex than that and I would recommend reusing existing APIs such as for instance `UriParser` for this.

IPv6 does not use colons to separate host and port numbers, but square brackets such as for instance

    https://[2001:db8:85a3:8d3:1319:8a2e:370:7348]:443/

and there can also be a scoped literal appended to it

    fe80::1ff:fe23:4567:890a%eth2

See https://en.wikipedia.org/wiki/IPv6_address for details.

akoeplinger:
Closing this PR for now as it needs some changes and there was no activity in a few months. Feel free to reopen the PR later :)

PR #15406, opened 2019-06-25
[aot] Save the number of locals into the aot debug info so the method header doesn't have to be decoded when its loaded.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #15412, opened 2019-06-25
[wasm] make packager.exe depend on release/ and debug/ because it may…

lewing:
… use either


lewing:
@monojenkins build OS X x64

PR #15415, opened 2019-06-26
Make codeman not public.

jaykrell:
Seems a bit obscure to be public.
The header is not public --  could still dlsym + cast.

### DIFF
--- a/mono/utils/mono-codeman.h
+++ b/mono/utils/mono-codeman.h
@@ -14,22 +14,22 @@ typedef struct {
 	void (*chunk_destroy) (void *chunk);
 } MonoCodeManagerCallbacks;
 
-MONO_API MonoCodeManager* mono_code_manager_new     (void);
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

CoffeeFlux:
Seems to have been made public with https://github.com/mono/mono/commit/4beada28924879ede55f3bf8b3f5b798d0f4637e#diff-59aef377aba8d07b41b1b15da3b84f88, a commit that I know made public some other stuff that shouldn't have been. I don't want to approve since I don't know the history here, but this seems sane to me given the header isn't public.

lateralusX:
Functions doesn't look to be referenced from any of the Xamarin repros and not part of the public header files.

CoffeeFlux:
@monojenkins build failed

PR #15420, opened 2019-06-26
[wasm][sdk][framework] Add versioning to the sdk and framework project files.

kjpou1:
- Add a central versioning file `common\Versions`
- Modify projects to use the central version file
- Auto generate a versions file for the SDK versions so that they can be updated automatically.
- Update Template SDK version from central version file.
- Fix build problem `The local source '/mnt/jenkins/workspace/test-mono-pull-request-wasm/sdks/wasm/sdk/packages' doesn't exist.`


kjpou1:
Was not sure if it was a bit too much or not so decided to err on completeness.  Will change this to use just the SDK version.

kjpou1:

PR #15423, opened 2019-06-26
Add support for TLS callbacks on Windows.

lateralusX:
Mono runtime depends on DllMain to be called by OS in order to correctly detach threads from the runtime. If this doesn't happen, threads from native thread pools (not owned by runtime) and attached
using native->managed callback, won't detach resulting in attached threads no longer running. Since threads are still attached to runtime, next GC will try to suspend the thread and that will fail and that in turn will put the state machine in an incorrect state, bringing down the runtime at next GC (when the thread state is revisited).

This problem is currently handled by DllMain, but since DllMain only exist in DLL's, statically link Mono runtime will get us into the above scenario.

This commit add support to hook up a TLS callback using the same mechanism used by MSVC linker and c-runtime (also available under MINGW). The callback will be included in a section/segment of an object file and included in final image (DLL or EXE) by linker and called by OS loader, solving the problem with static linked Mono runtime not being able to detach terminating threads.

The commit keeps current DllMain method when building a DLL but includes a mechanism making sure we only use one of the callback techniques at runtime, if both have been included in final image. This will simplify the build of libmini since we can always build the object including the callback, regardless
how the library will be consumed.

### DIFF
--- a/mono/mini/mini-windows-tls-callback.c
+++ b/mono/mini/mini-windows-tls-callback.c
@@ -29,10 +29,8 @@ mono_win32_handle_tls_callback_type (MonoWin32TLSCallbackType callback_type)
 	return TRUE;
 }
 
-MONO_EXTERN_C
 VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID reserved);
 
-MONO_EXTERN_C
 VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID reserved)
 {
 	mono_win32_runtime_tls_callback ((HMODULE)module_handle, reason, reserved, MONO_WIN32_TLS_CALLBACK_TYPE_LIB);
@@ -44,12 +42,12 @@ VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID res
 /* by OS loader (part of TLS Directory PE header), regardless if runtime is used */
 /* as a static or dynamic library. */
 #if (_MSC_VER >= 1400)
-#pragma const_seg(".CRT$XLX")
-MONO_EXTERN_C const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback = mono_win32_tls_callback;
-#pragma const_seg()
+#pragma const_seg (".CRT$XLX")
+extern const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback = mono_win32_tls_callback;
+#pragma const_seg ()
 #elif defined(__MINGW64__) || (__MINGW32__)
-MONO_EXTERN_C const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback __attribute__ ((section (".CRT$XLX"))) = mono_win32_tls_callback;
+extern const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback __attribute__ ((section (".CRT$XLX"))) = mono_win32_tls_callback;
 #else
-#pragma message("TLS callback support not included in .CRT$XLX segment. Static linked runtime won't add callback into PE header.")
+#pragma message ("TLS callback support not included in .CRT$XLX segment. Static linked runtime won't add callback into PE header.")
 #endif
 #endif
--- a/mono/mini/mini-windows-tls-callback.c
+++ b/mono/mini/mini-windows-tls-callback.c
@@ -1,0 +1,55 @@
+/**
+ * Copyright 2019 Microsoft Corporation
+ * Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ */
+
+#include "mini-runtime.h"
+
+#if defined(HOST_WIN32)
+#include "mini-windows.h"
+#include <windows.h>
+
+/* Only reset by initial callback from OS loader with reason DLL_PROCESS_ATTACH */
+/* both for DLL and LIB callbacks, so no need to protect this variable. */
+static MonoWin32TLSCallbackType mono_win32_tls_callback_type = MONO_WIN32_TLS_CALLBACK_TYPE_NONE;
+
+/* NOTE, this function needs to be in this source file to make sure linker */
+/* resolve this symbol from mini-windows.c, picking up callback and */
+/* included it in TLS Directory PE header. Function makes sure we only activate */
+/* one of the supported mono callback types at runtime (if multiple one has been used) */
+gboolean
+mono_win32_handle_tls_callback_type (MonoWin32TLSCallbackType callback_type)
+{
+	if (mono_win32_tls_callback_type == MONO_WIN32_TLS_CALLBACK_TYPE_NONE)
+		mono_win32_tls_callback_type = callback_type;
+
+	if (callback_type != mono_win32_tls_callback_type)
+		return FALSE;
+
+	return TRUE;
+}
+
+MONO_EXTERN_C
+VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID reserved);
+
+MONO_EXTERN_C
+VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID reserved)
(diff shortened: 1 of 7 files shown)


mono/mini/mini-windows-tls-callback.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/* If we are building a static library that won't have access to DllMain we can't */
+/* correctly detach a thread before it terminates. The MSVC linker + runtime (also applies to MINGW) */
+/* uses a set of predefined segments/sections where callbacks can be stored and called */
+/* by OS loader (part of TLS Directory PE header), regardless if runtime is used */
+/* as a static or dynamic library. */
+#if (_MSC_VER >= 1400)
+#pragma const_seg(".CRT$XLX")
+MONO_EXTERN_C const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback = mono_win32_tls_callback;
```
jaykrell: Why `extern "C"` here? I don't think it is needed.

mono/mini/mini-windows.c
```diff
@@ -444,3 +445,29 @@ mono_thread_state_init_from_handle (MonoThreadUnwindState *tctx, MonoThreadInfo
...
+		return FALSE;
+
+	switch (reason)
+	{
+	case DLL_PROCESS_ATTACH:
+		mono_install_runtime_load (mini_init);
+		break;
+	case DLL_PROCESS_DETACH:
+		if (coree_module_handle)
+			FreeLibrary (coree_module_handle);
```
jaykrell: Probably should check `reserved` and do nothing if it indicates ExitProcess, instead of FreeLibrary.
It isn't actually `reserved. It has a documented meaning.
Cleanup within ExitProcess tends not to work.

mono/mini/mini-windows-tls-callback.c
```diff
@@ -0,0 +1,55 @@
...
+	if (mono_win32_tls_callback_type == MONO_WIN32_TLS_CALLBACK_TYPE_NONE)
+		mono_win32_tls_callback_type = callback_type;
+
+	if (callback_type != mono_win32_tls_callback_type)
+		return FALSE;
+
+	return TRUE;
+}
+
+MONO_EXTERN_C
```
jaykrell: Why `extern "C"`?

mono/mini/mini-windows-tls-callback.c
```diff
@@ -0,0 +1,55 @@
...
+VOID NTAPI mono_win32_tls_callback (PVOID module_handle, DWORD reason, PVOID reserved);
+
+MONO_EXTERN_C
```
jaykrell: Why `extern "C"`?

jaykrell:
You might also consider `FlsAlloc`'s callback as a place to cleanup.
At least for threads, not for FreeLibrary. (`FlsAlloc` is equivalent to `TlsAlloc`, if fibers are not in used, except it provides the cleanup callback.)
`FlsAlloc` is "interesting" because the callback is outside the loader lock.

jaykrell:
Also please keep in mind that none of this stuff works the way people think.
It is documented correctly.

https://jaykrell2.blogspot.com/2017/11/dllthreadattach-and-dllthreaddetach-are.html


You do not get DLL_ATTACH for threads that exist before the runtime is loaded, except the loading thread.
You do not get DLL_DETACH for threads that remain existing when a .dll is unloaded.

jaykrell:
The way you can cleanup all your thread state, is by maintaining a separate list of threads.
In the FreeLibrary callback, walk them.
MFC of all things provides support for this in its TLS helpers.

Another popular strategy is give up on FreeLibrary.
Including by calling `GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_PIN)` to prevent the reference count from ever reaching 0.

mono/mini/mini-windows-tls-callback.c
jaykrell: No need for parens on `#if`.

mono/mini/mini-windows-tls-callback.c
```diff
@@ -0,0 +1,55 @@
...
+gboolean
+mono_win32_handle_tls_callback_type (MonoWin32TLSCallbackType callback_type)
+{
+	if (mono_win32_tls_callback_type == MONO_WIN32_TLS_CALLBACK_TYPE_NONE)
+		mono_win32_tls_callback_type = callback_type;
+
+	if (callback_type != mono_win32_tls_callback_type)
+		return FALSE;
+
+	return TRUE;
```
jaykrell: `return callback_type == mono_win32_tls_callback_type;`

Maybe the idemptotence belongs at the next level down though, I.e. mono should tolerate multiple attach and detach?

mono/mini/mini-windows-tls-callback.c
jaykrell: Spaces after parens? (`const_seg (`)

mono/mini/mini-windows-tls-callback.c
```diff
@@ -0,0 +1,55 @@
...
+
+/* If we are building a static library that won't have access to DllMain we can't */
+/* correctly detach a thread before it terminates. The MSVC linker + runtime (also applies to MINGW) */
+/* uses a set of predefined segments/sections where callbacks can be stored and called */
+/* by OS loader (part of TLS Directory PE header), regardless if runtime is used */
+/* as a static or dynamic library. */
+#if (_MSC_VER >= 1400)
+#pragma const_seg(".CRT$XLX")
+MONO_EXTERN_C const PIMAGE_TLS_CALLBACK __mono_win32_tls_callback = mono_win32_tls_callback;
+#pragma const_seg()
```
jaykrell: There is also `__declspec(allocate)` but I'm not sure the advantages/disadvantages.

jaykrell:
Oh darn, the reserved parameter to DllMain is defined and useful, but the one for TLS callbacks is not. That seems like a mistake. :(

mono/mini/mini-windows-tls-callback.c
lateralusX: Just prevents unnecessary duplicate calls for mono related callbacks (since we now can have to active in the process).

mono/mini/mini-windows-tls-callback.c
lateralusX: Does similar job, but a little harder to open/close sections using that method (depending on where you are in the source), but the result is similar.

mono/mini/mini-windows-tls-callback.c
lateralusX: extern "C" needed when building as C++ or it won't be picked up by linker.

jaykrell: Maybe it just needs to be extern, not externC? Linker is just using section name. Static is probably removed by compiler though.

mono/mini/mini-windows-tls-callback.c
lateralusX: Was just kept inline with DllMain, but since it's just used within this source, it could do without it.

mono/mini/mini-windows-tls-callback.c
lateralusX: Was just kept inline with DllMain, but since it's just used within this source, it could do without it.

mono/mini/mini-windows-tls-callback.c
lateralusX: Not sure, not using extern keyword at all still works for C but not for C++ compiler.

jaykrell: This is because `const` implies `static` in C++ but `extern` in C.
So you have to say `extern const` to get the same meaning in both.
And then the compiler will optimize away the unreferenced static.
We were pretty careful to only apply `externC` where needed.

lateralusX:
@monojenkins rebuild failed

lewing:
monojenkins build Linux ARMv7 Interpreter

PR #15438, opened 2019-06-27
Fix suspend thread failure behavior on Windows.

lateralusX:
According to documentation of mono_threads_suspend_begin_async_suspend:

"If begin suspend fails the thread must be left uninterrupted and resumed."

Current implementation on Windows didn't resume thread in case GetThreadContext fails, leaving a thread in suspended state on failure.

Fix is to resume thread if GetThreadContext fails.

PR also fix the use of THREAD_SUSPEND_DEBUG macro in the same file, since current implementation had both compile error and warnings (not used before).

### DIFF
--- a/mono/utils/mono-threads-windows.c
+++ b/mono/utils/mono-threads-windows.c
@@ -196,6 +196,8 @@ mono_threads_suspend_begin_async_suspend (MonoThreadInfo *info, gboolean interru
 	context.ContextFlags = CONTEXT_INTEGER | CONTEXT_CONTROL;
 	if (!GetThreadContext (handle, &context)) {
 		THREADS_SUSPEND_DEBUG ("SUSPEND FAILED (GetThreadContext), id=%p, err=%u\n", GUINT_TO_POINTER (id), GetLastError ());
+		result = ResumeThread (handle);
+		g_assert (result == 1);
 		return FALSE;
 	}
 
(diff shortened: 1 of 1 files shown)


mono/utils/mono-threads-windows.c
```diff
@@ -193,6 +195,9 @@ mono_threads_suspend_begin_async_suspend (MonoThreadInfo *info, gboolean interru
 	CONTEXT context;
 	context.ContextFlags = CONTEXT_INTEGER | CONTEXT_CONTROL;
 	if (!GetThreadContext (handle, &context)) {
+		THREADS_SUSPEND_DEBUG ("SUSPEND FAILED (GetThreadContext), id=%p, err=%u\n", GUINT_TO_POINTER (id), GetLastError ());
+		result = ResumeThread (handle);
+		g_assert (result == 1);
```
jaykrell: Could anyone else have suspended it? Is this over-asserting?

lateralusX: We have had the same assert in case we fail to change state as well for a long time, so if they fire, some external code probably suspended our thread underneath us. If we suspend a thread using mono runtime (in one way or another) we already keep a suspend count on the thread, meaning that this method should only be called ones, so believe the assert could give us some valuable information if they fire.

lewing:
@monojenkins build failed

mono/utils/mono-threads-windows.c
jaykrell: Sometimes I suspend threads in debugger to aid single threaded debugging. Granted, I do it rarely. People also suspend threads sometimes when applying Detours. I wonder if it would ever be useful vs. rarely misfire and take things down unnecessarily.

PR #15447, opened 2019-06-27
Pass the --clr-memory-model flag on the command line instead of MONO_DEBUG so its not inherited by child processes.

vargaz:


lewurm:
good idea, but `--clr-memory-model` isn't a mono option yet, right?

lewurm:
nevermind, the parsing code is actually smart.

vargaz:
@monojenkins build failed

PR #15449, opened 2019-06-27
[debugger] Update client thread frames after SetIP.

### LINKED ISSUE #13408: [sdb] SetIP requires a resume/suspend to update client thread frames

thaystg:
PR of draft https://github.com/mono/mono/pull/14667

Fixes #13408

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -427,7 +427,7 @@ namespace Mono.Debugger.Soft
 		 * with newer runtimes, and vice versa.
 		 */
 		internal const int MAJOR_VERSION = 2;
-		internal const int MINOR_VERSION = 51;
+		internal const int MINOR_VERSION = 52;
 
 		enum WPSuspendPolicy {
 			NONE = 0,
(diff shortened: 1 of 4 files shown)


vargaz:
Looks ok, but wondering whenever this is a good idea. The ip will point to one location, but the locals etc. will correspond to another location.

thaystg:
@monojenkins build failed

jbevain:
@vargaz isn't it the desired behavior here? For our scenario:

1) You stop at a breakpoint at offset `A`
2) The variables you see are at the time of `A`
3) You tell the IDE that you'll resume from offset `B`
4) The IDE shows you that the thread is now positioned at offset B with variables having values as set at `A`.

It feels like it works for our use case.

The problem right now is that we have to handle the «set next statement» by either creating a fake breakpoint and doing a resume, another possibility would be to special case the current IP for a thread which is what MD does, but is annoying to implement.

jbevain:
@thaystg as far as I'm concerned, I'm OK with this.

thaystg:
@monojenkins build failed

PR #15450, opened 2019-06-27
Disable compiler server on darwin. Add force-enable option.

kg:


kg:
@monojenkins build OS X x64

kg:
@monojenkins build Linux x64 mcs

kg:
@monojenkins build compiler server

kg:
@monojenkins build OS X x64

kg:
@monojenkins build Linux x64 mcs

kg:
@monojenkins build compiler server

kg:
@monojenkins build Linux x64

jaykrell:
Rude questioning: Is there coverage on other platforms i.e. to keep it enabled?
What does CoreCLR do i.e. does it have working compiler server & on which platforms?

kg:
There's a CI lane that tests it on linux, but it doesn't run automatically. (Surprise: It broke)
We don't have any other coverage yet. Even if it's enabled and tests pass it's probably still broken because the roslyn compiler server is effectively broken by design.
One step at a time...

jaykrell:
That's kinda what I meant:

 - Does it work anywhere?
 - Is it enabled anywhere?

Hopefully the answer to both is the same.

And, is there really no way to ask csc to do multiple things
in the same process, either sequentially or in parallel?

Doesn't msbuild already use some in-proc form of csc?

kg:
> That's kinda what I meant:
>     * Does it work anywhere?

Yes, it works pretty consistently on Linux.

>     * Is it enabled anywhere?

Yes, we enabled it by default for all 'make' builds of the BCL a little while ago. That's why we need to disable it for darwin, because pipes on Mac (or something related) seem hosed.

> And, is there really no way to ask csc to do multiple things
> in the same process, either sequentially or in parallel?

There is not. It also insists on using a single compiler server process (w/shared GC heap, and threads) instead of round-robin processes unix style. Our GC doesn't currently like this very much.

> Doesn't msbuild already use some in-proc form of csc?

MSBuild uses a variant of the roslyn compiler server which is also broken right now. It does invoke CSC in-process (or in the compiler server). This PR is to fix make being slow on mac.

kg:
@monojenkins build

lewing:
@monojenkins backport 2019-06

PR #15452, opened 2019-06-27
[netcore] Fix Assembly.GetForwardedTypes()

### LINKED ISSUE #15078: [netcore] Make System.Reflection.Tests.AssemblyNetCoreAppTests.AssemblyGetForwardedTypesLoadFailure Pass

```
System.Reflection.Tests.AssemblyNetCoreAppTests.AssemblyGetForwardedTypesLoadFailure [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.Reflection.ReflectionTypeLoadException)
      Actual:   typeof(System.TypeLoadException): Could not load type '' from assembly ''.
      ---- System.TypeLoadException : Could not load type '' from assembly ''.
      Stack Trace:
           at System.Reflection.Tests.AssemblyNetCoreAppTests.<>c__DisplayClass1_0.<AssemblyGetForwardedTypesLoadFailure>b__0()
        ----- Inner Stack Trace -----
           at System.Reflection.Tests.AssemblyNetCoreAppTests.<>c__DisplayClass1_0.<AssemblyGetForwardedTypesLoadFailure>b__0()
```

Determine why and make it pass.

### LINKED ISSUE #15079: [netcore] System.Reflection.Tests.AssemblyNetCoreAppTests.AssemblyGetForwardedTypes

```
System.Reflection.Tests.AssemblyNetCoreAppTests.AssemblyGetForwardedTypes [FAIL]
      Assert.Equal() Failure
      Expected: Type[] [typeof(string), typeof(TypeInForwardedAssembly), typeof(TypeInForwardedAssembly+PublicInner), typeof(TypeInForwardedAssembly+PublicInner+PublicInnerInner)]
      Actual:   Type[] [typeof(string), typeof(TypeInForwardedAssembly)]
      Stack Trace:
           at System.Reflection.Tests.AssemblyNetCoreAppTests.AssemblyGetForwardedTypes()
```

CoffeeFlux:
Fixes #15078 
Fixes #15079 

This consists of two main changes: throwing the proper `ReflectionTypeLoadException` and fetching inherited types (which was done in managed).


netcore/System.Private.CoreLib/src/System.Reflection/RuntimeAssembly.cs
```diff
@@ -137,7 +137,48 @@ internal static AssemblyName CreateAssemblyName (string assemblyString, out Runt
...
+			Type[] nestedTypes;
+
+			try {
+				nestedTypes = type.GetNestedTypes (BindingFlags.Public);
+			} 
+			catch (FileLoadException e) { exceptions.Add (e); return; }
+			catch (FileNotFoundException e) { exceptions.Add (e); return; }
+			catch (TypeLoadException e) { exceptions.Add (e); return; }
+			catch (IOException e) { exceptions.Add (e); return; }
+			catch (UnauthorizedAccessException e) { exceptions.Add (e); return; }
```
EgorBo: maybe it's better to do:
```csharp
catch (Exception e) when (IsLoaderException(e)) { exceptions.Add (e); return; }
```

CoffeeFlux: I was told specifically not to do that because exception filters don't work properly on some platforms we support.

vargaz: That would use an exception filter which we are trying to avoid.

EgorBo: ah, good to know

mono/metadata/icall.c
```diff
@@ -6056,23 +6061,43 @@ ves_icall_System_Reflection_RuntimeAssembly_GetForwardedTypes (MonoReflectionAss
 
 		mono_assembly_load_reference (image, assembly_idx - 1);
 		g_assert (image->references [assembly_idx - 1]);
-		if (image->references [assembly_idx - 1] == (gpointer)-1)
+		if (image->references [assembly_idx - 1] == (gpointer) - 1) {
```
lambdageek: could use `REFERENCE_MISSING`

CoffeeFlux:
Rebased so I could re-enable tests without merge conflicts. Should be good now.

mono/metadata/icall.c
```diff
@@ -6056,23 +6061,43 @@ ves_icall_System_Reflection_RuntimeAssembly_GetForwardedTypes (MonoReflectionAss
...
+			MonoExceptionHandle ex = MONO_HANDLE_NEW (MonoException, mono_get_exception_bad_image_format ("Invalid image"));
+			MONO_HANDLE_ARRAY_SETREF (types, aindex, NULL_HANDLE);
+			MONO_HANDLE_ARRAY_SETREF (exceptions, aindex, ex);
+			exception_count++; aindex++;
 			continue;
+		}
 		MonoClass *klass = mono_class_from_name_checked (image->references [assembly_idx - 1]->image, nspace, name, error);
-		if (!is_ok (error))
+		if (!is_ok (error)) {
+			MonoExceptionHandle ex = mono_error_convert_to_exception_handle (error);
```
lambdageek: So there's a problem here in that it leaves `error` in an unusuable state (by design).  (Same in other places where you do `if (!is_ok (error)) { ... convert ... continue;}`)

In this case we _know_ we're not going to propagate the error out.  We want to capture it and store it.

In order to do that, a better approach is to make a new  local`ERROR_DECL (captured_error)` inside the loop body and pass that to all these functions.

```
   for (...) {
      ERROR_DECL (captured_error);

      do_something (..., captured_error);
      if (!is_ok (captured_error)) {
            MonoExceptionHandle ex = mono_error_convert_to_exception_handle (captured_error);
            MONO_HANDLE_ARRAY_SETREF (exceptions, aindex, ex);
            aindex++;
            continue;
      }
      ...
  }
```

CoffeeFlux:
@monojenkins build failed

PR #15457, opened 2019-06-28
[wasm][sdk] Fix generation of version.

kjpou1:
The version is incorrectly generated.

- The package reference of `ClientWebSocketPackageVersion` is generated with an appended `1` causing an error being generated when restoring.


lewing:
@monojenkins build failed

Issue #15441, opened 2019-06-27
"pos undeclared" compilation error


volo-zyko:
This code doesn't compile since 20c06cc0f40167a388be2f939b44d838876a979f when in `mono_arch_emit_epilog_sub` was removed `int pos = 0`

https://github.com/mono/mono/blob/886c49017470112af1a73034d8951d6ca00b641e/mono/mini/mini-mips.c#L5300

marek-safar:
@jaykrell could you please address the build breakage

PR #15459, opened 2019-06-28
Fix year+ old MIPS compile break.

jaykrell:
Fixes https://github.com/mono/mono/issues/15441


akoeplinger:
@monojenkins backport 2019-06

PR #15135, opened 2019-06-14
[System.Windows.Forms] Fix crash of DataGridView in case of assigning a custom Column Header.

NVoronchev:
This PR fix crash of `DataGridView` in case of assigning a custom Column Header.

PR #15461, opened 2019-06-28
[2019-06] [System.Windows.Forms] Fix crash of DataGridView in case of assigning a custom Column Header.

NVoronchev:
This PR fix crash of `DataGridView` in case of assigning a custom Column Header.

Backport of #15135 

/cc @marek-safar


marek-safar:
approve

PR #15464, opened 2019-06-28
[merp] Fix API regression

alexischr:
The regression is the accidental removal of RegisterReportingForNativeLib from the managed side, introduced by b919ac4#diff-939ea9baecd2e8b24260bcb33f531274

### DIFF
--- a/mcs/class/corlib/Mono/Runtime.cs
+++ b/mcs/class/corlib/Mono/Runtime.cs
@@ -194,6 +194,18 @@ namespace Mono {
 		}
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern void RegisterReportingForNativeLib_internal (IntPtr modulePathSuffix, IntPtr moduleName);
+
+		static void RegisterReportingForNativeLib (string modulePathSuffix_str, string moduleName_str)
+		{
+			using (var modulePathSuffix_chars = RuntimeMarshal.MarshalString (modulePathSuffix_str))
+			using (var moduleName_chars = RuntimeMarshal.MarshalString (moduleName_str))
+			{
+				RegisterReportingForNativeLib_internal (modulePathSuffix_chars.Value, moduleName_chars.Value);
+			}
+		}
+
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		static extern void EnableCrashReportLog_internal (IntPtr directory);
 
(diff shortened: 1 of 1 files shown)


alexischr:
@lambdageek please see new commit/description. I restored the missing icall, and followed its structure. Things should make more sense now.

This is a regression at least back to 2018-10, but at the same time, that API doesn't seem to have been used yet.

mono/mini/mini-exceptions.c
```diff
@@ -1446,6 +1454,21 @@ mono_crash_reporting_register_native_library (const char *module_path, const cha
...
+	// Example: "libsystem_pthread.dylib:pthread:libsystem_platform.dylib:platform" will register two modules ("pthread" and "platform")
+	char **tokens = g_strsplit (module_string, G_SEARCHPATH_SEPARATOR_S, 0);
+	while (*tokens) {
+		char *module_path = *tokens;
+		tokens++;
+		char *module_name = *tokens;
+		tokens++;
+		mono_crash_reporting_register_native_library (module_path, module_name);
+	}
+}
```
lambdageek: Needs a `g_strfreev` to free `tokens`.

mono/mini/mini-exceptions.c
```diff
@@ -1446,6 +1454,21 @@ mono_crash_reporting_register_native_library (const char *module_path, const cha
...
 
+static void
+mono_crash_reporting_batch_register_native_libraries (const char *module_string)
+{
+	// Accepts a ":"-delimited string where each pair of entries is module_path + module_string
+	// Example: "libsystem_pthread.dylib:pthread:libsystem_platform.dylib:platform" will register two modules ("pthread" and "platform")
+	char **tokens = g_strsplit (module_string, G_SEARCHPATH_SEPARATOR_S, 0);
+	while (*tokens) {
+		char *module_path = *tokens;
+		tokens++;
```
lambdageek:
```suggestion
		tokens++;
		g_assert (*tokens);
```
My comment got lost.  I think we could use an assert here in case the input is malformed.

lewing:
@monojenkins build failed

alexischr:
After discussion we decided against this approach so this will be just be a PR for the regression

lewing:
@monojenkins build failed

alexischr:
@monojenkins build failed

lewing:
@monojenkins build failed

PR #15465, opened 2019-06-28
[windows] Ensure that mono_native_tls_get_value restores last error value

filipnavara:
On MinGW64 builds the interpreter is incorrectly overwriting last error values on P/Invoke calls. I traced it down to the usage of `TlsGetValue` which always overrides the last error value. The documentation states the following: 

> Functions that return indications of failure call SetLastError when they fail. They generally do not call SetLastError when they succeed. The TlsGetValue function is an exception to this general rule. The TlsGetValue function calls SetLastError to clear a thread's last error when it succeeds. That allows checking for the error-free retrieval of zero values.

The TLS code path is hit from the following line:

https://github.com/mono/mono/blob/112297ef5c076a2716c020ddb0bbf1e0a2da7703/mono/mini/interp/interp.c#L2003

Note that this is not a performance concern for official builds since the MSVC builds always use `__declspec(thread)` feature. MinGW should probably support `__thread` but configure script doesn't detect it.

### DIFF
--- a/mono/utils/mono-tls.h
+++ b/mono/utils/mono-tls.h
@@ -55,7 +55,15 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
 #define mono_native_tls_alloc(key,destructor) ((*(key) = TlsAlloc ()) != TLS_OUT_OF_INDEXES && destructor == NULL)
 #define mono_native_tls_free TlsFree
 #define mono_native_tls_set_value TlsSetValue
-#define mono_native_tls_get_value TlsGetValue
+
+static inline LPVOID mono_native_tls_get_value (MonoNativeTlsKey key)
+{
+	DWORD last_error = GetLastError ();
+	LPVOID result = TlsGetValue (key);
+	g_assert (GetLastError () == 0);
+	SetLastError (last_error);
+	return result;
+}
 
 #else
 
--- a/mono/utils/mono-tls.h
+++ b/mono/utils/mono-tls.h
@@ -58,10 +58,10 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
 
 static inline LPVOID mono_native_tls_get_value (MonoNativeTlsKey key)
 {
-	DWORD last_error = GetLastError ();
+	W32_DEFINE_LAST_ERROR_RESTORE_POINT;
 	LPVOID result = TlsGetValue (key);
 	g_assert (GetLastError () == 0);
-	SetLastError (last_error);
+	W32_RESTORE_LAST_ERROR_FROM_RESTORE_POINT;
 	return result;
 }
 


lambdageek:
Made an issue about MinGW `__thread` --> https://github.com/mono/mono/issues/15466

mono/utils/mono-tls.h
```diff
@@ -55,7 +55,15 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
 #define mono_native_tls_alloc(key,destructor) ((*(key) = TlsAlloc ()) != TLS_OUT_OF_INDEXES && destructor == NULL)
 #define mono_native_tls_free TlsFree
 #define mono_native_tls_set_value TlsSetValue
-#define mono_native_tls_get_value TlsGetValue
+
+static inline LPVOID mono_native_tls_get_value (MonoNativeTlsKey key)
```
jaykrell: 1. Cosmetic: “L” is a 16bit-ism. Prefer modern “PVOID”.
2. Use __declspec(thread) under “#if _MSC_VER” it is tremendously more efficient, like 3 cheap instructions, no lock, no call.

filipnavara: 1. I used exactly what is in the documentation.
2. That is already done elsewhere. This code will not get hit if __declspec(thread) is supported.

jaykrell:
The documentation is based on 16bit-isms.
“L” is to aid portability to portability to 16bit.
Prefer PVOID or void* or gpointer.
They are all the same but “L” is pointless.

filipnavara:
@jaykrell I know the history behind it but I still prefer to use the types that are in the documentation, even if long pointers are a history.

filipnavara:
Apparently I broke all the AOT tests, will investigate.

jaykrell:
Breakpoint on SetLastError is somewhat useful. Suggest only call it if GetLastError shows value changed.

Also please check, I will also, I think you only need to do the restore if TLS value is zero. TLS non-zero seems to imply clear success and therefore LastError unchanged. I’ll double check and try to get documentation clarified.

This isn’t the only case like this btw. GetFileSize and SetFilePointer have similar problems when file size or new seek position is 4GB-1.

jaykrell:
Oh darn. On the one point, it does always zero lastError, even if returning non-zero.

filipnavara:
I was unable to reproduce the AOT failures locally. Rebased & force push to see if it still happens.

filipnavara:

filipnavara:

filipnavara:

filipnavara:

filipnavara:

jaykrell:
Would still be nice to avoid Set if second Get returns same as first Get, i.e. 0. Otherwise arm interpreter lane seems skipable.

mono/utils/mono-tls.h
```diff
@@ -55,7 +55,15 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
...
+{
+	DWORD last_error = GetLastError ();
+	LPVOID result = TlsGetValue (key);
+	g_assert (GetLastError () == 0);
+	SetLastError (last_error);
+	return result;
```
jaykrell: Would be nice to skip Set if Gets return same value, i.e. 0.

jaykrell: And both variables could be const.

lateralusX:
At other places having similar issues we have fixed it at call site not to affect all callers of this rather low level method, maybe we should do similar here? So instead of fixing at mono_native_tls_get_value we should do it at the interpreter method that would like to preserve last error from pinvoke. We have W32_DEFINE_LAST_ERROR_RESTORE_POINT and W32_RESTORE_LAST_ERROR_FROM_RESTORE_POINT that can be used to accomplish this.

filipnavara:
@lateralusX Fair enough, I am fine with changing it. The reason I went with this approach was to ensure the same behavior between different build configurations. Builds that use `__declspec(thread)` never hit this error and it was extremely difficult to diagnose and find the roor cause.

jaykrell:
I think it is right like this (except for const and avoiding the write if the value matches).

I think it belongs with the code corrupting the last error, not with the code requiring error be preserved, because it is surprising that the error gets corrupted here.
Because 1. People do not check if this function succeeds (though it can fail -- only if you give it a bad index -- not under low memory), and 2. It clears last error always.

3. This is only on the mingw path, not the msvc path.

lateralusX:
TlsGetValue is used under msvc as well.

jaykrell:
@filipnavara that is another good pont -- what? You would ifdef thread_keyword at the callers (or I guess introduce another set of lasterror preserving macros, with the ifdef builtin). I think it is good as it is. Mingw is bad, and now even worse. Maybe eventually mingw will get proper `__thread` support. (It has `__thread` support, and it is very poor).

jaykrell:
Ah. True, but TlsGetValue can and should be avoided under msvc.
Really, every use of it, under msvc, should be switched to `__declspec(thread)`.

jaykrell:
Imho if it needs to be fast, it is using msvc and __declspec(thread).

lateralusX:
We could look into switching for msvc, but has been this way for a long time. We have some users of mingw builds that doesn't want slow down in low level methods like this, so adding the additional amount of code that this PR introduce on all TLS access is not good.

jaykrell:
Imho they are already slow, and the right fix is to make them faster than they even were before.
Msvc should just skip this code always.
Otherwise, well, you have to review all the uses of mono_native_tls_get_value and decide if they need error preserved, which I guess you can. There are around 50 uses.
Also the uses that are conditionally __declspec(thread) need to be checked.

filipnavara:
@lateralusX The problem for me is maintainability, not performance. If we can find a way that achieves both goal then I am happy to do that. Right now the official builds are MSVC compiled and they didn't hit this error in the first place (AFAICT). The error was happening only for MinGW builds and it was hard to find and diagnose what was going wrong because it didn't assert/crash and continued executing code with wrong values. As long as the offical builds are MSVC only and the unit tests don't cover all the code path it is bound to break again.

jaykrell:
I see both sides but I agree more with @filipnavara.
1. Convert msvc to use __declspec(thread). msvc will be fast.
2. Leave this slow and now slower path for mingw only.

But I do understand -- not all code cares about last error, and it could be left mostly asis.

jaykrell:
I have a prototype where using __declspec(thread) is a little more convenient.
I was writing it, with the intent of switching mingw to it, until I realized mingw isn't good. But it'll still help convert remaining TlsGetValue uses to declspec(thread) for msvc.
Give me a few days/weeks..

mono/utils/mono-tls.h
jaykrell: If you are going to call GetLastError, then please compare to the original, and only call SetLastError if they vary. For example if it was already zero.
Or maybe remove the second GetLastError call?

Also, maybe, maybe, use a fs/gs write, instead of a function all.
I'm not sure the offset is public though. Need to check.
That could, could make this slowdown a total of two instructions -- two movs, no calls. But I need to check..

jaykrell: Is this good enough? https://en.wikipedia.org/wiki/Win32_Thread_Information_Block

Using that, you can change GetLastError and SetLastError to one instruction each, mov to or from fs:34 or gs:68.

filipnavara: @jaykrell I rewrote it using the macros used elsewhere. They do the appropriate checks. If we want to access TEB directly on some platforms it could be done in separate PR in the macros. I believe some versions of MSVC headers used to implement `GetLastError` as inline function. The MinGW header don't do that AFAIK.

jaykrell:
Another gross thing to consider, see if CoreCLR accesses Teb->TlsSlots and Teb->TlsExpansionSlots. They are in winternl.h. Then you get much faster TlsGetValue and no LastError corruption.

But really msvc __declspec(thread) is even faster and definitely public.

filipnavara:
> But really msvc __declspec(thread) is even faster and definitely public.

I believe it may be possible to achieve the same effect in MinGW with few hacks. However, the recent addition of TLS callbacks makes it somewhat tricky to do so. I can look into it later this week.

jaykrell:
Not conclusive, but I don't think CoreCLR access the TEB fields.
Anyway, msvc __declspec(thread), slow mingw imho.

lateralusX:
not sure how easy the switch to declspec thread will be for these TLS variables since they are using key as part of our codegen as well, meaning that we need a way to get hold of the tls keys.

jaykrell:
I don't think the TLS callbacks help MinGW. I think give up on MinGW until they fix __thread.
Well, maybe if you can control the PE fields well enough, but I don't think you can.

lateralusX:
So as part of our codegen we also access the TEB fields based on TLS keys.

jaykrell:
Why do you need the keys?
There are two ways to achieve "fast" access.

1. Call out to a helper function, that just uses __declspec(thread).
This will be better than it is now. Where maybe it calls out TlsGetValue, slower.

2. You can inline the access using public information, like so.

Teb->TlsVector is an array with a pointer per .dll
_tls_index is a global variable.

Every variable is at Teb->TlsVector[_tls_index] + some offset.

You find the offset by finding it in any thread.

&foo - Teb->TlsVector[_tls_index];

Store that in a global.

And then you can have fast Win32 TLS inlined JIT.
Not for AOT though. That'd be different.

filipnavara:
> So as part of our codegen we also access the TEB fields based on TLS keys.

It's rather trivial to codegen the same thing as compiler does for `__declspec(thread)` variable access. I am not suggesting to do it since it would be a big change but it's an option.

jaykrell:
Oh, we access TlsSlots or TlsExpansionSlots?
Are you sure that code is in there?
I think it got removed years ago.
There is no Win32 fast TLS in the current tree.


Well, anyway, I outlined how you do that with __declspec(thread).
The result is maybe one instruction longer, but uses all public information (it mimics what the native compiler outputs).

jaykrell:
@filipnavara Yes it is a good option. But only with msvc and __declspec(thread), so that the allocation is handled correctly. TlsAlloc values are applied differently, and you shouldn't really use that except with TlsGetValue / TlsSetValue. I know CLR used to and mono used to, but I think those are no longer that way.

jaykrell:
For AOT you just all out to a helper that uses __declspec(thread), or the AOT accesses the global variable that you initialize early. You don't actually inline the offset with AOT, but at best a relocated pointer to it. Still pretty good.

jaykrell:
@lateralusX I think e7e6e6c697227e79030dc75cd51541d7844b156e removed the code you are thinking of. Right? I do not suggest bringing that back.

jaykrell:
Oh, maybe here:
```
Searching for 'TlsExpansionSlots'...
C:\s\mono2\mono\mini\mini-amd64.c(3791):		/* Load TEB->TlsExpansionSlots */
C:\s\mono2\mono\mini\mini-x86.c(2219):		/* Load TEB->TlsExpansionSlots */
2 occurrence(s) have been found.
```

jaykrell:
So then there becomes the question of MinGW JIT TLS quality.

lateralusX:
Was looking in the source and it looks like we still have the get implementation in our codegen mono_amd64_emit_tls_get used when lowering OP_TLS_GET.

jaykrell:
@lateralusX agreed. It is kinda gross, since these are internals.
Arguably it is the basis of TlsAlloc + fast/lasterror-preserving TlsGetValue in C/C++ though. Right?
You know -- what the JIT does, you can do in C/C++.

vargaz:
Instead of changing TLS, the special case should be fixed in the interpreter and the jit imho. I assume its SetLastError on DllImport declarations.

lateralusX:
That was my point as well, fix it on call site (as we done in the past), not affecting all calls to low level API's. Optimize TLS and make it work under all scenarios we support could be a different (bigger) work item for future.

filipnavara:
@vargaz @lateralusX #15534... still need to test it on my use cases until I drop it from Draft status.

jaykrell:
The JIT code looks suspicious when combined with this:

filipnavara:
> Instead of changing TLS, the special case should be fixed in the interpreter and the jit imho.

It does not affect JIT, at least not without an attached debugger.

> I assume its SetLastError on DllImport declarations.

Correct.

lateralusX:
Looked a little deeper at this since later changes around fast TLS didn't really made sense. Looks like we indeed use declspec(thread) under MSVC for most TLS keys used by runtime and the old codegen lowering OP_TLS_GET has been eliminated due to the way MONO_THREAD_VAR_OFFSET gets defined on Windows (not very clear or visible) so it will always fallback using icalls for TLS access from generated code. Fallback to icall is always the case for AOT on all platforms, put currently also in play on Windows for JIT.

I believe we should look into optimize the TLS access on Windows for JIT scenarios going forward as well (similar to what we had in the past with OP_TLS_GET) but now also matching the MSVC implementation of declspec (thread).

jaykrell:
Yeah, I had come to this conclusion a while ago, but not so quickly/easily today.

I guess it is because of this:

```
/* Runtime offset detection */
#if defined(TARGET_AMD64) && !defined(TARGET_MACH) && !defined(HOST_WIN32) /* __thread likely not tested on mac/win */
```

I suggest we `g_assert_not_reached ()` the TLS lowering for now (in Windows).
And go ahead and remove it probably.

jaykrell:
Also AOT Windows TLS access can be well optimized not just JIT.

lateralusX:
Yes, but currently looks like all AOT platforms fallback to icalls in this scenario.

PR #15453, opened 2019-06-27
Use /mnt/scratch for Android SDK, to match https://github.com/xamarin/xamarin-android/pull/3125

directhex:
Ought to cause fewer failures during archive job.

PR #15469, opened 2019-06-28
Correct env vars used for Android directory override

directhex:
This fixes https://github.com/mono/mono/pull/15453

The issue is the env var names differ vs. Xamarin.Android

this value comes from sdks/paths.mk


directhex:
jenkins isn't testing anything I changed, so squashing it early.

directhex:

PR #15470, opened 2019-06-28
[System.Net.Http]: Delete broken `Send_Transfer_Encoding_Custom()` test.

baulig:
This test was checking a broken Mono-specific behavior, which has been removed after we switched to the new corefx-based handler.  It also didn't make much sense to use a `Transfer-Encoding` header with a GET request.


baulig:
I'm going to remove the locally added tests again, since my upstream PR as already been approved.

steveisok:
@baulig Please move this out of draft and get the right tests in today if you could.

baulig:
> @baulig Please move this out of draft and get the right tests in today if you could.

I'm waiting for the upstream PR to land.

baulig:
@steveisok There actually won't be anything left in the PR except for bumping CoreFX once they've merged my PR.

baulig:
The new tests will be in https://github.com/dotnet/corefx/pull/39039, we need to backport that and bump.

akoeplinger:
@monojenkins backport 2019-06

PR #15473, opened 2019-06-28
Add TPN entry for GCC

richlander:


marek-safar:
@richlander could you please resolve the conflict

PR #15478, opened 2019-06-29
Fix const warning.

jaykrell:


jaykrell:
@monojenkins build failed

PR #15480, opened 2019-06-30
Update zlib from 1.2.5 to 1.2.11.

jaykrell:
Circa 2010 to circa 2017.

Retrieved with apt-get source zlib on WSL Ubuntu 18.04.2 LTS.

This matches Debian or Ubuntu (they differ by one commit, and one of them restores a dummy function for compat).


jaykrell:
Let’s cycle for jenkins.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

jaykrell:

CoffeeFlux:
Did you get a chance to go through the zlib changelogs and see if there's anything that could be an issue? I'll try to do so myself once this exists draft status.

jaykrell:
This does add functions, that we don't use.
I experimented with not doing that, but instead the only diff against upstream is the `ZCALLBACK` one.
So easier maintenance.

I think we should rename stuff to start `mono_`, but we use the names asis.
There is provision in zlib for renaming stuff reasonably easily, all in one place (`zconf.h` I think it was).

I'd prefer to use the zlib configure and Makefile, maybe lightly modified, but that'd be additional change.
Because of otherwise this wonky extract/subset of zlib/configure into mono/configure.

The configure flags affect all of mono, which is good and bad.
It provides consistency, good, is hypothetically extra change, bad, but is probably no change at all really, ok. They could be scoped to support, or even files therein if desired, trading consistency for change.

The largefile thing is wonky, in that I think it only affects code we don't use.
I went far down the rabbit hole of removing code we don't use, but eventually gave up.
Maybe revisit that.

jaykrell:
Debian matches github.com/madler/zlib/commits/master

Ubuntu 1804 matches same
 plus one commit from github.com/madler/zlib/commits/develop
 plus restoring one function for compat.

github.com/madler/zlib/commits/develop is supposed to released soon.

This matches Ubuntu or Debian, plus bringing forward one minor Mono change (see the commits).

jaykrell:
The first commit is just a copy, not a fancy merge, plus adding gzguts.h, plus the manual configure.ac stuff. The second commit is the only merge.

jaykrell:
I could move configure to separate commit if desired.

jaykrell:
@monojenkins build failed

jaykrell:
/cc @CoffeeFlux looking good, finally.
I had to hack on ioapi.c quite a bit for Android.
Probably should look into what zlib Android does ship, though maybe they just don't ship minizip -- it is "contrib" not "real". But we do use at least some of it.

jaykrell:
Having looked at AOSP, `#define USE_FILE32API` was the right solution. Let me go back to that..

jaykrell:
https://android.googlesource.com/platform/external/zlib suggests USE_FILE32API, but what I have seems maybe better. Maybe.

jaykrell:
The android use of USE_FILE32API is from 2014. I wonder if that is best now.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
After this goes in @grendello might fix autoconf to use system zlib on Android.
Anyway this will be used on Windows, maybe WebAssembly, and wherever autoconf says.

configure.ac
```diff
@@ -668,8 +668,20 @@ AC_PROG_LD_GNU
 
 AC_CHECK_HEADERS(sys/filio.h sys/sockio.h netdb.h utime.h sys/utime.h semaphore.h sys/un.h linux/rtc.h sys/syscall.h sys/mkdev.h sys/uio.h sys/param.h sys/sysctl.h libproc.h sys/prctl.h copyfile.h)
 AC_CHECK_HEADERS(sys/param.h sys/socket.h sys/ipc.h sys/utsname.h alloca.h ucontext.h pwd.h sys/select.h netinet/tcp.h netinet/in.h unistd.h sys/types.h link.h asm/sigcontext.h sys/inotify.h arpa/inet.h complex.h unwind.h)
-AC_CHECK_HEADER(unistd.h, [HAVE_UNISTD_H=1], [HAVE_UNISTD_H=0])
+
+# zlib/configure checks for stdarg.h existance and defines HAVE_STDARG_H on compiler
+# command line (not config.h) if it is present.
+CFLAGS="$CFLAGS -DHAVE_STDARG_H"
```
CoffeeFlux: Is there a reason you're not using `AC_CHECK_HEADER` here but are for unistd.h?

mcs/class/System/System.IO.Compression/DeflateStream.cs
```diff
@@ -611,6 +611,7 @@ static int WriteZStream (SafeDeflateStreamHandle stream, IntPtr buffer, int leng
 			}
 		}
 #else
+		// FIXME? Names should start "mono"?
```
CoffeeFlux: If we're going to do this, it should probably be in a separate PR (though I'm not familiar enough to have a strong opinion on the matter), so I don't see any use to these comments.

configure.ac
```diff
@@ -299,6 +299,9 @@ case "$host" in
 		CFLAGS="$CFLAGS -DPAGE_SIZE=4096 -DS_IWRITE=S_IWUSR"
 		CXXFLAGS="$CXXFLAGS -DPAGE_SIZE=4096 -DS_IWRITE=S_IWUSR"
 
+		# FIXME?  Only if __ANDROID_API__ < 24?
```
CoffeeFlux: An Android person should probably look at this - I don't know enough to sign off one way or another.

configure.ac
jaykrell: zlib does not use config.h.

configure.ac
jaykrell: If you read the Android headers, as of ABI 24, they have the `o` functions (`o` is for offset, vs. regular `long`).
`32` means `long`, which is 32 on 32bit, 64 on 64bit, and never widened on 32bit.
`o` stands a chance of being widened on 32bit.

Android's zlib seems stuck at `32` despite better being available.
Ultimately @grendello is likely to fixup our build after this PR to never use the zlib we carry.

It all affects the ability to use compressed files larger than 4gb.

jaykrell: (@grendello on Android specifically; maybe we should ifdef android error in our zlib, now or in that)

jaykrell:
This should be rebased to master also given the lag. Maybe wait a few hours to lighten Jenkins load.

mcs/class/System/System.IO.Compression/DeflateStream.cs
jaykrell: Comments are to form a memory. A github issue should be another option.

configure.ac
jaykrell: The comment explains -- "compiler command line...not config.h"

CoffeeFlux: I'm confused because you put the same comment next to unistd.h and then used AC_CHECK_HEADER anyway, but I guess you were just preserving the existing behavior in our configure?

CoffeeFlux: Maybe I'm just misunderstanding something - I haven't dealt with autotools much.

jaykrell: Oh, sorry. This is because I am assuming stdarg.h exists. Which mono already does in a few places. And should exist in any implementation since at least 1989. Maybe some day we port to K&R, the older varargs form?
zlib does not assume stdarg.h exists though.

unistd.h is at least Posix only, and sometimes mono assumes it exists, sometimes not.
There a few forms to unistd.h in mono which seems dubious.

jaykrell: Maybe this is overkill though.

```
#if defined(STDC) || defined(Z_HAVE_STDARG_H)
#  ifndef Z_SOLO
#    include <stdarg.h>         /* for va_list */
#  endif
#endif

#if !defined(STDC) && (defined(__STDC__) || defined(__cplusplus))
#  define STDC
#endif
```

configure.ac
```diff
@@ -670,8 +673,20 @@ AC_PROG_LD_GNU
 
 AC_CHECK_HEADERS(sys/filio.h sys/sockio.h netdb.h utime.h sys/utime.h semaphore.h sys/un.h linux/rtc.h sys/syscall.h sys/mkdev.h sys/uio.h sys/param.h sys/sysctl.h libproc.h sys/prctl.h copyfile.h)
 AC_CHECK_HEADERS(sys/param.h sys/socket.h sys/ipc.h sys/utsname.h alloca.h ucontext.h pwd.h sys/select.h netinet/tcp.h netinet/in.h unistd.h sys/types.h link.h asm/sigcontext.h sys/inotify.h arpa/inet.h complex.h unwind.h)
-AC_CHECK_HEADER(unistd.h, [HAVE_UNISTD_H=1], [HAVE_UNISTD_H=0])
+
+# zlib/configure checks for stdarg.h existance and defines HAVE_STDARG_H on compiler
+# command line (not config.h) if it is present.
+CFLAGS="$CFLAGS -DHAVE_STDARG_H"
```
lambdageek: I think we should put all the zlib defines into `ZLIB_EXTRA_CFLAGS` (or whatever name you like) and pass them to the compiler in `support/Makefile.am` using a target specific cflags argument.   I don't think we should put a bunch of stuff into the global CFLAGS just because zlib don't use a config file.

configure.ac
jaykrell: And the only use is here:

```
#if defined(STDC) || defined(Z_HAVE_STDARG_H)
#  ifndef Z_SOLO
ZEXTERN int            ZEXPORTVA gzvprintf Z_ARG((gzFile file,
                                                  const char *format,
                                                  va_list va));
#  endif
#endif
```

which is unused and I deleted the file I think. So the configure.ac change might be doubly unneeded, and also _fairly_ harmless. I do like short compiler command lines though and this regresses that.

configure.ac
jaykrell: That creates a dependency on including config.h before zlib.h.
Which appears to already be met but previously wasn't required.
Or knowing which files use zlib and using ZLIB_EXTRA_CFLAGS only on them.

I was aiming for a simple consistency, at the cost of possibly unused options.
Though on the other hand, I didn't want to add `#include config.h` to zlib either, giving it even more defines, in case they risky.

We do have the same `HAVE_STDARG_H` as them, so it kinda isn't pollution -- compiler command line vs. config.h, but shared valid semantic.

The stdarg.h one I can just remove it appears.

Mono misuses CFLAGS also as I understand.
It is supposed to be for the user, like -O0 -g, not things mono actually requires.

jaykrell: Oh and the `HAVE_HIDDEN` and `_LARGEFILE64_SOURCE`. `_LARGEFILE64_SOURCE` seems like possible trouble either way, I'd have to read up on it again.

configure.ac
```diff
@@ -6591,9 +6605,25 @@ if test x"$GCC" = xyes; then
...
+# defines _LARGEFILE64_SOURCE on the compiler command line (not config.h).
+#
+AC_MSG_CHECKING([for large file support])
+AC_COMPILE_IFELSE(
+	[
+		#include <sys/types.h>
+		off64_t dummy = 0;
+	], [
+		AC_MSG_RESULT(yes)
+		CFLAGS="$CFLAGS -D_LARGEFILE64_SOURCE=1"
```
jaykrell: Still need to reread what all this does. Messy stuff.

jaykrell:
Maybe a separate PR for the configure.ac/Makefile changes?

CoffeeFlux:
@monojenkins build failed

jaykrell:
Let's take https://github.com/mono/mono/pull/15697/files first and then rebase this. Just for readable history. And I guess dial back the "name should start mono_" comments here, so this PR becomes only updating/copying source.
It is a little tenuous in general, in that configure/makefile could be mutually dependent on the source change, or be meaningless on its own, but in this case, the first is not true, and the second imho is ok and I like the history to read "that way" -- edits of our code vs. copies of not our code.

CoffeeFlux:
Sounds good - I've just been going through and re-triggering the failed CI lanes for a ton of PRs since I want to get stuff merged.

CoffeeFlux:
Okay merged #15697, so we should be good to rebase this.

jaykrell:
I missed the point about added files and I think you are right, for the tarball build.

jaykrell:
Ok rebased, and carefully cultivated the commits, so it can/should go in unsquashed, no commit will be broken, except *maybe* one commit on Visual C++, even that I doubt, and the history should be fairly clean and readable. i.e. all edits that aren't just file copies are in separate commits.

Alternatively, the Makefile.am edits and a placeholder gzguts.h could in via a separate PR.
Arguably this is what I should have done with the earlier configure/Makefile change, sorry.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Okay, this looks fine except that I think you need to add gzguts.h to the msvc solution as well as the makefile, ideally just amending the last commit. After that, I'll rebase-merge the whole batch.

jaykrell:
There is no real need to add .h files to the .proj files. They just pretty up the gui.

jaykrell:
There is a need to in Makefile.am though, so it gets included in tarball. I'll check.

jaykrell:
I added to the project files. It is just noise. And I rebased to master.

CoffeeFlux:
@monojenkins build failed

support/README.zlib
```diff
@@ -5,3 +5,17 @@ systems that might have shipped older versions of zlib and to reduce
...
+Sources of truth:
+	https://packages.ubuntu.com/disco/zlib1g (or newer, but you get the idea)
+	https://packages.debian.org/stretch/zlib1g-dev (or newer, but you get the idea)
+	https://github.com/madler/zlib/tree/master
+	https://github.com/madler/zlib/tree/develop
+	mono commit b5a0b8e0a8bb2fe790dc2f34a1a846f8efc470db (57ea99c5011aa2d9ce2be308041aafe76a469e5c)
+	Possibly other mono commits.
+
+Debian matches master
+Ubuntu 1804 has one commit from master and one extra function for compat.
```
jaykrell: Should say devel not master.

akoeplinger: Fixed it in master.

PR #15482, opened 2019-07-01
Align native_handle open/close for MonoThreadInfo/MonoInternalThread on Windows.

lateralusX:


mono/utils/mono-threads.c
```diff
@@ -1775,6 +1775,39 @@ mono_threads_close_thread_handle (MonoThreadHandle *thread_handle)
...
+MonoNativeThreadHandle
+mono_threads_open_native_thread_handle (MonoNativeThreadHandle thread_handle)
+{
+#ifdef HOST_WIN32
+	BOOL success = FALSE;
+	HANDLE new_thread_handle = NULL;
+
+	g_assert (thread_handle && thread_handle != INVALID_HANDLE_VALUE);
+	success = DuplicateHandle (GetCurrentProcess (), thread_handle, GetCurrentProcess (), &new_thread_handle, 0, FALSE, DUPLICATE_SAME_ACCESS);
+	g_assertf (success, "Failed to duplicate thread handle");
```
jaykrell: This can fail under low memory and imho should be removed.

jaykrell: `return success ? new_thread_handle : NULL;`

jaykrell: And then really `return DuplicateHandle (…) ? new_thread_handle : NULL;`

mono/utils/mono-threads.c
```diff
@@ -1775,6 +1775,39 @@ mono_threads_close_thread_handle (MonoThreadHandle *thread_handle)
...
+	return NULL;
+#endif
+}
+
+void
+mono_threads_close_native_thread_handle (MonoNativeThreadHandle thread_handle)
+{
+#ifdef HOST_WIN32
+	g_assert (thread_handle != INVALID_HANDLE_VALUE);
+	if (thread_handle)
```
jaykrell: There is a sort of confusing sublte inconsistency between `NULL` and `INVALID_HANDLE_VALUE`.
I get it though -- `NULL` is the "real" invalid value, that could route through mono and back to here. `INVALID_HANDLE_VALUE` should never make it out of Win32-specific code.

You might assert that `DuplicateHandle` never returns `INVALID_HANDLE_VALUE`, at least on the success path, but that would be pointless really.

You might put in a comment acknowledging or explaining the different handling of `NULL` and `INVALID_HANDLE_VALUE`.

mono/utils/mono-threads.c
lateralusX: This code was mostly moved from old implementation, but I can make it more compact.

lateralusX: The assert has been there for a long time, so lets keep it for now.

mono/utils/mono-threads.c
jaykrell: Maybe just remove the assert about `INVALID_HANDLE_VALUE`.
Don't mention `INVALID_HANDLE_VALUE` at all.

mono/utils/mono-threads.c
jaykrell: I think it is all ok, just subtle in the NULL vs. INVALID_HANDLE_VALUE handling, except that low memory should not lead to assertion failures.

mono/utils/mono-threads.c
lateralusX: NULL is what's used by mono to detect invalid value or lack of value, so that's the most important one. Added check for INVALID_HANDLE_VALUE since it should never really happen, but is a valid error state mostly indicating an implementation error elsewhere in the code (using a different way to setup handle), so keeping it in the Win32 specific code adds value.

mono/utils/mono-threads.c
lateralusX: Checked history of the assert and it looks "quite" recently added, 2883e562e2d2, so agree that we should get rid of it.

lateralusX:
@monojenkins rebuild failed

lateralusX:
@monojenkins rebuild failed

lateralusX:
@lambdageek approve/merge?

mono/utils/mono-threads.c
```diff
@@ -1775,6 +1775,36 @@ mono_threads_close_thread_handle (MonoThreadHandle *thread_handle)
...
+mono_threads_open_native_thread_handle (MonoNativeThreadHandle thread_handle)
+{
+#ifdef HOST_WIN32
+	BOOL success = FALSE;
+	HANDLE new_thread_handle = NULL;
+
+	g_assert (thread_handle && thread_handle != INVALID_HANDLE_VALUE);
+	return DuplicateHandle (GetCurrentProcess (), thread_handle, GetCurrentProcess (), &new_thread_handle, 0, FALSE, DUPLICATE_SAME_ACCESS) ? new_thread_handle : NULL;
+#else
+	return MONO_GPOINTER_TO_NATIVE_THREAD_HANDLE (NULL);
```
jaykrell: just `return 0` here?

PR #15483, opened 2019-07-01
[stats] distinguish between mini and LLVM AOT method

lewurm:
Also print the stats on shutdown when configured with `--with-core=only`


lewurm:
@monojenkins squash

lewing:
@monojenkins build

lewurm:
@monojenkins build failed

PR #15486, opened 2019-07-01
Abort async suspend request on failure doing preemptive suspend.

lateralusX:
When a preemptive suspend fails, state machine will get into an invalid state and the thread is in an unknown state (running/self-suspended). In case of failure in mono_threads_suspend_begin_async_suspend, a new method should be called, mono_threads_transition_abort_async_suspend making sure state machine reflects correct thread state as well as reporting correct result to initial caller of mono_threads_suspend_begin_async_suspend.


lateralusX:
@monojenkins rebuild failed

lateralusX:
@monojenkins rebuild failed

Issue #15153, opened 2019-06-14
[netcore] Make System.Tests.ActivatorTests.TestingCreateInstanceObjectHandleFullSignature Pass


steveisok:
The test currently fails with:

```
System.Tests.ActivatorTests.TestingCreateInstanceObjectHandleFullSignature(assemblyName: "mscorlib", type: "System.Nullable`1[[System.Int32, mscorlib, Version"..., ignoreCase: True, bindingAttr: Instance | Public, binder: DefaultBinder { }, args: [], culture: , activationAttributes: null, returnedFullNameType: "", returnNull: True) [FAIL]
      System.TypeLoadException : Could not load type 'System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
      Stack Trace:
           at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
           at System.Activator.CreateInstanceInternal(String assemblyString, String typeName, Boolean ignoreCase, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes, StackCrawlMark& stackMark)
           at System.Activator.CreateInstance(String assemblyName, String typeName, Boolean ignoreCase, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
           at System.Tests.ActivatorTests.TestingCreateInstanceObjectHandleFullSignature(String assemblyName, String type, Boolean ignoreCase, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes, String returnedFullNameType, Boolean returnNull)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

PR #15488, opened 2019-07-01
[netcore] Don't default to mscorlib in Assembly.GetType

### LINKED ISSUE #15075: [netcore] Make System.Reflection.Tests.AssemblyTests.GetType_DoesntSearchMscorlib Pass

```
Starting:    System.Reflection.Tests (parallel test collections = on, max threads = 12)
    System.Reflection.Tests.AssemblyTests.GetType_DoesntSearchMscorlib [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.TypeLoadException)
      Actual:   (No exception was thrown)
      Stack Trace:
           at System.Reflection.Tests.AssemblyTests.GetType_DoesntSearchMscorlib()
```

CoffeeFlux:
Fixes #15075

The `else if` block didn't appear to be hit regardless when running the netcore tests, but I don't know for sure whether it's dead code in classic Mono so I've just ifdef'd it.

### DIFF
--- a/mono/metadata/reflection-internals.h
+++ b/mono/metadata/reflection-internals.h
@@ -41,7 +41,7 @@ MonoType *
 mono_reflection_type_handle_mono_type (MonoReflectionTypeHandle ref_type, MonoError *error);
 
 MonoType*
-mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, mono_bool ignorecase, gboolean search_mscorlib, mono_bool *type_resolve, MonoError *error);
+mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, gboolean ignorecase, gboolean search_mscorlib, gboolean *type_resolve, MonoError *error);
 
 MonoType*
 mono_reflection_type_from_name_checked (char *name, MonoImage *image, MonoError *error);
--- a/mono/metadata/reflection-internals.h
+++ b/mono/metadata/reflection-internals.h
@@ -41,7 +41,7 @@ MonoType *
 mono_reflection_type_handle_mono_type (MonoReflectionTypeHandle ref_type, MonoError *error);
 
 MonoType*
-mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, mono_bool ignorecase, mono_bool *type_resolve, gboolean searchMscorlib, MonoError *error);
+mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, mono_bool ignorecase, gboolean search_mscorlib, mono_bool *type_resolve, MonoError *error);
 
 MonoType*
 mono_reflection_type_from_name_checked (char *name, MonoImage *image, MonoError *error);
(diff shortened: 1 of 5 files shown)


EgorBo:
Does it also fix https://github.com/mono/mono/issues/15153 ? Seems related.

CoffeeFlux:
It does not, and it seems to break a bunch of other netcore tests, so I've added do-not-merge while I figure things out.

CoffeeFlux:
"If the type is in the currently executing assembly or in Mscorlib.dll, it is sufficient to supply the type name qualified by its namespace." Seems that while Assembly.GetType should not search mscorlib by default, Type.GetType does.

CoffeeFlux:
Should be fixed now - I've added an argument to a bunch of internal/static functions saying whether or not to fall back to mscorlib. If called anywhere but Assembly_InternalGetType I've preserved the old behavior, but for Assembly.GetType() it should only search the current assembly.

CoffeeFlux:
Oh, and it still doesn't fix #15153, so no 2-for-1 deal :(

mono/metadata/reflection-internals.h
```diff
@@ -41,7 +41,7 @@ MonoType *
 mono_reflection_type_handle_mono_type (MonoReflectionTypeHandle ref_type, MonoError *error);
 
 MonoType*
-mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, mono_bool ignorecase, mono_bool *type_resolve, MonoError *error);
+mono_reflection_get_type_checked (MonoImage *rootimage, MonoImage* image, MonoTypeNameParse *info, mono_bool ignorecase, mono_bool *type_resolve, gboolean searchMscorlib, MonoError *error);
```
lambdageek: I need to go through the whole thing in mode detail, but two quick nits: s/`searchMscorlib`/`search_mscorlib` - mono uses lowercase and underscores, mostly.  Aand: `type_resolve` is an out argument (I think...) so I think `search_mscorlib` should come before it.

CoffeeFlux: Yeah whoops, not sure what I was thinking on the name. Updated.

lewing: what is the difference between mono_bool vs gboolean?

jaykrell: `mono_bool` is public.
`gboolean` is private.
The choice of `int32_t` is poor though.
The type of Boolean expressions in C is `int`.
The type of Booleans in managed code is MonoBoolean -- unsigned char.
So this is a potentially third type (it could be `long`, not `int` -- it should just be `int`).

C:\s\mono2\mono\utils\mono-publib.h(88):typedef int32_t		mono_bool;
C:\s\mono2\mono\eglib\glib.h(198):typedef int32_t        gboolean;

jaykrell: Using both in the same signature seems pointless, agreed.

jaykrell: The only nice thing about `int32_t` here, is consistency on systems with 16bit `int`.
Anyway, no real difference, cosmetic. Usually the private names are shorter so preferred.
What to use on public signature implementations, not clear.
Maybe should rename it to `gbool` for brevity.

CoffeeFlux: That function using `mono_bool` at all is a mistake - what Jay said about public vs private usage is correct, and this is an internal function. The signature is even correct in reflection.c, so I assume the header was just an oversight in refactoring at some point. Updated.

CoffeeFlux:
@monojenkins build failed

steveisok:

PR #4484, opened 2017-03-07
[runtime] Implement support for embedded ppdb, when the ppdb blob is …

vargaz:
…compressed and embedded into the pe image.

PR #15493, opened 2019-07-01
Disable zlib on mingw Windows build to avoid dependency on zlib1.dll.

lateralusX:
https://github.com/mono/mono/pull/4484 broke Windows mingw build due to new dependency, zlib1.dll not available in PATH on build bots. Disabling for now until we know how this should be handled (and also fixed for MSVC build runtime).

### DIFF
--- a/mono/metadata/debug-mono-ppdb.c
+++ b/mono/metadata/debug-mono-ppdb.c
@@ -29,7 +29,7 @@
 #include <mono/utils/bsearch.h>
 #include <mono/utils/mono-logger-internals.h>
 
-#if defined (HAVE_SYS_ZLIB)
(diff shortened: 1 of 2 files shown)


jaykrell:
I think making sure HAVE_SYS_ZLIB is not defined for HostWin32 is the solution.
Basically what you have here but a little less hacky -- configure.ac/winconfig.h instead.
And statically link zlib. That we have a copy of.

lateralusX:
I did not have local repro of the problem so just did something that I know would work to unblock the broken builds. We still need to revisit this since no support was done for MSVC build, so then we can look into alternatives handling this both on mingw and msvc.

PR #15499, opened 2019-07-02
AC_DEFINE instead of AC_SUBST MONO_DL_NEED_USCORE.

jaykrell:
i.e. so it can be used in C instead of make.
i.e. so it can actually work.
Alternatively since it is not working, could remove it.

The lone use is here:
```
C:\s\mono2\mono\utils\mono-dl.c(271):#if MONO_DL_NEED_USCORE
```


jaykrell:
@monojenkins build failed

PR #15500, opened 2019-07-02
HAVE_DL_LOADER is never used so remove it.

jaykrell:


jaykrell:
@monojenkins build failed

PR #15502, opened 2019-07-02
Follow convention: "static inline" is for .h, "static" is for .c.

jaykrell:
inlining is up to the compiler to decide.
It is used in .h files for linkage/dedup, not to affect codegen.

Establishing the consistency is worth the "line damage" imho.


jaykrell:
@monojenkins build failed

BrzVlad:
@monojenkins build failed

PR #15505, opened 2019-07-02
Cleanup zlib-helper.c.

jaykrell:
- Cleanup allocation -- leave integer overflow checks to `calloc`.
 - Assume "recent" zlib (10 years old is recent from this point of view).

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -7893,6 +7893,7 @@ ves_icall_System_IO_get_temp_path (MonoError *error)
 
 #if defined(ENABLE_MONODROID) || defined(ENABLE_MONOTOUCH) || defined(TARGET_WASM)
 
+// FIXME? Names should start "mono"?
 G_EXTERN_C gpointer CreateZStream (gint32 compress, MonoBoolean gzip, gpointer feeder, gpointer data);
 G_EXTERN_C gint32   CloseZStream (gpointer stream);
 G_EXTERN_C gint32   Flush (gpointer stream);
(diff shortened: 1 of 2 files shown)


jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

mcs/class/System/System.IO.Compression/DeflateStream.cs
```diff
@@ -611,6 +611,7 @@ static int WriteZStream (SafeDeflateStreamHandle stream, IntPtr buffer, int leng
 			}
 		}
 #else
+		// FIXME? Names should start "mono"?
```
akoeplinger: I don't think this is worth fixing as I've seen external user code with a `DllImport ("MonoPosixHelper", ..)` that would be broken by this.

jaykrell: Yeah, it is a tough situation, agreed.

Because, now (and forever already) mono can end up replacing system libz.so.
You know, if it loads first, on most Unix systems, with their flat namespace.
So the responsibility to export a very good version is high.
I think.

I can just remove the comments then.

jaykrell: All of the names?

jaykrell: i.e. just the zlib ones or also the helper ones?

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

Issue #15503, opened 2019-07-02
Fix zlib linkage on mingw/msvc builds.


lateralusX:
https://github.com/mono/mono/pull/4484 added dependency on zlib for systems supporting it. Problem on mingw builds is that it will add a dependency on zlib1.dll (since we are building runtime as a dynamic library). That caused build breakage on CI since, by default it couldn't locate zlib.dll. The PR did not include any solution for the MSVC build regarding zlib either. We should come up with a solution for mingw/msvc on how we should consume zlib and we already include parts of needed library in mono posix helper library, so maybe a better alternative is to utilize that than the system zlib, giving use some more control.

The build break was temporary fixed by https://github.com/mono/mono/pull/15493, but ones we have a strategy in place handling zlib on Windows, that fix should be revisited.

PR #15509, opened 2019-07-02
Win32 zlib support.

jaykrell:
Fixes https://github.com/mono/mono/issues/15503

### DIFF
--- a/mono/metadata/Makefile.am
+++ b/mono/metadata/Makefile.am
@@ -99,10 +99,10 @@ if ENABLE_MONODROID
 support_sources += \
 	../../support/nl.c \
 	../../support/zlib-helper.c
-if HAVE_ZLIB
-# See libmonoruntime_support_la_LDFLAGS
-else
-support_sources += \
+endif
+endif
+
+zlib_sources = \
 	../../support/adler32.c \
 	../../support/crc32.c \
 	../../support/crc32.h \
@@ -122,7 +122,17 @@ support_sources += \
 	../../support/zlib.h \
 	../../support/zutil.c \
 	../../support/zutil.h
-endif
+
+#
+# Support is a separate library because it is not valid C++.
+#
+if HOST_WIN32
+support_sources += $(zlib_sources)
+else
+if HAVE_ZLIB
+# See libmonoruntime_support_la_LDFLAGS
+else
+support_sources += $(zlib_sources)
 endif
 endif
 
@@ -151,8 +161,7 @@ libmonoruntime_config_la_SOURCES = \
 libmonoruntime_config_la_CPPFLAGS = $(AM_CPPFLAGS) -DMONO_BINDIR=\"$(bindir)/\" -DMONO_ASSEMBLIES=\"$(assembliesdir)\" -DMONO_CFG_DIR=\"$(confdir)\" -DMONO_RELOC_LIBDIR=\"../$(reloc_libdir)\" @CXX_ADD_CFLAGS@
 
 #
-# This library is only used to compile the required support/ sources. We can't use them directly in platform_sources
-# because they aren't C++ compatible.
+# Support is a separate library because it is not valid C++.
 #
 libmonoruntime_support_la_SOURCES = $(support_sources)
 libmonoruntime_support_la_CFLAGS = $(filter-out @CXX_REMOVE_CFLAGS@, @CFLAGS@)
--- a/msvc/libmonoruntime.targets
+++ b/msvc/libmonoruntime.targets
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)monozlib.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmonoruntime-common.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmonoruntime-win32.targets" />
   <Import Project="$(MSBuildThisFileDirectory)libmonoruntime-posix.targets" />
--- a/configure.ac
+++ b/configure.ac
@@ -684,23 +684,32 @@ AC_CHECK_HEADERS(sys/user.h, [], [],
 
 AC_CHECK_HEADERS(linux/serial.h)
 
-AC_CHECK_HEADER(zlib.h, [have_zlib=yes], [have_zlib=no])
-if test x$have_zlib = xyes; then
-   AC_TRY_COMPILE([#include <zlib.h>], [
-   #if defined(ZLIB_VERNUM) && (ZLIB_VERNUM >= 0x1230)
-   #else
-   #error No good zlib found
-   #endif
-   ],[
-	AC_MSG_RESULT(Using system zlib)
-	zlib_msg="system zlib"
-	AC_DEFINE(HAVE_SYS_ZLIB,1,[Have system zlib])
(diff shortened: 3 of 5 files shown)


lateralusX:
I think we should do something more structured around this since we already have similar things in posix helper library used on Windows platforms. There is also another pattern for zlib features used on mobile that could be used as an alternative, so I don't think we should add a third solution into the mix, instead we should adjust to use one of the other approaches for this solution.

jaykrell:
It should not be a separate .dll imho.
I don't know why mobile does that. It seems like a bad idea.
It could be a separate static .lib though.
I didn't want to write .proj file is all. I can try copying the eglib one..

jaykrell:
I really don't like the msbuild proj files.
Why does it take 10 of them to build eglib?
```
 Directory of C:\s\mono5\msvc

04/14/2019  01:15 AM             2,299 eglib-common.targets
04/14/2019  01:15 AM             5,139 eglib-common.targets.filters
04/14/2019  01:15 AM             1,278 eglib-posix.targets
04/14/2019  01:15 AM             1,306 eglib-posix.targets.filters
04/14/2019  01:15 AM               830 eglib-win32.targets
04/14/2019  01:15 AM             1,798 eglib-win32.targets.filters
04/14/2019  01:15 AM               376 eglib.targets
04/14/2019  01:15 AM               915 eglib.targets.filters
04/14/2019  02:15 AM             8,810 eglib.vcxproj
04/14/2019  02:15 AM               629 eglib.vcxproj.filters
              10 File(s)         23,380 bytes
```

jaykrell:
I'll try it. I think it can be just one proj file for zlib.

lateralusX:
Project files have been structured like this to closer assembler makefiles and make it easier to reuse them for other build configurations (where you can't reuse the project files as such). It is used by external users as well inside their builds, so the structure adds value both to them and us.

It's in a DLL since we are using p/invoke to consume it from C# code. Mobile platforms uses icalls for some of the zlib support, so that's a second way of doing it and this is a third way So this was my thought, we should move into a solution that could be shared by all of them, probably a static lib that could be shared by both runtime and posix helper library.

lateralusX:
Believe we should make a zlib project, but do one .target file similar to the common one, that way we could reuse it in other build configurations, like in the dynamic library build configuration as well as on none desktop Windows platforms.

lateralusX:
We would need something similar for our mingw builds as well, not just for msvc.

jaykrell:
I think it should be a static lib in all cases, linked into the runtime, and always icalls, never p/invoke.
But for some reason Unix uses icalls sometimes, p/invoke others.
I'm not going to change that today.
This is neither icall nor p/invoke.
Imho it should either build in mini\Makefile or metdata\makefile or support\makefile or maybe a new zlib\makefile.
But I'm not going to change the Makefiles today either.

For Windows today, a static lib, either msvc\support.proj or msvc\zlib.proj or msvc\monozlib.proj.
For now I have msvc\zlib.proj.
I don't think it should be lumped with the rest of the "support" stuff. It should be a separate zlib.
I'm not going to move any of the files, but just in terms of naming the .proj.

jaykrell:
mingw: good point/reminder thank you.

I think for those what I was saying where we just tweak the HAVE_SYS_ZLIB will do.
We want that to be false, even if there is a sys_zlib. Don't detect it, just say it isn't there.
Unless there is a static .lib option.
And then, for Win32, compile zlib, like we do already, at least sometimes.

jaykrell:
Hm, we do already have a support.proj. So maybe just use that..maybe it already works actually, for non-mingw.. I'd prefer zlib.proj, but if support.proj is there and works..

jaykrell:
Except that it is a .dll which seems wrong.

lateralusX:
Problem with mingw build is that just using -lz as it currently does will use toolchains pick of zlib and since we are building a shared library of mono runtime it will pick the dynamic, not static version of zlib (if found during configure). So either we need to make sure we do static link of these libraries under mingw when picking up system version of zlib, or always fallback to our own build zlib version, regardless if we found a system wide zlib version or not.

vargaz:
We should build and link our own zlib then, the last version is 1.2.11 from 2017.

lateralusX:
Where did you find "support" project file? We currently build zlib support into posix helper and that is always a DLL, so we should share the zlib support library between both runtime as well as posix helper library.

lateralusX:
@vargaz are you thinking about doing the same for all platforms (building and link our own zlib), not just for Windows?

vargaz:
At least do it on windows.

jaykrell:
I'm updating to latest zlib separately (I have a draft PR, doesn't build yet).
On most Unix platforms it is ok to depends on the dynamic version, I think.
And usually we do, I think.
But static would be ok too.

lateralusX:
OK, then let's focus on Windows first.

* Build own static version of zlib for both mingw and msvc build.
* Make sure posix library uses static build zlib library.
* Make sure runtime uses static build zlib library.
* The new VS project should use a file for sources, similar to common in other projects so we can reuse the source configuration in the build configuration of libmono-dynamic.

jaykrell:
I don't want to touch MonoPosixHelper at this time.
But yes it already has zlib in it.
Probably we should remove the code from it.

jaykrell:
Hm. There is a question as to if the symbols can/should be renamed.

I kinda think they should, but in the current rendition I have, that does change MonoPosixHelper.dll.
We can do anything there -- change one, the other, both, neither, to the same new symbols or not.
Whatever is decided.

jaykrell:
For this PR I've left the symbol names unchanged.
In another I rename them.

mono/metadata/Makefile.am
```diff
@@ -20,7 +20,25 @@ win32_sources = \
 	w32process-win32-internals.h \
 	w32socket-win32.c \
 	w32error-win32.c \
-	w32subset.h
+	w32subset.h \
```
jaykrell: This might get compiled twice, Boehm and sgen, due to how it is structured.
That could be easily fixed.

msvc/libmonoruntime.targets
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)monozlib.targets" />
```
jaykrell: Can I just inline the ItemGroup here?

lateralusX: In theory yes, but it would be good to keep it in the file, so it can be shared with other project configurations going forward. Any specific reason why you would do that?

jaykrell: Fewer files is nice. We have so many in this directory. But ok.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This should go after https://github.com/mono/mono/pull/15535.

lateralusX:
Maybe its better that #15535 is done in this PR to remove any dependency between them?

jaykrell:
The code is unused either way. I wanted to stop building it in one step, then “move” building of the live code.

jaykrell:
@monojenkins build failed

jaykrell:
Please remove label.

jaykrell:
@monojenkins build failed

jaykrell:
The other PR went in first.

jaykrell:
@monojenkins build failed

PR #15514, opened 2019-07-02
Remove minizip.

jaykrell:


jaykrell:
It turns out this is used.
Which is unfortunate but not likely to change.

PR #15522, opened 2019-07-02
[netcore] Make assembly name parsing and comparison case-insensitive

### LINKED ISSUE #11269: Type.GetType doesn't return type for full type name in upper case
## Steps to Reproduce

Type.GetType takes type full name in upper case and should ignore case.

```c#
using System;
using System.Reflection;

namespace test
{
    public class MainClass
    {
        {
            string typeName = typeof(MyNamespace1.MynAmespace2.Foo<int>).FullName.ToUpper();
            Type type = Type.GetType(typeName, throwOnError: false, ignoreCase: true);
            Console.WriteLine(type);        
        }
    }

    namespace MyNamespace1
    {
        namespace MynAmespace2
        {
            public class Foo<T> { }
        }
    }
}
```

Seems that possible reason is that `PUBLICKEYTOKEN` substring in type full name ```"TEST.MYNAMESPACE1.MYNAMESPACE2.FOO`1[[SYSTEM.INT32, MSCORLIB, VERSION=4.0.0.0, CULTURE=NEUTRAL, PUBLICKEYTOKEN=B77A5C561934E089]]";``` starts from upper `P`.
At least for `pUBLICKEYTOKEN` it works.

### Current Behavior
Empty output
### Expected Behavior
```
test.MyNamespace1.MynAmespace2.Foo`1[System.Int32]
```

[ ] macOS
[X] Linux
[X] Windows

**Version Used**:

5.16.0 or the latest master

## Stacktrace

```
```

### LINKED ISSUE #15077: [netcore] Make System.Reflection.Tests.GetTypeTests.GetType Pass
The test throws a NRE:

```
System.Reflection.Tests.GetTypeTests.GetType(typeName: "SYSTEM.REFLECTION.TESTS.MYNAMESPACE1.MYNAMESPACE3."..., expectedResult: typeof(System.Reflection.Tests.MyNamespace1.MynAmespace3.Goo<int>)) [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object
      Stack Trace:
           at System.Reflection.Tests.GetTypeTests.GetType(String typeName, Type expectedResult)
```

Determine why and make it pass.

CoffeeFlux:
Fixes #15077
Fixes #11269
Possibly fixes other things?

I've broken this into separate commits to make clearer what's going on.

The first issue present is that the assembly info was parsed case-insensitively, so setting a lowercase 'p' in PublicKeyToken would result in it not being parsed. When fixing this, the test then consistently failed whether or not the first letter was capitalized.

The second issue, somewhat hidden by the first, is that the public key token is copied directly from the input string and then compared case-sensitively in the assembly lookup. I've added two commits to deal with this, one lowercasing the PKT on parse and one making the later comparison case-insensitive. Either of them individually solves the issue, but I think it's worth doing both for good measure.

Glancing through, the parsing and comparison of the culture also would appear to be case-sensitive for everything other than 'neutral'. I'm not sure if that is intended behavior or not, but if not a separate issue should be opened.

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -428,7 +428,7 @@ encode_public_tok (const guchar *token, gint32 len)
 gboolean
 mono_public_tokens_are_equal (const unsigned char *pubt1, const unsigned char *pubt2)
 {
-	return memcmp (pubt1, pubt2, 16) == 0;
+	return g_ascii_strncasecmp ((const char*) pubt1, (const char*) pubt2, 16) == 0;
 }
 
 /**
--- a/mono/metadata/reflection.c
+++ b/mono/metadata/reflection.c
@@ -1520,7 +1520,7 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 	while (g_ascii_isspace (*p))
 		p++;
 	while (*p) {
-		if (*p == 'V' && g_ascii_strncasecmp (p, "Version=", 8) == 0) {
+		if ((*p == 'V' || *p == 'v') && g_ascii_strncasecmp (p, "Version=", 8) == 0) {
 			p += 8;
 			assembly->major = strtoul (p, &s, 10);
 			if (s == p || *s != '.')
@@ -1538,7 +1538,7 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 			if (s == p)
 				return 1;
 			p = s;
-		} else if (*p == 'C' && g_ascii_strncasecmp (p, "Culture=", 8) == 0) {
+		} else if ((*p == 'C' || *p == 'c') && g_ascii_strncasecmp (p, "Culture=", 8) == 0) {
 			p += 8;
 			if (g_ascii_strncasecmp (p, "neutral", 7) == 0) {
 				assembly->culture = "";
@@ -1549,7 +1549,7 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 					p++;
 				}
 			}
-		} else if (*p == 'P' && g_ascii_strncasecmp (p, "PublicKeyToken=", 15) == 0) {
+		} else if ((*p == 'P' || *p == 'p') && g_ascii_strncasecmp (p, "PublicKeyToken=", 15) == 0) {
 			p += 15;
 			if (strncmp (p, "null", 4) == 0) {
 				p += 4;
--- a/mono/metadata/reflection.c
+++ b/mono/metadata/reflection.c
@@ -1562,7 +1562,9 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 				len = (p - start + 1);
 				if (len > MONO_PUBLIC_KEY_TOKEN_LENGTH)
 					len = MONO_PUBLIC_KEY_TOKEN_LENGTH;
-				g_strlcpy ((char*)assembly->public_key_token, start, len);
+				char* pkt_lower = g_ascii_strdown ((char*) assembly->public_key_token, len);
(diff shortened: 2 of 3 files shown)


mono/metadata/reflection.c
```diff
@@ -1562,7 +1562,9 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 				len = (p - start + 1);
 				if (len > MONO_PUBLIC_KEY_TOKEN_LENGTH)
 					len = MONO_PUBLIC_KEY_TOKEN_LENGTH;
-				g_strlcpy ((char*)assembly->public_key_token, start, len);
+				char* pkt_lower = g_ascii_strdown ((char*) assembly->public_key_token, len);
+				g_strlcpy (pkt_lower, start, len);
```
lambdageek: I think `pkt_lower` and `start` are swapped: the arguments are dest then src.

CoffeeFlux: Oh yeah, my logic here was screwy in a couple different ways. Not sure how it worked when I tested... will fix.

mono/metadata/assembly.c
```diff
@@ -428,7 +428,7 @@ encode_public_tok (const guchar *token, gint32 len)
 gboolean
 mono_public_tokens_are_equal (const unsigned char *pubt1, const unsigned char *pubt2)
 {
-	return memcmp (pubt1, pubt2, 16) == 0;
+	return g_ascii_strncasecmp ((const char*) pubt1, (const char*) pubt2, 16) == 0;
```
jaykrell: The previous case code handled embedded nuls, and always compared all 16 bytes.
Please consider add `g_memicmp`. On Windows it will just call `_memicmp`.
And then as a bonus, you won't need the casts.

mono/metadata/reflection.c
```diff
@@ -1549,7 +1549,7 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 					p++;
 				}
 			}
-		} else if (*p == 'P' && g_ascii_strncasecmp (p, "PublicKeyToken=", 15) == 0) {
+		} else if ((*p == 'P' || *p == 'p') && g_ascii_strncasecmp (p, "PublicKeyToken=", 15) == 0) {
```
jaykrell: Please consider a macro:
#define G_STRING_CONSTANT_AND_LENGTH(x) (x), sizeof (x) - 1
to replace `8` and `15`.

mono/metadata/reflection.c
```diff
@@ -1562,7 +1562,9 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 				len = (p - start + 1);
 				if (len > MONO_PUBLIC_KEY_TOKEN_LENGTH)
 					len = MONO_PUBLIC_KEY_TOKEN_LENGTH;
-				g_strlcpy ((char*)assembly->public_key_token, start, len);
+				char* pkt_lower = g_ascii_strdown (start, len);
+				g_strlcpy ((char*) assembly->public_key_token, pkt_lower, len);
+				g_free (pkt_lower);
```
jaykrell: Please consider `g_ascii_strdown_copy` or something to avoid the extra `malloc` / `free`.

mono/metadata/assembly.c
jaykrell: Well, hey, that's weird/good, g_ascii_strncasecmp doesn't stop at a terminal nul. So I guess nevermind.

jaykrell: Oh, `g_ascii_strncasecmp` does not stop at nuls I guess so somewhat nevermind.

lambdageek:
@monojenkins build failed

lambdageek:
@monojenkins squash

PR #15523, opened 2019-07-02
Print two bytes at a time in mono_pe_file_map conversion error

CoffeeFlux:
I didn't think this through with the initial attempt. Since this is UTF-16 the string comes in pairs of two-bytes, with one of them often being null when converted up from ASCII. This should print the string properly.

### DIFF
--- a/mono/utils/mono-proclib.c
+++ b/mono/utils/mono-proclib.c
@@ -1032,7 +1032,9 @@ mono_pe_file_map (gunichar2 *filename, gint32 *map_size, void **handle)
 		GString *raw_bytes = g_string_new (NULL);
 		char *p = (char*)filename;
 		while (*p != 0) {
-			g_string_append_printf (raw_bytes, "%02X ", (int)*p);
+			g_string_append_printf (raw_bytes, "%02X ", (uint)*p);
+			p++;
+			g_string_append_printf (raw_bytes, "%02X ", (uint)*p);
 			p++;
 		}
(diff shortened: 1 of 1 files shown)


mono/utils/mono-proclib.c
```diff
@@ -1032,7 +1032,9 @@ mono_pe_file_map (gunichar2 *filename, gint32 *map_size, void **handle)
 		GString *raw_bytes = g_string_new (NULL);
 		char *p = (char*)filename;
 		while (*p != 0) {
-			g_string_append_printf (raw_bytes, "%02X ", (int)*p);
+			g_string_append_printf (raw_bytes, "%02X ", (uint)*p);
+			p++;
+			g_string_append_printf (raw_bytes, "%02X ", (uint)*p);
```
jaykrell: This is confusing. The print is the same, except for the terminal condition.
I suggest maybe instead:
```
		const gunichar2 *p = (gunichar2*)filename; 
		while (*p)
			g_string_append_printf (raw_bytes, "%04X", *p++);
```
unless you really need the chars split into two bytes each.

jaykrell: Or unless you really want to end on a single zero instead of a double zero.
It seems wrong actually, like it'll terminate early, your version, if it happens to get a character with the low byte 0, which, granted, is rare.

lambdageek: Good point, let's iterate over `guinchar2`

lambdageek:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins backport 2019-06

PR #15524, opened 2019-07-02
Convert dl-test-assets.py to Python 3

CoffeeFlux:
Not sure if this will cause issues with the bots, but Python 2 is unsupported in a few months so this is going to have to happen sooner or later.


EgorBo:
@CoffeeFlux CI fails with `ImportError: No module named request`

vargaz:
@monojenkins build failed

EgorBo:
this broke netcore CI lane, e.g. see https://github.com/mono/mono/pull/15603

PR #15527, opened 2019-07-02
[loader] Add MonoAssemblyLoadContext; stop sharing MonoImages on netcore

lambdageek:
Implement a couple of things:

1. Add a `MonoAssemblyLoadContext` opaque struct
2. On netcore Mono, use it to keep track of  which `MonoImage`s are loaded.  (Addresses part of #13891) 

In a bit more detail:

- Add a `MonoLoadedImages` struct to hold the hash tables that we use for checking if a `MonoImage` is already loaded
- Add an opaque `MonoAssemblyLoadContext` typedef
- (Framework Mono) Add a `get_global_loaded_images()` function to get the shared `MonoLoadedImages`
- (netcore Mono) Add a `MonoAssemblyLoadContext` struct that owns a `MonoLoadedImages` and add a `default_alc` member to `MonoDomain
- Pass a `MonoAssemblyLoadContext*` to most `MonoImage`-loading functions - on Framework Mono its usually NULL and we fall back to the global loaded images, on netcore Mono it's usually `domain->default_alc->loaded_images`
- (netcore Mono) Create a default Assembly Load Context when a `MonoDomain` is created
- Add a `MonoAssemblyLoadContext*` field to `MonoAssemblyLoadRequest` and pass an ALC in a few places.

---

Remaining work:

(netcore Mono) There are still places where a `MonoAssemblyLoadRequest` or a `MonoAssemblyOpenRequest` has a `NULL` ALC - those will need to be tracked down and fixed.

(netcore Mono) Eventually `MonoAssemblyLoadContext` should keep track of the loaded `MonoAssembly`s, not just images.  One goal on netcore Mono is to ifdef out `MonoDomain:domain_assemblies`.

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -2495,7 +2495,8 @@ ves_icall_System_AppDomain_LoadAssemblyRaw (MonoAppDomainHandle ad,
 	mono_gchandle_free_internal (gchandle); /* unpin */
 	MONO_HANDLE_ASSIGN (raw_assembly, NULL_HANDLE); /* don't reference the data anymore */
 	
-	MonoImage *image = mono_image_open_from_data_internal (assembly_data, raw_assembly_len, FALSE, NULL, refonly, FALSE, NULL);
+	MonoAssemblyLoadContext *alc = mono_domain_default_alc (domain);
+	MonoImage *image = mono_image_open_from_data_internal (alc, assembly_data, raw_assembly_len, FALSE, NULL, refonly, FALSE, NULL);
 
 	if (!image) {
 		mono_error_set_bad_image_by_name (error, "In memory assembly", "0x%p", raw_data);
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3242,27 +3242,3 @@ mono_runtime_install_appctx_properties (void)
 
 #endif
 
-#ifdef ENABLE_NETCORE
-gpointer
-ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalInitializeNativeALC (gpointer this_gchandle_ptr, MonoBoolean is_default_alc, MonoBoolean collectible, MonoError *error)
-{
-	/* If the ALC is collectible, this_gchandle is weak, otherwise it's strong. */
-	uint32_t this_gchandle = (uint32_t)GPOINTER_TO_UINT (this_gchandle_ptr);
-	if (collectible) {
-		mono_error_set_execution_engine (error, "Collectible AssemblyLoadContexts are not yet supported by MonoVM");
-		return NULL;
-	}
-
-	MonoDomain *domain = mono_domain_get ();
-	MonoAssemblyLoadContext *alc = NULL;
-
-	if (is_default_alc) {
-		alc = mono_domain_default_alc (domain);
-		g_assert (alc);
-	} else {
-		/* create it */
-		alc = mono_domain_create_individual_alc (domain, this_gchandle, collectible, error);
-	}
-	return alc;
-}
-#endif
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -2243,6 +2243,7 @@ mono_domain_assembly_preload (MonoAssemblyLoadContext *alc,
 	MonoDomain *domain = mono_domain_get ();
 	MonoAssembly *result = NULL;
 #ifdef ENABLE_NETCORE
+	g_assert (alc);
 	g_assert (mono_alc_domain (alc) == domain); /* should be true everywhere, not just on netcore */
 #endif
 
@@ -2366,8 +2367,13 @@ ves_icall_System_Reflection_Assembly_InternalLoad (MonoStringHandle name_handle,
 	gboolean parsed;
 	char *name;
 
+	MonoAssemblyLoadContext *alc = (MonoAssemblyLoadContext *)load_Context;
+	if (!alc)
+		alc = mono_domain_default_alc (domain);
+	g_assert (alc);
 	asmctx = MONO_ASMCTX_DEFAULT;
 	mono_assembly_request_prepare (&req.request, sizeof (req), asmctx);
+	req.request.alc = alc;
 	req.basedir = NULL;
 	req.no_postload_search = TRUE;
 
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -103,9 +103,12 @@ static const char *
 mono_check_corlib_version_internal (void);
 
 static MonoAssembly *
-mono_domain_assembly_preload (MonoAssemblyName *aname,
+mono_domain_assembly_preload (MonoAssemblyLoadContext *alc,
+			      MonoAssemblyName *aname,
 			      gchar **assemblies_path,
-			      gpointer user_data);
+			      gboolean refonly,
+			      gpointer user_data,
+			      MonoError *error);
 
 static MonoAssembly *
 mono_domain_assembly_search (MonoAssemblyName *aname,
@@ -294,8 +297,8 @@ mono_runtime_init_checked (MonoDomain *domain, MonoThreadStartCB start_cb, MonoT
 	mono_marshal_init ();
 	mono_gc_init_icalls ();
 
-	mono_install_assembly_preload_hook (mono_domain_assembly_preload, GUINT_TO_POINTER (FALSE));
-	mono_install_assembly_refonly_preload_hook (mono_domain_assembly_preload, GUINT_TO_POINTER (TRUE));
+	mono_install_assembly_preload_hook_v2 (mono_domain_assembly_preload, GUINT_TO_POINTER (FALSE), FALSE);
+	mono_install_assembly_preload_hook_v2 (mono_domain_assembly_preload, GUINT_TO_POINTER (TRUE), TRUE);
 	mono_install_assembly_search_hook (mono_domain_assembly_search, GUINT_TO_POINTER (FALSE));
 	mono_install_assembly_refonly_search_hook (mono_domain_assembly_search, GUINT_TO_POINTER (TRUE));
 	mono_install_assembly_postload_search_hook ((MonoAssemblySearchFunc)mono_domain_assembly_postload_search, GUINT_TO_POINTER (FALSE));
@@ -2230,13 +2233,18 @@ real_load (gchar **search_path, const gchar *culture, const gchar *name, const M
  * might hold the loader lock. Thus, this function must not acquire the domain lock.
  */
 static MonoAssembly *
-mono_domain_assembly_preload (MonoAssemblyName *aname,
+mono_domain_assembly_preload (MonoAssemblyLoadContext *alc,
+			      MonoAssemblyName *aname,
 			      gchar **assemblies_path,
-			      gpointer user_data)
+			      gboolean refonly,
+			      gpointer user_data,
+			      MonoError *error)
 {
 	MonoDomain *domain = mono_domain_get ();
 	MonoAssembly *result = NULL;
-	gboolean refonly = GPOINTER_TO_UINT (user_data);
+#ifdef ENABLE_NETCORE
+	g_assert (mono_alc_domain (alc) == domain); /* should be true everywhere, not just on netcore */
+#endif
 
 	set_domain_search_path (domain);
 
@@ -2250,6 +2258,7 @@ mono_domain_assembly_preload (MonoAssemblyName *aname,
 #endif
 	MonoAssemblyOpenRequest req;
 	mono_assembly_request_prepare (&req.request, sizeof (req), refonly ? MONO_ASMCTX_REFONLY : MONO_ASMCTX_DEFAULT);
+	req.request.alc = alc;
 	req.request.predicate = predicate;
 	req.request.predicate_ud = predicate_ud;
 
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3235,3 +3235,28 @@ mono_runtime_install_appctx_properties (void)
 }
 
 #endif
+
+#ifdef ENABLE_NETCORE
+gpointer
+ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalInitializeNativeALC (gpointer this_gchandle_ptr, MonoBoolean is_default_alc, MonoBoolean collectible, MonoError *error)
+{
+	/* If the ALC is collectible, this_gchandle is weak, otherwise it's strong. */
+	uint32_t this_gchandle = (uint32_t)GPOINTER_TO_UINT (this_gchandle_ptr);
+	if (collectible) {
+		mono_error_set_execution_engine (error, "Collectible AssemblyLoadContexts are not yet supported by MonoVM");
+		return NULL;
+	}
+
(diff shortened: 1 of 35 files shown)


lambdageek:
For now, just running this to get some test results to fixup anything I missed

lambdageek:
Ok, I think this PR is big enough.   It's passing the netcore tests locally, so I think it's ready for review.

lambdageek:
netmodules failure is related.

vargaz:
Would it be possible to remove the caching etc. from image.c and do it in higher levels ?

lambdageek:
> Would it be possible to remove the caching etc. from image.c and do it in higher levels ?

We can't get rid of it entirely - if two threads open the same image at the same time in the same `AssemblyLoadContext`, we want to have only one `MonoImage` created since it's one of the keys to `MonoClass` identity.

Also at least on framework Mono getting rid of loaded images caching would break some embedder APIs.

vargaz:
I mean moving the caching to the loader context level, and not doing it anymore in image.c.

mono/metadata/dynamic-image.c
```diff
@@ -357,11 +357,13 @@ mono_dynamic_image_create (MonoDynamicAssembly *assembly, char *assembly_name, c
 	image->image.references = g_new0 (MonoAssembly*, 1);
 	image->image.references [0] = NULL;
 
+#ifdef ENABLE_NETCORE
 	/* HACK: only want to crerate modules_loaded_images in the initial image for this dynamic assembly, not in the added modules. */
 	if (!assembly->assembly.image) {
```
vargaz: Typo.

lambdageek:
In the `mono-coreonly-CI` lane:
* [`GetTypeByName_NoSuchType_ThrowsTypeLoadException`](https://github.com/dotnet/corefx/blob/b8a7f34e1b64e5c594f591aa8683f34506eb9736/src/System.Runtime/tests/System/Type/TypeTests.cs#L331-L343) - did not expect this one to fail
* [`GetTypeByNameCaseSensitiveTypeloadFailure`](https://github.com/dotnet/corefx/blob/b8a7f34e1b64e5c594f591aa8683f34506eb9736/src/System.Runtime/tests/System/Type/TypeTests.cs#L345-L373) - did not expect this one to fail
* 15 other tests like [`MemberInfoGeneric_IsCollectibleTrue_WhenUsingAssemblyLoadContext`](https://github.com/dotnet/corefx/blob/30ca4113ed07fc78060a0c9d3f95eee4fe8f8ee3/src/System.Runtime/tests/System/Reflection/IsCollectibleTests.cs#L202-L233) - these are failing because [`TestAssemblyLoadContext`](https://github.com/dotnet/corefx/blob/30ca4113ed07fc78060a0c9d3f95eee4fe8f8ee3/src/System.Runtime/tests/System/Reflection/IsCollectibleTests.cs#L15-L19) is collectible and MonoVM now throws an ExecutionEngineException when you try to create a collectible assembly.

lambdageek:
The two suspicious failing tests are failing because we don't fire an `AssemblyLoadContext.AssemblyResolve` event yet.  This is expected, for now.

The tests also fail with `master`.

lambdageek:
@monojenkins build failed

lambdageek:
Tried to do what @vargaz suggested: moved image hashing out of `image.c` to a few separate source files.

lambdageek:
The msvc build is broken for coreonly - how do I conditionally include source files with our msvc files? I need `assembly-load-context.c` and `loaded-images-netcore.c` for the netcore build, and `loaded-images-global.c` for the regular Mono msvc build.

vargaz:
Its easier to add both and #ifdef out the whole file.

jaykrell:
The preprocessor is the portable well known flexible build system, agreed.

lateralusX:
FYI, there is also an option to do conditional includes of files in the msvc build, like what's being done in msvc/libmini-llvm.targets, so based on properties passed to project, you can exclude/include source files. For configuration currently picked up from configure.h (not passed as properties to msbuild) it is a little more complicated (I have plan to simplify that going forward) so for this example, doing #ifdef is probably the easiest way to go and we already do that for a lot of other source files in the build.

mono/metadata/appdomain.c
```diff
@@ -3192,3 +3241,28 @@ mono_runtime_install_appctx_properties (void)
 }
 
 #endif
+
+#ifdef ENABLE_NETCORE
```
marek-safar: Would be better to have this in a separate file

lambdageek:
@monojenkins build failed

mono/metadata/loaded-images.c
```diff
@@ -0,0 +1,101 @@
...
+
+#include "mono/metadata/loaded-images-internals.h"
+#include "mono/metadata/metadata-internals.h"
+#include "mono/utils/mono-logger-internals.h"
+
+void
+mono_loaded_images_init (MonoLoadedImages *li, MonoAssemblyLoadContext *owner)
+{
+	li->owner = owner;
+	for(int hash_idx = 0; hash_idx < MONO_LOADED_IMAGES_HASH_COUNT; hash_idx++)
```
CoffeeFlux: Style nit: shouldn't we have a space before the parenthesis?

mono/metadata/loaded-images.c
```diff
@@ -0,0 +1,101 @@
...
+		MonoImage *image;
+
+		// If an assembly image is still loaded at shutdown, this could indicate managed code is still running.
+		// Reflection-only images being still loaded doesn't indicate anything as harmful, so we don't check for it.
+		g_hash_table_iter_init (&iter, mono_loaded_images_get_hash (li, FALSE));
+		while (g_hash_table_iter_next (&iter, NULL, (void**)&image))
+			mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_ASSEMBLY, "Assembly image '%s' [%p] still loaded at shutdown.", image->name, image);
+	}
+
+	for(int hash_idx = 0; hash_idx < MONO_LOADED_IMAGES_HASH_COUNT; hash_idx++) {
```
CoffeeFlux: Same as earlier, nit with the space

mono/metadata/image.c
```diff
@@ -1723,21 +1722,41 @@ find_by_guid (gpointer key, gpointer val, gpointer user_data)
...
 MonoImage *
 mono_image_loaded_by_guid_full (const char *guid, gboolean refonly)
 {
+	return mono_image_loaded_by_guid_internal (guid, refonly);
+}
+
+/**
+ * mono_image_loaded_by_guid_internal:
+ *
+ * Do not use.  Looks only in the global loaded images hash, will miss Assembly
```
CoffeeFlux: Shouldn't this comment be somewhere that it'll show up in the documentation? The same caveat applies to anyone using the external version.

mono/metadata/image.c
```diff
@@ -719,6 +700,13 @@ mono_image_load_module_checked (MonoImage *image, int idx, MonoError *error)
 	if (image->modules_loaded [idx - 1])
 		return image->modules [idx - 1];
 
+#ifdef ENABLE_NETCORE
+	/* SRE still uses image->modules, but they are not loaded from files, so the rest of this function is dead code for netcore */
+	g_assert_not_reached ();
+#else
+
+	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_ASSEMBLY, "Loading module %d of %s (%s)", idx, image->assembly ? image->assembly->aname.name : "some assembly", image->name);
```
CoffeeFlux: What's the scenario where the assembly doesn't exist?

mono/metadata/loaded-images-internals.h
```diff
@@ -0,0 +1,67 @@
...
+enum {
+	MONO_LOADED_IMAGES_HASH_PATH = 0,
+	MONO_LOADED_IMAGES_HASH_PATH_REFONLY = 1,
+	MONO_LOADED_IMAGES_HASH_NAME = 2,
+	MONO_LOADED_IMAGES_HASH_NAME_REFONLY = 3,
+	MONO_LOADED_IMAGES_HASH_COUNT = 4
+};
+
+struct _MonoLoadedImages {
+	MonoAssemblyLoadContext *owner; /* NULL if global */
```
CoffeeFlux: To make sure I'm clear on this, it's only global on legacy, right?

msvc/libmonoruntime-common.targets
```diff
@@ -54,6 +55,10 @@
     <ClInclude Include="$(MonoSourceLocation)\mono\metadata\image-internals.h" />
     <ClInclude Include="$(MonoSourceLocation)\mono\metadata\jit-icall-reg.h" />
     <ClCompile Include="$(MonoSourceLocation)\mono\metadata\jit-info.c" />
+    <ClCompile Include="$(MonoSourceLocation)\mono\metadata\loaded-images.c" />
+    <ClInclude Include="$(MonoSourceLocation)\mono\metadata\loaded-images.h" />
+    <ClCompile Include="$(MonoSourceLocation)\mono\metadata\loaded-images-global.c" />
+    <ClCompile Include="$(MonoSourceLocation)\mono\metadata\loaded-images-netcore.c" />
```
CoffeeFlux: I think you're supposed to add all this to the .filters file too.

mono/metadata/loaded-images-internals.h
lambdageek: Yes.

mono/metadata/image.c
lambdageek: Someone using the embedding API directly to only open images, not assemblies.

mono/metadata/loaded-images.c
```diff
@@ -0,0 +1,101 @@
...
+loaded_images_get_owner (MonoImage *image)
+{
+	/* image->alc could be NULL if we're closing an image that wasn't
+	 * registered yet (for example if two threads raced to open it and one
+	 * of them lost) */
+	MonoAssemblyLoadContext *alc = mono_image_get_alc (image);
+	return mono_alc_get_loaded_images (alc);
+}
+
+/* LOCKING: assumes the images lock is locked */
```
CoffeeFlux: I'm not sure I follow why the caller has to take the lock instead of doing it in this function. The only other thing it seems to be doing in the call site is modifying the refcount of the image, and that's even atomic.

mono/metadata/appdomain.c
```diff
@@ -103,9 +103,12 @@ static const char *
...
 
 static MonoAssembly *
-mono_domain_assembly_preload (MonoAssemblyName *aname,
+mono_domain_assembly_preload (MonoAssemblyLoadContext *alc,
+			      MonoAssemblyName *aname,
 			      gchar **assemblies_path,
-			      gpointer user_data);
+			      gboolean refonly,
+			      gpointer user_data,
+			      MonoError *error);
```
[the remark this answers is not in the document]
lambdageek: That's correct. the v2 callbacks have a MonoError argument.  But this particular callback doesn't have anything to report.

mono/metadata/appdomain.c
```diff
@@ -2230,13 +2233,19 @@ real_load (gchar **search_path, const gchar *culture, const gchar *name, const M
...
+			      gboolean refonly,
+			      gpointer user_data,
+			      MonoError *error)
 {
 	MonoDomain *domain = mono_domain_get ();
 	MonoAssembly *result = NULL;
-	gboolean refonly = GPOINTER_TO_UINT (user_data);
+#ifdef ENABLE_NETCORE
+	g_assert (alc);
+	g_assert (mono_alc_domain (alc) == domain); /* should be true everywhere, not just on netcore */
```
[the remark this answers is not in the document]
lambdageek: That's an outdated comment. I'll fix it.

(On non-netcore `alc == NULL`, so `mono_alc_domain (alc)` doesn't make sense)

mono/metadata/appdomain.c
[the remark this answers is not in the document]
CoffeeFlux: Oooh okay, I see.

mono/metadata/appdomain.c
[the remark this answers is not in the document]
CoffeeFlux: Right. Sounds good.

mono/metadata/loaded-images.c
lambdageek: The issue is we need multiple things to happen together without another thread racing with us: 1. decrement the refcount down to zero, 2. remove the MonoImage from the hash tables.

The issue is that if another thread opens the same image after we decrement the refcount to == 0 and we continue to try to close it, the other thread will find the image in the hash and increment the refcount to > 0 but we will continue with unloading it because we think the refcount is == 0.

Another way to solve it would be to take the images lock here in this function and then check the refcount again, but that's not how it used to work, and I mostly just moved the code from `image.c` to this file (and changed it to use `MonoLoadedImages`)

CoffeeFlux: Gotcha. The alternative sounds preferable to me, mostly because I think it's worth minimizing functions that assume locks are already held (particularly non-static ones), but I'm fine either way. If you do keep it as-is, it would be nice if the comment was expanded a bit to justify the caller having to take the lock.

lambdageek: I changed `mono_loaded_image_remove_image` to take the lock and to do the decrement and remove together.  It actually cleans up `mono_image_close_except_pools` nicely.  👍 Thanks for the suggestion!

mono/metadata/assembly-load-context.c
```diff
@@ -0,0 +1,50 @@
...
+#include "mono/metadata/loader-internals.h"
+#include "mono/metadata/loaded-images-internals.h"
+#include "mono/utils/mono-error-internals.h"
+
+#ifdef ENABLE_NETCORE
+/* MonoAssemblyLoadContext support only in netcore Mono */
+
+void
+mono_alc_init (MonoAssemblyLoadContext *alc, MonoDomain *domain, gboolean default_alc)
+{
```
vargaz: The default_alc argument is unused.

mono/metadata/domain-internals.h
```diff
@@ -454,6 +455,12 @@ struct _MonoDomain {
 	gboolean throw_unobserved_task_exceptions;
 
 	guint32 execution_context_field_offset;
+
+#ifdef ENABLE_NETCORE
+	GSList *alcs;
```
vargaz: Why is this stored in MonoDomain, when there is only one domain in netcore ?

lambdageek: I could make it global, if you prefer.

I was hoping to also modularize non-netcore Mono's loader using ALCs in some follow-up work.  (They wouldn't have managed counterparts, just internal to the runtime).  Would be less of a hack for `LoadFile`, `LoadFrom` etc.

lambdageek: Oh, one other thing: it makes it easier to catch lifecycle issues.

The default ALC gets created when the root domain is created, if you don't have a domain yet, you can't get the default ALC either.

And if you're using the embedding API and try to get the default ALC from a thread that's not attached, you can't do it.

vargaz:
I think this PR is already pretty big, we should merge it, and make further changes later.

CoffeeFlux:
I'm planning to make another pass on it today and then I'll approve after whatever final feedback I have is addressed. If you could do similarly that would be great, and then we can get it merged and work from there.

lambdageek:
@monojenkins build failed

PR #15530, opened 2019-07-03
Cleanup and augment string handling based on https://github.com/mono/mono/pull/15522.

jaykrell:
1. When comparing strings, check pointer equality.
Faster when they match, slower when they do not.

2. Provide string lowercasing into a provided buffer, which can equal
the input, to avoid extra malloc/free.

3. Provide string compare case insensitive that accepts an exact length,
not a length that will be dominated by a nul.
On Windows this just calls _memicmp.
Arguably it should accept two lengths.
Caller could compare those, but such inlining at caller gets annoying.
Will revisit that as needed.

4. define G_STRING_CONSTANT_AND_LENGTH(x) (x), G_N_ELEMENTS (x) - 1
e.g. strncmp (foo, G_STRING_CONSTANT_AND_LENGTH ("version"))

### DIFF
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -415,6 +415,32 @@ gint    g_ascii_xdigit_value (gchar c);
 gboolean g_utf16_ascii_equal (const gunichar2 *utf16, size_t ulen, const char *ascii, size_t alen);
 gboolean g_utf16_asciiz_equal (const gunichar2 *utf16, const char *ascii);
 
+static inline
+gboolean g_ascii_equal (const char *s1, gsize len1, const char *s2, gsize len2)
+{
+    return len1 == len2 && (s1 == s2 || memcmp (s1, s2, len1) == 0);
+}
+
+static inline
+gboolean g_asciiz_equal (const char *s1, const char *s2)
+{
+    return s1 == s2 || strcmp (s1, s2) == 0;
+}
+
+static inline
+gboolean
+g_ascii_equal_caseinsensitive (const char *s1, gsize len1, const char *s2, gsize len2)
+{
+    return len1 == len2 && (s1 == s2 || g_ascii_strncasecmp (s1, s2, len1) == 0);
+}
+
+static inline
+gboolean
+g_asciiz_equal_caseinsensitive (const char *s1, const char *s2)
+{
+    return s1 == s2 || g_ascii_strcasecmp (s1, s2) == 0;
+}
+
 /* FIXME: g_strcasecmp supports utf8 unicode stuff */
 #ifdef _MSC_VER
 #define g_strcasecmp _stricmp
--- a/mono/eglib/glib.h
+++ b/mono/eglib/glib.h
@@ -212,6 +212,9 @@ typedef guint32 gunichar;
  */
 #define G_N_ELEMENTS(s)      (sizeof(s) / sizeof ((s) [0]))
 
+// e.g. strncmp (foo, G_STRING_CONSTANT_AND_LENGTH ("version"))
+#define G_STRING_CONSTANT_AND_LENGTH(x) (x), G_N_ELEMENTS (x) - 1
+
 #define FALSE                0
 #define TRUE                 1
 
@@ -400,6 +403,7 @@ gchar  *g_stpcpy             (gchar *dest, const char *src);
 gchar   g_ascii_tolower      (gchar c);
 gchar   g_ascii_toupper      (gchar c);
 gchar  *g_ascii_strdown      (const gchar *str, gssize len);
+void    g_ascii_strdown_no_alloc (char* dst, const char* src, gsize len);
 gchar  *g_ascii_strup        (const gchar *str, gssize len);
 gint    g_ascii_strncasecmp  (const gchar *s1, const gchar *s2, gsize n);
 gint    g_ascii_strcasecmp   (const gchar *s1, const gchar *s2);
--- a/mono/eglib/gstr.c
+++ b/mono/eglib/gstr.c
@@ -821,11 +821,20 @@ g_ascii_tolower (gchar c)
 	return c >= 'A' && c <= 'Z' ? c + ('a' - 'A') : c;
 }
 
+void
+g_ascii_strdown_no_alloc (char* dst, const char* src, gsize len)
+{
+	// dst can equal src. no_alloc means this function does no
+	// allocation; caller may very well.
+
+	for (gsize i = 0; i < len; ++i)
+		dst [i] = g_ascii_tolower (src [i]);
+}
+
 gchar *
 g_ascii_strdown (const gchar *str, gssize len)
 {
 	char *ret;
-	int i;
 	
 	g_return_val_if_fail  (str != NULL, NULL);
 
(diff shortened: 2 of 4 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/eglib/glib.h
```diff
@@ -399,6 +402,7 @@ gchar  *g_stpcpy             (gchar *dest, const char *src);
 gchar   g_ascii_tolower      (gchar c);
 gchar   g_ascii_toupper      (gchar c);
 gchar  *g_ascii_strdown      (const gchar *str, gssize len);
+void    g_ascii_strdown_no_alloc (char* dst, const char* src, gsize len);
```
CoffeeFlux: Is there a reason you're using `char` instead of `gchar` here? Also style nit: should be `char *dst`.

mono/eglib/gstr.c
```diff
@@ -803,21 +803,29 @@ g_ascii_tolower (gchar c)
 	return c >= 'A' && c <= 'Z' ? c + ('a' - 'A') : c;
 }
 
+void
+g_ascii_strdown_no_alloc (char* dst, const char* src, gsize len)
```
CoffeeFlux: Same thoughts as with the definition.

mono/eglib/gstr.c
```diff
@@ -869,21 +901,22 @@ g_ascii_strncasecmp (const gchar *s1, const gchar *s2, gsize n)
...
-	while (*sp1 != '\0') {
-		char c1 = g_ascii_tolower (*sp1++);
-		char c2 = g_ascii_tolower (*sp2++);
-		
-		if (c1 != c2)
-			return c1 - c2;
+
+	char c1;
+
+	while ((c1 = *s1)) {
```
CoffeeFlux: Style nit: are the double parens doing anything here?

mono/eglib/gstr.c
```diff
@@ -841,26 +849,50 @@ g_ascii_strup (const gchar *str, gssize len)
...
+
 	g_return_val_if_fail (s1 != NULL, 0);
 	g_return_val_if_fail (s2 != NULL, 0);
 
 	for (i = 0; i < n; i++) {
-		gchar c1 = g_ascii_tolower (*s1++);
-		gchar c2 = g_ascii_tolower (*s2++);
-		
-		if (c1 != c2)
-			return c1 - c2;
```
CoffeeFlux: Did anything rely on the subtraction? Should check the callers. This is a significant change to function behavior, though I think it's a reasonable one.

mono/eglib/glib.h
jaykrell: There is no point to `gchar`. Prefer `char`.

mono/eglib/gstr.c
jaykrell: This is the idiom to quash gcc warnings about assignment within conditional expressions.
Unfortunately Visual C++ does not understand it.

mono/eglib/gstr.c
jaykrell: Highly unlikely.
The usual idiom is <0, 0, or >0 and the actual values <0 and >0 are not important.

jaykrell: Subtract is often-not-always incorrect due to overflow and hopefully nobody depended on incorrect results there.

mono/eglib/glib.h
jaykrell: `char* dst` vs. `char *dst` is debatable.
The second tends to be preferred by people who want to highlight C's backwardness.
The first is for people who want it to look sane despite the underlying craziness.
C's declaration style is a failed experiment in coming up with grammars that require right to left parsing instead of the intuitive left to right.

mono/eglib/gstr.c
CoffeeFlux: There are tests that assume the behavior and should be updated, though I don't see anything else (and am unsure of whether or not these tests are ever actually run). https://github.com/mono/mono/blob/ab69814598fbdc5fb66c59a0709eda2a61c3511f/mono/eglib/test/string-util.c#L627

mono/eglib/glib.h
CoffeeFlux: Yes, but most of the file (and the runtime as a whole) uses the second option. I personally prefer the first but consistency is what's important here.

CoffeeFlux: As far as the `gchar` thing, I was going to say it's for consistency there as well but looking at the header we seem to be all over the place for no good reason. Is there any logic to it? cc @lambdageek

jaykrell: Maybe sometime styles will be revisited and a better consistency will be established instead of being stuck with old less favored styles forever. :(

mono/eglib/gstr.c
jaykrell: Wow that is a bad test. I've never seen anyone depend on other than <0, 0, >0 for "compare" for `strcmp`, `strncasecmp`, `_stricmp`, and that is all they guarantee. The actual values are not guaranteed. (See
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp-mbsicmp-stricmp-l-wcsicmp-l-mbsicmp-l?view=vs-2019, http://pubs.opengroup.org/onlinepubs/009695399/functions/strncmp.html, http://pubs.opengroup.org/onlinepubs/009695399/functions/strcmp.html http://pubs.opengroup.org/onlinepubs/009695399/functions/strcasecmp.html
etc.).

This is a custom mono routine so it could have a more "precise" but bad interface, but it would no longer resemble what it is based on and people could not switch back and forth.

There could be some sort of tie-breaking when chars vary only in case, but I don't think the old code worked that way. And things like qsort/bsearch would not notice.

I have seen the overflow bug repeatedly, for example comparing a GUID a ULONG at a time. The bug only shows up when all the bits are actually used. For example if the text was strictly ASCII and not UTF8, it'd be ok. But this seems like too much trouble to understand and code, vs. doing simple compares. And while it says "ascii", we'd have to look more closely to rule out UTF8 or other. I think "ASCII" might be more the algorithm and not the data.

jaykrell: I believe the glib tests are not run. I made some improvements there, i.e. fixed them to be easily built, but stopped there.

CoffeeFlux: Yeah, the test definitely isn't great. I guess if they're not being run it doesn't matter too much, though in lieu of a fixed test a note would be nice.

jaykrell: The test should be fixed with this PR. I hadn't noticed it.
People should occasionaly run these tests. I didn't. And confirm it fails.
We should probably enable the tests for real too.

mono/eglib/glib.h
CoffeeFlux: Okay I think char is fine, so just the minor stylistic nit with the asterisk

jaykrell:
The starts seem dumb when they appear as:

`a = (char *)b`

or

```
char *
foo (...)
```


`char *a, *b` makes them make sense, but most style guides legislate that away and require:

```
char* a;
char* b;
```

which seems to be Stroustrup's preference, if it matters.

CoffeeFlux:
@jaykrell any chance you could update this soonish? I can if you don't have time, but I'm holding up another PR for it that I'd like to land.

jaykrell:
Sorry, yes, I kinda it backwards in my head.
The test isn't run so maybe punt that briefly?

jaykrell:
Either way, tonight or this weekend.

CoffeeFlux:
@monojenkins build failed

jaykrell:
Yeah I forgot too. Rebased. Thank you.

jaykrell:
@monojenkins build failed

PR #15534, opened 2019-07-03
Don't clobber last error in mono_set_lmf

filipnavara:
Still not sure about the best layer where to place it. I can sprinkle it around `interp_pop_lmf` calls in `interp.c` and hope for the best or put it on this one place.

### DIFF
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -757,7 +757,14 @@ mono_get_lmf (void)
 void
 mono_set_lmf (MonoLMF *lmf)
 {
+	/*
+	 * TLS access resets the last error. This is called from several places
+	 * in the interpreter and debugger where we need to preserve the value (eg.
+	 * to record it for P/Invoke calls).
+	 */
+	W32_DEFINE_LAST_ERROR_RESTORE_POINT;
 	(*mono_get_lmf_addr ()) = lmf;
+	W32_RESTORE_LAST_ERROR_FROM_RESTORE_POINT;
 }
 
 static void
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -757,6 +757,9 @@ mono_get_lmf (void)
 void
 mono_set_lmf (MonoLMF *lmf)
 {
+#ifdef MONO_KEYWORD_THREAD
+	(*mono_get_lmf_addr ()) = lmf;
+#else
 	/*
 	 * TLS access resets the last error. This is called from several places
 	 * in the interpreter and debugger where we need to preserve the value (eg.
@@ -765,6 +768,7 @@ mono_set_lmf (MonoLMF *lmf)
 	W32_DEFINE_LAST_ERROR_RESTORE_POINT;
 	(*mono_get_lmf_addr ()) = lmf;
 	W32_RESTORE_LAST_ERROR_FROM_RESTORE_POINT;
+#endif
 }
 
 static void


mono/mini/mini-runtime.c
```diff
@@ -757,7 +757,14 @@ mono_get_lmf (void)
 void
 mono_set_lmf (MonoLMF *lmf)
 {
+	/*
+	 * TLS access resets the last error. This is called from several places
+	 * in the interpreter and debugger where we need to preserve the value (eg.
+	 * to record it for P/Invoke calls).
+	 */
+	W32_DEFINE_LAST_ERROR_RESTORE_POINT;
```
jaykrell: Only needed for MinGW.

filipnavara: That's what I thought. I checked again and apparently the code is not using `__declspec(thread)` on MSVC here either. It boils down to this:

https://github.com/mono/mono/blob/cad60fce03f1014f0a6b7794d8ca9e28afcc78aa/mono/utils/mono-tls.c#L298-L301

If that is reworked then this can be changed to avoid it on MSVC. I am trying to make a minimal fix and possibly improve upon it later.

jaykrell: winconfig.h:

```
/* Have __thread keyword */
#define MONO_KEYWORD_THREAD __declspec (thread)
```

mono-tls.c

```
#ifdef MONO_KEYWORD_THREAD
#define MONO_TLS_GET_VALUE(tls_var,tls_key) (tls_var)
#define MONO_TLS_SET_VALUE(tls_var,tls_key,value) (tls_var = value)
#else
#define MONO_TLS_GET_VALUE(tls_var,tls_key) (mono_native_tls_get_value (tls_key))
#define MONO_TLS_SET_VALUE(tls_var,tls_key,value) (mono_native_tls_set_value (tls_key, value))
#endif
```

Unless msvc is getting the MinGW config.h or something. ?

filipnavara: Ah, you are right.

jaykrell:
The fact that mono's JIT accesses `TlsSlots` and `TlsExpansionSlots` suggests another possibility.
 -> Do that in C/C++.
 -> No LastError globbering.
 -> faster, maybe (can inline it, but it isn't tiny, so then, it is the same, except for the instruction to clobber LastError)
 -> In MSVC or MinGW.

jaykrell:
Btw that code to access TlsSlots is wrong if this happens:

```
```

jaykrell:
Please take this opportunity to add `const` to `W32_DEFINE_LAST_ERROR_RESTORE_POINT.
It wasn't there before, because there were sort of but not really other assignments.

lateralusX:
> Btw that code to access TlsSlots is wrong if this happens:
> #if G_HAVE_API_SUPPORT(HAVE_UWP_WINAPI_SUPPORT)
> #if WINDOWS_SDK_BUILD_VERSION < 14393
> #define TlsAlloc() FlsAlloc(NULL)

After some close look it seems like the "old" way of access TEB in codegen has been eliminated when building under MSVC, it will be converted to icalls instead of inlined TLS access. We don't have any build confiugrations using HAVE_UWP_WINAPI_SUPPORT and it will be eliminated going forward so that should not come in play at the moment.

lateralusX:
Another option is to put it in interp_pop_lmf or just close to the pinvoke failing. We have a test trying to test failures around last error and pinvokes, mono/mono/tests/last-error.cs, it should be run under interpreter as well on CI, but CI is MSVC only so it wouldn't detect this since it was a mingw specific problem.

filipnavara:
`interp_pop_lmf` sounds like a reasonable option to me. There seems to be more than one P/Invoke path and not all of them were triggered in my tests.

vargaz:
The right place to fix this is the code which is executed right after a pinvoke call. In the interp case, its easy, with the JIT, its kinda complicated.

filipnavara:
Damn. This didn't work. It still gets clobbered elsewhere on the way.

> with the JIT, its kinda complicated.

JIT is not affected.

vargaz:
Th relevant code looks like interp.c:1383 in ves_pinvoke_method ().

jaykrell:
Here is draft optimized `GetLastError` / `SetLastError` if interested..

https://github.com/mono/mono/pull/15540

lateralusX:
Looks like it can get in using this method as well, do_icall_wrapper, MINT_CALLI_NAT_FAST.

filipnavara:
@vargaz, that was not the code path I hit. The one @lateralusX mentions is the other one that gets hit in my case.

I am a bit tired of trying to fix it. I'll probably close this since it doesn't work correctly anyway and open an issue with steps to reproduce the error.

jaykrell:
@filipnavara Did the earlier PR work, that preserved LastError for all `TlsGetValue()`?

filipnavara:
@jaykrell Yes, it did.

vargaz:
Note that SetLastError is implement by the wrapper calling the 'mono_set_last_error' function after the pinvoke call. In the interpreter, calling this function involves a lot of code which can also overrride errno.
So this method needs to be changed to something else. I'd suggest calling mono_set_last_error () manually after the pinvoke call, and somehow nullifying the later set_last_error () call made by the wrapper.

lateralusX:
Looks like interpreter does a additional run of the GetLastErrorValue call that will push and pop several of the instructions before reading out the error value, in JIT this a call just after p/invoke, but the interpreter does more related to this bigger problem around when interpreter store last error for p/invokes.

jaykrell:
Which reminds me, why is there a second last error:

```
void
mono_marshal_set_last_error_windows (int error)
{
#ifdef WIN32
	/* This icall is called just after a P/Invoke call before the P/Invoke
	 * wrapper transitions the runtime back to running mode. */
	MONO_REQ_GC_SAFE_MODE;
	mono_native_tls_set_value (last_error_tls_id, GINT_TO_POINTER (error));
#endif
}
```

Why isn't that just `SetLastError()`?

lateralusX:
Since we need to keep correct last error that can be read from marshaller, ves_icall_System_Runtime_InteropServices_Marshal_GetLastWin32Error, and at that point a lot of code has been executed.

jaykrell:
You mean, because we clobber the real Win32 LastError, so we need to preserve it elsewhere?
Ugh.

jaykrell:
Perhaps we should preserve it in a few more places. Perhaps @filipnavara earlier PR is most of that.

lateralusX:
That is not Windows specific, after p/invoke we set last error for all platforms, since by the time we run managed code that will access it, last error is potentially long gone on any platform.

jaykrell:
I understand Unix would maybe need something else, either `__thread` or `pthread_getspecific`, or perhaps `errno`. Well, this is wonky though -- because P/Invoke's set lasterror somehow on Unix?

But anyway, on Windows, it should be enough to use GetLastError, and avoid clobbering it, or preserve it a in few places. Maybe switching to `__declspec (thread)` did or will make this easier..

lateralusX:
It is implemented on all platforms, mono_marshal_set_last_error/mono_marshal_set_last_error_windows. The real fix is to do something in the interpreter around this as in JIT, making the interpreter pinvoke call implementation setting this closer then how it currently works, since it will be subject for last error clobbering by any function changing last error between return of native p/invoke and when interpreter loop has run the instructions needed to call GetLastError on Windows.

jaykrell:
What makes it long gone? Other than `TlsGetValue`?

jaykrell:
And, if it is marshaling, couldn't the wrapper save it in a local instead of a thread local, and then back to `SetLastError` at the end of unmarshaling?

And if it must be a thread local, `__declspec(thread)` should work right?

lateralusX:
As I said, this is not Windows specific and implemented on all platforms (look at mono_marshal_set_last_error, uses errono on other platforms), it needs to be preserved since it could be accessed from managed code through Marshal.GetLastWin32Error. Fix is to set it differently from interpreter so it doesn't need to run additional interpreted instructions to set it.

jaykrell:
But where do we run a lot of code? Just unmarshalling right? Instead of GetLastError and set other thread local, how about GetLastError into a regular local, run that bunch of code, and then put it back into regular SetLastError? I understand the need to capture it and save it away, but why can't its final (temporary) resting place be its original place, after running whatever code might clobber it?

Is there no clear end to the code that could clobber it, so it needs some extra special safe location?

What if I were to turn around and p/invoke a function of mine called PrintLastError?
Isn't it reasonable to expect LastError to flow unmodified from P/Invoke, to managed, and to another P/Invoke? That is, the runtime should generally preserve it, in its "normal" place, not in a "special" place?

The point is, it'd seem better to have one location than two, even if it requires the same amount of get/set. i.e. making the normal place runtime-transparent should be about the same cost as making a special place runtime-transparent.

Does CoreCLR-or-desktop also setup a separate location, or instead preserve the one?

filipnavara:
> I understand the need to capture it and save it away, but why can't its final (temporary) resting place be its original place, after running whatever code might clobber it?

For one, because CLR API is designed that way. You can specify `SetLastError` property on the `DllImport` attribute for P/Invoke calls. Only when it's set to true you want it to be reflected by `Marshal.GetLastWin32Error`. So, you either take a penalty on `SetLastError = true` calls by storing the error elsewhere and running additional code to do so, or you take the penalty for `SetLastError = false` calls and all runtime code by keeping the last error consistently in the TEB all the time. Turns out the first option is less expensive one.

jaykrell:
"all runtime code", or just one get/set around the unmarshaling?
I need to poke aroun CoreCLR I guess..

jaykrell:
Similar looking in CoreCLR:

```
FCIMPL0(void, StubHelpers::SetLastError)
{
    // Make sure this is the first thing we do after returning from the target, as almost everything can cause the last error to get trashed
    DWORD lastError = ::GetLastError();

    FCALL_CONTRACT;

    GetThread()->m_dwLastError = lastError;
}
FCIMPLEND
```

jaykrell:
I still wonder if it can't go back to the normal place soon after though, living only in a local (returned from this function).

PR #15537, opened 2019-07-03
Make sure TLS (thread local storage) is always TLS and not FLS (fiber local storage).

jaykrell:
The Windows SDKs are inconsistent (and clearly documented as so), and the JIT assumes TLS.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15538, opened 2019-07-03
Readonly codemanager callbacks.

jaykrell:
The callbacks cannot be made directly, because codeman is used by metadata, and linked to monodis, but mini is not linked to monodis. The pointers provide a "weak" linkage. That is preserved.


jaykrell:
@monojenkins build failed

PR #15539, opened 2019-07-03
Remove unused Windows JIT TLS code.

jaykrell:
It should be replaced with something different.

### DIFF
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -3781,22 +3781,8 @@ static guint8*
 mono_amd64_emit_tls_get (guint8* code, int dreg, int tls_offset)
 {
 #ifdef TARGET_WIN32
-	if (tls_offset < 64) {
-		x86_prefix (code, X86_GS_PREFIX);
-		amd64_mov_reg_mem (code, dreg, (tls_offset * 8) + 0x1480, 8);
-	} else {
-		guint8 *buf [16];
-
-		g_assert (tls_offset < 0x440);
-		/* Load TEB->TlsExpansionSlots */
-		x86_prefix (code, X86_GS_PREFIX);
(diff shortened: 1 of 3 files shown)


jaykrell:
Agreed.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This code is not dead as we thought.

https://jenkins.mono-project.com/job/xa/1233/parsed_console/log.html

/r:mscorlib.dll /out:build/deps/basic-profile-check.exe build/common/basic-profile-check.cs
* Assertion: should not be reached at /cygdrive/d/j/workspace/xa/mono/mini/mini-amd64.c:3785


I will revise in:

https://github.com/mono/mono/pull/15567

lateralusX:
OK, so should we close this one then in favor of the other PR and do change as part of that instead?

jaykrell:
Yeah I think so. The other one is looking good, but some sequence point test fails.
I don't think this could should be replace per se, but add to it.
With MSVC we should use __declspec(thread), add to this code.
With MinGW, with poor implementation of __thread, continue using TlsAlloc, and this code.

The PR is "close". Maybe look at it. Or maybe it'll change still and wait.
There is some sequence point test failure I don't understand. Should be able to repro locally at least.

Alternatively, but I don't like it, use `TlsAlloc` either way.
For JIT code, the difference is smaller. We can inline either.
For C/C++, `__declspec(thread)` is much better.
For setting, which JITed code never does, `__declspec (thread)` is also much better -- you can't inline TlsSetValue, but `__declspec (thread)` is inlined either way.

lateralusX:
Code still in use.

PR #15552, opened 2019-07-03
[ilasm] [System.Runtime]System.String wasn't generating string primitive type

### LINKED ISSUE #13923: [DIM] [CoreCLR Test] boring not working
IL Example:

https://github.com/dotnet/coreclr/blob/e95c1f8fe6ae185b90b807170445c2e26cfbed66/tests/src/Regressions/coreclr/16355/boring.il

thaystg:
Was crashing when was executing the boring.il test from coreclr, the problem was that using [System.Runtime]System.String the type was not understood as the primitive type string.

Using ilasm from coreclr and running on mono it works, because of that I decided to fix on ilasm and not on runtime.

Fixes #13923

### DIFF
--- a/mcs/ilasm/codegen/GenericArguments.cs
+++ b/mcs/ilasm/codegen/GenericArguments.cs
@@ -40,7 +40,11 @@ namespace Mono.ILASM {
 
 			if (type_list == null)
 				type_list = new ArrayList ();
-			type_list.Add (type);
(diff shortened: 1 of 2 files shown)


thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins backport 2019-06

PR #15553, opened 2019-07-03
[android] switch cross compilers host from 32bit to 64bit

lewurm:
Untested, but we need to do it anyway soon. This will help us to understand the situation on Xamarin.Android while finishing the work for Xamarin.iOS.


jaykrell:
@monojenkins build failed

lewurm:
@monojenkins backport 2019-06

PR #15559, opened 2019-07-04
Distribute sgen-dynarray.h via source not extra.

jaykrell:


jaykrell:
@monojenkins build failed

PR #15560, opened 2019-07-04
Inline GetLastError and SetLastError.

jaykrell:
Context: Would make https://github.com/mono/mono/pull/15534 cheaper (and only MinGW is affected).

### DIFF
--- a/mono/utils/mono-error-internals.h
+++ b/mono/utils/mono-error-internals.h
@@ -284,4 +284,46 @@ mono_error_set_specific (MonoError *error, int error_code, const char *missing_m
 void
 mono_error_set_first_argument (MonoError *oerror, const char *first_argument);
 
+#if HOST_WIN32
+
+// FIXME Change all GetLastError / SetLastError to these.
+// FIXME Inline function should work.
+
+#if HOST_X86
+#if __GNUC__
+#define mono_get_last_win32_error()              \
+    ({ unsigned long err;                        \
+    __asm__ ("movl %%fs:0x34, %0" : "=r" (err)); \
+    err; })
+#define mono_set_last_win32_error(err) __asm__ ("movl %0, %%fs:0x34" : : "r" (err))
+#elif _MSC_VER
+#define mono_get_last_win32_error()     (__readfsdword (0x34))
+#define mono_set_last_win32_error(err)  (__writefsdword (0x34, (err)))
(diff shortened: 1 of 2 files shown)


jaykrell:
@monojenkins build failed

mono/utils/mono-error-internals.h
```diff
@@ -284,4 +284,80 @@ mono_error_set_specific (MonoError *error, int error_code, const char *missing_m
...
+}
+
+static inline
+void
+__stdcall
+SetLastError (unsigned long err)
+{
+#if __GNUC__
+#if HOST_X86
+    __asm__ ("movl %0, %%fs:0x34" : : "r" (err));
```
jaykrell: I think gcc has the intrinsics actually, so can remove `#if`.

jaykrell: Or at least macros with this interface that generate the same assembly.

filipnavara: I think it may support it but I will rather check that it works on both x86/x64.

mono/utils/mono-error-internals.h
```diff
@@ -284,4 +284,49 @@ mono_error_set_specific (MonoError *error, int error_code, const char *missing_m
...
+unsigned long
+__stdcall
+GetLastError (void)
+{
+#if HOST_X86
+    return __readfsdword (0x34);
+#elif HOST_AMD64
+    return __readgsdword (0x68);
+#else
+#error unknown architecture
```
filipnavara: Can we just rely on the system headers for unknown architecture instead of generating errors? (Windows/ARM64)

filipnavara: Oh, now I see that is actually done. It is a bit misleading :)

jaykrell: It was probably accurate earlier. I changed the message.
Perhaps could use more commenting/clarity.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Approve?

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15562, opened 2019-07-04
Turn off incremental linking for release builds.

jaykrell:
It has a runtime cost.

Also incremental linking does not mean anything for static .libs.
  Unless maybe the consumer inherits that -- let consumer decide if so.

Also genmdesc is rarely built so turn it off there too.

Larger context is:
 1. Systems I usually work with never use incremental linking. Linking perf isn't noticeable.
Linking perf is noticeable in large systems though.
2. Unix doesn't have incremental linking?
3. Incremental linking is a bit non-deterministic as compile/link cycles gradually build up state, perturbing the output in a non-repeatable way, until eventually incremental linking gives up and you get a full link again, starting the cycle over.
4. And still, keep it for debug builds, fast iteration, which is the usual convention.
5. Clean debug builds as well remain more deterministic.

Signs of an incremental link:
 1. Large *.ilk files dropped next to .exe and .dll files (but not .libs).
 2. All function calls go through an ILT incremental link thunk.

### DIFF
--- a/msvc/genmdesc.vcxproj
+++ b/msvc/genmdesc.vcxproj
@@ -80,11 +80,9 @@
     <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(MONO_BUILD_DIR_PREFIX)$(Platform)\obj\$(ProjectName)\$(Configuration)\</IntDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <LinkIncremental>true</LinkIncremental>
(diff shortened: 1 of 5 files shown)


lateralusX:
Easier to change this in the UI, since it will be consistent with how it handles the configuration settings for different properties.

lateralusX:
The project files didn't explicitly set incremental linking in release builds, but was setup to inherit from parent project. Looks like this might behave differently between VS versions, but when building release builds on VS2015 (the version we currently support), we get build without incremental linkage for the important mono-2.0-sgen.dll, but I agree that we should be explicit about it.

PR #15567, opened 2019-07-04
[JIT] [x86] [amd64] Fast TLS for Visual C++.

jaykrell:
Supercedes https://github.com/mono/mono/pull/15539.

### DIFF
--- a/mono/mini/mini-amd64.c
+++ b/mono/mini/mini-amd64.c
@@ -3781,6 +3781,15 @@ static guint8*
 mono_amd64_emit_tls_get (guint8* code, int dreg, int tls_offset)
 {
 #ifdef TARGET_WIN32
+
+#ifdef MONO_KEYWORD_THREAD
+
+	return mono_windows_emit_tls_get (code, dreg, tls_offset);
+
+#else
+
+	// FIXME refactor x86/amd64
+
 	if (tls_offset < 64) {
 		x86_prefix (code, X86_GS_PREFIX);
 		amd64_mov_reg_mem (code, dreg, (tls_offset * 8) + 0x1480, 8);
@@ -3797,6 +3806,9 @@ mono_amd64_emit_tls_get (guint8* code, int dreg, int tls_offset)
 		amd64_mov_reg_membase (code, dreg, dreg, (tls_offset * 8) - 0x200, 8);
 		amd64_patch (buf [0], code);
 	}
+
+#endif
+
 #elif defined(TARGET_MACH)
 	x86_prefix (code, X86_GS_PREFIX);
 	amd64_mov_reg_mem (code, dreg, tls_gs_offset + (tls_offset * 8), 8);
--- a/mono/mini/mini-windows.c
+++ b/mono/mini/mini-windows.c
@@ -471,3 +471,115 @@ mono_win32_runtime_tls_callback (HMODULE module_handle, DWORD reason, LPVOID res
 	}
 	return TRUE;
 }
+
+#if HOST_WIN32 && defined (MONO_KEYWORD_THREAD)
+
+guint8*
+mono_windows_emit_tls_get (guint8* code, int dreg, int tls_offset)
+{
+	// This is documented here:
+	//
+	// https://docs.microsoft.com/en-us/windows/win32/debug/pe-format
+	//
+	// Executable code accesses a static TLS data object through the following steps:
+	//
+	// At link time, the linker sets the Address of Index field of the TLS directory.
+	//
+	// This field points to a location where the program expects to receive the TLS index.
+	//
+	// The Microsoft run-time library facilitates this process by defining a memory image
+	// of the TLS directory and giving it the special name "__tls_used" (Intel x86 platforms)
+	// or "_tls_used" (other platforms).
+	//
+	// The linker looks for this memory image and uses the data there to create the TLS directory.
+	//
+	// Other compilers that support TLS and work with the Microsoft linker must use this same technique.
+	//
+	// When a thread is created,
+	//	[jaykrell: Or when a .dll is loaded, after thread create, retroactively for existing threads.]
+	// the loader communicates the address of the thread's TLS array by
+	// placing the address of the thread environment block (TEB) in the FS register. A pointer to
+	// the TLS array is at the offset of 0x2C from the beginning of TEB. This behavior is Intel x86-specific.
+	//	[jaykrell: For Win64 it is 0x2C times 2. For AMD64 it is GS:. This is obvious from compiler output].
+	//
+	// The loader assigns the value of the TLS index to the place that was indicated by the Address of Index field.
+	//
+	// The executable code retrieves the TLS index and also the location of the TLS array.
+	//
+	// The code uses the TLS index and the TLS array location (multiplying the index by 4 [jaykrell: or 8]
+	// and using it as an offset to the array) to get the address of the TLS data area for the given program
+	// and module. Each thread has its own TLS data area, but this is transparent to the program,
+	// which does not need to know how data is allocated for individual threads.
+	//
+	// An individual TLS data object is accessed as some fixed offset into the TLS data area.
+	//
+	// The TLS array is an array of addresses that the system maintains for each thread. Each address in
+	// this array gives the location of TLS data for a given module (EXE or DLL) within the program.
+	// The TLS index indicates which member of the array to use. The index is a number (meaningful only
+	// to the system) that identifies the module.
+	//
+	// [End of documentation]
+	//
+	// The documentation is out of date on some details that do not matter.
+	//   - It immediately prior contains a caveat that was fixed in Vista.
+	//   - Exactly when the operating system does the allocation/free does not matter.
+	//     That is, this works even if a thread is created before a .dll is loaded.
+	//
+	//   - The linker/compiler interface to the operating system remains unchanged and correct.
+	//
+	// For example:
+	//
+	// __declspec (thread) int a [10];
+	//
+	// __declspec (dllexport)
+	// int f1 (void)
+	// {
+	//	return a [0] + a [1] + a [2];
+	// }
+	//
+	// cl /O2 /Zi /MD /LD 1.c /link /incremental:no
+	// link /dump /disasm 1.dll | more
+	//
+	// f1:
+	// 0000000180001000: 8B 0D 3A 30 00 00  mov         ecx,dword ptr [_tls_index] ; ecx = _tls_index
+	// 0000000180001006: 65 48 8B 04 25 58  mov         rax,qword ptr gs:[58h]     ; rax = Teb->TlsVector
+	//	00 00 00
+	// 000000018000100F: BA 08 00 00 00     mov         edx,8                      ; edx = tls_offset common subexpression (tls_offset of a)
+	// 0000000180001014: 48 8B 04 C8        mov         rax,qword ptr [rax+rcx*8]  ; rax = Teb->TlsVector [_tls_index]
+	// 0000000180001018: 8B 4C 10 08        mov         ecx,dword ptr [rax+rdx+8]  ; ecx = *(Teb->TlsVector [_tls_index] + 16)
+	// 000000018000101C: 03 4C 10 04        add         ecx,dword ptr [rax+rdx+4]  ; ecx += *(Teb->TlsVector [_tls_index] + 12)
+	// 0000000180001020: 03 0C 10           add         ecx,dword ptr [rax+rdx]    ; ecx += *(Teb->TlsVector [_tls_index] + 8)
+	// 0000000180001023: 8B C1              mov         eax,ecx
+	// 0000000180001025: C3                 ret
+	//
+	// Use dreg as the only temp.
+	// _tls_index and tls_offset can be immediates in JIT.
+	//
+	// FIXME For AOT, we could export the offsets from mono runtime
+	// and access [__imp__tls_index] and [__imp__tls_offset] (for each tls_offset)
+	//
+	// tls_offset is determined at mono link-time but indirection is still a good idea.
+	// _tls_index is determined at load time and cannot be output to AOT.
+	//
+	const guint size = sizeof (char*); // documentation says "4" here
+
+ #if TARGET_X86
+	x86_prefix (code, X86_FS_PREFIX);
+
+	// FIXME Enable x86/amd64 code sharing. The interfaces match.
+#define amd64_mov_reg_mem       x86_mov_reg_mem
+#define amd64_mov_reg_membase   x86_mov_reg_membase
+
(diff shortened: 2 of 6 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
ifdefs not quite right for mingw. Will adjust.

`Mono implements fast TLS for Visual C++ only, pending verification/improvement of other compilers.`

lateralusX:
How does this work with different versions of msvc, the same implementation on all versions (2015/2017/2019) or differences?

jaykrell:
Yes. The compiler and linker have never ever changed here.

jaykrell:
Nor the runtime.
The main change in this area is __declspec(thread) in Vista, but it wasn't a compiler/linker change.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Slight problem on x86 tests so disabling.
amd64 looking good.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

BrzVlad:
I'm not convinced that the added complexity, which relies upon reverse engineering, is worth it.

I'd like to see first how much this improves performance on a microbenchmark (let's say accessing a ThreadStatic variable in a loop) compared to fallback to calling tls access functions in mono-tls.c (I'd expect negligible). Also what difference would it make compared to the simple straightforward way of accessing these tls variables with TlsGet (even if we have __thread), which we already have implemented since a long time (and I'm assuming it was enabled and working?) (I'd expect absolutely no diference here). If that is the case, I think a better and safer approach would be to just use TlsGet for the very few tls variables accessed from managed code.

jaykrell:
This is just reading the compiler output -- hardly reverse engineering.
`__declspec(thread)` is better because of the automatic initialization and cleanup it offers, and because that is how we are using to access the variables in the runtime itself already. We wouldn't want to change them to `TlsGetValue` imho.

jaykrell:
`__thread` should be the way forward. Like C++11 has `thread_local` (though not supported on all of our targets. I tried). And can apparently be seen in CoreCLR (again, fewer targets).

The two ways imho should remain abstracted in the code and both supported as they are now, until/unless __thread / __declspec(thread) becomes widespread.

jaykrell:
Which part is complex btw? It seems pretty simple. Except there is that sequence point failure I have to figure out. :(

BrzVlad:
There is no guarantee that the compiler will always generate thread local access that way. We had code like this in the past and we ended up just removing it because maintaining it was just painful. That's why we should do a performance comparison first. Otherwise what is the point to risk breaking everything on newer releases if we don't get any performance gain.

jaykrell:
The compiler is definitely guaranteed to behave this way.

It has been this way forever for all 32bit and 64bit Windows and it is part of the PE format and runtime loader. Some new way could come along, for use only with new compilers and new operating systems, but it has been this since version 1 and it works very well (at last since Vista, less well before then, but still the same interface). It won't change.


The only tiny "problem" is the name of the variable `_tls_index`. That is an interface between this compiler and this C runtime. The actual architectural interface is you get your PE headers and find the variable the same way as the loader. But it has always been this way for Visual C++.

With this linker, you can get your own PE headers via `__ImageBase`. Other linkers might use a different name.

There is also linker-independent Win32 API to get module base from any address in it.

I think referencing `_tls_index` is good tradeoff for now, for a while.
If MinGW gets this support, and uses a different variable name, we can ifdef that or use the headers.

jaykrell:
It isn't just read the compiler output, I guess I said that wrong, sorry.
It also discussed in the documentation some:

https://docs.microsoft.com/en-us/windows/win32/debug/pe-format

Though for example they fail to give the 64bit offset.

```
Executable code accesses a static TLS data object through the following steps:

At link time, the linker sets the Address of Index field of the TLS directory. This field points to a location where the program expects to receive the TLS index.

The Microsoft run-time library facilitates this process by defining a memory image of the TLS directory and giving it the special name "__tls_used" (Intel x86 platforms) or "_tls_used" (other platforms). The linker looks for this memory image and uses the data there to create the TLS directory. Other compilers that support TLS and work with the Microsoft linker must use this same technique.

When a thread is created, the loader communicates the address of the thread's TLS array by placing the address of the thread environment block (TEB) in the FS register. A pointer to the TLS array is at the offset of 0x2C from the beginning of TEB. This behavior is Intel x86-specific.

The loader assigns the value of the TLS index to the place that was indicated by the Address of Index field.

The executable code retrieves the TLS index and also the location of the TLS array.
The code uses the TLS index and the TLS array location (multiplying the index by 4 and using it as an offset to the array) to get the address of the TLS data area for the given program and module. Each thread has its own TLS data area, but this is transparent to the program, which does not need to know how data is allocated for individual threads.

An individual TLS data object is accessed as some fixed offset into the TLS data area.
The TLS array is an array of addresses that the system maintains for each thread. Each address in this array gives the location of TLS data for a given module (EXE or DLL) within the program. The TLS index indicates which member of the array to use. The index is a number (meaningful only to the system) that identifies the module.

```

I find the compiler output easier to understand and summarize from, the documenation is so verbose, but there it is.

jaykrell:
__declspec(thread) also never trashes LastError
as TlsGetValue always does.

And for writes, TlsSetValue has to 
lock & heap allocate 
sometimes.
A hidden coop hazard.

__declspec(thread) never does.
It is in .dll load or thread create instead.

jaykrell:
@monojenkins build failed

jaykrell:
CI passed on retry. Wierd.
I added the documentation.
I'm confused about the painful maintenance, given that this has never changed.
The non-Windows version does look painful agreed.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build failed

jaykrell:
The seqpoint breaks were because run to run generates different sized code. The test wants same size. There is a flag to stabilize size but it isn’t fully implemented and implementing it breaks things. I tried fixing but gave up for & worked around.

jaykrell:
@monojenkins build failed

jaykrell:
https://github.com/mono/mono/pull/15599 is a presumed better?preferred fix for the problems this PR was seeing, with the sequence point test, and then the workaround in this PR can go away.

jaykrell:
@monojenkins build failed

jaykrell:
Will require https://github.com/mono/mono/pull/15623 and then undo small part of this.

jaykrell:
@monojenkins build failed

lambdageek:
@jaykrell I don't think we'll take this PR without benchmarks that show a clear win over using TlsGet for those variables.

jaykrell:
It doesn’t use TlsGet currently. That would be a different change.

jaykrell:
The Unix systems have runtime disassembly against presumed undocumented but Windows can’t follow the simple disasm-free documented interface? Are there benchmarks to reuse from the Unix work? I’ll check history.

jaykrell:
To be clear, TlsGetValue and __declspec (thread) 
work differently. Their inputs are not interopable.
TlsGetValue is a little slower.
In order JIT TlsGetValue, the runtime native
code would also have to switch to it.
Currently I believe we JIT icalls.

vargaz:
So we used to have faster tls support, but dropped it because the effort required to support it was just not worth it.
In mono, fast tls needs to support:
- access from generated JITted code
- access from other threads
etc.
Its just too much work for too little gain.

jaykrell:
For access from other threads, 
you can take address on 
current thread & store that, i.e. in another
per-thread data.

They don’t move and portable C can do that.
It is quite easy. 
You can even have a struct,
 not limited to pointer-sized data, and 
there is no overall size limit to take from (1088 pointers
is current limit, per thread, for the process.)

The advantages of slow TLS is possibility 
to turn into FLS, 
and then the FLS destructor callback
 (which is useful even w/o fibers.).

Anyway it seems quite easy,
 not really any trouble.
I’m surprised this is deemed at all 
difficult and not just done.

It would also be trivial to AOT and
 I was going to do that next.
There would just be an indirection
 getting the module index and variable offset.

Pre-Vista was another matter.
I figured this was just leftover from then.

jaykrell:
```
/* This benchmark cheats/incomplete in multiple ways.

1. It includes TlsSetValue, not just TlsGetValue.

TlsGetValue is *sort of* inlinable, by using *sort of* documented.
Documentation says:
"You should not directly access this structure. To access the values of the TlsSlots and TlsExpansionSlots members, call TlsGetValue."
TlsSetValue is less inlinable, because the first call potentially
takes a lock and does an allocation, really undocumented.

2. And it doesn't inline TlsGetValue. Which officially
is not inlinable but mono now does.
Mono does not inline TlsSetValue.

Ultimately it would be nice to remove mono's TlsGetValue / TlsSetValue.
However mingw does not implement __declspec (thread) properly. It is
a wrapper over TlsGetValue / TlsSetValue, in an additional .dll, in the
version I checked.

3. Pure native, not managed/JIT.

4. It is particularly amenable to optimization -- the compiler keeps the address of the thread local in the loop.

The two loops are:

slow:

  0000000140001030: 8B CB              mov         ecx,ebx
  0000000140001032: FF 15 D0 7F 04 00  call        qword ptr [__imp_TlsGetValue]
  0000000140001038: 8B CB              mov         ecx,ebx
  000000014000103A: 48 8D 50 01        lea         rdx,[rax+1]
  000000014000103E: FF 15 CC 7F 04 00  call        qword ptr [__imp_TlsSetValue]
  0000000140001044: 48 83 EF 01        sub         rdi,1
  0000000140001048: 75 E6              jne         0000000140001030

fast:

  0000000140001090: 48 8B 13           mov         rdx,qword ptr [rbx]
  0000000140001093: 48 FF C2           inc         rdx
  0000000140001096: 48 89 13           mov         qword ptr [rbx],rdx
  0000000140001099: 48 83 E9 01        sub         rcx,1
  000000014000109D: 75 F1              jne         0000000140001090

I can try dialing down the cheats.
*/

#include <windows.h>

// volatile to cheat less -- otherwise this is even more optimized, the compiler halves the loop iterations and adds 2 instead of 1
__declspec (thread) volatile size_t fast;

int main()
{
int slow = TlsAlloc();

// Be nice and put the slower first call outside of timed.
TlsSetValue(slow, 0);

int counter;
const int N = 1 << 25;

for (int i = 0; i < 3; ++i)
{

__int64 startSlow = GetTickCount64 ();

for (counter = 0; counter < N; ++counter)
  TlsSetValue(slow, (void*)((size_t)TlsGetValue(slow) + 1));

__int64 endSlow = GetTickCount64 ();

for (counter = 0; counter < N; ++counter)
  ++fast;

__int64 endFast = GetTickCount64 ();

double durationSlow = endSlow - startSlow;
double durationFast = endFast - endSlow;

printf("%f %f %f\n", durationSlow, durationFast, durationSlow / durationFast);

}
}

/*
cl /Zi /O2 1.c /link /incremental:no

Z:\s>1.exe
156.000000 47.000000 3.319149
156.000000 47.000000 3.319149
125.000000 47.000000 2.659574

Z:\s>1.exe
125.000000 47.000000 2.659574
234.000000 32.000000 7.312500
140.000000 47.000000 2.978723

Z:\s>1.exe
125.000000 47.000000 2.659574
125.000000 46.000000 2.717391
125.000000 47.000000 2.659574

Z:\s>1.exe
140.000000 47.000000 2.978723
203.000000 47.000000 4.319149
219.000000 47.000000 4.659574

Z:\s>1.exe
141.000000 47.000000 3.000000
125.000000 47.000000 2.659574
125.000000 47.000000 2.659574

Difference is typically 3x, sometimes less sometimes more.
*/
```

jaykrell:
`TlsGetValue` also isn't currently what managed code does.

lambdageek:
We're not going forward with this.

jaykrell:
That's disappointing.

jaykrell:
This PR was not really about `ThreadStatic` or `ThreadLocal`.
They were mentioned but aren't really relevant.
It was meant just for the thread locals in the runtime.

It does appear that `ThreadStatic` would be easy to greatly optimize and benchmark under AOT, at least.
It does not appear much could be done with `ThreadLocal`, or either under JIT. i.e. due to dynamic vs. static discoverability.

BrzVlad:
Here is my opinion on the fast tls for windows subject.

Firstly, as far as I am concerned, this should be mainly about `ThreadStatic`. The tls variables from mono-tls.c are mainly accessed by key places in jitted code. Such places would be the managed allocator, m2n wrappers, `ThreadStatic`. These should have very good performance. In the runtime we don't really use them as much, or use it in slow places (like exception handling), so it is ok if they are slower (when using TlsGet). Comparing the code in this PR with a very simple alternative that reuses the mechanism that we had for a while (which seems hackish, but has worked for over a decade), aka https://github.com/BrzVlad/mono/commit/f2c4b5ee153e5699f745e78b8fc2c3f85281a2b4, it turns out that, while both options are much faster compared to the current state, the latter is even faster for a `ThreadStatic` microbenchmark. The simplicity and performance improvement would indicate that the commit in question would be better way to go.

The problem with it is that it decreases performance in the aot scenario because it disables `__thread` while also not providing fast tls access in the aot case. This current PR also does not provide support for aot case, which seems to be the main focus of mono nowadays. My opinion on the fast tls on windows is that we should use TlsGet, but only after we have provided support for aot.

jaykrell:
There are three or for things to do.
1. Access to runtime thread locals from JIT. This PR.
2. Access to runtime thread locals from AOT. Building on this PR a little.
3. Access to `ThreadLocal`. No changes, except for their very slight improvement here. They are too dynamic and not statically discoverable I think.
4. Access to `ThreadStatic` from JIT. Also too dynamic, nothing to do, except this small change.
5. Access to `ThreadStatic` from AOT. Agreed this is not the approach for that. This is the approach for access to the runtime's thread locals. The approach for `ThreadStatic` AOT could/should *resemble* how this access the runtime's thread locals. Discover a module's `ThreadStatic`s (outside of generics over value types), sort by alignment, add up their size, fill in the PE metadata for the loader, and access them *like* this PR does, but with the per-variable offset computed by the AOT compiler, along with a per-assembly `_tls_index`. You could possibly leverage the static linker to do some of this -- i.e. like the C compiler.

jaykrell:
We should do this PR first -- runtime thread locals from JIT. Easy.
And then runtime thread locals from AOT (importing _tls_index)
And then `ThreadStatic` from AOT.
This would just be the first step -- the work gets a more valuable and more difficult as it progresses. Start with the easy one.

Issue #15541, opened 2019-07-03
Interpreter doesn't preserve last error on P/Invoke in MinGW builds


filipnavara:
## Steps to Reproduce

1. Compile Mono with MinGW on Windows
2. Run the following program with redirected standard output and `MONO_ENV_OPTIONS=--interp`:

```csharp
using System;
using System.Runtime.InteropServices;

namespace HelloWorld
{
    class Program
    {
        [StructLayoutAttribute(LayoutKind.Sequential)]
        internal struct CONSOLE_SCREEN_BUFFER_INFO
        {
            internal COORD dwSize;
            internal COORD dwCursorPosition;
            internal short wAttributes;
            internal SMALL_RECT srWindow;
            internal COORD dwMaximumWindowSize;
        }

        [StructLayoutAttribute(LayoutKind.Sequential)]
        internal partial struct COORD
        {
            internal short X;
            internal short Y;
        }

        [StructLayoutAttribute(LayoutKind.Sequential)]
        internal partial struct SMALL_RECT
        {
            internal short Left;
            internal short Top;
            internal short Right;
            internal short Bottom;
        }

        internal partial class HandleTypes
        {
            internal const int STD_INPUT_HANDLE = -10;
            internal const int STD_OUTPUT_HANDLE = -11;
            internal const int STD_ERROR_HANDLE = -12;
        }

        [DllImport("kernel32.dll", SetLastError = true)]
        internal static extern bool GetConsoleScreenBufferInfo(IntPtr hConsoleOutput, out CONSOLE_SCREEN_BUFFER_INFO lpConsoleScreenBufferInfo);

        [DllImport("kernel32.dll")]
        internal static extern IntPtr GetStdHandle(int nStdHandle);

        static void Main(string[] args)
        {
            CONSOLE_SCREEN_BUFFER_INFO csbi;
            var outputHandle = GetStdHandle(HandleTypes.STD_OUTPUT_HANDLE);
            Console.WriteLine(GetConsoleScreenBufferInfo(outputHandle, out csbi) + " (expected false)");
            Console.WriteLine(Marshal.GetLastWin32Error() + " (expected 6)");
        }
    }
}
```

### Current Behavior

Prints `false` and `0`.

### Expected Behavior

Prints `false` and `6`.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[x] Windows

**Version Used**:

Mono 6.5 e301847

filipnavara:
[...]

PR #15568, opened 2019-07-04
mono_native_tls_get_value that does not change LastError.

jaykrell:
Should fix https://github.com/mono/mono/issues/15541
without the overhead of https://github.com/mono/mono/pull/15465.

### DIFF
--- a/mono/utils/mono-tls.h
+++ b/mono/utils/mono-tls.h
@@ -52,7 +52,24 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
 #define mono_native_tls_alloc(key,destructor) ((*(key) = TlsAlloc ()) != TLS_OUT_OF_INDEXES && destructor == NULL)
 #define mono_native_tls_free TlsFree
 #define mono_native_tls_set_value TlsSetValue
-#define mono_native_tls_get_value TlsGetValue
+
+#include <winternl.h>
+
+// TlsGetValue always writes 0 to LastError. Which can cause problems. This never changes LastError.
+//
+static inline
+void*
+mono_native_tls_get_value (unsigned index)
+{
+	PTEB const teb = NtCurrentTeb ();
+
+	if (index < TLS_MINIMUM_AVAILABLE)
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/utils/mono-tls.h
```diff
@@ -52,7 +52,32 @@ g_static_assert (TLS_KEY_DOMAIN == 0);
...
+//
+#include <winternl.h>
+
+static inline
+void*
+mono_native_tls_get_value (unsigned index)
+{
+	PTEB const teb = NtCurrentTeb ();
+
+	g_assert (index < TLS_MINIMUM_AVAILABLE + 1024);
```
filipnavara: I see you changed this... I assume `TLS_EXPANSION_SLOTS` was undefined so you replaced it with 1024. Perhaps `8 * sizeof(NtCurrentTeb()->Peb->TlsExpansionBitmapBits)` or something similar is an option?

jaykrell: It is defined but only in obscure headers. It wasn't compiling.
winternal.h does not reveal this either.

jaykrell: i.e. Peb->Bitmap.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Let's remove the assert about 1024 actually, in case the limit is ever raised.

mono/utils/mono-tls.h
jaykrell: Let's remove it anyway, in case limit is raised.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #15570, opened 2019-07-04
[wasm] General Docs

kjpou1:
- Reformatted to add more details.
- Add section on threading.


sdks/wasm/README.md
```diff
@@ -1,34 +1,104 @@
 # Contents
-- **sdk/out/wasm-bcl/wasm** directory: Core libraries to be used with the runtime.
-- **driver.c**, **libmonosgen-2.0.a**, **library_mono.js**, **binding_support.js**: Source / Binaries for custom building the runtime. See compilation instructions down.
-- **debug**, **release** directories: Pre-compiled runtimes using the above driver in release and debug configurations.
-- **sample.html**, **sample.cs**: Sample code, see sample details below.
 
+> **_Note:_** Recent changes have occurred to the directory structure for `Available WebAssembly Runtime Builds` (see below diagrams).  All builds can now be found in one directory `sdk/wasm/builds`.  This was necessary to start supporting more options and functionality.
```
vargaz: This should be in the release notes if we have one instead.

kjpou1: Will create one

kjpou1: Actually, now that I read this better, I read change log. Should we be adding a release notes file if we do not actually have a release yet? We are in discussions of this now how we will be publishing a release.

lewing: ok with letting this go in as is, but I think we should move the note section to a separate file with timestamps that we link to.

PR #15571, opened 2019-07-04
Interpreter reads last error to late after pinvoke.

lateralusX:
This could lead to clobbered last error since interpreter loop needs to run two additional IL instructions in order to read out the value. This works on JIT since the lowering saving the last error will happen right after pinvoke.

This fix makes sure interpreter will get needed information to read out last error right after pinvoke (if requested).

It adds a new IL prefix issued in managed->native wrapper that will tell codegen (and interpreter) that upcoming calli instruction needs to save last error, meaning that it could be done right after issuing the call, no need to run any additional IL instructions to do that work, closing the gap between the return of pinvoke call and save of last error.

Should fix, https://github.com/mono/mono/issues/15541.


mono/mini/mini-x86.c
```diff
@@ -4856,6 +4882,9 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 			x86_mov_membase_reg (code, ins->sreg1, MONO_STRUCT_OFFSET (MonoContext, esi), X86_ESI, sizeof (target_mgreg_t));
 			x86_mov_membase_reg (code, ins->sreg1, MONO_STRUCT_OFFSET (MonoContext, edi), X86_EDI, sizeof (target_mgreg_t));
 			break;
+		case OP_GET_LAST_ERROR:
+			code = emit_get_last_error (code, ins->dreg);
```
vargaz: Needs an entry in cpu-x86.md.

lateralusX: I added:

get_last_error: dest:i len:32

and it runs on 32-bit, anything else missing?

mono/mini/mini-x86.c
```diff
@@ -2293,6 +2293,32 @@ emit_setup_lmf (MonoCompile *cfg, guint8 *code, gint32 lmf_offset, int cfa_offse
...
+#define TEB_LAST_ERROR_SLOT 0x18
+#define TEB_LAST_ERROR_OFFSET 0x34
+
+static guint8*
+emit_get_last_error (guint8* code, int dreg)
+{
+	/* Threads last error value is located in TEB_LAST_ERROR_OFFSET. */
+	x86_prefix (code, X86_FS_PREFIX);
+	x86_mov_reg_mem (code, dreg, TEB_LAST_ERROR_SLOT, sizeof (gpointer));
+	x86_mov_reg_membase (code, dreg, dreg, TEB_LAST_ERROR_OFFSET, sizeof (guint32));
```
jaykrell: This function is materializing the TEB self pointer, and then using that to read the last error.
That is kinda bogus. Instead of two isnstructions, it should just be one.

See https://github.com/mono/mono/pull/15560.
It should be:

```
#define TEB_LAST_ERROR_OFFSET 0x34 

static guint8* 
emit_get_last_error (guint8* code, int dreg) 
{ 
	// Thread's last error value is located in TEB_LAST_ERROR_OFFSET.

	x86_prefix (code, X86_FS_PREFIX); 
	x86_mov_reg_mem (code, dreg,, TEB_LAST_ERROR_OFFSET, sizeof (guint32)); 
}
```

and remove `#define TEB_LAST_ERROR_SLOT`.

It isn't needed and is poorly named. It is the TEB self pointer, and not specific to LastError.

And try to share code between x86 and amd64.

jaykrell: See https://github.com/mono/mono/pull/15574.

lateralusX: This mimics the exact same assembler that you get when calling GetLastError:

mov         eax,dword ptr fs:[00000018h]  
mov         eax,dword ptr [eax+34h]

jaykrell: That isn't optimized either. It can be one instruction instead of two and it must work.
fs: is the TEB. fs:18 points to itself, but using no segment. So for example, you can take the address that way, or use it for multiple default-segmented instructions. But for just accessing one field, the self pointer isn't needed.

`fs:18:x` is the same as `fs:x`.

lateralusX: So I had it like that for x64 in the past, but at one point it started to fail, not getting correct values anymore, so that's when I reverted over to the same solution as used in GetLastErrorValue and the resolved the problem as well as adding an explicit test around this, https://github.com/mono/mono/commit/4ba673343877d2b30d931193af09efe5b8b449b8.

I agree that the name of the first offset is incorrect, could change that, but since what you propose didn't work on x64 and that GetLastError is using the less optimized version make me a little hesitant to change it again.

jaykrell: The code you reference is different.

`amd64_mov_reg_membase (code, dreg, TEB_LAST_ERROR_OFFSET, 0, sizeof (guint32))`
vs.
`x86_mov_reg_mem (code, dreg, TEB_LAST_ERROR_OFFSET, sizeof (guint32)) `

The `membase` and the `0`.
The non-`membase` and non-`0` mimics fetching the self pointer, except size=4 instead of size=pointer.

vargaz: I seem to remember problems in this area wrt virtualization, i.e. see the optimize_for_xen conditionals in mini-amd64.c. Also, this is not perf sensitive code at all, so optimizing it is not very useful.

jaykrell: The membase instruction from the old version is just different and wrong. It takes another register, the base register. What is supposed to be used used here is just a small offset/address.

What I suggest is correct and mimics what is being done to fetch the self pointer, but just skips that and gets the field directly.

Xen isn't relevant as the addresses here are not controlled by the JIT.
The "address", the `mem` parameter, is just a small constant offset.

I realize it might be rare code but the optimization is trivial. Unconditional. No logic to figure it out, etc.

lateralusX: If you are sure it will work out as described above, I can go for the more optimized version. We now have tests on all configurations as well, so believe we should spot any potential problems on CI.

lateralusX: Any idea why GetLastError doesn't go with the more optimized version if it should be straightforward?

jaykrell: I can't 100% guarantee anything in this world but yes I'm pretty certain.

Notice that I do similarly here:

https://github.com/mono/mono/pull/15567/files#diff-93c7ce07dad64cf9f617ea3e79ac8937R528

(though against, pointer-sized vs. 4)
and if that doesn't work, it is catastrophic, and it is working, I ran helloworld x86 and amd64 and verified this code ran.

jaykrell: Why the previous referenced code didn't work or why the native compiler doesn't optimize it?

The previous referenced code is just incorrect.
It takes another register (0) to add to the address.

The compiler, I think, isn't given enough information to optimize it.

It doesn't know, I guess, that fs: and the self pointer are the same thing.
We can ask though.

lateralusX: The old code also ran successfully for a long time both in JIT and AOT scenarios before it showed itself, that's why I switched to what GetLastError did, but lets re-retry the optimization.

filipnavara: > Any idea why GetLastError doesn't go with the more optimized version?

Tough question that is hard to answer without having someone on the kernel team. It is what you would get with `NtCurrentTeb()->LastErrorValue` in C compiler. Although unlikely, maybe nobody just bothered to optimize it.

jaykrell: I'm really confused by that old code.

```
amd64_mov_reg_membase (code, dreg, TEB_LAST_ERROR_OFFSET, 0, sizeof (guint32));
#define amd64_mov_reg_membase(inst,reg,basereg,disp,size)
basereg==ofset?
disp==0?
```

If it was the other way around, I'd hypothesize that reg==rax often == 0 or 1, esp. 0 when the error matters, and so it'd tend to work. But with the parameters also mixed up, what does it do?

jaykrell: I guess the compiler doesn't know that the self pointer is equal to fs:.
It is just two addresses.

lateralusX: > I'm really confused by that old code.

Yes looking at it now doesn't look correct, so something must have changed at some point. When it was original done it resulted in the correct assembler:

mov         ecx,dword ptr gs:[68h]

But then something changed exposing the error.

jaykrell: Ok, so I was confusing x86 and amd64.
0x68 and 0x34.
0x68 & 7 == 0.
So this would tend to load gs:rax, which would tend to be gs:0.
The TEB is largely architecture independent, and the x86 TEB starts with the EH linked list, which isn't used by other architectures, so, I looked once, and it is zero.

Is zero what the error case tended to see?

Granted, besides bool/lasterror, there is also hresult/lasterror, which would go much differently.

Maybe we should sync the tree back and debug it..

lewing:
@monojenkins build failed

jaykrell:
http://www.nynaeve.net/?p=181#comment-19195
dicsusses the two forms of TEB access.

Issue #15533, opened 2019-07-03
Replace mono_native_tls_get_value with MONO_KEYWORD_THREAD when possible.


jaykrell:
There are two ways to use thread locals.

mono_native_tls_get_value
MONO_KEYWORD_THREAD

Some code uses MONO_KEYWORD_THREAD if it is available, with a fallback to mono_native_tls_get_value. Some code just always uses mono_native_tls_get_value.

On some systems, for example Visual C++ and I suspect Linux, MONO_KEYWORD_THREAD is much faster.

As well, mono_native_tls_get_value destroys Win32 LastError (maybe getting fixed, maybe not, being debated).

Proposal is that, for the most part, all uses of mono_native_tls_get_value be converted to use MONO_KEYWORD_THREAD if it is available. That is, converge on the form that is already sometimes used.

And at the same time, when targeting Windows, be sure to never use MONO_KEYWORD_THREAD, even though it is available, as presently it is implemented even worse than mono_native_tls_get_value -- it uses TlsGetValue under the covers, and goes through an extra layer in an extra .dll.

PR #15572, opened 2019-07-04
Mechanize and increase use of thread keyword.

jaykrell:
https://github.com/mono/mono/issues/15533

### DIFF
--- a/mono/metadata/jit-icall-reg.h
+++ b/mono/metadata/jit-icall-reg.h
@@ -51,8 +51,10 @@ MONO_JIT_ICALL (generic_trampoline_vcall)	\
 MONO_JIT_ICALL (mono_tls_get_domain) \
 MONO_JIT_ICALL (mono_tls_get_jit_tls) \
 MONO_JIT_ICALL (mono_tls_get_lmf_addr) \
+MONO_JIT_ICALL (mono_tls_get_marshal_last_error) \
 MONO_JIT_ICALL (mono_tls_get_sgen_thread_info) \
 MONO_JIT_ICALL (mono_tls_get_thread) \
+MONO_JIT_ICALL (mono_tls_get_thread_small_id) \
 	\
 MONO_JIT_ICALL (__emul_fadd)	\
 MONO_JIT_ICALL (__emul_fcmp_ceq)	\
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -89,8 +89,6 @@ enum {
 static MonoCoopMutex marshal_mutex;
 static gboolean marshal_mutex_initialized;
 
-static MonoNativeTlsKey last_error_tls_id;
-
 static MonoNativeTlsKey load_type_info_tls_id;
 
 static gboolean use_aot_wrappers;
@@ -141,7 +139,9 @@ mono_signature_no_pinvoke (MonoMethod *method)
 void
 mono_marshal_init_tls (void)
 {
-	mono_native_tls_alloc (&last_error_tls_id, NULL);
+#ifndef MONO_KEYWORD_THREAD
+	mono_native_tls_alloc (&mono_tls_key_marshal_last_error, NULL);
+#endif
 	mono_native_tls_alloc (&load_type_info_tls_id, NULL);
 }
 
@@ -269,7 +269,9 @@ mono_marshal_cleanup (void)
 	mono_cominterop_cleanup ();
 
 	mono_native_tls_free (load_type_info_tls_id);
-	mono_native_tls_free (last_error_tls_id);
+#ifndef MONO_KEYWORD_THREAD
+	mono_native_tls_free (mono_tls_key_marshal_last_error);
+#endif
 	mono_coop_mutex_destroy (&marshal_mutex);
 	marshal_mutex_initialized = FALSE;
 }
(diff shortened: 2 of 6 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

vargaz:
So we got rid of most of the __thread support, because it only worked on a few arches, and it added a lot of complexity. I don't think we should add it back.

jaykrell:
We still use it a bunch and it is faster.
We have a mix of styles. Sometimes we use it, sometimes we don't.

Some uses are "adaptive" (at compile time) but repeat the work of the centralized adaptive stuff.

It works well on Windows (since Vista), and I believe it always works well on all Linux systems (and Android?), and "now" I believe it finally works well on Apple systems -- but I not sure if we have to wait for our minimum supported version to be new enough. It doesn't work on Watch/x86 for example.

It does not work well on MinGW, true. We don't use it there.

This makes the existing uses more mechanical and then makes it easier to use.
Even if we don't use, it makes it easier to use the alternative.
But I think we should use it more not less.

jaykrell:
I don't like that `FAST_TLS` stuff though. Want to get rid of that? Except, I have it implemented now on Windows and it is pretty simple and I understand it, so let's accept/keep that at least.
Windows doesn't have all those modes like Linux does and forming the offset is just a pointer subtract, no disassembly.

jaykrell:
I'd also be ok with splitting this up -- mechanization first, then increased use.

jaykrell:
I think I should dial it back some for the first PR -- one test failure, probably related.

jaykrell:
Another funny thing here, more about reuse than thread keyword per se, is fine grained control of initialization. The ad-hoc forms move the initializations around in time/place. Probably not useful, but maybe -- probably we might as well initialize them all, early.

Also, the mechanization isn't completely transparent as I think demonstrated by the one test failure and the ifdef in cleanup that I had removed but now put back for a retry.

jaykrell:
The ad-hoc initialization is a small thing, the tests initialize just part of the runtime, using private functions. I'll see about making it compatible but still relatively clean.

jaykrell:
1. I added initialization in the test, that is doing private stuff.
2. I added leaking to preserve historical behavior.
I don't think precise private initialization needs to be preserved.

jaykrell:
Hm:
```
Searching for '-with-tls'...
C:\s\mono2\sdks\builds\android.mk(141):	--with-tls=pthread \
C:\s\mono2\sdks\builds\android.mk(361):	--with-tls=pthread
C:\s\mono2\sdks\builds\android.mk(436):	--with-tls=pthread
C:\s\mono2\sdks\builds\desktop.mk(23):	--with-tls=pthread \
C:\s\mono2\sdks\builds\ios.mk(104):	--with-tls=pthread \
C:\s\mono2\sdks\builds\ios.mk(262):	--with-tls=pthread \
C:\s\mono2\sdks\builds\wasm.mk(227):	--with-tls=pthread \
```
I think these can mostly be switched or let autoconf decide, but, later and after more investigation.
I imagine CI often tests the other setting.
I believe Debian has required working __thread for many years (and dropped any architecture that lacked it).

jaykrell:
@monojenkins build failed

BrzVlad:
The tls keys in mono-tls are accessed both by runtime and compiled managed code, and we have special code there to let the runtime know how to generate OP_TLS_GET. For any other runtime thread local variable we should use whatever is faster (probably __thread), separate from the code in mono-tls.c

jaykrell:
Ah, I see your point -- I was noticing I was imbuing more variables with that feature, of computing and storing the offsets.
So maybe just kinda do almost the same thing twice, once w/o the offsets and their computation, once w/o? Two lists? I can do that.

jaykrell:
Similarly I was adding icalls unnecessarily. That will be fixed by doing what I think you are suggesting. And the enum won't change, ditto.

jaykrell:
For my next revision, I will use __thread on the second list if it is available.

Which is kinda but not exactly what you said.

I can imagine that might want to be a separate control, for the entire list though.
i.e. __thread might be available but not fastest/preferred.

It could also be a separate control on a variable by variable basis, but that is probably done just by leaving them on their own, off of both lists.

For example, only pthread has the destructor callback. Anything needing that should probably remain that way.

Also I picked two variables kinda at random here.
There are more, but I didn't want to keep the PR growing.

jaykrell:
Also I think the native-only ones can/should have two refinements:

1. When they are __thread, declare them in the .h.
2. Inline functions in the .h, esp. for __thread.

For the managed ones to do that requires moving the MONO_TLS_FAST to .h, which I was going to do, but then didn't.

Another reason against the inline functions, is if mono and its client disagree on availability of __thread. Such as by using different compiler. To serve that, the functions must be not inline.
But I'd have to check which, if any, of these are public to outside mono.

For now I think, just not-inline, in the .c. Can change that later.

BrzVlad:
I think having first some defines to abstract away the availability of __thread for declaration, access etc. and use them for more tls variables (for perf and ugliness avoidance, like tls_small_id) should be enough. If we want to refactor anything, move variables around, centralize them in lists etc is for another story.

jaykrell:
I think it is mostly very related -- the macros to produce the variables and functions, might as well be used all at once, via a list(s).

The initialization and cleanup can remain separate, as moving that around proved a bit sensitive.
I put booleans in the list to selectively suppress the initialization/cleanup.

Or you could provide the macros that are being applied to the list and have each variable producer invoke them where they chose. It is *almost* the same thing.
Difference being link granularity. The lists violate link granularity.

Another difference is providing the separate macros might preserve the ability to use file-level static.
i.e. of the keys/variables, that are only used by the functions.
What I have now took that away.

Centralizing initialization/cleanup is nice, because clients can just initialize "the runtime" and not each little piece. But that isn't the present reality.

The list makes sense for the managed ones, because they are already together.
Less sense for the others that are already scattered.

Let me see what I can do.
Probably both -- list the managed ones, don't list the native ones.
Maybe close this an open a new draft as well.

BrzVlad:
As mentioned above, MonoTlsKey is only for tls variables that are accessed by managed code. If we will do a refactoring of other tls accesses in the runtime, we should leave them alone for now.

PR #15574, opened 2019-07-04
emit_get_last_error should emit one instruction, not two.

jaykrell:


jaykrell:
Maybe should lower the size in the .md file.

jaykrell:
Also to be clear, it was correct before.
It was materializing the TEB flat/self pointer, and then using that.
But it isn't needed, it can read the TEB fields directly without the flat/self pointer.

jaykrell:
@monojenkins build failed

jaykrell:
Yes. This came first, but yes.

jaykrell:
I cleaned it up a bit here too, cosmetics.

jaykrell:
Arguably nicer to split it up, since the other PR is really doing other stuff.

jaykrell:
Less interesting since the other went it ahead of it.

PR #15576, opened 2019-07-04
[wasm] Add BlazingPizza.Client to size tests

lewing:


lewing:
@monojenkins build failed

PR #15582, opened 2019-07-05
[sdks] sanity check for maximum number of open file descriptors

lewurm:
Default for `ulimit -n ` seems to be 256, that isn't enough for the Android SDK build.


jaykrell:
Possible to do it in configure? Or don't know what the person building?
How many people configure + make vs. sdks, I wonder.
I kinda think configure should have more of the logic/defaulting, instead of the sdks thing on the side. 'cause that is what I understand how to use and presumably many other people. Alas.

Issue #14857, opened 2019-06-06
[netcore]Make System.Collections.Tests.*.ICollection_NonGeneric_CopyTo_ArrayOfEnumType Pass


steveisok:
This is a common error to a whole bunch of tests, so ideally one fix should fix them all. 

Here is the output of one failure:

```
System.Collections.Tests.ArrayListReadOnlyTests.ICollection_NonGeneric_CopyTo_ArrayOfEnumType(count: 1) [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.InvalidCastException)
      Actual:   (No exception was thrown)
      Stack Trace:
           at System.Collections.Tests.ICollection_NonGeneric_Tests.ICollection_NonGeneric_CopyTo_ArrayOfEnumType(Int32 count)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

According to @filipnavara , the reason for the failures is:

_**Mono currently supports Array.Copy from object[] array containing boxed ints to float[] array with primitive widening**_

Determine if these tests are passable and pass them.

EgorBo:
It's @filipnavara's comment :)

steveisok:
HA!  Sorry @filipnavara, I will change :-)

PR #15585, opened 2019-07-05
[netcore] Throw ICE for Array.Copy(object[], enum[])

EgorBo:
Fixes https://github.com/mono/mono/issues/14857
```csharp
Array src = new object[] { (object)1 };
Array dst = new DateTimeKind[] { DateTimeKind.Local };
src.CopyTo(dst, 0); // should throw InvalidCastException
```

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -23,14 +23,6 @@
 -noclass System.CodeDom.Compiler.Tests.ExecutorTests
 
 ####################################################################
-##  System.Collections.NonGeneric.Tests
-####################################################################
-
-# Mono currently supports Array.Copy from object[] array containing boxed ints to float[] array with primitive widening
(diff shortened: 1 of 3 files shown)


steveisok:
@monojenkins build failed

netcore/CoreFX.issues.rsp
```diff
@@ -68,9 +60,6 @@
 ##  System.Diagnostics.TraceSource.Tests
 ####################################################################
 
-# requires precise gc (should be ignored in corefx)
```
vargaz: These changes appear to be unrelated.

EgorBo: @vargaz I grouped them (and added a few more) in order to remove them all once we get https://github.com/dotnet/corefx/pull/39176 (turns out I should mark more tests with that attribute)

steveisok:
@monojenkins build failed

PR #15591, opened 2019-07-05
[debugger] Add intrinsic for creating a byte array

### LINKED ISSUE #15268: [debugger] Add intrinsic for creating a byte array
We switched to [`CreateByteArray`](https://github.com/mono/debugger-libs/blob/0030e7c2a2760015531dfe2c5c68ae6b99d0f136/Mono.Debugging.Soft/SoftDebuggerAdaptor.cs#L571) from `CreateString`, expecting it to be faster for some serialized data, but it turned out to be slower.

The amazing @vargaz uncovered the reason:

Zoltan Varga [10:40 AM]
the CreateByteArray () code doesn't look _too_ bad.
it does do some invocations into the debuggee.
but at least the array is transferred using one call.
the version using the string only does 1 roundtrip.
so its most likely faster.
the debugger doesn't have an api to create a byte array and initialize it with one call.
so it needs to go through all kinds of hoops to do it.
I can try adding one, but it will take a _long_ time until it propagates through all the projects/products.

Alex Corrado [12:08 PM]
Ahh that explains it. Thanks for checking.
We are using it to pass serialized payloads. Most are pretty small, so the overhead of creating the array for those is probably greater than the overhead of base64 encoding a string. But some payloads could be much larger

## Steps to Reproduce

1. 
2. 
3. 

- [x] macOS
- [ ] Linux
- [ ] Windows

**Version Used**: 5.18

## Stacktrace

```
```

thaystg:
Fixes #15268

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -1905,8 +1905,6 @@ namespace Mono.Debugger.Soft
 		}
 
 		internal long Domain_CreateByteArray (long id, byte [] bytes) {
-			if (!Version.AtLeast (2, 52))
-				throw new NotSupportedException ("This request is not supported by the protocol version implemented by the debuggee.");
 			var typ = (byte)ElementType.U1;
 			var w = new PacketWriter ().WriteId (id).WriteInt (0).WriteInt (bytes.Length);
 			for (int i = 0; i < bytes.Length; i++) {
--- a/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
+++ b/mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
@@ -1026,6 +1026,16 @@ namespace Mono.Debugger.Soft
 				offset += b.Length;
 				return this;
 			}
+			public PacketWriter WriteBytes (byte[] b) {
+				if (b == null)
+					return WriteInt (-1);
+				MakeRoom (4);
+				encode_int (data, b.Length, ref offset);
+				MakeRoom (b.Length);
+				Buffer.BlockCopy (b, 0, data, offset, b.Length);
+				offset += b.Length;
+				return this;
+			}
 
 			public PacketWriter WriteBool (bool val) {
 				WriteByte (val ? (byte)1 : (byte)0);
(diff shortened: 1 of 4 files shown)


thaystg:
@monojenkins build failed

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
```diff
@@ -1903,6 +1904,18 @@ internal void SetSocketTimeouts (int send_timeout, int receive_timeout, int keep
 			return SendReceive (CommandSet.APPDOMAIN, (int)CmdAppDomain.CREATE_STRING, new PacketWriter ().WriteId (id).WriteString (s)).ReadId ();
 		}
 
+		internal long Domain_CreateByteArray (long id, byte [] bytes) {
+			if (!Version.AtLeast (2, 52))
```
vargaz: This can be checked in the caller by calling vm.CheckProtocolVersion ().

mono/mini/debugger-agent.c
```diff
@@ -7207,6 +7208,29 @@ domain_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
...
 		break;
 	}
+	case CMD_APPDOMAIN_CREATE_BYTE_ARRAY: {
+		ERROR_DECL (error);
+		MonoArray *arr;
+		gpointer elem;
+		domain = decode_domainid (p, &p, end, NULL, &err);
+		uintptr_t size = 0;
+		printf("antes do primeiro decode_int\n");
+		int index = decode_int (p, &p, end);
```
vargaz: Debug stuff.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/Connection.cs
```diff
@@ -1903,6 +1904,18 @@ internal void SetSocketTimeouts (int send_timeout, int receive_timeout, int keep
...
+				throw new NotSupportedException ("This request is not supported by the protocol version implemented by the debuggee.");
+			var typ = (byte)ElementType.U1;
+			var w = new PacketWriter ().WriteId (id).WriteInt (0).WriteInt (bytes.Length);
+			for (int i = 0; i < bytes.Length; i++) {
+				w.WriteByte (typ);
```
vargaz: Wouldn't be easier to write just bytes ?

thaystg:
@monojenkins build failed

mono/mini/debugger-agent.c
```diff
@@ -7207,6 +7208,21 @@ domain_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
...
+		ERROR_DECL (error);
+		MonoArray *arr;
+		gpointer elem;
+		domain = decode_domainid (p, &p, end, NULL, &err);
+		uintptr_t size = 0;
+		int len = decode_int (p, &p, end);
+		size = len;
+		arr = mono_array_new_full_checked (mono_domain_get (), mono_class_create_array (mono_get_byte_class(), 1), &size, NULL, error);
+		elem = (gpointer*)arr->vector;
+		memcpy(elem, p, len);
```
vargaz: This should be 
elem = mono_array_addr_internal (arr, guint8, 0);

jaykrell: Space after `memcpy`.
No need for cast, or `elem` temporary.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins commit apidiff

vargaz:
@monojenkins build failed

thaystg:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins commit apidiff

PR #15595, opened 2019-07-06
experiment with single_imm_size

jaykrell:
Act as if single_imm_size is always true, to confirm my understanding.

This will not be commited to master.
Unless we want some gross simplification at the cost of larger code, i.e. 6 bytes when anywhere from 1-6 are required.


jaykrell:
Setup to do this locally, and giving up for now.

PR #15599, opened 2019-07-07
[JIT][amd64][x86] Partial single_imm_size support

jaykrell:
This is a flag to stablilize codegen size, for test purposes.
It not being implemented makes the sequence point test not reliable.
Esp. with https://github.com/mono/mono/pull/15567.
For purposes of #15567 this could be less.

There are multiple other options if #15567 it is the only problem.
1. Don't do #15567.
2. Fill in nops dynamically if the value is zero, only in that path (it is probably zero/non-zero, not imm8/non-imm8).
3. Duplicate just the macro it uses.

But this PR does seem most correct as the design intends and could stave off future similar problems.

### DIFF
--- a/mono/arch/x86/x86-codegen.h
+++ b/mono/arch/x86/x86-codegen.h
@@ -268,6 +268,8 @@ typedef union {
 #define x86_modrm_reg(modrm) (((modrm) >> 3) & 0x7)
 #define x86_modrm_rm(modrm) ((modrm) & 0x7)
 
+#define x86_use_imm8(x) (!mini_debug_options.single_imm_size && x86_is_imm8 (x))
+
 #define x86_address_byte(inst,m,o,r) do { *(inst)++ = ((((m)&0x03)<<6)|(((o)&0x07)<<3)|(((r)&0x07))); } while (0)
 #define x86_imm_emit32(inst,imm)     \
 	do {	\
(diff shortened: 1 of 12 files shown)


jaykrell:
@monojenkins build failed

mono/arch/x86/x86-codegen.h
```diff
@@ -309,7 +316,7 @@ typedef union {
 		x86_address_byte ((inst), 0, (r), (basereg));	\
```
jaykrell: Might need to disallow the 0 also, under single_imm_size.

jaykrell: That can be done separately if needed.
I'd rather keep this PR reduced to "imm8" and ignore "0".

vargaz:
I'd prefer removing those tests, they are very brittle.

jaykrell:
Is the test useful? Catch bugs?

I can dial back this change if you want to keep the test.

In particular, in codegen.h, can limit it to just one macro.
The .md changes can all or all-but-one be undone, since I gave up and just double.

I think the asserts throughout this change are ok/good independently.
I can PR those separately.

There is also a narrow workaround in the other PR.

It does seem a *little* gross all the size dependencies in the trampolines -- they depend on "optimization", granted, a very simple and always-on optimization, that is fairly stable, unless some offsets crossover the boundaries.

They are also fixable otherwise, like by making two passes or using an auto-growing buffer (like std::vector).

I'm ok either way, it'd just be lame imho if this was why __declspec(thread) access didn't get inlined.

jaykrell:
Hey, it is already disabled for ARM, I just noticed..

jaykrell:
Closed in deference to https://github.com/mono/mono/pull/15623.

jaykrell:
Closed in deference to https://github.com/mono/mono/pull/15623.

PR #15604, opened 2019-07-08
Insertion 3.3.0 beta1 19327 03

agocke:


configure.ac
```diff
@@ -6026,8 +6026,8 @@ fi
 AC_SUBST(mono_runtime)
 AC_SUBST(mono_runtime_wrapper)
 
-CSC_LOCATION=`cd $srcdir && pwd`/external/roslyn-binaries/Microsoft.Net.Compilers/3.1.0-release/csc.exe
-VBCS_LOCATION=`cd $srcdir && pwd`/external/roslyn-binaries/Microsoft.Net.Compilers/3.1.0-release/VBCSCompiler.exe
+CSC_LOCATION=`cd $srcdir && pwd`/external/roslyn-binaries/Microsoft.Net.Compilers/latest/csc.exe
```
radical: IMHO, we should continue to use the exact version here instead of `latest`. That makes it easy to just glance at this to see which roslyn version is being used.

agocke: In that case could we make the `configure.ac` variable more machine-parseable? I'm not familiar with the syntax here. Could we just rip out the number and call it "version number"?

One negative is that it would also cause significant churn in this file for every insertion, instead of just the submodule. Is that desirable?

PR #15605, opened 2019-07-08
[wasm][sdk][tests] Use the SDK to build the browser tests

kjpou1:
# Fix the runtime paths evaluation.
- If running from a master build the runtime builds path is different from the nuget paths.
  - From a master build the runtime builds are found in `wasm\builds` now.

# Browser Tests modifications:
- Moved all sources to a sources directory `tests\browser\src\`.
- Create a solutions project that can be built by VSfMac or VS using the SDK
- Create separate projects for each TestSuite.
- Change makefile target for the browser tests to use `dotnet` to build instead of the `packager.exe`

# SDK modifications and fixes:

Fix sdk build to run `monolinker.exe` using `dotnet`
- Run `mono` as task tool if running from `dotnet` and not running under Windows instead of executing the `monolinker.exe` directly
- Clean up WasmLinkAssemblies code.
   - Add routine to determine if build was initiated from `dotnet` instead of msbuild.
   - Add common routine to find the correct monolinker.exe to execute depending on running local or nuget.

Without the previous fixes, `monolinker` step can fail with the following message when using `dotnet build`

```
.nuget/packages/mono.webassembly.build/0.1.0-preview1/build/netstandard2.0/Mono.WebAssembly.Build.targets(76,3): error MSB6006: "monolinker.exe" exited with code 8.
```

The faling msbuild target is:

```
      _MonoWasmLink:
         /Users/Jimmy/.nuget/packages/mono.webassembly.build/0.1.0-preview1/build/tools/monolinker.exe --verbose --exclude-feature
         ... the rest of the command is cut off from here to save space

```

With the fixes command above becomes the following which execute the `monolinker` with `mono`. :

```
      _MonoWasmLink:
         /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono /Users/Jimmy/.nuget/packages/mono.webassembly.build/0.1.0-preview1/build/tools/monolinker.exe --verbose --exclude-feature remoting --exclude-feature com

         ... the rest of the command is cut off from here to save space
```


kjpou1:
@monojenkins Linux WebAssembly

kjpou1:

kjpou1:

kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build

vargaz:
The sources directory is usually called 'src' by convention.

mhutch:
`MonoWasmCLIMode` shouldn't be necessary, `ToolTask` starts the process with `Process.Start`, which under Mono will run an exe using Mono (when `ProcessStartInfo.ShellExecute` is false, which it is here).

The only exception is if the `ToolTask.UseCommandProcessor` property is `true`, but I don't see why it should be in this case.

kjpou1:
@monojenkins build failed

kjpou1:
> The sources directory is usually called 'src' by convention.

Will change this.

kjpou1:
> The sources directory is usually called 'src' by convention.

Changed in recent commit

kjpou1:
> @radical can you take a look at this and see if there is some other way we should be dealing with this? Particularly the ToolPath parts.

> `MonoWasmCLIMode` shouldn't be necessary, `ToolTask` starts the process with `Process.Start`, which under Mono will run an exe using Mono (when `ProcessStartInfo.ShellExecute` is false, which it is here).
> 
> The only exception is if the `ToolTask.UseCommandProcessor` property is `true`, but I don't see why it should be in this case.


Change `<MonoWasmCLIMode>` to false so that we can get the actual err…
…or message.

- This will also print a log message with some info:
   - CLI Mode used
   - The assembly location directory
   - The `monolinker.exe` tools path.

kjpou1:

kjpou1:
> > @radical can you take a look at this and see if there is some other way we should be dealing with this? Particularly the ToolPath parts.
> 
> > `MonoWasmCLIMode` shouldn't be necessary, `ToolTask` starts the process with `Process.Start`, which under Mono will run an exe using Mono (when `ProcessStartInfo.ShellExecute` is false, which it is here).
> > The only exception is if the `ToolTask.UseCommandProcessor` property is `true`, but I don't see why it should be in this case.
> 
> Change `<MonoWasmCLIMode>` to False so that we can get the actual error message.
> 
> * This will also print a log message with some info:
>   
>   * CLI Mode used
>   * The assembly location directory
>   * The `monolinker.exe` tools path.

Just a little more information.

Using `msbuild /r tests/browser/src` instead of `dotnet build tests/browser/src` executes the `monolinker` correctly.  In some instances using `dotnet msbuild /r tests/browser/src` works correctly but in this case neither one of the `dotnet` commands work. 

Executing with  `dotnet msbuild /r tests/browser/src` gives the following:

```
error MSB6003: The specified task executable "monolinker.exe" could not be run. System.ComponentModel.Win32Exception (13): Permission denied 
```

mhutch:
Ok, so the issue is that when running with `dotnet`, the .NET Core runtime doesn't know how to execute a net472 exe when not on Windows. That makes sense.

I don't think the right solution is to depend on Mono being in `PATH` though. That's going to complicate things like CI a lot.

I would suggest instead shipping a version of `monolinker.exe` that can run on .Net Core (i.e. a published netcoreapp shared framework build) or integrating the linker into the task directly.

If you think it make sense to continue with this approach for now as a temporary workaround, I would suggest using something more explicit:

```
<_MonoWasmLink
    InvokeLinkerUsingMono="$(MSBuildRuntimeType)=='Core' And $([MSBuild]::IsOSUnixLike())"
    ...
```

kjpou1:
> Ok, so the issue is that when running with `dotnet`, the .NET Core runtime doesn't know how to execute a net472 exe when not on Windows. That makes sense.
>

On later versions of `dotnet` this works.  On this version of the dotnet framework, 2.2.203, it unfortunately does not work.
 
> I don't think the right solution is to depend on Mono being in `PATH` though. That's going to complicate things like CI a lot.
> 

I agree, was hoping it was something else so that mod was not actually needed.

> I would suggest instead shipping a version of `monolinker.exe` that can run on .NET Core (i.e. a published netcoreapp shared framework build) or integrating the linker into the task directly.
> 

Thought about it but in the end concluded it was too much but now not so sure.

> If you think it make sense to continue with this approach for now as a temporary workaround, I would suggest using something more explicit:
> 
> ```
> <_MonoWasmLink
>     InvokeLinkerUsingMono="$(MSBuildRuntimeType)=='Core' And $([MSBuild]::IsOSUnixLike())"
>     ...
> ```

Like that idea and may be the solution for now though as a stop gap.  Will look into the possibility of the idea above.

Thanks for the feedback Mikayla.

kjpou1:
@monojenkins build failed

lewing:
@monojenkins build failed

Issue #15188, opened 2019-06-16
[netcore] Make System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo Pass


steveisok:
The test currently fails with:

```
System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo(skipFrames: 1, fNeedFileInfo: True) [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
           at System.Linq.Enumerable.Select[StackFrame,MethodBase](IEnumerable`1 source, Func`2 selector)
           at System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo(Int32 skipFrames, Boolean fNeedFileInfo)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo(skipFrames: 1, fNeedFileInfo: False) [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
           at System.Linq.Enumerable.Select[StackFrame,MethodBase](IEnumerable`1 source, Func`2 selector)
           at System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo(Int32 skipFrames, Boolean fNeedFileInfo)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

Issue #15182, opened 2019-06-15
[netcore] Make System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames Pass


steveisok:
The test currently fails with:

```
System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames(skipFrames: 1) [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
           at System.Linq.Enumerable.Select[StackFrame,MethodBase](IEnumerable`1 source, Func`2 selector)
           at System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames(Int32 skipFrames)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

PR #15607, opened 2019-07-08
[netcore] Do not inline ThrowHelpers, fix SkipFrames handling.

EgorBo:
1) Do not inline ThrowHelpers (e.g. `static void ThrowException() => throw new Exception();`)
2) `get_trace` incorrectly handles `skip` argument
3) fix broken ci (`dl-test-assets.py` is not python2 compatible any more)

Fixes https://github.com/mono/mono/issues/15182
Fixes https://github.com/mono/mono/issues/15188


netcore/Makefile
```diff
@@ -126,7 +126,7 @@ update-tests-corefx: corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION)
 
 corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
 	curl $(TEST_ASSETS_URL) --output corefx-test-assets.xml
-	python dl-test-assets.py corefx-test-assets.xml $(FEED_BASE_URL) corefx/tests
```
vargaz: Unrelated changes ?

mono/mini/mini-exceptions.c
```diff
@@ -1141,7 +1141,7 @@ ves_icall_get_trace (MonoException *exc, gint32 skip, MonoBoolean need_file_info
 		}
 
 		mono_debug_free_source_location (location);
-		mono_array_setref_internal (res, i, sf);
+		mono_array_setref_internal (res, i - skip, sf);
```
vargaz: Why did this work previously ?

EgorBo: I guess it never worked if `skip` was > 0, e.g. this crashes on regular mono:
```csharp
using System;
using System.Diagnostics;

namespace HelloWorld
{
    class Program
    {
        private static Exception InvokeException()
        {
            try
            {
                ThrowException();
                return null;
            }
            catch (Exception ex)
            {
                return ex;
            }
        }
        
        private static void ThrowException() => throw new Exception();
        
        {
            int skipFrames = 1;
            Exception ex = InvokeException();
            var exceptionStackTrace = new StackTrace(ex, skipFrames);
            Console.WriteLine(exceptionStackTrace);
        }
    }
}
```

filipnavara: Presumably because in most cases it was called with zero. In non-netcore build all the default constructors pass `METHODS_TO_SKIP == 0`.

netcore/Makefile
EgorBo: @vargaz on our CI `python` stands for Python 2 (default for macOS) and https://github.com/mono/mono/pull/15524 made that script python3 only

vargaz: I mean the change is unrelated to the PR.

EgorBo: I understand but without it I will not be able to see CI results for this PR 🙂 all netcore lanes are currently crash on the `download corefx tests` step in master. I can file a separate PR for it.

EgorBo:
/azp run
@monojenkins build failed

vargaz:
@monojenkins build failed

steveisok:
@monojenkins build failed

Issue #14611, opened 2019-05-23
Compile time regression on interp.c


vargaz:
Compiling interp/interp.c with xcode10.2 takes more than a minute.

Seems to be related to computed goto's, if thats disabled, then compile time goes down to about 20s.

lewurm:
That's with `CFLAGS=-O2` only, right?

vargaz:
Yes.

SamMonoRT:
@lewurm - the referenced issue is Merged, any action needed on this one now ?

vargaz:
It has nothing to do with this issue, this still happens.

vargaz:
On wasm-runtime-release, the build still takes about 2 minutes, or sometimes it gets stuck forever.
And wasm has computed gotos disabled, so the issue is somewhere else.
A stack trace of clang shows its stuck here:
  * frame #0: 0x0000000106a22f00 libLLVM.dylib`llvm::SlotIndexes::getIndexAfter(llvm::MachineInstr const&) const + 240
    frame #1: 0x0000000106c4a546 libLLVM.dylib`(anonymous namespace)::RegisterCoalescer::joinCopy(llvm::MachineInstr*, bool&)::$_2::operator()(llvm::MachineInstr&) const + 118
    frame #2: 0x0000000106c41f4c libLLVM.dylib`(anonymous namespace)::RegisterCoalescer::joinCopy(llvm::MachineInstr*, bool&) + 5612
    frame #3: 0x0000000106c40034 libLLVM.dylib`(anonymous namespace)::RegisterCoalescer::copyCoalesceWorkList(llvm::MutableArrayRef<llvm::MachineInstr*>) + 420
    frame #4: 0x0000000106c3ec67 libLLVM.dylib`(anonymous namespace)::RegisterCoalescer::runOnMachineFunction(llvm::MachineFunction&) + 4247
    frame #5: 0x0000000106af8c8d libLLVM.dylib`llvm::MachineFunctionPass::runOnFunction(llvm::Function&) + 349

lewurm:
Sounds like a compile-time regression in LLVM to me.

Here's a report with `-mllvm -time-passes`:
```
[...]
===-------------------------------------------------------------------------===
                      ... Pass execution timing report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 63.8235 seconds (63.8362 wall clock)

   ---User Time---   --System Time--   --User+System--   ---Wall Time---  --- Name ---
[...]

PR #15613, opened 2019-07-08
[interp] Avoid declaring MonoError's in interp_exec_method_full (), it increases the native stack frame size because the optimizer can't seem to figure out live ranges.

vargaz:


lewurm:
does that help with https://github.com/mono/mono/issues/14611 ?

vargaz:
No, but it should increase perf a little bit, and prevent possible stack overflows on some platforms like wasm.

mono/mini/interp/interp.c
```diff
@@ -992,7 +991,7 @@ ves_array_calculate_index (MonoArray *ao, stackval *sp, InterpFrame *frame, gboo
 	return pos;
 }
 
-static void
+static MONO_NEVER_INLINE void
```
jaykrell: These might merit
comments — “to avoid growing caller’s frame” or somesuch.
NT does similar.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #15619, opened 2019-07-08
Remove support for llvm36, its not used anymore.

vargaz:


lewurm:
what about all the LLVM version ifdefs in `mini-llvm.c`?

vargaz:
Will remove those as well.

vargaz:
@monojenkins build failed

PR #15622, opened 2019-07-09
[merp] Add API method that whitelists all native libraries

alexischr:
This is so that complex consumers such as Visual Studio for Mac can forego attempting to enumerate all loaded libraries, register them one at a time and potentially have to keep the whitelist updated.

### DIFF
--- a/mcs/class/corlib/Mono/Runtime.cs
+++ b/mcs/class/corlib/Mono/Runtime.cs
@@ -194,6 +194,14 @@ namespace Mono {
 		}
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern void RegisterReportingForAllNativeLibs_internal ();
+
+		static void RegisterReportingForAllNativeLibs ()
+		{
+			RegisterReportingForAllNativeLibs_internal ();
+		}
+
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		static extern void RegisterReportingForNativeLib_internal (IntPtr modulePathSuffix, IntPtr moduleName);
 
 		static void RegisterReportingForNativeLib (string modulePathSuffix_str, string moduleName_str)
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -6380,6 +6380,15 @@ ves_icall_Mono_Runtime_RegisterReportingForNativeLib (const char *path_suffix, c
 }
 
 void
+ves_icall_Mono_Runtime_RegisterReportingForAllNativeLibs ()
+{
+#ifndef DISABLE_CRASH_REPORTING
+	if (mono_get_eh_callbacks ()->mono_allow_all_native_libraries)
+		mono_get_eh_callbacks ()->mono_allow_all_native_libraries ();
+#endif
+}
+
+void
 ves_icall_Mono_Runtime_EnableCrashReportingLog (const char *directory, MonoError *error)
 {
 #ifndef DISABLE_CRASH_REPORTING
--- a/mono/mini/mini-exceptions.c
+++ b/mono/mini/mini-exceptions.c
@@ -133,6 +133,7 @@ static void mono_summarize_managed_stack (MonoThreadSummary *out);
 static void mono_summarize_unmanaged_stack (MonoThreadSummary *out);
 static void mono_summarize_exception (MonoException *exc, MonoThreadSummary *out);
 static void mono_crash_reporting_register_native_library (const char *module_path, const char *module_name);
+static void mono_crash_reporting_allow_all_native_libraries (void);
 
 static gboolean
 first_managed (MonoStackFrameInfo *frame, MonoContext *ctx, gpointer addr)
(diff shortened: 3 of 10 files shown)


lambdageek:

mono/mini/mini-exceptions.c
```diff
@@ -1446,12 +1449,20 @@ mono_crash_reporting_register_native_library (const char *module_path, const cha
...
+	allow_all_native_libraries = TRUE;
+}
+
 static gboolean
 check_whitelisted_module (const char *in_name, const char **out_module)
 {
 #ifndef MONO_PRIVATE_CRASHES
 		return TRUE;
 #else
+	if (allow_all_native_libraries)
```
leculver: This code has to set out_module, or the caller of this function is a no-op.

alexischr: @leculver could you please explain? From what I see, it is benign for callers to not set out_module, including when the frame information is written out. That said, we may just want to set it to `g_strdup (in_name)`.

leculver: I may be misinterpreting this, I haven't had a chance to step through/debug it.  If out_module isn't set I think we do not take this branch, leaving `has_name` to false:

https://github.com/mono/mono/blob/cb83160ea4d60f030e71bcea60b19baa06f82192/mono/mini/mini-exceptions.c#L1749

If `has_name` isn't set, we end up bypassing writing anything out here:

https://github.com/mono/mono/blob/d54a6544b3165f13bb9cd9b5585648b0ac708622/mono/utils/mono-state.c#L532

Unless `str_descr` is originally set to a meaningful value before the function call to `check_whitelisted_module`?

In any case, testing should show whether things are being written out or not.

lambdageek:
@monojenkins build failed

mono/mini/mini-exceptions.c
lambdageek: This code is a bit confusing, so I'm not sure how it will behave, but I just want to mention @alexischr that you shouldn't call `g_strdup (in_name)` - all the crash reporting stuff is running in async (signal handler) context, so no allocating memory.

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

lambdageek:

lambdageek:
@monojenkins build

lambdageek:
@monojenkins build OS X package

mono/mini/mini-exceptions.c
```diff
@@ -133,6 +133,7 @@ static void mono_summarize_managed_stack (MonoThreadSummary *out);
 static void mono_summarize_unmanaged_stack (MonoThreadSummary *out);
 static void mono_summarize_exception (MonoException *exc, MonoThreadSummary *out);
 static void mono_crash_reporting_register_native_library (const char *module_path, const char *module_name);
+static void mono_crash_reporting_allow_all_native_libraries (void);
```
lambdageek: Windows builds are unhappy about this declaration (should be under the same ifdef as the definition):

```
..\mono\mini\mini-exceptions.c(4004): error C2129: static function 'void mono_crash_reporting_allow_all_native_libraries(void)' declared but not defined
```

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

alexischr:

alexischr:
@monojenkins build failed

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build OS X package

mono/metadata/icall.c
```diff
@@ -6379,6 +6379,15 @@ ves_icall_Mono_Runtime_RegisterReportingForNativeLib (const char *path_suffix, c
 #endif
 }
 
+void
+ves_icall_Mono_Runtime_RegisterReportingForAllNativeLibs ()
```
jaykrell: `(void)`

mcs/class/corlib/Mono/Runtime.cs
```diff
@@ -193,6 +193,14 @@ static void EnableMicrosoftTelemetry (string appBundleID_str, string appSignatur
 			return new Tuple<String, ulong, ulong> (payload_str, portable_hash, unportable_hash);
 		}
 
+		[MethodImplAttribute (MethodImplOptions.InternalCall)]
+		static extern void RegisterReportingForAllNativeLibs_internal ();
+
+		static void RegisterReportingForAllNativeLibs ()
```
jaykrell: Why extra layer?

kdubau:
Should this be backported to 2019-06 for VSMac 8.3?

marek-safar:
@monojenkins backport 2019-08

marek-safar:
@alexischr please backport 2019-08

@kdubau we could do 2019-06 but it feels a little bit late for a new feature to propagate through

PR #15624, opened 2019-07-09
Cleanup trampoline size checks.

jaykrell:
- named `int` everywhere
- `assert` everywhere
- assert `<=` everywhere instead of `<`
- `const` almost everywhere
Names not consistent, following existing code (`size`, `buf_len`, `kMaxCode`.)

Only touching x86/amd64, granted.

Could be better -- make them two pass and avoid the presizing.

### DIFF
--- a/mono/mini/exceptions-amd64.c
+++ b/mono/mini/exceptions-amd64.c
@@ -81,8 +81,10 @@ get_win32_restore_stack (void)
 	if (start)
 		return start;
 
+	const int size = 128;
+
 	/* restore_stack (void) */
-	start = code = mono_global_codeman_reserve (128);
+	start = code = mono_global_codeman_reserve (size);
 
 	amd64_push_reg (code, AMD64_RBP);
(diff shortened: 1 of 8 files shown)


mono/mini/exceptions-amd64.c
```diff
@@ -108,7 +110,7 @@ get_win32_restore_stack (void)
 	amd64_mov_reg_imm (code, AMD64_R11, mono_restore_context);
 	amd64_call_reg (code, AMD64_R11);
 
-	g_assert ((code - start) < 128);
+	g_assertf ((code - start) <= size, "%d %d", (int)(code - start), size);
```
vargaz: Does this really fail often enough for this message to be useful ?

jaykrell: Only if you try to enable `single_imm_size` globally as I tried.
Just `assert` but not `assertf`?
Or dismiss entire PR?

jaykrell:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #15626, opened 2019-07-09
g_strnlen for portability.

jaykrell:
See https://github.com/mistydemeo/tigerbrew/issues/605.
See https://www.gnu.org/software/gnulib/manual/html_node/strnlen.html.

Note that the patch referred to using max and strlen isn't quite right.
strnlen is not supposed to scan past the specified number of bytes.


jaykrell:
@monojenkins build failed

NattyNarwhal:
I'll need to get out my iBook (which is a bit buried ATM) to test these changes.

PR #15629, opened 2019-07-09
[cxx] Portably pass a full pointer and not just an integer to terminate

jaykrell:
vararg pointer lists.

i.e. because of define NULL 0.
Though I think at least Visual C++ special cases this and passes a full pointer to varargs.


CoffeeFlux:
Was this causing issues somewhere? I would think it's fine as-is.

jaykrell:
I didn't see a runtime malfunction but this is not portable C++.
It is valid and commin in C++ to have `#define NULL 0` that will only pass a 32bit integer, not a full 64bit pointer. The upper bits of such a parameter are not defined. Maybe they will also be zero, maybe they will be leftover garbage.
Producing zero on amd64 at least _tends_ to produce a 64bit zero, because 32bit operations are zero extended.

jaykrell:
@monojenkins build failed

jaykrell:
Sorry, in registers, amd64 tends to widen 32bit to 64bit.
In memory, less so.
At the end of a long parameter list, could be memory.

This really isn't portable reliable C++.
It skates by usually in C because `#define NULL ((void*)0)` is valid C, because `void*` converts to any pointer silently.
Another good option is to overload on parameter count, or hardcode a small limit and default the parameters.

Or pass an array of pointers. Really.

varargs is not great ever.

CoffeeFlux:
@monojenkins build failed

jaykrell:
The code as written is not defined. It can crash. The fix is simple and obvious. Surprising it is not taken. I'll check the breaks and rebase to master.

CoffeeFlux:
Yeah, I was hesitant to review until it was passing CI since those didn't look like random failures.

jaykrell:
Yeah, sorry, there were a few clerical errors.
```
mono_foo_varargs_expecting_const_char_pointer (… (const char*)NULL) // good
mono_foo_not_varargs_not_const_char_pointer (… (const char*)NULL); // oops
```

jaykrell:
Also there are other good fixes, I should say. This isn't clearly best, but it is adequate and correct.
```
#define g_strconcat (...) do { char const * const params ] = { __VA_ARGS__ }; g_strconcatn (G_N_ELEMENTS (params), params); } while (0);```

is arguably better -- avoid runtime varargs.

jaykrell:
I'm very open to that last option, to be clear.
Notice it'd still have to check for null, to match current semantics.
There are other options too.

jaykrell:

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build failed

PR #15632, opened 2019-07-09
Cleanup mono_threads_platform_get_stack_bounds.

jaykrell:
And fix mingw warning unused function __readfsdword.
Prepare for, but don't actually, bumping minimum from Windows 7 to Windows 8. If there is a UWP-specific build, it will use the Windows 8 version.


jaykrell:
@monojenkins build Linux x64

jaykrell:
@monojenkins build failed

PR #15636, opened 2019-07-09
Fix build on Autoconf 2.63

directhex:
This was broken in e7f9dbe73c3344c96a13892a6c4c45f8e5c35aa3, because
AS_VAR_APPEND does not exist until v2.63b (?!)


directhex:
@monojenkins build failed

Issue #15348, opened 2019-06-21
[netcore] Make System.Reflection.Tests.TypeTests_HiddenFields.GetFieldDoesNotHideHiddenFields Pass


steveisok:
The test currently fails with:

```
System.Reflection.Tests.TypeTests_HiddenFields.GetFieldDoesNotHideHiddenFields [FAIL]
      Assert.Equal() Failure
      Expected: 4
      Actual:   0
      Stack Trace:
           at System.Reflection.Tests.TypeTests_HiddenFields.GetFieldDoesNotHideHiddenFields()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

Issue #15347, opened 2019-06-21
[netcore] Make System.Reflection.Tests.AssemblyTests.CrossAssemblyTypeRefToNestedType Pass


steveisok:
The test currently fails with:

```
System.Reflection.Tests.AssemblyTests.CrossAssemblyTypeRefToNestedType [FAIL]
      Assert.Equal() Failure
      Expected: typeof(Outer+Inner)
      Actual:   typeof(Outer+Inner+ReallyInner)
      Stack Trace:
           at System.Reflection.Tests.AssemblyTests.CrossAssemblyTypeRefToNestedType()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

Issue #15346, opened 2019-06-21
[netcore] Make System.Reflection.Tests.TypeTests_HiddenProperties.GetPropertyHidesPropertiesByNameAndSigAndCallingConventionCompare Pass


steveisok:
The test currently fails with:

```
System.Reflection.Tests.TypeTests_HiddenProperties.GetPropertyHidesPropertiesByNameAndSigAndCallingConventionCompare [FAIL]
      Assert.Equal() Failure
      Expected: String[] ["Item", "MyInstanceThenStaticProp", "MyStaticThenInstanceProp", "MyStringThenDoubleProp"]
      Actual:   String[] []
      Stack Trace:
           at System.Reflection.Tests.TypeTests_HiddenProperties.GetPropertyHidesPropertiesByNameAndSigAndCallingConventionCompare()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

PR #15641, opened 2019-07-09
[netcore] Fix IsAssignableFrom

EgorBo:
Mono used to return `true` for these three cases (see System.Reflection.Tests.TypeTests.IsAssignableFrom tests):
```csharp
using System.Reflection;
using Xunit;

namespace ConsoleApp
{
    public static class Program
    {
        {
            Assert.False(typeof(TI_Interface1[]).GetTypeInfo().IsAssignableFrom(
                         typeof(TI_StructWithInterface[]).GetTypeInfo()));

            Assert.False(typeof(int?[]).GetTypeInfo().IsAssignableFrom(
                         typeof(int[]).GetTypeInfo()));

            Assert.False(typeof(int[]).GetTypeInfo().IsAssignableFrom(
                         typeof(int?[]).GetTypeInfo()));
            
        }
    }
    
    public interface TI_Interface1 { }
    public struct TI_StructWithInterface : TI_Interface1 { }
}
```
Fixes https://github.com/mono/mono/issues/10848
Fixes https://github.com/mono/mono/issues/15080

Unrelated  System.Reflection.MetadataLoadContext.Tests tests that don't fail anymore:
Close https://github.com/mono/mono/issues/15351
Close https://github.com/mono/mono/issues/15348
Close https://github.com/mono/mono/issues/15347
Close https://github.com/mono/mono/issues/15346
Close https://github.com/mono/mono/issues/15345
Close https://github.com/mono/mono/issues/15343
Close https://github.com/mono/mono/issues/15342
Close https://github.com/mono/mono/issues/15341
Close https://github.com/mono/mono/issues/15338
Close https://github.com/mono/mono/issues/15337
Close https://github.com/mono/mono/issues/15336

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -3820,10 +3820,15 @@ mono_class_is_assignable_from_checked (MonoClass *klass, MonoClass *oklass, gboo
 		if (m_class_is_valuetype (eoclass)) {
 			if ((eclass == mono_defaults.enum_class) || 
 			    (eclass == m_class_get_parent (mono_defaults.enum_class)) ||
-			    (eclass == mono_defaults.object_class)) {
+			    (!m_class_is_valuetype (eclass))) {
 				*result = FALSE;
 				return;
 			}
+		}
+
+		if (mono_class_is_nullable (eclass) ^ mono_class_is_nullable (eoclass)) {
+			*result = FALSE;
+			return;
 		}
 
 		mono_class_is_assignable_from_checked (eclass, eoclass, result, error);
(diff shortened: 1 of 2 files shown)


EgorBo:
@monojenkins build failed

EgorBo:
@monojenkins build failed

lambdageek:
@EgorBo I think we should also add a case for generic type variables (and generic method variables).  The following example prints `True` and `False` on CoreCLR but `True` and `True` on Mono:

```csharp
using System;
using System.Reflection;

public interface I {
	void M1 ();
}

public class G<T> where T : struct, I {
}

public class Genny {
	public static void Main ()
	{
		var t = typeof (G<>).GetTypeInfo().GenericTypeParameters [0];

		var ta = t.MakeArrayType ();

		var i = typeof (I).GetTypeInfo ();

		var ia = i.MakeArrayType ();

		Console.WriteLine (ta);

		Console.WriteLine (i.IsAssignableFrom (t));

		Console.WriteLine (ia.IsAssignableFrom (ta));
	}
}
```

lambdageek:
Hm. Actually my example mono returns the wrong thing even without the `struct` constraint. I'll open a separate issue

EgorBo:
@lambdageek do you mind if we do it in a separate PR?

EgorBo:
@monojenkins build failed

lambdageek:
@EgorBo yea - sorry I wasn't clear, I agree. We don't need to fix #15647 in this PR.

PR #15642, opened 2019-07-10
Martin tests amazing things

baulig:


baulig:
This is based on the 2019-06 commit that the current integration is using, to test the xamarin-macios side against a stable base.  We'll need to rebase this on top of master before it can land; I'll create a new PR for that when we're done, with a real name.

Issue #15323, opened 2019-06-21
[netcore] Make System.Reflection.Emit.Tests.DynamicMethodToString.ToStringTest Pass


steveisok:
The test currently fails with:

```
System.Reflection.Emit.Tests.DynamicMethodToString.ToStringTest(returnType: typeof(string), parameterTypes: [typeof(string), typeof(int), typeof(System.Reflection.Emit.Tests.TestClass)], expected: "System.String MethodName(System.String, Int32, Sys"...) [FAIL]
      Assert.Equal() Failure
                 ↓ (pos 1)
      Expected: System.String MethodName(System.String, In···
      Actual:   String MethodName(String, Int32, TestClass···
                 ↑ (pos 1)
      Stack Trace:
           at System.Reflection.Emit.Tests.DynamicMethodToString.ToStringTest(Type returnType, Type[] parameterTypes, String expected)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    System.Reflection.Emit.Tests.DynamicMethodToString.ToStringTest(returnType: typeof(string), parameterTypes: [typeof(System.Reflection.Emit.Tests.GenericClass<>)], expected: "System.String MethodName(System.Reflection.Emit.Te"...) [FAIL]
      Assert.Equal() Failure
                 ↓ (pos 1)
      Expected: System.String MethodName(System.Reflection···
      Actual:   String MethodName(GenericClass`1)
                 ↑ (pos 1)
      Stack Trace:
           at System.Reflection.Emit.Tests.DynamicMethodToString.ToStringTest(Type returnType, Type[] parameterTypes, String expected)
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Looks like we need to put System on String :-).  Try to make the test pass.

EgorBo:
Filed https://github.com/dotnet/corefx/issues/39360

MaximLipnin:
Fixed in https://github.com/mono/mono/pull/16783

PR #15645, opened 2019-07-10
[netcore] Fix DynamicMethodToString test

EgorBo:
Fixes https://github.com/mono/mono/issues/15323

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -953,15 +953,6 @@
 -nomethod System.Reflection.Tests.AssemblyTests.LoadFromStream_Location_IsEmpty
 
 ####################################################################
-##  System.Reflection.Metadata.Tests
-####################################################################
-
-# Expected null, got a double byte array
-# https://github.com/mono/mono/issues/15193
--nomethod System.Reflection.Metadata.Tests.AssemblyDefinitionTests.ValidateAssemblyNameWithCultureSet
--nomethod System.Reflection.Metadata.Tests.AssemblyDefinitionTests.ValidateAssemblyNameForAssemblyDefinition
-
-####################################################################
 ##  System.Reflection.TypeExtensions.CoreCLR.Tests
 ####################################################################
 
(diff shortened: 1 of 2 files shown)


netcore/System.Private.CoreLib/src/System.Reflection.Emit/DynamicMethod.cs
```diff
@@ -316,14 +317,25 @@ internal override int GetParametersCount ()
...
 			}
-			return ReturnType.Name+" "+Name+"("+parms+")";
+
+			string returnTypeName;
+			if (ReturnType == typeof (void))
+				returnTypeName = "Void";
+			else if (ReturnType.IsPrimitive)
+				returnTypeName = ReturnType.Name;
+			else
+				returnTypeName = ReturnType.ToString ();
```
filipnavara: CoreCLR also has exception for `TypedReference` (https://github.com/dotnet/coreclr/blob/2ec0bcfefe0bccb9a7023b8d528c775349d93458/src/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs#L3904). Do we want to special case that one too?

netcore/System.Private.CoreLib/src/System.Reflection.Emit/DynamicMethod.cs
```diff
@@ -316,14 +317,25 @@ internal override int GetParametersCount ()
...
+			var paramsSb = new StringBuilder ();
+			ParameterInfo [] parameters = GetParametersInternal ();
+			for (int i = 0; i < parameters.Length; ++i) {
+				ParameterInfo p = parameters [i];
 				if (i > 0)
-					parms = parms + ", ";
-				parms = parms + p [i].ParameterType.Name;
+					paramsSb.Append (", ");
+
+				paramsSb.Append (p.ParameterType.IsPrimitive ? p.ParameterType.Name : p.ParameterType.ToString ());
```
filipnavara: Same nit about `TypedReference` as below. ByRef handling is also missing (https://github.com/dotnet/coreclr/blob/9773db1e7b1acb3ec75c9cc0e36bd62dcbacd6d5/src/System.Private.CoreLib/src/System/Reflection/MethodBase.CoreCLR.cs#L73-L77) but the usability of that code is questionable.

netcore/System.Private.CoreLib/src/System.Reflection.Emit/DynamicMethod.cs
EgorBo: Good catch!

netcore/CoreFX.issues.rsp
```diff
@@ -952,15 +952,6 @@
 # https://github.com/mono/mono/issues/15192
 -nomethod System.Reflection.Tests.AssemblyTests.LoadFromStream_Location_IsEmpty
 
-####################################################################
-##  System.Reflection.Metadata.Tests
-####################################################################
-
-# Expected null, got a double byte array
-# https://github.com/mono/mono/issues/15193
```
vargaz: Unrelated change ?

EgorBo: yeah, I'll revise all `-nomethod` in a separate PR

EgorBo:
Closing it - going to move needed pieces from CoreCLR to shared and re-use them after.

PR #15651, opened 2019-07-10
Misc AIX/PASE tweaks

NattyNarwhal:
Minor changes/patches I had sitting around for a few weeks, mostly preparing for things to be done in the future. Basically:

1. Set up the infrastructure for sigaltstack. Turns out it had similar limitations to macOS. This is useful in the future, but don't enable it just yet. There's a few deadlocks in the crash reporter with it on that I'd want to look at before turning it on by default..

2. Set up stuff for i getifaddrs equivalent. This will be useful for NetworkInface when it gets converted to CoreFX. Unfortunately for me, none of the codepaths using this are actually compiled in out of the box, because SIOCGIFADDR is hoisted above getiffaddrs in `networking-posix`, which may not be preferred, and the other usage is only for macOS/FreeBSD (though I haven't determined if a similar problem exists for i too).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -525,7 +525,9 @@ case "$host" in
 		with_gc=sgen
 		need_link_unlink=yes
 		use_sigposix=yes
-		dnl the valloc call to alloc the guard page bombs out, even with extending the data area
+		dnl Similar limitation to macOS about the first thread and the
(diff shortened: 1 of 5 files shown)


vargaz:
@monojenkins build failed

PR #15656, opened 2019-07-10
[sdb] Prevent runtime crash when accessing invalid pointer value

### LINKED ISSUE #15612: [sdb] Runtime crash when accessing invalid address under debugging
Unzip the following repro:

[reprosdb6.zip](https://github.com/mono/mono/files/3369391/reprosdb6.zip)

`./build.sh && mono debugger.exe`

Will result in a VM crash.

In this scenario, we're getting a `PointerValue` at an invalid address, and we're trying to dereference the pointer.

The debugger runtime should protect against a crash (~~could pointer evaluation be moved to managed?~~), and return an error to the debugger client access akin to an `AccessViolationException`.

jbevain:
Fix #15612

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
+++ b/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
@@ -1984,7 +1984,18 @@ public class Tests : TestsBase, ITest2
 	}
 
 	public static unsafe void pointer_arguments (int* a, BlittableStruct* s) {
+		int i = 12;
+		int *j = &i;
+		int *p = (int*)0x00000007;
+		int *np = (int*)0;
 		*a = 0;
+
+		pointer_get_value ();
+	}
+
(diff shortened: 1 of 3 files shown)


jbevain:
Ok so reading a value byte per byte like that is probably not ideal, but at least it fully validates that the debugger client is not going to take down the debugger server. Open to suggestions, but at least it avoids introducing a new method in the BCL that we would depend on to access from managed an entire address space.

The worst case here is a large struct where that we'd check byte per byte.

jbevain:
Also I'm trying to fix this without bumping the debugger protocol.

jbevain:
Discusse this offline with @vargaz. That implementation could trigger potential problems by running managed code on the debugger thread.

For instance, if the client has requested to handle exceptions, or if the user managed to put a breakpoint in Marshal.ReadByte, this will deadlock.

The suggestion is to move the checks to the client prior to dereferencing the address.

PR #15658, opened 2019-07-11
[netcore] Make System.Linq.Expressions.Tests.BindTests.GlobalField Pass

### LINKED ISSUE #14917: [netcore] Make System.Linq.Expressions.Tests.BindTests.GlobalField Pass

```
    System.Linq.Expressions.Tests.BindTests.GlobalField [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.ArgumentException)
      Actual:   (No exception was thrown)
      Stack Trace:
           at System.AssertExtensions.Throws[ArgumentException](String paramName, Func`1 testCode)
           at System.Linq.Expressions.Tests.BindTests.GlobalField()
```

MaximLipnin:
`Expression.Bind` should throw `ArgumentException` if called with `FieldInfo` for a global field. The exception is thrown from validation of the (null) `DeclaringType` (see https://github.com/dotnet/corefx/pull/15318).
Mono returns `"<Module>"` in this case and doesn't throw so the test fails. 

Fixes #14917 .

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -88,10 +88,6 @@
 # https://github.com/mono/mono/issues/14912
 -nomethod System.Linq.Expressions.Tests.ArrayBoundsTests.NewArrayBounds*
 
-# Expected exception to be thrown.  None is
-# https://github.com/mono/mono/issues/14917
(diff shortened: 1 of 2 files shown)


MaximLipnin:
@monojenkins build failed

Issue #14913, opened 2019-06-10
[netcore] Make System.Linq.Expressions.Tests.ConvertCheckedTests.ConvertCheckedNullableFloatToULongTest Pass


steveisok:
The test crashes the process.  See https://gist.github.com/steveisok/c27463427034160a44117aef10901f78 for more info.

Determine why the test crashes, try to fix it, and make the test pass.

akoeplinger:
The test was reenabled by https://github.com/mono/mono/pull/15664.

PR #15664, opened 2019-07-11
[netcore] Revise items in CoreFX.issues.rsp

EgorBo:


netcore/CoreFX.issues.rsp
```diff
@@ -984,10 +923,6 @@
 ##  System.Threading.Overlapped.Tests
 ####################################################################
 
-# Overflow Exception
-# https://github.com/mono/mono/issues/15197
--nomethod OverlappedTests.PropertyTest2
```
filipnavara: Still fails on non-Windows, as expected.

netcore/System.Private.CoreLib/src/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -409,6 +409,16 @@ public static void WriteInt64(Object ptr, int ofs, long val)
 			throw new PlatformNotSupportedException ();
 		}
 
+		public static int GetHRForException(Exception? e)
```
marek-safar: Why do you need this, the methods already exist

EgorBo: it exists in `Marshal.NoCom.cs` for unix. so ideally these methods should be moved to `Marshal.Windows.cs` but I'll rollback these changes - there is no sense in declaring `FeatureCominterop` and not implementing it

steveisok:
@monojenkins build failed

MaximLipnin:
Seems that https://github.com/mono/mono/issues/14913 can be closed now

Issue #15092, opened 2019-06-13
[netcore] Make System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_Decimal_ReturnsExpected Pass


steveisok:
The test currently fails with:

```
System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_Decimal_ReturnsExpected [FAIL]
      Assert.Equal() Failure
      Expected: 96
      Actual:   88
      Stack Trace:
           at System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_Decimal_ReturnsExpected()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Figure out why the offsets are different and make it pass.[

EgorBo:
Repro:
```csharp
using System;
using System.Runtime.InteropServices;

class Program
{
    static void Main(string[] args)
    {
        Console.WriteLine(Marshal.SizeOf<WithDecimal>());   // prints: 24 (on Mono: 20)
        Console.WriteLine(Marshal.SizeOf<WithMyDecimal>()); // prints: 20 (on Mono: 20)
    }
}

[StructLayout(LayoutKind.Sequential)]
public struct WithDecimal
{
    public decimal d; // 16 byte = 4 int * 4
    public byte b;    // 1 byte
}

[StructLayout(LayoutKind.Sequential)]
public struct WithMyDecimal
{
    public MyDecimal d; // 16 byte = 4 int * 4
    public byte b;      // 1 byte
}

[StructLayout(LayoutKind.Sequential)]
public struct MyDecimal
{
    public int a;
    public int b;
    public int c;
    public int d;
}
```
When there is a decimal field in a struct, the alignment on x64 should be 8 bytes.

Issue #15091, opened 2019-06-13
[netcore] Make System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NonRuntimeField_ThrowsArgumentException Pass


steveisok:
The test currently fails with:

```
System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NonRuntimeField_ThrowsArgumentException [FAIL]
      Assert.Equal() Failure
                ↓ (pos 0)
      Expected: fieldName
      Actual:   type
                ↑ (pos 0)
      Stack Trace:
           at System.AssertExtensions.Throws[ArgumentException](String paramName, Func`1 testCode)
           at System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NonRuntimeField_ThrowsArgumentException()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

Issue #15090, opened 2019-06-13
[netcore] Make System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NullFieldName_ThrowsArgumentNullException Pass


steveisok:
The test currently fails with:

```
System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NullFieldName_ThrowsArgumentNullException [FAIL]
      Assert.Equal() Failure
      Expected: (null)
      Actual:   fieldName
      Stack Trace:
           at System.AssertExtensions.Throws[ArgumentNullException](String paramName, Func`1 testCode)
           at System.Runtime.InteropServices.Tests.OffsetOfTests.OffsetOf_NullFieldName_ThrowsArgumentNullException()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

PR #15666, opened 2019-07-11
[netcore] decimal should have alignment=8

EgorBo:
Fixes https://github.com/mono/mono/issues/15092 (see my comment there)
Fixes https://github.com/mono/mono/issues/15091
Fixes https://github.com/mono/mono/issues/15090

btw, do I need to wrap minor changes with `#if ENABLE_NETCORE`? such as different strings in exceptions.

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -5272,12 +5272,12 @@ ves_icall_System_Runtime_InteropServices_Marshal_OffsetOf (MonoReflectionTypeHan
 		return 0;
 	}
 	if (MONO_HANDLE_IS_NULL (field_name)) {
-		mono_error_set_argument_null (error, "fieldName", "");
+		mono_error_set_argument_null (error, NULL, "");
 		return 0;
 	}
 
 	if (!m_class_is_runtime_type (MONO_HANDLE_GET_CLASS (ref_type))) {
-		mono_error_set_argument (error, "type", "");
+		mono_error_set_argument (error, "fieldName", "");
 		return 0;
 	}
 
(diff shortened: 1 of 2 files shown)


vargaz:
Preferably yes, we don't want to break non-netcore by accident.

mono/metadata/marshal.c
```diff
@@ -5935,9 +5943,17 @@ mono_marshal_type_size (MonoType *type, MonoMarshalSpec *mspec, guint32 *align,
 		klass = mono_class_from_mono_type_internal (type);
 		if (klass == mono_defaults.object_class &&
 			(mspec && mspec->native == MONO_NATIVE_STRUCT)) {
-		*align = 16;
-		return 16;
+			*align = 16;
+			return 16;
+		} 
+#if ENABLE_NETCORE
+		else if (strcmp (m_class_get_name_space (klass), "System") == 0 &&
```
marek-safar: what does assume this alignment? wouldn't be better to use .Pack ?

EgorBo: @marek-safar If in a struct there is a decimal field - the minimal alignment should be 8 bytes at least (on x64).
When Pack is set - this logic is ignored. e.g.:
```csharp
[StructLayout(LayoutKind.Sequential)]
public struct WithDecimal
{
    public decimal d; 
    public byte b;
} // Size: 24 (Mono used to return 20 before my fix)

[StructLayout(LayoutKind.Sequential, Pack = 1)]
public struct WithDecimal
{
    public decimal d;
    public byte b;
} // Size: 17 
```

EgorBo: I don't know the story behind it in .NET Core, probably the last int field is used as long in some cases or some compatibility issues

jaykrell:
```
typedef struct tagDEC {
    USHORT wReserved;
    union {
        struct {
            BYTE scale;
            BYTE sign;
        };
        USHORT signscale;
    };
    ULONG Hi32;
    union {
        struct {
#ifdef _MAC
            ULONG Mid32;
            ULONG Lo32;
#else
            ULONG Lo32;
            ULONG Mid32;
#endif
        };
        ULONGLONG Lo64;
    };
} DECIMAL;
```

EgorBo:
@monojenkins build failed

steveisok:
@EgorBo I can't tell from the comments, is this ready to go?

EgorBo:
@steveisok yeah I think so

steveisok:
/azp run

@monojenkins build failed

steveisok:
@monojenkins build failed

Issue #14839, opened 2019-06-06
[xamarin] TimeZoneInfo.Local always returns "Local" name


filipnavara:
## Steps to Reproduce

1.  `Console.WriteLine(TimeZoneInfo.Local.Id);`

### Current Behavior

Always returns "Local".

### Expected Behavior

Returns the name of the current time zone.

## On which platforms did you notice this

[x] macOS
[x] iOS
[ ] Linux
[ ] Windows

**Version Used**:

5.18.1, 6.0, 6.5 (master)

## Description

Essentially this is the same issue as https://github.com/mono/mono/issues/8267. It was fixed on desktop but it remains an issue on Xamarin macOS/iOS platforms due to this code:

https://github.com/mono/mono/blob/515234a128318cc199df0cca9cccdbf06fe72fdb/mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs#L49-L51

artur-malendowicz-gain:
@jonpryor any ETA for this?

PR #15667, opened 2019-07-11
[ios] Ensure Local TimeZone Id is Set For TimeZoneInfo.Local 

steveisok:
Fixes https://github.com/mono/mono/issues/14839

The local timezone name was not known and defaulted to "Local". 
This change adds an sdk function that returns the local name and sets TimeZoneInfo.Local accordingly.

### DIFF
--- a/mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs
+++ b/mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs
@@ -71,7 +71,7 @@ namespace System {
 				}
 			}
 		}
-				
+
 		[DllImport ("__Internal")]
 		extern static IntPtr xamarin_timezone_get_names (ref int count);
 
--- a/mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs
+++ b/mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs
@@ -44,10 +44,13 @@ namespace System {
 
 	public partial class TimeZoneInfo {
 
+		[DllImport ("__Internal")]
+		extern static string xamarin_timezone_get_local_name ();
+
 		static TimeZoneInfo CreateLocal ()
 		{
 			using (Stream stream = GetMonoTouchData (null)) {
-				return BuildFromStream ("Local", stream);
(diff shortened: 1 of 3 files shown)


mcs/class/corlib/Test/System/TimeZoneInfoTest.cs
```diff
@@ -115,6 +115,12 @@ public void GetLocal ()
 				Assert.IsTrue (true);
 			}
 
+			[Test]
+			public void LocalIdCheck ()
+			{				
+				Assert.IsTrue (TimeZoneInfo.Local.Id != "Local", "Local timezone id should not be \"Local\"");
```
akoeplinger: I dug a bit deeper and I think this will fail in cases where we read the local time zone from `/etc/localtime`, like https://github.com/mono/mono/blob/2f1b50bfbf9db129d844dad8e45ec248f0c877e4/mcs/class/corlib/System/TimeZoneInfo.cs#L146-L147.

I assume that was the reason why we added the `readlink()` check to the other test.

So instead of adding this new test we can reuse the `LocalId` test and simply ifdef out the readlink on `#if MONOTOUCH || XAMMAC`

mcs/class/corlib/System/TimeZoneInfo.MonoTouch.cs
```diff
@@ -68,7 +71,7 @@ static void GetSystemTimeZonesCore (List<TimeZoneInfo> systemTimeZones)
 				}
 			}
 		}
-		
+
```
akoeplinger: unrelated change

sdks/ios/runtime/runtime.m
```diff
@@ -355,6 +359,20 @@
 	return result;
 }
 
+//
+// Returns the geopolitical region ID of the local timezone.
+// Previously we just provided the data to TimeZoneInfo.MonoTouch.cs and that
```
akoeplinger: nit: a few years from now readers of this code won't care what we used to do a long time ago (that is what git history is for), I'd remove this sentence.

akoeplinger:
iOS jobs passed (which is the only thing touched by this), merging.

akoeplinger:
@monojenkins backport 2019-06

PR #15668, opened 2019-07-11
[netcore] Use the strict loader

lambdageek:
Mono historically was quite lax in how it resolved assembly references - it
would pick the first assembly with a matching simple name.  The "strict"
mode (available in framework Mono with `--assembly-loader=strict`) follows the
spec behavior: the version and public key token of the candidate assembly must
match what is requested.

### DIFF
--- a/mono/mini/main-core.c
+++ b/mono/mini/main-core.c
@@ -227,6 +227,12 @@ int STDAPICALLTYPE coreclr_initialize (const char* exePath, const char* appDomai
 	if (native_lib_paths != NULL)
 		mono_set_pinvoke_search_directories (native_lib_paths->dir_count, native_lib_paths->dirs);
 
+	/*
+	 * Don't use Mono's legacy assembly name matching behavior - respect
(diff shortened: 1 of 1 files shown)


lambdageek:
I guess this is wrong, actually.
We need a third policy - see https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-version

Policies:
* Strict - version and public key token has to match exactly
* Mono legacy - yolo - if the simple name matches, just load it and ignore version and tokens.
* Core - version (and public key token?) are checked - higher version than what's requested is allowed.

lambdageek:
We still need this - see https://github.com/mono/mono/issues/15669#issuecomment-510932593

lewing:
@monojenkins build failed

lambdageek:
@monojenkins build failed

lewing:
@monojenkins build failed

lambdageek:
@monojenkins build failed

PR #15671, opened 2019-07-11
[cominterop] Always marshal a class as a COM interface if it has the InterfaceType attribute set.

zfigura:
This fixes Game Maker Studio 2 Desktop under Wine.

### DIFF
--- a/mono/metadata/cominterop.c
+++ b/mono/metadata/cominterop.c
@@ -3756,3 +3756,32 @@ mono_cominterop_get_com_interface_internal (gboolean icall, MonoObjectHandle obj
 	g_assert_not_reached ();
 #endif
 }
+
+gboolean
+mono_cominterop_is_interface (MonoClass* klass)
+{
+#ifndef DISABLE_COM
+	ERROR_DECL (error);
+	MonoCustomAttrInfo* cinfo = NULL;
+	gboolean ret = FALSE;
+	int i;
+
+	cinfo = mono_custom_attrs_from_class_checked (klass, error);
+	mono_error_assert_ok (error);
+	if (cinfo) {
+		for (i = 0; i < cinfo->num_attrs; ++i) {
+			MonoClass *ctor_class = cinfo->attrs [i].ctor->klass;
+			if (mono_class_has_parent (ctor_class, mono_class_get_interface_type_attribute_class ())) {
+				ret = TRUE;
+				break;
+			}
+		}
+		if (!cinfo->cached)
+			mono_custom_attrs_free (cinfo);
+	}
+
+	return ret;
+#else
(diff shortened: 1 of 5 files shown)


marek-safar:
approve

marek-safar:
@vargaz @lambdageek could you please review

mono/metadata/cominterop.c
```diff
@@ -3756,3 +3756,26 @@ mono_cominterop_get_com_interface_internal (gboolean icall, MonoObjectHandle obj
...
+#ifndef DISABLE_COM
+	ERROR_DECL (error);
+	MonoCustomAttrInfo* cinfo = NULL;
+	MonoInterfaceTypeAttribute* itf_attr = NULL;
+
+	cinfo = mono_custom_attrs_from_class_checked (klass, error);
+	mono_error_assert_ok (error);
+	if (cinfo) {
+		itf_attr = (MonoInterfaceTypeAttribute*)mono_custom_attrs_get_attr_checked (cinfo, mono_class_get_interface_type_attribute_class (), error);
+		mono_error_assert_ok (error); /*FIXME proper error handling*/
```
lambdageek: What do we want proper error handling to be?  Can we just return false if trying to look for the custom attribute fails?  In that case could do something like this:

```c
if (!is_ok (error)) {
    mono_error_cleanup (error);
    itf_attr = NULL;
}
```

zfigura: I essentially just copy-pasted that part from cominterop_get_com_slot_begin (admittedly, I should have at least separated it out into a common helper). I wasn't really sure what proper error handling would be, but that seems reasonable enough. I'll resend the patch.

lambdageek:
@monojenkins build failed

marek-safar:
@monojenkins build

lewing:
@monojenkins build failed

akoeplinger:
This seems to cause two test failures, one during `aot-test` when trying to AOT System.Data.dll and one during `fullaot-mixed` which crashes with this error when AOT-ing System.dll:

```
Invoking method 'System.Runtime.InteropServices.InterfaceTypeAttribute:.ctor (System.Runtime.InteropServices.ComInterfaceType)' when running in no-exec mode.
```

zfigura:
Indeed, thanks. In `mono_marshal_get_native_wrapper()` I see a comment "The COM code is not AOT compatible, it calls mono_custom_attrs_get_attr_checked ()". Is there by any chance a simple explanation for why this is the case?

zfigura:
Okay, thanks to some hints from Zoltan Varga, I believe I've figured out the right way to resolve this. I've pushed a new version that avoids using `mono_custom_attrs_get_attr_checked()`.

PR #15674, opened 2019-07-12
Update Haiku support

NattyNarwhal:
* Build System.Native.

* Disable `pthread_mutexattr_setprotocol` use, since it returns EINVAL on Haiku.

* Disable the `strerror` cache on Haiku. On Haiku, errors begin at `0x8000_0000` like a Win32 HRESULT. This means the indexing would be on negative numbers, which is a no-no. There are a few alternative ways to keep it though: hashtable, tuple array, offset...

Unfortunately, doesn't build all the way through. Has a tendency to bomb out during building `build-unix`. Will investigate why.


mono/eglib/gstr.c
```diff
@@ -205,6 +205,13 @@ Last this was checked was June-2017.
 
 Apple is at 106.
 Android is at 133.
+
+Haiku starts numbering at 0x8000_7000 (like HRESULT on Win32) for POSIX errno,
```
jaykrell: HRESULTs do not start at 0x8000_7000. Please remove bogus comment.
HRESULTs span the entire 32bit range.

mono/eglib/gstr.c
```diff
@@ -213,6 +220,14 @@ Android is at 133.
 static pthread_mutex_t strerror_lock = PTHREAD_MUTEX_INITIALIZER;
 #endif
 
+#if defined(__HAIKU__)
```
jaykrell: I suggest `#if __HAIKU__` for brevity.

mono/eglib/gstr.c
```diff
@@ -213,6 +220,14 @@ Android is at 133.
...
+const gchar *
```
jaykrell: I suggest `gchar` is legacy, use regular `char`.

mono/eglib/gstr.c
```diff
@@ -213,6 +220,14 @@ Android is at 133.
...
+g_strerror (gint errnum)
+{
+	/* returns a const char* on Haiku */
+	return strerror(errnum);
```
jaykrell: Space before paren?

mono/eglib/gstr.c
```diff
@@ -213,6 +220,14 @@ Android is at 133.
 static pthread_mutex_t strerror_lock = PTHREAD_MUTEX_INITIALIZER;
 #endif
 
+#if defined(__HAIKU__)
+const gchar *
+g_strerror (gint errnum)
+{
+	/* returns a const char* on Haiku */
```
jaykrell: Comment doesn't seem useful.

jaykrell: Ok slightly useful, as it isn't const on other platforms.

mono/eglib/gstr.c
NattyNarwhal: Comment is unclear. It starts at `0x8000_0000` but POSIX errno begins at the +`0x7000` region.

jaykrell: HRESULT starts at 8000000 just because it is 32bit signed integer. But comparing Haiku errno to Win32 HRESULT does not seem useful.

Issue #15669, opened 2019-07-11
[netcore] Add a netcore loader policy


lambdageek:
We need a third loader policy - see https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-version

Policies:
* Strict - version and public key token has to match exactly
* Mono legacy - yolo - if the simple name matches, just load it and ignore version and tokens.
* **new** netcore - version is checked - higher version than what's requested is allowed. Public key tokens are ignored.

lambdageek:
@jkotas - any subtleties I should be aware of about the netcore assembly version policy?

jkotas:
.NET Core assembly loader throws exception by default when the available assembly version is lower then the requested assembly version.

Public key tokens are ignored.

lambdageek:
More details from chat:

> > so what happens if a lower version was already correctly loaded then a higher version is requested later?
> 
> The load fails. We only allow loading one version for given simple name (in each assembly load context).

This is a good policy.  I think we might need a little bit of work to make Mono behave this way: I think our "is this assembly already loaded" check can't cause the whole load request to fail - it can only say "yes it's already loaded, return it" or "no, it's not already loaded, go probe the filesystem".

filipnavara:
This is already implemented - https://github.com/mono/mono/blob/18ce65575142cc10c3313f3a8cb281e5c5c17d61/mono/metadata/assembly.c#L4297-L4301. Can you be more specific about what is still missing?

lambdageek:
Okay, so to summarize:

1. we have code in [`exact_sn_match`](https://github.com/mono/mono/blob/18ce65575142cc10c3313f3a8cb281e5c5c17d61/mono/metadata/assembly.c#L4295) to do the netcore-policy version check.
2. #15678 enhances `exact_sn_match` to ignore public key tokens.
3. we have code in [`mono_core_preload_hook`](https://github.com/mono/mono/blob/643b2ebfc6ad8a7cb2c21d3cac1a89ca3ceb3c9f/mono/mini/main-core.c#L124) to use the version check predicate when trying to open trusted platform assemblies.
4. After #15668 we will use the version check predicate in other places in the loader (for example loading non-TPA assemblies using reflection, etc).

That should be everything.

CoffeeFlux:
Everything got merged, so I think we're good here.

PR #15678, opened 2019-07-12
[netcore] Ignore public key when matching assemblies

filipnavara:
Related to #15669


filipnavara:
It is hit in practice and it was hit early on when we started running the CoreFX tests.

lambdageek:
Ah right... I forgot that we use the strong name matching in the netcore preload hook:
https://github.com/mono/mono/blob/202dca075b5110398a77e659e1bc128e6eda51e5/mono/mini/main-core.c#L123-L131

lewing:
@monojenkins build failed

PR #15680, opened 2019-07-12
[ios] Ensure Local TimeZone Id is Set For TimeZoneInfo.Local

steveisok:
Expands on https://github.com/mono/mono/pull/15667 by adding a null check on the name.


steveisok:
Nevermind on this one... Thought I could get away w/ the same branch.

PR #15682, opened 2019-07-12
[ios] Nil Check to Ensure Local TimeZone Id is Set For TimeZoneInfo.Local

steveisok:
In runtime.m added nil check in xamarin_timezone_get_local_name. Returns Local as the Id in the case of nil.


steveisok:
@monojenkins backport 2019-06

PR #15693, opened 2019-07-15
[corlib] Foward System.Type.GetTypeFromProgID() to the correct function.

zfigura:
Fixes a stack overflow when running Staxel under wine-mono.

### DIFF
--- a/mcs/class/corlib/Test/System/TypeTest.cs
+++ b/mcs/class/corlib/Test/System/TypeTest.cs
@@ -1950,6 +1950,29 @@ namespace MonoTests.System
 			else
 				throw new COMException ();
 		}
+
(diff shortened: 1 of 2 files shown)


marek-safar:
approve

steveisok:
@monojenkins build failed

akoeplinger:
The fix looks good but the new test fails with `System.NotImplementedException : Unmanaged activation is not supported`. I assume this exception is caught by your app?

I'd be fine with merging the patch if we disable the test or catch the NotImplementedException.

zfigura:
Indeed, my apologies for missing that. (I think the corlib test failed to run to completion on my machine, and so I didn't catch it while testing.)

PR #15695, opened 2019-07-15
[interp] Make the code GC safe on wasm.

vargaz:


BrzVlad:
@vargaz Could we keep the volatility of `o` and indirection to `frame->o` just for TARGET_WASM ?

vargaz:
Sure, but it would look a big ugly. Also, I don't think this change will hurt perf too much, its just one memory load from the stack.

BrzVlad:
For the local obj maybe we can define something like local_obj to be either o or frame->o, while for the volatility is straight forward. Shouldn't impact the readability too much.

vargaz:
Long term, we should centralize the handling of objrefs which would make it possible to do precise stack scanning in the interpreter, its much easier to do than in the jit.

mono/mini/interp/interp-internals.h
```diff
@@ -137,7 +137,11 @@ struct _InterpFrame {
 	 * Storing into this field will keep the object pinned
 	 * until the objref can be stored into stackval->data.o.
 	 */
+#ifdef TARGET_WASM
 	MonoObject* volatile o;
+#else
```
BrzVlad: This should go also for `stackval` structure. Well I guess only there, since `o` is dead here.

vargaz:
@monojenkins build failed

lewing:

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #15699, opened 2019-07-16
There is no need to zero static data, and it is racy.

jaykrell:
if there are multiple threads calling (unlikely).
This zeroing was made moot by 470354fa817716739615ff0c6cfefca42ec8b385 January 2019.

### DIFF
--- a/mono/utils/mono-merp.c
+++ b/mono/utils/mono-merp.c
@@ -49,7 +49,9 @@ os_version_string (void)
 	static const char *version_string;
 
 	if (!version_string) {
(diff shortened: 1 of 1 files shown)


CoffeeFlux:
@monojenkins build failed

jaykrell:
Note that the static data is still potentially racy, depending on if ‘uname’ writes just once. For example I think WSL zeros it all (just like this) and then copies the strings.

jaykrell:
Rebased to master.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

Issue #15340, opened 2019-06-21
[netcore] Make System.Reflection.Tests.CustomAttributeTests.TestDllImportPseudoCustomAttribute Pass


steveisok:
The test currently fails with:

```
System.Reflection.Tests.CustomAttributeTests.TestDllImportPseudoCustomAttribute [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object
      Stack Trace:
           at System.Reflection.CustomAttributeData.get_AttributeType()
           at System.Reflection.Tests.CustomAttributeInstantiator.Instantiate[DllImportAttribute](CustomAttributeData cad)
           at System.Reflection.Tests.TestUtils.UnprojectAndInstantiate[DllImportAttribute](CustomAttributeData cad)
           at System.Reflection.Tests.CustomAttributeTests.TestDllImportPseudoCustomAttribute()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

lambdageek:
This is preventing `MetadataLoadContext` from working properly with netcore Mono. Mono's `CustomAttributeData` implementation is different from CoreCLR and `MetadataLoadContext` subclasses `CustomAttributeData` and doesn't call our internal constructors.

grendello:
If MLC worked on Mono, it could greatly improve performance on mobile devices in situations when we can't avoid using reflection (not just to retrieve custom attributes) for various reasons.

lewing:
@mrward this is the mono issue

Issue #15339, opened 2019-06-21
[netcore] Make System.Reflection.Tests.CustomAttributeTests.CustomAttributeNamedArguments_Field Pass


steveisok:
The test currently fails with:

```
System.Reflection.Tests.CustomAttributeTests.CustomAttributeNamedArguments_Field [FAIL]
      System.InvalidOperationException : Sequence contains no elements
      Stack Trace:
           at System.Linq.ThrowHelper.ThrowNoElementsException()
           at System.Linq.Enumerable.Single[CustomAttributeData](IEnumerable`1 source)
           at System.Reflection.Tests.CustomAttributeTests.CustomAttributeNamedArguments_Field()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

MaximLipnin:
I mentioned this test with `-method System.Reflection.Tests.CustomAttributeTests.CustomAttributeNamedArguments_Field` and it passes locally on macOS.

steveisok:
OK, let's take it out of the rsp file and see if it passes CI

MaximLipnin:
It was checked in https://github.com/mono/mono/pull/15700.

PR #15700, opened 2019-07-16
[netcore] Clean up rsp-file

MaximLipnin:
`System.Reflection.Tests.CustomAttributeTests.CustomAttributeNamedArguments_Field` test from https://github.com/mono/mono/issues/15339 passes locally with latest master on macOS.

`System.Reflection.Tests.CustomAttributeTests.TestDllImportPseudoCustomAttribute` and 
`System.Reflection.Tests.CustomAttributeTests.TestMarshalAsPseudoCustomAttribute` fail with the same exception (see https://github.com/mono/mono/issues/15340).

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -1119,11 +1119,9 @@
 ##  System.Reflection.MetadataLoadContext.Tests
 ####################################################################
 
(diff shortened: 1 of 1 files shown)


filipnavara:
Most of the netcore tests failed with the remote executor process failing with SIGUSR2. It looks unrelated to the change but should be investigated.

PR #15707, opened 2019-07-16
Enable compiler server 2019-06

lewing:
Turn on the compiler server by default in 2019-06 for all platforms except darwin


lewing:
I've updated the msbuild hash to pull in what should become mono-2019-06

lewing:
Closing this as we have decided to just enable in msbuild by default

PR #15713, opened 2019-07-16
[ios][android] disable gac in ios and android sdk builds

lewing:


lewing:

lewurm:
by how much does this reduce the size of `libmonosgen.*`?

lambdageek:
@lewurm I don't have numbers, but it's probably not significant.

There's probably a speedup to assembly loading on slow Android devices since we don't go looking for non-existent directories anymore.

Issue #15646, opened 2019-07-10
MERP indefinitely hangs on application quit


Therzok:
## Steps to Reproduce

1. Not available, came in via MERP report: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/942255

## On which platforms did you notice this

[*] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Running on Mono 6.0.0.296 (2019-02/1e9f3bfbc5a) (64-bit)

## Stacktrace

```

Pkilling 0x700008680000 from 0x10bcb85c0
Pkilling 0x700004e23000 from 0x10bcb85c0
Pkilling 0x700002ef0000 from 0x10bcb85c0
Pkilling 0x70000644d000 from 0x10bcb85c0
Pkilling 0x700008c89000 from 0x10bcb85c0
Pkilling 0x7000034f9000 from 0x10bcb85c0
Pkilling 0x700006a56000 from 0x10bcb85c0
Pkilling 0x700003b02000 from 0x10bcb85c0
Pkilling 0x70000512c000 from 0x10bcb85c0
Pkilling 0x70000705f000 from 0x10bcb85c0
Pkilling 0x70000410b000 from 0x10bcb85c0
Pkilling 0x700005735000 from 0x10bcb85c0
Pkilling 0x700007668000 from 0x10bcb85c0
Pkilling 0x700004714000 from 0x10bcb85c0
Pkilling 0x700005d3e000 from 0x10bcb85c0
Pkilling 0x700007c71000 from 0x10bcb85c0
Pkilling 0x700002aea000 from 0x10bcb85c0
Pkilling 0x700008883000 from 0x10bcb85c0
Pkilling 0x70000275e000 from 0x10bcb85c0
Pkilling 0x7000030f3000 from 0x10bcb85c0
Pkilling 0x700006650000 from 0x10bcb85c0
Pkilling 0x7000036fc000 from 0x10bcb85c0
Pkilling 0x700006c59000 from 0x10bcb85c0
Pkilling 0x700003d05000 from 0x10bcb85c0
Pkilling 0x70000532f000 from 0x10bcb85c0
Pkilling 0x700007262000 from 0x10bcb85c0
Pkilling 0x70000430e000 from 0x10bcb85c0
Pkilling 0x700005938000 from 0x10bcb85c0
Pkilling 0x700004917000 from 0x10bcb85c0
Pkilling 0x700005f41000 from 0x10bcb85c0
[...]

PR #15715, opened 2019-07-17
[merp] Use a separate program as the hang supervisor.

alexischr:
Fixes https://github.com/mono/mono/issues/15646

macOS does not like signals being sent from the forked supervisor process, possibly towards anywhere but definitely when sent to the parent process. The following message is currently spewed after the supervisor process attempting to send a SIGSEGV to a hanged Mono process:
"The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec()."

We follow that direction and introduce a new binary that, when available, is used to abort the parent process for us.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -58,6 +58,7 @@
 #include <mono/metadata/exception-internals.h>
 #include <mono/utils/mono-state.h>
 #include <mono/metadata/w32subset.h>
+#include <mono/metadata/mono-config.h>
 
 #ifdef HAVE_SYS_WAIT_H
 #include <sys/wait.h>
@@ -6262,7 +6263,6 @@ mono_threads_summarize_one (MonoThreadSummary *out, MonoContext *ctx)
 	return success;
 }
 
-#define TIMEOUT_CRASH_REPORTER_FATAL 30
 #define MAX_NUM_THREADS 128
 typedef struct {
 	gint32 has_owner; // state of this memory
@@ -6278,7 +6278,7 @@ typedef struct {
 	gboolean silent; // print to stdout
 } SummarizerGlobalState;
 
-#if defined(HAVE_KILL) && !defined(HOST_ANDROID) && defined(HAVE_WAITPID) && ((!defined(HOST_DARWIN) && defined(SYS_fork)) || HAVE_FORK)
+#if defined(HAVE_KILL) && !defined(HOST_ANDROID) && defined(HAVE_WAITPID) && defined(HAVE_EXECVE) && ((!defined(HOST_DARWIN) && defined(SYS_fork)) || HAVE_FORK)
 #define HAVE_MONO_SUMMARIZER_SUPERVISOR 1
 #endif
 
@@ -6289,8 +6289,9 @@ typedef struct {
 } SummarizerSupervisorState;
 
 #ifndef HAVE_MONO_SUMMARIZER_SUPERVISOR
-static void
-summarizer_supervisor_wait (SummarizerSupervisorState *state)
+
+void
+mono_threads_summarize_init (const char *timeline_dir)
 {
 	return;
 }
@@ -6309,23 +6310,14 @@ summarizer_supervisor_end (SummarizerSupervisorState *state)
 }
 
 #else
-static void
-summarizer_supervisor_wait (SummarizerSupervisorState *state)
+static const char *hang_watchdog_path;
+
+void
+mono_threads_summarize_init (const char *timeline_dir)
 {
-	sleep (TIMEOUT_CRASH_REPORTER_FATAL);
-
-	// If we haven't been SIGKILL'ed yet, we signal our parent
-	// and then exit
-#ifdef HAVE_KILL
-	g_async_safe_printf("Crash Reporter has timed out, sending SIGSEGV\n");
-	kill (state->pid, SIGSEGV);
-#else
-	g_error ("kill () is not supported by this platform");
-#endif
-
-	exit (1);
+	hang_watchdog_path = g_build_filename (mono_get_config_dir (), "..", "bin", "mono-hang-watchdog", NULL);
+	mono_summarize_set_timeline_dir (timeline_dir);
 }
-
 static pid_t
 summarizer_supervisor_start (SummarizerSupervisorState *state)
 {
@@ -6352,6 +6344,13 @@ summarizer_supervisor_start (SummarizerSupervisorState *state)
 
 	if (pid != 0)
 		state->supervisor_pid = pid;
+	else {
+		char pid_str[20]; // pid is a uint64_t, 20 digits max in decimal form
+		sprintf (pid_str, "%llu", (uint64_t)state->pid);
+		const char *const args[] = { hang_watchdog_path, pid_str, NULL };
+		execve (args[0], (char * const*)args, NULL); // run 'mono-hang-watchdog [pid]'
+		g_assert_not_reached ();
+	}
 
 	return pid;
 }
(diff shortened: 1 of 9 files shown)


jaykrell:
Is it possible to limit what the supervisor uses, i.e. not use CoreFoundation?
Otherwise, ok, this is a good idea -- fork w/o exec is not a great idea and you are combating that.

mono/metadata/threads.c
```diff
@@ -6350,8 +6326,14 @@ summarizer_supervisor_start (SummarizerSupervisorState *state)
 	g_assert_not_reached ();
 #endif
 
-	if (pid != 0)
+	if (pid != 0) {
 		state->supervisor_pid = pid;
+		char pid_str[20]; // pid is a uint64_t, 20 digits max in decimal form
```
jaykrell: I'd be simpler to just say `65` instead of knowing how many decimal digits a 64bit number can expand to.

tools/mono-hang-watchdog/mono-hang-watchdog.c
```diff
@@ -0,0 +1,50 @@
...
+#include "config.h"
+
+#ifdef HAVE_SIGNAL_H
+#include <signal.h>
+#endif
+
+#define TIMEOUT 30
+
+static char* program_name;
+void program_exit (int exit_code, const char* message);
```
Therzok: Is the forward declaration needed?

alexischr:
@jaykrell The "CoreFoundation" functionality triggering this is actually just `kill()` (or possibly `sleep()`). Nothing else happens in the supervisor forked process.

tools/mono-hang-watchdog/mono-hang-watchdog.c
alexischr: Yup, we use `-Wmissing-prototypes`

mono/metadata/threads.c
```diff
@@ -6350,8 +6326,14 @@ summarizer_supervisor_start (SummarizerSupervisorState *state)
...
+		sprintf (pid_str, "%llu", (uint64_t)state->pid);
+		const char *const args[] = { "mono-hang-watchdog", pid_str };
+		execvp (args[0], (char * const*)args); // run 'mono-hang-watchdog [pid]'
```
lambdageek: Is it reasonable to rely on `$PATH` to find `mono-hang-watchdog`?  Maybe we should make some path relative to the `mono` binary?  (`../libexec/mono-hang-watchdog` perhaps?)

lambdageek: I think this is doing the `exec` in the parent process.

The child process would get `pid == 0` (ie the "false" case of the caller of `summarizer_supervisor_start`).

mono/metadata/threads.c
```diff
@@ -6665,8 +6647,6 @@ mono_threads_summarize (MonoContext *ctx, gchar **out, MonoStackHash *hashes, gb
 				g_assert (mem);
 				success = mono_threads_summarize_execute_internal (ctx, out, hashes, silent, mem, provided_size, TRUE);
 				summarizer_supervisor_end (&synch);
-			} else {
-				summarizer_supervisor_wait (&synch);
```
lambdageek: I think this is the child process - ie this is where the `exec` would need to happen.

tools/mono-hang-watchdog/mono-hang-watchdog.c
```diff
@@ -0,0 +1,50 @@
...
+#include <errno.h>
+#include <unistd.h>
+
+#include "config.h"
+
+#ifdef HAVE_SIGNAL_H
+#include <signal.h>
+#endif
+
+#define TIMEOUT 30
```
lambdageek: Can we use `TIMEOUT_CRASH_REPORTER_FATAL` for this define?  And can we get it from configure.ac?

mono/metadata/threads.c
alexischr: Yup I got it reversed - though the patch will still end up not having `summarizer_supervisor_wait` and the change will be in `summarizer_supervisor_start`

tools/mono-hang-watchdog/mono-hang-watchdog.c
alexischr: `TIMEOUT_CRASH_REPORTER_FATAL` had this one use, so it's actually removed with this patch.

I am not sure we use `configure.ac` for any other such overrides of integer values, if I am understanding you.

lambdageek: Yea, that's a fair point - maybe this doesn't need to be in configure.

I weakly prefer a longer name for the define - I'm concerned a little bit that it'll be difficult to find the place where this is hardcoded - `TIMEOUT_CRASH_REPORTER_FATAL` is maybe a bit easier to grep than just a generic `TIMEOUT` - not sure.

tools/mono-hang-watchdog/mono-hang-watchdog.c
```diff
@@ -0,0 +1,50 @@
...
+    if (errno)
+        program_exit (2, "Invalid pid");
+
+    sleep (TIMEOUT);
+
+    /* if we survived the timeout, we consider the Mono process as hung */
+
+#ifndef HAVE_KILL
+    /* just inform the user */
+    printf ("Mono process with pid %llu appears to have hanged");
```
lambdageek: missing the argument corresponding to `%llu`

mono/utils/mono-state.c
```diff
@@ -135,6 +137,13 @@ file_for_summary_stage (const char *directory, MonoSummaryStage stage, gchar *bu
 	g_snprintf (buff, sizeof_buff, "%s%scrash_stage_%d", directory, G_DIR_SEPARATOR_S, stage);
 }
 
+void
+mono_summarize_init (const char *timeline_dir)
+{
+	hang_watchdog_dir = mono_config_get_bin_dir ();
```
lambdageek: `hang_watchdog_dir` is assigned but not used anywhere.  I think you want a `merp_hang_watchdog_path = g_build_filename (hang_watchdog_dir, "mono-hang-watchdog");` and then use `merp_hang_watchdog_path` in the call to `execv` in `summarizer_supervisor_start`

mono/metadata/threads.c
```diff
@@ -6309,23 +6310,14 @@ summarizer_supervisor_end (SummarizerSupervisorState *state)
...
-	kill (state->pid, SIGSEGV);
-#else
-	g_error ("kill () is not supported by this platform");
-#endif
-
-	exit (1);
+void
+mono_threads_summarize_init (const char *timeline_dir)
+{
+	hang_watchdog_path = g_build_filename (mono_config_get_bin_dir (), "mono-hang-watchdog");
```
lambdageek: This needs to be NULL-terminated, I think.

lambdageek: This doesn't work for in-tree builds when there's no mono-hang-watchdog binary installed with system Mono.

I think we need to set this as a relative directory relative to the value of `mono_get_config_dir()` because that's set to something like "build_root/runtime/etc" for in-tree runs.

1. Use `g_build_filename (mono_get_config_dir (), "..", "bin", "mono-hang-watchdog", NULL)`.

2. Add a `mono/runtime/bin/mono-hang-watchdog.in`:
```
#! /bin/sh
r='@mono_build_root@'
exec "$r/tools/mono-hang-watchdog/mono-hang-watchdog" "$@"
```

3. Add to `configure.ac`
```
AC_CONFIG_FILES([runtime/bin/mono-hang-watchdog],[chmod +x runtime/bin/mono-hang-watchdog])
```

alexischr:
@monojenkins build OS X package

alexischr:
@monojenkins build failed

alexischr:
@monojenkins backport 2019-08

PR #15717, opened 2019-07-17
ld configure: Add another dash since PASE merely warns.

jaykrell:


NattyNarwhal:
Didn't have the opportunity to hit build, but the configure script output looks fine.

jaykrell:
GNU ld accept single or double dash, means the same thing.
AIX ld warns on single dash, errors on double dash, in autoconf, and then errors on single dash later.
We want the double dash so autoconf rejects it on AIX.
Apple presumably is consistent between them but that could use a double check
That CI passes is some confirmation.

PR #15721, opened 2019-07-17
Add com.apple.security.cs.allow-unsigned-executable-memory exception to hardened runtime entitlements

akoeplinger:
This avoids the issue from https://github.com/mono/mono/issues/13445 when hardened runtime is enabled.

### DIFF
--- a/mono/mini/mac-entitlements.plist
+++ b/mono/mini/mac-entitlements.plist
@@ -1,7 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 	<dict>
-		<key>com.apple.security.cs.allow-jit</key><true/>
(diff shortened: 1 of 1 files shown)


akoeplinger:

akoeplinger:
@monojenkins build OS X x64 Hardened Runtime

akoeplinger:
@monojenkins backport 2019-06

filipnavara:
Doesn't this defeat the whole point of using the hardened runtime?

With this entitlement the runtime won't detect it is running under hardened runtime:
https://github.com/mono/mono/blob/a97408ac1e9b1e0dedbd67424ca0b0a6fc7b73b6/mono/utils/mono-mmap.c#L294-L302

akoeplinger:
Yes, it defeats the memory protection of hardened runtime but this is arguably of little use for Mono's desktop use case where we allow running arbitrary assemblies on purpose.

This entitlement will be applied to the system Mono that ships in the .pkg, the embedded Mono used e.g. in Xamarin.Mac will use different entitlements so it has more protection.

We need to enable hardened runtime as Apple requires it for notarization.

PR #15724, opened 2019-07-17
[sdb] Fix TypeMirror.CSharpName computation

sailro:
Fixing two issues in TypeMirror.CSharpName computation (not sure if this property is actively used):
- there is a typo on S**B**yte primitive, preventing System.SByte -> sbyte.
- array rank display is wrong. For example System.Byte[,,,] has GetArrayRank() == 4, so the previous computation gave byte[,,,,] instead of byte[,,,]


sailro:
cc @jbevain

marek-safar:
whitelist

vargaz:
@monojenkins build failed

PR #14166, opened 2019-04-22
Enable disabled DIM test

kg:
cc @marek-safar are there other DIM tests that are still disabled? I couldn't find any

PR #15729, opened 2019-07-17
[mini] Runtime check if boxing is needed for DIM constrained calls

lambdageek:
This supersedes #14166 - enable `dim-constrainedcall.exe` and also make a fix so it works with FullAOT. 

---

This is to handle the following example.  The issue is that in `Check()` we have
a constrained call where the called method is ``IAdder`1<!!U>::PlusPlus()`` which
is a default interface method, and we need to determine at runtime whether to
box the this argument.

```csharp
using System;

public interface IAdder<T> {
	int Add (int x);

	int PlusPlus () {
		return Add (1);
	}
}

interface IGen3<T> { }

struct Adder<T> : IAdder<IGen3<T>> {
	int _field;
	public int Add (int x) {
		_field = x + _field;
		return _field;
	}
}

public class P {
	public static int Check<T, U>(T t) where T : IAdder<U> {
		return t.PlusPlus () + t.PlusPlus ();
	}

	public static void Main () {
		var x = new Adder<object> ();
		int y = Check<Adder<object>, IGen3<object>> (x);
		Console.WriteLine ("expect 2, got {0}", y);
	}
}
```

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -5636,10 +5636,12 @@ handle_constrained_call (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSignat
 		 * A simple solution would be to box always and make a normal virtual call, but that would
 		 * be bad performance wise.
 		 */
-		if (mono_class_is_interface (cmethod->klass) && mono_class_is_ginst (cmethod->klass)) {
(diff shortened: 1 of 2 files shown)


lambdageek:
The testcase `bad-constrainedcall.il` has the same code as the c# example (and a couple of other cases), but it's written in IL.

lambdageek:

Issue #15731, opened 2019-07-18
[wasm][bindings] WebAssembly.Core.Array.Pop() throws error if the array contains no values.


kjpou1:
The Pop method throws an invalid cast exception if the WebAssembly.Core.Array contains no elements.

PR #15732, opened 2019-07-18
[wasm][bindings] WebAssembly.Core.Array.Pop should return null 

kjpou1:
if the WebAssembly.Core.Array is empty instead of throwing error.

- Add test to the BindingsTestSuite.

Fixes https://github.com/mono/mono/issues/15731


kjpou1:
@monojenkins build failed

PR #15743, opened 2019-07-18
[loader] Pass ALC to mono_domain_loaded_internal, and then to domain search hook

lambdageek:
Replace runtime callers of `mono_domain_loaded_full` by `mono_domain_loaded_internal` - pass a `MonoAssemblyLoadContext` to the callers.  Update the search hook to take an ALC argument (for now it's not used for anything, but it will be needed to check if an assembly is already loaded in the correct ALC).

Add a temporary `mono_domain_ambient_alc` function which just returns the default ALC, but is meant to mark a place where we should be getting an ALC from the callers.

### DIFF
--- a/mono/dis/main.c
+++ b/mono/dis/main.c
@@ -1942,7 +1942,7 @@ monodis_assembly_load_hook (MonoAssembly *assembly, gpointer user_data)
 }
 
 static MonoAssembly *
-monodis_assembly_search_hook (MonoAssemblyName *aname, gpointer user_data)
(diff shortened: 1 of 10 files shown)


vargaz:
It might be a good idea to centralize loading into the alc object, i.e.
instead of mono_image_load (mono_get_alc (domain)), you would do
mono_alc_image_load (mono_get_alc (domain)).

lambdageek:
@monojenkins build failed

lambdageek:
@vargaz yea, we'll try to move more of the loading code to the ALC.  I didn't have a good idea of how intrusive the ALCs were going to be.

PR #15744, opened 2019-07-18
[ios] switch to python offset tool

lewurm:
The C# version doesn't work on 64bit. It would require to update its dependencies, but Zoltan already added a Python version for WebAssembly which works fine.

Some notes:
* changed the semantics of `--xcode-path` in order to match the output of `xcode-select -p`
* switched to python3/pip3
* removed hack for arm64_32, it works properly now because the Python version uses stock `clang`
* removed some leftover usages of `XCODE32_DIR`.

Manually verified by comparing output with the C# version.

This will unblock the iOS team when building on Catalina/Xcode11, as they can't consume SDK archives there. See https://github.com/xamarin/xamarin-macios/pull/6603#issuecomment-512803082

### DIFF
--- a/sdks/builds/wasm.mk
+++ b/sdks/builds/wasm.mk
@@ -20,10 +20,7 @@ ZLIB_HEADERS = \
 	$(MONO_SUPPORT)/zlib.h  	\
 	$(MONO_SUPPORT)/zutil.h
 
-ifeq ($(UNAME),Darwin)
-# The c# offsets tool is 32 bit, and the 64 bit version doesn't work
 USE_OFFSETS_TOOL_PY = 1
-endif
 
 $(TOP)/sdks/builds/toolchains/emsdk:
 	git clone https://github.com/juj/emsdk.git $(EMSCRIPTEN_SDK_DIR)
@@ -154,7 +151,7 @@ endif
 #  $(6): offsets dumper abi
 define WasmCrossTemplate
 
-_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$$(EMSCRIPTEN_SDK_DIR)/upstream/emscripten"
+_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$$(EMSCRIPTEN_SDK_DIR)/upstream/emscripten" --libclang-path /home/lewurm/work/mono/sdks/builds/toolchains/emsdk/upstream/lib/
 
 _wasm-$(1)_CONFIGURE_FLAGS= \
 	--disable-boehm \
--- a/sdks/builds/wasm.mk
+++ b/sdks/builds/wasm.mk
@@ -151,7 +151,7 @@ endif
 #  $(6): offsets dumper abi
 define WasmCrossTemplate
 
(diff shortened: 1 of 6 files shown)


tools/offsets-tool-py/offsets-tool.py
```diff
@@ -183,7 +225,7 @@ def run_clang(self):
 							continue
 						if child.is_bitfield ():
 							continue
-						rtype.fields.append (FieldInfo (child.spelling, child.get_field_offsetof () / 8))
+						rtype.fields.append (FieldInfo (child.spelling, int (child.get_field_offsetof () / 8)))
```
lewurm: on Python3 `child.get_field_offsetof () / 8` yields a float, so you would have entries like this:
```c
DECL_OFFSET2(MonoClass,interface_offsets_count,100.0)
```

tools/offsets-tool-py/offsets-tool.py
```diff
@@ -50,18 +63,45 @@ def parse_args(self):
...
-		if not args.abi in self.supported_abis:
-			print ("ABI '" + args.abi + "' is not supported.", file=sys.stderr)
-			sys.exit (1)
-
 		self.sys_includes=[]
 		self.target = None
+		self.target_args = []
+
+		sysroot_ios = "/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
+		sysroot_watchos = "/Platforms/WatchOS.platform/Developer/SDKs/WatchOS.sdk"
```
akoeplinger: should we pass in the xcode + sysroot paths from https://github.com/mono/mono/blob/master/sdks/builds/ios.mk#L144-L147 instead of duplicating it here? e.g. if you want to build with a different xcode/sdk

tools/offsets-tool-py/offsets-tool.py
```diff
@@ -6,9 +6,23 @@
 import argparse
 import clang.cindex
 
+IOS_DEFINES = ["HOST_DARWIN", "TARGET_MACH", "MONO_CROSS_COMPILE", "USE_MONO_CTX", "_XOPEN_SOURCE"]
```
vargaz: Should probably be changed to /usr/bin/env python3.

tools/offsets-tool-py/offsets-tool.py
vargaz: Could use // for division.

sdks/builds/wasm.mk
```diff
@@ -154,7 +151,7 @@ endif
 #  $(6): offsets dumper abi
 define WasmCrossTemplate
 
-_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$$(EMSCRIPTEN_SDK_DIR)/upstream/emscripten"
+_wasm-$(1)_OFFSETS_DUMPER_ARGS=--emscripten-sdk="$$(EMSCRIPTEN_SDK_DIR)/upstream/emscripten" --libclang-path /home/lewurm/work/mono/sdks/builds/toolchains/emsdk/upstream/lib/
```
vargaz: This is not good.

lewurm:
Thanks Zoltan, updated the PR.

lewurm:
@monojenkins squash

lewurm:

lewurm:
@monojenkins backport 2019-06

PR #15745, opened 2019-07-18
[netcore] Fix CI failures

EgorBo:


EgorBo:
@monojenkins build failed
/azp run

PR #15748, opened 2019-07-18
[netcore] Implement Thread.CurrentOSThreadId

### LINKED ISSUE #15728: [netcore] Implement System.Threading.Thread.CurrentOSThreadId
Some platforms (Windows) return a 32-bit thread ID, while others return a 64-bit one, so be careful about that.

CoreCLR uses GetCurrentThreadId on Windows, and [this](https://github.com/dotnet/coreclr/blob/835836c9d34409af0f31529201dfd57cb2bd053c/src/pal/src/include/pal/thread.hpp#L811) for non-Windows. Notably, the distinction between `gettid` and `pthread_self` is unintuitive and it's probably best to just match CoreCLR's behavior.

Not sure how this works with wasm.

CoffeeFlux:
Fixes #15728

Not sure what to do for wasm and Haiku, so I've put a fairly boring attempt down for now (that's almost certainly wrong in Haiku's case). Most of the platform-specific behavior is exactly copied from CoreCLR.

Not sure if there are any tests I need to re-enable now that this is added, but I don't think so.

### DIFF
--- a/mono/metadata/icall-def-netcore.h
+++ b/mono/metadata/icall-def-netcore.h
@@ -501,6 +501,7 @@ HANDLES(SEMA_3, "ReleaseSemaphore_internal", ves_icall_System_Threading_Semaphor
 ICALL_TYPE(THREAD, "System.Threading.Thread", THREAD_1)
 HANDLES(THREAD_1, "ClrState", ves_icall_System_Threading_Thread_ClrState, void, 2, (MonoInternalThread, guint32))
 HANDLES(ITHREAD_2, "FreeInternal", ves_icall_System_Threading_InternalThread_Thread_free_internal, void, 1, (MonoInternalThread))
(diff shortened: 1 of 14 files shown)


lambdageek:
I think this needs a pretty long comment somewhere explaining that it's different from `mono_native_thread_id_get()` and that it's intended for different things.

CoffeeFlux:
Worth noting that I haven't tested this on anything but OSX, though I've made a good effort to ensure it'll work on all platforms. My main concern is other iDevices using Mach, as I'm not sure how the kernel differs on them and documentation seems to be sparse.

Also, the reason I'm using `syscall()` on Linux/Android is because most libc implementations (including Bionic) don't seem to include `gettid()`.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

NattyNarwhal:
Compiles fine on AIX.

PR #15749, opened 2019-07-18
 Special-case arrays of interfaces/generic parameters in IsAssignableFrom

### LINKED ISSUE #15647: Array of interface shouldn't be IsAssignableFrom an array of generic type variable

```csharp
using System;
using System.Reflection;

public interface I {
	void M1 ();
}

public class G<T> where T : I {
}

public class Genny {
	public static void Main ()
	{
		var t = typeof (G<>).GetTypeInfo().GenericTypeParameters [0];

		var ta = t.MakeArrayType ();

		var i = typeof (I).GetTypeInfo ();

		var ia = i.MakeArrayType ();

		Console.WriteLine (ta);

		Console.WriteLine (i.IsAssignableFrom (t));

		Console.WriteLine (ia.IsAssignableFrom (ta));
	}
}
```

With Mono the output is:
```
T[]
True
True
```

With CoreCLR and .NET Framework the output is:
```
T[]
True
False
```

---

Mono version:

```
Mono JIT compiler version 6.5.0.358 (master/6ae04f597d0 Thu Jun 27 11:50:29 EDT 2019)
	TLS:           
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(610)
	Suspend:       hybrid
```

CoffeeFlux:
Fixes #15647

I don't actually like this patch, as it's another special-case in an already patchwork function, but it fixes the immediate issue. At some point, mono_class_is_assignable_from_checked should probably just be refactored, but doing that without breaking anything is a rather daunting task.

### DIFF
--- a/mcs/class/corlib/Test/System/TypeTest.cs
+++ b/mcs/class/corlib/Test/System/TypeTest.cs
@@ -443,16 +443,16 @@ namespace MonoTests.System
 			Assert.IsFalse (typeof (Array).IsAssignableFrom (gparams[0]), "#45");
 			Assert.IsFalse (typeof (Delegate).IsAssignableFrom (gparams[0]), "#46");
 
-			// Arrays of generic parameters and arrays of interfaces, see #15749
+			// Arrays of generic parameters and arrays of interfaces, see https://github.com/mono/mono/pull/15749
 			var t = typeof (SampleGeneric<>).GetTypeInfo ().GenericTypeParameters [0];
 			var ta = t.MakeArrayType ();
 			var i = typeof (IFace1);
 			var ia = i.MakeArrayType ();
-			var t2 = typeof (SampleGenericConstrained<>).GetTypeInfo.GenericTypeParameters [0];
+			var t2 = typeof (SampleGenericConstrained<>).GetTypeInfo ().GenericTypeParameters [0];
 			var ta2 = t2.MakeArrayType ();
 			Assert.IsTrue (i.IsAssignableFrom (t), "#47");
 			Assert.IsFalse (ia.IsAssignableFrom (ta), "#48");
-			Assert.IsTrue (ia.IsAssignableFrom (t2a), "#49");
+			Assert.IsTrue (ia.IsAssignableFrom (ta2), "#49");
 		}
 
 		[Test]
--- a/mcs/class/corlib/Test/System/TypeTest.cs
+++ b/mcs/class/corlib/Test/System/TypeTest.cs
@@ -61,6 +61,14 @@ namespace MonoTests.System
 		C
 	};
 
+	class SampleGeneric<T> where T : IFace1
+	{
+	}
+
+	class SampleGenericConstrained<U> where U : class, IFace1
+	{
+	}
+
 	abstract class Base
 	{
 		public int level;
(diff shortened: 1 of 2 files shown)


mono/metadata/class.c
```diff
@@ -3826,6 +3826,20 @@ mono_class_is_assignable_from_checked (MonoClass *klass, MonoClass *oklass, gboo
...
 
+		/* 
+		 * a is b does not imply a[] is b[] in the case where b is an interface and
+		 * a is a generic parameter that implements that interface,.
+		 */
+
+		if (MONO_CLASS_IS_INTERFACE_INTERNAL (eclass)) {
+			MonoType *eoclass_byval_arg = m_class_get_byval_arg (eoclass);
+			if (mono_type_is_generic_argument (eoclass_byval_arg) && 
+				MONO_CLASS_IMPLEMENTS_INTERFACE (eoclass, m_class_get_interface_id (eclass))) {
```
lambdageek: why do we need this last check?  wouldn't we return false for all interfaces - whether or not the gparam implements it.

CoffeeFlux: Yeah, not sure what my thought process was yesterday, this is too specific. I probably wanted it to fall through for... some reason, but we can just fail it here and the result will be the same.

EgorBo:
I'd add an nunit/xunit test since CoreFX tests don't cover this case/

lambdageek:
Complications:

```csharp
>using System;

interface I {
}

class B {
	public static void Foo (I [] b) { }
}

class G<T> where T : I {
	public static void Foo (T[] b) {
		// B.Foo (b);
	}

}

class H<U> where U : class, I {
	public static void Foo (U[] b) {
		B.Foo (b);
	}
}

public class P {
	public static void Main ()
	{
		var t = typeof (G<>).GetGenericArguments () [0];
		var ts = t.MakeArrayType ();

		var eyes = typeof (I).MakeArrayType ();
		Console.WriteLine ("I[] assignable from T[] ? {0}", eyes.IsAssignableFrom (ts));

		
		var us = typeof (H<>).GetGenericArguments () [0].MakeArrayType ();

		Console.WriteLine ("I[] assignable from U[] ? {0}", eyes.IsAssignableFrom (us));
	}
}
```

`dotnet run` prints:

```
I[] assignable from T[] ? False
I[] assignable from T[] ? True
```

mono/metadata/class.c
```diff
@@ -3826,6 +3826,41 @@ mono_class_is_assignable_from_checked (MonoClass *klass, MonoClass *oklass, gboo
...
+		 * This should print:
+		 * I[] assignable from T[] ? False
+		 * I[] assignable from U[] ? True
+		 */
+
+		if (MONO_CLASS_IS_INTERFACE_INTERNAL (eclass)) {
+			MonoType *eoclass_byval_arg = m_class_get_byval_arg (eoclass);
+			MonoGenericParam *eoparam = eoclass_byval_arg->data.generic_param;
+			MonoGenericParamInfo *eoinfo = mono_generic_param_info (eoparam);
+			int eomask = eoinfo->flags & GENERIC_PARAMETER_ATTRIBUTE_SPECIAL_CONSTRAINTS_MASK;
```
lambdageek: You shouldn't look at `eoclass_byval_arg->data.generic_param` until you know that `mono_type_is_generic_argument (eoclass_byval_arg)` returns TRUE.  In particular accessing `eoinfo->flags` before we know that we're looking at a generic param could crash.

CoffeeFlux: Whoops. Also cleaned up some of the other logic there and commented the class constraint check, since that's definitely not obvious.

mcs/class/corlib/Test/System/TypeTest.cs
```diff
@@ -435,6 +443,16 @@ public void TestIsAssignableFrom ()
 			Assert.IsFalse (typeof (Array).IsAssignableFrom (gparams[0]), "#45");
 			Assert.IsFalse (typeof (Delegate).IsAssignableFrom (gparams[0]), "#46");
 
+			// Arrays of generic parameters and arrays of interfaces, see #15749
```
lambdageek: nit: we've had a bunch of bug trackers for Mono over the years.  If you get a chance to update the PR one more time, put in a full URL.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

Issue #15720, opened 2019-07-17
[wasm][bindings] Implement JavaScript managed object reference counting


kjpou1:
The current implementation of the bindings framework creates a handle to each managed JavaScript object using the `GCHandle.Alloc` method, which prevents the managed object from being collected.  This leads to the developers code being sprinkled with calling `Dispose` directly or the 'using' construct so that the JavaScript manage objects can be collected.

The bindings should be modified to implement reference counting to facilitate the correct object collection.

TheXenocide:
Is this for references to JS Objects to clean up their managed interop representation or to prevent managed objects exposed to JS from being GC'd? I was thinking that managed objects exposed to JS might be able to use JS WeakSet/Map and Finalizer functionality to track whether or not there's still a reference and then notify the WASM runtime that an object is no longer referenced by JS. Perhaps there's a way to leverage that native functionality the other way around too?

PR #15754, opened 2019-07-19
[wasm][bindings] Implement SafeHandle for framework objects

kjpou1:
Closed in favor of https://github.com/mono/mono/pull/17303

## Wrap the javascript handle in a SafeHandleZeroOrMinusOneIsInvalid implementation.

- AnyRef
   - Modify `AnyRef` to be abstract and deriving from `SafeHandleZeroOrMinusOneIsInvalid`
   - Add extra parameter for `ownsHandle`
   - The `GCHandle.Alloc` creates a normal allocation or a weak allocation depending on the ownsHandle value passed.
      - true - GCHandleType.Weak
      - false - GCHandleType.Normal

- JSObject
   - Implements `ReleaseHandle` that is overridden from `AnyRef`
   - This removes the `Dispose` and Finalizers definitions from the objects.  This will be taken care of by `AnyRef` which derives from `SafeHandleZeroOrMinusOneIsInvalid`
   -  Add extra parameter for `ownsHandle`

- Runtime
   - The `bound_objects` dictionary that tracks all the objects that have been bound from JavaScript objects to managed code is now a `WeakReference`.
   - Implement methods for obtaining a handle, adding a reference to a handle and releasing a handle.  These are called from the javascript `binding_support` lib.
   - Refactored the releasing a `JSObject.`
      * Involved multiple calls to the bindings and managed code.
      * Streamlined to only call the bindings code once to free handles.

## Add support for marshaling the `AnyRef` SafeHandle.

- When calling into the bindings from managed code the` JSObject` reference handle is incremented.
- After execution of the call the `JSObject` reference is decremented.
- Special processing for the `GlobalObject` where the object is set with `ownsHandle` = false so that it is not released.
- Release reference counts on javascript exceptions.

## Support freeing delegates marshaled to the bindings.

Every delegate that was passed to the bindings code was kept forever causing leaks.

- Delegates are now not kept alive when GC'd.
- Delegates using lambda expressions forced the developer to explicitly free the delegate.

   * If a local is captured (used) by a lambda it becomes heap memory as we translate them into fields on an object.
   * A delegate that used lambda expressions was never collected and never released the heap so it could be collected.
   * This was a cause of memory leaks forcing the developer to explicitly call `WebAssembly.Runtime.FreeObject (foreachAction);`

- If a delegate is collected and the javascript code calls that delegate function an exception is thrown instead of inexplicably crashing with null exception error.
   * throw new Error("The delegate target that is being invoked is no longer available.  Please check if it has been prematurely GC'd.");
   * Uncaught Error: The delegate target that is being invoked is no longer available.  Please check if it has been prematurely GC'd.

close https://github.com/mono/mono/issues/15720


kjpou1:

kjpou1:
Closed in favor of https://github.com/mono/mono/pull/17303 which is a cleaner version.

cc @lewing

kjpou1:
Closed in favor of https://github.com/mono/mono/pull/17303 which is a cleaner version.

Issue #15760, opened 2019-07-19
TypeLoadException thrown for recursive type that .NET allows


UnityAlex:
When running an executable that uses recursively defined types a TypeLoadException is thrown by mono where .NET does not.

## Steps to Reproduce

1. Compile and run the sample code. No additional arguments required.
2. Observe exception thrown. 

```csharp
using System;

namespace TestRecursiveType
{
    class Program
    {
        static void Main(string[] args)
        {
            SubClass subC = new SubClass();
            Console.WriteLine(subC.GetTest());
        }
    }

    public struct ValueTest<T>
    {
        public readonly T value;
    }

    public abstract class BaseClass<T> where T : BaseClass<T>
    {
        public ValueTest<T> valueTest = default(ValueTest<T>);
    }

    public class SubClass : BaseClass<SubClass>
    {
        private String test = "test";


        public string GetTest()
        {
            return test;
        }
    }
}
```

### Current Behavior

```>mono.exe Program.exe

Unhandled Exception:
System.TypeLoadException: Could not set up parent class, due to: Generic Type Definition failed to init, due to: Recursive type definition detected TestRecursiveType.BaseClass`1 assembly:C:\Users\alexthibodeau\source\repos\TestRecursiveType\TestRecursiveType\Program.exe type:BaseClass`1 member:(null) assembly:C:\Users\alexthibodeau\source\repos\TestRecursiveType\TestRecursiveType\Program.exe type:BaseClass`1 member:(null)
[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeLoadException: Could not set up parent class, due to: Generic Type Definition failed to init, due to: Recursive type definition detected TestRecursiveType.BaseClass`1 assembly:C:\Users\alexthibodeau\source\repos\TestRecursiveType\TestRecursiveType\Program.exe type:BaseClass`1 member:(null) assembly:C:\Users\alexthibodeau\source\repos\TestRecursiveType\TestRecursiveType\Program.exe type:BaseClass`1 member:(null)
```

### Expected Behavior

Expecting "test" to be written to console with no exceptions thrown.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[X ] Windows

**Version Used**:

```>mono.exe --version
Mono JIT compiler version 6.5.0 (Visual Studio built mono)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       normal
[...]

PR #15761, opened 2019-07-19
[metadata] Fields whose types are gparams with a reference type constraint aren't blittlable.

lambdageek:
Don't try to layout the field to find out if it's blittable.

Fixes certain recursive examples.

```csharp
using System;

namespace TestRecursiveType
{
    class Program
    {
        {
            SubClass subC = new SubClass();
            Console.WriteLine(subC.GetTest());
        }
    }

    public struct ValueTest<U>
    {
        // When U is instantiated with T, from BaseClass, we know it'll be a
	// reference field, so we know the instantiation ValueTest<T> won't
	// be blittable.
        public readonly U value;
    }

    public abstract class BaseClass<T> where T : BaseClass<T>
    {
        public ValueTest<T> valueTest = default(ValueTest<T>);
    }

    public class SubClass : BaseClass<SubClass>
    {
        private String test = "test";

        public string GetTest()
        {
            return test;
        }
    }
}
```

Fixes https://github.com/mono/mono/issues/15760

---

### TODO ###
- [x] Add a test case

### DIFF
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -3501,6 +3501,41 @@ type_has_references (MonoClass *klass, MonoType *ftype)
 	return FALSE;
 }
 
+static gboolean
+mono_class_is_gparam_with_reference_parent (MonoClass *klass)
+{
+	MonoType *type = m_class_get_byval_arg (klass);
+	g_assert (mono_type_is_generic_parameter (type));
+	MonoGenericParam *gparam = type->data.generic_param;
+	if ((mono_generic_param_info (gparam)->flags & GENERIC_PARAMETER_ATTRIBUTE_REFERENCE_TYPE_CONSTRAINT) != 0)
+		return TRUE;
+	if ((mono_generic_param_info (gparam)->flags & GENERIC_PARAMETER_ATTRIBUTE_VALUE_TYPE_CONSTRAINT) != 0)
+		return FALSE;
+
+	/* FIXME: is this reasonable? */
+	if (mono_generic_param_owner (gparam)->is_anonymous)
+		return FALSE;
+
+	/* We could have:  T : U,  U : Base.  So have to follow the constraints. */
+	MonoClass  *parent_class = mono_generic_param_get_base_type (klass);
+
+	g_assert (!MONO_CLASS_IS_INTERFACE_INTERNAL (parent_class));
+
+	/* Parent can only be: System.Object, System.ValueType or some specific base class.
+	 *
+	 * If the parent_class is ValueType, the valuetype constraint would be set, above, so
+	 * we wouldn't get here.
+	 *
+	 * If there was a reference constraint, the parent_class would be System.Object,
+	 * but we would have returned early above.
+	 *
+	 * So if we get here, there is either no base class constraint at all,
+	 * in which case parent_class would be set to System.Object, or there is none at all.
+	 */
+	return klass->parent != mono_defaults.object_class;
+
+}
+
 /*
  * mono_class_layout_fields:
  * @class: a class
@@ -3612,6 +3647,8 @@ mono_class_layout_fields (MonoClass *klass, int base_instance_size, int packing_
 			continue;
 		if (blittable) {
 			if (field->type->byref || MONO_TYPE_IS_REFERENCE (field->type)) {
+				blittable = FALSE;
+			} else if (mono_type_is_generic_parameter (field->type) && mono_class_is_gparam_with_reference_parent (mono_class_from_mono_type_internal (field->type))) {
 				blittable = FALSE;
 			} else {
 				MonoClass *field_class = mono_class_from_mono_type_internal (field->type);
--- a/mono/metadata/class-init.c
+++ b/mono/metadata/class-init.c
@@ -3501,8 +3501,22 @@ type_has_references (MonoClass *klass, MonoType *ftype)
 	return FALSE;
 }
 
+/**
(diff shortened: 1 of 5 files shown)


mono/metadata/class-init.c
```diff
@@ -3501,6 +3501,41 @@ type_has_references (MonoClass *klass, MonoType *ftype)
...
+	g_assert (mono_type_is_generic_parameter (type));
+	MonoGenericParam *gparam = type->data.generic_param;
+	if ((mono_generic_param_info (gparam)->flags & GENERIC_PARAMETER_ATTRIBUTE_REFERENCE_TYPE_CONSTRAINT) != 0)
+		return TRUE;
+	if ((mono_generic_param_info (gparam)->flags & GENERIC_PARAMETER_ATTRIBUTE_VALUE_TYPE_CONSTRAINT) != 0)
+		return FALSE;
+
+	/* FIXME: is this reasonable? */
+	if (mono_generic_param_owner (gparam)->is_anonymous)
+		return FALSE;
```
lambdageek: I had to exit early if the generic container is anonymous here in order for the AOT step not to crash - I guess we could be instantiating something with a gshared type.

I think I need to look at the gshared constraint to find out if it's a valuetype or a reference type.

filipnavara:
This sounds tricky. You can have reference generic parameter and never use it in the struct. Technically such structure could still be blittable. An example is in the CoreCLR issue.

lambdageek:
@filipnavara The failure is happening when we are doing `mono_class_setup_fields ("BaseClass<T>")` which needs to decide for each field whether it is blittable or not.  So therefore we are trying to decide if `ValueTest<T>` (that is: the `ValueTest<U>` field inside `BaseClass<T>`) is blittable or not.

So we instantiate `U` with `T`.  
Now to decide whether `VaueTest<T>` is blittable or not, we look at every field.  
So then we look at `T value`.  
To decide if T is blittable we first check if it's a reference type.

That check is currently inadequate for generic parameters - what the PR adds is the ability to see if theres a `T : class` constraint or a `T : C` constraint - where `C` is some class.  As soon as we know that `T`'s constraint will force it to be a reference type we can definitely say that `T` won't be blittable without having to initialize `C`, at all.

Previously, Mono would see that `T` is some kind of type for which it couldn't definitively decide that it's a reference type and it would call: `mono_class_setup_fields (field_class)` which would then try to setup the fields of the parent class `BaseClass<T>`.  And that would hit the recursion check.

filipnavara:
Thanks for explanation and for coming up with solution so quickly. Now it sounds reasonable to me.

I was worried that you actually tried to short-circuit the evaluation one step earlier - ie. implying that `ValueType<T>` is not blittable for reference `T` without looking at the actual fields (`T value`).

Such assumption would not be true for a struct like this one:
```csharp
struct A<T> where T : class
{
    byte a;
}
```

lambdageek:
@monojenkins build failed

lambdageek:
Various lanes failing with ``* Assertion at method-to-ir.c:12345, condition `var->opcode == OP_REGOFFSET' not met``

Not sure if that's related...

lambdageek:
Yea the AOT crashes are related to changing `mono_type_is_refrence` to return true for generic parameters with a `GENERIC_PARAMETER_ATTRIBUTE_REFERENCE_TYPE_CONSTRAINT` flag.  Going to revert that part.

mono/metadata/class-init.c
```diff
@@ -3501,6 +3501,41 @@ type_has_references (MonoClass *klass, MonoType *ftype)
...
+	 * we wouldn't get here.
+	 *
+	 * If there was a reference constraint, the parent_class would be System.Object,
+	 * but we would have returned early above.
+	 *
+	 * So if we get here, there is either no base class constraint at all,
+	 * in which case parent_class would be set to System.Object, or there is none at all.
+	 */
+	return klass->parent != mono_defaults.object_class;
+
```
vargaz: Extra whitespace.

mono/metadata/class-init.c
```diff
@@ -3613,6 +3648,8 @@ mono_class_layout_fields (MonoClass *klass, int base_instance_size, int packing_
 		if (blittable) {
 			if (field->type->byref || MONO_TYPE_IS_REFERENCE (field->type)) {
 				blittable = FALSE;
+			} else if (mono_type_is_generic_parameter (field->type) && mono_class_is_gparam_with_reference_parent (mono_class_from_mono_type_internal (field->type))) {
+				blittable = FALSE;
```
vargaz: Would it hurt things if we would set blittable=FALSE for all gparams ?

filipnavara: Yes (https://github.com/mono/mono/pull/15761#issuecomment-513379827)

lambdageek: We're not looking at whether a gparam like `T` is blittable here.  Rather, we're inside a GTD like `G<T>` that has some field `T field` and we're trying to decide whether that makes `G<>` blittable or not.

What I wonder is whether we actually care about GTDs like `G<>` or open instantiations like `G<List<T>>` being blittable, or if we only care about closed instantiations.

And I think we the answer is we do care, because of generic sharing.  Something like `G<List<T_REF>>` needs to have a valid `klass->blittable` value.

vargaz: Note that the gshared type parameters have this encoded in param->gshared_constraint.

lambdageek:
@monojenkins build failed

PR #15765, opened 2019-07-19
Miscellaneous loader-related fixes

CoffeeFlux:
Didn't want to mix these in with a larger PR.

The domain_assemblies appends were changed to prepends because doing so is significantly faster for large numbers of assemblies and will help with ordering consistency between the various places loaded assemblies are stored.

### DIFF
--- a/mono/metadata/metadata-internals.h
+++ b/mono/metadata/metadata-internals.h
@@ -1200,7 +1200,7 @@ static inline
 MonoAssemblyLoadContext *
 mono_assembly_get_alc (MonoAssembly *assm)
 {
-	return assm->image->alc;
+	return mono_image_get_alc (assm->image);
 }
 #endif
 
--- a/mono/metadata/metadata-internals.h
+++ b/mono/metadata/metadata-internals.h
@@ -1195,13 +1195,11 @@ mono_image_get_alc (MonoImage *image)
 #endif
 }
 
-#ifdef ENABLE_NETCORE
 static inline
 MonoAssemblyLoadContext *
 mono_assembly_get_alc (MonoAssembly *assm)
 {
 	return mono_image_get_alc (assm->image);
 }
-#endif
 
 #endif /* __MONO_METADATA_INTERNALS_H__ */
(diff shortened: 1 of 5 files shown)


vargaz:
This might cause weird issues because the assemblies are returned in a different order to clients, like in the debugger.

CoffeeFlux:
Hmm, is there any way I could easily test that? If there's no way to be confident I can remove the commit, but if we have code dependent on the order of that list I feel like that should be corrected if reasonably possible.

mono/metadata/metadata-internals.h
```diff
@@ -1200,7 +1200,7 @@ static inline
 MonoAssemblyLoadContext *
 mono_assembly_get_alc (MonoAssembly *assm)
 {
-	return assm->image->alc;
+	return mono_image_get_alc (assm->image);
```
lambdageek: Now that `mono_assembly_get_alc` has a reasonable definition for non-netcore builds (it'll just return NULL), you can get rid of the `#ifdef ENABLE_NETCORE` around it.  It'll help keep the overall number of ifdefs down, I think.

vargaz:
We can change the list to GList so the append performance is ok. If there are any issues, they will probably only show up in apps.

CoffeeFlux:
Isn't performance going to be the same with a doubly-linked list if we're only storing the pointer to the first element? I don't think reference GList is circular, and would assume ours is the same. I can just get rid of the commit.

My main motivation was that we're going to have three separate lists of loaded assemblies in the runtime, each with their own quirks, and I'm worried about subtle bugs resulting from that/any effort to simplify things.

lewing:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

lewing:
@monojenkins build Linux x64

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #12044, opened 2018-12-13
[Coop] Convert System.Diagnostics.{FileVersionInfo,Process}.

jaykrell:

PR #15775, opened 2019-07-22
[Coop] Convert System.Diagnostics.{FileVersionInfo,Process}.

jaykrell:
Continuation of https://github.com/mono/mono/pull/12044.


jaykrell:

jaykrell:
@monojenkins build failed

vargaz:
Looks ok to me. The _assign functions look a bit ugly, but without them, its hard to create objects in loops, either you allocate an unbounded number of handles, or have to use the non-handle version of the object creation functions.

jaykrell:
There are two other ways to handle the loops, but yes I like this way too.

1. You split the loop body into a function, you know.

2. I eventually tired of that -- more tedious, more churn, hard diffs to read --  `HANDLE_LOOP_PREPARE`, `SETUP_ICALL_FRAME`, `CLEAR_ICALL_FRAME`. See mono-perfcounters.c for example. 

I still like `_assign` but I have to be sure you know what options are here and elsewhere.
`_assign` is potentially much more efficient.
The ultimate efficiency is managed passes down handles for native to use as temporaries.

Not all of the code is perf-sensitive, but `_assign` is meant to make it easy to write faster code everywhere without worrying about the necessity.

mono/metadata/w32process-win32.c
```diff
@@ -363,37 +380,40 @@ ves_icall_System_Diagnostics_Process_CreateProcess_internal (MonoW32ProcessStart
...
+	gboolean success;
 
 	do {
 		pids = g_new0 (DWORD, count);
-		if (!mono_process_win_enum_processes (pids, count * sizeof (guint32), &needed)) {
-			// FIXME GetLastError?
-			mono_error_set_not_supported (error, "This system does not support EnumProcesses");
-			mono_error_set_pending_exception (error);
+
+        MONO_ENTER_GC_SAFE;
```
jaykrell: Editor put in spaces, need to fix.

mono/metadata/w32process.c
```diff
@@ -124,137 +124,97 @@ get_process_module_class (void)
...
 	g_assert (domain);
 
-	klass = mono_object_class (obj);
-	g_assert (klass);
-
-	field = mono_class_get_field_from_name_full (klass, fieldname, NULL);
-	g_assert (field);
-
-	string = mono_string_new_utf16_checked (domain, val, len, error);
+	mono_string_new_utf16_assign (str, domain, val, len, error);
```
lambdageek: `MONO_HANDLE_ASSIGN (str, mono_string_new_utf16_handle (domain, val, len, error))` and wrap the whole thing in `HANDLE_FUNCTION_ENTER`/`HANDLE_FUNCTION_RETURN`.  Same in several other places.

jaykrell: That allocates excess handles and continues to encourage doing so.
There should be primitives for easily being efficient instead of primitives only for inefficient.

jaykrell: That also takes away `return`, which we are used to, but is nice to avoid where possible.

jaykrell: It is disappointing that there is preference to do this less efficiently when it is easy to do better.
See https://github.com/mono/mono/pull/15789.

PR #15776, opened 2019-07-22
Consolidate UWP implementation with non-UWP implementation.

jaykrell:
Prelude to https://github.com/mono/mono/pull/15775.


filipnavara:
@monojenkins build failed

jaykrell:
@monojenkins build failed

vargaz:
@monojenkins build failed

lambdageek:
I'd like to hold off on this until @lateralusX gets a chance to take a look.

lateralusX:
This will break platforms that supply their own version of all these methods (that was the purpose of breaking them out like this). You can expect that their exist an alternative implementation of w32process-win32-uwp.c elsewhere, so we need a way to exclude these functions when building upstream mono sources and include alternative ones in closed repro builds. I think a better approach is to continue a long the lines we started with the API subset and see if that could be applied to these functions or leave as is and I will make a pass over all this later this fall (as part of a different initiative).

jaykrell:
Wasn't the factoring basically replacing the entire file, but one level below the icalls?
So you could still do that, but at the icall level instead of one small level below?
I'll look again..

jaykrell:
I think we could maybe have  trilevel enum for each icall.
1. classic
2. unsupported
3. supplied elsewhere

but maybe your point is that 2 and 3 are kinda the same, as in elsewhere can just provide the unsupported one.

But I still think my previous point -- that the very slight factoring below the icall level isn't too useful.
Perhaps that is a different cleanup though and already done, let me check..

jaykrell:
It was partly done. The UWP file had a mix of toplevel icalls, and slightly-below icalls. Maybe I clean that part up then, soon-ish..

BrzVlad:
Looks like we are not going with this ?

PR #15777, opened 2019-07-22
[wasm][tests] Add running the browser tests for threads.

kjpou1:
- Build and run the tests using threads


kjpou1:
@monojenkins build failed

sdks/wasm/Makefile
```diff
@@ -410,13 +416,17 @@ test-debugger: build-debugger-test-app build-dbg-testsuite
 run-browser-tests: $(BROWSER_TEST)/.stamp-browser-test-suite
 	(cd $(BROWSER_TEST) && npm test)
 
+run-browser-threads-tests: $(BROWSER_TEST_THREADS)/.stamp-browser-test-threads-suite
+	(cd $(BROWSER_TEST_THREADS) && npm test)
+
 clean: clean-sdk
```
lewing: Maybe run clean-browser-tests here as well and get rid of the duplication in the commands

sdks/wasm/Makefile
```diff
@@ -230,11 +231,16 @@ bin/test-suite/mono.js: packager.exe $(WASM_FRAMEWORK)/.stamp-framework binding_
 
 build-test-suite: bin/test-suite/mono.js
 
-$(BROWSER_TEST)/.stamp-browser-test-suite: packager.exe $(WASM_FRAMEWORK)/.stamp-framework $(BROWSER_TEST_SOURCES) build-sdk
+$(BROWSER_TEST)/.stamp-browser-test-suite: packager.exe $(WASM_FRAMEWORK)/.stamp-framework $(BROWSER_TEST_SOURCES) clean-browser-tests build-sdk
```
lewing: are we cleaning here because the builds of the different configs end up in the same place?

Issue #15008, opened 2019-06-12
[netcore] Make System.Net.Tests.HttpListenerResponseCookiesTests Pass


steveisok:
This is a whole suite of tests that fail in the same way.  The skip file includes the whole class and shows the following error:

```
System.Net.Tests.HttpListenerResponseCookiesTests.AppendCookie_ValidCookie_AddsCookieToCollection [FAIL]
      System.ArgumentException : Cannot bind to the target method because its signature is not compatible with that of the delegate type.
      Stack Trace:
           at System.Delegate.CreateDelegate(Type type, Object firstArgument, MethodInfo method, Boolean throwOnBindFailure, Boolean allowClosed)
           at System.Delegate.CreateDelegate(Type type, MethodInfo method, Boolean throwOnBindFailure)
           at System.Delegate.CreateDelegate(Type type, MethodInfo method)
           at System.Reflection.RuntimeMethodInfo.CreateDelegate(Type delegateType)
           at System.Net.CookieExtensions.IsRfc2965Variant(Cookie cookie)
           at System.Net.HttpListenerResponse.ComputeCookies()
           at System.Net.HttpListenerResponse.SendHeaders(Boolean closing, MemoryStream ms, Boolean isWebSocketHandshake)
           at System.Net.HttpResponseStream.GetHeaders(Boolean closing, Boolean isWebSocketHandshake)
           at System.Net.HttpResponseStream.DisposeCore()
           at System.Net.HttpResponseStream.Dispose(Boolean disposing)
           at System.IO.Stream.Close()
           at System.Net.HttpConnection.Close(Boolean force)
           at System.Net.HttpListener.Cleanup(Boolean close_existing)
           at System.Net.HttpListener.Close(Boolean force)
           at System.Net.HttpListener.Dispose()
           at System.Net.HttpListener.System.IDisposable.Dispose()
           at System.Net.HttpListener.Close()
           at System.Net.Tests.HttpListenerFactory.Dispose()
           at System.Net.Tests.HttpListenerResponseTestBase.Dispose()
           at ReflectionAbstractionExtensions.DisposeTestClass(ITest test, Object testClass, IMessageBus messageBus, ExecutionTimer timer, CancellationTokenSource cancellationTokenSource)
```

Determine why the test fails and make it pass.

PR #15787, opened 2019-07-22
[netcore] Allow different enums as return type in CreateDelegate

EgorBo:
Fixes https://github.com/mono/mono/issues/15008

Test case:
```csharp
using System;

public enum Enum1 : long
{
    A,
    B,
    C
}

public enum Enum2 : long
{
    A,
    B,
    C
}

public sealed class Program
{
    public Enum1 Test => Enum1.A;

    {
        var del1 = typeof(Program).GetProperty("Test").GetGetMethod().CreateDelegate(typeof(Func<Program, Enum1>));
        
        // the Test property is of Enum1 type but we can use Enum2 too
        var del2 = typeof(Program).GetProperty("Test").GetGetMethod().CreateDelegate(typeof(Func<Program, Enum2>));
    }
}
```

Also, ignore `System.Transactions.Tests` - extremely slow on our CI (slow hardware?) and even timeouts sometimes

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -226,21 +226,6 @@
 #-nomethod System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandler_ClientCertificates_Test.AutomaticOrManual_DoesntFailRegardlessOfWhetherClientCertsAreAvailable
 
 ####################################################################
-##  System.Net.HttpListener.Tests
-####################################################################
-
-# TODO:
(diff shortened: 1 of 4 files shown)


EgorBo:
ARM64 lane is broken on CI everywhere.

marek-safar:
@EgorBo this seems to break quite a few existing tests

EgorBo:
@marek-safar yeah, will push a fix.

EgorBo:
@marek-safar fixed.

PR #15788, opened 2019-07-22
[draft] Replace code manager on Win32 with executable heap.

jaykrell:


jaykrell:
@monojenkins build failed

akoeplinger:

PR #15790, opened 2019-07-22
[netcore] Enable tests on Windows CI

EgorBo:


EgorBo:
I don't know how to fix ARM64 - someone broke it.

PR #15792, opened 2019-07-22
Remove unneeded auto cast with codeman.

jaykrell:
The uses are all void* or casted.


vargaz:
@monojenkins build failed

PR #15804, opened 2019-07-23
[tests] Return non-zero exit code if from MERP crash test on failure

### LINKED ISSUE #15781: MERP Crash Test Can Never Fail
Take a look at the merp crash test line here:

https://github.com/mono/mono/blob/master/mono/tests/merp-crash-test.cs#L332-L333

I don't believe the test runner will ever know if the test passed or failed.  I'd like to have this fixed so we can better validate MERP on Catalina w/ a hardened runtime.

lambdageek:
If any of the stress tests fail, return a non-zero exit code from the
supervising process.

Fixes #15781

### DIFF
--- a/mono/tests/merp-crash-test.cs
+++ b/mono/tests/merp-crash-test.cs
@@ -292,7 +292,7 @@ class C
 		}
 	}
 
-	public static void Main (string [] args)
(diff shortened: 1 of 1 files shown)


akoeplinger:

PR #15808, opened 2019-07-23
dladdr shim for gmodule; try to enable crash reporter on AIX

NattyNarwhal:
Some thoughts:

* put it in eglib since that's where it seems to make sense. the AIX reimpl is big enough to justify its own file.

* nothing really used gmodule other than libtest, preferring to access libdl/win32 directly.

* API for g_module_address is my own invention; IIRC real glib doesn't have this. tries to make it more piecemeal.

* the dladdr implementation returns a non-const string for file name (very annoying, must special-case freeing) because it lies on an allocated buffer. theoretically, it exists in memory as part of the (AFAIK fixed-address, and only existent on 64-bit) user-space loader, but accessing its memory sounds highly risky, being undocumented and likely ABI unstable.

* the fact said implementation allocs memory might be scary. do most dladdr implements guarantee that it won't? (i know most don't guarantee re-entrant behaviour)

* also is simplistic because it only reads symbols from text, hoping they have a traceback. a proper implementation parsing the XCOFF image in memory would be more complex, though arguably worth it, since function descriptors are in the data section and mixups are unfortunately common.

* the crash reporter seems to not break anything and its symbols are present, and it defaults to on, but I don't notice anything different happening. anything need to be done here? I noticed we don't have backtrace/backtrace_symbols on AIX and that's required for native dumps; perhaps a reimpl/wrapper is needed for that too

* implementation for windows/wasm not present. theoretically possible, though! also not tested on not-AIX, but the codepaths that do compile for me should compile elsewhere


NattyNarwhal:
not sure how I would test the crash reporter or if it's working; would it make a report and send it?

I tried the `merp-crash-test` but it bombs out if not on macOS (managed part of MERP); and tangentially it did expose another flaw in my workaround logic for not being able to get the executable path.

CoffeeFlux:
@alexischr can probably comment on how to test the crash reporter - not sure myself.

mono/eglib/gmodule-win32.c
```diff
@@ -137,6 +137,16 @@ g_module_symbol (GModule *module, const gchar *symbol_name, gpointer *symbol)
 	}
 }
 
+gboolean
+g_module_address (void *addr, char **file_name, void **file_base, char **sym_name, void **sym_addr)
+{
+	/*
+	 * XXX: For an actual implementation, file_name can be filled in from
+	 * GetModuleFileName, at least?
+	 */
```
jaykrell: `GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS)`.

jaykrell: And then `GetModuleFileName` agreed.

NattyNarwhal: I don't have a Windows system anymore to test this on, so I would be flying blind. I did do some research into what a Windows implementation would likely use though, so I did leave the comment here.

NattyNarwhal:
Turns out you need `backtrace_symbols` for it to actually make a native trace (and maybe actually any dumping, need to try that). i offers a compat stub, AIX does not. I have uncommited changes to fixup linkage to that for i, and just brutishly hacked it onto an AIX build. Unfortunately, looks like the signal handler's backchain doesn't go to what it interrupted, which means frames basically look like they're only the signal handler, and on AIX, something's corrupting/mixing up the areas, so it gets lost in the weeds and segfaults in a segfault handler. Eep.

Current hell: https://gist.github.com/NattyNarwhal/a3083df8925af5c4b1f40d29dfc6ac02

NattyNarwhal:
I have a working-ish native backtrace after fixing up the IBM shim to support signals. AIX will likely need some EGlibification of it.

NattyNarwhal:
BTW, this patch is a little more interesting for BSD/Haiku users because it should detect backtrace on those platforms now.

Also, I noticed comments a bit inaccurate - Win32 would have the same behaviour as the AIX compat shim. I'll get to updating those.

NattyNarwhal:
bumping; last I checked this should be functional, but I'm curious about anything related to crash reporting. MERP tests seem to only work on macOS, and I suspect that the crash reporting may be so. however, this PR is still useful, particularly for BSD users.

mono/eglib/gmodule-win32.c
```diff
@@ -140,6 +140,45 @@ g_module_symbol (GModule *module, const gchar *symbol_name, gpointer *symbol)
...
 
+gboolean
+g_module_address (void *addr, char **file_name, void **file_base, char **sym_name, void **sym_addr)
+{
+	HMODULE module;
+	/*
+	 * We have to cast the address because usually this func works with strings,
+	 * this being an exception.
+	 */
+	BOOL ret = GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, (LPCWSTR)addr, &module);
```
kg: I believe GetModuleHandleEx increments the refcount of the module, so unless we then decrement the refcount at the end of this function we're leaking a ref. Not sure how much that matters in practice, but we should call FreeLibrary on the handle after we attempt to get the filename.

jaykrell: It does. I added that in XP.
But FreeLibrary is also over-rated. A lot of .dlls don't behave well when unloaded anyway.

NattyNarwhal:
Tests pass; only reservation I have is the API contract is per-platform because of crumbly details. I'm tempted to make it `strdup` unconditionally in APIs where it can return `const char*` and remove `ifdef` guards.

NattyNarwhal:
Changed API impl and consumers; I feel this should reduce errors in the future for consumers of the API; you shouldn't have to ifdef out a free.

mono/mini/mini-exceptions.c
```diff
@@ -1512,20 +1512,24 @@ mono_get_portable_ip (intptr_t in_ip, intptr_t *out_ip, gint32 *out_offset, cons
 	// Note: it's not safe for us to be interrupted while inside of dl_addr, because if we
 	// try to call dl_addr while interrupted while inside the lock, we will try to take a
 	// non-recursive lock twice on this thread, and will deadlock.
-	Dl_info info;
-	gboolean success = dladdr ((void*)in_ip, &info);
+	char *sname = NULL, *fname = NULL;
+	void *saddr = NULL, *fbase = NULL;
+	gboolean success = g_module_address ((void*)in_ip, &fname, &fbase, &sname, &saddr);
```
lambdageek: So the fact that `g_module_address` dynamically allocates makes it unsuitable to be used right here.  `mono_get_portable_ip` is called by the crash reporter which runs in a signal handler (and also might be called when the malloc heap is corrupted) so it can't allocate.

mono/mini/mini-posix.c
```diff
@@ -938,12 +936,16 @@ dump_native_stacktrace (const char *signal, MonoContext *mctx)
 
 	for (int i = 0; i < size; ++i) {
 		gpointer ip = array [i];
-		Dl_info info;
-		gboolean success = dladdr ((void*) ip, &info);
+		char *sname = NULL, *fname = NULL;
+		gboolean success = g_module_address ((void*)ip, &fname, NULL, &sname, NULL);
```
lambdageek: `dump_native_stacktrace` is also called from a signal handler.

NattyNarwhal:
gitter convo says stack allocations for both the callee and caller would make sense due to the limitations, and I agree. will change API impls, signature and callers for that.

NattyNarwhal:
Oh, I forgot; Win32 is de facto nopped because u16to8 mallocs.

PR #15816, opened 2019-07-24
[netcore] Update CoreFX.issues.rsp

filipnavara:
Remove exclusions fixed by #14967


EgorBo:
oh, it's draft because you are checking them on CI? 🙂

filipnavara:
> oh, it's draft because you are checking them on CI?

yep. I checked locally but I want to be double sure.

PR #15826, opened 2019-07-25
mono_thread_set_name_internal failure should not be fatal,

jaykrell:
except as mandated by public interface.

### DIFF
--- a/mono/metadata/gc.c
+++ b/mono/metadata/gc.c
@@ -943,13 +943,15 @@ mono_runtime_do_background_work (void)
 static gsize WINAPI
 finalizer_thread (gpointer unused)
 {
-	ERROR_DECL (error);
 	gboolean wait = TRUE;
 
-	MonoString *finalizer = mono_string_new_checked (mono_get_root_domain (), "Finalizer", error);
-	mono_error_assert_ok (error);
-	mono_thread_set_name_internal (mono_thread_internal_current (), finalizer, FALSE, FALSE, error);
-	mono_error_assert_ok (error);
+	{
+		ERROR_DECL (error);
+		MonoString *finalizer = mono_string_new_checked (mono_get_root_domain (), "Finalizer", error);
+		if (is_ok (error))
+			mono_thread_set_name_internal (mono_thread_internal_current (), finalizer, FALSE, FALSE, NULL);
+		mono_error_cleanup (error);
+	}
 
 	/* Register a hazard free queue pump callback */
 	mono_hazard_pointer_install_free_queue_size_callback (hazard_free_queue_is_too_big);
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -1886,18 +1886,23 @@ ves_icall_System_Threading_Thread_GetName_internal (MonoInternalThreadHandle thr
 void 
 mono_thread_set_name_internal (MonoInternalThread *this_obj, MonoString *name, gboolean permanent, gboolean reset, MonoError *error)
 {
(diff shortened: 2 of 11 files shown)


mono/metadata/gc.c
```diff
@@ -943,13 +943,15 @@ mono_runtime_do_background_work (void)
...
-	ERROR_DECL (error);
 	gboolean wait = TRUE;
 
-	MonoString *finalizer = mono_string_new_checked (mono_get_root_domain (), "Finalizer", error);
-	mono_error_assert_ok (error);
-	mono_thread_set_name_internal (mono_thread_internal_current (), finalizer, FALSE, FALSE, error);
-	mono_error_assert_ok (error);
+	{
+		ERROR_DECL (error);
+		MonoString *finalizer = mono_string_new_checked (mono_get_root_domain (), "Finalizer", error);
```
lambdageek: Can we extract this recurring pattern into a function?

mono/metadata/threads.c
```diff
@@ -1886,18 +1886,23 @@ ves_icall_System_Threading_Thread_GetName_internal (MonoInternalThreadHandle thr
 void 
 mono_thread_set_name_internal (MonoInternalThread *this_obj, MonoString *name, gboolean permanent, gboolean reset, MonoError *error)
 {
+	// Mostly setting thread name is best-effort, and failure does not matter,
+	// but it is also a public API with a documented failure mode.
+
 	MonoNativeThreadId tid = 0;
 
-	LOCK_THREAD (this_obj);
+	if (error)
```
lambdageek: I weakly-dislike this.  We don't use a NULL MonoError as a sentinel value anywhere else in the runtime. 

What's the savings; we avoid one call to `mono_error_set_*`?

jaykrell:
Optional parameters don’t bother me.
They are very common outside of mono and work fine for everyone.
The copy/paste bothered me too.
I’ll revisit.

I considered set_name_noncrit but didn’t love the name.

Could be that null works but w/o copy/paste.
Like have function provide it and recurse.

The assert_ok common in runtime bothers me, which is what this combats microscopically.

jaykrell:
The booleans can be enum flags.
The constant parameter pattern is factored in the coop conversion.
So maybe combine the PRs.
Which I don’t love either.
The name is set_constant, and *coincindentally* are the noncrit ones, maybe a useful coincidence also, but I don’t love that either. But maybe.

mono/metadata/threads.c
jaykrell: If the error is going to be ignored, it seems better to avoid preparing it, and possibly that failing.
But I guess it could go either way, in terms of, when debugging it, if you want to hit mono_error_prepare or not.
The point is to avoid failability and noise, not cycles per se.
Debuggability is subtle and rarely considered and I not necessarily good at it.

jaykrell:
Subsumed by https://github.com/mono/mono/pull/15859.
Which cleans up much more.

It still uses a null `MonoError*` which I think is ok. This pattern is widespread outside mono and nobody seems to mind. It is arguably cleaner than mono's `foo` and `foo_full` pattern but the difference is small. I don't like having more functions like that because it is more to debug and search through, but ultimately `foo` will call `bar`, there will always be more than one function to search/read, it just cuts it down slightly.

However #15859 does add a flags enum which you might want to put this in.
And it coincides with the flag `constant` I put there.
Because managed doesn't pass constants and native does. Mostly but not entirely coincidence.

jaykrell:
#15859 was too large for reviewer. Got split up across several PRs and weeks.
Now subsumed by https://github.com/mono/mono/pull/16272.

PR #15827, opened 2019-07-25
[Coop] Convert ves_icall_System_Threading_Thread_SetName and usually ignore errors.

jaykrell:
This is done with an eye toward setting the unicode Windows 10 thread name.
There are two sources of thread names:
 public API with unicode
 internal constants that are 7bit clean

Extra conversions are avoided, and any representation we have we keep.
In particular, we do not convert the managed unicode to UTF8 and throw the Unicode.

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3010,12 +3010,11 @@ unload_thread_main (void *arg)
 	int i;
 	gsize result = 1; // failure
 
-	internal = mono_thread_internal_current ();
-
-	MonoString *thread_name_str = mono_string_new_checked (mono_domain_get (), "Domain unloader", error);
-	if (is_ok (error))
-		mono_thread_set_name_internal (internal, thread_name_str, TRUE, FALSE, error);
-	if (!is_ok (error)) {
+	// Domain unloader
+	mono_thread_set_name_constant (mono_thread_internal_current (),
+		MonoSetThreadNameFlag_Permanent, error,
+		'D','o','m','a','i','n',' ','u','n','l','o','a','d','e','r');
+	if (!is_ok (error)) { // FIXME: Does failure here matter?
 		data->failure_reason = g_strdup (mono_error_get_message (error));
 		goto failure;
 	}
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3003,7 +3003,6 @@ deregister_reflection_info_roots (MonoDomain *domain)
 static gsize WINAPI
 unload_thread_main (void *arg)
 {
-	ERROR_DECL (error);
 	unload_data *data = (unload_data*)arg;
 	MonoDomain *domain = data->domain;
 	MonoInternalThread *internal;
@@ -3011,13 +3010,9 @@ unload_thread_main (void *arg)
 	gsize result = 1; // failure
 
 	// Domain unloader
-	mono_thread_set_name_constant (mono_thread_internal_current (),
-		MonoSetThreadNameFlag_Permanent, error,
+	mono_thread_set_name_constant_ignore_error (mono_thread_internal_current (),
+		MonoSetThreadNameFlag_Permanent,
 		'D','o','m','a','i','n',' ','u','n','l','o','a','d','e','r');
-	if (!is_ok (error)) { // FIXME: Does failure here matter?
-		data->failure_reason = g_strdup (mono_error_get_message (error));
-		goto failure;
-	}
 
 	/* 
 	 * FIXME: Abort our parent thread last, so we can return a failure 
@@ -3080,7 +3075,6 @@ unload_thread_main (void *arg)
 
 	result = 0; // success
 exit:
-	mono_error_cleanup (error);
 	mono_atomic_store_release (&data->done, TRUE);
 	unload_data_unref (data);
 	return result;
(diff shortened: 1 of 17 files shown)


jaykrell:

mono/metadata/appdomain.c
```diff
@@ -3010,12 +3010,11 @@ unload_thread_main (void *arg)
...
-	internal = mono_thread_internal_current ();
-
-	MonoString *thread_name_str = mono_string_new_checked (mono_domain_get (), "Domain unloader", error);
-	if (is_ok (error))
-		mono_thread_set_name_internal (internal, thread_name_str, TRUE, FALSE, error);
-	if (!is_ok (error)) {
+	// Domain unloader
+	mono_thread_set_name_constant (mono_thread_internal_current (),
+		MonoSetThreadNameFlag_Permanent, error,
+		'D','o','m','a','i','n',' ','u','n','l','o','a','d','e','r');
```
vargaz: Can we just pass a normal c string here ? What does this change achieve ?

jaykrell: It gets us utf8 and utf16 static const. No copy, no alloc, no conversion.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build deb with monolith

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build deb with monolith

jaykrell:
monojenkins build deb with monolith

jaykrell:
monojenkins build deb with monolith

jaykrell:

jaykrell:
I had `SetName` vs. `SetName_icall` throwing everything off.

mono/metadata/appdomain.c
vargaz: I don't think is really neccessary, it looks ugly, and it saves like 1 allocation during the whole lifetime of the program.

jaykrell: I changed every constant thread name.
For example, `worker_callback` sets the thread name in a loop.
I think that is kinda terrible but it was defended as correct a while ago.

I have a larger PR where we can pass around utf8 and/or utf16, keep whatever we are given, convert to what we aren't, which we kinda do anyway. And support Windows native thread names, which are always unicode. So I'd like to start with this one and build up to that.

I agree it is a little ugly. I'd like to save whatever allocate and copy is easy. This is easy. Albeit ugly.
In C++11 you can say u"foo" to get a ucs2 string on all systems. That might be nice to make this pretty again some day.
On Win32 you could always say L"foo" but unfortunately that is ucs4 on others.

And this is crashing now I have to figure out why.

jaykrell:

mono/metadata/appdomain.c
jaykrell: Ok I have another idea to solve a few matters here. Please wait a day.
The strings will look “normal”.
Windows under ifdef will also use L.
Managed will pass utf8 & utf16.
Only utf8 will be retained.
Still a flag to distinguish constants.
Still curious why this crash.

jaykrell: Managed still just pass utf16, but the conversion will be moved up to `ves_icall_System_Threading_Thread_SetName_icall` instead of `mono_thread_set_name_internal`. Subtle.
This way native owns the alloc and free.

jaykrell:
Oh and I finally noticed that `mono_thread_get_name` returns unichar2.
Which moots a lot of what I was thinking.

jaykrell:
And `ves_icall_System_Threading_Thread_GetName_internal`. But perhaps that is rare and just retaining utf8 ok. Otherwise I'd like to retain both forms..

mono/metadata/icall-def.h
```diff
@@ -1050,7 +1050,7 @@ HANDLES(THREAD_12, "JoinInternal", ves_icall_System_Threading_Thread_Join_intern
 NOHANDLES(ICALL(THREAD_13, "MemoryBarrier", ves_icall_System_Threading_Thread_MemoryBarrier))
 HANDLES(THREAD_14, "ResetAbortNative", ves_icall_System_Threading_Thread_ResetAbort, void, 1, (MonoThreadObject))
 HANDLES(THREAD_15, "ResumeInternal", ves_icall_System_Threading_Thread_Resume, void, 1, (MonoThreadObject))
-ICALL(THREAD_18, "SetName_internal(System.Threading.InternalThread,string)", ves_icall_System_Threading_Thread_SetName_internal)
+HANDLES(THREAD_18, "SetName_icall", ves_icall_System_Threading_Thread_SetName_icall, void, 3, (MonoThreadObject, const_gunichar2_ptr, gint32))
```
jaykrell: InternalThread vs. Thread mixup here. I finally got a working system mono to see it.

jaykrell:
Subsumed by https://github.com/mono/mono/pull/15859 which doesn't have this ugliness and fixes more.

Issue #10748, opened 2018-09-21
Mono.Posix: Syscall.chown has incorrect signature


Nicholi:
## Steps to Reproduce

The signatures for `Mono.Unix.Native.Syscall.chown` is

`public static extern int chown([MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof (FileNameMarshaler))] string path, uint owner, uint group);`

Specifying both UID and GID as unsigned integers. However the native documentation for chown remarks this about the owner and group fields (man 2 chown).

> If the owner or group is specified as -1, then that ID is not changed.

Likewise Mono.Posix's own documentation further mentions this.

>             One of the owner or group id's
>             may be left unchanged by specifying it as -1.

Fortunately this **only** effects the newer preferred chown call in `Mono.Unix.Native`. The deprecated call available in `Mono.Posix` still correctly has the fields listed as signed integers.

`public static extern int chown (string path, int owner, int group);`

### Expected Behavior

Either remove the documentation specifying that -1 can be specified (it can't if the variables are uint), or change the signature of the calls to correlate with the underlying system calls (change fields to int).

**Version Used**:
1.0.0

PR #15828, opened 2019-07-25
Parameters for chown should be signed integers not unsigned

Nicholi:
Fixes https://github.com/mono/mono/issues/10748

The calls for chown, lchown, fchown, and fchownat (from `Mono.Unix.Native.Syscall`) are all declared with unsigned integers for the `owner` and `group` parameters. This is incorrect.

Directly from the manpage
`man 2 chown`
<pre><code>
...
       int chown(const char *pathname, uid_t owner, gid_t group);
       int fchown(int fd, uid_t owner, gid_t group);
       int lchown(const char *pathname, uid_t owner, gid_t group);

...

       int fchownat(int dirfd, const char *pathname,
                    uid_t owner, gid_t group, int flags);

...
...
       <b>If the owner or group is specified as -1</b>, then that ID is not changed.
...
</pre></code>

Emphasis added on the documentation explaining how the parameters can take the value -1, clearly not an unsigned integer. The `uid_t` struct is not an unsigned integer. I am assuming this was just an oversite on whoever re-wrote `Mono.Unix.Native.Syscall`.

The original calls from `Mono.Posix.Syscall` (now marked as Obsolete) all correctly use signed integers.
```
                [DllImport ("libc", SetLastError=true)]
                public static extern int chown (string path, int owner, int group);
                [DllImport ("libc", SetLastError=true)]
                public static extern int lchown (string path, int owner, int group);
```

Even the documentation directly from `mcs/class/Mono.Posix/Documentation/en/Mono.Unix.Native/Syscall.xml`
```
          <para>
            One of the owner or group id's
            may be left unchanged by specifying it as -1.
          </para>
```

I have changed the signatures from uint to int, the relevant XML documentation, and also 3 usages of chown within the project itself (switch from Convert.ToUInt32 to Convert.ToInt32). Hopefully this PR won't sit here for years, but who knows.

### DIFF
--- a/mcs/class/Mono.Posix/Documentation/en/Mono.Unix.Native/Syscall.xml
+++ b/mcs/class/Mono.Posix/Documentation/en/Mono.Unix.Native/Syscall.xml
@@ -695,8 +695,8 @@
       </Docs>
     </Member>
     <Member MemberName="chown">
-      <MemberSignature Language="C#" Value="public static int chown (string path, uint owner, uint group);" />
-      <MemberSignature Language="ILAsm" Value=".method public static hidebysig pinvokeimpl (&quot;libc&quot; as &quot;chown&quot; winapi lasterr)int32 chown(string path, unsigned int32 owner, unsigned int32 group) cil managed" />
+      <MemberSignature Language="C#" Value="public static int chown (string path, int owner, int group);" />
+      <MemberSignature Language="ILAsm" Value=".method public static hidebysig pinvokeimpl (&quot;libc&quot; as &quot;chown&quot; winapi lasterr)int32 chown(string path, int32 owner, int32 group) cil managed" />
       <MemberType>Method</MemberType>
       <AssemblyInfo>
         <AssemblyVersion>1.0.5000.0</AssemblyVersion>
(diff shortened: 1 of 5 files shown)


akoeplinger:
@monojenkins build

Nicholi:
Can't believe I totally spaced on the return types there.

akoeplinger:
@Nicholi it does introduce an API breaking change though so I think we'll need to add an overload for the existing uint-taking public methods.

Nicholi:
@akoeplinger Semi-related, would it be posible to request a more recent build uploaded to NuGet?

https://www.nuget.org/packages/Mono.Posix.NETStandard

The last one was quite awhile ago. Though it doesn't look like too many serious changes have been committed since then. Other than this one possibly.

Nicholi:
> @Nicholi it does introduce an API breaking change though so I think we'll need to add an overload for the existing uint-taking public methods.

Yeah I wasn't too sure on that, since the uint versions were technically incorrect all along. Re-added them back in marked regions, and also with Obsolete attribute. That seems most appropriate correct?

akoeplinger:
> Re-added them back in marked regions, and also with Obsolete attribute. That seems most appropriate correct?

Yes that looks fine.

> would it be posible to request a more recent build uploaded to NuGet?

I'll ping the relevant people after this was merged.

akoeplinger:
@monojenkins approve

akoeplinger:
@monojenkins commit apidiff

akoeplinger:
@Nicholi 	thank you for the contribution!

PR #15832, opened 2019-07-25
[netcore] Implement AssemblyLoadContext.GetLoadedAssemblies()

### LINKED ISSUE #14855: [netcore]Make Microsoft.VisualBasic.ApplicationServices.Tests.AssemblyInfoTests.LoadedAssemblies pass
The current test fails with:

```
   Microsoft.VisualBasic.ApplicationServices.Tests.AssemblyInfoTests.LoadedAssemblies [FAIL]
      System.NotImplementedException : The method or operation is not implemented.
      Stack Trace:
           at Microsoft.VisualBasic.ApplicationServices.AssemblyInfo.get_LoadedAssemblies()
           at Microsoft.VisualBasic.ApplicationServices.Tests.AssemblyInfoTests.LoadedAssemblies()
  Finished:    Microsoft.VisualBasic.Core.Tests
```

Determine why this is failing and try to make it pass.

CoffeeFlux:
Fixes #14855

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -1282,35 +1282,13 @@ leave:
 	HANDLE_FUNCTION_RETURN_VAL (is_ok (error));
 }
 
-MonoArrayHandle
-ves_icall_System_AppDomain_GetAssemblies (MonoAppDomainHandle ad, MonoBoolean refonly, MonoError *error)
+static MonoArrayHandle
+get_assembly_array_from_domain (MonoDomain *domain, MonoBoolean refonly, MonoError *error)
 {
-	error_init (error);
-	MonoDomain *domain = MONO_HANDLE_GETVAL (ad, data);
-	MonoAssembly* ass;
-	GSList *tmp;
 	int i;
 	GPtrArray *assemblies;
 
-	/* 
-	 * Make a copy of the list of assemblies because we can't hold the assemblies
-	 * lock while creating objects etc.
-	 */
-	assemblies = g_ptr_array_new ();
-	/* Need to skip internal assembly builders created by remoting */
-	mono_domain_assemblies_lock (domain);
-	for (tmp = domain->domain_assemblies; tmp; tmp = tmp->next) {
-		ass = (MonoAssembly *)tmp->data;
-		MonoBoolean ass_refonly = mono_asmctx_get_kind (&ass->context) == MONO_ASMCTX_REFONLY;
-		/* .NET Framework GetAssemblies() includes LoadFrom and Load(byte[]) assemblies, too */
-		/* For netcore, this should return all the assemblies in the domain - from every ALC */
-		if (refonly != ass_refonly)
-			continue;
-		if (ass->corlib_internal)
-			continue;
-		g_ptr_array_add (assemblies, ass);
-	}
-	mono_domain_assemblies_unlock (domain);
+	assemblies = mono_domain_get_assemblies (domain, refonly);
 
 	MonoArrayHandle res = mono_array_new_handle (domain, mono_class_get_assembly_class (), assemblies->len, error);
 	goto_if_nok (error, leave);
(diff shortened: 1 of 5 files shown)


mono/metadata/appdomain.c
```diff
@@ -1324,6 +1302,24 @@ ves_icall_System_AppDomain_GetAssemblies (MonoAppDomainHandle ad, MonoBoolean re
 	return res;
 }
 
+MonoArrayHandle
+ves_icall_System_AppDomain_GetAssemblies (MonoAppDomainHandle ad, MonoBoolean refonly, MonoError *error)
```
lambdageek: Do we need this one in netcore? can we ifdef it out?

mono/metadata/appdomain.c
```diff
@@ -1324,6 +1302,24 @@ ves_icall_System_AppDomain_GetAssemblies (MonoAppDomainHandle ad, MonoBoolean re
...
+	MonoDomain *domain = MONO_HANDLE_GETVAL (ad, data);
+	return get_assembly_array_from_domain (domain, refonly, error);
+}
+
+#ifdef ENABLE_NETCORE
+MonoArrayHandle
+ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalGetLoadedAssemblies (MonoBoolean refonly, MonoError *error)
+{
+	error_init (error);
+	MonoDomain *domain = mono_domain_get ();
```
vargaz: No need to init the error.

vargaz: I think netcore has no refonly assemblies.

mono/metadata/appdomain.c
```diff
@@ -1324,6 +1302,24 @@ ves_icall_System_AppDomain_GetAssemblies (MonoAppDomainHandle ad, MonoBoolean re
...
+	return get_assembly_array_from_domain (domain, refonly, error);
```
lambdageek: I think we can specialize to `refonly == FALSE` for netcore. looks like `ReflectionOnlyGetAssemblies` just always returns an empty array.
https://github.com/mono/mono/blob/090dffb9a09f26dc250d8d09a46819d192290b0b/netcore/System.Private.CoreLib/shared/System/AppDomain.cs#L183

CoffeeFlux: I left the parameter and just pass false, but I can remove it entirely from the icall.

mono/metadata/appdomain.c
CoffeeFlux: Do I not need to init the error for both this and the AppDomain call, or just this one? I'm not sure I understand why it's not needed here.

mono/metadata/appdomain.c
lambdageek: I'd just get rid of it entirely.

mono/metadata/appdomain.c
lambdageek: You shouldn't need it in both.

We changed how MonoError works at one point but didn't finish updating the code.  used to be that you had to call `error_init (error)` on entry to any function with a `MonoError *error` argument.

These days you no longer need to do that.  Instead the initialization is done once when the error is declared using `ERROR_DECL (error)`. In icalls it's done with macros when we make the handles wrapper https://github.com/mono/mono/blob/090dffb9a09f26dc250d8d09a46819d192290b0b/mono/metadata/icall-table.h#L482

mono/metadata/icall-def.h
```diff
@@ -177,7 +177,9 @@ HANDLES(KPAIR_5, "_ProtectUser", ves_icall_Mono_Security_Cryptography_KeyPairPer
 ICALL_TYPE(APPDOM, "System.AppDomain", APPDOM_23)
 HANDLES(APPDOM_23, "DoUnhandledException", ves_icall_System_AppDomain_DoUnhandledException, void, 2, (MonoAppDomain, MonoException))
 HANDLES(APPDOM_1, "ExecuteAssembly", ves_icall_System_AppDomain_ExecuteAssembly, gint32, 3, (MonoAppDomain, MonoReflectionAssembly, MonoArray))
+#ifndef ENABLE_NETCORE
```
lambdageek: Not needed.  This whole file is inside an ifdef.

CoffeeFlux: Oh right. Oops.

CoffeeFlux:

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #15837, opened 2019-07-26
Add more complete *freebsd* CI support

rootwyrm:
This adds more comprehensive functionality to support `*freebsd*` targets on Jenkins. Specific additions include checking pkgng function with pkg.FreeBSD.org, installing build and run dependencies, loading kernel modules, and mounting supporting filesystems. This should ensure successful builds, full coverage, and a reduced false-positive error rate. Note that this is based on the assumption of a 'clean' Azure FreeBSD 12.x or later VM (any architecture) which has already bootstrapped pkgng and uses pkg.FreeBSD.org. It should not be used in poudriere environments.

Pinging @akoeplinger for review


rootwyrm:
Whoops, I seem to have somehow managed to collide with someone else's work. Closing this PR and will re-submit without the XCODE collision.

PR #15839, opened 2019-07-26
Add more complete *freebsd* CI support

rootwyrm:
This adds more comprehensive functionality to support *freebsd* targets on Jenkins. Specific additions include checking pkgng function with pkg.FreeBSD.org, installing build and run dependencies, loading kernel modules, and mounting supporting filesystems. This should ensure successful builds, full coverage, and a reduced false-positive error rate. Note that this is based on the assumption of a 'clean' Azure FreeBSD 12.x or later VM (any architecture) which has already bootstrapped pkgng and uses pkg.FreeBSD.org. It should not be used in poudriere environments.

Pinging @akoeplinger for review, this time without accidentally stepping on his toes.

### DIFF
--- a/scripts/ci/run-jenkins.sh
+++ b/scripts/ci/run-jenkins.sh
@@ -354,6 +354,125 @@ if [[ ${CI_TAGS} == *'win-amd64'* ]];
     export MONO_EXECUTABLE_WRAPPER="${MONO_REPO_ROOT}/msvc/build/sgen/x64/bin/Release/mono-sgen-msvc.sh"
 fi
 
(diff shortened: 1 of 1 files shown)


steveisok:
@akoeplinger Please review

akoeplinger:
@rootwyrm I moved the CI provisioning scripts into the repo with #16326, can you please update that file instead? We probably don't need as much error handling there since it's fine to just fail if some error occurs.

rootwyrm:
Closing as this was addressed more appropriately by #16326

Issue #15824, opened 2019-07-25
[wasm][sdk] Error referencing System.CodeDom


kjpou1:
Hey there,

We're facing a similar issue:

![imagen](https://user-images.githubusercontent.com/1142782/61291529-05453000-a7cf-11e9-95a9-457a35c23d46.png)

Here you can repro it —just build and run, you can serve the web with `xsp4`, for instance:

https://github.com/MarcosCobena/WasmApp1/tree/bug-missing-system-codedom-dll

Our guess comes from System.CodeDom having a small piece bundled with mscorlib.dll (or System.dll), thus doesn't like the incoming System.CodeDom.dll NuGet. Does it make sense? In such case, would there be any workaround or similar to get it working? We're not so much interested in compiling code actually, but it's somewhat used for generating C# on the fly.

Thanks in advance!

—@sescalada & @MarcosCobena

_Originally posted by @MarcosCobena in https://github.com/mono/mono/issues/15404#issuecomment-511781275_

kjpou1:
See issue: https://github.com/mono/mono/issues/15823

MarcosCobena:
Thanks guys! :-)

Issue #15823, opened 2019-07-25
[wasm][sdk] WasmLinkAssemblies task wrongly uses reference assemblies


kjpou1:
Referencing `System.CodeDom` package from the SDK produces the following error.

```
Uncaught (in promise) Error: System.IO.FileNotFoundException: Could not load file or assembly 'System.CodeDom, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies.
File name: 'System.CodeDom, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
    at Object.call_method (mono.js:8687)
    at Object.call_static_method (mono.js:8752)
    at Object.init ((index):105)
    at runtime.js:9
    at mono.js:7981
```

---

The cause of this is that during the `WasmLinkAssemblies` the  resolved references from the msbuild task `ResolveAssemblies` references the `/ref/netstandard2.0` library instead of the actual `/lib/netstandard2.0` library path.

---

Example:

`       _MonoWasmLink:
         Running monolinker from /Users/Jimmy/websharp/projects/mono/sdks/out/wasm-bcl/wasm_tools/monolinker.exe.
         InvokeLinkerUsingMono True
[...]

PR #15841, opened 2019-07-26
[wasm][sdk] Fix issue with WasmLinkAssemblies task wrongly using refe…

kjpou1:
…rence assemblies

The [issue](https://github.com/mono/mono/issues/15823) is that when executing the `WasmLinkAssemblies` task the reference assemblies are currently used as input to the linker step and output in the published application instead of the actual framework library file.

My findings so far on this is that the package references are resolved in the MSBuild [ResolvePackageAssets](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs#L205.)

Those are then remapped from [here](https://github.com/dotnet/sdk/blob/ec9f5fc673f798dd6a89c7d11a4e62727837757c/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets#L270) to the `RuntimeCopyLocalItems`.

What I did was pass the `RuntimeCopyLocalItems` to the `WasmLinkAssemblies` task and do a remap to the runtime versions.  This works against everything I have thrown at it so far.

Of course this could fail if the nuget has somehow set `copy to local` false maybe but I do not see anything in the code referenced above checking for that.

Fixes https://github.com/mono/mono/issues/15824
Fixes https://github.com/mono/mono/issues/15823


kjpou1:
@monojenkins build failed

kjpou1:
@monojenkins build failed

kjpou1:
@radical does the above seem reasonable?

mhutch:
This seems reasonable to me, but runtime asset copying is an area of MSBuild I'm not super familiar with

PR #15842, opened 2019-07-26
[netcore] Add a missing make dependency.

vargaz:


vargaz:
@monojenkins build failed

PR #15844, opened 2019-07-26
[wasm][cli] Fix http framework module that errors using SDK 2.1.801

kjpou1:
- 2.1.801 seems to have a problem with:

```
   using (var uint8Buffer = Uint8Array.From(await request.Content.ReadAsByteArrayAsync ()))
```


lewing:
@monojenkins backport 2019-06

PR #15850, opened 2019-07-26
[netcore] Enable ALC loaded_assemblies

CoffeeFlux:
Notably, this PR does not touch the debugger or some of the other places that currently read domain_assemblies. Those will have to be made ALC-aware at a later date.

I'm not entirely convinced the way assemblies are freed in ALCs is correct, both due to locking and interaction with the GC, but this is an initial attempt at it. It shouldn't matter for now, but that will become relevant with collectability.

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -128,6 +128,9 @@ mono_domain_asmctx_from_path (const char *fname, MonoAssembly *requesting_assemb
 static void
 add_assemblies_to_domain (MonoDomain *domain, MonoAssembly *ass, GHashTable *hash);
 
+static void
+add_assembly_to_alc (MonoAssemblyLoadContext *alc, MonoAssembly *ass);
+
 static MonoAppDomainHandle
 mono_domain_create_appdomain_internal (char *friendly_name, MonoAppDomainSetupHandle setup, MonoError *error);
 
@@ -1449,6 +1452,43 @@ add_assemblies_to_domain (MonoDomain *domain, MonoAssembly *ass, GHashTable *ht)
 }
 
 static void
+add_assembly_to_alc (MonoAssemblyLoadContext *alc, MonoAssembly *ass)
+{
+	gint i;
+	GSList *tmp;
+
+	g_assert (ass != NULL);
+
+	if (!ass->aname.name)
+		return;
+
+	mono_alc_assemblies_lock (alc);
+
+	for (tmp = alc->loaded_assemblies; tmp; tmp = tmp->next) {
+		if (tmp->data == ass) {
+			mono_alc_assemblies_unlock (alc);
+			return;
+		}
+	}
+
+	mono_assembly_addref (ass);
+	// Prepending here will break the test suite with frequent InvalidCastExceptions, so we have to append
+	alc->loaded_assemblies = g_slist_append (alc->loaded_assemblies, ass);
+	mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_ASSEMBLY, "Assembly %s[%p] added to ALC (%p), ref_count=%d", ass->aname.name, ass, (gpointer)alc, ass->ref_count);
+
+	if (ass->image->references) {
+		for (i = 0; i < ass->image->nreferences; i++) {
+			if (ass->image->references[i] && ass->image->references [i] != REFERENCE_MISSING) {
+				// TODO: remove all this after we're confident this assert isn't hit
+				g_assert_not_reached ();
+			}
+		}
+	}
+
+	mono_alc_assemblies_unlock (alc);
+}
+
+static void
 mono_domain_fire_assembly_load (MonoAssembly *assembly, gpointer user_data)
 {
 	HANDLE_FUNCTION_ENTER ();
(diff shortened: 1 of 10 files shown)


mono/metadata/assembly-load-context.c
```diff
@@ -15,14 +16,79 @@ mono_alc_init (MonoAssemblyLoadContext *alc, MonoDomain *domain)
...
 void
 mono_alc_cleanup (MonoAssemblyLoadContext *alc)
 {
+	// TODO: get someone familiar with the GC to look at this
+	GSList *tmp;
+	MonoDomain *domain = alc->domain;
+
 	mono_loaded_images_free (alc->loaded_images);
+
+	for (tmp = alc->loaded_assemblies; tmp; tmp = tmp->next) {
```
lambdageek: @CoffeeFlux and I talked offline: this code in `mono_alc_cleanup` isn't ready yet.  We're going to document our assumptions about the refcount on each assembly in the ALC and add a `g_assert_not_reached` for now and revisit once we get started on collectible ALC support after we grok the CoreCLR collectible assemblies design.

mono/metadata/appdomain.c
```diff
@@ -2348,6 +2393,20 @@ mono_domain_assembly_search (MonoAssemblyLoadContext *alc, MonoAssembly *request
 	const MonoAssemblyNameEqFlags eq_flags = (MonoAssemblyNameEqFlags)(strong_name ? MONO_ANAME_EQ_IGNORE_CASE :
 		(MONO_ANAME_EQ_IGNORE_PUBKEY | MONO_ANAME_EQ_IGNORE_VERSION | MONO_ANAME_EQ_IGNORE_CASE));
 
+#ifdef ENABLE_NETCORE
+	mono_alc_assemblies_lock (alc);
+	for (tmp = alc->loaded_assemblies; tmp; tmp = tmp->next) {
+		ass = (MonoAssembly *)tmp->data;
+		g_assert (ass != NULL);
+		// TODO: can dynamic assemblies match here for netcore?
+		if (assembly_is_dynamic (ass) || !mono_assembly_names_equal_flags (aname, &ass->aname, eq_flags))
```
filipnavara: I realize you are not changing anything here compared to the previous status quo but it made me wonder anyway. Should not this mimic the logic of `exact_sn_match` more closely? What are the rules for the name matching here in .NET Core?

mono/metadata/appdomain.c
```diff
@@ -1462,6 +1503,7 @@ mono_domain_fire_assembly_load (MonoAssembly *assembly, gpointer user_data)
 	static MonoMethod *assembly_load_method;
 	ERROR_DECL (error);
 	MonoDomain *domain = mono_domain_get ();
+	MonoAssemblyLoadContext *alc = mono_domain_default_alc (domain);
```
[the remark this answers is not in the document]
CoffeeFlux: Fixed the function, though wiring up the invoke hook will be done in a later PR (even if this means assemblies will be put in the wrong place for now).

CoffeeFlux:
Should actually run the tests on netcore now...

mono/metadata/appdomain.c
CoffeeFlux: That's a fair point, and I'm not sure on the rules in Core myself, but I'm not inclined to hold this PR up for that since I have further ALC work to do that requires this PR as a base. I think for this PR, the TODO should be fine, and then I'll perhaps revisit.

CoffeeFlux:
@monojenkins build failed

mono/metadata/assembly-load-context.c
```diff
@@ -15,14 +16,56 @@ mono_alc_init (MonoAssemblyLoadContext *alc, MonoDomain *domain)
...
+	 * + Release GC roots for all assemblies in the ALC
+	 * + Iterate over the domain_assemblies and remove ones that belong to the ALC, which will probably require
+	 *     converting domain_assemblies to a doubly-linked list, ideally GQueue
+	 * + Close dynamic and then remaining assemblies, potentially nulling the data field depending on refcount
+	 * + Second pass to call mono_assembly_close_finish on remaining assemblies
+	 */
+
 	mono_loaded_images_free (alc->loaded_images);
+
+	g_slist_free (alc->loaded_assemblies);
```
lambdageek: Not sure if it's worth keeping this line if it's wrong anyway (need to iterate over the loaded_assemblies and decrement the refcounts before deleting the list).

mono/metadata/assembly-load-context.c
```diff
@@ -15,14 +16,56 @@ mono_alc_init (MonoAssemblyLoadContext *alc, MonoDomain *domain)
...
+	 * Missing steps:
+	 * 
+	 * + Release GC roots for all assemblies in the ALC
+	 * + Iterate over the domain_assemblies and remove ones that belong to the ALC, which will probably require
+	 *     converting domain_assemblies to a doubly-linked list, ideally GQueue
+	 * + Close dynamic and then remaining assemblies, potentially nulling the data field depending on refcount
+	 * + Second pass to call mono_assembly_close_finish on remaining assemblies
+	 */
+
 	mono_loaded_images_free (alc->loaded_images);
```
lambdageek: This should be one of the last things called in this function, not the first.

CoffeeFlux:
Rebased to deal with the conflict. Now to just wait on CI.

CoffeeFlux:
@monojenkins build failed

PR #15851, opened 2019-07-26
Update nuget.exe to 5.2.0, matching msbuild

radical:
This is a bit delayed as the 5.2.0 `nuget.exe` wasn't available earlier.


akoeplinger:
please make sure the msbuild+nuget updates are forward ported to master

PR #15857, opened 2019-07-27
Fix off by one bug in mono_thread_set_name_internal.

jaykrell:
In C++ you'd just write string::operator= once and be done.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -1905,8 +1905,9 @@ mono_thread_set_name_internal (MonoInternalThread *this_obj, MonoString *name, g
 		this_obj->name_len = 0;
 	}
 	if (name) {
-		this_obj->name = g_memdup (mono_string_chars_internal (name), mono_string_length_internal (name) * sizeof (gunichar2));
-		this_obj->name_len = mono_string_length_internal (name);
+		const size_t length = mono_string_length_internal (name);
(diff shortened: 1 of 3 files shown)


mono/metadata/threads.c
```diff
@@ -1905,8 +1905,9 @@ mono_thread_set_name_internal (MonoInternalThread *this_obj, MonoString *name, g
 		this_obj->name_len = 0;
 	}
 	if (name) {
-		this_obj->name = g_memdup (mono_string_chars_internal (name), mono_string_length_internal (name) * sizeof (gunichar2));
-		this_obj->name_len = mono_string_length_internal (name);
+		const size_t length = mono_string_length_internal (name);
+		this_obj->name = (gunichar2*)g_memdup (mono_string_chars_internal (name), (length + 1) * sizeof (gunichar2));
+		this_obj->name_len = length;
```
vargaz: Why is this needed ? c# strings are not null terminated. Do we want 'name' to be null terminated ?

jaykrell: 1. They are null terminated. So this is safe.
2. Oh, but I think you're right, we don't need it currently, the length suffices, but we probably will "need" it shortly (midly, could do without). This can wait and be part of needing it. I was trying to split up a larger PR.

jaykrell:
Isn't motivated on its own and will conflict with https://github.com/mono/mono/pull/15859.

PR #15867, opened 2019-07-28
ArgumentNullException when pointer is null for Marshal.Release

### LINKED ISSUE #15853: `Marshal.Release(IntPtr.Zero)` throws `ArgumentException` on Mono, `ArgumentNullException` on CLR
I was once told on this repo that any functional difference between Mono and the CLR is considered a bug in Mono... so I'd like to report one:

## Steps to reproduce

1. Create and run the following console program:

   ```csharp
   using System;

   class Program
   {
       static void Main()
       {
           Marshal.Release(IntPtr.Zero);
       }
   }
   ```

   <sub>**Background:** In case you're wondering why anyone would even do that, see this real-world example: https://github.com/castleproject/Core/commit/a0bdbb3d461e2bf5f25c27bbffa67da0620511be, where I codified an assumption about the above call in the form of a unit test; I noticed that test fails when accidentally executing it on Mono. I'll admit that I could simply rewrite that test&mdash;the important thing is that the call throws, not what precise exception type it throws.</sub>

2. Observe the exception being thrown.

### Current Behavior

Mono throws the following `ArgumentException`:

> ```
> System.ArgumentException: Value cannot be null.
> Parameter name: pUnk
>   at System.Runtime.InteropServices.Marshal.Release (System.IntPtr pUnk) [0x0000d] in <81342d83acda4c5590ec19c7afdf26b9>:0
>   at ConsoleApp3.Program.Main
> ```
> 

From here: https://github.com/mono/mono/blob/216fad51777dc5eb4f62ab43d8109629d4c13157/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs#L1114-L1115

### Expected Behavior

Mono should behave like the CLR, which throws an `ArgumentNullException` for the same program (I used .NET 4.7.2):

> ```
> System.ArgumentNullException: Value cannot be null.
> Parameter name: pUnk
>    at System.Runtime.InteropServices.Marshal.Release(IntPtr pUnk)
>    at ConsoleApp3.Program.Main
> ```

[ ] macOS
[ ] Linux 
[X] Windows 10
[X] Docker CE on Windows 10, using Docker image `mono:6`

**Version Used**:

```
Mono JIT compiler version 6.0.0.313 (tarball Sun Jul 14 09:55:10 UTC 2019)
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug
        Interpreter:   yes
        LLVM:          yes(600)
        Suspend:       hybrid
```
## Stacktrace

```
  at System.Runtime.InteropServices.Marshal.Release (System.IntPtr pUnk) [0x0000d] in <81342d83acda4c5590ec19c7afdf26b9>:0
  at ConsoleApp.Program.Main
```

NattyNarwhal:
Fixes #15853

### DIFF
--- a/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
+++ b/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
@@ -1112,7 +1112,7 @@ namespace System.Runtime.InteropServices
 		{
 #if !MOBILE || WINAOT
 			if (pUnk == IntPtr.Zero)
-				throw new ArgumentException ("Value cannot be null.", "pUnk");
+				throw new ArgumentNullException ("Value cannot be null.", "pUnk");
 
 			return ReleaseInternal (pUnk);
 #else
--- a/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
+++ b/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
@@ -66,7 +66,7 @@ namespace System.Runtime.InteropServices
 		{
 #if !MOBILE || WINAOT
 			if (pUnk == IntPtr.Zero)
-				throw new ArgumentException ("Value cannot be null.", "pUnk");
+				throw new ArgumentNullException ("pUnk");
 			return AddRefInternal (pUnk);
 #else
 			throw new NotImplementedException ();
@@ -939,7 +939,7 @@ namespace System.Runtime.InteropServices
 		{
 #if !MOBILE || WINAOT
 			if (pUnk == IntPtr.Zero)
-				throw new ArgumentException ("Value cannot be null.", "pUnk");
+				throw new ArgumentNullException ("pUnk");
 			return QueryInterfaceInternal (pUnk, ref iid, out ppv);
 #else
 			throw new NotImplementedException ();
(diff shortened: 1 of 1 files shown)


steveisok:
@monojenkins build failed

steveisok:
@monojenkins build failed

steveisok:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
```diff
@@ -1112,7 +1112,7 @@ public static int Release (IntPtr pUnk)
 		{
 #if !MOBILE || WINAOT
 			if (pUnk == IntPtr.Zero)
-				throw new ArgumentException ("Value cannot be null.", "pUnk");
+				throw new ArgumentNullException ("Value cannot be null.", "pUnk");
```
jonorossi: I've been caught by this before but the constructors of these 2 exceptions are reversed:

```csharp
public ArgumentException (string message, string paramName);
public ArgumentNullException (string paramName, string message);
```

akoeplinger: Good catch, thank you! I updated to the single-argument constructor since we don't need to set the message.

Issue #14342, opened 2019-05-03
Mono.Debugger.Soft tests assert that condition `!image->reflection_info_unregister_classes || mono_runtime_is_shutting_down ()' not met


lambdageek:
Seen this as a flaky failure few times on the "OS X x64 Android SDK" PR lanes.

For example: https://jenkins.mono-project.com/view/Pull%20requests/job/test-mono-pull-request-amd64-osx-products-sdks-android/11578/parsed_console/log.html

The last couple of things that happen in the debuggee are:

```
Unloading domain TEST RUNNER[0x765662ec3280], assembly foo[0x765662e68b00], ref_count=1
Unloading assembly foo [0x765662e68b00].
Unloading image foo [0x76564eec6000].
Unloading image foo [0x76564eeeb000].
* Assertion at /Users/builder/jenkins/workspace/test-mono-pull-request-amd64-osx-products-sdks-android/mono/metadata/image.c:2428, condition `!image->reflection_info_unregister_classes || mono_runtime_is_shutting_down ()' not met
```

So it looks like a non-root domain is unloading and but some SRE image hasn't properly unregistered its classes.

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-osx-products-sdks-android/14319/parsed_console/log_content.html

https://github.com/mono/mono/issues/15891

akoeplinger:
I _think_ this might also be the reason behind other Mono.Debugger.Soft flaky failures e.g. like https://jenkins.mono-project.com/job/test-mono-pull-request-interpreter-armhf/17946/ since it happens in the same methods as on that Android SDK lane.

We probably just don't see the assert since it happens in the child process and we only log stdout/stderr for the Android SDK lane, so it shows up as a regular Mono.Debugger.Soft.VMDisconnectedException in other lanes.

PR #15868, opened 2019-07-28
[Coop] Convert mono_attach_load_agent, fix leaks, reduce memory allocation/copy.

jaykrell:


jaykrell:
@monojenkins build failed

mono/metadata/attach.c
```diff
@@ -577,8 +571,11 @@ receiver_thread (void *arg)
 			agent_args = decode_string_value (p, &p, p_end);
 
 			printf ("attach: Loading agent '%s'.\n", agent_name);
-			mono_attach_load_agent (mono_domain_get (), agent_name, agent_args, &exc);
-
+			ERROR_DECL (error);
+			mono_attach_load_agent (mono_domain_get (), agent_name, agent_args, error);
+			mono_error_cleanup (error);
+			g_free (agent_name);
```
jaykrell: Cmd also leaks.
Protocol should probably null terminate to skip the alloc+copy+leak.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

mono/metadata/attach.c
```diff
@@ -293,51 +289,44 @@ mono_attach_load_agent (MonoDomain *domain, char *agent, char *args, MonoObject
...
 		}
-		mono_array_set_internal (main_args, MonoString*, 0, args_str);
+		MONO_HANDLE_ARRAY_SETREF (main_args, 0, args_str);
 	}
 
-
-	pa [0] = main_args;
-	mono_runtime_try_invoke (method, NULL, pa, exc, error);
+	pa [0] = MONO_HANDLE_RAW (main_args);
+	mono_runtime_try_invoke_handle (method, NULL_HANDLE, pa, error);
```
jaykrell: This loss of `exc` parameter always bugs me.

jaykrell: I think it is ok. Having both parameters is confusing. Changing code with both also confusing.

jaykrell: The more I convert other code, the more I dislike that I removed this parameter.
I will add it back in another PR, and maybe this should get it too, for clarity.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Hit https://github.com/mono/mono/issues/14342.

mono/metadata/object-internals.h
```diff
@@ -2143,12 +2143,15 @@ mono_runtime_object_init_checked (MonoObject *this_obj, MonoError *error);
...
+// The exc parameter was deliberately missing and so far this has proven to reduce code duplication.
 // In particular, if an exception is returned from underlying otherwise succeeded call,
 // is set into the MonoError with mono_error_set_exception_instance.
 // The result is that caller need only check MonoError.
+//
+// FIXME The removal of exc parameter confuses ongoing coop conversion. It has been restored.
+// Prior conversion should be revisited (despite long term success).
 MonoObjectHandle
-mono_runtime_try_invoke_handle (MonoMethod *method, MonoObjectHandle obj, void **params, MonoError* error);
+mono_runtime_try_invoke_handle (MonoMethod *method, MonoObjectHandle obj, void **params, MonoObjectHandleOut exc, MonoError* error);
```
lambdageek: I liked the fixes of the memory leaks in `mono_attach_load_agent`.

1. I don't understand why the seemingly unrelated change to `mono_runtime_try_invoke_handle` is also in this PR.  What am I missing?

2. We _want_ all exceptions to go via `MonoError` in this function.  Having both a `MonoError` and an exception out-arg is a step back.

jaykrell: `mono_runtime_try_invoke_handle`

This is a confusing thing agreed.
I've been round and round on this, at least with myself.

In the past I have come upon `mono_runtime_try_invoke`
and a need to convert pointer to handle, at least twice.

The first time I looked closely, removed the exc parameter,
and convinced myself it was semantica preserving.

The second time I was again puzzled, looked closely, and
remained convinced it was a correct change.

Here we are a third time.
Again the code is not obvious and I get tired of trying
to prove to myself what it does.

I suggest a conversion should have and now can be done
that more closely matches the original.

I believe the old conversion should be revisited.

This PR converts the new case, but does not yet reconvert the old ones.

You say it is a step back. Maybe you are right.
But I find it is a confusing change and would prefer to preserve
the old semantics, assuming they were ok.

jaykrell:
I still think 
https://github.com/mono/mono/pull/15868/commits/16aeab943e22fa9195633ef838029971943491f1

is the right idea.

However I have split this up into:
https://github.com/mono/mono/pull/15954
and
https://github.com/mono/mono/pull/15950

and since the exception here was never initialized nor used, and the other raw pointer is `NULL`, it is kinda ok and I still use the same raw pointer invoke.

Only kinda. In that all the `invoke` calls probably should be "fully" converted too.

Maybe redo those few invoke conversions before messing with more of them.

lambdageek:
This was split up into two PRs (referenced in the comment above)

PR #15869, opened 2019-07-28
[jit] Optimize the initialize of MonoError locals.

vargaz:
Previously, we would emit a call to memset since the struct is large (> 100 bytes). Only the error code

### DIFF
--- a/mono/metadata/object-offsets.h
+++ b/mono/metadata/object-offsets.h
@@ -136,6 +136,8 @@ DECL_OFFSET(MonoProfilerCallContext, method)
 DECL_OFFSET(MonoProfilerCallContext, return_value)
 DECL_OFFSET(MonoProfilerCallContext, args)
 
+DECL_OFFSET(MonoError, init)
+
 #ifdef HAVE_SGEN_GC
 DECL_OFFSET(SgenClientThreadInfo, in_critical_region)
 DECL_OFFSET(SgenThreadInfo, tlab_next)
--- a/mono/mini/decompose.c
+++ b/mono/mini/decompose.c
@@ -1249,7 +1249,13 @@ mono_decompose_vtype_opts (MonoCompile *cfg)
 					g_assert (ins->klass);
 
 					EMIT_NEW_VARLOADA_VREG (cfg, dest, ins->dreg, m_class_get_byval_arg (ins->klass));
-					mini_emit_initobj (cfg, dest, NULL, ins->klass);
+
+					if (m_class_get_image (ins->klass) == mono_defaults.corlib && !strcmp (ins->klass->name, "MonoError")) {
+						// Used in icall wrappers, optimize initialization
+						MONO_EMIT_NEW_STORE_MEMBASE_IMM (cfg, OP_STOREI4_MEMBASE_IMM, dest->dreg, MONO_STRUCT_OFFSET (MonoError, init), 0);
+					} else {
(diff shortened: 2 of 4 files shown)


jaykrell:
Would be good to optimize the strcmp’s in general.
Like at least store lengths or the lower bits of length.

jaykrell:
@monojenkins build failed

jaykrell:
I had tried to do the alloc & init on native side but a test was sensitive to that. Still maybe to try, as the error might be sometimes essentially unused.

vargaz:
@monojenkins build

vargaz:
@monojenkins build failed

jaykrell:
icall-table.h:
It was the allocation not the initialization that was fragile.
Could maybe try again to move it to native.
@lambdageek had some reason he preferred it in managed, but I removed it there only to satisfy fragile test.
It could also be that the test needed the zeroing, not the placement. I don't know.

```
// Declare the function wrapper that takes raw handles and a MonoError and returns a raw pointer.
//
// FIXME The error variable is on the managed side instead of native
// only to satisfy fragile test external/coreclr/tests/src/CoreMangLib/cti/system/weakreference/weakreferenceisaliveb.exe.
// I.e. We should have ERROR_DECL instead of error_init and MonoError parameter

// should be a local instead of a parameter. The different is minor.
#define MONO_HANDLE_DECLARE_RAW(id, name, func, rettype, n, argtypes)	\
ICALL_EXPORT MONO_HANDLE_TYPE_RAWPOINTER (rettype)				\
func ## _raw ( MONO_HANDLE_FOREACH_ARG_RAW_ ## n argtypes MONO_HANDLE_COMMA_ ## n MonoError *error)

// Implement ves_icall_foo_raw over ves_icall_foo.
// Raw handles are converted to/from typed handles and the rest is passed through.
// This is for functions in icall-def.h.

#define MONO_HANDLE_IMPLEMENT(id, name, func, rettype, n, argtypes)	\
										\
MONO_HANDLE_DECLARE_RAW (id, name, func, rettype, n, argtypes)			\
{										\
	HANDLE_FUNCTION_ENTER ();						\
										\
	/* FIXME Should be ERROR_DECL but for fragile test. */			\
	error_init (error);							\
```

jaykrell:
https://jenkins.mono-project.com/job/test-mono-pull-request-coop/17773/parsed_console/log.html
```
    System.Threading.ThreadPools.Tests.ThreadPoolTests.QueueRegisterPositiveAndFlowTest [FAIL]
      Assert.True() Failure
      Expected: True
      Actual:   False
      Stack Trace:
          at System.Threading.Tests.ThreadTestHelpers.CheckedWait (System.Threading.WaitHandle wh) [0x00000] in /mnt/jenkins/workspace/test-mono-pull-request-coop/external/corefx/src/Common/tests/System/Threading/ThreadTestHelpers.cs:125 
          at System.Threading.ThreadPools.Tests.ThreadPoolTests+<>c__DisplayClass12_0.<QueueRegisterPositiveAndFlowTest>b__1 (System.Boolean isWaitForRegisteredWaitCallback) [0x0000f] in /mnt/jenkins/workspace/test-mono-pull-request-coop/external/corefx/src/System.Threading.ThreadPool/tests/ThreadPoolTests.cs:274 
 ```

jaykrell:
@monojenkins build failed

PR #15871, opened 2019-07-28
[jit] Align float/double Nan/Inf to uint conversion behavior with CoreCLR/LLVM/Roslyn

### LINKED ISSUE #14931: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToUIntTest Pass

```
System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToUIntTest(useInterpreter: True) [FAIL]
      Assert.Equal() Failure
      Expected: 2147483648
      Actual:   0
      Stack Trace:
           at System.Linq.Expressions.Tests.ConvertTests.VerifyFloatToUInt(Single value, Boolean useInterpreter)
           at System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToUIntTest(Boolean useInterpreter)
```

### LINKED ISSUE #14932: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertNullableFloatToUIntTest Pass
The failure is the same as the one in https://github.com/mono/mono/issues/14931.

Only reason this issue exists is to track individual failures.

### LINKED ISSUE #14942: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertNullableFloatToNullableUIntTest Pass
Fails w/ same stack trace as https://github.com/mono/mono/issues/14931.

Only reason this issue exists is to track individual failures.

### LINKED ISSUE #14943: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToNullableUIntTest Pass
Fails w/ same stack trace as #14931.

Only reason this issue exists is to track individual failures.

filipnavara:
Fixes #14931
Fixes #14932
Fixes #14942
Fixes #14943

The conversions for `unchecked((ushort)double.NaN)` and `unchecked((byte)double.NaN)` produce zero. This aligns the behaviour for `unchecked((uint)double.NaN)` and `unchecked((ulong)double.NaN)` to also produce zero. This is important because the C# compiler does constant folding and the results for `return unchecked((uint)double.NaN)` and `var a = double.NaB; return unchecked((uint)a)` would be different.

For cross-reference, [the CoreCLR code](https://github.com/dotnet/coreclr/blob/ef7767a3ba1c0a34b55cbd5496b799b17218ca14/src/jit/codegenxarch.cpp#L6833-L6881) on x64 always generates 64-bit conversion instruction for float -> unsigned type conversion. This ensures that the exceptional NaN value (0x8000_0000_0000_0000) gets truncated to zero for data type smaller than 64 bits. The conversion to `ulong` is handled by a [helper call](https://github.com/dotnet/coreclr/blob/9bd2787a9dd2aa4d2b7d4f72afebc3dbe896e896/src/vm/jithelpers.cpp#L597-L612), just like in Mono.


filipnavara:
Here's one more reason why this needs to be fixed (even though the behavior may be unspecified):
```
csharp> float value = float.NaN;
csharp> (uint)value 
2147483648
csharp> unchecked((uint)float.NaN)
0  
```

The constant value produced is different from the runtime value produced from the same input.

filipnavara:
For cross-reference, [the CoreCLR code](https://github.com/dotnet/coreclr/blob/ef7767a3ba1c0a34b55cbd5496b799b17218ca14/src/jit/codegenxarch.cpp#L6833-L6881) on x64 always generates 64-bit conversion for float -> unsigned type conversion. The conversion to `ulong` is handled by a helper call because it could not be done in one instruction.

filipnavara:
Test program (should print zeros only):
```csharp
using System;

namespace HelloWorld
{
    class Program
    {
        static double a = double.NaN;
        static float fa = float.NaN;

        {
            Console.WriteLine(unchecked((ulong)double.NaN));        
            Console.WriteLine(unchecked((ulong)a));        
            Console.WriteLine(unchecked((uint)double.NaN));        
            Console.WriteLine(unchecked((uint)a));        
            Console.WriteLine(unchecked((ulong)float.NaN));        
            Console.WriteLine(unchecked((ulong)fa));        
            Console.WriteLine(unchecked((uint)float.NaN));        
            Console.WriteLine(unchecked((uint)fa));        
        }
    }
}
```

filipnavara:
The WebAssembly build failure is related. I will fix it.

filipnavara:
Linux AArch64 FullAOT failed on use case that I intentionally modified to behave same as CoreCLR. I don't have access to Windows ARM64 machine but I am reasonably sure the behavior of full .NET Framework is identical (if there is even ARM64 NetFX). I am going to fix the test unless someone objects to it.

filipnavara:

filipnavara:
D'oh. The WASM failure is in one of the new tests: `WASM: Test 'test_0_rconv_u4' failed result (got 2, expected 0).`

filipnavara:

filipnavara:
@monojenkins build Linux x64

PR #15873, opened 2019-07-29
HANDLE_FUNCTION_RETURN_VAL must not use handles in current frame.

jaykrell:


jaykrell:
@monojenkins build failed

lambdageek:

mono/metadata/handle.h
```diff
@@ -217,6 +217,8 @@ Icall macros
 	MONO_RESTORE_WARNING
 
 // Return a non-pointer or non-managed pointer, e.g. gboolean.
+// VAL should be a local variable or at least not use handles in the current frame.
+// i.e. it is "val", not "expr".
 #define HANDLE_FUNCTION_RETURN_VAL(VAL)		\
```
jaykrell: Would have been a great place for C++ auto, if a temporary was even needed. i.e. really just a good place for `return` and destructor.

PR #15883, opened 2019-07-29
[netcore] Replace dl-test-assets.py with msbuild

akoeplinger:
The python file was unreliable on macOS, let's see whether a simple msbuild replacement is better.

Also fixed dependencies on some of the Makefile targets so that they correctly detect when they need to run.


filipnavara:
I, for one, welcome this. Python3 on macOS Catalina betas had terribly broken mutiprocessing module to the point that it was constantly crashing and I had to rewrite the script locally as a workaround. This will also remove the Python dependency on Windows and we can finally enable the test runs there.

PR #15884, opened 2019-07-29
Remove unused libdl setting

coypoop:
We test dlopen with no library and in libdl later, without the use of this setting.

Got the hint since I'm building for a configuration which had it set wrong.


marek-safar:
approve

CoffeeFlux:
@monojenkins build failed

PR #15894, opened 2019-07-30
Move MonoError from managed wrappers to native wrappers.

jaykrell:
Move MonoError from managed wrappers (ilgen) to native wrappers (C preprocessor).

Alternative to https://github.com/mono/mono/pull/15869.
Which says: Previously, we would emit a call to memset
since the struct is large (> 100 bytes). Only the error code
Which that PR and this PR both fix, in different ways.

In the past:
- This did not work due to a possible dependency on precise GC,
  interacting with what gets zeroed or not here.
  That is why I put it on the managed side.
  We'll see if that reproduces.
  (https://github.com/mono/mono/pull/11294/files#diff-efe0070415f3f0c0f24ed0221aa1962aR339
  external/coreclr/tests/src/CoreMangLib/cti/system/weakreference/weakreferenceisaliveb.exe.)

- Some indication that embedding API should expose MonoError
  and not set_pending_exception. This does not seem relevant
  presently, and the code seems better structured and
  more efficient this way. It can revisited if/when new public API
  is the dominant concern, or do it this way for these functions,
  and another way for actual public functions.
  (https://github.com/mono/mono/pull/11294#issuecomment-432443485)

### DIFF
--- a/mcs/class/corlib/Mono/RuntimeStructs.cs
+++ b/mcs/class/corlib/Mono/RuntimeStructs.cs
@@ -48,14 +48,6 @@ namespace Mono {
 			internal IntPtr* data;
 			internal int len;
 		}
-
-		// mono-error.h MonoError
-		struct MonoError {
-			ushort error_code;
-			ushort hidden_0;
-			IntPtr hidden_1, hidden_2, hidden_3, hidden_4, hidden_5, hidden_6, hidden_7, hidden_8;
-			IntPtr hidden_11, hidden_12, hidden_13, hidden_14, hidden_15, hidden_16, hidden_17, hidden_18;
-		}
 	}
 
 	//Maps to metadata-internals.h:: MonoAssemblyName
--- a/configure.ac
+++ b/configure.ac
@@ -63,7 +63,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This line is parsed by tools besides autoconf, such as msvc/mono.winconfig.targets.
 # It should remain in the format they expect.
 #
-MONO_CORLIB_VERSION=811DFDEB-0230-4C72-91AD-A1AC9CFDA823
+MONO_CORLIB_VERSION=4e2d4257-1f2b-418c-9829-dd006fe09df5
 
 #
 # Put a quoted #define in config.h.
--- a/mcs/class/corlib/LinkerDescriptor/mscorlib.xml
+++ b/mcs/class/corlib/LinkerDescriptor/mscorlib.xml
@@ -843,9 +843,6 @@
 		-->
 		<type fullname="Mono.Runtime" preserve="nothing" />
 
-		<!-- marshal-ilgen.c (emit_native_icall_wrapper_ilgen) -->
-		<type fullname="Mono.RuntimeStructs/MonoError" />
-
 		<type fullname="System.Threading.WasmRuntime"/>
 
 		<!-- mini-generic-sharing.c -->
--- a/mono/metadata/icall-table.h
+++ b/mono/metadata/icall-table.h
@@ -367,9 +367,9 @@ typedef MonoStringHandle MonoStringOutHandle;
 #define MONO_HANDLE_FOREACH_TYPE_TYPED_8(t0, t1, t2, t3, t4, t5, t6, t7)     MONO_HANDLE_FOREACH_TYPE_TYPED_7 (t0, t1, t2, t3, t4, t5, t6)	,MONO_HANDLE_TYPE_TYPED (t7)
 #define MONO_HANDLE_FOREACH_TYPE_TYPED_9(t0, t1, t2, t3, t4, t5, t6, t7, t8) MONO_HANDLE_FOREACH_TYPE_TYPED_8 (t0, t1, t2, t3, t4, t5, t6, t7)	,MONO_HANDLE_TYPE_TYPED (t8)
 
-// Generate a parameter list, types and names, for a function accepting raw handles and a MonoError,
-// and returning a raw pointer. MonoError is not here, but added elsewhere.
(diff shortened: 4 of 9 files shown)


jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This produces some new warnings maybe, I’ll fix.

jaykrell:
A reason against this in the past was https://github.com/mono/mono/pull/11294#issuecomment-432443485
but I think it is still preferable this way for the forseeable future.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build

jaykrell:
@monojenkins build failed

jaykrell:
There is just the one error, signature mismatch on WebAssembly, that is difficult to understand.
The reported signatures match. If they didn't, the other platforms would have runtime problems possibly.
I have little experience building WebAssembly locally but am trying.

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
- Slightly less IL generated. Faster to interpret.
 - Less offsets.
 - Less special casing in method-to-ir.
  - Possibly faster, if the definition and wrapper and set_pending are visible together, and there is no error path, the MonoError will go away.
 - How I originally intended, but a precise-GC test required otherwise.
I like it.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
This needs a newer wasm-ld to link w/o error, and maybe run, maybe not.
I am unable to correlate emsdk to wasm-ld, except I can see it is a few months behind and this area of wasm-ld is under active development.

jaykrell:
@monojenkins build OS X WebAssembly

jaykrell:
wasm-ld in emsdk 1.38.43 still too old for this.
Maybe can workaround, by marking the wrapper as visibility=default (C) instead of visibility=hidden (IL) The mismatch seems to be visibility hidden vs. default, and I not sure changing default to hidden will work. Maybe try that. Hm.

jaykrell:
emcc 1.38.43 still not new enough.
But eventually wasm-ld does fix this the remaining problem here.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
Still fails, will try more locally. It'd seem like it should work-around-able, while waiting for wasm-ld fix.

jaykrell:
Try again with the wasm bump (should see what wams-ld it has, really).

jaykrell:
Nice, it passes WebAssembly now.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
/cc @lambdageek 
You had a point against this in the past but I still prefer it.
There was a precise-gc-dependent test, and a wams-ld bug in the way.
The precise-gc thing isn't showing.
The wasm-ld bug got fixed.
As you pointed out at the time, this will help the interpreter.
You objected that you wanted a public API with MonoError.
I suggest we can fix this if/when that API materializes.
And we can have both, really. JIT can still call these.
Public could be another set, but granted, that is wasteful.

I somewhat hope public API has faster to produce coop handles, however that contradicts abstracting them at the same time.

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
@monojenkins build failed

mcs/class/corlib/Mono/RuntimeStructs.cs
```diff
@@ -48,14 +48,6 @@ internal unsafe struct GPtrArray {
...
 			internal int len;
 		}
-
-		// mono-error.h MonoError
-		struct MonoError {
-			ushort error_code;
-			ushort hidden_0;
-			IntPtr hidden_1, hidden_2, hidden_3, hidden_4, hidden_5, hidden_6, hidden_7, hidden_8;
-			IntPtr hidden_11, hidden_12, hidden_13, hidden_14, hidden_15, hidden_16, hidden_17, hidden_18;
-		}
```
lambdageek: Follow-up task: 
1. remove `MonoError` from `mcs/class/corlib/LinkerDescriptor/mscorlib.xml`
2. remove `MonoError` from `netcore/System.Private.CoreLib/src/LinkerDescriptor/System.Private.CoreLib.xml`

jaykrell:
Is 10 days old also ok?
Maybe rebase to master *now* and do the followup at the same time?

jaykrell:
Rebased to master, did followup, revised corlib again.

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins squash

lambdageek:

jaykrell:
monolite is supposed to suffice now, new thing, does less, faster, than "deb with monolite".
I haven't used it yet.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

Issue #15125, opened 2019-06-14
csc crashes randomly in get_image_for_container


rolfbjarne:
## Steps to Reproduce

Unfortunately I don't have a reliable way to reproduce this, it's a random issue (so far I've seen it twice on bots only).

Possible way to reproduce: build xamarin-macios' `xcode11` branch in a loop, something like this (untested, but should be close enough):

```
git clone git@github.com:xamarin/xamarin-macios
cd xamarin-macios
git checkout -b xcode11 origin/xcode11
while make world; do
    echo "Trying again"
    make git-clean-all
done
```

### Expected Behavior

No crash.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

```
mono --version
Mono JIT compiler version 6.0.0.286 (2019-02/c22486e39d1 Thu Jun  6 16:31:00 EDT 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)

```

## Stacktrace

I have two stack traces that look identical (note that the crash occurred when compiling different assemblies, so it's not a particular csc input that makes it crash).

[Crash 1 build log](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10287/consoleText)
[Crash 2 build log](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10282/consoleText)

<details>
<summary>Crash 1</summary>

```
=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x10f0b5644 - /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/builds/tools64/mono/mini/mono : dump_native_stacktrace
	0x10f0b54af - /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/builds/tools64/mono/mini/mono : mono_dump_native_crash_info
[...]

PR #15895, opened 2019-07-30
Wrap more external-only functions in GC mode transitions.

jaykrell:
mono_compile_method
mono_runtime_set_main_args
mono_runtime_run_main
mono_runtime_exec_managed_code
mono_load_remote_field
mono_load_remote_field_new
mono_store_remote_field
mono_store_remote_field_new

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -744,9 +744,16 @@ mono_set_always_build_imt_trampolines (gboolean value)
 gpointer 
 mono_compile_method (MonoMethod *method)
 {
+	gpointer result;
+
+	MONO_ENTER_GC_UNSAFE;
+
 	ERROR_DECL (error);
-	gpointer result = mono_compile_method_checked (method, error);
+	result = mono_compile_method_checked (method, error);
 	mono_error_cleanup (error);
+
+	MONO_EXIT_GC_UNSAFE;
+
 	return result;
 }
 
@@ -4505,6 +4512,8 @@ free_main_args (void)
 int
 mono_runtime_set_main_args (int argc, char* argv[])
 {
+	MONO_ENTER_GC_UNSAFE;
+
 	MONO_REQ_GC_NEUTRAL_MODE;
 
 	int i;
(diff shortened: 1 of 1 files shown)


jaykrell:
https://github.com/mono/mono/issues/15125

jaykrell:
@monojenkins build failed

CoffeeFlux:
I don't really see a compelling reason to rename the macros, and having a mix used throughout the codebase seems unnecessarily confusing.

jaykrell:
If they need to do more in future would be the reason.
Truths: mixed like you say; hypothetical
I can undo that part.
Not sure about mode & exceptions.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@lambdageek I wonder if these were omitted on purpose.
And about test coverage, of extern-only.
Or perhaps it is obviously correct to you.

mono/metadata/object.c
```diff
@@ -4645,17 +4656,25 @@ int
 mono_runtime_run_main (MonoMethod *method, int argc, char* argv[],
 		       MonoObject **exc)
 {
+	int res;
+
 	MONO_REQ_GC_UNSAFE_MODE;
```
lambdageek: nit: Should remove this.  It doesn't make sense to require the thread to already be in GC Unsafe and then immediately switch to GC Unsafe

mono/metadata/object.c
```diff
@@ -4505,6 +4512,8 @@ free_main_args (void)
 int
 mono_runtime_set_main_args (int argc, char* argv[])
 {
+	MONO_ENTER_GC_UNSAFE;
+
 	MONO_REQ_GC_NEUTRAL_MODE;
```
lambdageek: Doesn't make sense to require GC Neutral mode immediately after switching to GC Unsafe.

But also this function doesn't need to be in GC Unsafe - it only interacts with eglib stuff.

jaykrell: Oh, the second part. So undo this? I guess I'll look more closely, instead of just following "external only" marks, and replace with a comment.

mono/metadata/object.c
jaykrell: I didn't get around to the nit. :(

mono/metadata/object.c
jaykrell: I didn't get around to this either.

PR #15898, opened 2019-07-30
If either sched_get_priority_{min,max} failed, don't try to set a priority

coypoop:
netbsd and linux both require that if SCHED_OTHER is used, a fixed value is used for sched_param ([0 for linux](http://man7.org/linux/man-pages/man7/sched.7.html), -1 for netbsd)

the netbsd case can also be detected by the [POSIX-defined failure](https://pubs.opengroup.org/onlinepubs/007908799/xsh/sched_get_priority_min.html) return value for sched_get_priority_min, a return value of -1.

### DIFF
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -791,7 +791,6 @@ mono_thread_internal_set_priority (MonoInternalThread *internal, MonoThreadPrior
 	if ((min == -1) || (max == -1))
 		return;
 
-
 	if (max > 0 && min >= 0 && max > min) {
 		double srange, drange, sposition, dposition;
 		srange = MONO_THREAD_PRIORITY_HIGHEST - MONO_THREAD_PRIORITY_LOWEST;
--- a/mono/metadata/threads.c
+++ b/mono/metadata/threads.c
@@ -787,6 +787,11 @@ mono_thread_internal_set_priority (MonoInternalThread *internal, MonoThreadPrior
 #endif
 	MONO_EXIT_GC_SAFE;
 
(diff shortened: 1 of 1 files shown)


coypoop:
Another possibly reasonable option, is that if we don't know how to map mono priorities to pthread posix ones, don't do anything. It's described in [this commit](https://github.com/coypoop/mono/commit/2f048940a5ae636f89c759887237427c2b485bdc).

But I imagine it is a scarier commit that needs to be performance-tested.

coypoop:
To clarify, this is because NetBSD returns an error for
````pthread_setschedparam(..., SCHED_OTHER, sched_priority=0)````
And this is treated as a fatal error by Mono.

File: mono/metadata/threads.c [hunk 1]
```diff
@@ -784,6 +784,10 @@ mono_thread_internal_set_priority (MonoInternalThread *internal, MonoThreadPrior
 #else
 	min = sched_get_priority_min (policy);
 	max = sched_get_priority_max (policy);
+
+	/* Not tunable. Bail out */
+	if ((min == -1) || (max == -1))
+		return;
 #endif
 	MONO_EXIT_GC_SAFE;
```
akoeplinger: We need to ensure we do the MONO_EXIT_GC_SAFE even if we return from the `-1` case.

I'd just move the checks here outside the #if, assuming PRIORITY_MIN isn't -1 on `__PASE__` which is probably a safe bet.

steveisok:
@monojenkins build failed

akoeplinger:
@monojenkins build

PR #15899, opened 2019-07-30
Move mono_invoke_unhandled_exception_hook from exception.h to -internals.h.

jaykrell:
It isn't marked MONO_API.
People usually dynamicall link, right?

I'm converting to coop, and if this flies, the name can be recycled.


jaykrell:
I guess it is for the statically linked driver, ugh.

PR #15900, opened 2019-07-30
[Coop] [wip] [draft] @monojenkins skip Conversion work in progress.

jaykrell:
[Coop] Reduce raw pointers esp. from mono_string_new_checked.
[Coop] Remove raw pointers leftover in locales.c.
[Coop] Convert or partially convert mono_domain_try_unload.
[Coop] Convert or partially convert mono_domain_unload.
[Coop] Convert or partially convert mono_gc_run_finalize.
[Coop] Convert or partially convert mono_invoke_unhandled_exception_hook.
[Coop] Convert or partially convert mono_jit_exec_internal.
[Coop] Convert or partially convert mono_load_remote_field_checked
[Coop] Convert or partially convert mono_message_init.
[Coop] Convert or partially convert mono_object_try_to_string[_handle]
[Coop] Convert or partially convert mono_print_unhandled_exception_internal.
[Coop] Convert or partially convert mono_remoting_invoke.
[Coop] Convert or partially convert mono_runtime_class_init_full.
[Coop] Convert or partially convert mono_runtime_exec_main
[Coop] Convert or partially convert mono_runtime_run_main.
[Coop] Convert or partially convert mono_runtime_try_invoke_handle.
[Coop] Convert or partially convert mono_runtime_try_run_main.
[Coop] Convert or partially convert mono_string_new.
[Coop] Convert or partially convert mono_string_new_internal.
[Coop] Convert or partially convert mono_string_new_utf32.
[Coop] Convert or partially convert mono_string_new_utf32_checked.
[Coop] Convert or partially convert mono_string_new_wrapper_internal_impl.
[Coop] Convert or partially convert prepare_run_main.
[Coop] Convert or partially convert ves_icall_System_AppDomain_InternalUnload.
[Coop] Convert or partially convert ves_icall_System_Runtime_Remoting_Messaging_AsyncResult_Invoke.

PR #15901, opened 2019-07-30
[Coop] Convert ves_icall_string_new_wrapper_impl.

jaykrell:
Static analysis prefers not even these short lived raw pointers.
Extracted from https://github.com/mono/mono/pull/15900.


jaykrell:
@monojenkins build API Diff

PR #15902, opened 2019-07-30
[Coop] Convert mono_string_from_byvalwstr_impl.

jaykrell:
Remove the raw pointer that static analysis will flag.
Extracted from https://github.com/mono/mono/pull/15900.


jaykrell:
Convert as in the remove the raw pointer that static analysis will not like.

jaykrell:
@monojenkins build failed

Issue #14951, opened 2019-06-10
[netcore] Make System.Linq.Expressions.Tests.CallTests.MethodName_TypeArgsDontMatchConstraints_ThrowsArgumentException Pass


steveisok:
Test currently fails with:

```
System.Linq.Expressions.Tests.CallTests.MethodName_TypeArgsDontMatchConstraints_ThrowsArgumentException [FAIL]
      Assert.Equal() Failure
      Expected: (null)
      Actual:   typeArguments
      Stack Trace:
           at System.AssertExtensions.Throws[ArgumentException](String paramName, Func`1 testCode)
           at System.Linq.Expressions.Tests.CallTests.MethodName_TypeArgsDontMatchConstraints_ThrowsArgumentException()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

Issue #14950, opened 2019-06-10
[netcore] Make System.Linq.Expressions.Tests.CallTests.EnumArgAndReturn Pass


steveisok:
Test currently fails with:

```
System.Linq.Expressions.Tests.CallTests.EnumArgAndReturn(useInterpreter: True) [FAIL]
      System.ArgumentException : Cannot bind to the target method because its signature is not compatible with that of the delegate type.
      Stack Trace:
           at System.Delegate.CreateDelegate(Type type, Object firstArgument, MethodInfo method, Boolean throwOnBindFailure, Boolean allowClosed)
           at System.Delegate.CreateDelegate(Type type, MethodInfo method, Boolean throwOnBindFailure)
           at System.Delegate.CreateDelegate(Type type, MethodInfo method)
           at System.Reflection.RuntimeMethodInfo.CreateDelegate(Type delegateType)
           at System.Linq.Expressions.Interpreter.FuncCallInstruction`2[[System.Byte, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Int32, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]..ctor(MethodInfo target)
           at System.Linq.Expressions.Interpreter.CallInstruction.FastCreate[Byte,Int32](MethodInfo target, ParameterInfo[] pi)
           at System.Linq.Expressions.Interpreter.CallInstruction.FastCreate[Byte](MethodInfo target, ParameterInfo[] pi)
           at System.Linq.Expressions.Interpreter.CallInstruction.FastCreate(MethodInfo target, ParameterInfo[] pi)
           at System.Linq.Expressions.Interpreter.CallInstruction.Create(MethodInfo info, ParameterInfo[] parameters)
           at System.Linq.Expressions.Interpreter.InstructionList.EmitCall(MethodInfo method, ParameterInfo[] parameters)
           at System.Linq.Expressions.Interpreter.LightCompiler.CompileMethodCallExpression(Expression object, MethodInfo method, IArgumentProvider arguments)
           at System.Linq.Expressions.Interpreter.LightCompiler.CompileMethodCallExpression(Expression expr)
           at System.Linq.Expressions.Interpreter.LightCompiler.CompileNoLabelPush(Expression expr)
           at System.Linq.Expressions.Interpreter.LightCompiler.Compile(Expression expr)
[...]

Issue #14909, opened 2019-06-10
[netcore] Make BasicEventSourceTests.TestsWrite.Test_Write_T_EventListener Pass


steveisok:
The test currently fails with:

```
    BasicEventSourceTests.TestsWrite.Test_Write_T_EventListener [FAIL]
      BasicEventSourceTests.EventTestHarness+EventTestHarnessException : Error Detected in EventTestHarness.RunTest
      FAILURE IN SUBTEST: "Write/Enum/EnumEventBasicEventSourceTests.Color"
      ************* EXCEPTION INFO ***************
      System.ArgumentException: Cannot bind to the target method because its signature is not compatible with that of the delegate type.
         at System.Delegate.CreateDelegate(Type type, Object firstArgument, MethodInfo method, Boolean throwOnBindFailure, Boolean allowClosed)
         at System.Delegate.CreateDelegate(Type type, MethodInfo method, Boolean throwOnBindFailure)
         at System.Delegate.CreateDelegate(Type type, MethodInfo method)
         at System.Reflection.RuntimeMethodInfo.CreateDelegate(Type delegateType)
         at System.Diagnostics.Tracing.PropertyValue.TypeHelper.GetGetMethod(PropertyInfo property, Type propertyType)
         at System.Diagnostics.Tracing.PropertyValue.ReferenceTypeHelper`1[[<>f__AnonymousType369`3[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[BasicEventSourceTests.Color, System.Diagnostics.Tracing.Tests, Version=4.2.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51],[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Diagnostics.Tracing.Tests, Version=4.2.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51]].GetPropertyGetter(PropertyInfo property)
         at System.Diagnostics.Tracing.PropertyValue.GetReferenceTypePropertyGetter(PropertyInfo property)
         at System.Diagnostics.Tracing.PropertyValue.GetPropertyGetter(PropertyInfo property)
[...]

PR #15905, opened 2019-07-30
[netcore] Minor fixes around System.Reflection

EgorBo:
Fixes https://github.com/mono/mono/issues/14909 (fixed by my previous PR)
Fixes https://github.com/mono/mono/issues/14950
Fixes https://github.com/mono/mono/issues/14951


netcore/sample/HelloWorld/HelloWorld.csproj
```diff
@@ -4,6 +4,7 @@
     <OutputType>Exe</OutputType>
     <OutputPath>bin</OutputPath>
     <TargetFramework>netcoreapp3.0</TargetFramework>
+    <DebugType>full</DebugType>
```
akoeplinger: what's the reason for this change? netcore should use portable pdb everywhere

EgorBo: unwanted change, removed

vargaz:
@monojenkins build failed

PR #15859, opened 2019-07-27
Thread name improvements including coop conversion.

jaykrell:
- Convert setting thread name to be coop-compatible.

 - Set proper Windows kernel unicode thread names, not just the debugger ones.
  These don't require a debugger to be useful, can be retrieved, show up in ETW, etc.

 - For constant thread names, just retain the constant, not a copy.
   This includes producing constant unicode thread names on Windows.

 - For non-constant thread names, retain the UTF8, that it was going
   to produce anyway, instead of the unicode. While Unicode
   benefits Windows, retaining it does not.
   A downside here is that it moves around failure points.
   In particular, setting a managed thread name involved a memdup,
   now it involves utf16 to utf8 conversion, which hopefully succeeds.

 - Make setting a thread name often non-fatal, except where
   mandated by public API.

 - Switch two boolean parameters, that could use extension to three, to enums, for
   call site readability.

- Retain nul in thread name. It didn't matter before but is useful now.

- This does penalize ves_icall_System_Threading_Thread_GetName_internal, the one
  code path that benefited from retaining UTF16 here -- lossless/conversion-free passthrough from Set to Get, is nice.
  Arguably we should retain both.
  Notice that netcore removes this function.

- Factor copy/pasted code into new small function mono_gstring_append_thread_name.

- Remove unused fields from MonoInternalThread. But synchronize netcore and non-netcore size with the dream of netcore-ness being a command line switch.

Alternative to https://github.com/mono/mono/pull/15827.

PR #15910, opened 2019-07-30
Cleanup and resynchronize Thread/InternalThread fields.

jaykrell:
- Remove unused.
 - Add missing that just happened to work.
 - Synchronize size between netcore and regular. `#if` considered bad.

### DIFF
--- a/netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
+++ b/netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
@@ -20,7 +20,6 @@ namespace System.Threading
 		// stores a thread handle
 		IntPtr handle;
 		IntPtr native_handle; // used only on Win32
-		IntPtr unused3;
 		/* accessed only from unmanaged code */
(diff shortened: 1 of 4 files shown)


jaykrell:

jaykrell:
@monojenkins build failed

netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
```diff
@@ -59,6 +57,7 @@ partial class Thread
 		private IntPtr suspended_event;
 		private int self_suspended;
 		private IntPtr thread_state;
+
 		private Thread self;
 		private object pending_exception;
 		private object start_obj;
```
lambdageek: I realize this is not part of the PR, but`pending_exception` and `start_obj` are swapped between netcore managed and native.

jaykrell: Good catch. Maybe should really mechanize this soon.

PR #15915, opened 2019-07-30
Fix `Socket.ConnectAsync(SocketAsyncEventArgs)` behavior.

baulig:
Fix `Socket.ConnectAsync(SocketAsyncEventArgs)` behavior.

We need to clearly distinguish between synchronous and asynchronous completions and avoid any ambiguities and race conditions when it comes to error handling.

On synchronous completion, we must not invoke the async completion delegate.

In `System.Net.Sockets/Socket.cs`, the internal `BeginSConnect()` and `BeginMConnect()` now return a `bool` indicating whether or not an async operation is pending.

If any error happens prior to starting the async operation, then we shall call `SocketAsyncResult.Complete(..., true)` and return `false`.  And since this is a synchronous completion, it's `AsyncCallback` will not be invoked.

In `referencesource/System/net/System/Net/Sockets/_MultipleConnectAsync.cs`, I compared the entire file against the CoreFX version and then copied the `AttemptConnection()` method from their implementation to adapt their behavior.

The difference is that on synchronous completion,  the `InternalConnectCallback(null, args)` should be invoked.

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -912,20 +912,61 @@ namespace System.Net.Sockets
 			try {
 				IPAddress [] addresses;
 				SocketAsyncResult ares;
+				bool pending;
+
+				/*
+				 * Both BeginSConnect() and BeginMConnect() now return a `bool` indicating whether or
+				 * not an async operation is pending.
+				 */
 
 				if (!GetCheckedIPs (e, out addresses)) {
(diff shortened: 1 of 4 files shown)


baulig:
@monojenkins backport 2019-02

baulig:
@monojenkins backport 2019-06

baulig:
@monojenkins build pkg

steveisok:
@monojenkins build failed

baulig:
Recreated the same thing as https://github.com/mono/mono/pull/15947.  I was simply too lazy about manually editing both back ports.

PR #15919, opened 2019-07-31
Support proper Windows kernel Unicode thread names.

jaykrell:
These do not require a debugger, can be retrieved, appear in ETW, etc.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Oops. The macro that I thought was a macro is an inline function.
So let's instead have a macro to decide static vs. dynamic (GetProcAddress) vs. none.
Describing the same three cases, but with different names and comments.

jaykrell:
This time I stepped it in the debugger.
Though cdb only shows the other thread name -- the exception-based on -- even if I skipped over setting it, then it shows nothing.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/utils/mono-windows-thread-name.c
```diff
@@ -0,0 +1,149 @@
+// mono-windows-thread-name.c
+//
+// There several ways to set thread name on Windows.
```
jaykrell: `There are`

mono/utils/mono-windows-thread-name.c
```diff
@@ -0,0 +1,149 @@
...
+//
+// Historically: Raise an exception with an ASCII string.
+//  This is visible only in a debugger. Only if a debugger
+//  is running when the exception is raised. There is
+//  no way to get the thread name and they do not appear in ETW traces.
+//
+// XboxOne: SetThreadName(thread handle, unicode)
+//   This works with or without a debugger and can be retrieved with GetThreadName.
+//   Sometimes an inline function SetThreadDescription for source compat with next.
```
jaykrell: `compatibility`

mono/utils/mono-windows-thread-name.c
```diff
@@ -0,0 +1,149 @@
...
+//   would require LoadLibrary / GetProcAddress, but LoadLibrary is not allowed.
+//
+// Author:
+//  Jay Krell (jaykrell@microsoft.com)
+//
+// Copyright 2019 Microsoft
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+#include "config.h"
+#include "mono-threads.h"
```
jaykrell: `#include` should be move to after `#if`, but it doesn't matter, the file is also filtered in existing Makefiles.

mono/utils/mono-windows-thread-name.c
```diff
@@ -0,0 +1,149 @@
...
+
+	// Checking for IsDebuggerPresent here would be reasonable, for
+	// efficiency, and would let this work with other compilers (without
+	// structured exception handling support), however
+	// there is a race condition then, in that a debugger
+	// can be detached (and attached) at any time.
+	//
+	// A vectored exception handler could also be used to handle this exception.
+
+	__try {
```
jaykrell: The debugger check would ease kernel debugging.
But it'd also prevent a vectored handler from seeing the exception which is arguably useful.

jaykrell: In either case, this PR doesn't change the behavior here, and this is likely what 100% of code out there does, except that we also compile with gcc.

mono/metadata/w32subset.h
```diff
@@ -210,3 +210,24 @@
 #define HAVE_API_SUPPORT_WIN32_IS_WOW64_PROCESS 0
 #endif
 #endif
+
+#ifndef HAVE_LOADLIBRARY
+#if G_HAVE_API_SUPPORT(HAVE_CLASSIC_WINAPI_SUPPORT)
+#define HAVE_LOADLIBRARY 1
+#else
+#define HAVE_LOADLIBRARY 0
```
jaykrell: LoadLibrary is coming to UWP so we can fix this.

jaykrell: And it is an immediately effective policy, nothing technical, can be fixed right away.

PR #15921, opened 2019-07-31
mono_thread_set_name_internal change boolean parameters to enum flags.

jaykrell:


jaykrell:
@monojenkins build failed

vargaz:
Looks ok but why is this needed, are we adding more flags ?

jaykrell:
I want to add another flag called `constant`.
And I want to retain UTF8 instead of UTF16.
So, given UTF8, it will just store it.
Given "constant", when it is set again, it will not free it. The flag will be remembered as a boolean.
@CoffeeFlux was a little reluctant to review the larger PR.

https://github.com/mono/mono/pull/15859/files#diff-f1a9287559d3aa620b058619f47f00c9R1874
etc.

jaykrell:
@monojenkins build failed

PR #15922, opened 2019-07-31
[wasm][bcl] Change how HttpClient finds custom message handler.

kjpou1:
- Remove the `private static Func<HttpMessageHandler> GetHttpMessageHandler;` from the WASM HttpClient.
- Add an environment variable read (`WASM_HTTP_CLIENT_HANDLER_TYPE`) to override the default handler.
   - The `WASM_HTTP_CLIENT_HANDLER_TYPE` environment variable controls the default System.Net.Http.HttpMessageHandler implementation which will be used by the System.Net.Http.HttpClient default constructor. The value is an assembly-qualified type name of an HttpMessageHandler subclass, suitable for use with System.Type.GetType(string).
   - Example code for setting:

```
      Environment.SetEnvironmentVariable("WASM_HTTP_CLIENT_HANDLER_TYPE", typeof(MyCustomWasmHandler).AssemblyQualifiedName);
```
   - On an error the following may be seen:

```
      Uncaught Error: System.PlatformNotSupportedException: XXXXXXXXXXXXX is not a valid HttpMessageHandler
```

**Note** This is a breaking change to client code.

### DIFF
--- a/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
+++ b/mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
@@ -6,49 +6,46 @@ namespace System.Net.Http
 {
     public partial class HttpClient
     {
-
-#pragma warning disable 649
-        private static Func<HttpMessageHandler> GetHttpMessageHandler;
-#pragma warning restore 649
-
         internal static HttpMessageHandler CreateDefaultHandler()
         {
 
-            if (GetHttpMessageHandler == null)
+            string envvar = Environment.GetEnvironmentVariable ("WASM_HTTP_CLIENT_HANDLER_TYPE")?.Trim ();
+            if (String.IsNullOrEmpty (envvar))
             {
                 Type type = Type.GetType("WebAssembly.Net.Http.HttpClient.WasmHttpMessageHandler, WebAssembly.Net.Http");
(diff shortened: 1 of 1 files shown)


mcs/class/System.Net.Http/System.Net.Http/HttpClient.wasm.cs
```diff
@@ -6,49 +6,46 @@ namespace System.Net.Http
...
-
-#pragma warning disable 649
-        private static Func<HttpMessageHandler> GetHttpMessageHandler;
-#pragma warning restore 649
-
         internal static HttpMessageHandler CreateDefaultHandler()
         {
 
-            if (GetHttpMessageHandler == null)
+            string envvar = Environment.GetEnvironmentVariable ("WASM_HTTP_CLIENT_HANDLER_TYPE")?.Trim ();
```
marek-safar: Do we really need this?

kjpou1: It is something that @lewing and I were discussing on an alternate way to remove the static method.  

There is also this PR: https://github.com/mono/mono/pull/15051 that is more the XI style but it requires Blazor and others to buy into our full bindings stack.

kjpou1:
@monojenkins build failed

PR #15926, opened 2019-07-31
Save 64 bytes from most Linux/amd64 interpreter frames.

jaykrell:
Probably same on all 64bit architectures. half on 32bit.

before:
```
0000000000003ce0 <interp_exec_method_full>:
    3ce0:	55                   	push   %rbp
    3ce1:	48 89 e5             	mov    %rsp,%rbp
    3ce4:	41 57                	push   %r15
    3ce6:	41 56                	push   %r14
    3ce8:	41 55                	push   %r13
    3cea:	41 54                	push   %r12
    3cec:	49 89 f4             	mov    %rsi,%r12
    3cef:	53                   	push   %rbx
    3cf0:	48 89 fb             	mov    %rdi,%rbx
    3cf3:	48 81 ec 78 01 00 00 	sub    $0x178,%rsp
```

after:
```
0000000000003ce0 <interp_exec_method_full>:
    3ce0:	55                   	push   %rbp
    3ce1:	48 89 e5             	mov    %rsp,%rbp
    3ce4:	41 57                	push   %r15
    3ce6:	41 56                	push   %r14
    3ce8:	41 55                	push   %r13
    3cea:	41 54                	push   %r12
    3cec:	49 89 f4             	mov    %rsi,%r12
    3cef:	53                   	push   %rbx
    3cf0:	48 89 fb             	mov    %rdi,%rbx
    3cf3:	48 81 ec 38 01 00 00 	sub    $0x138,%rsp
```

It was 0x158 recently.

Some frames will be larger, i.e. dynamic + pinvoke.

Classic technique:
 pointer instead of struct
 initialized to null
 alloca it when needed and not null

A more detailed analysis might enable better, such as making
the parameter optional, and/or having the callee default it to its
local storage.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -235,7 +235,7 @@ static void debug_enter (InterpFrame *frame, int *tracing)
 static void
 set_resume_state (ThreadContext *context, InterpFrame *frame)
 {
-	/* We have thrown an exception from a finally block. Some of the leave targets were unwinded already */ \
+	/* We have thrown an exception from a finally block. Some of the leave targets were unwinded already */
 	while (frame->finally_ips &&
 		   frame->finally_ips->data >= context->handler_ei->try_start &&
 		   frame->finally_ips->data < context->handler_ei->try_end)
@@ -942,13 +942,9 @@ interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, gco
 
 #define THROW_EX(exception,ex_ip) THROW_EX_GENERAL ((exception), (ex_ip), FALSE)
 
-#define THROW_EX_OVF(ip) do { \
-	THROW_EX (mono_get_exception_overflow (), ip); \
-	} while (0)
+#define THROW_EX_OVF(ip) THROW_EX (mono_get_exception_overflow (), ip)
 
-#define THROW_EX_DIV_ZERO(ip) do { \
-	THROW_EX (mono_get_exception_divide_by_zero (), ip);	\
-	} while (0)
+#define THROW_EX_DIV_ZERO(ip) THROW_EX (mono_get_exception_divide_by_zero (), ip)
 
 #define NULL_CHECK(o) do { \
 	if (G_UNLIKELY (!(o))) \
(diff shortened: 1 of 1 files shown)


mono/mini/interp/interp.c
```diff
@@ -3259,14 +3257,15 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
 				MonoMarshalSpec **mspecs;
-				MonoMethodPInvoke piinfo;
 				MonoMethod *m;
 
 				/* Pinvoke call is missing the wrapper. See mono_get_native_calli_wrapper */
 				mspecs = g_new0 (MonoMarshalSpec*, csignature->param_count + 1);
-				memset (&piinfo, 0, sizeof (piinfo));
 
-				m = mono_marshal_get_native_func_wrapper (m_class_get_image (imethod->method->klass), csignature, &piinfo, mspecs, code);
+				piinfo = piinfo ? piinfo : g_newa (MonoMethodPInvoke, 1);
```
filipnavara: Where is it `free`d?

jaykrell: `a` is for `alloca`. When the function returns.

lewurm: https://github.com/mono/mono/blob/a6e6c72c8fc3702db37e1572694a49902f63195f/mono/eglib/glib.h#L292

probably not the best name.

filipnavara: Oh, I see.

jaykrell: It is subtle I agree.
Should I just spell out `alloca` and the cast?
`g_allocate_stack`? :)

filipnavara: It's fine I guess. The name `g_newa` is confusing since it doesn't imply it's `alloca` under the covers.

vargaz: Lets hope this is not called in a loop.

lewurm: @vargaz why would that matter?

vargaz: alloca cannot be called in a loop because it will cause a stack overflow.

lewurm: `piinfo` wouldn't be `NULL` anymore.

filipnavara: It is guarded to do the allocation only once so it would not matter here.

PR #15930, opened 2019-07-31
[jit] Align float/double Nan/Inf to uint conversion behavior with CoreCLR/LLVM/Roslyn

### LINKED ISSUE #14931: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToUIntTest Pass

### LINKED ISSUE #14932: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertNullableFloatToUIntTest Pass

### LINKED ISSUE #14942: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertNullableFloatToNullableUIntTest Pass

### LINKED ISSUE #14943: [netcore] Make System.Linq.Expressions.Tests.ConvertTests.ConvertFloatToNullableUIntTest Pass

filipnavara:
Fixes #14931
Fixes #14932
Fixes #14942
Fixes #14943

The conversions for `unchecked((ushort)double.NaN)` and `unchecked((byte)double.NaN)` produce zero. This aligns the behaviour for `unchecked((uint)double.NaN)` and `unchecked((ulong)double.NaN)` to also produce zero. This is important because the C# compiler does constant folding and the results for `return unchecked((uint)double.NaN)` and `var a = double.NaB; return unchecked((uint)a)` would be different.

For cross-reference, [the CoreCLR code](https://github.com/dotnet/coreclr/blob/ef7767a3ba1c0a34b55cbd5496b799b17218ca14/src/jit/codegenxarch.cpp#L6833-L6881) on x64 always generates 64-bit conversion instruction for float -> unsigned type conversion. This ensures that the exceptional NaN value (0x8000_0000_0000_0000) gets truncated to zero for data type smaller than 64 bits. The conversion to `ulong` is handled by a [helper call](https://github.com/dotnet/coreclr/blob/9bd2787a9dd2aa4d2b7d4f72afebc3dbe896e896/src/vm/jithelpers.cpp#L597-L612), just like in Mono.

Re-opening of #15871 (GitHub screwed up and stopped running new builds and then complained re-opening of PR is impossible because of forced push).


mono/metadata/jit-icall-reg.h
```diff
@@ -116,6 +116,7 @@ MONO_JIT_ICALL (__emul_rconv_to_i8) \
 MONO_JIT_ICALL (__emul_rconv_to_ovf_i8) \
 MONO_JIT_ICALL (__emul_rconv_to_ovf_u8) \
 MONO_JIT_ICALL (__emul_rconv_to_ovf_u8_un) \
+MONO_JIT_ICALL (__emul_rconv_to_u4)	\
 MONO_JIT_ICALL (__emul_rconv_to_u8) \
```
vargaz: Need to bump the aot file format version in aot-runtime.h.

filipnavara: Good point!

filipnavara:
Keeping it as draft until the CI goes green.

Issue #15687, opened 2019-07-15
[debugger] Update client thread frames after SetIP.


thaystg:
On interpreter it's not working. The test case MonoTests.DebuggerTests.SetIP2 should be enabled.

SamMonoRT:
@imhameed - please can you take a look

PR #15936, opened 2019-07-31
[sdb][interp] Use the interpreter's resume state to compute the IL offset in compute_frame_info.

imhameed:
This recovers the relevant bits of the "resume state" from the interpreter and uses that to recalculate the `il_offset` for the top frame.

Fixes https://github.com/mono/mono/issues/15687.

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -3093,6 +3093,21 @@ no_seq_points_found (MonoMethod *method, int offset)
 	printf ("Unable to find seq points for method '%s', offset 0x%x.\n", mono_method_full_name (method, TRUE), offset);
 }
 
+static int
+calc_il_offset (MonoDomain *domain, MonoMethod *method, int native_offset, gboolean is_top_frame)
+{
+	int ret = -1;
+	if (is_top_frame) {
+		SeqPoint sp;
+		/* mono_debug_il_offset_from_address () doesn't seem to be precise enough (#2092) */
+		if (mono_find_prev_seq_point_for_native_offset (domain, method, native_offset, NULL, &sp))
+			ret = sp.il_offset;
+	}
+	if (ret == -1)
+		ret = mono_debug_il_offset_from_address (method, domain, native_offset);
+	return ret;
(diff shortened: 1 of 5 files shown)


lambdageek:
@thaystg could you review this

lewurm:
@monojenkins build failed

mono/mini/debugger-agent.c
```diff
@@ -3331,6 +3339,21 @@ compute_frame_info (MonoInternalThread *thread, DebuggerTlsData *tls, gboolean f
...
+	if (CHECK_PROTOCOL_VERSION (2, 52)) {
+		MonoJitTlsData *jit_data = thread->thread_info->jit_data;
+		gboolean has_interp_resume_state = FALSE;
+		MonoInterpFrameHandle interp_resume_frame = NULL;
+		gpointer interp_resume_ip = 0;
+		mini_get_interp_callbacks ()->get_resume_state (jit_data, &has_interp_resume_state, &interp_resume_frame, &interp_resume_ip);
+		if (has_interp_resume_state && tls->frame_count > 0) {
+			StackFrame *top_frame = tls->frames [0];
+			if (interp_resume_frame == top_frame->interp_frame) {
+				int native_offset = (int) ((uintptr_t) interp_resume_ip - (uintptr_t) top_frame->de.ji->code_start);
```
lewurm: why does `top_frame->native_offset` not have the correct value here?

imhameed: `top_frame->de.native_offset` does not have the correct value here because it's populated by `interp_frame_iter_next`, which does not retrieve data from the interpreter's resume state (if present and valid). An alternative would instead entail altering `interp_frame_iter_next`, so that it does retrieve data from the interpreter's resume state, although then (if https://github.com/mono/mono/pull/14667#issuecomment-496553352 is important), I'd need a way to thread a "should read from the resume state" flag through.

lewurm:
@monojenkins squash

thaystg:
@monojenkins build failed

thaystg:

Issue #15912, opened 2019-07-30
csc flaky on arm


jaykrell:
## Steps to Reproduce

1.  Submit a PR.
Here a trivial one https://github.com/mono/mono/pull/15909

### Current Behavior

For example:

```
r:./../../class/lib/net_4_x-linux/System.Configuration.dll -r:./../../class/lib/net_4_x-linux/System.Web.dll -r:./../../class/lib/net_4_x-linux/System.Web.ApplicationServices.dll -r:./../../class/lib/net_4_x-linux/mscorlib.dll  -sourcelink:./../../build/common/sourcelink.json /keyfile:../ecma.pub  -target:library -out:../../class/lib/net_4_x-linux/System.IdentityModel.dll  @./../../build/deps/linux_net_4_x__System.IdentityModel.dll.response
CLR: Managed code called FailFast, saying "System.NullReferenceException: Object reference not set to an instance of an object
  at System.Collections.Generic.EnumerableHelpers.ToArray[T] (System.Collections.Generic.IEnumerable`1[T] source) [0x00000] in <bacbf551d6714e2ca6541c4da125153e>:0 
  at System.Linq.Enumerable.ToArray[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00018] in <bacbf551d6714e2ca6541c4da125153e>:0 
  at Microsoft.CodeAnalysis.CSharp.CodeGen.Optimizer.FilterValidStackLocals (System.Collections.Generic.Dictionary`2[TKey,TValue] info) [0x0000c] in <6ad4116dfb37422ca62221cd1c4a63d8>:0 
  at Microsoft.CodeAnalysis.CSharp.CodeGen.Optimizer.Optimize (Microsoft.CodeAnalysis.CSharp.BoundStatement src, System.Boolean debugFriendly, System.Collections.Generic.HashSet`1[Microsoft.CodeAnalysis.CSharp.Symbols.LocalSymbol]& stackLocals) [0x00015] in <6ad4116dfb37422ca62221cd1c4a63d8>:0 
[...]

PR #15943, opened 2019-07-31
Remove error_init from native icall wrappers since managed does it.

jaykrell:
Preferable would be native only, i.e. for interpreter and for when
it is actually unused, but ok for now.

Between removing the managed and the native, managed is preserved
in case there uses of MonoError in C# other than icall wrappers.
Keep them safe.

But again ideally move it to native and remove the managed type,
pending debugging WebAssembly failure doing so.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
https://github.com/mono/mono/issues/15912

PR #15944, opened 2019-07-31
Add Roslyn maestro dependency

agocke:
This should bring down the Microsoft.Net.Compilers package,
which contains the Roslyn binaries needed for mono packaging.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -9,6 +9,10 @@
       <Uri>https://github.com/dotnet/core-setup</Uri>
       <Sha>2977763fd35b8dc120f4eac94314aba7b05ae9ee</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.Net.Compilers" Version="3.3.0-beta2-19381-14">
+      <Uri>https://github.com/dotnet/roslyn</Uri>
+      <Sha>ef3a7a3863ae53b610a4b0c3682a35cad0829583</Sha>
+    </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
     <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="1.0.0-beta.19379.1">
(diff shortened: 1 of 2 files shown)


agocke:
@akoeplinger This should add the necessary dependency for Microsoft.Net.Compilers via Maestro. I've also added a subscription to darc, so you should get the same version as .NET Core in the future.

I took a look and the biggest problem seems to be that you rely on autoconf and automake to lay out the Roslyn binaries. That'll probably need to at least be altered to grab the appropriate version out of Version.Details.xml and call NuGet restore to bring down the right package.

eng/Version.Details.xml
```diff
@@ -9,6 +9,10 @@
       <Uri>https://github.com/dotnet/core-setup</Uri>
       <Sha>2977763fd35b8dc120f4eac94314aba7b05ae9ee</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.Net.Compilers" Version="3.3.0-beta2-19381-14">
+      <Uri>https://github.com/dotnet/roslyn</Uri>
+      <Sha>ef3a7a3863ae53b610a4b0c3682a35cad0829583</Sha>
+    </Dependency>
```
akoeplinger: I think you also need to add an entry into `eng/Versions.props`

agocke:
Added to Versions.props as well. The .NET Core 3 version is the right one, because these are binaries mono should ship, which should match what's being put in the CLI.

How did you want to handle consumption? Do you have a plan for how to hook MSBuild/NuGet into autoconf?

akoeplinger:
> How did you want to handle consumption? Do you have a plan for how to hook MSBuild/NuGet into autoconf?

Yeah my plan is to grep the version from Versions.props and then download from the nuget feed directly (e.g. 
https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/microsoft.net.compilers/3.3.0-beta2-19381-14/microsoft.net.compilers.3.3.0-beta2-19381-14.nupkg)

agocke:
sounds good

akoeplinger:
Merged as no build is affected by this yet and I want to see whether tonight's maestro/darc update includes this.

Issue #15195, opened 2019-06-16
[netcore] Make System.Runtime.Loader.Tests.DefaultLoadContextTests.LoadInDefaultContext Pass


steveisok:
The test currently fails with:

```
System.Runtime.Loader.Tests.DefaultLoadContextTests.LoadInDefaultContext [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.IO.FileNotFoundException)
      Actual:   (No exception was thrown)
      Stack Trace:
           at System.Runtime.Loader.Tests.DefaultLoadContextTests.LoadInDefaultContext()
           at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

Determine why and make the test pass.

CoffeeFlux:
Failure is unchanged as of now.

CoffeeFlux:
This now passes outside of the initial failure, which is the result of the assembly being loaded before the test is actually run. Unsure of why this happens, but it's the only remaining issue.

CoffeeFlux:
The initial failure appears to be our preload hook finding the test noop assembly via the domain search path when it shouldn't(?)

PR #15946, opened 2019-08-01
[netcore] Implement referenced assembly resolving and ALC.GetLoadContext

lambdageek:
1. Implement the ALC referenced assembly resolving logic from https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-3.0#usage-in-the-runtime
2. Implement `AssemblyLoadContext GetLoadContext (Assembly assembly)`

---

The managed code is from https://github.com/dotnet/coreclr/blob/8ba2e15201361402acd0ae9710bd37d50785cdfa/src/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyLoadContext.CoreCLR.cs#L253-L281

And from https://github.com/dotnet/coreclr/blob/8ba2e15201361402acd0ae9710bd37d50785cdfa/src/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyLoadContext.CoreCLR.cs#L84-L187

### DIFF
--- a/mono/metadata/assembly-load-context.c
+++ b/mono/metadata/assembly-load-context.c
@@ -47,4 +47,12 @@ ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalInitializeNativeALC 
 	return alc;
 }
 
+gpointer
+ves_icall_System_Runtime_Loader_AssemblyLoadContext_GetLoadContextForAssembly (MonoReflectionAssemblyHandle assm_obj, MonoError *error)
+{
+	MonoAssembly *assm = MONO_HANDLE_GETVAL (assm_obj, assembly);
+	MonoAssemblyLoadContext *alc = mono_assembly_get_alc (assm);
+
+	return GUINT_TO_POINTER (alc->gchandle);
+}
 #endif /* ENABLE_NETCORE */
--- a/mono/metadata/assembly-load-context.c
+++ b/mono/metadata/assembly-load-context.c
@@ -68,7 +68,7 @@ ves_icall_System_Runtime_Loader_AssemblyLoadContext_GetLoadContextForAssembly (M
 gboolean
 mono_alc_is_default (MonoAssemblyLoadContext *alc)
 {
-	return alc == alc->domain->default_alc;
+	return alc == mono_alc_domain (alc)->default_alc;
 }
 
 static MonoAssembly*
--- a/mono/metadata/assembly-load-context.c
+++ b/mono/metadata/assembly-load-context.c
@@ -1,12 +1,19 @@
 #include "config.h"
+#include "mono/metadata/assembly.h"
 #include "mono/metadata/domain-internals.h"
 #include "mono/metadata/icall-decl.h"
 #include "mono/metadata/loader-internals.h"
 #include "mono/metadata/loaded-images-internals.h"
 #include "mono/utils/mono-error-internals.h"
+#include "mono/utils/mono-logger-internals.h"
 
 #ifdef ENABLE_NETCORE
 /* MonoAssemblyLoadContext support only in netcore Mono */
+
+static
+GENERATE_GET_CLASS_WITH_CACHE_DECL (assembly_load_context);
+
+GENERATE_GET_CLASS_WITH_CACHE (assembly_load_context, "System.Runtime.Loader", "AssemblyLoadContext");
 
 void
 mono_alc_init (MonoAssemblyLoadContext *alc, MonoDomain *domain)
@@ -40,6 +47,8 @@ ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalInitializeNativeALC 
 	if (is_default_alc) {
 		alc = mono_domain_default_alc (domain);
 		g_assert (alc);
+		if (!alc->gchandle)
+			alc->gchandle = this_gchandle;
 	} else {
 		/* create it */
 		alc = mono_domain_create_individual_alc (domain, this_gchandle, collectible, error);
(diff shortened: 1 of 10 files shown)


lambdageek:
Draft until I see that I didn't break anything.

I don't think I fixed any of the disabled tests (I tried `System.Runtime.Loader.Tests.DefaultLoadContextTests.LoadInDefaultContext` https://github.com/mono/mono/issues/15195 but it's still failing).

mono/metadata/assembly.c
```diff
@@ -1679,9 +1738,16 @@ mono_assembly_load_reference (MonoImage *image, int index)
...
 		}
+#else
+		reference = netcore_load_reference (&aname, mono_image_get_alc (image), image->assembly, &status);
+#endif
 	} else {
+#ifndef ENABLE_NETCORE
 		/* FIXME: can we establish that image->assembly is never NULL and this code is dead? */
 		reference = load_reference_by_aname_default_asmctx (&aname, mono_image_get_alc (image), image->assembly, &status);
+#else
+		g_assertf (image->assembly, "While loading reference %d MonoImage %s doesn't have a MonoAssemly", index, image->name);
```
jaykrell: `MonoAssemly`

mono/metadata/assembly-load-context.c
```diff
@@ -1,13 +1,20 @@
 #include "config.h"
+#include "mono/metadata/assembly.h"
 #include "mono/metadata/domain-internals.h"
 #include "mono/metadata/icall-decl.h"
 #include "mono/metadata/loader-internals.h"
 #include "mono/metadata/loaded-images-internals.h"
 #include "mono/utils/mono-error-internals.h"
+#include "mono/utils/mono-logger-internals.h"
```
jaykrell: `#include` possibly under `#if` but there are reasons why not: sorting or don't-care or order-dependent.

mono/metadata/assembly-load-context.c
```diff
@@ -40,11 +47,120 @@ ves_icall_System_Runtime_Loader_AssemblyLoadContext_InternalInitializeNativeALC
...
+
+MonoAssembly*
+mono_alc_invoke_resolve_using_resolving_event_nofail (MonoAssemblyLoadContext *alc, MonoAssemblyName *aname)
+{
+	MonoAssembly *result = NULL;
+	ERROR_DECL (error);
+
+	result = mono_alc_invoke_resolve_using_resolving_event (alc, aname, error);
+	if (!is_ok (error))
+		mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_ASSEMBLY, "Error while invoking ALC Load(\"%s\") method: '%s'", aname->name, mono_error_get_message (error));
```
lambdageek:
```suggestion
		mono_trace (G_LOG_LEVEL_INFO, MONO_TRACE_ASSEMBLY, "Error while invoking ALC Resolving(\"%s\") event: '%s'", aname->name, mono_error_get_message (error));
```

mono/metadata/assembly.c
lambdageek:
```suggestion
		g_assertf (image->assembly, "While loading reference %d MonoImage %s doesn't have a MonoAssembly", index, image->name);
```

filipnavara:
I took a first peek and it and it looked ok. As far as test suite goes I am surprised I didn't find any coverage for this. These tests succeed on the branch now and they were disabled before:

- System.Runtime.Loader.Tests.AssemblyLoadContextTest.PublicConstructor_Default
 System.Runtime.Loader.Tests.AssemblyLoadContextTest.GetAssemblyNameTest_NullParameter
System.Runtime.Loader.Tests.AssemblyLoadContextTest.LoadFromAssemblyName_AssemblyNotFound
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.PublicConstructor_Theory (with isCollectible: False)
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.GetAssemblyNameTest_ValidAssembly
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.GetLoadContextTest_SystemPrivateCorelibAssembly
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.DefaultAssemblyLoadContext_Properties
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.GetAssemblyNameTest_AssemblyNotFound 
- System.Runtime.Loader.Tests.AssemblyLoadContextTest.GetLoadContextTest_ValidTrustedPlatformAssembly
- System.Tests.AppDomainTests.GetAssemblies

Not all of them were fixed by this PR though (eg. the last one was fixed earlier by @CoffeeFlux).

CoffeeFlux:
Will review more properly in a moment, but is there a reason to copy-paste methods from CoreCLR instead of having them moved to the shared partition? Most of them can be brought over untouched.

lambdageek:
>  is there a reason to copy-paste methods from CoreCLR instead of having them moved to the shared partition? Most of them can be brought over untouched.

Mostly because I wasn't sure how to do that.

Also while the `Resolve` methods can be moved verbatim, the `GetLoadContext` code needs a bit of massaging if we want to share it - the CoreCLR version uses their QCall helpers which don't really make sense for mono.

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
```diff
@@ -87,5 +114,129 @@ internal static Assembly DoAssemblyResolve (string name)
...
+		{
+			AssemblyLoadContext context = (AssemblyLoadContext)(GCHandle.FromIntPtr(gchManagedAssemblyLoadContext).Target)!;
+
+			// Invoke the AssemblyResolve event callbacks if wired up
+			return context.ResolveUsingEvent(assemblyName);
+		}
+
+		// This method is invoked by the VM to resolve a satellite assembly reference
+		// after trying assembly resolution via Load override without success.
+		private static Assembly? ResolveSatelliteAssembly(IntPtr gchManagedAssemblyLoadContext, AssemblyName assemblyName)
```
CoffeeFlux: This is being copied but is unused, though looking at CoreCLR it seems it should be? See https://github.com/dotnet/coreclr/blob/007c6bc5d749c6cbd5fd106a2053ef5563d12c1d/src/vm/appdomain.cpp#L6599-L6624 and https://github.com/dotnet/coreclr/blob/007c6bc5d749c6cbd5fd106a2053ef5563d12c1d/src/binder/clrprivbinderassemblyloadcontext.cpp#L61-L68

CoffeeFlux: Whoops, didn't grab all the code I meant to, updated links. Maybe chat at lunch about the satellite assembly stuff.

CoffeeFlux:
Also regarding the disabled tests, I think a bunch of them should start to pass between this, #15850, and my upcoming PR to make the load hook ALC-aware. After all of those have landed, it's probably worth assessing the tests and seeing what can be re-enabled with confidence that our behavior is reasonably correct.

filipnavara:
@CoffeeFlux there is an open PR by @steveisok to reasses all the test exclusions. The plan is to do that more regularly which is why I am not submitting a separate PR now for every case I found.

CoffeeFlux:
Makes sense - I'll keep an eye on his PR then, I just don't want to enable more tests until we're sure they're passing for the right reasons, as until all of those PRs land our behavior is definitely wrong in a lot of common scenarios.

filipnavara:
> re: moving things to the shared partition still apply.
>
> I asked in Slack on Tuesday (since I was looking at how to implement this myself) and was told the best approach is to just open an issue in the CoreCLR repo saying what you want moved over.

Feel free to tag me on any such issue. I have been through the process before and I can help with it.

You start by submitting PR for CoreCLR to move the necessary parts from `src/*` to `shared/*`. The change has to be tested on CoreCLR, Mono and CoreRT to ensure it does not break builds. Sharing between two runtimes only is also possible using exclusions in the .csproj files. It is not advised though. Once the CoreCLR PR is accepted it gets mirrored to Mono and CoreRT. Any necessary changes have to be applies on top of this mirror PR which is merged then.

I am not necessarily convinced that in this particular case the sharing would be beneficial but it may be too early to evaluate it.

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
lambdageek: Thanks for tracking this down. I admit I punted on satellite assembly stuff. I'll take a look at wiring it up

mono/metadata/assembly.c
```diff
@@ -1578,6 +1579,63 @@ load_reference_by_aname_individual_asmctx (MonoAssemblyName *aname, MonoAssembly
...
+	reference = mono_assembly_loaded_internal (alc, aname, FALSE);
+	if (reference)
+		goto leave;
+
+	if (!mono_alc_is_default (alc))
+		reference = mono_alc_invoke_resolve_using_load_nofail (alc, aname);
+	if (reference)
+		goto leave;
+
+	MonoDomain *domain = mono_domain_get ();
```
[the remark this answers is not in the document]
CoffeeFlux: To add on to this, there are a bunch of places in the runtime where we currently get the domain with `mono_domain_get()` despite being passed an ALC. Perhaps worth making those get it from the ALC instead at some point.

lambdageek:
Added some logic to do satellite assembly loading and enabled some of the tests in System.Runtime.Loader.AssemblyLoadContextTests

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
```diff
@@ -129,6 +129,14 @@ internal static Assembly DoAssemblyResolve (string name)
 			return ResolveUsingResolvingEvent (gchALC, new AssemblyName (assemblyName));
 		}
 
+		// Invoked by Mono to resolve using the Resolving event after
+		// trying the Load overried and TPA load context without
```
CoffeeFlux: Typo, but also - are we going to use the CoreCLR terminology here, notably "TPA load context"? I'd prefer to just say the default since that's what we've used in the runtime.

Issue #15843, opened 2019-07-26
[wasm][sdk] Support for Custom Linker Configuration


kjpou1:
The sdk should support a way to add extra definitions to the linker to ensure the type, methods and/or fields are not eliminated.

Should follow the same pattern as IOS and Android [Custom Linker Configuration](https://docs.microsoft.com/en-us/xamarin/cross-platform/deploy-test/linker)

PR #15953, opened 2019-08-01
[wasm][sdk] Support for Custom Linker Configuration

kjpou1:
- Provide extra definitions to the linker to ensure the type, methods and/or fields are not eliminated from your application.

* Use a <LinkDescription> item as follows:

```
  <ItemGroup>
    <LinkDescription Include="customlinker.xml" />
  </ItemGroup>
```

These will then be added to the `WasmLinkAssemblies` task step.

Example xml:

```

<linker>
        <assembly fullname="mscorlib">
                <type fullname="System.Environment">
                        <field name="mono_corlib_version" />
                        <method name="get_StackTrace" />
                </type>
        </assembly>
        <assembly fullname="My.Own.Assembly">
                <type fullname="Foo" preserve="fields">
                        <method name=".ctor" />
                </type>
                <type fullname="Bar">
                        <method signature="System.Void .ctor(System.String)" />
                        <field signature="System.String _blah" />
                </type>
                <namespace fullname="My.Own.Namespace" />
                <type fullname="My.Other*" />
        </assembly>
</linker>

```

Close https://github.com/mono/mono/issues/15843

### DIFF
--- a/sdks/wasm/sdk/Mono.WebAssembly.Build/Mono.WebAssembly.Build.targets
+++ b/sdks/wasm/sdk/Mono.WebAssembly.Build/Mono.WebAssembly.Build.targets
@@ -95,6 +95,7 @@
 			LinkSkip="$(MonoWasmLinkSkip)"
 			Debug="$(_DebugSymbolsProduced)"
 			I18n="$(MonoWasmI18n)"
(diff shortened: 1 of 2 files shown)


kjpou1:
@monojenkins build failed

PR #15957, opened 2019-08-01
[Coop] Convert first parameter to mono_remoting_invoke.

jaykrell:
Baby steps.

### DIFF
--- a/mono/metadata/marshal.c
+++ b/mono/metadata/marshal.c
@@ -1175,13 +1175,17 @@ mono_delegate_begin_invoke (MonoDelegate *delegate, gpointer *params)
 	MonoMulticastDelegate *mcast_delegate;
 	MonoClass *klass;
 	MonoMethod *method;
+	MonoAsyncResult *result = NULL;
+
+	HANDLE_FUNCTION_ENTER ();
 
 	g_assert (delegate);
+
 	mcast_delegate = (MonoMulticastDelegate *) delegate;
 	if (mcast_delegate->delegates != NULL) {
 		mono_error_set_argument (error, NULL, "The delegate must have only one target");
 		mono_error_set_pending_exception (error);
-		return NULL;
+		goto return_null;
 	}
 
 #ifndef DISABLE_REMOTING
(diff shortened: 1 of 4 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

vargaz:
Still think we should make handles pin their objects, so you can pass native objrefs into functions as long as you keep a handle pointing to it. It would massively simplify these changes.

jaykrell:
I don’t disagree.
It could reduce huge change by way over half, maybe 90%, and resolve several things like all the residual ‘MONO_HANDLE_RAW’.

1. Static analysis might suffer but maybe ok.

2. This would enable a nice approach in C++ and further ease the work.
3. I suggested this too.
4. Let’s discuss it?
5. I think it is a tiny PR, I could float.

@lambdageek?

jaykrell:
I think it has far reaching *mostly* positive implications.
For example if you have a handle to a struct, then I think all of its fields would be safe to use raw.
You could stop creating handles for them.
You would declare them all as volatile, well, the pointers.
Not because they could move, but to ensure the in-memory values are up to date, which handles do.

And the handle construction itself could be pretty easy C++. I had a prototype.

Furthermore, if we ever more to, C++14 or 17, not sure, the order of evaluation restrictions mean the pinning isn't even any longer required.

But even just in C I believe it'd cut down the change dramatically.
I do think human and computer static analysis would need attention, to track that things are safe by other than type.
Possibly just a typedef that equates to the same raw pointers, that only helps human and is not checked by computer, maybe dangerous.

I did look at the GC code but not super obvious how to do this, but surely not difficult.

jaykrell:

jaykrell:
@monojenkins build failed

lambdageek:
I think I'd prefer not to merge this:
1. We agreed to proceed with #15970 (coop handles will pin managed objects).  Therefore the onus is on the caller of `mono_remoting_invoke` to pin the "this" object.
2. This doesn't address the `args` array which is still full of raw pointers.  (And with #15970 we won't have to).

So in the long term, using a handle for "this" but not the contents of "args" is confusing and unnecessary.  In the short term we can live with this being an unconverted function.

jaykrell:
I'd just as soon do a little more old fashioned conversion before adjusting to the new ways.

jaykrell:
And I wasn't meaning to stop at the first parameter, but the changes were getting large, so it was just a point to pause and get to master with a non-huge PR.

jaykrell:
This was in fact extracted from a much larger PR, that again, I'd just as soon continue to whittle away at.
I think I started by looking at all `mono_string_new_checked` or such and followed that graph. I was going to eliminate all of its uses. Allocators do need attention even in the new pinned world I think.

PR #15965, opened 2019-08-01
[netcore] Make System.Private.CoreLib.csproj IDE friendly

EgorBo:
Now it's possible to open `System.Private.CoreLib.csproj` in an IDE and build (use Roslyn version from `eng/Versions.props`).


directhex:
Can't repro the AzDO failures.

PR #15972, opened 2019-08-01
Use `MultipleConnectAsync` from CoreFX.

baulig:


baulig:
@monojenkins commit csproj

baulig:
@monojenkins build failed

steveisok:

steveisok:
@baulig do you think there would be any value in running this on the designer bot?  I'm wondering if some kind of periodic stress test would be worth it?

baulig:
> @baulig do you think there would be any value in running this on the designer bot? I'm wondering if some kind of periodic stress test would be worth it?

I actually thought about periodically running some stress tests myself, but we probably shouldn't use `nuget restore` for that and instead come up with a better solution.  Let's also have a look at CoreFX to see whether they might have any kind of stress tests for the socket code.

As for this particular PR, all I did was committing the work I did earlier on comparing the two implementations, to make it more obvious that they are.

steveisok:
Yeah, I don't think nuget restore is ideal either.  Stress tests in general would be a good idea, so let's look at corefx and see 👍 .

PR #15974, opened 2019-08-01
Fix patch step in mdoc tests for FreeBSD and others

rootwyrm:
Non-Linux systems largely do not support `--binary`. Since the patch appears to be text anyway, it's probably safe to remove for all platforms.

With `patch -p0` , mdoc test now provides accurate pass/fail on FreeBSD.


directhex:
@monojenkins build Linux x64

directhex:
@monojenkins build OS X x64

PR #15977, opened 2019-08-01
Fix usage of mono_debug_get_seq_points to avoid leaks

jbevain:
Audit usage of `mono_debug_get_seq_points` to either avoid passing unused variables that are not freed or properly free usage of returned variables.


jbevain:
Originally reported and fixed downstream in https://github.com/Unity-Technologies/mono/pull/1213

lewing:
@monojenkins build failed

PR #15978, opened 2019-08-01
[ci] Skip jenkins lanes for PRs which only contain changes to the netcore/ subdir.

vargaz:


vargaz:
Was tested on this lane:
https://github.com/mono/mono/pull/15978

vargaz:
@monojenkins build failed

filipnavara:
Can we do the same for the `eng` directory (ie. Maestro version updates)?

akoeplinger:
@filipnavara we're just about to start consuming Roslyn via arcade (https://github.com/mono/mono/pull/15944) so I don't think we can do that...

PR #15986, opened 2019-08-02
Relax locking

CoffeeFlux:
Ran into some suspicious locks and attempted to relax them a bit. Not totally sure I got the lock-free code right - @jaykrell can you take a look?

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3035,11 +3035,8 @@ unload_thread_main (void *arg)
 	}
 
 	/* Clear references to our vtables in class->runtime_info.
-	 * We also hold the loader lock because we're going to change
-	 * class->runtime_info.
 	 */
 
-	mono_loader_lock (); //FIXME why do we need the loader lock here?
 	mono_domain_lock (domain);
 	/*
 	 * We need to make sure that we don't have any remsets
@@ -3060,7 +3057,6 @@ unload_thread_main (void *arg)
 	mono_assembly_cleanup_domain_bindings (domain->domain_id);
 
 	mono_domain_unlock (domain);
-	mono_loader_unlock ();
 
 	domain->state = MONO_APPDOMAIN_UNLOADED;
 
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3035,7 +3035,7 @@ unload_thread_main (void *arg)
 	}
 
 	/* Clear references to our vtables in class->runtime_info.
-	 * We also hold the loader lock because we're going to change		
+	 * We also hold the loader lock because we're going to change
  	 * class->runtime_info.
 	 */
 
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -3035,8 +3035,11 @@ unload_thread_main (void *arg)
 	}
 
 	/* Clear references to our vtables in class->runtime_info.
+	 * We also hold the loader lock because we're going to change		
+ 	 * class->runtime_info.
 	 */
 
+	mono_loader_lock ();
 	mono_domain_lock (domain);
 	/*
 	 * We need to make sure that we don't have any remsets
@@ -3057,6 +3060,7 @@ unload_thread_main (void *arg)
 	mono_assembly_cleanup_domain_bindings (domain->domain_id);
 
 	mono_domain_unlock (domain);
+	mono_loader_unlock ();
 
 	domain->state = MONO_APPDOMAIN_UNLOADED;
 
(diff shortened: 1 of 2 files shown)


mono/metadata/appdomain.c
```diff
@@ -3035,11 +3035,8 @@ unload_thread_main (void *arg)
 	}
 
 	/* Clear references to our vtables in class->runtime_info.
-	 * We also hold the loader lock because we're going to change
-	 * class->runtime_info.
 	 */
 
-	mono_loader_lock (); //FIXME why do we need the loader lock here?
 	mono_domain_lock (domain);
```
vargaz: loader lock is the global runtime lock, while domain lock is a lower level lock. So its usually a good idea to take the loader lock before calling into complicated code like the one below, since it might take the loader lock itself.

mono/mini/debugger-agent.c
```diff
@@ -2228,22 +2228,34 @@ static AgentDomainInfo*
...
-	if (!info) {
-		info = g_new0 (AgentDomainInfo, 1);
-		domain_jit_info (domain)->agent_info = info;
-		info->loaded_classes = g_hash_table_new (mono_aligned_addr_hash, NULL);
-		info->source_files = g_hash_table_new (mono_aligned_addr_hash, NULL);
-		info->source_file_to_class = g_hash_table_new (g_str_hash, g_str_equal);
-		info->source_file_to_class_ignorecase = g_hash_table_new (g_str_hash, g_str_equal);
+
+	if (info) {
+		mono_memory_read_barrier ();
```
CoffeeFlux: Is this needed even if I'm not dereferencing the pointer?

mono/metadata/appdomain.c
CoffeeFlux: Oh yeah, this commit is bad, I didn't look closely enough. I'll just remove the FIXME I guess.

mono/mini/debugger-agent.c
lambdageek: Yes it's needed. Because the caller will access the memory that `info` points to.

CoffeeFlux: That's what I figured, but I didn't know if the fact you're coming out of a call means anything in terms of concurrency guarantees.

mono/mini/debugger-agent.c
```diff
@@ -2228,22 +2228,34 @@ static AgentDomainInfo*
...
+	info->source_files = g_hash_table_new (mono_aligned_addr_hash, NULL);
+	info->source_file_to_class = g_hash_table_new (g_str_hash, g_str_equal);
+	info->source_file_to_class_ignorecase = g_hash_table_new (g_str_hash, g_str_equal);
+
+	mono_memory_write_barrier ();
+
+	gpointer other_info = mono_atomic_cas_ptr (&jit_info->agent_info, info, NULL);
+
+	if (other_info != NULL) {
+		g_hash_table_destroy (info->loaded_classes);
```
jaykrell: I think this code should be shared with elsewhere, like `debugger_agent_free_domain_info`.

jaykrell:
Overall, it feels a *little* dodgy, because
all of the readers take a lock.

But I think it is ok.

A more conservative form initializes the empty
thing outside lock, and the replaces the cas
with lock + cas.

But I think it is ok.

All of the lockers can only see a fully "initialized" thing.

And all the initializing writes to the thing, are before
the pointer to the thing is published.


I also don't think any of these lines:
```
		info->loaded_classes = g_hash_table_new (mono_aligned_addr_hash, NULL);
		info->source_files = g_hash_table_new (mono_aligned_addr_hash, NULL);
		info->source_file_to_class = g_hash_table_new (g_str_hash, g_str_equal);
		info->source_file_to_class_ignorecase = g_hash_table_new (g_str_hash, g_str_equal);
```

are really needed here.

A null pointer can/should be equated with an empty hash table,
and the `new` be done on demand at the first insert.

But it is also ok.

And that would make the locking more convincing.

Furthermore, Mono over uses pointers.
agent_info does not need to be a pointer, I would think,
unless it is either larger and tends to be unused,
or its lifetime exceeds its apparent owner, such as by independent
ref-counting. Or if you really want to hide its size from its container.
It does appear hidden here, but that seems to be using indirection
and inefficiency to enforce abstraction boundary.


In C++ you should tend away from this pointer.
Unless you were really optimizing compile time and wanted the "pimpl" pattern,
which is certainly a thing in the literature, and I do not see used much,
except for separately compiled projects.


So, taken to extreme, this function goes away and is replaced by just the new0 of its container.
And the existing locks for the next level down mutation.

I suggest that be done next.

And consider going one level further don't "new" the hashtables either.

Unless the data really tends to be large and uninstantiated.
Then use a pointer.

Don't systematically use pointers for abstraction boundary enforcement. Imho.

mono/mini/debugger-agent.c
CoffeeFlux: The code in `debugger_agent_free_domain_info` is significantly more complicated and overkill for this imo

PR #15988, opened 2019-08-02
Don't use clock_nanosleep on PASE

NattyNarwhal:
PASE exports this AIX syscall so autoconf detects it, but using it
will trigger SIGILL (for unimplemented syscall), which emits a LIC
log entry type of 4700-000F.


akoeplinger:
FYI it's also used in mono/mini/mini-posix.c

PR #15997, opened 2019-08-02
[llvm] Add support for JIT + LLVM 9.0.

imhameed:


imhameed:
Bummer: On my workstation, I'm getting intermittent crashes with LLVM 9 + JIT that I didn't see last night. I'd like to figure those out before this is merged.

PR #15998, opened 2019-08-02
[netcore] Fix comments in ALC to use Mono terminology

lambdageek:
As suggested [in a comment on 15946](https://github.com/mono/mono/pull/15946#discussion_r310227893), fix inadvertantly copy-pasted comment and make sure to use Mono terminology.


netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
```diff
@@ -122,16 +122,14 @@ internal static Assembly DoAssemblyResolve (string name)
 		}
 
 		// Invoked by Mono to resolve using the Resolving event after
-		// trying the Load overried and TPA load context without
+		// trying the Load overried and default load context without
```
CoffeeFlux: Still missed "overried"

PR #16001, opened 2019-08-03
`SocketAsyncEventArgs` cleanups and preparations for the real thing.

baulig:
Cleaning up the `SocketAsyncEventArgs` class and preparing it for it's eventual
adaption from CoreFX.

There are a few instance fields and property accessors with `internal` visibility,
which the CoreFX version does not have.  Replaced these with accessor methods.

The `PolicyRestricted` property and it's related `SocketAsyncEventArgs (bool)`
constructor is not used anywhere and the CoreFX version has a constructor with the
same signature, but different semantics.

In the CoreFX version, `Complete ()` does not call `DoComplete (this)` - renamed it
into `Complete_internal ()` to avoid any ambiguities and help the transitioning.

### DIFF
--- a/mcs/class/System/System.Net.Sockets/Socket.cs
+++ b/mcs/class/System/System.Net.Sockets/Socket.cs
@@ -608,15 +608,15 @@ namespace System.Net.Sockets
 				throw new InvalidOperationException ("No operation in progress");
 
 			try {
-				e.AcceptSocket = e.current_socket.EndAccept (ares);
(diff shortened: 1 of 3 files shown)


baulig:
@monojenkins build failed

baulig:
@monojenkins build failed

akoeplinger:
The Android SDK failure is because of https://github.com/mono/mono/pull/16048

PR #16003, opened 2019-08-03
[windows gc] Trim one syscall per thread in garbage collection.

jaykrell:
Specifically do not query for guard pages.
There are two better ways:
 1. Visual C++ specific, __try / __except, get the guard page exception,
    put back the guard page. This is a little slower (exception + two syscalls),
    but exceedinly rare. __try itself is super cheap.
    This is also not allowed on UWP.
 2. Record the TEB and get from it the maximum commited page and just use MAX, trivial.

The comments are also not completely clear as to what is going on here, but they were close.
Adjust them.


jaykrell:
We could use a stress test for this, old and new. It seems like a very difficult condition to hit, but real.

jaykrell:
@monojenkins build failed

mono/metadata/sgen-mono.c
```diff
@@ -2217,14 +2217,127 @@ sgen_client_scan_thread_data (void *start_nursery, void *end_nursery, gboolean p
...
+#if 0
 		MEMORY_BASIC_INFORMATION mem_info;
-		SIZE_T result = VirtualQuery(info->client_info.stack_start, &mem_info, sizeof(mem_info));
+		const SIZE_T result = VirtualQuery(info->client_info.stack_start, &mem_info, sizeof(mem_info));
 		g_assert (result != 0);
 		if (mem_info.Protect & PAGE_GUARD) {
 			aligned_stack_start = ((char*) mem_info.BaseAddress) + mem_info.RegionSize;
 		}
+#else // old code vs. new code
+#if 0 // _MSC_VER
```
CoffeeFlux: Did you mean to uncomment this?

mono/metadata/sgen-mono.c
```diff
@@ -2217,14 +2217,127 @@ sgen_client_scan_thread_data (void *start_nursery, void *end_nursery, gboolean p
...
+		__try {
+			*(volatile char*)aligned_stack_start;
+		} __except (GetExceptionCode () == STATUS_GUARD_PAGE_VIOLATION) {
+			MEMORY_BASIC_INFORMATION mem_info;
+			const SIZE_T result = VirtualQuery(aligned_stack_start, &mem_info, sizeof(mem_info));
+			g_assert (result >= sizeof (mem_info));
+			VirtualProtect (aligned_stack_start, 1, mem_info.Protect | PAGE_GUARD, &mem_info.Protect);
+		}
+#else // clang / gcc
+		aligned_stack_start = MAX (aligned_stack_start, ((PNT_TIB)info->client_info.info.windows_teb)->StackLimit);
```
CoffeeFlux: I thought Windows tries to discourage people from accessing members in a TEB directly? Does that not apply here because you just care about the TIB? I'm far from an expert on Windows, so I'd be curious what the specifics are for that.

mono/metadata/sgen-mono.c
jaykrell: Not sure compiler-specificity is worthwhile. The compiler-portable code is good.

mono/metadata/sgen-mono.c
jaykrell: I pretty sure TIB is public, TEB is private. Winnt.h vs. winternl.h. I can try to check.

jaykrell: I pretty sure TIB is public, TEB is private. Winnt.h vs. winternl.h. I can try to check.

jaykrell: I pretty sure TIB is public, TEB is private. Winnt.h vs. winternl.h. I can try to check.

jaykrell: I confirmed it is ok.

mono/utils/mono-threads.h
```diff
@@ -200,9 +200,13 @@ typedef struct _MonoThreadInfo {
 	/*Tells if this thread was created by the runtime or not.*/
 	gboolean runtime_thread;
 
-	/* Max stack bounds, all valid addresses must be between [stack_start_limit, stack_end[ */
+	/* Max stack bounds, all valid addresses must be between [stack_start_limit, stack_end) */
 	void *stack_start_limit, *stack_end;
 
+#if HOST_WIN32
+	struct _TEB* windows_teb;
```
lateralusX: If TEB is private and TIB is public, shouldn't we store TIB here instead, since we are casting it to a TIB when used?

lateralusX: Maybe put it together with the other Windows specific members under USE_WINDOWS_BACKEND?

mono/utils/mono-threads-windows.c
```diff
@@ -470,6 +470,8 @@ mono_threads_platform_get_stack_bounds (guint8 **staddr, size_t *stsize)
 	//
 	VirtualQuery (&info, &info, sizeof (info));
 	*staddr = (guint8*)info.AllocationBase;
+	// TEB starts with TIB.
```
lateralusX: Combine comment into one line?

mono/metadata/sgen-mono.c
```diff
@@ -2217,14 +2217,127 @@ sgen_client_scan_thread_data (void *start_nursery, void *end_nursery, gboolean p
 		   will raise a STATUS_GUARD_PAGE_VIOLATION and the process will crash. This code uses
 		   VirtualQuery() to determine whether stack_start points into the guard page and then
 		   updates aligned_stack_start to point at the next non-guard page. */
+		//
```
lateralusX: When we pick solution, rewrite and reduce comment to correctly match selected solution.

mono/metadata/sgen-mono.c
```diff
@@ -2217,14 +2217,127 @@ sgen_client_scan_thread_data (void *start_nursery, void *end_nursery, gboolean p
...
+		__try {
+			*(volatile char*)aligned_stack_start;
+		} __except (GetExceptionCode () == STATUS_GUARD_PAGE_VIOLATION) {
+			MEMORY_BASIC_INFORMATION mem_info;
+			const SIZE_T result = VirtualQuery(aligned_stack_start, &mem_info, sizeof(mem_info));
+			g_assert (result >= sizeof (mem_info));
+			VirtualProtect (aligned_stack_start, 1, mem_info.Protect | PAGE_GUARD, &mem_info.Protect);
+		}
+#else // clang / gcc
+		aligned_stack_start = MAX (aligned_stack_start, ((PNT_TIB)info->client_info.info.windows_teb)->StackLimit);
```
lateralusX: Shouldn't we be able to just pick this solution a cross the board? If so we should probably store the TIB directly in client_info.info since that is part of the public API surface.

jaykrell: Minor. I’m used to calling it TEB and that name is widely known, and it is read with NtCurrentTeb(). But I guess..

mono/utils/mono-threads.h
jaykrell: I don’t see the point of that define. Is It because there are pthreads libraries for Windows?? That would make it wrong. But I don’t know what it means.

mono/metadata/sgen-mono.c
jaykrell: Yeah, it was a long comment, and almost correct. It was easier to amend than to edit. I can try.

mono/utils/mono-threads.h
jaykrell: Oh you mean struct arrangement. Sure.

jaykrell: I thought you meant different ifdef.

lateralusX: It includes things that are windows specific and is set when HOST_WIN32 so instead of having two windows defines in this struct, I was thinking that we could just add it bellow the existing one USE_WINDOWS_BACKEND that is set when we HOST_WIN32 is set.

jaykrell: Yes I think I understand and will do (didn't look at the code yet).

mono/metadata/sgen-mono.c
jaykrell: The previous long comment was duplicated twice.
New comment is longer. Put in one place and reference from other?
For now I put it in guard-page.txt.

mono/metadata/guard-page.txt
```diff
@@ -0,0 +1,100 @@
+// Due to the guard page mechanism providing gradual commit of Windows stacks,
```
jaykrell: I guess could put one or two two small functions here to justify its existance?
i.e. the line that calls max?

lateralusX: You mean under the 4:th way of doing it and the way we use here? Yeah that would make sense.

mono/utils/mono-threads.h
```diff
@@ -200,7 +200,7 @@ typedef struct _MonoThreadInfo {
 	/*Tells if this thread was created by the runtime or not.*/
 	gboolean runtime_thread;
 
-	/* Max stack bounds, all valid addresses must be between [stack_start_limit, stack_end[ */
+	/* Max stack bounds, all valid addresses must be between [stack_start_limit, stack_end) */
```
lateralusX: If we are going to touch this line, then maybe we should use either ( ) or [ ], right now we ended up with [ )

mono/metadata/guard-page.txt
jaykrell: I mean, this single line:

```
aligned_stack_start = MAX (aligned_stack_start, info->client_info.info.windows_tib->StackLimit);
```


could be a function:

static gsize
mono_thread_aligned_stack_start (info)
{
 void *aligned_stack_start aligned_stack_start = (void*)(mword) ALIGN_TO ((mword)info->client_info.stack_start, SIZEOF_VOID_P);
#if _WIN32
 // Huge comment, single-intsanced
 aligned_stack_start  = (aligned_stack_start, info->client_info.info.windows_tib->StackLimit);
#endif
 return aligned_stack_start ;
}

mono/utils/mono-threads.h
jaykrell: `[)` is the correct mathematical notation, for inclusive at the start, exclusive at the end.

jaykrell: My edit is specifically fixing the comment.

lateralusX: Ah, OK, missed that part.

mono/metadata/guard-page.txt
lateralusX: Not a big fan of having this huge comment with code samples included in source code, then I guess the way you have it now is OK (in separate file) or reduce the comment to smaller size (without source code example) and do as stated above.

jaykrell: The comment was already large, and matter is very esoteric. I agree the comment is then really really large. I could blog it perhaps.. I'm about to put the comment in the code. But ok either way.

jaykrell: i.e. very esoteric merits larger comment imho, but maybe I exceeded the limit.

mono/metadata/sgen-mono.c
```diff
@@ -1643,6 +1643,116 @@ report_handle_stack_roots (GCRootReport *report, SgenThreadInfo *info, gboolean
 	mono_handle_stack_scan (info->client_info.info.handle_stack, report_handle_stack_root, &ud, ud.precise, FALSE);
 }
 
+static void*
+get_aligned_stack_start (SgenThreadInfo *info)
+{
+	void* aligned_stack_start = (void*)(mword) ALIGN_TO ((mword)info->client_info.stack_start, SIZEOF_VOID_P);
```
jaykrell: This alignment is debatable.
1. On amd64 it will already be pointer aligned.
2. Are there systems where stack might not be pointer-aligned?
3. It is not correct for grow-up stack (HPPA).

PR #16004, opened 2019-08-03
 assembly-load-context.obj : warning LNK4221: This object file does n…

jaykrell:
…ot define any previously undefined public symbols, so it will not be used by any link operation that consumes this library

### DIFF
--- a/mono/metadata/assembly-load-context.c
+++ b/mono/metadata/assembly-load-context.c
@@ -1,4 +1,8 @@
 #include "config.h"
+#include "mono/utils/mono-compiler.h"
+
+#ifdef ENABLE_NETCORE // MonoAssemblyLoadContext support only in netcore Mono
+
 #include "mono/metadata/assembly.h"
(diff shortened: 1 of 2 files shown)


CoffeeFlux:
The change itself looks good, just needs to be rebased and working on non-netcore.

jaykrell:
How would it not work on non-netcore?

jaykrell:
Huh, breaks all over, will rebase agreed & look further.

mono/metadata/assembly-load-context.c
```diff
@@ -1,4 +1,8 @@
 #include "config.h"
+#include "mono/utils/mono-publib.h"
```
lambdageek:
```suggestion
#include "mono/utils/mono-compiler.h"
#include "mono/utils/mono-publib.h"
```

jaykrell: Agreed of course.

Issue #15992, opened 2019-08-02
Mono crashes during precompiling dll in fullaot mode


kbaladurin:
## Steps to Reproduce

1. Build mono (40f28b2e6371c708989ff59dd1cb4f7dcc22aedc)
```
$ cd <mono_repo>/netcore
$ ./build.sh
```
2. Build application:
```
$ cd <app_project_dir>
$ <mono_repo>/.dotnet/dotnet build
```
3. AOT image
```
$ export MONO_PATH=<mono_overlay>
$ <mono_repo>/mono/mini/mono -v --runtime=mobile -O=gsharedvt --aot=full <app_project_dir>/bin/Debug/netcoreapp3.0/helloworld.dll
```
Program.cs:
```cs
using System;
using System.Reflection;
using System.Runtime.CompilerServices;

namespace helloworld
{
    class Program
    {
        static T TryGetValueWorker<T>()
            where T : class
        {
            return Unsafe.As<T>(null);
        }

        static void Main(string[] args)
        {
            Console.WriteLine("Hello world!");
        }
    }
}
```
helloworld.csproj:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
  </PropertyGroup>

  <PropertyGroup>
   <LangVersion>preview</LangVersion>
  </PropertyGroup>

</Project>
```

### Current Behavior

Mono gets a SIGSEGV when it tries to compile gsharedvt method `T_GSHAREDVT helloworld.Program:TryGetValueWorker<T_GSHAREDVT> ()`.

### Expected Behavior

AOT image is created successfully. Also I think as `T` in `T TryGetValueWorker<T>()` is always class gsharedvt version shouldn't be compiled at all.

## On which platforms did you notice this

[ ] macOS
[x] Linux
[ ] Windows

**Version Used**:

```
$ <mono_repo>/mono/mini/mono --version
Mono JIT compiler version 6.5.0 (master/40f28b2e637 Fri Aug  2 15:41:47 UTC 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
[...]

PR #16006, opened 2019-08-03
[jit] Avoid asserting on Unsafe.As<T> in gsharedvt methods.

vargaz:
Fixes https://github.com/mono/mono/issues/15992.

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -268,6 +268,9 @@ emit_unsafe_intrinsics (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSignatu
 		g_assert (ctx);
 		g_assert (ctx->method_inst);
 
+		t = ctx->method_inst->type_argv [0];
+		if (mini_is_gsharedvt_variable_type (t))
+			return NULL;
 		if (ctx->method_inst->type_argc == 2) {
 			dreg = alloc_preg (cfg);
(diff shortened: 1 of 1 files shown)


mono/mini/intrinsics.c
```diff
@@ -268,6 +268,9 @@ emit_unsafe_intrinsics (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSignatu
 		g_assert (ctx);
 		g_assert (ctx->method_inst);
 
+		t = ctx->method_inst->type_argv [0];
+		if (mini_is_gsharedvt_variable_type (t))
+			return NULL;
 		if (ctx->method_inst->type_argc == 2) {
```
CoffeeFlux: Nit: else if?

lewurm:
I think the fix is wrong: We have to intrinsify `Unsafe.As` in all scenarios as there isn't any managed implementation for it:
https://github.com/mono/mono/blob/6630bcc67f904b257c90c4feec507a7a93a6b246/netcore/System.Private.CoreLib/shared/Internal/Runtime/CompilerServices/Unsafe.cs#L76-L103

@akoeplinger is running into this issue on iOS SDK (with netcore) in this code snippet:
https://github.com/dotnet/corefx/blob/15e7e1569a8c3a5502c476c31b60e06d633b3efa/src/Common/src/CoreLib/System/Runtime/CompilerServices/AsyncTaskMethodBuilderT.cs#L109-L113

@vargaz how hard would it be to implement the `gsharedvt` case?

Issue #15908, opened 2019-07-30
Reduce size of MonoVM for .NET5


marek-safar:
We can build much smaller Mono runtime for .NET Core purposes by simply removing code we don't need in this configuration. We did a few easy initial steps but we can go much further.

This list is not comprehensive but parts like 
- [ ] Culture Data
- [ ] Any PAL related code
- [ ] Unused icalls

CoffeeFlux:
There are a ton of icalls that have no prototype on netcore and raise a warning as a result, at least when building for OSX, and any efforts to clean that up would be greatly appreciated.

jaykrell:
For every warning you can `#if netcore` out the function, and what it uses, if there is no other use.

Or you can just squash the warning. Imho it isn't actually a useful warning.
The warning should only occur on the use, not the implementation, and if there is no use, as there isn't, then no warning.

And let the static linker do its job. It removes dead code well.
I don't see this actually making the runtime smaller, and it will mean that building for netcore gets less coverage of building for regular.
If regular is to be deleted in time then the `#if` helps that later effort.

jaykrell:
Part of the point being, is that `#if` to remove code and other purposes is overused.
The code is often removed anyway by the static linker, and this means that building for configuration X fails to show errors from configuration Y.
It is better imho to `if`.
Many people agree, but not around here.
We should check if netcore is already smaller, i.e. due to `icall-def-netcore.h`.

EgorBo:
Also, related: https://github.com/dotnet/coreclr/issues/25684 .

PR #16009, opened 2019-08-03
[netcore] ifdef out unused icalls/files.

vargaz:
Contributes to #15908


mono/metadata/mono-security.c
```diff
@@ -640,3 +642,5 @@ ves_icall_System_Security_SecureString_EncryptInternal (MonoArrayHandle data, Mo
 {
 	mono_invoke_protected_memory_method (data, scope, "Protect", &mono_method_securestring_encrypt, error);
 }
+
+#endif /* ENABLE_NETCORE */
```
jaykrell: MONO_EMPTY_SOURCE_FILE?
Can be under ‘#else’ or just always.

mono/metadata/rand.c
```diff
@@ -46,3 +48,5 @@ ves_icall_System_Security_Cryptography_RNGCryptoServiceProvider_RngClose (gpoint
 {
 	mono_rand_close (handle);
 }
+
+#endif /* ENABLE_NETCORE */
```
jaykrell: MONO_EMPTY_SOURCE_FILE?
Can be under ‘#else’ or just always.

vargaz:
@monojenkins build failed

PR #16012, opened 2019-08-04
[netcore] Make the load hook ALC-aware

CoffeeFlux:
The meat of this PR is what the title suggests, and with this commit all the basic ALC functionality should be working properly. Probably worth checking if this makes any new tests pass.

Additionally, this PR includes a second commit that switches over the search hook to use loaded_assemblies from the ALCs. This may break some existing tests due to remaining unimplemented ALC behavior, notably the native loading mechanism needed for Openssl in some of the cryptography tests. My inclination is to include the commit regardless and just disable the failing tests so that any future ALC work can watch for regressions in the loader, but I would appreciate input from @steveisok on how important those tests are. I've not disabled anything yet in order to see exactly what new failures are introduced.

### DIFF
--- a/mono/dis/main.c
+++ b/mono/dis/main.c
@@ -1936,7 +1936,7 @@ monodis_preload (MonoAssemblyName *aname,
 static GList *loaded_assemblies = NULL;
 
 static void
-monodis_assembly_load_hook (MonoAssembly *assembly, gpointer user_data)
+monodis_assembly_load_hook (MonoAssemblyLoadContext *alc, MonoAssembly *assembly, gpointer user_data, MonoError *error)
 {
 	loaded_assemblies = g_list_prepend (loaded_assemblies, assembly);
 }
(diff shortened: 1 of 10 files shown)


filipnavara:
The non-netcore builds are not compiling (`mono_alc_domain` not available in appdomain.c). On the other hand, the netcore tests are passing on most platform. The System.ComponentModel.Composition failures on ARM64 are likely the flaky tests that fail quite often. It has to be investigated but it is most likely not caused by this PR.

CoffeeFlux:
I'm surprised the crypto tests are passing on here - they appear to still have issues on my local machine, which I suspect indicates issues with the native loader. If it's passing on CI though that's convenient since I can just merge this for now and then hunt down what's going on locally. For reference, the output when running `make run-tests-corefx-System.Security.Cryptography.Algorithms.Tests`: https://gist.github.com/CoffeeFlux/c2c94b37015018f0a73cd748cfb1c511

I've opted to make the non-netcore behavior of `mono_alc_domain` to return the result of `mono_domain_get` instead of NULL since I think this will make it less likely for a stray NULL to slip in somewhere on legacy mono and cause issues. I am open to the alternative of returning NULL and adding asserts at callers, but I think it carries significant capacity for mistakes.

CoffeeFlux:
@monojenkins build failed

Issue #15994, opened 2019-08-02
Coverity: derefenrecing null in profiler (CID: 1307043)


migueldeicaza:
In the file `mono/profiler/mprof-report.c` in `dump_samples` the call to `qsort (cachedm,...)` could end up with `cachedm` equal to a null pointer if the samples are zero.

PR #16016, opened 2019-08-04
Skip degenerate qsort: num < 2, size == 0, base == 0

jaykrell:
Fixes https://github.com/mono/mono/issues/15994

Some is redundant for some qsort implementations.

### DIFF
--- a/mono/dis/main.c
+++ b/mono/dis/main.c
@@ -1782,9 +1782,9 @@ sort_filter_elems (void)
 
 	for (item = filter_list; item; item = item->next) {
 		ifilter = (ImageFilter *)item->data;
-		qsort (ifilter->types.elems, ifilter->types.count, sizeof (int), int_cmp);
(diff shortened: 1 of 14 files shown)


mono/utils/mono-qsort.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <stdlib.h>
+
+static inline
+void
+mono_qsort (void* base, size_t num, size_t size, int (*compare)(const void*, const void*))
+{
+	g_assert (compare);
+	g_assert (size);
+	if (num < 2 || !size || !base)
```
lambdageek: `!size` is redundant - you assert that size is non-zero on the previous line.

jaykrell: Code as if assert might be disabled?

jaykrell:
Some changes:
 moved to glib.h; still called mono_qsort
 restored some of the `if` that guarded some qsort, redundant but maybe worthwhile

jaykrell:
@monojenkins build failed

PR #16030, opened 2019-08-05
Swap bytes for tcp port number because windows API gives us port in n…

UnityAlex:
…etwork order.

Addresses the FIXME in Win32IPGlobalProperties.cs :: FillTcpTable

Fixed Unity issue: https://issuetracker.unity3d.com/issues/ipglobalproperties-dot-getactivetcpconnections-return-different-port-numbers-when-compared-to-cli-netstat-a-output


CoffeeFlux:
@monojenkins build failed

Issue #15575, opened 2019-07-04
Crash in FileSystemWatcher processing code


Therzok:
## Steps to Reproduce

No repro steps available, see https://devdiv.visualstudio.com/DevDiv/_workitems/edit/916147

### Current Behavior
Crash in FSW processing code

### Expected Behavior
No crash.

## On which platforms did you notice this

[*] macOS
[ ] Linux
[ ] Windows

**Version Used**:
5.18.1.28 (15), 6.0.0.296 (6), 6.0.0.283 (4), 5.18.1.24 (2)

## Stacktrace

```
System.ThrowHelper.ThrowArgumentOutOfRangeException()
System.IO.CoreFX.FileSystemWatcher.RunningInstance.ProcessEvents(Int32,Byte**,FSEventStreamEventFlags[],UInt64[],FileSystemWatcher)
System.IO.CoreFX.FileSystemWatcher.RunningInstance.<>c__XXX.<FileSystemEventCallback>b__XXX(Object)
System.Threading.ExecutionContext.RunInternal(ExecutionContext,ContextCallback,Object,Boolean)
System.Threading.ExecutionContext.Run(ExecutionContext,ContextCallback,Object,Boolean)
System.Threading.ExecutionContext.Run(ExecutionContext,ContextCallback,Object)
System.IO.CoreFX.FileSystemWatcher.RunningInstance.FileSystemEventCallback(IntPtr,IntPtr,IntPtr,Byte**,FSEventStreamEventFlags[],UInt64[])
System.IO.CoreFX.FileSystemWatcher.RunningInstance.StaticWatcherRunLoopManager.WatchForFileSystemEventsThreadStart(Object)
System.Threading.ThreadHelper.ThreadStart_Context(Object)
System.Threading.ExecutionContext.RunInternal(ExecutionContext,ContextCallback,Object,Boolean)
System.Threading.ExecutionContext.Run(ExecutionContext,ContextCallback,Object,Boolean)
System.Threading.ExecutionContext.Run(ExecutionContext,ContextCallback,Object)
System.Threading.ThreadHelper.ThreadStart(Object)
```

steveisok:
@alexischr Can you think of a reason why this happens?  If it's out of bounds on an array, the two passed in to the function are the only candidates I see.

marek-safar:
We should at least throw a better exception

slluis:
FYI, this is the 3rd top crash in VS for Mac 8.2.

steveisok:
[...]

PR #16049, opened 2019-08-06
[System] Guard against attempting to creating out-of-range spans in m…

alexischr:
…acOS FileSystemWatcher

Should fix https://github.com/mono/mono/issues/15575


marek-safar:
Should we upstream the fix?

steveisok:
@marek-safar Since this is intended to fix a crash in VSM, can we move this through and consider upstreaming separate?

steveisok:
@monojenkins build failed

marek-safar:
Yes but I am not sure it's the right fix if CoreFX is not experiencing the same crash

alexischr:
@marek-safar @steveisok I brought this to upstream attention here: https://github.com/dotnet/corefx/issues/40054 - they don't have an issue reported of this same problem. There is no opinion so far.

I should add a check that these event paths start with `_fullDirectory` (not just that the lengths are okay), but it would be nice to get a nod that this is an expected problem since I have not been able to reproduce the issue yet.

steveisok:
Is CoreFX actively testing against Catalina?  All but a few cases of the crash are happening there.

alexischr:
@monojenkins build failed

steveisok:
@monojenkins backport 2019-06

steveisok:
@monojenkins backport 2019-08

alexischr:
Failures in `System.ComponentModel.Composition.Tests` are unrelated

Issue #16046, opened 2019-08-06
[netcore] AOT with unresolved dependencies hits asserts


filipnavara:
The code introduced in https://github.com/mono/mono/pull/15946 runs managed callbacks for assembly resolution. Unfortunately that doesn't work when running in `--aot` mode. We should probably gurad it with `mono_runtime_get_no_exec()` and bail out without trying to run managed code.

```
Invoking method 'System.Runtime.Loader.AssemblyLoadContext:MonoResolveUsingResolvingEvent (intptr,string)' when running in no-exec mode.
```

/cc @lambdageek @CoffeeFlux

PR #16055, opened 2019-08-06
[netcore] Avoid calling load hooks during AOT.

vargaz:
Fixes https://github.com/mono/mono/issues/16046.


vargaz:
@monojenkins build failed

PR #16056, opened 2019-08-06
Use Memcpy instead of memcpy1 in System.Buffer.MemoryCopy (..., long, long)

mandawah:
Correct what appears to be a typo in System.MemoryCopy with "long, long" size arguments: use Memcpy instead of memcpy1 in the same way it is done in System.Buffer.MemoryCopy (void*, void*, ulong, ulong).

Consequently, copy is not done byte per byte for the last int.MaxValue bytes when MemoryCopy is called with long argument (most common case I guess). This should provide noticeable performance increase.


akoeplinger:
@monojenkins build

Issue #14829, opened 2019-06-06
[netcore]Implement ThreadPool.CompletedWorkItemCount


steveisok:
This may or may not help, but here's how coreclr does it:

https://github.com/dotnet/coreclr/blob/5d16005b984b9b27ce59245bca87aa225b2ca0b7/src/vm/threads.cpp#L8026

PR #16076, opened 2019-08-06
[netcore] Implement missing ThreadPool methods (for metrics)

EgorBo:
Fixes https://github.com/mono/mono/issues/14829

### DIFF
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -356,6 +356,25 @@ mono_threadpool_worker_request (void)
 	mono_refcount_dec (&worker);
 }
 
+#if ENABLE_NETCORE
+gint64 mono_threadpool_worker_get_completed_threads_count (void)
+{
+	return worker.heuristic_completions;
+}
+
+gint64 mono_threadpool_worker_get_pending_threads_count (void)
+{
+	ThreadPoolWorkerCounter const counter = COUNTER_READ ();
+	return counter._.starting;
+}
+
+gint32 mono_threadpool_worker_get_threads_count (void)
+{
+	ThreadPoolWorkerCounter const counter = COUNTER_READ ();
+	return counter._.working;
+}
+#endif
+
 /* return TRUE if timeout, FALSE otherwise (worker unpark or interrupt) */
 static gboolean
 worker_park (void)
--- a/mono/metadata/threadpool-worker-wasm.c
+++ b/mono/metadata/threadpool-worker-wasm.c
@@ -75,3 +75,20 @@ mono_threadpool_worker_notify_completed (void)
 {
 	return FALSE;
 }
+
+#if ENABLE_NETCORE
+gint64 mono_threadpool_worker_get_completed_threads_count (void)
+{
+	return 0;
+}
+
+gint64 mono_threadpool_worker_get_pending_threads_count (void)
+{
+	return 0;
+}
+
+gint32 mono_threadpool_worker_get_threads_count (void)
+{
+	return 0;
+}
+#endif
--- a/mono/metadata/threadpool-worker.h
+++ b/mono/metadata/threadpool-worker.h
@@ -21,6 +21,17 @@ mono_threadpool_worker_request (void);
 gboolean
 mono_threadpool_worker_notify_completed (void);
 
+#if ENABLE_NETCORE
+gint64
+mono_threadpool_worker_get_completed_threads_count (void);
+
+gint64
+mono_threadpool_worker_get_pending_threads_count (void);
+
+gint32
+mono_threadpool_worker_get_threads_count (void);
+#endif
+
 gint32
 mono_threadpool_worker_get_min (void);
 gboolean
--- a/netcore/System.Private.CoreLib/System.Private.CoreLib.csproj
+++ b/netcore/System.Private.CoreLib/System.Private.CoreLib.csproj
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\..\eng\Versions.props"/>
   <PropertyGroup>
     <EnableDefaultItems>false</EnableDefaultItems>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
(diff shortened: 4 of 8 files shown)


netcore/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\..\eng\Versions.props"/>
```
EgorBo: fixes a warning during build (already imported )

mono/metadata/threadpool-worker.h
```diff
@@ -21,6 +21,17 @@ mono_threadpool_worker_request (void);
 gboolean
 mono_threadpool_worker_notify_completed (void);
 
+#if ENABLE_NETCORE
```
lambdageek:
```suggestion
#ifdef ENABLE_NETCORE
```

mono/metadata/threadpool-worker-wasm.c
```diff
@@ -75,3 +75,20 @@ mono_threadpool_worker_notify_completed (void)
 {
 	return FALSE;
 }
+
+#if ENABLE_NETCORE
```
lambdageek:
```suggestion
#ifdef ENABLE_NETCORE
```

netcore/System.Private.CoreLib/src/System.Threading/ThreadPool.cs
```diff
@@ -73,6 +74,16 @@ internal static void NotifyWorkItemProgress ()
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
 		static extern void GetAvailableThreadsNative (out int workerThreads, out int completionPortThreads);
 
+		[MethodImpl(MethodImplOptions.InternalCall)]
+		static extern long GetCompletedWorkItemCount ();
+
+		[MethodImpl(MethodImplOptions.InternalCall)]
+		static extern long GetPendingWorkItemCount();
```
lambdageek: nit: Can we name it `GetPendingUnmanagedWorkItemCount`?

mono/metadata/threadpool-worker.h
```diff
@@ -21,6 +21,17 @@ mono_threadpool_worker_request (void);
...
+gint64
+mono_threadpool_worker_get_completed_threads_count (void);
+
+gint64
+mono_threadpool_worker_get_pending_threads_count (void);
```
lambdageek: why isn't this called `mono_threadpool_worker_get_pending_unmanaged_work_item_count`?

mono/metadata/threadpool-worker-default.c
```diff
@@ -356,6 +356,25 @@ mono_threadpool_worker_request (void)
...
+#if ENABLE_NETCORE
+gint64 mono_threadpool_worker_get_completed_threads_count (void)
+{
+	return worker.heuristic_completions;
+}
+
+gint64 mono_threadpool_worker_get_pending_threads_count (void)
+{
+	ThreadPoolWorkerCounter const counter = COUNTER_READ ();
+	return counter._.starting;
```
lambdageek: I don't think this is right:
1. this is called for `GetPendingUnmanagedWorkItemCount` not for pending threads.
2. The CoreCLR impl is counting tasks I think, not worker threads that are just starting up: https://github.com/dotnet/coreclr/blob/9bd2787a9dd2aa4d2b7d4f72afebc3dbe896e896/src/vm/comthreadpool.cpp#L208-L213

mono/metadata/threadpool-worker-default.c
```diff
@@ -356,6 +356,25 @@ mono_threadpool_worker_request (void)
 	mono_refcount_dec (&worker);
 }
 
+#if ENABLE_NETCORE
+gint64 mono_threadpool_worker_get_completed_threads_count (void)
```
lambdageek: should be called `mono_threadpool_worker_get_completed_work_items_count`, I think

mono/metadata/threadpool-worker.h
EgorBo: @lambdageek let me double check, the MetricTest is quite complicated and according to logs all three values had the same values on both coreclr and mono in different steps

mono/metadata/threadpool-worker-default.c
EgorBo: @lambdageek 
```csharp
Console.WriteLine($"1. tc:{ThreadPool.ThreadCount},comp:{ThreadPool.CompletedWorkItemCount},pend:{ThreadPool.PendingWorkItemCount}");
ThreadPool.QueueUserWorkItem(_ => {  while (true) ; });
Console.WriteLine($"2. tc:{ThreadPool.ThreadCount},comp:{ThreadPool.CompletedWorkItemCount},pend:{ThreadPool.PendingWorkItemCount}");
```
With my changes it prints the same what coreclr does

EgorBo: However, I am still checking...

EgorBo: CoreRT doesn't implement it: https://github.com/dotnet/corert/blob/c880f866597fc5b6f9348c60ee21b63ff2a893de/src/System.Private.CoreLib/src/System/Threading/ThreadPool.CoreRT.cs#L58

lambdageek: I think we should also return 0.

The only instance I found so far in CoreCLR where they queue an unmanaged work item is in their tiered JIT: https://github.com/dotnet/coreclr/blob/060c8a228fbe397b4207eaed2aedff7bbbf32586/src/vm/tieredcompilation.cpp#L550

(That ends up calling QueueUnmanagedWorkRequest which increments `m_NumRequests` which is the counter that's returned by `GetPendingUnmanagedWorkItemCount`)

EgorBo: I agree

steveisok:
@EgorBo The CI lane seems to be just a flaky test.  @lambdageek is this GTG?

lambdageek:
@monojenkins build

mono/metadata/threadpool-worker-default.c
jaykrell: Can this be faster somehow? Provide an address that managed holds on to and returns repeatedly? An icall to read a global variable seems, slow. Perhaps there can be some data driven relationship with method-to-ir to inline these, without being a special case? Similar to the slow increment I introduced as part of my first or second mono commit, that I still regret, perf-wise.

jaykrell: ps: Get it to master independent of my question, of course, but it might be nice to have a better system sooner than later, for binding to and incrementing and atomically incrementing global int32, int64, gsize, something like that.

EgorBo:
@monojenkins build failed

Issue #16034, opened 2019-08-05
MSBuild fatal unhandled exception


mrward:
## Steps to Reproduce

Install Mono 6.4.0.148 (2019-06/84bc821cdc3 Thu Aug  1 07:34:53 EDT 2019)
Install .NET Core SDK 2.1.701

mkdir angular-test
cd angular-test
dotnet new angular
msbuild
msbuild /t:Rebuild

### Current Behavior

Build succeed.
Fatal MSBuild error.
Exit code 255 returned from msbuild.

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.ObjectDisposedException: Safe handle has been closed
  at System.Runtime.InteropServices.SafeHandle.DangerousReleaseInternal (System.Boolean dispose) [0x00031] in <d826688bf0184e629043e8d66af1d42b>:0 
  at System.Runtime.InteropServices.SafeHandle.InternalFinalize () [0x00008] in <d826688bf0184e629043e8d66af1d42b>:0 
  at System.Runtime.InteropServices.SafeHandle.Dispose (System.Boolean disposing) [0x0000a] in <d826688bf0184e629043e8d66af1d42b>:0 
  at System.Runtime.InteropServices.SafeHandle.Finalize () [0x00000] in <d826688bf0184e629043e8d66af1d42b>:0 
```

We believe that the above exception is crashing MSBuild when running the VS Mac tests sometimes.

### Expected Behavior

MSBuild succeeds with no errors. Exit code is 0.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

6.4.0.148 (2019-06/84bc821cdc3 Thu Aug  1 07:34:53 EDT 2019)

We believe 6.4.0.110 to be OK.

## Stacktrace

```
[ERROR] FATAL UNHANDLED EXCEPTION: System.ObjectDisposedException: Safe handle has been closed
  at System.Runtime.InteropServices.SafeHandle.DangerousReleaseInternal (System.Boolean dispose) [0x00031] in <d826688bf0184e629043e8d66af1d42b>:0 
  at System.Runtime.InteropServices.SafeHandle.InternalFinalize () [0x00008] in <d826688bf0184e629043e8d66af1d42b>:0 
  at System.Runtime.InteropServices.SafeHandle.Dispose (System.Boolean disposing) [0x0000a] in <d826688bf0184e629043e8d66af1d42b>:0 
[...]

PR #16077, opened 2019-08-06
Bump msbuild to 'cd5b32d' to disable compiler server

lewing:


lewing:
@monojenkins build failed

lewing:
In response to https://github.com/mono/mono/issues/16034 while we dig into the problem

PR #16082, opened 2019-08-07
[llvm] Explicitly check for errors when calling LegacyIRCompileLayer::addModule.

imhameed:
LLVM built with LLVM_ENABLE_ASSERTIONS=On will crash if errors are not
explicitly consumed.


vargaz:
@monojenkins squash

vargaz:
@monojenkins build failed

PR #16089, opened 2019-08-07
Fix G_FILE_TEST_IS_EXECUTABLE test on win32.

madewokherd:
This was obviously meant to check for equality, but it actually
checks for inequality.

Sometimes Process.Start fails to add the filename of an executable with a relative path to the command line, and CreateProcess fails with "file not found", because it fails this test.

### DIFF
--- a/mono/eglib/gfile-win32.c
+++ b/mono/eglib/gfile-win32.c
@@ -115,7 +115,7 @@ g_file_test (const gchar *filename, GFileTest test)
 
 	if ((test & G_FILE_TEST_IS_EXECUTABLE) != 0) {
 		size_t len = strlen (filename);
-		if (len > 4 && strcmp (filename + len-3, "exe"))
+		if (len > 4 && !strcmp (filename + len-3, "exe"))
 		    return TRUE;
 		    
 		return FALSE;


jaykrell:
CreateProcess does not give this meaning to "exe" so Mono should not either.
Besides, the handling:
 foo.EXE fails
 fooexe passes

madewokherd:
I was trying to make the minimal change, but I'm honestly not sure why we have checks for the executable at all. Can't we rely on CreateProcess to find it if it's in the path?

akoeplinger:
Yeah I agree, I'd just remove this check for the `.exe` extension.

jaykrell:
Here is old Visual C++ 5.0 logic in `stat` though, ugh.
```
        /* see if file appears to be executable - check extension of name */

        if (p = _tcsrchr(name, _T('.'))) {
            if ( !_tcsicmp(p, _T(".exe")) ||
                 !_tcsicmp(p, _T(".cmd")) ||
                 !_tcsicmp(p, _T(".bat")) ||
                 !_tcsicmp(p, _T(".com")) )
                uxmode |= _S_IEXEC;
```
still dubious.
You can "run" more than this. .com is for example more.com.

I should check current.

madewokherd:
According to MSDN, batch files aren't executable: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

One could use SHGetFileInfo with SHGFI_EXETYPE to test for an executable file, but I'm still not sure why we'd want to do this.

jaykrell:
Sure they are.

copy c:\windows\system32\notepad.exe 1.bat
\bin\amd64\cdb .\1.bat

As well I think CreateProcess special cases .bat and .cmd and sends them to cmd.exe, if in fact they are not PEs.

A PE can have any extension at all, or even no extension.

madewokherd:
> As well I think CreateProcess special cases .bat and .cmd and sends them to cmd.exe, if in fact they are not PEs.

MSDN specifically claims in CreateProcess documentation that it does not do this and explains how to start a batch file with cmd.exe.

jaykrell:
MSDN is incorrect.

```
C:\s>type 1.bat
echo %1


C:\s>type 1.c

#include <windows.h>
#include <stdio.h>

int main()
{
BOOL success;
STARTUPINFOW startInfo = {sizeof(startInfo)};
PROCESS_INFORMATION processInfo = { 0 };

static const WCHAR app[] = L"c:\\s\\1.bat";
WCHAR command[] = L"c:\\s\\1.bat foo";

success = CreateProcessW(
  L"c:\\s\\1.bat",
  L"c:\\s\\1.bat foo",
  0, // process security
  0, // thread security
  0, // inherit
  0, // flags
  0, // env
  0, // curdir
  &startInfo,
  &processInfo);

DWORD error = GetLastError();

wprintf(L"%X %X %p\n", success, GetLastError(), processInfo.hProcess);

if (success)
{
 WaitForSingleObject(processInfo.hProcess, INFINITE);
 GetExitCodeProcess(processInfo.hProcess, &error);
}
exit(error);
}

C:\s>cl 1.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.21.27702.2 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

1.c
Microsoft (R) Incremental Linker Version 14.21.27702.2
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:1.exe
1.obj

C:\s>.\1.exe
1 0 00000000000000AC

C:\s>echo foo
foo

```

madewokherd:
g_find_program_in_path specifically looks for .exe, .bat, .com, and .cmd suffixes. We only search the path if the filename, exactly as given, doesn't exist. Should suffixes be appended to relative paths? Should this behavior be affected by SetSearchPathMode? Should a relative path containing a directory part really search $PATH? How much of this does CreateProcess take care of?

I was really hoping to not have to go down this rabbit hole.

process_get_shell_arguments is also broken because it treats a FALSE result from process_complete_path as success.

madewokherd:
I did some testing of CreateProcess.

If lpApplicationName is specified, it uses that exact filename.

If lpApplicationName is NULL, it appends .exe (if there's no extension) and then searches the current directory and PATH. SetSearchPathMode doesn't seem to have any effect.

madewokherd:
Testing the .NET side:
```cs
using System;
using System.Diagnostics;

class CreateProcessTest
{
    static void Main(string [] argv)
    {
        ProcessStartInfo si = new ProcessStartInfo(argv[0]);
        if (argv.Length >= 2)
            si.Arguments = argv[1];
        si.UseShellExecute = false;
        Process p = new Process();
        p.StartInfo = si;
        p.Start();
        p.WaitForExit();
        Console.WriteLine(p.ExitCode);
    }
}
```

Compiled as nettest.exe.

"nettest.exe hostname" works on Windows, so .exe is appended when searching the path.
"nettest.exe hostname.exe" works.
"nettest.exe more" fails, so .com is not appended.
"nettest.exe more.com" works.
"nettest.exe ./more.com" fails, so paths with a directory name shouldn't go to a PATH search.
"nettest.exe battest.bat" works if the file exists.
"nettest.exe cmdtest.cmd" works if the file exists.
"nettest.exe nosuffix" fails if nosuffix is a PE executable.
"nettest.exe nosuffix." works.
"nettest.exe ./nosuffix" works.
"nettest.exe nettest.exe" works (and the child process raises an exception because I didn't give it any arguments).
"nettest.exe nettest" works.
"nettest.exe ./nettest" works.

Testing CreateProcess's behavior with a NULL application name gives me identical results for all of those cases. So I think process_complete_path should always quote the application name and prepend it to the command-line as is on Windows.

madewokherd:
I'll make a new PR which removes the executable path checking/searching on Windows. Closing this one.

PR #16091, opened 2019-08-07
Fix broken OS-provided zlib linking, and assign z_ prefix to bundled zlib

directhex:
This adds a `z_` prefix to all symbols from Zlib.

This should make it easier to filter out zlib symbols, since it gives a common prefix to them all (e.g. for building Xamarin.MaciOS which cares about all symbols)

### DIFF
--- a/mono/metadata/debug-mono-ppdb.c
+++ b/mono/metadata/debug-mono-ppdb.c
@@ -32,7 +32,10 @@
 #if HOST_WIN32
 #include "../../support/zlib.h"
 #elif HAVE_SYS_ZLIB
+#undef Z_PREFIX
 #include <zlib.h>
+#define	z_inflate(strm, flush)			inflate(strm, flush)
+#define	z_inflateInit2(strm, windowBits)	inflateInit2(strm, windowBits)
 #endif
 
 #include "debug-mono-ppdb.h"
--- a/mono/metadata/debug-mono-ppdb.c
+++ b/mono/metadata/debug-mono-ppdb.c
@@ -184,9 +184,9 @@ mono_ppdb_load_file (MonoImage *image, const guint8 *raw_contents, int size)
 		stream.next_in = ppdb_data;
 		stream.avail_out = ppdb_size;
 		stream.next_out = data;
-		int res = inflateInit2 (&stream, -15);
+		int res = z_inflateInit2 (&stream, -15);
 		g_assert (res == Z_OK);
-		res = inflate (&stream, Z_NO_FLUSH);
+		res = z_inflate (&stream, Z_NO_FLUSH);
 		g_assert (res == Z_STREAM_END);
 
 		g_assert (ppdb_size > 4);
--- a/configure.ac
+++ b/configure.ac
@@ -695,7 +695,7 @@ case "$host" in
 		have_zlib=no
 		zlib_msg="bundled zlib"
 		;;
-	default)
+	*)
 		AC_CHECK_HEADER(zlib.h, [have_zlib=yes], [have_zlib=no])
 		if test x$have_zlib = xyes; then
 			AC_TRY_COMPILE([#include <zlib.h>], [
--- a/configure.ac
+++ b/configure.ac
@@ -708,10 +708,14 @@ case "$host" in
 				zlib_msg="system zlib"
 				have_sys_zlib=yes
 			],[
-				AC_MSG_RESULT(Using embedded zlib)
+				AC_MSG_RESULT(zlib too old, using embedded zlib)
 				have_zlib=no
 				zlib_msg="bundled zlib"
 			])
+		else
+			AC_MSG_RESULT(zlib not found, using embedded zlib)
+			have_zlib=no
+			zlib_msg="bundled zlib"
 		fi
 		;;
 esac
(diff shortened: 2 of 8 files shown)


jaykrell:
I suggest `monoz` or `mono_z` or `monozlib` or mono`-something.

mono/metadata/debug-mono-ppdb.c
```diff
@@ -184,9 +184,9 @@ mono_ppdb_load_file (MonoImage *image, const guint8 *raw_contents, int size)
 		stream.next_in = ppdb_data;
 		stream.avail_out = ppdb_size;
 		stream.next_out = data;
-		int res = inflateInit2 (&stream, -15);
+		int res = z_inflateInit2 (&stream, -15);
```
jaykrell: Shouldn't this be conditional if we use bundled zlib or not?

directhex:
@jaykrell I don't disagree, but we need to modify zconf.h to do that, since it's "if Z_PREFIX is defined, use `z_`" not "use the value of Z_PREFIX

mono/metadata/debug-mono-ppdb.c
directhex: Probably.

jaykrell:
Most likely zlib has a configure step that runs sed against the header, that I omitted.
Maybe bring that over.

jaykrell:
Or not:
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/configure#L596

visibility might make it not matter though?
We could sed the file ourselves, or define the z_ symbols to monoz. (define is the partial sed)

directhex:
I'm good with literally any solution which means I don't need to manually add every symbol from Zlib to https://github.com/xamarin/xamarin-macios/blob/master/tests/mtouch/MiscTests.cs#L84-L199

jaykrell:
What does it do? For what builds?
I'd like config.log to be easily seen in CI.
Bundled zlib I'd'a thought is mainly or only for Win32.

directhex:
One of Xamarin.MacIOS's tests is "which symbols exist in the .a files, and are they supposed to be there?" - so e.g. http://xamarin-storage/jenkins/xamarin-macios/PR-6723/69917933f62daede9ff57236af6ac62db4f8267a/1/jenkins-results/tests/index.html is failing.

jaykrell:
I guess this is a significant improvement, and possibly ideal even.
Assuming nobody really uses the static lib, and "visibility" hides the symbols.
Still should figure out why bundled zlib ever used, except on Win32.
And Win32 "visibility" does work well, for dynamic linkage.

directhex:
Well the OSX failures are fascinating.

directhex:
OKAY.

There are essentially three ways we can potentially consume zlib:

* dynamic linking against libz.so
* static linking a specified libz.a
* compiling in the bundled libz.c

Due to errors in configure, and Makefile.am, and ifdefs in .c files, we have not been correctly handling all three cases.

In addition, we want to be able to handle case 3 by renaming the functions from foo to z_foo (which is supported by zlib via one define, Z_PREFIX).

directhex:
The wasm failures are caused by a static linking problem - it links zlib twice (once as a standalone, once as part of libmonoruntime, which now has zlib inside it)

directhex:
@monojenkins build failed

vargaz:
So on wasm, we have 2 zlibs:
- the one used by the runtime for embedded pdbs.
- the one linked into the app to support System.Compression.

directhex:
@vargaz I may have broken that, then.

How have you overcome:

```
wasm-ld: error: duplicate symbol: z_errmsg
>>> defined in /mnt/jenkins/workspace/test-mono-pull-request-wasm/sdks/out/wasm-runtime-threads-release/lib/libmonosgen-2.0.a(libmonoruntime_support_la-zutil.o)
>>> defined in /mnt/jenkins/workspace/test-mono-pull-request-wasm/sdks/builds/toolchains/emsdk/.emscripten_cache/wasm-obj/libz.a(zutil.c.o)
```

directhex:
@vargaz I don't see how the embedded PPDB stuff would ever have intentionally worked on wasm. https://github.com/mono/mono/blob/master/mono/metadata/debug-mono-ppdb.c#L177 should not have been true on wasm (but may have, inadvertently, due to build system errors)

directhex:
@monojenkins build failed

directhex:
@monojenkins build failed

kjpou1:
> So on wasm, we have 2 zlibs:
> 
> * the one used by the runtime for embedded pdbs.
> * the one linked into the app to support System.Compression.

The one linked in to support System.Compression is actually the one from emscripten support libs so are probably different from the version used for embedded pdbs.  When I implemented this for wasm, if I remember correctly, the HAVE_SYS_ZLIB was not set.

kjpou1:
What is the difference between a pdb and embedded pdb?

steveisok:
@directhex I'm assuming we still want this?

directhex:
Let's close it for now. The original bug that inspired it has been resolved, without the need for `z_`

jaykrell:
You listed all the symbols?

PR #13857, opened 2019-04-04
[metadata] Create strongly typed versions of mono_assembly_request_prepare - Coverity #1443346

migueldeicaza:
Addresses the family of issues similar to Coverity #1443346.

This came up as part of the Coverity review, and Coverity notices that this is
writting beyond the end of the structure.  While this is correct, this requires
that the idiom be properly followed - that is, that the structure being passed
is of the given size.

Replaced instead with a method that eliminates the room for error with methods
that will initialize the whole structure, and the actual size computation is
done inside the method with the proper type sizes.

While I used methods, I feel that this should use macros.

Update: This both prevents Coverity from reporting a potential overflow, it allows coverity to reason that the values have been properly initialized, so it also removes another class of warnings,

PR #16094, opened 2019-08-07
[metadata] Create strongly typed versions of mono_assembly_request_prepare

lambdageek:
This is #13857 with conflicts resolved. Also fixed up a couple of references to `mono_assembly_request_prepare` that were under ifdefs.

---

Addresses the family of issues similar to Coverity #1443346.

This came up as part of the Coverity review, and Coverity notices that this is
writting beyond the end of the structure. While this is correct, this requires
that the idiom be properly followed - that is, that the structure being passed
is of the given size.

Replaced instead with a method that eliminates the room for error with methods
that will initialize the whole structure, and the actual size computation is
done inside the method with the proper type sizes.

While I used methods, I feel that this should use macros.

Update: This both prevents Coverity from reporting a potential overflow, it allows coverity to reason that the values have been properly initialized, so it also removes another class of warnings,

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -753,21 +753,51 @@ assembly_names_compare_versions (MonoAssemblyName *l, MonoAssemblyName *r, int m
 }
 
 /**
- * mono_assembly_request_prepare:
- * \param req the request to be initialized
- * \param req_size the size of the request structure
+ * mono_assembly_request_prepare_load:
+ * \param req the load request to be initialized
  * \param asmctx the assembly load context kind
  * \param alc the AssemblyLoadContext in netcore
  *
- * Initialize an assembly loader request.  The passed structure \p req must be
- * of size \p req_size.  Its state will be reset and the assembly context kind will be prefilled with \p asmctx.
+ * Initialize an assembly loader request.  Its state will be reset and the assembly context kind will be prefilled with \p asmctx.
  */
 void
-mono_assembly_request_prepare (MonoAssemblyLoadRequest *req, size_t req_size, MonoAssemblyContextKind asmctx, MonoAssemblyLoadContext *alc)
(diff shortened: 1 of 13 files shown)


mono/metadata/assembly.c
```diff
@@ -4339,9 +4369,9 @@ mono_assembly_load_from_gac (MonoAssemblyName *aname,  gchar *filename, MonoImag
 	g_free (culture);
 
 	MonoAssemblyOpenRequest req;
-	mono_assembly_request_prepare (&req.request, sizeof (req),
-	                               refonly ? MONO_ASMCTX_REFONLY : MONO_ASMCTX_DEFAULT,
-	                               mono_domain_default_alc (mono_domain_get ()));
+	mono_assembly_request_prepare_open (&req,
+					    refonly ? MONO_ASMCTX_REFONLY : MONO_ASMCTX_DEFAULT,
+					    mono_domain_default_alc (mono_domain_get ()));
```
CoffeeFlux: Whitespace

mono/metadata/assembly.c
```diff
@@ -4695,9 +4725,9 @@ mono_assembly_load_full (MonoAssemblyName *aname, const char *basedir, MonoImage
 	MonoAssembly *res;
 	MONO_ENTER_GC_UNSAFE;
 	MonoAssemblyByNameRequest req;
-	mono_assembly_request_prepare (&req.request, sizeof (req),
-	                               refonly ? MONO_ASMCTX_REFONLY : MONO_ASMCTX_DEFAULT,
-	                               mono_domain_default_alc (mono_domain_get ()));
+	mono_assembly_request_prepare_byname (&req,
+					      refonly ? MONO_ASMCTX_REFONLY : MONO_ASMCTX_DEFAULT,
+					      mono_domain_default_alc (mono_domain_get ()));
```
CoffeeFlux: Whitespace

Issue #15259, opened 2019-06-19
Client-Side Blazor OrderByDescending


mkArtakMSFT:
_From @matheusjulio96 on Wednesday, June 19, 2019 2:02:04 PM_

The bug: Execute an OrderByDescending query throw the error 'Unhandled exception rendering component: System.NullReferenceException: Object reference not set to an instance of an object.'

1. Using this version of ASP.NET Core 3.0.0-preview6.19307.2 with Client-Side project
2. Run this code:
```
@code{
    class Pet
    {
        public string Name { get; set; }
        public int Age { get; set; }
    }

    protected override void OnInit()
    {
        Pet[] pets = { new Pet { Name="Barley", Age=8 },
                       new Pet { Name="Boots", Age=4 },
                       new Pet { Name="Whiskers", Age=1 } };

        var query = pets.AsQueryable().OrderByDescending(pet => pet.Age);

        foreach (Pet pet in query)
            Console.WriteLine("{0} - {1}", pet.Name, pet.Age);
    }
}
```
3. See error Screenshot
![image](https://user-images.githubusercontent.com/29152563/59771996-55d57600-9281-11e9-832c-100eba869c03.png)

Expected behavior: Execute the query and print the values.

Ps: OrderBy query works fine.

_Copied from original issue: aspnet/AspNetCore#11371_

matheusjulio96:
Seems related with https://github.com/mono/mono/issues/8872

lewing:
```using System;
using System.Runtime.InteropServices;
using System.Linq;

public class TestQueryableOrderBy
{
    public static void Main (String[] args) {
        OnInit();
        Console.WriteLine ("Hello, World!");
    }

    class Pet
    {
        public string Name { get; set; }
        public int Age { get; set; }
    }

    static void OnInit()
    {
        Pet[] pets = { new Pet { Name="Barley", Age=8 },
                       new Pet { Name="Boots", Age=4 },
                       new Pet { Name="Whiskers", Age=1 } };

        var query = pets.AsQueryable().OrderByDescending(pet => pet.Age);

        foreach (Pet pet in query)
            Console.WriteLine("{0} - {1}", pet.Name, pet.Age);
    }
}
```


```Exception:System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #16107, opened 2019-08-08
Bump corefx for #15259 fix

marek-safar:


marek-safar:
@monojenkins commit csproj

PR #16112, opened 2019-08-08
[aprofutil] Install the tool correctly

akoeplinger:
It needed a few tweaks to be usable when Mono is installed to the prefix.


directhex:
Remind me this'll need adding to packages.

PR #16118, opened 2019-08-08
[netcore] Make System.Runtime.Intrinsics.X86.*.IsSupported intrinsic

EgorBo:
We don't support SRI.X86 yet but at least we should be able to eliminate the `IsSupported` branches, e.g.:
```csharp
    static float TestSRI(float x)
    {
        if (Sse2.IsSupported)
        {
            // we don't support SIMD yet so I expect this 'if' and 'sqrt' to be eliminated
            return MathF.Sqrt(x);
        }
        return x * x * x;
    }
```

Current codegen (llvm):
```asm
_Program_TestSRI_single:
	pushq	%rax
	movss	%xmm0, 4(%rsp)
	movq	1850(%rip), %rax
	cmpq	$0, (%rax)
	je	6 <_Program_TestSRI_single+0x1a>
	callq	*1942(%rip)
	callq	225 <plt_System_Runtime_Intrinsics_X86_Sse2_get_IsSupported>
	testb	%al, %al
	je	12 <_Program_TestSRI_single+0x2f>
	movss	4(%rsp), %xmm0
	sqrtss	%xmm0, %xmm0
	popq	%rax
	retq
	movss	4(%rsp), %xmm1
	movaps	%xmm1, %xmm0
	mulss	%xmm0, %xmm0
	mulss	%xmm1, %xmm0
	popq	%rax
	retq
```

After my change:
```asm
_Program_TestSRI_single:
	movaps	%xmm0, %xmm1
	mulss	%xmm1, %xmm1
	mulss	%xmm0, %xmm1
	movaps	%xmm1, %xmm0
	retq
```

Also, minor changes around CoreFX.issues.rsp

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -1689,6 +1689,18 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		}
 	}
 
+#ifdef ENABLE_NETCORE
+	// Return false for IsSupported for all types in System.Runtime.Intrinsics.X86 
+	// as we don't support them now
+	if (in_corlib && 
+		!strcmp ("System.Runtime.Intrinsics.X86", cmethod_klass_name_space) && 
+		!strcmp (cmethod->name, "get_IsSupported")) {
+		EMIT_NEW_ICONST (cfg, ins, 0);
+		ins->type = STACK_I4;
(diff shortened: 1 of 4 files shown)


steveisok:
Would it be more convenient if we pulled in the submodules when llvm is enabled?

vargaz:
@monojenkins build failed

PR #16127, opened 2019-08-08
[netcore] Don't return REFERENCE_MISSING in netcore_load_reference

CoffeeFlux:
As-is, error handling (among other things) will be totally broken.

@monojenkins backport 2019-08


lambdageek:

PR #16129, opened 2019-08-08
Add extra assertion to prevent invoking non-static methods w/o an object

CoffeeFlux:
I had added this while debugging something, but I think it's more broadly useful.

### DIFF
--- a/mono/metadata/object.c
+++ b/mono/metadata/object.c
@@ -3119,6 +3119,7 @@ mono_runtime_try_invoke (MonoMethod *method, void *obj, void **params, MonoObjec
 	MONO_REQ_GC_UNSAFE_MODE;
 
 	g_assert (exc != NULL);
+	g_assert (obj || (method->flags & METHOD_ATTRIBUTE_STATIC));
 
 	if (mono_runtime_get_no_exec ())
 		g_warning ("Invoking method '%s' when running in no-exec mode.\n", mono_method_full_name (method, TRUE));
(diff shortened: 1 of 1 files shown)


mono/metadata/object.c
```diff
@@ -3119,6 +3119,7 @@ mono_runtime_try_invoke (MonoMethod *method, void *obj, void **params, MonoObjec
 	MONO_REQ_GC_UNSAFE_MODE;
 
 	g_assert (exc != NULL);
+	g_assert (obj || (method->flags & METHOD_ATTRIBUTE_STATIC));
```
lambdageek: can this be "if and only if"?
```suggestion
	g_assert ((obj == NULL) == ((method->flags & METHOD_ATTRIBUTE_STATIC) != 0));
```

I dunno.  Too esoteric?

CoffeeFlux: I think that's reasonable so long as you're sure there isn't some weird case where that doesn't hold. The main reason my check isn't as strict is because I was solely worried about a non-static method being passed without an object.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Failures are related - seems to have uncovered some issue in the debugger.

CoffeeFlux:
I realized a check happens later in the callstack with some slightly more complicated logic and warns instead of asserts. I'm still concerned that the child thread in the debugger is hitting it and the warning is just getting swallowed up as a result, but don't have time to deal with it at the moment so this is getting shelved. Maybe revisit later.

Issue #15637, opened 2019-07-09
Support for turning off any processing on native crashes


rolfbjarne:
mono has always tried to print useful information in case of native crashes, but unfortunately this has become more buggy recently, and now it seems the process is just as likely to hang/deadlock, and when that happens we the information we get can be _much_ worse than the standard crash report from the OS (at least on Apple platforms).

In general Apple's crash reports are quite useful, reliable (as long as the process actually crashes and doesn't hang), and when symbolicated contain useful information.

Thus I become sad every time I see mono trying to print unnecessary helpful information, and just makes everything worse 😒

Example output:

```
=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Basic Fault Adddress Reporting
=================================================================
Memory around native instruction pointer (0x102a93610):0x102a93600  fd c3 00 91 f3 03 01 aa f4 03 00 aa 16 00 40 f9  ..............@.
0x102a93610  c8 06 40 f9 e9 03 28 2a 3f 05 40 f2 e9 07 9f 1a  ..@...(*?.@.....
0x102a93620  ea 00 98 52 08 01 0a 8a 1f 15 00 f1 e8 07 9f 1a  ...R............
0x102a93630  17 01 09 0a e0 03 16 aa e1 03 14 aa 45 00 00 94  ............E...

=================================================================
	Native stacktrace:
=================================================================
	0x102982000 - /var/containers/Bundle/Application/460FC1AE-C29E-4A5C-A553-23A5D25A0A32/BCL tests group 4.app/BCL tests group 4 : (null)
	0x102978658 - /var/containers/Bundle/Application/460FC1AE-C29E-4A5C-A553-23A5D25A0A32/BCL tests group 4.app/BCL tests group 4 : (null)
	0x1029857e0 - /var/containers/Bundle/Application/460FC1AE-C29E-4A5C-A553-23A5D25A0A32/BCL tests group 4.app/BCL tests group 4 : (null)
	0x1b7def3b4 - /usr/lib/system/libsystem_platform.dylib : <redacted>
	0x102a92c28 - /var/containers/Bundle/Application/460FC1AE-C29E-4A5C-A553-23A5D25A0A32/BCL tests group 4.app/BCL tests group 4 : (null)
[...]

PR #16135, opened 2019-08-09
[runtime] Disable the rest of Mono's native crash reporting when conf…

alexischr:
…igured with --disable-crash-reporting

Fixes https://github.com/mono/mono/issues/15637

Also unless I am missing something, it's still possible to end up in mono_handle_native_crash even when cross-compiled, so remove an assert.

### DIFF
--- a/mono/mini/mini-exceptions.c
+++ b/mono/mini/mini-exceptions.c
@@ -3292,7 +3292,7 @@ print_stack_frame_to_string (StackFrameInfo *frame, MonoContext *ctx, gpointer d
 	return FALSE;
 }
 
-#ifndef MONO_CROSS_COMPILE
+#if !defined(DISABLE_CRASH_REPORTING) && !defined(MONO_CROSS_COMPILE)
 static gboolean handle_crash_loop = FALSE;
 
 /*
@@ -3369,7 +3369,7 @@ mono_handle_native_crash (const char *signal, MonoContext *mctx, MONO_SIG_HANDLE
 void
 mono_handle_native_crash (const char *signal, MonoContext *mctx, MONO_SIG_HANDLER_INFO_TYPE *info)
 {
-	g_assert_not_reached ();
+	return;
 }
 
 #endif /* !MONO_CROSS_COMPILE */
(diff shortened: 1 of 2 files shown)


mono/mini/mini-exceptions.c
```diff
@@ -3369,7 +3369,7 @@ mono_handle_native_crash (const char *signal, MonoContext *mctx, MONO_SIG_HANDLE
 void
 mono_handle_native_crash (const char *signal, MonoContext *mctx, MONO_SIG_HANDLER_INFO_TYPE *info)
 {
-	g_assert_not_reached ();
```
lambdageek: I think `MONO_CROSS_COMPILE` doesn't mean that the build and host platforms are different.  It means the host and target platforms are different: e.g we're building an OSX amd64 AOT compiler that can target iOS arm64.  So in that case we don't expect to run any managed code on the host, so we don't want to use our own crash handler.

Something like that.

So I think we still want `g_assert_not_reached` here for `MONO_CROSS_COMPILE`.  And for `DISABLE_CRASH_REPORTING` we want a `return` (or possibly `mono_post_native_crash_handler (signal, mctx, info, mono_do_crash_chaining)` - I'm not sure what that does exactly)

lambdageek:
@vargaz Could you review this, too.

vargaz:
On the AOT compiler, we want the old crash handler which prints out native stack traces using lldb.

mono/mini/mini-exceptions.c
alexischr: That was my understanding of `MONO_CROSS_COMPILE` also, but cannot see how that categorically excludes the AOT compiler itself from crashing.

alexischr: Like @vargaz mentioned (if I understand "old crash handler" correctly), perhaps the #ifndef there was wrong in the first place.

lewing:
@monojenkins build failed

SamMonoRT:
@alexischr - any update on this ?

lambdageek:
We're going to revisit this once we get through the current crop of crash reporting work.

PR #16136, opened 2019-08-09
[netcore] Fix inlining

EgorBo:
It turns out my previous change broke inlining for netcore.
I am not sure how to properly fix the case (don't inline throw helpers)

cc @vargaz 

I suspect it might affect some StackTrace related tests, let's check CI.


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #16137, opened 2019-08-09
Update dependencies from corefx

akoeplinger:
Manual update until the arcade fix rolls out that reenables the update bot for our repo.


EgorBo:
Zero failures!

PR #16141, opened 2019-08-09
[System.Drawing.Design] Allow `context` be a null in the `ColorEditor…

NVoronchev:
If one call `ColorEditor.EditValue(IServiceProvider, Object)` ([MSDN](https://docs.microsoft.com/ru-ru/dotnet/api/system.drawing.design.uitypeeditor.editvalue?view=netframework-4.8#System_Drawing_Design_UITypeEditor_EditValue_System_IServiceProvider_System_Object_)), then `context` is `null` and drop-down menu doesn't appear.


akoeplinger:
@monojenkins build

PR #16142, opened 2019-08-09
Fix use of Span.Slice(int start() introduced in https://github.com/mo…

alexischr:
…no/mono/pull/16049

Related to #15575


akoeplinger:
Manually bumped corefx in all the branches.

PR #16145, opened 2019-08-09
[interp] Fix stub_get_resume_state write to out parameter

lambdageek:
Addresses Coverity CID 1452245


lambdageek:
@monojenkins build failed

PR #16149, opened 2019-08-09
[runtime] Initialize thread suspend policy at startup, inline getter

lambdageek:
On some benchmarks we saw `mono_threads_suspend_policy()` using a lot of CPU time.

Rather than doing an init-on-first-use, initialize the policy once upfront when the JIT it starting up and make the getter method static inline.

### DIFF
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -4080,6 +4080,8 @@ mini_init (const char *filename, const char *runtime_version)
 
 	CHECKED_MONO_INIT ();
 
+	mono_threads_suspend_policy_init ();
+
 #if defined(__linux__)
 	if (access ("/proc/self/maps", F_OK) != 0) {
 		g_print ("Mono requires /proc to be mounted.\n");
--- a/mono/mini/mini-runtime.c
+++ b/mono/mini/mini-runtime.c
@@ -4080,8 +4080,6 @@ mini_init (const char *filename, const char *runtime_version)
 
 	CHECKED_MONO_INIT ();
 
-	mono_threads_suspend_policy_init ();
-
 #if defined(__linux__)
 	if (access ("/proc/self/maps", F_OK) != 0) {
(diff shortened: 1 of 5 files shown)


mono/mini/mini-runtime.c
```diff
@@ -4080,6 +4080,8 @@ mini_init (const char *filename, const char *runtime_version)
 
 	CHECKED_MONO_INIT ();
 
+	mono_threads_suspend_policy_init ();
+
```
vargaz: Shouldn't this be called from like mono_thread_info_init () ?

lambdageek: Yea makes sense. I'll move it.

mono/utils/mono-threads-coop.c
```diff
@@ -703,7 +702,7 @@ mono_threads_suspend_validate_policy (MonoThreadsSuspendPolicy policy)
 void
 mono_threads_suspend_override_policy (MonoThreadsSuspendPolicy new_policy)
 {
-	threads_suspend_policy = (char)mono_threads_suspend_validate_policy (new_policy);
+	mono_threads_suspend_policy_hidden_dont_modify = (char)mono_threads_suspend_validate_policy (new_policy);
```
jaykrell: Avoid contractions as too informal? `do_not` instead of `dont`.

jaykrell: There are other names for this in practise also, like "mostly readonly". abbreviated imho obscurely as "mr".
Often times such data is actually made read only after initialization with VirtualProtect. Though there are considerations for that, i.e. isolating it to a page with also mostly readonly.

jaykrell: You can also play useful games like declaring it as const but implementating it as non const.
With externC. i.e. to enforce not writing to it, except where allowed.

lambdageek:
@monojenkins build

PR #16151, opened 2019-08-09
[sdk] Enable the SRE runtime code inside iOS builds

spouliot:
This will allow the interpreter to use System.Reflection.Emit.

Note: There is a small size increase (about 50kb) for an application
built in release mode with the LLVM backend (even if the interpreter
is now used).

### DIFF
--- a/sdks/builds/ios.mk
+++ b/sdks/builds/ios.mk
@@ -99,7 +99,7 @@ _ios-$(1)_CONFIGURE_FLAGS = \
 	--enable-dtrace=no \
 	--enable-icall-export \
 	--enable-maintainer-mode \
-	--enable-minimal=ssa,com,interpreter,jit,reflection_emit_save,reflection_emit,portability,assembly_remapping,attach,verifier,full_messages,appdomains,security,sgen_remset,sgen_marksweep_par,sgen_marksweep_fixed,sgen_marksweep_fixed_par,sgen_copying,logging,remoting,shared_perfcounters,gac \
+	--enable-minimal=ssa,com,interpreter,jit,portability,assembly_remapping,attach,verifier,full_messages,appdomains,security,sgen_remset,sgen_marksweep_par,sgen_marksweep_fixed,sgen_marksweep_fixed_par,sgen_copying,logging,remoting,shared_perfcounters,gac \
 	--enable-monotouch \
 	--with-lazy-gc-thread-creation=yes \
 	--with-tls=pthread \


spouliot:
c.c. @lewurm @chamons

sdks/builds/ios.mk
```diff
@@ -99,7 +99,7 @@ _ios-$(1)_CONFIGURE_FLAGS = \
 	--enable-dtrace=no \
 	--enable-icall-export \
 	--enable-maintainer-mode \
-	--enable-minimal=ssa,com,interpreter,jit,reflection_emit_save,reflection_emit,portability,assembly_remapping,attach,verifier,full_messages,appdomains,security,sgen_remset,sgen_marksweep_par,sgen_marksweep_fixed,sgen_marksweep_fixed_par,sgen_copying,logging,remoting,shared_perfcounters,gac \
+	--enable-minimal=ssa,com,interpreter,jit,portability,assembly_remapping,attach,verifier,full_messages,appdomains,security,sgen_remset,sgen_marksweep_par,sgen_marksweep_fixed,sgen_marksweep_fixed_par,sgen_copying,logging,remoting,shared_perfcounters,gac \
```
vargaz: reflection_emit_save should still be disabled imho.

spouliot: the old preview (and sample code) supported it

vargaz: How was it supported if it was disabled before ? Also, what is the use case for supporting sre save on ios ?

spouliot: We had a preview last fall for the interpreter which had SRE (including saving) enabled. Our sample shows this worked.

The main use case is compatibility with netstandard code that is not aware that SRE could work without saving.

vargaz: Wouldn't be better to just map AssemblyBuilder.Save to AssemblyBuilder.Run in that case ?

spouliot: Better ? no. Less worse ? maybe :)

We're doing this (interpreter, SRE...) to increase compatibility between .net runtimes. Why go half way and impose a different set of limitations ? 

Can you explain the problem(s) where remove this sre-save would solve ?

vargaz: Well, sre.save is for saving assemblies to disk, i.e. AssemblyBuilder.Save (filename), but afaik ios has no file system, so this functionality won't work.

spouliot: > Well, sre.save is for saving assemblies to disk, i.e. AssemblyBuilder.Save (filename), 

yup, exactly

> but afaik ios has no file system, so this functionality won't work.

hmm.. wrong. iOS has a file system and, like I said above, that's a feature that worked in the old preview.

You might be confusing that you can't write executable code (well you can but it cwon't execute). However since it's interpreted that restriction does not exists - it's simply data like any text/binary file an app would write to / read from.

PR #16156, opened 2019-08-09
[w32process-win32] Don't duplicate CreateProcess search logic.

madewokherd:
When given a null application name, CreateProcess already
searches the path and appends .exe in some cases. .NET's
Process class behaves in the same way (see
https://github.com/mono/mono/pull/16089#issuecomment-519984244).

The duplicated logic is unnecessary and broken in several ways.
Failures are ignored and the command line arguments are
executed. Relative paths are expected not to end in "exe".
Many corner cases work differently from .NET and CreateProcess.

This change makes it impossible for process_complete_path to fail,
so I removed the gboolean return and pulled out the output
parameter. Its caller wasn't handling the return value correctly
anyway.

### DIFF
--- a/mono/metadata/w32process-win32.c
+++ b/mono/metadata/w32process-win32.c
@@ -198,48 +198,22 @@ process_quote_path (const gchar *path)
 }
 
 /* Only used when UseShellExecute is false */
-static gboolean
-process_complete_path (const gunichar2 *appname, gchar **completed)
+static gchar*
+process_complete_path (const gunichar2 *appname)
 {
 	// FIXME This function should stick to gunichar2.
 
 	char *utf8app;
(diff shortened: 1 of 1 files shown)


akoeplinger:
@madewokherd it would be great if you could still send a PR which fixes the G_FILE_TEST_IS_EXECUTABLE strncmp comparison in g_file_test() even though it's no longer used :)

jaykrell:
Or maybe remove that code, no longer used?

madewokherd:
Well, the strcmp isn't a useful thing to do anyway. A proper fix for G_FILE_TEST_IS_EXECUTABLE would probably check for file existence or execute permission. Something analogous to what it does on Linux.

PR #16157, opened 2019-08-10
Split interpreter into recursive vs. non-recursive.

jaykrell:
This should greatly reduce frame size in recursive cases,
and even speed things up due to cache locality against a smaller stack.
And allow deeper recursion.
Still some analysis and cleanup to do.

This diff involves moving code around and is completely unreadable, even by me.
As such, the PR needs to be tested, and discussed, and then *something* done.
Split into multiple commits perhaps, or multiple PRS.
Including gross options like duplicating the function entirely but wrapping one instance in #if 0.


vargaz:
I think an easier solution would be to merge all the local variables into InterpFrame, i.e. the
			int ret_size = * (guint16 *)(ip + 1);
			unsigned char *ret_vt_sp = vt_sp;
etc. ones.

jaykrell:
So far, this isn’t helping. The recursive frame is much larger than InterpFrame. 

The compiler does a good-not-perfect job of enregistering/overlapping the non recursive variables. I’ll look more though as there still seems to be more than 100 extra bytes.

I’m reluctant to add to InterpFrame as that guaranteeably grows frame size.

The approach here is “easy” in some sense, but impossible to read the diff.

vargaz:

jaykrell:
Maybe: sp and ip and probably vt_sp are hard to eliminate anyway, so I guess can be in InterpFrame. ip is already there anyway. clause_args could be. or we could duplicate the interpreter w/ and w/o clause_args, i.e. implied null and not null. The not null one can be in InterpFrame maybe.

jaykrell:
The non-recursive cases might not really
 contribute to the frame, I now realize.
At least with gcc.
I need to look at clang a bit.
We probably care little about msvc.

jaykrell:
I think I can make this work, i.e. not crash, but geez the compiler wants to introduce 100 bytes of temporaries to implement just about anything. I'll take another crack at it now that master has most of other changes or after one other round of such.

jaykrell:
I will try this again, a third time. I previously hadn’t studied the code enough.

A  recursive frame size of maybe 0x88 bytes is looking likely, plus the nonvolatile pushes.

PR #16158, opened 2019-08-10
Cleanup alloca: call twice instead of once, and shrink allocation slightly.

jaykrell:
It is an actual function call at least on Windows.

Extracted from larger PR that reduced auto cast.


jaykrell:
@monojenkins build failed

PR #16160, opened 2019-08-10
[wasm] Xunit testing work.

vargaz:


vargaz:
@monojenkins build failed

PR #16161, opened 2019-08-10
[corlib] Improve file:// url handling in AppDomainSetup.

nsivov:
WiX 3 installer toolkit uses file:/// prefix when setting ApplicationBase for newly created domain.
Currently this leads to invalid path exception:

file:///C:\... -> /C:\... -> full path expansion with current workdir on different drive -> Y:\C:\...

Later it fails check for multiple ':' -> NotSupportedException ("The given path's format is not supported.").

This patch uses logic already present in AppDomain::Load().

### DIFF
--- a/mcs/class/corlib/System/AppDomainSetup.cs
+++ b/mcs/class/corlib/System/AppDomainSetup.cs
@@ -143,7 +143,7 @@ namespace System
 			if (appBase == null)
 				return null;
 
-			if (appBase.ToLower (CultureInfo.InvariantCulture).StartsWith ("file://")) {
+			if (appBase.StartsWith ("file://", StringComparison.OrdinalIgnoreCase)) {
 				appBase = new Mono.Security.Uri (appBase).LocalPath;
 				if (Path.DirectorySeparatorChar != '/')
 					appBase = appBase.Replace ('/', Path.DirectorySeparatorChar);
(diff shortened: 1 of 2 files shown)


steveisok:
@monojenkins build failed

steveisok:
@nsivov Thanks for the PR!  It would be great if we could have a test cover this scenario as well.  

@akoeplinger, any suggestions on where the best place for that would be?

nsivov:
@steveisok sure, I'll look into it. There is AppDomainSetupTest.cs with some cases already. My question is what's the proper way to run this test on Windows?

nsivov:
@steveisok Push a simple test case for that.

mcs/class/corlib/System/AppDomainSetup.cs
```diff
@@ -142,9 +143,8 @@ static string GetAppBase (string appBase)
 			if (appBase == null)
 				return null;
 
-			int len = appBase.Length;
-			if (len >= 8 && appBase.ToLower ().StartsWith ("file://")) {
-				appBase = appBase.Substring (7);
+			if (appBase.ToLower (CultureInfo.InvariantCulture).StartsWith ("file://")) {
```
akoeplinger: This would be better:
```suggestion
			if (appBase.StartsWith ("file://", StringComparison.OrdinalIgnoreCase)) {
```

nsivov:
@akoeplinger thanks, I'll take a look at that test failure.

nsivov:
@akoeplinger I pushed some fixes, please take a look

PR #16163, opened 2019-08-11
[System.Drawing] Use CoreFX implementation of BufferedGraphics on Windows

nsivov:
The issue was made visible with .NET Core WinForms. Some controls draw text with DrawTextEx to intermediate Graphics based on Bitmap, that later it's moved back to main Graphics with DrawImage. This results in a lost of transparency component along the way, making antialiased glyph bitmaps draw and B/W. Corruption is visible with gray antialiasing, but is more severe in subpixel modes.

I verified that same happens with gdiplus on Windows, code from CoreFX works correctly.


akoeplinger:
@monojenkins commit csproj

PR #16165, opened 2019-08-11
[interp] Reduce frame by 16 bytes usually by outlining NEWOBJ.

jaykrell:
Tested with Linux/amd64/gcc.
A previous almost identical rendition saved 32 but it is lost for now.


jaykrell:
Likely the next 16 byte savings can be had by uninlining LEAVE but let's do that separately and after this.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Closing in derefence to larger savings.

PR #16168, opened 2019-08-11
[interp] Reduce frame by 16 bytes.

jaykrell:
At least with gcc/Linux/amd64.
Not with msvc, but probably if we used __declspec (noinline).
Clang not checked.


jaykrell:
@monojenkins build failed

jaykrell:
Closing in derefence to larger savings.

PR #16169, opened 2019-08-11
[interp] Remove unnecessary loops, that optimizer might handle.

jaykrell:


jaykrell:
@monojenkins build failed

PR #16170, opened 2019-08-11
[draft][interp]

jaykrell:
Doing this gradually now to keep it working and start reducing frame size.

Split interpreter into recursive vs. non-recursive.
This should greatly reduce frame size in recursive cases,
and even speed things up due to cache locality against a smaller stack.
And allow deeper recursion.
Still some analysis and cleanup to do.


jaykrell:
The reduction we can expect here, if I can get it to work, on Linux/gcc/amd64, is around 72 bytes.
Probably similar on all 64bit platforms and half on 32bit. And more with MSVC since it seems to not stack-pack as well.
Beyond appriximately that will get more difficult -- need to look into reducing InterpFrame after that.
For example maybe computing values cheaply instead of storing them.
Some operations, i.e. call, will get slower.
All will gain performance because using less stack leads to performance due to locality, TLB hits.

I also wonder if there is another approach involving a more manual maintenance of stack.
Basically alloca, but keeping a higher water mark and only alloca when exceeding it.
For "direct" recursions, of interp calling interp.
Not for "indirect" where interp calls out to invoke and such.

That should provide the same savings.

jaykrell:
This might not be profitable anyway, since calls have the most locals.

PR #16171, opened 2019-08-11
[interp] Reduce frame by 48 bytes.

jaykrell:
- Outline NEWOBJ (https://github.com/mono/mono/pull/16165).
- Avoid address-taking two locals in HasFlag (https://github.com/mono/mono/pull/16168).
- Remove an unnecessary loop while moving code (https://github.com/mono/mono/pull/16169).
- Outline LEAVE.


jaykrell:

jaykrell:
I can save another 32 easily, closing this in deference to that.

jaykrell:
@monojenkins build failed

jaykrell:
1. Savings beyond this is proving difficult.
2. The recursive/non-recursive split is looking less interesting, as generally the calls use the most stack for their locals. Outlining the "larger" calls might help the smaller calls. i.e. the dynamic + pinvoke case is the least efficient.

jaykrell:
@monojenkins build failed

jaykrell:
This brings clang/amd64/Linux from 0x178 to 0x128, saving 80 bytes.
I'll see about lowering that more.

jaykrell:
Clang/amd64/Linux now from 0x178 to 0x118, so the same 96 byte savings as gcc, but lagging a little.

jaykrell:
The opcodes that are outlined, could instead be only partly outlined, to what is needed to conserve stack. Arguably that'd be nicer. I have such a change I'm trying out.
Though that change is more ad-hoc and the lengths goes to to save stack is much more fragile.

BrzVlad:
@jaykrell I have mixed feelings about some of the changes in the PR so I think it would be better to split it. At least the bigger changes (like outlining). I have concerns about performance for some opcodes if we do the entire outlining since it adds indirection (with a lot of arguments). I can run some benchmarks there (although the outlined doesn't look hot, if I didn't miss any change). Also it doesn't look super nice, I would be down for partial outlining if possible.

I think we should have a more specific plan. Overall, for starters, I think we should decide how many locals should a switch case use. I'd say about 4 stack slots is a pretty good tradeoff (it wouldn't make sense to go through a bunch of trouble, refactoring and indirection, compiler hacks to save 8-16 bytes on the stack, out of 2-300). So this would mean making sure no switch case uses more than that and make sure all the locals that are used are properly reused by the compiler in other switch cases (maybe the compiler is stupid here and there). For cases where we can't achieve this we can do some partial outlining (like NEWOBJ which is pretty big) or other tweaks to save some space. We should make the **i32**, **o** and **c** local to every switch case where they are used. This should enable us from the start to free up 3 slots for reusing in other places. Maybe just that, would make some of the changes in this PR redundant. And for opcodes that need more stack space, we can figure out something from case to case, preferably in separate PRs.

jaykrell:
Allocating/counting locals manually yourself is very tricky.
Because the compilers are quite good.

What takes up stack is:
 - address-taken locals -- which is what I was mostly going after, at least initially, and you can still see here.
 - If you don't say gcc -fconserve-stack, common sub expressions, like sp [-1].
 - alloca -- notice how outlining is probably an optimization for the dynamic/pinvoke case as I changed a malloc to an alloca.
 - and then, yes, locals, if you store them, then make a function call, and then read them, which is sometimes but not always; so in many cases you can recompute them after a function call, cheaply; or in some cases just do the one read later; both occur here

gcc and clang are pretty good I have found.
Not clear about msvc, I didn't look much.

I can split this up, but there is overlap once you have `CHECK_RESUME_STATE` et. al. in function form.

Saving 8 or 16 bytes at a time is how I got over 100.

jaykrell:
https://github.com/mono/mono/pull/16190

jaykrell:
https://github.com/mono/mono/pull/16196

jaykrell:
@monojenkins build failed

jaykrell:
CI looking good! But splitting up and maybe revising.

jaykrell:
Much of this is in master. Not all of it. Master has more savings than this PR though.
In particular some of the "refetch" is missing.
This drove frame size to 0xD8.
Master I think is it 0xB8, and close to getting to 0x98.

PR #16173, opened 2019-08-12
[remoting+interp] [remoting+debugger] Fix use of dangling local.

jaykrell:


jaykrell:
This costs 8 or 16 bytes of stack. A much larger conflicting PR will be presented that avoids that. We should maybe take this first, depending on timing.

jaykrell:
@monojenkins build failed

Issue #16179, opened 2019-08-12
ves_icall_TypeBuilder_create_runtime_class_raw USED n handles (continued)


stakx:
AppVeyor recently upgraded their Ubuntu image to use Mono 6.0.0.313 instead of 5.20.1.19. Since that upgrade, I am seeing repeated messages of the form _"ves_icall_TypeBuilder_create_runtime_class_raw USED n handles"_ in the CI build log; see e.g. https://ci.appveyor.com/project/castleproject/core/builds/26641027/job/37hyn438mke9aoct#L115.

(Yes, that project *is* using `TypeBuilder` to create types; the issue is that this debug message shouldn't be printed to the console no matter what.)

In issue #15247, this was reported as fixed in `master`. Has there been a regression?

## Steps to Reproduce

1. Check out https://github.com/castleproject/Core.
2. Build the project on a machine with Mono 6 and the .NET CLI using the build script (`./build.sh`).

Alternatively, check out the build log linked to above.

### Current Behavior

Lots of messages of the form _"ves_icall_TypeBuilder_create_runtime_class_raw USED <n> handles"_ are printed to the console.

### Expected Behavior

These debug messages are too verbose, they shouldn't be printed to the console unless some verbosity level is explicitly raised to some debug/diagnostic level.

## On which platforms did you notice this

[ ] macOS
[X] Linux: AppVeyor for Linux (Ubuntu 16.04 image)
[ ] Windows

**Version Used**:

```
Mono JIT compiler version 6.0.0.313 (tarball Sun Jul 14 09:59:19 UTC 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

lambdageek:
There are two work items for mono here:
1. Only print the warning if verbose tracing is turned on.
2. Identify why `ves_icall_TypeBuilder_create_runtime_class_raw` is using so many handles and either (a) decide that the warning limit needs to be raised, or (b) fix TypeBuilder to use fewer handles.

stakx:
Thanks for looking into this so promptly.

> Identify why `ves_icall_TypeBuilder_create_runtime_class_raw` is using so many handles

[...]

PR #16181, opened 2019-08-12
[coop] Reduce verbosity if too many handles used

lambdageek:
Instead of issuing a warning, only show the message if the trace level is low
enough.

Addresses part of https://github.com/mono/mono/issues/16179

### DIFF
--- a/mono/metadata/handle.c
+++ b/mono/metadata/handle.c
@@ -19,6 +19,7 @@
 #include <mono/metadata/gc-internals.h>
 #include <mono/utils/atomic.h>
 #include <mono/utils/mono-lazy-init.h>
+#include <mono/utils/mono-logger-internals.h>
 #include <mono/utils/mono-threads.h>
 #ifdef HAVE_BACKTRACE_SYMBOLS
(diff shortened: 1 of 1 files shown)


stakx:
Thanks for looking into this so promptly.

I suppose I don't know enough about Ref.Emit on Mono, just FYI that project's unit test suite fires up a lot of `AssemblyBuilder`s and `TypeBuilder`s (about 1 of the former and >1 of the latter per unit test, of which there are around 1,000.)

jaykrell:
This has always been a fixable bug.
And it already goes unnoticed.
I suggest increasing its visibility not decreasing it.
At least in CI and non-release builds.
Is there a plan for that?

stakx:
I think I agree with @jaykrell: If this is fixable, then that might be preferable to suppressing a warning that's there for good reason. If properly fixed, this present proposed change here might not be necessary.

lambdageek:
Closing in favor of #16184 which should fix one instance of the handles warning that motivated this PR.  We should revisit if we see more warnings that are less straightforward to fix properly.

PR #16184, opened 2019-08-12
[sre] Outline typebuilder_setup_fields loop body

lambdageek:
The loop body uses coop handles, so wrap it in HANDLE_FUNCTION_ENTER/HANDLE_FUNCTION_RETURN

Addresses part of #16179


lambdageek:

PR #16190, opened 2019-08-13
[interp] [work in progress] Manage error init/swallow/reuse properly.

jaykrell:
I suggest `mono_error_cleanup` do the reinit and this not be an easy to make mistake.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -4519,6 +4519,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 						THROW_EX (mono_error_convert_to_exception (error), ip);
 					frame_objref (frame) = mono_object_new_checked (imethod->domain, newobj_class, error);
 					mono_error_cleanup (error); /* FIXME: don't swallow the error */
+					error_init_reuse (error);
 					EXCEPTION_CHECKPOINT;
 					sp->data.o = frame_objref (frame);
 #ifndef DISABLE_REMOTING
@@ -4621,6 +4622,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 					/* slow path */
 					isinst = mono_object_isinst_checked (o, c, error) != NULL;
 					mono_error_cleanup (error); /* FIXME: don't swallow the error */
+					error_init_reuse (error);
 				} else {
 					isinst = FALSE;
 				}
@@ -4659,6 +4661,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 			if ((o = sp [-1].data.o)) {
 				MonoObject *isinst_obj = mono_object_isinst_checked (o, c, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
 				if (!isinst_obj) {
 					if (isinst_instr)
 						sp [-1].data.p = NULL;
@@ -4779,6 +4782,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 
 				addr = (char*)mono_load_remote_field_checked (o, klass, field, &tmp, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
 			} else
 #endif
 				addr = (char*)o + field->offset;
@@ -4805,6 +4809,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 				MonoClass *klass = ((MonoTransparentProxy*)o)->remote_class->proxy_class;
 				addr = (char*)mono_load_remote_field_checked (o, klass, field, &tmp, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
 			} else
 #endif
 				addr = (char*)o + field->offset;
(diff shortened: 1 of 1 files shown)


mono/mini/interp/interp.c
```diff
@@ -4519,6 +4519,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 						THROW_EX (mono_error_convert_to_exception (error), ip);
 					frame_objref (frame) = mono_object_new_checked (imethod->domain, newobj_class, error);
 					mono_error_cleanup (error); /* FIXME: don't swallow the error */
+					error_init_reuse (error);
```
lambdageek: IMO the right approach is to use a local `MonoError` for the two calls in this branch that need to ignore the error.

The lifecycle of a `MonoError` should be simple.  `error_init_reuse` is a code smell.

mono/mini/interp/interp.c
```diff
@@ -4621,6 +4622,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 					/* slow path */
 					isinst = mono_object_isinst_checked (o, c, error) != NULL;
 					mono_error_cleanup (error); /* FIXME: don't swallow the error */
+					error_init_reuse (error);
```
lambdageek: Same here. could use a local `MonoError`

mono/mini/interp/interp.c
```diff
@@ -4659,6 +4661,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 			if ((o = sp [-1].data.o)) {
 				MonoObject *isinst_obj = mono_object_isinst_checked (o, c, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
```
lambdageek: Same

mono/mini/interp/interp.c
```diff
@@ -4779,6 +4782,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 
 				addr = (char*)mono_load_remote_field_checked (o, klass, field, &tmp, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
```
lambdageek: Same

mono/mini/interp/interp.c
```diff
@@ -4805,6 +4809,7 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 				MonoClass *klass = ((MonoTransparentProxy*)o)->remote_class->proxy_class;
 				addr = (char*)mono_load_remote_field_checked (o, klass, field, &tmp, error);
 				mono_error_cleanup (error); /* FIXME: don't swallow the error */
+				error_init_reuse (error);
```
lambdageek: etc

mono/mini/interp/interp.c
jaykrell: We are super concerned with stack usage here.
This is an intense loop and reuse is appropriate.

mono/mini/interp/interp.c
jaykrell: I'm *reducing* stack use here. Many MonoError have already been removed. 100 more bytes will be removing. Adding more MonoError would go in the wrong direction.

jaykrell: "here" meaning in other PRs. This PR does not reduce stack.

lambdageek: There's already a function call here (`mono_object_isinst_checked`).  Could be `mono_object_isinst_nofail` that sets up a local MonoError.

reusing MonoError is a last resort sort of thing.  There's no reason to do it here.

mono/mini/interp/interp.c
lambdageek: The interpreter doesn't recurse here.  But it does make a call to another function already.  It's perfectly reasonable to call a `_nofail` version instead of `_checked` one here.

mono/mini/interp/interp.c
jaykrell: Ignoring errors isn't the right fix either imho. Reusing variables is a normal idiom..
```
s = strup("foo");
free(s)
s = strdup("bar");
```

I know there is functional programming that avoids it, but I've seen very little of that in mono or anywhere and I think for easy efficiency we should be willing to reuse.

mono/mini/interp/interp.c
jaykrell: Some of these are in `interp_exec_method_full`, some are not.
`ERROR_DECL` must not be used in `interp_exec_method_full` (not even in rare/slow paths), but it can be used in non-inlined functions that it calls.
So yeah we'd have to introduce a few helper functions.
I suggest `mono_interp_foo` and not `mono_foo_nofail`, as these are also anti-patterns -- ignoring errors, don't want to bless it.
I still think reuse is ok, but non-reuse should be viable here.

jaykrell: I can do it.

jaykrell: I still think reuse is the way.
The interpreter went out of its way already to reuse, just was slightly off.
The interpreter I think is unusual in its quest for performance since it really wants to be a JIT or AOT.
I worry that exiting to runtime& recursing
back to interp will double error use, though maybe never deeply?
Contructors bottom out quickly?

BrzVlad:
this looks outdated

PR #16191, opened 2019-08-13
Llvm optimizations

vargaz:

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -7779,6 +7779,10 @@ call_end:
 			if (cmethod)
 				ins = handle_call_res_devirt (cfg, cmethod, ins);
 
+			if (noreturn) {
+				MONO_INST_NEW (cfg, ins, OP_NOT_REACHED);
+				MONO_ADD_INS (cfg->cbb, ins);
+			}
 calli_end:
 			if ((tailcall_remove_ret || (common_call && tailcall)) && !cfg->llvm_only) {
 				link_bblock (cfg, cfg->cbb, end_bblock);
@@ -7864,11 +7868,6 @@ calli_end:
 						emitted_funccall_seq_point = TRUE;
 				}
 				emit_seq_point (cfg, method, next_ip, FALSE, TRUE);
-			}
-
-			if (noreturn) {
-				MONO_INST_NEW (cfg, ins, OP_NOT_REACHED);
-				MONO_ADD_INS (cfg->cbb, ins);
 			}
 			break;
 		}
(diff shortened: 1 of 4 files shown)


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

mono/mini/method-to-ir.c
```diff
@@ -3759,6 +3759,15 @@ mono_emit_load_got_addr (MonoCompile *cfg)
 	MONO_ADD_INS (cfg->bb_exit, dummy_use);
 }
 
+static gboolean
+method_does_not_return (MonoMethod *method)
```
marek-safar: Could we use better logic than method name? I guess checking for no `ret` should be sufficient for what we need? 

Note: [DoesNotReturn] is C# compiler flow-analysis attribute and we cannot rely on it tools like linker will try to remove it.

vargaz: Checking for no ret would involve parsing the IL for every method we want to inline, so if DoesNotReturn is not usable, then this hack would probably need to stay.

vargaz:
@monojenkins build failed

mono/mini/method-to-ir.c
EgorBo: Maybe it should be implemented in Roslyn (detect and mark such methods with DoesNotReturn automatically?)

marek-safar: That was discussed and rejected by Roslyn folks

vargaz: This is just an internal class, if something changes, it will be pretty easy to detect the breakage. Can also check that it throws when we compile the called method.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

jaykrell:
Relevant failure?

https://jenkins.mono-project.com/job/test-mono-pull-request-amd64-fullaot+llvm/15454/parsed_console/log_content.html

```
block_guard_restore_aligment_on_exit.exe                                                                        failed, time: 00:00.686, exit code: 4

Time: 00:19.693

 519 test(s) passed
   1 test(s) failed
   0 test(s) timed out

Failed test(s):

block_guard_restore_aligment_on_exit.exe
=============== block_guard_restore_aligment_on_exit.exe.stdout ===============
EEE
What now?

=============== block_guard_restore_aligment_on_exit.exe.stderr ===============

Makefile:2869: recipe for target 'runtest-managed' failed
make[5]: Leaving directory '/mnt/jenkins/workspace/test-mono-pull-request-amd64-fullaot+llvm/mono/tests'
make[5]: *** [runtest-managed] Error 1
```

jaykrell:
Or the other failed lane? I don't kno.

lewurm:
Jay is right, this PR introduced the failure on `block_guard_restore_aligment_on_exit.exe` with Linux/AMD64 FullAOT+LLVM. I'll look into it.

PR #16193, opened 2019-08-13
[interp] Replace `mono_error_ok` with `is_ok`.

jaykrell:
The locally unknown function, despite being small/fast,
forces saving values in stack and nonvolatile registers around calling it.

Both nonvolatile registers and stack are at a premium here, whereas usually stack is cheap.
The macro is just a read from offset 0 and checking for zero.

The entire runtime should be repaired in this way.


jaykrell:
Part of https://github.com/mono/mono/pull/16171..

Issue #16172, opened 2019-08-11
Interpreter stack size is too large


vargaz:
Currently, on amd64+osx, the interpeter main function stack size is 392 bytes:
```otool -tv mono-sgen | grep -10 _interp_exec_method_full:```

Thats in addition to the dynamically allocate stack+locals etc. InterpFrame itself is 96 bytes, the rest is local variables etc. This is probably not good for performance due to cache locality etc.

On 32 bit webassembly, its 400 bytes, plus about 27 wasm locals. There are reports of stack
overflows on some browsers.

jaykrell:
Linux/gcc/amd64 is better, 312 and I have a PR to get it to 264.
However a goal of around 96 would be good.

GerritBergen:
@vargaz is this what might be causing this issue? #15989

jaykrell:
I can get the 392 down to 216 (0xD8) on Linux/amd64/gcc and clang, and pass CI.
Maybe significantly lower, we'll see, I have ideas, but I'd like to circa 0xD8 first, as the other ideas might fail.

jaykrell:
Should be below 256 now. Might be able to save another 100+.

jaykrell:
Could the stack of 8 byte 

unions be transformed into actual 4 & 8 byte locals, like the JIT does?

jaykrell:
master/Linux/amd64/gcc down to 0x88.
There is a chance for 0x78.

SamMonoRT:
@BrzVlad - please follow up on this, if more fixes are needed.

BrzVlad:
I'd say we reached the limit on the improvements with regard to stack size with our current execution mode.

We could continue at some point with non recursive design for further gains. That is a more complex change, that would have perf improvement as the main driver, rather than stack usage. Personally, I'd close this unless the webassembly team explicitly requests lower stack usage. I think some investigations are pending there. cc @lewing

SamMonoRT:
@lewing - is there any web assembly investigation pending to keep this issue open ?

SamMonoRT:
Assigning to @lewing - please close/re-assign appropriately.

kjpou1:
verified via: https://github.com/mono/mono/issues/17938 and closed by: https://github.com/mono/mono/pull/17953

jaykrell:
https://github.com/mono/mono/pull/18756

jaykrell:
https://github.com/mono/mono/pull/18756
https://github.com/mono/mono/pull/18766
[...]

PR #16194, opened 2019-08-13
[interp] Reduce frame size by move/recompute locals.

jaykrell:
Compute or recompute locals closer to their use, so they don't have to survive function calls.

Baby steps from https://github.com/mono/mono/pull/16171.

Ultimately over 100 bytes are savable but it requires some change.


jaykrell:
@monojenkins build failed

jaykrell:
Close at least for now.

PR #16196, opened 2019-08-13
[interp] Reduce frame size by partially outlining load remote field.

jaykrell:
Extracted from https://github.com/mono/mono/pull/16171.
Ultimately over 100 bytes can be saved.
This change on its own probably saves 16 bytes, but over 100 bytes are savable.

Also fixes the error reuse problem nearby but not overall (https://github.com/mono/mono/pull/16190).

Outlining is partial to avoid "trouble" around:
 vt_sp update
 ip update
 exception handling

But those are more elegantly shown in other PR.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -2888,6 +2888,71 @@ static int opcode_counts[512];
 		} \
 	} while (0);
 
+static
+#ifndef DISABLE_REMOTING
+MONO_NEVER_INLINE // To reduce stack.
+#endif
+void
+mono_interp_load_remote_field (
+	InterpMethod* imethod,
+	MonoObject* o,
+	const guint16* ip,
+	stackval* sp,
+	MonoError* error)
+{
+	g_assert (o); // Caller checks and throws exception properly.
+
+	void* addr;
+	MonoClassField* const field = (MonoClassField*)imethod->data_items[* (guint16 *)(ip + 1)];
+
+#ifndef DISABLE_REMOTING
+	gpointer tmp;
+	if (mono_object_is_transparent_proxy (o)) {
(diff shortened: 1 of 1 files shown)


mono/mini/interp/interp.c
```diff
@@ -2888,6 +2888,72 @@ static int opcode_counts[512];
 		} \
 	} while (0);
 
+static
+#ifndef DISABLE_REMOTING
```
BrzVlad: Why guarding here ? Let's never inline this method.

jaykrell: It only has the address-taken local under this ifndef.
Otherwise it inlines reasonably and contributes nothing to the caller's stack usage.
That is the reason. If it is too confusing, or if you are concerned about the inlined code bytes polluting cache, I can see that. But for stack usage, when there is no remoting,..

mono/mini/interp/interp.c
```diff
@@ -2888,6 +2888,72 @@ static int opcode_counts[512];
...
+	if (mono_object_is_transparent_proxy (o)) {
+		klass = ((MonoTransparentProxy*)o)->remote_class->proxy_class;
+		addr = mono_load_remote_field_checked (o, klass, field, &tmp, error);
+		mono_error_cleanup (error); /* FIXME: don't swallow the error */
+		error_init_reuse (error);
+	} else
+#endif
+		addr = (char*)o + field->offset;
+	sp [-1].data.p = vt_sp;
+	vt_sp = vt_sp + ALIGN_TO (i32, MINT_VT_ALIGNMENT);
```
jaykrell: Accidentally changed `+=`. Code movement hard to read. :( Will fix. Just cosmetic.

jaykrell:
@monojenkins build failed

PR #16198, opened 2019-08-13
[interp] Outline `get_virtual_method_fast` to save 16 bytes of stack.

jaykrell:

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -584,7 +584,7 @@ alloc_method_table (MonoVTable *vtable, int offset)
 	return table;
 }
 
-static InterpMethod*
+static MONO_NEVER_INLINE InterpMethod* // Inlining causes additional stack use in caller.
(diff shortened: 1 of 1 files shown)


mono/mini/interp/interp.c
```diff
@@ -584,7 +584,9 @@ alloc_method_table (MonoVTable *vtable, int offset)
 	return table;
 }
 
-static InterpMethod*
+static
+MONO_NEVER_INLINE // Inlining causes additional stack use in caller.
```
BrzVlad: I'd really love it if we had these on the same line.

jaykrell: And where is the comment? Above, nowhere (an ok answer), further to the right?
I like the function name at start so I can search for defs and decls, but not calls, with regexp ^foo.
If only we were consistent. :(
`static` and return type, either way.

jaykrell: I admit disappointment in using outlining here. :(

jaykrell:
@monojenkins build failed

PR #16200, opened 2019-08-13
[interp] Outline `set_resume_state` to conserve stack in main interpreter loop.

jaykrell:
This is not intuitive but I definitely saw it.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -229,7 +229,7 @@ int mono_interp_traceopt = 0;
 #define MINT_IN_DEFAULT default:
 #endif
 
-static void
+static MONO_NEVER_INLINE void // Inlining this causes caller to use more stack.
 set_resume_state (ThreadContext *context, InterpFrame *frame)
 {
 	/* We have thrown an exception from a finally block. Some of the leave targets were unwinded already */


mono/mini/interp/interp.c
```diff
@@ -229,7 +229,9 @@ int mono_interp_traceopt = 0;
 #define MINT_IN_DEFAULT default:
 #endif
 
-static void
+static
+MONO_NEVER_INLINE // Inling this causes caller to use more stack.
```
BrzVlad: single line if possible

jaykrell:
```
static MONO_NEVER_INLINE void
foo
```
ok, I can search for ^foo.


```
static MONO_NEVER_INLINE void foo
```
less ok but still ok

jaykrell:
@monojenkins build failed

mono/mini/interp/interp.c
```diff
@@ -229,7 +229,7 @@ int mono_interp_traceopt = 0;
 #define MINT_IN_DEFAULT default:
 #endif
 
-static void
+static MONO_NEVER_INLINE void // Inlining this causes caller to use more stack.
 set_resume_state (ThreadContext *context, InterpFrame *frame)
```
vargaz: The comment is kinda not neccessary, we do it for the same reason for all functions.

jaykrell: It isn't obvious here since there are no locals.
I guess it increases register pressure in caller.

jaykrell: But ok either way, I can remove it.

PR #16202, opened 2019-08-13
[interp] Reduce stack usage, such as by `endsp` removal.

jaykrell:
In particular by removing the `endsp` locals that the compiler
struggles to preserve across entire handlers without using stack.
What it is holding can be efficiently recomputed instead.

This is just one a few steps.


jaykrell:
@monojenkins build failed

jaykrell:
Does not currently help, will redo maybe soon.

PR #16208, opened 2019-08-13
[debugger] Debugger tries to execute a single step when the thread is already terminated

thaystg:
I couldn't create a small test case that reproduces, I reproduced debugging (Xamarin.Mac) ios-sim-sharp.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/942093

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -7266,6 +7266,14 @@ event_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
 				return err;
 			}
 
+			DebuggerTlsData *tls = (DebuggerTlsData *)mono_g_hash_table_lookup (thread_to_tls, THREAD_TO_INTERNAL(step_thread));
+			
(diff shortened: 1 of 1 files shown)


mono/mini/debugger-agent.c
```diff
@@ -7266,6 +7266,14 @@ event_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
 				return err;
 			}
 
+			DebuggerTlsData *tls = (DebuggerTlsData *)mono_g_hash_table_lookup (thread_to_tls, THREAD_TO_INTERNAL(step_thread));
+			
+			if (tls->terminated) {
```
lambdageek: Can `tls` be null here?

thaystg: Yes, you are right, I changed how to get and the inserted the assert like is done in other places.

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

PR #16213, opened 2019-08-13
[interp] Split leave into pieces, partially outlined, to conserve stack,

jaykrell:
but without changing how exception handling macros access labels and locals.

This saves 16 bytes of stack on Linux/amd64/gcc and likely similar everywhere,
as it takes an address-taken local out of interp_exec_method_full (stackval tmp_sp).

See
https://github.com/mono/mono/pull/16171
https://github.com/mono/mono/pull/16171/commits/a6a8f0d5ffa875da9cda63444e401c37810a690f for a
full outlining.

This is just one step toward that.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -891,17 +891,17 @@ stackval_to_data_addr (MonoType *type_, stackval *val)
  *   Throw an exception from the interpreter.
  */
 static MONO_NEVER_INLINE void
-interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, gconstpointer ip, gboolean rethrow)
+interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, const guint16* ip, gboolean rethrow)
 {
 	ERROR_DECL (error);
 	MonoLMFExt ext;
 
 	interp_push_lmf (&ext, frame);
-	frame->ip = (const guint16*)ip;
+	frame->ip = ip;
 	frame->ex = ex;
 
 	if (mono_object_isinst_checked ((MonoObject *) ex, mono_defaults.exception_class, error)) {
-		MonoException *mono_ex = (MonoException *) ex;
+		MonoException *mono_ex = ex;
 		if (!rethrow) {
 			mono_ex->stack_trace = NULL;
 			mono_ex->trace_ips = NULL;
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

mono/mini/interp/interp.c
```diff
@@ -2907,13 +2907,29 @@ static int opcode_counts[512];
 #define frame_objref(frame) o
 #endif
 
+// Leave is split into pieces in order to consume less stack,
+// but not have to change how exception handling macros access labels and locals.
+static MONO_NEVER_INLINE MonoException*
+mono_interp_leave (InterpFrame* child_frame)
```
BrzVlad: Maybe name this mono_interp_get_undeniable_exception or something like that. The current name is not really representative of what this method is doing.

jaykrell: To naïve/shallow/empirical me, it is `mono_interp_outlined_mechanical_part_of_leave_whatever_it_takes`, shortened to `mono_interp_leave`, not anything more general or more specific, not be used anywhere else. But of course I try to be flexible. :)

jaykrell: `mono_interp_leave_outlined`??
There might be another chunk, I was going to call them `mono_interp_leave1` and `mono_interp_leave2`.
I need to double check that it useful for purposes of stack.

jaykrell: I see what you mean though, it is one line doing just that.
Nitty question, then..
```					child_frame.imethod = NULL; 
					child_frame.imethod = NULL;
```
should be moved into this function, for it to be more complete/sensible per your name?
For stack preservation they can go in either place. I only moved the minimal amount, didn't attempt to make a sensible function.

mono/mini/interp/interp.c
```diff
@@ -2907,13 +2907,29 @@ static int opcode_counts[512];
...
+{
+	stackval tmp_sp;
+	/*
+	 * We need for mono_thread_get_undeniable_exception to be able to unwind
+	 * to check the abort threshold. For this to work we use child_frame as a
+	 * dummy frame that is stored in the lmf and serves as the transition frame
+	 */
+	do_icall_wrapper (child_frame, NULL, MINT_ICALL_V_P, &tmp_sp, (gpointer)mono_thread_get_undeniable_exception, FALSE);
+
+	return (MonoException*)tmp_sp.data.p;
```
jaykrell: Feels like `p` should be `o` for wasm GC-safety. But I didn't change that.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Oops

jaykrell:
@monojenkins build failed

PR #16215, opened 2019-08-13
[interp] Remove address-taking of vt_sp.

jaykrell:
*Theoretically* *could* help reduce frame size but it has not been seen to.

Frame has been gradually reduced over 100 bytes to 0xD8 bytes but this change is *not* clearly important for that, and *might* be contradicted by other changes that take address of vt_sp.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -2354,10 +2354,9 @@ do_transform_method (InterpFrame *frame, ThreadContext *context)
 		interp_pop_lmf (&ext);
 }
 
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

BrzVlad:
looks cleaner anyway

jaykrell:
@monojenkins build failed

PR #16217, opened 2019-08-14
Completely replace mono_error_ok with is_ok and make first external_only.

jaykrell:
Unless some things fail then dial it back a little.


jaykrell:
@monojenkins build failed

PR #16218, opened 2019-08-14
Reduce size of internal MonoErrors by 3 pointers, while trying to be compatible with old size.

jaykrell:
This is compatible enough?
We don't hand off MonoError externally, that anyone can/should use the size of?
As long as embedders give us a large enough one, and we retain the size for any that they embed in their own data structures?

### DIFF
--- a/mono/metadata/object-offsets.h
+++ b/mono/metadata/object-offsets.h
@@ -136,7 +136,7 @@ DECL_OFFSET(MonoProfilerCallContext, method)
 DECL_OFFSET(MonoProfilerCallContext, return_value)
 DECL_OFFSET(MonoProfilerCallContext, args)
 
-DECL_OFFSET(MonoError, init)
+DECL_OFFSET(MonoErrorExternal, init)
 
 #ifdef HAVE_SGEN_GC
 DECL_OFFSET(SgenClientThreadInfo, in_critical_region)
--- a/mono/utils/mono-error.h
+++ b/mono/utils/mono-error.h
@@ -65,10 +65,16 @@ typedef union _MonoError {
 		uint16_t private_flags; /*DON'T TOUCH */
 		void *hidden_1 [12]; /*DON'T TOUCH */
 	};
-} MonoError;
+} MonoErrorExternal;
 
 #ifdef _MSC_VER
 __pragma(warning (pop))
+#endif
+
+#ifdef MONO_INSIDE_RUNTIME
+typedef union _MonoErrorInternal MonoError;
+#else
+typedef MonoErrorExternal MonoError;
 #endif
 
 /* Mempool-allocated MonoError.*/
--- a/mono/mini/decompose.c
+++ b/mono/mini/decompose.c
@@ -1252,7 +1252,7 @@ mono_decompose_vtype_opts (MonoCompile *cfg)
 
 					if (m_class_get_image (ins->klass) == mono_defaults.corlib && !strcmp (m_class_get_name (ins->klass), "MonoError")) {
 						// Used in icall wrappers, optimize initialization
-						MONO_EMIT_NEW_STORE_MEMBASE_IMM (cfg, OP_STOREI4_MEMBASE_IMM, dest->dreg, MONO_STRUCT_OFFSET (MonoError, init), 0);
+						MONO_EMIT_NEW_STORE_MEMBASE_IMM (cfg, OP_STOREI4_MEMBASE_IMM, dest->dreg, MONO_STRUCT_OFFSET (MonoErrorExternal, init), 0);
 					} else {
 						mini_emit_initobj (cfg, dest, NULL, ins->klass);
 					}
--- a/mono/unit-tests/test-mono-callspec.c
+++ b/mono/unit-tests/test-mono-callspec.c
@@ -5,6 +5,19 @@
  *
  */
 
+// Embedders do not have the luxury of our config.h, so skip it here.
+//#include "config.h"
+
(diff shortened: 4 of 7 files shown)


jaykrell:
Will probably split this up & rework it slightly but it isn’t awful.

jaykrell:
@monojenkins build failed

jaykrell:

mono/utils/mono-error.h
```diff
@@ -50,27 +50,36 @@ enum {
...
 /*Keep in sync with MonoErrorInternal*/
-typedef union _MonoError {
+union _MonoErrorExternal {
 	// Merge two uint16 into one uint32 so it can be initialized
 	// with one instruction instead of two.
 	uint32_t init;
 	struct {
 		uint16_t error_code;
-		uint16_t private_flags; /*DON'T TOUCH */
+		uint16_t hidden_0; /*DON'T TOUCH */
```
jaykrell: This is how it used to be. I changed it. I don't remember why. I'll check git blame.

jaykrell: No reason obvious in the PR. Maybe to see better in debugger?

mono/metadata/class-internals.h
```diff
@@ -1257,7 +1257,7 @@ mono_class_get_checked (MonoImage *image, guint32 type_token, MonoError *error);
 MonoClass *
 mono_class_get_and_inflate_typespec_checked (MonoImage *image, guint32 type_token, MonoGenericContext *context, MonoError *error);
 
-MonoClass *
+G_EXTERN_C MonoClass* // FIXME? G_EXTERN_C due to MonoError varying
```
lambdageek: Confused why _this_ function has to be `G_EXTERN_C` but none of the others functions in this header do too.

mono/metadata/object-offsets.h
```diff
@@ -136,7 +136,8 @@ DECL_OFFSET(MonoProfilerCallContext, method)
 DECL_OFFSET(MonoProfilerCallContext, return_value)
 DECL_OFFSET(MonoProfilerCallContext, args)
 
-DECL_OFFSET(MonoError, init)
+//FIXME
+//DECL_OFFSET(MonoError, init)
```
lambdageek: Could be `DECL_OFFSET(MonoErrorInternal, init)` and then `MONO_STRUCT_OFFSET (MonoErrorInternal, init)` at the use site in `decompose.c`, right?

mono/metadata/class-internals.h
jaykrell: I not sure -- "because there was an error in CI and I was lazy about it"?
Maybe "because profiler not C++"? I can undo it and get crispness on it.
I think it was driver and dis though, which I think are C++, so that does not hold.
I can revisit, before or after this PR, as you wish.

mono/metadata/object-offsets.h
jaykrell: Again I was being lazy. I will try that.
I'd actually like to remove this, but I have some inexplicable wasm-only error:
https://github.com/mono/mono/pull/15894

jaykrell:
We'll see if it builds though.

jaykrell:
Oh right it was tests not being C++.
I had converted them before but it was rejected.
I'll PR that.

```
.libs/libtestlib_la-test-mono-callspec.o: In function `test_mono_class_from_name':
/mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mono/unit-tests/test-mono-callspec.c:148: undefined reference to `mono_class_from_name_checked(_MonoImage*, char const*, char const*, _MonoErrorExternal*)'
/mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mono/unit-tests/test-mono-callspec.c:148: undefined reference to `mono_class_from_name_checked(_MonoImage*, char const*, char const*, _MonoErrorExternal*)'
/mnt/jenkins/workspace/test-mono-pull-request-amd64-cxx/mono/unit-tests/test-mono-callspec.c:148: undefined reference to `mono_class_from_name_checked(_MonoImage*, char const*, char const*, _MonoErrorExternal*)'
/mn
```

mono/metadata/class-internals.h
jaykrell: Tests do not include config.h.
So they don't have ifdef inside_runtime.
And they test external_only, which including config.h would break (deprecation error).

Solutions:

1. extern "C" as I had, but with comments explaining.

*or*

2. include config.h and muck with external_only

*or*

3. Still do not include config.h, define inside_runtime, muck with external_only

*or*

4. Another define to guide the internals, like, `MONO_ALLOW_EXTERNAL_ONLY`.

I pushed #3.

jaykrell:
@monojenkins build failed

mono/metadata/class-internals.h
jaykrell: https://github.com/mono/mono/pull/16218/commits/23a75a132bc2dd2b9642eb30f70f4fc80fabeeb2

jaykrell:
@vargaz quick tips for offsets-tool?

mono/utils/mono-error.h
```diff
@@ -56,7 +56,7 @@ __pragma(warning (disable:4201))
 #endif
 
 /*Keep in sync with MonoErrorInternal*/
-typedef union _MonoError {
+typedef union MonoErrorExternal {
```
jaykrell: Do we care about C++ name mangling compatibility?
If so put this back to `_MonoError`.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16221, opened 2019-08-14
Cleanup mono_error_assert_msg_ok, based on g_assertf.

jaykrell:


jaykrell:
Looks right but does not work.

jaykrell:
The problem is that layering over `g_assertf` leads to the double comma, that msvc does not eliminate.
You'd have to not build on `g_assertf` I guess. gcc/clang handle it.

PR #16226, opened 2019-08-14
[netcore] Ignore failing StackTraces tests

EgorBo:
See https://github.com/mono/mono/pull/16191/files#r313441821

(Here is the custom throw helper which is inlined: https://github.com/dotnet/corefx/blob/de38804d52f6b65f0f290b81383f01e6943a6d8f/src/System.Diagnostics.StackTrace/tests/StackTraceTests.cs#L326 but should not)

### DIFF
--- a/netcore/CoreFX.issues.rsp
+++ b/netcore/CoreFX.issues.rsp
@@ -713,6 +713,10 @@
 # https://github.com/mono/mono/issues/15187
 -nomethod System.Diagnostics.Tests.StackFrameTests.Ctor_SkipFrames_FNeedFileInfo
 
+# JIT should not inline custom throw helpers
+-nomethod System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames
+-nomethod System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames_FNeedFileInfo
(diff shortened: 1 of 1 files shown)


netcore/CoreFX.issues.rsp
```diff
@@ -713,6 +713,10 @@
 # https://github.com/mono/mono/issues/15187
 -nomethod System.Diagnostics.Tests.StackFrameTests.Ctor_SkipFrames_FNeedFileInfo
 
+# JIT should not inline custom throw helpers
+-nomethod System.Diagnostics.Tests.StackTraceTests.Ctor_Exception_SkipFrames
```
marek-safar: What about fixing the tests with `System.Runtime.CompilerServices.MethodImplOptions.NoInlining` ?

EgorBo: @marek-safar ok, will add this change to my next set of `dotnet/corefx` changes for tests but can we merge this to keep the CI green?

CoffeeFlux:
Is this only disabled temporarily or should we open an issue to track it?

EgorBo:
@CoffeeFlux I've already applied the workaround in my dotnet/corefx PR locally

PR #16230, opened 2019-08-14
[netcore] Mark "Run Tests" step as failed if there are failures

EgorBo:
Currently "Run CoreFX tests" step is always green


EgorBo:
Example: https://dev.azure.com/dnceng/public/_build/results?buildId=311329

PR #16231, opened 2019-08-14
[interp] Compute locals later to reduce register/stack pressure.

jaykrell:


mono/mini/interp/interp.c
```diff
@@ -5279,15 +5283,16 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
-			MonoClass *klass = (MonoClass*)imethod->data_items [*(guint16 *) (ip + 1)];
 			guint16 numargs = *(guint16 *) (ip + 2);
 			ip += 3;
 			sp -= numargs;
 
 			o = sp [0].data.o;
 			NULL_CHECK (o);
+
+			MonoClass *klass = (MonoClass*)imethod->data_items [*(guint16 *) (ip - 3 + 1)];
+			const gboolean needs_typecheck = ip [-3] == MINT_LDELEMA_TC;
```
jaykrell: For example these should not matter because they are only after a null check -- if there is a function call, it get cannot get here. I'm not sure the compiler was being that good.

mono/mini/interp/interp.c
```diff
@@ -5231,20 +5233,22 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
-			gsize offset_length = (gsize) *(gint16 *) (ip + 2);
-			gsize offset_pointer = (gsize) *(gint16 *) (ip + 3);
+			guint8 * const span = (guint8 *) sp [-2].data.p;
+			const int index = sp [-1].data.i;
 			sp--;
 
 			NULL_CHECK (span);
 
-			gint32 length = *(gint32 *) (span + offset_length);
+			const gsize offset_length = (gsize) *(gint16 *) (ip + 2);
```
jaykrell: Another example that might not matter -- the only intervening code is a null check.

jaykrell:
@monojenkins build failed

PR #16242, opened 2019-08-14
OP_LLVM_OUTARG_VT doesn't always create temporary vt needed by some ABI's.

lateralusX:
LLVM lowering passes vt as reference to managed methods. On platforms where the value type ABI is passing an address to the value type, either in register or on stack (Windows x64) we must make sure we always use a temporary of the value type or we will be breaking the value type semantics.

In JIT lowering of OP_OUTARG_VT we always make sure we use a temporary in cases where argument storage is set to ArgValuetypeAddrInIReg/ArgValuetypeAddrOnStack. On LLVM lowering this was currently not the case, but it still worked in many scenarios. This was because IL load of arguments before doing a call did a local copy and then we pass that byref. However, there are cases where this copy could get optimized away and when that happens, we will change the
local value type in the caller’s frame, breaking the value type semantics.

Added test case in winx64structs hits such a scenario that will generate IL breaking the value type semantics before the fix:

...
notail call monocc void @winx64_vector3Struct_Add_winx64_vector3Struct_winx64_vector3Struct
(i64 %49, %winx64_vector3Struct* %3, %winx64_vector3Struct* %16)
...
notail call monocc void @winx64_vector3Struct_Add_winx64_vector3Struct_winx64_vector3Struct
(i64 %64, %winx64_vector3Struct* %3, %winx64_vector3Struct* %15)
...

The two add calls will use the same reference (to a local value type in caller frame) on both calls, this will break the value type semantics since they should both be temporaries, like second argument is.

After the fix, the same code will be lowered into:

...
notail call monocc void @winx64_vector3Struct_Add_winx64_vector3Struct_winx64_vector3Struct
(i64 %55, %winx64_vector3Struct* %19, %winx64_vector3Struct* %18)
...
notail call monocc void @winx64_vector3Struct_Add_winx64_vector3Struct_winx64_vector3Struct
(i64 %72, %winx64_vector3Struct* %16, %winx64_vector3Struct* %15)
...

Each argument gets its own temporary that, according to the ABI, gets passed by address down to function.

There will be situations where we could end up with two temporaries, one from the load and one from the lowering of OP_LLVM_OUTARG_VT (same as we get on JIT), this will however be detected by LLVM optimization pass, removing one of the copies.

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -1569,6 +1569,7 @@ sig_to_llvm_sig_full (EmitContext *ctx, MonoMethodSignature *sig, LLVMCallInfo *
 				param_types [pindex] = LLVMArrayType (IntPtrType (), ainfo->nslots);
 			pindex ++;
 			break;
+		case LLVMArgVtypeAddr:
 		case LLVMArgVtypeByRef:
 			param_types [pindex] = type_to_llvm_arg_type (ctx, ainfo->type);
 			if (!ctx_ok (ctx))
@@ -3414,6 +3415,7 @@ emit_entry_bb (EmitContext *ctx, LLVMBuilderRef builder)
 			}
 			break;
 		}
+		case LLVMArgVtypeAddr:
 		case LLVMArgVtypeByRef: {
 			/* The argument is passed by ref */
 			ctx->addresses [reg] = LLVMGetParam (ctx->lmethod, pindex);
@@ -3922,6 +3924,7 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
 			g_assert (addresses [reg]);
 			args [pindex] = addresses [reg];
 			break;
+		case LLVMArgVtypeAddr :
 		case LLVMArgVtypeByRef: {
 			g_assert (addresses [reg]);
 			args [pindex] = convert (ctx, addresses [reg], LLVMPointerType (type_to_llvm_arg_type (ctx, ainfo->type), 0));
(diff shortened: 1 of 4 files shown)


lateralusX:
@monojenkins rebuild failed

mono/mini/mini-llvm.c
```diff
@@ -6335,8 +6338,8 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 					g_assert (values [ins->sreg1]);
 					LLVMBuildStore (builder, convert (ctx, values [ins->sreg1], type_to_llvm_type (ctx, t)), addresses [ins->sreg1]);
 					addresses [ins->dreg] = addresses [ins->sreg1];
-				} else if (values [ins->sreg1] == addresses [ins->sreg1]) {
-					/* LLVMArgVtypeByRef, have to make a copy */
+				} else if (ainfo->storage == LLVMArgVtypeAddr || values [ins->sreg1] == addresses [ins->sreg1]) {
```
imhameed: `LLVMArgVtypeAddr` is only ever used on amd64 and mostly serves the same role as `LLVMArgVtypeByRef`, except for this explicit check and for its use in `process_call`. Is there any harm in dropping `LLVMArgVtypeAddr`, and adding an test here for `LLVMArgVtypeByRef` instead? This would then also cover arm, armv7k, arm64, which also have native ABIs that sometimes pass aggregates via an indirection:

https://github.com/mono/mono/blob/2d06126dc97d09749dae7722fc1a8da44f60e129/mono/mini/mini-arm.c#L1516-L1527

https://github.com/mono/mono/blob/2d06126dc97d09749dae7722fc1a8da44f60e129/mono/mini/mini-arm.c#L2372-L2375

https://github.com/mono/mono/blob/2d06126dc97d09749dae7722fc1a8da44f60e129/mono/mini/mini-arm64.c#L1200-L1204

https://github.com/mono/mono/blob/2d06126dc97d09749dae7722fc1a8da44f60e129/mono/mini/mini-arm64.c#L2435-L2438

lateralusX: Tried to avoid hitting cases where OP_LLVM_OUTARG_VT is used but not asking for a valuetype copy and pass by reference of the copy (due to ABI) semantics, so the idea was to make a distinction between the two scenarios all way up to the place where we initially identify that this is coming from the use of ArgValuetypeAddrInIReg/ArgValuetypeAddrOnStack. The other alternative is to add something to the arguments passed along with LLVMArgVtypeByRef, but I don't think that is as clear as having an explicit storage type, the same as we have on mono regular JIT AOT. If the same behavior is needed on other CPU architectures (and that also suffers from the same issue), I believe it is better to use the new storage type for that specific mapping and leave the behavior of ArgVtypeByRef as is (unless it sole purpose was to model this specific case, triggering a valuetype copy and pass that byref according to ABI and not just pass a valuetype byref) . The only place where this bug has been identified so far is on Windows x64 (amd64) and that's where the current fix is applied. If other platforms might be a victim for the same issue, we could leverage the same fix there and make the test triggering the issue more generic.

lateralusX: @imhameed ^^^^?

lateralusX: @imhameed, thanks! Then I guess this PR is good to go, possible for you to approve it?

Issue #16032, opened 2019-08-05
IOException: Sharing violation on path ...


andreysavihin:
Hello!
I got exception in my code when i try to copy files using FileStream with FileShare parameter.

## Steps to Reproduce

1. create Program.cs file

```
using System;
using System.IO;

namespace FileCopy
{
    class Program
    {
        static void Main(string[] args)
        {
            try
            {
                File.WriteAllText("source.txt", "text");
                var fileStream = new FileStream("source.txt", FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite, 2048, FileOptions.DeleteOnClose);
                File.Copy(fileStream.Name, "target.txt", true);
            }
            catch (Exception e)
            {
                File.WriteAllText("error.txt", e.Message + Environment.NewLine + e.StackTrace);
            }
        }
    }
}
```

2. run build command

`mcs -out:Test.exe Program.cs`

3. execute 

`mono Test.exe`

### Current Behavior

got file error.txt with error & stack trace 
IOException: Sharing violation on path .../source.txt

### Expected Behavior

got file target.txt with "text" content.
file source.txt will be removed, because used FileOptions.DeleteOnClose parameter

[ ] macOS
[x] Linux
[ ] Windows

andreysavihin:
Mono JIT compiler version 6.0.0.313

tobiasschulz:
We are encountering this issue with several customers now. (Xamarin.Android 9.4, Mono 2019-02)

Is there any workaround? How can something like this break?

steveisok:
The first thing I did was look to see if we have a test for this.  *Seems* that we do not.  

I'll take a closer look and let you know.

steveisok:
And I can at least reproduce the same exception in regular mono, so this may be easier to track down.

steveisok:
So it appears before 6.0, File.Copy was more or less a direct line to MonoIO.CopyFile:

https://github.com/mono/mono/blob/3a07bd426d3dae0d9b6c7891e7bb11f487b18484/mcs/class/corlib/System.IO/File.cs#L72-L111

In 6.0, we switched more to CoreFX, which tries to create two FileStreams in order to pass handles into the file copy:

https://github.com/mono/corefx/blob/dbfa00cdb2f6e8423f805f9abffe94665cd5c22f/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs#L41-L59

FileStream in mono then throws an exception when it hits this line:

[...]

PR #16247, opened 2019-08-14
[corlib] Fix sharing mode validation rules 

steveisok:
Fixes https://github.com/mono/mono/issues/16032

When we switched to corefx File.Copy on unix, the behavior changed where it opened a source and destination stream in order to pass the safe handles to Interop.Sys.CopyFile.  This exposed a flaw in our sharing validation rules in MonoIO.Open, which threw an exception in this scenario:

```
new FileStream("source.txt", FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite, 2048, FileOptions.DeleteOnClose);
new FileStream("source.txt", FileMode.Open, FileAccess.Read, FileShare.Read, 2048, FileOptions.None);
```

This should now correctly line up w/  https://docs.microsoft.com/en-us/dotnet/api/system.io.fileshare?view=netframework-4.8#fields


mcs/class/corlib/Test/System.IO/FileStreamTest.cs
```diff
@@ -1650,6 +1650,30 @@ public void DeleteOnClose ()
...
+			File.WriteAllText (source, "text");
+			FileStream fs = new FileStream (source, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite, 2048,
+							FileOptions.DeleteOnClose);
+
+			Assert.AreEqual (true, File.Exists (source), "DCFC#1");
+
+			try {
+				File.Copy (fs.Name, target, true);
+			}
+			finally {
```
lambdageek: nit: `} finally {`

mono/metadata/w32file-unix.c
```diff
@@ -1874,16 +1874,15 @@ static gboolean share_allows_open (struct stat *statbuf, guint32 sharemode,
...
 		}
 
-		if (((file_existing_access & GENERIC_READ) &&
-		     !(sharemode & FILE_SHARE_READ)) ||
-		    ((file_existing_access & GENERIC_WRITE) &&
-		     !(sharemode & FILE_SHARE_WRITE))) {
-			/* New share mode doesn't match up */
-			mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_FILE, "%s: Access mode prevents open: requested share: 0x%" PRIx32 ", file has access: 0x%" PRIx32, __func__, sharemode, file_existing_access);
+		if (file_existing_share != sharemode &&
+				((sharemode | FILE_SHARE_READ) != sharemode) &&
```
lambdageek: Why do we need this?  It's checking that the new requested sharing mode `sharemode` doesn't include `FileShare.Read`.  Why?

Isn't it enough to just check that `file_existing_share != sharemode` (new is different from existing) and `(file_existing_share | sharemode) == file_existing_share` (new it at most as lax as existing - it doesn't include).  That together means that new `sharemode` is a strict subset of the existing `file_existing_share`.  And therefore we take the branch and return FALSE.

tobiasschulz:
How soon can we get this into xamarin.android stable? It is broken since VS release 16.2 on production.

mono/metadata/w32file-unix.c
jaykrell: This change/code confuses me.
Why do the share modes have to line up, as long as they grant each other access?
Don't sharemodes get anded upon successful open, to weigh against later opens?
I might try it myself if that not too rude.

like:
  - open readwrite, share readwrite
  - open read, share read; success, resulting share is read 
  - open readwrite; fail
 
or:
    - open read, share read
    -  open read, share readwrite; success, resulting share is read
    - open readwrite; fail

Note this cannot reall workout underlying kernel support.
We end up presumably only enforcing it within the process, amongst mono users, which is a lot less than the original intent.

steveisok: @lambdageek From our conversation, the `((sharemode | FILE_SHARE_READ) != sharemode) &&` line is to close the gap w/ ReadWrite.  Without it, FIle.Copy would fail (ReadWrite to Read). 

The more I look at this and the corefx version, the more I think we should just follow in kind. 
 Which is, FileShare.None locks exclusively and that's about it.

https://github.com/dotnet/corefx/blob/b990ece3f8cd7999ad62483f1bd814ec586895de/src/Common/src/CoreLib/System/IO/FileStream.Unix.cs#L76-L88

jaykrell: Yeah, since it requires kernel/filesystem to really support, i.e. to play "correctly" with non-mono callers.
A coworker of mine years ago suggested this sharing stuff, while it seems like a good idea, isn't really all that great even on Windows and perhaps is pointless. Unix might have its flaws, but this missing feature you don't hear about much.

My simple attempt broke tests.
https://github.com/mono/mono/pull/16271

steveisok: The check is pretty simple now.

lambdageek:

lambdageek:

mono/metadata/w32file.h
```diff
@@ -262,6 +262,7 @@ mono_filesize_from_fd (int fd);
 #define GENERIC_EXECUTE 0x20000000
 #define GENERIC_ALL     0x10000000
 
+#define FILE_SHARE_NONE   0x00000000
```
jaykrell: This is not actually a symbol in Windows, but ok.

mono/metadata/w32file-unix.c
jaykrell: Need to "subscribe" to that file/line and get a notification if they decide to work on it.
Maybe the AI will notice the copied code and notify us, or update ours.

steveisok:
@monojenkins build failed

Issue #16248, opened 2019-08-14
mono_thread_set_name_internal needs to be optimized for multi-thread


fanyang-mono:
### Current Behavior

I profiled mono-netcore with AOT=LLVM on macOS with the following benchmark.
https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks

After discussing with the team, we found that mono_thread_set_name_internal needs to be optimized to improve the performance of Mono-netcore. We probably don't need to set a new name for every thread.

<img width="656" alt="Screen Shot 2019-08-14 at 3 47 35 PM" src="https://user-images.githubusercontent.com/52458914/63051370-202cc200-beab-11e9-9c79-93b85e74a760.png">

**Mono version used**:
36af44e3be4a0f73586e886b96f86ee500f8ab32

## Steps to Reproduce
Reproduce by following **Profiling workflow (Method 1)** from the following link:
https://paper.dropbox.com/doc/Profile-Mono-on-macOS--Aijv~7jvs5XNRvKUtb_7UzTYAg-ryHeobv8okzUs1Fa9TfBG

jaykrell:
How about just not setting it in a loop, but just once per thread?

jaykrell:
After https://github.com/mono/mono/pull/16250

introduce a local variable so that this:
```
C:\s\mono4\mono\metadata\threadpool.c(357):		mono_thread_set_name_internal (thread, thread_name, MonoSetThreadNameFlag_Reset, error);
```
is only once instead every time the loop runs.

Setting a thread name *once per thread* will be cheap. Setting it per threadpool work item, not so much. I suggested this long ago but was told it is per-spec or something -- threadpool is specified to be slow.

Really I suggest removing it from non-netcore and will PR that now.

jaykrell:
I also have an optimization in a PR that removes the extra copying and string conversion here.
I was waiting for other PRs to be commited first.

See here:
https://github.com/mono/mono/pull/15859


which was requested to be split up.
The next in line for the split has been sitting for weeks here:

https://github.com/mono/mono/pull/15859

After that we'd just retain the constant w/o copying, for threadpool.

lambdageek:
> Setting a thread name _once per thread_ will be cheap. Setting it per threadpool work item, not so much. I suggested this long ago but was told it is per-spec or something -- threadpool is specified to be slow.

[...]

PR #15945, opened 2019-07-31
Retain thread names as UTF8 instead of UTF16.

jaykrell:
Windows can still have its Unicode thread names, since we have UTF16 initially, just don't retain it.

This does penalize ves_icall_System_Threading_Thread_GetName_internal (non-netcore only),
which has to convert back to UTF16 each time.

But seems like a net win overall.

There is also a little prep work here for constant thread names and some temporary measures as multiple changes head toward the same code.

Rename `mono_thread_set_name_internal` to `mono_thread_set_name`.
Close possible race conditions.

Extracted from https://github.com/mono/mono/pull/15859.

PR #16252, opened 2019-08-14
Do not allocate and set thread name in loop.

jaykrell:
This was rejected a while ago.

There are some middle grounds.

In particular, we can retain utf8 instead of a managed string, no copy or conversion.
That is almost done, awaiting PR.

### DIFF
--- a/mono/metadata/threadpool.c
+++ b/mono/metadata/threadpool.c
@@ -325,6 +325,8 @@ worker_callback (void)
 
 	previous_tpdomain = NULL;
 
+	gboolean set_thread_name = TRUE;
+
 	while (!mono_runtime_is_shutting_down ()) {
 		gboolean retire = FALSE;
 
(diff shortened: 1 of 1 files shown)


jaykrell:
Substantially addresses https://github.com/mono/mono/issues/16248.

vargaz:
Yes, the thread name needs to be reset, there are tests for it.

jaykrell:
Understood.
I wouldn't do a full get though, maybe a counter incremented every time it is changed.

vargaz:
That would be fine.

mono/metadata/threadpool.c
```diff
@@ -352,10 +354,13 @@ worker_callback (void)
 
 		domains_unlock ();
 
-		MonoString *thread_name = mono_string_new_checked (mono_get_root_domain (), "Thread Pool Worker", error);
-		mono_error_assert_ok (error);
-		mono_thread_set_name_internal (thread, thread_name, MonoSetThreadNameFlag_Reset, error);
-		mono_error_assert_ok (error);
+		if (set_thread_name) {
+			MonoString *thread_name = mono_string_new_checked (mono_get_root_domain (), "Thread Pool Worker", error);
```
marek-safar: Could we also avoid allocation this? Maybe introduce a new MonoSetThreadNameFlag_Init flag which only sets thread flags and uses static UTF data for the native name?

jaykrell: Yes. I have that in another PR already.

lambdageek: I think @jaykrell wanted `MonoSetThreadNameFlag_Constant` to do precisely that.  not sure if one of his open PRs already does it.

jaykrell: https://github.com/mono/mono/pull/15945
and then the rest of this:
https://github.com/mono/mono/pull/15859 which was requested to be split up and is almost done.

jaykrell:
Closed in deference to https://github.com/mono/mono/pull/16253.

PR #16253, opened 2019-08-14
Reduce redundant setting of thread name.

jaykrell:
Maintain a counter and only set if the counter has changed.
This cannot be fully thread safe, so is not.
Any thread can set any other thread name at any time.
You could be atomic to do better, and loop to do even better,
but the thread name could still be changed again right after the loop exists.

The previous code was also not reliable in this way, since again the name
could be changed right away.

https://github.com/mono/mono/issues/16248

### DIFF
--- a/mcs/class/corlib/System.Threading/Thread.cs
+++ b/mcs/class/corlib/System.Threading/Thread.cs
@@ -55,6 +55,7 @@ namespace System.Threading {
 		IntPtr native_handle; // used only on Win32
 		/* accessed only from unmanaged code */
 		private IntPtr name;
+		private IntPtr name_generation;
 		private int name_len; 
 		private ThreadState state;
 		private object abort_exc;
--- a/mono/metadata/object-internals.h
+++ b/mono/metadata/object-internals.h
@@ -574,6 +574,7 @@ struct _MonoInternalThread {
 	MonoThreadHandle *handle;
 	gpointer native_handle;
 	gunichar2  *name;
+	volatile gsize name_generation;
 	guint32	    name_len;
 	guint32	    state;      /* must be accessed while longlived->synch_cs is locked */
 	MonoException *abort_exc;
--- a/mono/metadata/threadpool.c
+++ b/mono/metadata/threadpool.c
@@ -325,6 +325,8 @@ worker_callback (void)
 
 	previous_tpdomain = NULL;
 
+	gsize name_generation = ~thread->name_generation;
+
 	while (!mono_runtime_is_shutting_down ()) {
 		gboolean retire = FALSE;
 
@@ -352,10 +354,17 @@ worker_callback (void)
 
 		domains_unlock ();
 
-		MonoString *thread_name = mono_string_new_checked (mono_get_root_domain (), "Thread Pool Worker", error);
-		mono_error_assert_ok (error);
-		mono_thread_set_name_internal (thread, thread_name, MonoSetThreadNameFlag_Reset, error);
-		mono_error_assert_ok (error);
+		// Any thread can set any other thread name at any time.
+		// So this is unavoidably racy.
+		// This only partly fights against that -- i.e. not atomic and not a loop.
+		// It is reliable against the thread setting its own name, and somewhat
+		// reliable against other threads setting this thread's name.
+		if (name_generation != thread->name_generation) {
+			MonoString *thread_name = mono_string_new_checked (mono_get_root_domain (), "Thread Pool Worker", error);
+			mono_error_assert_ok (error);
+			name_generation = mono_thread_set_name_internal (thread, thread_name, MonoSetThreadNameFlag_Reset, error);
+			mono_error_assert_ok (error);
+		}
 
 		mono_thread_clear_and_set_state (thread,
 			(MonoThreadState)~ThreadState_Background,
--- a/netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
+++ b/netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
@@ -22,6 +22,7 @@ namespace System.Threading
 		IntPtr native_handle; // used only on Win32
 		/* accessed only from unmanaged code */
 		private IntPtr name;
+		private IntPtr name_generation;
 		private int name_len;
 		private ThreadState state;
 		private object abort_exc;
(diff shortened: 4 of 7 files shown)


jaykrell:

jaykrell:

mono/metadata/threadpool.c
```diff
@@ -352,10 +354,17 @@ worker_callback (void)
...
-		MonoString *thread_name = mono_string_new_checked (mono_get_root_domain (), "Thread Pool Worker", error);
-		mono_error_assert_ok (error);
-		mono_thread_set_name_internal (thread, thread_name, MonoSetThreadNameFlag_Reset, error);
-		mono_error_assert_ok (error);
+		// Any thread can set any other thread name at any time.
+		// So this is unavoidably racy.
+		// This only partly fights against that -- i.e. not atomic and not a loop.
+		// It is reliable against the thread setting its own name, and somewhat
+		// reliable against other threads setting this thread's name.
+		if (name_generation != thread->name_generation) {
```
lambdageek: Can we check for `thread->flags & MONO_THREAD_FLAG_NAME_SET` ? Wouldn't that catch calls to `ves_icall_System_Threading_Thread_SetName_internal` ?

mono/metadata/object-internals.h
```diff
@@ -574,6 +574,7 @@ struct _MonoInternalThread {
 	MonoThreadHandle *handle;
 	gpointer native_handle;
 	gunichar2  *name;
+	gsize       name_generation;
```
jaykrell: Make this `volatile` once monolite finishes.

jaykrell: There are same-thread function calls generally in the way, of the compiler knowing it can cache, so *should* be ok either way. Some compilers over-pessimize volatile, at least in some modes (msvc arm modal).

mono/metadata/threadpool.c
jaykrell: Maybe. I never understood those flags.
This is also pretty simple and classic.

You are saying, the only intervening change can be from managed, and managed always leaves a flag behind, that native clears?

And then we'd need a boolean for the first call too, which I had, but tilde'ed away.

vargaz:
@monojenkins build failed

mcs/class/corlib/System.Threading/Thread.cs
```diff
@@ -55,6 +55,7 @@ sealed class InternalThread : CriticalFinalizerObject {
 		IntPtr native_handle; // used only on Win32
 		/* accessed only from unmanaged code */
 		private IntPtr name;
+		private IntPtr name_generation;
```
jaykrell: Also netcore/System.Private.CoreLib/src/System.Threading/Thread.cs after monolite finishes.

mono/metadata/threadpool.c
lambdageek: yea exactly that.

jaykrell: I'm on the fence. I found the flags confusing until now.
Maybe rename "set" to "managed"?
Hypothetically the counter is more general.
I also suggest, for a later commit, pointer or strcmp compare, but it might be moot.
But for example, if managed does a get and a set sometimes, and then we do another set, we could skip it. But I'm just making that up, and replacing managed with native constant is reasonable too.

mono/metadata/object-internals.h
```diff
@@ -574,6 +574,7 @@ struct _MonoInternalThread {
 	MonoThreadHandle *handle;
 	gpointer native_handle;
 	gunichar2  *name;
+	volatile gsize name_generation;
```
jaykrell: `volatile` ok? Unclear if needed. We do a read outside the lock.

netcore/System.Private.CoreLib/src/System.Threading/Thread.cs
```diff
@@ -22,6 +22,7 @@ partial class Thread
 		IntPtr native_handle; // used only on Win32
 		/* accessed only from unmanaged code */
 		private IntPtr name;
+		private IntPtr name_generation;
```
jaykrell: `int` would probably suffice. Cycle counters take a few seconds to hit 4GB, this would be 10 or 100x that.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build fialed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/metadata/threadpool.c
marek-safar: It looks to me we can optimize this even more so this path is basically never hit for ThreadPool worker threads which I'd expect to be the common case.

- set the native name in the thread-init only (no managed allocation) and do something like `thread->flags = MONO_THREAD_FLAG_NAME_TP_DEFAULT;`
- check here only for `thread->flags & MONO_THREAD_FLAG_NAME_SET` and use the current name (should not be racy).

jaykrell: I think this PR, plus my other two PRs, is plenty efficient, and allows the wonky existing behavior of arbitrary change and restore (which I'm all for removing also, but it has been declined twice).

Once the three PRs are in, the behavior will be:
  - store a string constant, no copy, no convert, no allocation, neither native nor managed  
  - do unlocked read of an integer and compare (volatile?) 
  - when they are equal in the overwhelming case, do nothing more 
  - when they are unequal, restore the name, as is apparently expected 

The use of the counter is very efficient and idiomatic. The bunch of flags I find a *little* more complicated and not really any more efficient, just saving a few bits in the Thread struct.

We could avoid the string altogether and just set a bit, but, we want native debuggers and ETW to see the string, so we have to pass it on somehow right away. We just don't actually have to store for GetName, which does not exist in netcore, but it is only a constant pointer, no copy/alloc/convert, so plenty cheap.

Netcore could possibly not both storing anything, except the counter.
And use the counter to gate the native call.
Since there is no GetName.
But this is not a huge difference really.
Not sure what to do about the flags though in netcore.

jaykrell: Probably keep the ability to get the name, as it is used internally.
Really, I don't know how to politely disagree/insist, but I think what I have is pretty good.
This stuff had been bugging me for a very long time, since I tried converting setname to coop, and I think we are on the cusp of it being pretty nice all around -- coop, Win32 names, constant names, threadpool loop, efficiency, etc.

Issue #16246, opened 2019-08-14
[netcore] Set up static ALC events


CoffeeFlux:
Tests:
- [x] Make System.Tests.AppDomainTests.TypeResolve pass (#18141)
- [x] Make System.Tests.AppDomainTests.ResourceResolve pass (https://github.com/dotnet/runtime/pull/36658)
- [x] Make System.Tests.AppDomainTests.AssemblyLoad pass (#18143)
- [x] Make System.Tests.AppDomainTests.FirstChanceException_Called pass (https://github.com/dotnet/runtime/pull/36658)
- [ ] Make System.Tests.AppDomainTests.AssemblyResolve_FirstChanceException pass (dependent on FirstChanceException)
- [x] Make System.Tests.AppDomainTests.AssemblyResolveInvalidAssemblyName pass (#18157)

Tasks (in flux):
- [x] Copy CoreCLR code into our ALC implementation, marking the region (#16256)
- [x] PR CoreCLR to move the code to shared (https://github.com/dotnet/coreclr/pull/26248)
- [x] Remove our copied implementation (#17767)
- [x] Wire up OnAssemblyResolve (see #15081)
- [x] Wire up OnAssemblyLoad (#18143)
- [ ] Only set RequestingAssembly when appropriate (see https://docs.microsoft.com/en-us/dotnet/api/system.resolveeventargs.requestingassembly?view=netcore-3.0#remarks)
- [x] Wire up OnTypeResolve (#18141)
- [x] Wire up OnResourceResolve (https://github.com/dotnet/runtime/pull/36658)
- [x] Wire up FirstChanceException (https://github.com/dotnet/runtime/pull/36658)
- [x] Remove old MonoDomain Do*Resolve methods and ensure nothing calls them from netcore (#18139, #18141, #18143, https://github.com/dotnet/runtime/pull/36658)

CoffeeFlux:
PRs are out now for the high priority events/tests here. The remaining ones, ResourceResolve and FirstChanceException, are some combination of more difficult/unlikely to be used/unlikely to be fully supported, so I'll get to them later.

PR #16256, opened 2019-08-14
[netcore] Enable netcore resolution logic for mono_assembly_byname_lookup

### LINKED ISSUE #15081: [netcore] Make System.Resources.Tests.ResourceManagerTests.GetString_ExpectEvents Pass

```
System.Resources.Tests.ResourceManagerTests.GetString_ExpectEvents [FAIL]
      Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.
      Stack Trace:
        
        Child exception:
          Xunit.Sdk.EqualException: Assert.Equal() Failure
        Expected: 2
        Actual:   0
           at System.Resources.Tests.ResourceManagerTests.Remote_ExpectEvents()
           at System.Resources.Tests.ResourceManagerTests.<>c.<GetString_ExpectEvents>b__10_0()
        
        Child process:
          System.Resources.ResourceManager.Tests, Version=4.1.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 System.Resources.Tests.ResourceManagerTests+<>c Void <GetString_ExpectEvents>b__10_0()
```

The thought is that netcore is missing assembly resolve events on AppDomain and AssemblyLoadContext.  Determine if that's the case and make the test pass.

CoffeeFlux:
Draft for now, as things are ~~probably~~definitely broken.

Addresses parts of #16246.
FIxes #15081.

Remaining todos:
- [x] Fix legacy behavior
- [ ] Figure out exact conditions under which the postload hook should be run
- [x] Rebase and resolve conflicts
- [x] Ensure the only caller of `netcore_load_reference` is `mono_assembly_request_byname`

### DIFF
--- a/mono/metadata/assembly.c
+++ b/mono/metadata/assembly.c
@@ -1611,10 +1611,8 @@ load_reference_by_aname_individual_asmctx (MonoAssemblyName *aname, MonoAssembly
 }
 #else
 static MonoAssembly*
-netcore_load_reference (MonoAssemblyName *aname, MonoAssemblyLoadContext *alc, MonoAssembly *requesting, MonoImageOpenStatus *status)
+netcore_load_reference (MonoAssemblyName *aname, MonoAssemblyLoadContext *alc, MonoAssembly *requesting, gboolean postload)
 {
-	g_assert (status != NULL);
-	g_assert (requesting != NULL);
 	g_assert (alc != NULL);
 
 	MonoAssemblyName mapped_aname;
@@ -1632,13 +1630,16 @@ netcore_load_reference (MonoAssemblyName *aname, MonoAssemblyLoadContext *alc, M
 	/*
 	 * Try these until one of them succeeds (by returning a non-NULL reference):
 	 * 1. Check if it's already loaded by the ALC.
+	 *
 	 * 2. If it's a non-default ALC, call the Load() method.
 	 *
-	 * 3. Try to load using the default ALC.
+	 * 3. Try to load using the default ALC (except for satellite requests).
 	 *
-	 * 4. Call ALC Resolving event.
+	 * 4. Call ALC ResolveSatelliteAssembly method (for satellite requests).
 	 *
-	 * 5. Return REFERNCE_MISSING
+	 * 5. Call ALC Resolving event.
+	 *
+	 * 6. Return NULL.
 	 */
 
 	reference = mono_assembly_loaded_internal (alc, aname, FALSE);
@@ -1654,7 +1655,7 @@ netcore_load_reference (MonoAssemblyName *aname, MonoAssemblyLoadContext *alc, M
 		MonoAssemblyByNameRequest req;
 		mono_assembly_request_prepare_byname (&req, MONO_ASMCTX_DEFAULT, mono_domain_default_alc (mono_alc_domain (alc)));
 		req.requesting_assembly = requesting;
-		reference = mono_assembly_request_byname (aname, &req, status);
+		reference = mono_assembly_request_byname (aname, &req, NULL);
 		if (reference)
 			goto leave;
 	}
@@ -1730,7 +1731,7 @@ mono_assembly_load_reference (MonoImage *image, int index)
 {
 	MonoAssembly *reference;
 	MonoAssemblyName aname;
-	MonoImageOpenStatus status;
+	MonoImageOpenStatus status = MONO_IMAGE_OK;
 
 	/*
 	 * image->references is shared between threads, so we need to access
@@ -1777,7 +1778,7 @@ mono_assembly_load_reference (MonoImage *image, int index)
 			break;
 		}
 #else
-		reference = netcore_load_reference (&aname, mono_image_get_alc (image), image->assembly, &status);
+		reference = netcore_load_reference (&aname, mono_image_get_alc (image), image->assembly, TRUE);
 #endif
 	} else {
 #ifndef ENABLE_NETCORE
@@ -4569,7 +4570,7 @@ mono_assembly_request_byname_nosearch (MonoAssemblyName *aname,
 				       const MonoAssemblyByNameRequest *req,
 				       MonoImageOpenStatus *status)
 {
-	MonoAssembly *result;
+	MonoAssembly *result = NULL;
 	MonoAssemblyName maped_aname;
 	MonoAssemblyName maped_name_pp;
 
@@ -4591,8 +4592,10 @@ mono_assembly_request_byname_nosearch (MonoAssemblyName *aname,
 		return result;
 	}
 
-	/* TODO: for netcore can we avoid calling this method?  There is no GAC and we also shouldn't need to load anything from the default path */
-	return mono_assembly_load_full_gac_base_default (aname, req->basedir, req->request.alc, req->request.asmctx, status);
+#ifndef ENABLE_NETCORE
+	result = mono_assembly_load_full_gac_base_default (aname, req->basedir, req->request.alc, req->request.asmctx, status);
+#endif
+	return result;
 }
 
 /* Like mono_assembly_request_byname_nosearch, but don't ask the preload look (ie,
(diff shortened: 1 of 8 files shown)


mono/metadata/assembly.c
```diff
@@ -4662,14 +4671,21 @@ mono_assembly_request_byname (MonoAssemblyName *aname, const MonoAssemblyByNameR
...
+	result = mono_assembly_request_byname_nosearch (aname, req, status);
 	const gboolean refonly = req->request.asmctx == MONO_ASMCTX_REFONLY;
 
 	if (!result && !req->no_postload_search) {
 		/* Try a postload search hook */
 		result = mono_assembly_invoke_search_hook_internal (req->request.alc, req->requesting_assembly, aname, refonly, TRUE);
 		result = prevent_reference_assembly_from_running (result, refonly);
 	}
+#else
+	result = netcore_load_reference (aname, req->request.alc, req->requesting_assembly, !req->no_postload_search);
```
lambdageek: What do you think about getting rid of `netcore_load_reference` entirely and replacing its other callers by a call to `mono_assembly_request_byname`?

CoffeeFlux: It's definitely an option. I'm not super happy with what the call stack looks like right now (since it's somewhat convoluted) but I also don't really want to require building a bunch more requests for simple calls within assembly.c, so I've been trying to think on how I want to organize things.

CoffeeFlux:
I've decided to hold off on System.Tests.AppDomainTests.AssemblyResolveInvalidAssemblyName until a later PR, since other work is being done to reflection to pass along an ALC. The core issue is that when calling `Type.GetType()`, we end up in `type_from_parsed_name()`, which does not set a requesting_assembly despite firing off the postload hook. It should call byname directly and make sure to set that as well as the ALC. This will maybe combine well with wiring up the TypeResolved event.

CoffeeFlux:
Rebased version on netcore is hitting a stack overflow. Oops.

CoffeeFlux:
Ah, it was me not being careful about resolving conflicts.

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
```diff
@@ -134,5 +134,53 @@ internal static Assembly DoAssemblyResolve (string name)
...
+			return null;
+		}
+#endregion
+#region Copied from AssemblyLoadContext.CoreCLR.cs, modified until our AssemblyBuilder implementation is functional
+		private static RuntimeAssembly? GetRuntimeAssembly(Assembly? asm)
+		{
+			return
+				asm == null ? null :
+				asm is RuntimeAssembly rtAssembly ? rtAssembly :
+				//asm is System.Reflection.Emit.AssemblyBuilder ab ? ab.InternalAssembly :
```
lambdageek: Is the only issue that we don't have an `InternalAssembly` property? could we add one that throws NIE and a FIXME to note to come back to it?

netcore/CoreFX.issues.rsp
```diff
@@ -382,7 +382,7 @@
 
 # TODO: Missing assembly resolve events on AppDomain and AssemblyLoadContext
 # https://github.com/mono/mono/issues/15081
--nomethod System.Resources.Tests.ResourceManagerTests.GetString_ExpectEvents
+#-nomethod System.Resources.Tests.ResourceManagerTests.GetString_ExpectEvents
```
lambdageek: Can delete this line and the two above it, no?

CoffeeFlux: Yeah all that'll get deleted before I mark this as ready. I just leave things commented while working.

mono/metadata/assembly.c
```diff
@@ -1669,6 +1672,13 @@ netcore_load_reference (MonoAssemblyName *aname, MonoAssemblyLoadContext *alc, M
 	if (reference)
 		goto leave;
 
+	// See: https://github.com/dotnet/coreclr/blob/0a762eb2f3a299489c459da1ddeb69e042008f07/src/vm/appdomain.cpp#L5178-L5239
+	if (!(strcmp (aname->name, MONO_ASSEMBLY_CORLIB_NAME) == 0 && is_satellite) && postload) {
```
CoffeeFlux: I'm still unsure on under what conditions postload should be set or not, but maybe it's best to get this merged and fix issues as they come up about it being run when it shouldn't? I'll take another look at CoreCLR tomorrow and see if it's clearer to me.

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
CoffeeFlux: I can do that if we're sure we want to use an InternalAssemblyBuilder the same way CoreCLR has. See: https://github.com/dotnet/coreclr/blob/52651008a5fb72eb678467cd2eb42aac4e5334e8/src/System.Private.CoreLib/src/System/Reflection/Emit/AssemblyBuilder.cs#L33

filipnavara:
> Addresses parts of #16256.

Self-reference? Was this supposed to be different number?

CoffeeFlux:
Yep, oops. Fixed.

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
lambdageek: I see.  Well then I guess we'll have to come back here when we work on SRE for netcore.

lambdageek:
@monojenkins build failed

netcore/System.Private.CoreLib/src/System.Runtime.Loader/AssemblyLoadContext.cs
CoffeeFlux: Yeah, looking at it again I think the best option here is to not move this function to shared yet so we don't play ping-pong with the code, then re-evaluate at a later date. I'll update the comment to be clearer though.

CoffeeFlux:
```
using System;
using System.Globalization;
using System.Resources;
using System.Runtime.Loader;
using System.Reflection;

namespace HelloWorld
{
    class Program
    {
        static System.Reflection.Assembly AssemblyResolveEventHandler(object sender, ResolveEventArgs args)
        {
            Console.WriteLine(sender == null);
            Console.WriteLine(AppDomain.CurrentDomain == sender);
            return null;
        }

        {
            Console.WriteLine("starting");
            AppDomain.CurrentDomain.AssemblyResolve += new ResolveEventHandler(AssemblyResolveEventHandler);
            ResourceManager resourceManager = new ResourceManager(typeof(Program));
            var culture = new CultureInfo("af-ZA");
            string actual = resourceManager.GetString("One", culture);
        }
    }
}
```
Output of this branch:
```
starting
False
True
False
```
.NET Core output:
```
starting
True
False
True
False
```
There are a couple other scenarios where the sender/requesting assembly seems to be set more aggressively here than in .NET Core. Somewhat amusingly, the AssemblyResolve event seems to almost always be triggered, which is what I was originally trying to test here, so I'm actually less worried about that for the moment.

CoffeeFlux:
Oh, maybe scratch that? I'm getting output like ours on preview8 - I was using preview6 before.

CoffeeFlux:
This is still different on preview8 though:
```
using System;
using System.Globalization;
using System.Resources;
using System.Runtime.Loader;
using System.Reflection;

namespace HelloWorld
{
    class Program
    {
        static System.Reflection.Assembly AssemblyResolveEventHandler(object sender, ResolveEventArgs args)
        {
            Console.WriteLine(args.RequestingAssembly != null);
            return null;
        }

        {
            Console.WriteLine("starting");
            AppDomain.CurrentDomain.AssemblyResolve += new ResolveEventHandler(AssemblyResolveEventHandler);
            AssemblyLoadContext.Default.LoadFromAssemblyName (new AssemblyName ("asdf"));
        }
    }
}
```
Us:
```
starting
True
```
Core:
```
starting
False
```

CoffeeFlux:
I think we're okay to merge this, and then I'll clarify/attempt to fix when we set RequestingAssembly in a followup PR. I'm still also unsure of whether AssemblyResolve should be triggered when resolving references, but in a similar vein I think we're better off getting this merged and dealing with issues as they come up.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
[This](https://docs.microsoft.com/en-us/dotnet/api/system.resolveeventargs.requestingassembly?view=netcore-3.0#remarks) seems to be the main documentation on when RequestingAssembly should or should not be set.

PR #16261, opened 2019-08-14
[runtime] Use a PRNG in the threadpool instead of the os random number generator.

### LINKED ISSUE #16260: mono_rand_try_get_uint32 takes too long to get a random value
### Current Behavior

I profiled mono-netcore with AOT=LLVM on macOS with the following benchmark.
https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/CSharp/aspnetcore/PlatformBenchmarks

After discussing with the team, we found that mono_rand_try_get_uint32 took too much time to just get a random number. We need to find a faster way to do that.

Screen Shot 2019-08-14 at 6 58 41 PM

**Mono version used**:
36af44e3be4a0f73586e886b96f86ee500f8ab32

## Steps to Reproduce
Reproduce by following **Profiling workflow (Method 1)** from the following link:
https://paper.dropbox.com/doc/Profile-Mono-on-macOS--Aijv~7jvs5XNRvKUtb_7UzTYAg-ryHeobv8okzUs1Fa9TfBG

vargaz:
Fixes #16260

### DIFF
--- a/mono/metadata/threadpool-worker-default.c
+++ b/mono/metadata/threadpool-worker-default.c
@@ -30,7 +30,6 @@
 #include <mono/utils/mono-proclib.h>
 #include <mono/utils/mono-threads.h>
 #include <mono/utils/mono-time.h>
-#include <mono/utils/mono-rand.h>
 #include <mono/utils/refcount.h>
 #include <mono/utils/w32api.h>
 #include <mono/utils/mono-complex.h> // This header has defines to muck with names, so put it late.
@@ -104,7 +103,6 @@ typedef struct {
 	gdouble *thread_counts;
 
 	guint32 current_sample_interval;
-	gpointer random_interval_generator;
 
 	gint32 accumulated_completion_count;
 	gdouble accumulated_sample_duration;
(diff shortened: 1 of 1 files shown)


jaykrell:
Maybe explain more in the description, like, "it is faster".

jaykrell:
Related failures.

```
00:13:37.750   libmono-static-sgen.lib(threadpool-worker-default.obj) : error LNK2019: unresolved external symbol random referenced in function rand_next [D:\j\workspace\x\msvc\monodis.vcxproj]
00:13:37.750   D:\j\workspace\x\msvc/./build/sgen/x64\bin\Release\monodis-sgen.exe : fatal error LNK1120: 1 unresolved externals [D:\j\workspace\x\msvc\monodis.vcxproj]
```

vargaz:
@monojenkins build failed

jaykrell:
Ah, but the PRNG should probably be fixed too.

mono/metadata/threadpool-worker-default.c
```diff
@@ -194,19 +192,17 @@ COUNTER_READ (void)
...
-	return mono_rand_init (NULL, 0);
-}
-
 static guint32
-rand_next (gpointer *handle, guint32 min, guint32 max)
+rand_next (guint32 min, guint32 max)
 {
 	ERROR_DECL (error);
-	guint32 val;
-	mono_rand_try_get_uint32 (handle, &val, min, max, error);
```
marek-safar: It looks like the whole thing was needed just for TP, can we remove mono_rand_try_get_uint32 + rest?

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

jaykrell:
For posterity let me note: I suspect the OS RNG use can be fixed to also be much faster.
I think we have a configuration problem in it, and I think it'd end up very similar to this.
Unless arc4rand is much slower than random() maybe.

Even if the OS is built on a slow mechanism, and reads from a driver even, it likely prefetches blocks of upcoming entropy per thread or per process.

We should consider updating OS RNG borrowing from Corefx too I think, as I showed above.
Granted, Corefx might target fewer platforms.

PR #16262, opened 2019-08-15
[interp] Fix more MonoError reuse (mono_object_isinst_checked).

jaykrell:
Whittling away at https://github.com/mono/mono/pull/16190.


jaykrell:
@monojenkins build failed

PR #16263, opened 2019-08-15
[interp] Fix more MonoError reuse (mono_object_new_checked).

jaykrell:
From https://github.com/mono/mono/pull/16190.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -2903,6 +2903,15 @@ static int opcode_counts[512];
 		} \
 	} while (0);
 
+static MONO_NEVER_INLINE MonoObject*
+mono_interp_new (MonoDomain* domain, MonoClass* klass)
+{
+	ERROR_DECL (error);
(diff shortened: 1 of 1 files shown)


jaykrell:
Merge conflict for me to resolve still.

mono/mini/interp/interp.c
```diff
@@ -3115,8 +3124,7 @@ mono_interp_newobj (
 				g_assert (exc);
 				return exc;
 			}
-			frame_objref (frame) = mono_object_new_checked (imethod->domain, newobj_class, error);
-			mono_error_cleanup (error); /* FIXME: don't swallow the error */
+			frame_objref (frame) = mono_interp_new (imethod->domain, newobj_class); // FIXME: do not swallow the error
```
jaykrell: nit due to merge: Previously this was inline and need to call to get the error allocated.
Now it is not inlined, so we could inline the "regular" call with a local MonoError.
i.e. save one function call. Probably.

jaykrell:
@monojenkins build failed

PR #16265, opened 2019-08-15
[interp] [wasm] p vs. o sledgehammer

jaykrell:
Too heavy handed, or reasonable?


BrzVlad:
I'd rather have us write to the correct field instead

jaykrell:
Yeah, I'm sympathetic to that. So *maybe* I go through and *guess* and have some PRs.

PR #16267, opened 2019-08-15
[interp] Outline newobj and newobj_vt to conserve stack.

jaykrell:
In this variation, the exception handling and resume are split between the main function and the helper function, and no address-taken occurs on ip/sp/vt_sp.

The order is a little tangled but convincing.
The helper function returns MonoException* or checks context->has_resume_state and then returns, and then the caller does the checks again.

Based on past analysis this will conserve stack, but that was not double checked here.
i.e. due to `&valuetype_this`, though it is conditional and the code could be split up into inline and outline.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -270,7 +270,13 @@ set_resume_state (ThreadContext *context, InterpFrame *frame)
 			else \
 				goto exit_frame; \
 		} \
-	} while (0);
+	} while (0)
+
+// In a void function, leave ret empty.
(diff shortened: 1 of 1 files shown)


jaykrell:
Hm, maybe it doesn't work. :(

jaykrell:
I see the problem.

jaykrell:
CI looking good!

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16268, opened 2019-08-15
[interp] Localize c, i32, o to conserve stack as suggested.

jaykrell:
This saves 16 stack bytes for clang, doesn't help gcc, it probably figured it out itself.


jaykrell:
I resolved the merge conflicts but I'll review it again myself later more carefully.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lewurm:
We should have done this way earlier, We had a couple bugs where we used for example an `i32` value set by a different opcode. Good change 👍

PR #16271, opened 2019-08-15
[draft] My own attempt at sharing mode..

jaykrell:
Took Steve's test case though.
Should maybe write some Win32 C experiments to demonstrate how it works.


jaykrell:
@monojenkins build failed

PR #16274, opened 2019-08-15
Fix failing OSX Android lane

directhex:
Linker order matters when using .a files, so zlib.a MUST come after
sgen.a if sgen.a was compiled with support for compressed ppdb


directhex:
I hadn't considered linker order when I modified dis/Makefile.am, ~~because I'm an idiot~~ because the fact that computers need this degree of handholding in the Year of Luigi 2019 is baffling.

directhex:

PR #16278, opened 2019-08-15
Exclude Authenticode tests on Watch

directhex:
1002a4de2b0d added Authenticode cert support on non-WatchOS. The tests,
however, were not non-WatchOS-only. Fix this, so Watch doesn't try to
run Authenticode tests.


directhex:

PR #16286, opened 2019-08-16
[netcore] Add a netcore version of simd-intrinsics.c.

vargaz:


vargaz:
This can run all the tests in System.Numerics.Vectors.Tests in netcore.
It also speeds up the vectorized operations in corelib like Array.IndexOf.

mono/mini/simd-intrinsics-netcore.c
```diff
@@ -0,0 +1,438 @@
+/**
+ * SIMD Intrinsics support for netcore
+ */
+
+#include <config.h>
+
+/*
+ * On LLVM is supported as a backend.
```
filipnavara: `On LLVM` -> `Only LLVM`?

mono/mini/simd-intrinsics-netcore.c
```diff
@@ -0,0 +1,438 @@
...
+	const char *class_ns;
+	MonoImage *image = m_class_get_image (cmethod->klass);
+
+	if (image != mono_get_corlib ())
+		return NULL;
+	if (!COMPILE_LLVM (cfg))
+		return NULL;
+	// FIXME:
+	if (cfg->compile_aot)
+		return NULL;
```
filipnavara: Why is it disabled for AOT btw? I guess it's ok to do it for now but a comment with explanation would be very welcome.

jaykrell: Depends on the current CPU is why?

steveisok:
@monojenkins build failed

mono/mini/simd-intrinsics-netcore.c
EgorBo: Isn't SSE2 minimal requirement for x86_64?

filipnavara: I'd expect the LLVM CPU selection should kick in but maybe it doesn't fallback quite well with current settings. Anyway, just guessing.

This makes it quite hard to test the impact. Running performance tests with `--llvm` is non-starter at the moment, it is too slow and has low test coverage. I do run performance pipeline with `-mcpu=native` for LLVM AOT and would like to see the impact there.

jaykrell: Automatic selection is troublesome for AOT and copy binaries to other system.
Yes SSE2 is required for amd64.

filipnavara: @jaykrell I understand that. That's why there's some `-mcpu=generic` default which I though would still work for this. It can be overridden if desired (eg. doing AOT on deployment).

vargaz:
@monojenkins build failed

lewurm:
```
*** start: check-prefix-llvmaot
Mono Ahead of Time compiler - compiling assembly /mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/tmp/hello.exe
AOTID E676B1BD-6147-3DC6-9167-0E3E7A3E2773
llvm load failed: /mnt/jenkins/workspace/test-mono-pull-request-amd64-prefix/tmp/monoprefix/lib/libmono-llvm.so: undefined symbol: mini_ins_info
```

EgorBo:
I guess, the netcore arm64 lane failures are also related:
```
 ^
simd-intrinsics-netcore.c: In function ‘type_to_expand_op’:
simd-intrinsics-netcore.c:139:10: error: ‘OP_EXPAND_I1’ undeclared (first use in this function)
   return OP_EXPAND_I1;
...
```

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #16288, opened 2019-08-16
Should fix many warnings: sgen-protocol-def.h adding unsigned long to string does not append.

jaykrell:


jaykrell:
@monojenkins build failed

PR #16291, opened 2019-08-16
[draft try again] Move MonoError from managed wrappers to native wrappers.

jaykrell:
Alternative to https://github.com/mono/mono/pull/15869.
 Which says: Previously, we would emit a call to memset
   since the struct is large (> 100 bytes). Only the error code
   Which that PR and this PR both fix, in different ways.

In the past:
 - This did not work due to a possible dependency on precise GC,
   interacting with what gets zeroed or not here.
   That is why I put it on the managed side.
   We'll see if that reproduces.
   (https://github.com/mono/mono/pull/11294/files#diff-efe0070415f3f0c0f24ed0221aa1962aR339
   external/coreclr/tests/src/CoreMangLib/cti/system/weakreference/weakreferenceisaliveb.exe.)

 - Some indication that embedding API should expose MonoError
   and not set_pending_exception. This does not seem relevant
   presently, and the code seems better structured and
   more efficient this way. It can revisited if/when new public API
   is the dominant concern, or do it this way for these functions,
   and another way for actual public functions.
   (https://github.com/mono/mono/pull/11294#issuecomment-432443485)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,7 +62,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This line is parsed by tools besides autoconf, such as msvc/mono.winconfig.targets.
 # It should remain in the format they expect.
 #
-MONO_CORLIB_VERSION=1b5be199-6637-495c-9baa-a95056f2137f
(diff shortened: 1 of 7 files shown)


jaykrell:
One inexplicable wasm problem in there. :(

PR #16293, opened 2019-08-16
[interp] Restore error reuse and fixup more, again.

jaykrell:
Interpreter goes to some length to only have one MonoError.
We should play along.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -2903,15 +2903,6 @@ static int opcode_counts[512];
 		} \
 	} while (0);
 
-static MONO_NEVER_INLINE MonoObject*
-mono_interp_new (MonoDomain* domain, MonoClass* klass)
-{
-	ERROR_DECL (error);
-	MonoObject* const object = mono_object_new_checked (domain, klass, error);
-	mono_error_cleanup (error); // FIXME: do not swallow the error
-	return object;
-}
-
 static
 #ifndef DISABLE_REMOTING
 MONO_NEVER_INLINE // To reduce stack.
@@ -3133,9 +3124,8 @@ mono_interp_newobj (
 				g_assert (exc);
 				return exc;
 			}
-			ERROR_DECL (error);
 			frame_objref (frame) = mono_object_new_checked (imethod->domain, newobj_class, error);
-			mono_error_cleanup (error); // FIXME: do not swallow the error
+			mono_error_cleanup (error); /* FIXME: don't swallow the error */
 			EXCEPTION_CHECKPOINT_IN_HELPER_FUNCTION;
 			sp->data.o = frame_objref (frame);
 #ifndef DISABLE_REMOTING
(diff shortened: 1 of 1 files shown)


jaykrell:
/cc @vargaz @lewurm @BrzVlad

Lambageek doesn't like `reuse` but I think it ok or preferred for interpreter.
There was another like this.

mono/mini/interp/interp.c
```diff
@@ -475,10 +476,10 @@ get_virtual_method (InterpMethod *imethod, MonoVTable *vtable)
 	MonoMethod *m = imethod->method;
 	MonoDomain *domain = imethod->domain;
 	InterpMethod *ret = NULL;
+	ERROR_DECL (error);
```
jaykrell: I don't love this, but it is what was there before.

BrzVlad:
@jaykrell why is this reverting the previous commits ? It looked fine before ?

jaykrell:
I didn’t like those changes originally or now but was temporarily convinced.

We allocate MonoError upon entry to interpreter. Allocate upon exit also seems redundant. I suppose there is an approach where we only do exit. The change/structure for entry seems to have been fairly deliberate and now we kinda contradict and moot it. Granted, we did already somewhat as noted in this PR.

There is guidance for this simple MonoError lifetime but I suggest the interpreter is not typical due to performance obsession-where-reasonable.

jaykrell:
Specific concern is paths that exit & reenter interpreter. Want one MonoError for that not two. This new pattern risks two but I’d either have to verify no reentrance or just assume reentrance which is easy enough imho.

lewurm:
could you have a look at this @lambdageek?

lambdageek:
Ok, well it's up to you folks to decide how you want to deal with errors in the interpreter @BrzVlad, @lewurm.

Jay's right that the guidance for `MonoError` is simplistic: give it a simple lifetime - in the same function that declares the error, the lifetime is terminated by mono_error_assert_ok, mono_error_cleanup, or by converting to a managed exception.  In between, pass it down to callees and ensure that at most one of the callees calls `mono_error_set_*`.

If you want to do something else in the interpreter, I suggest making the intent clearer: 
```
/* Ignore the error and reinitialize it for reuse.  The interpreter wants to save stack space, so we use a single error that might be set and reset multiple times. */
static inline void
mono_interp_error_ignore (MonoError *error)
{
   mono_error_cleanup (error);
   mono_error_reuse (error);
}
```

jaykrell:
I'd rather not think too much about what is hot or not, if writing simple enough faster code is easy enough, which I think is the case here. I realize it is subjective and more information swings it, i.e. if lack of reentrance is more or less obvious.

@lambdageek I think the appreciated guidance is if
specifically
https://github.com/mono/mono/pull/16190/commits/38aebeae38704df53ef8f2991ff63e293a3ac4de
is technically correct, and "acceptable", if in *slightly* poor taste (or *very*).

Everyone else has to decide if they can maintain the pattern -- `error_init_reuse`, by whatever name, when an error is "finished" and will be input again to a fallable path.

I think that commit makes the pattern clear enough.

BrzVlad:
I'd say it is ok how it is now. It's not really a problem to have multiple MonoError. If you really want this, I'm also open to follow aleksey's suggestion of having something like *mono_interp_error_ignore*.

PR #16295, opened 2019-08-16
eglib: Remove broken executable test on Windows.

madewokherd:
The strcmp is wrong, and it doesn't test anything meaningful
anyway.

This flag is unused since #16156 was merged.

### DIFF
--- a/mono/eglib/gfile-win32.c
+++ b/mono/eglib/gfile-win32.c
@@ -114,11 +114,8 @@ g_file_test (const gchar *filename, GFileTest test)
 	}
 
 	if ((test & G_FILE_TEST_IS_EXECUTABLE) != 0) {
-		size_t len = strlen (filename);
-		if (len > 4 && strcmp (filename + len-3, "exe"))
-		    return TRUE;
-		    
-		return FALSE;
+		/* Testing executable permission on Windows is hard, and this is unused, treat as EXISTS for now. */
+		return TRUE;
 	}
 
 	if ((test & G_FILE_TEST_IS_REGULAR) != 0) {


madewokherd:
For reference, a proper check would involve GetNamedSecurityInfo, OpenProcessToken, AccessCheck, and the FILE_EXECUTE and FILE_LIST_DIRECTORY access flags, unless I missed something simpler.

jaykrell:
I think you mean just `CreateFile`, and then perhaps `CreateFile(GENERIC_EXECUTE)` but really I'd double check if that actually matters to `CreateProcess`, if `CreateProcess` can run a non-executable file. I don't offhand.

mono/eglib/gfile-win32.c
```diff
@@ -114,11 +114,8 @@ g_file_test (const gchar *filename, GFileTest test)
 	}
 
 	if ((test & G_FILE_TEST_IS_EXECUTABLE) != 0) {
-		size_t len = strlen (filename);
-		if (len > 4 && strcmp (filename + len-3, "exe"))
-		    return TRUE;
-		    
-		return FALSE;
+		/* Testing executable permission on Windows is hard, and this is unused, treat as EXISTS for now. */
```
jaykrell: I'd strike `for now`. I don't see it ever changing. Granted msvcrt `stat` does `strcmp`. Does coreclr have anything similar?

madewokherd:
I was thinking CreateFile would have side-effects, but maybe not if you don't ask for read/write access.

jaykrell:
Well that is a good point or two. If you did ask for read access you would race with anyone just happening to read/file_share=0 at about the same time, or fail because they holding on to it (sharing violation though, not access denied -- so you infer access success?)

A non-racy test could be written to confirm that behavior. And then I hope CreateFile doesn't always internally ask for read access, I recall it does.

And then, does GetNamedSecurityInfo do similar anyway? I not sure.

In any case I think your fix is good.
Just *curious* in general.

There is an age old question, like, how to tell if a file exists -- the answer is you open it.
And then hold on to the handle and proceed to do what you were going to do.
(unless file existance is on its own a thing, and you weren't even going to read it)

If you preflight with e.g. GetFileAttributes, it is racy, and the file can be deleted right away.

This question of executableness seems to be about the same as existness.

"time of check time of use" bugs and all that.

So maybe, again, the solution would be CreateFile(executable) and hold it open through to the CreateProcess call, and then close it.

Or just leave it to CreateProcess. Unless you are searching path, and there are duplicates, and you are skipping non-executable, and really need to predict CreateProcess. Hopefully not.

madewokherd:
> "time of check time of use" bugs and all that.

Maybe you just want to know if you had permission to execute the file at some point between when you called a function and when it returned.

jaykrell:
We can try what I said -- actually CreateFile(executable).
Verify that:
 it succeeds if someone has it open for read + file_share=0
 verify it can still be open for read + file_share=0
  return the handle

The caller can decide if he wants to close it right away or not.

vs. your other way.

But your change is perfectly fine.

A caller that would hold on to the handle, can also just do the open himself.
This entire function is dubious. Given that `open` already exists and caller can just do that himself, with whatever flags (if there is a portable set..).

Maybe audit the code and remove all uses of this function, executable and otherwise, replace with `open`. Seriously. Don't enable time of check vs. use bugs. :)

lateralusX:
@monojenkins rebuild failed

PR #16297, opened 2019-08-16
[merp] Print error information when not able to start the watchdog pr…

alexischr:
…ocess


lambdageek:
Closing in favor of #16298 which also initializes the hang_watchdog_path in more cases

PR #16298, opened 2019-08-16
[merp] Initialize hang_watchdog_path earlier; be lax if it's missing

lambdageek:
Initialize hang_watchdog_path earlier - not just when
ves_icall_Mono_Runtime_EnableCrashReportingLog is called.  That function is
only expected to be called by apps that explicitly want progress reports from
MERP when it's collecting a crash report.  But we want the mono-hang-watchdog
to execute for all mono crashes (when crash reporting isn't ifdef'd out).

Also if we try to exec mono-hang-watchdog and fail for some reason, print a
nice message and then exit.  We used to call g_assert_not_reached which would
kick off another MERP of the _forked child process_ which would then get
confused because it really can't run in the forked process at all.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -6382,7 +6382,7 @@ void
 ves_icall_Mono_Runtime_EnableCrashReportingLog (const char *directory, MonoError *error)
 {
 #ifndef DISABLE_CRASH_REPORTING
-	mono_threads_summarize_init (directory);
+	mono_summarize_set_timeline_dir (directory);
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


lambdageek:

lambdageek:
@monojenkins build failed

lambdageek:
@monojenkins build failed

lambdageek:

filipnavara:

filipnavara:

filipnavara:

filipnavara:

PR #16301, opened 2019-08-16
fix minizip compile on Haiku

NattyNarwhal:
it doesn't have f*64 funcs, it seems


jaykrell:
Please also contribute upstream, or I will. 
See the readme.

PR #16303, opened 2019-08-17
[interp] Reduce register/stack pressure.

jaykrell:
No immediate savings on Linux/amd64/gcc/clang.


jaykrell:
@monojenkins build failed

PR #16305, opened 2019-08-17
[interp] Move EH data out of InterpFrame back to locals, saving 16 bytes of stack on Linux/amd64/clang.

jaykrell:
gcc is unmoved.

This is a partial undo of f8fd60b5c088753a4a0ccb79483ebae73489e7a2.
It is best to leave the register allocation to the compiler,
and keep address-taken structs to a minimum size, since compilers
do not do well with them.


jaykrell:
This needs more work.

Issue #16178, opened 2019-08-12
Math.Round doesn't emit SSE41 vround for AOT mode


EgorBo:
The following code doesn't emit SSE41 `vroundsd` or `vroundss` in AOT mode:
```csharp
    [MethodImpl(MethodImplOptions.NoInlining)]
    static float Test1(float x)
    {
        return MathF.Round(x);
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    static double Test2(double x)
    {
        return Math.Round(x);
    }
```
CoreCLR: [sharplab.io](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunykGAZgBsI2DAwBKUgPIBVAHIARABTDR4hAEouPTtR4GmAdgYBZMQAsAYgDoJEAK4A7ACbKtAbh3cAvjW+8/C6OwEIwkjIKiqTKwQ6h4VoBeoY8xCbmGBZ2jq7uml76PH7UPkA)

Tested with `--aot`, `--aot=llvm,llvmllc="-mcpu=native"`

EgorBo:
Yeah, unlike the CoreCLR, Mono is able to fold `Math.Round(3.599)` but doesn't generate SSE41 for variables

EgorBo:
hm.. it looks like it was disabled intentionally for AOT mode (portability?)
Even if I set `-mcpu=native?`

PR #16309, opened 2019-08-17
LLVM: Add more Math intrinsics, introduce `--cpu` flag

EgorBo:
1) Introduce `mcpu` flag for AOT mode, e.g.:
```
mono --aot=llvm,mcpu=native p.exe
```
will set `-mcpu=native` for both `opt` and `llc` and also, will provide an API for runtime (`use_current_cpu`) to allow it to emit any instructions current CPU supports (by default we use 'generic' CPU and aren't allowed to use anything above SSE2 on amd64)

2) Use LLVM intrinsics for some Math/MathF methods:
```csharp
static double MyCeiling(double x) => Math.Ceiling(x);
static float  MyCeilingf(float x) => MathF.Ceiling(x);

static double MyFloor(double x) => Math.Floor(x);
static float  MyFloorf(float x) => MathF.Floor(x);

static double MyFma(double x, double y, double z) => Math.FusedMultiplyAdd(x,y,z);
static float  MyFmaf(float x, float y, float z) => MathF.FusedMultiplyAdd(x,y,z);
```
LLVM IR:
```llvm

; Function Attrs: nounwind readnone uwtable
define hidden monocc double @P_MyCeiling_double(double %arg_x) #5 {
BB0:
  %t19 = call double @llvm.ceil.f64(double %arg_x)
  ret double %t19
}

; Function Attrs: nounwind readnone uwtable
define hidden monocc float @P_MyCeilingf_single(float %arg_x) #5 {
BB0:
  %t19 = call float @llvm.ceil.f32(float %arg_x)
  ret float %t19
}

; Function Attrs: nounwind readnone uwtable
define hidden monocc double @P_MyFloor_double(double %arg_x) #5 {
BB0:
  %t19 = call double @llvm.floor.f64(double %arg_x)
  ret double %t19
}

; Function Attrs: nounwind readnone uwtable
define hidden monocc float @P_MyFloorf_single(float %arg_x) #5 {
BB0:
  %t19 = call float @llvm.floor.f32(float %arg_x)
  ret float %t19
}

; Function Attrs: nounwind readnone uwtable
define hidden monocc double @P_MyFma_double_double_double(double %arg_x, double %arg_y, double %arg_z) #5 {
BB0:
  %t23 = call double @llvm.fma.f64(double %arg_x, double %arg_y, double %arg_z)
  ret double %t23
}

; Function Attrs: nounwind readnone uwtable
define hidden monocc float @P_MyFmaf_single_single_single(float %arg_x, float %arg_y, float %arg_z) #5 {
BB0:
  %t23 = call float @llvm.fma.f32(float %arg_x, float %arg_y, float %arg_z)
  ret float %t23
}
```

Output for LLVM JIT and LLVM AOT with cpu=native (fixes https://github.com/mono/mono/issues/16178):
```asm
_P_MyCeiling_double:
    18e0:	vroundsd	xmm0, xmm0, xmm0, 10
    18e6:	ret

_P_MyCeilingf_single:
    18f0:	vroundss	xmm0, xmm0, xmm0, 10
    18f6:	ret

_P_MyFloor_double:
    1920:	vroundsd	xmm0, xmm0, xmm0, 9
    1926:	ret

_P_MyFloorf_single:
    1930:	vroundss	xmm0, xmm0, xmm0, 9
    1936:	ret

_P_MyFma_double_double_double:
    1940:	vfmadd213sd	xmm0, xmm1, xmm2
    1945:	ret

_P_MyFmaf_single_single_single:
    1950:	vfmadd213ss	xmm0, xmm1, xmm2
    1955:	ret
```

Clang: https://godbolt.org/z/BX_Ood
CoreCLR: [sharplab.io](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKAAJ7biBmR02gYVoG8aG/be/esXJJaAEwgBXYABsYtALIBPdjACWs9QDsA5gApJM+bQQBKWgF4AfEuwYAFgDo1mnQfMBuQUJFiAZrIQ9vQqrlp6/vqBwRimFjZ2jgBiLhoRHmbe1D78fhLScgoqAErS2uKGhSbmVraK9s5lUhX6Xrl8+TEhSsrNFVHdcbWJDSlO/ZXtOXRCjKIFxsXKyUHQVUvxdUnOqxDr03PzAUE9KnvQg6fDCfWNqRdQbVk0HQwA9O+0uA7YUAAOsmwwAA5LhvjAoAA3SEFGC4bQguK4KT/f7QOL+fDYN7CBZGIq9ZLYjaEhBoRaE5QUgkmABetx2TmSUlwMHEiiksgw6kBygAguIpmhqQzskcutdQitsVdYqYKUNaNTaEqGdsxg4Hqz2ZzubzZAKhW0RWgxTQAL5AA==)


EgorBo:
A lot of other Math methods can be made llvm intrinsics (even without HW support).
It allows LLVM to apply a lot of InstCombine optimizations over math calls.
e.g.
```csharp
abs(X)  * abs(X)  => X * X
sqrt(X) * sqrt(X) => X
exp(X)  * exp(Y)  => exp(X + Y)
sin(X)  / cos(X)  => tan(X)
sin(-X)           => -sin(X)
logN(sqrt(X)) => 0.5 * logN(X)

etc..
```
^ these are real examples from [InstCombineCalls.cpp](https://github.com/llvm-mirror/llvm/blob/master/lib/Transforms/InstCombine/InstCombineCalls.cpp#L2311)
(of course if fast math is enabled)

vargaz:
I think the argument should have the same name as the llvm/clang argument i.e. mcpu=.

vargaz:
Also, this should probably be the default instead of being an option, aot images are not expected to be distributed afaik.

alexrp:
> (of course if fast math is enabled)

In case you are thinking of trying that, keep in mind that `-ffast-math` includes a *lot* of semantics-altering options that I doubt are valid for managed code. For one thing, NaNs and infinities are always a thing in managed land. You can also observe a difference between positive and negative zero if you try hard enough:

```csharp
> float f1 = +0.0f;
> float f2 = -0.0f;
> f1 == f2
true
> unsafe { fixed (float* p1 = &f1) { fixed (float* p2 = &f2) { Print(*(int*)p1 == *(int*)p2); } } }
false
```

[None of the other flags look particularly safe either.](https://llvm.org/docs/LangRef.html#fast-math-flags)

EgorBo:
@vargaz I agree on making it default. /cc @lewurm (who made it generic by default)

@alexrp yeah I understand, that's why it won't be enabled by default (and yeah developers should not compare floats with == even with strict mode 😉)

filipnavara:
> aot images are not expected to be distributed afaik.

Are they not? I assumed it is quite common to do so in Xamarin.Mac and we certainly do it for our product. Granted, I would expect the Xamarin tooling (mmp?) to take care of using "cpu=generic" for this case but I blindely assumed that is what Mono does by default and Xamarin inherits.

NattyNarwhal:
PPC has ceil/floor special intrins for itself, might want to convert those to the new common ones like that exist for round since this patch introduces them

mono/mini/intrinsics.c
```diff
@@ -119,17 +119,27 @@ llvm_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 			opcode = OP_ABSF;
 		else if (!strcmp (cmethod->name, "Sqrt"))
 			opcode = OP_SQRTF;
+		else if (!strcmp (cmethod->name, "Floor"))
+			opcode = OP_FLOORF;
+		else if (!strcmp (cmethod->name, "Ceiling"))
+			opcode = OP_CEILF;
+		else if (!strcmp (cmethod->name, "FusedMultiplyAdd"))
+			opcode = OP_FMAF;
```
EgorBo: I don't think I need to check for SSE41 and FMA capabilities here -- these are high level intrinsics with fallbacks, e.g. I tested it on a Ivy Bridge CPU without FMA -- the following test worked there https://gist.github.com/EgorBo/8f190158748968c9348cc04ee803b2a5

EgorBo:
@vargaz could you review please? If it looks ok I can make other Math methods LLVM intrinsics (e.g. Pow, Exp, etc).

mono/mini/mini-llvm.c
```diff
@@ -8466,6 +8520,12 @@ static IntrinsicDesc intrinsics[] = {
 	{INTRINS_SIN, "llvm.sin.f64"},
 	{INTRINS_COS, "llvm.cos.f64"},
 	{INTRINS_SQRT, "llvm.sqrt.f64"},
+	{INTRINS_FLOOR, "llvm.floor.f64"},
+	{INTRINS_FLOORF, "llvm.floor.f32"},
+	{INTRINS_CEIL, "llvm.ceil.f64"},
+	{INTRINS_CEILF, "llvm.ceil.f32"},
+	{INTRINS_FMA, "llvm.fma.f64"},
+	{INTRINS_FMAF, "llvm.fma.f32"},
 	/* This isn't an intrinsic, instead llvm seems to special case it by name */
```
EgorBo: I double checked -- floor and ceil have just one argument, fma - three (https://llvm.org/docs/LangRef.html)

EgorBo:

vargaz:
@monojenkins build failed

filipnavara:
I see pretty significant performance regression on the Burgers benchmarks which seems to have started with this PR.

Before:
|    Method |       Mean | Error |     Median |        Min |        Max |      Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|---------- |-----------:|------:|-----------:|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|
| Burgers_0 | 2,599.7 ms |    NA | 2,599.7 ms | 2,599.7 ms | 2,599.7 ms | 46000.0000 | 46000.0000 | 46000.0000 |         - |
| Burgers_1 | 2,238.5 ms |    NA | 2,238.5 ms | 2,238.5 ms | 2,238.5 ms |          - |          - |          - |         - |
| Burgers_2 |   144.3 ms |    NA |   144.3 ms |   144.3 ms |   144.3 ms |          - |          - |          - |         - |
| Burgers_3 | 2,265.6 ms |    NA | 2,265.6 ms | 2,265.6 ms | 2,265.6 ms |          - |          - |          - |         - |

After:
|    Method |       Mean |     Error |    StdDev |     Median |        Min |        Max |      Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|---------- |-----------:|----------:|----------:|-----------:|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|
| Burgers_0 | 3,279.4 ms | 280.56 ms | 311.85 ms | 3,234.5 ms | 3,003.1 ms | 3,898.9 ms | 47000.0000 | 47000.0000 | 47000.0000 |         - |
| Burgers_1 | 2,782.2 ms | 249.75 ms | 287.61 ms | 2,688.4 ms | 2,546.6 ms | 3,429.6 ms |          - |          - |          - |         - |
| Burgers_2 |   200.6 ms |  46.25 ms |  53.27 ms |   167.3 ms |   167.2 ms |   286.9 ms |          - |          - |          - |         - |
| Burgers_3 | 3,319.0 ms | 479.65 ms | 552.36 ms | 3,316.1 ms | 2,732.0 ms | 4,566.4 ms |          - |          - |          - |         - |

Both runs were done with `MONO_ENV_OPTIONS=--llvm`.

EgorBo:
@filipnavara this PR disables Max and Min intrinsics by default if `--ffast-math` is not enabled (due to IEEE754)

filipnavara:
According to the changes visible on GitHub diffs the Max/Min was already modified before.

EgorBo:
@filipnavara btw, Burgers benchmark doesn't use any of these intrinsics

filipnavara:
I misidentified the problematic PR. It is #16323 indeed. The timings between runs vary greatly now and the iterations run by BDN are significantly longer, suggesting BDN itself is affected. Running with `--ffast-math` gets rid of the problem.

filipnavara:
Turns out this was a false alarm. Sorry!

While I run the benchmarks on dedicated machine used only for that purpose it looks like some previous run of the pipeline caused a `lldb` process to be spawned. This process was eating 100% of the CPU time (one core) and causing heavily skewed results.

As of 52c1d4acfc751dd83f1964648cd91288988d7eec the results are back within my expectations:

|    Method |       Mean |   Error |  StdDev |     Median |        Min |        Max |      Gen 0 |      Gen 1 |      Gen 2 | Allocated |
|---------- |-----------:|--------:|--------:|-----------:|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|
| Burgers_0 | 2,665.4 ms | 5.39 ms | 4.21 ms | 2,665.7 ms | 2,654.4 ms | 2,669.3 ms | 46000.0000 | 46000.0000 | 46000.0000 |         - |
| Burgers_1 | 2,251.6 ms | 5.40 ms | 4.79 ms | 2,250.4 ms | 2,245.7 ms | 2,264.1 ms |          - |          - |          - |         - |
| Burgers_2 |   149.9 ms | 0.52 ms | 0.48 ms |   149.7 ms |   149.3 ms |   150.6 ms |          - |          - |          - |         - |
| Burgers_3 | 2,414.9 ms | 5.80 ms | 5.14 ms | 2,412.9 ms | 2,411.1 ms | 2,429.8 ms |          - |          - |          - |         - |

PR #16315, opened 2019-08-18
[draft] [interp] Try replacing InterpFrame.varargs with stack_args and a boolean.

jaykrell:
This does not provide any savings currently but has some potential.
i.e. might contribute to https://github.com/mono/mono/issues/16172, but the boolean + alignment make it possibly not. The boolean is used only for an assert. The boolean could probably be replaced with a bit at the bottom of an aligned pointer. i.e. add accessor macros frame_stack, frame_ip, etc. and you reclaim a few bits.


jaykrell:
Needs more work. Can replace with 32 bits though, 1 bit Boolean and 31bit offset, assuming stack is contiguous and fits within 2GB.

PR #16316, opened 2019-08-18
[interp] Replace frame.args with frame.stack, the same except for type.

jaykrell:


jaykrell:
@monojenkins build failed

PR #16321, opened 2019-08-19
[interp] Compress stack pointers to 32bit offsets.

jaykrell:
This assumes stacks are less than 4GB in size and contiguous, or at least that the alloca values and locals are within 2GB of each other.

This saves 16 bytes from interpreter size (20 aligned).

### DIFF
--- a/mono/mini/interp/interp-internals.h
+++ b/mono/mini/interp/interp-internals.h
@@ -128,12 +128,8 @@ typedef struct _InterpMethod
 } InterpMethod;
 
 struct _InterpFrame {
-	InterpFrame *parent; /* parent */
 	InterpMethod  *imethod; /* parent */
-	stackval       *retval; /* parent */
-	char           *varargs;
-	stackval       *stack_args; /* parent */
-	stackval       *stack;
+
 	/*
 	 * For GC tracking of local objrefs in exec_method ().
 	 * Storing into this field will keep the object pinned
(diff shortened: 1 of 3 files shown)


mono/mini/interp/interp-internals.h
```diff
@@ -145,9 +141,57 @@ struct _InterpFrame {
...
+	//
+	// As the pointers can be beyond 4GB, NULL has a special representation
+	// of 0. The values cannot coincide with frame itself, so that is not
+	// ambiguous.
+	//
+	// These are at the end of InterpFrame to sort by size, to minimize padding for alignment.
+	//
+	gint32 parent_offset;     // parent
+	gint32 retval_offset;     // parent
+	gint32 stack_args_offset; // parent
```
jaykrell: Probably `stack_args` should be put back due to frequent use, or interpreter should cache the full pointer in a local.
The others I think are rare enough.

Also this could use and ifdef 64bit since it only costs on 32bit, buys nothing. If we care about 32bit, 32bit interpreter.

Also you could quibble over how many bits these should be. Default Linux stack being 8MB, like 23 bits, and alignment, but 32 is convenient and a reduction.

lewurm:
It's an interesting idea, but for now I wouldn't take it because (as you said) it hurts debugability. Also I'm not sure if the GC will like it that much, as it can't follow those pointers anymore. We could add support for compressed pointers though.

Our main target regarding stack saving is WebAssembly, which is 32bit anyway. So if we need those 20bytes really badly on a 64bit platform we can reconsider.

jaykrell:
Are these pointers relevant to the GC?
I mean, they are to stack. Which it finds anyway.
Granted 32bit, does not benefit from this, and I'm reluctant to compact th pointers further.

BrzVlad:
I'm not a big fan of this. Maybe we will switch to a non recursive design and won't have to worry about it. Let's close for now and keep the option in mind.

Issue #16318, opened 2019-08-18
LLVM: MathF.Max(-0.0f, 0.0f) should return 0.0


EgorBo:
```csharp
Console.WriteLine(MathF.Max(-0.0f, 0.0f));
```
Prints `-0` when LLVM is enabled (and just `0` when any other runtime is used)
LLVM optimizes `Math.Max` into a single instruction - `vmaxss	%xmm0, %xmm1, %xmm0`
which doesn't respect this case. For exactly the same reason CoreCLR folks decided not to use it https://github.com/dotnet/coreclr/pull/22965#issuecomment-468936555

PR #16300, opened 2019-08-16
Use basic LLVM optimizations (PassManager) in JIT mode

EgorBo:
Contributes to https://github.com/mono/mono/issues/16235 

I just copied pieces from `opt.cpp` and old mono (old mono llvm 3.6 jit [used to support it](https://github.com/mono/mono/blob/1ec822976b92527fb57ef6d3ba6ccda2609d7db1/mono/mini/llvm-jit.cpp#L820-L838))
LLVM has tons of optimizations but it looks like not so much of them actually change anything for our IR yet, it can be checked this way: generate IR for different methods and feed `opt -O2 -print-before-all -print-after-all` with it. It should generate diffs for each transformation pass, e.g see: https://godbolt.org/z/QUszzv (`** IR Dump Before %pass%, ** IR Dump After  %pass%`...)

So this PR is just a minimal base (better than nothing, see `const char *opts =`)

PR #16323, opened 2019-08-19
Fast Math mode

EgorBo:
Introduce a flag to enable fast math mode (similiar to `-ffast-math` in clang/gcc and `/fp:fast` in msvc). I am not sure it should be named like this and probably should be put into `-O=` list so it needs some discussion.

This flag tells the compiler to perform aggressive FP optimizations and be less strict around IEEE and corner cases (e.g. `-0.0, Inf, NaN`). E.g. in normal mode FP operations are not associative so e.g. you can't optimize `a + b - a` into `b`. Another example is:
```csharp
static float Test(float x)
{
    return x / 10f;
}
```
```asm
; mono --llvm p.exe
movabsq	$0x7f97ee1c3a50, %rax
vdivss	(%rax), %xmm0, %xmm0  ; x / 10
retq

; mono --llvm --ffast-math p.exe
movabsq	$0x7f824bd8bad0, %rax
vmulss	(%rax), %xmm0, %xmm0 ; x * 0.1
; mul has better latency and throughput than div
retq
```

See more details in my "Fast Math" proposal for CoreCLR: https://github.com/dotnet/coreclr/issues/24784

Also put our intrinsic for `MathF.Max` under `fast math` condition. Fixes https://github.com/mono/mono/issues/16318
The flag works for both LLVM JIT and LLVM AOT modes

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -1116,6 +1116,10 @@ arch_init (MonoAotCompile *acfg)
 	acfg->align_pad_value = 0x90;
 #endif
 
+	if (mono_use_fast_math) {
+		g_string_append (acfg->llc_args, " -fp-contract=fast");
+	}
+
 #ifdef TARGET_ARM
 	if (acfg->aot_opts.mtriple && strstr (acfg->aot_opts.mtriple, "darwin")) {
 		g_string_append (acfg->llc_args, "-mattr=+v6");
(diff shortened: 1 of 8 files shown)


EgorBo:
Some benchmark results can be found here: https://github.com/mono/mono/pull/16300#issuecomment-523420709
but I used quite old CPU without FMA for them

EgorBo:
@monojenkins build failed

pr0vieh:
is cpu=native now default ? or missed this change ?

EgorBo:
@pr0vieh no, it's not (yet?)

pr0vieh:
not default is ok
but i like the idea with option mcpu=native to unleash CPU Optimizations
i like your llvm work 👍 to unleash the beast mono 😎

EgorBo:
@steveisok yeah, if the command line switch looks OK we can merge it

lewurm:
The change itself looks okay, but we should do something about the parameter. It doesn't fit well into `-O=`, since that is for mini specific optimization.

What do you think about extending `--llvm`, so for example this would be:
```
$ mono --llvm=fast-math yourprogram.exe
# or for AOT
$ mono --llvm=fast-math --aot yourprogram.exe
```
?

EgorBo:
@lewurm um.. this flag currently affects only LLVM but in future we can also do some FP related optimization in mini if it's set.

PR #16326, opened 2019-08-19
[ci] Add CI VM provisioning script for FreeBSD to the repo

akoeplinger:
We'll download it from GitHub instead of storing it in Jenkins admin config.

Trial run with FreeBSD for moving all the other provisioning scripts to the repo too.

/cc @rootwyrm


akoeplinger:
Merging since it doesn't affect other PR CI jobs.

Issue #16024, opened 2019-08-05
Native crash in ves_icall_System_Net_Sockets_Socket_Connect_internal


Therzok:
## Steps to Reproduce

N/A, see https://devdiv.visualstudio.com/DevDiv/_workitems/edit/919108

### Current Behavior

From the looks of it, an error code does not have a translation. Possible ones below.

`connect`
* EIO
* ELOOP
* ENAMETOOLONG
* ENOENT
* ENOTDIR

Other findings:
`setsockopt`
* EDOM

`socket`
* ENFILE

### Expected Behavior

It should not crash at all. Maybe throw a less descriptive exception rather than bringing down the application

## On which platforms did you notice this

[*] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Not sure, not contained in the bug report.

## Stacktrace

```
Thread 14 Crashed:: Thread Pool Worker
0   libsystem_kernel.dylib            0x00007fff6f4262c6 __pthread_kill + 10
1   libsystem_pthread.dylib           0x00007fff6f4e1bf1 pthread_kill + 284
2   libsystem_c.dylib                 0x00007fff6f3906a6 abort + 127
3   libmonosgen-2.0.dylib             0x000000010511728e mono_post_native_crash_handler + 14
4   libmonosgen-2.0.dylib             0x00000001050ba7d0 mono_handle_native_crash + 448 (mini-exceptions.c:3321)
5   libsystem_platform.dylib          0x00007fff6f4d6b5d _sigtramp + 29
6   ???                               0x00007fa73ad1b7e0 0 + 140356223088608
7   libsystem_c.dylib                 0x00007fff6f3906a6 abort + 127
8   libxammac.dylib                   0x0000000103ef1430 log_callback(char const*, char const*, char const*, int, void*) + 64
9   libmonosgen-2.0.dylib             0x000000010530610e monoeg_g_logv_nofree + 190 (goutput.c:150)
10  libmonosgen-2.0.dylib             0x00000001053061c2 monoeg_g_log + 130 (goutput.c:165)
11  libmonosgen-2.0.dylib             0x0000000105178c4c mono_w32socket_convert_error + 124 (w32socket-unix.c:1540)
12  libmonosgen-2.0.dylib             0x0000000105179139 mono_w32socket_connect + 377 (w32socket-unix.c:224)
[...]

PR #16331, opened 2019-08-19
[w32socket] Translate some errno codes to WSA_ errors imprecisely.

lambdageek:
These aren't a perfect match, but better than falling into the `default` case
where Mono asserts and crashes.

Addresses part of #16024

---

Also add `WSA_INVALID_PARAMETER` and `WSA_INVALID_HANDLE` defines

### DIFF
--- a/mono/metadata/w32socket-unix.c
+++ b/mono/metadata/w32socket-unix.c
@@ -1463,6 +1463,7 @@ mono_w32socket_convert_error (gint error)
 #ifdef ECONNRESET
 	case ECONNRESET: return WSAECONNRESET;
 #endif
+	case EDOM: return WSAEINVAL; /* not a precise match, best wecan do. */
 	case EFAULT: return WSAEFAULT;
 #ifdef EHOSTUNREACH
 	case EHOSTUNREACH: return WSAEHOSTUNREACH;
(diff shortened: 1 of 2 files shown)


mono/metadata/w32socket-unix.c
```diff
@@ -1463,6 +1463,7 @@ mono_w32socket_convert_error (gint error)
 #ifdef ECONNRESET
 	case ECONNRESET: return WSAECONNRESET;
 #endif
+	case EDOM: return WSAEINVAL; /* not a precise match, best wecan do. */
```
CoffeeFlux: Nit: space needed in "wecan".

mono/metadata/w32socket-unix.c
```diff
@@ -1472,11 +1473,12 @@ mono_w32socket_convert_error (gint error)
 #endif
 	case EINTR: return WSAEINTR;
 	case EINVAL: return WSAEINVAL;
-	/*FIXME: case EIO: return WSAE????; */
+	case EIO: return WSA_INVALID_HANDLE; /* not a precise match, best we can do. */
```
CoffeeFlux: Why this over WSAEBADF?

lambdageek: https://github.com/mono/mono/issues/16024#issuecomment-522653506 it's what w32error-unix.c does, kind of.

PR #16335, opened 2019-08-19
Pass -fvisibility=hidden to profiler modules.

directhex:
This appears to not break the profilers:

```
directhex@breakfast:/tmp$ nm -g /tmp/ponyponyponypony/lib/libmono-profiler-log.so | grep close_socket_fd
directhex@breakfast:/tmp$
directhex@breakfast:/tmp$ /tmp/ponyponyponypony/bin/mono --profile=log:report bottle.exe
99 bottles of beer on the wall, 99 bottles of beer.
Take one down and pass it around, 99 bottles of beer on the wall.

[..]

No more bottles of beer on the wall, no more bottles of beer.
Go to the store and buy some more, 99 bottles of beer on the wall.

Mono log profiler data
	Profiler version: 3.0
	Data version: 17
	Arguments: log:report
	Architecture: x86-64
	Operating system: linux
	Mean timer overhead: 86 nanoseconds
	Program startup: Mon Aug 19 14:18:30 2019
	Program ID: 17715
	Server listening on: 41539

JIT summary
	Compiled methods: 753
	Generated code size: 228046
	JIT helpers: 0
	JIT helpers code size: 0

GC summary
	GC resizes: 0
	Max heap size: 0
	Object moves: 0

Metadata summary
	Loaded images: 3
	Loaded assemblies: 3

Exception summary
	Throws: 0
	Executed finally clauses: 3

Thread summary
	Thread: 0x7fcd968c9700, name: "Finalizer"
	Thread: 0x7fcd966c8700, name: "Profiler Sampler"
	Thread: 0x7fcd9b036780, name: "Main"
	Thread: 0x7fcd964c7700, name: "Profiler Helper"
	Thread: 0x7fcd954c5700, name: "Profiler Dumper"
	Thread: 0x7fcd95cc6700, name: "Profiler Writer"

Domain summary
	Domain: (nil), friendly name: "bottle.exe"

Context summary
	Context: (nil), domain: (nil)

Counters:
	Mono System:
		User Time                      : 70.000ms
		System Time                    : 0.000ms
		Total Time                     : 140.000ms
		Working Set                    : 22032384
		Private Bytes                  : 63455232
		Virtual Bytes                  : 449630208
		Page Faults                    : 3004
		CPU Load Average - 1min        : 1.870000
		CPU Load Average - 5min        : 2.670000
		CPU Load Average - 15min       : 2.900000
	Mono JIT:
		Methods from AOT               : 536
		Methods JITted using mono JIT  : 154
		Methods JITted using LLVM      : 0

```

### DIFF
--- a/mono/profiler/vtune.c
+++ b/mono/profiler/vtune.c
@@ -33,6 +33,7 @@
 #include <mono/metadata/mono-debug.h>
 #include <mono/metadata/debug-internals.h>
 #include <mono/metadata/class-internals.h>
+#include <mono/utils/mono-publib.h>
 #include <string.h>
 #include <glib.h>
 
@@ -159,6 +160,9 @@ code_buffer_new (MonoProfiler *prof, void *buffer, int size, MonoProfilerCodeBuf
 		g_free (name);
 	}
 }
+
+MONO_API void
+mono_profiler_init_vtune (const char *desc)
 
 /* the entry point */
 void
--- a/mono/profiler/vtune.c
+++ b/mono/profiler/vtune.c
@@ -162,7 +162,7 @@ code_buffer_new (MonoProfiler *prof, void *buffer, int size, MonoProfilerCodeBuf
 }
 
 MONO_API void
-mono_profiler_init_vtune (const char *desc)
+mono_profiler_init_vtune (const char *desc);
 
 /* the entry point */
 void
--- a/mono/profiler/Makefile.am
+++ b/mono/profiler/Makefile.am
@@ -3,7 +3,8 @@ include $(top_srcdir)/mk/common.mk
 AM_CPPFLAGS = \
 	-DSUPPRESSION_DIR=\""$(datadir)/mono-$(API_VER)/mono/profiler"\" \
 	-I$(top_srcdir) \
-	$(GLIB_CFLAGS)
+	$(GLIB_CFLAGS) \
+	$(SHARED_CFLAGS)
 
 glib_libs = $(top_builddir)/mono/eglib/libeglib.la
 
--- a/mono/profiler/aot.c
+++ b/mono/profiler/aot.c
@@ -19,6 +19,7 @@
 #include <mono/metadata/assembly.h>
 #include <mono/metadata/class-internals.h>
 #include <mono/metadata/debug-helpers.h>
+#include <mono/utils/mono-publib.h>
 #include <mono/mini/jit.h>
 #include <mono/utils/mono-logger-internals.h>
 #include <mono/utils/mono-os-mutex.h>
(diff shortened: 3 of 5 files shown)


lambdageek:
I think `mono_profiler_init_aot` needs a `MONO_API`
https://github.com/mono/mono/blob/2f0cbb5aa5801144aa887a164f4c95ef146686a8/mono/profiler/aot.c#L340-L348

lambdageek:
Also vtune
https://github.com/mono/mono/blob/2f0cbb5aa5801144aa887a164f4c95ef146686a8/mono/profiler/vtune.c#L164-L165

directhex:
I thought this didn't break things - actually only doesn't break log profiler. Trying to figure out where to add some MONO_API

mono/profiler/vtune.c
```diff
@@ -160,6 +161,9 @@ code_buffer_new (MonoProfiler *prof, void *buffer, int size, MonoProfilerCodeBuf
 	}
 }
 
+MONO_API void
+mono_profiler_init_vtune (const char *desc)
```
akoeplinger:
```suggestion
mono_profiler_init_vtune (const char *desc);
```

directhex:
Remaining breakage not my fault https://github.com/mono/mono/issues/16338

jaykrell:
Please use `MONO_PROFIILER_API` instead.

jaykrell:
> I thought this didn't break things - actually only doesn't break log profiler. Trying to figure out where to add some MONO_API

`MONO_PROFILER_API`.
If profiler converted to C++, this can be visibility w/o extern "C".
I ported it twice and it was approved and rejected.

lambdageek:
I don't think the entrypoints for the profiler plugins should be `MONO_PROFILER_API`: 1. I think the entrypoints have to be C because we don't do any mangling when we form the name in the runtime when we try to load the plugin.  2. It's conceivable that third party profiler plugins are built with a compiler with different mangling rules.

akoeplinger:
Ok, merging this then to unblock the 2019-08 iOS integration.

jaykrell:
What is MONO_PROFILER_API for? I was guessing about externC.

lambdageek:
MONO_PROFILER_API is for profiler plugins calling into the runtime.  A superset of the public Mono API, including things that only make sense for the profiler or things that we don't necessarily want to make a part of the full public MONO_API.

In this PR we're looking at the entry points of the profiler plugins which have a name `"mono_profiler_init_"+pluginName`: these are called by the runtime when you say `mono --profiler=foo` - we'll look for `libprofiler-foo.dylib` and then dlsym `mono_profiler_init_foo`.  So the entrypoints have to be extern C and visible.

PR #16341, opened 2019-08-19
[interp] Relieve register/stack pressure around mono_thread_internal_current.

jaykrell:
Part of https://github.com/mono/mono/issues/16172 though not necessarily any gain from this change, at this time.
We might want to expose ifdef and inline mono_thread_internal_current also or instead.


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #16345, opened 2019-08-19
Return an empty string for assemblies loaded from stream

### LINKED ISSUE #15192: [netcore] Make System.Reflection.Tests.AssemblyTests.LoadFromStream_Location_IsEmpty Pass

```
System.Reflection.Tests.AssemblyTests.LoadFromStream_Location_IsEmpty [FAIL]
      System.IO.FileNotFoundException : 
      Stack Trace:
           at System.Reflection.Assembly.Load(AssemblyName assemblyRef, StackCrawlMark& stackMark, AssemblyLoadContext assemblyLoadContext)
           at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyName(AssemblyName assemblyName)
           at System.Reflection.Tests.AssemblyTests.LoadFromStream_Location_IsEmpty()
```

CoffeeFlux:
Fixes #15192 

The core of the issue here is that [this code](https://github.com/mono/mono/blob/7c3dfbcbdf46d057ae4f145a8ddfc8e193d90ef5/mono/metadata/image.c#L1833) sets the name field on assemblies to the pointer when loaded from a stream, and get_location simply returns that field. The [relevant docs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=netframework-4.8#System_Reflection_Assembly_Location) seem unambiguous that an empty string should be returned instead, even in framework.

If this ends up failing some legacy mono tests, I can either change the test or make this behavior netcore-only.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -5138,7 +5138,8 @@ ves_icall_System_Reflection_RuntimeAssembly_get_location (MonoReflectionAssembly
 {
 	MonoDomain *domain = MONO_HANDLE_DOMAIN (refassembly);
 	MonoAssembly *assembly = MONO_HANDLE_GETVAL (refassembly, assembly);
-	return mono_string_new_handle (domain, mono_image_get_filename (assembly->image), error);
+	const char *image_name = m_image_get_filename (assembly->image);
+	return mono_string_new_handle (domain, image_name != NULL ? image_name : "", error);
 }
 
 MonoBoolean
--- a/mono/metadata/image.c
+++ b/mono/metadata/image.c
@@ -1828,6 +1828,7 @@ mono_image_open_from_data_internal (MonoAssemblyLoadContext *alc, char *data, gu
 	image->storage = storage;
 	mono_image_init_raw_data (image, storage);
 	image->name = (name == NULL) ? g_strdup_printf ("data-%p", datac) : g_strdup(name);
+	image->filename = name ? g_strdup (name) : NULL;
 	iinfo = g_new0 (MonoCLIImageInfo, 1);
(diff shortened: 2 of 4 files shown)


mono/metadata/icall.c
```diff
@@ -5138,7 +5138,8 @@ ves_icall_System_Reflection_RuntimeAssembly_get_location (MonoReflectionAssembly
 {
 	MonoDomain *domain = MONO_HANDLE_DOMAIN (refassembly);
 	MonoAssembly *assembly = MONO_HANDLE_GETVAL (refassembly, assembly);
-	return mono_string_new_handle (domain, mono_image_get_filename (assembly->image), error);
+	const char *image_name = mono_image_get_filename (assembly->image);
+	return mono_string_new_handle (domain, strncmp (image_name, "data-", 5) != 0 ? image_name : "", error);
```
lambdageek: Can we change `mono_image_get_filename` to return `image->filename` instead of `image->name`, instead?

CoffeeFlux: Oh, yep. Didn't realize that existed since it isn't set at all in the functions I was looking at.

CoffeeFlux: Since it's public API, I've opted to leave it for now and instead add some `m_image_get_` functions for internal use. Probably not worth the line damage to update existing callers to `mono_image_get_filename`, unless you think it's worth marking external-only.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #16347, opened 2019-08-19
[interp] Specialize interpreter based on null-ness of clause_args.

jaykrell:
This saves 16 bytes of frame in the common case, and removes conditional branches in common and uncommon cases, but does cost code size.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -10,6 +10,9 @@
  *
  * (C) 2001, 2002 Ximian, Inc.
  */
+
+#ifndef interp_exec_method_inner
+
 #ifndef __USE_ISOC99
 #define __USE_ISOC99
 #endif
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

lewurm:
I'm against it:
1. Horrible usage of C macros.  Let's keep it simple.  If anything, move the interpreter main loop to a different file.
2. It's essentially duplicating the main interpreter loop, increasing the footprint even further (significantly).

jaykrell:
1. Can be fixed easily enough, but it makes for an easy to read diff.
2. Is the controversial part, agreed. Not everything is duplicated, i.e. where there are calls to functions, that do not take clause_args. That factoring can be increased for larger/rarer paths.

jaykrell:
This is btw, how you emulate templates in C. Albeit usually with a slightly different construction. You have to chose between generalized code full of size/conditionality/cost, or duplicating the code to propagate constants, paying code size and saving everything else. There is no free lunch and both options are frequently taken.

BrzVlad:
Looks like too much magic for nothing. If this is shown to have any sort of perf impact than we can revisit later.

PR #16348, opened 2019-08-19
[netcore] Consolidate running tests between AzDO pipeline and Makefile

akoeplinger:
There was some unnecessary duplication between the two that we can remove.

### DIFF
--- a/netcore/Makefile
+++ b/netcore/Makefile
@@ -188,7 +188,9 @@ run-tests-corefx: update-tests-corefx
 	tests_count=$(words $(dir $(wildcard corefx/tests/extracted/*/))); \
 	for testdir in corefx/tests/extracted/*; do \
 		counter=$$((counter+1)); \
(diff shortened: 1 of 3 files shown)


netcore/Makefile
```diff
@@ -183,14 +183,25 @@ corefx/.stamp-dl-corefx-tests-$(NETCORETESTS_VERSION):
 #
 
 run-tests-corefx: update-tests-corefx
+	@rm -f .failures; \
+	counter=0; \
+	tests_count=$(words $(dir $(wildcard corefx/tests/extracted/*/))); \
 	for testdir in corefx/tests/extracted/*; do \
-		$(MAKE) run-tests-corefx-$$(basename $${testdir}) || true; \
+		counter=$$((counter+1)); \
+		$(MAKE) run-tests-corefx-$$(basename $${testdir}) TEST_COUNTER="($$counter / $$tests_count) " || echo $$(basename $$testdir) >> .failures; \
```
EgorBo: the CI step used to run `run-step.sh` to execute each set in order to cancel if timeout is exceeded.

akoeplinger: Thanks, I added the timeout step back (using the `timeout` command directly to keep a separation between the Jenkins stuff and netcore).

PR #16350, opened 2019-08-20
[interp] Compress interpreter opcode names in usual way, offsets vs. pointers.

jaykrell:


BrzVlad:
This just makes debugging more cumbersome. This looks like it doesn't gain us anything

jaykrell:
The pointers are/can-be materialized as locals (one is hot and probably should be).
At least in the current frame. Maybe up the stack (if locals).
It does save 20 (or rather 16) bytes.
Doing the addition in the debugger isn't difficult, but yes would get cumbersome.

But ok. I'll see what I can do without it.

jaykrell:
This is the technique the rest of the runtime uses. It saves a few k.
700 opcodes
8 byte pointers compressed to 2 bytes, and 700 relocations removed.

```
const char*
mono_opcode_name (int opcode)
{
	return (const char*)&opstr + opidx [opcode];
}
```

mono/mini/interp/mintops.h
```diff
@@ -53,11 +53,18 @@ enum {
...
 extern MintOpArgType const mono_interp_opargtype[];
 extern char* mono_interp_dis_mintop(const unsigned short *base, const guint16 *ip);
 extern const guint16* mono_interp_dis_mintop_len (const guint16 *ip);
 
+// This, instead of an array of pointers, to optimize away a pointer and a relocation per string.
+extern const guint16 mono_interp_opname_offsets [ ];
+typedef struct MonoInterpOpnameCharacters MonoInterpOpnameCharacters;
+extern const MonoInterpOpnameCharacters mono_interp_opname_characters;
+
+#define mono_interp_opname(x) (((const char*)&mono_interp_opname_characters) + mono_interp_opname_offsets [x])
```
BrzVlad: Could you make this a normal global function ? It would be really annoying if I didn't have access to this while in debugger.

jaykrell:
@monojenkins build failed

PR #16354, opened 2019-08-20
[interp] The pattern of null_check and throw_ex where there is a goto *seems*

jaykrell:
to confound some of the compilers (gcc I use mostly, but could be others too) leading it to spill more to the stack.

Without being either too scientific or too thorough about it, this
adds "redundant" indentation amidst such error checks, and sometimes
redundant null checks also -- only slow path penalized by the extra checks.

This also does a little movement around of reads to be nearer next-reads,
i.e. to after the error checks.

This *seems* to contribute toward https://github.com/mono/mono/pull/16321.
but I admit I have not checked that each and every change helps,
and that I been using mostly only Linux/amd64/gcc, and a little Linux/amd64/clang.
Clang is noticably slower.
And amd64 is a bit register starved compared to arm, right?
i.e. a few more nonvolatile registers and the stack use would drop anyway.

This diff is mostly increased indentation, be sure to view w/o whitespace.

This is based on an extract from a larger PR that drives frame size to 0x78.
It is not random.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -5370,49 +5370,68 @@ main_loop:
 		}
 		MINT_IN_CASE(MINT_LDLEN) {
 			MonoObject* const o = sp [-1].data.o; // See the comment about GC safety above.
-			NULL_CHECK (o);
-			sp [-1].data.nati = mono_array_length_internal ((MonoArray *)o);
+			if (G_UNLIKELY (!o)) {
+				NULL_CHECK (o);
+			} else {
+				sp [-1].data.nati = mono_array_length_internal ((MonoArray *)o);
+			}
 			++ip;
 			MINT_IN_BREAK;
 		}
 		MINT_IN_CASE(MINT_LDLEN_SPAN) {
 			MonoObject* const o = sp [-1].data.o; // See the comment about GC safety above.
-			NULL_CHECK (o);
-			gsize offset_length = (gsize) *(gint16 *) (ip + 1);
-			sp [-1].data.nati = *(gint32 *) ((guint8 *) o + offset_length);
+			if (G_UNLIKELY (!o)) {
+				NULL_CHECK (o);
+			} else {
+				gsize offset_length = (gsize) *(gint16 *) (ip + 1);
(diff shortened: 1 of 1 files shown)


BrzVlad:
I don't really understand how this reduces stack. Anyhow, this would be logically equivalent with letting the compiler know that instructions following NULL_CHECK are unreachable if o is null. Would that work, so we don't need to do this big diff, and we would just tweak NULL_CHECK instead ?

lewurm:
`NULL_CHECK` is defined like this:

https://github.com/mono/mono/blob/ea29349f77df3a4eb553300c1b694afe656cd56a/mono/mini/interp/interp.c#L959-L962
Putting `eg_unreachable()` after `THROW_EX` should lead to the same result?

jaykrell:
_unreachable I believe is merely unlikely, not impossible.
I was looking into it for other reasons recently.
I understand these are impossible but compiler does not seem to.
Maybe there is a tweak to the macro though. Like maybe a goto w/o a condition before it.
I can double check.
The diff is mostly whitespace.
I can double check that this helps. It is less likely in master, on its own, where other cases will already use as much stack. You have to drive down equally or more-than critical paths.

jaykrell:
I can also do manual refetches, like `o = sp [n].data.p`. which will generate less good code, slightly, but smaller diff. I think that is what compiler ended up doing. if/else avoided it though.
i.e. compiler preserving stuff around `interp_throw`.

jaykrell:
Somewhere I’ve lost 16 bytes and am back up to 0x88. They could be gone a while or forever.

I am no longer able to demonstrate the value of this change. Possibly due to that.

Though arguably it makes the code clearer.
If you can stand loss of horizontal space.

I suggest and will PR, based on what this was attempting:

1. Remove the “if has_resume” in the null/throw cases, after interp_throw. Which should achieve what this was attempting.

2. Possible code size optimization.

Specifically “goto resume”.

Later maybe I can confirm if either/both of these actually achieve what I was after.

Also separately the moved reads.

Closing.

jaykrell:
https://github.com/mono/mono/pull/16371 is the follow up to this.

PR #16357, opened 2019-08-20
[interp] Outline more of leave.

jaykrell:
In my work to drive frame size to 0x78 it appears leave
was part of the critical path, consuming more stack than most cases.

This outlines more of it.

leave is the most onerous of opcodes because, besides using
many locals (old_list, ip_offset), it has four-seeming exit cases:
  throw_ex, goto exit_frame, goto main_loop, abort.
  (granted, ves_abort is throw_ex, throw_ex is either
   goto exit_frame or goto main_loop; but it is four w/o
   more surgery)

It has many outputs:
 finally_ips
 endfinally_ip
 ip
 sp
 vt_sp

What we do here is therefore a little ugly and non-ideal:
  finally_ips and endfinally_ip we take by address, forcing
    them to the stack, possibly somewhat defeating this effort,
    but only possibly -- i.e. register-rich archictures
   sp, vt_sp: caller updates (typical stategy employed elsewhere)
   ip: return new value or null, indicating use it or abort

Split into multiple pieces:
   leave1
   inlined
   leave2

It'd be nice if all the exit points were elegantly usable
 with outlined code, but it isn't easy to do w/o address-taking
 and most opcodes can get by with simpler shortcuts.

Leave is the most complicated.

Though ultimately a InterpContext with ip/sp/vt_sp/locals/stack_args passed all around by address might be good -- or put them in InterpFrame rather.

Was meant to contribute to https://github.com/mono/mono/issues/16172 but maybe not effective.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3048,7 +3048,7 @@ mono_interp_calli_nat_dynamic_pinvoke (
 // Leave is split into pieces in order to consume less stack,
 // but not have to change how exception handling macros access labels and locals.
 static MONO_NEVER_INLINE MonoException*
-mono_interp_leave (InterpFrame* child_frame)
+mono_interp_leave1 (InterpFrame* child_frame)
 {
 	stackval tmp_sp;
 	/*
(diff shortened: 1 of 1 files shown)


jaykrell:
Hm, it is looking like for efficacy, and really cleanliness and efficiency, this should be fully outlines. I will attempt.

jaykrell:
Well, it is difficult/impossible. Too many inputs/outputs and function calls.

jaykrell:
This could probably use a confirmation that it is useful before commiting.
It isn't horrible, but might not be helpful.

jaykrell:
This leave code I suspect is ultimately going to limit the frame reduction, to 0x98 bytes (unless clause_args specialization or stack pointer compression is taken). Or 0xA8 if imethod vs. frame->imethod isn't taken.

jaykrell:
Closing in deference to https://github.com/mono/mono/pull/16464.

PR #16358, opened 2019-08-20
[draft] [interp] Outline STFLD_VT.

jaykrell:
In a larger PR driving frame size down to 0x78, this case seemed to be one on the critical path.

I addressed it by refetching/recomputing values, but this should achieve similar more palably.
i.e. because klass survives mono_class_value_size, and i32 survives mono_value_copy_internal.

Also possibly related to compiler not following well the goto out of the null_check.


BrzVlad:
Could you run a simple c# program that does stfld.vt in a loop, just to see any potential perf impact. Since this is a hot opcode.

jaykrell:
Yes will do.

PR #16363, opened 2019-08-20
[runtime] Avoid resetting the tp thread name on every call to worker_callback ().

vargaz:


jaykrell:
This doesn't look right. Did I make a mistake in the generation check? I"ll step through that later to confirm.

vargaz:
It was fine, but worker_callback is called multiple times for the same thread, and it was setting the name for each invocation.

jaykrell:
The generation check is supposed to make once per thread. Maybe I messed up.

vargaz:
It was initialized with a generation = &thread->generation.

mono/metadata/threadpool.c
```diff
@@ -320,12 +326,15 @@ worker_callback (void)
 	 */
 	mono_defaults.threadpool_perform_wait_callback_method->save_lmf = TRUE;
 
+	gsize name_generation = thread->name.generation;
+	/* Set the name if this is the first call to worker_callback on this thread */
+	if (name_generation == 0)
+	   name_generation = set_thread_name (thread);
```
jaykrell: This shouldn't be needed. Unless you want it set outside the loop the first time. The tilde should make the first time through the loop set it, and each time through the loop if there was an intervening change.

You might as well change this to just:
`name_generation = set_thread_name` w/o the `if`.

Generation should also be more opaque. It could start at `-1` if that was useful.

jaykrell: The tilde is supposed to force it to run the first time through the loop. Perhaps that is too obscure? Or broken?

jaykrell: I'll step through it later today, make sure it runs once and only once per thread in master.

vargaz:
Thats what this PR does.

jaykrell:
Oh. Does `worker_callback` return? That would explain it. I thought it never returned.
Several options then.
1. Your PR.
2. Make the call unconditionally the first time in worker_callback. Maybe excess locking.
3. In set_thread_name do a flag compare and strcmp, under the lock, and call unconditionally. I guess that is too expensive though.

If worker_callback is really called, and returns, a lot, then yeah, what you have.

jaykrell:
```
	while (!mono_runtime_is_shutting_down ()) {
		gboolean retire = FALSE;
```
?

vargaz:
@monojenkins build failed

jaykrell:
Because there is break.
https://github.com/mono/mono/blob/0b6d95e8ce7c882e2db9e89fa0dec850c687f65e/mono/metadata/threadpool.c#L405

vargaz:
@monojenkins build failed

PR #16367, opened 2019-08-20
[profiler] Rename helper functions to have common prefix

akoeplinger:
Unfortunately https://github.com/mono/mono/pull/16335 didn't fix the new symbols showing up in the log profiler module.
The reason is that we actually pass `--disable-visibility-hidden` for iOS builds in https://github.com/mono/mono/blob/0b6d95e8ce7c882e2db9e89fa0dec850c687f65e/sdks/builds/ios.mk#L98

Rename the helper functions instead to have a common prefix so they can be ignored in xamarin-macios tests.

### DIFF
--- a/mono/profiler/aot.c
+++ b/mono/profiler/aot.c
@@ -231,10 +231,10 @@ helper_thread (void *arg)
 
 			FD_ZERO (&rfds);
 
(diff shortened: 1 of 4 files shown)


akoeplinger:
@monojenkins backport 2019-08

PR #16371, opened 2019-08-20
[interp] Error handling refactor (optimize non-exception path) (helps clang save stack).

jaykrell:
While working on stack reduction, I believe I found that
the compiler could not tell how THROW_EX and/or NULL_CHECK terminate.

That is, essentially, the success paths preserve volatile registers.
They do not have to be spilled/filled.
This tweak should make it clearer to the compiler, optimized common paths, and not slow down uncommon paths.

This also cleans it up a little and reduces macro expansion, and then possibly should allow for reinlining what is now used only once.

Also add some const so it can be casted away less.

Believed to save 32 bytes of stack with Linux/amd64/clang, nothing for gcc.
And a nice cleanup.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -84,13 +84,13 @@
 /* Arguments that are passed when invoking only a finally/filter clause from the frame */
 typedef struct {
 	/* Where we start the frame execution from */
-	guint16 *start_with_ip;
+	const guint16 *start_with_ip;
 	/*
 	 * End ip of the exit_clause. We need it so we know whether the resume
 	 * state is for this frame (which is called from EH) or for the original
 	 * frame further down the stack.
 	 */
-	guint16 *end_at_ip;
+	const guint16 *end_at_ip;
 	/* When exiting this clause we also exit the frame */
 	int exit_clause;
 	/* Exception that we are filtering */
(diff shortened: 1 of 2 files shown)


mono/mini/interp/interp.c
```diff
@@ -947,7 +940,7 @@ interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, con
 #define THROW_EX_GENERAL(exception,ex_ip, rethrow)		\
 	do {							\
 		interp_throw (context, (exception), (frame), (ex_ip), (rethrow)); \
-		CHECK_RESUME_STATE(context); \
+		goto resume;							  \
```
jaykrell: `if` is removed here, maybe helping compiler.

mono/mini/interp/interp.c
```diff
@@ -6603,6 +6599,28 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
 
+resume:
+	g_assert (context->has_resume_state);
+
+	if (frame == context->handler_frame && (!clause_args || context->handler_ip < clause_args->end_at_ip)) {
+
+		ip = context->handler_ip;
+		/* spec says stack should be empty at endfinally so it should be at the start too */
+		sp = frame->stack;
+		vt_sp = (guchar*)sp + imethod->stack_size;
```
jaykrell: `imethod` should be `frame->imethod` to reduce register pressure.

jaykrell:
@monojenkins build failed

mono/mini/interp/interp.c
```diff
@@ -6603,6 +6599,28 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
+			sp->data.p = frame->ex;
+			++sp;
+		}
+
+		// FIXME Reevaluate if this should be inlined.
+		finally_ips = set_resume_state (context, frame, finally_ips);
+		// goto main_loop instead of MINT_IN_DISPATCH helps the compiler and therefore conserves stack.
+		// This is a slow/rare path and conserving stack is preferred over its performance otherwise.
+		//
+		goto main_loop;
```
jaykrell: This also merits rechecking.

PR #16352, opened 2019-08-20
Inline mono_string_length_internal and mono_string_chars_internal everywhere.

jaykrell:

PR #16374, opened 2019-08-20
inline string functions but also verify GC

jaykrell:
https://github.com/mono/mono/pull/16352 but with verify.


jaykrell:
@monojenkins help

jaykrell:
What is the command to do the verified build?

lambdageek:
@monojenkins build Linux x64 Checked

lewurm:
@monojenkins build Linux x64 Checked

lewurm:
I don't think this triggers any lane 😕

PR #16378, opened 2019-08-21
[interp] Replace endsp with child_frame.retval.

jaykrell:
This should conserve stack. I didn't verify it.
child_frame being address-taken, compiler would struggle to remove it.
So we have the same value in two places.


jaykrell:
@monojenkins build failed

jaykrell:
You recall I removed endsp a while ago, in an uglier slower way.
I just didn't see how easy it is at the time.

PR #16383, opened 2019-08-21
Remove deprecated Canada/Eastern time zone check

rootwyrm:
I am proposing the removal of the Canada/Eastern zone from this test. IANA deprecated this time zone and so it is not present on *BSD, resulting in a false failure. On Linux-based systems, it is actually just a convenience copy of America/Toronto (the canonical zone.)


akoeplinger:
I'm confused, this code only runs on Windows?

rootwyrm:
Closing as in error - I misunderstood layout and this is a Windows test. The correct file is `TimeZoneInfo.Unix.cs`. Will submit a new PR.

Issue #6397, opened 2018-01-04
Inlining code size


Therzok:
## Current Behavior

https://github.com/mono/mono/blob/7a5df3c524d5eb4cf70744dc69c4803eb70b283b/mono/mini/method-to-ir.c#L81

Currently, methods of IL code size < 20 are probed for inlining.

## Expected Behavior

From this [article](https://blogs.msdn.microsoft.com/davidnotario/2004/11/01/jit-optimizations-inlining-ii/) linked by @alexischr, it seems that .NET on Windows inlines methods < 32, (or 0x20).

CoreCLR has a different algorithm to decide whether inlining is [profitable](https://github.com/dotnet/coreclr/blob/ef74ed68fdb108ed9432c40eb2acae4bf7c35bd3/src/jit/inlinepolicy.cpp#L2014)

### On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 5.11.0.26 (pull-request-6203/f74a01add57 Tue Dec 12 18:08:19 EST 2017)

**Questions**:
Should the inlining limit be bumped to 0x20?

Would this impact mobile and AOT in code size?

lewurm:
In my opinion increasing code size isn't that much of an issue on targets where we JIT, so I would be okay with increasing that number for JIT-only scenario. We shouldn't increase it for AOT/FullAOT though, because that would increase the size of binaries that we have to ship (or our customers).

But then again: We shouldn't arbitrarily tweak that number without running proper experiments. CoreCRL, as you said, uses a completely different heuristic to do the inlining decisions. Some implications of changing that number:
- A higher threshold means that we would inline more methods, and done naively, this can lead to trashing of the instruction cache.
- More inlining means more work to do for the compiler, thus it increases the compilation time of that specific method.

alexischr:
I do agree with @lewurm, the original value may have been a typo (or not!) but this is a far-reaching change. 

On the other hand, a very low "max length" value may be severely handicapping inlining, and CoreCLR (in the non-`LEGACY_BACKEND` configuration, which I assume is the current one), set the default max size to 100 (https://github.com/dotnet/coreclr/blob/ef74ed68fdb108ed9432c40eb2acae4bf7c35bd3/src/jit/compiler.h#L9302)

I think opening a PR and profiling performance in XS could be useful and fun (and AOT sizes matter less on desktop), and that can result in the inline length limit becoming a runtime or build option.

[...]

PR #16386, opened 2019-08-21
[llvm] Use a different inline limit (30) for llvm+aot.

vargaz:


EgorBo:
Related: https://github.com/mono/mono/issues/6397

mono/mini/method-to-ir.c
```diff
@@ -3812,18 +3814,31 @@ mono_method_check_inlining (MonoCompile *cfg, MonoMethod *method)
...
 			g_free (inlinelimit);
 		} else {
 			inline_limit = INLINE_LENGTH_LIMIT;
 			llvm_jit_inline_limit = LLVM_JIT_INLINE_LENGTH_LIMIT;
+			llvm_aot_inline_limit = LLVM_AOT_INLINE_LENGTH_LIMIT;
 		}
 		inline_limit_inited = TRUE;
 	}
 
+#ifdef ENABLE_NETCORE
```
lewurm: I would rather `ifdef` the definition of `LLVM_AOT_INLINE_LENGTH_LIMIT` than duplicating this code.

lewurm:
we should gather some numbers instead of arbitrary tweaking knobs

mono/mini/method-to-ir.c
vargaz: This is not really duplicate code, its more like #ifdef <future> #else <past>. The purpose is to not break existing code.

lewurm: it's kind of worse, because `llvm_aot_inline_limit` is ignored for `!ENABLE_NETCORE`, so its existence is misleading.

I'm proposing to use the block in `ENABLE_NETCORE` for everything and instead define `LLVM_AOT_INLINE_LENGTH_LIMIT` like this:

```c
#ifdef ENABLE_NETCORE
#define LLVM_AOT_INLINE_LENGTH_LIMIT 30
#else
#define LLVM_AOT_INLINE_LENGTH_LIMIT 20
#endif
```

vargaz: Thats not the same, the old code was only using the llvm limit for llvm+!aot.

lewurm: yeah, and otherwise it would fallback to `INLINE_LENGTH_LIMIT`? to make it more obvious, maybe this:

```c
#ifdef ENABLE_NETCORE
#define LLVM_AOT_INLINE_LENGTH_LIMIT 30
#else
#define LLVM_AOT_INLINE_LENGTH_LIMIT INLINE_LENGTH_LIMIT
#endif
```
anyway, it doesn't really matter.

vargaz:
@monojenkins build failed

PR #16388, opened 2019-08-21
Naricc/get total allocated bytes

naricc:
Add an API for GC.GetTotalAllocatedBytes(bool). This method returns a count of total allocation so far in the lifetime of the program. If the parameter is false, the count is imprecise. If true, it is precise but may incur a significant performance penalty.


mono/metadata/icall-decl.h
```diff
@@ -168,6 +168,7 @@ ICALL_EXPORT int ves_icall_System_Environment_get_Platform (void);
 ICALL_EXPORT int ves_icall_System_GC_GetCollectionCount (int);
 ICALL_EXPORT int ves_icall_System_GC_GetMaxGeneration (void);
 ICALL_EXPORT gint64 ves_icall_System_GC_GetAllocatedBytesForCurrentThread (void);
+ICALL_EXPORT gint64 ves_icall_System_GC_GetTotalAllocatedBytes (MonoBoolean precise, MonoError*);
```
lambdageek: nit: I don't think you need a declaration here if you're using the `HANDLES()` form in icall-def.h

mono/metadata/gc-internals.h
```diff
@@ -313,6 +313,8 @@ int mono_gc_get_los_limit (void);
 
 guint64 mono_gc_get_allocated_bytes_for_current_thread (void);
 
+guint64 mono_gc_get_total_allocated_bytes (MonoBoolean precise);
```
lambdageek: Need an implementation of this in boehm-gc.c and null-gc.c, too.

Could be just a trivial one

mono/metadata/icall-def.h
```diff
@@ -333,12 +333,12 @@ HANDLES(ENV_17, "internalGetEnvironmentVariable_native", ves_icall_System_Enviro
 HANDLES(ENV_18, "internalGetGacPath", ves_icall_System_Environment_GetGacPath, MonoString, 0, ())
 HANDLES(ENV_19, "internalGetHome", ves_icall_System_Environment_InternalGetHome, MonoString, 0, ())
 NOHANDLES(ICALL(ENV_20, "set_ExitCode", mono_environment_exitcode_set))
-
 ICALL_TYPE(GC, "System.GC", GC_10)
 NOHANDLES(ICALL(GC_10, "GetAllocatedBytesForCurrentThread", ves_icall_System_GC_GetAllocatedBytesForCurrentThread))
 NOHANDLES(ICALL(GC_0, "GetCollectionCount", ves_icall_System_GC_GetCollectionCount))
 HANDLES(GC_0a, "GetGeneration", ves_icall_System_GC_GetGeneration, int, 1, (MonoObject))
 NOHANDLES(ICALL(GC_0b, "GetMaxGeneration", ves_icall_System_GC_GetMaxGeneration))
+HANDLES(GC_11, "GetTotalAllocatedBytes", ves_icall_System_GC_GetTotalAllocatedBytes, gint64, 1, (MonoBoolean))
```
lambdageek: On the C side you declare the icall for non-netcore `System.GC`, but you didn't make the corresponding changes to `mcs/class/referencesource/mscorlib/system/gc.cs`.

Do we need this icall in non-netcore Mono?

mono/sgen/sgen-alloc.c
```diff
@@ -507,15 +508,20 @@ sgen_alloc_obj_mature (GCVTable vtable, size_t size)
 void
 sgen_clear_tlabs (void)
 {
+	guint64 total_bytes_allocated_globally = 0;
```
lambdageek: `total_bytes_allocated` is signed.

mono/metadata/sgen-mono.c
```diff
@@ -2505,6 +2515,12 @@ mono_gc_get_allocated_bytes_for_current_thread (void)
 	return info->total_bytes_allocated + info->tlab_next - info->tlab_start;
 }
 
+guint64
+mono_gc_get_total_allocated_bytes(MonoBoolean precise)
+{
+	return total_bytes_allocated;
```
lambdageek: returns a signed `gint64`, return type is unsigned `guint64`.

mono/metadata/icall.c
```diff
@@ -9326,6 +9326,12 @@ ves_icall_System_GC_GetAllocatedBytesForCurrentThread (void)
 	return mono_gc_get_allocated_bytes_for_current_thread ();
 }
 
+gint64
+ves_icall_System_GC_GetTotalAllocatedBytes (MonoBoolean precise, MonoError* error)
+{
+	return mono_gc_get_total_allocated_bytes (precise);
```
lambdageek: Calls a function that returns `guint64` then returns a signed `gint64`

mono/sgen/sgen-alloc.c
```diff
@@ -507,15 +508,20 @@ sgen_alloc_obj_mature (GCVTable vtable, size_t size)
...
 		/* A new TLAB will be allocated when the thread does its first allocation */
 		info->total_bytes_allocated += info->tlab_next - info->tlab_start;
+		total_bytes_allocated_globally += info->total_bytes_allocated;
 		info->tlab_start = NULL;
 		info->tlab_next = NULL;
 		info->tlab_temp_end = NULL;
 		info->tlab_real_end = NULL;
 	} FOREACH_THREAD_END
-}
+
```
BrzVlad: This doesn't account for threads that allocated and have died / detached.

This updates the counter only during a collection, so it is not precise. I also don't understand why would we do it this way, when for the thread specific counter we were already providing real time information. We could just iterate over all the threads during the icall and use the same computation from *mono_gc_get_allocated_bytes_for_current_thread* while keeping an incrementing counter for the allocation of threads that have died (I think in sgen_client_thread_detach_with_lock)

naricc: So, based on the API here: https://docs.microsoft.com/en-us/dotnet/api/system.gc.gettotalallocatedbytes?view=netcore-3.0 . it takes a "precise" flag. If its true, you get an accurate count, but there may be a performance penalty. If false, its fast but maybe inaccurate.

I figured being accurate up to the last minor collection was good enough for precise=false.  Can I iterate over all the threads with out taking out a lock on the list of threads, thereby blocking new thread creation? 

I will definately fix the missing counts from dead/detached threads though.

BrzVlad: I didn't notice that the PR was not ready. I personally prefer to PR when the code is ready for final review. If you prefer to do it this way (jenkins might hate you), make sure to add WIP or Draft tag just to make it clear.

I'd say it's ok to have a counter updated at collection time and for precise stop the world and iterate all threads.

mono/metadata/sgen-mono.c
```diff
@@ -820,6 +824,20 @@ sgen_finish_concurrent_work (const char *reason, gboolean stw)
 	sgen_major_collector.finish_sweeping ();
 }
 
+void
+mono_gc_stop_world ()
```
BrzVlad: I don't see why we need to add this api here

naricc: @lambdageek suggested I take this API from one of Miguel's branches, and it is a clean way to do stop the world. But I guess I do not need to actually expose it as a GC API in this PR?

lambdageek: It's from https://github.com/migueldeicaza/mono/commit/4f4934f00b959911e7879216d4664b7b54184a74. It's a pretty nice thing to have if we're going to be pausing the GC for various activities

BrzVlad: in this case we are only suspending from within sgen so we can use directly sgen_stop/restart. If we need it some other time to expose this we can add it then.

mono/metadata/icall-def.h
naricc: My understand is that we do not need it for non-netcore Mono, so I will eliminate this icall.

mono/metadata/sgen-mono.c
naricc: Remove the API.

mono/sgen/sgen-alloc.c
naricc: I will do it that way in the future, but is there another way to share partial changes to discuss stuff?

mono/sgen/sgen-gc.h
```diff
@@ -477,6 +477,8 @@ int sgen_get_current_collection_generation (void);
 gboolean sgen_collection_is_concurrent (void);
 gboolean sgen_get_concurrent_collection_in_progress (void);
 
+void sgen_set_bytes_allocated_attached(guint64);
```
lambdageek: nit: space before `(`

mono/sgen/sgen-alloc.c
```diff
@@ -507,16 +508,34 @@ sgen_alloc_obj_mature (GCVTable vtable, size_t size)
...
+void sgen_update_allocation_count (void)
+{
+	guint64 total_bytes_allocated_globally = 0;
+
+	FOREACH_THREAD_ALL (info) {
+		total_bytes_allocated_globally += info->tlab_next - info->tlab_start;
+		total_bytes_allocated_globally += info->total_bytes_allocated;
+	} FOREACH_THREAD_END
+
+	sgen_set_bytes_allocated_attached(total_bytes_allocated_globally);
```
lambdageek: nit: space before `(`

mono/sgen/sgen-alloc.c
```diff
@@ -507,16 +508,34 @@ sgen_alloc_obj_mature (GCVTable vtable, size_t size)
...
 		/* A new TLAB will be allocated when the thread does its first allocation */
 		info->total_bytes_allocated += info->tlab_next - info->tlab_start;
+		total_bytes_allocated_globally += info->total_bytes_allocated;
 		info->tlab_start = NULL;
 		info->tlab_next = NULL;
 		info->tlab_temp_end = NULL;
 		info->tlab_real_end = NULL;
 	} FOREACH_THREAD_END
+
+	sgen_set_bytes_allocated_attached(total_bytes_allocated_globally);
```
lambdageek: nit: space before `(`

mono/metadata/sgen-mono.c
```diff
@@ -2505,6 +2520,26 @@ mono_gc_get_allocated_bytes_for_current_thread (void)
 	return info->total_bytes_allocated + info->tlab_next - info->tlab_start;
 }
 
+guint64
+mono_gc_get_total_allocated_bytes(MonoBoolean precise)
+{
+	if (precise) 
+	{
```
lambdageek: nit: `{` should be on the same line as `if`

mono/metadata/sgen-mono.c
```diff
@@ -2505,6 +2520,26 @@ mono_gc_get_allocated_bytes_for_current_thread (void)
...
 
+guint64
+mono_gc_get_total_allocated_bytes(MonoBoolean precise)
+{
+	if (precise) 
+	{	
+		LOCK_GC;
+		sgen_stop_world (0, FALSE);
+
+		sgen_update_allocation_count();
```
lambdageek: nit: space before `(`

mono/metadata/sgen-mono.c
```diff
@@ -2505,6 +2520,26 @@ mono_gc_get_allocated_bytes_for_current_thread (void)
 	return info->total_bytes_allocated + info->tlab_next - info->tlab_start;
 }
 
+guint64
+mono_gc_get_total_allocated_bytes(MonoBoolean precise)
```
lambdageek: nit: space before `(`

mono/metadata/sgen-mono.c
```diff
@@ -2495,6 +2504,12 @@ mono_gc_get_los_limit (void)
 	return SGEN_MAX_SMALL_OBJ_SIZE;
 }
 
+void
+sgen_set_bytes_allocated_attached(guint64 bytes)
```
lambdageek: nit: space before `(`

lambdageek:
Oh. failing to build.  still needs a Boehm placeholder.

naricc:
@monojenkins build failed

naricc:
@monojenkins build failed

mono/metadata/sgen-mono.c
```diff
@@ -56,6 +56,13 @@ gboolean sgen_mono_xdomain_checks = FALSE;
 /* Functions supplied by the runtime to be called by the GC */
 static MonoGCCallbacks gc_callbacks;
 
+/* The total number of bytes allocated so far in program exection by all attached threads.
+ * This is not constantly syncrhonized, but only updated on each GC. */
+static guint64 bytes_allocated_attached = 0;
+
+/* Total bytes allocated so far in program exevution by detached threads */ 
+static guint64 bytes_allocated_detached = 0;
```
BrzVlad: I don't think these variables logically belong here. GC specific stuff should stay in sgen folder, and is for the best to keep sgen-mono as small as possible, since it serves as a glue.  Let's move them somewhere else, in sgen-alloc for example, expose them with extern in sgen-gc.h for the whole gc and have mono_gc_get_total_allocated_bytes  call sgen_gc_get_total_allocated_bytes which does the work. I think this would look slightly better.

naricc: I've put the variables in sgen-alloc, but added accessor functions instead of exposing them with extern, if that is acceptable. I prefer to control access to them in this way.

mono/metadata/sgen-mono.c
```diff
@@ -2055,6 +2055,8 @@ sgen_client_thread_detach_with_lock (SgenThreadInfo *p)
 
 	mono_tls_set_sgen_thread_info (NULL);
 
+	increment_bytes_allocated_detached(p->total_bytes_allocated);
```
BrzVlad: space here

mono/sgen/sgen-alloc.c
```diff
@@ -45,9 +45,15 @@
...
 static guint64 stat_bytes_alloced_los = 0;
-
 #endif
 
+/* The total number of bytes allocated so far in program exection by all attached threads.
+ * This is not constantly syncrhonized, but only updated on each GC. */
+static guint64 bytes_allocated_attached = 0;
+
+/* Total bytes allocated so far in program exevution by detached threads */ 
+static guint64 bytes_allocated_detached = 0;
```
BrzVlad: A bit too many typos for my liking. Pay closer attention when typing :)

mono/sgen/sgen-alloc.c
```diff
@@ -507,16 +514,62 @@ sgen_alloc_obj_mature (GCVTable vtable, size_t size)
...
 }
 
+void
+sgen_set_bytes_allocated_attached (guint64 bytes)
+{
+	bytes_allocated_attached = bytes;
+}
+
+void
+increment_bytes_allocated_detached(guint64 bytes)
```
BrzVlad: space here. We have the general convention to have prefixes (like sgen) for methods that are not static.

mono/sgen/sgen-gc.h
```diff
@@ -477,6 +477,9 @@ int sgen_get_current_collection_generation (void);
 gboolean sgen_collection_is_concurrent (void);
 gboolean sgen_get_concurrent_collection_in_progress (void);
 
+void sgen_set_bytes_allocated_attached (guint64 bytes);
+void increment_bytes_allocated_detached(guint64 bytes);
```
BrzVlad: space here

mono/metadata/sgen-mono.c
```diff
@@ -2505,6 +2509,15 @@ mono_gc_get_allocated_bytes_for_current_thread (void)
 	return info->total_bytes_allocated + info->tlab_next - info->tlab_start;
 }
 
+guint64
+mono_gc_get_total_allocated_bytes (MonoBoolean precise)
+{
+	return sgen_get_total_allocated_bytes (precise);
+}
+
```
BrzVlad: white space damage

mono/metadata/sgen-mono.c
```diff
@@ -2495,6 +2497,8 @@ mono_gc_get_los_limit (void)
 	return SGEN_MAX_SMALL_OBJ_SIZE;
 }
 
+
+
```
BrzVlad: white space damage

naricc:
@monojenkins build failed

naricc:
@monojenkins build failed

BrzVlad:
@naricc please fix the code formatting issues

naricc:
@monojenkins build failed

PR #16391, opened 2019-08-21
[netcore] use Release configuration for default System.Private.CoreLib build

lewurm:
For some reason it defaults to `Debug' otherwise.


akoeplinger:
the csproj sets the config to `Release`, we need to figure out what overrides it.

akoeplinger:
I looked into it a bit more and it seems the `Configuration` set inside of the System.Private.CoreLib.csproj is just completely ignored. Maybe because we're using `Microsoft.NET.Sdk` ?

PR #16394, opened 2019-08-21
Make System.Private.CoreLib Release by default

EgorBo:
No idea why it's debug by default despite the https://github.com/mono/mono/blob/master/netcore/System.Private.CoreLib/System.Private.CoreLib.csproj#L18 (and empty BuildType and CORLIB_BUILD_FLAGS)
Also, remove the Checked config (coreclr specific)


EgorBo:
Didn't notice https://github.com/mono/mono/pull/16391

Issue #15931, opened 2019-07-31
1024 hard limit of open file descriptors


mrward:
# Steps to Reproduce

Seen some reports of this with VS Mac, so I created a console project that triggers the http client error by creating a lot of FileSystemWatchers and then disposing them. The HttpClient does not seem to recover and gets stuck throwing an error.

```
HttpClient request failed. The SSL connection could not be established, see inner exception.
Unable to write data to the transport connection: The socket is not connected.
```

https://github.com/mrward/TestManyFileWatchers

Output:

```
mono TestManyFileWatchers/bin/Debug/TestManyFileWatchers.exe
10 file watchers created
20 file watchers created
30 file watchers created
40 file watchers created
50 file watchers created
60 file watchers created
70 file watchers created
80 file watchers created
90 file watchers created
100 file watchers created
110 file watchers created
120 file watchers created
130 file watchers created
140 file watchers created
150 file watchers created
160 file watchers created
170 file watchers created
180 file watchers created
190 file watchers created
200 file watchers created
210 file watchers created
220 file watchers created
230 file watchers created
240 file watchers created
250 file watchers created
HttpClient request failed. The SSL connection could not be established, see inner exception.
Unable to write data to the transport connection: The socket is not connected.
Watchers created: 254
WorkerThreadsMax: 800 CompletionPortThreadsMax 200
WorkerThreadsAvailable: 799 CompletionPortThreadsAvailable 200

Disposing all file watchers.
HttpClient request failed. The SSL connection could not be established, see inner exception.
Unable to write data to the transport connection: The socket is not connected.
Watchers created: 0
WorkerThreadsMax: 800 CompletionPortThreadsMax 200
WorkerThreadsAvailable: 799 CompletionPortThreadsAvailable 200
```

I suspect there are other ways to trigger this problem, but creating too many FileSystemWatchers is one way.

### Current Behavior

HttpClient gets stuck in a state where it cannot make any http requests:

```
HttpClient request failed. The SSL connection could not be established, see inner exception.
Unable to write data to the transport connection: The socket is not connected.

```

### Expected Behavior

HttpClient can recover from this lack of resources problem.

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono JIT compiler version 6.0.0.311 (2019-02/494641b300c Mon Jul  1 20:30:26 EDT 2019)
[...]

PR #16396, opened 2019-08-21
[threadpool-io] Throw if backend won't be able to register a FD

lambdageek:
Draft pull request to see how this behaves on other platforms.

**TODO**: Add a test case

---

The poll+select i/o selector backend can't handle file descriptor ids greater
than FD_SETSIZE.  This can happen if too many files are open and we want to
wait on it.

Previously, mono would fail in the i/o selector thread by which point it was
too late to do anything.

With this change we will fail eagerly on the thread that calls IOSelector.Add
by throwing a NotSupportedException.

Addresses https://github.com/mono/mono/issues/15931


lambdageek:
@monojenkins build failed

PR #16397, opened 2019-08-21
[netcore] Loader cleanup and LoadInDefaultContext test fixes

CoffeeFlux:
Should probably rebase this instead of squashing - the changes are largely isolated but I didn't want to make two pull requests.

The latter commit fixes the main remaining issue with the System.Runtime.Loader.Tests.DefaultLoadContextTests.LoadInDefaultContext test. Sadly, the test still can't be re-enabled because it fails the very first assert, which checks that a target assembly has never been loaded before, but should be otherwise functional.

Hopefully the order shuffling on the search hooks will not cause issues with legacy Mono, but we'll see.

### DIFF
--- a/mono/metadata/appdomain.c
+++ b/mono/metadata/appdomain.c
@@ -308,13 +308,16 @@ mono_runtime_init_checked (MonoDomain *domain, MonoThreadStartCB start_cb, MonoT
 	mono_gc_init_icalls ();
 
 	mono_install_assembly_preload_hook_v2 (mono_domain_assembly_preload, GUINT_TO_POINTER (FALSE), FALSE);
-	mono_install_assembly_preload_hook_v2 (mono_domain_assembly_preload, GUINT_TO_POINTER (TRUE), TRUE);
 	mono_install_assembly_search_hook_v2 (mono_domain_assembly_search, GUINT_TO_POINTER (FALSE), FALSE, FALSE);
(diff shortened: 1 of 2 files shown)


CoffeeFlux:
In case it's of any use, this is the Program.cs I used to run the test outside of the test harness itself: https://gist.github.com/CoffeeFlux/bd2f96c4975c49304765872b6a067d55

mono/metadata/appdomain.c
```diff
@@ -2467,9 +2467,14 @@ ves_icall_System_Reflection_Assembly_InternalLoad (MonoStringHandle name_handle,
...
 	char *name;
 
+	MonoAssembly *requesting_assembly = mono_runtime_get_caller_from_stack_mark (stack_mark);
 	MonoAssemblyLoadContext *alc = (MonoAssemblyLoadContext *)load_Context;
+
+	if (!alc)
+		alc = mono_assembly_get_alc (requesting_assembly);
 	if (!alc)
-		alc = mono_domain_default_alc (mono_domain_get ());
+		g_assert_not_reached ();
```
lambdageek: this is `g_assert (alc)`.  The same as line 2479.  Maybe just move that up to 2475?

CoffeeFlux: Ah whoops, yep.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

PR #16401, opened 2019-08-22
[llvm] Mark Vector128/Vector256 types as SIMD.

vargaz:
These are not currently supported, but marking them as simd helps llvm eliminate dead variables with these types, since
XZERO is easier to eliminate than VZERO.


vargaz:
@monojenkins build failed

PR #16408, opened 2019-08-22
[jit] Special case the ThrowHelper methods only on netcore.

vargaz:


vargaz:

vargaz:
@monojenkins build failed

vargaz:
This indeed seems to fix the i386 failures.

lateralusX:
Good, this some good input that I can use when looking into the underlying issue.

vargaz:
Merged this to fix the CI lane, but it needs to be investigated.

lateralusX:
Filed https://github.com/mono/mono/issues/16433 that I will take a look at.

PR #16420, opened 2019-08-22
Re-enable compiler server on OSX

steveisok:


steveisok:
Something is fishy still with osx.  My local works great ws CS on.

PR #16422, opened 2019-08-22
[2019-06] Add Compiler Server Workaround for 32-bit Linux

steveisok:
Currently, if the process is 32-bit, roslyn tries to determine how much
memory is available. To do that, it tries to pinvoke into GlobalMemoryStatusEx,
which is not supported on mono.

Without it, the compiler server bombs on https://github.com/dotnet/roslyn/blob/0e63260c5afb3fb5b74c357dd250e500172bcd63/src/Compilers/Server/VBCSCompiler/NamedPipeClientConnection.cs#L55-L5

Backport of https://github.com/mono/mono/pull/16405

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -1684,6 +1684,15 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
Do we need this in 2019-06 given the compiler server is disabled there?

steveisok:
Yes, @lewing would like it enabled if we can.

Issue #16395, opened 2019-08-21
[Mobile] DateTime.Now returns incorrect local time for one hour after entering DST in 39 locales.


jgold6:
## Steps to Reproduce

1. Open the attached test project
2. Set the device time on your target device to 3/8/2020 2:00 AM Central Standard Time. Note that it jumps to 3:00 AM as there is no 2:00 AM on 3/8/2020. 
If deploying to a simulator, you will have to set the time on the host workstation to set the system time for an emulator or simulator, whether iOS or Android. depending on the device/workstation, you may not be able to directly select Central Standard Time for the time zone, but may have to select, e.g. Kansas City, MO. Worth noting is that CST in Canada works correctly, and it is only CST in the US where this issue occurs, along with 38 other locales noted below.
3. Deploy the test app to your target device.
4. Observe the time in the app UI compared to the system time in the device status bar.

[DateTimeNowDSTTest.zip](https://github.com/mono/mono/files/3527388/DateTimeNowDSTTest.zip)

### Current Behavior

Time shown is one hour earlier than is shown by the system time. 

### Expected Behavior

Time shown in the app matches the system time in the system status bar.

### Notes

Although this issue only occurs on Xamarin.iOS and Xamarin.Android, it is a mono issue. 
I reported this issue against Xam.iOS and Xam.Android in two DevDiv reports (958935 and 958936), and an iOS engineer (@mandel-macaque ) confirmed this is a mono issue and believes he identified the problem area. Thus I am opening this new issue against Mono and will be closing the DevDiv issues against XI and XA.
This issue has been reported for other locales and DST change times/dates in this closed mono issue:
https://github.com/mono/mono/issues/9664
Initially the above issue was about 2019 Spring and Fall changeovers. The Fall 2019 DST changeover seems to be correct, but I still notice issues for Spring 2020 as noted below.

### Affected locales and details of when changeover is supposed to occur for each locale:

```
Cuba, Havana (Cuba Standard Time):    Jumps ahead at 12:00 AM on 3/8/2020 to 1:00 AM
US, Kansas City, MO (US Central Time):    Jumps ahead at 2:00 AM on 3/8/2020 to 3:00 AM
Anchorage, AK (Alaska Time):    Jumps ahead at 2:00 AM on 3/8/2020 to 3:00 AM
Azores ST (Ponta Delgada, Portugal):    Jumps ahead at 12:00 AM on 3/29/2020 to 1:00 AM
[...]

PR #16430, opened 2019-08-23
Fix time zone issue when jumping into DST 

MaximLipnin:
There is an issue with jumping into DST for some time zones when the incorrect date-time offset is returned for date-time in UTC (which comes from DateTime.Now). The fix is to just check if the incoming date-time is in UTC
Also I added set of tests for some time zones but they verify jumping into DST in general.
Fixes https://github.com/mono/mono/issues/16395


MaximLipnin:
@monojenkins build failed

marek-safar:
@steveisok what's the plan with this PR?

steveisok:
@MaximLipnin Asked me to test and then we'll move it forward.

steveisok:
Ran local device tests with the sample in https://github.com/mono/mono/issues/16395.  Works as expected.

steveisok:
@monojenkins backport 2019-08

steveisok:
@monojenkins backport 2019-06

Issue #16369, opened 2019-08-20
Relocations in .text on arm


radekdoulik:
It looks like AOT compiler produces shared libraries on XA/arm, which contains relocations in .text section.

Discussed it with @vargaz on slack and he thinks it was introduced recently by https://github.com/mono/mono/commit/9ff3b0d65ee4

I have tried older XA d16.1 and it indeed didn't contain relocations in `.text`, while current XA with mono-2019-06 does.

d16.1:

```
arm-linux-androideabi-readelf  -a /Users/rodo/git/monodroid-samples/android5.0/Topeka/Topeka/obj/Release/aot/armeabi-v7a/libaot-Topeka.dll.so |grep TEXT
```
Relevant part of source
```
.text 1
        .align 3
method_addresses:
        .local method_addresses
        .type method_addresses,#function
bl .Lm_0
bl .Lm_1
bl .Lm_2
...
```
No relocations in .text:
```
Section Headers:
...
  [ 5] .text             PROGBITS        00000240 000240 022a30 00  AX  0   0 16
...
Relocation section '.rel.dyn' at offset 0x178 contains 24 entries:
 Offset     Info    Type            Sym.Value  Sym. Name
00024008  00000017 R_ARM_RELATIVE   
0002401c  00000017 R_ARM_RELATIVE   
00024020  00000017 R_ARM_RELATIVE   
00024024  00000017 R_ARM_RELATIVE   
00024028  00000017 R_ARM_RELATIVE   
0002402c  00000017 R_ARM_RELATIVE   
00024030  00000017 R_ARM_RELATIVE   
00024034  00000017 R_ARM_RELATIVE   
00024038  00000017 R_ARM_RELATIVE   
0002403c  00000017 R_ARM_RELATIVE   
00024040  00000017 R_ARM_RELATIVE   
00024044  00000017 R_ARM_RELATIVE   
0002404c  00000017 R_ARM_RELATIVE   
00024050  00000017 R_ARM_RELATIVE   
00024054  00000017 R_ARM_RELATIVE   
00024058  00000017 R_ARM_RELATIVE   
0002405c  00000017 R_ARM_RELATIVE   
0002407c  00000017 R_ARM_RELATIVE   
00024080  00000017 R_ARM_RELATIVE   
00024084  00000017 R_ARM_RELATIVE   
00024088  00000017 R_ARM_RELATIVE   
0002408c  00000017 R_ARM_RELATIVE   
00024090  00000017 R_ARM_RELATIVE   
[...]

PR #16432, opened 2019-08-23
[android] remove text relocations in AOT/arm32

lewurm:
Remove support for "big assemblies" for AOT on Android running on ARM 32bit. This will remove relocations in `.text` which are not allowed in the upcoming Android Q release anymore.

_If_ a customer should run into this problem, the suggested workarounds are:
* Switch to ARM64, or
* Do not use AOT compilation.

Fixes https://github.com/mono/mono/issues/16369

### DIFF
--- a/mono/mini/aot-compiler.c
+++ b/mono/mini/aot-compiler.c
@@ -6172,7 +6172,7 @@ emit_and_reloc_code (MonoAotCompile *acfg, MonoMethod *method, guint8 *code, gui
 				if (direct_call) {
 					int call_size;
 
-					arch_emit_direct_call (acfg, direct_call_target, external_call, FALSE, patch_info, &call_size);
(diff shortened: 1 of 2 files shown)


lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2019-08

lewurm:
@radekdoulik @grendello how far do we need to backport this? The original change was landed up to `2019-02`.

radekdoulik:
Current XA master uses 2019-06, so we would like to have it at least as far as that. d16-3 uses 2019-06 as well, d16-2 uses 2019-02 and supports Android Q as well, so I think 2019-02 would need it too.

lewurm:
@monojenkins backport 2019-06

lewurm:
@monojenkins backport 2019-02

lewurm:
not sure if `2019-02` is still open.  @marek-safar ^

PR #16439, opened 2019-08-23
Pass --clr-memory-model in csc/vbc wrapper script

akoeplinger:
Should help with random CSC crashes on ARM (we've been using it in the Mono build for a while).


akoeplinger:
@BrzVlad do you think this will have negative performance effect on amd64 platforms?

BrzVlad:
@akoeplinger No. It doesn't impact code generation in any way on strong memory model platforms

akoeplinger:
@monojenkins backport 2019-08
@monojenkins backport 2019-06

Issue #15970, opened 2019-08-01
Proposal: local coop handles should be pinned.


lambdageek:
# Summary

A reference to an object on the managed heap from a local (coop) handle will pin the object, allowing raw pointers `MonoObject*` to be passed around as long as the local handle is active.

# Motivation

Right now almost all coop handles (except those that are created in the icall marshalling layer using https://github.com/mono/mono/pull/11294/files#diff-51871b3c9f77c36adec88b19019cb69aR6395) are treated by the GC like normal managed reference and do not pin the object that they point to.

If handles are pinning, a lot of code will not need to be adapted to use coop handles - as long as the callee is only manipulating objects passed from the caller, it will not need to use handles.

The only functions that will need to use handles are places that pass managed references to callees: `foo (x->field)` where `field` is also a managed object.  The caller will need to create a local handle for `x->field` and before passing the raw pointer to `foo`:
```
   MonoObject *ptr = x->field;
   MonoObjectHandle h = MONO_HANDLE_NEW (MonoObject, ptr);
   foo (ptr);
```

# Proposed behavior

1. Change the behavior of the handles to pin the objects.
2. Allow an object for which a handle is active to be passed down to callers as a raw pointer to a `MonoObject` (or one of its subclasses).
3. Each caller may assume that that any arguments passed to it are pinned.
4. To return a managed object one of these strategies must be used:
  a. the caller must allocate a local handle and intialize it and then pass the handle (`MonoObjectHandleOut`) to the callee; or,
  b. the callee may be declared with a return type like `MonoObjectHandle` and use `HANDLE_FUNCTION_ENTER`/`HANDLE_FUNCTION_RETURN_REF`  to return a handle allocated in the caller's handle stack (ie same mechanism as today).

# Arguments against

## Missing handles are hard to pin down

With this proposal, it will be difficult to find places where we forgot to pin some object. 
Right now any code that manipulates `MonoObject*` directly is suspect.  We're in the process of rewriting all of it to use the macros from `handles.h`.  If we adopt this proposal, we will have a lot of `x->y->z` raw object manipulation that is benign.  It will be more difficult to spot the needle in the haystack.  This is similar to the issues we've had in locating missing write barriers: `x->f = y` looks innocent and it's easy to forget to write `mono_object_setref (x, f, y)` instead.

# Details

## Fully preemptive suspend

(Fully preemptive suspend is Mono's historical behavior - it's what we had with Boehm and it's what we used by default with Sgen initially.)

[...]

PR #16451, opened 2019-08-24
[runtime] Make the local gc handles pin the object they point to by scanning the handle area conservatively.

vargaz:


mono/metadata/handle.c
```diff
@@ -177,6 +171,9 @@ mono_handle_new (MonoObject *obj, MonoThreadInfo *info, const char *owner)
 	mono_handle_chunk_leak_check (handles);
 #endif
 
+	// FIXME: Since we scan the handle stack inprecisely, some of the
```
jaykrell: ‘n’ to ‘m’

mono/metadata/handle.c
```diff
@@ -177,6 +171,9 @@ mono_handle_new (MonoObject *obj, MonoThreadInfo *info, const char *owner)
...
+	// membars could be removed
```
jaykrell: ‘a’ to ‘e’

jaykrell:
Looks like just comments?

vargaz:
@monojenkins build failed

mono/metadata/handle.c
```diff
@@ -319,14 +316,11 @@ mono_handle_stack_scan (HandleStack *stack, GcScanFunc func, gpointer gc_data, g
...
-	  the object.
-
-	  Note that if we're running, we know the world is stopped.
-	*/
-	if (!precise)
+	 * We're called twice, on the precise pass we do nothing.
+	 * On the inprecise pass, we pin the objects pointed to by the handles.
+	 * Note that if we're running, we know the world is stopped.
+	 */
+	if (precise)
```
jaykrell: I think this is too subtle.
Can the comment say more like:
We are called twice, first with !precise, second with precise (or whatever is true).
Being called with precise was historically useful.
Now everything needed is on the imprecise pass, and nothing is done on the precise pass.
In the imprecise pass, we pin the objects pointed to by the handles.

BrzVlad:
I don't understand this. Why do we need to pin them ? Wasn't the point of coop handles to have precise scanning of objects passed from managed to native and avoid pinning of everything, maybe in the future avoid scanning the native frames ? If we do this, then whats the point of coop handles ? (aside from a lot of indirection and performance hits) :confused:

vargaz:
The problem with precise handles is that you have to use them everywhere, needing to rewrite all the runtime code, and the resulting code looks very ugly. With pinning handles, you can create a handle to an object, then pass the naked objref to a function which can use it normally without having to worry about it being collected.

https://github.com/mono/mono/issues/15970

jaykrell:
Agreeing with @vargaz: The handles still need to exist and their lifetime covers the pointers, but the pointers can then be used like before.

We should also try to use the more efficient managed handles more, cheaper than the native ones. Esp. in JIT icalls.

BrzVlad:
@vargaz @lambdageek So the only use of the handles as of right now is to prevent the object in question from being collected. Were there cases where the normal conservative pinning from stack/regs was failing to detect an object ?

BrzVlad:
sorry, misclick

PR #16456, opened 2019-08-25
[interp] Outline box.

jaykrell:

Must be paired with additional savings, to get 16 bytes saved on Linux/amd64/clang.
Such as https://github.com/mono/mono/pull/16530.
No change with gcc.


lewurm:
@monojenkins squash

PR #16458, opened 2019-08-25
[interp] Cleanup stackval_to_data and stackval_to_data_addr.

jaykrell:
No functional change, except for debugging, look up a frame
for the prior value.


jaykrell:
@monojenkins build failed

PR #16461, opened 2019-08-25
[interp] Non-recursive interpreter

vargaz:

PR #16464, opened 2019-08-25
[interp] Conserve stack in leave.

jaykrell:
And free up some space for other opcodes (e.g. enum_hasflag).

This should be more effective than https://github.com/mono/mono/pull/16357.
This turns child_frame into a union that can be *carefully* reused in non-recursive paths for locals.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3251,7 +3251,14 @@ mono_interp_store_remote_field_vt (InterpFrame* frame, const guint16* ip, stackv
 static void
 interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClauseArgs *clause_args, MonoError *error)
 {
-	InterpFrame child_frame;
+	union {
+		InterpFrame child_frame;
+		struct {
+			GSList *old_list;
+			guint32 ip_offset;
+		} leave;
+	} u;
+	InterpFrame* const child_frame = &u.child_frame;
 	GSList *finally_ips = NULL;
 	const guint16 *endfinally_ip = NULL;
 	const guint16 *ip = NULL;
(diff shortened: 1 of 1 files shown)


lewurm:
that's a bit too clever imho. is it really worth it?

jaykrell:
Imho yes.

Most opcodes do not use child_frame, but they all kinda pay for it.
That space might as well be repurposed.


It is somewhat "important" and clear distinction in the code, once you study it a bit.

1. leave gets in the way of "everything else".
It is kinda the most complicated opcode, with its heavy use and mutation of locals and several exit conditions.

2. This can be used a bit more, like for enum_hasflag
 and maybe others that I outlined.

Or else https://github.com/mono/mono/pull/16461.
 Which obviates a lot, but not all, of my changes.
 
Or else, the recursive/nonrecursive split.
 I think w/o that split I can get to 0x98 (https://github.com/mono/mono/pull/16457)
 Or maybe 0x88 bytes w/ the split -- which granted is a heavy price,
 but I'll try it, maybe get lower.
 but again #16461 pretty much obviates everything else, except small code quality fixes.

BrzVlad:
@jaykrell Does this save two stack slots ? Is child_frame pointer allocated into a stack_slot or is the structure referenced directly from frame base pointer offset ?

BrzVlad:
I was wondering if when doing _child_frame->field1_, _child_frame->field2_ the compiler loads it through _mov rbp(offset1)_, _mov rbp(offset2)_  directly or it first loads the child_frame pointer from local variable (_mov rbp(child_frame_ptr_offset), rax_; _mov rax(field1_offset)_, _mov _rax(field2_offset)_). Or I'm asking if a stack_slot (local) is allocated for child_frame*

BrzVlad:
Yeah, I just checked and no change on gcc. Check clang please. No local is created for child_frame* so at least that is good. The change looks decent, but let's make sure that there is improvement.

BrzVlad:
It's ok to have it outlined for now

PR #16467, opened 2019-08-26
[draft] Make each dynamic codeman its own mass-destroyed heap.

jaykrell:
This seems reasonable but something like this was crashing before.
I suspect this is hitting a use after free bug, but finding it going to be onerous given the Windows build situation.


jaykrell:
This should work. That it doesn't, suggests a problem elsewhere. Needs debugging.

akoeplinger:

PR #16468, opened 2019-08-26
CodeManager cleanup.

jaykrell:
- Chunks do not need flags, only codemanagers do.
 - Sort by size to minimize padding for alignment.
 - Booleans only need one bit.
 - Chunk is not spelled chunck.
 - Suggest foo accessors be named foo() and set_foo(), not get_foo() and set_foo().
   i.e. the more common "get" is implied.
- Allow free(0) w/o initialization.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16470, opened 2019-08-26
[draft] Cleanup arm64 SeqPointInfo.

jaykrell:


jaykrell:
Hm, just for cross builds to get the offset. Will revisit.

PR #16471, opened 2019-08-26
Fix assert in x86_patch regarding second byte of Jcc instructions being 0x80-0x8F vs. 0x70-0x8F (6byte conditional branch with 32bit offset).

jaykrell:


BrzVlad:
this really needs an explanation

BrzVlad:
@monojenkins build failed

PR #16474, opened 2019-08-26
Cleanup x86_patch and amd64_patch.

jaykrell:
amd64_patch has two incorrect looking probably unused cases.
 g_assert(0) and fix them.
 call and jmp indirect (one case)
 move reg RIP relative

x86_patch is large-ish inline macro.
Presumably for sharing with amd64, not inlining.
Restructure as an inline function, called once.
Try to make the code a little clearer also, with same functionality and efficiency
 (modulo the optimized perf of "if i == 4" vs. "if i")
i.e. speak of instruction_size and offset_size, not pos and boolean size.

Add to x86_patch the ability to patch jmp/call indirect,
as amd64_patch has, but which was incorrect.
This would seem useful for a later change (using "thunks"
to patch nearby data instead of patching aligned instructions,
indirect call/jmp vs. direct).

amd64_patch largely delegates to x86_patch.
One case asserts and delegates.
Move the assert to the x86 side, where an optimizing
compiler should know to remove it for x86 anyway.

Change some assert to g_assert.

Remove some unnecessary casts.

Fix an assert (https://github.com/mono/mono/pull/16471).

### DIFF
--- a/mono/arch/x86/x86-codegen.h
+++ b/mono/arch/x86/x86-codegen.h
@@ -359,29 +359,65 @@ typedef union {
  * the instruction is inspected for validity and the correct displacement
  * is inserted.
  */
-#define x86_patch(ins,target)	\
-	do {	\
-		unsigned char* pos = (ins) + 1;	\
-		int disp, size = 0;	\
-		switch (*(unsigned char*)(ins)) {	\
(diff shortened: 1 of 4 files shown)


jaykrell:
ARMv7 crashed in csc.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

Issue #16433, opened 2019-08-23
Windows x86 hits assert in trampoline when using OP_NOT_REACHED in some scenarios.


lateralusX:
When running xunit corelib tests we hit the following assert (more likely on 32-bit coop suspend):

Invalid trampoline sequence: 0 0 0 0 0 0 cc
* Assertion: should not be reached at ..\mono\mini\tramp-x86.c:122

Currently disguised after https://github.com/mono/mono/pull/16408, but needs to be investigated.

lateralusX:
Should be fixed by https://github.com/mono/mono/pull/16476.

PR #16476, opened 2019-08-26
Fix special case tramp assert in x86/amd64 when using OP_NOT_REACHED.

lateralusX:
https://github.com/mono/mono/pull/16408 temporary fixed a tramp issue on x86 (also exist on amd64). This commit fixes the underlying issue and re-enable the use of noreturn outside of netcore.

The problem hit by the use of OP_NOT_REACHED on x86/amd64 is due to a specific case when the address to patch happens to fall at the start of a different managed method. Since the use of OP_NOT_REACHED can end a method with the call instruction on x86/amd64 in combination with how patch location (using the return address from stack) is resolved, the implementation in
mono_arch_patch_callsite didn't take this case into account, reading incorrect patch data, triggering assert on x86. So, based on timing we could end up with the following code:

025715B3 call 02570D98
025715B8 push ebp

since the patch target is 025715B8, but also the start of a completely different method, method_start == origin_code. The fix is to detect this case on x86/amd64 (done on each arch since the way origin_code is detected is arch specific) and then use a method_start of NULL, that is already a supported scenario.

This fixes https://github.com/mono/mono/issues/16433.

### DIFF
--- a/mono/mini/tramp-amd64.c
+++ b/mono/mini/tramp-amd64.c
@@ -136,7 +136,12 @@ mono_arch_patch_callsite (guint8 *method_start, guint8 *orig_code, guint8 *addr)
 {
 	guint8 *code;
 	guint8 buf [16];
-	gboolean can_write = mono_breakpoint_clean_code (method_start, orig_code, 14, buf, sizeof (buf));
+
+	// Since method_start is retrieved from function return address (below current call/jmp to patch) there is a case when
+	// last instruction of a function is the call (due to OP_NOT_REACHED) instruction and then directly followed by a
+	// different method. In that case current orig_code points into next method and method_start will also point into
+	// next method, not the method including the call to patch. For this specific case, fallback to using a method_start of NULL.
+	gboolean can_write = mono_breakpoint_clean_code (method_start != orig_code ? method_start : NULL, orig_code, 14, buf, sizeof (buf));
 
 	code = buf + 14;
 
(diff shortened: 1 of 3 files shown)


mono/mini/tramp-amd64.c
```diff
@@ -136,7 +136,12 @@ mono_arch_patch_callsite (guint8 *method_start, guint8 *orig_code, guint8 *addr)
 {
 	guint8 *code;
 	guint8 buf [16];
-	gboolean can_write = mono_breakpoint_clean_code (method_start, orig_code, 14, buf, sizeof (buf));
+
+	// Since method_start is retrieved from function return address (bellow current call/jmp to patch) there is a case when
```
jaykrell: "below"

jaykrell:
I suggest op_notreached end with a hardcoded breakpoint or other failfast kind of thing, like int3.
As occurs for example after you call abort() in C.
That way a return address will always be in same function.

>type 1.c
#include <stdlib.h>

void f()
{
 abort();
}

C:\s\mono2>cl /c /O2 1.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.21.27702.2 for x64

It actually does not do it when not optimizing.


>link /dump /disasm 1.obj

f:
  0000000000000000: 48 83 EC 28        sub         rsp,28h
  0000000000000004: E8 00 00 00 00     call        abort
  0000000000000009: CC                 int         3

Are there other occurences and problems with call at end of function, that this would paper over everywhere?

PR #16478, opened 2019-08-26
Map OP_NOT_REACHED to OP_BREAK, similar to C abort().

jaykrell:
Alternative to https://github.com/mono/mono/pull/16476.
Only for x86/amd64, should be for all.


SamMonoRT:
@jaykrell @vargaz  - any update ?

jaykrell:
It seems like this would improve a fragility in the system, where OP_NOT_REACHED leads to corner cases. Handling it this way might simplify elsewhere.
i.e. https://github.com/mono/mono/pull/16476

PR #16483, opened 2019-08-26
[cominterop] Implement native-to-managed safearray marshalling.

nsivov:
This is a case of CCW calling managed method with object[] argument.

### DIFF
--- a/mono/metadata/cominterop.c
+++ b/mono/metadata/cominterop.c
@@ -3239,7 +3239,7 @@ mono_cominterop_emit_marshal_safearray (EmitMarshalContext *m, int argnum, MonoT
 						}
 						while (mono_marshal_safearray_next(safearray, indices));
 					} // label2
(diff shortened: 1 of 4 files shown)


nsivov:
I don't quite understand why it fails, looks like it only happens with cygwin builds? Is this the problem?

> cominterop.exe
> =============== cominterop.exe.stdout ===============
> cant resolve internal call to "Tests/_TestClass::.ctor" (tested without signature also)
> 
> Your mono runtime and class libraries are out of sync.
> The out of sync library is: D:\j\workspace\w\mono\tests\cominterop.exe

nsivov:
Ok, I'll take a closer look.

nsivov:
@vargaz please take a look, I fixed broken wrapper logic that corrupted user-allocated argument.

vargaz:
@monojenkins build failed

PR #16484, opened 2019-08-26
Fix many printf warnings, for example:

jaykrell:
Do not `printf` `long long` or `unsigned long long`.

*Do* printf `gint64` or `guint64`.

Do not printf %ll.
Do not printf %z.

Each of both of them warns with Mingw, many times.

 gint64 ==  int64_t, which could be long, or long long, or maybe other
guint64 == uint64_t, ditto

Windows had %I64 and %I in 1999.
64bit Unix gets by with %l and %ll.
%ll is only needed on 32bit Unix.
  (Or you can split the integer up and use %x or format it yourself, easy, faster).

For printing gint64/guint64, use PRId64, PRIx64, etc.
For printing gsize/gssize, use `G_GSIZE_FORMAT` (seems like excess `G`).

%z definitely would warn for every single use, with Mingw.

And %ll might also, since on 64bit platforms, gint64 tends to be
long, not long long, and would merit %l not %ll.
(No matter that %ll is merited on 32bit, and the behavior
would then be correct on both.).

If there becomes a system w/o PRId64 etc., and it is a 64bit system,
then we could use %l under a portable name.
Or %I64. This appears unlikely (unless reach is extended to older Windows runtime).

Note that PRIxPTR for gsize/intptr_t may be a good idea too (again with portability layers among capital eye %I for old Windows, lowercase ell %l for all Unix).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4741,17 +4741,31 @@ esac
 
 case "$TARGET" in
 *32* | ARM | X86 | POWERPC | MIPS | SPARC | WASM | ARM6432)
-    AC_DEFINE([MONO_HOST32], [1], [Target is 32bit])
-    AC_DEFINE([MONO_HOST64], [0], [Target is not 64bits])
+    AC_DEFINE([MONO_TARGET32], [1], [Target is 32bit])
+    AC_DEFINE([MONO_TARGET64], [0], [Target is not 64bits])
     target_sizeof_void_p=4
     ;;
 *64* | S390X)
-    AC_DEFINE([MONO_HOST32], [0], [Target is not 32bit])
-    AC_DEFINE([MONO_HOST64], [1], [Target is 64bits])
+    AC_DEFINE([MONO_TARGET32], [0], [Target is not 32bit])
+    AC_DEFINE([MONO_TARGET64], [1], [Target is 64bits])
     target_sizeof_void_p=8
     ;;
 *)
-    AC_MSG_ERROR([unknown target])
+    AC_MSG_ERROR([Unable to determine if target is 32bit or 64bit.])
+    ;;
+esac
+
+case "$HOST" in
+*32* | ARM | X86 | POWERPC | MIPS | SPARC | WASM | ARM6432)
+    AC_DEFINE([MONO_HOST32], [1], [Host is 32bit])
+    AC_DEFINE([MONO_HOST64], [0], [Host is not 64bits])
+    ;;
+*64* | S390X)
+    AC_DEFINE([MONO_HOST32], [0], [Host is not 32bit])
+    AC_DEFINE([MONO_HOST64], [1], [Host is 64bits])
+    ;;
+*)
+    AC_MSG_ERROR([Unable to determine if host is 32bit or 64bit.])
     ;;
 esac
 
--- a/configure.ac
+++ b/configure.ac
@@ -4741,9 +4741,13 @@ esac
 
 case "$TARGET" in
 *32* | ARM | X86 | POWERPC | MIPS | SPARC | WASM | ARM6432)
+    AC_DEFINE([MONO_HOST32], [1], [Target is 32bit])
+    AC_DEFINE([MONO_HOST64], [0], [Target is not 64bits])
     target_sizeof_void_p=4
     ;;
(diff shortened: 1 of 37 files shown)


mono/eglib/glib.h
```diff
@@ -1386,7 +1386,22 @@ glong     g_utf8_pointer_to_offset (const gchar *str, const gchar *pos);
...
+#define MONO_HOST64 1
+#elif _WIN32 || __ILP32__ || __WORDSIZE == 4
+#define MONO_HOST32 1
+#define MONO_HOST64 0
+#else
+//#error unknown mono32/64
+#define MONO_HOST32 1
+#define MONO_HOST64 0
+#endif
+#endif
```
jaykrell: This isn't used, but I like the idea. Instead of `#if SIZEOF_VOID_P == 4`.

configure.ac
```diff
@@ -4706,9 +4706,13 @@ esac
 
 case "$TARGET" in
 *32* | ARM | X86 | POWERPC | MIPS | SPARC | WASM | ARM6432)
+    AC_DEFINE([MONO_HOST32], [1], [Target is 32bit])
```
jaykrell: This isn't used but I like the idea instead of `#if SIZEOF_VOID_P == 4`.
Note that in the Windows world this is summed up as `#if _WIN64` or `#ifdef _WIN64`.

lambdageek:
LGTM now (sorry took so long to return to it). has some merge conflicts by now

jaykrell:
@monojenkins build failed

configure.ac
akoeplinger: if this is about the target, shouldn't the variable be called `MONO_TARGET32` and `MONO_TARGET64` then?

jaykrell: It is supposed to be about the host but it looks like I messed it up.

jaykrell: Both would be reasonable, if done correctly.

configure.ac
```diff
@@ -4741,13 +4741,31 @@ esac
...
+*32* | ARM | X86 | POWERPC | MIPS | SPARC | WASM | ARM6432)
+    AC_DEFINE([MONO_HOST32], [1], [Host is 32bit])
+    AC_DEFINE([MONO_HOST64], [0], [Host is not 64bits])
+    ;;
+*64* | S390X)
+    AC_DEFINE([MONO_HOST32], [0], [Host is not 32bit])
+    AC_DEFINE([MONO_HOST64], [1], [Host is 64bits])
+    ;;
+*)
+    AC_MSG_ERROR([Unable to determine if host is 32bit or 64bit.])
```
jaykrell: Is it ok for the default to be an error?
Or just remove all of this "for now"? (forever?)

CoffeeFlux:
It would be nice if these rules surrounding printf were included on the website's coding guidelines. If you don't want to PR that I can handle it, but someone should or violations will repeatedly be reintroduced.

configure.ac
lambdageek: I lean slightly toward removing this.  Mostly because it's somewhat ambiguous what a "32-bit host" is.  I'm thinking of ARM64_32 for example.  Currently we're more specific (but somewhat - decentralized): we ask "is `sizeof(void*) == 8`" or about register sizes, or about alignment.  "64-bit host" could be shorthand for a number of conflated concerns and it might be difficult to untangle on a new architecture.

akoeplinger: Yeah I agree, let's wait until we have a good reason to use these defines. The important part of this PR are the printf warning fixes so let's get those in :)

jaykrell: 1. I pushed before seeing some of this, not all of it.
2. I cannot see yet what lambdageek pushed. Still looking.
3. Question is -- keep the MONO_TARGET stuff or not? Understood to remove MONO_HOST.

jaykrell: Ok, while I cannot find the commit, the description is indicatative:
```
Pushed a commit to remove MONO_{HOST,TARGET}{32,64}
```
so I'll do what it says.

jaykrell: Though I think the `MONO_TARGET{32,64}` stuff is reasonable, and CoreCLR has similar to some/all of this.

PR #16487, opened 2019-08-26
[runtime] Convert mono_reflection_create_custom_attr_data_args () to use handles.

vargaz:


filipnavara:
Looks like this broke `mono-coreonly-CI (Build Linux arm64)` builds:

```
../../mono/metadata/icall-def-netcore.h:197:49: error: ‘ves_icall_System_Reflection_CustomAttributeData_ResolveArgumentsInternal’ undeclared here (not in a function)
 ICALL(CATTR_DATA_1, "ResolveArgumentsInternal", ves_icall_System_Reflection_CustomAttributeData_ResolveArgumentsInternal)
                                                 ^
icall-table.c:149:43: note: in definition of macro ‘ICALL’
 #define ICALL(id, name, func) { (gpointer)func },
```

PR #16490, opened 2019-08-26
Allow file:// based Uri's to be used in HttpClient when the BaseAddress is set

steveisok:
Fixes https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/967582

Bumps the HttpClient part in mono/corefx at: https://github.com/mono/corefx/commit/ee3acec16b37bec5016da129e1e3840097ca629b

The test that would have caught this failure was disabled previously via:
https://github.com/mono/mono/commit/2f411750acfd2132f9b4970ef8cf11d9a0f62df9


akoeplinger:
Manually pushed to 2019-08 and -06.

PR #16494, opened 2019-08-26
[corlib] Simplify file: protocol check in Assembly::Load().

nsivov:
Mirrors change made with 068660732fbf10cae688182ef67f72cf80ebb632.


lambdageek:
@monojenkins build failed

PR #16498, opened 2019-08-26
[coop] Allow SETUP_ICALL_FRAME (i.e. a loop) to reference calling function.

jaykrell:
w/o HANDLE_FUNCTION_ENTER or SETUP_ICALL_FUNCTION.

This is based on my read of https://github.com/mono/mono/pull/16487.


jaykrell:
/cc @lambageek something to consider but I guess not really needed.

It's so functions with loops don't need enter/exit, or even prepare.
They'd just use their caller's frame "automatically".

jaykrell:
I have a plan that will require reverting this, but ok for now. The plan might not work out.

Issue #16310, opened 2019-08-17
LLVM: "".AsSpan()[0] doesn't throw IndexOutOfRangeException


EgorBo:
```csharp
char c = "".AsSpan()[0];
```
CoreCLR:
Mono:
```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
```

Mono-LLVM (both AOT and JIT):
```
(no exception was thrown)
```

imhameed:
This is caused by the unconditional use of abcrem when compiling with LLVM; the same problem occurs when using abcrem without LLVM. I should have a PR out for this soon.

PR #16499, opened 2019-08-26
[jit][llvm] Track variable nullness separately from array length ranges/index ranges in abcrem.

imhameed:
"Nullness" is tracked via a separate field that is stored adjacent to
variable/array value ranges; values may be "maybe-null" or "not-null";
the intersection of maybe-null and not-null yields not-null, the union of
maybe-null and not-null yields maybe-null. The existing graph traversal logic
is reused as much as possible, which makes the implementation of this
null value approximation slightly more complicated than it would be
otherwise.

Fixes https://github.com/mono/mono/issues/16310, which was caused by
the use of non-empty array length ranges as a way to encode non-null
values.

This also allows null check elision for zero-sized arrays.

### DIFF
--- a/mono/mini/abcremoval.c
+++ b/mono/mini/abcremoval.c
@@ -182,7 +182,9 @@ print_evaluation_context_status (MonoRelationsEvaluationStatus status) {
 
 static void
 print_evaluation_context_ranges (MonoRelationsEvaluationRanges *ranges) {
-	printf ("(ranges: zero [%d,%d], variable [%d,%d])", ranges->zero.lower, ranges->zero.upper, ranges->variable.lower, ranges->variable.upper);
+	printf ("(ranges: zero [%d,%d] (not-null = %d), variable [%d,%d])",
+		ranges->zero.lower, ranges->zero.upper, ranges->zero.nullness,
+		ranges->variable.lower, ranges->variable.upper);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


lewurm:
@monojenkins build failed

imhameed:
@monojenkins build failed

lewing:
What is the next step here?

imhameed:
I took a look at some of the differences in abcrem elisions between 85265d884cf31f8779daba2d639eb795fbc9bd5e (baseline) and 3df3300ea2fbc00d76c71fde218348dcd10d3ca4 (the latest version of this change).

`mono --llvm --aot out/lib/mono/4.5/mscorlib.dll` with `REPORT_ABC_REMOVAL` = 1 with a baseline copy of mono yields 7737 removals. A patched copy of mono yields 7554 removals. 182 baseline removals are caused by `ldaddr` implying that the register's array range has size at-least-one (i.e. bug #16310); these removals do not occur after this change. My change introduces a few new removals that wouldn't be generated by baseline mono; they're caused by `ldlen` implying not-null.

imhameed:
@monojenkins build failed

vargaz:
So the missing removals were incorrect ?

imhameed:
There were a few other missing removals in your run that were actually correct, but cb7f5baf461eec6e80479f961452453eefc8af9e and 3df3300ea2fbc00d76c71fde218348dcd10d3ca4 fixed that problem. The remaining missing removals are all incorrect.

vargaz:
@monojenkins build failed

PR #16504, opened 2019-08-27
[coop] Convert System_Runtime_Remoting_Messaging_AsyncResult, etc.

jaykrell:
There are several interesting/new aspects here, given the pinned handles.

Mainly, that I just create the handles and do not even store them.
I'm also loose with comparing contexts, and following pointers.
I think it is all correct.

If you store the handles, then you get a warning they aren't used.
So you can take their address, or think about using them.
The first thing you use them for is raw() but it is a slippery slope.
So this PR is very "optimized" or "less churn", but maybe too tricky?

I also optimized handle creation by manually passing a frame pointer, along with skipping a frame.
I think that should be done more often, though you have to watch out for loops, and yet another pattern could be done, where a frame is setup, but TLS is not fetched.


lambdageek:
Superseded by #16526 and future work

jaykrell:
It is not superceded.

PR #16505, opened 2019-08-27
[coop] Volatile handles less well motivated with pinning.

jaykrell:


jaykrell:
I am inclined against this but others *might* want it.

PR #16511, opened 2019-08-27
[wasm] update emscripten to 1.38.42

lewing:


lewing:
bumped to .43 instead

PR #16524, opened 2019-08-27
[netcore] Add LLVM to nuget

EgorBo:
I suggest to keep the runtime-with-llvm with regular runtime in the same nuget package, thus the package contains:
```
System.Private.CoreLib.dll (~3mb)
libmonosgen-2.0.dylib  (~4mb)
libmonosgen-2.0-llvm.dylib (~33Mb)
```
There is a case when we need both runtimes: LLVM AOT (we compile a self-contained app ahead of time with help of LLVM and then put the 4mb runtime there since we don't need LLVM anymore after).

Also, we need to setup a separate CI pipeline for LLVM JIT (at least for System.Runtime.Tests)


filipnavara:
I like the fact that I would just be able to download a precompiled runtime version with LLVM. It can save me some 20 minutes on automated builds.

I gave it some second thought and it made me wonder whether loadable LLVM would work here better though. I didn't actually try it yet but it would likely make the deployment slightly easier.

scripts/ci/pipeline-netcore-runtime.yml
```diff
@@ -73,7 +74,9 @@ stages:
 
         - bash: |
             cd netcore
-            ./build.sh -c $(_BuildConfig)
+            ./build.sh -c $(_BuildConfig) --llvm --rebuild
+            cp ../mono/mini/.libs/libmonosgen-2.0.so ../mono/mini/.libs/libmonosgen-2.0-llvm.so
```
marek-safar: This should be a new build step at least but we should probably clone the whole lane to make the execution of the tests easier

netcore/runtime-llvm.nuspec
```diff
@@ -0,0 +1,16 @@
...
+        <version>$VERSION$</version>
+        <description>Internal implementation package not meant for direct consumption.  Please do not reference directly. 
+The Mono runtime, and the base library, called mscorlib. It includes the garbage collector, JIT compiler, base .NET data types and many low-level classes.</description>
+        <authors>Microsoft</authors>
+        <projectUrl>https://www.mono-project.com/</projectUrl>
+    </metadata>
+    <files>
+        <file src="System.Private.CoreLib\bin\$COREARCH$\System.Private.CoreLib.dll" target="runtimes\$RID$\native" />
+        <file src="..\mono\mini\.libs\$PLATFORM_AOT_PREFIX$monosgen-2.0$PLATFORM_AOT_SUFFIX$" target="runtimes\$RID$\native" />
+        <file src="..\mono\mini\.libs\$PLATFORM_AOT_PREFIX$monosgen-2.0-llvm$PLATFORM_AOT_SUFFIX$" target="runtimes\$RID$\native" />
```
marek-safar: I think producing separate package would be better at this point

EgorBo: Yeah we had a meeting where we decided to make a separate package too

scripts/ci/pipeline-netcore-runtime.yml
EgorBo: that was just an experiment to see the CI results, we indeed need a separate lane for llvm

netcore/metapackage-llvm.nuspec
```diff
@@ -0,0 +1,16 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>Microsoft.NETCore.Runtime.MonoLLVM</id>
+        <version>$VERSION$</version>
+        <description>Internal implementation package not meant for direct consumption.  Please do not reference directly. 
+The Mono runtime, and the base library, called mscorlib. It includes the garbage collector, JIT compiler, base .NET data types and many low-level classes.</description>
+        <authors>Microsoft</authors>
+        <projectUrl>https://www.mono-project.com/</projectUrl>
+        <dependencies>
+            <dependency id="runtime.linux-x64.Microsoft.NETCore.Runtime.MonoLLVM" version="$VERSION$" />
```
filipnavara: s/linux-x64/$RID$/ ?

EgorBo: @filipnavara only linux will be supported for now :)

filipnavara: Is there any reason for that? I'd consume the packages on macOS if they were available. Besides, the other .nuspec has $RID$ already.

filipnavara: Oh, wait, the metapackage should reference all the available platforms, right? (embarrased look)

EgorBo: exactly, I'll add macos later (windows is not supported at all)

EgorBo:
@marek-safar @steveisok can this be merged?
It adds `Linux x64 LLVM` lane. 
MacOS will be added later - currently `--enable-llvm` on macOS has to build the whole LLVM from sources (Linux downloads pre-built binaries).
Also, tests are currently disabled for LLVM - we need to fix RemoteTestExecutor-based tests (they are fine on macOS-LLVM tho).

EgorBo:
@monojenkins build failed

PR #16530, opened 2019-08-28
[interp] Asserts pessimize clang. Removing saves 16 bytes on Linux/amd64/clang.

jaykrell:
Paired with https://github.com/mono/mono/pull/16456, saves 16 bytes on Linux/amd64/clang.
No change with gcc.


lewurm:
@monojenkins squash

PR #16531, opened 2019-08-28
[interp] Possible code size savings via goto instead of macros.

jaykrell:
Possibly the compilers do this already.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -299,7 +299,7 @@ ves_real_abort (int line, MonoMethod *mh,
 #define ves_abort() \
 	do {\
 		ves_real_abort(__LINE__, frame->imethod->method, ip, frame->stack, sp); \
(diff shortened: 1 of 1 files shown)


mono/mini/interp/interp.c
```diff
@@ -6583,6 +6579,18 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 
 	g_assert_not_reached ();
 
+abort_common:
+	THROW_EX (mono_get_exception_execution_engine (NULL), ip);
+null:
```
lewurm: `null` is problematic? Anyway I would suggested to prefix every label here with something like `ex_`. Looks good otherwise.

lewurm:
@monojenkins squash

lewurm:
Please do not merge manually, I'm trying to fix @monojenkins's automerge capability and using this as a test subject.

Issue #14828, opened 2019-06-06
[netcore]Implement Monitor.LockContentionCount


steveisok:
This may or may not help, but here's the link to coreclr's implementation

https://github.com/dotnet/coreclr/blob/82bd51f001c8df3def3b327bb31fc7611539cbd8/src/vm/threads.h#L3966

filipnavara:
And CoreRT implementation (managed code) for reference: https://github.com/dotnet/corert/pull/7066

filipnavara:
Returning 0 allows BDN to proceed as a temporary measure. It looks like implementing the counter should not be too hard though.

lambdageek:
We're going to try to schedule @alexischr to do this follow-up work: https://github.com/mono/mono/pull/16538#issuecomment-529388153:
> Unfortunately the [test](https://github.com/dotnet/corefx/blob/34fe566b6f6bd739a876e7e1dc48cf631647e5d7/src/System.Threading/tests/MonitorTests.netcoreapp.cs) would not work yet. Performance counters are globally disabled in netcore build in configure script. We'd either need to re-enable it or offer an alternate implementation here.

PR #16538, opened 2019-08-28
[netcore] Add Monitor.LockContentionCount icall

filipnavara:
Contributes to https://github.com/mono/mono/issues/14828.

The PR just wires up the Monitor.LockContentionCount API to an icall. The actual implementation is not done yet.

I wanted to reuse the existing `thread_contentions` performance counter but all performance counters unconditionally disabled on netcore builds. Also, the counter is 32-bit only while the managed API returns 64-bit values.

The CoreCLR implementation uses a neat trick to reduce the cost of counting. It stores a per-thread contention counter (eg. in `MonoInternalThread` field) and a global overflow counter. Up to 2^32 contentions could be recorded in the per-thread field with simple volatile incrementation. Once an overflow happens a slow path is taken that updates a global overflow counter. The `LockContentionCount` API enumerates all threads and adds up together all the per-thread values along with the global overflow counter. I tried to reimplement that in Mono but I didn't find any way to enumerate all threads and get `MonoInternalThread` from them.

### DIFF
--- a/mono/metadata/monitor.c
+++ b/mono/metadata/monitor.c
@@ -1438,3 +1438,15 @@ ves_icall_System_Threading_Monitor_Monitor_Enter (MonoObject *obj)
 {
 	mono_monitor_enter_internal (obj);
 }
+
+#if ENABLE_NETCORE
+gint64
+ves_icall_System_Threading_Monitor_Monitor_LockContentionCount (void)
+{
+#ifndef DISABLE_PERFCOUNTERS
+	return mono_perfcounters->thread_contentions;
+#else
+	return 0;
+#endif
+}
+#endif
--- a/mono/metadata/icall-def-netcore.h
+++ b/mono/metadata/icall-def-netcore.h
@@ -487,7 +487,7 @@ ICALL(MONIT_3, "Monitor_pulse_all", ves_icall_System_Threading_Monitor_Monitor_p
 ICALL(MONIT_4, "Monitor_test_owner", ves_icall_System_Threading_Monitor_Monitor_test_owner)
 ICALL(MONIT_5, "Monitor_test_synchronised", ves_icall_System_Threading_Monitor_Monitor_test_synchronised)
 ICALL(MONIT_7, "Monitor_wait", ves_icall_System_Threading_Monitor_Monitor_wait)
-ICALL(MONIT_8, "get_LockContentionCount", ves_icall_System_Threading_Monitor_Monitor_LockContentionCount)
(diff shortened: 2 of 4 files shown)


filipnavara:
I guess I need to bump `MONO_CORLIB_VERSION`, right?

filipnavara:
In general, `MONO_CORLIB_VERSION` is also used on netcore builds, but there is no monolite bootstrapping issue that would force me to use it.

mono/metadata/monitor.c
```diff
@@ -1438,3 +1438,15 @@ ves_icall_System_Threading_Monitor_Monitor_Enter (MonoObject *obj)
 {
 	mono_monitor_enter_internal (obj);
 }
+
+#if ENABLE_NETCORE
+gint64
+ves_icall_System_Threading_Monitor_Monitor_LockContentionCount (void)
+{
+#ifndef DISABLE_PERFCOUNTERS
```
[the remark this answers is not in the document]
filipnavara: That is preexisting code that is already compiled-out in this way.

mono/metadata/monitor.c
```diff
@@ -1438,3 +1438,15 @@ ves_icall_System_Threading_Monitor_Monitor_Enter (MonoObject *obj)
 {
 	mono_monitor_enter_internal (obj);
 }
+
+#if ENABLE_NETCORE
```
EgorBo: nit: `#ifdef` (we have a mix of `#if ENABLE_NETCORE` and `#ifdef ENABLE_NETCORE` makes sense to unify it

filipnavara: Yeah, I wanted to do that... but then I reverted the whole experiment and lost this small change. :)

I noticed that there's a mix already. Unless I will need to amend the PR due to some other change I'd prefer to do the `#if`/`#ifdef` unification in separate PR once there is agreement on the preferred way.

filipnavara: @jaykrell I don't disagree, in general. In this specific instance it is really `#if` vs `#ifdef` though. The netcore version is significantly different from the regular runtime. It is not just an optional feature where we can easily rely on compiler dead code elimination and regular `if`.

vargaz:
@monojenkins build failed

marek-safar:
@steveisok could we enable corefx tests for this?

filipnavara:
@marek-safar Unfortunately the [test](https://github.com/dotnet/corefx/blob/34fe566b6f6bd739a876e7e1dc48cf631647e5d7/src/System.Threading/tests/MonitorTests.netcoreapp.cs) would not work yet. Performance counters are globally disabled in netcore build in `configure` script. We'd either need to re-enable it or offer an alternate implementation here.

PR #16548, opened 2019-08-28
Build pkg-config as 64bit only, libtiff as universal and remove gtk-quartz-engine.py

akoeplinger:
See https://github.com/mono/bockbuild/pull/116


akoeplinger:
@monojenkins build pkg

akoeplinger:
@monojenkins build pkg

akoeplinger:
@monojenkins build pkg

PR #16552, opened 2019-08-28
Turned the compiler server off on Windows because Named Pipes aren't fully working

steveisok:


steveisok:
I don't think this saves us a thing time wise, but technically it should be off on windows.

jaykrell:
Surely named pipes are easy to get working on Windows?
Granted that doesn't mean the server is fast.

steveisok:
@monojenkins build failed

filipnavara:
Hmm, the named pipe code is from CoreFX and I am pretty damn sure it works there for the compiler server. In fact, that was one of the motivations for adapting the code. Is there some issue filed for the problem?

steveisok:
@filipnavara The main issue is mono's WindowsIdentity is incomplete. 

https://github.com/mono/corefx/blob/ee3acec16b37bec5016da129e1e3840097ca629b/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Windows.cs#L51-L72

And even if you comment that block out, it still throws. 

I haven't created an issue for it yet.

steveisok:
Additionally, the compiler server is never actually started because MONO_PATH is passed as cygwin paths.

lewing:
@monojenkins build failed

steveisok:
@monojenkins build failed

PR #16553, opened 2019-08-28
Remove MAX_PATH usage in w32process

### LINKED ISSUE #14730: Assertion at /Users/builder/jenkins/workspace/archive-mono/2019-02/mac/release/mono/utils/strenc.c:184, condition `utf8!=NULL' not met

CoffeeFlux:
Fixes #14730

Not totally sure my Windows code is correct here, so it's a draft until I can look at CI results.

### DIFF
--- a/mono/metadata/w32process-unix.c
+++ b/mono/metadata/w32process-unix.c
@@ -919,6 +919,11 @@ mono_w32process_module_get_filename (gpointer handle, gpointer module, guint32 *
 	gchar *path;
 	gunichar2 *proc_path;
 
+	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_PROCESS, "%s: Getting module file name, process handle %p module %p " G_GUINT32_FORMAT,
+	            __func__, handle, module);
+
+	*len = 0;
+
 	pid = mono_w32process_get_pid (handle);
 	if (pid == 0)
 		return NULL;
@@ -950,6 +955,8 @@ mono_w32process_module_get_name (gpointer handle, gpointer module, guint32 *len)
 
 	mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_PROCESS, "%s: Getting module base name, process handle %p module %p " G_GUINT32_FORMAT,
 		   __func__, handle, module);
+
+	*len = 0;
 
 	if (!mono_w32handle_lookup_and_ref (handle, &handle_data)) {
 		mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_IO_LAYER_PROCESS, "%s: unknown handle %p", __func__, handle);
--- a/mono/metadata/w32process-unix.c
+++ b/mono/metadata/w32process-unix.c
@@ -911,82 +911,57 @@ mono_w32process_try_get_modules (gpointer handle, gpointer *modules, guint32 siz
 	return TRUE;
 }
 
-guint32
-mono_w32process_module_get_filename (gpointer handle, gpointer module, gunichar2 *basename, guint32 size)
+gunichar2 *
+mono_w32process_module_get_filename (gpointer handle, gpointer module, guint32 *len)
 {
-	gint pid, len;
-	gsize bytes;
+	gint pid;
+	gsize bytes = 0;
 	gchar *path;
 	gunichar2 *proc_path;
 
-	size *= sizeof (gunichar2); /* adjust for unicode characters */
-
-	if (basename == NULL || size == 0)
-		return 0;
-
 	pid = mono_w32process_get_pid (handle);
+	if (pid == 0)
+		return NULL;
 
 	path = mono_w32process_get_path (pid);
 	if (path == NULL)
-		return 0;
+		return NULL;
 
 	proc_path = mono_unicode_from_external (path, &bytes);
+
+	*len = bytes / 2;
+
 	g_free (path);
-
-	if (proc_path == NULL)
-		return 0;
-
-	len = (bytes / 2);
-
-	/* Add the terminator */
-	bytes += 2;
-
-	if (size < bytes) {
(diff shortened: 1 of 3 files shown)


jaykrell:
```
gboolean
mono_get_module_filename (HMODULE mode, gunichar2** pstr, guint32* plength)
{
  gunichar* str= = NULL;
  guint32 capacity = 64;
  guint32 length = 0;
  gboolean success = FALSE;

  while (true)
  {
     length = 0;
     str = g_new (gunichar2, capacity);
     if (!str)
        break;
     length = GetModuleFileName (mod, str, capacity);
     success = length && length < capacity;
     if (success)
        break;
     g_free(str); // error or too small
     str = NULL; 
     capacity *= 2;
     if (!length) // error 
        break;
  }
  *pstr = str;
  *plength = length;
  return success;
}
```
?

One call to each function.
One write through each out parameter.
One exit path.
This function can be slow, but it should also contribute minimal size to the overall runtime.

CoffeeFlux:
Oh also, @rolfbjarne how far will you need this backported?

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

mono/metadata/w32process.c
```diff
@@ -561,9 +595,9 @@ ves_icall_System_Diagnostics_Process_GetModules_internal (MonoObjectHandle this_
 	return_val_if_nok (error, NULL_HANDLE_ARRAY);
 
 	for (i = 0; i < module_count; i++) {
-		if (mono_w32process_module_get_name (process, mods [i], modname, MAX_PATH)
-			 && mono_w32process_module_get_filename (process, mods [i], filename, MAX_PATH))
-		{
+		modname = mono_w32process_module_get_name (process, mods [i], &modname_len);
+		filename = mono_w32process_module_get_filename (process, mods [i], &filename_len);
+		if (modname && filename) {
```
jaykrell: If modname is null, should you call get_filename?

mono/metadata/w32process.c
```diff
@@ -561,13 +595,19 @@ ves_icall_System_Diagnostics_Process_GetModules_internal (MonoObjectHandle this_
...
-			return_val_if_nok (error, NULL_HANDLE_ARRAY);
+			if (!is_ok (error)) {
+				g_free (modname);
+				g_free (filename);
+				return NULL_HANDLE_ARRAY;
+			}
 			MONO_HANDLE_ARRAY_SETREF (temp_arr, num_added++, module);
 		}
+		g_free (modname);
+		g_free (filename);
```
jaykrell: You don't need to free twice.
Instead:
```
  if (!is_ok (error))
    break;
  array setref
 }
g_free twice
if (!is_ok())
  return
```

mono/metadata/w32process-unix.c
```diff
@@ -911,82 +911,64 @@ mono_w32process_try_get_modules (gpointer handle, gpointer *modules, guint32 siz
...
+		return NULL;
 
 	proc_path = mono_unicode_from_external (path, &bytes);
-	g_free (path);
-
-	if (proc_path == NULL)
-		return 0;
 
-	len = (bytes / 2);
+	*len = bytes / sizeof (gunichar2);
```
jaykrell: Aside: Windows code used to always `<< 1` and `>> 1` but that fell out of fashion.

rolfbjarne:
@CoffeeFlux back porting back to 2019-08 would be nice.

mono/metadata/w32process.c
CoffeeFlux: I don't think it makes much of a difference, given that's not a common scenario.

mono/metadata/w32process.c
CoffeeFlux: I don't think the code you suggested will work since they have to be freed each iteration of the loop, but you're right that this can be cleaned up a bit to only free once.

jaykrell: The old code stopped at the first error and returned.
So either undo this part, sorry, or try for single point of exit.

CoffeeFlux: Check my latest commit and see if you think that's fine, I tried to clean it up.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:

mono/metadata/w32process.c
jaykrell: I don’t think so but hard to see in github on phone. The old code stopped the loop upon error. This does not seem to.

jaykrell: Oh oops probably. Just hard to see indent on phone.

jaykrell:
I suggest you take helper function(s) from other PR. Backport should not be a problem.

Issue #16018, opened 2019-08-04
Tiered compilation for Mono runtime


migueldeicaza:
This issue is intended to be a location where we can discuss the specifics needed for tiered compilation in Mono.

# Background

Mono today offers two code generation engines, one that produces code very quickly, but performs poorly in various CPU intensive benchmarks and one powered by LLVM that is very slow, but incredibly efficient code.  Additionally, Mono can either just-in-time compile, or ahead-of-time compile.    On the desktop, ahead-of-time compilation is mostly used to reduce startup time, and reduce memory usage, and rarely to produce better code - as AOT code pays a price that JIT code does not pay.  Additionally, AOT can be a slow process.

Mobile users for example use the fast compiler during development, and use LLVM for Release builds.   But desktop users tend to not do this, as the cost of using LLVM for running their applications is too high.

CoreCLR calls the code generated quickly with low quality the Tier0, and the code generated with a higher optimization level Tier1.  It makes sense to embrace the same naming convetion.

One useful scenario to support would be tiered compilation.   One where we use the Mono built-in code generator for the startup sequence, and based on a heuristic that determines just how important a method is, we recompile the code with LLVM.

One specific user that could benefit from this is VSMac, which currently only uses the quick code generator, where using LLVM on startup is deemed too slow.   This would also help the Unity editor ([an example showing Unity](https://tirania.org/blog/archive/2018/Apr-11.html))

# Strategies

Tiered compilation is made up of a couple of pieces:

* Instrumenting the slow code to track candidates to be recompiled.
* Heuristics that determine when a method is suitable for recompilation.
* The actual recompilation changes.
* Redirecting any calls to the original low-quality generated code to the high-quality generated code.

While there are many different approaches for instrumenting the code, and many difference heuristics that can be used, we could start by embracing the current  [heuristics and processes used by the CoreCLR](https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/tiered-compilation.md), saving us design and experimentation time.   We could later adjust those parameters to better suit the Mono workloads.

While there are some very advanced use cases for tiered compilation that allow replacements of methods on the stack.   For interactive applications like VSMac, where control goes back to a mainloop to process events, the need to replace live methods is not as important.  Because even the mainloop is ran by unmanaged code.

The CoreCLR implementation replaces methods during a GC pause, ensuring that it can patch existing code without requiring atomic changes.

# Plan

- [x] Implement instrumentation that counts the number of invocations to a method.
[...]

PR #16556, opened 2019-08-29
Indirect calls for easier atomic patching and some day read only code.

jaykrell:
I don't think direct calls are worthwhile.

They are never used on Mac or Windows. Only Linux.

When Xen is absent.
So the extra map32 and Xen checks.
Map32 recently having caused problems on Apple systems.

The instructions need to be and are padded for alignment.

Is the check for upper bits being zero sufficient?
It doesn't mean distance is within 2GB.

They aren't necessarily faster than direct.

nops consume processor resources.

Indirect is a good approach also I think to fix
 PowerPC, which isn't atomic and has known
 but not understood consistent failures.

Eventually, if the thunks are moved further away,
code pages could become readonly once full.
Granted, direct calls could be patched with an mprotect cycle also
with synchronization.

Separately, indirect need not use r11,
but can be thunks like this PR does.


jaykrell:
@monojenkins build failed

jaykrell:
Hm, Mac worked locally.

jaykrell:
In one of the crashes, instruction pointer was off by one, leading to sigill.

jaykrell:
Local test was with configure disable-system-aot.
That probably was it.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Any chance of this going in?

It simplifies the code a fair amount.
  Unifies more of the platforms -- no more Linux-specificity.
  Conserves a register in some paths (r11).
And *might* make it faster. Reports vary.
And is a hypothetical step towards readonly code pages.
And provides a model for PowerPC to follow, which kinda needs it.
But I've said all this before.

lewurm:
My intuition tells me direct calls are worth it. Please backup your PR with some actual benchmark numbers.

jaykrell:
https://github.com/mono/mono/issues/16018#issuecomment-521097802

"In the past I believe some other developers have looked at call indirect vs call + jmp with patchable target. Results varied based on hardware as you might expect, I think indirect call was concluded to be a little faster typically."


https://github.com/mono/mono/issues/16018#issuecomment-521107149

Also, it depends on how many indirect calls you got on the hot path. If there is less than 10, the indirect call is faster thanks to indirect call predictor. More than 10 and less than 1000, direct call to direct jump is faster because of the indirect call predictor gets filled in and does not help anymore. More than 1000, the performance becomes equal. L0/L1 caches are filled in and the performance is dominated by cache misses. 10 and 1000 thresholds vary based on hardware.

RIP relative call/jmp also has a one cycle penality I recall reading, but have to find it.
One would think the RIP relative+indirect has the same penality. I'll have to find the reference.

But there is also that nops require decoding.
I think we output multiple nops instead of multi byte nops.

And read only pages might help cache coherence. But if that does matter, it won't for a while, i.e not with this PR.

I'll try to collect some data.

SamMonoRT:
@jaykrell - any benchmarking data to back these changes ?

jaykrell:
No. Are there any mono has already?
How about taking https://github.com/mono/mono/pull/16699 instead/first? It is a subset of this. It still uses direct calls, on Linux, sometimes, but the indirect calls through a register are replaced by indirect calls through memory.
More so, PowerPC should adopt this approach, to fix race conditions, and resolve unresolved frequently seen bugs in the direct call patching.

jaykrell:
Not having to align things also simplifies instruction sizing.

akoeplinger:

PR #16559, opened 2019-08-29
[WinForms] Fix UIA event on node insertions into a `TreeView` object

NVoronchev:
The UIA event doesn't emited by means of insertion a new `TreeNode` to a `TreeNodeCollection` of the respective `TreeNode`.

### DIFF
--- a/mcs/class/System.Windows.Forms/System.Windows.Forms/TreeNodeCollection.cs
+++ b/mcs/class/System.Windows.Forms/System.Windows.Forms/TreeNodeCollection.cs
@@ -91,9 +91,10 @@ namespace System.Windows.Forms {
 				if (index < 0 || index >= Count)
 					throw new ArgumentOutOfRangeException ("index");
 				
-				OnUIACollectionChanged (CollectionChangeAction.Remove, nodes [index]);
+				var removedNode = nodes [index];
 				nodes [index] = value;
-				
+
+				OnUIACollectionChanged (CollectionChangeAction.Remove, removedNode);
 				SetupNode (value);
 			}
 		}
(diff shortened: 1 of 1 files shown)


akoeplinger:
@monojenkins build

mcs/class/System.Windows.Forms/System.Windows.Forms/TreeNodeCollection.cs
```diff
@@ -90,8 +90,11 @@ internal TreeNodeCollection (TreeNode owner)
 			set {
 				if (index < 0 || index >= Count)
 					throw new ArgumentOutOfRangeException ("index");
-				SetupNode (value);
+				
+				OnUIACollectionChanged (CollectionChangeAction.Remove, nodes [index]);
 				nodes [index] = value;
```
filipnavara: The order of operations is reversed here. While the UIA API is not well documented this seems like a breaking change. 

```csharp
var removed = nodes [index];
nodes [index] = value;
OnUIACollectionChanged (CollectionChangeAction.Remove, removed);
```

NVoronchev: @filipnavara, thank you! You are right, one should raise "remove"- event after the node is removed. How do you like this approach:
```
    set {
        if (index < 0 || index >= Count)
            throw new ArgumentOutOfRangeException ("index");
        
-       OnUIACollectionChanged (CollectionChangeAction.Remove, nodes [index]);
-       nodes [index] = value;
-       
-       SetupNode (value);
+       RemoveAt (index, false);
+       Insert (index, value);
    }
```

filipnavara: That will cause unnecessary array resizing which could be costly. I'd go with the less disruptive change.

NVoronchev: Please, find the PR update

akoeplinger:
@monojenkins build

PR #16569, opened 2019-08-29
[configure] remove autodection of llvm-config in $PATH

lewurm:
Instead `./autogen.sh --with-llvm=<path>` should be used if a build from `external/llvm` is not preferred.

The old behavior was confusing when a `llvm-config` is on `$PATH`, which can be the case when you have LLVM packages installed on a Linux distribution. It picked that up when `./autogen.sh --enable-llvm` was used.

/cc @brianrob


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

lewurm:

PR #16571, opened 2019-08-29
[Mono.Posix] Add FcntlCommand.F_OFD_{GETLK,SETLK,SETLKW}

steffen-kiess:
Add F_OFD_GETLK, F_OFD_SETLK and F_OFD_SETLKW to FcntlCommand. These
commmands are similar to F_GETLK, F_SETLK and F_SETLKW but operate on a
per-open-file-description lock instead of a per-process lock and avoid the
problem that the locks are released when close(dup(fd)) is called. See
https://lwn.net/Articles/586904/ for more information.

This commit also reruns create-native-map but keeps manual changes from
ed892ccf27849c082ce6ca46fa8b96d86ca7c329,
b522eab5ff5466debaacf9e971e26cfc464ebba5 and
22b6b9581418260397b701c17b16c3eb55136de7.

In [create-native-map.diff.txt](https://github.com/mono/mono/files/3557077/create-native-map.diff.txt) you can see the difference between the commited `map.[ch]` and the one `create-native-map.exe` wants to create:
- Manually inserted changes for NetBSD (`MAYMOVE` is turned into a negated `FIXED`)
- `Mono_Posix_ToSockaddrIn6()` and friends get skipped for `HOST_WIN32` (not sure why, when `sockaddr_in6` exists then `Mono_Posix_ToSockaddrIn6()` should work)
- At the bottom of `map.h` some functions are reordered and the parameters change (`const char*` vs. `char*`)

Running create-native-map.exe also moved some O_NOATIME-related stuff (from 8b9033e4d24115190c06f775d18ef3a40cbca876) to the correct position and added it to Mono_Posix_ToOpenFlags(), which I kept in the commit.

### DIFF
--- a/mcs/class/Mono.Posix/Mono.Posix_test.dll.sources
+++ b/mcs/class/Mono.Posix/Mono.Posix_test.dll.sources
@@ -9,6 +9,7 @@ Mono.Unix/UnixPathTest.cs
 Mono.Unix/UnixSignalTest.cs
 Mono.Unix/UnixUserTest.cs
 Mono.Unix.Android/TestHelper.cs
+Mono.Unix.Native/OFDLockTest.cs
 Mono.Unix.Native/RealTimeSignumTests.cs
 Mono.Unix.Native/SocketTest.cs
 Mono.Unix.Native/StdlibTest.cs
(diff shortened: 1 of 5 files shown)


akoeplinger:
@monojenkins build

akoeplinger:
@monojenkins commit apidiff

PR #16585, opened 2019-08-30
Test LLVM 9.0 JIT on CI

filipnavara:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,7 +61,7 @@
 [submodule "external/llvm"]
 	path = external/llvm
 	url = git://github.com/mono/llvm.git
-	branch = release_60
+	branch = release_90
 [submodule "external/helix-binaries"]
 	path = external/helix-binaries
 	url = git://github.com/mono/helix-binaries.git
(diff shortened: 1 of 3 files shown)


filipnavara:
Looks like the whole test suite finished with pretty much the same failures as with LLVM 6.0 (could be related to the abcrem issue).

What were the blockers for moving to LLVM 9.0? /cc @imhameed @vargaz

imhameed:
I think these are blockers for moving to LLVM 9.0:
1. [`MonoException`](https://github.com/mono/llvm/blob/release_90/lib/CodeGen/AsmPrinter/MonoException.cpp) frequently causes assertion failures in LLVM 9 when LLVM debug assertions are enabled. Disabling LLVM assertions sometimes "works", and sometimes causes other failures or crashes later on during JIT compilation and, less frequently, during AOT compilation. This is pretty gross. To be fair, our current use of LLVM 6 also suffers from the same problem, but is mostly masked because we don't build LLVM with assertions enabled by default.
2. JIT + LLVM 9 is much more unstable than AOT + LLVM 9. But it's hard to disentangle this from issue 1; every JIT crash I've seen so far is related to `MonoException`.
3. The [release_60](https://github.com/mono/llvm/commits/release_60) branch has a few patches and bug fixes that haven't yet been cherry-picked into the [release_90](https://github.com/mono/llvm/commits/release_90) branch.

filipnavara:
Tracked in #16893 now.

PR #16596, opened 2019-08-30
[debugger][android] Fixing single step inside Exception on Android

### LINKED ISSUE #14772: Null Pointer Exception after debugger pauses and then continues
### Steps to Reproduce
Start debugging the attached test case either from Visual Studio or on the command line using sdb.
Tap the Button button in the app.
Once the debugger pauses on the exception, continue running the app from the debugger.
### Current Behavior
With Xamarin.Android 9.3.0.19 (3f5ec6d95c3...), the application process exits uncleanly due to a native segmentation fault in mono_method_get_class().
image

[*] Android 9.0 (API level 28) x86 emulator Nexus 6
[ ] macOS
[ ] Linux
[ ] Windows

[NestedExceptionDebugCrash.zip](https://github.com/mono/mono/files/3248735/NestedExceptionDebugCrash.zip)

thaystg:
Fixing single step inside Exception on Android...
The method is null because on ss_create_init_args the IL is -1 when calling method Mono_UnhandledException_internal what I think it's expected, because of that I ignore the single step on android if there is no method. (debugger-engine.c)
After that the sequence point cannot be found because it's calling a method of class DynamicMethodNameCounter what I think it's expected, because of that I remove the assert on android. (debugger-agent.c)
Fixes #14772

### DIFF
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -4970,7 +4970,7 @@ ss_create_init_args (SingleStepReq *ss_req, SingleStepArgs *args)
 #if !defined(HOST_ANDROID) && !defined(TARGET_ANDROID)					
 				g_assert (found_sp);				
 #else
-				if (found_sp) //it will not find the next sp if it is a "DynamicMethodNameCounter" from Android.Runtime
+				if (found_sp) //it will not find the sp if it is a method of class Android.Runtime.DynamicMethodNameCounter
 #endif				
 					method = frame->de.method;
 			}
--- a/mono/mini/debugger-agent.c
+++ b/mono/mini/debugger-agent.c
@@ -4967,8 +4967,12 @@ ss_create_init_args (SingleStepReq *ss_req, SingleStepArgs *args)
 				found_sp = mono_find_prev_seq_point_for_native_offset (frame->de.domain, frame->de.method, frame->de.native_offset, &info, &args->sp);
 				if (!found_sp)
 					no_seq_points_found (frame->de.method, frame->de.native_offset);
-				g_assert (found_sp);
-				method = frame->de.method;
+#if !defined(HOST_ANDROID) && !defined(TARGET_ANDROID)					
+				g_assert (found_sp);				
+#else
+				if (found_sp) //it will not find the next sp if it is a "DynamicMethodNameCounter" from Android.Runtime
+#endif				
+					method = frame->de.method;
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


mono/mini/debugger-agent.c
```diff
@@ -4967,8 +4967,12 @@ ss_create_init_args (SingleStepReq *ss_req, SingleStepArgs *args)
 				found_sp = mono_find_prev_seq_point_for_native_offset (frame->de.domain, frame->de.method, frame->de.native_offset, &info, &args->sp);
 				if (!found_sp)
 					no_seq_points_found (frame->de.method, frame->de.native_offset);
-				g_assert (found_sp);
-				method = frame->de.method;
+#if !defined(HOST_ANDROID) && !defined(TARGET_ANDROID)					
+				g_assert (found_sp);				
+#else
+				if (found_sp) //it will not find the next sp if it is a "DynamicMethodNameCounter" from Android.Runtime
```
lambdageek: This comment isn't very clear.  Do you mean this: https://github.com/xamarin/xamarin-android/blob/0ec01060f6bc16c55fdc2e2dae9c21b34a94e7bb/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs#L52 ?

thaystg: The name of the class of the method that is on stack is DynamicMethodNameCounter and the name of the method is in this case 5. I'm not sure if this is the case...

lambdageek: Yea, I think it's that line.  Note that the "owner" parameter is `typeof (DynamicMethodNameCounter)`: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.dynamicmethod.-ctor?view=netframework-4.8#System_Reflection_Emit_DynamicMethod__ctor_System_String_System_Reflection_MethodAttributes_System_Reflection_CallingConventions_System_Type_System_Type___System_Type_System_Boolean_

And then in `reflection_create_dynamic_method` in Mono we use that `owner` parameter to set the `klass` of the method.  And I think the debugger loses track of it.

thaystg: Thanks for the explanation! :)

lambdageek:
I think we should backport this to 2019-08 to get a simple workaround for now, but I think the real fix is to figure out how to single step throw `DynamicMethod`s that have an `owner`

steveisok:
I'm approving and merging b/c @lambdageek  approved the backport.

PR #16601, opened 2019-08-31
[RFC] Assert in get_call_target.

jaykrell:
This fails. But it should work? `method_addresses` being assembled counter to expectations? Does it matter? Need to coerce the assembler or use `db`?


akoeplinger:

PR #16602, opened 2019-08-31
Remove Xen and Mmap32 stuff.

jaykrell:
- Xen is test matrix expansion. Is it tested and worthwhile?
- Mmap32 is absent on Apple and Windows systems. Is Linux variation worthwhile?
  And it is affected by Xen.

There are contradictory comments about this stuff as well.
We'll see how it goes.


jaykrell:
@monojenkins build failed

jaykrell:
This overlaps a lot with https://github.com/mono/mono/pull/16556
and this one (https://github.com/mono/mono/pull/16602) should go first since it does less.

jaykrell:
@monojenkins build failed

vargaz:
map32 allows us to make direct calls, this is why its there.

jaykrell:
Are the direct calls woth this trouble and platform variation and test matrix expansion?
You know, Mac and Windows doesn't do this, Android unclear (is it Linux?).

I suggest those be removed too:
 1. The alignment logic (and extra space) can be removed.
 2. The code pages might become readonly eventually (when the thunks are moved to separate pages and the code pages fill up).
 3. As a prelude to fixing PowerPC patching problems.
 4. Are they so slow? Reportedly sometimes they are faster.

https://github.com/mono/mono/pull/16556 is a superset of this PR that removes Xen, Mmap32, and direct.

jaykrell:
Also, map32, and the upper bits being 0, does not ensure the addresses are within 2GB of each other. Does it really work?

jaykrell:
Windows and Mac do not use this code. I don't think it worth having for Linux.

PR #16609, opened 2019-08-31
[netcore] Tweak inlining heuristic for memcpy

filipnavara:
The overall `memcpy`/`memset` situation could use a better overhaul when compiling with LLVM since we can convert directly to intrinsics and the optimizer could make better choices and automatic vectorization for us.

This tweaks single heuristic that is affecting codegen for `System.Buffer.Memmove` (and by proxy `Buffer.BlockCopy` and some string operations). The `Memmove` code uses an `HAS_CUSTOM_BLOCKS` optimization that tries to copy memory block by 16-byte or 16-dword chunks. This hit the previous limit for inlining and resulted in a call being emitted instead of the intended inline code.

Furthermore, `Buffer.Memmove` got compiled to something that uses `string.memcpy`. `string.memcpy` is implemented as wrapper around `Buffer.Memcpy`, and `Buffer.Memcpy` calls `Buffer.Memmove`. The infinite recursion seems to be prevented only because small blocks (<= 16 bytes) are treated specially in `Buffer.Memmove`.

### DIFF
--- a/mono/mini/memory-access.c
+++ b/mono/mini/memory-access.c
@@ -16,7 +16,11 @@
 #include "ir-emit.h"
 #include "jit-icalls.h"
 
+#ifdef ENABLE_NETCORE
+#define MAX_INLINE_COPIES 16
+#else
 #define MAX_INLINE_COPIES 10
+#endif
 #define MAX_INLINE_COPY_SIZE 10000
 
 void 


filipnavara:

filipnavara:

EgorBo:
@monojenkins build failed

mono/mini/memory-access.c
```diff
@@ -16,7 +16,11 @@
 #include "ir-emit.h"
 #include "jit-icalls.h"
 
+#ifdef ENABLE_NETCORE
+#define MAX_INLINE_COPIES 16
```
marek-safar: Why 16?

filipnavara: I used the number used in Memmove specifically to avoid the recursive call issue. The likely choice for that is that x64 could do that with one unaligned load into xmm register and then one unaligned store from xmm register, effectively moving 16 bytes with two instructions (or multiples of 16 bytes for bigger data types). Mono doesn't do this optimization now but it would be fairly easy to add (and LLVM is already special cased).

filipnavara: Here's a link to the CoreCLR limits: https://github.com/dotnet/coreclr/blob/84d4ed44b5c7f1b3db6e9ae3c71e8e1bb585bbaf/src/jit/target.h#L503

marek-safar: Does it show any improvements in the microbenchmarks?

filipnavara: Yes, it does... and in the generated code that avoids recursive calls. If you are interested in specific number I can trigger the microbenchmark run. Then again, this is not primarily a performance fix as much as it is a correctness fix for the code that depends on this inlining behavior and which is located in the shared CoreLib partition.

PR #16610, opened 2019-08-31
LLVM: Make Buffer.Memmove intrinsic

EgorBo:
[Buffer.MemoryCopy](https://github.com/dotnet/coreclr/blob/d951f5f0ee8862f954c69a4fd9e05184bad15d25/src/System.Private.CoreLib/shared/System/Buffer.cs#L93-L100) calls [Memmove](https://github.com/dotnet/coreclr/blob/d951f5f0ee8862f954c69a4fd9e05184bad15d25/src/System.Private.CoreLib/shared/System/Buffer.cs#L144) which is manually unrolled and calls [__Memmove](https://github.com/dotnet/coreclr/blob/d951f5f0ee8862f954c69a4fd9e05184bad15d25/src/System.Private.CoreLib/shared/System/Buffer.cs#L307-L308) (InternalCall) if array is bigger than [MemmoveNativeThreshold](https://github.com/dotnet/coreclr/blob/52651008a5fb72eb678467cd2eb42aac4e5334e8/src/System.Private.CoreLib/shared/System/Buffer.Unix.cs#L24). I am not sure should we replace the whole `Memmove` with `llvm.memmove.p0i8.p0i8.i64` (see [memmove](https://llvm.org/docs/LangRef.html#llvm-memmove-intrinsic) doc) or just `__Memmove` I need to benchmark it.

```csharp
static unsafe void BufferMemoryCopy(byte* src, byte* dst, ulong len)
{
    Buffer.MemoryCopy(src, dst, len, len);
}
```
Before:
```llvm
define monocc void @"P:BufferMemoryCopy (byte*,byte*,ulong)"(i8* %arg_src, i8* %arg_dst, i64 %arg_len) #0 {
BB0:
  %0 = call i1 @llvm.expect.i1(i1 true, i1 true) ; `-lower-expect` will remove it
  br i1 %0, label %BB9, label %BB10

BB9:                                              ; preds = %BB0
  %1 = load void (i8*, i8*, i64)*, void (i8*, i8*, i64)** @tramp_937, align 8
  notail call monocc void %1(i8* %arg_dst, i8* %arg_src, i64 %arg_len)
  ret void

BB10:                                             ; preds = %BB0
  %2 = load void (i32)*, void (i32)** @tramp_938, align 8
  notail call monocc void %2(i32 43)
  unreachable
}
```
After
```llvm
define monocc void @"P:BufferMemoryCopy (byte*,byte*,ulong)"(i8* %arg_src, i8* %arg_dst, i64 %arg_len) #0 {
BB9:
  call void @llvm.memmove.p0i8.p0i8.i64(i8* %arg_dst, i8* %arg_src, i64 %arg_len, i32 1, i1 false)
  ret void
}
```
godbolt: https://godbolt.org/z/5RY5b1 (clang also emits `1` for alignment argument and `0` for is_volatile). Alignment 0 or 1 means we don't guarantee that both arrays are aligned (and we actually really can't)

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -222,6 +222,17 @@ llvm_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		}
 	}
 
+	if (in_corlib && !strcmp (m_class_get_name (cmethod->klass), "Buffer")) {
+		if (!strcmp (cmethod->name, "Memmove") && fsig->param_count == 3 && fsig->params [0]->type == MONO_TYPE_PTR && fsig->params [1]->type == MONO_TYPE_PTR) {
+			opcode = OP_MEMMOVE;
+			MONO_INST_NEW (cfg, ins, opcode);
+			ins->sreg1 = args [0]->dreg; // i1* dst
+			ins->sreg2 = args [1]->dreg; // i1* src
+			ins->sreg3 = args [2]->dreg; // i32/i64 len
+			MONO_ADD_INS (cfg->cbb, ins);
+		}
+	}
+
 	return ins;
 }
 
--- a/mono/mini/llvm-jit.cpp
+++ b/mono/mini/llvm-jit.cpp
@@ -289,7 +289,7 @@ public:
 		if (opts == NULL) {
 			// FIXME: find optimal mono specific order of passes
 			// see https://llvm.org/docs/Frontend/PerformanceTips.html#pass-ordering
-			opts = " -simplifycfg -sroa -instcombine -gvn";
+			opts = " -simplifycfg -sroa -lower-expect -instcombine -gvn";
 		}
 
 		char **args = g_strsplit (opts, " ", -1);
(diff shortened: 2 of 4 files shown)


EgorBo:
A naive benchmark:
```csharp
using System;
using System.Diagnostics;

class P
{
    static unsafe void Main() => RunBenchmark();

    static readonly int arrayLength = 100;
    static readonly int iterations = 10000000; // static readonly in order to prevent loop unrolling

    [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.NoInlining)]
    static unsafe void RunBenchmark()
    {
        byte[] src = new byte[arrayLength];
        src.AsSpan().Fill(42);

        byte[] dst = new byte[arrayLength];

        Stopwatch sw = new Stopwatch();
        while (true)
        {
            sw.Restart();
            for (int i=0; i<iterations; i++){
                fixed (byte* s = src)
                fixed (byte* d = dst)
                {
                    Buffer.MemoryCopy(s, d, dst.Length, src.Length);
                }
                //still slow:
                //Buffer.BlockCopy(src, 0, dst, 0, src.Length);
            }
            sw.Stop();
            Console.WriteLine("Elapsed: " + sw.ElapsedMilliseconds);
        }
    }
}
```
llvm.memmove is faster for any `arrayLength` than CoreCLR (especially under 1000)

EgorBo:
@monojenkins build failed

PR #16612, opened 2019-08-31
[Coop] Fully inline NULL_HANDLE (do not even access a global variable).

jaykrell:


jaykrell:
@monojenkins build failed

PR #16617, opened 2019-09-01
[netcore] Add intrinsics for Vector.GreaterThanOrEqual/LessThanOrEqual

filipnavara:


filipnavara:
This fixed nearly all performance gaps on `String` class except one `String.Contains` benchmark and the ones related to manipulation using `Buffer.BlockCopy`.

vargaz:
@monojenkins build failed

filipnavara:
I just realized that there are CMP_GE/CMP_LE/CMP_GE_UN/CMP_LE_UN operands so the code could be even simpler. I will fix it to use that. One benefit is that it's easier to use [this trick](https://stackoverflow.com/questions/16204663/sse-compare-packed-unsigned-bytes) with converting to min/max + equals.

PR #16614, opened 2019-09-01
[Coop] ves_icall_property_info_get_default_value.

jaykrell:
This includes resolving a somewhat deep problem involving
a void* that is usually to a int/float, but can be a MonoObject** or at least MonoString**.
An additional MonoStringHandleOut parameter is passed around to address this.
Perhaps it should be MonoObjectHandleOut for more generality.

PR #16620, opened 2019-09-01
[Coop] Debugger single point of exit.

jaykrell:
Extracted from https://github.com/mono/mono/pull/16614.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

CoffeeFlux:
What's the motivation for this? The commit itself seems fine, just want to understand the intent behind it to avoid unnecessary churn.

jaykrell:
It's in the description.
It is so coop can add function_enter/exit in a larger PR.
Also, maybe the err= + goto should just be goto w/ different label.? i.e. to avoid braces and reduce code? I kinda like that, will probably do.

vargaz:
I think it would be easier to extract the stuff which does objref manipulation into separate functions.

jaykrell:
I've done/attempted function splitting recently, to avoid single point of exist, and @lambageek seemed to disagree, so went back to single point of exist. I think it is pretty trivial either way. But yes function splitting is actual smaller diff.

jaykrell:
You don't even really need separate functions for separate purpose, you can just wrap the existing function in a helper, and have everyone calls the helper, with the old name.

jaykrell:
Actually, to make handle allocation be just alloca(fixed size), and never heap, I'd like to keep frames within functions that allocate handles. So that'd encourage a finer grained split. Which we can consider.
Anyway, I'll make a label per error at least, it makes the code a little smaller.

jaykrell:
Is this ok? I think in the short term this is easier than splitting up the functions, and in the long term doesn't make anything worse, if the functions are later split up.

Single point of return is certainly a little tedious to switch to but shouldn't imho be considered difficult or complex or anything.

jaykrell:
@monojenkins build failed

Issue #16623, opened 2019-09-02
Hebrew and other non-Gregorian calendars are broken


filipnavara:
Mono uses embedded culture data in form of `culture-info-tables.h`. The code that populates the managed data from it gets two things wrong:

- It ignores the calendar id and only considers the locale. So, for Hebrew locale and Hebrew calendar you'd still get resources for the Hebrew locale and Gregorian calendar. Naturally this results in incorrect month names and other inaccuracies.
- Leap month names are never filled in. This causes the managed code to assume that the leap month names are identical with the regular ones. Consequence is that 13th month in a leap year has no name and round-tripping of `DateTime`s fails.

The code around `ves_icall_System_Globalization_CalendarData_fill_calendar_data` will likely need to be refactored to fix the problem.

On September 1, 2019 this broke the CI on xunit tests that exercise the round-tripping of current date in the Hebrew calendar.

jaykrell:
Can you characterize what is wrong with the workaround? Is it ok just at the wrong layer?

filipnavara:
There should be a matrix of data. Hebrew calendar should have month names in English, Hebrew, and other languages.

I hard-coded English names for the stuff that was wrong and breaking the unit tests, but it is not entirely correct (names should be Hebrew for that particular case) and there is high likelyhood a similar problem exist with other data but was not discovered yet.

filipnavara:
Repro:
```
using System;

public static class Program {
    public static void Main() {
        var ci = new System.Globalization.CultureInfo("he-IL");
        ci.DateTimeFormat.Calendar = new System.Globalization.HebrewCalendar();
        Console.WriteLine(ci.DateTimeFormat.GetMonthName(13)); // should be non-empty
    }
}
```

jaykrell:
Is the non-leap year data correct? Duplicate it and add a month, and rename a month? Possibly by running the repro on desktop to get those exact names? Or maybe it is in the source data already that the non-leap derived data comes from.

jaykrell:
For the sake of the repro, also, something about plugging in the year 2019? And two of the months? And iterating languages, to compare with desktop? And maybe can manage to find equivalent C/Win32/go/Java, etc.

I also wonder if we can defer to another layer. Maybe try C runtime or Win32 first else fallback to limited (English) data.

What is really the matrix? How many languages are each calendar localized too? Any “tourist” or just the typical intersection of the calendar users and their languages?

filipnavara:
[...]

PR #16624, opened 2019-09-02
[corlib] Add hack for Hebrew calendar data

filipnavara:
Workaround for https://github.com/mono/mono/issues/16623. Unblocks the CI, should be back-ported to other branches.

/cc @akoeplinger @jaykrell

### DIFF
--- a/mcs/class/corlib/ReferenceSources/CalendarData.cs
+++ b/mcs/class/corlib/ReferenceSources/CalendarData.cs
@@ -43,7 +43,17 @@ namespace System.Globalization
 		static bool nativeGetCalendarData (CalendarData data, string localeName, int calendarId)
 		{
 			// TODO: Convert calendar-id to mono runtime calendar-id when it's used
-			return data.fill_calendar_data (localeName.ToLowerInvariant (), calendarId);
+			if (data.fill_calendar_data (localeName.ToLowerInvariant (), calendarId)) {
+				if ((CalendarId)calendarId == CalendarId.HEBREW) {
+					// HACK: Generate leap month names because the native code does not do it
+					// and tests would fail to round-trip dates on leap years.
+					// https://github.com/mono/mono/issues/16623
+					data.saMonthNames = new[] { "Nisan", "Iyar", "Sivan", "Tammuz", "Av", "Elul", "Tishrei", "Cheshvan", "Kislev", "Tevet", "Shevat", "Adar", "" };
+					data.saLeapYearMonthNames = new[] { "Nisan", "Iyar", "Sivan", "Tammuz", "Av", "Elul", "Tishrei", "Cheshvan", "Kislev", "Tevet", "Shevat", "Adar I", "Adar II" };
+				}
+				return true;
+			}
+			return false;
 		}
 
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
--- a/mcs/class/corlib/ReferenceSources/CalendarData.cs
+++ b/mcs/class/corlib/ReferenceSources/CalendarData.cs
@@ -34,6 +34,36 @@ namespace System.Globalization
 	[StructLayout (LayoutKind.Sequential)]
 	partial class CalendarData
 	{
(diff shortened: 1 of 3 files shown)


mcs/class/corlib/ReferenceSources/CalendarData.cs
```diff
@@ -43,7 +43,17 @@ public static int nativeGetTwoDigitYearMax (int calID)
...
 		{
 			// TODO: Convert calendar-id to mono runtime calendar-id when it's used
-			return data.fill_calendar_data (localeName.ToLowerInvariant (), calendarId);
+			if (data.fill_calendar_data (localeName.ToLowerInvariant (), calendarId)) {
+				if ((CalendarId)calendarId == CalendarId.HEBREW) {
+					// HACK: Generate leap month names because the native code does not do it
+					// and tests would fail to round-trip dates on leap years.
+					// https://github.com/mono/mono/issues/16623
+					data.saMonthNames = new[] { "Nisan", "Iyar", "Sivan", "Tammuz", "Av", "Elul", "Tishrei", "Cheshvan", "Kislev", "Tevet", "Shevat", "Adar", "" };
+					data.saLeapYearMonthNames = new[] { "Nisan", "Iyar", "Sivan", "Tammuz", "Av", "Elul", "Tishrei", "Cheshvan", "Kislev", "Tevet", "Shevat", "Adar I", "Adar II" };
```
filipnavara: The data may be in incorrect order (Biblical vs. Civil). I did some experiments but I have no idea how we end up in the month that it tries to locate. Also, NetFX seems to allow only combination of the Hebrew calendar with the `he-IL` locale so I should change the names to the Hebrew ones.

filipnavara: The order is indeed incorrect. We should be in  `Elul` now which is the 13th month on a leap year. Will fix it.

filipnavara: Fixed. Turned out that I had one more bug in there.

Apparently `MonthNames` always return 13 months and it's essential for the date/time parsing code. `LeapMonthNames` also has 13 months but the names are slightly different. The regular names are "Adar" and "Adar II", the leap names are "Adar I" and "Adar II".

filipnavara:
@monojenkins commit csproj

filipnavara:
@monojenkins build Linux AArch64 FullAOT+Interpreter+LLVM

mcs/class/corlib/ReferenceSources/CalendarData.cs
```diff
@@ -34,6 +34,36 @@ namespace System.Globalization
 	[StructLayout (LayoutKind.Sequential)]
 	partial class CalendarData
 	{
+		private static string[] HEBREW_MONTH_NAMES = new[] {
```
jaykrell: Maybe a comment as to roughly what these are? Perhaps the identifier suffice.
or a comment saying ideally it is in native, to avoid fix startup.

steveisok:

filipnavara:

mcs/class/corlib/ReferenceSources/CalendarData.cs
EgorBo: It's a hack anyway, we need to find out why our "ICU" analogue doesn't handle it and fix there (this PR just fixes CI)

filipnavara: It doesn't handle it because the `calendarId` is completely ignored, both in managed code and the native code. It was there as TODO for couple of years - `TODO: Convert calendar-id to mono runtime calendar-id when it's used`. Basically any combination of locale and non-default calendar for given locale is broken.

No one noticed it until now because Hebrew is the only supported calendar with leap months and it was the first leap year since the code was introduced. The produced `DateTime.ToString` results for the given combination were incorrect since the very beginning but they were round-trippable because the same wrong names were used for parsing.

directhex:

directhex:

steveisok:
@monojenkins backport 2019-02

PR #16628, opened 2019-09-02
Remove static from mingw tools.

jaykrell:
It doesn't work asis for me, or @MaximLipnin.


MaximLipnin:
@monojenkins build OS X WebAssembly

jaykrell:
I this is required by some configuration, but for me it only breaks things, and sound weird.
Static linking is to be minimize for performance.
But dynamic linking does breed dll hell, granted.

steveisok:
@monojenkins build failed

steveisok:
For OSX, applying this change was the only way I could move forward.

jaykrell:
Closing.

PR #16630, opened 2019-09-02
[netcore] Makefile clean up

EgorBo:
Remove ASP.NET sample (we currently have to download apsnet runtime everytime)
instead we can do:
```
dotnet new webapp
```
and do `make patch-app APP_PATH=/path/to/myasp` (works for any kind of apps/templates)


steveisok:
azp /run

PR #16449, opened 2019-08-24
[llvm] Implement more SIMD intrinsics on netcore.

vargaz:

PR #16631, opened 2019-09-02
[netcore] Implement Vector.get_Item intrinsic

filipnavara:


vargaz:
Conflicts with:
https://github.com/mono/mono/pull/16449

filipnavara:
Yeah, let's go with your version.

PR #16635, opened 2019-09-03
Brew zip stuff redundant with building source?

jaykrell:


jaykrell:
@grendello

grendello:
Mono requires `zlib` and the MinGW distribution that comes with `brew` does not come with it (unlike on Linux) and so we need to get it from somewhere - brew seemed to be the best place for it. Alternatively, Mono could be changed to build its own `zlib`, but would that really be better?

jaykrell:
Don't we already build our own?

grendello:
it wasn't the case when I implemented code to switch to mingw instead of using MXE - at least not when cross-building for Windows.

jaykrell:
Understood, but now it is. With the compressed/portable pdb support.

grendello:
If it works for mingw builds, then by all means the brew code is no longer needed.

jaykrell:
CI suggests yes? Or unclear, since autoconf could just be falling back to degraded functionality?
Oh, it is not clear.

We do unconditionally build and use zlib, in places, but it could be some code is disabled.
Needs more investigation I guess, or configure.ac/Makefile.am tweaks.

Where was it needed?

jaykrell:
Close for now as it is likely incomplete.
Configure.ac/ifded changes later maybe.

PR #16636, opened 2019-09-03
[wasm] Update emscripten to 1.38.43.

vargaz:


vargaz:
Causes assertions in run-sm-mini:
WASM-ERR: ﻿* Assertion at /Users/vargaz/git/wasm/mono/metadata/sgen-stw.c:69, condition `<disabled>' not met

vargaz:
@monojenkins build failed

lewing:
@kg because since part of the beta regressions are fixed by updating emscripten getting to the point where we can do that without other regressions is probably important

vargaz:
@monojenkins build failed

PR #16637, opened 2019-09-03
Set thread pool thread name just based on string pointer check.

jaykrell:
When the generation-based solution went in, there were not
yet constant thread names, so a pointer based approach could fail
due to reuse. We can do *slightly* simpler now.


marek-safar:
@vargaz could you please review that

jaykrell:

mono/metadata/object-internals.h
```diff
@@ -572,7 +572,7 @@ struct _MonoThreadInfo;
 
 typedef struct MonoThreadName {
 	char* volatile chars;      // null check outside of lock
-	gsize volatile generation; // read outside of lock
+	gsize volatile generation; // hypothetically read outside of lock; no longer actually used
```
filipnavara: Can we remove it?

jaykrell: Yes we could. It is a reasonable idea, but yes, removable.
It predates the ability to just do the pointer compare.

jaykrell: To be clear, w/o this PR, it is used. This PR replaces the use.

filipnavara: IMHO it should be removed if it's unused. We can always bring it back from history if necessary. Otherwise the change LGTM.

jaykrell: I'd be resolving merge conflicts in configure.ac every few days for months, had I made that change.

jaykrell: I’ll remove it today. Hopefully without having it sit months longer.

mono/metadata/threadpool.c
```diff
@@ -288,10 +288,19 @@ try_invoke_perform_wait_callback (MonoObject** exc, MonoError *error)
 	HANDLE_FUNCTION_RETURN_VAL (res);
 }
 
-static gsize
+static void
 set_thread_name (MonoInternalThread *thread)
 {
-	return mono_thread_set_name_constant_ignore_error (thread, "Thread Pool Worker", MonoSetThreadNameFlag_Reset);
+	// Any thread can set any other thread name at any time.
```
jaykrell: Given this duplicates knowledge from set_constant, merits a comment or a helper macro/function.

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:

jaykrell:
@monojenkins build monolite

mono/metadata/threadpool.c
jaykrell: I added a funny named flag to centralize the knowledge.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

filipnavara:
Azure Pipelines bot triggering is broken, it was already reported upstream by the dotnet people.

jaykrell:
@monojenkins build monolite

akoeplinger:
@jaykrell can you please rebase this PR to master?

lambdageek:

lambdageek:
@monojenkins build

mono/metadata/threads-types.h
```diff
@@ -342,12 +342,13 @@ typedef enum {
     MonoSetThreadNameFlag_Permanent = 0x0001,
     MonoSetThreadNameFlag_Reset     = 0x0002,
     MonoSetThreadNameFlag_Constant  = 0x0004,
+    MonoSetThreadNameFlag_RepeatedlyButOptimized = 0x0008,
```
jaykrell: I don't like this name.
Some might shorten it to "Fast" (the "optimized" part, omitting the "repeatedly" part).
I don't like that either, popular but vague.
I couldn't come up with anything else.
It could also be, instead of the behavior, the user -- "ThreadPoolWorker".
But I just pushed past the analysis/naming paralysis. :)

PR #16649, opened 2019-09-03
[netcore] Enable Vector<T>.get_Item () intrinsics for floats/doubles.

vargaz:


EgorBo:
Just for reference, accessing Vector's Item via const index: 
image
(boundexception is not possible here, but I guess LLVM will eliminate the one you emit anyway).

vargaz:
@monojenkins build failed

vargaz:
LLVM generates about the same code,  because it eliminates the switch we generate.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #16650, opened 2019-09-03
LLVM: Throw NRE if src or dst in Buffer.Memmove are nulls

EgorBo:
Fixes `System.Text.Tests.StringBuilderTests.Append_CharPointer_Null_ThrowsNullReferenceException` test for LLVM.
```csharp
var builder = new StringBuilder();
builder.Append(null, 2);
```
will end up in `Buffer.Memmove` and throw NRE in `*dst = xx` or `*src = xx` for regular mono and coreclr but not for `@llvm.memmove` intrinsic added [recently](https://github.com/mono/mono/pull/16610) (it crashes instead)


lewing:
@monojenkins build failed

PR #16652, opened 2019-09-03
Make case-insensitive class lookup work correctly with exported types

### LINKED ISSUE #15153: [netcore] Make System.Tests.ActivatorTests.TestingCreateInstanceObjectHandleFullSignature Pass

CoffeeFlux:
Fixes #15153 

This changes `mono_class_from_name_case_checked` to call into `mono_class_from_name_checked_aux`, which I made case-aware. The old code was broken in a couple different ways and almost never hit, as the only way to do so from C# was via reflection.

This is neither terribly pretty nor fast, but hopefully it sees little use outside of the test suite.

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -2946,6 +2946,20 @@ mono_image_add_to_name_cache (MonoImage *image, const char *nspace,
 
 typedef struct {
 	gconstpointer key;
+	GSList *values;
+} FindAllUserData;
+
+static void
+find_all_nocase (gpointer key, gpointer value, gpointer user_data)
+{
+	char *name = (char*)key;
+	FindAllUserData *data = (FindAllUserData*)user_data;
+	if (mono_utf8_strcasecmp (name, (char*)data->key) == 0)
+		data->values = g_slist_prepend (data->values, value);
+}
+
+typedef struct {
+	gconstpointer key;
 	gpointer value;
 } FindUserData;
 
@@ -3132,22 +3146,25 @@ mono_class_from_name_checked_aux (MonoImage *image, const char* name_space, cons
 		if (nspace_table)
 			token = GPOINTER_TO_UINT (g_hash_table_lookup (nspace_table, name));
 	} else {
-		FindUserData user_data;
-		user_data.key = name_space;
-		user_data.value = NULL;
+		FindAllUserData all_user_data = { name_space, NULL };
+		FindUserData user_data = { name, NULL };
+		GSList *values;
 
-		g_hash_table_foreach (image->name_cache, find_nocase, &user_data);
+		// We're forced to check all matching namespaces, not just the first one found,
+		// because our desired type could be in any of the ones that match case-insensitively.
+		g_hash_table_foreach (image->name_cache, find_all_nocase, &all_user_data);
 
-		if (user_data.value) {
-			nspace_table = (GHashTable*)user_data.value;
-			user_data.key = name;
-			user_data.value = NULL;
-
+		values = all_user_data.values;
+		while (values && !user_data.value) {
+			nspace_table = (GHashTable*)values->data;
 			g_hash_table_foreach (nspace_table, find_nocase, &user_data);
-
-			if (user_data.value)
-				token = GPOINTER_TO_UINT (user_data.value);
+			values = values->next;
 		}
+
+		g_slist_free (all_user_data.values);
+
+		if (user_data.value)
+			token = GPOINTER_TO_UINT (user_data.value);
 	}
 
 	mono_image_unlock (image);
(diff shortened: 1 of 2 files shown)


CoffeeFlux:
The issue I'm hitting with this PR is that when you use the cache, you have to check *all* matching namespace caches rather than just the first. I've pushed a commit to fix that, but it runs into an important issue: what order do you search the namespaces in? 

This is somewhat of a mess with all the case-insensitive lookup, and the CoreFX test suite recognizes it as such and only ensures we return a *valid* match, not some specific one as that's very heavily implementation dependent. This implementation, then, should pass all the tests.

However, it leads to use having different behavior from Core and in general somewhat strange selection behavior dependent on the hash function. For example, [this snippet](https://gist.github.com/CoffeeFlux/c8a4f584caf32163bfe0fb6578be512e) returns the following:
```
HelloWorld.MyNamespace1.MynAmespace3.gOo`1[System.Int32]
```

The alternative is to fall back to reiterating over the typedef and exported types tables, which is slower but will not change the order we iterate in. I've implemented that as well and pushed it to a branch [here](https://github.com/CoffeeFlux/mono/tree/case-insensitive-names-alt). When running the above sample, it returns the following:
```
HelloWorld.MyNamespace1.MynAmespace3.Goo`1[System.Int32]
```

I think with the inherent ambiguity of the case-insensitive lookup, the cache-based implementation is fine and significantly faster, but I've written this in an effort to document that decision should someone (including myself) run into compatibility-related issues down the line.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Remaining failure occurs when you build the entire project (make -j12 in the git root), but if you rebuild the runtime itself it goes away. It's probably invalidating some AOT'd image and the solution is to fix `get_class_from_name` to be case-insensitive as well, but I'll have to test.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Oh, forgot to link it back but I also pr'd corefx with a test addition that covers one of the scenarios fixed here: https://github.com/dotnet/corefx/pull/40788

kg:
@monojenkins build failed

CoffeeFlux:
Test failure might be relevant? I haven't the faintest idea why though, since the AOT compiler shouldn't be using case-insensitive lookup.

`Unable to compile method 'void MyType:Method ()' due to: 'Could not load file or assembly 'Dep, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.'.`

CoffeeFlux:
@monojenkins build failed

mono/metadata/class.c
```diff
@@ -3157,10 +3140,32 @@ mono_class_from_name_checked_aux (MonoImage *image, const char* name_space, cons
...
+			token = GPOINTER_TO_UINT (g_hash_table_lookup (nspace_table, name));
+	} else {
+		FindAllUserData all_user_data = { name_space, NULL };
+		FindUserData user_data = { name, NULL };
+		GSList *values;
 
-	if (nspace_table)
-		token = GPOINTER_TO_UINT (g_hash_table_lookup (nspace_table, name));
+		// We're forced to check all matching namespaces, not just the first one found,
+		// because our desired type could be in any of the ones that match case-insensitively.
```
lambdageek: Wow

PR #16655, opened 2019-09-04
[Wasm] Restore release-dynamic configuration

### LINKED ISSUE #16633: [Wasm] Restore release-dynamic configuration support
Since the update to the latest LLVM emscripten backend, the `release-dynamic` runtime configuration has been removed because of `-fPIC` issues.

Emscripten 1.38.43 is supposed to have fixes to address that, but an LLVM issue (https://github.com/emscripten-core/emscripten/issues/9317) is making compilation fail.

Another avenue is to avoid generating WASM object files, by using  `-s WASM_OBJECT_FILES=0` as [suggested by Zoltan here](https://gitter.im/mono/mono?at=5d389a4bd1cceb1a8db54627).

My working branch to try those avenues: https://github.com/jeromelaban/mono/tree/dev/jela/release-dynamic

jeromelaban:
Restores the release-dynamic configuration for mono-wasm, which was removed when changing the emscripten backend to LLVM.

Fixes #16633

### DIFF
--- a/sdks/builds/wasm.mk
+++ b/sdks/builds/wasm.mk
@@ -57,8 +57,8 @@ endif
 WASM_RUNTIME_AC_VARS= \
 	ac_cv_func_shm_open_working_with_mmap=no
 
-WASM_RUNTIME_BASE_CFLAGS=-fexceptions $(if $(RELEASE),-Os -g,-O0 -ggdb3 -fno-omit-frame-pointer)
-WASM_RUNTIME_BASE_CXXFLAGS=$(WASM_RUNTIME_BASE_CFLAGS) -s DISABLE_EXCEPTION_CATCHING=0
+WASM_RUNTIME_BASE_CFLAGS=-fexceptions $(if $(RELEASE),-Os -g,-O0 -ggdb3 -fno-omit-frame-pointer) -s WASM_OBJECT_FILES=0
+WASM_RUNTIME_BASE_CXXFLAGS=$(WASM_RUNTIME_BASE_CFLAGS) -s DISABLE_EXCEPTION_CATCHING=0 -s WASM_OBJECT_FILES=0
 
 WASM_RUNTIME_BASE_CONFIGURE_FLAGS = \
 	--disable-mcs-build \
--- a/sdks/builds/wasm.mk
+++ b/sdks/builds/wasm.mk
@@ -57,8 +57,8 @@ endif
 WASM_RUNTIME_AC_VARS= \
 	ac_cv_func_shm_open_working_with_mmap=no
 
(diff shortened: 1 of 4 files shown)


sdks/builds/wasm.mk
```diff
@@ -57,8 +57,8 @@ endif
 WASM_RUNTIME_AC_VARS= \
 	ac_cv_func_shm_open_working_with_mmap=no
 
-WASM_RUNTIME_BASE_CFLAGS=-fexceptions $(if $(RELEASE),-Os -g,-O0 -ggdb3 -fno-omit-frame-pointer)
-WASM_RUNTIME_BASE_CXXFLAGS=$(WASM_RUNTIME_BASE_CFLAGS) -s DISABLE_EXCEPTION_CATCHING=0
+WASM_RUNTIME_BASE_CFLAGS=-fexceptions $(if $(RELEASE),-Os -g,-O0 -ggdb3 -fno-omit-frame-pointer) -s WASM_OBJECT_FILES=0
+WASM_RUNTIME_BASE_CXXFLAGS=$(WASM_RUNTIME_BASE_CFLAGS) -s DISABLE_EXCEPTION_CATCHING=0 -s WASM_OBJECT_FILES=0
```
vargaz: This should be done by adding a new runtime build, similarly to the runtime-threads build, preferably optional, controlled by a make define.

sdks/wasm/Makefile
```diff
@@ -155,15 +150,12 @@ endef
 
 $(eval $(call InterpBuildTemplate,debug,wasm-runtime-release,$(EMCC_DEBUG_FLAGS)))
 $(eval $(call InterpBuildTemplate,release,wasm-runtime-release,$(EMCC_RELEASE_FLAGS)))
+$(eval $(call InterpBuildTemplate,release-dynamic,wasm-runtime-release,$(EMCC_RELEASE_DYNAMIC_FLAGS)))
 ifdef ENABLE_WASM_THREADS
```
vargaz: This should be controlled by an ifdef as well. We can define the ifdef in our CI in scripts/ci/run-jenkins.sh so the packages would contain the build.

sdks/builds/wasm.mk
jeromelaban: I added the configuration as an opt-in, but should it be an opt-out like the rest of the configuration ? I'm confused by the pthreads define name.

jeromelaban:

jeromelaban:
@monojenkins build mono-coreonly-CI

jeromelaban:
@monojenkins build failed

jeromelaban:
Looks like I do not have the permissions to rerun ADO builds with `/azp run`, where the three mono-coreonly builds fail.

jeromelaban:
Aside from the the two builds failing for unrelated reasons, the generated packages release-dynamic packages are running as they should.

PR #16663, opened 2019-09-04
[netcore] Fix the range checks for Vector.CopyTo().

vargaz:


filipnavara:
Verified locally that it fixes the `CopyTo` test failures.

mono/mini/simd-intrinsics-netcore.c
```diff
@@ -332,10 +332,14 @@ emit_sys_numerics_vector_t (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSig
...
 			break;
-		default:
+		case MONO_TYPE_I4:
+		case MONO_TYPE_U4:
 			opcode = OP_XEXTRACT_I32;
 			dreg = alloc_ireg (cfg);
 			break;
+		default:
+			// FIXME:
+			return NULL;
```
filipnavara: This still leaves out the exception checks generated above. I haven't checked if they get optimized out.

PR #16668, opened 2019-09-04
[llvm] Use branch weight metadata in some cases.

vargaz:


EgorBo:
@imhameed @vargaz  As far as I understand it applies some weight to the safepoint branches?
I thought we already use `llvm.expect` there which is converted into weights by `-lower-expect` pass.

vargaz:
Some parts of llvm, not sure which don't know about llvm.expect so the safe point call is not moved out from inside the loop when using the llvm jit, with this branch metadata, it is moved out.

EgorBo:
@vargaz um.. so `-lower-expect` (we probably need to make it the first pass) should just remove the `llvm.expect` completely and replace with !weight, example: https://godbolt.org/z/IS7Qqk

mono/mini/mini-llvm.c
```diff
@@ -6345,7 +6345,7 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 			args [1] = LLVMConstInt (LLVMInt1Type (), 1, FALSE);
 			cmp = LLVMBuildCall (ctx->builder, get_intrins (ctx, INTRINS_EXPECT_I1), args, 2, "");
 
-			LLVMBuildCondBr (builder, cmp, cont_bb, poll_bb);
+			mono_llvm_build_weighted_branch (builder, cmp, cont_bb, poll_bb, 64, 4);
```
lewurm: why 64 and 4?  is there a documentation for that somewhere?

vargaz: @imhameed ^^

marek-safar: I agree, make them constants at least

EgorBo: In my sample https://godbolt.org/z/IS7Qqk `-lower-expect` used `1` and `2000` weights for branches.

EgorBo: and I am actually not sure we use weights here in a correct order ("4" should be applied for the "we need to call state_pool" branch

vargaz: cmp is the result of expect, so its expected to be 1.

PR #16673, opened 2019-09-05
[cominterop] Default to SAFEARRAY(VARIANT) as CCW argument for object[].

nsivov:
Array argument marshalling logic depends on valid marshal spec, by default object[] arrays are mapped to SAFEARRAY(VARIANT) in ccw. This change implements such fallback; it was tested with _MethodInfo.Invoke() where method parameters are passed as a SAFEARRAY from ccw side, being object[] on receiving end.

### DIFF
--- a/mono/metadata/cominterop.c
+++ b/mono/metadata/cominterop.c
@@ -2021,6 +2021,7 @@ static MonoMarshalSpec*
 cominterop_get_ccw_default_mspec (const MonoType *param_type)
 {
 	MonoMarshalSpec *result = NULL;
+	MonoMarshalVariant elem_type = 0;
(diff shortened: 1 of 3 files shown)


marek-safar:
@vargaz @lateralusX please review

nsivov:
@lateralusX current behavior is to throw exception in emit_marshal_array_ilgen() for MARSHAL_ACTION_MANAGED_CONV_IN case on spec == NULL. So basically it does not work at all without attributes. With this change it will only handle szarray of objects, leaving the rest as it was before. I think regressions are unlikely, as it never worked before.

lateralusX:
@monojenkins rebuild failed.

PR #16674, opened 2019-09-05
[support] Remove some dead code

nsivov:
Both 'ptv' and 'tv' are unused since select() was replaced with poll() in 57b521e.


mono/metadata/threadpool-io.h
```diff
@@ -15,6 +15,10 @@ typedef struct _MonoIOSelectorJob MonoIOSelectorJob;
 
 TYPED_HANDLE_DECL (MonoIOSelectorJob);
 
+ICALL_EXPORT
+void
+ves_icall_System_IOSelector_Add (gpointer handle, MonoIOSelectorJobHandle job_handle, MonoError* error);
```
CoffeeFlux: Icalls should be defined in the icall-def files. This particular function appears to already be defined there - do the signatures not match? I don't see any error when building on OSX at least.

support/signal.c
```diff
@@ -481,13 +481,6 @@ wait_for_any (signal_info** signals, int count, int *currfd, struct pollfd* fd_s
...
 	// Once a second, stop to check if the VM is shutting down.
 	do {
-		struct timeval tv;
-		struct timeval *ptv = NULL;
-		if (timeout != -1) {
-			tv.tv_sec  = timeout / 1000;
-			tv.tv_usec = (timeout % 1000)*1000;
-			ptv = &tv;
-		}
 		r = poll (fd_structs, count, timeout);
```
jaykrell: Is `poll` expecting `ptv`?

jaykrell: Er, nevermind that, But the intent was to loop while `time remaining` and reduce it as it goes?

mono/metadata/threadpool-io.h
nsivov: They do match, but apparently threadpool-io.c does not include icall-decl.h header like, I'll fix it up. I get this warning with gcc 9.2.1, that's probably not what you're using on osx. To clean it up one step further, should I remove ves_icall_System_IOSelector_Remove() aswell, moving it to icall-decl.h?

support/signal.c
nsivov: With select() it was called with same timeout every time, same as it is now with poll(). So I'm not sure what you're suggesting.

mono/metadata/threadpool-io.h
jaykrell: No. remove is different, as I recall -- NOHANDLES vs. HANDLES.

CoffeeFlux: I think you can move the NOHANDLES ones as well? Not sure what the convention is though.
Looking further, the warning also appears to show up on netcore, but in that case I think the implementation needs to be ifdef'd out? Not sure what IOSelector is, but it looks like some mono-specific legacy thing.
```
threadpool-io.c:603:1: warning: no previous prototype for function 'ves_icall_System_IOSelector_Add' [-Wmissing-prototypes]
ves_icall_System_IOSelector_Add (gpointer handle, MonoIOSelectorJobHandle job_handle, MonoError* error)
```

nsivov: Just to make sure I understand, do you want me to move _Remove() to icall-decl.h? (out of threadpool-io.h is the point, I think)

nsivov:
Thanks for taking care of this @akoeplinger

PR #16677, opened 2019-09-05
[interp] Remove unused parameter to ves_pinvoke_method.

jaykrell:
It might actually cause stack pressure (but on its own, this PR does not save stack).


lewurm:
@monojenkins squash

PR #16681, opened 2019-09-05
[netcore] Bump CoreFX, core-setup dependencies

filipnavara:
Should fix some of the flaky tests by picking up dotnet/corefx@a192823.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Private.CoreFx.NETCoreApp" Version="5.0.0-alpha1.19409.2">
+    <Dependency Name="Microsoft.Private.CoreFx.NETCoreApp" Version="5.0.0-alpha1.19453.12">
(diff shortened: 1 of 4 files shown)


filipnavara:
Looks like couple of regressions popped up, most likely due to some packaging errors:
- `System.Json` and `System.Data.SqlClient` libraries are not found. The should come from the `Microsoft.Private.CoreFx.OOB` package but it looks broken.
- Tests added in https://github.com/dotnet/corefx/pull/40365 are not properly guarded for multiple platforms. It is not obvious whether that is intentional and whether they are supposed to be distributed.

filipnavara:
The failures to load assemblies are happening for tests run through `Microsoft.DotNet.RemoteExecutor`. Looks like something with runtimeconfig.json may be broken again.

filipnavara:
Apparently the downloaded `System.Data.SqlClient.Tests.deps.json` file has no reference to `System.Data.SqlClient` library. It may be loaded dynamically but since we don't use `CoreRun` the runtime doesn't have a way to locate these dependencies.

filipnavara:
Deps are sorted out (to be pushed), looking into the few remaining failures.

marek-safar:
Thank you!

PR #16682, opened 2019-09-05
[wasm] Add build.sh

MaximLipnin:
This build script contains a copy of build commands from existing wasm CI.

`cd sdks/wasm && ./build.sh`

### DIFF
--- a/sdks/wasm/.gitignore
+++ b/sdks/wasm/.gitignore
@@ -8,3 +8,4 @@ Mono.Cecil*
 emsdk_env.sh
 builds/threads-debug
 builds/threads-release
+.configured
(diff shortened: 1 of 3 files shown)


steveisok:
This is good since I always consult run-jenkins.sh for the up to date steps.  Are we ok with starting builds outside of netcore w/ build.sh?

kg:
This looks useful but does it really make sense to have a separate script that does Mostly what run-jenkins does? Can we just make the run-jenkins flow usable in such a way that this script can invoke it with the right arguments instead?

steveisok:
I prefer run-jenkins to be isolated.  I think this helps streamline the local sdk build, so I'm ok with it.

PR #16685, opened 2019-09-05
RFC: [draft][wasm] improve error message when unable to read pdbs

ghuntley:
Howdy folks,

Opening this early for feedback. I think this error message could be improved to add some helpful suggestions such:

- "Ensure DebugType is set to portable"
- "Ensure DebugSymbols is set to true"
- "Ensure DEBUG is defined in DefineConstants"

Sure folks like us know what this message means because we are digging in the source-code but this error message is starting to be seen by end-user folks who use the Uno Platform w/wasm.

https://github.com/unoplatform/uno/pull/1489/files

Opening up early, so we can discuss and then if 👍 please provide suggestions as to language and I'll rebase my PR.

### DIFF
--- a/sdks/wasm/Mono.WebAssembly.DebuggerProxy/MonoProxy.cs
+++ b/sdks/wasm/Mono.WebAssembly.DebuggerProxy/MonoProxy.cs
@@ -759,6 +759,9 @@ namespace WsProxy {
 				}
 			} catch (Exception e) {
 				var o = JObject.FromObject (new {
+					// discuss: "Ensure DebugType is set to portable"
+					// discuss: "Ensure DebugSymbols is set to true"
+					// discuss: "Ensure DEBUG is defined in DefineConstants"
 					scriptSource = $"// Unable to read document ({e.Message})\n" +
 								$"Local path: {src_file?.SourceUri}\n" +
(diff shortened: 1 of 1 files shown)


lewing:
I'm absolutely in favor of surfacing better errors here.  If you are looking for feedback on the specific error strings my only suggestion would be to make them more directive toward solving the problem.

ghuntley:
Should the error string contain a error code (if so, how does that work) and direction to an article on docs.microsoft.com vs encoding direction in the string itself?

kg:
Historically BCL exceptions do not contain error codes or URLs. I'd be fine with embedding a URL but we'd need to be sure that the URL won't break and I don't know if there's any sort of BCL policy about this.

Since we control this code I don't see anything wrong with merging it containing a URL as long as we can make sure the URL won't change.

There are a few historical examples in msbuild I can recall that provide advice on how to fix the problem, though the advice is often wrong. So based on that (and just generally my opinion + larry's) I think it makes sense to try and provide basic guidance. Maybe don't be specific like 'set DebugType to portable' since the correct choices there may change and invalidate the exception, but be more general like 'Check your settings to ensure your build is configured for debugging'

akoeplinger:
Looks like there's been no activity on this for a few months and there are no open questions at the moment, closing this for now :)

Issue #16678, opened 2019-09-05
Buffer.BlockCopy is slow


EgorBo:
see https://github.com/mono/mono/blob/master/netcore/System.Private.CoreLib/src/System/Buffer.cs#L19-L56
this method does 4 icalls (`IsPrimitiveTypeArray`, `IsPrimitiveTypeArray`, `ByteLength`, `ByteLength`) before calling Memmove (which is fast itself). At least we can merge these 4 calls.

Also, note https://github.com/mono/mono/blob/master/netcore/System.Private.CoreLib/src/System/Buffer.cs#L58-L65

Related benchmarks:
![image](https://user-images.githubusercontent.com/523221/64333063-77bbda80-cfde-11e9-8226-1ac4884a3a11.png)

marek-safar:
@vargaz @BrzVlad is this something you would like to look at?

EgorBo:
@marek-safar I think I can fix it

PR #16692, opened 2019-09-05
[netcore] Optimize Buffer.BlockCopy and Buffer.IsPrimitiveArray

EgorBo:
Move them to icalls.
Fixes https://github.com/mono/mono/issues/16678

### DIFF
--- a/mono/metadata/icall-def-netcore.h
+++ b/mono/metadata/icall-def-netcore.h
@@ -57,7 +57,9 @@ HANDLES(ARRAY_12, "SetValue",         ves_icall_System_Array_SetValue, void, 3, 
 HANDLES(ARRAY_13, "SetValueImpl",  ves_icall_System_Array_SetValueImpl, void, 3, (MonoArray, MonoObject, guint32))
 HANDLES(ARRAY_14, "SetValueRelaxedImpl",  ves_icall_System_Array_SetValueRelaxedImpl, void, 3, (MonoArray, MonoObject, guint32))
 
-ICALL_TYPE(BUFFER, "System.Buffer", BUFFER_1)
+ICALL_TYPE(BUFFER, "System.Buffer", BUFFER_0)
+HANDLES(BUFFER_0, "BlockCopy", ves_icall_System_Buffer_BlockCopy, void, 5, (MonoArray, int, MonoArray, int, int))
+HANDLES(BUFFER_0a, "IsPrimitiveTypeArray", ves_icall_System_Buffer_IsPrimitiveTypeArray, MonoBoolean, 1, (MonoArray))
 HANDLES(BUFFER_1, "_ByteLength", ves_icall_System_Buffer_ByteLengthInternal, gint32, 1, (MonoArray))
 NOHANDLES(ICALL(BUFFER_2, "__Memmove", ves_icall_System_Runtime_RuntimeImports_Memmove))
 
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -7157,6 +7157,52 @@ mono_array_get_byte_length (MonoArrayHandle array)
 	}
 }
 
+void
+ves_icall_System_Buffer_BlockCopy (MonoArrayHandle src, int src_offset, MonoArrayHandle dst, int dst_offset, int count, MonoError *error)
+{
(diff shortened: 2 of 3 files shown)


jaykrell:
Suggest take advantage of the new pinning-ness of coop handles?
Stlll take handles, but at the start of the function, assign them to raw pointers and usually use those?
i.e. I don't think you need the safepoint stuff any longer.

mono/metadata/icall-def-netcore.h
```diff
@@ -57,7 +57,9 @@ HANDLES(ARRAY_12, "SetValue",         ves_icall_System_Array_SetValue, void, 3,
 HANDLES(ARRAY_13, "SetValueImpl",  ves_icall_System_Array_SetValueImpl, void, 3, (MonoArray, MonoObject, guint32))
 HANDLES(ARRAY_14, "SetValueRelaxedImpl",  ves_icall_System_Array_SetValueRelaxedImpl, void, 3, (MonoArray, MonoObject, guint32))
 
-ICALL_TYPE(BUFFER, "System.Buffer", BUFFER_1)
+ICALL_TYPE(BUFFER, "System.Buffer", BUFFER_0)
+HANDLES(BUFFER_0, "BlockCopy", ves_icall_System_Buffer_BlockCopy, void, 5, (MonoArray, int, MonoArray, int, int))
```
Therzok: Why only on .netcore? refsources version is identical.

mono/metadata/icall-def-netcore.h
```diff
@@ -57,7 +57,9 @@ HANDLES(ARRAY_12, "SetValue",         ves_icall_System_Array_SetValue, void, 3,
...
+HANDLES(BUFFER_0a, "IsPrimitiveTypeArray", ves_icall_System_Buffer_IsPrimitiveTypeArray, MonoBoolean, 1, (MonoArray))
```
Therzok: BUFFER_0 not 0a?

mono/metadata/icall.c
```diff
@@ -7157,6 +7157,52 @@ mono_array_get_byte_length (MonoArrayHandle array)
...
+
+	MONO_EXIT_NO_SAFEPOINTS;
+
+	return;
+}
+
+MonoBoolean
+ves_icall_System_Buffer_IsPrimitiveTypeArray (MonoArrayHandle array, MonoError* error)
+{
+	MonoClass * const klass = m_class_get_element_class (MONO_HANDLE_GETVAL (array, obj.vtable)->klass);
```
vargaz: This should be a JIT intrinsics, will look into it.

mono/metadata/icall-def-netcore.h
EgorBo: "Legacy" Mono's BlockCopy is already optimized, in this version is based on it.

mono/metadata/icall.c
EgorBo: Ok, I'll remove it once your PR appear

filipnavara:
Here is a result of benchmark run (without noise filtered out): https://gist.github.com/filipnavara/b0f6d433b7f0e67f452c4f115ad4c5f1

The performance for `BlockCopy` has indeed improved significantly. However, on closer inspection `Span.Sorting.QuickSortArray`/`Span.Sorting.QuickSortSpan` looks like a regression. I have no idea how could it be caused by this PR because there's no obvious use of either `BlockCopy` or `IsPrimitiveArray` but I could not attribute it to a BDN in-process mode quirk based on closer inspection of the numbers.

EgorBo:
@jaykrell could you please explain how it works, don't I have to "pin" the managed arrays to memmove them?

jaykrell:
You used to have but I don't think so any more, recent change.

jaykrell:
Just to clarify what I know and am not sure of:

Having a handle to an "object" (MonoObject, etc.) now pins it.
There is no need to use HANDLE_GETVAL etc. You cay say:

```
FooObject fooh = xxx;
Foo* foo = MONO_HANDLE_RAW(fooh);
foo->bar; // previously not
```

What I am not 100% of is how this applies to the elements of an array.
I believe you can now access them more directly.

kg:
@monojenkins build failed

filipnavara:
Can we get this in? I'd be happy to look into `IsPrimitiveTypeArray` improvements later but it is already a big improvement as is.

PR #16695, opened 2019-09-05
All existing safepoints appear to be mooted by pinned coop handles.

jaykrell:
Perhaps they are never needed then.

### DIFF
--- a/mono/utils/mono-threads-api.h
+++ b/mono/utils/mono-threads-api.h
@@ -144,6 +144,7 @@ mono_threads_enter_no_safepoints_region (const char *func);
 void
 mono_threads_exit_no_safepoints_region (const char *func);
 
+#if 0
 #define MONO_ENTER_NO_SAFEPOINTS						\
 	do {										\
 		do {									\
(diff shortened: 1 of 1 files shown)


lambdageek:
Are there any other reasons, besides managed memory safety, that we may want a runtime check that ensures a given region of code doesn't run anything that may safepoint?

If not, then this seems fine.

jaykrell:
I am not sure. This is kinda an RFC not just a PR.
I think someone with more expertise (you) could/should invent more fine grained "safepoint" macros, if even called that, that may or may not do anything.

```
SAFEPOINT_BEGIN (array access with pinning handle)
SAFEPOINT_END()
```

or such.

also, relevant question:

A MonoArray conceptually or factually is:

```
struct Array
{
  MonoObject base;
 int lenth;
  T data[]; // array
};
```

and NOT

```
struct Array // NOT this
{
  MonoObject base;
  int length;
  T  * data; // pointer, not this 
};
```

right? That is, pinning any array (creating a coop handle, rather) pins its elements, right?

lambdageek:
Pinning an array will pin its contents, yes. So if it's an array full of valuetypes, those values are fixed in memory. However if those contents are object references, the objects aren't pinned.

PR #16705, opened 2019-09-06
[interp] Outline source to stfld_vt.

jaykrell:
Eventually this could maybe contribute to stack savings.
For now it is just splitting up the interpreter slightly, arguably for the better, arguably for the worse.


BrzVlad:
I think I mentioned in some other PR. I'd rather not outline potentially hot opcodes without a clear purpose / perf check. (even if this might not matter too much since it's doing another call)

PR #16706, opened 2019-09-06
[interp] Reduce register pressure in endfinally.

jaykrell:
Eventually this could contribute to stack savings. On its own, nothing.
The compiler apparently likes to create temporaries, like ip and ip - 1.
Which then have to survive functions calls, on the stack, if nonvolatile registers are exhausted.
This combats that and in a larger PR helped.


lewing:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16707, opened 2019-09-06
[intep] Outline source to stobj_vt.

jaykrell:
In time this might be leveraged to conserve stack,
due to i.e. c and size traversing function calls, but for now,
without an always_inline marker, it should only be cosmetic.

Or hey, when not optimizing, you can set a breakpoint on the function
which could be nice.


BrzVlad:
I think I mentioned in some other PR. I'd rather not outline potentially hot opcodes without a clear purpose / perf check. (even if this might not matter too much since it's doing another call)

jaykrell:
To be clear: it isn’t marked outline, is likely to be inlined, I can check. Though the point is, if enough other changes taken, to then mark outline in a smaller PR.

PR #16708, opened 2019-09-06
[interp] Embrace that ip is const guint16* and remove casts to that type.

jaykrell:
or to gint16*. Or is there a reason for this style?

Maybe could change type but still want to cast for the deref?


lewurm:
@monojenkins build failed

lewurm:
@monojenkins squash

lewurm:

jaykrell:
I have a feeling the opcodes used to be single byte.
Those days are gone forever?

PR #16717, opened 2019-09-07
[draft] [draft] [draft] [interp] Save 48 bytes of stack usually.

jaykrell:
and relieve pressure on most of the code. This is the recursive/non-recursive split.


jaykrell:
I am confused about something & valgrind highlights it well.

jaykrell:
I was missing the alloca in localalloc. This now works with csc so stands a chance.
Needs more work though.

jaykrell:
Closing for now, try again later.
Savings likely 32 bytes, and then can respend
stack for most opcodes.

PR #16718, opened 2019-09-07
[wasm] Initial netcore support.

vargaz:

### DIFF
--- a/sdks/wasm/netcore-corefx.diff
+++ b/sdks/wasm/netcore-corefx.diff
@@ -1,0 +1,55 @@
+diff --git a/src/Native/Unix/CMakeLists.txt b/src/Native/Unix/CMakeLists.txt
+index f1d6a9ac59..4a8c99e140 100644
+--- a/src/Native/Unix/CMakeLists.txt
++++ b/src/Native/Unix/CMakeLists.txt
+@@ -18,7 +18,7 @@ add_compile_options(-Wno-empty-translation-unit)
+ add_compile_options(-Wno-cast-align)
+ add_compile_options(-Wno-typedef-redefinition)
+ add_compile_options(-Wno-c11-extensions)
+-add_compile_options(-fPIC)
++#add_compile_options(-fPIC)
+ add_compile_options(-I${CMAKE_CURRENT_SOURCE_DIR}/Common)
+ add_compile_options(-I${CMAKE_CURRENT_BINARY_DIR}/Common)
+ add_compile_options(-g)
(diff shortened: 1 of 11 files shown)


vargaz:
This is initial support for wasm+netcore, controlled by the ENABLE_WASM_NETCORE make flag in Make.config. This is an optional build configuration which can work alongside a normal build.

vargaz:
See the run-hello-netcore target for usage.

vargaz:
Note that this cannot be enabled on CI, since it depends on an external corefx repo to compile a wasm version of System.Native plus some System.Native patches.

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

sdks/wasm/netcore-corefx.diff
```diff
@@ -0,0 +1,55 @@
+diff --git a/src/Native/Unix/CMakeLists.txt b/src/Native/Unix/CMakeLists.txt
```
marek-safar: Why is this needed? We should not be building this under Mono

vargaz: There are no wasm binaries for System.Native.

filipnavara: We should bring them back and publish through NuGet. That is no trivial effort though. At some point the WASM targets existed in CoreFX but they were removed and never made it to Azure Pipelines. I've been told that there should be a machine pool with emscripten and other relevant tools installed but I know no details about that.

marek-safar: I agree we need to start the effort on corefx side as well (I'll look into that)

filipnavara: For reference, the initial WebAssembly support was added with https://github.com/dotnet/corefx/pull/28936. Various parts were removed by https://github.com/dotnet/corefx/pull/36739.

filipnavara: Here's the definition for the Azure Pipelines pool:
```
        pool:
            name:  NetCorePublic-Pool
            queue: buildpool.windows.10.wasm.open
```

jkotas: > Various parts were removed by dotnet/corefx#36739

The initial WebAssembly support was on top of ProjectN. The removed parts were the ProjectN specific plumbing.

vargaz:
System.Native needs to be built with the same emscripten version we are using, the same flags, etc. We might need more than one version depending on flags.

PR #16719, opened 2019-09-07
[sdks] Change the DISABLE_ conditionals to ENABLE_ to avoid building extra stuff.

vargaz:


jaykrell:
Thank you. Negative settings beget double negatives, which become confusing.

vargaz:
@monojenkins build failed

PR #16721, opened 2019-09-07
[interp] Improve debuggability with enum.

jaykrell:


mono/mini/interp/interp.c
```diff
@@ -211,17 +211,17 @@ int mono_interp_traceopt = 0;
 #define USE_COMPUTED_GOTO 1
 #endif
 #if USE_COMPUTED_GOTO
-#define MINT_IN_SWITCH(op) COUNT_OP(op); goto *in_labels[op];
+#define MINT_IN_SWITCH(op) COUNT_OP(op); goto *in_labels[opcode = (op)];
 #define MINT_IN_CASE(x) LAB_ ## x:
 #define MINT_IN_DISPATCH(op) goto *in_labels[op];
 #if DEBUG_INTERP
 #define MINT_IN_BREAK if (tracing > 1) { MINT_IN_DISPATCH(*ip); } else { COUNT_OP(*ip); goto *in_labels[*ip]; }
```
lewurm: should do it here as well

lewurm:
did you verify if it doesn't add an additional load to each dispatch? especially important for the computed goto case

jaykrell:
1. It should be optimized away, because it is never read, but ok I will check.
2. At least the typedef? `gdb> p (MintOpcode)*ip`

jaykrell:
Comparing just gcc/Linux/amd64, -O2, the code (`objdump -d`) appears to be the same before/after.
(I added a blank line to normalize line numbers).

jaykrell:
Actually the last cleanup is there, removal of define defualt, but I can still break it up if desired.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16725, opened 2019-09-07
Fix Xaml reader to parse escaped values correctly

DavidKarlas:
This makes it possible to parse escaped values https://docs.microsoft.com/en-us/dotnet/framework/xaml-services/escape-sequence-markup-extension


marek-safar:
@DavidKarlas could you please add a test for the change?

DavidKarlas:
Added unit test

PR #16726, opened 2019-09-08
[interp] Fix buffer overreads and overwrites found by valgrind.

jaykrell:


mono/mini/interp/transform.c
```diff
@@ -3260,7 +3260,7 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
 			}
 			break;
 		case CEE_LDC_I4_1:
-			if (!td->is_bb_start[td->ip + 1 - td->il_code] && 
+			if (td->ip - td->il_code + 1 < td->code_size && !td->is_bb_start [td->ip + 1 - td->il_code] &&
 				(td->ip [1] == CEE_ADD || td->ip [1] == CEE_SUB) && td->sp [-1].type == STACK_TYPE_I4) {
```
jaykrell: `sp [-1] ` looks suspicious too.

jaykrell:
`valgrind /s/mono2/mono/mini/mono-sgen --interp /i/2/lib/mono/4.5/csc.exe /s/1.cs`

and probably overkill:

```
$ cat /s/1.cs
using System;

public struct big
{
 public int a,b,c;
};

class A
{
[ThreadStatic] static big a,b;

     static void Main (string[]args)
{
 int i = 0;
 a.a = ++i;
 a.b = ++i;
 a.c = ++i;
 b.a = ++i;
 b.b = ++i;
 b.c = ++i;
 Console.WriteLine("{0} {1} {2} {3} {4} {5}", a.a, a.b, a.c, b.a, b.b, b.c);
}

}
```

jaykrell:
@monojenkins build failed

mono/mini/interp/transform.c
```diff
@@ -5359,8 +5359,9 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
 			td->sp = td->stack;
 			SIMPLE_OP (td, MINT_ENDFINALLY);
 			td->last_ins->data [0] = td->clause_indexes [in_offset];
-			// next instructions are always part of new bb
-			td->is_bb_start [td->ip - header->code] = 1;
+			// next instructions, if they exist, are always part of new bb
+			if (td->ip - td->il_code < td->code_size)
+				td->is_bb_start [td->ip - header->code] = 1;
```
jaykrell: Could also grow the allocation of `is_bb_start` by one.

mono/mini/interp/transform.c
BrzVlad: This just accesses the next instruction following a LDC. I don't think is possible for such an instruction to be the end of the code stream. Where is this exactly triggered ? Aside from that, the access for is_bb_start was and is still wrong. We should never access it if we are inlining (it is only declared for the original method IL code).

Also, your change is wrong because apparently td->code_size is only declared for the orginal method, while td->ip and td->il_code are for the current method to be compiled. Hope we are not making this mistake elsewhere.

mono/mini/interp/transform.c
BrzVlad: How does this even happen ? Can a method end with endfinally ? :confused:

If not, I would rather throw invalid IL.

jaykrell: I don't know. Let me try to see.

debug_interp also has similar bugs, that make more sense, i.e. during inlining.
```
#0  0x000000000035b720 in generate_code (td=0x1ffeff6ab0, method=0x9c2bb70, header=0xb603530, generic_context=0x9c2bbb0, error=0x1ffeff6d90)
    at interp/transform.c:3102
#1  0x00000000003562f7 in interp_inline_method (td=0x1ffeff6ab0, target_method=0x9c2bb70, header=0xb603530, error=0x1ffeff6d90) at interp/transform.c:1790
#2  0x00000000003578d3 in interp_transform_call (td=0x1ffeff6ab0, method=0xade97f0, target_method=0x9c2bb70, domain=0x6075fc0, generic_context=0xade9830,
    is_bb_start=0xb600970 "\001", constrained_class=0x0, readonly=0, error=0x1ffeff6d90, check_visibility=1, save_last_error=0) at interp/transform.c:2086
#3  0x000000000035dcb6 in generate_code (td=0x1ffeff6ab0, method=0xade97f0, header=0xb601370, generic_context=0xade9830, error=0x1ffeff6d90)
    at interp/transform.c:3371
#4  0x0000000000370db6 in generate (method=0xade97f0, header=0xb601370, rtm=0x1ffeff6c90, generic_context=0xade9830, error=0x1ffeff6d90)
    at interp/transform.c:6376
#

		if (td->verbose_level > 1) {
			g_print ("IL_%04lx %s %-10s, sp %ld, %s %-12s vt_sp %u (max %u)\n",
				td->ip - td->il_code,
				td->is_bb_start [td->ip - td->il_code] == 3 ? "<>" :
```

mono/mini/interp/transform.c
jaykrell: Aha, you are right. It is inlining. Probably for all of them.

```
Thread 1 received signal SIGTRAP, Trace/breakpoint trap.
0x0000000000354f1f in generate_code (td=0x1ffefdf5f0, method=0xa5e91c8, header=0xa2809f0, generic_context=0x0, error=0x1ffefdf8d0) at interp/transform.c:3253
3253                            if (!td->is_bb_start[td->ip + 1 - td->il_code] && td->ip [1] == 0xfe && td->ip [2] == CEE_CEQ &&
(gdb) bt 9
#0  0x0000000000354f1f in generate_code (td=0x1ffefdf5f0, method=0xa5e91c8, header=0xa2809f0, generic_context=0x0, error=0x1ffefdf8d0) at interp/transform.c:3253
#1  0x000000000034eb53 in interp_inline_method (td=0x1ffefdf5f0, target_method=0xa5e91c8, header=0xa2809f0, error=0x1ffefdf8d0) at interp/transform.c:1790
#2  0x000000000035012f in interp_transform_call (td=0x1ffefdf5f0, method=0xa5e91a0, target_method=0xa5e91c8, domain=0x6075fc0, generic_context=0x0,
    is_bb_start=0xa628760 "\001", constrained_class=0x0, readonly=0, error=0x1ffefdf8d0, check_visibility=1, save_last_error=0) at interp/transform.c:2086
#3  0x0000000000356512 in generate_code (td=0x1ffefdf5f0, method=0xa5e91a0, header=0xb7728c0, generic_context=0x0, error=0x1ffefdf8d0) at interp/transform.c:3371
#4  0x000000000034eb53 in interp_inline_method (td=0x1ffefdf5f0, target_method=0xa5e91a0, header=0xb7728c0, error=0x1ffefdf8d0) at interp/transform.c:1790
#5  0x000000000035012f in interp_transform_call (td=0x1ffefdf5f0, method=0xa28ac40, target_method=0xa5e91a0, domain=0x6075fc0, generic_context=0xa28ac80,
```

mono/mini/interp/transform.c
jaykrell: See `TryDequeueSlow`.

Here is a reduced form:

```
class A
{
void a(int b)
{
  while (true)
    lock (this) { }
}
}

csc 3.cs /t:library /optimize

.method private hidebysig instance void  a(int32 b) cil managed
{
  // Code size       24 (0x18)
  .maxstack  2
  .locals init (class A V_0,
           bool V_1)
  IL_0000:  ldarg.0
  IL_0001:  stloc.0
  IL_0002:  ldc.i4.0
  IL_0003:  stloc.1
  .try
  {
    IL_0004:  ldloc.0
    IL_0005:  ldloca.s   V_1
    IL_0007:  call       void [mscorlib]System.Threading.Monitor::Enter(object,
                                                                        bool&)
    IL_000c:  leave.s    IL_0000
  }  // end .try
  finally
  {
    IL_000e:  ldloc.1
    IL_000f:  brfalse.s  IL_0017
    IL_0011:  ldloc.0
    IL_0012:  call       void [mscorlib]System.Threading.Monitor::Exit(object)
    IL_0017:  endfinally
  }  // end handler
} // end of method A::a
```

```

File: mono/mini/interp/transform.c (hunk 1 above)
Comment (BrzVlad) [reply mono/mini/interp/transform.c]: You should make sure that it works for inlining and that code_size is properly used (either kill it from TransformData and use header->code_size, or just keep it updated for inlined method to avoid further problems. Can't say I care too much whether it reports invalid IL or not.

File: mono/mini/interp/transform.c (hunk 2 above)
Comment (BrzVlad) [reply mono/mini/interp/transform.c]: I thought ret is always required for method end, I guess not. Make sure the code_size thing. Note that we never inline methods that have clauses.

File: mono/mini/interp/transform.c (hunk 1 above)
Comment (jaykrell) [reply mono/mini/interp/transform.c]: `is_bb_start` needs computing for inlining, right?
Only at most once per method? i.e. not for each inline attempt, or inline and regular?
Is there provision for that?

How about..I know this is lame..well, for inlining, for this specific code, we should just skip the optimization. But we should do better.

Ok, reading more code, it looks like we don't inline anything with a branch, to greatly simplify?

But I don't see what `is_bb_start`, um, it looks like it is allocated and sized once per call tree. And that IL offsets fro multiple functions are indexed in to the same array.
That seems very wrong, but I could also be reading it wrong.

I have a feeling that not inlining anything with a branch provides an easy fix though.
i.e. there isn't a need to have multiple is_bb_starts and look up from the inlined code. Just that the inlined code is considered all to occur at the point of the call to it.

So maybe just or in `inlining` before any `is_bb_start` check, or make a function out of that.

**Comment by jaykrell** [comment]
Updated, still needs testing.

File: mono/mini/interp/transform.c [hunk 3]
```diff
@@ -6226,14 +6241,16 @@ interp_cprop (TransformData *td)
...
 
 	for (ins = td->first_ins; ins != NULL; ins = ins->next) {
 		int pop, push;
 		int il_offset = ins->il_offset;
 		// Optimizations take place only inside a single basic block
 		// If two instructions have the same il_offset, then the second one
 		// cannot be part the start of a basic block.
-		gboolean is_bb_start = il_offset != -1 && td->is_bb_start [il_offset] && il_offset != last_il_offset;
+		// Only single basic blocks are inlined.
+		gboolean is_bb_start = il_offset != -1 && (inlining || td->is_bb_start [il_offset]) && il_offset != last_il_offset;
```
BrzVlad: We don't need this. Optimizations happen once all code has been generated. The concept of inlining no longer exists. Also il_offsets are only for the original method (if we inline a call, all instructions generated from the inlined method will have the same il_offset as the call instruction from the parent)

mono/mini/interp/transform.c
BrzVlad: we don't need to compute is_bb_start for inlined methods, because nobody is bb start there. Maybe this logic should be extracted in a separate method at some point.

mono/mini/interp/transform.c
```diff
@@ -984,6 +984,9 @@ emit_store_value_as_local (TransformData *td, MonoType *src)
 static gboolean
 interp_is_bb_start (TransformData *td, InterpInst *start, InterpInst *end)
 {
+	// Only single basic blocks are inlined.
+	if (td->inlined_method)
+		return FALSE;
```
BrzVlad: I don't think this is needed, il_offsets should be correct. I might have to update this method at some point since its looks fishy

jaykrell: Ah, this is also after IL transformed to IR, and inlining finished?
I missed the distinction in the code where it moves from IL to IR.

jaykrell: I undid it.

jaykrell:
@monojenkins build failed

mono/mini/interp/transform.c
BrzVlad: We could do inlining but the assigned il_offset to an instruction is always from the parent method, so it's fine to access is_bb_start with those offsets (it might be redundant but it is safe)

jaykrell:
Possibly relevant failure.

jaykrell:

jaykrell:
@monojenkins build failed

Issue #16701, opened 2019-09-06
C# REPL Unusable on Mono 6 / Ubuntu 19


nkosi23:
## Steps to Reproduce

1. Launch the REPL with the "csharp" command on Ubuntu 19 / Mono 6.0.0.327
2. Type any C# statement in the REPL

### Current Behavior

The cursor isn't are right place, it is always at the top of screen therefore the text and the output of statements always overlap with the previously written text, making the shell unusable in practice.

![2019-09-06_10-34-46](https://user-images.githubusercontent.com/6731436/64413647-2a994080-d092-11e9-8a24-713ee607a29f.png)
![2019-09-06_10-35-21](https://user-images.githubusercontent.com/6731436/64413657-2ec55e00-d092-11e9-95be-9ed731facd3b.png)

![2019-09-06_10-26-03](https://user-images.githubusercontent.com/6731436/64413396-b9598d80-d091-11e9-9d40-1d865a0847ba.png)
![2019-09-06_10-30-04](https://user-images.githubusercontent.com/6731436/64413407-bd85ab00-d091-11e9-89b1-fa758d678652.png)
![2019-09-06_10-31-27](https://user-images.githubusercontent.com/6731436/64413419-c1193200-d091-11e9-87db-00e201358023.png)
![2019-09-06_10-32-22](https://user-images.githubusercontent.com/6731436/64413434-c4142280-d091-11e9-90a1-332b3cdade8c.png)

### Expected Behavior

The output and cursor should be on the last line of the screen

## On which platforms did you notice this

[ ] macOS
[ X] Linux
[ ] Windows

**Version Used**:

Mono 6.0.0.327. Reproduced on both the xcfe terminal and the xterm terminal (only two terminals I have installed), so the problem comes from the C# REPL and not from the terminal.

lewurm:
does `$TERM` match with the terminal you are using?

Therzok:
@lwurm i've seen this on macos too

nkosi23:
@lewurm What do you mean? Typing $TERM in the REPL throws an error

>Error CS1056: Unexpected Character '$'

By removing the '$' I get an error CS0103 "The name TERM does not exist in the current context.

lewurm:
huch... I couldn't reproduce it at first:
* running it from a `screen(1)` session + iTerm2 + `TERM=rxvt` works _fine_.
* running from a plain iTerm2 session + `TERM=rxvt` behaves like reported.

[...]

PR #16745, opened 2019-09-09
[TermInfoDriver] update top/left value on Console.{CursorTop,CursorLeft} access

lewurm:
It can be outdated due to writes to `stdout`.

Fixes https://github.com/mono/mono/issues/16701

### DIFF
--- a/mcs/class/corlib/System/TermInfoDriver.cs
+++ b/mcs/class/corlib/System/TermInfoDriver.cs
@@ -554,6 +554,8 @@ namespace System {
 			get {
 				if (!inited) {
 					Init ();
+				} else {
+					GetCursorPosition ();
 				}
 
 				return cursorLeft;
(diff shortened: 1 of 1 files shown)


lewurm:
Not sure about this one, see also #16701. Was it always broken like this or did something else change?

From debugging it looks like `TermInfoDriver` doesn't track the cursor position properly which is kinda hard, because it gets completely by-passed when writes to `stdout` happen. So I wonder how that could ever have worked. I didn't find anything obvious with `git blame`/`git log`.

migueldeicaza:
I do not think this is a good idea, because it will produce a very heavy sequence every time it is queried.

If you write behind the back of the console, the results are not guaranteed.

lewurm:
Coincidentally I experienced the same issue with `sdb` yesterday. It turns it uses the same copy of `getline.cs` as `csharp` does, so maybe that's where the bug is really buried?

At least in terms of "UX performance" this change isn't noticeable (contrary to when the query is done after each keypress, then the auto completion gets all messed up while continuing typing).

@migueldeicaza If you have an idea how to fix that in a better way, feel free to revert this change.

PR #16746, opened 2019-09-09
Clean up map.c /map.h

steffen-kiess:
These commits remove all manual changes from map.c / map.h (which should be autogenerated by create-native-map).


marek-safar:
approve

marek-safar:
@lambdageek please review

steffen-kiess:
The windows build was broken because I didn't remove `Mono_Posix_FromMremapFlags` and `Mono_Posix_ToMremapFlags` from `msvc/monoposixhelper.def`, should be fixed now.

support/stdio.c
```diff
@@ -202,7 +202,7 @@ Mono_Posix_Stdlib_fgetc (void* stream)
 	return fgetc (stream);
 }
 
-char*
+void*
```
lambdageek: Why `void*`?  `fgets` returns a `char*`

steffen-kiess: The return type has to match the prototype in `map.h`, and the prototype in `map.h` is autogenerated from `Stdlib.sys_fgets` which returns `IntPtr` (which is mapped to `void*`): https://github.com/mono/mono/blob/ad2132e611a7e8b7c7dcaed1c3ecc23d40dc17de/mcs/class/Mono.Posix/Mono.Unix.Native/Stdlib.cs#L888

akoeplinger: We only check the pointer against null in Stdlib.cs so the type isn't really relevant :)

akoeplinger:
Thanks for the contribution!

PR #16748, opened 2019-09-10
[netcore] Disable more features.

vargaz:


configure.ac
```diff
@@ -1361,6 +1361,8 @@ elif test x$with_runtime_preset = xnetcore; then
    mono_feature_disable_com='yes'
    mono_feature_disable_gac='yes'
    mono_feature_disable_perfcounters='yes'
+   mono_feature_disable_attach='yes'
+   mono_feature_disable_gac='yes'
```
marek-safar: mono_feature_disable_gac='yes' is already in the list

PR #16753, opened 2019-09-10
Avoid creating un-representable DateTime for transition point

MaximLipnin:
Fixes https://github.com/mono/mono/issues/16742

### DIFF
--- a/mcs/class/corlib/System/TimeZoneInfo.cs
+++ b/mcs/class/corlib/System/TimeZoneInfo.cs
@@ -1251,8 +1251,11 @@ namespace System
 
 		private static DateTime TransitionPoint (TransitionTime transition, int year)
 		{
-			if (transition.IsFixedDateRule)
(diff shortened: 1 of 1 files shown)


MaximLipnin:
@monojenkins build failed

marek-safar:
Please add a test case for it

MaximLipnin:
I'm not sure that I have a good test case which would work w/o additional set up because I had to change my system time in order to get repro.

steveisok:
You can add a test case with something like:

```
var parsedTime = DateTime.Parse("1948-02-19T23:00:00Z", CultureInfo.InvariantCulture);
var newTime = TimeZoneInfo.ConvertTime(parsedTime, TimeZoneInfo.FindSystemTimeZoneById("Europe/Rome"));
```

MaximLipnin:
@monojenkins build failed

PR #16760, opened 2019-09-10
[wasm] Fix several typos

MaximLipnin:

### DIFF
--- a/sdks/wasm/xunit-exclusions.rsp
+++ b/sdks/wasm/xunit-exclusions.rsp
@@ -9,7 +9,7 @@
 -noclass System.Tests.ProgressTests
 -nonamespace System.Threading.Tests
 -nonamespace System.Threading.Tasks
--nonamespace Sytem.Threading.Tasks
+-nonamespace System.Threading.Tasks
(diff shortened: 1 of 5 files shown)


sdks/wasm/xunit-exclusions.rsp
```diff
@@ -9,7 +9,7 @@
 -noclass System.Tests.ProgressTests
 -nonamespace System.Threading.Tests
 -nonamespace System.Threading.Tasks
--nonamespace Sytem.Threading.Tasks
+-nonamespace System.Threading.Tasks
```
vargaz: Thats a typo in the test class name, not here.

MaximLipnin: Added a comment

MaximLipnin: Removed the comment because its can be passed on the xunit command line. The original issue with name can be fixed in related corefx sources.

MaximLipnin: Filed in https://github.com/dotnet/corefx/issues/40973

PR #16771, opened 2019-09-11
[wasm] update run-jenkins.sh

MaximLipnin:
- replaced set of `run-*-mini` targets with more common `run-all-mini` one;
- removed redundant `run-v8-corlib` (there is `run-all-corlib`)


MaximLipnin:
@monojenkins build failed

PR #16782, opened 2019-09-11
[mini] Fix couple of R4 constant folding checks

filipnavara:


mono/mini/cfold.c
```diff
@@ -97,13 +97,13 @@ mono_constant_fold_ins (MonoCompile *cfg, MonoInst *ins, MonoInst *arg1, MonoIns
 	switch (ins->opcode) {
 	case OP_FADD:
 	case OP_FMUL:
-		if (arg2->opcode == OP_R8CONST && arg2->opcode == OP_R8CONST){
+		if ((arg1->opcode == OP_R8CONST || arg1->opcode == OP_R4CONST) && arg2->opcode == arg1->opcode){
 			ALLOC_DEST (cfg, dest, ins);
```
vargaz: Needs an extra space at the end.

vargaz: This won't work because OP_R4CONST has its inst_p0 field pointing to a float, not a double, so FOLD_BINOP will not work.

mono/mini/cfold.c
```diff
@@ -258,6 +258,14 @@ mono_constant_fold_ins (MonoCompile *cfg, MonoInst *ins, MonoInst *arg1, MonoIns
 			dest->inst_c0 = arg1->inst_c0;
 		}
 		break;
+	case OP_FMOVE:
+		if (arg1->opcode == OP_R8CONST || arg1->opcode == OP_R4CONST) {
+			ALLOC_DEST (cfg, dest, ins);
+			dest->opcode = arg1->opcode;
+			dest->sreg1 = -1;
+			dest->inst_p0 = arg1->inst_p0;
+		}
```
vargaz: This might lead to worse code because it needs to load the same r4/r8 immediate many times.

vargaz: Nm, it was already there.

mono/mini/cfold.c
filipnavara: Thanks for noticing it, I fixed it. I don't think I have actually ever hit this case since Roslyn already does this optimization (for non-inlined code). The main issue I saw was the cut & paste error in the condition.

mono/mini/cfold.c
filipnavara: I just moved it next to the other OP_xMOVE cases and added same treatment as for OP_R8CONST.

filipnavara:

PR #16784, opened 2019-09-11
[interp] Cleanup dispatch macros.

jaykrell:
- Factor out reptition into already existing macros.
 - Remove the default for computed goto which is dead.
   Computed goto has no notion of default. It is just
   an unchecked array access.

Extracted from https://github.com/mono/mono/pull/16721.


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lewurm:
@monojenkins squash

PR #16786, opened 2019-09-12
[interp] recurse/norecurse split -- save 32 bytes and relieve pressure in non-recursive cases.

jaykrell:
This saves 32 bytes in the recursive cases *and* relieves pressure of stack use in the non-recursive cases.
For example the imethod local could come back, which would be a register on arm64.

The recursive cases become a little trickier, but the non-recursive cases can be easier and perhaps more optimized.

There was a 16 byte regression, which this fixes, which is the point -- we can stop worrying about it in most of the code, and then use even more.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -290,7 +290,7 @@ mono_interp_error_cleanup (MonoError* error)
 	error_init_reuse (error); // one instruction, so this function is good inline candidate
 }
 
-static MONO_NEVER_INLINE void
+static void
 ves_real_abort (int line, MonoMethod *mh,
 		const unsigned short *ip, stackval *stack, stackval *sp)
 {
@@ -455,7 +455,7 @@ interp_pop_lmf (MonoLMFExt *ext)
 	mono_pop_lmf (&ext->lmf);
 }
 
-static MONO_NEVER_INLINE InterpMethod*
+static InterpMethod*
 get_virtual_method (InterpMethod *imethod, MonoVTable *vtable)
 {
 	MonoMethod *m = imethod->method;
@@ -579,7 +579,7 @@ alloc_method_table (MonoVTable *vtable, int offset)
 	return table;
 }
 
-static MONO_NEVER_INLINE InterpMethod* // Inlining causes additional stack use in caller.
+static InterpMethod*
 get_virtual_method_fast (InterpMethod *imethod, MonoVTable *vtable, int offset)
 {
 	gpointer *table;
@@ -1003,7 +1003,7 @@ ves_array_calculate_index (MonoArray *ao, stackval *sp, gboolean safe)
 	return pos;
 }
 
-static MONO_NEVER_INLINE MonoException*
+static MonoException*
 ves_array_set (InterpFrame *frame, stackval *sp, MonoMethodSignature *sig)
 {
 	MonoObject *o = sp->data.o;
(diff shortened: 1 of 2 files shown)


mono/mini/interp/interp.c
```diff
@@ -3262,21 +3244,219 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
+			vtalloc,
+#endif
+			error);
+
+		finally_ips = child_frame.finally_ips;
+		ip = child_frame.ip;
+		sp = child_frame.sp;
+		vt_sp = child_frame.vt_sp;
+		child_frame.parent = frame;
+
```
jaykrell: Might want null these fields in child_frame here. I had that in some version.

BrzVlad:
I am not really fond of this change since it complicates the interpreter loop.

Having said that, one of the biggest sources of slowness in the interpreter is the call overhead, since a lot of work needs to be done when calling this huge method with a lot of prolog. This PR looks like it doubles the call overhead since for each call we additionally do a transition between `interp_exec_method_full` and `interp_exec_method_inner` which doesn't feel like a good idea. Benchmarking this just to be sure, it shows a 20% performance regression. This split idea looks like a no go.

We should instead try to get rid of the call overhead. One way I did it was through inlining. The second might be to make the interpreter non recursive.

I'm closing this since I don't think any performance improvement can come through this idea. Feel free to reopen if new developments appear.

jaykrell:
Darn. Is this a "real world mix" or "call intensive"?
Can you share it?
There might be a way to reduce the overhead but probably not eliminate it.

For example, we could un-union ip/vt_sp/sp, and then use the address-taken fields, moving the overhead from copy back and forth to indirection, which might be less. We'd still have the extra call/ret/prologue/epilogue though.

I wouldn't think real world code would be very impacted, but an artificial benchmark I guess.
And real world code could benefit from optimizing the other opcodes while still using less stack as they recurse (locality), but granted that is mostly already done.

jaykrell:
This structure really eases maintenance of all the non-recursive cases.
It is unfortunate that call perf is so sensitive that it can't tolerate the extra round of return/call/epilogue/prologue. I not sure I can do the split and optimize that away. :(

BrzVlad:
https://gist.github.com/BrzVlad/e1a3fe3370477e6c8d041d20245819d4

The fact that a few benchmarks get considerable perf improvement is an encouraging sign for experimenting with a non-recursive interpreter; which should have even better stack usage than this PR, while having less call overhead.

jaykrell:
This is with the InterpFrame child_frame allocated off of the machine frame, Zoltan's draft?
Or something else?
Yes I guess so, either way -- faster and less stack than this PR, should be possible.
I have another idea that I don't have working, involving alloca, and possibly alloca of negative numbers, or just alloca and tracking a high water mark. Probably the same perf as moving off the machine stack.
The 0x78 in my PR is still high seeming -- the compiler wants to create like 0x58 for almost no code.
Can you share how to run the benchmarks?

BrzVlad:
You can see here the location and command line for every benchmark https://github.com/BrzVlad/benchmarker/tree/master/benchmarks

Yes, it would be based on that draft. The advantage of that is for each InterpFrame we have just the InterpFrame allocated in the mempool and then we don't care at all about the stack usage of interp_exec_method_full, because we will have a single C frame, with the entire stack always in the fastest cache.

jaykrell:
Might still want to try this, or https://github.com/mono/mono/pull/16880, on wasm/ios13.

jaykrell:
@kjpou1 @vargaz

jaykrell:
Reopening this at least temporarily. WebAssembly might still benefit significantly, like over 100 bytes saved.

jaykrell:
Same thing against newer master has been seen to fix the plague of WebAssembly problems.

PR #16787, opened 2019-09-12
[interp] In jmp, only alloca the additional space needed, not all of the space.

jaykrell:
Repeated alloca are adjacent, and this is the last alloca.
Unless there was a localalloc.
This is a little confusing, because you can jmp through multiple functions, with a mix of localloc. However only the most recent one should matter, so this seems correct.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3441,7 +3441,7 @@ main_loop:
 				MONO_PROFILER_RAISE (method_tail_call, (frame->imethod->method, new_method->method));
 
 			if (!new_method->transformed) {
-				MONO_API_ERROR_INIT (error);
+				error_init_reuse (error);
 
 				frame->ip = ip;
 				mono_interp_transform_method (new_method, context, error);
@@ -3449,20 +3449,22 @@ main_loop:
 				if (ex)
 					THROW_EX (ex, ip);
 			}
-			ip += 2;
 			const gboolean realloc_frame = new_method->alloca_size > frame->imethod->alloca_size;
-			frame->imethod = new_method;
 			/*
 			 * We allocate the stack frame from scratch and store the arguments in the
 			 * locals again since it's possible for the caller stack frame to be smaller
 			 * than the callee stack frame (at the interp level)
 			 */
 			if (realloc_frame) {
-				frame->stack = (stackval*)g_alloca (frame->imethod->alloca_size);
-				memset (frame->stack, 0, frame->imethod->alloca_size);
+				if (frame->imethod->has_localloc)
+					frame->stack = (stackval*)g_alloca (new_method->alloca_size);
+				else
+					frame->stack = (stackval*)g_alloca (new_method->alloca_size - frame->imethod->alloca_size);
+				memset (frame->stack, 0, new_method->alloca_size);
 				sp = frame->stack;
 			}
-			vt_sp = (unsigned char *) sp + frame->imethod->stack_size;
+			frame->imethod = new_method;
+			vt_sp = (guchar*)sp + new_method->stack_size;
 #if DEBUG_INTERP
 			vtalloc = vt_sp;
 #endif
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

mono/mini/interp/interp.c
```diff
@@ -3441,28 +3441,30 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
...
 			 * locals again since it's possible for the caller stack frame to be smaller
 			 * than the callee stack frame (at the interp level)
 			 */
 			if (realloc_frame) {
-				frame->stack = (stackval*)g_alloca (frame->imethod->alloca_size);
-				memset (frame->stack, 0, frame->imethod->alloca_size);
+				if (frame->imethod->has_localloc)
+					frame->stack = (stackval*)g_alloca (new_method->alloca_size);
+				else
+					frame->stack = (stackval*)g_alloca (new_method->alloca_size - frame->imethod->alloca_size);
```
BrzVlad: I can't say I understand much from this PR. Here we set frame->stack with allocated memory on the stack that is not enough (shouldn't frame->stack always point to at least new_method->alloca_size amount of memory). What if new_method->alloca_size < frame->imethod->alloca_size. This PR cares about LOCALLOC but not about other sources of alloca ?

jaykrell: I think there are just two sources: `localloc` and `alloca_size`.

The point is that `alloca` is adjacent/additive.
You can reuse the `alloca` of the conceptual caller towards production of the alloca for the callee.
If the caller has 10 bytes and the callee needs 30, you can get by with just 20 more.

Unless there is another allocation in between the localloc -- which might have other problems, but ignored here.

I think those are the only two allocas. If I messed some, then that is a mistake.

lewurm: > Unless there is another allocation in between the localloc -- which might have other problems, but ignored here.

That sounds very easy to break for little benefit.

jaykrell: Jmp isn’t supposed to keep growing the stack.
To really implement this correctly,
 need to return to caller and then do the full alloca
there and then call back to here.
Kinda like the “split” but only for jmp.

BrzVlad: You haven't answered my first 2 questions

jaykrell:
I think I need to try something a bit better anyway. The cost can be likely absorbed into the branch around do_transform, i.e. no cost for normal functions.

> What if new_method->alloca_size < frame->imethod->alloca_size. 

It allocates the extra needed.

> This PR cares about LOCALLOC but not about other sources of alloca ?

If there are others then this wrong. I thought this was the only one.

Ok to close though until/unless I test & fix the larger problem here. It still I think uses unbounded stack, just less.

BrzVlad:
What do you mean it allocates the extra needed ? You would setup the frame->stack of the new frame with the result of alloca of a negative number. So all the locals and the execution stack of the method we are jumping to resides in that space. I don't know what that means.

jaykrell:
Maybe I wrote it backwards?? Let me check..

jaykrell:
```
const gboolean realloc_frame = new_method->alloca_size > frame->imethod->alloca_size;

if (realloc_frame) {
  ...
  frame->stack = (stackval*)g_alloca (new_method->alloca_size - frame->imethod->alloca_size);
```

jaykrell:
But yeah my answer was wrong.
If it is less, it doesn't do any alloca, not "it allocates the extra".

jaykrell:
I confused. Closed because there is a lingering already existing problem? Or closed due to some misunderstanding?

If the newly required space is more than it already has, it allocates it.
If the newly required space is less than it already has, it reuses what it has.
In neither case does it need to do the complete allocation, unless there is an intervening localloc.

jaykrell:
i.e. this should improve things some, but likely more is needed, granted.

BrzVlad:
I don't think this change is worth it. It is hackish to make internal assumptions about the result of alloca, that it is referencing more free memory on the stack than what we ask for (even if it might be always true, it is just confusing). We also have other sources of alloca, or might have in the future. All of this to marginally optimize an opcode that is not even generated by the c# compiler.

PR #16788, opened 2019-09-12
[draft] [interp] Add assert in jmp that sp == stack at start.

jaykrell:
It makes it true sometimes, only if not calling alloca.
Throwing invalid IL if false might be preferred here (or maybe it is ruled out by earlier analysis).


jaykrell:
@monojenkins build failed

PR #16789, opened 2019-09-12
[interp] Cleanup comments about o/frame_objref/GC safety.

jaykrell:
Not every o is related to this and the comment isn't always above.


jaykrell:
Closing in deference to https://github.com/mono/mono/pull/16790.

PR #16798, opened 2019-09-12
Explicit disable hybrid suspend when cooperative suspend is enabled.

lateralusX:
hybrid suspend won't get disabled when using --enable-cooperative-suspend triggering configure error:

Hybrid suspend and Cooperative suspend cannot be both enabled.

Fix explicit set hybrid suspend default behavior to no if cooperative suspend has been configured.


lateralusX:
@monojenkins rebuild failed

PR #16801, opened 2019-09-12
Fix warnings in eglib gspawn.c on platforms lacking fork/execv.

lateralusX:


lateralusX:
@monojenkins rebuild failed

PR #16802, opened 2019-09-12
Fix build warnings in libmonoruntime.

lateralusX:


lateralusX:
@monojenkins rebuild failed

mono/metadata/w32process-unix.c
```diff
@@ -1185,6 +1174,20 @@ mono_w32process_signal_finished (void)
 	mono_coop_mutex_unlock (&processes_mutex);
 }
 
+#if defined (HAVE_FORK) && defined (HAVE_EXECVE)
+static void
+switch_dir_separators (char *path)
+{
+	size_t i, pathLength = strlen (path);
+
+	/* Turn all the slashes round the right way, except for \' */
```
jaykrell: I know you only moved it, but the comment is not clear.

mono/metadata/w32process-unix.c
```diff
@@ -1185,6 +1174,20 @@ mono_w32process_signal_finished (void)
...
+	/* There are probably other characters that need to be excluded as well. */
+	for (i = 0; i < pathLength; i++) {
+		if (path[i] == '\\' && i < pathLength - 1 && path[i + 1] != '\'')
```
jaykrell: The code and the comment do not agree.
Specifically, the code is also not converting the last character in the string.

lateralusX: Code was just moved as part of this PR, guess we should fix this in separate PR.

jaykrell: Understood and I meant to acknowledge that, sorry.

jaykrell: Quite possibly the function can be removed.
It depends.
A lot of code, e.g. CreateProcess, CreateFile, does not care.
But the strings will maybe make it to argv[] in child and maybe it is picky.
But you could also run such programs w/o mono, with forward slashes.
Unclear whose job it it is to convert (nobody?), and who can/should be sensitive (nobody?).

Issue #16805, opened 2019-09-12
Casting a constant to IntPtr generates redundant movs


EgorBo:
```csharp
static IntPtr Test1(int x) => (IntPtr)42;
static IntPtr Test2(int x) => (IntPtr)x;
```

CoreCLR generates:
```asm
; Test1(Int32)
mov eax, 0x2a
ret

; Test2(Int32)
movsxd rax, ecx
ret
```

Mono-LLVM JIT generates:
```asm
; Test1(Int32)
mov    QWORD PTR [rsp-0x8],0x0
mov    QWORD PTR [rsp-0x8],0x4
mov    rax,QWORD PTR [rsp-0x8]
ret
 
; Test2(Int32)
movsxd rax,edi
ret
```

Something wrong with the case where I use a constant

EgorBo:
Fixed by https://github.com/mono/mono/pull/16812

PR #16807, opened 2019-09-12
Cast constants to (U)IntPtr in a single mov

EgorBo:
Fixes: https://github.com/mono/mono/issues/16805

Test case:
```csharp
static unsafe IntPtr FromConst() => (IntPtr)42;
```
CoreCLR:
```asm
mov eax, 0x2a
ret
```

Mono-LLVM (Before):
```asm
mov    QWORD PTR [rsp-0x8],0x0
mov    QWORD PTR [rsp-0x8],0x2a
mov    rax,QWORD PTR [rsp-0x8]
ret 
```

Mono-LLVM (After):
```asm
mov eax, 0x2a
ret
```

It's mostly a hack (workaround) but according to Zoltan, a proper fix is quite complicated:

> the problem is that IntPtr is a valuetype.
> so even through the call is inlined, we still have to take the address of a local etc.
> which leads to worse code.
> since we don't have a pass to get rid of those things, or the pass is not good enough.
> i.e. here is how the code looks:
> IL_0003: ldloca.s V_0
> IL_0005: ldc.i4.4
> IL_0006: call native int [System.Runtime]System.IntPtr::op_Explicit(int32)
> since IntPtr is a valuetype, the 'this' argument is passed by ref.
> so the easy solution is to add an intrinsics.
> the more complicated one is to fix the 'alias-analyis' pass in the jit to get rid of the ldloca.

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -1705,6 +1705,25 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 				return ins;
 			}
 		}
+	} else if ((cmethod->klass == mono_defaults.int_class || cmethod->klass == mono_defaults.uint_class) && 
+		!strcmp (cmethod->name, "op_Explicit") && fsig->param_count == 1) {
+
+		int from_type = fsig->params [0]->type;
+		if (from_type == MONO_TYPE_I4 || from_type == MONO_TYPE_U4) {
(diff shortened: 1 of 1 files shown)


filipnavara:
@marek-safar I believe the fix is simple enough that it's not worth writing an issue for that. The general case is already handled in the alias analysis and I submitted #16812 to fix this specific issue.

marek-safar:
@filipnavara are you saying this fix is no longer needed after your fix?

filipnavara:
@marek-safar Yes, assuming my fix is correct.

EgorBo:
Closing in favor of https://github.com/mono/mono/pull/16812

PR #16813, opened 2019-09-13
Fix warnings in libmono.

lateralusX:


mono/mini/mini-native-types.c
```diff
@@ -95,6 +95,22 @@ typedef struct {
 
 #endif
 
+gsize
+mini_magic_type_size (MonoCompile *cfg, MonoType *type)
```
jaykrell: I realize you are just moving code, but a comment near the word "magic" sure might be nice.
What is "magic"?

lateralusX:
@monojenkins rebuild failed

PR #16817, opened 2019-09-13
[draft] Prune atomic icalls.

jaykrell:
I wonder if reflection moots all this?
i.e. get an address of the things, not just inline intrinsics?

arm64 and amd64 can just about everything
x86 can do everything 32bit.
x86 could do 64bit load/store as floating point
and 64bit otherwise with lock cmpxch8b on current processors
but that isn't done here

float==int32 and double==int64, except maybe alignment, so unchanged

s390x can add and exchange, int and intptr.
PowerPC can add and cas, int and intptr.

arm32 requires armv7 to do much, so unchanged

And really, if we could convincingly efficiently do the lingering
coop work here, that maybe preferred.

Here I mainly go after the icall table, and less so the implementation,
but could/should do both.

Granted, #if considered bad.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,7 +62,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This line is parsed by tools besides autoconf, such as msvc/mono.winconfig.targets.
 # It should remain in the format they expect.
 #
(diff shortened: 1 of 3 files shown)


jaykrell:
@monojenkins build failed

jaykrell:

jaykrell:
Maybe these all have to be present.

jaykrell:

jaykrell:
I guess these have to stay. Closing.

PR #16823, opened 2019-09-13
[llvm] Convert byref args to i64 stack type

### LINKED ISSUE #16808: LLVM failed for String.GetHashCode
`String.GetHashCode` calls `Marvin.ComputeHash32` which can not be converted to LLVM, a simple repro:

```csharp
using System;

unsafe class Program
{
    {
        byte[] a = new byte[100];
        ComputeHash32(ref a[0]);
    }
    
    public static unsafe void ComputeHash32(ref byte data)
    {
        for (int i=0; i<1; i++)
            data = ref Unsafe.Add(ref data, 1);
            // or:
            // data = ref Unsafe.AddByteOffset(ref data, (IntPtr)1);
    }
}
```
`--llvm` output contains:

```
LLVM failed for 'Program.ComputeHash32': incoming phi arg type mismatch
```
It explains why `string.GetHashCode` is still slow even if LLVM mode (JIT) is enabled

filipnavara:
Fixes #16808

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -3578,9 +3578,12 @@ emit_entry_bb (EmitContext *ctx, LLVMBuilderRef builder)
 			/* The IR treats these as variables with addresses */
 			ctx->addresses [reg] = LLVMGetParam (ctx->lmethod, pindex);
 			break;
-		default:
-			ctx->values [reg] = convert_full (ctx, ctx->values [reg], llvm_type_to_stack_type (cfg, type_to_llvm_type (ctx, ainfo->type)), type_is_unsigned (ctx, ainfo->type));
+		default: {
+			/* Use the type of the assigned class to match the types used by OP_PHI and mono_compile_create_var */
+			MonoType *var_type = m_class_get_byval_arg (cfg->args [i + sig->hasthis]->klass);
+			ctx->values [reg] = convert_full (ctx, ctx->values [reg], llvm_type_to_stack_type (cfg, type_to_llvm_type (ctx, var_type)), type_is_unsigned (ctx, var_type));
 			break;
+		}
 		}
 	}
 	g_free (names);


mono/mini/mini-llvm.c
```diff
@@ -3579,7 +3579,10 @@ emit_entry_bb (EmitContext *ctx, LLVMBuilderRef builder)
 			ctx->addresses [reg] = LLVMGetParam (ctx->lmethod, pindex);
 			break;
 		default:
-			ctx->values [reg] = convert_full (ctx, ctx->values [reg], llvm_type_to_stack_type (cfg, type_to_llvm_type (ctx, ainfo->type)), type_is_unsigned (ctx, ainfo->type));
+			if (ainfo->type->byref)
+				ctx->values [reg] = convert_full (ctx, ctx->values [reg], IntPtrType(), FALSE);
+			else
+				ctx->values [reg] = convert_full (ctx, ctx->values [reg], llvm_type_to_stack_type (cfg, type_to_llvm_type (ctx, ainfo->type)), type_is_unsigned (ctx, ainfo->type));
```
filipnavara: Match the logic here:

https://github.com/mono/mono/blob/9608debd7bdf94fe27258d7365d7c04d1c648d90/mono/mini/ssa.c#L440-L443

filipnavara:
@monojenkins build OS X i386

filipnavara:

filipnavara:
@vargaz @imhameed Can you review please? I rewrote the original hack into something that I believe to be correct.

vargaz:
Would prefer:
https://github.com/mono/mono/pull/16874

kg:
I'm not super familiar with our LLVM infrastructure but do we pipe our wasm through llvm? If so, the correct stack type would be pointer not i64, since in wasm pointers are 32-bit

filipnavara:
> Would prefer: #16874

That is exactly what I originally used. Only downside is that I was not sure if it properly covers all the cases...

filipnavara:
>  If so, the correct stack type would be pointer not i64, since in wasm pointers are 32-bit

@kg it would actually use whatever the native pointer type is. I should have been more clear about that. The problem was that it used pointer type (i64*) where pointer-sized integer type was expected.

PR #16825, opened 2019-09-13
[debugger] New way to filter exceptions to support VSWin features

thaystg:
Creating a new version of MOD_KIND_EXCEPTION_ONLY to support VsWin features, discussed with Joaquin Jares.

Because we don't have this functionality of get every other exception that is not specified, VsWin was getting every exception and filtering later. But this causes a slowness during the debugger because we stop all thread on each exception that we get, even if they are caught and we don't want to stop.

This fix a bug of slowness on Android Debugger.


marek-safar:
@vargaz please review

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ExceptionEventRequest.cs
```diff
@@ -3,10 +3,10 @@
 
 namespace Mono.Debugger.Soft
 {
-	public sealed class ExceptionEventRequest : EventRequest {
+	public class ExceptionEventRequest : EventRequest {
```
vargaz: Wouldn't be easier to add the new property to the existing class ? Here and everywhere else ? And just pass it when the debuggee has a newer protocol version ?

thaystg: The thing is, VSMac will continue using the older exception filter, because VsMac features are supported with that, but VSWin needs another implementation, because of that I decided to keep both. But I could keep both and receive a flag to decide which one should I use... I don't know if this would be easier to understand.

vargaz: The flag would have the default value which is the old behavior and the ide could set it to a different value to get the new behavior.

thaystg: Done

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ExceptionEventRequest.cs
```diff
@@ -6,7 +6,7 @@ namespace Mono.Debugger.Soft
 	public sealed class ExceptionEventRequest : EventRequest {
 
 		TypeMirror exc_type;
-		bool caught, uncaught, subclasses;
+	    bool caught, uncaught, subclasses, win_features, everything_else;
```
vargaz: Indentation change.

vargaz: This should have some descriptive name instead of win_features.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ExceptionEventRequest.cs
```diff
@@ -19,6 +19,23 @@ public sealed class ExceptionEventRequest : EventRequest {
 			this.caught = caught;
 			this.uncaught = uncaught;
 			this.subclasses = true;
+			this.win_features = false;
+			this.everything_else = false;
+		}
+
+		internal ExceptionEventRequest (VirtualMachine vm, TypeMirror exc_type, bool caught, bool uncaught, bool win_features, bool everything_else) : base (vm, EventType.Exception) {
+			if (exc_type != null) {
```
vargaz: These ctors are internal, they can be merged.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
```diff
@@ -277,6 +277,13 @@ public void SetSocketTimeouts (int send_timeout, int receive_timeout, int keepal
 			return new ExceptionEventRequest (this, exc_type, caught, uncaught);
 		}
 
+		public ExceptionEventRequest CreateExceptionRequest2 (TypeMirror exc_type, bool caught, bool uncaught, bool everything_else) {
+			if (Version.AtLeast (2, 54))
```
vargaz: This could have the name same since it has a different signature.

mono/mini/debugger-agent.c
```diff
@@ -3609,7 +3609,14 @@ create_event_list (EventKind event, GPtrArray *reqs, MonoJitInfo *ji, EventInfo
 
 	if (!reqs)
 		return NULL;
-
+	gboolean hasEverythingElse = FALSE;
+	gboolean isNewFilteredException = FALSE;
```
vargaz: This should use the mono naming convention i.e. underscores.

mono/mini/debugger-agent.c
```diff
@@ -3722,12 +3754,22 @@ create_event_list (EventKind event, GPtrArray *reqs, MonoJitInfo *ji, EventInfo
...
 
-			if (!filtered) {
+			if (!filtered && !isNewFilteredException) {
 				*suspend_policy = MAX (*suspend_policy, req->suspend_policy);
 				events = g_slist_append (events, GINT_TO_POINTER (req->id));
 			}
 		}
 	}
+	if (hasEverythingElse && !isAlreadyFiltered)
+	{
```
vargaz: Coding style.

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/ExceptionEventRequest.cs
thaystg: Done

mcs/class/Mono.Debugger.Soft/Mono.Debugger.Soft/VirtualMachine.cs
thaystg: Done

thaystg:
@monojenkins commit apidiff

thaystg:
@monojenkins backport 2019-08

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

PR #16830, opened 2019-09-13
[WIP] Mechanism for probing js and linear stack depths in wasm code

kg:
This adds a simple way for JS to probe the current JS and linear memory stack depths when we enter/exit WASM. Right now we don't seem to have any stack depths but it would be easy to regress to the point where we would - there are no guarantees that a given runtime will give us lots of stack space and what happens in the event of an overflow is poorly defined.

Right now based on some other analysis I've done most of our functions have stack frames of up to 1KB in size, but we have a few functions with frame sizes in excess of 8kb. My thinking is we might want to use this for an automated test to keep track of how much stack we use in specific scenarios, or at least provide this as a debugging tool if users are hitting stack overflows in production.

Still unfinished. I'm thinking it might be good to implement probes somewhere in the interpreter or for EM_ASM blocks or something, but I haven't figured out how best to do that.

Right now in a test blazor app I'm seeing JS stack depths of around 80 at max and linear + shadow stack sizes of like 120kb as a result.

### DIFF
--- a/sdks/wasm/Makefile
+++ b/sdks/wasm/Makefile
@@ -122,7 +122,7 @@ emsdk_env.sh:
 
 MONO_LIBS = $(TOP)/sdks/out/wasm-runtime-release/lib/{libmono-ee-interp.a,libmono-native.a,libmono-icall-table.a,libmonosgen-2.0.a,libmono-ilgen.a}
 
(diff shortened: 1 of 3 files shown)


jaykrell:
How do you measure shadow stack?
Counting locals?
What functions use more than 1k?
Fix them?

kg:
You'd pair the fn #s from the JS stack with local counts from analyzing the .wasm module, but that doesn't account for bad codegen in the JIT. I'm increasingly convinced it isn't possible to probe reliably with any method other than attaching a debugger

PR #16833, opened 2019-09-14
[interp] Restore 16 bytes of frame due to change to leave.

jaykrell:
Outline more of leave, as one piece.
Hypothetically this is easier to maintain, but it still kinda on the edge.
Just putting in a memset pushes it over.

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -2991,20 +2991,101 @@ mono_interp_calli_nat_dynamic_pinvoke (
 	interp_exec_method (child_frame, context, error);
 }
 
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

BrzVlad:
While I'm not particularly against this PR, I don't like big changes, especially if the alternative is a one line revert. https://github.com/mono/mono/pull/16887

jaykrell:
It is a little noisy but conceptually simple and small -- union + outline.

jaykrell:
This form also more easily affords more changes in leave, which is hypothetical. If such changes materialize, can still do this, easily enough.

PR #16834, opened 2019-09-14
[interp] Combine common code paths in calls.

jaykrell:
Note that CHECK_RESUME_STATE can be deferred to after return value placed. I think.

There are 3 related PRs.
One minimizes size, one minimizes conditional branches, one strikes a balance.
My favorate is the balance: https://github.com/mono/mono/pull/16836

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3028,11 +3028,7 @@ mono_interp_newobj_vt (
 	// stack for all the other recursive cases.
 	interp_exec_method (child_frame, context, error);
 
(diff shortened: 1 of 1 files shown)


jaykrell:
Choice is kinda between:

faster/bigger: https://github.com/mono/mono/pull/16835

or smaller: https://github.com/mono/mono/pull/16834

The size difference is not huge.
But the size difference for spltting the function wasn't huge either.
But probably can have it both ways.

One way would be, if the function were a macro, and each instance had different labels, then the other cases could jmp to the right label.
I can try that.

PR #16835, opened 2019-09-14
[interp] Remove a few instructions from calls.

jaykrell:
Classic size vs. speed tradeoff.
This makes four copies of call, across void/non-void, virtual/non-virtual,
so the conditionals become constants.

There are 3 related PRs.
One minimizes size, one minimizes conditional branches, one strikes a balance.
My favorate is the balance: https://github.com/mono/mono/pull/16836


jaykrell:
Codegen looks like an improvement.
One benchmark tested (sat solver) seemed like only noise.
A micro benchmark around call, esp. with random call patterns, would be good.

PR #16837, opened 2019-09-14
LLVM: Dump Module instead of Function in JIT mode

EgorBo:
Since each function has it's own module (in JIT mode) it makes sense to dump the whole module for `MONO_VERBOSE_METHOD` instead of just LLVM::Function.
It adds metadata, attributes and global variables (so it will be possible to copy-paste it to godbolt)

Example
```csharp
static int Test(int x)
{
    return x / 10;
}
```

Diff for Unoptimized IR: https://www.diffchecker.com/NrF413Ts 
Diff for Optimized IR: https://www.diffchecker.com/jbEkeFFS
So now we can see global variables and metadata for each function (and can easily paste them to godbolt.org)

NOTE: optimized IR for this case contains "dead" variables (`dump_module` allowed us to see such), we probably need some Module-wide optimizations to strip dead code such as `-globaldce`
NOTE2: why do we emit exceptions in the first place for this case at all?

### DIFF
--- a/mono/mini/mini-llvm-cpp.cpp
+++ b/mono/mini/mini-llvm-cpp.cpp
@@ -58,7 +58,8 @@ mono_llvm_dump_value (LLVMValueRef value)
 void
 mono_llvm_dump_module (LLVMModuleRef module)
 {
-	LLVMDumpModule (module);
+	fflush (stdout);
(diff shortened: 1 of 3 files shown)


EgorBo:
@monojenkins build failed

vargaz:
@monojenkins build failed

PR #16841, opened 2019-09-14
[BE] Fix small sized argument output with tracing enabled

NattyNarwhal:
This is a legacy of before the profiler was used for the trace
output; it would apparently get the raw stack pointer and try to
dereference things by recalculating the offset as needed (no for
little, yes for big). However, because the profiler returns the
actual fixed-up pointer, this recalculation isn't needed on BE,
but instead grabs garbage a bit ahead on the stack.

To fix, use what was the little endian case unconditionally.

Example before:

`[1: 20.70598 14] ENTER: System.Globalization.CultureInfo:.ctor (int,bool,bool)(this:700000000408fb0[System.Globalization.CultureInfo platform.exe], 271824864, -32, -32, )`

Example after:

`[1: 1.93377 14] ENTER: System.Globalization.CultureInfo:.ctor (int,bool,bool)(this:700000000408fb0[System.Globalization.CultureInfo platform.exe], 127, 0, 1)`

### DIFF
--- a/mono/mini/trace.c
+++ b/mono/mini/trace.c
@@ -105,20 +105,11 @@ string_to_utf8 (MonoString *s)
 }
 
 /*
- * cpos (ebp + arg_info[n].offset) points to the beginning of the
- * stack slot for this argument.  On little-endian systems, we can
- * simply dereference it. On big-endian systems, we need to adjust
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

PR #16846, opened 2019-09-15
[interp] Speed up forwarding this parameter.

jaykrell:


jaykrell:
Two nits, but kinda big ones:
1. Might want an underscore between P and 0.
2. Might want to put this after _vt and after a newline, because of the math that is done on opcodes. This just happens to work because it is at the end, subtle.

Issue #16716, opened 2019-09-06
[llvm] Emit intrinsics for mini_emit_memset/mini_emit_memcpy


filipnavara:
LLVM has intrinsics for memset/memcpy:

[‘llvm.memset.*’ Intrinsics](https://llvm.org/docs/LangRef.html#llvm-memset-intrinsics)
[‘llvm.memcpy’ Intrinsic](https://llvm.org/docs/LangRef.html#llvm-memcpy-intrinsic)

Currently Mono emits unrolled loops which are not too bad. However, it lacks vectorization which could improve the performance significantly.

Furthermore, `string.memset` should be made into intrinsic similarly to `Buffer.Memmove`.

EgorBo:
In theory LLVM is able to recognize some patterns and replace them with memset/memmove/memcpy :) see https://github.com/mono/mono/issues/16251 (`ZeroArray` sample)

filipnavara:
@EgorBo Aside from the volatile load/store we are also hitting manually unrolled loops here.

SamMonoRT:
Seems like we don't want to port this to dotnet\runtime repo. Fixes for mono\mono seem completed by issue linked above. Closing this issue. We'll work on any more Intrinsics in the runtime repo.

PR #16847, opened 2019-09-15
LLVM: use @llvm.memset where possible

EgorBo:
TODO: make sure the following methods end up as `@llvm.memset` (for primitive types)

- [x] **RuntimeImports.ZeroMemory**
- [ ] **Unsafe.InitBlockUnaligned** (currently emits `string.memset` call)

E.g. `new Span<byte>(data).Clean()` calls `SpanHelpers.ClearWithoutReferences`  which either calls ZeroMemory or InitBlockUnaligned https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/SpanHelpers.cs#L29-L32  (@llvm.meset shows better performance results than both)


filipnavara:
Issue #16716, for reference.

mono/mini/mini-ops.h
```diff
@@ -768,6 +768,7 @@ MINI_OP(OP_SAVE_LMF, "save_lmf", NONE, NONE, NONE)
 MINI_OP(OP_RESTORE_LMF, "restore_lmf", NONE, NONE, NONE)
 
 MINI_OP3(OP_MEMMOVE, "memmove", NONE, IREG, IREG, IREG)
+MINI_OP3(OP_MEMSET64, "memset64", NONE, IREG, IREG, IREG)
```
vargaz: The last one should be LREG.

vargaz:
Looks ok but some issues:
- gc polling for large sizes ?
- llvm might end up calling memset without a wrapper which can causes problems with unwinding/stack walks. This is one of the reasons we called a managed method.

akoeplinger:
Should we port this to dotnet/runtime? I think we probably don't need it in mono/mono.

EgorBo:
> Should we port this to dotnet/runtime? I think we probably don't need it in mono/mono.

Ah I don't think so, Zoltan mentioned possible issues with this so it's probably better to leave as is to avoid long gc pauses in some cases.

PR #16848, opened 2019-09-16
[interp] Use +=, *= etc. Seems to generate identical code though.

jaykrell:
i.e. a = a op b is *not* double evaluating a.


jaykrell:
oops wrong branch

PR #16683, opened 2019-09-05
[draft] [interp] Frame to 0x78 bytes.

jaykrell:

PR #16852, opened 2019-09-16
[netcore] Make Environment calls thread-safe

filipnavara:
Fixes #16683 

Closely mimics the CoreRT fix from dotnet/corert#7001. After first call to `GetEnvironmentVariables` or `SetEnvironmentVariable` all the environment variables are reflected only on the managed side. New processes inherit those.


jaykrell:
Why not interoperate with other code in the process?
i.e. everyone should call GetEnvironmentVariable, every time.
I think Windows C runtime has similar bugs though, i.e. it caches and does not interoperate.
Possibly obsolete behavior but cannot change w/o breaking things.
Possibly a mess.
Or maybe well done, but I doubt it.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
```diff
@@ -13,36 +13,91 @@ namespace System
...
+			Debug.Assert(variable != null);
+
+			if (s_environment == null) {
+				using (var h = RuntimeMarshal.MarshalString (variable)) {
+					return internalGetEnvironmentVariable_native (h.Value);
+				}
+			}
+
+			lock (s_environment) {
+				variable = TrimStringOnFirstZero (variable);
```
marek-safar: TrimStringOnFirstZero should not be under lock
Should it be called in managed for internalGetEnvironmentVariable_native too?

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
```diff
@@ -13,36 +13,91 @@ namespace System
...
+		static unsafe void SetEnvironmentVariableCore (string variable, string value)
+		{
+			Debug.Assert(variable != null);
 
-		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		private extern static string [] GetEnvironmentVariableNames ();
+			EnsureEnvironmentCached ();
+			lock (s_environment) {
+				variable = TrimStringOnFirstZero (variable);
+				value = value == null ? null : TrimStringOnFirstZero (value);
```
marek-safar: Remove the null check, it's not needed due to `string.IsNullOrEmpty` check bellow

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
```diff
@@ -13,36 +13,91 @@ namespace System
...
+				if (string.IsNullOrEmpty (value)) {
+					s_environment.Remove (variable);
+				} else {
+					s_environment[variable] = value;
```
marek-safar: Could be just `s_environment [variable] = TrimStringOnFirstZero (value);`

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
```diff
@@ -13,36 +13,91 @@ namespace System
...
-				return internalGetEnvironmentVariable_native (h.Value);
+			Debug.Assert(variable != null);
+
+			if (s_environment == null) {
+				using (var h = RuntimeMarshal.MarshalString (variable)) {
+					return internalGetEnvironmentVariable_native (h.Value);
+				}
+			}
+
+			lock (s_environment) {
```
marek-safar: Are you missing `EnsureEnvironmentCached ();` call?

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
```diff
@@ -13,36 +13,91 @@ namespace System
...
+			return results;
 		}
 
-		static unsafe void SetEnvironmentVariableCore (string variable, string value)
+		private static string TrimStringOnFirstZero (string value)
 		{
-			fixed (char *fixed_variable = variable)
-			fixed (char *fixed_value = value)
-				InternalSetEnvironmentVariable (fixed_variable, variable.Length, fixed_value, value?.Length ?? 0);
+			int index = value.IndexOf('\0');
```
marek-safar: Follow the coding style with space before `(`

filipnavara:
@jaykrell Because the Unix API is not thread-safe by design. This way already debated many times in the context of CoreCLR and there is no good solution. If you lock around the native APIs someone can still P/Invoke and make it a reliability problem that is difficult to diagnose. This merely follows the original design decisions and aligns the behaviour with CoreRT/CoreCLR more closely.

This was already blowing up in the Xunit tests many times and I am sick of re-running the CI lanes unnecessarily.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
filipnavara: Not missing, `GetEnvironmentVariableCore` has a fast path for `s_environment == null`.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
filipnavara: The null check prevents a crash in `TrimStringOnFirstZero`.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
filipnavara: No, the trimming could make the value empty and unset the environment variable.

jaykrell:
The C runtime is not thread safe? Ugh, crazy.
I understand a lock in one caller does not protect all callers.
I've seen that anti-pattern enough.

Is the initial call thread safe somehow?

But anyway, on Windows, GetEnvironmentVariable is thread safe.
I'll have to look otherwise..

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
filipnavara: The trimming on zero merely mimics what the C APIs already do implicitly in native code. They accept zero-terminated strings.

filipnavara:
> Is the initial call thread safe somehow?

No, it is not, in general. It's a single point of failure though which is better than random failures. CoreCLR always performs the initialization at predefined point in runtime startup. We defer it to first use which is not optimal but matches whatever CoreRT does.

> But anyway, on Windows, GetEnvironmentVariable is thread safe.

...and on Windows we use it and depend on it.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
filipnavara: Fixed.

netcore/System.Private.CoreLib/src/System/Environment.Unix.cs
marek-safar: Then please update the method signature to be `SetEnvironmentVariableCore (string variable, string? value)`

filipnavara: Done.

jaykrell:
Port to non-netcore?

filipnavara:
> Port to non-netcore?

Maybe. I will need to check the `Process` code to ensure it would not break anything.

filipnavara:

PR #16855, opened 2019-09-16
[runtime] Add portable cached array creation functions and replace gcc-specific impl.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:
I don't know, Just preserving semantics with imho cleaner code, but arguably less clean, granted.
I was raised, perhaps incorrectly, not to depend on gcc.

PR #16862, opened 2019-09-16
[debugger] Remove extraneous ErrorCode declaration

CoffeeFlux:
With the single point of exit, this means the main ErrorCode isn't being written to and uninitialized memory is returned.


CoffeeFlux:
@monojenkins build failed

PR #16863, opened 2019-09-16
[cominterop] Default to [in] parameter direction when not specified explicitly

nsivov:
Treat [In, Out], [In] and unspecified as input arguments. Important change from
current behavior is to allow parameter without directional attributes.

Use case example is _MethodInfo.Invoke() method called from native client.


marek-safar:
@lambdageek please review

vargaz:
@monojenkins build failed

Issue #16864, opened 2019-09-16
[arm64_32] fix remaining regression tests


lewurm:
The following regression tests are still failing with arm64_32 + interpreter:
* test_0_array_delegate_full_aot
* test_0_exception_in_cctor
* test_1_basic_filter_catch
* test_1234_complicated_filter_catch
* test_0_filter_caller_area

This is blocking Blocks #10641

SamMonoRT:
@SamMonoRT fyi

SamMonoRT:
@imhameed - please can you check with @lewurm if any action is needed for this issue or can be closed.

imhameed:
Quick note: this is pretty low priority right now. Ping me if you think otherwise.

Issue #16819, opened 2019-09-13
[coop][interp] mono_threads_enter_gc_safe_region Cannot transition thread 0x1952b000 from STATE_BLOCKING with DO_BLOCKING


lewurm:
Full trace: https://gist.github.com/lewurm/793bafc873abc3cc8d828c53eb1379c7

Environment: watch arm64_32 with interpreter.

It happens for example on this test:
https://github.com/xamarin/xamarin-macios/blob/fe9e1c42481b7d47563601912c541575699153ee/tests/monotouch-test/Foundation/ObjectTest.cs#L140-L150

Since it's working with watch armv7k fullAOT, I suspect that some transition is missing in the interpreter? 

The pattern is:
1. interpreter throws an exception
2. exception handling calls back into interpreter for running `finally` code
3. managed `finally` code does an icall
4. ends up calling `xamarin_release_managed_ref`/`xamarin_framework_peer_lock`

I suspect we need to do some state transition here around here?
https://github.com/mono/mono/blob/6908a1941b7278fe6711c75976f2434db55136f0/mono/mini/mini-exceptions.c#L2960-L2971

This blocks https://github.com/mono/mono/issues/16819

lewurm:
Another trace without finally:
https://gist.github.com/lewurm/7fb2f6c111f25d997d92a56068ccdf7e

lewurm:
Fixed by https://github.com/xamarin/xamarin-macios/pull/7036 and https://github.com/mono/mono/pull/16955

Issue #10641, opened 2018-09-17
New arm64_32 architecture (watchOS 5) 


marek-safar:
This is tracking task to collect information about arm64_32 architecture and what will be needed to support it.

It was introduced with Xcode 10 GM but there is no documentation yet. Example of disassembled code https://gist.github.com/rolfbjarne/24465d3fd9ff53e1a8919b7ea602d976

[ILP32 Whitepaper](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0490a/ar01s01.html)

## Milestone I (2018-10)

- [x] Release mode support (llvm_only/bitcode). This is done using the the AOT compiler for `armv7k` (the old watch) and let LLVM take care to generate correct code for `arm64_32`.

## Milestone II (2019-08)

- [x] Debug mode support for arm64_32 backend. This requires changes in our arm64 backend to be aware of `arm64_32`. It's needed because our managed debugger doesn't work with LLVM generated code.
- [ ] ~Extend iOS SDK such that we can use it for watchOS too. This allows use to develop `arm64_32` related changes outside of `Xamarin.iOS`.~

lewurm:
As a note: While it's tempting to ignore this in mono and use `bitcode-build-tool` as described here https://mobile.twitter.com/jckarter/status/1044719452418322432 it unfortunately isn't a long-term solution for us:
* Apple requires `arm64_32` bitcode in March next year
* Xcode 10 already requires that (you can hack around it by using Xcode 9.4 for _now_ )
* debugging doesn't work
* build time would be even slower with `bitcode-build-tool`

source: https://xamarinhq.slack.com/archives/C03CELHQQ/p1538456914000100

/cc @rolfbjarne

lewurm:
Using this comment to track related PRs (in case we need to backport changes at some point):
* https://github.com/mono/mono/pull/11652

rolfbjarne:
@marek-safar the `Debug mode support for arm64_32 backend` task clearly didn't make 2019-04, do you have an updated milestone for that task?

lewurm:
@rolfbjarne that's my fault, I abandoned that work unintentionally and never got back to it. Sorry about that. I do have some patches that make Debug mode work with the interpreter, that might be worthwhile enough. We can then evaluate if we need to make AOT work as well.

Will do it this or next sprint, so it should be 2019-08.

lewurm:
Related changes have been landed into 2019-08.

[...]

PR #16865, opened 2019-09-16
[arm64_32] make "Debug Mode" work on Watch series 4 with --interpreter

lewurm:
Main objective for reviewers: This PR shouldn't change anything for `arm64`.

Context: "Debug Mode" on the Apple Watch allows execution of regular machine code and thus is pretty similar to FullAOT on iPhone, i.e. it doesn't require bitcode, but _still_ doesn't allow `signal(3)` so cooperative suspend is required.

Apple Watch Series 4 and onward use an ARM64 chip, but with a different ABI, also known as `arm64_32`. This PR changes the AOT compiler and related trampolines so that the AOT compiler can generate all necessary code to run the interpreter on this target.

Changes:

--------------------------------------

[arm64_32] set register size to 8 in cross compiler too 

-----------------------------------------

 [machine] update `target_mgreg_t` definition

`SIZEOF_REGISTER` is usually the same as `TARGET_SIZEOF_VOID_P`, except when `MONO_ARCH_ILP32 is defined.

`target_mgreg_t` should represent the width of the hardware register, that is, `SIZEOF_REGISTER`.

---------------------------------------------

[arm64_32] fix AOT image layout 

---------------------------------------------

 [arm64_32] make trampolines bitwidth aware

* replace pointer loads/stores with bitwidth-dependent macro
* reflect `specific` trampolines so that only two instructions are used for ILP32, since the related pair in the data page only needs 2 * 32bits too
* replace some `sizeof (target_mgreg_t)` with `TARGET_SIZEOF_VOID_P` (former is hardware register size, latter is pointer size)

---------------------------------------------
These regression tests are _not_ passing: https://github.com/mono/mono/issues/16864 Also some of monotouch-tests are crashing, see https://github.com/mono/mono/issues/16819

Here is a screenshot of a debug session in VSMac:
Screenshot 2019-09-16 at 19 13 22

Contributes to https://github.com/mono/mono/issues/10641

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4607,6 +4607,7 @@ if test "x$host" != "x$target"; then
 	aarch64*darwin*_ilp32)
 		TARGET=ARM6432;
 		AC_DEFINE(MONO_ARCH_ILP32, 1, [64 bit mode with 4 byte longs and pointers])
+		sizeof_register=8
 		AC_DEFINE(TARGET_WATCHOS, 1, [...])
 		;;
 	aarch64-*)
--- a/mono/arch/arm64/arm64-codegen.h
+++ b/mono/arch/arm64/arm64-codegen.h
@@ -857,4 +857,14 @@ arm_encode_arith_imm (int imm, guint32 *shift)
 
 #define arm_mrs(p, rt, sysreg) arm_format_mrs ((p), (sysreg), (rt))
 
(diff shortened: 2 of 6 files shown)


lewurm:
@monojenkins squash

lewurm:
@monojenkins backport 2019-08

Issue #16373, opened 2019-08-20
Mono broke xunit AppDomain marshalling


jaredpar:
This bug is fairly serious. It's blocking our standard CI runs. Temporarily disabling Mono until we can get this resolved. 

## Steps to Reproduce

1. Enlist in https://github.com/dotnet/roslyn
2. Run restore.sh
3. Run build.sh
4. Run `test.sh --testMono`

### Current Behavior

The Roslyn tests which engage `AppDomain` remoting are all failing with the following:

```
    [FATAL ERROR] System.InvalidOperationException
      System.InvalidOperationException : Type 'System.Object' of type 'System.RuntimeType' does not derive from System.Object
      Stack Trace:
        
        
        Server stack trace: 
          at System.Runtime.Serialization.FormatterServices.GetParentTypes (System.RuntimeType parentType, System.RuntimeType[]& parentTypes, System.Int32& parentTypeCount) [0x0003f] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
          at System.Runtime.Serialization.FormatterServices.InternalGetSerializableMembers (System.RuntimeType type) [0x00089] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
          at System.Runtime.Serialization.FormatterServices+<>c__DisplayClass9_0.<GetSerializableMembers>b__0 (System.Runtime.Serialization.MemberHolder _) [0x00000] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
          at System.Collections.Concurrent.ConcurrentDictionary`2[TKey,TValue].GetOrAdd (TKey key, System.Func`2[T,TResult] valueFactory) [0x00034] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
          at System.Runtime.Serialization.FormatterServices.GetSerializableMembers (System.Type type, System.Runtime.Serialization.StreamingContext context) [0x0005e] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
          at System.Runtime.Serialization.Formatters.Binary.WriteObjectInfo.InitMemberInfo () [0x0003d] in <59b1fe55a7ce47ca99c88b42a0622585>:0 
[...]

PR #16868, opened 2019-09-16
Temporarily disable embedded ppdb data decompression

CoffeeFlux:
Will need to put out a new Ubuntu nightly after this is merged, and then they can re-add us to CI.

### DIFF
--- a/mono/metadata/debug-mono-ppdb.c
+++ b/mono/metadata/debug-mono-ppdb.c
@@ -172,9 +172,10 @@ mono_ppdb_load_file (MonoImage *image, const guint8 *raw_contents, int size)
 		return NULL;
 	}
 
+// Temporarily disabled to unblock Roslyn
+#if HOST_WIN32 //|| HAVE_SYS_ZLIB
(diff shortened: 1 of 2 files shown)


lambdageek:
This addresses #16373. We will work on a proper fix, but this at least unblocks Roslyn CI

mono/metadata/debug-mono-ppdb.c
```diff
@@ -172,9 +172,10 @@ mono_ppdb_load_file (MonoImage *image, const guint8 *raw_contents, int size)
 		return NULL;
 	}
 
+// Temporarily disabled to unblock Roslyn
+#if 0//HAVE_SYS_ZLIB || HOST_WIN32
```
jaykrell: Isn’t Win32 ok?

CoffeeFlux: Yeah, should be. Will fix.

marek-safar:
@monojenkins backport 2019-08

PR #16874, opened 2019-09-16
[llvm] Use IntPtrType () for byref arguments instead of a pointer, the latter causes problems when pointer operations are used along with phi nodes. Fixes https://github.com/mono/mono/issues/16808.

vargaz:


vargaz:
@monojenkins build failed

Issue #16867, opened 2019-09-16
stackalloc always zeros memory


EgorBo:
CoreCLR's stackalloc is up to 10x times faster than Mono's one when it's allowed to allocate data without zeroing (but Mono always zeros it).

```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
static unsafe void Process(byte* data) {}

[MethodImpl(MethodImplOptions.NoInlining)]
static unsafe void DoWork(int size)
{
    byte* data = stackalloc byte[size];
    Process(data);
}
```

The `DoWork` method doesn't have `InitLocals` section in IL (there are no local vars).
Also .NET Core has a [linker-step](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/ILLink.targets#L83) to drop all `InitLocas` in all methods in BCL libs (at least for System.Private.CoreLib.dll)
It allows CoreCLR to stackalloc arrays without zeroing them (if InitLocals is empty).
Mono always zeros such arrays (emits @llvm.memset) and it hurts performance in some cases a bit.
According to the spec, stackalloc is not required to zero data.

So I am not saying we should not zero (memset) stackalloc, but only in cases when InitLocals is not empty just like CoreCLR does. 
BTW, when RuyJIT actually zeros data for stackalloc, our llvm.memset looks a bit more efficient for arrays > 100 items, see https://github.com/dotnet/coreclr/issues/13827#issuecomment-531947296

EgorBo:
asm for DoWork (RuyJIT):
```asm
; Method StackAlloc_VarSize:DoWork(int)

G_M8267_IG01:
       push     rbp
       sub      rsp, 48
       lea      rbp, [rsp+20H]
       mov      rax, 0xD1FFAB1E
       mov      qword ptr [rbp+08H], rax

G_M8267_IG02:
       mov      ecx, ecx
       test     rcx, rcx
       je       SHORT G_M8267_IG04
       mov      rax, rcx
       add      rax, 15
       and      rax, -16
       add      rsp, 32
       neg      rax
       add      rax, rsp
       jb       SHORT G_M8267_IG03
       xor      rax, rax

G_M8267_IG03:
       test     dword ptr [rsp], esp
       mov      rdx, rsp
       sub      rdx, 0x1000
       mov      rsp, rdx
       cmp      rsp, rax
       jae      SHORT G_M8267_IG03
       mov      rsp, rax
       test     dword ptr [rsp], esp
       sub      rsp, 32
[...]

PR #16881, opened 2019-09-17
Do not always zero localalloc.

jaykrell:
Note that we will still touch the pages (mono_emit_stack_alloc).

Should fix https://github.com/mono/mono/issues/16867.

### DIFF
--- a/mono/mini/method-to-ir.c
+++ b/mono/mini/method-to-ir.c
@@ -10929,7 +10929,7 @@ mono_ldptr:
 			MONO_ADD_INS (cfg->cbb, ins);
 
 			cfg->flags |= MONO_CFG_HAS_ALLOCA;
-			if (init_locals)
+			if (header->init_locals)
 				ins->flags |= MONO_INST_INIT;
 
 			MONO_START_BB (cfg, end_bb);
--- a/mono/tests/Makefile.am
+++ b/mono/tests/Makefile.am
@@ -1027,6 +1027,7 @@ TESTS_IL_SRC=			\
 	gh-13057_mono_local_emulate_ops_av.il \
 	module-cctor-entrypoint.il \
 	bug-gh-9706.il \
+	localloc-noinit.il \
 	calli_native.il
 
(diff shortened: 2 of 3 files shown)


marek-safar:
Could you think of any test for this?

jaykrell:
Fair point, yes, I think I can. Not totally portable, but portable enough.

jaykrell:
Specifically one should be able to observe the "uninitialized" values by first calling a function that initializes them, and then accessing the stale values on the stack. I can try that.

jaykrell:
I don't really like "writing" IL when it is actually C# + ildasm.
Anyway there is a test there now.

jaykrell:
Note that this an observable behavior change. But one difficult to see in C#. Still spec conformant, and faster (I didn't verify that).

jaykrell:
Fails interpreter, not surprising.

jaykrell:
Follow the links. This seems a thorny matter. Not really sure should change anything here and now. There was supposed to be an attribute for this but GitHub issue still open.

jaykrell:
Then again, if this PR matches coreclr? Also the check for unsafe seemed reasonable and suggested elsewhere but I think it is the wrong unsafe — command line option, not function flag.

jaykrell:
@monojenkins build failed

jaykrell:
Test is very sensitives, fails many configurations.

jaykrell:
The test is much more sensitive than I'd think and rarely works.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build

jaykrell:
I wonder if the conditions should be combined.
I.e. to restore the previous unsafe optimization.
I think it was counter-spec sort
of, but more deterministic in C#.

Basic problem is C# spec is
reasonable, .NET is semi-reasonable
(should be two flags, not one, or per instruction),
and C# implementation is not reasonable,
basically leaving no choice at all,
except corner case.

jaykrell:
@monojenkins build failed

PR #16882, opened 2019-09-17
Inline TLS access.

jaykrell:
The same functions are referenced by JIT and runtime.
The JIT uses cannot be inlined. The runtime can.
Since I could not get anything else to build and the change is simple and mechanical, now two sets of functions are provided, inline and extern. Extern just calls inline.
This PR appears to change all the uses, but it is only the JIT uses. The runtime ones just work and keep the old names.

We should also consider building desktop with LTO/LTCG, which would do all this automatically, smarter, more generally. A quick experiment shows it grows the runtime size, which is why I say desktop.

### DIFF
--- a/mono/utils/mono-tls.h
+++ b/mono/utils/mono-tls.h
@@ -16,6 +16,15 @@
 #include <config.h>
 #include <glib.h>
 #include <mono/utils/mono-forward-internal.h>
+#include <mono/utils/mach-support.h>
+
+#if __cplusplus
+#define MONO_INLINE inline
+#elif _MSC_VER
+#define MONO_INLINE __inline
+#else
+#define MONO_INLINE static inline
+#endif
 
 /* TLS entries used by the runtime */
 // This ordering is mimiced in MONO_JIT_ICALLS (alphabetical).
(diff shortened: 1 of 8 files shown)


jaykrell:
Huh? This somehow trips the wasm-ld bug. Will retry rebase to master or something.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mono/utils/mono-tls.h
```diff
@@ -101,18 +110,190 @@ mono_native_tls_set_value (MonoNativeTlsKey key, gpointer value)
...
-void mono_tls_set_lmf_addr 	   (MonoLMF           **value);
+
+G_EXTERN_C MonoInternalThread *mono_tls_get_thread_extern (void);
+G_EXTERN_C MonoJitTlsData     *mono_tls_get_jit_tls_extern (void);
+G_EXTERN_C MonoDomain *mono_tls_get_domain_extern (void);
+G_EXTERN_C SgenThreadInfo *mono_tls_get_sgen_thread_info_extern (void);
+G_EXTERN_C MonoLMF       **mono_tls_get_lmf_addr_extern (void);
+
+/*
+ * On all platforms we should be able to use either __thread or __declspec (thread)
```
lambdageek: Did you mean to also leave the original copy of these paragraphs in `mono-tls.c`

jaykrell: I meant to move almost everything to the .h.
I kinda didn't think about where is most readable.
And larger .h is bad for compile perf perhaps, at least it used to be.
I can look at it again and try to decide what is most readable.

I'd really like to remove the non-default `__thread` stuff, but I kept it.

PR #16883, opened 2019-09-17
[netcore] Fix compiler warnings

akoeplinger:
This fixes the compiler warnings when building netcore-only on macOS.

Mostly warnings about functions without prototypes for icalls that don't exist on netcore so wrapped them in `ENABLE_NETCORE` ifdef.

### DIFF
--- a/mono/metadata/mono-route.c
+++ b/mono/metadata/mono-route.c
@@ -9,6 +9,7 @@
 
 #include "config.h"
 
+#ifndef ENABLE_NETCORE
 #if HOST_DARWIN || HOST_BSD
 
 #include <sys/types.h>
@@ -154,6 +155,6 @@ gateway_from_rtm(struct rt_msghdr *rtm)
 }
 
 #endif
+#endif
 
 extern const char mono_route_empty_file_no_warning;
-extern const char mono_route_empty_file_no_warning = 0;
--- a/mono/metadata/mono-route.c
+++ b/mono/metadata/mono-route.c
@@ -158,3 +158,4 @@ gateway_from_rtm(struct rt_msghdr *rtm)
 #endif
 
 extern const char mono_route_empty_file_no_warning;
+const char mono_route_empty_file_no_warning = 0;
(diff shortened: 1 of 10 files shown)


mono/metadata/w32socket.c
```diff
@@ -742,6 +742,7 @@ get_socket_assembly (void)
 	return domain->socket_assembly;
```
marek-safar: I am surprised we need this file at all (is that for debugger?)

mono/sgen/sgen-debug.c
```diff
@@ -944,20 +944,20 @@ sgen_scan_for_registered_roots_in_domain (MonoDomain *domain, int root_type)
 static gboolean
 is_xdomain_ref_allowed (GCObject **ptr, GCObject *obj, MonoDomain *domain)
 {
+#ifndef ENABLE_NETCORE
 	MonoObject *o = (MonoObject*)(obj);
 	size_t offset = (char*)(ptr) - (char*)o;
 
-#ifndef ENABLE_NETCORE
```
jaykrell: netcore firmly implies no remoting?

mono/metadata/mono-route.c
```diff
@@ -153,7 +154,7 @@ gateway_from_rtm(struct rt_msghdr *rtm)
 	return 0;
 }
 
+#endif
 #endif
 
 extern const char mono_route_empty_file_no_warning;
-extern const char mono_route_empty_file_no_warning = 0;
```
jaykrell: My mistake. The correct fix is to remove "extern" on the second line.
Otherwise you are going to get link/lib warnings.
The reason for the two lines is ridiculous gcc warnings.
Otherwise just the last line would suffice.

mono/metadata/w32socket.c
filipnavara: Yes IIRC

mono/sgen/sgen-debug.c
filipnavara: Yes

PR #16894, opened 2019-09-17
[corlib] Fix building nunit-lite twice

akoeplinger:
We sometimes saw the following build error in Mono SDKs builds on Jenkins:

```
[2019-09-16T01:03:30.709Z] CSC     [xammac] nunit-lite-console.exe
[2019-09-16T01:03:30.709Z] error CS0009: Metadata file '/Users/builder/jenkins/workspace/archive-mono/2019-08/mac/release/mcs/class/lib/xammac/nunitlite.dll' could not be opened -- PE image doesn't contain managed metadata.
```

Taking a closer look we found out that nunit-lite was trying to be built twice for the same profile which caused a race where the file was corrupted.

The reason for this is that the corlib Makefile defines two test assemblies: the normal `PROFILE_corlib_test.dll` and `BinarySerializationOverVersionsTest.dll` and
integrating the BinarySerializationOverVersionsTest.dll via the `test-vts` target incorrectly overwrote the `test` target instead of using `test-local`.

/cc @rolfbjarne


mcs/class/corlib/Makefile
```diff
@@ -298,11 +298,8 @@ NO_VTS_TEST = yes
 endif
 
 ifndef NO_VTS_TEST
-test: test-vts
-run-test: run-test-vts
-else
-test:
-run-test:
+test-local: test-vts
+run-test-local: run-test-vts
```
akoeplinger: this is the important change, the other changes are just for consistency/cosmetic.

akoeplinger:
@monojenkins backport 2019-08

akoeplinger:
@monojenkins backport 2019-06

PR #16898, opened 2019-09-17
Optimize "constant_string".Length

EgorBo:
```csharp
static int Test() => "Hello".Length; // replace with just `5`
```
Before:
```asm
movabs rax,0x7fd6c1758160
mov    eax,DWORD PTR [rax]
ret 
```
Now:
```asm
mov    eax,0x5
ret
```

Also, handles cases after inlining, e.g.:
```csharp
bool Validate(string str)
{
    return str.Length > 0 && str.Length <= 100;
}

bool Test()
{
    return Validate("Hello");  // Validate() will be inlined
}
```
New codegen for `Test`:
```asm
mov    eax, 1     ; return true
ret  
```

I have a similar PR for CoreCLR: https://github.com/dotnet/runtime/pull/1378
jit-diff report found >30k bytes of improvements.


kg:
@monojenkins build failed

mono/mini/decompose.c
```diff
@@ -1587,8 +1587,20 @@ mono_decompose_array_access_opts (MonoCompile *cfg)
 					}
 					break;
 				case OP_STRLEN:
-					MONO_EMIT_NEW_LOAD_MEMBASE_OP_FLAGS (cfg, OP_LOADI4_MEMBASE, ins->dreg,
-														 ins->sreg1, MONO_STRUCT_OFFSET (MonoString, length), ins->flags | MONO_INST_INVARIANT_LOAD);
+					if (ins->prev && ins->prev->opcode == OP_I8CONST) {
+						// Optimize String.get_length for ldstr, e.g.:
```
vargaz: Need to check that ins->prev->dreg == ins->sreg1.

mono/mini/decompose.c
```diff
@@ -1587,8 +1587,20 @@ mono_decompose_array_access_opts (MonoCompile *cfg)
...
+						// "hello".Length => 5
+						mono_domain_lock(cfg->domain);
+						MonoGHashTable* table = cfg->domain->ldstr_table;
```
vargaz: Coding style.

PR #16905, opened 2019-09-17
Clean up assembly name parsers' handling of special chars

CoffeeFlux:
Resolves parser issues discussed in https://github.com/mono/mono/pull/16425#issuecomment-529661033, along with some minor cleanup. 

The commit making TypeNameParser consistent can be removed if the line damage isn't worth it, but the file is currently a mess, so I'm of the opinion that it is.

This is maybe worth rebasing for the history, perhaps with the last two commits squashed?

### DIFF
--- a/mono/metadata/class.c
+++ b/mono/metadata/class.c
@@ -289,31 +289,14 @@ mono_type_name_check_byref (MonoType *type, GString *str)
 		g_string_append_c (str, '&');
 }
 
-/**
- * mono_identifier_escape_type_name_chars:
- * \param str a destination string
- * \param identifier an IDENTIFIER in internal form
- *
- * \returns \p str
- *
- * The displayed form of the identifier is appended to str.
- *
- * The displayed form of an identifier has the characters ,+&*[]\
- * that have special meaning in type names escaped with a preceeding
- * backslash (\) character.
- */
-static GString*
-mono_identifier_escape_type_name_chars (GString* str, const char* identifier)
+static char*
+escape_special_chars (const char* identifier)
 {
-	if (!identifier)
-		return str;
-
-	size_t n = str->len;
-	// reserve space for common case: there will be no escaped characters.
-	g_string_set_size(str, n + strlen(identifier));
-	g_string_set_size(str, n);
-
-	for (const char* s = identifier; *s != 0 ; s++) {
+	size_t id_len = strlen (identifier);
+	// Assume the worst case, and thus only allocate once
+	char *res = g_malloc (id_len * 2 + 1);
+	char *res_ptr = res;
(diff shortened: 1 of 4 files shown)


mono/metadata/class.c
```diff
@@ -289,31 +289,48 @@ mono_type_name_check_byref (MonoType *type, GString *str)
 		g_string_append_c (str, '&');
 }
 
+static char*
+escape_special_chars (const char* identifier)
+{
+	// Assume the worst case, and thus only allocate once
+	char *res = g_malloc (strlen (identifier) * 2 + 1);
```
jaykrell: Please store the length in a local.

jaykrell: Consider making two passes to save memory.

mono/metadata/class.c
```diff
@@ -289,31 +289,48 @@ mono_type_name_check_byref (MonoType *type, GString *str)
...
+	char *res_ptr = res;
+	for (const char *s = identifier; *s != 0; s++) {
+		switch (*s) {
+		case ',':
+		case '+':
+		case '&':
+		case '*':
```
jaykrell: Consider a bitset with these characters. Or some way to avoid duplicating them. A macro. Etc.

mono/metadata/class.c
CoffeeFlux: Yeah, wasn't sure on this. I opted to just allocate aggressively since that's what the eglib string functions seemed to do.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
Okay, this seems to be causing way too many nasty regressions in weird edge cases. I guess for now the best option is to document that we don't properly escape the names and leave it be. The rest of the changes are still good.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:

CoffeeFlux:

netcore/System.Private.CoreLib/src/System/TypeNameParser.cs
```diff
@@ -157,25 +157,40 @@ static Type ResolveType (Assembly assembly, List<string> names, Func<Assembly, s
...
+
+			return sb.ToString ();
+		}
+
+		static string UnescapeTypeName (string name)
+		{
+			if (name.IndexOfAny (SPECIAL_CHARS) < 0)
+				return name;
 
 			var sb = new StringBuilder ();
```
lambdageek: nit: can give the string builder an initial capacity here so that it doesn't have to reallocate too often.

CoffeeFlux:
@monojenkins build failed

PR #16873, opened 2019-09-16
[Coop] InterlockedCompare/Exchange (object/T)

jaykrell:
- Usually intrinsified.
 - Always?
 - Avoid coop handles by manually always using ref to object.
 - For efficiency on input, and efficiency wrt MonoError.
 - For correctness on output (due to aliasing in the wrappers).
 - To probably avoid assertion failure in runtime around icall marshaling of generics.
 - Unclear on location.

PR #16908, opened 2019-09-18
[RFC] Pass through out and inout parameters in coop marshaling, instead of using local intermediates.

jaykrell:
This reverses a decision that I, Aleksey, Rodrigo made about a year ago.

The concern, which I still have, is that the out/ref parameters could live in native heap/stack/global, unlikely but possible, and that probably does not suffice for GC.
The current and previous coop marshaling makes it work.
This passthrough pattern does not.

I think the native code in those cases is supposed to create a gchandle or coop handle itself.
But the current/previous coop marshaling provides a large window of coverage.

The non-pass through has some cost, and breaks the ability to pass interlocked inout parameters, because, "identity". (see https://github.com/mono/mono/pull/16873)

## Same question restated.

Let's say I have:

```
MonoObject* native_global;

struct {
 MonoObject *o;
} *native_heap;


void native ()
{
  native_global = gc_alloc...
  native_heap = malloc(...)
  native_heap->o = gc_alloc...
  MonoObject *native_stack = gc_alloc...

 // note the ampersands
  managed (&native_global, &native_heap->o, &naive_stack);
}

void managed(ref object a, ref object b, ref object c)
{
 ...
}
```

Is that ok and correct and all?

Does the ref in managed frame suffice?

Should/must native make gchandles for the global and heap
and gchandle or coop handle for the stack?
(or even coop handles for the global/heap?)

If it is correct, it is a nice little reduction in coop cost,
and would let coop marshal things where the intermediate local
breaks semantics (InterlockedCompareExchange), though the generics
might also not work, and the MonoError is excess, and
I believe can be worked around anyway, and granted, inout/out parameters
are the vast minority anyway.

--

Historically, intermediates were required, because coop handles were pointers to TLS.
The "fast" managed coop handles enable this form then.
Either fast pointers to intermediate locals, or fast passthrough out/inout.
(and pointer to param for in).

--

There are very few out/ref parameters to icalls anyway.

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6339,6 +6339,8 @@ signature_param_uses_handles (MonoMethodSignature *sig, MonoMethodSignature *gen
 		return ICALL_HANDLES_WRAP_NONE;
 }
 
+const char mono_coop_marshal_out_inout_through_intermediate_locals = FALSE;
+
 static void
 emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, MonoMethodSignature *csig, gboolean check_exceptions, gboolean aot, MonoMethodPInvoke *piinfo)
 {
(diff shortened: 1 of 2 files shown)


jaykrell:
@monojenkins build failed

akoeplinger:

PR #16912, opened 2019-09-18
Add .NET 4.8 reference assemblies

akoeplinger:
Brings in https://github.com/mono/reference-assemblies/pull/20


akoeplinger:
@directhex this will need the usual updates to the linux packaging config.

PR #16914, opened 2019-09-18
Fix StackOverflow in ApplicationTrustEnumerator

EgorBo:
`ApplicationTrustEnumerator ` used to call itself. The following snippet crashed with SO:
```csharp
using System.Security.Policy;
public class Program {
        public static void Main (string[] args)
        {
            ApplicationTrustCollection col = ApplicationSecurityManager.UserApplicationTrusts;
            col.GetEnumerator ();
        }
}
```


EgorBo:
@monojenkins build failed

PR #16915, opened 2019-09-18
Optional only build MSVC runtime using make file system on Windows.

lateralusX:
Add a new option to only build MSVC version of mono runtime and native tools, --enable-msvc-only. In the past, --enable-msvc build both mingw as well as MSVC runtime, this commit adds an option to only build MSVC version of runtime and native tools using make file system.

Commit updates all native make file making sure mingw/gcc builds are not done if --enable-msvc-only has been used. It also makes sure we build and use MSVC based tools as part of build and tests.

Result will be a build without any mingw/gcc artifacts, but where all existing make files still works with the MSVC build artifacts. NOTE, this PR is just a first step, optional disabling the gcc/mingw build on Windows. It does not look into the install targets or other automake features, except making sure everything builds and can be tested in build tree.

This commits is also preparing to switch the .NETCore build over to use MSVC build runtime in order to get LLVM support, something that is currently only supported on Windows MSVC build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6027,12 +6027,24 @@ AC_DEFINE(MONO_INSIDE_RUNTIME,1, [Disable banned functions from being used by th
 mono_build_root=`pwd`
 AC_SUBST(mono_build_root)
 
+AC_ARG_ENABLE(msvc_only,[  --enable-msvc-only    Build only with Visual Studio (msvc) (defaults to no). Disable gcc build.], enable_msvc_only=$enableval, enable_msvc_only=no)
+if test x$host_win32 = xno; then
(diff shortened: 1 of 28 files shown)


lateralusX:
@lambdageek Review?

directhex:
@monojenkins build failed

lateralusX:
> LGTM
> I'm surprised you didn't need to modify mono/profiler/Makefile.am

Looks like it is not build at all under Windows, if !HOST_WIN32. I ported log profiler over to MSVC a while back, so we at least build that, don't know why not profiler libs have never been build on Windows in the past though, probably could/should at some point.

PR #16917, opened 2019-09-18
[bcl/netcore] Implement Marshal.SetLastWin32Error

filipnavara:
It is used by `SafeHandle` to restore last error from `Release`. This is especially important when doing P/Invoke marshalling where the error of the P/Invoked API would get overwritten if it was holding last reference for the particular `SafeHandle`.

### DIFF
--- a/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
+++ b/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs
@@ -1936,9 +1936,8 @@ namespace System.Runtime.InteropServices
 			return GetFunctionPointerForDelegateInternal ((Delegate)(object)d);
 		}
 
-		internal static void SetLastWin32Error (int error)
-		{
-		}
+		[MethodImplAttribute(MethodImplOptions.InternalCall)]
+		internal static extern void SetLastWin32Error (int error);
 
 #if NETCORE
 		internal static IntPtr AllocBSTR (int length)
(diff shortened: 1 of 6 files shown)


filipnavara:

jaykrell:
I also wonder if mono_marshal_set_last_error should also preserve errno on Windows.

jaykrell:
I guess do whatever coreclr does though.

configure.ac
```diff
@@ -62,7 +62,7 @@ MONO_VERSION_BUILD=`echo $VERSION | cut -d . -f 3`
 # This line is parsed by tools besides autoconf, such as msvc/mono.winconfig.targets.
 # It should remain in the format they expect.
 #
-MONO_CORLIB_VERSION=8E172F4D-82A3-4F55-8D7C-C8E8E0FA58AE
+MONO_CORLIB_VERSION=8E172F4D-82A3-4F55-8D7C-C8E8E0FA58AF
```
vargaz: No need for a corlib bump.

filipnavara: Oh, alright, I will revert it when I get back to my computer.

jaykrell: This confuses me. While a newer same-guid BCL will continue to work with old and new runtime, there exist older same-guid BCL that will not work with this newer runtime.

vargaz:
It won't work, but the runtime will detect the missing icall and print an error. The corlib version is for detecting changes which cannot be detected by the runtime.

vargaz:
Bumping the corlib version is painful so we shouldn't do it for new icalls etc.

jaykrell:
Ah.

So, do you mean, you still have to build deb with monolite for the CI to succeed?

Which to me is the pain, and it is borne by the PR author.

I thought maybe there is a time window where the bump affects everyone, waiting for CI to build master after the PR commits, but I'm not sure I've seen it.

vargaz:
I mean bumping corlib version is painful in all kinds of ways, so we should avoid it if the problems caused by not bumping are not big, like in this case, since the runtime will detect the problem anyway.

filipnavara:
It would be helpful to write some guideline for the version bump (or did I miss it is already written somewhere?!). I thought I may not need for adding icall but I was not sure.

vargaz:
There was a comment:
- * Changes which are already detected at runtime, like the addition
- * of icalls, do not require an increment.
which got deleted when the version number was moved to configure.ac.

filipnavara:

filipnavara:

filipnavara:
@monojenkins build OS X x64

filipnavara:

PR #16562, opened 2019-08-29
MAX_PATH reduction.

jaykrell:
Steps on toes somewhat of https://github.com/mono/mono/pull/16553 but tried not to entirely.

PR #16920, opened 2019-09-18
Use eglib win32 helper functions without MAX_PATH limit

CoffeeFlux:
Update to use Jay's helper functions introduced in #16562


CoffeeFlux:
Oops, forgot we were using the mono_w32process stuff for cross-platform compat. Will update the signatures for the relevant functions.

CoffeeFlux:
Okay, less small than I'd hoped for, but still should be pretty quick to review.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:

CoffeeFlux:
@monojenkins build failed

jaykrell:
I'm not sure careful maintenance of declarations is worth the tedium and noise, vs.  deferring the check to be a link error. At least on Windows it is always a static link error. Maybe Unix allows the unresolved external when building .so files.

CoffeeFlux:
@monojenkins build failed

PR #16921, opened 2019-09-18
Use new eglib string helper functions

CoffeeFlux:
Another quick PR to use new functions added by Jay a little while back, this time in #15530. The most important change here is avoiding the extra unnecessary allocation.

### DIFF
--- a/mono/metadata/reflection.c
+++ b/mono/metadata/reflection.c
@@ -1521,7 +1521,7 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 	while (g_ascii_isspace (*p))
 		p++;
 	while (*p) {
-		if ((*p == 'V' || *p == 'v') && g_ascii_strncasecmp (p, "Version=", 8) == 0) {
+		if ((*p == 'V' || *p == 'v') && g_ascii_strncasecmp (p, G_STRING_CONSTANT_AND_LENGTH ("Version=")) == 0) {
 			p += 8;
 			assembly->major = strtoul (p, &s, 10);
 			if (s == p || *s != '.')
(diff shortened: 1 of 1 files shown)


jaykrell:
It'd be nice to take out `+= 8`.

Like:

```
gboolean
g_ascaii_skip_prefix_case (char **pp, gsize *plength, const char *prefix, gsize prefix_length)
{
 gsize length = *plength;
 if (length < prefix_length)
   return FALSE;
 char *p = *pp;
 if (g_ascii_equal_caseinsensitive (p, length, prefix, prefix_lengt) {
    *pp = p + prefix_length;
    *plength = len - prefix_length;
    return TRUE;
  }
  return FALSE;
}
```

or maybe:

```
gsize
g_ascaii_check_prefix_case (const char *p, gsize length, const char *prefix, gsize prefix_length)
{
 return g_ascii_equal_caseinsensitive (p, length, prefix, prefix_length) ? prefix_length : 0;
}
```

```
if ((*p == 'V' || *p == 'v') && g_ascii_strncasecmp (p, G_STRING_CONSTANT_AND_LENGTH ("Version=")) == 0) { 
			p += 8; 
```

becomes then, either:

```
len = strlen (p);
g_ascaii_skip_prefix_case (&p, &len, G_STRING_CONSTANT_AND_LENGTH ("Version="));
```

or:

```
gsize match = g_ascaii_skip_prefix_case (p, strlen (p), G_STRING_CONSTANT_AND_LENGTH ("Version="));
if (match)
 p += match;
```

jaykrell:
The lengths are still duplicated, so the helpers don't help enough, in this case (they can be useful otherwise).

mono/metadata/reflection.c
```diff
@@ -1550,9 +1550,9 @@ assembly_name_to_aname (MonoAssemblyName *assembly, char *p) {
 					p++;
 				}
 			}
-		} else if ((*p == 'P' || *p == 'p') && g_ascii_strncasecmp (p, "PublicKeyToken=", 15) == 0) {
+		} else if ((*p == 'P' || *p == 'p') && g_ascii_strncasecmp (p, G_STRING_CONSTANT_AND_LENGTH ("PublicKeyToken=")) == 0) {
 			p += 15;
-			if (strncmp (p, "null", 4) == 0) {
+			if (strncmp (p, G_STRING_CONSTANT_AND_LENGTH ("null")) == 0) {
```
lambdageek: Build fails on Linux.  apparently `strncmp` is a macro and the preprocessor is unhappy with this line.

CoffeeFlux: Yeah, I just never got around to fixing this. I'm cleaning up various backlogged issues right now, will get to this eventually.

jaykrell: Yeah, we discussed this, a while ago. To wrap it in g_strncmp or such.

Issue #16709, opened 2019-09-06
Disposing the FileSystemWatcher does not close associated file descriptors


mrward:
## Steps to Reproduce

Related problem - https://github.com/mono/mono/issues/15931

1. Sample project - https://github.com/mrward/test-file-watcher-dispose - multi-target project that builds .NET 4.7.2 and .NET Core 2.1
2. Build this project.
3. From the command line run `mono bin/debug/net472/FileWatcherDisposeTest.exe`
4. Run `lsof -p pid`. Process id is output from the console app.
5. See that the file watcher creates file descriptors for several directories.
6. After 10 seconds the file watcher will be disposed.
7. Run `lsof -p pid` again.

### Current Behavior

lsof still lists the directories.

```
mono-sgen 3972 user    4r DIR   1,12      170   51740695 /Volumes/Drive/projects/tests/FileWatcherDisposeTest/FileWatcherDisposeTest/bin/Debug/net472
mono-sgen 3972 user    5r DIR   1,12      136   51740600 /Volumes/Drive/projects/tests/FileWatcherDisposeTest/FileWatcherDisposeTest/bin/Debug
mono-sgen 3972 user    6r DIR   1,12      102   51740599 /Volumes/Drive/projects/tests/FileWatcherDisposeTest/FileWatcherDisposeTest/bin
mono-sgen 3972 user    7r DIR   1,12      204   51740594 /Volumes/Drive/projects/tests/FileWatcherDisposeTest/FileWatcherDisposeTest
mono-sgen 3972 user    8r DIR   1,12      170   51740593 /Volumes/Drive/projects/tests/FileWatcherDisposeTest
mono-sgen 3972 user    9r DIR   1,12     6324     185300 /Volumes/Drive/projects/tests
mono-sgen 3972 user   10r DIR   1,12     2312        278 /Volumes/Drive/projects
mono-sgen 3972 user   11r DIR   1,12      646          2 /Volumes/Drive
mono-sgen 3972 user   12r DIR    1,4      160     420284 /Volumes
```
If you change the console project to run GC.Collect after the dispose then lsof will not list the directories.

Also affects .NET Core 2.1 (bug the GC.Collect does not workaround the problem there)

### Expected Behavior

File watcher directories not listed by lsof.

Note that 

## On which platforms did you notice this

[x] macOS
[ ] Linux
[ ] Windows

**Version Used**:

[...]

PR #16922, opened 2019-09-18
[System] Make FileSystemWatcher backend non-static

lambdageek:
The FileSystemWatcher `watcher` field is the backend IFileWatcher that's
supposed to be used for the actual operations.

Each backend has a `bool GetInstance (out watcher)` method that returns a
singleton object, so each new instance of FileSystemWatcher gets the same
IFileWatcher instance.  (They will get different `watcher_handle` instances
which are used by some backends to uniquely identify this FileSystemWatcher).

However when the first FileSystemWatcher instance is `Dispose`d, it will set
`watcher = null` which will prevent the remaining instances from calling

    watcher?.StopDispatching (watcher_handle);
    watcher?.Dispose (watcher_handle);

which means that the backend won't properly cleanup resources for any other FSW
instances that have other `watcher_handle` values.

Addresses https://github.com/mono/mono/issues/16709

### DIFF
--- a/mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
+++ b/mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
@@ -135,20 +135,19 @@ namespace MonoTests.System.IO
 
 				var fsw1 = new FileSystemWatcher (tmp.Path, "*");
 				var fsw2 = new FileSystemWatcher (tmp.Path, "*");
+				// at this point watcher and watcher_handle should be set
 
-				// Can't check for internal_map earlier - it is lazily created when the first FSW instance is created
-				var internal_map = proxyTypeInternalMapField.GetValue (null)
-					as global::System.Collections.Generic.IDictionary<object, global::System.IO.CoreFX.FileSystemWatcher>;
-				Assert.IsNotNull (internal_map);
-
+				global::System.Collections.Generic.IDictionary<object, global::System.IO.CoreFX.FileSystemWatcher> internal_map = null;
 				object handle1 = null;
 				object handle2 = null;
 
 				// using "using" to ensure that Dispose gets called even if we throw an exception
 				using (var fsw11 = fsw1)
 				using (var fsw22 = fsw2) {
-					// at this point watcher and watcher_handle should be set
 
+					// Once at least one FSW is initialized, watcher should be set.  But if the
+					// wrong backend is getting used, ignore this test because the other checks
+					// (internal_map in particular) won't be valid.
 					var watcher = watcherField.GetValue (fsw1);
 					Assert.IsNotNull (watcher);
 					if (!proxyType.IsAssignableFrom (watcher.GetType ()))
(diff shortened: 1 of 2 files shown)


Therzok:
@monojenkins backport 2019-06

lambdageek:
I'm not sure how to write a good test case for this (how do you write a test to verify that Dispose was called?).

The [sample test code](https://github.com/mrward/test-file-watcher-dispose/blob/e832e854712e813795d7c198948dc2588bc6d630/Program.cs) relies on manually observing the result of `lsof -p <pid>` which is not convenient to do in a regression test.

Therzok:
@lambdageek FileSystemWatcher is not sealed, so you can subclass and override.

steveisok:
@monojenkins backport 2019-08

lambdageek:
Added a test case that's using reflection to peek at the `internal_map` inside `CoreFXFileSystemWatcherProxy` to check that after `Dispose` it's not holding any more entries keyed on the FSW's `watch_handle` values.

steveisok:
@monojenkins backport 2019-08

mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
```diff
@@ -109,7 +110,72 @@ public void LargeNumberOfInstances ()
...
+				}
+
+				// Dispose was called, now watcher_handle should be null
+
+				Assert.IsNull (watcherHandleField.GetValue (fsw1));
+				Assert.IsNull (watcherHandleField.GetValue (fsw2));
+
+				// and moreover, the CoreFXFileSystemWatcherProxy shouldn't have entries for either handle1 or handle2
+
+				var proxyTypeInternalMapField = proxyType.GetField ("internal_map", BindingFlags.Static | BindingFlags.NonPublic);
```
Therzok: can we check the values are there _before_ disposing?

steveisok:
@monojenkins build failed

mcs/class/System/Test/System.IO/FileSystemWatcherTest.cs
```diff
@@ -109,7 +110,72 @@ public void LargeNumberOfInstances ()
...
+				Assert.IsNotNull (proxyTypeInternalMapField);
+				var internal_map = proxyTypeInternalMapField.GetValue (null)
+					as global::System.Collections.Generic.IDictionary<object, global::System.IO.CoreFX.FileSystemWatcher>;
```
alexischr: I would think this would not work on BCL profiles where the FSW namespace shift doesn't happen. You probably need this: 

https://github.com/mono/corefx/blob/master/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.cs#L15

lambdageek: @alexischr This entire file is inside `#if !MOBILE`.  

Also I found that conditional confusing - which of our configuration define `MONO`, `MOBILE` and `MOBILE_DESKTOP_HOST`?  Is that Android Designer?

```csharp
#if MONO && (!MOBILE || MOBILE_DESKTOP_HOST)
```

steveisok:
@lambdageek Looks like windows doesn't your test :-(

lambdageek:
```
Assert.IsNotNull (internal_map);
```
this line is failing on Windows.  I guess because we don't use the `CoreFXFileSystemWatcherProxy`  as the backend on Windows. so the `internal_map` is never initialized

steveisok:
@monojenkins backport 2019-06

steveisok:
@monojenkins backport 2019-06

steveisok:
3rd time is the charm

@monojenkins backport 2019-06

PR #16925, opened 2019-09-18
[netcore] Add dllmap

### LINKED ISSUE #16521: [netcore] Implement DllMap
Run the following snippet on macOS:
```csharp
using System;

class Program
{
    [DllImport("msvcrt")]
    public static extern int rand();

    {
        Console.WriteLine("Rand: " + rand());
    }
}
```

legacy mono will redirect "msvcrt" to "libc" (see https://github.com/mono/mono/blob/master/data/config.in#L18) and return some random value,

Mono-netcore:
```
Unhandled Exception:
System.DllNotFoundException: msvcrt assembly:<unknown assembly> type:<unknown type> member:(null)
   at HelloWorld.Program.Main(String[] args)
[ERROR] FATAL UNHANDLED EXCEPTION: System.DllNotFoundException: msvcrt assembly:<unknown assembly> type:<unknown type> member:(null)
   at HelloWorld.Program.Main(String[] args)
```

CoffeeFlux:
Fixes #16521

I'm not convinced this is particularly robust or the best way to do things, but I'm not an autotools pro. If there's a better way to pass down variables here than copying them into config.make, please let me know.

I've opted to start as conservatively as possible, with just libc, in an effort to avoid compatibility shims with .NET 5. Additional entries can be added in the same fashion as this one should need arise.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6986,6 +6986,7 @@ if test "x$with_core" = "xonly"; then
   echo "COREARCH = $COREARCH" >> netcore/config.make
   echo "CORETARGETS = $CORETARGETS" >> netcore/config.make
   echo "MONO_CORLIB_VERSION = $MONO_CORLIB_VERSION" >> netcore/config.make
+  echo "LIBC = $LIBC" >> netcore/config.make
 
   if test x$build_darwin = xyes; then
     echo "HOST_PLATFORM ?= macos" >> netcore/config.make
--- a/netcore/Makefile
+++ b/netcore/Makefile
@@ -92,12 +92,14 @@ patch-app: prepare
 	cd $(APP_PATH) && $(DOTNET) publish -c Release -r $(RID) -f netcoreapp3.0
 	cp ../mono/mini/.libs/libmonosgen-2.0$(PLATFORM_AOT_SUFFIX) $(APP_PATH)/bin/Release/netcoreapp3.0/$(RID)/publish/$(PLATFORM_AOT_PREFIX)coreclr$(PLATFORM_AOT_SUFFIX)	
 	cp System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.dll $(APP_PATH)/bin/Release/netcoreapp3.0/$(RID)/publish
+	cp System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.dll.config $(APP_PATH)/bin/Release/netcoreapp3.0/$(RID)/publish
 	cp System.Private.CoreLib/bin/$(COREARCH)/System.Private.CoreLib.pdb $(APP_PATH)/bin/Release/netcoreapp3.0/$(RID)/publish
 
 # makes $(DOTNET) to use mono runtime (to run real-world apps using '$(DOTNET) run')
 patch-local-dotnet: prepare
 	cp ../mono/mini/.libs/libmonosgen-2.0$(PLATFORM_AOT_SUFFIX) ../.dotnet/shared/Microsoft.NETCore.App/$(BOOTSTRAP_RUNTIME)/$(PLATFORM_AOT_PREFIX)coreclr$(PLATFORM_AOT_SUFFIX)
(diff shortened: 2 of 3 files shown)


CoffeeFlux:
NativeLibrary is on my todo list for this sprint, so that should be coming soon at least.

filipnavara:
For the record, CoreCLR has two versions of the name resolution algorithm - [a managed one ](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Runtime/Loader/LibraryNameVariation.Unix.cs) and [native one](https://github.com/dotnet/coreclr/blob/ab2f9caaf35c96d029b96aa171ee65d04253cf7c/src/vm/dllimport.cpp#L6375-L6436). They do handle fallback from something like "libc.so.42" to "libc.so" but not the other way around. I'd be curious to see some real-world example of .NET code that currently fails without the dllmap so I can see how it behaves on CoreCLR.

akoeplinger:
I share @filipnavara's concern on this one, we shouldn't be implementing things which deviate from the other .NET Core runtimes without a good reason.

lambdageek:
I find the CoreCLR hardcoded resolutions algorithm somewhat unsatisfactory - a data-driven approach seems more predictable than a hardcoded algorithm.

That said, I agree with @filipnavara - https://github.com/mono/mono/issues/16521#issuecomment-533172829 - let's see a real-world example where netcore mono needs a dllmap for libc before we the dllmap.  Maybe we just need to finish the rest of the netcore native library loading support first.

CoffeeFlux:
Per https://github.com/mono/mono/issues/16521#issuecomment-533176960, we *are* going to need some sort of manual transition for libc. I'll adjust this to not use an external file for now.

filipnavara:
@CoffeeFlux I checked some code on GitHub and it explicitly specified `libc.so.6` in the library name. CoreCLR can fallback from that to `libc.so` or even `libc.dylib`.

jaykrell:
Linux upgraded from libc5, which I suspect wasn't even GLIBC, to libc6, which is glibc (ignoring muslc, etc.), circa 1996, and it'll presumably never change again. It was painful then. It'd be much worse now.

CoffeeFlux:
@filipnavara i think you have the issue backwards - the concern here isn't getting libc from someone specifying libc.so.6, but the reverse. The comment I linked indicates CoreCLR has a manual translation to deal with that singular case. We will need something equivalent, because some code just specifies 'libc'. This entire issue came up because someone in Slack was trying to run benchmarking code doing exactly that and it was failing :(

filipnavara:
@CoffeeFlux I am suggesting that the original issue has it backwards. While searching for code on GitHub the usages on .NET Core friendly libraries use the numbered form (eg. https://github.com/tmds/Tmds.LibC/blob/43eb8b525a49a4e3a179a2f7fa533f18c69273ee/src/Sources/linux.specific/LibraryNames.glibc.cs#L5). I understand that Mono historically supported specifying "libc" but it looks to me like the .NET Core standard way is to specify "libc.so.6" and let it fallback to something that is actually available.

filipnavara:
Ah, I missed the last comment on the issue. I see, we should implement that.

CoffeeFlux:
I see. That seems crazy though - why would the *default* be to specify "libc.so.6" instead of "libc" for a cross-platform application... I guess we have to be able to map both ways? This is getting messier by the minute :(

vargaz:
These dllmaps were added so users can fix these problems themselves because the logic is not in C code. So having that functionality in the netcore version of mono is useful imho.

CoffeeFlux:
@marek-safar I'm going to need to somehow implement the libc transition and this seems like a reasonable approach instead of hard-coding something into the runtime. Can we merge this if I update the PR to ensure it's included in the NuGet package?

marek-safar:
@CoffeeFlux I think for the libc case we should go with the platform-specific handling compiled in. I am not sure the complexity of handling this required mapping through e.g. CLI packaging. That does not mean we have to rip out the dllmap I'd just not use it for this common library

PR #16927, opened 2019-09-19
Partially inline mono_method_signature_checked/internal.

jaykrell:
Or consider LTO/LTCG on desktop.

### DIFF
--- a/mono/profiler/coverage.c
+++ b/mono/profiler/coverage.c
@@ -278,7 +278,7 @@ dump_method (gpointer key, gpointer value, gpointer userdata)
 	image = mono_class_get_image (klass);
 	image_name = mono_image_get_name (image);
 
-	method_signature = mono_signature_get_desc (mono_method_signature_profiler (method), TRUE);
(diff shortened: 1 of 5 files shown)


jaykrell:
@monojenkins build failed

mono/profiler/coverage.c
```diff
@@ -278,7 +278,7 @@ dump_method (gpointer key, gpointer value, gpointer userdata)
 	image = mono_class_get_image (klass);
 	image_name = mono_image_get_name (image);
 
-	method_signature = mono_signature_get_desc (mono_method_signature_internal (method), TRUE);
+	method_signature = mono_signature_get_desc (mono_method_signature_profiler (method), TRUE);
```
lambdageek: Why can't the profiler use the `static inline` definitions?

jaykrell: Initially I had separate out of line profiler functions and profiler couldn’t use unexporteds. Now, agreed, should work, will apply that. Thank you/sorry.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

PR #16928, opened 2019-09-19
[draft] Reduce BCL/test sensitivity to tailcall optimizations,

jaykrell:
only where it is fairly obvious what the problem is.

There are other problems but they aren't understood.

In particular code likes to walk stack to precisely verify
it is "as expected" or to report data based on its caller.

One unusual API fails to take a parameter but instead the parameter
is the calling class, called from .cctor.

The calls to "NoTailcall" strewn around, are not a special symbol that a runtime/compiler should special case. Rather, they remove the ability for the preceding call to actually be a tailcall, because it is no longer followed by a ret, which is the most basic requirement for "simple auto tailcall" (in the absence of fancier optimization, such as making removing the call code that does nothing, etc.)

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Makefile
+++ b/mcs/class/Mono.Debugger.Soft/Makefile
@@ -1,6 +1,14 @@
 thisdir = class/Mono.Debugger.Soft
 include ../../build/rules.make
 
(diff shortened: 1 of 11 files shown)


akoeplinger:

PR #16929, opened 2019-09-19
Remove switch_dir_separators. Leave strings alone. Compat?

jaykrell:
Note that the code does not even quite do what the comment says -- it skips the last character always.

This function seems questionable, but maybe it helps Windows code run on Unix?

### DIFF
--- a/mono/metadata/w32process-unix.c
+++ b/mono/metadata/w32process-unix.c
@@ -1175,19 +1175,6 @@ mono_w32process_signal_finished (void)
 }
 
 #if defined (HAVE_FORK) && defined (HAVE_EXECVE)
-static void
-switch_dir_separators (char *path)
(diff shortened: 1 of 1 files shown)


jaykrell:
@monojenkins build failed

lateralusX:
I guess there is a rather big risk introducing regression with this change, since this function get used by managed Process class. Even if CI doesn't rely on switching dir separators, there is certainly code out there that do, that would be broken by this change. I suggest that we just fix the identified issue not replacing the dir separator, if it is the last one in the string, but keep the function around.

jaykrell:
On the other hand, how do you pass through backward slash?
Why does it alone not make it through unchanged?
What does CoreCLR do?
I wonder if there is a spec for this.

PR #16933, opened 2019-09-19
[2019-06] [offsets-tool] Install clang into the user-specific python directory.

rolfbjarne:
Because the system location isn't world writeable on Catalina.


lewurm:
@monojenkins backport 2019-08
@monojenkins backport master

PR #16942, opened 2019-09-19
[interp] omit safepoint insertion for managed-to-native wrappers

lewurm:
Like the JIT does too:
https://github.com/mono/mono/blob/1256d6c8cf149306032677fa8e060a8de9e69de2/mono/mini/mini.c#L2879-L2890

Contributes to https://github.com/mono/mono/issues/16819


BrzVlad:
I don't understand how this is related to what the JIT does. The JIT was skipping safepoints in the wrapper because a safepoint can call another wrapper which would also have safepoints. In the interpreter the safepoint is not calling any other wrappers. Also this commit disables safepoints for all m2n wrappers ?

lewurm:
Yes, it's wrong.

PR #16946, opened 2019-09-19
Fix check for SafeHandle types

filipnavara:


filipnavara:

PR #16947, opened 2019-09-19
Made Write and ReadCore functional within DeflateStream.  

steveisok:
For now, it'll route to standard read and write methods.


steveisok:
Closing... Using https://github.com/mono/mono/pull/16951 instead.

PR #16949, opened 2019-09-19
[netcore] Propagate ALCs through reflection functions

CoffeeFlux:
Draft because I still have ambient calls and I'm not convinced all the sites where I get the ALC from the image are correct. I'm also not sure how this should work with the marshaling code.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -1772,7 +1772,6 @@ type_from_parsed_name (MonoTypeNameParse *info, MonoStackCrawlMark *stack_mark, 
 	g_assert (rootimage);
 
 	*caller_assembly = assembly;
-	alc = rootimage->alc;
 
 	if (info->assembly.name) {
 		MonoAssemblyByNameRequest req;
@@ -1782,9 +1781,11 @@ type_from_parsed_name (MonoTypeNameParse *info, MonoStackCrawlMark *stack_mark, 
 		assembly = mono_assembly_request_byname (&info->assembly, &req, NULL);
 	}
 
-	/* When loading from the current assembly, AppDomain.TypeResolve will not be called yet */
-	type = mono_reflection_get_type_checked (rootimage, assembly->image, info, ignoreCase, TRUE, &type_resolve, error);
-	goto_if_nok (error, fail);
+	if (assembly) {
+		/* When loading from the current assembly, AppDomain.TypeResolve will not be called yet */
+		type = mono_reflection_get_type_checked (alc, rootimage, assembly->image, info, ignoreCase, TRUE, &type_resolve, error);
+		goto_if_nok (error, fail);
+	}
 
 	// XXXX - aleksey -
 	//  Say we're looking for System.Generic.Dict<int, Local>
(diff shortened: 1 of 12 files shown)


CoffeeFlux:
The image->alc issue is easily fixed. The biggest remaining question is whether or not `mono_reflection_type_from_name_checked` should take an ALC argument or just assume the one from the image. Currently, I think every caller is going to just want to get the ALC from the image, but I'm not sure if there's a scenario where that is not the case. Will have to think about it.

CoffeeFlux:
I think I'm going to assume that no, `mono_reflection_type_from_name_checked` should not take the ALC as a separate argument for now. It's easier to go back and add that afterwards if needed than to remove it, so start conservatively and adjust later on.

CoffeeFlux:
We'll go with this for now. Adds a lot of new ambient calls, but it's a good start on threading the ALC through. This is another potential candidate for rebasing, and if we want to go that route I can amend the commit names.

mono/metadata/icall.c
```diff
@@ -1719,23 +1720,28 @@ type_from_parsed_name (MonoTypeNameParse *info, MonoStackCrawlMark *stack_mark,
...
+	g_assert (rootimage);
+
+	*caller_assembly = assembly;
+
+	if (info->assembly.name) {
+		MonoAssemblyByNameRequest req;
+		mono_assembly_request_prepare_byname (&req, MONO_ASMCTX_DEFAULT, alc);
+		req.requesting_assembly = assembly;
+		req.basedir = assembly ? assembly->basedir : NULL;
+		assembly = mono_assembly_request_byname (&info->assembly, &req, NULL);
```
lambdageek: This can fail, right?  So then when you call `mono_reflection_get_type_checked (..., assembly->image, ....)` I think you could be dereferencing a null pointer.

I think you need to keep the `if (assembly) { ... }` stuff from before.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:

CoffeeFlux:
@monojenkins build Windows x64 Interpreter

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
@monojenkins build failed

lambdageek:

CoffeeFlux:
wasm failure is relevant; will fix later today. Pretty sure the rest is noise.

CoffeeFlux:
@monojenkins build failed

CoffeeFlux:
wasm is still failing, but with Zoltan's latest PR it should offer a more useful stack trace.

CoffeeFlux:
@monojenkins build failed

PR #16953, opened 2019-09-19
Fix sign error in GetAllocatedBytesForCurrentThread

naricc:
Fix to use an unsigned type (guint64) for tracking the nubmer of bytes allocated by the current thread.

### DIFF
--- a/mono/metadata/icall.c
+++ b/mono/metadata/icall.c
@@ -9320,7 +9320,7 @@ ves_icall_System_GC_GetMaxGeneration (void)
 	return mono_gc_max_generation ();
 }
 
-gint64
+guint64
 ves_icall_System_GC_GetAllocatedBytesForCurrentThread (void)
 {
 	return mono_gc_get_allocated_bytes_for_current_thread ();
--- a/mono/sgen/sgen-gc.h
+++ b/mono/sgen/sgen-gc.h
@@ -416,7 +416,7 @@ struct _SgenThreadInfo {
 	char *tlab_real_end;
 
 	/* Total bytes allocated by this thread in its lifetime so far. */
-	gint64 total_bytes_allocated;
+	guint64 total_bytes_allocated;
 };
 
 gboolean sgen_is_worker_thread (MonoNativeThreadId thread);


jaykrell:
So. I think this makes sense. I was trained to prefer unsigned numbers, in general because most quantities in life and software cannot be negative. File sizes, array indices, but not pointer indices.

But a lot of people, projects, code, prefer signed. Including mono.

So I wonder what is the rationale?

mono/metadata/icall.c
```diff
@@ -9320,7 +9320,7 @@ ves_icall_System_GC_GetMaxGeneration (void)
 	return mono_gc_max_generation ();
 }
 
-gint64
+guint64
```
lambdageek: it's declared as `long` on the managed side
https://github.com/mono/mono/blob/8fed6666b080d3d5634fd9068cd42f273f819eeb/mcs/class/referencesource/mscorlib/system/gc.cs#L108
and we can't change the signature https://docs.microsoft.com/en-us/dotnet/api/system.gc.getallocatedbytesforcurrentthread?view=netframework-4.8#System_GC_GetAllocatedBytesForCurrentThread

jaykrell:
Signed numbers have been known to be security problems.
Because people check for  < N and not >= 0.
And also signed numbers lead to CLR playing tricks for perf, when unsigned would have been faster. Alas.

mono/metadata/icall.c
naricc: Does changing it to gint64 actually fix anything? You still end up with a weird negative value sometimes, its just a matter of when.

jaykrell: If 63 bits isn't enough, then 64 probably isn't either?, is kinda my saying, usually with 31 and 32, widen to 63 or 64..

jaykrell: Managed tends to prefer signed, because I guess, to avoid too many confusing integer types, and I not sure I sympathetic to that or not.
(similar to array/string lengths not being 64 bits, alas).

jaykrell: And then if managed is "wrong", should native go down with the ship, or correct where possible?
And is improvement possible? I mean, the main consumer of the value will be the managed side anyway?

When the gint64 overflows, the result is undefined.
Some implementations might trap. Anything mono runs on? Anything real?
When guint64 overflows, the result is well defined, silent modulo math.
Not sure it matters.
Will the 63 bits overflow in our lifetime?

jaykrell: I still wonder if mono's predilection for signed numbers is wrong or harmful.
Google style guide is similar though, prefers signed 32bit `int` above all other integer types.

naricc: 63 bits is already overflowing in a test suite, which is where this PR comes from. I guess if other tests run and do lots of allocation before hand.

jaykrell: How long before 64 bits overflows?
Is something larger needed?

BrzVlad: @naricc are you sure the number is overflowing and there isn't any bug in the implementation. What suite is overflowing ?

naricc: Well, any way, I don't think this PR actually can fix the issue. So I am abandoning it.

Therzok:
> So I wonder what is the rationale?

I _guess_ it's CLS compliance based? Some managed languages cannot express unsigned values, so interop with the runtime can be broken with unsigned values.

PR #16958, opened 2019-09-20
[debugger] Crash when debugging iOS application that throws

### LINKED ISSUE #16824: Crash when debugging iOS application that throws on DispatchQueue
_From @chamons on Wed, 11 Sep 2019 17:41:55 GMT_

### Steps to Reproduce

1.  Create New iOS App
2. Replace AppDelegate FinishedLaunching with:
```
		static void BackgroundMethod()
		{
			throw new Exception("Exception in background queue");
		}

		public override bool FinishedLaunching (UIApplication application, NSDictionary launchOptions)
		{
			// Override point for customization after application launch.
			// If not required for your application you can safely delete this method

			backgroundQueue = new DispatchQueue ("nameOfDispatchQueue", false);
			backgroundQueue.DispatchAsync (BackgroundMethod);

			return true;
		}
```
3. Debug (Simulator is fine)

### Expected Behavior

A popup like we get on macOS

image

### Actual Behavior

Crash: https://gist.github.com/chamons/b80a5140227cd4fd4358f39a3d0f5985

### Environment

Xcode11-Preview
https://gist.github.com/chamons/bb6df5a45bc81f95755350fda2e34a17

_Copied from original issue xamarin/xamarin-macios#6982_

thaystg:
Doing the same fix from Android to IOS, I removed the ifdef, I don't think that this will happen when not using IOS nor Android, but I prefer to let the if there to avoid the exception.

On mini-exceptions.c check if there is no ji we can continue and generate the exception without checking wrapper_type.

Fixes #16824

### DIFF
--- a/mono/mini/debugger-engine.c
+++ b/mono/mini/debugger-engine.c
@@ -1277,11 +1277,9 @@ mono_de_ss_start (SingleStepReq *ss_req, SingleStepArgs *ss_args)
 	int nframes = ss_args->nframes;
 	SeqPoint *sp = &ss_args->sp;
 
-#if defined(HOST_ANDROID) || defined(TARGET_ANDROID)
-	/* this can happen on a single step in a exception on android (Mono_UnhandledException_internal) */
+	/* this can happen on a single step in a exception on android (Mono_UnhandledException_internal) and on IOS */
 	if (!method)
 		return;
-#endif		
 
 	/*
 	 * Implement single stepping using breakpoints if possible.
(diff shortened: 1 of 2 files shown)


thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

marek-safar:
@monojenkins backport 2019-08

Issue #16918, opened 2019-09-18
XmlSerializer Deserializing Property with same name as its enum type no longer works


devobra:
## Steps to Reproduce

Android
1. Download attached project and launch debug for BugReproduction-Deserialize-Android
2. Press the + floating action button to serialize/deserialize the test class that has 5 enum properties with the same name as the enum.
3. A toast will inform you which of the 5 properties deserialized properly and which didn't.

iOS
Same as Android, just debug BugReproduction-Deserialize-iOS and press the Button

[BugReproduction-Deserialize.zip](https://github.com/mono/mono/files/3627930/BugReproduction-Deserialize.zip)

### Current Behavior

Deserializing an Object Property which has the same name as the type of enum it contains has unpredictable results. It can succeed or fail depending on the specific number of fields in the Class and whether those fields have the same name as their type.

In addition, if the object being created has a Property that accepts objects, the XmlSerializer may set one of the XmlNodes it failed to read as the value of such a Property.

### Expected Behavior

Deserializing an Object Property which has the same name as the type of enum it contains works like any other property.

## On which platforms did you notice this

[X] Xamarin iOS
[X] Xamarin Android
[X] Xamarin Mac
[  ] .Net Console Project on Visual Studio for Mac

**Version Used**:

Mono Framework MDK
Runtime:
	Mono 6.0.0.334 (2019-02/6256b82d62f) (64-bit)
	Package version: 600000334

This specific problem with deserialization appears to be new. I do not know when it broke exactly but I believe it was working with Mono 5.14 Commit 000780ca. It does not happen on a .Net console project that uses Mono, so it may be specific to the Mono version that is bundled with Xamarin.

## Stacktrace

no stack trace.

The XmlSerializer fires the UnknownElement event for each Property that fails to deserialize.

Edit : updated to add that the bug also occurs on Xamarin Mac but not in a Dot Net console application that uses Mono.

marek-safar:
@steveisok, please check if it's a regression

devobra:
Just a hunch, but given that :

-the Serializer code doesn't seem to have changed in a long time
-the bug can cause objects to end up in the wrong field (an XmlNode ends up in one of the fields of the object being created)

The problem could be a bug with Reflection rather than with the Serializer itself.

steveisok:
We think the bug was at least surfaced w/ this commit and are trying to figure out if it applies the mobile profile only.

[...]

Issue #16473, opened 2019-08-26
XmlSerializer Leaving Nodes Serialized


rolfbjarne:
_From @AceCoderLaura on Mon, 26 Aug 2019 02:45:31 GMT_

### Steps to Reproduce

1. Use the below two files in a Xamarin.iOS project
#### AppDelegate.cs (Compile)
```
using Foundation;
using System;
using System.Collections.Generic;
using System.Diagnostics;
using System.IO;
using System.Runtime.Serialization;
using System.Xml;
using System.Xml.Serialization;
using UIKit;
using Xamarin.Forms.Platform.iOS;

namespace Adapt.Presentation.iOS
{
    [Register("AppDelegate")]
    public class AppDelegate : FormsApplicationDelegate
    {
        public override bool FinishedLaunching(UIApplication app, NSDictionary options)
        {
            try
            {
                var stream = System.Reflection.Assembly.GetExecutingAssembly().GetManifestResourceStream("Adapt.Presentation.Xivic.iOS.Some.xml");
                var reader = new StreamReader(stream);
                var doc = reader.ReadToEnd();
                var deserialized = DeserializeMyObject(doc);
                if (deserialized.Contents[0] is XmlNode[]) throw new Exception("Failed to deserialize.");
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex);
            }

            Debugger.Break();
            return true;
        }

        private static WhereList DeserializeMyObject(string markup)
        {
            var serializer = new XmlSerializer(typeof(WhereList));
            return (WhereList)serializer.Deserialize(new StringReader(markup));
        }
    }

    public class WhereList
    {
        public List<object> Contents { get; set; }
    }

    /// <summary>
    /// Represents a query Where Clause
    /// </summary>
    [XmlInclude(typeof(WhereClauseType))]
    [XmlInclude(typeof(List<string>))]
    [XmlInclude(typeof(int))]
    [XmlInclude(typeof(string))]
    [XmlInclude(typeof(decimal))]
    [XmlInclude(typeof(double))]
    [XmlInclude(typeof(bool))]
    public class WhereClause
    {
        public object LeftExpression { get; set; }

        public WhereClauseType WhereClauseType { get; set; } = WhereClauseType.Equals;

        public object RightExpression { get; set; }
    }

    /// <summary>
    /// The SQL Comparison operator to be used for the where clause expression
    /// </summary>
    [DataContract]
    public enum WhereClauseType
    {
        Is,
        Equals,
        Like
[...]

PR #16965, opened 2019-09-20
[System.Xml] Fix deserialization issue

MaximLipnin:
[This ](https://github.com/mono/mono/pull/11194/files#diff-ae54d9565e735005686191de0f19ed50R431-R435)change in `XmlTypeMapElementInfo.GetElement`  introduced the logic for handling the case of derived types deserialization. 
For some cases (like https://github.com/mono/mono/issues/16918) it breaks existing functionality because it returns wrong element for given element name after searching withing derived types collection.

image

Unfortunately, what I can do here is just reverting the related changes.

Fixes https://github.com/mono/mono/issues/16918


MaximLipnin:
@monojenkins build failed

MaximLipnin:
Possibly https://github.com/mono/mono/issues/16473 is related

akoeplinger:
@majellin24 FYI we unfortunately had to revert your pull request https://github.com/mono/mono/pull/11194 as it caused a regression.

marek-safar:
@monojenkins backport 2019-08

PR #16971, opened 2019-09-20
[checked-build] update allowed thread states

lewurm:
@lambdageek does it still make sense?


marek-safar:
@lambdageek please review

PR #16977, opened 2019-09-21
[netcore] Implement System.Runtime.Intrinsics.Arm.ArmBase (ARM64)

EgorBo:
This PR implements basic scalar [System.Runtime.Intrinsics.Arm.ArmBase](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Runtime/Intrinsics/Arm/ArmBase.cs) intrinsics for Arm64 (AArch64) via LLVM. Other SIMD stuff including `Vector<T>` is not enabled yet (but I made some preparations).

ArmBase sample:
```csharp
// CLZ 32
static int MyTest1(int x)   => ArmBase.LeadingZeroCount(x);
static int MyTest2(uint x)  => ArmBase.LeadingZeroCount(x);

// CLS 64
static int MyTest2(long x)  => ArmBase.Arm64.LeadingZeroCount(x);
static int MyTest3(ulong x) => ArmBase.Arm64.LeadingZeroCount(x);

// CLS 32/64
static int MyTest4(int x)   => ArmBase.Arm64.LeadingSignCount(x);
static int MyTest5(long x)  => ArmBase.Arm64.LeadingSignCount(x);
```
Codegen for `--aot=llvm,mcpu=native` on arm64.
```asm
0000000000000720 <HelloWorld_Program_MyTest1_int>:
 720:   5ac01000        clz     w0, w0
 724:   d65f03c0        ret
 728:   d503201f        nop
 72c:   d503201f        nop

0000000000000730 <HelloWorld_Program_MyTest2_uint>:
 730:   5ac01000        clz     w0, w0
 734:   d65f03c0        ret
 738:   d503201f        nop
 73c:   d503201f        nop

0000000000000740 <HelloWorld_Program_MyTest2_long>:
 740:   dac01000        clz     x0, x0
 744:   d65f03c0        ret
 748:   d503201f        nop
 74c:   d503201f        nop

0000000000000750 <HelloWorld_Program_MyTest3_ulong>:
 750:   dac01000        clz     x0, x0
 754:   d65f03c0        ret
 758:   d503201f        nop
 75c:   d503201f        nop

0000000000000760 <HelloWorld_Program_MyTest4_int>:
 760:   5ac01400        cls     w0, w0
 764:   d65f03c0        ret
 768:   d503201f        nop
 76c:   d503201f        nop

0000000000000770 <HelloWorld_Program_MyTest5_long>:
 770:   dac01400        cls     x0, x0
 774:   d65f03c0        ret
 778:   d503201f        nop
 77c:   d503201f        nop
```
Tested on the `softiron-1` machine with Cortex-A53.
`ArmBase` should be supported on any Arm64 and in order to turn this set off: `--aot=llvm,mattr=-base`. These intrinsics will be used in CoreLib like this: https://github.com/dotnet/coreclr/pull/26815 (improves performance in various places e.g. number formatting)


mono/metadata/w32process-unix.c
```diff
@@ -4279,4 +4279,28 @@ mono_w32process_signal_finished (void)
...
+{
+	return TRUE;
+}
+
+gboolean
+mono_w32process_ver_query_value (gconstpointer datablock, const gunichar2 *subblock, gpointer *buffer, guint32 *len)
+{
+	return TRUE;
+}
+
```
EgorBo: it's not relevant but the debug builds complain (crash) on missing symbols.

mono/mini/mini-llvm.c
```diff
@@ -7553,7 +7545,44 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
...
+
+#if defined(ENABLE_NETCORE) && defined(TARGET_ARM64)
+		case OP_CLS32:
+		case OP_CLS64: {
+			// %shr = ashr i32 %x, 31
+			// %xor = xor i32 %shr, %x
+			// %mul = shl i32 %xor, 1
+			// %add = or i32 %mul, 1
+			// %0 = tail call i32 @llvm.ctlz.i32(i32 %add, i1 false)
+			LLVMValueRef shr = LLVMBuildAShr (builder, lhs, ins->opcode == OP_CLS32 ?
```
EgorBo: Copied from https://github.com/microsoft/checkedc-llvm/blob/d8d78c7a13472b344e429d1ae320bc4f01ecb6db/test/CodeGen/AArch64/arm64-clrsb.ll
(LLVM recognizes it as `cls`)

mono/mini/local-propagation.c
```diff
@@ -841,7 +841,7 @@ static gboolean
 mono_is_simd_accessor (MonoInst *ins)
 {
 	switch (ins->opcode) {
-#ifdef MONO_ARCH_SIMD_INTRINSICS
+#if defined(MONO_ARCH_SIMD_INTRINSICS) && !defined(TARGET_ARM64)
```
vargaz: Why is this needed ?

mono/mini/mini-llvm.c
```diff
@@ -10734,9 +10763,18 @@ MonoCPUFeatures mono_llvm_get_cpu_features (void)
...
+		{ "crypto",	MONO_CPU_ARM64_CRYPTO },
+		{ "crc",	MONO_CPU_ARM64_CRC },
 #endif
 	};
 	if (!cpu_features)
 		cpu_features = MONO_CPU_INITED | (MonoCPUFeatures)mono_llvm_check_cpu_features (flags_map, G_N_ELEMENTS (flags_map));
+#if defined(TARGET_ARM64)
+	cpu_features |= MONO_CPU_ARM64_BASE;	// arm doesn't have an alias for it but we want users to be able to turn ArmBase off if they want
+											// it's always supported on Arm64 by default
+#endif
```
vargaz: Indentation.

mono/mini/mini.h
```diff
@@ -314,7 +314,11 @@ enum {
 #define MONO_IS_REAL_MOVE(ins) (((ins)->opcode == OP_MOVE) || ((ins)->opcode == OP_FMOVE) || ((ins)->opcode == OP_XMOVE) || ((ins)->opcode == OP_RMOVE))
 #define MONO_IS_ZERO(ins) (((ins)->opcode == OP_VZERO) || ((ins)->opcode == OP_XZERO))
 
+#ifndef TARGET_ARM64
 #define MONO_CLASS_IS_SIMD(cfg, klass) (((cfg)->opt & MONO_OPT_SIMD) && m_class_is_simd_type (klass))
```
vargaz: Why is this needed ?

mono/mini/mini.h
```diff
@@ -2845,7 +2849,13 @@ typedef enum {
 	MONO_CPU_X86_AVX2_COMBINED        = MONO_CPU_X86_AVX_COMBINED   | MONO_CPU_X86_POPCNT | MONO_CPU_X86_AVX2,
 	MONO_CPU_X86_FMA_COMBINED         = MONO_CPU_X86_AVX_COMBINED   | MONO_CPU_X86_POPCNT | MONO_CPU_X86_FMA,
 	MONO_CPU_X86_FULL_SSEAVX_COMBINED = MONO_CPU_X86_FMA_COMBINED   | MONO_CPU_X86_AVX2,
+#elif defined(TARGET_ARM64)
+	MONO_CPU_ARM64_BASE		= 1 << 1,
+	MONO_CPU_ARM64_NEON		= 1 << 2,
+	MONO_CPU_ARM64_CRYPTO	= 1 << 3,
+	MONO_CPU_ARM64_CRC		= 1 << 4,
```
vargaz: Indentation.

mono/mini/mini-ops.h
```diff
@@ -1381,6 +1383,9 @@ MINI_OP(OP_ARM64_CBZX, "arm64_cbzx", NONE, IREG, NONE)
 /* Branch if sreg1 != 0 */
 MINI_OP(OP_ARM64_CBNZW, "arm64_cbnzw", NONE, IREG, NONE)
 MINI_OP(OP_ARM64_CBNZX, "arm64_cbnzx", NONE, IREG, NONE)
+
+MINI_OP(OP_CLS32, "cls32", IREG, IREG, NONE)
```
vargaz: Shouldn't this be named something else ? And moved outside of #ifdef ARM64 ?

EgorBo: It's Arm specific - Count leading sign bits http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0802a/a64_general_alpha.html

mono/mini/local-propagation.c
EgorBo: these ops are defined for x86 only in mini-ops but yeah I'll share them with arm and remove this if-def

EgorBo:
@vargaz could you please review?
cc @imhameed @lewurm @filipnavara

mono/mini/mini.h
```diff
@@ -315,7 +315,7 @@ enum {
 #define MONO_IS_ZERO(ins) (((ins)->opcode == OP_VZERO) || ((ins)->opcode == OP_XZERO))
 
 #ifdef TARGET_ARM64
-#define MONO_CLASS_IS_SIMD(cfg, klass) (((cfg)->opt & MONO_OPT_SIMD) && COMPILE_LLVM (cfg) && m_class_is_simd_type (klass))
+#define MONO_CLASS_IS_SIMD(cfg, klass) (0)
 #else
```
vargaz: Why is this needed ?

vargaz: I think we should get rid of MONO_ARCH_SIMD_SUPPORTED, its not very clear what it means any more.

mono/mini/mini-arm64.h
```diff
@@ -19,6 +19,13 @@
 
 #define MONO_MAX_IREGS 32
 #define MONO_MAX_FREGS 32
+#define MONO_MAX_XREGS 0
```
filipnavara:
```
#define MONO_MAX_XREGS 32
```

also,

```
#define MONO_ARCH_NEED_SIMD_BANK 1
#define MONO_ARCH_USE_SHARED_FP_SIMD_BANK 1
```

mono/mini/mini-arm64.h
```diff
@@ -19,6 +19,13 @@
...
+#define MONO_ARCH_CALLEE_SAVED_XREGS 0
+#define MONO_ARCH_CALLEE_XREGS 0
+
+#if !defined(DISABLE_SIMD) && defined(ENABLE_NETCORE) && defined(ENABLE_LLVM)
+#define MONO_ARCH_SIMD_INTRINSICS 1
```
filipnavara: This should not be guarded by ENABLE_LLVM. Instead we should guard by `COMPILE_LLVM (cfg)` for `supported = ...` in `emit_sys_numerics_vector` (with special case for x64).

mono/mini/mini.h
filipnavara: We should keep `MONO_CLASS_IS_SIMD` as is.

Getting rid of `MONO_ARCH_SIMD_SUPPORTED` would be nice. It doesn't save that much code and it would be nice to unconditionally enable it everywhere and guard few places with runtime check (eg. `mono_arch_is_simd_enabled (cfg)`).

vargaz: Agree.

mono/mini/mini-arm64.h
```diff
@@ -19,6 +19,13 @@
 
 #define MONO_MAX_IREGS 32
 #define MONO_MAX_FREGS 32
+#define MONO_MAX_XREGS 0
+#define MONO_ARCH_CALLEE_SAVED_XREGS 0
+#define MONO_ARCH_CALLEE_XREGS 0
```
filipnavara:
```
#define MONO_ARCH_CALLEE_XREGS MONO_ARCH_CALLEE_FREGS
#define MONO_ARCH_CALLEE_SAVED_XREGS MONO_ARCH_CALLEE_SAVED_FREGS
```

may have to be reordered in the file to place where FREGs are defined. The registers are mirrored, ie. XREGs and FREGs are the same ones on the hardware.

Aside from that, the ABI states: "Additionally, only the bottom 64 bits of each value stored in v8-v15 need to be preserved [7]; it is the responsibility of the caller to preserve larger values.". I am not sure if we will need to adjust for that? Maybe `MONO_ARCH_CALLEE_SAVED_XREGS` should be 0 after all?

filipnavara: Ref: https://developer.arm.com/docs/ihi0055/d/procedure-call-standard-for-the-arm-64-bit-architecture#id47

mono/mini/mini.h
EgorBo: If you don't mind I'll keep it `(0)` as a TODO - won't enable Vector<T> as part of this PR, but at least it prepares stuff and enables ArmBase

EgorBo:

EgorBo:

Issue #16935, opened 2019-09-19
LLVM: Improve codegen for Vector<T>.EqualsAll


EgorBo:
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
static bool Test(Vector<float> a, Vector<float> b) => a == b;
```
Here is what Mono-LLVM JIT emits:
```asm
   0:   c4 e1 f9 6e c6          vmovq  %rsi,%xmm0
   5:   c4 e1 f9 6e cf          vmovq  %rdi,%xmm1
   a:   c5 f1 6c c0             vpunpcklqdq %xmm0,%xmm1,%xmm0
   e:   c4 e1 f9 6e c9          vmovq  %rcx,%xmm1
  13:   c4 e1 f9 6e d2          vmovq  %rdx,%xmm2
  18:   c5 e9 6c c9             vpunpcklqdq %xmm1,%xmm2,%xmm1
  1c:   c5 f8 c2 c1 00          vcmpeqps %xmm1,%xmm0,%xmm0
  21:   c4 e3 79 04 c8 0e       vpermilps $0xe,%xmm0,%xmm1
  27:   c5 f0 54 c0             vandps %xmm0,%xmm1,%xmm0
  2b:   c4 e3 79 04 c8 e5       vpermilps $0xe5,%xmm0,%xmm1
  31:   c5 f8 54 c1             vandps %xmm1,%xmm0,%xmm0
  35:   c4 e3 79 14 c0 00       vpextrb $0x0,%xmm0,%eax
  3b:   c3                      retq
```
The codegen looks a bit odd.
Ideally it should be something like this
```asm
; both vectors are passed via xmm registers
vcmpneqps xmm0, xmm0, xmm1 ; compare two vectors
vpmovmskb eax, xmm0 ; movemask for the result (eax will be 0 if vectors are equal)
test      ax, ax
sete      al
ret
```

LLVM IR we emit:
```llvm
; ModuleID = 'jit-module-Test'
source_filename = "jit-module-Test"

; Function Attrs: noinline uwtable
define monocc i8 @"Program:Test"(i64 %arg_a, i64, i64 %arg_b, i64) #0 {
BB0:
  %2 = alloca <4 x float>, align 16
  %3 = alloca <4 x float>, align 16
  %4 = bitcast <4 x float>* %3 to i8*
  %5 = getelementptr i8, i8* %4, i32 0
  %6 = bitcast i8* %5 to i64*
  store i64 %arg_a, i64* %6
[...]

PR #16984, opened 2019-09-22
LLVM: Improve Vector<T>== Vector<T> (OP_XEQUAL)

EgorBo:
Fixes https://github.com/mono/mono/issues/16935
```csharp
static bool Foo(Vector<float> vec1, Vector<float> vec2) // any type (long, byte, etc)
{
    return vec1 == vec2;
}
```
Was:
```asm
  ...
  1c:   c5 f8 c2 c1 00          vcmpeqps xmm0,xmm0,xmm1
  21:   c4 e3 79 04 c8 0e       vpermilps xmm1,xmm0,0xe
  27:   c5 f0 54 c0             vandps xmm0,xmm1,xmm0
  2b:   c4 e3 79 04 c8 e5       vpermilps xmm1,xmm0,0xe5
  31:   c5 f8 54 c1             vandps xmm0,xmm0,xmm1
  35:   c4 e3 79 14 c0 00       vpextrb eax,xmm0,0x0 ;; requires SSE4.1
  3b:   c3                      ret
```
Now (SSE2 friendly and faster):
```asm
  ...
  1c:   c5 f9 74 c1             vpcmpeqb xmm0,xmm0,xmm1
  20:   c5 f9 d7 c0             vpmovmskb eax,xmm0
  24:   3d ff ff 00 00          cmp    eax,0xffff
  29:   0f 94 c0                sete   al
  2c:   c3                      ret
```
So basically I do:
```csharp
static bool Compare<T>(Vector128<T> v1, Vector128<T> v2)
{
    Vector128<byte> lfs8 = v1.AsByte();
    Vector128<byte> rfs8 = v2.AsByte();
    Vector128<byte> cmp = Sse2.CompareEqual(lfs8, rfs8);
    return Sse2.MoveMask(cmp) == 0xFFFF;
}
```
OP_XEQAUL for double and float can be a bit improved more - to use MoveMask(CompareNotEqual)==0 -- it gets rid of that `cmp`.

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -7341,6 +7341,27 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 			break;
 		}
 		case OP_XEQUAL: {
+#if defined(TARGET_AMD64) || defined(TARGET_X86)
+			int nelemsi8 = 16; // TODO: calculate vector width in bytes for current CPU
+			LLVMValueRef lhs8 = LLVMBuildBitCast (ctx->builder, lhs, LLVMVectorType (LLVMInt8Type (), nelemsi8), "");
+			LLVMValueRef rhs8 = LLVMBuildBitCast (ctx->builder, rhs, LLVMVectorType (LLVMInt8Type (), nelemsi8), "");
+
+			//%5 = icmp eq <16 x i8> %3, %4
+			LLVMValueRef cmp = LLVMBuildICmp (builder, LLVMIntEQ, lhs8, rhs8, "");
+
+			//%6 = sext <16 x i1> %5 to <16 x i8>
+			//%7 = tail call i32 @llvm.x86.sse2.pmovmskb.128(<16 x i8> %6) #2
+			LLVMValueRef args [1];
+			args [0] = LLVMBuildSExt (builder, cmp, LLVMVectorType (LLVMInt8Type (), nelemsi8), "");
+			LLVMValueRef movmsk = LLVMBuildCall (builder, get_intrins (ctx, INTRINS_SSE_PMOVMSKB), args, 1, "");
+
+			//%8 = icmp eq i32 %7, 65535
+			LLVMValueRef icmpeq = LLVMBuildICmp (builder, LLVMIntEQ, movmsk, LLVMConstInt (LLVMInt32Type (), 0xFFFF, FALSE), "");
+			
+			//%9 = zext i1 %8 to i32
+			values [ins->dreg] = LLVMBuildZExt (ctx->builder, icmpeq, LLVMInt32Type (), "");
+			break;
+#else
 			LLVMTypeRef t;
 			LLVMValueRef cmp, mask [32], shuffle;
 			int nelems;
@@ -7369,6 +7390,7 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 			values [ins->dreg] = LLVMBuildExtractElement (builder, cmp, LLVMConstInt (LLVMInt32Type (), 0, FALSE), "");
 			// Maybe convert to 0/1 ?
 			break;
+#endif
 		}
 		case OP_XBINOP: {
 			switch (ins->inst_c0) {


mono/mini/mini-llvm.c
```diff
@@ -7341,6 +7341,27 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 			break;
 		}
 		case OP_XEQUAL: {
+#if defined(TARGET_AMD64) || defined(TARGET_X86)
+			int nelemsi8 = 16; // TODO: calculate vector width in bytes for current CPU
```
vargaz: LLVM is supposed to convert the code we generate to this intrinsic.

EgorBo: but it doesn't? 🙂 . I ported this code https://godbolt.org/z/-GEmBj (after opts: https://godbolt.org/z/dIzTDE)

vargaz: I.e.:
https://github.com/llvm-mirror/llvm/blob/master/test/CodeGen/X86/vector-compare-all_of.ll#L35

EgorBo: @vargaz hm.. comments state that it should generate movmskpd but it doesn't. Instead it generate something similar to what AVX-512 should generate (but my CPU doesn't have it). Let me benchmark both my and the current impl.

EgorBo: hm.. will be fixed once we move to LLVM 9 ? https://godbolt.org/z/ITU_2g

vargaz: It generated that when I wrote that code, maybe something changed in the meantime.

filipnavara: I think it was generating it on my machine too...

EgorBo: my Fix makes `==` 50% faster according to some benchmarks. Here is the IR we emit (unoptimized):
```llvm
source_filename = "jit-module-Foo"
target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

define i8 @"HelloWorld.Program:Foo (System.Numerics.Vector`1<long>,System.Numerics.Vector`1<long>)"(i64 %arg_vec1, i64, i64 %arg_vec2, i64) local_unnamed_addr #0 {
BB0:
  %.0.vec.insert = insertelement <2 x i64> undef, i64 %arg_vec1, i32 0
  %.8.vec.insert = insertelement <2 x i64> %.0.vec.insert, i64 %0, i32 1
  %.0.vec.insert4 = insertelement <2 x i64> undef, i64 %arg_vec2, i32 0
  %.8.vec.insert6 = insertelement <2 x i64> %.0.vec.insert4, i64 %1, i32 1
  %2 = icmp eq <2 x i64> %.8.vec.insert, %.8.vec.insert6
  %3 = sext <2 x i1> %2 to <2 x i8>
  %4 = shufflevector <2 x i8> %3, <2 x i8> undef, <2 x i32> <i32 1, i32 undef>
  %5 = and <2 x i8> %4, %3
  %6 = extractelement <2 x i8> %5, i32 0
  ret i8 %6
}

attributes #0 = { noinline norecurse nounwind readnone uwtable "target-cpu"="skylake" }
```

filipnavara: Can you check what `Vector<double>` used to generate?

EgorBo: @filipnavara same codegen

imhameed: The existing code uses a vector of `i8` for each of the intermediate values (instead of a vector of integers with width equal to that of the input vector's element type), which results in bad generated code:
https://godbolt.org/z/3JNyZy

So we could instead use the type of `lhs` directly for vectors of integers, use `i64` for `double`, and use `i32` for float, and get reasonable code without needing to manually emit sequences of intrinsics.

mono/mini/mini-llvm.c
```diff
@@ -7341,6 +7341,27 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
...
+			LLVMValueRef lhs8 = LLVMBuildBitCast (ctx->builder, lhs, LLVMVectorType (LLVMInt8Type (), nelemsi8), "");
+			LLVMValueRef rhs8 = LLVMBuildBitCast (ctx->builder, rhs, LLVMVectorType (LLVMInt8Type (), nelemsi8), "");
+
+			//%5 = icmp eq <16 x i8> %3, %4
+			LLVMValueRef cmp = LLVMBuildICmp (builder, LLVMIntEQ, lhs8, rhs8, "");
```
imhameed: `pcmpeqb` is not a floating point equality comparison, and will not correctly handle NaNs or negative zero.

PR #16985, opened 2019-09-22
LLVM: Make string.FastAllocateString nonnull and noalias

EgorBo:
`string.FastAllocateString` is guaranteed to return a non-null handle (or OOM) and allocates a new object. It's used a lot internally.
```llvm
%str = notail call monocc noalias nonnull i64* %0(i32 %str_length)
```

### DIFF
--- a/mono/mini/mini-llvm.c
+++ b/mono/mini/mini-llvm.c
@@ -4102,6 +4102,11 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
 	if (call->method && call->method->wrapper_type == MONO_WRAPPER_ALLOC)
 		mono_llvm_set_call_noalias_ret (lcall);
 
+	if (call->method && call->method->klass == mono_get_string_class () && !strcmp(call->method->name, "FastAllocateString")) {
+		mono_llvm_set_call_nonnull_ret (lcall);
+		mono_llvm_set_call_noalias_ret (lcall);
+	}
+
 	/*
 	 * Modify cconv and parameter attributes to pass rgctx/imt correctly.
 	 */


mono/mini/mini-llvm.c
```diff
@@ -4102,6 +4102,11 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
 	if (call->method && call->method->wrapper_type == MONO_WRAPPER_ALLOC)
 		mono_llvm_set_call_noalias_ret (lcall);
 
+	if (call->method && call->method->klass == mono_get_string_class () && !strcmp(call->method->name, "FastAllocateString")) {
```
marek-safar: Hmm not very keep on hardcoding this in runtime (please add at least comment)

Do you have any perf numbers backing this?

jaykrell: Better alternative would be attributes on the function?

EgorBo: yeah I was thinking of special attributes for LLVM engine but I suspect CoreCLR/FX folks won't be happy about that. @marek-safar it removes a lot of redundant null checks (nonnull), Not sure about nonalias but I suspect it should allow LLVM to apply some vectorization or/and use memcpy instead of memmove (on *nix).

EgorBo: `FastAllocateString` has 50 usages inside corelib (String, StringBuilder. Convert, Guid, NumberFormatting, ParseNumber, Enum, DateTimeFormat)

marek-safar: We could for CoreLib recognize (C# non-nullable reference types hint) and apply nonnull everywhere but not sure how impactful is `nonalias` here

EgorBo: cc @vargaz

jaykrell: How about "LLVM sets the precedent"?

A lot of thought and design goes into LLVM.

They have these attributes.

There is a relationship between .NET and LLVM.
 For example LLVM being used to JIT or AOT compile .NET.

.NET should have these attributes too.

PR #16987, opened 2019-09-23
[Coop] Implicit pin all byref parameters in non-handles icall.

jaykrell:
This will allow a more efficient handling of coop. i.e this should cover all input parameters and moot much-but-not-all other icall conversion.

There are a few ways we could do this:
 1 mark the arguments as volatile
 2 store the arguments into volatile locals
 3 mark the functions noinline, noopt, not aggressive inline
 4 store the argument addresses into locals, volatile or not

Each on its own should suffice, if things mean what I think they mean.
Here I tried 1 and 3. There were failures.
Now trying 4 and 3.

Note that this would resolve for example:
```
// Generic ref/out parameters are not supported by HANDLES(), so NOHANDLES().
// FIXME HANDLES should handle this trivially. Do it separately.
NOHANDLES(ICALL(ARRAY_5, "GetGenericValueImpl", ves_icall_System_Array_GetGenericValueImpl))

// Generic ref/out parameters are not supported by HANDLES(), so NOHANDLES().
// FIXME HANDLES should handle this trivially. Do it separately.
NOHANDLES(ICALL(ARRAY_11, "SetGenericValueImpl", ves_icall_System_Array_SetGenericValueImpl))
```

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6425,6 +6425,8 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 		mono_mb_patch_branch (mb, pos);
 	}
 
+	gboolean tailcall = TRUE;
+
 	if (uses_handles) {
 		MonoMethodSignature *generic_sig = NULL;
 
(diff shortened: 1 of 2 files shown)


jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Due to the failures, might want to alter the tactic here among 1-3.

jaykrell:
As well the volatile seems like overkill.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
This seems like a really good idea to me.
Coop conversion can be greatly reduced/undone/optimized as a result.
CI seems to be in a bad state for any PR so hard to guage if it is working.
/cc @vargaz @lambdageek

mono/metadata/marshal-ilgen.c
```diff
@@ -6483,7 +6485,25 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
+		for (i = 0; i < param_count; i++) {
+
+			// Take address of some parameters to pin them.
+			// Making them volatile should also work but there were failures.
+
+			if ((i == 0 && csig->hasthis) || MONO_TYPE_IS_REFERENCE (csig->params [i]) || mono_type_is_byref (csig->params [i])) {
+				if (!tailcall) {
+					tailcall = FALSE;
+					mb->method->iflags = (mb->method->iflags & ~METHOD_IMPL_ATTRIBUTE_AGGRESSIVE_INLINING)
+						| MONO_METHOD_IMPL_ATTR_NOOPTIMIZATION | MONO_METHOD_IMPL_ATTR_NOINLINING;
```
jaykrell: Mono doesn't appear to use `MONO_METHOD_IMPL_ATTR_NOOPTIMIZATION` at all.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

SamMonoRT:
@monojenkins build failed

jaykrell:
This is a proposal, and I still suspect, maybe a good one.
/cc @lambdageek

lambdageek:
I think we want to continue to have some explicit marker on the caller side to indicate that the icalls are doing something special with their arguments (namely, pinning them).

PR #16994, opened 2019-09-23
[Coop] Resolve fixme comments in generic array icalls.

jaykrell:
Depends on https://github.com/mono/mono/pull/16987 or similar.


lambdageek:
Hm. This is superceded by https://github.com/mono/mono/pull/17034

jaykrell:
Agreed, and it had problems and was further fixed up, oops darn.

Issue #16974, opened 2019-09-20
File.GetCreationTimeUtc() returns the time last modified, not time created


ryanching:
## Steps to Reproduce

1. Create a file that was modified at a time later than it was created. 
2. In a fresh console project, call "File.GetCreationTimeUtc('path/to/file');"
3. Notice that the function returns the time that the file was last modified rather than the time at which it was created. 

### Current Behavior

Function returns time file was last modified. 
### Expected Behavior

Function should return the time the file was created. 
## On which platforms did you notice this

[X ] macOS
[ ] Linux
[ ] Windows

**Version Used**:

Mono 6.4.0.198

Aguilex:
The VSMac Telemetry Monitor app reports the time of events incorrectly as a result of this bug... while this tool is only for internal use, it does impact us significantly, since we use this monitor for a significant number of VSMac issues.

marek-safar:
@steveisok please priority this

steveisok:
So, under the hood we call `lstat` on the file in order to get all the creation/modification details.  If the birth time is not set on the file, then we will fall back to the oldest of the two between the last modification and last status change dates.  

What's strange is that the birth time is reporting 0 in code, but I running `stat -f %SB <file>` shows a creation date.  Once I dig into our interop routine, I'll know a bit more.

steveisok:
It looks like this is our culprit:

https://github.com/mono/corefx/blob/65712075da8062eb2d9304104c110185001f8541/src/Native/Unix/System.Native/pal_io.c#L160-L168

While it seems to be a reasonable assumption for some linux versions, osx does track the birth time. I need to validate a change and then I'll post a PR.

akoeplinger:
@Aguilex this is fixed in Mono 6.6.0.90 from the 2019-08 branch. Do you need the fix in 2019-6 too?

Aguilex:
@akoeplinger yes please... since 2019-06 is used for VSM 8.3, it would be good to be able to use the Telemetry Monitor, and have it report time correctly, for any 8.3 investigations the team may need to do in the coming weeks/months. This assumes that the change is low-risk and will not destabilize the product. VSM 8.3 shipped to stable yesterday, so we'd likely ship this fix in a VSM service release, which typically have much shorter test cycles (and no dogfooding or public previews). 

cc @slluis

marek-safar:
Why the internal Telemetry Monitor app cannot run with 2019-08 Mono?

Aguilex:
[...]

PR #16999, opened 2019-09-23
Correctly check for HAVE_STAT_BIRTHTIME in configure.ac checks for System.Native

akoeplinger:
The AC_CHECK_MEMBER check was done incorrectly, the include files need to be separated by newlines otherwise we would get a false negative result.

config.log contained this:

```
conftest.c:279:23: warning: extra tokens at end of #include directive [-Wextra-tokens]
#include <sys/types.h>, #include <sys/stat.h>
                      ^
                      //
```

This means that when doing File.GetCreationTime() we'd fall back to returning the last modified time.

It regressed when we switched System.IO.File to the CoreFX implementation i.e. System.Native in https://github.com/mono/mono/commit/8f5cef936491e8b20888bbf18d426482c890637c.

Fixes https://github.com/mono/mono/issues/16974


marek-safar:
@monojenkins backport 2019-08

steveisok:
@monojenkins build failed

akoeplinger:
@monojenkins backport 2019-06

PR #17005, opened 2019-09-23
Enable hw intrinsics in AOT mode

EgorBo:
Currently we don't emit hw intrinsics/SIMD code in AOT mode and it affects performance because a lot of stuff nowadays in CoreLib rely on `System.Runtime.Vector<T>` or `System.Runtime.Intrinsics.*`. We don't emit them because we know nothing about the target at compile time (unlike JIT) especially in cross-compile mode.
This PR introduces `-mattr` flag (similar to opt/llc `-mattr` [semantics](https://godbolt.org/z/SdbE53)) which allows to specify cpu attributes the codegen can use. Let me explain it in examples:

```csharp
{
    Console.WriteLine($"Lzcnt:  {System.Runtime.Intrinsics.X86.Lzcnt.IsSupported}");
    Console.WriteLine($"Popcnt: {System.Runtime.Intrinsics.X86.Popcnt.IsSupported}");
    Console.WriteLine($"Bmi1:   {System.Runtime.Intrinsics.X86.Bmi1.IsSupported}");
    Console.WriteLine($"Bmi2:   {System.Runtime.Intrinsics.X86.Bmi2.IsSupported}");
}
```
`mono --llvm p.exe` (JIT always uses current cpu):
```csharp
Lzcnt:  True
Popcnt: True
Bmi1:   True
Bmi2:   True
```
`mono --aot=llvm p.exe` (--aot uses generic CPU by default): 
```csharp
Lzcnt:  False
Popcnt: False
Bmi1:   False
Bmi2:   False
```
`mono --aot=llvm,mcpu=native p.exe` (use current cpu in AOT mode):
```csharp
Lzcnt:  True
Popcnt: True
Bmi1:   True
Bmi2:   True
```
`mono --aot=llvm,mattr=+avx2 p.exe` (allow AOT to use `AVX2` which also includes `AVX1`, `SSEx`, but doesn't include `lzcnt` and `bmi`.
```csharp
Lzcnt:  False
Popcnt: True
Bmi1:   False
Bmi2:   False
```
`mono --aot=llvm,mattr=+avx2,mattr=+lzcnt,mattr=+bmi p.exe`: (allow `AVX2`, `lzcnt` and `bmi1`)
```csharp
Lzcnt:  True
Popcnt: True
Bmi1:   True
Bmi2:   False
```
`mono --aot=llvm,mcpu=native,mattr=-lzcnt p.exe`: (use current cpu but without `lzcnt`)
```csharp
Lzcnt:  False
Popcnt: True
Bmi1:   True
Bmi2:   True
```
It works for non-llvm AOT too. If LLVM AOT is selected, it also passes `mattr` strings to `llc` and `opt`. See https://godbolt.org/z/SdbE53

TODO: revise all System.Numerics.Vector APIs to correctly detect CPU features (in a separate PR, this PR keeps Vector disabled for AOT)


mono/mini/llvm-jit.cpp
```diff
@@ -466,39 +464,6 @@ mono_llvm_dispose_ee (MonoEERef *eeref)
 {
 }
 
-MonoCPUFeatures
```
EgorBo: Moved `mono_llvm_get_cpu_features ` from llvm-jit.cpp to mini-llvm.cpp (because it's not JIT-specific and also can be used in AOT (use_current_cpu=true)

mono/mini/aot-compiler.c
```diff
@@ -7936,6 +7939,101 @@ mono_aot_split_options (const char *aot_options)
...
 }
 
+static gboolean
+parse_cpu_features (MonoCPUFeatures *cpu_features_enabled, MonoCPUFeatures *cpu_features_disabled, const gchar *str)
+{
+	*cpu_features_enabled = (MonoCPUFeatures) 0;
+	*cpu_features_disabled = (MonoCPUFeatures) 0;
+
+	gchar **attrs = g_strsplit (str, ";", -1);
+
```
vargaz: Semicolons are not good separators since they also used by the shell. Id suggest simply using multiple -mattr arguments if needed.

mono/metadata/domain-internals.h
```diff
@@ -463,6 +463,11 @@ struct _MonoDomain {
 
 	guint32 execution_context_field_offset;
 
+	/* cpu features requested via -mattr */
+	gint32 cpu_features_enabled; 
+	/* cpu features disabled via -mattr */
+	gint32 cpu_features_disabled;
+
```
vargaz: We already have an infrastructure in utils/mono-hwcap for these.

vargaz: These should be globals in mini/ somewhere, or stored in MonoJitDomainInfo.

mono/mini/aot-compiler.c
EgorBo: Done, now the syntax is `mattr=+avx2,mattr=-lzcnt,mattr=+bmi2`

EgorBo:
@monojenkins build failed

PR #17006, opened 2019-09-23
[Coop] Monitor coop reduction.

jaykrell:
Depends on https://github.com/mono/mono/pull/16987.
Not yet handled is Enter itself due to various details like
JIT intrinsic and external_only API.


jaykrell:

PR #17007, opened 2019-09-23
[Coop] Use indirect instead of volatile to presumably avoid barriers.

jaykrell:


jaykrell:
@monojenkins build failed

jaykrell:

mono/mini/mini.c
```diff
@@ -1983,7 +1983,7 @@ mono_add_var_location (MonoCompile *cfg, MonoInst *var, gboolean is_reg, int reg
 static void
 mono_apply_volatile (MonoInst *inst, MonoBitSet *set, gsize index)
 {
-	inst->flags |= mono_bitset_test_safe (set, index) ? MONO_INST_VOLATILE : 0;
+	inst->flags |= mono_bitset_test_safe (set, index) ? MONO_INST_INDIRECT : 0;
```
SamMonoRT: The function name mono_apply_volatile suggests we do volatile ? This change is not aligning with function name. Also, am getting ramped up on the code, but what barriers are we talking about here with the change ?

jaykrell: 1. I agree, and the variables too.
2. I'd have to double check, but uses of parameters/locals in the icall managed to native wrappers.

jaykrell: The point was "don't optimize", not "flush to memory across threads".

SamMonoRT: Ok, we should do the change correctly updating function names/vars too. Please update the PR.

imhameed:
From my reading of this code, `mono_apply_volatile` only ever affects `OP_ARG` or `OP_LOCAL` opcodes, which shouldn't result in any fences being emitted for subsequent loads or stores that target these variables.

`MONO_INST_INDIRECT` looks like it has the same effect w.r.t. optimization, but, as Sam says, it'd be nice if the name of this function was updated to label what effect it actually has (i.e. forces the variable to be allocated somewhere in memory).

Aside: the use of `MONO_INST_VOLATILE` to mean "acquire loads/release stores" and "the may not be duplicated, elided, or reordered relative to other volatile instructions", and "variable must exist in memory" is a bit confusing. And it's also a little weird that CIL-volatile loads and stores emit explicit fences instead of lowering to an mini-ir atomic operation; while it's probably not a huge deal, the machine code that all of this generates (if I'm reading this correctly) is not the same as the code generated by our intrinsic implementation for `System.Threading.Volatile`. As an example, CIL-volatile loads will result in an `ldr` + `dmb ish` pair on ARMv8, while `Volatile.Read` will result in `lda`.

jaykrell:
@imhameed good point, I hadn't looked closely enough, I had *assumed* and I think you are right.
I skimmed the code, and I wrote a small test with a volatile field, and the accesses are marked volatile.
I should compare the wrappers before/after, but I guess they will be the same.
I guess close this.

PR #17009, opened 2019-09-24
Another attempt to pin icall parameters.

jaykrell:
Alternative to https://github.com/mono/mono/pull/16987.

### DIFF
--- a/mono/metadata/marshal-ilgen.c
+++ b/mono/metadata/marshal-ilgen.c
@@ -6539,8 +6539,21 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
 		}
 		mono_mb_emit_ldloc_addr (mb, error_var);
 	} else {
-		for (int i = 0; i < csig->param_count; i++)
+		for (int i = 0; i < csig->param_count; i++) {
 			mono_mb_emit_ldarg (mb, i);
+
+			// Make parameters volatile to pin them.
+
+			if (get_method_image (method) && ((i == 0 && csig->hasthis) || MONO_TYPE_IS_REFERENCE (csig->params [i]) || mono_type_is_byref_internal (csig->params [i]))) {
+				if (!mb->volatile_args) {
+					gpointer const mem = mono_image_alloc0 (get_method_image (method), mono_bitset_alloc_size (csig->param_count, 0));
+					mb->volatile_args = mono_bitset_mem_new (mem, csig->param_count, 0);
+					mb->method->iflags = (mb->method->iflags & ~METHOD_IMPL_ATTRIBUTE_AGGRESSIVE_INLINING)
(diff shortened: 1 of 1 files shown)


mono/metadata/marshal-ilgen.c
```diff
@@ -6483,8 +6483,21 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
 	} else {
-		for (int i = 0; i < csig->param_count; i++)
+		for (int i = 0; i < csig->param_count; i++) {
 			mono_mb_emit_ldarg (mb, i);
+
+			// Make parameters volatile to pin them.
+
+			if (get_method_image (method) && ((i == 0 && csig->hasthis) || MONO_TYPE_IS_REFERENCE (csig->params [i]) || mono_type_is_byref (csig->params [i]))) {
+				if (!mb->volatile_args) {
+					gpointer const mem = mono_image_alloc0 (get_method_image (method), mono_bitset_alloc_size (csig->param_count, 0));
```
jaykrell: Probably should just use another allocator if there is no image. Like for dynamic? Like for Tasklets?

jaykrell: But at least this should cover the runtime/mscorlib.dll.

jaykrell:
Normal arm failure.

jaykrell:

mono/metadata/marshal-ilgen.c
```diff
@@ -6483,8 +6483,21 @@ emit_native_icall_wrapper_ilgen (MonoMethodBuilder *mb, MonoMethod *method, Mono
...
+					mb->volatile_args = mono_bitset_mem_new (mem, csig->param_count, 0);
+					mb->method->iflags = (mb->method->iflags & ~METHOD_IMPL_ATTRIBUTE_AGGRESSIVE_INLINING)
+						| MONO_METHOD_IMPL_ATTR_NOOPTIMIZATION | MONO_METHOD_IMPL_ATTR_NOINLINING;
```
jaykrell: The noopt flag does nothing in the runtime.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

lambdageek:
I think we want to continue to have some explicit marker on the caller side to indicate that the icalls are doing something special with their arguments (namely, pinning them).

jaykrell:
"continue to"?
But historically there was "nothing", right?
Well, sometimes -- calls to "internal" (that you'd have to notice is marked icall) or sometimes "icall".
And granted, the pinning-ness isn't super-historical, but recent.
Now there is my "ref", which you propose and I do not oppose, something more explicit.

I suppose if you change the rationale from "continue to" to "introduce", it makes more sense.

And then I wonder -- does caller really tend to care?
Isn't there like, a deep callgraph, and ultimately, most of the code is very unaware?

PR #17013, opened 2019-09-24
[offsets-tool] install clang python package into a virtualenv

lewurm:
So it doesn't mess with the Python installation of the user.

/cc @mandel-macaque


jaykrell:
I kinda want to rewrite offset tool a la https://github.com/jaykrell/j/blob/master/jconf/jconf1.c but not right now.

mandel-macaque:
@lewurm per se looks good, but you might want to deactivate once you are done. Since else the shell of the user will be using the python in the virtualenv.

lewurm:
I don't think we have to deactivate it because the activation is done for the current shell environment, but `make` spawns a new shell for each command.

```console
$ cat Makefile
all:
	@export HELLO=hi && echo $$HELLO Mono
	@echo $$HELLO Mono
$ make
hi Mono
Mono
```

lewurm:
@monojenkins squash

MaximLipnin:
seems that it should also be git-ignored

PR #17016, opened 2019-09-24
[interp] always require GC Unsafe mode in exception checkpoint

lewurm:
Follow-up for https://github.com/mono/mono/pull/16955

/cc @BrzVlad

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -944,6 +944,15 @@ interp_throw (ThreadContext *context, MonoException *ex, InterpFrame *frame, con
 		}									\
 	} while (0)
 
+/* Don't throw exception if thread is in GC Safe mode. Should only happen in managed-to-native wrapper. */
+#define EXCEPTION_CHECKPOINT_GC_UNSAFE	\
+	do {										\
+		if (mono_thread_interruption_request_flag && !mono_threads_is_critical_method (frame->imethod->method) && mono_thread_is_gc_unsafe_mode ()) { \
+			MonoException *exc = mono_thread_interruption_checkpoint ();	\
+			if (exc)							\
+				THROW_EX (exc, ip);					\
+		}									\
+	} while (0)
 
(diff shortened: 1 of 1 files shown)


BrzVlad:
Overall the code is a little fishy when it comes to this unsafe checking. Normally the wrapper does all the necessary transitions and we would know when to check and when not to check for exception inside the wrapper (I think this is how it works on jit). In the interpreter, I think we use this opcode for multiple scenarios, some of which require exception checking, some of which must not (and therefore the bug from the other PR).

Regardless, we should probably have a separate define, that does the state check, only for these icall opcodes. The check is redundant for other uses of EXCEPTION_CHECKPOINT (since we are by definition in unsafe state) and it might also confuse readers of code.

lewurm:
You are right, the wrapper _could_ do it:
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/metadata/marshal.c#L3447-L3449
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/metadata/marshal-ilgen.c#L2073-L2074

It looks like it would do the thread interruption regardless of the GC state?
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/metadata/marshal-ilgen.c#L1188-L1218

Not sure how this can work in the JIT then. Anyway, we specifically ask to not do generate those checks when building the wrapper for interpreter uses:
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/mini/interp/interp.c#L510
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/mini/interp/interp.c#L1683
https://github.com/mono/mono/blob/aeca4a6cc3c5fa29b29000ba52a7de7e2468181e/mono/mini/interp/interp.c#L3460

lewurm:
@monojenkins squash

PR #16919, opened 2019-09-18
[netcore] Implement missing Bmi1/Bmi2 intrinsics

EgorBo:
Implement all missing Bmi1/Bmi2 intrinsics except `MultiplyNoFlags` (no idea how to properly implement it yet).

Tested on a CPU with both bmi1 and bmi2 sets:
```csharp
using System;
using System.Runtime.CompilerServices;
using System.Runtime.Intrinsics.X86;

public class Program
{
    public static void Main(string[] args)
    {
        Console.WriteLine(Bmi1.IsSupported);
        Console.WriteLine(Bmi2.IsSupported);
        Console.WriteLine();

        // BMI1
        Console.WriteLine(AndNot1(1234, 3));
        Console.WriteLine(AndNot2(1234, 3));
        Console.WriteLine(BitFieldExtract3(1234, 3));
        Console.WriteLine(BitFieldExtract4(1234, 3));
        Console.WriteLine(ExtractLowestSetBit7(1234));
        Console.WriteLine(ExtractLowestSetBit8(1234));
        Console.WriteLine(GetMaskUpToLowestSetBit9(1234));
        Console.WriteLine(GetMaskUpToLowestSetBit10(1234));
        Console.WriteLine(ResetLowestSetBit11(1234));
        Console.WriteLine(ResetLowestSetBit12(1234));
        Console.WriteLine(TrailingZeroCount13(1234));
        Console.WriteLine(TrailingZeroCount14(1234));

        // BMI2
        Console.WriteLine(ZeroHighBits15(1234, 3));
        Console.WriteLine(ZeroHighBits16(1234, 3));
        Console.WriteLine(ParallelBitDeposit17(1234, 3));
        Console.WriteLine(ParallelBitDeposit18(1234, 3));
        Console.WriteLine(ParallelBitExtract19(1234, 3));
        Console.WriteLine(ParallelBitExtract20(1234, 3));
    }

    // BMI1
    [MethodImpl(MethodImplOptions.NoInlining)] static uint AndNot1(uint a, uint c) => Bmi1.AndNot(a, c);
    [MethodImpl(MethodImplOptions.NoInlining)] static ulong AndNot2(ulong a, ulong c) => Bmi1.X64.AndNot(a, c);
    [MethodImpl(MethodImplOptions.NoInlining)] static uint BitFieldExtract3(uint a, ushort c) => Bmi1.BitFieldExtract(a, c);
[...]

PR #17026, opened 2019-09-24
[netcore] Fix typo

EgorBo:
Accidentally commented out `System.Runtime.Intrinsics.X86.Bmi2.ZeroHighBits` case. See https://github.com/mono/mono/pull/16919#discussion_r326860408 🙂


EgorBo:
Oh, was fixed in https://github.com/mono/mono/pull/17031/files#diff-aa6079874d24b5a98a375b092f912f49L731

PR #16990, opened 2019-09-23
[draft] [Coop] Remove handles from Monitor.

jaykrell:
Depends on https://github.com/mono/mono/pull/16987.

PR #16980, opened 2019-09-22
[draft] [Coop] Remove handles from Monitor.

jaykrell:
Replace with ref to ref and set_pending.

PR #17032, opened 2019-09-25
[Coop] Mostly remove handles from Monitor.

jaykrell:
This is an alternative to https://github.com/mono/mono/pull/16980
or https://github.com/mono/mono/pull/16990.
It does *not* depend on
   https://github.com/mono/mono/pull/16987
or https://github.com/mono/mono/pull/17009
or similar.

It does *not* yet remove handles from Enter, due to Enter being intrinsic.
I still like the others but the PRs have been ignored.


jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
Because compacting/moving GC?
I guess sgen does move?
Only when all references are from heap?
And so want a more significant record
of references from stack?

jaykrell:
Maybe something in marshal-ilgen instead?
Something like https://github.com/mono/mono/pull/17009 ?

jaykrell:
The CoreCLR pattern embodies the stack requirement in the name?
Harder for people to misuse?
Enforceable?

jaykrell:
/cc @lambdageek 
(Granted a merge conflict, but not a significant one probably.)

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

akoeplinger:
@jaykrell FYI you can do `@monojenkins build monolite` now to just build monolit

jaykrell:
@akoeplinger nice thank you! That should be much quicker. I recall it taking 30 minutes just to tar up stuff unnecessarily.

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build failed

jaykrell:
/cc @lambdageek
Previously was “ok”. Good enough?

jaykrell:

jaykrell:
Old workflow just in case.

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build monolite

jaykrell:
@monojenkins build monolite

netcore/System.Private.CoreLib/src/System/Threading/LowLevelLifoSemaphore.Unix.Mono.cs
```diff
@@ -31,14 +31,18 @@ public void Dispose ()
...
 		private bool WaitCore (int timeoutMs)
 		{
 			WaitEntry wait_entry = new WaitEntry ();
 			bool mutexLocked = false;
 			bool waitEntryLocked = false;
+			bool allowInterruption = false;
 
 			try {
-				Monitor.try_enter_with_atomic_var (mutex, Timeout.Infinite, false, ref mutexLocked);
+				Monitor.try_enter_with_atomic_var (ref mutex, Timeout.Infinite, allowInterruption, ref mutexLocked);
```
lambdageek: `ref mutex` here is a ref to a field in `this` (which is on the heap), so we don't actually have a guarantee that whatever `mutex` points to is pinned.

netcore/System.Private.CoreLib/src/System/Threading/LowLevelLifoSemaphore.Unix.Mono.cs
```diff
@@ -60,9 +64,9 @@ private bool WaitCore (int timeoutMs)
 			}
 
 			try {
-				Monitor.try_enter_with_atomic_var (wait_entry.condition, Timeout.Infinite, false, ref waitEntryLocked);
+				Monitor.try_enter_with_atomic_var (ref wait_entry.condition, Timeout.Infinite, allowInterruption, ref waitEntryLocked);
```
lambdageek: this is pretty fragile - `wait_entry` happens to be a struct, so it's heap allocated, so `ref wait_entry.condition` is a pointer to the stack (and so whatever `condition` points to is pinned), but if `wait_entry` were a class, it wouldn't be.

netcore/System.Private.CoreLib/src/System/Threading/LowLevelLifoSemaphore.Unix.Mono.cs
jaykrell: I didn't realize this, but, really? It doesn't suffice?
Solution is to copy it to a local and ref that I guess, and then no need to answer prior question.

akoeplinger:

PR #17034, opened 2019-09-25
[Coop] Convert Get/SetGenericValueImpl.

jaykrell:
This is an alternative to https://github.com/mono/mono/pull/16994
and does not depend on
   https://github.com/mono/mono/pull/16987
or https://github.com/mono/mono/pull/17009
or similar.

This is not necessarily preferred. Other seems like a good idea.


jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
Always beware intrinsics and AOT special cases.

jaykrell:

jaykrell:

jaykrell:

jaykrell:

mono/mini/intrinsics.c
```diff
@@ -22,12 +22,12 @@
...
 static GENERATE_TRY_GET_CLASS_WITH_CACHE (math, "System", "Math")
 
-/* optimize the simple GetGenericValueImpl/SetGenericValueImpl generic icalls */
+/* optimize the simple GetGenericValueImpl/SetGenericValueImpl generic calls */
 static MonoInst*
 emit_array_generic_access (MonoCompile *cfg, MonoMethodSignature *fsig, MonoInst **args, int is_set)
 {
 	MonoInst *addr, *store, *load;
-	MonoClass *eklass = mono_class_from_mono_type_internal (fsig->params [2]);
+	MonoClass *eklass = mono_class_from_mono_type_internal (fsig->params [1]);
```
jaykrell: This is surprising. We'll see if it works.

netcore/System.Private.CoreLib/src/System/Array.cs
```diff
@@ -496,15 +496,29 @@ internal int GetElementSize ()
...
+
+		// This is a special case in the runtime.
+		internal void GetGenericValueImpl<T> (int pos, out T value)
+		{
+			var self = this;
+			GetGenericValue_icall (ref self, pos, out value);
+		}
+
+		// This is a special case in the runtime.
+		internal static void SetGenericValueImpl<T> (int pos, ref T value)
```
jaykrell: Remove `internal` that got added. And elsewhere.

jaykrell:

mono/mini/intrinsics.c
jaykrell: It might also make sense to reference the generic/inflated stuff?

jaykrell:

jaykrell:

jaykrell:

jaykrell:

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
/cc @lambdageek 
(Granted a merge conflict, but not a significant one probably.)

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

mcs/class/corlib/System/Array.cs
```diff
@@ -204,16 +208,31 @@ internal void InternalArray__set_Item<T> (int index, T item)
...
-		internal extern void GetGenericValueImpl<T> (int pos, out T value);
+		extern static void GetGenericValue_icall<T> (ref Array self, int pos, out T value);
 
 		// CAUTION! No bounds checking!
 		[MethodImplAttribute (MethodImplOptions.InternalCall)]
-		internal extern void SetGenericValueImpl<T> (int pos, ref T value);
+		extern static void SetGenericValue_icall<T> (ref Array self, int pos, ref T value);
+
+		// This is a special case in the runtime.
+		internal void GetGenericValueImpl<T> (int pos, out T value)
```
lambdageek: Do we need to protect this from being inlined so that the call gets replaced by an intrinsic?

mcs/class/corlib/System/Array.cs
```diff
@@ -204,16 +208,31 @@ internal void InternalArray__set_Item<T> (int index, T item)
...
+		{
+			var self = this;
+			GetGenericValue_icall (ref self, pos, out value);
+		}
+
+		// This is a special case in the runtime.
+		internal void SetGenericValueImpl<T> (int pos, ref T value)
```
lambdageek: Protect from inlining?

jaykrell: Good point, I hadn't thought about it or verified it, that intrinsification should dominate inlining, but, um 1. I think it is fairly clearly ordered that way in the JIT 2. All the other functions, many of which, I think also have IL.
It is worth I guess some asserts, however, that will require sharing intrinsic-identification logic.

lambdageek: > I think it is fairly clearly ordered that way in the JIT 
Ah ok, I wasn't sure.

> It is worth I guess some asserts

I wasn't necessarily suggesting that.  More just wondering whether you've seen evidence that  `emit_array_generic_access` is getting called after your changes.

jaykrell: I admit, I didn't look, just assumed.
Except, I do know the signature change/index is needed.
Things failed without it.
I should do some verbose or aot dumps.

PR #17035, opened 2019-09-25
  BeginConnect complete to early when not using callback.

lateralusX:
The following commit, 068e8fe changed the way we do DNS resolve in BeginConnect using host name. Instead of doing a blocking DNS resolve, it switched to an async DNS resolve, followed by an async connect. The returned IAsyncResult was just representing the completion of the DNS resolve task and not the underlying connect. It
also turns out that part of doing EndConnect, we did some validation of the used IAsyncResult making sure it was an instance of SocketAsyncResult, meaning that if EndConnect was called on the returned IAsyncResult from BeginConnect, EndConnect would throw and exception.

NOTE, this only happens if you use pattern like this:

var result = sock.BeginConnect (hostName, port, null, null);
result.AsyncWaitHandle.WaitOne (2000);
sock.EndConnect (result);

It did not happened if you used callbacks, since then EndConnect would be called as part of callback and that IAsyncResult was of correct type, or if you did a BeginConnect with an EndPoint or IPAddress[] since that won't trigger the DNS resolve.

Fix is to make sure we hand out the SocketAsyncResult instance that will be used in our internal connect calls from BeginConnect. We also make sure, we chain the DNS resolve task together with a task doing the async connect (using the same SocketAsyncResult instance). This will make sure
the IAsyncResult returned from BeginConnect is the same one as used in internal connect call, so it will only complete when the connect call completes.


mcs/class/System/Test/System.Net.Sockets/SocketTest.cs
```diff
@@ -512,7 +512,7 @@ public void SocketErrorTest ()
 			sock.BeginConnect (ep, new AsyncCallback(SocketError_callback),
 				sock);
 
-			if (SocketError_event.WaitOne (2000, false) == false) {
+			if (SocketError_event.WaitOne (5000, false) == false) {
```
MaximLipnin: Is it the only place where the timeout needs to be increased?

lateralusX: This was just something I saw run local tests on Windows. Not related to this change, but since I hit it locally I though we could just increase it. I have not seen issues with this on CI.

steveisok:
@baulig Please review

mcs/class/System/System.Net.Sockets/Socket.cs
```diff
@@ -999,7 +999,7 @@ public static void CancelConnectAsync (SocketAsyncEventArgs e)
 			}
 		});
 
-		public IAsyncResult BeginConnect (string host, int port, AsyncCallback requestCallback, object state)
+		public IAsyncResult BeginConnect (string host, int port, AsyncCallback callback, object state)
 		{
```
baulig: Verified this; the new name matches the one in CoreFx.  https://github.com/dotnet/corefx/blob/8d6320189d6df167680a203422a2235d9befb9f4/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L2069

mcs/class/System/Test/System.Net.Sockets/SocketTest.cs
baulig: Just cosmetic; but we might want to extract those timeouts into constants to make it easier to change them.

lateralusX:
@monojenkins commit csproj

baulig:
We will need the parameter rename when landing my socket changes (and it will also simplify testing) - could you please take that last commit out again?

lateralusX:
Name change of parameter restored.

lateralusX:
@monojenkins commit apidiff

steveisok:
@monojenkins backport 2019-08

Issue #8404, opened 2018-04-22
Verify intrinsics work as expected


marek-safar:
We are missing tests to ensure we don't regress in this area. We could mitigate that by following.

- Write C# tests which call intrinsic implementation with error conditions and expect certain stack trace in the exception which should be different to the managed implementation body. This won't be always possible but many cases have that.
- When we are not running with disabled intrinsic any method decorated with IntrinsicAttribute attribute and hit by JIT (or AOT compiled) should throw execution exception or abort the compilation. We are already using this attribute we just need to add it everywhere.

PR #17047, opened 2019-09-25
Optimize and extend System.Array intrinsic identification.

jaykrell:
i.e. do the strcmp last.
Add GetRawArrayData.

### DIFF
--- a/mono/mini/intrinsics.c
+++ b/mono/mini/intrinsics.c
@@ -736,11 +736,11 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
 		} else
 			return NULL;
 	} else if (cmethod->klass == mono_defaults.array_class) {
-		if (strcmp (cmethod->name, "GetGenericValueImpl") == 0 && fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt)
+		if (fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt && strcmp (cmethod->name, "GetGenericValueImpl") == 0)
 			return emit_array_generic_access (cfg, fsig, args, FALSE);
-		else if (strcmp (cmethod->name, "SetGenericValueImpl") == 0 && fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt)
+		else if (fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt && strcmp (cmethod->name, "SetGenericValueImpl") == 0)
 			return emit_array_generic_access (cfg, fsig, args, TRUE);
-		else if (!strcmp (cmethod->name, "GetRawSzArrayData")) {
+		else if (!strcmp (cmethod->name, "GetRawSzArrayData") || !strcmp (cmethod->name, "GetRawArrayData")) {
 			int dreg = alloc_preg (cfg);
 			EMIT_NEW_BIALU_IMM (cfg, ins, OP_PADD_IMM, dreg, args [0]->dreg, MONO_STRUCT_OFFSET (MonoArray, vector));
 			return ins;
--- a/netcore/System.Private.CoreLib/src/System/Array.cs
+++ b/netcore/System.Private.CoreLib/src/System/Array.cs
@@ -429,12 +429,14 @@ namespace System
 			return GetLowerBound (dimension) + GetLength (dimension) - 1;
 		}
 
+		// This is intrinsic.
 		[MethodImpl (MethodImplOptions.AggressiveInlining)]
 		internal ref byte GetRawSzArrayData ()
 		{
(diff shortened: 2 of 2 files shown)


netcore/System.Private.CoreLib/src/System/Array.cs
```diff
@@ -429,12 +429,14 @@ public int GetUpperBound (int dimension)
...
 		}
 
+		// This is intrinsic.
 		[MethodImpl (MethodImplOptions.AggressiveInlining)]
 		internal ref byte GetRawSzArrayData ()
 		{
 			return ref Unsafe.As<RawData>(this).Data;
 		}
 
+		// This is intrinsic.
```
EgorBo: There is a special attribute for it: `[Intrinsic]`

jaykrell: We seem to have far more intrinsics than `[Intrinsic]`s.
I just keep tripping over them, figured I should start documenting.
i.e. I had changed the signature here -- crash.

marek-safar: Please use it (so we can use it by other tools)

jaykrell: It is just a hint to humans, right?
I mean, if I use it accidentally too much or too little, no harm done, no big concern about correctness?

I guess we could actually try to get it correct, and speed up all the checks as a result.

It almost seems like, hint to linker -- remove this IL. But I don't think that ends up working. Unless maybe reflection is also removed? Is that the idea? Like, actually more than a hint, but significant behavior?

EgorBo: @jaykrell Currently it's just a hint but I guess we indeed can use them to speed up checks. 👍

marek-safar: yes, we plan to use it as an optional hint for linker. For the rest check #8404

EgorBo: the only problem:
Some `[Intrinsics]` are CoreCLR only, some are Mono only, some are LLVM only (e.g. `Buffer.Memmove` lives in coreclr/shared and it's only [intrinsic] for Mono-LLVM, etc.

marek-safar: That's fine, the attribute should be used when it's intrinsic on most configurations

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins build failed

jaykrell:
@monojenkins rebuild failed

jaykrell:
@monojenkins build

mono/mini/intrinsics.c
```diff
@@ -736,11 +736,11 @@ mini_emit_inst_for_method (MonoCompile *cfg, MonoMethod *cmethod, MonoMethodSign
...
 			return NULL;
 	} else if (cmethod->klass == mono_defaults.array_class) {
-		if (strcmp (cmethod->name, "GetGenericValueImpl") == 0 && fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt)
+		if (fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt && strcmp (cmethod->name, "GetGenericValueImpl") == 0)
 			return emit_array_generic_access (cfg, fsig, args, FALSE);
-		else if (strcmp (cmethod->name, "SetGenericValueImpl") == 0 && fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt)
+		else if (fsig->param_count + fsig->hasthis == 3 && !cfg->gsharedvt && strcmp (cmethod->name, "SetGenericValueImpl") == 0)
 			return emit_array_generic_access (cfg, fsig, args, TRUE);
-		else if (!strcmp (cmethod->name, "GetRawSzArrayData")) {
+		else if (!strcmp (cmethod->name, "GetRawSzArrayData") || !strcmp (cmethod->name, "GetRawArrayData")) {
```
jaykrell: Reviewers please do note that I have opportunistically added an intrinsic here.
This is the "extended" part of the PR description.
I just happened to notice, that its current implementation is identical to the other function.
I do not know:
 1. If it is called often, is performance critical, what difference this made.
 2. If it is a temporary coincidence or long term architectural thing. It does seem just like a rename of the other, so long term architectural.

However the intrinsic notation (or even a comment) should assist in case I am wrong.
i.e. anyone changing the implementation needs to heed the comment/attribute and either leave it alone, or update the JIT, or, worst case, stop it from being an intrinsic and just use the C#, i.e. undo this change as needed.

jaykrell:
@monojenkins build failed

PR #17062, opened 2019-09-25
[2019-06] Fix time zone issue when jumping into DST

steveisok:
There is an issue with jumping into DST for some time zones when the incorrect date-time offset is returned for date-time in UTC (which comes from DateTime.Now). The fix is to just check if the incoming date-time is in UTC

Also I added set of tests for some time zones but they verify jumping into DST in general.

Fixes https://github.com/mono/mono/issues/16395

Backport of https://github.com/mono/mono/pull/16430


steveisok:
@monojenkins build failed

Issue #16460, opened 2019-08-25
System.IO.Path.GetFullPath("\\\\.\\pipe\\pipename"): different results on Mono and .NET


jspam:
## Steps to Reproduce

1. Compile the following test program:
```cs
class MainClass
{
    public static void Main (string[] args) {
        System.Console.WriteLine(System.IO.Path.GetFullPath("\\\\.\\pipe\\pipename"));
    }
}
```
2. Execute it with `mono Program.exe`
3. Execute it with `Program.exe` (using .NET)

### Current Behavior

Output with `mono Program.exe` is: `\\.\pipe\\pipe\pipename`
Output with `Program.exe` is: `\\.\pipe\pipename`

### Expected Behavior

Output is the same in both cases. The output by .NET (leaving the path unchanged) seems the correct one.

## On which platforms did you notice this

[ ] macOS
[ ] Linux
[X] Windows

**Version Used**:

Mono 6.7.0 (nightly)

## Note

`GetFullPath` is used in this fashion internally by `NamedPipeClientStream`, so this behavior basically makes `NamedPipeClientStream` unusable for me.

directhex:
Looks like some bad assumptions in CanonicalizePath in Path.cs which don't account for the named pipe naming format (UNC in the broad sense works, it's a specific mishandling of the syntax used for pipes)

directhex:
```
c:\Users\joshield\Desktop>mono test.exe
UNC pipe name (should be \\.\pipe\pipename): \\.\pipe\\pipe\pipename
UNC network name (should be \\servername\sharename\dirname): \\servername\sharename\dirname
UNC DOS name (should be \\?\C:\Users): \\?\C:\Users

c:\Users\joshield\Desktop>test.exe
UNC pipe name (should be \\.\pipe\pipename): \\.\pipe\pipename
UNC network name (should be \\servername\sharename\dirname): \\servername\sharename\dirname
UNC DOS name (should be \\?\C:\Users): \\?\C:\Users

c:\Users\joshield\Desktop>type test.cs
using System.Runtime.Versioning;

[assembly:TargetFramework(".NETFramework,Version=v4.7")]

class MainClass
{
    public static void Main (string[] args) {
        System.Console.WriteLine("UNC pipe name (should be \\\\.\\pipe\\pipename): {0}", System.IO.Path.GetFullPath("\\\\.\\pipe\\pipename"));
[...]

PR #17066, opened 2019-09-25
Special case in GetFullPath for UNC roots of \\.\

directhex:
A valid path might be something like \\\\.\\pipe\\pipename (i.e. \\\\.\\ is a valid root). If we blindly strip directories named `.` then we end up confusing our path cleaning algorithm. So permit a directory component of `.` in UNC path roots.

Closes: https://github.com/mono/mono/issues/16460

```
UNC pipe name (input is \\.\pipe\pipename): \\.\pipe\pipename
UNC network name (input is \\servername\sharename\dirname): \\servername\sharename\dirname
UNC DOS name (input is \\?\C:\Users\.\joshield): \\?\C:\Users\joshield
```

### DIFF
--- a/mcs/class/corlib/System.IO/Path.cs
+++ b/mcs/class/corlib/System.IO/Path.cs
@@ -734,7 +734,7 @@ namespace System.IO {
 				if (Environment.IsRunningOnWindows)
 					dirs[i] = dirs[i].TrimEnd ();
 				
-				if (dirs[i] == "." || (i != 0 && dirs[i].Length == 0))
+				if ((!(isUnc && i == 2) && dirs[i] == ".") || (i != 0 && dirs[i].Length == 0))
 					continue;
 				else if (dirs[i] == "..") {
 					// don't overwrite path segments below the limit


jaykrell:
It's not a special case, it's just the right way.

directhex:
It is special cased, in that our CanonicalizePath code splits paths into directory components and drops all cases where the component is just `.` - the special case is "except where it's a UNC path, and the . is the third character i.e. the UNC path begins `\\.`"

directhex:
@monojenkins build failed

kg:
It might be good to try and dig up a spec for how these paths work and make sure we have coverage for other interesting cases

directhex:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#namespaces ?

kg:
Seems right. From that page I don't see anything unique that merits tests other than parsing GLOBAL?? paths correctly, but I also feel like nobody would ever end up actually doing that so I think we're set.

directhex:
I'm not opposed to adding more tests, I just took one look at GetFullPath_Windows in mcs/class/corlib/Test/System.IO/PathTest.cs and nope'd out of there

jaykrell:
One wonders about dot at the end.

c:\foo\.
c:\foo

Some people like to put trailing slashes to mean directory. (not me)
Some people like to trim trailing slashes. (me)
Sometimes the dot might cause to keep the trailing slash.
But on the other hand, I don't know, this stuff is just really messy.

And explicit dot at the start:

LoadLibrary(.\foo.dll) is different than LoadLibrary(foo.dll).

LoadLibrary in particular. Not CreateFile. Implies a search path.
Either .\ means the one and only current directory, or it is shopped around and appended to each element in %PATH%. My best is the first, but it could be the second, or other.

directhex:
@jaykrell we should check whether on Windows we correctly ignore all trailing dots, but that's a distinct bug from this one IMHO

directhex:
@jaykrell right now in GetFullPath we do NOT strip trailing dots, and GetFullPath(".\foo.dll") is treated the same as GetFullPath("foo.dll"), i.e. both prepend cwd

jaykrell:
I wonder why Windows doesn't defer to GetFullPathname.

directhex:
Performance

jaykrell:
And bugs?

directhex:
Bugs aren't real, they can't hurt you

PR #17075, opened 2019-09-26
Revert "[2019-06][metadata] Fix leaks when handling a few attributes (#16675)"

marek-safar:
Reverts mono/mono#16851


thaystg:
@monojenkins build failed

PR #17081, opened 2019-09-26
[2019-06] Update NETStandard 2.1 APIs

EgorBo:
Contributes to #17064


marek-safar:

marek-safar:

mcs/class/Facades/netstandard/Stubs.cs
```diff
@@ -0,0 +1,87 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Data
```
akoeplinger: I think we need to make sure these stubs are only included on profiles that don't have System.Data.DataSetExtensions.dll otherwise we'd end up with the same type in different assemblies.

So wrap this in `#if MOBILE || XAMMAC_4_5` and add the typeforwarders + assembly reference for the net_4_x case.

EgorBo: @akoeplinger something like this ? https://github.com/mono/mono/pull/17081/commits/d6678820a961fa4fbbe7426e2198a412bba50d37

akoeplinger: @EgorBo yes, except that xammac_net_4_5 doesn't have System.Data.DataSetExtensions.dll either so you'll need to remove the LIB_REFS there.

And we also need to add typeforwards for these types in the `!MOBILE && !XAMMAC_4_5` case.

EgorBo: ok, I hope it's now fixed

akoeplinger:

marek-safar:
@akoeplinger could you please validate the diff after this change

PR #17097, opened 2019-09-27
[netcore] Add target for running coreclr tests

marek-safar:


filipnavara:
On a related note, do we want to use `corerun` to run the CoreFX tests after this PR lands?

marek-safar:
@filipnavara no, keeping them to run same in the same setup as in corefx repo would be better.

filipnavara:
@marek-safar Both CoreFX and CoreCLR run them using `corerun` AFAIK

filipnavara:
The documentation in CoreFX looks a bit mixed on what is currently used but it looks like it uses `Microsoft.NETCore.TestHost` package which comes from CoreCLR repository and contains pre-build `corerun`.

netcore/Makefile
```diff
@@ -234,6 +239,36 @@ build-test-corefx-%:
 xunit-summary:
 	./xunit-summary.py "corefx/tests"
 
+
+update-tests-coreclr: coreclr/.stamp-tests
+
+coreclr/.stamp-tests:
+	rm -rf coreclr/tests
```
filipnavara: The `coreclr` directory is never created on my machine and all the successive step consequently fail.

netcore/Makefile
```diff
@@ -234,6 +239,36 @@ build-test-corefx-%:
...
+	curl -L "$(CORECLR_TESTS)" --output coreclr/tests.zip
+	$(UNZIPCMD) coreclr/tests.zip -C coreclr/
```
filipnavara: It downloads a `.zip` file but tries to extract it with `tar` on Linux which fails.

PR #17100, opened 2019-09-27
Fix mono-extensions build

akoeplinger:
It needed an update because https://github.com/mono/mono/commit/5b108b5726ec196f4b61d30f406919ce70211754 touched code the patch modified.


akoeplinger:
@monojenkins build pkg

Issue #15235, opened 2019-06-18
Implement System.GC.AllocateUninitializedArray


naricc:
Implement System.GC.AllocateUninitializedArray.

https://github.com/mono/mono/blob/master/netcore/System.Private.CoreLib/src/System/GC.cs#L245

This should return a new array with a best-effort attempt to avoid the cost of zeroing it. Optimized for large arrays. It should only work if T is a non-reference type, otherwise it should throw an exception.

naricc:
I have been looking into this issue, and I am not confident it can actually improve performance. Here is the only place in the GC where we explicitly zero memory before allocating: https://github.com/mono/mono/blob/6023f36749c56bf5c29fd457dc039204f68015bc/mono/sgen/sgen-los.c#L400

In other cases, memory is zeroed because it comes from the nursery (which we cannot avoid zeroing) or from the OS directly (in which case the OS zeroes it). 

So this zeroing only happens for objects sized big enough that we allocate them in the large object space, but small enough that we do not need to allocate directly from the OS.

I won't know for sure unless I implement it, but I think the over head of the call would swamp any improvement from reduced zeroing. Does it actually make sense to implement this as a call into the runtime?

@BrzVlad @marek-safar @vargaz

BrzVlad:
There shouldn't be any overhead if we really want to do it. Personally I don't think it is worth it, it might add some ugliness. I'm not sure our LOS allocator is that good anyway so even if we do this it might not matter. I recommend for now to check the performance impact of not doing the zeroing. Just have some C# microbenchmark that allocates large enough array in a loop. Check the time it takes with default runtime vs a hacked runtime that does no zeroing and report back the numbers. Then we can decide if it really is worth it.

marek-safar:
I think for large arrays (MBs) this can make a difference. The API was introduced for performance reasons.

naricc:
So, for arrays larger than 1 mb (Actually, larger than 1 MB - 4kb - some object header stuff) we grab a chunk of memory directly from the OS which comes pre-zerod. There is no way to save zeroing there.

For objects large enough to be allocated in the LOS, but less than that size, we do zero them explictly. I did the experiment Vlad suggested and wrote a program that allocates 100,000 arrays of that size in a loop, and just removed the zeroing. 

Results:

Best time with zeroing: 55.237 seconds
Best time with out zeroing: 55.235 seconds

It seems like this means its not worth implementing the change, although we may revisit it if we ever change the LOS.

[...]

PR #17105, opened 2019-09-27
Add a comment explaining why GC.AllocateUninitializedArray is not an icall

naricc:
I experimentally determined that implementing GC.AllocateUninitializedArray() is unlikely to save any time. To do this I constructed a program that allocates 100k arrays of the maximum size for which we do explicit zeroing, and modified the VM to skip zeroing. This resulted in no measurable speed up.

This change is a result of this issue: https://github.com/mono/mono/issues/15235

### DIFF
--- a/netcore/System.Private.CoreLib/src/System/GC.cs
+++ b/netcore/System.Private.CoreLib/src/System/GC.cs
@@ -244,7 +244,11 @@ namespace System
 
 		internal static T[] AllocateUninitializedArray<T> (int length)
 		{
-			// TODO: Implement
+			// Mono only does explicit zeroning if the array is to big for the nursery, but less than 1 Mb - 4 kb.
+			// If it is bigger than that, we grab memroy directly from the OS which comes pre zeroed.
+			// Experimentation shows that if we just skip the zeroing in this case, we do not save a measurable
+			// amount of time. So we just allocate the normal way here.
+			// Revist if we change LOS implementation.
 			return new T [length];
 		}
 	}
--- a/netcore/System.Private.CoreLib/src/System/GC.cs
+++ b/netcore/System.Private.CoreLib/src/System/GC.cs
@@ -245,7 +245,7 @@ namespace System
 		internal static T[] AllocateUninitializedArray<T> (int length)
 		{
 			// Mono only does explicit zeroning if the array is to big for the nursery, but less than 1 Mb - 4 kb.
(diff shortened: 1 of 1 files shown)


netcore/System.Private.CoreLib/src/System/GC.cs
```diff
@@ -244,7 +244,11 @@ public static GCMemoryInfo GetGCMemoryInfo ()
 
 		internal static T[] AllocateUninitializedArray<T> (int length)
 		{
-			// TODO: Implement
+			// Mono only does explicit zeroning if the array is to big for the nursery, but less than 1 Mb - 4 kb.
+			// If it is bigger than that, we grab memroy directly from the OS which comes pre zeroed.
+			// Experimentation shows that if we just skip the zeroing in this case, we do not save a measurable
+			// amount of time. So we just allocate the normal way here.
+			// Revist if we change LOS implementation.
```
Therzok:
```suggestion
			// Revisit if we change LOS implementation.
```

naricc:
@monojenkins build failed

netcore/System.Private.CoreLib/src/System/GC.cs
```diff
@@ -244,7 +244,11 @@ public static GCMemoryInfo GetGCMemoryInfo ()
 
 		internal static T[] AllocateUninitializedArray<T> (int length)
 		{
-			// TODO: Implement
+			// Mono only does explicit zeroning if the array is to big for the nursery, but less than 1 Mb - 4 kb.
+			// If it is bigger than that, we grab memroy directly from the OS which comes pre zeroed.
```
lambdageek:
```suggestion
			// If it is bigger than that, we grab memory directly from the OS which comes pre zeroed.
```

netcore/System.Private.CoreLib/src/System/GC.cs
```diff
@@ -244,7 +244,11 @@ public static GCMemoryInfo GetGCMemoryInfo ()
 
 		internal static T[] AllocateUninitializedArray<T> (int length)
 		{
-			// TODO: Implement
+			// Mono only does explicit zeroning if the array is to big for the nursery, but less than 1 Mb - 4 kb.
```
lambdageek:
```suggestion
			// Mono only does explicit zeroning if the array is too big for the nursery, but less than 1 Mb - 4 kb.
```

filipnavara: Also, s/zeroning/zeroing/.

PR #15234, opened 2019-06-18
[debugger] Debugger doesn't break on unhandled exceptions 

thaystg:
When the try catch found is in a MONO_WRAPPER_RUNTIME_INVOKE method (in this case in "runtime_invoke_void_object" method) the exception should be treated as an UNHANDLED by the debugger.

Fixes #15203

PR #17106, opened 2019-09-27
[debugger][exception] Debugger breaks on handled exceptions

### LINKED ISSUE #17083: Regression: when debugger encounters a not-unhandled exception, it stops and shows "Dns.cs file not found" (Dns being the BCL class used)
I believe this is a regression because I haven't seen this behaviour in previous versions of VisualStudio for Mac (now using 8.2.4; not sure the exact version when this started happening, possibly when Mono 6.0 was bundled).

Steps to reproduce:
1. Open VS4Mac, create new project/solution, Console type, language: C#.
2. Use this simple snippet:

```
using System;
using System.Linq;
using System.Net.Sockets;
using System.Threading.Tasks;

namespace TestResolveCSharpConsole
{
    class MainClass
    {
        {
            try
            {
                var results = ResolveAsync().GetAwaiter().GetResult();
            }
            catch (SocketException sockEx)
            {
                Console.WriteLine("correctly handled");
            }
        }

        public static async Task<List<string>> ResolveAsync()
        {
            var addresses = await System.Net.Dns.GetHostAddressesAsync("foo.bar.baz");
            return new List<string>(addresses.Select(addr => addr.ToString()));
        }
    }
}
```

3. Compile it. (If you run it outside VS4Mac, result should just be the printing of the string "correctly handled".)
4. Run it with the debugger inside VS4Mac.

Expected results:
Should just run it and not stop, finish execution at the end like it's the case when being run outside of VS4Mac.

Current results:
Debugger seems to bomb when a `SocketException` is thrown, even if it's properly handled with a try-catch (this is the first bug). Then it tries to open the source code of the BCL where this exception happened (Dns.cs), telling the weird error "Source Not Found" to the user.

See screenshot:
Screenshot 2019-08-28 at 3 56 16 PM

> VS bug [#973786](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/973786), VS bug [#989287](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/989287)

thaystg:
If there is a perform_wait_callback in the stack there will be another catch generated by the method called in the owner thread, so we don't need to throw as unhandled exception, we can continue and find the next catch.

Fixes #17083

### DIFF
--- a/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
+++ b/mcs/class/Mono.Debugger.Soft/Test/dtest-app.cs
@@ -507,6 +507,10 @@ public class Tests : TestsBase, ITest2
 			unhandled_exception_wrapper ();
 			return 0;
 		}
+		if (args.Length >0 && args [0] == "unhandled-exception-perform-wait-callback") {
+			unhandled_exception_perform_wait_callback ();
+			return 0;
+		}
 		if (args.Length >0 && args [0] == "unhandled-exception-endinvoke") {
 			unhandled_exception_endinvoke ();
 			return 0;
@@ -1647,6 +1651,24 @@ public class Tests : TestsBase, ITest2
 	}
 	[MethodImplAttribute (MethodImplOptions.NoInlining)]
 	public static void unhandled_exception_endinvoke_2 () {
+	}
+
+	[MethodImplAttribute (MethodImplOptions.NoInlining)]
+	public static void unhandled_exception_perform_wait_callback () {
+		try
+		{
+			var results = ResolveAsync().GetAwaiter().GetResult();
+		}
+		catch (SocketException sockEx)
+		{
+			//Console.WriteLine("correctly handled");
+		}
+	}
+
+	public static async Task<List<string>> ResolveAsync()
+	{
+		var addresses = await System.Net.Dns.GetHostAddressesAsync("foo.bar.baz");
+		return new List<string>(addresses.Select(addr => addr.ToString()));
 	}
 
 	[MethodImplAttribute (MethodImplOptions.NoInlining)]
--- a/mcs/class/Mono.Debugger.Soft/Test/dtest.cs
+++ b/mcs/class/Mono.Debugger.Soft/Test/dtest.cs
@@ -4282,6 +4282,23 @@ public class DebuggerTests
 	}
 
 	[Test]
+	public void UnhandledException4 () {
+		vm.Exit (0);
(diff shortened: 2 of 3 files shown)


thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build OS X x64

thaystg:
@monojenkins build failed

thaystg:
@monjenkins build Linux ARMv5

marek-safar:
@monojenkins build failed

marek-safar:
@monojenkins backport 2019-08

vargaz:
Is perform_wait_callback a special method so it needs to be special cased in the runtime ?

lambdageek:
@vargaz `PerformWaitCallback` is a little bit special to the runtime.  The threadpool code sets `save_lmf` to TRUE because the worker loop wants to mess with the exception handling.

https://github.com/mono/mono/blob/7f37f76b6d75ed6236f6f33958784180e6f20446/mono/metadata/threadpool.c#L326-L330

vargaz:
I guess this is ok then but it still looks kinda weird.

knocte:
I don't want to delay this fix to be committed (in fact I was the bug reporter), but hardcoding threadpool_perform_wait_callback_method in this bug fix would mean that the bug would still be present when the thread is not a threadpool thread?

thaystg:
It's a side effect of #15234. We should backport to 2019-06 and 2019-02.

lambdageek:

lambdageek:
@monojenkins build failed

akoeplinger:
@monojenkins build

akoeplinger:
@monojenkins backport 2019-08

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

thaystg:
@monojenkins build failed

lambdageek:
Relevant failures: x64 and arm64 FullAOT+Interpreter lanes consistently failing the abort-cctor test.  Also on the 2019-08 backport.

thaystg:
@monojenkins build failed

thaystg:
@monojenkins backport 2019-08

thaystg:
@monojenkins backport 2019-06

thaystg:
@monojenkins backport 2019-02

PR #17108, opened 2019-09-28
[WIP] [interp] Add an interp-max-frames= option to MONO_DEBUG to limit the number of frames in the interpreter.

vargaz:

### DIFF
--- a/mono/mini/interp/interp.c
+++ b/mono/mini/interp/interp.c
@@ -3269,6 +3269,14 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 		sp++;
 	}
 
+	context->nframes ++;
+	if (G_UNLIKELY (mini_get_debug_options ()->interp_max_frames)) {
+		if (context->nframes == mini_get_debug_options ()->interp_max_frames) {
+			mini_get_debug_options ()->interp_max_frames = 0;
+			THROW_EX (mono_get_exception_stack_overflow (), ip);
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


mono/mini/interp/interp.c
```diff
@@ -3269,6 +3269,14 @@ interp_exec_method_full (InterpFrame *frame, ThreadContext *context, FrameClause
 		sp++;
 	}
 
+	context->nframes ++;
+	if (G_UNLIKELY (mini_get_debug_options ()->interp_max_frames)) {
+		if (context->nframes == mini_get_debug_options ()->interp_max_frames) {
+			mini_get_debug_options ()->interp_max_frames = 0;
```
kg: Is there a particular reason why you clear this after the first time it activates? It feels unnecessary and in the context of wasm (where you could be calling in from outside even after a throw has happened) it means you'd have no way to re-enable the recursion limit after the first time it's hit.

Also, a stack overflow exception can't be caught unless it's thrown by "user code", IIRC - will this behave like a user throw or a runtime throw? If the former that would mean the executing code could unintentionally trap the stack overflow and clear the flag. I'm guessing this is a runtime throw, in which case this should reliably throw and abort when the limit is first hit.

BrzVlad:
In general, I'm paranoid about adding any unnecessary stuff to the interpreter execution loop. Don't know exactly what is the target use case for this, but it would be great if it would be part of compile time constant. We could use this constant to also enable, in the future, other features that normally slow down interpreter, like opcode counts, opcode tracing etc.

vargaz:
This is just a diagnostics tool to help diagnose some wasm problems.

PR #17110, opened 2019-09-28
Factor out the context capturing on entering the GC safe region.

filipnavara:
The compiler is free to save registers to the stack and reuse them for other
variables. If the context is captured at that point we may inadvertently create
GC hole because the stack is later overwritten when returning out of
`mono_threads_enter_gc_safe_region_*` calls and hence a reference won't
be captured in neither the registers nor the stack.

This is NOT a full fix, just intermediate step to verify that the problem
goes away.

### DIFF
--- a/mono/dis/main.c
+++ b/mono/dis/main.c
@@ -1979,9 +1979,10 @@ usage (void)
 	exit (1);
 }
 
-static void
+static gboolean
 monodis_thread_state_init_from_monoctx (MonoThreadUnwindState *ctx, MonoContext *mctx)
 {
+	return FALSE;
 }
 
 #define monodis_setup_async_callback          NULL
--- a/mono/dis/main.c
+++ b/mono/dis/main.c
@@ -1980,7 +1980,7 @@ usage (void)
 }
 
 static void
-monodis_thread_state_init (MonoThreadUnwindState *ctx)
+monodis_thread_state_init_from_monoctx (MonoThreadUnwindState *ctx, MonoContext *mctx)
 {
 }
 
--- a/mono/unit-tests/test-conc-hashtable.c
+++ b/mono/unit-tests/test-conc-hashtable.c
@@ -319,7 +319,7 @@ benchmark_glib (void)
 }
 
 static void
-monotest_thread_state_init (MonoThreadUnwindState *ctx)
+monotest_thread_state_init_from_monoctx (MonoThreadUnwindState *ctx, MonoContext *mctx)
 {
 }
 
--- a/mono/unit-tests/test-conc-hashtable.c
+++ b/mono/unit-tests/test-conc-hashtable.c
@@ -318,9 +318,10 @@ benchmark_glib (void)
 	g_hash_table_destroy (h);
 }
 
-static void
+static gboolean
 monotest_thread_state_init_from_monoctx (MonoThreadUnwindState *ctx, MonoContext *mctx)
 {
+	return FALSE;
 }
 
 #ifdef __cplusplus
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -130,6 +130,10 @@ mono_threads_state_poll (void)
 static void
 mono_threads_state_poll_with_info (MonoThreadInfo *info)
 {
+	MonoContext ctx;
+
+	MONO_CONTEXT_GET_CURRENT (ctx);
+
 	g_assert (mono_threads_is_blocking_transition_enabled ());
 
 	++coop_do_polling_count;
@@ -147,7 +151,7 @@ mono_threads_state_poll_with_info (MonoThreadInfo *info)
 		return;
 
 	++coop_save_count;
-	mono_threads_get_runtime_callbacks ()->thread_state_init (&info->thread_saved_state [SELF_SUSPEND_STATE_INDEX]);
+	mono_threads_get_runtime_callbacks ()->thread_state_init_from_monoctx (&info->thread_saved_state [SELF_SUSPEND_STATE_INDEX], &ctx);
 
 	/* commit the saved state and notify others if needed */
 	switch (mono_threads_transition_state_poll (info)) {
@@ -266,7 +270,7 @@ copy_stack_data (MonoThreadInfo *info, MonoStackData *stackdata_begin)
 #endif
 
 static gpointer
-mono_threads_enter_gc_safe_region_unbalanced_with_info (MonoThreadInfo *info, MonoStackData *stackdata);
+mono_threads_enter_gc_safe_region_unbalanced_with_info (MonoThreadInfo *info, MonoStackData *stackdata, MonoContext *ctx);
 
 gpointer
 mono_threads_enter_gc_safe_region_internal (MonoStackData *stackdata)
@@ -282,15 +286,18 @@ mono_threads_enter_gc_safe_region (gpointer *stackpointer)
 	return mono_threads_enter_gc_safe_region_internal (&stackdata);
 }
 
-gpointer
+gpointer MONO_NO_OPTIMIZATION
 mono_threads_enter_gc_safe_region_with_info (MonoThreadInfo *info, MonoStackData *stackdata)
 {
 	gpointer cookie;
 
+	MonoContext ctx;
+	MONO_CONTEXT_GET_CURRENT (ctx);
+
 	if (!mono_threads_is_blocking_transition_enabled ())
 		return NULL;
 
-	cookie = mono_threads_enter_gc_safe_region_unbalanced_with_info (info, stackdata);
+	cookie = mono_threads_enter_gc_safe_region_unbalanced_with_info (info, stackdata, &ctx);
 
 #ifdef ENABLE_CHECKED_BUILD_GC
 	if (mono_check_mode_enabled (MONO_CHECK_MODE_GC))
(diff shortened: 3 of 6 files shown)


jaykrell:
Please consider adding the reduced test case, and I guess, including running it on a separate thread for..I don't know..a minute seems too long.

We should have some collection of tests, that run nightly or weekly, but not per PR and maybe not even daily. Then this test could run for, say, 5 minutes, and be affdordable

PR/CI tests should be well under a second each if possible.

filipnavara:
@jaykrell This is not supposed to be a final fix. I just want to ensure that it fixes the original failures on the CI machines. I do agree that the stress test case should be added.

EgorBo:
@filipnavara hm.. we need a stress-test mode for CI where we run each test suite (or specific one) several times and expect 0 failures.

filipnavara:
@EgorBo there are three things we can test here:
1) The original reduced test case run in GC stress mode.
2) The actual CoreFX/CoreCLR tests in GC stress mode.
3) A purpose built test case for this exact bug which is quick. We can handcraft some code that will set all non-volatile registers to known values, enter the GC safe mode through the Mono API, exit it, and check the saved values in the context against the known good values.

filipnavara:
We can setup custom (parametrized) pipelines that are optional and trigger them with `/azp run 《name》`

lambdageek:
> The compiler is free to save registers to the stack and reuse them for other
variables. If the context is captured at that point we may inadvertently create
GC hole because the stack is later overwritten when returning out of
mono_threads_enter_gc_safe_region_* calls and hence a reference won't
be captured in neither the registers nor the stack.

Maybe this is why `copy_stack_data` was enabled even on platforms where we could capture the context.  We missed this and https://github.com/mono/mono/commit/ad90ebd3abf06a33174ebf0979950025e7381f50 was too aggressive /cc @lateralusX

jaykrell:
Undoing that does seem to fix it, at least. Could be there are other ways to fix.

lateralusX:
Yes, copy_stack_data would take an address in callers frame so it would cover this case (as well as the solution used on platforms that can not take full context, only having access to __builtin_unwind_init). I assume it is only needed for GC safe transitions since GC unsafe should only be suspended at known locations either at enter/exit or during mono_threads_state_poll_with_info and then we should see the complete context and stack on platforms supporting these features (and where copy stack data has been disabled). The fix would be to at least re-enable the copy stack data for enter/exit GC safe regions.

lateralusX:
Another alternative would be to just copy the stack and not include the context on platforms that already supports taking the context before suspending a thread. That would handle the issue with compiler optimizations as part of native functions called when enter safe/unsafe mode. Doing it that way would reduce the amount of stack data that needs to be copied on every transition, and not include the additional context data on platforms that already capture the context as part of suspending a thread.

lateralusX:
A quick fix to see if this solves the issue you should be able to use --enable-copy-stack-data as default on all platforms.

jaykrell:
How is the complete context on the stack, w/o `__builtin_unwind_init`?`

lateralusX:
So for GC safe we will need the context on stack, but for unsafe we should only suspend in known places where we already take context on platforms supporting it and not currently using copy_stack_data.

lateralusX:
On platforms not supporting __builtin_unwind_init or using alternative implementation (as done on Windows), the context, won't be added as part of copy_stack_data.

jaykrell:
I find the names gc safe/unsafe confusing.
I guess this is, preemptive and cooperative suspend?
Preemptive basically doesn't have to do anything. Suspend + get thread context (registers) is reliable and "automatic", from the point of view of the suspended thread.
But cooperative needs to stores its context kinda manually.
?

lateralusX:
So how could I test this, just do a PR with the alternative change and enable the test, System.Security.Cryptography.Rsa.Tests.RSAKeyFileTests.ReadEncryptedRsa16384. Enough to run it on CI or does it needs to be run many times locally in order to reproduce?

lateralusX:
This is just cooperate and hybrid suspend. When using preemptive suspend this code won't be used.

jaykrell:
It is "easy" and "quick" to test locally.
"Just" Ubuntu 16.04, with a very small repro, run it, repro rate is 100%, takes less than a minute each time. Success is running w/o error or exception. Failure is any exception.
It is more likely gcc version than Ubuntu version, they just have, you know, preconfigured coincidences.
There might very well be a Windows repro, but..the divergent workflows bother me..

filipnavara:
The repro is in issue #16859. It's quite sensitive to specific environment (Ubuntu 16.04) but once you get that up and running it's very reproducible. I will be happy to test any fix or provide access to Azure machines that are setup with the test case and correct software versions.

lateralusX:
Thanks @filipnavara. I will try to run this locally to see if I get a repro and fix. As already said above seems like the copy_stack_data is needed for safe mode, since stack gets lost in that case upon return.

lateralusX:
Fixing the GC safe to use copy_stack_data solves the problem.

jaykrell:
Is that equivalent to undoing "that" change? I guess I just wait and see the PR.

lateralusX:
So far I'm still keeping the optimization for GC unsafe mode transition, since when doing that transition we shouldn't have references in registers or on stack (not seen by us) that is not already pinned, unless we are already in unsafe mode, and then we can only be preempted at known locations where already take full context + wait so all stack should be visible at that point, so partially undoing the PR.

lateralusX:
https://github.com/mono/mono/pull/17150

lambdageek:
> I find the names gc safe/unsafe confusing.

"GC Safe": it is safe for the GC to modify this thread's view of the heap.
"GC Unsafe" it is not safe for the GC to modify this thread's view of the heap.

If a thread is in GC Safe mode, it does not have any references to unpinned managed references.  So the thread can keep running as long as it doesn't try to interact with Mono.

If a thread is in GC Unsafe mode it is doing something that would interfere with the GC - maybe it's accessing a heap object that the GC needs to move, maybe it wants to take the GC lock because it is trying to allocate. etc.

> I guess this is, preemptive and cooperative suspend?

Thread state transitions are independent of the mechanism that we use for physically suspending the thread.

By default if we're using fully preemptive suspend everywhere, we can avoid explicitly tracking GC Safe / GC Unsafe transitions - the GC can preempt the thread, suspend it, do anything that it needs to do and resume the thread whether it was in some third party native code (GC Safe) or whether it's in managed code.

For fully cooperative suspend we keep track explicitly of GC Safe / GC Unsafe transitions, and once a thread is in a GC Safe state, we allow it to continue running in parallel with the GC because it can't interfere.  When it tries to explicitly return to GC Unsafe mode, it will block until the GC is done.

For hybrid suspend we have to keep track of GC Safe / GC Unsafe transitions explicitly, but we relax the rules of what a thread can do in GC Safe mode because we will preempt it (So they can hold references to managed objects across a GC safepoint without explicitly using handles, for example).  GC Unsafe threads are still cooperatively suspended, however, because found that preempting threads when they're manipulating the runtime can be a source of bugs.

> Preemptive basically doesn't have to do anything. Suspend + get thread context (registers) is reliable and "automatic", from the point of view of the suspended thread. But cooperative needs to stores its context kinda manually?

This is two kind of unrelated ideas actually. We have several approaches to preemptive suspend.
1. If we're on a platform where there's a syscall that lets one thread suspend another thread _and_ we have the ability to get the context of that other thread (so Win32 with `SuspendThread` and `GetThreadContext`) we can: suspend the other thread, get its context, and conservatively scan its stack and registers.
2. If we're on a platform where there is a way to send signals to another thread and we can get the context of the current thread  We can: 1. `pkill` the mutator thread. 2. the thread runs a signal handler.  3. the signal handler saves the context (actually we could probably just spill the registers to the stack and only capture the SP, but I think, we have to worry about sigaltstack) and signals to the GC thread that it suspended itself. 4. the signal handler goes to sleep in some manner until the GC thread wakes it up.  5. the GC thread scans the saved context reads out the SP and scans the stack and registers.

On platforms (like watchOS or WASM) where we can't do either of those things we use cooperative suspend.

With coop suspend we only need the capability to spill the current thread's registers to the stack.  And we only suspend by periodically polling.  The GC Safe / GC Unsafe thing is an optimization in a sense - when a thread is in GC Safe the GC doesn't have to wait for the thread to reach a safepoint, it can ignore it instead.  
So when a thread reaches a safepoint, we: 1. spill the registers. 2 take a snapshot of the stack. 3. block on a semaphore that the GC will use to wake us up.

If a thread reaches a GC Safe transition we do a similar thing: 1. spill the registers, take a snapshot of the stack. 3. keep running until we reach an exit from GC Safe at which point if the GC is active, we block until the GC is ready to wake everyone up.

(I put a copy of this [in a public gist](https://gist.github.com/lambdageek/3b59172074070bee52654fa88c3c8251))

jaykrell:
The names are still in the way, in my head.
I already knew the two mechanisms to preemptively suspend.
That still leaves much to understand.

I understand now, the GC_SAFE mode, but it'd seem a rare-ish condition.
Syscalls are kinda it, but not entirely.
Syscalls can make callbacks to usermode, like win32k callbacks or I/O completion, and those could take managed data as parameters, and further manipulate in the callback.
Perhaps the callback can safely transition nestedly back to GC_UNSAFE?
I guess hybrid lets that work anyway, can still have managed references.
But fully cooperative GC_SAFE seems like, it almost doesn't exist.

Ok, so rererereading this several times.
The problem is what to do in the transition from GC_UNSAFE to GC_SAFE with hybrid.
And that such a transition does not guarantee GC will occur, will vastly outnumber it, must be fast.
However the later suspend can still be preemptive.
So it has a difficult time balancing fast because not needed vs. slower and needed.

And another question -- is hybrid, where, we will preemptively suspend like with SuspendThread or signal, but not get context "that way", not use GetThreadContext?
So the "spill" transitioning to GC_SAFE really has to be complete.
Do the registers have to be addressable, or just available in a conservative stack scan?

I think for most systems we can just hand code assembly to spill the nonvolatile registers.
And the function that does, can take a continuation, and jmp do it.
Exactly like @lateralusX did for NT/AMD64.
It can be overkill I guess and spill excess, but it will be a hot-ish path and shouldn't do too much.

jaykrell:
This phrase "take a snapshot of the stack" sounds bad.

lateralusX:
As soon as we do a preemptive suspend (either as part of normal preemptive suspend or hybrid suspend mode), we will take threads full context using GetThreadContext on Windows or similar on other platforms.

The spill when entering gc safe needs to be as complete to make sure any managed reference in registers are placed into the memory region scanned by GC. On platforms doing conservative stack scan, we will also see all stack before call to enter gc safe, meaning that any references on that part of the stack will already be scanned as apart of normal GC conservative stack scan. Also note the caller of enter gc safe can pass an adress to a local and copy_stack_data will copy the entier memory region, including the spilled registers. In most cases (at least within the runtime) that local is just in the frame above, but could in theory cover more of the callers stack.

The registered currently spilled in copy_stack_data is without order, so just scanned as regular memory region by GC. Context retrieved using Mono context (part of thread_state_init) is however structured where we know where each individual register is located and used in things like unwinding operations.

lateralusX:
> This phrase "take a snapshot of the stack" sounds bad.

That was one of the things I tried to eliminate in the original optimization. On platforms where we already take full context and do conservative stack scan, taking any snapshot of the stack should not be needed, but I missed out the case where we get clobbered registers in native code between call to enter gc safe and where we snap full context, so at least taking a copy of the spilled register is currently needed to cover that case. I believe we could rewrite this to work with the original optimization, but before doing that, or any rewrite in this area, I would like to have a better test framework, covering as many cases we can come up with, both used within the runtime, but also how the semi private API is used by our products. Initial optimization eliminated configurations and usage we are currently not aware of, by just disable copy_stack_data on platforms we have better control around, so that could still be an alternative.

Issue #17059, opened 2019-09-25
LLVM JIT doesn't use -place-safepoints pass


EgorBo:
Let's say I have a simple method:
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
static int Foo()
{
    int x = 0;
    for (int i=0; i<10; i++) 
        x++;
    return x;
}
```
As you can see it's pretty simple and LLVM should be able to replace it all with just `return 10;`
If I run it with `--llvm` (JIT) it compiles to:
```asm
0000000000000000 <loWorld_Program_Foo___>:
<BB>:1
   0:   50                      push   %rax
   1:   48 b9 68 fe 9c b0 57    movabs $0x7f57b09cfe68,%rcx
   8:   7f 00 00
   b:   48 83 39 00             cmpq   $0x0,(%rcx)
   f:   75 1a                   jne    2b <loWorld_Program_Foo___+0x2b>
  11:   ba 0a 00 00 00          mov    $0xa,%edx
  16:   48 be 38 44 60 02 00    movabs $0x2604438,%rsi
  1d:   00 00 00
  20:   85 d2                   test   %edx,%edx
  22:   75 33                   jne    57 <loWorld_Program_Foo___+0x57>
  24:   b8 0a 00 00 00          mov    $0xa,%eax
  29:   59                      pop    %rcx
  2a:   c3                      retq
  2b:   48 b8 30 44 60 02 00    movabs $0x2604430,%rax
  32:   00 00 00
  35:   ff 10                   callq  *(%rax)
  37:   ba 0a 00 00 00          mov    $0xa,%edx
  3c:   48 be 38 44 60 02 00    movabs $0x2604438,%rsi
  43:   00 00 00
  46:   85 d2                   test   %edx,%edx
  48:   75 0d                   jne    57 <loWorld_Program_Foo___+0x57>
  4a:   eb d8                   jmp    24 <loWorld_Program_Foo___+0x24>
  4c:   0f 1f 40 00             nopl   0x0(%rax)
  50:   83 c2 ff                add    $0xffffffff,%edx
  53:   85 d2                   test   %edx,%edx
  55:   74 cd                   je     24 <loWorld_Program_Foo___+0x24>
  57:   48 83 39 00             cmpq   $0x0,(%rcx)
  5b:   74 f3                   je     50 <loWorld_Program_Foo___+0x50>
  5d:   ff 16                   callq  *(%rsi)
[...]

PR #17116, opened 2019-09-29
[llvm] Use -place-safepoints in JIT mode too

EgorBo:
Fixes: https://github.com/mono/mono/issues/17059

Example:
```csharp
static int Test()
{
    int sum = 0;
    for (int i = 0; i < 42000; i++)
        sum++;
    return sum;
}
```
### Current behavior:
```llvm
define monocc i32 @"Program:Test ()"() #0 {
BB0:
  %const = bitcast i64 140647394572600 to i64
  %0 = inttoptr i64 %const to i64*
  %1 = load volatile i64, i64* %0, align 8
  %2 = icmp eq i64 %1, 0
  br i1 %2, label %BB2, label %POLL_BB1, !prof !1

POLL_BB1:                                         ; preds = %BB0
  %3 = load void ()*, void ()** @mono_threads_state_poll, align 8
  call coldcc void %3()
  br label %BB2

BB2:                                              ; preds = %POLL_BB1, %BB0
  br label %BB4

BB4:                                              ; preds = %CONT_BB4, %BB2
  %lsr.iv = phi i32 [ %lsr.iv.next, %CONT_BB4 ], [ 42000, %BB2 ]
  %exitcond = icmp eq i32 %lsr.iv, 0
  br i1 %exitcond, label %BB1, label %BB5

BB5:                                              ; preds = %BB4
  %4 = inttoptr i64 %const to i64*
  %5 = load volatile i64, i64* %4, align 8
  %6 = icmp eq i64 %5, 0
  br i1 %6, label %CONT_BB4, label %POLL_BB3, !prof !1

POLL_BB3:                                         ; preds = %BB5
  %7 = load void ()*, void ()** @mono_threads_state_poll.1, align 8
  call coldcc void %7()
  br label %CONT_BB4

CONT_BB4:                                         ; preds = %POLL_BB3, %BB5
  %lsr.iv.next = add nsw i32 %lsr.iv, -1
  br label %BB4

BB1:                                              ; preds = %BB4
  ret i32 42000
}
```
(loop was not optimized away due to early inserted safepoints)

### New behavior:
```llvm
define monocc i32 @"Program:Test ()"() #0 gc "coreclr" {
BB0:
  %0 = load volatile i64, i64* inttoptr (i64 139875651025208 to i64*)
  %1 = icmp eq i64 %0, 0
  br i1 %1, label %gc.safepoint_poll.exit, label %gc.safepoint_poll.poll.i, !prof !1

gc.safepoint_poll.poll.i:
  %2 = load void ()*, void ()** @mono_threads_state_poll
  call coldcc void %2() #1
  br label %gc.safepoint_poll.exit

gc.safepoint_poll.exit: 
  ret i32 42000 ;;; NOTE: return 42000!
}
```
(LLVM itself put safepoints after optimizations so the loop was optimized!)

Should significantly improve LLVM JIT performance in benchmarks (in non-preemptive mode)


EgorBo:
Oops, it supposed to be a draft PR.

mono/mini/llvm-jit.cpp
```diff
@@ -284,13 +285,14 @@ class MonoLLVMJIT {
...
 		initializeTarget(registry);
+		linkCoreCLRGC();
 
 		const char *opts = g_getenv ("MONO_LLVM_OPT");
 		if (opts == NULL) {
 			// FIXME: find optimal mono specific order of passes
 			// see https://llvm.org/docs/Frontend/PerformanceTips.html#pass-ordering
 			// the following order is based on a stripped version of "OPT -O2"
-			opts = " -simplifycfg -sroa -lower-expect -instcombine -licm -simplifycfg -lcssa -indvars -loop-deletion -gvn -memcpyopt -sccp -bdce -instcombine -dse -simplifycfg";
+			opts = " -simplifycfg -sroa -lower-expect -instcombine -licm -simplifycfg -lcssa -indvars -loop-deletion -gvn -memcpyopt -sccp -bdce -instcombine -dse -simplifycfg -place-safepoints";
```
filipnavara: Don't we need to add `-place-safepoints` unconditionally irrespective of `MONO_LLVM_OPT`?

EgorBo: Makes sense! Fixed

mono/mini/llvm-jit.cpp
```diff
@@ -284,26 +285,34 @@ class MonoLLVMJIT {
...
-			// see https://llvm.org/docs/Frontend/PerformanceTips.html#pass-ordering
-			// the following order is based on a stripped version of "OPT -O2"
-			opts = " -simplifycfg -sroa -lower-expect -instcombine -licm -simplifycfg -lcssa -indvars -loop-deletion -gvn -memcpyopt -sccp -bdce -instcombine -dse -simplifycfg";
-		}
+		if (opts == NULL)
+			opts = default_opts;
+		else if (opts[0] == '+') // Append passes to the default order if starts with '+', overwrite otherwise
+			opts = g_strdup_printf ("%s %s", default_opts, opts + 1);
+		else if (opts[0] != ' ') // pass order has to start with a leading whitespace
+			opts = g_strdup_printf (" %s", opts);
```
EgorBo: Unrelated change, just improved `MONO_LLVM_OPT` a bit (it's now possible to append passes to the default order instead of completely overwrite them.

mono/mini/mini-llvm.c
```diff
@@ -4676,6 +4676,56 @@ get_float_const (MonoCompile *cfg, float val)
 		return LLVMConstFPExt (LLVMConstReal (LLVMFloatType (), val), LLVMDoubleType ());
 }
 
+static void
+emit_safepoint_poll_jit (EmitContext *ctx)
+{
```
vargaz: Can't this be merged with the aot version ?

mono/mini/mini-llvm.c
```diff
@@ -8118,8 +8168,11 @@ emit_method_inner (EmitContext *ctx)
 		}
 	}
 
-	if (!cfg->llvm_only && cfg->compile_aot && mono_threads_are_safepoints_enabled () && requires_safepoint)
```
vargaz: The OP_GC_SAFE_POINT code should be removed.

filipnavara: Still used for the LLVM-only mode.

EgorBo: Yeah I initially removed it but then Filip reminded me the llvm-only case

mono/mini/mini-llvm.c
EgorBo: Ok, will try

EgorBo:
@monojenkins build failed

vargaz:
Also, wouldn't this increase memory usage because every method will have a safepoint function emitted ?

filipnavara:
@vargaz I thought so but apparently it gets inlined and the result is not worse than it was before. I didn't run it locally to prove this but the theory seemed plausible.

vargaz:
I mean every jit module is going to have that extra unused function, or does it get removed ?

EgorBo:
@vargaz I've just pushed a change,
now the function is deleted after optimizations (`-place-safepoints` is the last pass) and should not affect codegen size/time/memory anyhow.

Also, I've put "optnone" attribute on it so the optimizer doesn't need to waste time on it.

EgorBo:
Here is a diff for the change: https://www.diffchecker.com/DEiZVPkP
(Full Module IR after optimizations), the function is deleted.

EgorBo:
@filipnavara honestly, no idea why loaded llvm doesn't work (it failed with "symbol not found" something like that) I thought "loaded llvm" mode was removed.

EgorBo:
@monojenkins build failed

EgorBo:

EgorBo:
hm.. System.Runtime.Tests fail at some point with 
```
Cannot transition thread 0x7fa3af7b0700 from SELF_SUSPENDED with REQUEST_PULSE
```

lambdageek:
@EgorBo if this happens again, could you grab the stack trace - that's very suspicious it seems like a race condition somewhere
> `Cannot transition thread 0x7fa3af7b0700 from SELF_SUSPENDED with REQUEST_PULSE`

EgorBo:
@lambdageek it happens always in my branch 🙁 
stack-traces:
https://gist.github.com/EgorBo/d1b4caf4c92bf17dd8b6cb9d5a1fefad
or
https://gist.github.com/EgorBo/d6be13a0f8d12a2af0d90a090dc7355e

Maybe I emit safepoints where I should not?

EgorBo:
Smaller repro:
```csharp
using System;
using System.Threading;

namespace HelloWorld
{
    class Program
    {
        {
            ThreadPool.QueueUserWorkItem(_ => {
                GC.Collect();
            });

            for (int i = 0; i < 10; i++) {
                ThreadPool.QueueUserWorkItem(_ => {
                    while (true) {
                        var o = new object();
                    }
                });
            }

            Thread.Sleep(100000);
        }
    }
}
```

EgorBo:
I suspect I should ignore `cfg->method->wrapper_type == MONO_WRAPPER_ALLOC` methods (it somehow helps with the repro but the tests still crash)

EgorBo:
Fixed with `cfg->method->wrapper_type != MONO_WRAPPER_ALLOC` check (do not emit sp in allocators)

EgorBo:
@monojenkins build failed

lewing:
Where are we with this?

EgorBo:
@lewing the PR is finished.

SamMonoRT:
@EgorBo  - all clear to merge ?

PR #17120, opened 2019-09-30
[wasm][xunit test] Exclude RemoteExecutorTests in bcl profile

MaximLipnin:
`System/RemoteExecutorTests.cs` tests are excluded in the mobile test profiles. It makes sense to do the same for wasm.

Fixes https://github.com/mono/mono/issues/17038.


MaximLipnin:
@monojenkins build failed

MaximLipnin:
@monojenkins build failed

Issue #17040, opened 2019-09-25
[wasm][xunit tests] System.Net.Security.Tests.SslStreamAlpnTests - MissingMethodException: Default constructor not found


MaximLipnin:
According to recent `Linux WebAssembly` CI lane logs there are several `System.Net.Security.Tests.SslStreamAlpnTests` failures:
```
WASM: FAIL: System.Net.Security.Tests.SslStreamAlpnTests.SslStream_StreamToStream_DuplicateOptions_Throws
WASM: System.MissingMethodException: Default constructor not found for type System.Net.Security.Tests.SslStreamAlpnTests
WASM:   at System.RuntimeType.CreateInstanceMono (System.Boolean nonPublic, System.Boolean wrapExceptions) <0x3076f28 + 0x000a0> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at System.RuntimeType.CreateInstanceSlow (System.Boolean publicOnly, System.Boolean wrapExceptions, System.Boolean skipCheckThis, System.Boolean fillCache) <0x30718a0 + 0x0001a> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at System.RuntimeType.CreateInstanceDefaultCtor (System.Boolean publicOnly, System.Boolean skipCheckThis, System.Boolean fillCache, System.Boolean wrapExceptions, System.Threading.StackCrawlMark& stackMark) <0x30717b0 + 0x00036> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at System.Activator.CreateInstance (System.Type type, System.Boolean nonPublic, System.Boolean wrapExceptions) <0x3071678 + 0x0006a> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at System.Activator.CreateInstance (System.Type type, System.Boolean nonPublic) <0x3c1c290 + 0x0000a> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at System.Activator.CreateInstance (System.Type type) <0x3c1c258 + 0x00006> in <07d23935c0704ba392a4a5c8da243ff4>:0 
WASM:   at WasmRunner.Run () <0x30192d0 + 0x00656> in <4b89f01df6184c3c8dd5484ffef1c42a>:0 
WASM: System.Net.Security.Tests.SslStreamAlpnTests.SslStream_StreamToStream_Alpn_Success [6]
[...]

PR #17122, opened 2019-09-30
[wasm][xunit tests] Other fix for "MissingMethodException: Default constructor not found" issue

MaximLipnin:
Fixes https://github.com/mono/mono/issues/17040


MaximLipnin:
There are no `MissingMethodException: Default constructor not found` failures in `Linux WebAssembly` lane logs.

/cc @steveisok

Issue #17017, opened 2019-09-24
Out of bounds unhandled exception


AndreiDiaconu1:
## Steps to Reproduce

1. Using the file `Program.cs`:
```
using System;
using System.Security.Permissions;

class Test
{
    public static void Main()
    {
        SiteIdentityPermission sip = new SiteIdentityPermission("site*");
    }
}
```
2. `mcs Program.cs`
3. `mono Program.exe`

The problem I believe lies in the file `mcs/class/corlib/System.Security.Permissions/SiteIdentityPermission.cs`, in the function `bool IsValid (string s)`. When handling the wildcard character, there is an access ,`s[i + 1]`, which will go out out of bounds if the wildcard is the last character (there is no check that `i + 1` is in bounds).

### Current Behavior

It throws an unhandled  `System.IndexOutOfRangeException` exception.

### Expected Behavior

Should throw `System.ArgumentException`.

## On which platforms did you notice this

[ ] macOS
[ x ] Linux
[ ] Windows

**Version Used**:

version `6.0.0.334`

## Stacktrace

```
Unhandled Exception:
System.IndexOutOfRangeException: Index was outside the bounds of the array.
  at System.Security.Permissions.SiteIdentityPermission.IsValid (System.String s) [0x00033] in <b0dd258eb94b44a48f94659a34d3da02>:0 
  at System.Security.Permissions.SiteIdentityPermission.set_Site (System.String value) [0x00000] in <b0dd258eb94b44a48f94659a34d3da02>:0 
  at System.Security.Permissions.SiteIdentityPermission..ctor (System.String site) [0x00006] in <b0dd258eb94b44a48f94659a34d3da02>:0 
  at Test.Main () [0x00000] in <c4ca752b96e4474cbaf6732a7e6b9bbf>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.IndexOutOfRangeException: Index was outside the bounds of the array.
  at System.Security.Permissions.SiteIdentityPermission.IsValid (System.String s) [0x00033] in <b0dd258eb94b44a48f94659a34d3da02>:0 
  at System.Security.Permissions.SiteIdentityPermission.set_Site (System.String value) [0x00000] in <b0dd258eb94b44a48f94659a34d3da02>:0 
  at System.Security.Permissions.SiteIdentityPermission..ctor (System.String site) [0x00006] in <b0dd258eb94b44a48f94659a34d3da02>:0 
[...]

PR #17124, opened 2019-09-30
Re-ordered the wildcard IsValid check in SiteIdentityPermissions

steveisok:
We were throwing an IndexOutOfRangeException with a string like 'www.mono-project.com*' when an ArgumentException was expected.

Fixes https://github.com/mono/mono/issues/17017

### DIFF
--- a/mcs/class/corlib/System.Security.Permissions/SiteIdentityPermission.cs
+++ b/mcs/class/corlib/System.Security.Permissions/SiteIdentityPermission.cs
@@ -197,7 +197,7 @@ namespace System.Security.Permissions {
 				if (x == 42) {
 					// special case for wildcards (*)
 					// must be alone or first and followed by a dot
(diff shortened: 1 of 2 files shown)


steveisok:
@monojenkins build failed

mcs/class/corlib/System.Security.Permissions/SiteIdentityPermission.cs
```diff
@@ -197,7 +197,7 @@ private bool IsValid (string s)
 				if (x == 42) {
 					// special case for wildcards (*)
 					// must be alone or first and followed by a dot
-					if ((s.Length > 1) && ((s [i + 1] != '.') || (i > 0)))
```
lambdageek: It took me a long time to understand the comment above.  So if there's a '*', the string must be either '*' or '*.' or else it's invalid?

PR #16899, opened 2019-09-17
[merp] Output native library module name when 'whitelist all' mode is…

alexischr:
… enabled

Partially fixes https://github.com/mono/mono/issues/16689

PR #17128, opened 2019-09-30
[2019-08] [merp] Add API method that whitelists all native libraries

alexischr:
Support native call data from all native modules. Backport of  https://github.com/mono/mono/pull/16899/ and https://github.com/mono/mono/pull/15622


lambdageek:

lambdageek:
@monojenkins build

PR #17130, opened 2019-09-30
[merp] Use macOS version not Darwin version in MERP reports

### LINKED ISSUE #17004: [merp] crash reports have Darwin kernel version in OSVersion field
## Steps to Reproduce

1. Induce a crash from VSM (so merp is enabled)
2. Look at the merp payload available in the merp dialog

### Current Behavior

The parameters field currently looks something like this: 
```	"parameters" : 
        {
		"ApplicationBundleId" : "com.microsoft.visual-studio",
		"ApplicationVersion" : "8.5.0.114",
		"ApplicationBitness" : "x64",
		"ApplicationName" : "Visual Studio",
		"BlameModuleName" : "Mono Exception",
		"BlameModuleVersion" : "6.7.0.562 (master/db626fe9d37 Thu Sep  5 00:00:34 EDT 2
                 .....
		"OSVersion" : "18.7.0",
		"LanguageID" : "0x7f",
		"SystemManufacturer" : "apple",
		"SystemModel" : "MacBookPro15,1",
		"EventType" : "AppleAppCrash"
	}
```

Notice OS Version is `18.7.0` which is actually the kernel version.

### Expected Behavior

I expected the OS Version to be the macOS version, for example `10.14.6`.   There is no clear mapping of kernel version to OS version as some macOS versions share kernel versions.   

I would also be okay with capturing both versions!

[x] macOS
[ ] Linux
[ ] Windows

kdubau:
Fixes #17004

### DIFF
--- a/mono/utils/mono-merp.c
+++ b/mono/utils/mono-merp.c
@@ -393,8 +393,11 @@ mono_init_merp (const intptr_t crashed_pid, const char *signal, MonoStackHash *h
 	merp->moduleOffset = 0;
 
 	merp->uiLidArg = MONO_LOCALE_INVARIANT;
-
+#if defined (TARGET_OSX)
 	merp->osVersion = macos_version_string ();
+#else
+	merp->osVersion = kernel_version_string ();
+#endif
 
 	// FIXME: THis is apple-only for now
 	merp->systemManufacturer = "apple";
--- a/mono/utils/mono-merp.c
+++ b/mono/utils/mono-merp.c
@@ -86,17 +86,12 @@ macos_version_string (void)
 	/* macOS 10.13.6 or later */
 	if (!version_string) {
 		size_t size = 0;
-		if (sysctlbyname ("kern.osproductversion", NULL, &size, NULL, 0) < 0) {
-			buf[0] = '\0';
-			version_string = &buf[0];
+		if (sysctlbyname ("kern.osproductversion", NULL, &size, NULL, 0) < 0 || size > buf_size) {
+			/* if we couldn't get the size or if it needs more space that we have in buf, leave it empty */
+			version_string = "";
 			return version_string;
 		}
 
-		if (size > buf_size) {
-			buf[0] = '\0';
-			version_string = &buf[0];
-			return version_string;
-		}
 		if (sysctlbyname ("kern.osproductversion", (char*)buf, &size, NULL, 0) >= 0)
 			version_string = &buf[0];
 	}
--- a/mono/utils/mono-merp.c
+++ b/mono/utils/mono-merp.c
@@ -65,21 +65,40 @@ kernel_version_string (void)
 }
 
 static gboolean
-starts_with(const char *pre, const char *str)
+starts_with (const char *pre, const char *str)
 {
-    return strncmp(pre, str, strlen(pre)) == 0;
+    return strncmp (pre, str, strlen (pre)) == 0;
 }
 
 static const char *
 macos_version_string (void)
 {
+	/* Can't allocate in here, could be called from a signal handler in a
+	 * crashed process.
+	 */
 	static const char *version_string;
+	static char buf[256];
+	static const size_t buf_size = sizeof (buf);
+
+	if (version_string)
+		return version_string;
+
 	/* macOS 10.13.6 or later */
 	if (!version_string) {
-		char s[256];
-		size_t size = sizeof (s);
-		if (sysctlbyname ("kern.osproductversion", s, &size, NULL, 0) >= 0)
-			version_string = strdup(s);
+		size_t size = 0;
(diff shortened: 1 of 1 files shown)


mono/utils/mono-merp.c
```diff
@@ -64,6 +64,48 @@ os_version_string (void)
...
+macos_version_string (void)
+{
+	static const char *version_string;
+	/* macOS 10.13.6 or later */
+	if (!version_string) {
+		char s[256];
+		size_t size = sizeof (s);
+		if (sysctlbyname ("kern.osproductversion", s, &size, NULL, 0) >= 0)
+			version_string = s;
+		free(s);
```
leculver: Should you be freeing a string from the stack?

mono/utils/mono-merp.c
```diff
@@ -64,6 +64,48 @@ os_version_string (void)
...
+static const char *
+macos_version_string (void)
+{
+	static const char *version_string;
+	/* macOS 10.13.6 or later */
+	if (!version_string) {
+		char s[256];
+		size_t size = sizeof (s);
+		if (sysctlbyname ("kern.osproductversion", s, &size, NULL, 0) >= 0)
+			version_string = s;
```
leculver: This is making a _pointer_ _copy_ of the location pointed to by s, which is on the stack.  Returning version_string after this line will result in pointing to negative stack space, which will be overwritten eventually by the thread using this space.

You likely want strdup, which will 'leak' this memory but since it's so small it's likely fine.

kdubau:
@leculver I am not sure :) ... unmanaged code is foreign to me, I asked @alexischr (offline) to give me a close review for that reason.

mono/utils/mono-merp.c
kdubau: removed, thanks.

mono/utils/mono-merp.c
kdubau: Updated, thanks!

kdubau:
Applied @leculver's feedback. Thanks, Lee.

lambdageek:
@kdubau This is a great start. It needs a bit of work: I believe this won't compile on non-OSX right now, and also the memory management isn't quite right.

I'll take over the PR, if you don't mind.

kdubau:
@lambdageek that would be great! thanks!

alexischr:
@monojenkins build pkg

lambdageek:
@monojenkins build pkg

mono/utils/mono-merp.c
```diff
@@ -65,21 +65,40 @@ kernel_version_string (void)
...
+
 	/* macOS 10.13.6 or later */
 	if (!version_string) {
-		char s[256];
-		size_t size = sizeof (s);
-		if (sysctlbyname ("kern.osproductversion", s, &size, NULL, 0) >= 0)
-			version_string = strdup(s);
+		size_t size = 0;
+		if (sysctlbyname ("kern.osproductversion", NULL, &size, NULL, 0) < 0) {
+			buf[0] = '\0';
```
Therzok: Should this be buf[size]?

Therzok: Nope nevermind, this is the error case.

lambdageek: No, this is the error path.  If we can't even get the expected size of the product version value, set buf to `""` and return that.

lambdageek:
@monojenkins build failed

mono/utils/mono-merp.c
```diff
@@ -64,6 +64,61 @@ os_version_string (void)
 	return "";
 }
 
+static gboolean
+starts_with (const char *pre, const char *str)
```
CoffeeFlux: We know the length of constant strings, so pass it as an argument.
There's actually a macro for this (G_STRING_CONSTANT_AND_LENGTH), and though it's currently broken the fix is easy, I just haven't pushed it to one of my PRs yet. If you want to use the macro I can push the fix to this PR to get it in, or you can just hard-code the length, whatever works.

mono/utils/mono-merp.c
```diff
@@ -64,6 +64,61 @@ os_version_string (void)
...
+	static char buf[256];
+	static const size_t buf_size = sizeof (buf);
+
+	if (version_string)
+		return version_string;
+
+	/* macOS 10.13.6 or later */
+	if (!version_string) {
+		size_t size = 0;
+		if (sysctlbyname ("kern.osproductversion", NULL, &size, NULL, 0) < 0 || size > buf_size) {
```
CoffeeFlux: I realize the buffer being filled is an unlikely scenario, but unless you're positive that `size` will include the null terminator i would use `size >= buf_size` to be safe.

lambdageek:
@CoffeeFlux fixed the buf_size comparison and added string lengths

lambdageek:

PR #17131, opened 2019-09-30
Update MERP event type to MonoAppCrash

kdubau:
To enable backend analysis, we need to update our parameters to match our back end event type, which is MonoAppCrash.

It doesn't make sense to me that the previous implementation constructed the XML here in Mono, but depended on the user (VSMac in this case) to add the "correct" event type.  The parameters in the XML must match exactly what is expected for the given EventType, and so we should ensure they are always in-sync by defining them here.  Thus, I removed the usage of `eventType` in `mono_merp_enable` and hard coded the expected EventType.


lambdageek:
@monojenkins build pkg

mono/utils/mono-merp.c
```diff
@@ -359,7 +359,7 @@ mono_init_merp (const intptr_t crashed_pid, const char *signal, MonoStackHash *h
 	merp->systemManufacturer = "apple";
 	get_apple_model ((char *) merp->systemModel, sizeof (merp->systemModel));
 
-	merp->eventType = config.eventType;
+	merp->eventType = "MonoAppCrash";
```
lambdageek: This comes from the `eventType` argument to the call to `EnableMicrosoftTelemetry` in managed, which is called by VSfM. I would prefer that we update the event type there instead of hardcoding something in here.

leculver: The interesting bit of this is that the layout of the XML file here (with its specific parameters in a specific order) is deeply tied to the `EventType` given to Watson.  It actually doesn't make sense to have VSfM be able to control that value, since if you mis-match the two you will get weird behavior from Watson...which we don't want.

I think it would make more sense to remove the `EventType` parameter from the pinvoke that VSfM calls into Mono with to remove that parameter and actually hard code it here as a result...  Though changing that pinvoke signature means we have to be sure to update the two in parallel?

In any case, I leave it to you all to work out the right information here, but I did want to add some context around why it makes sense for mono to control the `eventType` here itself and not let VSfM change it: Because the bucketing parameters/ordering/values are deeply tied to the `eventType`...you can't drop any other name into `eventType` other than `MonoAppCrash` and expect anything other than terrible results.

kdubau: Glad you brought it up...

So that causes a bit of a problem - if the parameters provided do not match what the back-end expects for a given `eventType` it will get rejected.  I am not sure why the the previous implementation put the XML payload in Mono, and the event type in VSfM when they _need_ to match.  AFAIU, there is no a scenario where the `eventType` can be set to something different and the upload still work.

If I update the parameters here, then we run the risk of an old VSM (with the old event type) running on this newer Mono and all it's events rejected.  Same goes the other way, if we change it in VSM and it's running on Mono from before this change, all the events will be rejected.  And so on...

Of course, we can coordinate the changes to land simultaneously, but that doesn't stop creative users from mix-matching things. 

There is no good answer to be honest, because with my change, the `EnableMicrosoftTelemetry` now has a parameter for `eventType` that get's ignored. And changing that signature is probably worse than just ignoring the parameter. 

I decided that the best solution to minimize rejected events, and prevent this scenario in the future, was to set the event type in the same place (meaning Mono/MERP) that we construct the XML/parameters (and ignore the one set by VSM). But of course, I am open to other suggestions and discussion how to best go forward (even if you think we should leave it as is, it just doesn't make sense to me 😛)

kdubau: GH should have collaborative editing...

lambdageek: I see.  That makes sense. Thanks @leculver and @kdubau


@kdubau I'm going to push some updates to this PR as well:
1. Add a comment summarizing why we're hardcoding the event type.
2. Update `EnableMicrosoftTelemetry` to call the parameter `unused` instead of `eventType`
3. I'll update the internal call signatures (`EnableMicrosoftTelemetry_internal`) to drop the `eventType` parameter entirely.

kdubau: Sounds good, I'm still testing these changes and may have more depending how that goes.

lambdageek:
@monojenkins build

lambdageek:
@monojenkins build pkg

lambdageek:
Waiting on some more changes from @kdubau who is validating the data we're dumping against the backend.

kdubau:
After testing these changes with Watson, it looks like we may end up not changing the XML schema after all.  However, we will still need to change our EventType to `MonoAppCrash` and it still doesn't make sense for VSMac to be able to control this value.

So we should either keep this PR for the bits that @lambdageek added to cleanup the icalls and hardcode `MonoAppCrash`, and revert the XML changes.  

Or we can abandon entirely, and go change VSMac to set `MonoAppCrash`.

In my opinion, we've already done the work here to cleanup the icalls and remove the ability for an external program to set the event type, so we may as well keep it.

kdubau:
@monojenkins build pkg

kdubau:
@monojenkins build pkg

alexischr:

alexischr:
@monojenkins build pkg

kdubau:
@lambdageek @alexischr we are good to go on this, it has been tested and is looking good. 

We will need backports back to 2019-08

lambdageek:
@monojenkins build failed

akoeplinger:

lambdageek:
Since there's a corlib bump, I'll do the backports by hand for 2019-10 and 2019-008

Issue #16803, opened 2019-09-12
mono_gc_finalize_notify  spinning after process exit


Therzok:
## Steps to Reproduce

1. N/A, Chris Hardy might have some insight on slack
2. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/975740

### Current Behavior

VSMac process hangs

### Expected Behavior
VSMac process does not hang

## On which platforms did you notice this

[*] macOS
[ ] Linux
[ ] Windows

**Version Used**:

=== Mono Framework MDK ===

Runtime:
	Mono 6.4.0.179 (2019-06/3608a495413) (64-bit)
	Package version: 604000179

## Stacktrace

Something happens on process exit and the process spins indefinitely:
```
2246 objc_msgSend  (in libobjc.A.dylib) + 29  [0x7fff6f9a085d]
  2246 altstack_handle_and_restore  (in libmonosgen-2.0.1.dylib) + 71  [0x104917bc7]  exceptions-amd64.c:870
    2246 mono_handle_native_crash  (in libmonosgen-2.0.1.dylib) + 213  [0x1048b68d5]  mini-exceptions.c:3328
      2246 mono_dump_native_crash_info  (in libmonosgen-2.0.1.dylib) + 995  [0x10491dd93]  mini-posix.c:1083
        2246 ???  (in <unknown binary>)  [0x103ed5000]
          2246 _sigtramp  (in libsystem_platform.dylib) + 29  [0x7fff70f0ab1d]
            2246 mono_sigchld_signal_handler  (in libmonosgen-2.0.1.dylib) + 18  [0x104980252]  w32process-unix.c:1159
               2246 mono_gc_finalize_notify  (in libmonosgen-2.0.1.dylib) + 80  [0x104aa9be0]  mono-os-semaphore.h:167

```
```
Please paste the stack trace here if available.
```

marek-safar:
@BrzVlad any thoughts on this?

BrzVlad:
This doesn't look GC related. Looks like something else crashes, that I can't really tell from the stacktrace. Maybe MERP ? @alexischr Did you investigate the other vsmac crash related to MERP ?

lambdageek:
[...]

PR #17002, opened 2019-09-23
[crashing] Remove Mono signal handlers when starting to handle a crash

alexischr:
Fixes https://github.com/mono/mono/issues/16803

PR #17143, opened 2019-10-01
[2019-06][crashing] Remove Mono signal handlers when starting to handle a crash

lambdageek:
Fixes https://github.com/mono/mono/issues/16803

Backport of https://github.com/mono/mono/pull/17002


lambdageek:
cc @kdubau

@marek-safar This is pretty low-risk, IMO and will improve VSfM hangs after crashes.

PR #17150, opened 2019-10-02
Always do copy_stack_data on entering GC safe/unsafe.

lateralusX:
This reverts https://github.com/mono/mono/commit/ad90ebd3abf06a33174ebf0979950025e7381f50 that seems to be a little to aggressive in its optimization.

It also enables a flaky test in CoreFx test suite that seems to hit the case where optimization of native code in mono_threads_enter_gc_safe_region_unbalanced_with_info, lost a reference to a managed SafeHandle only kept in a register. That reference needs to be spilled as part of doing copy_stack_data, making sure it is still visible to GC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1340,7 +1340,6 @@ with_wasm_default=no
 with_bitcode_default=no
 enable_cooperative_suspend_default=no
 enable_hybrid_suspend_default=no
-enable_copy_stack_data_default=yes
 
 # For the sake of clearer error messages, these numbers should all be different from each other.
 INVARIANT_AOT_OPTIONS=nimt-trampolines=2000,ntrampolines=10000,nrgctx-fetch-trampolines=256,ngsharedvt-trampolines=4400,nftnptr-arg-trampolines=4000
@@ -5680,38 +5679,6 @@ fi
 AM_CONDITIONAL([ENABLE_HYBRID_SUSPEND], [test x$enable_hybrid_suspend != xno])
 
 dnl End of thread suspend policy
-
-dnl *******************************
-dnl *** Copy stack data support ***
-dnl *******************************
-
-dnl Disable copy_stack_data on platforms supporting full context and
-dnl having access to full stack of running threads.
-if test x$enable_copy_stack_data_default != xno; then
-	case $HOST in
-	X86 | AMD64)
-		if test x$target_osx = xyes; then
-			dnl Some host/target confusion, there's no host_osx (and
-			dnl host_darwin would be true on iOS not just macOS).
-			enable_copy_stack_data_default=no
-		elif test x$host_linux = xyes -o x$host_win32 = xyes; then
-			enable_copy_stack_data_default=no
-		fi
-		;;
-	esac
-fi
-
-AC_ARG_ENABLE(copy_stack_data, [ --enable-copy-stack-data     Enable copy_stack_data feature for hybrid/cooperative suspend scenarios, (defaults to yes)], [], [enable_copy_stack_data=default])
-
-if test x$enable_copy_stack_data = xdefault; then
-   enable_copy_stack_data=$enable_copy_stack_data_default
-fi
-
-if test x$enable_copy_stack_data != xno ; then
-	AC_DEFINE(ENABLE_COPY_STACK_DATA,1,[Enable copy_stack_data feature for hybrid/cooperative suspend scenarios])
-fi
-
-dnl End of Copy stack data support
 
 dnl ***************************
 dnl *** feature experiments ***
--- a/mono/utils/mono-threads-coop.c
+++ b/mono/utils/mono-threads-coop.c
@@ -34,12 +34,6 @@
 #include <mono/utils/mach-support.h>
 #endif
 
-/* On platforms that doesn't have full context support (or doesn't do conservative stack scan), use copy stack data */
-/* when entering safe/unsafe GC regions. For platforms with full context support (doing conservative stack scan), */
-/* there is already logic in place to take context before getting in a state where thread could be conservative */
-/* scanned by GC. Avoiding doing additional stack copy will increse performance when entering safe/unsafe regions */
-/* when running in hybrid/cooperative supspend mode. */
-#if defined (ENABLE_COPY_STACK_DATA)
 #ifdef _MSC_VER
 // __builtin_unwind_init not available under MSVC but equivalent implementation is done using
 // copy_stack_data_internal_win32_wrapper.
@@ -49,9 +43,6 @@
 #define SAVE_REGS_ON_STACK do {} while (0)
 #else 
 #define SAVE_REGS_ON_STACK __builtin_unwind_init ();
-#endif
-#else
-#define SAVE_REGS_ON_STACK do {} while (0)
 #endif
 
 volatile size_t mono_polling_required;
@@ -207,7 +198,6 @@ copy_stack_data_internal (MonoThreadInfo *info, MonoStackData *stackdata_begin, 
 	state->gc_stackdata_size = stackdata_size;
 }
 
-#if defined (ENABLE_COPY_STACK_DATA)
 #ifdef _MSC_VER
 typedef void (*CopyStackDataFunc)(MonoThreadInfo *, MonoStackData *, gconstpointer, gconstpointer);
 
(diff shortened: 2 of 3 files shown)


BrzVlad:
I find stuff confusing. What is this copy_stack_data stuff, that saves stuff somewhere ? Where does it save it ? If we can do this on more platforms, even where we can't use MONO_CONTEXT_GET_CURRENT, then what is the point of thread_state_init ? Shouldn't we just use this API instead ? And what is the difference between this and thread_state_init ? Looks to me it just calls a function that stores the context somewhere and then it returns back. Which looks exactly like what thread_state_init does. Are we just fixing the issue by storing the context twice and for reasons, this means we don't lose any managed references ? Sounds hackish if that is so.

lateralusX:
So that was the observation I did when I set out doing https://github.com/mono/mono/commit/ad90ebd3abf06a33174ebf0979950025e7381f50, why do we take context twice, both in copy_stack_data as well as in thread_state_init? One of the problem with doing that optimization was the limited information around why it was implemented this way to begin with. So I was quite conservative in the optimization (but apparently not enough), to only optimize away copy_stack_data  on platforms where we have full context support and do conservative stack scanning. But as the issue indicates, I was missing the case where we get optimization in mono_threads_enter_gc_safe_region_unbalanced_with_info, since caller of this method (or one of the wrappers/macros calling it) seems to rely on the fact that all registers gets preserved when calling the method, but the context we take in thread_state_init happens to late, making the logic sensitive to optimizations in the native code between the call to mono_threads_enter_gc_safe_region_unbalanced_with_info or one of its wrappers/macros and thread_state_init.

Another confusing thing with enter_gc_safe/unsafe methods is that caller can pass a stackdata or stackpointer that might point to a bigger area of the stack, meaning that copy_stack_data will copy more than just the spilled registers, so that is also something that differs compared to our regular thread_state_init that just takes a snapshot of the context (with a specific layout).

This fix reverts part of that optimization back to how it used to work for enter GC safe, taking the context both in copy_stack_data as well as thread_state_init, fixing the identified issue.

I agree that we could look at a solution where we just take the context once on platforms where we can take full context and doing it where we today setup the MONO_STACKDATA to cover cases where registers gets clobbered in native code. I could potentially (as I said in this PR) look at that as a separate PR, or if the fix is not urgent, continue working on this PR.

BrzVlad:
I don't think the context information provided by thread_state_init is of any use in this case. Can we unwind anything from it given those are just random registers in a random function ? So for the coop checkpoint I think it makes sense to only use this copy_stack_data, which the GC scans to pin objects. Maybe we could clear the code up a bit around this observation and we also wouldn't care if MONO_CONTEXT_GET_CURRENT works or not on some platforms.

filipnavara:
> we use thread_state_init for different things

Not really, there are other overloads (eg. `thread_state_init_from_sigctx`) that are used elsewhere. The plain one is used on very few places. That's why I prototyped a different fix in #17110.

> Also copy_stack_data takes a snapshot of the entire stack (from the previous snapshot)

That sounds like very heavy overhead for this kind of transition. Also, the window for any problem is reduced but the capture of the stack and the registers still has to be somewhat atomic and internally consistent. I am not sure that's the case now. I'll read the PR and code again in the evening but I believe it may suffer from the same general problem (except that `__builtin_unwind_init` is one of the workarouds that makes it go away).

In any case we should make a test for this. My idea of the test is the following:
1) P/Invoke into native code
2) In the native function exit GC safe mode through Mono API
3) Set all callee preserved registers to known values
4) Enter GC safe mode again
5) Verify that the known values were captured in the context (or copied stack)

lateralusX:
The current copy_stack_data takes a copy of stack from where MONO_STACKDATA is captured and where copy_stack_data gets called, spilling the registers in the process, so it will see temporary stores done by compiler optimization. All that happens on the stack before calling macros like ENTER_GC_SAFE, will be covered by the conservative stack scanning as well.

filipnavara:
@lateralusX On paper that sounds correct and that it should work. I just spent too much time on this one to just accept it without some scrutiny. The previous code was broken even without optimization, it just happened to trigger different register allocation (`ebx` vs `edi`) that matched the generated code (`SafeHandle` stored in `edi`). It took 12 days to find the root cause of the problem so I believe investing some effort in additional testing and review is worth the time.

filipnavara:
LGTM, it seems to resolve the issue. I would still love to have a test for it since it's really easy to screw up and really hard to debug.

lateralusX:
I have given this some additionell thoughts. I will revert https://github.com/mono/mono/commit/ad90ebd3abf06a33174ebf0979950025e7381f50 as part of this PR to get back to known behavior (and aslo enable the failing test in corefx test suite). After that I will look into writing a low level test suite covering the various combinations of enter/exit safe/unsafe that we have within the runtime, but also covering the semi private API used by products, making sure we have a good test framework around this feature. That will be an attempt to collect as much knowledge as possible around these scenarios making sure any future attempt to change/rewrite/optimize won't hit similar or other issues.

jaykrell:
I still find the names of things confusing. Perhaps the overall architecture.
I think I understand, that in hybrid suspend, the thread is sometimes going to be cooperatively suspended, and sometimes preemptively suspended. When it is open for preemptive suspend, it doesn't have to do anything special. When it cooperatively suspended, it must save its registers itself somewhere, where the GC can see them, at least for pinning, perhaps for movement.

I think the way this could work is as follows:


managed
native_poll
native_definitely_will_suspend


managed occasionally either checks a variable,
or makes a call.

Let's model first the "make a call to poll" even
though that isn't ideal.

We can assume, "native codegen spills registers
somewhere eventually", and while "eventually"
can be "never", we can force.

So something like this:


native_poll shall take an address of a local from managed.
natve_poll when not suspending, just return, nothing done.

native_poll when deciding to really suspend, shall call
native_definitely_will_suspend.


native_definitely_will_suspend shall do whatever it takes
to get the compiler to spill all nonvolatile registers.

On most platforms (except watch and wasm), this can be
posix getcontext or setjmp.
Or _unwind_init.
(Posix getcontext is kinda deprecated, and overkill, it probably
has volatiles.)
Win32 RtlCaptureContext works, but is also overkill.
Win32 setjmp works and is minimal and sufficient.

This is if the registers need to be seen, but not changed.

native_definitely_will_suspend shall then call 
native_definitely_will_suspend2
which shall pass an address of local to
native_definitely_will_suspend3

which then will actually wait to be unsuspended.

The registers will be between the address of a local in native_definitely_will_suspend3
and the parameter passed from managed.

Do a conservative scan of this relatively narrow region to find them.

Maybe I have some excess frames here.
Maybe alloca(1) can replace some of the frames.

Does this work?
And suffices?

Do we just need to find a small conservative area of stack that will
have the registers, or must they be changable?

If the registers need to be precisely found, and updated, this does not work.


Making the poll be inline in managed is a simple transform of this.

lateralusX:
@lambdageek wrote some more info here:
https://github.com/mono/mono/pull/17110#issuecomment-537564274

lateralusX:
The poll is called when needed (both from managed as well as native code). When it decide to preempt a thread it will take context using existing context mechanism on platforms supporting it (or on platforms not supporting that, it still have registers in the slot copied by copy_stack_data when entering the curent mode) and wait for resume. At that point, on platforms with full context support we will see the complete stack + registers, on platforms without that support, we only see what copy_stack_data stored in specific memory region scanned by GC, upon entering the existing mode.

The copy_stack_data already does something similar to your proposal. Caller will pass an address to local and copy_stack_data will spill registers to stack and then copy the complete region between current stack pointer and the address to local  passed to the function.

Also note that when entering safe mode, thread won't be preempted, so it can run in parallel with GC until switching mode (at that point it will be preempted if GC is running). That means that the registers spilled to stack as part of copy stack data is no longer available and most likely overwritten, that's why it gets copied into a separate memory region. Since we won't know if we will need that information in future when entering safe mode, it will always be stored in specific memory region scanned by GC.

I believe reverted implementation does the right thing, but what we could look into is optimizations around not doing stack copy when entering unsafe mode, only take context once and not doing it in both copy_stack_data as part of register spill and then in thread_state_init on platforms supporting Mono context etc.

lateralusX:
@monojenkins rebuild failed

lateralusX:
@monojenkins backport 2019-08

PR #17153, opened 2019-10-02
Ignore python files in tools/offsets-tool-py/offtool

MaximLipnin:
During the wasm build I observe many python files in tools/offsets-tool-py/offtool folder.


MaximLipnin:
see https://github.com/mono/mono/pull/17168

PR #17159, opened 2019-10-02
[mixed] Add interp entry wrappers for devirtualized calls as well.

vargaz:


vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

vargaz:
@monojenkins build failed

sdks/wasm/print-table.py
```diff
@@ -6,7 +6,7 @@
 
 import sys
 
-prefix=" (elem (get_global $tableBase) "
+prefix=" (elem (i32.const 1) "
```
kg: What's this?

vargaz: Its a tool to print out the indirect function table from a wasm text file so if you print out a function pointer in C code, you can find the corresponding wasm function.

vargaz:
@monojenkins build failed

sdks/wasm/print-table.py
kg: I just meant specifically that it wasn't clear why you were replacing a get_global with a constant, since it didn't look like you changed the global. Mostly curious

vargaz: The wasm generated by the recent emscripten is different for the table:
(elem (i32.const 1) $mono_wasm_invoke_js $mono_wasm_invoke_js_marshalled...

vargaz:
@monojenkins build failed

PR #17162, opened 2019-10-03
Now IsExpired property for FormsAuthenticationTicket compares two dates with same kind (UTC)

### LINKED ISSUE #17161: IsExpired value returning wrong result when ExpirationDate is a DateTime with Kind UTC.
When calculating if a FormAuthenticationTicket is Expired or not, it's only working properly if value set for ExpirationDate has Local Kind. If you set the value for ExpirationDate with an UTC value, it's returning a wrong value saying that Ticket is Expired even if it's not.

https://github.com/mono/mono/blob/d8040ba86520ae584f36628bd7eece1558970f41/mcs/class/System.Web/System.Web.Security/FormsAuthenticationTicket.cs#L190

dbuenor:
Now IsExpired property compares two dates (actual datetime and expiration datetime) with same kind (UTC) due to converting both dates to UTC. Fixes #17161

### DIFF
--- a/mcs/class/System.Web/System.Web.Security/FormsAuthenticationTicket.cs
+++ b/mcs/class/System.Web/System.Web.Security/FormsAuthenticationTicket.cs
@@ -187,7 +187,7 @@ namespace System.Web.Security
 		}
 
 		public bool Expired {
-			get { return DateTime.Now > expiration; }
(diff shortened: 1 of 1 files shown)


akoeplinger:
@monojenkins build

PR #17164, opened 2019-10-03
Set default value for --host in autogen.sh on cygwin when missing.

lateralusX:
Previous attempt to do this as part of configure.ac didn't cover all cases since the auto generated parts of configure script already setup some values before we had the chance to set --host-alias. In order to fix this, the check for --host is moved to autogen.sh, making sure we pass --host to configure step, if missing.

NOTE, this will only apply to cygwin build that, by default, wants a different host/target toolchain compared to what config.guess delivers. By default, we will use --host=x86_64-w64-mingw32 when running autogen.sh without passing --host parameter on cygwin. NOTE, there is already a error on cygwin for this case in configure.ac, so it is currently not possible without passing --host to configure.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -160,7 +160,19 @@ if test x$MONO_EXTRA_CONFIGURE_FLAGS != x; then
 	echo "MONO_EXTRA_CONFIGURE_FLAGS is $MONO_EXTRA_CONFIGURE_FLAGS"
 fi
 
-conf_flags="$MONO_EXTRA_CONFIGURE_FLAGS --enable-maintainer-mode --enable-compile-warnings" #--enable-iso-c
+host_conf_flag=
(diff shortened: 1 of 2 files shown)


lateralusX:
@monojenkins rebuild failed

jaykrell:
What about people running configure, not autogen?

jaykrell:
Maybe that can be done early in configure.ac before defaulting?

lateralusX:
That's what I did before, but appears that it still sets some variables pretty early so I couldn't correctly default early enough in configure.ac, that's why I moved it into autogen.sh. We could try to hunt down exactly what variables that needs to be adjust in configure, if doing it early in configure.ac, but then we might be broken due to changes.

jaykrell:
Doing it in autogen.sh is ok, doesn't hurt in the long run, whether or not configure.ac does it also.

